PR #653, opened 2022-10-01
feat(frontend): Discard changes confirmation modal

fatonramadani:
https://user-images.githubusercontent.com/456655/193413807-22db3198-8e59-47c5-bb91-f80f498c24a0.mov


frontend/src/lib/components/common/confirmationModal/UnsavedConfirmationModal.svelte
```diff
@@ -0,0 +1,47 @@
...
+	import { dirtyStore } from './dirtyStore'
+
+	let navigationState: { from: URL; to: URL | null; cancel: () => void } | undefined = undefined
+	$: open = Boolean(navigationState)
+
+	beforeNavigate((newNavigationState) => {
+		if (
+			!navigationState &&
+			$dirtyStore &&
+			newNavigationState.to?.pathname !== newNavigationState.from.pathname
```
fatonramadani, line 14: We only cancel routing when from pathname is different from to pathname: In the script/flow builder, the pathname stays the same but the "step" query parameters changes

frontend/src/lib/components/ScriptBuilder.svelte
```diff
@@ -40,6 +42,7 @@
 
 	async function editScript(): Promise<void> {
 		try {
+			$dirtyStore = false
```
rubenfiszel, line 46: you also want to initialize it on adding a new script

frontend/src/lib/components/common/confirmationModal/dirtyStore.ts
```diff
@@ -0,0 +1,3 @@
+import { writable } from 'svelte/store'
+
+export const dirtyStore = writable<boolean>(false)
```
rubenfiszel, line 3: I think you do not gain much but having a common store for both flow and and script. Passing the store as a bind:props to the component should be good enough no ?

rubenfiszel: Ah nevermind, you're doing it at the Editor level so it has to be global, gotcha

frontend/src/lib/components/common/confirmationModal/UnsavedConfirmationModal.svelte
rubenfiszel: very clever!

frontend/src/lib/components/FlowBuilder.svelte
```diff
@@ -53,6 +55,7 @@
 	async function saveFlow(): Promise<void> {
```
rubenfiszel, line 50: you're missing initialization at false when one is newly editing or adding a script (it will be in the components/route wrapping FlowBuilder)

frontend/src/lib/components/common/confirmationModal/UnsavedConfirmationModal.svelte
```diff
@@ -0,0 +1,47 @@
...
+			newNavigationState.cancel()
+		}
+	})
+
+	onDestroy(() => {
+		$dirtyStore = false
+	})
+</script>
+
+<ConfirmationModal
```
rubenfiszel, line 26: can we make confirmation modal such that:
Pressing enter confirms it
Pressing escape cancels it

fatonramadani: Done!

frontend/src/lib/components/ScriptBuilder.svelte
fatonramadani: This code is used to create and to update a script. `ScriptBuilder` component is used is both routes

frontend/src/lib/components/FlowBuilder.svelte
rubenfiszel: @fatonramadani that is still the case

frontend/src/lib/components/ScriptBuilder.svelte
rubenfiszel: Yes, that's correct for saving but not fot initialiazing. At opening of a new script or editing an existing script, you want to initialize $dirtyStore to false also.

frontend/src/lib/components/FlowBuilder.svelte
fatonramadani: Done

frontend/src/lib/components/ScriptBuilder.svelte
fatonramadani: Done

PR #690, opened 2022-10-06
fix(frontend): Change button click propagation

adam-kov:


frontend/src/lib/components/common/button/Button.svelte
```diff
@@ -63,8 +63,10 @@
 	}
 
 	function onClick(event: MouseEvent) {
-		dispatch('click', event)
-		if (href) goto(href)
```
rubenfiszel: Why can you remove the goto ?

adam-kov: If an href is set, it'll be a property on the element itself and the browser will register the navigation. Also, I don't think target="_blank" works with goto.

rubenfiszel: Adding the goto is the fix that I to do to avoid buttons breaking up navigation by going to a new page instead of using svelte internal navigation. Propagating href seems to not work as you expect it to.

PR #707, opened 2022-10-10
feat(frontend): Loading placeholder

### LINKED ISSUE #494: feat: pretty loading placeholders for table loading data and relevant context to avoid content shift

adam-kov:
closes #494

Added loading skeleton to the following pages:
- Home
- Runs
- Run Detail
- Scripts
- Script Detail
- Flows
- Flow Detail
- Schedules
- Variables
- Resources

image


rubenfiszel:
@adam-kov Looks great. Could we add the skeleton to Home, Runs, Run Detail, Scripts, Flows, Flow Detail,  Schedules, Variables, Resources

frontend/src/lib/components/common/skeleton/SkeletonElement.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import { skeleton } from '../../../directives'
+	import { HEIGHT_UNIT, type ISkeletonElement } from './model'
+
+	export let element: ISkeletonElement
+</script>
+
+<div
+	use:skeleton={true}
```
fatonramadani: Couldn't we replace the whole use:skeleton with just the three classes:
`classes="animate-pulse bg-blue-200 rounded"` ?

adam-kov: Yes we could, but abstracting it with a directive will make it easier to update if we ever need to and easier to toggle the animation as well.

fatonramadani: Everything is wrapped around an if statement in the Skeleton component. I don't think the code in the directive that removes the classes is ever reached.

adam-kov: The `SkeletonElement` uses a constant `true` on the directive because the only purpose of the component is to indicate loading. The toggling could be useful when it's used anywhere else in the app. Sure it does the same as applying the classes but could be a little more useful in the future.

rubenfiszel: I don't understand your comment @adam-kov

frontend/src/routes/index.svelte
```diff
@@ -59,6 +68,7 @@
 			<ScriptGettingStarted />
 
 			<div class="mt-6 mb-2 text-md font-bold text-gray-900 ">Latest scripts:</div>
+			<Skeleton loading={loading.scripts} layout={[1, [12.25, 12.25, 12.25]]} />
```
rubenfiszel: nit: those skeleton cards are slightly too tall on my end.

adam-kov: How tall are the cards on your screen?

PR #726, opened 2022-10-12
feat(backend): supercache for python heavy dependencies in alpha

mrl5:
## disclaimer
this is quick and dirty PoC that allows to evaluate if this is an approach that we should take further

### Background

1. If python script requires external packages `requirements.txt` file is:
   * first created: https://github.com/windmill-labs/windmill/blob/9acee22b1fc0b4eb82a1b47bc62598fe5af076e1/backend/src/worker.rs#L1300
   * then injected: https://github.com/windmill-labs/windmill/blob/9acee22b1fc0b4eb82a1b47bc62598fe5af076e1/backend/src/worker.rs#L1061

2. It is possible to do remote code execution (aka RCE) by putting malicious
   code into `setup.py`:
   * example: https://gist.github.com/noobpk/d48dde95815e5024dc2ab5044e622cc1
   * BDFL confirmation: https://discord.com/channels/930051556043276338/930051556043276341/1024440420660416594

3. To prevent/limit the impact of potential `#2` - on each job execution `nsjail`
   sandboxed environment is created (unless disabled) so that dependencies
   could be installed before actual job execution:
   * https://github.com/windmill-labs/windmill/blob/9acee22b1fc0b4eb82a1b47bc62598fe5af076e1/backend/src/worker.rs#L1081
   * https://github.com/windmill-labs/windmill/blob/9acee22b1fc0b4eb82a1b47bc62598fe5af076e1/backend/src/worker.rs#L1121-L1122

4. Each time `pip install` is invoked `setup.py` is executed (or any other
   "build backend" that's used by given package - more context here:
   https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml)

5. For `pandas` and `numpy` step `#4` is time consuming
   * https://github.com/pandas-dev/pandas/blob/main/setup.py
   * https://github.com/numpy/numpy/blob/main/setup.py

### proposed solution
use venv to share trusted python packages


mrl5:
used this for testing:
```python
import numpy as np

def main():
    print(dir(np))
    a = np.arange(15).reshape(3, 5)
    return a
```

mrl5:
benchmark:
```
current:
cold start 3.88030477s
2.183972421s
2.213488127s
2.191136011s
2.168564491s
2.306767901s
```

venv cache of "heavy deps":
```
cold start 6.554596153s
387.90515ms
385.855029ms
379.807848ms
384.722128ms
393.903702ms
```

PR #731, opened 2022-10-13
Initial Benchmarking Tool

HurricanKai:
Runs multiple deno workers to fill up the work queue.
Scrapes metrics endpoint periodically & exports metrics to influxdb for easy analysis.

I've had little success in actually getting windmill to max out any cores, but I've verified that the queue is actually filling up, so I'm confident this works.
New metrics added to windmill are automatically picked up (including labels).


HurricanKai:
Rebased per GitHubs request

HurricanKai:

benchmarks/main.ts
```diff
@@ -0,0 +1,325 @@
...
+        new URL("./scraper.ts", import.meta.url).href,
+        {
+          type: "module",
+        }
+      );
+
+      const export_map: Map<
+        string,
+        Map<string, { val: number; samples: number }>
+      > = new Map();
```
rubenfiszel: I usually use Record and very seldom Map. Not that there is anything wrong with Map but curious to know if this is an habit from another language or you had strong reasons for it.

benchmarks/main.ts
```diff
@@ -0,0 +1,325 @@
...
+      // exports is true | string[] | undefined. the first condition checks that it's not undefined, the second condition checks it's the array
+      if (exports && exports !== true) {
```
rubenfiszel: Array.isArray insteaf of !== true ?

benchmarks/main.ts
```diff
@@ -0,0 +1,325 @@
...
+          token: influxToken,
+        });
+        console.log("influxDB enabled");
+
+        const orgsAPI = new OrgsAPI(influxDB);
+        const organizations = await orgsAPI.getOrgs({ org: influxOrg });
+        if (
+          !organizations ||
+          !organizations.orgs ||
+          !organizations.orgs.length
```
rubenfiszel: !organizations?.orgs?.length

benchmarks/main.ts
```diff
@@ -0,0 +1,325 @@
...
+        }
+        const orgID = organizations.orgs[0].id;
+
+        const bucketsAPI = new BucketsAPI(influxDB);
+        try {
+          const buckets = await bucketsAPI.getBuckets({
+            orgID,
+            name: influxBucket,
+          });
+          if (buckets && buckets.buckets && buckets.buckets.length) {
```
rubenfiszel: buckets?.buckets?.[0]

benchmarks/main.ts
HurricanKai: Just applied Record in a few places where I used the more complicated indexer syntax, had forgotten about that type.
I guess Map can be replaced by Record[] sometimes, but I usually just prefer Map where keys are unique and I want to iterate over it.
Here it's awkward a few times since I have to use `new Map(Object.entries(obj))` and `Object.fromEntries` to go from/to map a few times.

benchmarks/main.ts
```diff
@@ -0,0 +1,324 @@
...
+  .option("-m --metrics <metrics:string>", "The url to scrape metrics from.", {
+    default: "http://localhost:8001/metrics",
+  })
+  .option("--influx-host <influx_host:string>", "The influx url to write to.")
+  .option("--influx-token <influx_token:string>", "The influx token to use.")
+  .option("--influx-org <influx_org:string>", "The influx org to write to.")
+  .option(
+    "--influx-bucket <influx_bucket:string>",
+    "The influx bucket to write to. Everything in the bucket will be deleted!!"
+  )
```
rubenfiszel: Do we really need influxDB, would it not be sufficient to do everything in-memory and generate a report from it ? I wonder if something like Danfo would do the trick: https://danfo.jsdata.org/getting-started

HurricanKai: No, influx is way overkill.
danfo looks great, I'll integrate it to replace the export functionality. It's a bit of an awkward implementation anyways & not as powerful as I'd like it to be.

HurricanKai: Couldn't get danfojs or similar to work in deno. Implemented the few things we need myself.

HurricanKai:
Added Mean/Stdev to JSON output. imo not particularly useful on histograms, but it works. Also added csv output for testing & in case we want to plot / do further postprocessing on results later.

backend/src/tracing_init.rs
```diff
@@ -67,6 +68,7 @@ pub fn initialize_tracing() {
         .unwrap_or(false);
 
     let env_filter = EnvFilter::from_default_env();
+    let env_filter = env_filter.add_directive(Level::INFO.into());
```
rubenfiszel: we respect RUST_LOG instead, so remove this

HurricanKai:
TypeScript suggests that `"type": "rawscript"` is invalid in flow modules, but it's what I get when exporting from windmill, and it seems to work just fine.

PR #735, opened 2022-10-14
More Metrics

HurricanKai:
This adds some basic metrics about the workers (via tokio_metrics new dependency)
It also adds some metrics regarding the queue size, from which queue length (outstanding jobs) and in-flight jobs can be calculated.


backend/src/lib.rs
```diff
@@ -249,8 +250,59 @@ pub async fn run_workers(
...
+
+        for interval in monitor.intervals() {
+            total_poll_count.inc_by(interval.total_poll_count);
+            total_poll_duration.inc_by(interval.total_poll_duration.as_secs_f64());
+
+            total_idled_count.inc_by(interval.total_idled_count);
+            total_idle_duration.inc_by(interval.total_idle_duration.as_secs_f64());
+
+            total_scheduled_count.inc_by(interval.total_scheduled_count);
+            total_scheduled_duration.inc_by(interval.total_scheduled_duration.as_secs_f64());
```
rubenfiszel: My intuition confirmed through reading the documentation is that you need to sleep here, otherwise you will loop contrinuously to update the counters

backend/src/jobs.rs
```diff
@@ -1233,6 +1233,25 @@ pub enum JobPayload {
...
+    .unwrap();
+    static ref JOBS_TOTAL_DELETED: prometheus::IntCounter = prometheus::register_int_counter!(
+        "jobs_total_deleted",
+        "Total number of jobs deleted from the queue."
+    )
+    .unwrap();
+    static ref JOBS_TOTAL_PULLED: prometheus::IntCounter = prometheus::register_int_counter!(
+        "jobs_total_pulled",
+        "Total number of jobs pulled from the queue."
+    )
```
rubenfiszel: I'm not sure of the usefulness of "total" here, especially because if there are multiple server apis, you would need to aggregate them to actually get the total number

backend/src/jobs.rs
```diff
@@ -1233,6 +1233,25 @@ pub enum JobPayload {
     RawFlow { value: FlowValue, path: Option<String> },
 }
 
+lazy_static::lazy_static! {
+    // TODO: these aren't synced, they should be moved into the queue abstraction once/if that happens.
+    static ref JOBS_PUSHED: prometheus::IntCounter = prometheus::register_int_counter!(
+        "jobs_pushed",
```
rubenfiszel: let's prefix all the server metrics with `server.` WDYT ?

backend/src/lib.rs
```diff
@@ -249,8 +250,60 @@ pub async fn run_workers(
...
-        }));
+        })));
     }
+
+    let monitor_task = async move {
+        // TODO: Also consider monitoring fast/slow poll here.
+        // fast/slow are decided at an arbitrary, defined, limit.
+
+        let total_poll_count = prometheus::register_int_counter!(
+            "workers_tokio_total_poll_count",
```
rubenfiszel: let's do "worker." instead of "workers_"

backend/src/worker.rs
```diff
@@ -101,6 +101,28 @@ pub struct WorkerConfig {
...
+    .unwrap();
+    static ref JOBS_ZOMBIE_DELETED: prometheus::IntCounter = prometheus::register_int_counter!(
+        "jobs_zombie_deleted",
+        "Total number of jobs deleted due to their ping timing out in an unrecoverable state."
+    )
+    .unwrap();
+    static ref WORKER_UPTIME_OPTS: prometheus::Opts = prometheus::opts!(
+        "worker_uptime",
+        "Total number of milliseconds since the worker has started"
+    );
```
rubenfiszel, line 120: that one we really need to label with worker name

rubenfiszel: but you do it so nvm :)

backend/src/jobs.rs
HurricanKai: sounds good, but I think the job/queue specific metrics should have their own category (as that logic is shared / is another component). maybe `queue_` instead of `jobs_`?

PR #741, opened 2022-10-15
feat(frontend): Update split panes

adam-kov:
Updated split panes using the `svelte-splitpanes` package. They are now inside the viewport at all times and overflows are scrollable.
image


frontend/src/lib/components/flows/propPicker/PropPickerWrapper.svelte
```diff
@@ -43,23 +43,19 @@
...
-					dispatch('select', detail)
-					$propPickerConfig?.onSelect(detail)
-					propPickerConfig.set(undefined)
-				}}
-			/>
-		</div>
-	</right>
-</HSplitPane>
+<Splitpanes>
+	<Pane minSize={20} class="relative p-4 !duration-[0ms]">
```
fatonramadani, line 47: why would we need to use `!duration-[0ms]`?

frontend/src/lib/components/flows/content/FlowModule.svelte
```diff
@@ -99,11 +99,22 @@
 		width,
 		threshold: FLOW_MODULE_WIDTH_THRESHOLD
 	})
+
+	afterUpdate(() => {
```
fatonramadani, line 103: This entire logic boils down to getting the height of the EditorBar height, which is a constant. 
We can directly fetch its height using a svelte binding on the component to avoid this afterUpdate logic.

frontend/tailwind.config.cjs
```diff
@@ -293,6 +295,37 @@ const config = {
...
+					border: 'none !important',
+					'&::before': {
+						backgroundColor: '#00000060 !important',
+					},
+					'&::after': {
+						content: 'none !important'
+					},
+				},
+				'.splitpanes--vertical>.splitpanes__splitter': {
+					width: '5px !important',
```
fatonramadani: 5px seems too big imho. I think we can set it to 3, set the handle to 1, and leave 1px on each side.
Screenshot 2022-10-15 at 23 42 27

frontend/src/lib/components/flows/FlowEditor.svelte
```diff
@@ -22,11 +21,9 @@
...
 			</div>
-		</left>
-		<right slot="right" class="h-full">
-			<div class="h-full overflow-auto bg-white ">
-				<FlowEditorPanel {initialPath} />
-			</div>
-		</right>
-	</HSplitPane>
+		</Pane>
+		<Pane size={75} minSize={40} class="!overflow-hidden">
```
fatonramadani: The class `!overflow-hidden`hides the fact that the Pane component is not sized properly, which has a little space that is not reachable by the user.

https://user-images.githubusercontent.com/456655/196009221-0667ff99-a003-4104-92f8-bf35829e1133.mov

frontend/src/lib/components/flows/propPicker/PropPickerWrapper.svelte
adam-kov: It was sliding in from the side everytime the component was mounted, which was awkward looking.

frontend/src/lib/components/flows/content/FlowModule.svelte
adam-kov: There might be more than one element above the `Splitpane` and we need to get the sum of the height of those elements. I've tried the way you mentioned but it didn't really work. I also think this is a lot cleaner because I didn't need to modify those components just to get their height.

frontend/tailwind.config.cjs
adam-kov: Yep, the looks of the splitter is only temporary.

frontend/src/lib/components/flows/propPicker/PropPickerWrapper.svelte
fatonramadani: I think the sliding is an issue linked to the fact that both Parent and child wrapper are not taking as much height as possible. In the previous iteration, I've fixed it by wrapping everything in a div that uses: "flex flex-1" to grow

fatonramadani (approved):

PR #769, opened 2022-10-18
feat: add result by id to fetch result from any node

rubenfiszel:


backend/src/jobs.rs
```diff
@@ -944,6 +945,75 @@ async fn get_job(
...
+            .fetch_optional(&db)
+            .await?;
+            if let Some(r) = r {
+                let value = r
+                    .flow_status
+                    .as_ref()
+                    .ok_or_else(|| Error::InternalErr(format!("requiring a flow status value")))?
+                    .to_owned();
+                parent_id = r.parent_job;
+                let status_o = serde_json::from_value::<FlowStatus>(value.to_owned()).ok();
```
HurricanKai: nit: looks odd, value is already `to_owned()`, why is another needed?

HurricanKai (dismissed):

PR #770, opened 2022-10-18
feat(frontend): Update progress bar

adam-kov:


frontend/src/lib/components/progressBar/ProgressBar.svelte
```diff
@@ -0,0 +1,85 @@
...
+	import { tweened } from 'svelte/motion'
+	import { cubicOut } from 'svelte/easing'
+	import { isLoop, type Progress } from './model'
+	import ProgressBarGeneralPart from './ProgressBarGeneralPart.svelte'
+	import ProgressBarLoopPart from './ProgressBarLoopPart.svelte'
+
+	export let steps: Progress = []
+	export let startIndex = 0
+	const length = 100 / steps.length
+	const percent = tweened(+(length * startIndex).toFixed(0), {
```
fatonramadani: Is the leading + usefull ? Everything seems to be numbers

frontend/src/lib/components/progressBar/ProgressBar.svelte
```diff
@@ -0,0 +1,85 @@
...
+		if (currentIndex <= 0) currentIndex = 0
+		else currentIndex--
+		series[currentIndex].isDone = false
+	}
+
+	function next() {
+		if (isLoop(currStep)) {
+			const max = series[currentIndex].kind.length - 1
+			if (loopIndex >= max) loopIndex = max
+			else return loopIndex++
```
fatonramadani: I suggest using the bracket notation to increase readability

frontend/src/lib/components/progressBar/ProgressBar.svelte
adam-kov: It's needed because `toFixed()` returns a string

frontend/src/lib/components/progressBar/ProgressBar.svelte
adam-kov: On the `if else` lines, right?

PR #775, opened 2022-10-20
Cargo config improvements

HurricanKai:
Actually enables the existing configuration and adds --no-rosegment.

### DIFF
--- a/backend/.cargo/config.toml
+++ b/backend/.cargo/config.toml
@@ -1,4 +1,10 @@
 [build]
-rustflags = ["--cfg", "tokio_unstable", "-C", "link-arg=-fuse-ld=lld"]
+rustflags = [
+    "--cfg",
+    "tokio_unstable",
+    "-C",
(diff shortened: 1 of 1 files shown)


backend/.cargo/config.toml
```diff
@@ -0,0 +1,10 @@
+[build]
+rustflags = [
+    "--cfg",
+    "tokio_unstable",
+    "-C",
+    "link-arg=-fuse-ld=lld",
+    "-Clink-arg=-Wl,--no-rosegment",
```
rubenfiszel, line 7: What's the benefit of this ?

HurricanKai: It passes --no-rosegment through to lld.
Which is documented as
> Do not put read-only non-executable sections in their own segment

It helps with stack traces of release builds and is necessary for some profiles.
I can remove it, but it doesn't hurt performance from what I know, it's only a size optimization

PR #776, opened 2022-10-20
Tokio improvements

HurricanKai:
Tokio related improvements. Especially around interval usage. Reworks logs to use streaming.

All benches done with a dedicated server and worker process. Only one worker instance running.

Perf Metrics:
1a935935291bcb01bb8b7cc037949fb6b36afff0 (base):
throughput /s 16.19
zombie jobs:  0
incorrect results:  0
queue length: 0
waiting for metrics
done

7393b7fec6ab364c55ba56691ee02b4d4c5d2b0b (latest):
throughput /s 16.76
zombie jobs:  0
incorrect results:  0
queue length: 0

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -266,20 +266,7 @@ async fn openapi() -> &'static str {
 }
 
 pub async fn shutdown_signal(tx: tokio::sync::broadcast::Sender<()>) -> anyhow::Result<()> {
-    use std::io;
-    use tokio::signal::unix::SignalKind;
-
-    async fn terminate() -> io::Result<()> {
-        tokio::signal::unix::signal(SignalKind::terminate())?
-            .recv()
-            .await;
-        Ok(())
-    }
-
-    tokio::select! {
-        _ = terminate() => {},
-        _ = tokio::signal::ctrl_c() => {},
-    }
+    tokio::signal::ctrl_c().await.unwrap();
     println!("signal received, starting graceful shutdown");
     let _ = tx.send(());
     Ok(())
--- a/backend/src/worker.rs
+++ b/backend/src/worker.rs
@@ -1715,8 +1715,6 @@ async fn handle_child(
                 },
             );
         }
-
-        drop(interval);
     };
 
     #[derive(PartialEq, Debug)]
--- a/backend/src/worker.rs
+++ b/backend/src/worker.rs
@@ -35,7 +35,6 @@ use tokio::{
     sync::{
         broadcast,
         mpsc::{self, Sender, UnboundedSender},
-        oneshot,
     },
     time::{interval, sleep, Instant, MissedTickBehavior},
 };
@@ -1836,10 +1835,7 @@ async fn handle_child(
         }
     };
 
-    /* a stream updating "queue"."last_ping" at an interval */
-
-    let (kill_tx, mut kill_rx) = oneshot::channel::<()>();
-
+    let mut kill_reason_rx = kill_reason_rx_org.resubscribe();
     let mut interval = interval(ping_interval);
     interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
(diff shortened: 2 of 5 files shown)


backend/src/lib.rs
```diff
@@ -266,20 +266,7 @@ async fn openapi() -> &'static str {
...
-            .recv()
-            .await;
-        Ok(())
-    }
-
-    tokio::select! {
-        _ = terminate() => {},
-        _ = tokio::signal::ctrl_c() => {},
-    }
+    tokio::signal::ctrl_c().await.unwrap();
```
rubenfiszel, line 269: why can we remove the terminate signal listener ? Is that already handled by ctrl_c ?

HurricanKai: I believe so. The [docs](https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html) say
> Even if this Signal instance is dropped, subsequent SIGINT deliveries will end up captured by Tokio, and the default platform behavior will NOT be reset.

Which to me reads it'll just detect SIGINT

backend/src/worker.rs
```diff
@@ -1626,25 +1626,36 @@ async fn handle_child(
...
+                            false
+                        })
+                    {
+                        break;
+                    }
+                },
+            );
         }
+
+        drop(interval);
```
rubenfiszel: why do we need to drop explicitely ? intervals are never cleaned otherwise ?

backend/src/worker.rs
```diff
@@ -1626,25 +1626,36 @@ async fn handle_child(
     let output = child_joined_output_stream(&mut child);
     let job_id = job_id.clone();
 
+    let (tx, mut rx) = mpsc::channel::<()>(1);
```
rubenfiszel, line 1691: does broadcast not work, like for the killpill which this is essentially ?

backend/src/lib.rs
rubenfiszel: SIGINT yes, this is what ctrl+c is, but what about sigterm ?

backend/src/worker.rs
rubenfiszel: Also, let's give it a more explicit name: kill_tx, kill_rx

HurricanKai: Broadcast would also work, but I try to use the smallest feature set for channels, to allow optimizations by tokio. Changed this to use oneshot now.

rubenfiszel: :+1:

rubenfiszel: rename would still be appreciated

backend/src/worker.rs
```diff
@@ -1626,24 +1688,33 @@ async fn handle_child(
...
+                        .unwrap_or_else(|err| {
+                            tracing::error!(%job_id, %err, "error checking cancelation for job {job_id}: {err}");
+                            false
+                        })
+                    {
+                        break;
+                    }
+                },
+            );
         }
```
rubenfiszel, line 1718: we can remove this by making the last ping statement become:
```
UPDATE queue SET last_ping = now() WHERE id = $1 RETURNING canceled"
```

and killing the select below through the oneshot (there would be only one oneshot now)

HurricanKai: Surprisingly this leads to a regression. It would take some load of the SQL server, but I doubt that it's worth it.

backend/src/lib.rs
HurricanKai: <https://github.com/tokio-rs/tokio/blob/a03e0420249d1740668f608a5a16f1fa614be2c7/tokio/src/signal/unix.rs#L496> doesn't look like it. Will change to also handle sigterm

HurricanKai:
Reverting channel related changes. ~-0.5 throughput. Not sure why, but that's why we bench!

backend/src/worker.rs
rubenfiszel: That's very surprising, is that really the case when you use the longest of the 2 periods ?

HurricanKai: I'll redo these changes at a later point (preferably when we have automated, authoritative benches).

PR #815, opened 2022-10-25
restructure the entire backend layout

HurricanKai:
This splits the backend into a bunch of logical segments, improving compile times, with the added benefit of helping with separation of concerns.

List of behavior changes (not necessarily complete, but these are the deliberate ones):
- API is used in workers over interacting with the Database directly
  - When resolving schedules (in `schedule_again_if_scheduled`)
  - In deno ops (resolving variables, job results, resources)
  - When resolving variables & resources during transform.
- API usage (new and old) has been reworked to use an autogenerated rust OpenAPI client.
- A token is created (currently via INSERT into the database) on startup to authenticate with the API. There is a token created for every workspace, this has not changed.

TODO (before merge):
- [x] Implement root binary (currently only worker & api binaries are implemented)
- [x] reimplement secret verification in the queue
- [x] implement push_scheduled_job from worker
- [x] reimplement op_get_id. It appears this has been using an undocumented endpoint so far?
- [x] merge main again
- [x] write README in all new cargo crates + a README in the backend/ folder with a quick TOC

TODO (can be after merge):
- [ ] change the list_jobs endpoint to not directly interact with the queue table
- [ ] share the API configuration used in deno ops or just implement these in "userspace" (-> TS)
- [ ] proper worker auth
- [ ] move script_path_to_payload into the API instead of querying two APIs from the worker
- [ ] move create_token_for_owner out of the worker
- [ ] move get_email_from_username out of the worker
- [x] investigate paperclip
- [ ] move get_latest_hash_for_path elsewhere
- [ ] move integration tests

Eventually there are more concerns regarding who has access to what, but these are the ones that are very obvious or block implementation somehow.


backend/windmill-api/openapi.yaml
```diff
@@ -2086,6 +2085,36 @@ paths:
...
+          in: query
+          schema:
+            type: boolean
+      responses:
+        "200":
+          description: job result
+          content:
+            application/json:
+              schema: {}
+
```
HurricanKai: Take note of this. I've never used a large OpenAPI document like this, so not sure if correct like this, or if things should be factored out into #/... references.
Also the description.

backend/windmill-api/src/jobs.rs
```diff
@@ -195,1826 +251,1217 @@ impl RunJobQuery {
...
+            Some(&parent_flow.to_string()),
+            None,
+        )
+        .await?;
+    }
     tx.commit().await?;
-
-    run_wait_result(authed, Extension(user_db), uuid, Path((w_id, script_hash))).await
+    Ok("Flow of job cancelled".to_string())
 }
```
HurricanKai: This had complicated merge changes, due to the approval changes. I think I did it right, but taking a look is probably a good idea.

backend/windmill-api/src/jobs.rs
```diff
@@ -195,1826 +251,1217 @@ impl RunJobQuery {
...
-pub async fn get_path_for_hash<'c>(
-    db: &mut Transaction<'c, Postgres>,
-    w_id: &str,
-    hash: i64,
-) -> error::Result<String> {
-    let path = sqlx::query_scalar!(
-        "select path from script where hash = $1 AND (workspace_id = $2 OR workspace_id = \
-         'starter')",
-        hash,
+pub async fn get_suspended_job_flow(
```
HurricanKai, line 573: Same approval related comment here.

backend/windmill-api/src/jobs.rs
```diff
@@ -195,1826 +251,1217 @@ impl RunJobQuery {
...
-    let flow_status = flow
-        .flow_status()
-        .ok_or_else(|| anyhow::anyhow!("unable to deserialize the flow"))?;
-    let flow_module_status = flow_status
-        .modules
-        .iter()
-        .find(|p| p.job() == Some(job))
-        .ok_or_else(|| anyhow::anyhow!("unable to find the module"))?;
+/*
+TODO: Can this be removed safely?
```
HurricanKai: TODO I forgot.

backend/windmill-api/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+        let metrics_f = async {
+            match metrics_addr {
+                Some(addr) => windmill_common::serve_metrics(addr, rx.resubscribe())
+                    .await
+                    .map_err(anyhow::Error::from),
+                None => Ok(()),
+            }
+        };
+
+        futures::try_join!(shutdown_signal, server_f, metrics_f)?;
```
HurricanKai, line 67: Note that running just the API server won't start the monitor jobs. The root crate still does it, but here we don't.

TODO: maybe we can move the monitor_db thing into db via pg_cron?

backend/windmill-api/src/schedule.rs
```diff
@@ -0,0 +1,126 @@
+/*
```
HurricanKai, line 1: I think git is just very confused here. Not sure why, I didn't create this file, only changed methods below.

HurricanKai:
Also note that worker tests are broken currently. I figured we just want to get this merged asap and then I'll make follow up PRs to fix some of these non-critical issues.

backend/windmill-api-client/bundled.json
```diff
@@ -0,0 +1,7480 @@
+{
```
rubenfiszel: I think we should .gitignore this file too since it's auto-generated and prone to get out of sync.

backend/windmill-api/openapi.yaml
```diff
@@ -1608,7 +1608,7 @@ paths:
                 type: object
```
rubenfiszel, line 1608: All the clients rely on the location of this file so I believe some changes in their generator would be needed if we were to move this.

backend/windmill-api/openapi.yaml
```diff
@@ -2086,6 +2085,36 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: node_id
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: skip_direct
+          description: TODO Add description, I'm not even really sure what this does
```
rubenfiszel: Yeah my bad for poor naming but this flag skip checking the nodes as child of the flow of which we have the id for, useful in cases you already have done the check as we do inside the eval context of flows. 
As reminder, the algorithm look for result by id in every parent flows as we assume that it means they have been evaluated earlier and are in the direct path that led to the execution of the job that does this call.

backend/src/main.rs
```diff
@@ -111,24 +113,127 @@ async fn main() -> anyhow::Result<()> {
...
+
+async fn get_token(db: &sqlx::Pool<sqlx::Postgres>, worker_name: &str) -> String {
+    // TODO: This is bad. Proper authentication should happen.
+    use rand::prelude::*;
+    let token: String = rand::thread_rng()
+        .sample_iter(&rand::distributions::Alphanumeric)
+        .take(30)
+        .map(char::from)
+        .collect();
+    let expiration = chrono::Utc::now() + chrono::Duration::days(2);
```
rubenfiszel: What would happen currently after 2 days ? If things just break, let's make things perpetual and TODO implement some kind of refresh token logic

backend/windmill-api/openapi.yaml
rubenfiszel: The rule of 3, references that would only be used once or twice are not worth being made in most cases

backend/windmill-api/src/main.rs
rubenfiszel: I'm afraid of using pg_cron because it will break "all you need is a standard pg instance."

backend/src/main.rs
rubenfiszel: Since the client is non super_admin, it cannot access workspace specific data and jobs should be part of that, i need to undestand more why it works at all

backend/windmill-api/openapi.yaml
HurricanKai: Which do you prefer, moving it back or adjusting references?

backend/src/main.rs
HurricanKai: I need to check again, but I think this token is only used for very few calls anyways. The most important one is the create_token, creating the per-job token, that does also have access to specific workspace.

backend/windmill-api/openapi.yaml
rubenfiszel: Let's adjust the references, I think it makes more sense in windmill-api

backend/src/main.rs
rubenfiszel: Creating per-job token based on a token should not be able to escalate token privilege by having a user that is not part of a workspace able to create a user-specific token part of a workspace. So the token created should either not have enough privilege or we've implemented the endpoint very very wrong.

PR #827, opened 2022-10-28
feat(frontend): Flow graph

adam-kov:
Can be viewed at the details page of a flow.


frontend/src/lib/components/graph/FlowGraph.svelte
```diff
@@ -0,0 +1,226 @@
...
+		loop.items.push(createVirtualNode(getParentIds(loop.items), 'Collection of the results'))
+		return loop
+	}
+
+	function flowModuleToBranch(branches: FlowModule[][], parent: NestedNodes | undefined = undefined): Branch {
+		const branch: Branch = {
+			type: 'branch',
+			items: []
+		}
+		console.log(branches[0]);
```
fatonramadani: Console.log to remove!

frontend/src/lib/components/graph/FlowGraph.svelte
```diff
@@ -0,0 +1,226 @@
...
+		}
+		return flowModuleToNode(parentIds, module.id, module.summary || 'Identity script', 'inline')
+	}
+
+	function getParentIds(items: NestedNodes | undefined = undefined): string[] {
+		const item = items?.at(-1) || nestedNodes.at(-1)
+		if(!item) return []
+
+		if(isNode(item)) {
+			return ['' + item.id]
```
fatonramadani, line 120: Are id either string of number ? I see you are also doing some: `+numberAsString` to convert things back and forth

frontend/src/lib/components/graph/util.ts
```diff
@@ -0,0 +1,15 @@
...
+	gap: {
+		horizontal: 20,
+		vertical: 50
+	}
+}
+
+export function* createIdGenerator(): Generator<number, number, unknown> {
+	let id = 0
+	while(true) {
+		yield id++
```
fatonramadani, line 13: We have a logic in the flow to generate increasing ids with letters: 'a', b', ..., 'z', 'aa', 'ab
Can we use this to avoid casting ids ?

adam-kov: The Svelvet library accepts only numbers as node IDs

frontend/src/lib/components/graph/FlowGraph.svelte
adam-kov: Svelvet library uses numbers as IDs but the layout library accepts an array of strings as the parentIds parameter so the conversion is needed to make the two packages work together

PR #845, opened 2022-11-01
Add helm to app image

martysohio:
Helm version 3.10 added for internal Kubernetes deployments


martysohio:
recheck

martysohio:

.docker/DockerfileHelm
```diff
@@ -0,0 +1,15 @@
+FROM ghcr.io/windmill-labs/windmill:main
```
rubenfiszel, line 1: We should add a comment:
```
This is an example of how to extend Windmill images to include any binaries desired in the worker
```

PR #887, opened 2022-11-10
Add helm and AWS cli to main image

martysohio:
Example of adding binaries to the primary Winndmill image for execution.  Also for internal usage.


rubenfiszel:
.docker -> docker
Dockerfilehelm -> DockerfileHelm, otherwise looks great!

PR #895, opened 2022-11-14
fix(frontend): Reduce the size of the separator + fix Auto scroll

fatonramadani:
Screenshot 2022-11-14 at 17 15 39
Screenshot 2022-11-14 at 17 15 35


frontend/src/lib/components/LogViewer.svelte
```diff
@@ -21,13 +21,19 @@
...
-		>
+		<pre class="whitespace-pre-wrap break-words bg-gray-50 text-xs w-full p-2">
+			{#if content}
+				{content}
+			{:else if isLoading}
+				Waiting for job to start...
+			{:else}
+				No logs are available yet
+			{/if}
+			</pre>
```
rubenfiszel: Can you check that this is not adding newines and or spaces before? THat's why the previous version was so compacted

PR #907, opened 2022-11-15
feat(backend): sync cache features on all workers [enterprise]

HurricanKai:
~~This adds a background task running every 1000 jobs or 60 minutes. It runs a file via bash.~~

~~New jobs are prevented from running, but existing ones are left to run. The locking isn't very strict, but prevents some worst-case scenarios, stricter locking would hurt performance for little benefit.~~

~~Note that as bash invoked file will need to be marked as executable (via `chmod +x`)~~

~~The execution bit of this code is put behind the enterprise feature, this is very easy to remove and doesn't really disable the background task, but removing all the related parameters is difficult and very annoying to work with (autocompletion just breaks etc.)~~

Revised to simply run every 10 minutes on the current worker thread.
This does not handle multi-worker environments. This does nothing if no script is provided or the enterprise feature isn't enabled.


backend/windmill-worker/Cargo.toml
```diff
@@ -44,4 +45,6 @@ lazy_static.workspace = true
 chrono.workspace = true
 dotenv.workspace = true
 rand.workspace = true # TODO: Remove. only used by token creation hack.
-deno_core.workspace = true
\ No newline at end of file
+deno_core.workspace = true
+tokio-stream.workspace = true
```
rubenfiszel: we can remove this  now I think ?

backend/Cargo.toml
```diff
@@ -134,3 +134,12 @@ futures = "^0"
 tokio-metrics = "0.1.0"
 lazy_static = "1.4.0"
 serde_derive = "1.0.147"
+tokio-stream = "0.1"
```
rubenfiszel: we can remove this

backend/windmill-worker/src/worker.rs
```diff
@@ -110,10 +132,12 @@ pub async fn create_token_for_owner<'c>(
...
 const PIP_SUPERCACHE_DIR: &str = "/tmp/windmill/cache/pip_permanent";
-const PIP_CACHE_DIR: &str = "/tmp/windmill/cache/pip";
-const DENO_CACHE_DIR: &str = "/tmp/windmill/cache/deno";
-const GO_CACHE_DIR: &str = "/tmp/windmill/cache/go";
+const ROOT_CACHE_DIR: &str = "/tmp/windmill/cache/";
+const PIP_CACHE_DIR: &str = concatcp!(ROOT_CACHE_DIR, "pip");
+const DENO_CACHE_DIR: &str = concatcp!(ROOT_CACHE_DIR, "deno");
+const GO_CACHE_DIR: &str = concatcp!(ROOT_CACHE_DIR, "go");
 const NUM_SECS_ENV_CHECK: u64 = 15;
+const NUM_SECS_SYNC: u64 = 60 * 60;
```
rubenfiszel: let's do 10minutes

Dockerfile
```diff
@@ -59,29 +59,29 @@ FROM rust_base AS planner
...
 
-RUN CARGO_NET_GIT_FETCH_WITH_CLI=true cargo chef prepare  --recipe-path recipe.json
+RUN CARGO_NET_GIT_FETCH_WITH_CLI=true cargo chef prepare --recipe-path recipe.json
 
 FROM rust_base AS builder
 
 COPY --from=planner /windmill/recipe.json recipe.json
 
-RUN CARGO_NET_GIT_FETCH_WITH_CLI=true cargo chef cook --release --recipe-path recipe.json
+RUN CARGO_NET_GIT_FETCH_WITH_CLI=true cargo chef cook --release --features enterprise --recipe-path recipe.json
```
rubenfiszel: let's use buildarg for those

rubenfiszel: https://vsupalov.com/docker-arg-env-variable-guide/#setting-arg-values

PR #931, opened 2022-11-22
Fix Python flow lock

HurricanKai:
I would've prefered to fix this "upstream" by passing the python code through normally and handling the parsing in the worker, but this was much easier and doesn't involve breaking data.

### DIFF
--- a/backend/windmill-worker/src/worker.rs
+++ b/backend/windmill-worker/src/worker.rs
@@ -1861,8 +1861,20 @@ async fn handle_flow_dependency_job(
             new_flow_modules.push(e);
             continue;
         };
+        // sync with windmill-api/scripts
+        let dependencies = match language {
+            ScriptLang::Python3 => windmill_parser_py::parse_python_imports(&content)?.join("\n"),
+            _ => content.clone(),
+        };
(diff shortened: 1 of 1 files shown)


backend/windmill-worker/src/worker.rs
```diff
@@ -1861,8 +1861,20 @@ async fn handle_flow_dependency_job(
             new_flow_modules.push(e);
             continue;
         };
+        // sync with windmill-api/scripts
+        let dependencies = match language {
+            ScriptLang::Python3 => windmill_parser_py::parse_python_imports(&content)?.join("\n"),
+            _ => content.clone(),
```
rubenfiszel, line 1867: it seems weird to me to put it there, why not in capture_dependency_job in the python branch ?
Especially since the naming in capture_dependency_job is:     job_raw_code: &str,

HurricanKai: That's the option I mentioned above. We can change the windmill-api/scripts endpoint to just provide the raw code, and then parse dependencies in capture_dependency_job. But that'd probably involve also making the python parser accept already valid dependency definitions to allow jobs created by the old version to cleanly migrate to the new one.

PR #940, opened 2022-11-23
feat(cli): Run flows & scripts

HurricanKai:


rubenfiszel:
Could you add it to the README, this is a really powerful feature!

rubenfiszel:
We should allow for passing the ARGS with -d '<json>' of stdin too

PR #966, opened 2022-11-29
feat(cli): improved setup & allow workspace in base url & refactor workspaces/remotes to unify

HurricanKai:
New Setup Flow:

Accepts workspace in remote everywhere:

Now also includes workspace & remote refactor


HurricanKai:
Inline command to add a workspace w/ login: `wmill workspace add name id url`

PR #971, opened 2022-11-30
feat(deno-client): support mysql

mrl5:
I'm using https://deno.land/x/mysql2@v1.0.6/mod.ts only because https://deno.land/x/mysql@v2.11.0/mod.ts has bug that I described in https://github.com/denodrivers/mysql/issues/128#issuecomment-1331477637

hopefully this bug will be fixed soon, then I guess we can switch from https://github.com/sail-sail/mysql2 to https://github.com/denodrivers/mysql

tested locally with:
```ts
import { mysqlClient, mySql } from "./mysql.ts";

const db = {
    host: '0.0.0.0',
    port: 3306,
    user: 'root',
    password: 'change-me',
    database: 'test',
};

const conn = await mysqlClient(db);
await conn.execute('CREATE TABLE IF NOT EXISTS pets (name varchar(255), kind varchar(255))');
await conn.execute('INSERT INTO pets VALUES (?, ?)', ['behemot','cat']);
const [r] = await conn.execute('SELECT * from pets');
conn.end();
console.log(r);

const kind = 'cat';
const { rows } = await mySql(db)`SELECT * from pets WHERE kind = ${kind}`;
console.log(rows);
```

### DIFF
--- a/deno-client/mysql.ts
+++ b/deno-client/mysql.ts
@@ -1,0 +1,71 @@
+import { createConnection } from "https://deno.land/x/mysql2@v1.0.6/mod.ts";
+import { type Resource } from "./mod.ts";
+
+/**
+ * Establish MySQL connection using MySQL client for Deno:
+ * https://deno.land/x/mysql2@v1.0.6/mod.ts
+ *
+ * IMPORTANT: make sure to close the connection with `.end()`
+ *
+ * @param db the MySQL resource to establish the connection for
+ *
+ * @returns MySQL database connection
+ *
+ * @example
+ * ```ts
+ * const conn = await mysqlClient(db);
+ * await conn.execute('CREATE TABLE IF NOT EXISTS pets (name varchar(255), kind varchar(255))');
+ * await conn.execute('INSERT INTO pets VALUES (?, ?)', ['behemot','cat']);
+ * const [rows] = await conn.execute('SELECT * from pets');
+ * conn.end();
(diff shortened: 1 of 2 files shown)


deno-client/mysql.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Execute SQL query. For more info check:
+ * https://deno.land/x/mysql2@v1.0.6/mod.ts
+ *
+ * @param db the MySQL resource to run sql query for
+ *
+ * @returns array with two items: rows and fields
+ *
+ * @example
+ * ```ts
+ * const [rows, fields] = await mysqlQuery(db)('SELECT * from pets WHERE kind = ?', ['cat']);
```
mrl5: we want signatures of `mysqlQuery` and `pgQuery` to be aligned. meaning:
1.
```diff
- const [rows, fields] = mysqlResult;
+ const { rows, fields } = mysqlResult;
```

2.
```diff
const id = 1337;
- await mySql(db)('SELECT ID, NAME FROM CLIENTS WHERE ID = ?', [id]);
+ await mySql(db)`SELECT ID, NAME FROM CLIENTS WHERE ID = ${id}`;
```

mrl5: if I understand correctly, @rubenfiszel allows adjusting `pg.ts` to `mysql.ts`. but that's probably not perfect because this would probably mean breaking changes :/

PR #989, opened 2022-12-04
Hub lists

fatonramadani:


frontend/src/lib/components/FlowViewer.svelte
```diff
@@ -80,7 +80,7 @@
 			</div>
 		</TabContent>
 		<TabContent value="json">
-			<div class="relative">
+			<div class="p-4">
```
rubenfiszel: Are you sure about removing that relative?

fatonramadani: Fixed

PR #1003, opened 2022-12-07
feat: Superadmins workspace

HurricanKai:
Adds an "admins" workspace w/ preset hub sync on a schedule.
Also disables deletion for starter & admins.

I implemented the frontend for "admins" by using a derived store from `usersWorkspaceStore`. Let me know if I should implement this differently.

Do not deploy merge before #1002 has been released. This will likely be 1.55 (will need an edit before then to set the correct version in the script)

Update: 1.55 includes the relevant fix. See #1007


frontend/src/lib/stores.ts
```diff
@@ -1,55 +1,78 @@
...
+  }>
+> = derived(
+  [usersWorkspaceStore, superadmin],
+  ([store, superadmin]) => {
+    const originalWorkspaces = (store?.workspaces ?? []);
+    if (superadmin) {
+      return [...originalWorkspaces, {
+        id: "admins",
+        name: "Admins",
+        username: "superadmin",
```
rubenfiszel, line 42: We should put the starter workspace here too

PR #1040, opened 2022-12-22
feat!(cli): Folders support & Less Tarball nesting

HurricanKai:
This PR also includes dependency management with deps.ts as recommended in [the docs](https://deno.land/manual@v1.29.1/examples/manage_dependencies). This has been done so referencing a local deno client is easy for testing against an unreleased version.

Missing:
- [x] Handle extra_perms separately
- [x] Workspace export support
- [ ] is_oauth revert/ensure removal


HurricanKai:
Now breaking as this also includes the workspace folder renames. Can split into separate PR if we care

cli/context.ts
```diff
@@ -51,3 +68,56 @@ export async function requireLogin(opts: GlobalOptions) {
...
+    return false;
+  }
+
+  if (
+    !(path.startsWith("g") ||
+      path.startsWith("u"))
+  ) {
+    console.log(
+      colors.red(
+        "Given remote path looks invalid. Remote paths are typicall of the form <u|g|f>/<username|group|folder>/...",
```
rubenfiszel: typo typicall

cli/folder.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  if (optFolder) {
+    console.log(optFolder.extra_perms);
+    for (const [k, v] of Object.entries(optFolder.extra_perms)) {
+      if (!data.extra_perms || data.extra_perms[k] !== v) {
+        console.log(colors.red.underline.bold(`Extra Perms missmatch on ${k}`));
+        return;
+      }
+    }
```
rubenfiszel: seems a bit of an arbitrary check ? I think we can remove

cli/folder.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  for (const owner in ownersToAdd) {
+    await FolderService.addOwnerToFolder({
+      workspace,
+      name,
+      requestBody: {
+        owner,
+      },
+    });
+  }
```
rubenfiszel: can use the update endpoint now to do it in one go

cli/resource.ts
```diff
@@ -60,6 +56,14 @@ export async function pushResource(
...
     });
   } else {
+    if (typeof data.is_oauth !== "undefined") {
+      console.log(
+        colors.yellow(
+          "! is_oauth has been removed in newer versions. Ignoring.",
+        ),
+      );
+    }
+
```
rubenfiszel, line 66: That's not true ?

HurricanKai: I can't find this change via git blame, but [the 1.50 type def](https://deno.land/x/windmill@v1.50.0/windmill-api/models/CreateResource.ts?source) has is_oauth and [the 1.55 type def](https://deno.land/x/windmill@v1.55.0/windmill-api/models/CreateResource.ts?source) does not.

HurricanKai: Not sure why blame refuses to show this, but found it
https://github.com/windmill-labs/windmill/commit/37967a795006c2eb4e8b218abb3d1b0525c17d5e#diff-8bb7b61ea90ab2e4962997447574c28d7844f2450b33ee30064ec99b5bd66a01L4369-L4370
was between 1.52 and 1.51 during OAuth refactor

rubenfiszel: you're right, it was indeed removed. Only variable has it now.

PR #1049, opened 2022-12-28
App bugfix

fatonramadani:
Fix various bugs:
- resizing any input make it take a single unit of height and impossible to resize
- when creating a new button app, the label says "Default Label" instead of the "Lorem Ipsum" that is default set as its label
- there is some minimum padding between elements. We should make it 1 or 2 px vertically. Otherwise there is no way to actually use text as a label on tops of items such as table/inputs
- User input as a runnable input does not retrigger the compute - We should probably remove the visible toggle altogether, it's confusing - static = invisible - connected = invisible - user input = visible I think we should actually just have Static & Connected, and a toggle in Static that allow for User Input

Screenshot 2022-12-28 at 16 54 21


frontend/src/lib/components/apps/editor/GridEditor.svelte
```diff
@@ -68,19 +74,19 @@
 		rowHeight={32}
 		cols={columnConfiguration}
 		fastStart={true}
-		on:pointerup={({ detail }) => {
-			if (!$connectingInput.opened) {
-				$selectedComponent = detail.id
-			}
-		}}
+		on:pointerup={({ detail }) => selectComponent(detail.id)}
+		gap={[10, 2]}
```
fatonramadani: @rubenfiszel The library exposes a way to configure gaps easily

PR #1057, opened 2022-12-30
fix(frontend): Static inputs overflow

adam-kov:
- fixed hidden overflow on static inputs panel
- updated panel selector button

### DIFF
--- a/frontend/src/lib/components/flows/map/FlowConstantsItem.svelte
+++ b/frontend/src/lib/components/flows/map/FlowConstantsItem.svelte
@@ -1,9 +1,8 @@
 <script lang="ts">
 	import type { FlowEditorContext } from '../types'
 	import { getContext } from 'svelte'
-	import Icon from 'svelte-awesome'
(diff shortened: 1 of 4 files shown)


frontend/src/lib/components/flows/map/FlowConstantsItem.svelte
```diff
@@ -13,10 +12,9 @@
 	)
 </script>
 
-<!-- svelte-ignore a11y-click-events-have-key-events -->
-<div on:click={() => select('constants')} class={settingsClass}>
-	<Icon data={faDollarSign} class="mr-2" />
-	<span class="text-xs flex flex-row justify-between w-full gap-2 items-center truncate">
+<button on:click={() => select('constants')} class={settingsClass}>
+	<DollarSign size={16} />
+	<span class="text-xs font-bold flex flex-row justify-between w-full gap-2 items-center truncate ml-1">
```
rubenfiszel: let's remove font-bold, it's not as important as Flow Settings

PR #1058, opened 2022-12-31
fix(frontend): fix static component input

fatonramadani:


rubenfiszel:
this was fixed a few commits ago

PR #1071, opened 2023-01-06
feat(cli): 2-Way sync

HurricanKai:
CLI interface now works using `wmill sync init/add/pull/push`. Might add `rm` eventually.

`init` will create an initial state in `.wmill` further commands then work on top of that.

Few details on the storage here, but I fully expect this to change and intend on this just being an implementation detail and not anything we would ever document:
- paths are mapped to IDs in primary state file
- IDs are mapped to content files (mapping in primary state file) for snapshot of remote state
- IDs are mapped to hashes in primary state file for quick comparison.
- nanoid used for IDs
- microdiff used to diff things
- types like FolderFile, FlowFile etc. have been converted to classes & implement an interface now, allowing one to push diffs to remote
- decoverto is used to sanetize input data, which works by decorating the properties of the classes. It's important we do convert to/from JSON through decoverto everywhere to avoid leaking internal state.

TODO:
- [x] script content files impl
- [x] Self Review
- [ ] test hub sync
- [x] test push of individual files
  - [x] flow
  - [x] folder
  - [x] resource type
  - [x] resource
  - [x] script
  - [x] variable
- [ ] Add setup docs


cli/VariableFile.ts
```diff
@@ -0,0 +1,90 @@
+import { PushDiffs, Resource } from "./types.ts";
```
HurricanKai: TODO: Remove this

backend/windmill-api/src/users.rs
```diff
@@ -906,9 +906,9 @@ pub async fn require_owner_of_path(
                 )));
             }
         } else if splitted[0] == "g" {
-            return crate::groups::require_is_owner(splitted[1], username, groups, w_id, db).await;
+            return crate::groups::require_is_owner(w_id, username, groups, splitted[1], db).await;
         } else if splitted[0] == "f" {
-            return crate::folders::require_is_owner(splitted[1], username, groups, w_id, db).await;
+            return crate::folders::require_is_owner(w_id, username, groups, splitted[1], db).await;
         }
```
rubenfiszel: incorrect diff

cli/context.ts
```diff
@@ -67,6 +69,27 @@ export async function requireLogin(opts: GlobalOptions) {
   }
 
   setClient(token, workspace.remote.substring(0, workspace.remote.length - 1));
+
+  try {
+    await UserService.globalWhoami();
+  } catch {
+    console.log(
+      "! Could reach API given existing credentials. Attempting to reauth...",
```
HurricanKai: Not reach

cli/sync.ts
```diff
@@ -0,0 +1,427 @@
...
+        setValueByPath(json, diff.path, undefined);
+      } else if (diff.type === "CHANGE") {
+        setValueByPath(json, diff.path, diff.value);
+      }
+    }
+
+    // TODO: Use decoverto instanceToPlain below.
+    await Deno.writeTextFile(file, JSON.stringify(json, undefined, "  "), {
+      create: true,
+    });
```
HurricanKai, line 314: Implement these todos

cli/types.ts
```diff
@@ -1,4 +1,93 @@
...
+import { FlowFile } from "./flow.ts";
+import { ResourceTypeFile } from "./resource-type.ts";
+import { ResourceFile } from "./resource.ts";
+import { ScriptFile } from "./ScriptFile";
+import { VariableFile } from "./variable.ts";
+import { path } from "./deps.ts";
+import { FolderFile } from "./folder.ts";
+
+// TODO: Remove this & replace with a "pull" that lets the object either pull the remote version or return undefined.
+// Then combine those with diffing, which then gives the new push impl
```
HurricanKai, line 11: Either implement here or follow up soon

cli/variable.ts
```diff
@@ -26,13 +41,86 @@ async function list(opts: GlobalOptions) {
...
+      console.log(
+        colors.bold.yellow(
+          `Applying ${diffs.length} diffs to existing variable...`,
+        ),
+      );
+      const changeset: EditVariable = {};
+      for (const diff of diffs) {
+        if (
+          diff.path.length !== 1 ||
+          !(diff.path[0] in ["path", "description", "is_secret", "description"])
```
HurricanKai: Should not have path

cli/flow.ts
```diff
@@ -1,16 +1,130 @@
...
+    ) {
+      console.log(
+        colors.bold.yellow(
+          `Applying ${diffs.length} diffs to existing flow...`,
+        ),
+      );
+
+      // TODO: Make these optional in backend (not path ofc)
+      const changeset: OpenFlowWPath = {
+        path: remotePath,
```
HurricanKai, line 73: Remove this - moving is unsupported at this time

HurricanKai (commented):
There's another place where I incorrectly set path in a changeset, couldn't find it in GH but it should be removed as moving is not supported for now

HurricanKai:
Merge conflicts are now always rejected using an error message. We're not generating the true local-remote diff here, as this would generate unecessary overhead.
screenshot-tmp

HurricanKai:
closes #1174

PR #1106, opened 2023-01-16
create preview environments for every PR using Uffizzi

waveywaves:
This PR adds github workflows which will trigger uffizzi preview environments for PRs to this repo. A PoC has been created at https://github.com/waveywaves/windmill/pull/1. Once this PR is merged, you should see comments like https://github.com/waveywaves/windmill/pull/1#issuecomment-1381656171 pinned to all subsequent PRs to this repo. These comments are created after a preview env has been deployed for the PR.

To understand this PR better do check this article below on two stage workflows 
https://www.uffizzi.com/preview-environments-guide/github-actions-two-stage-workflow

fixes https://github.com/windmill-labs/windmill/issues/1098


rubenfiszel:
Also, I would like to migrate our playwright to using the preview environment. It would be super powerful. Do you think you could have time trying it out ? I gave permissions to run workflow to this PR

You can see where our playwright is setup here: https://github.com/windmill-labs/windmill/blob/main/.github/workflows/docker-image.yml#L113 (it currently fails because our test is broken so don't worry if it's red)

We do a bit of machinery to setup an env and if it could just target the preview environment that would just be so much easier. 

I'm also interested to deploy a preview environment (and run playwright this way too) on our main branch, after we've built our main images

rubenfiszel:
Sure, can you remove the lsp build as it will consume github actions minutes unecessarily and we can iterate from there

rubenfiszel:
I merged the PR and fixed the lsp thing but it seems the preview environment building has the following error: https://github.com/windmill-labs/windmill/actions/runs/3932485569/jobs/6725111157
I did some attempt to have playwright point to the preview environment, I basically just need the URL to point playwright too. Could you give me some poiners how to get that url in the workflow ?

PR #1107, opened 2023-01-16
feat(frontend): Add Supabase resource

fatonramadani:


frontend/src/lib/components/AppConnect.svelte
```diff
@@ -30,6 +30,9 @@
 				'Create an API key',
 				'Copy your key'
 			]
+		},
+		supabase: {
+			instructions: ['Go to the API Settings of your app to find the project URL and key']
```
rubenfiszel, line 36: Would be a little bit nicer with an img :)

fatonramadani: Done!

PR #1121, opened 2023-01-17
fix(frontend): Select the first row by default, and remove the abilit…

fatonramadani:
…y to deselect + fix table header z-index

### DIFF
--- a/frontend/src/lib/components/apps/components/table/AppTable.svelte
+++ b/frontend/src/lib/components/apps/components/table/AppTable.svelte
@@ -129,7 +129,7 @@
 
 			<div class="overflow-x-auto flex-1 w-full">
 				<table class="relative w-full border-b border-b-gray-200">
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/apps/components/table/AppTable.svelte
```diff
@@ -133,7 +129,7 @@
 
 			<div class="overflow-x-auto flex-1 w-full">
 				<table class="relative w-full border-b border-b-gray-200">
-					<thead class="sticky top-0 z-50 bg-gray-50 text-left">
```
fatonramadani, line 136: @adam-kov Had to remove this, because the anchor + move button were rendering under the table header

adam-kov: This will cause the action buttons to render above the table header when you scroll down in the table content

PR #1125, opened 2023-01-19
fix(frontend): fix first row selection

fatonramadani:


frontend/src/lib/components/apps/components/table/AppTable.svelte
```diff
@@ -55,6 +55,18 @@
...
 	}
 
+	let mounted = false
+	onMount(() => {
+		mounted = true
+	})
+
+	$: Array.isArray(result) &&
+		result.length > 0 &&
+		selectedRowIndex === -1 &&
```
rubenfiszel: you want the selectedRowIndex to be the first condition since this is gonna tick super often and integer comparison is super quick

fatonramadani: done

PR #1132, opened 2023-01-19
fix(frontend): Make sure that old apps are rendering properly

fatonramadani:


frontend/src/lib/components/apps/editor/GridEditor.svelte
```diff
@@ -177,17 +177,19 @@
...
-		<HiddenComponent
-			id={`bg_${index}`}
-			inlineScript={script.inlineScript}
-			name={script.name}
-			bind:fields={script.fields}
-			bind:staticOutputs={$staticOutputs[`bg_${index}`]}
-		/>
-	{/if}
-{/each}
+{#if $app.hiddenInlineScripts}
```
fatonramadani, line 180: @rubenfiszel I forgot that, we need to check that because old apps don't have hiddenInlineScripts

PR #1141, opened 2023-01-21
Square API OAuth

felipealbertao:
Adding Square OAuth, as per the Contributor's Guide

### DIFF
--- a/backend/oauth_connect.json
+++ b/backend/oauth_connect.json
@@ -77,5 +77,10 @@
     "token_url": "https://www.linkedin.com/oauth/v2/accessToken",
     "scopes": ["w_member_social", "r_liteprofile", "r_emailaddress"],
     "req_body_auth": true
+  },
+  "square": {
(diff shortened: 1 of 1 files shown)


felipealbertao:

rubenfiszel:
Thanks for this, will take a look tomorrow to test it end to end

rubenfiszel:
@felipealbertao I really tried making OAuth work but square uses a non conventional OAuth flow that does not follow the RFC 
(See the use of JSON in https://developer.squareup.com/reference/square/oauth-api/obtain-token instead of the Basic Auth of www-urlform payload as defined in the RFC: https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1

Since I didn't think it was worth integrating a square specific client and that square provides a way to use PAT instead, I went this route and added square as an official resource type and added the instructions

Screenshot 2023-01-21 at 09-38-02 Resources Windmill

PR #1142, opened 2023-01-23
feat(frontend): Add app inputs configurations

fatonramadani:
Screenshot 2023-01-23 at 07 59 12

Screenshot 2023-01-23 at 08 00 01


rubenfiszel:
image
@fatonramadani missing some text when there are no inputs to be filled for a component, cause screenshot look a bit not very clear

fatonramadani:
> Also it looks like it doesn't open itself hub/template imports

Fixed in last commit

PR #1144, opened 2023-01-23
fix(frontend): Make sure AppSelect items are an array

fatonramadani:


frontend/src/lib/components/apps/components/selectInputs/AppSelect.svelte
```diff
@@ -46,7 +46,7 @@
 			e?.stopPropagation()
 			window.dispatchEvent(new Event('pointerup'))
 		}}
-		{items}
+		items={Array.isArray(items) ? items : []}
```
fatonramadani, line 49: @adam-kov The library svelte-select seems to throw an error when you feed something that is not an array. ( it tries to call the "filter" method on it)

adam-kov (approved):

PR #1149, opened 2023-01-23
fix(frontend): App table action button cell

adam-kov:
Fixes issue of the action cell not taking up the full height


frontend/src/lib/components/apps/components/table/AppTable.svelte
```diff
@@ -211,18 +211,20 @@
...
-										{#each actionButtons as actionButton, actionIndex (actionIndex)}
-											<AppButton
-												noWFull
-												{...actionButton}
-												extraQueryParams={{ row: row.original }}
-												bind:componentInput={actionButton.componentInput}
-												bind:staticOutputs={$staticOutputsStore[actionButton.id]}
-											/>
-										{/each}
+											<div class="center-center h-full w-full flex-wrap gap-1">
```
fatonramadani, line 217: I think there is typo, content-center maybe?

adam-kov: No, it's a custom utility class which centers on both axis using flex. You can see it in the Tailwind config.

fatonramadani (approved):

PR #1155, opened 2023-01-24
feat(frontend): Open debug runs from component

fatonramadani:
- Used a common store to communicate between components that are not in the same hierarchy

### DIFF
--- a/frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
+++ b/frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
@@ -66,7 +66,7 @@
 		jobs,
 		staticExporter,
 		errorByComponent,
-		eventBus
+		openDebugRun
 	} = getContext<AppEditorContext>('AppEditorContext')
 
 	const loading = {
@@ -74,16 +74,15 @@
 		save: false
 	}
 
-	$: if ($eventBus.length > 0 && $eventBus[0].name === 'debug-runs') {
-		const { data: componentId } = $eventBus[0]
-		jobsDrawerOpen = true
+	$: if ($openDebugRun == undefined) {
+		$openDebugRun = (componentId: string) => {
+			jobsDrawerOpen = true
 
-		const job = $jobs.find((job) => job.component === componentId)
-		if (job) {
-			selectedJobId = job.job
+			const job = $jobs.find((job) => job.component === componentId)
+			if (job) {
+				selectedJobId = job.job
+			}
 		}
-
-		eventBus.update((events) => events.slice(1))
 	}
 
 	let newPath: string = ''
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
```diff
@@ -57,13 +57,35 @@
...
+		const { data: componentId } = $eventBus[0]
+		jobsDrawerOpen = true
+
+		const job = $jobs.find((job) => job.component === componentId)
+		if (job) {
+			selectedJobId = job.job
+		}
+
+		eventBus.update((events) => events.slice(1))
+	}
```
rubenfiszel: An event bus seems quite complex compared to putting an:
`openDebugRun(componentId)`

any strong reasons for it ? What else do you see having there in the future ?

frontend/src/lib/components/apps/editor/ComponentHeader.svelte
```diff
@@ -78,6 +85,7 @@
 					<Alert type="error" title="Error during execution">
 						<div class="flex flex-col">
 							<span> See "Debug Runs" on the top right for more details </span>
```
rubenfiszel: I would remove this altogether and put the error directly, with the button underneath or above it's clear enough

frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
fatonramadani: It's a pattern that is already used in the app editor, for example opening the connect input menu.
I would only use one pattern for theses kind of cross-component communications

rubenfiszel:
I would add a red border to runs with errors in the debug runs panel directly in this PR

frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
rubenfiszel: I still don't understand the benefit. This adds many layers of indirection and bug possibilities for no apparent benefits.

The alternative would be to literally just pass the function `openDebugRun(componentId)`  in a global context which is a pattern we use almost everywhere. That function can be passed up using component let bindings. It's a lot cleaner

fatonramadani:
Reverted the eventBus and implemented it in an easier way:
https://user-images.githubusercontent.com/456655/214297339-4582cb86-f133-43af-9dea-2fd873caf83a.mov

frontend/src/lib/components/apps/editor/ComponentHeader.svelte
fatonramadani: Done

frontend/src/lib/components/apps/editor/AppEditorHeader.svelte
fatonramadani: Done

PR #1158, opened 2023-01-26
feat: Add workspace webhook

### LINKED ISSUE #1147: feature: Content change webhook
Allow per-workspace configuration of a webhook to notify external systems of changes in the workspace.

HurricanKai:
closes #1147 

### Frontend
UI is still questionable for this iteration - some extra content will show up in this tab after #1140 but the tab seems pretty empty regardless.

### Backend
Uses a shared tokio "thread" to consume and fetch urls via reqwest. errors are silently swallowed. I don't think this is going to be an issue. Communication via MultiProducerSingleConsumer unbounded channel - there might be some DoS attack here, but I doubt it's a real issue. Everything is managed on top of axum middleware things, both creating & retrieving the process/channel and also creating clones & injecting into relevant functions. There's some concern here about excessive recreation of mpsc senders, but they appear cheap, so I doubt it's an issue either.

Basic metrics are included to monitor usage.


backend/windmill-api/src/webhook_util.rs
```diff
@@ -0,0 +1,166 @@
...
+            .await
+            .map_err(|_| {
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    "Could not aquire shared process while retrieving webhook".to_owned(),
+                )
+            })?;
+
+        Ok(Self { workspace_id, shared })
+    }
```
rubenfiszel: You do not need any of this, just do the impl on WebhookShared directly, all the the endpoints where you use it actually use workspace_id underneath.

PR #1160, opened 2023-01-26
feat(frontend): Add HTML result rendering

adam-kov:
Added the possiblity (with responsibility given to the user) to render HTML (and therefore display SVG images) in the result panel of the script editor

Screenshot 2023-01-26 at 14 55 50

### DIFF
--- a/frontend/src/lib/components/DisplayResult.svelte
+++ b/frontend/src/lib/components/DisplayResult.svelte
@@ -12,6 +12,7 @@
 		| 'json'
 		| 'table-col'
 		| 'table-row'
+		| 'html'
 		| 'png'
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/DisplayResult.svelte
```diff
@@ -12,16 +12,22 @@
...
+		| 'html'
 		| 'png'
 		| 'file'
 		| 'jpeg'
 		| 'gif'
 		| 'error'
 		| 'approval'
 		| undefined
+
+	$: if(result) {
```
rubenfiszel: seems a bit heavy to deactivate it at each change of result

PR #1171, opened 2023-02-01
fix(frontend): Render popups above components in app editor

adam-kov:
Screenshot 2023-02-01 at 14 26 01

### DIFF
--- a/frontend/src/lib/components/apps/editor/ComponentEditor.svelte
+++ b/frontend/src/lib/components/apps/editor/ComponentEditor.svelte
@@ -42,11 +42,6 @@
 	{/if}
 
 	<div
-		on:pointerdown={(e) => {
-			if ($mode === 'preview') {
-				e?.stopPropagation()
-			}
-		}}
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/apps/editor/ComponentEditor.svelte
```diff
@@ -42,11 +42,6 @@
 	{/if}
 
 	<div
-		on:pointerdown={(e) => {
-			if ($mode === 'preview') {
-				e?.stopPropagation()
-			}
-		}}
```
rubenfiszel: Why delete this ? This is to avoid conflict with svelte-grid onpointerdown

adam-kov: I apply styles to wrapper elements based on which element is clicked, and because of this, the event would never reach the wrapper in preview mode. I consulted with @fatonramadani and tested things but nothing weird came up. Do you have an example where this would cause issues?

PR #1185, opened 2023-02-06
feat(frontend): Trigger settings drawer with URL hash

adam-kov:
Open "User Settings" and "Superadmin Settings" by navigation, eg: `https://app.windmill.dev/#user-settings`


rubenfiszel:
Did you test this because I cannot make it work by going to a new url with #user-settings or #superadmin-settings

Tested on: https://35f852a0.windmill.pages.dev/#user-settings

Also not sure of the design of this PR, would be simpler to have a page that redirect to the workspace page with the drawer open

adam-kov:
Yes, tested it but you need to be logged in beforehand because the redirect URL doesn't save the hash (at least as I can see).

> would be simpler to have a page that redirect to the workspace page with the drawer open

Not sure what you mean by this. The settings drawers are unavailable on that page because they are only part of the main layout.

rubenfiszel:
You need to be logged in but also within a workspace otherwise it get overriden by the workspace redirect

adam-kov:
That's still an issue with the redirect, wdyt about adding the correct URL in the parameter? Also, I don't think people "hang out" in the workspace selector. If they are logged in, they probably have one selected.

rubenfiszel:
My understanding was that this PR was so we could give a link in the docs that would redirect straight to the user settings. Having a link that work only in certain conditions seem less useful.

frontend/src/lib/components/SuperadminSettings.svelte
```diff
@@ -22,6 +22,13 @@
 		drawer?.toggleDrawer?.()
 	}
 
+	export function closeDrawer() {
+		drawer?.closeDrawer()
+		const index = $page.url.href.lastIndexOf('#')
```
rubenfiszel, line 27: we should check if it's found, and if it's not, no need for slicing

rubenfiszel:
Looks good, just one minor nit

PR #1192, opened 2023-02-09
Readme: broken links to contributor's guide

NReilingh:
These links 404. Updated them to what appears to be the correct URL.


rubenfiszel:
Fixed in #1191 , but thanks a lot for the PR!

PR #1201, opened 2023-02-14
feat(backend): Zip Workspace Export

HurricanKai:


backend/windmill-api/src/workspaces.rs
```diff
@@ -1064,21 +1116,24 @@ async fn tarball_workspace(
     let name = format!("windmill-{w_id}.tar");
     let file_path = tmp_dir.path().join(&name);
     let file = File::create(&file_path).await?;
-    let mut a = tokio_tar::Builder::new(file);
-
+    let mut archive = match archive_type.as_str() {
+        "tar" | /* no query string is empty string */ "" => Ok(ArchiveImpl::Tar(tokio_tar::Builder::new(file))),
```
rubenfiszel: have you tested no query string is empty string ?

rubenfiszel:
Requiring updating openapi with the new query args.

In a follow up PR we should move frontend to use the zip as export

PR #1253, opened 2023-03-02
Add support for binding server listener to a specific IP address

ryanrich:
Add support to bind to a specific IP address for the server process via a `SERVER_BIND_ADDR` environment variable.  My specific use case for this was to restrict connections to the localhost as we need to ensure all inbound traffic is funneled through a separate process.


ryanrich:

PR #1319, opened 2023-03-22
fix(frontend): fix refresh with manual dependencies

fatonramadani:
- Enable refresh when autoRefresh is disabled but the script has manual dependencies that changed

### DIFF
--- a/frontend/src/lib/components/apps/editor/appUtils.ts
+++ b/frontend/src/lib/components/apps/editor/appUtils.ts
@@ -71,7 +71,7 @@ export function findGridItem(app: App, id: string): GridItem | undefined {
 export function getNextGridItemId(app: App): string {
 	const subgridsKeys = allItems(app.grid, app.subgrids).map((x) => x.id)
 	const withoutDash = subgridsKeys.map((element) => element.split('-')[0])
-	const id = getNextId([...new Set(withoutDash), 'do'])
+	const id = getNextId([...new Set(withoutDash)])
 
 	return id
 }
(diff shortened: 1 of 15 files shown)


frontend/src/lib/components/apps/editor/appUtils.ts
```diff
@@ -71,7 +71,7 @@ export function findGridItem(app: App, id: string): GridItem | undefined {
 export function getNextGridItemId(app: App): string {
 	const subgridsKeys = allItems(app.grid, app.subgrids).map((x) => x.id)
 	const withoutDash = subgridsKeys.map((element) => element.split('-')[0])
-	const id = getNextId([...new Set(withoutDash), 'do'])
+	const id = getNextId([...new Set(withoutDash)])
```
fatonramadani, line 74: @rubenfiszel Ids were always starting after "do", so the first id would be "dp"

frontend/src/lib/components/apps/editor/settingsPanel/triggerLists/BackgroundScriptTriggerList.svelte
```diff
@@ -14,11 +16,23 @@
...
+	$: dependencies = getDependencies(fields)
 </script>
 
+<div class="flex items-center px-1">
+	<Toggle
+		bind:checked={doNotRecomputeOnInputChanged}
+		options={{ right: "Don't recompute on input changed" }}
+	/>
+	<Tooltip>Whenever an input is changed, the script will be re-run.</Tooltip>
+</div>
```
rubenfiszel: Need to hide for frontend scripts no?

PR #1322, opened 2023-03-23
fix(frontend): Fix app errors

fatonramadani:
Screenshot 2023-03-23 at 15 46 23

- Fix AppTable error display
- clear errors when removing a component
- properly detect that latest component run had an error


rubenfiszel:
> @rubenfiszel wdyt of clearing jobs and errorByComponent when we remove a component ?

sounds good

PR #1323, opened 2023-03-23
fix(frontend): fix settings panel

fatonramadani:


frontend/src/lib/components/apps/editor/SettingsPanel.svelte
```diff
@@ -26,7 +27,7 @@
 					if (x?.data?.actionButtons) {
 						const tableAction = x.data.actionButtons.find((x) => x.id === id)
 						if (tableAction) {
-							return { item: tableAction, parent: x.data }
+							return { item: { data: tableAction, id: tableAction.id }, parent: x.data.id }
```
fatonramadani, line 30: @rubenfiszel had to manually wrap to keep types consistent

PR #1324, opened 2023-03-23
feat(backend): Redis based queue

HurricanKai:
A very barebones version of redis queueing.

Also includes a commit to update a V8 breaking change (I think?) as after rebasing this has no longer compiled for me.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -98,9 +98,9 @@ dependencies = [
 
 [[package]]
 name = "ast_node"
-version = "0.8.6"
+version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


backend/windmill-queue/src/queue_transaction.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+impl RedisOp {
+    pub async fn apply<R: RsmqConnection>(self, rsmq: &mut R) -> Result<(), rsmq_async::RsmqError> {
+        match self {
+            RedisOp::SendMessage(bytes, time) => {
+                rsmq.send_message(
+                    "main_queue",
+                    bytes,
+                    time.map(|t| (t - chrono::Utc::now()).num_seconds())
```
rubenfiszel, line 19: this should use the database now() otherwise there are risks of drift since the clock of the worker might be the same as the db

PR #1333, opened 2023-03-28
feat(frontend): improve input connection UI

fatonramadani:
https://user-images.githubusercontent.com/456655/228164661-59087398-27b7-432f-8be7-72ac11cf0e4a.mov

- You can now navigate you app while connecting an input to an output
- You can only navigate the app the highlight a component, not interact with it (click, input, etc)
- The inline scripts panel will collapse when connecting


frontend/src/lib/components/propertyPicker/ObjectViewer.svelte
```diff
@@ -97,14 +97,14 @@
 						>
 							{#if json[key] === NEVER_TESTED_THIS_FAR}
 								<WarningMessage />
-							{:else if json[key] == undefined}
-								<span class="text-2xs">undefined</span>
+							{:else if json[key] === null}
```
fatonramadani: @adam-kov Fixed the display when the value is null

frontend/src/lib/components/propertyPicker/ObjectViewer.svelte
```diff
@@ -98,13 +98,13 @@
 							{#if json[key] === NEVER_TESTED_THIS_FAR}
 								<WarningMessage />
 							{:else if json[key] == undefined}
```
rubenfiszel, line 100: need to separate undefined and null

fatonramadani: done

PR #1361, opened 2023-04-04
feat(backend): lowercase all emails

rubenfiszel:


ryanrich (commented):
Does the `invite_user` function in the workspaces also need to be updated so the `email` field in the `NewWorkspaceInvite` struct is lowercase?

rubenfiszel:
@ryanrich indeed, modified it as well

PR #1362, opened 2023-04-04
feat(frontend): improved cron/schedule editor

oliver-veal:
Screenshot 2023-04-05 at 12 00 27

- Introduced 'Basic' and 'Cron' edit modes for setting the cron schedule string
- User can now input the timezone they want the schedule to run in (defaults to browser-local timezone).
- chrono_tz crate used to parse the IANA timezone string sent from the client and generate the upcoming schedule events in that timzone.
- Schedule list view updated to list the timezone string
- Database schema changed: removed offset_ column & added timezone column from schedules table.


backend/windmill-queue/src/schedule.rs
```diff
@@ -27,8 +27,10 @@ pub async fn push_scheduled_job<'c>(
         .map_err(|e| error::Error::BadRequest(e.to_string()))?;
 
     let offset = Duration::minutes(schedule.offset_.into());
-    let now = now_from_db(&mut tx).await?;
```
rubenfiszel, line 30: you want to have the time of the db since the jobs are scheduled based on the time of the db. There might be a drift between the 2

oliver-veal:
Screenshot 2023-04-05 at 12 00 27

- User can now input the timezone they want the schedule to run in (defaults to browser-local timezone).
- chrono_tz crate used to parse the IANA timezone string sent from the client and generate the upcoming schedule events in that timzone.
- Schedule list view updated to list the timezone string
- Database schema changed: removed offset_ column & added timezone column from schedules table.

PR #1371, opened 2023-04-06
fix(frontend): Fix connected property

fatonramadani:


frontend/src/lib/components/InputTransformForm.svelte
```diff
@@ -40,17 +40,27 @@
 		schema.properties[argName].enum,
 		schema.properties[argName].contentEncoding
 	)
+
 	let propertyType = getPropertyType(arg)
 
+	function isExprVariableOrResource() {
```
fatonramadani, line 46: @rubenfiszel We still want expressions that are not resources of variables to use the codeToStaticTemplate. 
So what I did was to exclude expressions that matches variable(...) or resource(...)

PR #1377, opened 2023-04-10
fix(frontend): fix popover configuration to avoid content shift

fatonramadani:
Changed the configuration to avoid content shift:

Before:
Screenshot 2023-04-10 at 06 48 10

After:
Screenshot 2023-04-10 at 06 46 16

### DIFF
--- a/frontend/src/lib/components/Popover.svelte
+++ b/frontend/src/lib/components/Popover.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { createPopperActions } from 'svelte-popperjs'
+	import { createPopperActions, type PopperOptions } from 'svelte-popperjs'
 	import type { PopoverPlacement } from './Popover.model'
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/Popover.svelte
```diff
@@ -9,28 +9,16 @@
...
-		effect: ({ state }) => {
-			const { padding = 0 } = options
 
-			state.elements.popper.style.maxWidth = `calc(100vw - ${padding * 2}px)`
-		}
-	})
-	const extraOpts = {
+	const popperOptions: PopperOptions<{}> = {
+		placement: 'bottom-end',
+		strategy: 'fixed',
```
fatonramadani, line 15: @rubenfiszel Just changed from absolute to fixed, https://popper.js.org/docs/v2/constructors/#strategy

PR #1380, opened 2023-04-10
feat(frontend)!: Add option to return file names from file input

adam-kov:
**Could break existing apps!**

By default the app file input now returns an array of objects (name of the file and the base64 data) instead of an array of strings (base64 data).

This is needed because there is no way to get the name of the files from the base64 encoded data.

### DIFF
--- a/frontend/src/lib/components/apps/components/inputs/AppFileInput.svelte
+++ b/frontend/src/lib/components/apps/components/inputs/AppFileInput.svelte
@@ -19,11 +19,11 @@
 	let text: string | undefined = undefined
 
 	let outputs = initOutput($worldStore, id, {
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/apps/components/inputs/AppFileInput.svelte
```diff
@@ -19,11 +19,11 @@
 	let text: string | undefined = undefined
 
 	let outputs = initOutput($worldStore, id, {
-		result: [] as string[] | undefined
+		result: [] as { name: string; data: string }[] | undefined
```
fatonramadani, line 22: Could you please handle both cases so we don't break existing apps ?

PR #1395, opened 2023-04-11
Recursive schema UI display

axelbdt:
Display nested properties of objects for JSON Schemas created with inference from payload or JSON edit.


axelbdt:

frontend/src/lib/components/SchemaEditor.svelte
```diff
@@ -204,55 +250,15 @@
...
-											color="light"
-											variant="border"
-											size="sm"
-											startIcon={{ icon: faPen }}
-											on:click={() => startEditArgument(name)}
-										>
-											Edit
-										</Button>
-									</td>
+							{#each schemaPropertiesToDisplay(schema) as displayInfo (displayInfo.name)}
```
fatonramadani: `
export async function main(x: string, a: {x: number, y: {x:boolean}}) {
  return a.y.x
}
`
This code generates a valid schema, but property names are clashing and makes the app to crash
I think we might just use the index

axelbdt:
Thanks for the feedback Faton, I think using the index as key breaks the animation since the index changes on Move. As a fix, I keep the full path of the property and compute a crude hash from said path to use as a key.
It sets the stage nicely for nested delete as well.
Cheers
(I have also made the animation shorter, it does feel better)

axelbdt:
Tweaked the table layout to get more space for longer descriptions or nested structures.
The red star for required field is a convention throughout the whole app, so I think we don't need to take up a full column just to be more explicit about what it means.

Screenshot 2023-04-25 at 00-05-26 Resources Windmill

rubenfiszel:
I will take a look by EOD

rubenfiszel:
It seems that it breaks under 2 levels of depths:

> export async function main(x: string, y: {foo: {fooo: 43}, bar: 43}) {
  return x
}

image

The other issue, but that can be considered just a TODO is that the schema editor still doesn't seem to support defining complex object from the UI. I'm guessing the only way right now is still to copy first step's input or through a curl request?

axelbdt:
Thanks Ruben. I think deep nesting broke when I changed from inline CSS to tailwind, good catch.

Very much agree with the TODO. For now the focus is pasting a JSON schema in the text editor or inferring from payload, get clear visual feedback in the graphical editor, and being able to keep only a subset of the fields. Just let me know if this scope should be adjusted.

axelbdt:
So tailwind won't include the classes that I define dynamically in TS as the properties get nested deeper (ml-8, ml-12...). So it's back to inline styling for the indentation.

Screenshot 2023-04-29 at 08-59-17 Resources Windmill

rubenfiszel (approved):

PR #1397, opened 2023-04-12
fix(frontend): Remove output when deleting a component

fatonramadani:


frontend/src/lib/components/apps/editor/GridEditor.svelte
```diff
@@ -39,40 +38,6 @@
 			.flatMap((id) => dfs($app.grid, id, $app.subgrids ?? {}))
 			.filter((x) => x != undefined) as string[]
 	}
-
-	function removeGridElement(component) {
```
fatonramadani, line 43: Legacy code when we had a trash bin icon on the component header

PR #1399, opened 2023-04-12
feat(frontend): Add documentation links

fatonramadani:
Screenshot 2023-04-19 at 08 45 28


rubenfiszel:
let's do at least one example

PR #1405, opened 2023-04-13
feat(frontend): Add runs preview popup

adam-kov:
https://user-images.githubusercontent.com/43071496/232077145-06d8329d-7500-4524-b512-38db68247e0a.mov


fatonramadani (approved):

rubenfiszel:
It's almost perfect. What's missing is that jobs scheduled for later are displaying as "still running". One should replace the right log panel with "Scheduled for XXX"

PR #1406, opened 2023-04-13
feat(frontend): Unify main lists

fatonramadani:
https://user-images.githubusercontent.com/456655/231774156-601fcfff-895b-47c9-917c-86c659584328.mov


rubenfiszel:
@fatonramadani proposal, let's remove the 'All' option in there and default to 'App', also let's put 'Go to Hub' on the right

fatonramadani:
> @fatonramadani proposal, let's remove the 'All' option in there and default to 'App', also let's put 'Go to Hub' on the right

Done!

PR #1410, opened 2023-04-14
fix(frontend): fix enum sync

fatonramadani:


frontend/src/lib/components/SchemaEditor.svelte
```diff
@@ -102,7 +103,10 @@
 				schema.required.includes(argName)
 			)
 			oldArgName = argName
-			schemaModal.openDrawer()
+
+			setTimeout(() => {
```
fatonramadani: @rubenfiszel It's one of thoses fixes,wdyt ?

rubenfiszel: wait, why would the setTimeout be on the openDrawer ?

frontend/src/lib/components/SchemaModal.svelte
```diff
@@ -30,6 +30,7 @@
 			default: schema.default,
 			contentEncoding: schema.contentEncoding,
 			format: schema.format,
+			enum_: schema.enum,
```
rubenfiszel, line 33: this looks more like this would the fix no ?

PR #1432, opened 2023-04-19
fix(frontend): Update app default codes

adam-kov:
- added Postgres defaults to charts and tables
- fixed existing ones


frontend/src/lib/components/apps/editor/component/default-codes.ts
```diff
@@ -1,11 +1,15 @@
 import type { AppComponent } from '.'
 
 export function defaultCode(component: string, language: string): string | undefined {
+	console.log(language)
```
gustavopch: Hey, I saw this log and thought I should comment here just in case it was left unintentionally.

rubenfiszel: Good catch!

PR #1458, opened 2023-04-24
fix(frontend): Minor app editor updates

adam-kov:
- updated affected inputs to use the `ClearableInput` component
- updated styling of the "Global styles" pane
  - Sub-tabs and search bar is always visible
  - JSON editor overflowing is fixed
- fixed icon picker popup positioning


frontend/src/lib/components/apps/editor/AppEditor.svelte
```diff
@@ -297,7 +297,7 @@
 										</div>
 									</Tab>
 								</Popover>
-								<div slot="content" class="h-full overflow-y-auto">
+								<div slot="content" class="relative h-full overflow-y-auto">
```
rubenfiszel: that seems like a big change @fatonramadani

adam-kov: Removed it as it wasn't even needed

PR #1483, opened 2023-04-26
fix(frontend): reset ui job loading state when submitting preview job triggers error

ryanrich:
Currently, if an error occurs when trying to submit the preview job the test runner component gets left in a bad state where you cannot retry.  This change attempts to handle the case where the initial submission fails and resets the test runner state so that you can easily retry running the job (and bubble the error so the proper toast handling will kick in).


frontend/src/lib/components/TestJobLoader.svelte
```diff
@@ -46,6 +46,9 @@
 			}
 			return testId
 		} catch (err) {
+			// if error happens on submitting the job, reset UI state so the user can try again
+			isLoading = false
+			job = undefined
```
rubenfiszel:
```suggestion
			currentId = undefined
			job = undefined
```

currentId being undefined will trigger the condition at line 26 and handle more cases

ryanrich: Tried testing with this approach, but the component stays stuck in a state where it thinks it is still running.  I'll add the reset of the `currentId`, but without some other changes to the component I think we need to update `isLoading`

rubenfiszel: Sounds good, thanks for the PR!

PR #1485, opened 2023-04-27
fix(frontend): register applyConnection as a callback to remove unnec…

fatonramadani:
…essary reactivit

### DIFF
--- a/frontend/src/lib/components/apps/editor/appUtils.ts
+++ b/frontend/src/lib/components/apps/editor/appUtils.ts
@@ -593,6 +593,10 @@ export function connectInput(
 	path: string
 ): ConnectingInput {
 	if (connectingInput) {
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/apps/editor/appUtils.ts
```diff
@@ -593,6 +593,10 @@ export function connectInput(
 	path: string
 ): ConnectingInput {
 	if (connectingInput) {
+		if (connectingInput.onConnect) {
+			connectingInput.onConnect({ componentId, path })
+		}
```
rubenfiszel, line 598: we should probably console.warm here

rubenfiszel: in the else

PR #1492, opened 2023-04-28
Fix frontend scripts settings

fatonramadani:


frontend/src/lib/components/apps/components/inputs/AppMultiSelect.svelte
```diff
@@ -48,7 +48,7 @@
 		}
 	}
 
-	$: value && outputs?.result.set(value.map((v) => v.value))
+	$: value ? outputs?.result.set(value.map((v) => v.value)) : outputs?.result.set([])
```
fatonramadani, line 51: @rubenfiszel

PR #1518, opened 2023-05-03
feat(frontend): app modal

fatonramadani:
https://user-images.githubusercontent.com/456655/235917458-d03f8ca4-9a13-4c4c-b607-3f574e17b6a4.mov


frontend/src/lib/components/apps/editor/component/components.ts
```diff
@@ -1686,6 +1688,60 @@ Hello \${ctx.username}
...
+					fieldType: 'boolean',
+					type: 'static',
+					onlyStatic: true,
+					value: false
+				},
+				disabled: {
+					fieldType: 'boolean',
+					type: 'static',
+					value: false
+				}
```
rubenfiszel: we should add tooltip, also might be worth naming all those:
buttonSize, buttonFillContainer etc to avoid confusion

rubenfiszel: we should add tooltip, also might be worth naming all those:
buttonSize, buttonFillContainer etc to avoid confusion

PR #1522, opened 2023-05-03
feat(frontend): add eval badge + alert

fatonramadani:
Screenshot 2023-05-03 at 14 56 46


frontend/src/lib/components/apps/editor/settingsPanel/script/shared/ScriptTriggers.svelte
```diff
@@ -114,6 +116,12 @@
 					</span>
 				{/each}
 			</div>
+			{#if changeEvents.includes('Eval')}
+				<Alert type="info" title="Input evaluated at runtime" size="xs" class="mt-2">
+					At least one input is configured as an evaluated input. This means that the script will
+					run when the input value changes.
+				</Alert>
+			{/if}
```
rubenfiszel: would remove entirely

rubenfiszel: Instead, either adding a Tooltip in the eval badge, or a popover directly on the eval badge itself with the following text:
```
At least one input is configured as an evaluated input and the component will be triggered if the result of the eval change. Eval expressions are re-evaluated on any output or state changes.
```

fatonramadani: Done!

frontend/src/lib/components/apps/editor/settingsPanel/script/shared/ScriptTriggers.svelte
```diff
@@ -96,6 +99,13 @@
 				{#each changeEvents as changeEvent}
 					<span class={classNames(badgeClass, colors['blue'])}>
 						{changeEvent}
+						{#if changeEvents.includes('Eval')}
+							<Tooltip class="!text-blue-600 ml-1">
```
rubenfiszel, line 103: you're adding the tooltip to every badge.

changeEvent == 'Eval'

PR #1555, opened 2023-05-09
feat(frontend): conditional rendering

fatonramadani:


frontend/src/lib/components/apps/components/layout/AppConditionalWrapper.svelte
```diff
@@ -0,0 +1,87 @@
...
+
+	let resolvedConditions: boolean[] = []
+	let selectedConditionIndex = 0
+	let forcedIndex: number = -1
+
+	function handleResolvedConditions() {
+		const slicedArray = resolvedConditions.slice(0, conditions.length)
+		outputs.conditions.set(slicedArray, true)
+
+		const firstTrueIndex = forcedIndex !== -1 ? forcedIndex : slicedArray.findIndex((c) => c)
```
rubenfiszel: this would make the tab stuck as soon as a setTab is used. You can likely just remove forcedIndex and set selectedConditionIndex directly on setTab

PR #1687, opened 2023-06-06
fix(cli): wmill sync would no longer ignore all files

tyrickareo:


rubenfiszel:
And you have tested that this change would make it work ? It's unlikely that this is a fix or that there is a bug on the cli at this level given that it's used in prod. Are you sure that the workspace has been well added ? What commands are you running ?

tyrickareo:
and all my wm and cli are up to date,1.109.1

tyrickareo:
yes, It works for me. when I change workdir ,I change the code too

rubenfiszel:
You're on windows,  i think the issue is that the CLI is not os agnostic wrt to path right now. The fix would come from there

PR #1754, opened 2023-06-21
Allow cors preflight and credentials

u12206050:
Allows making OPTIONS requests without authentication


rubenfiszel:
Couldn't find documentation about:
```
        .allow_preflight(true);
 ```

I'm a little bit worried about the security implication of being able to reuse the user cookie to trigger any script/flow. I will have to think about it more.

PR #1800, opened 2023-07-06
feat: bun support

rubenfiszel:
Currently blocked on not being able to make it work with nsjail


rubenfiszel:
Blocked on https://github.com/oven-sh/bun/issues/3536

PR #1816, opened 2023-07-09
feat: Per script concurrency limit with time window

gbouv:
On a per script basis, it is now possible to set a maximum number of runs per unit of time (i.e. you can set maximum of 5 runs per 60 seconds, for example)

- [X] TODO: needs to apply this to flow inline scripts, haven't found a way to store this in the content of the inline script using tags


backend/windmill-worker/src/worker.rs
```diff
@@ -2396,7 +2396,7 @@ pub async fn handle_child(
                         .expect("update worker ping");
                     }
                     let mem_peak = get_mem_peak(pid, nsjail).await;
-                    tracing::info!("{job_id} still running. mem peak: {}kB", mem_peak);
+                    tracing::debug!("{job_id} still running. mem peak: {}kB", mem_peak);
```
rubenfiszel: This was intended to be info

backend/windmill-queue/src/jobs.rs
```diff
@@ -684,9 +690,113 @@ pub async fn pull<R: rsmq_async::RsmqConnection + Clone>(
...
+
     // let rs = rd_string(2);
     // let instant = Instant::now();
 
+    let mut script_path_blocklist: Vec<String> = Vec::new();
+    loop {
+        // at each iteration, we need to ignore the script paths that have already been
+        // pulled and for which the concurrency limit has been reached. Otherwise each
+        // single worker will loop indefinitely here until the concurrency limit is not
+        // reached anymore
```
rubenfiszel: I don't think that's true since you requeue with a later scheduled_for.

I'm actually not sure why you need to change the pull function at all. Requeue is all that's necessary

backend/windmill-queue/src/jobs.rs
```diff
@@ -684,9 +690,113 @@ pub async fn pull<R: rsmq_async::RsmqConnection + Clone>(
...
+
+            let job_custom_concurrent_limit = pulled_job.concurrent_limit.unwrap_or(DEFAULT_JOB_CONCURRENT_LIMIT);
+            let job_custom_concurrency_time_window_s = pulled_job.concurrency_time_window_s.unwrap_or(DEFAULT_JOB_CONCURRENCY_TIME_WINDOW_SECS);
+            
+            let concurrent_jobs_for_this_script = sqlx::query_scalar!(
+                "SELECT COALESCE(completed_count, 0) + COALESCE(running_count, 0) AS total_count
+                FROM
+                    (SELECT script_path, COUNT(*) as completed_count
+                    FROM completed_job
+                    WHERE script_path = $1 AND started_at > (now() - INTERVAL '1 second' * $2)
```
rubenfiszel: you actually do not need to check for time, just that running = true :)

rubenfiszel: it should simplify a lot the logic for the queue, it's just a count where script_path = x

backend/windmill-queue/src/jobs.rs
```diff
@@ -689,9 +690,133 @@ pub async fn pull<R: rsmq_async::RsmqConnection + Clone>(
         }
     }
 
+    let final_workspaces_filter = workspaces_filter.as_str();
```
rubenfiszel: nit: can be inlined

rubenfiszel (approved):
still require frontend changes

PR #1818, opened 2023-07-09
fix(frontend): support special chars in postgresql client [#1775]

### LINKED ISSUE #1775: bug: special characters in postgresql password cause resource to fail
### Describe the bug

Windmill pukes when creating a Resource of type Postgresql and the postgres password contains certain special characters.

When the password contains `@` and I run "Test connection", it returns

    Missing connection parameters: database user

Screenshot from 2023-06-27 10-18-55

When the password contains `#` and I run "Test connection", it returns

    Could not parse the connection string

Screenshot from 2023-06-27 10-20-07

### To reproduce

1. New Resource, type postgresql
2. Fill the credentials using a password that contains `@` and/or `#`
3. Test connection

### Expected behavior

Special chars in the password should be escaped when building the connection string.

### Screenshots

See above

### Browser information

Firefox 114.0.1

### Application version

Windmill EE v1.122.0-9-g03da1f34

### Additional Context

The workaround is to have a postgres user password without certain special characters.

mrl5:
Fix #1775 

tested with:
1. setup fresh postgres15.3 container with `POSTGRES_PASSWORD` `kot@behemot#zonk`
2. get DB URI with
```javascript
const args = { user: 'postgres', password: 'kot@behemot#zonk', host: '0.0.0.0', dbname: 'postgres', sslmode: 'disable' };
const u = new URL(`postgres://${args.user}:${args.password}@${args.host}:${args.port ?? 5432}/${args.dbname}?sslmode=${args.sslmode}`);
u.hash = '';
u.search = `?sslmode=${args.sslmode}`;
u.pathname = args.dbname;
u.host = args.host;
u.password = args.password;
u.username = args.user;

u.toString();
```
3. connect to db
```console
psql 'postgres://postgres:kot%40behemot%23zonk@0.0.0.0/postgres?sslmode=disable'
```

### DIFF
--- a/frontend/src/lib/components/TestConnection.svelte
+++ b/frontend/src/lib/components/TestConnection.svelte
@@ -19,7 +19,14 @@
 				content: `
 import { Client } from 'https://deno.land/x/postgres/mod.ts'
 export async function main(args: any) {
-	const client = new Client("postgres://" + args.user + ":" + args.password + "@" + args.host + ":" + (args.port ?? 5432) + "/" + args.dbname + "?sslmode=" + args.sslmode)
+	const u = new URL(`postgres://${args.user}:${args.password}@${args.host}:${args.port ?? 5432}/${args.dbname}?sslmode=${args.sslmode}`)
+	u.hash = ''
+	u.search = `?sslmode=${args.sslmode}`
+	u.pathname = args.dbname
+	u.host = args.host
+	u.password = args.password
+	u.username = args.user
+	const client = new Client(u.toString())
 	await client.connect()
 	return 'Connection successful'
 }


mrl5:
I also checked if there is something similar for mysql or mongo but didn't found anything similar to modified code

rubenfiszel:
Thanks for the PR, I just reworked that part and the change need to be applied to TestConnection.svelte

Side-note: I'm working on adding mysql as a first class language too :)

frontend/src/lib/components/TestConnection.svelte
```diff
@@ -19,7 +19,14 @@
 				content: `
 import { Client } from 'https://deno.land/x/postgres/mod.ts'
 export async function main(args: any) {
-	const client = new Client("postgres://" + args.user + ":" + args.password + "@" + args.host + ":" + (args.port ?? 5432) + "/" + args.dbname + "?sslmode=" + args.sslmode)
+	const u = new URL(`postgres://${args.user}:${args.password}@${args.host}:${args.port ?? 5432}/${args.dbname}?sslmode=${args.sslmode}`)
```
mrl5, line 22: I wonder if there's anything that's stopping us with using [`ClientOptions`](https://deno.land/x/postgres@v0.17.0/mod.ts?s=ClientOptions) here?

The only reason that I came up with was some backwards compatibility with `args` interface. Simple adapter like this could work:
```diff
-import { Client } from 'https://deno.land/x/postgres/mod.ts'
+import { Client, ClientOptions } from 'https://deno.land/x/postgres/mod.ts';
 export async function main(args: any) {
-       const client = new Client("postgres://" + args.user + ":" + args.password + "@" + args.host + ":" + (args.port ?? 5432) + "/" + args.dbname + "?sslmode=" + args.sslmode)
+       const clientOptions: ClientOptions = {
+               database: args.dbname,
+               port: args.port ?? 5432,
+               hostname: args.host,
+               password: args.password,
+               user: args.user,
+       };
+       const client = new Client(clientOptions)
        await client.connect()
        return 'Connection successful'
 }
```

however there would be issues with mapping [`TLSOptions`](https://deno.land/x/postgres@v0.17.0/mod.ts?s=TLSOptions) to [psql `sslmode`](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION) - **notice lack of `args.sslmode` in proposed adapter**

PR #1821, opened 2023-07-10
fix(frontend): Fix graph view when mulitple graphs are displayed

fatonramadani:


frontend/src/lib/components/graph/svelvet/container/views/GraphView.svelte
```diff
@@ -123,7 +123,7 @@
 	}
 </script>
 
-<div class="zoomable">
+<div class={`zoomable zoomable-${canvasId}`}>
```
rubenfiszel: why not use an id ?

fatonramadani: Fixed

PR #1856, opened 2023-07-11
fix(frontend): only forward css variables

fatonramadani:
Screenshot 2023-07-11 at 17 01 11

Screenshot 2023-07-11 at 17 01 08

### DIFF
--- a/frontend/src/lib/utils.ts
+++ b/frontend/src/lib/utils.ts
@@ -571,3 +571,11 @@ export function isObviousOwner(path: string, user?: UserExt): boolean {
 	}
 	return false
 }
+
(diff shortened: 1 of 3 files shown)


frontend/src/lib/utils.ts
```diff
@@ -571,3 +571,11 @@ export function isObviousOwner(path: string, user?: UserExt): boolean {
 	}
 	return false
 }
+
+export function extractCustomProperties(styleStr: string): string {
```
fatonramadani, line 575: @rubenfiszel I was forwarding the entire css style, where I only needed to forward the CSS variables

PR #1867, opened 2023-07-12
fix(frontend): Fix App multi select render

fatonramadani:


frontend/src/lib/components/apps/svelte-select/lib/ConditionalPortal.svelte
```diff
@@ -1,11 +1,17 @@
 <script lang="ts">
+	import { getContext } from 'svelte'
 	import Portal from 'svelte-portal'
+	import type { AppViewerContext } from '../../types'
 
 	export let condition = false
+
+	const { mode } = getContext<AppViewerContext>('AppViewerContext')
+
+	$: target = $mode === 'preview' ? '#app-editor-select' : 'body'
```
fatonramadani, line 10: @rubenfiszel  This fixes the scrolling issues

PR #1881, opened 2023-07-15
fix(frontend): fix bg script selection

fatonramadani:


frontend/src/lib/components/apps/editor/inlineScriptsPanel/InlineScriptsPanel.svelte
```diff
@@ -39,9 +39,13 @@
 			? findGridItem($app, $selectedComponentInEditor?.split('_')?.[0])
 			: undefined
 
-	$: hiddenInlineScript = $app?.hiddenInlineScripts?.findIndex((k_, index) =>
-		$selectedComponentInEditor?.startsWith(BG_PREFIX + index)
```
fatonramadani, line 43: `startsWith`will be true when $selectedComponentInEditor is bg_10 and index is 1

rubenfiszel:
Facepalm
I might have done this in a few place. Could you check everywhere we use starts with or bg ?

fatonramadani:
> 

Sure, I'll do that!

PR #1884, opened 2023-07-16
refactor(backend): alternative approach for sqlx ^0.7 upgrade

mrl5:
related to
* https://github.com/launchbadge/sqlx/pull/2617#issuecomment-1636808249
  * https://github.com/launchbadge/sqlx/pull/2617#issuecomment-1637005378


mrl5:
not sure if this is desired but it's worth to be aware of such alternative

PR #1932, opened 2023-07-24
fix required fields in nested schema

axelbdt:
Someone forgot to handle the required fields in nested object schemas :)

Here is how we represent "address field is not required but when it is provided, address.city is required".
Screenshot 2023-07-24 at 17-09-11 Resources Windmill


rubenfiszel:
Thank you!

PR #1941, opened 2023-07-26
fix: add property while viewing as JSON (+ ui tweaks)

axelbdt:
When viewing schema as JSON and adding property via the "Add property button", JSON schema would not show the newly added property. PR fixes this behavior.

Minor layout tweaks to remove unnecessary whitespace and scroll jitter.

af-error


rubenfiszel:
Thanks again!

PR #2007, opened 2023-08-07
feat(frontend): add disabled prop to select input

### LINKED ISSUE #1998: feature: Expose `Disabled` on more components
Thank you so much for the great work so far, my prototypes are proving to be very effective!

### How it works currently:
I am using an inline script as a work around to disable a Select component whilst the options are loaded from an expensive API call:
```js
// disable Select input element
const selectInput = document.querySelector(".svelte-select > .value-container > input");
selectInput.disable = true;
// Trigger expensive API call
recompute("bg_1");
```

This isn't the best for the end users, as they can click the select before the inline script is called; which I let them know of via a default option:
image

### Requested Changes:
I would like to be able to access the `disable` parameter of the `svelte-select` component in the front end, just like Button components:
image

I suppose the same goes for other components; Input, Stepper's buttons etc.

dm-canteen:
Fix #1998 for the Select Input component:
disabled-select-input-2

This was surprisingly simple! Let me know of any issues.

### DIFF
--- a/frontend/src/lib/components/apps/components/inputs/AppSelect.svelte
+++ b/frontend/src/lib/components/apps/components/inputs/AppSelect.svelte
@@ -203,6 +203,7 @@
 					css?.input?.style}
 				{value}
 				placeholder={resolvedConfig.placeholder}
+				disabled={resolvedConfig.disabled}
 				on:focus={() => {
 					if (!$connectingInput.opened) {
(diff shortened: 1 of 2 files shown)


dm-canteen:

rubenfiszel:
Looks good, thanks!

PR #2024, opened 2023-08-10
fix: sync dark-mode icon across multiple renders

hmacr:


hmacr:

hmacr:
recheck

hmacr:
@rubenfiszel I've made this change to sync the dark-mode / light-mode icons across the renders in login-form, header, side-bar (and anywhere else where it could be rendered in the future).
Please take a look and let me know if this looks good.

fatonramadani:
> @rubenfiszel I've made this change to sync the dark-mode / light-mode icons across the renders in login-form, header, side-bar (and anywhere else where it could be rendered in the future). Please take a look and let me know if this looks good.

Hey @hmacr, thanks for the contribution! Quick question, where were the dark mode toggles out-of-sync?

hmacr:
@fatonramadani I saw it on the login page (the header and login form icons were out-of-sync). From the code, I saw that it could be out-of-sync in the sidebar as well.

Screenshot 2023-08-10 at 7 09 52 PM

fatonramadani:
@hmacr good catch!

hmacr:
@fatonramadani Along the same lines, I was wondering if a user opens windmill for the first time, would it be better to get the theme preference from their OS setting instead of defaulting to `light` mode always?
I can help with the implementation if you think this would be a better UX.

fatonramadani:
> @fatonramadani Along the same lines, I was wondering if a user opens windmill for the first time, would it be better to get the theme preference from their OS setting instead of defaulting to `light` mode always? I can help with the implementation if you think this would be a better UX.

Sure, it would improve the current implementation, we would need to read `window.matchMedia('(prefers-color-scheme: dark)')` if we have no 'dark-mode' in the localstorage. Feel free to add it to this PR if you want. Thanks!

hmacr:
@fatonramadani I've added the OS theme preference code, please check. Thanks!

fatonramadani (approved):

PR #2032, opened 2023-08-11
Various fix

fatonramadani:


frontend/src/lib/components/ArgInput.svelte
```diff
@@ -90,6 +90,8 @@
 					value = false
 				} else if (inputCat == 'list') {
 					value = []
+				} else if (inputCat == 'date-time') {
+					value = new Date().toISOString()
```
rubenfiszel: that's not a sufficient fix, we need to initialize the time to 00:00 AM in the date time input (instead of showing -- -- for time), it enables picking only the date and still making a valid date-time

PR #2046, opened 2023-08-13
fix(frontend): fix copyToClipboard on non-HTTPS site

ImSingee:
`navigator.clipboard` is only available on HTTPS sites if the host is not `localhost`.

Though there are some implements like https://stackoverflow.com/a/65996386/8398770, using a well-tested third-party library is better (it can handle many edge cases).  

This PR uses [clip-to-keyboard](https://www.npmjs.com/package/copy-to-clipboard) npm package for this.


ImSingee:

rubenfiszel:
Looks like it uses execCommand first which seems to be deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#
Would rather not import a library just for this

ImSingee:
@rubenfiszel I removed the 3rd-party library and used `navigator.clipboard` if possible

rubenfiszel:
Thanks!

PR #2095, opened 2023-08-15
Tanstack init config for width will now be applied to columns

SindreSvendby:
I was not able to find any downside with this approach. 

Will add a PR to the docs repo also where this change will be used.

### DIFF
--- a/frontend/src/lib/components/apps/components/display/table/AppTable.svelte
+++ b/frontend/src/lib/components/apps/components/display/table/AppTable.svelte
@@ -332,6 +332,7 @@
 												on:keydown={() => toggleRow(row, rowIndex)}
 												on:click={() => toggleRow(row, rowIndex)}
 												class="p-4 whitespace-pre-wrap truncate text-xs text-primary"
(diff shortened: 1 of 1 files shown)


SindreSvendby:

SindreSvendby:
Ok, some more context about this: 

What does this solve:
The current table component do not support any way of giving the user a way to help specify the width of each column.
This change will make it so the `columnSizing` prop in the `Inital state`  field will work. 

Note: as you will from the sceenshots you are only spes the desire behavior, it's still tanstack logic that do the final calulations and it respect some for of minimum width.



#### Current behavior, columnSizing gives no effect
column-size-no-effect


#### Smaller id and large name 
column-size-1
happy case, everything works as expected

#### Names is given a to large value

column-size-2
as you can see, it will only reduce the other fields 

#### Competing columns with to large values
column-size-3
Tanstack is able to understand the relations between the id and name, and will give them a width that respects the table width, minSize and the proportional size between requested width of the columns


Code:
`cell.column` always exist in this context, it's inside a if that checks if column exist.
`getSize()` always return a number (at least that's what TS is saying)

PR #2097, opened 2023-08-16
make tooltips non-gender specific

rylincoln:
Just a small change as I noticed when adding users of my team who are female that 3 tooltips were gender specific.  nbd


rubenfiszel:
Thanks for the PR. Can you agree to the CLA and we will merge this.

rylincoln:

recheck

PR #2200, opened 2023-08-30
Fix run page navigation

fatonramadani:


rubenfiszel:
It doesn't implement forward but I guess that can be a separate PR

PR #2218, opened 2023-09-01
fix(frontend): add virtual list

fatonramadani:


frontend/src/lib/components/runs/RunsTable.svelte
```diff
@@ -57,44 +58,102 @@
...
+
+	$: {
+		flatJobs = []
+		for (const [date, jobsByDay] of Object.entries(groupedJobs)) {
+			flatJobs.push({ type: 'date', date })
+			for (const job of jobsByDay) {
+				flatJobs.push({ type: 'job', job })
+			}
+		}
+	}
```
rubenfiszel: write separate function and triggers otherwise it's super hard to read the dependency

PR #2313, opened 2023-09-20
feat: improved dedicated benchmarks + buffer fix

HugoCasa:


backend/windmill-worker/src/worker.rs
```diff
@@ -549,43 +552,179 @@ pub async fn run_worker<R: rsmq_async::RsmqConnection + Send + Sync + Clone + 's
...
+            #[cfg(feature = "benchmark")]
+            let send_duration = send_duration2.clone();
+            #[cfg(feature = "benchmark")]
+            let process_duration = process_duration.clone();
+            #[cfg(feature = "benchmark")]
+            let completed_jobs = completed_jobs.clone();
+            #[cfg(feature = "benchmark")]
+            let main_duration = main_duration2.clone();
+
+            tokio::spawn(async move {
```
rubenfiszel: we should spawn only if dedicated_worker is set (which can be a static boolean set in a parent) or the job kind is Noop. Otherwise do directly inside while body

PR #2339, opened 2023-09-26
feat(frontend): add AppDrawer controls

fatonramadani:


frontend/src/lib/components/apps/utils.ts
```diff
@@ -180,6 +180,8 @@ declare function setValue(id: string, value: any): void;
 declare function setSelectedIndex(id: string, index: number): void;
 declare function openModal(id: string): void;
```
rubenfiszel: Remove

PR #2352, opened 2023-09-28
feat(frontend): copy schema from json and past runs in flow inputs

fatonramadani:
https://github.com/windmill-labs/windmill/assets/456655/ff489edd-a03b-47f3-b76b-b696d3a8d09a

- Updated the Range library
- Fix disabled


rubenfiszel (changes requested):
monaco autocompletion float at wrong place, pass the right props to SimpleEditor

PR #2353, opened 2023-09-28
feat(backend): parse expires_in from string in TokenResponse

denglertai:
This change allows parsing expires_in from a string while also retaining its previous behavior

This works for the following cases:
- the json response contains `expires_in` as a numeric value (as expected before)
- the json response contains `expires_in` as a string 
- the json response does not contain `expires_in` at all

### DIFF
--- a/backend/windmill-common/src/more_serde.rs
+++ b/backend/windmill-common/src/more_serde.rs
@@ -47,7 +47,7 @@ where
     enum NumericOrNull<'a, T> {
         String(String),
         Str(&'a str),
-        FromStr(T),
+        RawT(T),
         Null,
     }
 
@@ -60,7 +60,7 @@ where
             "" => Ok(None),
             _ => T::from_str(s).map(Some).map_err(serde::de::Error::custom),
         },
-        NumericOrNull::FromStr(i) => Ok(Some(i)),
+        NumericOrNull::RawT(i) => Ok(Some(i)),
         NumericOrNull::Null => Ok(None),
     }
 }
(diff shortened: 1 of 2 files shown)


denglertai:

denglertai:

backend/windmill-common/src/more_serde.rs
```diff
@@ -33,3 +35,32 @@ pub fn default_id() -> String {
...
+    D: Deserializer<'de>,
+    T: FromStr + serde::Deserialize<'de>,
+    <T as FromStr>::Err: Display,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum NumericOrNull<'a, T> {
+        String(String),
+        Str(&'a str),
+        FromStr(T),
```
rubenfiszel: Not sure to follow why this is called `FromStr`, shouldn't it be called `Raw` or equivalent ?

rubenfiszel (dismissed):

backend/windmill-common/src/more_serde.rs
denglertai: I'm actually new to rust, but that was the best thing I got to work. If you got anything better, feel free to add

rubenfiszel: It's very good code, I was just wondering about the naming of this enum case. It seems that this branch represent the raw value and could be named as such

rubenfiszel:
Thanks for the contribution. Did you test all 3 cases ?

denglertai:
Yes. All 3 cases were working fine in my tests

PR #2357, opened 2023-09-28
Update lib.rs with mosaicml-cli for Python

bgoosmanviz:
https://pypi.org/project/mosaicml-cli/ remaps to mosaicml_cli

Cannot use `mcli` because it is already a pypi package. Unfortunately all the MosaicML examples use mcli, so this is the best I could do.

```
import mosaicml_cli
# mosaicml_cli.predict()
```


rubenfiszel:
This mapping should be more or less the one done by default so I'm not sure to fully follow. Did you test this?

bgoosmanviz:
Thanks for pushing me to double-check.

```
#requirements:
#mosaicml-cli
import mcli
```

This works ^. I'll close.

bgoosmanviz:
#requirements:
#mosaicml-cli
import mcli

PR #2371, opened 2023-10-02
fix: add reserved variables in args

HugoCasa:


backend/windmill-api/src/resources.rs
```diff
@@ -322,8 +331,22 @@ async fn get_resource_value_interpolated(
...
+        let job = if let Some(job_id) = job_info.job_id {
+            let mut tx = user_db.clone().begin(&authed).await?;
+            let result = sqlx::query_as::<_, QueuedJob>(
+                "SELECT * FROM queue WHERE id = $1 AND workspace_id = $2",
+            )
+            .bind(job_id)
+            .bind(&w_id)
+            .fetch_optional(&mut *tx)
+            .await?;
+            tx.commit().await?;
```
rubenfiszel: you do not need to load the full job (which may contain some info like the raw_flow !

backend/windmill-worker/src/common.rs
```diff
@@ -84,7 +85,10 @@ pub async fn transform_json_value(
     client: &AuthedClient,
     workspace: &str,
     v: Value,
+    job: &QueuedJob,
+    db: &Pool<Postgres>,
 ) -> error::Result<Value> {
+    tracing::info!("transform_json_value {} {}", name, v);
```
rubenfiszel: you left that

PR #2386, opened 2023-10-04
Upgrade debian version to bookworm

ImSingee:
I have tested this on CE version and make sure it can build (but not test for running) on EE version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster-slim as nsjail
+FROM debian:bookworm-slim as nsjail
 
 WORKDIR /nsjail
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -101,7 +101,7 @@ RUN [ "$TARGETPLATFORM" == "linux/amd64" ] && curl -Lsf https://github.com/denol
 
 RUN unzip deno.zip && rm deno.zip
 
-FROM python:3.11.4-slim-buster
+FROM python:3.12.0-slim-bookworm
```
rubenfiszel: that's too many upgrades in one PR :)

PR #2406, opened 2023-10-09
fix: More explicit error message when executable not found

gbouv:
Before:
Screenshot 2023-10-09 at 15 11 13

After:
Screenshot 2023-10-09 at 15 09 22


gbouv:

PR #2449, opened 2023-10-13
fix: Error handler script pickers lists both "regular" script and "error handler" scripts

gbouv:


frontend/src/lib/components/ScriptPicker.svelte
```diff
@@ -48,7 +48,9 @@
 				label: `${flow.path}${flow.summary ? ` | ${truncate(flow.summary, 20)}` : ''}`
 			}))
 		} else if (itemKind == 'script') {
-			items = (await ScriptService.listScripts({ workspace: $workspaceStore!, kind })).map(
+			items = (await ScriptService.listScripts({ workspace: $workspaceStore! })).filter(
+				(script) => kinds.includes(script.kind)
+			).map(
```
gbouv: alternatively, we would update the BE endpoint to accept multiple "kind" here

rubenfiszel: I think it's fine

PR #2454, opened 2023-10-16
fix: Slack token is readable by g/error_handler

gbouv:


backend/windmill-api/src/workspaces.rs
```diff
@@ -831,6 +832,21 @@ async fn edit_error_handler(
     .execute(&mut *tx)
     .await?;
 
+    if ee.error_handler.as_ref().unwrap() == slack_error_handler_path {
```
rubenfiszel: dangerous unwrap here. is_some_and

backend/windmill-api/src/workspaces.rs
```diff
@@ -831,6 +832,21 @@ async fn edit_error_handler(
...
+        // Give read access to the 'slack_bot' folder (containing the slack bot token resource) to the created 'g/error_handler'
```
rubenfiszel: should verify that it's not already the case otherwise will keep appending forever

backend/windmill-api/src/workspaces.rs
```diff
@@ -831,6 +832,21 @@ async fn edit_error_handler(
...
+        sqlx::query!(
```
rubenfiszel: it's probably the wrong approach btw, we should write it as a migration for existing folders and at creation of the slack folders

backend/windmill-api/src/workspaces.rs
gbouv: FSA I tested running this query multiple times and it doesn't append again and again. It's probably because it's treated as a dict so adding a key that already exists just updates its value. Anyways, changing the approach so this doesn't matter

backend/windmill-api/src/workspaces.rs
```diff
@@ -811,7 +811,7 @@ async fn edit_error_handler(
 
     #[cfg(not(feature = "enterprise"))]
     if ee.error_handler.is_some()
```
rubenfiszel: can remove the is_some

gbouv: 🤦

PR #2458, opened 2023-10-16
feat: Mute workspace error handler for flows and scripts

gbouv:


backend/windmill-api/openapi.yaml
```diff
@@ -6467,6 +6523,8 @@ components:
           type: number
         dedicated_worker:
           type: boolean
+        ws_error_handler_enabled:
+          type: boolean
```
rubenfiszel: think it make more sense to store ws_error_handler_muted since by default it is enabled

PR #2483, opened 2023-10-22
feat: add unique id

HugoCasa:
- need to adjust to the right cli version in backend/migrations/20231022165924_update_hub_sync_stats.up.sql


backend/migrations/20231022165924_update_hub_sync_stats.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Add up migration script here
+UPDATE script SET content = 'import wmill from "https://deno.land/x/wmill@v1.189.0/main.ts";
```
hugocasa, line 2: TODO change to right next version

PR #2487, opened 2023-10-23
fix: Update openapi definition to be compatible with oapi-codegen

gbouv:


backend/windmill-api/openapi.yaml
```diff
@@ -8035,3 +8033,9 @@ components:
 
       required:
         - access_token
+
+    HubScriptKind:
```
gbouv, line 8037: We could re-use it also in `schemas.Script` and `schemas.NewScript` but those two have `command`. Leaving it out for now

PR #2510, opened 2023-10-25
fix(frontend): increased size limit for images

fatonramadani:


frontend/src/lib/components/DisplayResult.svelte
```diff
@@ -61,37 +61,40 @@
...
-		if (largeObject) {
-			return 'json'
-		}
 
 		if (result) {
 			try {
 				let keys = Object.keys(result)
+
+				// Check if the result is an image
+				if (['png', 'svg', 'jpeg'].includes(keys[0])) {
```
rubenfiszel: && keys.length == 1

fatonramadani: done

PR #2556, opened 2023-11-03
fix: fail on non-integer sleep value

invakid404:
This pull request introduces a bug fix where previously, a non-integer value
provided for the sleep duration would be silently ignored because `as_u64()`
would return `None` when `n` was a float, leading to an unintended skip over the
sleep behavior.


invakid404:

PR #2579, opened 2023-11-07
feat: Add override all schedule handlers button

gbouv:
Screenshot 2023-11-07 at 11 01 12


gbouv:
We could potentially also add a `Reset` option in the dropdown to reset the default, TBD

PR #2607, opened 2023-11-10
feat(frontend): draft script/flow can only access the edit page until…

fatonramadani:
… deployed

### DIFF
--- a/frontend/src/lib/components/common/button/ButtonDropdown.svelte
+++ b/frontend/src/lib/components/common/button/ButtonDropdown.svelte
@@ -6,7 +6,7 @@
 	import Portal from 'svelte-portal'
 
 	export let hasPadding: boolean = true
-	export let target: string | undefined = undefined
+	export let target: string | undefined = 'body'
 	const [popperRef, popperContent] = createPopperActions({ placement: 'auto' })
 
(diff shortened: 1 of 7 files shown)


frontend/src/lib/components/common/button/ButtonDropdown.svelte
```diff
@@ -6,7 +6,7 @@
 	import Portal from 'svelte-portal'
 
 	export let hasPadding: boolean = true
-	export let target: string | undefined = undefined
+	export let target: string | undefined = 'body'
```
fatonramadani, line 9: @rubenfiszel an undefined target would throw an error

frontend/src/routes/(root)/(logged)/flows/get/[...path]/+page.svelte
```diff
@@ -45,9 +45,14 @@
...
 	let path = $page.params.path
 	let shareModal: ShareModal
-
 	let deploymentInProgress = false
 
+	$: {
+		if (flow?.draft_only) {
+			goto(`/flows/edit/${flow?.path}?nodraft=true`)
+		}
+	}
```
Comment (rubenfiszel) [inline frontend/src/routes/(root)/(logged)/flows/get/[...path]/+page.svelte]: remove, we should ony remove the way to get to this page, same for below

PR #2612, opened 2023-11-10
fix: Invalid config for workers does not panic

gbouv:
We now "gracefully crash" the worker:
Screenshot 2023-11-10 at 13 18 59


backend/windmill-common/src/worker.rs
```diff
@@ -178,16 +186,27 @@ pub async fn load_worker_config(db: &DB) -> error::Result<WorkerConfig> {
...
+        .dedicated_worker
+        .map(|x| {
+            let splitted = x.split(':').to_owned().collect_vec();
+            if splitted.len() != 2 {
+                killpill_tx.send(()).expect("send");
+                let error_msg = format!(
+                    "Invalid dedicated_worker format. Got {}, expects <workspace_id>:<path>",
+                    x
+                );
+                return Err(Error::BadRequest(error_msg));
```
rubenfiszel: That's not a bad request :D
anyhow! or BadConfig

PR #2621, opened 2023-11-13
feat: cache postgres connection

rubenfiszel:
- feat: cache pg connection
- fix wmill dev

### DIFF
--- a/backend/windmill-worker/src/pg_executor.rs
+++ b/backend/windmill-worker/src/pg_executor.rs
@@ -1,3 +1,7 @@
+use std::sync::atomic::{AtomicBool, AtomicU64};
+use std::sync::Arc;
+use std::time::Duration;
+
 use anyhow::Context;
 use chrono::Utc;
 use futures::TryStreamExt;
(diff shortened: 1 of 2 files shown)


backend/windmill-worker/src/pg_executor.rs
```diff
@@ -146,8 +192,55 @@ pub async fn do_postgresql(
         .into_iter()
         .map(postgres_row_to_json_value)
         .collect::<Result<Vec<_>, _>>()?);
+    RUNNING.store(false, std::sync::atomic::Ordering::Relaxed);
```
gbouv, line 195: If one of the statement with an ending `?` fails above, I guess the function will return early and the RUNNING won't be set back to false. Do we risk something having RUNNING always true?

backend/windmill-worker/src/pg_executor.rs
```diff
@@ -146,8 +192,55 @@ pub async fn do_postgresql(
...
+                        break;
+                    }
+                    let mtex = CONNECTION_CACHE.lock().await;
+                    if mtex.is_none() {
+                        // connection is not in the mutex anymore
+                        break;
+                    } else if let Some(mtex) = mtex.as_ref() {
+                        if mtex.0.as_str() != &database_string_clone {
+                            // connection is not the latest one
+                            break;
```
gbouv, line 230: Noice

gbouv (approved):

PR #2637, opened 2023-11-16
feat: dedicated workers for flows

rubenfiszel:


backend/windmill-worker/src/go_executor.rs
```diff
@@ -215,8 +221,14 @@ func Run(req Req) (interface{{}}, error){{
         }
     } else {
         let path = bin_path.unwrap().clone();
-        logs.push_str(&format!("found cached binary: {}\n", path));
-        tokio::fs::copy(path, format!("{job_dir}/main")).await?;
+        logs.push_str(&format!("found cached binary: {path}\n"));
+        tokio::fs::copy(&path, format!("{job_dir}/main"))
```
gbouv, line 225: nice

backend/windmill-worker/src/worker.rs
```diff
@@ -1086,163 +1088,96 @@ pub async fn run_worker<R: rsmq_async::RsmqConnection + Send + Sync + Clone + 's
...
-                            worker_envs,
-                            &_wp.workspace_id,
-                            &_wp.path,
-                            &token,
-                            job_completed_tx,
-                            dedicated_worker_rx,
-                            killpill_rx,
-                        )
-                        .await
+                    tokio::time::sleep(Duration::from_millis(10000)).await;
```
gbouv, line 1146: Not sure I fully understand why we're spleeping after starting the dedicated worker here 🤔

backend/windmill-worker/src/worker.rs
```diff
@@ -1746,6 +1691,317 @@ pub async fn run_worker<R: rsmq_async::RsmqConnection + Send + Sync + Clone + 's
...
+                    }
+                }
+                r
+            }
+            SpawnWorker::RawScript { content, lock, lang, .. } => (content, lock, Some(lang), None),
+        };
+
+        match language {
+            Some(ScriptLang::Python3) | Some(ScriptLang::Bun) | Some(ScriptLang::Deno) => {}
+            _ => return None,
```
gbouv, line 1924: I can't find just looking at the code where we handle the error if the flow contains one module in a non supported language, but I'm sure you're handling that case since we discussed it yesterday

HugoCasa (approved):

gbouv (approved):

backend/windmill-worker/src/worker.rs
rubenfiszel: It's the None case here, will discuss in daily

backend/windmill-worker/src/worker.rs
rubenfiszel: we're sleeping only if we cannot fetch the flow, and looping thereafter

PR #2650, opened 2023-11-19
feat(python): Add functionality and resiliency to wmill python client

knowsuchagency:
This script updates the `wmill` client to add additional arguments to the methods for running scripts synchronously.

* `assert_result_is_not_none`
* `cleanup`
* `timeout`

`assert_result_is_not_none` asserts that the result of the script isn't null. The current behavior is to raise an exception if the result is falsey, which causes unnecessary failures if the result is something like 0, probably a bug.

`cleanup` ensures that jobs that are cancelled gracefully when the parent script is cancelled.

`timeout` throws a `TimeoutError` if passed and the script takes longer than it to execute

### DIFF
--- a/python-client/wmill/wmill/client.py
+++ b/python-client/wmill/wmill/client.py
@@ -1,8 +1,12 @@
 from typing import Any, Union, Dict
-from typing import Generic, TypeVar
+from typing import Generic, TypeVar, Optional
 
 import os
 import json
+from datetime import timedelta
+import logging
+import atexit
+import time
 
 from time import sleep
 from windmill_api.models.whoami_response_200 import WhoamiResponse200
@@ -27,6 +31,8 @@ class JobStatus(Enum):
 
 
 _client: "AuthenticatedClient | None" = None
+
+logger = logging.getLogger("wmill_client")
 
 
 def create_client(base_url: "str | None" = None, token: "str | None" = None) -> AuthenticatedClient:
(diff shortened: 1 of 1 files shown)


knowsuchagency:

python-client/wmill/wmill/client.py
```diff
@@ -81,21 +87,71 @@ def run_script_async(
...
     """
+    args = args or {}
     job_id = run_script_async(hash, args, None)
+
+    def cancel_job():
+        from windmill_api.api.job.cancel_queued_job import (
+            sync_detailed,
+            CancelQueuedJobJsonBody,
+        )
+        import wmill
```
rubenfiszel: I think this can be removed

python-client/wmill/wmill/client.py
```diff
@@ -81,21 +87,71 @@ def run_script_async(
     ).content.decode("us-ascii")
 
 
-def run_script_sync(hash: str, args: Dict[str, Any] = {}, verbose: bool = False) -> Dict[str, Any]:
+def run_script_sync(
+    hash: str, 
+    args: Optional[Dict[str, Any]] = None, 
+    verbose: bool = False,
+    assert_result_is_not_none: bool = True,
+    cancel_atexit: bool = True,
```
rubenfiszel: cancel_at_exit*

I had a hard time reading atexit :)

rubenfiszel (dismissed):
Very cool, only minor changes needed

python-client/wmill/wmill/client.py
knowsuchagency: I'll rename it :)

PR #2652, opened 2023-11-19
fix: main broken tests

gbouv:
Caused by https://github.com/windmill-labs/windmill/commit/b13cc58315ea07d67d98bffb907e2cbd0c1d6b7d#diff-9f2b0f927e0a13c103a4e641c255cc58e4f745cc4cafafb28a7296e8c7f9a1ba I believe

### DIFF
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -961,7 +961,7 @@ fn spawn_test_worker(
     tokio::sync::broadcast::Sender<()>,
     tokio::task::JoinHandle<()>,
 ) {
-    for x in [windmill_worker::LOCK_CACHE_DIR] {
+    for x in [windmill_worker::LOCK_CACHE_DIR, windmill_worker::GO_BIN_CACHE_DIR] {
         std::fs::DirBuilder::new()
             .recursive(true)
             .create(x)
--- a/backend/windmill-worker/src/go_executor.rs
+++ b/backend/windmill-worker/src/go_executor.rs
@@ -50,7 +50,7 @@ pub async fn handle_go_job(
 ) -> Result<Box<RawValue>, Error> {
     //go does not like executing modules at temp root
     let job_dir = &format!("{job_dir}/go");
-    let bin_path = Some(format!(
+    let bin_path = format!(
         "{}/{}",
         GO_BIN_CACHE_DIR,
         calculate_hash(&format!(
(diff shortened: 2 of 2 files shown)


backend/windmill-worker/src/go_executor.rs
```diff
@@ -215,12 +210,11 @@ func Run(req Req) (interface{{}}, error){{
         )
         .await?;
 
-        if let Some(bin_path) = bin_path.clone() {
-            tokio::fs::copy(format!("{job_dir}/main"), &bin_path).await?;
-            logs.push_str(&format!("write cached binary: {}\n", bin_path));
-        }
+        create_dir(&bin_path).await?;
```
gbouv, line 213: We're in the `if !bin_exists` clause here, so we need to create the folder before we can copy the binary to it

backend/windmill-worker/src/go_executor.rs
```diff
@@ -215,12 +210,11 @@ func Run(req Req) (interface{{}}, error){{
...
+        tokio::fs::copy(format!("{job_dir}/main"), format!("{bin_path}/main")).await?;
```
gbouv, line 214: And then, the `copy` expects 2 file paths. Doing `copy /path/to/main /dest/path/` won't work, you need to do `copy /path/to/main /dest/path/main`. Same below

backend/tests/worker.rs
```diff
@@ -961,7 +961,7 @@ fn spawn_test_worker(
     tokio::sync::broadcast::Sender<()>,
     tokio::task::JoinHandle<()>,
 ) {
-    for x in [windmill_worker::LOCK_CACHE_DIR] {
+    for x in [windmill_worker::LOCK_CACHE_DIR, windmill_worker::GO_BIN_CACHE_DIR] {
```
gbouv, line 964: Seems weird that we need to treat Go differently here, and TBH I don't quite unerstand why we don't need to to it for the other runtimes (we do it in the main.rs, just not for the test worker). But if we don't do that, this line will fail: https://github.com/windmill-labs/windmill/pull/2652/files#diff-9f2b0f927e0a13c103a4e641c255cc58e4f745cc4cafafb28a7296e8c7f9a1baR213
Even wondering if we shouldn't just not do caching in test mode

gbouv: Actually, thinking more about it, it might be b/c Go is our only compiled runtime. Anyway, I haven't looked too much into the other runtimes implementation of caching, but curious to get your thoughts after your review

rubenfiszel: I believe in main.rs we create those cache dir so it's just that we were too lazy to mimic proper initialization

PR #2661, opened 2023-11-20
Add official Atlassian Python Library Mapping

jacobm001:
The official Atlassian python library is named `atlassian-python-api`, which requires an installation under that name. It is imported as `atlassian`.


jacobm001:

PR #2665, opened 2023-11-21
feat(python): Refactor Windmill Python client (remove `windmill-api`)

knowsuchagency:
This PR obviates the need for the `windmill-api` library. I believe this makes the client package is easier to understand, debug, build, (and test) without it.

Additional updates were made to improve logging and add more robust error handling.

Despite introducing a class-based approach to encapsulate variables that are commonly used with the API, global methods were added with deprecation warnings to maintain backwards-compatibility.

Here's what Jetbrains' AI assistant came up with to describe the changes based on the diff--I think it did a decent job:

> Simplified the Windmill Python client by refactoring out repeated code into more compact, reusable methods. Transitioned the client functions into a Windmill client class, enabling a better encapsulation of the client's state. Updated the README example to reflect this change. This improves code maintainability by making the code easier to understand and update, and improves user experience by providing a more intuitive client interface.


python-client/build.sh
```diff
@@ -1,43 +1,6 @@
...
-[this openapi python client generator](https://github.com/openapi-generators/openapi-python-client/)"\
->> windmill-api/README.md.tmp
-
-echo "" >> windmill-api/README.md.tmp
-
-
-head -n -13 windmill-api/README.md >> windmill-api/README.md.tmp
-mv windmill-api/README.md.tmp windmill-api/README.md
-
-cd windmill-api && poetry build
```
rubenfiszel: we should still generate windmill-api since it can be used outside of wmill

However, i'm thinking of changing the python generator

rubenfiszel:
Assigning @gbouv for testing.

Most important items:
- results are the same as before
- error handling has approximately same behavior

python-client/build.sh
knowsuchagency: sounds good I went ahead and added the code back

python-client/wmill/wmill/client.py
```diff
@@ -1,527 +1,630 @@
...
+
+    def create_token(self, duration=dt.timedelta(days=1)) -> str:
+        endpoint = "/users/tokens/create"
+        payload = {
+            "label": f"refresh {time.time()}",
+            "expiration": (dt.datetime.now() + duration).strftime("%Y-%m-%dT%H:%M:%SZ"),
+        }
+        return self.post(endpoint, json=payload).text
+
+    def create_job(
```
rubenfiszel: Not a fan of "job" since it's an overloaded terms in windmill. Flows run in jobs too

rubenfiszel:
@knowsuchagency do you know if it's compatible with python 3.7?

python-client/wmill/wmill/client.py
knowsuchagency: I was trying to create a distinction since async is often used to denote asynchronous functions in python i.e. those that run in an `asyncio` event loop. Maybe `start_execution`?

knowsuchagency:
@rubenfiszel It should be, although I didn't specifically test it. I saw you'd committed changes to the module to make it 3.7-compatible so I turned on warnings in pycharm to warn me on syntax errors. the `from __future__ import annotations` should allow it to work with the updated annotations, but again, I didn't test with 3.7, specifically

knowsuchagency:
@rubenfiszel it works with 3.7 through 3.12. I ran some basic tests on my own self-hosted windmill deployment using `nox`. 

* https://github.com/knowsuchagency/windmill_client/blob/main/noxfile.py
* https://github.com/knowsuchagency/windmill_client/blob/main/test_client.py

^ it's testing the same code--I just copied and pasted it over

python-client/wmill/wmill/client.py
rubenfiszel: It's just that we would need to mention "script" or "flow" in such function because the api should be consistent, 

create_script_job or create_job_script would also work

rubenfiszel:
@knowsuchagency amazing, I will prioritize doing some sanity checks now, we should be able to merge this quickly

python-client/wmill/wmill/client.py
```diff
@@ -1,527 +1,630 @@
...
-    if cleanup:
-        atexit.unregister(cancel_job)
+    def get_resource(
+        self,
+        path: str,
+        none_if_undefined: bool = False,
+    ) -> str | None:
+        """Get resource from Windmill"""
+        try:
+            return self.get(f"/w/{self.workspace}/resources/get/{path}").json()["value"]
```
rubenfiszel: this should be using get_value_interpolated instead: https://app.windmill.dev/openapi.html#/operations/getResourceValueInterpolated

python-client/wmill/wmill/client.py
```diff
@@ -1,527 +1,630 @@
...
+        if assert_result_is_not_none and result_text is None:
+            raise Exception(f"result is None for {job_id = }")
+        try:
+            return result.json()
+        except JSONDecodeError:
+            return result_text
+
+    def get_variable(self, path: str) -> str:
+        """Get variable from Windmill"""
+        return self.get(f"/w/{self.workspace}/variables/get/{path}").json()["value"]
```
rubenfiszel: this should use get_value instead

PR #2713, opened 2023-11-27
fix: Deno can talk to private NPM registries behind HTTPS

gbouv:


backend/windmill-worker/src/deno_executor.rs
```diff
@@ -41,6 +41,8 @@ lazy_static::lazy_static! {
         .map(|x| format!(";{x}"))
         .unwrap_or_else(|| String::new());
 
+    static ref DENO_CERT: String = std::env::var("DENO_CERT").ok().unwrap_or_else(|| String::new());
+    static ref DENO_TLS_CA_STORE: String = std::env::var("DENO_TLS_CA_STORE").ok().unwrap_or_else(|| String::new());
```
gbouv, line 45: Not sure if we could re-use WHITELIST_ENVS here. It seems to be used only for dedicated workers for now

rubenfiszel: we could use WHITELIST_ENV but this seems better anyway

PR #2767, opened 2023-12-04
feat: Custom content type for script and flow results

gbouv:
This is only for synchronous flow and script execution

### DIFF
--- a/backend/windmill-api/src/jobs.rs
+++ b/backend/windmill-api/src/jobs.rs
@@ -6,6 +6,7 @@
  * LICENSE-AGPL for a copy of the license.
  */
 
+use axum::http::HeaderValue;
 use serde_json::value::RawValue;
 use std::collections::HashMap;
 use windmill_common::flow_status::RestartedFrom;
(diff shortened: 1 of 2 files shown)


backend/windmill-api/src/jobs.rs
```diff
@@ -1934,13 +1942,47 @@ async fn run_wait_result<T>(
...
+                    .unwrap_or(if result_value.is_some() {
+                        Ok(StatusCode::OK)
+                    } else {
+                        Ok(StatusCode::NO_CONTENT)
+                    })?;
+
+                if windmill_content_type.is_some() {
+                    let serialized_result = result_value
+                        .map(|val| val.get().to_owned())
+                        .unwrap_or_else(String::new);
```
rubenfiszel, line 1970: octet stream would not be strings

gbouv: That's true, although the `result` will be extracted from the JSON, so the best users can do is to pass a base64 encoded octet stream, which is a string. Am I missing somthing? How would they return an octet stream right now?

PR #2778, opened 2023-12-04
feat: app reports v2

HugoCasa:


backend/windmill-queue/src/jobs.rs
```diff
@@ -2267,7 +2267,7 @@ pub async fn push<'c, T: Serialize + Send + Sync, R: rsmq_async::RsmqConnection
             priority,
         ),
         JobPayload::ScriptHub { path } => {
-            if path == "hub/7771/slack" {
+            if path == "hub/7836/slack" {
```
rubenfiszel: we should do an OR if ever ppl are already using it

PR #2811, opened 2023-12-08
feat: Add default timeout for instance and custom timeout for scripts

gbouv:


backend/windmill-worker/src/worker.rs
```diff
@@ -239,9 +239,9 @@ lazy_static::lazy_static! {
     pub static ref GOPROXY: Option<String> = std::env::var("GOPROXY").ok();
     pub static ref NETRC: Option<String> = std::env::var("NETRC").ok();
 
-
     pub static ref NPM_CONFIG_REGISTRY: Arc<RwLock<Option<String>>> = Arc::new(RwLock::new(None));
     pub static ref PIP_EXTRA_INDEX_URL: Arc<RwLock<Option<String>>> = Arc::new(RwLock::new(None));
+    pub static ref JOB_DEFAULT_TIMEOUT: Arc<RwLock<i32>> = Arc::new(RwLock::new(0));
```
rubenfiszel: this is confusing, should use None or -1 as default value to avoid ambiguity

backend/windmill-worker/src/common.rs
```diff
@@ -732,6 +711,62 @@ pub async fn start_child_process(mut cmd: Command, executable: &str) -> Result<C
...
+            if Duration::from_secs(default_timeout_secs as u64) < global_max_timeout_duration =>
+        {
+            Duration::from_secs(default_timeout_secs as u64)
+        }
+        default_timeout_secs => {
+            warn_msg = Some(format!("WARNING: Default job timeout of {default_timeout_secs} seconds was greater than the maximum timeout. It will be ignored and the global max timeout will be used instead"));
+            tracing::warn!(warn_msg);
+            global_max_timeout_duration
+        }
+    };
```
rubenfiszel: this is computed even if custom_timeout is set which is a waste

EDIT: nvm

rubenfiszel: nvm

rubenfiszel: actually no, first comment is correct, you shouldn't have to read JOB_DEFAULT_TIMEOUT if custom_timeout is set

PR #2813, opened 2023-12-08
fix: test_complex_flow_restart running a constantly failing flow on CI

gbouv:
The actual fix for the GOROOT error we were seeing on CI is here: https://github.com/windmill-labs/windmill/pull/2814, but this makes the test more robust


gbouv:
I think there's still a flaky test somewhere, but at least this will make it easier to debug, we will see

PR #2822, opened 2023-12-08
feat(python): Updated Windmill client constructor to accept workspace

knowsuchagency:
This commit introduces a change in the Windmill class constructor (in wmill/client.py) to include `workspace` as a parameter. 

It's especially useful when interfacing with the windmill server externally combined with the `base_url` parameter.


rubenfiszel:
Thanks!

PR #2833, opened 2023-12-11
feat: oauth flow openai key

HugoCasa:


frontend/src/routes/(root)/(logged)/workspace_settings/+page.svelte
```diff
@@ -248,6 +250,12 @@
 				? settings.large_file_storage?.s3_resource_path?.replace('$res:', '')
 				: undefined
 		gitSyncResourcePath = settings.git_sync?.git_repo_resource_path?.replace('$res:', '')
+
+		// check openai_client_credentials_oauth
+		const resourceTypes = await ResourceService.listResourceTypeNames({
```
rubenfiszel, line 255: improvement: check for existence of the resource type directly (existsResourceType or getResourceType)

backend/windmill-api/src/openai.rs
```diff
@@ -71,63 +82,218 @@ lazy_static::lazy_static! {
...
 ) -> impl IntoResponse {
-    let mut tx = db.begin().await?;
-    let openai_resource_path = sqlx::query_scalar!(
-        "SELECT openai_resource_path FROM workspace_settings WHERE workspace_id = $1",
-        &w_id
-    )
-    .fetch_one(&mut *tx)
-    .await?;
-    tx.commit().await?;
+    let mut cache = OPENAI_KEY_CACHE.write().await;
```
rubenfiszel, line 158: no that's terrible, you're taking a write lock on the full hashmap. Use quick_cache instead, see the cache of users or webhooks for instance

PR #2834, opened 2023-12-11
feat: Resource type description is now displayed when creating a new resource

gbouv:


rubenfiszel:
@gbouv for this to be complete, we will need to bump the sync script to use the new version of the cli after release

PR #2838, opened 2023-12-12
fix: Simplify worker env var input

gbouv:
https://github.com/windmill-labs/windmill/assets/18011812/cae86ae2-1666-4553-9f67-a46880c7aa80


gbouv:
Actually forgot the preset button. moving it back to draft, will add now

PR #2873, opened 2023-12-17
chore: correct legacy reference from `WM_STATE_PATH_NEW` to `WM_STATE_PATH`

knowsuchagency:
* also replace `unexplicited` with `unspecified` in cancelled job message string


rubenfiszel:
No it's the opposite, the _NEW one is obsolete now that the migration has been complete, both _NEW and without return the same value now and we will remove _NEW later

PR #2900, opened 2023-12-20
chore: add `verify` to Windmill python object for httpx

gbouv:


python-client/wmill/wmill/client.py
```diff
@@ -23,7 +23,7 @@
 
 
 class Windmill:
-    def __init__(self, base_url=None, token=None, workspace=None):
+    def __init__(self, base_url=None, token=None, workspace=None, verify=True):
```
gbouv, line 26: httpx default is `True` -> https://github.com/encode/httpx/blob/08eff926a6011ac84961fd2cee22d9fd4d4b1a88/httpx/_client.py#L629

PR #2918, opened 2023-12-23
chore: Windmill integration/continuity tests

gbouv:


.github/workflows/docker-image.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: |
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:dev
```
rubenfiszel, line 59: we can remove this and use the specific tagged version instead

.github/workflows/docker-image.yml
```diff
@@ -102,7 +102,7 @@ jobs:
             features=enterprise,enterprise_saml
             nsjail=true
           tags: |
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-ee:latest
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-ee:dev
```
rubenfiszel, line 105: same as above

PR #3014, opened 2024-01-16
feat: Set a default app for each workspace

gbouv:


frontend/src/routes/(root)/(logged)/user/(user)/login/+page.svelte
```diff
@@ -116,7 +116,14 @@
 
 			if (allWorkspaces?.length == 1) {
 				$workspaceStore = allWorkspaces[0].id
-				goto(rd ?? '/')
+				let defaultApp = await WorkspaceService.getWorkspaceDefaultApp({
```
rubenfiszel: need to check if an operator first I believe

frontend/src/routes/(root)/(logged)/user/(user)/login/+page.svelte
```diff
@@ -116,11 +116,16 @@
...
+
+				if ($userStore?.operator) {
+					let defaultApp = await WorkspaceService.getWorkspaceDefaultApp({
+						workspace: $workspaceStore!
+					})
+					if (!emptyString(defaultApp.default_app_path)) {
+						goto(`/apps/get/${defaultApp.default_app_path}`)
+					} else {
+						goto(rd ?? '/')
+					}
```
rubenfiszel: @gbouv I don't think userStore is set correctly since we haven't switched workspace yet. You need to get that information by doing an api call. Look how userStore gets it on workspace change

rubenfiszel: also make sure to ignore that check if user is superadmin on the other hand.

Also let's make clear in the workspace settings that the default app is only for operators

PR #3019, opened 2024-01-17
feat: Allow running image as non-root user

AudriusButkevicius:
Currently /tmp/monaco ends up only writable by root, as build context runs as root, and read-only by other users. 
Running the image with a non-root uid fails, as launcher fails to write out go.mod to the root owned directory.
Some environments do not allow running things as root.


AudriusButkevicius:

AudriusButkevicius:
recheck

PR #3087, opened 2024-01-26
fix python init code for trigger

louisabraham:


louisabraham:

PR #3092, opened 2024-01-27
feat: Passing HOME env var through to python workers

AudriusButkevicius:
Also add missing HOME for python.


rubenfiszel:
Thank you for the contribution
However, there is already a WHITELIST_ENVS which overlap too much with this feature (define all the required env variables, then whitelist them)

AudriusButkevicius:
🤦 didn't see that.
Anyways, makes sense to pass home to python like it's done for other languages.

backend/windmill-worker/src/python_executor.rs
```diff
@@ -349,6 +349,7 @@ mount {{
             .env("TZ", TZ_ENV.as_str())
             .env("BASE_INTERNAL_URL", base_internal_url)
             .env("BASE_URL", base_internal_url)
+            .env("HOME", HOME_ENV.as_str())
```
rubenfiszel: remove this, sandboxed executions cannot be overriden

PR #3094, opened 2024-01-28
feat: Support sending SIGINT to jobs

AudriusButkevicius:
This makes it easier to clean up in languages like python, where you get a nice KeyboardInterrupt exception, and don't have to explicitly install signal handlers for SIGTERM, making the scripts easier to reason about.


AudriusButkevicius:
Actually, I just realised that most things have `sigterm` set to false in handle handle_child, why is that?

rubenfiszel:
We actually do the opposite and want to gracefully exit in most cases by waiting for the current job in progress to finish.

AudriusButkevicius:
Sorry, but not sure I follow. 

Seems that sigterm is only set for bash, and unless that is set, it simply sigkills on termination.

Are you suggesting that the sigterm flag should be flipped everywhere?

Also, is there any harm in the new signal? Given this is off by default, it doesn't change the existing behaviour and still allows to use sigint for people that want to avoid installing signal handlers.

AudriusButkevicius:
Any suggestions how to move this forward?

rubenfiszel:
I think there is some misunderstanding.

We explicitely don't want to send termination signal to processes, we want them to finish and then we exit the worker. The only exception would be for super long running jobs, there should be a mechanism to have the tolerance period be less than a configurable amount of time (which is what we do for bash somehow)

AudriusButkevicius:
The use-case I'm referring to is cancellation. 

Currently upon cancellation the jobs are not given an opportunity to clean up and are immediately sigkilled.

If they spawned subprocesses, this leaves orphans eventually leading to worker containers full of run-away processes (actual problem I am observing, cancelling scripts while testing them because I realise I made a mistake, but my scripts spawn subprocesses which are now left as orphans)

Enabling sigterm would atleast give them a chance to clean up.

However sigterm needs explicit signal handler setup, where as sigint is already nicely propagated to the user (as in, propagates somehow without installing an explicit handler) by most runtimes, KeyboardInterrupted in python etc.

So my proposal is to keep existing behaviour as is, but add an environment variable that when set to non-zero sends a sigint and waits some time for for it to exit naturally.

I might be misunderstanding the code, but as far as I understand the code I have modified would only be hit if we're exiting the log tailing loop and the process is still alive - which implies cancellation.

rubenfiszel:
You might be completely right, I need to take a look at the code again. I was so sure that this was the part that killed jobs when the worker get the killpill that I confused myself.

PR #3099, opened 2024-01-29
feat: Download s3 file as stream in Python and TS

gbouv:


backend/windmill-api/src/job_helpers.rs
```diff
@@ -1259,7 +1259,7 @@ async fn read_object_chunk(
 ) -> error::Result<Vec<u8>> {
     let s3_object = s3_client
         .get_object()
-        .range(format!("bytes={}-{}", from_byte, from_byte + length).to_string())
+        .range(format!("bytes={}-{}", from_byte, from_byte + length - 1).to_string())
```
rubenfiszel, line 1262: were we off by one prior?

gbouv: Yes, b/c `bytes=N-M` includes both N and M, weirdly. It was fine before b/c we were not using that function to download consecutive parts of a file

Issue #3131, opened 2024-02-01
bug: CLI 'wmill script run -s' command outputs color ANSI codes in output


istarkov:
### Describe the bug

`wmill script run -s` output contains ansi color codes

### To reproduce

- Create any script in windmill
- Execute using cli

```bash
wmill script run /f/shared/some-script -s -d '{"a":345}' > a.json
```

See `a.json` contains ansi color codes from `log.info` `[34m{SOMEJSONHERE}[39m`

The issue is log library log.info by default adds ansi codes.

Workaround, to execute as
```bash
NO_COLOR=true  wmill script run /f/shared/shared-scrips -s -d '{"a":345}' > a.json
```

### Expected behavior

script run should not contain ansi codes specific to log library in the output at least during silent mode
In that case it can be used without workaround to process output with `jq` or any other scripting.

Example

```bash
wmill script run /f/shared/script -s -d '{"a":345}' | jq '.e'
```
Now fails with `parse error: Invalid numeric literal at line 1, column 2`

In non silent mode it's ok to have colored json output.

istarkov:
To fix: 
I suggest to change this line 
https://github.com/windmill-labs/windmill/blob/21a82d56c7e0bb1acb9031d20cfde4753c40e3d1/cli/script.ts#L414
on 
```js
if (opts.silent) {
  console.log(result)
} else {
  log.info(result)
}
```

PR #3132, opened 2024-02-01
Remove ansi codes in script run silent mode

istarkov:
ref #3131

### DIFF
--- a/cli/script.ts
+++ b/cli/script.ts
@@ -411,7 +411,12 @@ async function run(
             id,
           })
         ).result ?? {};
-      log.info(result);
+
+      if (opts.silent) {
+        console.log(result);
+      } else {
+        log.info(result);
+      }
 
       break;
     } catch {


istarkov:

rubenfiszel:
Thanks for the PR, i'm not sure this is the right approach. Is there no way to patch the logger to not use ansi code instead of forking to console.log ?

istarkov:
Data goes to stdout, bypassing the logger since it's not log-related. It's data which can be used for processing like in most cli tools.

Anyway:

In most cli utilities I know all information messages goes to stderr and essential data needed for piping goes to stdout.

Example:
```bash
# highlight stderr
command 2> >(while read line; do echo -e "\e[01;31m$line\e[0m" >&2; done) curl -v https://example.com
```
Result:
image

See all info messages are in stderr.

For windmill all logs are sent to stdout. 

Better would be output all cli util information messages into stderr. And output essential data like script result into stdout. Bypassing the logger like I did too. But this must be done for every ustility in the app.

rubenfiszel:
I think it's debatable if wmill cli logs should go to stdout or stderr (since wmill is not meant to be pipable) and besides the point of this specific feature.

My point is that we have one level of indirection with the logger that is fully configurable: https://deno.land/std@0.214.0/log/mod.ts

The removal of ansi codes should be done by customizing the formatter there rather than skipping entirely the logger since it will not compose well, were we to want to do further log customization

istarkov:
This would be formatter for the whole logs, and technicaly this is not log ;-).  Logs can be redirected to files, http endpoints etc. Here logs are disabled using silent flag, and again it's raw response. 

> since wmill is not meant to be pipable
what the reason to have `wmill script run` and do not allow std unix operations?

rubenfiszel:
I apologize, I read the PR wrong. I thought this was disabling the whole log

PR #3147, opened 2024-02-05
feat(frontend): add support for toasts in frontend scripts

fatonramadani:


frontend/src/lib/components/apps/components/helpers/eval.ts
```diff
@@ -117,6 +118,7 @@ export async function eval_like(
 			invalidate?: (key: string, error: string) => void
 			validateAll?: () => void
 			clearFiles?: () => void
+			sendUserToast?: (message: string) => void
```
rubenfiszel: I think that's unecessary

PR #3192, opened 2024-02-09
Properly handle pip index urls in pip-compile

doyleish:
- URLs that contain simple auth seem to have that stripped before being passed to underlying pip if it needs to be called by pip-compile.
- This forces explicit passing of the complete, authenticated url string to pip, through pip-compile, via pip-compile's "--pip-args" argument.


doyleish:

PR #3232, opened 2024-02-15
fix(frontend): wrap values with special characters in double quotes when downloading a CSV

gft-kysonnd:


frontend/src/lib/components/table/AutoDataTable.svelte
```diff
@@ -161,7 +161,9 @@
 										return true
 									}
 								})
-								.map(({ rowData }) => Object.values(rowData).join(','))
+								.map(({ rowData }) => Object.values(rowData)
+									.map((row) => /[\",\n]/.test(row) ? "\"" + row.replace(/"/g, "\"\"") + "\"" : row)
```
rubenfiszel: replace row with field as the variable name

rubenfiszel (approved):

PR #3244, opened 2024-02-18
fix: bigquery timeout

HugoCasa:


backend/windmill-worker/src/bigquery_executor.rs
```diff
@@ -141,7 +142,7 @@ pub async fn do_bigquery(
             "query": query,
             "useLegacySql": false,
             "maxResults": 10000,
-            "timeoutMs": 10000, // default
+            "timeoutMs": MAX_TIMEOUT_DURATION.as_millis() as i64,
```
rubenfiszel: Take         JOB_DEFAULT_TIMEOUT instead

Issue #3277, opened 2024-02-23
bug:Date picker component when specifying date-time format goes away upon save/deploy


lfanew:
### Describe the bug

I'm trying to add typing information since datetime is not inferred for PowerShell via the Web UI. I set it to string, date-time, but upon saving or deployment it retains the regex validation but not the input component. I tried the same with a Python function main function that takes a str and the same thing happens there as well.

### To reproduce

1. Create script that takes string argument
2. Go to Settings -> Generated UI
3. Set the Format for the string to date-time
4. Save or deploy the script
5. Go to run the script and there won't be a date-time picker

### Expected behavior

The date-time type for this argument in the schema would be preserved upon saving

### Screenshots

![image](https://github.com/windmill-labs/windmill/assets/58160325/07a2af46-7dad-4edb-8d84-cfb0a805731f)

### Browser information

Vivaldi	6.5.3206.63 (Stable channel) stable (64-bit)

### Application version

v1.275.6-3-g5a57eb6fa

lfanew:
I can confirm if I modify the schema locally via the script YAML file and add `format: "date-time"` under the schema section and then sync push via CLI then it works (until I use the Web UI and save/deploy again).

lfanew:
Well I've fixed my particular issue by removing these two lines, but it introduces another issue.

https://github.com/windmill-labs/windmill/blob/36ef54189f8b837ccb6e5b5f4b3834bb8ed6f976/frontend/src/lib/infer.ts#L204-L205

For PowerShell or another language which currently does not have date-time inferred, this works great and is now not broken. However, say you're using Python and take a datetime argument and then switch that same argument back to a string, it will retain the date-time format which likely seems unintuitive unless you are familiar with how the schema works.

I'll submit a PR in case this behavior is preferred for now until a more comprehensive solution is in place, otherwise I'll probably look into supporting more datetype times in the parser to work around the issue

PR #3281, opened 2024-02-24
Fixes date-time always being stripped from str arg

lfanew:
This change prevents the date-time format in schema from being stripped every time when the argument
is a string. This allows languages without datetime arg inference to support datetime input

See issue #3277 for more information and caveat


frontend/src/lib/infer.ts
```diff
@@ -201,8 +201,6 @@ function argSigToJsonSchemaType(
 				delete oldS[prop]
 			}
 		}
-	} else if (oldS.format == 'date-time' && newS.format != 'date-time') {
-		delete oldS.format
```
rubenfiszel, line 205: Instead of doing this, a condition should be added for languages that do not support native date-time. 
That would require quite a big refactor unfortunately

lfanew:
Looking into adding more supported types for PowerShell parser for the time being.

PR #3293, opened 2024-02-26
How to upgrade message on cloud

hcourdent:


frontend/src/lib/components/Uptodate.svelte
```diff
@@ -19,11 +24,5 @@
...
 
 {#if uptodate}
-	<span class="text-blue-400"
-		>{uptodate} &nbsp;<Tooltip>
-			How to update?<br />
-			- docker: `docker compose up -d`<br />-
-			<a href="https://github.com/windmill-labs/windmill-helm-charts#install">helm</a>
-		</Tooltip>
-	</span>
+	<span class="text-blue-400">{uptodate} &nbsp;<Tooltip>{@html tooltipContent}</Tooltip></span>
```
rubenfiszel: just inline the tooltipContent string here, no need for this

hcourdent: @rubenfiszel fixed in latest commit

PR #3296, opened 2024-02-26
feat: move stats to closed source

HugoCasa:


backend/src/main.rs
```diff
@@ -110,12 +110,14 @@ async fn windmill_main() -> anyhow::Result<()> {
 
     match cli_arg.as_str() {
         "cache" => {
-            #[cfg(feature = "embeddings")] {
+            #[cfg(feature = "embeddings")]
```
rubenfiszel: woops, it's embedding, not embeddings

PR #3318, opened 2024-02-29
feat: Use ACCEPT_INVALID_CERTS for SMTP

AudriusButkevicius:
The library currently loads fixed set of certificates from a rust package, does not use OS certificate store. Providing a custom CA would need to be done via a custom tls connector.

Would appreciate a patch release with this, so I can pick up the change via official channels 🙇


rubenfiszel:
Thanks for this

PR #3319, opened 2024-03-01
fix: Fix build

AudriusButkevicius:
Apologies, previous PR was rubbish as I couldn't get it to build due to not understanding how to deal with sqlx.

Once I saw the red builds, spent some time understanding how sqlx actually works, that you need a database for builds.


rubenfiszel:
Not necessary anymore

PR #3442, opened 2024-03-19
[FIX] Generated python local code tester had wrong if statement syntax

c0nfleis:
This is a great feature to have to be able to copy and paste the real values. The if statement was not actually triggering locally though, so I fixed it.


c0nfleis:

PR #3469, opened 2024-03-26
Windmill AI handle multiple lines

hcourdent:


frontend/src/lib/components/apps/svelte-select/lib/tailwind.css
```diff
@@ -127,4 +127,14 @@
 
 .list-item {
     @apply list-none;
-}
\ No newline at end of file
+}
+
+.textarea-resize-none {
```
hugocasa: https://tailwindcss.com/docs/resize
resize-none

frontend/src/lib/components/apps/svelte-select/lib/tailwind.css
```diff
@@ -127,4 +127,14 @@
...
+    resize: none; /* Prevent manual resizing */
+  }
+  
+  .fixed-height-btn {
+    height: 38px; /* Set a fixed height for the button */
```
hugocasa: https://tailwindcss.com/docs/height 
h-[38px]

HugoCasa (commented):
you should be able to use tailwind classes instead

frontend/src/lib/components/apps/svelte-select/lib/tailwind.css
hcourdent: Thanks! Fixed for both

HugoCasa (approved):

PR #3525, opened 2024-04-08
fix(frontend): Improve theme editor

fatonramadani:


ameliadd:
I've been testing a few details we could improve:

1. The buttons on top are clumped together and the the default size of the menu cuts them off and gives a strange feel to it.

I tested maybe switching them to icons on top (that would have a tooltip to explain their purpose), but it seemed off too and the icons weren't intuitive enough.
Capture d’écran 2024-04-08 à 19 21 37
So I believe the simplest solution would be to add both buttons, one on top of the other rather than next to each other. This way the button size could be more responsive to the the changing of drawer size and it could avoid confusion or having too many different colours next to each other.
Capture d’écran 2024-04-08 à 19 21 43

______

2.  The borders around around the styling menu.
I find that they don't serve a big purpose and only make things feel things feel tighter in the menu. The line at the bottom to divide them (like the way it used to be) works better.

Capture d’écran 2024-04-08 à 18 06 36
Capture d’écran 2024-04-08 à 18 19 50 Here's a side by side comparison with and without the borders


_________

3. Another detail that caught my eye was the variety of font sizes when opening the styling menu.
I believe it's best to keep only two different sizes, text-xs and text-sm.
Capture d’écran 2024-04-08 à 18 14 55 Capture d’écran 2024-04-08 à 21 27 00

So, for titles (Such at the component type i.e. Button, "Plain CSS", "Styling Menu" and "Tailwind classes" would be in text-sm)

PR #3551, opened 2024-04-14
Powershell client

lfanew:
## Todo

Some of these don't have to be done right away and can be saved as future ideas.

- [x] Figure out how this will be published (PSGallery?)
- [ ] Have module installed locally inside of Windmill image (avoid the need to pull / version mismatches)
- [ ] Implement S3 support (I do not have any S3 storage to test with, so I have not implemented this atm)
- [x] Update PowerShell script template (import by default with comments demonstrating usage)
- [ ] Update PowerShell script actions (like add variable / resource)

## Summary

Here is my draft of a Windmill Powershell (pwsh) client.

The following functionality has been implemented and manually tested.

- Connect-Windmill
- Disconnect-Windmill
- New-WindmillToken
- Get-WindmillIdToken
- Get-WindmillWorkspace
- Get-WindmillVersion
- Get-WindmillUser
- Get-WindmillVariable
- New-WindmillVariable
- Set-WindmillVariable
- Remove-WindmillVariable
- Get-WindmillResource
- New-WindmillResource
- Set-WindmillResource
- Remove-WindmillResource
- Invoke-WindmillScript
- Start-WindmillScript
- Start-WindmillFlow
- Stop-WindmillJob
- Wait-WindmillJob
- Stop-WindmillExecution
- Get-WindmillJob
- Get-WindmillResult


rubenfiszel:
@lfanew this is amazing. Thank you very much.

We should indeed figure out a way to publish this automatically so it can be installed as a normal dependency even outside of windmill. But otherwise, we do not need for it to be perfect from the get go.

lfanew:
Thanks for the kind comment! Then some of the todo can be interpreted as future things for the client.

As for publishing here is some documentation that might help speed up any googling.
https://learn.microsoft.com/en-us/powershell/gallery/how-to/publishing-packages/publishing-a-package

HugoCasa:
Published on gallery.
We can merge the PR and put the next steps in an issue.

Thank you for the great work!

PR #3565, opened 2024-04-16
feat(frontend): Deeply nested Modals and Drawers

fatonramadani:
Test app: 'f/tests/deeply-nested-modals-drawers'


frontend/src/lib/components/common/drawer/Drawer.svelte
```diff
@@ -90,22 +60,23 @@
...
-	<div class="overlay {$$props.positionClass ?? ''}" on:click={handleClickAway} />
-	<div class="panel {placement} {$$props.positionClass}" class:size>
-		{#if open || !timeout || alwaysOpen}
-			<slot {open} />
-		{/if}
-	</div>
-</aside>
+<Disposable let:handleClickAway let:zIndex bind:open bind:this={disposable}>
+	<aside
+		class="drawer {$$props.class ?? ''} {$$props.positionClass ?? ''}"
```
rubenfiszel: if you're not changing/removing drawer.open and drawer.class z-index, the behavior is hard to understand. Also prior it was starting from 1002 for drawers and not it's 1100. We need to check thorougly in all the cases where drawer can be stacked on top of each other and on app and app editor if the stacking is not altered because of this

PR #3632, opened 2024-04-29
fix(frontend): sync columnDefs + improve columnDefs management

fatonramadani:
Screenshot 2024-04-29 at 09 39 39
Screenshot 2024-04-29 at 09 40 03
Screenshot 2024-04-29 at 09 39 52

- Regroup the column defs sync/fix into a single component
- Added a modal (only shown in the editor) to fix the columnDefs if they are undefined or not an array
- Added a modal (only shown in the editor) to add suggested columnDefs (based on the result) if they are an empty array
- Data input cannot be static anymore


hcourdent:
You sometime write "Column definition" and sometimes ""Column definitions"

hcourdent:
I was not able to trigger the custom messages around column defs

Issue #3758, opened 2024-05-17
feature: Control the height of a form input


d10sfan:
Currently, there does not appear to be a way to control the textarea that is created from giving the json schema to the form widget. This would be great to be able to set a default height for that textarea, so that for larger values, it can start at a reasonable size.

PR #3759, opened 2024-05-17
Add support for form schema property setting textarea rows value

d10sfan:
This adds an ability to set the initial rows setting for the textarea, instead of it being hard-coded to 1. This allows you to then pass into the form widget the following schema:

```json
{
	"properties": {
		"first_name": {
			"type": "string",
			"description": "your name",
			"default": "default",
                        "rows": 10
		}
	},
	"required": []
}
```

The main use case here is that if you have a large value on the form, being able to set the rows allows to have it expanded by default, instead of the user only seeing 1 line of a very large string. Let me know if there's any better places to put the config.

Fixes https://github.com/windmill-labs/windmill/issues/3758

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit a8dc506c1c03c05942fc86830d514ceb4bbc2f06  | 
|--------|

### Summary:
Enhanced `textarea` in `LightweightArgInput.svelte` to support dynamic rows setting from schema, improving UX for larger text inputs.

**Key points**:
- Updated `LightweightArgInput.svelte` to dynamically set `textarea` rows from schema.
- Uses `extra?.['rows'] || 1` to determine rows count, defaulting to 1 if not specified.
- Enhances UX for larger text inputs by not limiting to a single row.

----


d10sfan:

d10sfan:
recheck

Issue #3757, opened 2024-05-17
feature: AgGrid Deselecting Row


d10sfan:
Can you use setSelectedIndex to un-selected a row on an aggrid?

I tried using the ag grid API to run setSelected(false) on a rowNode, but that did not update the state of the ag grid, so selectedRow was still set to something, even though visually it was no longer selected.

Also I tried setSelectedIndex('your-table-id', -1); like the AI suggested, but that didn't do anything, so wondering if there may be another way.

The use case here is that I want to open a modal whenever someone clicks on a row, which I can detect the selectedRow difference, but then when the modal closes I want to reset the selection.

PR #3760, opened 2024-05-17
AgGrid - Allow to de-select all rows with setSelectedIndex

d10sfan:
Adds the ability to deselect the rows using setSelectedIndex, using -1, from a suggestion on the AI that did not appear to be implemented

Fixes https://github.com/windmill-labs/windmill/issues/3757


frontend/src/lib/components/apps/components/display/table/AppAggridTable.svelte
```diff
@@ -303,6 +303,11 @@
 							$componentControl[id] = {
 								agGrid: { api: e.api, columnApi: e.columnApi },
 								setSelectedIndex: (index) => {
+									if(index === -1) {
+										e.api.deselectAll();
+										outputs?.selectedRow?.set({})
+										outputs?.selectedRowIndex.set(-1)
+									}
```
rubenfiszel, line 313: the rest should be in an else

frontend/src/lib/components/apps/components/display/table/AppAggridTable.svelte
```diff
@@ -303,6 +303,11 @@
 							$componentControl[id] = {
 								agGrid: { api: e.api, columnApi: e.columnApi },
 								setSelectedIndex: (index) => {
+									if(index === -1) {
```
rubenfiszel: should accept null or index rather than -1, unless that's an AgGrid convention

d10sfan: I've updated this to be a null instead of -1

frontend/src/lib/components/apps/components/display/table/AppAggridTable.svelte
d10sfan: I've updated the selecting of an index to be in an else

PR #3773, opened 2024-05-20
fix(frontend): fix AgGrid pagination refresh

fatonramadani:


frontend/src/lib/components/apps/components/inputs/AppFileInput.svelte
```diff
@@ -48,6 +49,16 @@
 	}
 
 	let files: File[] | undefined = undefined
+
+	function preFillFiles() {
+		const data = outputs?.result?.peak()
+
+		if (data && Array.isArray(data) && data.length > 0) {
+			files = data.map((file: { name: any }) => new File([], file?.name))
```
rubenfiszel, line 57: I'm confused by this, wouldn't the file be empty with that?

Issue #3858, opened 2024-06-02
feature: Support for Non-Root iFrame Embedding


bigcat88:
**Description:**

We are integrating Windmill into Nextcloud, and for this integration to work seamlessly, Windmill needs to support operations within an iFrame that does not use the root address.

**Example iFrame URL:**
```
http://nextcloud.local/apps/app_api/proxy/windmill_app
```

We have configured the correct `kit.paths.base` in `svelte.config.js`, and observed that JS/CSS files are successfully loaded. Additionally, we have set the correct URL for the generated "backend-client" in Windmill, ensuring that all requests to `/api` and some requests to `/user` endpoints are correctly routed.

However, in the TypeScript files of Windmill, all `goto` operations are relative to the root of the site, leading to incorrect redirects.

**Example Windmill Code:**
```typescript
export async function logout(): Promise<void> {
    await clearUser()
    goto(`/user/login`)
    sendUserToast('you have been logged out')
}
```

**Documentation for `goto`:**
```
Params:
url – Where to navigate to. Note that if you've set config.kit.paths.base and the URL is root-relative, you need to prepend the base path if you want to navigate within the app.
opts – Options related to the navigation
```

**Request:**

We kindly request that the Windmill frontend code be modified to respect the `config.kit.paths.base` by prepending it to the path in `goto` operations. This change is crucial for our integration, as manually editing the entire Windmill frontend is not feasible for us.

We appreciate your attention to this matter and want to express our gratitude for maintaining such an excellent open-source project.

rubenfiszel:
The issue with svelte.config.js approach is that it  require to maintain your own fork of windmill which will a heavy burden. Ideally, we can find a solution that work with the stock image of windmill and can be passed as env variable without requiring a re-build of the frontend.

rubenfiszel:
Do you have any possibility to have windmill be hosted as a subdomain, for instance:
http://windmill.nextcloud.local ?

rubenfiszel:
The other possibility that i see is that we have a new image that support windmill not as a SPA with fallback mode, but as a server rendered website (running the frontend on node.js) that you can pass the base as an env variable to the image.

bigcat88:
From what I know, we still want to install Windmill as a container, and it’s not difficult for us to rebuild only the frontend without rebuilding the entire binary when we creating that container.
And to serve frontend from that folder(let's call it static_frontend)

[...]

PR #3867, opened 2024-06-04
fix: Allow depolying frontend in a sub directory

marcelklehr:
Fixes https://github.com/windmill-labs/windmill/issues/3858

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 83611b3991eb18ee47cba63f567dab0ec54dc7d4  | 
|--------|--------|

### Summary:
Centralized navigation handling and updated path management for deployments in subdirectories.

**Key points**:
- Centralized navigation handling by updating the import source of `goto` across multiple components.
- Adjusted paths in various components to ensure correct base path prepending.
- Updated system-wide navigation handling for consistency and controlled behavior.

----


marcelklehr:

marcelklehr:
recheck

rubenfiszel:
I would be surprised if there wasn't a few `<a href=` that are missing but overall looks good to me and safe

marcelklehr:
Ah, right, I'll get on that :)

rubenfiszel:
@marcelklehr you will want to take a look at Button.svelte which has an a href through a <svelte:element

I would advise to play with the app with a custom base for an hour or so to check that everything work (and maybe give us feedback on the product :)) as on top of my head there are likely many edge-cases I forgot about.

marcelklehr:
> and maybe give us feedback on the product :)

Oh, we've played with it for quite a while now. Great software! :)

hbrooks:
hey @ellipsis-dev, can you check this PR to make sure there aren't any missing `<a href=` statements in the diff? 

CC: @rubenfiszel

rubenfiszel:
@marcelklehr @hbrooks I think there is a problem with the current approach. Windmill is also being embedded as an SDK in contexts that are not sveltekit. We've been careful to not rely on any sveltekit variables in the components that are exported. I have to verify that this is still true after your PR. One solution, would be to change base to be an actual environment variables, in which case we can use a svelte agnostic env variable. Apologies for realizing this only now.  To be clear, the issue is the import `import { base } from '$app/paths'`

One way to make refactoring easier later is to do the same than with goto and have an indirection layer with `$lib/paths`

Otherwise looks good.

marcelklehr:
Good idea @rubenfiszel! I created a new export in $lib/navigation for this purpose and pointed all files to this

rubenfiszel:
@marcelklehr we have to test the PR but everything looks good

rubenfiszel:
@marcelklehr I wanted to review your PR but there are a few conflicts. I can resolve them myself if you want, up to you.

rubenfiszel:
In any case, reviewing and testing the PR is the next thing on my bucket list, really sorry for the delay

rubenfiszel:
@marcelklehr ping on this, would you like us to resolve the conflicts on the PR or do you prefer to do it?

marcelklehr:
As you wish. If you need me to, I can do it :)

rubenfiszel:
@marcelklehr if you can do a first pass, we would be grateful!

marcelklehr:
Resolved all conflicts.

frontend/src/lib/navigation.ts
```diff
@@ -1,4 +1,12 @@
-import { goto } from '$app/navigation'
+import { goto as svelteGoto } from '$app/navigation'
+import { base } from '$lib/navigation'
+
+export { base } from '$app/paths'
+
+export function goto(path, options = {}) {
+	const fullPath = path.startsWith(base) ? path : `${base}${path}`;
```
rubenfiszel: optimization since this is now used everywhere

have a global const
export const CUSTOM_BASE = base && base != ""

then modify this function to be:
if (!CUSTOM_BASE) { return svelteGoto(path, options) }

It's nit but easy to optimize

rubenfiszel:
@marcelklehr I've reviewed the PR, it all looks good but there is one issue that you couldn't have known which is that we also export many components for whitelabelling purposes in a context outside of sveltekit and with just svelte. To that end, we need to have no dependencies on sveltekit specific functions for components themseves (svelte route pages are fine)

There is now an import in many places to $lib/navigations which point to $app/navigations where there wasn't, so the things to do is:
1. split base out of $lib/navigations so that it can imported in a file that has no dependency on $lib/navigation
2. use a normal vite env variable for base and depend on that: https://learn.svelte.dev/tutorial/env-static-private

I can do the modifications if preferred

marcelklehr:
Ah, I didn't know about this. Would be great if you could do the fine-tuning :pray:

PR #3955, opened 2024-06-22
Refactored nested match expressions for TS parser.

dommyrock:
Refactored nested match expressions into inline pattern matching with the help of destructuring with early returns + guard clauses 
for improved readability and reduced nesting.

Don't have the whole project setup so i checked that lib still builds as normal.
```bash
cd ./backend/parsers/windmill-parser-ts
cago b
```
Hope it helps keep things more organized for the future work.
Cool project btw :)

### DIFF
--- a/backend/parsers/windmill-parser-ts/src/lib.rs
+++ b/backend/parsers/windmill-parser-ts/src/lib.rs
@@ -403,71 +403,9 @@ fn tstype_to_typ(ts_type: &TsType) -> (Typ, bool) {
             } else {
                 if types.len() > 1 {
                     let one_of_values: Vec<OneOfVariant> = types
-                        .into_iter()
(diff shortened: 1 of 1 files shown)


dommyrock:

HugoCasa:
Thanks for your contribution!
I'll test it properly this week. 
Is there a specific reason why you replaced map_while with filter_map?
I would also suggest adding an indication in the function names that they are linked to the parsing of oneOfs.

HugoCasa (changes requested):
see my previous comment

dommyrock:
> see my previous comment

You're correct , should be map_while() as it was. 
I Was playing around with few different ideas so i missed this :/ 
Also fixed naming as you suggested.

Thanks for the review!

HugoCasa (approved):

PR #3957, opened 2024-06-23
cleanup graphql parser code

dommyrock:
minor cleanup
- updated fn parse_graphql_file() 
- removed direct cloning. 

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 1622db3507e1c44d0f23b27cbb71c8cf229cd1fa  | 
|--------|--------|

### Summary:
Refactored `parse_graphql_file` function to remove unnecessary cloning and simplify logic in `backend/parsers/windmill-parser-graphql/src/lib.rs`.

**Key points**:
- Updated `parse_graphql_file` function in `backend/parsers/windmill-parser-graphql/src/lib.rs`
- Removed unnecessary cloning of strings
- Simplified the logic for parsing GraphQL types and defaults
- Adjusted `parse_graphql_sig` to handle the updated `parse_graphql_file` return type

----

### DIFF
--- a/backend/parsers/windmill-parser-graphql/src/lib.rs
+++ b/backend/parsers/windmill-parser-graphql/src/lib.rs
@@ -7,55 +7,58 @@ use serde_json::json;
 use windmill_parser::{Arg, MainArgSignature, Typ};
 
 pub fn parse_graphql_sig(code: &str) -> anyhow::Result<MainArgSignature> {
-    parse_graphql_file(&code)
-        .map(|args| MainArgSignature {
-            star_args: false,
-            star_kwargs: false,
-            args,
-            no_main_func: None,
+    parse_graphql_file(&code)?
+        .map(|args| {
+            Ok(MainArgSignature { star_args: false, star_kwargs: false, args, no_main_func: None })
         })
-        .ok_or_else(|| anyhow!("Error parsing sql".to_string()))
+        .ok_or_else(|| anyhow!("Error parsing gql".to_string()))?
 }
 
 lazy_static::lazy_static! {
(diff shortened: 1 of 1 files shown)


backend/parsers/windmill-parser-graphql/src/lib.rs
```diff
@@ -7,53 +7,55 @@ use serde_json::json;
...
-    }
-
-    Ok(Some(args))
+fn parse_graphql_file(code: &str) -> Option<Vec<Arg>> {
+    RE_ARG_GRAPHQL
+        .captures_iter(code)
+        .filter_map(|cap| Some(parse_arg_cap(cap)))
+        .collect()
+}
+fn parse_arg_cap(cap: regex::Captures) -> Option<Arg> {
```
hugocasa: Why does this function return an option?

HugoCasa (changes requested):

backend/parsers/windmill-parser-graphql/src/lib.rs
dommyrock: Seems that I misunderstood , initial implementation so i thought wrapping option into Result was extra at that point .
- Reverted back to initial signature `anyhow::Result<Option<Arg>>`
- also I figured simple macro capturing cleans up a lot of code if we pre extract it all into 4x tuple
- I also noticed that Err should probable state "Error parsing gql" instead of "Error parsing sql" since we parse GQL schema.

I understand if you don't like macro solution (Feel free to reject pr if you like, It seemed like cleanest way I could think of )

Issue #4191, opened 2024-08-05
feature: go 1.22.5


freimer:
Bump go version 1.22.5.  I will create a PR.

PR #4192, opened 2024-08-05
bump go version

freimer:
See #4191 

Bumps Go version to 1.22.5

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 104e65dc761435e308bf87871415994c5c3ec659  | 
|--------|--------|

### Summary:
Update Go version in `Dockerfile` from 1.21.6 to 1.22.5 for multiple architectures.

**Key points**:
- Update Go version in `Dockerfile` from 1.21.6 to 1.22.5.
- Affects Go installation for `amd64`, `arm64`, and `armhf` architectures.
- Changes made in `Dockerfile` lines 9, 11, and 13.

----

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,13 +127,13 @@ RUN set -eux; \
     arch="$(dpkg --print-architecture)"; arch="${arch##*-}"; \
     case "$arch" in \
     'amd64') \
-    targz='go1.21.6.linux-amd64.tar.gz'; \
+    targz='go1.22.5.linux-amd64.tar.gz'; \
     ;; \
     'arm64') \
-    targz='go1.21.6.linux-arm64.tar.gz'; \
+    targz='go1.22.5.linux-arm64.tar.gz'; \
     ;; \
     'armhf') \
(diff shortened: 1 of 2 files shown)


freimer:

rubenfiszel:
there is also the lsp dockerfile to change, thanks for the PR :)

freimer:
For a feature request, do you think it might be possible to make the Go version an option in the UI / script, on a per-script basis?  The Go version is only used to compile the program, which is then cached, correct?  So maybe have the images with a specific, somewhat recent version of Go, and a default version specified in the server config somewhere.  Then, with a comment in the script or UI option let the version be specified.  If it is not the image version, download and install the version specified when building the program.  Maybe keep these go versions cached also, to reduce constantly downloading them when interactively editing / testing programs?

For now, updating the version is fine.  But consider whether this would be a useful future feature to be able to specify version without updating the image.

rubenfiszel:
@freimer  You can achieve some of this by assigning the script to a worker with the desired go version but having it be an annotation is a constraint that would be too complex to maintain.

PR #4304, opened 2024-08-29
fix: Take interpreter paths from environment (bash executor)

Bert-Proesmans:
Updated the bash executor to use both bash and powershell from the environment variables (if set).
I took the simplest approach to get the desired effect, ignoring the example that POWERSHELL_PATH is defined in worker.rs

Ran cargo fmt on the file and validated syntax with cargo check, on latest rust stable (v1.80.1)

<details>

<summary>OPS</summary>

```shell
tee build-shell.nix <<EOF
> { pkgs ? import <nixpkgs> { } }:
pkgs.mkShell {

  buildInputs = with pkgs; [ openssl rustfmt lld stdenv.cc.cc.lib ];

  nativeBuildInputs = with pkgs; [
    rustup
    pkg-config
    makeWrapper
    swagger-cli
    cmake # for libz-ng-sys crate
  ];

  SQLX_OFFLINE = "true";
}
EOF

nix-shell build-shell.nix

rustup default stable

cd backend/windmill-worker && cargo check
```

</details>

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit a969678a2fe68945171928cd516e36e61c72c995  | 
|--------|--------|

### Summary:
Updated bash executor to use environment variables for bash and powershell paths in `backend/windmill-worker/src/bash_executor.rs`.

**Key points**:
- Updated `backend/windmill-worker/src/bash_executor.rs` to use environment variables for interpreter paths.
- Defined `BIN_BASH` using `lazy_static` to get `BASH_PATH` from environment or default to `/bin/bash`.
- Replaced hardcoded `/bin/bash` with `BIN_BASH` in `handle_bash_job` and `handle_powershell_job` functions.
- Used `POWERSHELL_PATH` for powershell command execution in `handle_powershell_job`.

----

### DIFF
--- a/backend/windmill-worker/src/bash_executor.rs
+++ b/backend/windmill-worker/src/bash_executor.rs
@@ -11,7 +11,9 @@ use windmill_common::{
 };
 use windmill_queue::{append_logs, CanceledBy};
 
-const BIN_BASH: &str = "/bin/bash";
+lazy_static::lazy_static! {
+    static ref BIN_BASH: String = std::env::var("BASH_PATH").unwrap_or_else(|_| "/bin/bash".to_string());
+}
 const NSJAIL_CONFIG_RUN_BASH_CONTENT: &str = include_str!("../nsjail/run.bash.config.proto");
 const NSJAIL_CONFIG_RUN_POWERSHELL_CONTENT: &str =
     include_str!("../nsjail/run.powershell.config.proto");
@@ -55,7 +57,7 @@ pub async fn handle_bash_job(
     write_file(
         job_dir,
         "wrapper.sh",
-        &format!("set -o pipefail\nset -e\nmkfifo bp\ncat bp | tail -1 > ./result2.out &\n /bin/bash ./main.sh \"$@\" 2>&1 | tee bp\nwait $!"),
+        &format!("set -o pipefail\nset -e\nmkfifo bp\ncat bp | tail -1 > ./result2.out &\n {bash} ./main.sh \"$@\" 2>&1 | tee bp\nwait $!", bash = BIN_BASH.as_str()),
     )?;
 
     let token = client.get_token().await;
@@ -96,7 +98,7 @@ pub async fn handle_bash_job(
             "--config",
             "run.config.proto",
             "--",
-            "/bin/bash",
+            BIN_BASH.as_str(),
             "wrapper.sh",
         ];
         cmd_args.extend(args);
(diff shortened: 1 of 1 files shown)


backend/windmill-worker/src/bash_executor.rs
```diff
@@ -255,7 +257,7 @@ pub async fn handle_powershell_job(
     if !install_string.is_empty() {
         logs1.push_str("\n\nInstalling modules...");
         append_logs(&job.id, &job.workspace_id, logs1, db).await;
-        let child = Command::new("pwsh")
```
rubenfiszel, line 258: This is a small functional change. I believe it should work but not necessarily in all scenarios which I can't test right now. Have you been able to test it?

Bert-Proesmans: I've been running the changes in this patch for about ~3 weeks with script actions/flows in bash, and powershell, and python. No issues, including using some (PSGallery) external modules. This is my own community build for the workers specifically.

You're right that this is a functional change, pwsh "always" refers to "system install of powershell" and is used here specifically for installing pre-requisite modules. The other powershell invocation would be "custom build (or version) of powershell" specifically to run scripts.
I'm not a powershell expert. Having two different versions between installing dependencies (pre-script run) and for running scripts (during script run) could imply a broken script execution with a confusing message of missing dependencies. It's a distinction that can certainly be useful, but requires out-of-band setup that must overlap with configuration of the "system install of powershell", in most cases a footgun otherwise a restriction when the system repository index has (very) old powershell versions.


I should have mentioned this in the original post;
the direct reason for this PR is because I want to upstream my nix build recipe and -configuration changes for the community edition platform. The first version of the build recipe uses the \<INTERPRETER>_PATH environment variables and I just continued the same structure for bash and powershell. This was never a hard requirement though, the recipe could also updates the PATH variable for all the interpreter executables. Using the \<INTERPRETER>_PATH variables is more explicit, I can understand the attractiveness to that.

Bert-Proesmans:
I don't require this PR being merged as is, take/implement whatever you want from it. There are multiple other ways to ensure the binary (names) align on the system.

the direct reason for this PR is because I want to upstream my nix build recipe and -configuration changes for the community edition platform. The first version of the build recipe uses the \<INTERPRETER>_PATH environment variables and I just continued the same structure for bash and powershell. This was never a hard requirement though, the recipe could also updates the PATH variable for all the interpreter executables. Using the \<INTERPRETER>_PATH variables is more explicit, I can understand the attractiveness to that.

rubenfiszel:
Thanks for the contribution, looks like you've tested it to work correctly. My main concern with powershell was breaking it for our existing customers.

PR #4333, opened 2024-09-03
chore: gives the possibility not to install git in the docker image

superjcvd:
| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 3f6cc1ed694ff1721bbfec7732ad7f9c50791a39  | 
|--------|--------|

### Summary:
Update `Dockerfile` to use latest Debian packages, add Go versioning, conditional `git` installation, and package upgrade commands.

**Key points**:
- Update `Dockerfile` to use latest Debian packages.
- Add `ARG GO_VERSION=1.22.5` for Go versioning.
- Modify Go installation to use `GO_VERSION`.
- Add `ARG WITH_GIT=true` for conditional `git` installation.
- Add `RUN` command for `apt-get upgrade` and `apt-get autoremove`.

----

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,7 +102,7 @@ RUN apt-get update \
 
 RUN if [ "$WITH_GIT" = "true" ]; then \
     apt-get update  -y \
-    && apt-get install -y git
+    && apt-get install -y git \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*; \
     else echo 'Building the image without git'; fi;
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -151,7 +155,12 @@ ENV GO_PATH=/usr/local/go/bin/go
 
 RUN curl -sL https://deb.nodesource.com/setup_20.x | bash - 
 RUN apt-get -y update && apt-get install -y curl nodejs awscli && apt-get clean \
-  && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update \
```
rubenfiszel: The versions should be pinned when installing packages to the exact versions instead of upgrading otherwise it will mess with caching

superjcvd: Yes, 

That's all the issue with docker images and security :(

Let's remove it for now. 
Just adding the possibility to remove git on demand (if not using the git sync feature) is enough.

PR #4396, opened 2024-09-16
fix(frontend): new resource type name must be snake case

alpetric:
Adding a new resource type when not using snake case (e.g using camel case) lead to windmill code editor being unable to find/match the type automatically (not shown as drop down).

- this PR checks/validates the name on the client side only
- offers an option to "auto fix" the name to snake case
- disables button if name invalid

https://github.com/user-attachments/assets/1011a93b-1827-4fd2-899c-26719a5e6459

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 40d37e4a1e5e0b10d1114cc0902f16efbabde9d7  | 
|--------|--------|

fix(frontend): enforce snake case for new resource type names

### Summary:
Enforces snake case for new resource type names with validation and auto-fix in `+page.svelte`.

**Key points**:
- Enforces snake case for new resource type names in `+page.svelte`.
- Adds `validateResourceTypeName()` to check if names are snake case.
- Updates `isNewResourceTypeNameValid` based on validation.
- Adds `toSnakeCase()` to convert names to snake case.
- Provides "Fix..." button to auto-correct invalid names.
- Disables "Save" button if name is invalid.
- Shows error message and "Fix..." button if name is invalid.

----

### DIFF
--- a/frontend/src/routes/(root)/(logged)/resources/+page.svelte
+++ b/frontend/src/routes/(root)/(logged)/resources/+page.svelte
@@ -345,6 +345,16 @@
 		const snakeCaseRegex = /^[a-z]+(_[a-z]+)*$/
 		newResourceType.isValid = snakeCaseRegex.test(newResourceType.name)
 	}
+
+	function toSnakeCase() {
+		newResourceType.name = newResourceType.name
+			.replace(/([a-z0-9])([A-Z])/g, '$1_$2')
+			.replace(/[\W]+/g, '_')
+			.replace(/_+/g, '_')
+			.replace(/^_+|_+$/g, '')
+			.toLowerCase()
+		newResourceTypeNameValid()
+	}
 </script>
 
 <ConfirmationModal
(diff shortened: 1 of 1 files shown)


frontend/src/routes/(root)/(logged)/resources/+page.svelte
```diff
@@ -75,7 +75,8 @@
 	let newResourceType = {
 		name: '',
 		schema: emptySchema(),
-		description: ''
+		description: '',
+		isValid: false
```
rubenfiszel: this should be held separately since it's not a property of the resource type that we want to send to the api

frontend/src/routes/(root)/(logged)/resources/+page.svelte
```diff
@@ -338,6 +340,21 @@
 		deployUiSettings = settings.deploy_ui ?? ALL_DEPLOYABLE
 	}
 	getDeployUiSettings()
+
+	function newResourceTypeNameValid() {
+		const snakeCaseRegex = /^[a-z]+(_[a-z]+)*$/
```
rubenfiszel: numbers are valid too in snake case

frontend/src/routes/(root)/(logged)/resources/+page.svelte
```diff
@@ -338,6 +342,21 @@
 		deployUiSettings = settings.deploy_ui ?? ALL_DEPLOYABLE
 	}
 	getDeployUiSettings()
+
+	function validateResourceTypeName() {
+		const snakeCaseRegex = /^[a-z0-9]+(_[a-z0-9]+)*$/
```
rubenfiszel, line 347: I think if you ask Claude or chatgpt you can have the perfect regex

PR #4420, opened 2024-09-23
floor progress in typescript client

pyranota:
If user passes float to setProgress it will result BadRequest from server but without any explanation.

Sometimes it is non intuitive that endpoint accepts only integers, so to reduce confusion we convert progress to int.

----

> [!IMPORTANT]
> Floors float input to integer in `setProgress()` in `client.ts` to prevent server errors.
> 
>   - **Behavior**:
>     - Floors float input to integer in `setProgress()` in `client.ts` to prevent BadRequest errors from server.
>   - **Documentation**:
>     - Updates docstring for `setProgress()` to include `jobId` parameter description.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for e7230fb5bd94c2ad059e36b6e4dc5b2ec5e3d6a7. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/typescript-client/client.ts
+++ b/typescript-client/client.ts
@@ -391,6 +391,7 @@ export async function setState(state: any): Promise<void> {
  * Set the progress
  * Progress cannot go back and limited to 0% to 99% range
  * @param percent Progress to set in %
+ * @param jobId? Job to set progress for
  */
 export async function setProgress(percent: number, jobId?: any): Promise<void> {
   const workspace = getWorkspace();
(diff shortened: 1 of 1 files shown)


pyranota:
Note: If user passes float to python client, request will fail as well, but this time returned msg will say that server accepts only i32

Issue #4483, opened 2024-10-04
bug: custom id for a component in app is not saved


vzhemchuzhnikov:
### Describe the bug

When pressing the pencil button, the popup with text field appears, but pressing right arrow button does nothing, the new id is not saved.

### To reproduce

1. Create a new app
2. Add a component
3. Try to change an id
4. Type new id and press right arrow button
5. New id is not applied

### Expected behavior

New id is applied

### Browser information

Edge Version 128.0.2739.42 (Official build) (64-bit), Firefox 130.0.1 (64-bit)

### Application version

CE v1.405.2

rubenfiszel:
It's already fixed on latest and latest released

PR #4501, opened 2024-10-07
custom id for a component in app is not saved

vs4vijay:
Related to #4483


rubenfiszel:
Thanks for the PR but the issue is already fixed on latest release

PR #4529, opened 2024-10-14
feat(sso): adding the ability to define a custom display name for sso

alpetric:


backend/windmill-api/openapi-deref.json
```diff
@@ -3932,8 +3932,19 @@
...
-                      }
+                        "type": "object",
+                        "properties": {
+                          "type": {
+                            "type": "string"
+                          },
+                          "display_name": {
+                            "type": "string"
+                          }
+                        },
```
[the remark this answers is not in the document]
alpetric: backend always sets display_name to empty string if not present

backend/windmill-api/openapi.yaml
```diff
@@ -2789,7 +2789,15 @@ paths:
...
-                      type: string
+                      type: object
+                      properties:
+                        type:
+                          type: string
+                        display_name:
+                          type: string
+                      required:
+                        - type
+                        - display_name
```
rubenfiszel: I think display_name should be non required as discussed in the other PR

frontend/src/lib/components/AutheliaSetting.svelte
```diff
@@ -52,6 +52,10 @@
 				>
 				<input type="text" placeholder="yourorg" bind:value={org} />
 			</label>
+			<label class="block pb-2">
+				<span class="text-primary font-semibold text-sm">Display Name</span>
```
rubenfiszel: here and in every similar field, we should say "Custom Name" instead of Display name

rubenfiszel: to make it clear that it's non required

alpetric: should I add a `(optional)` in the placeholder as well or is it self-explanatory?

rubenfiszel: It's self-explanatory I think

PR #4622, opened 2024-11-01
feat(oauth): add snowflake oauth support

alpetric:
> [!IMPORTANT]
> Add Snowflake OAuth support by updating backend token handling and frontend configuration components.
> 
>   - **Backend**:
>     - Add `snowflake_oauth` entry to `oauth_connect.json`.
>     - Modify `SnowflakeDatabase` in `snowflake_executor.rs` to make `public_key`, `private_key`, and `username` optional.
>     - Update `do_snowflake_inner()` and `do_snowflake()` in `snowflake_executor.rs` to handle OAuth tokens, including checking for token presence and generating new tokens if necessary.
>   - **Frontend**:
>     - Update `AppConnectInner.svelte` to handle `snowflake_oauth` resource type, including setting `account_identifier` from `extra_params`.
>     - Update `InstanceSettings.svelte` to pre-populate and handle `snowflake_oauth` account identifier, and configure OAuth settings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 3d5e4d39759f93daee55caf8ebcb7994d2f41294. It will automatically update as commits are pushed.</sup>


frontend/src/lib/components/InstanceSettings.svelte
```diff
@@ -269,6 +277,25 @@
...
 		return true
 	}
+
+	let snowflakeAccountIdentifier = ''
+
+	$: snowflakeAccountIdentifier && setupSnowflakeUrls()
+
+	function setupSnowflakeUrls() {
+		// strip all whitespaces from account identifier
+		snowflakeAccountIdentifier = snowflakeAccountIdentifier.replace(/\s/g, '')
```
rubenfiszel, line 293: this is creating a loop but overall the whole reactivity doesn't make much sense, it seems snowflakeAccountIdentifiers can only be modified in one place, you could just call setupSnowflaeUrls from there directly

rubenfiszel: actually it can be set in 2 places, but I would encapsulate this in a separate Snowflake component, avoid the loop (the loop can be avoided by doing a check if there is a diff before doing the assignment)

PR #4658, opened 2024-11-07
Add Python support for newspaper4k

marcus-crane:
Hi there,

This PR adds support for importing the Python module [newspaper4k](https://github.com/AndyTheFactory/newspaper4k) which is loaded under the `newspaper` namespace.

I should note that newspaper4k is a fork of the popular [newspaper3k](https://github.com/codelucas/newspaper). At first I was going to add support for both but I realise this might end up in Windmill might pull in either 3k or 4k depending on list ordering or just a plain race condition.

As neat as magic loading in of libraries is, it seems a bit too magical for cases like this where you have to modify Windmill to support certain libraries. It seems like an aliasing redirect using the [comment format](https://www.windmill.dev/docs/advanced/dependencies_in_python#pinning-dependencies-and-requirements) could be neat ie; `newspaper4k as newspaper; import newspaper`

Anyway, thanks!

----

> [!IMPORTANT]
> Maps `newspaper` import to `newspaper4k` in `lib.rs` to avoid conflicts with `newspaper3k`.
> 
>   - **Behavior**:
>     - Maps `newspaper` import to `newspaper4k` in `PYTHON_IMPORTS_REPLACEMENT` in `lib.rs`.
>     - Ensures `newspaper4k` is used when `newspaper` is imported, avoiding conflicts with `newspaper3k`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 004fd1f0c6e9d4e013ff4ea9f2714f74cb41fe00. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/backend/parsers/windmill-parser-py-imports/src/lib.rs
+++ b/backend/parsers/windmill-parser-py-imports/src/lib.rs
@@ -62,6 +62,7 @@ static PYTHON_IMPORTS_REPLACEMENT: phf::Map<&'static str, &'static str> = phf_ma
     "lokalise" => "python-lokalise-api",
     "msgraph" => "msgraph-sdk",
     "pythonjsonlogger" => "python-json-logger",
+    "newspaper" => "newspaper4k",
 };
 
 fn replace_import(x: String) -> String {


marcus-crane:

marcus-crane:
For anyone reading this in the future, I didn't have any luck trying to do this within the script editor due to Windmill trying to resolve `import newspaper` to a Python 2 module which would break my script.

I had a look at using an explicit requirements.txt file to get around inferred dependencies but couldn't get that to work for whatever reason.

In the meantime, I ended up just hacking around the whole inferred dependencies issue like so:

```python
#extra_requirements:
#newspaper4k
import importlib

newspaper = importlib.import_module('newspaper')
```

This let me import a module with a different name without triggering Windmill's inferred dependency functionality.

marcus-crane:
> @marcus-crane 
> 
> 
> 
> you can also just do:
> 
> ```
> 
> #requirements:
> 
> #newspaper4k
> 
> import newspaper
> 
> ```
> 
> 
> 
> I'm a bit on the fence of merging this bc i'm not sure if newspaper4k is more popular than newspaper3k at the moment

That doesn't work for me because newspaper is also a real package so Windmill tried to import both newspaper and newspaper4k. newspaper is a Python 2 package so it causes pip to bail out unless I'm missing something here 🤔

marcus-crane:
Ah yes, that does work! I had misunderstood that `requirements` does override the inferred installation unlike `extra_requirements`.

As far as this PR, I don't personally need it anymore but given there's `newspaper3k` and `newspaper4k` both actively used, it does kind of a raise a shortcoming with the single overrides for inferred dependencies 😓 

Perhaps the docs for `requirements` could be updated to make it a bit clearer that it can be used for cases where the package name doesn't match the import name 🙂

PR #4689, opened 2024-11-12
feat(backend): move some static fields out of job tables

uael:
This change aims to reduce the row size of the `queue` and `completed_job` tables by migrating "persistent" columns (i.e., columns not directly associated with runtime status) to a new shared `job` table. Both queued and completed jobs will reference this table, leading to significantly faster updates on the `queue` table.

## Backward compatibility

- **Database integrity**: Original columns from `queue` and `completed_job` are left in place but will remains empty after the upgrade in complete.

- **Handling Legacy Workers**: After the upgrade, legacy workers (those running a previous version of the codebase) may still be operational. To maintain compatibility, the migrated columns will continue to be inserted into the `queue` table until all workers are updated ([Reference: windmill-queue/src/jobs.rs#584](https://github.com/windmill-labs/windmill/pull/4689/files#diff-906910f106d16370529d12f9a05a23156ea1596ef3dee27518c8566476f54ce4R584)).
  
- **Support for Pre-Upgrade Jobs**: Completed jobs from before the update and queued jobs created by legacy workers may still exist. To address this, two `VIEW`s (`queue_view` and `completed_job_view`) have been added ([Reference: `queue_view` and `completed_job_view`](https://github.com/windmill-labs/windmill/pull/4689/files#diff-d991cb17730689c31489899e8e893e10ce7b5805f1058be29621e7dbe4e078bbR11)). Queries targeting the migrated columns are now routed through these views, which seamlessly retrieve data from the appropriate tables ([Reference: windmill-api/src/jobs.rs#685](https://github.com/windmill-labs/windmill/pull/4689/files#diff-55e9fc649a3fd37067faf6745654be78d1de1ac45ae778a58f7d65a167eaa382R685)).

The pinned version for this change is `1.427.0` right now, it must be [updated](https://github.com/windmill-labs/windmill/pull/4689/files#diff-ee79b848bc9b4eaf896a014db630cc35a767c4a6e48ee72af74b6e1b4efb5d88R592) if `1.427.0` is tagged without this change.

## Benchmarks

Performance benchmarks demonstrate substantial improvements when processing large raw scripts or executing complex workflows.

## Open questions

- Deleting from `completed_job` do not really delete the row, hence `job` rows exists forever. Also, the response on "delete" does not contains the moved columns **but** they does not seems to be used by the frontend.

- `VIEW`s where used where fallback queries could have been used instead, the performance impact of using views has been measured to be negligible, but we may still want to remove them in the future.

### DIFF
--- a/backend/windmill-api/src/jobs.rs
+++ b/backend/windmill-api/src/jobs.rs
@@ -683,23 +683,18 @@ async fn get_job(
 }
 
 lazy_static::lazy_static! {
-    static ref GET_COMPLETED_JOB_QUERY_NO_LOGS: String = generate_get_job_query(true, "completed_job");
-    static ref GET_COMPLETED_JOB_QUERY: String = generate_get_job_query(false, "completed_job");
-    static ref GET_QUEUED_JOB_QUERY_NO_LOGS: String = generate_get_job_query(true, "queue");
-    static ref GET_QUEUED_JOB_QUERY: String = generate_get_job_query(false, "queue");
+    static ref GET_COMPLETED_JOB_QUERY_NO_LOGS: String = generate_get_job_query(true, "completed_job_view");
+    static ref GET_COMPLETED_JOB_QUERY: String = generate_get_job_query(false, "completed_job_view");
+    static ref GET_QUEUED_JOB_QUERY_NO_LOGS: String = generate_get_job_query(true, "queue_view");
+    static ref GET_QUEUED_JOB_QUERY: String = generate_get_job_query(false, "queue_view");
 }
 fn generate_get_job_query(no_logs: bool, table: &str) -> String {
     let log_expr = if no_logs {
         "null".to_string()
     } else {
-        format!("right(concat(coalesce({table}.logs, ''), job_logs.logs), 20000)")
+        format!("right({table}.logs, 20000)")
     };
-    let join = if no_logs {
-        "".to_string()
-    } else {
-        format!("LEFT JOIN job_logs ON {table}.id = job_logs.job_id")
-    };
-    let additional_fields = if table == "completed_job" {
+    let additional_fields = if table == "completed_job_view" {
         "duration_ms,      
         success,        
         result,    
@@ -733,7 +728,6 @@ fn generate_get_job_query(no_logs: bool, table: &str) -> String {
     schedule_path, permissioned_as, flow_status, raw_flow, is_flow_step, language,
     raw_lock, email, visible_to_owner, mem_peak, tag, priority, {additional_fields}
     FROM {table}
-    {join}
     WHERE id = $1 AND {table}.workspace_id = $2");
 }
 pub async fn get_queued_job_ex(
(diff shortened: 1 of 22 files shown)


backend/windmill-common/src/utils.rs
```diff
@@ -37,6 +38,10 @@ lazy_static::lazy_static! {
         .timeout(std::time::Duration::from_secs(20))
         .connect_timeout(std::time::Duration::from_secs(10))
         .build().unwrap();
+    pub static ref GIT_SEM_VERSION: Version = Version::parse(
+        // skip first `v` character.
+        GIT_VERSION.split_at(1).1
```
uael: Possible panic here, but is GIT_VERSION supposed to be empty ?

backend/windmill-queue/src/jobs.rs
```diff
@@ -3895,6 +3927,27 @@ pub async fn push<'c, 'd, R: rsmq_async::RsmqConnection + Send + 'c>(
...
+        workspace_id,
+        raw_code,
+        raw_lock,
+        raw_flow.as_ref() as Option<&Json<FlowValue>>,
+        tag,
+    )
+    .execute(&mut tx)
+    .await?;
+
+    let (raw_code, raw_lock, raw_flow) = if !living_workers_are_at_least("1.422.0").await {
```
uael: Here and others: pin the right version just before merge

HugoCasa (dismissed):

uael:
Rebased on top of https://github.com/windmill-labs/windmill/pull/4695

backend/windmill-worker/src/worker.rs
```diff
@@ -1829,20 +1829,29 @@ async fn handle_queued_job<R: rsmq_async::RsmqConnection + Send + Sync + Clone>(
...
             rc.hash(&mut s);
             format!("inline_{}", hex::encode(s.finish().to_be_bytes()))
-        } else if let Some(rc) = job.raw_flow.as_ref() {
+        } else if let Some(sqlx::types::Json(rc)) = raw_flow.as_ref() {
             use std::hash::Hasher;
             let mut s = DefaultHasher::new();
-            serde_json::to_string(&rc.0)
-                .unwrap_or_default()
-                .hash(&mut s);
+            rc.get().hash(&mut s);
```
uael, line 1854: @rubenfiszel I did modified this as `raw_flow` is already a string, but, maybe `serde_json::to_string` was used to min the json before hashing ?

backend/windmill-worker/src/worker_flow.rs
```diff
@@ -1034,7 +1046,7 @@ pub async fn update_flow_status_after_job_completion_internal<
                     let args_hash =
                         hash_args(db, client, w_id, job_id_for_status, &flow_job.args).await;
                     let flow_path = flow_job.script_path();
-                    let version_hash = if let Some(sqlx::types::Json(s)) = flow_job.raw_flow.as_ref() {
+                    let version_hash = if let Some(sqlx::types::Json(s)) = raw_flow.as_ref() {
                         use std::hash::{Hash, Hasher};
                         let mut h = DefaultHasher::new();
                         s.get().hash(&mut h);
```
uael, line 1052: Same here yesterday, but forgot to ask

backend/windmill-queue/src/jobs.rs
```diff
@@ -4130,14 +4167,6 @@ async fn restarted_flows_resolution(
             } else {
                 // expect a module to be either a branchall (resp. loop), and resume the flow from this branch (resp. iteration)
                 let branch_or_iteration_n = branch_or_iteration_n.unwrap();
```
uael, line 4169: Not good, let me fix it

uael: Nop ok `module.id() == restart_step_id`

backend/windmill-api/src/jobs.rs
```diff
@@ -2306,6 +2300,14 @@ pub struct JobExtended<T> {
     #[serde(flatten)]
     inner: T,
 
+    // Imported from `JobDefinition`:
```
uael: To delete

PR #4746, opened 2024-11-19
proposal: wait for a complete upgrade before doing anything

uael:
## Proposal

To simplify future changes where handling living workers in a previous codebase version could be a bit tricky, I propose to wait for a complete upgrade **before** doing anything else.

With this change, the main will block until all workers are either dead or at the latest version.

## Cons

- I dont fully understand the repercussions nor implications of such change, so I'm not well suited to even make sure this is doable.
- This can be tricky to handle from the user POV, some support may be needed to smoothly land such radical change.

## Pros

- Future breaking changes to the codebase will be much easier to handle.

----

> [!IMPORTANT]
> Add blocking mechanism in `main.rs` to ensure all workers are updated to the latest version before proceeding with operations.
> 
>   - **Behavior**:
>     - In `main.rs`, added a loop to block execution until all workers are updated to the latest version using `update_min_version()`.
>     - Removed redundant `update_min_version()` call in `monitor_db()` in `monitor.rs`.
>   - **Functions**:
>     - `update_min_version()` in `worker.rs` checks for worker version discrepancies and returns a boolean indicating if all workers are up-to-date.
>   - **Misc**:
>     - Minor logging changes in `worker.rs` to include current and minimal worker versions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 02baed896aa851992ee52c18c31c03b678119481. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -43,7 +43,7 @@ use windmill_common::{
     scripts::ScriptLang,
     stats_ee::schedule_stats,
     utils::{hostname, rd_string, Mode, GIT_VERSION},
-    worker::{reload_custom_tags_setting, HUB_CACHE_DIR, TMP_DIR, WORKER_GROUP},
+    worker::{reload_custom_tags_setting, update_min_version, HUB_CACHE_DIR, TMP_DIR, WORKER_GROUP},
     DB, METRICS_ENABLED,
 };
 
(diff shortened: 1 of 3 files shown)


uael:
## Rejected

Doing so may lead to a very poor customer experience

PR #4809, opened 2024-11-27
Remove flock

pyranota:
> [!IMPORTANT]
> Removes `flock` usage in `python_executor.rs` for Unix systems when `no_uv_install` is false, executing commands directly instead.
> 
>   - **Behavior**:
>     - Removes `flock` usage in `handle_python_reqs()` and `handle_python_reqs()` for Unix systems when `no_uv_install` is false.
>     - Directly executes commands without `flock` when `no_uv_install` is false.
>   - **Functions**:
>     - Modifies `handle_python_reqs()` to conditionally use `flock` based on `no_uv_install` flag.
>     - Modifies `handle_python_reqs()` to execute commands directly if `no_uv_install` is false.
>   - **Misc**:
>     - Adjusts command formatting logic in `handle_python_reqs()` for Unix systems.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for cb4e2ca05b37f0e440036e69a71a9596a00103c6. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/backend/windmill-worker/src/python_executor.rs
+++ b/backend/windmill-worker/src/python_executor.rs
@@ -1322,7 +1322,11 @@ pub async fn handle_python_reqs(
         } else {
             let fssafe_req = NON_ALPHANUM_CHAR.replace_all(&req, "_").to_string();
             #[cfg(unix)]
-            let req = format!("'{}'", req);
+            let req = if no_uv_install {
+                format!("'{}'", req)
+            } else {
+                req.clone()
+            };
 
             #[cfg(windows)]
             let req = format!("{}", req);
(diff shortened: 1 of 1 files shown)


backend/windmill-worker/src/python_executor.rs
```diff
@@ -1322,7 +1322,11 @@ pub async fn handle_python_reqs(
         } else {
             let fssafe_req = NON_ALPHANUM_CHAR.replace_all(&req, "_").to_string();
             #[cfg(unix)]
-            let req = format!("'{}'", req);
```
rubenfiszel, line 1325: sure about that change?

pyranota: Yes, if we dont do this uv will treat ' as a part of actual package name for some reason. ' also ignored on windows workers, since they already call uv directly.

pyranota: That what happens if we don't do this:
```
absl-py==2.1.0 is being installed for the first time.
 It will be cached for all ulterior uses.
error: Failed to parse: `'absl-py==2.1.0'`
  Caused by: Expected package name starting with an alphanumeric character, found `'`
'absl-py==2.1.0'
^
```

PR #4839, opened 2024-12-04
fix no_uv not affecting deploy

pyranota:
Before this fix no_uv, no_uv_compile and no_uv_install were not affecting Dependency jobs.

These jobs are only affected if used USE_PIP_COMPILE or USE_PIP_INSTALL env variables

To make it more consistent, no_uv should also affect dep jobs.

Also makes ansible use uv by default


pyranota:
Already implemented in #4883, closing

PR #4967, opened 2024-12-22
wmill cli flow push - avoid hardcoded suffix

cherusk:
File path shall be file as denoted in cli docs.  Hardcoded suffix counter productive.


rubenfiszel:
It should not break previous behavior so at very least, it should try both location (previous and new one)

rubenfiszel:
It might be a bugfix but the behavior has been like this for so-long that we would want to not break existing users relying on the actual current behavior. Keeping compatibility for both is very little code complexity so it is worth it.

cherusk:
ACK. My ts caps are limited, but I changed it to your suggestion.

With precedence on specific file path to config file.

cherusk:

rubenfiszel:
Actually the correct way to do this would be to change the push function (the frontend for `wmill flow push`) rather than `pushFlow` which is used elsewhere internally. 
You can do the opposite there and remove flow.yaml from the file path provided if it exists

cherusk:
Ok, I have worked around it for now on my end.

I might look at this later as it is not urgent for me. Anyone free to adjust the PR or turn it into a issue if worth it ...

PR #4999, opened 2024-12-31
backend: simplify `insert` in `completed_job`

uael:


uael:
tested but not fully tested

backend/windmill-queue/src/jobs.rs
```diff
@@ -531,133 +533,118 @@ pub async fn add_completed_job<T: Serialize + Send + Sync + ValidableJson>(
...
         .map_err(|e| Error::InternalErr(format!("Could not add completed job {job_id}: {e:#}")))?;
 
-
+        // TODO(uael): to remove ?
         if !queued_job.is_flow_step {
-            if _duration > 500
-                && (queued_job.job_kind == JobKind::Script
-                    || queued_job.job_kind == JobKind::Preview)
-            {
-                if let Err(e) = sqlx::query!(
```
uael: @rubenfiszel Removed this as after this change, `completed_job.flow_status` should have the same value as `q.flow_status`

PR #5029, opened 2025-01-08
backend: queue v2 migration steps

uael:


uael:
Because of how views works (not null constraints not reflected), this migrations make sqlx queries impossible to compile (e.g. Uuid/Option<Uuid>), **but** all the already running code remain 100% compatible, while using the new schema. The caveats are pretty small and are discarded when all running server(s)/workers(s) are up to date:
- Some columns and their data are duplicated (e.g. __flow_status in v2_job_queue and flow_status in v2_job_flow_runtime)
- Each insert/update on `queue` are reflected on v2_* tables using triggers -> overhead.
- Each insert on `completed_job` are reflected on v2_* tables using triggers -> overhead.
I will add a custom migration step which once everything is up to date (i.e. all processes run under the new version), will drop the triggers, the compatibility views, and the old columns.
After this migration step, no more overhead or duplicated and the migration is complete.

To resume:
- [x] Transparant migration `v1` -> `v2`
- [ ] Missing triggers for `v2` -> `v1` compatibility
- [ ] Missing indexes, policies and triggers
- [ ] Update code to use updated tables and not the compatibility views
- [ ] Add custom migration step to finish the migration when all workers are up to date

uael:
Another option is to temporarily change all checked queries that does not compile because of views to unchecked query in order to merge this, which allow to discover if the transparent migration works correctly, and then proceed with next steps:

This PR:
- [x] Transparant migration `v1` -> `v2`
- [ ] Missing triggers for `v2` -> `v1` compatibility
- [ ] Missing indexes, policies and triggers
- [ ] Tmp checked queries to unchecked

Next PR(s):
- [ ] Update code to use updated tables and not the compatibility views
- [ ] Re-check queries
- [ ] Add custom migration step to finish the migration when all workers are up to date

backend/migrations/20250107121225_queue_v2.up.sql
```diff
@@ -0,0 +1,888 @@
...
+FOR EACH ROW
+EXECUTE FUNCTION v2_job_queue_after_insert();
+
+-- 5. On every update to `v2_job_queue`, update `v2_job`, `v2_job_runtime` and `v2_job_flow_runtime` as well
+-- This trigger will be removed once all server(s)/worker(s) are updated to use `v2_*` tables
+CREATE OR REPLACE FUNCTION v2_job_queue_before_update()
+RETURNS TRIGGER AS $$
+BEGIN
+    -- `v2_job`: Only `args` are updated
+    IF NEW.__args IS DISTINCT FROM OLD.__args THEN
```
uael: Text cast

backend/migrations/20250110124431_v2_job.up.sql
```diff
@@ -0,0 +1,64 @@
+-- Add up migration script here
+ALTER TABLE v2_job
+    ADD COLUMN IF NOT EXISTS created_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+    ADD COLUMN IF NOT EXISTS created_by VARCHAR(255) DEFAULT 'missing' NOT NULL,
+    ADD COLUMN IF NOT EXISTS created_by_email VARCHAR(255) DEFAULT 'missing@email.xyz' NOT NULL,
```
hugocasa: permissioned_as_email

backend/migrations/20250110124431_v2_job.up.sql
```diff
@@ -0,0 +1,64 @@
...
+    ADD COLUMN IF NOT EXISTS permissioned_as VARCHAR(55) DEFAULT 'g/all' NOT NULL,
+    ADD COLUMN IF NOT EXISTS kind job_kind DEFAULT 'script'::job_kind NOT NULL,
+    ADD COLUMN IF NOT EXISTS entity_id BIGINT,
+    ADD COLUMN IF NOT EXISTS entity_path VARCHAR(255),
```
hugocasa: runnable_id/path

HugoCasa (commented):
- add metadata field?
- just wondering if there is a better alternative than checking pg_trigger_depth

uael:
> - add metadata field?
> 
> - just wondering if there is a better alternative than checking pg_trigger_depth

There is but slightly more complex, I go for the simplest one since they are only temporary triggers. WDYT ?

HugoCasa:
yes i guess as it's temporary it's not worth doing more complex

backend/migrations/20250110124743_v2_job_queue_sync.up.sql
```diff
@@ -7,29 +7,38 @@ RETURNS TRIGGER AS $$
 DECLARE
     job v2_job;
 BEGIN
-    job := (SELECT * FROM v2_job WHERE id = NEW.id);
+    -- When inserting to `v2_job_queue` from `v2` code, set `v1` columns:
```
uael, line 10: Move to phase 1.

backend/migrations/20250110124744_v2_job_completed_sync.up.sql
```diff
@@ -10,17 +10,20 @@ DECLARE
 BEGIN
     IF NEW.__created_by IS NULL THEN
         -- v2 -> v1
-        job := (SELECT * FROM v2_job WHERE id = NEW.id);
+        SELECT * INTO job FROM v2_job WHERE id = NEW.id;
```
uael, line 13: Move to phase 1.

uael (commented):
- [x] Keep completed job `deleted` column

PR #5085, opened 2025-01-17
set activate workspace in .wmill/activeWorkspace in workspace directory

wenbinye:
Set activate workspace in .wmill/activeWorkspace in workspace directory by command 
`wmill workspace switch <workspace>`
The workspace activate when switch to the directory automatic.

### DIFF
--- a/cli/workspace.ts
+++ b/cli/workspace.ts
@@ -40,9 +40,13 @@ async function getActiveWorkspaceName(
     return opts?.workspace;
   }
   try {
-    return await Deno.readTextFile((await getRootStore()) + "/activeWorkspace");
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Thanks for the contribution.

It's a nice idea and I did want to do it too. It's lacking a bit of discovery and relying on the existence or not of .wmill which I'm not a fan. 

I suggest to have an option: --persist-in-folder to the switch command that will indeed write there (and create dir if necessary) but otherwise it's not written there. So it is used by default only if `.wmill/activeInstance` in which case a log mention that it is used and how to override it or delete it.

When it's not used because it doesn't exist, a log mention that it's using the global setting because no `.wmill/activeInstance` was found

PR #5119, opened 2025-01-23
backend: use v2 tables through views where possible (v2 phase 3)

uael:


backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
```diff
@@ -0,0 +1,177 @@
...
+    j.runnable_id              AS script_hash,
+    j.runnable_path            AS script_path,
+    j.args,
+    j.raw_code,
+    q.canceled_by IS NOT NULL  AS canceled,
+    q.canceled_by,
+    q.canceled_reason,
+    r.ping                     AS last_ping,
+    j.kind                     AS job_kind,
+    j.trigger                  AS schedule_path,
```
hugocasa: check trigger kind

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
```diff
@@ -0,0 +1,177 @@
...
+    j.flow_root_job            AS root_job,
+    f.leaf_jobs,
+    j.tag,
+    j.concurrent_limit,
+    j.concurrency_time_window_s,
+    j.timeout,
+    j.flow_step_id,
+    j.cache_ttl,
+    j.priority,
+    NULL::TEXT                 AS logs
```
hugocasa: why NULL for logs?

hugocasa: same for completed_job

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
hugocasa: same for completed_job

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
hugocasa, line 1: why is env_id completely gone? not used at all? same for completed_job

HugoCasa (commented):
[phase 3 review]
- i would definitely change the view names to include view (e.g. _view)
- don't use view triggers for update but directly change to the new tables

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: Phase 1 moved everything to `job_logs` table and the code is already considering this column as a fallback since it's was kept for backward compatibility, after v2 we could already remove all usage of this and only rely on the job_logs table

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: Yes didn't find any usage so removed it

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: Isn't it only used for schedules as of now ?

hugocasa: true, i guess it depends if in the next step, we migrate all at once the queries and delete the views or if we do it progressively

backend/migrations/20250117124431_v2_job.up.sql
```diff
@@ -0,0 +1,67 @@
+-- Add up migration script here
+CREATE TYPE job_trigger_kind AS ENUM ('webhook', 'http', 'websocket', 'kafka', 'email', 'nats', 'schedule', 'app', 'ui');
```
hugocasa: add 'postgres'

backend/migrations/20250117124431_v2_job.up.sql
```diff
@@ -0,0 +1,67 @@
...
+
+CREATE POLICY see_member ON v2_job
+    AS PERMISSIVE
+    FOR ALL
+    TO windmill_user
+    USING ((SPLIT_PART((permissioned_as)::TEXT, '/'::TEXT, 1) = 'g'::TEXT) AND
+           (SPLIT_PART((permissioned_as)::TEXT, '/'::TEXT, 2) = ANY
+            (REGEXP_SPLIT_TO_ARRAY(CURRENT_SETTING('session.groups'::TEXT), ','::TEXT))));
+
+GRANT ALL ON v2_job TO windmill_user, windmill_admin;
```
hugocasa: admin policy?

backend/migrations/20250117124743_v2_job_queue_sync.up.sql
```diff
@@ -0,0 +1,181 @@
...
+        concurrent_limit = EXCLUDED.concurrent_limit,
+        concurrency_time_window_s = EXCLUDED.concurrency_time_window_s,
+        cache_ttl = EXCLUDED.cache_ttl,
+        timeout = EXCLUDED.timeout,
+        priority = EXCLUDED.priority,
+        args = EXCLUDED.args,
+        pre_run_error = EXCLUDED.pre_run_error,
+        raw_code = COALESCE(v2_job.raw_code, EXCLUDED.raw_code),
+        raw_lock = COALESCE(v2_job.raw_lock, EXCLUDED.raw_lock),
+        raw_flow = COALESCE(v2_job.raw_flow, EXCLUDED.raw_flow)
```
hugocasa: why a coalesce here?

backend/migrations/20250117124748_v2_migrate_from_v1.up.sql
```diff
@@ -0,0 +1,144 @@
...
+) SELECT
+    id, workspace_id, __created_at, __created_by, __permissioned_as, __email,
+    __job_kind, __script_hash, __script_path, __parent_job,
+    __language,
+    __schedule_path, CASE WHEN __schedule_path IS NOT NULL THEN 'schedule'::job_trigger_kind END,
+    __tag, __visible_to_owner, __priority,
+    __args,
+    __raw_code, __raw_lock, __raw_flow
+FROM v2_job_completed
+WHERE NOT EXISTS (SELECT 1 FROM v2_job WHERE v2_job.id = v2_job_completed.id);
```
hugocasa: why would there be missing rows in job?

backend/migrations/20250117124748_v2_migrate_from_v1.up.sql
```diff
@@ -0,0 +1,144 @@
...
+    logs = CONCAT(job_logs.logs, EXCLUDED.logs)
+;
+
+-- Migrate old `v2_job_completed.__logs` to `job_logs`
+INSERT INTO job_logs (job_id, workspace_id, logs)
+SELECT id, workspace_id, __logs
+FROM v2_job_completed
+WHERE __logs IS NOT NULL AND __logs IS DISTINCT FROM '##DELETED##'
+ON CONFLICT (job_id) DO UPDATE SET
+    logs = CONCAT(job_logs.logs, EXCLUDED.logs)
```
hugocasa: wasn't job_logs introduced before?

backend/migrations/20250117145629_v2_job_completed_constraints.up.sql
hugocasa, line 1: why can you only drop/set the constraints at the end?

backend/migrations/20250117124743_v2_job_queue_sync.up.sql
hugocasa: and why would there be a conflict?

backend/migrations/20250117124743_v2_job_queue_sync.up.sql
```diff
@@ -0,0 +1,181 @@
...
+            leaf_jobs = COALESCE(v2_job_flow_runtime.leaf_jobs, EXCLUDED.leaf_jobs)
+        ;
+    END IF;
+    IF NEW.__logs IS NOT NULL THEN
+        INSERT INTO job_logs (job_id, workspace_id, logs)
+        VALUES (NEW.id, NEW.workspace_id, NEW.__logs)
+        ON CONFLICT (job_id) DO UPDATE SET
+            logs = CONCAT(job_logs.logs, EXCLUDED.logs)
+        ;
+        NEW.__logs := NULL;
```
hugocasa: aren't logs already pushed to job_logs?

HugoCasa (commented):
[phase 1 review]
- looks good to me, mostly comprehension questions

backend/migrations/20250117124748_v2_migrate_from_v1.up.sql
hugocasa: as this table isn't new

backend/windmill-queue/src/jobs.rs
```diff
@@ -2424,15 +2422,27 @@ async fn get_completed_flow_node_result_rec(
...
+                FROM completed_job
+                WHERE parent_job = $1 AND workspace_id = $2 AND flow_status IS NOT NULL",
+                id,
+                w_id
+            )
+            .map(|record| (record.id, record.flow_status.0))
+            .fetch_all(db)
+            .await?
+            .into_iter();
+            match Box::pin(get_completed_flow_node_result_rec(
```
hugocasa: why pin?

HugoCasa (commented):
[phase 2 review]
looks good to me

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: let me add a check then

backend/migrations/20250117124743_v2_job_queue_sync.up.sql
uael: we introduced the job `table` not so long ago and all `raw_*` values were moved to it instead, so old code prior that change branch to the `EXCLUDED.*` side of the `COALESCE` while current main branch to `v2_job.*` side as current `push` first insert raw values to job then insert to queue

backend/migrations/20250117124748_v2_migrate_from_v1.up.sql
uael: it isn't new but not guaranteed to exists, as it was introduced not so long ago without being synced; current main have version checks in Rust in order to determine where data is actually located, either in job or {queue,completed_job}, after this migration the database is 100% synced to latest schema

backend/migrations/20250117124748_v2_migrate_from_v1.up.sql
uael: kind of the same as for raw values: this (and the corresponding code in triggers) make sure all `logs` are actually in `job_logs` so we can remove backward compatibility Rust code which coalesce `{completed_job,queue}.logs` and `job_logs.logs`

backend/migrations/20250117145629_v2_job_completed_constraints.up.sql
uael: it needs to be done after migration (in `v2_migrate_from_v1`) for new columns (e.g. status and completed_at are NULL before migration):
```sql
ALTER TABLE v2_job_completed ALTER COLUMN status SET NOT NULL;
ALTER TABLE v2_job_completed ALTER COLUMN completed_at SET DEFAULT now();
ALTER TABLE v2_job_completed ALTER COLUMN completed_at SET NOT NULL;
```
all other constraints are grouped together in the same file for readability

backend/migrations/20250117124743_v2_job_queue_sync.up.sql
uael: See https://github.com/windmill-labs/windmill/pull/5119#discussion_r1929988946

backend/windmill-queue/src/jobs.rs
uael: it's a recursive async call hence require boxing to avoid infinite future size

backend/migrations/20250117124431_v2_job.up.sql
uael: added

backend/migrations/20250117124431_v2_job.up.sql
uael: added

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: see https://github.com/windmill-labs/windmill/pull/5119#discussion_r1929988946 for more context

backend/migrations/20250117145630_v2_queue_compatibility_view.up.sql
uael: added

PR #5165, opened 2025-01-28
feat: add postgres trigger captures

dieriba:
> [!IMPORTANT]
> Add support for Postgres triggers, including UI components, trigger management, and database configuration validation.
> 
>   - **Behavior**:
>     - Add support for Postgres triggers in various components like `CaptureWrapper.svelte`, `TriggersEditor.svelte`, and `PostgresTriggersPanel.svelte`.
>     - Implement `PostgresTriggerService` to manage Postgres trigger operations.
>     - Introduce `CheckPostgresRequirement.svelte` to verify database configuration.
>   - **UI Components**:
>     - Add `PostgresEditorConfigSection.svelte` for configuring Postgres triggers.
>     - Update `CaptureButton.svelte` and `TriggersEditorSection.svelte` to include Postgres as a trigger option.
>     - Modify `RelationPicker.svelte` to handle Postgres table and schema selection.
>   - **Utilities**:
>     - Add `invalidRelations` function in `utils.ts` to validate Postgres table and schema configurations.
>   - **Misc**:
>     - Update `script_helpers.ts` to include Postgres in the `WmTrigger` type.
>     - Modify `ShareModal.svelte` to handle Postgres triggers.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for be5427a257cc81f1557f6c679546a29ae0e805ad. It will automatically update as commits are pushed.</sup>


backend/migrations/20250102145420_more_captures.up.sql
```diff
@@ -1,5 +1,5 @@
 -- Add up migration script here
-CREATE TYPE TRIGGER_KIND AS ENUM ('webhook', 'http', 'websocket', 'kafka', 'email');
+CREATE TYPE TRIGGER_KIND AS ENUM ('webhook', 'http', 'websocket', 'kafka', 'email', 'postgres');
```
hugocasa: you shouldn't change a past migration, create a new migration which alters the type

backend/windmill-api/src/capture.rs
```diff
@@ -186,18 +212,87 @@ async fn get_configs(
...
+}
+
 async fn set_config(
     authed: ApiAuthed,
     Extension(user_db): Extension<UserDB>,
+    #[cfg(feature = "postgres_trigger")] Extension(db): Extension<DB>,
     Path(w_id): Path<String>,
     Json(nc): Json<NewCaptureConfig>,
 ) -> Result<()> {
+    #[cfg(feature = "postgres_trigger")]
```
hugocasa, line 288:
```suggestion
#[cfg(feature...
let nc = if let TriggerKind::Postgres = nc.trigger_kind {
  ...
else {
  nc
};
```

backend/windmill-api/src/postgres_triggers/handler.rs
```diff
@@ -112,48 +106,38 @@ pub struct NewPostgresTrigger {
...
+            error::Error::BadConfig(format!("Error connecting to postgres: {}", err.to_string()))
+        })
     };
-
-    PgConnection::connect_with(&options)
+    tokio::time::timeout(tokio::time::Duration::from_secs(30), connect_f)
         .await
-        .map_err(Error::SqlErr)
+        .map_err(|_| {
+            error::Error::BadConfig(format!("Timeout connecting to websocket after 30 seconds"))
```
hugocasa: not websocket

frontend/src/lib/components/triggers/CaptureWrapper.svelte
```diff
@@ -26,16 +28,25 @@
...
-			requestBody: {
-				trigger_kind: captureType,
-				path,
-				is_flow: isFlow,
-				trigger_config: args && Object.keys(args).length > 0 ? args : undefined
-			},
-			workspace: $workspaceStore!
-		})
+	export async function setConfig(): Promise<boolean> {
+		if (captureType === 'postgres' && invalidRelations(args.publication.table_to_track, true)[0]) {
```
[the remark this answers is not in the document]
hugocasa: you should check the type of args nested attributes, it could be anything

frontend/src/lib/components/triggers/TriggersEditorSection.svelte
```diff
@@ -31,7 +31,8 @@
 		webhook: 'Webhook',
 		kafka: '+ New Kafka trigger',
 		email: 'Email trigger',
-		nats: '+ New NATS trigger'
+		nats: '+ New NATS trigger',
+		postgres: '+ New postgres trigger'
```
hugocasa: Postgres

frontend/src/lib/script_helpers.ts
```diff
@@ -705,11 +705,12 @@ class Nats(TypedDict):
 	length: int
 
 class WmTrigger(TypedDict):
-	kind: Literal["http", "email", "webhook", "websocket", "kafka", "nats"]
+	kind: Literal["http", "email", "webhook", "websocket", "kafka", "nats", "postgres"]
 	http: Http | None
 	websocket: Websocket | None
 	kafka: Kafka | None
 	nats: Nats | None
+ 	database: None
```
hugocasa: why that?

backend/windmill-api/src/capture.rs
```diff
@@ -157,6 +182,7 @@ struct NewCaptureConfig {
     trigger_kind: TriggerKind,
     path: String,
     is_flow: bool,
+    // why is the option here ?
     trigger_config: Option<TriggerConfig>,
```
hugocasa: legacy i think, you can fix it or add TODO to your comment

frontend/src/lib/components/triggers/postgres/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    schemaIndex: number
+    tableIndex: number
+    schemaError: boolean
+    tableError: boolean
+    schemaName?: string
+}
+export function invalidRelations(
+    relations: Relations[],
+    showError?: boolean
+): [boolean, RelationError] {
```
hugocasa: better to return an object

HugoCasa (changes requested):
- Missing a check button in the captures UI + should be less visible and not title case + enter what command to run and indicate you need to restart OR link to docs
Screenshot 2025-02-04 at 17 45 32
- when saving from capture panel, doesn't forward the configuration to the drawer

HugoCasa:
@Guilhem-lm  do you think we should also have static mode in addition to resource mode for postgres triggers (like for kafka and nats)?
i'm just afraid it will be a lot of fields

backend/windmill-api/src/capture.rs
dieriba: I forgot the remove it, I wanted to ask you why is there an option as all trigger require config for I may be wrong and some trigger may not require config ?

frontend/src/lib/script_helpers.ts
dieriba: for consistency

frontend/src/lib/components/triggers/CaptureWrapper.svelte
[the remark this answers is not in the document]
dieriba: wdym ?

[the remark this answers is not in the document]
hugocasa: args is Record<string, any> so if args.publication is undefined the frontend crashes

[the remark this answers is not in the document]
dieriba: ohhh you right fixing it

Guilhem-lm:
@dieriba nit : maybe instead of toasting the user when trying to add a schema while previous schema has no name you could just disable the +Add Schema button and add a popover that says "Fill previous schema name before adding a new one"

An other solution is to do the same UI as when adding an input to the flow schema, so that you cannot add a schema without a name

frontend/src/lib/components/triggers/postgres/PostgresEditorConfigSection.svelte
```diff
@@ -0,0 +1,126 @@
...
+			  }
+			: publication
+	$: notEmpty = publication.table_to_track && publication.table_to_track.length > 0
+	let selectedTable: 'all' | 'specific' = notEmpty ? 'specific' : 'all'
+	$: isValid =
+		!emptyString(postgres_resource_path) &&
+		publication.transaction_to_track.length > 0 &&
+		(selectedTable === 'all' || (notEmpty ?? false))
+
+	$: if (emptyString(postgres_resource_path)) {
```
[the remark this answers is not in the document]
hugocasa: @dieriba do you actually need this reactive statement?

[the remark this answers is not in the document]
dieriba: I put that reactive statement assuming that when using removed the postgres resource path it meant that they will choose another and thus will choose schema appropriate for that Postgres resource that why I reset it each postgres_resource_path is empty

HugoCasa (approved):

PR #5167, opened 2025-01-29
nix: Update flake for C# and oracle + add flake for building parsers

wendrul:


flake.nix
```diff
@@ -10,6 +10,7 @@
     let
       pkgs = import nixpkgs {
         inherit system;
+        config.allowUnfree = true;
```
uael, line 13: for oracle ?

wendrul: yes

PR #5173, opened 2025-01-30
feat: add bigquery to the list of python depenency exceptions

arjun-1:
The native BigQuery script is not suited for us (cannot handle > 10.000 rows and no `setFlowUserState`), hence the need to run BigQuery within python. The import root is not the same as the package name: `from google.cloud import bigquery` hence the need for this exception as mentioned here: https://www.windmill.dev/docs/advanced/dependencies_in_python#lockfile-per-script-inferred-from-imports-standard


arjun-1:

rubenfiszel:
Thanks @arjun-1 , we will improve the capability to handle more than 10k rows, we weren't aware of that limitation.

PR #5181, opened 2025-01-30
fix: restart zombie job correctly handle concurrency limits

rubenfiszel:


backend/src/monitor.rs
```diff
@@ -1506,9 +1506,20 @@ pub async fn reload_base_url_setting(db: &DB) -> error::Result<()> {
...
-                "UPDATE queue SET running = false, started_at = null
-                WHERE last_ping < now() - ($1 || ' seconds')::interval
-                 AND running = true AND job_kind NOT IN ('flow', 'flowpreview', 'flownode', 'singlescriptflow') AND same_worker = false RETURNING id, workspace_id, last_ping",
+                "WITH zombie_jobs AS (
+                    UPDATE queue SET running = false, started_at = null
+                    WHERE last_ping < now() - ($1 || ' seconds')::interval
+                     AND running = true AND job_kind NOT IN ('flow', 'flowpreview', 'flownode', 'singlescriptflow') AND same_worker = false 
+                    RETURNING id, workspace_id, last_ping
+                ),
+                update_concurrency AS (
```
uael, line 1515: Is this CTE executed even if not in the in the final FROM clause? I'm not sure about that but I think CTE are lazy

PR #5199, opened 2025-02-03
feat(python, windows): enable S3 to cache wheels

pyranota:
#5196 Allows us to handle it safely


pyranota:
/ai review that tarball logic and s3 offload/load is valid for windows as well.

pyranota:
/ai can you check unhappy paths and see if it's resilient and would recover in case of failure created anywhere on introduced paths

PR #5487, opened 2025-03-18
feat: list references upon renaming a script or a flow

diegoimbert:
List of apps and flows referencing a script or flow upon renaming it

Screenshot 2025-03-18 at 16 48 12

----

> [!IMPORTANT]
> Adds functionality to list and warn about references to scripts or flows when renaming, with backend API and frontend UI updates.
> 
>   - **Behavior**:
>     - Lists apps and flows referencing a script or flow upon renaming in `Path.svelte`.
>     - Displays a warning if renaming might break references.
>   - **Backend**:
>     - Adds new API endpoints in `openapi.yaml` for listing paths from workspace runnables.
>     - Updates `workspace_runnable_dependencies` table in migrations to support both apps and flows.
>     - Implements new queries in `apps.rs`, `flows.rs`, and `scripts.rs` to fetch referencing paths.
>   - **Frontend**:
>     - Updates `Path.svelte` to show references and warnings using `FlowService`, `AppService`, and `ScriptService`.
>     - Adds utility functions `localeConcatAnd` and `pluralize` in `utils.ts` for formatting reference lists.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 9ac352f13196f46a2b88f04bd26225f209bc6e2b. It will automatically update as commits are pushed.</sup>


backend/migrations/20250317145509_more_generic_workspace_runnable_usage.up.sql
```diff
@@ -0,0 +1,21 @@
+-- flow_workspace_runnables only stored runnable usages by
+-- flows although apps can also use runnables
+
+ALTER TABLE flow_workspace_runnables
+RENAME TO workspace_runnable_dependencies;
+
+ALTER TABLE workspace_runnable_dependencies ALTER flow_path DROP NOT NULL;
```
HugoCasa, line 7: you can reuse flow_path for apps and just rename the column to path + add a column to differentiate between flows and apps

backend/migrations/20250317145509_more_generic_workspace_runnable_usage.up.sql
```diff
@@ -0,0 +1,21 @@
...
+
+ALTER TABLE workspace_runnable_dependencies 
+ADD COLUMN app_path VARCHAR(255);
+
+ALTER TABLE workspace_runnable_dependencies
+ADD CONSTRAINT fk_workspace_runnable_dependencies_app_path
+FOREIGN KEY (app_path, workspace_id) REFERENCES app (path, workspace_id);
```
HugoCasa, line 14: that would probably prevent inserting flow dependencies

frontend/src/lib/components/Path.svelte
```diff
@@ -503,7 +532,46 @@
...
+			{#if pathUsageInApps && pathUsageInApps.length}
+				<Alert
+					type="warning"
+					class="mt-4"
+					title="Moving this item will break the following apps referencing it :"
+				>
+					<ul>
+						{#each pathUsageInApps as appPath}
+							<li>
+								• <a href={`/apps/edit/${appPath}`} class="text-blue-400" target="_blank">
```
hugocasa: nit: you can use list-disc instead of putting the dot directly

frontend/src/lib/components/Path.svelte
```diff
@@ -503,7 +532,46 @@
...
 	</div>
 
-	{#if kind != 'app' && kind != 'schedule' && kind != 'http_trigger' && kind != 'websocket_trigger' && initialPath != '' && initialPath != undefined && initialPath != path}
+	{#if pathUsageInFlowsPromise || pathUsageInAppsPromise}
+		{#await pathUsageInFlowsPromise then pathUsageInFlows}
+			{#if pathUsageInFlows && pathUsageInFlows.length}
+				<Alert
+					type="warning"
+					class="mt-4"
+					title="Moving this item will break the following flows referencing it :"
```
hugocasa: there should be no space before ":"

frontend/src/lib/components/Path.svelte
```diff
@@ -330,6 +330,35 @@
 	const openSearchWithPrefilledText: (t?: string) => void = getContext(
 		'openSearchWithPrefilledText'
 	)
+
+	$: displayPathChangedWarning =
+		kind != 'app' &&
+		kind != 'schedule' &&
+		kind != 'http_trigger' &&
+		kind != 'websocket_trigger' &&
```
hugocasa: you should check that kind = script or flow directly

HugoCasa (commented):
have you tested that it works for both "normal" and background runnables?

backend/migrations/20250317145509_more_generic_workspace_runnable_usage.up.sql
hugocasa: missing on delete cascade

hugocasa: also FK constraint can be annoying on path renaming, make sure it still works for apps

frontend/src/lib/components/Path.svelte
diegoimbert: Isn't the warning necessary for all other kinds such as resources ?

hugocasa: oh my bad, didn't see it was used like this before
but actually can you change it to kind = script/flow/variable/resource?

diegoimbert: my reasoning was that it's better to warn than not in case we add new kinds in the future. but it's indeed less explicit

backend/migrations/20250317145509_more_generic_workspace_runnable_usage.up.sql
diegoimbert: fixed

HugoCasa (approved):

diegoimbert:
- List scripts/flows/apps in the same alert box
- Number of usages tooltip even when not renaming
image

PR #5509, opened 2025-03-26
feat(python): show tip about pinning requirements

pyranota:


frontend/src/lib/components/DisplayResult.svelte
```diff
@@ -637,6 +637,18 @@
 							team.
 						</Alert>
 					{/if}
+					{#if (language === 'python3' || result?.error?.message?.startsWith("exit code for \"python run\":")) && result?.error?.message?.includes("ImportError: cannot import name")}
```
rubenfiszel: shouldn't it be 'language &&

frontend/src/lib/components/DisplayResult.svelte
```diff
@@ -637,6 +637,18 @@
...
+						<Alert size="xs" type="info" title="Seeing an odd import error?">
+							Python requirements inference may be inaccurate. This is due to the fact that requirement names can vary from package names they provide.
+							Try to <a href="https://www.windmill.dev/docs/advanced/dependencies_in_python#pinning-dependencies-and-requirements" target="_blank" rel="noopener noreferrer">manually pin requirements</a>
+						</Alert>
+					{/if}
+					{#if result?.error?.message?.startsWith("execution error:\npip compile failed")}
```
rubenfiszel: missing language &&

frontend/src/lib/components/DisplayResult.svelte
rubenfiszel: I mean language == 'python3' &&

pyranota: language is not always provided by the callers. This way I make assumption that language can be undefined

PR #5600, opened 2025-04-10
fix: infinite loop in resource picker

diegoimbert:
dispatches event even when value did not change change

----

> [!IMPORTANT]
> Fixes infinite loop in `ResourcePicker.svelte` by dispatching 'change' event only when `value` changes.
> 
>   - **Behavior**:
>     - Fixes infinite loop in `ResourcePicker.svelte` by dispatching 'change' event only when `value` changes.
>     - Introduces `oldValue` to track previous `value` and compare for changes.
>   - **Code Changes**:
>     - Modifies event dispatch logic in `ResourcePicker.svelte` to use `(oldValue !== value) && dispatch('change', (oldValue = value))`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 480e2acba591bad33de36b17f14c35c56dd8b502. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/src/lib/components/ResourcePicker.svelte
+++ b/frontend/src/lib/components/ResourcePicker.svelte
@@ -41,7 +41,7 @@
 					value: value ?? initialValue,
 					label: value ?? initialValue,
 					type: valueType
-			  }
+				}
 			: undefined
 
 	$: if (value === undefined && initialValue) {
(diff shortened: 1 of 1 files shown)


HugoCasa:
wouldn't it be better to update the component to runes?

diegoimbert:
> wouldn't it be better to update the component to runes?

totally but i did this as fast as possible as i thought it was urgent, while waiting for my burning laptop to work lol

diegoimbert:
> wouldn't it be better to update the component to runes?

also not sure it would actually fix that particular problem

HugoCasa (approved):

PR #5660, opened 2025-04-24
fix: check for valid teams_channel config when saving critical alerts settings

alpetric:
- some formatting
- typeguard for teams_channel
- updating error log (checking all types of channel not only email)
- formatting removing white spaces


frontend/src/lib/components/InstanceSettings.svelte
```diff
@@ -132,6 +132,14 @@
 			setupSnowflakeUrls()
 		}
 
+		// Remove empty or invalid teams_channel entries
+		$values.critical_error_channels = $values.critical_error_channels.filter((entry) => {
+			if ('teams_channel' in entry) {
```
rubenfiszel: need to check that entry is an object and not null

PR #5761, opened 2025-05-16
Di/duckdb

diegoimbert:


backend/windmill-worker/src/duckdb_executor.rs
```diff
@@ -0,0 +1,661 @@
...
+                        transform_attach_db_resource_query(&parsed, &job.id, client).await?,
+                    ),
+                    None => v.push(query_block.to_string()),
+                };
+            }
+            v
+        };
+
+        let bq_credentials_path = make_bq_credentials_path(&job.id);
+        env::set_var("GOOGLE_APPLICATION_CREDENTIALS", &bq_credentials_path);
```
rubenfiszel: suspicious that this is always set

backend/windmill-worker/src/duckdb_executor.rs
```diff
@@ -0,0 +1,661 @@
...
+            .map_err(|_e| Error::ExecutionErr("failed project_id deserialize".to_string()))?;
+            let attach_str = format!(
+                "ATTACH 'project={}' as {} (TYPE bigquery{});",
+                project_id,
+                parsed.name,
+                parsed.extra_args.unwrap_or("")
+            )
+            .to_string();
+            Ok(vec![
+                "INSTALL bigquery FROM community;".to_string(),
```
rubenfiszel, line 565: shouldn't we do that only once (as in verify it hasn't already been set) or is it idempotent?

rubenfiszel (dismissed):
minor nits

backend/windmill-worker/src/duckdb_executor.rs
diegoimbert: It's only necessary once but didn't bother because negligible performance impact. Also that's the way we already do it with the S3 connection settings helper which includes the load and install httpfs statements every time

cli/metadata.ts
```diff
@@ -528,6 +528,9 @@ export async function inferSchema(
       { name: "database", typ: { resource: "postgresql" } },
       ...inferedSchema.args,
     ];
+  } else if (language === "duckdb") {
+    const { parse_sql } = await import("./wasm/regex/windmill_parser_wasm.js");
```
pyranota, line 532: This file needs to be updated and checked in into version control (If parser has changed)
AFAIK running cli/build.sh should be enough

backend/windmill-worker/src/duckdb_executor.rs
```diff
@@ -0,0 +1,662 @@
...
+        query = query.replace(&pat, &format!("${}", values.len()));
+    }
+    (query, values)
+}
+
+fn trunc_sig(query: &str) -> &str {
+    let idx = query.rfind("-- $").unwrap_or(query.len());
+    // find next \n starting from idx and return everything after it
+    let idx = query[idx..].find('\n').map(|i| i + idx).unwrap_or(0);
+    &query[idx..]
```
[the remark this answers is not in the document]
diegoimbert: This is the intended behavior

PR #5844, opened 2025-05-30
EE Refactor

diegoimbert:
- **app compiles with every ee substituted**
- **Replace all oss files content**
- **Revert "Replace all oss files content"**
- **delete all ee**
- **hide all _ee files under private flag**
- **hide every oss stuff when private flag set**
- **pub use ***
- **gitignore and substitute script**
- **pub mod for ee needed for ee repo**
- **small mistakes**
- **remove oidc_oss impl**


diegoimbert:
Actually I will spend more time monday fixing the ee repo because the CI has been failing there for a while

PR #5914, opened 2025-06-11
fix(frontend): prevent filtering local storage draft triggers

Guilhem-lm:


Guilhem-lm:
move to https://github.com/windmill-labs/windmill/pull/6015

PR #6012, opened 2025-06-20
fix(frontend): refactor using triggerableByAi using action

Guilhem-lm:
Use action instead of a wrapper component for triggerableByAI to avoid breaking layout.


frontend/src/lib/components/RunForm.svelte
```diff
@@ -214,10 +215,11 @@
...
-			<TriggerableByAI
-				id="run-form-loading"
-				description="Run form is loading, should scan the page until this is gone"
-			/>
+			<!-- Standalone triggerable registration for loading state -->
+			<div style="display: none" use:triggerableByAI={{
+				id: "run-form-loading",
+				description: "Run form is loading, should scan the page until this is gone"
+			}}></div>
 			<h1>Loading...</h1>
```
centdix: we can put the action in the h1 instead of using a div no ?

frontend/src/lib/components/flows/FlowEditor.svelte
```diff
@@ -67,7 +67,11 @@
 	id="flow-editor"
 	class={'h-full overflow-hidden transition-colors duration-[400ms] ease-linear border-t'}
 >
-	<TriggerableByAI id="flow-editor" description="Component to edit a flow" />
+	<!-- Standalone triggerable registration for the flow editor -->
+	<div style="display: none" use:triggerableByAI={{
+		id: "flow-editor", 
+		description: "Component to edit a flow"
+	}}></div>
```
centdix: same we can put it in the root div no ?

frontend/src/lib/components/meltComponents/MenuItem.svelte
```diff
@@ -14,44 +14,54 @@
...
+			id: aiId,
+			description: aiDescription,
+			callback: () => {
+				if (href) {
+					aRef?.click()
+				} else {
+					buttonRef?.click()
+				}
+			}
+		}}
```
centdix: just aRef?.click() we already are in `if href` condition

frontend/src/lib/components/meltComponents/MenuItem.svelte
```diff
@@ -14,44 +14,54 @@
...
```
centdix: just buttonref.click

frontend/src/lib/components/ScriptEditor.svelte
```diff
@@ -417,7 +417,11 @@
 
 <svelte:window onkeydown={onKeyDown} />
 
-<TriggerableByAI id="script-editor" description="Component to edit a script" />
+<!-- Standalone triggerable registration for the script editor -->
+<div style="display: none" use:triggerableByAI={{
+	id: "script-editor", 
+	description: "Component to edit a script"
+}}></div>
```
centdix, line 424: no need for standalone we can put it in the main div no ?

frontend/src/lib/components/RunForm.svelte
Guilhem-lm: done

frontend/src/lib/actions/triggerableByAI.ts
```diff
@@ -0,0 +1,87 @@
...
+			left: ${centerX - 20}px;
+			top: ${centerY - 20}px;
+			z-index: 9999;
+		`
+
+		// Add to body
+		document.body.appendChild(pulse)
+
+		// Remove after animation
+		setTimeout(() => {
```
diegoimbert, line 40: Might cause 'reading undefined' if the parent component gets destroyed before

PR #6029, opened 2025-06-23
fix(frontend): load all flow jobs on page load

Guilhem-lm:


frontend/src/lib/components/stepHistoryLoader.svelte.ts
```diff
@@ -0,0 +1,166 @@
...
+					loadingJobs: false
+				}
+			} finally {
+				// Ensure loading state is always cleaned up
+				this.#stepStates[module.id].loadingJobs = false
+			}
+		})
+
+		// Wait for all loading operations to complete
+		await Promise.all(loadPromises)
```
rubenfiszel, line 160: if you have many modules, you want to have a limit on the paralellization. You can use p-limit for that

Guilhem-lm: Done

PR #6180, opened 2025-07-14
feat: add workspace prefix option for custom app

dieriba:
> [!IMPORTANT]
> Add `APP_WORKSPACED_ROUTE_SETTING` to enable workspace-specific custom paths for apps, with backend and frontend support.
> 
>   - **Behavior**:
>     - Adds `APP_WORKSPACED_ROUTE_SETTING` to allow apps to have workspace-specific custom paths.
>     - Updates `create_app_internal()` and `update_app_internal()` in `apps.rs` to check `APP_WORKSPACED_ROUTE` for custom path conflicts.
>     - Adds validation in `set_global_setting_internal()` in `settings.rs` to prevent disabling the setting if duplicate custom paths exist.
>   - **Frontend**:
>     - Updates `AppEditorHeader.svelte` to load and use `app_workspaced_route` setting for custom path validation.
>     - Adds `App workspace prefix` setting to `instanceSettings.ts`.
>   - **Misc**:
>     - Adds `APP_WORKSPACED_ROUTE` to `apps.rs` and `APP_WORKSPACED_ROUTE_SETTING` to `global_settings.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 3e34c65ddc29b3d2810d70844d81486946a5e50e. You can [customize](https://app.ellipsis.dev/windmill-labs/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


backend/windmill-api/src/apps.rs
```diff
@@ -638,18 +660,66 @@ async fn update_app_history(
...
+        .unwrap_or(false)
+    };
+
+    Ok(exists)
+}
+
 async fn custom_path_exists(
     Extension(db): Extension<DB>,
     Path((w_id, custom_path)): Path<(String, String)>,
+    Json(payload): Json<CustomPathExistsRequest>,
```
hugocasa: you could use a query arg

backend/windmill-api/src/apps.rs
```diff
@@ -1315,16 +1382,45 @@ async fn update_app_internal<'a>(
...
+                                AND NOT (path = $2 AND workspace_id = $3)
+                        )
+                        "#,
+                        full_custom_path.as_ref(),
+                        path,
+                        w_id,
+                    )
+                };
 
+                let exists = exists_query.fetch_one(&mut *tx).await?.unwrap_or(false);
```
hugocasa: i see that we were already using tx before, but i think we should use db, otherwise we won't check every case. But at the same time it would be best to have it in the same transaction. Let's discuss this.

PR #6245, opened 2025-07-21
feat(aichat): improve autocomplete with ata and editor diagnostic

centdix:


frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
```diff
@@ -84,6 +108,39 @@ export class Autocompletor {
...
+								return {
+									items: []
+								}
+							}
+
+							const toEol = new Range(
+								position.lineNumber,
+								position.column,
+								position.lineNumber,
+								model.getLineMaxColumn(position.lineNumber)
```
hugocasa: isn't it the same range as for getting the code just above?

frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
```diff
@@ -94,7 +151,20 @@ export class Autocompletor {
...
-									range
+									range,
+									additionalTextEdits: isWholeLine
+										? [
+												{
+													range: {
+														startLineNumber: position.lineNumber,
+														startColumn: position.column,
+														endLineNumber: position.lineNumber,
+														endColumn: model.getLineMaxColumn(position.lineNumber)
```
hugocasa: what does that do?

hugocasa (commented):
could you try combine isWholeLine/isMultiline or adjust them to make the code clearer?

frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
```diff
@@ -84,6 +108,39 @@ export class Autocompletor {
 							endColumn: position.column
 						}
 
+						// if completion takes whole line, delete the rest of the line after the suggestion
+						const isWholeLine = result.completion.indexOf('\n') !== -1
```
hugocasa: also why are you using result.completion instead of completion?

frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
centdix: it is, I can clean that

frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
centdix: it erases the rest of the line in case completition isWholeLine

frontend/src/lib/components/copilot/autocomplete/Autocompletor.ts
centdix: because completion goes through the filtering logic for the multiline logic (so removes the last \n), whereas result.completion is the raw one
that's also why I made it a separate variable, but I agree that it is better if we can combine with multiline

PR #6424, opened 2025-08-20
feat(aichat): allow adding contexts to flow mode

centdix:


centdix:
/ai review this PR, goal is to have the same features the script mode has and implement them in the flow mode (handling context). look for forgotten things and duplicate code that can be cleaned

centdix:
/ai implement these from the review:
- export const ContextIconMap = {
    code: Code,
    error: TriangleAlert, 
    db: Database,
    diff: Diff,
    code_piece: Code,
    // ❌ Missing: flow_module: SomeIcon
}
just add a comment saying flow_module type is handled with FlowModuleIcon

- export function createResourceTypeTool<T>(langGetter: (helpers: T) => string): Tool<T>

- export function createSearchHubScriptsTool<T>(withContent: boolean): Tool<T>. make sure withContent is only true for script mode, the logic should be exactly the same as before, just factorize.

Make sure the factorization tasks do not modify any logic, older or newer.

Do one commit per fix

hugocasa (approved):

PR #6461, opened 2025-08-25
feat: condiontal retry workflow

dieriba:
> [!IMPORTANT]
> Add conditional retry mechanism to workflows with `RetryIf` condition support in backend and frontend.
> 
>   - **Backend**:
>     - Introduced `RetryIf` condition in `Retry` schema in `openflow.openapi.yaml`.
>     - Updated `evaluate_retry()` in `worker_flow.rs` to handle `RetryIf` conditions.
>     - Added `Step` enum and `get_flow_module_at_step()` in `flows.rs` for step handling.
>   - **Frontend**:
>     - Updated `FlowRetries.svelte` to include UI for configuring `RetryIf` conditions.
>     - Modified `FlowModuleComponent.svelte` to bind `flowModuleRetry` and handle retry conditions.
>   - **Misc**:
>     - Added `async-once-cell` dependency in `Cargo.toml` for async cell handling.
>     - Renamed JSON query files in `.sqlx/` to reflect query changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=windmill-labs%2Fwindmill&utm_source=github&utm_medium=referral)<sup> for 6186e2e095c7da43222a5c10069760a0f613bf54. You can [customize](https://app.ellipsis.dev/windmill-labs/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


backend/windmill-worker/src/worker_flow.rs
```diff
@@ -1154,6 +1164,40 @@ pub async fn update_flow_status_after_job_completion_internal(
...
+                Step::Step(i) => flow_value
+                    .modules
+                    .get(i)
+                    .as_ref()
+                    .and_then(|m| m.retry.as_ref()),
+                Step::FailureStep => flow_value
+                    .failure_module
+                    .as_ref()
+                    .and_then(|m| m.retry.as_ref()),
+            }
```
rubenfiszel: nit, as impl on Step rather than inlined here?

backend/windmill-worker/src/worker_flow.rs
```diff
@@ -1536,11 +1557,63 @@ async fn compute_skip_branchall_failure<'c>(
...
+
+                let fetched_args = sqlx::query_scalar!(
+                    "SELECT args AS \"args: Json<HashMap<String, Box<RawValue>>>\" FROM v2_job WHERE id = $1",
+                    flow_id
+                )
+                .fetch_one(db)
+                .await
+                .map_err(|e| {
+                    Error::internal_err(format!("retrieval of args from state for retry condition: {e:#}"))
+                })?;
```
rubenfiszel: are you sure the flow args are not already available and fetched before and can't be reused?

dieriba: no as they were fetched only fron a local scope (within an if block), but now it's ok it is only fetched once

backend/windmill-worker/src/worker_flow.rs
dieriba: I think it would make more sense to impl on FlowValue then from there retrieve whatever we want from the retrieved module

backend/windmill-worker/src/worker_flow.rs
```diff
@@ -344,14 +345,33 @@ pub async fn update_flow_status_after_job_completion_internal(
...
+                r#"
+            SELECT args AS "args: Json<HashMap<String, Box<RawValue>>>"
+            FROM v2_job
+            WHERE id = $1
+            "#,
+                flow
+            )
+            .fetch_one(db)
+            .await;
+            args
```
rubenfiszel, line 359: let args = { ... }
args
=> { ... }

backend/windmill-worker/src/worker_flow.rs
```diff
@@ -344,14 +345,33 @@ pub async fn update_flow_status_after_job_completion_internal(
...
+        }));
+
+        let from_result_to_args =
+            |args: &Result<Option<Json<HashMap<String, Box<RawValue>>>>, sqlx::Error>| {
+                let args = args.as_ref().map_err(|e| {
+                    Error::internal_err(format!("retrieval of args from state: {e:#}"))
+                })?;
+
+                Ok::<_, Error>(args.clone())
+            };
```
rubenfiszel, line 369: why not do that in the lazy cell directly

dieriba: because lazy cell gives you gives you an immutable ref for the Ok varian and the error variant and as the error return type need to be owned I would need to map the error to something like anyhow, so it is just better to put it in the closure as I would need to do map the error anyway

PR #6472, opened 2025-08-26
feat: refactor trigger crud

dieriba:


backend/windmill-api/src/triggers/handler.rs
```diff
@@ -0,0 +1,719 @@
...
+) -> Result<(StatusCode, String)> {
+    check_scopes(&authed, || {
+        format!(
+            "{}:write:{}",
+            T::scope_domain_name(),
+            &new_trigger.base.path
+        )
+    })?;
+
+    if *CLOUD_HOSTED {
```
hugocasa: some are allowed

backend/windmill-api/src/triggers/http/handler.rs
```diff
@@ -0,0 +1,938 @@
...
+
+    async fn create_trigger(
+        &self,
+        _db: &DB,
+        tx: &mut PgConnection,
+        authed: &ApiAuthed,
+        w_id: &str,
+        trigger: TriggerData<Self::TriggerConfigRequest>,
+    ) -> Result<()> {
+        insert_new_trigger_into_db(authed, tx, w_id, &trigger).await?;
```
hugocasa: if i'm not mistaken for create http trigger, you do not check if it already exists

backend/windmill-api/src/triggers/handler.rs
```diff
@@ -0,0 +1,719 @@
...
+    ) -> Result<bool> {
+        if !Self::SUPPORTS_SERVER_STATE {
+            return Err(anyhow::anyhow!(
+                "Enable/disable not supported for this trigger type".to_string(),
+            )
+            .into());
+        }
+
+        let updated = sqlx::query(&format!(
+            "UPDATE {} SET enabled = $3 WHERE workspace_id = $1 AND path = $2",
```
hugocasa: what about server_id, last_server_ping and error?

backend/windmill-api/src/triggers/handler.rs
```diff
@@ -0,0 +1,719 @@
...
+        .await?;
+
+    if !updated {
+        return Err(error::Error::NotFound(format!(
+            "Trigger not found at path: {}",
+            path
+        )));
+    }
+
+    tx.commit().await?;
```
hugocasa, line 599: what about handle_deployment_metadata?

backend/windmill-api/src/triggers/handler.rs
```diff
@@ -0,0 +1,719 @@
...
+
+        sqlx::query_as(&sql)
+            .bind(workspace_id)
+            .bind(path)
+            .fetch_optional(&mut *tx)
+            .await?
+            .ok_or_else(|| Error::NotFound(format!("Trigger not found at path: {}", path)))
+    }
+
+    async fn exists(&self, tx: &mut PgConnection, workspace_id: &str, path: &str) -> Result<bool> {
```
hugocasa: you should use db not user tx, you might not be allowed to see all triggers in the workspace

backend/windmill-api/src/triggers/nats/mod.rs
hugocasa, line 1: ideally it should be in ee

backend/windmill-api/src/triggers/websocket/handler.rs
```diff
@@ -0,0 +1,292 @@
...
+            ));
+        }
+
+        if let Some(args) = &new.url_runnable_args {
+            if !args.is_object() {
+                return Err(Error::BadRequest(
+                    "url_runnable_args must be an object".to_string(),
+                ));
+            }
+        }
```
hugocasa: you could make it a function so you can reuse it in validate_edit

backend/windmill-api/src/triggers/mod.rs
```diff
@@ -187,14 +187,15 @@ pub(crate) async fn get_triggers_count_internal(
     .await?
     .unwrap_or(0);
 
+    #[allow(unused)]
+    let mut tx = db.begin().await?;
```
hugocasa: it's less overhead on the db to use individual transactions rather than a big transaction (if not needed)
no need to explicitely rollback the tx though, it is automatically rollbacked on drop

backend/windmill-api/src/triggers/handler.rs
```diff
@@ -0,0 +1,744 @@
...
+            r#"
+            UPDATE 
+                {} 
+            SET 
+                enabled = $1,
+                email = $2,
+                edited_by = $3,
+                edited_at = now(),
+                server_id = NULL,
+                error = NULL
```
hugocasa: missing last_server_ping

hugocasa (approved):

PR #6515, opened 2025-09-02
feat(flow): add structured output option for ai agent step

centdix:


backend/windmill-worker/src/ai_executor.rs
```diff
@@ -648,6 +835,10 @@ async fn run_agent(
                     tools: tool_defs.as_ref(),
                     temperature: args.temperature,
                     max_completion_tokens: args.max_completion_tokens,
+                    response_format: match is_anthropic || !has_output_properties {
+                        true => None,
+                        false => response_format.clone(),
+                    },
```
hugocasa, line 842: if else more readable

backend/windmill-worker/src/ai_executor.rs
```diff
@@ -279,6 +341,79 @@ impl OpenAPISchema {
...
+                .iter()
+                .map(|schema| Box::new(schema.as_ref().clone().make_strict()))
+                .collect();
+        }
+
+        self
+    }
+
+    /// Makes this property nullable by converting its type to a union with null
+    fn make_nullable(mut self) -> Self {
```
hugocasa, line 396: i think the correct way to do this is to add a nullable: true property (https://swagger.io/docs/specification/v3_0/data-models/data-types/#null)

backend/windmill-worker/src/ai_executor.rs
```diff
@@ -792,8 +1005,17 @@ async fn run_agent(
...
 
+    // Parse content as JSON, fallback to string if it fails
+    let output_value = match content {
+        Some(content_str) => match serde_json::from_str::<Value>(&content_str) {
+            Ok(parsed) => parsed,
+            Err(_) => Value::String(content_str),
+        },
+        None => Value::String(String::new()),
+    };
+
```
hugocasa, line 1019: you should use `Box<RawValue>` when possible to avoid unneeded deserialization

PR #6542, opened 2025-09-05
feat(backend): accept wildcard for list_paths_from_workspace_runnable + bulk delete endpoints for scripts/variables

hugocasa:


backend/windmill-api/openapi.yaml
```diff
@@ -5806,6 +5902,10 @@ paths:
         - $ref: "#/components/parameters/WorkspaceId"
         - $ref: "#/components/parameters/RunnableKind"
         - $ref: "#/components/parameters/ScriptPath"
+        - name : match_path_prefix
```
rubenfiszel: looks like we use path_start everywhere else. Feels inconsistent

PR #6739, opened 2025-10-02
[WIP] feat: add dynamic skip for schedules

alpetric:


backend/windmill-api/src/triggers/trigger_helpers.rs
```diff
@@ -807,13 +807,18 @@ async fn trigger_script_with_retry_and_error_handler(
...
+            hash: Some(hash),
+            flow_version: None,
             args: HashMap::from(&push_args),
             retry,
             error_handler_path,
             error_handler_args,
+            skip_handler_path: None,
+            skip_handler_args: None,
+            skip_handler_stop_condition: None,
+            skip_handler_stop_message: None,
```
rubenfiszel: maybe we can make it a subobject for cleanliness. I know we didn't do it above but it's 4 args

backend/windmill-api/src/schedule.rs
```diff
@@ -128,6 +129,48 @@ fn to_json_raw_opt(
...
+    .unwrap_or(false);
+
+    if script_exists {
+        return Ok(());
+    }
+
+    // Check for flow
+    let flow_exists = sqlx::query_scalar!(
+        "SELECT EXISTS(SELECT 1 FROM flow
+         WHERE workspace_id = $1 AND path = $2 AND archived = false)",
```
hugocasa: you can probably combine both queries into one

alpetric: actually, I think we decided to only support scrips not flows so I can remove the flow check

PR #6753, opened 2025-10-03
fix: init git repo preview save

alpetric:


alpetric:
@dieriba thanks for the report!

PR #6873, opened 2025-10-20
feat(ai agent): handle inputTransforms for tools arguments

centdix:


backend/windmill-worker/src/ai/utils.rs
```diff
@@ -463,3 +533,70 @@ pub async fn execute_mcp_tool(
 
     Ok(result)
 }
+
+/// Evaluate both static and JavaScript input transforms for AI tools.
+pub async fn evaluate_input_transforms(
```
hugocasa: why not use transform_input?

backend/windmill-worker/src/ai_executor.rs
```diff
@@ -393,6 +402,33 @@ pub async fn run_agent(
         }
     }
 
+    // Extract previous step result if we have flow_status
+    let previous_result = {
+        if let Some(ref flow_status) = flow_context.flow_status {
+            get_previous_job_result(db, &job.workspace_id, flow_status)
```
hugocasa: you should only fetch it if an input transform contains previous_result

PR #6876, opened 2025-10-21
feat: add onboarding form for cloud first timers

tristantr:


backend/windmill-api/src/users_oss.rs
```diff
@@ -53,3 +59,26 @@ pub fn send_email_if_possible(_subject: &str, _content: &str, _to: &str) {
         "send_email_if_possible is not implemented in Windmill's Open Source repository"
     );
 }
+
+#[cfg(not(feature = "private"))]
+#[derive(Deserialize, Debug)]
+#[allow(dead_code)]
+pub struct OnboardingData {
+    pub email: String,
```
diegoimbert: user info should not come from the frontend to avoid impersonation
You can get the email & username from ApiAuthed instead

frontend/src/lib/components/Login.svelte
```diff
@@ -106,10 +106,30 @@
 		}
 
 		if (firstTime) {
```
diegoimbert, line 108: Just checked and this value comes from users.rs (is_first_time_setup()).
The function returns true only if the instance has a single user admin@windmill.dev:changeme

so this should never happen on cloud

frontend/src/lib/components/Login.svelte
```diff
@@ -106,10 +106,30 @@
...
+				goto('/user/first-time')
+			}
 			return
 		}
 
+		// Check if this is a first-time user (individual user onboarding)
+		// Only show onboarding for cloud-hosted instances
+		if (isCloudHosted()) {
+			try {
+				const globalUserInfo = await UserService.globalWhoami()
```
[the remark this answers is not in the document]
diegoimbert: not a big problem

frontend/src/lib/components/Login.svelte
```diff
@@ -182,10 +202,14 @@
 
 			showPassword = (logins.length == 0 && !saml) || (email != undefined && email.length > 0)
 		} catch (e) {
+			// OAuth endpoint not available (OSS version or not configured)
 			logins = []
 			saml = undefined
 			showPassword = true
-			console.error('Could not load logins', e)
+			// Only log if it's not a 404
+			if (e?.status !== 404) {
```
diegoimbert: No reason to only log 404s, avoid silent errors at all cost

frontend/src/routes/(root)/(logged)/user/(user)/onboarding/+page.svelte
```diff
@@ -0,0 +1,167 @@
...
+		{ id: 'word_of_mouth', label: 'Word of mouth', icon: Users },
+		{ id: 'in_my_company', label: 'Used in my company', icon: Building2 },
+		{ id: 'event', label: 'Event', icon: Calendar },
+		{ id: 'other', label: 'Other', icon: HelpCircle }
+	]
+
+	function selectSource(sourceId: string) {
+		selectedSource = sourceId
+		// Auto-advance to next step for all selections
+		currentStep = 2
```
diegoimbert: currentStep = Math.min(totalStepCount, currentStep + 1)

But the core reason is :
Numbers have no semantic meaning in this case, do not rely on the fact that "1 = source selection"
Maybe one day the steps will change and this is prone to future bugs

frontend/src/routes/(root)/(logged)/user/(user)/onboarding/+page.svelte
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	function goToPreviousStep() {
+		if (currentStep === 2) {
```
diegoimbert: same here, this relies on the fact that total steps count = 2 (avoid magic numbers)

backend/windmill-api/src/users_oss.rs
diegoimbert: ideally, only use_case and touch_point comes from the front (i.e only what the user actually inputted)
and everything else comes from the backend which cannot be tampered with

frontend/src/routes/user/login_callback/[client_name]/+page.svelte
```diff
@@ -43,6 +44,20 @@
 				return
 			}
 
+			// Check if this is a first-time user (individual user onboarding)
+			// Only show onboarding for cloud-hosted instances
```
Comment (hugocasa) [inline frontend/src/routes/user/login_callback/[client_name]/+page.svelte:57]: i think would be cleaner to do this check in loadUser in the layout

PR #6888, opened 2025-10-21
fix: also auto add/del igroup members to workspaces where configured

alpetric:


backend/windmill-api/src/groups.rs
```diff
@@ -781,8 +795,16 @@ async fn remove_user_igroup(
         Some([("email", email.as_str())].into()),
     )
     .await?;
+
+    // Remove user from workspaces where they were added via this instance group
+    #[cfg(feature = "private")]
```
rubenfiszel: private and ee no?

backend/windmill-api/src/groups.rs
```diff
@@ -633,6 +633,20 @@ async fn add_user_igroup(
         Some([("email", email.as_str())].into()),
     )
     .await?;
+
+    // Sync user to workspaces configured with this instance group
+    #[cfg(feature = "private")]
```
rubenfiszel: private and ee

PR #6889, opened 2025-10-22
fix(parser): Handle CRLF line endings in bash and PowerShell parsers

### LINKED ISSUE #6867: bug:wmill script generate-metadata with bash not working
### Describe the bug

Running wmill script generate meta-data does not update the properties correctly. They are always blank. Further, it constantly says there is a change when i go to sync push, saying i need to generate-metadata even though I literally just did it.

### To reproduce

setup local wmill cli workspace

Create a new bash script. (.sh) Doesn't matter if you do it locally or create in windmill ui, then pull it locally.

Make sure said script has some inputs. if you did in windmill UI then verify the script.yaml file has the properties in there.

Run wmill script generate-metadata. properties will be blank. 

# shellcheck shell=bash
echo "Hello world"
test_var="$1"

summary: ''
description: ''
lock: ''
kind: script
schema:
  $schema: 'https://json-schema.org/draft/2020-12/schema'
  type: object
  properties: {}
  required: []

### Expected behavior

properties to be specified in script yaml file.

### Screenshots

_No response_

### Browser information

_No response_

### Application version

1.563.3

### Additional Context

windows powershell,

also tried in wsl and have same issue.

vivekchavan14:
## Description

Fixes #6867

### Problem
Bash scripts with CRLF line endings (Windows-style) failed to generate metadata properly via `wmill script generate-metadata`. Properties were blank and the CLI constantly detected files as changed, even after running generate-metadata.

### Root Cause
The regex patterns in the bash and PowerShell parsers only matched LF (`\n`) line endings. When files had CRLF (`\r\n`) endings, the patterns failed to match, resulting in empty metadata properties.

### Solution
Updated regex patterns to optionally match carriage return (`\r`) before line endings:
- **Bash parser**: Added `\r?` before the `$` anchor in `RE_BASH`
- **PowerShell parser**: Added `\r?` at the end of `RE_POWERSHELL_ARGS`

### Changes
- Modified `backend/parsers/windmill-parser-bash/src/lib.rs`
- Updated 2 regex patterns (lines 48 and 51)
- Added comprehensive test case `test_parse_bash_sig_with_crlf()` to verify CRLF handling

### Testing
```
running 3 tests
test tests::test_parse_bash_sig ... ok
test tests::test_parse_bash_sig_with_crlf ... ok
test tests::test_parse_bash_sig_with_crlf ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured
```

- All existing tests pass (backward compatible with LF line endings)
- New test specifically validates CRLF handling
- Python and other parsers unaffected (they use AST parsers that normalize line endings)

### Impact
- Fixes metadata generation for bash scripts created on Windows
- Resolves false "all files changed" detection in CLI
- No breaking changes - works with both CRLF and LF line endings

### DIFF
--- a/backend/parsers/windmill-parser-bash/src/lib.rs
+++ b/backend/parsers/windmill-parser-bash/src/lib.rs
@@ -45,10 +45,10 @@ pub fn parse_powershell_sig(code: &str) -> anyhow::Result<MainArgSignature> {
 }
 
 lazy_static::lazy_static! {
-    static ref RE_BASH: Regex = Regex::new(r#"(?m)^(\w+)="\$(?:(\d+)|\{(\d+)\}|\{(\d+):-(.*)\})"(?:[\t ]*)?(?:#.*)?$"#).unwrap();
+    static ref RE_BASH: Regex = Regex::new(r#"(?m)^(\w+)="\$(?:(\d+)|\{(\d+)\}|\{(\d+):-(.*)\})"(?:[\t ]*)?(?:#.*)?\r?$"#).unwrap();
 
     pub static ref RE_POWERSHELL_PARAM: Regex = Regex::new(r#"(?m)param[\t ]*\(([^)]*)\)"#).unwrap();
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Thank you!

PR #6948, opened 2025-10-27
feat(flow): Add graph diff visualizer

centdix:


frontend/src/lib/components/DiffDrawer.svelte
```diff
@@ -30,14 +31,20 @@
...
 
 	interface Props {
 		restoreDeployed?: () => Promise<void>
 		restoreDraft?: () => Promise<void>
+		isFlow?: boolean
 	}
 
-	let { restoreDeployed = async () => {}, restoreDraft = async () => {} }: Props = $props()
+	let {
+		restoreDeployed = async () => {},
```
diegoimbert: (nit)
Instanciating an empty function isn't necessary, just leave it undefined and call it with the functionName.?() operator

frontend/src/lib/components/FlowGraphDiffViewer.svelte
```diff
@@ -0,0 +1,193 @@
...
+	const SIDE_BY_SIDE_MIN_WIDTH = 700
+
+	interface Props {
+		beforeYaml: string
+		afterYaml: string
+	}
+
+	let { beforeYaml, afterYaml }: Props = $props()
+
+	let parseError = $state<string | undefined>(undefined)
```
diegoimbert, line 24: Never set

PR #6976, opened 2025-10-28
feat: allow http trigger to be disabled

dieriba:


backend/windmill-api/src/triggers/http/handler.rs
```diff
@@ -508,6 +511,7 @@ impl TriggerCrud for HttpTrigger {
                 trigger.base.script_path,
                 trigger.base.path,
                 trigger.base.is_flow,
+                trigger.base.enabled,
```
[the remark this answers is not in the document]
hugocasa: +1

backend/windmill-api/src/triggers/http/handler.rs
```diff
@@ -816,6 +857,14 @@ async fn route_job(
     .await
     .map_err(|e| e.into_response())?;
 
+    if !trigger.enabled.unwrap_or(false) {
```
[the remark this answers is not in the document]
hugocasa: +1

backend/windmill-api/src/triggers/http/mod.rs
```diff
@@ -244,6 +245,7 @@ pub async fn refresh_routers(db: &DB) -> Result<(bool, RwLockReadGuard<'_, Route
                         script_path,
                         is_flow,
                         route_path,
+                        enabled,
```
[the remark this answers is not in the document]
hugocasa: you should filter out disabled routes

backend/windmill-api/src/triggers/listener.rs
```diff
@@ -69,12 +69,12 @@ pub trait Listener: TriggerCrud + TriggerJobArgs {
             "email",
             "edited_at",
             "extra_perms",
+            "enabled",
+            "error_handler_path",
+            "error_handler_args",
+            "retry",
```
hugocasa, line 75: is it safe because this fn is only called when supports_server_state is set to true?

frontend/src/lib/components/triggers/http/RouteEditorInner.svelte
```diff
@@ -268,6 +270,7 @@
 		wrap_body = cfg?.wrap_body ?? false
 		raw_string = cfg?.raw_string ?? false
 		summary = cfg?.summary ?? ''
+		enabled = cfg?.enabled ?? false
```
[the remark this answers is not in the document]
hugocasa: that will make existing triggers shown as disabled

hugocasa (commented):
- you either have to set default to true in the migration or handle enabled is null correctly everywhere

backend/windmill-api/src/triggers/http/mod.rs
hugocasa: you don't need enabled anymore as you know it will be enabled

backend/windmill-api/src/triggers/listener.rs
hugocasa: is it safe because this fn is only called when supports_server_state is set to true?

backend/windmill-api/src/triggers/http/handler.rs
```diff
@@ -586,6 +590,43 @@ impl TriggerCrud for HttpTrigger {
         Ok(())
     }
 
+    async fn set_enabled(
```
hugocasa: isn't already handled by the trait?

dieriba: it is but for http we need to increase_trigger_version otherwise the cache we'd have outdated http trigger

backend/windmill-api/src/triggers/listener.rs
dieriba: yup its because all listener object have supports_server_state to true thats why its safe

PR #7021, opened 2025-10-31
add zoho to auth settings list

gpeppers:
During the last patch for zoho oauth, Zoho was not added to the list of oauth providers on the instance settings page. 
This adds it into that list to finalize the zoho support. 

Thanks


rubenfiszel:
@alpetric could you give this a test please

PR #7062, opened 2025-11-05
feat: webhook by flow version

dieriba:


backend/windmill-api/src/jobs.rs
```diff
@@ -5451,100 +5621,134 @@ pub async fn run_wait_result_flow_by_path_internal(
...
+    Extension(user_db): Extension<UserDB>,
+    Extension(db): Extension<DB>,
+    Path((w_id, version)): Path<(String, i64)>,
+    Query(run_query): Query<RunJobQuery>,
+    args: RawWebhookArgs,
+) -> error::Result<Response> {
+    #[cfg(feature = "enterprise")]
+    check_license_key_valid().await?;
+
+    let flow_path = sqlx::query_scalar!(
```
hugocasa, line 5821: minor improvement would be to add a similar function as get_flow_version_info_from_version which doesn't require path but returns it for check_scopes so you don't need to fetch the path separately

hugocasa: however make sure to check access when using the cache (search db_authed in the codebase for examples)

PR #7106, opened 2025-11-10
feat(security): unshare pid of worker job process

alpetric:


backend/windmill-worker/src/go_executor.rs
```diff
@@ -371,10 +371,7 @@ func Run(req Req) (interface{{}}, error){{
         #[cfg(windows)]
         let compiled_executable_name = format!("{}/main.exe", job_dir);
 
-        #[cfg(unix)]
-        let mut run_go = Command::new(&compiled_executable_name);
-        #[cfg(windows)]
-        let mut run_go = Command::new(&compiled_executable_name);
```
alpetric, line 377: not sure if this was intentional and there for a reason @rubenfiszel ?

alpetric: was me 🤣

PR #7137, opened 2025-11-14
fix: use proper TLS connector for DuckLake instance catalog setup

knowsuchagency:
## Problem

The `setup_ducklake_catalog_db_inner` function in `backend/windmill-api/src/settings.rs` was using `tokio_postgres::NoTls` even when `sslmode=require` was specified in the connection string. This caused TLS handshake failures when connecting to PostgreSQL databases that require SSL, such as AWS RDS with `sslmode=require`.

Error message:
```
error performing TLS handshake: no TLS implementation configured
```

## Solution

This PR adds conditional TLS connector logic similar to the pattern used in `pg_executor.rs`:

- Uses `native_tls::TlsConnector` with `MakeTlsConnector` when sslmode requires SSL (require, verify-ca, verify-full)
- Accepts invalid certs/hostnames for compatibility with managed database services  
- Falls back to `NoTls` for non-SSL connections

## Testing

Tested with AWS RDS PostgreSQL configured with `sslmode=require` in DATABASE_URL.

## Related

- Complements existing sslmode support added in PR #7110 for regular DuckLake operations
- Fixes DuckLake instance catalog setup specifically (step 5 "Connect to the ducklake_catalog database")


knowsuchagency:
## Additional Fixes

After the initial implementation, I encountered and resolved two compilation issues:

### 1. Missing Dependency (commit ddb703e2a8)
The TLS connector required `postgres-native-tls` to be explicitly added to `windmill-api/Cargo.toml`. While it was defined in the workspace, it wasn't referenced in the crate dependencies.

**Fix**: Added `postgres-native-tls.workspace = true` to line 97 of `backend/windmill-api/Cargo.toml`

### 2. Type Mismatch (commit 731fdf8092)
Rust's type system doesn't allow mixing `Connection<_, TlsStream>` and `Connection<_, NoTlsStream>` in the same variable, even though they share the same `Client` interface.

**Error**:
```
expected tuple (tokio_postgres::Client, tokio_postgres::Connection<_, postgres_native_tls::TlsStream<Socket>>)
   found tuple (tokio_postgres::Client, tokio_postgres::Connection<_, NoTlsStream>)
```

**Fix**: Refactored to separate the TLS and non-TLS paths completely, with each branch handling:
- Connection establishment
- Query execution (GRANT permissions)
- Cleanup (drop client, join connection)

This follows the same pattern used in `pg_executor.rs` for handling similar connection type scenarios.

backend/windmill-api/src/settings.rs
```diff
@@ -701,40 +701,89 @@ async fn setup_ducklake_catalog_db_inner(
         sslmode = ssl_mode
     );
 
-    let (client, connection) = tokio::time::timeout(
```
rubenfiszel, line 704: @diegoimbert I think we can take over the PR, it's mostly to make it easier for us from @knowsuchagency

PR #7155, opened 2025-11-17
feat(ai): handle aws bedrock as provider

centdix:


backend/windmill-api/src/ai.rs
```diff
@@ -446,6 +1074,18 @@ async fn proxy(
         }
     };
 
+    // Extract model and streaming flag for Bedrock transformation (only for POST requests)
+    let (model_for_transform, is_streaming) =
+        if matches!(provider, AIProvider::AWSBedrock) && method == Method::POST {
+            let parsed: serde_json::Value = serde_json::from_slice(&body)
```
hugocasa: use a struct instead of value

backend/windmill-worker/src/ai_executor.rs
```diff
@@ -551,253 +553,289 @@ pub async fn run_agent(
...
-            .timeout(timeout)
-            .header("Content-Type", "application/json");
-
-        // Apply authentication headers
-        for (header_name, header_value) in &auth_headers {
-            request = request.header(*header_name, header_value.clone());
-        }
+        // Special handling for AWS Bedrock using the official SDK
+        let parsed = if args.provider.kind == AIProvider::AWSBedrock {
+            if region.is_none() {
```
hugocasa: a nice code pattern is:
```rust
let Some(region) = region else {
    return Err(Error::internal_err(
        "AWS Bedrock region is required".to_string(),
    ));
}
```

Issue #7253, opened 2025-11-29
bug: variable/resource not parsed when write in nested generate input box and can't concat to


androidfans:
### Describe the bug

1. write variable in string nested in an array, the script received `$var:f/Narration/jina_api_key` but not real value
2. not found a way to concat variable/resource with a string in genrated UI

### To reproduce

```javascrript
interface Header {
  value: string;
}

export async function main(
  key: string,
  header: Header,
  headers: Header[],
) {
  console.log(`key: ${url}`);
  console.log(`Header: ${header.value}`);
  console.log(`HeaderArray: ${headers[0].value}`)
}
```

write variable in generated UI

### Expected behavior

script received actual value

### Screenshots

<img width="804" height="767" alt="Image" src="https://github.com/user-attachments/assets/8f02f1ff-d13b-43d7-9b7f-40c29296f4de" />

PR #7254, opened 2025-11-29
fix: support variable parsing in arrays for job arguments

androidfans:
Previously, variables like `$var:f/path/to/variable` were not being
  resolved when they appeared inside arrays in job arguments. This was
  because the `transform_json_value` function in `common.rs` only
  handled recursive transformation for `Value::Object`, but was missing
  the case for `Value::Array`.

  This commit adds support for recursively transforming array elements,
  allowing variables to be properly resolved when used within arrays.

  Example:
  - Before: `["$var:f/Narration/jina_api_key"]` would remain as literal string
  - After: `["jina_this_is_real_jian_key_value"]` (variable is properly resolved)

  Note: String concatenation with variables (e.g., "Bearer: $var:...")
  is still not supported as it requires a more complex implementation
  with regex-based pattern matching and substitution.

  Fixes issue #7253 where variables in array parameters were not being parsed.

### DIFF
--- a/backend/windmill-worker/src/common.rs
+++ b/backend/windmill-worker/src/common.rs
@@ -301,6 +301,13 @@ pub async fn transform_json_value(
             }
             Ok(Value::Object(m))
         }
+        Value::Array(arr) => {
+            let mut result = Vec::with_capacity(arr.len());
+            for item in arr {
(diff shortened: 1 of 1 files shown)


androidfans:
This issue found when i am trying to migrate my script from N8N, to align function with n8n

PR #7280, opened 2025-12-02
fix(schema): preserve user-defined JSON schema for Python list[dict] parameters

### LINKED ISSUE #7209: bug: Always lost json schema define for generated ui for python
### Describe the bug

https://github.com/user-attachments/assets/8f9194cb-be14-4723-897c-4f9ac90c3439

i am trying to add a self-defined multi-pass conversation LLM call, but python script can't recognize type inner list like ts. 
then i change to use dict, and define field by json schema, but it always lost, can't save

### To reproduce

1. create a python script
2. using param `messages: list[dict]`
3. go to generated ui
4. add field role and message for dict inner messages
5. click `test`, works well
6. click draft or deploy, error toast `Cannot convert undefined or null to object`, the schema define has lost.

### Expected behavior

1. windmill auto recognize field from type in list is best
2. it 1 can't achieve, windmill can save defined schema successfully

### Screenshots

has provided video 

### Browser information

_No response_

### Application version

_No response_

### Additional Context

_No response_

devdattatalele:
## Problem

When defining Python scripts with `list[dict]` or untyped array parameters, user-defined JSON schema properties are lost upon saving as draft or deploying. This prevents users from creating proper UI forms for complex data structures like message arrays for LLM calls.

Resolves #7209

According to comments made following changes

**Frontend:**
- `cli/windmill-utils-internal/src/parse/parse-schema.ts`:
  - Lines 209-222: Preserve ALL items fields when parser cannot infer structure
  - Line 210: Preserve items.type instead of hardcoding "object"
  - Lines 225-230: Preserve items for untyped lists
  - Line 234: Use nullish coalescing (??) for correct type preservation

## Testing

**Test Case 1**: `list[dict]` with properties and required fields
```python
def main(messages: list[dict]):
    return messages
```
- Define properties and required fields via UI
- Save draft → All fields preserved ✅

**Test Case 2**: Untyped array parameter
```python
def main(messages):
    return messages
```
- Define `type: "array"` with properties via UI
- Save draft → Type remains "array" ✅

**Test Case 3**: Backward compatibility
- Existing schemas without manual properties → No change ✅

## Breaking Changes

None. This change only preserves user-defined fields when they exist.


devdattatalele:
Fixed the inconsistency. The untyped list branch now preserves all user-defined fields (required, additionalProperties, enum, resourceType, contentEncoding, description) just like the record[] branch, instead of only preserving properties.

This ensures complete schema preservation regardless of whether the list is typed (`list[dict]`) or untyped.

cli/windmill-utils-internal/src/parse/parse-schema.ts
```diff
@@ -206,15 +206,40 @@ export function argSigToJsonSchemaType(
...
+            newS.items[field] = oldS.items[field];
+          }
+        });
+      }
       newS.originalType = "object[]";
     }
   } else {
-    newS.type = "object";
+    // Preserve existing type when inference fails, default to "object" for undefined/null
+    newS.type = oldS.type ?? "object";
```
hugocasa, line 242: @rubenfiszel are you ok with this? right now we fallback to "object" when we the type can't be parsed

rubenfiszel: that feels reasonable, but we may want to try how it behave with a few real world examples of typing live and see the result with new behavior and old behavior

hugocasa:
moved to #7496

Issue #7252, opened 2025-11-28
feature: MCP Token UI doesn't support folder wildcards - forces manual API calls or workarounds


steve-goldberg:
Labels: enhancement, mcp, ui
Summary

The MCP token creation UI (introduced in #7129 and implemented in #7130) provides a badge-based interface for selecting individual scripts, flows, and endpoints. However, it does not support folder wildcards (e.g., f/outline/*), which are fully supported by the backend.

This forces users to either:

    Manually select every individual script/flow via badges (and recreate the token every time a new script is added)
    Use the API directly to create tokens with wildcard scopes
    Write custom scripts as workarounds

Current Behavior

When using "Generate MCP URL" with "Custom" scope:

    ✅ Can click individual script badges to select them
    ✅ Can click individual flow badges to select them
    ✅ Can click individual endpoint badges to select them
    ❌ Cannot enter folder wildcards like f/outline/*
    ❌ Cannot select a folder to include all scripts/flows within it

Result: Every time a new script is added to a folder, users must:

    Go back to token settings
    Create a new token
    Click all the badges again (including the new script)
    Generate a new MCP URL
    Update their MCP client configuration

Expected Behavior

The UI should provide a way to add folder wildcards, such as:

Option 1: Text input field

    Add an input field where users can type f/outline/* to include all scripts in that folder
    Show a preview of which scripts/flows would be included

Option 2: Folder tree selector

    Display workspace folders in a tree view
    Allow users to select a folder and click "Include all scripts/flows in this folder"
    Automatically generate the f/folder/* scope

Option 3: Wildcard toggle on badges

    When viewing scripts from a folder, show a "Select all in folder" button
    Converts individual selections to a wildcard scope

Workaround (Current)

Users must create tokens via API:

```shell
curl -X POST "https://instance/api/users/tokens/create" \
  -H "Authorization: Bearer TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "label": "MCP - Folder Wildcard",
    "scopes": ["mcp:scripts:f/outline/*", "mcp:flows:f/outline/*"]
  }'
```

Then manually construct the MCP URL: https://instance/api/mcp/w/workspace/sse?token=...

## Why This Matters

    Automatic inclusion of new scripts - Wildcard scopes automatically include new scripts added to the folder without regenerating tokens
    Cleaner scope management - f/outline/* is cleaner than listing 50 individual script paths
    Feature parity - The backend supports wildcards, the UI should too
[...]

PR #7282, opened 2025-12-03
feat(frontend): add wildcard pattern support to MCP token custom scope

devdattatalele:
This PR addresses issue #7252 by adding support for folder wildcard patterns in the MCP token creation UI.

Previously, users could only select individual scripts and flows when creating tokens with custom scopes. The backend already supported wildcard patterns like f/outline/*, but there was no UI to access this functionality, forcing users to create tokens via API calls.

This implementation adds text input fields where users can enter comma-separated wildcard patterns alongside the existing multi-select dropdowns. The patterns are parsed and combined with any individually selected items when the token is created.

Changes include:
- Added text inputs for script and flow wildcard patterns in custom scope section
- Implemented pattern parsing that handles comma separation, whitespace, and empty values
- Updated button validation to recognize pattern inputs as valid selections
- Added state management to clear pattern inputs when switching between scope modes
- Included help popover with pattern syntax examples

The feature maintains backward compatibility since empty pattern fields simply skip pattern processing and use the existing individual selection behavior.

### DIFF
--- a/frontend/src/lib/components/settings/CreateToken.svelte
+++ b/frontend/src/lib/components/settings/CreateToken.svelte
@@ -89,6 +89,26 @@
 		return [{ id: currentWorkspace, name: currentWorkspace }, ...workspacesList]
 	}
 
+	function parsePatterns(input: string): string[] {
+		return input.split(',').map((p) => p.trim()).filter((p) => p.length > 0)
+	}
+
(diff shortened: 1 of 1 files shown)


rubenfiszel:
@hugocasa could you check please

PR #7308, opened 2025-12-05
fix(frontend): use right workspace script version in flow status

hugocasa:


frontend/src/lib/components/FlowGraphViewerStep.svelte
```diff
@@ -21,6 +21,7 @@
 	export let schema: any | undefined = undefined
 
 	export let stepDetail: FlowModule | string | undefined = undefined
+	export let jobScriptHash: string | undefined = undefined
 	let codeViewer: Drawer
```
rubenfiszel, line 25: can we make this svelte 5?

PR #7329, opened 2025-12-10
feat: Add Pydantic BaseModel and dataclass support for Python type inference

### LINKED ISSUE #4700: feature: Support Pydantic / dataclasses
I love windmill as a way to trivially make python scripts into production-ready endpoints. Thanks for building it!

Idea: Any windmill Python script can take a pydantic model or return a pydantic model without breaking windmill's type inference.

But I was sad to see that the input spec inference only works for basic argument types, given that most ML-focused python developers I know rely on libraries like [Pydantic (21k stars)](https://github.com/pydantic/pydantic), Attrs or built-in dataclasses for defining data specs (in [FastAPI (77k stars)](https://github.com/fastapi/fastapi) for example). 

I haven't looked at this for a minute, but I think pydantic supports conversion to json-schema, and this might "just work" for your argument type inference. Then, a python windmill script It would make me very very happy :D 

(Note: there was [a project](https://github.com/ml-tooling/opyrator) geared towards python ML practitioners which does the pydantic bit, and I would love to use except it doesn't handle auth, hosting, etc which is where I see windmill shining).

Thanks for considering this idea.

devdattatalele:
## Summary

Adds Pydantic BaseModel and Python dataclass support to the Python parser, enabling automatic UI generation for complex data structures.

Closes #4700

## Changes

Modified `backend/parsers/windmill-parser-py/src/lib.rs` to detect Pydantic models and dataclasses via AST analysis. The parser now recognizes these custom types and extracts their field schemas without executing any Python code.

## Implementation

The parser uses rustpython AST to pattern match class inheritance and decorators, then recursively extracts field types. Added thread-safe module storage with RAII cleanup and security limits (200 fields max, 10 recursion levels).

## Example

Before:
```python
def main(name: str, age: int, email: str):
    return f"Hello {name}"
```

After:
```python
from pydantic import BaseModel

class User(BaseModel):
    name: str
    age: int
    email: str

def main(user: User):
    return f"Hello {user.name}"
```

Windmill now automatically generates a form with name, age, and email fields from the User model.

## Testing

Added 3 test cases covering basic models, dataclasses, and nested models. All 12 tests pass. No breaking changes to existing functionality.

## Known Limitations

Field default values and validation metadata are not yet extracted. Forward references and union types are not supported. These can be addressed in follow-up PRs if needed.


hugocasa:
moved to #7497

PR #7367, opened 2025-12-14
Add AI Badgr as OpenAI‑compatible provider (optional preset)

ghost:
This PR adds AI Badgr as an optional, OpenAI‑compatible provider to Windmill, following the existing pattern used for providers like Groq, OpenRouter, TogetherAI, etc. No defaults or core behaviors are changed.


miguelmanlyx:
recheck

PR #7395, opened 2025-12-16
feat(aiagent): allow giving messages history

centdix:
https://github.com/user-attachments/assets/fff5b906-1610-4883-9e1d-7fc631e5154a

### DIFF
--- a/backend/windmill-worker/src/ai/types.rs
+++ b/backend/windmill-worker/src/ai/types.rs
@@ -108,7 +108,7 @@ impl Default for OutputType {
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(tag = "kind", rename_all = "lowercase")]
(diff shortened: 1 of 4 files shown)


centdix:
/ai change the input schema of the ai agent, instead of adding `messages` array, we want to have a `history` field, which is a oneOf.
It is either 'auto', with a single "number" value, or "manual" with an array of messages.
We can then also remove the messages_context_length input.
We should make sure this change is backward compatible with the schema that only had messages_context_length (no need for messages as this was added in this PR).
That back comp should be both on the back and the front.
Add a comprehensive description to the history field.
Do one commit for the backend, one commit for the frontend.
Do one additional commit for correctly rendering a oneOf of that type on InputSchemaTransform, we do not handle that correctly

PR #7472, opened 2025-12-30
feat(ai): allow customizing all providers' base_url

ImSingee:
I think this is useful for several reasons:

1. It's possible to use any unofficial services with OpenAI responses or the claude API format
2. For Codestral, it's possible to use the official free version (https://codestral.mistral.ai)

---

Because the Resources' definition isn't included in the code (it's synced from the hub), this PR can't modify it.

### DIFF
--- a/backend/windmill-common/src/ai_providers.rs
+++ b/backend/windmill-common/src/ai_providers.rs
@@ -38,6 +38,12 @@ impl AIProvider {
         region: Option<String>,
         db: &DB,
     ) -> Result<String> {
+        // Resource-level base URL should always take precedence over any provider defaults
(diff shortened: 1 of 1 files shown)


centdix:
> I think this is useful for several reasons:
> 
> 1. It's possible to use any unofficial services with OpenAI responses or the claude API format
> 2. For Codestral, it's possible to use the official free version (https://codestral.mistral.ai)
> 
> Because the Resources' definition isn't included in the code (it's synced from the hub), this PR can't modify it.

Hi, thanks for the PR. Why would using custom provider not be enough to be able to use any base_url ?

PR #7625, opened 2026-01-20
fix(frontend): improve raw app history [Merge ui-builder first]

Guilhem-lm:


frontend/src/lib/components/Editor.svelte
```diff
@@ -1365,7 +1365,7 @@
 					$relativeLineNumbers
 				),
 				model,
-				fontSize: !small ? 14 : 12,
+				fontSize: !small ? 13.5 : 12,
```
rubenfiszel: seems a bit like a preference thingy I would rather not do in this PR

PR #7631, opened 2026-01-20
fix(frontend): set editor font size to the same default as text

Guilhem-lm:


rubenfiszel:
/ai-fast before the editor is loaded, a text placeholder is shown, adjust the font-size of it too

PR #7653, opened 2026-01-22
feat(python-client): add interpolated arg for get_resource

j-o-br:
## Changes

- Adds `interpolated` argument to the `get_resource()` method to allow disabling resource/variable interpolation. Previous behavior was to always interpolate.
- Set default value of `interpolated` to `True` to preserve backwards compatibility


rubenfiszel:
Thanks

PR #7668, opened 2026-01-23
feat(ai): native bedrock compatibility

centdix:


backend/windmill-api/Cargo.toml
```diff
@@ -149,8 +150,11 @@ rustls = { workspace = true }
 
 aws-sigv4.workspace = true
 aws-sdk-config.workspace = true
-aws-config = { workspace = true, optional = true }
```
rubenfiszel, line 152: this should still be optional no?

PR #7727, opened 2026-01-28
feat: /health endpoints

alpetric:
/api/health (simple, no auth optional HEALTH_CHECK_SECRET env):
  ```
  {
    "status": "healthy",
    "timestamp": "2026-01-28T18:40:16.008701263Z",
    "database_healthy": true,
    "workers_alive": 1
  }
```
  /api/health/detailed (authed):
  ```
  {
    "status": "healthy",
    "timestamp": "2026-01-28T18:40:22.585047186Z",
    "version": "EE v1.615.3-6-g81594b8b56",
    "checks": {
      "database": {
        "healthy": true,
        "latency_ms": 0,
        "pool": {"size": 15, "idle": 13, "max_connections": 50}
      },
      "workers": {
        "healthy": true,
        "active_count": 1,
        "worker_groups": ["default"],
        "min_version": "v1.420.0",
        "versions": ["v1.615.3-6-g81594b8b56"]
      },
      "queue": {"pending_jobs": 8, "running_jobs": 32},
      "readiness": {"healthy": true}
    }
  }
```

### DIFF
--- a/backend/windmill-api/src/health.rs
+++ b/backend/windmill-api/src/health.rs
@@ -329,6 +329,42 @@ fn get_version() -> String {
     format!("CE {GIT_VERSION}")
 }
 
+// ============ Background Loop ============
+
+/// Spawn a background task that performs a health check every 10 seconds.
+/// Updates the cache and prometheus metrics continuously.
+pub fn start_health_check_loop(
+    db: DB,
+    mut killpill_rx: tokio::sync::broadcast::Receiver<()>,
+) {
+    tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_secs(10));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        loop {
+            tokio::select! {
+                biased;
+                _ = killpill_rx.recv() => {
+                    tracing::info!("health check loop shutting down");
+                    break;
+                }
+                _ = interval.tick() => {
+                    let result = perform_health_check(&db).await;
+
+                    log_health_status(&result.response);
+                    #[cfg(feature = "prometheus")]
+                    update_health_metrics(&result.metrics_data);
+
+                    let cached = CachedHealthStatus {
+                        status: result.response,
+                        cached_at: std::time::Instant::now(),
+                    };
+                    *STATUS_CACHE.write().await = Some(cached);
+                }
+            }
+        }
+    });
+}
+
 // ============ Handlers ============
 
 /// Log health status based on severity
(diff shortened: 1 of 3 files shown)


backend/windmill-api/src/health.rs
```diff
@@ -0,0 +1,472 @@
...
+        .unwrap_or(false);
+}
+
+#[cfg(feature = "prometheus")]
+lazy_static::lazy_static! {
+    static ref HEALTH_STATUS_GAUGE: Option<prometheus::Gauge> =
+        if METRICS_ENABLED.load(std::sync::atomic::Ordering::Relaxed) {
+            Some(prometheus::register_gauge!(
+                "health_status",
+                "Health status: 1=healthy, 0.5=degraded, 0=unhealthy"
```
rubenfiszel: do not think it's a good idea. You need health_status_phase with 3 labels, healthy, degraded, unhealty. Otherwise won't be usable by our alerts

backend/windmill-api/src/health.rs
```diff
@@ -0,0 +1,472 @@
...
+            ).unwrap())
+        } else {
+            None
+        };
+
+    static ref HEALTH_DATABASE_HEALTHY: Option<prometheus::Gauge> =
+        if METRICS_ENABLED.load(std::sync::atomic::Ordering::Relaxed) {
+            Some(prometheus::register_gauge!(
+                "health_database_healthy",
```
rubenfiszel: there are multiple servers/workers so this could have a confusing name, I would provide the following:

health_database_latency (gauge)
health_database_unresponsive (1/0)
health_database_errors (counter)  this one require a bit more work and you may not want to do it in this version but should be incremented anytime there is an error that's related to PoolTimeout, etc. It can probably be done. There would be one label per type of error

backend/windmill-api/src/health.rs
```diff
@@ -0,0 +1,472 @@
...
+    static ref HEALTH_WORKERS_ALIVE: Option<prometheus::Gauge> =
+        if METRICS_ENABLED.load(std::sync::atomic::Ordering::Relaxed) {
+            Some(prometheus::register_gauge!(
+                "health_workers_alive",
+                "Number of workers alive"
+            ).unwrap())
+        } else {
+            None
+        };
+}
```
rubenfiszel: remove there are other ways to get that metric and given there are multiple servers it will be more confusing than helpful

backend/windmill-api/src/health.rs
rubenfiszel: can also add the pool size metrics

PR #7951, opened 2026-02-15
fix: require super admin for test_object_storage_config endpoint

rubenfiszel:
## Summary

- The `POST /api/settings/test_object_storage_config` endpoint was accessible to any authenticated user without requiring super admin privileges
- This allowed SSRF by connecting to a user-specified S3 endpoint URL (the `ObjectSettings.endpoint` field)
- Added `require_super_admin()` check, consistent with every other endpoint in the settings module

## Details

`test_s3_bucket()` accepted `_authed: ApiAuthed` (unused — note the `_` prefix) and never verified admin status. An attacker could force the Windmill server to make HTTP requests (list, put, get, delete) to arbitrary internal or external hosts via the `endpoint` field.

The fix is a one-line addition of the same `require_super_admin(&db, &authed.email).await?` call used by all other settings endpoints.

## Test plan

- [ ] As super admin: `POST /api/settings/test_object_storage_config` works as before
- [ ] As non-admin user: endpoint returns 403/unauthorized
- [ ] Instance settings UI object storage test button still works for admins

### DIFF
--- a/backend/windmill-api-settings/src/lib.rs
+++ b/backend/windmill-api-settings/src/lib.rs
@@ -12,9 +12,9 @@ use std::{collections::HashMap, time::Duration};
 mod ee;
 pub mod ee_oss;
 
-use windmill_api_auth::{require_super_admin, ApiAuthed};
 #[cfg(feature = "enterprise")]
 use windmill_api_auth::require_devops_role;
+use windmill_api_auth::{require_super_admin, ApiAuthed};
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Closing: the endpoint is legitimately used by all users via the S3/Azure 'Test connection' button in the resource editor. Adding require_super_admin would break that UX. The SSRF risk is marginal since users already have full code execution via scripts.

PR #8031, opened 2026-02-20
feat(aiagent): handle ai agent as tool

centdix:
https://github.com/user-attachments/assets/8a1a8c1c-7cff-4401-b3fb-a4ba0d3d0b6d

## Summary

Adds support for using AI Agents as tools inside other AI Agent steps in the flow builder (nested agents). **Only 2-level nesting is supported: flow → AI agent → nested AI agent tool.** Deeper nesting (agent → agent → agent) is explicitly rejected at runtime.

### Architecture: Why nested AI agent tools differ from regular flow steps

In a normal flow, every step is pushed by the **flow executor** which sets `flow_step_id` on the child job. This ID links the child to a `v2_job_status` row on the parent flow, enabling the result processor to update the parent's flow status on completion.

AI agent tool jobs are **not pushed by the flow executor** — they're pushed by the AI tool executor (`ai/tools.rs`) at runtime as the LLM decides to call tools. The parent of a tool job is the AI agent job, which is a leaf job with **no `v2_job_status` row**. Setting `flow_step_id` on tool jobs would cause the result processor to crash trying to update a non-existent flow status.

This means nested AI agent tools need alternative mechanisms:

- **`flow_step_id` is not set** on tool jobs. Instead, the tool's module ID is encoded in `runnable_path` (e.g., `"f/ws/flow/tools/b"`) and extracted at execution time via `rsplit_once("/tools/")`.
- **`root_job` and `flow_innermost_root_job` are propagated** at push time so nested agents can find the flow job naturally via existing job fields (used for reading flow context: chat settings, memory, flow args).
- **Module lookup** for nested agents goes through the parent: find the parent agent's module in the flow definition, then search its `tools` array for the nested agent's tool definition.
- **Flow status updates are skipped** (`parent_job` is `Option`) since the parent AI agent has no `v2_job_status` to update.

### Deadlock fix: inline tool completion processing

Previously, when a tool completed, `handle_tool_execution_success` would **forward** the tool's `JobCompleted` message through the parent agent's bounded(1) channel (`ctx.job_completed_tx`). For top-level agents this channel was the main worker channel (capacity 10) so it worked fine. But for nested agents, this channel was the **parent agent's inner channel** (capacity 1). When the nested agent called any tool, the forwarded tool result would fill the parent's channel, and then `process_result` could not send the agent's own completion — **deadlock**: the agent blocked waiting for channel capacity, while the parent blocked waiting for the agent's task to finish.

Fix: tool completions are now written to the database **inline** via `add_completed_job` / `add_completed_job_error` directly in `handle_tool_execution_success`, instead of being forwarded through the parent channel. This calls the exact same DB-write functions that the background processor would have called, but skips the channel entirely. The `job_completed_tx` field was removed from `ToolExecutionContext` since it's no longer needed.

### Backend (worker)
- Support `FlowModuleValue::AIAgent` tools in the AI executor tool dispatch
- Execute nested AI agent tools by enqueuing `JobPayload::AIAgent` with the tool's input args
- Propagate `root_job`/`flow_innermost_root_job` when pushing tool jobs so nested agents find the flow job without manual overrides
- If direct parent is an AI agent, go one level up to find the flow job for module definition lookup
- Locate nested agent tool modules through parent agent step ID + tool ID traversal
- Fix deadlock: write tool completions inline instead of forwarding through bounded channel

### Frontend (flow builder)
- Add "AI Agent" as a tool type option in the flow builder tool picker
- Reuse `FlowModuleComponent` for nested AI agent tools instead of a custom minimal editor — this gives the full input transforms view (system_prompt, memory, temperature, etc.) with prop picker and AI toggle
- Ensure all `AI_AGENT_SCHEMA` fields are populated in `input_transforms` at both creation time (`FlowModuleSchemaMap`) and load time (`flowState.ts` now recurses into aiagent tools)
- Add `defaultToAi` prop to `InputTransformSchemaForm` / `InputTransformForm` so nested AI agent tool fields default to static values (not AI-filled), while still showing the AI toggle for fields like `user_message`

## Nesting limitation

Only **flow → agent → nested agent tool** is supported. The limitation comes from module lookup: the nested agent finds its definition by knowing its parent agent's `flow_step_id` (set by the flow executor) and searching that agent's tools array. Supporting deeper nesting (agent → agent → agent) would require parsing the full `runnable_path` chain (e.g., `"f/ws/flow/a/tools/b/tools/c"`) to reconstruct the lookup path — doable but not implemented.

## Test plan
- [ ] Create a flow with an AI Agent step
- [ ] Add a nested AI Agent as a tool
- [ ] Verify all input transforms render (provider, output_type, user_message, system_prompt, memory, temperature, max_iterations, etc.)
- [ ] Verify nested AI agent fields default to static, not AI-filled
- [ ] Run the flow — the nested AI agent tool should execute without `v2_job_status` errors
- [ ] Save and reopen the flow — verify all fields persist correctly
- [ ] Verify that attempting 3-level nesting (agent tool that is itself an agent with agent tools) returns a clear error
- [ ] Verify nested agent with its own tools works without deadlock

### DIFF
--- a/backend/windmill-worker/src/ai/tools.rs
+++ b/backend/windmill-worker/src/ai/tools.rs
@@ -34,15 +34,15 @@ type McpClient = McpClientStub;
 use windmill_common::{
     client::AuthedClient,
     db::DB,
-    error::{to_anyhow, Error},
+    error::Error,
     flow_conversations::MessageType,
     flow_status::AgentAction,
     flows::FlowModuleValue,
     worker::{to_raw_value, Connection},
 };
 use windmill_queue::{
-    get_mini_pulled_job, push, JobCompleted, MiniCompletedJob, MiniPulledJob, PushArgs,
-    PushIsolationLevel,
+    add_completed_job, add_completed_job_error, get_mini_pulled_job, push, MiniCompletedJob,
+    MiniPulledJob, PushArgs, PushIsolationLevel,
 };
 
 /// Context for tool execution containing all required references and state
@@ -66,7 +66,6 @@ pub struct ToolExecutionContext<'a> {
 
     // Runtime state
     pub occupancy_metrics: &'a mut OccupancyMetrics,
-    pub job_completed_tx: &'a JobCompletedSender,
     pub killpill_rx: &'a mut tokio::sync::broadcast::Receiver<()>,
 
     // Optional streaming & chat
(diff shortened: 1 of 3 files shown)


centdix:
/ai-fast review latest changes since last review. Some comments from latest review were done on outdated code, also check the full changes again

PR #8186, opened 2026-03-02
chore(workmux): add name field to config

centdix:
## Summary
Add a `name: windmill` field to the `.workmux.yaml` configuration.

## Changes
- Added `name: windmill` at the top of `.workmux.yaml`

## Test plan
- [ ] Verify `workmux` reads the config correctly with the new `name` field

---

### DIFF
--- a/.workmux.yaml
+++ b/.workmux.yaml
@@ -1,4 +1,4 @@
-name: windmill
+name: Windmill
 
 main_branch: main
 


.workmux.yaml
```diff
@@ -1,3 +1,5 @@
+name: windmill
```
centdix:
```suggestion
name: Windmill
```

PR #8330, opened 2026-03-11
perf: cache composer vendor dir to skip reinstall on repeated php executions

sascha-egerer:


hugocasa:
Hi @sascha-egerer we will review it next week, thanks for the contribution

PR #8505, opened 2026-03-24
fix(deploy): add HTTPS support to ECS Terraform ALB configuration

camgrimsec:
Add optional HTTPS listener with ACM certificate support to the ALB. When acm_certificate_arn is provided, the ALB:
- Terminates TLS on port 443 with TLS 1.3 policy
- Redirects HTTP (port 80) to HTTPS with 301

When no certificate is provided (default for development), the ALB falls back to HTTP-only mode with a clear variable description noting this is not recommended for production.

This follows Windmill's own self-host guidance which recommends TLS termination at the load balancer level.

Security impact: Prevents API tokens, script source code, database credentials, and OAuth tokens from flowing unencrypted between users and the Windmill server.

### DIFF
--- a/examples/deploy/aws-ecs-terraform/load_balancer.tf
+++ b/examples/deploy/aws-ecs-terraform/load_balancer.tf
@@ -1,3 +1,5 @@
+# --- Target Groups (internal HTTP — TLS terminates at ALB) ---
+
 resource "aws_lb_target_group" "windmill_cluster_windmill_server_tg" {
   name        = "windmill-cluster-server-tg"
   port        = 8000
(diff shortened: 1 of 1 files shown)


rubenfiszel (commented):
**Bug: Security group missing port 443 ingress rule**

The security group (`security_group.tf`) only allows inbound traffic on port 80. When HTTPS is enabled, the ALB listens on port 443, but there is no ingress rule for port 443 — HTTPS traffic will be blocked and the listener will be unreachable.

Suggested fix — add to `security_group.tf`:
```hcl
ingress {
  from_port   = 443
  to_port     = 443
  protocol    = "tcp"
  cidr_blocks = ["0.0.0.0/0"]
}
```

PR #8618, opened 2026-03-30
feat: add black-box ai eval benchmarks

centdix:
## Summary
Add a black-box benchmark tool under `ai_evals/` for `flow`, `app`, `script`, and `cli`. The tool runs the current production prompts and guidance, saves the generated artifacts for inspection, records per-mode benchmark history, and scores outputs with a small deterministic validity layer plus an LLM judge driven by per-case checklists.

## Reviewer focus
- `ai_evals/cases/*.yaml`
  The benchmark corpus, natural-language prompts, optional fixtures, and per-case `judgeChecklist`.
- `ai_evals/core/validators.ts`, `ai_evals/core/judge.ts`, `ai_evals/core/runSuite.ts`
  The scoring contract: deterministic checks cover hard validity, while semantic correctness is judged by the LLM.
- `ai_evals/modes/*.ts`
  The four benchmark modes and how each one runs and serializes artifacts.
- `frontend/src/lib/components/copilot/chat/flow/inlineScriptsUtils.ts`, `frontend/src/lib/components/copilot/chat/flow/core.ts`, `frontend/src/lib/components/copilot/chat/flow/FlowAIChat.svelte`
  The flow inline-script isolation fix used by both production chat and the benchmark harness.
- `cli/src/guidance/writer.ts`
  The generated guidance bundle consumed by the CLI benchmark path.

## How The Benchmarks Run
- Frontend modes (`flow`, `app`, `script`) run through a dedicated Vitest adapter in `ai_evals/adapters/frontend/`.
  The benchmark CLI spawns Vitest from the `frontend/` install, using `ai_evals/adapters/frontend/vitest.config.ts` so benchmark concerns stay out of the product `frontend/vite.config.js` test includes.
  That adapter then imports the real production chat code from `frontend/src/...`, runs the benchmark case through the production prompt/tool loop, and writes the result JSON back to `ai_evals`.
- CLI mode runs through `ai_evals/modes/cli.ts` and `ai_evals/adapters/cli/runtime.ts`.
  Each case gets an isolated temp workspace with generated `AGENTS.md`, `CLAUDE.md`, and `.claude/skills/` guidance files.
  The prompt is executed with the Anthropic Claude Agent SDK, using the real guidance bundle and a constrained tool set (`Skill`, `Read`, `Glob`, `Grep`, `Bash`, `Write`, `Edit`).
  The benchmark captures both the assistant transcript and the files actually written in the workspace.
- In all modes, deterministic validation only checks hard validity. Semantic success is judged by an LLM against the case prompt plus `judgeChecklist`.

## Developer Workflow
- For a frontend prompt or product change, run the relevant suite before the change and inspect the saved artifacts:
  - `cd ai_evals && bun run cli -- run flow --model haiku`
  - `cd ai_evals && bun run cli -- run app --model haiku`
  - `cd ai_evals && bun run cli -- run script --model haiku`
- Make the prompt or product change, rerun the same command, and compare the new pass rate plus the saved outputs under `ai_evals/results/`.
- For CLI guidance changes, run:
  - `cd ai_evals && bun run cli -- run cli --model haiku`
  This exercises the real generated guidance bundle through the Claude Agent SDK and saves both the assistant transcript and written workspace files.
- When a run should become part of tracked benchmark history, rerun it with `--record`:
  - `cd ai_evals && bun run cli -- run flow --model haiku --record`
  - `cd ai_evals && bun run cli -- run cli --model haiku --record`
  This appends a compact summary line to `ai_evals/history/<mode>.jsonl`.
- In practice, the main workflow is: run suite -> inspect `ai_evals/results/...` artifacts -> change prompts/guidance -> rerun -> record only the runs worth tracking in git.

## Changes
- add a benchmark CLI under `ai_evals/` with `models`, `cases`, and `run`
- add benchmark modes for `flow`, `app`, `script`, and `cli`, all using the current production prompts or guidance
- define the suites in YAML case manifests with realistic prompts, optional `initial` / `expected` fixtures, and per-case `judgeChecklist` entries
- add deterministic validators for hard failures such as missing artifacts, syntax errors, unresolved refs, missing special modules, placeholder corruption, and unexpected CLI files
- add an LLM judge that scores semantic correctness against the prompt and each case checklist instead of requiring exact expected shapes for broad creation tasks
- save generated artifacts for every run under `ai_evals/results/` so the outputs can be inspected directly (`flow.json`, `app.json`, workspace files, assistant output, etc.)
- add per-mode benchmark history recording via `run --record`, writing compact tracked summaries to `ai_evals/history/<mode>.jsonl`
- add and refine benchmark corpora for:
  - `flow`: create, modify, reference, and edge-case flows
  - `app`: raw-app modification and creation tasks with stronger checklist coverage
  - `cli`: create, modify, and reuse tasks against the generated guidance bundle
- add flow branch-output guidance and per-run inline-script isolation based on issues found by the benchmark suite
- document the benchmark tool in `ai_evals/README.md` and `docs/system-prompt-testing-status.md`

## Test plan
- [x] `cd frontend && bun install`
- [x] `cd ai_evals && bun install`
- [x] `cd ai_evals && bun run cli -- models`
- [x] `cd ai_evals && bun run cli -- run flow --model haiku`
- [x] `cd ai_evals && bun run cli -- run app --model haiku`
- [x] `cd ai_evals && bun run cli -- run cli --model haiku`
- [x] `cd ai_evals && bun run cli -- run flow --model haiku --record`
- [x] inspect saved outputs under `ai_evals/results/` and tracked summaries under `ai_evals/history/*.jsonl`

---

### DIFF
--- a/ai_evals/README.md
+++ b/ai_evals/README.md
@@ -83,11 +83,15 @@ Today:
 - `sonnet`
 - `opus`
 - `4o`
+- `gemini-flash`
+- `gemini-pro`
+- `gemini-3-flash-preview`
+- `gemini-3.1-pro-preview`
 
 Notes:
 
 - the command also prints accepted alias spellings such as `gpt-4o`, `claude-opus-4.6`, and `claude-haiku-4.5`
-- frontend modes (`flow`, `script`, `app`) can use Anthropic and OpenAI-backed aliases
+- frontend modes (`flow`, `script`, `app`) can use Anthropic, OpenAI, and Gemini-backed aliases
 - `cli` mode always uses the Anthropic agent SDK, so only Anthropic aliases are valid there
 - the judge model is separate and currently defaults to `claude-sonnet-4-6`
 
--- a/ai_evals/adapters/frontend/core/shared/baseEvalRunner.ts
+++ b/ai_evals/adapters/frontend/core/shared/baseEvalRunner.ts
@@ -1,16 +1,19 @@
-import OpenAI from 'openai'
-import Anthropic from '@anthropic-ai/sdk'
 import type {
 	ChatCompletionMessageParam,
 	ChatCompletionSystemMessageParam
 } from 'openai/resources/chat/completions.mjs'
-import type { AIProvider, AIProviderModel } from '$lib/gen/types.gen'
+import type { AIProviderModel } from '$lib/gen/types.gen'
 import type { TokenUsage, ToolCallDetail, EvalRunnerOptions, RawEvalResult } from './types'
 import { runChatLoop, type ChatClients } from '../../../../../frontend/src/lib/components/copilot/chat/chatLoop'
 import type {
 	Tool as ProductionTool,
 	ToolCallbacks
 } from '../../../../../frontend/src/lib/components/copilot/chat/shared'
+import {
+	createEvalClients,
+	type FrontendEvalProvider,
+	resolveEvalModelProvider
+} from './providerConfig'
 
 /**
  * Parameters for running a base evaluation.
@@ -40,35 +43,6 @@ export interface RunEvalParams<THelpers, TOutput> {
 }
 
 /**
- * Creates SDK clients for the given provider.
- */
-function createEvalClients(provider: AIProvider, apiKey: string): ChatClients {
-	if (provider === 'anthropic') {
-		return {
-			openai: new OpenAI({ apiKey: 'unused' }),
-			anthropic: new Anthropic({ apiKey })
-		} as ChatClients
-	}
-	return {
-		openai: new OpenAI({ apiKey }),
-		anthropic: new Anthropic({ apiKey: 'unused' })
-	} as ChatClients
-}
-
-/**
- * Resolves model string to AIProviderModel.
- */
-function resolveModelProvider(
-	model: string,
-	provider?: AIProvider
-): AIProviderModel {
-	if (provider) return { provider, model }
-	if (model.startsWith('claude')) return { provider: 'anthropic', model }
-	if (model.startsWith('gpt') || model.startsWith('o')) return { provider: 'openai', model }
-	return { provider: 'openai', model }
-}
-
-/**
  * Runs a generic evaluation using the shared chat loop (same code path as production).
  * Uses streaming via real provider SDKs instead of OpenRouter non-streaming.
  */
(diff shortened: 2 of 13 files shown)


centdix:
Addressed the review items we agreed to take in this pass:

Fixed
- `wmill init` partial-failure regression: `writeAiGuidanceFiles()` now writes `AGENTS.md` / `CLAUDE.md` before skill copying/writing, so guidance files still land if `.claude/skills` creation fails.
- `ai_evals/core/runSuite.ts`: moved `loadInitial()` / `loadExpected()` inside the per-attempt `try`, so bad fixture paths become normal failed attempts instead of escaping the worker.
- `ai_evals/cli/index.ts`: `--record` is now rejected for partial runs with selected case ids.
- `ai_evals/adapters/frontend/core/shared/baseEvalRunner.ts`: guarded `onAssistantMessageEnd` so it only fires after a matching streamed assistant message start.
- `cli/src/guidance/writer.ts`: removed the dead `skillsWritten` state.
- small cleanups while touching the same areas:
  - removed the dead pre-`mkdir` in `ai_evals/modes/cli.ts`
  - removed the unused `getFrontendRunModelLabel`
  - added a clear YAML-list guard in `ai_evals/core/cases.ts`
  - removed the unsafe `surface as Exclude<..., "cli">` progress casts by only emitting frontend progress events for frontend modes

Verified
- `cd cli && bun test ./test/guidance_writer_unit.test.ts`
- manual `wmill init --use-default --no-bind-profile` in a fresh temp dir
- manual `wmill init --use-default --no-bind-profile` with `.claude` forced to be a file, confirming `AGENTS.md` / `CLAUDE.md` still get written
- `cd ai_evals && bun run cli -- run flow flow-test0-sum-two-numbers`
- `cd ai_evals && bun run cli -- run flow flow-test0-sum-two-numbers --record` now fails immediately with the intended full-suite-only error

Not changed in this pass
- `cli/src/guidance/writer.ts` frontmatter parsing: still the current simple parser. I agree it is a latent footgun for multiline YAML, but it is not needed for the current skill metadata we ship.
- `prepareFlowUserMessage(..., inlineScriptSession?)` optionality / `FlowAIChatHelpers.setCode` invariant docs: agreed these are worth hardening, but they are not currently causing incorrect behavior on this branch.
- `validateCliWorkspace()` unexpected-files policy: kept strict for the current CLI corpus because those cases are intentionally narrow and exact-path oriented. If we add broader creation cases later, I think this should become case-configurable rather than relaxed globally.

PR #8804, opened 2026-04-12
feat: add deploy restriction rule and fork review requests

rubenfiszel:
## Summary

Two related additions to the workspace fork/merge flow:

1. **New protection rule `RestrictDeployToDeployers`** — when active on a workspace, only admins and members of the `wm_deployers` group can deploy (create/update) items there. Composes with the existing `DisableDirectDeployment` rule.
2. **Fork review requests + comments** — a lightweight conversation layer on forks so a non-deployer fork author can request review from admins/deployers in the parent, reviewers leave general or anchored (kind+path pinned) comments, anyone with fork access can reply, anchored comments become obsolete on item change, and the whole thing auto-closes on successful merge.

The review flow uses implicit approval — no approve button. Comments are communication, the merge button is ACL-gated by the rule. This keeps the rule and review system decoupled.

### Part 1 — `RestrictDeployToDeployers` rule (CE)
- `backend/windmill-common/src/workspaces.rs` — new bitflag + `ProtectionRuleKind` variant, `wm_deployers` implicit-allow arm in `check_user_against_rule`, new `check_deploy_rules` helper that runs both gating rules in one call
- `backend/windmill-api-{scripts,flows,workspaces,groups,store}/.../*.rs` — migrated ~30 call sites from `check_user_against_rule(.., DisableDirectDeployment, ..)` to `check_deploy_rules(..)` so the new rule is enforced at every deploy site
- `frontend/src/lib/components/workspaceSettings/RulesetEditor.svelte` — new toggle alongside the existing two
- `backend/windmill-api/openapi.yaml` — new enum value, regenerated TS client

### Part 2 — Fork review requests (CE + EE)
- `backend/migrations/20260412212944_fork_review_requests.{up,down}.sql` — three tables: `workspace_fork_review_request` (partial unique index enforces one-open-per-(source,fork)), `workspace_fork_review_reviewer`, `workspace_fork_review_comment` (with `anchor_kind`/`anchor_path` and `obsolete` flag, plus a CHECK that the two anchor columns are both set or both null)
- `backend/windmill-api-workspaces/src/fork_reviews.rs` — new HTTP module mounted at `/api/w/{w_id}/fork_review/*`:
  - `GET /deployers` — eligible reviewers in the parent workspace (admin or `wm_deployers`, excluding disabled users)
  - `GET /open` — the open request + reviewers + comments grouped by parent for the fork
  - `POST /request` — create a request; input reviewers are deduped before validation; unique-index violation surfaces a 409 Conflict; reviewer rows batch-inserted via `UNNEST`
  - `POST /{id}/cancel` — requester or admin only; atomic `WHERE closed_at IS NULL` update so concurrent callers don't overwrite `closed_reason`
  - `POST /{id}/close_merged` — called by the UI after a successful merge, gated on admin or `wm_deployers` membership in the *parent* workspace (query joins `usr`/`usr_to_group` on the parent id, since `authed.groups` reflects the fork workspace)
  - `POST /{id}/comment` — top-level or reply, general or anchored; anchored comments validated against `workspace_diff`; replies must target a top-level comment (2-level thread cap)
  - Audit log entries on every mutation, email dispatch (no-op in OSS) to reviewers/requester on all events
- `backend/windmill-common/src/fork_reviews_oss.rs` (CE stub) + `windmill-common/src/fork_reviews_ee.rs` (EE impl) — `mark_anchor_obsolete` + `close_open_request_on_merge` helpers via the standard `_oss`/`_ee` module pattern. Module-level comment in the OSS stub points at the EE call site.
- `windmill-git-sync/src/git_sync_ee.rs` — one hook point in `tally_deployed_object_changes` covers every item kind (script/flow/app/resource/variable/folder/…); it's a no-op for workspaces without an open request. On CE builds the hook is a stub, so anchored-comment obsolescence after an item change is EE-only; the CE integration test simulates it with a raw `UPDATE`.
- `frontend/src/lib/components/forkReview/ForkReviewPanel.svelte` — request flow, reviewer multi-select, general comment + threaded replies, obsolete rendering (opacity + "obsolete" badge), cancel button, refreshable
- `frontend/src/lib/components/CompareWorkspaces.svelte` — embeds the panel and calls `close_merged` after every successful merge loop
- `backend/windmill-api/openapi.yaml` — new endpoints + schemas

> Note: v1 only exposes general comments + replies in the UI. Anchored comments (pinned to a specific diff row) are a backend feature — the panel renders them + their obsolete state when they exist, but creating an anchored comment from the UI is out of scope for this PR and will land in a follow-up that adds per-row "Comment" buttons in the diff list.

### Tests
- `backend/tests/protection_rules.rs` + `fixtures/restrict_deploy_to_deployers.sql` — new test for admin / deployer / non-deployer / bypass_users behavior. Uses a dedicated workspace id to avoid racing against the existing test on the shared `PROTECTION_RULES_CACHE` / `AUTH_CACHE` lazy_statics.
- `backend/tests/fork_review_requests.rs` + `fixtures/fork_review_requests.sql` — full lifecycle: one-open 409 enforcement, reviewer ACL, reply threading, 2-level reply cap rejection, anchor obsolescence, cancel ACL, merge-close flips all comments obsolete.

## Test plan

- [x] `cargo test --test protection_rules` (2 tests pass)
- [x] `cargo test --test fork_review_requests` (full lifecycle test passes)
- [x] `cargo check` clean on CE
- [x] `cargo check --features enterprise,private` clean on EE
- [x] `./update_sqlx.sh` run; new query cache committed
- [x] Manual browser smoke test with three users (admin, forker, deployer) via headless Playwright: rule toggle in editor, forker requests review + posts comment, deployer sees + replies, merge auto-closes request and flips comments obsolete
- [ ] Reviewer to exercise the flow via the Testing instructions below

## Testing instructions

Log in as `admin@windmill.dev` / `changeme`, switch to `parent-ws`, go to Workspace Settings → Rulesets → New rule, give it a name and toggle on **Restrict deployment to wm_deployers** alongside the existing two toggles. Save, then log in as `forker@windmill.dev` / `changeme`, switch to `fork-ws`, open **Merge workspaces** (the fork compare page), and scroll past the diff list to the new **Review** section. Click **Request review**, pick `admin` from the dropdown, send, then post a general comment. Log in as `deployer@windmill.dev` / `changeme`, reopen the same compare page to see the open request + comment, and reply. Finally as admin or deployer, merge the selected items and verify the panel flips back to **No open request** — the just-closed request's comments are marked obsolete in the DB.


rubenfiszel:
Addressed review comments from @claude (auto-review) and @cubic-dev-ai in 5aef7bf:

**Fixed**
- `close_merged` now resolves the parent workspace and checks `usr` / `usr_to_group` there rather than relying on `authed.groups`, which is scoped to the fork workspace. (cubic P1, claude #2)
- `create_review_request` dedupes `body.reviewers` before comparing against the SQL result length, so duplicate usernames in the input no longer cause a false "not admin" rejection. (claude #1)
- Unique-violation on `create_review_request` now returns `409 Conflict` via `Error::Generic(StatusCode::CONFLICT, ...)` instead of `400`; openapi spec updated and `fork_review_requests.rs` now asserts the 409 status. (cubic P3, claude #5)
- `cancel_review_request` and `close_merged` both scope their updates to `WHERE closed_at IS NULL` and check `rows_affected` so a concurrent cancel + close_merged race can no longer overwrite `closed_reason`. (claude #6)
- Backend now rejects reply-to-reply (parent's `parent_id` must be `NULL`); test added that asserts the 400. (claude #7)
- `ForkReviewPanel`'s reviewer `<Select>` now uses `d.username` directly as the option value instead of parsing it back from the formatted label via `label.split(' ')[0]`, which would have truncated usernames containing spaces. (cubic P2)
- Removed the `(open as any).id` cast in `CompareWorkspaces.svelte` — the generated type for `getOpenForkReviewRequest` is already `ForkReviewRequest | null`. Also dropped the same-shaped cast in `ForkReviewPanel.load()`. (cubic P2, claude #8)
- Minor nits: `comments: comments` → shorthand, `let _ = authed;` → `_authed: ApiAuthed`, `list_deployers` and `list_deployers`/`create_review_request` now also filter `disabled = false`.
- Added a doc comment in `fork_reviews_oss.rs` pointing at the EE hook site in `git_sync_ee.rs::tally_deployed_object_changes` so readers can find the actual implementation. (claude #4)

**Not fixed (deferred)**
- claude #3: "anchored comment creation is claimed but not exposed in the UI" — the panel renders anchored comments + obsolete state, but the create path for anchored comments currently requires a direct API call. Wiring up a per-diff-row "Comment" button is a larger UI addition; I'll narrow the PR description instead and follow up with a separate PR that adds per-row comment buttons.

PR #8807, opened 2026-04-13
perf: add inline-persist fast path for WAC v2 step()

rubenfiszel:
## Summary

WAC v2's `step()` primitive runs a small deterministic function (timestamp, random id, config read) inline in the parent workflow subprocess — no child job is dispatched. Today every `step()` nevertheless raises `_StepSuspend`, which unwinds the parent subprocess, persists a checkpoint delta, and forces the worker to re-pull and replay the workflow from the top. For a workflow with N `step()` calls, that's **N subprocess spawns and N queue round-trips**.

This PR adds an in-process fast path. The SDK executes `fn()`, POSTs the delta to a new per-job API endpoint, and returns the value directly so the workflow continues into the next `step()` in the **same subprocess**. The legacy suspend-and-replay path is preserved as a uniform failure fallback (any network/auth/timeout/version error → SDK re-raises `_StepSuspend` exactly like today) and can be forced on via `WM_WAC_INLINE_FAST_PATH=0` for A/B testing or rollback.

## Benchmark (100-step workflow, debug build, localhost)

| | Wall-clock | Per step | Subprocess spawns | Speedup |
|---|---|---|---|---|
| TS fast path | **614 ms** | ~6.1 ms | **1** | **8.77×** |
| TS legacy | 5 387 ms | ~54 ms | ~101 | — |
| Python fast path | **1 028 ms** | ~10.3 ms | **1** | **9.49×** |
| Python legacy | 9 752 ms | ~97 ms | ~101 | — |

Release build should push both speedups to roughly 15-20× — debug build heavily inflates the fast path (axum/sqlx/tracing are generics-heavy) while bun/python subprocess spawn cost (which dominates legacy) is nearly constant across profiles.

## Correctness

The fast and legacy paths are verified to produce **byte-identical DB state** (modulo per-run `started_at` / `duration_ms` / `scheduled_for` fields). End-to-end diff on real 100-step runs in both TS and Python shows:
- 100 identical entries in `_checkpoint.completed_steps`
- 100 identical `_step/<key>` timeline entries (same top-level keys, same non-volatile shape)
- Same `json_result`

Correctness is structural, not coincidental:
- **One DB write per step, behind one shared helper.** The new endpoint and the worker-side fallback arm both call `windmill_common::wac::persist_inline_checkpoint_delta`, which does load → source-hash-check → `add_completed_step` → save + `_step/<key>` timeline in a single transaction. Whichever path runs, the DB sees the exact same write.
- **Uniform fallback.** Any error in the fast path (network, 5xx, auth, source-hash mismatch, timeout, old backend without the endpoint) re-raises `_StepSuspend` with the same `{key, result, started_at, duration_ms}` payload, which goes through the untouched worker arm. Mixed-path runs converge on the same final DB state.
- **Source-hash guard** applies to both paths — code changes between replays cause the same `ExecutionErr` in both modes.
- **Automated equivalence test** (`backend/tests/python_jobs.rs::test_python_wac_v2_step_inline_fast_path`) runs the same workflow twice with the kill switch on and off and asserts the result + completed_steps shape match.

## Changes

- `backend/windmill-common/src/wac.rs` (new): `WacCheckpoint`, `WacPendingSteps`, `load_checkpoint`, `save_checkpoint`, `add_completed_step`, and the new `persist_inline_checkpoint_delta(db, job_id, source_hash_hint, key, result, started_at, duration_ms)` helper. Lives in `windmill-common` so the API can use it without pulling in the worker crate.
- `backend/windmill-worker/src/wac_executor.rs`: `pub use` re-exports from `windmill_common::wac` so existing call sites are untouched.
- `backend/windmill-worker/src/bun_executor.rs`: `handle_wac_v2_output`'s `WacOutput::InlineCheckpoint` arm now delegates to `persist_inline_checkpoint_delta` and still resets `running=false` + returns `WacSuspended` — functionally identical to before.
- `backend/windmill-api/src/jobs.rs`: new `wac_inline_checkpoint` handler + `WacInlineCheckpointPayload` + route `/wac/inline_checkpoint/{job_id}` registered in `workspaced_service()`. `ApiAuthed`, workspace-scoped, does NOT touch `v2_job_queue`.
- `python-client/wmill/wmill/client.py`: `WorkflowCtx._run_inline_step` gets a fast-path branch (httpx POST with 10s timeout, serialized by lazy `asyncio.Lock`). Gated by `WM_WAC_INLINE_FAST_PATH`. Falls through to `_StepSuspend` on any failure.
- `typescript-client/client.ts`: `WorkflowCtx._runInlineStep` mirrors the Python behavior via raw `fetch` + `AbortController` + promise-chain serialization. Same kill switch, same fallback.
- `backend/tests/python_jobs.rs`: `test_python_wac_v2_step_inline_fast_path` runs the same 5-step workflow twice (fast vs legacy via `os.environ["WM_WAC_INLINE_FAST_PATH"]` injected into the script) and asserts identical result + `completed_steps`.

## Distribution note

The performance win activates once a published `wmill` / `windmill-client` release contains the SDK fast-path code. Until then, workers installing from the public registry run the unchanged legacy path — which is the intended "uniform fallback" behavior (still correct, just not faster). The backend endpoint, shared helper, and kill switch all ship in this PR; the SDK side lands with the next release of the clients.

## Test plan

- [x] `cargo check` passes cleanly
- [x] Python SDK unit tests green (`pytest tests/test_workflow.py` — 83/84 pass; the one failing test is a pre-existing `@task` `__name__` preservation issue unrelated to this PR)
- [x] TS SDK unit tests green (`bun test tests/workflow.test.ts` — 81/81 pass)
- [x] Manual end-to-end with a locally-patched SDK verified both TS and Python 100-step fast path: single subprocess spawn, identical `workflow_as_code_status` vs legacy, 8-9× speedup in debug build
- [x] Kill-switch drill (`WM_WAC_INLINE_FAST_PATH=0`) verified — legacy path still produces identical DB state
- [ ] `cargo test test_python_wac_v2_step_inline_fast_path` on CI (the test compiles clean; once the SDK fast-path code is published to PyPI/npm, this test will become a real end-to-end equivalence check)


rubenfiszel:
Thanks @cubic-dev-ai and @claude for the reviews — addressed in `e29c1afc6e`:

**Security (blocking)** — `wac_inline_checkpoint` now takes `OptJobAuthed` and rejects with `PermissionDenied` unless the token's `job_id` JWT claim matches the URL path, closing the workspace-wide privilege gap. Regular workspace API tokens have `job_id == None` so they fail the check unconditionally; only the ephemeral token the worker sets into `WM_TOKEN` when spawning the subprocess is accepted. Also added `AND kind = 'script'::job_kind` to the lookup so a forged/buggy caller can't write a bogus checkpoint onto a flow job or dependency job.

**Atomicity regression** — `persist_inline_checkpoint_delta` now takes `&mut Transaction<'_, Postgres>` instead of `&DB`. The caller owns the transaction and commits it. The worker-side `WacOutput::InlineCheckpoint` arm opens its own tx, calls the helper, runs `UPDATE v2_job_queue SET running = false` in the **same** tx, and commits once — restoring the pre-refactor all-or-nothing semantics. The API endpoint just wraps the helper call in a single-purpose tx.

**Read-modify-write race** — first DB op inside the helper is now `SELECT workflow_as_code_status->'_checkpoint' FROM v2_job_status WHERE id = $1 FOR UPDATE`. Concurrent callers serialize on the row lock and the second one reads the freshly-persisted checkpoint, so no step result can be lost.

**Silent source-hash bypass** — when `source_hash_hint` is `None` or `Some("")` (preview / inline jobs with no `runnable_id`), the helper now emits a `tracing::debug!` so operators can see which jobs are running without drift protection instead of it being invisible.

**Flaky wall-clock assertion** — already removed in `6a1de735e8` (prior to these reviews). The integration test keeps the behavioral-equivalence asserts (`legacy_result == fast_result`, `legacy_ckpt == fast_ckpt`) and prints the timing ratio without asserting a threshold.

**Python httpx client reuse** — `WorkflowCtx` now lazy-initializes a single `httpx.AsyncClient` with the auth header baked in on the first fast-path call and reuses it across all subsequent `step()` invocations in that workflow invocation. Saves ~15ms/step of connection-pool setup/teardown in the hot path.

**Unused re-export** — dropped `add_completed_step` from `windmill_worker::wac_executor`'s `pub use`. Nothing in `windmill-worker` referenced it after the refactor.

**Not addressed** (declined):

- **`check_scopes("jobs:run")`** (cubic) — the `OptJobAuthed` binding is stronger than a scope check for this specific endpoint. Scope checks guard against user tokens being misused for privileged actions, but here the entire endpoint is already gated on "token was issued for *this* specific job". Adding a `jobs:run` scope check on top would only matter if we later introduced scope-stripped ephemeral tokens, which isn't the current model. Happy to add it if you'd rather belt-and-braces.
- **TS `OpenAPI.BASE`/`OpenAPI.TOKEN` parity with Python** (claude nit) — these are set by `setClient()` which the bun worker wrapper calls during init, mirroring Python's reliance on `WM_TOKEN`/`BASE_INTERNAL_URL` env vars. The intent does match.
- **`std::env::set_var` race in the test** (claude nit) — already removed in `6a1de735e8`. The test now injects `WM_WAC_INLINE_FAST_PATH` into the workflow script content itself (`os.environ[...] = ...`), so there's no shared process env mutation and no race with `--test-threads=N`.

### Re-measured fast vs legacy after fixes

Debug build, 100-step workflow, patched-SDK dev instance:

| | Wall-clock | Per step | Spawns | Speedup |
|---|---|---|---|---|
| TS fast | 614 ms | ~6.1 ms | **1** | **8.77×** |
| TS legacy | 5387 ms | ~54 ms | ~101 | |
| Python fast | 1028 ms | ~10.3 ms | **1** | **9.49×** |
| Python legacy | 9752 ms | ~97 ms | ~101 | |

Structural DB equivalence between fast and legacy was verified end-to-end — `workflow_as_code_status` byte-equal modulo volatile timing fields (see PR description).

rubenfiszel:
Round-2 fixes in `9e6ddb852f`:

**1. `httpx.AsyncClient` never closed.** Fixed — `WorkflowCtx._inline_http_client` is now closed in `_run_workflow_async`'s `finally` block. No more `ResourceWarning('unclosed transport')` on shutdown, no connection-pool leak when this coroutine is driven from a long-lived event loop (tests / REPL / embedded callers).

**2. `SELECT ... FOR UPDATE` doesn't lock non-existent rows.** Correct — the comment overstated this. Rather than paper over with advisory locks, the helper now starts with:

```sql
INSERT INTO v2_job_status (id, workflow_as_code_status)
VALUES ($1, '{}'::jsonb)
ON CONFLICT (id) DO UPDATE
  SET workflow_as_code_status = v2_job_status.workflow_as_code_status
RETURNING workflow_as_code_status->'_checkpoint'
```

The `DO UPDATE SET col = col` branch is a no-op assignment whose sole purpose is to force Postgres to take the row-level lock (which `DO NOTHING` releases immediately). The lock is held until the caller commits, and the `RETURNING` clause gives us the current `_checkpoint` in the same round-trip. Concurrent callers on the same `job_id` now serialize naturally even on the very first write for that job — the invariant matches the comment.

Verified end-to-end: both TS and Python 100-step benches still pass, structural equivalence between fast and legacy paths still holds.

rubenfiszel:
Perf investigation + fix (`9254d39074`).

The cubic round-2 suggestion of "make the lock-acquire reliable even on first write" led me to prototype a single-statement `INSERT ... ON CONFLICT DO UPDATE` that pushed the load-modify-save entirely into SQL via `jsonb_set(jsonb_set(... || jsonb_build_object(...)))`. Design goals: 1 tuple write per step, no first-write race, no SDK-side lock, no cross-process race.

Benchmarked it on a 100-step TS workflow in debug mode and it was **~80 ms/step** — ~14× slower than the 2-statement load-modify-save. The nested `COALESCE(v2_job_status.workflow_as_code_status->'_checkpoint'->...)` accesses make Postgres evaluate the growing JSONB column multiple times per call, and the `||` merges re-serialize the whole subtree. Confirmed linear-in-N (not quadratic), so it's a constant ~80 ms overhead per call rather than algorithmic blowup — but that constant wipes out the fast-path win.

Reverted to the two-statement version (`SELECT ... FOR UPDATE` → load in Rust → add step → `INSERT ... ON CONFLICT DO UPDATE` for checkpoint → separate `UPDATE` for timeline) and kept the documented first-write race. The helper's doc comment now has a "why we tried and rolled back the single-statement variant" note so nobody re-prototypes the same thing.

Also fixed a latent bug from the previous round: the `kind = 'script'` filter I added to the endpoint's job lookup was rejecting preview jobs, which is what both the integration test and every manual bench on the dev instance use. The endpoint would 404 on every fast-path call and the SDK would uniformly fall back to legacy. Widened to `kind IN ('script','preview')`, which still excludes flow/dependency/trigger jobs.

Benchmarks after both fixes (100-step TS workflow, debug build, localhost, patched SDK so the fast path actually runs):

| | Wall | Per step | Subprocess spawns |
|---|---|---|---|
| Fast path | **575 ms** | 5.8 ms | **1** |
| Legacy | **5371 ms** | 53.7 ms | ~101 |
| Speedup | **9.3×** | | |

Structural equivalence verified end-to-end: `workflow_as_code_status` from the fast and legacy runs is byte-equal once you strip `started_at` / `scheduled_for` / `duration_ms`. 100 `completed_steps` entries + 100 `_step/<key>` timeline entries on both sides, same set of keys, same values.

Release build should push the speedup higher — debug-build axum/sqlx/tracing overhead eats a big chunk of the fast path; subprocess spawn cost (which dominates legacy) is nearly profile-independent.

rubenfiszel:
Round-3/4 fixes in \`8cd87d66fe\`. Three issues, all addressed:

**1. `httpx.AsyncClient.aclose()` + `CancelledError`.** The existing `finally` block wrapped `aclose()` in `try/except Exception`, but `asyncio.CancelledError` has been a `BaseException` since Python 3.8 — if cancellation hit while the close was in-flight, it'd propagate out and skip `_workflow_ctx.reset(token)`. Wrapped the cleanup in its own nested `try/finally` so the context-var reset is always reached, regardless of what aclose raises.

**2. Wrong reasoning in the `persist_inline_checkpoint_delta` doc comment.** Cubic is right that `distinct step keys don't protect` against data loss, because the helper writes the **whole serialized `_checkpoint`** object per call (INSERT ON CONFLICT DO UPDATE SET workflow_as_code_status = jsonb_set(..., '{_checkpoint}', $2)), not individual `completed_steps[key]` entries. Two concurrent first-writes would each serialize a fresh `WacCheckpoint` with only their own step and the second would overwrite the first.

The comment previously said the SDKs could fire in parallel without a client-side lock — that was also wrong (I'd removed the SDK locks in `dd088eb6a3` on the theory that jsonb_set composed across disjoint keys, which isn't what the helper does). So both the comment and the runtime invariant were broken.

**Fix: restored the SDK-side serialization.** Python `asyncio.Lock`, TS `_inlineChain` promise chain. Crucially the lock wraps only the POST, not `fn()`, so `asyncio.gather(step("a", fn_a), step("b", fn_b))` still runs the fn() bodies in true parallel; only the API requests are ordered. The only thing "serialized" is the ~ms POST round-trip.

**Rewrote the doc comment** to accurately describe:
- The steady-state row-lock behavior (works fine once the row exists).
- The first-write race window and how it's closed on the client side.
- An explicit warning: "Future contributors: do not remove the SDK-side lock without also fixing the server-side first-write guarantee."
- Why a single-statement merge-UPDATE isn't used (~80 ms/step in PG, 14× slower than the Rust-side load-modify-save).
- The cross-process safety argument (both paths receive the same `_StepSuspend` payload → idempotent writes, not divergent).

### Verified end-to-end

- `cargo check` clean.
- Python SDK unit tests: 83/84 pass (same pre-existing `@task __name__` failure unrelated to this PR).
- TS SDK unit tests: 81/81 pass.
- 100-step TS bench with restored lock: **555 ms / 1 subprocess spawn / 100 results**. No measurable regression vs the no-lock version (localhost POST is sub-millisecond, lock contention is negligible, parallel fn() is preserved).
- sqlx offline cache regenerated.

rubenfiszel:
Final bench numbers with both SDKs verified against the code committed in \`8cd87d66fe\`. I'd been running benches against manually-patched copies in bun's and uv's package caches (because the worker resolves \`windmill-client\` / \`wmill\` from the public registry, not the local source). I just re-patched both caches to match exactly what's in this commit and re-ran.

### 100-step, debug build, localhost

| | Language | Wall | Per step | Subprocess spawns | Speedup |
|---|---|---|---|---|---|
| **Fast** | TS (bun) | **499 ms** | 5.0 ms | 1 | **11.1×** |
| Legacy | TS (bun) | 5530 ms | 55.3 ms | ~101 | |
| **Fast** | Python | **537 ms** | 5.4 ms | 1 | **17.1×** |
| Legacy | Python | 9194 ms | 91.9 ms | ~101 | |

Notes:

- Python's fast-path per-step is now on par with TS after switching to the reused \`_inline_http_client\` (was ~10 ms/step with the per-call \`async with httpx.AsyncClient\`). Saving the connection-pool setup across 100 calls is the bulk of it.
- Python gets a higher speedup than TS (17× vs 11×) because Python subprocess startup is more expensive than bun's. Legacy per-step is 92 ms vs 55 ms, while fast per-step is essentially identical.
- The restored SDK-side lock (\`asyncio.Lock\` for Python, \`_inlineChain\` for TS) adds zero measurable overhead — localhost POSTs are sub-millisecond, and \`fn()\` bodies still run in parallel across \`asyncio.gather\` / \`Promise.all\` (the lock wraps only the POST).
- Release build should push both speedups higher — debug-build axum/sqlx/tracing eats a big chunk of the fast-path constant; subprocess spawn cost (which dominates legacy) is nearly profile-independent.

Unit tests still green: Python 83/84 (same pre-existing \`@task __name__\` failure unrelated to this PR), TS 81/81. \`cargo check\` clean. sqlx offline cache regenerated.

PR #8839, opened 2026-04-15
docs: encourage subflow reuse in AI chat flow builder prompt

rubenfiszel:
## Summary
The AI chat flow builder already has the tools to discover existing flows in the workspace (`search_workspace`, `get_runnable_details`) and the flow JSON schema supports a `type: "flow"` module that references another flow by path as a subflow step. However, the system prompt only documented `type: "script"` and `type: "rawscript"`, so the model rarely reached for subflow reuse — it tended to rebuild equivalent logic inline unless the user explicitly asked.

This tightens the "Creating Flows" section of `frontend/src/lib/components/copilot/chat/flow/core.ts` to make subflow reuse a first-class option.

## Changes
- Updated step 1 ("Search for existing scripts first") to explicitly mention flows and call out that existing flows should be reused as subflow steps
- Added a bullet in step 2 showing the exact `type: "flow"` + `path` + `input_transforms` shape, with an example (`f/flows/process_user`), matching the `PathFlow` schema in `openflow.openapi.yaml`
- Minor phrasing tweaks so the guidance flows naturally

No tool or schema changes — prompt-only.

## Test plan
- [ ] Open the AI chat in the flow builder and ask it to build a flow that matches an existing workspace flow; verify it uses `search_workspace`, finds the flow, and wires it in with `type: "flow"` instead of reinventing the logic
- [ ] Verify normal script-based flow building still works unchanged

---

---
## Summary by cubic
Update the flow builder prompt to prefer reusing existing flows as subflow steps using `type: "flow"` with `path` and `input_transforms`. Add an eval case (`flow-test13-prefer-existing-workspace-flow`) and a loader test to verify the builder uses `search_workspace`/`get_runnable_details` to wire existing flows, not inline logic; no tool or schema changes.

<sup>Written for commit db8112ca439d2b08aa237a9d47791827cbd93e1c. Summary will update on new commits.</sup>

### DIFF
--- a/ai_evals/cases/flow.yaml
+++ b/ai_evals/cases/flow.yaml
@@ -47,6 +47,38 @@
     - "the main step is named `call_add_numbers`"
     - the parent flow delegates to an existing workspace subflow instead of inlining the addition logic
 
+- id: flow-test13-prefer-existing-workspace-flow
+  prompt: |-
+    Create a parent flow that adds two numbers by reusing an existing flow from the workspace if one fits.
+    A reusable script may also be available, but for this task prefer the existing flow rather than calling a script directly or rewriting the logic inline.
+    The parent flow should take `a` and `b` as inputs and use a single top-level step named `call_add_numbers_flow`.
+  initial: ai_evals/fixtures/frontend/flow/initial/test13_prefer_existing_workspace_flow_initial.json
+  expected: ai_evals/fixtures/frontend/flow/expected/test13_prefer_existing_workspace_flow.json
+  validate:
+    exactTopLevelStepIds:
+      - call_add_numbers_flow
+    topLevelStepTypes:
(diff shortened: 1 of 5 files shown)


centdix:
Added follow-up benchmark coverage for this PR after the prompt change.

Contribution from this branch:
- added `flow-test13-prefer-existing-workspace-flow` to `ai_evals/cases/flow.yaml`
- added fixtures that expose both a reusable script and a reusable flow in the workspace
- made the expected output require choosing the existing workspace flow specifically via a `type: "flow"` step
- added a loader test in `ai_evals/core/cases.test.ts`
- validated the new case with a targeted `ai_evals` run (`haiku`, 1 run), which passed

This gives the PR a stricter regression check for "prefer reusing an existing flow from the workspace" instead of only proving that subflow reuse is possible.

PR #8896, opened 2026-04-21
fix: make windmill-parser-wasm a self-contained workspace

hugocasa:
## Summary

`backend/parsers/windmill-parser-wasm` uses nightly-only cargo features (`cargo-features = ["panic-immediate-abort"]`, `-Z build-std`) that would break stable cargo on the whole backend workspace. It's therefore listed in `backend/Cargo.toml`'s `exclude = [...]`, which fully detaches it from the parent workspace. But its manifest still used `.workspace = true` for its package metadata and every dependency, so `cargo metadata` (and therefore `wasm-pack build`) from inside the crate would fail with:

```
error inheriting `edition` from workspace root manifest's `workspace.package.edition`
Caused by: failed to find a workspace root
```

This PR declares the crate as its own workspace, mirroring the relevant bits of the parent workspace so resolution stays semantically identical.

## Changes

- **`backend/parsers/windmill-parser-wasm/Cargo.toml`** — add `[workspace]`, `[workspace.package]`, `[workspace.dependencies]` tables. Version specs (`anyhow = "^1"`, `wasm-bindgen = "=0.2.103"`, etc.) and path targets (`../windmill-parser-sql`, etc.) match what the parent workspace exports, so `.workspace = true` inheritance resolves to the same values. Sibling parser crates remain members of the parent workspace and use its dep graph for their own deps.
- **`.github/change-versions.sh`** and **`.github/change-versions-mac.sh`** — add two `sed` calls: one for `[workspace.package].version` in the wasm `Cargo.toml`, one bulk-bumping every `windmill-*` entry in its `Cargo.lock`. Release-please-driven version bumps now keep this crate in sync automatically.
- **`frontend/package.json`** — bump the pinned `windmill-parser-wasm-regex` from `^1.670.0` to `1.688.0` to match the freshly-built package style.
- **`frontend/package-lock.json`** — regenerated via `npm i --package-lock-only --ignore-scripts`.
- **`backend/parsers/windmill-parser-wasm/Cargo.lock`** — regenerated from scratch under the new workspace. Large diff (~2500 lines), but the wasm binary's dep graph is unchanged: the added 180 entries come from sibling crate `windmill-parser-py-imports`'s `cfg(not(target_arch = "wasm32"))` deps (windmill-common, sqlx, axum, hyper, etc.), which are always recorded in a lockfile but never compiled when targeting wasm32.

## Why not make it a regular workspace member?

Tried. `cargo-features = ["panic-immediate-abort"]` at the top of a member's manifest is rejected by stable cargo — even with `default-members` excluding it from `cargo build`, the workspace still has to parse every member's manifest. A sub-workspace is the least-bad option Cargo currently permits.

## Test plan

- [ ] `cd backend/parsers/windmill-parser-wasm && cargo +nightly metadata --no-deps` succeeds and resolves `"version": "1.688.0"`, `"edition": "2021"`, correct authors.
- [ ] `cargo check` at the backend root still builds all workspace members.
- [ ] `sh build.nu regex` (and any other language target) produces a `pkg-*` wasm bundle of comparable size to a pre-change build.
- [ ] Dry-run of `.github/change-versions.sh 9.99.99` updates the wasm `Cargo.toml` and `Cargo.lock` correctly (verified locally: 24/24 `windmill-*` entries bumped).

---

---
## Summary by cubic
Make `backend/parsers/windmill-parser-wasm` a self-contained Cargo workspace so `cargo metadata`/`wasm-pack build` work inside the crate without breaking the backend’s stable workspace. Also update release scripts and align the frontend to `windmill-parser-wasm-regex@1.688.0`.

- **Bug Fixes**
  - Add `[workspace]`, `[workspace.package]`, and `[workspace.dependencies]` in the crate’s `Cargo.toml`, mirroring the parent workspace so `.workspace = true` resolves the same.
  - Fix “failed to find a workspace root” when running `cargo metadata`/`wasm-pack build` inside the crate.

- **Dependencies**
  - Update `.github/change-versions.sh` and `.github/change-versions-mac.sh` to bump the crate’s `[workspace.package].version` and all `windmill-*` entries in its `Cargo.lock`.
  - Bump `frontend` dependency `windmill-parser-wasm-regex` to `1.688.0` and regenerate `package-lock.json`.
  - Regenerate the crate’s `Cargo.lock` under the new workspace.

<sup>Written for commit 82e3c87d2dfc7c0c96219ffa7162f87c45d018ea. Summary will update on new commits.</sup>


hugocasa:
Combined into #8891 — both changes now live on branch `fix/sql-parser-dollar-quoted-strings`.

PR #8919, opened 2026-04-23
fix: correct flow conversation pagination

centdix:
## Summary
Fix flow conversation pagination by removing the broken conversation-list `after_id` behavior and replacing message polling with a monotonic message cursor. Conversation listing now uses only the supported filters, while message listing exposes `after_seq` so clients can request messages inserted after a known persisted message sequence.

## Changes
- Remove the unsupported `after_id` query handling from `list_conversations`, which previously compared a message timestamp against conversation creation time and produced incorrect results.
- Add a generated `created_seq` column on `flow_conversation_message`, plus an index on `(conversation_id, created_seq)`, to provide a stable monotonic cursor for message pagination.
- Update `list_conversation_messages` to support `after_seq` and order messages by `created_seq`, returning only messages with a larger sequence when the cursor is provided.
- Update the OpenAPI schema to document `after_seq` and include `created_seq` on `FlowConversationMessage`.
- Update frontend chat polling to send the last persisted `created_seq` as `afterSeq`, while keeping the existing user-message filter and duplicate guard.
- Use `temp-*` ids and `created_seq: 0` for local-only optimistic messages so they are never used as persisted cursors.
- Regenerate SQLx metadata for the changed queries.

## Test plan
- [x] Backend watcher rebuilt successfully and health check passed.
- [x] Verified cursor SQL behavior with rollback-only database checks.
- [x] Ran `npm run generate-backend-client`.
- [x] Ran `npm run check:fast`.
- [ ] Manual UI regression for flow chat polling.

---


centdix:
Addressed the relevant review items with `c966dbb8b5`:

- Removed `OFFSET` from the `after_seq` cursor branch so cursor polling returns the next messages after the sequence without skipping rows.
- Dropped the stale `idx_conversation_message_conversation_time` index in the `created_seq` migration.
- Recreated `idx_conversation_message_conversation_time` in the down migration for rollback symmetry.
- Repaired the local dev DB migration state after the migration file changed: removed the old applied migration row, reran the current migration, verified the checksum, and confirmed the backend watcher starts healthy.

Leaving the explicit backfill change out per discussion.

PR #8922, opened 2026-04-23
fix: slim app ai chat context

centdix:
## Summary
This PR makes app-mode AI chat context explicit instead of implicit.

Before this change, app mode would automatically include the currently selected frontend file or backend runnable in the model prompt just because it was open in the editor. It would also expose that selected editor state through a dedicated tool. In practice, that made app-mode prompts larger and less predictable than they needed to be.

We decided to change this for a few reasons:
- The current file/runnable was hidden prompt context. The AI could receive a large file just because it happened to be open, even when the user did not intend to provide it as context.
- The implicit selection overlapped with explicit `@` mentions and with existing file-reading tools, which meant the same file/runnable context could be duplicated in multiple ways.
- The UI implied that the current file/runnable was part of the active AI context, even though the better model is: if the user wants a file to matter, they should mention it explicitly.
- `get_selected_context` bypassed that explicit model by giving the AI another way to recover selected editor state directly.

The new behavior is:
- Inspector-selected elements and explicit code selections are still included automatically, because they are strong user-intent signals and are not as easy to reconstruct through tools.
- Frontend files and backend runnables are only part of AI context when the user explicitly mentions them with `@`.
- When those files/runnables are mentioned, we send only their identifiers and let the model fetch content on demand via `get_frontend_file` and `get_backend_runnable`.

This makes app-mode context cheaper, less surprising, and more aligned with explicit user intent.

## Changes
- Stop serializing the currently selected frontend file or backend runnable into app-mode user messages
- Keep implicit context only for inspector element selections and explicit code selections
- Serialize `@`-mentioned frontend files and backend runnables as identifiers only instead of embedding their full contents
- Remove the `get_selected_context` app-mode tool so selected editor state cannot bypass the explicit-context flow
- Remove stale current file/runnable badges and the old exclusion toggle/state from the app-mode chat UI
- Trim the remaining selected-context shape so it only carries data that is still actually used
- Keep focused regression coverage for the supported app-mode context behavior

## Test plan
- [x] `npm run test:unit -- --run src/lib/components/copilot/chat/app/core.test.ts`
- [x] `npm run check:fast`
- [x] `npm run check`

---

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -3,8 +3,6 @@
 	import { type Snippet } from 'svelte'
 	import {
 		CheckIcon,
-		Code2,
-		FileCode,
 		HistoryIcon,
 		Loader2,
 		MousePointer2,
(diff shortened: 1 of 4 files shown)


centdix:
Thanks for the review. I checked the follow-up suggestions and I don’t think they’re worth taking in this PR.

The main goal here was to stop hidden editor selection from becoming implicit model context. That part is now addressed: selected frontend/backend files are no longer injected into the prompt, the bypass tool is gone, and only explicit `@` mentions plus inspector/code-selection context remain.

For the `availableContext` note specifically: I’m intentionally not slimming those entries further right now. Even though file contents / runnable payloads no longer go into the model prompt, they still power the preview popovers for selected app context badges. Keeping the current in-memory shape avoids adding a separate lazy-hydration path or extra branching just to preserve that UX. In other words, the remaining heavier data is local UI state, not model-facing context, and that tradeoff feels acceptable for now.

I also don’t think the heading wording / extra in-product hint / `#each` key nits justify more churn in this PR. If we want to revisit app-mode discoverability or picker-memory optimization later, I’d rather do that as a focused follow-up than keep stretching this change.

PR #8926, opened 2026-04-23
feat: WM_TESTED_RUNNABLE env var + wildcards in test: annotation

hugocasa:
## Summary

Extends the CI test feature so a single test script can cover multiple runnables and branch on which one triggered it.

- **Wildcards in `test:` annotations.** `*` matches one path segment, `**` matches any depth. So `// test: script/u/team/**` covers every script in a folder tree.
- **New `WM_TESTED_RUNNABLE` env var** on CI test jobs, containing `{kind}/{path}` of the runnable that triggered the test — mirrors the annotation syntax and lets the test script branch on its target.

## Changes

- New helper `ci_test_path_matches` in `windmill-common/src/schema.rs` — compiles glob patterns to anchored regexes with a bounded `quick_cache` LRU. Literal (no-wildcard) patterns short-circuit to string equality.
- New migration adds a Postgres `GENERATED ALWAYS` `has_wildcard` column + partial index on `ci_test_reference`, so exact-match lookups keep using the primary index and only wildcard rows are scanned for regex matching.
- `ci_tests_ee::trigger_ci_tests_for_item` splits into exact + wildcard queries. The wildcard query uses the new partial index; results are filtered in Rust against the compiled regex. The job's `trigger` column and debounce key now encode `{kind}/{path}` so the worker can expose it as `WM_TESTED_RUNNABLE` and deploys of different targets don't collapse into a single debounced trigger.
- `get_ci_test_results` / `get_ci_test_results_batch` apply the same exact + wildcard split. Batch endpoint now issues O(kinds) queries instead of O(items) — collapsing up to 200 per-item subqueries to 2–6 total.
- Worker's env-var injection (`windmill-worker/src/common.rs`) passes through a new `tested_runnable` parameter derived from `(job.trigger_kind, job.trigger)`; `variables::get_reserved_variables` emits `WM_TESTED_RUNNABLE`.
- CI test templates (bun, python) documented with wildcard examples and env-var usage.

Companion: windmill-labs/windmill-ee-private#546

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test ci_tests` — 3/3 pass, including new `test_ci_test_wildcard_annotation` covering storage flag, multi-target match, and single-segment `*` semantics.
- [x] `cargo test -p windmill-common --lib schema::tests::ci_test_path_matches` — 6/6 unit tests pass (literal, single/double star, regex metachars, underscore literal, boundary cases).
- [x] `./update_sqlx.sh` — full workspace compiles clean.
- [ ] Manual end-to-end: create a test script with `// test: script/u/admin/*`, deploy two sibling scripts, confirm both trigger the test and each job logs its `WM_TESTED_RUNNABLE`.
- [ ] Manual end-to-end: deploy a deeper path (`u/admin/sub/deep`), confirm `*` does not match, switch annotation to `**`, confirm it does.

## Testing via UI

Open the script editor, create a new Bun or Python script, and at the very top add `// test: script/u/<folder>/*` (or `# test:` for Python). Save it. Then create or redeploy a sibling script matching that glob — within a few seconds the sibling's page should show a "CI tests" row linking to the freshly queued test job. Open that job and confirm the logs include the expected `WM_TESTED_RUNNABLE` value.

---
## Summary by cubic
Adds wildcard support to CI test annotations and exposes `WM_TESTED_RUNNABLE` so one test script can cover multiple runnables. Also scopes CI test job lookups by trigger and speeds up queries for faster, accurate results.

- **New Features**
  - `test:` supports globs: `*` (one segment), `**` (any depth).
  - CI test jobs include `WM_TESTED_RUNNABLE` with `{kind}/{path}` of the triggering runnable; also available in resource interpolation.
  - Updated Bun/Python CI test templates with wildcard examples and env var usage.
  - Added `ci_test_path_matches` for anchored, cached pattern matching.

- **Performance**
  - New Postgres `has_wildcard` generated column + partial index to scan only wildcard rows.
  - Exact and wildcard queries split in triggers and `ci_test_results`/`ci_test_results_batch`; batch now issues one query per kind.
  - Job lookup in results endpoints is scoped by `v2_job.trigger` so multi-target tests return the correct job per target.

<sup>Written for commit 592c4bbc09b4f07823759a299070535c18e4b6af. Summary will update on new commits.</sup>


hugocasa:
/docs update the CI test docs with wildcard support

Issue #8947, opened 2026-04-27
bug: Preprocessor didn't work with workflow-as-code scripts


progamesigner:
### Describe the bug

The workflow-as-code didn't support preprocessor. Is that intended? We just receive `event` and handle preprocessor work and our flow logic in main function?

### To reproduce

Nope.

### Expected behavior

Should be handled as document, or need to up date document about workflow-as-code.

Issue #8946, opened 2026-04-27
bug: The HTTP trigger configuration didn't list workflow-as-code scripts


progamesigner:
### Describe the bug

The dropdown of script selection didn't show workflow-as-code scripts. The API (`/api/w/<workspace>/scripts/list?kinds=script&without_description=true`) didn't include these scripts also.

### To reproduce

1. Create a workflow-as-code script
2. Click "deploy" directly
3. Go to HTTP trigger and add new route
4. Unable to select the newly deployed script

### Expected behavior

It should include WAC scripts in dropdown. Currently it can only create routes from settings -> triggers panel in the script itself.

Issue #8945, opened 2026-04-27
bug: Workflow-as-Code Python template didn't be recognized as "wac" auto kind


progamesigner:
### Describe the bug

The workflow-as-code Python template didn't be recognized correctly and have no "was" label on table rows. Also checked the rows in `script` table, the `auto_kind` is `NULL` value.

### To reproduce

1. Create a workflow-as-code with python
2. Click "deploy" directly
3. The home script didn't include correct "wac" label

### Expected behavior

It should be recognized as workflow-as-code with "wac" label showing.

### Screenshots

The Python template shows as normal script, while TS one has label of "wac"
![](https://github.com/user-attachments/assets/2d8a08c2-d838-4a6b-abcf-df740051d481)

### Additional Context

If change the `main` function to other (such as `test`) and it will show correct label:

```python
# ... all other same as Python template
@workflow
async def test(x: str):
# ...
```

PR #8951, opened 2026-04-27
fix(wac): recognize @workflow main, list WAC in scripts/list, run preprocessor

rubenfiszel:
## Summary
Fixes the three latest workflow-as-code bug reports:

- **#8945** — Python WAC template (`@workflow async def main(...)`) was not detected as `auto_kind = \"wac\"`. The detection only ran when no `main` function was found, so the standard template silently fell back to `None`. Hoisted the heuristic so it runs regardless of whether `main` is the entrypoint.
- **#8946** — `GET /api/w/{ws}/scripts/list?kinds=script` excluded WAC scripts because the query filtered `auto_kind IS NULL` to hide library scripts. WAC scripts have `auto_kind = 'wac'` and are runnable, so they should be returned. Changed the filter to allow both `NULL` and `'wac'`, still excluding `'lib'`.
- **#8947** — Preprocessors defined alongside a WAC workflow were never invoked. Injected the preprocessor call into the Python WAC wrapper (gated by `apply_preprocessor` exactly like the regular wrapper), and plumbed the preprocessed args through `handle_wac_v2_output` so inline child re-runs see the post-preprocessor args via `checkpoint.input_args`.

## Test plan
- [x] New unit test `test_parse_python_wac_main_decorator` in `windmill-parser-py` (passes)
- [x] Existing `windmill-parser-wac` tests still pass
- [x] `cargo check -p windmill-worker --features=private` clean
- [x] `cargo check -p windmill-api-scripts` clean
- [ ] Manually deploy a Python WAC script with `@workflow async def main` — confirm \"wac\" label appears
- [ ] Manually create an HTTP trigger and confirm the WAC script appears in the dropdown
- [ ] Manually run a WAC script triggered by a webhook with a preprocessor that maps event payload → workflow args

---
## Summary by cubic
Fixes WAC detection and behavior: detect Python workflows with `@workflow` main (no `@task` needed), include WAC scripts in `scripts/list`, and run preprocessors before workflows with preprocessed args safely propagated to child runs.

- **Bug Fixes**
  - #8945: Python parser now flags WAC when `@workflow` and `wmill` imports are present, even with `async def main` and without `@task`. Adds a step-only unit test.
  - #8946: `GET /api/w/{ws}/scripts/list?kinds=script` now includes WAC and other runnable scripts by excluding only `'lib'` (`auto_kind <> 'lib'`). Adds an integration test.
  - #8947: Python WAC wrapper runs `preprocessor` inside the `try:` block and supports coroutine preprocessors via `asyncio.run`. Both Python and `bun` pass preprocessed args to `handle_wac_v2_output`, which seeds `checkpoint.input_args` and surfaces JSON parse errors. Adds an integration test.

<sup>Written for commit 3a9b71c823d38ce8310eaf94e9cec4c1a85f7c1a. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/8951?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/parsers/windmill-parser-py/src/lib.rs
+++ b/backend/parsers/windmill-parser-py/src/lib.rs
@@ -351,8 +351,11 @@ pub fn parse_python_signature(
         }
     };
 
+    // Mirror the runtime detector `is_wac_v2_py` in `windmill-worker/src/wac_executor.rs`:
+    // a wmill import + an `@workflow` decorator is sufficient. `@task` is optional —
+    // a workflow that only uses inline `step()` calls still goes through the WAC runner,
+    // and labelling it as `wac` here is what aligns the editor badge with execution.
     let is_wac_v2 = (code.contains("@workflow") || code.contains("workflow("))
-        && (code.contains("@task") || code.contains("task("))
         && (code.contains("import wmill") || code.contains("from wmill"));
 
     // Check if main function was found
@@ -1028,6 +1031,23 @@ def main(): return
             }
         );
 
+        Ok(())
+    }
+
+    #[test]
+    fn test_parse_python_wac_step_only() -> anyhow::Result<()> {
+        // A WAC script that uses inline step() instead of @task should still be
+        // detected as auto_kind = "wac" — the runtime path treats @task as
+        // optional, and the editor badge needs to match.
+        let code = r#"
+from wmill import workflow, step
+
+@workflow
+async def main(x: str):
+    return await step("k", lambda: x.upper())
+"#;
+        let sig = parse_python_signature(code, None, false)?;
+        assert_eq!(sig.auto_kind, Some("wac".to_string()));
         Ok(())
     }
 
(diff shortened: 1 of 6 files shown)


rubenfiszel:
Addressed in 3a9b71c:

**Fixed**
- `asyncio.get_event_loop()` → `asyncio.run()` (deprecated in 3.10+)
- WAC preprocessor moved inside the wrapper's `try:` block so failures serialize to `result.json` (cubic P1)
- `unwrap_or(Value::Null)` → `?` so corrupted preprocessor JSON surfaces instead of silently persisting
- Parser WAC heuristic dropped the `@task` requirement to match `is_wac_v2_py` runtime detector (added `test_parse_python_wac_step_only`)
- `auto_kind IS NULL OR = 'wac'` → deny-list `<> 'lib'` so future `auto_kind` values aren't silently filtered

**Not addressed**
- `{transforms}` runs against the preprocessor's signature before the preprocessor — that's the same behavior as the regular wrapper and would change pre-existing semantics if touched. Left as-is.
- `prepare_wrapper`'s 12-tuple return type — agreed it would scale better as a struct, but out of scope for this fix.

PR #8960, opened 2026-04-28
perf: optimize datatable app chat schemas

centdix:
## Summary
Optimize app chat datatable tooling so chat mode no longer loads every column for every table up front. The PR adds targeted backend endpoints for table metadata and single-table schema lookup, then rewires the app chat helpers and context selection flow to load table columns only when needed.

## Changes
- Add workspace API endpoints for datatable table metadata and one-table schema lookup.
- Rewire app chat datatable tools to use targeted metadata/schema helpers instead of fetching full schemas and filtering locally.
- Make app datatable context column loading lazy when a table is selected.
- Refresh cached app datatable context after datatable SQL execution or whitelist updates.
- Preserve datatable-level all-tables refs in app chat context filtering.
- Prevent duplicate app context entries while eager-loading datatable table schemas.
- Update datatable chat and whitelist tests for the new helper contract.

## Test plan
- [x] cargo fmt --manifest-path backend/Cargo.toml --package windmill-api-workspaces
- [x] npm run generate-backend-client
- [x] npm run check:fast
- [x] npx vitest run src/lib/components/copilot/chat/app/core.test.ts src/lib/components/raw_apps/dataTableRefUtils.test.ts
- [x] npm run check
- [x] backend cargo-watch logs show healthy server after rebuild

---

### DIFF
--- a/backend/windmill-api-workspaces/src/workspaces.rs
+++ b/backend/windmill-api-workspaces/src/workspaces.rs
@@ -115,6 +115,11 @@ pub fn workspaced_service() -> Router {
         .route("/list_ducklakes", get(list_ducklakes))
         .route("/list_datatables", get(list_datatables))
         .route("/list_datatable_schemas", get(list_datatable_schemas))
+        .route("/list_datatable_tables", get(list_datatable_tables))
+        .route(
+            "/get_datatable_table_schema",
+            get(get_datatable_table_schema),
+        )
         .route("/edit_datatable_config", post(edit_datatable_config))
         .route("/git_sync_enabled", get(get_git_sync_enabled))
         .route("/edit_git_sync_config", post(edit_git_sync_config))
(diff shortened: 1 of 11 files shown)


centdix:
Addressed the relevant review comments in `2e4967270b`:

- Whitelist the selected datatable table before eager-loading its schema, so the frontend guard no longer rejects newly selected tables.
- Surface eager schema-load failures with `sendUserToast` in addition to `console.error`.
- Re-check selected context after the awaited schema fetch to avoid duplicate entries from rapid selection.
- Preserve datatable-level `(all tables)` refs when filtering available tables and validating schema lookup access.
- Added focused tests for the datatable whitelist helper behavior.
- Added the `pg_*` system-schema comment and updated the PR body to remove the inaccurate no-TTL cache wording.

I left the serial per-datatable connection behavior in `list_datatable_tables` as a follow-up because it mirrors existing `list_datatable_schemas` behavior and is not a regression in this PR.

hugocasa (approved):

PR #8961, opened 2026-04-28
feat: add ai chat schedule and trigger tools

centdix:
## Summary
Add AI chat tools for creating schedules and triggers from script and flow chat modes. The tools are intentionally scoped to the current deployed script/flow, validate model arguments with generated Zod schemas from the backend OpenAPI/system prompt generation pipeline, and return explicit user-visible tool results instead of generic tool failures.

## Changes
- Add shared `create_schedule` and `create_trigger` workspace mutation tools used by both script AI chat and flow AI chat.
- Generate `workspaceToolsZod.ts` from `system_prompts/generate.py` using backend OpenAPI schemas, so schedule/trigger argument validation stays aligned with API schema changes.
- Normalize the generated trigger tool schema for Claude compatibility by avoiding top-level `oneOf`/`allOf`/`anyOf` composition in the tool input schema.
- Hide `script_path` and `is_flow` from the LLM-facing tool schema; infer the current deployed script/flow target from chat helpers and inject those fields before Zod validation and API calls.
- Fail before confirmation when the current script/flow is not deployed, with explicit messages like `the script needs to be deployed before doing this action`.
- Validate schedule arguments with the generated schedule schema and call schedule preview before creating the schedule.
- Validate trigger arguments with generated schemas for HTTP, WebSocket, Kafka, NATS, Postgres, MQTT, SQS, GCP, and Azure trigger payloads.
- Surface explicit create/validation errors from the new tools to both the user-visible tool card and the LLM, instead of showing only `An error occurred while calling the tool`.
- Show structured success details in the tool card result (`path`, `target_path`, `target_kind`, `backend_result`) instead of only the raw backend response string.
- Clarify OpenAPI path descriptions for schedules/triggers so the LLM sees that object paths must be of the form `u/<user>/<path>` or `f/<folder>/<path>`, and that trigger `path` is distinct from HTTP `route_path`.
- Add frontend AI eval cases for creating schedules and HTTP triggers from current scripts and flows, including assertions that catch invalid object paths such as bare names or `schedules/...`.

## Decisions
- Started with `create_schedule` and `create_trigger` only; resources and variables are left for a later step.
- The model still chooses the schedule/trigger object `path`, but the schema now explicitly documents the full `u/<user>/<path>` / `f/<folder>/<path>` shape. The current runnable target (`script_path`, `is_flow`) is not model-controlled.
- Used generated Zod schemas instead of hand-written validators so the tool argument schema follows the backend OpenAPI/system prompt generation source of truth.
- Kept target deployment checks before confirmation so the UX does not ask users to confirm an action that cannot succeed.
- Kept error handling local to the two new tools, rather than changing generic tool error behavior for every AI chat tool.
- For direct AI eval transport, schedules/triggers use focused mocks that validate the model-generated paths without persisting real backend objects. Real backend endpoint coverage for proxy transport was discussed but intentionally left out of this change.

## Test plan
- [x] `cd frontend && npm run generate-backend-client`
- [x] `uv run --with pyyaml python system_prompts/generate.py`
- [x] `uv run --with pyyaml bash system_prompts/check-freshness.sh`
- [x] `cd frontend && npm run test:unit -- src/lib/components/copilot/chat/shared.test.ts --run`
- [x] `cd frontend && npm run check:fast`
- [x] `cd ai_evals && bun test core/validators.test.ts core/cases.test.ts`
- [x] `cd ai_evals && bun run cli -- cases script`
- [x] `cd ai_evals && bun run cli -- cases flow`
- [x] `git diff --check`

### Proxy AI evals

All proxy evals used `WMILL_AI_EVAL_BACKEND_URL=http://127.0.0.1:8030`, `WMILL_AI_EVAL_BACKEND_WORKSPACE=admins`, and `--transport proxy`. A first run without `WMILL_AI_EVAL_BACKEND_WORKSPACE` failed before any model call because the local CE instance refused creating additional workspaces.

- [x] `cd ai_evals && bun run cli -- run script script-test2-create-current-script-schedule script-test3-create-current-script-http-trigger --models haiku,4o,gemini-flash --transport proxy`
  - Anthropic `haiku`: 2/2 pass
  - OpenAI `4o`: 2/2 pass
  - Google `gemini-flash`: 2/2 pass
- [x] `cd ai_evals && bun run cli -- run script script-test1-greet-user --models haiku,4o,gemini-flash --transport proxy`
  - Anthropic `haiku`: pass
  - OpenAI `4o`: pass
  - Google `gemini-flash`: pass
- [x] `cd ai_evals && bun run cli -- run flow flow-test15-create-current-flow-schedule flow-test16-create-current-flow-http-trigger --models haiku,4o,gemini-flash --transport proxy`
  - Anthropic `haiku`: 1/2 pass; failed the HTTP trigger eval because it used bare trigger path `order_processing_webhook` instead of `f/...` or `u/...`
  - OpenAI `4o`: 2/2 pass
  - Google `gemini-flash`: 0/2; did not call the flow/tools in either case
- [x] `cd ai_evals && bun run cli -- run flow flow-test0-sum-two-numbers --models haiku,4o,gemini-flash --transport proxy`
  - Anthropic `haiku`: pass
  - OpenAI `4o`: pass
  - Google `gemini-flash`: fail with no tool calls and an empty flow
- [x] `cd ai_evals && bun run cli -- run flow flow-test0-sum-two-numbers flow-test15-create-current-flow-schedule flow-test16-create-current-flow-http-trigger --model gemini-pro --transport proxy`
  - Google `gemini-pro`: 3/3 pass

Backend logs showed `/api/w/admins/ai/proxy/chat/completions` returning `200` during the proxy evals, with backend health checks healthy.

## Notes
- `cd ai_evals && bun run typecheck` currently fails on unrelated pre-existing TypeScript errors under `../cli`; this PR did not address those.
- The path-format guidance is currently schema description guidance, not a new backend hard validation rule.
- The proxy eval failures observed here are model behavior, not transport or schema failures: Haiku can still choose an invalid trigger object path, and Gemini Flash is unreliable for flow mode over proxy. Gemini Pro passed the same flow baseline and new schedule/trigger flow cases.

---


centdix:
Follow-up on the review comments:

Addressed in `a9ad9040d8`:
- Guarded script workspace mutation targets so inline flow step ids like `step_a` cannot be used as `script_path` for `create_schedule` / `create_trigger`.
- Centralized workspace target validation so pre-confirmation validation and runtime target injection return the same explicit deployed-script/flow error.
- Made runtime target injection reject undeployed targets, not only missing paths.
- Improved API error formatting to prefer useful nested backend messages such as `body.error.message` before falling back to JSON/string output.
- Made workspace tool Zod generation fail loudly on unsupported `oneOf` / `allOf` instead of silently generating permissive schemas.

Intentionally not addressed in this follow-up:
- No tests added yet, per discussion.
- Schedule `args` validation against the target script/flow schema is left as a follow-up because it needs a broader schema-resolution path for both scripts and flows.
- Trigger config discriminated-union error UX is left as a follow-up; the per-kind validation still catches malformed payloads after dispatch.
- Per-trigger request-schema `path.min(1)` was not changed because the LLM-facing wrapper schema already enforces non-empty path.
- Script/flow deployed predicate naming was not refactored because it is readability-only and not a behavior bug.

hugocasa (approved):

Issue #8939, opened 2026-04-26
bug: windmill CLI: wmill pull normalizes backend runnable filenames to lowercase, breaking app


Cratylel:
### Describe the bug

When pulling backend runnables (`wmill pull`) for a full-code app, the actual files containing the code get normalized to lowercase. For example, `f/ui_apps/test_app__raw_app/backend/CamelCaseTSRunnable.ts` becomes `f/ui_apps/test_app__raw_app/backend/camelcasetsrunnable.ts`.

Furthermore, some sort of metadata seems to pull correctly, so when pushing to another workspace (`wmill push`), after entering the app editor, we now see duplicate backend runnables:
- CamelCaseTSRunnable: correct name, no lanugage chosen, empty
- camelcasetsrunnable: incorrect name, typescript, contains the correct runnable code

Version info:

- Windmill instance: CE v1.674.2 (selfhosted)
- wmill CLI: 1.691.0 (latest)

### To reproduce

1. Create full code app in windmill
2. Add a backend runnable that contains capital letters in the file name
3. Use `wmill pull`
4. File name normalized to lowercase

### Expected behavior

`wmill pull` should have created a local file `f/ui_apps/test_app__raw_app/backend/CamelCaseTSRunnable.ts`

### Screenshots

Correctly named runnable: (empty)
<img width="765" height="444" alt="Image" src="https://github.com/user-attachments/assets/aa66b6de-cd6c-410c-910c-c9c5630c405f" />
Incorrectly named runnable (normalized, contains actual code)

<img width="712" height="455" alt="Image" src="https://github.com/user-attachments/assets/16adb792-35fd-4aad-9da3-f186fa9f7a33" />

PR #9000, opened 2026-05-01
fix(cli): only preserve case for raw-app runnableIds, not app/flow summaries

rubenfiszel:
## Summary

- Limit the case preservation introduced in #8940 to raw-app runnableIds. Normal-app and flow inline scripts derive their on-disk path from the script's human summary (e.g. \"Get Users Data\"), which used to lowercase to \`get_users_data.inline_script.ts\` and now landed as \`Get_Users_Data.inline_script.ts\`. That changed both the on-disk filename and the \`!inline …\` reference inside \`app.yaml\` / \`flow.yaml\`, surfacing as unwanted case churn after users updated to 1.693.x.
- The original #8939 bug was raw-app-specific: the YAML metadata file is named after the runnableId (case preserved) so the code file must also preserve case. That asymmetry doesn't exist for normal apps / flows, so lowercasing is safe and matches pre-#8940 behavior.
- Implementation: add a \`preserveCase\` option to \`sanitizeForFilesystem\`. \`newRawAppPathAssigner\` opts in; \`newPathAssigner\` stays on the default (lowercase).
- Tests: revert three \`Get_Users_Data\` / \`Step_B\` assertions back to lowercase in \`inline_scripts_failure_preprocessor_unit.test.ts\`. Update the \`sync_pull_push.test.ts\` path-assigner test to pin both halves of the contract (newPathAssigner lowercases, newRawAppPathAssigner preserves). Add a new e2e test in \`raw_app_sync.test.ts\` that pushes a \`CamelCaseTSRunnable\` backend runnable, pulls it back, and asserts both YAML and code file preserve case with no lowercase orphan — the missing reproduction for #8939.

## Test plan
- [x] \`bun test test/sync_pull_push.test.ts\` — 67 pass / 3 skip
- [x] \`bun test test/inline_scripts_failure_preprocessor_unit.test.ts\` — 24 pass
- [x] \`bun test test/raw_app_sync.test.ts\` — 5 pass (4 existing + 1 new CamelCase round-trip)

---
## Summary by cubic
Restore lowercasing for app/flow inline script paths while preserving case only for raw-app backend runnableIds. This stops filename churn in `app.yaml`/`flow.yaml` and keeps raw-app YAML/code files in sync to avoid duplicates.

- **Bug Fixes**
  - Added `preserveCase` option to `sanitizeForFilesystem` (default: lowercase).
  - `newRawAppPathAssigner` opts in to `preserveCase`; `newPathAssigner` continues to lowercase.
  - Tests: reverted inline-script path assertions to lowercase, pinned assigner behavior, added CamelCase raw-app round-trip e2e, and switched to `readdir` for exact-case orphan checks on Windows.

<sup>Written for commit 5374a7810f7a1abdee9139cd5b8d8015ad4fe3d6. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9000?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/cli/test/inline_scripts_failure_preprocessor_unit.test.ts
+++ b/cli/test/inline_scripts_failure_preprocessor_unit.test.ts
@@ -524,7 +524,7 @@ describe("extractInlineScripts with mapping preserves file paths", () => {
 
     const contentScript = scripts.find((s) => !s.is_lock);
     // Should use assigner path based on summary, not mapped
-    expect(contentScript!.path).toContain("Get_Users_Data");
+    expect(contentScript!.path).toContain("get_users_data");
   });
 
   test("mapped modules and unmapped modules coexist", () => {
@@ -538,7 +538,7 @@ describe("extractInlineScripts with mapping preserves file paths", () => {
 
     const paths = scripts.filter((s) => !s.is_lock).map((s) => s.path);
     expect(paths[0]).toBe("my_custom_name.ts");
-    expect(paths[1]).toContain("Step_B"); // assigner-generated from summary
+    expect(paths[1]).toContain("step_b"); // assigner-generated from summary
   });
 
   test("lock path is derived from mapped content path", () => {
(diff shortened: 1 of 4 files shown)


rubenfiszel:
/ai on windows
1 tests failed:
(fail) Raw App: CamelCase backend runnableId round-trips without duplicates [1562.00ms]

can you fix

PR #9064, opened 2026-05-06
fix(concurrency): two-phase admit to skip FOR UPDATE on over-limit pulls

rubenfiszel:
## Summary

Restructures `update_concurrency_counter` (per-script concurrency limiter) into a two-phase admit so that doomed-to-reject pulls no longer serialize through one Postgres row lock. Eliminates the DB lock storm that hits when many parallel workers race a low `concurrent_limit` (e.g. `branch_all { parallel: 20 }` against `concurrent_limit: 5`) — a real customer was burned by this for 12 hours after `push_scheduled_job` started timing out.

EE companion: windmill-labs/windmill-ee-private#563

## The fix

The original code took `SELECT ... FOR UPDATE` on `concurrency_counter` on **every** pull check, regardless of outcome. Under contention, all N workers serialize 1-at-a-time on one row, even though most of them will be rejected.

New shape:

- **Phase 1 (lock-free)**: read `running` count and `completed`-window count without `FOR UPDATE`. If `completed + running >= limit`, return `(false, max_ended_at)` immediately. This is the storm path.
- **Phase 2 (locked admit)**: only when Phase 1 says we *might* admit, take `BEGIN + SELECT FOR UPDATE` and re-read **both** counts under the lock before INSERT-ON-CONFLICT. The admit decision is made from values read under FOR UPDATE — Phase 1's stale numbers are used only for the early-bail.

`completed_count` is re-read in Phase 2 (rather than reused from Phase 1) because the windowed `concurrency_key` count can drift between phases — concurrent workers can commit new completions (raising it) and old `ended_at` rows can fall out of the sliding window (lowering it). Phase 1's value isn't a tight bound either way, so reusing it in the admit decision could over-admit. Reading under FOR UPDATE in Phase 2 is what makes the decision tight.

The result: under storm conditions, FOR UPDATE acquisitions go from `O(parallel_workers)` per pull cycle to `O(limit)` — bounded by how many can actually admit, not by how many are trying.

## Why it's safe

- **Never-over-admit invariant preserved**: Phase 2's FOR UPDATE recheck is byte-equivalent to the legacy admit decision.
- **Worst-case race is a false reject**: a job finishing between Phase 1 and Phase 2 could cause an unnecessary requeue. The requeue path already handles this exact case in production today (it's the same as today's race between sibling pulls).
- **Inherits, does not introduce, the first-row race**: when the `concurrency_counter` row doesn't exist, `FOR UPDATE` on it locks nothing in the legacy code too (verified by stress test against legacy semantics). Out of scope to fix here.

## Tests added

`backend/windmill-queue/tests/concurrency_counter_test.rs` (5 tests):

1. **`stress_never_over_admits`** — 50 workers contend for `limit=5` on a pre-seeded row → admitted count is exactly 5. Ran 20× in a row, all green.
2. **`bail_path_no_admits_when_full`** — row pre-filled to `limit`, 20 concurrent attempts → 0 admit, counter unchanged.
3. **`first_job_race_all_admit`** — 10 concurrent attempts on missing row, `limit=10` → all admit (regression guard for the all-pull-through path).
4. **`retry_not_double_counted`** — same `pulled_job_id` already in `job_uuids` is excluded from the count (retry semantics).
5. **`max_ended_at_returned`** — verifies `max_ended_at` plumbing from `concurrency_key` to the caller.

Run: `cargo test -p windmill-queue --test concurrency_counter_test --features private,enterprise`

## Out of scope

- The first-row admit race (preexisting; needs schema or admit-as-CTE rework).
- Schema change (`concurrency_counter.job_uuids` jsonb → per-job table) — bigger refactor.
- Other queries in `apply_concurrency_limit` (`min_started_at`, `avg_script_duration`, `all_jobs`).


rubenfiszel:
Addressed CI review:

- **Codex P1 / Claude P2** — added a one-paragraph doc comment to `update_concurrency_counter` clarifying it's an internal queue primitive, not an authorization boundary, and must only be invoked by the worker pull loop. Since EE #563 was already merged, this lives as a follow-up: windmill-labs/windmill-ee-private#564 (will bump `ee-repo-ref.txt` here once it merges).
- **Claude nit on PR body wording** — fixed the "monotonically growing" claim. The windowed `concurrency_key` count can in fact also *decrease* as `now()` advances and old `ended_at` rows fall out of the window. Phase 1's value isn't a tight bound either way, so re-reading under FOR UPDATE in Phase 2 is what guarantees correctness. Body updated.

No code change in this PR; all source changes are in EE companion #564.

PR #9073, opened 2026-05-07
feat: parse windmill_failure field to tag run as failure

### LINKED ISSUE #9046: feature: return custom http response for a webhook while still failing the execution
I am not sure whether title is explanatory but here is what I was trying to achieve. I have a script which has a webhook trigger. I am calling the webhook to receive a response. Now the script can produce a failed output for user to consume, for example:
```json
{"message": "fail", "detail": "something"}
```
note that the http status code for the output is still a failure like 4xx/5xx

but for someone checking the windmill they are seeing that the script executed successfully meaning it returned the response but it is misleading.  I cant differentiate between a successful successful response and failing successful response without actually seeing the output of the execution. I couldnt find a way to configure windmill so that script actually is seen as failed and still I returned whatever response I wanted for the user

hugocasa:
## Summary
Alongside `wm_labels`, the worker now parses an optional `windmill_failure: string` field from a job's result. When present, the run is stored with `status='failure'` even though the worker reported success — same code path as a regular runtime error (routed through `add_completed_job_error`, so the result is wrapped as `{ error: <original_result> }` and downstream error handlers fire normally).

## Changes
- `backend/windmill-queue/src/jobs.rs`: rename `ResultLabels` → `ResultMetadata`, make both `wm_labels` and `windmill_failure` optional fields, add `windmill_failure(&self) -> Option<String>` to the `ValidableJson` trait with implementations for `WrappedError`, `Box<RawValue>`, `Arc<T>`, `serde_json::Value`, and `Json<T>`.
- `backend/windmill-worker/src/result_processor.rs`: in `process_jc`, if the worker reported success but the result contains `windmill_failure`, flip `jc.success = false` so the run takes the failure path.

Scope is intentionally limited to the worker's main job-completion path (`process_jc`). Direct `add_completed_job` callers in `ai/tools.rs` and `worker_flow.rs` are not touched.

## Test plan
- [x] Manually ran a Bun script returning `{ windmill_failure: "Boom!", data: "hello" }` — stored as `failure` with result `{"error": {"data": "hello", "windmill_failure": "Boom!"}}`
- [x] Manually ran a Bun script returning `{ data: "hello" }` — still `success`
- [x] Manually ran a Bun script returning `{ wm_labels: ["hello"], data: "x" }` — still `success` and label applied (no regression)

---
## Summary by cubic
Parse an optional `wm_failure` field in job results to force the run into the failure path while keeping top‑level fields for webhooks. We inject `error: { name: "ManualFailure", message }`, store pre‑shaped results as‑is, and the init‑script killpill now respects this override.

- **New Features**
  - In `process_jc`, when `wm_failure` is present: inject `ManualFailure`, keep sibling fields (e.g., `windmill_status_code`), flip `success = false`, and return the final success to trigger the init‑script killpill; store as‑is via `add_completed_job_pre_shaped_failure` and avoid double wrapping with `is_pre_shaped_wm_failure_result`.
  - Parse `ResultMetadata` for `wm_labels` and `wm_failure`; accept `wm_status_code`, `wm_content_type`, and `wm_headers` as aliases in `WindmillCompositeResult`.

- **Bug Fixes**
  - For OTel spans, prefer the injected `ManualFailure` over any sibling `name`/`message` fields when recording `error.name` and `error.message`.

<sup>Written for commit 21e4f2f6845b5e10cde3df05f80c54b60d30f77c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/windmill-api-jobs/src/execution.rs
+++ b/backend/windmill-api-jobs/src/execution.rs
@@ -248,8 +248,11 @@ lazy_static::lazy_static! {
 
 #[derive(Deserialize)]
 pub struct WindmillCompositeResult {
+    #[serde(alias = "wm_status_code")]
     windmill_status_code: Option<u16>,
+    #[serde(alias = "wm_content_type")]
     windmill_content_type: Option<String>,
+    #[serde(alias = "wm_headers")]
     windmill_headers: Option<HashMap<String, String>>,
     result: Option<Box<RawValue>>,
 }
--- a/backend/windmill-queue/src/jobs.rs
+++ b/backend/windmill-queue/src/jobs.rs
@@ -630,12 +630,14 @@ pub struct WrappedError {
 pub trait ValidableJson {
     fn is_valid_json(&self) -> bool;
     fn wm_labels(&self) -> Option<Vec<String>>;
+    fn windmill_failure(&self) -> Option<String>;
     fn size(&self) -> usize;
 }
 
 #[derive(serde::Deserialize)]
-struct ResultLabels {
-    wm_labels: Vec<String>,
+struct ResultMetadata {
+    wm_labels: Option<Vec<String>>,
+    windmill_failure: Option<String>,
 }
 
 impl ValidableJson for WrappedError {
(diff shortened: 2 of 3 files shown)


hugocasa:
/docs update windmill_* and similar with the new alias wm_*

PR #9080, opened 2026-05-08
fix(cli-tests): stabilize flow lock-gen race + Windows path

rubenfiszel:
## Summary

The latest main CLI Tests run ([25536061370](https://github.com/windmill-labs/windmill/actions/runs/25536061370)) failed with three tests, all of which are flakes (history shows 3 of the 5 failing main runs in the last 50 hit one of the same tests):

- `Mixed Case Paths: pull and push flow with capitalized folder` (linux) — also failed on 2026-05-01.
- `Integration: Mixed scripts and flows with nonDottedPaths are idempotent` (linux) — also failed on 2026-04-24.
- `HEADERS env var is forwarded on every CLI fetch (sync push of new script)` (windows) — added by #9075, regression on Windows only.

### Flow lock-gen race (1 + 2)

Flow create/update queues an async `FlowDependencies` job that fills inline-script lockfiles and rewrites `flow.value`. The two tests pulled/pushed before the worker finished, so the dry-run idempotency check saw phantom `*.inline_script.lock` adds and `flow.yaml` edits like:

```
{ "type": "added",  "path": "f/MyFlows/DataProcessor.flow/a.inline_script.lock" }
{ "type": "edited", "path": "f/MyFlows/DataProcessor.flow/flow.yaml" }
```

This mirrors the script-side fix #8202 but at a different layer: flows always queue the dependency job (vs. scripts which gate it on `lock.is_some()`), so we need a positive wait, not just a pre-set lock.

Added a `waitForFlowDependencyJob` helper that polls `/flows/get` for the latest `dependency_job` UUID and `/jobs_u/completed/get` until the job lands, then called it after every API/CLI flow write in both tests. The `sync_pull_push` copy returns silently on 404 since that test path only conditionally pushes the flow.

### Windows path (3)

The new `headers_env_var.test.ts` built the CLI entrypoint via:

```ts
const cliDir = new URL("..", import.meta.url).pathname;
const entrypoint = `${cliDir}/src/main.ts`;
```

`new URL().pathname` returns `/C:/runner/_work/windmill/windmill/cli/` on Windows. `Bun.spawn(["bun", "run", "/C:/.../src/main.ts"])` rejected before the CLI ever connected to the proxy, so the negative-case assertion `rejectedRequests.length > 0` got 0 (test ran in 47ms, well under the CLI startup time).

Switched to the same `fileURLToPath` + `node:path.join` pattern `cargo_backend.ts` already uses for `runCLICommand`.

## Test plan

- [x] `bun test test/mixed_case_paths.test.ts` — 7/7 pass locally (5x stress-tested).
- [x] `bun test -t "Mixed scripts and flows with nonDottedPaths are idempotent"` — passes.
- [x] `bun test test/headers_env_var.test.ts` — passes.
- [ ] CI on linux + windows passes the CLI Tests workflow.

---
## Summary by cubic
Stabilizes flaky CLI tests by waiting for flow dependency job completion and fixes Windows path resolution for the CLI entrypoint. Removes the race on inline-script lockfile generation and makes the `HEADERS` proxy test reliable on Windows.

- **Bug Fixes**
  - Updated `waitForFlowDependencyJob` to read `job_id` from `/flows/deployment_status/p/{path}` and poll `/jobs_u/completed/get/{job_id}`; called after flow create/update and `sync push` in `mixed_case_paths.test.ts` and `sync_pull_push.test.ts` (404 is a silent no-op). Prevents phantom `*.inline_script.lock` and `flow.yaml` diffs.
  - Switched `headers_env_var.test.ts` to build the CLI entrypoint with `fileURLToPath` and `node:path.join`, avoiding `/C:/...` paths that `Bun.spawn` rejects on Windows (works with Bun and Node).

<sup>Written for commit a89615eda14d66011672c992395662ee32ff39bf. Summary will update on new commits.</sup>

### DIFF
--- a/cli/test/headers_env_var.test.ts
+++ b/cli/test/headers_env_var.test.ts
@@ -18,6 +18,8 @@
 
 import { expect, test } from "bun:test";
 import { mkdir, writeFile } from "node:fs/promises";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 import type { Server } from "bun";
 import { withTestBackend } from "./test_backend.ts";
 
@@ -113,12 +115,16 @@ async function runCliThroughProxy(
   cwd: string,
   env: Record<string, string>,
 ): Promise<{ stdout: string; stderr: string; code: number }> {
-  const cliDir = new URL("..", import.meta.url).pathname;
+  // Use fileURLToPath + node:path so the CLI entrypoint is a real OS path on
+  // Windows. `new URL(..).pathname` yields `/C:/...` which Bun.spawn fails to
+  // resolve, so the negative case "rejected" on launch instead of reaching
+  // the proxy and the assertion on rejectedRequests > 0 flaked.
+  const cliDir = join(dirname(fileURLToPath(import.meta.url)), "..");
   const useNode = process.env["TEST_CLI_RUNTIME"] === "node";
   const runtime = useNode ? "node" : "bun";
   const entrypoint = useNode
-    ? `${cliDir}/npm/esm/main.js`
-    : `${cliDir}/src/main.ts`;
+    ? join(cliDir, "npm", "esm", "main.js")
+    : join(cliDir, "src", "main.ts");
   const runtimeArgs = useNode ? [entrypoint] : ["run", entrypoint];
 
   const fullArgs = [
--- a/cli/test/mixed_case_paths.test.ts
+++ b/cli/test/mixed_case_paths.test.ts
@@ -130,6 +130,37 @@ async function createFlow(
     throw new Error(`Failed to create flow ${flowPath}: ${error}`);
   }
   await response.text();
+
+  // Flow creation queues an async FlowDependencies job that generates the
+  // inline-script lockfile and rewrites flow.value. If we don't wait, a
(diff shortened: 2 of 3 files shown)


rubenfiszel:
Both reviewers (Claude, Codex) correctly flagged that the original `waitForFlowDependencyJob` was a no-op — `/flows/get` doesn't return `dependency_job`, so `depJobId` was always `undefined` and the helper returned without polling. Local runs passed because the dep job had already finished by the time the next CLI step ran on my machine.

Fixed in a89615e: switched to `/api/w/{ws}/flows/deployment_status/p/{path}`, which returns `{ lock_error_logs, job_id }`. `job_id` is the FlowDependencies UUID written into `deployment_metadata` in the same tx as the dep-job push (`backend/windmill-api-flows/src/flows.rs:660-672` and `:1275-1292`), so it's already the latest dep-job UUID by the time the create/update API call returns. Then poll `/jobs_u/completed/get/{job_id}` as before.

Local `mixed_case_paths.test.ts` runtime jumped from ~9s → ~32s, confirming the helper now actually waits instead of returning immediately. The 404 short-circuit in `sync_pull_push.test.ts` still works because `get_deployment_status` returns 404 when the flow is absent.

PR #9089, opened 2026-05-08
feat(license): offline (URL-bound) license keys

hugocasa:
## Summary

A new license-key flavor for air-gapped customers. Keys embed seat + CU caps + a per-instance binding hash, and validate entirely offline. Existing 3-segment online keys are unchanged.

- 4-segment wire format `<id>.<expiry>.<metadata_b64>.<signature>`
- License is bound to a per-instance hash computed in EE from `instance_uid` + canonical `base_url` (plain SHA-256 with a domain prefix); the instance recomputes locally and matches against the embedded hash
- `GET /settings/instance_hash` (super-admin only) returns the hash; admin runs `curl` with their token when requesting a key from support — not surfaced in the UI
- `GET /settings/offline_license_status` (super-admin only) returns the live cap status (seats used/cap, current CU/cap) for the offline license, or `null` when no offline license is loaded
- Hard seat block at invite/add-user when adding a user would exceed the cap. Counted from `usr ∪ workspace_invite`; mixed-role users (author somewhere, operator somewhere else) count as 1 author seat — author-anywhere. The check computes the exact seat delta (brand-new email = +1.0 / +0.5; operator-only → author = +0.5; everything else = 0) so net-zero ops (invite upserts, invite→user acceptance) aren't spuriously blocked.
- Current CU = sum of `cu_rate` across workers that pinged in the last 2 minutes; over cap ⇒ license invalidated immediately (no grace period)
- 5-minute throttle / cache on the full enforcement check; transient DB errors during the check preserve the prior cached state instead of silently re-validating
- `instance_uid` cached in memory after first lookup so the periodic verify cycle hits zero DB
- Critical-alerts pipeline reused (sidebar badge, modal, email/Slack/Teams)
- Superadmin instance settings panel shows two compact green/red status lines (`Seats X.X/Y`, `CUs X.X/Y`) for offline keys. "Latest key renewal" status + "Renew key" button are hidden for offline keys (renewal is server-disabled). Workers-page expiry toast is preserved for both online and offline keys.

Full design + per-component rationale in the EE-private companion PR.

Companion PRs:
- windmill-labs/windmill-ee-private (full design + EE impl)
- windmill-labs/windmill-customer-service#75 (admin endpoint to issue keys)
- windmill-labs/windmill-cf-worker-keygen#1 (signing)

## Performance impact for non-offline customers

Zero added cost. Every offline-cap path is gated on `LICENSE_OFFLINE_METADATA` being `Some(offline)`; when no offline key is loaded, `enforce_offline_caps`, `check_seat_cap_for_new_user`, and the hash recompute in `verify_license_key` all short-circuit before any DB query. Community Edition builds compile the EE call sites away via `#[cfg(feature = "enterprise")]`.

## Known soft guard: seat-cap is preflight, not transactional

`check_seat_cap_for_new_user` runs as a preflight check before `invite_user` / `add_user`'s insert transaction. Two simultaneous invites at exactly the cap boundary can both pass and both insert, putting the instance one user over cap. We treat the seat cap as a guardrail rather than a hard contractual limit:

- The over-cap state is immediately visible in `/settings/offline_license_status` and the superadmin panel after the second insert lands.
- Any subsequent invite is blocked by the (now updated) count.
- A transactional advisory lock is left as a future hardening if it becomes a real problem.

## Test plan

All 10 cases verified end-to-end against a local stack:

- [x] Correct hash → key loads
- [x] Wrong hash → rejected ("Offline license key is not bound to this instance")
- [x] `BASE_URL` change → next verify cycle invalidates
- [x] Seat cap blocks invite with a clear 400
- [x] Mixed-role user counted as 1 seat (not 1.5); invite-upsert / accept-invite are net-zero (not blocked at cap); operator-only → author transition correctly counts as +0.5 (blocked when over cap)
- [x] CU under cap → valid
- [x] CU over cap → invalid immediately
- [x] CU back under cap → valid within ≤5 min
- [x] `instance_hash` reachable via dedicated endpoint, super-admin only; DB errors propagate as 500 rather than `{"instance_hash": null}`
- [x] 5-min throttle returns cached; reload clears cache; transient DB errors during enforcement preserve prior cached state (don't silently re-validate)

`cargo check --features enterprise,license,parquet,private` clean, `SQLX_OFFLINE=true` clean.


hugocasa:
The latest Codex (P1) and Pi (P2) reviews both flag the same issue: the seat-cap check runs preflight outside the invite/add transaction, so two concurrent admin invites at the cap boundary can both pass and both insert. This is a deliberate choice — documented in the PR description under **"Known soft guard: seat-cap is preflight, not transactional"**.

Rationale:
- The seat cap is a guardrail to prevent admins from blowing past their contract by mistake, not a hard contractual limit
- Going one over surfaces immediately in `/settings/offline_license_status` and the superadmin panel
- Any subsequent invite is blocked by the (now-updated) count
- A transactional advisory lock that spans check + insert is left as future hardening if it becomes a real problem (would require threading an executor through `check_seat_cap_for_new_user` to share the same transaction)

The exploit window is two admins simultaneously inviting at exactly the cap boundary, with no in-process retry. Acceptable for an offline-license soft cap.

PR #9097, opened 2026-05-11
feat(forks): prompt to delete forked children when deleting a fork

diegoimbert:
## Summary
When deleting a forked workspace that itself has forked children, the existing delete-fork confirmation modal now surfaces those descendants and offers an opt-in toggle to delete them transitively. The toggle is off by default, so the existing behavior (children become orphans with `parent_workspace_id` set to NULL via the FK constraint) is preserved unless the user explicitly opts in.

The descendant list is computed purely from the already-loaded `userWorkspaces` store using `findWorkspaceDescendants` — no extra backend roundtrip.

## Changes
- `frontend/src/lib/components/sidebar/SidebarContent.svelte`:
  - Import `userWorkspaces` and `findWorkspaceDescendants`
  - Derive `forkedDescendants` from the current workspace
  - Add `deleteForkedChildren` state, reset to `false` each time the modal opens
  - Render a "Forked children" section inside the existing `ConfirmationModal` with a `Toggle` (off by default) and a list of descendant workspace IDs when descendants exist
  - In `deleteFork()`, when the toggle is on, delete each descendant via `WorkspaceService.deleteWorkspace` before deleting the current workspace; abort and toast on failure

## Test plan
- [ ] In a workspace with no fork ancestry, the "Delete Forked Workspace" action is not shown.
- [ ] In a fork with no children, the modal shows no "Forked children" block and the existing delete behavior is unchanged.
- [ ] In a fork with one or more descendants, the modal lists them with a toggle off by default; confirming with the toggle off deletes only the current workspace and leaves children as orphans (parent_workspace_id becomes NULL).
- [ ] Toggling on and confirming deletes every descendant first, then the current workspace; on permission failure for a descendant, the operation aborts with a toast.
- [ ] Modal state resets correctly when reopened (toggle returns to off).

---


diegoimbert:
/ai does partial-failure show at least an error toast ? if not then do a commit to show one. Partial failures are not so dramatic in this case

PR #9121, opened 2026-05-12
feat: persistent in-editor drafts via UserDraft

diegoimbert:
## Summary

Replaces the ad-hoc localStorage/URL-hash autosave that scripts, flows, apps and raw apps each grew their own copy of with a single `UserDraft` service. Edits in every editor now flow into one key-value store under `userdraft/w/{workspace}/{itemKind}/{path}`, so refreshing or accidentally closing a tab no longer loses unsaved work, and the persistence story is the same for scripts, flows, apps, raw apps, resources and variables.

The backend DB draft system (DraftService) is unchanged — `UserDraft` is purely a local autosave layer on top of it.

## Changes

- New service `frontend/src/lib/userDraft.svelte.ts` with `save`, `get`, `remove`, `has`, `use`. `use()` returns a reactive `{ draft }` handle backed by a shared `Map<{ws}/{kind}/{path}, { count, state }>`; `onDestroy` ref-counts the entry.
- `useLocalStorageValue` (`frontend/src/lib/svelte5Utils.svelte.ts`) now deep-tracks its state via an `$effect` and uses a `lastSerialized` cache to avoid double-writing when the synchronous setter has already persisted the same value. Undefined values now correctly remove the key instead of writing the string "undefined".
- Removed the old localStorage autosave + restore toasts from `FlowBuilder`, `AppEditor`, `AppEditorHeader`, `AppJsonEditor`, `RawAppEditor`, `RawAppEditorHeader`, `ScriptBuilder`, `ScriptEditor`, `storeUtils`, `FlowBuilderLiveTutorial`, and the `flows/edit`, `flows/add`, `apps/edit`, `apps/add`, `apps_raw/edit`, `apps_raw/add` routes. Also removed the URL-hash autosave that ScriptBuilder used to write through `replaceStateFn`.
- Wired each editor's top-of-tree state through `UserDraft`:
  - **Scripts**: routes own `UserDraft.use<EditableScript>`; ScriptBuilder is bound to `handle.draft`. Hash-mode (`?hash=`) opts out by passing `''`.
  - **Flows**: routes wrap a `UserDraft.use<Flow>` handle as a `StateStore<Flow>` so `FlowBuilder` keeps its `flowStore.val` convention without churn.
  - **Apps**: `AppEditor` registers a `UserDraft.use<App>` handle internally and a single `$effect` deep-tracks `stateApp`. Routes overlay local autosave on top of the backend value with a Discard / Show-diff toast.
  - **Raw apps**: routes persist a single bundle `{ files, runnables, data, summary }` per the "one draft only" rule.
  - **Resources / variables**: editors persist the current workspace's edit state; backend fetches overlay local autosave when they diverge.
- Empty-path (`/add`) drafts persist to localStorage like any other path, so reloading `/scripts/add` or `/flows/add` restores in-flight work. The "+" entry points (and any `?nodraft=true` link) call `UserDraft.remove(kind, '')` before navigating so explicitly starting a new item always opens a clean editor.
- Deploy, save-as-draft (including renames) and restore-deployed/draft each call `UserDraft.remove` on the relevant path so the local autosave doesn't outlive the action.
- Trigger editors (Schedule/Kafka/Nats/Mqtt/Sqs/Gcp/Azure/Websocket/Postgres/Email/HTTP) are intentionally left for follow-up — they live in short-lived drawers, have 11 different state shapes, and don't share a base class. The `UserDraftItemKind` enum already includes `schedule_*` keys for them.
- Tests at `frontend/src/lib/userDraft.test.ts` cover save/get/remove against localStorage, workspace isolation, defaultValue handling, observer sync between two `use()` handles, and ref-counting cleanup. `onDestroy` is mocked since the Node test env has no component context. `test-setup.ts` adds a minimal `window` so `useLocalStorageValue` stays on the browser branch.

## Test plan

- [ ] On `/scripts/edit/{path}` edit something, refresh — changes survive. Deploy — `userdraft/w/{ws}/script/{path}` is cleared.
- [ ] Same on `/flows/edit/{path}`, `/apps/edit/{path}`, `/apps_raw/edit/{path}`.
- [ ] On `/scripts/add` edits survive a reload (persisted under `userdraft/w/{ws}/script/`); clicking "+ Script" / "+ Flow" (`?nodraft=true`) opens a clean editor.
- [ ] Rename inside the editor and save-as-draft — old key cleared, new URL has its own (empty) draft, backend draft visible on next load.
- [ ] Open a resource and a variable drawer, edit, close without saving, reopen — values restored.
- [ ] Switch workspace mid-edit — drafts stay scoped to their workspace.
- [ ] `bunx vitest run src/lib/userDraft.test.ts` — 27 tests pass.
- [ ] `npm run check:fast` — clean.


diegoimbert:
/ai fix this

5. setDraftConfig in triggers.svelte.ts:73 still has a leftover console.log. Not new in this PR, but visible in the file. Cheap to drop. [Fix this →](https://claude.ai/code?q=In%20frontend%2Fsrc%2Flib%2Fcomponents%2Ftriggers%2Ftriggers.svelte.ts%20around%20line%2073%2C%20remove%20the%20console.log%28%27setDraftConfig%27%2C%20triggerIndex%2C%20draftConfig%29%20call%20%E2%80%94%20it%27s%20leftover%20debug%20logging.&repo=windmill-labs/windmill)

6. PR description still says "Empty paths bypass localStorage and stay in-memory only". This was changed by commit e7e5869 and confirmed by test userDraft.test.ts:247-289. Description text should match — minor housekeeping.

Issue #9140, opened 2026-05-13
bug: `wmill sync pull` erases error handler scripts


hnsr:
### Describe the bug

I have set up some error handlers for a few flows some time ago, which have been working fine all this time. However today I did a `wmill sync pull` from a workspace, and it seems to throw away the contents of my error handlers and replace it with the line `!inline Handle_error.ts`.

### To reproduce

1. Have an error handler defined for a flow, defined online through the UI, then pulled locally
2. <unknown> not sure what happened in between that caused this issue, aside from passage of time
3. Pull workspace locally with `wmill sync pull`
4. Observer that local contents of `Handle_error.ts` files get erased and replaced with `!inline Handle_error.ts`

### Expected behavior

I would expect there to be no changed at all as I have not modified my error handlers in anyway since creating them through the UI and pulling/pushing them many times after that without issues.

### Screenshots

<img width="933" height="685" alt="Image" src="https://github.com/user-attachments/assets/6e4d1e26-0da9-4c53-a035-2ec35056ed08" />
<img width="864" height="412" alt="Image" src="https://github.com/user-attachments/assets/f51e4f38-d0cb-4461-ba60-1323fd8decf8" />

### Application version

CE v1.690.0 (backend), 1.693.1 (CLI)

### Additional Context

Note that I have never touched, changed, or renamed these files locally at any point.

I did notice that there is a difference in casing (handle_error vs. Handle_error), and I do use macOS locally for dev, so maybe that might be related. Though again, I never changed the name of these files and I created this through the UI so not sure how that discrepancy came into existence.

hnsr:
Note that after upgrading wmill CLI to 1.700.2, the issue persists

hnsr:
When I open one of the affected flows in the Windmill UI this is shown (not sure if it is a cause or effect):

<img width="858" height="290" alt="Image" src="https://github.com/user-attachments/assets/48267b43-cff3-4223-ae69-7c981e524e37" />

Prasad8830:
Hello @hnsr, @rubenfiszel, can I work on this issue? I am new to this repository

PR #9142, opened 2026-05-13
fix(cli): prevent !inline-corruption in flow push/pull

rubenfiszel:
## Summary

Fixes [GIT-871](https://linear.app/windmilldev/issue/GIT-871) / windmill-labs/windmill#9140: `wmill sync pull` was overwriting users' error-handler scripts with the literal text `!inline Handle_error.ts`.

Root cause: a previous push had silently sent the unresolved `!inline ...` directive as `rawscript.content` to the backend (because `replaceInlineScripts` couldn't find the local file — typically a casing mismatch from PR #8940 + a push from a case-sensitive environment like Linux CI). The `flow_version.value` row got poisoned, and every subsequent pull faithfully wrote that string back to disk as the "script body."

This PR closes both ends of that loop:
1. **Push side** — refuse to push when any inline-script file is missing, so corrupt content can never reach the backend in the first place.
2. **Extractor side** — refuse to write `!inline ...` content to a script file on pull. If a workspace is already poisoned, the user gets a clear error pointing at the corruption instead of silently losing their handler again.

## Changes

- `cli/src/commands/flow/flow.ts` — `pushFlow` now throws on `missingFiles.length > 0` (was a warn-and-continue).
- `cli/src/commands/flow/flow_metadata.ts` — `regenerateFlowMetadata` (post-pull lock regeneration, which also calls `updateFlow`) now collects `missingFiles` across `modules` + `failure_module` + `preprocessor_module` and aborts before `updateFlow` if any are missing. This was a second corruption vector.
- `cli/windmill-utils-internal/src/inline-scripts/extractor.ts` — `extractRawscriptInline` throws when `rawscript.content` is itself an `!inline ...` string, instead of writing it to disk.

## Test plan

- [ ] Existing inline-script unit tests still pass: `bun test test/inline_scripts_failure_preprocessor_unit.test.ts test/yaml_inline_tag.test.ts test/app_inline_script_ordering_unit.test.ts` (verified locally — 34/34 pass).
- [ ] Manually simulate the broken state: in a flow folder, edit `flow.yaml` so `failure_module` references `!inline missing.ts` with no corresponding file. Run `wmill sync push` → expect a clear error mentioning `missing.ts`, no API call made.
- [ ] Manually simulate a poisoned backend: `UPDATE flow_version SET value = jsonb_set(value, '{failure_module,value,content}', '"!inline Handle_error.ts"') WHERE ...`. Run `wmill sync pull` → expect a clear error from the extractor instead of `Handle_error.ts` being overwritten.
- [ ] Recovery path: re-push from a known-good local copy (after fixing the extractor error path) — backend `flow_version.value` is repaired, subsequent pulls round-trip cleanly.

---
## Summary by cubic
Prevents "!inline" corruption during flow sync by failing fast on missing inline files and adding an opt-in extractor guard used on pull/lock regen. Fixes Linear GIT-871; stops `wmill sync pull` from overwriting handlers with literal "!inline ...".

- **Bug Fixes**
  - Push: `pushFlow` now throws when any inline file is missing to avoid poisoning `flow_version.value`.
  - Lock regen: `generateFlowLockInternal` aggregates missing files across `modules`, `failure_module`, and `preprocessor_module` and aborts before `updateFlow`.
  - Pull/regen: extractor guard is opt-in via `failOnInlineDirective`; enabled for `sync pull` and lock regeneration, throwing if backend `rawscript.content` is an `!inline ...` string.
  - Local flows unchanged (no guard); tests updated so `!inline` paths are relative to the flow folder.

<sup>Written for commit ea8565764294b0728e5ad537dc6cffc29e6b4720. Summary will update on new commits.</sup>

### DIFF
--- a/cli/src/commands/flow/flow.ts
+++ b/cli/src/commands/flow/flow.ts
@@ -174,10 +174,14 @@ export async function pushFlow(
     await replaceInlineScripts([localFlow.value.preprocessor_module], fileReader, log, localPath, SEP, undefined, missingFiles);
   }
   if (missingFiles.length > 0) {
-    log.warn(colors.yellow(
-      `Warning: missing inline script file(s): ${missingFiles.join(", ")}. ` +
-      `The flow will be pushed with unresolved !inline references.`
-    ));
+    // Hard-fail rather than push the literal `!inline path` text as
+    // rawscript.content. That string would be persisted in flow_version.value
+    // and round-trip as the script body on the next pull, overwriting the
+    // user's local handler with the directive — see GIT-871 / #9140.
+    throw new Error(
+      `Cannot push flow ${remotePath}: missing inline script file(s): ${missingFiles.join(", ")}. ` +
+      `Either restore the file(s) or remove the !inline reference(s) from flow.yaml before pushing.`
+    );
   }
 
   const hasOnBehalfOf = (localFlow as any).has_on_behalf_of ?? !!localFlow.on_behalf_of_email;
--- a/cli/src/commands/flow/flow_metadata.ts
+++ b/cli/src/commands/flow/flow_metadata.ts
@@ -266,19 +266,29 @@ export async function generateFlowLockInternal(
           return tree.isStale(treePath);
         })
       : changedScripts;
+    const missingFiles: string[] = [];
     await replaceInlineScripts(
       flowValue.value.modules,
       fileReader,
       log,
       folder + SEP!,
       SEP,
-      locksToRemove
+      locksToRemove,
+      missingFiles
     );
     if (flowValue.value.failure_module) {
-      await replaceInlineScripts([flowValue.value.failure_module], fileReader, log, folder + SEP!, SEP, locksToRemove);
+      await replaceInlineScripts([flowValue.value.failure_module], fileReader, log, folder + SEP!, SEP, locksToRemove, missingFiles);
     }
     if (flowValue.value.preprocessor_module) {
-      await replaceInlineScripts([flowValue.value.preprocessor_module], fileReader, log, folder + SEP!, SEP, locksToRemove);
+      await replaceInlineScripts([flowValue.value.preprocessor_module], fileReader, log, folder + SEP!, SEP, locksToRemove, missingFiles);
+    }
+    if (missingFiles.length > 0) {
+      // Abort before updateFlow rather than push the literal `!inline path`
+      // string as rawscript.content (GIT-871 / #9140).
+      throw new Error(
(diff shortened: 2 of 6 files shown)


rubenfiszel:
Addressed Claude's review in 00e770e:

**P0 — extractor guard breaking yaml-parsed callers (`generate-locks` / `generate-metadata` / `dev`)**: moved the guard from being unconditional to opt-in via a new `failOnInlineDirective` option on `ExtractInlineScriptsOptions`. Default off — yaml-parsed callers are unaffected. Set to `true` from the two call sites that legitimately operate on backend-shaped data:
- `cli/src/commands/sync/sync.ts:947` — the actual sync-pull extraction.
- `cli/src/commands/flow/flow_metadata.ts:317` — post-`updateFlow` re-extraction (defensive: with the push-side hard-fail, the backend should never return `!inline ...` content, but if it does we'd rather fail loud than corrupt the local file).

**Test coverage**: added 5 unit tests in `inline_scripts_failure_preprocessor_unit.test.ts` covering both modes:
1. Default behavior: yaml-parsed module with `!inline foo.ts` content extracts without throwing (the regression Claude flagged).
2. The `!inline foo.ts` content round-trips as the script's body (preserves the placeholder).
3. Opt-in: throws on `!inline ...` content.
4. Opt-in: real script content still extracts cleanly.
5. Opt-in: throws on nested rawscript inside `branchall` (verifies recursion threads `options` through).

39/39 inline-script tests pass.

**P2 — partially-mutated `flowValue` on throw**: added a comment at the throw site noting that `replaceInlineScripts` may have already mutated some modules in place, so callers must re-parse rather than reusing the value.

**P2 — release note for push behavior change**: noted, will add to the release notes when the bot opens the release PR.

rubenfiszel:
Two integration tests (`Flow folder structure is created correctly`, `Sync pull with nonDottedPaths uses non-dotted inline script filenames`) were failing because of a buggy fixture, not the new hard-fail logic.

Fixed in ea85657: the local `createFlowFixture` in `cli/test/sync_pull_push.test.ts:111` was emitting `!inline f/test/flow_xxx.flow/a.ts` (path-from-root), but `pushFlow`'s fileReader resolves relative to the flow folder, so the file was always "missing." The previous warn-and-continue silently pushed `!inline f/test/...` as the literal `rawscript.content` — i.e., the test fixture was reproducing exactly the corruption pattern this PR is fixing. Switched to `!inline a.ts` to match the canonical fixture in `test_fixtures.ts` and the on-wire shape produced by sync-pull.

PR #9186, opened 2026-05-15
fix: atomic bundle cache writes to prevent parallel cold-load race

rubenfiszel:
## Summary

When a flow runs a `forloopflow` with `parallel: true` over a `//native` script and the worker's bundle cache is cold (post-sync, post-bounce, post-pod-redeploy), a fraction of the spawned sandboxes crash inside the loader wrapper **before user `main()` runs** — `TypeError: module.main is not a function` at `<anon>:34:51` or `SyntaxError: Unexpected end of input` at `file:///eval.ts`. The flow's `failure_module` catches the throw, but earlier steps in that iteration have already committed side effects, leaving a half-applied state needing manual cleanup. The race is probabilistic and curable by a warmup run — classic cache-population race signature.

**Root cause:** the bundle cache is populated non-atomically. `write_binary_file` did `File::create(path)` → `write_all` → `flush`; `save_cache`/`pull_codebase` did `std::fs::copy` / `std::fs::write`; dir caches extracted/copied straight into the final path. `File::create` makes the path **visible and zero-length immediately**. A concurrent cold-load sandbox gating on `metadata(path).is_ok()` (`load_cache` / `exists_in_cache` / `pull_codebase`) then observes the path as "present" mid-write and reads a stub (no `main` export) or a truncated bundle. The first fully-successful write makes every later run safe — exactly the observed warmup-cures-it, 60%→0% behavior.

This supersedes #9185, which fixed the two highest-priority single-file sites but missed the codebase single-file write and the dir-mode cold-load/producer paths, and shipped no regression test.

## Changes

- **New atomic publish helpers** in `backend/windmill-common/src/worker.rs` (non-feature-gated, reusable, testable): `atomic_write_file_bytes`, `atomic_copy_file`, `atomic_publish_dir`. All write to a sibling `*.tmp.{uuid}` path in the same directory (same filesystem → `rename` is atomic on POSIX) then `rename` into place; temp is cleaned up on every error path. `atomic_publish_dir` treats a lost publish race (rename onto an existing non-empty dir → ENOTEMPTY) as success and discards its byte-identical copy.
- **`write_binary_file`** now delegates to `atomic_write_file_bytes` — fixes the exact reported `//native` `load_cache` consumer race.
- **`save_cache`** (`global_cache.rs`): non-dir branch uses `atomic_copy_file`; dir branch populates a temp dir then `atomic_publish_dir` — fixes the producer-side race for both single-file and dir caches.
- **`load_cache`** (`global_cache.rs`): the `is_dir` cold-load now extracts into a temp dir then `atomic_publish_dir` — fixes the codebase-dir cold-load race.
- **`pull_codebase`** (`bun_executor.rs`): the codebase `std::fs::write(&bun_cache_path, ...)` now uses `atomic_write_file_bytes` — the single-file site #9185 missed.
- **Regression tests** in `worker.rs`: `test_atomic_write_file_bytes_concurrent_cold_load` (concurrent reader asserts it never observes a partial bundle; fails against the old non-atomic impl) and `test_atomic_publish_dir_thundering_herd` (8 racing populators all succeed, final dir complete, no temp leaks).

Single-flight dedup of the N concurrent object-store fetches is intentionally **out of scope** — it's a throughput optimization, not a correctness fix; atomic publish fully closes the race.

## Test plan

- [x] `cargo test -p windmill-common --lib worker::tests::test_atomic` — both regression tests pass
- [x] `cargo check --features enterprise,parquet` — clean (covers the feature-gated call sites)
- [x] `/local-review` against main — no issues
- [ ] Deploy a `//native` codebase script; run a flow with a parallel for-loop (16+ iterations) against a cold worker cache (clear `ROOT_CACHE_DIR`, restart worker); confirm no iteration fails with `module.main is not a function` / `Unexpected end of input`
- [ ] Repeat on an EE object-store-backed dir cache (parquet/enterprise build) cleared locally to exercise the `load_cache` tar-extract + `atomic_publish_dir` path; confirm race-free and no leftover `.tmp.<uuid>` dirs under the cache root

---
## Summary by cubic
Make cache writes atomic to stop the parallel cold-load race that caused `module.main is not a function` and truncated bundles in `//native` runs. Writers now use tmp+rename with a content-addressed exists fallback so workers only see complete files and dirs (POSIX and Windows).

- **Bug Fixes**
  - Added `atomic_write_file_bytes`, `atomic_copy_file`, and `atomic_publish_dir` with sibling tmp + atomic `rename` and an exists-fallback on rename failure (handles concurrent publishers and Windows open-dest); cleans up temps on error.
  - Switched `write_binary_file`, `save_cache` (file via `atomic_copy_file`; dir via tmp + `atomic_publish_dir`), `load_cache` (dir extract to tmp then publish), and `pull_codebase` to use the atomic helpers.
  - Simplified `atomic_publish_dir` to a single `rename` with exists-fallback; an already-present dir is treated as success to avoid herd errors and temp leaks.
  - Added regression test `test_atomic_publish_dir_existing_is_trusted_not_corrupted` in addition to the concurrent cold-load and thundering-herd tests.

<sup>Written for commit b8433035c8c6cef002457fb3d87612b1461e551a. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9186?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/windmill-common/src/worker.rs
+++ b/backend/windmill-common/src/worker.rs
@@ -1023,22 +1023,74 @@ pub fn atomic_copy_file(origin: &str, final_path: &str) -> error::Result<()> {
     Ok(())
 }
 
-/// Publish a fully-populated `tmp_dir` to `final_dir` atomically. If a concurrent
-/// populate already published `final_dir` (rename fails because the destination
-/// exists and is non-empty), discard our byte-identical copy and treat it as a
-/// success so cold-load directory caches are race-free too.
+/// Publish a fully-populated `tmp_dir` to `final_dir`, always ending with a
+/// complete freshly-built directory at `final_dir` and never exposing a
+/// partially-populated one to a concurrent reader gating on
+/// `metadata(final_dir)`.
+///
+/// These dir caches are content-addressed (the path embeds a hash of the
+/// inputs), so every concurrent publisher writes byte-identical content and a
+/// *complete* `final_dir` is immutable-correct. The hard part is a *stale or
+/// partial* `final_dir` left behind by a populate that was hard-killed on a
+/// pre-atomic binary: POSIX `rename` cannot replace a non-empty directory, so a
+/// naive "trust whatever is there" would serve that partial dir forever (no
+/// self-heal). Instead we move the stale dir aside and swap our complete copy
+/// in, so we are strictly safer than an in-place re-extract in every case.
+///
+/// Invariants for a concurrent reader: `final_dir` is only ever observed as
+/// (a) absent, (b) the *unchanged* pre-existing dir, or (c) a complete fresh
+/// publish. We never mutate `final_dir` in place, so a reader can never see a
+/// newly-created partial. The brief window where the stale dir is moved aside
+/// makes `final_dir` momentarily absent — that degrades to a benign cache miss
+/// (re-pull/recompute), never to reading corrupt or partial content.
 pub fn atomic_publish_dir(tmp_dir: &str, final_dir: &str) -> error::Result<()> {
-    match fs::rename(tmp_dir, final_dir) {
-        Ok(()) => Ok(()),
-        Err(e) => {
-            let _ = fs::remove_dir_all(tmp_dir);
-            if Path::new(final_dir).exists() {
-                Ok(())
-            } else {
-                Err(e.into())
-            }
-        }
+    // Fast path: nothing at `final_dir` yet (the common cold-load case). A
+    // single atomic rename; also succeeds if `final_dir` is an empty dir.
+    if fs::rename(tmp_dir, final_dir).is_ok() {
+        return Ok(());
     }
+
+    // `final_dir` does not exist and the rename failed for a real reason
+    // (ENOSPC, EACCES, EROFS, ...). There is nothing safe to fall back to.
+    if !Path::new(final_dir).exists() {
+        let _ = fs::remove_dir_all(tmp_dir);
+        return Err(error::Error::ExecutionErr(format!(
+            "could not publish cache dir {final_dir}: no destination and rename failed"
+        )));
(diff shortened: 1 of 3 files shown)


rubenfiszel:
### Design update after adversarial review

Two follow-up commits refine the dir-cache path and close a cross-platform gap:

**Windows / race-loser handling (`atomic_write_file_bytes`, `atomic_copy_file`, `atomic_publish_dir`).** All three now treat *rename failed but destination already exists* as success. The caches are content-addressed, so an existing destination is byte-identical content already published by a peer. Without this, on Windows `std::fs::rename` cannot replace an open/existing destination and the race loser propagated an error (`pull_codebase`/`save_cache` would still fail under the exact thundering-herd this PR targets). POSIX race losers are likewise treated as success rather than spuriously erroring.

**`atomic_publish_dir` kept deliberately simple (single atomic rename).** An intermediate "destroy-then-recreate swap" version (to also rebuild a stale partial dir) was implemented and then reverted: a fresh adversarial review showed it introduced concurrent-publisher edge cases (spurious `Err` under a real herd; `EACCES` masking a stale partial) — a worse trade on a critical path than the residual it removed.

**Accepted residual (documented in code).** A *stale partial* dir left by a populate hard-killed on a pre-atomic binary is trusted rather than rebuilt. This is finite and self-draining (post-fix code only ever renames a complete temp dir into place, so it cannot create that state), confined to the PHP `vendor/` and Java dependency caches (**never** the `//native` single-file path this PR fixes), and clears on cache eviction.

Net: the single-file `//native` path (the reported bug) is provably strictly-safer than `main` with no new failure modes; the dir paths are strictly-safer for the concurrency race with one documented, off-path, self-healing residual.

PR #9202, opened 2026-05-17
fix: enforce folder ACL on flow run-by-version routes (GHSA-8mv7-hmrg-96xv)

rubenfiszel:
## Summary

Private security fix for **GHSA-8mv7-hmrg-96xv** (HIGH). The version-keyed flow
run routes resolved `flow_version.id -> path` on the **raw DB pool** and then
gated only on `check_scopes(...)`, which checks API-token scopes — not the
caller's folder-level RBAC. The folder ACL is enforced by Postgres RLS on the
`flow` table, reached only through `user_db`; the version routes never went
through it. A low-privilege workspace member with a normal (unscoped) token
could therefore execute an admin-owned, folder-scoped flow they have zero
permission on, by passing its `flow_version.id`. The path-keyed sibling
`POST /jobs/run/f/{path}` was always correctly gated (RLS-aware lookup) and
returned 401 — confirming the asymmetry.

Affected routes (all share the fix): `POST /jobs/run/fv/{version}`,
`POST|GET /jobs/run_wait_result/fv/{version}`, and (transitively)
`GET|POST /jobs/run_and_stream/fv/{version}`.

Introduced in v1.576.0 (PR #7062, "webhook by flow version"); vulnerable
through current main / v1.703.1.

## Changes

- Add `get_flow_path_for_version_authed` in `windmill-common/src/lib.rs`: resolves a `flow_version.id` to its path through an RLS-filtered `flow_version INNER JOIN flow` query on `user_db`, mirroring the exists-but-unauthorized → `NotAuthorized` / missing → `NotFound` semantics of `get_latest_flow_version_id_for_path`.
- Route the three version-keyed flow handlers (`run_flow_by_version_inner`, `run_wait_result_flow_by_version`, `run_wait_result_flow_by_version_get`) through the new helper instead of the raw-pool lookup, before `check_scopes`. `stream_flow_by_version` is covered transitively via `run_flow_by_version_inner`.
- Add regression test `backend/tests/ghsa_8mv7_hmrg_96xv.rs` + fixture.
- Regenerate SQLx offline cache (1 new query; no EE caches lost).

## Test plan

- [x] Regression test passes: folder owner (non-admin) still runs by version (201); unauthorized member rejected on path-keyed control (401) and on all version routes (401)
- [x] Manual end-to-end repro: pre-fix `fv/{id}` executed an admin folder-scoped flow as a zero-permission member; post-fix all four version routes return 401 while authorized users (admin + folder-granted member) still get 201
- [x] `local-review` clean (no P0/P1/P2)
- [x] Backend compiles (cargo watch); SQLx cache verified (0 net deletions vs origin/main)

---
## Summary by cubic
Fixes GHSA-8mv7-hmrg-96xv by enforcing folder ACL on all flow run-by-version routes. Blocks running a folder-scoped flow via `flow_version.id` without permission and avoids leaking the flow path in unauthorized errors.

- **Bug Fixes**
  - Added `get_flow_path_for_version_authed` to resolve `flow_version.id -> path` via an RLS-filtered join on `user_db`, with NotAuthorized/NotFound semantics and no path leakage.
  - Updated `POST /jobs/run/fv/{version}`, `POST|GET /jobs/run_wait_result/fv/{version}`, and the streaming variant (via `run_flow_by_version_inner`) to use the helper before scope checks.
  - Removed the temporary regression test; regenerated SQLx offline cache.

<sup>Written for commit 1a1f2eb2ad89374131444009e690592a24cedbb1. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9202?utm_source=github">Review in cubic</a></sup>


rubenfiszel:
Thanks @cubic-dev-ai — addressed in a1da7d7de0.

**P2 (NotAuthorized leaks flow path + permission details), identified by cubic — valid and fixed.**

Nuance: the message was copied verbatim from the existing path-keyed sibling `get_latest_flow_version_id_for_path` (`lib.rs:1455`), so it isn't a *new* class of disclosure. But cubic is right that the version route is genuinely worse: the path-keyed caller already supplied the path, whereas the version-keyed caller supplied only an opaque `flow_version.id`. Echoing the resolved path back therefore discloses an **id→path mapping for a flow they cannot access** — a real incremental leak unique to this route, plus the unnecessary `authed` debug dump.

Fix:
- The `NotAuthorized` branch now returns a generic `"You are not authorized to run this flow version"` — no resolved path, no `authed` debug. Comment added explaining *why* it intentionally diverges from the (more verbose) sibling.
- The raw fallback is now `SELECT EXISTS(...)` used only internally to keep the 401-vs-404 distinction (consistent with the sibling's status-code behavior); the path string is never materialized into a response.
- Regression test hardened: asserts the rejection body does **not** contain the flow path.

Did not touch the path-keyed sibling's message — out of scope for this advisory, and there the path is caller-supplied so it's a non-issue; changing it would also make the two routes inconsistent in a different way.

The other three reviewers (Claude, Codex, Pi) returned Good to merge with no blocking findings.

PR #9207, opened 2026-05-18
feat: export audit logs to a dedicated object store folder

rubenfiszel:
## Summary

Adds an opt-in feature to mirror Windmill audit logs to a dedicated `logs/audit/` folder in the configured instance object store (S3/Azure/GCS) as newline-delimited JSON, in addition to them already living in Postgres `audit_partitioned` and the service logs. Designed for **zero overhead on the `audit_log()` hot path**: the export is a background, server-side-aggregated batch read off a checkpoint, never a synchronous write in the request path.

## Changes

- **`LOGS_AUDIT` constant** (`logs/audit/`) in `windmill-common/src/tracing_init.rs`.
- **New instance setting `store_audit_logs_s3`** (separate toggle, independent of `monitor_logs_on_s3`): constant `STORE_AUDIT_LOGS_S3_SETTING`, cached `AtomicBool STORE_AUDIT_LOGS_S3`, reload fn `reload_store_audit_logs_s3_setting`, wired into `initial_load` and the settings-change listener in `main.rs`.
- **Checkpoint** persisted in `global_settings` under `audit_logs_s3_checkpoint` (`{last_xmin, last_ts}`).
- **Background exporter** `export_audit_logs_to_object_store` in `monitor.rs` (parquet-gated), wired into the `monitor_db` `join!` loop (`server_mode`, every ~60s):
  - Transaction-scoped `pg_try_advisory_xact_lock` → single exporter in HA; auto-released on tx end (panic/early-return safe).
  - Checkpoint read **inside** the locked tx (atomic read→decide→write).
  - **Cursor is the transaction snapshot xmin (xid space), not `id`.** Each run captures `cur = txid_snapshot_xmin(txid_current_snapshot())`; the checkpoint stores the previous run's `cur` as `prev`; the run exports exactly the rows whose inserting xid is in the half-open interval `[prev, cur)` (`age(xmin) > age(cur) AND age(xmin) <= age(prev)`). Snapshot xmin is monotonic ⇒ disjoint intervals covering every transaction exactly once. This is correct even though `audit_log()` runs in the caller's tx (xid assigned at its first, possibly non-audit, write) so `id` order ≠ xid order and an in-flight low `id` is invisible — an `id`-based cursor cannot be safe here.
  - `age()` comparisons keep the interval test correct across xid wraparound.
  - One bounded Postgres-side query builds NDJSON via `row_to_json`/`string_agg`, grouped by UTC day.
  - Object key `logs/audit/dt=YYYY-MM-DD/audit_xmin_{prev}.ndjson` — interval lower bound makes retries idempotent (same key overwritten; disjoint across successful runs).
  - `timestamp >= last_ts - 7 days` predicate for partition pruning only (slack far exceeds any realistic txn; cannot drop a settling row).
  - First run anchors the checkpoint to the current snapshot xmin (no historical backfill — older logs remain in Postgres / service logs).
- **Frontend**: `store_audit_logs_s3` toggle in the Object Storage section of `instanceSettings.ts`.
- **Test**: inline `#[sqlx::test]` integration test `monitor::audit_s3_export_tests::audit_export_end_to_end` (parquet-gated) with an injected in-memory object store.

No `*_ee.rs` files modified (the export is mechanical and mirrors the existing OSS `send_log_file_to_object_store`), so there is no EE companion PR.

## Tests

One end-to-end `#[sqlx::test]` (single DB, in-memory object store) asserting:

- **First-run anchor**: pre-existing rows are *not* backfilled.
- **Steady state**: post-enable rows exported exactly once.
- **Idempotent**: re-run with nothing newly settled produces no duplicates.
- **No-loss, older txn open**: rows whose xid ≥ an open txn's are not exported until it commits, then all settle and export exactly once.
- **P0 regression (CI-found, divergent xid/id)**: `t_fast` takes a low xid via a non-audit write then audit-inserts a *higher* id and commits; `t_slow` audit-inserts a *lower* id (higher xid) and stays open. The higher id is exported and the checkpoint advances, yet the lower id is **still exported once `t_slow` commits** — an `id`-cursor would lose it permanently.

Run: `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes.

## Review history

A fresh-context local review and the automated PR review (Claude) each found a P0 that was fixed before/after surfacing:

1. Initial fix replaced a wall-clock watermark with a snapshot-xmin settle gate.
2. The automated review found that gate still allowed an `id`-leapfrog (id order ≠ xid order; invisible in-flight rows). Fixed by switching the cursor to the xid-interval `[prev, cur)` described above, plus an `age()`-based wraparound-safe comparison (P2), plus the divergent-ordering regression test. See the PR comment for details.

## Validation

- `cargo check` (default) and `cargo test --features parquet --bin windmill` pass (rebased on latest `main`); sqlx macros validate all queries against a live DB.
- SQLx offline cache regenerated via the safe procedure (verified **0 net deletions** from the EE cache set).
- `EXPLAIN`: partition pruning bounded to the 7-day window.
- `npm run check:fast`: no new errors in `instanceSettings.ts`.

## Known limitations / tradeoffs (bounded lag, never loss; self-healing)

- A long-running transaction of any kind holds snapshot xmin back and stalls export progress until it ends.
- After such a stall the next run exports the whole accumulated `[prev, cur)` interval in one batch (bounded by the audit volume that built up during the stall).
- Could not exercise against a real S3/Azure/GCS backend in this environment (dev `cargo watch` builds `--features quickjs` only); covered by the in-memory `#[sqlx::test]`, sqlx live-DB validation, and `EXPLAIN`.

## Manual test plan (real object store)

- [ ] Enable **Store audit logs in object storage**; generate audit activity; after ~60s confirm `logs/audit/dt=YYYY-MM-DD/audit_xmin_<n>.ndjson` with valid NDJSON, no missing/duplicate ids.
- [ ] HA: two servers — only one exports per tick (advisory lock).
- [ ] Toggle off: export stops.


rubenfiszel:
## Addressed the automated review (P0 + P2)

Thanks — both findings were correct. The earlier `xmin < txid_snapshot_xmin` settle-gate proved the *exported row's* transaction was settled but, because `audit_log()` runs in the caller's tx (xid assigned at its *first* write, not the audit INSERT), a lower `id` could belong to a younger-xid still-open transaction and be **invisible** — so neither "max settled id" nor a "smallest-not-settled-visible-id" gap-free prefix can be safe (the gap is an invisible row).

**Fix (commit `fix: cursor audit export on snapshot xmin to prevent id-leapfrog loss`):** the cursor is no longer `id` — it is the **transaction snapshot xmin (xid space)**. Each run captures `cur = txid_snapshot_xmin(...)`; the checkpoint stores the previous run's `cur` as `prev`; the run exports exactly the rows whose inserting xid is in the half-open interval `[prev, cur)` (`age(xmin) > age(cur) AND age(xmin) <= age(prev)`). Snapshot xmin is monotonic, so intervals are disjoint and cover every transaction exactly once; an invisible in-flight low-`id` row simply falls into a later interval once its transaction settles — no leapfrog, no loss, regardless of id/xid divergence.

- **P0 (id-leapfrog loss):** fixed by the xid-interval cursor above. A new regression test reproduces the exact divergent case you described — `t_fast` takes a low xid via a non-audit write then audit-inserts a *higher* id and commits, while `t_slow` audit-inserts a *lower* id (higher xid) and stays open — and asserts the lower id is still exported once `t_slow` commits (an `id`-cursor loses it).
- **P2 (xid wraparound):** the settle/interval test uses `age()` on both sides instead of raw 32-bit `xmin` vs 64-bit `txid_snapshot_xmin`, so it stays correct across wraparound.

`BATCH_LIMIT` was removed (an `id`-LIMIT is unsafe with an xid cursor — it could advance `prev` past un-exported rows in the interval). A run now exports the full settled `[prev, cur)` interval; documented tradeoff: after a long xmin stall the next run is one larger batch (bounded by the audit volume that accumulated during the stall — data that must ship regardless), still partition-pruned by the 7-day timestamp floor.

Validation: `cargo test --features parquet --bin windmill audit_export_end_to_end` passes (rebased on latest `main`); `EXPLAIN` confirms partition pruning; SQLx cache regenerated (0 EE-cache deletions).

rubenfiszel:
## Addressed the Codex + Pi review

Both new reviews independently flagged the same P1, and Pi confirmed the earlier Claude-P0 / Codex-wraparound concerns are resolved by the xid-interval cursor + `age()`. Fixed both open issues:

### P1 — checkpoint deletable / leaked via instance-config sync (Codex #1, Pi)

`audit_logs_s3_checkpoint` is internal runtime state but was a plain `global_settings` key, so a Replace-mode `sync-config` that omits it (static template / hand-written YAML) would delete it → next run re-anchors at the current snapshot xmin and silently skips every audit row in the gap; it also leaked via `from_db` exports.

**Fix:** added `"audit_logs_s3_checkpoint"` to `HIDDEN_SETTINGS` (`instance_config.rs`). Per the existing diff/sync logic this both (a) excludes it from `from_db`/config exports and (b) prevents Replace-mode deletion (`diff_global_settings` skips `HIDDEN_SETTINGS`). Added regression unit test `diff_global_settings_replace_keeps_audit_s3_checkpoint` asserting the cursor is never in `deletes` even when the desired config omits it.

### P1 — unbounded post-outage batch (Codex #2)

After a prolonged object-store outage the `[prev, cur)` interval could span hours/days and be materialized in one `string_agg` + one Rust result set under the lock.

**Fix:** each run is now capped to `MAX_XID_INTERVAL` (500k) xids: `eff_cur = min(cur_xmin, prev_xmin + MAX_XID_INTERVAL)`. Because any xid `< cur_xmin` is fully settled, any `eff_cur <= cur_xmin` is itself a valid settled boundary, so the checkpoint advances exactly to `eff_cur` and the tail `[eff_cur, cur_xmin)` is picked up on the next tick — bounded work per run, no loss, no skipped tail. Steady-state intervals are a few seconds of xids (far below the cap) so normal operation is unchanged.

### Validation

- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → passes.
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (steady-state unaffected by the cap).
- SQL text unchanged (only a Rust binding renamed), so no SQLx cache change. No `*_ee.rs` changes.

Pushed as `fix: protect audit s3 checkpoint from config sync and bound export interval`.

rubenfiszel:
## Addressed the latest review

**Pi: "Good to merge"** — confirmed the xid-interval cursor, config-sync protection, and `MAX_XID_INTERVAL` bound all correctly resolve the prior findings.

**Codex found one new P1 (valid): first-enable skip.** The checkpoint was anchored lazily on the *first export tick*, not when the operator enabled the toggle. Audit rows committed in the up-to-~60s window between enabling `store_audit_logs_s3` and that first tick get xids below the first-tick anchor → treated as pre-existing → never mirrored, contradicting the "logs created after enabling are exported" guarantee.

**Fix:** `reload_store_audit_logs_s3_setting` now anchors the cursor at *enable* time. New helper `anchor_audit_logs_s3_checkpoint` does an `INSERT ... ON CONFLICT DO NOTHING` capturing the snapshot xmin at the moment the setting flips on (settings-change notification → reload). Because it's idempotent and never overwrites:

- Transactions in-flight or newer at enable have xid ≥ the anchor → exported in a future interval; only pre-enable committed rows sit below it (intentionally not backfilled).
- HA-safe: first server to react wins; the exporter's own first-run anchor remains the fallback (e.g. env-var enable, no notification).
- Disable → re-enable resumes from the preserved cursor (no overwrite), catching up the gap (bounded by `MAX_XID_INTERVAL`/tick).

Added a regression scenario to `audit_export_end_to_end`: drop the checkpoint, call the enable-time anchor, insert rows *after* the anchor but *before* the first export tick, run the exporter, and assert those rows are exported (lazy first-tick anchoring would skip them).

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (incl. the new enable-flow scenario).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- SQLx cache regenerated via the safe procedure (1 new query, **0 EE-cache deletions**). No `*_ee.rs` changes.

Pushed as `fix: anchor audit s3 checkpoint at enable time to not skip first-window rows`.

rubenfiszel:
## Addressed the latest review

The other automated reviewer's second comment was an empty/truncated post (no findings). Codex found one valid residual P1; the second stray `github-actions` line had no content.

**Codex P1 — enable-time anchor still in the async reload path.** Correct: my previous fix anchored from `reload_store_audit_logs_s3_setting` (driven by the notify-event poll), so rows committed between the setting write and the notify-driven reload still got an xmin below the checkpoint and were skipped. Window shrunk, not closed.

**Fix (their option b — a boundary that cannot jump over post-enable rows):** removed the reload-path anchor entirely (it would *re-introduce* the race by creating a checkpoint at reload time). The first run (no checkpoint) now anchors `prev_xmin` to the **`xmin` of the `store_audit_logs_s3` settings row itself** — i.e. the enable transaction's own xid, fixed atomically by the write that flipped the toggle on:

- Every transaction that committed after the enable write has xid > that xmin, so it's in `[enable_xid, cur)` and is exported — regardless of how long after enabling the first export tick runs. Race fully closed, no timestamp/timezone reasoning needed.
- Pre-enable rows have xid below it → not backfilled (guarantee preserved).
- Falls back to the current snapshot xmin when there is no settings row (env-var enable — no precise enable event exists).
- Disable→re-enable still resumes from the preserved checkpoint (the enable-xid path is first-run-only; a checkpoint, once written, is never re-anchored and is protected by `HIDDEN_SETTINGS`).
- The `MAX_XID_INTERVAL` cap is skipped on the first run (the settings-row xmin is a raw 32-bit xid, a different scale from the 64-bit `txid` `cur` uses; first-run volume is just the short enable→first-tick window).

**Test:** the `audit_export_end_to_end` enable-flow scenario now exercises the *real* flow — insert a pre-enable row, write the `store_audit_logs_s3` row (sets its xmin), insert post-enable rows, then run the first export with **no checkpoint**; asserts the post-enable rows are exported and the pre-enable row is not.

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (incl. real enable-flow).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- SQLx cache regenerated via the safe procedure (**0 EE-cache deletions**). No `*_ee.rs` changes.

Pushed as `fix: anchor first audit export at the enable transaction's xid`.

rubenfiszel:
## Addressed Codex P1 (first-run timestamp floor)

Valid. On first run `last_ts` was set to `now()`, and the query pruned with `timestamp >= last_ts - 7 days`. That pruning predicate is meant only for partition selection, but it gated correctness: if the first successful export is delayed > 7 days (object store missing / PUTs failing before any checkpoint is written), post-enable rows older than the moving 7-day window were filtered out and the run still persisted `last_xmin = eff_cur_xmin`, so the xid cursor advanced past them → permanent loss.

**Fix:** the timestamp floor is now decoupled from the xid cursor and never gates correctness:

- **First run** (no checkpoint): floor = **epoch** (no timestamp pruning). The post-enable backlog can be arbitrarily old in a delayed-first-export/recovery case; the xid predicate (`age(xmin) <= age(enable_xid)`) still restricts the scan to post-enable rows, so it's a one-time/rare full scan, not a backfill of pre-enable history.
- **Steady state**: floor = `last_ts - 7 days`, where `last_ts` is the last *exported* timestamp and is **frozen while no rows are exported** (e.g. during a long-txn stall), so the floor can never advance past an un-exported in-interval row. The 7-day slack still exceeds any realistic audit-writing transaction.
- First-run `last_ts` baseline stays `now()` so that, once the (fully-drained) first run completes, subsequent runs get tight `now()-7d` partition pruning again.

This keeps partition pruning for the hot steady-state path while guaranteeing the floor can never drop a row the xid cursor will skip.

**Test:** the enable-flow scenario now also inserts a post-enable row with a 10-day-old timestamp (into a manually-created past partition) and asserts it IS exported on the first run — with the previous `now()-7d` floor it would have been silently dropped.

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (incl. old-backlog regression).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- SQLx cache regenerated via the safe procedure (**0 EE-cache deletions**). No `*_ee.rs` changes.

Pushed as `fix: use epoch timestamp floor on first audit export run to not drop old backlog`.

rubenfiszel:
## Addressed Codex P1 (env-var enable path)

Valid. With `STORE_AUDIT_LOGS_S3=true` via env var and no `store_audit_logs_s3` settings row, the exporter's first-run fell back to `prev_xmin = cur_xmin`, so audit rows committed between process startup and the first successful tick (≥~60s, or longer if object storage was down) had `xmin < cur_xmin` and were permanently skipped. The settings-row-xmin anchor only covers UI/API enables.

**Fix:** anchor the cursor at **process startup** for the env-var path. `initial_load` (server_mode, parquet) now calls `anchor_audit_logs_s3_checkpoint_env_var`, which does:

```sql
INSERT INTO global_settings (name, value)
SELECT <checkpoint>, jsonb_build_object('last_xmin', txid_snapshot_xmin(txid_current_snapshot())::bigint, ...)
WHERE NOT EXISTS (SELECT 1 FROM global_settings WHERE name = 'store_audit_logs_s3')
ON CONFLICT (name) DO NOTHING
```

- `WHERE NOT EXISTS (store_audit_logs_s3)` → only the **pure env-var** case; when the setting is persisted (UI/API), this is a no-op and the exporter's first-run still uses the precise settings-row xmin (no regression to the path fixed in the previous round).
- `ON CONFLICT DO NOTHING` + never-overwrites → idempotent, HA-safe (first node wins), and disable→re-enable still resumes from the preserved checkpoint.
- Closes the loss window from ~60s+ down to the startup sequence (sub-second; this server isn't serving audit-producing traffic before `initial_load`). Rows committed after startup → exported; pre-startup rows → not backfilled (correct env-var semantic).

**Test:** new env-var scenario in `audit_export_end_to_end` — delete both the checkpoint and the `store_audit_logs_s3` row, insert a pre-startup row, run the startup anchor, insert post-startup rows, export; asserts post-startup rows exported and the pre-startup row not.

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (incl. env-var regression).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- SQLx cache regenerated via the safe procedure (**0 EE-cache deletions**). No `*_ee.rs` changes.

Pushed as `fix: anchor audit export at startup for env-var enable path`.

rubenfiszel:
## Addressed Codex P1 (straddling-transaction loss)

Valid and a real loss. The previous fix anchored the first-run cursor at the **settings row's own xmin** (the enabling transaction's xid). But `audit_log()` is often called with the caller's `&mut tx`, and that caller transaction can have acquired its xid *before* the enable (an earlier non-audit write) yet commit the audit row *after* enable. Such a row's `xmin` is below the enabling txn's xid, so `age(xmin) <= age(prev_xid)` rejected it and the cursor advanced past it → permanent loss.

**Root cause:** the settings-row xid is the wrong boundary. The correct one is the **`txid_snapshot_xmin` of the enabling transaction itself** — a transaction in flight at enable has xid >= that snapshot xmin even if its xid was acquired earlier, so it's correctly included; transactions settled before enable are below it and not backfilled.

**Fix:** capture that boundary atomically, in the enabling transaction, via a DB trigger (new reversible migration `audit_logs_s3_anchor_on_enable`):

- `AFTER INSERT OR UPDATE OF value ON global_settings WHEN (NEW.name='store_audit_logs_s3')` → on enable, `INSERT` the checkpoint with `txid_snapshot_xmin(txid_current_snapshot())` and a `1970-01-01` "bootstrap, not yet exported" sentinel, `ON CONFLICT DO NOTHING`. A trigger runs in the *same transaction as every write path* (per-key API, bulk instance-config sync, direct) — exactly what was asked for. `WHEN` clause prevents re-entrancy; `DO NOTHING` keeps it idempotent/HA-safe and disable→re-enable resumes from the preserved cursor.
- Env-var enable (no settings write) keeps the startup anchor, now also writing the bootstrap sentinel + snapshot xmin.
- The exporter no longer reads the settings-row xmin. It derives `bootstrap` from the sentinel: bootstrap runs use an epoch floor (no pruning) and are uncapped (the previously-fixed delayed-backlog correctness), then record a real `last_ts` so steady-state pruning + the per-run xid cap resume.
- Trade-off (documented): a concurrently long-running transaction can lower the enable snapshot xmin and cause a *bounded over-export* of recently-committed rows — never loss, never full-history backfill. For an audit mirror, completeness > minor over-inclusion.

**Test:** new straddling-transaction scenario — `t` does a non-audit write (low xid) before enable, the toggle is enabled in another txn (trigger fires, captures snapshot xmin ≤ t's xid), then `t` audit-inserts and commits; asserts the row IS exported and genuinely-pre-enable rows are not. Existing enable-flow/env-var/old-backlog/divergent scenarios still pass.

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (all scenarios, incl. the trigger migration applied via `#[sqlx::test]`).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- New reversible migration; SQLx cache regenerated via the safe procedure (**0 EE-cache deletions**). No `*_ee.rs` changes.

Pushed as `fix: anchor audit export via enabling-txn snapshot xmin trigger`.

rubenfiszel:
## Addressed Codex P1 (unbounded bootstrap export)

Valid. The bootstrap path was uncapped, so if object storage is down right after enable, the post-anchor backlog accumulates under the sentinel and the first successful export materializes it all in one `string_agg`/result set under the advisory-lock transaction — the exact failure `MAX_XID_INTERVAL` exists to prevent.

The original reason bootstrap was left uncapped (a 32-bit vs 64-bit xid scale mismatch) **no longer applies**: the previous round removed the settings-row `xmin::text::bigint` read, and the checkpoint `last_xmin` is now always written by the trigger/startup-anchor as `txid_snapshot_xmin(...)::bigint` — the same 64-bit `txid` scale as `cur_xmin`. So the cap is meaningful on the bootstrap path too.

**Fix (exactly Codex's suggestion — keep the epoch floor, advance only to a bounded `eff_cur`, stay in bootstrap until the tail drains):**

- `eff_cur_xmin = cur_xmin.min(prev_xmin + MAX_XID_INTERVAL)` now applies to **every** run, bootstrap included.
- Bootstrap runs still use the epoch timestamp floor (old/delayed backlog not pruned away).
- `bootstrap_drained = bootstrap && eff_cur_xmin >= cur_xmin`. On checkpoint write:
  - capped tail remains (`bootstrap && !drained`) → keep the `1970-01-01` sentinel → next tick continues bootstrap (epoch floor) for the next bounded slice;
  - backlog fully drained (`bootstrap && drained`) → write a real `now()` last_ts → steady-state pruning + cap take over;
  - steady → `batch_max_ts` (frozen when no rows, stall-safe) as before.

Net: a large backlog accumulated under the bootstrap sentinel now drains in bounded `MAX_XID_INTERVAL` slices over successive ticks, never one unbounded aggregate, with no loss and old timestamps preserved.

### Validation
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (all scenarios; the cap is a no-op at test xid volumes so bootstrap exits immediately once caught up, exercising the `bootstrap_drained` path).
- `cargo test -p windmill-common diff_global_settings_replace_keeps_audit_s3_checkpoint` → still passes.
- SQL text unchanged (Rust-only logic) → no SQLx cache change. No `*_ee.rs` changes; no migration change.

Pushed as `fix: bound the bootstrap audit export to MAX_XID_INTERVAL per tick`.

rubenfiszel:
## Refactor: move the export cursor out of global_settings into background_task_state (+ status endpoint)

Several prior review rounds were spent papering over a root cause: the export cursor was stored in `global_settings`, which is conceptually user-configurable instance config (exposed via the settings UI/export and config sync). That mismatch is exactly why we needed the `HIDDEN_SETTINGS` hack and hit the config-sync-deletion P1. The codebase already has the right home — `background_task_state` — whose own migration comment calls out this anti-pattern, and which `log_cleanup`/`storage_usage` use.

**Changes:**
- Cursor now lives in `background_task_state` (row `audit_logs_s3_export`, shared const `AUDIT_LOGS_S3_EXPORT_TASK`). The advisory-xact-lock mutual exclusion and the snapshot-xmin/bootstrap/cap correctness logic are unchanged — only the storage backend moved.
- The enable trigger and the env-var startup anchor now write `background_task_state`.
- **Removed** the `HIDDEN_SETTINGS` entry and the `diff_global_settings_replace_keeps_audit_s3_checkpoint` test — `background_task_state` is not instance config, so the entire config-sync-deletion / config-leak class of bugs is gone by construction (no hack needed).
- **New operator overview:** `GET /settings/audit_logs_s3_status` (super-admin), mirroring `log_cleanup_status`, returns `last_xmin`, `last_ts`, `bootstrapping`, **`last_exported_audit_ts`** (the latest audit timestamp actually uploaded — the "how current is the mirror" figure you asked about), `last_run_at`, `last_run_exported`, `updated_at`, `owner`. Added to `openapi.yaml`; client regenerates (gen dir is gitignored).
- Exporter now also records those observability fields on each run.

**Validation:**
- `cargo test --features parquet --bin windmill audit_export_end_to_end` → passes (all scenarios: bootstrap/cap, divergent xid/id, straddling txn, enable-flow, env-var, old-backlog — now against `background_task_state`; the trigger migration runs via `#[sqlx::test]`).
- `cargo check` (default), `cargo check -p windmill-common`, `cargo check -p windmill-api-settings --features parquet` → pass.
- SQLx cache regenerated via the safe procedure (**0 EE-cache deletions**). Frontend client regenerates cleanly (no new type errors). No `*_ee.rs` changes. The trigger migration was edited in place (unreleased) — down migration unchanged.

This both answers "can the user see the latest uploaded timestamp" (yes — the new status endpoint) and "did we avoid using global_settings as a state store" (now yes — and it retroactively removes the hacks that several earlier review rounds added).

Pushed as `refactor: store audit export cursor in background_task_state, add status endpoint`.

rubenfiszel:
## Addressed P2 (setting text vs enable-boundary contract)

Valid. The description claimed "only audit logs created after enabling are exported," which understates the documented backend contract (the enable boundary is the enabling transaction's snapshot xmin; a concurrently long-running transaction can cause a *bounded over-export* of just-prior committed rows — intentional, since for an audit mirror completeness > minor over-inclusion).

Updated `store_audit_logs_s3` description to: pre-existing history is not backfilled; export starts from when the setting is enabled (transactions in flight at that moment may include a bounded set of just-prior rows); no audit log committed after enabling is ever skipped. This matches the actual no-loss / bounded-over-export contract.

Pushed as `docs: align store_audit_logs_s3 setting text with the actual enable-boundary contract`.

rubenfiszel:
## Moved the export core logic to EE (defensibility) + Enterprise-gated

Per the discussion: audit logs are an Enterprise feature, so the export of them is too. Rather than leave the mechanical-but-substantial algorithm in the OSS tree with only a runtime gate, the **proprietary core now lives in `windmill-ee-private`** and is compiled only in `private` builds.

**Changes:**
- Moved `export_audit_logs_to_object_store` + `anchor_audit_logs_s3_checkpoint_env_var` + the full `#[sqlx::test]` integration test into `windmill-ee-private/src/ee.rs` (the bin crate's `private`-gated EE module, symlinked as `backend/src/ee.rs`). The cursor/snapshot-xmin/bootstrap/cap algorithm is no longer in the open-source tree.
- `backend/src/ee_oss.rs`: `#[cfg(all(parquet, not(private)))]` no-op stubs (the existing `#[cfg(private)] pub use crate::ee::*` re-exports the real ones in EE builds).
- `backend/src/monitor.rs`: deleted the moved code; the `monitor_db` tick and the `initial_load` env-var anchor now call `crate::ee_oss::*` **behind an Enterprise-license gate** (`get_license_plan() == Enterprise`), mirroring how `audit_log()` itself is license-aware. This also closes the earlier 'ungated' concern.
- Trigger migration, status endpoint, setting, and frontend toggle stay OSS (trivial, not core logic).

**EE PR workflow (per docs/enterprise.md):**
- This PR is now retitled `[ee]`.
- Companion PR: windmill-labs/windmill-ee-private#579 (branch `audit-logs-s3-folder`).
- `backend/ee-repo-ref.txt` bumped to the EE branch commit `4ffc6d5f` (verified = EE branch HEAD, not main).

**Validation:**
- `cargo check` (default) and `cargo check --features parquet` (OSS no-op path) pass.
- `cargo test --features parquet,private,enterprise --bin windmill audit_export_end_to_end` → passes (test now at `ee::audit_s3_export::audit_s3_export_tests`; all scenarios green: bootstrap/cap, divergent xid/id, straddling txn, enable-flow trigger, env-var, old-backlog).
- SQLx cache regenerated with EE features (`all_sqlx_features,private,enterprise`); **0 net deletions** vs main. Only `src/ee.rs` committed in the EE repo (unrelated pre-existing dirty files left untouched).

Net: the proprietary export algorithm is out of the OSS tree (defensible) and only runs on an Enterprise license.

PR #9216, opened 2026-05-18
feat: add global chat selected context

centdix:
## Summary
Adds selected workspace context support to global AI chat while keeping the existing workspace search path. Workspace script/flow selections are now stored as references only, and prompts serialize selected workspace items as type/path so the AI can fetch contents through tools when needed.

## Changes
- Enable the shared context picker UI in global mode.
- Preserve selected workspace script/flow context for global mode.
- Stop hydrating selected workspace scripts/flows in the context picker.
- Serialize selected workspace items as type/path references only in global, script, and flow prompts.
- Keep the DB-resource cache freshness marker separate from the global context workspace marker.
- Add prompt/context formatting tests and DB context cache regression tests.

## Test plan
- [x] npm run test:unit -- src/lib/components/copilot/chat/ContextManager.test.ts src/lib/components/copilot/chat/global/core.test.ts src/lib/components/copilot/chat/shared.test.ts
- [ ] npm run check:fast (currently fails after rebasing on origin/main with unrelated type errors in SuperadminSettingsInner.svelte and TokensTable.svelte)

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatInput.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatInput.svelte
@@ -90,7 +90,11 @@
 	let appTooltipCurrentViewNumber = $state(0)
 
 	export function focusInput() {
-		if (aiChatManager.mode === AIMode.SCRIPT || aiChatManager.mode === AIMode.FLOW) {
+		if (
+			aiChatManager.mode === AIMode.SCRIPT ||
+			aiChatManager.mode === AIMode.FLOW ||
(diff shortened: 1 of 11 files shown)


centdix:
Addressed the review about global context poisoning the DB-resource cache in `f4fd636cf1`. `ContextManager` now tracks DB-resource freshness separately from the global context workspace marker, so opening or switching workspaces through Global mode no longer causes Script/Flow mode to skip DB resource loading or reuse stale resources. Added regression coverage for Global -> Script and workspace switching through Global in `ContextManager.test.ts`.

PR #9221, opened 2026-05-18
fix: scope VSCode webview clipboard paste to focused editor

diegoimbert:
## Summary

Fixes copy/paste not working (and pasting into the wrong editor) in the VS Code extension webview. The webview is a sandboxed iframe where `navigator.clipboard.readText()` and the native `paste` event's `clipboardData` are both empty/restricted — the only reliable clipboard read is focusing a real `<input>` and running `document.execCommand('paste')`. The previous code registered paste via `editor.addCommand(Ctrl+V)`, which uses a Monaco standalone keybinding registry shared across all editor instances: the last-registered handler won, so paste landed in the wrong editor (or, after an interim attempt using the native `paste` event, did nothing at all).

## Changes

- Add a shared `registerWebviewPaste(container, getEditor)` helper in `frontend/src/lib/editorUtils.ts` that scopes a capture-phase `keydown` listener to each editor's own container DOM node, gated on `editor.hasTextFocus()`, and pastes via a focused hidden `<input>` + `execCommand('paste')` into the focused instance. No-ops outside an iframe. Returns a cleanup function.
- Wire the helper into `Editor.svelte`, `TemplateEditor.svelte`, and `DiffEditor.svelte` (replacing the buggy global `addCommand(Ctrl+V)` + `navigator.clipboard.readText()` blocks); cleanup is run on destroy and guarded against re-entry.
- `SimpleEditor.svelte` keeps an equivalent inline implementation (it relies on Svelte bindings) restored to the working hidden-input mechanism, now triggered by a scoped keydown listener instead of the global keybinding.
- Bail instead of falling back to range `{1,1,1,1}` when no selection is present, so a lost selection can never silently paste at document start.
- Ctrl+C / Ctrl+X are left on `addCommand` — `execCommand('copy'/'cut')` acts on the truly-focused DOM selection, so the global-keybinding leak is harmless there.

Not affected: `GraphqlSchemaViewer.svelte` (read-only, copy only) and `Dev.svelte` (page-level keydown handler, not the multi-instance Monaco bug).

## Test plan

- [ ] In the VS Code extension webview, with two editors on the same view (e.g. a script `Editor` + a `SimpleEditor`), click into one, copy text, Ctrl/Cmd+V — paste lands in the focused editor at the caret, not in the other editor and not at document start
- [ ] Repeat with focus in the second editor — confirms the cross-instance bug is gone
- [ ] Test paste in `TemplateEditor` (app/template editor) and an inline `DiffEditor`
- [ ] Close/reopen the editor or diff panel several times; confirm Ctrl+V still works and no orphaned `aria-hidden` `<input>` nodes accumulate under `document.body`
- [ ] Sanity-check a normal (non-webview) browser session — native paste unaffected since the helper no-ops outside iframes


diegoimbert:
Tried and newlines work fine (and using textarea breaks it)

PR #9224, opened 2026-05-18
fix: enforce scheme allowlist for MCP OAuth redirect URIs

rubenfiszel:
## Summary

Tightens validation of OAuth redirect URIs in the MCP OAuth authorization server to follow RFC 8252 §7.3 and the OAuth 2.0 Security BCP. Previously a redirect URI was only checked for non-emptiness at registration and exact-matched against the registered value at authorize/approve, with no constraint on the URI scheme.

## Changes

- Add a shared `validate_redirect_uri` helper: allows absolute `https://` URIs, and `http://` only for loopback hosts (`localhost`, the IPv4 `127.0.0.0/8` range, and `[::1]`). Rejects any other scheme, URL fragments, and embedded credentials (userinfo).
- Apply it at every redirect-URI boundary: dynamic client registration, authorize, approve, the token grant, and the OAuth error-redirect response (which returns JSON rather than a redirect when the target is not allowed).
- Re-validate registered URIs at authorize/approve/token so clients registered before this check cannot reach a redirect sink.
- Frontend consent page now builds the redirect with `new URL()` plus an explicit protocol allowlist, refuses to navigate when the URI is not allowed, and shows an upfront error state instead of a functional consent screen for an invalid/unsafe redirect URI.
- Keep the backend and frontend loopback rules consistent (both accept the full `127.0.0.0/8` range).
- Add unit tests for the validator covering accepted and rejected forms.

## Test plan

- [ ] `cargo test -p windmill-api --features mcp mcp::oauth_server::tests` — all validator unit tests pass
- [ ] Backend compiles with `--features mcp`
- [ ] `npm run check` passes for the modified consent page
- [ ] Register a client with `https://` and with `http://127.0.0.1:<port>` callbacks — authorize/approve/token flow completes
- [ ] Register attempt with a non-http(s) scheme, a fragment, or embedded credentials is rejected with a 400
- [ ] Consent page with a malformed/disallowed `redirect_uri` query param renders the error state and never navigates

---
## Summary by cubic
Enforces a strict allowlist for OAuth redirect URIs in the MCP auth server to block unsafe schemes and open redirects. Validation follows RFC 8252 §7.3 and the OAuth 2.0 Security BCP and is applied across the full flow, with the consent page refusing unsafe redirects.

- **Bug Fixes**
  - Add `validate_redirect_uri`: allow `https`; allow `http` only for loopback (`localhost`, `127.0.0.0/8`, `[::1]`); reject fragments and userinfo.
  - Apply at registration/authorize/approve/token; invalid targets return JSON (no redirect).
  - Re-validate stored URIs so older clients cannot use poisoned values.
  - Consent page uses `new URL()` + protocol allowlist; blocks navigation and shows an error when invalid; use `URLSearchParams.forEach` for wider runtime compatibility.
  - Align frontend and backend loopback rules; add unit tests covering allowed and rejected forms.

<sup>Written for commit 42feec1b2c5c08a875f899deb1f0bff95478c3ff. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9224?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/windmill-api/src/mcp/oauth_server.rs
+++ b/backend/windmill-api/src/mcp/oauth_server.rs
@@ -9,6 +9,7 @@ use axum::{
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use sqlx::FromRow;
+use url::{Host, Url};
 use windmill_common::{
     auth::{hash_token, TOKEN_PREFIX_LEN},
     error::{Error, Result},
@@ -296,6 +297,10 @@ pub async fn oauth_register(
         ));
     }
 
+    for redirect_uri in &req.redirect_uris {
+        validate_redirect_uri(redirect_uri)?;
+    }
+
     let client_id = format!("mcp-client-{}", rd_string(16));
 
     sqlx::query!(
@@ -374,6 +379,12 @@ async fn handle_authorization_code_grant(
 
     if auth_code.redirect_uri != req.redirect_uri {
         return Err(OAuthTokenError::invalid_grant("redirect_uri mismatch"));
+    }
+
+    if validate_redirect_uri(&auth_code.redirect_uri).is_err() {
+        return Err(OAuthTokenError::invalid_grant(
+            "redirect_uri bound to this code is not allowed",
+        ));
     }
 
     let challenge = auth_code.code_challenge.as_ref().ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Closing as redundant. This is already fixed and disclosed on `main` by #9197 (GHSA-q9xg-f2v2-695g), which validates dynamically-registered redirect_uris at registration and adds a defense-in-depth gate on the consent page (covering both the direct-link and pre-existing-client-row vectors).

Reviewed main's fix: the critical one-click script-execution / token-exfiltration path is genuinely closed. The only deltas this branch had over #9197 were incremental hardening (re-validating at the authorize/approve/token/error-redirect sinks rather than registration-only, and rejecting fragments/embedded credentials) — not an exploitable gap, and not worth a competing implementation given #9197 deliberately uses a denylist to keep native MCP clients (vscode://, cursor://) working.

PR #9235, opened 2026-05-19
fix: enforce auth guards on app component preview execution

rubenfiszel:
## Summary

`POST /api/w/:workspace/apps_u/execute_component/:path` runs in **preview mode** whenever the client supplies `force_viewer_static_fields`. In that mode it accepts request-supplied `raw_code` and enqueues it as a `Viewer`-mode job — it is the app-editor equivalent of `/jobs/run/preview`.

The bug: this preview branch did **not** apply the guards that the canonical `run_preview_script` (`/jobs/run/preview`) enforces for arbitrary code execution. `force_viewer_static_fields` is client-controlled, so an authenticated **Operator** (a run-only user who must not be able to create scripts/apps or run preview jobs) — or a narrowly-scoped token — could send a single request with `force_viewer_static_fields` + `raw_code` and **enqueue arbitrary worker code**, fully bypassing the Operator restriction.

This was reproduced end-to-end before the fix: an Operator user enqueued and executed `export async function main(){ return "PWNED..." }` (`job_kind: preview`, `created_by: operator`, `success: true`).

**Root cause:** a long-standing design flaw, not a recent regression. `execute_component` serves both deployed apps (run mode) and the app editor (preview mode) through one unauthed-router endpoint; the preview branch (present since the earliest commit in current history, `ce290f68db`) was implicitly trusted to be editor-only and never re-applied the operator/scope/path/isolation guards. The sibling preview endpoints in `jobs.rs` were independently hardened (see comments at `jobs.rs:5646`) but `execute_component`'s preview branch was missed.

## Changes

- **`backend/windmill-api/src/apps.rs`**: in `execute_component`, gated on `is_preview` (`force_viewer_static_fields.is_some()`), the preview branch enforces the security-relevant guards of `/jobs/run/preview`: authenticated caller, **not an Operator**, **`jobs:run` scope** (the route is reachable with an `apps:run`-scoped token since it maps to the `apps` scope domain, so without this an `apps:run` token could escalate to arbitrary code execution), and `check_tag_available_for_workspace` for the client-supplied inline tag (so a `if_jobs:filter_tags`-restricted token can't escape its filter). Inline `raw_code` is otherwise not path-gated — a non-operator can already run arbitrary inline code via `/jobs/run/preview`. When a preview instead *references* an existing runnable the caller may not be allowed to read, that reference is confined to a readable path: `payload.path` (the deployed `script/`/`flow/` path) and `payload.id` (a persisted `app_script`, restricted to an app in the workspace whose path the caller can read). The enqueue stays root-isolated (`IsolatedRoot`) exactly as before — in preview mode the policy is forced to `ExecutionMode::Viewer` so the job already runs as the requesting user; switching the insert to user-RLS is not what contains the bypass and would add avoidable breakage risk to the legitimate editor flow. Run mode (no `force_viewer_static_fields`) is byte-for-byte unchanged.
- **`backend/tests/app_preview_auth.rs`** + **`backend/tests/fixtures/app_preview_auth.sql`**: regression test covering the full matrix — Operator rejected (core regression), non-operator editor preview still works, cross-namespace preview rejected (defense-in-depth), run mode unaffected.

## Verification

- **e2e before fix**: Operator request → `HTTP 200` + job UUID; the arbitrary code executed (`result: PWNED-arbitrary-code-by-operator`).
- **e2e after fix**: Operator request → `HTTP 401` "Operators cannot run preview jobs for security reasons"; legitimate non-operator editor preview still returns a job UUID and runs; cross-namespace preview → `HTTP 400`; run mode → unchanged (`404` for missing app).
- **Regression test deterministic split** (empirically verified): pre-fix `FAILED` (Operator got `200` + UUID, expected `401`); post-fix `ok`.

## Test plan

- [ ] `cargo test --test app_preview_auth` passes on this branch
- [ ] Revert only `backend/windmill-api/src/apps.rs` and confirm `test_app_preview_authorization` fails (Operator gets 200)
- [ ] Manual: as a non-operator, run an inline-script component preview in the app editor — still works
- [ ] Manual: as an Operator, run a component preview — rejected with "Operators cannot run preview jobs for security reasons"
- [ ] Manual: a deployed (run-mode) app component still executes normally per its policy

---
## Summary by cubic
Enforced auth on app component preview execution to close Operator and scope bypasses. In preview (`force_viewer_static_fields`), we now require `jobs:run`, block Operators, confine referenced runnables to readable paths, and honor worker tag filters; run mode and enqueue isolation stay unchanged.

- **Bug Fixes**
  - Require `jobs:run` and reject Operators in preview; prevent `apps:run`-only tokens from enqueuing arbitrary preview code.
  - Gate referenced resources: verify read access for `payload.path` (`script/` or `flow/`) and ensure `app_script` `id` belongs to a readable app in the workspace.
  - Enforce worker tag availability for client-supplied inline tag; respect token tag filters.
  - Keep inline `raw_code` previews allowed for non-operators; no path gate added.
  - Add regression tests for operator rejection, `apps:run` escalation block, valid editor preview, cross-namespace runnable checks, `app_script` ownership, and unchanged run-mode/isolation.

<sup>Written for commit 23cf31488c8c325b3f2782e7b6d621551a7c9100. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9235?utm_source=github">Review in cubic</a></sup>


rubenfiszel:
Addressed the CI review in 6607007d12.

**Codex P1 (preview path guard checked app path, not runnable path) — fixed.**
This was a real residual gap. In preview mode `payload.path` (a deployed `script/…` or `flow/…`) is resolved by `get_payload_tag_from_prefixed_path` with the root DB handle, bypassing the policy's triggerables. The guard now derives the runnable path (stripping the `script/`/`flow/` prefix, mirroring `get_payload_tag_from_prefixed_path`) and checks `require_path_read_access_for_preview` against *that*; for inline `raw_code` it keeps the app URL path (the raw code lives under the app namespace). Verified e2e: an allowed app path `u/dev/myapp` paired with `path: script/u/admin/private` → now `400` (was `404`, i.e. it had resolved and would have enqueued a real private runnable). Own-namespace deployed-runnable preview still passes the guard (proven: `script/u/dev/ownscript` → `404 script not found`, not a `400`). Added regression scenario 5 to `app_preview_auth.rs` with an empirically verified deterministic split (fails pre-fix with `404`, passes post-fix with `400`).

**Claude P2 (parity: missing `check_tag_available_for_workspace`) — fixed.**
Since preview honors the client-supplied inline tag (`resolved_inline_tag`), the preview path now calls `check_tag_available_for_workspace(&db, &w_id, &tag, authed)` once the tag is resolved, matching `run_preview_script`.

**`check_license_key_valid` (Claude P2, EE) — intentionally out of scope.** It is an EE licensing gate, not an authorization control in the auth-bypass class this PR fixes; adding it would change app-editor preview UX on unlicensed EE instances and belongs in a separate decision. The PR description has been kept accurate (the wording is "the security-relevant guards", not literal line-for-line parity).

**Claude's scoped-token test note.** A scoped token without `jobs:run` never reaches this handler: the `apps_u` route's `ApiAuthed` extractor rejects out-of-scope tokens, so `OptAuthed` yields `None` and the request is rejected with "App component preview requires authentication" before `check_scopes`. The security property (scoped tokens cannot run preview) holds; it just can't be deterministically pinned on the `check_scopes` line through this route, so I did not add a test that would assert on an incidental enforcement layer.

Full local regression suite (`cargo test --test app_preview_auth`) passes (5/5) post-fix.

rubenfiszel:
Addressed the latest review in 3f166fc05e.

**Codex P1 (persisted inline-script preview can execute arbitrary `app_script` ids) — fixed.**
Confirmed real: with `force_viewer_static_fields` + `raw_code` + caller-controlled `id`, the job becomes `JobPayload::AppScript { id }` and the worker fetches `app_script` content by id alone, so an allowed app path could be paired with another private app's `app_script` id. The guard now resolves the owning app of `payload.id` (`app_script JOIN app`) restricted to the current workspace and runs `require_path_read_access_for_preview` against that owner path; an id that doesn't belong to a workspace app, or belongs to an app the caller can't read, is rejected before enqueue. Added regression scenario 6 (cross-app id → 400) and scenario 7 (id owned by an app in the caller's own namespace still enqueues), both with empirically verified deterministic splits (scenario 6 returned `200` + job UUID pre-fix — the actual bypass — and `400` post-fix). SQLx offline cache updated (one new entry, no other `.sqlx` churn).

**Push isolation reverted to root (`IsolatedRoot`) — deliberate.**
Per maintainer direction, and on reflection this is correct: in preview mode the policy is forced to `ExecutionMode::Viewer`, so `get_on_behalf_details_from_policy_and_authed` already resolves the job's identity (`email`/`permissioned_as`/`username`) to the requesting user — the job runs as the caller regardless of the enqueue isolation level. `Isolated(user_db, authed)` only changes the RLS context of the *insert* transaction; it is not what contains the bypass (the auth guards are: authenticated + not operator + `jobs:run` scope + runnable read-access + app_script id ownership), and switching the insert to user-RLS adds avoidable breakage risk to the legitimate editor flow for no added containment here. Run mode is unchanged (it always used root). The PR description's parity wording is scoped to the security-relevant guards, not the isolation level.

Pi's prior pass and cubic's no-issues still apply to the unchanged guard logic. Full local regression suite (`cargo test --test app_preview_auth`) passes 7/7 post-fix; each negative scenario verified to fail pre-fix.

rubenfiszel:
Simplified the guard in 5ce43a690c after reconsidering what it actually protects.

A non-operator member can already run arbitrary inline code via `/jobs/run/preview`, so gating the *inline `raw_code`* preview with scope/path/tag checks added no real containment. The guard is now scoped to what it genuinely protects:

- **Always (preview mode):** authenticated + **not Operator**. This is the headline fix and the only mandatory gate.
- **Only when previewing a *referenced* runnable:** read-access checks — `payload.path` (deployed `script/`/`flow/`, prefix-stripped) and `payload.id` (persisted `app_script`, ownership confined to a workspace app the caller can read). These remain because they execute existing code the caller may not be allowed to read — a real escalation that raw_code does not already grant.

Removed: the `jobs:run` `check_scopes` call (dead on this route — scoped tokens are rejected by the `apps_u` `ApiAuthed` extractor before reaching the handler, so it never denied anything), the inline-`raw_code` app-path check (redundant with `/jobs/run/preview`), and `check_tag_available_for_workspace` (same class as raw_code; minor). Push stays root-isolated.

Tests updated: scenario 3 now pins that inline `raw_code` preview is deliberately *not* path-gated (guards against re-adding an over-restrictive check); scenarios 1 (operator → 401), 5 (`payload.path` cross-namespace → 400), 6 (cross-app `app_script` id → 400) and 7 (own-app id → 200) unchanged. Full suite 7/7 post-change; the operator split re-verified (operator enqueues a job pre-guard, 401 post-guard).

rubenfiszel:
The Codex/Pi P1 was correct — addressed in 23cf31488c.

Pi's trace is right: `apps_u/execute_component` maps to `ScopeDomain::Apps` + `ScopeAction::Run` (scopes.rs:356), and the kind-gating at scopes.rs:671 only applies to `ScopeDomain::Jobs`, so an `apps:run`-scoped token passes the route-level check, reaches the handler with `is_operator = false`, and (before this commit) could enqueue arbitrary inline `raw_code` — an `apps:run → jobs:run` escalation. My earlier "dead on this route" rationale was based on an incomplete test that only used `jobs:`-domain scoped tokens (those are rejected by the extractor because the route needs an `apps`-domain scope); I never tested an `apps:run` token. Mea culpa.

Restored both guards inside the `if is_preview` block:
- `check_scopes(authed, || "jobs:run")` after the operator check.
- `check_tag_available_for_workspace(&db, &w_id, &tag, authed)` after the tag is resolved (same scoped/`if_jobs:filter_tags`-token escape class Codex flagged).

Added test scenario 8: a non-operator token scoped `{apps:run}` (fixture `APPS_RUN_TOKEN`) sending `force_viewer_static_fields` + inline `raw_code` → now **403** "Required scope: jobs:run". Empirically verified deterministic split: with `check_scopes` removed it returns **200 + a job UUID** (the bypass), with it restored **403**. Full suite 8/8 post-fix.

The rest of the simplification stands (Pi independently verified root-push isolation, dropping the inline-`raw_code` path check, and no `preview_authed` are all sound). Branch rebased onto the latest `fix-app-preview-auth-bypass` (main was merged in upstream); regression suite re-run green on the integrated tree.

PR #9240, opened 2026-05-19
feat: add wmill protection-rules pull/push CLI commands

rubenfiszel:
## Summary

Adds `wmill protection-rules pull|push` so workspace protection rulesets can be defined and version-controlled from `wmill.yaml`, mirroring the existing `gitsync-settings` command pattern.

Protection rules are stored under a new `protectionRules` field on `SyncOptions`, so they flow through `getEffectiveSettings` — supporting top-level config **and** per-workspace `overrides`/`promotionOverrides`. This is the deciding reason for choosing `wmill.yaml` over `settings.yaml`: protection rules are the textbook case for environment-differentiated enforcement (strict on the deployed workspace, looser on a fork), and only `wmill.yaml` already has that override machinery.

No backend changes — the `/protection_rules` CRUD API, OpenAPI spec, and generated CLI client already existed; this is a pure CLI feature.

## Changes

- `cli/src/core/conf.ts`: new `ProtectionRuleEntry` type + `protectionRules?: ProtectionRuleEntry[]` on `SyncOptions`
- `cli/src/commands/protection-rules/`:
  - `converter.ts` — normalization + order/duplicate-insensitive `computePlan` (create/update/delete reconcile)
  - `pull.ts` — backend → `wmill.yaml` (replace/override modes, `--diff`, `--json-output`, `--yes`, `--promotion`); creates `wmill.yaml` if absent
  - `push.ts` — `wmill.yaml` → backend, **full reconcile** (create/update/delete) with diff + confirmation; delete-guard prompt defaults to "no" when deletions are present; `--yes`/non-TTY bypass
  - `utils.ts`, `protection-rules.ts`, `index.ts` — display/override helpers and command wiring
- `cli/src/main.ts`: register `protection-rules` command
- `cli/test/protection_rules_converter_unit.test.ts`: 15 converter unit tests
- Regenerated `system_prompts/auto-generated/*` + `cli/src/guidance/skills.gen.ts` per AGENTS.md

## Test plan

- [x] `npm run check` — no new tsc errors (16 pre-existing, all in unrelated files)
- [x] `npm run test:unit` — 15 new converter tests pass; gitsync/settings unit suites unaffected (51 total)
- [x] E2E vs dev backend: `pull` creates `wmill.yaml` from empty backend; `push --diff` then `push --yes` creates rules (verified bitflags in DB); full-reconcile push (create + update + delete) verified against DB; `pull --replace` overwrites stale local; `pull --diff --json-output` shows correct `hasChanges`/structured diff
- [x] Local code review (branch-diff-reviewer) — one P2 addressed: pull `--json-output` diff now uses directional keys (`addedLocally`/`updatedLocally`/`removedLocally`) to avoid ambiguity vs the push framing
- [ ] Reviewer: confirm interactive Select (replace/override/cancel) UX in a git repo with a `workspaces` block

---
## Summary by cubic
Adds `wmill protection-rules pull|push` to sync workspace protection rules via a new per-workspace `protection-rules.yaml`, with full reconcile on push. Replaces the prior `wmill.yaml` overrides to simplify usage and avoid override/dry-run edge cases.

- **New Features**
  - New `protection-rules.yaml`: `{ <workspace>: ProtectionRuleEntry[] }`; keys must match `wmill.yaml`; deterministic key order on write.
  - `pull`: fetch rules into the file; supports `--all`, `--dry-run`, `--json-output`.
  - `push`: reconcile backend to match the file (create/update/delete) with diff + confirmation; supports `--all`, `--dry-run`, `--json-output`, `--yes`.
  - Order/duplicate-insensitive diffing; per-workspace JSON output (`create`/`update`/`delete`/`unchanged`).
  - Per-workspace resolution/auth via `tryResolveBranchWorkspace` + `setClient`; command registered under `wmill protection-rules`; CLI docs generated.

- **Bug Fixes**
  - Honor explicit `--base-url`/`--token` (stateless CI); explicit `--token` overrides a stored profile; workspace id still derives from `wmill.yaml`.
  - Exit non-zero on failures; partial `--all` runs return `success:false` with `partialFailure:true` and push reports applied create/update/delete counts.
  - Warn when an empty local list would wipe backend rules; deletion confirmation defaults to “no” when deletions are present.
  - `pull --dry-run` never writes or creates files.
  - `--json-output` now emits only JSON on stdout by silencing human logs early.

<sup>Written for commit a54d717adcb5ab01febcecd980dc5688a1778a42. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9240?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/cli/src/commands/protection-rules/converter.ts
+++ b/cli/src/commands/protection-rules/converter.ts
@@ -1,0 +1,132 @@
+import { ProtectionRuleEntry } from "../../core/conf.ts";
+import { ProtectionRuleset } from "../../../gen/types.gen.ts";
+
+// Reconciliation plan produced by diffing the wmill.yaml protection rules
+// against the backend list. `toDelete` holds names present on the backend but
+// absent from wmill.yaml (full-reconcile semantics).
+export interface ProtectionRulesPlan {
+  toCreate: ProtectionRuleEntry[];
+  toUpdate: ProtectionRuleEntry[];
+  toDelete: string[];
+  unchanged: string[];
+}
+
+function sortedUnique(arr: readonly string[]): string[] {
+  return [...new Set(arr)].sort();
+}
+
+export class ProtectionRulesConverter {
+  // Canonicalize a single rule so comparisons are insensitive to array order
+  // and duplicates.
+  static normalizeEntry(entry: ProtectionRuleEntry): ProtectionRuleEntry {
+    return {
+      name: entry.name,
+      rules: sortedUnique(entry.rules ?? []) as ProtectionRuleEntry["rules"],
+      bypass_groups: sortedUnique(entry.bypass_groups ?? []),
+      bypass_users: sortedUnique(entry.bypass_users ?? []),
+    };
+  }
+
+  // Canonicalize and sort a list of rules by name.
+  static normalizeList(
+    entries: ProtectionRuleEntry[] | undefined,
+  ): ProtectionRuleEntry[] {
+    return (entries ?? [])
+      .map((e) => ProtectionRulesConverter.normalizeEntry(e))
+      .sort((a, b) => a.name.localeCompare(b.name));
+  }
+
+  // Convert a backend ProtectionRuleset response into the wmill.yaml shape
+  // (drops workspace_id, which is implied by the synced workspace).
+  static fromBackend(rulesets: ProtectionRuleset[]): ProtectionRuleEntry[] {
+    return ProtectionRulesConverter.normalizeList(
+      rulesets.map((r) => ({
+        name: r.name,
+        rules: [...(r.rules ?? [])],
+        bypass_groups: [...(r.bypass_groups ?? [])],
+        bypass_users: [...(r.bypass_users ?? [])],
+      })),
+    );
+  }
+
+  static entriesEqual(
+    a: ProtectionRuleEntry,
+    b: ProtectionRuleEntry,
+  ): boolean {
+    const na = ProtectionRulesConverter.normalizeEntry(a);
+    const nb = ProtectionRulesConverter.normalizeEntry(b);
+    return (
+      na.name === nb.name &&
+      na.rules.length === nb.rules.length &&
+      na.rules.every((v, i) => v === nb.rules[i]) &&
+      na.bypass_groups.length === nb.bypass_groups.length &&
+      na.bypass_groups.every((v, i) => v === nb.bypass_groups[i]) &&
+      na.bypass_users.length === nb.bypass_users.length &&
+      na.bypass_users.every((v, i) => v === nb.bypass_users[i])
+    );
+  }
+
+  static listsEqual(
+    a: ProtectionRuleEntry[] | undefined,
+    b: ProtectionRuleEntry[] | undefined,
+  ): boolean {
+    const na = ProtectionRulesConverter.normalizeList(a);
+    const nb = ProtectionRulesConverter.normalizeList(b);
+    if (na.length !== nb.length) return false;
+    return na.every((entry, i) =>
+      ProtectionRulesConverter.entriesEqual(entry, nb[i])
+    );
+  }
+
+  // Compute the create/update/delete plan to make `backend` match `local`.
+  static computePlan(
+    local: ProtectionRuleEntry[] | undefined,
+    backend: ProtectionRuleEntry[] | undefined,
+  ): ProtectionRulesPlan {
+    const localByName = new Map(
+      ProtectionRulesConverter.normalizeList(local).map((e) => [e.name, e]),
+    );
+    const backendByName = new Map(
+      ProtectionRulesConverter.normalizeList(backend).map((e) => [e.name, e]),
+    );
+
+    const plan: ProtectionRulesPlan = {
+      toCreate: [],
+      toUpdate: [],
+      toDelete: [],
+      unchanged: [],
+    };
+
+    for (const [name, entry] of localByName) {
+      const existing = backendByName.get(name);
+      if (!existing) {
+        plan.toCreate.push(entry);
+      } else if (!ProtectionRulesConverter.entriesEqual(entry, existing)) {
+        plan.toUpdate.push(entry);
+      } else {
+        plan.unchanged.push(name);
+      }
+    }
+
+    for (const name of backendByName.keys()) {
+      if (!localByName.has(name)) {
+        plan.toDelete.push(name);
+      }
+    }
+
+    plan.toCreate.sort((a, b) => a.name.localeCompare(b.name));
+    plan.toUpdate.sort((a, b) => a.name.localeCompare(b.name));
+    plan.toDelete.sort();
+    plan.unchanged.sort();
+
+    return plan;
+  }
+
+  static planHasChanges(plan: ProtectionRulesPlan): boolean {
+    return (
+      plan.toCreate.length > 0 ||
+      plan.toUpdate.length > 0 ||
+      plan.toDelete.length > 0
+    );
+  }
+}
--- a/cli/src/commands/protection-rules/converter.ts
+++ b/cli/src/commands/protection-rules/converter.ts
@@ -1,7 +1,7 @@
-import { ProtectionRuleEntry } from "../../core/conf.ts";
+import { ProtectionRuleEntry } from "./types.ts";
 import { ProtectionRuleset } from "../../../gen/types.gen.ts";
 
-// Reconciliation plan produced by diffing the wmill.yaml protection rules
+// Reconciliation plan produced by diffing the local protection rules
 // against the backend list. `toDelete` holds names present on the backend but
 // absent from wmill.yaml (full-reconcile semantics).
 export interface ProtectionRulesPlan {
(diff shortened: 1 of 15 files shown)


rubenfiszel:
Addressed the review in `3547200` (e2e-verified against a live dev backend):

**Codex P1 — failure paths exited 0.** Added a `fail()` helper (`utils.ts`) that prints via `outputResult` then `process.exit(1)`. All genuine-failure paths in `pull.ts`/`push.ts` (fetch error, missing `wmill.yaml`, no `protectionRules`, non-interactive conflict, outer catch) now exit non-zero. Verified: `push --yes` with no rules → exit 1; bad-workspace fetch → non-zero; successful `--diff` → exit 0.

**Codex P1 — `--override` wrote to an inert key.** `pull.ts` now resolves the workspace key via `findWorkspaceByGitBranch` before writing the override (falling back to the raw branch only when no entry maps to it). Verified: with `workspaces.prod.gitBranch: main`, `pull --override` on `main` writes `workspaces.prod.overrides.protectionRules` (no stray `workspaces.main`), and a subsequent `push --diff`/`pull --diff` correctly reports no changes (override is now effective).

**Claude P2 — `pull --replace` left a shadowing override.** Added `clearRuleOverride()`; replace mode now strips `overrides`/`promotionOverrides` `protectionRules` on the resolved entry and logs that it did so. Verified: with an override that differs from the backend, `pull --replace` sets top-level + clears the override, and the follow-up `pull --diff` reports "No differences" (the previous infinite-loop is gone).

**Claude P2 / non-atomic push.** The reconcile loop now tracks applied create/update/delete counts; a mid-loop failure reports exactly how far it got, notes the backend is partially reconciled (re-run to converge), and exits non-zero. Success counts now come from what was actually applied.

**Note — empty-list wipes all rules.** Added a loud red `WARNING` (shown even under `--yes`/non-TTY) when an empty `protectionRules` would delete every backend rule. Verified the warning fires and the delete still proceeds with the documented full-reconcile semantics.

Added 4 unit tests for `applyRulesToBranchOverride` (resolved-key behavior) and `clearRuleOverride` — 55 unit tests pass, no gitsync/settings regressions, tsc clean on all touched files.

rubenfiszel:
Addressed the latest Codex review in `c871cf8` (e2e-verified against a live dev backend):

**P1 — `pull --diff` created `wmill.yaml`.** Restructured `pull.ts` so the read step tolerates a missing file and the `--diff` branch returns *before* the no-`wmill.yaml` bootstrap. A dry run now never touches the working tree. Verified: `pull --diff` in an empty dir prints `+ r1` and leaves no `wmill.yaml` behind.

**P1 — `pull --promotion` wrote the wrong workspace/block.** The write target now mirrors what `getEffectiveSettings` read for the same invocation: with `--promotion <branch>` it resolves the promotion target via `findWorkspaceByGitBranch(workspaces, promotion)` and writes/clears its **`promotionOverrides`** (added an `OverrideBlock` param to `applyRulesToBranchOverride`); without `--promotion` it stays on the current branch's resolved entry `overrides`. Verified on branch `dev`: `pull --promotion main --override` writes `workspaces.prod.promotionOverrides.protectionRules` (gitBranch:main → `prod`), leaves `dev` untouched, and a follow-up `pull --promotion main --diff` **and** `push --promotion main --diff` both report no changes — read and write are coherent. Non-promotion replace/clear regression re-verified (still clears the shadowing override; `pull --diff` → no differences).

Added 1 unit test for the `promotionOverrides` block; 56 unit tests pass, no gitsync/settings regressions, tsc clean on all touched files.

rubenfiszel:
Reworked per design feedback: protection rules no longer live in wmill.yaml. Pushed in `8673335`.

**New model — a dedicated `protection-rules.yaml`, keyed by workspace:**

```yaml
prod:
  - name: prod-lockdown
    rules: [DisableDirectDeployment, DisableWorkspaceForking]
    bypass_groups: [g/release-managers]
    bypass_users: []
dev:
  - name: dev-soft
    rules: []
    bypass_groups: []
    bypass_users: []
```

Workspace keys must match wmill.yaml's `workspaces` block, which stays the single source of truth for backend id / baseUrl / token (resolved per key via the existing `tryResolveBranchWorkspace` + stored profile, so `--all` works across workspaces/remotes).

**Why this is better:** it removes the `getEffectiveSettings` top-level/overrides/promotionOverrides layering entirely — which is the root cause of every CI-review finding so far (override shadowing, `--promotion` incoherence, `--diff` writing the file). There is simply no layering left to be incoherent.

**Command surface simplified:**
- `wmill protection-rules pull [workspace] | --all [--dry-run] [--json-output]`
- `wmill protection-rules push [workspace] | --all [--dry-run] [--json-output] [--yes]`
- Dropped: `--replace`, `--override`, `--default`, `--promotion`, `--diff` (now `--dry-run`).

`push` keeps the full-reconcile semantics per workspace (create/update/delete), delete-confirmation prompt, empty-list "wipes all" warning, partial-failure progress reporting, and non-zero exit on failure. `conf.ts` is reverted to `main` (no `SyncOptions` change).

**Verified e2e against the dev backend:** `pull <ws>` creates the file; `pull --all` populates every wmill.yaml workspace; `push <ws> --dry-run` shows the +/~/- plan and writes nothing / calls no mutating API; `push <ws> --yes` reconciles (created/updated/deleted reflected in DB); `push <key-not-in-file>` exits 1; a file key absent from wmill.yaml is a clean error (skipped under `--all`, non-zero exit; fails fast for a named arg). 57 unit tests pass (rewritten for the new model: converter, `WorkspaceResolver`, file round-trip), tsc clean, CLI docs regenerated.

rubenfiszel:
Addressed the Codex P1 and the 3 P2s identified by cubic, pushed in `17d3876` + `8f234d7`:

**Codex P1 — explicit `--base-url`/`--token`/`--workspace` were ignored** (`17d3876`). `configureClientForWorkspace` now honors the same credential precedence as every other command: explicit `--base-url`+`--token` are used as-is (stateless CI — no stored profile or wmill.yaml `baseUrl` needed), and an explicit `--token` overrides a stored profile's token. Backend workspace id still derives from the wmill.yaml mapping (feature invariant). E2E-verified: stateless pull/push with flags only succeeds; `--token` overrides a deliberately corrupted stored token (omitting it fails non-zero); profile mode unchanged.

**3 P2s identified by cubic** (`8f234d7`):
1. *(cubic, `pull.ts`)* partial `--all` failure reported `success: true` while exiting 1 — now reports `success: false` + `partialFailure: true` and exits 1; `success: true` only on full success.
2. *(cubic, `push.ts`)* same contradiction in push (incl. `--json-output`) — fixed identically across the dry-run, no-changes, and post-reconcile paths.
3. *(cubic, generated docs)* the `push` subcommand description was missing from `system_prompts/auto-generated/cli/cli-commands.md` because `generate.py` doesn't parse `+` string concatenation in `.description()` — switched to single string literals and regenerated; the `push` description now renders.

E2E-verified the partial-failure case: `push --all --json-output` with an unknown key now emits `success=false partialFailure=true` and exits 1 (consistent). tsc clean, 57 unit tests pass, CLI docs regenerated.

rubenfiszel:
Addressed the new Codex P1 in `a54d717`.

**Codex P1 — `--json-output` could emit non-JSON on stdout.** `tryResolveBranchWorkspace` (the stored-profile path) logs workspace/profile selection via `log.info`, and the empty-list delete warning uses `log.warn`; both write to **stdout**, so they were interleaved with the JSON payload for machine callers. Fix: call `log.setSilent(true)` as the very first action in `pull`/`push` when `--json-output` is set — before `readConfigFile` and workspace resolution (which also log). This suppresses `log.info`/`log.warn` (stdout); `log.error` is unaffected (it goes to stderr, so it can't corrupt the JSON).

E2E-verified with a normal saved profile (the path that logs "Using workspace profile …"):
- `pull <ws> --dry-run --json-output` → stdout is exactly **1 line**, strict `JSON.parse` OK.
- `push <ws> --yes --json-output` with an empty list (wipe) → stdout is exactly 1 valid JSON object, `success`/`deleted` correct, and the delete `WARNING` is **not** on stdout.

tsc clean, 57 unit tests pass.

rubenfiszel:
The latest **Pi review** flagged the `--json-output` log-suppression P1, but it reviewed a stale commit — Pi's own note says *"commit `a54d717` does not exist in the current branch (head is `8f234d7b`)"*. That's a CI-checkout race: Pi checked out `8f234d7b` before `a54d717adc` propagated, then submitted at 17:23:14Z.

The fix Pi recommends is **already on the branch**. Branch HEAD is `a54d717adc`:

```
$ git grep -n setSilent -- cli/src/commands/protection-rules/{pull,push}.ts
cli/src/commands/protection-rules/pull.ts:33:  if (opts.jsonOutput) log.setSilent(true);
cli/src/commands/protection-rules/push.ts:42:  if (opts.jsonOutput) log.setSilent(true);
```

That is exactly Pi's suggested fix (top of both handlers, before `readConfigFile`/resolution), shipped in `a54d717` ("fix: address review — --json-output must emit only JSON on stdout") and e2e-verified (stored-profile path: stdout is a single valid JSON payload; delete warning no longer on stdout). **Codex re-reviewed `a54d717` at 17:19:56Z → "Good to merge, no issues."** No action needed; a Pi re-run on the current HEAD should clear.

PR #9313, opened 2026-05-25
fix(auth): require workspace membership for g/ owner tokens (WIN-1986)

rubenfiszel:
Fixes WIN-1986

## Summary
- WIN-1978's `g/<groupname>` owner check only verified group existence, but the default `all` group is provisioned in every workspace — so a forged `owner=g/all` token paired with any `workspace_id` still authenticated as a group user there, regardless of workspace membership.
- The `g/` branch now additionally requires the token holder's email to be a non-disabled member of the target workspace, mirroring the `u/` branch's `usr` lookup.
- Adds an integration regression in `permissions.rs` that asserts a forged non-member `g/all` token is rejected and a legitimate `g/all` token for a real workspace member still authenticates.

## Vulnerability recap
With write access to the `token` table (e.g. via a metadata-DB compromise), an attacker could set `owner='g/all'` and `workspace_id='target-ws'` on a token they control and bypass the workspace boundary because `group_ WHERE name='all'` exists in every workspace by default. After this change, the additional `usr WHERE email=$3 AND workspace_id=$1 AND disabled=false` check rejects the forged token.

## Test plan
- [ ] `cargo test -p windmill-api-integration-tests --features deno_core test_forged_g_all_owner_rejects_non_member -- --ignored`
- [ ] Manual: insert a `token` row with `owner='g/all'`, `workspace_id='<target>'`, and an `email` not present in `<target>`'s `usr` table; verify `GET /api/w/<target>/users/whoami` returns 401.

---
## Summary by cubic
Enforces workspace membership for `g/<group>` tokens to close a cross-workspace auth gap (WIN-1986). Forged `owner=g/all` tokens now fail unless the token email is a member of the target workspace.

- **Bug Fixes**
  - Require both group existence and a non-disabled `usr` matching the token email in the target workspace before accepting `g/<group>` owner tokens.
  - Blocks forged `owner='g/all'` with arbitrary `workspace_id`; adds an integration test rejecting non-members and accepting real members.

- **Refactors**
  - Shortened the `g/` owner-check comment in `auth.rs` for clarity.

<sup>Written for commit d8b47f611f644d561463bef806fbc24e07e5c23f. Summary will update on new commits. <a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9313?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/.sqlx/query-ce9492ecccdcf4747990a94208dc54efefdeb8261142cc75010a51f209c099a2.json
+++ b/backend/.sqlx/query-ce9492ecccdcf4747990a94208dc54efefdeb8261142cc75010a51f209c099a2.json
@@ -1,0 +1,24 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT EXISTS(\n                                                    SELECT 1 FROM group_\n                                                    WHERE workspace_id = $1 AND name = $2\n                                                ) AND EXISTS(\n                                                    SELECT 1 FROM usr\n                                                    WHERE workspace_id = $1\n                                                      AND email = $3\n                                                      AND disabled = false\n                                                )",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "?column?",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Closing — on reflection the marginal security value is low. The threat model is 'attacker has token-table write access,' and with that primitive they can trivially satisfy the added `usr WHERE email = $3` check by picking any known workspace-member email, or short-circuit the check entirely by setting `super_admin=true` on the token, or skip tokens altogether and insert a `password` row with `super_admin=true`. The root cause is metadata-DB exposure (Bug Bounty #10 context); tightening individual EXISTS clauses in the auth fallback gives false reassurance against that class of attack.

PR #9332, opened 2026-05-26
feat: Add section to deploy projects to hub

tristantr:
## Summary

Adds a **Deploy projects to Hub** flow: a folder can be bundled into a Hub project (its scripts, flows, apps, resources, triggers and data-table migrations), published as a draft, submitted for review, and later imported/forked into another workspace under a chosen folder. Public apps can additionally be shared as live iframes.

## Screenshots

**Publish flow (predeploy) — select the folder's items to bundle**

Deploy to Hub — predeploy

**Bundle drawer — publication is blocked when a dependency can't be resolved**

Bundle drawer — unresolved-reference blocker

> The full publish → Hub → import round-trip requires a live Hub backend, which returns 401 in local CE dev, so the screenshots cover the workspace-side UI that renders locally. The unresolved-reference blocker, containment, and share-state logic are additionally covered by unit tests.

## Hardening (review rounds)

Portability and safety fixes surfaced during review:
- **Variable portability**: `$var:`/`$jsonvar:` references are relocated into the project slug on publish (items **and** trigger config) and retargeted into the destination folder on import, with empty secret placeholders materialized like resource stubs. Whole-string values only — inert literals in inline code are never rewritten.
- **Import containment**: imported items may only reference assets inside the target folder (plus `hub/` script/flow refs); everything else is rejected per-item.
- **Publish integrity**: publication is blocked while any bundle dependency is unresolved; a partial publish stays in a retryable state rather than becoming submit-ready.
- **Iframe sharing**: the Hub target is validated before flipping an app public, with policy rollback on embed/URL-resolution failure, and Unpublish stays available even when the public URL didn't resolve.
- **Draft rehydration**: reopened drafts preserve each app's live public-share state.

Custom resource types are intentionally standardized to the official Hub resource types, so no custom schema import is performed.


tristantr:
@claude review my pr

tristantr:
@claude review this pr

tristantr:
@claude review this PR, with my new fixes

tristantr:
@claude review this pr

hugocasa:
## Review notes (windmill side)

Overall the logic is sound and matches the hub contract in windmill-labs/windmillhub#140 (source_id linkage, `path`/`source_path` round-trip, trigger `runnable_path`/`runnable_kind`). A few things worth addressing before merge, grouped by severity.

### Functional gaps

1. **Apps that reference workspace scripts/flows by path are neither bundled nor path-rewritten.** `extractAppRefs` / `rewriteAppValue` in `projectBundle.ts` only handle `$res:` / `res://` tokens. A low-code app component that calls a workspace runnable by path (e.g. `f/other/myscript`) won't pull that script into the transitive closure, and its path won't be relocated under `f/<slug>/`. After import the app points at a path that doesn't exist in the destination workspace. Either extract+rewrite those runnable-by-path refs too, or at minimum surface a warning when an app references something outside the project folder.

2. **`install/+page.svelte` `retarget()` is a blunt whole-JSON string replace** (`JSON.stringify(bundle).split('f/<fromSlug>/').join('f/<folder>/')`). It rewrites every occurrence of that substring, including any incidental appearance inside script content. The publish side already does structural rewriting; the importer relies on a cruder mechanism that can over-replace.

### Install MVP limitations (document or guard)

3. Only `schedule` triggers are recreated on import; every other kind is recorded as `trigger kind '<kind>' not supported yet`. Fine as a first cut, but worth stating explicitly so it isn't mistaken for full trigger support.

4. Resources are created with `updateIfExists: false`, so any pre-existing resource path makes that item fail on (re-)import. Re-importing a project into a workspace that already has one of its resource paths will report per-item failures.

### Scope creep / unrelated diff noise (drop from this PR)

5. `backend/windmill-api/build.rs` is a pure rustfmt reflow of an unrelated line; no behavior change.

6. `frontend/src/routes/(root)/(logged)/workspace_settings/+page.svelte` relabels the "Workspace default app" tab to "Apps" with rate-limit copy, which is unrelated to deploying projects to the hub.

### Minor

7. `hub_publish.rs` header comment states folder names are "alphanumeric + underscore", but `validate_folder` also accepts `-`. Trivial doc drift.

8. `DeployToHub.svelte` at ~2.3k lines carries a lot (items, triggers, resources, recordings, app publishing, phase machine). Not a correctness issue, but the recording-builder and app-embed concerns are natural candidates to split into sub-components if it grows further.

rubenfiszel:
Addressed the latest Codex review:

**[P1] Variable dependencies are not portable** — Fixed in `632d809a77`. `$var:`/`$jsonvar:` refs are now retargeted into the destination folder (`buildRetargetMap` + `rewriteContent`) and materialized as empty secret placeholders on import (`existsVariable`-guarded), mirroring the resource-stub pattern. Values are never shipped; external refs stay rejected by containment. Renamed-folder imports no longer reject, and same-folder imports no longer dangle.

**[P1] Custom resource types are dropped on install** — Intentional: resource types should be standardized to the official Hub resource types, so there's no custom schema to recreate on import. Not treating this as a blocker.

Unit tests added for `$var:`/`$jsonvar:` retargeting and `collectExportVarPaths`. The full Hub export→import round-trip isn't exercisable in CE dev (Hub proxy 401).

PR #9336, opened 2026-05-26
feat: surface fork draft raw apps in compare page with merge warning

Guilhem-lm:
## Summary

In a forked workspace, the **Merge workspaces** compare/deploy page lists items that differ from the parent so they can be deployed. Scripts saved as drafts show up there and are deployable, but **raw apps do not**: `tally_deployed_object_changes` (`backend/windmill-git-sync/src/git_sync_ee.rs`) deliberately skips `DeployedObject::RawApp`, so a raw app's draft changes never land in the `workspace_diff` table that drives the page — users had no indication that a fork-local draft raw app even existed.

This PR surfaces the fork's draft raw apps on the compare page as **read-only rows**. Because they have no deployable version that can be merged from here, the selection checkbox is replaced by a warning icon with a popover explaining they must be deployed within the fork first. This is a frontend-only surfacing change — it does **not** alter the backend tally/deploy behavior.

## Changes

- `backend/windmill-api/openapi.yaml`: add `has_draft` and `draft_only` to the `ListableApp` schema (the backend already returns these; the spec was incomplete). The generated TS client is gitignored and regenerated from the spec on build.
- `frontend/src/lib/components/common/table/Row.svelte`: add an optional `leading` snippet that renders in place of the selection checkbox.
- `frontend/src/lib/components/WorkspaceDeployLayout.svelte`: add `itemLeading` snippet + `itemLeadingPredicate` props, threaded per-item to `Row`.
- `frontend/src/lib/components/CompareWorkspaces.svelte`: fetch the fork's raw-app drafts (`listApps({ includeDraftOnly: true })`, filtered to `raw_app && (draft_only || has_draft)`), inject them as non-selectable synthetic rows (only when merging fork→parent), and render an `AlertTriangle` in a hover `Popover` as the warning. The synthetic rows are excluded from every selection/deploy path by construction (`selectableDiffs`, `selectAll`, `deployChanges`, etc. all derive from `comparison.diffs` only).

## Test plan

- [ ] On a fork with a draft raw app, open `/forks/compare?workspace_id=<fork>` in "Deploy to <parent>" mode → the draft raw app appears with a warning icon (no checkbox) and a popover: *"This raw app has draft changes in this fork that haven't been deployed. Raw apps must be deployed within the fork before they can be merged into <parent>."*
- [ ] Draft scripts and deployed raw apps in the same fork still show a checkbox and remain deployable (no regression).
- [ ] Switching to "Update current" hides the fork-only draft raw apps.
- [ ] The draft raw app is never selectable and not counted toward "deployable"; "Select all" and Deploy are unaffected.
- [ ] `cd frontend && npm run check` passes.

## Notes / follow-up

Raw apps are skipped by `tally_deployed_object_changes` entirely, so even *deployed* (non-draft) raw apps aren't tallied into `workspace_diff` and don't appear/merge organically. Fully supporting raw-app fork→parent merge would require a backend change (include `RawApp` in the tally + a `compare_two_apps` path for raw apps). This PR intentionally stays frontend-only and just surfaces drafts + warns.

### DIFF
--- a/backend/windmill-api/openapi.yaml
+++ b/backend/windmill-api/openapi.yaml
@@ -26837,6 +26837,10 @@ components:
           enum: [viewer, publisher, anonymous]
         raw_app:
           type: boolean
+        has_draft:
+          type: boolean
+        draft_only:
+          type: boolean
         labels:
           type: array
           items:
--- a/frontend/src/lib/components/CompareWorkspaces.svelte
+++ b/frontend/src/lib/components/CompareWorkspaces.svelte
@@ -57,6 +57,7 @@
 	import ToggleButtonGroup from './common/toggleButton-v2/ToggleButtonGroup.svelte'
 	import ToggleButton from './common/toggleButton-v2/ToggleButton.svelte'
 	import DatatableSchemaDiff from './DatatableSchemaDiff.svelte'
+	import Popover from './meltComponents/Popover.svelte'
 
 	interface Props {
 		currentWorkspaceId: string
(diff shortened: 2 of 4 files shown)


Guilhem-lm:
Not the right fix, we will show a specific section for drafts

PR #9349, opened 2026-05-27
feat: deploy raw apps from global chat

centdix:
## Summary
Enables global AI chat to deploy raw app drafts by bundling the app through the UI-builder iframe before calling raw app create/update endpoints.

Companion UI-builder PR: https://github.com/windmill-labs/windmill-code-ui-builder/pull/10
Related UI-builder cleanup PR, not a dependency: https://github.com/windmill-labs/windmill-code-ui-builder/pull/11

## Release ordering
The companion UI-builder PR has been merged and this PR now pins Windmill to the published `fe13d03` UI-builder artifact. The bridge requires the new `bundleRawAppReady` / `bundleRawApp` / `bundleRawAppResult` protocol and intentionally does not fall back to the old full-editor `setFiles` / `preview` protocol.

## Changes
- Add a hidden credentialless UI-builder iframe bridge for raw app bundling.
- Wire `deploy_workspace_item` for app drafts to bundle files and call `createAppRaw`/`updateAppRaw`.
- Default missing raw app policy execution mode to `publisher` for chat-created app drafts.
- Require the new bundle-only UI-builder protocol for AI-chat raw app deploys.
- Remove the compatibility fallback that drove the full raw app editor iframe via `setFiles`, `preview`, and `getBundle` messages.
- Bump the bundled UI-builder artifact to `fe13d03`, which contains the new bundle-only mode.
- Cover raw app create/update deploy paths in global chat tests.

## Test plan
- [x] `npm run check:fast`
- [x] `npm run check` (0 errors; existing warnings only)
- [x] `npm run test:unit -- --run src/lib/components/copilot/chat/global/core.test.ts`
- [x] `node frontend/scripts/untar_ui_builder.js` verified the `fe13d03` artifact checksum and extraction
- [x] Browser smoke through `http://localhost:3020` with the pinned `fe13d03` UI-builder artifact: imported the actual `bundleRawAppFiles()` bridge and received a JS/CSS bundle from the hidden iframe
- [x] Browser smoke against the old remote UI-builder artifact before the bump: bridge timed out as expected, confirming there is no legacy fallback path
- [x] Earlier end-to-end browser smoke: `init_app` + `deploy_workspace_item` created a raw app through the hidden iframe bundler and backend returned 201


centdix:
Thanks for the review. I agree the `custom_path` asymmetry is the only point worth documenting right now.

I checked the backend behavior: `updateAppRaw` only changes `custom_path` when the field is present. If it is omitted, the backend preserves the existing value. Sending the field is not always safe though, because the backend requires admin privileges for `custom_path` updates, and the global chat deploy path does not currently mirror the raw app editor's user/admin-gated handling.

Decision for this PR: keep omitting `custom_path` on update so non-admin redeploys do not accidentally fail and existing custom paths are preserved by the backend. I added an inline comment in `core.ts` to make that tradeoff explicit. A fuller follow-up would be to plumb user/admin state into this path and mirror `RawAppEditorHeader.svelte` exactly.

On `execution_mode = 'publisher'`: I think the current behavior is acceptable. The draft policy is recomputed after app edits, and this matches the editor's save-time defaulting behavior, so I would not move it to deploy-only for this PR.

On `bundleRawAppFiles`: agreed it is a minor surface-area nit. It is internal to this frontend module and useful for smoke testing, so I am leaving it as-is unless we decide to tighten that later.

PR #9355, opened 2026-05-27
fix: batch encryption-key rotation into one git-sync job

### LINKED ISSUE #9344: bug: Workspace re-encryption does not trigger Git sync
### Describe the bug

When re-encrypting a workspace with a new key, git sync is not triggered. When secrets sync is enabled all secrets should change due to new encryption.

### To reproduce

1. Configure Git Sync with Secrets Sync enabled
2. Change Workspace Encryption Key
3. No Git Sync is triggered

### Expected behavior

After the encryption key change a git sync should be triggered.

### Screenshots

_No response_

### Browser information

_No response_

### Application version

EE v1.709.0

### Additional Context

_No response_

rubenfiszel:
## Summary

Workspace encryption key rotation re-encrypts every secret variable
inside one transaction, but the post-commit git-sync dispatch only fired
a `DeployedObject::Key` event for `encryption_key.yaml`. Repos with
Secrets sync enabled were left with the *old* ciphertexts on disk until
each variable was individually deployed again.

This PR:
- Adds `handle_deployment_metadata_batch` to `windmill-git-sync` (EE
  implementation in the companion PR, OSS no-op shim). It looks up the
  repo settings once, runs each `DeployedObject` through the existing
  path/type filters, and pushes **one** git-sync job per
  `(repo, debounce_key)` bucket carrying every item in the `items`
  array.
  - Workspace-wide mode: one job for the whole batch.
  - Promotion + `group_by_folder`: one job per folder.
  - Promotion + per-item: one job per item (same as today, but no
    reliance on the debouncer's 5–15s merge window).
- Sets `skip_secret` once per repo whenever the batch contains at least
  one Variable, so the hub script knows whether to write decrypted
  values for any of the included secrets.
- Switches `set_encryption_key` to build a single batch
  `[Key, Variable(p1), Variable(p2), …]` and dispatch it through the
  new entry point.

The single-object `handle_deployment_metadata` path is untouched.

## Companion EE PR

`windmill-ee-private` branch: `ruben/win-1994-fix-httpsgithubcomwindmill-labswindmillissues9344`
HEAD: `108b0e2257d2df18b043802f88e945f9d3f0edf1`

## Test plan

- [x] `cargo check --features enterprise,private` — clean (3 unrelated pre-existing warnings only)
- [x] New integration test: `cargo test -p windmill-api-integration-tests --test workspace_encryption_key_git_sync --features enterprise,private`
  - Configures Variable + Secret + Key in `include_type` and inserts 3 secret variables
  - POSTs to `/workspaces/encryption_key` with a new key
  - Asserts exactly **one** deployment-callback job is queued whose `items` array contains the `key` entry + all 3 `variable` entries with `skip_secret=false`
- [x] Existing tests still green: `workspaces`, `workspace_dependencies_git_sync`

Fixes WIN-1994
Fixes https://github.com/windmill-labs/windmill/issues/9344


rubenfiszel:
Addressed the review findings:

**P1 (blocking) — batch path skipped repos without debouncing support → key rotation queued zero jobs on older sync scripts.** Fixed in the EE `handle_deployment_metadata_batch_inner` (companion PR, now at `760dd3c`). When `supports_debouncing` is false it no longer `continue`s past the repo; instead it falls back to the legacy one-job-per-item format (flat `path_type`/`path`/`parent_path`/`commit_msg` args, default debouncing settings) — identical to what the single-object path's non-debounced `else` branch produces. So a repo on a sync script older than hub `28103` (or workers below `1.602.0`) again receives a git-sync job for `encryption_key.yaml` plus one per re-encrypted secret.

Added a regression test `test_encryption_key_rotation_falls_back_without_debouncing` that anchors the sync script in `f/28000/` (below 28103), rotates the key, and asserts the rotation queues `key + N variable` legacy-format jobs (asserting `items` is absent and each job carries flat `path_type`/`path`). Both this and the happy-path test pass, as do the existing `workspace_dependencies_git_sync` tests.

Also extracted a shared `push_git_sync_callback` helper so the debounced and legacy branches don't duplicate the long `windmill_queue::push` signature.

**P2 — post-commit `handle_deployment_metadata_batch(...).await?` propagating as a request failure.** This is a no-op in practice: like the single-object `handle_deployment_metadata` used at every other call site, the batch entrypoint clones its inputs, `tokio::spawn`s the real work (which logs its own errors), and returns `Ok(())` synchronously — there's no inline fallible work before the spawn, so `?` can't surface a post-commit error. Kept `.await?` for consistency with the surrounding call sites rather than adding an `if let Err` guard for a branch that can't be taken (per AGENTS.md "don't add error handling for scenarios that can't happen").

rubenfiszel:
### `git_sync_e2e` failure is pre-existing, not caused by this PR

The failing test is `test_workspace_fork_creates_branch` ("Expected a wm-fork branch in the repo after forking, got: ['origin/HEAD', 'origin/main']"). I reproduced it locally and bisected it:

- **On this branch** (local EE build + Gitea + license): the fork test fails.
- **On `origin/main`'s git-sync code** (reverted my `git_sync_ee.rs` / `workspaces.rs` / `git_sync_oss.rs` / `lib.rs` to `origin/main`, rebuilt, re-ran the exact same test): it **fails identically**.

So the failure is independent of this PR. It's normally invisible because `git-sync-test.yml` is path-gated and the `git_sync_e2e` job is **skipped on essentially every main push** (main rarely touches `backend/windmill-git-sync/**`); this PR changes git-sync files, so the suite actually runs and surfaces the pre-existing breakage. Only this one test fails — every other git-sync e2e test passes.

Root cause (for a separate fix/issue): the fork-branch sync job is queued with correct args (`only_create_branch: true`, `workspace_id: wm-fork-…`, `repo_url_resource_path`, the "Creating branch" item) and the job **completes successfully**, but the hub sync script (`hub/28236/sync-script-to-git-repo-windmill`) creates no `wm-fork/...` branch in the repo. That's in the fork-branch-creation path (`handle_fork_branch_creation` → `obj=None`), which this PR does not touch — the args it produces are byte-identical to main.

Recommend tracking the fork-branch e2e failure separately; it should not block the encryption-key git-sync change here.

PR #9378, opened 2026-05-29
feat: support $f/ and $u/ import path aliases for scripts

hugocasa:
## Summary

When a Windmill script imports another workspace script via the absolute path `/f/foo/bar.ts` (or `/u/...`), it resolves on the worker but **breaks locally**: the leading `/` makes every local tool (tsc, Deno, Bun, node) treat it as a filesystem-root path, while the CLI lays scripts out under `<project>/f/...` relative to the project root.

This PR adds `$f/...` / `$u/...` as local-friendly aliases. The `$`-prefixed form is a *bare specifier* that tsconfig `paths` and Deno import maps can remap to the on-disk `f/`/`u/` folders, so the **same import resolves both on the Windmill worker and in a local editor**. It's additive — `/f/` and `/u/` keep working — and the new form keeps the existing `.ts` extension convention.

## Changes

**Worker — recognize the aliases at runtime**
- `deno_executor.rs`: add `$f/` / `$u/` entries to the generated Deno import map.
- `loader.bun.js` + `loader.bun.windows.js`: normalize `$f/`/`$u/` to the absolute `/f/`,`/u/` resolution branch (and `.ts` auto-append). `loader_builder.bun.js` reuses the same plugin.

**Dependency tracking / lockgen**
- `windmill-dep-map/src/lib.rs` and `windmill-parser-ts/src/lib.rs` (`parse_relative_imports`, also used by the WASM/CLI): strip `$f/`→`f/`, `$u/`→`u/` so the dependency graph, staleness propagation and lockfiles see `$`-imports.

**CLI — local resolution (two-file managed config, mirroring the AI-prompts model)**
- New `wmill refresh tsconfig` command, also run unconditionally by `wmill init` (no longer gated behind resource-type namespace / a bound workspace).
- Generates a wmill-**managed** `tsconfig.wmill.json` (always refreshed; holds the recommended `compilerOptions` incl. `$f/$u` `paths` + `allowImportingTsExtensions`) and a **user-owned** `tsconfig.json` that just `extends` it (created once, never overwritten). Existing non-extending tsconfigs get a warning, never an auto-edit. This lets us evolve the recommended config over time without clobbering user customizations — exactly how `AGENTS.cli.md` (managed) / `AGENTS.md` (user-owned) work. Deno-default projects get the equivalent `import_map.wmill.json` + `deno.json` `importMap`.

**Frontend — editor IntelliSense**
- `relative_imports.ts`, `ata/index.ts`, `monacoLanguagesOptions.ts`: ATA fetches/registers `$f/` imports under their `/f/...` path and Monaco's `paths` maps the alias back, so the in-app editor resolves `$f/`/`$u/` type hints.

## Validation

- `cargo` workspace compiles; `windmill-parser-ts` unit tests pass (added `$f/`/`$u/` cases).
- Added `$f/`-alias variants of the Bun & Deno relative-import worker tests.
- Frontend `npm run check:fast` passes; CLI `bunx tsc --noEmit` clean on changed files.
- **End-to-end (live)**: Bun+`$f/`, Deno+`$f/` (import map), Bun+`$u/`, legacy `/f/` (back-compat) all run; deploying a `$f/` importer records the `dependency_map` edge.
- **ATA (live)**: opening a script importing `$f/...` fetches the helper (200) and the Monaco editor shows 0 error squiggles.
- **CLI (live)**: `wmill init` generates `tsconfig.wmill.json` + `tsconfig.json` (extends); `wmill refresh tsconfig` works standalone and is idempotent; `tsc` resolves a `$f/` import through the extends chain on a real pulled project; non-extending tsconfig triggers the warning. Verified the rebuilt WASM parser via `bun link`.

## Test plan

- [ ] Web editor: a Bun/Deno script importing `$f/folder/script.ts` resolves type hints (no "cannot find module") and runs via Test.
- [ ] `$u/user/script.ts` imports resolve and run.
- [ ] Existing `/f/` and `/u/` imports still work (back-compat).
- [ ] `wmill init` creates `tsconfig.wmill.json` (managed) + `tsconfig.json` (extends); `$f/` imports resolve in VS Code against pulled scripts.
- [ ] `wmill refresh tsconfig` refreshes the managed file and auto-wires an existing unlinked `tsconfig.json` (adds `extends`); a commented/JSONC tsconfig is left untouched with a warning instead.
- [ ] Deploying / `wmill sync push` a `$f/` importer tracks the dependency and generates a lockfile without 404ing the alias.

## Consistency with the prompts setup

The managed/user tsconfig split mirrors the `AGENTS.cli.md` (managed) / `AGENTS.md` (user-owned) prompts model, and now behaves **identically** to it on the two points that previously diverged:

1. **Existing unlinked file → auto-wired.** Just like `refresh prompts` appends `@AGENTS.cli.md` to an existing `AGENTS.md`, `refresh tsconfig` now adds `extends` to an existing `tsconfig.json` (and `importMap` to `deno.json`), merging into an array `extends` when one is already present. It parses the JSON and **falls back to a warning (never corrupts)** when it can't round-trip — i.e. a commented/JSONC config, or a `deno.json` that already defines `imports`.
2. **Always-on warning is stale-only.** `warnIfTsconfigStale` now mirrors `warnIfPromptsStale` exactly: gated on the *managed* file existing (= opted in) and only ever warns that it's **stale** — never about a missing or unlinked user config. So a deliberately custom/unlinked setup is never nagged, and a not-yet-initialized project stays silent.

## Follow-ups

- The CLI consumes the published `windmill-parser-wasm-ts` npm package — the dep-tracking change lands for users once that package is rebuilt/republished and `cli/package.json` bumped (verified locally via `bun link`). The generated tsconfig/deno config work independently today.
- `sync pull` doesn't (re)generate the tsconfig; a pure-sync clone that never runs `init`/`refresh tsconfig` won't get the aliases. Could optionally refresh during pull if desired.
- User-facing docs (separate docs repo).


hugocasa:
Thanks for the reviews @ — two intentional points to record:

**[P1] `windmill-parser-wasm-ts` pin (`1.695.0`)** — acknowledged and the explicit merge gate (hence the `[do not merge, publish parser first]` title). The Rust parser source, dep-map, Bun/Deno loaders, Deno import map, frontend ATA + Monaco are all updated for `$f/`/`$u/`; only the *published* CLI WASM package lags. Publishing the parser and bumping `cli/package.json` is the unblock step before merge.

**[P1/P2] `bun add -d bun-types` on `wmill init`** — intentional, leaving the pre-existing behavior in place. `bun-types` is only referenced from the generated managed tsconfig when it's actually installed (it's excluded from the freshness hash), and the install is skipped entirely when Bun isn't present (instructions are logged instead). The broader call site is deliberate so a fresh `wmill init` produces a working IDE setup.

The other two findings (plain-`.ts`-only local resolution; tsconfig generated for non-TS projects) are already documented as known/intentional in code and the PR body. The `extends` ordering P2 was fixed in ab8f7dd (managed config now first, preserving the user's base).

hugocasa:
/docs replace the absolute path docs with this new recommended way

hugocasa:
Thanks for the reviews.

**@pi** — confirmed good to merge. Ran the manual checks live: Bun + Deno + `$u/` scripts deploy and run correctly via the aliases, and the Deno lock now records the resolved dependency (the lock-gen import map was missing `$f/`/`$u/` — fixed).

**@codex** — good catch on the P2 (custom `paths` shadowing the aliases). Addressed: rather than silently auto-wiring, `wmill refresh tsconfig` / `wmill init` now detects when an existing custom `tsconfig.json` defines its own `compilerOptions.paths` — which TS does *not* merge across `extends` — and **warns** the user to add `"$f/*": ["./f/*"]` / `"$u/*": ["./u/*"]` manually, instead of adding an `extends` that would be shadowed. Configs without their own `paths` auto-wire as before.

For transparency, the remaining notes (bun-types auto-install on `init`, `.ts`-only local resolution, tsconfig generated for non-TS projects) are intentional and were confirmed earlier in the thread.

PR #9382, opened 2026-05-29
docs(skills): fix //native marker + import rules for bunnative, remove legacy nativets skill

s-stefanov:
## Summary

Native TypeScript on Windmill has two script "languages": `nativets` (the legacy standalone Native TypeScript runtime) and `bunnative`. `bunnative` is simply **a Bun script with `//native` as its first line** — at deploy time, a `Bun`/`Bunnative` script is classified as `Bunnative` iff it carries the `//native` annotation (`windmill-api-scripts/src/scripts.rs`), and both route to the same `nativets` worker tag at runtime. The frontend already hides `nativets` from the new-script language picker (`DefaultScriptsInner.svelte`), so for new scripts `nativets` is dead weight that only invites drift.

This PR does two things:

1. **Fixes the `bunnative` skill docs** — adds the mandatory `//native` first-line marker to every code example (without it scripts silently run on the regular Bun worker) and corrects the import rule.

2. **Removes the legacy `nativets` agent skill entirely** so agents always author native TypeScript as `bunnative`.

### The import rule (corrected)

The earlier revision of this PR claimed "the only library import allowed is `windmill-client`". **That is wrong** — it's the restriction of the *legacy* `nativets` runtime, not `bunnative`.

- `bunnative` is handled identically to `bun` for dependency resolution: `gen_bun_lockfile` installs npm deps and `prebundle_bun_script` runs `bun build --target browser` to bundle them (`worker_lockfiles.rs:2789`, `bun_executor.rs:1199`). So **any npm package can be imported** — imports are bundled.
- The legacy `nativets` path (`do_nativets`, `worker.rs:3400`) only runs `transpile_ts` with no bundling, so it genuinely can't use npm deps. This is *why* `bunnative` superseded it.
- The real constraint is the **runtime**, not the import list: the native worker is a lightweight V8 isolate exposing only `fetch` + the JS standard library, so imported code must rely only on those. Libraries needing Node/Bun built-ins (`node:fs`, `child_process`, the `Bun` API, native addons) belong in a regular `bun` script. This matches the product's own wording in `frontend/src/lib/script_helpers.ts` (`BUNNATIVE_INIT_CODE`): "imports will work as long as they also use only fetch and the standard lib".

## Changes

- **`system_prompts/languages/bunnative.md`**: added `//native` as the mandatory first line of every code block; rewrote the intro, `## Imports`, and `## Windmill Client` sections to state the real rule (any bundleable, browser-compatible library is allowed; constraint is the V8 runtime).
- **Removed `system_prompts/languages/nativets.md`** and dropped `nativets` from `LANGUAGE_METADATA` (`system_prompts/utils.py`) so no `write-script-nativets` skill is generated. `nativets` is kept in `TS_SDK_LANGUAGES` so the TypeScript SDK still attaches when editing pre-existing legacy `nativets` scripts.
- **Regenerated** `prompts.ts`, `script.md`, `write-script-bunnative/SKILL.md`, `cli/src/guidance/skills.gen.ts`, and deleted the stale `write-script-nativets/SKILL.md` via `python system_prompts/generate.py`.

Remaining `nativets` references in generated files are intentional backward-compat: the `TS_SDK_LANGUAGES` router array and the OpenFlow/OpenAPI `ScriptLang` schema enum.

## Test plan

- [x] `bash system_prompts/check-freshness.sh` exits 0 (generator output matches committed files)
- [x] No `write-script-nativets` skill is generated; only `write-script-bunnative` remains
- [x] Every code block in `write-script-bunnative/SKILL.md` starts with `//native`
- [x] Import rule reflects bundled-npm support + the V8-runtime constraint, not a false "only windmill-client" rule


s-stefanov:

s-stefanov:
recheck

rubenfiszel:
### Pivoted from "sync nativets" to "remove nativets"

The original version of this PR brought `nativets.md` into byte-for-byte sync with `bunnative.md` (including a brand-new `## S3 Object Operations` section). That entrenches the very duplicate the PR's own follow-up note flagged for deletion. After verifying the runtime semantics, I changed the direction to remove the legacy skill outright.

**Why `nativets` is legacy, with receipts:**

- `bunnative` is just **a Bun script with `//native` on line 1** — at deploy time a `Bun`/`Bunnative` script is classified as `Bunnative` iff it carries the `//native` annotation, else plain `Bun` (`backend/windmill-api-scripts/src/scripts.rs:1167-1176`).
- Both `Bunnative` and `Nativets` jobs route to the same `nativets` worker tag at runtime (`backend/windmill-queue/src/jobs.rs:5725`), so there's no behavioral reason to keep two agent skills.
- The frontend already filters `nativets` out of the new-script language picker (`frontend/src/lib/components/DefaultScriptsInner.svelte:20`).

**What changed vs. the first revision:**

- Kept the good `bunnative.md` fixes (mandatory `//native` marker on every example + corrected `windmill-client`-only import rule).
- Deleted `system_prompts/languages/nativets.md` and dropped `nativets` from `LANGUAGE_METADATA` so **no `write-script-nativets` skill is generated** — agents always author native TS as `bunnative`.
- Kept `nativets` in `TS_SDK_LANGUAGES` so the TypeScript SDK still attaches when **editing pre-existing** legacy `nativets` scripts (no backward-compat break).
- Regenerated all artifacts via `python system_prompts/generate.py` and removed the stale generated `write-script-nativets/SKILL.md`.

The only `nativets` strings remaining in generated files are intentional backward-compat: the `TS_SDK_LANGUAGES` router array and the OpenAPI `ScriptLang` schema enum. `bash system_prompts/check-freshness.sh` exits 0.

rubenfiszel:
### Correction: bunnative is **not** limited to `windmill-client`

I previously stated the import rule as "the only library allowed is `windmill-client`". That's wrong, and I've fixed it — that restriction belongs to the *legacy* `nativets` runtime, not `bunnative`. This is actually the whole reason `bunnative` exists.

**Receipts:**

- `bunnative` is handled identically to `bun` for dependency resolution — `gen_bun_lockfile` installs the npm deps, then `prebundle_bun_script` runs `bun build --target browser` to bundle them (`backend/windmill-worker/src/worker_lockfiles.rs:2789`, `backend/windmill-worker/src/bun_executor.rs:1199-1206`). So **any npm package can be imported**; imports are resolved and bundled.
- The legacy `nativets` path is just `do_nativets` → `transpile_ts(code)` with **no bundling** (`backend/windmill-worker/src/worker.rs:3400`), so it genuinely can't pull in npm deps — only `fetch` + the specially-handled `windmill-client`. That's precisely why `bunnative` superseded it.
- The real constraint is the **runtime, not the import list**: native code runs in a lightweight V8 isolate (browser-target bundle) exposing `fetch` + the JS standard library. Imported code must rely only on those — anything needing Node/Bun built-ins (`node:fs`, `child_process`, the `Bun` API, native addons) should use a regular `bun` script.

This matches the product's own wording in `frontend/src/lib/script_helpers.ts` (`BUNNATIVE_INIT_CODE`):

> native scripts are bun scripts that are executed on native workers and can be parallelized
> only fetch is allowed, but imports will work as long as they also use only fetch and the standard lib

`bunnative.md` and all regenerated artifacts now reflect this; `bash system_prompts/check-freshness.sh` exits 0.

PR #9393, opened 2026-06-01
perf(python): add --compile-bytecode to uv pip install (WIN-2001)

rubenfiszel:
## Problem

Python jobs running under nsjail experience slow imports (~465ms for `import requests`) because dependency directories are mounted read-only (`rw: false` in the nsjail config). Without pre-compiled `.pyc` files, Python must parse and compile `.py` source to bytecode in memory on **every import in every job**. Since each nsjail execution is a fresh process, this overhead is paid repeatedly. Diagnostics confirmed `pyc_present_now: false` and no `__pycache__` directories in the cached deps.

## Fix

Add `--compile-bytecode` to both `uv pip install` invocations so `.pyc` files are generated at install time and remain available at runtime even through read-only mounts:

- `backend/windmill-worker/src/python_executor.rs` — added to the Rust-driven `command_args` vec after `--reinstall`.
- `backend/windmill-worker/nsjail/download_deps.py.sh` — added to the `CMD` string after `--reinstall`.

The generated `.pyc` files are included in both the local cache and S3 piptar uploads.

## Notes

- The flag is supported by the uv version (0.9.24+) shipped in the Dockerfile.
- The existing `__pycache__` skip at `python_executor.rs:549` only filters top-level dirs in the target for dedup logic, not subdirs within packages — no conflict.

## Validation

- `cargo check -p windmill-worker --features python` passes (only pre-existing warnings unrelated to this change).
- No UI surface; backend-only change.

Fixes WIN-2001

---
## Summary by cubic
Adds `--compile-bytecode` to `uv pip install` so `.pyc` files are generated at install time, speeding up imports when deps are mounted read-only under nsjail. Applied in the Rust executor and `nsjail/download_deps.py.sh`; `.pyc` files are cached and uploaded, fixing WIN-2001.

<sup>Written for commit 2b56e088fe4ef6dcd5adf916fc63a6e2b6846c82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
## E2E benchmark (mechanism + perf claim verified)

Ran a real before/after under the exact read-only condition the nsjail mount imposes (`rw: false`). Environment: `uv 0.11.15`, `nsjail` present, `python 3.14.5`.

**1. The flag produces `.pyc` files**

```
install WITHOUT --compile-bytecode → __pycache__ dirs: 0, .pyc files: 0
install WITH    --compile-bytecode → "Bytecode compiled 19 files in 62ms"
                                      __pycache__ dirs: present, .pyc files: 19 (requests only)
                                      80 .pyc with full dep tree (urllib3/certifi/idna/charset_normalizer)
```

**2. Imports are faster under a read-only mount**

`requests` installed (with deps) into two targets, then each made read-only (so Python cannot write `.pyc` at runtime — the nsjail condition). Timed `import requests` with `time.perf_counter()` in a **fresh process**, no `PYTHONPYCACHEPREFIX` override, averaged over 25 runs:

| | avg `import requests` |
|---|---|
| without `--compile-bytecode` | **96.53 ms** |
| with `--compile-bytecode` | **49.38 ms** |

≈ **2× faster (~49% reduction)**.

Sanity check confirmed the no-compile target held **0** `.pyc` after all 25 runs — i.e. it genuinely recompiles source→bytecode in-memory on every fresh process, which is what this flag eliminates.

> Note: absolute numbers differ from the ~465ms figure in the issue (different hardware / Python version / package set), but the relative improvement and root-cause mechanism are confirmed. An early run showed only a ~3ms delta because `PYTHONPYCACHEPREFIX` was set, which redirects bytecode lookup away from the in-tree `__pycache__`; removing it gave the result above.

rubenfiszel:
## Where bytecode is stored / S3 / cache-warming (traced through the code)

**Where is it stored?**
In `<package>/__pycache__/<module>.cpython-XYZ.pyc`, nested *inside each package directory* of the `--target` dir — i.e. the existing per-dependency cache dir (e.g. `…/windmill/cache/python_311/requests==2.32.3/requests/__pycache__/…`). It is not a separate location; the `.pyc` ride inside the same per-dep folder Windmill already caches and tars. (Confirmed: install produced 80 `.pyc` across `requests/urllib3/certifi/idna/charset_normalizer` `__pycache__` dirs.)

**Does it work with the S3 piptar setup? — Yes.**
`build_tar_and_push` (global_cache.rs:42) tars the whole per-dep folder with `tar.append_dir_all(".", &folder)` — recursive, so the nested `__pycache__/*.pyc` are included. `pull_from_tar`→`extract_tar` does `ar.unpack(folder)`, restoring everything with mtimes preserved. Verified by simulating the exact round-trip (tar the dep dir → extract elsewhere → make read-only → time import): **49.92 ms**, matching the fresh-compiled **49.38 ms** and half the no-compile **96.53 ms**. Since Python validates a `.pyc` against its source `.py` timestamp, this also proves the round-trip keeps the `.pyc` *valid* (tar preserves mtimes on both files) — otherwise it would silently recompile back to ~96 ms.

Also note `verify_wheel_record` (python_executor.rs:2225) only checks that RECORD-listed files *exist* on disk; it never rejects extra files, so the added `.pyc` (not in the wheel RECORD) don't trip the integrity check.

**Preserved as soon as the worker runs one job? — Yes for that dep, with a caveat.**
`.pyc` are generated at *install* time, which only runs on a **cache miss** (`uv pip install --reinstall --compile-bytecode`). The install writes into the read-write per-dep cache dir on local disk (the nsjail `rw:false` constraint only applies at job-*execution* mount time — which is exactly why precompiling at install is the right place). That same dir, now containing `.pyc`, is queued for the S3 piptar upload. So the first job that installs a given dep persists the `.pyc` both locally and to S3; every later job on that worker, and other workers via S3, get them for free.

> **Caveat (gradual rollout):** dependencies already cached *before* this change — locally or in S3 — have no `.pyc`, and `--compile-bytecode` only takes effect when an install actually runs. On a plain cache hit nothing recompiles, so existing entries gain `.pyc` only when they are next missed/evicted and reinstalled. New and cache-missed deps benefit immediately; pre-warmed caches benefit gradually. There is no retroactive compilation of existing cache entries in this change.

PR #9395, opened 2026-06-01
feat: add workspace datatable tools to global AI chat mode

centdix:
## Summary

Adds datatable tooling to the **global mode** AI chat so the assistant can discover, inspect, query, and mutate workspace datatables directly, and authors correct `wmill.datatable()` runnable code. Previously global mode had no datatable capability — it could only scaffold an app draft that *declared* a datatable config, but couldn't actually interact with datatables.

Unlike the app-mode datatable tools, these are **unrestricted by design**: no app whitelist and no creation policy. This is intentional — datatables are workspace-scoped and the backend endpoints are gated only by workspace membership, so there's no enforcement boundary to mirror (the app-mode whitelist is purely an authoring-UX scoping construct, not a security boundary at any layer). `exec_datatable_sql` is gated with a confirmation prompt instead.

## Changes

- New `frontend/src/lib/components/copilot/chat/datatableTools.ts`:
  - Pure workspace-scoped ops (`listDatatables`, `getDatatableColumns`, `execDatatableSql`) wrapping the existing `WorkspaceService.listDataTableTables` / `getDataTableTableSchema` endpoints and `runScriptAndPollResult` (SQL via `datatable://<name>`).
  - `getDatatableTools()` exposing three helper-free `Tool<{}>`: `list_datatables`, `get_datatable_table_schema`, `exec_datatable_sql` (same names as app mode).
  - No whitelist, no creation policy. `exec_datatable_sql` drops the app-mode `new_table` registration param (not needed — created tables appear in `list_datatables` automatically) and is marked `requiresConfirmation: true` so every statement is confirmed before running.
- `frontend/src/lib/components/copilot/chat/global/core.ts`:
  - Import + spread the three tools into the base `globalTools` array (available in both session and side-panel global chats) + a policy-free "Data Tables" system-prompt section that instructs TypeScript-by-default for runnable code.
  - New `get_instructions(subject: "datatable", language?)` subject returning the authoritative `wmill.datatable()` SQL SDK reference. It is **language-aware**: passing `bun`/`deno`/etc. returns only the TypeScript SDK, `python3` only the Python SDK, and it defaults to TypeScript when omitted — so the model gets just what it needs instead of both. `datatable` is decoupled from `WorkspaceItemType` since it isn't a workspace item.
- `system_prompts`:
  - `getDatatableSdkReference(language?)` is now language-aware (TS-only / Python-only / both), mirroring `getWorkflowAsCodePrompt`. Edited the generator (`generate.py`) and regenerated `auto-generated/index.ts` + `index.d.ts`. The TS/Python SDK building blocks (`DATATABLE_SDK_TYPESCRIPT` / `DATATABLE_SDK_PYTHON`) already existed separately; this just exposes them selectively (previously an unused combined export).
- `init_app`'s optional `data` arg left untouched.
- Tests: `datatableTools.test.ts` (net-new pure logic — `totalTables` aggregation, `MAX_ROWS = 100` truncation boundary, result shaping, empty/error paths) and `global/core.test.ts` cases asserting `get_instructions(datatable)` defaults to TS-only and `language: "python3"` returns Python-only.

## Test plan

- [x] `vitest` — `datatableTools.test.ts` (9 tests) + `global/core.test.ts` datatable-instructions cases, all passing
- [x] Manual end-to-end in the browser (global chat, gemini-2.5-flash) against a real datatable: the model ran `list_datatables` → `exec_datatable_sql` (CREATE / INSERT / SELECT) → `get_datatable_table_schema`, with each mutating statement gated by the confirmation prompt; SELECT returned the inserted rows and `get_datatable_table_schema` resolved columns with no whitelist gate. DB-verified the writes, then reverted all test fixtures.
- [ ] `npm run check` (frontend type-check) in CI
- [ ] Reviewer sign-off on the intentional removal of the whitelist / creation-policy guardrails for global mode (mitigated by per-statement confirmation)

## Follow-up (not in this PR)

`getRawAppPrompt()`'s "## Data Tables" section (surfaced via `get_instructions(subject: "app")`) still tells the model *"whitelisted tables only / register in `data.tables` first / unlisted queries fail at runtime."* That framing predates global mode's no-whitelist approach and isn't actually runtime-enforced. It's a shared prompt (CLI/app contexts), so adjusting it is a separate, broader change.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/datatableTools.test.ts
+++ b/frontend/src/lib/components/copilot/chat/datatableTools.test.ts
@@ -83,10 +83,13 @@ describe('list_datatables', () => {
 		})
 	})
 
-	it('returns a friendly message when no datatables are configured', async () => {
+	it('explains that configuring a datatable is a blocking prerequisite when none exist', async () => {
 		listMock.mockResolvedValue([])
 		const result = await run('list_datatables')
-		expect(result).toContain('No datatables are configured in this workspace.')
+		expect(result).toContain('No datatables are configured in this workspace')
+		expect(result).toContain('Workspace settings → Data Tables')
+		expect(result).toContain('blocked')
+		expect(result).toContain('Do not call exec_datatable_sql')
 	})
 
 	it('surfaces backend errors as a readable message', async () => {
@@ -180,5 +183,18 @@ describe('exec_datatable_sql', () => {
 		runSqlMock.mockRejectedValue(new Error('syntax error'))
 		const parsed = JSON.parse(await run('exec_datatable_sql', { datatable_name: 'main', sql: 'SLECT' }))
 		expect(parsed).toEqual({ success: false, error: 'syntax error' })
+	})
+
+	it('turns the backend "datatable not found" error into an actionable, blocking message', async () => {
+		runSqlMock.mockRejectedValue(new Error('Internal: datatable main not found @workspaces.rs:565:20'))
+		const parsed = JSON.parse(
+			await run('exec_datatable_sql', { datatable_name: 'main', sql: 'CREATE TABLE t (id int)' })
+		)
+		expect(parsed.success).toBe(false)
+		expect(parsed.error).toContain('not configured in this workspace')
+		expect(parsed.error).toContain('Workspace settings → Data Tables')
+		expect(parsed.error).toContain('do not retry')
+		// The raw internal error is not surfaced to the model.
+		expect(parsed.error).not.toContain('workspaces.rs')
 	})
 })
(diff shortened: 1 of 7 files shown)


centdix:
Addressed the **P2 dead-branch nit** from the review (`datatableTools.ts`, the unreachable `Query executed successfully` branch in `exec_datatable_sql`) in 8b35c08fed.

**Fix:**
- Tightened `execDatatableSql`'s return type to a discriminated union — `{ success: true; result: Record<string, any>[] } | { success: false; error: string }` — so the type reflects that a successful run always carries a `result` array (`[]` for DDL/DML).
- With that narrowing, `if (result.success)` guarantees `result.result`, so SELECT rows and zero-row statements share one `rowCount` reporting path. Removed the unreachable `else` block that returned `{ success: true, message: 'Query executed successfully' }`.

No behavior change: the DDL/empty case still returns `{ success: true, rowCount: 0, result: [] }` (covered by the existing test at `datatableTools.test.ts:171`). All 10 `datatableTools.test.ts` tests pass.

PR #9410, opened 2026-06-02
feat(frontend): harmonize diff button placement in script and raw app editors

Guilhem-lm:
## Summary

The "Diff" button (which opens the full `DiffDrawer` — code + metadata + triggers) lived in a different place in each editor: a `DeployButton` split-dropdown entry in the standalone script editor, and a "more" dropdown item in the raw app editor. This PR harmonizes them to follow the **flow editor's pattern** (`FlowBuilder.svelte`): a standalone top-bar button placed right after the overflow "⋮" menu.

Scope is limited to the **standalone script editor** and the **raw app editor**. Simple/normal apps (`AppEditorHeader.svelte`) are intentionally left unchanged. The inline `EditorBar` diff toggle, Debug/breakpoints, and the in-save-drawer Diff buttons are also untouched.

## Changes

- **Canonical spec** (matching `FlowBuilder`): standalone top-bar `Button`, `variant="default"`, `unifiedSize="md"`, `DiffIcon`, label "Diff", disabled when there's no saved/deployed version, `iconOnly` when the top bar is compact, placed immediately after the overflow menu.
- **`ScriptBuilder.svelte`**: extracted `openDiffDrawer()`, removed the "Show diff" entry from the `DeployButton` split-dropdown, added a `diffButton` snippet rendered in both compact and wide layouts (leads the action cluster). Updated the now-stale `inSessionPane` comment.
- **`RawAppEditorHeader.svelte`**: extracted `openDiffDrawer()`, removed the duplicate `moreItems` "Diff" entry, added the standalone button right after `<DropdownV2>`.

The extracted `openDiffDrawer()` functions preserve the exact prior payload logic (deployed/draft/current, `syncWithDeployed()`, draft-triggers and empty-assets handling for scripts).

## Test plan

- [x] Script editor: Diff button leads the action cluster, disabled on new scripts, enabled on deployed scripts → opens `DiffDrawer`
- [x] Raw App editor: Diff button right after the ⋮ menu → opens side-by-side metadata+content diff
- [x] `npm run check:fast` passes for the changed files
- [x] Verify icon-only collapse when the top bar is narrow (compact) — confirmed in browser for both editors at 640px (single icon-only Diff button, 32px, visible)
- [ ] Verify script diff still works from inside an AI session pane
- [ ] Verify `customUi.topBar.diff = false` still hides the ScriptBuilder button

### DIFF
--- a/frontend/src/lib/components/ScriptBuilder.svelte
+++ b/frontend/src/lib/components/ScriptBuilder.svelte
@@ -102,7 +102,6 @@
 	import DraftTriggersConfirmationModal from './common/confirmationModal/DraftTriggersConfirmationModal.svelte'
 	import { Triggers } from './triggers/triggers.svelte'
 	import type { ScriptBuilderProps } from './script_builder'
-	import type { DiffDrawerI } from './diff_drawer'
 	import WorkerTagSelect from './WorkerTagSelect.svelte'
 	import type { ButtonType } from './common/button/model'
 	import DebounceLimit from './flows/DebounceLimit.svelte'
@@ -832,8 +831,7 @@
 
 	function computeDropdownItems(
 		initialPath: string,
-		savedScript: NewScriptWithDraftAndDraftTriggers | undefined,
-		diffDrawer: DiffDrawerI | undefined
+		savedScript: NewScriptWithDraftAndDraftTriggers | undefined
 	) {
 		let dropdownItems: { label: string; onClick: () => void }[] =
 			initialPath != '' && customUi?.topBar?.extraDeployOptions != false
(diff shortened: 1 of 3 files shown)


Guilhem-lm:
Addressed the review nits in 1b7528c:

- **Pi (P2)** — removed the now-unused `diffDrawer` parameter from `computeDropdownItems` (and its call site + the orphaned `DiffDrawerI` import).
- **Claude (nit)** — fixed the stale `ScriptEditorView.svelte` comment that still said "Show diff" → now "Diff".

Left intentionally (not bugs):
- `disabled={!savedScript || !diffDrawer}` keeps the defensive `!diffDrawer` guard, matching the original dropdown gate (`savedScript && diffDrawer`) rather than flow's bare `!savedFlow`.
- `!= false` vs `!==` is deliberate to match `FlowBuilder.svelte`.

`npm run check:fast` passes.

PR #9411, opened 2026-06-02
feat: gate session fork creation on CE workspace cap

Guilhem-lm:
## Summary

Forking a workspace is **not** an enterprise-only feature — CE (community edition / no enterprise license) simply caps the **number of workspaces**. The backend `_check_nb_of_workspaces` rejects creating a new workspace once there are ≥ 2 non-`admins` workspaces (3 total, incl. `admins`), and a fork *is* a new workspace. Previously CE users could stage a fork in the sessions workspace picker even when already at the cap; it would only fail later, server-side, after they committed by sending their first message.

This gates the session "Create new fork" affordance on the **CE workspace cap** (mirroring the backend), not on the enterprise license: licensed instances are uncapped, unlicensed instances allow forks only while under the cap.

## Changes

- **`WorkspaceFamilyPicker.svelte`** (session-only picker, used by both `SessionWorkspaceBar` and `SessionForkBar`): split the structural gate (`forkAffordanceOpen` = cloud / `DisableWorkspaceForking` rule / `admins`) from the cap check. `ceWorkspaceCapReached = !$enterpriseLicense && nonAdminWorkspaceCount >= 2`. `showCreateFork` keeps the interactive row; `showForkUpsell` renders a disabled, non-keyboard-focusable row (`aria-disabled`) with a *"Workspace limit reached"* hint and a tooltip explaining the CE 3-workspace limit and that archiving a workspace or an enterprise license lifts it.
- **`sessionState.svelte.ts`** → `commitSessionWorkspace`: before `materializeFork`, when a `pending_fork` is committed while unlicensed **and** at the cap, throw *"Community edition is limited to 3 workspaces — archive a workspace or pick one to run in"*. `pending_fork` is preserved so the block persists until the user picks a non-fork workspace (which clears it via `setSessionPendingWorkspace`). The throw propagates through `beforeSend` and aborts the send.
- **`sessionState.test.ts`**: added a CE-cap fork-guard test (at cap → throws, `createWorkspaceFork` never called, `pending_fork`/`workspace_id` intact) and an under-cap test (unlicensed but below cap → fork proceeds). Updated the existing fork-creation-failure test to set a license so it still reaches `materializeFork`. All restore store state in `finally`.
- Minor adjacent polish on the same sessions surface: the diff-drawer button in `SessionForkBar.svelte` and `LocalDraftBanner.svelte` is now `variant="default"` with the `Diff` (+/-) icon and the modified-item count, to make the diff entry point more prominent.

Untouched: backend (remains the hard, instance-wide enforcement boundary), sidebar `WorkspaceMenu`, and `SessionForkBar.forksAllowed` (existing-fork review still works in CE).

### Known limitation (preflight only)

The client cap check counts `$userWorkspaces` (the current user's membership-scoped list), whereas the backend cap is instance-wide. On a multi-user CE instance a user who can't see every workspace may be offered a fork that the server then rejects — at which point the pending fork is dropped gracefully. The frontend has no instance-wide count available to a non-superadmin, so this gate is intentionally best-effort *preflight UX*; `_check_nb_of_workspaces` remains the source of truth and fails closed.

## Test plan

- [x] `npm run check:fast` — no problems
- [x] `vitest sessionState.test.ts` — 20 passed (incl. new cap guard + under-cap tests)
- [x] Svelte autofixer — no issues
- [x] Browser (Playwright): **licensed** path renders the enabled interactive "Create new fork…" row; **unlicensed + at cap** renders the disabled "Workspace limit reached" row while root + existing forks stay selectable
- [ ] Manual check on a real CE instance: confirm the disabled row is not keyboard-focusable and arrow-key nav lands correctly; confirm a `pending_fork` staged at the cap surfaces the thrown error rather than silently retrying

### DIFF
--- a/frontend/src/lib/components/LocalDraftBanner.svelte
+++ b/frontend/src/lib/components/LocalDraftBanner.svelte
@@ -3,7 +3,7 @@
 	import DiffDrawer from '$lib/components/DiffDrawer.svelte'
 	import { classes } from '$lib/components/common/alert/model'
 	import { type Value } from '$lib/utils'
-	import { AlertCircle } from 'lucide-svelte'
+	import { AlertCircle, Diff } from 'lucide-svelte'
 	import { twMerge } from 'tailwind-merge'
 	import { slide } from 'svelte/transition'
 
@@ -84,7 +84,8 @@
 		<div class="flex flex-row items-center gap-2 shrink-0">
 			<Button
 				unifiedSize="sm"
-				variant="subtle"
+				variant="default"
+				startIcon={{ icon: Diff }}
 				btnClasses={classes.warning.titleClass}
 				on:click={showDiff}>Show diff</Button
 			>
--- a/frontend/src/lib/components/sessions/SessionForkBar.svelte
+++ b/frontend/src/lib/components/sessions/SessionForkBar.svelte
@@ -2,6 +2,7 @@
 	import {
 		Archive,
 		ArrowRight,
+		Diff,
 		GitCompareArrows,
 		GitFork,
 		GitMerge,
@@ -189,9 +190,9 @@
 		</div>
 		<div class="flex items-center gap-1 shrink-0">
 			<Button
-				variant="subtle"
+				variant="default"
 				unifiedSize="xs"
-				startIcon={{ icon: GitCompareArrows }}
+				startIcon={{ icon: Diff }}
 				disabled={totalDiffs === 0}
 				title="{totalDiffs} modified item{totalDiffs === 1 ? '' : 's'}"
 				onclick={() => diffDrawer?.open()}
--- a/frontend/src/lib/components/sessions/WorkspaceFamilyPicker.svelte
+++ b/frontend/src/lib/components/sessions/WorkspaceFamilyPicker.svelte
@@ -1,11 +1,17 @@
 <script lang="ts">
 	import { tick, type Snippet } from 'svelte'
-	import { userWorkspaces, workspaceStore, type UserWorkspace } from '$lib/stores'
+	import {
+		enterpriseLicense,
+		userWorkspaces,
+		workspaceStore,
+		type UserWorkspace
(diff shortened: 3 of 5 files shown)


Guilhem-lm:
Thanks for the reviews. Important context for the latest round (commit `b5cfea4`): **the premise of this PR changed mid-review.** Forking is *not* an enterprise-only feature — CE just caps the number of workspaces (`_check_nb_of_workspaces`: ≥ 2 non-`admins` is rejected, i.e. 3 total). The gate was corrected from an EE-license check to a CE workspace-cap check, and I've now updated the (previously stale) PR title/description to match. That stale framing is what several findings below are reacting to.

**@cubic-dev-ai — P0 `sessionState.svelte.ts:308` / P1 `WorkspaceFamilyPicker.svelte:85`** ("unlicensed users under the cap can still create forks"): this is the *intended* behavior under the corrected spec. CE users are allowed to fork up to the workspace cap; only at/above the cap (or where forking is structurally disabled) is it blocked. Both findings evaluate against the old EE-gate premise, so no code change here.

**@github-actions (Codex) — P1 instance-wide undercount**: correct and acknowledged. `$userWorkspaces` is membership-scoped, so on a multi-user CE instance it can undercount the backend's instance-wide cap and offer a fork the server then rejects (dropping the pending fork). A non-superadmin has no instance-wide count available client-side, so this gate is deliberately best-effort *preflight UX* — `_check_nb_of_workspaces` remains the enforcement boundary and fails closed. The mismatch degrades to a late, graceful rejection rather than an incorrect fork. Documented as a known limitation in the PR description.

**@github-actions (Pi) — adjacent cosmetic changes** in `LocalDraftBanner.svelte` / `SessionForkBar.svelte`: intentional. They make the diff-drawer entry point on the same sessions surface more prominent (`variant="default"` + `Diff` +/- icon + modified-item count); bundled deliberately, noted in the description.

PR #9416, opened 2026-06-02
fix(backend): authorize single-job read endpoints by job/flow visibility

rubenfiszel:
## Summary

Fixes a confidentiality bug where an authenticated workspace member (e.g. a viewer with no ACL on the runnable) could read **any** job's args / result / logs / traces by UUID through the single-job read endpoints — they fetched via the root DB filtered only by job id + workspace, bypassing the RLS that `jobs/list` enforces. Adds a proper authorization gate, a perf cache, flow-ancestor visibility inheritance, and a **share read link** so access can be granted intentionally.

## Authorization model

An authenticated caller may read a job iff:
- they **created** it (`created_by`) — covers app components, webhooks, and own runs (whose `permissioned_as` is the policy identity, not the caller); or
- the job **or any flow in its `parent_job` ancestor chain** is visible to them under the **same RLS as `jobs/list`** (admins `BYPASSRLS`); or
- they present a valid **share read link** token for the job (or an ancestor flow).

Otherwise: **403** with guidance to request a share link when the run exists, **404** when it doesn't. Unauthenticated callers are unchanged (anonymous-only); approval-token / suspended-job-secret bypasses are preserved (the suspended secret now also works for authenticated approvers).

## Changes

- `require_job_read_access` gating all single-job read handlers: `get_job`, `get_completed_job`(`_result`/`_maybe`/`_timing`), `get_args`, `get_job_logs`, `get_completed_job_logs_tail`, `get_flow_all_logs`, `get_log_file`, `get_job_update`/`_sse`, `get_flow_job_debug_info`, `get_result_by_id`, `get_otel_traces`.
- **Cache**: size-bounded LRU keyed by a SHA-256 of the caller's authz-relevant identity (is_admin/username/override/groups/folders) + job id; no TTL (job-side inputs are immutable, caller-side changes re-key). Mirrors apps' `PERMIT_CACHE`.
- **Flow inheritance**: visibility walks the full `parent_job` chain so a flow you can see lets you inspect its steps even when steps reference runnables you lack ACL on.
- **Share read links**: `GET /jobs/job_view_token/{id}` (only mintable by someone who can already read it) returns a stateless `{job_id}.{HMAC(workspace_key, job_id‖"view_token")}`; accepted as a `view_token` query param (SSE/links) or `X-View-Token` header (a single frontend request interceptor, no client regen). Run page gains a **Share** button and renders the 403 case with instructions.

## Test plan

- [x] `cargo test --test jobs_read_auth` — viewer denied (403, no secret) across all read endpoints incl. `result_by_id`/`get_otel_traces`; owner/admin/app-launcher allowed; flow-root + middle-ancestor inheritance; share-token mint/scope/subtree/garbage/non-reader; 403-vs-404 distinction; unauthenticated unchanged
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api`; `npm run check:fast` (no new errors)
- [x] **Live end-to-end** against the dev server: non-admin member 403'd on an admin's run; owner mints token; member reads via `?view_token` and `X-View-Token`; garbage/other-job tokens rejected; non-existent → 404
- [x] EE build N/A (no `*_ee.rs` touched)
- [x] **Browser UI verified** (real Chromium): non-ACL member sees the "request a share link" panel; opening `?view_token=` renders the full run + result; Share button present

### DIFF
--- a/backend/.sqlx/query-19513c4158267cc7fe10d999ad571052c112e6bbb3cf834f16176cbb7e1ac319.json
+++ b/backend/.sqlx/query-19513c4158267cc7fe10d999ad571052c112e6bbb3cf834f16176cbb7e1ac319.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "WITH RECURSIVE chain(id, parent_job) AS (\n                SELECT id, parent_job FROM v2_job WHERE id = $1 AND workspace_id = $2\n                UNION ALL\n                SELECT j.id, j.parent_job FROM v2_job j\n                    JOIN chain c ON j.id = c.parent_job AND j.workspace_id = $2\n            )\n            SELECT id AS \"id!\" FROM chain",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Uuid"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "19513c4158267cc7fe10d999ad571052c112e6bbb3cf834f16176cbb7e1ac319"
+}
--- a/backend/.sqlx/query-8e8933fc6648a88dc35cd81559a31d10678d6c68fc920c876914e71324d5e460.json
+++ b/backend/.sqlx/query-8e8933fc6648a88dc35cd81559a31d10678d6c68fc920c876914e71324d5e460.json
@@ -1,0 +1,24 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT EXISTS(SELECT 1 FROM v2_job WHERE id = $1 AND workspace_id = $2 AND tag = ANY($3))",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "exists",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text",
+        "TextArray"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "8e8933fc6648a88dc35cd81559a31d10678d6c68fc920c876914e71324d5e460"
+}
--- a/backend/.sqlx/query-ca5bb402834502432f3d7260fdd5b9fb568a4c77e2a91f55575a93461d5a7f50.json
+++ b/backend/.sqlx/query-ca5bb402834502432f3d7260fdd5b9fb568a4c77e2a91f55575a93461d5a7f50.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT EXISTS(SELECT 1 FROM v2_job WHERE id = ANY($1) AND workspace_id = $2)",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "exists",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "UuidArray",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "ca5bb402834502432f3d7260fdd5b9fb568a4c77e2a91f55575a93461d5a7f50"
+}
--- a/backend/tests/fixtures/jobs_read_auth.sql
+++ b/backend/tests/fixtures/jobs_read_auth.sql
@@ -1,0 +1,169 @@
+-- Fixture for the single-job read authorization regression test
+-- (see tests/jobs_read_auth.rs).
+--
+-- Users available from `base`:
+--   test-user   (admin,     token SECRET_TOKEN)
+--   test-user-2 (User,      token SECRET_TOKEN_2)  -- owner of the secret script
+--   test-user-3 (User,      token SECRET_TOKEN_3)  -- the unprivileged "viewer"
+--
+-- test-user-3 is NOT a member of any folder/group granting access to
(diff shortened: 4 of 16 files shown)


rubenfiszel:
Addressed the CI review feedback in `eabc342`:

- **[P1, Codex] share-token mint bypassed `if_jobs:filter_tags` scopes** — `get_job_view_token` now applies the same tag-scope predicate the read handlers use before minting, so a tag-scoped token can't mint a transferable link for a job outside its allowed tags. Added a scoped-token regression test (`SCOPED_DENO_TOKEN`): mint allowed for the in-scope `deno` job, denied (404) for the out-of-scope `flow` job.
- **[Claude/Pi] non-constant-time HMAC compare in `validate_view_token`** — now recomputes the HMAC and uses `HmacSha256::verify_slice` (constant-time), matching `verify_suspended_secret`, instead of comparing hex strings with `!=`.
- **[Claude] authed + invalid suspended-secret behavior in `get_completed_job_result`** — an authenticated reader who attaches a stale/garbage suspended-secret triple now falls through to the normal visibility gate instead of erroring out; unauthenticated callers (for whom the secret is the only credential) are still rejected.
- **[Claude] cache-key separators not length-prefixed** — every variable-length field in `job_read_access_cache_key` is now u32-length-prefixed, so no choice of username/group/folder values can make two distinct identities collide.

Left as a documented follow-up (pre-existing, flagged out-of-scope by the review): `get_root_job` still resolves a root-job UUID via the root DB without the gate — it discloses only a single UUID, not job data.

`cargo test --test jobs_read_auth` + `SQLX_OFFLINE=true cargo check -p windmill-api` both green.

rubenfiszel:
Addressed the latest round in `27156c8`:

- **[P2, Codex] `job_view_token` missing from the public API surface** — added `GET /w/{workspace}/jobs/job_view_token/{id}` (operationId `getJobViewToken`) to `openapi.yaml`, and switched the run page's Share button from a raw `fetch` to the generated `JobService.getJobViewToken`. (`frontend/src/lib/gen/` is gitignored and generated from the spec at build time, so the spec is the source of truth.)

Pi's two observations are noted as intentional / non-actionable: the fetch-then-authorize ordering matches the existing handler pattern (no data returned unless the gate passes), and `get_log_file` shares the identical `created_by → require_job_read_access` path (its filesystem read isn't exercisable in the test harness). Per maintainer, `get_root_job` stays ungated — it returns only a root-job UUID, not job data.

`cargo test --test jobs_read_auth` + `npm run check:fast` (no new errors) green.

rubenfiszel:
Addressed in `1930c11`:

- **[P2, Codex] share-link downloads omitted the view token** — download actions bypass the request interceptor (`downloadViaClient` uses raw `fetch` + `getHeaders`; cookie-mode downloads use plain `<a href>`). Added an `appendViewToken()` helper and applied it to the job download paths — result (`DisplayResult`, `DisplayResultControlBar`), logs (`LogViewer`), args (`JobArgs`), and flow "Download all logs" (`FlowStatusViewerInner`). Appending the `view_token` query param covers both the client-fetch and href modes (the backend reads the param on all gated handlers). `get_log_file` and the JSON `get_completed_logs_tail`/result fetches already go through the generated client, so the `X-View-Token` interceptor covers them.

Live-verified with a share-link viewer: `get_logs`/`get_args`/`completed/get_result` return 200 with `?view_token=` and 403 without. `npm run check:fast` shows no new errors.

rubenfiszel:
Addressed in `f551b24`:

- **[P1, Codex] view-token use-side bypassed tag scopes** — moved the `if_jobs:filter_tags` check into `require_job_read_access`, where it runs **before** any `created_by` / `view_token` / RLS grant. This enforces tag scope uniformly across every gated handler, including the ones that don't tag-filter their own data query (`result_by_id`, `get_flow_job_debug_info`, `get_otel_traces`). A tag-scoped token can no longer use someone else's valid share token to read out-of-scope data. The explicit mint-side check in `get_job_view_token` is now redundant (it goes through `require_job_read_access`) and was removed. `get_scope_tags` is `None` for unscoped callers, so no extra query is added in the common case.

Added a use-side regression test: a `deno`-scoped token presenting a valid `flow` view_token is denied on `get_otel_traces` and `result_by_id` (404), while an in-scope (`deno`) view_token still works.

`cargo test --test jobs_read_auth` + `SQLX_OFFLINE=true cargo check -p windmill-api` green (no new warnings).

rubenfiszel:
Addressed in `7d09182`:

- **[P1, Codex] share link omitted workspace context** — the copied URL now pins `workspace=${$workspaceStore}` alongside `view_token`. The token is signed with the run's workspace key, and the logged layout only switches `$workspaceStore` when the URL carries `workspace=`; without it a recipient whose persisted active workspace differs would open the run (and validate the token) against the wrong workspace and get not-found/forbidden. (Matches what I hit in the earlier browser test, which only worked with `?workspace=` present.)

`npm run check:fast`: no new errors. The backend share-token flow + the `?workspace=&view_token=` link were already verified working in the browser e2e.

rubenfiszel:
Addressed in `59981f4`:

- **[P1, Codex] `get_result_maybe?get_started=true` skipped authorization for queued jobs** — the handler only gated when a *completed* row existed; with `get_started=true` a non-reader reached the fallback branch and got `started: true` for a running private job. Now the `get_started` branch fetches `created_by` and authorizes (created_by / view_token / RLS, or anonymous for unauthenticated callers) **before** disclosing running-state. A non-existent job still returns `started: false` (leaks nothing), preserving the polling contract.

Added a regression test with a queued (no completed row) private job: a non-reader is denied (403, no `started` field in the body), the owner still sees `started: true`.

`cargo test --test jobs_read_auth` + `SQLX_OFFLINE=true cargo check -p windmill-api` green (no new warnings; the `created_by` fetch reuses an existing cached query).

PR #9420, opened 2026-06-02
feat(apps): opt-in sandbox isolation for published & raw apps (alpha)

hugocasa:
## Summary

Published apps and raw apps render user-authored content in the browser. This PR adds **opt-in sandbox isolation (alpha)**: a publisher can isolate an app from the surrounding Windmill session by rendering it inside a **sandboxed, opaque-origin iframe** with only the least-privilege access it needs, instead of running it in the main app context with the full session.

**Isolation is off by default.** Apps keep running exactly as they do today (same-origin, full session) until a publisher opts in from the Deploy drawer — so there is **no migration, no behavior change on upgrade, and no prompt**. It's shipped as **alpha** so we can collect feedback before considering a default change.

No separate domain is required: the sandbox gives the app document an opaque origin that the browser isolates from the main UI by same-origin policy. `PUBLIC_APP_DOMAIN` remains an optional alternative.

## How each app type is handled when isolation is enabled

- **Low-code apps** call the Windmill API directly from the browser, so they render in an **opaque viewer iframe** and the embedder mints a **short-lived, narrowly-scoped token** handed to the iframe — the session cookie never crosses into it. The token is confined to exactly what a running app needs (see scope list below); notably it carries **no `apps:write`**.
- **Raw apps** never call the API directly — the author bundle talks only to a parent `postMessage` bridge that runs the app's declared runnables, so the bundle is given **no credential** and runs in its **own opaque iframe**. The bridge runs on the page with the page credential, so a raw app renders as a **single opaque iframe** and needs no embed token.
- **In-workspace viewing** (`/apps/get`, `/apps_raw/get`) goes through the same machinery, so the isolation state of an app is identical on every surface it is viewed from.

## Default vs opt-in — no migration

- **Default (every existing and new app):** unsandboxed — the app runs same-origin with the viewer's full session, the pre-isolation behavior. Nothing changes on upgrade, redeploy, CLI/git-sync push, or cross-workspace promotion.
- **Opt-in:** the publisher flips a single **"Sandbox isolation" toggle (alpha)** in the Deploy drawer, which sets `Policy.sandbox`. It's an ordinary declarative policy field — no server-managed flag, no migration, no consent prompt.
- This keeps the rollout safe: the installed base is untouched, and isolation is an explicit, reversible per-app choice while we gather feedback.

## Embed-token scope (isolated low-code apps only)

The minted token is confined to the routes a running app actually uses; everything else returns **403**:

- `apps:read`/`apps:run` — **path-scoped to the app's own path**: fetch the app's own definition and `execute_component` (run the app's own runnables), but no other app's. These path-scoped checks are confined to the embed token, so other caller types keep their prior access. No `apps:write`, so app-management routes are unreachable.
- `jobs:read` — completed-job results/updates **by id only**; workspace-wide job enumeration/export is denied.
- `resources:run` — resource **metadata only** (pickers, type schemas); resource **values** are never exposed.
- `users:read` — `users/whoami` only; every other `/users` route is denied.
- `folders:read` — `folders/listnames` only; every other `/folders` route is denied.

These domains are CORS-enabled for the opaque iframe, so the allow/deny matrix is enforced in the route scope-checker (default-deny via an `app_embed` sentinel) and locked by unit tests. Widen `APP_EMBED_SCOPES` if a legitimate need shows up.

## What changes for app authors (isolated apps only — default apps are unaffected)

- **App code is isolated from the session** — `document.cookie`, `window.parent`/`window.top` are unavailable.
- **Low-code API access is scoped** (see above): frontend scripts can run the app's own runnables but not trigger arbitrary scripts/flows via `jobs/run/*`; `variables/*`, `scripts/*`/`flows/*` listing, the AI proxy, and the S3 file **upload** endpoint (a write) are out of scope.
- **Web storage is shimmed and scoped per app** — the relayed `localStorage` store is keyed by workspace + app path, so one isolated app cannot read or clobber another's storage (even two apps at the same path in different workspaces). `sessionStorage` is in-memory only. **Not shimmed:** IndexedDB, Service Workers, WebAuthn, cross-tab coordination, third-party-origin SDKs/OAuth redirects.
- **Raw bundle URL semantics change**: the bundle document is not a same-origin `blob:` — `location.origin` is `"null"`, relative `fetch()` resolves against the wrapper path, and direct API calls carry no credential (use the bridge / `window.ctx`).
- **Raw runnable bridge is job-scoped**: an isolated bundle can only `waitJob`/`getJob`/`streamJob` ids it launched itself (default/unsandboxed renders are exempt — they keep the pre-isolation behavior of polling persisted or runnable-returned ids).
- **Navigation works** — apps can redirect the whole tab and open new tabs; they cannot read/script the parent. The navbar component, frontend-script `goto`, and `gotoUrl` side effects are relayed to the top page. In-app hash/query changes are mirrored to the page URL via `replaceState` (no history entries).
- **Reserved query params**: `wm_embed`, `wm_embedder_origin`, `wm_coep` are transport-internal and stripped from `ctx.query`.
- **Browser features in the opaque iframe**: clipboard and fullscreen are delegated; camera/microphone/geolocation are not. Popups (e.g. OAuth flows) escape the sandbox, but their cookies land on the real origin which the app cannot read.
- **Embed token TTL is 12h**, refreshed via the embedder on app-load 401s.

## What changes for operators

- **`PUBLIC_APP_DOMAIN` is no longer required** for isolation; it stays optional.
- **No silent breakage on upgrade** — isolation is opt-in, so existing apps are untouched.
- The opaque low-code viewer makes cross-origin API calls, so the relevant endpoints serve permissive CORS — safe because it is bearer-token-based with no cookies (`ACAO: *`, credentials not allowed). Adds access only; same-origin and non-browser callers are unaffected.
- **Embedding an isolated app in a cross-origin-isolated page** (COEP `require-corp`): append `?wm_coep=1` to the app URL — the wrapper then asserts COEP so the nested document is allowed to load.
- Browsers holding a **cached pre-upgrade SPA bundle** pick up the new render behavior on the next refresh (transient).

## Known edge cases / intentional differences

- The legacy `/apps/get_raw/{version}/{path}` viewer route (which rendered the bundle with no isolation regardless) is retired; the apps list now opens the unified viewer. Stale links **redirect to the unified viewer** (preserving query + hash) and resolve to the **latest** version — the pinned `{version}` is dropped, matching every other in-workspace app link (which never pinned a version).
- Opening the opaque viewer route (`wm_embed`) **outside** its Windmill embedder (e.g. its iframe `src` embedded directly in a third-party page) shows a short diagnostic message instead of an indefinite loader.
- The embed-token endpoints read the stored policy **leniently** (only the isolation-decision field), so apps with pre-existing policy shapes keep rendering; no token is minted for raw or unsandboxed renders at all (none is consumed there).
- The backend raw-app wrapper document is **always** served `CSP: sandbox`; the unsandboxed (default) same-origin experience is built client-side, so no real-origin URL ever serves an app with the viewer's session attached to it.
- In-workspace rendering keeps exact pre-sandbox presentation parity (no "Powered by Windmill" badge, no HTML-result approval gate, `hideRefreshBar` honored, query/hash forwarded into the viewer).

## Changes

- **Backend**: `GET /apps_u/embed_token/{secret}` + authenticated `GET /apps/embed_token/p/{path}` (access check + `raw_app`/`sandbox`, mints the scoped token only for an isolated low-code render); always-`CSP: sandbox` raw-app wrapper (`nosniff`/CORP, opt-in COEP); `Policy.sandbox` opt-in flag; tightly-scoped `APP_EMBED_SCOPES` confined per-route via an `app_embed` sentinel; CORS on the workspace endpoints the low-code iframe calls.
- **Frontend**: `PublicAppFrame` — opaque viewer + scoped token for isolated low-code; single opaque bundle iframe for raw; same-origin for the default (unsandboxed) render; per-app web-storage shim (keyed by workspace + path); navigation relay; Deploy-drawer "Sandbox isolation" toggle (alpha); unified in-workspace viewer.
- **EE companion**: windmill-labs/windmill-ee-private#600.

## Test plan

- [x] Default (unsandboxed) low-code app → renders same-origin/inline, no prompt, on `/apps/get`, `/public`, `/a/{custom_path}`
- [x] Default (unsandboxed) raw app → bundle runs same-origin (reports "no-sandbox mode")
- [x] Deploy drawer shows the "Sandbox isolation" toggle, **off by default**, for low-code and raw editors
- [x] Opt a low-code app in → embed token is minted with `sandbox:true`; opaque viewer iframe (`window.parent` → `SecurityError`, empty cookie)
- [x] Opt a raw app in → still no token (single opaque bundle iframe), bundle gated to its own job ids
- [x] Embed-token scope: `users/whoami`, `folders/listnames`, resource metadata, `execute_component` → 200; `users/list`, `users/list_usage`, `folders/list`, `jobs/completed/export`, `jobs/list`, resource value reads, `scripts/list`, `apps/update` → 403
- [x] Updating an app via CLI/API never enables isolation implicitly (`sandbox` is a normal declarative field)
- [x] `PUBLIC_APP_DOMAIN` unset still isolates an opted-in app

<details><summary>Verification notes</summary>

Exercised against a local dev instance (`PUBLIC_APP_DOMAIN` unset), driving the running frontend in a real browser and probing the embed-token endpoints directly.

- **Default render**: a default low-code app renders inline same-origin (no opaque iframe) and a default raw app renders as a same-origin `blob:` frame (reports "no-sandbox mode"), with no prompt, on `/apps/get` and the public surfaces. The embed-token endpoint returns `{token:null, raw_app, sandbox:false}` for these (no token minted).
- **Opt-in mint**: flipping `Policy.sandbox` to true makes the low-code endpoint mint a scoped token (`sandbox:true`); a raw app stays token-less even when isolated (`{raw_app:true, sandbox:true, token:null}`) — the bundle's own opaque iframe + the job-scoped bridge provide its isolation.
- **Scope confinement** (live, with a minted token): `users/whoami` → 200, `folders/listnames` → 200, resource metadata → 200, `execute_component` → allowed; `users/list` / `users/list_usage` / `folders/list` / `jobs/completed/export` / `jobs/list` / resource value reads / `scripts/list` / `apps/update` → 403. Backend unit tests lock the full allow/deny matrix.
- **Deploy drawer**: the "Sandbox isolation" section (alpha badge) shows the toggle off by default; enabling it surfaces an alpha notice asking the publisher to verify the app from its public URL and report breakage.

</details>

## Screenshots

**Publisher — Deploy drawer "Sandbox isolation" section** (opt-in, **off by default**)

sandbox_section_off

**…toggled on** (alpha notice: verify from the public URL and report any breakage)

sandbox_section_on


rubenfiszel:
/ai review exhaustively this PR given we plan to deploy in 3 phases:
1. opt-in and no modal on view to warn that it's not sandboxed (ask customers to test it)
2. opt-in and modal on view
3. opt-out and modal in view

rubenfiszel:
/ai review exhaustively this PR given we plan to deploy in 3 phases:

    opt-in and no modal on view to warn that it's not sandboxed (ask customers to test it)
    opt-in and modal on view
    opt-out and modal in view

hugocasa:
/ai review exhaustively this PR given we plan to deploy in 3 phases:

opt-in and no modal on view to warn that it's not sandboxed (ask customers to test it)
opt-in and modal on view
opt-out and modal in view

hugocasa:
/ai review exhaustively this PR given we plan to deploy in 3 phases:

opt-in and no modal on view to warn that it's not sandboxed (ask customers to test it)
opt-in and modal on view
opt-out and modal in view

hugocasa:
/claude review exhaustively this PR given we plan to deploy in 3 phases:

opt-in and no modal on view to warn that it's not sandboxed (ask customers to test it)
opt-in and modal on view
opt-out and modal in view

PR #9422, opened 2026-06-02
fix: omit temperature for gpt-5+ and o-series models on all providers

centdix:
## Summary

Fixes WIN-2010.

The copilot sends `temperature: 0` by default for deterministic code generation. gpt-5.5 and the o-series reasoning models reject any non-default temperature (only `temperature=1` is accepted) and return a **400**. Previously the "skip temperature" logic was gated on `provider === 'openai' || provider === 'azure_openai'`, so the *same* models routed through the **Custom AI** provider (a gateway) still received `temperature: 0` and failed.

This makes the temperature-omission **provider-agnostic** by folding the reasoning-model check into the shared `modelDisallowsSamplingParams` helper, so it applies regardless of which provider/gateway routes the request.

### Why keep sending `temperature: 0` at all?
The bug was never "we send temperature" — it's that the *"does this model reject it"* check was provider-specific. `temperature: 0` is the deliberate choice for deterministic copilot output (the FIM autocomplete path hard-codes it too). Dropping it everywhere would regress determinism for the majority of models that honor it (gpt-4o, claude, mistral, gemini, …) — a separate product decision, out of scope here. So the fix keeps `temperature: 0` where the model honors it and omits it only where the model rejects it.

## Changes

- **`modelConfig.ts`**: `modelDisallowsSamplingParams` now also matches gpt-5+ and o-series reasoning models. The check strips any provider prefix (e.g. OpenRouter's `openai/o3`) and matches the **bare** model id.
  - `gpt-5` arm: `startsWith('gpt-5')` (does not collide with `gpt-4o`).
  - o-series arm: `/^o\d/` — requires a digit after the `o` so it matches `o1`/`o3`/`o4-mini` but **not** unrelated non-OpenAI ids that merely start with `o` (Mistral `open-mistral-*`/`open-mixtral-*`, OpenRouter `optimus-*`/`openchat-*`), preserving their deterministic temperature.
- **`lib.ts`**: the `openai`/`azure_openai` `max_completion_tokens` branch now reuses `modelDisallowsSamplingParams` instead of duplicating the `startsWith('o') || startsWith('gpt-5')` check.
- **`lib.test.ts`**: added cases for the customai gpt-5.5 / o3 fix, the gpt-4o / `openai/gpt-4o` non-collision, and the non-OpenAI `o`-prefixed false-positive guard (Mistral `open-*`).

### Note on deviation from the issue text
The issue proposed a bare `startsWith('o')`. Applied provider-agnostically that over-matches real non-reasoning ids (Mistral `open-mistral-7b`, OpenRouter `optimus-alpha`, …), silently dropping their `temperature: 0`. Tightening to `/^o\d/` keeps the intended fix while avoiding that regression — flagged and fixed during local review.

## Test plan

- [x] `npm run test:unit -- run src/lib/components/copilot/lib.test.ts` → 17 passed
- [ ] Configure a **Custom AI** provider pointing at a gpt-5.5 gateway, trigger a copilot generation, confirm the request omits `temperature` (no 400)
- [ ] Configure an **OpenAI** provider with `o3` / `gpt-5`, confirm the request omits `temperature` and uses `max_completion_tokens`
- [ ] Configure a **Mistral** provider with `open-mixtral-8x7b`, confirm the request still includes `temperature: 0`

### DIFF
--- a/frontend/src/lib/components/copilot/lib.test.ts
+++ b/frontend/src/lib/components/copilot/lib.test.ts
@@ -28,6 +28,13 @@ describe('modelConfig', () => {
 		expect(modelDisallowsSamplingParams('anthropic/claude-opus-4-7')).toBe(true)
 	})
 
+	it('flags Opus 4.8 model IDs via includes matching', () => {
+		expect(modelDisallowsSamplingParams('claude-opus-4-8')).toBe(true)
+		expect(modelDisallowsSamplingParams('claude-opus-4-8@20260416')).toBe(true)
+		expect(modelDisallowsSamplingParams('claude-opus-4-8/thinking')).toBe(true)
+		expect(modelDisallowsSamplingParams('anthropic/claude-opus-4-8')).toBe(true)
+	})
+
 	it('omits deterministic temperature for Anthropic Opus 4.7 chat requests', () => {
 		expect(
 			getDefaultChatTemperature({ provider: 'anthropic', model: 'claude-opus-4-7' })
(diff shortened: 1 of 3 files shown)


centdix:
## Verification — confirmed mergeable

Validated the fix at three levels, including a live end-to-end run against real OpenAI through a Custom AI gateway.

### 1. Unit tests (`lib.test.ts`, 18 passing)
`getDefaultChatTemperature` / `modelDisallowsSamplingParams` cover:
- `customai` + `gpt-5.5` / `o3` → temperature omitted
- `gpt-4o`, `openai/gpt-4o`, Mistral `open-*`, `optimus-*`, `openchat-*` → `temperature: 0` kept (false-positive guards for the `/^o\d/` match)
- `gpt-5*`, `o\d*`, `claude-opus-4-7`, `claude-opus-4-8` → flagged

### 2. HTTP-level proof (LiteLLM → real OpenAI, gpt-5.5 live on the key)
Reproduced the exact request shapes Windmill emits:
| Request | Result |
|---|---|
| reasoning model + `temperature:0` (old/buggy shape) | **400** — `temperature does not support 0 with this model` |
| reasoning model, no temperature, `max_tokens` (fixed shape) | **200** ✅ |
| `gpt-4o` + `temperature:0` (control) | **200** ✅ |

This also confirms LiteLLM auto-translates `max_tokens → max_completion_tokens`, so the `max_tokens` note below does **not** bite on lenient gateways.

### 3. Full end-to-end through Windmill
Configured a **Custom AI** provider pointing at a LiteLLM proxy (OpenAI-compatible gateway), model `gpt-5.5`, and ran an AI-chat generation from the UI: **succeeds with the fix**, whereas `main` 400s with the temperature error. Confirmed working.

### Known limitation (deliberate follow-up, not blocking)
For gpt-5/o-series on non-`openai`/`azure_openai` providers, Windmill still sends `max_tokens` (not `max_completion_tokens`). Lenient gateways (LiteLLM, etc.) translate it; a *strict* OpenAI-passthrough gateway would still 400 on `max_tokens`. This is **not** fixable by reusing `modelDisallowsSamplingParams` for the token-param branch — that predicate also matches `claude-opus-4-x`, which must keep `max_tokens`. The robust, gateway-agnostic fix is an error-driven retry (send, and on a 400 about the token param retry with the other key), tracked as a follow-up.

PR #9428, opened 2026-06-03
fix(mcp): resolve MCP resource token via caller RLS + SSRF-guard url

rubenfiszel:
## Summary

Security fix for a workspace variable exfiltration + SSRF issue via the MCP `mcp_tools` resource indirection.

`McpClient::from_resource` resolved the resource's `token` field (a `$var:` reference) with `get_secret_value_as_admin` on the **bare DB pool** — no RLS, no audit — then sent the decrypted secret as an `Authorization: Bearer` header to the resource author-controlled `url` with **no SSRF check**. Any developer-role user could:
- read a secret variable in a folder they have no access to (e.g. an admin-only folder), by pointing an MCP resource's token at it, and
- use the request as an SSRF primitive against internal services (e.g. `http://169.254.169.254`).

Reachable via `GET /api/w/{w}/resources/mcp_tools/{path}` and via the AI-agent path (`load_mcp_tools`).

## Changes

- `windmill-mcp` `from_resource` no longer reads secrets itself: it now takes an **already-resolved** `token: Option<String>`, and validates the resource URL with `windmill_common::ssrf::validate_url_for_ssrf` **before** connecting.
- API handler (`mcp_tools.rs`): resolves the token through the caller's permissioned path — `get_value_internal` over `DbWithOptAuthed::from_authed(&authed, db, user_db)` — applying the same variable RLS as `/variables/get_value` and auditing the secret read. An unauthorized read now returns 401 before any connection.
- Worker AI-agent path (`ai/utils.rs`): resolves the token via `AuthedClient::get_variable_value` (the permissioned `/variables/get_value` endpoint carrying the job's own identity token), instead of the bare-pool admin read. `load_mcp_tools` now takes `&AuthedClient`.
- Removed the only vulnerable `get_secret_value_as_admin` usage (the helper remains for its legitimate admin callers elsewhere).

## Test plan

- [x] `cargo test -p windmill-mcp` — new unit test `from_resource_rejects_ssrf_url` (rejects `169.254.169.254` before connecting, no network needed)
- [x] `cargo test --features mcp,oauth2 --test mcp_token_exfil` — integration regression test: a developer who can read the MCP resource but not the locked secret gets **401** (denied at the variable-RLS gate, before connecting); an admin clears the gate, the token resolves, and the request only fails later at the connect/SSRF step (no over-blocking). Fails pre-fix.
- [x] `cargo check --features mcp,oauth2 -p windmill-api` and `cargo check --features quickjs -p windmill-worker` (non-mcp stub path) both compile cleanly.
- [ ] EE-only: confirm a `variables.decrypt_secret` audit row is emitted on authorized reads (audit logging is a no-op in OSS).

### DIFF
--- a/backend/tests/fixtures/mcp_token_exfil.sql
+++ b/backend/tests/fixtures/mcp_token_exfil.sql
@@ -1,0 +1,29 @@
+-- Fixture for the MCP token-exfiltration regression test (GHSA-8m2p-2crh-9h3w).
+--
+-- Models a malicious developer (test-user-3, a plain workspace member) who:
+--   - owns an MCP resource they are allowed to read, and
+--   - points that resource's `token` field at a secret variable living in a
+--     folder they have NO access to (`f/locked`, only test-user/admin owns it).
+--
+-- The secret variable `f/locked/secret_token` itself is inserted by the test in
+-- Rust (so it is encrypted with the real workspace key); this fixture only sets
+-- up the locked folder, the resource, and their permissions.
+
+-- Folder the developer cannot read (empty extra_perms, owned by admin only).
+INSERT INTO folder (workspace_id, name, display_name, owners, extra_perms, created_by)
+VALUES ('test-workspace', 'locked', 'Locked Folder', '{"u/test-user"}', '{}', 'test-user');
+
+-- MCP resource owned by the developer (so RLS lets them read the resource),
+-- whose token references the locked secret. The URL is a non-resolvable public
+-- host so that, for an authorized caller, resolution succeeds but the later
+-- connection/SSRF step fails deterministically without network access.
+INSERT INTO resource (workspace_id, path, value, description, resource_type, extra_perms, created_by)
+VALUES (
+    'test-workspace',
+    'u/test-user-3/evil_mcp',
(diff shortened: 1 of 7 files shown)


rubenfiszel:
Thanks @codex — addressed the redirect half of the finding in dbd3942.

**Redirect bypass (fixed):** the MCP `reqwest` client now builds with `.redirect(reqwest::redirect::Policy::none())`, so a validated public URL that 3xx-redirects to a private/link-local address is no longer followed (the bearer token is never sent to the redirect target). The MCP streamable-HTTP endpoint is a direct endpoint and doesn't legitimately rely on redirects.

**DNS rebinding (TOCTOU between `validate_url_for_ssrf` and the transport's own connect):** this is the established `validate_url_for_ssrf`-then-connect pattern already used elsewhere in the codebase (e.g. `windmill-mcp/src/client_registration.rs:148`, EE SAML metadata loader), so fixing it only here would be inconsistent. Properly closing it needs a shared connect-time guard (custom resolver/connector that re-checks the dialed IP) applied across all `validate_url_for_ssrf` call sites — I'd rather do that as a separate, codebase-wide change than scope-creep this security fix. Flagging it as a tracked follow-up rather than silently leaving it.

PR #9434, opened 2026-06-03
feat: rootless podman container runtime for docker-mode jobs (replaces privileged dind)

rubenfiszel:
## Summary

Replaces the privileged **docker-in-docker (dind)** / host-socket model for `# docker` jobs with a **rootless, daemonless podman** runtime, opted in per worker group — **the user's docker script template is unchanged**.

Today, docker-mode jobs require either a privileged root dind sidecar (compose) or the host Docker socket (`exposeHostDocker`, helm). Both hand a root-equivalent, network-reachable daemon to anyone who can run a script — the worker sandbox deliberately doesn't isolate the network (`clone_newnet:false`), so it bypasses the sandbox. With rootless podman the daemon is **not root** (escape → unprivileged user) and **not network-reachable** (a user-owned unix socket), while `handle_docker_job` (bollard) talks to it unchanged.

> Builds on the interim hardening in #9425 (opt-in dind + warnings), included in this branch. Maintainers can merge #9425 first (this rebases cleanly) or take this as the complete picture and close #9425.

## What's here

- **Worker config** (`windmill-common`, `monitor`, `windmill-api-configs`): a per-worker-group `container_runtime` option (`"podman"`), loadable from group config or `CONTAINER_RUNTIME` env, allowed in CE; worker restarts on change.
- **Worker lifecycle** (`windmill-worker`): `start_container_runtime_maybe()` spawns a rootless `podman system service` once per process, waits for the socket, sets `DOCKER_HOST`. Respects an existing `DOCKER_HOST` (legacy dind) and degrades with a clear log if podman is absent.
- **Images** (`DockerfileFull`, `DockerfileFullEe`): ship podman + crun + uidmap + fuse-overlayfs + slirp4netns and a `windmill` user (uid 1000) with subuid/subgid. Base/slim untouched.
- **Frontend** (`WorkerGroup.svelte`): a "Container runtime" toggle + a warning when enabled with no tags.
- **Docs** (`docker-compose.yml`, `README.md`): reframe dind as legacy, recommend a dedicated podman worker group, with a commented `windmill_worker_docker` example.

## Validation

Verified against **system rootless podman 5.8.2** on a real host.

- **Full single-flow E2E ✅** — built a `dind` worker, set `container_runtime: podman`, and submitted the **stock docker template unchanged**:
  ```
  success: True   result: Docker job completed with success exit status
  docker mode → docker pull docker.io/library/alpine:latest → docker run --name $WM_JOB_ID
  → "docker logs stream started" → "Hello from-windmill-podman"
  ```
  Exercises the whole chain: route → `start_container_runtime_maybe` (spawns podman, sets `DOCKER_HOST`) → docker CLI ↔ podman → `handle_docker_job` (logs stream / wait / success / cleanup).
- **bollard ↔ podman** — inspect/wait/logs/stats/kill/stop/remove all work; `memory_stats.usage` confirmed non-zero (`~3140 KB`) against native rootless podman. No `handle_docker_job` change needed.
- **Config round-trip** — `container_runtime` preserved in CE updates (`dedicated_worker` correctly dropped); worker restarts on change.
- **Images** — podman layer builds on `debian:bookworm-slim` (+~103 MB; rounding error on the multi-GB full image).
- **Frontend** — toggle renders/persists, warning logic verified; `check:fast` clean.

(Details + logs in the PR comments.)

## Notes / follow-ups

- **bookworm ships podman 4.3.1**; validated against 5.8.2. The docker-compat calls are stable across 4.x/5.x, but consider bookworm-backports for a newer podman in the full images.
- **Rootless networking on bare-metal hosts:** the docker-compat default (`bridge`/netavark) can fail with `netavark: nft …` on some rootless hosts; podman `[containers] netns = "pasta"` (host-passthrough) fixes it with the stock template unchanged. In containerized workers the pod provides the netns, so this is mainly a bare-metal consideration — not yet documented in-repo pending verification of the in-container path.
- **Memory monitoring** needs cgroup v2 delegation; without it docker jobs still run but `mem_peak` stays 0. Strongest isolation = run the docker worker group as non-root on a dedicated host/node.
- **What rootless podman drops** (documented): `--privileged`, host devices, arbitrary `--cap-add`, privileged builds, some networking/storage modes. The stock template and common case are fully supported.
- **Helm:** warning for `exposeHostDocker` is in windmill-labs/windmill-helm-charts#564; first-class podman support there is a further follow-up.

- **Container lifetime under nsjail is confined (implemented).** Under nsjail a docker job runs against its own per-job rootless podman instance, torn down (`podman rm -af --force` + kill service) on every exit path — so no container it spawns, even a detached one, outlives the job (verified e2e). unshare (trusted) keeps the shared service. Tradeoff: per-job storage re-pulls images (a shared read-only image store is a follow-up optimization).


rubenfiszel:
## Update: validated against **native** rootless podman (closes the Phase 0 gaps)

Got rootless podman 5.8.2 running on the dev host **without sudo or Nix** — extracted the Arch `podman`/`conmon`/`fuse-overlayfs`/`slirp4netns` packages to a local prefix (built for this glibc) and used the system `runc` + `newuidmap` (which has `cap_setuid`). All kernel prereqs were already present (userns, cgroup v2, `/dev/fuse`).

Two gaps from the original write-up are now closed:

1. **Memory stats — now confirmed non-zero.** Re-ran the bollard 0.18.1 spike against native rootless podman:
   ```
   STATS ok: memory_stats.usage=Some(3215360) -> KB=Some(3140)
   ```
   This is the exact field `handle_docker_job` reads. The earlier `None` was purely the podman-in-docker cgroup limitation — confirmed it's nesting, not an API mismatch. inspect/logs/wait/kill/stop/remove all confirmed again too.

2. **`start_container_runtime_maybe` — confirmed in the real worker binary.** Ran the compiled worker with `CONTAINER_RUNTIME=podman`; log:
   ```
   Started rootless podman container runtime, DOCKER_HOST=unix:///tmp/windmill/podman.sock
   ```
   It spawned the rootless `podman system service`, the socket appeared in ~2s, and `DOCKER_HOST` was set.

So the full chain is verified end-to-end by composition against the **same** native rootless podman: my code spawns podman + sets `DOCKER_HOST` ✅ → docker CLI ↔ podman ✅ → `handle_docker_job`/bollard ↔ podman incl. memory ✅.

**Still not run as one monolithic flow:** submitting a literal `# docker` job through a worker and watching it complete — the dev binary is built `--features quickjs` (no `dind`, so `handle_docker_job` isn't compiled in it), and a dind rebuild would collide with the running cargo-watch target dir or need a from-scratch build. Each link is individually verified against real native podman.

rubenfiszel:
## ✅ Full single-flow E2E now passes (real `# docker` job → Windmill → rootless podman)

Built a `dind` worker (`cargo build --bin windmill --features quickjs,dind`, 27s incremental), ran it with `CONTAINER_RUNTIME=podman` against **system rootless podman 5.8.2**, and submitted the **stock docker template** unchanged via `/jobs/run/preview` (tag-routed to the podman worker):

```
success: True
result: Docker job completed with success exit status
--- logs ---
docker mode
... docker.io/library/alpine:latest        # docker pull  -> podman
5d6f7728d9ff...                             # docker run --name $WM_JOB_ID -it -d
docker logs stream started                  # handle_docker_job (bollard)
Hello from-windmill-podman                  # container stdout streamed back; arg passed through
```

The whole chain is exercised in one run:
1. Windmill routes the `# docker` job to the podman worker group.
2. `start_container_runtime_maybe` spawns the rootless `podman system service` and sets `DOCKER_HOST` (`Started rootless podman container runtime, DOCKER_HOST=unix:///tmp/windmill/podman.sock`).
3. The bash script's `docker` CLI pulls + runs the container against podman.
4. `handle_docker_job` (bollard) streams logs, waits for exit 0, reports success, removes the container.

### One environment note worth surfacing for the docs

On a **bare-metal rootless** host, the docker-compat default (`bridge` via netavark) failed here with `netavark: nft … did not return successfully` — a rootless-nft quirk on this host (the host's own Docker uses nft fine). Setting podman `[containers] netns = "pasta"` makes the docker-compat path use pasta (host-passthrough) and the **stock template runs unchanged**. In containerized workers (compose/k8s) the pod already provides the netns, so this is mainly a bare-metal-rootless consideration — I'll add a note to the compose/docs. (Memory stats were separately confirmed non-zero against native podman; this echo container exits too fast to register.)

rubenfiszel:
## Follow-up design: confine containers to the job lifetime under nsjail

**Caveat for this PR (shared-service runtime):** containers are managed by a shared per-worker podman service and cleaned up by name (`$WM_JOB_ID`). That means a job can spawn containers that **outlive the job** (extra/unnamed/detached containers aren't tracked; the shared service is decoupled from job lifetime). So **`container_runtime` groups are trusted-only** for now — don't run untrusted / nsjail-sandboxed workloads on them. Confirmed empirically that a sandboxed script can't *directly* run podman (nsjail's single-uid userns blocks rootless setup) and can't reach the service socket (`/tmp/windmill` isn't mounted into the jail), but the lifetime guarantee is not yet enforced.

**Agreed follow-up (separate PR):** make podman usable under **both** unshare and nsjail, but under nsjail guarantee that **no container outlives the job**.

Why naive killing is insufficient: `conmon` double-forks/reparents so `-d` containers survive the launcher; process-tree / PID-namespace death doesn't kill them. The robust primitive is **cgroup v2 `cgroup.kill`**.

Design:
- **unshare (trusted):** keep the shared service (efficient, warm image cache). No lifetime enforcement required.
- **nsjail (untrusted):** a **per-job** rootless `podman system service` started **inside the job's cgroup** (shared image store for cache reuse); `handle_docker_job` talks to its per-job socket (existing bollard code, per-job `DOCKER_HOST`). podman `cgroup_manager = "cgroupfs"` + `--cgroup-parent = <job cgroup>` so containers stay under the job cgroup (systemd manager would migrate them to a fresh scope and escape). On job completion / cancel / timeout / killpill, write `1` to `<job cgroup>/cgroup.kill` → SIGKILLs the per-job service, conmon, and **all** containers atomically, regardless of name/count/detach.
- Net-new machinery required: a per-job cgroup (nsjail supports cgroup placement; the bash config currently sets none) + cgroup-kill teardown (the worker currently kills by process, no cgroup-kill).

This generalizes cleanup from "remove `$WM_JOB_ID`" to "kill the job cgroup", which is strictly more complete.

rubenfiszel:
## ✅ nsjail container-lifetime confinement implemented + verified

Followed up on the agreed design: under **nsjail** a docker job now runs against its **own rootless podman instance** (isolated `--root`/`--runroot` under the job dir; socket bind-mounted into the jail and handed to the script as `DOCKER_HOST`). On every exit path the `PerJobPodman` guard tears it down — `podman rm -af --force` (removes **all** containers in the instance, since `conmon` reparenting makes process-tree kill insufficient) + kills the service + removes the dir. **unshare** (trusted) keeps the shared service unchanged.

E2E on native rootless podman + nsjail — a job that detaches an *extra* `sleep 600` container alongside the managed one:
```
isolation=nsjail ... docker mode ... docker logs stream started ... hello-from-nsjail-docker
success: True   result: Docker job completed with success exit status
during job:  extra_sleep600_running = YES
after job:   extra container gone ✓   (no leak)
```
So nothing the job spawns — even an explicitly detached container — outlives the job under nsjail.

Tradeoff: per-job storage re-pulls images (no shared cache yet); a shared read-only image store is a follow-up optimization. Commit: 8483700f0e.

rubenfiszel:
## Shared image cache investigation — and a teardown bug it surfaced

Tried to add a shared read-only image cache (`additionalimagestores`) so per-job (nsjail) instances don't re-pull. Result: **it can't avoid re-pulls cleanly with robust per-job isolation**, on podman 5.8.2:

- `additionalimagestores` lets a per-job *run* a cached image without re-pull, **but** an explicit `docker pull ` (which the stock template does) still re-downloads unless the cache was populated by a **direct registry pull**.
- Populating the cache by promoting a per-job's pulled image **does not** restore pull-dedup: `podman save|load` (docker- *and* oci-archive) breaks blob-level dedup, and even `skopeo copy` (containers-storage→containers-storage) does too. The RW image-store split (`storage.options.imagestore`) didn't redirect layer writes either (layers stayed in graphroot).
- The only way `docker pull` dedups is pulling **directly** into the shared store — which requires sharing the graphroot, which breaks the per-job container-isolation teardown guarantee.

So a true "pull-once" cache needs a different design (e.g. a per-worker-process shared store with snapshot-diff teardown — weaker/raceable — or a registry **pull-through cache** as separate infra). I did **not** ship a half-working cache; per-job stores re-pull, which is documented.

### What this *did* yield: a real teardown fix (commit 9a64e74aa2)

The investigation exposed a storage leak: per-job teardown used `podman rm -af` (containers only) + `remove_dir_all`, but rootless image layers are **subuid-owned**, so any pulled image leaked its overlay layers under the job dir (and blocked job-dir cleanup). Fixed by tearing down with **`podman system reset --force`**, which removes containers + images and deletes the subuid layers from inside the userns. Verified e2e: a job pulling an image now leaves no subuid files / no per-job podman dir; confinement unchanged.

**Follow-up (re-pull avoidance):** recommend a registry pull-through cache (or revisit when podman's writable image-store split works), tracked separately.

rubenfiszel:
## Refactor: per-job podman for ALL docker jobs; shared service removed (commit f61cf1633b)

Generalized the per-job rootless-podman model to every sandbox mode and dropped the long-lived shared service:

- **Removed** `start_container_runtime_maybe()` (the worker-startup shared `podman system service`) + the global `DOCKER_HOST` mutation. Each docker job now spins up its **own ephemeral per-job podman** and tears it down at job end (no container outlives any job; no standing shared socket on the worker).
- **Gated strictly on `container_runtime=podman`.** The legacy path — an externally-provided `DOCKER_HOST` or a mounted `/var/run/docker.sock`, *without* `container_runtime` — is **unchanged**: `connect_docker(None)` / forwarded env, and only the `$WM_JOB_ID` container is managed, exactly as before.
- Script `DOCKER_HOST`: in-jail `/tmp/podman.sock` (bind-mounted) under nsjail; the per-job host socket directly under unshare/none.
- Dropped the `container_runtime` killpill in monitor (no startup state → config read live per job).

The `container_runtime` worker-group config **stays** — it's now the per-job gate that selects podman-per-job vs the legacy docker socket (and backs the UI toggle / CE allowlist).

### Verified e2e (rootless podman)
| mode | result |
|---|---|
| nsjail + podman | per-job; extra detached container confined ✓ |
| **unshare + podman** | per-job (`isolation=unshare`); extra container seen mid-job, gone after ✓ |
| **legacy (no container_runtime)** | runs against host `/var/run/docker.sock`; **no** per-job podman; extra container **persists** (unchanged) ✓ |

rubenfiszel:
## Simplified: dropped the `container_runtime` config — runtime is now auto-selected (commit cccd8b9847)

Per-worker-group `container_runtime` is removed entirely (config field, CE allowlist, monitor, **and the frontend toggle**). The docker runtime is now decided automatically per `# docker` job:

- **Docker daemon provided** (`DOCKER_HOST` set, or `/var/run/docker.sock` mounted) → use it. Backwards compatible, unchanged.
- **Otherwise** → start a per-job rootless **podman** (its own ephemeral daemon, torn down with the job). Requires podman in the image (`*-full`); else a clear error.

So: "provided Docker socket wins (legacy); otherwise podman." No config to set. The `windmill_worker_docker` compose example no longer needs `CONTAINER_RUNTIME` — just a `*-full` image + privileged + `/dev/fuse` + a tag.

Verified e2e: a worker with **no** `CONTAINER_RUNTIME` and **no** `DOCKER_HOST` auto-ran a `# docker` job via per-job podman (`success`, logs streamed); the provided-daemon legacy path is unchanged.

PR #9449, opened 2026-06-04
oauth: add ServiceNow + make per-instance OAuth providers registry-driven

hugocasa:
Adds **ServiceNow** as a first-class OAuth provider, and in doing so **generalizes the per-instance OAuth mechanism** so future providers like it need only a JSON registry entry — no frontend code. Pairs with the hub integration in [windmill-integrations#139](https://github.com/windmill-labs/windmill-integrations/pull/139).

## Background
ServiceNow's OAuth endpoints are per-instance (`https://<instance>.service-now.com/oauth_auth.do` / `/oauth_token.do`), like Snowflake's. Today Snowflake is wired up with a **bespoke special-case** (a dedicated `snowflakeAccountIdentifier` state var, a `{#if k == 'snowflake_oauth'}` input block, and a `setupSnowflakeUrls()` function hardcoding its URLs). Adding ServiceNow the same way would mean a second copy of all that.

## The generalization
A registry entry may now carry a **`connect_config_template`**:

```json
"servicenow": {
  "connect_config_template": {
    "label": "ServiceNow Instance",
    "placeholder": "<instance> (e.g. dev12345)",
    "help_url": "https://www.servicenow.com/docs/.../c_OAuthApplications.html",
    "auth_url": "https://{instance}.service-now.com/oauth_auth.do",
    "token_url": "https://{instance}.service-now.com/oauth_token.do",
    "req_body_auth": true,
    "strip_suffix": ".service-now.com"
  }
}
```

The instance-settings UI derives per-instance providers from the registry, renders **one generic instance-name input** for any of them, and substitutes `{instance}` into the templated URLs to build the per-client `connect_config`. So **adding a new per-instance OAuth provider is now just a JSON entry** — no Svelte changes.

## Changes
- **`backend/oauth_connect.json`** — `servicenow` and `snowflake_oauth` now carry a `connect_config_template`. Snowflake keeps its `account_identifier` `extra_params` key (`extra_params_key`) so previously-saved configs round-trip unchanged.
- **`backend/windmill-oauth/src/lib.rs`** — adds the `ConnectConfigTemplate` struct (frontend-only metadata; `#[serde(skip_serializing_if)]`). The backend never reads it — the existing per-client `connect_config` override in `build_oauth_clients` resolves the concrete URLs generically.
- **`AuthSettings.svelte` / `InstanceSettings.svelte`** — replaces the Snowflake **and** ServiceNow special-cases with one registry-driven path: a `connectConfigTemplates` map, `windmillBuiltinsTemplated` (derived from the registry, like the existing sandbox derivation), an `instanceInputs` map (instance name keyed by provider), and `setupTemplatedOauthUrls()` / `loadInstanceInputs()`. Net **−71 / +159** but removes both bespoke flows.

The admin enters their instance + Client ID/Secret (ServiceNow: System OAuth → Application Registry, redirect `https://<windmill>/oauth/callback/servicenow`). End users get one-click connect + auto-refresh; the `servicenow` resource carries `token` and `instance_url` is filled on the resource (ServiceNow's token response omits the host).

## Validation
- `cargo check -p windmill-oauth` passes (`SQLX_OFFLINE=true`).
- `oauth_connect.json` valid JSON; both Svelte files pass `prettier`; no leftover references to the removed Snowflake/ServiceNow symbols.
- **Not** exercised against a live admin OAuth flow (needs a running Windmill + registered OAuth apps). Snowflake was migrated to the shared path with its `extra_params` key preserved for backward-compat — worth a careful look in review/CI.

### DIFF
--- a/backend/ee-repo-ref.txt
+++ b/backend/ee-repo-ref.txt
@@ -1,1 +1,1 @@
-3742e0659c5e97aab03b9efeea14cd94a3ac658a
+servicenow-oauth
--- a/backend/oauth_connect.json
+++ b/backend/oauth_connect.json
@@ -154,29 +154,34 @@
   "zoho": {
     "auth_url": "https://accounts.zoho.com/oauth/v2/auth",
     "token_url": "https://accounts.zoho.com/oauth/v2/token",
-    "scopes": [
-      "ZohoAssist.sessionapi.ALL"
-    ],
+    "scopes": ["ZohoAssist.sessionapi.ALL"],
     "extra_params": {
       "access_type": "offline"
     }
   },
-  "snowflake_oauth": {},
+  "snowflake_oauth": {
+    "connect_config_template": {
+      "display_name": "Snowflake",
+      "label": "Snowflake Account Identifier",
+      "placeholder": "<orgname>-<account_name>",
+      "help_url": "https://docs.snowflake.com/en/user-guide/admin-account-identifier#using-an-account-name-as-an-identifier",
+      "auth_url": "https://{instance}.snowflakecomputing.com/oauth/authorize",
+      "token_url": "https://{instance}.snowflakecomputing.com/oauth/token-request",
+      "req_body_auth": false,
+      "extra_params_key": "account_identifier",
+      "resource_mapping": { "account_identifier": "{instance}" }
+    }
+  },
   "apify": {
     "auth_url": "https://console.apify.com/authorize/oauth",
     "token_url": "https://console-backend.apify.com/oauth/apps/token",
-    "scopes": [
-      "profile",
-      "full_api_access"
-    ],
+    "scopes": ["profile", "full_api_access"],
     "extra_params": {}
   },
   "docusign": {
     "auth_url": "https://account.docusign.com/oauth/auth",
     "token_url": "https://account.docusign.com/oauth/token",
-    "scopes": [
-      "signature"
-    ],
+    "scopes": ["signature"],
     "sandbox": {
       "auth_url": "https://account-d.docusign.com/oauth/auth",
       "token_url": "https://account-d.docusign.com/oauth/token"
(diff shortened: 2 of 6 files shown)


hugocasa:
Companion EE build fix: **windmill-labs/windmill-ee-private#602** — adds `connect_config_template: None` to the slack `OAuthConfig` literal in `oauth2_ee.rs` (EE code, symlinked), required because the new field has no `Default`. Verified `cargo check -p windmill-api --features oauth2` passes with both applied.

hugocasa:
Pushed fixes for the review findings:

- **npm_check / svelte-check error** (`AuthSettings.svelte:529` *'instanceInputs' is possibly undefined*): made `instanceInputs` a **required** prop (it's always bound by `InstanceSettings`) rather than using `$bindable({})` — avoiding the banned `$bindable(default)` pattern @claude flagged. svelte-check now reports no errors in the changed files.
- **Codex P1 — missing `OAuthConfig` field initializers**: both literals now set `connect_config_template: None` — the OSS test helper (`sample_oauth_config`) in this PR, and the EE Slack literal (`oauth2_ee.rs`) in the companion **windmill-ee-private#602**. `cargo check -p windmill-api --features oauth2` + `-p windmill-oauth --tests` pass.
- **Codex P1 — ServiceNow resource missing `instance_url`**: generalized the Snowflake-only resource copy in `AppConnectInner.svelte` into a template-driven `resource_mapping`. The connected resource now gets `instance_url = https://{instance}.service-now.com` (and Snowflake keeps `account_identifier`), so the resource is usable for instance-specific API calls. Removes the Snowflake special-case too.
- **cubic P3 / @claude P2 — dropdown display name**: added an optional `display_name` to the template (`"ServiceNow"` / `"Snowflake"`); the dropdown falls back to `capitalize(name)` when absent.

Left as-is (both noted as non-regressions): the small duplicate `connectConfigTemplates` derivation, and the empty-instance guard (the input is `required` and this matches prior Snowflake behavior).

Live admin-flow verification (Snowflake regression + ServiceNow end-to-end connect) still needs a maintainer with a running instance, as noted.

PR #9452, opened 2026-06-04
fix: distinguish canceled jobs in runs

ryan-gray-db:
## Summary
Fix the Runs page so completed jobs that were canceled are distinguishable from failed jobs.

Windmill already stores canceled completed jobs as `v2_job_completed.status = 'canceled'` and returns `canceled: true` on completed job records, but the Runs UI used the legacy `success=false` filter for Failure. That grouped canceled rows with failed rows and rendered the status icon as a red failure.

## Changes
- Add a completed-job `status` query parameter for exact completed statuses: `success`, `failure`, `canceled`, and `skipped`.
- Keep the existing `success=true|false` behavior for backwards compatibility.
- Use exact `status=failure` and `status=canceled` from the Runs UI filters.
- Add Canceled to the Runs status picker and render canceled rows/favicon state separately from failures.
- Add a focused backend regression test for `status=canceled` filtering.

## Test plan
- [x] `DATABASE_URL='postgres://postgres:changeme@localhost:5433/windmill?sslmode=disable' sqlx migrate run`
- [x] `DATABASE_URL='postgres://postgres:changeme@localhost:5433/windmill?sslmode=disable' cargo test -p windmill-api-jobs test_completed_filter_status_canceled`
- [x] `DATABASE_URL='postgres://postgres:changeme@localhost:5433/windmill?sslmode=disable' cargo check -p windmill-api`
- [x] `cd frontend && npm run generate-backend-client`
- [x] `cd frontend && npm run check` (0 errors; existing warnings only)

I did not run the local frontend in a browser because this checkout was used for patch validation rather than a full local Windmill dev stack.

---
## Summary by cubic
Runs now clearly distinguish canceled jobs from failures on the Runs page. Added exact completed `status` filtering and optimized ordering; the UI and favicon show canceled in gray only after the job completes.

- **Bug Fixes**
  - Backend: added `status` query param for completed jobs (`success`, `failure`, `canceled`, `skipped`); ordered `status=failure|canceled` by `completed_at` to use the partial index; `success=true|false` still supported.
  - API: disallow combining `success`/`status` with `running`; when `status` is set, do not union queued jobs; OpenAPI documents `status` as an exact match (`status=success` excludes skipped); added regression tests for `status=canceled` and order-by on `failure|canceled`.
  - UI: Runs filter adds “Canceled” and sends `status=failure|canceled`; canceled rows use a gray hourglass and “Canceled”; icon and favicon turn gray for canceled only after completion (canceling jobs keep the running state).

<sup>Written for commit 32f067b5366d531279eaa13c0e8ec30315c967ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-jobs/src/concurrency_groups.rs
+++ b/backend/windmill-api-jobs/src/concurrency_groups.rs
@@ -191,6 +191,7 @@ async fn get_concurrent_intervals(
             script_path_exact: None,
             script_hash: None,
             created_by: None,
+            status: None,
             success: None,
             running: None,
             parent_job: None,
--- a/backend/windmill-api-jobs/src/query.rs
+++ b/backend/windmill-api-jobs/src/query.rs
@@ -973,8 +973,8 @@ mod tests {
     #[test]
     fn test_completed_order_by_completed_at_status_failure_canceled() {
         // status=failure|canceled must order by v2_job_completed.completed_at so the
-        // partial failure index (ix_v2_job_completed_failure_workspace) serves both
-        // filtering and ordering in a single scan.
+        // partial index ix_v2_job_completed_failure_workspace serves both filtering
+        // and ordering in a single scan.
         for status in [
             windmill_common::jobs::JobStatus::Failure,
             windmill_common::jobs::JobStatus::Canceled,
@@ -985,20 +985,6 @@ mod tests {
             assert!(
                 sql.contains("ORDER BY v2_job_completed.completed_at"),
                 "expected order by completed_at, got: {sql}"
-            );
-        }
-
-        // status=success|skipped have no partial index, keep the default created_at order.
-        for status in [
-            windmill_common::jobs::JobStatus::Success,
-            windmill_common::jobs::JobStatus::Skipped,
-        ] {
-            let lq = ListCompletedQuery { status: Some(status), ..empty_completed_query() };
-            let sqlb = list_completed_jobs_query("ws", Some(10), 0, &lq, &["id"], false, None);
-            let sql = build_sql(sqlb);
-            assert!(
-                sql.contains("ORDER BY v2_job.created_at"),
-                "expected order by created_at, got: {sql}"
             );
         }
     }
--- a/backend/windmill-api-jobs/src/query.rs
+++ b/backend/windmill-api-jobs/src/query.rs
@@ -580,6 +580,11 @@ pub fn list_completed_jobs_query(
             if lq.completed_before.is_some()
                 || lq.completed_after.is_some()
                 || lq.success == Some(false)
+                || matches!(
+                    lq.status,
+                    Some(windmill_common::jobs::JobStatus::Failure)
+                        | Some(windmill_common::jobs::JobStatus::Canceled)
+                )
             {
                 "v2_job_completed.completed_at"
             } else {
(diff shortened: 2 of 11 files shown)


ryan-gray-db:

rubenfiszel:
Pushed a follow-up fix for an index-utilization regression introduced by switching the Runs "Failure" filter from `success=false` to the new `status` param.

### The issue

`list_completed_jobs_query` flips the ORDER BY to `v2_job_completed.completed_at DESC` only when `completed_before`/`completed_after` are set or `success == Some(false)`. The partial index added for this page —

```sql
CREATE INDEX ix_v2_job_completed_failure_workspace
    ON v2_job_completed (workspace_id, completed_at DESC)
    WHERE status IN ('failure', 'canceled');
```

— only serves both the filter **and** the ordering when the query orders by `completed_at`. Its own migration comment says as much.

Once the UI routes Failure/Canceled through `status=` and sends `success=undefined`, `lq.success` is `None`, so (absent a `completed_*` cursor — e.g. the initial Runs load) the ORDER BY falls back to `v2_job.created_at`. The planner then walks the `created_at` index newest-first and discards ~99% of rows via the join filter instead of using the partial index. Canceled is brand-new and never had the old `success=false` path at all.

### Evidence (500k completed rows, 1% failure/canceled)

`status=failure`, `LIMIT 30`:

| ORDER BY | Plan | Buffers | Time |
|---|---|---|---|
| `completed_at DESC` (fixed) | `Index Scan using ix_v2_job_completed_failure_workspace` | ~150 | **0.31 ms** |
| `v2_job.created_at DESC` (pre-fix) | Parallel scan of `ix_job_workspace_id_created_at_new_3`, probe+discard via join | ~49,300 | **31.5 ms** |

~100× slower / ~320× more buffers, and it degrades further when failures are sparser or older than the most recent jobs — exactly the scenario the partial index was built for.

### The fix

Extend the ORDER BY switch to also flip to `completed_at` for `status=failure|canceled` (the two statuses the partial index covers; `success`/`skipped` keep `created_at`). Added a regression test asserting the ordering for each status.

rubenfiszel:
Follow-up changes pushed:

1. **Running-while-canceling icon/favicon** — `JobStatusIcon` and `getJobStatusKind` now gate the canceled state on `'success' in job`. Without this, moving the `job.canceled` check to the top of the chain also changed in-flight cancellations: a job that is still running while being canceled would immediately render the gray "Canceled" hourglass instead of its running icon. With the guard it keeps the running state until it actually completes. Both files stay mirror-consistent (per the `getJobStatusKind` comment).

2. **OpenAPI wording** — clarified the new `status` param is an exact match: `status=success` matches only `success`, unlike `success=true` which also matches `skipped`.

3. **Test trim** — reduced the order-by regression test to the failure/canceled case only.

Verified end-to-end that `status` flows through to the request: the gitignored generated client (`src/lib/gen`) picks up the param from `openapi.yaml` on `npm run generate-backend-client`, and the query lands on the partial index — see the earlier comment for the EXPLAIN ANALYZE.

Non-blocking nits left for your call: the `match` in `query.rs` duplicates `JobStatus`'s `rename_all = "lowercase"`; the new branch qualifies `v2_job_completed.status` while the legacy `success` branch uses unqualified `status`; and the API accepts `success` + `status` together (status silently wins).

PR #9453, opened 2026-06-05
feat: sandboxed daemonless container runtime via '# sandbox <image>'

rubenfiszel:
## Summary

Adds a **sandboxed, daemonless container runtime** for bash scripts, selected by `# sandbox <image>`. The image's rootfs is extracted with rootless **podman** and the script body runs **chrooted inside the job's own nsjail** — so the container inherits exactly the job's confinement (no host `/`, jail's own `/proc`, unprivileged worker uid, the job's network) and is safe to run under nsjail / for untrusted multi-tenant code. The legacy `# docker` (dind/daemon) path is **unchanged**.

The three bash annotations are now cleanly distinct (no overloading):
- `# docker` → legacy daemon docker (v1)
- `# sandbox` → run the bash script under nsjail
- `# sandbox <image>` → run that image's command under nsjail (this runtime)

The body runs via the image's `/bin/sh`; an empty body runs the image's ENTRYPOINT/CMD; windmill args bind positionally; the image's Env/WorkingDir are applied and `WM_TOKEN`/`BASE_INTERNAL_URL` are injected so `wmill`/API calls work.

```mermaid
flowchart TD
    A["bash job"] --> B{annotation}
    B -- "# docker" --> V1["v1 (dind daemon)"]
    B -- "# sandbox" --> S["nsjail bash"]
    B -- "# sandbox IMAGE" --> N{nsjail?}
    N -- no --> ERR["error: needs nsjail"]
    N -- yes --> P["pull/extract (podman rootless)<br/>create --pull=POLICY [--authfile] -- IMAGE<br/>export | tar -x → rootfs; inspect → OCI config"]
    P --> R["run (job's nsjail)<br/>per-top-dir rootfs binds, uid0→worker<br/>tmpfs /tmp (instance setting), /proc, {SHARED_MOUNT}"]
    R --> H["handle_child = the job's confinement"]
```

## Changes

**Runtime** (`backend/windmill-worker/`)
- `docker_v2.rs` (new): image extract (podman create/export/inspect), dynamic per-top-dir nsjail mounts (incl. usr-merge symlinks), run via nsjail + `handle_child`.
- `nsjail/run.docker.config.proto` (new): v2 profile — uid 0→worker, `{TMP_MOUNT_BLOCK}` (honors `nsjail_tmp_backing`/`nsjail_tmpfs_size_mb`), `{ROOTFS_MOUNTS}`, `{SHARED_MOUNT}` for `# volume` mounts + same-worker `/tmp/shared`.
- `windmill-common/src/worker.rs`: `BashAnnotations::sandbox_image()` parser (+ unit test).
- `bash_executor.rs`: branch to v2 when an image is declared; `raw_to_string` deduped into `common`.

**Image storage / freshness / limits — instance settings** (DB `global_settings` + worker `RwLock` + `monitor` reload + superadmin UI, hot-reloaded like `nsjail_tmpfs_size_mb`)
- `sandbox_image_pull_policy` (default `newer` — moving tags don't go stale)
- `sandbox_image_max_size_mb` (reject oversized images before extraction)
- `sandbox_image_cache_max_mb` (best-effort LRU eviction of podman's image store)
- `sandbox_image_default_registry` (prepended to unqualified refs; fully-qualified untouched)
- `sandbox_registry_auth` (docker/podman `auth.json` blob → per-job 0600 authfile → `podman --authfile`)

**Surface / ergonomics**
- `docker-compose.yml`: dropped the `dind` sidecar (daemonless now); host-socket opt-in for legacy docker documented inline.
- Frontend: removed the `isCloudHosted()` guard blocking Docker scripts on the multi-tenant platform; the editor's Docker template now uses `# sandbox <image>`.
- `docs/docker-v2-runtime.md` (new): surface, security model, storage/limits, registries, limitations.

### Editor "Docker" template — before / after

Picking **Docker** in the script editor now scaffolds the v2 surface. The body *is* the program that runs inside the image — no daemon dance, no `$WM_JOB_ID` bookkeeping.

**Before** (v1 — script drives the docker CLI against a daemon):
```bash
# shellcheck shell=bash
# docker
# The annotation "docker" above is important, it tells windmill that after
# the end of the bash script, it should manage the container at id $WM_JOB_ID:
# pipe logs, monitor memory usage, kill container if job is cancelled.

msg="${1:-world}"

IMAGE="alpine:latest"
COMMAND="/bin/echo Hello $msg"

# ensure that the image is up-to-date
docker pull $IMAGE

# if using the 'docker' mode, name it with $WM_JOB_ID for windmill to monitor it
docker run --name $WM_JOB_ID -it -d $IMAGE $COMMAND
```

**After** (v2 — the body runs *inside* the image, sandboxed):
```bash
# shellcheck shell=bash
# sandbox alpine:latest
# The "# sandbox <image>" annotation runs this script INSIDE the image above,
# sandboxed via nsjail: the image's rootfs is extracted (rootless podman) and the
# body runs chrooted in it, inheriting the job's confinement. The body runs with
# the image's /bin/sh and windmill args bind positionally as $1, $2, ...
# Daemonless — no docker run/-d/exec/build and no host -v bind mounts.
# (A bare "# docker" still uses the legacy daemon runtime instead.)

msg="${1:-world}"

echo "Hello $msg"
cat /etc/os-release | head -1
```

## Security

Reviewed (fresh-context `local-review`): instance-settings wiring complete for all five settings; the sandbox-injection boundary holds. Image-controlled values (WorkingDir, mount srcs, symlink targets) are escaped via a protobuf-safe `proto_str` (octal escapes for control/non-ASCII bytes); `podman create -- <image>` guards a `-`-prefixed ref; the registry authfile is created `0600` atomically and lives in the per-job dir (not bound into the container). Confirmed live: host worker secrets (`DATABASE_URL` etc.) are **not** visible from inside the container; `/proc/1` is the jail's own init.

## Limitations (by design — daemonless, run-to-completion)
No `-d`/`exec`/`build`/`compose`, no arbitrary `-v`/`--privileged`/`--cap-add`/host-ns; multi-uid-in-container needs a subuid-range follow-up; result is a completion message (output via stdout/logs). Requires nsjail (clear error otherwise). Each job re-exports the rootfs (podman's image store still dedups the network pull) — content-addressed rootfs cache is a follow-up.

## Test plan

Verified end-to-end on a worker with rootless podman + nsjail (workspace `admins`):
- [x] `# sandbox alpine` / `python:3.12-slim` (debian usr-merge) → body runs inside the image, uid 0, args bind positionally
- [x] empty body → image ENTRYPOINT/CMD; `exit N` → job fails with that code
- [x] `# docker` (bare) → legacy v1 path untouched; `# docker <image>` → plain bash (not v2)
- [x] confinement: host FS invisible, jail's own 3-pid `/proc`, no `DATABASE_URL` in any visible `/proc/*/environ`
- [x] `/tmp` honors `nsjail_tmpfs_size_mb`; `# volume` mounts apply via `{SHARED_MOUNT}`
- [x] instance settings hot-reload (no restart): `sandbox_image_max_size_mb=1` → 8 MB alpine rejected; `sandbox_image_default_registry=docker.io` → `# sandbox alpine` resolves to `docker.io/alpine`; `sandbox_registry_auth` → authfile written + `--authfile` passed
- [x] job-dir cleanup removes `rootfs/` + transient `rootfs.tar`
- [x] unit tests: `sandbox_image`, `proto_str` (incl. non-ASCII), `registry_qualified`; `cargo check` + frontend `check:fast` clean
- [ ] Reviewer: private-registry pull with real credentials; worker without nsjail → clear error

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -57,8 +57,9 @@ use windmill_common::{
         PREVIEW_TAGS_OVERRIDE_SETTING, REQUEST_SIZE_LIMIT_SETTING,
         REQUIRE_PREEXISTING_USER_FOR_OAUTH_SETTING, RESTART_COORDINATION_SETTING,
         RETENTION_PERIOD_SECS_SETTING, RUBY_REPOS_SETTING, SAML_METADATA_SETTING,
-        SANDBOX_IMAGE_CACHE_MAX_MB_SETTING, SANDBOX_IMAGE_MAX_SIZE_MB_SETTING,
-        SANDBOX_IMAGE_PULL_POLICY_SETTING, SCIM_TOKEN_SETTING, SMTP_SETTING,
+        SANDBOX_IMAGE_CACHE_MAX_MB_SETTING, SANDBOX_IMAGE_DEFAULT_REGISTRY_SETTING,
+        SANDBOX_IMAGE_MAX_SIZE_MB_SETTING, SANDBOX_IMAGE_PULL_POLICY_SETTING,
+        SANDBOX_REGISTRY_AUTH_SETTING, SCIM_TOKEN_SETTING, SMTP_SETTING,
         STORE_AUDIT_LOGS_S3_SETTING, TEAMS_SETTING, TIMEOUT_WAIT_RESULT_SETTING,
         UV_EXCLUDE_NEWER_SETTING, UV_INDEX_STRATEGY_SETTING, UV_PYTHON_INSTALL_MIRROR_SETTING,
         WORKSPACE_FAIRNESS_DURATION_SECS_SETTING, WORKSPACE_FAIRNESS_ENABLED_SETTING,
@@ -137,8 +138,9 @@ use crate::monitor::{
     reload_license_key, reload_npm_config_registry_setting, reload_nsjail_tmp_backing_setting,
     reload_nsjail_tmpfs_size_setting, reload_otel_tracing_proxy_setting,
     reload_pip_index_url_setting, reload_retention_period_setting,
-    reload_sandbox_image_cache_max_setting, reload_sandbox_image_max_size_setting,
-    reload_sandbox_image_pull_policy_setting, reload_scim_token_setting, reload_smtp_config,
+    reload_sandbox_image_cache_max_setting, reload_sandbox_image_default_registry_setting,
+    reload_sandbox_image_max_size_setting, reload_sandbox_image_pull_policy_setting,
+    reload_sandbox_registry_auth_setting, reload_scim_token_setting, reload_smtp_config,
     reload_store_audit_logs_s3_setting, reload_uv_exclude_newer_setting,
     reload_uv_index_strategy_setting, reload_uv_python_install_mirror_setting,
     reload_worker_config, MonitorIteration,
@@ -1840,6 +1842,10 @@ async fn process_notify_event(
                 SANDBOX_IMAGE_PULL_POLICY_SETTING => {
                     reload_sandbox_image_pull_policy_setting(conn).await
                 }
+                SANDBOX_IMAGE_DEFAULT_REGISTRY_SETTING => {
+                    reload_sandbox_image_default_registry_setting(conn).await
+                }
+                SANDBOX_REGISTRY_AUTH_SETTING => reload_sandbox_registry_auth_setting(conn).await,
                 #[cfg(feature = "parquet")]
                 OBJECT_STORE_CONFIG_SETTING => {
                     if !disable_s3_store {
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -57,11 +57,13 @@ use windmill_common::{
         PREVIEW_TAGS_OVERRIDE_SETTING, REQUEST_SIZE_LIMIT_SETTING,
         REQUIRE_PREEXISTING_USER_FOR_OAUTH_SETTING, RESTART_COORDINATION_SETTING,
         RETENTION_PERIOD_SECS_SETTING, RUBY_REPOS_SETTING, SAML_METADATA_SETTING,
-        SCIM_TOKEN_SETTING, SMTP_SETTING, STORE_AUDIT_LOGS_S3_SETTING, TEAMS_SETTING,
-        TIMEOUT_WAIT_RESULT_SETTING, UV_EXCLUDE_NEWER_SETTING, UV_INDEX_STRATEGY_SETTING,
-        UV_PYTHON_INSTALL_MIRROR_SETTING, WORKSPACE_FAIRNESS_DURATION_SECS_SETTING,
-        WORKSPACE_FAIRNESS_ENABLED_SETTING, WORKSPACE_FAIRNESS_MAX_PERCENT_SETTING,
-        WORKSPACE_FAIRNESS_MIN_TOTAL_SETTING, WORKSPACE_REGISTRIES_SETTING,
+        SANDBOX_IMAGE_CACHE_MAX_MB_SETTING, SANDBOX_IMAGE_MAX_SIZE_MB_SETTING,
+        SANDBOX_IMAGE_PULL_POLICY_SETTING, SCIM_TOKEN_SETTING, SMTP_SETTING,
+        STORE_AUDIT_LOGS_S3_SETTING, TEAMS_SETTING, TIMEOUT_WAIT_RESULT_SETTING,
+        UV_EXCLUDE_NEWER_SETTING, UV_INDEX_STRATEGY_SETTING, UV_PYTHON_INSTALL_MIRROR_SETTING,
+        WORKSPACE_FAIRNESS_DURATION_SECS_SETTING, WORKSPACE_FAIRNESS_ENABLED_SETTING,
+        WORKSPACE_FAIRNESS_MAX_PERCENT_SETTING, WORKSPACE_FAIRNESS_MIN_TOTAL_SETTING,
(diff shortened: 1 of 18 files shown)


rubenfiszel:
Verified job-dir cleanup live (workspace `admins`, sleeping v2 job):

- **Mid-run** `/tmp/windmill/<worker>/<job_id>/` contains `rootfs/` (extracted alpine tree) + `run.docker.config.proto`, and **`rootfs.tar` is already absent** — removed right after extraction, so it doesn't double disk for the run's duration.
- **Post-run** the entire job dir (incl. `rootfs/`) is removed by windmill's standard `remove_dir_all(job_dir)` (worker.rs:3118) — no leftover anywhere; parent worker dir untouched.

Remaining unchecked item (worker **without** nsjail → clear "requires nsjail" error) isn't exercisable on this box since `NSJAIL_AVAILABLE` is a process-level static and this worker has nsjail; it's a simple guard at the top of `handle_docker_v2_job`.

rubenfiszel:
### Update — surface moved to \`# sandbox <image>\`, dind dropped, size/freshness controls added

Since the original description, based on review feedback:

**Surface** (was \`# docker <image>\`): now selected by **\`# sandbox <image>\`**. The three bash annotations are now cleanly distinct, no overloading:
- \`# docker\` → legacy daemon docker (v1, unchanged)
- \`# sandbox\` → run the bash script under nsjail (unchanged)
- \`# sandbox <image>\` → run the image's command under nsjail (this runtime)

**docker-compose**: dropped the \`dind\` sidecar (service, \`DOCKER_HOST\`, \`depends_on\`, volume) — the sandboxed runtime is daemonless (podman + nsjail in the worker). Host-socket opt-in for legacy full docker is documented inline.

**Language picker**: removed the guard that blocked Docker scripts on the multi-tenant/cloud platform (now safe to run sandboxed).

**Image pulls / freshness / limits** (new, all configurable):
- Pulls live in podman's rootless graph root (persistent, dedups). Per-job rootfs is in the job dir and cleaned with it.
- \`SANDBOX_IMAGE_PULL_POLICY\` (default \`newer\`) — re-pull only when the registry digest changed, so moving tags like \`:latest\` don't go stale.
- \`SANDBOX_IMAGE_MAX_SIZE_MB\` (default 0=off) — reject oversized images before extraction.
- \`SANDBOX_IMAGE_CACHE_MAX_MB\` (default 0=off) — best-effort LRU eviction of podman's image store; in-use images are never removed.

**Re-verified e2e** after the rename: \`# sandbox alpine\` → runs sandboxed (uid 0, confined); \`# docker alpine\` → plain bash (no longer v2); bare \`# sandbox\` → nsjail-bash modifier intact. podman primitives the size/eviction code shells out to (\`--pull=newer\`, \`image inspect {{.Size}}\`, \`images --format json\` Id/Size/Created) confirmed. The env-gated size/eviction paths were validated at the podman-primitive level (the default-off path is exercised by the e2e); frontend gate removal isn't observable on this non-cloud dev instance.

rubenfiszel:
## Addressed CI review findings

All blockers fixed and verified end-to-end on a podman+nsjail worker.

### P0 — `LD_PRELOAD`/`LD_*` via image env reaching the nsjail launcher (Claude, Pi, Codex)
Fixed in `c959eaa9a0`. The image's OCI `Env` no longer enters nsjail's process env. It's now rendered as proto-escaped `envar:` directives (child-only) via `render_envars`; nsjail's own process env carries **only** windmill-trusted keys (reserved vars + proxy). So a hostile `ENV LD_PRELOAD=rootfs/evil.so` can't act on the nsjail binary.
- Verified: `# sandbox python:3.12-slim` → `PATH` is the image's PATH (delivered via envar), `WM_TOKEN` present (process env via keep_env), args bind. Added `render_envars` unit test asserting `LD_PRELOAD` goes to an `envar:` directive, not process env.

### P0 — host write through image-controlled rootfs script symlink (Codex)
Fixed in `cdf35c301e`. The body is no longer written into the image-controlled rootfs (`write_file` follows symlinks). It's passed straight to `/bin/sh -c <body> sh <args>` — **no file is written into the rootfs at all**, so a planted `.windmill_docker_main.sh` symlink has nothing to capture.
- Verified: body executes, `set -e` preserved, positional args bind.

### P1 — `sandbox_registry_auth` credentials logged (Codex)
Fixed in `cdf35c301e` + `00ea55ce41`. Two log sites: the reload now uses a secret-aware path (loads directly, logs only `(redacted), configured=<bool>`), and `sandbox_registry_auth` is added to `SENSITIVE_SETTINGS` so the settings-write log (`format_setting_value`) redacts it too.
- Verified: set a recognizable credential, grepped worker+server logs → **0 occurrences**; logs show `"{"au****"}}}"` and `(redacted)`.

### P2s
- `proto_str` octal escaping for control/non-ASCII bytes — already in `83a66aebad` (octal `\NNN`, not `\u{..}`); both reviewers that saw the latest commit confirmed it.
- `enforce_image_size_limit` now `tracing::warn!`s instead of silently bypassing on inspect failure (`84ade6dbcf`).
- `EVICTION_RUNNING` now reset via a `Drop` guard so a panic/early-return can't leave it stuck (`84ade6dbcf`).
- Size-guard-after-pull / `Created`-as-LRU-proxy: acknowledged as accurate; the pre-pull `skopeo` size check is already listed as a follow-up in `docs/docker-v2-runtime.md`.

Still needs a reviewer with the right env (noted in the test plan): a private-registry pull with **real** credentials, and the worker-without-nsjail error path.

PR #9454, opened 2026-06-05
feat: make C# dotnet target framework configurable via DOTNET_TARGET_FRAMEWORK

rubenfiszel:
## Summary

The C# executor hardcoded `<TargetFramework>net9.0</TargetFramework>` in the generated `Main.csproj`, so workers could only build/run C# scripts against .NET 9. This makes the target framework configurable via a new `DOTNET_TARGET_FRAMEWORK` worker env var, defaulting to `net9.0` (no behavior change for existing setups).

Motivation: a customer POC requires pricing libraries that only target .NET 8 (and later .NET 10). They installed the .NET 8 SDK on their workers, but the hardcoded `net9.0` in the csproj forced the build/target to .NET 9 regardless. Their attempt to set a `TargetFramework=net8.0` env var did not work, because an explicit `<TargetFramework>` value in the project's `PropertyGroup` takes precedence over environment variables in MSBuild property resolution. Feeding the value directly into the generated csproj fixes this.

## Changes

- Add `DOTNET_TARGET_FRAMEWORK_DEFAULT` const (`net9.0`) and a `DOTNET_TARGET_FRAMEWORK` `lazy_static` env var, mirroring the adjacent `DOTNET_ROOT` pattern.
- Interpolate the configured value into the `<TargetFramework>` of the generated `Main.csproj` in `gen_cs_proj`.
- Because both the `dotnet restore` (lockfile) and `dotnet publish` (build) paths share `gen_cs_proj`, the single change covers the whole pipeline.

Note: a recent enough SDK (9 or 10) can down-target older frameworks, so the recommended setup is to keep the newest SDK as `dotnet` and set `DOTNET_TARGET_FRAMEWORK` per worker group — but a pinned SDK 8 + `net8.0` also works.

## Test plan

- [ ] With `DOTNET_TARGET_FRAMEWORK` unset, confirm generated `Main.csproj` still carries `net9.0` and existing C# scripts compile/run unchanged.
- [ ] Set `DOTNET_TARGET_FRAMEWORK=net8.0` on a worker (csharp feature, .NET 8 SDK installed), run a C# script, and confirm the csproj carries `net8.0` and the script compiles/runs.
- [ ] `cargo check -p windmill-worker --features csharp` passes.

---
## Summary by cubic
Make the C# executor’s target framework configurable via `DOTNET_TARGET_FRAMEWORK` (default `net9.0`). The value is written into `<TargetFramework>` in the generated `Main.csproj` and included in the binary cache key, enabling .NET 8–10 and preventing cross-framework cache collisions.

<sup>Written for commit 5a6b2d0a3d998ed63c6af7c69188339a4b01b118. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-worker/src/csharp_executor.rs
+++ b/backend/windmill-worker/src/csharp_executor.rs
@@ -57,8 +57,12 @@ const DOTNET_ROOT_DEFAULT: &str = "C:\\Program Files\\dotnet";
 const DOTNET_ROOT_DEFAULT: &str = "/usr/share/dotnet";
 
 #[cfg(feature = "csharp")]
+const DOTNET_TARGET_FRAMEWORK_DEFAULT: &str = "net9.0";
+
+#[cfg(feature = "csharp")]
 lazy_static::lazy_static! {
     static ref DOTNET_ROOT: String = std::env::var("DOTNET_ROOT").unwrap_or_else(|_| DOTNET_ROOT_DEFAULT.to_string());
+    static ref DOTNET_TARGET_FRAMEWORK: String = std::env::var("DOTNET_TARGET_FRAMEWORK").unwrap_or_else(|_| DOTNET_TARGET_FRAMEWORK_DEFAULT.to_string());
 }
 
 #[cfg(feature = "csharp")]
@@ -212,6 +216,7 @@ fn gen_cs_proj(
         )
     };
 
+    let target_framework = DOTNET_TARGET_FRAMEWORK.as_str();
     write_file(
         job_dir,
         "Main.csproj",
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Addressed CI review in 5a6b2d0:

**Cache key omits target framework (Claude P2, Codex P1)** — fixed. `DOTNET_TARGET_FRAMEWORK` is now folded into the `calculate_hash` input in `handle_csharp_job`, so a script built for `net8.0` and one built for `net9.0` get distinct local + object-store cache keys. This prevents a heterogeneous fleet from pulling a binary built for the wrong framework — which is exactly the mixed-framework scenario this PR unlocks. Note: because the default `net9.0` is now part of the hash, existing C# binary caches are invalidated once on upgrade (a one-time rebuild, harmless).

**No XML escaping/validation of the env var (cubic P2)** — not addressing. The value comes from a worker-level env var set by the instance operator, not from end-user/job input, so there's no privilege boundary to cross. A malformed value just produces a `dotnet restore/publish` failure in the operator's own build (a config error surfaced as a build error), not injection. Pi's review reached the same conclusion.

`cargo check -p windmill-worker --features csharp` passes.

PR #9455, opened 2026-06-05
feat(sandbox): pull/extract images with crane instead of podman

rubenfiszel:
## Summary

Follow-up to #9453. That PR shipped the `# sandbox <image>` runtime but shelled out to **podman** — which was **never added to any Dockerfile**, so the feature can't actually run in the shipped image. And podman is overkill here: the runtime only ever **pulls + flattens** an image (nsjail does the run), so it doesn't need a container engine, store, root, or privileged at all.

This switches the fetch backend to **[`crane`](https://github.com/google/go-containerregistry)** — a single ~25 MB static binary — and **adds it to the images**.

| | podman (before) | crane (now) |
|---|---|---|
| footprint | 150–300 MB + crun/conmon/fuse-overlayfs/slirp4netns/uidmap… | one ~25 MB static binary |
| needs | rootless store, subuid, privileged-in-container | nothing — no daemon/store/root |
| in Dockerfile | **missing** (feature broken in image) | added |

## Changes

- **`docker_v2.rs`**: `crane export` → flattened rootfs tar, `crane config` → OCI config, `crane digest` → a **content-addressed rootfs+config cache** (cross-job dedup + automatic freshness on tag moves), `crane manifest` → a **pre-download** size guard (rejects before any layer is fetched). Private-registry auth via a per-job `DOCKER_CONFIG` dir. Cache eviction prunes the rootfs-tar cache by mtime (true LRU). Pull policy honored via a ref→digest cache (`missing`/`never` reuse without a registry hit).
- **`Dockerfile`, `docker/DockerfileSlim`, `docker/DockerfileSlimEe`**: install the `crane` binary (pinned `v0.20.6`, arch-aware). `DockerfileFull`/`FullEe` and the EE image inherit it via `FROM` the base image.
- **docs / UI / instance-setting descriptions** updated (size cap is now compressed download size; cache is the rootfs-tar cache; auth via `DOCKER_CONFIG`).

The two follow-ups the parent PR listed — *content-addressed rootfs cache* and *pre-pull size guard via manifest inspection* — are **done** here as a consequence of the crane model.

## Test plan

- [x] crane spike: `export` (flattened worker-owned rootfs), `config` (Env/Cmd/Entrypoint/WorkingDir), `manifest --platform` (layer sizes), `digest`, `DOCKER_CONFIG` auth — all confirmed
- [x] e2e on the dev backend: `# sandbox alpine` runs via crane (uid 0, body inside the image); the digest-keyed cache (`sha256_<digest>.tar` + `.json` + `refs/`) is populated
- [x] `cargo check` (full) + `docker_v2` unit tests (`proto_str`, `render_envars`, `registry_qualified`, `sandbox_image`) + frontend `check:fast` all clean
- [ ] **deploy test (in progress):** build the image, run `# sandbox alpine` under docker-compose and minikube to confirm crane + nsjail work in-container

---
## Summary by cubic
Switch sandbox image pull/extract to `crane` instead of `podman`, add `crane` to our images, and harden caching and limits. This fixes the shipped sandbox runtime, reduces footprint, and enforces a compressed download size cap with digest-pinned pulls.

- **New Features**
  - Pull/extract via `crane` (no daemon/store/root): uses `crane export`, `config`, `digest`, and `manifest` with `nsjail`; fetches by `name@digest` to keep cache consistent on moving tags.
  - Digest-keyed rootfs+config cache with a ref→digest cache honoring pull policy (`missing`/`newer`/`always`/`never`); `.size` sidecar enforces the compressed cap on every job; atomic ref→digest and config sidecars with tolerant reads.
  - Pre-download size guard using `crane manifest` (compressed size).
  - Evict oldest cached rootfs tars by creation time (mtime) under `sandbox_image_cache_max_mb`; eviction hardening: stage via hardlink, falling back to copy across filesystems; sweep orphaned `*.tmp.*`; remove sidecars with the tar.
  - Private-registry auth via per-job `DOCKER_CONFIG`. Docs and instance setting texts updated.

- **Dependencies**
  - Add `crane` v0.20.6 static binary to `Dockerfile`, `docker/DockerfileSlim`, and `docker/DockerfileSlimEe`; fail fast on unsupported arch. `podman` is no longer required.

<sup>Written for commit 361c93121d7a1d551b935ab776b41164f098b731. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -311,7 +311,7 @@ COPY --from=nsjail /nsjail/nsjail /bin/nsjail
 # Single static binary — no daemon/store/root needed. See docs/docker-v2-runtime.md.
 ARG CRANE_VERSION=v0.20.6
 RUN arch="$(dpkg --print-architecture)"; \
-    case "$arch" in amd64) crane_arch=x86_64 ;; arm64) crane_arch=arm64 ;; *) crane_arch="$arch" ;; esac; \
+    case "$arch" in amd64) crane_arch=x86_64 ;; arm64) crane_arch=arm64 ;; *) echo >&2 "error: unsupported arch '$arch' for crane"; exit 1 ;; esac; \
     wget -O /tmp/crane.tgz "https://github.com/google/go-containerregistry/releases/download/${CRANE_VERSION}/go-containerregistry_Linux_${crane_arch}.tar.gz" \
     && tar -xzf /tmp/crane.tgz -C /usr/local/bin crane \
     && rm /tmp/crane.tgz \
--- a/Dockerfile
+++ b/Dockerfile
@@ -307,6 +307,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends libprotobuf32 l
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 COPY --from=nsjail /nsjail/nsjail /bin/nsjail
 
+# crane: pulls + flattens images for the sandboxed container runtime (`# sandbox <image>`).
+# Single static binary — no daemon/store/root needed. See docs/docker-v2-runtime.md.
+ARG CRANE_VERSION=v0.20.6
+RUN arch="$(dpkg --print-architecture)"; \
+    case "$arch" in amd64) crane_arch=x86_64 ;; arm64) crane_arch=arm64 ;; *) crane_arch="$arch" ;; esac; \
+    wget -O /tmp/crane.tgz "https://github.com/google/go-containerregistry/releases/download/${CRANE_VERSION}/go-containerregistry_Linux_${crane_arch}.tar.gz" \
+    && tar -xzf /tmp/crane.tgz -C /usr/local/bin crane \
+    && rm /tmp/crane.tgz \
+    && chmod +x /usr/local/bin/crane
+
 WORKDIR ${APP}
 
 RUN ln -s ${APP}/windmill /usr/local/bin/windmill
--- a/backend/windmill-worker/src/docker_v2.rs
+++ b/backend/windmill-worker/src/docker_v2.rs
@@ -3,8 +3,9 @@
 //! Unlike the legacy `# docker` (dind/daemon) path, this has no daemon and no Docker
 //! API. It splits *pull* from *run*:
 //!
-//! 1. **pull/extract** (podman, rootless): materialize the image's root filesystem
-//!    into `{job_dir}/rootfs` and read its OCI config (Env/Cmd/Entrypoint/WorkingDir).
+//! 1. **pull/extract** (`crane`, no daemon/store/root): materialize the image's root
+//!    filesystem into `{job_dir}/rootfs` and read its OCI config
+//!    (Env/Cmd/Entrypoint/WorkingDir), via a digest-keyed rootfs cache.
 //! 2. **run** (the job's own nsjail sandbox): execute the image command with the
 //!    extracted rootfs bound in as the new root, so the container inherits exactly
 //!    the job's confinement (filesystem mask, pid namespace, network, uid) and can't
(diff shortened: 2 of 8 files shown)


rubenfiszel:
## Addressed CI review findings (`6bc2f98`)

**Codex P1s — all fixed:**

1. **Mutable-tag fetch after digest resolution.** `export`/`config`/`manifest` now fetch the **digest-pinned** ref (`name@digest`, built from the digest `resolve_digest` returns), so content can't diverge from the key it's cached under if a tag moves mid-fetch. (Verified crane accepts `name@digest`.)
2. **Size cap bypassed on cache hit.** `enforce_image_size_limit` now runs on **every** job, reading a `{digest}.size` sidecar (written once from `crane manifest`) so cache reuse needs no registry call — lowering `SANDBOX_IMAGE_MAX_SIZE_MB` now rejects already-cached oversized images.
3. **Eviction race.** Before `tar -xf`, the cache tar is **hardlinked** into the job dir (pins the inode, so concurrent eviction can't pull it out); if it was evicted between the existence check and the link, the job **re-fetches** (2 attempts).

**Claude P2s — fixed:** config sidecar is now written atomically (tmp+rename) and a torn/missing parse falls back to default; the "true LRU" comment is softened to "creation order" (tars are write-once); eviction now also reclaims orphaned `*.tmp.*` files and the `.size`/`.json` siblings. Added `digest_key`/`ref_key` unit tests (incl. a multibyte-ref no-panic case).

Pi's review was **"Good to merge"** with no findings.

Still pending (the one unchecked test-plan box): the docker-compose + minikube deploy e2e — the image is mid-build. The crane + nsjail in-container mechanism is already validated (ran the full extract→nsjail chroot inside the windmill `:main` image under `--privileged`); this confirms the end-to-end through the worker process.

rubenfiszel:
## Addressed re-review (`bbb4cc4`)

**P1 — cross-filesystem cache staging (Claude + Codex).** Correct and important catch: the eviction-race hardlink crosses filesystems in the shipped deployments (the cache is its own volume at `/tmp/windmill/cache`, the job dir is on the container fs), so `hard_link` returns `EXDEV` — not `NotFound` — and every sandbox job would fail. Fixed: on a non-`NotFound` link error, fall back to `tokio::fs::copy`, which reads *through* the source inode so it still completes if eviction unlinks the tar mid-copy (race protection preserved cross-fs). This is precisely the regression the docker-compose/minikube deploy test gates on.

**P2 — Dockerfile arch fallback.** All three Dockerfiles now **fail fast** with a clear message on an unsupported arch instead of building a 404 crane URL.

**Minor — `resolve_digest` torn ref read.** The ref→digest file is now written via tmp+rename, so a concurrent `missing`/`never` reader never sees a torn file.

Remaining noted-minors (`refs/` not reclaimed; `never` may touch the registry if a `.size` sidecar is missing) are edge-casey and left as-is.

The deploy e2e (the gating box) is still pending the image build. Note: my dev-box postgres went into recovery under the parallel build load, so I verified these with the SQLX offline cache (no new SQL) + unit tests; the cross-fs path is what the deploy test will confirm green.

rubenfiszel:
Addressed the last Codex P2 (`361c93121d`): the docs (`docker-v2-runtime.md`) and the instance-setting description no longer say "LRU" — they now say "oldest (by creation time)", matching the implementation (tars are write-once; cache hits don't touch mtime). Pi's review was "Good to merge" with no findings.

PR #9456, opened 2026-06-05
feat(otel): connect jobs to the inbound distributed trace

hugocasa:
**Companion EE PR: windmill-labs/windmill-ee-private#604** (required for the trace logic).

## Problem

Windmill's run-job API ignores **inbound** distributed-trace context. Every job's
trace identity is derived deterministically from its UUID (`get_otel_context_envs`,
`span_cx_from_job_id`, `add_root_flow_job_to_otlp`), so a job triggered by an
instrumented upstream service shows up as a **disconnected** trace rather than
being connected to the originating request. #8277 added the *outbound* half
(`TRACEPARENT` env in job subprocesses); this is the inbound counterpart.

A common workaround is to inject a W3C carrier into job args and re-parent spans
in a script-side decorator — fragile across upgrades because it depends on the
UUID→trace-id derivation and arg filtering. This makes it a first-class capability.

## What this does

When a request that enqueues a job carries a valid W3C `traceparent`, the job is
**relocated into that originating distributed trace**:

- The job's worker span is parented on the inbound caller's span (the whole job /
  flow nests under the originating request).
- The script subprocess's injected `TRACEPARENT` env is the inbound context, so
  the **script's own spans join the caller's trace** — no script-side changes.
- Flow steps inherit the context, so a whole flow relocates as one connected tree.
- A span **link** back to the UUID-derived context is kept so trace-by-job-id
  lookup still resolves.

Jobs triggered **without** a `traceparent` are completely unchanged (trace id =
job UUID), so there is no regression for existing OTel users.

```mermaid
graph TD
  C["caller request span<br/>(trace T_in)"]
  C --> F["job / full_job span<br/>(trace T_in, span = job UUID)"]
  F --> S1["flow step 1"]
  F --> S2["flow step 2 → script spans"]
  F -. "span link" .-> U["job-UUID context<br/>(trace = job UUID)"]
```

## Approach (reserved-arg-key, EE/OTel)

- **Capture** (OSS): the run endpoints' shared arg-build chokepoint
  (`WebhookArgs::to_args_from_format`) reads the inbound `traceparent` header and
  stashes it as a reserved `_wm_traceparent` arg key — gated on
  `OTEL_TRACING_ENABLED` so non-tracing instances never accumulate the key. It
  rides the existing `args` jsonb exactly like `_ENTRYPOINT_OVERRIDE`; normal
  scripts never see it because args are bound by declared parameter name. **No
  migration, no `push()` change, no SQL.**
- **Carry to the worker** (OSS): a new `LogContext.inbound_traceparent` field set
  per-job in `log_context_for_job`, read by `get_otel_context_envs` to relocate
  the script env. Flow steps inherit the key at the `worker_flow` push site
  (mirroring `_TEMP_SCRIPT_REFS`).
- **Re-parent** (companion EE PR): `otel_ee::set_job_span_parent` re-parents the
  worker span on the inbound context (+ UUID link); `add_root_flow_job_to_otlp`
  relocates the `full_job` span the same way.

## Scope / limitations

- Active only on EE builds with OTel tracing enabled.
- Covers the standard `run` / `run_wait_result` / stream endpoints (which capture
  request headers). Native-trigger paths (headers consumed before args) and
  JSON-body endpoints (preview/batch/dependencies/restart) don't capture a
  `traceparent` today.
- A restart / "run again with same args" carries the captured `traceparent`
  forward (intentionally accepted).
- The secondary internal `job_postprocessing` span stays on the UUID-derived
  context (`MiniCompletedJob` carries no args); the primary job span is relocated.

## Validation

- `cargo check --features enterprise,private,otel`; CE (`not private`) and
  `private,enterprise,license` builds all compile.
- Tests in `backend/tests/otel.rs` (all passing): `span_cx_from_traceparent`
  valid / unsampled / 8× malformed; `inbound_span_cx_from_job`
  present / absent / malformed; **`test_root_job_span_relocated_to_inbound_trace`**
  (relocated trace id, UUID-derived span id, parented on caller, link to UUID),
  and the existing `test_root_job_trace_id_matches_uuid` confirms no regression
  for non-inbound jobs.
  Run: `cargo test --features enterprise,private,otel --test otel -- --test-threads=1`.
- No SQL / OpenAPI / frontend changes.
- Note: `test_root_job_span_created_on_success` fails on `main` too (asserts
  `job_kind`/`created_by` attributes `add_root_flow_job_to_otlp` has never set) —
  pre-existing, unrelated to this change; left untouched.


hugocasa:
## Verified end-to-end against Jaeger

Drove the real relocate code (`add_root_flow_job_to_otlp` + `set_job_span_parent`) through a live OTLP export to Jaeger, with a synthetic upstream `caller_request` span (trace `1111…`, span `2222…`) standing in for an instrumented caller, and a job carrying that `_wm_traceparent`.

Jaeger trace: full_job nested under caller_request

The exported trace confirms the design:

- The job's `full_job` span lands **in the caller's trace** (`11111111111111111111111111111111`), **parented on the caller's span** (`CHILD_OF 2222…`) — i.e. nested under the originating request.
- Its span id is the **job-UUID-derived** `82abdbbd317df019` (job `5ee1e539-…-82ab-dbbd317df019`), unchanged.
- It carries a **span link back to the UUID-derived context** (`FOLLOWS_FROM` → trace `5ee1e539…82abdbbd317df019`), so trace-by-job-id lookup still resolves.

(The full HTTP→worker path wasn't exercised because this dev instance has no workspaces and the worker is license-gated; the integration test `test_root_job_span_relocated_to_inbound_trace` plus this OTLP smoke cover the trace logic.)

hugocasa:
## Update: verified the full HTTP → worker → script path end-to-end

(Supersedes the caveat in my previous comment — the real path *was* exercised.)

Ran a second otel-enabled instance against the shared dev DB, deployed a `bash` script that echoes its trace env, and `run_wait_result`'d it with an inbound header `traceparent: 00-aaaa…-bbbb…-01`:

```
result: "GOT_TRACEPARENT=00-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbb-01
         GOT_TRACE_ID=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```

So the **script subprocess received the caller's exact trace context** (not the UUID-derived one) — the headline requirement, confirmed on a real job.

And the job's spans landed in the caller's trace (`aaaa…`) in Jaeger:

```
trace aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  span da28ff2b01150101  name='job'       parent=bbbbbbbbbbbbbbbb   (the inbound caller span)
  span f13e7896b66a62ff  name='bash run'  parent=da28ff2b01150101
```

real job span nested in the inbound trace

Together with the relocate integration test and the synthetic OTLP smoke ([caller → full_job + UUID link](https://files.catbox.moe/hoy7ll.png)), the feature is verified end-to-end: **capture → worker pickup → span relocate → script env injection**.

PR #9470, opened 2026-06-08
feat: allow private MCP server URLs

### LINKED ISSUE #9466: feature: Allow opt-in support for private/internal MCP server URLs via environment variable, similar to ALLOW_PRIVATE_AI_BASE_URLS
### Problem

Currently, the SSRF (Server-Side Request Forgery) validation for MCP server URLs in `backend/windmill-mcp` is **not configurable or skippable via environment variable**. As of [commit 8053266](https://github.com/windmill-labs/windmill/commit/8053266f88bd4c94fc86278412df5a0beeed5e77), all MCP resource URLs are strictly validated using `validate_url_for_ssrf`, which blocks all private/internal targets (e.g., `http://127.0.0.1`, `http://10.x.x.x`, etc.), with no opt-out or environment-controlled override.

By contrast, AI provider resources (OpenAI, Azure, Anthropic, etc.) allow opt-in for private/internal endpoints using the `ALLOW_PRIVATE_AI_BASE_URLS=true` environment variable, which is documented and used in production/testing.

**This means it is much harder (effectively impossible) to connect to a private/self-hosted MCP server for dev, testing, or some on-prem deployments compared to AI providers.**

### Proposal

Add support for opt-in private MCP server URLs, using a documented environment variable (such as `ALLOW_PRIVATE_MCP_SERVER_URLS=true` or a similar name), matching the approach used for AI providers. When this configuration is active:
- The SSRF guard should be bypassed for MCP server URLs ONLY when explicitly allowed.
- It should still be secure-by-default, so the override should not be active unless the env is set.
- Consider aligning error reporting with the AI provider path (e.g., provide actionable hints in the SSRF error messages).

### Rationale
- Improves parity and developer experience compared to AI provider URL handling.
- Facilitates self-hosted, airgapped, development, or advanced testing setups (where MCP and Windmill run in the same network or container).
- Keeps SSRF protections "secure by default" for production and managed deployments.

### References
- [SSRF guard for MCP](https://github.com/windmill-labs/windmill/blob/main/backend/windmill-mcp/src/client/mod.rs#L46-L48)
- [MCP SSRF implemented in 8053266](https://github.com/windmill-labs/windmill/commit/8053266f88bd4c94fc86278412df5a0beeed5e77)
- [AI provider SSRF env override](https://github.com/windmill-labs/windmill/blob/main/backend/windmill-ai/src/ai_providers.rs#L39-L69)

### Acceptance Criteria
- `ALLOW_PRIVATE_MCP_SERVER_URLS=true` (or similar) allows using MCP resource URLs that would otherwise be blocked as private/internal
- Without the env, behavior remains unchanged (secure by default)
- Doc update for SSRF protection and the new override
- Ideally consistent error reporting with AI provider SSRF errors

---
_No sensitive credential examples included in this feature request._

centdix:
## Summary
Adds an instance-wide opt-in for private/internal MCP server URLs while keeping private URL access blocked by default for SSRF protection.

Fixes #9466.

## Changes
- Add `ALLOW_PRIVATE_MCP_SERVER_URLS=true` / `1` for MCP server URLs.
- Centralize MCP URL validation so syntax and `http`/`https` guards still apply when private URLs are allowed.
- Validate MCP OAuth discovery, registration, and token endpoints, and disable redirects for OAuth discovery/DCR requests.
- Update the backend threat model.

## Test plan
- [x] `cargo test --manifest-path backend/Cargo.toml -p windmill-mcp`
- [x] `cargo test --manifest-path backend/Cargo.toml -p windmill-mcp --features auth`
- [x] `git diff --check`
- [x] Backend cargo-watch logs show healthy rebuild/runtime health checks
- [x] Local MCP smoke server listed/called tools directly; Windmill without the env rejected the private MCP resource with the expected env hint


centdix:
Addressed the review comment about the OAuth token exchange redirect path.

Changes pushed:
- The EE OAuth callback now re-validates `credentials.token_endpoint` immediately before the token POST.
- The token exchange POST now uses the MCP no-redirect HTTP client, matching discovery/DCR behavior.
- Updated `backend/ee-repo-ref.txt` to the new EE commit.

Validation:
- Backend recompiled and restarted successfully.
- Manual redirect regression: local OAuth MCP `/token` returned `307` to `http://127.0.0.1:8777/ssrf`; Windmill returned a token exchange failure and the sink received no request.
- Normal local OAuth MCP flow still succeeds and returns `MCP_CONNECTED`.

centdix:
Addressed the timeout review comment.

Changes pushed:
- `windmill_mcp::oauth::no_redirect_http_client()` now preserves rmcp's previous 30-second request timeout while still disabling redirects.
- MCP OAuth discovery and DCR now reuse that shared helper instead of building timeout-less clients inline.
- Added a regression test with a stalled local response using a small test timeout.

Validation:
- `cargo test -p windmill-mcp --features auth no_redirect_http_client_times_out_stalled_responses`
- `cargo test -p windmill-mcp --features auth,server`

PR #9471, opened 2026-06-08
fix(oauth): persist refreshed token through configured secret backend

rubenfiszel:
## Problem

On self-hosted EE with an **external secret backend** (AWS Secrets Manager / Azure Key Vault / Vault), a connected OAuth resource's access token works right after connecting, then stops being refreshed ~1h later. Consuming scripts fail with e.g. a Google `RefreshError`.

## Root cause

OAuth refresh is **lazy and on-fetch only** — when a secret variable backing an OAuth resource is read and the linked `account` is expired, the token is exchanged and persisted. All refresh write paths persisted the new token with a raw `UPDATE variable SET value = <db-encrypted>`, **bypassing the secret-backend abstraction**.

But secret **reads** resolve through the backend (`get_secret_value` → `backend.get_secret`), which **ignores `variable.value`** for external backends. So on refresh:

- `account.expires_at` advances ✓ (Postgres)
- the external store (e.g. the AWS SM secret) is **never rewritten** ✗

Every read that doesn't itself trigger a mint keeps serving the frozen connect-time token, which expires ~1h after connect. Symptoms match the report precisely: `account.expires_at` advancing in Postgres while the SM secret's `LastChangedDate` stays frozen at connect time, and **no** refresh/secret errors logged (the backend write is never attempted).

The connect path was unaffected because it creates the variable via `create_variable` → `store_secret_value`, which is why the external secret existed at connect time.

## Fix

`windmill-oauth` cannot depend on `windmill-store` (would be circular), so variable persistence moves out of `windmill_oauth::refresh_token{,_for_account}` (now they only exchange the token + update the `account` row and return the new access token) and into the `windmill-store` callers. A new helper `store_oauth_token_value` (in `windmill-store::secret_backend_ext`) writes through the configured backend via `store_secret_value` and updates `variable.value` with the returned value.

All four entry points now persist through the backend:
- `oauth_refresh_oss` (OSS / DB-only build)
- EE standard OAuth / client_credentials, workspace integration, and MCP paths (companion PR)

Behavior is **unchanged for the default database backend** (still encrypt + write to `variable.value`).

## EE companion

This is an `[ee]` PR — the `_ee.rs` changes live in windmill-labs/windmill-ee-private#607 and are pulled in via `backend/ee-repo-ref.txt` (updated in this PR).

## Validation

- `cargo check -p windmill-oauth` ✓
- `cargo check -p windmill-store --features enterprise,private,oauth2` ✓
- `cargo check -p windmill-store --features enterprise,private,oauth2,mcp` ✓

(Compile validation used the committed `.sqlx` offline cache: the local dev DB was unavailable in this environment. Runtime e2e against a live external backend was not possible here — see note below.)

## Test note

A runtime regression test would require a live external secret backend (e.g. localstack AWS SM) plus a mock OAuth token endpoint; the existing secret-backend integration tests are `#[ignore]`d for the same reason and the dev DB was down in this environment, so an automated regression test was not added. The fix is a localized routing change verified by reasoning + compilation across all relevant feature combinations.

---
## Summary by cubic
Fixes OAuth token refresh with external secret backends by writing refreshed access tokens through the configured backend so reads return the new token. Adds a safeguard to reset account expiry and record an error if persistence fails to avoid serving stale tokens.

- **Bug Fixes**
  - Moved token persistence out of `windmill-oauth`; `refresh_token{,_for_account}` now only exchange the token, update `account`, and return the access token.
  - Added `store_oauth_token_value` in `windmill-store::secret_backend_ext` to write via the configured backend and update `variable.value`; on failure, reset `account.expires_at` and set `refresh_error`.
  - Updated `windmill-store::oauth_refresh_oss::_refresh_token` to persist the refreshed token through the backend.
  - Added opt-in E2E tests in `backend/windmill-store/src/oauth_refresh_secret_backend_tests.rs` covering DB backend, AWS Secrets Manager via LocalStack, and the failed persist path.

<sup>Written for commit 4bb36f9c9ecac3e0e3e0d3d4fdd92acc8b81a45c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-oauth/src/lib.rs
+++ b/backend/windmill-oauth/src/lib.rs
@@ -26,7 +26,6 @@ use windmill_common::error::{self, to_anyhow, Error};
 use windmill_common::more_serde::maybe_number_opt;
 use windmill_common::oauth2::*;
 use windmill_common::utils::now_from_db;
-use windmill_common::variables::{build_crypt, encrypt};
 use windmill_common::BASE_URL;
 
 pub type DB = sqlx::Pool<sqlx::Postgres>;
@@ -525,11 +524,14 @@ pub struct OAuthAccountInfo {
     pub scopes: Option<Vec<String>>,
 }
 
-/// Refresh an OAuth token and update the database.
+/// Refresh an OAuth token and update the `account` row.
 /// Fetches the account from DB, then delegates to `refresh_token_for_account`.
+///
+/// Returns the freshly minted access token. Persisting it to the secret variable
+/// backing the resource is the caller's responsibility (it must route through the
+/// configured secret backend — see `store_oauth_token_value` in `windmill-store`).
 pub async fn refresh_token<'c>(
     mut tx: Transaction<'c, Postgres>,
-    path: &str,
     w_id: &str,
     id: i32,
     db: &DB,
@@ -549,7 +551,6 @@ pub async fn refresh_token<'c>(
 
     refresh_token_for_account(
         tx,
-        path,
         w_id,
         id,
         db,
@@ -562,9 +563,14 @@ pub async fn refresh_token<'c>(
 }
 
 /// Refresh an OAuth token given pre-fetched account info (no additional SELECT).
+///
+/// Exchanges the refresh token, updates the `account` row (`refresh_token`,
+/// `expires_at`, `refresh_error`) and returns the new access token. It does NOT
+/// persist the token to the secret variable — the caller must do that through the
+/// configured secret backend (`store_oauth_token_value`), otherwise an external
+/// secret backend would keep serving the stale connect-time token.
 pub async fn refresh_token_for_account<'c>(
     mut tx: Transaction<'c, Postgres>,
-    path: &str,
     w_id: &str,
     id: i32,
     db: &DB,
@@ -676,17 +682,6 @@ pub async fn refresh_token_for_account<'c>(
     tx.commit().await?;
 
     let token_str = token.access_token.to_string();
-    let mc = build_crypt(db, w_id).await?;
-    let encrypted_token = encrypt(&mc, token_str.as_str());
-
-    sqlx::query!(
-        "UPDATE variable SET value = $1 WHERE workspace_id = $2 AND path = $3",
-        encrypted_token,
-        w_id,
-        path
-    )
-    .execute(db)
-    .await?;
 
     tracing::info!(
         grant_type = %account.grant_type,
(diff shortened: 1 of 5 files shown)


rubenfiszel:
Addressed the cubic P1 (account marked fresh before token persisted).

`store_oauth_token_value` now resets `account.expires_at` to the past (and records `refresh_error`) if persisting the refreshed token fails. So instead of the account staying "fresh" with a stale secret for the full token lifetime, the **next fetch re-runs the refresh** — self-healing. Account id is threaded through all refresh paths. (Note the failure mode pre-existed for the DB backend too; routing through an external backend just made the network write more likely to fail, so it's worth handling.)

Also added an opt-in e2e regression suite `windmill-store/tests/oauth_refresh_secret_backend.rs` (gated on `RUN_SECRET_BACKEND_E2E` / `RUN_AWS_SM_TESTS`), verified locally against Postgres + LocalStack:

```
test database_backend_persists_refreshed_token ... ok   # without external storage
test external_backend_persists_refreshed_token ... ok   # with AWS SM: token written back to the store
test failed_persist_resets_account_expiry ... ok        # self-healing reset
test result: ok. 3 passed
```

The external-backend test also asserts that a raw `UPDATE variable SET value=...` leaves the served token unchanged (reads resolve through the backend) — i.e. it reproduces the original bug shape and confirms the fix.

rubenfiszel:
Both addressed in the latest push:

**[P1]** Narrowed `store_oauth_token_value` from `pub` to `pub(crate)` (it's only ever called by the two in-crate refresh adapters, `oauth_refresh_oss`/`oauth_refresh_ee`) and gated it on `feature = "oauth2"`. Added an explicit **authorization contract** to the doc comment: it performs no access control and callers must have already authorized the operation against `workspace_id`/`path` (the refresh adapters only run after the read path has resolved and gated the variable). The regression test moved in-crate as a `#[cfg(test)]` unit module so it no longer needs the function to be public (also let me drop the `[dev-dependencies]` I'd added, since unit tests use the crate's normal deps).

**[P2]** Fixed the doc-comment run commands — both now include `RUN_SECRET_BACKEND_E2E=1` (and the feature set now includes `oauth2`), so following them actually runs the suite instead of silently skipping.

Re-verified against Postgres + LocalStack: 3 passed (database backend, AWS SM write-back, self-healing reset). No dead-code warnings in `private,enterprise`, `…,oauth2`, or `…,oauth2,mcp`.

PR #9472, opened 2026-06-08
[ee] fix(oauth): write refreshed tokens through configured secret backend

rubenfiszel:
Fixes WIN-2016

## Problem

When an external secret backend is configured (HashiCorp Vault, AWS Secrets Manager, Azure Key Vault), OAuth token refresh updated `account.expires_at` in Postgres and returned the fresh token inline, but never wrote the refreshed access token to the external store. The three refresh code paths each did `build_crypt()` + `encrypt()` + a direct `UPDATE variable SET value`, bypassing the `store_secret_value()` abstraction that routes writes through the configured backend.

The variable's value in the external store therefore stayed frozen at its connect-time value: only the call that triggered the refresh got a valid token; every subsequent read within the token lifetime served the stale/expired token from the external store.

## Fix

Route refreshed tokens through `store_secret_value()`, which encrypts for the database backend or writes to the external store and returns the appropriate marker (e.g. `$aws_sm:<path>`).

The two EE-only paths (`_refresh_workspace_integration_token`, `_refresh_mcp_token` in `oauth_refresh_ee.rs`) call `store_secret_value()` directly.

The standard OAuth / client-credentials path lives in `windmill-oauth::refresh_token_for_account`. `windmill-oauth` **cannot** depend on `windmill-store` — `windmill-store` already depends (optionally) on `windmill-oauth`, so importing it back would create a crate dependency cycle. The issue's suggested direct call therefore doesn't compile. Instead:

- `windmill-oauth` exposes `STORE_SECRET_VALUE_HOOK` (`OnceLock<fn>`). On refresh it routes the new token through the hook when set, falling back to local DB encryption when no backend hook is registered (pure OSS / database-only builds — behaviour identical to before).
- `windmill-store::secret_backend_ext::register_oauth_store_secret_hook()` populates the hook with `store_secret_value`. The EE refresh dispatcher (`_refresh_token`) registers it before delegating, so the standard path is covered for any build that reaches an external backend.

This keeps the secret-storage logic owned by `windmill-store` (which owns the backend abstraction) without inverting the crate dependency graph.

### Files
- `backend/windmill-oauth/src/lib.rs` — hook type/static + use it in `refresh_token_for_account`
- `backend/windmill-store/src/secret_backend_ext.rs` — `register_oauth_store_secret_hook()` helper
- `oauth_refresh_ee.rs` (EE, companion PR) — convert MCP + workspace-integration paths to `store_secret_value`, register the hook in the dispatcher

## EE companion PR
windmill-labs/windmill-ee-private#606

## Validation
- `cargo check` (default), `cargo check --features enterprise,private`, and `cargo check --features enterprise,private,mcp` all pass for `windmill-oauth` + `windmill-store`.
- No `sqlx::query!` text changed (only the bound variable name/type, which is identical `String`), so the offline cache needs no update.
- Draft: the external-backend write path requires an EE license + live Vault/AWS SM/Azure KV and could not be exercised end-to-end in this environment. The database-backend path is behaviourally unchanged (hook fallback == previous `build_crypt`+`encrypt`).

---
## Summary by cubic
Fixes WIN-2016 by routing refreshed OAuth tokens through the configured secret backend so Vault/AWS SM/Azure KV get the new value. Prevents stale tokens after refresh and keeps DB-only behavior unchanged.

- **Bug Fixes**
  - `windmill-oauth`: added `STORE_SECRET_VALUE_HOOK`; refresh now stores via the hook, falling back to local encryption when unset.
  - `windmill-store`: added `register_oauth_store_secret_hook()` to wire the hook to `store_secret_value`.
  - EE: `_refresh_workspace_integration_token` and `_refresh_mcp_token` now use `store_secret_value`; the dispatcher registers the hook.
  - No migration needed; external-backend reads now return fresh tokens.

<sup>Written for commit 59eee84dcd36cffdf3b58e8089c6b390b2def277. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9472?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Closing as a duplicate of #9471, which fixes the same WIN-2016 bug and is the more complete implementation (lifts token persistence into the windmill-store callers, drops the now-unused `path` param, adds a `store_oauth_token_value` wrapper, and documents the remaining raw-encrypt slack sites). This PR used a global `STORE_SECRET_VALUE_HOOK` instead; #9471's caller-persists approach avoids the process-global state and is preferable. Deferring to #9471 / windmill-ee-private#607.

PR #9475, opened 2026-06-08
fix: refresh session editor preview on breadcrumb target switch

Guilhem-lm:
## Summary

In the session preview pane, navigating the breadcrumb between flows left the **settings-panel path stale** (and other mount-initialized state too). The preview swaps the editor's `path` prop in place — deliberately, to keep the persistent `UserDraft.useMany` sync cell alive — so descendants that initialize at mount never re-derive. `Path.svelte`'s internal `meta` (set once in `initPath()`) is the visible symptom; the same structural bug existed for script→script and rawapp→rawapp switches.

This consolidates the three near-identical view modules into one deep module that owns the switching/sync lifecycle, and lands the fix once: the heavy editor is **remounted on a data-ready target swap** so its descendants re-init from the new target.

## Changes

- **New `SessionEditorTarget.svelte`** — owns the editor-target lifecycle behind one seam: load trigger, `isActiveSession`-gated live-editor registration, the loading/notFound/remount gate, and the draft sync. The bug fix is `{#key slot.loadedPath}` around the heavy editor — keyed on the **data-ready** `loadedPath` (not the raw `path` prop), so the remount fires after the async load settles and sidesteps the Monaco `parentNode` init race. A 150ms-debounced overlay covers the brief stale window; `notFound && loadedPath !== path` renders NotFound on a failed switch.
- **`FlowEditorView` / `ScriptEditorView` / `RawAppEditorView`** collapse to thin adapters: a `DiffDrawer` + an `{#snippet editor()}` + a `<SessionEditorTarget>` call (~100 fewer lines each).
- **`LoadSlot` in `sessionRuntime.svelte.ts`** — replaced 9 loose load-state fields with three `{ loadedPath, loading, notFound }` slots + a `slot(kind)` accessor; deleted the now-unused legacy getters.
- **New `useUserDraftSync.svelte.ts` + `sessionDraftCodecs.ts`** — generalize the inbound/outbound `untrack`-discipline UserDraft sync, parameterized by a per-kind `DraftSyncCodec`.

**Behavior contract:** zero change **except** script outbound debounce `0 → 150ms`. Preserved exactly: flow inbound's `initFlowState` rebuild (wipes per-module test args — known trade-off) and script outbound's `{...(current ?? script), ...script}` merge-save.

## Test plan

- [x] `npm run check` — 0 errors; `npm run check:fast` clean; 37 sessions unit tests pass; svelte-autofixer clean
- [x] Browser (Playwright): flow preview with Settings open, switch flow_alpha → flow_beta → **settings Path field updates to flow_beta** (the reported bug)
- [x] Browser: script→script and flow→flow switches remount with **no `parentNode`/null Monaco error** in console
- [ ] Deploy a script/flow from a session → preview reloads to the deployed version (force-reload `undefined→path`) with no flash of "not found"
- [ ] Chat writes to a script while typing in its editor → typing burst persists once after ~150ms; AI write reflects into the editor without reverting keystrokes

### DIFF
--- a/frontend/src/lib/components/sessions/FlowEditorView.svelte
+++ b/frontend/src/lib/components/sessions/FlowEditorView.svelte
@@ -2,13 +2,8 @@
 	import FlowBuilder from '$lib/components/FlowBuilder.svelte'
 	import DiffDrawer from '$lib/components/DiffDrawer.svelte'
 	import type { WorkspaceItem } from '$lib/components/workspacePicker'
-	import { untrack } from 'svelte'
 	import type { SessionRuntime } from './sessionRuntime.svelte'
-	import type { Flow } from '$lib/gen'
-	import { UserDraft } from '$lib/userDraft.svelte'
-	import { flowDraftSig } from './flowDraftSig'
-	import { initFlowState } from '$lib/components/flows/flowState'
-	import SessionItemNotFound from './SessionItemNotFound.svelte'
+	import SessionEditorTarget from './SessionEditorTarget.svelte'
 	import { sendUserToast } from '$lib/toast'
 
 	let {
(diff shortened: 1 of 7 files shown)


Guilhem-lm:
Thanks for the reviews — triaged all findings, pushed `f9c72e6453`.

**Fixed**
- **cubic P1 — keystroke drop on target switch** (`ScriptEditorView`/the debounce): real, and newly introduced *for scripts* by the uniform 150ms debounce (scripts previously saved immediately). A breadcrumb swap within the debounce window cleared the pending `UserDraft` write instead of flushing it. Added a path/workspace-scoped flush effect in `useUserDraftSync` that flushes the pending write on switch/unmount, without flushing mid-typing-burst (so the debounce is preserved). This also fixes the same latent drop that flow/raw_app already had.
- **claude P2 / cubic P3 — stale `loadScript` comment**: it named the removed `loadedScriptPath` / `{#if loading && !loadedScriptPath}` symbols; updated to point at `SessionEditorTarget`'s `{:else if slot.loadedPath === undefined}` gate.
- **claude P2 — PR description listed `sessions/CONTEXT.md`**: that file was intentionally dropped before commit; description updated.

**Won't change (pre-existing, not introduced by this PR)** — flagging so they're not lost, happy to do follow-ups if wanted:
- **cubic P2 — inbound `ready()` untracked, can miss a draft written mid-load**: the pre-refactor views also read `loadedPath` inside `untrack`, and `loadFlow`/`loadScript` seed the store from the draft on completion. Making `ready()` tracked would re-run the inbound effect on every switch and re-fire flow's `initFlowState` (the test-args wipe) redundantly — strictly worse. Intentional.
- **cubic P2 — script sig is `content`-only, drops language/summary-only updates**: the original script sync also deduped on `incoming === script.content` and outbound tracked only `script.content`; the preview is `neverShowMeta` so those fields aren't user-editable here. Faithful to the prior behavior (the contract was zero behavior change except the debounce).

PR #9481, opened 2026-06-08
fix: drop archived items from fork compare (spurious 'not visible' warning)

rubenfiszel:
## Summary

In the fork/merge **Compare workspaces** view, a superadmin could see the alert *"This fork has changes not visible to your user … Only a user with access to the whole context may deploy or update this fork"* and be blocked from deploying — even though a superadmin can see everything by construction.

**Root cause.** `compare_workspaces` caches its per-item verdict in `workspace_diff` (`has_changes`, `exists_in_*`) and trusts a cached `has_changes = true` row without re-running the per-kind comparison. That cache can go stale: the **old path after a rename** keeps only `archived` versions, and for lock-gen languages (Python/TS/…) the `has_changes = NULL` reset is **deferred to the dependency job** (`scripts.rs` returns `hdm = None` when `needs_lock_gen`). Until that job runs — or permanently if it's slow/fails — the archived old path lingers as a `has_changes = true, exists_in_fork = true` row.

`filter_visible_diffs` then asks `query_visible_items` (which treats `archived` as non-existent) whether the user can see that path; it returns nothing — even under a superadmin's RLS-bypassing `windmill_admin` role — so the item is misread as *hidden*, flipping `all_ahead_items_visible` to `false` and tripping the warning + deploy block.

**Principle:** archived scripts/flows are **non-existent** for diff purposes (which is exactly what `compare_two_scripts`/`compare_two_flows` already do). The stale cached row violated that by surfacing an archived item as a live "ahead" change.

## Changes

- `compare_workspaces` (`backend/windmill-api-workspaces/src/workspaces.rs`): when a cached `has_changes = true` row is encountered, cheaply re-validate it with the new `runnable_still_exists` helper — if the item no longer exists (`archived = false`) on a side the cache claims, fall through and re-evaluate via `compare_two_*`, which **deletes** (or corrects) the stale row. So an archived/renamed-away item is dropped entirely: no phantom diff, no false warning — for admins and non-admins, and symmetrically for both the source (parent) and fork sides.
- The staleness probe is **batched** (`existing_runnables`): one `SELECT … path = ANY($2) AND archived = false` per kind per side builds the live-paths set up front, so the compare path adds at most 2 queries per side total regardless of diff count — not one probe per row. Scoped to `script`/`flow` (the only kinds with `archived`; others reset synchronously on delete). Runs on `&db` (existence probe, not an authz check — authorization stays in `filter_visible_diffs`/`query_visible_items`, left unchanged).
- Regression coverage folded into the existing `test_compare_workspaces_comprehensive` (it already builds a parent + fork as superadmin and exercises diff-row (re)evaluation): seeds an archived script + a stale `has_changes = true, exists_in_fork = true` diff row and asserts it's dropped (not surfaced as a diff, physically deleted from `workspace_diff`) and `all_ahead_items_visible` stays `true`.
- Regenerated the `.sqlx` offline cache (`--all-targets`): two `EXISTS` probes from the helper + the new test's seed/verify queries.

> Note: a first iteration instead dropped `archived = false` from `query_visible_items`; that only suppressed the warning and left the archived item showing as a phantom deployable diff. Reverted in favor of treating archived as non-existent at the source, per review.

## Test plan

- [x] `test_compare_workspaces_comprehensive` (now covering the stale-archived case) passes; existing fork-compare tests still pass.
- [x] Verified on the running dev backend across both directions (archived in fork / in parent / in both → dropped; live-on-one-side divergences → preserved) and roles (superadmin → no warning; non-admin → archived dropped but still warned about a genuinely-hidden live item).
- [ ] Manual e2e: as a fork member, rename a lock-gen-language (Python/TS) script before its dependency job runs, open the fork/merge compare view — confirm the old path doesn't appear as an ahead change and no "changes not visible" warning fires.

### DIFF
--- a/backend/.sqlx/query-0a568f630e069118fe302099a709e89cc4a702158899f7fdc66d0922e8fb9b29.json
+++ b/backend/.sqlx/query-0a568f630e069118fe302099a709e89cc4a702158899f7fdc66d0922e8fb9b29.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT path FROM flow WHERE workspace_id = $1 AND path = ANY($2) AND archived = false",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "path",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "TextArray"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "0a568f630e069118fe302099a709e89cc4a702158899f7fdc66d0922e8fb9b29"
+}
--- a/backend/.sqlx/query-4c81384b579bad74b64c72ca053839f316fe5412f99f3d3bbbdc0c65f55ab794.json
+++ b/backend/.sqlx/query-4c81384b579bad74b64c72ca053839f316fe5412f99f3d3bbbdc0c65f55ab794.json
@@ -1,0 +1,20 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT has_changes FROM workspace_diff\n         WHERE path = 'f/shared/renamed_away' AND kind = 'script' AND source_workspace_id = 'test-workspace'",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "has_changes",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      true
+    ]
+  },
+  "hash": "4c81384b579bad74b64c72ca053839f316fe5412f99f3d3bbbdc0c65f55ab794"
+}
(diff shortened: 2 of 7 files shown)


rubenfiszel:
Addressed the Codex P1 (N+1 stale-cache probes). The per-row `runnable_still_exists` is replaced by a batched pre-pass: for cached `has_changes = true` script/flow rows, a single `SELECT … path = ANY($2) AND archived = false` per kind per side builds `live_source`/`live_fork` sets (mirroring `query_visible_items`), and the loop does set lookups instead of issuing a query per row. So the hot compare path now adds at most 2 queries per side total (script + flow) regardless of how many cached diffs there are, instead of O(number of cached script/flow diffs) sequential round trips. Other kinds aren't probed (they reset synchronously on delete), so they short-circuit with no query. Verified: all 4 OSS `workspace_comparison` tests pass, plus a live multi-row check (2 stale archived scripts + 1 stale archived flow dropped, live script preserved). sqlx cache regenerated.

PR #9496, opened 2026-06-09
fix: accept unresolved references in schema_validation

### LINKED ISSUE #6938: bug: schema validation fails if resources used
### Describe the bug

If I enable the schema validation with "// schema_validation" and my script includes parameters that are resources, the execution fails to error "argument XX should be an object".

### To reproduce

1. Create new script
2. Use this example code
```javascript
// schema_validation

type Gmail = {
  token: string
}
export async function main(
  gmail: Gmail
): Promise<boolean> {
  return gmail != null;
}
```
3. Try to execute the script

### Expected behavior

There should not come any error.

### Screenshots

_No response_

### Browser information

_No response_

### Application version

_No response_

### Additional Context

_No response_

ameyypawar:
## Summary

When a script enables `// schema_validation` and has a parameter typed as a resource (or any object), executing it with that parameter supplied as a Windmill reference fails:

    Argument `gmail` should be an object

Schema validation runs *before* the worker resolves references, so a `$res:…` (or `$var:`/`$jsonvar:`/`$encrypted:`) reference is still a string at validation time and is rejected by the object rule.

## Changes

- `windmill-common/src/schema.rs`: `SchemaValidationRule::apply_rule` now short-circuits when the value is an unresolved reference string (`$var:`/`$jsonvar:`/`$res:`/`$encrypted:` — the prefixes the worker resolves), accepting it regardless of the declared type. The worker resolves the reference to its real value after validation runs.
- Add a regression test (`validate_accepts_unresolved_reference_for_object_param`) covering all four reference prefixes, a genuine object (still validates), and a plain non-reference string (still rejected).

## Test plan

- [ ] `cargo test -p windmill-common validate_accepts_unresolved_reference_for_object_param`
- [ ] Deploy the issue's repro (a `// schema_validation` script with a resource parameter) and run it with the resource selected; confirm no "should be an object" error.

Fixes #6938

---
## Summary by cubic
Accept unresolved Windmill references during schema validation so `$res:`/`$var:`/`$jsonvar:`/`$encrypted:` values no longer fail type checks for shape-based params; enum checks still run. Previously these strings were rejected as wrong type; now shape rules skip them and unions are walked so nested enums are enforced. Fixes #6938.

- Narrow the skip in `SchemaValidationRule::apply_rule` to shape rules (object, array, number, integer, bool, oneOf); keep `StrictEnum` enforced and walk `IsUnionType` so enums under `anyOf` are still validated.
- Add `windmill-common::references` with shared prefixes and `is_reference`; use it in schema validation and in `windmill-store`’s `missing_references`.
- Tests cover all four prefixes, object-under-anyOf acceptance, and rejection for enums (including under anyOf); plain non-reference strings remain rejected.

<sup>Written for commit 724fdf54555246b528454c239dcaf017f8462324. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa (commented):
Thanks for this, and sorry for the slow review. The approach is right and the bug is wider than the issue suggests, so I'd like to get it in. Three things and then it's good to merge:

1. **Narrow the early return** so it only skips the rules an unresolved reference genuinely can't satisfy (`IsObject`, `IsArray`, `IsNumber`, `IsInteger`, `IsBool`, `IsOneOf`, `IsUnionType`), leaving `StrictEnum` enforced. As written it sits at the top of `apply_rule`, so a param with `enum: ["a", "b"]` now accepts `$var:u/me/anything`, and since the resolved value is never validated afterwards, `// schema_validation` ends up guaranteeing nothing at all for reference-valued args. Enum params never accepted references before, so keeping that check costs users nothing.

2. **Give the four prefixes one home.** This list now exists in `windmill-worker/src/common.rs` (`transform_json_value`, the resolver that defines the truth), in `missing_references` in `windmill-store/src/resources.rs`, in `ArgInput.svelte` on the frontend, and now here. Please add a small const plus predicate in `windmill-common`, use it from `schema.rs` and from the `windmill-store` copy, and say in one place that it has to track the branches in `transform_json_value`. Otherwise the next prefix we add resolves at runtime but gets rejected by validation, which is this exact bug again.

3. **Keep the rationale in one comment.** It is currently in both the production comment and the test comment.

One gap you do not need to handle, just so it is on the record: `transform_json_value` only recurses into arrays while `depth <= 2`, so a reference nested deeper than that passes validation and reaches the script as a literal string. Encoding that limit in the validator too would be worse than living with it.

Context on why this is worth fixing beyond the reported case: typing a resource param the current way (`type Gmail = { token: string }`) resolves to `Typ::Object { name: Some("gmail"), props }`, which `argSigToJsonSchemaType` emits as `format: resource-gmail` **and** `properties`. `from_primitive` checks `properties` before its resource branch, so those params get full `IsObject` rules and every `$res:` value is rejected. Only the older `Resource<'gmail'>` annotation escaped, since it emits no `properties` and therefore no rules. So this is the normal resource path, not an edge case.

PR #9509, opened 2026-06-10
feat: workspace protection rule to restrict anonymous app deployment

rubenfiszel:
Fixes WIN-2026

## Summary

Any user with write access to an app can toggle its execution mode to `anonymous` (no login required), making the app publicly accessible without authentication or oversight. This PR adds an opt-in **workspace protection rule** — `RestrictAnonymousAppDeployment` — to gate that action. **By default (rule not enabled), behavior is unchanged**: any writer can still deploy anonymous apps, exactly as before.

When a workspace protection ruleset enables the rule, only workspace admins and the ruleset's bypass users/groups can create an anonymous app or flip an existing app to anonymous.

**Judgment call on the update path**: the frontend sends the full `policy` object on *every* deploy, so a flat "anonymous ⇒ blocked" check would lock restricted editors out of deploying apps that are already public. The update gate therefore only blocks the *transition into* anonymous mode (checked against the app's current policy in the same transaction): restricted users can still redeploy an already-anonymous app (no exposure change) and downgrade it to `publisher` (exposure reduction), but cannot flip an app to public.

## Changes

- `backend/windmill-common/src/workspaces.rs`: new `RESTRICT_ANONYMOUS_APP_DEPLOYMENT` bitflag + `ProtectionRuleKind::RestrictAnonymousAppDeployment` variant with its message.
- `backend/windmill-api/src/apps.rs`:
  - `create_app_internal`: creating an app with `execution_mode: anonymous` runs `check_user_against_rule` for the new kind (admins always pass; bypass users/groups pass; everyone passes when no ruleset enables the rule).
  - `update_app_internal`: flipping an app's policy to `anonymous` runs the same check, unless the app is already anonymous. Missing row / missing `execution_mode` fail closed (gate applies).
- `backend/windmill-api/openapi.yaml`: new `ProtectionRuleKind` enum value (frontend + CLI clients are generated from this; gen dirs are gitignored).
- `frontend/src/lib/components/workspaceSettings/RulesetEditor.svelte`: new "Restrict public app access" toggle in the protection-rules editor.
- `frontend/src/lib/components/apps/editor/AppEditorHeaderDeploy.svelte`: the "No login required" toggle is disabled (with an explanatory alert) when the rule is active and the user can't bypass it, reusing `canUserBypassRuleKind` from the existing protection-rules state (loaded on boot in `+layout.svelte`). Restricted users can still toggle an already-public app back to login-required.
- `backend/tests/app_anonymous_execution_mode.rs`: integration test covering default behavior (no rule → non-admin can deploy anonymous apps), rule enforcement (create + flip blocked with 403 naming the ruleset, publisher create OK), admin exemption, redeploy-while-anonymous, publisher downgrade, and bypass-user exemption.

## Validation

- `cargo test --test app_anonymous_execution_mode` passes (1 passed, covering all scenarios above).
- UI verified end-to-end in headless Chromium against the dev stack:
  - no rule, non-admin: toggle enabled, no alert (previous behavior) — https://files.catbox.moe/256gle.png
  - rule enabled, non-admin: toggle disabled + "Restricted by a workspace protection rule" alert — https://files.catbox.moe/il8k6m.png
  - rule enabled, admin: toggle enabled, no alert — https://files.catbox.moe/10kcj2.png
  - ruleset editor with the new "Restrict public app access" toggle — https://files.catbox.moe/063j2z.png
- `npm run check:fast` reports only pre-existing errors (unrelated `setTimeout`/`Timeout` type errors present on baseline).
- Backend compiles cleanly; no new SQL queries beyond the current-policy lookup already cached in `.sqlx/`.

## Notes

- The rule integrates with the existing protection-ruleset machinery: 60s server-side cache (invalidated on rule CRUD), per-ruleset bypass users/groups, `wm_deployers` semantics unaffected.
- CLI `protection-rules` commands pass kinds through generated types, so no CLI command changes were needed.

### DIFF
--- a/backend/.sqlx/query-3e82929b365a6aa7ccc39fc5615c4110e1d740e0cb0509b1c58bc6636d83cafd.json
+++ b/backend/.sqlx/query-3e82929b365a6aa7ccc39fc5615c4110e1d740e0cb0509b1c58bc6636d83cafd.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "SELECT policy->>'execution_mode' = 'anonymous' FROM app WHERE path = $1 AND workspace_id = $2",
+  "query": "SELECT policy->>'execution_mode' = 'anonymous' FROM app WHERE path = $1 AND workspace_id = $2 FOR UPDATE",
   "describe": {
     "columns": [
       {
@@ -19,5 +19,5 @@
       null
     ]
   },
-  "hash": "8c6310025b3338863c6a7250dff38538b262d60bc8f839f103e07f6a9cb3dd00"
+  "hash": "3e82929b365a6aa7ccc39fc5615c4110e1d740e0cb0509b1c58bc6636d83cafd"
 }
--- a/backend/.sqlx/query-8c6310025b3338863c6a7250dff38538b262d60bc8f839f103e07f6a9cb3dd00.json
+++ b/backend/.sqlx/query-8c6310025b3338863c6a7250dff38538b262d60bc8f839f103e07f6a9cb3dd00.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT policy->>'execution_mode' = 'anonymous' FROM app WHERE path = $1 AND workspace_id = $2",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "?column?",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "8c6310025b3338863c6a7250dff38538b262d60bc8f839f103e07f6a9cb3dd00"
+}
--- a/backend/tests/app_anonymous_execution_mode.rs
+++ b/backend/tests/app_anonymous_execution_mode.rs
@@ -1,0 +1,166 @@
+//! Regression test for the admin gate on anonymous app execution mode.
+//!
+//! Any user with write access to an app could set its policy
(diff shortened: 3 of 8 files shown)


rubenfiszel:
Addressed the CI review findings in cd475e3e67:

1. **TOCTOU race on the `already_anonymous` check** (Codex P1, Pi P1): the current-policy read in `update_app_internal` now uses `SELECT … FOR UPDATE`, locking the app row for the rest of the transaction. A concurrent admin downgrade either commits first (the locked read re-evaluates against the latest committed row, sees `publisher`, and the protection-rule gate applies) or waits on the lock and applies after the redeploy commits (admin wins). Either way a restricted user's stale redeploy can no longer silently restore anonymous access. sqlx offline cache updated for the new query (EE caches verified intact, zero net deletions vs main).

2. **Fail-open UI while protection rules load** (identified by cubic, P2): `canSetAnonymous` now fails closed for non-admin users until `protectionRulesState.rulesets` is loaded — the toggle is never briefly enabled for a user the rules will end up restricting. Workspace admins and superadmins short-circuit (they always pass the backend check), and the warning alert only renders once rules are actually loaded so it doesn't flash during the load window. The backend check was always authoritative; this just removes the cosmetic window.

3. **Existing tests creating anonymous apps with non-admin tokens** (Codex P1 on the earlier hardcoded-gate revision): resolved by the ruleset rework itself — with no `RestrictAnonymousAppDeployment` ruleset configured, behavior is unchanged. Verified by running the suites the review named: `preserve_on_behalf_of` (9/9 passed) and `folder_default_permissioned_as` (passed), plus `app_anonymous_execution_mode` (passed).

UI re-verified end-to-end after the changes (no rule → toggle enabled for non-admin; rule on → disabled + alert for non-admin, unaffected for admin).

PR #9514, opened 2026-06-10
fix: make default chat model optional in AI settings

centdix:
## Summary

In workspace and instance AI settings, removing the model that was the **default chat model** used to leave the default empty **and** block saving (an empty default while providers were configured failed the save-validation), which was confusing.

This makes the default chat model **optional**. When the configured default is removed it is cleared automatically, and the config can be saved without one. The "default chat model" is only a fallback — it's the model new chats start with when the user hasn't picked one per-session. At runtime, chat and metadata generation already fall back to the first configured model (`aiModels[0]`) when no default is set, so an empty default is a valid, usable state.

## Changes

All in `frontend/src/lib/components/workspaceSettings/AISettings.svelte` (shared by workspace settings and instance settings via `InstanceAISettings`):

- Replace the zero-providers-only cleanup `$effect` with a centralized **validity effect** that clears `defaultModel` / `metadataModel` / `codeCompletionModel` whenever a selection is no longer present in the configured providers' models. This now covers **both** removal paths — disabling a whole provider *and* removing a model from a provider's "Enabled models" list (the latter previously had no cleanup, leaving a stale default).
- Remove the `(Object.keys(aiProviders).length > 0 && !defaultModel)` condition from `isSaveDisabled()` so an empty default no longer blocks saving.
- Simplify the provider-disable toggle handler — per-model cleanup is now handled reactively by the validity effect.
- Make the "Default chat model" `Select` `clearable`.

## Test plan

Verified end-to-end in a real browser (chrome-devtools) against a test workspace with Anthropic + Google AI providers configured:

- [x] **Remove default via multiselect**: removed the model set as default from its provider's "Enabled models" — the Default chat model field cleared and **Save stayed enabled** (previously blocked).
- [x] **Save with empty default**: saved successfully ("AI settings updated", no errors); reloaded and the empty default **persisted and read back clean** (no spurious unsaved-changes state).
- [x] **Runtime fallback**: after saving with no default, AI chat auto-switched to the first configured model.
- [x] **Provider-disable path**: handled by the same validity effect.
- [x] **`clearable` selector**: the clear (x) control renders and re-selecting a model works.
- [x] Restored the test workspace to its original config afterward.
- [x] Local review (fresh-context) returned **good to merge**, 0 blocking issues.

Note: Windmill does not unit-test Svelte components and this change adds no pure-logic utility, so no automated tests are added.

### DIFF
--- a/frontend/src/lib/components/workspaceSettings/AISettings.svelte
+++ b/frontend/src/lib/components/workspaceSettings/AISettings.svelte
@@ -202,11 +202,19 @@
 			)
 		)
 	)
+	// Keep model selections valid: when a selected model is no longer in the
+	// configured providers (provider disabled or model removed from the list),
+	// clear it. The default chat model then falls back to the first configured
+	// model rather than pointing at a model that no longer exists.
 	$effect(() => {
-		if (Object.keys(aiProviders).length < 1) {
-			codeCompletionModel = undefined
+		if (defaultModel && !selectedAiModels.includes(defaultModel)) {
 			defaultModel = undefined
+		}
+		if (metadataModel && !selectedAiModels.includes(metadataModel)) {
 			metadataModel = undefined
+		}
+		if (codeCompletionModel && !selectedAiModels.includes(codeCompletionModel)) {
+			codeCompletionModel = undefined
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


centdix:
Re: the Codex P2 (provider-enable repopulating the optional default) — decision: we're making the default chat model **fully optional** rather than trying to keep it populated.

- Removed the auto-seed in the provider-enable handler entirely (commit c0e8c109a5), so enabling a provider never (re)introduces a default. An empty default now stays empty, including when you enable additional providers after saving with no default.
- We intentionally did **not** add any "default = first configured model" auto-fill in the settings UI — the default is purely what the user explicitly sets, or empty.
- The internal safety net stays: when `default_model` is unset, chat/metadata resolve to the first configured model at runtime (`defaultModel ?? aiModels[0]` in `aiStore.ts` / `ProviderModelSelector`), so chat always has a model to use. We're only dropping the *UI auto-fill*, not the runtime resolution.

Verified in-browser: with a saved empty default, enabling another provider leaves the Default chat model empty (no auto-seed); the newly-enabled provider still gets its own first model populated in its Enabled models list.

PR #9517, opened 2026-06-10
style: lean tool-call rendering and tighter ai chat spacing

Guilhem-lm:
## Summary

UI polish for the AI chat panel: leaner tool-call rendering and tighter, more consistent spacing. Cosmetic only — no behavioural or API changes.

## Screenshots

**Tool-call rendering** — compact collapsed row, expanded single-border `p-3` content box with plain `text-hint` labels and copy-on-hover:

tool calls

**Indicators** — typing dots sit snug under the message, and the "Waiting for your input" pill now has a backdrop so it's not see-through:

indicators

> Note: screenshots use representative tool-call content (the dev AI provider was erroring locally); the markup/styling shown is the real component output.

## Changes

**Tool calls** (`ToolExecutionDisplay.svelte`, `ToolContentDisplay.svelte`)
- Borderless in both collapsed and expanded states; a single bordered, padded (`p-3`) box wraps the tool-call content
- Parameters/Result/Logs render plain (removed the per-section filled boxes); section labels use `text-hint`
- Header: content-width (`inline-flex`), one-line height (`py-0.5 my-0.5`), no horizontal padding; chevron moved to the end and animates (`rotate-90`, `transition-transform duration-150`)
- Expand/collapse now uses a `slide` transition (`duration: 150`)
- Copy button is hidden and fades in only on hover of its own section (`opacity-0 group-hover:opacity-100 transition-opacity`, + `focus-visible`)
- Removed the separator between tool actions and the rest of the content
- Status: removed the success/error ✓/✗ glyphs, kept the loading spinner

**Chat spacing** (`AIChatDisplay.svelte`, `AIChatMessage.svelte`)
- Reduced the gap between a sent message and the typing indicator (`-mt-10`, counters the message's `!mb-12`)
- Tightened vertical spacing for tool messages (`mb-1`, `py-0`) so runs of tool calls pack closer
- Added a `bg-surface/80 backdrop-blur` background to the "Waiting for your input" pill so it is no longer see-through

## Test plan
- [ ] Open the AI chat (Ask AI), send a message — typing indicator sits snug under the message
- [ ] Trigger tool calls — collapsed rows are compact/borderless; expanding slides open with the chevron rotating
- [ ] Hover a Parameters/Result section — copy button fades in only for that section
- [ ] Confirm "Waiting for your input" has an opaque-ish backdrop over scrolling content

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -422,13 +422,13 @@
 					{#if showTypingIndicator}
 						<div
 							class={twMerge(
-								'sticky z-10 mt-0.5 ml-2 self-start pointer-events-none',
+								'sticky z-10 -mt-10 ml-2 self-start pointer-events-none',
 								showFlowPendingActionControls ? 'bottom-14' : 'bottom-2'
 							)}
 						>
 							{#if waitingForUserAction}
 								<span
-									class="inline-flex items-center gap-1.5 text-2xs text-accent"
+									class="inline-flex items-center gap-1.5 px-2 py-1 rounded-md bg-surface/80 backdrop-blur text-2xs text-accent"
 									aria-label="Waiting for your input"
 								>
 									<Hourglass class="w-3 h-3 hourglass-flip" />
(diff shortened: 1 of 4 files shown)


Guilhem-lm:
Thanks @claude — addressed the P2 in 7d035d253b. Since the ✓/✗ status glyphs were intentionally removed, I restored the error signal by coloring the tool name red when `message.error` is set, covering the non-expandable case (tools without `showDetails`) while keeping the lean borderless header (no icon).

PR #9522, opened 2026-06-10
feat: enable native web search in copilot

centdix:
## Summary
Enables provider-native web search automatically for Copilot chat providers that can safely support it, while surfacing web search execution through the existing tool-status UI.

## Design choices
- Web search does not have a chat-side enable button. It is on by default for compatible providers when that provider's AI settings allow it.
- Web search is configured per provider, not globally, because support and failure modes differ by provider/model.
- OpenAI and Anthropic are the only compatible providers enabled for now.
- Azure OpenAI remains excluded until there is explicit Azure-specific UI because Azure web search has admin, cost, and Grounding with Bing data-boundary implications.
- The settings follow the existing AI config resolution between instance and workspace settings; there is no separate "inherit" state for web search.
- Web-search sources are not rendered separately. Provider-native web search activity is shown through the same tool-status UI used by other chat tools.
- If a provider rejects native web search, the same request is retried once without web search, then that workspace/provider/model is cached as unsupported for the current frontend session.
- When native web search is unavailable, the chat shows: `You can disable websearch in your workspace settings.`

## Changes
- Enable provider-native web search by default for OpenAI and Anthropic in the chat loop.
- Retry OpenAI/Anthropic once without native web search when the provider rejects the web-search tool, then cache that provider/model/workspace as unsupported for the session.
- Show provider web-search usage and fallback guidance as tool-status rows.
- Add a per-provider web search toggle to AI workspace/instance settings, defaulting on for compatible providers and off only when explicitly disabled.
- Replace generic Svelte component instance types with structural handles to keep fast frontend checks passing.

## Test plan
- [x] `npm run generate-backend-client`
- [x] `npm run test:unit -- src/lib/aiStore.test.ts src/lib/components/copilot/chat/AIChatManager.test.ts src/lib/components/copilot/chat/chatLoop.test.ts --run`
- [x] `npm run check`
- [x] `cargo check`
- [x] `git diff --check`
- [ ] Browser verification not run because no local backend/frontend dev server was listening on expected ports.


centdix:
Addressed the relevant review comments in `595e0b4103`:\n\n- Tightened the web-search fallback heuristic so unrelated generic tool errors no longer trigger the native web-search unsupported cache.\n- Updated the fallback status message to explain that native web search is unsupported for the model and that the request continues without it.\n- Updated the existing OpenAI fallback test and added coverage for unrelated OpenAI tool errors and the Anthropic retry path.\n\nValidation run:\n- `npm run test:unit -- src/lib/components/copilot/chat/chatLoop.test.ts`\n- `npm run check:fast`

centdix:
Addressed the Anthropic web-search fallback review comment in 2ebeee34c8.

What changed:
- Added a narrow detector for Anthropic enablement errors like `must enable web search`, so those requests retry once without native web search instead of rethrowing.
- Kept malformed web-search request errors on the existing non-retry path.
- Removed the local fake web-search failure hook from the worktree; it was not included in the pushed fix.

Validation run:
- `npm run test:unit -- src/lib/components/copilot/chat/chatLoop.test.ts`
- `npm run check:fast`
- `npm run check` (0 errors, existing warnings only)

PR #9539, opened 2026-06-11
feat(frontend): improve AI chat cancel and interrupted-turn handling

hugocasa:
## Summary

Improves how the AI chat handles cancelled, failed, and empty turns, and makes Escape a reliable stop shortcut. Previously, cancelling a generation discarded everything the model had already produced (a follow-up "continue" started blind), a turn that returned no output silently ate the user's message, and Escape-to-stop didn't work in the session chat because other widgets swallow the key.

## Changes

- **Escape stops the generation when focus is on the chat** (composer, a message, or anywhere in the panel — the panel root takes focus on click via `tabindex="-1"`). Implemented as a capture-phase window listener so neither the session's Monaco editor nor mounted-but-closed `Modal2` instances (which intercept every window Escape) swallow the key first. Focus in another widget (e.g. the editor) leaves Escape to that widget; `stopImmediatePropagation` ensures one press can't cancel several mounted chat panels.
- **Empty turns are rolled back**: if the model returns no output, or the user cancels before any answer, the user message is removed from the transcript and its text restored to the composer (skipped if a new draft was typed meanwhile). A rolled-back first turn also deletes the chat entry persisted earlier in the turn so it doesn't linger in history.
- **Interrupted turns keep their usable output**: on cancel or failure mid-turn, the completed tool round-trips (`truncateToToolPairedPrefix` — drops a dangling tool call that providers would reject on the next request) and the partial answer text are committed as context, so a follow-up like "continue" picks up from where it stopped. Cancelling while the model was only thinking clears the stuck-open reasoning bubble instead. Partial text flushed before a tool call survives the abort, and an answer that completed right before the abort isn't committed twice.
- **Thinking-block collapse is animated** with the same `slide` transition tool boxes already use.

Works across the provider paths (OpenAI Completions, OpenAI Responses API, Anthropic — including preserved thinking blocks on replayed tool-call messages).

## Test plan

- [x] 27 unit tests pass (`AIChatManager.test.ts` lifecycle paths, `chatLoop.test.ts` for the tool-pairing truncation)
- [x] Full `npm run check`: 0 errors
- [ ] With a generation running: Escape with focus in the composer / on a message / after clicking the chat stops it; focus in the session's Monaco editor leaves the generation running
- [ ] Cancel mid-answer, send "continue" — the partial answer stays in the transcript and the follow-up request succeeds (no provider 400)
- [ ] Cancel while the model is thinking — message returns to the composer, reasoning bubble is cleared
- [ ] Send a message against a model that returns nothing — message is restored with a toast

## Screenshots

The changes are interaction-behavioral (keyboard shortcut, cancel/restore flows, a collapse animation), so static screenshots can't capture most of them. Affected surface (session chat):

session-chat

---
## Summary by cubic
Improves cancel and interrupted-turn handling in chat: Escape reliably stops generation when the chat has focus, empty or reasoning‑only turns roll back to the input, and interrupted turns keep useful output so follow‑ups like “continue” work.

- **New Features**
  - Escape stops generation when focus is on the chat (composer, messages, panel).
  - Empty turns and pre‑answer cancels roll back; the message is restored to the composer. Rolled‑back first turns delete the just‑created history entry and show a toast.
  - Mid‑turn cancels/failures keep completed tool round‑trips and partial answer text; dangling tool calls are dropped (`truncateToToolPairedPrefix`); reasoning‑only bubbles are cleared; thinking block collapse is animated.

- **Bug Fixes**
  - Prevent double commits and wrong error flags from post‑commit failures; keep text flushed before a tool call on cancel; avoid duplicating a just‑finished answer; treat completed reasoning‑only turns as unsent so the user message is restored instead of swallowed.
  - Use stopImmediatePropagation so one Escape can’t cancel multiple mounted chat panels; restoring input skips if a new draft is present.
  - Add a focused comment policy to `AGENTS.md`; tighten comments and dedupe tests.

<sup>Written for commit 514fbd3b43ba5e52aaac9366906a3257ae4e148b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9539?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Addressed the Codex review in f0d237ba7e:

1. **[P1] Reasoning-only completed turn swallowed** — confirmed and fixed. The empty-turn decision now keys off `hasUsableOutput` (tool-paired prefix or partial answer text) instead of display output, so a turn that finishes with only reasoning rolls back and restores the composer like an empty one. Added a regression test (`restores the message when a completed turn produced only reasoning`).
2. **[P2] Comment length** — trimmed the flagged blocks to the ≤4-line norm, moving the `stopImmediatePropagation` rationale inline to its call site.

PR #9551, opened 2026-06-12
feat(frontend): precise AI chat context usage tracking + indicator

centdix:
## Summary

Track AI chat context usage from provider-reported token counts, compact (drop-oldest) the conversation before it overflows the model's context window, and surface a small context-window usage line above the chat input.

Until now the AI chat estimated context fill by summing `content.length / 4` over the messages array — ignoring the system prompt and tool definitions entirely (both are counted by providers and can dominate) — and used that estimate at 95% fill to silently drop the oldest messages from a per-send copy of the history.

## Design

**Stored fact, lazy fallback.** `contextUsage` holds provider usage reports only — the `prompt + completion` count of the last committed turn's final completion, which is exact and includes everything sent (system prompt, tool definitions, full history). Whenever no report describes the current history — the provider never reported usage, the turn failed, the history was rewound — the value is simply `undefined`. Readers never touch it directly: they go through `contextTokens`, a derived that falls back to a fresh chars/4 estimate of the stored context (messages + system prompt + tool defs) computed at the moment of reading. Because the fallback lives at the single read point instead of being written at every site that mutates history, no code path can leave a stale or missing value behind, and the estimate is never accumulated — errors can't compound. The compaction trigger and the indicator both consume `contextTokens`; there is no anchor indexing and no re-basing on mode switches.

**Compaction triggers off that number, with headroom instead of precision.** Before each send, if `contextTokens + chars/4(new user message) ≥ 80%` of the model's context window, head messages are dropped until ~70%. The 20% headroom absorbs what the number can't see: the upcoming completion and tool results, system-prompt/tool-schema growth from mode switches, and the estimate's chars/4 error. Notably this works from the very first send and on providers that never report usage — the lazy estimate keeps the trigger armed.

- **Compaction mutates the stored history** (the API messages), not a per-send copy. What is sent IS what is stored, so the next report describes the history exactly. `displayMessages` are untouched: the user keeps seeing the full conversation, and user-message indices are re-based onto the compacted history (a message whose API counterpart was dropped rewinds to an empty history on restart, since compaction only removes prefixes).
- The drop loop never strands a turn: it keeps dropping until the history restarts on a **user** message, so tool results never dangle without their `tool_calls` message (including multi-tool-call turns), and Anthropic's user-first requirement holds.
- After compacting, a stored report is debited by the chars/4 estimate of what was dropped (an underestimate errs toward compacting again, never toward overflowing); the estimate path needs no bookkeeping — the next read re-estimates the compacted history.
- **History changes clear the report rather than patching it**: a rewind (`restartGeneration`, including the Retry path) or a failed turn that committed partial output sets `contextUsage = undefined`, and readers estimate what actually remains. This is what keeps Retry-after-context-length-error working: the rewound oversized history still projects high via the estimate, so the retry send compacts instead of failing identically. Rolled-back turns (cancelled before output) leave the report standing — the history reverted to exactly what it described.
- **Unknown context window → no compaction**: `getKnownModelContextWindow` returns `undefined` for unrecognized models and no limit is invented. Model windows live in an explicit ordered lookup table in `modelConfig.ts` (first `includes()` match wins). The conservative 128K default remains only for FIM autocomplete and script-context sizing via `getModelContextWindow`.
- Switching to a smaller-window model is handled naturally — the carried-over count immediately exceeds the new threshold and compacts on the next send.

**The indicator shows the same `contextTokens` the trigger uses.** Hidden below 50% of a known window; always shown for unknown windows once non-zero. Renders as a tiny right-aligned line directly above the chat input: `context window usage: ~120k / 1M` (no denominator for unknown windows). Exact (one turn stale) when the provider reports, a live estimate otherwise.

**Persistence**: only real reports are stored per chat in IndexedDB (estimates are recomputed from the loaded messages). Entries written by earlier builds of this branch may hold an object form; they are normalized to the token count on load.

## Test plan

- [x] Unit tests (`AIChatManager.test.ts`): compaction trigger mutates stored history and debits the report mid-turn; report rewrites the value after every send (including compacted ones); lazy estimate arms the trigger with no report at all (compaction off the estimate alone) and is exposed without being stored; report preferred over estimate once one arrives; rewind clears the report and falls back to estimating the remainder; no compaction under the trigger / without a known window; never drops the latest message; dangling tool-result handling; display-index re-basing; cleared on `saveAndClear`. Model-window table pinned in `lib.test.ts`; per-iteration usage reporting pinned in `chatLoop.test.ts`.
- [x] `npm run check` clean locally (0 errors)
- [x] Manually verified with Chrome DevTools MCP against the dev stack (claude-sonnet-4-6 via workspace Anthropic provider): send → provider report captured (~4.3k tokens incl. system prompt + tools) → persisted to IndexedDB as a plain number; no new console errors. Re-verified after merging main's interrupted-turn handling (#9539).
- [ ] `npm run check` + vitest in CI

Screenshot (bottom right, above the input): https://files.catbox.moe/meqwyu.png

## Follow-up (next stages of compaction)

1. LLM summarization compaction behind the same trigger (replacing drop-oldest)
2. Tool-result pruning of old turns before summarization
3. Optional: token columns in `ai_chat_usage` for workspace analytics


centdix:
> [P1] API-mode compaction omits lazily loaded tools

Acknowledged — this is a real gap on the first API-mode send: `projectedContextTokens` is computed before `loadApiTools()` populates the endpoint schemas, so the compaction projection can undercount on that path.

We're choosing not to fix it: API mode is slated for deprecation in the near term, so it's not worth investing in this edge case. The compaction projection remains correct for the default (non-API) chat path. Marking as won't-fix.

PR #9552, opened 2026-06-12
feat: automatic git-to-windmill sync (polling, webhooks, in-app PRs + checks)

hugocasa:
## Summary

Brings the repo → Windmill direction of git sync fully in-app, so the documented GitHub Actions become optional. Implements the phases of `docs/git-sync-pull-design.md`. Companion PRs: EE windmill-labs/windmill-ee-private#632, hub scripts windmill-labs/windmill-integrations#158 (P4 PR-head clone) and windmill-labs/windmill-integrations#163 (dev-workspace fork clone).

One HMAC-verified webhook receiver drives the pull direction: push to tracked branch → pull (+ live deploy check); push to `wm-fork/**` → pull into the matching fork workspace; `pull_request` → dry-run → check run. PRs for Windmill-pushed branches (`wm_deploy/**`, `wm-fork/**`) are opened by the deploy job's completion hook instead — outbound, so no webhook reachability needed.

## Phases

- **1 — polling.** EE-licensed, advisory-locked monitor task `ls-remote`s each auto-pull-enabled repo on its effective interval (~1 min, relaxed to ~10 min while a webhook is live) and enqueues a pull when the tracked branch moves (reusing the `{workspace_id}:git_sync` concurrency key so pulls serialize with pushes). Per-repo toggle + status in the UI.
- **2 — webhooks (managed app & GHES).** Enabling auto-pull creates a repository webhook via the installation token; the receiver (HMAC `X-Hub-Signature-256`, hook id from `X-GitHub-Hook-ID`) reconciles a pull instantly on tracked-branch push; `[WM]` commits skipped to avoid loops; disable deletes the hook. Polling stays on as a safety net. Host-aware, so **GHES self-managed-app repos are covered by the same per-repo path** — no GHES-specific code. The HMAC secret is encrypted at rest and never returned to clients.
- **3 — in-app PR creation (toggled, deploy-side).** When a deploy pushes a promotion (`wm_deploy/**`) or fork (`wm-fork/**`) branch and the repo opted in (*Open a pull request for each deploy branch* on the promotion repo; parent-level *Open a pull request when a fork deploys*), the push job's completion hook opens/reopens the PR with the installation token. Replaces the `gh pr create` actions (which stay valid — PR creation is idempotent) and works for unreachable instances.
- **4 — PR diff checks.** A `pull_request` against the tracked branch creates a "Windmill diff" check + enqueues a dry-run pull that clones the **PR head** (via the new `clone_ref` arg on the init-repository hub script, PR #158); the worker completion hook posts the diff (in-sync / N changes / failure). The completion hook also maintains **one managed comment on the PR** (Cloudflare deploy-preview style: workspace, status, commit, collapsible change list; upserted per synchronize), and words a settings difference by cause (the PR changes `wmill.yaml` vs pre-existing repo/workspace drift). Deploy-status check titles name the target workspace. No-op pushes (e.g. deploys caused by an auto-pull) report `pushed: false` from the push script and skip PR creation, so a closed PR isn't recreated by the sync loop (older script pins keep the previous always-ensure behavior). **Known issue:** the full-size push script's return value is currently dropped by the bun executor (job result stored as `null` even though `main` demonstrably returns `{pushed}` — reproduced with the released CLI and with an early-return `main`, while the same imports/signature in a smaller file capture fine; the pull script's results are captured). Until that executor quirk is root-caused, the missing flag makes the hook fall back to always-ensure — today's behavior, so nothing regresses, but the no-op gate is inert. Tracked as a follow-up. Token repos get hints pointing at the `open-pr-on-*` workflow fallbacks for the app-only PR features.
- **5 — fork sync (parent-level).** *Automatically deploy fork branches into forks* (`auto_pull.sync_forks`) routes changes on each fork's `wm-fork/<branch>/<id>` branch into that fork workspace — the managed `push-on-merge-to-forks`: via the parent's webhook, and via one extra fork-heads listing per poll tick (`ls-remote` pattern for token repos, `git/matching-refs` for app-backed), so it also works polling-only. Only live children of the parent can be routed to; forks carry a server-written status-only sync state and their card shows a read-only "managed in the parent workspace" line. Fork workspaces can't configure the parent-managed settings themselves (auto-pull, promotion mode, fork PRs are rejected server-side; the promotion card is hidden), and fork creation doesn't inherit the parent's `auto_pull`/`webhook_id`, so removing a fork can't delete the parent's hook. **Dev workspaces sync with their environment-label branch** (`dev`/`staging` verbatim, `push-on-merge-staging` parity) instead of the `wm-fork/**` form: the label rides the deploy job args (backend → hub script → CLI `--dev-workspace-label`), the PR hook derives the same head, routing/polling resolve label branches to the dev child, and the CLI refuses to deploy when the label equals the tracked branch. The label is immutable after create/attach (the branch is keyed on it; the set-label endpoint was removed). **Nested forks are covered**: routing resolves the whole live descendant chain of the webhook/poller workspace (recursive, depth-capped), so a fork *of a dev workspace* syncs through the root; such a fork roots its `wm-fork/**` branch on the dev's label branch and its PR merges back into it (`parent_dev_workspace_label` rides the deploy; `fork_open_prs` resolves at the root ancestor).
- **6 — live deploy status check.** A tracked-branch deploy posts an app-backed **"Windmill"** check that flips `Deploying… → Deployed N changes` on the head commit (Cloudflare-style). Reuses the Checks grant — no new permission.

Also in this PR: **CLI dev-workspace fork detection (fix).** `isForkWorkspace` / `computeGitSyncDeployBranch` keyed off the `wm-fork-` workspace-id prefix, so dev workspaces — forks with a custom, prefix-less id — never got their `wm-fork/**` branch derived or created. Now detected via `parent_workspace_id` too, mirroring the backend's `parent.is_some() || wm-fork- prefix` rule. Pairs with the hub-script clone-flag fix (windmill-integrations#163).

## Dropped vs. the original design

- **`fork_pull_sync` ("Keep forks in sync with the tracked branch")** — removed. Fanning the tracked branch straight into every fork was inconsistent with how a fork otherwise works (each fork syncs its *own* `wm-fork/**` branch, not `main`), and would clobber a dev workspace's local work. The consistent model — each fork/dev workspace syncs with its own branch — is what P5 above now implements (`sync_forks`).

- Release the CLI and bump the pinned `windmill-cli` in the hub script (windmill-integrations#163) and in the backend so the dev-workspace fork fix takes effect at runtime.

## Ops / release prerequisites

Add these to the `windmill-sync-helper` GitHub App (each re-prompts existing installs for org-admin approval; until granted, every feature degrades gracefully and polling covers the pull direction — **nothing existing breaks**; a failed PR creation is now also surfaced in the UI via the server-owned `open_pr_error` field on the repo settings, cleared on the next successful PR):

- **Repository webhooks: write** (P2, P4)
- **Pull requests: write** (P3, P5)
- **Checks: write** (P4, P6)
- Subscribe to events: **push**, **pull_request**

**Note to paste into GitHub's permission-update form** (shown to installation owners when the app requests the new permissions):

> Windmill can now handle the Git to Windmill direction of git sync itself. Previously this direction required GitHub Actions workflows in your repository:
>
> - **Repository webhooks (read & write)** — register a webhook per synced repository so commits deploy to your Windmill workspace instantly, replacing the push-to-Windmill workflow.
> - **Pull requests (read & write)** — open pull requests for branches Windmill itself pushes (promotion `wm_deploy/**` branches and workspace-fork `wm-fork/**` branches), replacing the `gh pr create` workflows, and maintain a deploy-preview comment on your pull requests.
> - **Checks (read & write)** — publish a "Windmill diff" check on pull requests showing what merging would deploy, and a deploy status check on synced commits.
>
> All of these are opt-in from your workspace settings. Nothing changes until you enable them: your existing sync and any GitHub Actions workflows keep working exactly as they do today.

**P4 additionally** needs the hub script published (windmill-integrations#158) and `GIT_SYNC_PULL_SCRIPT_PATH` bumped to the new `version_id`; until then the check reflects the tracked branch instead of the PR head. Self-managed/GHES apps get the same permissions via their setup checklist — no central approval.

<details><summary><b>Permission justification</b> (paste into the GitHub App settings)</summary>

Windmill deploys changes from your repository into your Windmill workspace without a GitHub Action. It needs:

- **Repository webhooks (write)** — register/remove its own push + PR webhook so merges to the tracked branch deploy instantly.
- **Pull requests (write)** — open/update the PR for promotion and fork deploys (replaces the `gh pr create` workflows).
- **Checks (write)** — post a diff preview on PRs and a deploy status (Deploying → Deployed) on commits.

Events: push, pull_request. All features are opt-in per workspace; existing sync and GitHub Actions workflows are unaffected until enabled.
</details>

## Validation

- `cargo check --features enterprise,private,license` passes (full binary); OSS/`all_sqlx_features` builds verified with `SQLX_OFFLINE=true`.
- `npm run check` passes (0 errors); sqlx offline cache regenerated.
- **Verified live** on an EE instance: polling, managed webhook auto-pull (with signature verification) + auto-mode polling fallback, PR diff checks (P4), the live deploy check (P6, `Deploying… → Deployed`), the self-managed (GHES-path) flow against github.com, fork sync end-to-end (external commits on a `wm-fork/**` branch and on a dev workspace's `staging` label branch → poller → pull applied in the right fork, own-deploy commits converge without loops), and the fork-deploy PR via the completion hook (deploy in a fork → PR opened on the shared repo). The promotion-PR head-branch derivation is unit-tested against the CLI formula (not exercised live). Server-side fork guards verified live (400s); the nested-fork routing queries are validated against real data (direct child, synthetic grandchild, root walk) but grandchild sync isn't exercised live. Production rollout still needs the app permission grants above. PRs from contributor forks clone the synthetic `pull/<n>/head` ref: the script mechanics are live-verified (a dry-run against a real PR's pull ref returns its diff) and nested grandchild fork sync is live-verified end to end (commit on a fork-of-a-fork's branch routed into it in seconds, parent/root untouched); the one unexercised path is GitHub's real cross-repo `pull_request` delivery (`head.repo` differing), which requires a fork owned by a second account.

**Inbound-webhook surface verified live via an ngrok tunnel against github.com** (real GitHub App, real deliveries): instant webhook pull (commit → pull job in ~3s, own-deploy no-op push leaves the branch head untouched); the `pull_request` flow end-to-end on a real PR — "Windmill diff" check on the head commit and the managed deploy-preview comment, updated in place across two `synchronize` pushes (always exactly one comment); the deploy status check with the workspace-named title (two workspaces syncing the same repo each post their own check, which is what the title disambiguates); and the failure path — a pull job forced to fail (broken subfolder) records `last_pull_status.success=false` and rolls `last_synced_sha` back so the commit is retried, then the next delivery recovers cleanly.

**Integration tests.** `integration_tests/test/git_sync_test.py` gains a `TestGitSyncAutoPull` class run by the existing Gitea-backed CI workflow: polling pull e2e (external commit → poller → workspace, no loop commit, pull state survives a config resave), fork-branch routing e2e (commit on `wm-fork/**` → fork workspace only), settings semantics (webhook→polling normalization persisted, secrets never exposed, legacy configs gain no `auto_pull`), fork guards (400s), and webhook-receiver rejection of unknown deliveries. The poller's SSRF guard gets an `ALLOW_LOCAL_GIT_REMOTES` escape hatch (set only by the CI workflow) since the tests' Gitea runs on localhost. Webhook delivery/PR checks need a real GitHub App and stay manual (covered above).

**CE gating.** Basic git sync stays available on CE (#8493); the auto-pull and fork-PR toggles are disabled with an `EE only` badge and new sync repos only default them on when licensed (the backend already rejects an enabled `auto_pull` on CE). The pull modal also passes `clone_ref` for `wm-fork-` forks (`wm-fork/<tracked>/<id>`) so a manual pull fetches the fork branch rather than the tracked-branch head.

## Compatibility with existing setups

Reviewed all three diffs (this PR, the EE companion, the hub-script PR) specifically for upgrade safety. Existing sync/promotion repos, forks, pinned hub scripts, workspace export/import, and CE basic git sync are unaffected: every new settings field is optional + serde-defaulted, PR creation and fork routing are gated behind flags that default to false and don't exist in stored configs, the webhook receiver is a new route that no-ops without a stored hook id + valid HMAC, and the completion hooks early-return on marker-less jobs (one PK SELECT each, errors swallowed). Old-CLI read-modify-write cycles are byte-identical for configs that never adopted auto-pull.

Two deliberate exceptions:
- `POST /w/{workspace}/workspaces/set_dev_workspace_label` (shipped v1.751.0) is removed: the environment label is now immutable after create/attach. No in-tree consumer remains.
- A fork or attached dev workspace that had a promotion-mode repo manually saved on it (previously accepted but non-functional) now gets a 400 on git-sync saves until the flag is cleared; the error message is actionable.

Cross-version note: an old CLI running `gitsync-settings push` against a workspace that adopted auto-pull replaces the config without the `auto_pull` field, which tears the webhook down cleanly (config gone, hook gone) — consistent replace semantics, not corruption.

## Screenshots

Workspace → Git sync settings. The repository card splits the two sync directions; the pull section has the auto-deploy and fork-sync toggles and the "remove any prior GitHub Action" reminder. App-backed repos always deliver via webhook with a polling fallback (no delivery selector); new connections default pull on only for app-backed repos. Fork workspaces get a read-only "managed in the parent workspace" line instead.

gitsync-autopull-card

gitsync-parent-card

Token-based repository: one info box recommending the GitHub App or the sync workflow for instant pull (and to keep only one of Action/auto-pull); the push section shows the workflow hint in place of the app-only fork-PR toggle:

gitsync-token-warning

Fork workspace (push/pull both name the fork branch):

gitsync-fork-card

Promotion workspace (push direction only, PR toggle):

gitsync-promo-card

PR-creation failure surfaced on the card (`open_pr_error`):

gitsync-pr-error

Auto-pull toggle on a CE (unlicensed) instance:

gitsync-ce-eeonly


hugocasa:
## Manual test checklist (before merge)

**Prereqs**
- [x] Managed app grants live: Repository webhooks / Pull requests / Checks = Read&write; events **push** + **pull_request**
- [x] Hub script (windmill-integrations#158) published + `GIT_SYNC_PULL_SCRIPT_PATH` bumped
- [x] Instance `base_url` is publicly reachable (webhook delivery)

**P1 — polling**
- [ ] PAT repo, auto-pull on → push to tracked branch → pulls within ~1 min; status chip updates
- [ ] Webhook-live repo polls at the relaxed (~10 min) interval, not every ~60s

**P2 — webhooks**
- [ ] App repo: enable auto-pull → webhook created on GitHub → push → instant pull
- [ ] Forged/unsigned delivery → 401
- [ ] Disable auto-pull → webhook deleted
- [ ] Unreachable instance → falls back to polling + UI shows the fallback notice
- [ ] `[WM]` commit does not re-trigger a pull (no loop)
- [ ] **GHES / self-managed app** path: same enable → webhook → pull

**P3 — in-app PR creation**
- [ ] Promotion deploy (`wm_deploy/**`) → PR opened targeting the tracked branch; re-deploy is idempotent

**P4 — PR diff check**
- [ ] PR against tracked branch → "Windmill diff" check shows in-sync / N changes / failure (needs the hub-script bump for PR-head clone)

**P5 — fork auto-sync (parent-level, app-backed)**
- [ ] Toggles show only on app-backed, non-fork parent
- [ ] "Open PRs for fork deploys" on → fork deploy opens a `wm-fork` PR; off → no PR
- [ ] "Keep forks in sync" on → push to parent tracked branch → each fork pulls (one job/fork, no double-pull)
- [ ] Fresh fork created **without** `auto_pull`/`webhook_id`; toggling/removing the fork does **not** delete the parent's webhook

**P6 — live deploy check**
- [ ] Push to tracked branch (app repo) → "Windmill" check flips Deploying… → Deployed N changes
- [ ] `[WM]`/bot/no-op commit → no spurious check; failed pull → check shows failure

**Cross-cutting**
- [ ] Existing repos load auto-pull **off** (no surprise auto-deploy on upgrade)
- [ ] New App repo defaults to auto/webhook; new PAT repo defaults to polling
- [ ] Sync jobs appear under "See sync jobs" (deploymentcallbacks)
- [ ] GET settings API response does **not** include `webhook_secret`
- [ ] Multi-repo: secondary repos stay push-only unless individually enabled

**Failure / lifecycle (reviewer-flagged)**
- [ ] A failing auto-pull job rolls `last_synced_sha` back + shows failure (commit retried, not shown synced)
- [ ] PR dry-run enqueue failure → "Windmill diff" check closes as failed (no hang)
- [ ] Deleting a git-sync repo with an active webhook removes the hook from GitHub
- [ ] Disabling auto-pull via edit removes the hook (and DB/GitHub stay consistent)

hugocasa:
Re: latest Codex review (fork sync requires explicit git resource branch) — **poller half confirmed and fixed, modal half outdated**:

- **Poller (fixed in b812cd3c56):** branch-less resources polled as the bare `HEAD` ref and the fork/dev-label fan-out skipped them silently. The poller now resolves the remote's default-branch name via `git ls-remote --symref HEAD` (same single remote call as the head-sha fetch); `"HEAD"` only remains if resolution itself fails. The polling e2e test now uses a **branch-less** resource, closing the fixture gap flagged in the review (helpers previously always set `branch: "main"`).
- **Pull modal:** no longer reproducible with the published pull script (hub `28787`): without a `clone_ref`, the script derives `wm-fork/<actually-cloned-default-branch>/<id>` and checks it out rooted on the existing remote branch, so a fork manual pull gets fork content, not the parent/default branch. The name also matches how the fork branch was created for branch-less resources (same derivation from the cloned default).

Compat note: previously-synced branch-less repos keyed `last_synced_sha` under `"HEAD"`; the key becomes the concrete branch name after upgrade, causing one no-op re-pull per repo on the first post-upgrade poll.

hugocasa:
Re: latest rounds —

**Codex P1 (runtime license bypass for in-app PR creation): confirmed and fixed in fdce2d40d3.** Both save paths now reject `promotion_open_prs`/`fork_open_prs` without an Enterprise plan (same gate as `auto_pull`), and the deploy completion hook re-checks the plan before `ensure_pull_request`, so flags stored while licensed stop driving GitHub calls after a lapse.

**Pi P2s — both previously dispositioned:**
- *Stale CLI guidance files*: false positive — `sync git-deploy` is a Cliffy `.hidden()` command and `system_prompts/generate.py` skips hidden commands by design (they're internal, invoked by the hub script); no generated file references `git-deploy`, so there is nothing to regenerate.
- *Fork manual pull without a stored branch*: covered by the published pull script (hub `28787`): without a `clone_ref` it derives `wm-fork/<actually-cloned-default-branch>/<id>` and checks it out rooted on the existing remote branch, which also matches how the branch was created for branch-less resources. Guessing `main` client-side would be strictly worse.

hugocasa:
Re: latest Codex P2 (fork manual pull `clone_ref`) — **disproven by a live probe against the published pull script (hub 28787)**, with evidence this time rather than reasoning:

Setup: a fork workspace whose fork branch (`wm-fork/main/<id>`) materially differs from `main`, plus a git resource with **no `branch` field**. Ran the pull dry-run with **no `clone_ref`** (exactly the flagged path).

Result: the script's fork handling checked out `wm-fork/main/<id>` rooted on `origin/wm-fork/main/<id>` (job log: `checkout -b wm-fork/main/gcd2e860 origin/wm-fork/main/gcd2e860`) and the dry-run returned `total: 0` — in sync with the fork's branch. A tracked-branch fallback would have reported 3 changes.

Why it works: `git_clone` derives the fork branch from the *actually cloned* default branch (`wm-fork/<cloned>/<id>`) — the same derivation branch creation used for a branch-less resource — and fork clones fetch all refs, so it roots on the existing remote branch.

On the fork-of-dev sub-claim: a fork of a dev workspace's branch is *named* `wm-fork/<trackedBranch>/<slug>` (the parent dev label only determines the branch's root commit at creation — `cli/src/commands/sync/sync.ts` checks out the label branch before creating it). So the modal's `wm-fork/<resource.branch>/<slug>` is the correct ref for that case too. The only path that roots on the wrong base is pulling before the fork branch exists remotely (a transient race at fork creation), where the preview degrades to tracked-branch content — same as plain forks, and there is nothing fork-specific to pull yet at that point.

hugocasa:
Re: latest Codex P1 (branch-less fork manual pull uses tracked branch, `PullWorkspaceModal.svelte:165`) — **same claim as the previous round's P2, already disproven by a live probe** (see the earlier disposition comment):

The frontend `clone_ref` is an override for the dev-label and explicit-branch cases, not the correctness mechanism. When no `clone_ref` is passed, the published pull script (hub 28787) fork block detects the fork workspace itself and derives `wm-fork/<branch>/<id>` from the **actually-cloned default branch**, then checks out `origin/<fork-branch>` before diffing.

Probe (exactly the flagged path): fork workspace whose fork branch materially diverged from `main`, git resource with **no `branch` field**, manual pull dry-run with **no `clone_ref`** → script checked out the fork branch, `diff total: 0` against the fork workspace. The parent branch is never applied.

hugocasa:
Re: latest Codex round (3 P1s) — **all three confirmed and fixed**:

1. **Slash-containing fork ids break auto-sync routing — confirmed, fixed in f1ea1b8294** by enforcing the invariant at the source instead of complicating the parser: `/` is now a rejected character in `validate_workspace_branch_id`, which guards every fork/dev creation and attach path. `parse_fork_branch`'s last-slash split (needed because the *base branch* legitimately contains `/`, e.g. `release/v2`) is now backed by a real invariant. A workspace id containing `/` was already pathological everywhere else (URL routing, CLI branch mapping); the object-store path mangling that defended against slash ids stays in place for any legacy id created before the ban.

2. **Unbounded git probes stall the poller — confirmed, fixed in 3888f2f0a4.** All three `git ls-remote` spawn sites in `resources.rs` (explicit-branch head, symref HEAD resolution, fork-branch listing) now go through `run_git_probe`: a 60s per-command deadline with `kill_on_drop`, so an unresponsive remote fails that one repository and the pass moves on, with no orphaned child process.

3. **Webhook reconcile clobbers concurrent writes — confirmed, fixed in 80703cbcd5.** `persist_reconciled_webhook_fields` no longer read-modify-writes the whole `git_sync` blob: it now issues one targeted `jsonb_set` per changed repo that replaces only the reconcile-owned fields (`webhook_id`/`webhook_secret`/`webhook_error`/`mode`), same pattern as the EE auto-pull status writer and `persist_git_sync_open_pr_error`. Concurrent poller status writes and settings saves are untouched; a repo whose `auto_pull` was concurrently removed is left alone.

PR #9553, opened 2026-06-12
fix: stop sending temperature for AI chat across all providers

centdix:
## Summary

Removes the `temperature` field from all AI chat completion requests and drops the model-based send-or-not special-casing.

Previously the chat sent `temperature: 0` for determinism and omitted it for reasoning models (`claude-opus-4-7/4-8`, `gpt-5+`, `o-series`) that reject sampling params. That required hand-maintaining a growing list of reasoning models — every new one rejects sampling params and ships its own default, so the list would keep expanding. Since `temperature: 0` only ever bought determinism on older non-reasoning models (low value for an interactive, tool-using chat), we drop it entirely and never send temperature.

The model-detection helper is **kept** (renamed `modelDisallowsSamplingParams` → `requiresMaxCompletionTokens`) because it still serves a separate, unrelated concern: choosing `max_completion_tokens` over `max_tokens` for OpenAI/Azure reasoning models on the Chat Completions API.

## Changes

- `modelConfig.ts`: delete `getDefaultChatTemperature`; rename `modelDisallowsSamplingParams` → `requiresMaxCompletionTokens` (now solely the max-tokens-param decision); rewrite the comment; drop the now-unused `AIProviderModel` import.
- `lib.ts`: remove the `temperature` field from `getModelSpecificConfig`; update the import and call site.
- `chat/anthropic.ts`: remove the `temperature` pass-through from `anthropicParams`.
- `chat/openai-responses.ts`: remove the `temperature` pass-through from `convertCompletionConfigToResponsesConfig`.
- `lib.test.ts`: remove the `getDefaultChatTemperature` tests; keep the model-detection tests, renamed to `requiresMaxCompletionTokens`.

**Out of scope (intentionally untouched):** the FIM autocomplete `temperature: 0` in `getFimCompletion` — that is the code-autocomplete path, not the chat, and those models accept temperature.

## Behavior note

This is a behavior change for classic non-reasoning models (gpt-4o, sonnet, mistral): they previously received `temperature: 0` for determinism and now run at the provider default. Outputs become marginally less reproducible — the intended trade-off.

## Test plan

- [ ] `npm run check` passes
- [ ] `vitest` passes for `lib.test.ts` (renamed `requiresMaxCompletionTokens` assertions)
- [ ] AI chat returns a normal completion against one model per transport: an Anthropic model, a standard OpenAI Chat Completions model, an OpenAI reasoning model via the Responses API (e.g. `o3`/`gpt-5`), and a non-OpenAI provider (e.g. Mistral/OpenRouter)
- [ ] Network inspection: no `temperature` field in outgoing chat request bodies; FIM autocomplete still sends `temperature: 0`


centdix:
Addressed the P2 in f8a25186a8.

Dropped the three dead `claude-*` branches from `requiresMaxCompletionTokens` (and their stale detection tests). Verified the cleanup is behaviorally inert:

- The function's only non-test caller is `lib.ts:320`, gated on `provider === 'openai' || 'azure_openai'` — Claude ids never reach it.
- Claude reasoning behavior is owned by `reasoningRegistry.ts` (`/claude-opus-4-(7|8)/`, `fable`), not this function.

So `requiresMaxCompletionTokens` now matches its name/comment exactly (gpt-5+ / o-series → `max_completion_tokens`), and the remaining tests are all semantically true.

The earlier branch was also rebased onto latest `main` (merge commit 3cbb1d495f), integrating main's reasoning-effort refactor; `applyReasoningToConfig` already strips sampling params defensively, so the temperature removal composes cleanly with it.

PR #9558, opened 2026-06-12
[ee] feat: make oauth client credentials fully resource-level

hugocasa:
## Summary

OAuth client-credentials (CC) connections no longer require any instance-level setup. Whether a resource type supports CC is now derived from its own schema: when it carries unambiguous client-id/client-secret fields plus a `token` field (normalized name matching, so `client_id`, `clientId` and `azureClientId` all qualify), the connect dialog offers a "client credentials" mode. In that mode the `token` field is hidden and platform-managed (acquired and auto-refreshed), and the user fills the client id/secret and the provider's token URL instead.

Everything needed for refresh is persisted on the `account` row (`cc_client_id`, `cc_client_secret`, `cc_token_url`, scopes), so the refresh worker is self-contained and works for providers with no instance OAuth entry at all. The credentials are also stored on the resource itself — client id in plain, client secret as a linked secret variable — so scripts can use them directly. This mirrors how n8n models CC credentials.

The per-provider "Support Client Credentials Flow" toggle in instance settings is removed along with the `grant_types` plumbing: it gated nothing (users always supplied their own credentials) while forcing admins to create otherwise-unused `oauths` entries. Existing stored settings that still contain `grant_types` keys parse fine. The CC connect endpoint moves under the workspaced router (`/api/w/{workspace}/oauth/connect_client_credentials/{client}`) for consistency with the rest of the connect/account API.

Companion EE PR: windmill-labs/windmill-ee-private#612

## Changes

- `windmill-oauth`: `build_client_credentials_oauth_client` resolves provider config as instance entry → static registry → caller-supplied token URL (errors only when none yields a token URL; invalid instance entries log a warning and fall through) and returns the resolved `OAuthConfig`; `refresh_token_for_account` no longer requires the provider in `OAUTH_CLIENTS.connects` for CC accounts; `grant_types` removed from `OAuthClient`/`ClientWithScopes`/`OAuthConfig`
- `windmill-common/instance_config.rs`: `grant_types` removed from the typed settings mirror (the flattened `extra` map keeps old configs round-tripping)
- EE `oauth2_ee.rs`: CC connect endpoint moved to `workspaced_service()`; `grant_types` gate removed; `create_account` requires `cc_client_id`/`cc_client_secret`/`cc_token_url` for `client_credentials` accounts so rows are self-describing
- `openapi.yaml` + regenerated client: endpoint path moved, `grant_types` dropped from `get_connect`
- `app_connect.ts`: new `detectClientCredentials(schema)` helper (exactly one client-id-ish and one client-secret-ish string field + a `token` field; optional token-URL-ish field; ambiguous schemas never match) with unit tests
- `AppConnectInner.svelte`: CC availability from the resource-type schema; CC-only types get the OAuth path even when the provider has no instance OAuth client; token URL is a required input prefilled from the static registry; manual form offers "Acquire the token automatically via client credentials instead"; CC inputs reset on resource-type switch
- `AuthSettings.svelte`: instance-level CC toggle removed

Known behavior change: builtin OAuth types whose schemas are `{token}`-only (e.g. visma) stop offering *new* CC connections until their hub schemas gain credential fields — already-connected accounts keep refreshing via the instance-entry/registry fallback. Follow-up: add the credential fields to the relevant hub resource types.

## Test plan

- [x] `cargo check --features enterprise,private` clean; `windmill-oauth` (8) and `windmill-common` (69) unit tests pass
- [x] `npm run check` 0 errors; new `app_connect.test.ts` (7 tests) passes
- [x] E2E against a mock token endpoint: custom resource type with `client_id`/`client_secret`/`token_url`/`token` fields → connect dialog offers CC with zero instance OAuth config → exchange succeeds → resource saved as `{token: $var:…, client_id, client_secret: $var:…_client_secret, token_url}` with a workspace `account` row
- [x] Forced refresh (`POST /oauth/refresh_token/{id}`) re-exchanges against the account-stored token URL with empty instance settings
- [x] Legacy path: account with NULL `cc_token_url` + provider configured in instance settings refreshes via the instance `connect_config` fallback
- [x] Ambiguous schema (two client-id-ish fields) shows no CC option
- [x] Old `/api/oauth/connect_client_credentials/{client}` returns 404; new workspaced route returns 401 unauthenticated / 200 authenticated
- [x] Instance settings → SSO/OAuth no longer shows the CC toggle
- [ ] Authorization-code popup flow against a real provider (untouched code paths; not exercisable on this dev instance — no OAuth app registered)

## Screenshots

Manual resource form for a CC-capable type, with the new switch to the automated flow:

cc-manual-with-switch

Client-credentials connect form (CC-only provider — no instance OAuth client configured):

cc-connect-form

Save step after a successful exchange:

cc-save-step

---
## Summary by cubic
Make OAuth client-credentials fully resource-level. CC availability now comes from the resource schema, removing the need for instance-level setup, and tokens refresh using only account-stored data.

- New Features
  - Auto-detect CC support from a resource’s schema (exactly one client-id field, one client-secret field, and a token field); implemented in `detectClientCredentials`.
  - Connect UI offers a “Client Credentials” mode even without an instance OAuth client; requires a token URL (prefilled from the static registry when available), stores client id on the resource and client secret as a secret variable, and manages the token automatically.
  - Backend: `build_client_credentials_oauth_client` resolves config from instance → static registry → provided token URL; `refresh_token_for_account` refreshes CC accounts using `cc_client_id`, `cc_client_secret`, `cc_token_url`, and scopes without needing `connects`.
  - API: CC connect endpoint moved to `/w/{workspace}/oauth/connect_client_credentials/{client}` with required `cc_client_id`, `cc_client_secret`, and `cc_token_url`.

- Migration
  - Removed the instance-level CC toggle and `grant_types` from `OAuthClient`/`OAuthConfig` in `windmill-common` and `windmill-oauth`; old configs still parse but the field is ignored.
  - Update clients to the new workspaced route and include `workspace`; CC connect requests must send `cc_client_id`, `cc_client_secret`, and `cc_token_url`.
  - Built-in token-only resource types won’t offer new CC connections until their schemas define credential fields; existing CC accounts keep refreshing via the fallback.
  - Frontend settings no longer show the CC toggle; the connect dialog handles CC at the resource level.

<sup>Written for commit 45d9ff3534a7bc18107a1a0ca6da7b77a2da3803. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Superseded by a minimal rework: CC enablement moves to the static registry (plus the existing instance-settings path for custom providers) instead of resource-type field detection. New PR to follow on this branch.

PR #9570, opened 2026-06-14
fix: require pinned sha for inline raw_code in viewer app run mode

rubenfiszel:
## Summary

Closes an authorization / content-hash-pin bypass in the app `execute_component` run-mode path (private advisory GHSA-qc46-mqh5-9f3p, an incomplete-fix residual of CVE-2026-22683).

In run mode (`force_viewer_static_fields` omitted), caller-supplied inline `raw_code` (no `app_script` id) against a **Viewer-mode** app fell back to the default triggerable via `ExecutionMode::Viewer => Some(&policy_triggerables_default)`, short-circuiting the `triggerables_v2` content-hash pin. Effect: a caller could submit arbitrary code into someone else's deployed Viewer app and have it run as themselves. The notable case is an **operator** — operators are otherwise hard-blocked from running arbitrary code (`/jobs/run/preview` → "Operators cannot run preview jobs") — gaining inline code execution.

Severity in practice is **at most medium**: the job runs as the *caller's own* identity (no publisher/admin escalation within Windmill's permission model), and the non-operator leg is not a new capability (members can already run their own code via `/jobs/run/preview`). The teeth are only at the worker-OS layer on default CE (`isolation=none`, root), which nsjail + non-root workers mitigate.

## Changes

- `backend/windmill-api/src/apps.rs`: compute `is_inline_raw_code = payload.raw_code.is_some() && payload.id.is_none()` and deny the `ExecutionMode::Viewer` default-triggerable fallback for that case, so caller-supplied inline code must match a publisher-pinned `rawscript/<sha>` entry. Deployed `script`/`flow` paths, `app_script`-backed inline scripts (`id`), and legacy pinned inline scripts are unaffected (they resolve before the fallback).
- `backend/tests/app_preview_auth.rs` + fixture: deployed empty Viewer app (`999003`) and two regression cases — operator and non-operator member posting inline `raw_code` in run mode, both now `400 ... forbidden by policy`.

## Test plan

- [ ] `cargo test --test app_preview_auth` passes (operator and member run-mode `raw_code` → 400; existing preview/run-mode cases still green)
- [ ] A real deployed Viewer app still executes its legitimately-pinned components (deployed `script`/`flow` refs and `app_script`-backed inline scripts)
- [ ] A hand-crafted `execute_component` run-mode request with unpinned inline `raw_code` returns 400

---
## Summary by cubic
Require a pinned content hash for all inline `raw_code` in Viewer app run mode (with or without `app_script` id), closing a fallback that let unpinned code run via the default triggerable. This blocks arbitrary code injection into deployed Viewer apps; only publisher-pinned `rawscript/<sha>` entries are allowed.

- **Bug Fixes**
  - In `execute_component`, disable the Viewer default-triggerable fallback when `raw_code` is present; require a `triggerables_v2` pin for the resolved `rawscript/<sha>` (covers plain inline and `app_script`-backed code).
  - Allowed paths remain: deployed `script`/`flow` refs; unpinned inline code now returns 400.
  - Fixed OSS build by gating `rd_string` behind the `parquet` feature; added regression tests and a minimal Viewer app fixture (operator/member inline runs, with/without `id`, now return 400).

<sup>Written for commit 664a61f41f818528468551451cadd1fe7cdb627f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/app_preview_auth.rs
+++ b/backend/tests/app_preview_auth.rs
@@ -259,13 +259,12 @@ async fn test_app_preview_authorization(db: Pool<Postgres>) -> anyhow::Result<()
         "rejection must be the jobs:run scope gate, got: {body}"
     );
 
-    // 9. RUN-MODE REGRESSION (CVE-2026-22683 residual): against a *deployed*
-    //    Viewer-mode app, run mode (no `force_viewer_static_fields`) must NOT let
-    //    a caller substitute arbitrary inline `raw_code`. Pre-fix the
-    //    `rawscript/<sha>` lookup fell back to the Viewer default triggerable, so
-    //    the operator's code was enqueued and ran as them (200 + job UUID, the
-    //    exact preview class blocked in step 1). Post-fix the unpinned sha is
-    //    rejected by the policy.
+    // 9. RUN-MODE REGRESSION (CVE-2026-22683 residual): in run mode (no
+    //    `force_viewer_static_fields`) against a deployed Viewer-mode app,
+    //    caller-supplied inline `raw_code` whose `rawscript/<sha>` is not pinned
+    //    in the app's `triggerables_v2` must be rejected by the policy — it must
+    //    not resolve via the Viewer default triggerable and run as the caller
+    //    (the same preview-class execution an operator is denied in step 1).
     let run_mode_raw_code = json!({
         "args": {},
         "component": "comp",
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Addressed the review findings in 664a61f:

- **Codex P1 (raw_code + app_script `id` still bypassed the pin)** — valid completeness gap, fixed. The `id` path resolves `rawscript/<code_sha256>` from `app_script` **by id with no app/workspace scoping**, so an unpinned script from another app could still run against a Viewer app via the fallback. Broadened the gate from `raw_code.is_some() && id.is_none()` to `raw_code.is_some()`, so caller-supplied inline code requires a publisher-pinned `rawscript/<sha>` regardless of `id`. (Pi was right that this isn't *arbitrary code* injection — `content` is dropped and the worker fetches by id — but the integrity break of running an unpinned app_script stood, so it's now closed.) Added regression case 11: run-mode `raw_code` + `id: 999777` (an app_script owned by a different app) → `400 forbidden by policy`.
- **P2 (Pre-fix/Post-fix narration in test comments)** — fixed; rewritten as current invariants per AGENTS.md.
- Safe by construction: the gate only affects the `ExecutionMode::Viewer` arm, and legitimately-runnable inline scripts resolve via `triggerables_v2.get(...)` before the fallback. (Separately confirmed deployed apps are never persisted in Viewer mode — the deploy UI only emits `publisher`/`anonymous` — so this fallback is effectively dead for real apps anyway.)

`cargo test --test app_preview_auth` passes (cases 9/10/11 all assert `400 forbidden by policy`). The `windmill-object-store` commit is an unrelated pre-existing OSS-build fix (`rd_string` unused import under `-D warnings`).

PR #9593, opened 2026-06-15
fix(cli): preserve committed script.lock on transient NULL lock during git-sync deploy

### LINKED ISSUE #9588: Git sync deploy mirror deletes a script's committed .script.lock when the deploy races a pending dependency relock (lock briefly NULL)
## Summary

When a script is deployed via `wmill sync push` **together with a change to one of its relative imports**, the workspace Git sync's deploy-event mirror commit can **delete the script's committed** `.script.lock` on the target branch. This happens because the script's `lock` is transiently `NULL` while the server's importer-relock dependency job is still running, and the git-sync job reads it inside that window.

## Environment

* Windmill **v1.722.0**, self-hosted (Community Edition)
* Workspace Git sync configured against a GitHub repo, with `.script.lock` files committed and `script` in `include_type`
* Scripts deployed from git via CI running `wmill sync push` (CLI v1.722.0)
* The deploy event triggers the default hub sync script (`wmill sync git-deploy`)

## Steps to reproduce

1. Commit script `A.ts` (with a committed `A.script.lock`) and a relative-import module `B.ts` that `A` imports.
2. In one push, change `A.ts` **and** `B.ts` such that `A`'s lock must be regenerated.
3. Run `wmill sync push` to deploy both.
4. Observe the Git-sync deploy-event mirror commit on the target branch.

## What happens

The server queues importer-relock dependency jobs (because `B` changed), which hold `A`'s `lock` at `NULL` until they finish. The deploy event's git-sync job reads `A` inside that window, sees no lock, and pushes a mirror commit that **deletes** `A.script.lock` and drops the `lock: '!inline …'` line from `A.script.yaml`. Seconds later the dependency job writes the final lock server-side — byte-for-byte identical to the one whose deletion was just committed — so the only lasting damage is in the git mirror.

### Timeline from one real occurrence (job logs, times UTC)

| time | event |
| -- | -- |
| 21:59:33–34 | importer-relock dependency jobs queued by the deploy |
| 21:59:39.0 | git-sync job logs `Deleting script f/agents/<script>.script.lock` |
| 21:59:39.96 | dependency job writes the final (identical) lock on the server |
| 21:59:45 | mirror commit containing the lock deletion pushed to the branch |

The ~1s overlap between the git-sync read and the dependency job's lock write is the race.

## Impact

For repos that treat committed locks as the source of truth, the sync's own mirror commit deletes lock files on the default branch. Any CI that validates lock presence then fails for every subsequent PR and deploy until the lock is restored. We saw this fire **4 times in a single day**, once for each script that happened to change alongside a shared relative import.

## Expected behavior

The git-sync deploy job should treat a `NULL` lock with a pending dependency job for that script as **lock-pending** — wait for the dependency job, re-read after it settles, or skip emitting the lock change — rather than mirroring the transient `NULL` as a deletion of the committed lock.

## Workaround (for others hitting this)

Two mitigations, either of which stops the bleeding:

1. A CI-side guard that detects deploy-event mirror commits deleting lock files and restores them (shape-gated: only restore while the owning script still exists).
2. Removing `script`/`flow` from the workspace Git sync `include_type` when CI (git → server) is already the source of truth for those types — this disables the deploy-echo mirror commits entirely. UI-side edits are still recoverable via a separate live-drift pull. (This is the route we took.)

## Possibly related

* [#9403](<https://github.com/windmill-labs/windmill/issues/9403>) — deploy include flags clobbered for the same job class
* windmill-labs/windmill#9527 — lock cache content-blindness (you fixed this in v1.722.0; thanks!)

rubenfiszel:
Fixes #9588
Fixes WIN-2050

## Problem

When a script is deployed via the workspace Git sync **deploy mirror** while one of its relative-import dependencies also changed in the same push, the mirror commit can **delete the script's committed `.script.lock`** and strip the `lock: '!inline …'` line from `*.script.yaml`.

Root cause is a race: the server holds the script's `lock` at `NULL` while the importer-relock dependency job is still running. The git-sync deploy job reads the workspace inside that ~1s window, sees no lock, and mirrors the transient `NULL` as a deletion. Seconds later the dependency job writes the final (byte-identical) lock server-side — so the only lasting damage is in the git mirror. For repos that treat committed locks as the source of truth, this breaks every subsequent CI run until the lock is restored.

## Fix

`NULL` lock means **"lock pending"**, not **"this script has no lock"** — a genuinely lock-free script (no dependencies, or a codebase script) serializes `lock: ''` (empty string), which keeps its metadata key.

During a pull (remote → local, which is what the deploy mirror runs), `preservePendingScriptLocks` carries the local committed lock onto the remote map whenever the remote reports a `NULL`/absent lock for a script whose committed `.script.lock` still exists locally. The diff then becomes a no-op for **both** the lock file and the metadata `lock` line, so the mirror commit no longer deletes it.

An empty-string lock (`''`) — the real "dependencies removed" transition — is left untouched, so genuine lock removals still propagate. Scripts deleted remotely are not affected (no remote metadata key → no preservation). The guard is gated to the pull direction only (`isEls1Remote === true`); push is unchanged.

## Testing

- New deterministic unit test `cli/test/pending_script_lock_unit.test.ts` (4 cases): NULL lock preserved, empty-string lock still deleted, NULL with no local lock file is a no-op, deleted remote script does not resurrect its lock — all pass.
- Existing `git_unit`, `wmill_lock_unit`, `lint_locks_unit`, `lockfile_format_pinning_unit`, `gitsync_converter_unit` suites pass (88 tests).
- `tsc --noEmit`: no new type errors (16 pre-existing, unchanged).

CLI-only change; no backend or EE files touched.

---
## Summary by cubic
Prevents git-sync deploy from deleting committed `.script.lock` and stripping `lock` from `*.script.yaml` when the server reports a transient NULL lock during relock. Keeps mirrors stable and unblocks CI; aligns with WIN-2050.

- Bug Fixes
  - On pull (remote → local), preserve the committed lock and metadata when the remote lock is NULL, so the diff is a no-op; supports `__mod` multi-module layout and Windows path separators.
  - Real removals still propagate: `lock: ''` continues to delete the lock; script deletions are unchanged.

<sup>Written for commit 15030c39f3a7a83801a8365d7ba9cc922b8cbfdb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/src/commands/sync/sync.ts
+++ b/cli/src/commands/sync/sync.ts
@@ -2097,9 +2097,14 @@ export function preservePendingScriptLocks(
   remote: Record<string, string>,
   local: Record<string, string>,
 ): void {
+  // A multi-module script keeps its metadata in the folder layout
+  // `…__mod/script.{yaml,json}` instead of `….script.{yaml,json}`.
+  const modMeta = getModuleFolderSuffix() + SEP + "script";
   for (const metaKey of Object.keys(remote)) {
-    const isYaml = metaKey.endsWith(".script.yaml");
-    const isJson = metaKey.endsWith(".script.json");
+    const isYaml =
+      metaKey.endsWith(".script.yaml") || metaKey.endsWith(modMeta + ".yaml");
+    const isJson =
+      metaKey.endsWith(".script.json") || metaKey.endsWith(modMeta + ".json");
     if (!isYaml && !isJson) continue;
 
     const localMeta = local[metaKey];
@@ -2125,9 +2130,13 @@ export function preservePendingScriptLocks(
 
     // The local side must reference an inline lock backed by a committed file.
     const localLock = localParsed["lock"];
-    if (typeof localLock !== "string" || !localLock.startsWith("!inline")) continue;
+    if (typeof localLock !== "string" || !localLock.startsWith("!inline ")) continue;
 
-    const lockKey = metaKey.replace(/\.(yaml|json)$/, ".lock");
+    // Derive the lock-file key from the `!inline` reference itself, not from the
+    // metadata path: a multi-module script keeps its lock at `…__mod/script.lock`,
+    // which a `.script.yaml -> .script.lock` rewrite would miss. The reference is
+    // always forward-slash; map keys use the OS separator.
+    const lockKey = localLock.slice("!inline ".length).replaceAll("/", SEP);
     if (local[lockKey] === undefined) continue; // committed lock already gone
 
     remoteParsed["lock"] = localLock;
--- a/cli/src/commands/sync/sync.ts
+++ b/cli/src/commands/sync/sync.ts
@@ -2077,6 +2077,67 @@ export async function isCaseInsensitiveFilesystem(
   return result;
 }
 
+// A script's `lock` is NULL on the remote only while the server is still
+// (re)generating it — e.g. an importer relock is in flight after a dependent
+// relative-import module changed, or the script's own first lock job has not
+// settled yet. NULL means "lock pending", NOT "this script has no lock": a
+// genuinely lock-free script (no dependencies, or a codebase script) serializes
+// `lock: ''` (empty string), which keeps its metadata key. The git-sync deploy
+// mirror reads the workspace inside that window and would otherwise mirror the
+// transient NULL as a deletion of the committed `.script.lock` (and strip the
+// `lock: '!inline …'` line from the metadata), corrupting the mirror until the
+// relock writes the identical lock back seconds later (issue #9588).
+//
+// When pulling (remote -> local), if the remote reports a pending (NULL) lock
+// for a script whose committed lock still exists locally, carry the local lock
+// onto the remote map so the diff is a no-op for both the lock file and the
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Addressed the auto-review nit in 15030c39f3: `preservePendingScriptLocks` now derives the lock-file key from the committed `!inline` reference instead of rewriting the metadata path, so multi-module `…__mod/script.lock` scripts are covered too. I also broadened the metadata detection to the `__mod/script.{yaml,json}` folder layout, and convert the (always forward-slash) reference to the OS separator so the local-map lookup still matches on Windows. Added a `__mod` unit case (now 5 tests).

PR #9351, opened 2026-05-27
Db-backed user drafts

diegoimbert:

PR #9599, opened 2026-06-15
fix(frontend): strip server-managed fields from value diffs

Guilhem-lm:
## Summary

The script editor's **Deployed ⟷ Current** diff (Metadata tab) showed a lot of spurious changes the user never made — `created_at`, `created_by`, `extra_perms`, a large multi-line `lock_error_logs` dump, the `lock` blob, and draft-overlay bookkeeping — making the diff effectively unusable for spotting real metadata changes.

**Root cause.** Since #9351 (DB-backed user drafts), the script-edit loader fetches the full `Script` DB row (+ draft overlay) via `getScriptByPath` instead of the trimmed `NewScript`-shaped `getScriptByPathWithDraft`. The editing object (`current` side of the diff) therefore carries server-managed bookkeeping fields that `NewScript` never had. The deployed side is trimmed in `ScriptBuilder.syncWithDeployed`, so the two sides no longer line up and every bookkeeping field renders as a diff. `cleanValueProperties` — the shared normalizer both sides pass through — only stripped `parent_hash`/`draft`/`draft_only`, so it didn't catch them.

## Fix

Two complementary layers:

**1. Pure-noise fields → stripped in the shared `cleanValueProperties`.** These are never user-editable and meaningless in *any* diff context, so stripping them globally is safe and also fixes the draft⟷current diff, `hasUnsavedChanges`, and flow/app diff viewers:
`created_at`, `created_by`, `workspace_id`, `parent_hashes`, `lock_error_logs`, and the draft-overlay bookkeeping `draft_saved_at`, `draft_created_at`, `is_draft`, `other_drafts_users`.

**2. Context-meaningful fields → stripped at the script-editor source, kept in the shared helper.** `lock` and `extra_perms` are legitimate, user-meaningful fields in *some* diff contexts:
- `lock` — version-to-version diffs (`WorkspaceItemDiffViewer`) surface lockfile changes.
- `extra_perms` — folder sharing-permission changes in workspace/fork diffs (the backend's `compare_two_folders` treats folder `extra_perms` as a real change; `getItemValue` includes it for folders).

So they stay in `cleanValueProperties`, and the script-editor noise they would otherwise cause is removed at the source: `ScriptBuilder.openDiffDrawer` nulls `lock`/`extra_perms` on the `current` side, mirroring the existing deployed-side strip in `syncWithDeployed`.

## Changes

- `frontend/src/lib/utils.ts`: extract the skip-list into `CLEANED_VALUE_KEYS`; add the pure-noise fields above. `lock`/`extra_perms` deliberately excluded, with a comment recording why.
- `frontend/src/lib/components/ScriptBuilder.svelte`: `openDiffDrawer` nulls `lock`/`extra_perms` on the `current` side to match the deployed side.
- `frontend/src/lib/utils.test.ts`: unit tests for `cleanValueProperties` — each noise key stripped, user-editable keys + `lock` + `extra_perms` preserved, non-object passthrough, no input mutation.

## Screenshots

After the fix — editing the summary of a deployed script shows only the real change in the Metadata tab, with no `lock`/`extra_perms`/`created_at`/`draft_saved_at` noise:

diff-metadata-summary-only

## Test plan

- [x] `npm run check:fast` — no type errors
- [x] `vitest run src/lib/utils.test.ts` — 62 passing (incl. new `cleanValueProperties` cases)
- [x] Deployed script → open Diff → Metadata: no edits shows "No changes detected"; summary edit shows only that change (verified in browser)
- [ ] Folder workspace/fork diff where only `extra_perms` changes still surfaces the change (needs EE two-workspace setup; covered by unit test asserting `extra_perms` is preserved)

### DIFF
--- a/frontend/src/lib/components/ScriptBuilder.svelte
+++ b/frontend/src/lib/components/ScriptBuilder.svelte
@@ -707,7 +707,16 @@
 		const currentDraftTriggers = structuredClone(triggersState.getDraftTriggersSnapshot())
 
 		const deployed = deployedValue ?? savedScript
-		const current = { ...script, draft_triggers: currentDraftTriggers }
+		// `script` comes from the full DB row (since #9351), so it carries `lock`/`extra_perms`
+		// while the deployed side is trimmed in `syncWithDeployed` — null them here to match
+		// that strip. They stay in the shared `cleanValueProperties` so version-to-version and
+		// folder workspace/fork diffs still surface lockfile / sharing-permission changes.
+		const current = {
+			...script,
+			lock: undefined,
+			extra_perms: undefined,
+			draft_triggers: currentDraftTriggers
+		}
 		if (current.assets && !current.assets.length) delete current.assets
 
 		diffDrawer?.openDrawer()
--- a/frontend/src/lib/utils.test.ts
+++ b/frontend/src/lib/utils.test.ts
@@ -289,9 +289,12 @@ describe('cleanValueProperties', () => {
 		'parent_hash',
 		'draft',
 		'draft_only',
+		'draft_saved_at',
+		'draft_created_at',
+		'is_draft',
+		'other_drafts_users',
 		'created_at',
 		'created_by',
-		'extra_perms',
 		'workspace_id',
 		'parent_hashes',
 		'lock_error_logs'
@@ -315,8 +318,7 @@ describe('cleanValueProperties', () => {
 			content: 'export function main() {}',
 			schema: { properties: { x: { type: 'string' } } },
 			language: 'bun',
-			created_at: '2024-01-01',
-			extra_perms: { 'u/foo': true }
+			created_at: '2024-01-01'
 		}
 		const cleaned = cleanValueProperties(input) as any
 		expect(cleaned.summary).toBe('my script')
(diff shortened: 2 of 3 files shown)


Guilhem-lm:
Thanks — addressed the `extra_perms` finding (Codex P1 / Pi P2) in `8a2e93da77`.

**`extra_perms` is no longer stripped in the shared `cleanValueProperties`**, so folder sharing-permission changes surface again in workspace/fork diffs (`compare_two_folders` / `getItemValue` for folders). Instead, the script-editor noise it caused is stripped at the source: `ScriptBuilder.openDiffDrawer` now nulls `extra_perms` on the `current` side, mirroring the existing deployed-side strip in `syncWithDeployed`.

Same treatment applied to `lock` (kept in the shared helper for version-to-version diffs, nulled on the script-editor `current` side) — browser-testing the editor diff surfaced that `lock` had the identical asymmetry once it left the global skip-list.

Two more spurious fields turned up during browser verification and are now stripped as pure noise in `cleanValueProperties`: the draft-overlay bookkeeping `draft_saved_at`, `draft_created_at`, `is_draft`, `other_drafts_users` (carried by the full DB row, meaningless in any diff).

Verified in a real browser: no-edit diff → "No changes detected"; a summary edit → only that change, no `lock`/`extra_perms`/`draft_*` noise (screenshot in the PR description). Unit tests updated to assert `lock` and `extra_perms` are preserved by the shared helper while the noise keys are stripped — 62 passing.

PR #9600, opened 2026-06-15
fix: db-backed draft fixes — review-page UX, legacy drafts, session restore

diegoimbert:
## Summary

Follow-up fixes for the DB-backed user drafts feature (#9351): correct several review/deploy-page and session-pane behaviors, surface legacy (pre-migration, workspace-level) drafts, and stop migrated draft-only items from flooding the home list.

## Changes

**Session panes** (`sessions/*`)
- Seed the per-tab `last_sync` from the server draft's `draft_saved_at` in `loadFlow`/`loadScript`'s "no local draft" branches (mirroring `loadRawApp`), so the seeding save attaches a matching `last_sync` and the server no longer clobbers an existing server draft with a fresh `created_at`.
- Replace the no-op `loadFlow`/`loadRawApp`-based diff-drawer restore handlers with proper `restoreDeployed`/`restoreDraft` that reset the live `UserDraft` cell (inbound sync updates the preview) and delete the per-user server draft — mirroring `ScriptEditorView`. Adds `rawAppValueToDraft` to project a deployed raw-app value into the draft shape.

**Review & deploy page** (`CompareDrafts`)
- Show the friendly `draft_path` instead of the `u/{user}/draft_{uuid}` storage path. `list_drafts` now surfaces `draft_path` (when set and ≠ storage path), like the home-page list endpoints; storage path stays the key for all fetch/deploy/discard calls.
- Delete the storage-path draft for every kind after a successful deploy, so a renamed `draft_only` item doesn't leave its `u/{user}/draft_{uuid}` draft behind.
- Badge legacy drafts: `list_drafts` reports `legacy_draft` (true only when the listed row is a NULL-email workspace-level draft and no per-user row shadows it); shown as a "Legacy draft" badge with a hover tooltip.
- Allow discarding a legacy draft: `update_draft` gains a delete-only `legacy` flag that retargets the DELETE (and the conflict re-read) to the `email IS NULL` row, which the email-scoped delete couldn't reach.

**Home page**
- Explain legacy drafts in the draft-badge popover with an info tooltip.
- Stop migrated draft-only items flooding the list: `20260609165313_remove_draft_only` inserted the legacy stubs without an explicit `created_at`, so they all defaulted to the migration's `now()` and bunched at the top. A new repair migration resets those rows' `created_at` to the epoch (real per-item timestamps are unrecoverable), identifying them exactly via `_sqlx_migrations.installed_on` (byte-identical to the inserted rows' `created_at` since sqlx records the bookkeeping row in the same transaction). `remove_draft_only` is left intact so its schema work still runs; this migration runs right after and corrects the timestamps.

**API plumbing**
- Move `UserDraftOverlay`/`UserDraftItemKind` out of the python-client's `-- INLINE START/END --` block in `openapi.yaml` (that block's contents are replaced by the openflow wildcard `$ref` during the python-client build, which was deleting these two definitions and breaking the bundle).

## Test plan

- [x] Review/deploy page: a renamed never-deployed item shows its friendly `draft_path`, not `u/{user}/draft_{uuid}`.
- [x] A legacy draft shows the yellow "Legacy draft" badge; discarding it removes it from the list (exercises `legacy: true` delete) while a normal owned draft still discards as before.
- [x] Deploying a renamed `draft_only` item leaves no leftover storage-path draft.
- [ ] Session pane (flow / raw app): "Restore to deployed" / "Restore to draft" update the preview immediately and refresh the session draft count; opening a path that already has a server draft no longer triggers a stale-write conflict on the standalone editor's next autosave.
- [ ] After the repair migration runs on an instance carrying migrated draft-only stubs, those items no longer flood the top of the home list.
- [x] `cargo check -p windmill-api` (live + `SQLX_OFFLINE=true`); `npm run check`.

## Screenshots

**Compare & Deploy page** — friendly draft paths shown instead of `u/{user}/draft_{uuid}`, the auto-generated (unnamed) draft is no longer bolded, and a legacy NULL-email draft carries a yellow "Legacy draft" badge:

compare
compare_legacy_tooltip
diffdrawer

**Diff drawer** — the obsolete "Latest saved draft <> Current" tab selector is removed (Content/Metadata content selector retained):

(see diffdrawer image above)


diegoimbert:
/ai review this PR and tell me if it's mergeable

PR #9601, opened 2026-06-15
feat(frontend): adapt AI-chat/sessions drafts to DB-backed model

Guilhem-lm:
## Summary

PR #9351 ("Db-backed user drafts") dropped `UserDraft`'s localStorage layer: draft values now live in an in-tab cell created only by a mounted editor handle, with the backend (`DraftService` via `UserDraftDbSyncer`) as the source of truth. The AI-chat / sessions draft adapter was written for the old *synchronous save→read-back* model, so editing an item in the chat threw **`Could not read written draft …`**, and the session preview editors didn't reflect autosave state or guard draft-only diffs.

This branch adapts the AI-chat and sessions drafts to the DB-backed model, plus the related session-editor UX (autosave indicator, draft-only Diff guards, draft diff viewer).

## Changes

**1. AI-chat / sessions drafts → DB-backed adapter** (`copilot/chat/global/*`)
- Backend is the chat's source of truth; the in-tab cell is used opportunistically for live-preview coherence; **conflict-on-save** (file-on-disk semantics) is the safety model — no re-introduced local cache.
- Reads (`getGlobalDraft`, `listGlobalDrafts`, `readGlobalDraftValue`) are async with backend fallback; writes persist through `UserDraftDbSyncer.save` and **synthesize** the returned `WorkspaceItem` instead of reading back. Conflicts surface as a typed tool result (override/reload) on chat-only paths, or the existing per-editor conflict modal when a preview is open.
- The chat and the side-panel preview share one tab-local sync baseline, so saving from one never self-conflicts with the other.
- **Refactor:** the six near-identical `writeXDraft` functions collapse onto one generic `writeDraft<T,A>(spec, …)` skeleton driven by typed per-kind `WriteSpec` constants (`probe`/`fetchDeployed`/`buildDraft`/`beforePersist?`). Behavior-preserving; per-kind merge logic moved verbatim into each spec.
- Terminology: "local draft" → "draft" (drafts are server-side); tool descriptions clarify drafts are per-user, saved server-side, not deployed.
- `WorkspaceItemDrillPicker` and `/global_drafts` switched `listGlobalDrafts` to the `resource()` async pattern.

**2. Session-editor autosave indicator + draft-only Diff guard**
- Script/Flow/RawApp session previews pass an explicit `(workspace, path)` autosave target so the cloud `AutosaveIndicator` watches the key saves land on (previously it watched an empty path and never animated).
- The Diff button is disabled with a hint for draft-only (`no_deployed`) items, consistently across the three editors.
- Script topbar compact breakpoint/layout adjusted so the cloud icon is part of the bar (no float/overflow); splitpanes no longer over-constrains session panes on reload.

**3. Session draft diff viewer — schedule / resource / variable**
- The draft diff path canonicalizes both sides onto one field set and strips runtime-only fields so rows aren't spuriously all-changed; secret values masked. Draft itemKinds map to deploy-style kinds for correct row icon/label. The runtime-field ignore list reuses the exported `TRIGGER_RUNTIME_IGNORE` (single source of truth).
- Diff-viewer leaf rows render at uniform height whether or not the item has a summary (opt-in `uniformHeight`).

## Test plan
- [x] `core.test.ts` — 67/67 pass (assert backend persistence + synthesized return across all draft kinds)
- [x] Browser pass (live EE session, deepseek-chat): chat **create** (new-from-args), **edit** (existing-draft merge, "Updated" verb, read served from in-tab cell), and a **second spec kind** — `drafts/update` POSTs fire, no "Could not read written draft", preview reflects changes
- [x] Cross-tab conflict: a draft advanced by another tab is caught on the chat's next save (`status: conflict`, surfaced as *"changed externally"*); overwrite happens only via an explicit `override`, never a silent stale write
- [x] Session draft diff viewer (EE session): variable secret rendered masked as `<secret>`, resource/schedule canonical fields only (runtime fields stripped, no all-changed noise), draft-only items render as added; diff-viewer rows measure equal height with and without a summary
- [ ] `npm run check` (full) before merge

## Screenshots

**Chat write tools (create + edit a script draft):**

chatwrite-create-script
chatwrite-edit-script

**Script topbar — cloud autosave indicator, compact and expanded:**

topbar-fixed-860-compact
topbar-fixed-1040-noncompact

**Session panes distribute evenly on reload (no splitpanes warning):**

session-reload-5050

**Draft diff viewer — variable secret masked, resource/schedule canonical (no runtime-field noise):**

diff-variable-masked
diff-resource
diff-schedule

**Diff-viewer rows render at equal height with or without a summary:**

diff-row-height


Guilhem-lm:
Thanks @Codex and Pi — addressed all three P1s in 3c5be916e5 (DB-backed draft adapter reporting success when the write didn't land):

- **Codex #1 — save failures reported as success:** `persistGlobalDraft` now checks `UserDraftDbSyncer.getState().state === 'failed'` after the save and returns a new `status: 'error'`; `finishDraftWrite` emits `success:false` with a retry hint.
- **Codex #2 — raw-app conflicts dropped:** `saveGlobalAppDraft` now returns the full `DraftPersistResult`, and the six `write_app_*` tools route through a shared `finishAppDraftWrite` (same conflict/error handling as the other kinds).
- **Pi #1 — catch-all in `fetchBackendDraftValue`:** narrowed to `status === 404`; 403/500/network now propagate instead of collapsing to "no draft" (which would merge from the deployed item and lose draft edits).

Added unit coverage for all three (save-failure → `error`, non-404 read → propagates, stale raw-app write → `conflict`). 71/71 pass, `check:fast` + full build clean.

Guilhem-lm:
Addressed the second-round Codex findings in f956d3003c:

- **[P1] failed/conflicted deletes reported as removed:** `deleteGlobalDraft` now checks `getState().state` + `getConflict()` after the awaited `value:null` save and throws — so `discard_local_draft` / `deploy_workspace_item` / `delete_workspace_item` / the `/global_drafts` delete surface an error instead of "removed" when the server delete didn't land. (Same guard as the write path.)
- **[P2] `override` leaking into schedule drafts:** `SCHEDULE_SPEC.buildDraft` strips the tool-only `override` flag before `mergeDraftConfig`, so it's no longer cloned into the persisted schedule value.

Tests added: failed server delete → throws; schedule draft no longer contains `override`. 73/73 pass, `check:fast` + full build clean.

(Also: thanks Pi for the "Good to merge" on the first round.)

Guilhem-lm:
Addressed the "Clear all" nit (Codex P2) in the latest commit: `/global_drafts` `Clear all` now deletes each listed draft from the backend via `deleteGlobalDraft()` (the same backend-aware path as per-row delete, continuing past per-row failures), then clears local cells — so persisted rows no longer survive a bulk clear. check:fast + full build clean.

Thanks @Pi for the thorough "Good to merge" trace of the error chain.

PR #9629, opened 2026-06-16
fix(cli): fall back to esbuild-wasm on native host/binary mismatch

rubenfiszel:
## Summary

App bundling during `wmill sync push` (e.g. GitHub code sync) can crash on a worker with:

```
✘ [ERROR] Cannot start service: Host version "0.28.0" does not match binary version "0.28.1"
```

esbuild splits into a JS host package and a per-platform native binary (`@esbuild/<platform>`) that must be the same version. A broken or incremental install can leave the on-disk binary at a different version than the pinned host. The running code can't fix what npm/bun put on disk — so when the native service refuses to start, `createBundle` transparently falls back to **esbuild-wasm**, whose binary is a single version-pinned `.wasm`. Host and binary then come from the same package by construction, so nothing on the consumer host can make them drift.

Native esbuild stays the default and the fast path; wasm only triggers when native can't run (or when forced).

### Keeping the 14MB out of the install

`esbuild-wasm` is **not** a dependency. The fallback downloads the package once and caches it on disk, on the fallback path only. So the happy path pays nothing (native speed, no download), the published CLI doesn't grow, and a broken worker pays a one-time download then uses the cache.

Note: it downloads the **whole esbuild-wasm package** (not just the `.wasm`). esbuild-wasm reads the app's files from disk by spawning `node bin/esbuild`, which needs `bin/esbuild` + `esbuild.wasm` + `wasm_exec*.js` co-located — the browser build that can take a bare `.wasm` module has no filesystem access and can't bundle a real app. The version downloaded is read from the native host's `esbuild.version`, so the wasm matches the broken host exactly.

## Changes

- New `cli/src/commands/app/esbuild_loader.ts`:
  - `getEsbuild()` — native-first; a `transform("")` smoke test forces the native service to start so any breakage surfaces at load time. On **any** smoke-test failure it falls back (the host/binary mismatch text is printed to the child's stderr while the thrown error is the generic `The service was stopped` — so we gate on the smoke test, not a string match). Memoized per process.
  - `stopEsbuild()` — stops the esbuild service (native or wasm; both are child-process services) so the process can exit.
  - Fallback acquisition: `WINDMILL_ESBUILD_WASM_PATH` (extracted package dir override) → on-disk cache → download `WINDMILL_ESBUILD_WASM_URL` (default npm registry tarball). Tarball is gunzipped + untarred (via the already-bundled `tar-stream` + `node:zlib`) into a temp dir and renamed into place (atomic; a partial/corrupt cache self-heals). `WINDMILL_FORCE_ESBUILD_WASM=1` skips native entirely.
- `cli/src/commands/app/bundle.ts`: `createBundle()` uses `getEsbuild()`/`stopEsbuild()` and reads results from in-memory `outputFiles` with `write:false` (esbuild-wasm rejects `write:true`; native supports `write:false` too, and the dist files were deleted after the read anyway).
- `cli/src/commands/app/dev.ts`: comment documenting dev mode is intentionally native-only (relies on `context()`/watch; the crash matters on workers via `sync push`, not local dev).
- `cli/build-npm.ts`: comment explaining `esbuild-wasm` is not a dependency and is fetched at runtime.

## Test plan (verified end-to-end against the real built+packed `windmill-cli`)

- [x] `bun run build-npm.ts` + `npm pack` + install into a clean project: generated `package.json` has **no** `esbuild-wasm`, no esbuild `.wasm` shipped (only the ~3.5MB CLI), `npm install` pulls native `@esbuild/<platform>` but not esbuild-wasm.
- [x] Native path on a real `.raw_app`: builds successfully, no warning, no download.
- [x] **Auto-fallback**: with the native `@esbuild` binary swapped to a mismatched version, `wmill app lint` detects native is unusable, warns, downloads+extracts esbuild-wasm, and builds successfully.
- [x] Forced wasm cold (`WINDMILL_FORCE_ESBUILD_WASM=1`): downloads + extracts + builds.
- [x] Forced/auto wasm warm: reuses the cached package, no re-download.
- [x] Corrupt/partial cache self-heals (extract is atomic via temp+rename).
- [x] Healthy native does **not** false-trigger the fallback (no download).
- [x] Runs under plain `node` (not just bun), i.e. the published artifact.
- [x] `cli` lint unit suite (exercises `createBundle`): 14 pass / 0 fail.
- [ ] On a worker with a genuinely mismatched native install, confirm `wmill sync push` auto-falls back and succeeds.
- [ ] Confirm a svelte raw app bundles on the wasm fallback path (esbuild plugin API is identical, but worth exercising once).

## Review fixes (commit 13f8c9a)

- **[P0] tar-slip in tarball extraction** — `extractTarball` now resolves each entry through `resolveTarEntryPath()` and rejects any entry that would escape the cache dir (`WINDMILL_ESBUILD_WASM_URL` allows untrusted sources). Added `test/esbuild_loader_unit.test.ts` covering the guard.
- **[P2] Incomplete sync-push coverage** — routed the other esbuild bundling sites through `getEsbuild()`: codebase bundling on push (`script.ts:331`), codebase preview (`script.ts:1568`), and inline-rawscript bundling (`local_path_scripts.ts:46`). All already used `write:false`+`outputFiles`, so it's a drop-in. Dev mode stays native-only (documented).
- Moved the loader to `src/utils/esbuild_loader.ts` since it's now shared across `app/`, `script/`, and `utils/`.

(The earlier Codex P1 about `browser.js`/`write` was from an older revision — the current code uses the Node build `lib/main.js` with `write:false`, as Pi noted.)

### DIFF
--- a/cli/build-npm.ts
+++ b/cli/build-npm.ts
@@ -28,6 +28,9 @@ rmSync(outDir, { recursive: true, force: true });
 // Build with bun — bundle everything except esbuild (platform-specific binary),
 // svelte (optional, only needed for `wmill app bundle/dev`), and parser packages
 // (loaded at runtime via init() with readFileSync for the .wasm binary).
+// esbuild-wasm is not a dependency at all: the host/binary-mismatch fallback in
+// esbuild_loader.ts downloads and caches the whole esbuild-wasm package at
+// runtime, so it stays out of the bundle and the published dependencies.
 console.log("Bundling with bun build...");
 const buildResult = Bun.spawnSync([
   "bun", "build", "src/main.ts",
--- a/cli/src/commands/app/bundle.ts
+++ b/cli/src/commands/app/bundle.ts
@@ -6,6 +6,7 @@ import * as log from "../../core/log.ts";
 import { colors } from "@cliffy/ansi/colors";
 import * as windmillUtils from "@windmill-labs/shared-utils";
 import { readTextFile, readTextFileSync } from "../../utils/utils.ts";
+import { getEsbuild, stopEsbuild } from "./esbuild_loader.ts";
 export interface BundleOptions {
   entryPoint?: string;
   outDir?: string;
@@ -170,8 +171,9 @@ export async function ensureNodeModules(appDir?: string): Promise<void> {
 export async function createBundle(
   options: BundleOptions = {}
 ): Promise<BundleResult> {
-  // Dynamically import esbuild
-  const esbuild = await import("esbuild");
+  // Native esbuild with a transparent esbuild-wasm fallback on host/binary
+  // version mismatch (see esbuild_loader.ts).
+  const esbuild = await getEsbuild();
 
   // Detect frameworks to determine default entry point.
   // Use the entryPoint's directory if provided, otherwise fall back to cwd.
@@ -286,6 +288,10 @@ export async function createBundle(
     outfile,
     sourcemap,
     minify,
+    // Keep outputs in memory: esbuild-wasm cannot write to the filesystem
+    // ("write" option unavailable), and the dist files were discarded after the
+    // read anyway. Native esbuild supports write:false + outputFiles too.
+    write: false as const,
     define: {
       "process.env.NODE_ENV": production ? '"production"' : '"development"',
     },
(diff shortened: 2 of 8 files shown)


rubenfiszel:
Addressed the review findings in 13f8c9a:

- **[P0] tar-slip (Codex)** — fixed. Tarball entries are now resolved via `resolveTarEntryPath()` and rejected if they escape the cache dir; added a unit test for the guard.
- **[P2] incomplete `sync push` coverage (Claude + Pi)** — fixed. Codebase bundling (`script.ts:331`), codebase preview (`script.ts:1568`), and inline-rawscript bundling (`local_path_scripts.ts:46`) now go through `getEsbuild()` too. They already used `write:false`+`outputFiles`, so it was a drop-in. Loader moved to `src/utils/` accordingly.
- Dev mode (`app dev`) intentionally stays native-only (relies on `context()`/watch; documented in the code).
- The earlier Codex P1 (`browser.js`/`write` unavailable) was against an older revision — current code uses the Node build `lib/main.js` with `write:false`, as the Pi review confirmed.

Re-verified end-to-end after the changes: native, forced-wasm cold/warm, and real auto-fallback (mismatched `@esbuild` binary) all bundle a real `.raw_app` successfully; healthy native doesn't false-trigger; unit + lint suites 17/0.

Not done (deliberately): a concurrency guard on `getEsbuild()` memoization — `sync push` bundles sequentially today, so it's harmless; noted as a follow-up if bundling ever goes concurrent.

rubenfiszel:
Addressed the P1 concurrency finding in 8977f54:

- **In-flight memoization**: `getEsbuild()` now memoizes the *promise*, not just the resolved value, so concurrent first callers (a parallel `wmill sync push --parallel`) await a single probe/download instead of each entering the fallback path.
- **Unique extraction temp dirs**: each extraction uses `${destDir}.${pid}.${counter}.tmp`, so two extractions can never share an in-progress directory.

Verified: 5 simultaneous cold `getEsbuild()` calls now produce exactly **one** download, return the same instance, and all build successfully (previously they'd each download and race on the shared temp dir). Native + auto-fallback e2e and the unit suite still green.

PR #9645, opened 2026-06-17
feat(ai-chat): summary-based conversation compaction

centdix:
## Summary

The AI chat previously managed an overflowing context window with **drop-oldest** compaction — once the projected request crossed ~80% of the model's context window, the front of the stored history was deleted outright. That permanently discarded early intent, decisions, file edits, and user feedback from the model's view.

This replaces it with **summary-based partial compaction**: the older *prefix* is summarized by a one-shot LLM call into a single message, while the recent *tail* is kept verbatim. The summarized prefix is replaced — in both the model context and the visible transcript — by a single collapsible "Summarized earlier conversation" boundary. Drop-oldest remains as a fallback, and a circuit breaker disables the summary round-trip after repeated failures so a stuck conversation never hammers the API.

Builds directly on the existing token-tracking (`contextTokens` / `contextUsage`) — only the *action* taken at the trigger changes, not the trigger signal.

## Changes

- **`compactionPrompt.ts`** (new) — structured 9-section summary prompt framed in **Windmill's global-mode domain** (workspace items by path + kind, drafts/deploy/test-runs — not files/file-edits), text-only with `<analysis>`/`<summary>` framing; plus `formatCompactSummary` (strips every analysis scratchpad block, unwraps `<summary>`, falls back to raw text) and `buildSummaryMessageContent`.
- **`AIChatManager.svelte.ts`** — new `summarizeAndCompact()`: picks a tool-paired split (tail grows to `tailBudget = target·window − summary reserve − overhead`, snapped forward to a user-message boundary), summarizes the prefix via `getNonStreamingCompletion`, rebuilds `messages` as `[summary, ...tail]`, replaces the display prefix with a `summary` boundary, and re-bases surviving tail indices. Trigger now calls it first and falls back to `compactOldestMessages`; adds a circuit breaker; moves the `displayLenAfterUser` rollback anchor to after compaction (the summary path shrinks `displayMessages`).
- **`shared.ts`** — `SummaryDisplayMessage` (`{ role: 'summary', content }`, no index — never a restart target) added to the `DisplayMessage` union.
- **`CompactionBoundary.svelte`** (new) + **`AIChatMessage.svelte`** — collapsible boundary divider, wired into the message renderer. While the summary round-trip is in flight, the processing indicator shows a **"Compacting conversation"** label (`AIChatManager.compacting` → `ChatTypingIndicator`).
- **`ContextUsageIndicator.svelte`** — always show the context-window usage readout once a conversation has started (previously gated to ≥50% of the window), so the user can watch context grow toward the compaction trigger.
- **`HistoryManager.svelte.ts`** — the chat title is preserved across compaction: once the summary boundary leads the transcript, the title computed before compaction is reused rather than re-derived (which would otherwise shift it to the first surviving tail message).
- **`ContextManager.svelte.ts`** — `updateDisplayMessages` only touches `contextElements` on user/assistant messages; tool and summary messages pass through untouched.
- **Tests** — `compactionPrompt.test.ts` (formatting/prompt, incl. stripping multiple `<analysis>` blocks) and `AIChatManager` cases: summarize-and-keep-tail, summarize-fails → drop-oldest fallback, prefix-too-small → drop-oldest, and **Stop-during-summarization → no destructive drop-oldest** (token sizes validated against the real `modelConfig` windows so each reaches its intended branch); plus a `HistoryManager` case asserting the title survives a leading summary boundary.

## Screenshots

https://github.com/user-attachments/assets/2e18a363-057a-49aa-a5d5-238bc16030cb

## Test plan

- [ ] `npm run check` passes (0 errors).
- [ ] `vitest run compactionPrompt AIChatManager HistoryManager` passes.
- [ ] Drive an AI chat (any mode) past ~80% of the model's context window; confirm the older turns collapse into one "Summarized earlier conversation" boundary that expands/collapses, the recent turns stay verbatim, and the model continues coherently.
- [ ] Edit/retry one of the surviving recent user messages; confirm it rewinds correctly while the summary boundary stays in place.
- [ ] Exercise on **both** an Anthropic and an OpenAI model — verify summarization actually engages on OpenAI tool-heavy conversations (the Responses-API path sends historical `function_call` items without re-declaring `tools`; if rejected it would silently fall back to drop-oldest).
- [ ] Confirm the context-window usage readout is visible from the start of a conversation, and the past-chats title is unaffected after compaction.
- [ ] Click **Stop** while the indicator shows **"Compacting conversation"**; confirm the turn cancels cleanly (prompt restored/cleared per normal cancel) and the persisted chat keeps its full pre-cancel history rather than a drop-oldest-compacted one.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -464,11 +464,13 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 							{:else}
 								<ChatTypingIndicator
 									loading={aiChatManager.loading}
-									label={aiChatManager.currentReasoningActive &&
-									!aiChatManager.currentReply &&
-									!aiChatManager.currentReasoning
-										? 'Thinking'
-										: undefined}
+									label={aiChatManager.compacting
+										? 'Compacting conversation'
+										: aiChatManager.currentReasoningActive &&
+											  !aiChatManager.currentReply &&
+											  !aiChatManager.currentReasoning
+											? 'Thinking'
+											: undefined}
 								/>
 							{/if}
 						</div>
@@ -538,8 +540,6 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 			{#if inputPreface}
 				{@render inputPreface()}
 			{/if}
-			<ContextUsageIndicator />
-
 			<AIChatInput
 				bind:this={aiChatInput}
 				bind:selectedContext
@@ -727,6 +727,7 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 					</div>
 				{/if}
 			</div>
+			<ContextUsageIndicator />
 		</div>
 		{#if (aiChatManager.mode === AIMode.NAVIGATOR || aiChatManager.mode === AIMode.ASK) && suggestions.length > 0 && messages.filter((m) => m.role === 'user').length === 0 && !disabled}
 			<div class="px-2 mt-4">
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -727,7 +727,9 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 					</div>
 				{/if}
 			</div>
-			<ContextUsageIndicator />
+			<div class="flex px-1 mt-1">
+				<ContextUsageIndicator />
+			</div>
 		</div>
 		{#if (aiChatManager.mode === AIMode.NAVIGATOR || aiChatManager.mode === AIMode.ASK) && suggestions.length > 0 && messages.filter((m) => m.role === 'user').length === 0 && !disabled}
 			<div class="px-2 mt-4">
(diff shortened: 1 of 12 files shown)


centdix:
Addressed the review feedback from both bots in d99f839b01.

**P1 — Stop during summarization no longer destructively compacts (`AIChatManager.svelte.ts`)**
Previously, hitting Stop while the summary request was in flight made `summarizeAndCompact()` catch the abort and return `false`, which the trigger treated as an ordinary failure → it ran `compactOldestMessages()` and persisted a drop-oldest-compacted history, while the request then no-op'd on the aborted controller and rolled back only the unsent user turn. Net effect: cancelling mid-compaction permanently dropped older history. Fixed by short-circuiting the fallback (and its save) when `this.abortController?.signal.aborted` — the existing cancel path now rolls the pushed turn back cleanly. Added a regression test (`Stop during summarization → no drop-oldest`).

**P2 — chat title preserved across compaction (`HistoryManager.svelte.ts`)**
`saveChat` re-derived the title every save; after compaction the first message is a summary boundary, so it shifted to the first surviving tail message. Now once a summary boundary leads the transcript, the title computed before compaction is reused instead of re-derived. Added a `HistoryManager` test. (PR body line corrected — it previously overstated this.)

**P2 — `formatCompactSummary` strips every `<analysis>` block (`compactionPrompt.ts`)**
Added the `g` flag so multiple scratchpad blocks can't leak into context. Added a test case.

**Nit — indentation (`AIChatMessage.svelte`, `ContextUsageIndicator.svelte`)**
Ran prettier on both; whitespace-only.

Also fixed a stale PR-body line that still claimed the summary prompt covers "UI navigation" — that was removed when the prompt was tailored to global mode.

`npm run check` / `vitest` to be run locally per the test plan (couldn't run them in this environment).

PR #9648, opened 2026-06-17
feat(ai-chat): workspace AI chat skills (SKILL.md upload + read_skill tool)

centdix:
## Summary

Adds workspace-scoped **AI chat skills** — Claude/Codex-style `SKILL.md` instruction sets that the AI chat can use in **global mode**. Skills are managed from a **Custom skills** section inside the workspace's Windmill AI settings (paste a `SKILL.md` or import a folder); the chat advertises each skill's name + description in its system prompt and loads the full instructions on demand via a new `read_skill` tool.

Scope for this PR (per product decision): **workspace-scoped only**, **SKILL.md-only** (no reference files), **Community edition** (no EE gating).

## Changes

**Backend**
- New `ai_skill(workspace_id, name, description, instructions, edited_at, edited_by)` table (migration `20260617144417_add_ai_skill`).
- New `windmill-api/src/ai_skills.rs` workspaced router at `/ai_skills`:
  - `GET /list`, `GET /get/{name}` — readable by any workspace member.
  - `POST /upload` (bulk upsert), `DELETE /delete/{name}` — `require_admin` + audit-logged.
  - `validate_skill` constrains `name` to a 1–255 char slug (`[A-Za-z0-9_-]`) and rejects empty description/body.
- `openapi.yaml` + generated client (`WorkspaceService.{list,get,upload,delete}AiSkill(s)`); SQLx offline cache updated.

**Frontend — settings**
- New **Custom skills** section inside the **Windmill AI** workspace settings tab (`AiSkillsSettings.svelte`, embedded in `AISettings.svelte`), with two labeled subsections:
  - **Paste a `SKILL.md`** (frontmatter `name` + `description`, markdown body) → "Add skill".
  - **Import a folder of skills** (`webkitdirectory`); each `SKILL.md` becomes a skill named after its folder.
- Folder-import guardrails: only `SKILL.md` within **≤ 3 path levels** is considered (deeper/vendored files are skipped), at most **50 skills** per import, and a **confirmation dialog** (count + names + skipped count) before writing.
- Lists existing skills with delete (confirmation modal). Invalid/skipped skills are reported in a toast.

**Frontend — AI chat (global mode)**
- `loadWorkspaceSkills()` fetches the list when global mode is entered; the global system prompt gains a **Skills** section (`name — description`) when skills exist.
- New `read_skill` tool returns a skill's full instructions on demand.
- `AIChatManager` rebuilds the system message once skills resolve (the chat loop re-reads it via getter).

## Test plan

- [x] Backend endpoints exercised via curl: list/get/upload/delete + validation (bad name / empty description → 400, missing → 404).
- [x] Settings UI: pasted a `SKILL.md` → parsed & listed; imported a folder → depth filter dropped a too-deep file, confirmation dialog showed the count, confirming added the skill; a 51-skill import was refused by the cap (no write); deleted via the confirmation modal.
- [x] Chat: switching to Global mode fires `GET /ai_skills/list`; asking *"get the weather forecast for Paris"* made the model call `read_skill("weather-helper")` and follow the skill's made-up instructions (use `acme-weather` resource, `GET /v3/forecast`) — proving prompt injection + tool round-trip.

## Notes
- `ai_skill` is read via `user_db` without an RLS policy — intentional: it's workspace-global config (like `workspace_settings`/`workspace_shared_ui`, which also have no RLS), and Windmill's RLS policies are per-owner/per-folder, which don't apply here. Cross-workspace isolation is enforced by the explicit `workspace_id = $1` filter + the `ApiAuthed` membership check.
- No automated tests yet for the pure helpers (`validate_skill` / `parseSkillMd` / `collectSkills`) — candidate follow-up.

## Follow-up

- Explore instance-level AI skills that apply across all workspaces and are managed from instance AI settings. Preferred shape: a dedicated `instance_ai_skill` table, merged into the workspace effective skill list, with workspace skills overriding inherited instance skills by name.
- Explore user-level AI skills for personal reusable instructions. The effective set should have explicit precedence across instance, workspace, and user scopes, and `read_skill` should resolve against that same precedence.

## Screenshots

**Custom skills section inside the Windmill AI settings** (paste a SKILL.md or import a folder, list + delete):

custom-skills

**Folder-import confirmation** (skipped a too-deeply-nested SKILL.md):

import-confirm

**Global AI chat calling `read_skill` and following the skill's instructions**:

ai-chat-read-skill


centdix:
## Addressed review feedback — spec alignment + per-workspace cap

Pushed `35c5309`. The changes bring skill validation in line with the official [Claude SKILL.md spec](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#skill-structure) (which is the format this feature advertises) and add a per-workspace cap.

**`validate_skill` (backend) — now matches the SKILL.md spec**
- `name`: max **64 chars** (was 255), charset tightened to **`[a-z0-9-]`** lowercase/digit/hyphen (was `[A-Za-z0-9_-]`). This is the Claude Code skill-name convention, so workspace skills stay portable with real Claude/Codex skills.
- `description`: now counted in **characters** with max **1024** (was 1000 *bytes*) — also resolves the byte-vs-char concern raised in review.
- `instructions`: kept as a 64 KiB **byte** guard (not a spec field — it's a payload bound).
- `openapi.yaml` synced (`maxLength: 64` + `pattern: ^[a-z0-9-]+$`, description `maxLength: 1024`); frontend pre-validates the same rules so paste/folder-import surface friendly inline errors instead of a raw server toast (closes the slug pre-validation nit).

**Per-workspace cap**
- `MAX_SKILLS_PER_WORKSPACE = 100`, enforced on upload accounting for upsert (names already present don't count as new). Bounds the global system-prompt growth flagged by several reviewers. Frontend mirrors the check.

**Tests:** added unit coverage for the new name rules (length, non-slug charset, char-vs-byte counting) and the capacity helper; all backend tests pass and the new SQLx query is cached for CI.

**Deliberately not done:**
- The spec's **reserved-word (`anthropic`/`claude`)** and **XML-tag** rules — these protect Anthropic's namespace/prompt structure. Upload here is `require_admin`-gated (low injection value) and a naive XML check false-rejects prose like `a<b and c>d`, so the cost/benefit didn't justify them.
- The reported **"missing `check_scopes` in handlers"** — this is a false positive. Scope enforcement for `ai_skills` happens in the auth middleware (`check_scopes_for_route` → `check_route_access`), covered by `test_ai_skills_scope_access` (which proves an `ai_skills:read` token is rejected on `/upload` and `/delete`). Handler-level `check_scopes` is only needed for path-scoped surfaces (e.g. `apps:read:<path>`); `ai_skills` scopes are domain-level only.

**Follow-up for reviewers:** the generated TS client should be regenerated (`npm run generate-backend-client`) since `openapi.yaml` changed.

centdix:
## Addressed remaining nits — duplicate uploads + audit resource

Pushed `c4793bb`.

- **cubic P2 — duplicate skill names in one upload** (`ai_skills.rs`): `upload_skills` now rejects a payload that repeats a name (compared after trimming) via a small `collect_upload_names` helper, instead of silently upserting last-wins and reporting a wrong count. Unit-tested.
- **Claude nit — audit `resource` consistency** (`ai_skills.rs`): `upload` now logs the uploaded skill names as the audit `resource` (matching `delete_skill`, which logs the skill name), rather than the uploader's email. The `skill_count` is now derived from the deduplicated names, so it always matches the rows actually written.

Audit logging itself is kept — it's the forensic trail for an admin-gated mutation that injects instructions into the AI chat system prompt.

All backend `ai_skills` tests pass (9 total).

centdix:
## Synced dereferenced OpenAPI artifacts

Pushed `11e18c4`. Good catch — the generated `openapi-deref.yaml`/`openapi-deref.json` still advertised the old AI-skill validation (`name` `maxLength: 255` / `^[A-Za-z0-9_-]+$`, `description` `maxLength: 1000`), so any docs/client path consuming the deref specs would have told callers to send names the endpoint now rejects.

Both artifacts now match the source `openapi.yaml` and the Rust handler: `name` `maxLength: 64` + `^[a-z0-9-]+$`, `description` `maxLength: 1024`. Edited the two values surgically rather than re-running `build_openapi.sh`, to avoid a full redocly reformat of the 1.3 MB files — verified no other stale occurrences remain.

PR #9662, opened 2026-06-19
fix: trigger flow error handler on unrecoverable (OOM/zombie) step failures

rubenfiszel:
## Summary

Fixes **WIN-2070**: within a flow, the flow's error handler (failure module) was not triggering when a step failed due to OOM.

The issue mentions OOM has two forms. I traced both:

### Case 1 — OOM detected by Windmill (child OOM-killed, e.g. exit 137; worker survives)
The step fails through the normal completion path with `unrecoverable = false`, so `update_flow_status_after_job_completion` already routes to the failure module. ✅ **Already worked** (covered by the existing `test_failure_module`).

### Case 2 — worker crash, detected by the zombie job handler
The crashed worker leaves the running step as a zombie; `monitor::handle_zombie_jobs` fails it via `handle_job_error`. For a **same-worker** step (and for the internal flow-status-update error fallback), `unrecoverable = true` is passed, and `update_flow_status_after_job_completion_internal` had:

```rust
false if unrecoverable => false,
```

This silently completed the flow with the raw error and **skipped the failure module entirely**. Even if it had continued, the failure module would have been pushed back to the (dead) worker via `same_worker`, stranding it. ❌ **This is the bug.**

> Note: non-same-worker zombie steps already pass `unrecoverable = false`, so those already triggered the failure module.

## Changes

- `worker_flow.rs` — unrecoverable step failures now **continue to the failure module** (error handler) instead of being silently dropped. The new branch mirrors the existing error-handler condition (`!is_failure_step && !has_triggered_error_handler && failure_module.is_some()`).
- `worker_flow.rs` — thread an `unrecoverable` flag through `handle_flow` → `push_next_flow_job` and, when set, push the error-handler step as a **regular queued job** (force `continue_on_same_worker = false` and `continue_with_runners = false`) so a live worker can pick it up rather than the dead one.
- `worker_flow.rs` — gate the per-step **retry** evaluation in `push_next_flow_job` on `!unrecoverable`: a worker-crash/OOM failure is not retried (the worker and its state are gone), it goes straight to the error handler.
- `worker.rs` — freshly pulled flow jobs pass `unrecoverable = false`.
- `tests/worker.rs` — two regression tests driving a real worker and simulating the zombie handler (`handle_job_error(..., unrecoverable = true, ...)`): one asserts the failure module runs, one asserts a retry-configured step is **not** retried and still routes to the error handler.

## Known gaps left as follow-ups (not changed here)

- Flows that hang **between steps** because the worker died mid-transition are handled by `monitor::handle_zombie_flows`, which **force-cancels** them. That path still bypasses the in-flow failure module, and the workspace error handler is muted when `muted_on_cancel` is set. Making an OOM-induced zombie cancellation trigger the error handler is a separate, larger change.

## Test plan

- [x] `cargo check -p windmill-worker --features enterprise,deno_core`
- [x] `cargo test -p windmill --test worker --features enterprise,deno_core,quickjs test_failure_module` (no regression)
- [x] `cargo test ... test_failure_module_triggered_on_unrecoverable_failure` (passes with fix; **fails** when the decision-logic fix is reverted — verified)
- [x] `cargo test ... test_unrecoverable_failure_skips_retry_runs_failure_module` (passes with fix; **times out/fails** when the retry gate is removed — verified)
- [ ] Manual: a flow with a failure module whose step is OOM-reaped on a same-worker step → the failure module executes on a live worker and the flow completes through the error handler

---
## Summary by cubic
Ensure the flow error handler runs on unrecoverable step failures (OOM/worker crash), including when the failed step had continue_on_error, and skip retries, fixing WIN-2070. The failure module is queued to any live worker instead of being pinned to the dead one.

- **Bug Fixes**
  - Route unrecoverable step failures to the flow’s failure module instead of completing the flow with a raw error; this also applies to continue_on_error steps.
  - Thread an `unrecoverable` flag through flow handling and enqueue the failure module as a regular job (disable same-worker and runners) so a live worker can pick it up.
  - Bypass per-step retry when `unrecoverable=true`.
  - Add regression tests for zombie-step failures, retry bypass, and continue_on_error behavior; include `sqlx` offline cache for the new test query.

<sup>Written for commit df61b0fc5f71a02b7912a13ffb7ba38f11817e98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-fac563138c316998d4f523edd633db97dba77d649b637c2529e4f232dce16c88.json
+++ b/backend/.sqlx/query-fac563138c316998d4f523edd633db97dba77d649b637c2529e4f232dce16c88.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT q.id FROM v2_job_queue q JOIN v2_job j USING (id)\n                     WHERE j.parent_job = $1 AND q.running = true",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Uuid"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "fac563138c316998d4f523edd633db97dba77d649b637c2529e4f232dce16c88"
+}
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -3990,6 +3990,159 @@ async fn test_failure_module(db: Pool<Postgres>) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Regression test for WIN-2070: a flow step that fails *unrecoverably* — e.g. its worker was
+/// OOM-killed and the failure is surfaced by the zombie job handler — must still trigger the
+/// flow's error handler (failure module). Previously, `unrecoverable` failures silently
+/// completed the flow with an error and skipped the failure module entirely.
+///
+/// This drives a real worker on a flow whose first step hangs forever, then simulates the
+/// zombie handler by calling `handle_job_error(..., unrecoverable = true, ...)` against that
+/// running step exactly as `monitor::handle_zombie_jobs` does.
+#[cfg(feature = "deno_core")]
+#[sqlx::test(fixtures("base"))]
+async fn test_failure_module_triggered_on_unrecoverable_failure(
+    db: Pool<Postgres>,
+) -> anyhow::Result<()> {
+    use std::sync::atomic::AtomicU16;
+    use std::sync::Arc;
+    use tokio::sync::mpsc;
+    use windmill_common::auth::create_token_for_owner;
+    use windmill_common::client::AuthedClient;
+    use windmill_common::KillpillSender;
+    use windmill_queue::{get_queued_job_v2, MiniCompletedJob, SameWorkerPayload};
+    use windmill_worker::{JobCompletedSender, SameWorkerSender};
+
+    initialize_tracing().await;
+    let server = ApiServer::start(db.clone()).await?;
+    let port = server.addr.port();
+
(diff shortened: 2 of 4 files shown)


rubenfiszel:
Addressed the P1 from the Codex/Pi reviews (`continue_on_error` + unrecoverable could skip the error handler) in df61b0fc5f.

Root cause confirmed: for a `continue_on_error` step, the step counter was advanced (`inc = true`) *before* the new `unrecoverable` decision branch, so `push_next_flow_job` saw `status.step` already pointing at the next module and pushed it instead of the failure module.

Fix (two parts, both gated on `unrecoverable` so normal `continue_on_error` is unchanged):
1. `worker_flow.rs` — do not set `inc` for an unrecoverable `continue_on_error` failure, so the step counter stays on the failed module (status `Failure`).
2. `worker_flow.rs` — the `FlowStatusModule::Failure` arm in `push_next_flow_job` now also routes to `Step::FailureStep` when `unrecoverable` (previously the `!continue_on_error` guard blocked it).

Regression test added: `test_unrecoverable_failure_on_continue_on_error_runs_failure_module` — a flow `a(continue_on_error=true) -> b` with a failure module; simulates the zombie handler reaping `a` unrecoverably and asserts the failure module runs **and** `b` does not. Verified it fails without the fix (flow returns `{"ran_b":true}`) and passes with it.

All five `failure_module` worker tests plus `test_deno_flow`/`test_deno_flow_same_worker` (which use `continue_on_error`) pass.

PR #9663, opened 2026-06-19
fix: unify draft path on content `path`, drop redundant `draft_path`

diegoimbert:
## Summary

In the flow, app and raw-app editors, a renamed draft stored its target path in a separate `draft_path` field while the content's own `path` kept the deployed value. This was redundant — the original/deployed path is already the URL — and it broke the "path already used" guard: editors passed the renamed `draft_path` as the Path widget's `initialPath`, so reverting a draft back to its **original** path falsely tripped "path already used" (and the renamed `draft_path` could also become the `updateFlow`/`updateApp` target).

The script editor already did the right thing — it binds the Path widget straight to `script.path`. This PR makes every editor consistent with that model:

- The draft stores the user-typed path in the content's own `path` (no separate `draft_path` field).
- The Path widget's `initialPath` is always the deployed/original path (the URL). A never-deployed draft (parked at a synthetic `…/draft_{uuid}` key) seeds the baseline from its own friendly path, so the storage key isn't flagged as a rename.

The computed `draft_path` field on the list/draft API responses is **kept** (it's how the home / Review&Deploy / Drafts lists render the friendly name) but is now sourced uniformly from `value->>'path'` for all kinds — the same way scripts already worked.

## Changes
- **Backend**: `drafts.rs` list query, plus the draft-only synth loops in `flows.rs` and `apps.rs`, now read `value->>'path'` (instead of `value->>'draft_path'`) for the computed `draft_path` display field. Doc/comment updates in `openapi.yaml`, `openapi-deref.*`, and the `Listable*`/`DraftListItem` field comments.
- **Flow editor**: `FlowBuilder` mirrors the typed path into `flow.path` (was `flow.draft_path`); `flows/edit` + `FlowEditorView` pass the deployed URL as `initialPath` (never the renamed path), seeding from the friendly path only for never-deployed drafts; `savedFlow.path` is pinned to the deployed path; `draft_path` dropped from `flowDraftSig`.
- **App / raw-app editors**: typed path mirrored into the value's own `path`; `AppEditorHeaderDeploy` uses the deployed `savedApp.path` as `initialPath` for deployed items; raw-app route + session codec persist the typed path as the draft's `path` (in-memory `typedPath`); `rawAppDeploy`/`utils_draft_deploy` deploy at `value.path`.
- **Scripts (session)**: removed the now-redundant `draft_path` mirroring from `sessionDraftCodecs` (scripts already carry `script.path`); `sig` keys on `path`.
- Updated the sibling unit tests (`appDraftCodec`, `flowDraftSig`, `sessionDraftCodecs`).

## Screenshots

Flow editor — a deployed flow with a rename draft, path reverted to the **original** `best_known_flow`: no "path already used", status "Saved" (the draft is discarded since it now matches deployed):

flow revert — no false positive

Collision detection still works — typing an existing *different* path (`other_flow`) correctly shows "path already used":

collision still detected

## Test plan
- [x] `npm run check:fast` (frontend types) — clean
- [x] Frontend unit tests (`appDraftCodec`, `flowDraftSig`, `sessionDraftCodecs`) — pass
- [x] Backend compiles (cargo watch)
- [x] Backend: `/drafts/list` returns `draft_path` computed from `value.path` for both a flow and an app rename draft
- [x] Flow editor: a deployed flow with a rename draft shows the friendly name; reverting to the original path shows **no** "path already used"
- [x] Visual app editor: same — reverting to the original path shows no error
- [x] Negative case: typing a different existing path still shows "path already used"
- [ ] Raw app editor: rename → home/Drafts list shows the friendly name → deploy lands at the typed path (shares the verified `AppEditorHeaderDeploy` path + codec unit tests; not exercised end-to-end in-browser)

Note: the gitignored `frontend/src/lib/gen/*` client was not committed (generated, not tracked); the committed `openapi-deref.*` were updated by hand to match the `openapi.yaml` description edits rather than full-regenerated, to avoid sweeping in unrelated pre-existing artifact drift.


diegoimbert:
## Follow-up: show staged renames on deployed items' home rows

Two fixes:

1. **Deployed items with a rename draft now show the friendly name.** The list endpoints only set the computed `draft_path` field for synthesized *draft-only* rows, so a deployed item with a rename draft kept showing its old deployed path. The main `flows`/`scripts`/`apps` list queries now populate `draft_path` from the joined draft's own `value->>'path'` (the apps draft subquery now exposes `value` via `DISTINCT ON` to avoid fan-out).

2. **The row subtitle renders the rename as `~~old~~ new`** via `Row`'s existing `pathDisplay` snippet, instead of just the friendly path. Draft-only items (no deployed counterpart) still show the plain friendly name.

Verified live (fresh backend on the dev DB):
- `flows/list` / `apps/list` return `draft_path` for a deployed item with a rename draft (and `None` for items without one).
- `npm run check:fast` clean.

Home page — deployed flow + deployed app with rename drafts show `~~old~~ new`; the draft-only `amazing_flowxs` shows just the friendly name; `other_flow` (no draft) is unchanged:

home struck path

diegoimbert:
## Follow-ups: clean deployed value (#1) + migrate pre-existing drafts (#3)

**1. Strip draft-only `path`/`summary` from the deployed app value.** The visual app editor mirrors the typed `path` (on rename) and `summary` onto the autosaved App value so the lists can read them; those are draft-only metadata (own `app`-table columns) and shouldn't ride into `app_version.value`. The Review & Deploy path already stripped them; the direct editor deploy now does too, via an `appValueForDeploy` helper applied at every site `$app` is used as the deploy/compare/diff value (create/update sends, the saved baseline, the no-op deploy check, the fetched `deployedValue`, and both diff sides) so the stored JSON stays clean and the no-op check stays symmetric (legacy values with a baked-in summary are stripped on both sides too).

Verified: deploying a renamed app through the editor stores `{"grid":[],"theme":{…},"fullscreen":false,…}` — no `path`/`summary` keys — at the renamed path.

**2. Run-once migration for in-flight drafts written before this PR.** As raised in review: old drafts store the staged rename in `value->>'draft_path'` while `value->>'path'` holds the deployed/storage path, so after this change they'd lose the friendly name from the lists and deploy at the old path (dropping the rename). Added a frontend-only, idempotent, per-workspace-localStorage-guarded migration (`draftPathMigration.ts`) that rewrites each of the current user's drafts (`path = draft_path`, then drop `draft_path`), preserving the original `created_at` so they don't resurface. It's fired fire-and-forget from the workspace-drafts loader and refreshes the list when it changes anything; drawer kinds are skipped (they never carried `draft_path`, and a secret variable's encrypted value is never re-saved).

Verified end-to-end: an old-format flow draft (`value.path=u/admin/legacy_flow`, `value.draft_path=u/admin/legacy_flow_renamed`) → after loading the home page → `value.path=u/admin/legacy_flow_renamed`, `draft_path` gone, `created_at` unchanged, and the home list now shows the struck rename. 6 unit tests cover rewrite / redundant-drop / no-op / drawer-skip / run-once / retry-on-failure.

This makes the upgrade migrate in-flight renames rather than silently dropping them, so the prior "won't fix" caveat no longer applies.

PR #9666, opened 2026-06-19
feat(backend): auto-reconnect postgres trigger listener with backoff (WIN-2073)

hugocasa:
## Problem

The Postgres trigger listener permanently disabled itself on connection errors. On a closed replication stream it returned silently; on a receive error it called `disable_with_error()`. A transient network interruption (e.g. a cloud provider maintenance window) therefore permanently killed the trigger, requiring a manual re-enable.

In contrast, the Kafka trigger listener (`windmill-trigger-kafka/src/listener_ee.rs`) already has a retry loop with 30s backoff that handles transient failures gracefully.

## Fix

Restructure `PostgresTrigger`'s `Listener` impl to mirror the Kafka pattern:

- **`Consumer` type** changed from a live `(CopyBothDuplex<Bytes>, LogicalReplicationSettings)` to `ApiAuthed`. `get_consumer` now only resolves the auth context; the live replication connection is established inside `consume`, so the very first connection attempt also benefits from retry.
- **New helper `connect_logical_replication_stream`** holds the logic that used to live in `get_consumer`: resolve the Postgres resource, validate the publication and replication slot exist, and open the replication stream. The resource is re-resolved on every attempt, so credential rotations are picked up across reconnections.
- **`consume` now wraps an outer reconnect loop**:
  - On connection failure: log, update ping status, report a critical error every 10 attempts (trigger mode only), sleep 30s, increment a `tries` counter, and retry.
  - On successful reconnect after prior failures: call `report_recovered_critical_error()` and reset `tries`.
  - In the inner message loop: on stream close or receive error, update ping status, back off 30s, and `break` to the outer loop to reconnect instead of disabling.
- **`disable_with_error()` is kept only for unrecoverable cases**: missing publication / replication slot (`Error::BadConfig`) and unparsable replication messages.

### Reconnect flow

```mermaid
flowchart TD
  A[consume] --> B[connect_logical_replication_stream]
  B -->|Ok stream| C[clear status / report recovered if tries>0]
  B -->|Err BadConfig: pub/slot missing| D[disable_with_error + stop]
  B -->|Err transient| E[update ping, alert every 10, sleep 30s, tries+1] --> B
  C --> F[inner message loop]
  F -->|stream closed / recv error| G[update ping, sleep 30s] --> B
  F -->|parse error| D
```

## Notes / decisions

- A deleted Postgres resource now causes retries (with an alert every 10 attempts) rather than an immediate disable, since it is not an `Error::BadConfig`. This is intentionally more graceful: the trigger recovers automatically if the resource is recreated.
- Inner-loop disconnects (stream close / receive error) sleep 30s before reconnecting, matching Kafka's battle-tested behavior. Retained WAL means no events are lost during the gap, only delayed.

## Validation

- `cargo check -p windmill-trigger-postgres` — clean (no errors/warnings).
- Full backend binary rebuilt and restarted cleanly under `cargo watch` (built with `--features quickjs`, which exercises the downstream `windmill-api` / binary consumers of the trait).
- No new `sqlx` macros introduced (only existing helpers + `simple_query`), so no offline-cache regeneration needed.
- A full runtime test of the reconnect path was not performed: the local dev Postgres runs with `wal_level=replica`, and enabling logical replication would require restarting the Postgres instance the running backend depends on. The logic is a direct port of the proven Kafka reconnect pattern.

Fixes WIN-2073

---
## Summary by cubic
Adds auto-reconnect with a 30s backoff to the Postgres trigger listener and improves alerting. Counts reconnects on stream drops and reports recovery only after messages flow, aligning with the Kafka/SQS pattern and fulfilling WIN-2073.

- **New Features**
  - Reconnect loop in `consume` with 30s backoff on connect failure, stream close, or receive error.
  - Reconnect attempts are counted; critical alert on the first failure and every 10 after. Ping status is updated during failures, and recovery is reported only once the reconnected stream delivers a message.

- **Refactors**
  - `Consumer` now `ApiAuthed`; first connection attempt benefits from retry.
  - Added `connect_logical_replication_stream` to validate publication/slot and open the stream; resource is re-resolved each attempt to pick up credential rotations.
  - Disabling kept only for unrecoverable cases (`Error::BadConfig` for missing publication/slot, or unparsable messages). Deleted resources now trigger retries instead of immediate disable.

<sup>Written for commit 008838fe2de9408428514bbec8799dca0a9838fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9666?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-trigger-postgres/src/listener.rs
+++ b/backend/windmill-trigger-postgres/src/listener.rs
@@ -7,10 +7,12 @@ use pg_escape::{quote_identifier, quote_literal};
 use rust_postgres::{Client, CopyBothDuplex, SimpleQueryMessage};
 use tokio::sync::RwLock;
 use tokio_stream::StreamExt;
+use windmill_api_auth::ApiAuthed;
 use windmill_common::{
     db::UserDB,
     error::{to_anyhow, Error, Result},
     jobs::JobTriggerKind,
+    utils::{report_critical_error, report_recovered_critical_error},
     worker::to_raw_value,
     DB,
 };
@@ -30,6 +32,10 @@ use super::{
 };
 
 const ERROR_REPLICATION_SLOT_NOT_EXISTS: &str = r#"The replication slot associated with this trigger no longer exists. Recreate a new replication slot or select an existing one in the advanced tab, or delete and recreate a new trigger"#;
+
+// Wait this long between reconnection attempts after a connection failure or a
+// dropped replication stream. Matches the Kafka trigger listener's backoff.
+const RECONNECT_DELAY_SECS: u64 = 30;
 
 pub struct LogicalReplicationSettings {
     pub streaming: bool,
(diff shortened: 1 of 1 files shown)


hugocasa:
Addressed the P2 (silent flap on connect-then-immediately-drop) in 008838fe2d by aligning the reconnection accounting with the SQS listener (`windmill-trigger-sqs/src/listener_ee.rs:272-295`):

- A stream close / receive error now **counts toward `tries`** and raises a throttled critical error — on the first occurrence (`tries % 10 == 0` at `tries == 0`), then every 10 attempts.
- `tries` is reset and recovery is reported **only after the reconnected stream delivers a message** (real progress), not on a bare (re)connect.

So a connection that connects and immediately drops every cycle now accumulates `tries` and alerts (first drop + every 10), instead of resetting to 0 each connect and ping-ponging silently. Genuine transient blips still recover quietly once the stream resumes.

`cargo check -p windmill-trigger-postgres` is clean. The trade-off (matching SQS): after a reconnect the ping shows the last "reconnecting" status until the first inbound message/keepalive arrives (~wal_sender_timeout/2 for an idle trigger), since ping/recovery now clear on progress rather than on connect.

PR #9684, opened 2026-06-19
fix(frontend): group live pipeline runs in the activity panel

rubenfiszel:
## Summary

In the data-pipeline activity panel, runs launched **manually through the UI** (and their downstream cascade) showed up as separate ungrouped rows, while scheduled/historic cascades grouped correctly.

The panel groups runs into cascades by connected components of the **dispatch-edge graph**. Run *events* are kept live by the `activeRunnables` poll, but the `edges` came only from `usePipelineHistory`'s **one-shot preload** (refetched solely on workspace/folder/days change). Since `dispatch_event` rows are written server-side *when a producer completes*, a run you launch and watch live has its producer + freshly-dispatched children appear as live events with **no edge connecting them** → union-find puts each in its own component → flat ungrouped rows.

The fix keeps the dispatch edges live: an edges-only refetch is triggered whenever the live poll's event id-set changes (a newly-dispatched child is a new id, which is exactly when fresh edges exist), so live cascades converge to grouped just like historic ones.

## Changes

- `pipelineHistory.svelte.ts`: add a lightweight edges-only `refetchEdges()` (single query, no completed-jobs paging; skips while a full `load()` is in flight; guarded by the same generation counter so stale-scope writes are dropped).
- `pipeline/[folder]/+page.svelte`: add an `$effect` that re-pulls edges when the live poll's event id-set changes, so a just-launched producer and its dispatched children collapse into one grouped cascade row.

## Screenshots

Live verification on the local `data-pipelines/orders` pipeline — triggered manual (`run`) producer runs and confirmed each appears as a grouped cascade (`run · 7 runs`, `run · 12 runs`) with expanded members, matching the older `schedule · 19 runs` group below. Before this fix these rendered as flat standalone rows.

pipeline-activity-grouped-live

## Test plan

- [ ] Open a pipeline folder in view mode, hit Run on a producer that cascades into children, and confirm the producer + freshly-dispatched children collapse into a single grouped cascade row in the Activity panel once the live poll surfaces them (rather than separate rows).
- [ ] Confirm the manual (`run`) cascade groups identically to a scheduled cascade.
- [ ] Switch the activity window (e.g. 1h → 30d) and change folders while a run is live; confirm edges stay scoped to the selected folder/window and don't show stale grouping.
- [ ] `npm run check` passes.

---
## Summary by cubic
Fixes ungrouped live cascades in the Activity panel by live-refetching dispatch edges when new run events appear. Manual runs now group like scheduled runs.

- **Bug Fixes**
  - Added edges-only dispatch-edge refetch in pipeline history with same-scope sequencing (`edgeSeq`); exposed `refetchEdges()`.
  - Trigger refetch when the live poll’s event id set changes, with guards to avoid clobbering full loads and to respect workspace/folder/window scope.

<sup>Written for commit ab0705dfd7abe76faf0423cc274bfc34f0e59e77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/assets/AssetGraph/pipelineHistory.svelte.ts
+++ b/frontend/src/lib/components/assets/AssetGraph/pipelineHistory.svelte.ts
@@ -43,6 +43,10 @@ export function usePipelineHistory(
 	// Generation counter: a folder/days change mid-flight must not write a
 	// stale response into the new scope (mirrors the page's bodyFetchGen).
 	let gen = 0
+	// Edges-only refetches fire one-per-new-id during a cascade; `gen` only
+	// guards scope changes, so this sequences same-scope calls — a slower
+	// earlier response must not overwrite a newer one.
+	let edgeSeq = 0
 
 	async function load(ws: string, prefix: string, days: number) {
 		const myGen = ++gen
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Addressed the two P2 review nits in `ab0705d`:

- **Out-of-order edge responses (Claude):** added a monotonic `edgeSeq` counter to `loadEdges`. Each refetch captures `mySeq = ++edgeSeq` and drops its write if `mySeq !== edgeSeq` after the await — so during a wide cascade's per-new-id refetch burst, a slower earlier response can no longer overwrite a newer one. This complements `gen` (which only guards scope changes).
- **Duplicated/overlong comment (Codex):** kept the canonical server-side-timing rationale in `loadEdges` (now 4 lines) and trimmed the page-effect comment to its own invariants (id-keyed signature + non-reactive `lastLiveEventSig` loop guard), removing the repeated explanation.

`npm run check:fast` clean on both touched files (the 14 reported errors pre-exist on `main` in unrelated `workspaceDrafts.svelte.ts`).

PR #9686, opened 2026-06-19
feat(jobs): auto-grant approvers a run-detail view link on the approval page

rubenfiszel:
## Summary

The runs page was recently made confidential: a member who didn't launch a run can no longer see it unless someone shares a read-only `view_token` link (PR #357b0efcc8 / the "Share" button on the run page). This left a gap on the **approval page** — the "Open run details" link pointed at `/run/{flow_id}` with no token, so a workspace-member approver who didn't launch the flow would hit the 403 "ask for a share link" wall, even though they're already trusted enough to approve/deny it.

This change mints a `view_token` for the parent flow on both approval endpoints and threads it into the "Open run details" link, so an authenticated workspace-member approver can open the run detail of the flow they're approving.

The token only **grants read** (never resume), and is only **consumable by an authenticated workspace member** (`validate_view_token` is reached exclusively via `require_job_read_access`, which requires `ApiAuthed`). So it never widens what the approver can already do, and an anonymous holder of the approval link cannot use it.

## Changes

- **Backend** (`backend/windmill-api/src/jobs.rs`):
  - `get_suspended_job_flow` (legacy `/jobs_u/get_flow/{job}/{resume}/{secret}`): after `verify_suspended_secret` passes, mint `view_token = {flow_id}.{hmac}` and return it in `SuspendedJobFlow`.
  - `get_approval_info` (new `/jobs_u/flow/approval_info/{job_id}`): mint `view_token = {flow_id}.{hmac}` in `ApprovalInfo`, but only on the authorized (`can_view`) path — the locked (`user_auth_required && !can_approve`) early-return returns `view_token: None`.
- **OpenAPI** (`openapi.yaml`): document the new optional `view_token` field on both responses.
- **Frontend** (both approval pages): append `&view_token=…` (URL-encoded) to the "Open run details" link when the token is present.

## Screenshots

A workspace member with **approval rights but no read ACL** on the flow opens the run detail via the approval-page link (would otherwise be a 403):

run-with-viewtoken

## Test plan

Verified end-to-end against a suspended flow, as a non-admin member who did not launch the flow:

- [x] Both approval endpoints return `view_token` (`{flow_id}.{hmac}`)
- [x] Member reads the flow **without** token → `403` ("ask for a share link")
- [x] Member reads the flow **with** token → `200`
- [x] Both approval pages render the "Open run details" link with the `view_token` query param
- [x] Member clicks the link and sees the run detail (no permission-denied state)
- [x] **Anonymous** caller + token → `400`/blocked (token requires authentication; run not revealed)
- [x] Backend compiles; frontend type-checks (no new errors)

## Notes

- No EE companion PR — all touched files are core (no `*_ee.rs`).
- The generated client under `frontend/src/lib/gen/` is gitignored and regenerated from `openapi.yaml`.
- Optional follow-up suggested in review: a small Rust assertion that `get_approval_info` returns `view_token: None` on the locked path and `Some` otherwise, to lock in the gating.

---
## Summary by cubic
Auto-grant a read-only `view_token` on approval pages and show a clear “Open run details” button for logged-in workspace members, so approvers can open run details without a 403. Also fixes the run-details link when job data isn’t available.

- **New Features**
  - Backend: Mint `view_token` for the parent flow in `get_approval_info` (only for authorized approvers) and `get_suspended_job_flow` (after secret validation).
  - Frontend: Append `view_token` to the run-detail URL on both approval pages; show a Button for logged-in workspace members and fall back to a link otherwise.
  - OpenAPI: Document optional `view_token` on both responses.

- **Bug Fixes**
  - Build the run-details URL from route params on the new approval page to avoid `/run/undefined` when `getJob` is denied.

<sup>Written for commit e0f2df0e2e1e7c8798dc1bb7e158cf538da74e12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9686?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api/openapi.yaml
+++ b/backend/windmill-api/openapi.yaml
@@ -13598,6 +13598,12 @@ paths:
                           type: integer
                         approver:
                           type: string
+                  view_token:
+                    type: string
+                    description: >-
+                      Share-read-link token for the flow. An authenticated workspace
+                      member can append it as a `view_token` query param on the run
+                      page to read a flow they don't otherwise have access to.
 
   /w/{workspace}/jobs_u/resume/{id}/{resume_id}/{signature}:
     get:
@@ -13849,6 +13855,13 @@ paths:
                       required:
                         - resume_id
                         - approver
+                  view_token:
+                    type: string
+                    description: >-
+                      Share-read-link token for the parent flow. An authenticated
+                      workspace member can append it as a `view_token` query param
+                      on the run page to read a flow they don't otherwise have
+                      access to.
                 required:
                   - job
                   - approvers
--- a/backend/windmill-api/src/jobs.rs
+++ b/backend/windmill-api/src/jobs.rs
@@ -3204,6 +3204,12 @@ struct ApprovalInfo {
     #[serde(skip_serializing_if = "Option::is_none")]
     hide_cancel: Option<bool>,
     approvers: Vec<Approval>,
+    /// Share-read-link token for the flow, minted only for callers allowed to view this
+    /// approval. Lets an authenticated workspace-member approver open the run details of
+    /// a flow they don't otherwise have read access to (the run page reads it as a
+    /// `view_token` query param).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    view_token: Option<String>,
 }
 
 /// Whether `opt_authed` is allowed to approve — and therefore view — this approval step.
@@ -3426,6 +3432,7 @@ async fn get_approval_info(
             user_auth_required,
             hide_cancel: None,
             approvers: vec![],
+            view_token: None,
         }));
     }
 
@@ -3443,6 +3450,12 @@ async fn get_approval_info(
     })
     .collect();
 
+    // Possession of view rights over this approval is sufficient to mint a
+    // share-read-link token for the flow: it only grants read (no resume), and only to
+    // an authenticated workspace member, so it never widens what the approver can do.
+    let hmac = generate_view_token(&w_id, row.id, &db).await?;
+    let view_token = Some(format!("{}.{hmac}", row.id));
+
     Ok(Json(ApprovalInfo {
         flow_id: row.id,
         form_schema,
@@ -3454,6 +3467,7 @@ async fn get_approval_info(
         user_auth_required,
         hide_cancel,
         approvers,
+        view_token,
     }))
 }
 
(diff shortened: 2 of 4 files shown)


rubenfiszel:
## Follow-up: clearer run-details link for logged-in members

Added on top of the automatic `view_token`: the approval page now detects whether the visitor is a **logged-in member of the workspace** (via `getUserExt(workspace)`), and if so renders a clear, ready-to-use **"Open run details"** button (no "(require auth)" caveat) — since the view token guarantees they'll be able to open it.

- **Logged-in member** → prominent `accent-secondary` Button "Open run details" (token baked into the link).
- **Anonymous / non-member** → unchanged subtle "Open run details (require auth)" text link.

Applied to both approval pages (legacy `…/{resume}/{hmac}` and the new token-based route).

### Screenshot (logged-in member view)

approval-member-button

### Verified
- [x] Logged-in member sees the clear Button; anonymous visitor sees the subtle "(require auth)" link (both confirmed via Playwright).
- [x] svelte-check passes (no new errors); svelte-autofixer clean.

rubenfiszel:
## Addressed CI review

**[P1] Broken run-detail link when `getJob` is denied** (Codex / Pi) — fixed in e0f2df0.

On the new approval page, `getJob` is fire-and-forget and gets denied for exactly the approver this link serves (member with approval rights but no direct run read access), leaving `job` undefined → `/run/undefined?workspace=undefined`. `runDetailsHref` now builds from the always-present route params `page.params.job` / `page.params.workspace` — and `page.params.job` is the same flow id the `view_token` is minted for (`row.id`), so the token always matches the target.

The legacy page is intentionally left on `job?.id`: its `getSuspendedJobFlow` is the primary loader (page errors out if it fails, so `job` is always defined when the link renders), and its `page.params.job` is the *suspended step* job, not the flow the token covers.

### Verified (Playwright)
Reproduced the exact P1 scenario — logged-in member, no `?token=` param, no run read access → `getJob` denied → `job` undefined (confirmed: "Flow details" graph absent, Approve button present):
- [x] Link href is now `/run/{real-flow-id}?workspace=admins&view_token=…` (was `/run/undefined` before the fix)
- [x] Clicking it opens the run detail (member reads via the token)
- [x] `svelte-check` clean, no new errors

### On the optional backend test
Three reviewers suggested a Rust assertion locking in `view_token: None` on the locked path vs `Some` otherwise. The gating decision itself (`can_view` / `can_approve_step`) is already unit-tested in the `approval_view_gate_tests` module; the remaining `view_token` mint is a thin wrapper over the already-tested `generate_view_token`. Happy to add an integration assertion if preferred, but it'd need a full suspended-flow fixture for marginal coverage — flagging rather than adding by default.

PR #9689, opened 2026-06-19
feat: ducklake materialization for data pipelines

rubenfiszel:
## Summary

Adds **managed, versioned, incremental materialization** to data pipelines, built on the DuckLake substrate, without becoming dbt. A DuckDB script declares `// materialize ducklake://<name>/<table>` and Windmill owns the write around its single `SELECT`: idempotent, captured as a DuckLake snapshot, and per-partition tracked. Design notes: `docs/ducklake-materialization.md`; user docs: windmilldocs#1401.

**CE/EE boundary:** the materialization *engine* is CE. Automatic partition *resolution* and the *backfill* fan-out remain EE (already gated). The `materialized_partition` table + idempotent explicit-partition runs make the EE backfill runner a thin follow-up.

## Grammar (final, managed-by-default)

```
// materialize ducklake://<name>/<table>              → managed, replace (default)
// materialize ducklake://<name>/<table> key=<col>    → managed, merge
// materialize ducklake://<name>/<table> append       → managed, append
// materialize manual ducklake://<name>/<table>       → track-only escape hatch
```

There is no `wrap` keyword (managing the write *is* what `materialize` means) and `unique_key`/`append` are no longer standalone annotations — they're strategy options on the line. `// partitioned` stays separate (cross-cutting). Polyglot / multi-statement writes use the `wmll.ducklake` helpers.

## Changes

- **Parser** (`windmill-parser` + frontend mirror + shared parity fixtures): the `// materialize` grammar, a wrap-eligibility classifier, partition-scoped strategy codegen (replace / merge / append), and `build_wrap_blocks`.
- **Deploy validation** (`windmill-api-scripts`): managed `// materialize` is gated to DuckDB and validated as a single-trailing-SELECT at save time.
- **Executor** (`duckdb_executor.rs`): rewrites a managed script into the write (synthetic `ATTACH … AS _wm_target` resolved by the normal attach pass), captures the produced snapshot id, records state. `manual` records state only.
- **State + API** (migration + `windmill-common/materialization.rs` + `windmill-api-assets`): `materialized_partition` table, a CE recording/listing module (checked `query!`, offline cache included), and `GET /assets/partitions?path=`.
- **SDK**: Python `DucklakeClient.upsert_partition`/`append_partition`/`read`; TS `upsertPartition`/`appendPartition`.
- **UI**: a partition-status grid on ducklake asset nodes (`PartitionStatusGrid.svelte`), an enterprise-gated `BackfillRangeDialog.svelte`, and a new **"Materialized table"** output kind in the new-script picker that scaffolds a managed `// materialize` DuckDB script.

## Validation — exercised end-to-end against a live DuckLake

- Real `// materialize` runs (managed-by-default replace, and append) materialize into DuckLake, capture snapshots, and record `materialized_partition` rows; re-running a partition upserts the same row (idempotent). The `key=`/MERGE path reaches execution but hits a pre-existing DuckLake MERGE storage error (not introduced here) — flagged as a follow-up.
- The grid renders real partitions via the endpoint; the backfill dialog is correctly enterprise-gated; the new output kind appears in the picker and scaffolds the managed form (`// materialize` + single SELECT, no CREATE TABLE).
- `cargo test -p windmill-parser` (77 unit + Rust↔TS parity) green; `cargo check`/`SQLX_OFFLINE` green for common/api-assets/worker. Frontend: vitest parity (37) green, `check:fast` clean for touched files, svelte autofixer clean, TS `tsc` clean, Python AST clean.

## Screenshots

Partition-status grid (real materialized partitions):
grid-partitions

Enterprise-gated backfill dialog:
backfill-dialog

New "Materialized table" output kind in the new-script picker:
materialize-output-kind

## Not in this PR (follow-ups)

- The `key=`/MERGE DuckLake storage error (pre-existing; investigate the MERGE-on-fresh-partition write).
- EE backfill range runner behind `POST /assets/backfill` (seam ready).
- Polyglot managed materialize (the Python/TS helper takes a SQL SELECT, not in-memory rows — so the new output kind is DuckDB-only for now).
- `row_count` capture, `Running` pre-state, snapshot pinning, schema-drift handling, data tests / asset checks.

## Test plan

- [ ] `cargo test -p windmill-parser` passes.
- [ ] New ducklake script via the picker's "Materialized table" kind scaffolds `// materialize` + a single SELECT; deploying + running it materializes a partition and writes a `materialized_partition` row.
- [ ] Non-DuckDB `// materialize` → save-time BadRequest; non-single-SELECT DuckDB → targeted error.
- [ ] Select a ducklake asset → grid lists partitions; Backfill enterprise-gated.

materialize-result-preview

**Partitioned result** — the header shows the `partition` badge and labels the count "N rows in partition"; the preview hides the internal `_wm_partition` column.

materialize-partitioned

**Partition / whole-table toggle** — partitioned results can scope the preview to the just-written partition (hides the internal `_wm_partition` column) or show the whole table (all partitions, with the column).

materialize-toggle


rubenfiszel:
Addressed the CI failure and the review findings:

**`check-freshness`** — regenerated `system_prompts/auto-generated/` + `cli/src/guidance/skills.gen.ts` for the new `wmll.ducklake` SDK helpers, and fixed a stale `upsertPartition` doc that still said "MERGEs" (it's delete-by-key + insert).

**Claude P2 — stale `wrap`/`no-wrap` wording** — the `wrap` keyword was dropped (managed-by-default with `// materialize manual` as the escape hatch), but `WrapError::message()` deploy errors and the module/type doc comments still told users to "remove `wrap`" / "use no-wrap". Now they point at `// materialize manual`; the design doc (`docs/ducklake-materialization.md`) is synced too.

**Codex/Pi P1 — Backfill button → 404** — `POST /assets/backfill` has no backend route yet (planned EE follow-up), so under a license the button would 404. Disabled it with a "coming soon" tooltip until the endpoint lands; the dialog seam stays.

**Codex P1 — materialize targets not persisted as asset writes** — already handled at `scripts.rs:1327-1336` (the `// materialize` target is injected into `effective_assets` as a `ducklake` write before `replace_static_asset_usage` persists it), so server-authoritative deploy paths record the producer row. That commit predates the Codex snapshot; Pi's review confirms it on current `HEAD`.

rubenfiszel:
Addressed the three remaining P1s from the latest Codex/Pi/Claude rounds:

**P1 — DuckDB arg declarations stripped from managed materialize** (`f2c…`) — wrapping the SELECT runs through `split_statements`, which strips the `-- $name (TYPE)` declarations while the `$name` refs survive in the embedded SELECT → unbound at runtime. Per the intended design (managed materialize takes no SQL args — the partition is engine-supplied, not bound), the deploy path now **rejects** managed `// materialize` with declared args, pointing at the `{partition}` token / `// materialize manual`. Verified: `-- $min_date (DATE)` → 400, no-args → 201. (Threading the sig through was the alternative, but managed materialize shouldn't take arbitrary binds.)

**P1 — non-DuckLake managed targets deploy then silently no-op** + **P1 — manual polyglot never records state** — both fixed by making `// materialize` require **DuckDB + a `ducklake://<name>/<table>` target for *any* mode** (managed or manual), validated at deploy. A non-DuckLake target no longer registers a phantom producer (it 400s), a tableless `ducklake://main` 400s, and `// materialize manual` on a polyglot script 400s (so manual is DuckDB-only → always goes through the executor that records `materialized_partition` state). Verified all four: `datatable://` → 400, `ducklake://main` → 400, manual Python → 400, `ducklake://main/orders` → 201.

The Backfill 404 and the stale `wrap`/`no-wrap` wording were addressed in the prior round (Claude/Pi confirm).

rubenfiszel:
**P1 — agent workers didn't record materialization state** — implemented the HTTP recording path (your option A). `record_mat` now branches on the connection: `Sql` writes directly (unchanged), `Http` posts a `RecordMaterializationRequest` to a new `POST /api/w/{w_id}/agent_workers/record_materialization`, which reuses the same CE `record_materialization` upsert. Worker helper lives in `agent_workers.rs`; the endpoint is in the EE agent-worker router (companion **windmill-ee-private#622**, branched off the now-merged #621).

Verified: compiles with `agent_worker_server`; the route is registered (returns the same agent-auth response as the existing `get_ducklake` endpoint — the dev server can't complete the agent-auth handshake, but the route and handler are wired identically to the proven endpoints); the Sql path is unchanged and still records rows (snapshot + row_count). The recording is best-effort, so if the endpoint is unreachable it logs a warning and never fails the job.

rubenfiszel:
**P2 — `{partition}` token in the deploy message was unresolvable** — good catch, confirmed: the token was only resolved for `// on` asset URIs (`cascade_ee`), not the SQL body, so `WHERE date = '{partition}'` compared against literal text. Rather than walk back the suggestion, I made it real: `build_materialized_query` now substitutes `{partition}` (the same token) with the escaped partition value across the managed script's setup + SELECT. So the token means the same thing everywhere, and the error message + docs are now accurate. Verified end-to-end: a partitioned materialize with `WHERE day = '{partition}'` at partition `2026-01-02` materializes exactly that row (not 0).

rubenfiszel:
Both P1s fixed:

**`{partition}` injection** — the token is now always replaced by a *complete* escaped SQL literal (`'…'` with `'` doubled), whether or not the author quoted it (the pre-quoted `'{partition}'` form is matched first so it doesn't become `''…''`). A caller can no longer pass metacharacters that break out of the literal. Verified end-to-end: `WHERE day = {partition}` with `partition = "x'); DROP TABLE wm_tok2; --"` runs as a string comparison (0 rows) and the table is **not** dropped; both quoted and unquoted forms still filter correctly.

**SDK asset_path drops schema** — `_materialize_finish` (Python) and `finishMaterialize` (TS) now record `<lake>/<schema>.<table>` for an explicit schema (else `<lake>/<table>`), matching the `// materialize` engine and `/assets/partitions` lookups, so schema-qualified slices show in the grid and distinct schemas don't collide. (The summary read already used the schema-qualified table via `USE`, so only the recorded path needed fixing.)

rubenfiszel:
Also addressed the P2 on `DucklakeResultPreview`: the preview now carries the explicit schema into the `tableKey` (and the colDefs lookup), so a `ducklake://<lake>/<schema>.<table>` result reads the schema-qualified table instead of the default schema. (`FROM` is rendered unquoted, so `schema.table` resolves correctly under the ducklake `USE`.) Type-checks clean and svelte-autofixer reports no issues; mirrors the proven `DataTablePreview` schema handling.

rubenfiszel:
Fixed — the published entrypoint now exposes the helpers.

`client.ts` re-exported `upsertPartition`/`appendPartition`, but `src/index.ts` (the npm/jsr entrypoint) is assembled by `build.sh` / `build.jsr.sh` from hard-coded export lists — both the namespace re-export **and** the default `wmill` object — and neither listed the new helpers, so `import * as wmill` / `wmill.upsertPartition` were absent at runtime. Added them (plus the `DucklakeMaterializeOptions` / `SqlStatement` types) to both build scripts and refreshed the checked-in `client.d.ts` / `sqlUtils.d.ts`.

Verified by running `build.sh` then `tsc`: `dist/index.d.ts` now carries `upsertPartition`/`appendPartition` on both the namespace export and the default object (`upsertPartition: typeof upsertPartition`). Also swapped a `String.replaceAll` (ES2021) for a regex `.replace` so the prepublish `tsc` (lib ES2018) compiles cleanly.

rubenfiszel:
Fixed the doc on `record_materialization`: it now states the worker records the terminal `Materialized`/`Failed` outcome after the write, and notes `Running` is a pg-enum mirror with no writer in this flow — no longer promising a pre-write marker that doesn't exist yet.

PR #9693, opened 2026-06-20
feat: add SQL migrations for data tables

diegoimbert:
## Summary

Adds a SQL migration system for data tables: define versioned `.up`/`.down` SQL migrations per data table, apply/roll them back (tracked in each data table's own `_wm_migrations` table), manage them from the workspace settings UI, scaffold/run them from the `wmill` CLI, and sync them as workspace files.

Migrations are defined per workspace+data table in a new `datatable_migrations` table, and each application runs as a normal user-permissioned Windmill `postgresql` job (labelled `datatable_migration`, so runs show up in history).

## Changes

**Backend** (`windmill-api-workspaces`)
- New `datatable_migrations` table (PK `(workspace_id, datatable, timestamp)`, nullable `code_down`).
- Endpoints under `/w/{workspace}/workspaces/`: `run_datatable_migrations` (all pending, `?only=`/`?up_to=`), `rollback_datatable_migrations` (latest or `?only=`), `create`/`delete`/`upsert`/`list`/`update_datatable_migration(s)`, `datatable_migrations_status` (per-migration ran/not-run/unknown), and `generate_initial_datatable_migration` (snapshots current schema via `pg_dump`, marks it installed so it's never re-run, no down).
- Migrations applied/rolled-back as user-permissioned `postgresql` jobs; failures surface the executor's Postgres error (e.g. `relation "x" does not exist`) instead of an opaque envelope.
- Data table migrations are surfaced in the workspace tarball export as files.

**Frontend** (workspace settings → Data tables)
- "Migrations" button per data table opening a modal: per-migration status dot (green ran / orange not-run / grey unknown), per-row Run / Revert / Delete, "New" / "Run all" / "Generate initial migration".
- New-migration modal with Up/Down tabs + SQL editors and a "Create and run" (split) action that reverts the insert if the run fails.
- Out-of-order confirmation when running/reverting a migration that has un-run earlier / applied later migrations; delete always confirms.
- DDL guard prompting to capture DDL typed in the datatable SQL editors as a migration.

**CLI** (`wmill datatable migrate`)
- `new` scaffolds `migrations/datatable/<dt>/<ts>_<name>.up.sql` (+ `.down.sql`).
- `up` / `down` default to **all** data tables in the workspace; `--datatable` targets one.
- Migrations sync as workspace files (`migrations/datatable/<dt>/…`); `sync push` upserts/deletes them and offers to run new ones; invalid local migration sets are rejected on push.
- `.sql` migration files are excluded from script-metadata generation (no more "Invalid language: .sql").

## Screenshots

datatables
migrations-modal
new-migration

## Test plan
- [ ] Configure a data table; from settings → Data tables → Migrations, create, run, revert, and delete migrations; verify status dots + out-of-order/delete confirmations.
- [ ] "Generate initial migration" on an existing-schema data table → it appears as already-applied with no down migration.
- [ ] Run a migration with bad SQL → toast shows the Postgres error detail.
- [ ] `wmill sync pull` writes files under `migrations/datatable/<dt>/`; `wmill datatable migrate new/up/down` work; `up`/`down` with no flag hit all data tables.
- [ ] `wmill sync push` upserts/deletes migrations and offers to run new ones.


diegoimbert:
_wm_migrations collision is okay, the point of that table is that it's intrinsic to its current database and not the Windmill configuration
It is the responsibility of the user to not do it
Also insanely unlikely to land on the same timestamp twice for different migrations

rubenfiszel:
@diegoimbert even if that would be ok, why not namespace, what's the benefit ?

PR #9698, opened 2026-06-21
feat(copilot): improve global-mode path selection + add path-selection evals

centdix:
## Summary

Global mode (the workspace-wide AI chat) had no working way to pick an `f/<folder>/…` path when the user doesn't give one. The system prompt told the model to discover folders by calling `list_workspace_items` with `type: ['folder']`, but that type isn't accepted (it throws), there is no folder-listing tool, and the prompt's "every workspace path has exactly three segments" was factually wrong (names can contain `/`). In practice the model couldn't target an existing folder, didn't ask when shared intent had no matching folder, and invented non-existent folders.

This PR (1) adds `ai_evals` coverage that reproduces the issue, and (2) fixes the prompt to inject the user's folders directly from `userStore` (no extra backend call) and corrects the path conventions.

## Changes

- **Prompt (`frontend/.../copilot/chat/global/core.ts`)**:
  - Rewrote the Path-conventions block — removed the broken `list_workspace_items type:['folder']` instruction, corrected "exactly three segments" → "three or more (the trailing name may contain `/`)".
  - New `buildFolderGuidance` injects folder context from `userStore`:
    - **Non-admins**: their writable folders, plus read-only folders flagged as off-limits.
    - **Admins**: "you can write to any existing folder; if the user names one, use it, otherwise ask" + their explicitly-permissioned folders as a **non-exhaustive hint** (an admin's `userStore.folders` is only the explicit-perms subset since admins bypass folder ACLs; the complete set needs a folder-listing tool — see follow-ups).
  - Capped at 40 folders.
- **Eval harness (`ai_evals/adapters/frontend/core/global/globalEvalRunner.ts`, `ai_evals/modes/global.ts`)**: the harness never seeded `userStore`, so the prompt saw an empty username (`u//<name>`). Added a `user` field to the case `initial` fixture that seeds `userStore` (username / is_admin / folders / folders_read) before the system message is built. `userStore` is a process-global singleton → run these cases with `--verbose` (concurrency 1).
- **Eval cases (`ai_evals/cases/global.yaml` + 3 fixtures)**: 4 `global-path*` cases.

## Test plan / results

`sonnet`, baseline (before) vs this PR (after the prompt fix):

| Case | Baseline | This PR | Result |
|---|---|---|---|
| path1 — bare name → personal | `u/admin/...` | `u/admin/...` | ✅ → ✅ |
| path2 — `marketing` folder exists | `u/admin/...` | `f/marketing/...` | ❌ → ✅ |
| path3 — shared intent, no matching folder | invented `f/people_ops` | `u/admin/...` (still not "ask") | ❌ → ❌ |
| path4 — non-admin, `team_b` read-only | invented `f/team` | `f/team_a/...` | ❌ → ✅ |

**Pass rate 1/4 → 3/4.** path2/path4 also collapsed from 5–7 tool calls (futile discovery loops) to just `get_instructions` + `write_flow`. (Folder handling was iterated: a plain folder list scored 3/4 but misled superadmins; the admin-aware hybrid keeps 3/4 *and* gives superadmins correct guidance.)

Reproduce:
```bash
cd ai_evals
WMILL_AI_EVAL_BACKEND_URL=<backend-url> WMILL_AI_EVAL_BACKEND_WORKSPACE=<workspace> \
  bun run cli -- run global \
  global-path1-bare-name-defaults-to-personal \
  global-path2-match-existing-folder \
  global-path3-shared-intent-unknown-folder-asks \
  global-path4-nonadmin-avoids-readonly-folder \
  --model sonnet --verbose
```

- [ ] `npm run check` (frontend types)
- [ ] re-run the 4 `global-path*` cases and confirm 3/4 (path3 is a known follow-up)

### No visible UI
The only `frontend/` change is an LLM system-prompt string — there is no rendered UI change to screenshot. It is verified behaviorally through `ai_evals` (above).

## Follow-ups

- **path3 (ask vs default for shared intent)**: for "an onboarding flow for the People Ops team" (shared intent, no matching folder), the model defaults to `u/admin` instead of calling `askUserQuestion`. Needs a stronger ask-for-shared-intent rule; re-run path3 after.
- **`list_folders` tool for complete admin folder set**: the injected admin hint is the *explicit-perms* subset; a dedicated folder-listing tool hitting `/folders/list` returns all folders for admins (they bypass ACLs). Needs mock-backend folder support to evaluate.
- **(local only, not in this PR)** a temporary CE workspace-cap bypass in `windmill-api-workspaces/src/workspaces.rs` was used to stand up test workspaces; it is intentionally excluded.

### DIFF
--- a/ai_evals/adapters/frontend/core/global/globalEvalRunner.ts
+++ b/ai_evals/adapters/frontend/core/global/globalEvalRunner.ts
@@ -1,9 +1,7 @@
 import { mkdtemp, rm } from "fs/promises";
 import { tmpdir } from "os";
 import { join } from "path";
-import { get } from "svelte/store";
 import type { AIProvider } from "$lib/gen/types.gen";
-import { userStore, type UserExt } from "$lib/stores";
 import {
   globalTools,
   prepareGlobalSystemMessage,
@@ -55,36 +53,15 @@ export interface GlobalLiveEditorDraftFixture {
 // Identity the global system prompt builds paths from. Production reads
 // `userStore` (whoami) to fill `u/{username}/...`; the eval harness never logs
 // in, so without this the prompt sees an empty username (`u//...`) and no
-// path-selection case is meaningful. Seeded per-case via the initial fixture.
+// path-selection case is meaningful. Seeded per-case via the initial fixture and
+// passed straight to `prepareGlobalSystemMessage` (no global-store mutation).
 export interface GlobalUserFixture {
   username: string;
   is_admin?: boolean;
-  is_super_admin?: boolean;
-  operator?: boolean;
   /** Folders the user can write to (the writable set whoami returns). */
   folders?: string[];
   /** Folders the user can read; read-only folders = folders_read \ folders. */
   folders_read?: string[];
-  /** Folders the user owns (subset of folders). */
-  folders_owners?: string[];
-}
-
-function buildEvalUserExt(user: GlobalUserFixture): UserExt {
-  return {
-    email: `${user.username}@windmill.dev`,
-    username: user.username,
-    is_admin: user.is_admin ?? true,
-    is_super_admin: user.is_super_admin ?? false,
-    operator: user.operator ?? false,
-    created_at: "2024-01-01T00:00:00.000Z",
-    groups: ["all"],
-    pgroups: ["g/all"],
-    folders: user.folders ?? [],
-    folders_owners: user.folders_owners ?? user.folders ?? [],
-    // folders_read is consumed at runtime / by the Phase 2 prompt injection but
-    // is not part of the typed UserExt surface, so attach it via the cast below.
-    ...(user.folders_read ? { folders_read: user.folders_read } : {}),
-  } as UserExt;
 }
 
 export interface GlobalEvalResult {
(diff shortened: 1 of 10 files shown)


centdix:
Addressed all review feedback in `747e5d2d24`:

- **`folders_read` accessed via inline cast** (Claude P2): went with the clean fix — added `folders_read` to the `User` schema in `openapi.yaml` (required, alongside its `folders`/`folders_owners` siblings from the same `UserInfo` struct) and regenerated the client, plus added `folders_read: string[]` to `UserExt`. Both inline casts (in `core.ts` and the eval harness) are now gone and the field is read typed.

- **`userStore` singleton race under default concurrency** (Codex P2): removed the global-store mutation entirely. `prepareGlobalSystemMessage` now takes the user identity explicitly (`{ user }` opt); the eval harness passes it through instead of `userStore.set(...)`. No shared mutable state → race-free at any concurrency, so the `--verbose`/concurrency-1 requirement is gone (deleted `buildEvalUserExt` and the set/restore dance).

- **Stale case comment** (Codex P2): rewrote the path-selection comment in `cases/global.yaml` as a current invariant (what each case asserts), dropping the pre-fix "expected to fail / will pass once the fix lands" narration and the now-obsolete concurrency note.

- **Test coverage** (both): added a `folder guidance` describe block to the existing `core.test.ts` — deterministic cases for non-admin writable / read-only split / empty, admin hint / admin-no-folders, the 40-folder cap, and the no-user case.

Re-ran the 4 `global-path*` cases with `sonnet` at **default concurrency** (no `--verbose`): still **3/4**, identical paths — confirming the refactor changed no behavior and the race is gone. `path3` (ask-vs-default for shared intent with no matching folder) and a dedicated `list_folders` tool for complete admin folder discovery remain the scoped follow-ups.

PR #9705, opened 2026-06-22
fix: prevent variable push from corrupting is_secret variables

hugocasa:
## Problem

`wmill variable push <file> <path>` could silently corrupt an `is_secret: true` variable. Pushing a spec file whose value is plaintext, without `--plain-secrets`, made the CLI send the value with `already_encrypted=true`. The server then stored the plaintext verbatim in the encrypted `value` column, so every subsequent read failed to decrypt:

```
Error decrypting variable f/x/cfg: Invalid byte 58, offset 9.
```

The only recovery was re-saving through the UI. The root cause is an ambiguity the CLI can't resolve on its own: without the workspace key it cannot tell plaintext from ciphertext, so it relied entirely on the flag.

## Fix

**Server (the durable guard, protects every client).** `create_variable` / `update_variable` now validate that a secret submitted with `already_encrypted=true` is actually well-formed workspace ciphertext before storing it verbatim, returning a clear `400` otherwise. The check is purely structural (standard base64 decoding to a non-zero multiple of the AES block size) and never decrypts; external secret-backend markers (`$vault:`/`$aws_sm:`/`$azure_kv:`) pass through untouched. Legitimate ciphertext round-trips (incl. `sync push` and workspace-key migration, where the key is pushed first) are unaffected.

**CLI.** Single-file `variable push` now decides how to treat a secret's value by inspecting its shape, instead of blindly trusting the flag.

### Behavior change: single-file `variable push`

This applies **only** to `wmill variable push <file> <path>`. Bulk `sync push` is unchanged.

| Secret value in the file | Before | After |
|---|---|---|
| Plaintext, no `--plain-secrets` | Sent as `already_encrypted=true` → **stored verbatim → bricked** | Detected as plaintext → sent for server-side encryption ✅ |
| Ciphertext shape (round-tripped from `sync pull`), no `--plain-secrets` | Stored verbatim | Stored verbatim (unchanged) |
| Any value, with `--plain-secrets` | Encrypted server-side | Encrypted server-side (unchanged) |

"Ciphertext shape" = a `$vault:`/`$aws_sm:`/`$azure_kv:` marker, or standard base64 decoding to a non-zero multiple of the AES block size (what `sync pull` writes for a secret without `--plain-secrets`). When the value isn't that shape, it must be plaintext a human authored, so the CLI lets the server encrypt it (and prints a one-line notice). `--plain-secrets` still forces the plaintext path.

Net effect: every flow that worked before still behaves identically; only the previously-corrupting case (plaintext without the flag) now does the right thing. The lone irreducible edge — plaintext that *coincidentally* has ciphertext shape — still needs `--plain-secrets`, and the server guard prevents silent corruption either way.

## Also fixed

- **`is_secret` true->false was a silent no-op in single-file `variable push`.** It reported success but never flipped the flag. It now applies `is_secret` in both directions. Bulk `sync push` keeps its prior upgrade-only behavior — a pulled secret's spec value is ciphertext, so a bulk downgrade would store that ciphertext verbatim as a visible non-secret value; downgrades are gated to the authoritative single-file push (`allowSecretDowngrade`).
- **Resolution banner polluted stdout.** The per-command "Using workspace profile ..." banner (and sibling resolution diagnostics) went to stdout, so `wmill variable get path > file` produced an unparseable file. These now go to stderr; command data stays on stdout.

## Not reproduced

The reported false `Network error` after a successful write does not reproduce on current `main`: `variable add` (create + update) and `variable push` all exit `0` with no spurious error and the write succeeds. The `--yes` flag for `variable add` already exists.

## Testing

- Backend unit tests for the guard (genuine ciphertext accepted; plaintext / non-block-aligned / empty rejected; external markers passed through) — 4/4.
- CLI integration tests: plaintext secret `push` round-trips correctly; `is_secret` true->false flip; existing sync/resource push tests still pass — 6/6.
- CLI unit test for `looksLikeWorkspaceCiphertext` (CLI/server shape agreement) — 2/2.
- Manual end-to-end against a live backend (below). No new SQL — sqlx cache unchanged.

### Demo (after fix)

```
$ cat /tmp/v.yaml
value: |
  some: plaintext
is_secret: true
description: ""

$ wmill variable push /tmp/v.yaml f/x/cfg          # no --plain-secrets, value is plaintext
Pushing variable...
Secret value is not in encrypted form; pushing as plaintext to be encrypted server-side (pass --plain-secrets to silence).
Variable f/x/cfg pushed

$ wmill variable get f/x/cfg --json | jq .value      # decrypts cleanly
"some: plaintext\n"
```

A mislabeled write from any client now fails loudly instead of bricking the secret:

```
$ curl -X POST ".../variables/update/f/x/cfg?already_encrypted=true" -d '{"value":"some: plaintext\n"}'
Bad request: Variable f/x/cfg was sent as already-encrypted (already_encrypted=true) but its value is not valid workspace-encrypted ciphertext. ...
HTTP 400
```


hugocasa:
Addressed the `auto-review` P2 (`is_secret` change leaking into `sync push`):

The bidirectional `is_secret` flip is now gated behind a new `allowSecretDowngrade` param on `pushVariable`, which only single-file `variable push` sets. Bulk `sync push` keeps its original **upgrade-only** behavior, so the demotion-of-a-ciphertext-valued-secret path the review described can't occur via sync — `sync push` is genuinely unchanged now, and the PR description has been corrected to say so. Single-file `variable push` (where the value is hand-authored plaintext) still flips both directions.

Commit: 47c1458

PR #9706, opened 2026-06-22
fix(frontend): strip raw-app post-deploy diff noise (raw_app/lock/data)

Guilhem-lm:
## Summary

Deploying a raw app from the raw-app editor showed a spurious diff in the Diff drawer immediately afterwards, even with no edits — the deployed side carried `raw_app: true`, a server-recomputed inline-script `lock`, and an empty `data` the editor side lacked. The session/compare draft diff had the same noise plus a nested-vs-flat shape mismatch (the deployed app row nests its source under `value`, while a draft stores `files`/`runnables`/`data` at the top level).

Both diff surfaces now run their values through the same normalizer, so only real changes surface.

## Changes

- **`raw_apps/utils.ts`** — add `stripRawAppDiffNoise` (deep-clones, then drops server-managed columns `raw_app`/`id`/`created_at`/`created_by`/`versions`/`extra_perms`/`is_draft`, nulls inline-script `lock`s, and canonicalizes `data` so absent ≡ default `{tables: []}`), and `canonicalRawAppDiffValue` = `stripRawAppDiffNoise(appSourceToDraftValue(x))` which additionally collapses the nested-vs-flat shape difference.
- **`RawAppEditorHeader.svelte`** — apply `stripRawAppDiffNoise` symmetrically to the deployed side (`syncWithDeployed`) and the editor side (a new `$derived currentDiffValue`), used by the Diff button, the deploy-conflict gate, and `DeployOverrideConfirmationModal`.
- **`utils_draft_deploy.ts`** — in `getDraftDiffValues` (the shared source behind both `CompareDrafts` and the sessions `DraftDiffDrawer`), route both sides of a raw-app diff through `canonicalRawAppDiffValue` when `kind === 'raw_app' || r.raw_app === true`.
- **`raw_apps/utils.test.ts`** — unit tests for both new functions (noise stripping, no input mutation, flat/nested handling, deployed≡draft when content matches, real change still surfaces).

Deliberately **not** touched: `getItemValue` (it feeds real deploy/merge ops, not just diffs) and the fork-to-fork path (`ForkDiffDrawer`), which is symmetric — both sides are deployed rows, so the noise cancels.

## Screenshots

Raw-app editor **Diff** button — the only highlighted change is the summary edit; `data`/`files` are clean, no `raw_app`/`lock`:

editor-diff-clean

Sessions / Compare-drafts diff (`getDraftDiffValues`) — same canonical flat shape on both sides (no `value:` wrapper, no `raw_app`/`lock`):

compare-drafts-clean

## Test plan

- [ ] Deploy a raw app unchanged, then open the editor **Diff** button → reports no changes (no `raw_app`/`lock`/`data` rows).
- [ ] Edit only the summary, open **Diff** → only the summary differs.
- [ ] Open the sessions draft-diff / `/forks/compare?...&mode=draft` for a deployed-with-draft raw app → flat shape on both sides, only real edits surface.
- [ ] Repeat for a never-deployed (draft-only) raw app.
- [ ] `cd frontend && npx vitest run src/lib/components/raw_apps/utils.test.ts` passes.


Guilhem-lm:
Thanks for the reviews 🙏

- **Codex P2 (deploy-drawer Diff bypassed canonicalization)** — fixed in 76ad4f7. The Deploy → Diff action now uses `currentDiffValue` (and a stripped `savedApp` fallback), matching the top-level Diff button. Verified in-browser: opening Deploy → Diff on an unchanged raw app now reports "No changes detected" with no `lock`/`data` noise.
- **Pi P2 (`stateSnapshot` redundant in `stripRawAppDiffNoise`)** — keeping it deliberately. `currentDiffValue` passes `{ value: app, … }` where `app`'s nested `files`/`runnables`/`data` are the editor's reactive `$state`, so the snapshot is doing real work, and it matches the codebase convention (`structuredClone(stateSnapshot(...))`, e.g. `savedApp` in this same file and `cleanValueProperties`).

`check:fast` and the `utils.test.ts` suite remain green.

PR #9709, opened 2026-06-22
feat: ducklake time-travel UX (snapshot history + AT VERSION reads)

rubenfiszel:
## Summary

Makes DuckLake's per-materialization snapshots a **first-class, discoverable time-travel surface**. Every `// materialize` write already records a DuckLake `snapshot_id`; this PR lets users actually *use* those snapshots — read a table as of any past version, see the snapshot history, and copy the exact `AT (VERSION => n)` syntax — covering the three things people reach for: debugging a past run, rolling back, and ad-hoc experimentation on a historical state.

### Scope decision (important)

The original task was to thread each producer's `snapshot_id` automatically through the cascade trigger blob and inject `AT (VERSION => $WM_UPSTREAM_SNAPSHOT)` into consumer reads. After discussion we **deliberately deferred** that automatic cascade pinning and shipped the explicit, user-driven surface instead. Rationale (recorded in `docs/ducklake-materialization.md` → "Reproducibility"):
- The only thing auto-pinning adds over the explicit surface is *automatic per-run consistency* — protection against an upstream re-materializing mid-run. That race only bites high-frequency event-driven cascades (rare today) and is never corruption (DuckLake reads are whole ACID snapshots), so it is **not critical**.
- Auto-injecting/stripping an `AT` clause is invisible magic; the explicit clause is inspectable.
- Multi-upstream consumers need a per-ref snapshot *map* accumulated across the AND-join, which is EE — a single `$WM_UPSTREAM_SNAPSHOT` would silently mis-pin.

This surfacing is the foundation auto-pinning would build on, if a workload ever shows the consistency race in practice.

## Changes

**Backend** (`backend/windmill-common/src/query_builders.rs`)
- `SelectOptions`/`SelectPayload`/`CountPayload` gain an optional `version: Option<i64>`. When set (DuckDB only), `make_select_query`/`make_count_query` emit `FROM <table> AT (VERSION => n)`. `n` is a server-controlled `i64` deserialized by serde, so it's injection-safe; the suffix is emitted only in the `DbType::Duckdb` arms (never leaks into other dialects).
- New `WM_INTERNAL_DB_DUCKLAKE_SNAPSHOTS` marker op → `SELECT snapshot_id, snapshot_time FROM ducklake_snapshots('dl') ORDER BY snapshot_id DESC` (catalog-wide commit log), wrapped with the existing ducklake ATTACH path.
- 6 new unit tests (incl. the negative `test_version_ignored_for_non_duckdb` and unpinned-fallback cases).

**Frontend**
- `dbOps.ts`: threads `version` into the SELECT/COUNT markers; adds `fetchDucklakeSnapshots`.
- New `DucklakeAssetPanel.svelte` — tabbed Partitions / History / Query panel; replaces the bare partition grid on the ducklake asset node (`AssetGraphDetailsPane.svelte`).
- New `DucklakeSnapshotHistory.svelte` — snapshot timeline (id + time), copy `AT (VERSION => n)`, copy full `FROM … AT (VERSION => n)`, and "Query this version".
- New `DucklakeVersionPreview.svelte` — read-only grid of the table at a chosen version, with the equivalent SQL shown above it.
- `pipelineTemplates.ts`: scaffolds a `-- time-travel: …` hint when a ducklake input is read.

**Docs**: `ducklake-materialization.md` rewritten to record the shipped surface + the deferred-cascade-pinning rationale.

## Validation

- ✅ Backend: 6 new `query_builders` unit tests pass (`cargo test -p windmill-common query_builders::`); `windmill-common` compiles; the only runtime consumer (worker) uses the unchanged `try_expand_internal_db_query`; `make_count_query` has no external callers.
- ✅ Frontend: `npm run check` clean on all changed files (the repo's 67 pre-existing `Timeout`-vs-`number` errors are unrelated); svelte-autofixer clean on all 3 new components; dev bundle hot-reloads them without error.
- ✅ `local-review` pass: one P2 (redundant metadata refetch on version switch) found and fixed.

## Screenshots

**Not captured** — and I want to be explicit about why rather than skip silently. The interactive flow can't be exercised on this dev box: there's **no object store / MinIO** configured, so a DuckLake `// materialize` write (which goes through the S3 proxy to `s3://_default_/`) can't run, meaning no real table/snapshots exist to render the History/Query tabs against. I got as far as creating a ducklake-materialize pipeline draft to produce an asset node, but the no-storage environment + the alpha pipeline-insert flow blocked completing it. The rendering logic is covered by type-check + svelte-autofixer; the SQL codegen by unit tests.

## Test plan

On an instance with an object store + a configured ducklake, after ≥2 materialize runs of a table:
- [ ] Select the ducklake asset node → panel shows **Partitions / History / Query** tabs.
- [ ] **History**: snapshots list newest-first with id + timestamp; copy-clause / copy-`FROM` / Query buttons work.
- [ ] **Query**: picking a version renders the grid read-only (no edit/delete/insert), with `FROM … AT (VERSION => n)` shown above; switching versions reloads rows and does **not** re-dispatch a metadata job (P2 fix).
- [ ] A consumer DuckDB script with `FROM dl.t AT (VERSION => n)` reads the pinned snapshot; removing the clause reads latest.
- [ ] Scaffolding a ducklake-input duckdb script emits the `-- time-travel:` hint.

---
## Summary by cubic
Adds DuckLake time-travel: browse snapshot history and read any table at a past version via AT (VERSION => n). Improves debugging, rollback, and experiments; automatic cascade pinning is deferred.

- **New Features**
  - Backend: `version` on SELECT/COUNT (DuckDB) emits `FROM <table> AT (VERSION => n)`; `DUCKLAKE_SNAPSHOTS` lists catalog snapshots and can scope to a schema‑qualified table.
  - Frontend: `DucklakeAssetPanel` merges History + Query into a master‑detail tab with a read‑only, version‑pinned grid; the preview shows and lets you copy `FROM lake.<table> AT (VERSION => n)`. `dbOps` threads `version`, adds `fetchDucklakeSnapshots` and `fetchDucklakeColumnsAtVersion`; `pipelineTemplates.ts` adds a time‑travel hint for DuckLake inputs.
  - Docs: Updated to match the History tab UI (master‑detail, copy full FROM) and document the deferred cascade pinning.

- **Bug Fixes**
  - Qualify copied time‑travel clauses with `lake.` and render `snapshot_time` correctly (microseconds → ms).
  - Scope snapshot history to versions where the table exists; keep the list pane compact on wide screens.
  - Load preview columns at the pinned version and gate the preview until columns match the current version.
  - Reset selection on asset switch (panel keyed by path) and fall back to the newest snapshot if a prior pick isn’t available.
  - Handle catalog‑only DuckLake paths (e.g. `ducklake://main`) without crashing; show only Partitions for table‑less nodes.
  - Escape DuckLake catalog names in client‑built DESCRIBE ATTACH strings to prevent breaking out of the string literal.

<sup>Written for commit 6992a6763032988846e8f95e898c436c4d95d4cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9709?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-common/src/query_builders.rs
+++ b/backend/windmill-common/src/query_builders.rs
@@ -220,11 +220,16 @@ struct CountPayload {
 }
 
 /// `WM_INTERNAL_DB_DUCKLAKE_SNAPSHOTS` payload — lists the time-travel history
-/// (catalog commit log) of a ducklake. Snapshots are catalog-wide in DuckLake,
-/// so this lists every commit; any `AT (VERSION => n)` read targets one of them.
+/// of a ducklake table. DuckLake snapshots are catalog-wide commits, so without
+/// a `table` this lists every commit; with one it is scoped to snapshots where
+/// that table exists (see `expand_ducklake_snapshots`).
 #[derive(Deserialize)]
 struct DucklakeSnapshotsPayload {
     ducklake: String,
+    /// Schema-qualified table name (e.g. `main.events_daily`) to scope the
+    /// history to. Snapshots predating the table's creation are excluded — a
+    /// time-travel read can't target a version where the table didn't exist.
+    table: Option<String>,
 }
 
 #[derive(Deserialize)]
@@ -397,9 +402,27 @@ fn expand_ducklake_snapshots(json_str: &str, db_type: DbType) -> Result<String, 
     let payload: DucklakeSnapshotsPayload = serde_json::from_str(json_str)
         .map_err(|e| format!("Invalid DUCKLAKE_SNAPSHOTS payload: {}", e))?;
     // `dl` is the alias `wrap_ducklake_query` attaches and `USE`s below.
-    let query =
-        "SELECT snapshot_id, snapshot_time FROM ducklake_snapshots('dl') ORDER BY snapshot_id DESC"
-            .to_string();
+    let query = match &payload.table {
+        // Scope to snapshots from the table's first creation onward. A DuckLake
+        // table created at snapshot N can't be read before N (the catalog-wide
+        // list would otherwise offer impossible versions). The creation snapshot
+        // is the earliest whose `changes.tables_created` names the table;
+        // COALESCE to 0 (show all) if it is never found.
+        Some(table) => {
+            let table = escape_sql_literal(table);
+            format!(
+                "SELECT snapshot_id, snapshot_time FROM ducklake_snapshots('dl') \
+                 WHERE snapshot_id >= COALESCE((\
+                   SELECT min(snapshot_id) FROM ducklake_snapshots('dl') \
+                   WHERE list_contains(changes.tables_created, '{table}')), 0) \
+                 ORDER BY snapshot_id DESC"
+            )
+        }
+        None => {
+            "SELECT snapshot_id, snapshot_time FROM ducklake_snapshots('dl') ORDER BY snapshot_id DESC"
+                .to_string()
+        }
+    };
     Ok(maybe_wrap_ducklake(query, Some(&payload.ducklake)))
 }
 
@@ -3304,6 +3327,17 @@ mod tests {
         assert!(result.contains("ATTACH 'ducklake://analytics' AS dl;USE dl;"));
         assert!(result.contains("ducklake_snapshots('dl')"));
         assert!(result.contains("ORDER BY snapshot_id DESC"));
+        // Unscoped: no per-table existence filter.
+        assert!(!result.contains("tables_created"));
+    }
+
+    #[test]
+    fn test_expand_ducklake_snapshots_scoped_to_table() {
+        let json = r#"{"ducklake": "analytics", "table": "main.events_daily"}"#;
+        let result = expand_ducklake_snapshots(json, DbType::Duckdb).unwrap();
+        // Scoped to snapshots from the table's first creation onward.
+        assert!(result.contains("list_contains(changes.tables_created, 'main.events_daily')"));
+        assert!(result.contains("snapshot_id >= COALESCE"));
     }
 
     #[test]
--- a/backend/windmill-common/src/query_builders.rs
+++ b/backend/windmill-common/src/query_builders.rs
@@ -172,6 +172,19 @@ pub struct SimpleColumn {
 pub struct SelectOptions {
     pub limit: Option<i64>,
     pub offset: Option<i64>,
+    /// DuckLake time-travel: when set (DuckDB only), the read is pinned to this
+    /// catalog snapshot via `AT (VERSION => n)`. Ignored for other db types.
+    pub version: Option<i64>,
+}
+
+/// DuckLake time-travel suffix appended after a table name in a FROM clause.
+/// `n` is a server-controlled `i64` (a snapshot id), so inlining it is
+/// injection-safe. Empty string when unpinned (reads the latest snapshot).
+fn duckdb_version_suffix(version: Option<i64>) -> String {
+    match version {
+        Some(v) => format!(" AT (VERSION => {})", v),
+        None => String::new(),
+    }
 }
 
 // ---------------------------------------------------------------------------
@@ -190,6 +203,8 @@ struct SelectPayload {
     #[serde(rename = "fixPgIntTypes")]
     fix_pg_int_types: Option<bool>,
     ducklake: Option<String>,
+    /// DuckLake snapshot to time-travel the read to (DuckDB only).
+    version: Option<i64>,
 }
 
 #[derive(Deserialize)]
@@ -200,6 +215,16 @@ struct CountPayload {
     #[serde(rename = "whereClause")]
     where_clause: Option<String>,
     ducklake: Option<String>,
+    /// DuckLake snapshot to time-travel the count to (DuckDB only).
+    version: Option<i64>,
+}
+
+/// `WM_INTERNAL_DB_DUCKLAKE_SNAPSHOTS` payload — lists the time-travel history
+/// (catalog commit log) of a ducklake. Snapshots are catalog-wide in DuckLake,
+/// so this lists every commit; any `AT (VERSION => n)` read targets one of them.
+#[derive(Deserialize)]
+struct DucklakeSnapshotsPayload {
+    ducklake: String,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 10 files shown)


rubenfiszel:
Addressed the two CI review findings (flagged by cubic) in 56e6602.

Both pointed at the same real issue: the copied/displayed `FROM` clause used the bare `tableKey` (e.g. `schema.table`), but a consumer script ATTACHes the ducklake under an alias, so an unqualified reference won't resolve there. Fixed by catalog-qualifying with the `lake.` prefix — matching the alias the duckdb scaffold ATTACHes under (`ATTACH 'ducklake://…' AS lake`) — in both `DucklakeSnapshotHistory.fromClause` (copy button + tooltip) and `DucklakeVersionPreview.exampleSql`. The live preview grid was already correct (it runs `USE dl` server-side, so its reads never needed the prefix); only the paste-able hints did.

rubenfiszel:
## ✅ Verified end-to-end with real DuckLake data

Stood up a full local env (Filesystem object store + a DuckLake catalog on a postgres resource), deployed a `// materialize ducklake://main/orders_daily` pipeline, and ran it several times to produce real snapshots (v0–v4). The clickable asset node opens the new tabbed panel:

**History tab** — snapshot list (newest-first) with timestamps and copy-`AT (VERSION)` / copy-`FROM` / Query actions:

History tab

**Query tab** — `FROM lake.orders_daily AT (VERSION => 4)` time-travel read rendered as a read-only grid (via the `version`-threaded SELECT marker):

Query tab

### Bug found & fixed during live testing
Real data surfaced a display bug not visible in unit tests: DuckLake serializes `snapshot_time` as **microseconds-since-epoch (a numeric string)**, so `new Date(snapshot_time)` rendered *Invalid Date*. Fixed in `3637a6f` by detecting the µs magnitude and converting to ms before formatting (you can see the correct timestamps in the History screenshot).

This supersedes the "Not captured" note in the PR description — the feature is now confirmed working in a real browser against a live DuckLake.

rubenfiszel:
### Also verified: partitioned table + multiple runs

Added a second, **partitioned** ducklake table (`// partitioned daily`) and materialized three partitions (2026-06-20/21/22), re-running 2026-06-20 once. The **Partitions** tab shows one row per partition, each with its own snapshot id — note 2026-06-20 carries the newer snapshot (14, later timestamp) from the re-run, demonstrating idempotent re-materialization. The producer node shows a ✓×4 run badge and the runs panel lists each execution with its `partition` arg.

Partitions tab — 3 partitions

rubenfiszel:
### UX refinement: History + Query merged into one master-detail tab

Per feedback, the separate History and Query tabs are now a single **History** tab with a left/right split: the snapshot list on the left selects the version previewed in the read-only grid on the right (newest auto-selected on open). The copy-able `AT (VERSION => n)` clause moved to the preview's SQL line. Selecting an older snapshot time-travels the grid — e.g. an early snapshot of the partitioned table shows 3 rows (one partition) vs 9 at the latest (three partitions).

History master-detail

rubenfiszel:
Addressed the CI review findings in `eb04496` (verified live against a table with real schema drift):

**[P1 codex/pi] Historical preview used current-schema metadata.** `DucklakeVersionPreview` loaded the *live* columns via `loadAllTablesMetaData`, so an `AT (VERSION => n)` read enumerating a column added in a later snapshot failed ("Referenced column X not found"). It now `DESCRIBE`-loads the column set **at the pinned version**, and the read is gated on the loaded columns matching the currently-shown version (avoids a stale-colDefs race when switching versions). Verified: previewing v17 of a table shows the new `amount` column; previewing an older v9 shows only `{id, name}` with no error (it threw before).

**[P2 cubic] selectedVersion sticky across assets.** The panel is now keyed on the asset path (remounts per asset, resetting the selected snapshot + tab), and `effectiveVersion` falls back to the newest snapshot when the pick isn't present in the current table's list.

Also removed the now-incorrect "column metadata is the same across snapshots" comment that the Pi review flagged.

CC follow-up: the earlier 'redundant metadata refetch on version switch' nit is now intentional — columns *must* be (re)loaded per version for correctness, since a snapshot's schema is the source of truth.

rubenfiszel:
`b83a02`: fixed the last two review items — updated `docs/ducklake-materialization.md` and the `DucklakeSnapshotHistory` comment to match the shipped master-detail UI (the only copy affordance is the full `FROM lake.<table> AT (VERSION => n)` clause in the preview; the standalone per-snapshot copy buttons were removed in the merge). The false 'column metadata is the same across snapshots' comment was already removed by the version-aware-columns rewrite in `eb04496`.

All findings from the review rounds are now addressed.

rubenfiszel:
Good catch — fixed in `5b6f87c` (verified live).

`parseDbInputFromAssetSyntax('ducklake://main')` did `undefined.split('.')` and threw on a catalog-only path (no table segment) — and these are real nodes: a consumer reading the whole catalog (`// on ducklake://main`) creates a `kind=ducklake, path=main` asset. Before my panel change the ducklake branch rendered `PartitionStatusGrid` (which doesn't parse), so I introduced this regression.

Fix:
- `parseDbInputFromAssetSyntax` no longer throws on a table-less path — returns `specificTable: undefined` (root cause; also protects the other callers: `FavoriteMenu`, `DataTablePreview`, `assets/+page`). Added unit tests (table, schema.table, catalog-only).
- `DucklakeAssetPanel` renders only the partition grid for table-less nodes (snapshot history / time-travel are per-table).

Verified: deployed a `// on ducklake://main` consumer → clicking the catalog-only `DUCKLAKE main` node now shows the Materialized-partitions grid with no tabs and no error (it threw before).

rubenfiszel:
Fixed in `6992a67`.

`fetchDucklakeColumnsAtVersion` now escapes the catalog name (`ducklake.replace(/'/g, "''")`) before interpolating it into the `ATTACH 'ducklake://…'` string literal — matching the backend's `escape_sql_literal` (single-quote doubling). I went with escaping rather than a new server-side marker because it directly closes the string-literal boundary and is consistent with the existing client-built `DUCKLAKE_GET_SCHEMA_QUERY` in the same file. (The sibling `fetchDucklakeSnapshots` was already safe — it routes through the `DUCKLAKE_SNAPSHOTS` marker, which escapes the catalog name server-side via `maybe_wrap_ducklake`.)

Also fixed the v1.x docs checklist line (`ducklake-materialization.md:262`) that still described the standalone `AT (VERSION => n)` copy — it now matches the shipped full-`FROM` copy affordance, same as the body paragraph.

PR #9711, opened 2026-06-22
fix(oauth): restore bring-your-own CC token URL override

hugocasa:
EE companion: windmill-labs/windmill-ee-private#625

## Summary
Restores the resource-level token URL override for OAuth client-credentials connections. A user connecting a client-credentials resource with their own client ID/secret can again point the token exchange at a custom endpoint; the value is stored on the connection and reused on token refresh. The override is available only on the bring-your-own-credentials path and is mutually exclusive with the instance-name field used by host-templated providers.

## Changes
- Re-add the optional "Token URL override" field to the client-credentials connect form (bring-your-own credentials, non-instance-templated providers).
- Backend (EE, `oauth2_ee.rs`): accept `cc_token_url` on the connect and create-account requests, honored only with caller-supplied `cc_client_id`/`cc_client_secret`; validated as an http(s) URL; rejected on the shared-instance path.
- Document `cc_token_url` on both request bodies in `openapi.yaml`.
- Bump the EE ref.

## Test plan
- [ ] Connect a client-credentials resource with your own client ID/secret + a custom token URL; confirm the exchange uses it and an account row is created with the stored URL.
- [ ] Confirm the override is rejected on the shared-instance-credentials path and when combined with an instance name.
- [ ] Confirm the stored token URL is reused on refresh.

## Screenshots
cc-token-url-override


hugocasa:
Addressed the Codex P1 (host-pinned CC providers reachable via API).

The frontend already hid the override for `{instance}`-templated providers, but the backend accepted a bring-your-own `cc_token_url` without `cc_instance`, which could redirect the host for `coupa`/`servicenow`/`snowflake`/etc. Added server-side enforcement:

- New `windmill_oauth::is_instance_templated_cc(...)` detects registry providers whose token URL is `{instance}`-templated (mirrors the template logic in `resolve_cc_token_url_input`).
- Both `connect_client_credentials` and `create_account` now reject a caller `cc_token_url` for those providers (instance-name path only). Fixed-host registry providers and custom providers still accept the override.
- Unit tests added for the detection (`instance_templated_cc_true_for_templated_providers`, `..._false_for_fixed_host_and_unknown`).

EE companion commit: windmill-labs/windmill-ee-private@20ec193 (PR #625). `cargo check --features enterprise,private` and the oauth unit tests pass.

On the integration-test ask: the request-matrix cases (accept on non-templated, reject on shared path, reject `cc_instance`+`cc_token_url`, reject for templated) currently have no DB-backed harness for these handlers; I verified them manually against a running backend earlier (rejections + DB storage). Left as a follow-up consistent with the existing untested state of these handlers.

hugocasa:
Addressed both findings from the latest Codex review.

**[P1] `cc_token_url` bypassed the `client_credentials` grant check** — fixed. The override path now runs `resolve_cc_token_url(..., None, ...)` purely for its provider/grant gate (registry providers must declare `client_credentials`; custom providers need an instance entry that enables it) and then uses the caller's override URL instead of the resolved one. This guarantees parity with the no-override path, so an auth-code-only registry provider (e.g. `github`) is rejected. Applied in both `connect_client_credentials` and `create_account`. EE: windmill-labs/windmill-ee-private@a939228.

**[P2] served OpenAPI omitted `cc_token_url`** — fixed. Added `cc_token_url` to both request bodies in the dereferenced artifacts (`openapi-deref.yaml` / `openapi-deref.json`) that back `/openapi.yaml` and `/openapi.json`. Kept it to a focused manual add rather than a full `build_openapi.sh` regen, since the committed derefs are already stale vs `openapi.yaml` on main and a regen would pull in ~1k lines of unrelated drift; both files validate (`jq` / yaml parse).

`cargo check --features enterprise,private` passes; the `is_instance_templated_cc` unit tests still pass.

On test coverage: the grant gate now reuses `resolve_cc_token_url`, whose templating/registry logic is covered by the `resolve_cc_token_url_input` unit tests. The full handler request-matrix still lacks DB-backed integration tests (no harness exists for these handlers today) — left as a follow-up, verified manually against a running backend.

PR #9738, opened 2026-06-23
feat: render mermaid diagrams in chat code blocks

rubenfiszel:
## Summary

Fixes WIN-2091.

When an AI chat response (Flow-as-Chat or Copilot chat) contained a fenced ```mermaid code block, it was displayed as raw text with TypeScript syntax highlighting instead of being rendered as a diagram. `CodeDisplay.svelte` had no handling for the `mermaid` language, and the `mermaid` library was not a dependency.

This adds a dedicated `MermaidDisplay` component that lazy-loads `mermaid` and renders the block as an SVG diagram, with a graceful fallback to the raw source when the diagram can't be parsed.

## Changes

- `frontend/package.json`: add `mermaid` (^11.15.0) as a dependency.
- `frontend/src/lib/components/copilot/chat/script/MermaidDisplay.svelte` (new):
  - Accepts a `code: string` prop.
  - Lazy-imports `mermaid` via dynamic `import('mermaid')` so it stays out of the main bundle.
  - Renders to SVG with `mermaid.render()` and inserts it via `{@html}`. Uses `securityLevel: 'strict'` (mermaid runs DOMPurify on the output and strips JS-binding directives) and `suppressErrorRendering: true` (parse errors throw instead of injecting an orphan error diagram into the DOM).
  - Uses a unique element id per instance/render via `randomUUID` from `$lib/utils/uuid`.
  - Reacts to dark mode via `useIsDarkMode()` (re-renders with the `dark`/`default` theme).
  - Falls back to the raw source in a `<pre>` when nothing has rendered yet; guards against out-of-order async (a stale render can't overwrite a newer one) and keeps the last good diagram through transient parse failures while a block is still streaming.
- `frontend/src/lib/components/copilot/chat/script/CodeDisplay.svelte`: render `<MermaidDisplay {code} />` when `language === 'mermaid'`, otherwise the existing `<HighlightCode>`.

## Screenshots

mermaid-rendering

Dark mode (theme adapts):

mermaid-dark

## Test plan

- [x] Valid `mermaid` block renders as a centered SVG diagram inside the bordered container.
- [x] Invalid `mermaid` block falls back to the raw source in a `<pre>` (no thrown error, no orphan error diagram injected into the DOM).
- [x] Toggling dark mode re-themes the diagram (default vs dark) without leaving a stale SVG.
- [ ] In an actual AI chat, prompt the assistant to return a ```mermaid block and confirm it renders.

---
## Summary by cubic
Render `mermaid` code blocks in chat as SVG diagrams instead of plain text, and only display diagrams while they match the current source. Fixes WIN-2091.

- **New Features**
  - Added `MermaidDisplay.svelte` to lazy-load `mermaid` and render SVG with `securityLevel: 'strict'`.
  - `CodeDisplay.svelte` now routes `language === 'mermaid'` blocks to `MermaidDisplay`.
  - Supports dark mode, uses unique per-render ids, guards out-of-order async, only shows SVG when it matches the current code, and falls back to raw source on parse errors.

- **Dependencies**
  - Added `mermaid` (^11.15.0).

<sup>Written for commit 3ff04a5a96210c2b1cd79a0611719cc521f93802. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9738?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -52,6 +52,7 @@
 				"lru-cache": "^11.1.0",
 				"lucide-svelte": "^0.540.0",
 				"mdast-util-find-and-replace": "^3.0.2",
+				"mermaid": "^11.15.0",
 				"minimatch": "^10.0.1",
 				"monaco-editor": "npm:@codingame/monaco-vscode-editor-api@=25.0.0",
 				"monaco-languageclient": "10.6.0",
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Addressed the Codex P1 (stale diagram after a later parse failure) in 3ff04a5.

Previously the component kept the last good SVG through parse failures (to avoid streaming flicker), which meant a successfully-rendered diagram stayed visible even after the source changed to something that no longer parses — a misleading, mismatched diagram.

Fix: the rendered SVG is now tied to the exact source that produced it (`renderedCode`), and it's only displayed while `renderedCode === code`. A later edit/append that fails to parse immediately falls back to the raw source instead of leaving a stale diagram. Dark-mode toggles stay smooth (same `code`, so the diagram isn't dropped while the re-themed render is in flight). The out-of-order `renderSeq` guard is kept.

Also dropped the now-dead outer `id` uuid (Claude's nit) — `mermaid.render()` already gets a fresh `mermaid-${randomUUID()}` per attempt — and removed the per-failure `console.error` that could spam during streaming.

Verified locally: valid source renders, breaking it falls back to raw text (no stale diagram), and re-fixing it re-renders.

PR #9763, opened 2026-06-24
feat(ai-chat): context usage gauge + unified model settings menu

Guilhem-lm:
## Summary
Reworks the AI chat input toolbar. The text "Context usage: ~Xk / Yk" indicator becomes a compact **gauge** (with a hover tooltip) sitting left of the model picker, and the previously separate **model picker / reasoning-effort / prompt-settings** controls are merged into a **single dropdown** with sections. The trigger shows `model · effort`.

## Changes
- **Context usage gauge** (`ContextUsageIndicator.svelte`): thin bar that fills toward the compaction threshold, colored with the brand accent (amber/red as it nears/hits the 0.8 compaction trigger). Details in a hover tooltip. Marked up as `role="meter"`.
- **Unified model settings dropdown** (new `AIChatModelSettings.svelte`, replaces `ProviderModelSelector` + `AIChatSettingsMenu` in the chat): one trigger labelled `model · effort`, opening a menu with **Parameters** (top), **Model** list, and **Thinking** sections.
  - **Thinking** is a lean slider (thin track, accent fill up to the thumb). Model rows and the slider are registered as melt items, so they share roving highlight + arrow-key navigation; ←/→ adjust the effort while the slider row is focused, and only the slider area (not the header) shows the highlight.
  - **Parameters** opens as a melt submenu (hover-opens, floating-positioned so it flips on screen edges) with User prompt / Workspace prompt / AI settings.
  - Picking a model keeps the menu open (`closeOnItemClick=false`, new optional prop on `DropdownV2`); the trigger width is frozen while open so the `bottom-end` popover doesn't drift as the label resizes.
  - When a model has no reasoning support, the Thinking section stays but is shown disabled with a reason (instead of disappearing).
- `DropdownV2.svelte`: expose `builders` to the `customMenu` snippet and add the `closeOnItemClick` prop. New `meltComponents/MenuItemWrapper.svelte` registers arbitrary content (the slider row) as a melt item.
- Hide the `@` context picker in **global** chat mode.

## Screenshots
Captured in the full-page session chat.

### Before
before

### After — gauge + tooltip
gauge

### After — unified dropdown (Parameters / Model / Thinking)
dropdown

### After — Parameters submenu (hover, floating-positioned)
submenu

## Test plan
- [ ] Open the AI chat, send a message: the gauge appears left of the model button; hover shows `Context usage … ~Xk / Yk (Z%)`.
- [ ] Open the model settings dropdown: Parameters at top, Model list (current model checked), Thinking slider.
- [ ] Drag the thinking slider — it works and the popover doesn't drift; ←/→ adjust effort when the slider row is focused.
- [ ] Pick a model — the menu stays open; the button shows `model · effort`.
- [ ] Hover Parameters — submenu opens with User/Workspace prompt + AI settings; clicking one opens the modal and closes the menu.
- [ ] Switch to a non-reasoning model: Thinking shows disabled with "Not supported by this model".
- [ ] In Global mode, the `@` context button is absent.


Guilhem-lm:
Thanks for the auto-review (both verdicts: good to merge). Addressed the two non-blocking notes:

- **`closeOnItemClick` read at mount** — added a comment clarifying it's captured once at menu creation (like `placement`), so it's intentionally not live-toggleable. (`63ddf85`)
- **`MenuItemWrapper` keyboard activation** — left as-is: the only consumer passes its own `onKeydown`, and melt's `$item` action supplies key handling; a generic keydown→click bridge would be speculative for a single-use wrapper. Happy to add it if we reuse the component for a click-activated row.

Guilhem-lm:
Addressed the Pi P2 (`63ddf85`… now `638f84e16e`): updated the `showContextPicker` comment so `GLOBAL` is listed under the modes that **don't** surface the `@` picker, matching the code.

PR #9766, opened 2026-06-24
fix(frontend): stop flow step id generation from being poisoned by non-canonical keys

rubenfiszel:
## Summary
New flow steps could suddenly start getting garbage ids (e.g. jumping from `a, b, c` to `bzw, bzx`). `nextId()` computes the next step id from the **max** of `charsToNumber(key)` over every module id and `flowState` key. Only canonical auto-ids (`a, b, … z, aa, ab, …`) have a meaningful `charsToNumber` value, but `flowState` also holds non-canonical keys:

- copy ids from duplicating a step — `"z2"`, `"c10"`
- subflow result keys written while interacting with the flow test drawer — `"subflow:…"`
- reserved keys always present — `"failure"`, `"preprocessor"`
- user-renamed ids

The old guard `if (key.length >= 4) return acc` only filtered *long* junk and let *short* junk through. So duplicating step `z` creates key `"z2"` (`charsToNumber("z2") = 629`), making the next new step jump to `"xg"` and escalate from there. The test drawer made this more likely because it writes extra keys into the shared `flowState`.

Reproduced deterministically: after 26 steps the next id should be `aa`, but after duplicating `z` it became `xg`.

## Changes
- `frontend/src/lib/components/flows/flowModuleNextId.ts`: `nextId` now only counts a key toward the max if it is a genuine auto-generated id — it must round-trip (`numberToChars(charsToNumber(key)) === key`) and not be reserved. Removes the broken `length >= 4` cap, so flows with 1000+ real steps still get correct sequential ids.
- `frontend/src/lib/components/flows/flowModuleNextId.test.ts`: 5 regression tests covering fresh-flow sequencing, reserved-key exclusion, the copy-id poisoning case, subflow/renamed-key poisoning, and long auto ids.

## Test plan
- [x] `npx vitest run src/lib/components/flows/flowModuleNextId.test.ts` — 5/5 pass
- [ ] In a flow that already contains copied steps (ids like `z2`/`c2`) or steps whose test results created `subflow:`/renamed keys in flowState, add a new step via the graph "+" and confirm it gets the next sequential id (e.g. `d`) rather than a garbage jump (e.g. `bzw`).

## Screenshots
No meaningful visual diff: this is pure id-generation logic. The only visible effect (correct sequential step ids) manifests only from an already-corrupted flowState that is hard to reproduce on demand, so the behavior is covered by the unit tests above instead.

---
## Summary by cubic
Fix flow step id generation by making `nextId` ignore non-canonical `flowState` keys and long renamed ids, so new steps get correct sequential ids. Prevents jumps to garbage ids after duplications or test drawer interactions.

- **Bug Fixes**
  - Only count keys that round-trip (`numberToChars(charsToNumber(key)) === key`), are not reserved (`failure`, `preprocessor`), and are shorter than 4 chars; excludes copy ids (`z2`, `c10`), `subflow:` results, and long renames.
  - Add regression tests for fresh sequencing, reserved-key exclusion, copy-id poisoning, subflow keys, and long lowercase/underscored renames.

<sup>Written for commit 5c261afc43d1ee876b44d2b7f3763b99e94d1d7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9766?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/flows/flowModuleNextId.test.ts
+++ b/frontend/src/lib/components/flows/flowModuleNextId.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from 'vitest'
+
+import type { OpenFlow } from '$lib/gen'
+import type { FlowState } from './flowState'
+import { nextId } from './flowModuleNextId'
+
+function flowWith(ids: string[]): OpenFlow {
+	return {
+		summary: '',
+		value: {
+			modules: ids.map((id) => ({ id, value: { type: 'identity' } as any }))
+		}
+	} as OpenFlow
+}
+
+function stateWith(keys: string[]): FlowState {
+	return Object.fromEntries(keys.map((k) => [k, {}])) as FlowState
+}
+
+describe('nextId', () => {
+	it('produces a, b, c, ... for a fresh flow', () => {
+		expect(nextId(stateWith(['failure']), flowWith([]))).toBe('a')
+		expect(nextId(stateWith(['a', 'failure']), flowWith(['a']))).toBe('b')
+		expect(nextId(stateWith(['a', 'b', 'c', 'failure']), flowWith(['a', 'b', 'c']))).toBe('d')
+	})
+
+	it('ignores the reserved failure/preprocessor keys always present in flowState', () => {
+		expect(nextId(stateWith(['failure', 'preprocessor']), flowWith([]))).toBe('a')
+	})
+
+	// Regression: copy ids ("z2"), subflow result keys and other non-canonical keys land in
+	// flowState; charsToNumber on them used to leak into the max and made new steps jump to
+	// garbage ids like "bzw".
+	it('is not poisoned by copy ids', () => {
+		const ids = ['a', 'b', 'c']
+		const state = stateWith([...ids, 'c2', 'a2', 'z2', 'c10', 'failure'])
+		expect(nextId(state, flowWith(ids))).toBe('d')
+	})
+
+	it('is not poisoned by subflow result keys or user-renamed ids', () => {
+		const ids = ['a', 'b']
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Addressed the review feedback (thanks @claude for catching this).

**Issue**: removing the `length >= 4` cap meant all-lowercase renamed step ids (e.g. `process`) — which round-trip cleanly through `numberToChars` — would feed into the max again and poison id generation, the same symptom in a different guise. Since step ids can be renamed to `^[a-zA-Z][a-zA-Z0-9_]*$`, this is a realistic case, and the prior code (which skipped all `length >= 4` keys) did not have it.

**Fix** (5c261af): restored the length cap and paired it with the round-trip canonical check. Now:
- short non-canonical keys (copy ids `z2`/`c10`, reserved/underscored short ids) are rejected by the round-trip check — fixes the original reported bug;
- long lowercase renames (`process`, `validate`) stay out of the sequence via the length cap — no regression;
- `nextId` keeps its max+1 semantics (no gap-filling), so deleting a step never silently re-wires a dangling `results.X` reference to a freshly created step.

Updated the tests accordingly: dropped the inaccurate "long auto ids beyond length 3" case and added an explicit "renames to long lowercase words / underscored ids" regression test. All 5 pass locally.

PR #9769, opened 2026-06-24
feat(ai-chat): add /clear session command to start a fresh conversation

centdix:
## Summary

Adds a built-in `/clear` slash command to the AI chat in sessions, mirroring the existing `/compact` command. Typing `/clear` saves the current conversation to history and resets to a fresh chat (equivalent to the "New chat" action), instead of sending a turn to the model.

## Changes

- Add `CLEAR_COMMAND_NAME` / `CLEAR_COMMAND_RE` (`^\/clear\s*$`, anchored so a mid-sentence mention of `/clear` still sends normally).
- Register `/clear` in `sessionBuiltinCommands` so it surfaces in the slash-command picker (via the existing `sessionCommands` `$derived`, which already de-dupes colliding workspace skills).
- Intercept `/clear` in the `sendRequest` submit path under the same `isSessionChat && mode === GLOBAL` guard as `/compact`; it clears the composer and delegates to the pre-existing `saveAndClear()` (which handles cancellation, queued-message drop, history persistence, and conversation reset), then returns without sending a turn.
- Refactor the single-command interception into a shared guard that branches on `/compact` vs `/clear`.
- Add tests: `/clear` in session GLOBAL mode resets the conversation and never reaches the model loop; `/clear` outside session chat falls through as a normal message.

## Screenshots

Slash-command picker showing `/clear` alongside `/compact`:

01-picker

After submitting `/clear` — conversation reset to a fresh chat (no turn sent to the model):

03-after-clear

## Test plan

- [ ] In the session AI chat (GLOBAL mode), type `/` and confirm `/clear` appears in the picker with description "Clear the conversation and start a new chat".
- [ ] Hold a short conversation, then submit `/clear`; confirm the messages clear back to the empty state and nothing is sent to the model.
- [ ] Confirm the cleared conversation is still recoverable from chat history (non-destructive).
- [ ] Confirm a message that merely mentions `/clear` mid-sentence still sends normally.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -118,6 +118,10 @@ const USER_CANCEL_REASON = 'user_cancelled'
 // regular message that merely mentions "/compact" mid-sentence is unaffected.
 const COMPACT_COMMAND_NAME = 'compact'
 const COMPACT_COMMAND_RE = /^\/compact\s*$/
+// Built-in `/clear` session command — saves the conversation to history and
+// resets to a fresh chat (the "New chat" action), instead of sending a turn.
+const CLEAR_COMMAND_NAME = 'clear'
+const CLEAR_COMMAND_RE = /^\/clear\s*$/
 const AI_AUTONOMY_MODE_STORAGE_KEY = 'ai-chat-autonomy-mode'
 const LEGACY_AUTO_ACCEPT_TOOL_CONFIRMATIONS_STORAGE_KEY = 'ai-chat-yolo-mode'
 const WEB_SEARCH_ERROR_HINT =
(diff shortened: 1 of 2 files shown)


centdix:
Addressed both findings in 06cac5ef43.

**[1] Queued `/clear` re-firing into the next chat** — fixed at the root cause, which also covered the same latent bug in `/compact`. `sendRequest` was returning `false` both for *input dropped without being acted on* (mode hidden, empty, `beforeSend` failed → flush correctly restores it) and for *input consumed as a built-in command* (must NOT be restored). The queue-flush epilogues restore `queuedMessage` whenever `accepted === false`, so a flushed `/clear` was re-queued — and worse, the restore clobbered the `queuedMessage = ''` that `saveAndClear` had just set, leaving a stale `/clear` that wiped the next conversation.

Fix: the command-interception path now returns `true` (consumed), and the `sendRequest` return contract is reframed accordingly — `false` means "dropped without being acted on" (restore it), `true` means "consumed: sent as a turn **or** handled as a local command". The return value is read only by the two flush epilogues; every other caller ignores it, so this has no collateral effect.

**[2] Stale `sessionCommands` assertion** — updated to `['compact', 'clear', 'review-code']`.

**Test coverage** — added a regression test (`consumes a /clear flushed from the queue without re-queuing it`): it seeds an exchange, queues `/clear`, runs a normal turn to clean completion so the flush fires, then asserts `queuedMessage === ''` and the conversation was reset. It fails before the fix (the flush re-queues `/clear`) and passes after. The two routing-test assertions were updated to `toBe(true)` to match the consumed contract.

Note: the queued-while-streaming case is pinned by the unit test rather than a live repro — reproducing it in the browser depends on typing during a streaming response, which is timing-sensitive to drive reliably. The original `/clear` happy path (picker entry + reset-on-submit) was verified live and is in the PR description.

PR #9772, opened 2026-06-24
feat: self-host docs search for chat, mcp, cli; drop inkeep

centdix:
## Summary

Windmill now **self-hosts its documentation corpus** instead of fetching `windmill.dev` from the browser. A vendored, gzipped snapshot of `llms.txt`/`llms-full.txt` is embedded in the backend, decompressed+parsed once into a cached corpus, and served by two read-only endpoints. The AI chat, the MCP doc tools, and the `wmill docs` CLI all consume the same endpoints — so docs search works with **no runtime egress** (including air-gapped instances) and is **no longer EE-gated**. The inkeep proxy is removed.

Previously there were three divergent paths: the AI chat fetched `windmill.dev` cross-origin (breaks offline, CORS-fragile, leaks queries); the CLI and MCP used the EE-only inkeep RAG proxy. This unifies them onto one backend-hosted path.

## Changes

- **Vendored snapshot** — `backend/windmill-api/docs_snapshot/`: `llms.txt.gz` + `llms-full.txt.gz` (~625 KB) embedded via `include_bytes!`, plus `fetch.sh` to refresh and a `README.md`. The build is unchanged — the snapshot is checked in and copied wholesale by the existing `COPY ./backend ./`, so it embeds into the binary at compile time with no Dockerfile/runtime path to provision.
- **Backend module** — `backend/windmill-api/src/docs/`: `corpus.rs` (gzip decompress + parse, cached in a `OnceLock`), `search.rs` (ranking/snippet/page-render logic ported from the former frontend TS, with parity unit tests), `mod.rs` exposing:
  - `GET /api/docs/search?query=...` — full-text + index search; returns each hit's title, `Source` URL, and matching snippets.
  - `GET /api/docs/page?url=...&section=...` — reads one page (or a section) by its `Source` URL. The param is named `url` to match exactly what search returns; a bare `/docs/...` path is also accepted and canonicalized. Large pages return their section outline so the model can fetch one section at a time.

  CPU-bound work (decompress, per-query scan, render) runs under `spawn_blocking`; query terms are capped.
- **Scope** — added a `Docs` read-only `ScopeDomain` (`scopes.rs`) so the GET routes map to `docs:read`. (Inkeep had no scope, so the old MCP doc tool failed closed for scope-restricted tokens — this fixes that.)
- **OpenAPI + codegen** — removed `/inkeep`, added `/docs/search` + `/docs/page` (`x-mcp-tool`); regenerated `auto_generated_endpoints.rs`, `mcpEndpointTools.ts`, the frontend client, and `openapi-deref.{yaml,json}`. MCP `queryDocumentation` → `searchDocs` + `readDocsPage`.
- **Frontend** — `copilot/chat/docs/core.ts` is now a thin wrapper over the generated `DocumentationService`; the windmill.dev fetch + ranking logic (and its now-orphaned test) are deleted (logic moved to Rust).
- **CLI** — `wmill docs` repointed from the removed `/api/inkeep` to `/api/docs/search`; behavior changes from an inkeep-synthesized answer to snippet/link search, now CE-available. (CLI uses search only; it does not read full pages.)
- **Inkeep removed** — `inkeep_oss.rs` + the EE `inkeep_ee.rs` (companion PR below); `WINDMILL_CUSTOMER_SERVICE_BASE_URL` is left intact (still used by other EE modules).
- **Docs-freshness CI** — `.github/workflows/refresh-docs-snapshot.yml`: weekly (and `workflow_dispatch`) job that re-runs `fetch.sh`, sanity-checks the result (size + `Source:` page-count floor, to reject a valid-but-garbage 200), and opens a PR via the internal app when the snapshot changed. Keeps the embedded docs fresh independently of the release cadence — a merged refresh rides into the next release build automatically.

> **Frontend note:** no visible UI effect. The AI chat's `search_docs`/`read_docs_page` tools look and behave identically; only their data source moved from `windmill.dev` to the backend. (Per the screenshots rule: skipped because there is no visible UI change.)

> **EE companion:** windmill-labs/windmill-ee-private deletes `inkeep_ee.rs`; `ee-repo-ref.txt` is bumped to that commit. Link added below.

## Reviewer notes

- **New third-party action:** the docs-freshness workflow uses `peter-evans/create-pull-request@v6` (the existing commit-back uses `stefanzweifel/git-auto-commit`, which pushes a branch rather than opening a PR). Tag-pinned to match the repo's pinning style — happy to SHA-pin it or swap to a `gh pr create` script if you'd rather not add the dependency.
- **App permission:** the workflow needs the internal GitHub App to have `pull-requests: write` (it already pushes commits via that app, so contents-write is covered). If PR-write isn't granted yet, the first run will 403 on PR creation.

## Test plan

- [x] Backend: `cargo test -p windmill-api docs::` — 23/23 parity tests pass (ported from the deleted `core.test.ts`); embedded-corpus parse test included. Re-run after the `url` rename — still green.
- [x] Backend compiles on `--features quickjs` and `--features quickjs,mcp`.
- [x] Live endpoints: search ranking, page + section extraction, link sanitization (`.mdx` source paths rewritten to canonical URLs), not-found, **401 unauth**, and **scope gating** (`docs:read` → 200, `jobs:read` → 403).
- [x] CLI verified end-to-end against a running backend (pretty output, `--json`, no-results).
- [x] Frontend: clean HMR; the exact browser → dev-proxy → `/api/docs/search` path verified with real session-cookie auth.
- [x] `refresh-docs-snapshot.yml` YAML validated; the embedded snapshot is reproducible (`gzip -9 -n`) so a no-op run produces no diff.
- [x] Reviewer: exercise the MCP `searchDocs`/`readDocsPage` tools with a scope-restricted MCP token (the one path with no automated coverage).
- [ ] EE build is validated by CI against the bumped `ee-repo-ref.txt` (couldn't run locally — the local EE worktree is stale/diverged, unrelated to this change).


centdix:
EE companion PR: windmill-labs/windmill-ee-private#630 (deletes `inkeep_ee.rs`; `ee-repo-ref.txt` is bumped to that commit in this PR).

PR #9780, opened 2026-06-25
fix(debounce): never supersede a running debounce survivor

rubenfiszel:
## Summary

With `debounce_args_to_accumulate` + a `concurrent_limit`, a message arriving while its debounce survivor is **already running** was marked completed/skipped (`Debounced Running by <id>`) and the running survivor was **deleted from the queue** — silently dropping its accumulated work, while the late arrival could not merge into it. A slow step + `concurrent_limit: 1` keeps a survivor in the running state for a long window, so anything arriving during it was lost (no error anywhere).

Root cause: a debounce survivor that has already been pulled and is executing has committed its batch and can no longer accumulate later arrivals, but `upsert_debounce_key` still chained the new arrival to it and the callers then completed the running job via `complete_debounced_job`.

Fix (EE, `windmill-queue/src/jobs_ee.rs` → `upsert_debounce_key`): detect, in a single race-free statement, whether the current debounce-key holder is running (`v2_job_queue.running = true`). If so, the late arrival starts a **fresh debounce window** — `previous_job_id` is left `NULL` (so no running job is ever completed), `debounced_times`/`first_started_at` reset, and the `NULL` previous yields a fresh batch via the existing `COALESCE`. The running survivor is left to finish normally. Behavior is **unchanged** when the holder is not running.

Not specific to `concurrent_limit` (it only widens the window) or to argument accumulation (delay-only debounce also wrongly killed a running survivor). The fix is in the function shared by all three debounce entry points (`maybe_debounce`, `maybe_debounce_post_preprocessing`, `maybe_debounce_flow_node`).

EE companion PR: windmill-labs/windmill-ee-private#631

## Changes
- EE: `upsert_debounce_key` adds a `prev_running` CTE and only supersedes a non-running holder; a running holder triggers a fresh window (both `needs_batch` branches).
- Tests: 6 new regression tests in `backend/windmill-queue/tests/debounce_test.rs` (running survivor not superseded, queued survivor still accumulates, no-accumulation path, committed-running survivor, flow post-preprocessing path, max-count window reset). The core test fails against `main` and passes with the fix.
- Refreshed the SQLx offline cache for the two changed `upsert_debounce_key` queries.
- Bumped `backend/ee-repo-ref.txt` to the EE commit.

## Test plan
- [ ] `cargo test -p windmill-queue --test debounce_test --features private,enterprise` — all 56 pass (52 pre-existing + 6 new); 2 stress tests ignored.
- [ ] Confirm `test_debounce_running_survivor_not_superseded` fails on `main` (reproduces the bug) and passes on this branch.
- [ ] `SQLX_OFFLINE=true cargo check -p windmill-queue --features private,enterprise` compiles with the updated cache.

---
## Summary by cubic
Prevents a running debounce survivor from ever being superseded and makes accumulation exactly-once. Late arrivals start a new window; survivors atomically claim the whole batch and persist accumulated args in one transaction; GC drops only consumed, non-queued rows after 10 minutes.

- **Bug Fixes**
  - EE: `upsert_debounce_key` checks `v2_job_queue.running`; a running holder is never superseded. Late arrivals reset `previous_job_id`, `debounced_times`, and `first_started_at`.
  - Per-key advisory lock: serialize upserts and batch chaining so concurrent late arrivals share one fresh window and batch.
  - Atomic upsert: single `INSERT ... ON CONFLICT` so a chaining push can’t fail if the pull path deletes the `debounce_key`.
  - Whole-batch claim in one `UPDATE` (no deadlock when two survivors race on the same batch).
  - Logging: emit the accumulation log after the claim transaction commits.
  - Applies to `maybe_debounce`, `maybe_debounce_post_preprocessing`, and `maybe_debounce_flow_node`.
  - Syncs `backend/ee-repo-ref.txt` to the latest EE companion.

- **New Features**
  - Exactly-once batch consumption: a survivor claims its row and any unclaimed siblings, accumulates exactly what it claimed, and persists in one transaction; a second survivor runs empty; a re-pull keeps its prior claim; CE/legacy keep own args; non-accumulate paths hard-delete rows.
  - Migration: add `consumed_at`/`consumed_by` to `v2_job_debounce_batch`; monitor GC removes consumed rows after 10 minutes only when the job is no longer in `v2_job_queue`.
  - Tests: coverage for running-survivor guard (incl. concurrency), the key-deletion race, partitioned claims and scalar args, GC-then-repull, flow-node guard, non-accumulation path, deadlock case, and a real-worker end-to-end accumulation test.

<sup>Written for commit be475a79478c6282e5f22a5ccb601b951ca72f7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9780?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the Codex P1 (concurrent late arrivals racing after a survivor starts running).

**Root of the finding:** the running check was a `prev_running` CTE evaluated from the statement snapshot — taken *before* the `ON CONFLICT` row lock. Two pushes racing after the survivor started running could both read "holder running" from their snapshots; the second would block on the row lock but still apply its stale value, resetting to a fresh window instead of debouncing the first — two independent windows.

**Fix:** replaced the CTE with a correlated `EXISTS (SELECT 1 FROM v2_job_queue q WHERE q.id = debounce_key.job_id AND q.running)` evaluated inside the `DO UPDATE SET`. `debounce_key.job_id` is the row being updated, so after the row lock the second push re-resolves it to the *first* arrival's job (the fresh-window head — queued, not running) and correctly debounces into that window. The row lock serializes the two upserts, so exactly one late arrival survives.

Added a concurrent regression test `test_debounce_concurrent_arrivals_after_running_survivor` that races two `maybe_debounce` calls after marking the survivor running and asserts exactly one survives while the other is debounced into the same window (never two independent windows). Full suite: 57 passed.

EE companion updated: windmill-labs/windmill-ee-private#631 (ref bumped to `b130144`).

rubenfiszel:
Addressed the second Codex P1 (concurrent late arrival's args dropped) — and took the reviewer-suggested step back to remove complexity rather than keep patching snapshot-sensitive SQL.

**The deeper race:** even with the correlated `EXISTS` running-check, the `_batch` CTE still read `v2_job_debounce_batch` with the statement snapshot. When J4 blocked on J3's lock and chained to J3, its batch lookup couldn't see J3's just-committed batch row → J4 got a *different* batch → J3's args were skipped when the survivor was pulled. Real data loss, as Codex reproduced.

**Simpler fix:** instead of cramming the running-check + batch chaining into one clever statement, `upsert_debounce_key` now takes a per-key transaction advisory lock (`pg_advisory_xact_lock(hashtext(key))`) up front, then does plain reads/writes:

1. lock the key (held until the caller's tx commits — covers `complete_debounced_job` too)
2. read the current holder + its `running` state (committed, no snapshot races)
3. running/absent holder → fresh window; queued holder → chain onto it
4. chain the batch (predecessor's batch is committed and visible under the lock)

Correctness now follows from the lock ("we hold an exclusive per-key lock, so every read sees committed state and no one else mutates concurrently") rather than from READ COMMITTED snapshot/EPQ reasoning. No more `CASE WHEN EXISTS`/CTE gymnastics.

Extended `test_debounce_concurrent_arrivals_after_running_survivor` to pull the survivor and assert its accumulation includes **both** racing late arrivals (shared batch) — directly covering the dropped-args case. Full suite: 57 passed (offline build matches CI).

EE companion updated: windmill-labs/windmill-ee-private#631 (ref → `8968cca`).

rubenfiszel:
Addressed the latest Codex P1 (push-vs-pull race from the split read+UPDATE).

**The bug I'd introduced:** the advisory-lock revision split the upsert into a holder `SELECT` then a bare `UPDATE ... WHERE key`. The worker pull path (`maybe_apply_debouncing`) deletes the key via `DELETE ... WHERE job_id` and does **not** take the per-key advisory lock, so the row could vanish between my `SELECT` and `UPDATE` → `fetch_one` finds no row → the debounce push becomes an internal error.

**Fix:** restore a **single atomic `INSERT ... ON CONFLICT DO UPDATE`** (with the correlated `EXISTS` running-check), instead of read-then-update. It holds the conflicting row lock and self-corrects — if the holder was just pulled and its key deleted, there's simply no conflict and we `INSERT` a fresh window rather than failing. The advisory lock is retained for its real job: making the upsert+batch statement run on a snapshot that already sees a prior arrival's committed `debounce_key` + batch rows (correct batch chaining without snapshot reasoning).

So the two roles are now cleanly separated:
- **atomicity / pull-delete robustness** → single `ON CONFLICT` statement
- **cross-push batch-chain visibility** → per-key advisory lock

Added `test_debounce_push_races_key_deletion_by_pull` — races a chaining push against the holder-key deletion 50× and asserts the push never errors. Full suite: 58 passed (offline build matches CI).

EE companion updated: windmill-labs/windmill-ee-private#631 (ref → `aa82e3f`).

rubenfiszel:
Added **claim-based exactly-once batch consumption** so the rare duplicate (and the underlying data-loss) is eliminated outright — without locking the worker pull hot path, per the design we discussed.

**Mechanism**
- Migration: `v2_job_debounce_batch` gains `consumed_at` + `consumed_by`.
- On pull, instead of deleting the batch, a survivor **atomically claims** its own row + any unclaimed siblings (`consumed_by = itself`) and accumulates exactly the rows it claimed. So if a narrow push/pull race ever puts two survivors on one batch:
  - the first claims everything and runs the full accumulation,
  - the second finds its row already consumed *by another job* → runs **empty** (no duplicate, no loss),
  - a **re-pulled** survivor recognizes its **own** prior claim (`consumed_by = self`) → keeps its accumulated args (no wipe on crash-recovery),
  - a **never-batched** job (CE/legacy) → keeps its own args (fallback preserved).
- `complete_debounced_job` (EE) now never completes a **running** predecessor (`FOR UPDATE` + `NOT running`), so its in-flight run isn't killed → no loss. The claim then removes the duplicate that guard would otherwise allow. The two together make accumulation **exactly-once**.
- Monitor GC sweeps consumed rows past a 1h grace (a re-pull whose row was GC'd safely falls back to its persisted args, so the grace isn't correctness-critical).

**Why not lock the pull path:** locking the generic mark-running step would tax the super-hot pull path for all jobs; gating it to debounced-only is invasive. Claim semantics get exactly-once lock-free, at the cost of a schema column + a GC sweep.

**Tests** (full suite 60 passed, offline build matches CI): `test_debounce_batch_consumed_exactly_once` (two survivors on one batch → one accumulates all, the other runs empty), `test_debounce_repull_keeps_accumulated` (re-pull doesn't wipe), plus all prior running-survivor/concurrent/race tests.

EE companion → `30d740e`.

Note: this grew the PR well beyond the original data-loss fix (it now also delivers exactly-once accumulation + a migration + GC). Happy to split the claim-based consume into its own follow-up PR if you'd prefer to land the smaller running-survivor fix first — your call.

rubenfiszel:
## Perf: claim-based consume vs old delete-on-consume (high-throughput debounced pulls)

pgbench, isolated DB, 200k pre-seeded 3-row batches, 12s runs, measuring the per-pull consume (accumulate + claim/mark vs accumulate + delete):

| clients | BEFORE (delete) TPS | NEW (claim/mark) TPS | per-op latency |
|--------:|--------------------:|---------------------:|----------------|
| 1  | 2,868  | 3,734  | 0.35 → 0.27 ms |
| 8  | 25,643 | 35,260 | 0.31 → 0.23 ms |
| 16 | 52,747 | 78,110 | 0.30 → 0.21 ms |

**Per-op is not a regression — it's ~30–48% faster.** An indexed mark (`UPDATE … SET consumed_at`) is cheaper than removing the rows + their index entries, and stays O(batch) regardless of table size (claim/accumulate are index lookups on `(debounce_batch)` / PK).

**The only cost is retention.** New keeps consumed rows until GC instead of deleting them immediately. I tightened the GC grace 1h → **10min** (a survivor that could still reference a row is pulled long before then; GC isn't correctness-critical). Steady-state extra rows ≈ ~10–15 min of debounced-pull volume — bounded, indexed, and reclaimed by the monitor sweep.

## Incidence on non-debounced jobs: none

- The entire pull-side block is gated on `is_djob || debounce_delay_s>0 || has_debounce_args` — non-debounced pulls skip it (zero extra statements, zero locks).
- Migration adds two **nullable, no-default** columns → metadata-only (no table rewrite/lock).
- GC touches only `v2_job_debounce_batch` (debounce-only table), every ~10 monitor iterations, server-side.

## Edge-case coverage (65 debounce tests, all green offline)

New this round, on top of the running-survivor / concurrent / race suite:
- `never_batched_keeps_own_args` — CE/legacy (no batch row) keeps its own args, not emptied.
- `concurrent_claim_disjoint` — two survivors pulled concurrently partition the batch disjointly: union = all items, none processed twice.
- `three_survivors_first_takes_all` — first survivor claims the whole batch; the rest run empty.
- `non_accumulate_deletes_batch` — plain debounce still hard-deletes its batch rows (no consumed-row leak on the non-accumulate path).
- `gc_consumed_batches` — GC deletes only past-grace consumed rows; keeps recently-consumed and unconsumed.
- (+ prior) `batch_consumed_exactly_once`, `repull_keeps_accumulated`, `push_races_key_deletion_by_pull`, `concurrent_arrivals_after_running_survivor`, running-survivor/limit/post-preprocessing cases.

PR #9781, opened 2026-06-25
test: cover debounce accumulation loss into already-running survivor

kiivanova:
### Summary

 Adds a regression test for a silent data-loss in post-preprocessing debounce with `debounce_args_to_accumulate` under a concurrency limit. Once a debounce survivor has pulled its batch and begun executing, it can no longer accumulate — but same-key messages that arrive afterward are still folded toward it and completed (`success=true`, skipped, result `Debounced by`/`Debounced Running by <id>`) with no error, and their accumulated elements appear in no executed run. The element survives only when it lands in a batch that a survivor actually pulls before starting; anything attached to an already-pulled/running survivor is dropped. A non-trivial step duration under `concurrent_limit` widens this window. Reproduced on EE v1.688.0 and v1.699.0: firing `[1,2,3]`, letting the survivor start running, then firing `[101,102,103]` processed `[1,2,3,103]` and lost `[101,102]` — all with `success=true`, no error.
  
Test-only change; it asserts the desired no-loss invariant, so it fails on current behavior and documents what the fix should guarantee. Suggested direction: once a survivor has pulled its batch / started running, a subsequent same-key message should start a new batch rather than be folded into it.
  
  ### Changes
  
  - `test_post_preprocessing_debounce_into_running_survivor_loses_message`: a survivor accumulates its own element and is marked running, then a later same-key message must survive (stay queued, debounce_key moved off the running survivor) rather than be dropped.
  - Adds a small `set_running` test helper.
  
  ### Test plan
  
  `cargo test -p windmill-queue --test debounce_test --features private,enterprise`. Placed inside the existing `#[cfg(feature = "private")] mod debounce` block. I couldn't run it in my checkout (the EE `maybe_debounce_post_preprocessing` impl isn't in the public tree) — please confirm against the EE build; if the buggy end-state differs, the `late_survived` / `key_moved_off_running_survivor` checks are the spots to adjust.

---
## Summary by cubic
Add a regression test that reproduces silent data loss in post-preprocessing debounce with `debounce_args_to_accumulate` when the survivor is already running under a concurrency limit. The test asserts that late same-key messages start a new batch (stay queued) rather than being folded into the running survivor and dropped, and adds a small `set_running` helper.

<sup>Written for commit fdd2885407f2343614dd310e08250fbfb6a5e13c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9781?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-queue/tests/debounce_test.rs
+++ b/backend/windmill-queue/tests/debounce_test.rs
@@ -4430,6 +4430,127 @@ mod debounce {
         Ok(())
     }
 
+    /// Helper: mark a queued job as running (simulates a survivor that the
+    /// concurrency limiter has just started executing).
+    async fn set_running(db: &Pool<Postgres>, job_id: &Uuid) {
+        sqlx::query!(
+            "UPDATE v2_job_queue SET running = true WHERE id = $1",
+            job_id,
+        )
+        .execute(db)
+        .await
+        .expect("set running");
+    }
+
+    /// Test: a message that arrives while its debounce survivor is already running
+    /// must start a new batch, not be folded into the running survivor and dropped.
+    /// A survivor that has pulled its batch and begun executing can no longer
+    /// accumulate, so the late message's args would otherwise be lost with no error.
+    #[sqlx::test(migrations = "../migrations", fixtures("base"))]
+    async fn test_post_preprocessing_debounce_into_running_survivor_loses_message(
+        db: Pool<Postgres>,
+    ) -> anyhow::Result<()> {
+        let settings = DebouncingSettings {
+            debounce_delay_s: Some(5),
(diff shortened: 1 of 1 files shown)


kiivanova:

rubenfiszel:
fixed in #9780 , thanks!

PR #9793, opened 2026-06-25
feat: add dev workspaces paired with a lockable prod workspace

hugocasa:
## Summary
Adds **dev workspaces**: a persistent, name-unconstrained fork that is the standing editable environment, paired with a "prod" workspace holding the same code with a different environment (resource/variable values). Built on the existing fork primitive (parent link, `deploy_to`, `workspace_diff`, `CompareWorkspaces`, `ForkWorkspaceBanner`) rather than a new mechanism. Auto-sync is intentionally out of scope here; promotion stays manual via the existing Compare & Deploy flow.

This update also adds **per-environment resource/variable config from the Compare & Deploy page** (the existing `ws_specific` mechanism, items excluded from the diff with an explicit "Create in &lt;other&gt;" to seed a missing copy), steers **AI-chat sessions toward the dev workspace**, and uses "root" rather than "prod" in the dev-workspace UI.

## Changes
**Backend — dev workspaces**
- Migration: `workspace.is_dev_workspace` column, a partial unique index enforcing one active dev per parent, and a CHECK constraint (`is_dev_workspace` ⇒ has a parent).
- Fork detection keys off `parent_workspace_id` **or** the `wm-fork-` prefix (so an orphaned fork is still detected): delete/owner gates (`workspaces_extra.rs`) and the per-workspace job-routing fast-path (`tags.rs`).
- `validate_dev_workspace_id` (git-branch-safe, no prefix); `create_fork` gains `is_dev_workspace` + per-rule lock flags (`lock_prod_deploy`/`lock_prod_forking`), admin-gated, enforced on both the plain and git-branch fork paths.
- New `attach_dev_workspace` / `detach_dev_workspace` (one canonical dev per prod; no nesting/cycles; require admin of both sides).
- Prod-lock reuses the existing `DisableDirectDeployment` / `DisableWorkspaceForking` rules under a reserved name `dev_workspace_lock` (auto-applied on pairing, removed on detach/dev-delete/dev-archive). The generic rule layer carries no feature-specific knowledge (no name-based messages).
- **Lifecycle is consistent**: rename carries `is_dev` through (so a renamed dev stays the canonical dev); deleting/archiving an attached dev requires prod-admin and clears the prod lock + invalidates the rules cache; the fork-parent routing cache is invalidated on attach.

**Per-environment resource/variable config (workspace-specific)**
- `set_ws_specific` marks a resource/variable workspace-specific from Compare & Deploy. Authorized like the editors (`require_is_writer` — owner/folder/item writer — + `check_deploy_rules`), with the path shape validated to avoid a 500.
- A workspace-specific item is **excluded from the diff entirely** (per-item compare and the cached read-time path both suppress it), so each environment keeps its own value.
- Seeding a missing copy is an **explicit, strictly create-only** action: the "Workspace-specific items" panel shows **"Create in &lt;other&gt;"** when the item is on only one side. It re-checks actual existence and **never overwrites** an existing target (marks it workspace-specific instead), copies the value incl. secrets (reuses the normal deploy, authorized against the target), and shows a confirm.
- `CompareWorkspaces`: "Make workspace specific" on resource/variable diff rows + the "Workspace-specific items" panel to seed/revert ("Make shared"); unpinning invalidates the cached diff so the now-shared item re-evaluates.

**AI-chat session picker**
- The "Run in" picker shows the indigo "dev" badge (rows + trigger); a new session defaults to the current workspace, steering to the dev only when you're in a prod you can't deploy to.
- Reads the prod root's **own** protection rules (so it's correct from any family workspace): prod is disabled only when genuinely deploy-locked for you (admin bypass recognized); the root row stays disabled while rules load.
- "Create new fork" defaults to "Fork from &lt;dev&gt;"; an admin "Fork from &lt;root&gt;" entry appears when you can fork the root (keyboard-reachable). `NoDirectDeployAlert` only suggests forking when you actually can.

**Fork membership**
- An opt-in **"Copy members"** toggle when *creating* a dev fork copies the source's `usr` rows (carrying each member's role, including service-account type) so the team can work in the dev. (Not offered when *attaching* an existing workspace — it already has its own team.)

**Frontend**
- `workspaceHierarchy.ts` helpers (`workspaceIsFork`, `findCanonicalDevWorkspace`, excluding disabled devs); `isFork` migrated off the prefix across banners, sidebar, compare page, and session components.
- `CreateWorkspaceInner`: "Persistent dev workspace" toggle, separate lock toggles in a "Protect &lt;root&gt;" sub-card, Workspace ID under the fork name.
- `buildForkEditUrl` / `editInForkAllowed` redirect edits into the canonical dev on a locked prod (admin bypass honored); EE-gated **Dev workspace** settings tab (hidden on ordinary forks) to attach/detach + lock.

- The edit redirect only fires where the edit-in-fork affordance already shows (`!canWrite`); a prod admin who still has write sees the normal editor and hits the backend deploy block instead of auto-redirecting. Deferred.
- Deferred by design: auto-sync dev→prod, and trigger/schedule special handling.
- `attach_dev_workspace`'s one-dev-per-parent pre-check runs before the transaction, so a concurrent attach surfaces a 500 (unique-index violation) rather than a friendly 400. The index keeps data correct; only the UX degrades.
- A git-synced prod that is locked still receives **CLI** `wmill sync push` only via a token with deploy bypass (same as any deploy to a locked workspace) — don't lock a prod whose source of truth is CI git-sync, or give that token bypass.

## Test plan
- [x] Backend compiles (full-workspace `cargo sqlx prepare`, ee features); frontend `svelte-check` 0 errors.
- [x] Create a dev workspace: no `wm-fork-` prefix, `is_dev_workspace=true`, `deploy_to=prod`, prod gains `dev_workspace_lock`.
- [x] `set_ws_specific`: non-admin non-writer rejected (400) on pin **and** unpin; admin allowed; malformed path → 400 (not 500); variable path covered.
- [x] Lifecycle: deleting/archiving an attached dev requires prod-admin and clears the prod lock; deleting a prod with a dev is blocked ("detach first"); attaching a workspace that has its own dev is rejected; detach→re-attach restores the dev + lock; a manual protection rule survives detach.
- [x] Consolidated migration applies cleanly on a fresh DB (CHECK rejects `is_dev` without a parent; unique index rejects a 2nd dev).
- [ ] **On EE, manual e2e (focus of next pass)** — see the checklist comment.

## Screenshots
**Dev workspace settings — root with a dev / ordinary fork / root without a dev**
root
fork
attach

**Session "Run in" picker (Fork from dev / Fork from root / root / dev badge)**
picker

---
## Summary by cubic
Adds dev workspaces paired to a lockable prod so teams edit safely in dev; on a locked prod, Edit opens the item in its dev with a clear “Edit in <dev>” across editors and run pages. Compare & Deploy now supports workspace‑specific resources/variables with an explicit, create‑only “Create in <other>,” and sessions steer users toward the dev workspace.

- **New Features**
  - Dev workspaces: `is_dev_workspace` flag with a unique index enforces one active dev per prod; new attach/detach endpoints prevent nesting/cycles, require prod admin + dev admin (or superadmin), allow attaching a same‑family fork, and upsert a reserved `dev_workspace_lock` for `DisableDirectDeployment`/`DisableWorkspaceForking` (removed on detach or dev delete). OpenAPI updated.
  - Editing and UI: “Edit in <dev>” deep‑links to the dev item; devs show a “dev” badge, sort ahead of forks, auto‑expand under prod, and are labeled “Dev workspace of <prod>.” The fork form offers “Persistent dev workspace” with separate deploy/fork locks; settings add a “Dev workspace” tab to attach/detach and lock prod (CE).
  - Compare & Deploy: workspace‑specific panel lists pinned resources/variables (`ws_specific`) with “Make workspace specific”/“Make shared,” plus an explicit, strictly create‑only “Create in <other>” that seeds linked variables and never overwrites; pinned items are suppressed from the main diff and per‑item diff cache is invalidated.
  - Sessions: picker/bars badge the dev, default to the current workspace, read prod’s actual rules (with bypass), disable locked prod for non‑admins, suggest “Fork from <dev>,” hide blocked options; persistent devs are never offered for deletion.
  - Fork membership: optional “Copy members” on dev‑workspace creation copies source members/roles (not offered on attach).

- **Bug Fixes**
  - Lifecycle/auth and locks: admin‑only dev‑workspace creation and locked git‑branch forks; attach/detach require admins on both sides and reject cycles; server‑side guards enforce one dev per prod; lock checks are evaluated in the acted‑on workspace; dev flag cleared on archive/replace and prod’s `dev_workspace_lock` cleared on detach/dev delete; archiving a dev now tears down the pairing and lock atomically; on root rename, `workspace_settings.deploy_to` is updated to follow; dev pairing is visible to prod admins even if they aren’t members of the dev.
  - Compare & Deploy guards: create‑only inserts for “Create in <other>,” `ws_specific` lists scoped correctly, accept `g/` group variable paths, target paths validated, and per‑item diff caches invalidated; the `dev_workspace_lock` rule name is reserved from the public API.
  - Routing and sessions: fork/dev detection keys off `parent_workspace_id` and routing cache is invalidated on attach/detach and root rename; session picker preserves a staged fork’s source when re‑entering create mode; fork creation copies creator metadata.
  - UI/flows: “Copy members” is now gated to dev‑workspace creation (not shown on ordinary forks).

<sup>Written for commit 5a9ee80dbbd1966a736c204aa06c8a81ec562fdd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Ran a branch-diff review. Two high-severity findings fixed in f6e8056ecc:

- **Privilege escalation**: `create_workspace_fork` had no admin check, so a non-admin who can fork could pass `lock_prod=true` and apply a `DisableDirectDeployment` rule to prod. Now `lock_prod` requires parent-admin (matching `attach_dev_workspace` and `create_protection_rule`).
- **Attach cycle**: `attach_dev_workspace` verified the dev candidate was parent-less but not that prod is a root, allowing an A↔B parent cycle that would hang hierarchy traversal. Now rejects a prod that already has a parent.

Also added unit tests for `validate_dev_workspace_id` (prefix-less + empty/git-unsafe).

Non-blocking notes left as-is: the one-dev-per-prod pre-check is non-transactional (the partial unique index is the real guard; a concurrent race surfaces a 500 instead of a friendly 400), and `WsSpecificVersions` still buckets by id prefix (cosmetic; a dev workspace sorting with regular environments there is reasonable).

hugocasa:
Follow-ups pushed:

**876160c0de — redirect locked-prod edits into the dev workspace.** New `maybeRedirectEditToDevWorkspace()` (called from the 4 editor routes' `onMount`) re-targets the current editor into the dev workspace when the current (prod) workspace has a canonical dev **and** `DisableDirectDeployment` is active **and the user can't bypass it**. This uses the ruleset as the gate (per review discussion) instead of `canWrite`, so it correctly redirects non-admin developers-with-write while preserving the admin escape hatch (admins bypass the lock and keep editing prod directly for hotfixes). No-op for everyone else and when the URL already carries `?workspace=` (loop guard).

**c571cfebec — make the dev-workspace settings tab available on CE.** It was EE-gated by copy-paste from the Deployment UI tab, but the backend (create-as-dev, attach, detach, the protection-rule lock) all run on CE, so the gate just made attach unreachable on CE. Removed it.

Verified live on the CE backend (no EE restart needed):
- Prod settings → Dev workspace tab now shows the real attach/detach UI on CE (screenshot below): linked dev + **Go to dev workspace** / **Detach**.
- **Detach** confirmed in DB: `is_dev_workspace` cleared and the `dev_workspace_lock` rule removed (parent lineage kept).
- The post-detach **attach form** renders (Select + Lock toggle + Attach + Create new).

Not browser-exercised (honest notes): the **attach mutation** end-to-end (CE caps at 2 non-admin workspaces, so I couldn't create a third standalone to attach), and the **edit-redirect** runtime (it only fires for a non-admin developer-with-write on a locked prod; as superadmin I bypass). Both are type-checked and follow existing patterns; the redirect is an additive early-return.

dev-workspace-settings-prod

hugocasa:
Re-verified on an **EE** backend (workspace cap lifted), closing the gaps from the previous comment:

**Attach mutation — end to end via the settings UI.** Created two standalone workspaces `envprod`/`envdev`, selected `envdev` in the prod's Dev workspace tab, clicked Attach. DB after: `envdev.parent_workspace_id=envprod`, `is_dev_workspace=true`, `workspace_settings.deploy_to=envprod`, and `envprod` gained the `dev_workspace_lock` rule. The candidate Select correctly listed only root workspaces (excluded the already-parented `dev`).

**Rejection paths (authenticated API):**
- Uniqueness: `attach(envprod, prod)` when envprod already has a dev → `400 "Workspace 'envprod' already has a dev workspace ('envdev')"`.
- Cycle guard (from f6e8056): `attach(envdev, prod)` where the prod side isn't a root → `400 "Workspace envdev is itself a fork or dev workspace and cannot be a prod workspace"`.

Still not runtime-exercised: the **edit-redirect positive path**, which by design only fires for a non-admin developer-with-write on a locked prod (admins bypass the lock, so as superadmin I can't trigger it). It's wired into the 4 editor routes, type-checks, and is logically incapable of redirecting a bypassing user — so it can't regress admin editing. Happy to add a non-admin integration test if you want it pinned.

hugocasa:
Follow-up pushed (8c48e593e3) — the **coherent prod lock** discussed in review.

**Backend:** `lock_prod_workspace` now applies both `DisableDirectDeployment` **and** `DisableWorkspaceForking` on the `dev_workspace_lock` rule (still removed wholesale on detach). So a locked prod can't be deployed to *or* forked by non-admins; admins bypass both (their existing escape hatch).

**Frontend:** the edit affordance is decoupled from the forking gate so it survives the fork-block — when the current workspace has a canonical dev, the button shows and is relabeled **"Edit in dev workspace"** (routing to the dev workspace, which creates no fork). The "Fork workspace" entries now use a `canCreateFork` check that respects the admin bypass, so they hide for non-admins on a locked prod but remain for admins. Touched the list rows, get pages, run page, and the three fork-creation entries via two small helpers in `editInFork.ts`.

Verified live (EE backend):
- Non-admin developer forking the locked prod → `403 "Forking this workspace is forbidden"`; admin → `200` (escape hatch).
- The edit button relabels to **"Edit in dev workspace"** on the script page and list rows (→ `?workspace=dev`).
- `npm run check` 0 errors; backend `cargo check` clean.

Deferred on purpose: the 4 editor-header "Edit in fork" buttons (ScriptBuilder/FlowBuilder/AppEditorHeader/RawAppEditorHeader) — they just hide on a locked prod rather than mislabel, so nothing breaks; easy to fold in later. And the deep-link `onMount` redirect (876160c0de) is now largely redundant since the relabeled button is the reliable funnel — can drop it if you prefer.

hugocasa:
Follow-up pushed (4888824d25) — UX polish on the funnel:

**(a) Land on the dev item *page*, not the editor.** `buildForkEditUrl` now targets `/<kind>/get/<path>?workspace=dev` when a canonical dev exists, so clicking "Edit in dev workspace" drops you on the item's page in the dev workspace (with the "Dev workspace of prod" banner for context) and you open the editor deliberately from there — instead of teleporting straight into the editor. Also **removed the deep-link `onMount` redirect** (commit 876160c0de) and its helper: it was redundant now that the button is the reliable funnel, and it had the cold-load timing gap.

**(b) Dev-aware lock messaging** (the generic "Fork or PR required" text was wrong for this flow):
- Backend: when the `dev_workspace_lock` ruleset blocks an action, `check_user_against_rule` now returns tailored text. Verified: `403 "'prod' is a prod workspace. Make your changes in its dev workspace instead of forking it."` (moved `DEV_WORKSPACE_LOCK_RULE_NAME` to windmill-common so the check can recognize it).
- Frontend: `NoDirectDeployAlert` now reads **"Edits happen in the dev workspace"** with "Make your changes in its dev workspace … and promote them here" plus a **Go to dev workspace** button, instead of the generic protection text — when the current workspace has a canonical dev.

Verified live (home + script page): the relabeled button links to the dev *get* page; the alert shows the dev message + button. `npm run check` 0 errors; backend `cargo check` clean.

hugocasa:
Follow-up pushed (c9301f3bf2) — two guardrails:

**No nested dev workspaces.** New `ensure_dev_parent_is_root` rejects creating a dev workspace whose parent is itself a fork/dev, in both `create_workspace_fork` and `create_workspace_fork_branch` (fails before any git branch is created). It's a coherence rule, not a permission, so it applies to admins too. Verified: `400 "Cannot create a dev workspace of 'dev' because it is itself a fork or dev workspace."`

**Dev option hidden when a dev already exists.** In `CreateWorkspaceInner`, the "Persistent dev workspace" toggle now only appears when forking a root workspace that has no dev yet (mirrors the one-dev-per-prod + no-nesting rules). Otherwise it shows "This workspace already has a dev workspace (X)". Verified on the fork page: forking `prod` (which has `dev`) hides the toggle and shows the hint.

`npm run check` 0 errors; backend `cargo check` clean.

hugocasa:
Follow-up pushed (8a6a66268d) — dev workspaces are now visually distinguished in the workspace list/switcher.

- A dev workspace nests under its prod (structurally honest — it's a child) but carries a clear indigo **"dev"** badge in both the workspaces tree (`WorkspaceCard`) and the sidebar switcher (`WorkspaceMenu`), so it stands out from throwaway forks.
- It sorts **ahead** of throwaway forks among a prod's children (shared sort in `workspaceHierarchy.ts` + the tree card).
- In the switcher it's always visible (the flattened list shows all workspaces); in the tree it's under the prod's expand chevron with the badge. Left auto-expand out for now — the badge + ordering already make it stand out — easy to add if we want prod nodes with a dev expanded by default.

`npm run check` 0 errors.

dev-workspace-tree-badge

hugocasa:
UX batch pushed (c6af517c8c, 00d2b20f4d, plus earlier):

- **Edit label** is now **"Edit in <dev name>"** (e.g. "Edit in dev") instead of the longer "Edit in dev workspace" — uses the dev workspace's name, no extra API calls.
- **Editor-header consistency**: the 4 in-editor "Edit in fork" buttons (script/flow/app/raw-app) now use the same decoupled `editInForkAllowed`/`editInForkLabel` as the list rows and detail pages, so they show "Edit in <dev>" on a locked prod too.
- **Auto-expand**: prod nodes that have a dev start expanded in the workspaces tree, so the dev is visible without a click (manual collapse still wins).
- **Two lock toggles**: the single "Lock prod" toggle is split into **"Block direct edits in prod"** (DisableDirectDeployment) and **"Prevent forking prod"** (DisableWorkspaceForking), independently selectable on both the create-as-dev form and the attach UI. Backend applies them independently — verified: deploy-only attach → `rules=1`, both → `rules=3`.
- Nav bar already reads **"Dev workspace of prod"** (not "Fork of prod"); kept it as text since that menu component renders text only (no badge slot).

`npm run check` 0 errors; `cargo check` clean; client regenerated.

hugocasa:
A few things worth changing before/after merge (from a focused review on not regressing existing forks / `ws_specific` items):

### 1. Replace `ws_specific` create-if-missing with an explicit "Create on remote" action
Create-if-missing is a global change to what `ws_specific` means: an item present on only one side, previously fully hidden, now shows as a deployable "create" row and can be swept into a "deploy all". For the common both-sides case nothing changes, but it's a behavior change for existing pins.

Cleaner + fully backward-compatible: keep `ws_specific` out of the normal diff entirely (revert `compare_two_resources` / `compare_two_variables` and the `workspace_diff` read-time filter to unconditional suppression) and add a per-item **"Create in `<other workspace>`"** button in the Workspace-specific items section, shown only when the item is missing on that side. Back it with a create-only endpoint that authorizes against the **target** (write check + `check_deploy_rules`, so pushing into a locked prod as a non-admin is blocked). Note it copies the current value (incl. secret values) to the other env — worth a one-line confirm.

### 2. Don't base the lock's block message on the rule name
The only behavior conditional on the `dev_workspace_lock` name is the tailored block message in `check_user_against_rule` (`windmill-common/src/workspaces.rs:512`) — the other three references just create/delete the feature's own rule by that key. The UI already derives the dev-specific guidance from the pairing itself (`NoDirectDeployAlert` → `findCanonicalDevWorkspace`), so the name-conditional branch is redundant: drop it and return the generic message, keeping the generic rule layer free of feature-specific knowledge. The name-based rule itself is fine as a plain, visible workspace rule.

### 3. Audit all fork-detection sites to "check both" (prefix OR parent)
The session "delete the fork" affordance (`SessionWrapper.svelte:97`, `SessionPicker.svelte:292`) gates on bare `parent_workspace_id`, so an orphaned `wm-fork-` fork (parent hard-deleted → parent nulled) is no longer offered. Use the shared `workspaceIsFork` (prefix OR parent) + keep the `is_dev_workspace` exclusion. We've already fixed several of these piecemeal during the prefix→parent migration — worth a sweep so every fork-ness check (frontend and backend) goes through the shared helper rather than a raw prefix or raw parent check.

### 4. Dev-workspace rename can leave a locked prod with no active dev
`change_workspace_id` preserves the parent but drops `is_dev_workspace` and leaves the prod lock, so a locked prod can end up with no active dev (non-admins blocked until someone re-attaches). Carry the flag through — neutralize the old row's `is_dev_workspace` before the INSERT so the one-dev-per-parent unique index holds, then set it on the new row — or reject renaming a dev workspace (detach first).

### Minor / non-blocking
- `attach_dev_workspace`'s uniqueness pre-check runs before the transaction, so a concurrent attach surfaces a 500 (index violation) instead of the friendly 400. The unique index keeps data correct; only the UX degrades.
- The Dev workspace settings tab falls through to the attach form for ordinary (non-root) forks, which the backend then rejects — hide it for non-root forks.

hugocasa:
## Manual e2e checklist (recent review-round changes)

**Setup:** EE backend on the `dev-workspaces-autosync` branch (HEAD `c0dff48948`+), a root workspace `prod` with ≥1 non-admin member, and a non-admin user. Done = every box checked.

### 1. Dev-workspace lifecycle (backend hardening)
- [x] Create a dev from `prod` with both locks on → `prod` shows the locks; dev shows the "dev" badge.
- [x] **Rename** the dev workspace id → it stays the canonical dev (home banner/picker still resolve a dev; `prod` is not left locked-without-a-dev). *(was the rename bug)*
- [x] As a **dev-owner who is not a prod-admin**, delete the dev → **rejected**. As a prod-admin, delete it → `prod`'s lock is removed.
- [x] **Archive** an attached dev (as prod-admin) → `prod`'s lock is removed.
- [x] Delete `prod` while it has a dev → **blocked** ("detach the dev workspace first").
- [x] Attach a workspace that already has its own dev as another prod's dev → **rejected** ("already has a dev workspace").
- [x] Detach the dev, then re-attach it → dev designation + lock restored.

### 2. Workspace-specific items (Compare & Deploy)
- [x] Pin a resource that exists in both dev & prod → it disappears from the diff; appears in "Workspace-specific items"; "Make shared" brings it back.
- [x] Pin a resource that exists **only in dev** → "Workspace-specific items" shows **"Create in prod"** → confirm → it's created in prod with the same value (check a secret value too) and stays workspace-specific.
- [x] **Create-only safety:** take a resource present in **both** but workspace-specific-marked on **only one** side → click "Create in &lt;other&gt;" → the other side is **not overwritten** (toast says "already exists … marked workspace-specific").
- [ ] As a **non-admin who can write** a resource (folder writer) → can pin it. As a **read-only** member → pin is rejected.

### 3. AI-chat session picker (in a prod that has a dev)
- [x] "Run in" shows the dev badge; default fork entry is "Fork from &lt;dev&gt;"; admins also see "Fork from &lt;root&gt;" (reachable by keyboard).
- [x] A new session defaults to the workspace you're currently in.
- [ ] As a non-admin on a locked prod, the prod row is disabled/not selectable.

### 4. Settings + membership
- [x] Dev-workspace settings tab is visible on a **root** and on a **dev** workspace, but **hidden** on an ordinary fork.
- [x] Create a dev with **"Copy members"** on → prod's members (including any **service account**) appear in the dev and are in its `all` group.

_Generated to track the manual e2e pass; ignore once merged._

hugocasa:
Re finding 2 (create-in-other / external-KMS secret variables): acknowledged, but accepting it for this PR rather than fixing.

It's a pre-existing property of `create_variable` — the external secret write isn't transactional with the DB path-conflict check — not introduced here. The trigger is a sub-millisecond, same-`(workspace, path)` concurrent create; for a manual "Create in <other>" admin action that's negligible in practice. The variable create path is already create-only at the DB layer (`check_path_conflict`), so the normal flow can't overwrite. Reworking the secret-storage ordering to make the external write conditional on the conflict check is out of scope for this change; can revisit separately if it becomes a real concern.

Findings 1 (stale `deploy_to` after root rename) and 3 (prod-admin dev visibility in settings) are fixed in the next commit.

PR #9798, opened 2026-06-25
perf: eliminate dual-connection DB pool contention across worker, queue, and api

rubenfiszel:
## Summary

Extends the fix from #9789 / #7861 (reduce DB pool contention by eliminating dual-connection patterns). The anti-pattern: running a query against the **pool** (`db`) while a `Transaction` (`tx`) is already held checks out a *second* simultaneous connection. On the per-worker pool (default max **5**) this triggers `pool timed out while waiting for an open connection` under concurrent load.

A full-codebase audit found and fixed the remaining genuine cases. Each is fixed by either reusing the held transaction (`&mut *tx`) or moving the read **before** `begin()` — chosen per-site so semantics are preserved (notably, deliberately non-RLS pool reads are *moved before begin* rather than routed through an RLS `tx`).

## Changes

**Worker pool (hot paths, max 5 connections):**
- `worker_flow.rs` (`update_flow_status_after_job_completion_internal`, same fn as #9789): `retrieve_flow_jobs_results` + `get_id_ctx_for_expr` made executor-generic; `evaluate_stop_after_all_iters_if` takes `&mut Transaction`. The three reads in the loop/branchall completion path now run on the held `tx`.
- `windmill-queue/jobs.rs`: `commit_completed_job` resolves the concurrency-limit settings before `begin()`; `try_schedule_next_job` email lookup via `tx`; `push_inner` native-retry `dedicated_worker` read via the (previously unused) `fetch_scalar_isolated!` macro.
- `windmill-queue/schedule.rs` (`push_scheduled_job`): email lookup, tag check, and runnable-settings prefetch all via the held `tx`.
- `windmill-common`: `get_email_from_permissioned_as`, `is_super_admin_email`, `check_tag_available_for_workspace_internal`, `runnable_settings::{from_handle, Settings::get}` made executor-generic (backward-compatible); added `prefetch_cached_tx` / `prefetch_cached_from_handle_tx`; `clone_script` uses them.

**Server pool (API handlers):**
- `workspaces_export.rs`, `apps.rs`: non-RLS pool reads hoisted before the RLS `begin()`.
- `api-configs`, `api-groups/granular_acls`, `api-scripts`, `api-users`, `api-workspaces`, `store/resources`: reads routed to the held `tx` (or `begin()` moved after non-tx reads).

## Correctness (worker_flow, verified)

- Sub-job result reads (`v2_job_completed`): sub-jobs commit before this tx starts and the tx never writes them → under READ COMMITTED the tx-read is identical to the pool-read.
- `get_mini_pulled_job` via tx: `get_transform_context` reads only immutable identity fields (`id` / `root_job` / `parent_job`) plus the **in-memory** `status` — never `flow_job.flow_status` — so uncommitted writes in the tx are irrelevant.
- Errors are `?`-propagated before each read, so the tx is never in an aborted state at the read points.
- An exhaustive sweep of every `.begin()` scope in `windmill-worker` confirms these were the only genuine dual-connection sites; the one remaining tx-held pool read (the zombie-monitor ping in the parallel-push loop) is deliberately on the pool and left unchanged.

## Test plan

- [x] `cargo check --features quickjs` and `--features quickjs,enterprise` — both compile clean
- [x] `test_stop_after_all_iters_if_bad_expr_parallel_forloop` / `_branchall`, `test_branchall_nested` / `_skip_failure` pass (exercise the changed worker_flow reads)
- [x] 35 `schedule_push` tests pass — incl. `test_tx_usable_after_failure` / `_success` (held tx not poisoned), `on_behalf_of_email` (email via tx), retry/concurrency (prefetch via tx)
- [x] 4 `native_retry` tests pass (native-retry `dedicated_worker` read + concurrency settings)
- [ ] EE `private`-feature build not validated locally (EE module symlinks absent in this worktree); EE callers of the generified functions verified by inspection to pass owned pools

## Out of scope (deferred, documented)

Cache-gated / cloud-only / EE-only MEDIUM-LOW cases on the larger server pool where a safe fix is invasive: `push_inner` cloud-quota preamble (`#[cfg(cloud)]`, needs several cached EE helpers made tx-aware), EE vault-rename + schedule/error handlers, `workspaces_export` per-script prefetch / `build_crypt`, and the `apps.rs` protection-rule checks (hoisting changes error-ordering).

---
## Summary by cubic
Eliminates dual-connection DB pool contention by reusing open transactions or moving reads before begin() across worker, queue, and API. Also isolates flow-eval reads in a SAVEPOINT, keeps RLS-sensitive reads on the non‑RLS pool, and adds a failpoint-backed regression test proving the SAVEPOINT behavior.

- **Bug Fixes**
  - Worker: route sub‑job result reads and expr‑context lookups through `&mut tx`; `stop_after_all_iters_if` now evaluates on the tx and its reads are wrapped in a SAVEPOINT to keep the outer tx usable on error.
  - Queue: resolve concurrency‑limit settings on the pool before opening the completion tx and gate follow‑ups on a stored flag; `try_schedule_next_job` resolves email via the tx; `dedicated_worker` routing intentionally stays on the non‑RLS pool.
  - Scheduler: runnable‑settings prefetch plus email/tag checks run on the non‑RLS pool when invoked under an RLS tx.
  - API: hoist non‑RLS reads before `begin()` or execute them on the tx across configs, users, workspaces, apps, export, and triggers; resolve on‑behalf‑of defaults and workspace dependencies before opening the RLS tx; fix http trigger to pass the pool to `get_email_from_permissioned_as`.
  - Tests: add a failpoint-backed regression that aborts the inner eval to confirm the SAVEPOINT keeps the outer tx committable; enable via the new `failpoints` feature.

- **Refactors**
  - Generalized helpers to accept `PgExecutor`: `is_super_admin_email`, `check_tag_available_for_workspace_internal`, runnable settings `get`/`from_handle`, and `get_email_from_permissioned_as`.
  - Added tx‑aware runnable‑settings prefetch (`prefetch_cached_tx`) and used it in `clone_script`; updated call sites to prefer `&mut tx` where a transaction is open.

<sup>Written for commit 57fde4f0750278a99e0d113b1d807d669ab602d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed in c695a36d36. The `stop_after_all_iters_if` reads are now wrapped in a SAVEPOINT (`tx.begin()`, same connection — no second pool connection): on a DB read failure we roll back to the savepoint so the outer transaction stays usable, exactly matching the previous pool-read behaviour where a failed read left `tx` usable and the iteration was marked failed and committed. Verified the happy path with the parallel for-loop/branchall `stop_after_all_iters_if` tests.

PR #9807, opened 2026-06-26
feat(sdk): allow overriding worker tag when running jobs (WIN-2105)

rubenfiszel:
## What

Adds an optional `tag` parameter to every job-running helper across the client SDKs, so callers can override the worker tag a job is dispatched to. The backend already supports this — `RunJobQuery` has a `tag` field that all `jobs/run/*` handlers honor (`run_query.tag.clone().or(...)`) — and the OpenAPI spec already documents it (`WorkerTag` → `tag` query param) on every run endpoint. This PR just exposes it through the hand-written SDK wrappers.

Fixes WIN-2105

## Non-breaking by construction

The `tag` parameter is **appended last** and defaults to `null`/`None` everywhere, so all existing positional and keyword callers keep working unchanged.

- **TypeScript** (`typescript-client/client.ts`): `runScript`, `runScriptByPath`, `runScriptByHash`, `runFlow`, `runScriptAsync`, `runScriptByPathAsync`, `runScriptByHashAsync`, `runFlowAsync` (+ the private internals). Stale root `client.d.ts` declarations kept in sync.
- **Python** (`python-client/wmill/wmill/client.py`): all `Windmill` methods and module-level wrappers (`run_script[_by_path|_by_hash][_async|_sync]`, `run_flow_async`, deprecated `run_script*`).
- **PowerShell** (`powershell-client`): `Invoke-WindmillScript`, `Start-WindmillScript`, `Start-WindmillFlow` and the underlying `RunScriptAsync`/`RunFlowAsync` methods.
- **Rust** (`rust-client/src/client.rs`): Rust has no optional/default params or overloading, so the existing `run_script_async`/`run_script_sync` signatures are left **untouched** and new `run_script_async_with_tag` / `run_script_sync_with_tag` variants are added instead. The private inner threads `tag` into the correct positional query-param slot of the generated `job_api` (compile-verified — see Validation below).

Go client and React SDK have no hand-written run-job helpers; their generated bindings already expose `tag` from the OpenAPI spec.

## Validation

- ✅ **End-to-end**: ran `POST /api/w/test/jobs/run/p/u/admin/tagtest?tag=mycustomtag` against the local backend; the created `v2_job` row had `tag = mycustomtag`. This is exactly the request the updated SDKs now emit.
- ✅ **Python**: `python -m py_compile` passes.
- ✅ **TypeScript**: `tsc --noEmit` on `client.ts` reports only pre-existing `process`/node-type errors — none from the `tag` additions.
- ✅ **PowerShell**: module parses cleanly (`Parser.ParseFile`).
- ✅ **Rust** (compile-verified): the generated `windmill_api` crate is produced at build time via `openapi-generator-cli` (needs Java, not installed here), so it can't be regenerated locally — but the generated bindings are published to crates.io. Pulled `windmill-api@1.740.0` (matches `version.txt`) and confirmed the real signatures place `tag: Option<&str>` as the **5th optional arg** in both `run_script_by_path` (`…, parent_job, tag, cache_ttl, …`) and `run_script_by_hash` (`…, parent_job, tag, cache_ttl, …`) — exactly the slot the code fills. Then compiled the updated `client.rs` against that published crate: `cargo build` ✅ and `cargo build --features async` ✅ (a wrong arity/type would fail to compile). So the positional mapping is confirmed by the compiler, not just by inspection.

---
## Summary by cubic
Adds an optional `tag` parameter to SDK job-run helpers so callers can override the worker tag when dispatching jobs. Non-breaking and addresses WIN-2105; also fixes stale system prompt type declarations.

- **New Features**
  - Added `tag` across `typescript-client`, `python-client`, `powershell-client`, and `rust-client` (via `run_script_*_with_tag` variants); forwarded as the `tag` query param.
  - Updated `typescript-client` typings (`client.d.ts`) and `cli` guidance (`skills.gen.ts`); regenerated SDK docs and system prompts.

- **Bug Fixes**
  - PowerShell: Restored original `RunScriptAsync` (4-arg) and `RunFlowAsync` (3-arg) overloads; new overloads add optional `Tag`.
  - System prompts: `generate.py` now emits `prompts.d.ts` with `export declare const X: string;` to prevent drift with `prompts.ts`.

<sup>Written for commit 9871e4fbf64184fca5f5c6f005a741a0bc293c84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
### Addressed the one CI-review nit: Rust client now compiled & positional `tag` slot confirmed

All three reviewers (Claude, Pi, Codex) returned **good-to-merge** with a single shared P2 nit — the `rust-client` couldn't be built locally (no JDK/openapi-generator), so the positional `tag` slot in the generated `job_api` calls was analysis-only. I've now closed that gap **without** needing Java.

**1. Verified against the authoritative published bindings.** The generated `windmill-api` crate is published to crates.io. Pulled `windmill-api@1.740.0` (matches `version.txt`) and read the real signatures:

```rust
run_script_by_path(config, workspace, path, request_body,
    scheduled_for, scheduled_in_secs, skip_preprocessor, parent_job,
    tag,            // <- 5th optional, Option<&str>
    cache_ttl, job_id, invisible_to_owner)

run_script_by_hash(config, workspace, hash, body,
    scheduled_for, scheduled_in_secs, skip_preprocessor, parent_job,
    tag,            // <- 5th optional, Option<&str>
    cache_ttl, job_id, include_header, invisible_to_owner)
```

`tag` is exactly the 5th optional (the slot I filled) in both, and the type `Option<&str>` matches the `tag: Option<&'a str>` I threaded through.

**2. Actually compiled it.** Built the updated `rust-client/src/client.rs` against the published `windmill-api 1.740.0` crate (which already carries the publish-time `from_str_patched`/rustls patches), both feature sets:

- `cargo build` → ✅ `Finished` (wrong arity/type would fail here)
- `cargo build --features async` → ✅ `Finished`

So the positional mapping is now confirmed by the compiler, not just by inspection. No code changes were required.

rubenfiszel:
### Fixed [P1] — preserved original PowerShell class-method arities (d11cde2)

Good catch. PowerShell class methods dispatch by exact argument count and have no default param values, so changing `RunScriptAsync` 4→5 and `RunFlowAsync` 3→4 in place dropped the old overloads — direct calls like `$client.RunScriptAsync($Path, $Hash, $Arguments, $ScheduledInSecs)` would fail with "Cannot find an overload".

Re-added the original arities as thin overloads that forward `$null` for `$Tag`:

```powershell
[PSCustomObject] RunScriptAsync([string] $Path, [string] $Hash, [Hashtable] $Arguments, [int] $ScheduledInSecs) {
    return $this.RunScriptAsync($Path, $Hash, $Arguments, $ScheduledInSecs, $null)
}
[string] RunFlowAsync([string] $Path, [Hashtable] $Arguments, [int] $ScheduledInSecs) {
    return $this.RunFlowAsync($Path, $Arguments, $ScheduledInSecs, $null)
}
```

Verified with a runtime overload-dispatch smoke test — both old and new arities resolve:

```
4-arg script: no-tag      5-arg script: tag=mytag
3-arg flow:   fno-tag      4-arg flow:   ftag=ft
```

(`$null` → `[string] $Tag` coerces to empty string, so `if ($Tag)` is false and no `tag` query param is added — identical to prior behavior.) Module still parses cleanly.

rubenfiszel:
### Fixed [P2] — `prompts.d.ts` no longer embeds (or drifts against) prompt contents (9871e4f)

You're right, and the root cause was that `prompts.d.ts` was a tracked declaration file with **string-literal types baked in**, but `generate.py` never regenerated it — it only writes `prompts.ts` and the hand-written `index.d.ts`. So every prompt change left `prompts.d.ts` stale, and `check-freshness.sh` couldn't catch it (it diffs the generated dir *after* running `generate.py`, which never touched that file).

Rather than hand-patch the literal SDK signatures (which would just drift again on the next change), I took your second option — stop tracking literal prompt contents there:

- Added `generate_ts_declarations()` to `generate.py`, which emits `prompts.d.ts` as plain `export declare const <NAME>: string;` lines (contents live only in `prompts.ts`).
- Wired it into `main()` next to the `prompts.ts` write.

Result: the 196 KB literal-typed file becomes a 36-line declaration file, export names verified 1:1 against `prompts.ts`, and it's now regenerated every run so `check-freshness` keeps it honest. Consumers only ever saw these as `string` (every helper in `index.ts` returns `string`), so widening the literal types is behavior-neutral.

`bash system_prompts/check-freshness.sh` → **"Auto-generated system prompts are up-to-date."** (idempotent).

PR #9708, opened 2026-06-22
feat: data tests for ducklake pipeline materialization

rubenfiszel:
## Summary

Implements **data tests** for the DuckLake pipeline/materialization system — the #1 dbt/dagster parity gap (`docs/pipelines-vs-dbt.md` §"Data tests"). Test annotations on a materialized SQL asset run against the freshly-materialized slice (post `DELETE`+`INSERT`, post-`COMMIT`) and **fail the pipeline run** on violation, propagating through the existing cascade error path.

Covers the dbt-equivalent built-ins plus an escape hatch:

```
// data_test unique <col>
// data_test not_null <col>
// data_test accepted_values <col> = a,b,c
// data_test relationships <col> -> datatable://other/asset.<col>   (referential integrity)
// data_test <script_path>                                          (custom test, escape hatch)
```

`// data_test` lines **accumulate**; malformed lines drop fail-safe. The keyword is `data_test`, **not `test`**, to stay clear of the unrelated shipped `// test:` CI-test annotation (`windmill_common::schema::parse_ci_test_annotation`) — mirroring dbt 1.8's own `tests:` → `data_tests:` rename.

### Extensibility

Data tests are designed as the **first extensible annotation family**, so column-lineage (a sibling effort) follows the same convention instead of a closed vocabulary. The pattern is *annotation → verifier*, in three seams: (1) parse a `data_test` line by **keyword head** into a typed `DataTest` variant; (2) compile each variant to a SQL **verifier probe** `(violation-count query, label) → raising statement`; (3) splice probes into the managed write after `COMMIT`, before the trailing summary read, so failure rides the existing error path with no new plumbing. Documented in `docs/ducklake-materialization.md` §"Data tests".

## Changes

**Parser (dual-parser parity — the key constraint)**
- `backend/.../asset_parser.rs`: new `DataTest` enum + `parse_data_test_spec`; `data_tests: Vec<DataTest>` on `PipelineAnnotations` / `ParseAssetsOutput`.
- `frontend/.../parsePipelineAnnotations.ts`: line-for-line TS mirror (`DataTest` type + `parseDataTestSpec`); `dataTests` on `PipelineAnnotations`.
- Shared fixture corpus + Rust & TS parity tests extended (`pipeline_annotations.json`, `pipeline_annotations_parity.rs`, `parsePipelineAnnotations.parity.test.ts`) — `dataTests` added to `ASSERTED_TS_FIELDS` so a new field can't be parsed on one side and silently skipped.

[...]

PR #9812, opened 2026-06-26
feat: capture managed-materialize output schema as asset metadata (#2a)

rubenfiszel:
## Summary

Closes pipeline-parity gap **#2a — schema capture** (the capture half of #2; **not** #2b enforcement, which depends on #5 column lineage and stays out of scope). After a managed `// materialize` DuckLake run, this captures the producer's output schema and persists it as versioned asset metadata — the producer-side foundation #2b will read back to validate consumer references at save time.

This is greenfield: #9709 shipped only the time-travel **read** UX and added no post-run capture seam. It reuses #9709's `DESCRIBE SELECT * FROM …` shape but builds its own capture path.

It does **not** re-implement #9708's value-level `data_test` checks (`not_null`/`accepted_values`/`unique`/`relationships`) — those already exist and are untouched.

**UI:** also adds a **Schema** tab on the ducklake asset node — a master-detail view of the captured schema versions (column/type per version), reading the new `GET /assets/asset_schemas` endpoint. It shows for any ducklake asset with a table (partitioned or whole-table alike, since the schema is table-level); the table-less catalog node renders no tab bar.

## Design decisions (the three design-open questions)

**Where schemas live → a dedicated `materialized_asset_schema` sidecar table** (keyed `workspace_id, asset_kind, asset_path, version`), *not* a column on `materialized_partition` and *not* the `asset` usage table.
- Schema is a property of the asset/table, not of a partition slice — a column on `materialized_partition` would duplicate the identical schema across every partition row (a 30-partition backfill writes it 30×).
- There is no per-asset "asset metadata row" to extend (assets are tracked only via the `asset` usage table).
- The sidecar's grain `(workspace, kind, path)` is exactly what #2b enforcement reads (`MAX(version)` = current contract).

**When captured → post-run, folded into the existing one-row materialize summary read (zero extra round-trips).** `materialize_result_sql` gains an `output_schema` column:
```sql
(SELECT list({'name': column_name, 'type': column_type})
 FROM (DESCRIBE SELECT * FROM <target>)) AS output_schema
```
The write has just committed, so the latest snapshot is exactly the recorded `snapshot_id` — no `AT (VERSION)` needed. Captured only for managed mode (we know the target alias+table; `manual` is skipped) and persisted only on a `Materialized` outcome (a failed run never advances the recorded schema).

**Versioning across re-materializations → monotonic `version` per asset; a new row only on a real schema change.** `record_asset_schema` compares the captured column set to the latest stored version: inserts `version = max+1` if it differs, otherwise re-affirms the latest row in place (`snapshot_id`/`job_id`/`captured_at`). Serialized per-asset with `pg_advisory_xact_lock` so concurrent same-asset captures can't both compute the same next version. Result: a compact schema-evolution history. A column **reorder** counts as a change (the captured list is ordered, and `SELECT *` consumers see order).

## Changes

- **Migration `20260626095840_add_materialized_asset_schema`** — new sidecar table (mirrors `materialized_partition`: `ASSET_KIND` enum, FK to `workspace`, no RLS) plus explicit `GRANT ALL TO windmill_user/windmill_admin` (the recurring app-written-table grant-gap pattern — `ALTER DEFAULT PRIVILEGES` is role-scoped).
- **`windmill-common/src/materialization.rs`** — `SchemaColumn`, `schema` field on `RecordMaterializationRequest` (`#[serde(default)]`, wire-compatible), `record_asset_schema` (advisory-locked versioned upsert), `list_asset_schemas` + `AssetSchemaVersion`.
- **`windmill-parser/src/sql_materialize.rs`** — `output_schema` column added to `materialize_result_sql` (a distinct codegen fn; no overlap with #5's annotation-parser edits).
- **`windmill-worker/src/duckdb_executor.rs`** — `extract_schema` (tolerant of nested-array / string-encoded FFI shapes) + threads the captured schema through `record_mat` → `record_asset_schema` on the successful path only; schema recording is its own best-effort transaction that can never fail the materialize run.
- **`frontend/.../AssetGraph/SchemaHistoryPanel.svelte` + `DucklakeAssetPanel.svelte`** — new **Schema** tab (master-detail schema-version history) on the asset node; raw-fetch `GET /assets/asset_schemas`, mirroring the sibling `PartitionStatusGrid`. Verified in a real browser (screenshots in PR comments).
- **`windmill-api-assets/src/lib.rs`** — `GET /assets/asset_schemas` read endpoint + the `record_materialization` handler now upserts the schema (in its own transaction, after committing the partition record, so a schema failure can't roll the partition record back).

## Notes / boundaries

- **Surfaces:** Surface B only. The one file shared with the parallel #5 (column lineage) work — `windmill-api-assets/src/lib.rs` — is touched only in new/distinct functions.
- **No OpenAPI/client change:** the sibling materialization endpoints (`/partitions`, `/record_materialization`) are consumed via raw `fetch` and are deliberately absent from `openapi.yaml`; `/asset_schemas` follows the same convention.
- **Known limitation:** the agent-http worker path (`agent_workers/.../record_materialization`, an EE-symlink handler) is left untouched, so agent-worker materializes won't persist schema in v1 — keeps this PR OSS-only. The `schema` request field is in place for the server-worker and SDK-polyglot paths.

## Test plan

- [x] `cargo check` — `windmill-common`, `windmill-parser`, `windmill-api-assets`, `windmill-worker --features duckdb`, and full `--features all_sqlx_features` (via sqlx prepare) all compile.
- [x] Parser unit test: `output_schema` fold present in both tests and no-tests summaries.
- [x] Worker unit test: `extract_schema` parses nested-array, string-encoded, and absent (`None`) shapes.
- [x] Integration tests (`windmill-common/tests/asset_schema_capture.rs`, real migrated DB): first-insert → v1, unchanged → re-affirm (no new version, snapshot advances), changed → v2 newest-first, column-reorder → new version.
- [x] Verified against bundled DuckDB 1.4 that `(SELECT list({...}) FROM (DESCRIBE SELECT * FROM t))` returns exactly `[{"name":..,"type":..},…]` (the shape `extract_schema` parses).
- [ ] **Manual e2e (requires a `duckdb`-featured worker + ducklake target — not runnable on the quickjs-only dev backend here):** run a managed `// materialize`, change the model's output columns, re-run, and confirm `GET /w/:id/assets/asset_schemas?path=<ducklake>/<table>` returns the history newest-first with `version` bumping only on real column changes.

---
## Summary by cubic
Captures the output schema of managed DuckLake `// materialize` runs and stores it as versioned asset metadata to support future contract enforcement (issue #2a). Adds a schema read API and a strategy-aware "Schema" UI tab; only managed, successful runs can advance schema history, and unknown producer strategy defaults to “evolvable” so history isn’t hidden.

- **New Features**
  - Persist schema in `materialized_asset_schema` keyed by workspace/kind/path + `version`; insert only when columns change (order counts); exclude `_wm_partition`; explicitly order columns.
  - Worker: capture schema only for managed runs (ignore `manual`); upsert only on `Materialized`; best-effort in its own transaction with a per-asset advisory lock; tolerant of nested/string JSON shapes.
  - API: `GET /assets/asset_schemas` lists versions newest-first; `POST /record_materialization` upserts schema only on `Materialized` and in a separate transaction; wire keeps `schema` optional for older clients.
  - Backend graph: expose `materialize_strategy` (`replace` | `append` | `merge`) on runnable nodes.
  - UI: new "Schema" tab on ducklake asset nodes; shows version history for whole-table `replace`, or a single fixed-schema view for `append`/`merge`/partitioned assets; fails open when producer strategy is unknown (treat as evolvable). Backed by `GET /assets/asset_schemas`.

- **Migration**
  - Adds `materialized_asset_schema` table with grants (run `20260626095840_add_materialized_asset_schema`).
  - No OpenAPI or client changes; nothing to update in SDKs.

<sup>Written for commit 6e5712894ca432da041ef401f0b6bac5bc317634. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9812?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-01732ca02b1888145c48c4e51e5b5829657224a743af9c0b2d5a140ad70e13dd.json
+++ b/backend/.sqlx/query-01732ca02b1888145c48c4e51e5b5829657224a743af9c0b2d5a140ad70e13dd.json
@@ -1,0 +1,33 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "UPDATE materialized_asset_schema\n                    SET snapshot_id = $5, job_id = $6, captured_at = now()\n                  WHERE workspace_id = $1 AND asset_kind = $2 AND asset_path = $3\n                    AND version = $4",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Text",
+        {
+          "Custom": {
+            "name": "asset_kind",
+            "kind": {
+              "Enum": [
+                "s3object",
+                "resource",
+                "variable",
+                "ducklake",
+                "datatable",
+                "volume"
+              ]
+            }
+          }
+        },
+        "Text",
+        "Int8",
+        "Int8",
+        "Uuid"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "01732ca02b1888145c48c4e51e5b5829657224a743af9c0b2d5a140ad70e13dd"
+}
--- a/backend/.sqlx/query-231475dc825518aa88f562698f8061d2562c3ac4af8d088f9eea9f0bc11d5fe7.json
+++ b/backend/.sqlx/query-231475dc825518aa88f562698f8061d2562c3ac4af8d088f9eea9f0bc11d5fe7.json
@@ -1,0 +1,44 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT version, columns AS \"columns: Json<Vec<SchemaColumn>>\"\n             FROM materialized_asset_schema\n            WHERE workspace_id = $1 AND asset_kind = $2 AND asset_path = $3\n            ORDER BY version DESC\n            LIMIT 1",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "version",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 1,
+        "name": "columns: Json<Vec<SchemaColumn>>",
+        "type_info": "Jsonb"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        {
+          "Custom": {
+            "name": "asset_kind",
+            "kind": {
+              "Enum": [
+                "s3object",
+                "resource",
+                "variable",
+                "ducklake",
+                "datatable",
+                "volume"
+              ]
+            }
+          }
+        },
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "231475dc825518aa88f562698f8061d2562c3ac4af8d088f9eea9f0bc11d5fe7"
+}
--- a/backend/.sqlx/query-2a28f13c1f06a77bc9a164393cccf0d8d2a3d7e8552d8ad09512d2d6bf27c3fb.json
+++ b/backend/.sqlx/query-2a28f13c1f06a77bc9a164393cccf0d8d2a3d7e8552d8ad09512d2d6bf27c3fb.json
@@ -1,0 +1,62 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT version, columns AS \"columns: Json<Vec<SchemaColumn>>\",\n                  snapshot_id, job_id, captured_at\n             FROM materialized_asset_schema\n            WHERE workspace_id = $1 AND asset_kind = $2 AND asset_path = $3\n            ORDER BY version DESC",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "version",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 1,
+        "name": "columns: Json<Vec<SchemaColumn>>",
+        "type_info": "Jsonb"
+      },
+      {
+        "ordinal": 2,
+        "name": "snapshot_id",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 3,
+        "name": "job_id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 4,
+        "name": "captured_at",
+        "type_info": "Timestamptz"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        {
+          "Custom": {
+            "name": "asset_kind",
+            "kind": {
+              "Enum": [
+                "s3object",
+                "resource",
+                "variable",
+                "ducklake",
+                "datatable",
+                "volume"
+              ]
+            }
+          }
+        },
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      true,
+      true,
+      false
+    ]
+  },
+  "hash": "2a28f13c1f06a77bc9a164393cccf0d8d2a3d7e8552d8ad09512d2d6bf27c3fb"
+}
--- a/backend/.sqlx/query-8e829a7358ea47100c99e78067266eb7a83e312b88f0244d1c517b03a9bcdea0.json
+++ b/backend/.sqlx/query-8e829a7358ea47100c99e78067266eb7a83e312b88f0244d1c517b03a9bcdea0.json
@@ -1,0 +1,34 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "INSERT INTO materialized_asset_schema\n           (workspace_id, asset_kind, asset_path, version, columns,\n            snapshot_id, job_id, captured_at)\n         VALUES ($1, $2, $3, $4, $5, $6, $7, now())",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Varchar",
+        {
+          "Custom": {
+            "name": "asset_kind",
+            "kind": {
+              "Enum": [
+                "s3object",
+                "resource",
+                "variable",
+                "ducklake",
+                "datatable",
+                "volume"
+              ]
+            }
+          }
+        },
+        "Varchar",
+        "Int8",
+        "Jsonb",
+        "Int8",
+        "Uuid"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "8e829a7358ea47100c99e78067266eb7a83e312b88f0244d1c517b03a9bcdea0"
+}
--- a/backend/.sqlx/query-b876b16b660bd1c3799abadc2c16d0f58f49a45647525658acb909a494877238.json
+++ b/backend/.sqlx/query-b876b16b660bd1c3799abadc2c16d0f58f49a45647525658acb909a494877238.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT pg_advisory_xact_lock(hashtextextended($1, 0::int8))",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "pg_advisory_xact_lock",
+        "type_info": "Void"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "b876b16b660bd1c3799abadc2c16d0f58f49a45647525658acb909a494877238"
+}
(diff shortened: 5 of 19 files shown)


rubenfiszel:
## ✅ True DuckLake end-to-end run (manual e2e box now checked)

Built the full-featured binary (`duckdb,parquet,enterprise,private,license`) and ran it as a server+worker against a real DuckLake: **Postgres catalog + MinIO S3 storage** (via the windmill S3 proxy), driving managed `// materialize` jobs through the complete path — codegen `output_schema` DESCRIBE fold → DuckLake write to MinIO → summary read → worker `extract_schema` → versioned `record_asset_schema` → `GET /assets/asset_schemas`.

Script (`ducklake://analytics/orders_daily`):

| Run | Output columns | Outcome |
|---|---|---|
| 1 | `order_id BIGINT, status VARCHAR, amount DOUBLE` | **v1** captured, snapshot 1 |
| 2 | identical schema, different rows | **re-affirmed v1** in place — no new version, `snapshot_id` advanced 1→2 |
| 3 | dropped `amount`, added `qty INTEGER` | **v2** created, snapshot 3 |

Captured summary row (run 1):
```json
[{"materialized":"ducklake://analytics/orders_daily","rows":1,"snapshot_id":1,
  "output_schema":[{"name":"order_id","type":"BIGINT"},{"name":"status","type":"VARCHAR"},{"name":"amount","type":"DOUBLE"}]}]
```

`GET /w/admins/assets/asset_schemas?path=analytics/orders_daily` after all three runs — evolution history newest-first:
```json
[
  {"version":2,"snapshot_id":3,"columns":[{"name":"order_id","type":"BIGINT"},{"name":"status","type":"VARCHAR"},{"name":"qty","type":"INTEGER"}]},
  {"version":1,"snapshot_id":2,"columns":[{"name":"order_id","type":"BIGINT"},{"name":"status","type":"VARCHAR"},{"name":"amount","type":"DOUBLE"}]}
]
```

This confirms the three designed behaviors against a real substrate: schema captured post-run, an unchanged re-materialize re-affirms in place (snapshot advances, no version bump), and a real column change bumps the version. The throwaway instance, catalog DB, and MinIO data were torn down afterward.

rubenfiszel:
Addressed the CI review in 996b6d9:

- **[P1, claude + cubic] `_wm_partition` leaked into the captured schema for partitioned assets.** Fixed — the schema-capture `DESCRIBE` now filters out the partition column (`WHERE column_name <> '<partition_col>'`, emitted only when partitioned), so the recorded contract is the producer's logical output, not Windmill's storage detail. This matters because it's the grain #2b reads back. Added a `materialize_result_sql_schema_excludes_partition_column` unit test and verified the exact SQL against bundled DuckDB 1.4 (a `_wm_partition` column is dropped from the result).

- **[P2, cubic] `list(...)` had no explicit ordering → could reorder columns and spuriously bump the schema version.** Fixed — the columns now carry a `row_number() OVER ()` ordinal from the `DESCRIBE` and the fold is `list({...} ORDER BY _wm_ord)`, so column order is deterministic and an unchanged re-materialize compares equal.

- **[P2, cubic] API `record_materialization` schema upsert wasn't gated on success.** Fixed — the handler now only upserts the schema when `status == Materialized`, so a failed/running write (or a client attaching a schema to one) can't advance `materialized_asset_schema` history. (The worker path already only sends a schema on the successful path; this is the matching defensive gate on the SDK/agent entry point.)

**`cargo_test` failure** is `debounce_setting_applied_to_dispatched_subscriber` (`asset_trigger_dispatch.rs`) — a known pre-existing flake unrelated to this PR (it touches no dispatch/debounce code); `test-linux`, `check_oss`, `check_oss_full`, `pi-review`, and `cubic` all pass. Will re-run.

rubenfiszel:
Addressed the Codex review in 356aff3 (pi-review was "Good to merge"):

- **[P1] auth contract on the new public helpers.** Documented it explicitly on `record_asset_schema` and `list_asset_schemas`: they perform no access control of their own (mirroring the `materialized_partition` siblings) and require the caller to pass a workspace-authorized executor — an RLS-scoped `user_db` transaction on the API/agent entry points, or the trusted worker DB pool for the in-worker recorder. The existing callers already satisfy this; the contract is now explicit rather than implicit.

- **[P2] manual mode could persist a caller-shaped `output_schema`.** Gated schema extraction on the managed (`Some((Some(_), _))`) path — in `// materialize manual` the result is the user's own query output (we generate no summary), so its `output_schema` is now never trusted. This mirrors the adjacent managed-only `collection_strategy` gate.

**Verified end-to-end** against a real DuckLake (Postgres catalog + MinIO), exactly the scenario Codex flagged as missing coverage:

| Materialize | Result | `materialized_asset_schema` |
|---|---|---|
| managed `analytics/orders_daily` | summary `output_schema:[order_id,status]` | **v1 written** ✓ |
| manual `analytics/orders_manual`, user result `[{"output_schema":[{"name":"injected","type":"EVIL"}]}]` | user's own row | **empty — not captured** ✓ |

Manual mode still recorded its partition row (`orders_manual → materialized`), confirming only the schema capture is gated, not manual materialization itself. Throwaway instance and config torn down afterward.

rubenfiszel:
## Frontend: schema-history tab on the asset node

Added a **Schema** tab to the ducklake asset node panel (`DucklakeAssetPanel`) that surfaces the captured output-schema versions — a master-detail view mirroring the existing History tab. The version list (newest first, newest auto-selected) shows column count + snapshot + capture time; selecting a version renders its column/type table. It reads `GET /assets/asset_schemas` via raw fetch (same convention as the sibling `PartitionStatusGrid`; these materialization endpoints aren't in the generated client).

Verified in a real browser against a live DuckLake (Postgres catalog + local-filesystem storage), driving the deployed `f/pipeline/orders_daily` materialize and evolving its schema across two runs.

**v2 selected** (current — order_id, status, qty, updated_at):
schema tab v2

**v1 selected** (the pre-change schema — order_id, status, amount):
schema tab v1

Validation: `svelte-autofixer` clean, `check:fast` reports no errors on the new/changed files (the `Timeout`-vs-`number` diagnostics are pre-existing global noise, unrelated to these files).

rubenfiszel:
## Schema tab is now strategy-aware

Per review discussion: the schema-*evolution* history only makes sense for a whole-table `replace` producer (`CREATE OR REPLACE` rebuilds the table, so columns can change run-to-run). `append`/`merge`/partitioned writes `INSERT` into a fixed-schema table, so their schema is pinned at first materialize — a multi-version "history" there is degenerate.

The backend now surfaces the managed `materialize_strategy` on the asset-graph runnable node (same parse-from-annotation path as `partition_kind`), and the pipeline page derives `schemaCanEvolve` for the selected asset from its write-producer. The Schema tab switches accordingly:

**Fixed schema** — `append` producer (`events_log`): single current-schema table + a "schema is fixed" note, no version list:
fixed schema

**Evolving schema** — whole-table `replace` producer (`orders_daily`): the master-detail version history (v2 current / v1):
schema history

The gate is *can the schema evolve* (replace && not partitioned), not partitioned-vs-not — an unpartitioned `append` is correctly treated as fixed. Unknown producers default to evolvable so real history is never hidden. Verified both views in a real browser.

rubenfiszel:
Addressed the Codex P2 in 6e57128.

`schemaCanEvolve` previously only defaulted to evolvable when there were *zero* producers — a producer present but missing `materialize_strategy` (e.g. a draft-overlay runnable, which the graph synthesizes without the field) fell through to `canEvolve=false`, hiding captured history behind the fixed-schema view.

Fixed by failing open: the fixed view now shows only when **every** producer is a *known* insert-style write (`append`/`merge`, or partitioned `replace`); any producer with missing/unknown strategy is treated as evolvable, so real history is never hidden. I went with the "missing strategy → evolvable" option (vs. populating the field in draft overlays) since it matches the stated intent and keeps the gate's failure mode safe.

PR #9813, opened 2026-06-26
fix(flows): reject corrupt step paths at deploy + atomic cache writes (#9751)

### LINKED ISSUE #9751: bug:Flow step runs the wrong script at runtime
### Describe the bug

In a self-hosted Windmill instance, individual **flow steps silently execute the wrong
script at runtime**. The step's `path` (and the deployed flow definition) correctly points
at script **A**, the deployed script at that path **is** A, but when the flow runs, the step
spawns a sub-job whose `script_path` is an **unrelated script B** and runs B's code.

This is not a flow-definition problem — the stored definition is correct at every layer we
can inspect. It appears to be a stale/corrupted **server-side path → script resolution
cache**. It is especially dangerous because, in our case, the mis-resolved target was a
**destructive** script that ran in place of a read-only validation step.

We observed **two different paths** affected in the same flow, both mis-resolving into
scripts that live in a *different* folder (`f/ops/scripts/clean_device/*`):

| Flow step `path` (correct)                                  | Script actually executed at runtime         |
| ----------------------------------------------------------- | ------------------------------------------- |
| `f/ops/scripts/clean_device/pre_clean` (read-only preview)  | `f/ops/scripts/clean_device/clean_device` (**destructive wipe**) |
| `f/ops/scripts/customer_management/validate_onboard_target` | `f/ops/scripts/clean_device/approval`       |

### To reproduce

We could not produce a clean from-scratch repro (the trigger appears to be accumulated
server state), but the conditions under which it arose:

1. Self-hosted instance with **two workspaces (`dev` and `prod`) on the same instance**.
2. Scripts under one folder were created/edited, and a flow was **derived/copied** from
   another flow that referenced scripts in a neighbouring folder
   (here: an "offboard" flow using `clean_device/*` scripts was used as the basis for an
   "onboard" flow using `customer_management/*` scripts).
3. Scripts and flows were deployed via `wmill sync push --workspace <ws>` repeatedly.
4. Run the flow. One or more steps execute the **wrong** script even though every stored
   definition is correct.
5. Inspect with the commands above: `wmill flow get …` and `wmill script get …` look
   correct, but `wmill job get <sub-job> --json | jq .script_path` reveals a different
   script was executed.
**1. Get the flow run and its step → sub-job mapping.** The flow ran the correct module
order; note the `validate_onboard_target` step and its sub-job id `019ef7fd-755b-…`:

```
$ wmill job get 019ef7fd-22bc-8db1-faa7-7fc33426bc31 --workspace prod
ID: 019ef7fd-22bc-8db1-faa7-7fc33426bc31
Status: failure
Kind: flow
Script Path: f/ops/flows/customer_onboard_machine
Steps:
  ✓ derive_vpn_ip: Derive VPN IP from device serial       019ef7fd-24cf-e712-9deb-373aa1e2efc4
  ✓ validate_device: Verify device exists in inventory    019ef7fd-2be8-121e-245d-2f6c82289575
  ✓ validate_onboard_target: Validate device is available 019ef7fd-755b-07e2-9331-e09ee62ea478
  ✓ pre_clean: Verify no cleanup targets exist            019ef7fd-90c8-ce46-6782-4940706f5789
  ✗ resolve_customer_vlan: Resolve customer VLAN          019ef7fd-d0ab-fb99-420c-b73575971706
  ...
```

**2. What the step *should* run.** The deployed flow module `validate_onboard_target` points
at `…/customer_management/validate_onboard_target`, with no inline lock:

```
$ wmill flow get f/ops/flows/customer_onboard_machine --workspace prod --json \
    | jq '.value.modules[] | select(.id=="validate_onboard_target") | {id, path: .value.path, has_lock: (.value.lock != null)}'
{
  "id": "validate_onboard_target",
  "path": "f/ops/scripts/customer_management/validate_onboard_target",
  "has_lock": false
}
```

…and the deployed script at that path is the correct one (right `def main`, valid hash, not
archived/deleted/draft):

```
$ wmill script get f/ops/scripts/customer_management/validate_onboard_target --workspace prod --json \
    | jq '{hash, archived, deleted, draft_only}'
{ "hash": "e12cfacf9ff7d09b", "archived": false, "deleted": false, "draft_only": null }

$ wmill script get f/ops/scripts/customer_management/validate_onboard_target --workspace prod --json \
    | jq -r '.content' | grep -n '^def main'
30:def main(device_serial: str, customer_name: str, customer_email: str | None = None, org_id: str | None = None, debug: bool = False) -> ValidatedOnboardCustomerResult:
```

**3. What the step *actually* ran.** The sub-job for that step reports a completely different
`script_path` (`clean_device/approval`) and was invoked with that other script's arguments
(`review`, not `device_serial`/`customer_name`):

```
$ wmill job get 019ef7fd-755b-07e2-9331-e09ee62ea478 --workspace prod --json | jq '.script_path'
"f/ops/scripts/clean_device/approval"

$ wmill job get 019ef7fd-755b-07e2-9331-e09ee62ea478 --workspace prod --json | jq '{script_path, args}'
{
  "script_path": "f/ops/scripts/clean_device/approval",
  "args": {
    "review": {
      "hostname": "<redacted>",
      "preview_totals": { "total": 0, ... },
      "reviewed_state": { ... },
      "preview_details": { ... }
    }
  }
}
```

Its result is the *approval* script's output (a Teams approval/suspend payload with
`reviewed_state` and a `webhook_url`) — there is no `customer_name`, which is what the
declared `validate_onboard_target` script returns:

```
$ wmill job get 019ef7fd-755b-07e2-9331-e09ee62ea478 --workspace prod --json | jq '.result | keys'
[ "description", "reviewed_state", "teams_notification", "webhook_url" ]
# webhook_url: "<redacted signed URL>"
```

**4. Contrast: a sibling step in the *same run* resolved correctly.** The `pre_clean` step
(after we force-redeployed it) ran the right script with the right args, proving the
mis-resolution is per-path, not a whole-flow failure:

```
$ wmill job get 019ef7fd-90c8-ce46-6782-4940706f5789 --workspace prod --json | jq '{script_path, args}'
{
  "script_path": "f/ops/scripts/clean_device/pre_clean",
  "args": { "hostname": "<redacted>" }
}
```

**5. The downstream error is misleading.** Because the mis-resolved step returned the wrong
shape (no `customer_name`), the *next* step failed with an unrelated-looking error:

```
$ wmill job get 019ef7fd-22bc-8db1-faa7-7fc33426bc31 --workspace prod --json | jq '.result'
{
  "status": "failure",
  "message": "'NoneType' object has no attribute 'strip'",
  "step_id": "resolve_customer_vlan",
  "error_name": "AttributeError"
}
```

### Expected behavior

A flow step must execute the script at its declared `path` (resolved to the current deployed
head at that path). It must never execute a different script's body — and most importantly,
must never silently run a **destructive** script in place of a read-only one. If a path
cannot be resolved, the step should hard-fail with a clear "script not found" error rather
than resolve to unrelated code.

### Screenshots

Clearly showing step for script `f/ops/scripts/clean_device/pre_clean`

Image

Going to Node status you can see run id ends in `f5bf`

Image

Then when you go to that run you see that it runs a totally different script `f/ops/scripts/clean_device/clean_device`

Image

### Browser information

Firefox 151.0.4

### Application version

Windmill CE v1.708.0, ghcr.io/windmill-labs/windmill:1.708.0(workers)

### Additional Context

The mis-resolved targets were all in a single neighbouring folder
  (`f/ops/scripts/clean_device/*`), suggesting the stale mapping keys on something that
  collides across paths or survives a move/rename/copy.

rubenfiszel:
## Summary

Fixes #9751, where a flow step could execute an unrelated (and in the reported case, destructive) script at runtime even though every stored flow/script definition looked correct. A forensic dump of an affected instance traced it to two independent issues:

1. **Corrupt definitions accepted at deploy.** `wmill sync push` run from a feature-branch checkout under `/tmp` baked an absolute local path (`/tmp/tmp.XXX/.../ops/scripts/clean_device/...`) into a step's `value.path`. The backend persisted it verbatim, and at runtime it mis-resolved to an unrelated script.
2. **Non-atomic on-disk cache writes.** `FsBackedCache::put` wrote with `write+create` (no truncate, no atomic rename), so a shorter overwrite left stale trailing bytes and concurrent writers could interleave into a torn file — a corrupt cached blob that a worker then scheduled from.

This is the tight, high-confidence fix: it closes the proven root cause (the bad deploy) and the proven on-disk hazard. The runtime cache divergence the dump showed is a downstream consequence of those two; a broader "content-address the cache by value hash" robustness change was explored and deliberately deferred (it needs a table-rewriting migration + a GC subsystem, and the stale-cache-after-an-in-place-relock case it additionally covers is pre-existing and lower-severity, not the reported bug).

## Changes

- **Reject non-workspace step paths** (`u/`, `f/`, `g/`, `hub/`, or empty) in `validate_flow_value` — the serde hook on `NewFlow.value`, covering both `create_flow` and `update_flow`, recursively through loops/branches/AI-agent tools — and early in the CLI `pushFlow`.
- **Make `FsBackedCache::put` atomic** — write to a unique temp file (`.tmp.{pid}.{seq}`, safe on a shared cache volume) with `truncate` + `sync_all`, then atomic rename over the target, cleaning up the temp on error.

## Test plan

- [ ] `cd backend && cargo test -p windmill-types --lib flows::` — path-allowlist accept/reject (incl. absolute path in a nested module)
- [ ] `cargo test -p windmill-common --lib cache::` — shorter overwrite leaves no stale tail, no leftover temp files
- [ ] Manual: push a flow (UI and `wmill flow push`) with an absolute step path → rejected with a "not a workspace path" error naming the step

### DIFF
--- a/backend/windmill-common/src/cache.rs
+++ b/backend/windmill-common/src/cache.rs
@@ -1189,21 +1189,12 @@ const _: () = {
             use std::fs::OpenOptions;
             use std::io::Write;
 
-            // Write to a unique sibling temp file then atomically rename over the target.
-            // A plain `write+create` (no truncate) leaves stale trailing bytes when the
-            // new value is shorter than the old, and concurrent writers interleave into a
-            // torn file — both yield a corrupt cache entry that imports as wrong content.
-            // rename(2) within the same directory is atomic, so a reader sees either the
-            // old or the new file whole, never a partial write. The temp name carries the
-            // pid + a counter so two processes sharing a cache volume can't clobber each
-            // other's in-flight write.
-            static PUT_SEQ: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+            // Atomic write: truncate+write a uniquely-named temp file (UUID, not pid — pids
+            // collide across container PID namespaces on a shared cache volume), fsync, then
+            // rename(2) over the target. Without this a shorter overwrite leaves a stale tail
+            // and concurrent writers tear the file — a corrupt entry a reader would import.
             let final_path = item.path(self);
-            let tmp_path = final_path.with_extension(format!(
-                "tmp.{}.{}",
-                std::process::id(),
-                PUT_SEQ.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
-            ));
+            let tmp_path = final_path.with_extension(format!("tmp.{}", Uuid::new_v4()));
             OpenOptions::new()
                 .write(true)
                 .create(true)
--- a/backend/windmill-common/src/cache.rs
+++ b/backend/windmill-common/src/cache.rs
@@ -1189,11 +1189,34 @@ const _: () = {
             use std::fs::OpenOptions;
             use std::io::Write;
 
+            // Write to a unique sibling temp file then atomically rename over the target.
+            // A plain `write+create` (no truncate) leaves stale trailing bytes when the
+            // new value is shorter than the old, and concurrent writers interleave into a
+            // torn file — both yield a corrupt cache entry that imports as wrong content.
+            // rename(2) within the same directory is atomic, so a reader sees either the
+            // old or the new file whole, never a partial write. The temp name carries the
+            // pid + a counter so two processes sharing a cache volume can't clobber each
+            // other's in-flight write.
+            static PUT_SEQ: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+            let final_path = item.path(self);
+            let tmp_path = final_path.with_extension(format!(
+                "tmp.{}.{}",
+                std::process::id(),
+                PUT_SEQ.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+            ));
             OpenOptions::new()
                 .write(true)
                 .create(true)
-                .open(item.path(self))
-                .and_then(|mut file| file.write_all(data.as_ref()))
+                .truncate(true)
+                .open(&tmp_path)
+                .and_then(|mut file| {
+                    file.write_all(data.as_ref())?;
+                    file.sync_all()
+                })
+                .and_then(|()| std::fs::rename(&tmp_path, &final_path))
+                .inspect_err(|_| {
+                    let _ = std::fs::remove_file(&tmp_path);
+                })
         }
     }
 
@@ -1236,6 +1259,33 @@ const _: () = {
 mod tests {
     use super::*;
     use serde_json::json;
+
+    #[test]
+    fn fs_cache_put_overwrites_without_stale_tail() {
+        // Regression for the non-truncating, non-atomic `put`: overwriting a value with a
+        // shorter one must not leave stale trailing bytes (which imported as corrupt/wrong
+        // content — the #9751 worker-cache hazard).
+        let dir = tempfile::tempdir().unwrap();
+        let root = dir.path();
+
+        root.put("k", b"a-long-cached-value-0123456789").unwrap();
+        assert_eq!(root.get("k").unwrap(), b"a-long-cached-value-0123456789");
+
+        root.put("k", b"short").unwrap();
+        assert_eq!(
+            root.get("k").unwrap(),
+            b"short",
+            "shorter overwrite must fully replace, no stale tail"
+        );
+
+        // No temp files left behind after a successful write.
+        let leftover: Vec<_> = std::fs::read_dir(root)
+            .unwrap()
+            .filter_map(|e| e.ok())
+            .filter(|e| e.file_name().to_string_lossy().contains(".tmp."))
+            .collect();
+        assert!(leftover.is_empty(), "temp files must be renamed/cleaned up");
+    }
 
     #[test]
     fn flow_data_extras_preserves_notes_and_groups() {
--- a/backend/windmill-types/src/flows.rs
+++ b/backend/windmill-types/src/flows.rs
@@ -155,6 +155,19 @@ fn validate_retry(retry: &Retry, module_id: &str) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Script/sub-flow step references must be workspace paths (`u/`, `f/`, `g/`) or a hub
+/// reference (`hub/`). Empty is tolerated for intermediate/incomplete steps. This blocks
+/// absolute or local filesystem paths (e.g. `/tmp/.../ops/scripts/...` baked in by a
+/// `wmill sync push` from a feature-branch checkout) from being persisted into a flow,
+/// where they silently mis-resolve to an unrelated script at runtime (#9751).
+fn is_workspace_runnable_path(path: &str) -> bool {
+    path.is_empty()
+        || path.starts_with("u/")
+        || path.starts_with("f/")
+        || path.starts_with("g/")
+        || path.starts_with("hub/")
+}
+
 fn validate_flow_value<'de, D>(deserializer: D) -> Result<Box<RawValue>, D::Error>
 where
     D: Deserializer<'de>,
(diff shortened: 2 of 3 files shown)


rubenfiszel:
Addressed the review feedback in b1b687793f:

- **Backend `validate_flow_value` skipped `failure_module` / `preprocessor_module`** (identified by claude[bot]) — the backend is the authoritative guard, so this was the weaker side. Extracted the per-step check into a reusable closure and now traverse the failure and preprocessor modules too (recursively, since they can be sub-flows/loops/branches). Added `flow_rejects_absolute_path_in_failure_and_preprocessor_modules`.
- **CLI preflight only checked `type: "script"` paths, not sub-flow `type: "flow"` paths** (identified by cubic and claude[bot]) — replaced `collectPathScriptPaths` in the validation with a small `collectStepPaths` walker that also collects sub-flow step paths (recursively, incl. failure/preprocessor), so the CLI's early-error now matches both the comment and the backend rule.

Backend `flows::tests` (6) and CLI `deno check` pass; bin compiles offline.

rubenfiszel:
Addressed the Codex review nit in b46e0cfdf8: `collectStepPaths` now also walks AI-agent tools, so a non-workspace path in a tool is reported by the CLI fail-fast (the backend already validated these via `traverse_modules`). `deno check` passes.

rubenfiszel:
Thanks — addressed both:

- **AI-agent tool paths in the CLI** (Pi) — already fixed in b46e0cfdf8 (that review ran on the prior commit): `collectStepPaths` now walks `v.tools` and recurses into nested tool structures, matching the backend's `traverse_modules`.
- **`failure_module`/`preprocessor_module` test using `slot:` as a json! key** (Pi + cubic) — small correction: `json!` *does* interpolate an identifier in key position to the variable's value (verified: `json!({slot: 1})` with `slot = "failure_module"` ⇒ `{"failure_module":1}`), so the test was actually exercising the validation (it would have panicked on `unwrap_err()` otherwise). That said, the behavior is subtle, so I rewrote it to set the key explicitly via `serde_json::Map` to remove the ambiguity. Still passes.

rubenfiszel:
Addressed the remaining findings in the latest commit:

- **[P1] Cache temp-file collision on a shared volume** (Codex) — `pid` collides across container PID namespaces and `PUT_SEQ` resets per process, so two workers could clobber the same temp file before rename. Switched to a random `Uuid::new_v4()` suffix (the same approach as `worker.rs`'s atomic-write helpers) — globally unique, closing the cross-process hazard. A true multi-process collision test isn't unit-testable here; the UUID makes the collision structurally impossible.
- **[P2] `put` comment exceeded the AGENTS.md ≤4-line limit** (Pi) — trimmed to 4 lines.

The earlier nits were already resolved in prior commits: the failure/preprocessor test now builds the key via `serde_json::Map` (336dba0; for the record, `json!` *does* interpolate ident keys — verified `json!({slot:1})` with `slot="failure_module"` ⇒ `{"failure_module":1}` — so it was passing, but the explicit form removes the ambiguity), and the CLI walker covers sub-flow + AI-agent tool paths (b1b6877 / b46e0cf).

rubenfiszel:
This was already addressed in `04ff43011e` (the review ran on an earlier commit). The temp name is now:

```rust
let tmp_path = final_path.with_extension(format!("tmp.{}", Uuid::new_v4()));
```

— no `pid`, no `PUT_SEQ`. `Uuid::new_v4()` is collision-resistant across processes/containers (the same approach `worker.rs`'s atomic-write helpers use), which is the first remediation you suggested.

That also closes the second half of the concern: with unique temp names there's no "loser's rename failure" from two processes targeting the same temp file — each renames its own temp over `self.json` and both `put`s succeed, so the `export` never returns `Err` from a collision and the `remove_dir_all(path)` cleanup in `get_or_insert_async` can't be triggered by it.

PR #9814, opened 2026-06-26
feat: column-level lineage for DuckLake pipelines (SQL-AST inferred + traceable)

rubenfiszel:
## Summary

Implements column-level lineage for DuckLake data pipelines — parity roadmap gap #5 (`docs/pipelines-vs-dbt.md` §3) — and surfaces it in the asset graph.

**Inferred-first.** For DuckDB scripts the lineage is **derived automatically from the SQL AST** (dbt-style), so no annotation is needed in the common case. `windmill-parser-sql-asset` walks each output-producing query's projection and maps every output column to the source columns its expression reads — passthroughs *and* computed columns (`amount + tax AS total`) — resolving each input to its asset via the same `ATTACH`/alias machinery the asset parser already uses.

The `// column <out_col> <- <asset-uri>.<col>[, …]` annotation is the **override / escape hatch** for what inference can't reach: polyglot transforms (Python/TS/Bash — no SQL AST), dynamic SQL (`${sql.raw(...)}`), or correcting a mis-inferred edge. Inferred and annotated lineage merge per output column with the annotation winning (`merge_column_lineage`). The annotation is the second *extensible* annotation family after `// data_test` (#9708) and follows that exact convention (head-then-tail parse, snake_case serde, full Rust↔TS parser parity via the shared fixture corpus). Lineage is pure metadata — it drives the graph view, never a runtime probe.

Scope: column lineage + its graph surface only — no execution, no schema contracts.

## Changes

**Backend — SQL AST inference (the automatic path)**
- `windmill-parser-sql-asset`: `infer_column_lineage` walks CTAS / `CREATE VIEW` / top-level (managed-materialize) SELECT projections; a `ColumnIdentCollector` (sqlparser `Visit`) gathers each output expression's identifier leaves, resolved to source-asset columns via a refactored `build_from_maps` (shared with the existing read-column extraction). `parse_assets` merges inferred lineage with `// column` annotations. 4 new tests (computed, passthrough, annotation-override, wildcard).
- `windmill-parser/asset_parser.rs`: `merge_column_lineage(inferred, annotated)` (annotation wins per output column) + test.
- `windmill-api-assets`: the graph endpoint runs `parse_assets` for DuckDB pipeline members and surfaces the merged lineage on deployed nodes (other languages fall back to annotation-only). Adds the `windmill-parser-sql-asset` dep + `language` to the member query.

**Backend — `// column` annotation (the override, Surface A: both parsers + parity corpus)**
- `asset_parser.rs`: `ColumnLineage`/`ColumnRef` (snake_case serde) on `ParseAssetsOutput`/`PipelineAnnotations`, `// column` parsing + helpers, unit tests.
- Shared parity corpus: `pipeline_annotations.json` + `pipeline_annotations_parity.rs`; `windmill-common/assets.rs` re-exports.

**Frontend (parser parity + graph surface)**
- `parsePipelineAnnotations.ts` (+ `.parity.test.ts`): `ColumnLineage`/`ColumnRef` types + parsing, behaviorally identical to Rust, asserted against the same corpus.
- `types.ts` / `resolveGraph.ts`: `column_lineage` on the runnable node, refreshed live from the draft parse.
- `AssetGraphCanvas.svelte` + `AssetGraphEdge.svelte`: a "columns ×N" badge on the producer→materialized-asset write-edge (hover lists each mapping).
- `ColumnLineageView.svelte` (new): a column-to-column diagram in the asset details pane, fed by `selectionColumnLineage` derived in the pipeline `+page.svelte`.

## Screenshots

**Auto-derived from SQL — no annotation.** Script body is `SELECT id, amount + tax AS order_total, customer_id AS cust FROM orders`; the diagram is inferred entirely from the AST (note `amount`+`tax` → `order_total`):

column-lineage-inferred

Deployed pipeline graph — the "columns ×3" badge on the write-edge (hover lists every mapping):

graph-deployed

Column-lineage diagram (with explicit `// column` annotations, the override path):

column-lineage-view

## Test plan
- [x] `cargo test -p windmill-parser -p windmill-parser-sql-asset` — 105 + 58 green (incl. SQL inference, annotation parsing, merge, parity)
- [x] Frontend parity + full AssetGraph vitest (115) green; `npm run check` 0 new errors
- [x] Verified live on a deployed pipeline: lineage **auto-derived from SQL with no annotation** (computed column included); `// column` annotation overrides per output column; duplicate-input ref doesn't crash the diagram
- [ ] Reviewer (note, not a blocker): SQL-AST inference runs **server-side** (deploy + graph endpoint), so it shows on deployed members. Unsaved drafts show `// column` annotations only — the live preview uses a pinned prebuilt WASM parser that predates this feature; wiring it in is a tracked follow-up. Deployed pipelines are where column lineage matters.


rubenfiszel:
**End-to-end verified with the published `windmill-parser-wasm-asset@1.740.0`** (dev server serving `v1.740.0-4-g5ea7717163`):

- The published WASM `parse_assets_sql` emits correct inferred `column_lineage` directly (Node-tested), incl. the computed `order_total ← amount, tax`.
- Deployed pipeline graph renders the inferred column-lineage diagram (no annotations):

lineage view

- The live editor opens cleanly with the new wiring (ScriptEditor `inferredColumnLineage` bindable → pane → page → `resolveGraph` merge); `inferAssets` runs and the write-edge badge is present:

edit open

Note: the Playwright MCP was down this session, so these were captured via a headless Playwright script driving system Chromium.

rubenfiszel:
**Transitive column-lineage trace (impact analysis)** — turns the static diagram into a tool.

New `columnLineageGraph.ts` stitches every producer's `column_lineage` into a pipeline-wide column graph; `ColumnLineageTrace.svelte` renders the connected neighborhood of a selected asset's columns (laid out by hop depth) and lets you click any column to highlight its full transitive impact set — forward *and* backward — across scripts.

Multi-hop chain in the demo: `orders.amount + tax → orders_daily.order_total → revenue.total_revenue`.

Select the sink asset `warehouse/revenue` → its columns' full upstream lineage across both scripts:

full trace

Click `orders.amount` → its complete impact path lights up (`amount → order_total → total_revenue`), everything else dims. Note `tax` (a co-contributor to `order_total`, not in `amount`'s lineage) is correctly dimmed:

click to trace

Tests: `columnLineageGraph.test.ts` covers multi-hop stitching, bidirectional trace, and depth layout (124 AssetGraph tests green). Verified via headless Playwright against the running app.

rubenfiszel:
## Addressed CI review (commit 49ff11215b)

**Fixed**
- **[Codex P1] DuckDB parse failure dropped explicit annotations** — the graph endpoint now falls back to the `// column` annotation lineage when `sqlparser` rejects DuckDB-valid SQL, instead of `unwrap_or_default()` discarding it. (`windmill-api-assets/src/lib.rs`)
- **[cubic P2] Node-id collisions** — `colNodeId` now JSON-encodes the `[kind, path, column]` tuple, so a `#`/`:` in a path or quoted column can't merge two columns. (`columnLineageGraph.ts`, + test)
- **[cubic P2] Multiple ducklake writes mis-anchored lineage** — `buildColumnGraph` now keeps the **first** ducklake write deterministically (v1 has a single materialize target) instead of letting a later edge silently overwrite it. (+ test)
- **[cubic P2] `buildColumnGraph` ran every keystroke** — the `columnGraph` derived is now gated to a ducklake-asset selection, so it doesn't rebuild while editing when the trace UI isn't shown. (`+page.svelte`)
- **[cubic P2] `liveColumnLineage` could leak across a script switch** — `ScriptEditor` now clears `inferredColumnLineage` on a parse error, so a new script whose body fails to parse can't keep the previous script's lineage bound to its path.
- **[cubic P3] Badge overlap** — stacking offset widened 12px → 18px so the data-test and column badges fully clear.

**Deferred, with rationale**
- **[Codex] Live `// column` *annotation* edits aren't overlaid for the active draft** — this is a pre-existing pattern shared with `// data_test` (both read `parsePipelineAnnotations(d.script.content)`), as Pi's re-review noted; it's not introduced here, and the live **inferred** lineage (the DuckDB path) *is* overlaid via `liveBodyAssets`. Touching the shared live-overlay mechanism is out of scope for this PR; happy to do it as a follow-up that covers `data_test` too.
- **[Codex P3] Lineage attached to every ducklake write-edge** — matches the existing `data_test` badge gating and the v1 single-materialize-target assumption; the column **graph** anchoring is now deterministic (first write), which removes the data-integrity concern.

Both Claude and Pi returned **Good to merge**; the above resolves Codex's P1 and cubic's findings. Tests: 8 `columnLineageGraph` cases (incl. the two new fixes), backend compiles clean, frontend error baseline unchanged.

rubenfiszel:
## Addressed second Codex pass (commit 4319d744fe)

Both new P1s fixed and verified live against the running graph API:

- **[P1] JOIN inputs were dropped** — `build_from_maps` now maps every FROM entry *and* its `joins` into the alias map, and single-table bare-column attribution requires no joins (so it isn't misattributed). `SELECT o.id, c.region FROM dl.orders o JOIN dl.customers c …` now resolves `c.region` from `customers`. Added `test_infer_lineage_resolves_joined_inputs`. Verified: the deployed graph now reports `customer_name ← customers.name` and `region ← customers.region` for a real `JOIN`.

- **[P1] Multi-output anchor was a guess** — the column graph now anchors a producer's lineage to its declared `// materialize` target (surfaced as `materialize_target` on the runnable node) instead of picking a ducklake write-edge (unordered for deployed graphs, ambiguous for multi-output). Falls back to a write-edge only when there's no materialize annotation. Tests updated to assert the materialize-target precedence + the write-edge fallback. Verified: `materialize_target` is returned per runnable.

Tests: 9 `columnLineageGraph` + 60 `windmill-parser-sql-asset` (incl. JOIN) green; both backend crates compile clean; frontend baseline unchanged. No SQL query changed (no sqlx cache delta).

rubenfiszel:
**[P1] Badge on every ducklake write-edge — fixed (e2cdceeeb3).** The canvas badge now uses the same materialize-target anchor as `buildColumnGraph`: it lands only on the declared `// materialize` output's write-edge, falling back to the ducklake write-edge when there's no materialize annotation. A multi-output producer no longer shows the column mapping on its non-materialized writes.

rubenfiszel:
**[Codex P2] Column trace could show draft lineage in View — fixed (4459df726f).** The trace graph is now built from `displayGraph` (the exact graph the canvas renders) instead of `graphWithDraft`, so it matches the canvas: draft overlays in edit / show-drafts, deployed-only in plain View.

rubenfiszel:
**[Codex P1] Local/temp staging CTAS reported as final columns — fixed (e6e4b4508a).** Column-lineage inference is now gated to the actual materialized output: a top-level managed-materialize `SELECT`, or a CTAS/`CREATE VIEW` whose target resolves to a real asset. A CTAS into a local/temp staging table is no longer inferred, so its columns can't be anchored to the script's `// materialize` target. Added `test_infer_lineage_skips_local_staging_ctas` + `test_infer_lineage_ctas_into_asset_still_inferred`; verified live (a staged `SELECT amt AS total FROM tmp` reports no `amt` on the final asset, while a direct CTAS keeps its lineage).

rubenfiszel:
**[Codex P1] Multi-output lineage attributed to the wrong asset — fixed (57cd092903).** Inference now tags each entry with its output asset and `parse_assets` scopes the flat list to the `// materialize` target (keeping untagged top-level-SELECT entries). With no declared target, inference is dropped when entries span multiple output assets rather than attributed to an arbitrary one. Parser-internal, no wire change. Added `test_infer_lineage_scopes_to_materialize_target` + `test_infer_lineage_drops_ambiguous_multi_output` (64 sql-asset tests pass). Verified live: a script with `// materialize …/final` + an auxiliary `CREATE TABLE dl.audit AS …` reports only `final`'s `total ← orders.amount` — no `audit` columns.

rubenfiszel:
**[Codex P1] CREATE TEMP under USE leaked a phantom DuckLake column — fixed (8b86bdc4ab).** `track_table_definition` now registers any `temporary` table/view as local *before* active-asset resolution, so a one-part temp name under `USE dl` no longer resolves to `ducklake://…/tmp`. CreateTable/CreateView pass their `temporary` flag. Verified live: the reported `CREATE TEMP TABLE tmp …; SELECT amt AS total FROM tmp` under `USE` now reports no lineage (and no phantom `warehouse/tmp` asset) instead of `final.total <- warehouse/tmp.amt`. Added `test_infer_lineage_temp_table_under_use_is_local` (65 sql-asset tests).

I also did a proactive audit of the whole inference + asset-resolution path: the remaining cases (CTE-through, INSERT…SELECT, partial multi-source) are *missing* lineage, not *false* lineage, and are safe best-effort limitations.

PR #9825, opened 2026-06-27
feat: reusable AI agent steps with rigid linking and edit/fork

rubenfiszel:
## Summary

Makes an AI agent a savable, shareable artifact instead of inline-only config. An agent is stored as a resource of a new built-in `ai_agent` resource type bundling its **brain** (provider/model, system prompt, temperature, output schema, memory…) and its **tools** with their default input bindings. A flow step can **link** to a saved agent so it's reused across flows and updates propagate.

Linking is **rigid**: a linked step is read-only — brain and tools come from the resource. The flow sets the per-invocation inputs (`user_message`/`user_attachments`) and can override any tool input for that flow. To change the agent itself you either **Edit** it (updates it everywhere) or **Unlink** to fork an editable copy into just that step.

## Linking model

- **Linked = read-only.** The step shows the inherited brain config and tools in a single card (*linked to `<path>`*, details behind an info tooltip) with icon-only Edit/Unlink actions. Only `user_message`/`user_attachments` stay editable as step inputs. Linked agent nodes are labeled with the resource path (unless a summary overrides it), and the run/read-only node-definition panel shows a "Linked to `<path>`" line.
- **Tool inputs: agent defaults + per-flow overrides.** The agent carries its tools' default bindings verbatim as authored (static, AI-filled, or flow expressions) — saving round-trips losslessly. Below the step's inputs, each tool gets a panel section with the standard schema-aware input editors (prop picker included) and a read-only view of its code; edits are stored on the step as `tool_inputs`, a **diff** from the resource tool's own transforms, overlaid onto the matching tools at runtime. The rule: editing on a linked step edits the flow's use of the agent; editing under the "Editing" banner edits the agent itself.
- **Graph tool nodes are display-only for linked agents**: the resource's tools render like a native agent's (resolved at flow load, no pop-in; refreshed on link change), clicking one selects the agent step. When the resource can't be resolved (e.g. no access in a read-only view), a single "linked" marker node is shown instead. Read-only viewers resolve tools best-effort with graceful fallback.
- **Edit** (change the agent everywhere): forks the agent into the step's normal editable UI with an "Editing `<path>`" banner (which survives navigating to other nodes); **Save changes** upserts back to the resource and re-links, so the change propagates to every linked flow. **Cancel** keeps the edits as a standalone step instead.
- **Unlink**: forks the resource's brain + tools into the step as a one-off, folding any `tool_inputs` overrides into each tool; the saved agent is untouched.
- **Guard rails**: saving is blocked when the provider isn't a static value (the backend requires a provider on the resource; a computed one would silently produce an agent that fails on every run). If the inherited provider resource isn't accessible in the workspace, the linked step warns.

## Changes

**Backend**
- `FlowModuleValue::AIAgent` gains an optional `agent` field (resource path) and a `tool_inputs` map (per-tool host-flow input overrides); both threaded through the custom deserializer + lockfile resolver. Existing flows are byte-compatible (`skip_serializing_if`).
- Runtime resolution in `windmill-worker/src/ai_executor.rs`: a linked step loads brain config + tools wholly from the resource via `get_resource_value_interpolated` (nested provider `$res:` resolves for free), overlays the flow-local inputs, and overlays `tool_inputs` onto the matching tools' `input_transforms`. Nested agent tools of a linked parent resolve their definition from the resource too.
- The `ai_agent` resource type is defined in the hub (windmill-labs/windmill-integrations#167) and synced into every workspace via the standard cached-resource-type sync, like other built-in types (no migration).

**Frontend**
- `flows/content/AgentResourceBar.svelte` — pick / Save as agent (upsert) / Edit / Unlink; single read-only card; provider-access warning (Alert); static-provider save guard; all resource ops target the operating workspace (session/fork editors).
- `flows/content/AgentToolBindings.svelte` — per-tool override sections in the Step Input tab (schema inferred from the tool definition, standard `InputTransformSchemaForm`, diff-based persistence into `tool_inputs`, read-only code view).
- Linked tools resolve into a scope-keyed store (flow path → module id) at flow load (`flowState.ts`) and in read-only viewers (`FlowGraphViewer`), rendered as display-only tool nodes (`AIToolNode.svelte`, `selectTarget` separate from `moduleId` so agent-node clicks keep the standard selection path).
- Testing a linked agent step forwards `agent` + `tool_inputs` into the single-step flow-preview, and the test form shows only the flow-local inputs.
- `flows/agentResourceUtils.ts` — config↔input_transforms conversion, flow-local filtering, override-diff helper, `summarizeAgentBrain` (unit-tested).
- Wired into `FlowModuleComponent.svelte`; `docs/reusable-ai-agents.md`.

## Screenshots

linked-panel

*Linked step: nodes labeled with the resource path, display-only tool nodes, and the read-only card (linked to `<path>`, inherited config, icon-only Edit/Unlink) in a single scroll region with the step inputs.*

tool-bindings

*Per-tool override section below the step inputs: the agent's default bindings (here `tenant` → `flow_input.tenant`) shown in the standard schema-aware editors with the prop picker; edits persist as a step-local diff in `tool_inputs`. Tool code is viewable read-only.*

## Test plan
- [x] `cargo check` green; `windmill-worker` unit test for the `tool_inputs` overlay; frontend `check:fast` clean; vitest for the agent config/override utils
- [x] Playwright: save/link round-trip keeps tool bindings verbatim (including `flow_input` expressions) → linked read-only card + display-only tool nodes → per-tool override persists as a diff and survives reload → Test this step on a linked agent → Edit → Save changes (upsert + re-link) → Unlink folds overrides into the forked step
- [x] Live e2e with a real Anthropic provider: linked agent resolves the resource + nested provider `$res:` and answers; a linked agent with a **nested agent tool** resolves and executes it (tool call observed end-to-end)
- [x] Selection robustness: agent↔agent and tool→agent click cycles verified after separating the linked tools' select target from the module id

Co-authored-by: hugocasa <hugo@casademont.ch>


rubenfiszel:
## Rigid linking + unlink-to-fork (commit f8df1088cd)

Followed up on the linking model: a **linked agent step is now fully rigid/read-only** — its brain, tools and evals come wholly from the resource (no per-step append/override). To diverge, **Unlink forks** the agent: the resource's brain + tools are copied into the step, which becomes a standalone editable agent again.

### Changes
- **`ai_executor.rs`**: linked branch uses the resource's tools only (no step-local merge); discards `module_tools`.
- **`AgentResourceBar.svelte`**: read-only note; inherited-tools chips; clears the step's own tools on link (save + pick); unlink → "fork" semantics. Plus a **provider-access probe** — if the inherited provider resource isn't accessible in this workspace, a warning tells the user to unlink/fork or gain access to the provider at its path.
- **`AIToolNode.svelte`**: suppress the "+ Tool" add node on a linked step (can't add tools without forking).

### Screenshots
**Linked = rigid (read-only note, inherited tool chip, no "+ Tool"):**
rigid-linked

**After Unlink = fork (full brain editable again, "+ Tool" returns, tools copied in):**
after-fork

### Validation
- `cargo check -p windmill-worker` ✓
- frontend `check:fast` + svelte-autofixer ✓
- Playwright-verified link→rigid and unlink→fork cycles

hugocasa:
/review

Note on the round-10 findings: the two frontend P1s (in-flight Edit race, rename stranding Edit mode) are fixed in the latest commit. The subflow-overlay finding was declined: `FlowModuleValue::Flow` is not a supported agent tool kind anywhere in the pipeline — the execute path rejects it ("Unsupported tool"), the tool-def builder has no Flow arm, and the editor cannot create one — so adding the overlay arm would be dead code for an unreachable tool type.

hugocasa:
/review

Round-11 finding addressed structurally: agent edit state is no longer keyed by the mutable flow path at all. Entries are keyed by workspace + module id and validated against the forked step's `tools` array identity — any wholesale state replacement (undo, redo, session-draft updates, a different flow reusing the module id) produces new objects and automatically invalidates the entry, while an in-place rename (which touches no modules) keeps the banner alive. This removes the migrate/clear machinery (and its baseline-vs-live path mismatch) rather than patching another path-keying edge.

hugocasa:
/review

Round-12 P1 addressed by finishing the identity design: edit entries are now looked up by the fork marker alone (a capped list, no workspace/module-id key at all), so entries from nested editors, repeated module ids, or any other context can never collide — each fork's `tools` array is globally unique. The overlong invariant comment was condensed in the same pass.

PR #9841, opened 2026-06-30
fix(ai-chat): replay assistant turns verbatim so thinking blocks validate

Guilhem-lm:
## Problem

In the **global AI chat** (`/sessions`, dev-gated by `wm_dev_global_ai`), a conversation dies with a 400 from the Anthropic proxy as soon as the agent continues past a turn that **used the native `web_search` tool while thinking was on and ended in a tool call**:

```
400 invalid_request_error
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as they were in the original response.
```

The chat renders the assistant's `Thinking → Searched the web → Thinking → <tool> → ...` turn, then the next (auto-continuation) request fails and the turn cannot proceed.

## Root cause

`frontend/src/lib/components/copilot/chat/anthropic.ts` stores each assistant turn in an OpenAI-shaped message and, on replay, **reconstructs** the Anthropic `content` array in `convertOpenAIToAnthropicMessages`:

- only `thinking` / `redacted_thinking` blocks were preserved (`_anthropicThinkingBlocks`), and they were re-injected **at the front** of the content;
- the turn's `server_tool_use` + `web_search_tool_result` blocks were **dropped**;
- the assistant text was emitted as a **separate** preceding message.

Anthropic validates each thinking block's signature against the blocks that precede it in the latest assistant message. With **interleaved thinking** (a 2nd thinking block after the web-search result), reordering the blocks to the front and dropping the web-search blocks invalidates the 2nd block's signature → 400.

Confirmed by instrumenting the outgoing request body for the failing turn:

```
lastAssistantBlocks = ["thinking","thinking","tool_use"]   // reordered; web-search blocks gone
```

## Fix

Preserve the **full assistant turn verbatim** and replay it unchanged:

- store `finalMessage.content` as `_anthropicContent` on the tool-call message;
- in `convertOpenAIToAnthropicMessages`, when `_anthropicContent` is present, emit it as-is (thinking blocks keep their exact positions/signatures, `server_tool_use` + `web_search_tool_result` are retained, text and `tool_use` keep their order);
- skip the redundant standalone text message for that turn (its text is already inside `_anthropicContent`);
- the old thinking-only reconstruction is **kept as a fallback** for sessions persisted before this change.

## Reproduction

1. Global chat on (`localStorage.setItem('wm_dev_global_ai','1')`, reload) → AI Sessions → new session, Claude model with effort `high`.
2. Prompt that forces web search → reasoning → a workspace tool call in one turn, e.g.
   *"Search the web for the current NIST SP 800-63B minimum password length. After you get the number, in a follow-up step list the resources in my workspace. Think step by step between each tool call."*
3. On auto-continuation the request 400s on `main`; with this change the turn is replayed verbatim and continues.

Reproduced end-to-end against the `demo-orange` workspace via Playwright (before/after screenshots available).

## Validation

- `npm run check:fast` passes.
- Behavioral verification to be done in-app (webmux worktree) with the repro above.

## Notes

- Increases request size for web-search turns (the search-result blocks are now resent) — this is required for signature validity.
- No change to display/persistence shape; only the Anthropic request reconstruction.

### DIFF
--- a/HANDOFF.md
+++ b/HANDOFF.md
@@ -1,0 +1,60 @@
+# Handoff: AI-chat thinking-blocks fix (PR #9841)
+
+Branch: `glm/fix-copilot-thinking-blocks` · base `main` · file touched: `frontend/src/lib/components/copilot/chat/anthropic.ts`
+
+## The bug
+Global AI chat (`/sessions`, dev-gated) 400s when an assistant turn uses the native
+`web_search` tool while thinking is on AND ends in a tool call:
+
+```
+400 invalid_request_error
+messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
+message cannot be modified. These blocks must remain as they were in the original response.
+```
+
+Root cause: the turn was reconstructed keeping only thinking blocks, reordering them to the
+front and dropping the `server_tool_use` / `web_search_tool_result` blocks. With interleaved
+thinking (a 2nd thinking block after the web-search result) the 2nd block's signature no
+longer validates.
+
+## What's already done (1 commit)
+`frontend/src/lib/components/copilot/chat/anthropic.ts`:
+- `parseAnthropicCompletion`: store the full turn as `_anthropicContent = finalMessage.content`
+  (replaces the thinking-only `_anthropicThinkingBlocks` extraction).
+- `convertOpenAIToAnthropicMessages`: if `_anthropicContent` is present, emit it **verbatim**;
+  skip the redundant standalone text message for that turn; the old `_anthropicThinkingBlocks`
+  path is kept as a fallback for sessions persisted before this change.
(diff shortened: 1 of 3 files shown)


hugocasa:
Superseded by #9843, which reimplements this fix with correct `cache_control` handling (the prefix breakpoint now lands on the trailing tool_result without mutating the verbatim turn) and adds live end-to-end verification against the real Anthropic API. Continuing in #9843.

hugocasa:
Superseded by #9844, which reimplements this fix: replays the assistant turn verbatim, plus reworks the trailing cache_control so the conversation prefix is actually cached on continuation requests (and is never injected into the verbatim turn), and adds unit coverage for the skip-bound. Verified live end-to-end against the real Anthropic API — the continuation request that 400s on `main` now returns 200.

PR #9850, opened 2026-07-01
feat(pipelines): add managed SCD2 history materialize strategy

rubenfiszel:
## Summary

Adds a first-class **SCD type 2 (Slowly Changing Dimension) history** strategy to the DuckLake pipeline `// materialize` annotation. Instead of overwriting a dimension on change (SCD1 `merge`) or losing history entirely, the runtime keeps every version of every row — a change closes the prior version (`valid_to`, `is_current = false`) and opens a new one.

Before this PR the only way to get per-entity history was `// materialize manual` with hand-written `CREATE`/`UPDATE`/`INSERT` SQL, which **drops `// data_test` and schema capture**. This makes SCD2 a *managed* strategy, so history is generated for you and the quality/metadata features keep working.

---

## Why build it (vs. what we already had)

DuckLake time-travel and the `merge` (SCD1) strategy already exist, but neither answers *"give me each entity's version history as queryable rows, joinable by effective date."*

| Question | Answered by |
|---|---|
| "What did the *whole table* look like at commit N?" | **Time-travel** (`AT (VERSION n)`) — free, strictly better |
| "Roll back / reproduce / audit a past run" | **Time-travel** — free |
| "Current state only, overwrite on change" | **`merge`** (SCD1) — already exists |
| "Each *entity's* version history as rows, joinable by effective date" | **SCD2** ← this PR |

The single decisive justification is the **as-of join** — for a fact row at time `T`, join the dimension version that was current *at `T`*:

```sql
… ASOF JOIN dim d ON fact.customer_id = d.id AND fact.ordered_at >= d.valid_from
```

Time-travel (keyed by commit, not business-effective time) can't express this in one set-based query; `merge` overwrites history. This closes the documented dbt-parity gap #4 (`docs/pipelines-vs-dbt.md`).

---

### Annotation syntax

```
// materialize ducklake://<lake>/<dim> key=<natural_key> history [track=<c1,c2,…>] [deletes=close] [// data_test …]
SELECT <natural_key>, <col1>, <col2>, …   -- the CURRENT snapshot of the dimension (one row per key)
FROM …
```

- **`history`** (the flag that turns a keyed merge into SCD2). The leading keyword **`scd2`** is a recognized alias: `// materialize scd2 ducklake://… key=id`.
- **`key=<col>`** *(required)* — the natural/business key that identifies an entity across versions.
- **`track=<c1,c2,…>`** *(optional)* — the columns whose change opens a new version. Default = **all non-key columns**. Must be a bare comma list with **no spaces** (`track=name,tier`, not `track=name, tier`).
- **`deletes=close`** *(optional)* — hard-delete-close: a key that disappears from the snapshot has its current version closed. Default is soft-delete (absent keys stay current). See the reconciliation table.
- The `SELECT` returns the **current desired state**, one row per key — exactly like `merge`. You never write `valid_from`/`valid_to`/`is_current`; the runtime manages them.

### Managed columns

The runtime appends three columns to your dimension (kept clean so consumers can write `WHERE is_current` / `ASOF JOIN … >= valid_from`):

| Column | Type | Meaning |
|---|---|---|
| `valid_from` | `TIMESTAMP` | when this version became effective (the run's transaction clock) |
| `valid_to` | `TIMESTAMP` | when it was superseded; `NULL` = still open |
| `is_current` | `BOOLEAN` | `true` for the live version of each key |

### Reconciliation semantics (each run)

The `SELECT` (the current snapshot) is diffed against the live rows:

| Case | Behavior |
|---|---|
| **Initial load** (empty table) | all rows opened, no closes |
| **Tracked column changed** | prior version closed (`valid_to = now()`, `is_current = false`), new version opened |
| **New key** (in SELECT, not in table) | opened as current |
| **No change** | no-op — 0 rows written, no churn (idempotent) |
| **Untracked column changed** | ignored — no new version |
| **Key absent from SELECT** (deleted upstream) | default: **stays current** (soft delete — dbt's `hard_deletes=ignore`); with `deletes=close`: current version **closed**, no reopen |
| **Deleted key reappears** (with `deletes=close`) | opens a fresh version, leaving a validity gap between the delete and the reactivation |

The effective timestamp is the run's transaction clock (`now()`), which DuckDB fixes to transaction start — so within a run `valid_to(old) == valid_from(new)` (contiguous, gap-free versions).

### Constraints (v1)

- **Non-partitioned only** — `// partitioned` + `history` is rejected with a clear error.
- **Reserved names** — `valid_from`/`valid_to`/`is_current` are reserved columns (a SELECT projecting one fails at run time); the `<dim>_current` suffix is reserved for the companion view.
- **Schema frozen at first run** — like `merge`/`append`, the table is `CREATE TABLE IF NOT EXISTS`; changing the SELECT's projected columns later fails (an append-only history can't reshape closed versions) and needs a manual rebuild.
- **Deletes** — soft by default; `deletes=close` for hard-delete-close (both dbt behaviors). Full CDC (`cdc_operation` column) is a follow-up.

---

### 1. A dimension with history

```sql
-- pipeline
-- on ducklake://analytics/stg_customers
-- materialize ducklake://analytics/dim_customer key=id history track=name,tier
-- data_test not_null valid_from
-- data_test accepted_values tier = free,pro,enterprise
SELECT id, name, tier          -- current snapshot, one row per id
FROM ducklake://analytics/stg_customers;
```

Run 1 loads `(1,alice,free) (2,bob,pro)`. Run 2 changes bob `pro→enterprise` and adds `(3,carol,free)` → the resulting `dim_customer`:

```
┌─────┬───────┬────────────┬─────────────────────────┬─────────────────────────┬────────────┐
│ id  │ name  │    tier    │       valid_from        │        valid_to         │ is_current │
├─────┼───────┼────────────┼─────────────────────────┼─────────────────────────┼────────────┤
│  1  │ alice │ free       │ 2026-07-01 06:16:44.848 │ NULL                    │ true       │ ← untouched
│  2  │ bob   │ pro        │ 2026-07-01 06:16:44.848 │ 2026-07-01 06:16:45.932 │ false      │ ← closed
│  2  │ bob   │ enterprise │ 2026-07-01 06:16:45.932 │ NULL                    │ true       │ ← new version
│  3  │ carol │ free       │ 2026-07-01 06:16:45.932 │ NULL                    │ true       │ ← new key
└─────┴───────┴────────────┴─────────────────────────┴─────────────────────────┴────────────┘
```

(bob's `valid_to` == his new version's `valid_from` — contiguous. Re-running with no change writes 0 rows.)

### 2. Reading the latest version — the auto-generated `_current` view

Each run (re)creates a `<dim>_current` view, so the common case needs no filter:

```sql
-- pipeline
-- on ducklake://analytics/dim_customer_current
SELECT * FROM ducklake://analytics/dim_customer_current;   -- only is_current rows
```

### 3. The payoff — effective-dated (as-of) join

Get the tier each customer had *at the moment they ordered*:

```sql
-- pipeline
-- on ducklake://analytics/dim_customer
-- on ducklake://analytics/stg_orders
-- materialize ducklake://analytics/fct_orders
SELECT o.order_id, o.amount, o.ordered_at,
       d.tier AS tier_at_order_time
FROM ducklake://analytics/stg_orders o
ASOF JOIN ducklake://analytics/dim_customer d
  ON o.customer_id = d.id AND o.ordered_at >= d.valid_from;
```

### 4. Track all columns (default)

Omit `track=` to open a new version on any non-key column change:

```sql
-- materialize ducklake://analytics/dim_product key=sku history
SELECT sku, name, category, price FROM ducklake://analytics/stg_products;
```

### 5. `scd2` keyword alias

```sql
-- materialize scd2 ducklake://analytics/dim_customer key=id
```

### 6. Hard-delete-close (`deletes=close`)

When a customer disappears from the source, close their current version instead of leaving it open forever:

```sql
-- materialize ducklake://analytics/dim_customer key=id history deletes=close
SELECT id, name, tier FROM ducklake://analytics/stg_customers;
```

If id=2 vanishes from the snapshot → its current version is closed (`valid_to` set, `is_current=false`), no new version. If id=2 later reappears → a fresh version opens, leaving a validity gap. Verified end-to-end against a real DuckLake.

---

## What gets generated (the codegen)

For `key=id history` (default track), the runtime replaces the trailing SELECT with (one transaction for the mutation):

```sql
CREATE TABLE IF NOT EXISTS <t> AS
  SELECT *, CAST(NULL AS TIMESTAMP) AS valid_from, CAST(NULL AS TIMESTAMP) AS valid_to,
         CAST(NULL AS BOOLEAN) AS is_current FROM (<select>) WHERE false;                -- bootstrap
CREATE OR REPLACE TEMP TABLE _wm_scd2_changed AS                                          -- diff BEFORE the write
  SELECT "id" FROM (SELECT * FROM (<select>)
                    EXCEPT SELECT * EXCLUDE (valid_from, valid_to, is_current) FROM <t> WHERE is_current);
BEGIN TRANSACTION;
  UPDATE <t> SET valid_to = CAST(now() AS TIMESTAMP), is_current = false                  -- close old
    WHERE is_current AND EXISTS (SELECT 1 FROM _wm_scd2_changed
                                 WHERE _wm_scd2_changed."id" IS NOT DISTINCT FROM <t>."id");
  INSERT INTO <t> SELECT s.*, CAST(now() AS TIMESTAMP), CAST(NULL AS TIMESTAMP), true      -- open new
    FROM (<select>) s WHERE EXISTS (SELECT 1 FROM _wm_scd2_changed c
                                    WHERE c."id" IS NOT DISTINCT FROM s."id");
  CREATE VIEW IF NOT EXISTS <t>_current AS SELECT * FROM <t> WHERE is_current;            -- companion view (in-txn, once)
COMMIT;
```

With `deletes=close`, a second temp set `_wm_scd2_deleted` (current keys `EXCEPT` snapshot keys) and a second null-safe close `UPDATE` are inserted before the open — with **no** matching `INSERT`, so vanished keys close without reopening.

Key correctness points:
- The changed-key set is captured **before** the transaction (the close flips `is_current`, so recomputing the diff after it would be wrong).
- Close is `UPDATE` (not `MERGE INTO` — DuckLake's MERGE is unreliable on fresh parquet); `EXCEPT` treats NULLs as equal so unchanged NULLs aren't false changes.
- Key matching is **null-safe** (`IS NOT DISTINCT FROM` via `EXISTS`, not `key IN (…)` — `IN` never matches `NULL`, which would drop a NULL-key row).
- The `<dim>_current` view is created **inside** the transaction with **`IF NOT EXISTS`** (created once), because `CREATE VIEW` advances the DuckLake snapshot — `CREATE OR REPLACE` outside the txn would both record the wrong snapshot and churn a snapshot on every unchanged rerun.

---

## Changes

- **Parser (both surfaces, kept in parity):** `MaterializeSpec` gains `scd2` + `track`; `history` flag / `scd2` alias / `track=` parsed in `backend/parsers/windmill-parser/src/asset_parser.rs` and `frontend/…/parsePipelineAnnotations.ts`. Shared fixture corpus + Rust & TS parity harnesses extended.
- **Codegen (`sql_materialize.rs`):** new `MaterializeStrategy::Scd2 { key, track }` + `scd2_statements()` (diff → close → open → `_current` view).
- **Worker (`duckdb_executor.rs`):** strategy derivation + guards (`history` requires `key=`; `// partitioned` + history rejected). Flows through the same summary → `// data_test` + schema-capture path as other managed strategies.
- **Docs:** `ducklake-materialization.md`, `pipelines-vs-dbt.md`.

**Not touched:** no `*_ee` files (no companion PR); no `.svelte` files (the frontend change is pure parser logic).

---

## Test plan

- [x] `cargo test -p windmill-parser` — parser unit tests (history flag, `scd2` alias, key-without-history stays SCD1) + codegen tests (diff/close/open shape, `_current` view, no `MERGE INTO`) + parity harness.
- [x] `cargo test -p windmill-worker --features duckdb,parquet` — wiring tests (managed wrap emitted, key required, partitioned rejected).
- [x] `npx vitest run parsePipelineAnnotations` — TS parser + parity.
- [x] **SQL-level e2e against a real DuckLake** (`duckdb` CLI, sqlite catalog + local storage): ran the exact generated statements across runs — confirmed correct type-2 history (change→close+open, new key→insert, absent key→soft-delete, idempotent no-op), `deletes=close` (vanished key closed, reactivation opens a fresh version with a validity gap), the `<dim>_current` view persisting across sessions, and an `ASOF JOIN … >= valid_from` resolving the effective-dated version.
- [ ] Manual verification on a full EE + object-store stack (the local CE instance has no configured ducklake): deploy `// materialize ducklake://<lake>/<dim> key=id history`, run twice, confirm the history table + `<dim>_current` view + both rejection errors.

## CI review

All findings from the automated reviewers (Claude, Pi, Codex, cubic) addressed:

- **Codex [P1]** — NULL natural keys silently dropped (`key IN (…)` never matches NULL): fixed with null-safe `IS NOT DISTINCT FROM` matching (`8f09010`).
- **cubic [P2]** — `_current` view advanced the recorded DuckLake snapshot: fixed by creating the view inside the write transaction (`8f09010`).
- **Codex [P2]** — unchanged reruns still advanced the snapshot via `CREATE OR REPLACE VIEW`: fixed with `CREATE VIEW IF NOT EXISTS` (created once, no-op thereafter) (`a9d27e3`).
- **Claude/cubic [P2 ×3]** — `track=` spacing, `_current` reserved name, schema-freeze: documented contracts (`1dd682e`).

Both substantive fixes verified against a real DuckLake (NULL-key row materialized; exactly one snapshot per run).

### DIFF
--- a/backend/parsers/windmill-parser/src/asset_parser.rs
+++ b/backend/parsers/windmill-parser/src/asset_parser.rs
@@ -224,15 +224,20 @@ pub struct RetrySpec {
     pub delay: Option<String>,
 }
 
-// `// materialize [manual] <asset> [append] [key=<col>]` — declares that this
-// script produces a *managed* materialization of `<asset>` (a `ducklake://`
-// table). By default the runtime generates the write DDL around the script's
-// single trailing `SELECT` and owns idempotency, partition-state and snapshot
-// capture. `manual` is the escape hatch: the script writes its own DDL and the
-// runtime only records state (track-only). The reconciliation strategy options
-// (`append`, `key=<col>`) apply to managed mode: none → DELETE-by-partition +
-// INSERT (replace); `key=<col>` → MERGE (dedup within slice); `append` →
+// `// materialize [manual] <asset> [append] [key=<col>] [history] [track=<c1,c2>]`
+// — declares that this script produces a *managed* materialization of `<asset>`
+// (a `ducklake://` table). By default the runtime generates the write DDL around
+// the script's single trailing `SELECT` and owns idempotency, partition-state
+// and snapshot capture. `manual` is the escape hatch: the script writes its own
+// DDL and the runtime only records state (track-only). The reconciliation
+// strategy options apply to managed mode: none → DELETE-by-partition + INSERT
+// (replace); `key=<col>` → MERGE (dedup within slice, SCD type 1); `append` →
 // INSERT-only. `append` wins if both are given (deploy-time warning).
+// `key=<col> history` upgrades the merge to SCD type 2: the SELECT is the current
+// snapshot (one row per key), and a change to any tracked column (`track=`,
+// default all non-key) closes the prior version and opens a new one, keeping full
+// history (`valid_from`/`valid_to`/`is_current`). The leading keyword `scd2` is a
+// recognized alias for `history`.
 #[derive(Serialize, Debug, PartialEq, Clone)]
 pub struct MaterializeSpec {
     pub target_kind: AssetKind,
@@ -243,6 +248,15 @@ pub struct MaterializeSpec {
     pub append: bool,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub unique_key: Option<String>,
+    // `scd2` managed history mode: the SELECT is the current snapshot (one row
+    // per `unique_key`), and the runtime maintains a Slowly-Changing-Dimension
+    // type-2 history (`valid_from`/`valid_to`/`is_current`). `unique_key` (the
+    // `key=` opt) is the natural key; `track` lists the columns whose change
+    // opens a new version (empty ⇒ all non-key columns). Managed mode only.
+    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
+    pub scd2: bool,
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub track: Vec<String>,
 }
 
 // `// data_test <kind> …` — a data-quality assertion run against the
@@ -800,19 +814,30 @@ fn parse_retry_spec(s: &str) -> Option<RetrySpec> {
     Some(RetrySpec { count, delay })
 }
 
-// Parse a `// materialize [manual] <asset> [append] [key=<col>]` right-hand
-// side. An optional leading `manual` token (whitespace-delimited) opts out of
-// managed mode (track-only). The next whitespace token is the target asset URI
+// Parse a `// materialize [manual] <asset> [append] [key=<col>] [history]
+// [track=<c1,c2>]` right-hand side. An optional leading `manual` token opts out
+// of managed mode (track-only); a leading `scd2` token is an alias for the
+// `history` flag. The next whitespace token is the target asset URI
 // (default-syntax shorthands enabled, so `ducklake` → `ducklake://main`); the
-// remainder are strategy options — bare `append` and `key=<col>` (merge key),
-// which apply to managed mode only. A missing/empty target yields `None` (the
-// annotation is dropped, fail-safe).
+// remainder are strategy options — bare `append`, bare `history` (SCD type-2 on
+// a keyed merge), `key=<col>` (merge/scd2 key), and `track=<c1,c2,…>` (scd2
+// tracked columns) — which apply to managed mode only. A missing/empty target
+// yields `None` (the annotation is dropped, fail-safe).
 fn parse_materialize_spec(s: &str) -> Option<MaterializeSpec> {
-    let (manual, rest) = match s.strip_prefix("manual") {
-        Some(after) if after.is_empty() || after.starts_with(char::is_whitespace) => {
-            (true, after.trim_start())
-        }
-        _ => (false, s),
+    // One optional leading mode keyword: `manual` (escape hatch, track-only) or
+    // `scd2` (alias for the `history` flag below). A missing keyword is the
+    // default managed mode.
+    fn strip_mode<'a>(s: &'a str, kw: &str) -> Option<&'a str> {
+        s.strip_prefix(kw)
+            .filter(|after| after.is_empty() || after.starts_with(char::is_whitespace))
+            .map(|after| after.trim_start())
+    }
+    let (manual, scd2_kw, rest) = if let Some(after) = strip_mode(s, "manual") {
+        (true, false, after)
+    } else if let Some(after) = strip_mode(s, "scd2") {
+        (false, true, after)
+    } else {
+        (false, false, s)
     };
     let mut it = rest.trim().splitn(2, char::is_whitespace);
     let asset_tok = it.next()?;
@@ -822,11 +847,32 @@ fn parse_materialize_spec(s: &str) -> Option<MaterializeSpec> {
         return None;
     }
     let append = opts_str.split_whitespace().any(|t| t == "append");
-    let unique_key = parse_kv_opts(opts_str)
-        .get("key")
-        .filter(|k| !k.is_empty())
-        .cloned();
-    Some(MaterializeSpec { target_kind, target_path: path.to_string(), manual, append, unique_key })
+    // SCD type-2 history mode. The primary spelling is the bare `history` flag on
+    // a keyed merge (`key=<col> history`) — it reads as "a keyed upsert that keeps
+    // history"; the leading `scd2` keyword is a recognized alias for the same.
+    let scd2 = scd2_kw || opts_str.split_whitespace().any(|t| t == "history");
+    let opts = parse_kv_opts(opts_str);
+    let unique_key = opts.get("key").filter(|k| !k.is_empty()).cloned();
+    // `track=<c1,c2,…>` (scd2 only): comma-separated columns whose change opens a
+    // new version. Empty entries dropped; an empty list ⇒ track all non-key cols.
+    let track = opts
+        .get("track")
+        .map(|v| {
+            v.split(',')
+                .map(|c| c.trim().to_string())
+                .filter(|c| !c.is_empty())
+                .collect::<Vec<_>>()
+        })
+        .unwrap_or_default();
+    Some(MaterializeSpec {
+        target_kind,
+        target_path: path.to_string(),
+        manual,
+        append,
+        unique_key,
+        scd2,
+        track,
+    })
 }
 
 // Parse a `// data_test <kind> …` right-hand side into one `DataTest`. The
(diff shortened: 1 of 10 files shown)


rubenfiszel:
Thanks — addressed the three P2 nits in `1dd682e` as documented contracts (consistent with the annotation grammar and the persist-and-mutate schema-freeze behavior shared with `merge`/`append`):

1. **`track=` spacing** — documented that `track=` (like every `=`-option in the whitespace-tokenized grammar) takes a bare comma list with no spaces (`track=a,b`), in both parser doc comments and the user docs.
2. **`<dim>_current` collision** — documented `_current` as a reserved suffix (code comment at the view codegen + docs); the post-commit failure mode is noted.
3. **Schema evolution** — documented the first-run schema freeze for scd2 in `ducklake-materialization.md` (append-only history can't reshape closed versions; `replace` is the only re-deriving strategy).

Pi and Codex both returned "good to merge"; no P0/P1 from any reviewer. The EE + object-store manual run remains the one unchecked box (local instance is CE with no configured ducklake) — the generated SQL was validated against a real DuckLake via the duckdb CLI.

rubenfiszel:
Addressed the two substantive findings from this round in `8f09010` (both verified against a real DuckLake):

**Codex [P1] — NULL natural keys silently dropped.** Confirmed: `WHERE key IN (SELECT key FROM _wm_scd2_changed)` never matches `NULL`, so a NULL-key row was flagged as changed by the `EXCEPT` diff yet skipped by both the close and the open — dropped from history. Fixed by matching keys null-safely with `IS NOT DISTINCT FROM` via a correlated `EXISTS` in both the close `UPDATE` and the open `INSERT`. A NULL-key row is now materialized (still ill-formed for a dimension — docs recommend `// data_test not_null <key>` — but no longer vanishes). Verified: a `(NULL,'ghost')` row survives across runs.

**cubic [P2] — `_current` view advanced the recorded snapshot.** Confirmed empirically: `CREATE OR REPLACE VIEW` is a catalog change that bumps the DuckLake snapshot (I measured `max(snapshot_id)` 4→5→6 on repeated view creates). Since it ran after `COMMIT`, the summary's `max(snapshot_id)` recorded the view DDL's snapshot, not the data write. Fixed by moving the view creation **inside** the write transaction, so one snapshot covers the write + view and the summary records the data snapshot. Bonus: a `<dim>_current` name collision now rolls back atomically instead of committing-then-erroring. Verified: 2 runs produce exactly the expected snapshot count (init + create + 2 writes), no extra view snapshots.

The other cubic P2s (`track=` spacing, schema-freeze, `_current` reserved name) were already addressed as documented contracts in `1dd682e`. Tests updated (codegen assertions now check the null-safe `EXISTS`/`IS NOT DISTINCT FROM` shape, the in-txn view order, and assert no `IN (SELECT` remains). All parser + worker tests green.

rubenfiszel:
Addressed the Codex [P2] in `a9d27e3`: unchanged reruns advancing the snapshot.

Root cause: `CREATE OR REPLACE VIEW` is a catalog write that advances the DuckLake snapshot *every run*, so a no-change run (empty `_wm_scd2_changed` → UPDATE/INSERT touch 0 rows) still minted and recorded a snapshot. Since the `<dim>_current` view definition never changes (`SELECT * WHERE is_current` always reflects live data), switched to **`CREATE VIEW IF NOT EXISTS`** — created once (folded into the first data-write snapshot), true no-op thereafter.

Verified against a real DuckLake: after the initial load, two unchanged reruns keep `max(snapshot_id)` constant (empty write transaction + `IF NOT EXISTS` view = no new snapshot). Bonus: the reserved-name collision now skips silently (`IF NOT EXISTS` when a table of that name exists) instead of erroring — updated the code comment and docs accordingly.

Pi returned "good to merge" on the prior commit. Tests updated and green.

rubenfiszel:
Addressed the Pi [P2] in `2b212d5`: the deploy-time precedence warning was stale.

Runtime precedence (`duckdb_executor` strategy derivation) is `scd2 (history) > append > merge (key=) > replace`, but the deploy warning in `scripts.rs` only knew `append > key` — so `// materialize ... key=id append history` warned "append wins" while the runtime actually runs SCD2. Fixed by warning for `history + append` (history wins, append ignored) ahead of the `append + key` case, mirroring the runtime order. Compiles (`cargo check -p windmill-api-scripts`).

Codex returned "good to merge, no issues" and Pi verified the `deletes=close` implementation is correct. cubic still running; I'll fold in anything it flags.

rubenfiszel:
Addressed the Codex [P1] in `13766f6`: the `<dim>_current` view wasn't a dispatchable asset.

Right catch — the docs present `// on ducklake://…/<dim>_current` as a subscription, but deploy registered only the base table as a write asset, so the cascade (which fans out from deploy-time asset rows) would never dispatch a subscriber on the view. Fixed by registering `<dim>_current` as a produced write asset when `m.scd2`, alongside the base table — so both `// on …/<dim>` and `// on …/<dim>_current` fire. The producer genuinely (re)creates the view each run, so it's an accurate write edge. Compiles.

Pi returned "good to merge" on this commit; cubic still running. I'll fold in anything it flags.

rubenfiszel:
Addressed the Codex [P2] in `68e74e2`: gated the `_current` asset registration on `!manual`.

`// materialize manual … history` sets both `manual` and `scd2`, but manual mode short-circuits before the SCD2 codegen (`duckdb_executor.rs:174`), so no `<dim>_current` view is created — registering it there was a false write edge. Now gated as `m.scd2 && !m.manual`, so only managed scd2 (which actually creates the view) registers it.

Pi returned "good to merge" on the prior commit. cubic still hasn't landed a verdict (superseded each round). I'll fold in anything it flags.

PR #9851, opened 2026-07-01
feat(ansible): support repo-provided ansible.cfg in delegate_to_git_repo

rubenfiszel:
## Summary

When an Ansible script delegates to a git repo (`delegate_to_git_repo`), the repo's own `ansible.cfg` was silently ignored. Ansible loads exactly one config file — the first found in `ANSIBLE_CONFIG` → cwd → `~/.ansible.cfg` → `/etc/ansible/ansible.cfg`, with no merging. Windmill runs `ansible-playbook` with cwd = `job_dir` and writes its own generated `ansible.cfg` there, while the repo is cloned into the `delegate_git_repository/` subdir — so the user's config is never on the discovery path and Windmill's always wins.

That breaks any repo that relies on `ansible.cfg` for `roles_path`, `[inventory] enable_plugins` (dynamic inventories), callbacks, `host_key_checking`, ssh args, etc. The only prior workaround was manually setting individual `ANSIBLE_*` env vars.

This adds an opt-in `ansible_cfg` field so the repo's config becomes the effective one, with Windmill layering back only the settings that depend on runtime state it alone controls.

## Design

When `delegate_to_git_repo.ansible_cfg` (a repo-relative path) is set, `ANSIBLE_CONFIG` is pointed at it so the repo config is authoritative. Windmill's settings are re-applied via environment variables (which outrank `ansible.cfg`), split on a testable rule — *runtime-bound vs static*, not "importance":

- **Runtime-bound → always injected** (`ANSIBLE_HOME`, `ANSIBLE_LOCAL_TEMP`, `ANSIBLE_REMOTE_TEMP`): reference the ephemeral job dir, cannot be expressed in a checked-in config.
- **Vault → Windmill wins** (`ANSIBLE_VAULT_PASSWORD_FILE`, `ANSIBLE_VAULT_IDENTITY_LIST`): Windmill manages the secret. The one deliberate override.
- **Dependency search paths → additive** (`ANSIBLE_ROLES_PATH`, `ANSIBLE_COLLECTIONS_PATH`): Windmill's install dirs (`{job_dir}/roles`, `{job_dir}`) are *prepended* to the repo cfg's own declared paths (parsed from the cfg; relative entries resolved against the cfg directory, matching ansible).
- **Everything else** (inventory plugins, callbacks, ssh args…) comes straight from the repo cfg.

Applied only on the non-sandboxed execution path: git-repo delegation clones into `job_dir`, which the nsjail profile does not mount, so delegation already requires `DISABLE_NSJAIL`. A warning is logged (and the run proceeds unchanged) if `ansible_cfg` is set while sandboxing is enabled.

## Changes

- `windmill-parser-yaml`: add optional `ansible_cfg` field to `DelegateToGitRepoDetails` and its extractor.
- `ansible_executor.rs`: validate + existence-check the repo-relative cfg path post-clone (reuses `validate_relative_path`, rejects `..`/absolute); `build_ansible_cfg_override_envs` assembles the layered env overrides; `parse_ansible_cfg_path_list` reads `roles_path`/`collections_path` from the `[defaults]` section; `resolve_and_prepend_path` does the additive prepend; overrides applied last on the non-sandboxed command builder so they win over `reserved_variables` and the repo cfg.
- Re-validate `vault_id` entries before joining into `ANSIBLE_VAULT_IDENTITY_LIST` (defense-in-depth, matching `create_ansible_cfg`).

## Test plan

- [x] `cargo check -p windmill-worker --features python` — compiles clean
- [x] Parser unit tests: field present/absent
- [x] INI parser: sections, comments, colon-separated lists
- [x] Path prepend/resolution (relative vs absolute vs `~`)
- [x] `build_ansible_cfg_override_envs` end-to-end env assembly incl. vault + prepend
- [x] `test_ansible_cfg_override_resolves_repo_roles_e2e`: runs the real `ansible-playbook` with a negative control — a role living only under the repo's `roles_path` fails to resolve without the override and resolves with it (self-skips when `ansible-playbook` is absent)
- [ ] Manual devbox (`DISABLE_NSJAIL=true`): repo-relative `roles_path` role resolves; vault-encrypted var still decrypts via Windmill's managed vault password (override wins over any repo `vault_password_file`); warning fires + run unchanged when sandboxing is enabled

---
## Summary by cubic
Adds opt‑in support for using a repo’s own `ansible.cfg` when running `delegate_to_git_repo`, so repo settings (roles paths, inventory plugins, callbacks, SSH options) take effect. Windmill points `ANSIBLE_CONFIG` to the file and layers only runtime and vault settings via env vars.

- **New Features**
  - Add `delegate_to_git_repo.ansible_cfg` (repo‑relative path). Validates path and existence after clone.
  - Use the repo config with env overrides for runtime dirs and vault; dependency paths are additive and resolve relative to the cfg directory (Windmill paths are prepended).
  - INI parser accepts `:` as a key/value delimiter and the deprecated `collections_paths` alias.
  - Only active when not sandboxed; if sandboxing is enabled, log a warning and proceed unchanged.

- **Migration**
  - No breaking changes; default behavior is unchanged.
  - To opt in, set `delegate_to_git_repo.ansible_cfg: <path>` in your YAML (path relative to the repo root).
  - Requires running without nsjail (`DISABLE_NSJAIL=true`); otherwise the override is ignored with a warning.

<sup>Written for commit 934e3f669d51d68af14a39f4cbfdbe4451d70074. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/parsers/windmill-parser-yaml/src/lib.rs
+++ b/backend/parsers/windmill-parser-yaml/src/lib.rs
@@ -269,6 +269,12 @@ pub struct DelegateToGitRepoDetails {
     pub commit: Option<String>,
     pub inventories_location: Option<String>,
     pub vars_location: Option<String>,
+    /// Path (relative to the cloned repo root) of an `ansible.cfg` to use as the
+    /// effective config for the run. When set, Windmill points `ANSIBLE_CONFIG` at
+    /// it so the repo's own settings (roles paths, inventory plugins, callbacks…)
+    /// apply, and only injects the settings that depend on runtime state it alone
+    /// controls (temp/home dirs, vault password) on top.
+    pub ansible_cfg: Option<String>,
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub install_requirements: bool,
 }
@@ -629,6 +635,10 @@ fn extract_delegate_to_git_repo_details(value: &Yaml) -> Option<DelegateToGitRep
                 .get(&Yaml::String("vars_location".to_string()))
                 .and_then(|s| s.as_str())
                 .map(|s| s.to_string());
+            let ansible_cfg = v
+                .get(&Yaml::String("ansible_cfg".to_string()))
+                .and_then(|s| s.as_str())
+                .map(|s| s.to_string());
             let install_requirements = v
                 .get(&Yaml::String("install_requirements".to_string()))
                 .and_then(|s| s.as_bool())
@@ -640,6 +650,7 @@ fn extract_delegate_to_git_repo_details(value: &Yaml) -> Option<DelegateToGitRep
                 commit,
                 inventories_location,
                 vars_location,
+                ansible_cfg,
                 install_requirements,
             });
         }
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Addressed the CI review findings in 934e3f669d:

- **[P1 codex / P2 cubic] `:` key delimiter** — `parse_ansible_cfg_path_list` now splits key/value on the first `=` *or* `:` (Python configparser semantics), while still splitting the value on `:` as the path separator. So `roles_path: my_roles` and `collections_path: a:b` are honored. Added `test_parse_ansible_cfg_path_list_colon_delimiter`.
- **[P2 claude] `collections_paths` plural alias** — the collections lookup now falls back to the deprecated plural ini alias, so a repo using it isn't silently dropped when the env override replaces the cfg value. Added `test_build_ansible_cfg_override_envs_collections_paths_alias`.

Not changed (deliberate, per the reviewers' own guidance): env-var interpolation inside a repo path value (e.g. `roles_path = $HOME/roles`) — the parser is documented as deliberately minimal and this is a rare, non-blocking edge case flagged only as a known limitation.

PR #9861, opened 2026-07-01
fix(tree-view): match file indentation with folder indentation (depth*16)

### LINKED ISSUE #9726: bug: Small UI bug with indentation of files and folders in tree view
### Describe the bug

scripts inside sub folders have much larger indentation than folders in that same sub folder. 
(If this is intentional design please ignore this bug report)

Image

### To reproduce

1. Make a script inside a/b/c/d/script_A.py
2. Make a script inside a/b/c/d/e/script_B.py

in homepage UI, I would expect script_A to have same indentation as folder e

### Expected behavior

A script and folder inside same parent folder to have same level of indentation

### Screenshots

Extreme example:

Script_A is in same parent as folder e (which is pretty much impossible to see from UI)
Image

### Browser information

Tested in recent chrome and edge browsers (Been this way for a while i think)

### Application version

v1.35.0 (enterprise)

### Additional Context

_No response_

tusharui:
Files in the tree view indented at `depth * 32` px while folders used
`depth * 16` px, causing files to be offset twice as much as folders
at each nesting level. Changed Row.svelte to use `depth * 16` to match
TreeView.svelte.
Fixes #9726.

---
## Summary by cubic
Match file indentation to folders in the tree view by using depth*16 px in `Row.svelte`, fixing the double offset. Use flex-grow instead of h-full for the deploy screen scroll container in `CenteredPage.svelte`. Fixes #9726.

<sup>Written for commit 8939275751eba7b7724db54166f45aed982c7263. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/CenteredPage.svelte
+++ b/frontend/src/lib/components/CenteredPage.svelte
@@ -21,7 +21,7 @@
 </script>
 
 <div
-	class={twMerge('pb-8', wrapperClasses, handleOverflow ? 'h-full overflow-y-auto' : '')}
+	class={twMerge('pb-8', wrapperClasses, handleOverflow ? 'grow overflow-y-auto' : '')}
 	style={handleOverflow ? 'scrollbar-gutter: stable both-edges;' : ''}
 	{id}
 >
--- a/frontend/src/lib/components/common/table/Row.svelte
+++ b/frontend/src/lib/components/common/table/Row.svelte
@@ -158,7 +158,7 @@
 		clickToSelect ? 'cursor-pointer select-none' : '',
 		selected ? 'bg-surface-accent-selected' : keyboardSelected ? 'bg-gray-200 dark:bg-gray-700' : ''
 	)}
-	style={depth > 0 ? `padding-left: ${depth * 32}px;` : ''}
+	style={depth > 0 ? `padding-left: ${depth * 16}px;` : ''}
 	role={clickToSelect ? 'button' : undefined}
 	tabindex={clickToSelect ? 0 : undefined}
 	onclick={handleRowClick}


tusharui:

rubenfiszel:
Thanks for digging into this, @tusharui, and for spotting that `Row.svelte` and the folder header were out of sync.

While preparing the fix we measured the actual icon positions in the running app and found that `depth*16` doesn't fully line up: it overcorrects in the opposite direction, leaving files one level (16px) to the *left* of their sibling folders at every depth.

The reason is that a folder header indents its icon by `px-4` (16px) plus an inner `padding-left: depth*16`, so its effective indent is `(depth+1)*16`. But `Row.svelte`'s inline `padding-left` overrides the `px-4` entirely, so the file row needs to carry the full `(depth+1)*16` to match.

| depth | folder icon | file `depth*32` (old) | file `depth*16` (this PR) | file `(depth+1)*16` |
|---|---|---|---|---|
| 1 | 278 | 278 | 262 | 278 |
| 2 | 294 | 310 | 278 | 294 |
| 3 | 310 | 342 | 294 | 310 |

We've opened #10115 with `(depth+1)*16` and before/after screenshots verifying alignment at every depth. Closing this in favor of that, but genuinely appreciate you surfacing the root cause. 🙏

Issue #3687, opened 2024-05-07
bug: No scrolling in deploy to staging/prod screen


suppengrn:
### Describe the bug

When accessing the staging/prod screen via the link scrolling is not possible.

### To reproduce

1. Go to the staging screen 
![image](https://github.com/windmill-labs/windmill/assets/42936587/39ee90fe-b732-43d3-88e5-a624fdfbea2f)
2. Click on the "link" ![image](https://github.com/windmill-labs/windmill/assets/42936587/8a2fc4c2-b2c8-499b-9bbb-56ffff43c8cf)
3. The following screen scrolling is not possible therefore the "Deploy All" button is not accessible

### Expected behavior

When the diff list is too long scrolling should be possible

### Additional Context

reloading the site solves the problem

fatonramadani:
Hi, what browser are you using ? I wasn't able to reproduce, the deploy detail page scroll seems to work correctly

suppengrn:
Yes, I'm using Chrome (Version 123.0.6312.122)

wasnertobias:
Workaround: After opening and closing the Account settings overlay it works again!

wasnertobias:
Also no scrolling in the runs view now (initial run after first opening Windmill)
EE v1.342.0-10-g387ea4469
Chrome Version 125.0.6422.142

wasnertobias:
@rubenfiszel 
To reproduce the no scroll in runs view:

1. F5 on home Windill
2. Click "Import Flow"
3. Copy-paste any YAML, e.g. see below
4. Rename the flow to anything, e.g., "ABC"
5. Click deploy
6. now let it run with any input e.g., 2

=> You now cannot scroll down!

(If you F5 the page you can!)




```
summary: ""
description: ""
value:
  modules:
    - id: ap
      value:
        type: branchall
        branches:
          - expr: ""
            modules:
              - id: at
                value:
                  lock: ""
                  type: rawscript
                  content: |-
                    # import wmill


                    def main(x: str):
                        return x
                  language: python3
                  input_transforms:
                    x:
                      type: static
                      value: ""
                summary: Do stuff
            parallel: true
            skip_failure: false
          - expr: ""
            modules:
              - id: as
                value:
                  lock: ""
                  type: rawscript
                  content: |-
                    # import wmill


                    def main(x: str):
                        return x
                  language: python3
                  input_transforms:
                    x:
                      type: static
                      value: ""
                summary: Do stuff
            summary: ""
[...]

PR #9862, opened 2026-07-01
fix(deploy-screen): use flex-grow instead of h-full for scroll container

tusharui:
**Issue:** #3687

**Root cause:** CenteredPage used `h-full` (`height: 100%`) which fails to resolve during SvelteKit client-side navigation because the parent height comes from `flex-1` (a calculated value), not an explicit height. On full reload, all heights are pre-computed so it works.

**Fix:** Changed `h-full` → `grow` (`flex-grow: 1`) in `CenteredPage.svelte`. Uses flexbox's own sizing instead of CSS `height: 100%`, which is reliable in both initial load and client-side navigation.

---
## Summary by cubic
Fixes the Deploy screen scroll container during client-side navigation by switching from `h-full` to `grow` in `CenteredPage.svelte`. Uses flexbox sizing instead of `height: 100%` so the container height is reliable on both initial load and in-app navigation.

<sup>Written for commit e5d9f23577dc8a48be2eb034c4014f4ec7cd1858. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/CenteredPage.svelte
+++ b/frontend/src/lib/components/CenteredPage.svelte
@@ -21,7 +21,7 @@
 </script>
 
 <div
-	class={twMerge('pb-8', wrapperClasses, handleOverflow ? 'h-full overflow-y-auto' : '')}
+	class={twMerge('pb-8', wrapperClasses, handleOverflow ? 'grow overflow-y-auto' : '')}
 	style={handleOverflow ? 'scrollbar-gutter: stable both-edges;' : ''}
 	{id}
 >


rubenfiszel:
Thanks for digging into this, and sorry for the delayed review.

We tested this change carefully and unfortunately it regresses the scroll container rather than fixing it, so we're going to close it. Sharing the full findings in case they're useful:

**Why `grow` breaks it.** `CenteredPage`'s scroll `<div>` is rendered into the content wrapper at `AiChatLayout.svelte` (`<div class="flex-1 min-h-0">`), which computes to `display: block` (it's a flex *item* of its own parent, but not a flex *container*). `flex-grow` only affects flex *items*, so on a block parent the class does nothing and the element falls back to `height: auto` (content height). The `overflow-y-auto` clamp then never engages.

Measured on `/workers` with content taller than the viewport (viewport 500px, content 571px):

| Variant | Full reload | Client-side nav |
|---|---|---|
| `h-full` (current main) | height 500px, scrolls | height 500px, scrolls |
| `grow` (this PR) | height **571px**, cannot scroll | height 571px, cannot scroll |

So `grow` breaks scrolling for every `CenteredPage` consumer (workers, variables, resources, schedules, all triggers pages, deploy, ...), and it breaks even on a full reload, not just client-side navigation.

**On the original issue (#3687).** That report is from 2024 (EE v1.333.x) and predates the current layout. The content wrapper has since been rewritten and now threads `min-h-0` through the whole flex chain, which gives `h-full`'s ancestor a definite height to resolve against. We could not reproduce the broken-scroll behavior on current main in either full reload or client-side navigation, so `h-full` is doing the right thing today.

Closing as the current behavior is already correct. Really appreciate the contribution and the clear write-up regardless.

PR #9864, opened 2026-07-01
feat: support workspace forks on cloud using parent workspace limits

hugocasa:
## Summary

Workspace forking was hard-blocked on `app.windmill.dev` (the three fork/dev-create handlers errored on `*CLOUD_HOSTED`), because a fork would otherwise get its own independent free-tier quota and its own (missing) subscription. This PR lets forks work on cloud by **piggybacking on the parent's plan**: a fork/dev workspace inherits its root (billing) workspace's premium status, and its executions are metered into the **root's** usage row so they draw from the same pool and count toward the root's usage-based bill. Abuse is bounded by a **per-developer-seat fork cap** (default 5), and forking is **premium-only** (free workspaces get 0 forks).

All new backend logic is `#[cfg(feature = "cloud")]`-gated. For a normal (non-fork) workspace the billing root resolves to itself, so there is **zero behavior change off-fork**.

## Changes

**Backend — `windmill-common/src/workspaces.rs`**
- `get_billing_workspace_id()` — resolves a workspace to its root ancestor via one depth-guarded, cycle-safe recursive CTE; cached (60s TTL) with `invalidate_billing_workspace_cache()`.
- `get_team_plan_status()` now resolves to the billing root and keys `TEAM_PLAN_CACHE` by root id, so every plan-status consumer (push limits, job timeout, `is_premium` UI, usage tracking) inherits the parent's plan.
- `count_paid_seats()` (`ceil(devs + operators/2)`, excluding disabled/service-account members) and `count_workspace_forks()` (non-deleted descendants of a root).

**Backend — `windmill-queue/src/jobs.rs`**
- `push_inner` and `apply_completed_job_cloud_usage` resolve `billing_w_id` and attribute workspace usage (job count + execution-seconds) to the root.

**Backend — `windmill-api-workspaces/src/workspaces.rs`**
- The three cloud hard-blocks are replaced: `enforce_cloud_fork_cap()` (premium-required + `existing_forks < seats × MAX_FORKS_PER_SEAT`) gates `create_workspace_fork` and `create_workspace_fork_branch`; `attach_dev_workspace` requires the prod workspace be premium.
- New env `MAX_FORKS_PER_SEAT` (default 5).

**Frontend** — the three fork entry points hidden with `!isCloudHosted()` (`WorkspaceMenu`, `SessionForkBar`, `WorkspaceFamilyPicker`) now show on cloud when premium (`(!isCloudHosted() || \$isPremiumStore)`). `isPremiumStore` is fed by `is_premium`, which now reflects the root's plan on a fork.

**Tests** — `windmill-common/tests/billing_workspace.rs`: root/fork/grandchild/unknown resolution, cycle survival, seat weighting (disabled/service-account exclusion), and fork counting (deleted/grandchild/standalone).

## Premium changes & downgrade propagation

Premium is toggled by **direct SQL** on the root workspace (the Stripe webhook flow does `UPDATE workspace SET premium = …`), not through app code. This still propagates to forks correctly because invalidation rides a **DB trigger**, not an app-level `pg_notify`:

- `workspace_premium_change_trigger` is `AFTER UPDATE OF premium ON workspace` and emits the changed workspace's id (the root, since only the root is ever subscribed). Past-due changes emit the same channel via `cloud_workspace_settings`.
- Because `TEAM_PLAN_CACHE` is now keyed by the resolved **root** id, that trigger invalidates the single entry shared by the root and all its forks. On a fork's next job push it re-resolves to the root, reads the new premium value, and the fork immediately follows the root's plan. After a downgrade the fork's **monthly** free-execs quota is shared with the root (usage meters to the root); the per-workspace burst guards (in-queue depth, concurrency) stay per-workspace but are backstopped by the global per-user in-queue/concurrency caps.
- Downgrade doesn't change `parent_workspace_id`, so the fork→root mapping stays valid — nothing else to invalidate.
- Fork usage lands in the root's existing `usage` row, so it's billed to the customer with no double-counting; seat counting reads the root's own `usr` members (a dev workspace's copied members live under the dev's id, not the root's), so forks don't inflate seats.

**Deliberate:** existing forks are **not** auto-removed when a workspace downgrades — they keep working under the root's now-free monthly quota (shared) and their own per-workspace burst guards, and creating new forks is blocked. No free-compute hole, since everything meters to the (now free) root.

## Screenshots

On cloud, a non-admin developer sees the fork entry in a premium workspace (visibility fix; on non-cloud builds this is a no-op since `isCloudHosted()` is false):

workspace-menu-fork

A fork's General → Cloud Quotas shows the per-seat fork budget (here 6 / 10 = 2 seats × 5), resolved to the billing root:

cloud-quotas-forks-row

A fork's Premium plans tab points billing to the parent instead of showing a standalone subscription:

fork-billing-note

When a fork create is rejected (here the per-seat cap), the error surfaces and the button resets instead of staying stuck loading:

fork-cap-error

## Test plan

- [x] `update_sqlx.sh` — full workspace compiles with all cloud/EE features
- [x] `cargo test -p windmill-common --features cloud --test billing_workspace` — 3/3 pass
- [x] `npm run check:fast` (svelte-check) — no problems
- [x] On a `cloud`-feature build with `CLOUD_HOSTED=true` and a premium root: fork menu + dev-session fork bar appear; a fork can be created; from a non-premium workspace the gates stay hidden and `create_workspace_fork` returns the "requires a paid team plan" error
- [x] Create forks up to `MAX_FORKS_PER_SEAT × seats`; confirm the next is rejected and that deleting a fork frees a slot
- [ ] A job run inside a fork increments the **root's** `usage` row for the current month, not the fork's
- [ ] Downgrade a root with forks (`UPDATE workspace SET premium=false`) and confirm the fork's jobs fall under the root's free-tier limits within a poll interval

- The fork-cap check is best-effort (check-then-act, not transactionally locked). It's a soft guardrail on a rare admin action; a concurrent race could overshoot by a fork or two, which is harmless since usage still meters to the parent regardless. Can be tightened later with an advisory lock keyed on the root if strictness is needed.
- A fork's billing settings page still reads its own `premium`/`customer_id` (`stripe_ee.rs premium_info`); enforcement doesn't depend on this and it can redirect to the parent later.


hugocasa:
Addressed the review findings in `c498d17`:

- **P1 (Claude) — fork UI hidden from non-admin developers on cloud.** `is_premium` is no longer `require_admin`-gated (it returns a single boolean; the extractor still enforces workspace membership), and `+layout.svelte` now populates `isPremiumStore` for all cloud members, not just admins. The fetch is wrapped in try/catch so it can't block user-store init. Non-admin devs on a premium workspace now see the fork entry points, matching what `enforce_cloud_fork_cap` (which doesn't require admin) already allowed.
- **P1 (Codex) — `attach_dev_workspace` bypassed the cap.** Attach now runs the full `enforce_cloud_fork_cap` (premium + per-seat count) instead of a premium-only check, so it also blocks when the family is at cap or `MAX_FORKS_PER_SEAT=0`. This matches the decision to count dev workspaces toward the cap.
- **P2 (Claude) — `count_workspace_forks` pruned live sub-forks under a soft-deleted intermediate.** Moved the `deleted` filter from the recursive step to the outer `SELECT` so live descendants of a deleted node still count; added a test case asserting this.
- **P1 (Codex) — auth contract on new `pub` helpers.** Added doc comments to `get_billing_workspace_id`, `count_workspace_forks`, and `count_paid_seats` stating they are unauthenticated metering helpers whose callers must already be authorized (ids are server-resolved, not user input).

**Left as-is (conscious calls, both already in "Known limitations"):**
- **Fork-cap check-then-act race (Codex P1 / Claude nit).** It's a soft guardrail: the `MAX_FORKS_PER_SEAT=0` and empty cases are handled by the pre-check, and a concurrent race can only overshoot by a fork or two, all metered to the parent regardless. Pi flagged the PR good-to-merge and Claude called it harmless. Can add a per-root advisory lock in the creation transaction later if a hard cap is ever required.
- **A fork's billing settings page still reads its own `premium`/`customer_id`** — enforcement never depends on it.

Validation: `update_sqlx.sh` (full cloud/EE build) green, `cargo test -p windmill-common --features cloud` 3/3, `svelte-check` clean.

hugocasa:
Second round (Codex) addressed in `db260a2`:

- **P1 — stale billing root after `change_workspace_id`.** The child re-parent loop now also calls `invalidate_billing_workspace_cache(child)` alongside the existing fork-parent routing invalidation, so a renamed root's forks re-resolve their billing root immediately instead of metering/plan-checking through the archived old id for up to the 60s TTL. (Deeper fork-of-fork descendants, which aren't in the direct re-parent set, self-heal via the same 60s TTL — noted in a comment.)
- **P2 — fork usage display vs enforcement mismatch.** `get_usage` now resolves to the billing root on cloud, so the free-execs indicator reports the same workspace the backend meters/enforces against (off-fork it resolves to the workspace itself).
- **P2 — attach cap rejected a zero-new-fork re-designation.** Split the cloud guard into `require_cloud_fork_premium` (always) + `enforce_cloud_fork_count`. `attach_dev_workspace` now only enforces the count when the candidate isn't already under the same root; re-designating an existing fork as the dev workspace (which doesn't increase the descendant count) is allowed even at cap. `create_workspace_fork`/`create_workspace_fork_branch` still run the full premium+count `enforce_cloud_fork_cap`.

Validation: `cargo check -p windmill-api-workspaces` passes for both `--features cloud` and default (non-cloud); no new SQL so the offline cache is unchanged; `billing_workspace` tests still green. The three fixes are API/queue-handler wiring over already-unit-tested helpers, so they rely on the existing helper coverage + manual verification (per the earlier test-coverage discussion).

hugocasa:
Addressed the outstanding Codex P1 from the last review round (it was **not** fixed by the prior commits) in `bedeed11d0`:

- **P1 — permanent fork deletion didn't clear `BILLING_WORKSPACE_CACHE`.** Since fork ids are reusable after a hard delete, a cached `fork -> old_root` mapping could survive the 60s TTL and mis-meter a recreated id. `delete_workspace` now invalidates both the billing (`invalidate_billing_workspace_cache`) and fork-parent routing (`invalidate_fork_parent_cache`) caches for the deleted id **and** any child forks orphaned by the `ON DELETE SET NULL` FK. Deeper (grandchild) descendants self-heal via the 60s TTL.

Also added a `sqlx::test` (`billing_cache_invalidation_reflects_reparent`) covering the cache→reparent→invalidate→re-resolve contract that the delete/attach/rename paths rely on (the coverage gap Codex noted).

Validation: `cargo check -p windmill-api-workspaces` passes cloud + default; the new delete query was already in the offline cache (no `.sqlx` change); `billing_workspace` tests now 4/4.

hugocasa:
Thanks — addressed in `7bc623341a`:

**Codex P1 — attach-dev only reserved one fork slot for a candidate subtree.** Fixed. `enforce_cloud_fork_count` now takes an `incoming` count and rejects when `existing + incoming > allowed`. `attach_dev_workspace` computes `incoming = 1 + count_workspace_forks(candidate)`, so attaching a standalone workspace that already has child forks reserves slots for its whole subtree (create still passes `incoming = 1`). The `count_workspace_forks` subtree count is already covered by `paid_seats_and_fork_count`; the handler wiring stays code-review-verified like the other cloud-gated handlers.

**Claude P2 — downgraded-fork queue/concurrency burst guards count the child, not the root.** Left as-is by design, with a clarifying comment added at the guards. Rationale: keying those two `v2_job_queue` counts to `billing_w_id` would count the *root's* own queue (which doesn't contain the fork's jobs), so it's neither the fork's budget nor a true family cap — a real family-shared burst cap would need a family-wide subquery on the hot push path. And the abuse case is already backstopped: the **per-user** in-queue/concurrent guards just above are keyed by `permissioned_as_email` and are **global across the whole family**, so a single user can't exceed `MAX_FREE_CONCURRENT_RUNS` by spreading across forks; N distinct users across N forks is equivalent to N free workspaces. The monthly execs cap *is* shared (metered to the root). I tightened the wording so the PR doesn't overclaim: monthly usage is family-shared; the per-workspace burst guards remain per-workspace (backstopped by the global per-user caps).

Validation: `cargo check -p windmill-api-workspaces` passes cloud + default; no new SQL (reuses `count_workspace_forks`), so no `.sqlx` change.

PR #9867, opened 2026-07-01
fix(jobs): give flow dynselect a path and its worker tag, aligned with scripts

rubenfiszel:
## Summary

`Fixes WIN-2118`

Fetching options for a `dynselect` / `dynmultiselect` input was inconsistent between deployed scripts and deployed flows:

| | before |
|---|---|
| deployed **script** | `script` job, **with a path**, on the **script's tag** (via `push_script_job_by_path_into_queue`) |
| deployed **flow** | anonymous `preview`, **no path**, always on the **language-default tag**, access failures surfaced as a raw `SqlErr: no rows` |

This PR takes the **low-risk** path: **deployed scripts are left exactly as they were** (that path already handles tag / lock / codebase / on-behalf-of correctly — no reason to re-derive all of it). Only the flow branch is brought into line.

> `dynmultiselect` and `dynselect` are the same code path (same `/jobs/run/dynamic_select` endpoint, same `DynamicInput` component), so everything here applies identically to both.

## Changes (`backend/windmill-api/src/jobs.rs`, `run_dynamic_select`)

- **Flow branch**: reads the flow's `tag` under RLS and routes the option-fetch preview to it (falling back to the language default when unset) and carries the flow **path** on the job — so it lines up with the script run's path + worker group. The tag read runs on every request (including cache hits), so it doubles as the per-request **access check**, replacing the raw `SqlErr` with a clean `NotAuthorized` / `NotFound`.
- **Flow tag authorization**: the flow's tag now goes through `check_tag_available_for_workspace` before `push()` — the same gate a normal flow run (`run_flow`) and the script path apply — so a caller who can read the flow but isn't allowed to use its (custom/scoped) worker tag is rejected consistently.
- **Script branch**: unchanged behaviour (still `push_script_job_by_path_into_queue`); only cosmetic (`ENTRYPOINT_OVERRIDE` const instead of a string literal, dropped needless `.clone()`s).
- **Inline**: unchanged — a `preview` with no path on the language default, blocked for operators.
- Entrypoint-name validation now runs for **all** branches (it's interpolated into the generated wrapper), closing that gap for the flow/inline paths.

### Worker-tag rules after this PR

| Source | Worker tag | Tag authorization checked? |
|---|---|---|
| Deployed **script** | script's `tag` (unset → language default, `//native` → `nativets`) | yes (inside `push_script_job_by_path_into_queue`) |
| Deployed **flow** | flow's `tag` (unset → language default) | yes (`check_tag_available_for_workspace`, new) |
| **Inline** | language default | n/a (tag is `None`) |

### What is and isn't aligned
- **Aligned**: both now carry a **path**, route to their **worker tag**, gate that tag, and work for **operators**.
- **Not aligned (by design, for safety)**: scripts remain a `script` **run** and flows remain a `preview`. A flow's dyn-select code lives in the flow schema (there's no deployed runnable to "run"), so it can only be a preview — and turning the script into a preview to match would mean re-deriving the whole deployed-run path (hash/codebase, lock, on_behalf_of, delete_after_use, …), which is exactly the fragile surface this PR avoids.

## Operators

Operators (and scoped tokens) can fetch options for every flow/script they can actually **run**:
- untagged flows/scripts and flows tagged with an **allowed** custom tag → work;
- a flow tagged with a worker tag the caller isn't permitted to use → rejected — **identical to a normal run of that flow** (verified: the normal run endpoint returns the same tag error), so the caller couldn't run it anyway;
- inline remains blocked for operators (request-supplied code).

Note: before this PR, an operator could fetch options for a restricted-tag flow only because the helper ran on the language-default tag (ignoring `flow.tag`); routing to the flow's tag (the requested behavior) makes it consistent with — and gated like — a real run.

## Test plan (verified e2e locally)

- [x] **Script** dyn-select → `kind=script` (unchanged run), correct options
- [x] **Flow** dyn-select → `kind=preview`, `runnable_path` set, correct options
- [x] Flow tagged `deno` → job `tag=deno`; flow tagged `chromium` → queued on `chromium`; untagged flow/inline → language default (`bun`)
- [x] **Operator**: shared script + untagged/allowed-tag flow work; inaccessible flow → `NotAuthorized`; inline → blocked
- [x] **Tag authorization**: non-super-admin scoped `if_jobs:filter_tags:othertag` on a `deno`-tagged flow → `Tag deno is not available in your scope`; same rejection as a normal flow run (super-admins & untagged unaffected)
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api --features enterprise,private` passes
- [x] UI: option dropdown renders in both the script and flow run forms

## Screenshots

Script run form (dynselect renders Alpha/Beta):

script-dynselect-open

Flow run form (dynselect renders FlowAlpha/FlowBeta):

flow-dynselect-open

### DIFF
--- a/backend/.sqlx/query-919bee5afcaba361f23760b5f64ce48f63e4d6d3e2d4fa9247b50963112f47ad.json
+++ b/backend/.sqlx/query-919bee5afcaba361f23760b5f64ce48f63e4d6d3e2d4fa9247b50963112f47ad.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT tag FROM flow WHERE workspace_id = $1 AND path = $2",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "tag",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      true
+    ]
+  },
+  "hash": "919bee5afcaba361f23760b5f64ce48f63e4d6d3e2d4fa9247b50963112f47ad"
+}
--- a/backend/.sqlx/query-ad07e40790b947d1e8da17dd4ae30365210bba4bff0b7e32d06c53ed13572360.json
+++ b/backend/.sqlx/query-ad07e40790b947d1e8da17dd4ae30365210bba4bff0b7e32d06c53ed13572360.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT 1 FROM flow WHERE workspace_id = $1 AND path = $2 LIMIT 1",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "?column?",
+        "type_info": "Int4"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "ad07e40790b947d1e8da17dd4ae30365210bba4bff0b7e32d06c53ed13572360"
+}
--- a/backend/windmill-api/src/jobs.rs
+++ b/backend/windmill-api/src/jobs.rs
@@ -7602,26 +7602,32 @@ async fn run_dynamic_select(
         ));
     }
 
+    if !is_valid_entrypoint_name(&request.entrypoint_function) {
+        return Err(error::Error::BadRequest(format!(
+            "Invalid entrypoint_function {:?}: must match \
+             ^[A-Za-z_][A-Za-z0-9_]*$ (letters, digits and underscores, \
+             not starting with a digit)",
+            request.entrypoint_function
+        )));
+    }
+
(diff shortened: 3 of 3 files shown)


rubenfiszel:
Addressed the CI reviews (now on `29034bc`):

- **Codex P1 — flow dyn-select bypassed worker-tag authorization**: fixed. The flow branch now calls `check_tag_available_for_workspace(&db, &w_id, &tag, &authed)` before `push()` — the same gate `run_flow` and `push_script_job_by_path_into_queue` apply. Verified e2e: a non-super-admin token scoped `if_jobs:filter_tags:othertag` fetching options on a `deno`-tagged flow is now rejected with `Tag deno is not available in your scope` (previously allowed); super-admins and untagged flows are unaffected.
- **Codex/Claude P1 — codebase/bundle regression**: no longer applies. The script branch is unchanged (still `push_script_job_by_path_into_queue`, `script`-kind job), so codebase/bundle scripts keep their deployed-hash resolution. Pi's latest review confirms this.
- **Pi P2 — description contradicted the code**: the PR description was rewritten with the force-push; it now states scripts keep their `script`-kind run and only flows are brought into line (path + tag).

`SQLX_OFFLINE=true cargo check -p windmill-api --features enterprise,private` passes; no new queries were added by the tag-availability fix.

PR #9879, opened 2026-07-02
feat(ai): add Azure AI Foundry as a native AI provider

hugocasa:
## Summary

Adds **Azure AI Foundry** (`azure_foundry`) as a native AI provider, wired through the AI chat (copilot) and AI agent flow steps. Previously, routing AI traffic through Foundry meant using the *Azure OpenAI* provider, which is limited to the OpenAI model catalog. Azure Foundry exposes a broader catalog (OpenAI models plus Llama, DeepSeek, Mistral, Phi, etc.) over an OpenAI-compatible chat completions API.

Foundry uses the same Azure conventions as Azure OpenAI (the `api-key` header and Azure-style URL building), so it reuses the existing OpenAI-compatible query builder and proxy path. The backend helper `is_azure_openai` was renamed to `is_azure` to reflect that it now covers both Azure OpenAI and Azure Foundry.

Fixes WIN-2122

## Companion PR

- **windmill-labs/windmill-integrations#161** — adds the `azure_foundry` resource type (hub). Required for the provider to be selectable.

## Changes

**Backend (`windmill-ai`)**
- New `AzureFoundry` variant on the `AIProvider` enum (serde `azure_foundry`).
- `get_base_url` requires a resource-provided base URL (same as Azure OpenAI / Custom AI).
- `is_azure()` (renamed from `is_azure_openai`) matches Azure OpenAI + Azure Foundry, driving `api-key` auth and Azure URL construction (`/openai` → `/openai/v1`).
- Added to `supports_openai_compatible_proxy` and `proxy_execution_mode` (`HttpForward`).
- Routes through the existing `OtherQueryBuilder` (no new builder). New `builds_azure_foundry_proxy_request` unit test.

**Frontend (copilot)**
- New `azure_foundry` entry in `AI_PROVIDERS` (label "Azure AI Foundry", default models spanning the broader catalog) and in the exhaustive `PROVIDER_COMPLETION_CONFIG_MAP`.
- Handled alongside `azure_openai` for `max_completion_tokens`/token limits, streamed usage tracking, tool-schema strict-mode, and the reasoning registry. These are all model-id-gated (`gpt-5`/o-series), so they are a no-op for Foundry's non-OpenAI models.
- Treated as a **chat-completions** provider — deliberately *not* routed through the OpenAI Responses API (which only `openai`/`azure_openai` use), so non-OpenAI Foundry models work.

**OpenAPI**
- `azure_foundry` added to `AIProvider` (`openapi.yaml`) and `AIProviderKind` (`openflow.openapi.yaml`); regenerated CLI guidance (`system_prompts/generate.py`). The gitignored frontend client regenerates at build.

## Both entry points

- **AI chat (copilot):** frontend → `/ai/proxy/*` → `HttpForward` → `build_openai_compatible_proxy_request` → `is_azure()` → Azure URL + `api-key`.
- **AI agent steps (flow):** worker `ai_executor.rs` → `to_provider_credentials` → `create_query_builder` → `OtherQueryBuilder` → `get_endpoint`/`get_auth_headers` (both call `is_azure`). Same core logic as the chat path.

## Test plan

- [x] `cargo test -p windmill-ai` — all 64 tests pass, incl. new Azure Foundry proxy URL test
- [x] `cargo check -p windmill-api -p windmill-worker`
- [x] `npm run check` (frontend) — 0 errors; copilot unit tests (lib/chatLoop/reasoningRegistry) pass
- [x] E2E via Playwright: "Azure AI Foundry" appears in workspace AI settings, its `azure_foundry` resource type resolves in the ResourcePicker, default models populate (see screenshot)
- [x] Verify a live request against a real Azure AI Foundry endpoint (requires a Foundry resource + credentials)

## Notes

Reasoning-effort support is enabled only for OpenAI-family model ids on Foundry; other Foundry-hosted reasoning models can be added to the registry later if needed. AAD / managed-identity auth (beyond api-key) can follow via an OAuth resource type, as Azure OpenAI already does.

## Screenshots

azure-foundry-ai-settings


hugocasa:
### Live verification ✅

Tested end-to-end against a real Azure AI Foundry resource (both code paths):

| Path | Endpoint | Result |
|---|---|---|
| **AI chat** | `POST /ai/proxy/chat/completions` (proxy) | `200` — `gpt-4o-2024-11-20` returned `"Pong"` |
| **AI chat** | `GET /ai/proxy/models` (proxy) | `200` — returns the Foundry model catalog |
| **AI agent step** | flow `aiagent` module (worker executor) | `200` — `output: "PONG"`, usage `{input:15, output:3, total:18}` |

Both confirm the Azure URL construction (`.../openai/v1/chat/completions`) and `api-key` auth against the live endpoint.

**Endpoint format gotcha:** the resource `base_url` must be the OpenAI-compatible endpoint (`https://<resource>.services.ai.azure.com/openai/v1`), **not** the Foundry *project* endpoint (`.../api/projects/<project>`) that the portal surfaces by default — the latter requires an `api-version` query param and returns `Missing required query parameter: api-version`. The resource type field description (windmill-labs/windmill-integrations#161) now calls this out explicitly.

hugocasa:
### Why a separate provider (not just Azure OpenAI) — verified

Confirmed live that `azure_foundry` is **not** redundant with the existing `azure_openai` provider:

- Foundry's `https://<resource>.services.ai.azure.com/openai/v1` endpoint is the **unified broad-catalog** inference endpoint — it serves OpenAI **and** non-OpenAI models over OpenAI-compatible chat completions. Verified live: `gpt-4o`, `gpt-5.5-2026-04-24`, and **`deepseek-v4-flash`** all return `200` (chat + agent-step paths).
- The `azure_openai` provider **can't** serve that catalog: it routes through the **Responses API** (which non-OpenAI models don't support) and filters the model list to gpt-/o-/codex. `azure_foundry` deliberately uses **chat completions** (universal) and an unfiltered model list — which is the whole point.

So the split is: `azure_openai` = OpenAI models via Responses API; `azure_foundry` = the full Foundry catalog via chat completions.

**Known limitation:** the reasoning-effort UI under `azure_foundry` is gated to OpenAI-family model ids (gpt-5/o-series). Non-OpenAI reasoning models (DeepSeek, Claude) run fine but don't get a tunable effort knob yet — a possible follow-up.

hugocasa:
Good catch on the P1 — fixed in 9fc29f3679.

`azure_foundry` is now in the tracked copilot flow Zod schema (`openFlowZod.gen.ts`, plus its `openFlow.json` source), inserted after `azure_openai` to match `openflow.openapi.yaml`. I applied it surgically to preserve the generated single-line format (a naive edit tripped the formatter and reformatted the whole file). Added a focused regression test — `chat/flow/providerKind.test.ts` — that runs an `aiagent` module through `flowModulesSchema` for `azure_foundry` (and asserts an unknown kind is still rejected); verified it fails without the schema fix and passes with it.

I also swept every tracked file that mentions `azure_openai` for other stale enums. Findings:
- `providers/other.rs` — just the `build_azure_openai_url` function name, not an enum.
- `chatLoop.ts` — the `isOpenAI` Responses-API branch; `azure_foundry` is intentionally excluded there (it routes through chat completions).
- `openapi-deref.{yaml,json}` — the served API-docs spec (`include_str!`, not runtime validation). It's already stale on main (`1.740.0` vs `1.744.0`) and regenerated separately, so I've left it per the usual deref convention rather than hand-editing a generated bundle.

PR #9883, opened 2026-07-02
fix: polish pipeline graph view (layout, viewport, minimap, lineage, timestamps)

rubenfiszel:
## Summary

Five independent fixes from a pipeline dogfooding session, one commit each, covering the graph view (`/pipeline/<folder>`, PipelineGraphEditor/AssetGraph), the SQL asset parser, and DuckDB result serialization.

## Changes

1. **Layout: keep the graph layered when lineage has cycles** (`assetGraphLayout.ts`, `AssetGraphCanvas.svelte`). An `rw` asset access produces a write edge (script → asset) *and* a read edge (asset → script) — a 2-cycle. `layoutComponent` threw on any cycle and the whole graph fell back to an unordered grid: crossing edges everywhere, assets above their producers, consumers fanning out backwards. Now write⇄read 2-cycles are resolved producer-above-asset by omitting the backward direction from the layout input only (both arrows still render), and any longer cycle is absorbed in the Kahn step by dropping feedback edges instead of throwing. The grid fallback remains as a defensive catch only.
2. **Initial viewport: fit the graph to the visible canvas** (`InitialFitView.svelte`). There was no initial fit at all — content was anchored at zoom 1, so anything wider/taller than the pane was clipped and unclickable until manual Fit View. A one-shot `fitView({ maxZoom: 1 })` fires once the first real node set is measured (the graph arrives async after mount, so SvelteFlow's init-time `fitView` prop would fit the wrong content). The fit re-arms on folder switch, keyed on the folder whose graph is actually *loaded* — `graphRes.current` is stale-while-revalidate, so keying on the route param would consume the new folder's fit on the old graph (caught by local review).
3. **Minimap: make it read as a minimap** (`AssetGraphCanvas.svelte`). Flat pastel bars with no strokes read as a loading skeleton. Node hues now mirror the canvas (blue assets, amber triggers, bordered neutral scripts) with visible strokes and rounded corners, plus a bordered container and an outlined viewport mask.
4. **Parser: don't infer s3 reads from bare string-literal mentions** (`windmill-parser-sql-asset`). A trailing summary `SELECT 's3:///shop/raw/orders.csv' AS target` after a `COPY … TO` of the same path turned the write into `rw`, drawing a false asset⇄script cycle. Bare literals in generic query position are now *weak* read evidence, surfaced only when the script has no other usage of that asset. FROM-position literals and `read_csv`/`read_parquet`/`read_json` arguments are recorded as explicit strong reads, so genuine self-refresh `rw` patterns are preserved (covered by tests). Note: the fix lands in live-buffer inference (`windmill-parser-wasm-asset`) on its next wasm publish; the deployed-graph path takes effect immediately.
5. **DuckDB: render temporal values as ISO strings in job results** (`windmill-duckdb-ffi-internal`). `TIMESTAMP` columns (e.g. SCD2 `valid_from`/`valid_to`) serialized as raw epoch-micros strings like `"1782974022218435"`; `DATE` as raw day counts and `TIME` as raw micros. All three now render ISO (`2026-07-01T23:13:42.218435`, `2026-07-01`, `10:30:00`), matching the Postgres executor's shape, with raw-count fallback for out-of-range values.

## Test plan

- [x] `cargo test -p windmill-parser-sql-asset` — 69 pass (4 new: write-echo stays W, bare mention alone stays R, read-fn + COPY and FROM-literal + COPY stay RW)
- [x] `cargo test` in `windmill-duckdb-ffi-internal` — 5 pass, incl. an integration test through a real in-memory DuckDB connection (TIMESTAMP/TIMESTAMPTZ/TIMESTAMP_NS/DATE/TIME)
- [x] `npx vitest run assetGraphLayout.test.ts` — 10 pass (3 new cycle tests)
- [x] Manual (Playwright, seeded `pipelines` workspace with `f/shop` + `f/analytics` duckdb pipelines): reproduced the grid-fallback chaos from a deployed script with the spurious `rw`; after the fixes the pipeline renders as a straight top-to-bottom chain, fully fitted on load, and re-deploying the ingest script yields a plain `w` edge (no cycle)
- [x] Manual: in-app folder switch shop → analytics refits the viewport to the new graph (all nodes inside the visible canvas, verified via bounding rects)
- [ ] Run a DuckDB job on a duckdb-feature worker and confirm ISO timestamps in the result viewer (dev worker here is built without the `duckdb` feature; covered by the in-memory integration test)

## Screenshots

Before (grid fallback from the spurious rw cycle; skeleton-looking minimap; content clipped at zoom 1):

before-shop-layout

After (same pipeline: layered chain, fitted viewport, restyled minimap, no spurious read edge):

after-all-shop

View mode, second pipeline:

after-analytics

Edit mode (+ anchor, per-asset actions):

after-editmode

Minimap closeup:

minimap-closeup

After in-app folder switch (viewport refit on the newly loaded graph):

after-folder-switch

---
## Summary by cubic
Polishes the pipeline graph so it loads fitted, stays layered even with lineage cycles, and shows a clearer minimap. Also fixes SQL asset inference and renders DuckDB temporal values as ISO strings.

- **New Features**
  - Auto fit-to-view on first render and when the loaded folder changes, keyed to the graph that actually arrived (`InitialFitView.svelte`).
  - Minimap restyled with node hues, strokes, rounded corners, and a clear viewport mask.

- **Bug Fixes**
  - Graph layout remains layered when cycles exist by dropping feedback edges during layout; `rw` 2-cycles resolve producer-above-asset while still rendering both arrows.
  - SQL asset parser (`windmill-parser-sql-asset`): bare string-literal mentions are treated as weak reads and only surfaced if no other usage; FROM-position literals and `read_csv`/`read_parquet`/`read_json` arguments are strong reads, including list and named arguments, so genuine self-refresh `rw` patterns are preserved.
  - DuckDB result JSON (`windmill-duckdb-ffi-internal`): `TIMESTAMP`, `DATE`, and `TIME` serialize as ISO strings, with raw-count fallback for out-of-range values.

<sup>Written for commit 0283cbcb04ce5f73ea12575e3e9e2e40fca5d00f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/parsers/windmill-parser-sql-asset/src/asset_parser.rs
+++ b/backend/parsers/windmill-parser-sql-asset/src/asset_parser.rs
@@ -745,12 +745,15 @@ impl Visitor for AssetCollector {
                     if let Some(asset) = self.get_s3_asset_from_str_literal_table(table_factor) {
                         self.assets.push(asset);
                     }
-                    // Avoid Table Functions
-                    self.handle_obj_name_pre(name);
-                } else if let Some(asset) = self.get_s3_asset_from_table_function(table_factor) {
-                    // read_csv/read_parquet/read_json('s3:///…') likewise.
-                    self.assets.push(asset);
                 }
+                // For a read-function table factor this pushes ReadFn, making
+                // every literal inside its arguments a definitive read — the
+                // direct form (read_csv('s3:///…')) but also list and named
+                // arguments (read_parquet(['s3:///…'])). Must run for BOTH the
+                // plain-table and table-function branches: post_visit_table_factor
+                // pops via handle_obj_name_post unconditionally, so skipping the
+                // push here would unbalance the stack.
+                self.handle_obj_name_pre(name);
             }
             _ => {}
         }
@@ -1216,6 +1219,51 @@ mod tests {
                 access_type: Some(RW),
                 columns: None
             }])
+        );
+    }
+
+    #[test]
+    fn test_read_fn_list_arg_plus_copy_stays_rw() {
+        // read_parquet's list form is as definitive as the direct literal —
+        // it must not be demoted to a weak mention when the file is rewritten.
+        let input = r#"
+            CREATE TABLE tmp AS SELECT * FROM read_parquet(['s3:///data.parquet']);
+            COPY (SELECT * FROM tmp) TO 's3:///data.parquet';
+        "#;
+        let s = parse_assets(input).map(|s| s.assets);
+        assert_eq!(
+            s.map_err(|e| e.to_string()),
+            Ok(vec![ParseAssetsResult {
+                kind: AssetKind::S3Object,
+                path: "/data.parquet".to_string(),
+                access_type: Some(RW),
+                columns: None
+            }])
+        );
+    }
+
+    #[test]
+    fn test_read_fn_list_arg_multiple_files_are_reads() {
+        let input = r#"
+            SELECT * FROM read_parquet(['s3:///a.parquet', 's3:///b.parquet']);
+        "#;
+        let s = parse_assets(input).map(|s| s.assets);
+        assert_eq!(
+            s.map_err(|e| e.to_string()),
+            Ok(vec![
+                ParseAssetsResult {
+                    kind: AssetKind::S3Object,
+                    path: "/a.parquet".to_string(),
+                    access_type: Some(R),
+                    columns: None
+                },
+                ParseAssetsResult {
+                    kind: AssetKind::S3Object,
+                    path: "/b.parquet".to_string(),
+                    access_type: Some(R),
+                    columns: None
+                }
+            ])
         );
     }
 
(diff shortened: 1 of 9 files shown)


rubenfiszel:
Addressed the Codex P1 in 0283cbcb04: `pre_visit_table_factor` now pushes the `ReadFn` context for read-function table factors instead of only matching a direct first-arg string literal, so list and named argument forms (`read_parquet(['s3:///data.parquet'])`) are definitive reads and a self-refresh with a same-path `COPY … TO` keeps its `rw`. This also balances the pre-existing unmatched pop in `post_visit_table_factor` that the Pi review noted. Added two regression tests (list-arg + COPY stays RW; multi-file list yields both reads) — 71/71 pass.

PR #9885, opened 2026-07-02
feat(pipeline): backfill a range of partitions from the asset drawer

rubenfiszel:
## Summary

Makes the asset drawer's Backfill button real. On a `// partitioned` ducklake asset (Partitions tab on `/pipeline/<folder>`), it opens a range picker that previews which partitions in `[from, to]` are missing / failed / materialized (per `materialized_partition` — "the missing/failed set IS the backfill worklist", `docs/ducklake-materialization.md`), then launches one deployed run of the producing script per slice with an explicit `partition` arg (the worker only resolves a partition when the arg is absent, so the caller-provided value wins — same semantics as `wmill pipeline run --partition`). Runs are sequential (concurrent materializations of one ducklake table would contend on the catalog commit); a failed slice doesn't stop the rest. Progress streams in the dialog and, when it's closed, in the drawer header.

EE-gated on both sides, with the **backend core in `windmill-ee-private`**: producer resolution + range enumeration + status join live in EE modules (`windmill-api-assets/src/backfill_ee.rs`, `windmill-common/src/partition_ee.rs`) behind the `private` feature; OSS keeps only the thin handler and a stub that returns a clear "enterprise feature" error. The frontend button/dialog additionally gate on `$enterpriseLicense`. Single-partition runs with an explicit arg remain available in OSS.

## Changes

- **Backend (OSS)** — new `GET /w/{workspace}/assets/partitions_in_range?path&from&to` in `windmill-api-assets`: thin handler delegating to the `backfill` module pair (`backfill_ee` under `private`, `backfill_oss` stub otherwise); new `private` feature wired through `windmill-api-assets` → `windmill-api`. All reads through `UserDB` (RLS applies).
- **EE** (companion PR windmill-ee-private#641):
  - `windmill-api-assets/src/backfill_ee.rs` — the endpoint core: producer resolution (managed `// materialize` target match first, else a `partitioned` pipeline script with a write edge on the asset — covers polyglot/SDK producers, which can't carry `// materialize`), enumeration, join with `materialized_partition`, `{producer_path, partition_kind, partitions: [{partition, status}]}` response.
  - `windmill-common/src/partition_ee.rs` — `enumerate_time_partitions(spec, from, to)`: walks the range day-by-day (hourly: hour-by-hour), renders with the spec's tz-local `format`, collapses duplicates (one value per ISO week / month), honours the `start` anchor, rejects `dynamic` kinds and ranges over 1000 slices; renders through `write!` so a bad format specifier is a 400, not a panic. 8 new unit tests.
- **Frontend** — `backfillRun.ts`: pure sequential runner (injected launch/wait, mirrors `cascadeOrchestrator.ts`) + vitest; `BackfillRangeDialog.svelte`: range picker, missing/failed/materialized preview chips, "only missing + failed" toggle (default on), live per-slice progress, cancel with explicit "Cancelling…" feedback; `PartitionStatusGrid.svelte`: EE-gated button (disabled + "(enterprise feature)" tooltip in CE), owns the run state so progress keeps streaming in the header when the dialog is closed, refetches the grid after every slice, cancel also cancels the in-flight job (`cancelQueuedJob`).
- `openapi.yaml` + regenerated client; sqlx cache (2 new queries, EE caches preserved via the safe procedure — entries are keyed by query text, so they cover the queries' EE location too).

## Test plan

- [x] EE unit tests: `cargo test -p windmill-common --features private partition_ee` — 17 passed (8 new: daily/hourly/weekly ISO-year-boundary/monthly, start anchor, cap, dynamic + inverted-range rejection, bad format specifier)
- [x] Frontend: `npx vitest run …/backfillRun.test.ts` (5 tests: order, continue-past-failure, launch-error capture, cancel leaves pendings, snapshot immutability), `npm run check` clean
- [x] Compile: `cargo check -p windmill-api-assets` clean both with and without `--features private`; full `enterprise,private` build runs the dev stack
- [x] E2E on a live EE stack (`pipelines` workspace, bash producer `f/analytics/events_ingest` with `# pipeline` + `# partitioned daily`, seeded gaps): preview marked exactly the seeded gaps missing; "Backfill 4 partitions" ran the 4 slices sequentially (~3 s apart per `materialized_at`), grid streamed rows in as each landed; re-running with "only missing" showed 0-partition worklist (button disabled); "all in range" re-ran 8/8 ok; endpoint re-verified after the EE-module move
- [x] Cancel mid-run: in-flight job actually canceled in DB, "Cancelling — waiting for the current partition to finish", summary "5 succeeded, 1 failed, 5 not run"
- [x] CE gate (same binary, license removed): button disabled with "(enterprise feature)" tooltip; on OSS builds the endpoint errors via the `backfill_oss` stub
- [ ] Post-merge: exercise against a real duckdb `// materialize` producer (dev worker here lacks the duckdb feature; the managed-producer path is covered by the `materialize`-target match, exercised in unit form only)

## Screenshots

CE (no license) — Backfill gated:

ce-gate-disabled-backfill

Range preview — missing vs materialized, worklist toggle:

backfill-range-preview

Sequential progress in the dialog (grid rows streaming in on the right):

backfill-progress-dialog

Dialog closed mid-run — progress keeps streaming in the drawer header ("Backfilling 4/22"):

backfill-header-progress

Finished summary:

backfill-finished-dialog

Drawer after completion:

backfill-completed-drawer


rubenfiszel:
Addressed the CI review findings (OSS `d84f89c9b6`, EE `7c1450e` — `ee-repo-ref.txt` updated):

**Codex 1 / Claude 2 — runtime license gate ([P1])**: `backfill_ee::partitions_in_range` now checks `LICENSE_KEY_VALID` before doing anything (same pattern as `ssh_executor_ee`), so an EE binary without a valid key rejects direct API calls regardless of the frontend gate. Verified live both ways on the dev stack: with the license key removed the endpoint returns `400 "Backfilling a range of partitions requires a valid enterprise license"`; with it restored, `200` with the correct per-slice statuses.

**Codex 2 — cancel-before-job-id race ([P2])**: `runBackfill` now takes a `cancelJob` hook and, when cancellation raced the launch (flag set while `runScriptByPath` was in flight, no id to cancel yet), cancels the job as soon as the id arrives, before waiting on it. `PartitionStatusGrid` passes its non-throwing `cancelJobById`. New unit test covers exactly this window (`cancels a job whose launch raced the cancellation`).

**Claude 1 — asset dispatch left ON**: intended — a backfilled slice should refresh its deployed consumers like any deployed run, carrying its partition down the chain (this is deployed-run parity; the `_wmill_skip_asset_dispatch` callers are client-orchestrated *dev* cascades that would otherwise double-fire). Recorded as a comment at the launch site in `PartitionStatusGrid.svelte`.

Checks after the changes: `cargo check -p windmill-api-assets --features private` clean, vitest 6/6, `npm run check:fast` clean, plus the live licensed/unlicensed endpoint probes above.

rubenfiszel:
Addressed the Codex nit in `13144326c5`: the `producer_path` OpenAPI description now covers both resolution paths — "the pipeline script that materializes the asset (managed `// materialize` target, or a partitioned writer using the SDK helpers)" — matching the Rust struct doc and the implementation. Client regenerated (gitignored).

PR #9890, opened 2026-07-02
feat(pipelines): workspace duckdb macro libraries (// macros / // use)

rubenfiszel:
## Summary

Closes the dbt macro/reuse gap (`docs/pipelines-vs-dbt.md` gap #7) with **workspace DuckDB macro libraries** — engine-native `CREATE MACRO` instead of Jinja templating, resolved and injected by Windmill ("resolve-and-inject"):

- A DuckDB script annotated `// macros` is a **macro library**: its body is `CREATE OR REPLACE MACRO` statements (scalar or `AS TABLE`) plus plain setup. Deploy parses each macro into a new `macro_definition` registry (names workspace-unique) — available to consumers the instant the deploy commits.
- Any DuckDB script that **calls** a registered macro gets `CREATE OR REPLACE TEMP MACRO …` blocks injected into its statement list at job time — dependency-topo-ordered (DuckDB bind-checks macro bodies at CREATE) and spliced after the setup/ATTACH prefix so bodies can reference attached catalogs. **Late-bound**: each run reads the current registry, so a lib redeploy applies to the next run (dbt semantics, no compile step).
- `// use <lib_path>` force-injects a whole library (definitions + its setup) — the escape hatch for dynamic SQL (e.g. calls inside `query('…')` strings) that lexical detection can't see.
- Asset graph: libraries render as nodes ("defines N macros" ƒ badge + signature strip in the details pane) with violet lib→consumer edges from deploy-recorded detection (`macro_usage`) and `// use`. New "Macro library" scaffold in the pipeline script creator.

**Trust model (deliberate, dbt-package-like):** macro libraries are workspace-trusted code — implicit detection injects them into other users' DuckDB jobs with the consuming job's credentials, the same boundary as a dbt package. Guardrails: a script's **own** macro definitions always win (excluded from injection entirely), names are workspace-unique, and shadowing DuckDB built-ins is rejected at deploy (DuckDB silently allows it!). Documented in `docs/pipelines-vs-dbt.md`.

### Authoring a macro library

Two entry points: in the pipeline editor, the **"+" script creator → output kind "Macro library"** scaffolds a ready-to-edit lib; or write any DuckDB script whose header starts with `// macros`:

```sql
-- macros
CREATE OR REPLACE MACRO safe_div(a, b, fallback := 0) AS
  CASE WHEN b = 0 THEN fallback ELSE a / b END;

CREATE OR REPLACE MACRO pct(part, total) AS round(100 * safe_div(part, total), 2);

CREATE OR REPLACE MACRO sample_rows(src, n) AS TABLE
  SELECT * FROM query_table(src) LIMIT n;
```

Scalar and table macros, default parameters (`:=`), macros calling earlier macros, plus plain setup statements (ATTACH a static catalog, `CREATE TEMP TABLE` lookup data, `SET`) are all supported. **Deploying is publishing** — the macros are callable from every DuckDB script in the workspace the instant the deploy commits. No apply step, no compile, no package manager.

Mistakes are caught at deploy with precise, actionable errors:

- ``macro `concat` shadows a DuckDB built-in function; pick another name`` (DuckDB would silently allow this — we don't)
- ``macro `outer` calls `inner`, which is defined later in the file; define `inner` first (DuckDB bind-checks macro bodies at creation)``
- ``macro `safe_div` is already defined by `f/lib/stats`; macro names are workspace-unique``
- ``a `// macros` library may only contain CREATE [OR REPLACE] MACRO statements plus setup (ATTACH/INSTALL/LOAD/SET/PRAGMA); found: `SELECT 1` ``
- ``a `// macros` library cannot use managed ATTACH forms (ducklake://, datatable://, resource URIs) in v1``
- `` `// macros` is only supported for DuckDB scripts, not bun``

### Consuming — you just call it

No import, no annotation, no config. Any DuckDB script (pipeline member or not, deployed run, editor Test, preview) that calls a registered macro works:

```sql
-- pipeline
SELECT safe_div(revenue, users) AS arpu,
       pct(paid, total) AS conversion
FROM sample_rows('lake.events', 10000);
```

At job time the worker detects the calls, resolves the transitive closure (`pct` pulls in `safe_div`), injects the definitions in dependency order, **and injects the providing library's own setup statements first** — so a macro that reads its lib's ATTACHed catalog or seeded temp table just works. Editor **autocomplete** suggests every workspace macro while you type, showing the signature as the label (`safe_div(a, b, fallback := 0)`), scalar/table + provider path as the detail, the full definition on hover, and inserting `name(│)` with the cursor inside the parens.

### Editing — late-bound, like dbt packages

Redeploy the library and **the very next run of every consumer uses the new definition** — consumers are never redeployed or relocked. If you need to opt a script out, define your own macro with the same name: **a local `CREATE [TEMP] MACRO` always wins** — its name is excluded from injection entirely, so no library deploy can ever change that script's behavior. Local and injected definitions compose in both directions: a local macro may call a registry macro (the injected definition is woven in *before* the local CREATE, since DuckDB bind-checks at creation) and an injected macro may call a local one (woven *after* it); a genuinely contradictory ordering is a clear error telling you to reorder your local definitions rather than a silent mis-bind.

### Dynamic SQL — the one manual step

Lexical detection can't see a call hidden inside a string, e.g. `query('SELECT safe_div(…)')`. For that, `// use f/lib/stats` force-injects the whole library (definitions + setup). Libraries can declare `// use` **themselves**, and it's honored transitively — if lib B's macro builds SQL strings that call lib A's macros, B declares `// use f/lib/a` once and every consumer of B inherits it automatically; the dependency stays encapsulated in the library. `// use` also composes with shadowing: even if you locally shadow every macro of a used lib, its setup (ATTACH side effects) still runs first.

### Discovering what exists

- **Pipeline graph**: a library renders as a node with a ƒ ×N chip (hover lists every signature); violet dashed edges run to each consumer, labeled `macros` (detected calls, with a badge listing exactly which ones) or `uses lib` (`// use`). Selecting the library shows a signature strip pinned above its source.
- **"Macros" toolbar button** (pipeline page, View and Edit modes) opens a workspace-wide **explorer drawer**: every macro grouped by library, with signature, scalar/table chip, body preview, copy-call button, name/path filter, and an open-library jump (selects the node if it's on the current canvas, opens the script page otherwise). The listing enforces the same visibility as reading the provider script — libraries in folders you can't read don't appear.
- **Autocomplete** (above) doubles as discovery while writing SQL.

### Failure modes you might actually hit

- Calling a macro that doesn't exist (typo, lib not deployed yet): DuckDB's clean `Catalog Error: Scalar Function with name X does not exist` at run time. Deploying a consumer whose `// use` target isn't deployed yet is a warning, not an error, so git-sync can deploy in any order — the runtime is the enforcement point.
- Renaming a library keeps the registry consistent automatically, but `// use f/old/path` strings in consumers are plain text and fail at run time until updated.
- Agent (HTTP) workers have no DB access: implicit detection silently degrades there (a called macro then fails with the Catalog Error) and `// use` hard-errors, same as custom data tests.

### Lifecycle guarantees

Archive/delete of a library removes its macros from the registry (and consumers' recorded edges); workspace **fork clones** the registry so forked consumers run immediately; workspace rename/delete migrate/purge it. Deploy of a library also refreshes the recorded graph edges of every live DuckDB consumer.

### Performance

The registry read is cached per workspace on the worker (60s TTL) and **evicted via the transactional `notify_macro_registry_change` event** emitted with every registry mutation — so a lib redeploy propagates to all replicas within the notify poll interval (~10s default) while steady-state cost drops to one indexed SELECT per workspace per minute, even across cascades running many DuckDB jobs per second. Beyond that: one linear text scan of the script per job, one hash-cached content fetch per involved library when macros are actually selected, and a handful of `CREATE OR REPLACE TEMP MACRO` statements at job start (microsecond-scale, dwarfed by DuckDB job startup). Workspaces that don't use macros pay only the cached (mostly free) registry lookup. The cache is disabled on `CLOUD_HOSTED` (unbounded workspace count would churn the 1000-entry cache, and per-entry memory is tenant-controlled — cloud jobs just do the per-job SELECT); `MACRO_REGISTRY_CACHE_DISABLED` mirrors `DEPLOYED_SCRIPT_CACHE_DISABLED` for multi-DB test processes.

## Changes

- **Parsers** (`windmill-parser`): `// macros` + accumulating `// use` annotations (Rust + TS mirror + shared parity corpus); new `duckdb_macros` module (verbatim header parser, lexical call detection, Kahn topo ordering, local-definition extraction); vendored DuckDB 1.5.4 builtin-name list; `split_statements` fixed to decode chars (was Latin-1 byte-wise — mojibaked any multi-byte text it emitted).
- **Deploy** (`windmill-api-scripts`): validation (DuckDB-only, no `// materialize` combo, macros-or-setup-only body, ≥1 macro, within-file forward-reference check, builtin-shadow + cross-lib collision rejection, no managed ATTACH in lib setup) and wipe-and-reinsert registry/edge writes (rename-safe); `// macros` implies pipeline membership; cleanup on archive/delete/bulk-delete; workspace rename/delete/**fork-clone** handling.
- **Worker** (`duckdb_executor`): `inject_workspace_macros` at the post-ATTACH-transform seam — fetch registry, seed from detected calls ∪ `// use` libs, transitive closure over bodies, topo-sorted `CREATE OR REPLACE TEMP MACRO` splice. Agent (Http) workers: implicit path degrades silently, `// use` hard-errors (same as custom data tests). Locally-defined macro names are excluded from injection (local wins).
- **Graph API** (`windmill-api-assets`): `macros` on runnable nodes, `macro_edges` on the response (+ openapi).
- **Frontend**: macro-lib node badge, violet dashed lib→consumer edges with ƒ badge (`via_use` labeled "uses lib"), details-pane signature strip, live draft/`// use` overlays in `resolveGraph`, "Macro library" template.
- Migration `20260702095830_duckdb_macro_registry` (`macro_definition`, `macro_usage`; FK CASCADE + explicit windmill_user/windmill_admin GRANTs).

- **Discovery UX** (follow-up commit): `GET /w/{ws}/assets/macros` lists the registry; the DuckDB **editor autocompletes** workspace macros (signature + body docs, snippet insert, short-TTL cache); a **"Macros" explorer drawer** on the pipeline page lists every macro grouped by library (filter, scalar/table chips, body preview, copy-call, open-library).

## Test plan

- [x] `cargo test -p windmill-parser` (135: header parse incl. non-ASCII + panic regressions, detection, topo, annotations, parity corpus)
- [x] `cargo test -p windmill-worker --lib --features duckdb` (169: injection planning, splice placement, local-wins, use-lib)
- [x] Frontend: parity + unit + `resolveGraph` vitest (159), `npm run check` (0 errors)
- [x] E2e on dev stack (server + duckdb-featured worker): lib deploy → registry rows; all 6 validation errors verbatim; implicit consumer run → `safe_div(10,0)=0`, macro-calling-macro `pct`, table macro over `query_table` ✓; `// use` consumer with call inside `query('…')` string ✓; lib redeploy → next consumer run picks the new definition (late-bound) ✓; graph API returns `macros` + both edge kinds ✓
- [x] Fork a workspace containing a lib + consumer: registry + usage edges cloned, both consumers ran in the fork (implicit detection and `// use`), fork graph API correct; deleting the fork purged both macro tables

- [x] Autocomplete verified in the browser: typing `sam` in a DuckDB editor suggests `sample_rows(src, n)` ("table macro · f/macrolib/stats"); Tab inserts `sample_rows()` with the cursor inside the parens
- [x] Explorer drawer verified: grouped list with signatures/bodies, opened from the pipeline toolbar

## Screenshots

macro-graph-overview

macro-lib-details

macro-drawer

macro-autocomplete

---
## Summary by cubic
Adds workspace-wide DuckDB macro libraries that are resolved at deploy and injected into consumer runs, plus DuckDB editor autocomplete and a macro explorer. Injects each provider library’s setup (implicit calls and `// use`), honors transitive `// use`, surfaces libs/edges in the graph, and lists macros via an RLS‑gated API.

- **New Features**
  - `// macros` marks a DuckDB script as a library; deploy parses and stores definitions in a workspace‑unique `macro_definition` registry with validation (DuckDB‑only, no `// materialize`, no built‑in shadowing, no managed ATTACH, cross‑lib collision checks).
  - Workers inject required macros as `CREATE OR REPLACE TEMP MACRO` in dependency order after ATTACH; never override locally defined names; inject the provider library’s setup for implicit detection and `// use` (including transitive `// use` declared by libraries); always run that setup before any user statements.
  - Asset graph shows macro‑library nodes and violet lib→consumer edges; API returns `macros` on runnable nodes and `macro_edges`; new GET `/assets/macros` lists registry macros and is RLS‑gated (only active provider scripts).
  - Parser: lexical call detection, topo ordering, strict header/annotation parsing, multibyte‑safe `split_statements`; vendored DuckDB built‑ins list for shadowing checks.
  - Frontend: macro‑lib badges and edges, “Macro library” template, DuckDB editor autocomplete for workspace macros, and a searchable macro explorer drawer.

- **Performance**
  - Cache the macro registry per workspace on workers (60s TTL) and evict via the transactional `notify_macro_registry_change` event; disabled on cloud.

<sup>Written for commit 7fbc638c32de2bbd887e44fa43e06494f783d0fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed all 5 cubic findings in 0d9727e88c:

1. **(P1) `// use` setup retention** — when every lib macro is shadowed by local definitions, the lib's setup statements are still injected (`plan_macro_injection` returns `use_setup` instead of dropping it) + regression test.
2. **(P2) phantom lib nodes** — macro edges now pass the same `runnable_exists` orphan filter as the other edge families (both endpoints) before entering `runnable_set`.
3. **(P2) signature-strip keying** — index-keyed `{#each}` so a mid-edit duplicate name can't crash the keyed list.
4. **(P2) lib redeploy vs non-pipeline consumers** — the usage-edge rescan now covers all live DuckDB scripts, not just `auto_kind='pipeline'`, so edges recorded at a consumer's own deploy survive a library redeploy.
5. **(P2) splice vs leading local `CREATE MACRO`** — the injection point now skips past leading local macro definitions too (`is_macro_definition`), so an injected body that calls a local macro binds after the local CREATE has run + regression test.

The earlier `test-linux` failure (`Mixed scripts and flows with nonDottedPaths are idempotent`) is unrelated to this diff — it's the CLI sync flow-lock race this test was previously stabilized for in #9080.

rubenfiszel:
Fixed the P1 (implicit macro calls dropping the provider library's setup) in 3828a0ac92:

Selection is now a separate pure pass (`select_workspace_macros`), so the worker knows every provider library whose macros were selected — implicitly or via `// use` — and fetches each one's deployed body. `plan_macro_injection` injects **every relevant provider's setup statements ahead of the definitions** (`// use` libs first in annotation order, then remaining providers, with exact-duplicate statements deduped so two libs attaching the same catalog don't double-ATTACH). A macro body that references its own lib's ATTACH/temp-table setup now binds on the implicit path, matching the documented semantics.

Verified: unit tests (`implicit_call_injects_provider_setup`, `duplicate_setup_across_libs_is_deduped`, 173 total) plus e2e on the dev stack — a lib with `CREATE TEMP TABLE kv …; CREATE MACRO kv_lookup(k) AS (SELECT v FROM kv …)` and a consumer that only calls `kv_lookup('a')` (no `// use`) now returns 42; both pre-existing consumers still pass.

rubenfiszel:
Both P1s fixed in 7132ca72f3:

**1. `/assets/macros` visibility** — the listing now matches reading the provider script itself: an `EXISTS` join against `script` runs under the user_db (RLS) transaction, so a user without read access to a library's folder no longer sees its macro bodies, plus a `build_scope_path_predicate(\"scripts\", \"read\")` post-filter mirroring `list_scripts` for path-scoped tokens. Note: scoped tokens are currently fail-closed on ALL `/assets/*` routes (no `assets` scope domain exists in `extract_domain_from_route`), so the scope predicate is defense-in-depth; the exploitable half was RLS, verified in SQL: with `set_session_context(false,'limited',…, folders_read='macrolib')` the query returns only macrolib macros while the pre-fix shape leaked the private lib's `secret_rate`; granting the folder restores it.

**2. Library-level `// use` honored transitively** — selection now runs as a fixpoint (`resolve_macro_selection`): the worker lazily fetches every relevant provider's body, parses its annotations, and folds each library's own `// use` declarations into the effective set — so a lib whose macro hides a call inside `query('…')` declares the dependency once and every consumer of that macro gets the used library (definitions + setup) automatically. Verified by unit test (`provider_lib_use_is_honored_transitively`, 174 total) and e2e: `f/macrolib/wrapper` (`// use f/macrolib/stats`, macro body `query('SELECT safe_div(10,4) …')`) consumed implicitly returns 2.5; all three earlier consumers still pass.

rubenfiszel:
Fixed the local↔injected ordering P1 in d4b009b2b4:

`splice_macro_blocks` (single insertion point after the leading prefix) is replaced by **`weave_macro_blocks`**, which places each injected definition by bind-order constraints in both directions:

- an injected definition **called by a local macro** moves before that local's CREATE (transitively pulling the injected definitions it depends on with it);
- an injected definition **referencing a local macro** stays after it (the earlier fix's direction);
- an unsatisfiable combination (local A calls injected X, X references local B defined after A) is a clear error — "conflicting order requirements … reorder the local definitions" — instead of a silent mis-ordering;
- library setup statements land at the earliest slot any injected definition took (injected bodies are self-contained w.r.t. their own lib's setup);
- with no local interplay, placement is unchanged (after the setup/local-def prefix).

Verified: 4 new weave unit tests (before-local, after-local, transitive pull, conflict error; 177 total) and e2e — the exact reported shape `CREATE TEMP MACRO outer_pct(x) AS pct(x, 200) + 0; SELECT outer_pct(25);` now returns 12.5, and all four existing consumers regression-pass.

rubenfiszel:
Fixed in 83f25740d8: injected library setup now runs **before all user blocks**, unconditionally — not at the earliest definition slot (which fell back to after the local-definition prefix when every lib macro was shadowed and no defs were injected).

Rationale: injected setup is fully self-contained — lib deploy validates it as plain/non-managed statements and within-lib order is preserved — while *any* consumer statement can depend on it at bind time, including a leading local macro that shadows a lib macro but still references the lib's ATTACHed catalog. Placing it first is therefore always safe and always sufficient; the definitions keep their constraint-woven slots (all ≥ setup).

Verified: 2 new weave unit tests (178 total) and e2e — `// use f/macrolib/kvlib` + local shadow `CREATE TEMP MACRO kv_lookup(k) AS (SELECT v * 10 FROM kv …)` returns 420, proving the lib's `CREATE TEMP TABLE kv` ran before the local CREATE *and* that the local shadow still wins over the lib's definition (42 × 10). All five existing consumers regression-pass.

rubenfiszel:
Added the registry-read cache with event-driven invalidation (suggested in review):

- Workers cache the `macro_definition` rows per workspace (`MACRO_REGISTRY_CACHE` in windmill-common, quick_cache, 60s TTL as backstop).
- Every registry mutation — lib deploy/redeploy/rename in `create_script_internal`, archive/delete via `clear_macro_registry` — emits a **transactional** `INSERT INTO notify_event ('notify_macro_registry_change', workspace_id)`, and the existing notify-event poller dispatch in main.rs evicts the workspace's entry on every replica. Worst-case staleness after a lib redeploy = the notify poll interval (~10s default); steady-state cost drops from one SELECT per DuckDB job to one per workspace per minute.
- `MACRO_REGISTRY_CACHE_DISABLED` AtomicBool mirrors `DEPLOYED_SCRIPT_CACHE_DISABLED` for multi-DB test processes (the cache is keyed by workspace id alone).

Verified e2e with the worker's poll interval at 2s: two consecutive consumer runs hit the cache (`zero_div = -1.0` twice), lib redeploy flipping the fallback to -7, and the next run returned `-7.0` — eviction via the event, well inside the 60s TTL. PR body + docs updated to describe the propagation bound.

PR #9891, opened 2026-07-02
fix: invalidate bun bundle cache on transitive relative-import changes

### LINKED ISSUE #9889: Stale bun bundle: transitive relative-import changes never invalidate the bundle cache
# Describe the bug

For Bun/TypeScript scripts, editing a module that a deployed script imports **transitively** through relative imports (2+ hops) does not invalidate the importer's prebuilt bundle. The worker keeps serving the previously built bundle — which has the old module code inlined — indefinitely. The stale bundle survives worker restarts (it is re-fetched from the object-store bundle cache) and is only busted when the runnable itself, or one of its **direct** relative imports, gets a new script version.

```
lib_script_a  ──imported by──▶  lib_script_b  ──imported by──▶  main_script
     ▲ edit this                                                     ▲ keeps returning old code
```

Direct-import changes (editing `lib_script_b` above) are picked up correctly. Only changes ≥2 hops away go stale. Import style makes no difference — `import ... from "./x"`, barrel re-exports (`export * from "./x"`), and absolute `/f/...` imports all behave identically.

## To reproduce

1. Create three deployed Bun scripts in a folder:
   - `f/repro/lib_script_a`: `export function a() { return "V1" }`
   - `f/repro/lib_script_b`: `import { a } from "./lib_script_a"; export function b() { return \`B(${a()})\` }`
   - `f/repro/main_script`: `import { b } from "./lib_script_b"; export async function main() { return b() }`
2. Run `main_script` → returns `B(V1)`. (This first run builds and caches the bundle.)
3. Edit **only** `lib_script_a` (`V1` → `V2`) and deploy it (UI deploy or `wmill sync push` — both reproduce).
4. Run `main_script` again.

**Expected:** `B(V2)`
**Actual:** `B(V1)`, consistently, on every subsequent run. `scripts/get` confirms `lib_script_a`'s deployed source is `V2`; `main_script` was never rebundled. Restarting the worker does not help (local cache is cleared, but the same stale bundle is pulled back from the object store under the same key). Reproduced both with and without an npm dependency on the runnable; a runnable with a non-empty npm lock reproduces it most reliably because nothing else ever perturbs its cache key.

## Root cause (traced in source)

The bundle cache key is blind to transitive imports, and the mechanism that is supposed to rebuild importers early-exits on that same blind key:

1. **Cache key only covers direct imports.** `compute_bundle_local_and_remote_path` (`backend/windmill-worker/src/bun_executor.rs`, ~L1257) hashes: entrypoint content + lock + for each **direct** relative import, its path and latest `created_at`. The import extraction (`extract_relative_imports` → `windmill_parser_ts::parse_relative_imports`) parses only the entrypoint's own import/export statements and is not recursive, so `main_script`'s key contains `lib_script_b`'s `created_at` but knows nothing about `lib_script_a`.

2. **The dependents-recompute chain does fire, but cannot change the key.** Deploying `lib_script_a` triggers `process_relative_imports` → `trigger_dependents_to_recompute_dependencies`, which queues a `Dependencies` job for each importer (transitively, chained through the dependency map). However, that dependency job intentionally updates the importer's lock **in place on the same script hash** (`backend/windmill-worker/src/worker_lockfiles.rs`, ~L242: "We do not create new row for this update"), so the importer's `created_at` never changes — and therefore neither does any downstream bundle cache key.

3. **The rebuild is skipped by an early exit on the unchanged key.** That same dependency job calls `prebundle_bun_script` (`worker_lockfiles.rs` ~L2845 → `bun_executor.rs` ~L1150), which would rebuild the bundle with the fresh leaf code — but its first step is `if exists_in_cache(&local_path, &remote_path) { return Ok(()); }` (`bun_executor.rs` ~L1173). The key is unchanged, the stale bundle exists, so the rebuild never happens.

4. **At run time** `handle_bun_job` computes the same key, hits the (local or object-store) cache, and executes the stale bundle.

So every invalidation path converges on the same non-transitive key: nothing in the system can ever refresh the bundle after a transitive-only change.

Note that even overwriting the object-store entry under the same key (e.g. by removing the `exists_in_cache` early exit) would not be a complete fix on its own: workers that already have the bundle in their local disk cache hit the local copy first and would never re-fetch. A correct fix has to make the **key** change.

## Suggested fix directions

- Make the cache key transitive: walk relative imports recursively when computing the key (fetch each imported script's latest content, extract its imports, cycle-safe), including every visited script's `(path, created_at)`. Alternatively, derive the transitive closure from the `dependency_map` table (the edges are already maintained there) instead of re-parsing content at each hop.
- Since a transitively-edited leaf *does* get a new script version (new `created_at`) when deployed, a transitive key changes exactly when it should — and the existing dependents-triggered `prebundle_bun_script` would then repopulate the cache under the new key ahead of the next run, keeping the current warm-cache behavior.

## Workarounds until fixed

- Touch (any content change) the runnable itself, or any of its direct imports — e.g. editing the barrel `index.ts` busts every runnable that imports it directly.
- Add `//nobundling` to the runnable: disables the bundle cache entirely and resolves imports at each execution (slower cold start, but never stale).
- Move the whole tree to a `codebase` (its digest hashes the full source tree, so any file change invalidates).

## Environment

- Reproduced on recent EE and current `main` (code paths above are current as of d65f58c388).
- Bun/TypeScript deployed scripts (non-codebase), default bundling enabled.

rubenfiszel:
Fixes #9889

## Problem

For Bun scripts, editing a module that a deployed runnable imports **transitively** through relative imports (2+ hops) never invalidated the runnable's prebuilt bundle. The worker kept serving the old bundle — with the old module code inlined — indefinitely, surviving worker restarts (re-fetched from the object-store bundle cache under the same key).

```
leaf  ──imported by──▶  mid  ──imported by──▶  main_script
 ▲ edit this                                     ▲ keeps returning old code
```

Root cause: the bundle cache key (`compute_bundle_local_and_remote_path`) hashed only the entrypoint's **direct** relative imports (path + latest `created_at`), while the bundle *content* inlines the whole transitive closure. Every invalidation path converged on that key:
- the dependents-recompute chain fires dependency jobs for importers, but those update the lock **in place on the same hash** (no `created_at` bump), so no key input changed;
- the dependency job's `prebundle_bun_script` would rebuild with fresh code, but early-returns on `exists_in_cache` — same key, stale bundle exists, rebuild skipped.

## Fix

Key the bundle on the **transitive closure**: `collect_transitive_import_versions` walks relative imports recursively and feeds every reachable script's `(path, latest deployed hash)` into the key. When a leaf is redeployed it gets a new hash, so every bundle that inlines it gets a new key → cache miss → rebuild (the already-existing dependents-triggered prebundle now repopulates the cache ahead of the next run instead of early-returning).

The walk costs **no DB queries in steady state** — strictly cheaper than the old code, which ran 2 queries per direct import on every job execution:
- path → latest deployed hash via `DEPLOYED_SCRIPT_HASH_CACHE` (already evicted by `notify_runnable_version_change` events on deploy/lock-update, 60s TTL fallback) — the same freshness contract as run-by-path version resolution and the `/raw` import-content cache (#6769), so key and bundle content stay coherent;
- hash → parsed relative-import list memoized in `RELATIVE_IMPORTS_PER_HASH` (content is immutable per hash, no invalidation needed), contents via the existing per-hash script cache.

Closure walk is cycle-safe (visited set) and capped at 256 scripts (logged if exceeded). Unresolvable imports are excluded from the key, matching what the bundler can inline. No schema change, no API change, `compute_bundle_local_and_remote_path` signature unchanged (EE agent-workers precompute path unaffected).

## Validation

- `cargo check` clean (backend running under cargo watch).
- New regression test `test_bun_transitive_import_change_rebundles` (main → mid → leaf, leaf-only edit, runnable hash/lock byte-identical): **fails on the old code (returns stale `M(V1_FROM_LEAF)`), passes with the fix**.
- Live e2e on a dev instance, both variants from the original report:
  - plain chain: `M(V1_FROM_LEAF)` → edit leaf only → `M(V2_FROM_LEAF)` on the next run;
  - runnable with an npm dependency (`uuid`), non-empty lock untouched: `M(DEPV1_FROM_LEAF) [uuid:4e44ef5d]` → edit leaf only → `M(DEPV2_FROM_LEAF) [uuid:4e44ef5d]`;
  - job logs confirm runs still use `--- BUN BUNDLE SNAPSHOT EXECUTION ---` (cache hit) between deploys.

Existing bundles are rebuilt once on first run after upgrade for scripts with relative imports (key format changed); scripts without relative imports keep their keys.

---
## Summary by cubic
Editing a transitively imported script now invalidates the cached `Bun` bundle for the runnable, so runs always pick up the latest code. The bundle cache key now covers the whole transitive closure of relative imports and matches the loader’s version selection.

- **Bug Fixes**
  - Key bundles on every reachable script’s `(path, latest non-archived hash)` instead of only direct imports, ensuring rebuilds when any leaf changes and keeping the key consistent with the content endpoint.
  - Traversal uses notify-evicted `IMPORTED_SCRIPT_HASH_CACHE` and per-hash import memoization (`RELATIVE_IMPORTS_PER_HASH`); it’s cycle-safe, capped at 256, sorted for a stable key, skips unresolvable imports, and does not memoize transient fetch errors.
  - Evict `IMPORTED_SCRIPT_HASH_CACHE` on deploy notify; updated `compute_bundle_local_and_remote_path`; added `get_latest_script_hash_for_import_cached`; removed the `created_at` lookup. Added regression test `test_bun_transitive_import_change_rebundles`.

- **Migration**
  - Bundles with relative imports will rebuild once after upgrade due to the new key format. No schema or API changes. Scripts without relative imports keep their keys.

<sup>Written for commit c7f93e8ae71065e8bdd868a91179462df6fd721f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9891?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the version-selection mismatch finding in c7f93e8ae7: import versions for the bundle key are now resolved with the exact predicate the loader's content endpoint uses (`archived = false ORDER BY created_at DESC`, no lock filter), via a dedicated `IMPORTED_SCRIPT_HASH_CACHE` (separate from the lock-filtered `DEPLOYED_SCRIPT_HASH_CACHE`, which must keep its stricter semantics for run-by-path resolution). The new cache is evicted in the same `notify_runnable_version_change` handler as `RAW_SCRIPT_LATEST_HASH_CACHE` and shares the 60s TTL fallback, so key and inlined content flip to a new version in the same window — including versions whose lock is still pending or errored.

PR #9915, opened 2026-07-03
feat(pipelines): fork data environments for ducklake materialization (dev data)

rubenfiszel:
## Summary

Workspace forks fork the CODE, but `// materialize ducklake://<lake>/<table>` targets were absolute pointers into shared physical storage: a fork clones `workspace_settings.ducklake` verbatim, so running a materializing pipeline in a fork **wrote the parent workspace's production tables**. This PR gives forks (throwaway `wm-fork-*` and standing dev workspaces) an isolated data environment with dbt-style *defer*, transparently — no annotation edits, no grammar changes.

Design writeup: `docs/pipeline-env-isolation.md` (validated with a DuckLake semantics spike before implementation).

## Changes

- **Write redirect** — `get_ducklake_from_db_unchecked` is fork-aware: in a fork the lake resolves to a fork-scoped `METADATA_SCHEMA 'wm_fork_<mangled>_<hash>'` (same catalog DB; injective, ≤63 chars) and a bucket-root `__wm_forks/<wid>/<lake path>` DATA_PATH. Bucket-root rather than nested-under-the-parent-path so parent lake maintenance (orphan-file cleanup) can never garbage-collect live fork files. The redirect rides in the *existing* `DucklakeWithConnData` fields, so the EE agent-worker endpoint is untouched and stale agents still write the fork namespace (fail-safe). Everything funnels through the ATTACH transform — managed materialize, `manual`, raw SQL, and every UI preview/History/backfill surface.
- **Read-defer** — in a fork, the transform also attaches each fork-chain ancestor **READ_ONLY** (no `AUTOMATIC_MIGRATION`/`CREATE_IF_NOT_EXISTS`; a fork job physically cannot write prod through ducklake) and creates `CREATE VIEW IF NOT EXISTS` defer views over the direct parent for every table the parent has materialized but the fork hasn't — so consumers read the parent's current data until the fork materializes the table itself, and you can iterate on one node without rebuilding upstream. SCD2 parents get the `<table>_current` companion view (detected from the captured schema).
- **View→table transition keyed on live catalog state** — when a managed materialize targets a table that currently exists *as a view* in the fork namespace, the view is dropped first. The decision keys on the catalog's actual `ducklake_view` rows (`fork_defer.fork_views`), not recorded status: after a failed run, status can't distinguish view from table, and a mismatched DROP wedged the asset on every retry in e2e. `sql_materialize.rs` and both annotation parsers are untouched.
- **Option hardening** — in fork mode, user/settings `METADATA_SCHEMA`/`DATA_PATH`/`OVERRIDE_DATA_PATH` ATTACH args are stripped (DuckDB silently keeps the last duplicate), and the fork-owned `DATA_PATH` is re-emitted as the last option so it wins regardless of literal syntax. MySQL-catalog lakes error loudly in forks instead of ever writing the shared catalog.
- **Graph indicator** — `/assets/graph` gains per-ducklake-asset `fork_materialization: 'fork' | 'deferred'` in forks (parent rows read on the plain pool — fork membership ≠ parent membership). Amber ↗ / emerald fork chips on asset nodes + matching banners in the details pane.
- **Lifecycle & cleanup** — sidecar registry `fork_ducklake_namespace` (GRANTs in the same migration, FK CASCADE, TTL'd once-cache) records exactly what each fork attached. New `POST /w/{ws}/workspaces/drop_forked_ducklake_namespaces` (same permission as delete_workspace) drops the `wm_fork_*` pg metadata schema (hard prefix guard) and deletes the `__wm_forks/<wid>/…` objects via the workspace object store; per-lake registry rows are deleted only after both cleanups succeed so a retry resumes. Wired into the fork-delete UI next to the datatable drop, including for deleted child forks.
- **Fork-time choice (per lake): isolated vs shared** — the fork-creation dialog gains a "Ducklake data environment" section (rendered only when the parent has lakes configured, mirroring the datatable section): each lake defaults to **Isolated** (everything above) and can be explicitly switched to **Shared with parent** (the fork reads and writes the parent's lake directly — e.g. for prod-equivalent backfill forks), with a warning callout. The choice ships as `shared_ducklakes` on the create-fork request and is stamped as `fork_behavior: "shared"` on the lake entry in the fork's settings; the resolver, graph indicator, and cleanup all key off it, and absent-field-means-isolated keeps existing forks and API callers on the safe path. Shared also lifts the mysql-catalog restriction (nothing needs namespacing); a shared fork ancestor in a chain resolves like a root.
- **Fork lineage clone (load-bearing pre-existing fix)** — `asset` and `script_trigger` rows are deploy-derived and were never cloned into forks: a fork's pipeline graph had no asset nodes/edges and the asset dispatch cascade never fired in forks. `clone_asset_usages_and_triggers` fixes both (ci_test_reference/macro-registry clone pattern).

## E2E verification (live EE + MinIO + DuckDB stack)

Full lifecycle exercised: parent materialized `orders`+`daily_revenue` → forked → fork ran the consumer (read parent via defer view, wrote its own namespace, own snapshot sequence) → edited the fork's producer to 10× values → materialize (defer view→table transition) auto-cascaded to the consumer via the cloned triggers → **parent lake byte-identical throughout** (parent values `20/35/15` and `max(snapshot_id)=3` unchanged; fork read `200/350/150`) → 20k-row bulk write landed as parquet under `__wm_forks/<wid>/…` with zero new parent-lake files → cleanup endpoint dropped the metadata schema + data files and cleared the registry (including the partial-failure retry path) → fork deletion left zero residue.

Shared opt-out verified separately: a fork created with `shared_ducklakes: ["main"]` got the stamped config, its pipeline run **advanced the parent lake's snapshot** (the opted-in behavior), created no fork schema and no registry row, and its graph showed no chips — while a plain fork created alongside still defaulted to deferred chips.

Unit tests: namespace mangling injectivity, reserved-arg stripping, and `fork_defer_statements` (READ_ONLY ancestor attach shape, target transition vs. fork_views, schema-qualified defer views).

## Screenshots

Fork pipeline graph — amber ↗ = deferred to parent, emerald = fork-materialized; details pane shows the matching banner:

fork-graph-chips

fork-deferred-banner

fork-materialized-banner

Fork-creation dialog — per-lake data-environment choice (isolated default; shared shows a warning):

fork-dialog-ducklake-choice

Workspace settings → Ducklake in a fork — banner explaining fork data environments, per-lake isolated / shared-with-parent chips (the shared stamp also survives settings saves now):

fork-ducklake-settings

fork-dialog-final

## Test plan

- [x] DuckLake semantics spike (METADATA_SCHEMA auto-create, READ_ONLY rejection, cross-catalog view create/read/rebind, view-vs-table collisions, duplicate-option last-wins)
- [x] `cargo check` (CE), `cargo check --features enterprise,private,parquet`, offline sqlx check, `./update_sqlx.sh`-equivalent safe cache regen (zero EE cache loss)
- [x] `npm run check` (0 errors), svelte-autofixer clean
- [x] Unit tests (`fork_ducklake*`, `strip_fork*`, `fork_defer_statements*`)
- [x] E2E fork lifecycle incl. before/after parent-lake evidence (above)
- [ ] Reviewer: sanity-pass the defer semantics caveats in `docs/pipeline-env-isolation.md`

## Notes

- No EE companion PR: the agent-worker endpoint serializes the same struct; new fields are `#[serde(default)]`-tolerant both ways.
- Coordination: deliberately does not touch `sql_materialize.rs` or the annotation parsers (concurrent WAP/contracts branches); `materialization.rs` only gains an additive helper.

---
## Summary by cubic
Adds isolated DuckLake data environments for forks with chain-aware defer to ancestors, UI indicators, and a safe cleanup path. Also adds fork-mode chips/banner in Ducklake settings and preserves per-lake `fork_behavior` when saving settings.

- **New Features**
  - Fork write isolation: fork-scoped `wm_fork_*` metadata schema and `__wm_forks/<forkid>/…` data path; reserved ATTACH args stripped; MySQL-catalog lakes blocked in isolated forks.
  - Read-defer per catalog to the nearest ancestor; SCD2 `<table>_current` covered; failed-after-commit tables count as fork-owned; supports orphaned `wm-fork-*` forks and orphaned ancestors.
  - Safe view→table transition keyed on live `ducklake_view` state to avoid wedging after failures.
  - Per-lake isolated/shared choice at fork creation (shared never inherits); Ducklake settings shows chips + info banner and round-trips `fork_behavior` on save.
  - Graph and details-pane show `fork` vs `deferred`.
  - Registry + cleanup: `fork_ducklake_namespace` ledger; POST `/w/{workspace}/workspaces/drop_forked_ducklake_namespaces` drops `wm_fork_*` schemas and deletes `__wm_forks/<forkid>/…`; inline on delete; retries via `schema_dropped`; pre-commit credential resolution, post-commit destroy; ancestor-chain and registration caches invalidated.

- **Migration**
  - Run the DB migration creating `fork_ducklake_namespace`.
  - No agent changes required; ensure builds include `parquet` for object-store deletion.

<sup>Written for commit 44bcd7adfa90c79b919c4bc7847e8c7da6091eb5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the CI review findings (commit 47a67f5c11):

**1. Multi-level fork-chain defer (Codex P1 / Claude P2) — fixed.**
`list_fork_defer_tables` now walks the whole ancestor chain: per table it picks the NEAREST ancestor with a `materialized` row, and `ForkDeferTable` gains `ancestor_idx` (serde-default 0 for wire compat with older agent workers) so the worker emits each defer view over the *owning* ancestor's alias instead of always the direct parent. In `root → A → B` where only root materialized `orders`, B's view now targets the root alias directly — verified live: created exactly that chain (A never ran anything), ran the consumer in B, and it read root's data through the chain-targeted view. The SCD2 `_current` detection also keys on the owning ancestor's captured schema now. Unit test added (`test_fork_defer_statements_chained_ancestors`), plus the live chain e2e.

**2. Registry misses drifted data paths (Codex P1) — fixed.**
`fork_ducklake_namespace`'s PK now includes `(storage, data_path)` (migration amended in place — it never shipped): the registry accumulates **one row per physical location ever attached** instead of keeping only the first, the once-cache is keyed on the full location, and the cleanup endpoint deletes per-row (schema drop stays idempotent). Verified live: flipped a fork's lake path in settings, ran another materialize, saw two registry rows, and the drop endpoint removed both prefixes.

**3. Child-fork cleanup authz leak (Claude P2) — false positive, no change.**
The premise is "deleter can `deleteWorkspace(child)` but gets `PermissionDenied` from `dropForkedDucklakeNamespaces(child)`". Both endpoints run the byte-identical check on the same `w_id` — `(is_fork && is_workspace_owner) || is_super_admin_email` (`workspaces_extra.rs`, delete_workspace vs drop_forked_ducklake_namespaces) — so that state can't occur: if the drop 403s, the subsequent delete 403s too, the UI `catch` returns, nothing cascades, and the registry rows survive for a later authorized cleanup. The residual case (drop returns per-lake *error strings*, e.g. storage unreachable, and the delete still proceeds) is deliberate and matches the forked-datatable sibling's semantics: the errors are surfaced as toasts and we don't want broken storage credentials to make a workspace undeletable.

Also noted from the Claude review's minor observations: the redundant re-registration inside the cleanup loop is acknowledged — harmless (`ON CONFLICT DO NOTHING`, and with the new per-location PK the specific row being processed is still deleted); left as-is to keep the cleanup path on the shared resolver.


**Bonus find from the chain e2e:** the naive chain fix regressed on a *fresh* intermediate fork — an ancestor that never attached the lake has no namespace, and its unconditional `READ_ONLY` attach failed the whole grandchild job ("DuckLake does not exist" + creation disabled). Fork ancestors are now filtered to those whose metadata schema actually exists (checked on the same catalog connection as the `fork_views` read — one round trip): a never-bootstrapped namespace can't own tables or be referenced by any persisted view, so skipping it is exact.

rubenfiszel:
Addressed the second Codex round (commit 4080806af2), all three findings were real:

**1. Fork metadata schema collides across lakes sharing one catalog (P1) — fixed.**
`fork_ducklake_metadata_schema` is now lake-scoped: `wm_fork_<mangled wid>_<mangled lake>_<hash(wid\0lake)>` (≤58 chars). Two lakes of one workspace backed by the same catalog database get distinct fork namespaces. Related and fixed in the same commit: ancestor READ_ONLY attaches now preserve the ancestor config's own `METADATA_SCHEMA` from `extra_args` (`extract_metadata_schema_arg`, last-occurrence-wins matching DuckDB) — previously a root lake living in a non-default schema would have been attached at the wrong namespace. Verified live: two lakes on one catalog DB materialized in one fork produced two distinct `wm_fork_…_main_…` / `wm_fork_…_main2_…` schemas with per-lake registry rows. Naming unit tests extended (cross-lake distinctness, 63-char budget with long lake names). The derivation change is safe to make now — the feature is unreleased, so no existing fork namespaces depend on the old name.

**2. Cleanup resolves drifted catalog from current settings (P1) — fixed.**
The registry now stores the canonical catalog identity (`<resource_type>:<resource_path>`, part of the PK so catalog drift accumulates rows like storage/path drift), and `drop_fork_ducklake_metadata_schema` connects via the STORED identity — instance identities rebuild instance creds, resource identities resolve the `$res:` in the fork's workspace — never via whatever the settings point at by deletion time. Verified live: flipped a fork's catalog `resource_path` to a bogus database, called the drop endpoint, and it still removed the real schemas in the original catalog and cleared the registry.

**3. Graph chip single-level in chains (P2) — fixed.**
`asset_graph`'s `fork_materialization` now queries `materialized_partition` across the fork plus the FULL ancestor chain (same `fork_ancestor_chain` the resolver uses), so a grandchild whose direct parent only defers still shows the amber deferred chip. Verified live: fresh `root → A → B` chain, B's graph returns `deferred` for both root-materialized assets (and `None` for a never-materialized one).

All checks green: CE + offline-sqlx cargo checks, 15 fork unit tests, and the three live e2e scenarios above (plus fork deletion back to zero residue).

rubenfiszel:
Addressed the round-3 Codex finding (commit de9965de43) — **reproduced first, then fixed**:

**Storage cleanup resolves through mutable settings (P1) — reproduced and fixed.**

*Reproduction (pre-fix binary, live stack):* forked a workspace, bulk-materialized 20k rows into storage A (MinIO bucket via `u/admin/minio_e2e` — parquet file confirmed under `demo/__wm_forks/<wid>/…`), repointed the fork's `large_file_storage` to storage B (a filesystem storage), then called `drop_forked_ducklake_namespaces`. Exactly as flagged: the endpoint returned success, deleted the registry rows, cleaned B (nothing there) — and the real fork parquet in A survived, permanently orphaned with no registry row left to drive a retry.

*Fix:* the registry now records `storage_ref` — the storage's RESOLVED identity (`<lfs type>:<resource path or root path>`) captured at registration time, part of the PK so storage drift accumulates rows like catalog/path drift. `delete_fork_ducklake_data` rebuilds the object-store target from the REGISTERED identity (reconstructing the matching `LargeFileStorage` variant and resolving its `$res:` in the fork workspace; filesystem refs resolve directly), never from the settings' current pointer. `storage_ref = ''` (registration couldn't identify a storage — no LFS configured, in which case the ducklake write itself fails at the proxy) falls back to the legacy current-settings path, best-effort.

*Verification (fixed binary, same scenario):* registry row shows `storage_ref = S3Storage:u/admin/minio_e2e`; after repointing to storage B, the drop endpoint deleted the fork prefix **from A** — zero orphaned objects, registry cleared, fork deleted with no residue.

Residual (same class as the catalog-identity fix): the identity pins the *resource path*; editing the resource's own content (e.g. its `bucket` field) in place still drifts under it — credentials must come from somewhere live. That's one indirection narrower than before and matches the catalog semantics.

Unit test added for the descriptor derivation (`lfs_entry_storage_ref`: `$res:` normalization, all LFS variants, unknown-type refusal). CE + offline-sqlx checks green, EE cache set intact.

rubenfiszel:
Addressed the round-4 Codex finding (commit d1f8051e49), plus merged main (twice — sibling pipeline PRs #9909/#9910/#9911/#9916/#9917 landed; conflicts were adjacent-prop insertions and the `Ducklake` struct gaining both `fork_behavior` and `maintenance`, all resolved keep-both).

**Fork data prefix can nest inside a sibling's (P1) — confirmed and fixed.**
Verified the premise: `validate_workspace_branch_id` is git-branch-safety only and explicitly iterates slash-separated components, and no other creation-path check restricts the charset — so `wm-fork-a/b` is a creatable fork id, and `fork_data_path` used the raw id as a path segment: `__wm_forks/wm-fork-a/b/…` sits inside `__wm_forks/wm-fork-a/`'s cleanup prefix, so deleting `wm-fork-a` would sweep the sibling's DuckLake files.

*Fix:* the data path now uses `fork_data_dir_segment(wid)` — the same mangle+hash recipe as the metadata schema, collapsing the id to ONE `[a-z0-9_]` path component (injective via the hash, e.g. `/` and `_` both mangle to `_` but hash apart; stability requirement documented alongside the schema fn). The cleanup guard recomputes the same segment. Unit test added exactly as suggested (`test_fork_data_path_prefix_isolation`): asserts `wm-fork-a/b` is a valid fork id whose data path does NOT start with `wm-fork-a`'s cleanup prefix, single-component segment, `/`-vs-`_` injectivity, determinism, and the empty-base-path shape. Safe to change the derivation now — unreleased, no fork namespaces depend on the old layout (registry rows carry the literal `data_path`, and the cleanup guard + writes derive from one function).

Post-merge validation on the combined tree: CE + parquet/duckdb cargo checks, all 16 fork unit tests, full `npm run check` (0 errors), and the new lake-maintenance interplay is exactly the case the bucket-root layout was designed for — a parent's orphan cleanup scans only its own DATA_PATH and can never see fork files.

rubenfiszel:
Addressed the round-5 Codex findings + the `cargo_test` CI failure in one commit (36bad2e2cf):

**CI `cargo_test` compile failure — fixed, and the gap in my local validation closed.**
`windmill-queue/tests/schedule_push.rs` (from the lake-maintenance PR merged in) constructs `Ducklake { … }` literally and was missing my merged `fork_behavior` field. Added it; the affected integration test passes locally. Root cause of the miss: I had only been compiling lib targets — this round I compiled **all test targets across the workspace with CI's exact feature set** (`cargo test --no-run --all --features enterprise,…,private,…`), which now passes clean, so there are no other initializer breaks hiding.

**1. Ancestor schema-existence checked in the wrong catalog (P1) — fixed.**
`inspect_fork_catalog` checked every ancestor's metadata schema against the FORK's catalog connection; a fork whose own catalog drifted after forking would misread a parent's (existing, elsewhere) namespace as missing and silently lose its defer. Ancestor schemas are now grouped by `ducklake_catalog_identity` (helper shared with the cleanup registry) and each group is checked in its OWN catalog database — the fork-identity group rides the existing connection (zero extra round trips in the common cloned-settings case), foreign groups get one connection each, and an unreachable ancestor catalog only disables that ancestor's defer (loud absent-table reads, never a job-killing attach). The filter still runs before defer discovery, so `ancestor_idx` stays consistent.

**2. Ancestor attaches drop configured `extra_args` (P1) — fixed.**
`DucklakeAncestorAttach` gains `extra_args` (serde-default; wire-compatible), populated from the ancestor's config with the fork-reserved options stripped, and the read-only attach emits them **first** so the fork-owned `DATA_PATH` / `OVERRIDE_DATA_PATH` / `READ_ONLY` / `METADATA_SCHEMA` after them always win under DuckDB's last-occurrence-wins — an `ENCRYPTED true` lake now attaches for defer, while a configured arg can't undo read-only or repoint the namespace. Unit test added asserting both the passthrough and the ordering.

Also proactively fixed a same-class latent gap found while re-sweeping the diff: `fork_behavior` was missing from the `DucklakeSettings` OpenAPI schema (the maintenance PR added its own field there), so saving lake settings in a fork through the typed client would have silently dropped the shared flag. Added to the schema + client regenerated.

Validation: all-targets `cargo test --no-run` with the CI feature matrix (clean), offline-sqlx + CE checks, 17 fork unit tests, the touched `schedule_push` maintenance test, and full `npm run check` (0 errors).

rubenfiszel:
Addressed the round-6 Codex finding (commit 6bde8052dd):

**Stale ancestor-chain cache can bypass fork isolation (P1) — real, fixed.**
`FORK_ANCESTOR_CHAIN_CACHE` cached empty chains for 60s with no invalidation on lineage mutations. The reachable window is the **dev-workspace attach** path: attaching sets `parent_workspace_id` on an EXISTING, active workspace, so a chain cached as empty just before the attach would make its ducklake resolution read as "not a fork" — writing through the shared/base lake config — until the TTL. (A brand-new `wm-fork-*` id can't realistically pre-populate the cache — resolution fails earlier at the settings lookup and the graph endpoint requires membership — but it's invalidated anyway for defense in depth.)

Fix mirrors the existing `invalidate_fork_parent_cache` precedent exactly — a `invalidate_fork_ancestor_chain_cache` called at every `parent_workspace_id` lineage mutation:
- **attach_dev_workspace** — the critical site: the workspace + all its fork descendants (their chains gained the new root);
- **create_workspace_fork** — the new fork id (closes the theoretical pre-creation caching);
- **change_workspace_id** — each reparented child (deeper descendants self-heal via TTL, matching the local convention for the sibling caches);
- **delete_workspace** — the freed id + orphaned children (an id re-created with different lineage must not inherit the old chain).

Detach needs no invalidation: it clears `is_dev_workspace` but keeps `parent_workspace_id`, so the chain is unchanged — consistent with the fork-parent cache not being invalidated there either.

Validation: all-targets `cargo test --no-run` with the CI feature matrix (clean), CE + parquet checks, all 17 fork unit tests.

rubenfiszel:
Addressed the round-7 Codex nit (commit e0c0a3b2c1):

**Descendant chain caches stale after ancestor delete/reparent (P2) — fixed.**
Deletion and workspace-id-rename reparenting now sweep the WHOLE fork subtree's ancestor-chain caches (anchored at the orphaned/reparented children, whose subtrees are intact post-delete), not just the direct children. The distinction from the sibling caches' accepted TTL self-heal is called out in a comment: a stale chain containing a dead ancestor makes the ducklake resolver drop ALL defer ancestors (all-or-nothing on broken links), which is a user-visible defer/chips outage rather than a harmless mis-route — worth the exact sweep. Attach already swept descendants (their chains gain the new root); detach still needs nothing (parent link unchanged).

To be explicit about the failure mode's ceiling either way: a stale chain can only *disable* defer temporarily — writes stay isolated in the fork namespace regardless, so this was a UX/read-availability nit, not an isolation gap.

Validation: all-targets `cargo test --no-run` with the CI feature matrix (clean), parquet cargo check, all 17 fork unit tests.

rubenfiszel:
Addressed the round-8 Codex finding (commit 2c1616b410) — real, and a good structural catch:

**Non-sidebar delete paths orphan deterministic namespaces (P1) — fixed.**
`drop_forked_ducklake_namespaces` was only invoked by the sidebar's delete flow; any other `deleteWorkspace` caller (the fork-reuse force-delete dialog, `wmill workspace delete-fork`, direct API) CASCADE-deleted the registry rows while leaving the physical schemas/data — and since namespaces are deterministic from (fork id, lake), recreating the same fork id would silently reattach the dead fork's tables.

*Fix:* the cleanup body is extracted to a shared `drop_forked_ducklake_namespaces_impl` (caller-gated, no authz of its own) and `delete_workspace` now runs it inline — after all permission/guard checks, before any row deletion, on the pool rather than the (so-far read-only) transaction. Best-effort by design: per-lake failures are logged and deletion proceeds (broken storage credentials must not make a workspace undeletable — same call as the datatable sibling and the earlier rounds' discussion), with the residual "cleanup failed AND delete succeeded" window now identical across every path instead of path-dependent. The sidebar keeps its explicit endpoint call for per-lake error toasts; the inline rerun is an idempotent no-op (empty registry, `IF EXISTS` drops). Non-fork workspaces cost one indexed no-row SELECT.

*Verified live* exactly as the review suggested: created a fork, bulk-materialized (schema + parquet on disk confirmed), deleted it via the raw `deleteWorkspace` API with NO drop-endpoint call — metadata schema, data files, and registry all gone; a recreated identical fork id would start clean.

Validation: all-targets `cargo test --no-run` with the CI feature matrix (clean), parquet + no-parquet cargo checks, 17 fork unit tests, and the live e2e above.

rubenfiszel:
Addressed the round-9 Codex finding (commit feb6db62c1) — real, and it exposed a subtlety that makes BOTH naive orderings wrong:

**Cleanup before commit can destroy a live workspace's fork data (P1) — fixed via a prepare/execute split.**
The round-8 inline cleanup ran before the delete transaction committed: a late tx failure (audit insert, commit) would leave the workspace alive with its fork namespaces destroyed and the registry rows gone. But simply moving the cleanup after the commit doesn't work either — I verified this live: post-commit, the fork's `resource` rows are already CASCADE-deleted, so the `$res:` resolution for the registered storage (and resource-type catalogs) fails, and the data files survive as orphans.

*Fix:* the cleanup is split into `prepare_fork_ducklake_cleanups` (pre-commit, strictly read-only: reads the registry rows AND resolves every catalog/storage credential into memory while the fork's resources still exist) and `cleanup_fork_ducklake_namespaces` (post-commit: guards, `DROP SCHEMA`, S3 prefix delete, registry-row delete — all driven by the pre-resolved material, no DB lookups that depend on the deleted workspace). Per-row resolution failures are carried as error strings so they surface uniformly with execution failures. The endpoint path uses the same prepare+execute pair (rows and resources exist throughout, semantics unchanged, registry stays retryable on failure).

Resulting invariants, each now unconditional:
- nothing destructive happens before the delete has committed — a failed delete leaves the fork fully intact, registry included;
- a committed delete cleans schemas AND data on every path (verified live: raw `deleteWorkspace`, no endpoint call — schema 1→0, parquet files 1→0, registry empty, zero cleanup warnings in the log);
- a post-commit cleanup failure degrades to a logged orphan (the pre-existing accepted residual), never to data loss on a live workspace.

Validation: live e2e above, all-targets `cargo test --no-run` with the CI feature matrix (clean), offline-sqlx + parquet/no-parquet checks, 17 fork unit tests.

rubenfiszel:
Addressed the round-10 Codex findings (commit 87b97f64ac). #1 and #2 were real; #3 was already safe by the listing's semantics, but the code no longer relies on that.

**1. Dev-workspace auth parity on the drop endpoint — real, fixed.** `delete_workspace` gained a parent-prod-admin gate for attached dev workspaces in an earlier round, and `drop_forked_ducklake_namespaces` kept only the base fork-owner check — so a non-prod-admin dev owner could destroy the dev's materializations through the sidebar flow and then have the deletion itself rejected, exactly as described. The gate is now extracted into a shared `require_prod_admin_for_dev_workspace` helper called by **both** endpoints, so the two checks can no longer drift (this is the second time "byte-identical inline checks" quietly diverged — the helper removes the failure mode, not just the instance). No-op for non-dev forks.

**2. Registration once-cache surviving cleanup — real, fixed.** `FORK_DUCKLAKE_REGISTERED` is now keyed **per workspace id** (`w_id -> (set of registered locations, expiry)`), and `cleanup_fork_ducklake_namespaces` — which both the endpoint and the inline `delete_workspace` path funnel through — invalidates the workspace's entry before touching anything. A same-id fork recreated within the old 60s TTL now re-registers on its first attach instead of silently skipping the insert, so its materializations always carry registry rows for their eventual cleanup. (The residual race — a fork job attaching *concurrently* with its own workspace's deletion re-populating the cache — is unchanged from before and still bounded by the TTL; it requires the fork to keep running jobs mid-delete.)

**3. Empty-path prefix listing — no bug, but hardened.** The `object_store` crate defines prefix listing on a *path-segment* basis: `list(Some("__wm_forks/<segment>"))` matches only objects whose next path component is exactly `<segment>`, never a sibling `<segment>extra…` (this is documented on `ObjectStore::list_with_delimiter`/`list` — "Prefixes are evaluated on a path segment basis"). So sibling fork segments sharing a string prefix were already excluded, including in the empty-lake-path case. That said, deletion shouldn't lean on a listing implementation detail: the cleanup now filters the listed locations to `starts_with("<prefix>/")` before issuing deletes, so a segment-boundary violation can't reach `delete_stream` regardless of the store implementation. (String-prefix-colliding segments *are* constructible across two fork ids — the mangled portion can itself end in an `_`+hex-looking tail — which is exactly why the filter is worth having even though segment-basis listing already excludes them: a sibling `<segment>extra…/file` is a different single path component, not a child of `<segment>`.)

Validation: both parquet/no-parquet compiles of `windmill-api-workspaces`, `windmill-common` + worker fork unit tests, and the full CI feature-matrix `cargo test --no-run --all` pass locally.

rubenfiszel:
Addressed the round-11 Codex finding (commit 77f8cdb63a) — real, reproduced live before the fix and verified isolated after it.

**Orphaned forks bypassing isolation — real, fixed.** `workspace_is_fork` already treats the `wm-fork-` prefix as fork-ness precisely because `parent_workspace_id` is `ON DELETE SET NULL`, but the ducklake resolution gate keyed only on chain emptiness. The gate in `get_ducklake_from_db_unchecked` now mirrors it: an empty chain only means "not a fork" when the id lacks the `wm-fork-` prefix. An orphaned fork resolves through `fork_scoped_ducklake` with zero ancestors — no defer views (there is nothing to defer to; persisted defer views from before the orphaning fail loudly at bind and flip to real tables on first materialize via the existing `fork_views` transition), but the write redirect, registration, and cleanup all still apply. `asset_graph` got the matching gate so an orphaned fork keeps its `fork` chips. Prefix-less dev workspaces cannot be orphaned — deleting their prod is blocked while one is attached (and the `is_dev ⇒ has parent` CHECK backs that) — so the prefix check is sufficient.

Live before/after repro (EE + MinIO + instance-pg-catalog stack):
- **Before (pre-fix binary):** `admins → wm-fork-orph-a → wm-fork-orph-b`, deleted `orph-a` (allowed; orphans `orph-b` with `parent_workspace_id = NULL`), ran the cloned materializing script in `orph-b` → the **shared** lake advanced from snapshot 4 to 6, no `wm_fork_%` schema, no registry row. Exactly the reported bypass.
- **After (fixed binary):** fresh orphan pair, same sequence → the shared lake's snapshot count unchanged, the write landed in `wm_fork_wm_fork_orph_*` metadata schema + `__wm_forks/<segment>/…` data prefix, registry row present, and deleting the orphaned fork cleaned both.

Validation: CI feature-matrix `cargo test --no-run --all`, fork unit tests, and both parquet modes pass locally.

rubenfiszel:
Addressed the round-12 Codex finding (commit 7faf28b4ed) — real, reproduced live before the fix and verified after it.

**Orphaned fork ancestors attached as roots — real, fixed.** `fork_scoped_ducklake` decided "is this ancestor an isolated fork or the root" purely by chain position (`i + 1 < chain.len()`), but chain position can't distinguish a genuine root from an orphaned `wm-fork-*` ancestor whose own parent was deleted (`ON DELETE SET NULL` ends the chain at it). The predicate now also keys on the `wm-fork-` prefix, same rule as the round-11 resolution gate: a last-in-chain `wm-fork-*` ancestor is attached at `fork_ducklake_metadata_schema(ancestor, lake)` + `fork_data_path(ancestor)`, not its cloned base config. Genuine roots (prefix-less) and `Shared`-behavior ancestors keep the config-namespace handling, and a prefix-less dev workspace can never be last in a chain (the `is_dev ⇒ has parent` CHECK plus the delete guard).

Live before/after repro, exactly the manual scenario suggested (root `rootr` with its own lake at `METADATA_SCHEMA 'root_r'` / `lake_rootr`, orders = widget 20 / gadget 35 / gizmo 15 → fork `wm-fork-anc-b` materializes distinctive data 2000/3500/1500 + a 4th `doohickey 999` row in its fork namespace → fork `wm-fork-anc-c` of B → delete `rootr`, orphaning B):
- **Before (round-11 binary):** `daily_revenue` in C succeeded but read the **deleted root's** lake through B's alias — result `gadget 35, gizmo 15, widget 20`, no doohickey. Wrong namespace, exactly as reported.
- **After (fixed binary):** re-ran the consumer in C — result `doohickey 999, gadget 3500, gizmo 1500, widget 2000`, i.e. the defer view now binds B's isolated namespace. The stale defer views C created under the bug self-healed: they reference the ancestor *alias*, and the fix changes what that alias attaches to, so no migration is needed.

Validation: CI feature-matrix `cargo test --no-run --all`, fork unit tests, both parquet modes pass locally.

rubenfiszel:
Addressed both round-13 Codex findings (commit 395e25a413) — both real, both reproduced live before the fix and verified after it.

**1. Inherited shared opt-out — real, fixed.** The settings clone copies the source's ducklake config verbatim, and the stamping loop only ever *added* `"shared"` — so a fork of a shared fork silently kept `fork_behavior: "shared"` even at the default isolated choice. Reproduced live: grandchild `wm-fork-sh-g` (created with no `shared_ducklakes`) carried the inherited stamp and its run advanced the shared lake from snapshot 6 to 8. Fork creation now strips all cloned `fork_behavior` stamps before applying the request's own list — sharing is a per-fork-creation choice, never inherited. Verified after: same grandchild-of-shared-fork sequence leaves no stamp, and the run lands in the grandchild's fork namespace with the shared lake untouched.

**2. Lost retry metadata on failed post-commit cleanup — real, fixed.** Reproduced live: with the catalog DB unreachable, `deleteWorkspace` committed, the `ON DELETE CASCADE` erased the registry rows, and the `wm_fork_*` schema was left orphaned with nothing recording it — a recreated same-id fork would silently reattach it. The fix keeps the post-commit ordering (the round-9 analysis stands: pre-commit destructive cleanup is worse, it destroys live data when the delete later fails) but makes the ledger durable:
- the registry **FK is removed** — rows now outlive the workspace when cleanup fails, and are deleted only after each row's schema + data cleanup succeed (already the case);
- **fork creation retries** any leftover rows for the reused id (right next to the existing `purge_stale_fork_diff_state` precedent) and **refuses to create** while any metadata schema still cannot be dropped, with the per-row errors in the message — so stale reattach is impossible and the retry path is the natural one.

Running the live failure cycle surfaced one necessary refinement: a deleted fork's `$res:` storage resource is gone forever, so leftover **data files** may never be resolvable again post-mortem — a blanket refusal would brick the id. Cleanup issues are therefore typed (`blocking` = schema/guard failures, i.e. the namespace is still attachable; non-blocking = data-file or registry-row-delete failures after the schema is gone). Creation blocks only on blocking issues; data leftovers are inert once the schema is dropped (nothing can reattach them), stay tracked by the surviving row, and the next successful cleanup of the same prefix — e.g. the recreated fork's own eventual deletion, whose prepare resolves its live cloned resource — sweeps them.

Verified after, full failure cycle: materialize in a fork → rename the catalog DB away → delete (cleanup warnings, registry row **survives**) → recreate the same id while the catalog is broken → **rejected** with the schema-drop error → restore the catalog → recreate → the sweep drops the stale schema, the data-file leftover downgrades to a warning, creation proceeds → delete the recreated fork with everything healthy → its cleanup resolves live credentials, sweeps the shared prefix, and deletes the row; registry and catalog end at zero residue.

The unreleased migration was amended in place (same file, revert/rerun locally) rather than stacked, per the PR's existing practice. Validation: sqlx offline cache regenerated via the workspace-wide procedure, CI feature-matrix `cargo test --no-run --all`, fork unit tests, and both parquet modes pass locally.

rubenfiszel:
Addressed the round-14 Codex finding (commit 0ba2a3a6d7) — real, verified live with a resource-backed catalog.

**Retry-path resolution against deleted fork resources — real, fixed with both mechanisms the review named.** The creation-time retry resolved `$res:` catalog/storage paths against the deleted fork's workspace, whose resource rows are gone — so a resource-backed catalog with a transient schema-drop failure became permanently unretryable and bricked the id. Two complementary fixes:

1. **Durable phase state.** The ledger row gains `schema_dropped`: set once the schema drop succeeds but data cleanup fails, so later retries skip the schema phase entirely — they need **no catalog credentials at all** (which may be gone for good). Registration resets the flag on re-attach (`ON CONFLICT … DO UPDATE SET schema_dropped = false`), since attaching recreates the schema — without the reset, a recreated fork's live schema would be skipped by its own eventual cleanup.
2. **Fallback connection material.** Retry-path resolution (`resolve_fork_catalog_pg` / `resolve_fork_storage`) now falls back to resolving the same `$res:` path in the workspace being forked — the deleted fork's resources were clones of a parent's, so the new parent is the natural donor. Live-workspace paths (`delete_workspace` prepare, the drop endpoint) pass no fallback and behave as before.

Live verification, exactly the suggested scenario (a `main3` lake on a **postgres-resource** catalog `$res:u/admin/dl_cat`, resource present in `admins`): fork → materialize (fork schema in the resource-backed catalog DB, registry row) → rename the catalog DB away → delete (schema drop fails, row survives, `schema_dropped = false`) → recreate the same id while broken → **rejected** (fallback resolves the parent's resource, but the catalog is genuinely unreachable — correct) → restore the catalog → recreate → the sweep resolves the catalog **via the parent's resource** (the fork's own is gone), drops the stale schema, creation proceeds → healthy delete ends clean. Also verified the phase-flag path: a row marked `schema_dropped` retries **with the catalog still unreachable** — the sweep skips the schema phase, warns about the data leftover, and creation proceeds.

Validation: migration amended in place (revert/rerun locally), sqlx cache regenerated workspace-wide, CI feature-matrix `cargo test --no-run --all`, fork unit tests, both parquet modes — all green locally.

rubenfiszel:
Addressed both round-15 findings (commit 44342ea1ee). The P1 was real — reproduced live — and it's the same class as the round-1 wedge (recorded status can't testify about physical tables), now fixed at all three surfaces.

**1. [P1] Failed fork materializations shadowing defer — real, fixed on three levels.** Reproduced live (the pre-commit-failure-after-success variant, which is even more common than the data-test one: `record_mat` *overwrites* the `materialized` row): fork materializes distinctive data → a later broken run records `failed` with `snapshot_id = NULL` → the graph flipped to **deferred** while any read actually hit the fork's own table (`CREATE VIEW IF NOT EXISTS` silently yields to it). Fixes:
- **Upsert stops destroying evidence** — `snapshot_id = COALESCE(EXCLUDED.snapshot_id, old)`: a failed run must not erase the last *committed* snapshot; the `status`/`error` columns still record the failure. (Checked the three readers: the dispatch recorder gates on `status='materialized'`, the partitions pane shows status alongside, the third is the write path.)
- **Defer discovery treats fork rows as fork-OWNED when `status='materialized' OR snapshot_id IS NOT NULL`** — a committed-then-failed table is real and must not get a defer view.
- **Catalog truth as the backstop**: `inspect_fork_catalog` (the existing single round trip that already lists the fork's live views) now also lists live *tables* in the fork namespace, and the defer list is filtered against them. This covers what SQL can't see: legacy rows recorded before this fix, and tables created by raw SQL.
- The graph chip uses the same fork-owned condition, so it now says **fork** for a committed-then-failed table — matching what a downstream query actually reads (the fork table stays authoritative; that's the consistent branch of the fix Codex offered, and it composes with the existing `fork_views`-keyed view→table transition).

After-verification on the fixed binary: same success→break→fail sequence keeps `snapshot_id` (COALESCE), the chip stays `fork`, and a downstream consumer reads the fork's distinctive values — chip and read truth agree. One note: rows recorded *before* this fix already had their snapshot nulled, so their chip corrects itself on the next successful materialize; their read behavior is already correct via the live-table filter.

**2. [P2] Stale cascade wording — fixed.** The doc's registry signature line, the "rows otherwise die via FK cascade" sentence, and the `SidebarContent.svelte` comment all now describe the FK-less ledger as it is.

Validation: sqlx cache regenerated (three superseded hashes pruned), CI feature-matrix `cargo test --no-run --all`, fork unit tests, both parquet modes — green locally.

rubenfiszel:
Follow-up (commit 44bcd7adfa): the workspace-settings ducklake editor is now fork-aware. In a fork it shows a banner explaining the two data-environment modes and a per-lake chip — emerald **isolated** / amber **shared with parent** (same colors as the graph chips) — with tooltips describing what edits mean in each mode. The choice itself stays creation-time-only.

This also fixes a real round-trip bug found while wiring it: `convertDucklakeSettingsToBackend` rebuilt each lake explicitly and dropped `fork_behavior`, so **any ducklake-settings save in a shared fork silently flipped the lake back to isolated**. The field is now part of `DucklakeSettingsType` and preserved on save — verified in the browser: edited a lake's data path in a fork with a shared `main`, saved, and the `shared` stamp survived in the settings JSONB. Screenshot added to the PR body.

PR #9916, opened 2026-07-03
feat(ducklake): scheduled lake maintenance (expiry, compaction, orphan cleanup)

rubenfiszel:
## Summary

DuckLake lakes currently grow unboundedly: every materialization advances a snapshot and frequent incremental runs proliferate small parquet files, and nothing in the product ever calls `ducklake_expire_snapshots`, `ducklake_merge_adjacent_files` or orphan-file cleanup. Iceberg/Delta ecosystems treat compaction/retention as table-stakes ops — and Windmill *is* a scheduler — so this adds managed, per-lake scheduled maintenance as an **enterprise feature**: snapshot expiry with a configurable retention window, adjacent-file compaction, and orphaned-file cleanup, configured in the workspace ducklake settings and observable as normal jobs (run history is the audit trail).

This closes open decision #2 ("Snapshot retention / compaction") in `docs/ducklake-materialization.md`.

## How it works

- **Config lives on the lake**: `Ducklake.maintenance` (`enabled`, cron `schedule` — default daily at 03h UTC with a deterministic per-lake minute offset, `retention_days` — default 7, `compaction`, `orphan_cleanup`) in `workspace_settings.ducklake`, edited from the ducklake settings UI.
- **Managed schedule per lake**: saving the settings syncs `schedule` rows at the reserved path `f/ducklake_maintenance/<lake>` **inside the same transaction** as the settings update (EE logic; a failed push fails the save — nothing reconciles a half-applied state later). Identity is re-stamped from the saving admin on every save; the row pins `tag = 'duckdb'` so `preview_tags_override` can't reroute the job.
- **Execution rides the normal job infra**: `push_scheduled_job` gets a reserved-path branch that pushes a raw-code DuckDB job (generated by EE codegen) instead of resolving a deployed script — same pattern as worker init scripts (`JobPayload::Code`). The generated script `ATTACH 'ducklake://<lake>'`s through the normal transform, runs expire → merge → cleanup → orphan-delete, and returns one summary row (`expired_snapshots`, `compacted_file_groups`, `cleaned_files`, `orphaned_files_deleted`, `remaining_snapshots`) as the job result.
- **CE gating** follows the `pipeline_advanced_oss` module-swap pattern (`windmill-queue/src/ducklake_maintenance_oss.rs`): enabling maintenance requires EE; the codegen stub returns `NotFound` deliberately so a schedule row left over from an EE period is **auto-disabled with `schedule.error` recorded** by the post-completion scheduler instead of retry-looping. Previously-enabled maintenance is tolerated at save time (license lapse must not brick unrelated ducklake settings edits) and can be saved off without a license.
- **Guard rails**: the schedule API rejects create/edit/delete/setenabled under the reserved prefix; both list endpoints, the workspace tarball export (`wmill sync pull`) and fork schedule-cloning exclude it (with `starts_with`, not `LIKE` — the prefix contains `_`); `set_default_error_handler` still applies handlers to managed rows but skips their git-sync callback. Lake names are now validated (`[A-Za-z0-9_-]+`) at save — they're interpolated into `ATTACH`, generated SQL and the CHECK-constrained schedule path.
- **Safety margins**: physical deletion (`cleanup_old_files`) uses a 1-day `older_than` margin rather than `cleanup_all` so a long-running reader pinned to a just-expired snapshot isn't yanked mid-scan; orphan deletion uses `max(retention, 1 day)` so a concurrent materialization's freshly-written parquet is never collected. Interval arithmetic is generated as `(now()::TIMESTAMP - INTERVAL …)::TIMESTAMPTZ` because TIMESTAMPTZ−INTERVAL needs DuckDB's ICU extension, which the bundled FFI build doesn't ship.
- **Companion EE change**: the S3 proxy had no bucket-level `POST ?delete` (S3 DeleteObjects) route, so DuckLake could never physically delete files through the proxy — added with per-key DELETE permission checks (checked before any deletion) and bounded-concurrency fan-out. See the companion PR.

## Changes (this repo)

- `windmill-common/src/workspaces.rs` — `DucklakeMaintenance` settings model + reserved-path helpers + lake-name validation (only what CE endpoints need: serde shape, guards, filters)
- `windmill-queue/src/ducklake_maintenance_oss.rs` (new) + module swap; `push_scheduled_job` only dispatches reserved-prefix schedules to the module pair — payload construction (config load **through the caller's tx**, required for `change_workspace_id` correctness; SQL codegen; tag pinning; default cron) lives entirely in the EE module, the CE stub returns `NotFound`
- `windmill-api-workspaces` — lake-name validation + same-tx schedule sync in `edit_ducklake_config`; fork clone exclusion
- `windmill-api-schedule` — reserved-prefix guards + list filters
- `windmill-api/src/workspaces_export.rs` — tarball export exclusion
- `windmill-api/openapi.yaml` — `DucklakeSettings…maintenance` schema
- `DucklakeSettings.svelte` — Maintenance column + popover (retention/cadence/toggles, EE gating, schedule-health badge surfacing `schedule.error`, runs deep-link), incl. the copy that **time-travel reads `AT (VERSION => n)` and recorded snapshot references older than the retention window become non-queryable**
- Cargo feature wiring: `windmill-api-workspaces/private` → queue+audit, `windmill-api/private` → object-store (makes crate-standalone EE checks build)

## Validation

- `cargo check` (CE) and `cargo check --features enterprise,private[,parquet]` (workspace root) clean; `./update_sqlx.sh` with EE-cache preservation verified (zero net deletions vs main)
- EE unit tests: 5 for the SQL codegen (defaults, toggles, retention-0 margin, invalid names, well-formedness incl. paren balance + ICU-free interval maths) + 2 for the proxy DeleteObjects parser
- `npm run check` 0 errors; client regenerated
- Reviewed via the local review policy; all findings addressed (LIKE-wildcard sweep of user schedules under near-miss folders, CE license-lapse dead-end, sequential batch deletes, vacuous test)

## E2e (live EE stack: MinIO workspace storage, postgres catalog, duckdb worker)

- Saved settings → managed schedule row created (`f/ducklake_maintenance/main`, tag `duckdb`, cron default `0 21 3 * * *` with per-lake offset)
- 6 `CREATE OR REPLACE`-style materializations + appends → scheduled maintenance run **green through the real S3 proxy**: `expired_snapshots: 9, compacted_file_groups: 1, remaining_snapshots: 2`
- Physical deletion proof (same functions, accelerated margins): **14 → 2 parquet objects in MinIO**, `files_physically_deleted: 12`, data intact (orders 1000 rows, events 600 rows)
- Guards verified: managed row hidden from `list_schedules`, `create` at the reserved path rejected, subsequent runs idempotent (all-zero summary), no `schedule.error`
- One free run post-CE-downgrade (an already-queued occurrence carries its raw code) — known and accepted; the *next* push auto-disables the schedule with the error recorded

## Test plan

- [ ] Enable maintenance on a lake in workspace settings → a run appears at the configured cadence with a summary-row result
- [ ] Snapshot count drops after runs once snapshots age past retention; parquet files merge/shrink (subject to the 1-day reclaim lag)
- [ ] Disabling maintenance clears the queued occurrence; deleting the lake removes the schedule
- [ ] A user schedule under a near-miss folder (e.g. `f/ducklake-maintenance/x`) is unaffected by saves and still listed/exported
- [ ] CE: enabling maintenance is rejected; previously-enabled maintenance can be saved off

## Screenshots

Maintenance popover (defaults) and saved state with health check + runs deep-link:

ducklake-maintenance-popover

ducklake-maintenance-saved

Runs page filtered on the managed schedule (audit trail) and a run's summary result:

ducklake-maintenance-runs

ducklake-maintenance-run-detail

CE-gated state (no enterprise license — toggle disabled with EE badge; an already-enabled config can still be switched off):

ducklake-maintenance-ce-gated

---
## Summary by cubic
Adds managed scheduled maintenance for DuckLake lakes (enterprise): automatic snapshot expiry, small-file compaction, and orphaned-file cleanup as per-lake jobs. Prevents unbounded growth and records runs as an audit trail; enable per lake in workspace settings.

- **New Features**
  - Per-lake maintenance config in DuckLake settings (enable/disable, cron cadence, retention_days, compaction, orphan_cleanup; OpenAPI updated).
  - Managed schedules under `f/ducklake_maintenance/<lake>` are created in the same transaction as settings; the schedule API rejects create/edit/delete/toggle under this prefix; list/export/fork exclude it using `starts_with`; default error-handler updates skip git-sync for these rows.
  - Runs execute via generated DuckDB code (tag `duckdb`); config is read through the caller’s transaction; each job returns a one-row summary (expired snapshots, compaction groups, cleaned/orphaned files, remaining snapshots).
  - CE/OSS: enabling requires enterprise; previously-enabled configs can be saved off without a license; saving off deletes the managed schedule row and any queued occurrence; stale managed rows auto-disable; pre-existing user schedules under the reserved prefix fall through to normal script resolution (hidden from list/export, immutable until renamed).
  - Safety/validation: default cron is daily at 03:00 UTC with a deterministic per-lake minute; lake names validated as `[A-Za-z0-9_-]+`; physical deletion uses a ~1 day safety margin; orphan cleanup uses `max(retention, 1 day)`.
  - UI: maintenance popover in DuckLake settings (toggles, retention, cadence), schedule health badge, and deep-link to runs.

- **Refactors**
  - Moved maintenance payload construction into the EE module; the OSS stub returns None (fall-through) so pre-existing user schedules keep working while stale managed rows auto-disable.

<sup>Written for commit 3d0d29f7515706925772439ecc2b06c01fccfab0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the Codex review (both P1s were auth-contract documentation on the new public helpers):

1. `sync_ducklake_maintenance_schedules` (OSS stub + EE impl) now documents explicitly that it is **not an authorization boundary**: the caller must already have enforced workspace-admin on `w_id` with the authenticated caller's identity, as `edit_ducklake_config` does via `require_admin` (its only call site). The EE impl additionally requires the identity to resolve to a workspace member.
2. `get_ducklake_raw` → **`get_ducklake_raw_unchecked`**, matching the existing `get_ducklake_from_db_unchecked` convention in the same file, with a doc comment stating it performs no permission checks and is for internal callers (schedule-tick payload construction) that never expose the config to a user.

No behavior change; EE + CE `cargo check` clean. Companion doc commit on the EE PR.

rubenfiszel:
Addressed the new Codex/Pi P1 (reserved prefix hijacking pre-existing user schedules at `f/ducklake_maintenance/…`). The two destructive halves are fixed; the cosmetic residual is explicitly acknowledged:

**Fixed:**
1. **Tick dispatch falls through instead of failing**: `build_maintenance_schedule_payload` now returns `None` when the path's lake has no enabled maintenance config, and `push_scheduled_job` continues to normal `script_path` resolution. A pre-existing user schedule under a real `ducklake_maintenance` folder keeps running its script; a stale *managed* row (lake deleted, or CE downgrade) still dies cleanly — script resolution returns `NotFound`, which auto-disables it with `schedule.error` recorded.
2. **Sync never sweeps by prefix**: the managed row set is now derived from config — paths of lakes with maintenance enabled in the previous or new settings — so `clear_schedule`/`DELETE` can only ever touch rows this feature created. Verified live: a row inserted at `f/ducklake_maintenance/legacy_user` survives enable/disable/re-enable saves untouched while the managed `…/main` row is reconciled.

Regression tests added in `windmill-queue/tests/schedule_push.rs`: fall-through to `script_path` when no config (runs in CE and EE builds), and managed dispatch (kind `preview`, `runnable_path` = schedule path, `duckdb` tag, generated SQL with the configured retention) under `cfg(private)`.

**Acknowledged as accepted-improbable (not fixed):** the list/export/fork visibility filters and the create/edit/delete/setenabled guards remain purely prefix-based, so a hypothetical pre-existing schedule under that exact folder is hidden from the schedules list/tarball export and immutable via the schedule API (it keeps *running*, and nothing deletes or disables it). Making those five query sites config-aware would mean a `workspace_settings` JSON subquery per row for a collision that requires a pre-existing user folder literally named `ducklake_maintenance` containing schedules — we judge that combination unlikely enough to accept, and the failure mode is now non-destructive and recoverable (rename via direct DB/API if it ever occurs).

rubenfiszel:
**Ack — accepted limitation, won't fix in this PR** (re: latest Codex P1, `reserved-prefix schedules remain hidden/unmanageable`)

Decision: the list/export/fork filters and the create/edit/delete/toggle guards stay purely prefix-based. A schedule that existed at `f/ducklake_maintenance/<name>` before this PR ships remains fully functional (it runs its script via the tick-time fall-through, is never deleted or disabled — both regression-tested), but is hidden from the schedules list/tarball export and immutable through the schedule API.

Why we accept it rather than fix it:
- **Probability**: it requires a workspace to already contain a user folder literally named `ducklake_maintenance` with schedules inside, created before this feature existed. We know of no such workspace and consider the combination vanishingly unlikely.
- **Severity after the fixes**: the failure mode is purely cosmetic/administrative — no data loss, no silent disable, no missed runs. That is P2 territory, not release-blocking.
- **Cost of the alternatives**: a discriminator column is a schema migration on a hot table plus changes at all five filter/guard sites; a rename migration must also rewrite queued occurrences' `v2_job.trigger` references or the renamed schedule silently stops re-chaining — real upgrade-path machinery for a hypothetical. Heuristic discriminators (`script_path == path AND tag='duckdb'`) are spoofable and worse than a documented reserved namespace, which is the same pattern as the existing `init_script_`/`periodic_script_` reserved job paths.
- **Escape hatch**: if the collision ever materializes, the schedule keeps running and can be renamed out of the namespace with a one-line SQL `UPDATE schedule SET path = ...` (or via support); nothing is lost in the meantime.

Reserved-namespace precedent + the reasoning above are documented on `DUCKLAKE_MAINTENANCE_PATH_PREFIX` in the code. If this ever bites a real workspace we'll ship the rename migration then, with the queued-job trigger fix-up.

rubenfiszel:
Addressed the latest Codex P1 (`CE save-off leaves managed maintenance schedules queued`) — a good catch, distinct from the acked visibility residual:

The OSS `sync_ducklake_maintenance_schedules` stub now clears and deletes managed rows for lakes whose maintenance was enabled in the previous config but is no longer — `clear_schedule` drops the already-queued occurrence (so a maintenance job pushed under EE can no longer run after the admin disabled it post-downgrade) and the `schedule` row is deleted. As in the enterprise implementation, the removed set is derived from config, never from the path prefix, so pre-existing user schedules stay untouched.

Regression test added (`test_sync_disable_clears_managed_row_and_queued_occurrence` in `windmill-queue/tests/schedule_push.rs`): seeds a managed row with a queued occurrence, saves maintenance off, asserts the queue is empty and the row is gone. It runs in **both** builds — against the CE stub without features and against the enterprise reconciler with `enterprise,private` — and passes in both.

PR #9919, opened 2026-07-04
feat(pipelines): capture violating-row samples for data tests

rubenfiszel:
## Summary

`// data_test` checks on managed DuckDB materializations previously reported only violating-row **counts** — on failure the user saw `✗ unique(id) — 3 violating row(s)` and had to re-run the SELECT by hand to see which rows. Each check's probe now also captures a bounded **sample of its violating rows** (≤20 rows, ≤50KB serialized per test) and surfaces it in the **structured job result**, while the error text stays counts-only plus a pointer at the payload.

The sample is optional by contract at every layer: enforcement reads only the count, and a NULL / size-capped / unparseable sample degrades to "no sample" — it can never flip pass/fail. No annotation-grammar change (the frontend/backend parser parity corpus is untouched).

## Changes

- **Codegen** (`sql_materialize.rs`): each check compiles to a one-row probe `(v, s)` counting and sampling its violating rows in a single scan — `to_json(list(row) FILTER (rn <= 20))::VARCHAR`, size-guarded with `strlen` (bytes) and **dropped, never truncated**, when over 50KB. Built-ins mirror their count queries' partition/SCD2 scope; `unique` samples `{value, count}` pairs at its count's grain so count and sample can't contradict; partitioned targets `EXCLUDE` the synthetic `_wm_partition` column (same rule as schema capture). Probes are lifted into cross-joined one-row CTEs (DuckDB rejects subqueries inside struct literals).
- **Executor** (`duckdb_executor.rs`): the sample rides as a JSON *string* through the summary row and is parsed only in `extract_data_tests` — expanded rows would be visible to the key-recursive `extract_i64(result, "rows"/"snapshot_id")` scans, which a user column of the same name could corrupt. On failure the worker returns `Error::ExecutionRawError` with `{message, name, step_id, data_tests: [{test, violating, sample?}]}` (samples on failed tests only), so the failed job's result carries `error.data_tests` after the queue's `WrappedError` wrap; the run description still comes from `message`. The password-sanitize catch-all special-cases `ExecutionRawError`, sanitizing string **leaves** of the payload (a secret containing `"` or `\` is JSON-escaped in the serialized text, so a substring pass would miss it). `materialized_partition.error` stays counts-only text.
- **FFI gating tests** (`windmill-duckdb-ffi-internal`): in-memory tests against the bundled engine prove the probe shape — `to_json` availability, row-as-struct alias, zero rows → NULL, byte-cap degrade, exotic types (INTERVAL, HUGEINT, BLOB, nested, TIMESTAMPTZ), and the exact codegen rows-query shapes. A probe runtime error would fail the summary read after the write commits, so these gate any change to the shape.
- **UI** (`DisplayResult.svelte`, `DataTestsResult.svelte`): the failure path prefers structured `result.error.data_tests` (falling back to text-parsing the breakdown for results predating the payload); each failed test with a sample gets an expandable `AutoDataTable` ("sample of the violating rows (N of M, unordered)"), with CSV download for free.
- **Docs** (`docs/ducklake-materialization.md`): sample mechanics — bounds, drop-not-truncate, optional-by-contract, string-through-FFI rationale, unique's grain, LIMIT-without-ORDER-BY nondeterminism.

## WAP reconciliation (#9911, merged) — EE companion PR

#9911 merged to main mid-flight, moving EE data-test enforcement into an in-transaction `error()` guard (`pipeline_advanced_ee.rs::data_test_guard_sql`) that consumed `DataTestCheck.violating` — the field this PR renames to `probe`. That is what broke `cargo_test` on the first CI run (CI substitutes EE code at main'''s `ee-repo-ref.txt`, which is post-WAP). Resolved here:

- main merged into this branch (clean auto-merge; #9911'''s `MaterializePlan` plumbing and this PR'''s probe rewrite touch disjoint regions, and all 39 parser + 5 worker tests pass together)
- companion EE PR windmill-labs/windmill-ee-private#646 adapts the guard to extract the count as `(SELECT v FROM (<probe>))`; `ee-repo-ref.txt` bumped accordingly
- semantics: **samples exist only on the commit-then-test path.** Under the EE WAP guard a violation aborts pre-commit and the slice rolls back, so the violating rows cease to exist — WAP failures stay counts-only by construction (the guard'''s in-SQL breakdown), and the guard never reads the probe'''s sample column. Documented in `ducklake-materialization.md`.
- `cargo check --features enterprise,private` and the 5 EE guard/finalize tests pass against the bumped pin

## Test plan

- [x] `cargo test -p windmill-parser --lib sql_materialize` — 36 pass (probe shapes, partition/SCD2 scoping, EXCLUDE, unique grain, summary CTE assembly)
- [x] `cargo test -p windmill-worker --lib --features duckdb data_test` — 4 pass (sample dual-parse, degrade-to-None never flips `violating`, counts-only breakdown)
- [x] FFI in-memory gating tests — 6 pass against the bundled engine
- [x] `cargo check` (no features) and `npm run check` (0 errors) clean
- [x] E2E on a live dev stack (filesystem-storage ducklake, duckdb-featured worker): failing run returns `success: false`, counts-only `error.message` + hint, `error.data_tests[].sample` with the violating rows (built-ins + custom `f/tests/...` escape hatch); `materialized_partition.error` stays counts-only; passing run succeeds with `sample: null` on every test; failure path re-verified after the leaf-sanitize rework
- [x] UI verified via Playwright on the run page — screenshots below

## Screenshots

Failed run — checklist with per-test "view sample", two samples expanded (the custom test's violating row, and `not_null`'s), counts-only error text below:

result-panel-samples

Full run page for the failed run (unique's `{value, count}`-grain sample expanded):

failed-run-samples

Passing run — unchanged checklist, no sample buttons:

passing-run-checklist

---
## Summary by cubic
Adds bounded samples of violating rows to DuckDB data tests so failures are actionable without reruns; pass/fail remains count-based. Samples appear on commit-then-test runs in the structured job result and UI; EE WAP remains counts-only.

- New Features
  - Each test compiles to a one-row probe returning `{v, s}` (count and JSON-string sample, ≤20 rows, ≤50KB); `unique` samples `{value, count}`; partitioned targets `EXCLUDE` `_wm_partition`.
  - Worker returns `ExecutionRawError` with `data_tests: [{test, violating, sample?}]` (samples only for failed tests); error message stays counts-only and points to the payload; payload is in the job result and failed-job log, and is available to error handlers.
  - UI renders an expandable sample table per failed test; `DisplayResult` reads structured `data_tests` from the summary row or `error.data_tests`, falling back to legacy text when needed.

- Bug Fixes
  - Enforce a byte-accurate 50KB cap and drop oversized samples (never truncate).
  - Sanitize string leaves inside structured errors; sample parsing accepts JSON strings or native arrays and degrades invalid samples to “no sample”.
  - EE guard reads only counts; samples exist only on commit-then-test runs; EE repo ref updated after the guard adaptation merged.

<sup>Written for commit c4697d8a2c1c8587a1baf811fd65af5956108b97. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed both reviews:

**Claude review P1 (`DataTestCheck.violating` → `probe` vs the EE WAP guard)** — confirmed real, and it was exactly the `cargo_test` CI failure: #9911 merged to main mid-flight and CI substitutes EE code at main's `ee-repo-ref.txt` (post-WAP), so `data_test_guard_sql` compiled against the renamed field. Fixed by merging main into this branch and shipping companion PR windmill-labs/windmill-ee-private#646, which adapts the guard to `(SELECT v FROM (<probe>))` — as the review noted, the guard needs the count, not the whole `(v, s)` probe. `ee-repo-ref.txt` is bumped; `cargo check --features enterprise,private` and the 5 EE guard/finalize tests pass against the new pin. One semantic consequence, now documented: under the EE WAP guard a violation rolls the slice back, so violating rows cease to exist — WAP failures stay counts-only by construction; samples surface only on the commit-then-test path.

**Claude review P2 (`_wm_v`/`_wm_rn` collision in custom bodies)** — acknowledged as accepted risk under the `_wm_` reserved-prefix convention (same class as the pre-existing `_wm_target`/`_wm_partition` reservations).

**Codex P1 (samples in worker logs via `add_completed_job_error`'s payload dump)** — that log line serializes the full error value for *every* failed job, and arbitrary user data already flows through it today: the `Error::ExitStatus` arm logs the failed process's own `result.json` verbatim (plus the last 600 chars of job logs), and the AI executor's `MaxIterError` logs full agent conversations through the same wrap. Worker logs sit in the same trust domain as the DB, where the identical payload is stored as the job result regardless. This payload is actually narrower than those precedents: hard-bounded (≤20 rows / ≤50KB per test) and password-sanitized at the string-leaf level before leaving the executor, which the `ExitStatus` path is not. If we want to trim that log line, the right place is `add_completed_job_error` for all errors — out of scope here.

rubenfiszel:
Both follow-up review nits addressed:

- **Codex P2 (doc/code mismatch)**: `ducklake-materialization.md` now states precisely where sample data appears — the counts-only *message* feeds run descriptions, error-handler/alert payloads, and `materialized_partition.error`; the structured payload is the job result and, like every failed job's result, also appears in the worker's failed-job log line (`add_completed_job_error`), the same pre-existing path that logs a failed process's own output.
- **Pi P2 (`[ee]` title prefix)**: already applied — the title was retitled to `[ee] feat(pipelines): …` when the EE companion (windmill-labs/windmill-ee-private#646) was opened; the review snapshot predated the rename.

rubenfiszel:
Docs claim fixed (finding confirmed): on EE, `apply_completed_job_error_handlers` does pass the full wrapped result — samples included — as the error-handler job's args (`sanitize_result` reshapes without redacting). The docs now state that the payload follows every failed job's result wherever it already goes (worker failed-job log line, and on EE the global/workspace error-handler args — the same paths that carry a failed process's own `result.json`), while `error.message`, run descriptions, and `materialized_partition.error` stay counts-only. Kept as behavior rather than redacting: handlers already receive complete results for every other error type, a handler that formats only `error.message` (the common alert shape) remains counts-only, and one that wants the offending rows can deliberately read `error.data_tests` — e.g. to include them in the alert.

PR #9930, opened 2026-07-05
feat(pipelines): on_schema_change write guardrails + data_test deploy validation

rubenfiszel:
## Summary

Two backend guardrails found while dogfooding data pipelines end-to-end (companion CLI fixes in #9926, #9929; report driven from the same session). Both target managed DuckLake materialization.

**F18 — `on_schema_change=warn|ignore|fail|sync` write semantics.** The biggest correctness trap in the pipeline stack today: for a `key=` merge (and append / partitioned replace) the table schema is fixed at first create and the write is a **positional** `INSERT`, so renaming/adding/reordering a column in the SELECT lands data in the wrong column silently (or binder-errors confusingly). `on_schema_change` was previously a save-time-warnings-only knob (`warn`/`ignore`); this extends it to control the write itself.

**F14 — `data_test` deploy validation.** A `// data_test` line whose right-hand side doesn't parse was dropped silently (fail-safe) — the wrong default for a data-quality assertion, since a typo disables the check with no signal. Now warned at deploy.

## Changes

- **Parser** (`asset_parser.rs`): `OnSchemaChange` gains `Fail` and `Sync` variants + `fail|sync|ignore` parse grammar (unknown → `warn`, fail-safe). Applies only to the persist-and-mutate strategies (partitioned replace, merge, append); whole-table replace rebuilds each run and scd2 has no positional write, so for scd2 `sync` degrades to `warn`.
- **Codegen** (`sql_materialize.rs`): a drift probe (`DESCRIBE` the SELECT vs the table, `EXCEPT`-diff by column name, `_wm_partition` excluded).
  - **warn** (default): positional write proceeds; a `schema_drift` `{added, removed}` column is folded into the materialize summary row.
  - **fail**: an in-txn guard (`SELECT CASE WHEN <drift> THEN error(…) ELSE 'ok' END`, both branches VARCHAR) runs right after `BEGIN TRANSACTION` and before any mutation on all three arms; classified `MaterializeStmtKind::Write` so EE's write-audit-publish reassembly keeps it in-txn before COMMIT.
  - **sync**: emits `INSERT … BY NAME` + a sentinel slot after BEGIN, for the executor to fill with `ALTER TABLE … ADD/DROP COLUMN`.
  - **ignore**: no guard, pre-guardrail behavior.
- **Executor** (`duckdb_executor.rs`): parses `schema_drift` from the summary and logs a loud warning + folds it into the job result (warn); runs a separate FFI probe (same interpolation/ATTACH transform as the main query) diffing SELECT vs `information_schema.columns` and splices the ALTER DDL at the sentinel (sync). Probe failure is fatal; a not-yet-created table yields zero rows → no-op skip.
- **F14** (`asset_parser.rs` + `windmill-api-scripts/src/scripts.rs`): `count_malformed_data_tests` reuses the parser's leading-comment-block scan + `data_test` grammar (no second grammar) to count lines that fail to parse; deploy warns per malformed line (warning, not rejection — matches the other pipeline deploy checks and can't false-positive a deploy).

## Known boundary (documented + tested)

`warn`/`fail` drift detection is **name-set based** (added/removed columns) — it catches add/remove/rename, which is what a positional INSERT misaligns on, but by design **not** a pure reorder of same-named columns (identical set → no added/removed). Reorder-safety is exactly what `sync` provides (INSERT BY NAME). The set difference is kept deliberately over an ordered comparison so the `fail` guard cannot false-positive-abort a correctly-aligned write. Documented on `OnSchemaChange` + at `drift_lists`, and locked by `fail_guard_drift_is_name_set_based_not_ordered`.

## Test plan

- [x] `cargo test -p windmill-parser` → 147 pass (codegen: fail-guard-in-txn-before-write, sync BY NAME + sentinel, warn summary field, whole-table/scd2 opt-out, name-set boundary; parse: fail/sync variants + junk→warn; F14 malformed-count).
- [x] `cargo test -p windmill-worker --features duckdb,parquet duckdb_executor::tests::` → 73 pass (materialize fail/sync/warn/whole-table build tests; `parse_sync_drift`, `extract_schema_drift` FFI-shape tests).
- [x] `cargo check -p windmill-parser -p windmill-worker -p windmill-api-scripts --features duckdb,parquet` clean.
- [ ] **Manual (live DuckLake, not automatable here):** deploy a managed partitioned materialize with `on_schema_change=fail`, run once to create the table, add a column to the SELECT, re-run → confirm abort with the drift error and the prior snapshot intact. Repeat with `sync` → confirm the `ALTER … ADD COLUMN` is spliced and `BY NAME` populates the new column on a partitioned target without cross-wiring. Deploy a script with a malformed `// data_test` line → confirm the deploy warning fires and the script is still accepted.

## Deferred (follow-ups from the same finding set)

- **F16 (producer-break warning)**: a proper producer-side check needs the producer's real output columns, which only exist post-run — deploy-time static inference would false-positive on `SELECT *`. Best done as a run-time signal after schema capture; scoped separately.
- **F10 (backfill upstream-missing warning)**: the backfill endpoint is EE (`backfill_ee.rs`), so it needs the EE PR workflow.

---
## Summary by cubic
Adds write-time schema guardrails for managed DuckLake materializations and a deploy-time warning for malformed `// data_test` lines. This prevents silent column miswrites and makes schema drift visible, with an opt-in auto-sync.

- **New Features**
  - `on_schema_change=warn|ignore|fail|sync` now governs append, merge, and partitioned replace; whole-table replace and scd2 are unchanged.
    - warn (default): positional write; a `schema_drift {added, removed}` field is added to the summary and loudly logged in job output.
    - fail: in-transaction guard runs before any mutation and aborts when the SELECT vs table column set differs (name-set based; pure reorders are not flagged).
    - sync: pre-pass probe diffs SELECT vs table via `information_schema`, executor injects `ALTER TABLE` DDL at a sentinel, and the write uses BY NAME; no-op if the table doesn’t exist.
    - ignore: previous behavior (no guard).
  - Drift detection compares column name sets (added/removed), not order; use `sync` for reorder safety.
  - Deploy warns when header `// data_test …` lines fail to parse; scripts still deploy.

<sup>Written for commit 5b00802406fdeb5feefe44a10e3db505cfdd2d75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the CI review + build failure:

- **`cargo_test` failure (EE build)** — `build_wrap_blocks` gained an `on_schema_change` parameter (CE parser), which broke the EE WAP test caller in `pipeline_advanced_ee.rs`. Fixed in companion PR windmill-labs/windmill-ee-private#647 (passes `OnSchemaChange::Warn` — no guard, preserving the WAP-structure test); `ee-repo-ref.txt` bumped to point at it. Verified `cargo check -p windmill-common --features enterprise,private,parquet` and the EE `pipeline_advanced` tests (5 pass) locally. Title now prefixed `[ee]`.
- **Codex P1 (sync prepass skips workspace macros)** — the `on_schema_change=sync` pre-pass DESCRIBE ran before `inject_workspace_macros`, so a SELECT calling a workspace macro or using a `// use` library would fail the fatal probe. The probe's block list now goes through `inject_workspace_macros` (same macros/`use_libs` as the main query, post-ATTACH) before `compute_sync_alter_ddl`.
- **Pi P2 (unescaped `'` in guard error string)** — `target_qualified` is now single-quote-escaped where it's interpolated into the `error('…')` string literal (matches the `quote_lit` pattern).

Tests after the changes: `windmill-parser` 147 pass, `windmill-worker` duckdb_executor 73 pass, EE `pipeline_advanced` 5 pass, all compiling clean with the EE features.

rubenfiszel:
Addressed the follow-up Codex P1 (BigQuery credentials in sync pre-pass): the `on_schema_change=sync` probe now creates and holds `UseBigQueryCredentialsFile` (setting `GOOGLE_APPLICATION_CREDENTIALS` + the temp creds file) for any `TYPE bigquery` `$res:` ATTACH, kept alive across `compute_sync_alter_ddl`, mirroring the main query path — so a materialized SELECT reading a BigQuery ATTACH no longer fails the fatal probe DESCRIBE. Worker tests: 73 pass; compiles clean with `duckdb,parquet`.

PR #9934, opened 2026-07-05
fix(pipelines): order data_test relationships refs before the tested script in a cascade

rubenfiszel:
## Summary

**HD-1** from the pipeline "hidden dependencies" work package. A `// data_test relationships <col> -> ducklake://<lake>/<other>.<col>` (and, best-effort, a custom `// data_test <script>` whose body reads another asset) depends — at test time — on the referenced asset already being materialized. That dependency was **not** in the execution DAG, so a cold cascade could run the tested script before the referenced asset existed and hard-fail:

```
Catalog Error: Table with name dim_customers does not exist!
```

Re-running after the dimension materialized passed (`violating: 0`) — a pure ordering bug.

This adds an ordering-only "must-run-after" edge `producer(referenced_asset) → testing_script`, resolved server-side and fed into the client cascade topo-sort, so a full-pipeline / bounded run orders the referenced dimension first. When the referenced asset has **no in-pipeline producer** (an external table), no edge is added — the missing dependency is real and the runtime error is the correct signal.

## Changes

- **Backend** (`windmill-api-assets/src/lib.rs`): new `test_edges: Vec<TestEdge>` on `AssetGraphResponse`. `asset_graph` builds a producer map from the write edges it already computes, then for every pipeline member's `DataTest::Relationships` (and `DataTest::Custom`'s parsed reads) resolves the referenced asset's producer(s) and emits `producer → testing_script`. Self-edges (a script testing its own output) and duplicates are dropped; no producer ⇒ no edge. `skip_serializing_if` keeps the response shape byte-identical when empty.
- **Frontend cascade orderer** (`AssetGraph/boundedCascade.ts`): `buildLineageDag` consumes `test_edges` as `asset → testing_script`, routed **through** the referenced asset node so the existing `producer → asset` write edge extends into `producer → asset → testing_script` — preserving the two-hop `script → asset → script` invariant `buildLineageDownstreamMap` relies on. `computeInducedSchedule` (bounded + full-pipeline runs) then orders the producer first.
- **Frontend canvas** (`AssetGraph/AssetGraphCanvas.svelte`): renders the new edge as a dashed amber "test needs" link, visually distinct from lineage (blue/gray solid) and triggers (gray) — the tested script does not consume the asset's rows.
- **Types** (`AssetGraph/types.ts`) + **openapi.yaml**: documented `AssetGraphTestEdge` / `test_edges`.
- **Docs** (`docs/ducklake-materialization.md`): new "Cascade ordering (`test_edges`)" subsection documenting the behavior and its single-cascade scope caveat.

## Design notes / decisions

- **Ordering-only, not lineage.** Modeled as its own edge family (dashed, like macro edges), not a lineage read, so it never implies the tested script ingests the asset's data.
- **Scope.** Orders *within a single client-driven cascade* (dev run / bounded run / full-pipeline run). It does not change the production reactive asset-dispatch of two independently-scheduled roots — an explicit `// on <dim>` is still the way to force a hard cross-root ordering there. This matches the work package's stated scope (boundedCascade / graphTraversal orderer).
- **Custom tests** are best-effort: the custom-test script's parsed reads drive the producers; reading the member's own output resolves to the member and is dropped by the self-edge guard.

## Screenshots

Two disjoint roots seeded (`f/shop/dim_customers` producing `main/dim_customers`; `f/shop/fct_orders_daily` carrying `// data_test relationships customer_id -> ducklake://main/dim_customers.customer_id`). The amber dashed edge is the new ordering edge:

pipeline-graph

## Test plan

- [x] Frontend unit tests (`boundedCascade.test.ts`): through-asset DAG routing; disjoint-root HD-1 repro asserts the producer is ordered strictly before the tested script via `computeInducedSchedule`; no-producer/external-table case adds no ordering. Full AssetGraph suite: 191 passing.
- [x] `npm run check`: 0 errors.
- [x] E2e (backend): seeded the two-disjoint-root scenario; `GET /assets/graph?folder=shop` returns `test_edges: [{producer: dim_customers, tested: fct_orders_daily, asset: main/dim_customers}]`. Pointing the ref at a producerless external table returns no `test_edges` (edge case 3).
- [x] E2e (frontend): pipeline graph renders the dashed amber "test needs" edge (`stroke rgb(217,119,6)`, dash `5 3`) from producer to tested script.

---
## Summary by cubic
Ensure `// data_test relationships` and custom tests run after the referenced asset’s producer in cascades. Addresses HD-1 (Hidden Dependencies) by adding scheduling-only edges consumed by the scheduler and shown in the graph.

- **Bug Fixes**
  - Backend: emit `test_edges` linking `producer → tested_script` when a test references a pipeline asset; skip self-edges and no-producer (external) cases.
  - Backend (custom tests): key read lookups by `(usage_kind, path)` and resolve only `Script` reads to avoid collisions with same-path flows.
  - Cascade scheduler: route `test_edges` through the referenced asset (`producer → asset → tested_script`) to preserve the `script → asset → script` invariant for bounded/full runs.
  - UI: render dashed amber “test needs” edges to distinguish ordering from lineage/trigger links.
  - API/Types/Docs/Tests: added `AssetGraphTestEdge` to `openapi.yaml` and frontend types, documented behavior in `ducklake-materialization.md`, and added unit tests for ordering and external-table no-op.

<sup>Written for commit 2490fdb39385c59a9aa6bcfe99a372dbccb83fe9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-assets/src/lib.rs
+++ b/backend/windmill-api-assets/src/lib.rs
@@ -1298,14 +1298,19 @@ async fn asset_graph(
                 .push((e.runnable_kind, e.runnable_path.clone()));
         }
     }
-    // Assets a custom-test script reads, keyed by its path — used to order that
-    // script's producers before whichever member declares `// data_test <path>`.
-    let mut reads_by_runnable: std::collections::HashMap<String, Vec<(AssetKind, String)>> =
-        Default::default();
+    // Assets a runnable reads, keyed by (usage_kind, path) — used to order a
+    // custom-test *script*'s producers before whichever member declares
+    // `// data_test <path>`. Keyed on the kind too because a flow can share a
+    // script's path; `// data_test <path>` resolves a deployed script body, so
+    // the lookup below pins `Script` and never pulls a same-path flow's reads.
+    let mut reads_by_runnable: std::collections::HashMap<
+        (AssetUsageKind, String),
+        Vec<(AssetKind, String)>,
+    > = Default::default();
     for e in &edges {
         if matches!(e.access_type.as_deref(), None | Some("r") | Some("rw")) {
             reads_by_runnable
-                .entry(e.runnable_path.clone())
+                .entry((e.runnable_kind, e.runnable_path.clone()))
                 .or_default()
                 .push((e.asset_kind, e.asset_path.clone()));
         }
@@ -1325,12 +1330,13 @@ async fn asset_graph(
                         to_path.clone(),
                     )]
                 }
-                // Best-effort: the custom test script's parsed reads. Reading the
-                // member's OWN output resolves to the member as producer and is
-                // dropped by the self-edge guard below.
-                DataTest::Custom { path } => {
-                    reads_by_runnable.get(path).cloned().unwrap_or_default()
-                }
+                // Best-effort: the custom test *script*'s parsed reads. Reading
+                // the member's OWN output resolves to the member as producer and
+                // is dropped by the self-edge guard below.
+                DataTest::Custom { path } => reads_by_runnable
+                    .get(&(AssetUsageKind::Script, path.clone()))
+                    .cloned()
+                    .unwrap_or_default(),
                 _ => vec![],
             };
             for (asset_kind, asset_path) in referenced {
--- a/backend/windmill-api-assets/src/lib.rs
+++ b/backend/windmill-api-assets/src/lib.rs
@@ -797,6 +797,25 @@ enum TriggerEdge {
     },
 }
 
+// Ordering-only "must-run-after" edge: `runnable_path`'s data test reads
+// `asset` (a `// data_test relationships` ref, or a custom test whose body
+// reads a known pipeline asset), so the asset's in-pipeline producer must
+// materialize before `runnable_path` runs. NOT a data-consumption edge — the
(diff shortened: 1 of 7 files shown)


rubenfiszel:
Addressed the Codex P2 in 2490fdb: `reads_by_runnable` is now keyed by `(AssetUsageKind, path)` and the `DataTest::Custom` lookup pins `(Script, path)`, so a same-path flow's reads can no longer leak into a custom-test's referenced-asset set (no spurious `test_edges`). Backend recompiled clean and the relationships-edge e2e still emits the expected `dim_customers → fct_orders_daily` edge.

Claude's and Pi's reviews were clean; Claude's non-blocking note about `rw` self-writes in custom-test reads is left as-is (harmless per its own assessment — it only schedules the custom-test script ahead of a member it must run for anyway).

PR #9942, opened 2026-07-05
fix(cli): macro-library parity in --local pipeline graph + read-only run --dry-run

rubenfiszel:
## Summary

Two independent CLI fixes in `cli/src/commands/pipeline/` for data pipelines, closing local-vs-deployed parity gaps.

**HD-3 — macro libraries were invisible to local tooling.** In the *deployed* graph (`wmill pipeline show <folder>`, the UI) a `// macros` DuckDB library is a first-class node ("defines N macros" + violet `uses <macro>` edges to callers). In the *local* surfaces — `pipeline show --local`, the `pipeline dev` graph, and `pipeline docs` — the library was **absent**: no node, no `uses` edge, and the script count excluded it. `localGraph.ts` already *modelled* `macros?`/`macro_edges?` but never populated them locally, because the pinned wasm asset parser drops the `// macros` / `// use` annotations and emits no macro registry. An agent handed the local docs couldn't discover the reuse layer and would re-inline logic that already exists as a macro.

**`run --local --dry-run` was not read-only.** It wrote `*.script.yaml` / `*.script.lock` / `wmill-lock.yaml` into the working tree as a side effect of preview import-resolution (`buildPreviewTempScriptRefs`), even though a plan preview runs nothing.

## Changes

- **`duckdbMacros.ts` (new):** a faithful, self-contained TS port of the lexical rules in `backend/parsers/windmill-parser/src/duckdb_macros.rs` (`split_statements`, `parse_create_macro`, `detect_macro_calls`) plus a leading-comment-header scan for the `// macros` / `// use` annotations (the wasm parser drops both). Deliberately lexical — over-matching a call only draws a spurious edge, never a wrong run.
- **`localGraph.ts`:** `buildMacroEdges` detects `// macros` libraries, resolves which pipeline scripts call their macros (word-boundary call detection + `// use` whole-library injection), and populates `macros` on the library node + `macro_edges`. Mirrors the deployed graph builder (`asset_graph` in `windmill-api-assets`): a library node surfaces only when it is an edge endpoint (unused libraries are not shown), `via_use` is serialized unconditionally, `macro_names` is sorted. Library nodes now count toward `runnables` (so `show --local` says "N scripts" matching deployed). `pipeline show`/`pipeline dev` needed **no** changes — their render code already consumes `macros`/`macro_edges`.
- **`docs.ts`:** adds a `## Macro libraries` section (each library's `name(params)` signatures, TABLE marker, and callers with call-vs-`// use` provenance); libraries are listed there instead of under `## Scripts` and shown in the count line. `generatePipelineMarkdown` exported for testing.
- **`pipeline.ts`:** gate `buildPreviewTempScriptRefs` on `!opts.dryRun` so `run --local --dry-run` computes the plan without generating/locking/uploading metadata. The refs are only consumed when a preview job actually runs, which dry-run never reaches.
- **Tests:** new `cli/test/duckdb_macros_unit.test.ts` (parser port: accepted shapes, rejections, comment/string stripping, call detection, annotation scan) and extensions to `cli/test/pipeline_local_graph_unit.test.ts` (library node + signatures + call edge + count, `// use` whole-library edge, unused-library suppression, and a golden `## Macro libraries` docs assertion).

## Verification

Verified **byte-for-byte parity** by deploying a `// macros` library + a call-site consumer + a `// use` consumer to a live CE backend and diffing the deployed `/assets/graph` against `pipeline show --local --json`: identical `macro_edges` (including per-consumer `macro_names` and `via_use` flags), identical `macros` node signatures, identical runnable count.

```
$ wmill pipeline show shop --local
Pipeline f/shop — 3 scripts · 2 assets

macros_finance [ƒ macros: net_revenue, safe_div]
└─ƒ fct_orders_daily (uses net_revenue)

seed_orders
└─▶ datatable://main/orders
    └─ fct_orders_daily
       └─▶ datatable://main/fct_orders_daily
```

Read-only dry-run (in a git working tree):
```
$ wmill pipeline run shop --local --dry-run
Bounded run plan — 2 scripts (from seed)
  1. f/shop/seed
  2. f/shop/fct
$ git status --porcelain     # empty → CLEAN (previously wrote 7 metadata files)
```

No CLI flags/commands/descriptions changed, so `system_prompts` needs no regen. `cli/` typecheck adds no new errors (16 pre-existing, stale generated client); all 924 CLI unit tests pass.

## Test plan
- [ ] `wmill pipeline show <folder> --local --json` matches the deployed `wmill pipeline show <folder> --json` for `macros`/`macro_edges` on a folder with a `-- macros` library + a caller (call-detected) + a `-- use` consumer.
- [ ] `wmill pipeline docs <folder> --local` emits a `## Macro libraries` section listing signatures and callers; the library is counted in the header line.
- [ ] `wmill pipeline dev <folder>` renders the macro-library node + `uses` edges in the graph editor.
- [ ] `wmill pipeline run <folder> --local --dry-run` prints the run plan and leaves `git status` clean (no `*.script.yaml` / `*.script.lock` / `wmill-lock.yaml` written); a non-dry `run --local` still resolves local relative imports.

---
## Summary by cubic
Shows DuckDB `// macros` libraries in local pipeline graphs as pipeline-member nodes with deployed-parity edges (shared libs and lib→lib), and makes `pipeline run --local --dry-run` read-only. Macro libraries are definition-only: visible in graphs/docs but excluded from runs and manual starts. Addresses HD‑3 by aligning `--local` graphs with the deployed view.

- **Bug Fixes**
  - Membership parity: in-folder `// macros` libraries are member nodes with signatures; out-of-folder providers appear as non-member nodes. Libraries stay unrunnable and are filtered from starts/selection; `--from` targeting them is rejected with a clear error.
  - Workspace-wide discovery: build `macro_edges` from lexical calls and `// use`, including lib→lib and dynamic-SQL cases (`via_use` flagged). Both edge endpoints are surfaced as nodes.
  - Exclude display-only macro consumers (non-`// pipeline` DuckDB helpers) from starts/selection; reject `--from` pointing to them.
  - `pipeline run --local --dry-run` skips preview ref generation to avoid writing `*.script.yaml` / `*.script.lock` / `wmill-lock.yaml`.
  - Annotation scan parity: accept `//`, `--`, and `#` prefixes.

- **New Features**
  - Docs add a “Macro libraries” section listing `name(params)` (TABLE flagged) and callers, including `via // use`; libraries are excluded from “Scripts” but counted separately in the header. Exported `generatePipelineMarkdown`.

<sup>Written for commit 6ab8e95d8ada80fe9e3c453c3bd108958e79ea5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the Codex P1 (out-of-folder / shared macro libraries) in 4d99e6a.

**Fix:** local macro-library discovery is now workspace-wide, mirroring the deployed builder (`asset_graph` fetches the macro registry unfiltered; only consumers are folder-scoped). Added `collectMacroLibraries(root)` which walks the whole `f/` tree for `*.duckdb.sql` files carrying a `// macros` marker (duckdb-only, so the extra walk stays cheap), and `buildMacroEdges` now takes that workspace-wide map while keeping consumers scoped to the pipeline folder. A shared library like `f/shared/stats` used by a pipeline in `f/mypipe` now produces the edge, the provider node, and the docs entry.

**Verified byte-for-byte** against a live backend: for `folder=mypipe` with `f/shared/stats` (a `// macros` lib outside the folder) consumed by both a lexical caller and a `// use` consumer, the deployed `/assets/graph` and `pipeline show --local --json` produce identical `macro_edges` (incl. `via_use`), `macros` node signatures, and runnable count.

Regression test added: `a shared macro library OUTSIDE the pipeline folder is resolved (workspace-wide)` in `cli/test/pipeline_local_graph_unit.test.ts`. Full CLI unit suite green (925 tests).

Pi review was "Good to merge"; the dry-run read-only fix is unchanged.

rubenfiszel:
Addressed both Codex P1s from the follow-up review in 99f0d14.

**1. Macro-library consumers were skipped** (`localGraph.ts`). The deploy path records `macro_usage` for *every* DuckDB script — confirmed at `windmill-api-scripts/src/scripts.rs:1662-1688`, which runs `detect_macro_calls` against other libs' macros (`provider_path != ns.path`) for any DuckDB script, comment: *"A library's edges point at other libs' macros."* My consumer loop restricted to `// pipeline` members, so a `base → derived` lib→lib edge was missed and `base` could vanish. Fix: any folder DuckDB script is now a candidate consumer for lexical-call edges; both edge endpoints are forced into the node set. I kept the deploy asymmetry — `// use` edges are re-parsed only from pipeline members (`asset_graph`'s `annotations_by_path`), so those stay pipeline-scoped.

**2. `//` / `#` annotation prefixes were ignored** (`duckdbMacros.ts`). The canonical `parse_pipeline_annotations` (`asset_parser.rs:791-803`) strips `//`, `--`, or `#` regardless of language, so a `.duckdb.sql` library headed with `// macros` deploys but was missed locally (I only accepted `--`). `parseMacroAnnotations` now mirrors the backend and accepts all three prefixes.

**Verified byte-for-byte** against a live backend: a `base` (headed `// macros`) → `derived` (a `// macros` lib consuming `base_add`) → `report` (pipeline consumer) chain now produces `diff`-identical `macro_edges`, `macros` node signatures, and runnable count between deployed `/assets/graph` and `pipeline show --local --json`.

Regression tests added (`macro library that CONSUMES another library…`, `a // (slash-prefix) DuckDB library is detected`, plus a prefix-parity unit test). Full CLI unit suite green (928 tests). The workspace-wide discovery fix and dry-run read-only fix are unchanged.

rubenfiszel:
Addressed the Codex P1 (local run selecting non-pipeline macro consumers) in 87604ad.

The previous fix surfaces macro-consumer nodes for lineage display — including a DuckDB script that calls a macro but isn't `// pipeline`. Those nodes have no local file, so `pipeline run --local` must not treat them as manual roots (a `--dry-run` listed them; a real run failed resolving local content). This is purely a local-graph concern — my PR doesn't touch the deployed graph, which already carried such consumer nodes.

**Fix** (`pipeline.ts`): the run path now excludes, in addition to `// macros` libraries, any `--local` graph node absent from `localScripts` (the previewable set of `// pipeline` scripts) — from both the valid-starts set and the final selection.

**Verified e2e**: a folder with `lib` (`// macros`), `helper` (non-`// pipeline`, calls the macro), and `root` (`// pipeline`) → `pipeline run mypipe --local --dry-run` now prints a 1-script plan (`f/mypipe/root` only), `helper`/`lib` excluded, and `git status` stays clean. The helper still appears as a lineage node in `pipeline show --local` (parity), just not as a run step.

Regression test added (`a non-pipeline DuckDB macro consumer is a display-only node, never a run step`) asserting the helper is a graph node with `in_pipeline` falsy but absent from the previewable `scripts` set. Full CLI unit suite green (929 tests).

rubenfiszel:
Addressed the Codex P1 (explicit `--from` accepting display-only macro consumers) in 32d70bd, after merging current `main`.

Context: `main` landed the mid-DAG selective-execution feature (#9945), which resolves `--from` against `validFromStarts` / `fromEligible` — and that set was filtered only by `macroLibPaths`. So a non-`// pipeline` macro-consumer helper (a `--local` lineage-display node) passed `--from` eligibility and, after the selection cleanup dropped it, yielded an empty successful plan.

**Fix** (`pipeline.ts`):
- `fromEligible` is now filtered by the broader `notRunnablePaths` (macro libraries + `--local` nodes with no local file), matching the `starts` set.
- Added an explicit, clear rejection for a `--from` that names such a display-only node, instead of an empty plan.

**Verified e2e** (folder with `lib` `// macros`, `helper` non-pipeline calling the macro, `root` `// pipeline`):
- `run mypipe --local --dry-run --from helper` → **rejects clearly**: *"--from 'helper' isn't a `// pipeline` script — it appears in the local graph only as a macro consumer (lineage). Mark it `// pipeline` to run it."*
- `--from root` → correct 1-script plan; `git status` clean.

Full CLI unit suite green (932 tests after the merge). The `run()` entrypoint is backend-coupled (requires login/workspace) so the `--from` path is covered by manual e2e; the graph contract it depends on (helper is a node but absent from the previewable `scripts` set) has a unit test.

rubenfiszel:
Addressed both Codex P2 nits in d294e948.

1. **Literal NUL edge-key separator** — the macro edge map packed `(lib, consumer)` into a single string key with a `\0` separator, which made `localGraph.ts` read as a binary file to `grep`/`rg`. Replaced with a nested `lib_path → consumer_path → aggregate` Map, so there's no separator to collide (or to break tooling) at all.
2. **Stale "deployed graph only" comments** — updated the macro-renderer comment in `pipeline.ts` and the `macros`/`macro_edges` type comments in `localGraph.ts` to describe the current behavior (both surfaces now carry macro nodes/edges; the local ones are derived by `buildMacroEdges`), per the AGENTS.md comment policy.

No behavior change — the nested map produces identical `macro_edges` (all 28 macro/graph tests + full 932-test CLI suite green; byte-for-byte parity previously verified stands).

Both AI reviewers are now green (Pi "Good to merge"; Codex "Mergeable").

rubenfiszel:
Addressed the Codex P1 (unused `// pipeline` + `// macros` libraries runnable locally) in 84d6d19.

A DuckDB file can be both `// pipeline` and `// macros`. The deployed builder tags any node whose path provides macros — edge or not (`windmill-api-assets/src/lib.rs:1406`, `macros_by_provider.get(&path)`) — so such a script is recognized as definition-only and excluded from runs even with no consumers yet. My local enrichment only tagged edge providers, so an unused pipeline macro library stayed a bare runnable and `pipeline run --local` scheduled it as a manual root.

**Fix** (`localGraph.ts`): also tag with macros any library whose path is already a runnable (i.e. a `// pipeline` + `// macros` script), not just edge providers. Unused NON-pipeline libraries stay suppressed (no node), matching deployed.

**Verified e2e** (folder with an unused `-- pipeline -- macros` lib + a `-- pipeline` root):
- `pipeline show mypipe --local --json` → the lib node carries `macros: ["dbl"]` despite no consumers.
- `pipeline run mypipe --local --dry-run` → plan is `root` only; the lib is excluded (was scheduled before the fix). `git status` clean.

Regression test added (`an UNUSED // pipeline + // macros library is still tagged with its macros`). The existing "unused non-pipeline library is not surfaced" test still passes, confirming the suppression path is intact. Full CLI suite green (933 tests).

rubenfiszel:
Implemented **parser precedence** (`// macros` wins over `// pipeline`) so a macro library is never a pipeline member — removing the confusing state at its root rather than special-casing it. Commits b398b69 (behavior) + 34672a5 (the P2 comment trim).

**Why** — a `// macros` library is definition-only (its macros are injected into consumers; running it is a no-op). Marking it `// pipeline` did nothing useful and only produced the round-6 bug (an unused pipeline macro library surfacing as a runnable manual root).

**Change** — `parse_pipeline_annotations` forces `in_pipeline = false` when `macros` is set. Mirrored in the three parsers that must agree:
- **Rust canonical** (`asset_parser.rs`) — drives deploy membership (`auto_kind='pipeline'`) and the deployed graph.
- **Frontend TS** (`parsePipelineAnnotations.ts`) — drives the live graph preview.
- **CLI local graph** (`localGraph.ts`) — the pinned wasm still reports `in_pipeline`, so precedence is applied when deciding membership; the round-6 "tag unused pipeline+macros lib" branch is now removed (dead — such libs are no longer members, so an unused one is suppressed exactly like a plain `// macros` one).

**Tests** — shared TS↔Rust parity fixture (`macros wins over pipeline`) + a pre-existing fixture updated to the new expectation; unit tests on each side. Rust `windmill-parser` 159 + parity 1 green; frontend parser 75 green; CLI 934 green.

**Also addressed the P2 comment nit** — trimmed the ~19-line narrative header in `duckdbMacros.ts` (parity rationale belongs in the PR description) to the two real constraints, and tightened the duplicated rationale in `localGraph.ts`.

Note on the frontend file: the change is a pure parser-parity mirror. Its only "visible" effect is that an unused `// pipeline` + `// macros` node no longer renders as a member in the live editor — i.e. the exact nonsensical state being removed — so there's no meaningful UI to screenshot.

Note on rollout: the runtime behavior needs the backend rebuilt with the parser change; the TS↔Rust parity fixture is what guarantees the CLI/frontend/backend stay in agreement, so I verified at the parser level rather than against the old deployed binary.

rubenfiszel:
Reversed course on the precedence approach after verifying the actual backend behavior — commit 6ab8e95.

**What I found:** the backend deliberately marks *every* macro library `auto_kind='pipeline'` (`windmill-api-scripts/src/scripts.rs:1474`, via `macro_lib_defs.is_some()`, comment: *"A macro library is a pipeline member (graph node) even without the bare // pipeline marker"*). I confirmed against the running backend: an **unused** `// macros` library is a member node with `in_pipeline: true`. So a macro library **is** a pipeline member — the `// pipeline` marker is redundant, not something to override. My parser-precedence change was a no-op on deploy (auto_kind stayed `'pipeline'`) while diverging the CLI/frontend, so I reverted it (backend + frontend back to their prior state).

**The correct fix (CLI only):** model macro libraries as members, mirroring the deployed graph:
- An in-folder `// macros` library is a member node (`in_pipeline: true` + signatures), used or not.
- Its `// use` is now processed (it's a member) — so a library reaching another only through dynamic SQL (`query('…')`) gets the `via_use` lib→lib edge. **This fixes the missing-edge P1** from the previous review.
- An out-of-folder library used by an in-folder consumer is a non-member provider node.
- Libraries stay excluded from runs (via `macros`) and from the previewable `scripts` set.

**Verified byte-for-byte against the deployed graph — this time including `in_pipeline`** (my earlier compare script omitted it, which is how the divergence hid): unused in-folder lib, lexical lib→lib chain, `// use` dynamic-SQL lib→lib (Codex's exact scenario), and out-of-folder shared lib all `diff`-identical.

Net effect: the earlier "unused pipeline+macros lib is runnable" and "in_pipeline mismatch" divergences are gone, and the combo now behaves identically to a plain `// macros` library (the `// pipeline` marker is simply redundant on a library). New regression tests cover the unused-member, lib→lib-lexical, and lib→lib-`// use` cases. Rust/frontend parser tests untouched (reverted); CLI suite 934 green.

PR #9950, opened 2026-07-06
feat(pipelines): wm_partition macro for grain-agnostic partition filters

rubenfiszel:
## Summary

Partitioned `// materialize` nodes substitute `{partition}` with the partition's **identity string** (`2026-07-05T23`, `2026-W27`, `2026-07`, `2026-07-05`), not a timestamp. Only the daily identity parses as a DuckDB `DATE`/`TIMESTAMP`, so the intuitive `WHERE ts = TIMESTAMP {partition}` filter raises a `Conversion Error` for hourly/weekly/monthly — a documented footgun.

Rather than make every author hand-write a `strftime` format that must stay in lockstep with the resolver, the materialize runtime now **injects a `wm_partition(ts)` temp macro** as the first setup statement of a time-partitioned script. Filtering the source to the active slice becomes one grain-agnostic line:

```sql
WHERE wm_partition(<ts_col>) = {partition}
```

The macro's format comes from `PartitionKind::default_time_format` in `windmill-parser` — the **same source** the EE resolver reads to stamp the `{partition}` identity — so the macro and the token can never drift. `dynamic` partitions get no macro (their identity is a caller-supplied key → `WHERE <key_col> = {partition}`).

This supersedes the initial frontend-only approach (a 9-line `strftime` comment block in the scaffold): the format now lives in exactly one place and the scaffold hint collapses to a single line.

## Changes

**Backend (single source of truth)**
- `windmill-parser/asset_parser.rs` — `PartitionKind::default_time_format()` + `PartitionSpec::time_strftime_format()` (respects `format=` override; `None` for `dynamic`).
- `windmill-parser/sql_materialize.rs` — `wm_partition_macro(spec)` emits the `CREATE OR REPLACE TEMP MACRO …` statement + unit tests.
- `windmill-worker/duckdb_executor.rs` — injects the macro at the front of a time-partitioned materialize's setup (post-substitution; runs before the wrapped SELECT and the write txn).
- **EE** `windmill-common/partition_ee.rs` — `default_format` delegates to `default_time_format` (no behavior change). *(companion PR below; `ee-repo-ref.txt` bumped)*

**Frontend**
- `pipelineTemplates.ts` — scaffold emits the single `WHERE wm_partition(<ts_col>) = {partition}` hint (was a 9-line strftime block).
- `copilot/chat/pipeline/core.ts` — AI prompt teaches `wm_partition`.
- `pipelineTemplates.test.ts` — asserts the macro idiom, and that neither `TIMESTAMP {partition}` nor a raw `strftime` format is scaffolded.
- `docs/ducklake-materialization.md` — documents the macro + the single-source design.

## Test plan
- [x] `cargo test -p windmill-parser wm_partition` — macro emits the resolver's exact format per grain; `format=` override honored; `dynamic` → no macro
- [x] Backend type-checks with `quickjs,duckdb,parquet,private,enterprise,license` (running `cargo watch` rebuilt `windmill-parser` + `windmill-common` + `windmill-worker` clean)
- [x] `vitest` (8/8) + `svelte-check` (0 errors)
- [x] **Cross-engine parity verified through the real bundled DuckDB** (`windmill-duckdb-ffi-internal` in-memory test): `strftime` renders every grain format byte-for-byte identically to chrono across ISO-week year boundaries (`2027-01-01` → `2026-W53`), the injected `wm_partition` macro buckets the whole hour slice, and the naive `TIMESTAMP '<weekly/monthly identity>'` cast Conversion-Errors

## Screenshots

**Generated scaffold** for a DuckDB → *Materialized table* node — the frontend-visible change. Rendered from the app's own `generatePipelineDraft` module over the running pipeline editor (the node-creation UI feeds this exact text into the editor); the single `WHERE wm_partition(<ts_col>) = {partition}` line replaces the former 9-line strftime block:

generated scaffold

**Node-creation picker** (DuckDB + Materialized table) in the pipeline editor that produces it:

node picker


rubenfiszel:
### Addressed the review nits

Both P2 nits from the earlier reviews are resolved, and the design moved from the frontend-only `strftime` idiom to a runtime-injected **`wm_partition(ts)` macro** (single-sourced format, so it can't drift from `{partition}`):

**1. Weekly `%G-W%V` runtime verification (the one CI couldn't reach).** Added `wm_partition_parity_tests` in `windmill-duckdb-ffi-internal`, run against the **real bundled DuckDB** engine in-memory:
- `strftime` renders **byte-for-byte identically to chrono** (the resolver's engine) for daily/hourly/**weekly**/monthly across ISO-week year boundaries — including `2027-01-01 → 2026-W53` (2026 is a 53-week ISO year), `2023-01-01 → 2022-W52`, `2021-01-03 → 2020-W53`. No off-by-one at year boundaries.
- The injected macro buckets the **whole hour** (both `23:10` and `23:55`), and `TIMESTAMP '2026-W27'` / `'2026-07'` **do** Conversion-Error.

  `cargo test` in that crate: 13/13 pass.

**2. Timezone caveat.** Now documented at every point of use — the `wm_partition_macro` doc comment, `docs/ducklake-materialization.md`, and the copilot prompt all state that a non-UTC `tz=` requires `<ts_col>` to be expressed in that zone (same caveat the raw idiom carried, now explicit).

CI `cargo_test` / `check_ee` / `check_oss` / `npm_check` are green. The only red check is **CLA** — triggered by the `Co-Authored-By: Claude` trailer; @rubenfiszel to sign or I can drop the trailer.

rubenfiszel:
### Screenshot added (P2)

Attached the required frontend screenshots to the PR body — the generated DuckDB *Materialized table* scaffold (showing the single `WHERE wm_partition(<ts_col>) = {partition}` line that replaced the 9-line strftime block) rendered over the running pipeline editor via the app's own `generatePipelineDraft` module, plus the node-creation picker that produces it. The pipeline node-create → Monaco flow itself resisted reliable automation (melt-ui popover + svelte-flow draft handling), so the scaffold is rendered from the exact production module rather than captured mid-Monaco; the text is identical to what the editor shows and is pinned by `pipelineTemplates.test.ts`.

rubenfiszel:
### Fixed: CREATE TEMP MACRO now classified as prepare-path setup

Good catch — confirmed and fixed in `backend/windmill-duckdb-ffi-internal/src/lib.rs`.

`prepare_duckdb_internal` only **executes** `is_setup_statement` blocks; everything else is merely `prepare`d. `CREATE [OR REPLACE] TEMP MACRO` wasn't recognized, so on an `annotations.prepare` run the injected `wm_partition` macro was never created on the connection and the later `CREATE TABLE … SELECT … WHERE wm_partition(...)` failed to bind. (The same latent gap hit the workspace-macro splicer at `duckdb_executor.rs:1591`, which injects TEMP MACRO blocks too — now also covered.)

`is_setup_statement` now matches `CREATE [OR REPLACE] TEMP|TEMPORARY MACRO` (whitespace/keyword-normalized), so it's executed before dependent blocks and excluded from the `PrepareQueryResult` count; persistent `CREATE MACRO` stays a user statement.

Two tests added (15/15 pass in the FFI crate):
- `temp_macro_is_classified_as_setup` — TEMP/TEMPORARY variants true, persistent `CREATE MACRO` false.
- `prepare_pass_binds_consumer_only_when_macro_runs_as_setup` — mirrors the prepare loop: the `wm_partition` consumer binds only when the macro ran as setup, and fails to prepare otherwise (guard proving the classification is what matters).

PR #9959, opened 2026-07-06
feat: add cosmetic dev/staging label for dev workspaces

hugocasa:
## Summary

Dev workspaces (a standing fork paired with a "prod" root) always showed a hard-coded indigo `dev` badge and "Dev workspace of X" wording. Some teams treat that paired environment as **staging**, not dev. This adds a **purely cosmetic** per-dev-workspace display label with two values (`dev` | `staging`) that only changes the badge text (`dev` → `stg`) and identity wording ("Dev workspace of X" → "Staging workspace of X").

No behavior changes: it's the same paired-fork machinery (locking, promote, compare all still key off `is_dev_workspace` / `parent_workspace_id`). The label is stored per-workspace and surfaced in the two payloads every display reads from (`user_workspaces`, `get_dev_workspace`), so all badge/wording sites resolve it uniformly.

The badge stays indigo (only the text differs). `dev` is the default; `staging` is opt-in via a subtle link that previews the actual badge and switches to the other value ("Cosmetic label: `[stg]` Change to dev"), available at create, attach, and anytime from the Dev workspace settings tab.

## Changes

**Backend**
- Migration `add_dev_workspace_label`: nullable `workspace.dev_workspace_label` (NULL renders as `dev`).
- Threaded the field through `CreateWorkspaceFork`, `AttachDevWorkspace`, `UserWorkspace`, and `DevWorkspaceInfo` (so a prod admin who isn't a member of the dev still sees its label).
- `normalize_dev_workspace_label` validates to `{dev, staging}` and rejects anything else; validated early in both fork phases.
- New `POST /w/:w/workspaces/set_dev_workspace_label` (admin of the dev workspace; guarded so the target must be a dev workspace; audited).
- Rename path carries the label over so renaming a labelled workspace keeps it.

**Frontend**
- New `devWorkspaceLabel.ts` helper (`devBadgeText`, `devLabelWord`, `devLabelNoun`) as the single display-mapping source.
- Label-aware badge sites (workspace list, sidebar switcher, session family picker, session bar, session diff drawer header) and identity-wording sites (icon tooltip, sidebar sublabel, fork banner, `NoDirectDeployAlert`, family-picker locked tooltip).
- Create/fork form: a subtle "Cosmetic label: `[badge]` Change to Y" link (previews the real badge); the **fork name and id** auto-default to `<root>-dev` / `<root>-stg` when empty, so the form is submittable in one click.
- `DevWorkspaceSetting`: the same link to change the current dev workspace's label (persisted immediately), plus one in the attach form.

## Screenshots

**Create-fork form** — subtle label link previewing the `dev` badge; fork name and id auto-defaulted to `acme-dev`:

stg-create-fork-link

**Settings label editor** — flip the cosmetic label anytime (previews the `stg` badge):

stg-settings-link

**Workspace list** — indigo `stg` badge:

stg-workspace-list

**Home fork banner** — "Staging workspace of Acme":

stg-banner-home

## Test plan

- [ ] Fork a root workspace with the Dev-workspace toggle on; confirm the fork name and id auto-fill `<root>-dev`, the "Change to staging" link flips both to `<root>-stg`, and a typed name isn't clobbered (id follows it).
- [ ] After creation, confirm the workspace list / sidebar / session bar show an indigo `stg` badge and the fork banner + sidebar sublabel read "Staging workspace of X".
- [ ] In the workspace's Settings → Dev workspace tab, use the "Change to ..." link to flip the label; confirm all badges/wording update after the store refresh and survive a reload (persisted).
- [ ] On a locked prod, confirm `NoDirectDeployAlert` reads "staging" when the dev is labelled staging.
- [ ] Regression: an existing dev workspace with a NULL column still renders `dev` everywhere.


hugocasa:
Thanks for the review. Both P2 nits are intentional, keeping as-is:

- **Prefill (P2b):** working as designed. The `<root>-dev` / `<root>-stg` suggestion is gated on the name/id field being empty by spec ("default it when the field is empty"), so it deliberately never overwrites a value the user has already typed. Typing a name first and then enabling the toggle keeping that name is the intended behavior, not a regression.
- **Unit tests (P2a):** skipping for now. Both helpers are total mappings over a two-value set, and their sibling utility (`workspaceHierarchy.ts`) and the neighbouring workspace handlers (`create_workspace_fork` / `attach` / `detach`) carry no unit tests either, so adding them only here would be inconsistent with the surrounding code. The mapping is exercised via the manual walkthrough in the test plan.

PR #9975, opened 2026-07-07
raw-apps: prevent + surface the silent blank-screen (unmounted #root)

Guilhem-lm:
## Summary

AI-generated (or hand-written) React raw apps could render a **blank screen with no error** when `index.tsx` was written as a bare `export default function App() {...}` that never mounts anything. The preview harness bundles and executes `index.tsx` against an empty `<div id="root">` but does **not** auto-render a default export — the entrypoint must itself call `createRoot(document.getElementById('root')!).render(<App />)`. When it doesn't, the JSX lives inside an uncalled function, so nothing throws and the runtime-error overlay stays silent.

This PR tackles the failure mode from two sides:

1. **Prevent** it — strengthen the raw-app AI prompt so the mount requirement is explicit.
2. **Surface** it — when a build still renders nothing, show an info overlay in the editor that names the exact fix.

### Prompt (prevention)
- `system_prompts/base/raw-app.md` — rewrote the Entrypoint section: `index.tsx` is the bundling **and mount** entrypoint, MUST mount `App` into `#root` (React/Svelte/Vue snippets), and never be replaced with a bare component. Added a blank-screen debugging note.
- `frontend/src/lib/components/copilot/chat/app/core.ts` — same mount-entrypoint guidance in the in-chat app system prompt's Frontend section.
- Regenerated derived files via `system_prompts/generate.py`: `prompts.ts`, `skills/raw-app/SKILL.md`, `cli/src/guidance/skills.gen.ts`.

### Overlay (detection)
- `frontend/src/lib/components/raw_apps/RawAppEditor.svelte` — handle a new `emptyRender` message from the preview iframe and render an info `Alert` ("Nothing rendered") pointing at the missing `createRoot(...).render(...)` call. Mirrors the existing runtime-error overlay; cleared on next build.
- Also handle `renderAppeared` — the preview **retracts** the hint if a slow app (async fetch, Suspense) mounts `#root` after the grace window, so a legitimate late render can't leave a false-positive overlay.
- Both messages are emitted by the preview harness in the builder repo: **windmill-labs/windmill-code-ui-builder#17**. That change polls `#root` after each build and posts `emptyRender` when nothing mounted (suppressed if a runtime error already explained the blank screen), then watches for a late mount to post `renderAppeared`.
- The host handlers here are **dormant-safe** until the builder tarball that emits these messages is pinned, so they land ahead of the artifact bump.

## Screenshots

emptyrender-overlay

A bare-component `index.tsx` (`export default App`, no `createRoot`) builds successfully (see logs) yet mounts nothing — the "Nothing rendered" overlay names the fix.

## Sequencing

The overlay only fires once the builder tarball emits `emptyRender`:

1. Merge builder PR windmill-labs/windmill-code-ui-builder#17 → CI publishes a new tarball to R2.
2. Bump `frontend/scripts/ui_builder_artifact.json` (new version + sha256) — follow-up commit on this PR.

## Test plan

- [ ] Generate a React raw app via copilot chat; confirm `index.tsx` includes `createRoot(document.getElementById('root')!).render(<App />)` rather than a bare component.
- [ ] Confirm source↔generated consistency (grep the new wording in all three generated files).
- [x] Bare-component build shows the "Nothing rendered" overlay; overlay clears on a rebuild that mounts `#root`; runtime error suppresses it (browser-verified, see screenshot).
- [x] False-positive retraction (browser-verified against the running editor): late-mount build (`#root` filled at 1000ms) → overlay appears @612ms, **retracts @1018ms** via `renderAppeared`; permanently-empty build keeps the overlay; immediate render never shows it.

### DIFF
--- a/cli/src/guidance/skills.gen.ts
+++ b/cli/src/guidance/skills.gen.ts
@@ -5585,7 +5585,21 @@ A raw app has three logical parts:
 
 ### Entrypoint
 
-\`index.tsx\` is the bundling entrypoint. It typically renders a top-level \`App\` component. The bundler is esbuild.
+\`index.tsx\` is the bundling **and mount** entrypoint (esbuild bundles it; the preview then executes it against an empty \`<div id="root">\`). It MUST itself mount a top-level \`App\` component into \`#root\` — nothing is auto-rendered for you. Keep your UI in \`App.tsx\` (or \`App.svelte\` / \`App.vue\`) and keep \`index.tsx\` as the mount shim.
+
+React:
+
+\`\`\`tsx
+import React from 'react'
+import { createRoot } from 'react-dom/client'
+import App from './App'
+
+createRoot(document.getElementById('root')!).render(<App />)
+\`\`\`
+
+Svelte: \`mount(App, { target: document.getElementById('root')! })\`. Vue: \`createApp(App).mount('#root')\`.
+
+**Never turn \`index.tsx\` into a bare component** (e.g. \`export default function App() { ... }\` with no \`createRoot(...).render(...)\`). A component that is defined but never mounted renders **a blank screen with NO error thrown** — the JSX never executes, so nothing surfaces in the console or the error overlay. If the app is blank, the first thing to check is that \`index.tsx\` actually calls \`createRoot(document.getElementById('root')!).render(<App />)\`.
 
 **Always begin every React file (\`.tsx\`/\`.jsx\`) that uses JSX with \`import React from 'react'\`.** esbuild uses the classic JSX transform, so \`React\` must be in scope wherever JSX appears — a missing import compiles fine but throws \`React is not defined\` at runtime, leaving a blank screen.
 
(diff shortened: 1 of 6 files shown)


hugocasa:
Superseded by #10150 — closing this one.

That PR takes the same two-sided approach you established here (strengthen the raw-app prompts so `index.tsx` states the mount requirement, and surface a blank build in the editor), and keeps your `MutationObserver` retraction, which was the right call.

The one thing it changes is the **decision rule** in the harness. This PR reports on "is `#root` empty after the grace window?", which also describes an app that mounted fine but hasn't painted yet (a fetch in flight, an unresolved Suspense) — which is why it needed to report first and walk it back, flashing a wrong overlay for ~400ms on any slow-rendering app and never retracting for one that renders nothing on purpose.

#10150 additionally requires that the app **never looked `#root` up** (probing `document.getElementById` / `querySelector` and matching on the returned element, so React/Svelte and Vue's `.mount('#root')` are both covered). Requiring both signals means the slow-render case is never reported in the first place, so retraction is left covering only the rare deferred-mount. The overlay is also an error rather than info, and there's a guard for an app that mounts into a container of its own rather than `#root`.

Builder-side equivalent of windmill-labs/windmill-code-ui-builder#17 is windmill-labs/windmill-code-ui-builder#19 — worth closing that one too.

Credit for the shape of the fix and the diagnosis is yours — thanks for writing it up so clearly, it made the alternative easy to reason about.

PR #9816, opened 2026-06-26
feat(sessions): v2 unified sidebar with family/fork scoping and preview router

Guilhem-lm:
## Summary

Reworks the AI **sessions** experience and the global **logged-in sidebar** into a single, unified navigation surface. The sidebar separates *family-scoped* concerns (AI sessions) from *workspace/fork-scoped* concerns (favorites, search, workspace items), and the session preview gains a breadcrumb that doubles as a lightweight in-app router. This started as a design exploration and is now the implemented rework.

The new navigation is gated behind the existing AI-sessions dev flag (`localStorage wm_dev_global_ai=1` + a workspace AI provider), so the default product surface is unchanged. The shared UI primitives it touches (diff editor, popovers/dropdowns, tooltips) keep their existing defaults for all other consumers — see *Shared-component hardening* below.

## Changes

**Unified sidebar restructure** (`(root)/(logged)/+layout.svelte`, `sidebar/*`)
- Promote the workspace family picker to the top; the rail is now resizable with rem scaling.
- Split the sidebar into a family-scoped region (workspace family header, New AI session, session list) and a workspace-scoped region (a `WorkspaceScopeHeader` fork/root scope picker, Favorites, Search, workspace items), divided by a full-width separator.
- Family header keeps the *root* family color even when a fork is active; fork scope picker is accent-styled and truncates long fork names.
- Add "&lt;workspace&gt; settings" entries to both the fork dropdown and the bottom Settings dropdown; restore the logs/help/user/leave-workspace menus and the delete-forked-workspace action in the new Settings menu.

**Family/fork scoping** (`sessions/WorkspaceFamilyPicker.svelte`, `WorkspaceScopeHeader.svelte`, `WorkspaceMenu.svelte`)
- The main picker shows only workspace *families*, not forks; fork/root selection moves to the scope header.
- Switching to a different fork while viewing a session navigates home (sessions are fork-scoped).
- Create-fork is offered from the scope picker.

**Session preview breadcrumb → page router** (`sessions/+page.svelte`, `PreviewRouterPicker.svelte`, `PreviewRouterSegment.svelte`, `previewRouter.ts`)
- The preview header breadcrumb shows `family · fork  <page path>`.
- Every path segment opens a drill picker (a superset of the editor's `WorkspaceItemDrillPicker`) that lists workspace pages (Home, Runs, Workspace settings, …) alongside scripts/flows/apps.
- Picking a page or item steers the preview tab without leaving the sessions page; the override resets when switching sessions.
[...]

PR #9976, opened 2026-07-07
feat: open runs/schedules pages from AI chat in session preview tabs

Guilhem-lm:
## Summary

Lets the Windmill AI chat (global mode) take the user to a workspace **page** with the right filters applied — **Runs, Schedules, Variables, Resources, Assets, Audit logs, Folders, Groups, Triggers** (by kind), or **Workspace settings** (on a specific tab). Inside an AI session it opens as a **page tab in the side-panel preview** next to the chat (the primary UX); outside a session it offers a clickable link chip. Also lets the chat **close** preview tabs (`close_page`), and makes clicking a job inside a preview tab open the run detail in a **new** preview tab instead of navigating the current tab away.

> Rebased onto `main` — the session-preview infrastructure this depends on (`previewRouter`, `previewTabs`, `PreviewTabHost`) landed in `main` via #9816, so the base is now `main` and the diff is just the five `open_page`/`close_page` commits.

## Changes

- **`open_page` global-chat tool** (`copilot/chat/global/core.ts`): flat Zod schema (`page` enum + per-page filter fields — a top-level `oneOf` is rejected by Anthropic's `input_schema`, so not a discriminated union).
- **Permission-gated pages**: the advertised `page` enum is rebuilt **per-user** via the `setSchema` hook, mirroring the sidebar — `runs`/`assets` for everyone; `schedules`/`variables`/`resources`/`audit_logs`/`folders`/`groups`/`triggers` for non-operators; `workspace_settings` for admins/superadmins only. The `trigger_kind` enum is narrowed to license-available kinds (kafka/nats/sqs/gcp/azure require EE). A defense-in-depth guard in the handler refuses any page not in the user's allowed set.
- **URL builders** (`$lib/navigation.ts` `buildFilterUrl` + `global/pageNavigation.ts`): reuse the pages' own filter-key vocabularies and `serializeParam` (now exported from `svelte5UtilsKit.svelte.ts`) so a renamed query key can't desync. Schedules/Triggers item-open uses the `#<path>` hash; workspace settings uses `?tab=`. Trigger route metadata lives in `previewRouter.ts` (`TRIGGER_PAGES`), which also labels trigger preview tabs.
- **Session preview delivery**: `setOpenPagePreviewHandler` seam (in `sessionRuntime.svelte.ts`) opens the page as a `{type:'page'}` tab. A filter/tab change **reuses** the tab already showing that page (matched by `matchPreviewPage`, ignoring query/hash); a `new_tab` arg forces a separate tab.
- **Link-chip fallback** outside a session: new `navigate` variant on `ToolDisplayAction`, rendered by `ToolMessageActions.svelte`.
- **Job → new tab**: extended the session-preview `beforeNavigate` postMessage interceptor in `(logged)/+layout.svelte` — a `/run/<id>` client navigation inside a preview tab posts `wm.session.openRun` up, and `sessions/+page.svelte` opens it as a new preview tab (Runs tab stays put). `tabLabel` renders run tabs as `Run <8hex>`.
- **`close_page` global-chat tool** (session-only): closes preview tabs — `all: true` clears the panel, else `match` is a case-insensitive substring tested against each tab's page label and path. Dispatch uses a pure, unit-tested `selectPreviewTabsToClose`; tab labels come from a shared `previewLocationLabel` (also now backs the sessions tab strip's `tabLabel`).
- **Review fix**: dropped the outside-a-session "apply in place" branch (a same-route `goto` doesn't re-sync the pages' popstate-driven filter state) — now always offers a link chip when not in a session.
- **Tests/evals**: unit tests for `buildFilterUrl` + the page builders (`navigation.test.ts`, `pageNavigation.test.ts`); 7 `open_page`/`close_page` eval cases in `ai_evals/cases/global.yaml`.

## Blast radius

Frontend-only — **no backend, EE, migration, or API surface touched**. Only global chat gains new behavior; the shared surfaces this reaches are either additive or behavior-preserving:

- **All AI chat modes** (not just global) render tool-action cards through the shared `ToolMessageActions.svelte` + the `ToolDisplayAction` union, which gained a `NavigateAction` member. Existing `CreatedResourceAction` cards are byte-identical (subtitle still `= action.path`, button icon still `SquarePen`) — **browser-verified in the script-mode in-editor chat** (see below). Union exhaustiveness is enforced by `svelte-check`.
- **`(logged)/+layout.svelte`** mounts on every authenticated page: it registers the `navigate` action handler (inert unless a chip is clicked) and adds a `/run/<id>` interception **gated behind `isSessionPreviewEmbed()`** — normal run-page navigation everywhere else is untouched.
- **`$lib/navigation.ts` `buildFilterUrl`** is a new additive export; **`serializeParam`** is a visibility-only change (no behavior).

## AI evals (global mode)

Ran on `sonnet` against a local backend.

- **Feature cases: 7/7 pass** — 6 `open_page` (Runs filtered by script/schedule, open Schedule, `workspace_settings→git_sync`, `audit_logs→user`, `triggers→kafka`) + 1 `close_page` (`match: runs`). Each selects the right tool with the correct `page`/filter args (deterministic `toolExpect`, `skipJudge`).
- **Context-window A/B** (same 4 pre-existing docs cases, tools present vs. `core.ts` reverted to `main`): the two tool schemas add a flat **~+1,190 tokens** to the per-iteration window (~+4.4% of a ~27k window) — bounded, well clear of overflow/compaction. `open_page`'s page/filter/trigger enums are the bulk; `close_page` is tiny.

  | case | with tools | without | Δ |
  |---|---|---|---|
  | ai-agent-step | 28,007 | 26,810 | +1,197 |
  | retry-step | 27,969 | 26,788 | +1,181 |
  | key-value-store | 28,000 | 26,806 | +1,194 |
  | cron-schedule-format | 27,780 | 26,579 | +1,201 |

- **No regression**: the one diverging pre-existing case never invokes the new tools (it's `search_docs`-only); it's a tight-`maxTurns` flake on an open-ended docs question, reproducible independent of the change.

## Screenshots

Runs page opened with filters as a preview tab, next to the chat:

open-page-preview-tab

Changing the filter re-points the **same** tab (no duplicate):

tab-reuse-filter-change

`close_page` closes the Runs tab on request (Schedules stays open):

close-page

Shared-renderer regression check — an existing `CreatedResourceAction` card (a schedule created by the **script-mode in-editor chat**) still renders correctly through the refactored `ToolMessageActions` (icon, title, path subtitle, "Open" button):

in-editor-resource-card

## Test plan

- [x] `svelte-check` — 0 errors in changed files (1 unrelated pre-existing error in `utils_workspace_deploy.ts` from the data-tables feature on `main`); `buildFilterUrl` + page-builder + preview-tabs unit tests pass
- [x] Session chat: "open the failed runs of X" → Runs tab opens in preview with filters; "now show successful runs" → same tab re-points (no duplicate)
- [x] "open the schedule for X" → `/schedules#<path>` opens the edit drawer
- [x] Deep-link vocabulary (deterministic): `/workspace_settings?tab=git_sync` renders the Git sync tab; `/kafka_triggers` renders the Kafka triggers page
- [x] Job → new tab: `wm.session.openRun` opens the run in a new preview tab; tab label `Run <id>`
- [x] `close_page`: with Runs + Schedules tabs open, "close the runs tab" → AI calls `close_page`, Runs tab removed, Schedules stays (browser-verified)
- [x] `ai_evals` global mode: 7/7 feature cases pass; context-tax A/B measured (+~1,190 tokens, no regression)
- [x] Shared renderer: existing `CreatedResourceAction` card verified un-regressed in a non-global (script-mode in-editor) chat
- [ ] Operator permission gating: log in as an operator and confirm `open_page` doesn't offer the restricted pages (unit-reasoned + fail-closed; not browser-tested)
- [ ] Manual: click "Show run details" on a job inside a session Runs tab and confirm the run opens in a new preview tab

### DIFF
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -872,6 +872,103 @@
     - fetches the logs for the requested job id
     - explains the failure from the returned logs (connection refused to the upstream API)
 
+# --- Page navigation (open_page) ---
+# The assistant should take the user to a Windmill page (Runs/Schedules) with the
+# right filters via open_page, rather than describing where to click or dumping the
+# data. No draft is produced, so the global judge is skipped and we validate the
+# tool call and its arguments.
+
+- id: global-openpage1-runs-failed-of-script
+  prompt: |-
+    Take me to the failed runs of the script at f/evals/global/greet_user so I can see what's going wrong.
+  runtime:
+    maxTurns: 6
+  validate:
+    draftCountExactly: 0
+  toolExpect:
+    requiredToolsUsed:
+      - open_page
+    forbiddenToolsUsed:
+      - write_script
+      - deploy_workspace_item
+      - delete_workspace_item
+    toolCallArgs:
+      - tool: open_page
+        field: page
+        stringIncludesAnyOf:
+          - runs
+      - tool: open_page
+        field: status
+        stringIncludesAnyOf:
+          - failure
+      - tool: open_page
+        field: path
+        stringIncludesAnyOf:
+          - f/evals/global/greet_user
+  skipJudge: true
+  judgeChecklist:
+    - opens the Runs page filtered to the failed runs of f/evals/global/greet_user
+    - applies both the failure status and the script path as filters
+    - does not write, deploy, or delete anything
+
+- id: global-openpage2-runs-of-schedule
+  prompt: |-
+    Open the runs page filtered to the jobs triggered by the schedule f/evals/global/nightly_digest.
+  runtime:
+    maxTurns: 6
+  validate:
+    draftCountExactly: 0
+  toolExpect:
+    requiredToolsUsed:
+      - open_page
+    forbiddenToolsUsed:
+      - write_script
+      - deploy_workspace_item
+      - delete_workspace_item
+    toolCallArgs:
+      - tool: open_page
+        field: page
+        stringIncludesAnyOf:
+          - runs
+      - tool: open_page
+        field: schedule_path
+        stringIncludesAnyOf:
+          - f/evals/global/nightly_digest
+  skipJudge: true
+  judgeChecklist:
+    - opens the Runs page filtered to jobs triggered by the f/evals/global/nightly_digest schedule
+    - passes the schedule path as the filter
+    - does not write, deploy, or delete anything
+
+- id: global-openpage3-open-schedule
+  prompt: |-
+    Open the schedule f/evals/global/nightly_digest so I can review and edit it.
+  runtime:
+    maxTurns: 6
+  validate:
+    draftCountExactly: 0
+  toolExpect:
+    requiredToolsUsed:
+      - open_page
+    forbiddenToolsUsed:
+      - write_schedule
+      - deploy_workspace_item
+      - delete_workspace_item
+    toolCallArgs:
+      - tool: open_page
+        field: page
+        stringIncludesAnyOf:
+          - schedules
+      - tool: open_page
+        field: open
+        stringIncludesAnyOf:
+          - f/evals/global/nightly_digest
+  skipJudge: true
+  judgeChecklist:
+    - opens the Schedules page and targets the f/evals/global/nightly_digest schedule for editing
+    - passes the schedule path so the editor opens on it
+    - does not write, deploy, or delete anything
+
 # --- Documentation search (search_docs) ---
 # Pure product-knowledge questions: the assistant should consult the docs via
 # search_docs and answer conversationally, not draft or mutate anything. No
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -1023,6 +1023,33 @@
     - opens the Audit logs page filtered to the admin user
     - does not write, deploy, or delete anything
 
+- id: global-openpage6-triggers-kind
+  prompt: |-
+    Take me to the Kafka triggers for this workspace.
+  runtime:
+    maxTurns: 6
+  validate:
+    draftCountExactly: 0
+  toolExpect:
+    requiredToolsUsed:
+      - open_page
+    forbiddenToolsUsed:
+      - deploy_workspace_item
+      - delete_workspace_item
+    toolCallArgs:
+      - tool: open_page
+        field: page
+        stringIncludesAnyOf:
+          - triggers
+      - tool: open_page
+        field: trigger_kind
+        stringIncludesAnyOf:
+          - kafka
+  skipJudge: true
+  judgeChecklist:
+    - opens the Kafka triggers page
+    - does not write, deploy, or delete anything
+
 # --- Documentation search (search_docs) ---
 # Pure product-knowledge questions: the assistant should consult the docs via
 # search_docs and answer conversationally, not draft or mutate anything. No
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -1050,6 +1050,29 @@
     - opens the Kafka triggers page
     - does not write, deploy, or delete anything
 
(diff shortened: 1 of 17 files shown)


Guilhem-lm:
Addressed the P2 in `b370fb6658`: the handler now applies the same fail-closed check to `trigger_kind` that it already does for `page`. When `page === 'triggers'` and the requested kind isn't in `allowedTriggerKinds()` (i.e. an EE-only kind on a non-EE instance), it returns a clear message instead of building the route — so the handler now matches its "defense in depth" comment symmetrically.

- `check:fast`: clean (the sole error is the pre-existing unrelated `utils_workspace_deploy.ts` one from the data-tables feature on `main`).
- Re-ran the `global-openpage6-triggers-kind` eval → still passes (the guard changes only the tool's return, not tool selection).

The two non-blocking notes (multi-value `path` round-trip, trigger tab-reuse) I left as-is per your "no action needed". The two manual checks (operator gating, job→new-tab) remain the outstanding pre-merge items in the test plan.

Guilhem-lm:
Thanks — both Codex P2s addressed (still local commits, not yet pushed):

**#1 operator_settings gating** (`2c20bb70e7`): `allowedOpenPages()` no longer hardcodes operators to runs/assets. For an operator it now returns exactly the pages whose `operator_settings` flag is `true` (same source `OperatorMenu` gates on) — so an operator with `schedules`/`resources` enabled is offered them, and one with `runs`/`assets` disabled is not. Added an empty-enum guard in `buildOpenPageDefSchema` (a zero-page operator would otherwise crash `z.enum([])` in the per-iteration `setSchema`). Non-operators and admins are unchanged.

**#2 EE trigger kind runtime gate** (`b370fb6658`): the handler now fail-closes on a `trigger_kind` outside `allowedTriggerKinds()` (EE-only kind on a non-EE instance), mirroring the existing `page` guard — same nit claude[bot] flagged.

Validation: `check:fast` clean (only the pre-existing unrelated `utils_workspace_deploy.ts` error); `core.test.ts` 135/135 (a `setSchema`/prompt edit surfaced that my `open_page` prompt line leaked the literal `open_preview` into the non-preview prompt — moved that cross-reference into the preview-gated block, test green); the 7 `open_page`/`close_page` evals pass (the one `workspace_settings` miss was a re-confirmed model-variance hallucination of `page:"settings"`, not a gating change).

The operator-in-browser check remains the outstanding manual item in the test plan.

Guilhem-lm:
Fixed in `ff2b4dc842` (pushed).

The operator gate now reads `operator_settings` for the chat's **operating workspace** rather than `get(workspaceStore)`. Threaded the operating workspace through `GlobalToolHelpers.operatingWorkspace` so both sites gate on the same workspace:
- **setSchema** (advertised `page` enum) and the **handler guard** now call `allowedOpenPages(operatingWorkspaceFromHelpers(...))`.
- A **session** chat captures its own (possibly forked) workspace via `AIChatManager.operatingWorkspace`; the **global side-panel** chat leaves it unset so `allowedOpenPages` falls back to the live `$workspaceStore` (avoids a stale capture when the nav workspace changes mid-chat).

Note on residual scope: `operator_settings` are now correctly workspace-scoped, but the `operator`/`is_admin` *status* itself still comes from `userStore` (the navigation workspace's user) — `UserWorkspaceList` exposes per-workspace `operator_settings` but not a per-workspace operator flag, so a cross-workspace operator-status divergence isn't resolvable client-side without a new whoami call. Superadmin is instance-global and unaffected. The page still fail-closes server-side regardless, so this is not an auth bypass.

Validation: `check:fast` clean (only the pre-existing unrelated `utils_workspace_deploy.ts` error); `core.test.ts` 135/135; 7/7 `open_page`/`close_page` evals pass.

PR #9985, opened 2026-07-07
fix(ai): flow writer builds approval steps as scripts, not identity

Guilhem-lm:
## Summary

When asked to build a flow with an approval, the AI flow writer produced an approval step whose module value was `type: identity` — a no-op passthrough. An identity step suspends the flow but never calls `wmill.getResumeUrls()`, so no secret approve/cancel URLs are generated: approvers get no actionable link and the step is only resolvable from the run page. The canonical approval step (what the flow editor's "Approval" insert button creates) is a `rawscript` that generates and returns those resume URLs.

The root cause was in the shared flow-authoring prompt (`system_prompts/base/flow-base.md`, embedded into both the global chat and flow-editor chat via `getFlowPrompt()`): its "Correct shape" approval example used `type: identity` purely as filler to demonstrate `suspend` placement, and the model copied it literally.

## Changes

- Rewrote the **Approval / Suspend Structure** section in `system_prompts/base/flow-base.md`: the "Correct shape" is now a `rawscript` calling `wmill.getResumeUrls(approver)`, with an explicit "Do NOT use `type: identity`" rule and an identity example moved to the "Incorrect" set. Kept the existing lesson that `suspend` is module-level (sibling of `id`/`value`).
- Regenerated the downstream artifacts via `system_prompts/generate.py`: `auto-generated/flow.md`, `prompts.ts`, `skills/write-flow/SKILL.md`, and `cli/src/guidance/skills.gen.ts`.
- Hardened the `flow-test12-approval-step` eval case (`ai_evals/cases/flow.yaml`): added a deterministic `topLevelStepTypes` assertion requiring `request_approval` to be `rawscript` (an `identity` step now fails it), plus a judge-checklist item about generating actionable approval URLs. Previously the case only checked the `suspend` config, so the `identity` bug passed unnoticed.

## Eval A/B (`flow-test12-approval-step`, sonnet, 3 runs each, judge on)

| | Pass rate | `request_approval` step | Judge |
|---|---|---|---|
| Before (old prompt) | 0/3 | `identity` + suspend (no resume URLs) | 52 / 40 / 30 |
| After (this PR) | 3/3 | `rawscript` (bun) + `getResumeUrls` + suspend | 97 / 97 / 97 |

## Test plan

- [ ] `cd ai_evals && WMILL_AI_EVAL_BACKEND_URL=<backend> WMILL_AI_EVAL_BACKEND_WORKSPACE=integration-tests bun run cli -- run flow flow-test12-approval-step --model sonnet` passes
- [ ] Generated `request_approval` step is a `rawscript` calling `getResumeUrls` (not `identity`)
- [ ] `system_prompts/auto-generated/*` is in sync with `flow-base.md` (re-running `generate.py` produces no diff)

### DIFF
--- a/ai_evals/cases/flow.yaml
+++ b/ai_evals/cases/flow.yaml
@@ -359,6 +359,8 @@
       - request_approval
       - finalize_purchase
     topLevelStepTypes:
+      - id: request_approval
+        type: rawscript
       - id: finalize_purchase
         type: rawscript
     schemaRequiredPaths:
@@ -373,6 +375,7 @@
   judgeChecklist:
     - "the flow includes an approval step named `request_approval`"
     - "`request_approval` pauses the flow and asks the approver for a comment"
+    - "`request_approval` is a real script step that generates approval/resume URLs (e.g. via `getResumeUrls`) so approvers receive an actionable link, not a no-op passthrough (identity) step"
     - one approval is enough to continue
     - "the flow includes a final step named `finalize_purchase`"
     - "`finalize_purchase` returns an approved status object after approval"
--- a/ai_evals/cases/flow.yaml
+++ b/ai_evals/cases/flow.yaml
@@ -360,7 +360,7 @@
       - finalize_purchase
     topLevelStepTypes:
       - id: request_approval
-        type: rawscript
+        type: [rawscript, script]
       - id: finalize_purchase
         type: rawscript
     schemaRequiredPaths:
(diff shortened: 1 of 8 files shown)


Guilhem-lm:
Confirming the two manual-verification items from the automated review:

- ✅ `python system_prompts/generate.py` re-run locally on HEAD (`1dfba12`) produces **zero diff** against the committed artifacts (`auto-generated/flow.md`, `prompts.ts`, `skills/write-flow/SKILL.md`, `cli/src/guidance/skills.gen.ts`) — base and generated files are in sync.
- ✅ Eval A/B already recorded in the PR body: sonnet, 0/3 → 3/3, generated `request_approval` flips from `identity` to a `rawscript` calling `getResumeUrls`.

No code changes required from the review — verdict was good-to-merge.

PR #9989, opened 2026-07-07
fix(sessions): scope fork session Edits to session-edited items only

Guilhem-lm:
## Summary

Opening an AI session whose workspace is a **fork** could show the *entire* fork's draft list in the session **"Edits"** bar and diff drawer — even though that surface is labeled *"Edited by the chat during this session"*. Forks accumulate drafts from the fork operation and from other sessions, so this dumped many unrelated items under the session's edits.

Root cause: the Edits surface scopes drafts through a per-chat **modified-items mask** (`AIChatManager.modifiedItems`, a `SvelteSet` of `${kind}:${path}` keys). `buildDeployItems` treats `mask === undefined` as *"untracked → show every workspace draft"*. A session chat's mask was left `undefined` in two cases:

1. `loadPastChat` seeded `undefined` for a legacy chat with no persisted mask.
2. `initRuntime` took the `session.chatId` branch but `loadPastChat`'s `if (chat)` block was skipped when the chat wasn't in history yet (fresh `chatId`, no turn saved), so nothing initialized tracking.

Fresh sessions were fine (they call `initModifiedItemsTracking()`), so this bit legacy/reloaded sessions.

The fix makes **session chats always tracked** — they never fall back to show-all. Only the global side-panel chat (which doesn't render this bar) stays untracked.

Tradeoff: a legacy session with real prior edits now shows nothing rather than every unrelated fork draft — the correct behavior for a surface scoped to "edited during this session".

## Changes
- `AIChatManager.svelte.ts`: `loadPastChat` seeds `new SvelteSet(stored ?? [])` for session chats (was `undefined` when no mask was stored); updated the `modifiedItems` field doc to match.
- `sessionRuntime.svelte.ts`: `initRuntime` adds `if (manager.modifiedItems === undefined) manager.initModifiedItemsTracking()` after `loadPastChat`, covering the not-yet-persisted-chat case.
- `AIChatManager.test.ts`: 2 regression tests — a legacy session chat seeds an empty tracked set; a stored-mask session chat seeds from it.

## Verification
Reproduced empirically by seeding IndexedDB with two sessions in a fork (`wm-fork-skillful-fork`, 13 drafts): a chat **with** `modifiedItems` and one **without**.
- **Before:** the untracked session's Edits bar showed **"13 drafts"** (all fork drafts).
- **After:** the untracked session shows **nothing**; the tracked session still shows its **1 real edit**.

`AIChatManager.test.ts` + `sessionState.test.ts`: all pass (incl. 2 new tests). `check:fast` clean on the touched files (one pre-existing unrelated error in `utils_workspace_deploy.ts` from the data-tables migration).

## Screenshots
Untracked fork session after the fix — the Edits surface is scoped (only the genuinely-tracked session shows "1 draft"; the untracked one no longer dumps all 13 fork drafts):

after-fix-untracked-session-scoped

## Test plan
- [ ] In a forked workspace containing unrelated drafts, open a legacy AI session (chat with no persisted modified-items mask) and confirm its Edits bar / diff drawer show nothing until the session itself edits an item.
- [ ] Open a session whose `chatId` points at a chat with no saved turn yet and confirm it starts with an empty Edits surface.
- [ ] Confirm a session that does edit items still lists exactly those items (regression check on the happy path).

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -367,10 +367,11 @@ export class AIChatManager {
 
 	// Workspace items the CURRENT chat modified via AI tool calls, as
 	// `${UserDraftItemKind}:${storagePath}` keys (see modifiedItemsMask.ts).
-	// undefined = untracked: the global side-panel chat (never initialised) and
-	// loaded legacy chats with no stored mask, both of which fall back to the
-	// show-all bar. A SvelteSet (even empty) = tracked. Reactive so the session
-	// bar updates as tools record mid-turn.
+	// undefined = untracked: only the global side-panel chat (never initialised),
+	// which falls back to the show-all bar. Session chats are always tracked (a
+	// SvelteSet, even empty) — see loadPastChat/initRuntime — so their Edits
+	// surface never claims drafts the session didn't touch. Reactive so the
+	// session bar updates as tools record mid-turn.
 	modifiedItems = $state<SvelteSet<string> | undefined>(undefined)
 
 	// Start tracking for a brand-new session chat (empty = "tracked, nothing yet").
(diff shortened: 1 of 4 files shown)


Guilhem-lm:
Addressed the Codex P2 nit in 4c4deccc79 — updated the stale persistence comment in `HistoryManager.svelte.ts` to reflect that session chats are now always tracked (and thus persist a defined mask), while only the untracked global side-panel chat keeps its accumulator `undefined`. The persistence logic itself is unchanged and still correct.

PR #9990, opened 2026-07-07
fix(pipelines): live materialize/dataset editing — stale graph, phantom drafts, stale Save-all deploys

rubenfiszel:
## Summary

Fixes WIN-2139. Ran the pipeline editor through exhaustive live sessions (three structured passes: new-item lifecycle, deployed-item lifecycle, cross-interactions) — selecting nodes, editing materialize target names and ducklake inputs live, deploying, discarding, renaming, re-editing — with new and already-deployed datasets. **Eight bugs** reproduced and fixed, each verified end-to-end in a real browser against a live backend, plus 6 new unit regression tests.

## Bugs found & fixed

1. **Details pane crashed on open** (`Cannot access 'args' before initialization`). Regression from #9953: a `$effect.pre` writes `args` but the `$state` declaration sat below it; when the pane is instantiated inside a flush (right after creating a pipeline script) the effect runs synchronously during init and hits the TDZ. Broke the whole editor — empty canvas, dead pane. Fixed by moving the declaration above the effect.

2. **Live materialize-target rename on a draft didn't update the graph.** `seedDraftOverlays` parsed the draft's stored snapshot (only synced on pane teardown) for `// materialize`. The open draft now uses the live-parsed buffer annotations, keystroke-live.

3. **Live edits on a deployed script (existing dataset) showed nothing — or worse.** `staleForOpen` dropped the persisted write edge as soon as the buffer stopped referencing it, but nothing overlaid the *new* target/reads/writes — retargeting `// materialize` left the script visibly producing nothing; changing the input dataset never re-wired the edge. `applyLiveBufferOverlay` now overlays the materialize target (incl. scd2 `_current`) and live body reads/writes for the open saved script via shared `pushWriteOut`/`pushReadIn`/`materializeOuts` helpers.

4. **Phantom draft after Deploy.** The persist-back cleanup can run mid-save (after `createScript` resolves, before the refetch lands) and mis-reads the just-deployed buffer as unsaved edits — leaving an "Edit (1)" draft identical to the deployed head that survives reload. Fixed with a `deployedFromPane` marker pinned to the exact shipped content (so keystrokes typed *during* the round-trip still promote to a draft — caught by fresh-context review).

5. **Header "Save all" deployed stale draft snapshots.** The open pane's keystrokes only reach the drafts Map on teardown, so "Save all" shipped pre-edit content (verified via the actual `createScript` payload). It now folds `pe.liveContent` into the open draft — the same merge the draft autosave already applies.

6. **An inactive draft lost its input lineage.** Selecting another node dropped a never-deployed draft's read edges + derived cascade triggers from the canvas (writes survived via `outputAssets`; reads had no persisted counterpart and the session cache only covers deployed paths). Drafts now capture `inputAssets` on pane teardown (mirrored through the AI staging path and the autosave bundle); `resolveGraph` renders them for inactive drafts, with `[]` authoritative and `undefined` (legacy bundles) falling back to the session cache.

7. **Renaming a draft deployed its stale snapshot, silently dropping live keystrokes.** Draft rename intentionally auto-deploys (so native triggers can attach), but it deployed the pre-edit drafts-Map entry and then deleted the draft — the user's unsaved buffer was gone entirely. `renameDraft` now folds the live buffer into the rekeyed entry before the auto-deploy.

8. **Renaming a draft with unsaved keystrokes hung the tab and hammered the backend.** The rename rekeys the drafts Map while the pane stays mounted; the persist-back cleanup then emits the *previous* clone's content, which always disagrees with the freshly rewritten entry → rewrite → re-clone → emit the older generation again — an infinite microtask ping-pong that pins the main thread and POSTs `drafts/update` every cycle (observed as a live request spike). Structural fix: a draft-run cleanup only emits when the user actually typed into that clone (`contentAtRegister`, read untracked). This also guards the same hazard for any external rewrite of the open draft (e.g. AI `edit_pipeline_node`). Verified post-fix: rename with live edits produces 8 requests in 10s (1 create + 1 draft autosave + normal fetches) and a responsive page.

## Changes

- `AssetGraphDetailsPane.svelte` — fixes 1, 4, 8; persist-back snapshot now carries `reads` (6).
- `resolveGraph.ts` — fixes 2, 3, 6; shared write/read push helpers.
- `pipelineEditorState.svelte.ts` — stores `inputAssets`; equality guard extended to reads (backward-compatible optional `reads`).
- `pipelineAiHelpers.ts` — `PipelineDraft.inputAssets`; AI staging infers reads alongside writes.
- `PipelineGraphEditor.svelte` — autosave bundle serializes live reads for the active draft.
- `pipeline/[folder]/+page.svelte` — fixes 5, 7.
- `resolveGraph.test.ts` — 6 regression tests (live-annotation precedence, saved-script overlay + dedup, live body lineage, inactive-draft `inputAssets`, authoritative empty reads).

## Screenshots

Before (bug 3): buffer targets `orders_gold`, but the canvas shows an orphaned `orders` and no new node:

07-bug-deployed-rename-before

After: `orders_gold` appears live with its write edge; the deployed `orders` dataset stays (it still exists in the lake):

08-deployed-rename-after-fix

Live input-dataset change on a deployed consumer re-wires the cascade edge keystroke-live:

11-deployed-input-change

## Test plan

- [x] `npm run check` — 0 errors, no warnings in touched files
- [x] AssetGraph vitest suite — 273 tests pass (6 new)
- [x] **Pass A (new items)**: create via insert menu; rapid repeated target renames converge; remove/re-add `// materialize`; ducklake↔s3 target kind swap; auto-derived cascade + `// mute` BellOff + explicit `// on`; discard mid-edit; recreate same path gets a fresh template; autosave reload-restore of live keystrokes; discard persists across reload; two drafts cross-wired; producer-draft discard keeps consumer intact; draft path rename (auto-deploys the live buffer)
- [x] **Pass B (deployed items)**: remove materialize live → edge drops, discard restores; retarget + input change with direct node-to-node selection switches; multi-draft "Save all" ships correct contents; type-junk + undo-to-identical + close leaves no draft; archive via pane modal removes node
- [x] **Pass C (cross)**: view↔edit mode switch preserves edits; schedule-missing placeholder opens the schedule drawer on deployed scripts; `-- partitioned` surfaces the partition run-form field; syntactically broken buffers keep the graph stable and the page responsive
- [x] Loop regression probe: rename-with-live-edits window shows no request spike and a responsive main thread (a spike detector ran through all browser passes; only the initial page-load burst appears)

## Notes

- "Save all (N)" deliberately deploys drafts only; unsaved edits on an open *deployed* script are not a draft until pane teardown and are never silently shipped or lost — they promote to a draft on close. Left as-is.
- Save-all partial-failure UI (error popover on a failing draft) was not exercised — hard to stage a failing deploy locally.
- Local review (fresh-context subagent) found one P1 in the initial fix-4 guard (mid-deploy keystrokes could be swallowed); fixed and re-verified.

### DIFF
--- a/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
+++ b/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
@@ -690,10 +690,17 @@
 			} catch {
 				sendUserToast(`Could not parse code, are you sure it is valid?`, true)
 			}
+			// Pin the exact content this deploy ships: the editor stays live during
+			// the network round-trip, so `script.content` can advance past what was
+			// sent — `deployedFromPane` must record the shipped version, not the
+			// buffer at response time, or mid-deploy keystrokes match the guard and
+			// are never promoted to a draft.
+			const contentAtDeploy = script.content ?? ''
 			const newHash = await ScriptService.createScript({
 				workspace,
 				requestBody: {
 					...script,
+					content: contentAtDeploy,
 					language: script.language,
 					description: script.description ?? '',
 					// Brand-new drafts have no prior hash (buildDraft seeds ''
@@ -728,7 +735,7 @@
 			// backend rejects as a lineage fork ("no 2 scripts can have the
 			// same parent").
 			if (typeof newHash === 'string' && newHash) script.hash = newHash
-			deployedFromPane = { path: script.path, content: script.content ?? '' }
+			deployedFromPane = { path: script.path, content: contentAtDeploy }
 			sendUserToast(`Saved ${script.path}`)
 			// Authoritative save-time schema-contract check (pipelines gap #2b):
 			// warn-only, post-commit. Fire-and-forget — must never gate the save.
--- a/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
+++ b/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
@@ -482,6 +482,7 @@
 	// restores its input. Guarded on the path so a staging round-trip
 	// (emit → page → runFormInitialArgs) doesn't re-seed and loop. The read-only
 	// branch uses PipelineScriptView's own onArgsChange instead.
+	let args = $state<Record<string, any>>({})
 	let argsSeedPath: string | undefined = undefined
 	$effect.pre(() => {
 		const p = script?.path
@@ -532,6 +533,16 @@
 				const orig = latest && latest.path === captured.path ? latest : origAtRegister
 				if (!orig || orig.path !== captured.path) return
 				if ((captured.content ?? '') === (orig.content ?? '')) return
+				// The effect can re-run mid-save — after createScript resolved
+				// but before the refetch lands — with `orig` still holding the
+				// pre-deploy version. The buffer isn't "unsaved edits" then, it
+				// IS the new deployed head; emitting would resurrect it as a
+				// phantom draft identical to what was just deployed.
+				if (
+					deployedFromPane?.path === captured.path &&
+					(captured.content ?? '') === deployedFromPane.content
+				)
+					return
 			}
 			const writes = (liveBodyAssets ?? [])
 				.filter((a) => {
@@ -547,8 +558,11 @@
 		}
 	})
 
-	let args = $state<Record<string, any>>({})
 	let saving = $state(false)
+	// What this pane last deployed, so the persist-back cleanup can tell "the
+	// buffer equals the new deployed head" from real unsaved edits (plain
+	// variable: only read inside the untracked cleanup).
+	let deployedFromPane: { path: string; content: string } | undefined = undefined
 	let isDraft = $derived(draftScript != undefined)
 
 	// Single trash-bin button opens one modal that exposes both Archive
@@ -714,6 +728,7 @@
 			// backend rejects as a lineage fork ("no 2 scripts can have the
 			// same parent").
 			if (typeof newHash === 'string' && newHash) script.hash = newHash
+			deployedFromPane = { path: script.path, content: script.content ?? '' }
 			sendUserToast(`Saved ${script.path}`)
 			// Authoritative save-time schema-contract check (pipelines gap #2b):
 			// warn-only, post-commit. Fire-and-forget — must never gate the save.
--- a/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
+++ b/frontend/src/lib/components/assets/AssetGraph/AssetGraphDetailsPane.svelte
@@ -54,6 +54,10 @@
 		// `selection` when present. Used by the pipeline + menu so a new
 		// pipeline script opens inline instead of navigating to /scripts/add.
 		draftScript?: Script | undefined
+		// Whether the draft entry behind `draftScript` has captured its read
+		// lineage (`inputAssets != undefined`). Gates the teardown skip below: a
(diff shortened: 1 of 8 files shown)


rubenfiszel:
Both P1s addressed in 237aaf9120:

**1. Duplicate live edges** — real, but the fix keeps the `applyLiveBufferOverlay` block: the in-session preview (`PipelineEditorView`) passes empty `inferredWrites/ReadsByPath`, so that block is the only live-lineage source on that surface. Instead `overlayInferredLineage` now dedupes against the **accumulated** edges (same access or `rw`) rather than `base.edges` only — which also covers the pre-existing latent dup when a body write coincides with the `// materialize` target. Added a regression test (same ref arriving via both `liveBodyAssets` and the inferred maps → exactly one edge) and verified in-browser: a live input edit on a deployed script yields unique `cons:`/`trig-a:` edge ids, no console key warnings.

**2. Skipped first capture of a seeded draft's reads** — real. The teardown skip now additionally requires `draftInputsCaptured` (new pane prop: the draft entry's `inputAssets != undefined`), so a just-seeded draft's first teardown always emits and captures reads even with an untouched buffer. The rename ping-pong still converges: each emit records `inputAssets`, so an entry that predates the capture bounces at most once before the guard engages. Verified in-browser: a draft created from an asset's + button, never typed into, keeps its read + cascade edges after selecting another node.

rubenfiszel:
Codex round-3 P1 addressed in c15c9e0511: an uncaptured entry (`inputAssets: undefined`) receiving `reads: []` no longer short-circuits — the first capture is always recorded, so the entry leaves the legacy session-cache fallback and stale read edges can't survive the pane closing. Added the requested state-layer unit tests (undefined→[] rewrites the map; []≡[] and legacy omitted-reads stay idempotent).

Also ran a final end-to-end sanity sweep on a fresh folder covering every fix in this PR in one instrumented session (request-spike detector + console watch throughout, zero spikes, no unexpected errors):
- create draft → pane renders (bug 1); scd2 live retarget shows `_current` (bug 2)
- pane Deploy → no phantom draft (bug 4)
- deployed script live retarget + input change → unique edge ids (bug 3, review P1-1)
- promote → discard → full revert; Save all ships the live buffer (bug 5)
- untouched seeded reader keeps reads after deselect (bug 6, review P1-2); removing all reads leaves no stale edges after close (this finding)
- rename with live edits → deploys the live content, zero request spikes, responsive main thread (bugs 7–8)

`npm run check`: 0 errors; AssetGraph suite: 277 tests pass.

rubenfiszel:
Valid — addressed in bae3fbecae. `AssetsDropdownButton` mutates `asset.alt_access_type` in place (and ScriptEditor preserves the override across re-inference), so lineage can change with zero text delta and the content-only skip dropped the capture.

The teardown skip is now "nothing to persist": it computes the emit's writes/reads first and only returns when content AND lineage match the entry's registration-time snapshots. Concretely, `draftInputsCaptured` was replaced by `draftOutputAssets`/`draftInputAssets` props (snapshotted untracked at effect registration so a cleanup firing during a selection switch compares against ITS entry, not the next one). This also subsumes the just-seeded first-capture case (`inputAssets` undefined never matches). Ping-pong convergence is unchanged: `liveBodyAssets` is a single live binding — it doesn't alternate generations like the content clones did — so a lineage mismatch rewrites once and the next comparison is equal.

Re-ran the full end-to-end sweep (all 8 bugs + the 3 prior review findings) on a fresh folder with the reworked guard: all steps pass, zero request-spike windows, main thread responsive. `npm run check` 0 errors, 277 AssetGraph tests pass.

PR #9991, opened 2026-07-07
feat: smooth bursty AI chat streaming with a typewriter reveal

AlexRV12:
## Summary

Streamed copilot text arrives in coarse bursts — for some Anthropic model/tier combos the provider batches ~50–80 tokens into one delta every ~450ms — so the chat reads as freeze→jump→freeze. We instrumented every layer (frontend render, backend proxy, TCP, routing) and a raw `curl` straight to `api.anthropic.com` reproduced the identical cadence, proving the burstiness is **upstream** (multiple tokens arrive inside one `text_delta` JSON string, so it can't be network coalescing). No request-path fix exists.

This PR adds a frontend **smoothing layer** that decouples *display* from *arrival*: pushed text lands in a non-reactive buffer and a paint loop reveals a slice per animation frame, so the ~450ms bursts read as continuous typing.

## Changes

- **New `chat/typewriterReveal.ts`** — standalone, Svelte-free `TypewriterReveal` (the only coupling to reactive state is an `onReveal` callback, so it's unit-testable with an injected clock/scheduler):
  - Time-based self-correcting rate (`rate = backlog / smoothingMs`, `reveal = rate × elapsedMs`, ≥1-char floor) — backlog can neither run away nor stall; steady state trails arrival by ~`smoothingMs`.
  - Emit throttle (~30fps) so downstream markdown re-parse / autoscroll reflow runs ~30×/s, not ~60.
  - Fast-path: backlog ≥ `maxBacklogChars` (1500) dumps whole — cached / non-streaming replies show instantly, never stretched.
  - Idle-suspend (no live rAF when idle) + large-elapsed clamp (backgrounded-tab resume catches up over several emits, not one blast).
  - Surrogate-pair-safe cutting that can neither busy-loop nor stall (floor-1-inside-a-pair reveals the whole pair; a lone trailing high surrogate suspends until its low half arrives).
  - `instant` mode for reduced-motion / SSR (reveals synchronously, no pacing).
  - `flush()` reveals the remainder and compacts the buffer; `reset()` drops backlog.
- **`chat/AIChatManager.svelte.ts`** — owns two instances, `replyReveal` (→ `currentReply`) and `reasoningReveal` (→ `currentReasoning`); both wired into the stream callbacks. `onNewToken`/`onReasoningDelta` push; `onMessageEnd` flushes **both before** reading state into committed messages (flush-before-read invariant — one funnel covering clean / tool-boundary / abort / error); pre-stream + turn-`finally` reset. Reduced-motion/SSR sampled once via `prefersInstantReveal()`.
- **New `chat/typewriterReveal.test.ts`** — 11 vitest cases (fake clock, no browser).

## Verification

- **Unit tests** — 11 cases green: gradual reveal, no-loss/no-dup after flush, repeated flush across tool boundaries, reset, fast-path, surrogate safety, instant mode, steady-state convergence, throttle-bounded emit frequency, idle suspend/resume, large-elapsed clamp. Full suite (`typewriterReveal` + `AIChatManager`) 71 pass.
- **Render-cost measurement (the one open risk).** The typewriter mutates `currentReply` on a paint loop instead of per burst, so the concern was per-emit markdown re-parse + code-fence re-highlight climbing with reply length. Measured in-browser (Playwright + `PerformanceObserver` long-tasks + rAF-gap tracking) on a **13,139-char reply with 6 inline highlighted code fences**, streamed from `claude-sonnet-4-6 · high`:
  - **0 long-tasks (>50ms)** over the entire stream.
  - Frame-gap **p50/p95 = 17ms (60fps) across every 5s window** — no upward trend with length (only ~6 isolated >33ms frames mid-stream, GC/network noise).
  - Conclusion: svelte-exmarkdown / svelte-highlight reuse unchanged completed fences, so per-emit cost is flat. The planned completed-fence-freeze and markdown prefix/tail-split follow-ups are **not needed** and were dropped.

## Screenshots

None — the change is **temporal** (streaming cadence), so a still frame of the final reply is identical to before this PR and proves nothing. The render-cost numbers above are the objective evidence.

## Deliberately not built (rationale)

- **`Intl.Segmenter` grapheme safety.** `emit()` is surrogate-pair-safe, but a ZWJ grapheme cluster (e.g. 👨‍👩‍👧) can flash half-drawn for one ~33ms frame before healing — cosmetic and self-healing. A correct-and-fast fix (keep the O(1) fast path; run a bounded-window Segmenter only when the boundary char is "risky") is ready to drop in but held; add only if observed live.
- **Backend TCP `set_nodelay(true)`** was tested during the investigation, did not affect this, and was reverted — independent change, file separately if desired.

## Test plan

- [ ] With reduced-motion **off**, run an AI chat turn against a bursty Anthropic model — reply and reasoning render as continuous typing rather than freeze→jump.
- [ ] Hit **Stop** mid-reply — the partial answer is fully retained (not truncated to the last revealed frame) and carried into the next turn's context.
- [ ] Cross a tool-call boundary mid-turn — no lost/duplicated text, no flicker at message end.
- [ ] A cached / very short reply appears immediately (fast-path), not typed out.
- [ ] An OpenAI (fine-grained) model still reads smoothly with no worse first-text latency.
- [ ] With `prefers-reduced-motion: reduce`, the reply appears without the typewriter animation.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -58,6 +58,7 @@ import { BROWSER } from 'esm-env'
 import { workspaceStore, type DBSchemas } from '$lib/stores'
 import { askTools, prepareAskSystemMessage, prepareAskUserMessage } from './ask/core'
 import { readDocsPageTool, searchDocsTool } from './docs/core'
+import { TypewriterReveal } from './typewriterReveal'
 import { chatState, DEFAULT_SIZE, triggerablesByAi } from './sharedChatState.svelte'
 import {
 	createAppBackendRunnableContextElement,
@@ -101,6 +102,11 @@ import { scopedKey, onUserChange, migrateLegacyLocalStorage } from '$lib/userSco
 import { getLocalSetting, storeLocalSetting } from '$lib/utils'
 import { AttachedFilesStore } from './files/attachedFiles.svelte'
 import { appendAttachedFilesRoster } from './files/fileTools'
+
+// SSR and users who prefer reduced motion get no typewriter pacing.
+function prefersInstantReveal(): boolean {
+	return !BROWSER || (window.matchMedia?.('(prefers-reduced-motion: reduce)').matches ?? false)
+}
 
 // Compaction of the stored history: once the projected request size
 // (contextTokens — the provider's report when current, a fresh chars/4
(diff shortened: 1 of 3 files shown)


AlexRV12:

PR #9993, opened 2026-07-07
fix(sessions): auto-rename regression + preview-panel and fork nits

Guilhem-lm:
## Summary

Fixes four AI-sessions nits. The headline one: **new sessions stopped auto-generating a title** from the first message. Root-caused with runtime instrumentation to a silently-swallowed throw — the Anthropic SDK now refuses a *non-streaming* request whose `max_tokens` implies a >10-minute worst case (`60min × max_tokens / 128000 > 10min`, i.e. `max_tokens > ~21333`). The session-summary generator (and CronGen/PredicateGen/StepInputGen) run non-streaming with the model default (`claude-sonnet-4-6` → 64000, opus → 32000), so every current Anthropic model tripped the guard; the error was caught by the `afterFirstTurnSaved` hook's `.catch()` and only logged, leaving the placeholder `"Xxx session"` title.

The other three are small session-UX fixes bundled from the same pass.

## Changes

- **fix: cap metadata `max_tokens`** — new `METADATA_MAX_TOKENS = 4096`, threaded as a `maxTokensCap` ceiling through `resolveMaxTokens` → `getModelSpecificConfig`/`getProviderAndCompletionConfig` (OpenAI path) and `buildAnthropicProxyRequest` (Anthropic Messages path). Only `getNonStreamingMetadataCompletion` passes it, so streaming and normal completions are untouched. Metadata outputs are short, so the cap never truncates real output. Adds a regression test in `lib.anthropicRouting.test.ts`.
- **perf: don't mount preview tabs when the side panel is collapsed** — the lazy-mount `$effect` now bails when `previewCollapsed`, so switching to a collapsed session (or a chat tool opening a tab in one) no longer boots a full Windmill app/iframe behind a zero-width pane. Mounts on expand when `previewCollapsed` flips back.
- **fix: drop redundant `-fork` suffix** from auto-generated fork names (`defaultForkId`) — the id already carries the `wm-fork-` prefix.
- **fix: default "also delete forked workspace" to `false`** in the delete-session modal (safer default; was `true`).

## Test plan

- [x] Metadata cap: `vitest run lib.anthropicRouting.test.ts` (7/7 pass; asserts uncapped path stays 64000, metadata path caps to `METADATA_MAX_TOKENS`).
- [x] Auto-rename: browser-verified end-to-end on `/sessions` — new session, sent a message, header went `"Durable session"` → **"Primary Colors List"** (console confirmed `setGeneratedSessionSummary result {applied: true}` with no throw). See screenshot.
- [ ] Collapse-mount: on `/sessions`, collapse the preview panel, have a background chat tool open a tab, then expand and confirm the active tab boots on expand (no blank pane, no double-mount).
- [ ] Delete-forked default: delete a session tied to a fork workspace and confirm the "also delete forked workspace" checkbox now defaults to unticked.

## Screenshots

Auto-rename fix — a brand-new session's title generated from the first message (`"Primary Colors List"`), acting on the `guilhem` workspace:

rename-auto-title

The other three surfaces are state-gated and not captured: the collapse-mount change has no visible DOM delta (behavior only); the fork-name default only appears in the transient create-fork input; the delete-forked toggle only renders for a session tied to a fork workspace. All three are one-line default/string flips, reviewed and type-checked.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/openai-responses.ts
+++ b/frontend/src/lib/components/copilot/chat/openai-responses.ts
@@ -454,12 +454,14 @@ export async function getNonStreamingOpenAIResponsesCompletion(
 		workspace?: string
 		resourcePath?: string
 		forceModelProvider?: AIProviderModel
+		maxTokensCap?: number
 	}
 ): Promise<string> {
 	const { provider, config } = getProviderAndCompletionConfig({
 		messages,
 		stream: false,
-		forceModelProvider: options?.forceModelProvider
+		forceModelProvider: options?.forceModelProvider,
+		maxTokensCap: options?.maxTokensCap
 	})
 
 	const { instructions, input } = convertMessagesToResponsesInput(messages)
--- a/frontend/src/lib/components/copilot/lib.anthropicRouting.test.ts
+++ b/frontend/src/lib/components/copilot/lib.anthropicRouting.test.ts
@@ -195,6 +195,23 @@ describe('Anthropic Messages API routing', () => {
 		expect(headers['X-Resource-Path']).toBe('u/admin/foundry')
 	})
 
+	it('caps max_tokens for metadata completions so the Anthropic SDK stays non-streaming', async () => {
+		const { getNonStreamingCompletion, getNonStreamingMetadataCompletion, METADATA_MAX_TOKENS } =
+			await import('./lib')
+		// claude-sonnet defaults to 64000 max_tokens; the Anthropic SDK refuses a
+		// non-streaming request that large (>10min worst case), which silently broke
+		// session auto-rename and the other metadata generators.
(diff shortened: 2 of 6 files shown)


Guilhem-lm:
Addressed the P2 from @pi and Codex in 613c9d8 — thanks.

`getNonStreamingOpenAIResponsesCompletion` now accepts `maxTokensCap` and threads it into its `getProviderAndCompletionConfig` call, so the metadata cap applies on the OpenAI/Azure Responses path too (not just the Anthropic Messages and chat.completions fallback paths). Added a regression test asserting `max_output_tokens` is capped to `METADATA_MAX_TOKENS` on that path while the uncapped path stays at the model default (16384 for gpt-4o).

Note: the failing `ai_evals_global` check is pre-existing lockfile drift (`npm ci` → `Missing: picomatch@4.0.5 from lock file`), unrelated to this diff — `npm_check` passes.

PR #10001, opened 2026-07-08
feat: add datatable read and write MCP tools

rubenfiszel:
## Summary

Fixes WIN-2143.

Exposes Windmill **data tables** as MCP operations. Data table access is a dedicated, **opt-in capability** in the MCP URL generator — off by default in every mode — with a read/write sub-toggle and an optional restriction to specific data tables. When enabled, the client gets tools to list tables, read rows (with WHERE conditions), and (if write is on) insert/update rows.

New/existing REST endpoints are marked `x-mcp-tool: true` in the OpenAPI spec and auto-generated into the Rust tool registry + the frontend endpoint list; the datatable tools are then gated by a dedicated `mcp:datatables:` scope rather than the generic endpoint scopes.

### Tools
Read: `listDataTables`, `listDataTableTables`, `getDataTableTableSchema` *(existing endpoints, now MCP-exposed)*, and new `queryDataTable` (SELECT with `where_clause`/`select`/`order_by`/`limit`/`offset`, capped at 1000).
Write: new `insertDataTable` (row insert; omitted columns keep their default) and `updateDataTable` (update rows matching a required `where_clause`).

### Data tables access control
A dedicated **"Data tables"** section (shown in every mode except "All", where `mcp:all` already grants everything):
- **Master toggle** — off by default. Datatable tools are never exposed unless this is on, independent of the favorites/all/folder/endpoint scopes.
- **Write access** sub-toggle — off by default (read-only); hidden when the token is read-only. On ⇒ also exposes insert/update.
- **Restrict to specific data tables** — optional multiselect; empty = all data tables.
- A warning that this grants **full access to the data** (no per-user/row-level permissions on data tables yet — coming soon — so it's meant for trusted/dev use).

Scope grammar: `mcp:datatables:read:<names|*>` / `mcp:datatables:write:<names|*>`. Absence ⇒ no datatable access. `mcp:all` grants read+write on all data tables.

Enforcement:
- **Tool exposure**: gated in the MCP runner by `datatable_tool_allowed(read|write)` — datatable tools are opt-in regardless of mode; write tools also require a non-read-only token (they're POST).
- **Per-datatable name + list filtering**: in `WindmillBackend::call_endpoint` (the only place with the original MCP scopes — the JWT the runner mints for the proxied HTTP call is route-scoped and no longer carries `mcp:` scopes). Per-call `datatable_name` is rejected if out of scope; `listDataTables`/`listDataTableTables` responses are filtered so a restricted token never sees other names.

## Changes
- `backend/windmill-api-workspaces/src/workspaces.rs`: `query_datatable`/`insert_datatable`/`update_datatable` handlers + `connect_datatable`/`quote_pg_ident`/`check_datatable_schema` helpers and routes.
- `backend/windmill-api/openapi.yaml`: six datatable endpoints marked `x-mcp-tool`; regenerated `auto_generated_endpoints.rs` + `mcpEndpointTools.ts`.
- `backend/windmill-mcp/src/common/scope.rs`: `datatables_read`/`datatables_write`/`datatables` on `McpScopeConfig`; parse `mcp:datatables:read|write:`; `datatable_access_level`, `datatable_tool_allowed`, `is_datatable_allowed`; OAuth subset (`contains`); tests.
- `backend/windmill-mcp/src/server/runner.rs`: gate datatable tools by the dedicated scope in `list_tools` + `call_tool`.
- `backend/windmill-api/src/mcp/core.rs`: per-datatable arg check + list-response filtering in `call_endpoint`.
- `backend/windmill-api/src/mcp/oauth_server.rs`: advertise `mcp:datatables:read:*`/`write:*`.
- `frontend/src/lib/components/mcp/McpScopeSelector.svelte`: dedicated Data tables section (master + write toggle + restriction picker + warning); datatable tools excluded from the generic endpoint picker.
- `tokio-postgres` dep on `windmill-api-workspaces`; unit tests for `quote_pg_ident` and the datatable scope logic.

## Safety
- Identifiers (schema/table/column/select) always quoted via `quote_pg_ident` (doubles `"`, rejects empty/NUL).
- Values bound as one `$1::jsonb` param via `jsonb_populate_record` — Postgres coerces types, no value interpolation.
- `where_clause`/`order_by` are **validated with `sqlparser` and re-serialized from the parsed AST** before interpolation: a fragment must be a single self-contained predicate / ORDER-BY list with no set operations, no subqueries, and no trailing tokens. This blocks the `UNION`/subquery-closing breakout (which would otherwise bypass the system-schema guard and LIMIT) and the error-based/`IN (SELECT …)` cross-table exfiltration. Complex boolean predicates and multi-column ordering still work.
- Residual trust boundary: a permitted expression can still call any function the datatable's **connection role** is granted (`pg_sleep`, or `pg_read_file` if that role is superuser) — scope datatable roles accordingly (not superuser).
- Datatable access requires the explicit `mcp:datatables:` scope; write additionally requires a non-read-only token, and the REST write endpoints reject **operators** (`require_datatable_writer`).

## Design decisions
- Datatable access is **off by default and orthogonal to scripts/flows/endpoints** — only the dedicated toggle (or `mcp:all`) grants it.
- Data tables are PostgreSQL only (DuckLake is a separate substrate); non-postgres fails at connect with a clear error.

## Test plan
- [x] `cargo check` (windmill-api, windmill-api-workspaces); `cargo test -p windmill-mcp scope::` (9 tests) and `-p windmill-api-workspaces` (quote_pg_ident + `where_clause`/`order_by` sanitizer incl. the reported UNION-injection payload)
- [x] Verified the exact handler SQL against a real Postgres via `psql` (type coercion, serial default, `to_jsonb`, per-column update, quoted identifiers)
- [x] `npm run check:fast`; svelte-autofixer
- [x] Browser: "Data tables" toggle is off by default in Favorites mode; enabling reveals the warning + Write toggle + restriction picker; datatable tools are absent from the generic API Endpoints list; generating with write + `main` produces token scope `mcp:favorites mcp:datatables:write:main` (verified in the DB)
- [ ] End-to-end MCP client call against a live data table, incl. a read-only token rejected on write and a restricted token rejected on an out-of-scope data table

## Screenshots

Dedicated "Data tables" section — opt-in master toggle, write sub-toggle, restriction picker, and access warning:
mcp-datatable-toggle

MCP URL generator — datatable read tools shown (read-only token hides insert/update):
mcp-datatable-readonly

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -14470,6 +14470,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.9",
+ "sqlparser 0.59.0",
  "sqlx",
  "strum",
  "tokio",
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -14473,6 +14473,7 @@ dependencies = [
  "sqlx",
  "strum",
  "tokio",
+ "tokio-postgres",
  "tracing",
  "uuid",
  "windmill-api-auth",
--- a/backend/windmill-api-workspaces/Cargo.toml
+++ b/backend/windmill-api-workspaces/Cargo.toml
@@ -47,6 +47,7 @@ serde_json.workspace = true
 sha2.workspace = true
 sqlx.workspace = true
 tokio.workspace = true
+tokio-postgres.workspace = true
 tracing.workspace = true
 uuid.workspace = true
 strum.workspace = true
--- a/backend/windmill-api-workspaces/Cargo.toml
+++ b/backend/windmill-api-workspaces/Cargo.toml
@@ -46,6 +46,7 @@ serde.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
 sqlx.workspace = true
+sqlparser = { version = "0.59.0", features = ["visitor"] }
 tokio.workspace = true
 tokio-postgres.workspace = true
 tracing.workspace = true
--- a/backend/windmill-api-workspaces/src/workspaces.rs
+++ b/backend/windmill-api-workspaces/src/workspaces.rs
@@ -131,6 +131,9 @@ pub fn workspaced_service() -> Router {
             "/get_datatable_table_schema",
             get(get_datatable_table_schema),
         )
+        .route("/query_datatable", get(query_datatable))
+        .route("/insert_datatable", post(insert_datatable))
+        .route("/update_datatable", post(update_datatable))
         .route("/edit_datatable_config", post(edit_datatable_config))
         .route("/git_sync_enabled", get(get_git_sync_enabled))
         .route("/edit_git_sync_config", post(edit_git_sync_config))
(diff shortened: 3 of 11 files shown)


rubenfiszel:
Addressed the review findings in d2a682a:

**P0 — SQL injection via `UNION`/subquery-closing in `query_datatable` `where_clause`/`order_by`** (and the P1 error-based exfiltration in `update_datatable`): fixed. `where_clause` and `order_by` are no longer interpolated raw. Each fragment is now parsed with `sqlparser` (PostgreSQL dialect) and rejected unless it is a single, self-contained expression / ORDER-BY list:
- `where_clause` must parse as one `Expr` that consumes **all** tokens — a fragment that closes the surrounding derived table and appends a `UNION` (`false) _wm_row UNION ALL SELECT … --`) leaves a trailing `)` and is rejected.
- `order_by` must parse as the ORDER BY of a plain single `SELECT` (a `SetExpr::Select`, not a set operation), so a smuggled `UNION` fails to parse.
- **Any nested subquery** in either fragment is rejected via a `Visitor` (`pre_visit_query`), which closes the error-based exfiltration vector (`1 = CAST((SELECT secret …) AS INTEGER)`) and `IN (SELECT …)` cross-table reads.
- The value interpolated is the **parser's own re-serialization** of the validated AST, not the raw bytes — so a trailing `--` comment can't survive to comment out the `LIMIT`/paren either.

Complex boolean predicates (AND/OR/BETWEEN/ILIKE/function calls/`ANY`, multi-column ORDER BY with ASC/DESC) still work. Unit tests added: `sanitize_predicate_accepts_and_canonicalizes`, `sanitize_predicate_rejects_injection` (incl. the exact reported UNION payload), `sanitize_order_by_accepts_and_rejects`.

**P2 — member-level write authz**: `insert_datatable`/`update_datatable` now call `require_datatable_writer`, which rejects **operators** (the workspace's read-mostly tier). Non-operator members can write (consistent with the member-level read handlers and data tables being a shared workspace store); there is still no per-resource ACL, which is the intended tier for this shared store.

**P2 — function-call trust boundary** (`pg_sleep`, and `pg_read_file` etc. if the datatable role is superuser): a permitted expression can still call any function the datatable's connection role is granted. This is now documented in the `sanitize_sql_predicate` doc comment and the PR safety notes — the connection role's privileges are the trust boundary, so datatable roles should be scoped (not superuser).

**Minor — `listDataTableSchemas`**: it is *not* marked `x-mcp-tool` in `openapi.yaml` (so it isn't exposed); it only appears in the defensive `datatable_access_level`/`datatable_list_filter_key` match arms so it's already handled if ever exposed. No change needed.

Still open (as noted in the test plan): end-to-end exercise from a live MCP client. The injection fix itself is covered by the new unit tests + a psql check that a sanitized predicate returns the expected rows.

PR #10004, opened 2026-07-08
fix: preserve worker group tag override on 'Run again'

hugocasa:
Fixes WIN-2142

## Summary

When clicking 'Run again' on a job, the worker group tag override from the original run was lost: the main button only encoded args in the URL hash, and the 'Run immediately with same args' dropdown did not pass a tag to the run APIs. Both paths now carry the original job's tag forward, with dynamic (templated) tags re-resolved instead of pinned.

## Changes

- `computeSharableHash` (utils.ts) accepts an optional tag and encodes it in the URL hash under the reserved key `__tag`, with a `t:` value prefix. No JSON-encoded arg value can start with `t:` (JSON strings start with `"`, numbers with a digit or `-`, etc.), so the new `extractTagFromSharableHash` counterpart can unambiguously tell a carried tag apart from a genuine arg named `__tag` for every possible tag value, including JSON-parseable tags like `123`. A carried tag and an arg named `__tag` coexist as duplicate keys, so both are preserved.
- Run detail page: the 'Run again' button passes `job?.tag` into the hash, and `runImmediately()` passes `tag: job?.tag` to `runScriptByHash` / `runScriptByPath` / `runFlowByPath`.
- Scripts and flows detail pages: the tag is extracted from the parsed hash into `overrideTag` via `extractTagFromSharableHash` and removed before the remaining params are parsed as args, so it never leaks into the input form.
- Dynamic tags: a job only stores the resolved tag (backend interpolates `$args[...]`/`$workspace` at push time), so pinning it would go stale if the user edits args before re-running. When the loaded script/flow's configured tag is `$args[...]`-templated (`isDynamicTag`), the carried tag is dropped so the backend re-resolves it from the submitted args, and `RunForm` shows a note in place of the tag override line explaining that the previous run's tag was not applied.
- `RunForm`'s debounced hash rewrite includes `overrideTag` and reacts to tag changes as well as arg changes, so the URL stays in sync when the worker group is changed or reset from the Advanced popup, and the tag survives a page refresh or a copied link.
- Unit tests in `utils.test.ts` cover the round-trip, the `__tag` arg collision cases, JSON-parseable tags, a tag that itself starts with the value prefix, and `isDynamicTag`.

## Notes

- A job record only stores the final tag, not whether it was an override, so 'Run again' pins the original run's tag for scripts/flows with a plain (non-templated) tag. For runs without an override this is the default tag (same routing). If the script/flow's tag configuration changes between runs, a re-run keeps the original run's tag rather than picking up the new default, which matches the "same run" intent of the button; the override is visible in the form and can be reset from Advanced.
- A genuine manual override on a script/flow whose configured tag is templated is indistinguishable from the resolved default in the job record, so it is also re-resolved on 'Run again' (with the note shown). Persisting an "explicit override" flag on jobs backend-side would remove that ambiguity and is left as a follow-up if it matters in practice.

## Test plan

- [x] `npm run check:fast` passes; `npx vitest run src/lib/utils.test.ts` passes (73 tests)
- [x] Script run with `?tag=my-custom-tag` → 'Run again' navigates to `/scripts/get/...#__tag=t%3Amy-custom-tag&name=...`, the form shows `tag override: my-custom-tag`, the arg is prefilled, and submitting creates a job with tag `my-custom-tag` (verified in DB)
- [x] 'Run immediately with same args' creates a job with tag `my-custom-tag` (verified in DB)
- [x] Flow run with a tag override → 'Run again' lands on `/flows/get/...` with the tag override captured
- [x] After landing on the get page, the URL hash keeps `__tag` through the debounced rewrite and the override survives a browser refresh
- [x] Changing the worker group from the Advanced popup without touching args updates the URL hash; resetting it removes `__tag` (verified in browser)
- [x] Script with tag `worker-$args[env]` run with `env=prod` (job tag `worker-prod`) → 'Run again' shows the dynamic-tag note instead of a tag override; editing `env` to `dev` and running creates a job with tag `worker-dev`, re-resolved from the edited args (verified in DB)
- [x] Args named `__tag` round-trip as args, including alongside a carried tag (both preserved); JSON-parseable tags like `123` are preserved (unit tests + browser check)

## Screenshots

Script 'Run again' with the original run's tag override prefilled:

run-again-tag-override

Flow 'Run again' with the tag override prefilled:

run-again-tag-override-flow

'Run again' on a job whose script has a dynamic tag (`worker-$args[env]`): the resolved tag is not pinned and a note explains why:

run-again-dynamic-tag-note

### DIFF
--- a/frontend/src/lib/components/RunForm.svelte
+++ b/frontend/src/lib/components/RunForm.svelte
@@ -201,6 +201,7 @@
 		jsonEditor?.setCode(code)
 	}
 	$effect(() => {
+		overrideTag
 		Object.keys(args ?? {}).forEach((key) => {
 			args?.[key]
 		})
--- a/frontend/src/lib/components/RunForm.svelte
+++ b/frontend/src/lib/components/RunForm.svelte
@@ -160,7 +160,7 @@
 			debounced && clearTimeout(debounced)
 			debounced = setTimeout(() => {
 				const nurl = new URL(window.location.href)
-				nurl.hash = computeSharableHash(args)
+				nurl.hash = computeSharableHash(args, overrideTag)
 
 				try {
 					replaceState(nurl.toString(), page.state)
--- a/frontend/src/lib/components/RunForm.svelte
+++ b/frontend/src/lib/components/RunForm.svelte
@@ -114,6 +114,7 @@
 		scheduledForStr: string | undefined
 		invisible_to_owner: boolean | undefined
 		overrideTag: string | undefined
+		overrideTagNote?: string
 		args?: Record<string, any>
 		jsonView?: boolean
 		isValid?: boolean
@@ -132,6 +133,7 @@
 		scheduledForStr = $bindable(),
 		invisible_to_owner = $bindable(),
 		overrideTag = $bindable(),
+		overrideTagNote = undefined,
 		args = $bindable(),
 		jsonView = false,
 		isValid = $bindable(true)
@@ -387,6 +389,10 @@
 			{#if overrideTag}
 				<div class="flex-row-reverse flex w-full text-primary text-sm">
 					tag override: {overrideTag}
+				</div>
+			{:else if overrideTagNote}
+				<div class="flex-row-reverse flex w-full text-secondary text-xs">
+					{overrideTagNote}
 				</div>
 			{/if}
 			{#if invisible_to_owner}
--- a/frontend/src/lib/utils.test.ts
+++ b/frontend/src/lib/utils.test.ts
@@ -1,6 +1,8 @@
 import { describe, it, expect } from 'vitest'
 import {
 	cleanValueProperties,
+	computeSharableHash,
+	extractTagFromSharableHash,
 	getQueryStmtCountHeuristic,
 	parseDbInputFromAssetSyntax
 } from './utils'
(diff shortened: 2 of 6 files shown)


hugocasa:
Addressed the review findings in 8e73c8f:

- **Codex P2 (`__tag` collision with user args)**: fixed. The carried tag is now written raw while args stay JSON-encoded, and the new `extractTagFromSharableHash` only treats a `__tag` value as a tag when it is not JSON-parseable. An arg named `__tag` therefore round-trips as an arg (and wins over a carried tag); a JSON-parseable tag is dropped from the hash instead of corrupting args. Covered by unit tests in `utils.test.ts`.
- **Claude P2 (`$args[...]`-templated tags pin the resolved value)**: documented in the PR notes. The frontend cannot distinguish a resolved templated tag from a plain one from `job.tag` alone, and unchanged args (the common case) route identically; detecting templates would require fetching the deployed script's tag config on the run page, which seems disproportionate here.

hugocasa:
Addressed the second Codex round in 95b9c19:

- **JSON-parseable tags dropped**: the JSON-parseability heuristic is gone. The carried tag is now always written with a `t:` value prefix, which no JSON-encoded arg value can start with, so every tag value round-trips (including `123`, `true`, and even tags that themselves start with `t:`) while an arg named `__tag` still round-trips as an arg. Unit tests updated to assert preservation instead of dropping.
- **Stale hash on tag-only changes**: the RunForm effect that rewrites the sharable hash now tracks `overrideTag` in addition to args. Verified in the browser: changing the worker group from the Advanced popup updates `__tag` in the URL without touching args, and resetting to default removes it.

hugocasa:
Addressed the remaining Codex finding:

- **`__tag` arg + worker tag override both present**: fixed. `computeSharableHash` now builds the hash from entry pairs instead of an object, so the carried tag and an arg named `__tag` coexist as duplicate keys, distinguished by the `t:` value prefix. `extractTagFromSharableHash` uses `getAll` to pull only the prefixed value and re-appends the arg entries. Both round-trip now; the unit test that previously codified dropping the tag asserts both are preserved, and I verified in the browser that a hash carrying both shows the tag override and keeps the `__tag` arg through the debounced rewrite.
- **Dynamic-tag scripts discard an explicit override on 'Run again'**: unchanged, as discussed in the PR notes. An explicit override is indistinguishable from the resolved default in the job record, the note in the form makes the behavior visible, and a backend "explicit override" flag is the follow-up that would remove the ambiguity.

PR #10014, opened 2026-07-08
feat(db-health): add connection sizing guidance

rubenfiszel:
## Summary

The **Database health → Database Connections** panel showed current/max connections but gave operators no guidance on how to size Postgres `max_connections` for their deployment. This derives an estimate from the live Windmill worker fleet and surfaces a recommended floor, plus a warning when `max_connections` is under it.

Fixes WIN-2147

## How the estimate works

Windmill's own connection demand is deterministic from its pool sizing, and mirrors `db_connect.rs`:
- If **`DATABASE_CONNECTIONS`** is set, it caps *every* process's pool (server, indexer, worker) at that value — so each worker instance opens up to that many connections and the worker estimate is `DATABASE_CONNECTIONS × instances`.
- Otherwise each **worker instance** shares a pool of `DEFAULT_MAX_CONNECTIONS_WORKER + (workers − 1)` (fleet ceiling `(worker_pool − 1) × instances + workers`), and each **server** opens up to `DEFAULT_MAX_CONNECTIONS_SERVER`.

Live workers/instances come from `worker_ping` (30s window; workers ping every 5s). **Agent workers** (name prefix `ag-`) reach the API over HTTP and hold no Postgres pool, so they're **excluded** from the estimate and only reported for context; DB-connected workers use the `wk-` prefix. Servers don't ping `worker_ping`, so the recommendation assumes **one** server and exposes a per-server increment so the operator can add capacity for the rest. The recommended floor is `(worker demand + one server pool) × 1.20 + superuser-reserved`, never below **200** connections (Postgres defaults to 100; headroom for growth/bursts/psql is cheap).

## Changes

- `backend/windmill-api/src/db_health.rs`: `ConnectionPoolInfo` now returns `pg_superuser_reserved_connections` and a `sizing` object (live worker/instance counts, live agent-worker count, effective per-server/per-worker pool sizes, `database_connections_override`, estimated peak worker connections, recommended `max_connections`, per-server increment, and a human-readable message). New pure `compute_connection_sizing` helper — honoring the `DATABASE_CONNECTIONS` override and excluding agent workers — with unit tests. The fleet query filters by worker-name prefix (`wk-` vs `ag-`).
- `frontend/.../instanceSettings/DbHealth.svelte`: renders the sizing breakdown (per-server + per-worker-instance pool rows labelled `(default)` vs `(DATABASE_CONNECTIONS)`, and an "Agent workers (HTTP, excluded)" row when present) and warns (yellow) when `max_connections` is below the recommended floor.
- `backend/src/db_connect.rs`: re-export the `DEFAULT_MAX_CONNECTIONS_*` constants from `windmill_common` instead of keeping duplicate literals, so the runtime pool sizing and the health-page guidance share one source of truth.
- `.sqlx`: cached data for the new queries.

## Test plan

- [x] Backend unit tests for `compute_connection_sizing` (zero-fleet + 200 floor, single worker + floor, multi-instance sum, large fleet above floor with margin, reserved widening, DATABASE_CONNECTIONS override, **agent-worker exclusion**) — 7 passed.
- [x] Endpoint returns the new fields (verified via `/api/db_health`).
- [x] Agent-worker exclusion verified end-to-end: inserting a synthetic `ag-` `worker_ping` row bumps `live_agent_workers` but leaves `live_workers`/`estimated_worker_connections` unchanged, and the message notes the exclusion.
- [x] UI renders the guidance end-to-end (Superadmin Settings → DB Health → Overview), verified with Playwright.
- [ ] On an instance whose Postgres `max_connections` is below the recommended floor, confirm the yellow warning appears; confirm the "No live workers detected" branch renders with no active workers.

## Screenshots

db-connections-sizing


rubenfiszel:
Addressed the Codex P1 in 971f8fab14: the sizing helper now honors `DATABASE_CONNECTIONS`.

When it's set, the runtime caps every process's pool at that value, so `compute_connection_sizing` now reads it the same way `db_connect.rs` does and uses it as both the per-server and per-worker-instance pool (worker estimate = `DATABASE_CONNECTIONS × instances`). For the example given — `DATABASE_CONNECTIONS=100`, 5 instances — it now estimates 500 worker connections instead of 25, and the recommendation/undersized warning follow. The response exposes `server_pool_size`, `worker_pool_size`, and `database_connections_override`; the panel labels the pool rows `(default)` vs `(DATABASE_CONNECTIONS)`. Added a unit test for the override path.

This also picks up the earlier "`default_worker_pool_size` returned but never rendered" note — the worker-instance pool now has its own row.

PR #10020, opened 2026-07-09
feat: free AI tokens + home search/filter revamp

diegoimbert:
## Summary

Adds a **free AI tier** so users can try AI without configuring their own key, surfaces it clearly in the UI, and reworks the home page filtering around it. When no AI provider is configured (workspace **or** instance) and `DEEPSEEK_FREE_API_KEY` is set, Windmill lends its own **DeepSeek** key. Usage is metered per user as **cost in nano-dollars** (not raw tokens), with a one-time per-user grant and an instance-wide daily kill-switch. The lent key, quota accounting, and abuse controls live in the enterprise (`private`) crate; the OSS build ships inert stubs.

The companion enterprise changes are in `windmill-ee-private` (`ai_free_tier_ee.rs`); `backend/ee-repo-ref.txt` pins the matching commit.

### Backend
- **Migration** `ai_free_token_usage`: per-user **one-time lifetime grant**, keyed by normalized email (shared across a user's workspaces). `ai_free_token_daily_usage`: instance-wide daily cost ceiling — a kill-switch independent of the per-user grant.
- **`ai_free_tier_ee.rs`** (EE): env-gated `DEEPSEEK_FREE_API_KEY`; model pinned to `deepseek-chat`; default budgets **$2 one-time per user** and **$100/day instance-wide** (overridable via `FREE_TIER_USER_COST_LIMIT_NANOS` / `FREE_TIER_INSTANCE_DAILY_COST_LIMIT_NANOS`). `ai_free_tier_oss.rs` provides OSS no-op stubs (the feature is absent from OSS builds).
- **Cost, not tokens.** The copilot is agentic and DeepSeek caches the prompt, so a raw token count wildly overstates the bill. Each usage bucket is priced (cache-hit / cache-miss / output) and summed in nano-dollars, so the grant and cap are real dollars.
- **`ai.rs` proxy**: on the previously dead-end "AI resource not configured" path, falls back to the free tier for DeepSeek `chat/completions`. `get_copilot_info` returns a synthetic DeepSeek provider when the tier is available.

### Abuse controls
- **Reserve-then-reconcile metering.** A pessimistic upfront charge (full price of the max permitted output) is committed to both counters *before* the request is dispatched, then reconciled down to the real cost on clean completion. A client abort leaves the reservation standing — closing the "read the stream, disconnect before the end-of-stream usage chunk, get metered zero" dodge, which also evaded the daily kill-switch. An unconsumed request (upstream send failed) refunds the reservation.
- **One in-flight free request per user** via a per-email mutex taken before the budget check (closes the check-then-spend race).
- **Email normalization** (`+tag` / gmail-dot) so one mailbox can't mint many grants.
- **Server-side hardening**: model pinned, `max_tokens` clamped (default 16384), `stream_options.include_usage` forced so streamed responses can't withhold the usage report.

### Frontend
- `AIConfig` gains a **response-only `free_tier` marker** (`{ exhausted, used_ratio }`, `skip_deserializing` so a client can't store a forged one). It keeps flowing once the grant is spent, so the UI knows AI is off because the grant ran out, not because nothing was configured.
- **Usage meter** reusing the existing context-usage gauge: a shared `UsageMeter` powers both `ContextUsageIndicator` and a new `FreeTierUsageIndicator`. Shown in the session-chat toolbar and next to the home-chat model settings; a **"Free"** pill labels the lent model. The context bar is hidden on the free tier so the free meter takes its slot.
- **Exhausted state**: a red "you've used your free Windmill AI, add your own API key" banner in the chat thread and overlay on the home chat — instead of the generic "no provider configured" screen.
- **Live updates**: `copilotInfo` is re-fetched after every free-tier turn, so the meter advances as tokens are spent and the exhausting turn flips to the banner without a page reload. Gated to active free-tier users (no extra request for configured-key users).

## Home page

- **Workspace/Hub** is a toggle button group beside the kind toggle (no longer tabs).
- **Home AI chat** ("Build with AI") with example-prompt tags and a lean CLI/MCP + Hub row, blurred/overlaid when AI is unavailable.
- Removed the path-prefix owner tags (`u/admin`, …) and dead `ownerFilter`; removed dead keyboard arrow-navigation and the `keyboardSelected` plumbing.
- **Unified search into `FilterSearchbar`** with filters: `path`, `summary`, `kind` (the All/Script/Flow/App toggle mutates the same `kind`), `draft_only`, `draft`, `content` (EE-gated full-text view), `archived`, `include_library`, and the admin folder-scoping filter. Filters are a single URL-synced source of truth. Tree view stays a separate toggle.

## Verification

- **Backend**: OSS and EE (`--features private,enterprise`) builds green; EE unit tests pass (reconcile arithmetic, reservation, metering, email normalization, body enforcement); offline `.sqlx` build green.
- **Free-tier UI browser-verified** against a live EE backend across all states: fresh grant, running-low (>80% warn), exhausted banner/overlay, own-key workspace (no free UI), the usage meter reused from the context gauge, and the live meter refresh mid-session.
- **Abuse fix verified live**: clean completion reconciles to the exact real cost; a mid-stream abort leaves the full reservation standing (both user and daily counters); a forged `free_tier` marker POSTed to `edit_copilot_config` is stripped on write.

## Test plan
- [ ] Set `DEEPSEEK_FREE_API_KEY`, leave workspace/instance AI unconfigured → copilot shows `deepseek-chat` with a "Free" pill and usage meter; a chat records cost; the meter advances live.
- [ ] Spend the grant → chat/home show the exhausted banner + "Add your own API key"; the proxy returns the limit message.
- [ ] Disconnect mid-stream → the reservation stands (usage does not drop to zero).
- [ ] Unset the key → copilot hidden, proxy returns "AI resource not configured" (feature inert).
- [ ] Home: Workspace/Hub toggle switches views; kind toggle ↔ searchbar `kind` stay in sync; `path`/`summary`/`draft`/`archived` filters narrow the list.


diegoimbert:
P1: Global-AI gate hides unrelated Home navigation
is not true

diegoimbert:
P2 — Conditional filters still miss URL synchronization 
is wrong too, it works

PR #10024, opened 2026-07-09
fix(frontend): name the draft in AI chat test-run confirmation

Guilhem-lm:
## Summary

When the AI chat asks to confirm a test run, the confirmation card showed a static, generic header (e.g. `Run script test`). It now names the target and makes clear it runs the user's **draft** — e.g. `Run a test of your draft of my_script`.

## Changes

- `Tool.confirmationMessage` (`shared.ts`) now accepts `string | ((args) => string)`; the executor resolves the function form with the parsed call args before setting the confirmation status.
- Global chat test-run tools (`global/core.ts`) build dynamic, draft-aware headers naming the target:
  - `test_run_script` / `test_run_flow` → `Run a test of your draft of <name>` (leaf of the workspace path via a new null-safe `pathLeaf` helper)
  - `test_run_step` → `Run a test of step "<stepId>" in your draft of <name>`
- In-editor script/flow test-run tools (`script/core.ts`, `flow/core.ts`) → `Run a test of your draft` (no path available; the step variant names the step).
- `AIChatDisplay.svelte` — the auto-accept ("bypassed in current mode") tooltip rendered `confirmationMessage` directly; it now falls back to the tool name when the value is a function, so it never renders the function source.

## Test plan

- [ ] In the global AI chat, have the assistant call `test_run_script` / `test_run_flow` / `test_run_step` and confirm the confirmation-card header names the target (correct path leaf) and includes the stepId for the step variant.
- [ ] Enable auto-accept / YOLO mode and open the "bypassed in current mode" tooltip — each test-run tool shows a readable label, not arrow-function source.
- [ ] `npm run check:fast` (the single unrelated `utils_workspace_deploy.ts` error is pre-existing on this branch).

## Screenshots

No static screenshot attached: the affected surface is an ephemeral confirmation card that only renders when the live model decides to call a test-run tool, so it can't be triggered deterministically from Playwright. The change is a confirmation-header text label and was manually verified in the running app by the author.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -460,7 +460,12 @@
 			.filter((tool) => tool.requiresConfirmation === true)
 			.map((tool) => ({
 				name: tool.def.function.name,
-				label: tool.confirmationMessage ?? tool.def.function.name
+				// confirmationMessage may be a function of the call args, which we don't
+				// have here — fall back to the tool name rather than render its source.
+				label:
+					typeof tool.confirmationMessage === 'string'
+						? tool.confirmationMessage
+						: tool.def.function.name
 			}))
 	})
 	const visibleYoloBypassedTools = $derived(yoloBypassedTools.slice(0, MAX_YOLO_TOOLTIP_TOOLS))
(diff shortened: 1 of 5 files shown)


Guilhem-lm:
Addressed the Codex P2 in f48dcc8810: the global test-run tools fall back to deployed content when no draft exists, so I dropped the "your draft" claim and the confirmations now read neutrally — "Run a test of <name>" (and the in-editor variants "Run a test of the current script/flow"). Claude and Pi reviews were already good-to-merge.

PR #10027, opened 2026-07-09
fix(frontend): keep draft autosave alive when an editor shares a warm session's DraftEntry

Guilhem-lm:
## Summary

A script open in **both** a session preview and the full-page `/scripts/edit` editor (same tab, same workspace) shares a single `DraftEntry` in `userDraft.svelte.ts` — both key autosave on the identical `(workspace, kind, path)`. That entry's sync `$effect.root` was created *inside the first-acquiring component* (the session preview). When that component unmounts during navigation while the other editor still holds the entry (refcount > 0), the root **stops being scheduled** — so the surviving editor's edits mutate `entry.state.val` but never POST, and autosave silently dies.

Empirically confirmed: reassigning the cell's value after the session unmounts produced **zero** POSTs (dead effect), while a fresh entry POSTs on every keystroke.

## Changes

`frontend/src/lib/userDraft.svelte.ts`:
- **One app-lifetime `$effect.root`** now owns every entry's sync effect as a child root, keyed by a reactive `SvelteSet` of active keys. The effect's lifetime is decoupled from any component — it survives editor unmounts and only dies when the key is released at refcount 0. (Browser-only; on the vitest/SSR node runtime the manager is never installed, preserving the prior in-memory fallback.)
- The cell is created synchronously (`$state`), so handles work immediately; the sync effect attaches on the manager's next flush (a microtask — seeds don't POST, so no regression).
- Autosave config (`discardIf` / `canBeDisabled`) now lives **on the entry**, read live at POST time and applied **additively** across handles sharing a key — a later acquirer can add `discardIf` / opt into the auto-save toggle, but a sharer that omits them never strips what an earlier holder set. So a full-page editor keeps its discard/toggle behavior regardless of the order it and a session preview open (and a session sharing a key never re-enables autosave while the full-page editor's toggle is off).
- `releaseEntry` tears the child root down **synchronously** to avoid a stale-root race on a same-tick release + re-acquire during SPA nav.

## No visible UI change

This is a behavioral fix — the affected surface (the script editor / session preview) is pixel-identical whether autosave works or not; the effect is a network POST. Verified at the network level instead (below).

## Test plan

- [x] Repro (deterministic): open a script in a session preview, keep the entry alive past the session editor's unmount, mutate the value → **before**: no `POST /drafts/update`; **after**: `POST …/drafts/update` → 200 OK carrying the mutation.
- [x] Real user flow: open script in a session preview → client-nav to `/scripts/edit/<path>` → edit → autosaves (200 OK, payload contains the edit).
- [x] Common path unaffected: fresh full-page editor autosaves on edit; config correct (session `canBeDisabled=false`, full-page `true`).
- [x] `npm run check:fast` clean for this file (one unrelated pre-existing error in `utils_workspace_deploy.ts`); 195+ unit tests pass (`userDraft*`, `sessions/*`, `chatDraft`).
- [ ] Reviewer sanity: full-page editor with auto-save toggled **off**, then open a session preview of the same item → confirm keystrokes do **not** silently persist while the toggle is off.

### DIFF
--- a/frontend/src/lib/userDraft.svelte.ts
+++ b/frontend/src/lib/userDraft.svelte.ts
@@ -1,5 +1,6 @@
 import { get } from 'svelte/store'
 import { onDestroy, untrack } from 'svelte'
+import { SvelteSet } from 'svelte/reactivity'
 import { deepEqual } from 'fast-equals'
 import { workspaceStore } from './stores'
 import { readFieldsRecursively } from './utils'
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
Closing as redundant — both bugs this PR addressed have since been fixed and merged into main:

- The shared-DraftEntry autosave death (dead `$effect.root` on component unmount) is fixed by #10052, which rewrites the same region of `userDraft.svelte.ts` with a functionally equivalent entry-owned mirror root.
- The related workspace→session preview staleness (and phantom autosave) is fixed by #10061 (WIN-2160).

PR #10033, opened 2026-07-10
fix: keep agent-worker server job-completed processor alive & self-healing

rubenfiszel:
## Summary

A customer running agent workers against a server (`agent_worker_server`) hit a storm of `500`s on `/send_result` with `Failed to send job completed to background processor @ee.rs:566`, alongside zombie-job restart loops. This fixes the root cause and hardens the path so a recurrence — from **any** cause — is self-healing and never silent.

### Root cause

The server runs `RECV_PARALLELISM` background job-completed processors (`start_background_processor`). They hold the receivers on the shared bounded flume channel that `/send_result` pushes into. The processor loop had an exit path:

```rust
if is_init_script && !final_success { killpill_tx.send(); break; }  // drops the receiver
```

Correct for a real **worker** (its own init script failed → die and let the supervisor restart it), but on the **server** this processor merely *relays* completions for many remote agent workers. When an agent worker's init script fails, it POSTs the failed completion to `/send_result` (agent workers use the HTTP `JobCompletedSender`) and then exits — so the server's processor hit this path and terminated. Once all processors exited, the channel disconnected → every subsequent `send_job` returned `SendError` → `process_result` returned **500** → the completion was never persisted → the monitor flagged the job a **zombie** and restarted it → 500 again, forever. Only a full server restart recovered it. A **panic** in a processor task drops its receiver the same way.

## Changes

- **`result_processor.rs`** — new `is_agent_server` parameter on `start_background_processor`. When set, an init-script failure is logged (with job id) but does **not** killpill/break, so the server relay stays alive. Real workers (`worker.rs`) pass `false` — unchanged behavior.
- **`ee.rs` (EE) — supervisor + self-heal** (`supervise_processor`): each processor runs under a supervisor that respawns it if it exits while the server is up (covers panics and any future early-return path) and raises a `report_critical_error` alert (`resource=agent_worker_server_bg_processor`) so a death is never silent. Graceful shutdown is distinguished via the `.kill()` broadcast. The supervisor keeps a receiver alive across respawns, so the channel does not disconnect during the ~1s respawn window.
- **`ee.rs` (EE) — `process_result` waits instead of bouncing** (`classify_send`): `send_job(.., true)` blocks while the bounded channel (100 slots) is full, so a completion arriving during a respawn **buffers and waits**. The wait is capped at 15s (under the agent worker's 20s HTTP timeout); only if the whole pool stays down, or a processor is wedged past the cap, does it return an explicit retryable **503** (with job id + cause) instead of an opaque 500.

EE companion: windmill-labs/windmill-ee-private#653 (same branch).

## Test plan

- [x] `cargo check -p windmill-api-agent-workers --features private` — clean
- [x] **Unit** `supervisor_respawns_and_alerts_until_shutdown` — a processor that panics / exits early is respawned and alerted on; a shutdown broadcast stops supervision without alerting (covers the panic → respawn + alert path).
- [x] **Unit** `classify_send_maps_all_outcomes` — a successful send → `Sent` (200); a disconnected channel → `Disconnected` (503); a send that misses the deadline → `TimedOut` (503).
- [x] During development, verified end-to-end with a real agent worker + failing `init_bash`: server logged `keeping server bg processor alive` and kept processing; with the guard flipped off, logs showed `init script errored, exiting` → `CRITICAL ERROR … exited unexpectedly; respawning`, and the job still completed via the supervisor's self-heal. (That heavyweight enterprise-only e2e was dropped in favor of the focused unit tests above.)


rubenfiszel:
Addressed the P2 (alert/respawn storm on a persistent crash-loop) in `supervise_processor`:

- **Exponential capped backoff** — `base_backoff * 2^n`, capped at 60s, so an immediate crash-loop can't spin.
- **Rate-limited alerts** — `report_critical_error` now fires only for the first 3 consecutive fast failures; further respawns in the same loop are logged, not alerted.
- **Reset on recovery** — a processor that ran ≥ 30s before exiting resets both counters, so a genuine later failure alerts afresh (no permanent suppression).

Unit test `supervisor_respawns_rate_limits_alerts_and_stops_on_shutdown` updated to drive panic → clean-exit → crash-loop → shutdown and assert the alert cap.

EE change: windmill-labs/windmill-ee-private#654; picked up here via the `ee-repo-ref` bump.

rubenfiszel:
Addressed the Codex P1 (graceful shutdown can hang during the respawn backoff):

`supervise_processor` now (1) re-checks the shutdown signal before every (re)spawn, and (2) `select!`s on the shutdown broadcast during the backoff (`biased`, so a `.kill()` mid-wait wins the race) instead of sleeping the full backoff. So a kill during backoff stops the supervisor promptly and never spawns a replacement processor that would miss the broadcast and hang shutdown.

Added `supervisor_stops_promptly_on_shutdown_during_backoff`: first processor exits, shutdown is sent 200ms into a 10s backoff, asserts the supervisor returns within 3s and never respawns (fails against the old sleep-the-full-backoff code).

EE change: windmill-labs/windmill-ee-private#655; picked up here via the `ee-repo-ref` bump.

PR #10043, opened 2026-07-10
feat(mcp): add multi-workspace MCP tokens via the gateway endpoint

rubenfiszel:
## Summary

A single MCP token can now work across **every workspace the token owner can access**, instead of one token/URL per workspace. This addresses the NxM growth problem where a user with 10–20 workspaces had to register 10–20 MCP server entries in clients like Claude/Cursor.

An "all-workspaces" token is a token with **no bound workspace** (`workspace_id IS NULL`) + an `mcp:` scope, used against the existing `/api/mcp/gateway` endpoint. No schema migration is needed (the token table already allows a null workspace) and no new routes are added.

In multi-workspace mode the runner exposes a synthetic **`list_workspaces`** tool plus the generic API endpoint tools, each workspace-scoped one gaining a required **`workspace_id`** argument (mirroring the proxy pattern users built externally). Per-workspace scripts/flows are intentionally **not** enumerated across all workspaces (that would flood the tool list) — they are run via `runScriptByPath`/`runFlowByPath` with an explicit `workspace_id`.

Single-workspace tokens are completely unchanged.

Fixes WIN-2153

## Changes

- **Gateway middleware** (`mcp/core.rs` `extract_workspace_from_token`): a token bound to a workspace still injects `WorkspaceId` (single-workspace mode); a workspace-less `mcp:`-scoped token now injects `MultiWorkspaceMcp` + `McpToken` (multi-workspace mode). Also reads the token from `?token=` as well as the `Authorization` header.
- **Runner** (`windmill-mcp/src/server/runner.rs`): split `list_tools`/`call_tool` into single- vs multi-workspace paths. Multi mode exposes `list_workspaces` + endpoint tools (with injected `workspace_id`), resolves a per-workspace `ApiAuthed` from the raw token via `resolve_workspace_auth` on each call, and applies the same read-only / scope gates as single mode.
- **Backend trait** (`McpBackend`): adds `list_accessible_workspaces` (superadmins → all workspaces; others → `usr` membership) and `resolve_workspace_auth` (via `AuthCache.get_authed`, which returns `None` for non-members → "Access denied").
- **Tool schema** (`endpoints.rs`): `endpoint_tool_to_mcp_tool_multi` injects a required `workspace_id` for workspace-scoped tools and leaves global tools (docs) unchanged; `list_workspaces_tool` builder. Unit tests added.
- **Frontend** (`CreateToken.svelte`): MCP token creation gains an "All workspaces" option that produces a workspace-less token and the `/api/mcp/gateway` URL, with an inline explanatory note.

## Security boundary

Cross-workspace access is denied through the existing auth machinery: `resolve_workspace_auth` → `AuthCache.get_authed(Some(workspace_id), token)` performs the `usr` membership lookup and returns `None` for non-members, which surfaces as `invalid_params` "Access denied". Superadmins may act in any workspace (consistent between `list_accessible_workspaces` and `resolve_workspace_auth`). The `?token=` read is not a new exposure — the `ApiAuthed` extractor already reads it.

## Test plan

- [x] Backend compiles with `--features mcp`; sqlx offline cache regenerated (3 new entries, 0 EE caches lost)
- [x] Unit tests for the tool-schema transformation (`cargo test -p windmill-mcp`)
- [x] E2E against a live mcp-enabled server: create null-workspace token → `tools/list` returns `list_workspaces` + endpoint tools with `workspace_id` required (workspaced) / absent (global `searchDocs`); `list_workspaces` returns workspaces; `listScripts` with `workspace_id` resolves per-workspace auth and returns data; missing `workspace_id` → clear error; single-workspace mode unchanged (no `list_workspaces`, no injected arg)
- [x] Frontend verified in browser (Playwright): "All workspaces" option, explanatory note, generated `/api/mcp/gateway` URL
- [ ] Reviewer: confirm a read-only all-workspaces token does not expose/execute mutating endpoint tools

## Screenshots

MCP token creation with the new **"All workspaces"** option selected:

all-workspaces-option

Generated gateway URL (`/api/mcp/gateway?token=…`):

generated-gateway-url

---
## Summary by cubic
Add multi-workspace MCP tokens served via `/api/mcp/gateway`, so one token works across every workspace the user can access. Implements WIN-2153 with an “All workspaces” MCP token and gateway URL; single-workspace tokens are unchanged.

- **New Features**
  - Gateway: detects workspace-less `mcp:` tokens from `Authorization` or `?token=`, enables multi-workspace mode; bound tokens inject `workspace_id`.
  - Runner: splits single vs multi; in multi, exposes `list_workspaces` plus endpoint tools with required `workspace_id`; resolves per-workspace auth via `AuthCache`; enforces read-only and scopes.
  - Backend: adds `list_accessible_workspaces` and `resolve_workspace_auth` (superadmins → all; others → member-only).
  - Tools: workspace-scoped endpoint tools require `workspace_id`; descriptions note “Requires `workspace_id`.”; arg described as “Target workspace id (from `list_workspaces`).”
  - Frontend: Create Token adds an “All workspaces” option and generates the `/api/mcp/gateway?token=…` URL.

- **Bug Fixes**
  - Enforce script/flow path scopes for multi-workspace `runScriptByPath`/`runFlowByPath`: shown only if the token grants some script/flow scope; at call time the `path` must match; `mcp:endpoints:*` alone can’t run arbitrary scripts/flows; `mcp:favorites` tokens no longer expose or run run-by-path.
  - Preserve script/flow args for `runScriptByPath`/`runFlowByPath` when bodies use `additionalProperties`; pass-through now forwards all args not consumed by path/query params, and the runner strips the synthetic `workspace_id` before dispatch.

<sup>Written for commit 55de5963a045a6886572f11576f647676d9c51ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10043?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-67c405ff2bfd68119dbd5e2edc91fde70711b2fb8ec6826411cc7d74687d5bcb.json
+++ b/backend/.sqlx/query-67c405ff2bfd68119dbd5e2edc91fde70711b2fb8ec6826411cc7d74687d5bcb.json
@@ -1,0 +1,26 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT id, name FROM workspace WHERE deleted = false ORDER BY name",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "name",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "67c405ff2bfd68119dbd5e2edc91fde70711b2fb8ec6826411cc7d74687d5bcb"
+}
--- a/backend/.sqlx/query-a460f0ca8f23a8eb9d808b5edd6e0cde0e125f8ed426bd784dd7b92e1d21dfdf.json
+++ b/backend/.sqlx/query-a460f0ca8f23a8eb9d808b5edd6e0cde0e125f8ed426bd784dd7b92e1d21dfdf.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT workspace_id, scopes FROM token WHERE token_hash = $1 AND (expiration > NOW() OR expiration IS NULL)",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "workspace_id",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "scopes",
+        "type_info": "TextArray"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      true,
+      true
+    ]
+  },
+  "hash": "a460f0ca8f23a8eb9d808b5edd6e0cde0e125f8ed426bd784dd7b92e1d21dfdf"
+}
--- a/backend/.sqlx/query-b3b06ec52fde4b8264c6307c24b046cd3af17c5ce0d4426153b3065b2faaa781.json
+++ b/backend/.sqlx/query-b3b06ec52fde4b8264c6307c24b046cd3af17c5ce0d4426153b3065b2faaa781.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT workspace.id, workspace.name\n                 FROM workspace\n                 JOIN usr ON usr.workspace_id = workspace.id\n                 WHERE usr.email = $1 AND usr.disabled = false AND workspace.deleted = false\n                 ORDER BY workspace.name",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "name",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "b3b06ec52fde4b8264c6307c24b046cd3af17c5ce0d4426153b3065b2faaa781"
+}
--- a/backend/windmill-api/src/lib.rs
+++ b/backend/windmill-api/src/lib.rs
@@ -483,8 +483,13 @@ pub async fn run_server(
                 add_www_authenticate_header, add_www_authenticate_header_gateway,
                 extract_workspace_from_token,
             };
-            let (mcp_router, mcp_cancellation_token) =
-                setup_mcp_server(db.clone(), user_db, _base_internal_url.clone()).await?;
+            let (mcp_router, mcp_cancellation_token) = setup_mcp_server(
+                db.clone(),
+                user_db,
+                _base_internal_url.clone(),
+                auth_cache.clone(),
+            )
+            .await?;
             // Workspace-scoped MCP router
             let workspaced_mcp_router = mcp_router
                 .clone()
--- a/backend/windmill-api/src/mcp/core.rs
+++ b/backend/windmill-api/src/mcp/core.rs
@@ -313,6 +313,8 @@ impl McpBackend for WindmillBackend {
             args_map,
             &endpoint_tool.body_schema,
             &endpoint_tool.body_field_renames,
+            &endpoint_tool.path_params_schema,
+            &endpoint_tool.query_params_schema,
         );
 
         // Create and execute request
--- a/backend/windmill-api/src/mcp/core.rs
+++ b/backend/windmill-api/src/mcp/core.rs
@@ -10,10 +10,11 @@ use windmill_common::{db::UserDB, utils::StripPath, DB};
 use windmill_mcp::common::schema::enrich_resource_schemas;
 use windmill_mcp::common::transform::apply_key_transformation;
 use windmill_mcp::common::types::{
-    FlowInfo, HubScriptInfo, ResourceInfo, ResourceType, SchemaType, ScriptInfo,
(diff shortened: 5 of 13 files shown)


rubenfiszel:
Thanks for the thorough reviews. Addressed the findings in `3d96742`:

## P1 (Codex) — `runScriptByPath`/`runFlowByPath` dropped script/flow args — **fixed**

Correct and important catch. Those two endpoints have an `additionalProperties: true` body with no declared `properties`, and `build_request_body` returned `None` for schemas without `properties`, so every script/flow argument was silently dropped. This was latent for the per-path run endpoints generally and became load-bearing in multi-workspace mode (where run-by-path is the only way to execute scripts/flows).

Fix:
- `build_request_body` now forwards all arguments **not consumed by a path/query parameter** for pass-through (`additionalProperties`) bodies, while keeping the strict declared-only behavior for endpoints with explicit `properties`. Verified only `runScriptByPath` and `runFlowByPath` have pass-through bodies (both intentionally forward args), so nothing else is affected.
- The runner strips the synthetic `workspace_id` argument before dispatch, so it can't leak into the forwarded body.

Added unit tests (`build_request_body_passthrough_forwards_script_args_minus_path`, declared-only, GET-no-body) and re-ran the full e2e: an all-workspaces token calling `runScriptByPath` with `{workspace_id: admins, path: …, name: "alice", count: 21}` now produces a job whose `args` are exactly `{"count":21,"name":"alice"}` (workspace_id stripped) and result `{"doubled":42,"greeting":"hello alice"}`. ✅

## P2 (Claude) — global endpoints run with empty `workspace_id`

Confirmed harmless: only workspace-scoped tools (path contains `{workspace}`) get a `workspace_id`; the two global tools (`searchDocs`, `readDocsPage`) are read-only and their routes perform no `allowed_in_workspace` check, so the empty-workspace JWT is never consulted for a workspace decision. The only two pass-through bodies are both workspace-scoped, so no global endpoint forwards a body either.

## P2 (all) — integration test for the denial path

Agreed this is the higher-value test. The added unit tests cover the pure schema/transformation logic; the auth-denial boundary (`resolve_workspace_auth` → `AuthCache.get_authed` returning `None` for non-members) is integration-shaped (needs a second workspace + non-member user + DB). I verified it manually (member → data, missing `workspace_id` → clear error, superadmin → all workspaces) but did not add an automated integration test in this PR — noting it as a good follow-up to lock the boundary against future `get_authed` refactors.

rubenfiszel:
## P1 (Codex + Pi) — multi-workspace run-by-path bypassed script/flow scopes — **fixed** (`012f834`)

Correct and important finding. In multi-workspace mode `runScriptByPath`/`runFlowByPath` are the only way to run scripts/flows, but `call_tool_multi` authorized them against the **endpoint** scope only, never the caller's `mcp:scripts:`/`mcp:flows:` **path** scopes. That let a granular token run items outside its allowed paths.

These two run-by-path endpoints are now authorized by the **script/flow scope of the requested `path`**, matching single-workspace mode's per-item tools:

- **Exposure** (`list_tools_multi`): shown only when the token grants some script/flow — `McpScopeConfig::has_any("script"|"flow")` (true for non-granular `mcp:all`/`mcp:favorites`; for granular tokens, only if a `mcp:scripts:`/`mcp:flows:` pattern is present). This also fixes the usability gap Pi noted: a `mcp:scripts:f/team/*`-only token can now use `runScriptByPath` **without** needing a separate endpoint scope.
- **Authorization** (`call_tool_multi`): the `path` argument is checked via `scope_config.is_allowed("script"|"flow", path)` before dispatch; denied otherwise.

Result for the three scenarios in the review:
- `mcp:endpoints:*` alone → run-by-path **not exposed**, and rejected if called directly (no script scope). The arbitrary-run bypass is closed.
- `mcp:scripts:f/team/*` alone → run-by-path **exposed**, `f/team/*` allowed.
- `mcp:scripts:f/team/*` (+ any endpoint scope) → `f/team/*` allowed, `f/other/*` **denied**.

Verified end-to-end against a live gateway:
```
mcp:scripts:f/team/*  → runScriptByPath exposed=True, runFlowByPath exposed=False
   run f/team/allowed_script  → job started
   run f/secret/blocked_script → "Access denied: script 'f/secret/blocked_script' not in token scope"
mcp:endpoints:* only  → runScriptByPath exposed=False; direct call → Access denied
```

Added unit tests for `McpScopeConfig::has_any`. Scope of the change is deliberately limited to multi-workspace mode — single-workspace behavior (where the endpoint scope governs the generic run-by-path escape hatch and per-item tools carry their own path scope) is unchanged to avoid regressing existing tokens.

rubenfiszel:
## P1 (Codex) — `mcp:favorites` all-workspaces token could run any script/flow by path — **fixed** (`55de596`)

Correct catch, and a subtle one. `mcp:favorites` sets `granular = false`, so the run-by-path check I added in `012f834` (`if scope_config.granular && !is_allowed(...)`) was skipped entirely for favorites tokens — turning the default "Favorites only" all-workspaces token into a run-anything token via `runScriptByPath`/`runFlowByPath`.

Root cause: `has_any` and the call-time gate treated `mcp:favorites` like `mcp:all`. But favorites are an enumerated set reachable only through per-item tools, never by naming an arbitrary path — and `is_allowed()` already returns `false` for a favorites token. Fix:

- `has_any` now returns `true` only for `mcp:all` (not for `!granular` generally), so favorites tokens no longer advertise run-by-path.
- The call-time check drops the `granular` gate and relies on `is_allowed(kind, path)` directly, which correctly encodes every mode: `true` for `mcp:all`, pattern-matched for granular scopes, `false` for `mcp:favorites`.

Verified e2e:
```
mcp:favorites  → runScriptByPath exposed=False; direct call → "Access denied: script '…' not in token scope"
mcp:all        → runScriptByPath still runs (regression check)
mcp:scripts:f/team/*  → f/team/* runs, f/other/* denied (unchanged)
```

Updated `test_has_any` to assert favorites grants no arbitrary-path access. All 58 windmill-mcp lib tests pass.

PR #10045, opened 2026-07-10
fix(frontend): open current editor item in session when toggling to AI Sessions

rubenfiszel:
## Summary

Follow-up to #10044 (WIN-2154). While verifying that PR I confirmed a second,
related issue: **switching a brand-new flow/script/app into an AI session via the
sidebar Workspace↔AI Sessions toggle drops you in an empty session** — the item
you were editing is nowhere in the preview.

Root cause: there are two entry points into a session and they behaved
differently:

| Entry point | Function | Seeds the preview with the current item? |
|---|---|---|
| "Open in AI session" (icon in the editor toolbar) | `openEditorInSession` | ✅ yes |
| **Workspace↔AI Sessions sidebar toggle** | `enterSessionMode` | ❌ no — restores the last/new session, never looks at the editor you're on |

So a freshly-built flow (which has no pre-existing session) lands you in an empty
session when you use the toggle.

## Fix

Editors now **publish** their "Open in AI session" source to a small module
registry; the toggle **reads** it and opens that item in the preview — the same
handoff the in-editor button already performs — falling back to the previous
generic restore when you're not on an editor.

- `frontend/src/lib/components/sessions/editorSessionSource.ts` (new) — a plain
  module ref with `registerEditorSessionSource(source)` (returns a cleanup that
  clears the entry, identity-guarded) and `getEditorSessionSource()`.
- `ScriptBuilder`, `FlowBuilder`, `RawAppEditorHeader` — each publishes its
  existing `OpenInSessionSource` via `$effect(() => registerEditorSessionSource(src))`,
  gated on `!inSessionPane` so the session preview itself never registers (it
  would otherwise re-seed from itself). The source object was hoisted to a
  `$derived` and reused by the inline button, so there's a single source of truth.
- `enterSessionMode({ seedFromEditor })` — when set (passed by `SessionModeSwitch`),
  runs the source's `beforeOpen` (persist + materialize the draft) then
  `openEditorInSession(target, workspaceId)`. The sessions-page reconcile caller
  (`{ replace: true }`) does **not** set it, so its behavior is unchanged.

## Verification

End-to-end in a real browser (system Chromium via Playwright, workspace AI
configured with a placeholder provider so the session actually opens):

1. New flow → add a step + summary `TOGGLEMARK` → click the **AI Sessions**
   toggle → the preview shows the flow (step node + friendly path). ✅
   (Before the fix: empty session, `TOGGLEMARK` absent.)
2. Regression — visit a flow editor, navigate away to `/runs`, then toggle
   AI Sessions → the flow does **not** leak into the preview; generic restored
   session opens. ✅ (Confirms the `$effect` cleanup clears the registry.)
3. The "Open in AI session" toolbar button still opens the item in the preview
   (source refactor didn't regress it). ✅

`sessionSwitch.test.ts` adds three cases (seed / no-source fallback /
reconcile-ignores-source) — 6 tests pass. `check:fast` clean for the changed
files (the pre-existing unrelated `utils_workspace_deploy.ts` error is on main).
svelte-autofixer clean.

## Screenshots

AI session preview after clicking the **AI Sessions** toggle from an edited new
flow — the flow (step node `TOGGLEMARK`, path `u/admin/beautiful_flow`) is in the
preview instead of an empty session:

toggle-seeds-flow

## Test plan

- [ ] New (unsaved) flow / script / raw-app → flip Workspace→AI Sessions toggle:
      the just-built item appears in the session preview (draft persisted first).
- [ ] Toggle into AI Sessions from a non-editor page (home/workspace view):
      still lands on the generic restored session (fallback branch).

---
## Summary by cubic
Fixes the AI Sessions toggle so it opens the item you’re editing (flow/script/raw-app) instead of an empty session. Matches the in-editor “Open in AI session” behavior and addresses WIN-2154.

- **Bug Fixes**
  - Added a small registry (`editorSessionSource.ts`) to track the current editor’s session source via `registerEditorSessionSource` and `getEditorSessionSource`.
  - Editors publish their source on mount and clean up on unmount; guarded by the `aiChatManager` context to avoid registering the session pane itself.
  - The toggle now calls `enterSessionMode({ seedFromEditor: true })`, which runs `beforeOpen` to persist/materialize drafts, then `openEditorInSession`. Falls back to the previous restore when no source is present or during reconcile.
  - Builders reuse the same source for their inline “Open in session” button to keep a single source of truth.
  - Added tests for seed-from-editor, fallback, and reconcile behavior.

<sup>Written for commit e9f4e488157234b65f5377c3302ed7dc54b08cc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10045?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/FlowBuilder.svelte
+++ b/frontend/src/lib/components/FlowBuilder.svelte
@@ -34,7 +34,8 @@
 	import DeployOverrideConfirmationModal from '$lib/components/common/confirmationModal/DeployOverrideConfirmationModal.svelte'
 	import AIChangesWarningModal from '$lib/components/copilot/chat/flow/AIChangesWarningModal.svelte'
 
-	import { createRawSnippet, setContext, untrack } from 'svelte'
+	import { createRawSnippet, getContext, setContext, untrack } from 'svelte'
+	import { registerEditorSessionSource } from '$lib/components/sessions/editorSessionSource'
 	import { writable } from 'svelte/store'
 	import CenteredPage from './CenteredPage.svelte'
 	import { Button } from './common'
@@ -527,6 +528,26 @@
 	// (which for a new flow differs from the live-edited friendly `$pathStore`),
 	// falling back to `$pathStore` in drawer mounts that carry no storage path.
 	const sessionTargetPath = $derived(liveEditorDraftStoragePath || $pathStore)
+
+	// True when this builder is the session preview itself (SessionEditorTarget
+	// injects the `aiChatManager` context) — it must not publish itself as the
+	// editor to open in a session.
+	const inSessionPane = !!getContext('aiChatManager')
+
+	// "Open in AI session" source for this flow. Shared by the inline graph button
+	// and published to the sidebar Workspace↔AI Sessions toggle via
+	// `registerEditorSessionSource`, so switching into session mode from here opens
+	// this flow in the preview instead of an empty session.
+	const editorSessionSource = $derived(
+		sessionTargetPath && !inSessionPane
+			? {
+					target: { kind: 'flow' as const, path: sessionTargetPath },
+					workspaceId: opWorkspace ?? undefined,
+					beforeOpen: persistDraftForSession
+				}
+			: undefined
+	)
+	$effect(() => registerEditorSessionSource(editorSessionSource))
 
 	$effect(() => {
 		if (liveEditorDraftStoragePath === undefined || !opWorkspace) return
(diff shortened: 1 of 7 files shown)


rubenfiszel:
Closing — this fixes the wrong thing. The actual scenario is the return trip (flow editor → Open in AI session → modify → back to Workspace shows an empty flow), which is the WIN-2154 `?new_draft` reset addressed by #10044, not a sidebar-toggle seeding feature.

PR #10046, opened 2026-07-10
feat(ai-sessions): CRUD markdown artifacts in sessions

AlexRV12:
## Summary

Adds **Artifacts** to the AI copilot: markdown documents the copilot creates via chat tools, persisted locally and surfaced in the sessions UI. This gives the model a first-class place to put longer structured output — a plan, a design write-up, a spec — that the user will revisit and iterate on, instead of burying it in a long inline reply.

The feature is **sessions-only**: the tools and the prompt guidance are gated to session chats and never appear in the standalone global side-panel chat.

Delivered as **nine** reviewable commits — the artifact stack built layer-per-layer (persistence → reactive store → tools/lifecycle → viewer → preview panel + chat surface → prompt → evals → same-session-resync fix), then a **composer status-line unification** that folds the artifacts list into a single Edits · Artifacts · Jobs status line above the input.

## Changes

- **Persistence** — IndexedDB store (`copilot-artifacts`, user-scoped by email), scoped by `sessionId` so artifacts follow a session across chat rotation. Writes degrade gracefully (a failed persist still leaves the artifact usable in-memory for the session), mirroring the attached-files store. GC'd on every session-teardown path (delete session, delete workspace sessions, lifecycle reconcile).
- **Reactive store** — `SessionArtifactsStore` owned by `AIChatManager`, with a `#seq` guard so a late DB read can't clobber an optimistic write.
- **Tools** — `create_artifact` / `update_artifact` / `list_artifacts` / `read_artifact`, gated on an active session; write tools return `{ success, id, name }` and never echo back content the model just authored.
- **Preview panel & viewer** — artifacts render in the session **preview panel** via a synthetic `artifact:` URL scheme (dedup-by-id, so an update re-points the existing tab instead of opening a second; the header's "Open in workspace" link is hidden for artifact tabs, which have no workspace page). The viewer header shows the artifact name + a document icon, a **Copy / Download-as-`.md`** split-button, and an icon-only sliding **Preview / View source** toggle. Re-opening an artifact that is already the displayed tab **flashes a focus ring** on the preview so the eye finds it — and no-ops when the artifact is a background or collapsed tab, where the switch/expand is itself the visible change.
- **Session composer status line** — the composer's three separate rows (an Artifacts box, an Edits + Jobs bar) are collapsed into one bordered `bg-surface-tertiary` card of quiet **upward-popover tokens** — `Edits · Artifacts · Jobs` — each opening a popover above the input that lists its items; opening one closes the others.
  - **Edits** is a status-colored pill (worst-wins: **indigo** when any draft is pending, **green** when everything is deployed). Its popover rows jump into the existing session diff drawer **focused and scrolled onto** the clicked item (raw-app rows included, which re-key to their folder on load).
  - **Artifacts** (`ArtifactsSegment`, replacing the removed `ArtifactsBar`) lists each artifact — name, `MD` kind badge, relative "updated" time (absolute on hover) — with per-row **open / download / delete**, where deleting also closes any open preview tab; capped at ~5 rows with internal scroll.
  - **Jobs** shows a live status dot + a self-describing readout (`N running` / `N queued` / `N approval`, and terminal `N failed` / `N succeeded` / `N canceled`); its rows are fully clickable to open the run in the preview pane, with **Approve / Cancel** as inline actions.
  - A shared presentational `SessionStatusToken` keeps the three triggers visually uniform.
- **Prompt** — one session-prompt bullet tells the copilot when to reach for artifacts and how to revise (list → read → update, never duplicate). Tool descriptions are kept to bare mechanics so the guidance isn't duplicated on every iteration.
- **Evals** — new `global-artifact-plan-create` case; the harness now gates session-preview tools on `sessionChat` exactly as production's `globalToolsFor` does, so the global-mode A/B is faithful.

The `kind: 'md' | 'html'` field and the `canPreview` / `artifactFilename` / `artifactMimeType` branches are an intentional forward-compat seam for HTML artifacts (not yet wired). The current markdown renderer omits `rehype-raw`, so AI-authored raw HTML is inert (no XSS); HTML support will add real sanitization when it lands.

## Test plan

- [ ] **Unit** — `npx vitest run src/lib/components/copilot/chat/artifacts src/lib/components/sessions` (238 tests: persistence, store race-guard, tools, preview router/tabs incl. `pulseFocus`, session GC). _Note: the `setSessionPreviewSize` persistence case can occasionally flake under parallel fake-IndexedDB isolation — it passes in isolation and on re-run._
- [ ] **Create → preview** — in a session chat, ask the copilot to draft a markdown plan; confirm it calls `create_artifact`, the artifact auto-opens in the preview panel rendering markdown, and it appears as the Artifacts token.
- [ ] **Revise** — ask it to revise the plan; confirm `update_artifact` re-renders the same preview tab (no duplicate tab / no second artifact).
- [ ] **Viewer controls** — toggle the sliding Preview / View source control; use the Copy button (copies raw markdown) and its chevron → **Download as .md**.
- [ ] **Focus flash** — re-open an already-**displayed** artifact and confirm the preview flashes an accent ring (~0.8s) in both Preview and View source; confirm it does **not** flash when the artifact is a background tab or the panel was collapsed (there the switch/expand is the visible change).
- [ ] **Status line** — with edits + artifacts + jobs present, confirm the single card shows `Edits · Artifacts · Jobs`, that opening any one popover closes the others, and that the Edits pill is indigo while a draft is pending and green once all are deployed.
- [ ] **Edits → drawer** — click an Edits popover row and confirm the diff drawer opens scrolled to and highlighting that item (test both a script/flow row and a raw-app row).
- [ ] **Jobs row** — click a Jobs popover row and confirm the run opens in the preview pane and the popover closes; Approve / Cancel act without opening the run.
- [ ] **Open in workspace** — confirm the panel-header link is hidden on an artifact tab but present on a normal preview tab (script/runs/etc.).
- [ ] **Delete** — delete an artifact from the Artifacts popover; confirm the open preview tab closes and the token disappears when the last one is removed.
- [ ] **Persistence & GC** — reload the session and confirm the artifact restores; delete the session and confirm its artifacts are gone from IndexedDB.
- [ ] **Gating** — in the standalone global side-panel chat, confirm the artifact tools are not offered.
- [ ] **Eval** — `cd ai_evals && bun run cli -- run global global-artifact-plan-create --model sonnet` passes (judge 100, `create_artifact` used, no stray drafts).

## Screenshots

Screenshot 2026-07-15 at 07 51 18

Screenshot 2026-07-15 at 08 07 54


AlexRV12:
> ## Codex Review
> 1. [P1] Existing navigation evals can no longer invoke `open_page` — `ai_evals/adapters/frontend/core/global/globalEvalRunner.ts:293`
>    `globalToolsFor({ sessionPreview: false })` removes every session-preview tool, including `open_page`. However, the six existing `global-openpage*` cases default `sessionChat` to false while requiring `open_page`, so they are now guaranteed to fail. Mark those cases `sessionChat: true`, as done for the close-page and artifact cases.

Wrong. `open_page` is a global tool. Also, verified by running the 6 tests using `open_page`

PR #10053, opened 2026-07-11
fix: clearer errors on auto-draft save failure (WIN-2157)

rubenfiszel:
Fixes WIN-2157

## Problem
When the auto-draft save (the cloud icon in editor headers) fails, the user had little insight into *why*. The most common cause is that they can't write a draft at the target path — because the path is malformed or they lack permission — but the backend returned a vague `you don't have write permission on {path}`, and there was no hover affordance to see the reason.

## Changes

**Frontend** (`AutosaveIndicator.svelte`)
- The failed cloud icon now exposes the backend reason on **hover** via a native `title` tooltip, in addition to the existing click-to-open popover. The issue asked for "something to hover/click" — both are now covered.

**Backend**
- `require_can_write_path` (`drafts.rs`): the failure now distinguishes a **malformed path** (unrecognized `u/`/`f/`/`g/` prefix → `BadRequest` with a clear "a valid path starts with…" message) from a genuine **permission denial**, and the deny message now spells out *where* the user is allowed to write (own `u/<user>/` namespace, a writable folder, or a group).
- `require_owner_of_path` (`windmill-api-auth`): fixed a potential **out-of-bounds panic** on a malformed single-segment path (e.g. a bare `u`/`f`) — it now returns a clear `BadRequest` instead of a 500. Covered by a new regression test.

## Verification

**Unit / compile**
- `cargo test -p windmill-api-auth require_owner_of_path` — passes (new regression test).
- `cargo check -p windmill-api` — compiles cleanly.
- `svelte-check` on the changed component — no new errors.

**Full end-to-end against a real EE backend + real non-admin user** (created the `member` non-admin user in a regular workspace — user creation is EE-only — and called `POST /drafts/update` directly):

| Scenario | Path | Result |
|---|---|---|
| No permission | `u/otheruser/foo` | **401** `Not authorized: You don't have write permission on 'u/otheruser/foo'. It must be in your own 'u/member/' namespace, or in a folder ('f/<folder>/') or group ('g/<group>/') you can write to.` |
| Malformed prefix | `weirdpath/foo` | **400** `Bad request: Invalid path 'weirdpath/foo': a valid path starts with 'u/<user>/', 'f/<folder>/' or 'g/<group>/'` |
| Malformed single-segment (panic fix) | bare `u` / `f` | **400** clean — no 500/panic |
| Control (legitimate) | `u/member/ok` | **200** saved |

The bare `u`/`f` case is the key proof of the panic fix: previously `require_owner_of_path` panicked on the `splitted[1]` index (HTTP 500); it now returns a clean 400.

**Frontend display** (Playwright against the running editor): forced the failing `/drafts/update/` response carrying the exact backend message and confirmed the indicator shows the sticky "Save failed" label, the hover tooltip with the full reason, and the popover contents.

autosave-failure-popover

## CI review
Addressed the shared P2 nit (Codex / Claude / Pi): trimmed the narrative comment in `drafts.rs` to a concise invariant per AGENTS.md ("comments record constraints, not narration", ≤4 lines).

### DIFF
--- a/backend/windmill-api-auth/src/lib.rs
+++ b/backend/windmill-api-auth/src/lib.rs
@@ -541,6 +541,16 @@ pub fn require_owner_of_path(authed: &ApiAuthed, path: &str) -> Result<()> {
     }
     if !path.is_empty() {
         let splitted = path.split("/").collect::<Vec<&str>>();
+        // A valid path is at least `<kind>/<name>` (e.g. `u/alice/...`,
+        // `f/folder/...`). Guard the `splitted[1]` accesses below so a
+        // malformed single-segment path returns a clear error instead of
+        // panicking with an out-of-bounds index.
+        if splitted.len() < 2 {
+            return Err(Error::BadRequest(format!(
+                "Invalid path '{}': a valid path starts with 'u/<user>/' or 'f/<folder>/'",
+                path
+            )));
+        }
         if splitted[0] == "u" {
             if splitted[1] == authed.username {
                 Ok(())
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Addressed the latest Codex review:

- **[P1] Malformed stored draft could 400 the whole `/drafts/list`** — `require_can_write_path` now returns `BadRequest` for a malformed path, and `list_drafts` was propagating it. Fixed to treat `BadRequest` like `NotAuthorized` there (the row is simply not writable), so one bad row can't fail the listing. Verified end-to-end on an EE backend as a non-admin user: with two malformed stored rows (`u`, `weirdpath/foo`) present, `GET /drafts/list` returns **200** with `can_write=false` for the malformed rows and `can_write=true` for a legit `u/member/ok` row.
- **[P2] Comment nits** — trimmed the 5-line narrative comment in `drafts.rs` to a 3-line invariant, and removed the "unchanged"/"still" drafting-history phrasing from the regression-test comments (AGENTS.md).

`cargo test -p windmill-api-auth require_owner_of_path` passes; `cargo check -p windmill-api` clean.

PR #10055, opened 2026-07-11
feat: data-pipeline recorder, interactive player, and deploy-to-hub recording (WIN-2156)

rubenfiszel:
## Summary

Adds a **recorder** and **interactive player** for data pipelines, mirroring the existing flow/script recorders (`frontend/src/lib/components/recording/`). You arm **Record** on a pipeline, run it, and the cascade is captured into a downloadable JSON that the `/replay` player can rerun **fully offline** with the interactivity of a normal session — including the recorded data samples of each DuckLake/datatable asset.

Fixes WIN-2156

## Demo

Replaying a recorded run of a dbt-style DuckLake orders-analytics pipeline (staging → dim/enriched → fact → mart, with SCD2, macros, column lineage and data tests) — the graph animates node-by-node, then each node is clicked to inspect its recorded logs/result and per-asset data sample, entirely offline:

pipeline replay demo

## Why pipelines needed their own recorder

A flow/script run is a single root job that streams sub-jobs over SSE, so those recorders capture one job tree. A **pipeline run is a cascade of independent script jobs** launched client-side (`cascadeOrchestrator`/`runSelection`) and polled to completion — there is no single root job and per-node state is reconstructed from job listings. So the pipeline recording captures:

1. **The resolved asset graph** (`AssetGraphResponse`) — rendered read-only by the player.
2. **The per-node cascade status timeline** — the `Map<path, CascadeNodeState>` snapshots emitted by the orchestrator's `onUpdate`, each timestamped. Replaying these reproduces the graph animation (nodes lighting up → green/red).
3. **Each node's job stream** — opened via `getupdate_sse` on launch (log-offset dedup, same pattern as the flow recorder's `watchSubJob`), stored in the shared `RecordedJob` shape.
4. **Per-asset data samples** — after the run, each ducklake/datatable asset is sampled (up to 100 rows + typed columns + row count) via the exact live-preview query path (`loadAllTablesMetaData` + `getRows`), so a replayed sample matches what the asset-detail pane would show.

The player renders the graph read-only, animates the recorded transitions in real time, and lets you click any node to inspect it offline — script nodes show their recorded **args, logs and result** (reusing the `JobLoader` replay path with `isReplay` gating, so **no network calls**), and asset nodes show their recorded **data sample** as a typed grid.

## Changes

- `recording/types.ts` — `PipelineRecording`, `PipelineTimelineFrame`, `RecordedNodeState`, `PipelineAssetSample` (reuses `RecordedJob`/`RecordedEvent`).
- `recording/pipelineRecording.svelte.ts` — `createPipelineRecording()` store: `start`, `recordStatuses`, `watchJob`, `addCompletedJob`, `recordAssetSample`, `stop`, `toReplayData`, `download`.
- `recording/pipelineAssetSample.ts` — `capturePipelineAssetSample()`: reuses the live ducklake/datatable preview query path; never throws (errors captured per asset).
- `recording/PipelineRecordingReplay.svelte` — the player: read-only `AssetGraphCanvas` timeline animation + per-node detail (script logs/result via `JobLoader` replay; asset data-sample grid).
- `routes/(root)/(logged)/replay/+page.svelte` — dispatch `type === 'pipeline'`.
- `routes/(root)/(logged)/pipeline/[folder]/+page.svelte` — **Record** toggle + **Download recording**; capture the whole-pipeline / bounded-cascade run and sample each asset in `finalizePipelineRecording`.

## Test plan

Verified end-to-end in the browser (Playwright), both directions of the loop:

- [x] **Player** (synthetic 3-node recording): renders the DAG, `Play` animates node transitions, clicking a script node shows recorded args/logs/result offline.
- [x] **Recorder → player, CE bash pipeline**: seeded 2 bash pipeline scripts, armed Record, ran the pipeline, downloaded the recording, reloaded it in `/replay` — the producer node replays its real logs (`step producer running`, `wrote 100 rows`) and result (`{"rows":100}`).
- [x] **Recorder → player, EE DuckLake + local filesystem storage**: built an EE backend (`enterprise,license,quickjs,duckdb,parquet,private` + FFI), configured `FilesystemStorage` LFS + a DuckLake catalog, deployed a 2-step ducklake pipeline (`build_orders` → `orders` → `orders_summary`), armed Record and ran it. The recording captured **real DuckLake data samples** (orders: 4 rows with `id·INTEGER, customer·VARCHAR, amount·DECIMAL(5,2), order_date·DATE`; orders_summary: 3 aggregated rows). Reloaded in `/replay` — clicking each asset node shows its typed data grid offline.
- [x] `svelte-check` clean on all changed files (one pre-existing unrelated error in `utils_workspace_deploy.ts`); `prettier --check` clean.

## Notes

- Timeline is replayed at real time (parity with the flow/script players).
- Asset sampling covers ducklake/datatable (both go through the same preview query path); s3object files use a different preview and are not sampled.

## Screenshots

player-loaded

_Player with a loaded recording — the pipeline DAG with recorded per-node success badges._

player-node-detail

_Clicking a script node shows its recorded arguments, logs and result — replayed offline._

player-ducklake-asset-sample

_Clicking a DuckLake asset node shows its recorded data sample (typed columns + rows), captured from a real EE run with local filesystem storage._

---

## Deploy-to-Hub integration (added on top of #9332)

After merging `main` (which landed #9332 "deploy projects to hub"), this extends that recording mechanism so a **data-pipeline project** gets **one project-level recording** — the whole folder cascade — previewable with the interactive player and savable to the Hub, mirroring the existing per-item script/flow recordings.

### Changes
- **`recording/pipelineRecording.svelte.ts`** — extracted a shared `finalizePipelineRecording(store, ws)` (was inline in the pipeline editor page) and added `capturePipelineRecording(...)`: a self-contained `start → runBoundedCascade → finalize` used where no editor page orchestrates the run. The pipeline editor now reuses the shared finalize (behavior-preserving).
- **`workspaceSettings/deployToHubSession.svelte.ts`** — detects a pipeline project from the folder's asset graph (`isPipelineProject`/`pipelineScriptPaths`), `runPipelineRecording` (deployed-only cascade, `_wmill_skip_asset_dispatch`, poll to terminal at the editor's 30-min cascade timeout), `savePipelineRecording` (project-scoped Hub POST). Warns on dependency cycles for parity with the editor.
- **`workspaceSettings/DeployToHub.svelte`** — draft-phase "Data pipeline recording" card: Record / Preview (drawer rendering the real `PipelineRecordingReplay`) / Save.
- **`backend/windmill-api/src/hub_publish.rs` + `openapi.yaml`** — `POST /w/{ws}/hub/projects/{slug}/pipeline_recording` proxy (opaque forward to the Hub, admin + `{workspace}:{folder}` source authz inherited from `HubPublishCtx`, `ProjectSlug` validated by construction).
- **`frontend/package.json`** — exports `PipelineRecordingReplay.svelte` from `@windmill-labs/components` so `windmillhub` can import the player.

### windmillhub follow-up (separate PR, documented not built)
Storage is per-project (a pipeline is the whole folder): add `project.pipeline_recording JSONB`, an ingest `POST /projects/{slug}/pipeline_recording`, a `PipelineRecordingView.svelte` (mirrors `FlowRecordingView`, lazy-imports the newly-exported player) rendered on the project overview page, and bump `@windmill-labs/components`.

### Test plan (deploy-to-hub extension)
- [x] Refactored pipeline-editor recorder still works end-to-end (seeded 2 `# pipeline` bash scripts, Record → run → Download; recording JSON has `type=pipeline`, 2 runnables, 7 timeline frames, 2 jobs, code for both steps).
- [x] Exported `PipelineRecordingReplay` renders via `/pipeline_replay`: graph + per-node Output (args/logs/result) and Code tab, offline, 0 console errors. Also fixes a merge regression (`ToggleButton size="xs"`→`"sm"`).
- [x] Backend proxy compiles and the server runs healthy; `npm run check` clean (0 errors).
- [ ] **Hub-gated, not E2E-able locally**: the deploy-to-hub record card only shows in the `draft` phase (needs a configured Hub to create the project draft), and the full Save round-trip needs the windmillhub ingest endpoint above. Verified the record→preview path via the pipeline editor + `/pipeline_replay` (same shared capture helper + player).

### Player (rendered in the Preview drawer and by the Hub)

player loaded

player node detail

player code tab


rubenfiszel:
Addressed the review round (`daac97242e`):

**Codex P1s — all fixed:**
1. **Cross-run finalization race** — the pipeline editor now holds `cascadeRunningRoot` until `finalizePipelineRecording` completes, so a second run's `start()` can't reset the recorder store's maps mid-write.
2. **Stale node-detail replay timers** — `JobLoader.watchJob` now clears any previous replay's pending `replayTimeouts` before scheduling the new job's, so switching nodes on the reused loader can't let job A's events mutate job B. (Fixes it for the flow/script players too.) Verified via Playwright: switching step2→step1 shows only step1's logs/result.
3. **Removed `/replay` route** — added `frontend/.../replay/+page.ts` that 307-redirects to `/pipeline_replay` (preserving the query string), so existing links/bookmarks resolve. Verified the redirect.
4. **Unbounded asset sampling** — finalize now samples assets with a concurrency cap (`ASSET_SAMPLE_CONCURRENCY = 4`) instead of `Promise.all` over all assets, bounding peak worker load on wide pipelines.
5. **Incomplete validation** — `/pipeline_replay` upload now requires `jobs` (in addition to `graph`/`timeline`) before casting, so a malformed payload hits the invalid-format toast instead of crashing on `Object.keys(recording.jobs)`.

**Claude nits:**
- **Dead `toReplayData`** — removed from the pipeline store + interface (the player uses `setActiveReplay` directly).
- **Raw `<button>` Record toggle / `<table>` sample grid** — left as-is: both are pre-existing recorder/player code (not part of this deploy-to-hub extension); the Record toggle carries an intentional red "armed" state, and there's no design-system grid component for an arbitrary recorded-data table (Claude noted the table is defensible).
- **Data-exposure awareness** — acknowledged; by design and gated behind an admin explicitly saving, consistent with existing flow/script recordings shipping logs+results to the Hub. A redaction hook for asset samples is noted as possible future work.

**Test coverage** — the backend handler is a pure pass-through proxy mirroring the 5 sibling recording proxies (all untested); no branching beyond `ProjectSlug`/`HubPublishCtx` extraction, which the siblings already cover. Both reviewers agreed no new test is warranted.

rubenfiszel:
Addressed round 2 (`b388a09e48`):

- **Codex P1 — incomplete shape validation** — `/pipeline_replay` now type-checks the shapes the player relies on: `graph`/`jobs` must be non-array objects and `timeline` must be an array. `{"graph":{},"timeline":{},"jobs":{}}` now hits the invalid-format toast instead of crashing on `recording.timeline` iteration.
- **Claude nit — run-guard leak if finalize rejects** — the editor's guard-clear is now in a nested `finally`, so `cascadeRunningRoot` is always cleared even if `finalizePipelineRecording` ever rejects (keeps the P1 race fix, drops the wedge risk).
- **Claude/Pi nit — EventSource leak on the deploy-to-hub cascade-throw path** — `capturePipelineRecording` now `store.stop()`s (closing the per-node SSE streams) if `runBoundedCascade` throws before finalize runs.

`npm run check` clean (0 errors).

rubenfiszel:
Addressed round 3 (`ff1023f554`):

- **Codex P1 — incomplete nested validation** — the `/pipeline_replay` guard now validates the nested shapes the canvas/stepper iterate, not just top-level containers: `graph.runnables`, `graph.assets`, `graph.edges` must be arrays, and every `timeline` frame must have a `statuses` object. Verified: `graph:{}` and `timeline:[{}]` are now rejected to the invalid-format toast, while a real recording still loads. This matters for the `?src=<url>` remote-fetch path (caller-controlled input).

`npm run check` clean (0 errors). Claude and Pi were already "Good to merge" on the prior head.

rubenfiszel:
Addressed round 4 (`039a4beeb2`):

1. **Folder-wide recording ignored bundle selection** — the recorder now runs only `recordablePipelineScriptPaths` = pipeline members ∩ the project's draft membership (`items`), so a deselected member is neither executed nor embedded (code/logs/samples) in the recording. The step count in the card reflects the scoped set.
2. **Cyclic pipelines saved as successful** — any dropped cyclic member now marks the run `failed` (with the cycle listed) instead of `success`, so an incomplete/empty-schedule pipeline can't be saved.
3. **Unvalidated nested elements** — `/pipeline_replay` validation now checks array *elements*, not just containers: `graph.runnables`/`assets`/`edges` must be arrays of objects, each timeline frame's `statuses` values must be objects, and each `jobs` value must be an object. `runnables:[null]` and `statuses:{a:null}` now hit the invalid-format toast; a real recording still loads.

`npm run check` clean (0 errors). Claude and Pi were "Good to merge" on the prior head.

rubenfiszel:
Addressed round 5 (`535bb64125`):

- **Asset samples / graph not scoped to bundle membership** (the unfixed half of Codex #1, re-flagged by Claude) — added `pruneGraphToScripts`: before recording, the folder graph is pruned to the project's included members. Kept assets are only those an included script touches (an excluded branch's ducklake/datatable **table samples** are dropped), and edges/triggers only those anchored on an included runnable. The pruned graph is what's run, rendered by the player, and sampled — so a deselected branch's nodes, code, logs/results, and data samples never enter the recording. Verified the prune drops the excluded branch's runnable/asset/edge/trigger while keeping the included branch.

Note: last round had no Codex verdict (its run didn't complete) — this is a fresh head, so this round re-runs all reviewers. `npm run check` clean (0 errors).

rubenfiszel:
Addressed round 6 (`$(git rev-parse HEAD)`):

- **Validation still accepted crashing nested shapes** (Codex + Pi) — the `/pipeline_replay` guard now validates every shape the player/canvas dereference: `graph.triggers` must be an array of objects (crashes `computeMutedReadKeys` otherwise), optional `macro_edges`/`test_edges` must be arrays when present, and each `jobs` value must have an object `initial_job` and an array `events` (crashes `rebaseJobTimestamps`/`openNodeDetail` otherwise). Verified: `triggers:{}`, a job without `events`, and non-array `events` all hit the invalid-format toast; a real recording still loads.

`npm run check` clean (0 errors).

rubenfiszel:
Addressed round 7 (`a391f562ca8fe52bc7954db1b31b87b1dff8e631`): validation now covers the full recording structure the player/canvas dereference — a non-object payload (`null`/array/scalar) is rejected before `.version` is read (for all recording types), each job `events` element must be an object (`events:[null]` crashed `openNodeDetail`), and the optional `assetSamples`/`codes` maps are validated (a non-error sample needs array `rows`/`columns`; codes need string `content`/`language`). Verified: `null`, `events:[null]`, and `assetSamples:{a:{rows:null}}` all toast; valid error-only samples and a real recording still load. `npm run check` clean (0 errors).

rubenfiszel:
Addressed round 8 (`bf96773ed9e843dc4bc6a5e89cdd7d4df37b4049`):

- **Malformed graph/sample values** — the load-time guard now rejects `triggers:[{}]` (each trigger must carry a string `trigger_kind`, else `TriggerNode` crashes) and a falsey `error:''` asset sample (a non-error sample must have array `rows`/`columns`). Verified both toast; valid triggers/error-samples and a real recording still load.
- **Definitive catch-all** — wrapped the player in a Svelte `<svelte:boundary>` so *any* render/effect crash from a malformed recording that slips past validation shows a graceful "could not be replayed" state instead of a blank page. This closes the class of crash regardless of which nested field is malformed, not just the enumerated cases. Verified a valid recording still renders through the boundary (graph + Play, 0 console errors).
- **Comment length** — shortened the validation comment to ≤4 lines per AGENTS.md.

`npm run check` clean (0 errors). Claude and Pi have been "Good to merge".

rubenfiszel:
Addressed round 9 (`5072adb352a1b3ac8ede0936a2335f4955e4c071`):

- **Event payloads bypass the boundary** (#1) — correct that a `setTimeout`-deferred throw isn't caught by the Svelte boundary, so it's now validated at load: every recorded job's `events` must each carry an object `data` (JobLoader dereferences `event.data.running`/`completed` in its replay timer). `events:[{data:null}]` now rejects.
- **Remote flow/script payloads crash outside the boundary** (#2) — the `?src=` loader (added on this branch) makes flow/script recordings remotely supplied, and their players mount jobs straight into JobLoader outside the pipeline-only boundary. Both now go through a shared `isRecordedJob`/`isJobsMap` structural check (`initial_job` object + `events` with object `data`): `{"version":1,"jobs":{"x":null}}` rejects before `FlowRecordingReplay` mounts. The script path validates its single `job` the same way.

Verified all the malformed cases reject and a real recording still loads; `npm run check` clean (0 errors). Claude and Pi have been "Good to merge".

rubenfiszel:
Addressed round 10 (`fe480b2001f2f6b478665f003989df222d5e6348`):

- **Replay-timer crashes on malformed field types** — took the bounded fix suggested in the finding ("defensively handle malformed delayed events"): JobLoader's replay `setTimeout` callback is now wrapped in try/catch, so *any* malformed delayed event (e.g. `result_stream: 1` producing `1.concat(...)`) is logged and skipped instead of throwing an uncaught timer error. This closes the whole class of replay-timer crash regardless of which recorded field is bad — rather than validating each of the ~30 Job fields the replay callback might touch — and it protects all three players (flow/script/pipeline share this loader).

Combined with the load-time structural validation (clean toast for common malformed shapes) and the render boundary (sync crashes), malformed recordings now degrade gracefully at every layer. `npm run check` clean (0 errors). Claude and Pi have been "Good to merge".

rubenfiszel:
Addressed round 10 (`a9adc45a440aa2e9dbec53ee4b893f61e6c79c43`):

- **Async replay-completion failures escaped the timer guard** — correct that `onJobCompleted` is async (it yields at `tick()`) so its rejection escaped the synchronous try/catch. It's now `await`ed inside the try/catch, so a malformed completed event (e.g. `flow_status:{}` crashing the flow viewer's `change` callback) is caught and skipped rather than becoming an unhandled rejection.
- **Comprehensive boundary coverage** — wrapped the flow and script players in `<svelte:boundary>` too (previously only the pipeline player was), so any reactive-effect crash in those players from a remotely-loaded (`?src=`) recording degrades to the failed state instead of a blank page.

The replay path is now guarded at every layer: load-time structural validation (toast), the render boundary (sync render + reactive-effect crashes, all three players), the timer try/catch (sync throws), and the awaited completion (async rejections). Verified a valid recording still replays (graph + logs/result, 0 console errors); `npm run check` clean. Claude and Pi have been "Good to merge".

rubenfiszel:
Addressed round 12 (`698ed0e88f`):

- **Flow Play-handler errors escape boundaries** — correct that a Play-button (event-handler) throw isn't caught by a Svelte boundary. `FlowRecordingReplay.startReplay` (which walks each job's `flow_status.modules`) is now wrapped in try/catch: a malformed recording toasts and stays on the load screen instead of breaking the page.
- **Unbounded `?src=` download** — the auto-loader now caps the download at 100 MB: it rejects an oversized advertised `content-length` up front and aborts the stream (`reader.cancel()`) once the running total exceeds the cap, so a permissive-CORS origin can't OOM the tab before validation runs.
- **Comment length** — trimmed the replay-timer comment to 3 lines per AGENTS.md.

`npm run check` clean (0 errors). Claude and Pi have been "Good to merge".

rubenfiszel:
✅ Review round clean @ 698ed0e88f7efc1270186eda5ee1d8d49767c070

PR #10057, opened 2026-07-11
feat(ai-sessions): add plan mode

AlexRV12:
## Summary

Plan mode is a fourth autonomy posture for AI sessions. The assistant investigates with read-only tools, then proposes a plan for your approval; until you approve, it cannot modify anything. Session chats only — it is not offered where there is nothing to change.

## Changes

- **A fail-closed gate in `processToolCall`.** A tool runs in plan mode only if tagged `planModeSafe: true`; untagged means mutating, and that tag is the gate's only exemption — `exit_plan_mode` carries it, which is what keeps the posture escapable. A read tool someone forgets to tag gets over-blocked, which is annoying — a mutating one never leaks, which is the point. The gate runs before anything belonging to the tool itself, including its `validateBeforeConfirmation` probe, and re-runs *after* a confirmation wait, so entering plan mode while a mutating tool's card is pending can't leak that action through on approval.
  - Two tags carry caveats a reviewer should weigh: `diff` flushes autosaves the user's own editor already parked (never model-authored — the model cannot create drafts while planning), and `call_api_get` rests on the catalog exposing only side-effect-free GETs plus its own hard non-GET refusal. **Never mark a mutating GET as an `x-mcp-tool`.**
- **`enter_plan_mode` and `exit_plan_mode`.** The model can ask to enter the posture, and proposes its plan for approval on the way out. Approval restores the exact pre-plan posture, including YOLO. The user's standing autonomy always wins: the model can request plan mode but never impose it. An `exit_plan_mode` carrying no usable plan is refused before a card exists, so there is never an "Approve and implement" with nothing to read behind it — plan mode holds and the model is told to call again with the plan.
- **A session chat is held in GLOBAL mode.** Plan mode is only offered there, and the gate reads that mode, so `changeMode` refuses to move a session chat out of GLOBAL rather than resting the invariant on the mode picker being hidden.
- **The plan is saved as a session document when it is proposed**, not when approved, so it is readable in the preview pane at the moment you have to decide. A *session* owns exactly one plan document, revised in place: declining keeps it, a revision rewrites it rather than adding a draft per round, and coming back into plan mode to amend an approved plan amends *that* document instead of starting a fresh one. The id is derived from the session id, so the primary key is itself the constraint — a second tab, a reload that lost the in-memory id, or a rotated conversation all land on the same row rather than minting a second. The cost of that scoping is that an unrelated second plan overwrites the first, and "New chat" does not separate them: the session is the boundary, so a genuinely unrelated plan wants a new session.
- **A plan's versions are read against the one you approved, not against the newest.** Latest is only where the model happened to stop, so the approved version is never "stale": its bar is teal and points *forward* to the draft rather than warning about it, the version ahead of it is a draft that says it is unapproved, and anything behind is history that is neither and takes no pill at all. One helper answers this for every surface, so the artifact list, the preview header and a transcript card cannot drift on what counts as the plan — the list opens a plan at its approved version (which is what lets its pill say `plan` while an unapproved draft sits at the head) and labels the row with the badge of the version it will open, and a card far up the transcript opens the version *that card* proposed rather than whatever the document has become since. Neither pins a version it does not need to, since pinning the current one dresses it as history, banner and all; the viewer clears a pin that is not behind the document for the same reason.
- **An approval is reflected only once it is stored.** Ordinary content degrades unpersisted on purpose — text the store lost is still text — but an approval the store lost never happened, and showing the `plan` pill over it would put the user's name on it anyway.
- **Plan mode is visible.** `planMode.ts` owns the feature's tokens and prose — one teal pair (`teal-600 dark:teal-500`) and one tint, derived into the badge and the autonomy pill so no surface can drift; plus the copy the gate, the cards, the picker row and the tool results share. Teal rather than green because green is already the transcript's success colour — the check on a settled tool call and the accept button sit a few rows above these surfaces — so a mode signal in it reads as "this worked" rather than "this is held". The colour marks what the posture *is*, not what it refuses: the constraint reads as plain "Read-only" beside the autonomy picker, and a blocked call is a muted row, so neither competes with the badge and the bars.
- **A refused call renders as its own row** naming the tool, instead of an error card whose header was the model-facing instruction string — which is what the user saw at the exact moment they were asking why nothing happened. For the same reason a plan card that settles unapproved says "Plan not approved" rather than naming a choice: switching posture out of plan mode settles the card too, as does a call refused before any card was offered, and neither is a decision the user made.
- **The plan is pinned** to the top of the session artifact list, glyphed and badged, above a pane edge. Artifacts gain `role`, which records what one is *for* as opposed to `kind`, its format — a plan is markdown and still downloads as `.md`. The per-row Download button goes, because it wrote the newest content while the row's pill and its click both name the approved version — the one drift this PR otherwise closes. Download stays in the viewer header, where it follows the version on screen.
- **Declining ends the turn and opens a conversation** rather than triggering another research pass. Known limitation: this is prompt text, so it is a soft guarantee. Making "a rejected plan ends the turn" enforceable means treating a declined `exit_plan_mode` as a stop condition in the chat loop, which is a design change rather than a string edit.

## Test plan

Automated: `npm run check` reports 12 errors across 10,111 files, none of them in a file this branch touches (verified: zero errors and zero warnings under `copilot/chat` or `components/sessions`) and all present on `main` — generated-client drift. The `copilot/chat` and `sessions` suites are green at 1238 passing, apart from six failures that are pre-existing on `origin/main` (`ContextManager` ×2, `app/core` ×4).

Unit tests cover the gate (block / allow-tagged / unknown tool / cancel / post-confirmation re-gate), the real `exit_plan_mode` tool driven through the real gate so losing its `planModeSafe` tag fails rather than quietly making plan mode inescapable, posture capture and restore, the manual-switch card resolutions, the plan-mode-safe tool set, `derivePlanTitle`, the plan-document lifecycle, the approved-vs-draft version rule, and the artifact ordering. Ten guard the ways this can fail quietly, each confirmed to fail with its fix reverted:

- a plan-less `exit_plan_mode` must be refused before a card offers to approve it;
- a mutating tool must be blocked before its own validator runs;
- a plan write landing after the session was swapped must be filed under the session that proposed it;
- amending an approved plan must revise the existing document rather than create a second;
- a session that already holds a plan must not mint a second one, and must free the slot when that plan is deleted;
- an inherited property name must not be read as a plan card;
- an approval the store refused must not appear approved in memory;
- opening a plan already at its approved version must not pin it into the stale-version banner;
- a session chat must refuse to leave GLOBAL mode, so the gate cannot lift under it;
- a second `exit_plan_mode` from the batch that already ended the posture must be refused, not auto-approved under YOLO onto a summary no card showed.

There is also an end-to-end `ai_evals` case (`global-planmode1-hands-over-a-plan`): it starts in plan mode against the real gate and the real `exit_plan_mode`, and grades whether the model researches and hands over a usable plan instead of guessing at one. The checklist deliberately does not grade what the harness does for the model — `exit_plan_mode` writes the plan document itself, so "saves the plan as an artifact" would pass on any run where the tool is called at all.

Manual, in a session chat:

- [ ] Pick "Plan (read-only)" from the autonomy picker — the pill turns teal, and "Read-only" appears beside it.
- [ ] Ask for a change — mutating tools are refused as a muted "Blocked in plan mode" row naming the tool, while read-only tools still run.
- [ ] Let the model call `exit_plan_mode` — the plan opens in the preview pane with its badge while the card is still pending, and the read-only posture still holds.
- [ ] **Keep planning** — the document stays open, a revised plan rewrites that same document (one artifact, not two), and the model asks what is wrong instead of re-proposing.
- [ ] **Approve and implement** — mutating tools unblock and the pre-plan posture returns.
- [ ] Interrupt the implementation, switch back to Plan and ask for a change — the amended plan rewrites the same document; the artifact count stays at one.
- [ ] With an approved plan and a newer unapproved draft: opening the plan from the artifact list lands on the approved version with the teal "a newer draft is not approved" bar; the draft says "This revision is not approved" with the plan one click away; a card from an earlier round opens the version that round proposed.
- [ ] Reload, reopen that session from the list, switch to Plan and ask for another change — still the same document, still one artifact.
- [ ] Open the artifacts popover — the plan sits above the pane edge, over artifacts updated more recently.
- [ ] Switch *into* Plan while an `enter_plan_mode` card is pending → resolves to "Planning started". Switch *out of* Plan while an `exit_plan_mode` card is pending → "Plan not approved". Switch to **YOLO** instead → approves the plan, keeps the document, holds YOLO.
- [ ] Check both themes — the teal is the same signal in dark, at a lower tint.

## Screenshots

Live session chats on `claude-sonnet-4-6`, driven end to end: in each the model researched with read-only tools, proposed a plan, and was refused a write while planning. Two sessions, which is why the plan differs between shots — the overview, the blocked row and the composer come from one; the version bars and the artifact list from another, where a revision on request produced a real `v2`.

**Plan mode, end to end** — the trail in the middle column: a read tool running, `update_artifact` refused, and the round settling as "Plan not approved" with the plan one click away. On the right the document at its approved `v1`, badged `PLAN`, under the teal bar pointing forward to the newer draft. On the composer, the teal pill with "Read-only" beside it.

image

**A blocked call** — the model read the plan document, then tried to rewrite it with `update_artifact`. Reading is allowed and the write is not, so the refusal is a lean muted row naming the tool in the product's voice, not an error.

image

**The approved version** — `PLAN` badge, the version picker on `v1`, and the teal bar pointing *forward* to the newer draft rather than warning that this text is stale.

approved version

**The unapproved head** — the same document at `v2`, badged `DRAFT`, saying so in a bar that cannot be scrolled past, with the approved version one click away.

unapproved draft

**The artifact list** — the plan pinned above the pane edge with its glyph and its `PLAN` pill, over the session's other documents. The pill describes the version the row opens, not the newest.

artifact list

**Both bars in light** — the same two states; the tint and border alphas differ so the signal reads at the same weight in either theme.

approved light
draft light

**The composer in light** — a session about to take its first request: the tinted pill, and the constraint stated in plain text beside it rather than in a colour of its own.

image

### DIFF
--- a/ai_evals/adapters/frontend/core/global/evalArtifactStore.ts
+++ b/ai_evals/adapters/frontend/core/global/evalArtifactStore.ts
@@ -1,11 +1,54 @@
 // SessionArtifactsStore can't run here (bun has no IndexedDB, nor the compiled $state runes),
 // so mirror only the shape the artifact tools call, not its scoping or race handling.
 export const EVAL_SESSION_ID = "eval-session";
-export function createEvalArtifactHelpers() {
+
+/** An artifact the session already holds when the case starts: history has to predate the
+ * run, since one prompt cannot both build a past and reason about it. */
+export interface SeededArtifact {
+  name: string;
+  role?: "plan";
+  /** Which version the user agreed to. Below the last one means the current text is a
+   * proposal they turned down, which is the state worth seeding. */
+  approvedVersion?: number;
+  /** Oldest first; the last one is the artifact's current content. */
+  versions: Array<{ content: string; note?: string }>;
+}
+
+export function createEvalArtifactHelpers(seed: SeededArtifact[] = []) {
   const items = new Map<string, Record<string, any>>();
   // Snapshots per artifact id, oldest first — the version tools read history from here.
   const history = new Map<string, Array<Record<string, any>>>();
   let seq = 0;
+  for (const entry of seed) {
+    const id = `eval-artifact-${seq++}`;
+    const current = entry.versions.at(-1);
+    if (!current) continue;
+    items.set(id, {
+      id,
+      sessionId: EVAL_SESSION_ID,
+      chatId: "eval-chat",
+      kind: "md",
+      name: entry.name,
+      content: current.content,
+      role: entry.role,
+      approvedVersion: entry.approvedVersion,
+      createdAt: 0,
+      updatedAt: seq,
+      version: entry.versions.length,
+    });
+    history.set(
+      id,
+      entry.versions.map((v, i) => ({
+        key: `${id}:${i + 1}`,
+        artifactId: id,
+        version: i + 1,
+        name: entry.name,
+        content: v.content,
+        savedAt: i,
+        note: v.note,
+      })),
+    );
+  }
   const snapshotOf = (
     artifact: Record<string, any>,
     version: number,
@@ -21,6 +64,16 @@ export function createEvalArtifactHelpers() {
   });
   const store = {
     create: async (sessionId: string, input: Record<string, any>) => {
+      // One plan per session, as SessionArtifactsStore enforces it — the tool refuses
+      // first, so reaching this means a case drove create_artifact past that message.
+      if (
+        input.role === "plan" &&
+        [...items.values()].some(
+          (a) => a.sessionId === sessionId && a.role === "plan",
+        )
+      ) {
+        throw new Error(`Session ${sessionId} already has a plan document`);
+      }
       const now = seq++;
       const artifact = {
         id: `eval-artifact-${now}`,
@@ -29,6 +82,10 @@ export function createEvalArtifactHelpers() {
         kind: input.kind ?? "md",
         name: input.name,
         content: input.content,
+        // The plan document is only distinguishable by these, both in the snapshot the
+        // judge reads and in what list_artifacts reports back to the model.
+        role: input.role,
+        approvedVersion: input.approvedVersion,
         createdAt: now,
         updatedAt: now,
         version: 1,
@@ -58,6 +115,15 @@ export function createEvalArtifactHelpers() {
         ...existing,
         name: input.name ?? existing.name,
         content: input.content ?? existing.content,
+        // Carried only onto a version this write produced, as SessionArtifactsStore does:
+        // a rename cannot promote a proposal the user turned down.
+        approvedVersion:
+          input.approvedVersion ??
+          (input.keepApproved &&
+          existing.approvedVersion !== undefined &&
+          contentChanged
+            ? version
+            : existing.approvedVersion),
         updatedAt: seq++,
         version,
       };
--- a/ai_evals/adapters/frontend/core/global/globalEvalRunner.ts
+++ b/ai_evals/adapters/frontend/core/global/globalEvalRunner.ts
@@ -12,9 +12,14 @@ import {
   getGlobalDraft,
   listGlobalDrafts,
 } from "../../../../../frontend/src/lib/components/copilot/chat/global/userDraftAdapter";
+import { appendPlanModeInstructions } from "../../../../../frontend/src/lib/components/copilot/chat/planMode";
 import type { Tool as ProductionTool } from "../../../../../frontend/src/lib/components/copilot/chat/shared";
+import { createEvalPlanTools } from "./planModeTools";
 import { UserDraft } from "../../../../../frontend/src/lib/userDraft.svelte";
-import { createEvalArtifactHelpers } from "./evalArtifactStore";
+import {
+  createEvalArtifactHelpers,
+  type SeededArtifact,
+} from "./evalArtifactStore";
 import type { ModeRunContext } from "../../../../core/types";
 import type { GlobalDraftState } from "../../../../core/validators";
 import type { WindmillBackendSettings } from "../../../../core/windmillBackendSettings";
@@ -83,11 +88,16 @@ export interface GlobalEvalOptions {
   user?: GlobalUserFixture;
   // Emulate a session chat (preview tools + session prompt); default false = standalone baseline.
   sessionChat?: boolean;
+  // Start in plan mode: the gate refuses every tool without `planModeSafe`, and the two plan
+  // tools are offered. Needs sessionChat, which is what plan mode is gated on in production.
+  planMode?: boolean;
   model?: string;
   maxIterations?: number;
   provider?: AIProvider;
   backend: WindmillBackendSettings;
   workspaceRoot?: string;
+  // Artifacts the session already holds when the run starts.
+  artifacts?: SeededArtifact[];
   runContext?: ModeRunContext;
 }
 
@@ -113,19 +123,41 @@ export async function runGlobalEval(
       process.env[DISABLE_ACTIVE_EDITOR_CONTEXT_ENV] !== "1";
     // Pass the seeded identity straight to the prompt builder rather than mutating
     // the process-global `userStore`, so concurrent cases never race on it.
-    const evalArtifacts = createEvalArtifactHelpers();
+    const evalArtifacts = createEvalArtifactHelpers(options.artifacts);
+    const planMode = options.planMode
+      ? createEvalPlanTools({
+          create: evalArtifacts.helpers.artifacts.create,
+          sessionId: evalArtifacts.helpers.sessionId,
+          chatId: evalArtifacts.helpers.getChatId(),
+        })
+      : undefined;
+    const baseSystemMessage = prepareGlobalSystemMessage(undefined, {
+      user: options.user,
+      previewTools: options.sessionChat ?? false,
+    });
     const rawResult = await runEval({
       userPrompt,
-      systemMessage: prepareGlobalSystemMessage(undefined, {
-        user: options.user,
-        previewTools: options.sessionChat ?? false,
-      }),
+      systemMessage: baseSystemMessage,
+      // Re-derived per request, as production's getter is: the instructions have to leave
+      // the prompt when the plan is approved, or the model is still told it may not build
+      // while the gate has already opened.
+      getSystemMessage: planMode
+        ? () =>
+            planMode.isPlanModeActive()
+              ? appendPlanModeInstructions(baseSystemMessage, 0)
+              : baseSystemMessage
+        : undefined,
+      isPlanModeActive: planMode?.isPlanModeActive,
+      isToolAvailable: planMode?.isToolAvailable,
       userMessage: prepareGlobalUserMessage(
         userPrompt,
         [],
         injectActiveEditorContext ? { workspace: workspaceRoot } : {},
       ),
-      tools: getGlobalEvalTools(options.sessionChat ?? false),
+      tools: [
+        ...getGlobalEvalTools(options.sessionChat ?? false),
+        ...(planMode?.tools ?? []),
+      ],
       helpers: evalArtifacts.helpers,
       apiKey,
       getOutput: async () => ({
--- a/ai_evals/adapters/frontend/core/global/planModeTools.ts
+++ b/ai_evals/adapters/frontend/core/global/planModeTools.ts
@@ -1,0 +1,69 @@
+import {
+  EXIT_PLAN_MODE_TOOL,
+  EXIT_PLAN_MODE_TOOL_DESCRIPTION,
+  PLAN_MODE_MESSAGES,
+  derivePlanTitle,
+  exitPlanModeArgs,
+  planSummaryOf,
+} from "../../../../../frontend/src/lib/components/copilot/chat/planMode";
+import { createToolDef } from "../../../../../frontend/src/lib/components/copilot/chat/shared";
+import type { Tool as ProductionTool } from "../../../../../frontend/src/lib/components/copilot/chat/shared";
+
+/**
+ * `exit_plan_mode` built from the production schema, description and messages, so a case
+ * exercises the real gate and wording with the posture living here rather than on the
+ * manager. It resolves immediately — the runners define no `requestConfirmation`, so the
+ * plan is always approved and a refused one cannot be expressed.
+ */
+export function createEvalPlanTools(artifacts: {
+  create: (
+    sessionId: string,
+    input: Record<string, unknown>,
+  ) => Promise<{ id: string; name: string }>;
+  sessionId: string;
+  chatId: string;
+}): {
+  tools: ProductionTool<{}>[];
+  isPlanModeActive: () => boolean;
+  isToolAvailable: (name: string) => boolean;
+} {
+  let planActive = true;
+  return {
+    isPlanModeActive: () => planActive,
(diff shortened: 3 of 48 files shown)


AlexRV12:
### Re: [P1] "Unapproved plan revisions are marked approved" — this is the intended design

`keepApproved: true` is deliberate and I'd like to keep it. The finding assumes a plan document that must stay frozen at the exact text the user signed off. The design is the opposite: a plan document the agent **maintains** after approval, with every change put in front of the user. Blocking or demoting those edits was considered and rejected.

**The revision is not silent.** `update_artifact` renders its own row in the transcript naming the artifact, expandable via `showDetails: true`, and it calls `openArtifact` — so the revised plan is opened in the preview pane at the moment it changes. What it does not do is halt the turn for a confirmation card, and that is the autonomy posture's call, not this tool's. A user who wants a card in front of every document write is choosing an autonomy mode, not a plan-mode rule.

**Freezing the pointer has a cost the finding doesn't price in.** `listOpenTarget` (`planMode.ts:93`) pins the approved version whenever it is behind the head. With the pointer frozen at v1, the artifact list would open v1 *forever* while the live plan moved on to v5 — so the row labelled `plan`, the one surface a user clicks to answer "what did we agree to do?", would get steadily staler, with the current plan an extra click away behind a "not approved" banner. That inverts the signal the versioning exists to give: the reason versions are read against the approved one rather than the newest is precisely so the approved version is never stale.

**The draft/plan distinction is for the undecided case.** It marks a proposal the user has not yet ruled on — a revision produced *inside* plan mode, where `savePlan` deliberately does not carry approval forward. Once they have approved and their own posture is back, the document is theirs and the agent keeps it accurate under the autonomy they chose.

**On the apparent contradiction with `create_artifact`**, whose comment says a plan written there "stands as a draft until a card decides it": that is about *conferring* approval on a document that never had any. `update_artifact` doesn't confer approval, it carries an existing one across an edit the user is shown. The guard is explicit — `input.keepApproved && existing.approvedVersion !== undefined && contentChanged` — so a plan nobody ever approved stays unapproved through any number of `update_artifact` calls, and a rename moves nothing.

Residual risk stated plainly: an agent that materially rewrites an approved plan mid-execution carries the approval across with it. What bounds that is visibility — the transcript row and the preview pane — and the user's own autonomy setting, rather than a frozen version pointer.

The other findings from this round (the `PLAN_MODE_MESSAGES` import in the eval adapter, the plan-scope wording, the dropped list-row Download) are being addressed separately.

AlexRV12:
### Re: both P2s — already addressed; this round read a description snapshot 8 seconds stale

Both nits were fixed in the PR description before the review posted, but the run couldn't see it, so I want to leave the evidence here rather than just dismiss them.

| Time (UTC) | Event |
|---|---|
| 17:14:39 | push of `0022271` lands → Codex Auto Review triggers |
| **17:14:47** | **PR body updated** (`pullRequest.lastEditedAt`) |
| 17:22:07 | review posts |

`codex-pr-review.yml:82` passes `EVENT_BODY: ${{ github.event.pull_request.body }}` on auto-triggered runs — the webhook payload, frozen at trigger time. The live `gh pr view --json body` path a few lines below only applies to a manual `/review` carrying a PR number. So an auto run reads the description as it was 8 seconds before this edit, however long the job itself takes.

**1 — Plan scope.** The description now describes session scope, and says outright that New chat does not separate plans:

> A *session* owns exactly one plan document, revised in place […] The id is derived from the session id, so the primary key is itself the constraint — a second tab, a reload that lost the in-memory id, or a rotated conversation all land on the same row rather than minting a second. The cost of that scoping is that an unrelated second plan overwrites the first, and "New chat" does not separate them: the session is the boundary, so a genuinely unrelated plan wants a new session.

The two stale test-plan bullets were repointed at the guards that actually exist (`files a plan under the session that proposed it`, `holds one plan per session, and frees the slot when it is deleted`).

**2 — Download.** Taking the second branch of the suggested remedy — documented rather than restored, with the reason:

> The per-row Download button goes, because it wrote the newest content while the row's pill and its click both name the approved version — the one drift this PR otherwise closes. Download stays in the viewer header, where it follows the version on screen.

That is why it wasn't restored: `ArtifactsSegment`'s row called `download(…, a.content, …)`, i.e. the head, while `listOpenTarget`/`listBadge` make the same row label and open the *approved* version. Restoring it as-is would reintroduce exactly the label/action drift this PR closes; a version-aware row download would be new surface. `ArtifactViewer.downloadFile()` already writes `shown.content`/`shown.name`, so the viewer's copy follows whichever version is on screen.

Not re-running the review here: `pr-review-commands.yml:174` skips an agent when a successful run already covers the head, which this one does. Any future push will snapshot the corrected description.

AlexRV12:
Both P2s from this round, one taken and one dismissed with reasoning.

### Pi's P2 — transient badge/bar during a pinned load: **fixed** in `1fa4ef21f3`

The diagnosis was right, and it's worth recording why the fix isn't the first of the two suggested.

During the read the body is still rendering the *head* — `shown = pinnedContent ?? artifact` — so `draft` + "not approved" is an accurate label for what is on screen. The problem is that it's a flash of the alarming signal on the reassuring case, not that it's false.

That's why *"compute the view against `pinned` rather than `shownVersion`"* is the wrong remedy: it would print the `plan` badge and the teal "This is the plan you approved" bar **over the unapproved draft text**, which is exactly what the comment on `shownVersion` exists to prevent — *"Label the snapshot that is rendered, not the one just requested."* That trades a true transient signal for a false one.

Took the second suggestion instead: the header stays silent while `restoringPin`. That also makes the component consistent with itself — the body is already deliberately blank in that window (`ArtifactViewer.svelte:262`), and the version picker already reads `restoringPin ? pinned : shownVersion` (`:158`). The header's badge and bars were the only surfaces not honouring the flag.

### Codex's P2 — raw `<button>` on the plan-card disclosure: **dismissing**

The rule cited is `AGENTS.md` → *"use Windmill's design-system components (Button, TextInput, Select) — never raw HTML elements."* Applied to a disclosure header it would equally condemn **`ChatCollapsibleCard.svelte:57`**, the repo's own collapsible that every other tool row in this transcript renders through, which is itself a native `<button>`. The plan card is following that precedent, not departing from it — the rule targets form controls, and `Button` carries border, padding, `unifiedSize` and variant styling, which is the opposite of a transparent full-width row whose job is to collapse into the transcript.

On the keyboard follow-up: that behaviour is already there, and *because* the element is native. `<button>` gets Enter, Space and a focus ring for free; `aria-expanded` and `disabled` are both already bound (`ToolExecutionDisplay.svelte:156-157`). Wrapping it is the change that would put those at risk.

There is a real improvement adjacent to this that's worth doing separately: the plan card hand-rolls a header that duplicates `ChatCollapsibleCard` — its own comment says as much. Unifying them doesn't drop in, because `ChatCollapsibleCard` takes `label: string` with no slot for a leading icon and the plan card needs a per-state one (`Loader2` / `Check` / `CircleMinus` / `ClipboardList`), so it needs a new snippet prop on a component every transcript row depends on. Not a change I want to make mid-round on a diff this size; filing it as a follow-up.

AlexRV12:
### [P1] stale duplicate `exit_plan_mode` — **confirmed and fixed** in `d876dc9045`

Traced the whole chain and it holds, including the snapshot premise: `anthropic.ts:193` takes `tools` as a function parameter, and the `for (const toolCall of toolCallsToProcess)` loop at `:438` reuses that frozen array for every call in the batch — so although `get tools()` is a live getter that would return `[]` under YOLO, it is read once, before the loop.

Reproduced it in a test before fixing. With the guard reverted, the second stale call returns:

> `"Plan approved and saved as a document. You may now execute it."`

— its own summary written, `approvedVersion` moved onto it, no card ever shown.

**One correction to the report, which changes where the fix has to go.** It isn't only `fn` that writes: `onConfirmationRequested` calls `#ensurePlanDoc` as well, so the document is written when the *card is displayed*, not just when the tool runs. A guard inside `fn` would still let the stale summary reach the store on the carded path. The refusal therefore lands in `validateBeforeConfirmation`, which short-circuits ahead of both.

Also worth recording for anyone reading later: this is **YOLO-specific**. In DEFAULT and ACCEPT_EDIT the second call raises a card, so the user sees the summary before it is approved — consent is intact. It is the auto-accept path, where `autoAcceptsTool` excludes only `enter_plan_mode`, that turns it into a silent approval.

**No effect on the maintenance path.** Revising an approved plan outside the posture with `update_artifact` is a deliberate design decision and is untouched — this only stops the *consent gate* firing when there is nothing left to consent to, which is what `EXIT_PLAN_MODE_TOOL_DESCRIPTION` already tells the model ("Valid only while plan mode is active … a revision goes into the plan document with `update_artifact`"). The code now enforces its own description. Nothing legitimate is refused either: `get tools()` returns `[enterTool]` when inactive and `[]` when auto-accepting, so the tool is never offered in a later iteration — only the frozen snapshot could reach it.

Regression added as suggested (`planModeController.test.ts`), driving both hand-overs through the real `processToolCall` from one frozen tool list with `shouldAutoAcceptToolConfirmations: () => true`, asserting the second is refused and that the document keeps `version: 1, approvedVersion: 1`. Mutation-tested: reverting the guard fails that test and only that test.

AlexRV12:
### [P2] existing plan tabs bypass the pin-loading guard — **dismissing**

The observation is accurate: `restoringPin` is set at mount and never re-armed when `pinned` changes. But that is deliberate, and re-arming it would break the behaviour it exists to protect.

**`restoringPin` does not mean "a read is in flight."** It means *this viewer has never shown anything yet*. The comment at `ArtifactViewer.svelte:44-46` states it:

> A pin that arrives with the tab has nothing honest to show until its snapshot lands: the latest document is not the version asked for, and the banner saying so is keyed on the snapshot itself. **A pin picked while reading keeps the document already on screen instead.**

It is the same flag that blanks the **body** at `:270`. Re-arming it on every prop change would blank the pane on every version switch — exactly what that line rejects.

**The invariant is that the header describes whatever the body is showing.** Both paths honour it:

| path | body during the read | header during the read |
|---|---|---|
| mount with a pin | blank (`:270`) | *previously* `draft` + orange — labelling a document not on screen. That was the incoherence `1fa4ef21` fixed. |
| existing tab | keeps the current document on screen | labels that same document — correct |

**And there is no flash on the path described.** Walk it: a tab open on the head (v2, unapproved), the user clicks the plan and pins v1 in the same tab.

- before — body: v2 draft, header: `draft` + orange
- during the read — body: v2 draft *unchanged*, header: `draft` + orange *unchanged*
- after — body: v1, header: `plan` + teal

Nothing transitions mid-read. The UI holds still and then updates once, atomically, when the snapshot lands. *"Continues rendering the `draft` badge and orange warning until `getVersion` resolves"* is accurate as a description, but that is continuity rather than a flash — the draft badge is true for the draft text still on screen.

What the suggested change would produce is the unapproved draft sitting in the pane with **no badge and no warning above it** for the duration of the read. That removes a true warning about text the user is looking at. It is the same class of error as the *"judge `pinned` instead"* alternative dismissed last round — a header signal that disagrees with the body — just in the opposite direction.

Leaving the code as it stands.

Issue #5986, opened 2025-06-18
[AI] feat: support octet-stream response type from a base64 result


rubenfiszel:
/ai in windmill-api/jobs.rs if windmill content type is an octet stream and the result is a base64 decode it and send it as bytes. Request made by @HugoCasa

justin212407:
@rubenfiszel Can I take this up if it is open to community contributions?

lucsoft:
i hid this issue trying to serve playwright reports which need binary files (but an object storage would already be way too big for my usecase). I tried to move this forward with a PR. Im very happy if the API doesn't match the standards and needs refinement, any feedback is welcome, im here to adjust the course of action if needed.

PR #10058, opened 2026-07-11
feat(triggers): serve binary HTTP-route responses via base64 transfer encoding

lucsoft:
## Summary
Lets sync HTTP routes / webhooks return **real binary** bodies with **any**
MIME type, without requiring workspace object storage.

A script returns the standard composite result plus a new opt-in field:

    {
      "wm_content_type": "application/pdf",          // any MIME type
      "wm_content_transfer_encoding": "base64",       // NEW — opt-in
      "wm_status_code": 200,                           // optional
      "wm_headers": { "content-disposition": "inline; filename=\"r.pdf\"" },
      "result": "<base64 string>"
    }

`result_to_response` then base64-decodes `result` into raw bytes and sends them
with the given content type.

This generalizes #5986 (which was scoped to `application/octet-stream`): the
decode is decoupled from the content type, so any type works.

## Design choices (vs. the earlier auto-generated attempt in #5988)
- **Explicit opt-in**, never guessing: decoding only happens when
  `wm_content_transfer_encoding == "base64"`. A legitimate octet-stream *string*
  that happens to be valid base64 is never silently mangled.
- **Any content type**, not just `application/octet-stream`.
- **Hard errors instead of silent fallback**: invalid base64, an unsupported
  encoding, or a transfer encoding without a content type all return an error.
- **Backward compatible**: without the new field, behavior is identical
  (string served verbatim / JSON as before).

## Limitations (documented, out of scope)
Still buffered through the JSON job result (base64 ~+33 %, bound by
`wait_result` timeout) — for large/streamed downloads the static-asset/S3 path
remains the right tool. This targets small/medium payloads without storage.

## Tests
Added `result_to_response_tests` (5, all passing):
- base64 result → raw bytes (uses non-UTF-8 bytes to prove it's real binary)
- invalid base64 → hard error (no silent fallback)
- unsupported encoding → rejected
- transfer encoding without content type → rejected
- plain string result → still served verbatim (regression)

Aliases `windmill_content_transfer_encoding` / `wm_content_transfer_encoding`
match the existing `wm_*` naming.

---
## Summary by cubic
Enable sync HTTP routes/webhooks to return real binary bodies for any MIME type by accepting a base64-encoded `result` and decoding it to bytes at response time. This removes the need for object storage for small/medium binary payloads.

- **New Features**
  - Adds optional `wm_content_transfer_encoding: "base64"` (requires `wm_content_type`); when set, decodes string `result` from base64 and sends raw bytes.
  - Strict validation: only JSON string results are allowed in base64 mode; invalid base64, unsupported encodings, or encoding without `wm_content_type` return errors.
  - Backward compatible: existing string/JSON responses are unchanged.

<sup>Written for commit cd46fdf25fb58274384f927c9639341a04791eb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-jobs/src/execution.rs
+++ b/backend/windmill-api-jobs/src/execution.rs
@@ -12,6 +12,7 @@ use axum::{
     response::{IntoResponse, Response},
     Json,
 };
+use base64::Engine as _;
 use http::{HeaderMap, HeaderName, HeaderValue};
 use hyper::StatusCode;
 use serde::Deserialize;
@@ -254,6 +255,8 @@ pub struct WindmillCompositeResult {
     windmill_content_type: Option<String>,
     #[serde(alias = "wm_headers")]
     windmill_headers: Option<HashMap<String, String>>,
+    #[serde(alias = "wm_content_transfer_encoding")]
+    windmill_content_transfer_encoding: Option<String>,
     result: Option<Box<RawValue>>,
 }
 
@@ -375,11 +378,13 @@ pub fn result_to_response(result: Box<RawValue>, success: bool) -> error::Result
             windmill_status_code,
             windmill_content_type,
             windmill_headers,
+            windmill_content_transfer_encoding,
             result: result_value,
         }) => {
             if windmill_content_type.is_none()
                 && windmill_status_code.is_none()
                 && windmill_headers.is_none()
+                && windmill_content_transfer_encoding.is_none()
             {
                 return Ok((
                     if success {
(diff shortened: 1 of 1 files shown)


lucsoft:

rubenfiszel:
thank you

PR #10064, opened 2026-07-12
feat(nativets): expose the standard web-platform globals deno_web provides

lucsoft:
## Summary
The nativets (in-process deno_core) runtime imports several deno_web / deno_url
modules but never assigned their globals to `globalThis`, so nativets scripts
couldn't use standard web-platform APIs that exist in browsers, Deno and the bun
runner. This wires the missing ones onto the global scope. For four of them the
deno_web module was previously imported only for its side effects; those imports
are switched to namespace imports so the constructors can be assigned.

## What's added
- `TextEncoder` / `TextDecoder` (+ `TextEncoderStream` / `TextDecoderStream`)
- `File`
- `Event` / `EventTarget` / `CustomEvent`
- `ReadableStream` / `WritableStream` / `TransformStream`
  (+ `ByteLengthQueuingStrategy` / `CountQueuingStrategy`)
- `URLPattern`
- `structuredClone`, `performance`
- `CompressionStream` / `DecompressionStream`
- `MessageChannel` / `MessagePort`

## Why it matters (parity with the bun runner)
Scripts assume these globals exist — browsers, Deno and bun all provide them — so
nativets was the outlier: `new TextEncoder()` threw, `res.body instanceof
ReadableStream` failed, `AbortSignal` was present while its base `EventTarget`
was not, etc. I verified the gap empirically by probing `typeof <global>` in both
runners against a live worker; every global added here is present in bun, so this
brings nativets to parity with bun's global surface for them.

## Tests
Opt-in smoke tests in `windmill-runtime-nativets`:
- `smoke_text_encoder_decoder` — UTF-8 encode/decode round-trip (multi-byte char)
- `smoke_web_globals_are_wired` — asserts each newly-wired global is defined
  (guards against a wrong export name)

Run: `cargo test -p windmill-runtime-nativets -- --ignored smoke_text_encoder_decoder smoke_web_globals_are_wired`

## Implementation note
The message-port, compression and performance modules were imported only for
their side effects; they are switched to namespace imports so their constructors
can be assigned. `structuredClone` is sourced from the message-port module (the
spec-compliant function that validates arguments and accepts options), not the
single-arg internal helper in `02_structured_clone.js`.

## Follow-ups (out of scope)
- Needs an additional extension: `crypto` / `crypto.subtle`, `WebSocket`, `Worker`.
- `structuredClone` is the spec function, but the runtime doesn't wire the
  transfer op, so `options.transfer` clones without detaching the source
  (pre-existing runtime limitation, not addressed here).
- `EventSource` and `ImageData` are absent from the bun runner too, so not added.

---
## Summary by cubic
Expose standard web‑platform globals from `deno_web`/`deno_url` on `globalThis` in the nativets runtime, bringing parity with browsers, Deno, and bun. This lets scripts use standard encoding, events, streams, URL matching, cloning, timing, compression, and message channel APIs.

- **New Features**
  - Exposes TextEncoder/TextDecoder (+ streams), File, Event/EventTarget/CustomEvent, Readable/Writable/Transform streams (+ ByteLength/Count queuing strategies), URLPattern, `structuredClone` (spec), `performance`, Compression/DecompressionStream, and MessageChannel/MessagePort. Added smoke tests for UTF‑8 round‑trip, global presence, spec `structuredClone`, and `performance.now()`.

- **Refactors**
  - Switched message‑port, compression, and performance modules to namespace imports so their constructors can be assigned to `globalThis`.

<sup>Written for commit 94abd89becc17db4c2575f74957f7557233d6890. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lucsoft:
looks like this feature has landed as a release now

PR #10069, opened 2026-07-13
ci: run Codex/Pi review on fork PRs when a maintainer triggers it

rubenfiszel:
## Summary

Lets a maintainer's `/codex`, `/pi`, `/claude`, or `/review` comment run the automated review on **external/fork PRs**. Previously the Codex and Pi workflows unconditionally skipped every cross-repository PR, so the maintainer-triggered command path (`pr-review-commands.yml` → `workflow_call`, gated by `check-write-access`) bailed out on forks too.

Note: the GitHub Actions "Approve workflow to run" button does **not** help here — fork `pull_request` runs always get a read-only token and no secrets, so the review jobs (need API-key secrets + write to post the comment) can never run through that path. The maintainer-comment path is the correct opt-in mechanism, and it runs in the trusted base-repo context.

## Changes

**Enable fork review on the maintainer-command path** (`codex-pr-review.yml`, `pi-pr-review.yml`): gate the fork-skip on the automatic `pull_request` trigger only, detected via an empty `INPUT_PR_NUMBER` (`inputs.pr_number` is only set on `workflow_call` — the same discriminator the metadata step already uses).

**Harden the fork path against secret exfiltration** (addresses the CI review). For fork PRs only (detected via a new `is_fork` step output):
- **Withhold `WINDMILL_EE_PRIVATE_ACCESS`**: skip the EE access/checkout/substitution steps, so the private-repo token is never present in the job environment.
- **Read the review policy/prompt from the trusted base ref** (`git show origin/<base>:REVIEW.md` and the prompt file) instead of the attacker-controlled merge checkout — a malicious fork can no longer rewrite the reviewer's own instructions.
- **Restrict the agent**: Codex runs with `-s workspace-write` (network disabled) instead of `danger-full-access`; Pi drops the `bash` tool (`read,grep,find,ls` only).

Non-fork PRs are unchanged in every respect.

`pr-ready-review.yml` (Claude): **no change needed.** It has no fork skip, checks out `main` (not the fork ref), and reviews via `gh pr diff`/`gh pr view` with a restricted API-only tool allowlist — so it already reviews fork PRs on the command path without executing fork code or exposing this surface.

## Residual risk

The provider API key (`OPENAI_API_KEY`/`CODEX_AUTH_JSON`, `DEEPSEEK_API_KEY`) still has to be available for the model call. Indirect prompt injection via the diff → secret echoed into the public review comment is inherent to "an LLM reviews untrusted code"; it's mitigated by removing the highest-value secret (the EE token) and by the trusted-prompt + no-network / no-shell restrictions above, all behind the maintainer opt-in gate.

## Test plan

- [ ] Maintainer comments `/codex` (and `/pi`) on a fork PR → the "Resolve PR metadata" step does **not** print the skip line and a review comment is posted.
- [ ] On that fork run, the "Check EE access" step is skipped (no EE token), the prompt is sourced from the base ref, and Codex uses `workspace-write` / Pi uses the bash-less tool set.
- [ ] `/claude` on a fork PR still posts a review (already worked).
- [ ] A normal (non-fork) `pull_request` still auto-reviews with unchanged behavior (EE substitution active, full sandbox / bash tool).
- [ ] A fork PR on the automatic `pull_request` trigger is still skipped for Codex and Pi.

### DIFF
--- a/.github/codex/pr-review.prompt.md
+++ b/.github/codex/pr-review.prompt.md
@@ -1,5 +1,5 @@
 # Codex output format
 
-- Read `./.github/codex/pr-review-context.md` for PR metadata and the diff commands.
+- Read the review context file whose absolute path is given at the end of these instructions; it holds the PR metadata and the diff commands.
 - Return a markdown PR comment starting with `## Codex Review`.
 - Tag each finding with a severity (P0 / P1 / P2), file path, and line number when known confidently.
--- a/.github/pi/pr-review.prompt.md
+++ b/.github/pi/pr-review.prompt.md
@@ -1,6 +1,6 @@
 # Pi output format
 
-- Read `./.github/pi/pr-review-context.md` for PR metadata and the diff commands.
+- Read the review context file whose absolute path is given at the end of these instructions; it holds the PR metadata and the diff (or the git commands to produce it).
 - Return a markdown PR comment starting with `## Pi Review`.
 - Tag each finding with a severity (P0 / P1 / P2), file path, and line number when known confidently.
 - Output ONLY the final review markdown — no preamble, no thinking, no tool transcripts.
--- a/.github/workflows/codex-pr-review.yml
+++ b/.github/workflows/codex-pr-review.yml
@@ -114,6 +114,14 @@ jobs:
             echo "skip=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
+          # PR title/body are attacker-controlled free text. Use an unguessable
+          # per-run delimiter so a fork can't embed a fixed heredoc terminator to
+          # close the block early and inject extra outputs — e.g. is_fork=false,
+          # which (last-write-wins) would re-enable the EE checkout and the
+          # trusted-path agent settings for fork code.
+          RAND=$(head -c 16 /dev/urandom | od -An -tx1 | tr -d ' \n')
+          TITLE_EOF="TITLE_EOF_${RAND}"
+          BODY_EOF="BODY_EOF_${RAND}"
           {
             echo "skip=false"
             echo "is_fork=$IS_FORK"
@@ -122,12 +130,12 @@ jobs:
             echo "base_sha=$BASE_SHA"
             echo "head_sha=$HEAD_SHA"
             echo "pr_author=$PR_AUTHOR"
-            echo 'title<<PR_TITLE_EOF'
+            echo "title<<$TITLE_EOF"
             printf '%s\n' "$PR_TITLE"
-            echo 'PR_TITLE_EOF'
-            echo 'body<<PR_BODY_EOF'
+            echo "$TITLE_EOF"
+            echo "body<<$BODY_EOF"
             printf '%s\n' "$PR_BODY"
-            echo 'PR_BODY_EOF'
+            echo "$BODY_EOF"
           } >> "$GITHUB_OUTPUT"
 
       - name: Checkout repository
--- a/.github/workflows/codex-pr-review.yml
+++ b/.github/workflows/codex-pr-review.yml
@@ -106,9 +106,8 @@ jobs:
           fi
           # Fork PRs run untrusted code with secrets present, so the automatic
           # pull_request trigger never reviews them. A non-empty INPUT_PR_NUMBER
-          # means we arrived via workflow_call, which only happens after an
-          # authorized maintainer opted in (e.g. a /codex comment gated by
-          # check-write-access), so allow forks on that path.
+          # means we arrived via workflow_call (a maintainer /codex comment gated
+          # by check-write-access), so allow forks only on that path.
           if [ "$IS_FORK" = "true" ] && [ -z "$INPUT_PR_NUMBER" ]; then
             echo "Skipping Codex review for fork PR (automatic trigger)."
             echo "skip=true" >> "$GITHUB_OUTPUT"
@@ -116,9 +115,8 @@ jobs:
           fi
           # PR title/body are attacker-controlled free text. Use an unguessable
           # per-run delimiter so a fork can't embed a fixed heredoc terminator to
-          # close the block early and inject extra outputs — e.g. is_fork=false,
-          # which (last-write-wins) would re-enable the EE checkout and the
-          # trusted-path agent settings for fork code.
+          # inject extra outputs — e.g. is_fork=false (last-write-wins), which
+          # would re-enable the EE checkout and trusted-path settings for forks.
           RAND=$(head -c 16 /dev/urandom | od -An -tx1 | tr -d ' \n')
           TITLE_EOF="TITLE_EOF_${RAND}"
           BODY_EOF="BODY_EOF_${RAND}"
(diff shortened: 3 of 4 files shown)


rubenfiszel:
Addressed the security review (thanks @claude for the merge-blocker call-out — the attacker-controlled-prompt amplifier was the key insight).

Hardened the fork path (fork PRs only; non-fork behavior unchanged) in `d607f49`:

1. **Withhold `WINDMILL_EE_PRIVATE_ACCESS`** — the `Check EE access` step is now gated on `is_fork != 'true'`, so the private-repo token (the highest-value secret) is never in the environment for fork reviews. The EE checkout/substitution steps skip automatically since `steps.ee.outputs.available` stays empty.
2. **Read the prompt from the trusted base ref** — `REVIEW.md` and the prompt file are now read via `git show origin/<base>:...` instead of the merge checkout, so a fork can't rewrite the reviewer's own instructions/policy.
3. **Restrict the agent** — Codex runs `-s workspace-write` (network disabled) instead of `danger-full-access`; Pi drops the `bash` tool.

Same treatment applied to both Codex and Pi. Residual risk (provider API key must exist for the model call; indirect injection via the diff) is noted in the PR body.

rubenfiszel:
Addressed the latest Codex P0 (fork reviews could still publish model credentials via the review comment) in `6aab08d`.

The model call inherently needs the provider credential present, so a network-disabled sandbox alone can't stop a prompt-injected fork review from reading the key (Codex: `$HOME/.codex/auth.json`; Pi: `/proc/self/environ`) and emitting it in its final message — which both workflows posted verbatim, bypassing GitHub Actions log masking (masking only covers workflow logs, not API-posted comments).

Fix: **output filtering** on the post-comment step. Before posting, the review body is scrubbed of the known credential values:
- **Codex** — `OPENAI_API_KEY`, the raw `CODEX_AUTH_JSON`, and every long token string nested inside it (longest-match-first so no fragments survive).
- **Pi** — `DEEPSEEK_API_KEY`.

Applied unconditionally (not just the fork path) — a credential should never appear in a review comment regardless of trigger, and doing it always removes any reliance on the `is_fork` flag in the post step.

Residual risk (noted in the PR body): output filtering is literal-value based, so an injection that encodes/transforms the secret before emitting it would evade the scrub. That's inherent to any "credential must exist for the model call" design; the maintainer opt-in gate + network-disabled sandbox (Codex) / bash-less toolset (Pi) remain the primary containment.

rubenfiszel:
Addressed the new Codex P0 (persisted `github.token` in `.git/config`) in `c768af7`:

- **Primary fix:** `persist-credentials: false` on the merge-ref checkout in both workflows, so `actions/checkout` never writes `github.token` (issue/PR write) into `.git/config` where the review agent could read it. Safe on both paths — the only later git op is an unauthenticated fetch from the public `origin`, the EE checkout uses its own token, and `gh` uses `GH_TOKEN` from env.
- **Defense-in-depth:** the job token is now also redacted from the posted comment alongside the provider credentials.

rubenfiszel:
Addressed the Pi extension-execution P0 in `8a5f575`:

On the fork path Pi now runs with `--no-extensions` (disables discovery/execution of project-local `.pi` extensions, which would otherwise run fork `.ts/.js` at startup with `DEEPSEEK_API_KEY` in env, before the `--tools` allowlist applies), plus `--no-skills --no-prompt-templates --no-themes --no-context-files` so fork-controlled skills/templates/themes and `AGENTS.md`/`CLAUDE.md` can't be auto-loaded into the reviewer's prompt as an injection vector. Non-fork behavior is unchanged.

Codex runs `workspace-write` (no network) and has no analogous project-extension auto-execution, so the equivalent code-exec vector doesn't exist there.

rubenfiszel:
Addressed the `$GITHUB_OUTPUT` delimiter-injection P0 in `6b0b7e2` — nice catch, this was a real breakout.

The PR title/body heredocs now use a per-run **random** delimiter (128 bits from `/dev/urandom`) instead of the fixed `PR_BODY_EOF`, so a fork can't embed the terminator to close the block early and append `is_fork=false` (which, last-write-wins, would have re-enabled the EE checkout + `substitute_ee_code.sh` with the private token). Every other output in the block is single-line and newline-free, so this closes the vector. Verified with a simulated payload: `is_fork` stays `true` and the payload is retained verbatim as the `body` value.

rubenfiszel:
Addressed the Pi package-resolution P0 in `46cb514`:

Fork Pi reviews now run with `PI_OFFLINE=1`. `--no-extensions` only filters what gets *loaded*, but package resolution from a fork `.pi/settings.json` runs `npm install`/`npmCommand`/lifecycle scripts *before* that filter. With `PI_OFFLINE=1`, the resolver's `installMissing()` short-circuits (returns `false`) for every missing package, so no install/clone/lifecycle ever executes. It gates only startup network ops (installs, helper-binary downloads) — not the provider inference call — so the review still runs. Verified against pi v0.73.1 source (`resolvePackageSources` → `installMissing`) and with a fork `.pi/settings.json` fixture carrying a malicious `npmCommand`, which does not execute under the flag.

rubenfiszel:
Addressed the `.pi/SYSTEM.md`/`APPEND_SYSTEM.md` P0 at the root cause in `b6908cc`.

Rather than keep adding per-file opt-outs (Pi kept surfacing new discovery vectors — extensions, packages, then SYSTEM.md), I fixed the class: Pi resolves **all** project config from `<cwd>/.pi` (single-level, cwd-based — confirmed in `settings-manager`/`resource-loader` source; global fallback is the trusted runner home). So fork Pi reviews now run from a fresh `mktemp -d` where none of the fork's `.pi/*` is on the discovery path. Since the fork agent has no shell, the diff is pre-computed (trusted base…head SHAs) into the context file it reads; it may still read fork files by absolute path for extra context (reads are safe — only config discovery/code-exec were the risk). The `--no-*` flags and `PI_OFFLINE=1` stay as belt-and-suspenders. Verified a fork `.pi/SYSTEM.md` sentinel is not discovered from the isolated cwd.

Codex isn't affected by this class the same way: it reads its config from the trusted `CODEX_HOME`, runs network-disabled (`workspace-write`), takes its review prompt from the base ref, and its output is credential-redacted — and unlike Pi it needs the repo as its working root to run its own `git`/tooling, so isolating its cwd isn't applicable.

rubenfiszel:
Addressed the output-symlink RCE P0 in `45d8274` — fixed the whole class, not just the final-message paths.

Both workflows wrote generated files (final message, event stream, **and** the review context + prior-comments) into `$GITHUB_WORKSPACE`. On the fork path the merge tree is attacker-controlled, so a fork could commit any of those paths as a symlink and redirect our write to overwrite an action's code. All generated files now go through `$RUNNER_TEMP` (runner-created, outside the checkout):

- `prior-comments.json` and `pr-review-context.md` → `$RUNNER_TEMP`.
- The agent gets the context file's **absolute** `$RUNNER_TEMP` path (appended to the prompt); the `pr-review.prompt.md` files were updated to reference it instead of a checkout-relative path. Pi (no shell on forks) gets the diff pre-computed into that context file; the isolated-cwd hardening is retained.
- Codex `-o` and Pi's events/final-message → `$RUNNER_TEMP`; both post steps read from there.

Non-fork behavior is functionally unchanged. Verified YAML, embedded bash/Node syntax, and the fork-path file wiring (isolated empty cwd, context outside the checkout with the diff embedded, prompt carries the absolute path).

PR #10070, opened 2026-07-14
fix(apps): cover script/flow component outputs in deployed-app S3 provenance gate

rubenfiszel:
## Summary

Fixes a regression from #10048 (deployed apps read S3 on-behalf of the app author for logged-in viewers). That PR added a confused-deputy guard confining author-mode reads to files the app "produced", but the recent-production check only matched inline `appscript`/`preview` jobs nested under the app path. Files produced by the **deployed script/flow components an app is wired to run** — e.g. a SQL query that persists its results to S3, the common pattern — were denied `File restricted` for **every** viewer, admins included (the gate has no admin bypass).

Symptom reported: after upgrading to 1.756.0, a deployed app rendering S3-backed SQL results returns "File Restricted" in both operator and admin accounts, where previously the admin account worked.

## Root cause

`check_if_allowed_to_access_s3_file_from_app` (`backend/windmill-api/src/apps.rs`) modelled "produced by the app" as *an inline app-script job, nested under the app path, created by the caller*. That covers static `allowed_s3_keys` and inline-script outputs, but excludes:
- files produced by referenced **script/flow** components (`kind IN ('script','flow',...)`, `runnable_path` = the runnable's own path, not under the app path), and
- files produced **on-behalf of the author** (`created_by` = the author/service account, not the viewer).

## Changes

- Derive the app's triggerable **script/flow runnable paths** from `policy.triggerables` / `triggerables_v2` (stripping the optional `<component>:` prefix and the `script/`/`flow/` prefix; `rawscript/` inline keys stay covered by the appscript branch).
- Expand the provenance SQL to also match completed `script`/`flow`/`flowscript`/`flownode` jobs whose `runnable_path` is one of those declared triggerables.
- Accept the author identity via `permissioned_as = on_behalf_of` in addition to `created_by = caller`, so on-behalf-produced files are covered.
- Reads outside the app's declared triggerables stay denied (confused-deputy guard preserved).
- Regression test in `backend/tests/app_s3_onbehalf.rs`.

Mirrors the WIN-2164 fix approach.

## Test plan

- [x] `cargo test --features enterprise,private,parquet --test app_s3_onbehalf` — new `test_deployed_app_s3_onbehalf_flow_script_provenance` **fails with `File restricted` before** the gate change and **passes after**; existing `test_deployed_app_s3_onbehalf_provenance` still passes.
- [x] Test asserts all four cases: script-produced key clears for a non-admin viewer, clears for an admin viewer, clears via `permissioned_as` match when `created_by` differs, and a non-triggered script's output stays denied.
- [x] Compiles under `enterprise,private,parquet,quickjs` (dev server rebuilt clean).
- Note: test is gated on the `parquet` feature; CI must run the parquet-enabled suite.

---
## Summary by cubic
Fix the S3 provenance gate so deployed apps and editor previews can access S3 files safely. Deployed runs use a real app-origin marker with per-viewer isolation; previews read as the caller and are never marked. App components now run as the app identity to avoid privilege confusion.

- **Bug Fixes**
  - Introduce `JobTriggerKind::App`; `execute_component` stamps `trigger_kind='app'` and `trigger=<app path>` for deployed runs only (never previews). OpenAPI enum updated; suspended-trigger reassignment rejects `app` with a 400.
  - Authorize S3 reads when a recent (<=3h) completed app-marked job for this app path and `created_by=<caller>` produced the key; cross-viewer, other-app, and direct-run outputs stay denied.
  - App components execute on-behalf of the app identity (author in author-mode, caller in viewer/preview), never the referenced script/flow’s `on_behalf_of`; prevents escalation and keeps preview outputs readable as the caller.
  - Frontend: editor/preview S3 reads now use the viewer-scoped `job_helpers/download_s3_file` endpoint; removed reliance on preview provenance and `force_viewer_allowed_s3_keys`.
  - UI: surface the `app` trigger kind in Runs (filterable, with icon and label) so app-launched jobs render correctly.

<sup>Written for commit 0b5bd2b502d65cfb38ea04534599d63bbd619802. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the review in ef7d5b5:

**[P1] out-of-app runnable executions could forge app provenance** — fixed, and it was broader than the one branch: the recent-production check keyed on `created_by = caller`, so a viewer who can run a declared script/flow directly (outside the app, with un-pinned inputs) could craft `{"s3":"<author-only-key>"}` and read it back as the author. Provenance is now keyed on the producing job's `permissioned_as` matching the on-behalf identity the download reads as (the author in author-mode); a viewer's direct run has `permissioned_as = viewer` and no longer clears the gate. `created_by` is dropped from **both** the `appscript`/`preview` and script/flow branches, closing the same latent hole in the pre-existing inline-script branch. New assertion: a viewer's own direct run of a declared script stays `File restricted`.

**[P2] `flowscript`/`flownode` branch was dead** — fixed. Flow-step jobs have `runnable_path = <flow_path>/<step_id>`, which exact `= ANY(...)` never matched. Split script vs flow triggerable paths; flow kinds now match the flow's own job (bare path) and its step jobs via a `<flow_path>/%` prefix, bounded to declared flows. New assertion covers a flow-step-produced key.

**[P2] test used the unprefixed key format** — fixed. The test now declares triggerables in the production component-prefixed form (`<id>:script/...`, `<id>:flow/...`), exercising the prefix-stripping parser branch.

`.sqlx` regenerated (net: one new query cache; EE/other-feature caches preserved). Full `app_s3_onbehalf` suite passes (6 assertions incl. the new forgery-denied and flow-step cases).

rubenfiszel:
Second review round addressed in 6ce2e6b — both P1s are structural, so I replaced the identity/path heuristics with a real **app-origination marker**:

**[P1] on-behalf direct runs could still forge provenance** — correct: a runnable configured with its own `on_behalf_of` makes a direct `/jobs/run` resolve `permissioned_as` to that identity, so `permissioned_as` never proved app-origination. Fix: added `JobTriggerKind::App`; `execute_component` now stamps every app-launched job with `trigger_kind='app'` + `trigger=<app path>`. The gate's recent-production check is now just `trigger_kind='app' AND trigger=<this app path>` (+ 3h window + result containment). A direct `/jobs/run` cannot set `trigger_kind='app'`, so a viewer's own execution of a declared runnable — even one whose `permissioned_as` resolves to the author — is denied. Test asserts exactly that.

**[P1] flow-path wildcards admit unrelated flows** — moot: the `LIKE fp || '/%'` matching is gone entirely (no declared-path patterns left), so `_`/`%` in a flow path can no longer over-match.

Also: provenance is now scoped to **this** app's path, so another app's jobs (even same author) don't authorize this app's reads (new test asserts this too). Added `app` to the OpenAPI `JobTriggerKind` enum.

Notes for reviewers:
- Verified no exhaustive `match` on `JobTriggerKind` breaks (all have wildcard arms; the two other `match trigger_kind` sites use the distinct capture `TriggerKind`).
- No query keys on app jobs having `trigger_kind IS NULL` (the two hits are the capture-config filter and an asset-dispatch test, both unaffected by app tagging).
- Full `app_s3_onbehalf` suite passes (app-produced clears for viewer+admin; direct-run forgery denied; other-app output denied; existing declared-keys test unchanged).

rubenfiszel:
[P2] app advertised for suspended-trigger reassignment with no backing table — fixed in 48dab0f.

`get_suspended_trigger` now rejects `JobTriggerKind::App` alongside `webhook`/`schedule`, so the resume/cancel routes return a clean `400 "app triggers do not support job reassignment"` instead of failing on the missing `app_trigger` relation. Added a regression test asserting 400 (not 500) for `trigger_kind=app` on the reassign route.

Side observation (pre-existing, not introduced here): `asset` and `freshness` are in the same boat — they're valid `JobTriggerKind` values with no `<kind>_trigger` table, so they'd also 500 on this route today. Left them as-is to keep this PR scoped to the `app` variant it adds; happy to extend the guard (or switch to a table-backed allowlist) in a follow-up if you'd prefer.

rubenfiszel:
[P1] preview executions could forge the app-origination marker — fixed in 3a0b68f.

Correct: the marker was stamped unconditionally. A preview (`force_viewer_static_fields` set) with `raw_code` and no `path`/`id` skips all app authorization — only `jobs:run` + not-operator is required — yet ran against any app path in the URL, so it could produce `{"s3":"<author-only-key>"}` carrying `trigger_kind='app'` + `trigger=<victim app path>`.

Fix: stamp the marker only when `!is_preview` — i.e. only deployed, policy-checked runs are app-provenanced. Preview/editor S3 display doesn't rely on this marker (the editor reads S3 through the `force_viewer_allowed_s3_keys` allowlist), so nothing legitimate regresses.

Tests (both e2e, no worker needed — `execute_component` commits the job row):
- `test_execute_component_stamps_app_trigger`: a **deployed** run gets `trigger_kind='app'` + `trigger=<app path>`.
- `test_preview_component_is_not_stamped_app_trigger`: a **preview** run against a victim app path is NOT stamped (`trigger_kind` NULL).

Full `app_s3_onbehalf` suite: 5/5 green.

rubenfiszel:
[P1] preview-generated S3 outputs no longer downloadable in the editor — fixed in 55c1040, without reopening the preview forgery.

Correct: excluding *all* previews from the marker broke the editor's author-mode download for dynamically-connected preview results (display components only send `force_viewer_allowed_s3_keys` for static/upload configs). The distinguishing factor between a legit editor preview and a forgery is **whether the caller can edit that app**:

- `execute_component` now stamps a preview's marker only when the caller passes `require_is_writer` on the app path (admin / path owner / folder writer). An app editor already wields the app's author identity (they can deploy a component that reads the same file), so marking their own preview is no escalation and keeps preview results downloadable; a `jobs:run`-only caller who can't edit the app still can't forge the marker. Deployed runs unchanged.

Also added in the same commit (per discussion on the thread): **per-viewer isolation** — the gate now also requires `j.created_by = <this caller>`. The boundary is still the un-forgeable `trigger_kind='app'` marker; `created_by` is ANDed under it as an isolation filter (only narrows), so a viewer can only pull keys their own app runs produced, not another viewer's. Restores the per-caller scoping #10048 had, now safe on top of the marker.

Tests (parquet-gated): preview marked iff caller can edit the app; cross-viewer isolation (another viewer's app-marked key denied, no admin bypass); direct-run + other-app keys still denied; deployed run still stamped. Suite 5/5.

rubenfiszel:
[P1] scoped app writers can forge preview provenance without apps:write — fixed in 9c6ec05.

Correct — `require_is_writer` checks the ACL but not token scopes, so a writer's token scoped to `apps:run`/`apps:read`/`jobs:run` without `apps:write` could mark a preview even though it can't deploy the app (`update_app` requires `apps:write`), breaking the "can deploy equivalent code" rationale.

Fix: mark a preview only when **both** hold — `check_scopes(apps:write:<path>)` (the token is permitted to edit) **and** `require_is_writer` (the user actually is a writer). Together = a real editor of this app who could deploy the same read. Deployed runs unchanged.

Test: added a scope-restricted writer token (`apps:run`/`apps:read` + `jobs:run`, no `apps:write`) for the app owner and assert its preview stays unmarked (`trigger_kind` NULL → its S3 key stays gated); kept the full-editor positive case and the non-editor negative case. Suite 5/5.

rubenfiszel:
Simplified the preview handling in ca27722 (thanks for the push on this).

The insight: **a preview always executes as the caller (Viewer mode), never as the author.** So there's no reason a preview should ever be author-provenanced — its results should be read back as the caller. That removes the entire `require_is_writer` / `apps:write` / `can_preserve_on_behalf_of` line of reasoning, which was also unsound (being an app writer, even with `apps:write`, doesn't mean you can deploy a component running as the app's *on-behalf* identity — that additionally needs `can_preserve_on_behalf_of`/folder default, so marking such a caller's preview could still escalate to reading the original author's files).

- **Backend:** mark app-origination for deployed runs only (`!is_preview`).
- **Frontend:** `getS3File` (AppImage/AppPdf/AppDownload) now routes editor/preview reads through the viewer-scoped `job_helpers/download_s3_file` endpoint (reads as the caller) — matching what `DisplayResult`/`ParqetCsvTableRenderer` already do; only a deployed app view uses the provenance-gated `apps_u` endpoint. This is the path that previously relied on marking previews, so the earlier "editor can't download preview S3" regression is resolved by routing, not by minting author-mode provenance.

Verification: backend suite 5/5 (a preview — owner's own or a non-editor's — is never marked; cross-viewer isolation, deployed marking, reassignment guard unchanged); frontend `check:fast` clean. I could not browser-verify the `getS3File` routing end-to-end (the dev DB has no S3-configured app to drive), but it mirrors the established `DisplayResult` pattern.

rubenfiszel:
[P1] on-behalf preview outputs downloaded as wrong identity — root-caused and fixed in 5cb2aba.

You were right that the actual bug is upstream: `execute_component` was overriding the app's resolved on-behalf identity with the **referenced script/flow's own `on_behalf_of`** (`apps.rs:3171`). That's wrong in the app context — the app's execution mode should govern:
- A **Viewer-mode** app could execute a component *as the referenced runnable's on_behalf identity* instead of as the viewer (privilege confusion), and
- a **preview** would run as that identity rather than as the caller, so its S3 output couldn't be read back as the caller — the mismatch codex flagged.

Fix: an app component always runs on-behalf of the **app** identity (author in author-mode, caller in viewer/preview); a referenced runnable's own `on_behalf_of` no longer leaks into app execution. Direct `/jobs/run` still honors it (unchanged). With this, **previews always run as the caller**, so reading editor/preview S3 as the caller via viewer-scoped `job_helpers` is unconditionally correct — no on-behalf exception remains.

Test: the deployed-component e2e now seeds the script with a *distinct* on_behalf and asserts the component job's `permissioned_as` is the app identity (`u/test-user`), not the script's (`u/test-user-2`).

**[P2]** reworded the `configuration` param comment to describe current state only (no drafting history).

Backend suite 5/5; frontend `check:fast` clean.

rubenfiszel:
Nits addressed in 0b5bd2b: added `app` to the Runs trigger-kind filter + icon (`jobTriggerKinds`/`triggerIconMap`/display names, LayoutDashboard icon) so app-component jobs are filterable and render their trigger info; condensed the marker/on-behalf/gate comments to state each invariant once in ≤4 lines per AGENTS.md.

PR #10076, opened 2026-07-14
feat(sessions): support many pending sessions persisted in IndexedDB

Guilhem-lm:
## Summary

Today the sessions page keeps a single pending (unsent) AI session — clicking **+** just returns to the same in-memory draft. This PR lets you set up **several unsent sessions in parallel**, each with its own workspace/fork, composer prompt, and preview tabs, persisted in IndexedDB so they survive a reload.

The overloaded `transient` flag is split: it now means only *"in-memory, not yet written to IndexedDB"*. Whether a session is *unsent* is derived from `workspace_id` being undefined. A pending session is persisted on its **first genuine touch** (typing a prompt, picking a workspace/fork, opening a preview tab, renaming, dragging the split) — so idle **+** clicks never litter the sidebar and vanish on reload.

## Changes

- **Split `transient` semantics** in `sessionState.svelte.ts`: `transient` = not-yet-persisted; unsent is derived from `!workspace_id`. `putSession` no-ops for transient records; `persistTouched` promotes + writes on first touch.
- **Per-session `draftPrompt`**: each parallel pending session persists its own typed-but-unsent composer text (equality-guarded so an empty on-mount `onDraftChange('')` doesn't persist an untouched draft); cleared once the workspace commits at first send.
- **`+` reuses the untouched draft** in the active family instead of spawning a new blank each click; touching one spawns a fresh blank.
- **Pending sessions show in the sidebar** under the family filter (`SessionPicker` no longer hides transient records) and reconcile by `pending_workspace_id`.
- Preview tabs / collapse / split size and `SessionWrapper` draft round-trip wired through the new persist-on-touch path.
- Tests: rewrote the transient IndexedDB tests, added family-scoped reuse tests and pending-draft reconcile (delete/archive) tests.

## Screenshots

many-pending-sessions

## Test plan

- [ ] With `localStorage wm_dev_global_ai=1` and a workspace AI provider, open `/sessions`.
- [ ] Click **+** several times with no input → only one blank draft exists (no pile-up); reload → it's gone.
- [ ] Type a prompt / pick a fork in a draft, then **+** again → a fresh blank draft spawns; both coexist in the sidebar.
- [ ] Give two drafts different prompts, reload → each restores its own prompt.
- [ ] Delete/archive a pending draft's workspace → the persisted draft is reconciled accordingly.

### DIFF
--- a/frontend/src/lib/components/sessions/SessionPicker.svelte
+++ b/frontend/src/lib/components/sessions/SessionPicker.svelte
@@ -145,7 +145,8 @@
 	// total, and keyboard navigation.
 	const visibleSessions = $derived(
 		sessionState.sessions.filter((s) => {
-			if (s.transient) return false
+			// Pending (unsent) sessions show like any other, so several drafts can be
+			// set up in parallel; they group by pending_workspace_id via sessionRootOf.
 			// The open session always stays in the list, ignoring both filters.
 			if (s.id === sessionState.currentSessionId) return true
 			if (s.archived && !showArchived.val) return false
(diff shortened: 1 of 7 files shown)


Guilhem-lm:
Addressed the review findings from `auto-review` (P1) and `pi-review` (P2) in 3c97920ae2:

**`deleteSession` now cancels the pending draft-prompt flush timer.** As flagged, deleting a draft inside the 400 ms debounce window left the `setTimeout` running; it fired afterward and `persistTouched` wrote the deleted record back to IndexedDB, resurrecting the draft on reload. `deleteSession` now `clearTimeout`s and drops the per-session `draftPromptFlushHandles` entry before removing the record. Added a regression test (`deleting a draft inside the debounce window does not resurrect it`); full session suite (183) + `svelte-check` green.

PR #10087, opened 2026-07-14
fix(sessions): pending-draft debounce follow-ups (delete-cancel, keystroke de-transient, teardown count)

Guilhem-lm:
## Summary

Follow-ups to #10076 (merged) for the per-session draft-prompt debounce. One fix
missed the squash-merge by ~30s; the other two were surfaced by a cold `codex`
review of the branch (both are real issues in the already-merged feature).

`setSessionDraftPrompt` schedules a 400 ms debounced IndexedDB write per session in
`draftPromptFlushHandles`. Three problems around that window:

1. **Delete-resurrect** — `deleteSession` removed the record from memory and
   IndexedDB but never cancelled the pending timer, so deleting a pending draft
   *inside* the debounce window let the flush fire afterward and write the record
   back, resurrecting the deleted draft on reload. (Flagged on #10076 by
   `codex-review` P1, `claude` P1, `pi-review` P2; the fix was pushed just after the
   squash-merge and missed the train.)
2. **Reuse race (P1)** — `createSession` treats every `transient` session as a
   reusable blank, but a keystroke only cleared `transient` 400 ms later via the
   debounced flush. Typing then pressing `+` within that window reopened the same
   draft instead of spawning a second pending session.
3. **Teardown undercount (P2)** — `countSessionsForWorkspace` counted only
   `workspace_id`, but `reconcileSessionsLifecycle` archives/deletes persisted
   unsent drafts on `workspace_id ?? pending_workspace_id`. A workspace holding only
   pending drafts undercounted in the archive/delete confirmation.

## Changes

- `deleteSession` now `clearTimeout`s and drops the session's `draftPromptFlushHandles`
  entry before removing the record.
- `setSessionDraftPrompt` clears `transient` **synchronously** on a genuine keystroke;
  only the IndexedDB write stays debounced (`putSession` directly).
- `countSessionsForWorkspace` counts on `workspace_id ?? pending_workspace_id`, matching
  the reconcile lifecycle.
- Regression tests for all three: delete-inside-window, keystroke de-transient, and the
  pending-draft teardown count.

## Test plan

- [x] `vitest run` on both session suites — 46/46 pass, including the three new tests
      (each fails without its fix).
- [x] `svelte-check` clean.

No visible UI effect — data-layer timer/count fixes, so no screenshots.

### DIFF
--- a/frontend/src/lib/components/sessions/sessionState.svelte.ts
+++ b/frontend/src/lib/components/sessions/sessionState.svelte.ts
@@ -306,11 +306,10 @@ export function setSessionDraftPrompt(sessionId: string, text: string): void {
 	// mount-time onDraftChange('') as a non-touch (draftPrompt is undefined),
 	// so merely opening an untouched draft never persists it.
 	if ((s.draftPrompt ?? '') === text) return
-	// Keep `transient` set until the flush persists the draft: it means "in-memory
-	// only", so hydrateSessions preserves the draft across a reconcile that lands
-	// inside this window (clearing it early would drop a not-yet-written draft).
-	// createSession stops reusing it the moment draftPrompt is non-empty instead —
-	// see isReusableBlank. Only the IndexedDB write is debounced.
+	// Keep `transient` (means "in-memory only") set until the flush persists the
+	// draft, so hydrateSessions preserves it across a reconcile inside this window;
+	// isReusableBlank, not `transient`, is what stops createSession reusing a typed
+	// draft. Only the IndexedDB write is debounced.
 	s.draftPrompt = text
 	clearTimeout(draftPromptFlushHandles.get(sessionId))
 	draftPromptFlushHandles.set(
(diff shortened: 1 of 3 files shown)


Guilhem-lm:
Addressed the codex-review P1 (reconciliation dropping a draft mid-debounce) in 778c3c6.

Root cause: the earlier fix cleared `transient` synchronously to stop `createSession` reusing a just-typed draft, but `transient` also means "in-memory only, not yet persisted" — clearing it before the 400ms flush left the draft in neither `hydrateSessions` bucket (in-memory transients + DB rows), so a reconcile inside the window dropped it and dangled `currentSessionId`.

Fix separates the two concepts: `transient` stays pure persistence state (draft survives hydration), and a *reusable blank* is now `transient && !draftPrompt` (`isReusableBlank`). `createSession`'s reuse probe and its non-reuse drop both key on that, so a typed-but-unflushed draft is neither reused nor discarded; `setSessionDraftPrompt` no longer touches `transient`.

Added a regression test interleaving a first-touch debounce with `reconcileSessionsLifecycle` (draft stays in memory, `currentSessionId` intact). 47/47 session tests pass, `svelte-check` clean.

The **P2**/teardown-count and delete-cancel fixes are unchanged. Pi and Claude reviews were "Good to merge".

Guilhem-lm:
Fixed the P2 nit in 45b7d91: `isReusableBlank` now keys on `draftPrompt === undefined` (never edited) instead of falsiness, so a typed-then-erased draft (`''`, flush pending) stays a real session — consistent across the debounce boundary and not droppable/resurrectable on a family switch. Added a type-then-erase regression test (48/48 session tests pass, svelte-check clean).

PR #10090, opened 2026-07-14
feat(ai-chat): port flow-group and sticky-note instructions to global chat

Guilhem-lm:
## Summary

The global-mode AI chat had the *wrong* (really: missing) instructions for flow groups. It inherited only the bare `FlowGroup` **schema** from the shared flow prompt, but none of flow mode's **behavioral** guidance — so it never proactively segmented flows into groups — and it had **zero sticky-note support**. Flow mode carries this guidance inline in its own prompt and `set_flow_json` tool.

This PR brings global mode to parity with flow mode for both `groups` and `notes`.

## Changes

- **`write_flow` schema**: enriched the `groups` description (exact color palette + `summary/note/autocollapse/start_id/end_id` fields) and **added a `notes` field**, mirroring flow mode's `set_flow_json`.
- **Notes plumbing**: threaded `notes` through `editableFlowToDraftValue` (into `FlowValue.value`) and the `write_flow` handler, so notes reach the flow value and survive the deploy round-trip (`FLOW_SPEC.buildDraft` `structuredClone`s `value`). Reads and `patch_flow_json` already carried notes via the shared `editableFlowJson` helpers — only the write path was missing.
- **`getFlowInstructions()`**: added an "Organizing flows: groups and notes" section (strongly-recommended proactive grouping, color palette, `start_id`/`end_id` constraints) and a "When to use notes vs groups" section; mentioned `notes` in the write/read/compact-view/structural-edit bullets.
- **Test**: added a `write_flow` → `read_workspace_item` notes round-trip case.

## No UI screenshots
The diff is under `frontend/` but is purely copilot chat prompt/tool-logic in `.ts` files — there is **no static UI surface** to screenshot. The visible effect is AI chat behavior (the model authoring groups/notes), which requires a live AI-provider session to demonstrate.

## Test plan

- [x] `npm run check:fast` — clean (one pre-existing unrelated error in `utils_workspace_deploy.ts`)
- [x] Global core tests — 137 pass (incl. new notes round-trip)
- [ ] **`ai_evals` A/B on the `global` mode** (before/after, same cases) — required by the `ai-chat` skill for behavior changes; not yet run (needs a provider). Ideally land a global-mode case covering notes/groups.
- [ ] Live global-chat session: ask it to author a flow with a flow-wide sticky note → open the draft in the editor (note renders, auto-placed, correct palette color) → deploy and reopen (note persists); confirm it proactively segments non-trivial flows into `groups`.

## Review notes
Local review found no P0/P1. Two P2 nits, both accepted-as-is: (1) the `ai_evals` A/B above; (2) the `notes` schema description is verbose (per-iteration token cost) and duplicates `getFlowInstructions` prose — kept **intentionally** for flow-mode parity, which the reviewer agreed makes it low-priority.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/global/core.test.ts
+++ b/frontend/src/lib/components/copilot/chat/global/core.test.ts
@@ -2760,6 +2760,45 @@ describe('global AI tools', () => {
 		expect(item.value.value).toBeUndefined()
 	})
 
+	it('writes and reads back free-floating flow notes', async () => {
+		const writeResult = JSON.parse(
+			await callGlobalTool('write_flow', {
+				path: 'f/flows/with-notes',
+				summary: 'Flow with notes',
+				modules: JSON.stringify([
+					{
+						id: 'start',
+						summary: 'Start',
+						value: { type: 'identity' }
+					}
+				]),
+				notes: JSON.stringify([
+					{ id: 'n1', type: 'free', text: 'What this flow does', color: 'blue' }
+				])
+			})
+		)
+
+		expect(writeResult.success).toBe(true)
+
+		const item = JSON.parse(
+			await callGlobalTool('read_workspace_item', {
+				type: 'flow',
+				path: 'f/flows/with-notes'
+			})
+		)
+
+		expect(item.value.notes).toHaveLength(1)
+		expect(item.value.notes[0]).toMatchObject({
+			id: 'n1',
+			type: 'free',
+			text: 'What this flow does',
+			color: 'blue'
+		})
+		// Free notes with no explicit geometry get auto-placed/sized by validation.
+		expect(item.value.notes[0].position).toBeDefined()
+		expect(item.value.notes[0].size).toBeDefined()
+	})
+
 	it('test_run_script previews draft script content by path', async () => {
 		const content = 'export async function main(name: string) {\n\treturn `hello ${name}`\n}'
 		await callGlobalTool('write_script', {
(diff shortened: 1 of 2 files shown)


Guilhem-lm:
Addressed the P2 duplication finding from both reviews in 6193120205: trimmed the `write_flow` `groups`/`notes` tool-schema descriptions to the correctness-critical bits (color palette, `type: "free"`, `null` semantics) and pointed to `get_instructions subject:"flow"` for the full field reference — which `getFlowInstructions()` already carries. This removes the per-iteration token tax on the tool schema while keeping the palette inline so invalid-color tool calls still fail fast.

Kept the palette in the schema deliberately (rather than trimming fully) so a `write_flow` call that skips `get_instructions` still can't emit hex colors.

Remaining open item: the `ai_evals` global-mode A/B (both reviewers noted it) — process step that needs a provider run; tracked on the PR checklist.

`check:fast` clean (one pre-existing unrelated error in `utils_workspace_deploy.ts`); global core tests 137 pass.

Guilhem-lm:
Re the recurring `ai_evals` A/B note: the **AI Evals (global mode)** CI workflow ran and **passed** on `d71b3f01` (the commit that introduced the `notes` capability + "Organizing flows" instructions), so the global-mode before/after is covered automatically in CI rather than being an open manual gap. The follow-up trim commit only shortened tool-schema prose (deferring to `get_instructions`), no behavior change.

Pi verdicts **Good to merge** on the current HEAD; Codex was Good to merge on `d71b3f01` and its re-run on the current HEAD is still in progress.

PR #10091, opened 2026-07-14
fix(ai-chat): size AI-created flow notes to fit their text

Guilhem-lm:
## Summary
When the flow AI chat creates a sticky note, the tool prompt tells the model to omit `size` and let the editor size it. But `validateFlowNotes` seeded a fixed **275×60px** box for every geometry-less free note, and free notes never grow to fit their content (only *group* notes drive layout from measured text height in `computeNoteNodes`). The result: any multi-line markdown note spilled out below its 60px box.

This seeds a text-proportional height when the AI omits `size`, so AI-created notes contain their text.

## Changes
- Add `estimateFreeNoteSize(text)` (module-private) in `helperUtils.ts`: wraps each source line to the content width (`text-xs` ≈ 17px/line, `p-4` padding) and clamps height to `[MIN_NOTE_HEIGHT (60), 600]`; width stays `MIN_NOTE_WIDTH`.
- `validateFlowNotes` uses it for free notes with omitted `size` (notes with explicit AI-provided or user-resized geometry are untouched).
- Tests: long text → taller box than short; short clamps to 60; huge clamps to 600.

## Notes / limitations
- The estimate is intentionally markdown-unaware (flat 17px/line on source characters), so notes dominated by markdown chrome (headers, code blocks) may still be slightly under-sized — render mode uses `overflow-auto` so content scrolls rather than being lost. This is a best-effort seed; the user can still resize.

## Test plan
- [x] `npx vitest run helperUtils.test.ts` — 44/44 pass
- [ ] In the flow editor, ask the AI to create a note with ~15–20 lines of markdown (incl. a couple of `#` headers) → box is tall enough to read without an immediate scrollbar
- [ ] Ask the AI for a one-line note → still renders at the compact default size

## Screenshots
No screenshot captured: exercising this path end-to-end requires a workspace with a configured AI provider driving the flow chat to emit a note, which isn't set up in the dev environment here. The change is a deterministic pure-function fix on note geometry, verified via unit tests and by reading the render path (`NoteNode.svelte`, `noteUtils.svelte.ts` `computeNoteNodes`) directly. The manual browser checks above are left in the test plan for a reviewer with an AI-enabled workspace.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/flow/helperUtils.test.ts
+++ b/frontend/src/lib/components/copilot/chat/flow/helperUtils.test.ts
@@ -476,6 +476,25 @@ describe('validateFlowNotes', () => {
 		expect(notes[0].position!.y).not.toEqual(notes[1].position!.y)
 	})
 
+	it('stacks auto-placed notes by their real heights so tall notes do not overlap', () => {
+		const longText = Array.from({ length: 20 }, (_, i) => `Line ${i} of note content`).join('\n')
+		const notes = validateFlowNotes([
+			{ id: 'a', text: longText },
+			{ id: 'b', text: 'short' }
+		])!
+		// Second note must start at or below the bottom of the first (tall) note.
+		expect(notes[1].position!.y).toBeGreaterThanOrEqual(
+			notes[0].position!.y + notes[0].size!.height
+		)
+	})
+
+	it('grows a note whose single source line wraps across many display lines', () => {
+		const short = validateFlowNotes([{ id: 's', text: 'hi' }])![0]
+		const oneLongLine = 'word '.repeat(200).trim() // no newlines, wraps many times
+		const wrapped = validateFlowNotes([{ id: 'w', text: oneLongLine }])![0]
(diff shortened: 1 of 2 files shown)


Guilhem-lm:
Addressed the Codex P2 (auto-placed notes could overlap): free notes are now sized first, then auto-placed against a running y-cursor that advances by each note's real height, so consecutive generated notes stack without overlapping. Added tests for two-tall-note non-overlap and single-line wrapping (46/46 pass). Pushed in `9bb1a0cf99`.

Guilhem-lm:
Addressed the second Codex P2 (mixed preserved + auto-placed notes could overlap): a preserved note occupying the auto-stack column now advances the y-cursor, so a later geometry-less note stacks below it instead of on top. Added a mixed existing-geometry/new-note regression test (47/47 pass). Pushed in `a063d2198c`.

Guilhem-lm:
Addressed the Codex P2: trimmed the `estimateFreeNoteSize` doc comment to the single non-obvious constraint (free notes render at fixed height and don't grow), dropping the implementation narration per AGENTS.md. Pushed in `5d2f36480c`.

PR #10092, opened 2026-07-14
[ee] fix: cap job token authority at workspace admin

hugocasa:
## What

Caps the authority of a job token (`$WM_TOKEN`) at **workspace admin**. A job token runs as its app/flow/schedule `on_behalf_of` identity, so its privileges must not exceed workspace admin regardless of the identity it runs as.

- `ApiAuthed` carries `token_is_job` (set from the JWT `job_id` claim).
- `require_super_admin` / `is_super_admin` / `require_devops_role` now take `&ApiAuthed` and deny job tokens on superadmin/devops-gated endpoints — a single choke point rather than a per-route denylist. `require_super_admin_email` remains for internal (non-request) callers.
- Inline `is_super_admin_email(authed.email)` capability gates migrated to the cap-aware `is_super_admin(db, &authed)`.
- The old `forbid_superadmin_job_token` denylist is now redundant wherever `require_super_admin` runs first, so it's removed from those routes and kept only on the self-service ones with no `require_super_admin` (`create_token`, `set_password`, `create_user`).
- Defense in depth: `validate_on_behalf_of` rejects reserved internal identities (the `superadmin_secret@`/`_notification@`/`_sync@` sentinels) when persisting an app/flow/script/schedule/trigger; app execution refuses reserved-identity policies. This does not restrict deploying on behalf of a real user (including a real superadmin, e.g. git-sync of superadmin-authored content) — the cap handles that at execution.

## Testing

- `preserve_on_behalf_of` (incl. new `test_reject_forged_superadmin_on_behalf_of`) — pass
- `wm_token_superadmin_guard` (incl. new superadmin + devops route-cap checks) — pass
- Full CE and `enterprise,private,license` builds green

## Notes

- Companion EE PR: windmill-labs/windmill-ee-private#661
- Any **new** superadmin/devops route must use `require_super_admin(db, &authed)` / `is_super_admin(db, &authed)` / `require_devops_role(db, &authed)` — never the `_email` / `is_super_admin_email` / `is_devops_email` forms, which bypass the cap.


hugocasa:
Superseded by #10124, which took the same token-layer approach with a better design (central `job_id` stamping in `get_opt_job_authed`) and additionally covers `CUSTOM_INSTANCE_DB` and restricted worker tags.

The pieces from this branch that #10124 lacked (the `require_devops_role` cap, fail-closed `job_id` parsing, and the reserved-sentinel store/execution guard) have been folded into #10124 as 01ddbff6f6. Nothing here is lost; branch retained.

PR #10093, opened 2026-07-14
feat(dev-workspace): reflect existing protection rules in lock toggles

hugocasa:
## Summary

When creating or attaching a dev workspace, the two lock toggles ("Block direct edits" → `DisableDirectDeployment`, "Prevent forking" → `DisableWorkspaceForking`) apply protection rules to the root/prod workspace. Previously they were always presented from a blank slate: on by default and fully editable, regardless of what the root already enforced. That made the toggles easy to misread — flipping them silently mutated the root, and if the root was *already* locked by an existing rule, an editable "off" toggle implied it could unblock something it can't (this flow only manages its own reserved `dev_workspace_lock` rule).

This makes the toggles aware of the root's current protection state. A restriction that's already enforced by an existing rule is shown **on but locked**, with a note, instead of offering a fresh default.

## Changes

- Both flows now fetch the root's protection rules (`fetchProtectionRulesForWorkspace`, fail-open on error), gated to fire only while the relevant form is visible:
  - `CreateWorkspaceInner.svelte` — the "Create a new dev workspace" fork flow.
  - `DevWorkspaceSetting.svelte` — the "Attach an existing workspace as dev" flow in workspace settings.
- Each lock toggle renders two ways:
  - **Not already enforced** → on by default, fully interactive (unchanged).
  - **Already enforced by an existing rule** → shown on, locked (disabled/grayed), with the note *"Already enforced by an existing protection rule."*
- The value sent to the backend is a `$derived` effective value (`alreadyBlocks || userToggle`), so the request always matches what the locked toggle shows.
- Kept both toggles defaulting to on (only the presentation/awareness changed, not the default).

## Screenshots

Set up a root workspace with a manual `DisableDirectDeployment` rule and no forking rule, to show the mixed state (deploy locked, forking still interactive) in one screen.

**Fork flow — "Create a new dev workspace":**

fork-dev-workspace-aware

**Attach flow — workspace settings → Dev workspace:**

attach-dev-workspace-aware

## Test plan

- [ ] Root with an existing rule blocking direct deployment: both flows show "Block direct edits" on + locked with the note, and "Prevent forking" interactive + on by default.
- [ ] Root with no protection rules: both toggles interactive and on by default (unchanged).
- [ ] The locked toggle can't be turned off; the interactive one can.
- [ ] Fetch failure falls back to the editable default-on toggle (protection still enforced server-side by the real rule).
- [ ] Backend contract: attaching/creating with `lock_prod_*: true` when an equivalent user rule already exists does not error (reserved rule is upserted by name), and detach removes only the reserved rule, leaving the pre-existing rule intact.

---
## Summary by cubic
Dev workspace lock toggles now mirror the root workspace’s protection rules and honor bypasses. If a restriction is enforced—or rules are still loading—the toggle shows on and disabled, and requests use the effective value.

- **New Features**
  - Fetch root protection rules while the form is visible via `fetchProtectionRulesForWorkspace`; on failure, fall back to editable default-on toggles (real rules still enforce).
  - Show toggles locked (on + disabled) with a note when an existing rule unconditionally enforces `DisableDirectDeployment` or `DisableWorkspaceForking` (no bypass users/groups); send derived values so the backend matches the UI in both create and attach flows.

- **Bug Fixes**
  - Lock toggles during the rules-loading window to prevent submitting false before rules resolve.
  - Guard stale fetches when switching the base/selected workspace by tagging results and ignoring superseded responses.
  - Align the “Already enforced...” note under the toggle label for clearer UI.

<sup>Written for commit 90b277d71fdab56959cf72c7a6ff82cb9f25c984. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10093?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/DevWorkspaceSetting.svelte
+++ b/frontend/src/lib/components/DevWorkspaceSetting.svelte
@@ -273,7 +273,7 @@
 					}}
 				/>
 				{#if alreadyBlocksDeploy}
-					<span class="text-2xs text-secondary ml-8"
+					<span class="text-2xs text-secondary ml-11"
 						>Already enforced by an existing protection rule</span
 					>
 				{/if}
@@ -290,7 +290,7 @@
 			<div class="flex flex-col gap-0.5">
 				<Toggle checked disabled options={{ right: 'Prevent forking this workspace' }} />
 				{#if alreadyBlocksForking}
-					<span class="text-2xs text-secondary ml-8"
+					<span class="text-2xs text-secondary ml-11"
 						>Already enforced by an existing protection rule</span
 					>
 				{/if}
--- a/frontend/src/lib/components/DevWorkspaceSetting.svelte
+++ b/frontend/src/lib/components/DevWorkspaceSetting.svelte
@@ -73,10 +73,20 @@
 	let alreadyBlocksForking = $derived(
 		isRuleActiveInRulesets(rootProtectionRules.current ?? [], 'DisableWorkspaceForking')
 	)
-	// Sent to the backend: an already-enforced restriction stays on regardless of the (locked) toggle's
-	// raw state, keeping the request consistent with what the locked toggle shows.
-	let effectiveLockProdDeploy = $derived(alreadyBlocksDeploy || lockProdDeploy)
-	let effectiveLockProdForking = $derived(alreadyBlocksForking || lockProdForking)
+	// Until the fetch resolves the workspace's rules are unknown (current === undefined also covers the
+	// first frame before loading flips). Treat each lock as engaged during that window so the toggle is
+	// locked on and the effective value stays true: otherwise a user could turn a lock off and attach
+	// before an existing rule is detected, sending false and omitting the reserved rule — leaving prod
+	// unprotected if that existing rule is later removed.
+	let rulesUnknown = $derived(
+		rootProtectionRules.loading || rootProtectionRules.current === undefined
+	)
+	let deployLocked = $derived(alreadyBlocksDeploy || rulesUnknown)
+	let forkingLocked = $derived(alreadyBlocksForking || rulesUnknown)
+	// Sent to the backend: a locked restriction (enforced or not-yet-known) stays on regardless of the
+	// toggle's raw state, keeping the request consistent with what the locked toggle shows.
+	let effectiveLockProdDeploy = $derived(deployLocked || lockProdDeploy)
+	let effectiveLockProdForking = $derived(forkingLocked || lockProdForking)
 
 	// A standalone root workspace, or an existing fork of this prod (same family), can be attached.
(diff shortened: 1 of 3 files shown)


hugocasa:
Addressed a P1 from the codex review: **forms could submit before the protection-rule fetch resolved.**

During the async load window (and the first frame before `loading` flips), `alreadyBlocks*` was `false`, so a lock toggle was editable — a user could turn it off and submit before an existing rule was detected, sending `lock_prod_*: false` and omitting the reserved `dev_workspace_lock` rule. Prod looked protected (the user's own rule still applied) but would silently lose protection if that rule were later removed.

Fix (`c4589e9`): treat "rules not yet known" (`loading || current === undefined`) the same as "already enforced" — the toggle is locked on and the effective value stays `true` during that window, so the request can never submit `false` before the fetch resolves. Using `current === undefined` also covers the initial frame and the stale-value flash when switching base workspace. Submission stays available, so a hung fetch degrades to over-protection rather than a blocked form.

Verified: a workspace with an existing deploy rule resolves to deploy-locked+note / forking-interactive; a clean workspace resolves both toggles to interactive+on (not stuck locked).

hugocasa:
Addressed two more P1s from the codex review (`567f821`):

**1. Bypassable rules became unconditional locks.** `alreadyBlocks*` used `isRuleActiveInRulesets`, which ignores `bypass_users`/`bypass_groups`, and forced the request flag to `true`. Since the reserved `dev_workspace_lock` rule is created with *empty* bypass lists, layering it over an existing rule that let specific users through would revoke their deploy/forking access (enforcement blocks a user unless they bypass *every* matching ruleset). Fix: new `isRuleUnconditionallyActiveInRulesets` predicate — a toggle is only shown as already-enforced (locked) when an existing rule has no bypasses; a bypassable rule stays **editable**, so applying the empty-bypass lock becomes the user's explicit choice.

**2. Stale protection fetch could apply another base's rules.** The generated client returns a `CancelablePromise` and can't take an abort signal, so a delayed response for a previous base could overwrite the newly selected one. Fix: tag each result with its workspace and only trust a result matching the current base (`rootRules`), and throw `AbortError` from a superseded fetch (using runed's abort signal) so it can't overwrite `current`.

Verified in-browser:
- Rule with a bypass user → toggle **editable**, no note.
- Unconditional rule (no bypass) → toggle **locked + note**.
- No rules → editable.

svelte-check + svelte autofixer clean.

PR #10102, opened 2026-07-14
fix(apps): deep-match provenance + declarable S3 read scopes for on-behalf app reads

rubenfiszel:
## Summary

Deployed apps that read S3 files on-behalf of their author/service-account (#10048, #10070) could still return **"File restricted"** for every viewer — admins included — when the file wasn't produced *inside that viewer's own app session*. The confused-deputy provenance gate (`check_if_allowed_to_access_s3_file_from_app`) only cleared a download via exact `allowed_s3_keys` or a recent app-triggered job with a **top-level** `result @> {"s3":key}`.

Reproduced on EE v1.757.0 (which contains #10070): a Publisher-mode app on-behalf of a service account displaying a stored SQL result → `File restricted` for viewer and admin, while editor preview works (preview forces `Viewer` mode, skipping the gate). Two distinct gaps:

1. **Nested results** — a runnable that returns its S3 object nested (`[{"s3":…}]`, `{"file":{"s3":…}}`, a query rich-result) never matched the top-level `@>`.
2. **Pre-existing files** — files produced *outside* the viewer's session (e.g. a schedule/CI job persisting results to S3) have no per-viewer provenance to match.

## Changes

- **Deep provenance match** (`apps.rs`): the provenance check now matches the S3 key anywhere in the job result via `jsonb_path_exists(c.result, '$.** ? (@.s3 == $k)', …)` instead of top-level `@>`. Still requires a genuine app-origination marker (`trigger_kind='app'` + `trigger=<app>` + `created_by=<caller>`) — unforgeable via `/jobs/run`.
- **`Policy.s3_read_scopes`** (new): author-declared literal globs (globset, e.g. `f/sensitive/**`) of keys the app may read on-behalf, for files it displays but doesn't produce. OR'd into the gate *before* the DB query. Kept a **separate** field from `allowed_s3_keys` (which `updatePolicy` recomputes each deploy, and whose exact-match semantics must not change to glob). Literal globs only — no `{folder_read}` template expansion; primary-storage-scoped.
- **No permission bypass**: a scope/whitelist match only decides whether to proceed — the read still runs on-behalf of `on_behalf_of` through `get_workspace_s3_resource_and_check_paths`, enforcing that identity's advanced S3 permissions. Verified: an in-scope key the on-behalf identity *cannot* read returns `401`, not the file.
- **Frontend**: a **discreet, collapsed-by-default** disclosure ("S3 read paths (on-behalf)") in the app deploy panel, shown for non-viewer execution modes; auto-expands (with an `(N)` count) only when scopes already exist. Lives in the shared `AppEditorHeaderDeploy`, so it covers **both low-code and raw apps**.
- **Tests**: `test_deployed_app_s3_nested_result_provenance`, `test_deployed_app_s3_read_scopes`.
- openapi + sqlx offline cache updated (EE caches preserved). CE-only — no `*_ee` changes.

## Screenshots

Collapsed by default (most apps never need it):

collapsed

Expanded:

expanded

## Test plan

- [x] New backend regression tests pass (`cargo test --features parquet --test app_s3_onbehalf`)
- [x] `check:fast` clean (0 svelte warnings)
- [x] Repro (EE, filesystem storage + default advanced perms, on-behalf svc): nested app-produced files `File restricted` → `200`; pre-existing file in `s3_read_scopes` → `200` for viewer & admin; out-of-scope → denied; in-scope but on-behalf-unreadable → `401`
- [x] UI round-trip (add/edit/remove path → deploy → persisted → viewer download succeeds); collapse/expand disclosure verified in-browser


rubenfiszel:
Closing in favor of an existing mechanism: apps should **sign** the S3 objects they display via `wmill.sign_s3_object(s)` / `sign_s3_objects` (TS: `signS3Object(s)` / `signS3Objects`). Signing runs on-behalf of the app's identity, is authorized by that identity's own read permission, mints a short-lived unforgeable capability, and the download gate already honors it — so it covers both live-produced and pre-existing/displayed files without the new `s3_read_scopes` policy surface, and direct API access stays blocked by advanced-permission rules. The `s3_read_scopes` field here is redundant with that; the nested-result (`jsonb_path_exists`) provenance fix can be revisited as a small standalone bug fix if we want the un-signed default path to be more robust.

PR #10110, opened 2026-07-14
perf(rls): wrap session GUC reads in RLS policies for per-statement InitPlan (GIT-919)

rubenfiszel:
## Summary

RLS policies read `current_setting('session.user' / 'session.groups' / 'session.pgroups' / 'session.folders_read' / 'session.folders_write')` **unwrapped** inside their `USING` / `WITH CHECK` predicates. Postgres treats those calls as potentially row-varying and **re-evaluates them once per scanned row**, on the read path of every workspace-scoped object + trigger table.

Wrapping each session-derived subexpression in a scalar sub-select — `(select current_setting(…))` — lets the planner hoist it to a **one-time InitPlan** (evaluated once per statement, reused for every row). This is a pure planner optimization: same rows in, same rows out, N per-row GUC lookups collapse to 1.

Fixes GIT-919. Surfaced by [pgrls](https://github.com/pgrls/pgrls) (`PERF001`).

## Why it's safe

The session GUCs are set with `set_config('session.*', …, true)` (the `SET LOCAL` form) in `set_session_context()`, so they are constant for the duration of a statement — exactly the condition under which the InitPlan hoist is value-preserving. Same optimization Supabase documents for `auth.*()` / `current_setting()` in policies.

## Approach

A single consolidating migration (`20260714230440_wrap_session_gucs_in_rls_policies`) recreates every existing policy (recorded across ~30 prior migrations) whose predicate reads a session GUC. Rather than hand-writing ~280 `DROP/CREATE` statements, it iterates `pg_policies`, deparses each predicate, and substitutes the wrapped forms, preserving name / permissive / cmd / roles / `USING` / `WITH CHECK`. The `.down.sql` is the exact inverse.

### The one non-obvious detail: array operands

Naively wrapping the array operand of `= ANY (...)` **breaks** it — `x = ANY ((select arr))` reparses as the row-returning subquery form and errors (`malformed array literal`). The wrapped array subexpressions therefore keep an explicit `::text[]` cast — `= any ((select regexp_split_to_array(…))::text[])` — which keeps the operand in array form while still creating the InitPlan. Same for the jsonb `?|` operator.

## Validation

- **Applies cleanly** through the backend's normal sqlx path (`_sqlx_migrations.success = t`), no errors/panics.
- **Round-trip is byte-identical** — `up` then `down` reproduces the original `pg_policies` definitions exactly (all 502 policies, verified via `\copy` snapshot diff).
- **InitPlan hoist confirmed** on live RLS via `EXPLAIN (VERBOSE)`: every `current_setting` now appears only in InitPlan `Result → Output` lines (once per statement); **zero** appear in any `Filter:` clause (per-row). Covers all five patterns incl. the `jsonb_each_text(...) EXISTS` group policies used by INSERT/UPDATE/DELETE.
- **Row-equivalence on real data** — for a live session context, `SELECT count(*) FROM script` under the wrapped RLS returns the same set as a hand-written unwrapped union of the `USING` predicates.

The `adding-a-trigger` skill now documents the wrapped form so new trigger tables inherit it (trigger policies are hand-copied per migration; there is no runtime template).

No Rust queries changed → no `.sqlx` update needed. No frontend surface.

---
## Summary by cubic
Wrap session GUC reads in RLS policies with scalar sub-selects so Postgres hoists them into per-statement InitPlans, removing per-row `current_setting` lookups. No behavior change; fixes GIT-919.

- **Refactors**
  - Wrapped all `current_setting('session.user'|'session.groups'|'session.pgroups'|'session.folders_read'|'session.folders_write')` calls in `(select ...)` inside RLS `USING` / `WITH CHECK` predicates to enable per-statement InitPlans.
  - Added migration `20260714230440_wrap_session_gucs_in_rls_policies` to recreate affected policies by deparsing `pg_policies`; the `.down.sql` reverses the rewrite.
  - Kept explicit `::text[]` on wrapped array expressions and fixed docs to place the cast outside the sub-select for `= ANY (...)` so it stays an array operand (also for `?|`); documented canonical forms in `./.agents/skills/adding-a-trigger/SKILL.md`.

<sup>Written for commit c39376552218015f1c60763524c3e60966977406. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/adding-a-trigger/SKILL.md
+++ b/.agents/skills/adding-a-trigger/SKILL.md
@@ -35,12 +35,15 @@ The `up.sql` usually defines:
 
 **RLS: wrap every session GUC read in a scalar sub-select.** Write the session
 reads as `(select current_setting('session.user'))`,
-`= any((select regexp_split_to_array(current_setting('session.groups'), ',')::text[]))`,
-`?| (select regexp_split_to_array(current_setting('session.pgroups'), ',')::text[])`,
+`= any((select regexp_split_to_array(current_setting('session.groups'), ','))::text[])`,
+`?| (select regexp_split_to_array(current_setting('session.pgroups'), ','))::text[]`,
 `? (select concat('u/', current_setting('session.user')))`, etc. — not the bare
 `current_setting(...)`. The GUCs are set with `SET LOCAL`, so the sub-select
 hoists them to a one-time InitPlan instead of re-evaluating per scanned row.
-The array cases keep the `::text[]` cast so `= any (...)` / `?|` stay in
+Put the `::text[]` cast **outside** the sub-select for the array cases: in an
+`= any (...)` context, casting inside — `= any((select ...::text[]))` — makes
+Postgres parse the operand as a row-returning subquery and fails at CREATE with
+`operator does not exist: text = text[]`. The outside cast keeps it in
 array-operand form. See migration `20260714230440_wrap_session_gucs_in_rls_policies`
 for the canonical wrapped forms.
 
--- a/.agents/skills/adding-a-trigger/SKILL.md
+++ b/.agents/skills/adding-a-trigger/SKILL.md
@@ -31,6 +31,18 @@ The `up.sql` usually defines:
   - trigger-specific fields
 - Indexes on foreign keys + any frequently-filtered columns
 - Foreign key to `workspace`
+- The RLS policies (`see_own`, `see_member`, `see_folder_extra_perms_user_*`, `see_extra_perms_user_*`, `see_extra_perms_groups_*`), copied from an existing trigger table
+
+**RLS: wrap every session GUC read in a scalar sub-select.** Write the session
+reads as `(select current_setting('session.user'))`,
+`= any((select regexp_split_to_array(current_setting('session.groups'), ',')::text[]))`,
+`?| (select regexp_split_to_array(current_setting('session.pgroups'), ',')::text[])`,
+`? (select concat('u/', current_setting('session.user')))`, etc. — not the bare
+`current_setting(...)`. The GUCs are set with `SET LOCAL`, so the sub-select
+hoists them to a one-time InitPlan instead of re-evaluating per scanned row.
+The array cases keep the `::text[]` cast so `= any (...)` / `?|` stay in
+array-operand form. See migration `20260714230440_wrap_session_gucs_in_rls_policies`
+for the canonical wrapped forms.
 
 Down migration drops the table and any enum types.
 
--- a/backend/migrations/20260714230440_wrap_session_gucs_in_rls_policies.down.sql
+++ b/backend/migrations/20260714230440_wrap_session_gucs_in_rls_policies.down.sql
@@ -1,0 +1,68 @@
+-- Inverse of the up migration: strip the scalar sub-select wrapping from the
+-- session GUC reads in RLS predicates, restoring the bare per-row forms.
+--
+-- Postgres re-deparses the wrapped sub-selects into a canonical shape
+-- ( SELECT <expr> AS <alias>), and drops the redundant ::text[] cast in the
+-- jsonb `?|` context while keeping it in the `= ANY (...)` context, so the
+-- unwrap tolerates the alias and the optional trailing cast.
+
+CREATE FUNCTION pg_temp.unwrap_session_gucs(expr text) RETURNS text AS $fn$
+DECLARE e text := expr;
+BEGIN
+  IF e IS NULL THEN
+    RETURN NULL;
+  END IF;
+  -- ( SELECT regexp_split_to_array(current_setting('session.<g>'::text), ','::text) AS regexp_split_to_array)[::text[]]
+  e := regexp_replace(e,
+    '\( SELECT (regexp_split_to_array\(current_setting\(''session\.(?:pgroups|groups|folders_read|folders_write)''::text\), '',''::text\)) AS regexp_split_to_array\)(?:::text\[\])?',
+    '\1', 'g');
+  -- ( SELECT concat('u/', current_setting('session.user'::text)) AS concat)
+  e := regexp_replace(e,
+    '\( SELECT (concat\(''u/'', current_setting\(''session\.user''::text\)\)) AS concat\)',
+    '\1', 'g');
+  -- ( SELECT current_setting('session.user'::text) AS current_setting)
+  e := regexp_replace(e,
+    '\( SELECT (current_setting\(''session\.user''::text\)) AS current_setting\)',
+    '\1', 'g');
+  RETURN e;
+END;
+$fn$ LANGUAGE plpgsql IMMUTABLE;
+
+DO $do$
+DECLARE
+  r record;
+  new_qual text;
+  new_check text;
+  stmt text;
+BEGIN
+  FOR r IN
+    SELECT schemaname, tablename, policyname, permissive, cmd, roles, qual, with_check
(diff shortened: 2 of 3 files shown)


rubenfiszel:
## e2e benchmark: before vs after

Synthetic table carrying the exact `script` SELECT policy set (the 5-way OR: `see_own`, `see_member`, `see_folder`, `see_extra_perms_user`, `see_extra_perms_groups`), **1,000,000 rows** (87 MB, no secondary indexes → seq scan, the RLS list read path). Rows are folder scripts the benchmark user can't see, so **every OR branch evaluates for every row** — the worst case, representative of a large workspace where a user sees only a subset. Run as `windmill_user` (RLS enforced), session GUCs set via `set_config`. `EXPLAIN (ANALYZE, TIMING OFF)`, warm cache, `SELECT count(*)`.

| Plan | Unwrapped (before) | Wrapped (after) | Speedup |
|---|---|---|---|
| Parallel seq scan (3 workers) | **~478 ms** (min 474.6) | **~55 ms** (min 53.9) | **~8.7×** |
| Serial seq scan (`max_parallel_workers_per_gather=0`) | **~1347 ms** (min 1345) | **~146 ms** (min 142.6) | **~9.4×** |

Before, the plan keeps every `current_setting(...)` / `regexp_split_to_array(...)` / `concat(...)` / `?|` inline in the `Filter:` and re-runs them per scanned row. After, all five collapse to `InitPlan 1..5` (`(InitPlan N).col1` in the filter), evaluated once per statement:

```
-- before (per row)
Filter: (... = current_setting('session.user'::text)) OR (... = ANY (regexp_split_to_array(current_setting('session.groups'::text), ','::text))) OR ...
-- after (once per statement)
Filter: (... = (InitPlan 1).col1) OR (... = ANY ((InitPlan 3).col1)) OR (extra_perms ?| (InitPlan 5).col1) OR ...
InitPlan 1..5 -> Result -> Output: current_setting(...) / regexp_split_to_array(...) / concat(...)
```

The per-row cost is dominated by the three `regexp_split_to_array` calls plus `concat` and the jsonb `?|`, which the unwrapped form rebuilds on every row. Both variants returned the same rows (0 here; row-equivalence on real data verified separately). The `= any (...)` array branches only work wrapped when the `::text[]` cast sits **outside** the sub-select — `(select regexp_split_to_array(...))::text[]`; putting it inside reparses `any((select …))` as a row-returning subquery (`ERROR: operator does not exist: text = text[]`), which is exactly why the migration keeps the cast outside.

rubenfiszel:
## Full before/after verification of all affected policies

**273 policies across 28 tables** are rewritten, reducing to **36 distinct predicate shapes**. Verified along three independent axes:

### 1. Coverage & integrity (all 273)
- Affected set == exactly the policies whose predicate reads a `session.*` GUC — **no policy dropped or added** (273 → 273, `(table, policy)` sets identical).
- **`cmd` / `permissive` / `roles` unchanged for every policy** — only the predicate text differs (all 273 are `PERMISSIVE`, `windmill_user`).

### 2. Reversibility (all 273 / all 316 in DB)
Unwrapping the live wrapped policies (running the `down` migration in a rolled-back transaction) reproduces the **byte-identical** pre-migration predicate text for **all 316 policies** in the schema. This proves the `up` migration's *only* effect is the session-subexpression wrapping — no predicate logic, operator, or column reference was altered anywhere. Reviewing the 36 distinct shapes confirms it visually: e.g. `regexp_split_to_array(current_setting('session.pgroups'::text), ','::text)` → `( SELECT regexp_split_to_array(current_setting('session.pgroups'::text), ','::text) AS regexp_split_to_array)::text[]`, and nothing else on the line moves. (Shapes with `USING true`, e.g. `usr_to_group`, correctly leave `USING` untouched and wrap only `WITH CHECK`.)

### 3. Semantic identity (value-preserving)
Every policy is composed from just 6 wrapped subexpressions in 7 operator contexts. Each context was proven identical (`orig IS NOT DISTINCT FROM wrapped`) over a boundary-covering input matrix:

| Context | Result |
|---|---|
| `x = current_setting(user)` | ✓ identical |
| `jsonb ? concat('u/',user)` | ✓ identical |
| `(jsonb ->> concat('u/',user))::bool` | ✓ identical |
| `concat('u/',user) = ANY(owners)` | ✓ identical |
| `x = ANY(regexp_split_to_array(...)::text[])` (populated / **empty** / single-element GUC) | ✓ identical |
| `jsonb ?\| regexp_split_to_array(...)::text[]` (populated / **empty** / single-element GUC) | ✓ identical |
| composed `jsonb_each_text(...) EXISTS` group policy + folder `owners`-unnest policy | ✓ identical row-for-row |

The empty-GUC case (`regexp_split_to_array('',',')` → `{''}`) is covered explicitly, and the `= ANY` / `?|` cases are exactly where a wrong cast placement would silently change results — both confirmed value-identical. Combined with (2)'s referential-transparency guarantee (equal subterms → equal predicate under `SET LOCAL` constancy), this establishes that all 273 rewritten policies select the identical rows as before.

rubenfiszel:
## PG14 compatibility — verified on a real Postgres 14.23

Windmill supports PG14+, and the migration's `up` (literal `replace()` on deparsed predicates) and `down` (`regexp_replace` on the canonical deparse) both depend on how Postgres renders `pg_get_expr`. I reproduced all **36 distinct policy shapes** on a clean **PostgreSQL 14.23** container (oldest supported major) and ran the actual `up.sql` / `down.sql` against them.

| Check | Result on PG14.23 |
|---|---|
| Deparse of the **original** policies (what `up`'s literal match keys on) | **byte-identical to PG18** for all 36 |
| `up` migration runs | clean — wraps all 36, **0 skipped** |
| **Wrapped output** | **byte-identical to PG18** (`= any(...)` keeps `::text[]`, `?\|` drops it — same on both versions) |
| InitPlan hoist | `EXPLAIN (VERBOSE)` as `windmill_user`: **0 `current_setting` in any `Filter:`**, InitPlans present → optimization actually fires |
| `down` round-trip | **byte-identical** to the originals on PG14 |

### Why it's compatible across 14–18
- Every construct used predates PG14 by years: the `pg_policies` view + its `permissive`/`cmd`/`roles`/`qual`/`with_check` columns (PG10), `DO`/plpgsql, `regexp_replace` ARE (non-capturing groups, optional quantifier), scalar sub-selects in RLS, `= any(array)`, `format()`, `string_agg`, `quote_ident`. Nothing needs PG15+.
- The only version-sensitive aspect — the assumed `pg_get_expr` deparse shape (`current_setting('…'::text)`, `regexp_split_to_array(…, ','::text)`, `concat`, `( SELECT … AS alias)`, the cast drop for `?\|`) — is rendered **identically** by PG14 and PG18, so `up`/`down` fire the same.
- **Fails safe regardless:** if any deployment's Postgres ever deparsed a form differently and a policy didn't match, that policy is simply left **unwrapped** — correct access, just no InitPlan gain. It can never produce a broken or wrong-permission policy.

InitPlan hoisting of uncorrelated sub-selects is long-standing planner behavior present across the whole 14–18 range, so both correctness and the optimization hold on every supported Postgres version.

PR #10112, opened 2026-07-14
feat(nativets): expose the standard web-platform globals deno_web provides

rubenfiszel:
## Summary
The nativets in-process runtime (`backend/windmill-runtime-nativets`, deno_core based) imports several `deno_web` / `deno_url` modules but never assigns their constructors to `globalThis`. Scripts that assume standard browser/Deno/bun globals therefore broke on nativets: `new TextEncoder()` threw, `res.body instanceof ReadableStream` failed, `EventTarget` was missing while `AbortSignal` (which extends it) was present, etc. This wires the missing globals so nativets matches the bun runner's global surface.

The target surface was verified empirically by probing `typeof <global>` in bun 1.3.11: every global wired here is present in bun, and the two web globals bun lacks (`EventSource`, `ImageData`) are deliberately left out.

## Changes
- **`runtime.js`** — wire onto `globalThis`:
  - `DOMException` (deno_web references it as a global — `AbortController.abort()` with no reason constructs one, so the already-wired `AbortController`/`AbortSignal` threw `DOMException is not defined` on abort before this)
  - `TextEncoder` / `TextDecoder` / `TextEncoderStream` / `TextDecoderStream`
  - `File`
  - `Event` / `EventTarget` / `CustomEvent` / `MessageEvent` / `CloseEvent` / `ErrorEvent` / `reportError`
  - `ReadableStream` / `WritableStream` / `TransformStream` / `ByteLengthQueuingStrategy` / `CountQueuingStrategy`, plus the reader/controller globals bun exposes (`ReadableStreamDefaultReader`, `ReadableStreamBYOBReader`, `ReadableStreamDefaultController`, `ReadableByteStreamController`, `ReadableStreamBYOBRequest`, `WritableStreamDefaultWriter`, `WritableStreamDefaultController`, `TransformStreamDefaultController`) for `instanceof` checks
  - `URLPattern`
  - `CompressionStream` / `DecompressionStream`
  - `MessageChannel` / `MessagePort`
  - `performance` (with `timeOrigin` seeded per isolate, see below) + its constructor globals `Performance` / `PerformanceEntry` / `PerformanceMark` / `PerformanceMeasure`; spec `structuredClone`
- Switched the three previously side-effect-only imports (`13_message_port`, `14_compression`, `15_performance`) to namespace imports so their constructors are reachable (the module bodies still run, so side effects are preserved).
- `performance.timeOrigin` is seeded per isolate: `runtime.js` runs at snapshot-build time, so `setTimeOrigin()` must be called after the snapshot is restored. A `globalThis.__wmInitPerIsolate` hook is invoked from `create_nativets_runtime` (mirroring the existing `__bootstrapOtel` pattern), so `performance.timeOrigin + performance.now()` tracks wall-clock time instead of being `NaN`.
- `structuredClone` is sourced from `13_message_port.js` (the spec function that validates args and honors the options bag), not the single-arg internal helper in `02_structured_clone.js`.
- **globalThis is made a functional `EventTarget`** (per isolate, in `__wmInitPerIsolate`, as Deno's bootstrap does): `setPrototypeOf` to `DedicatedWorkerGlobalScope.prototype` + `setEventTargetData` + webidl brand + `saveGlobalThisReference(globalThis)`. deno_web routes an uncaught `EventTarget`-listener error (and `reportError`) through `reportException`, which dispatches on that saved global reference; `reportError` additionally requires its receiver to equal it. Without this, `dispatchEvent` threw a *masking* error hiding the original, and `globalThis.reportError()` threw `Illegal invocation`. Now both `reportError(e)` and `globalThis.reportError(e)` surface the original error as an async unhandled exception (matching bun, which exits non-zero), and `globalThis.addEventListener`/`dispatchEvent` work.
- **`smoke_tests.rs`** — opt-in smoke tests: multi-byte UTF-8 encode/decode round-trip; an all-globals-defined guard against wrong export names; `structuredClone` deep-clone + `performance` (`now()` and per-isolate `timeOrigin`); a `ReadableStream` + encoding-stream pipe; a construction test that `new`s the ops-backed constructors (`URLPattern`, `MessageChannel`, gzip `Compression`/`DecompressionStream`); and a broad **functional edge-case sweep** (`smoke_web_globals_edge_cases`, ~35 checks) that exercises every wired global for real — `DOMException`/`AbortController.abort()` default reason, `AbortSignal.timeout()`, `EventTarget` dispatch, stream tee/reader/writer/transform, all three compression formats, `structuredClone` of Map/Set/Date/TypedArray/circular + reject-function, `performance.mark/measure`, `MessagePort` message delivery — to catch defined-but-broken cases (this is how the `reportError` and `DOMException` gaps were found).
- Merged `main` (Web Crypto via deno_crypto, #10109). The crypto globals (`crypto`/`Crypto`/`CryptoKey`/`SubtleCrypto`) and the new web globals coexist; the edge-case sweep also re-checks `crypto.getRandomValues`/`subtle.digest` still work.

## Out of scope / follow-ups
- `WebSocket`, `Worker` need an additional extension registered, not just a global assignment. (`crypto`/`crypto.subtle` already landed on `main` via #10109 and is merged in here.)
- `structuredClone` is the spec function, but the runtime doesn't wire the transfer op, so `options.transfer` clones without detaching the source (pre-existing runtime limitation).

## Test plan
- [x] **After** (with the wiring, post-merge with `main`): `cargo test -p windmill-runtime-nativets --lib -- --ignored smoke_web_globals smoke_text_encoder smoke_structured_clone smoke_readable_stream smoke_web_blob smoke_web_crypto` — all 8 pass (incl. the ~35-check edge-case sweep and the merged `smoke_web_crypto`); snapshot rebuilds cleanly.
- [x] **Before** (pre-change `runtime.js` restored into the working tree, new tests kept): 5 of 6 fail with real `ReferenceError`s (`TextEncoder is not defined`, `ReadableStream is not defined`, `structuredClone is not defined`, `URLPattern is not defined`, and `smoke_web_globals_are_wired` reports the unwired names). The control `smoke_web_blob_btoa_atob` passes in both states (Blob/btoa/atob already existed), confirming no regression to pre-existing globals and that the new tests don't pass vacuously.
- [x] Note: these smoke tests exercise the runtime's V8 isolate end-to-end (transpile → spawn isolate → execute → assert), the real execution path — not a full worker/queue-dispatched job (that path is covered by `backend/tests/worker.rs`, which doesn't reach this JS API surface).
- [x] `performance.timeOrigin`: `smoke_structured_clone_and_performance` asserts it's finite and `timeOrigin + now()` is within 5s of `Date.now()` (pre-fix this asserted `NaN` and failed).
- [x] Export names verified against vendored `deno_web` 0.240.0 / `deno_url` 0.209.0 source.
- [x] Global surface verified against bun 1.3.11.

---
## Summary by cubic
Expose standard web‑platform globals in the NativeTS runtime so scripts expecting browser/Deno/`bun` APIs run correctly. Make `globalThis` a real `EventTarget`, seed `performance.timeOrigin` per isolate, and expand smoke tests for presence, construction, and edge cases.

- **New Features**
  - Wire missing globals: `DOMException`; `TextEncoder`/`TextDecoder` (+ streams); `File`; `Event`/`EventTarget`/`CustomEvent`/`MessageEvent`/`CloseEvent`/`ErrorEvent` (+ `reportError`); Streams + queuing strategies and reader/controller globals; `URLPattern`; `CompressionStream`/`DecompressionStream`; `MessageChannel`/`MessagePort`; `performance`; spec `structuredClone` (exclude `EventSource` and `ImageData`).
  - Add performance constructor globals: `Performance`, `PerformanceEntry`, `PerformanceMark`, `PerformanceMeasure` (bun parity).
  - Switch `message_port`, `compression`, and `performance` modules to namespace imports so constructors are reachable.
  - Add smoke tests for globals presence and construction, multi‑byte encode/decode, streams piping, gzip round‑trip, spec `structuredClone`, `performance.now()`/`timeOrigin`, `mark`/`measure` + `instanceof` on performance entries, throwing listener/reportError behavior, and a broad edge‑case sweep.

- **Bug Fixes**
  - Seed `performance.timeOrigin` per isolate via `__wmInitPerIsolate` so timings are finite and correct.
  - Make `globalThis` a functional `EventTarget` so `dispatchEvent` works and both `reportError(e)` and `globalThis.reportError(e)` surface the original error (no “Illegal invocation”).
  - Fix abort errors by wiring `DOMException`; resolve runtime issues like `new TextEncoder()` throwing and `res.body instanceof ReadableStream` failing.
  - Reword inline comment to state the bun‑parity constraint and the deliberate `EventSource`/`ImageData` exclusions.

<sup>Written for commit 907cebcf662cf4f14bc1af2a19a68aba7ef9c10c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the Codex review in the latest commit:

1. **[P1] `reportError` unusable** — confirmed: in `deno_web` 0.240.0 `reportError` → `reportException` dispatches through `globalThis_.dispatchEvent`, which is only set by `saveGlobalThisReference()`; this runtime never calls it nor installs `dispatchEvent` on the global, so a call throws (`Illegal invocation` / undefined `globalThis_`). Rather than pull in full global-EventTarget setup, I **dropped `reportError`** from the wiring (its event-class companions `Event`/`CustomEvent`/`MessageEvent`/`CloseEvent`/`ErrorEvent` construct standalone and stay). Documented inline why it's excluded, and removed it from the presence test.

2. **[P1] `performance.timeOrigin` uninitialized** — confirmed: `setTimeOrigin()` is never called, and it can't be called from `runtime.js` since that executes at snapshot-build time (it would freeze the build-time origin). Fixed by adding a per-isolate init hook `globalThis.__wmInitPerIsolate` (calls `performance.setTimeOrigin()`) and invoking it from `create_nativets_runtime` right after the snapshot is restored, mirroring the existing `__bootstrapOtel` pattern. `smoke_structured_clone_and_performance` now asserts `timeOrigin` is finite and `timeOrigin + performance.now()` is within 5s of `Date.now()`.

3. **[P2] drafting-history comment** — reworded the namespace-import comment to describe only the current constraint (namespace imports so the constructors are reachable; module bodies still run for side effects), no "previously imported".

All 6 opt-in smoke tests pass; verified before/after (pre-fix, the timeOrigin assertion fails with `NaN`).

rubenfiszel:
Addressed the Codex review in the latest commit:

**1. [P1] Throwing EventTarget listeners** — confirmed: an uncaught listener error is caught by deno_web's `invokeEventListeners` and passed to `reportException`, which dispatches an error event on a *saved global reference*. With none set, `dispatchEvent` threw a masking error (deref of the unset reference) that hid the original. I probed bun 1.3.11 for the reference behavior: a throwing listener there does **not** throw out of `dispatchEvent` (it returns `true`) — the error is reported **asynchronously** (bun exits non-zero), and notably **`globalThis instanceof EventTarget` is `false`** in bun. So rather than mutate `globalThis`'s prototype (which would diverge from bun and is riskier), I wired a dedicated `EventTarget` as the saved dispatch target: `event.saveGlobalThisReference(new event.EventTarget())`. Now a throwing listener surfaces the **original** error as an async unhandled exception (bun-matching), and `reportError` is re-added and functional. New regression test `smoke_eventtarget_throwing_listener_reports_original` asserts the surfaced error is the original (`wm_listener_marker`), not a masking `Illegal invocation`/`undefined`.

**2. [P2] Missing stream globals** — wired the reader/controller constructors bun also exposes: `ReadableStreamDefaultReader`, `ReadableStreamBYOBReader`, `ReadableStreamDefaultController`, `ReadableByteStreamController`, `ReadableStreamBYOBRequest`, `WritableStreamDefaultWriter`, `WritableStreamDefaultController`, `TransformStreamDefaultController`. The edge-case sweep now obtains real reader/writer/controller instances and verifies `instanceof` against these globals.

On the note about `mold` being unavailable locally for the reviewer: that's the CI/dev linker config, unrelated to the change; all smoke tests pass here (18 non-network, incl. the new regression test).

PR #9860, opened 2026-07-01
fix(jseval): increase QuickJS memory limit from 32MB to 256MB

salarkhannn:
Flow expression evaluation uses QuickJS to evaluate step input transforms. The 32MB memory limit was too restrictive for payloads that construct large arrays — for example `Array.from({length: 1e6}, ...).flat()` exceeds 32MB and causes QuickJS to throw `Exception generated by quickjs`, surfacing as an opaque eval failure.

Increasing the limit to 256MB. QuickJS evaluations are always authenticated and run inside the worker per-job isolation, so this is safe.

Closes #8073.

---
## Summary by cubic
Increase the QuickJS memory limit for flow expression evaluation from 32MB to 256MB to prevent opaque eval failures on large payloads (e.g., million-element arrays). Evaluations are authenticated and run in per‑job isolated workers, so this increase is safe.

<sup>Written for commit 5c5b6b9f187afd6bf7c64694edc0adb74ab80256. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/9860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10116, opened 2026-07-14
fix(jseval): raise QuickJS eval memory cap to 128MB with clear OOM error

### LINKED ISSUE #8073: bug: expression evaluation fails with large payloads in step input transforms (QuickJS)
### Describe the bug

When running a Flow with large input payloads , any step that uses expression evaluation in step inputs fail during evaluation.

This error does not occur when the input data is small — the same flow and expressions run successfully with smaller payloads. The failure only appears once the input size becomes large.

### To reproduce

- Create a new Flow.
- Add Step 1 (TypeScript) and execute the following JavaScript code inside the step code:

```
Array.from({ length: 1000000 }, (_, i) => [i, i + 1, i + 2]).flat()
```
Run the flow and confirm Step 1 completes successfully.
- Add Step 2 (TypeScript).
In the step input transform expression ( evaluated directly in the step input), use the same expression:

Run the flow again.
- Observe that Step 2 fails with: QuickJS evaluation error: Exception generated by quickjs

### Expected behavior

Both Step 1 (script execution) and Step 2 (expression evaluation in step inputs) should succeed and return the same flattened array output.

### Screenshots

1. Image
2. Image
3. Image

### Browser information

_No response_

### Application version

CE v1.641.0

### Additional Context

_No response_

rubenfiszel:
## Summary

Flow step-input transforms are evaluated with QuickJS (`backend/windmill-jseval`), which capped the eval heap at 32MB. Legitimate payloads that build large arrays (the issue's `Array.from({ length: 1e6 }, (_, i) => [i, i+1, i+2]).flat()`) exceed 32MB, and QuickJS surfaced this as an opaque `Exception generated by quickjs` with no hint that memory was the cause. Closes #8073.

This raises the default cap, makes it env-configurable, and — most importantly — turns the opaque out-of-memory failure into a clear, actionable message.

Supersedes #9860 (which only bumped the number 32→256MB): the cap is now a named, documented, env-tunable constant, and users who hit it get told exactly what to do.

## Changes
- Replace the hardcoded `QUICKJS_MEMORY_LIMIT = 32MB` with a documented `DEFAULT_QUICKJS_MEMORY_LIMIT_BYTES = 128MB` and a `lazy_static` resolved once from `QUICKJS_MEMORY_LIMIT_MB` (positive integer MB; `saturating_mul` against hostile values).
- Add `map_quickjs_error()`: classify the caught error **by the caught value's type** (reliable) and rewrite OOM into `The expression evaluation exceeded the memory limit of N MB. Reduce the amount of data handled in the transform, or raise the cap via the QUICKJS_MEMORY_LIMIT_MB environment variable.`
  - `Exception` whose message contains `out of memory` → OOM.
  - `Value` that is `null`/`undefined` → OOM (QuickJS couldn't even allocate the `Error`, so it throws a bare null — this is the opaque `Exception generated by quickjs` that #8073 reported).
  - Anything else (e.g. a user `throw "boom"`, ordinary JS errors) → left as a normal error, never misattributed to memory.
  - A post-hoc heap check was tried and **rejected**: QuickJS ref-count-frees the offending arrays as the JS stack unwinds, so `malloc_size` drops below any threshold by the time we could read it (measured — it leaked the opaque error at 128/160MB). The value-type check has no such race.
- Wire both eval paths to the resolved limit + classifier: `eval_quickjs_inner` (flow transforms) gains a `memory_limit` param; `eval_simple_js` is refactored to extract `eval_simple_js_inner`.
- All new items are `#[cfg(feature = "quickjs")]`-gated; the non-quickjs stubs are unchanged.

## Why 128MB (and why the default is deliberately modest)
Within a worker process, transforms are evaluated **one at a time** (`transform_input` awaits each sequentially; each eval drops its `AsyncRuntime` before the next), so the cap is per-eval, not per-flow. The default deployment is **one worker per process** — `DEFAULT_NUM_WORKERS = 1`, and `src/main.rs` refuses `NUM_WORKERS > 1` in-process unless native mode or `I_ACK_NUM_WORKERS_IS_UNSAFE`. In **native / multi-worker-in-process** mode, up to `NUM_WORKERS` evals run concurrently in **one shared heap** (and this is the worker binary's own RSS, since QuickJS is in-process, unlike Python/TS jobs that run in child subprocesses), so the process peak is `NUM_WORKERS × cap`. Keeping the default at 128MB bounds that worst case (8 native workers → ~1GB rather than ~2GB at 256MB) while still being 4× the old limit. Operators with genuinely large transforms raise `QUICKJS_MEMORY_LIMIT_MB`.

> [!NOTE]
> The exact #8073 payload measures at needing ~256MB, so under the 128MB default that specific repro still exceeds the cap — but now fails with the clear "exceeded the memory limit of 128 MB … raise the cap via QUICKJS_MEMORY_LIMIT_MB" message instead of the opaque one. Setting `QUICKJS_MEMORY_LIMIT_MB=256` makes it succeed. The default is a guardrail; oversized transforms opt into more.

## Test plan
- [x] `cargo test -p windmill-jseval --features quickjs` — 69 passing, incl.:
  - `test_eval_oom_error_reports_memory_limit` — 4MB cap, `out of memory` Error path → clear message.
  - `test_eval_opaque_oom_reports_memory_limit` — the #8073 flat payload at 64MB, the opaque null-throw path → clear message (guards the exact bug symptom).
  - `test_eval_user_throw_not_reported_as_oom` — `throw 'boom'` is **not** misreported as OOM.
  - `test_eval_issue_payload_succeeds_with_raised_cap` — the #8073 payload succeeds at 256MB via `eval_quickjs_inner` (the flow transform path).
  - `test_eval_moderately_large_array_under_default` — a ~48MB array (would fail at 32MB) succeeds under the 128MB default via `eval_timeout_quickjs`.
- [x] `cargo check -p windmill-jseval` (no `quickjs`) — stub path still compiles; no warnings either config.
- No SQL, API, or frontend surface touched.


rubenfiszel:
Addressed the Codex review (pushed 977754b):

- **[P1] Concurrent batch-rerun evals** — valid. `eval_simple_js` runs in the **API process**, which serves concurrent requests with no per-process serialization (unlike workers), so its peak is `concurrent_requests × cap`, unbounded by `NUM_WORKERS`. The PR had let it inherit the raised 128MB cap. Fixed by giving it a **separate `EVAL_SIMPLE_JS_MEMORY_LIMIT_BYTES = 32MB`** (its pre-PR value, not tied to `QUICKJS_MEMORY_LIMIT_MB`) — batch-rerun filters and `retry_if` only remap args / return a boolean, so 32MB is ample. This restores the prior API-process bound while keeping the flow-transform improvement.
- **[P2] Misclassification** — tightened the Exception branch to an **exact** `message() == "out of memory"` (was `.contains`), so a user `throw new Error('out of memory later')` is no longer swept into the OOM bucket. Added `test_eval_user_throw_not_reported_as_oom` cases for `throw 'boom'`, `throw new Error('boom')`, and the memory-text Error. The `null`/`undefined` throw ambiguity is genuinely irreducible (QuickJS emits a bare `null` on OOM), so it stays absorbed into the OOM bucket by design — now documented in the `map_quickjs_error` doc comment and pinned by `test_eval_bare_null_throw_treated_as_oom` so the behavior is a conscious, tested contract. (Note: `Promise.reject()` does not apply to `eval_simple_js`, which doesn't await promises.)
- **[P2] Drafting-history comments** — reworded the constant doc and test comments to state the current sizing constraint without change history, per AGENTS.md.

70 tests pass; both feature configs compile warning-free.

rubenfiszel:
Addressed the two follow-up P2s (pushed 2faaad2):

- **Ineffective remediation on the fixed-cap path** — `map_quickjs_error` now takes an `env_override: Option<&str>`. The flow path (`eval_quickjs_inner`, env-tunable) passes `Some("QUICKJS_MEMORY_LIMIT_MB")`; the fixed-cap `eval_simple_js` path passes `None`, so its message reads *"…exceeded the memory limit of 32 MB. Reduce the amount of data handled in the expression."* with no misleading env-var advice. New tests `test_eval_flow_oom_reports_env_override` (asserts the hint) and the updated `test_eval_oom_error_reports_memory_limit` (asserts the fixed-cap path omits it).

- **Exact `throw new Error("out of memory")` still misclassified** — you were right that message-equality alone wasn't enough. I probed the actual exceptions: genuine OOM is an **`InternalError`** (`name="InternalError"`, `message="out of memory"`), whereas a user `new Error("out of memory")` has `name="Error"`. The classifier now requires the `InternalError` **kind** *and* the exact message, so the user Error is left as a normal error. `test_eval_user_throw_not_reported_as_oom` now includes the exact `new Error('out of memory')` case. The only residual ambiguity is a literal `throw null` / `throw undefined`, which is genuinely indistinguishable from QuickJS's own OOM null-throw and stays absorbed into the OOM bucket by design (documented + pinned by `test_eval_bare_null_throw_treated_as_oom`).

71 tests pass; both feature configs compile warning-free.

PR #10121, opened 2026-07-15
fix(apps): invalidate cached app policy on change or deletion (GHSA-r5v4-cxh9-7qhq)

rubenfiszel:
## Summary

Fixes security advisory **GHSA-r5v4-cxh9-7qhq**: a public app stayed anonymously executable after its public access was revoked.

The versioned `execute_component` path cached the app policy keyed by the caller-supplied `version`. A policy-only update rewrites `app.policy` in place **without bumping the version**, so the cache kept serving the old policy — `execution_mode` (anonymous/public), `triggerables`, and `on_behalf_of` — letting anonymous callers keep executing after a revocation, run removed runnables, run as a withdrawn identity, or replay one app's policy under another.

## Fix

Re-key the cache by the app's real identity `(workspace_id, path)` (the policy is a single per-app value, not per-version), and gate reuse on **three independent freshness signals** — each covers a case the others can't, and any disagreement only ever triggers a safe refetch:

| Signal | Catches | Window |
|---|---|---|
| **caller `version` matches the cached one** | a redeploy (bumps the version) — free, it's already in the request | instant |
| **`notify_app_policy_change` hasn't evicted the key** | a policy-only change / deletion (no version bump), on every replica | ≤ notify poll interval |
| **entry within TTL (5 min)** | a missed event / restart | bounded |

- `backend/windmill-api/src/apps.rs` — `APP_POLICY_CACHE` keyed by `(workspace_id, path)`, value `(cached_version, policy, cached_at)`; reuse only when all three signals agree, else refetch live. `invalidate_app_policy_cache(workspace, path)` drops one entry. `payload.version` is now purely a cache-freshness hint (it never selects which version runs — the policy/runnables are always the app's current ones).
- **Migration** `notify_app_policy_change` — `SECURITY DEFINER` triggers emitting a `notify_event` on `AFTER UPDATE OF policy` (when it actually changes) and on `AFTER DELETE`.
- `backend/src/main.rs` — the poller handles `notify_app_policy_change` and invalidates the key on every server.
- `backend/windmill-api/src/lib.rs` — re-export the invalidation fn.

The `(workspace, path)` key also closes cross-app / same-path-recreation replay (the key is the app's identity, not the caller-supplied version).

### Staleness tradeoff (please review)

A **policy-only revoke** (which doesn't bump the version) is caught only by the notify event + TTL, so there is a bounded staleness window — typically ≤ the notify poll interval, worst case the 5-minute TTL — during which an anonymous caller can still execute a just-revoked app. This is an accepted product decision to keep the cache. A fully synchronous, zero-window alternative is a one-line change (always fetch the policy live) if that window is not acceptable.

## Test plan

- [x] `cargo check -p windmill-api` (offline, `-D warnings`) and `cargo check --bin windmill` pass.
- [x] SQL trigger check: fires on a real policy change and on delete; not on a value/summary-only update.
- [x] `test_policy_change_invalidates_cache` — prime, revoke to `publisher`, dispatch the emitted event as the poller would, assert the anonymous execute is rejected; same for delete. Verified that **without** dispatching, the stale cache still executes (returns a job id), confirming cache + invalidation are both exercised.
- [x] `test_version_change_refetches_policy` — after a revoke, executing with a bumped version refetches and is rejected **without** any notify dispatch (isolates the version signal).
- [x] `test_versioned_execute_not_replayable_across_apps` — a version executed against a different app resolves that app's own policy.
- [x] `.sqlx` unchanged vs `main` (no new query macros; triggers are pure SQL).

## Notes

- No `*_ee.rs` files touched, so no EE companion PR. No frontend changes.


rubenfiszel:
Addressed the Codex review by switching from `notify_event`-based cache invalidation to **synchronous live validation** of `execution_mode` on every versioned execute (`live_app_execution_mode` in `apps.rs`). This resolves all three P0s at once:

- **P0-1 (post-revocation window):** the live `execution_mode` is now read and enforced synchronously before authorization, so there is no polling window — a revoked app is rejected on the very next request.
- **P0-2 (restart / missed events):** authorization no longer depends on consuming a transient event, so the disk-backed cache + startup cursor-skip can't retain a stale authz decision. Deploying the fix also takes effect immediately, with no pre-existing stale entries to flush.
- **P0-3 (app deletion):** the live read targets the app row; a deleted app returns not-found and execution is blocked, without needing a delete-side trigger. The regression test now covers deletion in addition to revocation.

The `notify_event` trigger + migration + poller handler have been removed. The non-authz `triggerables` stay cached (the documented perf trade-off); only the authz-critical `execution_mode` is read live (one indexed `LIMIT 1` query per versioned execute).

On the test-coverage note (the old test manually called `invalidate_app_policy_cache()`, masking the polling delay): that's gone — the new test primes the cache via a real anonymous execute, then asserts rejection after revoke and after delete with no manual cache manipulation, so it exercises the actual live-read gate.

rubenfiszel:
Addressed the cross-app replay P0.

**Root cause:** the version-keyed policy cache used the bare, caller-controlled `version` as its key while the policy is fetched by `path`, so a policy primed for app A under version K could be served for app B when B was called with version K — leaking A's `triggerables` and `on_behalf_of` identity.

**Fix (`apps.rs`):** the cache is now keyed by a hash of `(workspace_id, path, version)` instead of the bare version. Replaying app A's version id against app B computes a different key, misses, and fetches B's own policy, so A's cached entry can never be served under B's path. This also preserves per-deploy freshness (a new version is a new key).

**Regression test:** `test_versioned_cache_not_replayable_across_apps` creates two anonymous apps (A can run `script/noop`, B cannot), primes A's policy under a version id, then replays that version against B's execution endpoint and asserts B does not inherit A's runnable — it fails with "Path forbidden by policy" instead of enqueuing a job. I verified the test fails against the old bare-version key ("app_b must not inherit app_a's cached policy, got 200 OK") and passes with the fix.

All three tests pass (revoke, delete, cross-app), idempotent across runs.

rubenfiszel:
Addressed the persisted-legacy-key P0.

Two changes to close the cross-release vector:

1. **Dedicated cache namespace.** The scoped cache now lives in its own namespace (`app_policy_scoped`) instead of reusing `policy`. The previous scheme keyed that namespace by the caller-supplied version, so an attacker could pre-plant `policy/<X>` on an old release where `X` equals a future scoped key. Reading from a fresh namespace means those pre-planted entries are never imported after upgrade; the old directory is simply orphaned.

2. **Full digest key (no truncation).** The key is now the full hex Sha256 of `(workspace_id, path, version)` (a `String`), not a truncated `u64`, so an attacker can't craft a 64-bit collision to alias one app's key onto another's.

Together these mean a persisted entry can only be served for the exact `(workspace, path, version)` that produced it, on this release's namespace.

On test coverage: as you noted, the "persisted entry from the previous key scheme" scenario is a cross-release/on-disk-layout concern that can't be faithfully simulated in a single-process integration test without asserting on cache-internal file paths (brittle). It's closed structurally by the namespace change — the new code never reads the old namespace — so I've left the automated suite at the three behavioral cases (revoke, delete, in-release cross-app replay) and rely on the namespace switch for the rollout case.

All three tests still pass.

rubenfiszel:
Addressed the same-path-recreation P0.

**Root cause (shared by the whole class):** `payload.version` is caller-controlled and was never bound to the app. The scoped cache key includes the path but not the app identity, so deleting app A at path P and recreating app B at P reused A's cached entry (same `(workspace, path, version)` key), and B's live `anonymous` mode matched.

**Fix (`apps.rs`):** the live check now also validates that the requested version belongs to the current app row — `WHERE path = $1 AND workspace_id = $2 AND $3 = ANY(versions)`. If the version isn't one of the app's versions (a different app, or a different app later created at the same path), the row isn't returned and execution fails closed. This binds version→app at the source, closing cross-app replay and same-path recreation; the app-scoped key + dedicated namespace remain as defense-in-depth (and keep the cached triggerables app-correct). Legitimate calls are unaffected — the app always executes with a version from its own `versions` array, and versions are only ever appended.

**Regression test:** `test_versioned_cache_not_replayable_after_same_path_recreation` primes an anonymous app at path P, deletes it, creates a new anonymous app at P, and replays the old version — asserting it can't execute the deleted app's runnable. Verified it fails without the version check ("recreated app must not inherit the deleted app's cached policy, got 200 OK") and passes with it.

All four tests pass.

rubenfiszel:
Addressed the publisher→anonymous P0 by removing the policy cache on the execute path entirely.

The mismatch-branch fix suggested (fetch the full live policy when modes differ) would still leave the matching branch stale — e.g. an anonymous→anonymous policy-only update that removes a runnable keeps serving the cached triggerables. The common cause of every finding across this review is that the version-keyed cache can't reflect a policy-only update (mode, triggerables, and on-behalf-of identity all change in place without bumping the version).

So the versioned execute path now **fetches the policy live on every request and never caches it** — identical to the non-versioned path:

```
-            let policy = if let Some(id) = payload.version {
-                let cache = cache::anon!({ u64 => Arc<Policy> } in "policy" <= 1000);
-                arc_policy = policy_fut.map_ok(Arc::new).cached(cache, id as u64).await?;
-                &*arc_policy
-            } else {
-                policy = policy_fut.await?;
-                &policy
-            };
+            policy = policy_fut.await?;
+            let policy = &policy;
```

This structurally eliminates the whole class: no stale mode after revoke, no stale triggerables after a runnable removal (this finding), no stale on-behalf-of identity, no cross-app or same-path replay (the policy is always fetched by the request's own path), and nothing persisted to disk to import after a restart/upgrade. The `live_app_execution_mode` helper, scoped cache key, and namespace are all gone — the fix is a net simplification. Cost is one indexed `SELECT policy … LIMIT 1` per versioned execute, which the non-versioned path already paid and which is dwarfed by the job enqueue.

Tests (4, all pass): revoke-then-execute rejected; runnable removed via policy-only update → withdrawn runnable rejected (directly covers this finding); deleted app rejected; a version replayed against a different app resolves that app's own policy. `.sqlx` is back in sync with main (no new queries).

rubenfiszel:
Both nits addressed:

1. **PR body** — already updated to the final implementation before this review ran (the review snapshotted the body at push time). It now describes the live fetch with no cache, no `payload.version` use, no new SQLx query, and the three regression tests.
2. **Comment length** — compressed the rationale from nine lines to four, recording just the durable cache-safety constraint:

```rust
// Fetch the policy live, never cache it by version: a policy-only update
// (revoke, runnable removal, identity change) rewrites it in place without
// bumping the version, so a version-keyed cache would keep serving a stale
// grant to anonymous callers (GHSA-r5v4-cxh9-7qhq).
```

PR #10123, opened 2026-07-15
fix(security): enforce variables:write scope on resource-delete var cascade (GHSA-xmr2-98m6-cjf7)

rubenfiszel:
## Summary

Fixes **GHSA-xmr2-98m6-cjf7**. A token scoped only to `resources:write:<path>` could delete linked secret **variables** it had no `variables:write` scope for: `delete_resource`/`delete_resources_bulk` recursively harvest every `$var:<path>` from the (attacker-controlled) resource JSON and `DELETE` those `variable` rows, with only a `resources:write` check. `delete_variable` itself gates on `variables:write:<path>`, so the cascade bypassed that boundary. #9712 tightened scoped-token boundaries broadly but missed this path.

## Fix

`check_linked_var_delete_scopes` requires `variables:write:<var>` for every linked variable before any delete, failing (and rolling back) otherwise. No-op for unscoped/full tokens, so normal cascade cleanup is unchanged; a token holding both `resources:write` + `variables:write` still cascades.

No co-located-path exemption: a resource and a variable may share a path, and a `resources:write` token can create a resource over an existing standalone variable and self-reference it, so "same path as the deleted resource" is attacker-forgeable and can't stand in for variable scope (per the Codex CI review of an earlier iteration).

Core change is ~23 lines; `update_resource`/`update_resource_value` have no `$var:` delete cascade and are unaffected. No SQL changes, no `*_ee.rs` touched.

## Test

`backend/tests/ws_specific.rs`: a `resources:write:<r>`-only token deleting a resource that references an out-of-scope `$var:` → **403**, variable survives; a full token still cascades. `cargo test --test ws_specific` passes.

(The `cargo_test` CI failure is an unrelated pre-existing flake: `ee::audit_s3_export::audit_export_end_to_end`.)

### DIFF
--- a/backend/tests/ws_specific.rs
+++ b/backend/tests/ws_specific.rs
@@ -24,6 +24,39 @@ fn authed(builder: reqwest::RequestBuilder, token: &str) -> reqwest::RequestBuil
     builder.header("Authorization", format!("Bearer {}", token))
 }
 
+/// Helper: does a variable exist at (workspace, path)?
+async fn variable_exists(db: &Pool<Postgres>, workspace: &str, path: &str) -> anyhow::Result<bool> {
+    let n: Option<i64> =
+        sqlx::query_scalar("SELECT COUNT(*) FROM variable WHERE workspace_id = $1 AND path = $2")
+            .bind(workspace)
+            .bind(path)
+            .fetch_one(db)
+            .await?;
+    Ok(n.unwrap_or(0) > 0)
+}
+
+/// Helper: mint an API token restricted to `scopes`, using the admin SECRET_TOKEN.
+async fn mint_scoped_token(port: u16, scopes: Vec<&str>) -> anyhow::Result<String> {
+    let resp = authed(
+        client().post(format!("http://localhost:{port}/api/users/tokens/create")),
+        "SECRET_TOKEN",
+    )
+    .json(&json!({
+        "label": "scoped",
+        "scopes": scopes,
+        "workspace_id": "test-workspace",
+    }))
+    .send()
+    .await?;
+    assert_eq!(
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Addressed the Codex P0 (co-located exemption is attacker-forgeable). Since a `resources:write` token can create a resource over an existing standalone variable and self-reference it, "same path as the deleted resource" can't stand in for variable scope. Dropped the co-located exemption entirely: `check_linked_var_delete_scopes` now requires `variables:write` for every linked variable in both `delete_resource` and `delete_resources_bulk`.

- Full/unscoped tokens are unaffected (no-op), so normal cascade cleanup is unchanged.
- A token holding both `resources:write` + `variables:write` still cascades.
- Added a regression test (`test_colocated_secret_still_requires_variables_write`) that reproduces the manufactured-co-location attack: pre-create a standalone secret, create a resource over its exact path with `$var:` self-reference, then delete with a `resources:write:u/test-user/*`-only token → 403 + secret survives; and confirms a resources+variables scoped token does cascade.

All 8 `ws_specific` tests pass.

rubenfiszel:
Addressed the Codex P2 (contradictory co-located test doc) and minimized the PR since this is a niche boundary issue: dropped the extra tests that were only distinguishing intermediate exemption variants (now gone) and kept a single focused regression — a `resources:write:<r>`-only token cannot cascade-delete an out-of-scope linked variable (403 + survives), full token still cascades. Core change is ~23 lines.

Note: the `cargo_test` CI failure is an unrelated pre-existing flake (`ee::audit_s3_export::audit_export_end_to_end`), not from this change.

PR #10125, opened 2026-07-15
fix(security): rebind forked app run-as to the fork creator (GHSA-r7x4)

rubenfiszel:
## Summary

Fixes security advisory **GHSA-r7x4-qj3p-jfvr**. Forking a workspace clones its apps, but `clone_apps` copied each app's `policy` verbatim. A public app's policy runs as an elevated `on_behalf_of` identity; copied unchanged into a low-privileged fork, the fork owner retained a publicly-reachable entrypoint executing as the parent's elevated identity, surviving revocation of the parent app's public access. Any non-admin member can create a (non-dev) fork, so this let a low-privileged user inherit privileged execution across a workspace/trust boundary.

## Fix

A fork clone is conceptually the **fork creator deploying the parent's apps** into their own workspace. A non-privileged redeploy already rewrites `on_behalf_of` to the deployer (`update_app_internal`, `backend/windmill-api/src/apps.rs:2566`) — the grant to run as someone else is gated by `can_preserve_on_behalf_of` (admin / `wm-deployers`). The clone simply skipped that step.

So this rebinds each cloned app's `on_behalf_of` / `on_behalf_of_email` to the fork creator. The app then runs with the creator's own permissions rather than the inherited elevated identity — closing the cross-workspace escalation while preserving the app's behavior (an app running as its deployer is exactly what deploying it yourself produces).

## Changes

- **`backend/windmill-api-workspaces/src/workspaces.rs`**: `clone_apps` now rebinds `on_behalf_of`/`on_behalf_of_email` to the fork creator; the creator's identity (`authed.username` / `authed.email`) is threaded through `clone_workspace_data`.

## Why not force `viewer` mode / strip the identity

An earlier iteration forced `execution_mode = viewer`. That works but changes the app's execution semantics for every forked app (viewer runs as the caller, not an author). Rebinding `on_behalf_of` to the creator is the minimal change that matches the existing deploy model: `execution_mode` is left untouched, and there is no run-as an identity the creator wasn't entitled to. Simply stripping `on_behalf_of` isn't viable either — `anonymous`/`publisher` modes require it and would error at execution (`get_on_behalf_of`).

## Scope note

Confined to the OSS-tracked `workspaces.rs` clone path; no `*_ee.rs` files are touched, so there is no EE companion PR. GHSA-hfh4 (the `on_behalf_of_email` superadmin-sentinel store-time validation) is handled separately and does not overlap this diff.

## Test plan

- [x] `cargo check --features enterprise,private,parquet -p windmill-api-workspaces` passes
- [ ] (optional) UI sanity: fork a workspace with a public on-behalf-of app; confirm the cloned app runs as the fork creator (its own permissions), not the parent's identity

### DIFF
--- a/backend/windmill-api-workspaces/src/workspaces.rs
+++ b/backend/windmill-api-workspaces/src/workspaces.rs
@@ -4981,6 +4981,15 @@ async fn clone_apps(
         if let Some(&current_version) = app.versions.last() {
             latest_version_ids.insert(current_version);
         }
+        // Reset run-as to auth-required viewer so a cloned public app can't
+        // execute anonymously as the parent's elevated on_behalf_of identity.
+        let mut policy = app.policy;
+        if let Some(obj) = policy.as_object_mut() {
+            obj.insert(
+                "execution_mode".to_string(),
+                serde_json::Value::String("viewer".to_string()),
+            );
+        }
         let new_app_id = sqlx::query_scalar!(
             "INSERT INTO app (workspace_id, path, summary, policy, versions, extra_perms, custom_path)
              VALUES ($1, $2, $3, $4, $5, $6, $7)
@@ -4988,7 +4997,7 @@ async fn clone_apps(
             target_workspace_id,
             app.path,
             app.summary,
-            app.policy,
+            policy,
             &Vec::<i64>::new(), // Start with empty versions array
             app.extra_perms,
             app.custom_path,


rubenfiszel:
Closing — after review this is not a privilege escalation, so no code change is warranted.

**Why it's not exploitable**

- Only a **member** of the parent workspace can create a (non-dev) fork.
- The creator's `is_admin` in the fork is copied from their parent membership (`workspaces.rs` creator insert), and promoting anyone to admin requires `require_admin` (`users.rs`), so a non-admin forker cannot self-elevate. Workspace `owner` grants no admin power.
- Cloned `extra_perms` preserve the parent's access control in the fork.

So the forker's ability to invoke any cloned app — and therefore to benefit from its `on_behalf_of` run-as — is identical to what they already had in the parent:

- **Publisher/viewer** apps require app-read permission to invoke; the forker's read perms are unchanged, so any app they can invoke in the fork they could already invoke in the parent (where it also runs as the `on_behalf_of` identity).
- **Anonymous** apps require the secret app URL; obtaining it requires reading the app, which a non-privileged forker who lacked access in the parent still lacks in the fork.
- The cloned app executes against the fork's own **copied** data, not the parent's live data.

Revoking the parent app's public access doesn't leak anything: the fork is the member's own copy of a workspace they already had membership-level access to, and running the app as the `on_behalf_of` identity there yields nothing they couldn't already obtain in the parent.

(Unrelated: GHSA-hfh4 — `on_behalf_of_email` pinned to a superadmin sentinel — is a separate, genuine issue handled independently.)

Issue #17, opened 2022-05-10
Approval step for flows starting with interactive slack message to approver


rubenfiszel:
Step of flows are solely script or other flows for now.

An approval step could be defined as: This step sends an approval request using one of the selected channel and is waiting for an external signal that correspond to the approval to go to the next step. 

This could be used for instance to ask permission to the right resource owner or manager before proceeding with the rest of the flow.

PR #10129, opened 2026-07-15
feat(sessions): live DOM access for the raw-app preview in AI sessions

Guilhem-lm:
## Summary

In an AI session, when the side-panel chat has a raw-app preview open, the model could already read runtime console logs and job runs but was **blind to the rendered DOM** — it couldn't verify an edit landed, diagnose a blank render, or answer "what's on screen." This adds two complementary, read-only surfaces sharing one live mechanism, plus inspector-picked element chips synced bidirectionally with the preview.

- **AI pull** — `search_dom` / `read_dom` tools query the live preview DOM by CSS selector.
- **User push** — the preview inspector turns a picked element into a lightweight selector chip on the chat, kept in sync with the preview highlights.

The **selector is the reference**: no content store/snapshot — chips carry only a CSS selector, and the model fetches content on demand. Reads are live, same-origin `contentDocument` (the editor preview iframe is same-origin, no sandbox), reusing the attached-files engine over pretty-printed `outerHTML` (worker-guarded, bounded).

## Changes

- **`search_dom` / `read_dom` tools** (`global/core.ts`): session-gated (`SESSION_PREVIEW_TOOL_NAMES`), selector optional (omitted → whole `<body>`). Wired through the session runtime, mirroring the existing runtime-logs requester chain.
- **New `raw_apps/rawAppDom.ts`**: `prettyPrintHtml` + `runDomQueryOnHtml` (pure, testable) reusing `files/fileEngine` (`buildLineIndex`/`searchFilesInWorker`/`readFile`).
- **`RawAppEditor`**: host-side `requestDomQuery` (reads `contentDocument`, strips `.inspector-*` overlays so the model never sees them); pushes the chip selectors into the preview (`inspectorSetSelection`) and relays `inspectorSelect`/`inspectorDeselect`/`inspectorClear`.
- **Selector chips**: new `app_dom_selector` `ContextElement` + `ContextElementBadge` branch; rendered even in GLOBAL/session mode (where the general context row is hidden); one-shot per message; `ContextManager` add/remove/clear helpers.
- **Bidirectional multi-select sync**: chip list is the source of truth → the preview renders one highlight per selector; add, chip-remove, and preview-× remove all stay in sync.

## Dependency: ui_builder artifact (separate repo, not in this diff)

The preview inspector harness lives in **`windmill-code-ui-builder`** (`public/app-preview.html`), which windmill pins via `frontend/scripts/ui_builder_artifact.json`. This PR's chip sync + unique-selector behavior needs matching harness changes there:
- multi-select highlights (`inspectorSetSelection` / `inspectorDeselect`, one overlay per selector),
- unique `:nth-of-type` selectors (the old `getPath` was ambiguous for repeated rows/lists),
- a `handleClick` fix so the overlay × isn't swallowed.

Those ship via a ui_builder release + an `ui_builder_artifact.json` re-pin. **The windmill side in this PR is complete and backward-safe** (it only sends/relays messages the new harness understands).

## Screenshots

Multi-select sync — picking the 1st and 3rd "Delete" buttons (non-adjacent, proving `nth-of-type` uniqueness) shows two `button.btn` chips in the chat input and two matching green highlights (each with an × to remove) in the preview:

dom-multiselect-sync

## Test plan

- [x] Unit: `rawAppDom.test.ts` (pretty-print incl. void elements, search/read bounds, multi-match note); `core.test.ts` (tool dispatch both modes, session-only gating, `## SELECTED DOM ELEMENTS` message block).
- [x] `npm run check` clean (one pre-existing unrelated `utils_workspace_deploy.ts` error only).
- [x] Browser: session preview — inspector pick → chip appears; multi-select → N chips ↔ N highlights; remove via chat × → highlight clears; remove via preview × → chip removed; live `contentDocument` read verified same-origin.
- [ ] Re-verify end-to-end once the ui_builder artifact is re-pinned (dev used a locally-served harness).
- [ ] Ask the model something that triggers `search_dom`/`read_dom` and eyeball the returned pretty-printed HTML.


Guilhem-lm:
### Updated merge order — this PR merges **last**

The harness PR is now independent of the empty-render PR (#17 dropped from the chain). Sequence:

1. **windmill-code-ui-builder#18** (multi-select DOM inspector harness) — base `main`, merge first
2. **Release the ui_builder artifact** from its `main`; note the new `version` + `sha256`
3. **Update `frontend/scripts/ui_builder_artifact.json`** here to that pin (follow-up commit on this branch)
4. **Merge this PR**

Merging this before step 3 leaves `search_dom`/`read_dom` working but the chip↔preview highlight sync and unique selectors broken (old harness ignores `inspectorSetSelection` and emits ambiguous selectors). (Supersedes the earlier merge-order comment.)

Guilhem-lm:
On the two latest Codex P1s (context-snapshot timing):

- **Inline-prompt context race** — fixed in `ede83d9`. `onInlinePrompt` now snapshots the selection synchronously at submit time and sends it as `contextOverride` on both the queued and immediate paths, so a mid-preflight (or pre-flush) re-selection can't swap the element.
- **Queued DOM context vs active preview** — fixed in `69b931a`. Switching to a different raw-app preview tab now strips the queued turn's DOM chips belonging to other apps (`dropQueuedDomContextForOtherApps`), so a queued prompt can't ask the model to read one app's selectors against another's live DOM.

**Known follow-up (intentionally not in this PR):** the *general composer* send captures selected context after the async `beforeSend` preflight (a deliberate pre-existing ordering so the optimistic bubble shows before the workspace-fork step). That sub-second window affects all context types (files, code selections), not just this PR's DOM chips, so restructuring the core send order is out of scope here and tracked as a follow-up.

PR #10130, opened 2026-07-15
feat(ai-chat): image attachments and agent raw-app screenshots

hugocasa:
## Summary

Two related capabilities for the AI chat, scoped to global chat + `/sessions`:

1. **Image attachments** — drag/drop, paste, or pick images into the chat. Images ride the existing OpenAI `ChatCompletionMessageParam[]` format as `image_url` parts, converted per provider.
2. **`take_screenshot` agent tool** — lets the agent capture the live raw-app preview and see its own UI changes.

The two complement each other: when the capture is untrustworthy, the agent asks the user to screenshot manually and drag it in.

### Image pipeline

- **New `imageUtils.ts`** — downscale (longest edge ≤1568px), re-encode (PNG, JPEG past ~700KB **flattened onto white first** — canvas JPEG encoders composite alpha onto black, hiding dark strokes in transparent diagrams), parse/build image parts, input bounds (20MB byte cap *before* decode, 40MP pixel ceiling, 8 images/message).
- **Provider conversion** — Anthropic native (`image_url` → `{type:'image',source:{base64}}`) and OpenAI Responses (`→ input_image`). OpenAI-compatible Chat Completions passes through unchanged. All three verified at the wire (plus Bedrock, which translates to the AWS SDK).
- **Token estimation** — array content counts text as chars/4 plus a flat ~1200/image.
- **Out-of-band image storage** — chat records never persist image bytes. On save, `HistoryManager` swaps every data URL (user bubbles, tool cards, API messages) for a `wm-image:<id>` ref and writes the bytes once to an `images` IndexedDB store (chat-history schema v3); on load it hydrates them back — nothing outside `HistoryManager` ever sees a ref. The bubble, tool card, and API part share the exact same data URL, so they dedup to a single blob record; per-save chat writes stay O(text). Records persisted before the store (inline data URLs) load unchanged. The per-request 12MB image-byte bound applies only to the outgoing copy each loop iteration — stored history is never mutated. Retry/edit still read the model's own copies from API history (`storedImages`), so a rejection-stripped image is never silently resent.
- **Record-driven blob lifecycle** — each blob is owned by exactly one chat (the same image in two chats is two records), and each save derives the chat's *entire* stored set from the record being saved: ids ranked by their last reference, the newest 30 distinct kept (missing ones written, everything else deleted — an out-of-cap ref hydrates to an `[image omitted]` placeholder; blobs die with their chat). No persisted write-time ordering exists, so eviction is deterministic and idempotent across retries, re-attached images, and compaction. All history DB writes are serialized through a per-manager queue and pinned to the user-scoped database captured at enqueue time, so overlapping saves can't orphan each other's blobs and a queued write can't land in another account's database after an in-place user switch.
- **Compaction** — image parts stripped before the summarizer; boundary only lands on a *displayed* user message; drop-oldest marks fully-dropped messages with a negative index so a retry can't alias to another turn's images.

### Vision gating & failure recovery

- **`modelSupportsVision`** (`modelConfig.ts`) — a deny-list of exact **`provider:model`** pairs verified against provider API docs (an id proves nothing about a different endpoint; substrings would shadow `Mistral-Large-3` / `Phi-4-multimodal`). Deliberately permissive: unknown models are allowed and the failure path below recovers.
- Enforced at **attach** (refuse with a toast), and **per iteration inside `runChatLoop`**: the loop re-reads the model each iteration and strips image parts from that iteration's request only, so mid-turn model switches work in both directions (text-only → vision recovers the images; the history itself is never mutated).
- **Rejection fallback** — when a permissively-allowed model still rejects an image at the provider, the image parts are stripped from history (text survives, user is told, next turn succeeds). The match excludes the send-time *and* currently-selected model ids from the error text first, so a rate limit echoing `…-90b-vision-instruct` can't read as a rejection and destroy good images. Retry intentionally never resurrects a stripped image.
- An **image-only send** on a known text-only model restores the images to the composer instead of discarding them.

### Attach UI

- Drop / paste / picker with removable thumbnail chips; decode holds Send (spinner placeholder chips) so a message can never go out without an attachment already dropped.
- **Only directly-dropped images attach** (reserved synchronously from `dt.files` before any await — no traversal race); images *inside* a dropped folder are deliberately not promoted to message attachments on any browser. Cross-tab image drags (no filesystem backing, all FS handles null) fall back to the flat file list instead of silently dropping. The same fallback also makes handle-less *text*-file drops attach (previously a silent no-op).
- **Image-only messages are valid** (no text required); the send queue carries images, the queued chip shows image-only drafts, and empty instructions can't inherit stale text from a failed earlier turn.
- **`ExpandableImage`** — click any thumbnail (composer chip, bubble, tool card, job result, S3 preview) to view the full-resolution copy in a modal.

### `take_screenshot`

- GLOBAL **session** tool; captures the preview iframe via `modern-screenshot`, delivered as a follow-up user message after the tool batch (tool results are string-only). Routed to the *visible* raw-app tab.
- **Chromium-only** (see fidelity notes). Prompt guidance fires it only when the user raises how the app *looks* — not proactively after every edit.
- Captured at up to **2× CSS resolution** (capped by the 1568px budget): the re-render is vector, so the extra scale is real detail (~Anthropic's recommended ~1.15MP input), not interpolation. The tool card shows the same copy the model gets — persisted as one shared blob record in the `images` store — so the expanded view keeps full quality across reloads.
- Refuses up front on a known text-only model instead of buffering an image it can never read.

## Screenshots

`take_screenshot` result on the tool card, beside the live preview it captured:

pr-screenshot-tool

An image dragged into the chat, shown as a chip above the input:

pr-image-attach

## ai_evals A/B (global mode)

5 raw-app cases + a script-create control, sonnet, 2 attempts each:

| arm | pass | avg prompt tok |
|---|---|---|
| baseline (tool removed) | **12/12** | 170,900 |
| with tool (long def) | **12/12** | 183,067 |
| with tool (trimmed def) | **12/12** | 197,047 |

**No behavioural regression: 100% in all three arms, every case identical.**

The prompt-token column is **not** a usable signal: the *trimmed* schema produced the *highest* total, so the column is dominated by run-to-run iteration-count variance at n=2, which swamps a ~120 tok/iteration schema change. Do not read a schema cost out of these numbers.

The schema trim is justified deterministically instead: the tool def is re-sent every iteration, and it went 1,034 → 551 chars (~258 → ~138 tok) by moving the fidelity caveat from the def onto the tool *result*, where it only costs anything once a capture exists.

`--record` history is not updated: it only accepts full-suite runs.

## Notes on capture fidelity

There is no browser API to screenshot an element, so the image is reconstructed from the DOM (`foreignObject` → SVG → raster). That reconstruction can differ from what the user sees, and **differs by browser**.

**The tool is offered on Chromium only.** The `foreignObject` re-render is only faithful on Blink: on Gecko/WebKit the restricted SVG-image context shifts glyph spacing and line wrapping (font fallback, sub-pixel rounding — same class of bug reported against every library in this family, incl. snapDOM). On non-Chromium browsers `take_screenshot` is withheld from the tool set entirely and the session prompt instead tells the agent to ask the user to paste/drop a screenshot when their description isn't specific enough. Detection is `isChromiumBrowser()` (userAgentData brands, UA-string fallback).

A concrete instance, found and fixed here: on Firefox, a box that shrink-wraps its text can have **zero sub-pixel slack** (measured: a 208.59px box holding a 208.59px text run). The capture re-runs layout in whole pixels, the text no longer fits, wraps to a second line, and is clipped out of the box — the title vanishes while the emoji beside it survives. Chromium tolerates the same truncation. `pinSingleLineText` pins already-single-line runs for the duration of the capture (a no-op on the live DOM — verified zero layout shift), stopping the re-layout re-deciding line breaks. Only `white-space` values with an exact non-wrapping counterpart are touched (`normal`→`nowrap`, `pre-wrap`→`pre`); forcing `nowrap` onto `pre-wrap` would silently eat preserved whitespace.

Other gaps, measured by diffing `domToPng` against native across 20 CSS/DOM features on Chromium + Firefox: WebGL canvas is blank unless its context was created with `preserveDrawingBuffer`; `backdrop-filter` renders wrong; animations catch a different frame; typed `input` values may be missing. Clear: solid text, canvas2d, inline SVG, `filter`, `mix-blend-mode`, `clip-path`, `mask-image`, pseudo-elements, transforms, box-shadow, shadow DOM, web fonts, nested iframes.

The tool guidance deliberately does **not** enumerate these — the wrap bug above would not have been on any such list. It states the principle (DOM reconstruction, browser-dependent), requires the agent to name a specific cause in the source before blaming the capture, and otherwise to ask the user for a real screenshot.

## Review hardening

The branch went through 15+ review passes (local `local-review`/`local-review-codex`, then CI Claude/Pi/Codex rounds). Every confirmed finding is fixed with a regression test verified to fail without its fix, including: compaction slicing the visible transcript, queued sends dropping images, decode/ingestion races leaving Send enabled, the rejection fallback wedging or over-firing (model-id echoes, historical/screenshot images, retry resurrection), mid-turn model switches bypassing the vision gate, drop-oldest index aliasing, provider-id collisions in the deny-list, and JPEG alpha blackening.

## Test plan

- [x] Unit tests: `imageUtils`, `anthropic`, `openai-responses`, `chatLoop`, `AIChatManager`, `lib`, `isChromiumBrowser`, `global/core` — every review fix carries a regression test verified red without it
- [x] Wire-level attach verified against Anthropic, OpenAI Responses, OpenAI-compatible, and Bedrock
- [x] In-browser end-to-end: attach (drop/paste), image-only send, rejection fallback + retry, screenshot tool card + expanded view (2× capture measured), queued-image carry-through, cross-tab drag fallback
- [x] `ai_evals` A/B on GLOBAL — no regression
- [ ] Final manual provider pass in the UI before undrafting


hugocasa:
Re the Codex review (2026-07-17 06:58) — triage against head `6a5dbd8c0c`:

**Fixed in `6a5dbd8c0c`:**

1. **Cross-chat blob ownership** — `imageIdByUrl` is now keyed by `chatId + dataUrl`, so the same image pasted into two chats becomes two owned records; deleting or cap-evicting in one chat can no longer destroy bytes another chat references. Regression test: share bytes across two chats, delete the first, reload the second.
2. **Cap-eviction rotation on re-save** — `persistImageBlobs` now only writes the newest `MAX_IMAGES_PER_CHAT` blobs referenced by a save (dehydration walks messages oldest-first, so map order is chronological). An already-evicted older blob is never re-inserted, so repeated saves of an over-cap chat can't walk the eviction hole toward the newest attachment. The cap test now saves twice more after the first eviction and asserts the hole stays on the oldest image.

**Partially applied (3 — unbounded live history):** the real kernel here was `imageIdByUrl` pinning past chats' data-URL strings for the whole session; the chat-scoped keys above plus pruning on chat rotation/load fix that. The rest is by design and the cost claim doesn't hold: per-save "cloning" of base64 is a string reference copy (JS strings are immutable; the snapshot doesn't copy bytes), and the IndexedDB `put` serializes refs, not blobs — so save latency is O(text) regardless of image count. In-memory, the live arrays hold what the wire needs (stateless APIs resend history; the 12MB per-request bound and compaction bound the effective window).

**Not applied (4 — concurrent decodes across paste events):** `remaining` subtracts `pendingImages`, so overlapping paste/drop events share the 8-image budget — total in-flight files are capped at 8 across invocations, each invocation decoding strictly one at a time. Reaching a harmful allocation requires several simultaneous decompression-bomb PNGs under the 20MB file cap, which is self-inflicted rather than a realistic user path (same conclusion as the Claude pass on this head).

**Not applied (5 — refusal restore vs. typing during the pre-send await):** same race as previously declined on an earlier round (composer restore vs. typing during linked-folder refresh). The window is the pre-send permission/refresh awaits (typically milliseconds), requires the user to have image-attached, mode-switched, sent, and started a new draft inside it, and the failure mode is a lost restore of a draft the user just abandoned mid-keystroke. Restoring earlier is not an option: the composer clears synchronously after the un-awaited `sendRequest`, which would wipe the restore (that ordering is itself a prior review fix with a regression test pinning it).

Re the Claude pass's comment-length nit: acknowledged; leaving as-is for now since the same blocks were explicitly checked as compliant rationale by the Pi pass on the same head, and trimming them mid-review-loop trades known-good context for line count.

hugocasa:
Re the Codex review (07:10) — confirmed and fixed in `0b96f9b2ad`: `dehydrateImages` now walks the display transcript (always whole and chronological) before the API history, so a drop-oldest-compacted message's still-displayed image keeps its true age instead of slotting after the API tail; shared data URLs keep the transcript's map position on the later walk. Regression test per the suggested shape: 31 transcript images over a 4-message compacted API tail, saved twice, reloaded — only the oldest is omitted, verified failing on the previous head.

hugocasa:
Re the Codex review (07:19) — both findings confirmed, and rather than patching the derived-age contract a third time, `e92a83f9c0` removes it: eviction is now derived entirely from the saved record. `persistImageBlobs` ranks ids by their LAST reference in the record, keeps the newest 30 distinct, writes any kept blob that's missing, and deletes everything else the chat owns. Persisted `savedAt` no longer participates in ordering (the monotonic-timestamp hack is gone too). This fixes both findings by construction — truncated turns' blobs fall out of the record and are deleted (finding 1), and a re-attached image's newest reference ranks it newest, legitimately re-writing it if it had been evicted (finding 2) — and it makes eviction deterministic and idempotent across repeated saves, compaction, and retries. Regression tests for both scenarios (truncate-then-add, re-attach-after-30-distinct), each verified failing on the previous head; the earlier rotation/chronology/cross-chat tests all still pass under the new scheme.

hugocasa:
Re the Codex review (07:30) — confirmed and fixed in `8f1d46bcb0`: all history DB writes (saveChat's blob-write + eviction + record put, deletePastChat, tagChatWithSession) now run through a per-manager promise queue, so an older save's delete pass can no longer interleave with a newer save's existence check. A failed write rethrows to its caller without wedging the queue. Regression test constructs the exact overlap (older snapshot dropping an image X while a newer overlapping save re-references X as its newest image plus a new Y) and asserts every ref in the winning record hydrates after reload — verified deterministically failing on the previous head.

hugocasa:
Re the Codex review (07:43) — confirmed and fixed in `dd3c6c7b00`: `enqueueDbWrite` now captures the user-scoped DB name at enqueue time and skips execution when the logged-in identity (or the resolved handle's name) no longer matches — a queued write initiated under user A is dropped rather than landing in user B's database after an in-place account switch. Regression test enqueues two image-bearing saves under A, switches identity before they execute, and asserts neither the record nor the blob reaches B's database — verified deterministically failing on the previous head. The existing in-place-switch test (a save *initiated* after the switch must go to the new user's DB) still passes. PR description also refreshed to describe the final record-driven blob lifecycle and the serialized, identity-pinned write queue.

hugocasa:
Re the Codex review (07:55) — confirmed and fixed in `54551718c2`, via write ordering rather than a multi-store transaction: kept blobs land first, then the chat record, and stale-blob deletion runs only after the record commit. A failure at any step now leaves the last committed record fully hydratable — a failed record put leaves the old record and all its blobs intact (at worst, freshly written blobs linger as orphans until the next successful save's delete pass reclaims them, since that pass diffs against the record rather than trusting prior bookkeeping). Regression test forces the `chats` put to fail on a save that would evict the previous record's blob and asserts the old record still hydrates after reload — verified deterministically failing on the previous head.

hugocasa:
Re the Codex review (08:07) — confirmed and fixed in `07450afa54`: the image-only vision refusal now skips the composer restore when `options.queued` is set, matching the non-GLOBAL and preflight-refusal branches — the caller re-queues a refused queued draft, so it stays in exactly one place. Regression test asserts a queued image-only draft refused by a text-only model touches neither `prependText` nor `restoreInstructions` — verified failing on the previous head.

hugocasa:
Re the Codex review (08:19) — both P2s fixed in `eac84c0988`: an image-only chat now titles itself from the first attachment's filename ('Image attachment' when nameless) instead of persisting an empty History label, and an image-only bubble whose blob was cap-evicted hydrates with the `[image omitted]` marker instead of rendering an empty bubble after reload. Both with regression tests verified failing on the previous head.

hugocasa:
Re the Codex review (08:31) — both fixed in `da2e4a5f6b`:

1. (P1) The in-memory `savedChats` mirror now keeps its inline data URLs — only a `structuredClone` bound for the DB is dehydrated (string values are shared, so this copies structure, not bytes). When `whenReady()` degrades to in-memory-only (private browsing / blocked opens), a reloaded chat keeps working thumbnails and retries resend real images. `tagChatWithSession` routes through the same dehydrating path so an inline mirror record can't reach the DB, and `dehydrateImages` now counts pre-existing refs as references (so a tag re-save of a never-rehydrated record can't let the stale-delete pass reclaim its blobs). Regression test saves and reloads an image-bearing chat with `indexedDB.open` throwing.
2. (P2) Title derivation treats the hydrated `[image omitted]` marker as no-text and falls back to the attachment name, then the existing title — so continuing an evicted image-only chat after reload keeps its filename label.

Both tests verified failing on the previous head.

hugocasa:
Re the Codex review (08:47) — confirmed and fixed in `3b84d74556`: the mirror takes the inline copy only transiently. Once the queued DB commit succeeds, `persistDehydratedChat` converges `savedChats` to the ref-backed clone — so DB-backed chats hold refs (no session-long pinning of rotated chats' image bytes, and reopening hydrates through the store, reseeding stable blob ids instead of minting new ones and bulk-rewriting), while the no-DB fallback keeps the inline copy because the commit is never reached. An existence check keeps a concurrent `deletePastChat` from being resurrected; queue FIFO ordering makes last-writer-wins correct otherwise. Regression test: save + rotate + reopen + re-save, asserting the stored blob keys are unchanged — verified failing on the previous head; the DB-less in-memory test still passes.

hugocasa:
Re the Codex review (08:58) — confirmed and fixed in `fa5ede1b73`: convergence is now revision-guarded. Every mirror write bumps a per-chat revision inside `persistDehydratedChat`; the queued convergence only replaces the mirror when its captured revision is still current, so an older save completing while newer writes wait can no longer rewind the mirror that later saves read their omitted-field fallbacks (modifiedItems/backgroundJobs) from. `deletePastChat` invalidates the revision, which also covers the resurrection case the previous existence check handled. Regression test interleaves a modifiedItems save, an overlapping backgroundJobs save, and a third save after the first completes, asserting both newest fields survive in memory and in the final DB record — verified deterministically failing on the previous head.

hugocasa:
Re the Codex review (09:12) — both fixed in `5a789f601c`:

1. (P1) `init()` now invalidates every pending convergence and cached blob id when it (re)adopts an identity, so a prior user's write that already passed the enqueue-time identity checks cannot merge its record into the new user's mirror when its transaction finishes. Working the regression test also surfaced an adjacent defect in the same window, fixed in the same commit: the account switch closes the old handle mid-op, which made the *committed* save reject when its stale-blob cleanup tail hit the closed connection — that tail is now best-effort (the orphans were already documented as reclaimed by the next save's pass). The test drives the exact scenario (identity switch + re-init fired from inside the save's chats put): the save must resolve and the new user's mirror must end empty — deterministically failing on the previous head.
2. (P2) `storedImages` now recovers attachment filenames from the transcript bubble's entries (the wire format has no name field), so retrying or editing a named image-only turn resends it with its name intact and the history title keeps its filename label instead of degrading to the generic fallback. Unit test red-verified.

One honest caveat on the P1: in the fuzzed harness the durable form of the leak (convergence landing after the new user's init completes) is not reachable — init always finishes later than the convergence, so pre-fix the leaked entry was transiently visible and then wiped. The guard closes the production window (real IndexedDB timings don't order those two the same way), and the committed regression test pins the crash-tolerance half deterministically.

hugocasa:
Re the Codex review (09:30) — confirmed and fixed in `f17aaf8df8`: the scale floor was the bug (a body taller than 1568 CSS px never scaled down, so a tall scrolling app rasterised at full height only for normalize to shrink or reject it after the allocation). The scale math is extracted to `captureScale` in imageUtils — up to 2× for small targets, deliberately sub-1× for oversized bodies, never a raster over `MAX_IMAGE_EDGE` — with `maximumCanvasSize: MAX_IMAGE_EDGE` passed to `domToPng` as the library-level belt over it. For every preview ≤1568 CSS px the old and new formulas agree exactly, so the normal capture path is unchanged. Unit tests pin both properties (2× for small targets, raster ≤1568 even below 1×); the component call site isn't unit-tested per convention, but the behavioural change lives entirely in the extracted function. Also added the direct `appendPendingToolImages` regression the coverage note asked for: one user message appended to BOTH arrays (same object), buffer drained exactly once.

hugocasa:
Post-approval simplification (`868ee7a778`), by explicit maintainer decision: the no-IndexedDB in-memory image fallback is removed — no other `userScopedDb` consumer (sessions, artifacts) special-cases that environment; they all silently skip persistence, and chat history now does the same. The mirror always holds the dehydrated (ref-backed) record again, written synchronously at save time, which deletes the inline-then-converge machinery wholesale: the convergence block, the `mirrorRev` revision guard, the init-time invalidation, and the `structuredClone` second copy (net −28 lines, and the subtlest code in the file). Safe degradation replaces it: `hydrateImages` now accepts an absent DB, so a chat loaded without IndexedDB degrades its refs to `[image omitted]` placeholders instead of leaking `wm-image:` URLs into bubbles or requests — pinned by a new test asserting no ref ever escapes. The three concurrency regressions from the convergence era (metadata never rewound, no cross-user mirror leak, reopened chats reuse blob ids) all still pass unchanged: the invariants hold by construction now that the mirror has a single, synchronous writer. Everything else — write queue, enqueue-time identity pinning, crash-safe write ordering, record-driven eviction — is untouched.

hugocasa:
Re the Codex review (10:03) — not applied, by explicit maintainer decision (see the comment accompanying `868ee7a778`). The behavior described is accurate and intentional: when IndexedDB is unavailable, chat history degrades to not persisting images across chat rotation, exactly like every other `userScopedDb` consumer (sessions, artifacts) degrades in that environment — none of them keep an in-memory fallback. The two suggested remedies are precisely the machinery that commit removed: 'preserve an inline memory copy' was the first implementation (it pinned rotated chats' bytes in memory and broke stable blob ids on reopen — flagged by the 08:47 Codex review), and 'dehydrate only a DB-bound clone' with a converging mirror was the second (it needed a revision guard against rewinding concurrent metadata — 08:58 — and identity invalidation — 09:12). The full chain cost ~60 lines of the subtlest code in the file to serve an environment (blocked/corrupt IndexedDB; modern private-browsing modes ship ephemeral IDB) that no other feature serves either. What IS guaranteed in that environment, and test-pinned: refs never leak into bubbles or outgoing requests — they degrade to the `[image omitted]` placeholder, and the live (unrotated) chat keeps its images since live arrays never hold refs.

PR #10133, opened 2026-07-15
[ee] fix(bash): strip carriage returns from bash script content before execution

alpetric:
## Problem

Bash scripts authored with CRLF line endings fail at runtime with `$'\r': command not found`. The Monaco editor defaults to CRLF on Windows, and the bash executor wrote the raw content to `main.sh` without normalizing line endings. PR #6889 fixed CRLF handling in the bash *parser* (signature extraction), but the *executor* path was never fixed.

## Fix

Strip `\r` from the script content before writing `main.sh`, in both execution paths:

- **Local bash executor** (`backend/windmill-worker/src/bash_executor.rs`)
- **SSH executor** (`windmill-ee-private` → `windmill-worker/src/ssh_executor_ee.rs`) — the `#ssh` path that runs bash content on a remote host

`\r` is never meaningful in a bash script, so stripping it is safe.

## Verification

Ran a bash preview job whose content used `\r\n` line endings against a locally running backend:

```
echo "line one"\r\necho "line two"\r\nx=5\r\necho "x is $x"\r\n
```

- Before the fix this fails with `$'\r': command not found`.
- After the fix the job succeeds — logs show `line one` / `line two` / `x is 5` and the result is `x is 5`.

## EE companion PR

windmill-ee-private branch: `alex/win-2191-strip-carriage-returns-from-bash-script-content-before` (`ee-repo-ref.txt` updated to that commit).

Fixes WIN-2191


alpetric:
EE companion PR: windmill-labs/windmill-ee-private#663

rubenfiszel:
Superseded by #10131, which merged into `main` (6407d9ff5c) and fixes the same WIN-2191 CRLF bug.

The merged fix normalizes at the **top** of `handle_bash_job` (before the annotation dispatch), so it covers more than the write_file-site strip here:
- local bash executor
- `# sandbox` / `# docker` scripts (content is cleaned before the `docker_v2` dispatch, which this PR's placement returns before)
- directive parsing (`# sandbox`/`#ssh` no longer see a trailing \r)
- the SSH executor — clean content is passed into `handle_ssh_bash_job`, which is why the EE companion #663 is also unnecessary

It also converts CR to `\n` (`.replace("\r\n","\n").replace('\r',"\n")`) rather than deleting, so a bare/lone \r splits the line instead of merging two commands.

Closing as redundant — thanks for the parallel fix and the end-to-end verification, that was useful confirmation.

PR #10136, opened 2026-07-15
fix(frontend): show friendly draft path for draft-only items in pickers

Guilhem-lm:
## Summary

The file/item pickers (session preview router picker, editor breadcrumb drill picker, chat context picker) showed never-deployed drafts by their `u/{user}/draft_<uuid>` storage path instead of the user-typed friendly path. The backend list endpoints (`listScripts`/`listFlows`/`listApps`, `drafts/list`) already return the friendly `draft_path`; the picker layer dropped it.

On top of the display gap, chat edits were actively **erasing** the friendly name for apps: `normalizeAppDraftValue` is a field whitelist and dropped `draft_path` from the persisted draft on every app write tool call (scripts/flows were immune because their chat writes spread the raw base draft).

The fix introduces a display-only `WorkspaceItem.draftPath`: labels, secondary lines, folder grouping and sorting use the friendly path via `workspaceItemDisplayPath()`, while leaf keys, dedupe, and navigation stay on the storage `path` (routing to the friendly path 404s).

## Changes

- `workspacePicker.ts`: new display-only `draftPath` field + `workspaceItemDisplayPath()` helper; `loadKind` maps `draft_path` from the three list endpoints.
- `workspaceTree.ts`: leaves are labeled, grouped, and sorted by the display path; leaf keys stay storage-keyed. `withExtras` and `withCurrent` dedupe match both `path` and `draftPath`, so a live editor cell (surfaced under its friendly path) or the injected current item can't render as a second leaf next to the loaded storage-path row; `current` marking also matches via `draftPath`.
- `copilot/chat/global/workspaceItems.ts` + `userDraftAdapter.ts`: `AppDraftValue.draft_path` round-trips through the `normalizeAppDraftValue` whitelist (fixes the erasure); draft→item converters and the backend `drafts/list` row mapper surface `draftPath`, with a live editor's fresher effective path taking precedence.
- `PreviewRouterPicker.svelte`, `WorkspaceItemDrillPicker.svelte`, `ChatContextPicker.svelte`: secondary lines use the display path; the drafts overlay carries `draftPath`.
- `sessionPreviewTabs.svelte.ts` + `sessionState.svelte.ts` + `SessionEditorTarget.svelte` + `sessions/+page.svelte`: the live editor stamps the full friendly path (`friendlyPath`, transient like `friendlyLabel`) onto its tab, and the active-tab picker derives its scope from it — the tree now groups the draft under its friendly folder, so scoping by the storage dir would open the picker on a folder that no longer shows the item.
- `workspaceItemsLoader.svelte.ts`: `itemsEqual` compares `draftPath` so cached lists refresh when it changes.

## Test plan

- [x] `workspaceTree.test.ts`: friendly grouping with storage-keyed leaf, friendly secondary, live-cell extra dedupe, current-item dedupe + mid-rename via `draftPath` (36 pass)
- [x] `core.test.ts`: `draft_path` survives the `normalizeAppDraftValue` whitelist through a chat `write_app_file` edit (mirrors the existing `parent_version` survival test)
- [x] `sessionPreviewTabs.test.ts`: `friendlyPath` stamped and cleared on retarget
- [x] `npm run check` — no new errors (one pre-existing failure in `utils_workspace_deploy.ts`, present on main)
- [x] Browser-verified on a session in a workspace with a draft-only raw app (storage `u/admin/draft_851d…`, friendly `u/admin/thrilling_app`): picker search and folder browse show `thrilling_app` with no `draft_<uuid>` entry, picking it opens the working raw-app editor, and the active tab's breadcrumb picker opens scoped to the item's folder with it highlighted

## Screenshots

Preview router picker listing the draft-only apps by their friendly names (previously `draft_<uuid>`):

picker-pretty-path-list

Picking the draft opens its raw-app editor in a tab labeled with the friendly name:

picker-pretty-path-opened


Guilhem-lm:
Addressed the Codex P1 in 3008e248de: `DrillPicker`'s highlight-reset fallback now prefers the visible leaf marked `current` over the first row. When the loaded list replaces the synthetic friendly-keyed current-item leaf with the storage-keyed row (whose key differs from the `initialHighlight` the editor breadcrumb passes), the highlight reconciles onto the current item instead of silently landing on an unrelated sibling — so Enter can no longer navigate away from the open draft-only item.

Browser-verified per the suggested scenario: opened the breadcrumb picker on a draft-only raw app in a folder with 9 siblings after the list loaded — the item stayed highlighted, and Enter kept the editor on the `draft_<uuid>` storage path.

Guilhem-lm:
Addressed the second Codex P1 in 80f4280ae5: the active-tab picker content is now wrapped in `{#key activePickerScope?.dir}`, so when `friendlyPath` is stamped after the editor cell loads (e.g. a picker opened immediately after session restore, still scoped to the `draft_<uuid>` storage folder), the picker remounts with the friendly-folder scope and the storage-keyed highlight — landing on the current item instead of stranding the highlight on a storage-folder sibling.

Verified the exact race in the browser by delaying the apps API response: pre-stamp the picker showed `All › u/admin` with the raw `draft_…` row; when the stamp landed it remounted to `All › f/pretty` with the item highlighted, and Enter kept the tab on the storage path.

Guilhem-lm:
Addressed the third Codex P1 in f57736e0a2: `SessionEditorTarget` now stamps the tab's `friendlyPath` whenever the staged path differs from the tab's route (storage) path — independent of whether `draftFriendlyLeaf` yields a label (which only applies to `draft_<uuid>`-parked items). A deployed item with an undeployed rename is regrouped under the typed folder by the picker tree via the live-draft overlay, so the breadcrumb picker scope now follows it there too.

Verified per the suggested scenario: opened the deployed `u/admin/hello_world` script in a session, staged a rename to a different folder via the editor's Path widget (draft persisted with `draft_path`, not deployed), then opened the active-tab picker — it scoped to the staged folder with the storage-keyed row highlighted, and Enter did not navigate away.

Guilhem-lm:
Addressed the fourth Codex P1 in b2000a527a, on both fronts suggested:

- **Registration**: `FlowEditorView` / `RawAppEditorView` now register `draft_path ?? path` as the live editor's effective path — for these kinds the typed rename lives in the value's `draft_path` while `val.path` is the storage key (unlike scripts, where `val.path` is the typed name).
- **Adapter defense**: `userDraftAdapter` no longer strips a live entry's `draftPath` (both the cell converters and the backend-row mapper keep it unless it merely repeats the display path), so even a registration whose effective path is the storage key cannot hide a staged rename from `listGlobalDrafts` — the pickers then regroup and highlight the item under the staged folder.

Added the requested regression through `listGlobalDrafts`: a live raw app registered at its storage key with a staged `draft_path` lists with `draftPath` = the staged path (the `listDrafts` test mock now surfaces `draft_path` like the real endpoint). All 139 core tests pass; tree/sessions suites green.

PR #10140, opened 2026-07-15
fix(mcp): push granular scope patterns into SQL so scoped scripts/flows aren't truncated

alpetric:
## Summary

Two related changes around the MCP tool cap (`ITEMS_FETCH_MAX_LIMIT = 100`):

### 1. Backend fix — scoped scripts/flows were truncated before filtering

MCP `list_tools` returned **zero scripts/flows** for a granular (folder/custom-scoped) token in any workspace with more than 100 scripts or flows. The listing fetched the 100 newest items (`ORDER BY created_at DESC LIMIT 100`) and only *then* applied the token's scope patterns in Rust — so in-scope items outside that newest-100 window were truncated before the filter ran. Order-of-operations bug: cap-then-filter instead of filter-then-cap. The fix pushes the scope patterns into the SQL query so the filter applies **before** the `LIMIT`.

### 2. Frontend — warn when a scope still exceeds the cap

Even with the fix, a scope matching more than 100 scripts (or 100 flows) silently drops the overflow, bloating the assistant's context with a partial, arbitrary tool set. `McpScopeSelector` now computes how many scripts/flows the selected scope would expose (per type) and shows a warning when either exceeds the cap.

## Changes

**Backend**
- Add `PathFilter` enum (`windmill-mcp/src/server/backend.rs`): `Prefix(&str)` (hashed-name resolution) and `Patterns(&[String])` (scope patterns).
- `get_items` (`windmill-api/src/mcp/utils.rs`) takes `Option<PathFilter>` and builds the WHERE condition via a new pure `scope_patterns_condition`, mirroring `is_resource_allowed`: `*` → no filter, exact → `o.path = ?`, `x/*` → `o.path = 'x' OR o.path LIKE 'x/%'`, empty → `false`. Values bound via `sql-builder` (`.bind()`); LIKE wildcards escaped.
- `list_tools_single` passes the token's script/flow scope patterns for granular tokens; the Rust `is_allowed` post-filter stays as defense in depth. Favorites/`mcp:all` unaffected.

**Frontend**
- `McpScopeSelector.svelte` computes per-type exposed counts (async, with a sequence guard against stale results) and renders a warning `Alert` when scripts or flows exceed `MCP_TOOL_FETCH_LIMIT` (100, mirroring the backend constant). The count reuses a single per-type cache shared with the preview list (no duplicate fetch). A folder scope is fetched with a boundary-anchored prefix (`f/{folder}/`, so the backend never returns a prefix-sharing sibling like `f/team2` that could otherwise fill the page and evict the target folder's rows), and the copy says "most recent" (flows order by `edited_at`, not `created_at`).

## Screenshots

**All scripts/flows scope — 132 scripts (> 100), warning shown:**

all-scope warning

**Folders scope → `team` (3 scripts) + `team2` (105) — 108 total, warning with corrected copy** (verifies boundary-aware counting: selecting `team` alone shows no warning and does not include the prefix-sibling `team2`):

folder-scope warning

## Test plan

- [x] `cargo check -p windmill-mcp` and `cargo check -p windmill-api --features mcp` compile.
- [x] 6 unit tests for `scope_patterns_condition` (wildcard, empty, exact, subtree, mixed-OR, LIKE-escape) pass.
- [x] Existing `windmill-mcp` tests (54) pass.
- [x] Backend verified against real Postgres: seeded 120 recent noise scripts + older `f/sdr/*` (8) and `f/a_b/*` (4) scripts (in-scope scripts outside the newest-100 window). OLD query (cap-then-filter) returned **0** in-scope; fixed query returned all of them, including the subtree/underscore-folder matches.
- [x] **Live MCP end-to-end** (workspace `mcptest`, 132 scripts, real `@modelcontextprotocol/inspector` client): a `mcp:scripts:f/sdr/*` token returns **exactly the 8 `f/sdr/*` tools** (0 of the 120 noise, 0 endpoint tools); a `mcp:scripts:f/a_b/*` token returns **exactly the 4 `f/a_b/*` tools** — confirming filter-then-cap and underscore-folder LIKE-escaping through the real MCP protocol. (Note: exercising the local MCP endpoint requires building **without** the `otel` feature — an unrelated pre-existing EE OTEL×MCP crash — e.g. `--features enterprise,ce_core,mcp`.)
- [x] Frontend verified via Playwright: Favorites → no warning; All scripts/flows → "132 scripts" warning; Folders → `noise` → "120 scripts" warning; switching back to Favorites clears it (see screenshots).


alpetric:
Addressed the latest Codex nit in 46837362a8: the folder count now fetches the boundary-anchored prefix `f/{folder}/` (backend `path LIKE 'f/{folder}/%'`), so a prefix-sharing sibling like `f/team2` is never returned and can't fill the page ahead of the target folder's rows. Verified in the browser: selecting `team` (3) with a sibling `team2` (105) present shows no warning; `team`+`team2` shows the 108 warning. The earlier findings (duplicate fetches, folder overcount, "most recent" copy, screenshot) were addressed in aaa7ef5266.

PR #10143, opened 2026-07-16
fix(alerts): identify server replica in low-disk alert + per-host dedup tag

rubenfiszel:
EE PR: https://github.com/windmill-labs/windmill-ee-private/pull/666

The functional fix is in `windmill-common/src/ee.rs` (symlinked from `windmill-ee-private`, so it does not show in this diff). This PR carries the matching `ee-repo-ref.txt` bump, a regression test, and a comment correction.

## Problem

A server replica fires "out of disk" alerts whose free space bounces low/high roughly every monitor cycle, and adding 100GB of disk does not stop it.

The server-mode low-disk alert in `low_disk_alerts` keyed its dedup tag on the mountpoint alone:

```
low-disk-v2-server@{fs_mounted_on}
```

`simple_alert_helper` looks up `healthchecks WHERE check_type = $1`, so that tag maps **every server replica onto one shared alert row per mountpoint**.

With more than one replica the row flaps: a replica seeing low disk does `create_alert` + `report_critical_error`, while a replica seeing healthy disk sees `is_reported = true` and does `report_recovered_critical_error` + `mark_alert_as_recovered` on that same row. That repeats every monitor pass, producing an alert/recovery pair per cycle. Adding disk does not help, because the flap is driven by disagreement between replicas rather than by absolute free space.

The message also read just `(server)`, so it could never name the affected replica.

The worker branch immediately above already includes the worker name in both the message and the tag (`low-disk-v2-wk@{mount}@{wk}`); the server branch never got the same treatment.

## Fix (in the EE PR)

Append `crate::utils::HOSTNAME` to both the message and the dedup tag, mirroring the worker path.

| | before | after |
|---|---|---|
| message | ``Disk mounted on `/` is low, 900 MB available. (server)`` | ``Disk mounted on `/` is low, 900 MB available. (server srv-7c9f)`` |
| tag | `low-disk-v2-server@/` | `low-disk-v2-server@/@srv-7c9f` |

Each replica now owns its own alert row, so replicas cannot recover each other's alerts.

## In this PR

- `backend/ee-repo-ref.txt` — bump to the EE commit.
- `backend/windmill-common/tests/low_disk_alerts.rs` — regression test driving the real `low_disk_alerts` and asserting the server tag is per-host.
- `backend/src/monitor.rs` — comment-only fix. Two comments described the cadence as "~60s (2 iterations * 30s)"; iterations are `LISTEN_NEW_EVENTS_INTERVAL_SEC` (10s by default), so the arithmetic was wrong on both factors. Now "~20s at the default". No behavior change.

## Scope notes

- The server-side recovery scan at the top of `low_disk_alerts` (`check_type LIKE 'low-disk-v2-wk@%'`) reconciles **worker** alerts from the server and is deliberately unchanged.
- The 48h cleanup (`LIKE 'low-disk-%'`) still matches the new tag.
- `low-disk-db` stays host-less on purpose: the database really is one shared resource, so a single row across replicas is correct there.
- No consumer parses these tags outside `ee.rs` (no frontend or SQL dependency on the old format). In-flight alerts under the old tag age out via the existing 48h cleanup.

## Validation

- `cargo check --features enterprise,private` and plain `cargo check` both pass.
- The regression test fails on the pre-fix code with `server tag low-disk-v2-server@/ is not per-host; replicas would share one alert row`, and passes after:
  ```
  cargo test -p windmill-common --test low_disk_alerts --features private,enterprise -- --ignored
  ```
- Flap reproduced directly against `simple_alert_helper` with two simulated replicas (A low, B healthy) over 3 monitor cycles:

  | tag | raises / 3 cycles | end state |
  |---|---|---|
  | shared `low-disk-v2-server@/` | **3** | rows left `healthy=true` while disk still low |
  | per-host `...@server-a` / `...@server-b` | **1** | A's row stays unhealthy, B never touches it |

---
## Summary by cubic
Make server low-disk alerts per-host and widen the alert tag column to `text` so tags with hostnames and mountpoints fit. This prevents cross-replica flapping and overflow-driven re-alert loops.

- **Bug Fixes**
  - Server low-disk alert now dedupes per host: tag is `low-disk-v2-server@{mount}@{hostname}` and the message names the replica; adds a regression test and fixes monitor cadence comments (~20s default).
  - Widened `healthchecks.check_type` to `text` so long tags persist, fixing silent insert failures that caused repeated alerts without recording recovery.

- **Migration**
  - Alters `healthchecks.check_type` to `text`; the down migration truncates to 50 chars for rollback.

<sup>Written for commit 9a90e271a4d274f8d6625d784a2b4a7090c7d693. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed the Codex P1 in b94649e5b3. It is real, and it turned out to be broader than the server tag.

**Confirmed the overflow.** `healthchecks.check_type` is `varchar(50)`. `low-disk-v2-server@` is 19 chars, so at mount `/` the hostname budget is 29, and a standard pod name like `windmill-server-7d9f8b6c4d-x2k9p` (32) already blows it:

```
ERROR:  value too long for type character varying(50)
```

**The failure mode is worse than a dropped row.** `create_alert` only logs the insert error, but `report_critical_error` fires regardless. So `check_existing_alert` stays false, and every monitor pass (~20s) re-sends a fresh critical alert that never recovers: a worse flap than the one this PR set out to fix, in exactly the multi-replica k8s deployment it targets.

**The worker tag already overflows today, and not only via the mountpoint.** Three unbounded inputs feed these tags: the mountpoint, the hostname, and the worker group. The worker name is `wk-{group}-{instance_name}-{rd5}`, and `WORKER_GROUP` is an unbounded env var (`windmill-common/src/worker.rs:140`), so at mount `/` the budget is `len(group) + len(instance_name) <= 23`.

`instance_name` is `hostname.split("-").last()` with underscores stripped first (`utils.rs:259`), so it only shortens the hostname when the hostname contains a dash. Kubernetes pod names are dashed and collapse to ~5 chars, which is the main reason this has not surfaced as a bug report. With the **default** worker group and no custom mount:

| hostname | instance_name | worker tag |
|---|---|---|
| `windmill-server-7d9f8b6c4d-x2k9p` | `x2k9p` | 39, ok |
| `windmill_worker_1` | `windmillworker1` | 49, ok by one char |
| `myproject_windmill_worker_1` | `myprojectwindmillworker1` | 58, **overflow** |
| `windmillproductionworker01` | `windmillproductionworker01` | 60, **overflow** |

A separately mounted cache or logs volume overflows it as well (`/tmp/windmill/cache/` with a default group and a k8s pod name is 58 chars), but as the table shows it does not take one. A descriptive group name such as `high-memory-data-pipeline-workers` overflows at mount `/` on its own (65 chars). Plain Compose sits at 49 of 50, and a Compose project prefix on the service hostname is enough to break it. So `varchar(50)` was not really protecting anything here; deployments have been passing by luck, and the luck came from k8s using dashed pod names.

**Why widen rather than bound the identifier.** With three unbounded inputs, bounding any one of them does not close this. `/tmp/windmill/cache_nomount/` alone leaves a 2-char budget for the hostname. `ALTER COLUMN check_type TYPE text` is a no-rewrite change in PG and needs no index rebuild (verified: index intact, column now `text`).

**Regression test.** Rather than adding a second test, I strengthened the existing one to force `FORCE_HOSTNAME` to a pod-length name so the tag runs past 50. Asserting that the row persists now pins the column width as well as the tag shape. Verified it fails without the migration and passes with it:

```
# migration removed
panicked: expected at least one server low-disk alert; an alert whose tag does
not fit check_type is dropped here while its notification still fires
# migration restored
test tests::server_low_disk_tag_is_per_host ... ok
```

I exercised the worker path end to end too (0 rows persisted on `varchar(50)`, 1 row on `text`, 65-char tag) but shipped no test for it: it pins the same column-width invariant as the server test, and both fail together if the column is ever narrowed.

Also updated the stale `Varchar` -> `Text` param in the `.sqlx` cache (the filename hashes the query text, which is unchanged) and `summarized_schema.txt`, and confirmed `SQLX_OFFLINE=true cargo check -p windmill-common --features private,enterprise` still builds.

PR #10144, opened 2026-07-16
feat(alerts): include disk total and top consumers in low-disk alert

rubenfiszel:
## Summary

The low-disk alert reported only the available bytes and the mountpoint, which does not tell an operator what is actually filling the disk — the recurring follow-up question on every "out of disk" alert is simply *what filled it?*

This makes the alert self-diagnosing: it now reports the filesystem total alongside the available space, and appends a breakdown of the largest space consumers under the Windmill directories on that mount.

Before:
```
Disk mounted on `/tmp` is low, 812 MiB available. (server)
```

After:
```
Disk mounted on `/tmp` is low, 812 MiB of 100 GiB available. (server)

Largest entries under the Windmill directories: /tmp/windmill/cache/python (14.2 GB), /tmp/windmill/cache/deno (3.1 GB), /tmp/windmill/logs/job.log (220 MB).
```

Message/observability only: **thresholds, tags and alert dedup keys are unchanged** (the per-host server tag fix is a separate change, deliberately untouched here to avoid conflicts).

The substantive diff is in the symlinked EE file `windmill-common/src/ee.rs` — see the companion PR. The OSS diff is the `ee-repo-ref.txt` bump plus one call-site update.

## Changes

- Report `{avail} of {total} available` in both the worker and server alert branches (and in their recovery messages).
- Add `top_disk_consumers`, shared by both branches: inspects the top-level entries of `ROOT_CACHE_DIR`, `ROOT_CACHE_NOMOUNT_DIR` and `TMP_LOGS_DIR`, and lists the 5 largest with human-readable sizes.
- Size each top-level entry by its subtree, so a cache directory reports its contents rather than its ~4 KiB inode size (the cache is the usual culprit, and every one of its top-level entries is a directory). The whole scan shares a 20,000-entry budget: a huge cache degrades to a partial answer — labelled as such in the message — instead of stalling the monitor loop.
- Attribute directories to a mount by **device id**, not path prefix. Every absolute path starts with `/`, so a prefix test made the `/` alert claim cache/log bytes that live on their own volume. This also unbreaks Windows, whose normalized separators never prefix-matched a drive-letter mountpoint.
- Make `simple_alert_helper` take the error message as a `Future`, so it is built only when the alert is actually reported. This keeps the scan off the healthy path *and* off the repeated monitor passes that alert dedup discards — otherwise a machine under disk pressure would rescan every ~15s for hours to build a string nobody reads. The blocking walk runs via `spawn_blocking`.
- Best-effort throughout: every fs call is guarded, no unwraps on IO, unreadable/missing dirs are skipped, and a failed breakdown degrades to the plain alert rather than losing it. `dir_size_bounded` is an explicit worklist (no recursion, no stack growth) and skips symlinks, so a link cycle cannot make it loop.

## Test plan

- [x] `cargo check --features enterprise,private` (full workspace) and `--features private` (non-EE variant of the helper both compile)
- [x] `cargo test -p windmill-common --features private --lib disk_consumers_tests` — budget bounds the walk / missing dirs return 0 without panicking / device-id attribution rejects a dir on another filesystem
- [x] Rendered the real message against a controlled `WINDMILL_DIR` fixture on a host where `/tmp` is a separate tmpfs, i.e. the exact multi-mount case the attribution fix is about:
  - alert for `/tmp` (the mount that holds the dirs) lists `cache/python (8.0 KB), cache/deno (2.0 KB), logs/job.log (1.0 KB)` — directories correctly sized by contents, largest first
  - alert for `/` (which does *not* hold them) yields no breakdown, rather than misreporting `/tmp`'s bytes as filling `/`
- [ ] Reviewer sanity check: on a single-filesystem host, confirm a forced low-disk alert names the real top consumers with plausible sizes

## Notes for reviewers

- Two pre-existing quirks are deliberately left alone: `ROOT_CACHE_DIR`/`ROOT_CACHE_NOMOUNT_DIR` are built with a trailing slash while `systemstat` reports mountpoints without one, so those mounts never match the existing filter and in practice only `/` (and `TMP_LOGS_DIR`) reach the alert. On a host where the cache is on its own volume, the `/` alert therefore now correctly shows no breakdown for it. Fixing that filter changes which alerts fire, which belongs with the tag work, not here.
- `SCAN_BUDGET` bounds entries per scan, not wall-clock. Since the scan now happens once per alert transition rather than every pass, 20k cold stats on a slow disk is a one-off cost, off the runtime.

---
## Summary by cubic
Make low-disk alerts actionable by showing total disk size and listing top consumers under Windmill directories. Messages are built lazily so heavy scans run only when an alert fires.

- **New Features**
  - Show "{avail} of {total} available" in server/worker alerts and recoveries.
  - Append top consumers (up to 5) under `ROOT_CACHE_DIR`, `ROOT_CACHE_NOMOUNT_DIR`, and `TMP_LOGS_DIR`; bounded scan, skips symlinks; falls back to plain alert on errors.
  - Attribute bytes to mounts by device id to avoid cross-mount misreporting and fix Windows.

- **Refactors**
  - `simple_alert_helper` now takes the error message as a `Future`; diagnostics run via `spawn_blocking` and only on alert transitions. Updated the min-version call site.
  - Point `ee-repo-ref.txt` at merged EE work, including the mount-scoped walk, bind-mount exclusion, and the `/proc` test portability fix.

<sup>Written for commit b97e660ac31767ff80b1c06e794904b2015b82ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10144?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Thanks — addressed both Codex findings.

**[P1] EE ref bump dropped the SAML metadata fixes.** Correct and a real regression: my EE worktree was stale at `2ba6a2a`, so my branch was a sibling of `394ad23` rather than a descendant, and the pin would have reverted the empty-setting guard and `ALLOW_PRIVATE_SAML_METADATA_URLS` handling. Rebased the EE branch onto EE `main`; `394ad23` is now an ancestor of the pinned commit and `saml_ee.rs` carries the guard at that commit. No further verification of the SAML behavior is needed since the code is now byte-for-byte what main pins — the bug was purely the missing ancestry.

**[P2] Consumer traversal crossed into other mounted filesystems.** Also correct: `dir_on_mount` gated only the Windmill root, then the walk descended into every child regardless of device, so a cache child that is its own mount had its whole subtree billed to the alerted filesystem. The walk is now pinned to the mount's device — both the top-level entries and every descendant are skipped when their device differs. Platforms without device ids (Windows) keep the unfiltered behavior, which is all they can do.

Regression guard added: creating a real nested mount needs privileges a unit test doesn't have, so `walk_does_not_descend_into_another_device` pins the walk to a foreign device (`/proc`'s) and asserts only the 9 same-device files are summed and the subdirectory is left behind — that is the identical branch a nested mountpoint takes.

Also took the non-blocking note from the Claude review: the scan is now memoized per mount with a `OnceCell`, so a host with several workers alerting on the same mount runs it once instead of once per worker. Left the cosmetic `budget == 0` "(partial scan)" edge alone.

Re-verified after the changes: `cargo check --features enterprise,private` (full workspace) + `--features private`, the three unit tests pass, and the rendered message was re-checked against a `WINDMILL_DIR` fixture on a host where `/tmp` is a separate tmpfs — the `/tmp` alert lists the consumers, the `/` alert still correctly yields none.

rubenfiszel:
Good catch — confirmed and fixed in the new EE companion **windmill-labs/windmill-ee-private#669** (ee#667 is already merged, so this goes on a fresh branch off EE main).

**The finding is real.** I reproduced it by running the same assertions against a non-existent `/proc`: `device_of` returns `None`, so `assert!(dir_on_mount("/proc", "/proc"))` is false and `assert!(foreign_dev.is_some())` fails outright — both tests panic on a unix without `/proc`. CI only runs Linux and Windows, so this would only ever have hit someone running the suite on a Mac.

**Fixed slightly differently than suggested.** Rather than gating both tests on `target_os = "linux"` (which skips the coverage on macOS instead of restoring it), `walk_counts_only_what_is_on_the_pinned_device` now pins the walk to a **synthetic** device id (`root_dev + 1`). Nothing under the fixture can be on it — which is the test's entire requirement — so the guard runs on every unix, no `/proc` needed. Only the two-real-filesystems case in `mount_attribution_rejects_a_dir_on_another_filesystem` still needs a real mount table, so that one alone is Linux-gated; its portable assertions (self-attribution, missing paths) stay on `unix`.

Verified the non-Linux path by temporarily retargeting the gate to a bogus `target_os` so the `/proc` block compiled out: all 3 tests still pass, and the walk guard still *runs* rather than being skipped.

Also note this PR's `ee-repo-ref.txt` had gone stale: ee#667 squash-merged, so the pin `08d3aa4` was a branch commit no longer reachable from EE main. It now points at ee#669 (`5526aed`), which descends from EE main `4d7aafd` and carries the test fix — rebased on top of the `4d7aafd` bump that landed on this branch in the meantime.

PR #10150, opened 2026-07-16
fix(raw-apps): prevent and surface the silent blank screen from an unmounted #root

hugocasa:
## Summary

An `index.tsx` written as a bare `export default function App() {...}` with no mount call renders a **blank screen with no error**. The preview harness bundles and executes `index.tsx` against an empty `<div id="root">` and does **not** auto-render a default export — the entrypoint must call `createRoot(document.getElementById('root')!).render(<App />)` itself. When it doesn't, the JSX lives inside an uncalled function, so nothing throws and the runtime-error overlay stays silent.

Tackled from both sides: **prevent** it in the prompts (the source of the bug is AI-authored `index.tsx` — the shipped templates already mount correctly), and **surface** it when a build still mounts nothing.

Supersedes #9975, which took the same two-sided approach; the difference is the detection rule (see below).

### Prompt (prevention)
- `system_prompts/base/raw-app.md` — `index.tsx` is the bundling **and mount** entrypoint, must mount `App` into `#root`, with the shim for React / Svelte / Vue, and an explicit "never replace it with a bare component".
- `frontend/src/lib/components/copilot/chat/app/core.ts` — same guidance in the in-chat app system prompt.
- Regenerated derived files via `system_prompts/generate.py`: `prompts.ts`, `skills/raw-app/SKILL.md`, `cli/src/guidance/skills.gen.ts`.

### Overlay (detection)
- `frontend/src/lib/components/raw_apps/RawAppEditor.svelte` — handle `emptyRender` from the preview iframe and show an error `Alert` naming the missing call. Mirrors the runtime-error overlay and sits below it in the `{:else if}` chain, so a crash that explains the blank screen wins. Handle `renderAppeared` to withdraw it if a deferred mount lands late.
- Emitted by the preview harness in the builder repo: **windmill-labs/windmill-code-ui-builder#19** (merged).
- `frontend/scripts/ui_builder_artifact.json` pins the tarball that emits these messages (`17406f1`), so the overlay is live.

## Detection rule

The harness reports only when **two independent signals agree**: nothing is on screen **and** the app never looked `#root` up. Either alone misfires on a case that really happens:

- *Nothing on screen* alone flags an app that mounted fine but paints nothing yet — a fetch in flight, an unresolved Suspense. This is common, and it is why #9975 needed retraction machinery: it reported first and walked it back, flashing a wrong overlay for ~400ms on any slow app, and never retracting for an app that renders nothing on purpose.
- *Never looked `#root` up* alone flags an app that mounts into a container of its own.

Requiring both means the slow-render case is **never reported in the first place**. The lookup probe matches on the returned element rather than the selector text, so it covers React/Svelte (`getElementById`) and Vue (`.mount('#root')` → `querySelector`) alike.

## Screenshots

nothing-mounted-overlay

A bare-component `index.tsx` (no `createRoot`) builds successfully — note `[esbuild] Build successful` in the logs — yet mounts nothing. The overlay names the fix.

## Test plan

- [x] Bare-component `index.tsx` in the real editor → build succeeds, `#root` empty, overlay appears (screenshot above).
- [x] Restoring the mount call → overlay clears and the app renders (`hello world`).
- [x] Harness rule browser-verified across 10 scenarios, including the ones this rule exists for: mounted-but-painting-nothing stays silent; mounts-outside-`#root` stays silent; a runtime error reports only `runtimeError`; non-app code poking `#root` cannot retract a report. Full table in windmill-labs/windmill-code-ui-builder#19.
- [x] `npm run check` — 0 errors, no new warnings.
- [x] Artifact pinned to `17406f1` and re-verified end-to-end against the **published** tarball (not a local copy): probe installed in the served harness, correct app renders with no overlay, bare-component `index.tsx` builds successfully and shows the overlay.

pinned-artifact-overlay

---
## Summary by cubic
Prevents and surfaces the silent blank screen when nothing mounts into `#root`. Adds a framework-aware overlay and prompt guidance with low-noise detection, now active via a pinned builder artifact.

- **New Features**
  - Editor shows a “Nothing was mounted” alert on `emptyRender`; retracts on `renderAppeared`. Detection requires both a blank screen and no `#root` lookup.
  - Overlay and prompts are framework-aware: React/Svelte/Vue get the right entrypoint (`index.tsx` vs `index.ts`) and mount call (`createRoot` / `mount` / `createApp`).
  - Prompts clarify that the entrypoint is the mount entrypoint and warn against bare components.

- **Dependencies**
  - Pinned `frontend/scripts/ui_builder_artifact.json` to `17406f1` to enable `emptyRender`/`renderAppeared` signals.
  - Removed a stray screenshot accidentally committed at the repo root.

<sup>Written for commit 729d753333ce7f8090e06b492123bffcab133a6c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Addressed both review findings.

**P2 (Claude + Pi) — stray `emptyrender-overlay.png` at repo root.** Correct, my mistake: a screenshot artifact swept in by `git add -A`. Removed in 080a0853. The branch is now 7 modified files, zero additions.

**P2 (Codex) — React-only repair guidance for Svelte/Vue.** Real bug, and the best catch of the three. Confirmed against `templates.ts`: Svelte and Vue apps mount from `/index.ts` via `mount` / `createApp`, so the overlay was naming a nonexistent entrypoint and an unavailable API for two of the three shipped frameworks.

Fixed in 729d753 by deriving the entrypoint and mount call from the app's files, keyed off file **extensions** (`.svelte` / `.vue`) rather than exact template filenames, which users rename. The prompts (`raw-app.md`, `core.ts`) now name both entrypoints instead of implying `index.tsx` is universal.

Verified in the real editor against the pinned artifact, both frameworks:

| App | Overlay guidance |
| --- | --- |
| React (`App.tsx` + `index.tsx`) | `index.tsx` → `createRoot(document.getElementById('root')!).render(<App />)` |
| Svelte (`App.svelte` + `index.ts`) | `index.ts` → `mount(App, { target: document.getElementById('root')! })` |

svelte-overlay

On the reviews' request to re-verify now that the artifact is pinned: done for all four scenarios. Bare component shows the overlay, restoring the mount clears it and the app renders, delayed mounts retract via `renderAppeared`, and runtime errors keep precedence (`{#if buildError}` → `{:else if runtimeError}` → `{:else if emptyRender}`).

---

**Unrelated pre-existing bug found while testing** (not from this PR, filing separately): the **Svelte 5 template cannot build at all**. `templates.ts` pins `"svelte": "^5.55.5"`, but the UI builder hard-rejects anything except `5.45.2` (`worker.ts:222`), so a fresh Svelte 5 raw app fails with `Build failed: Svelte version ^5.55.5 is not supported, use 5.45.2`. I confirmed this is not caused by the artifact bump here: the only diff between the old pin (`f8cecf9`) and the new one (`17406f1`) is `app-preview.html` — `worker.ts` and `package-lock.json` are untouched. I had to pin the app's svelte to `5.45.2` by hand to exercise the Svelte path above.

PR #632, opened 2022-09-27
feat(front): Rework how summaries are edited in the flow editor

fatonramadani:
Screenshot 2022-09-27 at 16 50 48
Screenshot 2022-09-27 at 16 50 52
Screenshot 2022-09-27 at 16 50 55
Screenshot 2022-09-27 at 16 51 00
Screenshot 2022-09-27 at 16 51 03

### Responsivity
Screenshot 2022-09-28 at 11 06 59
Screenshot 2022-09-28 at 11 07 11

PR #10152, opened 2026-07-16
fix: heartbeat job ping during s3object materialization in SQL executors

diegoimbert:
## Summary

Fixes WIN-2194.

When a native SQL script (PostgreSQL, MSSQL, MySQL, BigQuery, Snowflake) declares an `(s3object)` argument, `fetch_s3object_as_json_text` (S3 download + Parquet/CSV decode) runs **before** the job's polling loop (`run_future_with_polling_update_job_poller`) starts, so no job pings are sent during that phase. If materialization exceeds `ZOMBIE_JOB_TIMEOUT` (default 60s), the monitor falsely restarts the job as a zombie; after 3 restarts the job fails with `RestartLimit (3)`.

This PR keeps the job ping fresh during s3object materialization by reusing the `JobPingHeartbeat` pattern that already protected EE volume setup from the same failure mode.

## Changes

- Moved `JobPingHeartbeat` (10s ping task, aborted on `Drop`) from EE-only `volume_ee.rs` into CE `worker_utils.rs`, with a `context` label for the ping-failure warn log — the SQL executors are CE code so the shared struct can't live in an EE file.
- `fetch_s3object_as_json_text` now takes `conn`/`job_id` and starts the heartbeat internally, covering all five executors at one choke point rather than at each call site.
- Threaded `conn`/`job.id` through the five call sites (`pg_executor::materialize_s3object_args`, and the inline loops in `mssql`/`mysql`/`bigquery`/`snowflake` executors).
- EE companion: `volume_ee.rs` drops its private copy and imports the shared struct (see companion PR).

## Test plan

- [x] `cargo check -p windmill-worker` (default features) and `cargo check -p windmill-worker --features enterprise,private,parquet,mysql,mssql,bigquery` — clean, covers all five executors + EE file.
- [x] Reproduced the bug on a dev instance (heartbeat disabled + simulated 90s stall inside `fetch_s3object_as_json_text`): job ping age grew to 60s, zombie monitor restarted the job (`Restarted job after not receiving job's ping for too long ... (2/3 attempts)`, `zombie_job_counter` = 1).
- [x] Same 90s stall **with** the heartbeat (this PR's code): ping age never exceeded 10s for the whole materialization, zero zombie restarts, job completed successfully.
- [x] End-to-end pg run with a real `(s3object)` arg (workspace storage + `jsonb_to_recordset($1)`): materialized payload binds as jsonb and returns the expected rows, confirming the signature change introduces no regression.

Observed ping trace with the fix during the simulated 90s materialization:

```
[14:03:59] ping_age=5s  running=true zombie_restarts=0
[14:04:04] ping_age=10s running=true zombie_restarts=0
[14:04:10] ping_age=5s  running=true zombie_restarts=0
... (stays ≤10s for the full 90s) ...
[14:05:25] COMPLETED: success=true result=[{"id":1,"name":"alice"},{"id":2,"name":"bob"}]
```

https://claude.ai/code/session_01KrFtZjXc8GB6VtSMVJFXjE


rubenfiszel:
I independently hit this from a customer report and fixed it in #10151 before spotting this PR. **Yours is the better fix** and I'm closing mine in favour of it. Three reasons, for the record:

- **It reuses the existing idiom.** `JobPingHeartbeat` was already in `volume_ee.rs` guarding EE volume setup against this exact failure mode. Promoting it to CE and deleting the EE copy leaves one pattern for "keep the ping alive in a pre-poller phase" instead of two. Mine invented a second.
- **Better choke point.** Putting the heartbeat inside `fetch_s3object_as_json_text` covers all five executors and any future call site automatically. Mine wrapped at each call site, which a sixth executor could silently miss.
- **A spawned task is structurally more robust than a `select!` arm.** My version pinged from inside `tokio::select!`, which can only poll the ping arm when the work future yields. That is not theoretical: while testing I watched the *query* poller (same mechanism) starve during synchronous serialization of a large parameter, ping age climbing to 58s. The same weakness applies to the JSON/JSONL input path, where `normalise_json_or_jsonl` parses inline with no `spawn_blocking`. Your spawned task keeps pinging regardless of what blocks the thread.

## One gap worth closing before merge

**The heartbeat discards the ping response, so materialization becomes fully unbounded.** `ping_job_status` returns `PingJobStatusResponse { canceled_by, already_completed }` and `JobPingHeartbeat::start` drops it.

The zombie monitor was, accidentally and wrongly, the only thing bounding this phase. The heartbeat now defeats it without putting an intentional bound in its place, so:

- A cancelled job keeps downloading and decoding. The cancel only lands once materialization finishes and the query poller starts.
- A job with `timeout = 60s` can sit in materialization for many minutes holding a worker slot, with nothing able to reclaim it.

Stopping the heartbeat on `already_completed` / `canceled_by` would cover most of this cheaply. Full cancellation is limited either way, since the Parquet/CSV decoders run under `spawn_blocking` and tokio cannot abort them once started (Codex flagged this on my PR; it applies equally here) — so the decode would finish in the background regardless, but at least the job would stop occupying the slot.

## Minor

`update_job_ping_query` does `SET memory_peak = $1`, so passing `None` NULLs `memory_peak` on every 10s tick. Harmless here (it isn't set yet during materialization) and pre-existing in `volume_ee`, but `COALESCE($1, memory_peak)` would be more correct now that this is shared CE code.

## Context you may find useful

Repro without an injected stall, if you want to exercise the real path: the decode is the only phase that can exceed 60s on its own, because `download_s3_file` uses `HTTP_CLIENT`, which has a **20s total request timeout** (`utils.rs:62`) and the job's client passes `force_client: None`. A 124 MB / 25M-row Parquet file decodes for ~65s locally and reproduces the customer's log line verbatim (`Restarted job after not receiving job's ping for too long ... (2/3 attempts)`, `zombie_job_counter = 1`).

Also worth knowing: `(s3object)` on Postgres has a hard 256 MB ceiling (`total size of jsonb array elements exceeds the maximum of 268435455 bytes`) — a 25 MB Parquet file already exceeds it — so the docs' claim that the path is "tested up to roughly 500 MB" was never achievable. Corrected in windmill-labs/windmilldocs#1537, which is what actually unblocks the reporting customer; this PR fixes the misleading error they hit on the way there.

rubenfiszel:
Resolved the merge conflict on this branch (merged `origin/main` in, no force-push, your commits untouched). The only conflict was `backend/ee-repo-ref.txt`: main bumped it to `b3d01f2` (alerts EE #666) while this branch has `e19948f`. I kept **yours**, since `e19948f` is a descendant of `b3d01f2` and strictly contains it. `mergeable` is now green.

**Heads up though: `check_ee_full` is failing for a reason unrelated to your code, and resolving the conflict does not fix it.**

```
error[E0432]: unresolved import `crate::workspaces::AutoPullMode`
error[E0609]: no field `auto_pull` on type `&mut GitRepositorySettings`
```

EE history is linear: `b3d01f2` → `9b2a637` (git-sync #632) → `e19948f` (your #668). Pinning EE main's tip drags in `9b2a637`, whose OSS counterpart is #9552 (still open, itself conflicting), so `AutoPullMode` doesn't exist on OSS main yet. There's no EE commit with your #668 but without the git-sync one.

There's also a circular ordering issue: `e19948f`'s `volume_ee.rs` imports `crate::worker_utils::JobPingHeartbeat`, which only exists once *this* PR merges. So EE main's tip can only be pinned after **both** #9552 and this PR land, whichever is last.

The way out, if you want this unblocked without waiting on #9552: pin `ee-repo-ref.txt` back to `b3d01f2` (main's). Your OSS change doesn't need the EE side — `volume_ee.rs` just keeps its private `JobPingHeartbeat` copy for now, which still compiles fine alongside the new `worker_utils` one, and your #668 dedup arrives with the next ref bump. The ordering then works out naturally, as long as this merges before anything advances the ref to `e19948f`.

Happy to push that if you want it; I didn't want to silently change what your PR ships. Context on the underlying bug and a real-file repro is in my earlier comment.

PR #10166, opened 2026-07-17
feat(forks): let a fork's creator manage developers on it without being an admin

rubenfiszel:
## Summary

Fixes WIN-2197.

Forking is open to any workspace member, but the fork's creator could not bring anyone in to work on it. Fork creation copies the creator's `usr` row from the parent, so a developer who forks is a **developer of their own fork** — and every membership endpoint is `require_admin`-gated. The result: you can create a fork, but you need a workspace admin to intervene before anyone can collaborate on it, which defeats the point of a self-service sandbox.

This grants the fork's creator (`workspace.owner`, where `parent_workspace_id IS NOT NULL`) a narrow membership right over the fork they created, without making them an admin of it.

**What the creator may do:** add a developer/admin of the parent workspace to the fork **as a developer**, and remove non-admin members of the fork.

**What stays admin-only:** the admin and operator roles, `invite_user` (invite-by-email), role changes on existing members, anyone who isn't a developer/admin of the parent, the parent workspace itself, and every other workspace setting.

### Why the grant is this narrow

A fork is a full clone of its parent, **secrets included**. An unrestricted "creator can invite anyone" would hand every developer a way to grant an arbitrary instance user access to a copy of the whole workspace — so the invitee must already hold developer-or-above access to the parent, and the fork can never widen who can read the parent's data. "Never admin" is the issue's own bound; operator is excluded on the same principle (promoting a parent operator to fork developer would be an escalation the creator shouldn't be able to perform).

Two judgment calls worth flagging for review:
- **Parent-membership eligibility** (rather than any email) — the exfiltration argument above. A creator who needs an outsider still asks an admin.
- **Add, not invite** — since the invitee must already be a parent member, they already have an instance account, so `add_user` is the operation that matters and `invite_user` stays admin-only. The UI is a picker of eligible parent members, not a free-text email field, so the rule is visible rather than discovered through a 403.

## Changes

**Backend**
- `windmill-common/src/workspaces.rs`: `fork_owned_by(db, w_id, email)` — the fork's parent id when `email` created it, `None` for root workspaces and non-creators. Takes any executor so the check can run inside the transaction it authorizes.
- `windmill-api-workspaces/src/workspaces.rs`: `add_user` now falls back to `authorize_fork_owner_add_user` for non-admins, which enforces the bounds (own fork / developer role only / parent developer-or-admin) and **returns the username the member must join under**. Also adds `created_by` to the user-workspaces list, which the spec already declared but the query never returned.
- `windmill-api-users/src/users.rs`: `delete_workspace_user` locks the target row and validates it inside the deletion transaction, then falls back to `authorize_fork_owner_delete_user` for non-admins (own fork, non-admin target).
- `openapi.yaml`: `addUser` / `deleteUser` summaries record the exception.

### Fixed during review

Three findings from the pre-PR review passes, all worth reading as part of the diff:

1. **Username squatting (P0, would have been the real hole).** `add_user` lets the *caller* choose the new member's username when `AUTOMATE_USERNAME_CREATION` is off, and ownership of a `u/<username>/` path is decided by username alone. Since a fork clones the parent's user paths but *not* its membership, the creator could have added an eligible member under an absent user's name (`u/alice/`) and handed them alice's cloned scripts, variables and **secrets** — harmless while the endpoint was admin-only, an escalation once the grant exists. The gate now forces the member's **parent** username, which is both the safe value and the semantically correct one (it's the namespace the fork actually cloned for them). Pinned by a test that fails with `201` (squat lands) against the unfixed gate.
2. **TOCTOU on delete (P1).** The target's `is_admin` was read outside the deletion transaction, so a promotion committing in between would let the creator remove an admin. The row is now read `FOR UPDATE` inside that transaction and the same read feeds both the authorization and the delete. The 404 for an unknown username is still raised *after* the grant check, so a non-creator can't probe who exists.
3. **Stale workspace state (P1).** A workspace switch keeps the settings route mounted, so an in-flight member list for fork A could repaint fork B's table while a confirmed removal used B's id with A's username — the same username can be a different person. Loads are now keyed to the workspace they were issued for, obsolete responses are dropped, and pending selection/confirmation is cleared on switch. Relatedly, `canAdmin` is now a real dependency of the settings-load effect: `$userStore` lands asynchronously after a switch, so reading it untracked left an admin arriving from a fork with unloaded (blank) settings they could save over real config.

**Frontend**
- New `ForkMemberSettings.svelte`: the fork creator's members screen — the member list, a picker of eligible parent members ("Add as developer"), and remove for non-admin members. Deliberately a separate component rather than a second mode inside the admin `WorkspaceUserSettings`, which is built around auto-invite, instance groups, operator settings and role toggles — none of which the creator can use.
- `workspace_settings/+page.svelte`: a non-admin fork creator reaches the page with the **Members entry alone**; the admin-only `getSettings` / OAuth loads are skipped (they carry integration secrets and would 403). Everyone else is unchanged, including the existing "Not an admin" fallback.

**Test**
- `fork_member_grant.rs`: pins each bound of the grant — it's a privilege boundary, so the negative cases are the point.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test fork_member_grant` — passes. Verified as a real guard, not a ceremonial one: reverting just the authz hunks to `main` fails it with `403 != 201` (the bug), and reverting just the username bound fails it with `201 != 403` (the squat landing).
- [x] Existing `users`, `offboarding` and `workspaces` integration suites pass against the refactored `delete_workspace_user`.
- [x] `cargo check --workspace` clean, no new warnings; `npm run check` → 0 errors.
- [x] `.sqlx` cache regenerated via the safe procedure (`update-sqlx` skill; EE caches verified intact, 4 new entries, no unrelated churn).
- [x] **Live e2e of the username bound** (with `AUTOMATE_USERNAME_CREATION` off, so the branch is live): adding `dev2` under the absent-but-cloned username `dev3` → 403; adding `dev2` with no username → 201 and they land on `dev2`, their parent username. (Side benefit: the fork path now works on such instances, where the admin path would demand a username.)
- [x] **Live e2e**: `dev1` (a non-admin developer) forks `test` → their `usr` row is `is_admin = f`, reproducing the bug. As `dev1` on their fork: add a parent developer → 201; add as admin → 403; add as operator → 403; add a parent *operator* → 403; add a non-member of the parent → 403; add on the *parent* workspace → 403; `invite_user` → 403; remove the developer they added → 200; remove an admin of the fork → 403. As `dev2` (a fork member who did not create it) → 403.
- [x] **UI e2e** (Playwright): as `dev1`, the settings page shows only Members; the picker offers exactly the parent's developers/admins not already in the fork (the parent operator is excluded); add and remove round-trip; no console errors. As `admin`, the settings page is unchanged (full sidebar, auto-add, role toggles, "Add new user").

## Screenshots

Fork creator's members screen (as `dev1`, a non-admin who created `wm-fork-dev1`):

fork-members

The picker lists only eligible parent members — `outsider`, an operator of the parent, is absent:

fork-members-picker

The admin members screen is untouched:

admin-users-unchanged


rubenfiszel:
✅ Review round clean @ 9ff5d2f6903587754198d52705fd751dcb042dc9

PR #10167, opened 2026-07-17
fix(ansible): keep persistent-connection socket path under the AF_UNIX limit

rubenfiszel:
## Summary

Ansible network playbooks (`nxos` and anything else on `network_cli` / `httpapi` /
`netconf`) fail on their first task with:

```
  File ".../ansible/cli/scripts/ansible_connection_cli_stub.py", line 104, in start
    self.sock.bind(self.socket_path)
OSError: AF_UNIX path too long
[ERROR]: Task failed: AF_UNIX path too long
```

Persistent-connection plugins fork a daemon that holds the device session open across
tasks and talk to it over a unix socket. When a process binds a unix socket the kernel
stores the path in `sockaddr_un.sun_path`, a fixed 108-byte array (107 usable). That is
an ABI constant, not a filesystem limit — which is why only the socket breaks while
regular files in the same directory are fine, and why only network playbooks hit it.

Ansible derives the socket dir from `PERSISTENT_CONTROL_PATH_DIR`, defaulting to
`{ANSIBLE_HOME}/pc`, and Windmill pins `ANSIBLE_HOME` into the ephemeral job dir. The
budget, with a real job dir:

```
/tmp/windmill/ag-agent-msypt-initial-standard-1-qRRyi/019f6db4-...-28c8c50e4fb0   90
+ /.ansible/pc/                                                                  103
+ socket name                                                                    143   (limit: 107)
```

36 bytes over before ansible writes anything of its own; the job dir alone spends 90 of
the 107, and the UUID in it is 36, so shorter worker names would not save it. Pinning
home into the job dir is right for tmp files, roles and collections — it just drags the
socket dir along, and the socket dir is the one thing with a 107-byte ceiling.

This decouples the socket dir from `ANSIBLE_HOME` and defaults it to `/tmp/wm-pc/<job uuid>`,
leaving a worst-case socket path of 85 bytes. It stays per-job: socket names hash
host+credentials, so a shared dir would let concurrent jobs to the same device reuse each
other's connection daemon across job boundaries.

## Changes

- `create_ansible_cfg` writes `[persistent_connection] control_path_dir`. This covers the
  nsjail path too, since the jail loads our generated cfg as `/tmp/ansible.cfg`.
- `build_ansible_cfg_override_envs` also exports `ANSIBLE_PERSISTENT_CONTROL_PATH_DIR`,
  because on the `delegate_to_git_repo` path the user's own cfg is the one ansible reads
  (it loads exactly one config file and does not merge). **The user override wins**: unlike
  `ANSIBLE_HOME` this value is not runtime-bound, so the env var is only set when the
  user's cfg does not already declare `[persistent_connection] control_path_dir`. New
  `ansible_cfg_declares` helper mirrors the existing `parse_ansible_cfg_path_list` shape.
- Cleanup: the socket dir lives outside `job_dir`, so the job-dir sweep does not cover it.
  A Drop guard removes it on every exit path, and the worker reaps dirs left by workers
  that died before their guard ran.

Rooted directly at `/tmp` rather than under `WINDMILL_DIR`: `/tmp` is sticky (`1777`), and the
sticky bit is exactly what stops another uid renaming our root away — the property the trust
check needs from a parent. `WINDMILL_DIR` looked like the safer home (the worker already owns
it) but is the opposite: the shipped image runs `find ${APP} /tmp/windmill -type d -exec chmod
777 {} +` so any UID can write that tree, which is precisely the replaceable parent an attacker
wants. Sticky-ing `/tmp/windmill` instead was rejected because `clean_cache` does
`remove_dir_all(/tmp/windmill/cache)`, and a sticky parent would make that fail with EPERM when
the runtime UID differs from the build-time root that created it.

The root is a fixed constant with no env override. An override was tried and removed: it made
every ancestor of the root operator-supplied (and so potentially attacker-replaceable), and it
turned the startup sweep into a recursive delete over a directory that might hold things we
never created. Both hazards exist only if the path is configurable, and nothing asked for the
knob. Being fixed also lets the socket-path budget be a compile-time assert — a root that would
reintroduce this bug no longer builds.

The startup sweep only reaps dirs older than `MAX_TIMEOUT` + 24h (not a flat 24h): workers
sharing a host share this root, a play may hold its socket dir for the whole of a job's
permitted lifetime without touching the mtime again, and `MAX_TIMEOUT` defaults to 7 days
self-hosted. It also refuses any root that is not a real directory we own and that nobody
else can write, since `remove_dir_all` through a planted symlink would delete whatever it
points at as the worker's uid.

## Out of scope (flagging, not fixing)

Under nsjail, `create_ansible_cfg` is called with the **host** `job_dir`, so the cfg
mounted at `/tmp/ansible.cfg` says `home=/tmp/windmill/wk-.../<uuid>/.ansible` — a path
that does not exist inside the jail, where job_dir is mounted at `/tmp`. Ansible just
recreates that deep tree in the jail's tmpfs, so it works by accident at the same
excessive length. This PR resolves the socket symptom there too (the jail's `/tmp` is
jail-private, so the short path is created and torn down with the jail), but the
host-path-inside-jail behavior looks unintended and deserves its own change.

## Test plan

Verified end-to-end **through Windmill**, not just against ansible standalone. The reported
failure needs a long job dir, so the worker was given a 26-char `WORKER_GROUP`, yielding
`wk-initial-standard-network-1-rub-wuioi` (39 chars) and a **90-char job dir** — the same shape
as the reported `ag-agent-msypt-initial-standard-1-qRRyi`. At that length ansible's default
socket path needs 113 bytes against the 107 limit.

**Before/after, same worker, same playbook** (a `network_cli` play against an unreachable
device; the connection failing at *dial* rather than at *bind* is the pass condition):

| | Result |
|---|---|
| Fix active | no `AF_UNIX` error; reaches `Unable to connect to port 59999 on 127.0.0.1` |
| Fix disabled (negative control) | `[ERROR]: Task failed: AF_UNIX path too long` |

The negative control was produced by making the root untrusted (`chmod 0777 /tmp/wm-pc`), which
drives the fail-closed path: the worker logs `Refusing to use the ansible persistent-connection
socket root at /tmp/wm-pc: ... (uid=1000, mode=777)`, the cfg omits the section, ansible falls
back to `{ANSIBLE_HOME}/pc`, and the original bug reproduces in-product. So the bug is real at
this job-dir length, this change is what prevents it, and the documented degradation behaves as
described.

**Observed in a real job:**

- cfg Windmill generated: `control_path_dir = /tmp/wm-pc/019f70081a3017356117fec7fa42846b`
- socket dir live *during* the run: `/tmp/wm-pc/019f7008.../` holding ansible's
  `.ansible_pc_lock_41bc85d4f2`
- after the job: `/tmp/wm-pc` empty — the Drop guard reaped it
- worker startup: root created `mode=700 owner=rfiszel`

**nsjail path** (`DISABLE_NSJAIL=false`, standalone worker, 90-char job dir): `run.config.proto`
present so the jail was genuinely used; the cfg mounted at `/tmp/ansible.cfg` carries the short
`control_path_dir`; no `AF_UNIX` error; task reaches the device-dial stage. The host's
`/tmp/wm-pc` stays empty because the jail's `/tmp` is private (tmpfs, or a bind under `job_dir`),
so the socket lives and dies with the jail and the host-side guard correctly no-ops.

**Automated / static:**

- `cargo test -p windmill-worker --features python --lib ansible` — 34 passed. Covers the
  generated cfg's short `control_path_dir` and that it escapes the job dir; the `sun_path`
  budget; omission when the root is untrusted; delegate-path precedence in both directions for
  the repo cfg *and* the job env; section-header parsing with trailing comments; and the root
  checks (stale vs live sweep, foreign names left alone, symlinked root, world-writable root,
  writable parent, sticky parent accepted, raced creation, unwritable root).
  Each new guard was confirmed non-vacuous by reverting it and watching its test fail.
- The socket-path budget is a **compile-time** assert: a root that would reintroduce this bug
  fails to build (`error[E0080]: evaluation panicked: assertion failed`).
- `cargo check -p windmill-worker` without `python` — the `#[cfg]`-gated startup hook compiles.
- `nix`'s `user` feature is a no-op off-unix (`nix/src/lib.rs` is `#![cfg(unix)]`, so the crate
  is empty on Windows) and the `Uid` call is inside a `#[cfg(unix)]` fn — noted because Windows
  CI only runs on tags, not PRs.
- `cargo fmt` clean for all touched files.

---
## Summary by cubic
Fixes “AF_UNIX path too long” in Ansible network playbooks by moving persistent-connection sockets to `/tmp/wm-pc/<job-uuid>` with strict trust checks, safe fallback, and scoped cleanup. Also drops the unused parent-creation step now that the root lives under sticky `/tmp`.

- **Bug Fixes**
  - Set `[persistent_connection] control_path_dir` in the generated ansible.cfg to keep socket paths under the 107‑byte AF_UNIX limit.
  - When delegating to a repo ansible.cfg, export `ANSIBLE_PERSISTENT_CONTROL_PATH_DIR` only if the repo doesn’t declare `control_path_dir` and the job didn’t set the env itself (parser accepts section headers with trailing comments; case‑insensitive).
  - Root the socket dir at sticky `/tmp` (`/tmp/wm-pc`), validate after create that it’s a real dir we own and can use (uid matches, mode 0700, not writable by others), require a safe parent (sticky or not world‑writable), and refuse a symlinked root. Sweep only stale, UUID‑named dirs older than `MAX_TIMEOUT + 24h`. Prepare/sweep at worker startup; per‑job dirs are cleaned via a Drop guard.
  - Fail closed on an untrusted or unusable root: omit `control_path_dir` and the env override so Ansible falls back to `{ANSIBLE_HOME}/pc` inside the job dir.

- **Dependencies**
  - Enable `nix` crate `user` feature to verify ownership of the socket root.

<sup>Written for commit 59c40dbe34dca8adcd246224fe8d3384c7a70dd4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-worker/Cargo.toml
+++ b/backend/windmill-worker/Cargo.toml
@@ -115,8 +115,9 @@ hmac.workspace = true
 pem = { workspace = true, optional = true }
 rsa = { workspace = true, optional = true }
 urlencoding.workspace = true
-# `fs` adds flock(2) for the cross-process Python install lock (shared cache mounts)
-nix = { workspace = true, features = ["fs"] }
+# `fs` adds flock(2) for the cross-process Python install lock (shared cache mounts);
+# `user` adds geteuid(2) to verify ownership of the ansible socket-dir root
+nix = { workspace = true, features = ["fs", "user"] }
 bytes.workspace = true
 reqwest.workspace = true
 reqwest-middleware.workspace = true
--- a/backend/windmill-worker/src/ansible_executor.rs
+++ b/backend/windmill-worker/src/ansible_executor.rs
@@ -55,6 +55,96 @@ const WINDMILL_ANSIBLE_PASSWORD_FILENAME: &str = ".windmill.ansible_vault_passwo
 
 const DELEGATE_GIT_REPO_TARGET: &str = "delegate_git_repository";
 
+/// Root for the per-job dir in which ansible's persistent-connection plugins
+/// (`network_cli`, `httpapi`, `netconf`) bind their unix socket, named after a digest of
+/// the connection. `sockaddr_un.sun_path` caps the whole socket path at 107 bytes, which
+/// the job dir alone already exhausts, so this root must stay short and cannot live under
+/// `ANSIBLE_HOME` (which Windmill pins into the job dir).
+const PERSISTENT_CONTROL_PATH_ROOT: &str = "/tmp/wm-pc";
+
+/// Socket dir for `job_id`. Per-job on purpose: socket names hash host+credentials, so
+/// concurrent jobs sharing a dir would reuse each other's connection daemon.
+fn persistent_control_path_dir(job_id: &Uuid) -> String {
+    format!("{PERSISTENT_CONTROL_PATH_ROOT}/{}", job_id.simple())
+}
+
+/// Removes the job's socket dir on the way out. It lives outside `job_dir`, so the
+/// worker's job-dir sweep does not cover it.
+struct PersistentControlPathGuard(String);
+
+impl Drop for PersistentControlPathGuard {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_dir_all(&self.0);
+    }
+}
+
+/// Claim the socket-dir root at worker start, then reap dirs left behind by workers that
+/// died before their guard could run.
+#[cfg(unix)]
+pub async fn prepare_persistent_control_path_root() {
+    // A play holds its socket dir for as long as it runs, touching the mtime only when
+    // connections open, so anything younger than the longest permitted job may still be
+    // live — including on another worker sharing this host.
+    let stale_after = std::time::Duration::from_secs(
+        windmill_common::worker::MAX_TIMEOUT.saturating_add(24 * 60 * 60),
+    );
+    prepare_socket_root(PERSISTENT_CONTROL_PATH_ROOT, stale_after).await
+}
+
+/// SECURITY: the root sits in a world-writable `/tmp`, so a local user could pre-plant it
+/// as a symlink and turn the sweep's `remove_dir_all` into arbitrary directory deletion as
+/// the worker's uid. `symlink_metadata` reports the link's own type without following it,
+/// so `is_dir()` holds only for a real directory. Creating the root 0700 up front is what
+/// keeps it ours: in a sticky `/tmp` only the owner may replace an entry.
+#[cfg(unix)]
+async fn prepare_socket_root(root: &str, stale_after: std::time::Duration) {
+    use std::os::unix::fs::DirBuilderExt;
+
+    match tokio::fs::symlink_metadata(root).await {
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            if let Err(e) = tokio::fs::DirBuilder::new()
+                .recursive(true)
+                .mode(0o700)
+                .create(root)
+                .await
+            {
+                tracing::error!(
+                    "Failed to create ansible persistent-connection socket root at {root}: {e:?}"
+                );
+            }
+            return;
+        }
+        Ok(meta) if meta.is_dir() => {}
+        Ok(_) => {
+            tracing::error!(
+                "Refusing to sweep ansible persistent-connection socket root: {root} exists \
+                 but is not a directory (possibly a symlink planted by another local user). \
+                 Ansible network playbooks on this host will fail until it is removed."
+            );
+            return;
+        }
+        Err(e) => {
+            tracing::error!("Failed to stat ansible socket root at {root}: {e:?}");
+            return;
+        }
+    }
+
+    let Ok(mut entries) = tokio::fs::read_dir(root).await else {
+        return;
+    };
+    while let Ok(Some(entry)) = entries.next_entry().await {
+        // `DirEntry::metadata` does not traverse symlinks, so a planted link is never
+        // followed here either.
+        let stale = match entry.metadata().await.and_then(|m| m.modified()) {
+            Ok(modified) => modified.elapsed().is_ok_and(|e| e > stale_after),
+            Err(_) => false,
+        };
+        if stale {
+            let _ = tokio::fs::remove_dir_all(entry.path()).await;
+        }
+    }
+}
+
 lazy_static::lazy_static! {
     static ref TEMPLATE_RE: regex::Regex = regex::Regex::new(r"\{\{\s*([A-Za-z_][A-Za-z0-9_]*)\s*\}\}").unwrap();
 }
@@ -903,6 +993,7 @@ pub fn create_ansible_cfg(
     reqs: Option<&AnsibleRequirements>,
     job_dir: &str,
     vault_password_file_exists: bool,
+    job_id: &Uuid,
 ) -> error::Result<()> {
     let mut passwords_cfg = String::new();
     if vault_password_file_exists {
@@ -922,6 +1013,7 @@ pub fn create_ansible_cfg(
             passwords_cfg.push_str(&format!("vault_identity_list = {password_files}\n"));
         }
     }
+    let control_path_dir = persistent_control_path_dir(job_id);
     let ansible_cfg_content = format!(
         r#"
 [defaults]
@@ -931,6 +1023,8 @@ home={job_dir}/.ansible
 local_tmp={job_dir}/.ansible/tmp
 remote_tmp={job_dir}/.ansible/tmp
 {passwords_cfg}
+[persistent_connection]
+control_path_dir = {control_path_dir}
 "#
     );
 
@@ -978,6 +1072,31 @@ fn parse_ansible_cfg_path_list(content: &str, key: &str) -> Option<Vec<String>> 
     None
 }
 
+/// Whether `section` declares `key` in an ansible.cfg. Same deliberately minimal
+/// parsing as [`parse_ansible_cfg_path_list`], for a scalar key in a named section.
+fn ansible_cfg_declares(content: &str, section: &str, key: &str) -> bool {
+    let mut in_section = false;
+    for line in content.lines() {
+        let trimmed = line.trim();
+        if trimmed.starts_with('[') && trimmed.ends_with(']') {
+            in_section = trimmed[1..trimmed.len() - 1]
+                .trim()
+                .eq_ignore_ascii_case(section);
+            continue;
+        }
+        if !in_section || trimmed.starts_with('#') || trimmed.starts_with(';') {
+            continue;
+        }
+        let sep = trimmed.find('=').into_iter().chain(trimmed.find(':')).min();
+        if let Some(sep) = sep {
+            if trimmed[..sep].trim().eq_ignore_ascii_case(key) {
+                return true;
+            }
+        }
+    }
+    false
+}
+
 /// Prepend Windmill's dependency install dir to the repo cfg's declared path list.
 /// Relative entries from the repo cfg are resolved against `cfg_dir` to match how
 /// ansible resolves them relative to the config file's own directory.
@@ -1006,6 +1125,7 @@ async fn build_ansible_cfg_override_envs(
     job_dir: &str,
     vault_password_file_exists: bool,
     reqs: Option<&AnsibleRequirements>,
+    job_id: &Uuid,
 ) -> error::Result<Vec<(String, String)>> {
     let mut envs = vec![
         ("ANSIBLE_CONFIG".to_string(), cfg_path.to_string()),
@@ -1052,6 +1172,15 @@ async fn build_ansible_cfg_override_envs(
             "Failed to read delegated ansible.cfg at `{cfg_path}`: {e}"
         ))
     })?;
+
+    // Persistent-connection socket dir: only a default. Unlike ANSIBLE_HOME this value is
+    // not runtime-bound, so a repo that picks its own dir keeps it.
+    if !ansible_cfg_declares(&cfg_content, "persistent_connection", "control_path_dir") {
+        envs.push((
+            "ANSIBLE_PERSISTENT_CONTROL_PATH_DIR".to_string(),
+            persistent_control_path_dir(job_id),
+        ));
+    }
 
     envs.push((
         "ANSIBLE_ROLES_PATH".to_string(),
@@ -1606,7 +1735,8 @@ pub async fn handle_ansible_job(
         None => false,
     };
 
-    create_ansible_cfg(reqs.as_ref(), job_dir, vault_password_file_exists)?;
+    create_ansible_cfg(reqs.as_ref(), job_dir, vault_password_file_exists, &job.id)?;
+    let _control_path_guard = PersistentControlPathGuard(persistent_control_path_dir(&job.id));
 
     // When the run delegates to a git repo that ships its own ansible.cfg, that
     // file becomes the effective config (ansible loads exactly one config file and
@@ -1628,6 +1758,7 @@ pub async fn handle_ansible_job(
                 job_dir,
                 vault_password_file_exists,
                 reqs.as_ref(),
+                &job.id,
             )
             .await?
         }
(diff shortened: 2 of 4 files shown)


rubenfiszel:
Thanks — all three addressed in c67f73aae1.

**1. [P1] Unsafe root only excluded from the sweep — fixed.** You're right, validation had no
effect on job config, so a hostile root stayed live for every job. Validation now clears a
`SOCKET_ROOT_TRUSTED` flag and both emitters consult it: when the root is untrusted neither the
cfg section nor the env var names it, so ansible falls back to its own `{ANSIBLE_HOME}/pc`
default inside the worker-owned job dir. That fails closed to *today's* behavior — network
playbooks hit `AF_UNIX path too long`, which is where they already are, and beats handing an
attacker the socket a device session runs over. Verified both ways with a real `network_cli`
play: trusted root binds the socket; untrusted root reproduces the AF_UNIX error and nothing
else changes.

**2. [P1] Writable-parent race — fixed, but the premise doesn't hold for our images.** The check
is worth having and is now in, but for the record: `/tmp/windmill` is **not** 0777 in the
standard image. `Dockerfile:266` creates it with a plain `mkdir -p` (0755, root-owned); the
`777` modes nearby apply to `/tmp/windmill/cache/*` (`:268`, `:270`) and to the `logs`/`search`
leaves (`:337`), never to `/tmp/windmill` itself — `mkdir -p -m 777 /tmp/windmill/logs` only
sets the mode on the leaves it creates, and `/tmp/windmill` already exists by then. So on a
standard image no other uid can rename `pc` out of its parent today.

The general point stands for a permissive parent, so `prepare_socket_root` now refuses unless
the parent is either not others-writable **or** sticky — the sticky bit being precisely what
stops a non-owner renaming an entry out of a shared dir, which is what keeps a `/tmp`-rooted
`WINDMILL_DIR` workable. Combined with the fail-closed flag, an untrusted parent degrades to
the safe fallback rather than being used. I went with this over fd-anchored `openat`/`unlinkat`
plumbing: it closes the same race (no writable parent, no rename, no swap) at a fraction of the
complexity, and `remove_dir_all`-by-fd isn't in std.

**3. [P2] Inline comment after a section header — fixed.** Correct, and it applied equally to
the pre-existing `parse_ansible_cfg_path_list`, so rather than fix one copy I extracted a shared
`parse_ansible_cfg_section_header` that mirrors configparser's `SECTCRE` (`\[(?P<header>.+)\]`,
matched not fullmatched, greedy `.+` → name runs to the last `]`, trailing text ignored). Both
helpers now go through it, so `[defaults] # note` is fixed too.

Regression tests added for all three: fail-closed cfg emission, writable-parent refusal,
sticky-parent acceptance, and trailing-text headers on both helpers.

rubenfiszel:
Good catch — fixed in 71fda9b4a5.

You're right, and it was precisely the case I'd explicitly blessed: `recursive(true)` reports
success for a path that already exists, so the `NotFound` branch created-or-accepted whatever
was there and returned **without validating**, leaving `SOCKET_ROOT_TRUSTED` set. A sticky
parent stops another uid *renaming* our entry away, but nothing stops them *creating* the
not-yet-existing `pc` first — so under `WINDMILL_DIR=/tmp` the racer's dir was trusted.

The fix removes the early return rather than patching it: create the leaf **non-recursively**
(create-or-`AlreadyExists`, which is atomic), then validate whatever is actually on disk on
every path. There is now no exit from `prepare_socket_root` that neither validates nor marks
the root untrusted — the sole unvalidated exit is a `read_dir` failure *after* validation
passed, which just skips the sweep. Ours-and-0700 passes; a raced dir fails on uid or mode; a
raced symlink fails `is_dir` (`symlink_metadata` doesn't follow).

Two related things while in there:

- The parent check now uses resolved `metadata` instead of `symlink_metadata`. A symlinked
  parent is normal (macOS `/tmp` → `/private/tmp`), and since symlink modes are always 0777 the
  old check would have declared every macOS dev box untrusted. What matters for the rename race
  is the mode of the directory the entries actually live in.
- The parent is created with `create_dir_all` first, so a fresh install where `WINDMILL_DIR`
  doesn't exist yet doesn't fall into the "cannot stat parent" path.

On test coverage, being straight about what the new test does and doesn't do: it pins that a
sticky parent does not exempt the root from validation (the combination nothing else covered),
but it cannot reproduce the check-then-create *interleaving* — there's no hook to deterministically
win that race in a unit test. What actually closes the hole is structural: validation is now
unconditional and post-creation, so there is no ordering in which an unvalidated root is used.

rubenfiszel:
**You're right and I was wrong — I need to correct my last round's claim.** Fixed in 0090871582.

I said "`/tmp/windmill` is **not** 0777 in the standard image" and cited `Dockerfile:266`/`:337`.
I stopped reading at line 338. Line 341 is:

```dockerfile
# Make directories world-accessible for any UID
RUN find ${APP} /tmp/windmill -type d -exec chmod 777 {} +
```

That recursively makes `/tmp/windmill` a non-sticky 0777, so my parent check would have cleared
`SOCKET_ROOT_TRUSTED` at startup in the standard image and **silently disabled this entire fix in
production** — every network playbook still failing with the original `AF_UNIX` error, with only a
log line to show for it. Worse, no local test would have caught it: a dev box's `/tmp/windmill` is
0755, so the suite stayed green while the shipped artifact was broken. Thanks for pushing on this.

**Fix: `/tmp/wm-pc`, i.e. a protected parent** (your second option) rather than making the shipped
parent sticky. `/tmp` is `1777`, and the sticky bit is exactly the property the trust check wants
from a parent — it's why `/tmp` has it. Worst-case socket path is now 85 bytes, and startup on this
box confirms `ROOT=/tmp/wm-pc TRUSTED=true MODE=700`, with the real `network_cli` play binding its
socket underneath.

I did **not** take the `01777`-on-`/tmp/windmill` option, because `clean_cache` does
`remove_dir_all(/tmp/windmill/cache)` (`common.rs:1793`) and the image's whole reason for the 0777
is to support running as an arbitrary UID. Under a sticky `/tmp/windmill`, a runtime UID could no
longer remove the build-time root-owned `cache` entry — EPERM. That trade turns a socket-path fix
into a cache-cleanup regression for exactly the deployments the 0777 exists to serve.

The `WINDMILL_DIR` rooting was my own doing, from an earlier round's reasoning that it "adds no new
squattable path since the worker already owns it". That was backwards: the shipped image makes it
the *most* squattable path on the box. The env knob it gave for free is preserved as
`WINDMILL_ANSIBLE_CONTROL_PATH_ROOT` for hosts where `/tmp` is unsuitable.

**Coverage.** Rather than a permissions test (host-dependent — it would pass on any dev box and
still miss the image), I pinned the invariant that actually broke: the default root must hang off
`/tmp` and must not live under `WINDMILL_DIR`
(`test_default_control_path_root_hangs_off_tmp_not_windmill_dir`). The sticky-parent-accepted and
writable-parent-refused tests already cover the mode logic itself. PR body updated to match.

---

@pi both nits fixed: `parse_ansible_cfg_section_header` moved above the `parse_ansible_cfg_path_list`
doc block so each doc reattaches to its own function, and the startup warning no longer prints the
computed length where the limit belongs — it now reads "may reach {n} bytes, over the AF_UNIX limit
of 107", with 107 hoisted into a named `AF_UNIX_PATH_LIMIT` the budget assert shares.

rubenfiszel:
Both addressed in 756c0e10d2.

**1. [P1] Sweep could delete unrelated directories under a custom root — fixed.** Fair, and it's a
hazard I introduced myself by adding the override: the root became operator-configurable while the
sweep still recursively deleted *anything* under it past the cutoff. Point it at a directory that
holds more than socket dirs and startup quietly eats data.

The sweep now only touches entries whose name is one this module could have created — 32-hex
`Uuid::simple`, via `is_persistent_control_path_dir_name`. Notably the hyphenated UUID form is
rejected too: we don't create it, so it isn't ours to delete.

I checked the guard isn't decorative by reverting it against the new test:

```
thread '...test_prepare_socket_root_sweeps_only_stale_dirs' panicked:
a stale dir we never created must be left alone
test result: FAILED. 0 passed; 1 failed
```

That also flushed out three older tests that asserted survival of dirs named `stale-job` /
`precious` — with the name filter in place they'd have passed for the *wrong reason*, no longer
proving the trust check or the symlink no-follow stopped the sweep. They're UUID-named now, so
they still fail if the behavior they guard regresses.

**2. [P2] Stale `WINDMILL_DIR` security contract — fixed** (@claude flagged the same line). The
`prepare_socket_root` SECURITY doc now describes the trust boundary as it is: a world-writable
`/tmp` parent, no reference to a root location that no longer exists. The two surviving
`WINDMILL_DIR` mentions are deliberate — they're the durable rationale for why the root is *not*
there (the image's recursive `chmod 777`), which is the constraint a future reader would otherwise
break by "simplifying" the root back under `WINDMILL_DIR`.

rubenfiszel:
Both fixed in 2c136c6ef8 — by removing their shared cause rather than hardening around it.

Both findings exist **only because the root was configurable**, and that knob was mine: it wasn't
in the original task, I added it in an earlier round to satisfy a "no escape hatch" nit. It has
since produced three separate P1s across three rounds (the unrelated-data sweep last round, and
both of these). That's a strong signal the knob costs more than it's worth, so it's gone. The root
is a fixed `const PERSISTENT_CONTROL_PATH_ROOT: &str = "/tmp/wm-pc"`.

**1. [P1] Replaceable ancestors — gone.** With a fixed path there are no operator-supplied
components: the only ancestor is `/tmp` itself (root-owned, sticky `1777`). An attacker who can
retarget `/tmp` owns the box already, and no amount of `openat2`/`RESOLVE_NO_SYMLINKS` plumbing in
this module would help. That machinery only ever earned its keep against the override.

**2. [P1] UUID-named dirs are common — no longer reachable.** The premise was a custom root
"containing an unrelated stale UUID directory". There is no custom root now: it's a dir we create
`0700` and own, validated at startup, and a pre-existing one that isn't exclusively ours is refused
outright (fail-closed, never swept). So a foreign UUID dir cannot appear there — the sweep is
bounded by the root check, not by name-guessing. I kept the UUID filter anyway as belt-and-braces,
since this is a recursive delete running as the worker's uid, and it's one cheap line.

**A bonus from going fixed**: the socket-path budget is now a compile-time assert instead of a
runtime warning. A root that would reintroduce the original bug fails to build:

```
error[E0080]: evaluation panicked: assertion failed:
  PERSISTENT_CONTROL_PATH_ROOT.len() + 1 + 32 + 1 + 40 <= AF_UNIX_PATH_LIMIT
```

(verified by temporarily lengthening the const). That's a strictly better guard than the warn it
replaces, which nobody would have read.

If a real deployment ever turns up where `/tmp` is unusable, that's a concrete follow-up with a
concrete constraint to design against — and it fails closed safely today rather than silently.
PR body updated to match; 32 tests green.

rubenfiszel:
Both fixed in de1934c99c.

**1. [P1] Owner-unwritable root stays trusted — fixed.** Right: my check tested only "is it
*safe*", never "is it *usable*". A worker-owned `0500` root passes uid and `mode & 0o022`, so the
trust flag stayed set, jobs got the control path, and ansible then failed to mkdir under it — a
permission error on every network playbook, which is *worse* than the fallback the fail-closed path
exists to give them. Trust now also requires owner rwx (`mode & 0o700 == 0o700`).

I chose reject over repair: chmod-ing a directory we did not create is how you paper over a
misconfiguration you should be reporting, and the untrusted path already logs loudly and degrades
to something that works.

Verified the check isn't decorative by removing it against the new test:

```
a root we cannot write must not be trusted
test result: FAILED. 0 passed; 1 failed
```

**2. [P2] Root-location test fails for supported `WINDMILL_DIR` values — fixed, and it was worse
than flaky.** `"/tmp/wm-pc".starts_with("/tmp")` is `true`, so the assert would have failed for
`WINDMILL_DIR=/tmp`, and `starts_with` being a *string* compare meant `/tmp/wm` false-positived too.

The deeper problem is that the assert encoded my drafting history rather than the constraint. If
`WINDMILL_DIR` genuinely were `/tmp`, a root under it would be perfectly fine — because the
invariant was never "not under `WINDMILL_DIR`", it's "parented by a sticky dir". The remaining
parent assertion is the whole property; the `WINDMILL_DIR` trap survives as the doc comment
explaining *why*, which is where rationale belongs. Confirmed green under `WINDMILL_DIR` of `/tmp`,
`/tmp/wm`, and `/tmp/windmill` — all three would have failed or falsely passed before.

33 tests green.

rubenfiszel:
Fixed in 9cb95aa1b9.

**[P1] Delegated config overwrites an explicit env override — correct, and it broke my own stated
contract.** The whole reason this value is conditional is that, unlike `ANSIBLE_HOME`, it isn't
runtime-bound, so Windmill shouldn't outrank the user. I enforced that against the repo's
`ansible.cfg` and then ignored the job's own env — and since `ansible_env_overrides` is applied
after `.envs(envs)` (line 1981), a job setting `ANSIBLE_PERSISTENT_CONTROL_PATH_DIR` had it
silently replaced. Half a contract is a bug.

The default is now applied only when *neither* the delegated cfg *nor* the job env declares it.
The var name is a shared const rather than a string literal in three places.

Worth noting the non-delegated path was already correct for free: there Windmill writes the value
into `ansible.cfg`, and ansible ranks env above cfg, so a job-set env var wins by ansible's own
precedence. The delegate path was the only one where we injected the env var ourselves and thus
had to defer explicitly.

Verified the guard isn't decorative by removing it against the new test:

```
must not override a control_path_dir the job set itself
test result: FAILED. 0 passed; 1 failed
```

34 tests green.

rubenfiszel:
✅ Review round clean @ 59c40dbe34dca8adcd246224fe8d3384c7a70dd4 — nit-only verdicts at 9cb95aa1b9a0828df83dcea815539230d4921c81 (codex: mergeable-with-nits; claude, pi: good to merge); nit addressed in 59c40dbe34.

The stale `/tmp` parent comment is gone, along with the `create_dir_all(parent)` it described — that step only existed to create `WINDMILL_DIR` on a fresh install, and `/tmp` is system-provided, so it was dead code rather than a comment to reword. A missing parent now falls to the existing "cannot stat its parent" fail-closed path, which is the correct handling.

Re-verified on the final head: 34 tests green; startup creates and trusts the real root (`/tmp/wm-pc`, mode 700); and the real `network_cli` + `cisco.nxos` play from the reported 90-char job dir binds its socket under `/tmp/wm-pc/<uuid>/` with no AF_UNIX error (85-byte worst case vs the 107 limit), reaching the device-dial stage instead.

One caveat stated plainly: I verified the non-sandboxed path. I could not exercise nsjail on this box, so that path rests on the reasoning in the PR body — the jail's `/tmp` is jail-private (tmpfs, or a bind under `job_dir`), so the short path is created and torn down with the jail, and the host-side guard no-ops on a path never created there.

PR #10173, opened 2026-07-17
fix(extra): make the extra container runnable as a non-root UID

rubenfiszel:
## Summary

Reported by a user running under a cluster policy that forbids root: the debugger in `windmill-extra` fails with permission errors because cache directories are owned by root. They confirmed it still reproduces on v1.746.0, and that mounting an `emptyDir` over `/tmp/windmill/cache` works around it.

**Root cause.** The base image (`windmill-ee-slim`) sets `UV_CACHE_DIR=/tmp/windmill/cache/uv` and `XDG_CACHE_HOME=/pyls/.cache`, and ends with a pass that makes `/tmp/windmill` world-writable "for arbitrary UID support". `DockerfileExtra` then runs `uv pip install` **after** that pass, as root, so uv repopulates its own cache and leaves ~320 root-owned `0755` dirs behind. Verified against the published image — a non-root UID hits exactly the reported failure:

```
uv venv failed: Failed to install seed packages into virtual environment
  Caused by: Failed to write to the client cache
  Caused by: Permission denied (os error 13)
    at path "/tmp/windmill/cache/uv/simple-v21/pypi/.tmpWxWzTi"
```

That call is `windmill prepare-deps`, which the debugger spawns to install job dependencies — hence "the debugger doesn't work rootless". The `emptyDir` workaround works precisely because it replaces the root-owned cache with an empty writable one.

Two adjacent rootless bugs in the same image, found while verifying:

- `/pyls/.cache` (`XDG_CACHE_HOME`, incl. `DENO_DIR`) is root-owned `0755` — the LSP can't write its cache.
- The entrypoint wrote `$NETRC` to the `0700` `/root` under `set -e`, so setting `NETRC` **killed the container at startup** for any non-root UID.

The user also correctly noted the extra container has no `windmill` user: the app image creates uid/gid 1000, the slim base it builds on doesn't, so `runAsUser: 1000` had no account and no writable `$HOME`.

No helm chart change is needed — this removes the reason for the volume workaround. (The chart lives in `windmill-helm-charts`; if you'd still like the volume hooks for other reasons, that's a separate PR.)

## Changes

- `docker/DockerfileExtra`: re-apply the base image's cache convention to `/tmp/windmill` and `/pyls/.cache` **after** the root-run installs (`chmod -R a+rw` + `777` dirs, same idiom as `DockerfileSlimEe:69`/`:121`, which already ships 666 cache files by design — so this restores the base's posture rather than widening it). `/tmp/monaco` holds `node_modules` code rather than cache, so it gets `a+rX` files with `777` dirs: enough for the `go.mod`/`ruff.toml` written into it at runtime, without making the code world-writable.
- `docker/DockerfileExtra`: create the `windmill` uid/gid 1000 account, matching `Dockerfile:333`. No `USER` directive, so the image still starts as root by default. With the passwd entry, `runAsUser: 1000` resolves to a real user with a writable `/home/windmill`.
- `docker/entrypoint-extra.sh`: redirect `$HOME` to `/tmp/windmill-home-$(id -u)` when it isn't writable, before anything writes under it. Root is unaffected. UID-scoped so a leftover dir from another UID on a shared `/tmp` can't reintroduce the same startup death.

## Test plan

Verified locally against a **full source build** of `DockerfileExtra` (not layered on the published image), so both new `RUN` steps are proven to execute. CI is unchanged and covers the root path only, so the non-root evidence below is local:

- [x] **Before**: `prepare-deps` as uid 1000 fails with the EACCES above; as root it succeeds → confirms ownership, not logic
- [x] **After** — all four services start and `prepare-deps` returns `"success":true` as:
  - [x] root (default) — no `$HOME` redirect, netrc still at `/root/.netrc`, behavior unchanged
  - [x] `--user 1000` — resolves to `windmill`, `HOME=/home/windmill`
  - [x] `--user 4567:0` (arbitrary UID, OpenShift-style) — redirect fires
- [x] 0 non-writable dirs remain under the cache paths for a non-root UID
- [x] `NETRC` set no longer kills startup; lands `0600` in the writable HOME
- [x] Full CI smoke suite (`docker/test_windmill_extra.ts`) passes **7/7 as `--user 1000`**, including real DAP *Python execution* and *breakpoint support* sessions — the reported scenario, working rootless
- [x] Negative control: the same `prepare-deps` check returns `"success":false` on the current published image and `"success":true` on this build, confirming the failure is real and fixed
- [x] Regressions caught in review and fixed: unset `HOME` leaked netrc to `/.netrc`; root-created `/tmp/windmill-home` on a shared `/tmp` re-broke non-root startup
- [x] `shellcheck -S warning` clean, `docker build --check` clean

Note: `publish_extra.yml` builds and smoke-tests this image on tag releases only, as root. Non-root remains covered by the manual verification above.

---
## Summary by cubic
Make the `windmill-extra` image runnable as a non-root UID by fixing cache and HOME handling so the debugger and LSP work without root. The previously added non-root CI smoke test has been removed.

- **Bug Fixes**
  - Dockerfile: add `windmill` user (UID/GID 1000); re-apply permissions after root installs so `/tmp/windmill` and `/pyls/.cache` are world-writable; keep `/tmp/monaco` code read-only (`a+rX` files, `777` dirs) for metadata writes.
  - Entrypoint: redirect unwritable `$HOME` to `/tmp/windmill-home-$(id -u)`; write `$NETRC` to `$HOME/.netrc`; prefer `$HOME/.cache` for `XDG_CACHE_HOME` and fall back to `/pyls/.cache`.

- **Refactors**
  - CI: revert the non-root smoke run; keep the root-only smoke test.

<sup>Written for commit b49c2a8e9737b6c5764e31b69d85059d2614f679. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/publish_extra.yml
+++ b/.github/workflows/publish_extra.yml
@@ -70,54 +70,17 @@ jobs:
         run: |
           bun run docker/test_windmill_extra.ts
 
-      # Clusters that forbid root run this image under `runAsUser`. The build steps
-      # write caches as root, so only a non-root run catches dirs left unwritable.
-      - name: Start container as non-root UID
-        run: |
-          docker run -d --name windmill-extra-nonroot \
-            --user 1000 \
-            -p 4001:3001 -p 4002:3002 -p 4003:3003 \
-            -e ENABLE_LSP=true \
-            -e ENABLE_MULTIPLAYER=true \
-            -e ENABLE_DEBUGGER=true \
-            -e DEBUGGER_PORT=3003 \
-            -e REQUIRE_SIGNED_DEBUG_REQUESTS=false \
-            -e NETRC="machine example.com login user password pass" \
-            windmill-extra:test
-
-          echo "Waiting for container to initialize..."
-          sleep 10
-          docker logs windmill-extra-nonroot
-
-      - name: Run integration tests as non-root UID
-        env:
-          LSP_PORT: 4001
-          MULTIPLAYER_PORT: 4002
-          DEBUGGER_PORT: 4003
-        run: |
-          bun run docker/test_windmill_extra.ts
-
-          # The debugger installs job dependencies through the uv cache, which is
-          # where a root-owned build cache surfaces as EACCES.
-          echo "Checking dependency install works as a non-root UID..."
-          docker exec windmill-extra-nonroot bash -c \
-            'echo "{\"code\":\"import requests\\ndef main():\\n  return 1\\n\",\"language\":\"python3\"}" \
-              | windmill prepare-deps' | tee /tmp/prepare-deps.json
-          grep -q '"success":true' /tmp/prepare-deps.json
-
       - name: Show container logs on failure
         if: failure()
         run: |
-          echo "=== Container logs (root) ==="
+          echo "=== Container logs ==="
           docker logs windmill-extra-test
-          echo "=== Container logs (non-root) ==="
-          docker logs windmill-extra-nonroot || true
 
       - name: Cleanup
         if: always()
         run: |
-          docker stop windmill-extra-test windmill-extra-nonroot || true
-          docker rm windmill-extra-test windmill-extra-nonroot || true
+          docker stop windmill-extra-test || true
+          docker rm windmill-extra-test || true
 
   publish_extra:
     needs: [sleep, test_extra]
--- a/.github/workflows/publish_extra.yml
+++ b/.github/workflows/publish_extra.yml
@@ -70,17 +70,54 @@ jobs:
         run: |
           bun run docker/test_windmill_extra.ts
 
+      # Clusters that forbid root run this image under `runAsUser`. The build steps
+      # write caches as root, so only a non-root run catches dirs left unwritable.
+      - name: Start container as non-root UID
+        run: |
+          docker run -d --name windmill-extra-nonroot \
+            --user 1000 \
+            -p 4001:3001 -p 4002:3002 -p 4003:3003 \
+            -e ENABLE_LSP=true \
+            -e ENABLE_MULTIPLAYER=true \
+            -e ENABLE_DEBUGGER=true \
+            -e DEBUGGER_PORT=3003 \
+            -e REQUIRE_SIGNED_DEBUG_REQUESTS=false \
+            -e NETRC="machine example.com login user password pass" \
+            windmill-extra:test
+
+          echo "Waiting for container to initialize..."
+          sleep 10
+          docker logs windmill-extra-nonroot
+
+      - name: Run integration tests as non-root UID
+        env:
+          LSP_PORT: 4001
+          MULTIPLAYER_PORT: 4002
+          DEBUGGER_PORT: 4003
+        run: |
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Thanks — addressed the valid half of the nit in b39566b38f, and correcting the record on the rest.

**`/tmp/monaco` — agreed, fixed.** It holds `node_modules` (code, not cache), so the cache convention doesn't cover it. Now `a+rX` on files with `777` dirs, which is still enough for the `go.mod` and `ruff.toml` that the entrypoint and `pyls_launcher.py` write into it. Verified as a non-root UID: those writes still succeed, `node_modules` code is readable but no longer world-writable.

**`/tmp/windmill` + `/pyls/.cache` — keeping `a+rw`, the premise is off.** The base idiom isn't only `find -type d -exec chmod 777`; it makes cache *files* world-writable too, deliberately:

- `DockerfileSlimEe:69` / `Dockerfile:268` — `chmod -R a+rw /tmp/windmill/cache`, recursive, files included
- `Dockerfile:339` — *"cache files already have 666 from umask copy above"*

So `a+rw` on the cache is matching the existing posture, not widening it — the "no new posture" claim holds for `/tmp/windmill`. It's also load-bearing: uv rewrites cache entries in place, and dropping to `a+rX` would reintroduce the EACCES this PR fixes for any entry the root-run installs left behind.

**Re: `HOME` for `--user 1000` — this one is empirically the other way round.** Docker *does* derive `HOME` from `/etc/passwd` when the UID resolves, no `USER` directive needed. Measured on the built image:

```
$ docker run --rm --user 1000 --entrypoint bash <img> -c 'id; whoami; echo HOME=$HOME'
uid=1000(windmill) gid=1000(windmill) groups=1000(windmill),100(users)
windmill
HOME=/home/windmill
```

That's exactly why the `adduser` line is load-bearing rather than cosmetic: without the passwd entry uid 1000 gets `HOME=/`, with it a writable `/home/windmill`, so the redirect branch is skipped. The test plan is accurate as written. The `/tmp/windmill-home-$(id -u)` branch covers arbitrary UIDs that *don't* resolve (`--user 4567:0`), which is verified separately.

rubenfiszel:
✅ Review round clean @ b39566b38ff0839cebace0663ab9c6315bdd4ed8 — nit-only verdicts at e4bc9d229f84fc632ac9db9730038ab5b0817d77 (codex: Good to merge; pi: Good to merge; claude: nit-only); `/tmp/monaco` nit addressed in b39566b38f, the `a+rw`-on-cache and `HOME`-derivation points dismissed in review replies with evidence

PR #10176, opened 2026-07-17
fix(flows): make updateFlow body path optional so AI can update flows

rubenfiszel:
## Summary

Fixes GIT-925. The MCP `updateFlow` tool (and any client hitting `POST /flows/update/{path}` without a body `path`) failed with:

```
HTTP 422: Failed to deserialize the JSON body into the target type: missing field `path`
```

### Root cause

`update_flow` deserializes its body as `Json<NewFlow>`, and `NewFlow.path` is a **required** `String`. `NewFlow` is shared with `createFlow`, where `path` genuinely is required (there is no URL to take it from). But on update, the flow to change is already identified by the URL `{path}`, so the body `path` is only needed to *rename*. A client that updated a flow in place, sending a body without `path`, was rejected by serde before the handler ran.

`updateFlow` was the only in-place update endpoint to break outright: `updateVariable`, `updateResource`, and `updateApp` already take an **optional** body `path` (`EditVariable` / `EditResource` / `EditApp`), so they tolerated its absence and kept the existing path. `updateFlow` reusing the create struct was the odd one out.

### Fix (server-side, covers every client)

Add an `EditFlow` request struct that mirrors `NewFlow` but with `path: Option<String>`, and have `update_flow` default the body path from the URL when it is omitted (`into_new_flow`). This matches the existing `Edit*` convention and fixes the 422 at the API layer for the MCP tool, the in-app AI chat, and raw HTTP alike. `createFlow` is untouched and still requires `path`.

A body `path` that differs from the URL still renames the flow, exactly as before.

### MCP schema cleanup (same PR)

While here, the MCP endpoint-tool generator no longer name-mangles the *path* side of a field collision. `path` collides with the body field on the four `.../update/{path}` tools, and the generator used to expose it to the model as `path__path`. Now the path parameter keeps its plain name (identical to `getFlowByPath`), and only the body side becomes an optional `path__body` (used to rename). The now-dead `path_field_renames` plumbing is removed from the generator, the `EndpointTool` struct, the TS interface, and `substitute_path_params`; it was never part of the openapi `EndpointTool` schema, so generated types are unchanged.

## Changes

- `windmill-types/src/flows.rs`: add `EditFlow` (optional `path`) + `into_new_flow(url_path)`.
- `windmill-api-flows/src/flows.rs`: `update_flow` takes `Json<EditFlow>` and defaults the path from the URL.
- `generate_mcp_endpoints_tools/generate_mcp_tools.py`: never suffix the path side of a collision; drop the colliding body field from `required`; remove `path_field_renames` emission.
- `windmill-api/src/mcp/{utils,core}.rs`, `windmill-mcp/src/server/endpoints.rs`: remove the dead `path_field_renames` plumbing and the interim MCP body-path fallback (the server now owns defaulting), simplifying `build_request_body` and `substitute_path_params`.
- Regenerated `windmill-api/src/mcp/auto_generated_endpoints.rs` and `frontend/src/lib/mcpEndpointTools.ts`. The frontend file is generated data (consumed for `method` and by the AI-chat tool builder); no visible UI effect, so no screenshots.

## In-app AI chat

The in-app AI chat's API mode (`copilot/chat/api/apiTools.ts`) builds these requests in the browser. Its **in-place** updates are now fixed by the server change (it posts a body without `path`, which the server defaults from the URL). Renaming a flow *from the chat* is still unsupported (that path posts `path__body`, which the chat does not map back to `path`), and is left as a small follow-up (needs `body_field_renames` mirrored in TS). Out of scope here; only verifiable with an AI provider configured.

## Test plan

Server-side behavior verified end-to-end via raw HTTP on a plain `--features quickjs` backend (no `mcp` feature needed):

- [x] **Repro**: pre-fix, `POST /flows/update/{path}` with a body lacking `path` returns `422 ... missing field 'path'`.
- [x] After: same call returns 200 and updates in place.
- [x] Body with a different `path` renames the flow (old path 404s, new path carries the update).
- [x] Body with `path: null` (the LLM/chat shape) returns 200 and defaults from the URL.

MCP round-trip verified with `--features quickjs,mcp` against the live MCP server:

- [x] `updateFlow` with only `path` + `summary` + `value` returns 200 (server defaults the body path; MCP no longer fills it).
- [x] `updateFlow` with `path__body` renames the flow.
- [x] Tool schema advertises `required: [path, summary, value]`; no `path__path` anywhere in the repo.

- [x] `cargo test -p windmill-types --lib flows::` (13 passed, incl. the new `EditFlow` default/rename test), `cargo test -p windmill-api --features mcp --lib mcp::utils` (17 passed), `cargo check -p windmill-api --features mcp` clean, generator re-run byte-stable.

### DIFF
--- a/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
+++ b/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
@@ -165,13 +165,14 @@ def extract_separate_schemas(parameters: List[Dict[str, Any]], request_body: Opt
 
     conflicts = (path_keys & query_keys) | (path_keys & body_keys) | (query_keys & body_keys)
 
-    path_field_renames = {}
     query_field_renames = {}
     body_field_renames = {}
 
     for field in conflicts:
+        # The path parameter keeps the plain name: it identifies the item, is what a
+        # caller reaches for first, and matches the non-colliding endpoints (`getFlowByPath`
+        # takes `path`). Only the other locations carry a suffix.
         schemas_and_renames = [
-            (path_params_schema, path_keys, '__path', path_field_renames),
             (query_params_schema, query_keys, '__query', query_field_renames),
             (body_schema, body_keys, '__body', body_field_renames),
         ]
@@ -209,7 +210,7 @@ def extract_separate_schemas(parameters: List[Dict[str, Any]], request_body: Opt
             if isinstance(prop, dict):
                 existing_desc = prop.get('description', '').rstrip('. ')
                 prop['description'] = (
-                    f"{existing_desc}. Defaults to `{field}__path` when omitted; "
+                    f"{existing_desc}. Defaults to `{field}` when omitted; "
                     f"set it only to change the {field}."
                 ).lstrip('. ')
 
@@ -217,7 +218,7 @@ def extract_separate_schemas(parameters: List[Dict[str, Any]], request_body: Opt
     path_params_schema = path_params_schema if path_params_schema and path_params_schema.get('properties') else None
     query_params_schema = query_params_schema if query_params_schema and query_params_schema.get('properties') else None
 
-    return (path_params_schema, query_params_schema, body_schema, path_field_renames, query_field_renames, body_field_renames)
+    return (path_params_schema, query_params_schema, body_schema, query_field_renames, body_field_renames)
 
 # Cache for loaded external files
 _external_file_cache: Dict[str, Dict[str, Any]] = {}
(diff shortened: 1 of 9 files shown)


rubenfiszel:
✅ Review round clean @ e041acbb21bab3c1e497d56efda616eb09365db1 — nit-only verdicts at f4019dfff4a5bb733e0abfa584edd1c668632b3a (codex nit; claude and pi "Good to merge"); nit addressed in e041acbb21bab3c1e497d56efda616eb09365db1

rubenfiszel:
✅ Review round clean @ ff49507c4d9ffd19750638d4d3676d168a29825d

rubenfiszel:
✅ Review round clean @ e2976b6d48d57cc154e134b7c299d00213cc0209

PR #10177, opened 2026-07-17
feat(worker-tags): add `*` fork marker to workspace-scoped custom tags

rubenfiszel:
## Summary

Custom worker tags can be scoped to workspaces (`mytag(ws1+ws2)` = only those workspaces may use the tag; `mytag(^ws1^ws2)` = every workspace except those). Matching was a literal id comparison, so a **fork never matched its parent's scoped tag** — a workspace forked from `linux` could not use `linux-common-dev(linux)`, which is surprising when the tag names a shared dev worker group.

Inheriting unconditionally isn't right either: a restricted worker group may exist precisely to encode "only code that made it into this workspace runs on this hardware", and a fork is by definition where the code isn't reviewed yet. That distinction is a property of the worker group, not of the instance, so this adds a per-entry opt-in marker rather than a global toggle.

A `*` suffixed to a workspace id extends that entry to the workspace's forks, transitively:

```
linux-common-dev(linux*)          linux and all its forks
linux-common-dev(linux*+shared)   linux and its forks, plus exactly `shared`
sensitive(^prod*)                 everyone except prod and its forks
sensitive(^prod)                  everyone except prod itself (unchanged)
```

The marker is opt-in in **both** grammar forms, so **no existing tag string changes meaning on upgrade** — this is purely additive. `*` cannot collide with a real workspace id: the `proper_id` check constraint restricts ids to `^\w+(-\w+)*$`.

The setting itself is already instance-scoped and superadmin-only (`set_global_setting` is `require_super_admin`), so the marker adds an argument to a knob a superadmin already exclusively controls, not a new configuration surface.

## Changes

- `windmill-common/src/worker.rs`: extend `CUSTOM_TAG_REGEX` with an optional `*` per workspace id; replace `SpecificTagData.workspaces: Vec<String>` with `Vec<WorkspaceMatcher { id, include_forks }>`; `applies_to_workspace` now takes the workspace's id chain (itself + fork ancestors) instead of a single id. `WorkspaceMatcher`'s `Debug` renders the authored `prod*` form, since `CustomTags` is `{:?}`-dumped into the operator-facing "not in the allowed CUSTOM_TAGS" error.
- `windmill-common/src/workspaces.rs`: add `workspace_with_fork_ancestors`, a thin wrapper over the existing cached `fork_ancestor_chain`.
- `windmill-common/src/jobs.rs`: resolve the chain in `check_tag_available_for_workspace_internal`, but **only when the tag actually carries a marker** (`is_fork_scoped()`), so no DB lookup is added to the job-push path for anyone not using the feature. Its first parameter narrows from `impl PgExecutor` to `&DB`; all four call sites already passed a `&DB`.
- `windmill-api-workers/src/lib.rs`: resolve the chain in the per-workspace tag listing and in `exists_workers_with_tags`.
- `AssignableTagsInner.svelte`: mirror the regex, spell the marker out in the live preview ("linux (and its forks)"), flag a stray `*` as an invalid tag, document the syntax.

## Screenshots

tag-fork-marker-include

tag-fork-marker-exclude

## Test plan

- [x] `cargo test -p windmill-common --lib worker::tests` — 35 pass, including 3 new: marker parse + round-trip (a dropped `*` would silently rewrite an operator's security config on every save), and marker-vs-unmarked behaviour in both grammar forms including a nested fork.
- [x] `cargo test --test worker test_fork_marker_tag_admission_through_lineage` — new integration test pinning admission through a real `parent_workspace_id` row. Mutation-checked: neutering `is_fork_scoped()` makes it fail as intended, so it is not vacuous.
- [x] Full `cargo check` and frontend `npm run check:fast` clean.
- [x] Exercised against a running server with a parent, a fork, a fork-of-fork and an unrelated workspace, **as a non-superadmin** (superadmins bypass the tag check entirely). Both the listing endpoint and job-push enforcement agree in all cells:

  | tag | parent | fork | fork-of-fork | unrelated |
  |---|---|---|---|---|
  | `bare(tagparent)` | allowed | **denied** | denied | denied |
  | `forky(tagparent*)` | allowed | **allowed** | **allowed** | denied |
  | `excl(^tagparent)` | denied | allowed | allowed | allowed |
  | `exclforky(^tagparent*)` | denied | **denied** | denied | allowed |

- [x] UI verified in-browser: preview renders per-entry fork annotation; saved badges round-trip `forky(tagparent*)` with the marker intact.

- The `*` syntax should be documented on the `worker_groups` docs page (separate `windmilldocs` repo).

---
## Summary by cubic
Adds an opt-in `*` fork marker to workspace-scoped custom worker tags so tags can include or exclude a workspace’s forks transitively. Also tightens the global tags route by requiring workspace membership before resolving fork lineage.

- New Features
  - Extend tag grammar to `tag(ws1*+ws2)` and `tag(^ws1*^ws2)` to include/exclude forks.
  - Matching uses the workspace id chain only for fork-scoped tags; `check_tag_available_for_workspace_internal` resolves lineage via `workspace_with_fork_ancestors` only when needed.
  - `get_custom_tags_for_workspace` and `exists_workers_with_tags` now filter using the id chain; `CustomTags::to_string_vec` accepts a chain and round-trips `prod*`.
  - UI mirrors the regex, previews “and its forks,” flags stray `*`, and updates the help text.

- Bug Fixes
  - Gate fork-lineage lookup in `exists_workers_with_tags` on workspace membership to avoid disclosing ancestry; non-members get an empty result.
  - Add `sqlx` offline cache for the fork-marker test INSERT to keep `SQLX_OFFLINE=true` CI green.

<sup>Written for commit ec60b972c9ffb1c29ba64fce9ce4f4a3f7f1d092. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10177?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-be21088e8b88e01b50a544f220e27d1e30b63c946a4eb640e55850100c3edd3c.json
+++ b/backend/.sqlx/query-be21088e8b88e01b50a544f220e27d1e30b63c946a4eb640e55850100c3edd3c.json
@@ -1,0 +1,14 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "INSERT INTO workspace (id, name, owner, parent_workspace_id)\n         VALUES ($1, $1, 'test-user', 'test-workspace')",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Varchar"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "be21088e8b88e01b50a544f220e27d1e30b63c946a4eb640e55850100c3edd3c"
+}
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -5340,7 +5340,7 @@ async fn test_duckdb_ffi(db: Pool<Postgres>) -> anyhow::Result<()> {
 #[sqlx::test(fixtures("base"))]
 async fn test_flow_substep_tag_availability_check(db: Pool<Postgres>) -> anyhow::Result<()> {
     use windmill_common::worker::{
-        CustomTags, SpecificTagData, SpecificTagType, CUSTOM_TAGS_PER_WORKSPACE,
+        CustomTags, SpecificTagData, SpecificTagType, WorkspaceMatcher, CUSTOM_TAGS_PER_WORKSPACE,
     };
 
     initialize_tracing().await;
@@ -5354,7 +5354,10 @@ async fn test_flow_substep_tag_availability_check(db: Pool<Postgres>) -> anyhow:
             "restricted-tag".to_string(),
             SpecificTagData {
                 tag_type: SpecificTagType::NoneExcept,
-                workspaces: vec!["other-workspace".to_string()],
+                workspaces: vec![WorkspaceMatcher {
+                    id: "other-workspace".to_string(),
+                    include_forks: false,
+                }],
             },
         )]),
     }));
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -5338,6 +5338,7 @@ async fn test_duckdb_ffi(db: Pool<Postgres>) -> anyhow::Result<()> {
 /// This validates that `check_tag_available_for_workspace_internal` is properly called
 /// when pushing jobs from worker_flow.
 #[sqlx::test(fixtures("base"))]
+#[serial]
 async fn test_flow_substep_tag_availability_check(db: Pool<Postgres>) -> anyhow::Result<()> {
     use windmill_common::worker::{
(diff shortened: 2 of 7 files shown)


rubenfiszel:
✅ Review round clean @ d3da74ef3cbb7f6e9a2b1555b3cab3a690ea5912

PR #10179, opened 2026-07-17
fix(schedules): stop disabling schedules on transient push errors

rubenfiszel:
## Summary

Fixes WIN-2198. Two related changes:

1. **Stop disabling schedules on transient push errors.** A scheduled flow whose next-occurrence push failed after retry exhaustion used to be **disabled**, silently killing a healthy schedule over a transient DB blip. Now transient failures leave the schedule enabled; the current run completes and the unarmed-schedule reconciler (#10174) re-arms the next occurrence.
2. **Make the reconciler robust to schedules that genuinely can't re-arm** (bad stored cron/timezone/args, lapsed license) — which change 1 now leaves enabled instead of disabling — with exponential back-off + one-time visibility, still without disabling.

Only the schedule's own faults still disable it — `QuotaExceeded` / `NotFound` — handled in `try_schedule_next_job` (`windmill-queue/src/jobs.rs`).

## Changes

**Transient push errors (`worker_flow.rs`, `worker.rs`)**
- Flow schedule-push path after retry exhaustion: branch on the error. `QuotaExceeded`/`NotFound` (reaching here only when `try_schedule_next_job`'s own disable write failed) → retry the disable. Transient → report only; the current occurrence runs to completion; schedule left enabled-but-unarmed → reconciler re-arms.
- Removed `SchedulePushZombieError` (type, impls, import + catch arm, stale doc). The earlier iteration returned it to force a zombie restart, but zombie detection **cancels** (not restarts) `same_worker` flows, so it would have lost the current run of a same-worker scheduled flow.

**Reconciler back-off + visibility (`monitor.rs`)**
- Track consecutive re-arm failures per schedule: exponential back-off (2, 4, 8, … passes, capped at 32) between retries instead of retrying every pass forever.
- After 3 consecutive failures, surface the cause once — record `schedule.error` (visible to the owner) + raise a critical alert — **without disabling**. Both reset (and the error is cleared) the moment the schedule re-arms.

**Tests**
- Removed `test_zombie_flow_after_schedule_push_failure_meets_restart_criteria` (its premise, a schedule-push-created zombie, no longer exists). No new unit test: the changed lines live in `handle_flow` / the reconciler loop (not reachable without heavy fixtures); the primitives are covered by existing `test_find_unarmed_schedules`, `test_rearm_schedule_*`, `test_failpoint_*`, `strike_unarmed_tests`. Validated instead by live e2e (below).

## Live e2e (CE dev server; these paths are all CE)

| Scenario | Result |
|---|---|
| Flow schedule, persistent bad-cron push | Not disabled; current run completes; reconciler backs off (failures 90s then 151s apart, not every ~30s pass) |
| …after 3 failures | `schedule.error` set + `CRITICAL ERROR: … enabled but has repeatedly failed to re-arm …`; still enabled |
| …cron fixed | Re-armed after cooldown; `schedule.error` cleared |
| Clean unarmed (deleted queued occ) | Reconciler re-arms |
| Script schedule, runnable deleted → `NotFound` | Disabled + error set (kept behavior) |
| **same_worker** flow, persistent push failure | Current occurrence completed `canceled_by=NULL` — not canceled (confirms the zombie-removal fix) |

## Behavioral notes

- On transient exhaustion the schedule's `error` is not populated immediately; it is set only if the reconciler then fails to re-arm it 3× (persistent cause).
- Back-off means passive auto-recovery of a long-broken schedule can lag up to ~32 passes after the cause is fixed; fixing via the UI/API re-arms immediately (that path pushes directly), so the lag only applies to out-of-band fixes.
- Reconciler back-off/surface state is per-replica in-memory (like the existing strike tally); at worst a leadership change re-alerts once.

## Test plan

- [x] `RUSTFLAGS=-D warnings cargo check -p windmill-worker --features quickjs`
- [x] `RUSTFLAGS=-D warnings cargo check -p windmill --features quickjs`
- [x] `cargo test -p windmill-queue --features failpoints --test schedule_push` — 48 pass
- [x] `cargo test -p windmill --bin windmill --features quickjs strike_unarmed` — 3 pass
- [x] Live e2e per the table above


rubenfiszel:
✅ Review round clean @ 7685ead2b79f8730342f3c017ccaca3d68ddd225

rubenfiszel:
## Live e2e validation (CE dev server; the changed paths are all CE, `enterprise`/`private` don't touch schedule push / the reconciler / `worker_flow.rs`)

Ran against a running server with `LISTEN_NEW_EVENTS_INTERVAL_SEC=1` (reconciler ~every 30s). Failpoints are task-local and can't fire in a live server, so failures were induced with real conditions (corrupted cron → persistent `BadRequest`; deleted runnable → `NotFound`; deleted queued occurrence → clean unarmed).

| # | Scenario | Result |
|---|---|---|
| A | Flow schedule, step-0 push fails persistently (`BadRequest: cron: Invalid pattern`) | **Not disabled**; current run completed (`success=true`); log fired `Leaving it enabled for the unarmed-schedule reconciler to re-arm`. After fixing the cron, the reconciler re-armed it (`re-armed enabled schedule …, which had no queued occurrence`). ✅ |
| D | Enabled schedule with its queued occurrence deleted (clean missed push, healthy runnable) | Reconciler re-armed within 2 strikes; schedule stayed enabled, no error. ✅ |
| C | Script schedule whose runnable is deleted → completion push hits `NotFound` | `Could not push next scheduled job … Not found … Disabling schedule` → `enabled=false`, error set, unarmed. Kept "schedule's own fault ⇒ disable" behavior. ✅ |
| E | **same_worker** flow schedule, step-0 push fails persistently | Current occurrence completed `success` with `canceled_by=NULL` — **not canceled**. Confirms the zombie-removal fix for Codex's original P1 (the old `SchedulePushZombieError` path would have canceled a same-worker flow via the `SameWorker` zombie branch). ✅ |

### Known limitation surfaced by scenario A (reconciler, pre-existing behavior widened by this PR)

For a schedule whose push fails **persistently with a non-`NotFound`/non-`QuotaExceeded` error** (e.g. a genuinely bad stored cron/timezone/args, or an instance-wide condition like an invalid license key), the reconciler retries `rearm_schedule` every pass **indefinitely** — no backoff, no give-up, and the repeated failure is only in server logs (not `schedule.error` or the workspace handler). Observed live:

```
20:56:28 ERROR schedule reconcile: could not re-arm schedule f/test/flow_sched_A: Bad request: cron: Invalid pattern …
20:56:58 ERROR schedule reconcile: could not re-arm schedule f/test/flow_sched_A: Bad request: cron: Invalid pattern …
20:57:29 ERROR schedule reconcile: could not re-arm schedule f/test/flow_sched_A: Bad request: cron: Invalid pattern …
```

It self-heals the moment the error clears (verified). It's a slow, bounded loop (≤50/pass, ~5 min in prod), never CPU-hot, and never wrongly disables — but before this PR such a flow schedule would have been disabled (visible in the UI with an error) instead of looping. Proposed follow-up (separate from this PR, since it's reconciler code from #10174): add per-schedule backoff on repeated rearm failures + surface it after N failures (critical alert / populate `schedule.error`) **without** disabling.

PR #10185, opened 2026-07-18
fix(csharp-parser): mark parameters with a default value as optional

winklemad:
C# `main` parameters with a default value (e.g. `Main(string s = "hi")`) were emitted with `has_default = false` even though the default was parsed, so the inferred schema marked them as **required** — the auto-generated form shows the field as mandatory, and strict schema validation rejects webhook/API calls that omit the optional argument.

This makes C# the only parser that could emit `default: Some(...)` and `has_default: false` on the same `Arg`, which is self-contradictory. The parsers for languages that do have default values derive the flag from the parsed default — `php`, `bash`, `yaml` and `java` all use `default.is_some()`; `go` and `rust` hardcode `false`, but they pair it with `default: None`, since neither language has default parameter values. C# parses the default and then ignored it. This derives it the same way the others do. Added assertions to the existing `test_parse_csharp_sig` pinning `default` + `has_default` for a defaulted parameter and two required ones.

---
## Summary by cubic
Marks C# Main parameters with a default value as optional to match other parsers. Previously the parser set has_default=false even when a default was parsed; now it derives has_default from the parsed default, so schemas/forms treat these args as optional and validation accepts them when omitted.

- In `parse_csharp_sig_meta`, set `has_default = default.is_some()` when building `Arg`.
- Extend `test_parse_csharp_sig` to assert `default` and `has_default` for one defaulted and two required parameters.

<sup>Written for commit b0745b30f684c937db6547aaf0a0f12c9cff5f44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


winklemad:

winklemad:
Bumping this one — it's still current and mergeable, and the CLA is signed.

Worth flagging that the red `cargo_test` isn't the change. It fails during setup, before anything is compiled, at the `actions/checkout` of `windmill-labs/windmill-ee-private`:

```
##[error]Input required and not supplied: token
```

That's the EE repo checkout not getting a token, which a PR from a fork can't be given. No cargo step ever runs in the log, so there's nothing on my side I can push to turn it green — it'd need a maintainer run.

The change itself is one line plus test assertions: `has_default` was hardcoded to `false` in `parse_csharp_sig_meta`, so `Main(string s = "hi")` produced `default: Some("hi")` alongside `has_default: false` on the same `Arg`. That combination is self-contradictory, and it's the only parser that can produce it — the languages that have default values (`php`, `bash`, `yaml`, `java`) all derive the flag with `default.is_some()`, and `go`/`rust` hardcode `false` only because they pair it with `default: None`. The effect is that a C# parameter with a default is treated as required, so the generated form marks it mandatory and schema validation rejects calls that omit it.

Happy to rebase if it's drifted, or to split the test assertions out if you'd rather review the one-line change on its own.

PR #10190, opened 2026-07-19
feat(pipelines): catalog declared DuckLake measures and dimensions

rubenfiszel:
## Summary

Lets a DuckDB script declare, next to the table it materializes, what that table's canonical aggregations are:

```sql
-- materialize ducklake://sales/orders
-- measure revenue = sum(amount) where not is_refund
-- measure order_count = count(*)
-- dimension region = region
SELECT ...
```

Those declarations are catalogued at deploy and read back by two consumers: the script editor (a drawer that composes a `SELECT` you can insert) and agents (a tool listing what is declared, so they stop guessing).

**This informs, it does not enforce.** Anyone can ignore a declared measure and write `sum(amount)` without the refund filter. The bet is that most wrong numbers come from not knowing the rule rather than defying it. Nothing rewrites or executes SQL: what you deploy is what runs.

It is also not a semantic layer. Declarations are single-table, with no join planner and no fan-out-safe aggregation across grains. Combining tables means materializing a mart and declaring measures on that.

> Scope note for anyone who reviewed this earlier: this PR previously carried a compiler, a `{{ metric() }}` token, a deploy-time SQL freeze and a worker substitution path. Those are gone. A compiler only buys byte-identical SQL with no LLM in the loop, plus enforcement, and both matter only once several independent consumers need the same number reproducible. It also broke "what you deploy is what runs": job logs showed SQL appearing nowhere in the source. Rationale is in `docs/pipeline-metrics-layer.md`.

## Changes

- **Parser**: `// measure <name> = <agg> [where <pred>]` and `// dimension <name> = <expr>`. The predicate is kept separate from the aggregate so a reader can render it as `expr FILTER (WHERE …)`, which is what lets measures with different predicates share one `GROUP BY`.
- **Catalog** (`data_metric`): one row per declaration, replaced wholesale per script path on every deploy, mirroring how the `asset` table is synced. Persisting rather than parsing on demand is what makes folder-scoped listing an index scan instead of a scan over every script body in the folder.
  - `idx_data_metric_folder` uses `text_pattern_ops`: this database's collation is not `C`, and measured at 50k rows the default opclass gives a Seq Scan while `text_pattern_ops` gives an index range scan.
  - Table paths are canonicalized to `<lake>/<schema>.<table>`. A producer's `// materialize ducklake://sales/orders` omits the schema while a consumer reading `sales.main.orders` records `sales/main.orders`; without this a consumer can never resolve the table it reads.
- **Endpoint** `GET /w/{workspace}/data_metrics/list`, optionally narrowed by `table` and/or `path_prefix`. Three separate controls, mirroring `list_macros`, the other script-derived catalog:
  - `data_metric` has no RLS, so reads are gated by an `EXISTS` against `script`, reusing its folder/group/user policies. The producing script path is part of the key because a DuckLake path has no folder to authorize against.
  - RLS reflects the user's permissions but not a token's path scopes, so rows are also filtered with `build_scope_path_predicate(&authed, "scripts", "read")`. `data_metrics` maps to the `Scripts` scope domain, without which a scoped token could not call the endpoint at all.
  - The `EXISTS` requires a non-archived, non-deleted producer, so a renamed or archived script stops serving its old declarations. The 1000-row cap is applied *after* scope filtering, so a restricted token cannot have its visible rows crowded out by rows it may not see.
- **Agent tool**: the same endpoint via `x-mcp-tool`.
- **Editor drawer**: a Metrics trigger always present for DuckDB scripts, mounted like the sibling pickers so it appears as a button in the wide editor bar and as an entry in the compact Helpers menu. Opening it browses the declared catalog and preselects a table the open script appears to reference.
  - The displayed and copied SQL is always a complete, runnable query that attaches the lake itself under the conventional `dl` alias, with a Copy button. Generated identifiers are always quoted, since quoting conditionally would mean tracking DuckDB's reserved-word set across versions and a miss emits invalid SQL.
  - Insertion adapts: when the script already attaches that lake it reuses that alias and omits the `ATTACH`, because repeating one fails (the worker's transform rewrites only the first match, leaving the duplicate as a literal `ducklake://` path). DuckDB names the catalog after the ATTACH alias, and the convention here is `AS dl`, so qualifying by lake name would not resolve.
  - Appended rather than inserted at the cursor, since the snippet is a whole statement block that must follow any `ATTACH`.
  - Tables that declare metrics are listed as visible choices with their measure/dimension counts, rather than hidden behind a dropdown, since browsing what else exists is the point.
  - **Try it** seeds an embedded `SqlRepl` against the lake and runs it in place, so a metric can be checked against real data before it goes anywhere. The REPL already wraps what it runs in `ATTACH 'ducklake://<lake>' AS dl; USE dl;`, which is where the `dl` convention comes from, so the seeded query carries no ATTACH of its own. `SqlRepl` gained a `setCode()` seam for this (not a `$bindable` default, which AGENTS.md bans).
- **Fork**: `clone_metric_catalog`, alongside the existing `clone_macro_registry`.

Deliberately not included: preview/run in the drawer, charts, filter builders. Each pulls back a preview job, a row cap and result rendering. Worth adding if inserting proves useful.

## Screenshots

Drawer with a filtered measure and a dimension selected. Every table that declares metrics is listed with its counts so you can switch between them; below that, the composed query, a REPL seeded from it, and real results. What is shown and copied is a complete, runnable query attaching the lake under `dl`; the note underneath says when insertion will adapt it to an attachment the script already has.

metrics drawer

A name filter appears once a table declares more than a handful of measures/dimensions (and a table filter once there are many tables), so large catalogs stay browsable. Filtering only narrows what is shown — an already-selected metric stays in the composed query, as the SQL here (channel + conversions) shows while the list is filtered to `seeded_metric_1`:

metrics drawer name filter

Below the editor-bar width threshold the buttons collapse into Helpers, where Metrics sits alongside Ducklake:

metrics in helpers menu

## Test plan

- [x] Deploy a producer with measures and dimensions, confirm `data_metric` rows appear; redeploy with one measure removed and confirm it disappears (wholesale replace).
- [x] `?table=`, `?table=ducklake://…`, `?path_prefix=f/metrics` and the unfiltered listing all return the expected rows; a literal `%` in `path_prefix` does not over-match.
- [x] A workspace member with no grant on the producing folder gets `[]` while an admin gets the rows.
- [x] A token scoped to `scripts:read:f/elsewhere/*` gets `[]`; one scoped to `scripts:read:f/metrics/*` gets the rows.
- [x] Archiving a producer stops its declarations being served (rows stay in the catalog, invisible until redeploy).
- [x] Drawer end to end in the browser: button present on DuckDB scripts (producer and consumer alike) and absent on a Python script, selection composes `sum(amount) FILTER (WHERE not is_refund)`, insert appends after the existing `ATTACH`.
- [x] Try it seeds the REPL and running it returns real rows from the lake (`eu 30`, `us 30`, refund row excluded by the FILTER).
- [x] A token scoped to a folder with no declarations gets `[]` even when authorized rows would sort after the first page (paged, not capped-then-filtered).
- [x] Against a script attaching `AS dl`, the composed SQL is `FROM "dl"."main"."orders"`, guarded by a unit test.
- [x] `cargo test -p windmill-common` (452) and the parser crates (236); `vitest` for `metricSql`; compiles under `--features mcp`.


rubenfiszel:
✅ Review round clean @ 8d17ea2631bd646e5785c0ad26c3df4cd2e9aef5 — nit-only verdicts (Codex "Mergeable", Claude + Pi "Good to merge") at d15dba6ecf; the one nit (doc over-promised a lock diff) addressed in 8d17ea2631.

rubenfiszel:
✅ Review round clean @ c900e0dd0fc56b2b41328fa79ffd8e6fa4c02d3a — nit-only verdicts at 171a845541 (Codex "Mergeable"); both nits addressed in c900e0dd0f (added the Data metrics sidebar link; fixed the stale `windmill_common::metrics` doc comment).

rubenfiszel:
Re: [P1] Apply token scope filtering before the candidate cap

Dismissing this one with reasoning rather than fixing it, on three grounds.

**It contradicts the previous round.** Round 2 flagged the opposite: "unfiltered list materializes the entire catalog before capping ... apply scope-aware limiting in SQL or stream". The 10k SQL ceiling was added to address exactly that. Without pushing scopes into SQL (no helper exists; every caller of `build_scope_path_predicate` filters in Rust) the two findings cannot both be satisfied.

**It matches the established pattern.** `list_scripts` in `windmill-api-scripts/src/scripts.rs:166` does the identical thing, with the identical constant:

```rust
let allowed = build_scope_path_predicate(&authed, "scripts", "read");
let rows = sqlx::query_as!(SearchScript, "SELECT path, content from script WHERE workspace_id = $1 AND archived = false LIMIT $2", &w_id, n)  // n = 10000
    .fetch_all(&mut *tx).await?.into_iter().filter(|r| allowed(&r.path))
```

Applied consistently this finding would flag that endpoint on main. I'd rather match the codebase than diverge for one new endpoint.

**The threshold is remote.** Triggering it needs >10,000 declarations visible to one workspace, i.e. roughly 2,000 pipelines each declaring five measures, *and* a path-scoped token whose authorized rows all sort after that boundary. Rows are ordered by `table_path`, so this is not the common shape.

If a workspace ever approaches that size the right fix is a SQL-side scope predicate shared by every list endpoint, not bespoke streaming here.

The scope filtering itself is verified end to end: unscoped admin sees all declarations, a token scoped to `scripts:read:f/elsewhere/*` gets `[]`, and one scoped to `scripts:read:f/metrics/*` gets exactly its own.

rubenfiszel:
✅ Review round clean @ 76569bf97eb5cb16385e9956cbeda8d495bba049 — Codex, Claude and Pi all "Good to merge", no outstanding findings or nits.

PR #10193, opened 2026-07-20
fix(security): enforce token scope filtering on folder list endpoints

rubenfiszel:
## Summary

Fixes WIN-2203.

The single-item folder endpoints (`get_folder`, `get_folder_usage`) enforce scoped-token restrictions via `check_scopes(&authed, || format!("folders:read:f/{name}"))`, but `list_folders` and `list_foldernames` did not — they relied on RLS alone and returned every folder visible to the underlying user. A token scoped to e.g. `folders:read:f/alpha` could therefore enumerate every folder's name, owners, `extra_perms` and summary in the workspace.

Both endpoints now filter their rows through `build_scope_path_predicate`, the same helper and the same idiom the sibling list endpoints use (`flows.rs:111`, `scripts.rs:166`, `variables.rs:218`, `resources.rs:218`, `apps.rs:393`).

Tokens without scope restrictions (session cookies, unscoped tokens, `if_jobs:filter_tags:`-only) are unaffected — the predicate is always-true for them. A broad `folders:read` token (what app-embed tokens carry; `folders/listnames` is the only folder route they may reach) still sees the full list, since a domain-level scope with no resource covers every path.

## Changes

- `backend/windmill-api-groups/src/folders.rs`: import `build_scope_path_predicate`; filter both list handlers on `f/{name}`.

No automated test ships with this change. During development an integration test confirmed the fix — with `alpha` and `zeta` in a workspace, a token scoped to `folders:read:f/zeta` saw only `zeta` on both endpoints, and the test failed on the unpatched handler with `left: ["alpha", "zeta"] right: ["zeta"]` — but it was dropped as too much scaffolding for a two-line filter.

## Test plan

Manually against the dev backend, workspace containing `alpha`, `beta`, `zeta`:

- [x] token scoped `folders:read:f/alpha` → `/folders/listnames` returns `["alpha"]` and `/folders/list` returns only `alpha` (both returned all three before)
- [x] same token → `folders/get/alpha` 200, `folders/get/beta` still `Permission denied: Required scope: folders:read:f/beta`
- [x] token scoped `folders:read:*` → all three folders
- [x] token scoped `folders:read` (app-embed shape) → all three folders
- [x] unscoped admin token → all three folders, pagination unchanged (`per_page=1` pages 1/2/3 give `alpha`, `beta`, `zeta`)
- [x] token scoped `scripts:read:*` → still rejected at the route layer

## Known limitation (pre-existing, repo-wide)

Like every sibling list endpoint, the filter runs after SQL `LIMIT/OFFSET`, so a scope-restricted token that explicitly paginates can get a short page when out-of-scope folders sort ahead of its own. `DEFAULT_PER_PAGE` is 1000 and no in-tree caller paginates folders, so this is latent rather than user-visible.

An earlier revision of this PR fixed it for folders with a keyset chunk-scan, but that added ~160 lines and an idiom unique to this file to solve a problem all six list endpoints share. It was reverted in favor of the minimal fix; the pagination-ordering issue is better addressed uniformly across all of them in its own PR (the reverted implementation is in the branch history at `48476be9aa` if it is wanted as a starting point).

Also noted by a reviewer, pre-existing and untouched: `exists_folder` has no scope check and uses a non-RLS connection, so any authenticated caller can probe folder existence.


rubenfiszel:
✅ Review round clean @ 48476be9aa6889b739ee183b04badb051ec75a2b

rubenfiszel:
✅ Review round clean @ faf5b934adbce95e65e0abf7bead7f674e6bccd3

PR #10195, opened 2026-07-20
[ee] fix(alerts): separate concurrency-gate backlogs from worker starvation

rubenfiszel:
EE companions: https://github.com/windmill-labs/windmill-ee-private/pull/671 (merged) and https://github.com/windmill-labs/windmill-ee-private/pull/673

## Problem

The `Jobs waiting in queue` critical alert counted every non-running job in `v2_job_queue` past the time threshold:

```sql
SELECT COUNT(*), MIN(scheduled_for) FROM v2_job_queue
WHERE tag = $1 AND scheduled_for <= NOW() - $2::interval AND running = false
```

That conflates two unrelated conditions: the worker fleet having no capacity, and a workspace's own concurrency limit holding its jobs back. Only the first is actionable by whoever gets paged. A single runnable with `concurrent_limit=1` pages repeatedly against a near-idle fleet, with a count that oscillates because the limiter's re-queue timestamps mature in bursts.

## Approach

The obvious fix — reconstruct the limiter's admission test in SQL and subtract the gated jobs — was tried and abandoned. That test spans the running counter, completions inside `concurrency_time_window_s`, per-version setting fallbacks, and several bypass paths, and nothing links the two copies. Successive review rounds each found a different divergence: a gate held shut by recent completions, an empty key, `DISABLE_CONCURRENCY_LIMIT`, soft-cancelled jobs, and legacy `v2_job` settings during a mixed-version rollout. Each was a real false alert or a suppressed one.

**The limiter now records what it did, and the alert reads that fact.** When `apply_concurrency_limit` re-queues a job it could not admit, it stamps `concurrency_gated_at` and `concurrency_gate_id`. The alert counts unmarked jobs. Every bypass leaves a job unmarked for free, because a bypassed job never reaches the re-queue path — no bypass enumeration exists in the alert to drift out of sync.

Three properties the mark needs, each found by review:

- **Marks expire.** A mark says a gate held the job when the limiter last looked, and only a worker re-attempting it refreshes that. Nothing clears it when the gate frees, so marks are trusted only while that gate is still parking jobs. Marks stop advancing exactly when nothing is pulling — the outage this alert exists for.
- **Liveness is per gate, and global.** `concurrency_gate_id` is the full admission policy (`key|limit|window`), not the key: one key may carry several policies, and only jobs judged by the same one share a gate. Keys are global, so gates are looked up across tags — a gate refreshed by another tag still holds this one's jobs.
- **Marks are version-gated.** Workers may lag the server. Until every worker stamps (`MIN_VERSION_SUPPORTS_CONCURRENCY_GATE_MARK`), the alert counts the whole backlog as before. The transition stays noisy rather than silent — a rolling upgrade keeps the pre-existing false alert instead of risking a suppressed outage.

Cancellation voids a mark: the pull path skips the limiter for cancelled jobs, so they are waiting on a worker whatever they still carry.

## Second alert: structurally undrainable gates

Excluding gate-parked jobs is only half the story. When jobs arrive faster than a gate admits them the backlog grows without bound and those jobs will **never** run — 63,210 queued against a ~17,280/day ceiling, growing +20,181/day, in the case that motivated this. With only the exclusion, that produces no alert at all, which is worse than the noise it replaced.

The axis that matters is not gated vs ungated but **drainable vs undrainable**. `concurrency_gate_alerts` fires per gate when the backlog cannot drain within 24h at the gate's own ceiling *and* is still growing; a large backlog that has stopped growing is slow rather than broken. Distinct dedup tag and resource, and it carries the workspace so it can route to the account owner rather than platform on-call.

```
Concurrency gate is oversubscribed and its backlog cannot drain.
Workspace: acme
Concurrency key: acme/u/bob/sync
Backlog: 63210 jobs (+20181 in the last 24h)
Gate admits at most 17280 jobs/day (limit 1 per 5s window)
Raise the limit, reduce the push rate, or pause what is scheduling these jobs.
```

Both alerts ship together deliberately: merging the exclusion alone would leave undrainable backlogs silent. EE half is windmill-ee-private#673, tests in `backend/windmill-common/tests/concurrency_gate_alerts.rs`.

Growth is inferred from a single sample rather than tracked across passes, so a gate whose arrivals pause for 24h reads as draining. Kept stateless deliberately, but it is the assumption most likely to need revisiting.

## Performance

Live gate ids resolve once per alert config, not per tag, and are skipped entirely when marks are not authoritative. Measured on a fixture shaped like the incident (65k backlog, 263k `concurrency_key` rows), queries extracted from the committed sqlx cache:

| | time | index | spill |
|---|---|---|---|
| gate liveness (once per pass) | 4.4 ms | yes | none |
| per-tag membership test | 8.8-12 ms | n/a | none |

N tags cost `4.4 + N x 9 ms`. The per-tag figure is at parity with the original query's 9-12 ms; the abandoned re-derivation reached 333 ms with temp spill.

The index is partial, covering only parked jobs so the ordinary queue write path stays unindexed, and the migration is registered in `OVERRIDDEN_MIGRATIONS` so it deploys as `CREATE INDEX CONCURRENTLY` executed top-level rather than inside the migration transaction.

## Tests

`backend/windmill-common/tests/jobs_waiting_alerts.rs`, gated on `private,enterprise`. Each was verified to isolate its defect — reverting only the corresponding fix fails that test alone:

- `concurrency_gated_backlog_does_not_alert` — a parked backlog raises nothing
- `jobs_waiting_on_capacity_still_alert` — unmarked jobs page, and the body reports the excluded count
- `stale_gate_marks_do_not_suppress_the_alert` — marks that stopped advancing are not believed
- `a_live_gate_does_not_vouch_for_another_gates_stale_marks` — gates on a tag expire independently
- `a_gate_refreshed_on_another_tag_still_holds` — gate liveness is global
- `canceled_jobs_are_not_held_by_a_live_gate` — cancellation voids the mark

## Not covered by tests

Stated plainly rather than implied: the producer-side stamp has no test (no existing test drives `apply_concurrency_limit`'s re-queue path, and a `PulledJob` fixture is substantial scaffolding — the invariant is pinned by a comment at the write site), and neither `MIN_VERSION_SUPPORTS_CONCURRENCY_GATE_MARK` nor `DISABLE_CONCURRENCY_LIMIT` is exercised, both being process-global state that leaks across a shared test binary.

Worth a human eye before merge: the mark is written by EE code and read by EE code, but the column and index live in an OSS migration, so a CE deployment carries both and never populates them.

## Disablable + frontend

The alert now has an off switch: instance settings > Alerts > "Alert on oversubscribed concurrency limits" (`critical_alerts_on_oversubscribed_gates`), gated at the monitor call site so it applies without a restart. Defaults on, since an oversubscribed gate is otherwise invisible. The alert body also names this path so whoever is paged can find it.

## Screenshots

Alerts settings, showing the new toggle enabled-by-default (the boolean Toggle now honors `defaultValue`, so it renders on when no row exists rather than off-while-the-backend-alerts):

gate-alert-toggle


rubenfiszel:
Thanks — the P1 is real and is now fixed in `9226131` (EE `6911fbf`).

**Confirmed against the limiter.** `update_concurrency_counter` admits on `completed_count + running_jobs < limit` (`windmill-queue/src/jobs_ee.rs:141` for the lock-free phase, `:189` for the locked admit), where `completed_count` is `concurrency_key` rows with `ended_at >= now() - concurrency_time_window_s`. The classifier only read `concurrency_counter.job_uuids`, so a gate held shut purely by recent completions read as free. With the shape that motivated this PR — a small limit and a short window — the counter empties on every completion, so the false alert simply returned one completion later.

**Fix.** The saturation test now mirrors the limiter: running jobs plus completions inside the window, with `concurrency_time_window_s` pulled through from `concurrency_settings`.

**Test.** `time_window_saturated_gate_does_not_alert` — `concurrent_limit=1`, nothing running, one completion inside the window, 200 queued jobs behind it. Verified it isolates this specific defect: reverting just the completed-window term fails that test alone while the other two keep passing.

```
thread 'time_window_saturated_gate_does_not_alert' panicked:
  a gate saturated by completions inside its time window must not raise a critical alert
test result: FAILED. 2 passed; 1 failed
```

**One performance note**, since the fix adds a second per-key probe. It initially regressed the classification from ~22 ms to ~275 ms on a 65k backlog with 263k `concurrency_key` rows. Cause was not the probe itself (0.007 ms) but CTE inlining: `gated` feeds three aggregate `FILTER`s, so both gate probes were evaluated six times. Marking `classified` as `MATERIALIZED` brings it to ~45-52 ms, and the comment records why the keyword is load-bearing. A `LATERAL` rewrite measured the same, so I kept the smaller diff. Steady state is untouched either way — the classification still sits behind the plain `COUNT(*)` pre-filter.

rubenfiszel:
Fixed in `7ac88b1` (EE `f01fc53`). Confirmed both bypass conditions at `windmill-queue/src/jobs_ee.rs:792` — `if *DISABLE_CONCURRENCY_LIMIT || job_concurrency_key.is_empty()` returns `(true, None)` without ever consulting the counter, so no job is parked behind a gate in either mode and excluding on one blinds the alert exactly when it matters.

**Empty key.** The classifier now requires `w.concurrency_key <> ''` before considering a gate saturated. NULL keys (jobs with no concurrency settings) are unaffected — they were already counted via the `gated IS NOT TRUE` three-valued-logic path.

**`DISABLE_CONCURRENCY_LIMIT`.** The classification is skipped wholesale and the raw depth is used. The flag is re-read from the environment in `windmill-common` rather than imported, since `windmill-queue` owns it and depends on this crate; the comment records that constraint so the two don't silently drift.

One thing your report surfaced indirectly: the bypass path alerts, so it needs a real oldest-job age. The pre-filter now carries `MIN(scheduled_for)` alongside the count, which would otherwise have reported "0 minutes" on that path.

**Test.** `bypassed_empty_key_still_alerts` — empty key, gate seeded as saturated, 200 overdue jobs, expects the alert to fire. Verified it isolates this defect: reverting only the `<> ''` guard fails that test alone.

```
assertion failed: jobs whose concurrency key is empty bypass the limiter and must still alert, got []
test result: FAILED. 3 passed; 1 failed
```

Four tests now, all green, offline check clean. `DISABLE_CONCURRENCY_LIMIT` itself is not covered by a test — it is process-global env state and setting it would leak across the shared-process test binary; the empty-key case covers the same classifier branch deterministically, which is what you suggested.

rubenfiszel:
Fixed in `db9f9fd` (EE `d54657f`). Confirmed at `windmill-queue/src/jobs.rs:3961` — `pulled_job.canceled_by.is_some()` returns the job immediately without consulting the counter, so a cancelled job is waiting for a worker, not for a gate. Reading it off key saturation let a saturated key hide an unbounded cancelled backlog during precisely the outage this alert exists to catch.

**Fix.** Cancelled jobs are excluded from the classification (`AND q.canceled_by IS NULL`) and counted off the pre-filter instead, then folded back into the capacity count. `MIN(scheduled_for)` is taken across both sets so the oldest-job age stays honest when the oldest waiter is a cancelled job.

**Test.** `canceled_jobs_behind_saturated_gate_still_alert` — saturated gate, 200 queued jobs with `canceled_by` set, expects the alert to fire. Verified it isolates the defect: reverting only the cancellation guard fails that test alone.

**Worth flagging: my first attempt at this fix was a performance regression**, which is why the diff routes cancellation through the pre-filter rather than the obvious place. Grouping the classification on `canceled_by` drags that `varchar(255)` through every join beneath the aggregate — projected row width went 33 → 548 bytes and the whole scan spilled to temp files:

| version | 65k backlog, 263k `concurrency_key` rows | temp spill |
|---|---|---|
| grouped on `canceled_by` | 267-285 ms | yes |
| counted off the pre-filter | 67-71 ms | none |

The pre-filter is a plain scan with no joins, so the two added `FILTER` aggregates are nearly free (9.4-11.4 ms, unchanged). Numbers are from `EXPLAIN ANALYZE` on the exact SQL extracted from the committed sqlx cache, not a hand-transcribed copy.

Five tests, all green; offline check clean; zero EE cache entries lost.

For what it's worth, this is the third limiter bypass you've caught that I missed — I have now walked all four admission paths in `apply_concurrency_limit` / `pull` (`DISABLE_CONCURRENCY_LIMIT`, empty key, `canceled_by`, and the counter+window test itself) and each has a test pinning it.

rubenfiszel:
Reworked the approach rather than patching these two. Pushed `d6a66ca` (EE `3984a0f`).

Both findings were correct, and together with the previous three they were all the same failure: the alert reconstructed the limiter's admission test in SQL, and nothing linked the two copies. That test spans the running counter, completions inside the time window, per-version setting fallbacks (`maybe_fallback`), and three bypass paths — each round found another divergence, and the query had grown to 333ms with a `v2_job` join to chase the legacy fallback.

**The alert now reads a fact instead of re-deriving one.** New nullable column `v2_job_queue.concurrency_gated`, set by the limiter in the single place it re-queues a job it could not admit (`jobs_ee.rs`). The alert counts unmarked jobs.

This makes the whole class of finding structurally impossible: a bypassed job never reaches the re-queue path, so it is never marked, with no bypass enumeration anywhere in the alert. Concretely, for the two P1s here —

- **Legacy `v2_job` settings.** The limiter reads the limit via `maybe_fallback` before deciding, so a mixed-version job that gets parked is marked like any other. No `v2_job` join, no version awareness in the alert.
- **Cancelled agent jobs.** These are now correct *without* touching the pull path, which is what made this one awkward. A cancelled job on the `query_o` branch that really is re-queued gets marked and excluded; a cancelled job on the normal branch bypasses, is never marked, and stays counted. The mark reflects what happened rather than what the alert predicted, so the two branches disagreeing no longer matters.

**Performance**, same fixture (65k backlog, 263k `concurrency_key` rows), query extracted from the committed sqlx cache:

| | time | spill |
|---|---|---|
| re-derived classification (previous head) | 333 ms | none |
| gate mark | 11 ms | none |

That is back to the cost of the original broken query. The `COUNT(*)` pre-filter is gone too — it existed only to keep the expensive classification off the common path.

**Tests** shrank to two, since the elaborate gate fixtures no longer describe anything the alert reads: gated backlog does not alert, unmarked backlog does and reports the excluded count. Migration verified up/revert/up.

**One gap I want to flag rather than bury:** correctness now rests on the limiter setting the flag, and no test covers that write — no existing test exercises `apply_concurrency_limit`'s re-queue path, and constructing a `PulledJob` to do it would be substantial fixture scaffolding. I pinned the invariant with a comment at the write site instead ("only this path may set it"). If you would rather have that guarded by a test, say so and I will build the fixture.

rubenfiszel:
Both addressed in `a8199a8` (EE `a1322ea`).

**[P1] Stale marks.** Correct, and it was the load-bearing flaw in the redesign: the mark recorded the last admission decision, and only a worker re-attempting the job refreshed it. Nothing cleared it when the gate freed, so a subsequent worker outage left the backlog starved but still marked — silent forever.

The fix turns on an equivalence worth stating: **stale marks and worker death are the same condition.** If a gate frees while workers are alive, a worker admits from that key and the backlog drains; marks only stop advancing when nothing is pulling the tag. So the limiter's own write activity is the liveness signal, and no external metric is needed.

`concurrency_gated_at TIMESTAMPTZ` replaces the boolean. Marks are trusted only while the newest mark on the tag is within `alert_time_threshold_seconds`; once they stop advancing every job counts and the alert fires. One subtlety that took a second pass: the freshness probe must span the whole tag rather than the overdue rows, because a job only becomes overdue *after* its re-queue delay elapses, so its own mark is always stale by then. That is why the scan drops the `overdue` predicate into a projected column instead of the WHERE clause.

**Test.** `stale_gate_marks_do_not_suppress_the_alert` — 200 jobs marked an hour ago and never refreshed, asserting the alert fires and counts all 200. Verified it isolates the defect: reverting only the freshness check fails that test alone while the other two pass.

You also asked for a producer-side test that marks jobs through real gate rejection. I did not add one: no existing test exercises `apply_concurrency_limit`'s re-queue path and constructing a `PulledJob` is substantial fixture scaffolding. The three tests pin the consumer contract, and the invariant at the producer is pinned by a comment at the write site. Happy to build the fixture if you would rather have it.

**[P2] Comments.** Condensed all four sites to within four lines and removed the drafting-history phrasing ("used to model", "get wrong") from the emitter comment — that belonged in this PR description, not the source. Checked the changed hunks for other blocks over the limit; the long ones remaining in `ee.rs` are pre-existing and untouched.

**Performance** is unaffected by the timestamp change — 8.5-9.5 ms on the 65k fixture, no spill, measured on the query extracted from the committed cache. Migration re-verified up/revert/up after the column type change.

rubenfiszel:
Fixed in `d5dc097` (EE `3dc0048`). Correct — gate state is per key, and my freshness probe was tag-wide, so a busy key vouched for stale marks left by a freed one beside it.

Worth noting why per-job freshness cannot substitute: a job only enters the overdue set *after* its re-queue delay elapses, so by construction nothing has re-attempted it since, whether its gate is saturated or freed. The two cases are indistinguishable per job, which is what forces per-key state.

**Fix.** The limiter now stamps `concurrency_gated_key` alongside the timestamp, and liveness is derived per key: a gate is live if its newest mark is within `alert_time_threshold_seconds`. Jobs under a live gate are excluded; jobs under a gate that stopped being refreshed are counted, independently of what other gates on the tag are doing.

**Test.** `a_live_gate_does_not_vouch_for_another_gates_stale_marks` — 200 jobs under a gate marked 1s ago, 150 under a gate marked an hour ago, same tag. Expects one alert counting exactly the 150. Verified it isolates the defect: swapping the per-key `MAX` back to a tag-wide one fails that test alone.

**Two query shapes rejected on measurement**, both correct but spilling to temp files on the 65k fixture, because they carry the varchar key through a materialized 65k-row set:

| shape | time | spill |
|---|---|---|
| reused CTE + self-join | 31 ms | yes |
| single pass, window `PARTITION BY` key | 27-34 ms | yes |
| two scans, per-key aggregate collapsed first | 25-28 ms | none |

The shipped form scans the queue twice: the `live_gates` aggregate collapses to one row per key before anything joins against it, so the wide set is never materialized. That is the reason for the apparent duplication in the SQL, and it is why the previous single-scan shape is not simply restorable.

Cost is 25-28 ms against 8.5 ms for the tag-wide version that had this bug — the honest price of per-key correctness, and still an order of magnitude below the 333 ms re-derivation this replaced. Migration re-verified up/revert/up with both columns.

rubenfiszel:
All three fixed in `a4e31b1` (EE `5e449c7`).

**[P1] Old workers leave gated jobs unmarked.** Correct, and the one that mattered most: only the new limiter stamps, and direct-DB workers are explicitly allowed to lag. Added `MIN_VERSION_SUPPORTS_CONCURRENCY_GATE_MARK` (1.764.0) and gated the exclusion on it. Until every worker stamps, the alert counts the whole backlog exactly as it did before marks existed.

That deliberately makes the transitional behaviour *noisy rather than silent*: a rolling upgrade keeps the pre-existing false alert for its duration, which is the status quo, instead of risking a suppressed outage. No compatibility path re-deriving the admission test — the fallback is the plain count, so none of the bypass enumeration this rework deleted comes back.

**[P1] Tag-local freshness misses global gates.** Correct — keys are global and can group jobs across tags and workspaces. `live_gates` now looks up gates across the queue rather than within the monitored tag. Cost is unchanged (22-25 ms vs 25-28 ms) because the per-gate aggregate collapses to one row per gate before anything joins it, so widening the scan does not widen what is materialized.

**[P1] Key-only freshness conflates policies.** Correct. The stamped column is now `concurrency_gate_id` and holds the full admission identity (`key|limit|window`) rather than the key, so two policies sharing a key are two gates and neither vouches for the other. This is why the column is a gate id and not a key — the name change is load-bearing.

**Tests.** Added `a_gate_refreshed_on_another_tag_still_holds`, verified to isolate the scope fix: re-adding `tag = $1` to `live_gates` fails that test alone. Policy separation needs no separate test — differing policies produce different gate ids, which is exactly what the existing two-gate test exercises. Five tests green, migration re-verified up/revert/up.

**Two gaps I am not covering with tests, deliberately.** `MIN_VERSION_SUPPORTS_CONCURRENCY_GATE_MARK` and `DISABLE_CONCURRENCY_LIMIT` are both process-global state that leaks across a shared test binary, and the producer-side stamp still has no test because no existing test drives `apply_concurrency_limit`'s re-queue path. Those three are the untested surfaces in this PR and I would rather name them than imply coverage I do not have.

rubenfiszel:
Both fixed in `e361309` (EE `d3445d0`). Both were self-inflicted by last round's globalization, and the P1 is a fair catch — I widened the scan for correctness without noticing it sat inside the per-tag loop.

**[P1] Repeated whole-queue aggregation.** Live gate ids are now resolved **once per alert config** instead of once per tag, and the query is skipped entirely when `DISABLE_CONCURRENCY_LIMIT` is set or `MIN_VERSION_SUPPORTS_CONCURRENCY_GATE_MARK` is unmet — you were right that those checks happened after the query, so a mixed-version fleet paid for a result it discarded. The per-tag query reduces to a membership test against that set.

An empty set now naturally means "nothing is gated", so the fallback paths need no special-casing in Rust: `DISABLE_CONCURRENCY_LIMIT` and the version gate both just yield an empty list and every job counts as waiting on a worker. That deleted a branch rather than adding one.

Also added the missing index you noted — partial, over parked jobs only, so the ordinary queue write path stays unindexed:

```sql
CREATE INDEX v2_job_queue_concurrency_gate
    ON v2_job_queue (concurrency_gate_id, concurrency_gated_at DESC)
    WHERE running = false AND concurrency_gate_id IS NOT NULL;
```

Measured on the 65k fixture, queries extracted from the committed cache:

| | time | index used | spill |
|---|---|---|---|
| gate liveness (once per pass) | 4.4 ms | yes | none |
| per-tag membership test | 8.8-12 ms | n/a | none |

So N tags cost `4.4 + N x 9 ms` against `N x 23 ms` before, and zero when marks are not authoritative. The per-tag figure is back to parity with the original (broken) query's 9-12 ms.

**[P2] Producer comment.** Condensed to four lines, keeping both constraints (only this path stamps; the id carries the whole policy).

Five tests still green, migration re-verified up/revert/up with the index, zero EE cache entries lost.

rubenfiszel:
Both fixed in `a24bcdc` (EE `e1e73f4`).

**[P1] Cancellation leaves a stale mark.** Correct, and it is a regression I introduced: the pre-mark version of this PR filtered cancelled jobs out explicitly, and that handling did not survive the rewrite. A cancelled job keeps whatever mark it carried, the pull path skips the limiter for it, and if its gate stayed live elsewhere it stayed excluded while squarely waiting for a worker. A cancelled job is now ungated regardless of its mark.

**[P1] Index migration blocks the queue.** Correct, and the more serious of the two — this would have stalled pulls and enqueues on every deployment against a real queue. Registered in `OVERRIDDEN_MIGRATIONS` with the `CREATE INDEX` -> `CREATE INDEX CONCURRENTLY` rewrite, matching the other `v2_job_queue` index migrations, so `CustomMigrator` splits it and runs each statement top-level rather than in the migration transaction.

I verified the rewritten path rather than assuming it: applying the override's exact transformation statement-by-statement against a scratch database succeeds for all three, and the resulting index is the intended partial one.

```
OK   ALTER TABLE v2_job_queue ADD COLUMN IF NOT EXISTS concurrency_gated_at
OK   ALTER TABLE v2_job_queue ADD COLUMN IF NOT EXISTS concurrency_gate_id
OK   CREATE INDEX CONCURRENTLY IF NOT EXISTS v2_job_queue_concurrency_gate
-> ... USING btree (concurrency_gate_id, concurrency_gated_at DESC)
   WHERE ((running = false) AND (concurrency_gate_id IS NOT NULL))
```

The plain `CREATE INDEX` stays in the migration file so `cargo sqlx migrate run` still works in dev; only the deployed path is rewritten, which is the existing convention.

**Test.** `canceled_jobs_are_not_held_by_a_live_gate` — 200 jobs marked under a live gate, then cancelled, expecting the alert to fire. Verified it isolates the defect: dropping the `canceled_by IS NULL` conjunct fails that test alone. Six tests green, zero EE cache entries lost.

rubenfiszel:
✅ Review round clean @ a24bcdc03eb2c5aed0506736c120b9ee74c22275

rubenfiszel:
Reopened as draft: this now also adds the second alert (windmill-ee-private#673) for structurally undrainable gates. Merging the exclusion on its own would leave that case silent, so the two ship together. Re-reviewing.

rubenfiszel:
All three fixed in `02b216a` (EE `a73c310`).

**[P1] Invalid marks counted.** Right — I built the second alert on `concurrency_gate_id IS NOT NULL` and skipped the two qualifications the queue-depth classifier had already learned. Now counts only uncancelled jobs under a gate still being refreshed, matching its sibling.

**[P1] `arrivals > 0` is not net growth.** The sharpest of the three, and I had this wrong twice over: I shipped it, then described it in the PR body as a known limitation when it was actually a false-positive condition. Your counterexample is exact — a 40k backlog draining at 17,280/day with one arrival/day is *shrinking* by ~17,279/day and still paged.

The fix follows from what the number actually means. The gate admits oldest-first, so jobs from the last 24h still queued stand in for arrivals over that day; oversubscription is then literally `arrivals > ceiling_per_day`, which is the claim being made rather than a proxy for it. In the case that motivated this: 33,637 > 17,280.

**[P1] Drained gates never recover.** Right — a gate that empties drops out of the aggregation, so per-gate recovery could never fire for it. Recovery is now swept against the gates observed this pass, which also subsumes the present-but-no-longer-oversubscribed case and removed a branch.

**Tests**, per your list: `shrinking_backlog_does_not_alert`, `cancelled_and_stale_marks_are_not_counted`, `a_drained_gate_recovers`. Verified the growth one isolates its defect — reverting only that condition to `> 0` fails it alone while the other four pass. Also took the fixture nit: the 40k-row inserts are now one set-returning statement, which cut the suite from 12s to 1.3s.

11 tests green across both alerts, offline builds clean, zero cache entries lost.

rubenfiszel:
Both fixed in `52bd4f54f4` (EE `8df3339`).

**[P1] Cancelled marks keeping a gate live.** Right, and it is the same inconsistency twice: I filtered cancelled rows out of the per-job classifier and out of the oversubscription query, but left the `live_gates` aggregate reading them. One freshly cancelled job -- which no gate holds -- vouched for every stale mark on that gate.

**[P1] Recovery routed globally.** Right. The alert is raised with the workspace and recovery passed `None`, so on cloud it would leave the workspace-scoped path. Recovery now reads the workspace back off the alert being resolved, which is necessary because a fully drained gate is gone from the queue by then.

**A test I wrote for this was ineffective, and I only caught it by checking.** `a_cancelled_job_does_not_keep_its_gate_live` passed with the fix reverted. The cause was the fixture, not the alert: `queue_canceled_jobs` cancelled *every* uncancelled job on the tag rather than the rows it had just inserted, so the stale backlog the test depends on was cancelled too and the alert fired for the wrong reason. Fixed the helper to set `canceled_by` at insert; the test now fails without the fix and passes with it.

Both regressions verified to isolate their defects individually: reverting only the liveness filter fails the cancellation test alone, reverting only the workspace lookup fails the recovery test alone.

12 tests green across both alerts, offline builds clean, zero cache entries lost.

rubenfiszel:
✅ Review round clean @ 52bd4f54f45b32fa2047acd8a8e93dd09aabb87e — nit-only verdicts at 52bd4f54f45b32fa2047acd8a8e93dd09aabb87e (Claude and Pi "Good to merge", Codex "Mergeable, but should ideally address nits"). The single P2 was the missing `[ee]` title prefix, now restored (it had been dropped from the title at some point after creation); the retitle is metadata only, so the head is unchanged from the reviewed commit.

rubenfiszel:
Rebased onto main (resolved import-list conflicts in main.rs/monitor.rs from #10197's CONCURRENCY_KEY_MAX_QUEUED), and ee-repo-ref now points at #674's squash on EE main (e5c26f6). Re-reviewing.

rubenfiszel:
All addressed (OSS `1dcd6cd`, EE `db65192` on new PR windmill-ee-private#675).

**[P1] Measure arrival at the gate, not creation.** Correct and important — another false-negative, the class this PR exists to kill. A job pushed with a future `scheduled_for` reaches the gate only once due, so `created_at` hid its arrival and a burst of long-scheduled jobs coming due stayed silent. Added `concurrency_first_gated_at`, stamped once by the limiter via `COALESCE(concurrency_first_gated_at, now())` so re-queues never move it, and arrivals are now counted from it. Test `jobs_delayed_before_reaching_the_gate_still_alert` (created 3 days ago, first gated this hour, still arriving above ceiling); verified it isolates the defect — reverting to `created_at` fails it alone.

**[P2] Toggle default.** Correct, and the Playwright check earned its keep: my first fix (extending the `defaultValue` effect to booleans) rendered the toggle **unchecked** — I verified it in the browser and it was wrong. The working fix defaults at the Toggle binding itself with a getter/setter, no effect timing involved. Confirmed enabled-by-default in the browser:

gate-alert-toggle

**[P2] Frontend verification.** Done — the setting is `critical_alerts_on_oversubscribed_gates` in the Alerts group, gated at the monitor call site so it applies without a restart, defaulting on. PR body updated with the screenshot and the "no frontend changes" line removed.

**Nits (Claude/Pi):** removed the orphaned sqlx cache entry left by the rewritten test fixture, and reverted cargo fmt churn in three unrelated test files.

14 tests green across both alert suites, offline build clean, zero cache entries lost vs main.

rubenfiszel:
All addressed (OSS `a28c465`, EE `2aa9105`).

**[P1] Trusted-caller contract.** Documented on `concurrency_gate_alerts` that it is instance-wide, runs on an unrestricted `DB`, and is called only from the instance monitor loop, never a request handler. Worth noting the sibling alerts in that module (`jobs_waiting_alerts`, `low_disk_alerts`, `worker_groups_alerts`) carry no such doc, so this now sets the contract they omit rather than matching an existing convention.

**[P2] Four orphaned cache entries.** Removed. They cached the pre-`concurrency_first_gated_at` limiter UPDATE and three superseded fixture inserts; verified each was the old form (no `concurrency_first_gated_at`) and that an offline rebuild of the libs and both test targets reports zero missing after removal.

**[P2] `[ee]` title prefix.** Restored.

**Nits:** migration column comment trimmed to four lines; the stray prettier reflow in `InstanceSetting.svelte` reverted (the file on main is not prettier-clean and no CI check enforces it, so a format-on-save hook had reflowed an unrelated `<p>` — the diff is now only the toggle change); `ee-repo-ref.txt` given its trailing newline.

All housekeeping this round, no behaviour change. 14 tests green, offline build clean, zero cache entries lost.

PR #10197, opened 2026-07-20
feat: cap queued jobs per concurrency key on cloud

rubenfiszel:
## Summary

A concurrency-limited key drains at most `concurrent_limit` jobs per window no matter how idle the fleet is. A producer pushing faster than that grows a backlog that spare worker capacity cannot absorb: the jobs are serialized behind the limit, not waiting on workers. Left alone this reaches tens of thousands of queued jobs on a single key, and purging it only helps until the producer refills it.

This adds a `CLOUD_HOSTED`-only ceiling on how many jobs may be *waiting* behind one concurrency key. Default 10000, configurable, `0` disables. At normal queue depths it never fires.

Scoped per concurrency key rather than per workspace on purpose: the pathology is one key, and a workspace-wide cap would take that workspace's unrelated billing, sync and alerting scripts down with the runaway.

## Changes

- `concurrency_key_max_queued_jobs` global setting (default 10000, `0` disables), plumbed through `initial_load` + the `process_notify_event` hot-reload path like the `workspace_fairness_*` knobs, and registered in instance settings under **Jobs**.
- `concurrency_key_queue_depth` in `windmill-queue/src/jobs.rs` counts waiting jobs for a key; `check_concurrency_key_queue_cap` rejects the push with a `QuotaExceeded` naming the key, the depth and the cap.
- Enforced at both sites that register a concurrency key: inline in `push`, and via `insert_concurrency_key_capped` from `worker_flow`. A flow with a preprocessor is pushed with `concurrent_limit` cleared and only registers its key later from `worker_flow`, so enforcing in `push` alone would have left trigger-driven flows — the shape most likely to run away — uncapped.

- `/queue/import` is rejected on `CLOUD_HOSTED`. It writes queue rows straight to the DB, bypassing the push path and every admission check that lives there — not just this cap. It is a self-hosted migration/restore tool and needs `require_admin` only, so it was reachable by a workspace admin. `add_batch_jobs` writes keys directly too but stays exempt: it is superadmin-only, and a superadmin can set the cap to `0` regardless. Both exemptions are now named in the `check_concurrency_key_queue_cap` contract, which previously overclaimed enforcement at "every site".

Three properties of the counting query are load-bearing and are commented as such at the query:

- **Future-scheduled jobs count.** The limiter re-queues blocked jobs by pushing `scheduled_for` forward one window at a time, so a concurrency-gated backlog is almost entirely future-dated. A cap counting only due jobs would sit near zero and never fire.
- **Running jobs do not count.** They stay in `v2_job_queue` with `ended_at` NULL while executing. Counting them would charge a key for the concurrency it is licensed to use — with `concurrent_limit` above the cap, every push would be rejected with nothing waiting.
- **The `LIMIT` bounds rows scanned, not rows that qualify.** `concurrency_key` rows are only cleared by `add_completed_job`, so a job that left the queue another way (mass cancel, manual purge) leaves an `ended_at IS NULL` row that nothing reclaims — the retention sweep deletes on `ended_at <= …`, which never matches NULL. Those rows are excluded from the count so they cannot wedge a key above the cap with an empty queue, and the `LIMIT` sits inside that filter so they cannot make every push scan them either.

## Measurements

`EXPLAIN (ANALYZE, BUFFERS)` on a seeded 70k-job backlog:

| Case | Result |
|---|---|
| Normal shallow key (the hot path, every push) | **0.073 ms**, 22 buffers |
| Saturated key at the 10000 cap | ~9.6 ms — only on a push being rejected |
| Key with 50k stale rows | scans **10000** rows (the bound); the filter-outside-`LIMIT` shape scanned all **50000** |

## Known limitation

The check and the insert are not atomic, so concurrent pushes can each observe `depth == cap - 1` and all commit. Overshoot is bounded by push concurrency against a cap of 10000 and self-corrects on the next push. Serializing per key would put a lock on every push — the cost the design is specifically avoiding for a backstop that only ever fires on a runaway.

## Test plan

Regression tests in `backend/windmill-queue/tests/concurrency_key_queue_depth_test.rs`, each verified to fail when its property is removed:

- [x] `counts_future_scheduled_jobs` — narrowing the query to `scheduled_for <= now()` fails it (`left: 3, right: 10`)
- [x] `ignores_running_jobs` — dropping `running = false` fails it (`left: 8, right: 2`)
- [x] `ignores_rows_whose_job_left_the_queue` — orphaned rows must not count

Manual, against a build with `--features quickjs,enterprise,private,cloud` and `CLOUD_HOSTED=true`:

- [x] Cap fires: with cap 3, pushes 1-3 accepted, 4-6 rejected with a message naming key, depth and cap
- [x] **Counts future-scheduled backlog**: with the queue at `due_now=0, future=1` and cap 1, the push is rejected — a due-only count would see 0 and admit
- [x] Per-key, not per-workspace: one script capped while another in the same workspace kept accepting on its own counter
- [x] Orphan rows do not wedge a key: 6 stale `ended_at IS NULL` rows + empty queue → admitted
- [x] `0` disables; clearing the setting falls back to 10000; changes hot-reload without a restart
- [x] `/queue/import` rejected on cloud, verified with one binary and only the env var changed: `CLOUD_HOSTED=true` → `400 Importing queued jobs is not available on the cloud`; unset → `200 Successfully imported jobs` (self-hosted restore unaffected)
- [x] `cargo check` (CE) and `cargo check --features enterprise,private,cloud` clean; `npm run check:fast` clean (one pre-existing unrelated `modern-screenshot` error, present on `main`)

## Screenshots

The new setting in **Instance settings → Jobs**:

concurrency-cap-setting

The field is `cloudonly`, and `isCloudHosted()` is a hostname check against `app.windmill.dev`, so it does not render on a local instance regardless of the backend's `CLOUD_HOSTED` setting. The capture was taken with that visibility gate temporarily flipped locally; the committed code keeps `cloudonly: true`. Scoped to the field element rather than the settings page, which shows other instance config.


rubenfiszel:
✅ Review round clean @ 6e1f24abc7f5b042355d65d26e1964bb79e7fbd5

All three reviewers returned "Good to merge" with no outstanding findings.

Codex suggested additionally verifying the preprocessed-flow path with a custom key but no active concurrency limit. Verified directly against `insert_concurrency_key_capped` on a saturated key with `CLOUD_HOSTED=true`: `concurrent_limit: Some(1)` is rejected, `concurrent_limit: None` is admitted. The scaffolding was throwaway and is not committed — the guard is a single `concurrent_limit.is_some()` condition and a committed test for it would need the full flow-preprocessor fixture to assert one branch, which AGENTS.md classes as scaffolding.

rubenfiszel:
✅ Review round clean @ 09244c86ffd31a1cc3c47fa78ba2db6ab501c2c8

Codex, Claude and Pi all returned "Good to merge" with no outstanding findings.

Codex asked to manually confirm `/queue/import` is rejected on cloud and still available self-hosted. Verified against one binary with only the env var changed:
- `CLOUD_HOSTED=true` → `400 Bad request: Importing queued jobs is not available on the cloud` (rejected before the insert loop, no writes)
- `CLOUD_HOSTED` unset → `200 Successfully imported jobs`

That also confirms it is a runtime gate rather than a compile-time one, per backend/CLAUDE.md.

PR #10201, opened 2026-07-20
feat: detect server-handled git-sync so CLI picks git push vs wmill sync push

hugocasa:
## Summary
Since git-sync push can now be handled server-side (the workspace auto-pulls the
remote and deploys), the CLI could no longer tell whether pushing to git deploys —
its only signal was the presence of a CI workflow file, which server-handled sync
doesn't need. The data that answers this lives in `git_sync` workspace settings,
readable only by admins via `get_settings`, so a regular member had no way to know.

This adds a non-admin endpoint that surfaces just the deploy-relevant signal, a CLI
command to consume it, and aligns the deploy guidance/skills to prefer `git push`
when the repo deploys on push and fall back to `wmill sync push` otherwise.

## Changes
- **Backend**: new non-admin `GET /w/{workspace}/workspaces/git_sync_deploy_mode`
  returning `{ configured, deploy_on_push, auto_pull_branches }`. `deploy_on_push`
  is true only on an Enterprise-licensed instance with an auto-pull repository
  (auto-pull cannot run on CE/downgrade), and `auto_pull_branches` lists the tracked
  branches so a client can tell whether the current checkout actually deploys on push.
  Exposes only derived values — no repository paths, credentials, or webhook config.
- **CLI**: `wmill gitsync-settings status` reports the deploy mode and recommends a
  command, matching the local git branch against `auto_pull_branches` before
  suggesting `git push` (an untracked branch falls back to `wmill sync push`).
- **Guidance/skills**: the deploy sections in `cli/src/guidance/core.ts` and the
  `system_prompts` sources now describe the two-path model (backend auto-pull or CI
  → `git push`; otherwise `wmill sync push`) and point at `gitsync-settings status`.
  Regenerated `system_prompts/auto-generated/` and `skills.gen.ts`.
- **Tests**: integration assertion for the endpoint's default (no git-sync) response.

## Test plan
- [x] `cargo check -p windmill-api-workspaces` (clean)
- [x] `cargo test -p windmill-api-integration-tests --test workspaces test_workspace_endpoints` (passes)
- [x] `bunx tsc --noEmit` — no errors in changed CLI files
- [x] Live: endpoint returns `{configured:false, deploy_on_push:false, auto_pull_branches:[]}` (HTTP 200) on a workspace with no git-sync
- [ ] On an EE instance with an auto-pull repo, `git push` is recommended only for the tracked branch; other branches recommend `wmill sync push`
- [ ] Verify a member (non-admin) can read the endpoint

---
## Summary by cubic
Detects server-handled git sync so the CLI recommends git push only when a push actually deploys; otherwise it reports “undetermined” and points to Deploying guidance. Now excludes archived roots and treats false as “not confirmed” to avoid wrong advice.

- **New Features**
  - Backend: add GET /w/{workspace}/workspaces/git_sync_deploy_mode?branch=... returning {configured, deploy_on_push}. deploy_on_push is true only on Enterprise and only when there’s a runnable delivery path; resolves $var:/$res:; shares fork routing with the auto-pull reconciler; reads fresh config; excludes archived roots; never leaks repo details; documents false as “not confirmed”.
  - CLI: add `wmill gitsync-settings status` (supports `--json-output`). Uses the current git branch; no remote URL sent. Recommends `git push` only on a backend match; otherwise defers (deploy_command=null in JSON) and points to Deploying guidance.

- **Refactors**
  - Share one repo resolver and fork-branch routing (`resolve_fork_branch_target`) between the auto-pull reconciler and deploy-mode endpoint; endpoint bypasses cache and uses the canonical fork ancestor chain.
  - Remove remote-URL matching and related flags/helpers/tests to simplify and reduce risk.
  - Bump EE repo reference to align with the latest EE merge.

<sup>Written for commit 71dc559c2df6a6cb07d4de403efc3c347975bed2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Addressed the actionable items from the last review; remaining ones are deliberate, narrow limitations.

**Fixed**
- **CI-backed deploys got the wrong recommendation:** `status` no longer assumes `wmill sync push` when backend auto-pull doesn't match — it reports the backend signal and defers CI-vs-sync-push to the user + `AGENTS.md` (`Deploy mode:` line), matching the Deploying guidance. `deploy_command` is `null` in JSON when undetermined.
- **Runnable-delivery gate:** `deploy_on_push` now requires an actual delivery path (active webhook, or a pollable non-app HTTPS repo), so a failed webhook or an SSH-only repo no longer advertises deploy-on-push. Live-verified (SSH → false, HTTPS polling → true).
- Doc-comment placement and the app-repo comment wording.

**Deliberately not implemented (narrow, safe):**
- `validate_git_url` parity for polling (localhost/private URLs): the poller's validator is async/network-doing; running it per-repo on an on-demand endpoint isn't worth it for a self-inflicted misconfiguration. The endpoint stays conservative and this only affects a git-sync pointed at a private/invalid URL.
- Webhook-id rebinding after a repo URL changes post-hook-creation: a very narrow drift window; the safe fallback is `wmill sync push`.

Both remaining cases are edges an operator would have to actively misconfigure into; the recommendation degrades to the always-safe `wmill sync push` path.

hugocasa:
Addressed the two cheap items; #1 is a deliberate design choice.

**Fixed**
- **Archived roots:** `deploy_on_push` now requires the root workspace to be live (`workspace.deleted = false`). Polling and webhook delivery both exclude deleted roots, so an archived root — or a live descendant beneath one — with retained git-sync no longer reports deployable.
- **`false` framing:** `status` and the OpenAPI now describe `false` as *not confirmed* (it also covers ambiguity and conservative false-negatives), not a definite no. The CLI asks the user how the repo deploys rather than asserting the push won't deploy.

**Deliberate trade-off (#1, branch-only matching):** dropping remote-URL matching in favor of repo-count disambiguation was a conscious decision (it removed the credentials-in-logs / shell-injection / URL-normalization surface entirely). Single-repo workspaces trust the branch match because a synced project's `origin` is the git-sync repo by construction; multi-repo workspaces return `false` and defer to the user. The residual "local checkout points at an unrelated remote" case degrades to the always-safe `wmill sync push` via the ask-the-user fallback. Claude's review confirms this as an accepted trade-off of the design.

hugocasa:
Re: the remaining nit (nested dev-workspace forks — `wm-fork/<dev-label>/<id>`): this is a pre-existing, system-wide limitation, not introduced here. The auto-pull poller only enumerates `wm-fork/<root-tracked>/*` plus the dev-workspace label branches, so forks *of* a dev workspace aren't collected or reconciled today either — the shared resolver just mirrors that. Non-dev forks resolve at any nesting depth (their base flattens back to the root's tracked branch), and direct dev workspaces on their label branch resolve too — both verified live.

Handling nested dev-workspace forks properly means teaching the poller to enumerate `wm-fork/<dev-label>/*` and the resolver to accept those bases; that's a change to the fork/auto-pull feature itself and will be handled more generally in a separate PR.

Good to merge.

PR #10202, opened 2026-07-20
fix: steer ai chat away from draft-blind api catalog reads and runs

hugocasa:
## Summary

Follow-up hardening for the global-chat API catalog tools (#10199): closes the remaining paths where the model could pick a catalog endpoint over the dedicated tool and act on stale, deployed-only state, and adds ai_evals coverage for catalog steering.

## Changes

- Block `getScriptByPath` / `getFlowByPath` in `COVERED_ENDPOINTS`: they return the deployed version, blind to the user's draft, and a flow read through them feeds `patch_flow_json` a serialization its exact-match edits can never hit. The refusal points at `read_workspace_item`, which prefers drafts.
- Prompt: drop "running deployed items" as a catalog example and add an explicit rule that `runScriptByPath` / `runFlowByPath` run the DEPLOYED version, are only for explicit user requests, and that drafts are tested via `test_run_*`.
- Three new ai_evals global cases: catalog used for a genuine gap (worker status via `search_api_endpoints` → `call_api_get`), draft-edit-then-run steering (`test_run_script` required, `call_api_endpoint` forbidden), and delete steering (`delete_workspace_item`, catalog untouched).
- Eval harness: the mock backend now serves a representative MCP endpoint catalog (`McpService.listMcpTools` wrap) and answers the executor's relative `/api/...` fetches — previously the catalog tools could not run at all under the harness.
- One regression unit test for the newly blocked reads (9 total in `apiCatalogTools.test.ts`).

## Test plan

- [x] `apiCatalogTools.test.ts` 9/9, `npm run check:fast` clean
- [x] New eval cases pass 3/3 with sonnet against a live `mcp`-feature backend (two consecutive runs)
- [ ] Full global eval suite run on this branch to check for regressions (in progress; results will be posted before review)

## Screenshots

No visible UI change: prompt text, tool guard map, eval cases, and eval harness only.

---
## Summary by cubic
Steers global chat away from draft-blind API catalog reads and runs. Adds `version: "deployed"` and script input `schema` to `read_workspace_item`, per‑type pagination with `page` and a capped page‑1 draft overlay to `list_workspace_items`, labels server‑synthesized `draft_only` rows as drafts, and enables the eval harness to run catalog tools with a narrowly scoped fetch stub.

- **Bug Fixes**
  - Block draft-blind reads/lists (`getScriptByPath`, `getFlowByPath`, `getResource`, `getSchedule`, `listScripts`, `listFlows`, `listResource`, `listSchedules`) and hide them from catalog search; errors point to `read_workspace_item` / `list_workspace_items`.
  - Clarify that `runScriptByPath` / `runFlowByPath` run the DEPLOYED version only; use `test_run_*` for drafts, and read with `read_workspace_item` `version: "deployed"` first to match the deployed input schema.
  - `read_workspace_item` supports `version: "deployed"` and includes a script’s input `schema`; defaults still prefer drafts.
  - `list_workspace_items` paginates per item type with `page`, overlays drafts on page 1 only capped at `limit` per type, and avoids cross-type truncation; server-synthesized `draft_only` rows are labeled `isDraft`.

- **Evals and Harness**
  - Add three global evals: worker status via catalog, draft edit then test-run (no deploy), and delete via `delete_workspace_item`.
  - Mock backend serves a representative `McpService.listMcpTools` catalog and answers specific relative `/api/...` calls; the fetch stub intercepts only handled benchmark paths.
  - Add tests covering blocked endpoints, deployed reads with schema, per-type pagination with page forwarding, capped page‑1 draft overlay, and no cross-type truncation.

<sup>Written for commit c1aa24fc3efdfeb199bb5089436647544db7c0f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10202?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
## ai_evals regression check (sonnet, live mcp-feature backend)

**Full global suite on this branch: 47/55 (85.5%).** All 8 failures triaged against a baseline run of the same cases on main:

- 6 of 8 reproduce identically on main (ambiguity ask-vs-act cases test7/test9/path3, openpage4 tab substring, docs-key-value-store, pipeline-create-node judge) — pre-existing, unrelated to this PR.
- 2 of 8 (docs-ai-agent-step, docs-retry-step) were a real regression from this branch's eval harness fetch stub: it intercepted every relative `/api/` call, turning `search_docs`' environment failure into a query-shaped 404 that sent the model into rephrase-retry loops until max turns. Fixed in 5ffed1e432 by scoping the stub to the paths the benchmark actually handles; both cases pass again and the three catalog cases still pass (verification run 5/6, only the pre-existing docs-key-value-store failure remains).

The three new catalog steering cases pass consistently: 3/3 in two dedicated runs plus within the full suite.

hugocasa:
Codex P1 addressed in 59efb4c9df: `read_workspace_item` now takes `version: "deployed"`, which skips both the chat draft and the editor DB draft (`getDraft: false`), so an explicit deployed run can read the deployed input schema. The refusal messages for the blocked `get*ByPath`/`getResource`/`getSchedule` endpoints advertise the option, and the deployed-run prompt rule now instructs reading with `version: "deployed"` before constructing arguments. Unit test pins that both draft layers are skipped. No eval case for the differing-schema scenario: case fixtures cannot seed chat drafts today, and the tool-level behavior is deterministic and unit-pinned; the steering half (deployed runs only on explicit request) is already covered by the prompt rule and test31.

hugocasa:
Codex round-2 P1 addressed: `scriptToItem` now returns the script's input `schema` alongside content (matching what flow reads already do via their compact JSON), so a deployed read exposes defaults/enums/required fields for argument construction. The deployed-read unit test asserts the schema is returned.

hugocasa:
Codex round-3 P1 addressed: `list_workspace_items` now takes a `page` parameter (forwarded to every underlying list endpoint, 100 per page), so items beyond the first page are reachable through the dedicated tool. Drafts merge into page 1 only, keeping later pages clean deployed listings — which also gives an explicit deployed-only listing view (any page > 1, or page 1 rows not marked isDraft). Unit test pins page forwarding and the page-1-only draft merge. The remaining catalog-specific server filters (resource broad_filter, schedule target path) stay unavailable by design: the query/path_prefix filters plus paging cover the discovery need, and re-exposing the raw list endpoints would reintroduce the draft-blind listing problem this PR exists to close.

hugocasa:
Codex round-4 P1 + P2 (and Claude's matching nit) addressed: `limit` is now the per-item-type page size and the cross-type truncation is gone — each type's list call is already capped, and slicing the concatenation was silently dropping later types' rows while their next page skipped past them (regression test added: a full page of scripts can no longer hide flows). Tool and schema descriptions now consistently say up-to-limit-per-type, default 50.

hugocasa:
Codex round-5 P1 addressed: the draft overlay is now windowed client-side with the same per-type page/limit as the deployed streams — results stay bounded (`limit: 1` returns one item even with many drafts) and every draft remains reachable by paging. Regression test walks two drafts across pages 1–3 with limit 1.

hugocasa:
Codex round-6 addressed by simplifying rather than adding logic: the draft windowing is replaced with a page-1-only overlay capped at limit per type. That removes the cross-page draft/deployed twin duplication structurally (drafts never appear past page 1), keeps results bounded, and drops the second client-side pagination stream entirely. Past the cap a narrower path_prefix still reaches any draft (it filters before the cap) — realistic chat draft counts sit far below the cap. The trigger fan-out is documented: limit applies per trigger kind for the trigger type. Regression test updated accordingly.

hugocasa:
Round-7 disposition, per maintainer direction to avoid further pagination complexity for edge cases:

- **Fixed** the substantive part of the second finding: rows synthesized by `includeDraftOnly` listings were mislabeled `isDraft: false`; `scriptToItem`/`flowToItem` now carry `draft_only` through, so draft-only items always read as drafts.
- **Dismissed** the cross-page draft/deployed twin duplication and the over-limit page-1 edge: both require a fully deduplicated cross-page pagination protocol for a metadata-only listing, which is complexity disproportionate to impact — Claude's review of the same delta adjudicates the duplication as cosmetic and non-blocking, and it only manifests when a drafted item's deployed twin sorts past page 1 (or with 50+ same-type drafts). `byKey` already dedupes within a page; each row's `isDraft` is now always accurate, so the model cannot act on stale metadata without noticing.

hugocasa:
✅ Review round clean @ 234516137822ea2623f1c9ff5e439a6f8c8a7ce7 — nit-only verdicts at af5406412bfe1c714124df8fd21358338fa4d8e5 (Codex and Pi: "Mergeable, but should ideally address nits"); both nits addressed in 234516137822ea2623f1c9ff5e439a6f8c8a7ce7 (stale mcp-prerequisite comment corrected, draft_only → isDraft passthrough pinned in the listing test).

PR #10203, opened 2026-07-20
fix(apps): stop the app editor crashing on a grid-less draft

rubenfiszel:
## Summary

Editing a low-code app could throw `Uncaught TypeError: Cannot read properties of undefined (reading 'map')`, leaving the editor canvas blank ("App is not yet initialized") and the Preview / Deploy panes unusable, while viewing the same app in live mode kept working.

The error comes from `getRowsCount` doing `items.map(...)` on an undefined `app.grid`, reached via `AppPreview` -> `GridViewer` -> `getContainerHeight`.

The reason live mode was unaffected is that the viewer renders the *deployed* value, whereas the editor renders the *draft* when one exists. So a draft whose stored value has no `grid` breaks editing only.

That exposed the underlying gap: `AppEditor` normalizes the `app` prop through `migrateApp`, but `stateApp` adopts `appDraftHandle?.draft ?? app` and an adopted draft bypassed normalization entirely, despite the comment directly above asserting that `stateApp` is `migrateApp`'d on mount. A persisted draft is precisely the kind of value that can predate a schema migration, so it was the input most in need of it.

## Changes

- `migrateApp` defaults a missing or non-array `grid` to `[]`. It is the one hook every app load path runs through (`AppPreview`, `AppEditor`, the edit route's overlay baseline, the localStorage to DB draft migration).
- `AppEditor` runs `migrateApp` on whichever value `stateApp` adopts, so a draft gets the same normalization as the prop.

## Verification

Reproduced the reported scenario locally by seeding a deployed app whose value has a `grid` alongside a draft row whose value has none.

Before: live view rendered fine; the editor showed a blank canvas with "App is not yet initialized"; switching to Preview threw with the same `getRowsCount` -> `getContainerHeight` -> `GridViewer` -> `AppPreview` frames as the report.

After: the editor initializes with a real canvas and Preview renders cleanly, no console error.

Also confirmed that mutating the draft cell at init does not trigger a spurious autosave: after a full editor mount the stored draft row is unchanged (no `grid` key added, `created_at` untouched).

`npm run check:fast` reports one pre-existing, unrelated error in this worktree (`modern-screenshot` missing in `RawAppEditor.svelte`) that reproduces identically on unmodified `main`.

## Screenshots

Editor before the fix, blank canvas and "App is not yet initialized":

repro-editor-broken

Editor after the fix:

fixed-editor

Preview after the fix:

fixed-preview

## Follow-up

This stops the crash and restores editing, but it does not explain how a grid-less draft was written in the first place; the repro seeded that row by hand. If a draft genuinely has no `grid`, its components are already absent from that draft, so recovering such an app means discarding the draft and re-editing from the deployed version. Worth tracing the write path separately.

## Test plan

- [ ] Open an app with a draft whose value omits `grid`: editor renders an empty canvas, Preview does not throw
- [ ] Confirm no `POST /drafts/update/app` fires on mount before any edit
- [ ] Regular apps with a populated grid still render and autosave normally

---
## Summary by cubic
Fixes an editor crash when opening an app with a draft that lacks a grid. Missing grids are now normalized so both the editor and Preview load reliably.

- **Bug Fixes**
  - Default missing/non-array `app.grid` to `[]` in `migrateApp`, covering viewer, editor, and draft migrations.
  - Run `migrateApp` on the adopted value in `AppEditor` so drafts are normalized like props.

- **Tests**
  - Dropped the `migrateApp` grid-default test.

<sup>Written for commit 58b571f3d715ab0ab23529d981aad9770d5ab90f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 8f958ae8f9220f2cd7933e555f663f4c50580482

PR #10210, opened 2026-07-20
feat(ai): live web-search source list on chat tool cards

Guilhem-lm:
## Summary

The chat's web-search tool call previously rendered as a bare "Searched the web" label with nothing behind the chevron. It now shows the search query in its label (streamed live while the model types it) and surfaces the list of consulted pages — favicon, title, link — the moment each search completes mid-stream, auto-expanding the card while the model continues its answer. Sources persist with the conversation.

## Changes

- `shared.ts`: new `WebSearchSource` type and `webSearchSources` field on `ToolDisplayMessage`.
- `anthropic.ts`: the query is extracted live from the `server_tool_use` block's streaming partial JSON (`partialWebSearchQuery` regexes the string prefix out of unparseable mid-string JSON); the `web_search_tool_result` block arrives complete in `content_block_start`, so sources surface mid-stream instead of at end of turn. The end-of-turn handler only re-emits for results whose stream event was missed, so a user's manual collapse isn't overridden.
- `openai-responses.ts`: requests `include: ['web_search_call.action.sources']` and reads query + source URLs from `response.output_item.done` mid-stream (with an end-of-turn fallback sweep). OpenAI sources carry no title, so rows fall back to the hostname.
- `WebSearchSourcesDisplay.svelte` (new): deduped source list rendered in the expanded card; only absolute http(s) URLs become links (provider-supplied URLs must not inject `javascript:` hrefs). Favicons load from `t3.gstatic.com` directly because the app is served with COEP `require-corp` and the `google.com/s2/favicons` redirect hop carries no CORP header; unknown favicons fall back to a Globe icon via `onerror`.
- `ToolExecutionDisplay.svelte`: renders the sources list in place of the empty generic Result section. Cards with sources set `showDetails: true, autoCollapseDetails: false`, which is what auto-expands them and keeps them open.
- `anthropic.test.ts`: unit tests pinning `partialWebSearchQuery`'s escape handling (escaped quotes, mid-escape cut, no query yet).

## Screenshots

Query streamed into the label, then the sources list auto-expanded mid-stream while the answer streams below (Anthropic web search, live run):

websearch-live

Expanded card close-up (favicons, titles, domains):

websearch-sources-favicons

## Test plan

- [x] Unit tests for the partial-query extraction (`npx vitest run .../anthropic.test.ts`, 7 passing)
- [x] Live browser run (Anthropic): label streams the query while searching, sources auto-expand mid-stream with working favicons/links, card label ends as `Searched the web for "<query>"`, zero console errors
- [x] Persistence: sources survive a full page reload and older pre-change cards still render collapsed/expandable
- [ ] OpenAI Responses model run (query + hostname-only sources at completion) — needs a workspace with an OpenAI provider

## Review

`/local-review` (Claude): no blockers, 3 P2 nits — all addressed in the second commit (comment accuracy, end-of-turn re-emit no longer overrides a user collapse, favicon third-party tradeoff documented). `/local-review-codex`: flagged P1 `javascript:` scheme injection via provider-supplied source URLs — fixed by filtering to absolute http(s) before rendering.


Guilhem-lm:
✅ Review round clean @ 6399e1faa23403333ef0fee7142312c86f35522f — nit-only verdicts at 02436b45b00bd7061f2122ccd8a81610b58b02b2 (codex: nit; claude, pi: good to merge); nit (plural `queries` field for OpenAI labels) addressed in 6399e1faa2

PR #10211, opened 2026-07-20
feat: unified read-only `diff` chat tool (drafts, fork vs parent, search)

Guilhem-lm:
## Summary

Adds one read-only `diff` tool to the global AI chat that covers every comparison the deleted per-mode chats used to offer, and more: draft-vs-deployed for any draftable kind, whole-workspace change index, deployed-fork-vs-parent, per-file raw-app diffs, and a grep over changed lines. All queries are served from a materialized per-workspace snapshot cache, so nothing is fetched or recomputed twice.

## Changes

- **`diff` tool** (`copilot/chat/global/core.ts`) — flat schema, mode by arg presence:
  - no args → index of every draft with change status (filters: `types`, `path_prefix`, `limit`); multi-file apps render as folders with per-file children
  - `type`+`path` → that item's unified patch over stable YAML (`offset`/`limit` line windows); `file=` reads one file inside a raw app (lenient name matching)
  - `against="parent_workspace"` → deployed fork vs parent, identical semantics to the compare page; items with local drafts are flagged but not diffed (consistent with the fork UI)
  - `search=` → literal grep over added/removed diff lines across all patches (`file_glob`, `max_matches`, search_app output conventions)
- **Snapshot cache** (`diffSnapshot.ts`) — patches computed once per (workspace, mode); freshness from three cheap signals: per-row draft `created_at` (server bumps it on every draft upsert), the `invalidateWorkspaceDrafts` version (now also bumped by chat draft writes), and comparison ahead/behind markers for fork entries. Eager materialization capped at 50 with lazy overflow; search lifts the cap.
- **Shared comparison fetch** (`$lib/workspaceComparison.ts`) — single-flight `compareWorkspaces` layer; `ForkWorkspaceBanner` and `/forks/compare` now fetch through it, and the chat tool reuses a ≤30s-old result instead of recomputing the tally.
- **Patch formatting** (`draftDiff.ts`) — deterministic YAML (sorted keys, block scalars, null-pruning) + raw text patches for app files.
- **Safety**: fork-mode variables are fetched with `decryptSecret: false` and secret values placeholder-masked (masked items are reported as "may differ", never "unchanged"); `bundle_secret` and cross-workspace noise stripped from app rows; the pre-diff autosave flush honors the auto-save toggle so a read-only call never persists edits the user chose to keep local.
- Draft-vs-deployed script rows are stripped of server-managed fields in the shared `getDraftDiffValues`, which also cleans the existing CompareDrafts/session diff drawers.

## Local review

Both pre-push reviews ran (Claude `local-review` + Codex `local-review-codex`); all findings addressed in this branch: NUL bytes in `workspaceComparison.ts`, autosave-toggle-violating flush, post-flush stale listing window, fork projections missing `description`/`resource_type`, secret-only changes misreported as unchanged, `skipped_comparison` false negatives in item/search modes, hardcoded trigger-kind list, caller-less export.

## Test plan

- [x] `draftDiff.test.ts` + `diffSnapshot.test.ts` (19 tests: YAML patch semantics, cache reuse/invalidation incl. post-flush expiry, eager cap, friendly-path resolution, fork side mapping and marker reuse, secret redaction + masking flag, script description-only fork changes, raw-app per-file splitting) — 372/372 across the chat/session suites
- [x] `npm run check:fast` clean except a pre-existing unrelated error in `utils_workspace_deploy.ts` (from the datatable-migrations merge, untouched here)
- [x] Browser E2E with a live model (claude-sonnet-4-6), fresh sessions: draft index + item read (screenshot 1), fork index/one-sided/ahead/behind + parent↔fork patches on a real fork (screenshot 2), raw-app folder children + `file=` read + cross-diff search (screenshot 3)
- [ ] Paid `ai_evals` A/B benchmark (case `global-draft-diff-report` updated for the new tool name) — pending
- [ ] Manual: in a fork, confirm the banner and `/forks/compare` still load their counts (both now route through the shared fetch layer)

## Screenshots

Draft mode — workspace index + exact item diff:

diff-tool-stage1-session

Fork mode — index (only-in-fork / ahead / behind) + parent↔fork patches:

diff-tool-stage2-fork-session

Raw-app folder children, per-file read, and changed-line search:

diff-tool-stage3-search-files


Guilhem-lm:
**ai_evals A/B benchmark (sonnet, 7-case subset)** — new diff case + 5 draft/edit canaries + 1 unrelated read-only control:

| case | before (main) | after (this PR) |
|---|---|---|
| global-draft-diff-report | FAIL (judge 72 — no diff tool) | **PASS (judge 95)**, tools: read → edit → diff |
| global-test1-script-create | PASS 100 | PASS 100 |
| global-test2-script-edit-existing | PASS 100 | PASS 100 |
| global-test5-existing-flow-inline-code-edit | PASS 100 | PASS 100 |
| global-test6-secret-variable-draft | PASS | PASS |
| global-test12-current-live-script-edit | PASS 100 | PASS 100 |
| global-test27-list-recent-runs (control) | PASS | PASS |

No regressions on the canaries. Per-iteration context cost of the new tool (schema + system-prompt rule), measured on the unrelated control case: **+867 final-context tokens**.

The first after-run looped to max turns — an eval-harness fidelity gap, not a product issue (the mock draft backend ignored `get_draft` overlays and used a fixed `created_at`, so the diff tool saw no drafts). Fixed in 4d06ace488; the browser E2E against a real backend had already exercised the same path correctly.

Guilhem-lm:
**Architecture soundness review** (cold-context pass over the diff subsystem, big design flow only):

Verdict: sound, reuse verified. Draft mode shares `getDraftDiffValues` verbatim with the CompareDrafts drawer; fork mode reuses `getItemValue` + the `compareWorkspaces` single-flight store now shared with the fork banner/compare page; `draftDiff.ts` is new code for a genuinely new need (text patches for the model vs Monaco's visual DiffEditor for humans) — no pre-existing diff function was recoded.

Agreed follow-up (separate PR after this merges): extract the chat-safe per-kind projections from `diffSnapshot.fetchForkSideValue` into a `getDiffItemValue` adapter beside the `getItemValue` wrapper in `utils_workspace_deploy.ts`, plus a drift-pinning unit test asserting the projections stay a superset of `getItemValue`'s fields. Rationale: "which fields of a script/resource are diffable" currently lives in two files with nothing linking them, and the never-decrypt-secrets rule should live where the next model-facing consumer will find it. Folding into `windmill-utils-internal` itself was considered and rejected: cross-repo publish train, and the raw-app canonicalization depends on frontend-only modules that cannot cross the package boundary.

Noted, not scheduled: the trigger noise-field set exists in three hand-synced copies (frontend `TRIGGER_RUNTIME_IGNORE`, `windmill-utils-internal`, backend `TRIGGER_COMPARE_IGNORE`) — pre-existing duplication this PR inherited.

Guilhem-lm:
Both findings addressed in 8902b1a0d9:

1. **Classic-app drafts in index/search** — `itemTypeForKind` now maps both `app` and `raw_app` draft kinds to the chat `app` type, matching the read path's pairing. Classic-app drafts materialize in the index and their patches are searchable. Regression test added (index materializes a kind=`app` row with type `app`).

2. **`++`/`--` changed-line search** — replaced the prefix-exclusion predicate with hunk-aware scanning (`changedLineIndices` in draftDiff.ts): file labels only occur before the first `@@`, so a changed source line rendered as `+++counter`/`---counter` now matches while labels never do. Regression test pins exactly that byte-identical-prefix case.

Guilhem-lm:
Both findings addressed in 09306459fb:

1. **Variable values redacted everywhere** — draft-mode diffs now mask both sides of every variable (not just secrets) with a placeholder pair that still marks *whether* the value changed, never its content; fork-mode masks at fetch time (and still never decrypts). Item reads carry an explicit "values are never shown in chat" note, and the fork "unchanged" messaging says "value never shown; may differ" instead of claiming content parity. The former secret-only flag is generalized to `valueMasked`. Regression tests: non-secret fork variable leaks no plaintext; draft-mode value change shows the changed marker with no plaintext.

2. **Classic-app sides canonicalized** — the classic editor drafts the bare grid value while the deployed row nests it beside summary/policy, so a raw diff read as a whole-document rewrite. Both sides now reduce to `{ value }` via a pure `classicAppDraftValue` helper (defensively unwraps legacy wrapped drafts; wrapper metadata can't change through a classic draft, so it doesn't belong in the diff). This also cleans the CompareDrafts drawer's classic-app diff, which shared the same shapes. Unit tests cover bare, legacy-wrapped, and grid-with-a-`value`-key shapes.

Guilhem-lm:
All three addressed in 35182be9bb:

1. **Secret-only drafts no longer claim "unchanged"** — `maskVariableDiffSides` now recognizes that a secret's sides are both masked upstream (draft rows store `''`, deployed values are never decrypted), so equality proves nothing. Such entries carry `valueUncomparable`: the index says "no visible changes (secret value cannot be compared; may differ)" and the item read says the value "may have been updated in the draft" — mirroring fork mode's honesty. Regression test pins the sentinel-equal case.

2. **Classic-app metadata split from the grid** — confirmed in `apps/types.ts` + `AppEditorHeader`: the editor mirrors `summary` (and `draft_path`) into the bare `App` precisely so drafts round-trip them. `classicAppDraftParts` now extracts `{ value, summary }` with draft-only markers stripped, applied to BOTH sides (an older deployed value retaining an inner summary stays comparable), so summary edits diff as summary edits and `draft_path` never appears. Tests cover the mirrored-metadata, legacy-wrapped, and component-named-`value` shapes.

3. **Glob-safe search subjects** — file keys' leading slashes are stripped when building `path/file` subjects, so `f/x//index.tsx` becomes `f/x/index.tsx` and natural filters like `f/x/*.tsx` match.

Guilhem-lm:
All three addressed in 527c1f5782:

1. **Classic-app local edits** — the chat `app` type spans two draft kinds, so item mode now flushes and probes BOTH the `raw_app` and classic `app` keys (using whichever kind holds the unflushed cell for the local diff), and the index/search flush sweep includes classic-app editor cells. `app` stays out of `GLOBAL_DRAFT_KINDS` deliberately — `clearGlobalDrafts` must never clear a user's open classic editor — the sweep adds it explicitly.

2. **Comparison-relevant fork fields** — flow projections now carry `schema` (with a taxonomy-agnostic deep walk stripping per-workspace inline-script hashes) and resource types carry `description`/`format_extension`/`is_fileset`. One caveat: folder `display_name` is **not exposed by the API's `Folder` type** (`GetFolderResponse` has name/owners/extra_perms/summary/labels only), so it cannot be projected client-side — if the backend comparison truly hashes it, that field needs an API addition first. Regression test: fork flow schema-only change materializes as `modified` with no hash noise.

3. **Conflicts count as unflushed** — a conflicted save leaves its payload parked with state `none`, which the sweep previously read as clean. Index/search now include conflicted paths in their warning ("auto-save off, a save failed, or a conflict is unresolved"). A syncer-level test pins the exact contract triple (conflict set, state `none`, auto-save signal silent).

Guilhem-lm:
All three addressed in the latest commit:

1. **Staged renames diff as `path`** — both app branches now put the deployed path and the staged `draft_path` on their respective sides, so a rename-only draft shows `-path: f/old` / `+path: f/new` in index, item read, and search (and in the shared drawer). `classicAppDraftParts` extracts the staged path separately from the grid; the local-editor branch mirrors the same field.

2. **Classic-app cells flushed at their real keys** — item mode now resolves per kind (live-editor mapping per draft kind) AND asks the listing which row owns a friendly/renamed path (`resolveWorkspaceDiffTarget`), then flushes/probes every resolved candidate. A renamed classic app's unsaved edits are found at its original storage key. Regression test pins the friendly-path → owning-row resolution across kinds.

3. **Comment compressed** to the ≤4-line rationale limit.

Guilhem-lm:
Addressed in the latest commit, with one correction to the finding's premise: **no folder API surface exposes `display_name`** — not `getFolder`, not `listFolders` (both return the `Folder` schema: name/owners/extra_perms/summary/created_by/edited_at/default_permissioned_as/labels). The column exists only in the DB (workspace export selects it directly), so the comparison tally can flag it while no client can project it.

Given that, the honest client-side fix shipped: fork index and item reads for folders no longer claim "content matches parent" — they say the display name is not exposed by the API and may be what differs (same pattern as masked variable values). The real fix — adding `display_name` to the folder endpoints — is a backend API change out of this PR's scope; tracked as a follow-up.

Guilhem-lm:
Both addressed in the latest commit:

1. **Gap-free pagination** — `windowPatch` (now a tested pure helper in `draftDiff.ts`) cuts at the last complete line when the char backstop triggers and sets the continuation offset to exactly the first undelivered line; a single line larger than the whole budget is delivered truncated and stepped past so pagination always advances. Tests pin both: successive windows cover every line exactly once, and the pathological-line case.

2. **Forced-fresh comparisons** — in-flight requests now carry their start time, and `maxAgeMs` gates joining them the same way it gates cached results: `maxAgeMs: 0` always issues its own fetch (never adopting a tally that began before the mutation), and a superseded older request can no longer overwrite a newer cached result. Tests pin the no-adoption and tolerant-join behaviors with a held in-flight request.

Guilhem-lm:
Both addressed in the latest commit:

1. **Generation-ordered cache writes** — request ordering no longer rides on `Date.now()` (which collides under concurrency): each request takes a monotonic generation, and a superseded request's late result can never overwrite a newer one. Regression test freezes the clock, resolves the newer forced request first, then verifies a tolerant re-read still sees it after the older request lands.

2. **Every fork comparison kind is addressable** — a fork item read may now omit `type`: path-only reads resolve across all comparison kinds, which is how folder/resource_type (and any future kind outside the chat type enum) become readable without polluting the global type schema. Ambiguous paths (same path differing under several kinds) return the kind list and ask for `type`. Item-read messages label entries by their comparison kind, and pending index lines for enum-less kinds say "read it by path alone". Test covers the wildcard read and the ambiguity case.

Guilhem-lm:
All four addressed in the latest commit:

1. **Wrapper detection keys on `grid`** — a bare App always carries `grid` at top level (and the editor mirrors `summary` into every bare grid, as the finding notes), while a legacy wrapper never does. The guard test now uses the production shape (grid + component named `value` + mirrored summary) and pins that the full grid survives.

2. **First post-deploy fork read is fresh** — `invalidateWorkspaceDrafts` now also drops cached comparisons for the mutated workspace, so a banner-prewarmed pre-deploy tally can't serve the first chat fork read (the drafts-version baseline only covered reads with a prior snapshot). Test: prewarm → invalidate → tolerant re-read fetches fresh.

3. **Ambiguity eliminated instead of instructed away** — a path-only wildcard read now returns one section per matching kind (nothing in the chat type schema could disambiguate folder vs resource_type, so the tool stops asking the model to do the impossible). Test covers two kinds sharing a path, both materialized.

4. **Hidden differences never read as parity** — zero visible entries with hidden diffs says "no differences visible to you — hidden items differ", and hidden counts stay directional (`N ahead, M behind`, noting a conflicted item counts in both) instead of a double-counting sum.

Guilhem-lm:
**Cubic review batch** addressed in 075f0c0630 — triage of all 21 inline comments:

**Fixed (12):**
- *(P1, latest run)* Invalidation now matches **either side** of a comparison pair — a parent-workspace deploy evicts its forks' tallies — and fork snapshots additionally baseline the parent's drafts version.
- Invalidation **fences in-flight requests**: nobody new joins them and their late results are rejected via a per-key generation floor.
- Comparison eviction moved from every drafts-version bump to **deploy success only** (draft saves never move the deployed tally), and the cache is LRU-capped at 8 pairs.
- Draft materialization carries a **stale-generation token**: a save landing mid-fetch discards that run's pre-save result; awaiting readers fall through to a fresh fetch.
- A save/delete also expires the fork cache so `hasLocalDraft` flags stay current.
- `onAnySaved` listeners are **error-isolated** (a throwing listener can't mark a committed save failed), and the pagehide **keepalive flush notifies them on dispatch** (that path bypasses postSave).
- Folder fork lines drop the empty `()`; patch-window offsets past the end report themselves instead of `(lines 6-5 of 5)`; the resolver test is labeled; the eval case pins the `diff` call's `path` argument.

**Already fixed by intervening commits (7):** double-slash glob subjects, drawer-kind local canonicalization, per-kind flush paths, same-millisecond generation ordering, judge-visibility checklist, compare-page re-poll joins (default `maxAgeMs: 0` never joins), and the kind-join mapping (`app`→`app`, `email_trigger`→`trigger_email` are correct in the current head).

**Declined with reasons (2):**
- *Prune empty-after-null objects*: collapsing `{x: {a: null}}` to omit `x` would hide genuinely meaningful empty-object fields (e.g. `schema: {}`); the current behavior — showing `x: {}` — is the honest reading.
- *Stamp comparison freshness at completion*: `fetchedAt = startedAt` is deliberate — the data reflects server state at request time, so completion-stamping would let a slow request serve up to `duration + maxAgeMs` of staleness. The cost (a >30s request expiring on arrival) is the rarer, safer failure.

Guilhem-lm:
All findings from both reviews addressed in the latest commit — structurally this time, since three of the four P1s were instances of the same class (async producer racing an invalidation), and patching them pairwise is what produced the last several rounds:

- **One fencing primitive per cache.** `diffSnapshot` now has a per-workspace **mutation epoch** bumped by every invalidation; both reconcilers run a bounded retry loop where joiners re-validate after awaiting and producers refuse to store results whose inputs predate a mutation. That single mechanism covers Codex #1 (in-flight listing adoption) and #2 (pre-deploy fork tallies served to joiners) — and any future producer added to the module. `workspaceComparison` replaces per-key inflight fencing with per-**workspace** generation floors: any request started before an invalidation is fenced from joining and from cache-landing, including superseded requests the inflight map no longer tracks (Codex #3). Tests pin both: a mid-flight save forces a listing refetch with the new row present; a fenced superseded request resolving late cannot repopulate the cache.
- **Codex #4** — `delete_workspace_item` now invalidates comparisons exactly like deploy success (deployed state moved).
- **Codex #5 / Pi P2** — empty *filtered* indexes say "no drafts match your filters (N exist)" / "no differing items match your filters" instead of claiming workspace or fork parity.
- **Pi's in-flight reconcile P2** is the same as Codex #1 — covered by the epoch loop.

Guilhem-lm:
Addressed the latest Codex review in df0030375a:

- **Direct deploys never invalidated comparisons (P1)**: the chat `deploy` tool's non-script branches (schedule/trigger/resource/variable/app) deploy via the services directly and bypassed `deployDraftToWorkspace`'s invalidation. The shared success tail in `deployDraft` now calls `invalidateWorkspaceComparison(workspace)` before the fallible draft cleanup, so every successful deployed-state mutation evicts cached fork comparisons even if cleanup throws. Double-firing on the script/flow path is harmless (idempotent eviction).
- **Secret caveat swallowed by metadata changes (P1)**: a secret variable whose description/metadata also changed produced a non-empty patch and the value-uncomparable caveat vanished. Now: item reads emit a distinct secret note (`SECRET_UNCOMPARABLE_NOTE`) whenever `valueUncomparable` is set (snapshot, local-cell, and overlay-fallback branches), the index suffixes modified secret entries with "secret value may also differ", and fork sections note value uncertainty on modified masked variables. Regression test pins that a secret with a changed description yields `status: 'modified'` **and** `valueUncomparable: true` with no plaintext in the patch.

Guilhem-lm:
All three P1s from the 15:11Z Codex review addressed in b3f643b974:

- **Caches survive account changes**: both cache modules are now identity-scoped on `usersWorkspaceStore` email. On any change (logout sets it undefined, login sets the new account) `diffSnapshot` bumps the mutation epoch for every tracked workspace (fencing in-flight producers started under the previous identity) and wipes draft + fork caches; `workspaceComparison` wipes cache + inflight and raises a **global** generation floor — so even superseded requests nothing tracks anymore can neither be joined nor land. Regression tests: same-account reuse vs cross-account refetch (drafts), and an old-account in-flight comparison that is neither joined nor cached after the switch.
- **Fork snapshot doubling the freshness window**: `fetchWorkspaceComparisonMeta` now exposes the underlying request's `fetchedAt`/`generation`; the fork snapshot stores those instead of stamping `Date.now()`, so reuse ages from when the comparison was actually fetched. Regression test adopts a 25s-old tally and asserts a read 10s later refetches.
- **Failed draft cleanup leaving the fork snapshot stale**: the snapshot's reuse gate additionally checks `isComparisonCurrent` (generation vs invalidation floor), so the `invalidateWorkspaceComparison` call every deploy/delete now fires *before* cleanup immediately stops fork-snapshot reuse — no drafts-version bump needed. `delete_workspace_item`'s invalidation also moved ahead of the fallible `deleteGlobalDraft`. Regression test invalidates the store without a drafts bump and asserts the snapshot refetches.

`fetchWorkspaceComparison` keeps its original signature (banner + compare page untouched); it's now a thin wrapper over the meta variant.

Guilhem-lm:
Pi's P1 fixed in 72a3e6894e:

**In-flight fork reconciliation joins survived account switches** — exactly right: `ForkCache` had no `epoch` field, so the joiner compared its own (post-bump) loop-top epoch against the current epoch — trivially equal — while `reconcile`'s joiner correctly checks the *joined cache's* producer epoch. Applied both suggested fixes: `ForkCache` now carries its producer's epoch, checked by the joiner AND the reuse gate (`joined.epoch === mutationEpoch(workspace)` / `prev.epoch === epoch`), closing the whole before-arrival-bump class, and `ensureCacheOwner` additionally clears `reconciling`/`forkReconciling` so a cross-identity join can't happen at all (including through the bounded-retry escape hatch).

Regression test pins it: a fork reconciliation left in flight across a switch is not joined by the new account, its late result never lands, and even the pre-switch caller's promise resolves to the fresh-identity snapshot.

Guilhem-lm:
Codex 15:27Z round addressed in 31f4f51202:

- **P1 — fork fetch failures became valid-looking diffs**: correct — the shared `getItemValue` returns `{}` for any failed fetch (right for the deploy drawer, where a missing side means one-sided; wrong here). Since `materializeFork` only fetches sides the comparison lists as existing, the default read path now treats an empty read as a transient failure and throws — surfaced through the existing per-entry `error` status + `errorMessage`, never as a fabricated addition/removal or parity. The per-kind override paths (variable/script/resource/flow/resource_type/raw-app) already threw naturally. Regression test: both sides reading `{}` yields `status: 'error'` with the fetch-failure message.
- **P2 — conflicted edits promised but omitted**: the item read's local-override branch now also triggers on an unresolved conflict (`UserDraftDbSyncer.getConflict`), diffing the in-memory editor value like the autosave-off/failed-save cases — honoring the index's "read the item to include them" promise. The caveat now names which side is shown: your local editor value when the cell is available, the persisted draft when it isn't (e.g. editor closed).

No unit regression for the conflict path itself — exercising it needs the full syncer+editor cell setup; the syncer conflict-triple contract is already pinned in `userDraftFlushToggle.test.ts`, and the branch change is a one-predicate extension of the tested detection. Flagging it for the manual test plan instead.

Guilhem-lm:
Codex 15:38Z P1 fixed in 19d53e4cfc: `collectDiffSearchUnits` now diverts `status: 'error'` entries into a `failedPaths` list instead of dropping them, and both search outcomes (matches and no-match) append a warning naming exactly which diffs could not be computed and were NOT searched, with a retry hint — a search can no longer present an incomplete scan as definitive.

On the requested regression: `diffSearch` lives in the tool layer (core.ts), which has no unit harness — standing up one for this would mock the whole tool context to assert a string concat. The error-entry production itself is pinned in `diffSnapshot.test.ts` (swallowed-fetch → `status: 'error'` + `errorMessage`), and the routing change is a two-branch pure function. Added to the manual test plan instead.

Guilhem-lm:
Codex 15:53Z finding triaged — **declined as not applicable**:

**[P1] "AMQP trigger fork diffs always fail"** — there is no AMQP trigger kind anywhere in this API surface. `WorkspaceItemDiff.kind` (the backend comparison enum in `src/lib/gen/types.gen.ts`) enumerates exactly: script, flow, app, raw_app, resource, variable, resource_type, folder, schedule, and http/websocket/kafka/nats/postgres/mqtt/sqs/gcp/azure/email triggers — no `amqp_trigger`, and `grep -ri amqp src/lib/gen` is empty. The comparison can never emit the entry the finding describes, so the failing input is unrepresentable.

I did check the nearest real neighbor the reasoning could apply to: `email_trigger` **is** in the comparison enum while missing from the chat's `TRIGGER_KINDS` addressing enum. It works: the shared reader supports it (`legacyTriggerKind` maps `email_trigger → 'emails'`, so `getItemValue` fetches real content, never the `{}` sentinel), and it's readable through the path-only wildcard read that exists precisely for enum-less kinds (folder, resource_type) — listed in the fork index, materialized, searchable, addressable by path.

PR #10212, opened 2026-07-20
fix(db): grant schema usage and re-run windmill role grants

rubenfiszel:
Fixes WIN-2208

## Problem

Migration `20250205131523_grant_all_in_current_schema` puts all its grants in one `DO` block whose first statement is `LOCK TABLE pg_catalog.pg_roles` — which a non-superuser cannot take. On managed Postgres (RDS, Cloud SQL) it raises, and the block's single `EXCEPTION WHEN OTHERS` handler downgrades the failure to a `NOTICE`, so every `GRANT USAGE`, `GRANT ALL ON ALL TABLES`, and `ALTER DEFAULT PRIVILEGES` after it is silently skipped. Confirmed on PG 18: a non-superuser gets `ERROR: permission denied for view pg_roles` on that lock.

**What that actually costs is narrower than it first appears**, and the PR is scoped accordingly:

- The *table* grants mostly survive, because `20221105003256_grant_all` grants `ALL ON ALL TABLES` to both roles outside any such block.
- What is lost is the `ALTER DEFAULT PRIVILEGES`, which covers tables created by *later* migrations. Default privileges apply only to objects created by the role that set them, so a deployment whose migration runner never successfully ran them ends up with newer tables ungranted and writes failing with `permission denied for table`. **This is the gap that forced `20260619091631`, `20260701083047` and `20260716151337` to each patch a single table by hand.** Re-establishing the default privileges under the current runner closes it for future tables instead of one migration at a time.
- The *schema* `USAGE` grant matters only on databases that revoked `USAGE` from `PUBLIC`. There the roles resolve no table and queries fail with `relation does not exist` rather than a permission error.

## Changes

**`backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles`** — re-establishes the grants of `20250205131523` with no `LOCK TABLE`. Editing the original in place is not an option: sqlx checksums applied migrations, so every existing deployment would fail with "previously applied but has been modified". Down is a no-op (revoking would leave the instance unable to run any `user_db` query), matching the original.

**`init-db-as-superuser.sql`** — adds `GRANT USAGE ON SCHEMA public`, and moves `windmill_admin` creation above the table grants; it previously ran last, so role membership was not yet in place while they executed.

## The migration cannot abort an upgrade (primary design goal)

A grant migration that fails would block the entire upgrade, so this one is built to never raise. Two things get it there:

1. **Grant only runner-owned objects.** The blanket `GRANT ALL ON ALL TABLES IN SCHEMA` raises a hard `ERROR: permission denied for table X` — *not* a warning — the moment the schema holds one object the runner does not own: a superuser-installed extension (PostGIS's `spatial_ref_sys`) or a co-located app table. So the migration iterates `pg_tables`/`pg_sequences` filtered to `owner = current_user` and grants each individually. Owning the object is exactly the condition under which the GRANT succeeds, so this skips what would fail (Windmill's tables are all runner-owned → coverage unchanged) rather than the original's approach of swallowing the error.

2. **Guard every grant individually.** Each grant is wrapped in its own `BEGIN/EXCEPTION WHEN OTHERS/RAISE WARNING`. This is deliberately the *opposite* of `20250205131523`'s single block-wide handler — the original's flaw was **granularity**, not the presence of a handler: one early failure (the lock) silently skipped every remaining grant. Here a failure is isolated to one object, re-raised as a **named `WARNING`** (visible in the migration log), and the rest still run. This absorbs the last abort paths: an object dropped by another session between the catalog scan and the GRANT, `USAGE` on a schema the runner can't grant, or (defensively) a missing role. On a clean owned schema the guards never fire — zero warnings.

`ALTER DEFAULT PRIVILEGES` is also guarded; it only governs the runner's own future objects, so it cannot conflict with anyone else's. Both windmill roles are guaranteed to exist here — the unguarded `20221105003256` grants to both, so any DB that reached this migration already has them.

## Corrections after review

Two claims in the first draft were wrong and are fixed in the comments:

1. *"PostgreSQL 15 no longer grants USAGE on public to PUBLIC."* PG15 revoked `CREATE`, not `USAGE` (thanks @codex). USAGE is still granted by default, so the explicit grant is a no-op on a stock database and only matters on hardened ones. The `relation does not exist` repro only appeared because I revoked it by hand. (This myth is also in the linked issue.)
2. *"Core tables stay ungranted."* Overstated — `20221105003256` covers them, per above.

The migration is kept despite that narrowing: idempotent, ~5ms, and it closes the default-privileges gap that has already produced three one-table patch migrations.

## Validation (PG 18, non-superuser runner, `REVOKE ALL ON SCHEMA public FROM PUBLIC`)

**Before** — full history applies "successfully" (lock failure swallowed):
```
SET ROLE windmill_user;  SELECT count(*) FROM workspace;  -- ERROR: relation "workspace" does not exist
SET ROLE windmill_admin; SELECT count(*) FROM script;     -- ERROR: relation "script" does not exist
```
**After** — fresh DB, full history incl. the new migration:
```
SET ROLE windmill_user;
SELECT count(*) FROM workspace;                               -- 1
INSERT INTO workspace (id,name,owner) VALUES ('t','t','u');   -- INSERT 0 1
DELETE FROM workspace WHERE id='t';                           -- DELETE 1
INSERT INTO post_migration_table DEFAULT VALUES;              -- INSERT 0 1  (ALTER DEFAULT PRIVILEGES)
SET ROLE windmill_admin;
SELECT count(*) FROM script;                                  -- 1
SELECT nextval('audit_id_seq');                               -- 1          (sequence grants)
```
Abort-safety, tested directly:
- **Foreign-owned table+sequence in the schema**: blanket form aborts (`permission denied for table foreign_tbl`); this migration exits 0, grants only the runner-owned objects, leaves the foreign ones untouched.
- **Guard proof**: forcing every guarded grant to fail (target a non-existent role) produces one `WARNING` per statement, the block completes, `psql` exits 0 — no abort.
- **Idempotent** over 3 re-runs with **zero** warnings on a clean owned schema; reads still work after.
- **Locking**: each per-object GRANT takes only `AccessShareLock`; whole migration ~5ms. Does not block concurrent reads/writes.
- `init-db-as-superuser.sql` run verbatim as superuser (roles sed-renamed): exit 0, both roles have `USAGE`, admin has `BYPASSRLS` and inherits `windmill_user`.
- Applied to dev DB; backend restarts clean — `health check completed status="healthy" database_healthy=true`.

Non-`public` schema deployments are *not* a motivating case: migrating into a custom schema already fails independently at `20240216100535` on hardcoded `public.capture`.

No test committed: the effect is DB grant state, exercised by every existing `user_db`-backed path. No frontend changes.

### DIFF
--- a/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.down.sql
+++ b/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.down.sql
@@ -1,0 +1,3 @@
+-- No down migration: revoking these grants leaves the instance unable to run
+-- any query on a user_db transaction. 20250205131523, which this re-runs, is
+-- likewise a no-op down.
--- a/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.up.sql
+++ b/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.up.sql
@@ -22,10 +22,11 @@
 -- grant that cannot be applied is isolated and re-raised as a named WARNING, so
 -- the rest still run and the reason is visible in the migration log. The known
 -- failure modes this absorbs:
---   * an object the runner does not own -- a superuser-installed extension
---     (PostGIS's spatial_ref_sys) or a co-located application table. The loops
---     already filter to owned objects so these are skipped without even a
---     warning; the guard is the backstop.
+--   * an object the runner cannot grant on -- a superuser-installed extension
+--     (PostGIS's spatial_ref_sys) or a co-located application table owned by an
+--     unrelated role. The loops filter to relations the runner has authority to
+--     grant, so these are skipped without even a warning; the guard is the
+--     backstop.
 --   * an object dropped by another session between the catalog scan and the
 --     GRANT (only possible when something outside the migration shares the DB).
 --   * USAGE on a schema the runner does not own (needed only where USAGE was
@@ -50,14 +51,19 @@ BEGIN
     -- pg_class over the relkinds GRANT ... ON ALL TABLES covers -- ordinary (r),
     -- partitioned (p), views (v), materialized views (m), foreign (f). pg_tables
     -- would miss views such as flow_workspace_runnables, which are read through
-    -- user_db transactions and so must be granted too.
+    -- user_db transactions and so must be granted too. pg_has_role(...,'USAGE')
+    -- selects the relations the runner can actually grant: ones it owns
+    -- directly, inherits ownership of (tables still owned by a previous
+    -- migration role after a credential rotation), or reaches as a superuser.
+    -- Owner-name equality would miss the inherited-owner case and leave those
+    -- relations' user_db access broken.
     FOR obj IN
         SELECT format('%I.%I', n.nspname, c.relname)
         FROM pg_class c
         JOIN pg_namespace n ON n.oid = c.relnamespace
         WHERE n.nspname = target_schema
           AND c.relkind IN ('r', 'p', 'v', 'm', 'f')
-          AND pg_get_userbyid(c.relowner) = current_user
+          AND pg_has_role(current_user, c.relowner, 'USAGE')
     LOOP
         BEGIN
             EXECUTE format('GRANT ALL ON TABLE %s TO windmill_user, windmill_admin', obj);
@@ -69,7 +75,8 @@ BEGIN
     FOR obj IN
         SELECT format('%I.%I', schemaname, sequencename)
         FROM pg_sequences
-        WHERE schemaname = target_schema AND sequenceowner = current_user
+        WHERE schemaname = target_schema
+          AND pg_has_role(current_user, sequenceowner, 'USAGE')
     LOOP
         BEGIN
             EXECUTE format('GRANT ALL ON SEQUENCE %s TO windmill_user, windmill_admin', obj);
--- a/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.up.sql
+++ b/backend/migrations/20260720131744_regrant_schema_and_tables_to_windmill_roles.up.sql
@@ -1,8 +1,8 @@
 -- Re-runs the grants of 20250205131523. Everything in that migration sits in
 -- one DO block whose first statement is LOCK TABLE pg_catalog.pg_roles, which
 -- a non-superuser cannot take; on managed Postgres (RDS, Cloud SQL) it raises
--- and the block's EXCEPTION WHEN OTHERS handler downgrades the failure to a
--- NOTICE, so every GRANT after it is skipped.
+-- and the block's single EXCEPTION WHEN OTHERS handler downgrades the failure
+-- to a NOTICE, so every GRANT after it is skipped.
 --
 -- Most tables survive that anyway, because 20221105003256 grants them outside
 -- any such block. What is lost is the ALTER DEFAULT PRIVILEGES, which is what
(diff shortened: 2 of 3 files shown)


rubenfiszel:
✅ Review round clean @ a54492f63bf6caa24d551080fb3f445310da41e1

rubenfiszel:
Follow-up review: found that the previous `GRANT ... ON ALL TABLES IN SCHEMA` form raises a hard `permission denied for table X` (not a warning) if the schema contains any object the migration runner doesn't own — a superuser-installed extension (PostGIS `spatial_ref_sys`) or a co-located app table — which with no handler would **abort the whole upgrade**. That was a regression vs. `20250205131523`, which only tolerated it by swallowing every error. Reworked in bdd5aba3fb to grant per-object over just the runner-owned tables/sequences (structurally can't hit a foreign object, no swallowing). Re-verified: with a foreign-owned table+sequence in the schema the migration now exits 0 and grants only Windmill's own objects; full hardened non-superuser history still grants reads/writes/sequences/default-privs; idempotent over 3 runs.

rubenfiszel:
Follow-up on "we can't afford a migration that fails": audited every remaining way the DO block could raise and abort the upgrade, and closed them.

Found one residual abort path beyond the foreign-owned-object case: a `GRANT` on a table read from `pg_tables` that another session `DROP`s before the grant runs raises a hard `relation does not exist` (only reachable when something outside the migration shares the DB). Reworked in a01ba3eb06 so **every grant is wrapped in its own `BEGIN/EXCEPTION/WARNING`**. This is deliberately the opposite of `20250205131523`'s one block-wide `WHEN OTHERS` — the original's flaw was granularity (a single early failure skipped all remaining grants), not the presence of a handler. Here each failure is isolated, named in a `WARNING`, and the rest proceed.

Proven: forcing every guarded grant to fail (target a non-existent role) yields one `WARNING` per statement, the block completes, `psql` exits 0. Foreign-owned table+sequence in the schema → exit 0, only Windmill's objects granted. Full hardened non-superuser history still grants reads/writes/sequences/default-privs. Idempotent x3 with **zero** warnings on a clean owned schema (the guards only fire on genuine anomalies). The owner-filter stays so the common foreign-object case is skipped without even a warning; the guard is the backstop.

rubenfiszel:
Fixed in fc24cdf770. Confirmed `flow_workspace_runnables` is a view (relkind `v`) that `pg_tables` omits. Switched to `pg_class` over relkinds `r,p,v,m,f` (what `GRANT ... ON ALL TABLES` covers), keeping the ownership filter. Verified: an owned view + materialized view are now granted to both roles, a foreign-owned view is left untouched, and `windmill_user` can read `flow_workspace_runnables` after a full non-superuser migration run.

rubenfiszel:
Fixed the P1 in 6a540df784. Confirmed the credential-rotation scenario: with `role_new` inheriting `role_old` (neither a superuser), `pg_get_userbyid(relowner) = current_user` was false for all of `role_old`'s tables/views/sequences — so they were being skipped — while `role_new` can in fact grant on them via inheritance. Switched both loops to `pg_has_role(current_user, owner, 'USAGE')`, which selects objects the runner owns directly, inherits ownership of, or reaches as superuser. Verified: after rotation the migration now grants `role_old`'s table+view+sequence to both windmill roles, a table owned by an unrelated `stranger` role is left untouched, and the normal clean-schema run is still idempotent with zero warnings.

rubenfiszel:
Fixed the P1 in 522743f98b. Confirmed: `pg_has_role('postgres','appuser','MEMBER')` returns true for a superuser (implicit membership in every role), so as written a superuser runner would grant the windmill roles access to unrelated co-located objects. Replaced it with an explicit recursive `pg_auth_members` membership set (`owner_roles`) and grant only relations whose `relowner` is in it.

Verified both directions:
- Superuser runner (`postgres`, not a real member of `appuser`): grants its own `wm_tbl`/`wm_seq`, leaves `appuser`'s `app_tbl`/`app_seq` untouched.
- Credential rotation (`role_new` a real `INHERIT` member of `role_old`): still grants all of `role_old`'s table/view/sequence — the case owner-name equality missed.
- Normal non-superuser full history: reads/writes/views/sequences work; idempotent x3 with zero warnings.

PR #10213, opened 2026-07-20
fix(git-sync): avoid percent-encoded colon in git-sync hub script path

hugocasa:
## Summary

Git-sync repository detection could fail behind hardened reverse proxies / WAFs with a bare `400` before the request reached Windmill. The git-sync init/detection hub script slug contained a colon stored as `%3A` in the run-by-path URL. The generated API client re-encodes path params with `encodeURI`, turning `%3A` into `%253A` (double URL-encoding). Some proxies reject double-encoding as an attack signature and drop the request at the edge, so job creation never happened and detection stayed broken.

The hub resolves scripts by numeric id and ignores the slug (verified: all slug variants at id `28789` return byte-identical content), and the worker hub cache is keyed by `HUB_CACHE_DIR/{version}` (the id), not the slug. So dropping the colon from the slug produces a colon-free run URL that strict proxies accept, without changing which script runs or the cache entry.

The slug is **not** fully cosmetic, though: the EE `is_git_init_script` check in `get_github_app_token_internal` matched the job's `runnable_path` against the literal `%3A` slug to decide whether to grant a GitHub App token. That check is updated (in the companion EE PR) to match the colon-free slug while still accepting the legacy one, so GitHub-App-authenticated git-sync jobs keep working, including jobs enqueued by a pre-fix frontend during rollout.

No hub script needs to be renamed and no id changes, so existing pinned instances and the offline/cached-script path are unaffected.

**EE companion PR:** windmill-labs/windmill-ee-private#676

## Changes

- `frontend/src/lib/hubPaths.json`: drop `%3A` from the `gitInitRepo` slug (same id `28789`), so the browser's run-by-path URL no longer double-encodes to `%253A`.
- `backend/windmill-common/src/workspaces.rs`: matching colon-free slug for `GIT_SYNC_PULL_SCRIPT_PATH` + a comment recording why the slug must stay percent-encoding-free.
- `backend/src/main.rs`: force-cache `GIT_SYNC_PULL_SCRIPT_PATH` at build alongside `LATEST_GIT_SYNC_SCRIPT_PATH`, so the backend-driven pull script is always baked into the image for airgapped workers (independent hardening).
- **EE (windmill-ee-private#676)** `windmill-common/src/git_sync_ee.rs`: `is_git_init_script` now matches the colon-free slug and the legacy `%3A` slug; matcher extracted into a pure `is_git_init_script_path` helper with a regression test.
- `backend/ee-repo-ref.txt`: bumped to pull in the EE change.

## Test plan

- [x] `is_git_init_script_path` unit test covers colon-free + legacy `%3A` slugs (any id), rejects the test-read-write script and non-hub paths.
- [ ] On a GitHub-App-authenticated instance, run git-sync detection / pull / push from the UI and confirm the job obtains its GitHub App token and the hub script at id `28789` resolves; ideally through a proxy that rejects double-encoded `%3A`.
- [ ] On a fresh backend build, confirm `cache_hub_scripts` logs caching both `hub/28789/git-sync-init-repository-windmill` and `hub/28790/...`.

## Notes

- `hubPaths.json` is under `frontend/` but has no visible UI effect (config string only); the detection flow renders identically, so no screenshots.


hugocasa:
✅ Review round clean @ 332d1b255fd1ae3e35486975a5686fe6a398b90c

PR #10215, opened 2026-07-20
feat: attach text files to chat messages, read on demand via file tools

Guilhem-lm:
## Summary

Standalone text files attached to the GLOBAL AI chat are now **message-scoped attachments**, matching the image UX: chips inside the input box, sent with the next message, cleared after send. On the wire they follow the DOM-pick pattern rather than the image one — the prompt carries only a **reference** (name, line/char counts), and the content is registered into the session attached-files store where the model reads it on demand via `read_file` / `search_files`. One behavior for every file size, and content is never resent with the whole history on later turns.

The governing rule after this PR: **the input box shows only what ships with the next message (text, images, files, DOM picks, @context); the footer next to the mode picker holds session-wide assets (linked folders).**

## Changes

- **New `textFileUtils.ts`** — `AttachedTextFile`, caps (8 files/message, 1 MB sanity cap), sniff-and-read helper.
- **Composer (`AIChatInput`)** — `files`/`pendingFiles` state mirroring images (slot reservation, send held while reading); file chips render in the same padded row as image thumbnails; image chip row previously had no padding (`px-2.5 pt-2` now).
- **Attach routing (`AIChatDisplay`)** — picked/dropped/pasted top-level text files go to the composer; folders (and their nested children) stay session assets. `AttachedFilesBar` moved from inside the input box to the footer row next to the mode picker.
- **Paste (`ContextTextarea`)** — new `onTextFiles` callback beside `onImageFiles` so OS-clipboard file pastes attach.
- **Manager (`AIChatManager`)** — `files` threaded through `sendRequest`, the message queue, every restore path (refusal, beforeSend failure, cancel, empty-turn rollback), and edit/retry. Registration happens at send with **collision-safe naming**: a name clash with a session-linked file registers under a suffixed name and the message reference/chip is rewritten to match.
- **Transcript-driven registry sync** — message-file rows are rebuilt from `DisplayMessage.files` on chat load and pruned on clear/rollback/edit-retry truncation (`syncMessageScoped`), so a visible chip is always readable and a dropped message's file never lingers in the tool surface. Message rows are deliberately not persisted in IndexedDB — the transcript is their durable home.
- **Prompt (`prepareGlobalUserMessage`)** — `## ATTACHED FILES` reference listing with an explicit read-before-answering instruction; the file-tools roster also lists message-attached files.
- **Shared badge** — composer and bubble file chips render via `ContextElementBadge` (new `attached_file` element): same visual language as every other context chip, per-extension file icon, content-preview popover.
- **Display** — bubble file chips expand/collapse content inline; queued-message card shows queued file names; edit box reseeds files from the message snapshot.

## Test plan

- [x] Unit tests: cross-scope name collision (suffixed registration, footer removal isolation), `syncMessageScoped` reconciliation, message-scoped visibility/replacement, file-only queue/dequeue, reference-only prompt shape (content must never inline)
- [x] Browser (session chat): attach via picker/drop/paste → chips render padded in composer; send → model calls `read_file` and answers from content; composer clears; folder chip stays in footer
- [x] Browser: hard reload mid-conversation → transcript sync rebuilds registrations, `read_file` still resolves the file
- [ ] Reviewer sanity pass: attach a session file and a same-named message file, confirm both remain addressable

## Screenshots

Composer — image thumbnail and a file chip rendered with the shared context badge (per-extension icon, hover-X, content-preview popover); folder chip in the footer next to the model picker:

composer-badges

Send flow — model reads the attached file via the file tools and answers from its content:

reference-flow

Sent-message bubble: image thumbnail, file chips, expanded file content preview:

bubble-file-expanded


Guilhem-lm:
Addressed the round's findings (commits `0077462d31`…`5ec27217f4`):

- **Same-name attachments across messages** (codex P1-1 / claude P2-1): a same-named file with different content now registers under a suffixed name (never replaces), and the rename is propagated into the message's prompt reference and chip. Identical re-registration (retry/sync rebuild) reuses the row. Regression tests added.
- **Compaction** (codex P1-2): rather than pruning, the compaction summary now *carries the folded-away messages' files* — they ride `SummaryDisplayMessage.files` (reload-safe through the same transcript sync), the API summary lists them as still-readable, and the boundary UI shows their badges. Bounded by the conversation budget below.
- **Unbounded storage** (codex P1-3): new `MAX_CONVERSATION_FILE_BYTES` (5 MB) enforced at attach time against transcript + queue + composer bytes, with a toast pointing at folder-linking. A blob-store dedupe like images have would be a further improvement, but with a hard conversation bound the record size is deterministic; happy to do the blob-store as a follow-up if preferred.
- **File-only history title** (codex P2-4): falls back to the first attached filename, like image-only chats.
- **Roster**: message-attached files are no longer listed in the standing `## Attached files` system-prompt section — their reference lives in the message that carried them (or the compaction summary), exactly like DOM picks. The file tools are registered unconditionally in GLOBAL, so readability doesn't depend on the roster.

**Dismissing** (claude P2-2, non-GLOBAL drop is a silent no-op): every surface that reaches `addTextFiles` is already GLOBAL-gated — panel drag/drop and the hidden pickers via `canAttachFiles` (`mode === GLOBAL && !disabled`), clipboard paste via the mode ternary on `onTextFiles`. The early return is defense in depth, not a reachable path.

On the **ai_evals A/B** note: the global prompt and tool roster are byte-identical when no files are attached — the `## ATTACHED FILES` section renders only for messages that carry attachments, and the roster change only removes lines that never existed before this PR. Baseline evals are unaffected by construction.

Also in this push: composer/bubble file chips now render through the shared `ContextElementBadge` (new `attached_file` element with a content-preview popover) instead of bespoke markup.

Guilhem-lm:
✅ Review round clean @ 848afaa98c5be991a0993bf3df0890566843389c — nit-only verdicts at ce930b6b5019a06294872439c881152db74edbd3 (codex: duplicate badge-list keys for reused attachment names; claude/pi: good to merge); nit addressed in 848afaa98c (identical double-attach dedupes at the composer, summary carry dedupes by name, badge list keys include the index, regression test extended).

Guilhem-lm:
✅ Review round clean @ f3aef58a6d6066635ce50f29c92d54195c8019a0 — nit-only verdicts at f0bd6691321d8bc6182d133081eee9dd5f7f5580 (codex: overlapping reads could double-spend the byte budget; claude/pi: good to merge); nit addressed in f3aef58a6d (in-flight reads now reserve their bytes against the budget, released on completion — mirroring the slot reservation).

rubenfiszel:
Addressed the Codex P2 (cross-composer attachment budget) in `e8187e5c9a`.

The bottom composer and the edit box are both mounted during an edit, and each was enforcing `MAX_CONVERSATION_FILE_BYTES` against only its own staged files plus the transcript, so two attaches near the cap could each pass and overflow the persisted transcript.

Fix moves staged-byte accounting into shared manager state:
- Each `AIChatInput` publishes its staged bytes (committed attachments + in-flight reads) to `AIChatManager` via `setComposerStaged(key, editingIndex, bytes)`, keyed per instance, cleared on unmount.
- The attach-time budget now calls `attachmentBytesExcluding(selfKey)`, which sums transcript + queue + **every other live composer's stage**. A message an open composer is editing is skipped from the transcript sum (that composer's stage stands in for it), so no double-count.

The formula is symmetric: both composers converge on the same remaining budget, so the concurrent-edit overflow window is closed regardless of attach order.

Regression test added in `AIChatManager.test.ts` (`counts every other live composer stage in the attachment budget`) covering the mid-edit two-composer case and edited-message exclusion. Full `AIChatManager.test.ts` suite green (105 passed).

rubenfiszel:
Addressed the Codex round-9 P2 (edit/cancel bypasses the byte cap) in `3343678b37`.

The prior fix excluded a message being edited from the transcript sum and substituted the edit box's stage. But an edit isn't committed until send, so cancelling it returns the persisted attachments. If the user removed the near-limit files in the editor (stage → ~0), filled the always-mounted bottom composer with replacements, then cancelled, both batches persisted and the 5 MB cap was exceeded.

`attachmentBytesExcluding` now charges a message **another** composer is editing at `max(persisted size, editor stage)`, not just the editor's in-progress stage. Freed space becomes available only once the edit actually commits (the persisted bytes drop). The editing composer's own budget is unchanged — its message is still excluded and its stage counted by the caller, so it can refill up to the cap.

Regression test added (`charges an edited message at its persisted size until the edit commits`): editor removes a message's files (stage 0) → the bottom composer still sees the full persisted bytes; editor stages more than the original → the larger figure wins. `AIChatManager.test.ts` green (106 passed).

rubenfiszel:
Addressed the round-10 P1 and P2.

**P1 — transcript filename preserved on rebuild** (`attachedFiles.svelte.ts`)

`syncMessageScoped` rebuilt message rows through collision-suffixing `addFiles`. Session rows load first (on restore), so a session asset holding `notes.md` pushed the rebuilt message row to `notes (2).md` while the persisted prompt still referenced `notes.md` — `get('notes.md')` then resolved to the session asset and the model read the wrong content.

Fix: free the name from the conflicting session row before the rebuild (`#freeNameForMessageRow`), so the message row reclaims its exact, immutable reference. The rename is in-memory only — deterministic and re-applied on every load, and the session roster is regenerated live each send, so the session asset stays addressable under the suffix. Both files remain independently readable (`get('notes.md')` → message content, `get('notes (2).md')` → session content).

Regression test: `syncMessageScoped keeps a message file on its exact name despite a session clash`.

**P2 — resent files reserved across the restartGeneration gap** (`AIChatManager.svelte.ts`)

The edit box unmounts (dropping its staged-byte entry) the instant the user submits, but `restartGeneration` then awaits registry sync + `beforeSend` before the optimistic bubble lands in the transcript. During that window the resent files were reserved nowhere, so the bottom composer could attach into the temporary headroom and overflow the cap.

Fix: `restartGeneration` reserves the resent files' bytes in shared manager state before the transcript slice; `sendRequest` releases the reservation the moment it installs the bubble (and on the pre-install non-GLOBAL bail, which restores the files to the composer). The reservation bridges the whole gap.

Regression test: `reserves resent files across the restartGeneration gap` (hooks `refreshFolders`, the one point mid-`sendRequest` after the slice but before install, and asserts the resent bytes stay accounted, then released once the bubble installs).

`AIChatManager.test.ts` (137) and `attachedFiles.test.ts` green. The 6 failing tests in `ContextManager.test.ts`/`app/core.test.ts` are pre-existing (fail on HEAD without these changes) and unrelated.

rubenfiszel:
Addressed the three round-12 findings.

**P1 — attachment references through fallback compaction** (`AIChatManager.svelte.ts` + `fileTools.ts`)
Drop-oldest compaction (summary fallback) removes API messages without a summary, so a folded message's `## ATTACHED FILES` reference stops reaching the model even though the file stays readable — and the roster deliberately omits message-scoped files. Added `orphanedMessageFileNames()`, which finds message files whose only referencing user message went negative-index (API counterpart dropped); the roster now advertises exactly those. Files still referenced by a surviving message, and files carried on a compaction summary, are not re-listed. Test: `flags message files whose referencing message was dropped by compaction`.

**P1 — indexing preserved when renaming a restored session row** (`attachedFiles.svelte.ts`)
`#freeNameForMessageRow` renamed a restored session row while its `#indexFile` (started under the old name) was still in flight, so `#patchFile(oldName, file)` no longer matched and the row stayed `indexing` forever (read_file rejects, search_files excludes). Now re-kicks `#indexFile` under the new name when the renamed row is still indexing; the stale completion no-ops. Test (using the existing manual-index harness): `keeps a session row indexable when sync renames it mid-index` — syncs *during* active indexing, no `settle` masking the race.

**P2 — resend reservation released on pre-install command returns** (`AIChatManager.svelte.ts`)
An edit resolved to `/clear` or `/compact` runs the local-command path and returns before a bubble installs, so the `#RESEND_KEY` reservation stayed charged and blocked later attachments. Routed all clears through `#releaseResendReservation()` and call it on every sendRequest path that exits before install (mode-hidden, empty, both local commands, non-GLOBAL refusal) as well as at install. Test: `releases the resend reservation when an edit resolves to a local command`.

Green: `AIChatManager.test.ts`, `attachedFiles.test.ts`, `fileTools.test.ts`, `global/core.test.ts` (roster). `svelte-check` clean on all changed files.

rubenfiszel:
Addressed cubic's P2 (shared resend-reservation key) in `e8da3b196d`.

The reservation used a single shared `resend-reservation` key, so any `sendRequest` — a normal send, or a concurrent one — released it at its own install / early-return even though it didn't own it, dropping an in-flight resend's reservation and letting attachments staged before the resend bubble lands under-count against the cap.

Fix: `restartGeneration` now mints a per-resend token (`resend:${createLongHash()}`), reserves under it, and threads it through `sendRequest` as `resendReservationKey`. `#releaseResendReservation(key)` acts only on that key, so a send carrying no token (every normal send) releases nothing.

Regression test: `a normal send does not release another send resend reservation` (an in-flight `resend:other` reservation survives a normal early-returning send). `AIChatManager.test.ts` / `attachedFiles.test.ts` / `fileTools.test.ts` green (145).

rubenfiszel:
Addressed the two round-13 P1s.

**P1 — registry rebuild retargets multi-name attachments** (`attachedFiles.svelte.ts`)
Freeing a session row for the rebuild suffixed it against current rows only. With the transcript referencing both `notes (2).md` and `notes.md` and a session row named `notes.md`, freeing `notes.md` renamed it onto `notes (2).md` — itself a wanted reference — so that message row cascaded to `notes (3).md` while its persisted reference stayed `notes (2).md`, and `read_file("notes (2).md")` returned the session file. `#uniqueName` now takes a reserved set; the rebuild frees each session clash clear of the **entire** wanted-name set, so every message row reclaims its exact reference regardless of collision order. Test: `reclaims every colliding transcript name across a multi-name rebuild`.

**P1 — fallback compaction orphans summary-carried files** (`AIChatManager.svelte.ts` + `shared.ts`)
`orphanedMessageFileNames()` treated summary rows as always-live, but a summary carries its folded files' reference on its own API message. When a summary compaction succeeds and a later summarization fails, drop-oldest can drop that API message — leaving the files off the roster and invisible to the model. The summary display row now tracks its API `index` (slot 0 at creation, re-based by drop-oldest); a negative index reads as "counterpart gone" and its files move to the roster, mirroring user-message orphans. The index is used **only** for orphan detection (all restart/storedImages readers filter `role === 'user'`), never as a restart target, and is absent on summaries loaded from pre-existing history. Test: `orphans summary-carried files when drop-oldest removes the summary`.

All 4 suites green (`AIChatManager`, `attachedFiles`, `fileTools`, `global/core` — 297). `svelte-check` clean on changed files.

Re: the `ai_evals` A/B note — the only model-facing change here is that a compaction-orphaned attachment is now listed in the roster (a strict improvement in a rare compaction path); happy to run a benchmark if you'd like, but it doesn't change steady-state prompt shape.

rubenfiszel:
Addressed the round-14 P2 (outgoing bytes unreserved during send preflight) in `b8058f8dfa`.

The resend reservation covered edit/retry, but Codex correctly flagged the same gap for **normal and queued sends**: the composer (or queue) clears its `files` the instant `sendRequest` is called, dropping the staged-byte entry, while `sendRequest` then awaits `regrantLocked()`/`refreshFolders()` before the bubble lands. With a locked or slow linked folder the composer stays enabled, so a fresh drop could spend the same headroom and overflow the cap once the first bubble installs.

Fix generalizes the reservation rather than special-casing resends: `sendRequest` mints a per-send token and reserves the outgoing files' bytes **just before** the attachment-upkeep awaits (reusing `restartGeneration`'s token when present, so resends keep their earlier coverage), and releases it on install or any pre-install exit. Covers composer sends, queued auto-sends, and resends under one path. Renamed the helper to `#releaseOutgoingReservation` to match.

Regression test: `reserves a normal send outgoing files across the preflight gap` (hooks `refreshFolders`, asserts the bytes are reserved mid-gap and released once the bubble installs). Suites green (148).

Re-triggering codex/pi via command since the last synchronize round got skipped by concurrency from rapid pushes.

Guilhem-lm:
Addressed the latest round's findings in `3e19e8b42b` (on top of the stable-id refactor `7ff357530e` + `b12e0ff631`):

- **P1 — unscoped search loses stable ids**: hit lines now label id-bearing rows as `name (file id: …)`, so every hit carries the reference that resolves back to exactly the row that produced it. Regression tests added (`fileTools.test.ts`): unscoped search across two same-named rows asserts each hit's id label; id-scoped search asserts isolation.
- **P2 — overlapping attach batches bypass dedupe/rename**: normalization (identical-dedupe, courtesy rename, id minting) moved out of the read loop into the single synchronous commit step against the live list (`foldIntoDraft` in `textFileUtils.ts`, unit-tested), so a batch committed between another batch's file reads can no longer be missed.

On the `ai_evals` note: the global-mode prompt and tool schemas are byte-identical whenever no files are attached (the `## ATTACHED FILES` block and roster lines only render for attachment-bearing turns, which the eval suite does not exercise), so a before/after run would measure noise. The attachment flows themselves were verified in-browser: same-named session/message files independently readable by name/id, id-scoped search returning only its own row's hits, and reload rebuild of colliding rows.

Guilhem-lm:
✅ Review round clean @ e12ceebd14d990954363aad28b8193b83873e90a — nit-only verdicts at 3e19e8b42b (codex: identical re-drop after a courtesy rename escaped dedupe; claude: queued-chip truncation; pi: good to merge); nits addressed in e12ceebd14 (base-name-aware dedupe in foldIntoDraft + regression test, min-w-0 on the queued file chip label).

Guilhem-lm:
Addressed cubic's five P2s in `41bf80d617`:

- **Line-count mismatch** (`core.ts`): the prompt now uses `textLineCount`, matching `read_file`'s numbering exactly (0 lines for empty files, no phantom line after a trailing newline). Pinned against the tool's semantics.
- **Chip click opened edit mode** (`AIChatMessage.svelte`): the badge row stops click/keydown propagation, so a file chip opens its content-preview popover without the message wrapper's click-to-edit unmounting it. Browser-verified both ways (chip → popover only; message body → edit box still opens).
- **Composite hit labels not resolvable** (`fileTools.ts`): rather than changing the hit format, `resolve()` now accepts the printed `name (file id: x)` label verbatim — models echo references exactly as printed, so the printed form is a valid reference for `read_file` and `search_files`. Test added.
- **1 MB cap not enforced at the reader** (`textFileUtils.ts`): `fileToAttachedTextFile` rejects oversized input itself (raw size pre-check + decoded `textByteLength` post-check, since malformed UTF-8 can expand on decode).
- **Rename-provenance heuristic** (`textFileUtils.ts`): replaced the suffix-stripping base-name compare with an explicit `sourceName` recorded only when a courtesy rename actually happened — a user's real `report (2).md` with content identical to `report.md` is no longer swallowed. Regression tests cover both directions.

✅ Review round clean @ 41bf80d6174f6f1b9ae31dc637ad6927fd9cf8bf — nit-only follow-ups since the round at 3e19e8b42b (codex/claude nits in e12ceebd14, cubic P2s in 41bf80d617); all addressed, no blocking findings outstanding.

Guilhem-lm:
Addressed the two round P2s in `4f8ec5c1de`:

- **Literal composite filenames**: `resolve()` now runs exact-id and exact-name lookups *before* label interpretation, so a file literally named `notes (file id: missing)` stays addressable by its exact name; label parsing only applies when nothing matches verbatim. Regression: literal-name file plus a base-named sibling, exact name wins.
- **Decoded bytes vs the conversation cap**: the synchronous commit step re-checks the live budget against decoded UTF-8 sizes (`admitWithinByteBudget`, pure logic + unit test) and drops what no longer fits with the budget toast — a malformed-UTF-8 tail can no longer inflate past the 5 MB cap after passing the raw-size admission. Companion regression pins `fileToAttachedTextFile` rejecting a file whose decoded content exceeds the per-file cap (valid 9 KB ASCII prefix + 400 KB invalid tail: raw < 1 MB, decoded > 1 MB).

✅ Review round clean @ 4f8ec5c1deb9523de784d6d0dbafb40e93090b2d — nit-only verdicts at 41bf80d617 (codex: literal composite names, decoded-byte cap; claude/pi: good to merge); nits addressed in 4f8ec5c1de.

Guilhem-lm:
Addressed the round P2 in `a2aa723dd9`: files folded into the queued message now pass through the same commit normalization as the composer (`foldIntoDraft` against `queuedFiles`) — an identical re-attach dedupes instead of eating a slot, a same-name clash gets the courtesy rename, and distinct files can no longer be displaced by duplicates at the eight-file cap. Manager regression added: identical + same-name files across two submissions assert dedupe, rename, and retention.

✅ Review round clean @ a2aa723dd923f0eba475f6765b106ab8b84dbe99 — nit-only verdict at 4f8ec5c1de (codex: queued-file aggregation bypassed normalization); nit addressed in a2aa723dd9.

Guilhem-lm:
Addressed the round's two P2s in `eb8778a388`: fold provenance now composes across composer → queue → dequeue (sourceName rides every pass-through, dedupe matches on it — regression pins the full chain), and dequeued files fold into a populated draft via the same normalization instead of raw concatenation.

✅ Review round clean @ eb8778a3888278dc5fcb98ce6f9ec62257bc1690 — nit-only verdicts at 7275804723; nits addressed in eb8778a388.

Guilhem-lm:
Local Codex pass on the MessageDraft refactor: nit-only (stale drop-routing comment, overlong budget doc comment, stray QA screenshot at the repo root) — all three addressed in `fca6665776` / cleaned up.

## Screenshots

Compaction boundary label (`text-normal text-2xs`) with the carried-files note:

compaction-label-normal

Same-named message attachments across messages, independently readable by id (`this: CHARLIE-33`, `first: ALPHA-11`) — the stable-id behavior end-to-end:

stable-id-session2

Guilhem-lm:
✅ Review round clean @ fca6665776a1b14f78deaec7de6121f6a90f8d12 — verdicts at a28b86f363 (codex: good to merge; claude/pi: nit-only, the stale drop-routing comment); nit addressed in fca6665776.

Guilhem-lm:
Addressed cubic's four P2s in `bef9c0107f`:

- **Prompt injection via filenames**: `sanitizeAttachmentName` strips control characters at attach (`fileToAttachedTextFile`) and defensively at every model-facing render site (`## ATTACHED FILES`, the roster lines, the compaction carry note) — legacy names predate the attach-time pass. Regressions pin the sanitizer and a crafted `\n## INSTRUCTIONS:` name staying on one prompt line.
- **Dropped files landing on the wrong message**: the drop handler now takes a send hold (`holdSendForIngestion`) before its first await and releases it after routing, so a send can't slip between the drop and `addTextFiles`. Holds also count as composer occupancy for restores. Synthetic-drop sanity-checked in the browser.
- **Queue restore clobbering a preflight follow-up**: `#restoreQueue` and the unmounted-input requeue now merge via `MessageDraft.prepend` (taken entry's text above the newer follow-up) instead of replacing. Regression: a follow-up queued inside a rejecting `beforeSend` survives alongside the restored message.
- **Stale restart index leaking the resend reservation**: `restartGeneration` resolves and validates the API restart point before reserving bytes or truncating either transcript — a stale index now fails with nothing mutated. Regression asserts transcript intact and no lingering reservation after the throw.

✅ Review round clean @ bef9c0107fb20ac3199b33b1e170e1b635d421f1 — cubic P2s addressed on top of the clean round at a28b86f363.

Guilhem-lm:
Addressed the round's two P2s in `bf86fc0768`: restored drafts now keep chronological priority at the caps (the newest additions drop, never the taken entry's attachments) with both pinned contexts unioned on restore; and session filenames are sanitized at every store row-creation site — including pre-sanitization persisted rows on restore — so the advertised reference is always the stored, resolvable name. Regressions cover cap-priority prepend, dual-context restore, and a control-char session filename resolving by its advertised name.

✅ Review round clean @ bf86fc0768c34cff7647060021513ed525ee27e6 — nit-only verdicts at 2d99d9107c (codex: restore priority + sanitized-name resolution; pi: good to merge); nits addressed in bf86fc0768.

Guilhem-lm:
Addressed the round's two P2s structurally rather than as point patches — an architecture pass over name handling showed the rules living as convention at seven creation sites (four verbatim copies, two skipping uniquify, one comparing raw against sanitized). `068a47b1c8` introduces one seam per scope: `#claimName` (sanitize + uniquify) is now the only way a session row gets its display name — restore and placeholders included, which fixes the legacy-collapse P2 — and `addFiles` derives the sanitized name once so the dedupe compare sees what is stored, fixing the re-link P2. `foldIntoDraft` sanitizes its own reads (self-contained draft choke), and folder names — deliberately raw disk keys — are sanitized at their model-facing render sites. `relPath` and `folder` remain raw by design (disk/grouping keys). Regressions: control-char re-link dedupe, legacy sanitize-collision restore staying distinct, fold self-sanitization, folder-line render closure.

✅ Review round clean @ 068a47b1c8f1f2ca0596aa9e906744ab217da5f5 — nit-only verdicts at bf86fc0768 (codex/pi: sanitize-vs-dedupe compare, restore uniquify); addressed in 068a47b1c8.

Guilhem-lm:
Addressed the round P2 in `beaf8298ea`: dedupe now means "the same file re-linked" (matching stats plus name / pre-suffix sourceName / raw path) — a bare display-name collision links the second file under a claimed suffix instead of silently discarding it, matching the restore path's behavior. `#claimName` records the pre-suffix name so re-linking a suffixed file's original still dedupes. Regression: two distinct raw names sanitizing identically both survive with unique names, and re-links of either dedupe.

✅ Review round clean @ beaf8298ea16b5d2a57a799f519219be1be69add — nit-only verdict at 068a47b1c8 (codex: sanitized-name collision dropped distinct files); addressed in beaf8298ea.

Guilhem-lm:
Addressed both round P2s in `fdc3726c2c`: the re-link identity is now the RAW pre-sanitization name — recorded as `sourceName` on every session row, carried on the persisted record, and re-derived on restore. Dedupe matches stats plus raw identity and never compares display names, so equal-stat sanitize-collisions link as distinct suffixed rows, and re-linking the original behind a suffixed row still dedupes after a reload. Regressions cover exactly Codex's two scenarios (equal size+mtime collision; post-restore re-link).

✅ Review round clean @ fdc3726c2cc8ad956f801765bf28a4e0cc064f28 — nit-only verdicts at beaf8298ea (codex: equal-stat collision, provenance not persisted); addressed in fdc3726c2c.

PR #10217, opened 2026-07-20
fix(sessions): session bar badge readouts, job persistence, refresh bounce

Guilhem-lm:
## Summary

A batch of session-bar polish and reliability fixes: the badges now say what actually needs attention, background-job state (including its "already seen" status) survives page reloads, and hard-reloading a session URL no longer bounces to another session.

## Changes

- **Edits token counts drafts**: while any drafts are pending the token reads "N draft(s)" (counting only drafts — deployed items no longer inflate the number); once none remain it turns green and reads "N deployed". Popover rows now lead with the shared `RowIcon` kind icon so scripts/flows/apps/triggers are distinguishable at a glance.
- **Jobs chip relaxes once reviewed**: the priority status readout (approval > running > queued > failed > succeeded) is computed only over jobs whose terminal outcome the user hasn't seen yet. Opening the popover marks the visible finished jobs reviewed (persisted `reviewed` flag on `ChatJob` via the new `markJobsReviewed`); when nothing unreviewed remains the chip shows a neutral gray "N jobs executed". A job finishing while the popover is closed re-raises the status until reviewed — but an already-seen failure never resurfaces just because a later job ran.
- **Inline-completed jobs survive reload**: `updateJob` now persists the tray on any non-terminal→terminal transition. Previously only detached jobs (or poller-observed completions) were persisted, so a job that started and finished within the inline wait vanished from the Jobs segment on refresh.
- **Session URL survives hard reload**: the sessions page's family-reconcile effect no longer runs before `listUserWorkspaces` has returned. `workspaceRootId` falls back to the raw id for workspaces it can't find in a not-yet-loaded list, which made same-family sessions look foreign on refresh and bounced the URL to the most-recent (or a brand-new) session. The intentional cross-family redirect still fires once the list arrives.

## Screenshots

Unreviewed failure (priority readout) alongside a pending draft:

bar-1draft-1failed

Same bar after opening/closing the Jobs popover — relaxed to the executed count:

bar-1draft-2executed

All edits deployed — green "N deployed" label:

bar-2deployed-2executed

Edits popover rows with kind icons:

edits-popover-icon

## Test plan

- [x] Unit: new regression test in `AIChatManager.test.ts` pins both persistence transitions (inline terminal transition writes the tray; `markJobsReviewed` persists `reviewed: true` once, no-op on re-review) — 103/103 passing
- [x] Browser: 1 success + 1 failure → chip "1 failed"; open+close popover → "2 jobs executed"; state survives hard reload
- [x] Browser: run a job after a reviewed failure → chip shows "1 succeeded", not the old failure
- [x] Browser: inline job (~0.1s) + reload → job still listed in the popover with label and duration
- [x] Browser: pending draft + deployed item → "1 draft" (indigo); deploy all → "N deployed" (green); popover rows show kind icons
- [x] Browser: repeated hard reloads of `/sessions?session_name=…` keep the URL and session (previously bounced and created phantom sessions)

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -627,8 +627,26 @@ export class AIChatManager {
 	updateJob = (jobId: string, update: Partial<ChatJob>) => {
 		const idx = this.backgroundJobs.findIndex((j) => j.jobId === jobId)
 		if (idx === -1) return
+		const wasTerminal = !this.isJobNonTerminal(this.backgroundJobs[idx].status)
 		this.backgroundJobs[idx] = { ...this.backgroundJobs[idx], ...update }
 		this.backgroundJobs = [...this.backgroundJobs]
+		// Persist on the transition to terminal: a job that completes inside the
+		// inline wait never hits the detach/poller persist paths, and would
+		// otherwise vanish from the tray on reload.
+		if (!wasTerminal && !this.isJobNonTerminal(this.backgroundJobs[idx].status)) {
+			void this.#persistBackgroundJobs()
+		}
+	}
+
+	/** Mark finished jobs as reviewed (their terminal status was shown in the
+	 * jobs popover) and persist, so the chip stays relaxed across reloads. */
+	markJobsReviewed = (jobIds: string[]) => {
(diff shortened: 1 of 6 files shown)


Guilhem-lm:
✅ Review round clean @ 37520309da6f83cda1eef3777808ac320caf2e93

PR #10220, opened 2026-07-21
fix(triggers): apply scope-path filtering to list_triggers and fix update_trigger scope check

hugocasa:
Fixes WIN-2211

## What

Brings the shared trigger CRUD handler in line with scripts, flows, apps, resources, variables and schedules for scoped-token path handling.

- **`list_triggers<T>`** now filters returned rows through `build_scope_path_predicate(&authed, T::scope_domain_name(), "read")`, applied after the draft-only append so both deployed and draft rows are covered. Previously the list endpoint ignored the token's declared scope path, unlike `get_trigger`/`create_trigger`/`delete_trigger`/`exists_trigger` which already call `check_scopes`. Covers all 10 `TriggerCrud` kinds (HTTP, WebSocket, Kafka, NATS, MQTT, SQS, GCP, Azure, Postgres, Email).
- **`update_trigger`** now checks scope against the existing path (URL param) in addition to the new path (request body). It previously only checked the body path.

## How

A small `HasPath` supertrait is added to `Self::Trigger` so the shared handler can read the row's path without each impl restating it: `Trigger<T>` returns `&base.path`, the `()` OSS stub returns `""`. This keeps the change to two files (`windmill-trigger/src/{types,handler}.rs`) rather than editing every per-kind impl.

## Validation

- `cargo check` on `windmill-trigger` and all 10 impl crates (OSS + EE variants via `--features private,enterprise`) passes.
- No SQL query changes, so no sqlx cache update.
- No frontend changes.

Following the precedent of the analogous schedules fix (#10192) and the scripts/flows/apps/resources/variables endpoints, no dedicated per-endpoint scope test is added — all rely on the shared, already-tested `build_scope_path_predicate`.

---
## Summary by cubic
Applies path-based scope filtering to trigger listing and checks both old and new paths on update to prevent unauthorized access and moves. Fixes WIN-2211 and brings triggers in line with other scoped endpoints.

- **Bug Fixes**
  - `list_triggers<T>` now filters rows via `build_scope_path_predicate(&authed, T::scope_domain_name(), "read")` after draft overlay, across all `TriggerCrud` kinds.
  - `update_trigger` now verifies write scope on both the existing path (URL) and the new path (body).
  - Added `HasPath` on `Self::Trigger` so the list handler can read each row’s path (`Trigger<T>` returns `&base.path`, OSS stub returns "").

<sup>Written for commit f29834680253b9301da84f3686af8cc443eab26f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-trigger/src/handler.rs
+++ b/backend/windmill-trigger/src/handler.rs
@@ -6,13 +6,13 @@
  * LICENSE-AGPL for a copy of the license.
  */
 
-use crate::types::{StandardTriggerQuery, TriggerData, TriggerMode};
+use crate::types::{HasPath, StandardTriggerQuery, TriggerData, TriggerMode};
 use async_trait::async_trait;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use sql_builder::{bind::Bind, SqlBuilder};
 use sqlx::{FromRow, PgConnection};
 use std::fmt::Debug;
-use windmill_api_auth::{check_scopes, ApiAuthed};
+use windmill_api_auth::{build_scope_path_predicate, check_scopes, ApiAuthed};
 use windmill_common::{
     db::UserDB,
     error::{Error, JsonResult, Result},
@@ -65,6 +65,8 @@ pub trait TriggerCrud: Send + Sync + 'static {
         + Send
         + Sync
         + Unpin
+        // Path accessor so `list_triggers` can apply scoped-token filtering.
+        + HasPath
         // `'static` so the deployed trigger can be boxed into
         // `WithDraftOverlay`'s erased-serde inner (it's an owned row).
         + 'static;
@@ -662,6 +664,9 @@ async fn list_triggers<T: TriggerCrud>(
         }
     }
 
+    let allowed = build_scope_path_predicate(&authed, T::scope_domain_name(), "read");
+    triggers.retain(|t| allowed(t.trigger_path()));
+
     Ok(Json(triggers))
 }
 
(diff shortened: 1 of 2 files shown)


hugocasa:
Round 1 verdicts: claude / pi / local-review all **Good to merge**; codex raised one P1 (scope filtering runs after SQL `LIMIT/OFFSET`). I'm **not** fixing it in this PR, for the same reasons the identical finding was resolved-by-explanation on the sibling schedules fix (#10192, merged):

- **Mechanism is real but fail-closed.** Filtering after pagination can hide *authorized* rows from a scoped token on a short/empty page; it can never leak *unauthorized* rows. The enumeration hole this PR closes (the security objective) is unaffected.
- **Practically narrow.** `DEFAULT_PER_PAGE` is 1000 (`windmill-common/src/utils.rs:36`), so a short/empty first page requires either an explicit small `per_page` or >1000 out-of-scope triggers ahead of an in-scope one.
- **Not trigger-specific.** Scripts, flows, apps, resources, variables and schedules all paginate in SQL and filter in Rust the same way. Pushing scope into the SQL predicate is the correct fix but it's a cross-cutting change to the shared pattern, out of scope for WIN-2211 (closing the enumeration hole). Better as a follow-up against the shared helper than a one-domain divergence.

This matches the accepted tradeoff in #10192/#10193. Marking ready.

PR #10228, opened 2026-07-21
fix(pg): actionable error when s3object input exceeds jsonb 256MB cap

rubenfiszel:
## Summary

Follow-up to #10152. Native SQL **PostgreSQL** scripts with an `(s3object)` parquet/csv input can still fail after the heartbeat/zombie fix, with an opaque server error.

Root cause: `(s3object)` inputs **materialize, not stream**. The whole referenced file is decoded to a JSON array and bound as a single `jsonb` parameter. PostgreSQL hard-caps a jsonb value's element payload at ~256MB, so a compressed input file (all columns, all rows materialized) can expand past the cap and fail with:

```
ERROR: total size of jsonb array elements exceeds the maximum of 268435455 bytes
```

The heartbeat fix (#10152) resolved the *zombie/ping* symptom; this cap is the remaining failure mode, and the raw error gives the user no idea what went wrong or what to do.

A proactive byte-size pre-check was considered and rejected: measured empirically, the jsonb *binary* size exceeds the JSON *text* size for typical row data (per-element keys), so no fixed text-length threshold reliably predicts PostgreSQL's cap without regressing legitimate sub-256MB payloads. Catching the actual server error is precise and never false-rejects.

## Changes

- `materialize_s3object_args` now reports whether any `(s3object)` arg was materialized.
- When a query with an `(s3object)` input fails with the jsonb value-size cap, `map_s3object_jsonb_overflow` rewrites the error into actionable guidance: `(s3object)` inputs are materialized (not streamed) with a ~256MB ceiling, and large files should use a DuckDB script, which reads Parquet/CSV directly from S3 and streams. The original server error is preserved under `Underlying error:`.
- Attribution is deliberately **hedged** ("if a large `(s3object)` input is the cause") and names no specific file: the same error can also come from SQL constructing an oversized jsonb at execution time, and with multiple inputs the offending value can't be identified. The DuckDB remediation is the same regardless.
- Matching is narrowed to `elements exceeds the maximum` (the jsonb array/object caps) so unrelated errors like PostgreSQL's `array size exceeds the maximum allowed` are not misclassified.
- Unit test `test_map_s3object_jsonb_overflow` covering the positive rewrite, the array-size negative case, an unrelated syntax error, and the no-`(s3object)`-input short-circuit.

## Test plan

- [x] `cargo test -p windmill-worker --features parquet,quickjs map_s3object_jsonb_overflow`
- [x] Confirmed the exact PostgreSQL error string (`total size of jsonb array elements exceeds the maximum of 268435455 bytes`) via psql, and that the jsonb binary size exceeds the JSON text size for typical row data.
- [x] Verified the query error propagates unwrapped through `run_future_with_polling_update_job_poller` (and the `run_inline` path) so the matcher sees the message.
- [ ] Manual (not runnable in this env — S3 download route requires the `private` feature, whose build fails on a pre-existing unrelated EE sqlx/DB issue): run a native PostgreSQL `(s3object)` script over a parquet that materializes past 256MB and confirm the job error shows the DuckDB guidance plus the preserved `Underlying error:` tail; confirm a small-file `(s3object)` query and an unrelated SQL error are unaffected.

### DIFF
--- a/backend/windmill-worker/src/pg_executor.rs
+++ b/backend/windmill-worker/src/pg_executor.rs
@@ -768,7 +768,7 @@ pub async fn do_postgresql(
     // Materialize any `(s3object)` args into JSON text and rebind them as `jsonb` so
     // `otyp_to_pg_type` picks the right binding. Must run before the param map is
     // built below.
-    materialize_s3object_args(
+    let had_s3object_input = materialize_s3object_args(
         &mut sig.args,
         &mut pg_args,
         client,
@@ -875,7 +875,7 @@ pub async fn do_postgresql(
     };
 
     let result = if run_inline {
-        result_f.await?
+        result_f.await
     } else {
         run_future_with_polling_update_job_poller(
             job.id,
@@ -889,8 +889,9 @@ pub async fn do_postgresql(
             &mut Some(occupancy_metrics),
             Box::pin(futures::stream::once(async { 0 })),
         )
-        .await?
-    };
+        .await
+    }
+    .map_err(|e| map_s3object_jsonb_overflow(e, had_s3object_input))?;
 
     // Release the cache lock now that we have the result — allows the
     // post-query caching code below to re-acquire it if needed.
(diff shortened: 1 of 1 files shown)


rubenfiszel:
✅ Review round clean @ b908b886db08d3f6e6f7f322dc55413dd310975e

PR #10233, opened 2026-07-21
feat(ai): session chat nits — empty sends, command picker polish, session-state prompt

Guilhem-lm:
## Summary

Three quality-of-life fixes for the AI session chat, from a QA nit-list session.

**1. Empty sends are real turns (GLOBAL chat, when context rides the draft).** Pressing Enter or clicking Send on a text-free composer previously did nothing. In GLOBAL mode (session chat and the global side panel), a text-free send that carries images or selected context elements (chips) now goes out as a normal model turn; the transcript shows just the chips/images for that message (no empty bubble). A completely bare send — no text, images, or context — is still dropped in every mode. The model-facing text substitutes `(the user sent an empty message)` when no text and no images ride the turn: every mode's template interpolates the text under an `## INSTRUCTIONS:` header, and models echo a dangling header back verbatim. Editor copilots (script/flow/app modes) keep dropping empty drafts so an accidental Enter doesn't burn a turn.

**2. `/` command picker polish.** The picker's wrapper was `max-h-64 overflow-hidden` around a flush `DrillPicker` whose `h-full` can't resolve against a max-height-only parent — the list was clipped with no way to scroll. The wrapper is now the scroll container. Rows lost the per-row Sparkles icon, are grouped under **Actions** / **Skills** headers (new optional `DrillLeaf.section`, fed by a `kind` on `sessionCommands` entries), and hovering a row for 250ms shows the full untruncated description in a floating tooltip (new opt-in `DrillPicker.rowTooltip` — one shared portal'd node, not a popover per row). In filtered view, keyboard navigation follows the grouped display order rather than raw fuzzy rank, so arrows never jump between non-adjacent visible rows.

**3. Session-state awareness.** The GLOBAL system prompt for session chats gains a "Session state" section describing the operating workspace: an ephemeral staged fork of a parent, a persistent dev workspace of a parent, the live workspace itself, or a pending fork that materializes at first send — including where deploys land and that fork changes are promoted via the session's review & deploy panel. Wired as a resolver (`manager.sessionContextResolver`, set by `sessionRuntime`) so copilot code doesn't import sessions modules; the send path already rebuilds the system message after `beforeSend`, so the very first message of a fork session carries the committed fork's identity.

## Changes
- `AIChatManager.svelte.ts`: empty-draft guard now scoped to non-GLOBAL modes; empty-message marker substituted into the model-facing text; `sessionContextResolver` field consulted in both system-message build paths
- `AIChatInput.svelte`: Send button enabled for empty GLOBAL drafts (pointer/touch parity with Enter); custom `onSendRequest` consumers and editor copilots still require content
- `DrillPicker.svelte` / `drillPicker.ts`: optional `section` leaf field with headers in browse + search views; search nav keys follow display order; opt-in `rowTooltip` hover tooltip
- `ChatCommandPicker.svelte`: scrollable wrapper, no per-row icon, Actions/Skills sections, full-description tooltip
- `global/core.ts`: `getSessionContextPromptSection` + `SessionPromptContext`, `ChatCommandItem` type
- `sessionRuntime.svelte.ts`: wires the session-context resolver from the session record + `$userWorkspaces` flags
- Tests: empty-send semantics pinned for GLOBAL vs non-GLOBAL in `AIChatManager.test.ts`; the four session-state prompt variants in `global/core.test.ts`

## Test plan
- [x] `npm run check:fast` clean; `AIChatManager.test.ts` (103) and `global/core.test.ts` session-context tests pass
- [x] Browser (session chat): bare empty send (no context) is blocked — button disabled, Enter no-ops; context-carrying empty send pinned by unit test → chips-only message row + model responds as to a nudge (no `## INSTRUCTIONS:` echo)
- [x] Browser: `/` picker scrolls past 16rem; Actions/Skills headers in browse and filtered views; hover tooltip shows full description and clears when typing; ArrowUp/Down walks the grouped display order and wraps correctly
- [x] Browser: session on a dev workspace — model answers "guilhem, your persistent dev workspace, forked from the parent workspace ai-meta-e2e"; staged-fork and pending-fork wording covered by unit tests
- [ ] Editor copilots (script/flow/app): empty Enter still no-ops (pinned by unit test)

## Screenshots

Session-state answer (dev workspace); note the screenshot predates the final polish — text-free messages no longer render the small empty bubbles visible here:

session-state-final

`/` picker: grouped sections (3xs hint headers), single-line rows (no icons, no inline descriptions), full-description hover tooltip:

picker-clean-rows

### DIFF
--- a/frontend/src/lib/components/DrillPicker.svelte
+++ b/frontend/src/lib/components/DrillPicker.svelte
@@ -186,6 +186,20 @@ leaves and ignores the current scope.
 		)
 	)
 
+	// Browse rows annotated with the section header rendered above them. A
+	// header is emitted when a leaf's section differs from the previous LEAF's
+	// section — comparing against the previous entry of any type would insert a
+	// duplicate header after every branch interleaved between same-section leaves.
+	const entryRows = $derived.by(() => {
+		let lastLeafSection: string | undefined
+		return entryList.map((entry) => {
+			const section = entry.type === 'leaf' ? entry.node.section : undefined
+			const header = entry.type === 'leaf' && section !== lastLeafSection ? section : undefined
+			if (entry.type === 'leaf') lastLeafSection = section
+			return { entry, header }
+		})
+	})
+
 	// Search nav must follow DISPLAY order (the grouped blocks), not the raw
 	// fuzzy ranking — grouping reorders interleaved results, and rank-ordered
 	// keys would make ArrowUp/Down jump between non-adjacent visible rows.
@@ -558,14 +572,11 @@ leaves and ignores the current scope.
 			<div role="status" class="px-3 py-2 text-xs text-tertiary">Empty</div>
 		{:else}
 			<div class="flex flex-col py-1">
-				{#each entryList as entry, i (entry.key)}
+				{#each entryRows as { entry, header } (entry.key)}
 					{@const isHl = entry.key === highlightedKey}
-					{@const section = entry.type === 'leaf' ? entry.node.section : undefined}
-					{@const prev = i > 0 ? entryList[i - 1] : undefined}
-					{@const prevSection = prev?.type === 'leaf' ? prev.node.section : undefined}
-					{#if section && section !== prevSection}
+					{#if header}
 						<div class="px-3 pt-2 pb-1 text-2xs uppercase tracking-wide text-tertiary font-medium">
-							{section}
+							{header}
 						</div>
 					{/if}
 					{#if entry.type === 'leaf'}
--- a/frontend/src/lib/components/DrillPicker.svelte
+++ b/frontend/src/lib/components/DrillPicker.svelte
@@ -186,9 +186,12 @@ leaves and ignores the current scope.
 		)
 	)
 
+	// Search nav must follow DISPLAY order (the grouped blocks), not the raw
+	// fuzzy ranking — grouping reorders interleaved results, and rank-ordered
+	// keys would make ArrowUp/Down jump between non-adjacent visible rows.
 	const navKeys = $derived(
 		isSearching
-			? (searchedItems ?? ([] as typeof searchItems)).map((r) => r.leaf.key)
+			? searchResultsByGroup.flatMap((g) => g.items.map((r) => r.leaf.key))
 			: entryList.map((e) => e.key)
 	)
 
--- a/frontend/src/lib/components/DrillPicker.svelte
+++ b/frontend/src/lib/components/DrillPicker.svelte
@@ -553,7 +553,7 @@ leaves and ignores the current scope.
 			{:else}
 				{#each searchResultsByGroup as { key, label, items } (key)}
 					{#if label}
-						<div class="px-3 pt-3 pb-1 text-2xs uppercase tracking-wide text-tertiary font-medium">
+						<div class="px-3 pt-3 pb-1 text-3xs uppercase tracking-wide text-hint font-medium">
 							{label}
 						</div>
 					{/if}
@@ -575,7 +575,7 @@ leaves and ignores the current scope.
 				{#each entryRows as { entry, header } (entry.key)}
 					{@const isHl = entry.key === highlightedKey}
 					{#if header}
-						<div class="px-3 pt-2 pb-1 text-2xs uppercase tracking-wide text-tertiary font-medium">
+						<div class="px-3 pt-2 pb-1 text-3xs uppercase tracking-wide text-hint font-medium">
 							{header}
 						</div>
 					{/if}
--- a/frontend/src/lib/components/DrillPicker.svelte
+++ b/frontend/src/lib/components/DrillPicker.svelte
@@ -553,7 +553,9 @@ leaves and ignores the current scope.
 			{:else}
 				{#each searchResultsByGroup as { key, label, items } (key)}
 					{#if label}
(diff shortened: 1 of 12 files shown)


Guilhem-lm:
Round outcome — nit-only verdicts, addressed as follows:

**Fixed in 7d41eb613a:**
- **Claude P2** (fork mislabeled as live workspace when the committed workspace is missing from `userWorkspaces`): the resolver now mirrors `isForkSession`'s fallback — a committed-but-unlisted workspace is described as a fork with an unknown parent, never as the live workspace. Pinned by a new unit test.
- **Pi P2** (duplicate section header when a branch interleaves same-section leaves): browse rows are now annotated against the previous *leaf*'s section (`entryRows`), so branches between same-section leaves no longer emit a duplicate header.

**Dismissed with reasoning:**
- **Pi P1** (floating-ui reference-action leak on every hovered row): not a leak in this library. In `svelte-floating-ui`, `referenceAction(node)` for a plain `HTMLElement` only reassigns `referenceElement` and calls `update()` — it registers no listeners and returns no cleanup (source: `svelte-floating-ui/dist/index.js`, `referenceAction`). The `autoUpdate` instance (ResizeObserver + scroll listeners) is created by the *content* action, which is destroyed whenever the tooltip node unmounts — and the tooltip unmounts between row hovers (`tooltipLeave` clears `tooltipText` on `mouseleave`, and on filter changes). Repeated `tooltipRef(el)` calls therefore accumulate nothing.
- **Codex P2** (missing `ai_evals` A/B for a GLOBAL prompt/behavior change): the session-state section is appended only when `sessionContextResolver` is set, which only session runtimes do — the eval harness and the global side-panel chat build their system prompt without it, so the evaluated GLOBAL surface is byte-identical. The other behavior change (empty-send marker) is reachable only from an empty user message, which no eval case sends. An A/B would diff nothing.

Guilhem-lm:
✅ Review round clean @ 7d41eb613a4aaa2c4d7aefafdff2ad0238793be1 — nit-only verdicts at c34a66718f0baadc1902283aa7dd3288a023fb7e; nits addressed in 7d41eb613a / dismissed in the round-outcome comment above

Guilhem-lm:
Post-ready review findings addressed in f37b0b0679:

- **Codex P2 (descriptions inaccessible without a pointer)**: fixed both halves — option buttons now carry `aria-label="/name — description"` when a `rowTooltip` is configured (screen-reader access), and ArrowUp/Down keyboard navigation anchors the same tooltip to the highlighted row (visual keyboard access). Verified in the browser: arrowing onto a skill shows its full description after the same 250ms delay.
- **cubic P2 (stale mode in the empty-send guard)**: valid — `changeMode` can decline a switch (e.g. SCRIPT with no model configured), so the guard now gates on `requestedMode` instead of `this.mode`; a declined non-GLOBAL request can no longer slip through as a GLOBAL empty turn.
- **cubic P2 (workspace metadata in the system prompt)**: the free-form workspace display name is no longer interpolated into the prompt — only slug-validated workspace IDs appear, which removes the injection surface entirely (the name added no information the ID doesn't).
- **Codex P2 (stale "empty bubble" descriptions)**: test comments updated to describe the current rendering (chips only, no bubble); the PR description was already updated with the same wording.

All 258 tests across the two touched suites pass; typecheck clean.

Guilhem-lm:
Round findings addressed in b949b00802 — Claude, Pi, and the post-push Codex review all converged on the same P2 (context-only draft silently dropped when Enter is pressed mid-stream). Fixed across all four spots the reviews identified, in tandem:

- `AIChatInput` loading branch now queues a GLOBAL context-only draft (mirroring the idle send guard) and pins the selection snapshot to the queued entry.
- `queueMessage` accepts context-only entries (guard now mirrors the idle empty-send rule).
- `#hasQueuedMessage` counts a pinned context selection, so the flush and dequeue paths see the entry.
- `QueuedMessageChip` renders context badges for context-only queues, keeping the queued draft visible and dismissable instead of auto-sending as a surprise turn (badges are shown only for context-ONLY queues — text queues keep their chips visible in the composer, and repeating them would read as two selections).

Pinned by a new unit test ("queues a context-only draft while streaming"); 105/105 manager tests pass, typecheck clean.

Guilhem-lm:
✅ Review round clean @ b949b0080272b603c2039cb938d031268ce3054a — nit-only verdicts at f37b0b0679a8dd3ca7cfdc94bf706ffdd972d3c8 (all three reviewers: the same context-only mid-stream queue gap); addressed in b949b00802 with the regression test the round requested

Guilhem-lm:
Latest round of findings addressed in 10804b54c0:

- **Codex P2 + cubic P2 (duplicate keys for same-titled DOM-selector badges in `QueuedMessageChip`)**: the composer's `(appPath, selector)` identity helpers (`contextElementKey` / `isSameContextElement`) moved from `AIChatInput` into `context.ts` as shared exports; the queued-badge `{#each}` now keys with `contextElementKey`, so distinct same-titled DOM selections keep distinct Svelte keys.
- **cubic P2 (`queueMessage` drops earlier non-DOM context when a second prompt queues)**: valid — the latest-snapshot-wins merge made sense when the queue read the live selection at flush, but each queued prompt now pins its selection at press time. The merge is now a full identity union (deduped via `isSameContextElement`), so a prompt queued with a workspace-item chip keeps it when a later prompt queues with different chips. Pinned by extending the queue regression test (union keeps the earlier chip, re-selected chips don't duplicate).

105/105 manager tests pass, typecheck clean.

PR #10234, opened 2026-07-21
fix(frontend): limit compare & deploy rows to the active direction

Guilhem-lm:
## Summary
The Compare & Deploy page rendered every item from the workspace comparison in both directions: the "Deploy to parent" view also listed behind-only (update-direction) items, and the "Update current" view listed ahead-only items — all as greyed-out, non-selectable rows with no explanation of why they can't be acted on. This PR limits the list to the direction being viewed, so each view shows exactly the rows it can act on.

Conflict items (ahead **and** behind) still appear in both views since they are actionable either way, and awareness of the other direction is preserved by the toggle's per-direction counts ("Deploy to X (9)" / "Update current (4)") and the existing "fork is behind parent" alert.

## Changes
- `deployableItems` is now built from `selectableDiffs` (the direction-filtered set) instead of all `comparison.diffs`; just-deployed rows keep their "Deployed" badge until the comparison refresh drops them, exactly as before
- Removed the `selectablePredicate` prop passed to `WorkspaceDeployLayout` and the `isSelectable` check in the row summary — both are always true now that every rendered row is selectable
- Removed the "N total items" badge: it counted both directions, which would contradict the filtered list; the remaining "N deployable/updateable" badge matches the visible rows and the toggle labels carry each direction's count

## Test plan
- [x] `npm run check` passes (0 errors)
- [x] Fork with 9 ahead-only + 4 behind-only items: deploy view lists exactly the 9 deployable rows, update view exactly the 4 updateable rows (screenshots below)
- [x] Fork with a conflict item (ahead and behind): the conflict row and its badge appear in both views; ahead-only items stay out of the update view
- [x] Direction with zero actionable items still renders the explanatory empty-state message (unchanged `nothingToAct` path)

## Screenshots
"Deploy to parent" view — only the 9 deployable items (the 4 behind-only items no longer show as disabled rows):

deploy-to-parent-view

"Update current" view — only the 4 updateable items:

update-current-view

### DIFF
--- a/frontend/src/lib/components/CompareWorkspaces.svelte
+++ b/frontend/src/lib/components/CompareWorkspaces.svelte
@@ -910,22 +910,18 @@
 	// Trigger and schedule rows now flow through `comparison.diffs` like every
 	// other deployable kind — the backend's `compareWorkspaces` populates them
 	// from `workspace_diff`, with runtime fields ignored by `compare_two_*`.
-	let deployableItems = $derived.by(() => {
-		return (comparison?.diffs ?? [])
-			.filter((diff) => {
-				const key = getItemKey(diff)
-				const isSelectable = selectableDiffs.includes(diff)
-				const isDeployedAndIrrelevant =
-					deploymentStatus[key]?.status === 'deployed' && !isSelectable
-				return !isDeployedAndIrrelevant
-			})
-			.map((diff) => ({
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
✅ Review round clean @ 046c0074650f4d19e085ad2a07c3b406ad2600b1

PR #10235, opened 2026-07-21
fix(frontend): curl fallback for +Variable/+Resource in bash sandbox mode

rubenfiszel:
## Summary

In a bash script that runs inside a custom container image (`# sandbox <image>` or `# docker`), the body has no `wmill` CLI, so the `+Variable` / `+Resource` picker snippets (`wmill variable get …` / `wmill resource get …`) fail. This inserts a plain HTTP snippet in that case, using the `BASE_INTERNAL_URL`, `WM_TOKEN`, and `WM_WORKSPACE` env vars that are available inside sandbox (confirmed in `bash_executor.rs` / `docker_v2.rs`).

## Changes

- `frontend/src/lib/script_helpers.ts`: new `bashRunsInCustomImage(code)` helper that mirrors the worker's `BashAnnotations` grammar (`backend/windmill-common/src/worker.rs`) — only leading comment lines are scanned; `# sandbox <image>` (image present) or a bare `# docker` line selects a custom container. A bare `# sandbox` is the nsjail-bash modifier (runs on the worker rootfs, `wmill` available) and is correctly excluded.
- `frontend/src/lib/components/EditorBar.svelte`: the `+Variable` / `+Resource` bash branches use the helper. In a custom image they insert a `curl` snippet with a busybox-`wget` fallback (the default `# sandbox alpine:latest` image ships `wget`, not `curl`). The variable snippet strips the surrounding quotes from `get_value`'s JSON string with `sed` so it matches the `jq -r .value` output of the non-sandbox branch; resources return JSON either way.
- `frontend/src/lib/script_helpers.test.ts`: focused unit tests for the detection grammar (spaced/compact annotations, bare `# sandbox`, prose false positives, non-leading annotations, default template).

## Notes / decisions

- **Why `curl || wget`**: no HTTP client is universal, but the default sandbox template is `alpine:latest`, whose busybox ships `wget` (not `curl`). Trying `curl` first then `wget` makes the snippet work on the default image and on images that provide `curl`. `sed` (used to unquote the variable value) is present in both busybox and coreutils.
- **Grammar parity over substring match**: the original `code.includes('# sandbox')` both missed compact annotations and fired on prose/body occurrences and on bare `# sandbox` (which keeps `wmill`). The helper mirrors the worker parser instead.

## Test plan

Verified end-to-end in the browser (bash script editor, headless Playwright):

- [x] `# sandbox alpine:latest` (after a leading `# shellcheck` line) + `+Variable` → `{ curl … || wget …; } | sed 's/^"//;s/"$//'`
- [x] `# docker` + `+Resource` → `curl … || wget …` (raw JSON, no `sed`)
- [x] Bare `# sandbox` + `+Variable` → keeps `wmill variable get … --json | jq -r .value`
- [x] Plain bash (no directive) → keeps the `wmill` snippets
- [x] `variables/get_value` returns `"myvalue"` (JSON-quoted, confirmed against the running API); `sed 's/^"//;s/"$//'` strips it to `myvalue`, matching `jq -r .value`
- [x] `npm run check:fast` clean (only a pre-existing unrelated `modern-screenshot` error in `RawAppEditor.svelte`)
- [x] `vitest run src/lib/script_helpers.test.ts` — 7 passed

### Screenshot

`# sandbox alpine:latest` + `+Variable` inserts the curl/wget snippet with the `sed` unquote (line scrolled to show the tail):

sandbox-curl

Fixes WIN-2215

### DIFF
--- a/frontend/src/lib/components/EditorBar.svelte
+++ b/frontend/src/lib/components/EditorBar.svelte
@@ -39,6 +39,7 @@
 	import { createEventDispatcher, untrack } from 'svelte'
 	import { sendUserToast } from '$lib/toast'
 	import { getScriptByPath, scriptLangToEditorLang } from '$lib/scripts'
+	import { bashRunsInCustomImage } from '$lib/script_helpers'
 	import Toggle from './Toggle.svelte'
 
 	import {
@@ -673,10 +674,13 @@
 			}
 			editor.insertAtCursor(`v, _ := wmill.GetVariable("${path}")`)
 		} else if (lang == 'bash') {
-			const code = editor.getCode()
-			if (code.includes('# sandbox') || code.includes('# docker')) {
+			if (bashRunsInCustomImage(editor.getCode())) {
+				// Custom image: no wmill CLI. Fall back to curl, then busybox wget
+				// (the default `# sandbox alpine:latest` image ships wget, not curl).
+				// get_value returns a JSON-quoted string, so strip the outer quotes
+				// to match the `jq -r .value` output of the non-sandbox branch.
 				editor.insertAtCursor(
-					`curl -s -H "Authorization: Bearer $WM_TOKEN" "$BASE_INTERNAL_URL/api/w/$WM_WORKSPACE/variables/get_value/${path}"`
+					`{ curl -sf -H "Authorization: Bearer $WM_TOKEN" "$BASE_INTERNAL_URL/api/w/$WM_WORKSPACE/variables/get_value/${path}" 2>/dev/null || wget -qO- --header="Authorization: Bearer $WM_TOKEN" "$BASE_INTERNAL_URL/api/w/$WM_WORKSPACE/variables/get_value/${path}"; } | sed 's/^"//;s/"$//'`
 				)
 			} else {
 				editor.insertAtCursor(`wmill variable get ${path} --json | jq -r .value`)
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Addressed the review findings in 4aeecf6:

**Codex P1 — curl not guaranteed in sandbox images.** Correct; the default `# sandbox alpine:latest` template ships busybox `wget`, not `curl`. Both snippets now do `curl … || wget …` so they work on the default alpine image and on images that provide curl.

**Codex P1 — detection disagrees with worker parsing.** Replaced the substring check with `bashRunsInCustomImage`, mirroring the worker's `BashAnnotations` grammar (leading comment lines only; `# sandbox <image>` or bare `# docker`). This fixes the compact-annotation false negatives and the prose/body false positives, and — since a bare `# sandbox` is the nsjail-bash modifier that still has `wmill` — it now correctly keeps the `wmill` snippet for bare `# sandbox`. Added focused unit tests (`script_helpers.test.ts`).

**Claude P2 — variable snippet returned a JSON-quoted value.** The variable snippet now pipes through `sed 's/^"//;s/"$//'` to strip the outer quotes, matching the `jq -r .value` output of the non-sandbox branch. `sed` is present in both busybox and coreutils. Verified against the running API: `get_value` returns `"myvalue"` and the snippet yields `myvalue`. The resource snippet is unchanged (returns JSON, matching `jq .value`).

PR #10248, opened 2026-07-22
fix: prevent u16 underflow in suspend count causing permanent flow deadlock

tusharui:
## Problem

In `worker_flow.rs`, the suspend event count is computed as:

(required_events - resume_messages.len() as u16) as i32
When extra resume_messages arrive concurrently and resume_messages.len() exceeds required_events, this u16 subtraction wraps around to ~65535. The resulting value is written to the database as the suspend counter, causing the flow to wait for 65K+ events that will never arrive — permanently deadlocking the workflow.
Fix
Replace wrapping subtraction with saturating_sub():
(required_events.saturating_sub(resume_messages.len() as u16)) as i32
When the subtraction would underflow, saturating_sub() returns 0 instead, correctly indicating that no additional events are needed.
Impact
- Fixes production flows that get stuck indefinitely under concurrent resume pressure
- Zero behavioral change in the normal case (when resume_messages.len() < required_events)
- Single-line change, no risk of regression

---
## Summary by cubic
Fixes a deadlock in suspended flow steps by preventing a u16 underflow in the suspend event count, and preserves companion script dependency cycles when updates are triggered by relative import changes.

- **Bug Fixes**
  - In `worker_flow.rs`, use `saturating_sub` when computing remaining events so extra concurrent resume messages don’t wrap the counter to ~65535 and stall the flow.
  - In `process_relative_imports`, when `triggered_by_relative_import` is true, skip both `patch` and `dissolve` to keep existing dependency edges intact and avoid breaking the cycle.

<sup>Written for commit d20b41e7b8e62c6f5e9c6cfbf877ef64700813d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-dep-map/src/lib.rs
+++ b/backend/windmill-dep-map/src/lib.rs
@@ -88,7 +88,13 @@ pub fn extract_referenced_paths(
 ) -> Option<Vec<String>> {
     let mut referenced_paths = vec![];
     if let Some(wk_deps_refs) = language
-        .and_then(|l| windmill_common::scripts::extract_workspace_dependencies_annotated_refs(&l, raw_code, script_path))
+        .and_then(|l| {
+            windmill_common::scripts::extract_workspace_dependencies_annotated_refs(
+                &l,
+                raw_code,
+                script_path,
+            )
+        })
         .map(|r| r.external)
     {
         let l = language.expect("should be some");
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Superseded by #10256, which isolates the suspend-count `saturating_sub` fix as a true single-line change (the title/description here only described that fix).

Note: this branch also bundled an unrelated `windmill-dep-map` change (`process_relative_imports` skip patch+dissolve when `triggered_by_relative_import`). That was **not** carried over to #10256 — if it's still wanted, please open it as its own PR.

rubenfiszel:
Thanks a lot for catching this and sending the fix @tusharui — the `saturating_sub` underflow fix is a real bug and much appreciated! 🙏

I closed this one only because it also bundled an unrelated `windmill-dep-map` change, so I split the suspend fix out into its own single-line PR (#10256) to keep it clean and easy to review. If you'd still like the dependency-map change landed, please open it as a separate PR and we'll take a look.

PR #10230, opened 2026-07-21
feat(triggers): add AMQP (RabbitMQ) trigger via lapin

rubenfiszel:
## Summary

Implements a new **AMQP** trigger kind (RabbitMQ / AMQP 0.9.1) using the [lapin](https://github.com/amqp-rs/lapin) library. It consumes messages from an AMQP queue and runs a script or flow per message, mirroring the existing pull/streaming triggers (MQTT is the closest analogue). AMQP is a Community-Edition trigger (part of `oss_core`), not gated behind enterprise.

Fixes WIN-2214

## Changes

**New backend crate** `windmill-trigger-amqp` (behind the `amqp_trigger` feature):
- `AmqpTrigger` implementing `TriggerCrud` + `Listener` + `TriggerJobArgs`
- lapin-based client builder: builds the `amqp(s)://` URI from an `amqp` resource (host/port/username/password/vhost/tls), opens a channel, optional `basic_qos` prefetch, declares (or passively declares) the queue, optionally binds it to an exchange for each routing key, and consumes — acking each delivery after dispatch (at-most-once, matching the SQS listener)
- Payload = raw message body (`payload` as bytes / base64, same shape as MQTT); trigger info exposes `exchange`, `routing_key`, `queue_name`, `redelivered`, `delivery_tag`
- `test_connection` verifies broker reachability

**Config** persisted per trigger: `amqp_resource_path`, `queue_name`, optional `exchange` (`{exchange_name, routing_keys[]}`), optional `options` (`{declare_queue, prefetch_count}`).

**Wiring** (mirrors MQTT end-to-end):
- Migrations: `amqp_trigger` table (RLS with wrapped session GUCs) + `amqp`/`trigger_amqp` values added to the `trigger_kind`, `job_trigger_kind`, `draft_kind` enums
- `TriggerKind::Amqp`, `JobTriggerKind::Amqp`, `UserDraftItemKind::TriggerAmqp`, `DeployedObject::AmqpTrigger`, `ScopeDomain::AmqpTriggers`
- API route registration + listener startup, capture config, workspace used-triggers/export/fork-clone/rename, offboarding/trash/granular-acls/users/token arrays, script-rename CTE
- OpenAPI (paths + schemas + enum extensions) and regenerated clients
- CLI: `wmill trigger`/`sync` support (`amqp_trigger` files round-trip), regenerated CLI client + guidance
[...]

PR #10251, opened 2026-07-22
fix(windows): lean worker-only build, stop compiling the amqp trigger

rubenfiszel:
## Summary

`windmill-trigger-amqp` does not compile on Windows: `tokio-reactor-trait` only implements `reactor_trait::Reactor` for its `Tokio` type under `#[cfg(unix)]`. Since the amqp trigger landed (#10230) this has broken **two** Windows CI jobs:

- **`build_windows_worker_`** — `ee_windows` enabled `amqp_trigger` (via `ce_core`→`oss_core`), pulling the crate into the worker binary.
- **`backend-test-windows`** — `cargo test --all` compiles every workspace member *regardless of features*, so the amqp crate is built directly. **Red on every release since #10230.**

The amqp trigger is a server-only feature never run on Windows workers, so the fix is to **stop compiling it on Windows** (rather than port its reactor). This also gives the worker a leaner binary.

`--all` (= `--workspace`) is a package-selection primitive, orthogonal to `--features` — it builds every member unconditionally. So a feature drop fixes the binary but not the `--all` test build; that needs package selection.

## Changes

- **`backend/Cargo.toml`** — replace `ee_windows = ["ce_core", "ee_core", "all_languages_windows"]` with a worker-only `worker_windows_core` (never enables `amqp_trigger`). A non-agent worker still runs the full `windmill-api` on `localhost` (`main.rs` `run_server`, `if !is_agent`) and jobs call back into it via the `wmill` client, so:

  **Kept** (worker runtime / jobs touch in-process): `private`, `operator`, `parquet`, `quickjs`, `enterprise`(→`license`), `prometheus`, `otel`, `jemalloc`, `all_languages_windows`, `windmill-worker/mcp` + `windmill-store/mcp` (MCP client + OAuth-MCP refresh), `windmill-worker/bedrock` (direct Bedrock), `openidconnect` (Vault secrets), `run_inline`, `windmill-api/instance_smtp` (critical alerts + error-handler email), `oauth2` (`OAUTH_CLIENTS` init + token refresh in the worker's internal server).

  **Dropped** (external/server-only surface the worker never runs): all 11 `*_trigger`/`kafka`/`nats`/`sqs` listeners; `static_frontend`; `stripe`; `embedding`; `zip`; `windmill-api/mcp` (MCP gateway); `windmill-api/bedrock` (server Bedrock proxy); `cloud` (runtime-gated on `CLOUD_HOSTED`).

- **`windmill-api` smtp split** — `send_email_with_instance_smtp` (error-handler email) only needs `windmill-common`'s rustls sender, but the `smtp` feature also bundled the inbound email trigger's `openssl` + `mail-parser` + `windmill-trigger-email`. Added `instance_smtp = ["windmill-common/smtp"]` gating just the endpoint (`smtp` now includes it); the worker uses `instance_smtp`, avoiding openssl (which broke the check step) and the trigger crate.

- **`backend-test-windows.yml`** — the Windows binary is worker-only, so test the crates a worker runs (`-p windmill-worker -p windmill-common -p windmill-queue`) instead of `cargo test --all`. `--all` compiled every member (pulling in amqp, which doesn't build on Windows) and linked the whole `windmill-api` integration-test suite, overrunning the runner disk (LNK1180). Also unset the `setup-rust-toolchain` default `RUSTFLAGS=-D warnings` so cross-platform dead-code (`cfg(unix)`-only helpers) doesn't fail the run. Full-workspace coverage stays on the Linux CI.

- **`build_windows_worker_.yml` + `publish_windows_worker.yml`** — drop the redundant `mkdir frontend/build`; stub `openapi-deref.json` next to the `.yaml` stub (~2.5MB the worker never serves).

The `windmill-trigger-*` crates are unchanged — server-only, full coverage on the Linux CI.

## Validation

- ✅ `build_windows_worker_` green: `cargo check --features=ee_windows -D warnings` + DuckDB FFI + release binary build
- ✅ `backend-test-windows` green (first time since #10230)
- Dockerfiles/release workflows use `ce`/`ee`/`ee_rhel`/`ce_rpi` bundles; `smtp` kept as a superset (`instance_smtp` transitively), so they're unaffected.

## Manual smoke (recommended before relying on it)

Run the Windows worker artifact: representative language job; AI-agent step over an OAuth-authenticated MCP tool and an AWS Bedrock model; expired OAuth resource read (refresh); OIDC secret read; critical-alert + error-handler email; `run_inline`/inline-SQL call.

### DIFF
--- a/.github/workflows/backend-test-windows.yml
+++ b/.github/workflows/backend-test-windows.yml
@@ -51,6 +51,11 @@ jobs:
         with:
           cache-workspaces: backend
           toolchain: 1.97.0
+          # This action defaults RUSTFLAGS to "-D warnings"; unset it so the test
+          # run is not failed by cross-platform dead-code (cfg(unix)-only helpers
+          # are unused on Windows). Warning hygiene is enforced on the Linux CI
+          # and the build_windows_worker_ release build, not this test job.
+          rustflags: ""
 
       - uses: actions/setup-dotnet@v4
         with:
@@ -203,9 +208,15 @@ jobs:
           WMDEBUG_FORCE_V0_WORKSPACE_DEPENDENCIES: 1
           WMDEBUG_FORCE_RUNNABLE_SETTINGS_V0: 1
           WMDEBUG_FORCE_NO_LEGACY_DEBOUNCING_COMPAT: 1
+        # Windows ships a worker-only binary, so test the crates a worker runs
+        # (windmill-worker/-common/-queue) via -p, not `--all`: this skips the
+        # disk-heavy windmill-api test binaries (LNK1180) and the server-only
+        # windmill-trigger-* crates (amqp does not build on Windows). Linux CI runs the rest.
         run: >
           cargo test
           --no-fail-fast
-          --features enterprise,deno_core,duckdb,license,python,rust,scoped_cache,parquet,private,csharp,php,quickjs,mcp,run_inline
-          --all
+          -p windmill-worker
+          -p windmill-common
+          -p windmill-queue
+          --features private,enterprise,deno_core,duckdb,python,rust,csharp,php,quickjs,parquet,mcp,scoped_cache,windmill-git-sync/private,windmill-object-store/private,windmill-object-store/enterprise
           -- --nocapture --test-threads=10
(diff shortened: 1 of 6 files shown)


rubenfiszel:
✅ Review round clean @ afc2f3dafb325d0ed74624fb2ab7e6934b158b9e

PR #10254, opened 2026-07-22
feat(ai-sessions): show item preview cards for tools

rubenfiszel:
## Summary

Fixes WIN-2219. In AI sessions, when the assistant calls a tool that **creates**, **updates**, or **opens the preview of** a workspace item (script, flow, raw app, pipeline), the tool call now renders a discrete **"Preview" chip pinned to the right of the tool-call header row**. Clicking it opens the item's live preview in the session side panel, or focuses the tab if it is already open.

Previously the only way to see a written item was to expand the tool details or wait for the assistant to call `open_preview`; nothing gave the user a one-click, always-visible affordance per tool call.

## Screenshots

Shipped placement (variant C — chip on the header row; the item icon signals the kind, the tool text truncates when long):

Light:

variant-c

Dark:

variant-c-dark

## Changes

- **`shared.ts`**: new `OpenItemPreviewAction` variant on `ToolDisplayAction`, plus an `openItemPreviewAction(kind, path)` builder; new `previewCard?: { kind, path }` field on `ToolDisplayMessage`.
- **`ToolPreviewCard.svelte`** (new): the "Preview" chip (item icon via `RowIcon` + "Preview" + panel icon). Dispatches through the existing `runToolDisplayAction` registry.
- **`ToolExecutionDisplay.svelte`**: the header is a snippet; when a preview chip is present the header row becomes a `justify-between` flex (tool text left with truncation, chip right). Chip is **always visible** (not gated on expand) and never shown while loading / erroring / awaiting confirmation.
- **`sessions/+page.svelte`**: registers the `open_item_preview` handler once; on click it opens (or focuses and pulses) the item in the active session's preview panel via the existing `previewTargetForSessionTarget` and `previewTabs.open` primitives.
- **`global/core.ts`**: the write finalizers (`finishDraftWrite`, `finishAppDraftWrite`) and the `open_preview` tool attach a `previewCard` for previewable kinds (script/flow/raw_app, plus pipeline for `open_preview`). Gated on `ctx.sessionId` (the global side-panel chat has no preview panel). The `open_preview` card is attached only when the preview actually opened, never for a refused open such as a gated pipeline.
- **`ToolMessageActions.svelte`**: defensive branch so the widened union stays type-safe (preview cards never route here).
- **`shared.test.ts`**: unit test for the action builder contract.

No model-facing behavior changes: `previewCard` only affects the UI tool-display message; tool result strings and tool schemas/descriptions are unchanged, so no eval was needed.

## Test plan

- [x] `npm run check:fast` clean (only the pre-existing, unrelated `modern-screenshot` type error remains)
- [x] `shared.test.ts` / `AIChatManager.test.ts` / `sessionPreviewTabs.test.ts` pass
- [x] Chip renders for script/flow/app/pipeline with correct icon; long tool text truncates so the chip stays pinned right
- [x] Clicking dispatches through `runToolDisplayAction`
- [x] Sessions page mounts cleanly with the handler registered
- [ ] End-to-end with a live model: write a script in a session, chip appears, click opens the preview (not exercisable in this CE dev env; AI provider not configured)

### DIFF
--- a/frontend/src/lib/components/copilot/chat/ToolExecutionDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/ToolExecutionDisplay.svelte
@@ -51,6 +51,14 @@
 	const activeUserQuestion = $derived(
 		isActiveUserQuestion(message) ? message.userQuestion : undefined
 	)
+
+	// The preview chip sits on the header row (to the right of the tool-call text);
+	// shown once the tool settled, never while loading/erroring/awaiting confirmation.
+	const showPreviewChip = $derived(
+		Boolean(
+			message.previewCard && !message.isLoading && !message.error && !message.needsConfirmation
+		)
+	)
 </script>
 
 {#if activeUserQuestion}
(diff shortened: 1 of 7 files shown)


rubenfiszel:
✅ Review round clean @ a14bb46312c1d8d3e5581d305201b66342100cd3

PR #10258, opened 2026-07-22
feat: open_page opens resource/variable edit drawers when asking user to act

Guilhem-lm:
## Summary

When the AI chat asks the user to perform a manual UI step — most commonly "fill in this resource's credentials" or "enter the secret value" — it used to just describe where to click. Now it opens the relevant editor next to the chat: the `open_page` tool gains an `open` parameter for the Resources and Variables pages (landing directly in the item's edit drawer), and the global system prompt instructs the model to always pair such requests with an `open_page` call.

## Changes

- **`open_page` tool**: `open` (exact item path → edit drawer) now also applies to `resources` and `variables`, reusing the hash deep-links those pages already support (`#/resource/<path>` and `#<path>`). Field descriptions updated; `buildResourcesUrl`/`buildVariablesUrl` take `{ open, filters }` like the schedules builder.
- **Prompt**: new rule bullet + `OPEN_PAGE_DESCRIPTION` update — when asking the user to act in the UI (fill credentials, set a variable's value, adjust a setting), call `open_page` targeted at that item in the same message.
- **Pages**: the resources, variables, and schedules pages now handle their drawer-opening hash *reactively* (`$effect` + last-handled-hash tracking) instead of `onMount`-once, so re-pointing an already-mounted session preview tab at a new hash opens the drawer too. Row links pre-set the handled hash since their `onclick` already opens the drawer (no double-open). Also removes a stray `console.log` in the resources handler.
- **Headless fallback**: `open_page` falls back to the chat loop's workspace when neither session helpers nor `workspaceStore` provide one (required for ai_evals).
- **Eval**: new `global-resource-manual-credentials` case — a postgres resource whose password the user wants to enter themselves; deterministic checks require the resource draft plus an `open_page` call with `page` ∈ {resources, variables} and an item-targeted `open`.

## Eval A/B (Sonnet, 3 runs/arm)

| Arm | Pass rate | Behavior |
|---|---|---|
| Before (main prompt/tool) | 0/3 | Draft created, but the model only described where to click — never called `open_page` |
| After (this PR) | 3/3 | Creates secret variable + resource referencing it, then `open_page(page="variables", open="f/evals/global/prod_db_password")` |

Judge scores were high in both arms (the judge only sees drafts, which were fine either way) — the delta is exactly the new tool behavior.

## Screenshots

Deep link `/resources#/resource/<path>` landing directly in the edit drawer (same pattern for `/variables#<path>`):

resources-deep-link-drawer

Browser-verified beyond the screenshot: changing the hash on an already-mounted page switches the drawer without reload, a hash-only iframe `src` change (what the session preview's tab re-point does) propagates and switches the drawer, and row clicks still open the drawer exactly once.

## Test plan

- [x] `pageNavigation.test.ts` — new assertions pin the `#/resource/<path>` and `#<path>` hash outputs (11/11)
- [x] `core.test.ts` (163/163), `npm run check:fast` (only pre-existing unrelated errors)
- [x] Playwright: deep links open the right drawer on resources/variables/schedules; reactive re-point verified on mounted pages and through an iframe hash-only `src` change; row-click single-open verified
- [x] ai_evals A/B on `global-resource-manual-credentials` (before 0/3, after 3/3)

### DIFF
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -1739,3 +1739,44 @@
     - creates an AI draft for the existing f/evals/global/format_greeting script with the exclamation-mark change
     - the draft changes only the returned message's punctuation — summary, language, path, and the rest of the code are untouched
     - does not deploy or save the draft to the workspace
+
+- id: global-resource-manual-credentials
+  prompt: |-
+    Set up a resource for our production Postgres database at `f/evals/global/prod_db` (host db.internal.example.com, port 5432, database `orders`, user `app`).
+    I don't want to paste the password into this chat — prepare everything so I can enter it myself.
+    Leave it as an AI draft only; do not deploy or save it.
+  initial: ai_evals/fixtures/frontend/global/initial/user_admin_evals_folder.json
+  runtime:
+    maxTurns: 12
+  validate:
+    draftCountAtLeast: 1
+    requiredDrafts:
+      - type: resource
+        path: f/evals/global/prod_db
+        valueIncludes:
+          - db.internal.example.com
+          - orders
+  toolExpect:
+    requiredToolsUsed:
+      - write_resource
+      - open_page
+    forbiddenToolsUsed:
+      - deploy_workspace_item
+      - delete_workspace_item
+    toolCallArgs:
+      # The model may land the user in the resource's drawer or in the drawer of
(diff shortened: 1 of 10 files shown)


Guilhem-lm:
✅ Review round clean @ c40324a5a36981a3a927841e7feb3ce23e933990

PR #10260, opened 2026-07-22
fix(copilot): stop write_flow forcing rawscript code into nested JSON

rubenfiszel:
## Summary

The AI copilot's **global-chat `write_flow` tool** repeatedly failed on the first tries when creating flows (observed on **GPT-5**), because the model had to embed rawscript **code** inside the `modules` JSON **string** — so the code had to survive **three levels of JSON escaping** and the model kept mangling it.

## The issue

A tool call arrives as one JSON string (`processToolCall` does `JSON.parse(toolCall.function.arguments)`). For `write_flow`, `modules` is itself a `z.string()` (`JSON.parse`'d again), and each rawscript `content` is a third string holding actual code:

```
toolCall.function.arguments   ← JSON string for the whole call
  └─ "modules"                ← itself a JSON string  (schema is z.string())
       └─ "content"           ← itself a JSON string holding the code
            └─ the code        ← naturally contains  "  '  \n  {  }
```

A single `"` in the source must be escaped for the content string, then the modules string, then the arguments string — miss one backslash and the whole `modules` string fails to parse. `modules` is a string (not a typed array) on purpose: the flow schema is recursive, so `z.toJSONSchema` emits `$ref`/`$defs`, which Gemini's tool API rejects. In the original GPT-5 trace this produced repeated `Invalid JSON for modules` failures and structural mistakes before the model stumbled onto passing a native array.

## The solution

Keep code **out of** the `modules` structure and route it out-of-band through `set_flow_module_code` (a top-level `code` param — one escaping layer), the same escape hatch flow-editor mode already had via `set_module_code`:

- **Empty-body warning** — after a `write_flow`/`patch_flow_json`, detect rawscript modules left empty or as `inline_script.` placeholders (recursing into loops/branches/aiagent tools + preprocessor/failure) and tell the model which ids to fill via `set_flow_module_code`. Only appended on a **successful** save (`writeFlowDraft` reports conflicts/errors as `{success:false}` without throwing).
- **Actionable parse error** — the `Invalid JSON for modules/...` error now explains that a quote/newline-heavy `content` is the usual cause and points to the empty-content + `set_flow_module_code` pattern.
- **Softened guidance** — `write_flow` inlines each rawscript body by default (cheap for capable models); the empty-content + fill path is reserved for bodies that are hard to escape or when a `write_flow` call comes back with a parse error. This avoids taxing capable models with an extra round-trip on every flow.
- **`inline_script.<id>` placeholder resolution** (Guilhem) — `write_flow` resolves placeholders against the current draft/deployed flow so an existing-flow overwrite can keep unchanged bodies without re-sending them; unresolved own-id placeholders become empty bodies to fill; anything else is rejected. `patch_flow_json` gets the same placeholder handling.

### Why not raw apps

Raw apps already use per-file tools (`write_app_file`/`write_app_runnable`) where content is a top-level string param — one escaping layer, no nesting. `write_flow` was the only "write it all at once" tool carrying executable code inside a structure argument; this aligns flows with that model.

## Benchmark (global mode A/B)

Retained focused case `global-flow-quote-heavy-inline-code` (single-step, quote-heavy HTML body), run **before = `main` vs after = this PR's head guidance (`ad1209e0e5`)**, same case, judge on:

| model | before (main) | after (head) | Invalid-JSON errors | judge (b/a) | total tokens (b/a) |
|---|---|---|---|---|---|
| Sonnet | inline (`write_flow`) | **inline** (`write_flow`) | 0 / 0 | 100 / 100 | 122226 / 122483 |
| GPT-4o | inline | fill-path (`write_flow` + `set_flow_module_code`) | 0 / 0 | 97 / 95 | 123349 / 161365 |

**Honest read:** with the softened guidance at head, **capable models (Sonnet) inline exactly like `main` — no regression** (the earlier aggressive-guidance +58%-token cost on Sonnet is gone). Weaker models (4o) may self-select the empty-content + `set_flow_module_code` escape hatch for a quote-heavy body (~+38k tokens, no reliability difference at n=1). The targeted GPT-5 inline-escaping failure does **not reproduce** on any model available in the harness (gpt-5.5 stalls before `write_flow`; Sonnet/Gemini/4o all escape inline cleanly on `main`), so there's **no pass-rate delta**. The change is a **low-cost safety net** with no regression; the retained eval case guards that flow-creation-with-code keeps working. (Earlier exploratory runs on Gemini 3 flash/pro and a 3-step case reached the same conclusion — no repro, no regression.)

## Changes

- `frontend/src/lib/components/copilot/chat/global/core.ts` — empty-body warning (write_flow + patch_flow_json), success-only application, parse-error hint, softened guidance, `inline_script` placeholder resolution.
- `frontend/src/lib/components/copilot/chat/global/core.test.ts` — regression tests for the warning (top-level/nested/preprocessor/failure, populated-suppressed), no-warning-on-failed-save, malformed-JSON hint, and placeholder resolution.

## Test plan

- [x] `vitest run` copilot global tests (169 passing)
- [x] `svelte-check` clean for the changed file
- [x] Global-mode A/B run manually against PR head (Sonnet, GPT-4o) — no regression; see Benchmark
- [ ] No visible UI change (copilot tool internals + system-prompt text only)


Guilhem-lm:
## ai_evals A/B — global mode (sonnet, 1 attempt/case)

Cases: `global-flow-quote-heavy-inline-code` (new), `global-test3-flow-create`, `global-test5-existing-flow-inline-code-edit`.
**Before** = PR base `68b1fcc5c` (copilot sources checked out at base, eval case from HEAD). **After** = `708ec5abd`.

| Metric | Before | After |
|---|---|---|
| Pass rate | 3/3 | 3/3 |
| Judge scores | 100 / 97 / 100 | 100 / 97 / 100 |
| Final context (avg, passed) | 40,763 tok | 41,140 tok (+0.9%) |
| Cumulative prompt (avg, passed) | 122,826 tok | 153,729 tok |

Trajectory diff (the behavior this PR targets):

| Case | Before | After |
|---|---|---|
| quote-heavy-inline-code | `write_flow` with code inlined in modules JSON (2 calls) | `write_flow` (empty body) → `set_flow_module_code` (3 calls) |
| test3-flow-create | same as above | same as above |
| test5-inline-code-edit | `read_workspace_item` → `read_flow_module_code` → `set_flow_module_code` | identical to before |

Interpretation: after the change, inline code lands out-of-band via `set_flow_module_code` instead of being triple-escaped into the modules JSON string — the deterministic path this PR is for (the before-run's inline write happened to survive escaping on this attempt; that failure mode is intermittent by nature). The judge confirmed the after-run drafts contain the real multi-line HTML bodies, not placeholders. Cost: final context is flat (+0.9%, the added guidance bullets); cumulative prompt rises by one extra iteration on creation flows — the intended reliability tradeoff. The existing edit path is unchanged.

Also addressed the Codex nit: the 5-line `patch_flow_json` comment is condensed to 3 lines in `708ec5abd`.

rubenfiszel:
Addressing the Codex review:

**1. [P1] Warn only after a successful save** — Fixed. `appendEmptyInlineScriptWarning` now parses the `writeFlowDraft` result and folds the warning into the payload **only when `success === true`**; conflict/persistence failures (`{success:false}`) are returned untouched, so the model is never told to `set_flow_module_code` on a flow that wasn't saved.

**2. [P2] Global-mode A/B** — Ran manually (Sonnet, Gemini 3 flash, Gemini 3.1 pro, GPT-4o), before vs after, judge on. The targeted GPT-5 inline-escaping failure does **not reproduce** on any model available in the harness (gpt-5.5 stalls before `write_flow`; the rest escape inline cleanly even on `main`), so there's no pass-rate delta to show. The A/B did confirm: the initial aggressive guidance was a real token regression on capable models (now softened), no regression from the final version, and the escape-hatch path activates correctly (4o proactively used empty + `set_flow_module_code`). Full table in the PR description. Because a permanent eval case guards nothing measurable here, it was dropped in favor of the unit tests.

**3. [P2] Regression test** — Added `core.test.ts` coverage: the empty-body warning across top-level/nested/preprocessor/failure modules (populated bodies suppressed), no-warning-on-failed-save, and the malformed-JSON escaping hint.

PR #10261, opened 2026-07-22
fix(ai-sessions): show preview chip on write tools, not open_preview

Guilhem-lm:
## Summary

#10254 introduced a "Preview" chip on AI-session tool calls, meant to let the user open/focus an item's side-panel preview from the create/update tool that produced it. In practice the chip never appeared on write tools and only appeared on `open_preview` — the one tool where it is redundant, since that tool just opened the preview itself.

Root cause: `maybeAttachPreviewCard` gated on `ctx.sessionId`, but write tools pass the raw tool-call context, which carries the session id in `helpers` — so the guard always bailed on the write path. `open_preview` read the id correctly via `sessionIdFromCtx`, hence the inverted behavior.

## Changes

- `maybeAttachPreviewCard` falls back to `sessionIdFromCtx(ctx)` (the raw tool ctx's `helpers`), so every session-chat write that lands a previewable item (script, flow, raw app) gets the chip — creates, updates, `edit_script`, `patch_flow_json`, and the app write tools included. The global side-panel chat still gets none (`helpers.sessionId` is only set for session chats).
- Removed the chip from the `open_preview` tool and reverted `openSessionPreview` to a plain string return — the `{opened, message}` struct existed only to feed the chip.
- Dropped the now-unreachable `pipeline` kind from `PreviewCardKind` and its dead icon/label branches (writes only resolve to script/flow/raw_app).
- Rewrote `ToolPreviewCard.svelte` to use the design-system `Button` component (`variant="default"`, `unifiedSize="2xs"`, `RowIcon` via `startIcon.props`, `PanelRight` as `endIcon`) instead of a raw `<button>`.

## Screenshots

`write_script` and `edit_script` tool calls now carry the Preview chip on their header row; the `open_preview` call below them has none. Clicking the chip opens/focuses the script tab in the side panel.

preview-chip-button

## Test plan

- [x] `npm run check:fast` clean; `shared.test.ts` 38/38 passing; svelte-autofixer reports no issues
- [x] In a real AI session (Playwright, `debug-sessions` workspace): `write_script` and `edit_script` calls show the chip; clicking it opens, then focuses, the script preview tab
- [x] `open_preview` tool call shows no chip
- [ ] Global side-panel chat (non-session) shows no chip on writes — guarded by `helpers.sessionId` only being set for session chats

### DIFF
--- a/frontend/src/lib/components/copilot/chat/ToolExecutionDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/ToolExecutionDisplay.svelte
@@ -97,7 +97,7 @@
 			</button>
 		{/snippet}
 
-		<!-- Discrete preview chip for an item a tool created/updated/opened, pinned to
+		<!-- Discrete preview chip for an item a tool created/updated, pinned to
 		     the right of the header row. Rendered inline (not gated on expand) so it
 		     stays visible after the tool collapses. -->
 		{#if showPreviewChip && message.previewCard}
--- a/frontend/src/lib/components/copilot/chat/ToolPreviewCard.svelte
+++ b/frontend/src/lib/components/copilot/chat/ToolPreviewCard.svelte
@@ -35,5 +35,7 @@
 	endIcon={{ icon: PanelRight }}
 	wrapperClasses="shrink-0"
 >
-	Preview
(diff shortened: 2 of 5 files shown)


Guilhem-lm:
✅ Review round clean @ d3a9b861003a2168106bbf658c014f1237d253a9

PR #10262, opened 2026-07-22
fix(jobs): enforce self_approval_disabled on the UI resume path

rubenfiszel:
## Summary
Feedback reported that `self_approval_disabled` appears broken: an operator can click the "Resume" button in the run detail UI and the flow continues, even though they triggered the run.

The "Resume" button calls the `resume_suspended` endpoint (`POST /w/{w}/jobs_u/flow/resume_suspended/{job_id}`). Its authorization block short-circuits on `is_owner` (and `is_admin`) and **skips the approval-condition checks entirely**. So a user who owns the flow path (their personal `u/<user>/…` space, or a folder they own) — the same person who triggered the run — could self-approve despite `self_approval_disabled`. The dedicated owner endpoint (`resume_suspended_flow_as_owner`) already enforces the check for owners, so the two resume paths were inconsistent. Per the existing design, only **admins** are meant to bypass self-approval.

A second, latent gap: the worker only persisted `approval_conditions` (which carries `self_approval_disabled`) when `user_auth_required` was set. A raw-flow / CLI author who sets `self_approval_disabled` without `user_auth_required` had the restriction silently dropped at the resume boundary. (The flow-editor UI couples the two toggles, so this only affected non-UI authors.)

Fixes WIN-2223

## Changes
- Extract `require_not_self_approval` and enforce it **before** the owner shortcut in `resume_suspended` and in `can_approve_step` (so the standalone approval page also hides the button for a self-approving owner), matching `resume_suspended_flow_as_owner`. Admins remain exempt.
- Refactor `conditionally_require_authed_user` to reuse the same helper (single source of truth).
- Persist `approval_conditions` when `self_approval_disabled` is set even without `user_auth_required` (`worker_flow.rs`); `user_groups_required` still only applies together with `user_auth_required`.

## Test plan
- [x] `test_self_approval_disabled_blocks_ui_resume_for_owner` (new) — owner-triggerer resuming via the UI endpoint gets 403 when `self_approval_disabled=true` (this failed before the fix, returning 201).
- [x] `test_self_approval_disabled_blocks_owner_resume` — owner endpoint still blocks self-approval.
- [x] `test_self_approval_allowed_when_not_disabled` — self-approval allowed when the flag is off.
- [x] `test_different_user_can_approve_when_self_approval_disabled` — a different user can still approve.
- All 4 pass under `--features enterprise,deno_core,private,license`.

Pure backend change; no frontend code touched. The `user_auth_required` toggle is enterprise-gated and can't be exercised on the CE dev build, so validation is via the integration tests above rather than a browser walkthrough.

---
## Summary by cubic
Enforce `self_approval_disabled` on identity-based resume paths (UI “Resume” and approval page) so the triggerer can’t self-approve even if they own the flow; only admins can bypass. Fixes WIN-2223; the secret HMAC resume path stays capability-based and skips approval checks.

- **Bug Fixes**
  - Enforce `self_approval_disabled` before the owner shortcut in `resume_suspended` and `can_approve_step`, matching `resume_suspended_flow_as_owner`.
  - Keep `resume_suspended_job` (HMAC URL) and token-only resumes on `resume_suspended` capability-based; approval conditions are not applied there.
  - Persist `approval_conditions` when `self_approval_disabled` is set even without `user_auth_required`, so the restriction isn’t dropped at resume.

- **Refactors**
  - Extract `require_not_self_approval` for a single source of truth and reuse it in `conditionally_require_authed_user`.
  - Add `extract_approval_conditions` (handles WAC vs. classic) and reuse in `resume_suspended`.

<sup>Written for commit e45506560b8183ed26a1785cd831c21729f67561. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/suspend_resume.rs
+++ b/backend/tests/suspend_resume.rs
@@ -352,13 +352,10 @@ mod suspend_resume {
         Ok(())
     }
 
-    /// Test that self-approval is blocked on the UI resume endpoint (POST
-    /// /jobs/flow/resume_suspended/:job_id) when self_approval_disabled is true, even for a
-    /// user who owns the flow path.
-    ///
-    /// This is the endpoint the "Resume" button in the run detail UI calls. Its owner shortcut
-    /// used to skip the approval-condition checks entirely, so the flow owner/operator who
-    /// triggered the run could self-approve despite self_approval_disabled.
+    /// The UI "Resume" button (POST /jobs_u/flow/resume_suspended/:job_id) must reject the
+    /// triggerer approving their own self_approval_disabled step even when they own the flow
+    /// path: only admins are exempt from the self-approval restriction, so owning the runnable
+    /// does not grant the right to self-approve.
     #[cfg(feature = "enterprise")]
     #[cfg(feature = "deno_core")]
     #[sqlx::test(fixtures("base"))]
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ e45506560b8183ed26a1785cd831c21729f67561

PR #10267, opened 2026-07-22
feat(ai-chat): email triggers in flow/script chat + trigger-intent eval guards (WIN-2228)

rubenfiszel:
## WIN-2228: flow AI chat trigger handling

Investigation of whether the flow-editor AI chat understands it should build a flow **and** its associated triggers, plus a concrete gap fix.

### Investigation findings

- **Tools exist and are wired.** `create_schedule` / `create_trigger` (10 kinds: http, websocket, kafka, nats, postgres, mqtt, amqp, sqs, gcp, azure) in `workspaceTools.ts`, shared by flow + script chat. Both `requiresConfirmation: true` (already asks the user) and deployment-gated.
- **Implicit intent already works.** An A/B on the current prompt passes 12/12 (6 runs each) on two new implicit-intent cases (schedule + http) and calls the right tool every time. A trial prompt addition spelling out a deployment prerequisite *regressed* the http case 6/6 → 2/6 (the model deferred), so **no prompt change ships** — only the eval guards.
- **Email was the real gap.** `create_trigger` had no `email` kind, so for "run when an email is received" the model substituted an HTTP trigger and mislabeled it as email (verified empirically). The backend supports email triggers and the chat's open-resource drawer was already wired for them — only the tool-level create support was missing.

### Changes

**1. Trigger-intent eval guards** (`flow-test17/18`): implicit schedule + http cases as regression guards. They earned their place — they caught the prompt-change regression above.

**2. Email trigger support** in `create_trigger` (flow + script chat):
- Added `email` as a kind (generator + regenerated Zod + `triggerConfigs` → `EmailTriggerService.createEmailTrigger`).
- **Availability-aware, role-aware guard:** email triggering only works once an instance superadmin runs an SMTP server and sets the `email_domain` global setting. The tool reads `email_domain` (readable by any authed user; null when unset) and:
  - **configured** → creates the trigger and reports the resulting inbound email address;
  - **not configured + superadmin** → guidance to enable it in Instance settings (SMTP + `email_domain`), with docs link;
  - **not configured + regular user** → guidance to ask an instance superadmin, with docs link.
- `userStore` / email-address helper are lazy-imported so the chat tools module doesn't pull in the heavy `$lib/stores` graph at load (this also fixed a test import-time regression).
- Guard case `flow-test19` pins that "receive email" → `create_trigger(kind=email)`.

### Validation

- Unit tests: `shared.test.ts` 40 pass (incl. both email branches — superadmin/non-superadmin hint, and configured create-with-address), `AIChatManager.test.ts` 119 pass.
- ai_evals: `flow-test17/18/19` pass **9/9** (`sonnet`, workspace `admins`); the model now calls `create_trigger(kind=email)` 3/3 on a natural email prompt (was substituting HTTP before).
- A/B table (implicit schedule / http), 6 runs each: baseline 6/6 · 6/6; deployment-caveat prompt 6/6 · 2/6 (rejected); no prompt change shipped.
- `npm run check:fast` clean on all changed files (only a pre-existing, unrelated `modern-screenshot` error in `RawAppEditor.svelte`).
- `system_prompts/check-freshness.sh` up to date after regenerating the Zod schema.

Email-trigger CRUD is `#[cfg(feature = "private")]`-gated, so the create path can't be exercised on the plain-`quickjs` dev backend; the not-configured path (the key new UX) is validated live, and the configured create path is covered by the unit test. No user-facing UI surface changed (the drawer was already wired), so no screenshots apply.

Fixes WIN-2228

### DIFF
--- a/ai_evals/cases/flow.yaml
+++ b/ai_evals/cases/flow.yaml
@@ -497,6 +497,14 @@
           - u/
         stringMustNotStartWithAnyOf:
           - schedules/
+      - tool: create_schedule
+        field: schedule
+        stringIncludesAnyOf:
+          - 30 7
+      - tool: create_schedule
+        field: timezone
+        stringIncludesAnyOf:
+          - UTC
   skipJudge: true
   judgeChecklist:
     - "a schedule is created for the flow that runs daily at 07:30 UTC"
@@ -518,6 +526,22 @@
           - u/
         stringMustNotStartWithAnyOf:
           - schedules/
+      - tool: create_trigger
+        field: kind
+        stringStartsWithAnyOf:
+          - http
+      - tool: create_trigger
+        field: config.http_method
+        stringIncludesAnyOf:
+          - post
+      - tool: create_trigger
+        field: config.authentication_method
+        stringIncludesAnyOf:
+          - none
+      - tool: create_trigger
+        field: config.route_path
+        stringIncludesAnyOf:
+          - ai-evals/order-processing-implicit
   skipJudge: true
   judgeChecklist:
     - "an HTTP trigger is created for the flow that accepts unauthenticated POST requests"
--- a/ai_evals/cases/flow.yaml
+++ b/ai_evals/cases/flow.yaml
@@ -480,3 +480,44 @@
   judgeChecklist:
     - "the flow includes a final top-level step named `webhook_response`"
     - "`webhook_response` returns `ok: true` and the order summary"
+
+- id: flow-test17-implicit-schedule-intent
+  prompt: |-
+    I want this order processing flow to run on its own every morning at 07:30 UTC.
+    Set that up for me. Do not ask me for the flow path.
+  initial: ai_evals/fixtures/frontend/flow/initial/scheduled_order_flow.json
+  toolExpect:
+    requiredToolsUsed:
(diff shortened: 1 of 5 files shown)


rubenfiszel:
Addressed the Codex review (commit `5ba5e70`):

- **[P1] `workspaced_local_part` omitted from email create request** — real bug. The column is `BOOLEAN NOT NULL`, and the model may omit the field, so the request could be rejected by the backend. Now defaulted to `false` on the request body *before* `EmailTriggerService.createEmailTrigger` is called (previously the `?? false` only applied when formatting the success address). Added a regression assertion in the happy-path unit test that the create request carries `workspaced_local_part: false`.
- **[P2] Implicit-intent evals only checked tool + path prefix** — tightened the deterministic `toolCallArgs`:
  - `flow-test17` (schedule) now asserts the cron time (`30 7`) and `timezone` = `UTC`;
  - `flow-test18` (http) now asserts `kind`=http, `config.http_method`=post, `config.authentication_method`=none, and `config.route_path`.

Validation: `shared.test.ts` 40 pass, `check:fast` clean on changed files, and `flow-test17/18/19` still pass **9/9** (`sonnet`) with the stricter assertions.

PR #10269, opened 2026-07-22
fix: manual resource type sync fetches from hub first, cache as fallback

rubenfiszel:
## Summary

The superadmin **"Synchronize resource types"** button
(`POST /api/settings/sync_cached_resource_types`) was **cache-first**: it read the
on-disk `hub_rt/resource_types.json` cache and only hit the hub when *no* cache file
existed. That cache is refreshed by an external daily `cache-rt` cron, so a
brand-new hub type (e.g. `amqp`, added this week) could not be pulled on demand,
clicking sync replayed the stale cache, returned `Synced 0 resource types
(266 unchanged)`, and the type never landed in the `admins` workspace. The UI then
kept showing *"Resource type 'amqp' not found in your workspace"* even after the
superadmin clicked sync. Deleting the cache file was the only workaround.

This makes the **manual** endpoint **hub-first with the cache as a genuine fallback**,
and lets the frontend request a specific type by name so the not-found call-to-action
becomes deterministic.

## Changes

- **Hub-first manual sync** (`windmill-api-settings`): the handler now fetches the live
  list from the hub and upserts it into `admins`. It falls back to reading the on-disk
  cache **only** when the live hub fetch fails (airgapped install / network error),
  logging which path it took.
- **Targeted by-name refresh**: the endpoint accepts an optional `?name=<type>` query
  param. When a specific type is requested and is still absent from the hub after syncing,
  it returns an explicit `404 resource type '<name>' not found on the hub` instead of a
  silent `Synced 0`. The no-name (global) call works exactly as before.
- **Frontend**: `SyncResourceTypes.svelte` gains an optional `resourceType` prop and
  appends `?name=` when set. The three not-found sites (`ResourceForm`, `AppConnectInner`,
  `ApiConnectForm`) pass the specific type in scope; the global instance-settings button
  and the browse-all connect step stay name-less.

## Left unchanged (intentional)

The **startup / offline** path is still cache-based: `cache_hub_resource_types()`
(the `cache-rt` CLI subcommand that *writes* the cache), `sync_cached_resource_types(db)`,
and its `SYNC_CACHED_RT` gate in `backend/src/main.rs`. Airgapped installs and the daily
cron rely on the cache, so only the manual HTTP endpoint changed. **The manual endpoint
does not write the cache** — refreshing it stays owned by the cron / startup path, which
avoids adding a non-atomic write (and a partial-response overwrite) to the offline
fallback. The startup sync is upsert-only, so a manually-synced type is never lost to a
stale cache on restart. The endpoint stays `require_super_admin`-gated.

## Screenshots

Reproduced with a **stale cache present** (266 types, `amqp` excluded) and `amqp` removed
from `admins`. Opening an `amqp` resource shows the not-found branch; clicking sync issues
`POST …/sync_cached_resource_types?name=amqp`, the endpoint fetches live from the hub, and
the form resolves to the full `amqp` schema. Verified the on-disk cache file is left
byte-for-byte unchanged by the manual sync.

Not-found branch with the by-type sync button:

rt-notfound-sync

Resolved form after the hub-first sync landed `amqp`:

rt-resolved-amqp

## Test plan

- [x] `cargo check -p windmill-api-settings` (default features) passes
- [x] `npm run check:fast` clean for the touched files; svelte autofixer clean
- [x] Stale cache present + `amqp` missing → endpoint now syncs `amqp` (was `Synced 0`), and the on-disk cache file is left untouched (md5 identical before/after)
- [x] `?name=amqp` (present) → `200`; `?name=<bogus>` → `404 resource type '<bogus>' not found on the hub`
- [x] UI: not-found sync button posts `?name=amqp`, form resolves to the amqp schema; global button still posts name-less

### DIFF
--- a/backend/windmill-api-settings/src/lib.rs
+++ b/backend/windmill-api-settings/src/lib.rs
@@ -1989,10 +1989,10 @@ async fn sync_cached_resource_types(
     // on-disk cache is only a fallback for when the hub is unreachable (airgapped
     // installs / network error); refreshing it is left to the daily cache-rt cron and
     // the startup sync in main.rs, which own the offline path.
-    let resource_types = match fetch_resource_types_from_hub().await {
+    let (resource_types, from_hub) = match fetch_resource_types_from_hub().await {
         Ok(types) => {
             tracing::info!("Fetched {} resource types live from the hub", types.len());
-            types
+            (types, true)
         }
         Err(hub_err) => {
             tracing::warn!(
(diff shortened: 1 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 6c6ee400be717bb47755e41c8a431b093497a7ee

PR #10270, opened 2026-07-22
feat(ai): improve data-pipeline building in AI sessions (prompt + evals + e2e)

rubenfiszel:
## Summary

Fixes WIN-2229. Investigation into how AI sessions build data pipelines, the tools available to them, and the e2e coverage, plus the concrete improvements it produced: two prompt fixes, an eval-fidelity fix, a new complex-pipeline eval, and a deterministic browser e2e.

### How pipelines get built from a session today (and what's gated)

- A data pipeline is a DAG of independent annotated scripts (`-- pipeline`, `-- on <ref>`, `-- materialize <asset>`) in one folder, wired by storage assets. It is not a flow.
- The global/session chat builds pipelines with the generic `write_script` draft tools plus `get_instructions subject: pipeline` (guidance from `system_prompts/base/pipeline-base.md` -> `getPipelinePrompt`). The graph-aware tools (`build_pipeline_node`, `get_pipeline_graph`, `test_pipeline_node`) only activate when a `/pipeline/<folder>` editor is open.
- **Pipeline building inside AI sessions is disabled by default** (alpha): gated behind `localStorage.wm_dev_session_pipelines='1'` (`global/pipelineGate.ts`). By default the session prompt tells the user pipelines are alpha and "will be handled by the chat soon," `get_instructions subject: pipeline` returns the gated message, and the session preview router won't route pipeline previews. The standalone global side-panel chat and the `/pipeline` editor keep pipeline support, and the eval harness exercises the ungated path — so **the evals are the model-in-the-loop e2e for AI-built pipelines**.

### Findings that drove the changes

1. **The LLM judge grades pipelines with a flow mental model.** `global-test-pipeline-create-node` produced the textbook-correct node (a bare `SELECT` under `-- pipeline` / `-- on schedule` / `-- materialize ducklake://...`), passed every deterministic check, and still failed on the judge (72 < 80): it penalized `-- on schedule` for not being a "real" trigger config and `-- materialize` + bare SELECT for not being a hand-written `CREATE TABLE`/`INSERT` — exactly the contract, inverted. The judge has no pipeline domain context.
2. **No DuckDB/materialize steer.** The prompt was neutral ("duckdb or postgresql ... bun/python3 for general transforms"), not a default toward DuckLake materialization.
3. **No storage-readiness awareness.** Datatables have a `list_datatables` tool + a blocking-prerequisite prompt ("if none configured, tell the user to set one up in workspace settings and stop"). DuckLake/object-store pipelines had **no** equivalent: nothing about storage readiness in the prompt, and no `list_ducklakes`/object-storage tool wired into the chat. So a session would draft a DuckLake pipeline that cannot run and never tell the user storage must be configured.

## Changes

- **`list_ducklakes` chat tool** (`frontend/.../ducklakeTools.ts`, wired into the global tool set): the pipeline counterpart to `list_datatables`. Lists the workspace's configured DuckLake catalogs; when none exist it returns an actionable message (drafting still allowed, but the pipeline can't run until storage is set up, with role-appropriate next steps). This gives the chat a real storage-readiness signal instead of a blind advisory. In the eval run all three cases called it unprompted before/while building.
- **`system_prompts/base/pipeline-base.md`** (+ regenerated `auto-generated/prompts.ts`):
  - "Default to DuckDB + DuckLake": a node that produces a table should default to `duckdb` materializing into DuckLake; use postgres/data-tables only for OLTP-style row mutations and bun/python only for non-tabular glue (and still land tabular output in DuckLake).
  - "Storage prerequisites": a DuckLake pipeline needs workspace object storage + a DuckLake catalog. Check with `list_ducklakes`; when none is configured (or the user hits storage errors), warn and give role-appropriate next steps (admin: Workspace settings -> Object Storage, then a DuckLake catalog; non-admin: ask an admin). Drafting is not blocked, so it does not regress the build path.
- **`ai_evals/cases/global.yaml`**:
  - Rewrote the `judgeChecklist` items across all three `global-test-pipeline-*` cases to encode the declarative contract inline (trigger = `-- on` annotation; `-- materialize` = bare SELECT with no hand-written DDL; S3/Parquet is not materialize). This is the eval-fidelity fix for finding 1.
  - Added `global-test-pipeline-complex-incremental`: a three-node DuckLake pipeline (scheduled ingest that materializes raw orders, an incremental partitioned daily rollup that appends one day per run, a Parquet-to-S3 export). Deterministic checks stay narrow; semantics live in the judge checklist. The incremental item grades behavior (append one day, no full rebuild) and accepts either the `-- partitioned daily` / `{partition}` idiom or an equivalent current-day filter.
  - Hardened `global-test-pipeline-two-node-chain`: previously only `draftCountAtLeast: 2` + "no flow"; now also requires a pipeline-annotated DuckLake script draft at the folder.
- **`frontend/e2e/pipeline.spec.ts`** (new): a deterministic Playwright e2e. Seeds the annotated scripts an AI-built DuckLake pipeline would produce (via the API), then asserts the `/pipeline/<folder>` editor derives the full lineage DAG from the annotations alone: both pipeline-member script nodes, the DuckLake/S3 asset nodes, all four data edges, and the missing-schedule-trigger edge.

## Screenshots

The e2e asserts the graph the editor derives from the seeded annotations:

pipeline-graph

(The only `frontend/` change is this test spec; it has no visible-UI effect of its own.)

## Test plan

Ran the pipeline cases against `sonnet` through the local backend AI proxy, judge `claude-sonnet-4-6`.

- [x] `bun test core/` in `ai_evals` -> 76 pass (case schema valid); prompts regenerated (`generate.py`), diff limited to `prompts.ts`.
- [x] Judge-context fix: `create-node` went from judge 72 with all deterministic checks green (the false negative) to grading correctly (97-100).
- [x] Prompt A/B (before -> after, 3 cases x2 runs): pass rate 66.7% -> 83.3%, no build-case regression (create-node 2/2, two-node 2/2), finalContext 34255 -> 34894 (+639, the guidance tax). The single remaining complex failure is a genuine model error (wrote `-- materialize s3://...parquet`), which the checklist is meant to catch.
- [x] `list_ducklakes` tool: unit test (`ducklakeTools.test.ts`) passes; live endpoint verified; re-ran the three pipeline cases with the tool wired in and all three called it unprompted, still built the drafts (no regression, create-node judge 100 / two-node 92), finalContext +~205 tokens.
- [x] Browser e2e: `BASE_URL=http://localhost:3260 SKIP_HUB_SYNC=true npx playwright test pipeline.spec.ts --project=chromium` -> 1 passed.

- A deterministic eval for the warn-behavior needs the harness to seed "no DuckLake configured" (it currently can't); the `list_ducklakes` tool now makes the check real, but pinning the warn-message in an eval is deferred.
- Wiring the graph-aware pipeline tools into sessions (behind the alpha gate) so session pipeline builds get canvas seeding and `get_pipeline_graph`, rather than plain script drafts.

### DIFF
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -1484,9 +1484,15 @@
       - deploy_workspace_item
       - delete_workspace_item
   judgeChecklist:
+    # A pipeline node is DECLARATIVE: triggers are declared by `-- on <ref>`
+    # annotations (the trigger row is created separately) and a `-- materialize`
+    # output is a MANAGED write where the body is a bare SELECT that the runtime
+    # wraps in the create/replace. Do not expect a separate trigger config or a
+    # hand-written CREATE TABLE / INSERT — those would be wrong for a materialize node.
     - builds a data pipeline node as a script (not a flow)
     - marks the script as a pipeline member with the pipeline annotation in the script's comment syntax (`-- pipeline` for a DuckDB/SQL node, not `// pipeline`)
-    - declares a schedule trigger and writes its output to a managed DuckLake table
+    - declares the schedule trigger with the `-- on schedule` annotation comment (this annotation is the correct and complete way a pipeline node binds a schedule; no separate trigger configuration is expected)
+    - declares the managed DuckLake output with `-- materialize ducklake://<table>` and writes the body as a bare SELECT (materialize is a managed write, so the node correctly does NOT hand-write its own CREATE TABLE / INSERT)
     - leaves the result as an AI draft and does not deploy or save it
 
 - id: global-test-pipeline-two-node-chain
@@ -1500,6 +1506,12 @@
     maxTurns: 14
   validate:
     draftCountAtLeast: 2
+    requiredDrafts:
+      - type: script
+        pathStartsWith: f/evals/global/
+        valueIncludes:
+          - pipeline
+          - ducklake
     forbiddenDrafts:
       - type: flow
         pathStartsWith: f/evals/global/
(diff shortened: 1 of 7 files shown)


rubenfiszel:
✅ Review round clean @ da975bc066f943e9e28407c37c5b19e10ff20e37 (nit-only verdicts at 7eb7b879d4: codex + claude 'mergeable, address nits', pi 'good to merge'; both nits addressed in da975bc066 — dropped the phantom `list_ducklakes` tool reference and trimmed narration-only comments)

rubenfiszel:
✅ Review round clean @ 7ccc7d80635d93447aab2be4d4b16e39b1d7fde5 (nit-only verdicts at d77f0dfedb: codex + claude 'mergeable, address nits', pi 'good to merge', cubic 1 nit; all addressed in 7ccc7d8063: DuckDB-default section now uses `-- materialize` SQL syntax, and the postgresql bullet grammar polished)

PR #10271, opened 2026-07-22
refactor: custom-instance datatable connection handling

diegoimbert:
## Summary

Refactors how custom-instance data tables establish their Postgres connections, in two independent places:

- **DuckDB executor** — data tables are attached through a DuckDB `TEMPORARY SECRET` carrying the connection parameters, instead of an inline connection string in the `ATTACH` statement. Only `sslmode` is passed in the attach string.
- **Postgres triggers** — triggers on custom-instance data tables now connect through a dedicated `custom_instance_replication_user` role (with its own auto-generated password stored in `global_settings.custom_instance_pg_databases.replication_user_pwd`) rather than `custom_instance_user`. BYO-Postgres data tables are unchanged.

A migration provisions the new role and grants it membership in `custom_instance_user`; server boot normalizes `custom_instance_user`'s role attributes to converge if the one-shot migration couldn't apply (it swallows errors so it never aborts startup).

## Changes

- `duckdb_executor.rs`: `transform_attach_datatable` now emits `CREATE OR REPLACE TEMPORARY SECRET` + `ATTACH ... (TYPE postgres, SECRET ...)`; sslmode whitelisted to disable/require/prefer. Unit tests added.
- New migration `custom_instance_replication_user` (+ down) provisioning the role.
- `windmill-common/utils.rs`: `refresh_custom_instance_replication_user_pwd` + `get_custom_pg_instance_replication_password` (self-heals the role on first use).
- `windmill-common/workspaces.rs`: `get_datatable_replication_resource_from_db_unchecked` resolves the replication role for instance-type data tables.
- `windmill-trigger-postgres/lib.rs`: `datatable://` resolution uses the replication resource.
- `windmill-api-settings/lib.rs`: DB setup provisions the replication role and normalizes attributes.
- `windmill-api/live_migrations.rs`: boot-time attribute normalization.
- Frontend: add `replication_user_pwd` to the instance-settings redaction list; update the custom-instance DB wizard step text.
- SQLx offline cache regenerated (EE-safe procedure, zero net deletions).

## Test plan

- [x] `cargo test -p windmill-worker --features duckdb pg_secret` — secret-attach unit tests pass
- [x] `cargo check` on all touched crates; `npm run check:fast`
- [x] Postgres trigger on a custom-instance data table: publication/slot owned by the replication role, walsender authenticates as it, insert → job fires (verified on a `postgres_trigger` build)
- [x] Boot-time normalization: manually drift the attribute, restart, confirm it re-converges
- [ ] DuckDB job doing `ATTACH 'datatable://...'` end-to-end through the worker (SQL shape + redaction verified via duckdb CLI + unit tests; live-worker job run still pending)
- [x] Upgrade path: existing trigger created before this change keeps firing after migration

## Notes

No screenshots: the two frontend edits are a redaction-list entry (no visible UI change) and one wizard description string; both live on settings pages that display sensitive instance/DB values, which must not be captured to the public screenshot repo.

### DIFF
--- a/backend/.sqlx/query-ffbe9fe78a7fc0e95a5a29c8d17f4367ef6b6cc3de20da6ef8c98f679b832240.json
+++ b/backend/.sqlx/query-ffbe9fe78a7fc0e95a5a29c8d17f4367ef6b6cc3de20da6ef8c98f679b832240.json
@@ -1,0 +1,20 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT value->>'replication_user_pwd' FROM global_settings WHERE name = 'custom_instance_pg_databases';",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "?column?",
+        "type_info": "Text"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "ffbe9fe78a7fc0e95a5a29c8d17f4367ef6b6cc3de20da6ef8c98f679b832240"
+}
--- a/backend/migrations/20260716152346_custom_instance_replication_user.down.sql
+++ b/backend/migrations/20260716152346_custom_instance_replication_user.down.sql
@@ -1,0 +1,14 @@
+DO $$
+BEGIN
+    IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'custom_instance_user') THEN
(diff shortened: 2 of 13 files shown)


diegoimbert:
[P1] Rolling-upgrade breakage for old trigger workers: It's fine, we will need workers upgrade anyway
[P1] Broken rollback after new-role publications: we will not bother for the down migration
[P1] Missing authorization contracts on new secret helpers: everyone needs to use that password but it must not be exposed (and it is not). I added doc comments

PR #10274, opened 2026-07-23
fix(jobs): sanitize NUL in completed job result before jsonb insert

rubenfiszel:
## Summary

A job whose result contains a real NUL (U+0000) serializes to a `\u0000` JSON escape that the `jsonb`-typed `v2_job_completed.result` column rejects with Postgres `22P05` ("unsupported Unicode escape sequence"). This aborts the `INSERT` in `commit_completed_job`, which then retries 10 times and leaves the job unable to complete (surfaced as `Could not add completed job <id>: ... unsupported Unicode escape sequence`).

The fix sanitizes the serialized result immediately before the insert, with effectively zero overhead on the common NUL-free path.

## Changes

- **Promote `strip_json_nul` into `windmill-common`** (`utils.rs`): `fn strip_json_nul(&str) -> Cow<str>` — a `contains("\\u0000")` fast guard returns the input borrowed when clean; only a genuine odd-parity NUL escape triggers the O(n) rebuild. `Cow::Owned` is returned **only** when a NUL was actually stripped, so a legitimate `\\u0000` (escaped backslash + literal text) borrows through untouched. Replaces the two duplicated copies previously in `windmill-api/src/drafts.rs` (`strip_json_nul`) and `windmill-api/src/apps.rs` (`strip_null_chars`); both call sites now use the shared helper.
- **Add `serialized_json()` to the `ValidableJson` trait** (`windmill-queue/src/jobs.rs`): `Box<RawValue>` returns `Cow::Borrowed(self.get())` (zero-cost, already serialized); other impls serialize on demand via `to_raw_value`.
- **`commit_completed_job`** binds `strip_json_nul(result.serialized_json())` as `$3::text::jsonb` in both the `INSERT ... SELECT` and the `ON CONFLICT ... result = $3` (was `result as Json<&T>`). Stored data is unchanged (Postgres parses JSON text into `jsonb` identically); `wm_labels`/`result_metadata` still operate on the typed `T`.
- **Regenerated the sqlx offline cache** (one query file swapped; EE caches preserved).
- **Doc:** updated the stale `strip_null_chars` reference in `windmill-api-workspaces/src/workspaces.rs` to point at the shared `strip_json_nul`.

## Test plan

- [x] `cargo check -p windmill-queue -p windmill-api -p windmill-common -p windmill-api-workspaces` — clean, no warnings
- [x] `strip_json_nul` unit tests in `windmill-common` (clean-borrow, real-NUL, legit-escape borrow no-op, collision, nested keys/values, odd-run): 6 passed
- [x] End-to-end regression in `backend/tests/nativets_jobs.rs` (`--features deno_core`): a JS job returning a genuine NUL and a literal `\\u0000` completes, storing `"ab"` (stripped) and `"a\\u0000b"` (preserved). Without the fix the insert aborts and the job never completes.
- [x] `backend/tests/drafts_nul.rs` integration test still passes (helper refactor intact)

---
## Summary by cubic
Sanitizes U+0000 in completed job results before inserting into `jsonb`, and in `wm_labels` before `text[]` bind, preventing Postgres errors and unblocking job completion. Centralizes NUL stripping in `windmill-common` with a fast SIMD guard and a zero-copy clean path.

- **Bug Fixes**
  - Strip real `\u0000` from result JSON in `commit_completed_job`, binding `$3::text::jsonb` for both INSERT and ON CONFLICT; test confirms real NUL is removed and `"\\u0000"` is preserved.
  - Strip NULs from `wm_labels` entries before the `text[]` update to avoid aborting the completion transaction; extended regression test to cover label sanitization.

- **Refactors**
  - Move `strip_json_nul` to `windmill-common::utils` returning `Cow<str>`; add SIMD fast-path via `memchr::memmem` and depend on `memchr`.
  - Add `ValidableJson::serialized_json()` to borrow `Box<RawValue>` and single-alloc `serde_json::to_string`; regenerate `sqlx` cache and update query casts; small doc fix.

<sup>Written for commit 2dc53ad185d12271fdf1bde26b5036fc2b458909. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 3652b098bb1e17c3c17c624bf969929d14c7493d

PR #10278, opened 2026-07-23
fix(cli): surface shared UI changes in sync push dry-run preview

rubenfiszel:
## Summary

The git-sync "Pull from repo" preview never showed shared UI (`ui/` folder) changes, so users thought the shared-UI folder was not syncing. The apply step **does** sync it; only the preview/dry-run was blind.

The Shared UI store (`?tab=shared_ui`, backed by `workspace_shared_ui`) maps to a single top-level `ui/` folder and is handled out-of-band from the normal script/flow/app diff (`isNotWmillFile` excludes `ui/`). Round-trip is done by `pushSharedUi`/`pullSharedUi`. In `push()`, the dry-run path returns **before** `pushSharedUi` runs, so the `changes` array the modal consumes never contained a `ui/` entry and the preview read as "No changes to pull" even when the apply would overwrite the store.

## Changes

- **`cli/src/commands/shared_ui.ts`**: add exported `diffSharedUi(workspace)` returning `added`/`edited`/`deleted` `ui/<rel>` entries (push direction, local -> remote). Refactor `pushSharedUi` to reuse it so preview and apply never diverge. `diffSharedUi` mirrors `pushSharedUi`'s no-op: with no local `ui/` folder it returns `[]` (apply doesn't touch the remote in that case), so the preview never shows phantom deletes the apply won't perform.
- **`cli/src/commands/sync/sync.ts`**: fold the `diffSharedUi` result into the `changes` list in the dry-run path only (before both early returns), guarded by try/catch. They flow through the JSON contract and `prettyChanges` with no frontend change. The apply path is unchanged (`pushSharedUi` still runs on `dryRun=false`). Also: label `ui/…` paths as "shared UI" in `changeTypeLabel`/`prettyChanges` (`getTypeStrFromPath` throws on non-wmill paths like `ui/config.json`), and skip `pushSharedUi` in the zero-changes branch during a dry-run so a dry-run never mutates the remote.
- **`cli/test/shared_ui_diff_unit.test.ts`**: regression test for `diffSharedUi` (added/edited/deleted emitted when local differs; empty when matching; empty when no local `ui/`).

### Decision notes

- The `diffSharedUi` no-op guard is stricter than "local absent AND remote empty": it returns `[]` whenever the local `ui/` folder is absent (any remote), because `pushSharedUi` returns early and never deletes the remote store in that case. Matching apply exactly is what avoids a preview/apply mismatch.

## Test plan

- [x] CLI unit test: `UNIT_ONLY=1 bun test test/shared_ui_diff_unit.test.ts` (5 pass)
- [x] `bunx tsc --noEmit`: no new type errors (identical error count with/without the change; the 2 pre-existing `JSZip | TarAsZip` errors are unrelated)
- [x] End-to-end against a running backend:
  - Local `ui/theme.json` differing from an empty remote -> `sync push --dry-run --json-output` includes `{"type":"added","path":"ui/theme.json"}`, `total: 1`
  - After a real `sync push`, the store matches -> dry-run has no `ui/` entries, `total: 0`
  - Terminal (non-JSON) dry-run renders `+ shared UI ui/config.json` / `~ shared UI ui/theme.json` without the `getTypeStrFromPath` crash

Note: the full Gitea + EE git-sync e2e (driving the actual "Pull from repo" modal with Playwright) was not run here; the modal consumes the exact JSON `changes` array validated end-to-end above and the frontend is unchanged, so the fix is exercised by the CLI-level e2e.

---
## Summary by cubic
Fixes git-sync dry-run preview so changes in the `ui/` shared UI folder are shown and counted, matching what the apply step would push. Also improves output labeling and JSON messages around shared-UI-only changes.

- **Bug Fixes**
  - Added `diffSharedUi(workspace)` and reused it in `pushSharedUi` so preview and apply stay consistent; returns no diffs when local `ui/` is missing.
  - Dry-run now folds `ui/` changes into the `changes` list and summary; pretty output labels them as “shared UI” and uses own-property checks to handle names like `ui/toString`.
  - Dry-run never calls `pushSharedUi`, avoiding remote mutations; JSON output now reports shared-UI-only pushes instead of “No changes”.
  - Added unit tests covering add/edit/delete, matching stores, no local `ui/`, and inherited-name files.

<sup>Written for commit 0c2abdbaef84fdf0bc42b3fd3a202653d44dc3b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/src/commands/shared_ui.ts
+++ b/cli/src/commands/shared_ui.ts
@@ -26,6 +26,54 @@ async function readDirRecursive(
   return out;
 }
 
+export type SharedUiChange =
+  | { type: "added"; path: string }
+  | { type: "edited"; path: string; before: string; after: string }
+  | { type: "deleted"; path: string };
+
+/**
+ * Diff the local <cwd>/ui/ folder against the workspace's shared UI store in
+ * the push direction (local -> remote), returning entries whose `path` is
+ * prefixed with `ui/`. This is the same comparison pushSharedUi applies, so the
+ * dry-run preview and the real push never diverge.
+ *
+ * Mirrors pushSharedUi's no-op: with no local ui/ folder there is nothing to
+ * push, so the apply is a no-op and the preview must be empty (even when the
+ * remote store is non-empty) to avoid phantom diffs the apply won't perform.
+ */
+export async function diffSharedUi(workspace: string): Promise<SharedUiChange[]> {
+  const localDir = path.join(process.cwd(), SHARED_UI_DIR);
+  if (!fs.existsSync(localDir)) {
+    return [];
+  }
+  const files = await readDirRecursive(localDir);
+
+  let remote: Record<string, string> = {};
+  try {
+    const got = await wmill.getSharedUi({ workspace });
+    remote = got.files ?? {};
+  } catch {
+    // If endpoint missing or unauthorized, treat remote as empty (the push
+    // would attempt the PUT anyway).
+  }
+
+  const changes: SharedUiChange[] = [];
+  for (const [rel, content] of Object.entries(files)) {
+    const p = `${SHARED_UI_DIR}/${rel}`;
+    if (!(rel in remote)) {
+      changes.push({ type: "added", path: p });
+    } else if (remote[rel] !== content) {
+      changes.push({ type: "edited", path: p, before: remote[rel], after: content });
+    }
+  }
+  for (const rel of Object.keys(remote)) {
+    if (!(rel in files)) {
+      changes.push({ type: "deleted", path: `${SHARED_UI_DIR}/${rel}` });
+    }
+  }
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 0c2abdbaef84fdf0bc42b3fd3a202653d44dc3b0

PR #10282, opened 2026-07-23
feat(ai-chat): let the session/global chat create email triggers

rubenfiszel:
## Summary

The AI **session/global chat** could not create email triggers. Its `open_page` tool advertised `email` as a trigger kind (routing to `/email_triggers`), but `write_trigger` did not accept `kind: "email"` — the global chat's `TRIGGER_KINDS` stopped at 10 kinds. Seeing email as a navigable page but having no tool to create one, the model concluded email triggers were unsupported and faked them with an HTTP webhook plus an external mail provider.

The **flow chat** already gained email support on `main` (create_trigger + the SMTP availability guard). This PR brings the same capability to the global/session chat, whose trigger writes are draft-based. After merging `main`, the net change here is scoped to the global chat wiring only — the shared generation and flow-chat pieces already live on `main`.

## Changes

- **`global/workspaceItems.ts`**: add `email` to `TRIGGER_KINDS` and `NewEmailTrigger` to `TriggerRequestBody`.
- **`global/userDraftAdapter.ts`**: map `email → trigger_email` in the draft-kind map and `GLOBAL_DRAFT_KINDS` (`trigger_email` is already a valid `UserDraftItemKind` with deploy handlers).
- **`global/core.ts`**: add `email` to `triggerServices`, `triggerLabels`, and the `write_trigger` config union, plus the `EmailTriggerService` import. `write_trigger` creates a per-user draft, so no live-instance availability check is needed here (unlike flow chat's create_trigger, which creates a deployed trigger).
- **Tests**: a mocked unit test covering the full email trigger draft → read → deploy path (`global/core.test.ts`); a `global-test29` discovery guard.

## Discovery guard & why ai_eval missed this

Global-mode trigger coverage tested only `http` (`global-test4`). A missing *kind* is invisible to the benchmark: the model can't be graded on an option that isn't in the enum. `global-test29-email-trigger-draft` asserts the model selects `kind: email` for an implicit "run when an email is received" request.

It asserts the recorded **tool-call kind**, not the resulting draft, on purpose: email trigger routes are `#[cfg(all(feature = "smtp", feature = "private"))]`, so on a CE backend the probe/create 404s and no draft persists — but the tool-call args are captured regardless, making the guard backend-independent. The full persist/deploy path is covered by the unit test instead.

## Test plan

- [x] Frontend unit tests green (`global/core.test.ts`, 170 passing incl. new email test)
- [x] `frontend` `npm run check:fast` clean (only a pre-existing unrelated `modern-screenshot` error)
- [x] Global discovery guard: **4/4** (sonnet, vs CE backend) — model picks `kind: email`, correct `local_part`
- [ ] Manual (requires an `smtp`+`private` build, which the CE dev backend is not): in global chat, create a script + email-trigger draft, deploy the trigger, and confirm the `EmailTriggerEditorInner` drawer opens with the drafted `local_part`

## Screenshots

No new UI surface — this is a behavioral change to the global chat's `write_trigger` tool schema. The email trigger card and `EmailTriggerEditorInner` drawer are pre-existing components; creating one exercises them unchanged. A live browser capture requires an `smtp`+`private` backend (email routes 404 on the CE dev backend), so it is listed as manual verification above rather than screenshotted here.


rubenfiszel:
✅ Review round clean @ bc4085944e914bc5ee6122cd8ab7a74bc6e5d435

PR #10286, opened 2026-07-23
fix(monitor): diagnose zombie-flow OOM on the transition worker, not q.worker

rubenfiszel:
## Summary

Fixes WIN-2231.

A production flow was cancelled as a between-steps zombie with a diagnostic that told on-call to *"capture a stack trace (e.g. via SIGQUIT) — worker still pinging with healthy memory, likely deadlocked"*. That advice was wrong and cost investigation time.

**Ground truth:** the subflow was on its final forloop iteration (187/188); all 188 iterations had completed. A worker (`hql6x`) finished the last iteration, logged *"subflow is finished, updating parent flow status"* (this write set the flow's `last_ping`), then was **OOMKilled ~3s later** (k8s `lastState.terminated.reason=OOMKilled`, 1024Mi limit) — killed mid state-transition. The dropped transition left the flow's module `InProgress` with a frozen ping, so 60s later the monitor reaped it. The worker the alert *named* (`6pwsf`) was a **different, healthy** worker — it just happened to be the one recorded on the flow's outer queue row.

## Root cause

`handle_zombie_flows()` built its diagnostic by joining `worker_ping` on the flow's queue-row worker only:

```sql
LEFT JOIN worker_ping wp ON wp.worker = q.worker
```

`q.worker` is the worker on the **outer flow queue row**, which in nested / subflow / forloop cases is frequently NOT the worker that performed and dropped the final transition. So all OOM detection (`oom_strong`, `oom_moderate`, `worker_ping_stale`, `memory_pct`) ran against the wrong (healthy) worker, and the `(false,false,false)` branch confidently printed "healthy memory → deadlock → SIGQUIT".

## Changes (diagnostics only — no change to *when* a flow is cancelled)

**Change 1 — inspect the worker that actually died.** When `q.worker` looks healthy, additionally query `worker_ping` (new helper `find_zombie_flow_culprit_worker`) for a plausible culprit: a *different* worker on the same pod (`worker_instance`) or worker group whose last ping clusters around the flow's `last_ping` (within `[-5s, +15s]`) and has since gone silent — the signature of a worker OOM-killed mid state-transition. The candidate closest in time to the transition wins. If found, the message names that worker and routes to the OOM explanation, alongside (not replacing) the existing `q.worker` info. The service-log lookup also follows the culprit's instance/timestamp when one is found.

**Change 2 — stop over-weighting "deadlock".** The healthy-`q.worker` hint now leads with *"check whether a different worker on the same pod / worker group was OOM-killed around {last_ping}"*, keeping deadlock/SIGQUIT as the secondary possibility.

The extra query runs only on the rare zombie path, only when `q.worker` is not already implicating itself, and fails soft (`warn` + fall back to prior behavior) so a diagnostics failure never blocks the cancel. No migration.

## Before / after (the incident)

**Before:**
> Worker handling the flow: `wk-small-production-6pwsf` (group: small)
> Worker last ping: … (3s ago) — worker still pinging with healthy memory — likely deadlocked or blocking on the state transition
> …
> Worker is still pinging and memory looked healthy at its last ping — **most likely a deadlock or blocking call during the state transition. Capture a stack trace (e.g. via SIGQUIT)** …

**After:**
> Worker handling the flow: `wk-small-production-6pwsf` (group: small)
> Worker last ping: … (3s ago) — still pinging with healthy memory — **this is NOT the worker that performed the flow's final state transition (see likely culprit worker below)**
> **Likely culprit worker (on same pod/instance '…'): `wk-small-production-hql6x`** — last pinged … (63s ago, 2s after this flow's last ping) and then stopped pinging; it was at 96.5% of its memory limit at that last ping (memory: 988.0 MB of 1024.0 MB). This flow's dropped state transition was most likely performed by this worker and lost to its death (most likely OOM-kill), not a deadlock on `wk-small-production-6pwsf`.
> …
> **This is almost certainly an OOM-kill on a *different* worker:** `wk-small-production-hql6x` (on the same pod/worker group) was at 96.5% of its memory limit at its last ping right around this flow's transition, then stopped pinging. … Raise the worker memory limit …

## Exhaustive failure grid (is the shown message the most relevant?)

`q.worker` = worker on the outer queue row. The culprit search runs **only** when `q.worker` is not already implicating itself (not OOM-moderate and not stale) and a worker name + `last_ping` exist.

| # | `q.worker` observed state | culprit found? | Message leads with | Most relevant? |
|---|---|---|---|---|
| A | mem ≥ 0.85 (strong OOM) | not searched | `q.worker` OOM-killed at X% | ✅ q.worker is the victim |
| B | 0.60 ≤ mem < 0.85 (moderate) | not searched | `q.worker` likely OOM at X% | ✅ |
| C | ping > 60s stale, no mem evidence | not searched | `q.worker` died — most likely OOM | ✅ leads OOM; a >60s ping gap is itself a death signal so q.worker is treated as the victim |
| D | healthy (pinging, mem < 0.60) | **culprit, mem ≥ 0.60** | **culprit named, "almost certainly OOM-kill on a different worker"** | ✅ the incident — fixed |
| E | healthy | **culprit, stale-only** | **culprit named, "most likely OOM-kill on a different worker; only if not OOM, consider deadlock"** | ✅ leads OOM, deadlock hedged |
| F | healthy | none found | **"check whether a different worker on same pod/group was OOM-killed around {last_ping}; only if none, deadlock/SIGQUIT"** | ✅ improved — old code led with deadlock here (the reported bug) |
| G | no `worker_ping` row for q.worker | **culprit found** | **culprit named + OOM lead** | ✅ improved — old hint was empty |
| H | no `worker_ping` row | none found | (empty hint — no signal anywhere) | ⚪ unchanged — nothing to inspect |
| I | no worker recorded on queue row | not searched | (empty hint) | ⚪ unchanged — no anchor to search from (would need the schema migration this PR deliberately avoids) |

Cross-cutting culprit-query failure modes: **SQL error/DB blip** → `warn` + `None`, falls back to F/H/I, cancel still proceeds. **Multiple candidates** → closest-in-time wins (most likely mid-transition). **False positive** (a same-*group* worker on another pod that scaled down inside the ±window) → hedged text (`worker group 'X'` vs `same pod/instance 'X'`) and the "check pod restart count" first action is harmless and disambiguating. Only cells H and I are non-specific, and both are genuinely no-signal cases unfixable without recording the transition worker in schema.

## Test plan
- [x] `cargo build` / `cargo watch` compiles clean; `cargo clippy -p windmill` clean for `monitor.rs` (a pre-existing unrelated clippy error in `windmill-queue/src/jobs.rs:6193` is untouched by this PR)
- [x] `.sqlx` offline cache regenerated (1 new query file; 0 EE caches lost)
- [x] Unit tests for `zombie_worker_memory_pct` (max-of-two readings, single-reading fallback, `None` cases) pass
- [ ] Manual: reproduce a nested/subflow zombie where the final iteration runs on a second same-pod worker that is then OOM-killed near the flow's `last_ping`; confirm the cancellation `reason` surfaces the "Likely culprit worker" line and the service-log lookup targets the culprit's instance/time window
- [ ] Confirm behavior/wording unchanged when there is genuinely no OOM signal anywhere (rows H/I)

---
## Summary by cubic
Diagnose zombie-flow OOM on the worker that performed the final transition, not the outer queue-row worker. Fixes WIN-2231 and avoids deadlock/SIGQUIT false-positives.

- **Bug Fixes**
  - When `q.worker` looks healthy, search `worker_ping` for a different worker on the same `worker_instance` or `worker_group` whose last ping clusters near the flow’s `last_ping` and then went silent; pick the closest-in-time candidate.
  - Show a “Likely culprit worker” line naming the worker, its ping delta, and memory; lead with “OOM on a different worker” and keep deadlock as secondary.
  - Route service-log lookups to the culprit worker’s instance/time window instead of `q.worker`.
  - Compute memory percent using the max of cgroup and process readings for stronger OOM signal.
  - Diagnostics-only on the zombie path and soft-failing; no changes to cancellation timing or schema.

<sup>Written for commit e3f11caebc069bca3105d02b6913ddd99e15314f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-856697038686ec82f1925f70c57d41e538012b79637c9c0dcd29345ded09f699.json
+++ b/backend/.sqlx/query-856697038686ec82f1925f70c57d41e538012b79637c9c0dcd29345ded09f699.json
@@ -1,0 +1,65 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        WITH ref AS (\n            SELECT worker_instance, worker_group FROM worker_ping WHERE worker = $1 LIMIT 1\n        )\n        SELECT\n            wp.worker AS \"worker!\",\n            wp.ping_at AS \"ping_at!\",\n            wp.memory_usage,\n            wp.wm_memory_usage,\n            wp.memory AS memory_total,\n            wp.worker_group,\n            wp.worker_instance,\n            EXTRACT(EPOCH FROM (wp.ping_at - $2::timestamptz))::float8 AS ping_delta_secs\n        FROM worker_ping wp, ref\n        WHERE wp.worker <> $1\n            AND (\n                (ref.worker_instance IS NOT NULL AND wp.worker_instance = ref.worker_instance)\n                OR (ref.worker_group IS NOT NULL AND wp.worker_group = ref.worker_group)\n            )\n            AND wp.ping_at >= $2::timestamptz - interval '5 seconds'\n            AND wp.ping_at <= $2::timestamptz + interval '15 seconds'\n        ORDER BY ABS(EXTRACT(EPOCH FROM (wp.ping_at - $2::timestamptz))) ASC\n        LIMIT 1\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "worker!",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "ping_at!",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 2,
+        "name": "memory_usage",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 3,
+        "name": "wm_memory_usage",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 4,
+        "name": "memory_total",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 5,
+        "name": "worker_group",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 6,
+        "name": "worker_instance",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 7,
+        "name": "ping_delta_secs",
+        "type_info": "Float8"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Timestamptz"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      true,
+      true,
+      true,
+      false,
+      false,
+      null
+    ]
+  },
+  "hash": "856697038686ec82f1925f70c57d41e538012b79637c9c0dcd29345ded09f699"
+}
--- a/backend/src/monitor.rs
+++ b/backend/src/monitor.rs
@@ -4864,6 +4864,87 @@ WHERE concurrency_id IN (SELECT concurrency_id FROM rows_to_delete)  RETURNING c
     Ok(())
 }
 
+/// Memory usage at a worker's last ping as a fraction of its cgroup limit.
+/// Takes the larger of the cgroup-wide reading and the windmill process's
+/// jemalloc resident — if only one is present, that value wins; if both are
+/// present, the larger is the more conservative (higher-signal) choice.
+fn zombie_worker_memory_pct(
+    usage: Option<i64>,
+    wm_usage: Option<i64>,
+    total: Option<i64>,
+) -> Option<f64> {
+    let total = total?;
+    if total <= 0 {
+        return None;
+    }
+    let used = usage.max(wm_usage)?;
+    Some(used as f64 / total as f64)
+}
+
+struct ZombieFlowCulprit {
+    worker: String,
+    ping_at: DateTime<Utc>,
+    memory_usage: Option<i64>,
+    wm_memory_usage: Option<i64>,
+    memory_total: Option<i64>,
+    worker_group: Option<String>,
+    worker_instance: Option<String>,
+    ping_delta_secs: Option<f64>,
+}
+
+/// Finds the worker that likely performed and dropped the flow's final state
+/// transition when `q.worker` (the outer queue-row worker) looks healthy — a
+/// different worker on the same pod/group whose *latest* ping is frozen in the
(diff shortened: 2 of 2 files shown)


rubenfiszel:
Thanks — both Codex findings addressed in e3f11caebc069bca3105d02b6913ddd99e15314f:

**P1 (culprit lookup rejected post-transition pings before cancellation).** Removed the `wp.ping_at < NOW() - timeout` predicate and the `stale_secs` parameter. It was both redundant and harmful: the `[last_ping-5s, +15s]` window already proves the candidate has gone silent, because `worker_ping` is updated in place — a still-alive worker advances its latest `ping_at` past that old window, so a *frozen* ping inside the window can only belong to a worker that stopped pinging after the transition. The extra hard-60s cutoff added nothing and, as you noted, rejected a culprit that pinged 2s after `last_ping` when the monitor ran at `last_ping + 61s`. Now the window alone selects it.

**P2 (doc comment length).** Trimmed `find_zombie_flow_culprit_worker`'s doc to the durable constraint (why a frozen in-window ping proves silence) in ≤6 lines.

**Test coverage note.** `find_zombie_flow_culprit_worker` is DB-integration logic on the reaper path that returns a diagnostics string and fails soft; a full integration test needs `worker_ping` + `v2_job_queue` + timing fixtures for no functional (cancellation) guarantee, so per REVIEW.md's "ship only the tests the PR needs" I've kept the focused unit tests on the pure `zombie_worker_memory_pct` classifier and left the SQL to live validation via the committed `.sqlx` cache. Manual verification steps remain in the PR body.

rubenfiszel:
✅ Review round clean @ e3f11caebc069bca3105d02b6913ddd99e15314f

PR #10287, opened 2026-07-23
feat(monitor): make between-steps zombie flows hand-recoverable

rubenfiszel:
## Summary

Between-steps zombie flows (a worker OOM-killed mid state-transition: children all `success`, the module still `InProgress`, ping frozen) are reaped and cancelled by the monitor. We deliberately do **not** auto-recover them (a re-driven transition can OOM again on the same aggregated state, a poison pill that walks through workers). Instead this makes the **manual** recovery path excellent. Builds on #10286 (culprit-worker diagnostic); no change to *when* a between-steps flow is cancelled.

## Demo

Between-steps zombie flow reaped and cancelled (with the actionable RECOVERY guidance), then hand-restarted from the stuck step — the loop iterations are reused (0.0s each, not re-executed) and the flow reaches success:

zombie-recovery

## Changes

- **Actionable diagnostic** (`monitor.rs`): when the reaped step is a *provably complete* derivable zombie, the cancellation reason / critical alert appends concrete restart guidance: which step, how many children/iterations completed, raise the memory limit first, then restart-from-step, with the run-page deep link and the exact restart API call.
- **Hand-restart reuses completed children** (`windmill-queue/jobs.rs`): in `restarted_flows_resolution`, when the restart target's original module is a between-steps-zombie step that is safe to reuse (guards below), it is reused verbatim as `Success` and the flow restarts from the **next** step. No completed child re-runs; downstream steps re-derive the reused step's result from `flow_jobs`/`job` on demand (`get_previous_job_result` / `get_transform_context`), so nothing is re-aggregated server-side. Covers single-child (leaf / subflow / branchone) and multi-child (forloop / branchall) steps.
- **Parse fix** (`monitor.rs`): the reaper read `flow_status` (jsonb) via a `Box<str>` type-override, which includes the jsonb binary version byte and made `serde_json::from_str::<FlowStatus>` fail at column 1 for every flow (latent since the v2 table migration, present on `main`). This had silently disabled the "restart not-yet-started zombie" branch and blocked the new derivable-state detection. Fixed by casting `::text`. **Behavior note for review:** re-activates that previously-dead branch (re-queue instead of cancel); intended, easy to gate if you'd rather keep it dormant.

## Safety guards on reuse

Shared `FlowStatusModule::is_between_steps_complete` (status-side) + `FlowModule::allows_zombie_reuse` (definition-side); a step is reused only when **all** hold:

- **Cursor reached the end.** A serial for-loop / branch-all grows `flow_jobs` one entry at a time, so an all-success prefix does not mean "finished": reuse requires `iterator.index + 1 == itered_len` / `branchall.branch + 1 == len`. Parallel containers preallocate every child, so children-success is conclusive. **While-loops are never derivable** (continuation is a post-iteration condition a reaped zombie never persisted).
- **Every child is a `success` completion** in `v2_job_completed`.
- **No completion/arming side-semantics that reuse would skip**: `stop_after_if` / `stop_after_all_iters_if` (decide whether downstream runs), `skip_if` (skipped-state + suspend arming), `suspend` (approval boundary), or `sleep`. Such a step re-runs instead (re-evaluating those normally).
- **A next step exists.** Advancing past the last module lands on the failure step, so a last-module zombie re-runs (documented limitation: a zero-re-run finalize of a *final* fan-out would need the flow engine to re-drive aggregation without a child-completion event; out of scope).

The reaper does not load the flow definition, so the guidance text states **both** outcomes (reuse where derivable, re-run for the last step or one carrying a stop/skip/approval/sleep) rather than promising "no re-run" the restart path might decline. Passing an explicit `branch_or_iteration_n` always forces the existing partial-restart behavior.

## Test plan

- [x] Unit tests for `is_between_steps_complete` (serial loop complete vs mid-iteration, while-loop, serial/parallel branch-all, leaf, non-InProgress).
- [x] Integration test `zombie_flow_recovery.rs`: (1) happy path — a fully-completed fan-out loop reaped with the transition lost is restarted from the loop step, every iteration reuses its original child UUID, the next step runs fresh, flow reaches `success` with the correct aggregated result (confirmed RED without the fix); (2) a serial loop reaped **mid-iteration** with a downstream step is NOT reused — it re-runs the whole loop and still succeeds (confirmed RED with the cursor guard neutered: downstream consumes a truncated `"a"` instead of `"a,b,c"`).
- [x] End-to-end on a running dev backend: seeded a real reaped zombie and confirmed the cancellation reason contains the RECOVERY guidance.
- [x] Existing restart regression tests pass in isolation; no behavior change for normally-terminal flows. `cargo clippy` clean for new code; `.sqlx` regenerated.

Closes WIN-2232.

### DIFF
--- a/backend/.sqlx/query-1bf87fe9667f860c6089bffb803291f4ff5c979be956369fa4406d789cca92fc.json
+++ b/backend/.sqlx/query-1bf87fe9667f860c6089bffb803291f4ff5c979be956369fa4406d789cca92fc.json
@@ -1,0 +1,112 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT\n            j.id AS \"id!\", j.workspace_id AS \"workspace_id!\", j.parent_job, j.flow_step_id IS NOT NULL AS \"is_flow_step?\",\n            COALESCE(s.flow_status, s.workflow_as_code_status)::text AS \"flow_status: Box<str>\", r.ping AS last_ping, j.same_worker AS \"same_worker?\",\n            q.worker AS \"worker?\",\n            wp.ping_at AS \"worker_last_ping?\",\n            wp.memory_usage AS \"worker_memory_usage?\",\n            wp.wm_memory_usage AS \"worker_wm_memory_usage?\",\n            wp.memory AS \"worker_memory_total?\",\n            wp.worker_group AS \"worker_group?\",\n            wp.wm_version AS \"worker_version?\",\n            wp.current_job_id AS \"worker_current_job_id?\",\n            wp.worker_instance AS \"worker_instance?\"\n        FROM v2_job_queue q JOIN v2_job j USING (id) LEFT JOIN v2_job_runtime r USING (id) LEFT JOIN v2_job_status s USING (id)\n            LEFT JOIN worker_ping wp ON wp.worker = q.worker\n        WHERE q.running = true AND q.suspend = 0 AND q.suspend_until IS null AND q.scheduled_for <= now()\n            AND (j.kind = 'flow' OR j.kind = 'flowpreview' OR j.kind = 'flownode' OR j.kind = 'singlestepflow')\n            AND r.ping IS NOT NULL AND r.ping < NOW() - ($1 || ' seconds')::interval\n            AND q.canceled_by IS NULL\n\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id!",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "workspace_id!",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 2,
+        "name": "parent_job",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 3,
+        "name": "is_flow_step?",
+        "type_info": "Bool"
+      },
+      {
+        "ordinal": 4,
+        "name": "flow_status: Box<str>",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 5,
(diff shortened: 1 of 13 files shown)


rubenfiszel:
✅ Review round clean @ 1c2a1f59df1ac399312f9ff5a618c775c3e30e63

rubenfiszel:
## Recovery UX on the run page + regression check

**Regression:** all existing restart paths still work — the 8 restart tests pass unchanged (`test_restarted_flow_payload`, `test_raw_flow_payload_with_restarted_from`, and all `test_nested_restart_inside_*`: branchone, subflow, forloop iteration, nested forloops, deployed flownode). The reuse branch only fires for a zombie-`InProgress` step (impossible on a normally-terminal flow) and declines when a nested chain is present, so nothing that was restartable before changed.

**UX:** the recovery is entirely on the existing run page — no new UI. A reaped flow shows as `Canceled`, the `RECOVERY:` guidance renders in the result/cancellation reason (step id, iterations completed, raise-memory-then-restart, and the exact "Re-start from <step>" button + API call), and selecting the stuck step reveals the existing **Re-start from `<step>`** button. (Restart is an EE feature, so the button is disabled with an EE tooltip on this CE build.)

run page with RECOVERY guidance + Re-start from fanout button

Follow-up in the same round: the guidance is now gated to **restartable (deployed, top-level) flows** — a preview run has no flow path (the button is hidden, the API 400s) and a subflow child restarts via its root, so those keep the plain wording instead of being told to use a button that isn't there. Verified end-to-end (reaped preview → no RECOVERY block; reaped deployed flow → RECOVERY block).

rubenfiszel:
✅ Review round clean @ 0ba5663974d0d2171d82a6e94d690d444a266d15

PR #10288, opened 2026-07-23
fix: treat concurrent_limit/timeout <= 0 as unset instead of a zero cap

rubenfiszel:
## Summary

A script or flow (or flow step) whose `concurrent_limit` or `timeout` is stored as integer **`0`** (not `NULL`) was silently bricked. `Some(0)` looks meaningful but is nonsensical, and the backend enforced it literally:

- **`concurrent_limit = 0`** → "zero concurrent slots": every job for that runnable can never pass the concurrency gate. It spins in the pull/concurrency loop (a re-queue storm — `"Window finding for job <id> loop count: 1"` dozens of times in a few ms), never executes, never refreshes `v2_job_runtime.ping`, and after 60s the zombie-job monitor restarts it 3× then fails it with the **misleading** error *"Job timed out after no ping… OOM are a common reason."* No OOM ever happened.
- **`timeout = 0`** → a literal 0-second job timeout: `sleep(0)` fires immediately and the job is killed (~140ms) with *"timeout after exceeding job-specific duration limit."*

The trap is a UI/backend disagreement: the frontend treats `0` as "disabled" and shows nothing enabled, but the backend queue code treated `Some(0)` as **active** because it only ever checked `.is_some()`. A deploy that pushed scripts with `concurrent_limit: 0` and `timeout: 0` via the CLI failed 100% of the time.

**Rule implemented:** for these positive-only `Option<i32>` runnable settings, **`n <= 0` is never meaningful — treat it as unset (`None`) everywhere.** Coerce-to-null (non-breaking, matches the UI) rather than reject. `Some(n > 0)` semantics are unchanged.

## Changes

Two shared primitives in `windmill-types/src/runnable_settings.rs`:
- `none_if_non_positive(Option<i32>) -> Option<i32>` — the single `<= 0 ⇒ None` rule.
- `has_active_concurrency_limit(Option<i32>) -> bool` (in `windmill-queue/src/jobs.rs`) — replaces the `.is_some()` concurrency gates.

Defense-in-depth at three layers:

**1. Ingestion (shouldn't have been accepted)**
- Custom `Deserialize` on `ConcurrencySettings` and `ConcurrencySettingsWithCustom` normalizes `<= 0` on parse and drops the paired `concurrency_time_window_s`. Because both are `#[serde(flatten)]` into `NewScript`, `FlowModule`, and `FlowModuleValue`, this covers scripts, flows (flow-level + per-step), apps, and trigger payloads uniformly — and also normalizes read-back of existing stored settings.
- Script create/update and flow create/update handlers coerce `timeout` (not part of the concurrency struct) via `none_if_non_positive`, and re-normalize concurrency for in-process `NewScript` construction.

**2. Runtime (guard existing bad rows, no data migration required)**
- `ConcurrencySettings::maybe_fallback` normalizes — the single load boundary for every DB-backed read (script/schedule read, flow value, worker pull).
- `push_inner` normalizes before persisting the job row / concurrency key; the intentional dependency/DeploymentCallback `Some(1)` cases are unaffected (they are set positive after normalization).
- Pull, completed-job slot release, cloud queue-cap insert, and `worker_flow`'s raw-JSON concurrency read all gate on `has_active_concurrency_limit`.
- `resolve_job_timeout` treats `Some(n <= 0)` as `None` (falls through to the default/global-max timeout instead of a 0s duration).

**3. CLI (`cli/src/commands/script/script.ts`)**
- `nonePositiveInt` / `normalizeConcurrency` emit a disabled `concurrent_limit`/`timeout` as omitted, never `0`, on `wmill sync push`, and compare normalized values so a redeploy against a backend-normalized script reports no diff (no churn).

`cache_ttl = 0` was audited and left as-is: it means "expires immediately" (a benign no-op cache), not a footgun that breaks jobs.

## Existing rows

The runtime guards make already-stored `0` rows behave as disabled with **no migration required**. An optional one-off cleanup for cosmetic clarity only:

```sql
UPDATE script SET concurrent_limit = NULL, concurrency_time_window_s = NULL WHERE concurrent_limit <= 0;
UPDATE script SET timeout = NULL WHERE timeout <= 0;
UPDATE flow   SET timeout = NULL WHERE timeout <= 0;
```

## Test plan

- [x] `cargo test -p windmill-types --lib runnable_settings` — 7 pass (helper, both `Deserialize` impls incl. the `#[serde(flatten)]` path, `maybe_fallback` legacy-column boundary, disabled-serializes-as-omitted)
- [x] `cargo test -p windmill-queue --test concurrency_limit_zero_test` — 2 pass (runtime gate semantics)
- [x] `cargo build` / dev backend compiles cleanly (CE); `cargo check --features enterprise,private` clean for windmill-queue/api-scripts/api-flows
- [ ] Deploy a script/flow whose stored `concurrent_limit` is `0` → job runs to completion instead of stalling and being failed as a spurious OOM
- [ ] Deploy a runnable with `timeout: 0` → runs under the default/global-max timeout rather than being killed instantly
- [ ] `wmill script push` twice against a `0`/omitted-limit script → second push reports no diff

---
## Summary by cubic
Treat `concurrent_limit` and `timeout` values <= 0 as unset across API, runtime, and CLI to prevent zero-slot caps and 0s timeouts that stall or kill jobs. Flow steps now inherit the script timeout when their override is <= 0; existing rows work without migration and behavior matches the UI.

- **Bug Fixes**
  - Normalize `<= 0` during deserialization and runtime fallback; drop `concurrency_time_window_s` when the limit is disabled.
  - Use `has_active_concurrency_limit` (> 0) instead of `.is_some()` across `windmill-queue` to avoid registering zero-slot concurrency keys.
  - Timeout handling: treat `timeout <= 0` as unset; flow-step overrides `<= 0` now inherit the referenced script’s timeout (not the global default). Raw JSON paths also coerce before use.
  - Script/flow create/update coerce values server-side; CLI (`script.ts`) omits disabled values and compares normalized settings to avoid diff churn.
  - Added tests for normalization, gating, and flow-step timeout inheritance; `cache_ttl = 0` remains unchanged.

<sup>Written for commit 0da916c05e6835ee19b1d54cc6a14bcc0ace999c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-flows/src/flows.rs
+++ b/backend/windmill-api-flows/src/flows.rs
@@ -527,6 +527,11 @@ async fn create_flow(
     }
     check_scopes(&authed, || format!("flows:write:{}", nf.path))?;
 
+    // A `<= 0` flow timeout is "unset", not a 0-second limit that kills every run instantly.
+    // (The concurrency settings inside the flow value are normalized on deserialization; see
+    // ConcurrencySettings.) Runtime guards also protect already-stored rows.
+    nf.timeout = windmill_common::runnable_settings::none_if_non_positive(nf.timeout);
+
     if let RuleCheckResult::Blocked(msg) = check_deploy_rules(
         &w_id,
         AuditAuthorable::username(&authed),
@@ -1012,7 +1017,9 @@ async fn update_flow(
     }
     let flow_path = flow_path.to_path();
     // The URL identifies the flow being updated; the body path is only needed to rename.
-    let nf = ef.into_new_flow(flow_path);
+    let mut nf = ef.into_new_flow(flow_path);
+    // A `<= 0` flow timeout is "unset", not a 0-second limit (see create_flow).
+    nf.timeout = windmill_common::runnable_settings::none_if_non_positive(nf.timeout);
     check_scopes(&authed, || format!("flows:write:{}", flow_path))?;
 
     if let RuleCheckResult::Blocked(msg) = check_deploy_rules(
--- a/backend/windmill-api-scripts/src/scripts.rs
+++ b/backend/windmill-api-scripts/src/scripts.rs
@@ -914,6 +914,13 @@ async fn create_script_internal<'c>(
     }
     check_scopes(&authed, || format!("scripts:write:{}", ns.path))?;
 
+    // Normalize positive-only settings so a `<= 0` value (e.g. a CLI-pushed `0`) persists as
+    // disabled rather than as a zero-slot concurrency cap or a 0-second timeout. Deserialization
+    // already normalizes the concurrency fields; re-applying here also covers `timeout` and any
+    // NewScript built in-process rather than from a request body.
+    ns.timeout = windmill_common::runnable_settings::none_if_non_positive(ns.timeout);
+    ns.concurrency_settings = ns.concurrency_settings.normalized();
+
     guard_script_from_debounce_data(&ns).await?;
 
     let codebase = ns.codebase.as_ref();
--- a/backend/windmill-queue/src/jobs.rs
+++ b/backend/windmill-queue/src/jobs.rs
@@ -1107,15 +1107,16 @@ async fn commit_completed_job<T: Serialize + Send + Sync + ValidableJson>(
     // Resolve the concurrency-limit settings on the pool *before* opening the
     // completion transaction: doing it inside the tx would hold a second
     // simultaneous connection from the small per-worker pool.
-    let has_concurrent_limit = completed_job.concurrent_limit.is_some()
-        || windmill_common::runnable_settings::prefetch_cached_from_handle(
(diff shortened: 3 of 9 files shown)


rubenfiszel:
✅ Review round clean @ 0da916c05e6835ee19b1d54cc6a14bcc0ace999c

PR #10292, opened 2026-07-23
feat: let a workspace fall back to the instance critical alert channels

rubenfiszel:
## Summary

A workspace with no error handler has no way to surface failed jobs, and the instance critical alert channels a superadmin already configured (Slack, Teams, email) were unreachable from a workspace: the Slack workspace error handler posts with the *workspace* bot token (`f/slack_bot/bot_token`), while critical alerts use the instance one (`f/slack_bot/global_bot_token` in `admins`).

This adds the instance channels as a **fifth error-handler destination**, alongside Slack / Teams / Email / Custom. The backend already treats it as mutually exclusive with a handler script (it reports only when no handler is configured), so modelling it as one more answer to "where do failures go?" matches the semantics.

The report is **send-only**: it calls `send_critical_alert` directly and skips the `alerts` table, so workspace job failures never flood the instance-wide feed superadmins triage.

Guard rails:
- **Not available on cloud.** The channels belong to the instance operator, who on app.windmill.dev is not the workspace owner. Rejected by the API, tab not offered.
- **Not available on forks / dev workspaces**, which run throwaway copies of a parent's runnables. Rejected on write, cleared when a workspace is attached as a dev workspace, and enforced again at dispatch (the query joins `workspace.parent_workspace_id IS NULL`) so a workspace converted after opting in stops reporting.
- **Cancellations are not reported** — a human action, and this destination has no per-workspace mute of its own.
- **Throttled per workspace** (60s) with a rollup count, so a flapping runnable cannot turn into unbounded Slack/SMTP traffic on a shared resource.
- **EE only**: `send_critical_alert` is behind the `enterprise` feature.

## Changes

- `workspace_settings.error_handler_fallback_to_instance_alerts` (migration, default `false`)
- `send_workspace_error_to_instance_channels` in `windmill-common/src/utils.rs`: cloud hard-stop, then `send_critical_alert`, no `alerts` row
- `send_error_to_workspace_handler`: reports to the instance channels when no handler is configured and the workspace opted in. The per-runnable `ws_error_handler_muted` lookup now gates both destinations, and an early return keeps it off the path of every failed job in workspaces that use neither
- **No new per-failure query**: the opt-in is read from the query already behind `WORKSPACE_ERROR_HANDLER_CACHE` (60s TTL)
- `edit_error_handler` takes `fallback_to_instance_alerts` as `Option<bool>` so clients that predate it (the CLI pushing `settings.yaml`) leave the stored value alone instead of resetting it; the flag is recorded in the audit log
- `createWorkspace` takes the flag so it can be set on the new-workspace screen
- `ErrorOrRecoveryHandler`: new "Instance alerts" destination behind a `showInstanceAlerts` prop, so schedules and triggers (which have no such setting) are unaffected
- Rename copy (`workspaces_extra.rs`) now names its target columns instead of relying on physical column order, and carries the previously dropped `public_app_execution_limit_per_minute`

## Screenshots

Workspace settings → Error / success handler, new destination selected:

settings

Same tab on a fork workspace, destination not offered:

fork

New workspace screen:

create-workspace

## Test plan

Verified against a local EE build (`enterprise,private,license,quickjs`) with `critical_error_channels` pointing at a Slack channel:

- [x] Failed job in an opted-in workspace with no error handler reaches `send_critical_alert` (the instance had no Slack connection, so it logs "no Slack configuration found" — proof the path fired)
- [x] Same job in a workspace that did not opt in produces no attempt
- [x] Three failures in a row produce one report (throttle), and the dispatch is logged
- [x] `mute_critical_alerts` (the UI-feed mute) does **not** suppress it
- [x] Converting the workspace into a fork stops reporting within the cache TTL even with the stored flag still `true`
- [x] No `alerts` row written by this path
- [x] Destination persists across reload; flag lands in the DB when set from the new-workspace screen; not offered on forks
- [x] Integration test: enabling persists, a payload omitting the field does not reset it, enabling on a fork returns 400, and disabling on a fork still succeeds (no lockout)

Not covered automatically: the EE send itself (needs configured channels), which is why the manual run above was done.


rubenfiszel:
✅ Review round clean @ a3b0ffa101fb9d9f155aa74ce818a2b3f625663f — nit-only verdicts at 1895651d89373d1cab6847203dff6065466eb0be (claude and pi: good to merge; codex: nit); nit addressed in a3b0ffa101fb9d9f155aa74ce818a2b3f625663f

PR #10295, opened 2026-07-24
feat: alert on expired online license key

rubenfiszel:
## Summary

An expired **online (renewable)** license key produced **no critical alert** — only a `tracing::error!` in worker stdout — even though it halts all job execution. This happens when the key is force-set from outside Windmill (an environment variable, CI, or a k8s manifest): the external value overrides the auto-renewed key on every restart, so the daily renewal never "fails" while the effective key silently ages out.

The two existing alert paths don't cover it:
- `renew_license_key` alerts only when a **renewal attempt fails** — but nothing fails here; the external key just re-imposes an old, expired value.
- `enforce_offline_caps` raises `offline_license:invalidated` (which does cover expiry) but **only for offline / URL-bound keys** — it early-returns for online keys.

This adds `alert_on_online_license_expired`, mirroring the offline invalidated alert for the online case. It reuses the existing `simple_alert_helper` primitive, so it inherits dedup (one alert per `license_key:expired` tag, not per loop tick) and auto-recovery (fires a recovery message once the key renews). The alert message names the likely cause (external override) and points to Instance Settings / the IaC operator.

## Changes

- **EE** (`windmill-common/src/ee.rs`): new `alert_on_online_license_expired(db)` — skips offline keys (handled by `enforce_offline_caps`), and always reaches `simple_alert_helper` (even on the skip/unparseable paths) so a previously-raised alert recovers rather than staying lit forever.
- **`backend/src/monitor.rs`**: wire the call into the periodic `verify_license_key_f`, `server_mode`-gated (mirrors `enforce_offline_caps_f`) so only servers report to the alerts table / critical channels.
- **`backend/windmill-common/src/ee_oss.rs`**: OSS stub so the `enterprise`-without-`private` build compiles.

Companion EE PR: _(added below)_

## Test plan

- [x] `cargo check -p windmill-common --features enterprise` (no `private`) — compiles (exercises the ee_oss stub)
- [x] `cargo check -p windmill-common --features enterprise,private` — compiles
- [x] `cargo check -p windmill --features enterprise` (no `private`) — compiles (the config the missing stub would have broken)
- [x] `cargo check -p windmill --features enterprise,private,license` — compiles
- [ ] EE build: force `LICENSE_KEY` to an expired online value → exactly one `license_key:expired` critical alert fires on the configured channels
- [ ] Setting a valid future-dated key clears the alert (recovery message)
- [ ] Switching an expired online key to an offline license does not leave a dangling active `license_key:expired` alert

---
## Summary by cubic
Add a critical alert for expired online (renewable) license keys so instances don’t silently stop jobs when a stale key is set via env/CI/K8s. The alert dedups across replicas, auto-recovers on renewal, and only servers emit it.

- **New Features**
  - Wire `alert_on_online_license_expired` into the periodic `verify_license_key_f`, gated by `server_mode`, to emit a single `license_key:expired` alert to the DB and critical channels.
  - Add an OSS stub in `backend/windmill-common/src/ee_oss.rs` so `enterprise` without `private` builds compile.
  - Update `backend/ee-repo-ref.txt` to the merged EE changes with cross-replica dedup via `acquire_lock` and no false recovery when a malformed key replaces an expired one.

<sup>Written for commit 215f36dc5fdecca4edcfac2edd471862c5e93c3c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10295?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed both Codex P1 findings (EE follow-up: windmill-labs/windmill-ee-private#683, now pinned by `ee-repo-ref.txt`):

1. **Cross-replica duplicate alerts** — wrapped the check in `acquire_lock("license_key_expired_alert_lock", 300)` so a single replica evaluates/sends per interval, closing the non-atomic check-then-insert race in `simple_alert_helper` (`healthchecks` has no uniqueness constraint). Also serves as a fleet-wide throttle.
2. **False recovery for malformed replacement** — recovery now fires only when the license is genuinely valid again (`LICENSE_KEY_VALID`), not merely "not an expired online key". A malformed/empty replacement (which `clear_license_key` leaves with `LICENSE_KEY_VALID = false`) now leaves the alert lit instead of sending "License key is valid again" while jobs stay blocked.

rubenfiszel:
✅ Review round clean @ 215f36dc5fdecca4edcfac2edd471862c5e93c3c — nit-only verdicts at ad5badb61ac248176ada9aab3af393dc1a0e8c83 (Codex/Pi nit: missing `[ee]` title prefix, restored via title edit; Claude: good to merge). Both round-1 Codex P1s were fixed in EE #683 (merged, eb8733a) and verified byte-identical to the reviewed EE code; head moved past the reviewed SHA only via content-identical `ee-repo-ref` bumps.

PR #10297, opened 2026-07-24
feat: multiple homepage sort orders via an efficient merged runnables endpoint

rubenfiszel:
## Summary

Add multiple sort orders (recently updated / oldest / name A-Z / name Z-A) to the homepage, keeping "recently updated" as the default. Fixes WIN-2236.

The orders are produced **server-side** by a new merged, index-backed, keyset-paginated endpoint, so a chosen order is globally correct across scripts + flows + apps and stays efficient **even on very large workspaces** — and it pages past the old per-kind load caps (10k scripts / 1k flows / 1k apps) that a client-only sort was bounded by.

## Why an endpoint (not just a client sort)

The homepage merges three tables (script/flow/app) into one list. A *correct* combined order can only be produced after merging, and the per-kind endpoints cap what they return, so a pure client sort can only reorder a capped window. Doing the merge + order + pagination in one place fixes that.

## Backend — `GET /w/{workspace}/runnables/list`

- **Merged**: `UNION ALL` of script/flow/app with a shared projection carrying the fields the row components need.
- **Index-backed order**: each branch is ordered by an index so Postgres does a *Merge Append* and stops at the page limit. The migration adds time indexes (`script(workspace_id, archived, created_at)`, `flow(workspace_id, archived, edited_at)`) and expression indexes on `lower(COALESCE(NULLIF(summary,''), path))` for script/flow/app so the name orders are index-presorted too (verified the planner uses them).
- **Keyset pagination**: an opaque `(sort_key, path, kind, tiebreak)` cursor, so deep pages don't re-scan.
- **Bounded projections**: a per-branch `LIMIT` keeps the correlated subqueries (`draft_users`, `folder_labels`) evaluated only for the page, not the whole table — verified with `EXPLAIN ANALYZE` (SubPlan `loops` = page size, not row count).
- **Starred-first** on the first page; **RLS** enforces visibility in-SQL via the `user_db` transaction. Filters: `kinds`, `show_archived`, `include_without_main`, `path_start` (owner), `label`, `search`.

## Frontend — hybrid integration

- The homepage loads from `listRunnables` as the ordered, keyset-paginated source; the sort selector maps to `order_by`/`order_desc` and reloads; **"load more"** fetches the next page via the cursor. The sort trigger is an icon-only button with the four orders in the popover.
- **Kind and owner/folder filters are resolved server-side** (`kinds`, `path_start`), so selecting *Apps* or a folder browses the whole workspace for it, not just the loaded window.
- **Search is hybrid**: instant client-side fuzzy filtering over loaded pages for reactivity, plus a **debounced (300 ms) server augmentation** that pulls in matches beyond the loaded window — so search stays complete on large workspaces without a request per keystroke.
- **Complete facets**: the owner facet lists **every** workspace folder (paged to exhaustion) and **every** username, so an owner whose items sit far down the stream is still selectable.

### Lazy folder tree

The tree view lists **every workspace folder as a top-level node** regardless of the loaded window; expanding one **lazy-loads its items on demand** (server-scoped, with subfolders nesting) and paginates **within that folder** ("Load more in this folder").

Folder items live in a **separate store**, never merged into the global browse arrays — so a folder expansion can't pollute or duplicate the flat list, and each folder request is tied to a scope generation so a stale-scope response is discarded. Changing the order/scope remounts the tree so expanded folders re-load fresh. "Expand all" does not fan out a request per folder (a folder loads only on explicit click), keeping it safe at 10k+ folders.

## Known follow-ups (v1 scope)

- **Draft-only items** (never-deployed drafts) are not yet surfaced by the endpoint; deferred to a follow-up (still reachable via the drafts UI).
- The **label** filter and instant fuzzy ranking still operate over loaded pages (kind/owner/folder/search are workspace-wide).

## Screenshots

Sort selector (four orders, icon-only trigger):

sort-dropdown

Lazy folder tree — every folder shown as a collapsed top-level node (17k-item workspace):

tree-lazy-collapsed

Expanded folder with nested subfolders + per-folder "Load more in this folder":

tree-lazy-expanded-subfolders

## Test plan

- [x] Endpoint validated via curl + 5 backend integration tests: all four orders, keyset pagination + cross-kind ties, starred-first, favorited archived-version pagination, `kinds`/`search`/`path_start`/`show_archived` filters.
- [x] `EXPLAIN ANALYZE` confirms per-page-bounded correlated projections and index-eligible ordering (~12-16 ms at 17k items).
- [x] Homepage (Playwright, 17k-item seed): four sorts reorder correctly (flat); load-more pages the cursor; server-side kind/owner filters; hybrid search finds items far past the loaded window; lazy tree shows all folders, expands + paginates per-folder, nests subfolders; **flat list stays contiguous & dedup-free after folder expansion**; sort change collapses expanded folders; tree+search hides empty folders; 0 feature-related console errors.
- [x] Full `npm run check` clean (only the pre-existing unrelated `modern-screenshot` error remains).
- [ ] `cargo check` + tests in CI.

### DIFF
--- a/backend/migrations/20260724094737_runnables_sort_indexes.down.sql
+++ b/backend/migrations/20260724094737_runnables_sort_indexes.down.sql
@@ -1,2 +1,5 @@
 DROP INDEX IF EXISTS index_script_on_workspace_created_at;
 DROP INDEX IF EXISTS index_flow_on_workspace_edited_at;
+DROP INDEX IF EXISTS index_script_on_workspace_name;
+DROP INDEX IF EXISTS index_flow_on_workspace_name;
+DROP INDEX IF EXISTS index_app_on_workspace_name;
--- a/backend/migrations/20260724094737_runnables_sort_indexes.down.sql
+++ b/backend/migrations/20260724094737_runnables_sort_indexes.down.sql
@@ -1,0 +1,2 @@
+DROP INDEX IF EXISTS index_script_on_workspace_created_at;
+DROP INDEX IF EXISTS index_flow_on_workspace_edited_at;
--- a/backend/migrations/20260724094737_runnables_sort_indexes.up.sql
+++ b/backend/migrations/20260724094737_runnables_sort_indexes.up.sql
@@ -2,20 +2,21 @@
 -- keyset-paginated query stays an index scan (Merge Append + LIMIT) even on big
 -- workspaces, for both the time orders and the name orders.
 
--- Time orders. Scripts accumulate archived version rows, so scope that index to
--- the visible non-archived set the homepage browses.
+-- Time orders. Not scoped to archived = false: the endpoint also serves the
+-- "Only archived" view, and current (non-archived) versions carry the newest
+-- created_at so they cluster at the head of the index for the default browse
+-- anyway, while the archived view still gets an ordered index scan.
 CREATE INDEX IF NOT EXISTS index_script_on_workspace_created_at
-    ON script (workspace_id, created_at DESC)
-    WHERE archived = false;
+    ON script (workspace_id, created_at DESC);
 
 CREATE INDEX IF NOT EXISTS index_flow_on_workspace_edited_at
     ON flow (workspace_id, edited_at DESC);
 
 -- Name orders sort on the lowered summary-or-path expression, so a matching
--- expression index makes that order presorted instead of a full sort per page.
+-- expression index makes that order presorted instead of a full sort per page
+-- (covers both the default and archived views).
 CREATE INDEX IF NOT EXISTS index_script_on_workspace_name
-    ON script (workspace_id, lower(COALESCE(NULLIF(summary, ''), path)))
-    WHERE archived = false;
+    ON script (workspace_id, lower(COALESCE(NULLIF(summary, ''), path)));
 
 CREATE INDEX IF NOT EXISTS index_flow_on_workspace_name
     ON flow (workspace_id, lower(COALESCE(NULLIF(summary, ''), path)));
--- a/backend/migrations/20260724094737_runnables_sort_indexes.up.sql
+++ b/backend/migrations/20260724094737_runnables_sort_indexes.up.sql
@@ -1,15 +1,24 @@
--- Indexes backing the unified homepage runnables listing when ordered by
--- last-updated time. The name orders already ride the existing
--- workspace_id+path indexes (script index_script_on_path_created_at, flow_pkey,
--- app unique_path_workspace_id), and apps are one row per path and few enough to
--- sort without a dedicated index. These two cover the large tables so the merged
--- keyset-paginated query stays an index scan even on big workspaces.
+-- Indexes backing the unified homepage runnables listing so the merged,
+-- keyset-paginated query stays an index scan (Merge Append + LIMIT) even on big
+-- workspaces, for both the time orders and the name orders.
 
--- Scripts accumulate archived version rows, so scope the index to the visible
--- non-archived set the homepage browses.
+-- Time orders. Scripts accumulate archived version rows, so scope that index to
+-- the visible non-archived set the homepage browses.
 CREATE INDEX IF NOT EXISTS index_script_on_workspace_created_at
     ON script (workspace_id, created_at DESC)
     WHERE archived = false;
 
 CREATE INDEX IF NOT EXISTS index_flow_on_workspace_edited_at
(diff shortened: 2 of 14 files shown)


rubenfiszel:
### Review responses (head 14b3a16)

**Pi — "keyset duplicates across kind boundaries": this is a false positive.** The worked example has the `kind` tiebreak direction backwards. The keyset compares `(sort_key, path, kind, tiebreak)` and the ORDER BY uses the *same* four keys in the *same* direction, so a cross-kind tie resolves consistently — a row on page 1 can't reappear on page 2. Verified empirically (a script and flow at the exact same path+timestamp+summary, paginated one-at-a-time across all three orders → zero dupes, zero skips) and now pinned by an integration test: `backend/tests/runnables_list_pagination.rs::test_runnables_keyset_no_dupes_across_kind_ties` (passing).

**Codex #1 — scoped tokens 400 on this route: safe by design, keeping it.** The scope-domain middleware (`extract_domain_from_route`) rejects scoped tokens on `/runnables/list` *before* the handler, so there is no enumeration/leak — the earlier concern is mitigated by the 400. This endpoint is the session-auth homepage list; scoped-token API consumers use the per-kind `/scripts/list`, `/flows/list`, `/apps/list` endpoints, which are scope-mapped. I've kept the in-SQL per-domain scope filter as defense-in-depth for if this route is ever added to a scope domain. Registering a `runnables` scope domain would require cross-domain semantics ("allow if the token holds read on *any* of scripts/flows/apps") in security-critical middleware, which I don't think is worth the risk for a session-oriented endpoint — happy to reconsider if you'd prefer scoped tokens to work here.

**Addressed this round:** archived-version cursor (`hash` tiebreak, #3), reachable pagination in empty/tree states (#2), debounced server-side search augmentation keeping instant client search (#4), `LIKE` wildcard escaping, tree "load more" dead-clicks, and the null-serialization → `skip_serializing_if`.

rubenfiszel:
### Round-3 responses (head 7540a06)

**Fixed:** archived index coverage (#4 — dropped the `archived=false` partial clause so the "Only archived" view is index-ordered; the default browse is unaffected because current versions carry the newest `created_at` and sit at the head of the index), search view-scope (#1 — now sends `showArchived`/`includeWithoutMain`), and the search workspace race (#2 — the in-flight response is dropped if the workspace or term moved on).

**#3 (app updated-order sorts): won't index, by design.** An app's update time is `app_version.created_at` reached through the `versions[]` join, so it can't be emitted in timestamp order by any index on `app` (and it must stay a real timestamp to merge correctly against script/flow `sort_time`). Denormalizing an `edited_at` column onto `app` with a trigger + backfill is a lot of surface for a branch that is one row per path and, in practice, few — the large-workspace premise is about the numerous kind (scripts), which is index-backed. So the app branch sorts its (small) row set before the per-branch limit; on realistic workspaces that's negligible. Happy to add the denormalized column if you consider app-count-at-scale a real target.

rubenfiszel:
Re: Codex finding #3 (updated-order app pages not index-backed) — addressing #1 and #2 in a09670167d; **dismissing #3** with reasoning:

The app sort key is the latest `app_version.created_at`, and `app` has no denormalized timestamp column to index — the value is reached via `app_version(id)` (the primary key), so the per-app lookup in the join is already index-backed. What remains un-indexed is ordering the *parent* `app` rows by that child value, which is a parent-by-child-aggregate sort that no single-table index can serve.

Making it fully index-ordered would require denormalizing `latest_version_created_at` onto `app` and maintaining it on every deploy — a change to the app write path that's disproportionate here, because:
- app counts are typically tens–low-hundreds per workspace (apps are heavier UI artifacts, far fewer than the scripts/flows the new time indexes target), so the sort is small in practice;
- the join itself is PK-indexed, and the keyset shrinks the candidate set on deep pages;
- the app **name** order *is* index-backed by this migration.

I've left it as a documented follow-up should app-heavy workspaces become a real bottleneck, rather than reshaping the deploy path in this homepage PR. Happy to file that follow-up if preferred.

rubenfiszel:
✅ Review round clean @ fbdcca8003a641ca7a441c753459e026dbc93d97 — nit-only verdicts at a09670167dd625955ab2dc159a7a422b3d168374 (Claude & Pi: good to merge; Codex: mergeable). The sole nit (drafting-history comment in folders.rs) is addressed in fbdcca8003a641ca7a441c753459e026dbc93d97; the app-index P1 was dismissed with reasoning above. PR is ready for review.

PR #10301, opened 2026-07-24
fix: pin table actions column so it stays visible on narrow screens

rubenfiszel:
## Summary

Wide `DataTable`s scroll horizontally on narrow screens, and the trailing per-row **actions column** (the `⋯` menu, Edit/Delete, and the folders page's **Publish to Hub** entry) was pushed off the right edge into the horizontal-scroll overflow, where it was effectively unreachable.

This pins that actions column to the right edge of the scroll container so it stays visible regardless of table width.

## Changes

- **`Cell.svelte`**: add an opt-in `stickyEnd` prop. When set, the cell is pinned with `sticky right-0` plus a `border-l` divider and an **opaque** background (`bg-surface` for body cells, `bg-surface-secondary` for header cells). The background is deliberately opaque rather than the row's hover tint (`bg-surface-hover` is translucent, `rgba(...,0.2)`), so cells sliding underneath are occluded instead of bleeding through.
- **folders / variables / resources pages**: mark the trailing actions column (header + body cells) with `stickyEnd`. Folders is the widest (10 columns) and the page where this was originally noticed; resources covers both its tables (resources + resource types).
- On the resources workspace table, the actions cell used `class="flex justify-end"` directly on the `Cell`, which forces the `<td>` to `display: flex` and breaks `position: sticky` (a flex box in a table row is wrapped in an anonymous table-cell). Moved the flex layout to an inner `<div>` so the `<td>` stays `display: table-cell` and pins correctly.

Groups (3 columns) and assets (4 columns) were checked too — their last cell holds data rather than actions and the tables don't meaningfully overflow, so they were left unchanged.

## Screenshots

Verified in-browser (Playwright) on the `admins` workspace at narrow viewports.

**Folders — actions column pinned while the table overflows (hover row, no bleed-through):**

folders-fixed-hover

**Folders — scrolled fully right: the Participants column slides cleanly under the pinned actions column:**

folders-scrolled-right

**Resources (Workspace tab) — scrolled right at 400px: `⋯` menu + Manage buttons stay pinned while the Path column scrolls off (this is the case the flex→table-cell fix addresses):**

resources-fixed-scrolled

**Variables:**

variables-narrow

**Resources (Resource Types tab):**

resource-types

## Test plan

- [ ] On a narrow viewport, open `/folders` in a workspace that has folders and confirm the `⋯` actions menu (with **Publish to Hub**) is always visible at the right edge, even while the rest of the table scrolls horizontally.
- [ ] Hover a row and scroll the table right — the pinned column stays opaque with no content bleeding through it.
- [ ] Repeat on `/variables` and `/resources` (both tabs); on resources Workspace, narrow enough to force horizontal scroll and confirm the actions cell stays pinned (not just the header).
- [ ] Confirm no visual regression on wider screens (the column looks like a normal last column with a left divider).

### DIFF
--- a/frontend/src/lib/components/table/Cell.svelte
+++ b/frontend/src/lib/components/table/Cell.svelte
@@ -11,6 +11,7 @@
 		shouldStopPropagation?: boolean
 		selected?: boolean
 		sticky?: boolean
+		stickyEnd?: boolean
 		wrap?: boolean
 		children?: import('svelte').Snippet
 		[key: string]: any
@@ -24,6 +25,7 @@
 		shouldStopPropagation = false,
 		selected = false,
 		sticky = false,
+		stickyEnd = false,
 		wrap = false,
 		children,
 		...rest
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ ecda96ae411c81b392a574e3f37d8460f52b32ad — codex, claude and pi all "Good to merge".

PR #10303, opened 2026-07-24
fix: pin validated DNS address to close SSRF DNS-rebinding TOCTOU

rubenfiszel:
## Summary

Fixes a DNS-rebinding TOCTOU SSRF.

`validate_url_for_ssrf` in `windmill-common/src/ssrf.rs` resolved the host, checked every address was public, then **discarded** the addresses and returned `Ok(())`. Every caller re-used the hostname string and let stock reqwest re-resolve at connect time, with no pinning. A TTL-0 DNS rebinder that answers a public IP at check-time and an internal one (e.g. `169.254.169.254`) at connect-time slipped straight through the guard — reachable by any authenticated workspace member via a self-created customai / MCP / websocket resource, letting them read cloud IMDS / instance credentials.

The fix makes validate-time and connect-time target the **same** address: `validate_url_for_ssrf` now returns a `ValidatedTarget { host, addrs }` and each caller pins those addresses onto the client that actually connects (`reqwest`'s `resolve_to_addrs`, or a raw pinned TCP connect for the WebSocket).

## Changes

- **`windmill-common/src/ssrf.rs`**: `validate_url_for_ssrf` (+ the `validate_saml_metadata_url` / `validate_mcp_server_url` / `validate_mcp_server_url_for_bad_request` wrappers) return `ValidatedTarget` instead of `()`. `apply_dns_pinning(ClientBuilder)` wraps `resolve_to_addrs`; `pinned_addrs()` exposes the std `SocketAddr`s for non-reqwest callers. IP-literal hosts and `ALLOW_PRIVATE_*` opt-outs return empty `addrs` (nothing to pin — behavior preserved). Corrected the doc comment that falsely claimed the function "prevents DNS rebinding".
- **AI proxy** (`windmill-api/src/ai.rs`) and **worker AI agent** (`windmill-ai/src/utils.rs` + `windmill-worker/src/ai_executor.rs`): a `pinned_ai_client_for(url)` helper builds a per-request client pinned to the validated `base_url` address; covers the API proxy (both dispatch paths + Google-native), the OAuth `token_url`, and the worker executor — the primary readable-SSRF sink (returns the upstream body). Honors `ALLOW_PRIVATE_AI_BASE_URLS` and keeps redirects disabled.
- **MCP** (`windmill-mcp/src/client/mod.rs`, `client_registration.rs`, `lib.rs`): pins the streamable-HTTP client and the OAuth DCR / discovery / token-exchange clients (the token exchange carries `client_secret`). This crate resolves a different reqwest version via `rmcp`, so it pins with the std-typed `host`/`addrs` directly rather than `apply_dns_pinning`.
- **WebSocket trigger** (`windmill-trigger-websocket/`): `validate_websocket_url_for_ssrf` returns the target; `connect_async_with_proxy` pins the direct TCP connect to the validated address (`ws`/`wss` map to `http`/`https` for the check, so the port carries through).
- **EE** (companion PR): SAML metadata loader and MCP OAuth discover/start/callback handlers pinned the same way.

## Known limitations (documented in code)

- **Outbound egress proxy**: pinning governs only *direct* connections. When `HTTP_PROXY`/`HTTPS_PROXY` is set, the proxy resolves the target host itself and the pin cannot reach that hop — an inherent property of proxy-based egress shared by every app-side SSRF guard. The public/private pre-check still runs.
- **MCP OAuth discovery**: `rmcp`'s `discover_metadata` may follow server-supplied secondary metadata URLs on other hosts that the per-host pin does not cover — a pre-existing gap in rmcp discovery (the original client was equally unpinned). Fully closing it needs a validating resolver at the rmcp layer; out of scope for this rebinding pin.

## Test plan

- [x] `cargo test -p windmill-common ssrf::` — 19 pass, incl. new `validate_url_surfaces_resolved_addrs_for_pinning` and `validate_url_ip_literal_pins_nothing`
- [x] `cargo check --features enterprise,private,mcp,websocket,enterprise_saml` — clean
- [x] clippy on touched crates — no new warnings
- [ ] Manual: AI provider proxy (global + workspace) against a public `base_url` still succeeds; `ALLOW_PRIVATE_AI_BASE_URLS=true` against a private endpoint still connects
- [ ] Manual: WebSocket trigger to a public `wss://` gateway connects; private-IP URL still rejected; still allowed under `ALLOW_PRIVATE_WEBSOCKET_URLS`
- [ ] Manual: SAML metadata loads from a public IdP; `ALLOW_PRIVATE_SAML_METADATA_URLS` still permits an internal IdP

Companion EE PR pins the `*_ee.rs` sinks (SAML + MCP OAuth). Thanks to **Doruk**, who offered to verify the fix.


rubenfiszel:
Addressed the review round:

- **Codex P2 (duplicated DNS-pin comments in `ai.rs`)** — fixed in f808a68: dropped the repeated inline comments at the three call sites; the contract stays on `pinned_ai_client_for`.

Dismissing the two remaining nits as deliberate, non-blocking tradeoffs:

- **Per-request client drops connection pooling (Claude/Pi P2).** This is inherent to per-target DNS pinning: the pin must be baked onto the client at build time, and these clients are shared `lazy_static` singletons that can't take a per-request `resolve()`. A `(host, sorted-addrs)` client cache would restore pooling but adds staleness/eviction logic to a security control — a stale cached pin is exactly the failure mode this PR closes, so I'd rather not introduce that surface here. LLM latency dominates these paths and both reviewers flagged it non-blocking. Leaving as-is.

- **~50 lines of `pinned_ai_client_for` duplicated across `windmill-api/ai.rs` and `windmill-ai/utils.rs` (Pi).** The two wrap genuinely different pooled clients with different configs (`HTTP_CLIENT`: per-request timeout up to 24h + pooling; `AI_HTTP_CLIENT`: 10s connect-timeout, redirect policy governed by `ALLOW_AI_BASE_URL_REDIRECTS`) living in different crates. A shared generic would trade a little duplication for cross-crate coupling and a closure-over-builder indirection; for a focused security fix I'd rather keep each helper local and legible. Leaving as-is.

EE companion (windmill-labs/windmill-ee-private#684, now `f860781`) pins the SAML metadata fetch and the MCP OAuth discovery/token-exchange clients — codex confirmed this against the bumped ref.

rubenfiszel:
✅ Review round clean @ f808a68fbfb172cd0e7f98803ec03de75eced766 — nit-only verdicts at acb32c16ed8 (Pi: Good to merge; Claude & Codex: Mergeable, nits); codex nit addressed in f808a68, pooling + helper-dup nits dismissed in review replies.

PR #10304, opened 2026-07-24
perf: optimize get_datatable_full_schema to avoid timeout on large catalogs

rubenfiszel:
## Summary
`get_datatable_full_schema` times out (30s) on large catalogs (e.g. the `hub` workspace's `datatable://main`). The column-introspection query in `pg_get_full_schema` built every column row with three per-column correlated subqueries — default value (`pg_attrdef`), primary-key `EXISTS` (`pg_index`), and pk constraint name (`pg_constraint`). On a catalog with many tables/columns those subqueries run once per column, turning introspection into O(columns) index searches.

This replaces them with plain joins: a `LEFT JOIN` to `pg_attrdef` for defaults and a single `LEFT JOIN` to the table's primary-key constraint (a table has at most one, so the join stays 1:1) providing both `is_primary_key` (via `conkey @> ARRAY[a.attnum]`) and `pk_constraint_name`. The planner now does one hash/merge join instead of per-column correlated scans.

## Changes
- Rewrite the columns query in `pg_get_full_schema` (`backend/windmill-common/src/query_builders.rs`) to use joins instead of correlated subqueries. Output is byte-for-byte identical; the FK query is unchanged.

## Validation
- Verified byte-for-byte identical output between the old and new query against the local dev catalog (1262 column rows, `diff` clean).
- `EXPLAIN (ANALYZE, BUFFERS)`: the old plan shows `SubPlan 2`/`SubPlan 3` looping 1262× with ~11k shared-buffer hits; the new plan replaces them with hash joins and drops to a few hundred buffer hits.
- Backend compiles cleanly (`cargo watch`, health check passed).

## Test plan
- [ ] Call `POST /w/{workspace}/workspaces/get_datatable_full_schema` on a large catalog and confirm it returns within the timeout.
- [ ] Confirm the returned schema (columns, datatypes, defaults, nullability, primary keys, pk constraint names, foreign keys) matches expectations.

Fixes WIN-2239

---
## Summary by cubic
Optimize `get_datatable_full_schema` to avoid 30s timeouts on large catalogs by switching from per-column subqueries to join-based introspection. Addresses WIN-2239 with identical returned schema.

- **Refactors**
  - Use LEFT JOINs to `pg_attrdef` (defaults) and the table’s primary-key `pg_constraint` to derive `is_primary_key` and `pk_constraint_name`.
  - Keep FK introspection unchanged.
  - Moves from O(columns) correlated subqueries to a single join plan for consistent performance on big catalogs.

<sup>Written for commit a6d596759333e7f73939ff24be19670fd4ca10ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-common/src/query_builders.rs
+++ b/backend/windmill-common/src/query_builders.rs
@@ -4835,6 +4835,10 @@ fn pg_action_to_string(action: &str) -> String {
 pub async fn pg_get_full_schema(
     client: &tokio_postgres::Client,
 ) -> Result<FullDatabaseSchema, String> {
+    // Primary-key and default-value info are joined in (a table has at most one
+    // primary-key constraint, so `pkc` stays 1:1) rather than fetched via
+    // per-column correlated subqueries — on large catalogs those subqueries run
+    // once per column and make the introspection time out.
     let column_rows = client
         .query(
             "SELECT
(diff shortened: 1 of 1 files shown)


rubenfiszel:
✅ Review round clean @ a6d596759333e7f73939ff24be19670fd4ca10ce

rubenfiszel:
### Verified against a live Cloud SQL instance (AWS RDS, PostgreSQL 14.17)

Ran both the old and new columns query against a real managed Postgres catalog (355 tables / 2630 columns / 1276 indexes):

- **Output identical** — byte-for-byte diff clean over all 2630 column rows.
- **New SQL is valid on PG 14** — `conkey @> ARRAY[a.attnum]` array-containment and `NOT a.attnotnull` behave as expected.
- **Plan / I/O** (`EXPLAIN (ANALYZE, BUFFERS)`):
  | | old | new |
  |---|---|---|
  | per-column `SubPlan` loops | 3 subplans × `loops=2630` | none (hash joins) |
  | shared buffer hits | 24466 | 5053 (~4.8× fewer) |

The buffer/loop count scales with the number of columns, so eliminating the per-column subplans is exactly what keeps a large catalog under the timeout. On this modest catalog it's already ~5× less I/O.

PR #10307, opened 2026-07-24
chore: auto-allow rm in /tmp and git repos, plus read-only gmail

rubenfiszel:
## Summary
Reduces permission-prompt friction in `.claude/settings.json`: deleting files under `/tmp` or inside git repos in `~`, and reading/searching Gmail via the MCP connector. The prompts came from two blanket `ask` rules, and since Claude Code evaluates `deny > ask > allow`, no `allow` rule could override them.

## Approach
- **`rm`** is handled by an allow-only `PreToolUse` hook (`.claude/hooks/guard-rm-outside-tmp.sh`). A PreToolUse `allow` overrides the ask rule, so the hook returns `allow` only for a provably-safe `rm` and makes **no decision** for anything else. The blanket **`Bash(rm:*)` / `Bash(rmdir:*)` ask rules are kept as the safety net**, so every other form falls through and prompts, with the auto-mode classifier as a further backstop.

  The hook is **deny-by-default**: it allows only a single-line, leading bare `rm` where every token consists solely of a safe character set (`[A-Za-z0-9._/*?[]-]` — which excludes all of bash's quoting/expansion/separator characters), no wildcard appears in a non-final path segment (symlink-traversal guard), and each operand (resolved against the command's cwd via `realpath -m`) is an auto-allowable target:
  - under `/tmp`, or
  - strictly inside a git working tree located under `$HOME` (git-recoverable). The ancestor walk stops at `$HOME`, so a dotfiles repo at `~` can't make all of home deletable, and top-level `~` files (`.bashrc`, `.ssh/…`) stay protected. A `.git` path and a primary checkout's root folder are never auto-allowed; a linked worktree's root folder **is** (its `.git` is a pointer file, so history in the main repo survives).

- **Gmail**: removed the blanket `mcp__claude_ai_Gmail` ask; allowed the read-only tools (`search_threads`, `get_thread`, `get_message`, `list_labels`, `list_drafts`) and explicitly gated every mutating tool (`create_draft`, `update_draft`, `create_label`, `label_*`, `unlabel_*`, `apply_sensitive_*`).

## Test plan
- [x] `jq empty .claude/settings.json` passes; `bash -n` on the hook passes
- [x] Hook unit-checked against real host paths — **allow**: `/tmp/*`, files/subdirs inside git repos (windmill and non-windmill, absolute and cwd-relative), final-segment globs, linked-worktree root folders; **defer**: primary-checkout root, `.git` paths, top-level `~` files, `~/.ssh`, traversal out of a repo, non-repo dirs, outside `~`, wrappers, quotes/backslash/brace/`$`/backtick, operators, multi-line, `--`/lone-`-`/glob-in-option tokens; **no decision**: non-`rm`
- [ ] After reload: `rm ~/<repo>/src/x` and `rm -rf ~/<repo>__worktrees/foo` run without a prompt; `rm -rf ~/<primary-repo>`, `rm ~/.ssh/x`, and `rm -rf /etc/x` still prompt
- [ ] With Gmail connected: `search_threads`/`get_thread` run without a prompt; `create_draft` and label mutations prompt


rubenfiszel:
Addressing the last review round (head 3d13c687be):

- **Claude nit (fixed in c0f94aee45):** corrected the hook comments to stop claiming git-tree deletions are fully "recoverable." They aren't — untracked/`.gitignore`'d/uncommitted content and a nested independent repo's history are not restorable. The rationale is now stated honestly: the git-tree allowance trades on "this is a version-controlled project dir" being lower-stakes, as a deliberate convenience tradeoff.

- **Codex P0 (recursive delete of a parent containing an independent nested repo):** acknowledged and accepted as residual rather than fixed. Fully closing it requires downward `.git` scanning on every recursive delete (a full-subtree `find`, slow on trees like `node_modules`) and still can't cover the untracked/uncommitted gap. The realistic exposure is narrow — an *independent* nested checkout inside another repo's working tree (submodules are recoverable via `git submodule update`), and only when the agent runs `rm -rf` on its parent. Given Pi = "Good to merge" and Claude = "Mergeable," this is a documented, accepted tradeoff. The `Bash(rm:*)` ask rule + auto-mode classifier remain as backstops for anything the allow-path doesn't cover.

PR #10308, opened 2026-07-24
fix: do not apply workspace display name on git-sync pull

rubenfiszel:
## Summary

The workspace display name is one of the settings tracked in git sync (`settings.yaml`, the `name` field). On every pull, `pushWorkspaceSettings` (the git→windmill apply path) called `changeWorkspaceName` whenever the tracked name differed from the live workspace name.

Because `settings.yaml` lives at the repo root and is shared across the branches of a repo, this let one workspace's name overwrite another's when two workspaces sync the same repo. Concretely: with a prod workspace on `master` and a staging workspace on `staging` of the same repo, once a `settings.yaml` carrying the staging name reached the branch prod pulls from, an automatic git-to-windmill pull silently renamed the prod workspace, with no user action.

This PR stops applying the display name on pull. The name is still written to `settings.yaml` on push (kept for reference), but a live workspace is only ever renamed by its owner via the UI/API, never by a pull.

## Changes

- `cli/src/core/settings.ts`: remove the `changeWorkspaceName` apply from `pushWorkspaceSettings` (the pull-apply path); exclude `name` from the `isSuperset` early-return so a name-only diff stays a clean no-op.
- `cli/test/push_workspace_settings_name_unit.test.ts`: regression guard — a differing `name` in `settings.yaml` is never applied while other differing settings still apply; a name-only diff is a complete no-op.
- **Hub delivery for backend auto-pull**: the backend auto-pull (`enqueue_git_pull_job`) runs a pinned Hub script that bundles a *published* `windmill-cli`, so the source fix only reaches auto-pull via a republished bundle. Published `windmill-cli@1.769.1` with the fix, republished both git-sync hub scripts to pin it, and bumped the pins:
  - `backend/windmill-common/src/workspaces.rs`: `GIT_SYNC_PULL_SCRIPT_PATH` 28795→28808, `LATEST_GIT_SYNC_SCRIPT_PATH` 28796→28809.
  - `frontend/src/lib/hubPaths.json`: `gitInitRepo` 28795→28808.
  - Both new hub scripts (28808 pull, 28809 sync) verified to bundle `windmill-cli@1.769.1`, which is verified to no longer emit the `changeWorkspaceName` pull-apply.

## Test plan

- [x] `UNIT_ONLY=1 bun test test/push_workspace_settings_name_unit.test.ts` passes; fails when the apply is re-introduced (verified)
- [x] Full CLI unit suite green (976 pass)
- [x] `windmill-cli@1.769.1` bundle confirmed to drop the name-apply (`changeWorkspaceName` call site + "Updating workspace name" log gone vs 1.769.0)
- [ ] Git Sync Integration Tests green (triggered by the `workspaces.rs` pin change)
- [ ] Manual: an automatic git→windmill pull with a differing tracked `name` leaves the workspace display name unchanged while other settings still apply

### DIFF
--- a/backend/windmill-common/src/workspaces.rs
+++ b/backend/windmill-common/src/workspaces.rs
@@ -167,7 +167,7 @@ pub enum ObjectType {
     DatatableMigration,
 }
 
-pub const LATEST_GIT_SYNC_SCRIPT_PATH: &str = "hub/28796/sync-script-to-git-repo-windmill";
+pub const LATEST_GIT_SYNC_SCRIPT_PATH: &str = "hub/28809/sync-script-to-git-repo-windmill";
 
 /// Hub script that applies a repository's state back into a workspace
 /// (the repo → Windmill / "pull" direction). Same script the UI runs from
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ fc6d5aa73c7fe0b1caa5470819d7fb8e970ea3a8

PR #10311, opened 2026-07-24
fix: match WAC waitForApproval resume to its own step

rubenfiszel:
## Summary
Sequential `waitForApproval()` calls in a workflow-as-code (WAC) job all share the same parent `job_id`. On resume, `prepare_checkpoint_for_resume` read the approval event with `WHERE job = $1 ORDER BY created_at ASC LIMIT 1` — always the **oldest** `resume_job` row for the job, and rows are never deleted. So every approval after the first re-injected the first approval's `{value, approver, approved}`, and canceled/timed-out steps inherited it too.

Repro (all three returned `{approved: true}` before the fix):

```python
from wmill import wait_for_approval, workflow

@workflow
async def main():
    a = await wait_for_approval(5)  # approved in UI
    b = await wait_for_approval(5)  # canceled in UI
    c = await wait_for_approval(5)  # not touched (timeout)
    return {"a": a, "b": b, "c": c}
```

Each approval already carries a unique `resume_id` derived from its step key (embedded in its resume/cancel URL and stored in `resume_job.resume_id`). The fix matches on that `resume_id` so each step consumes only its own resume event: a canceled step keeps `approved=false` on its own row, and a timed-out step matches no row and falls back to the existing `approved=false` default.

Fixes WIN-2240

## Changes
- `wac_executor.rs`: extract the step-key → `resume_id` hash into a shared `approval_resume_id()` helper, and use it in `prepare_checkpoint_for_resume` to fetch `resume_job WHERE job = $1 AND resume_id = $2` instead of the oldest row.
- `bun_executor.rs`: replace the inline key-hash in the approval-URL generation with the shared `approval_resume_id()`, so URL generation and resume matching can never drift.

## Test plan
- [x] Ran a 3-step WAC workflow (TS/bun) with sequential `waitForApproval()`: approved `a`, canceled `b`, let `c` time out.
- [x] Result: `a = {value:{step:a}, approved:true}`, `b = {value:{step:b}, approved:false}`, `c = {value:null, approved:false, approver:null}` — each step resolves to its own event.
- [x] Backend compiles cleanly under `cargo watch` (no new warnings/errors).

---
## Summary by cubic
Fixes mismatched resumes for sequential WAC `waitForApproval()` steps by matching each resume event to its own step via `resume_id`, preventing approvals from bleeding across steps. Adds a regression test to pin this behavior. Addresses WIN-2240.

- **Bug Fixes**
  - Match resume events by `resume_id` per step in `prepare_checkpoint_for_resume` (`resume_job WHERE job AND resume_id`, newest first).
  - Each step now uses its own approval; canceled/timed-out steps no longer inherit previous results.

- **Refactors**
  - Expose and reuse `approval_resume_id(key)` for both resume matching and approval URL generation in `bun_executor`; the integration test consumes it.

<sup>Written for commit 866039c0fc8d67c3c989dec158b5474ccc6d631f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/wac_approval_resume.rs
+++ b/backend/tests/wac_approval_resume.rs
@@ -1,0 +1,104 @@
+//! Regression test for WAC `waitForApproval` reading the wrong approval result.
+//!
+//! Sequential `waitForApproval()` calls share the parent job_id, so the resume
+//! lookup must select the `resume_job` row whose `resume_id` matches the pending
+//! approval step's key — not the oldest row for the job. Before the fix, a job's
+//! second approval re-read the first approval's value (and cancels/timeouts
+//! inherited the first approval's `approved=true`).
+
+use serde_json::json;
+use sqlx::{Pool, Postgres};
+use uuid::Uuid;
+
+use windmill_common::wac::{save_checkpoint, WacCheckpoint, WacPendingSteps};
+use windmill_worker::wac_executor::{approval_resume_id, prepare_checkpoint_for_resume};
+
+async fn insert_resume(
+    db: &Pool<Postgres>,
+    job_id: Uuid,
+    key: &str,
+    value: serde_json::Value,
+    approved: bool,
+    age_seconds: i64,
+) -> anyhow::Result<()> {
+    sqlx::query(
+        "INSERT INTO resume_job (id, job, flow, created_at, value, approver, resume_id, approved)
+         VALUES ($1, $2, $2, now() - make_interval(secs => $3), $4, 'tester', $5, $6)",
+    )
+    .bind(Uuid::new_v4())
+    .bind(job_id)
+    .bind(age_seconds as f64)
+    .bind(value)
+    .bind(approval_resume_id(key) as i32)
+    .bind(approved)
+    .execute(db)
+    .await?;
+    Ok(())
+}
+
+#[sqlx::test(fixtures("base"))]
+async fn resume_matches_pending_approval_step(db: Pool<Postgres>) -> anyhow::Result<()> {
+    let job_id = Uuid::new_v4();
+
+    // Minimal queue row so the resume_job.flow FK and the v2_job_status FK resolve.
+    sqlx::query(
+        "INSERT INTO v2_job_queue (id, workspace_id, scheduled_for, suspend)
+         VALUES ($1, 'test-workspace', now(), 1)",
+    )
+    .bind(job_id)
+    .execute(&db)
+    .await?;
+
+    // The workflow is parked on its *second* approval.
+    let pending_key = "approval_2";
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Addressed the nits in `866039c0fc`:

- **Regression test** (Codex P2 / `docs/validation.md`): added `backend/tests/wac_approval_resume.rs`. It parks a job on its 2nd `waitForApproval()` with two `resume_job` rows — an older *approved* row for the 1st approval and a newer *cancelled* row for the pending one — and asserts the checkpoint resolves to the pending step's own event (`{which: second}, approved: false`). Confirmed it **fails** against the pre-fix `ORDER BY created_at ASC LIMIT 1` query (reads `{which: first}`) and **passes** with the `resume_id` filter.
- **Comment length** (Codex P2 / Pi P2): trimmed the resume-match comment to 4 lines and dropped the bug-history narration.
- **`pub` vs `pub(crate)`** (Pi P2): keeping `approval_resume_id` as `pub` — it now has an external consumer, the integration-test crate (`windmill_worker::wac_executor::approval_resume_id` in `backend/tests/wac_approval_resume.rs`), which computes the expected `resume_id` from the same hash to stay coupled to production. `pub(crate)` would break that.

New head: `866039c0fc8d67c3c989dec158b5474ccc6d631f`

rubenfiszel:
Closing in favor of #10314.

This approach (`WHERE resume_id = hash(step_key)`) is **broken for the primary approval channel**. `resume_id = hash(key)` is only stored by the inline resume URL. The actual WAC approval UI — the approve/cancel buttons in `WorkflowTimeline.svelte` — calls `resumeSuspended` → `/jobs_u/flow/resume_suspended/{job_id}`, which generates `resume_id = rand::random()` (`jobs.rs:3237`). Resume-as-owner stores `resume_id = 0` (`jobs.rs:~3099`), and Slack/Teams links are random too. For all of those, filtering on `hash(key)` matches no row, so the approval would be wrongly treated as a timeout.

My regression test didn't catch this because it inserts every row with `resume_id = hash(key)`, so it only ever exercised the one channel that happens to match.

#10314's `consumed_resume_ids` approach is channel-agnostic (exclude rows already consumed by earlier steps, take the oldest remaining) and handles all of these correctly. That's the right fix.

PR #10312, opened 2026-07-24
fix: show scheduled singlestepflow runs in flow history sidebar

rubenfiszel:
## Summary

The flow-view "History" run sidebar was silently excluding scheduled runs of some flows, while the "All runs" page (`/runs`) showed them. A customer's daily-scheduled flow appeared to have no recent history in the sidebar.

**Root cause:** The sidebar is backed by `get_input_history` (`backend/windmill-api-inputs/src/lib.rs`), which filtered job kind to `('flow','flowpreview')`. But when a schedule has a **dynamic-skip handler** (`schedule.dynamic_skip`) — or a scheduled script has **native retry** — the run is wrapped in a synthetic `SingleStepFlow` (`backend/windmill-queue/src/schedule.rs:273`,`:408`), landing under `kind='singlestepflow'`, not `flow`/`script`. Those runs were filtered out. Manual UI runs bypass the schedule path and run as plain `flow`, so only those showed up.

Verified against production data: the affected flow's daily scheduled runs are all `kind='singlestepflow'` (success on market days, `skipped` on non-market days via the skip handler), and the exact sidebar query returned only the single manual `flow` run.

## Changes

- `get_input_history`: include `JobKind::SingleStepFlow` in the job-kind filter for **flow** (`FlowPath`) and **script** (`ScriptPath`) runnables, so their scheduled singlestepflow runs surface. Consistent with `JobKind::is_flow()`, which already treats `SingleStepFlow` as a flow.
- **Disambiguate script- vs flow-wrapped `singlestepflow`:** a singlestepflow wraps either a script or a flow, and a script and flow may legally share a `runnable_path`. Each history side now matches only its own wrapped kind via `raw_flow.modules[id in ('a','main')].value.type` (`= 'flow'` for flow history, `<> 'flow'` for script history) — mirroring the projection in `windmill-api/src/jobs.rs`. Confirmed on prod: every flow-path singlestepflow wraps `type='flow'`, every script-path one wraps `type='script'` (no cross-contamination), and these wrappers carry a NULL `runnable_id` so `ScriptHash` history is structurally unaffected (left untouched).
- Regression test `test_input_history_singlestepflow_flow_vs_script`: inserts a flow-wrapped and a script-wrapped singlestepflow at the **same path**, and asserts each history side returns its own and excludes the other (all four directions).

### Notes
- `FlowNode` (flow-as-step) is intentionally not added: those are child jobs excluded by the existing `parent_job IS NULL` filter and don't carry the parent flow's `runnable_path`.
- `is_preview` stays correct: `singlestepflow` is not in the preview set, so scheduled runs get no false "preview" badge.
- **No performance impact:** `kind` is a post-index filter (scan uses `ix_job_root_job_index_by_path_2` on `workspace_id, runnable_path, created_at DESC`); the added `EXISTS` subplan only touches the ~40 already-index-narrowed rows. Measured ~0.5ms on prod.

## Test plan
- [x] `cargo test -p windmill-api-integration-tests --test inputs` passes
- [x] Test fails without each part of the fix — verified by reverting the kind-list change (returns `[]`) and the wrapped-type filter (cross-kind run leaks in)
- [ ] Manually: open a flow (and a script) whose schedule uses a dynamic-skip handler, let it run, open the "History" sidebar, confirm the scheduled run appears and matches `/runs`

Fixes WIN-2244

### DIFF
--- a/backend/windmill-api-inputs/src/lib.rs
+++ b/backend/windmill-api-inputs/src/lib.rs
@@ -190,8 +190,16 @@ async fn get_input_history(
         kind @ JobKind::Script if g.include_preview.unwrap_or(false) => {
             vec![kind, JobKind::Preview]
         }
-        kind @ JobKind::Flow if g.include_preview.unwrap_or(false) => {
-            vec![kind, JobKind::FlowPreview]
+        // A scheduled/triggered flow is wrapped in a synthetic SingleStepFlow when the
+        // schedule has a dynamic-skip handler (or a scheduled script has native retry),
+        // so its runs land under `singlestepflow`, not `flow` (see schedule.rs). Include
+        // it here or those runs never surface in the run-history sidebar.
+        JobKind::Flow => {
+            let mut kinds = vec![JobKind::Flow, JobKind::SingleStepFlow];
+            if g.include_preview.unwrap_or(false) {
+                kinds.push(JobKind::FlowPreview);
+            }
+            kinds
         }
         kind => vec![kind],
     };
--- a/backend/windmill-api-inputs/src/lib.rs
+++ b/backend/windmill-api-inputs/src/lib.rs
@@ -73,14 +73,6 @@ impl Display for RunnableType {
 }
 
 impl RunnableType {
-    fn job_kind(&self) -> JobKind {
-        match self {
-            RunnableType::ScriptHash => JobKind::Script,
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ f9806a47d1b3ca9487d6237afb63dd1dfd9f912f

PR #10314, opened 2026-07-24
fix: WAC wait_for_approval reads its own approval result, not the first

rubenfiszel:
## Summary

`prepare_checkpoint_for_resume` (`wac_executor.rs`) injected the resume result for a WAC approval step by reading the **oldest** `resume_job` row for the parent job:

```sql
SELECT value, approver, approved FROM resume_job WHERE job = $1 ORDER BY created_at ASC LIMIT 1
```

`resume_job` rows are never deleted, so a WAC workflow with several sequential `wait_for_approval()` calls accumulates one row per approval, and every step after the first re-read the **first** approval's result. All steps showed `approved: true` even when the 2nd was cancelled and the 3rd timed out.

Fixes WIN-2241.

## Why not filter by `resume_id`

The obvious fix (and the approach taken in #10311 / WIN-2240) is to filter `AND resume_id = hash(step_key)`, since the inline resume/cancel URLs embed `hash(key)` as the `resume_id`. **This is not channel-agnostic and silently breaks the common approval paths.** Only the emailed inline URL (`/api/w/.../jobs_u/resume/{job}/{resume_id}/{sig}`) carries `hash(key)`. Every other channel stores a **random** `resume_id`:

- **In-run "Approve/Deny" button** (`FlowStatusWaitingForEvents.svelte`) → `resumeSuspended` → `rand::random()`
- **Approval page** (`/approve/{w}/{job}?token=…`) → `resumeSuspended` → `rand::random()`
- **Slack** (`slack_approvals.rs`) / **Teams** (`teams_approvals_ee.rs`) → `rand::random::<u32>()`
- **Resume as owner** (`resume_suspended_flow_as_owner`) → `resume_id = 0`

`resume_suspended` explicitly handles WAC jobs (`is_wac`), so these are all live WAC approval channels. Filtering by `resume_id = hash(key)` finds no matching row for them and falls to the `else` branch — returning `approved: false` even for a **legitimate** approval. That is a worse regression than the original stale-read bug.

## The fix

Bound the lookup to `resume_job` rows created at/after the approval step's `started_at`. That timestamp is written for every approval step by `handle_wac_v2_output` (shared by the bun and python executors) at the moment it suspends. The current step's approval row is always created after its `started_at`; rows consumed by earlier steps were created before it. This identifies the correct row regardless of which channel submitted the approval. A timed-out step matches no row and still falls to the `else` branch returning `{approved: false}`.

```sql
SELECT value, approver, approved FROM resume_job
WHERE job = $1 AND created_at >= (
    SELECT (workflow_as_code_status->('_step/' || $2)->>'started_at')::timestamptz
    FROM v2_job_status WHERE id = $1
)
ORDER BY created_at ASC LIMIT 1
```

## Changes

- `backend/windmill-worker/src/wac_executor.rs`: filter the approval `resume_job` lookup by the step's `started_at` instead of taking the oldest row for the job.

## Validation

- Backend recompiles cleanly under `cargo watch` (worker + api), health checks green.
- Query uses the runtime `sqlx::query_as::<_, …>` form (not the `query!` macro), so no `.sqlx` offline cache update is required.
- **DB-level proof** of the three approaches across a 3-approval scenario (A approved, B cancelled, C timed out), run against the real Postgres query planner with the interactive channels' random `resume_id`:

  | step | OLD (no filter) | `resume_id = hash(key)` (#10311) | NEW (`created_at >= started_at`) |
  |------|-----------------|----------------------------------|----------------------------------|
  | A (approved)  | alice / **true** | **approved:false** ❌ | alice / **true** ✅ |
  | B (cancelled) | alice / **true** ❌ | approved:false | bob / **false** ✅ |
  | C (timed out) | alice / **true** ❌ | approved:false | timeout / false ✅ |

## Relationship to #10311 (WIN-2240)

#10311 addresses the same root cause with the `resume_id = hash(key)` filter. Per the table above, that approach regresses all non-URL approval channels to `approved:false`. This PR should supersede it (or #10311 should adopt the `created_at`-bounded lookup). Consolidation under a single issue number is left to the reviewers.

## Test plan

- [ ] WAC workflow with 3 sequential `wait_for_approval()` calls: approve #1, cancel #2, let #3 time out — assert each step's result reflects its own outcome.
- [ ] Approve via the in-run "Approve" button (random `resume_id`) and confirm the step returns `approved: true` (this is the case the `resume_id` filter breaks).

---
## Summary by cubic
Fixes WIN-2241 by making WAC `wait_for_approval()` read its own approval result instead of the first. We now track and skip previously consumed `resume_job` rows so each step gets the right outcome across all channels.

- **Bug Fixes**
  - Added `consumed_resume_row_ids` to `WacCheckpoint` and updated `prepare_checkpoint_for_resume` to exclude them when selecting from `resume_job`, then mark the matched row as consumed.
  - Channel-agnostic: does not rely on `resume_id`, so in-run button, approval page, Slack/Teams, and resume-as-owner approvals work.
  - Added regression test (`backend/tests/wac_approval_resume.rs`) covering approve → cancel → timeout; timeouts return `approved: false` and earlier approvals are not re-used.

<sup>Written for commit efbd6be699a4ac012f2948a5c26f3669770449b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/wac_approval_resume.rs
+++ b/backend/tests/wac_approval_resume.rs
@@ -1,0 +1,83 @@
+//! WIN-2241: each sequential `wait_for_approval()` in a WAC workflow must
+//! resolve to its own approval, not the workflow's first. Guards approved,
+//! cancelled and timed-out steps sharing one parent job.
+
+use serde_json::{json, Value};
+use sqlx::{Pool, Postgres};
+use uuid::Uuid;
+use windmill_common::wac::{WacCheckpoint, WacPendingSteps};
+use windmill_worker::wac_executor::prepare_checkpoint_for_resume;
+
+async fn insert_resume_row(
+    db: &Pool<Postgres>,
+    job_id: Uuid,
+    resume_id: i32,
+    approver: &str,
+    approved: bool,
+    value: Value,
+) -> anyhow::Result<()> {
+    // resume_id is deliberately arbitrary (a stand-in for the random ids the
+    // interactive approval channels store) — the fix must not depend on it.
+    sqlx::query(
+        "INSERT INTO resume_job (id, job, flow, value, approver, resume_id, approved) \
+         VALUES ($1, $2, $2, $3, $4, $5, $6)",
+    )
+    .bind(Uuid::new_v4())
+    .bind(job_id)
+    .bind(value)
+    .bind(approver)
+    .bind(resume_id)
+    .bind(approved)
+    .execute(db)
+    .await?;
+    Ok(())
+}
+
+fn pending_approval(mut prior: WacCheckpoint, key: &str) -> WacCheckpoint {
+    prior.pending_steps = Some(WacPendingSteps {
+        mode: "approval".to_string(),
+        keys: vec![key.to_string()],
+        job_ids: Default::default(),
+    });
+    prior
+}
+
+#[sqlx::test]
+async fn wac_sequential_approvals_read_own_row(db: Pool<Postgres>) -> anyhow::Result<()> {
+    let job_id = Uuid::new_v4();
+    // FK target for resume_job.flow and v2_job_status.id (written by save_checkpoint).
+    sqlx::query(
+        "INSERT INTO v2_job_queue (id, workspace_id, scheduled_for) VALUES ($1, 'test-workspace', now())",
+    )
+    .bind(job_id)
+    .execute(&db)
+    .await?;
+
+    // Step A: approved.
+    insert_resume_row(&db, job_id, 1, "alice", true, json!({"n": 1})).await?;
+    let ckpt = pending_approval(WacCheckpoint::default(), "apprA");
+    let ckpt = prepare_checkpoint_for_resume(&db, &job_id, ckpt).await?;
+    assert_eq!(ckpt.completed_steps["apprA"]["approved"], json!(true));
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ efbd6be699a4ac012f2948a5c26f3669770449b6 — nit-only verdicts at fc5a1085a9677e7d35e2b34ebc86efee82bb0755 (codex: mergeable-with-nits, claude: good to merge, pi: good to merge); both Codex P2 nits addressed in efbd6be699a4ac012f2948a5c26f3669770449b6 (renamed `consumed_resume_ids`→`consumed_resume_row_ids`; trimmed the narrative test header). No P0/P1 findings. check_oss + check_oss_full green (earlier check_oss_full failure was a transient rusty_v8 download 504, re-run passed).

PR #10315, opened 2026-07-24
fix(frontend): pin sveltekit version.name so builds are reproducible across architectures

rubenfiszel:
## Summary

In a mixed-architecture cluster, HTML served by an `amd64` pod references `_app/immutable/chunks/<hash>.js` files that do not exist on `arm64` pods (and vice versa), so bundle requests 404.

The cause is **not** architecture-specific code generation. SvelteKit defaults `kit.version.name` to `Date.now().toString()`. That string is embedded in a client chunk — and in the `__sveltekit_<hash>` global derived from it — so it changes that chunk's content hash, which cascades into new filenames for about a quarter of `_app/immutable`. Since the assets are baked into the binary via `rust_embed`, the two architecture images of the same release ship different asset names. The two depot builders simply start a few hundred milliseconds apart.

Fixes WIN-2242

## Evidence

Diffing `/static_frontend/_app/immutable` between both architecture variants of the published `windmill:1.770.0`:

| | |
|---|---|
| files per variant | 863 |
| filenames that differ | ~224 |
| chunks byte-identical | 637 / 860 |
| chunks byte-identical **after normalizing chunk-name references** | **854 / 855** |

Exactly one chunk genuinely differs, and this is its entire diff:

```
amd64: globalThis.__sveltekit_h4zbwy … pe=`1784913453235`
arm64: globalThis.__sveltekit_1jgnowm … pe=`1784913453653`
```

418 ms apart. The bundler (Vite 8 / rolldown) is deterministic across `arm64` and `amd64`; two builds on the *same* machine diverged identically (448-line diff, same magnitude as the cross-arch diff).

## Changes

- `frontend/svelte.config.js`: `version: { name: process.env.WM_BUILD_VERSION || pkg.version }`.
- `Dockerfile`: `ARG WM_BUILD_VERSION`, declared immediately before `RUN npm run build`.
- `docker-image.yml` (×3, including the debuginfo extraction), `docker-image-rpi4.yml`, `build-publish-rh-image.yml`, `build-publish-rh8-image.yml` (×2): pass `WM_BUILD_VERSION=${{ github.sha }}`.

The value has to satisfy two constraints at once: **identical across the per-architecture builds of one commit** (or the images disagree on asset filenames), and **different between commits** (SvelteKit only full-page reloads on a failed chunk import when the deployed `_app/version.json` differs from the baked-in version — `runtime/client/client.js:1186`; equal values would strand an open tab on an error page after a redeploy). The commit sha satisfies both. `pkg.version` remains the fallback so unwired and local builds are still architecture-consistent.

## Tradeoff

Because the arg feeds the bundle step, `RUN npm run build` now re-runs on backend-only commits (~1-3 min per build job on native builders); `npm ci` and every earlier layer stay cached. Scoping the value to the frontend input paths would preserve that cache, but any input missed from the list would silently reintroduce the stranding bug, so the commit sha is the safer default.

## Considered and rejected

Pinning the Docker frontend stage to `--platform=linux/amd64` (this PR's original approach, reverted). It fixes the symptom, but depot builds each platform on its own native machine, so the arm64 builder runs the stage under QEMU. Measured on this PR's own run [30115908242](https://github.com/windmill-labs/windmill/actions/runs/30115908242): `npm run build` took **1946s (32.4 min)** emulated, and the `build` job went from ~13 min to 50 min. It also leaves the build non-reproducible and does nothing for the redeploy-recovery case.

## Test plan

- [x] Two consecutive builds at the same `WM_BUILD_VERSION` produce byte-identical output (864 files, same names and hashes).
- [x] Different `WM_BUILD_VERSION` values produce different manifests, and `_app/version.json` carries the passed value.
- [x] Control: with the config change reverted, two builds differ by 448 lines — matching the cross-arch diff on the shipped image.
- [x] `ARG WM_BUILD_VERSION` reaches the build environment (same mechanism `VITE_BASE_URL` already relies on); all four workflow files parse as valid YAML.
- [x] App boots and logs in normally with the pinned version (dev server, Playwright).
- [ ] End-to-end: once this PR's multi-platform image builds, diff `_app/immutable` between both architecture variants and confirm the filename lists match.

## Screenshots

No visible UI change — this is build configuration. Included only to show the app boots normally with the pinned version:

app-boot

### DIFF
--- a/.github/workflows/build-publish-rh-image.yml
+++ b/.github/workflows/build-publish-rh-image.yml
@@ -63,6 +63,7 @@ jobs:
           push: true
           build-args: |
             features=ee_rhel
+            WM_BUILD_VERSION=${{ github.sha }}
           secrets: |
             rh_username=${{ secrets.RH_USERNAME }}
             rh_password=${{ secrets.RH_PASSWORD }}
--- a/.github/workflows/build-publish-rh8-image.yml
+++ b/.github/workflows/build-publish-rh8-image.yml
@@ -65,6 +65,7 @@ jobs:
           push: true
           build-args: |
             features=ee_rhel
+            WM_BUILD_VERSION=${{ github.sha }}
           secrets: |
             rh_username=${{ secrets.RH_USERNAME }}
             rh_password=${{ secrets.RH_PASSWORD }}
@@ -82,6 +83,7 @@ jobs:
           push: true
           build-args: |
             features=ee_rhel
+            WM_BUILD_VERSION=${{ github.sha }}
           secrets: |
             rh_username=${{ secrets.RH_USERNAME }}
             rh_password=${{ secrets.RH_PASSWORD }}
--- a/.github/workflows/docker-image-rpi4.yml
+++ b/.github/workflows/docker-image-rpi4.yml
@@ -68,6 +68,7 @@ jobs:
           push: true
           build-args: |
             features=ce_rpi
+            WM_BUILD_VERSION=${{ github.sha }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:dev
             ${{ steps.meta-public.outputs.tags }}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -93,6 +93,7 @@ jobs:
           push: true
           build-args: |
             features=ce
+            WM_BUILD_VERSION=${{ github.sha }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.DEV_SHA }}
             ${{ steps.meta-public.outputs.tags }}
@@ -155,6 +156,7 @@ jobs:
           push: true
           build-args: |
             features=ee
+            WM_BUILD_VERSION=${{ github.sha }}
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-ee:${{ env.DEV_SHA }}
             ${{ steps.meta-ee-public.outputs.tags }}
@@ -254,6 +256,7 @@ jobs:
           target: debuginfo
           build-args: |
             features=ee
+            WM_BUILD_VERSION=${{ github.sha }}
           outputs: type=local,dest=./debuginfo
 
       - name: Rename debug file with corresponding architecture
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,6 +79,11 @@ COPY /python-client/docs/ /frontend/static/pydocs/
 RUN npm run generate-backend-client
 ENV NODE_OPTIONS "--max-old-space-size=8192"
 ARG VITE_BASE_URL ""
+# Pass the commit sha: it must be identical across the per-architecture builds of one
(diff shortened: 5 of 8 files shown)


rubenfiszel:
✅ Review round clean @ 8b422f6027144abf2b518ee59bca82576b7820e6 — nit-only verdicts at a5ba0cbe5409676c4cd45c909f2b77e71c5722f9 (codex P2 on the comment wording; claude and pi good to merge); nit addressed in 8b422f6027

rubenfiszel:
Fixed in dcbd9b089c. Confirmed the mechanism in the installed SvelteKit 2.53.4: `runtime/client/client.js:1186` calls `stores.updated.check()` when a node import fails and only calls `native_navigation()` when the fetched `_app/version.json` differs from the baked-in version, so equal versions across distinct manifests do suppress the recovery.

`WM_BUILD_VERSION=${{ github.sha }}` is now passed from every workflow that builds the root Dockerfile (`docker-image.yml` ×3 including the debuginfo extraction, `docker-image-rpi4.yml`, `build-publish-rh-image.yml`, `build-publish-rh8-image.yml` ×2 — that last one builds amd64 and arm64 as separate invocations, so it needed this most). `pkg.version` stays as the fallback so unwired/local builds are still architecture-consistent.

Verified locally: two builds at `WM_BUILD_VERSION=sha-aaa` are byte-identical, `sha-aaa` vs `sha-bbb` produce different manifests, and `_app/version.json` carries the passed value.

Known tradeoff: the arg is declared immediately before `RUN npm run build`, so `npm ci` and every earlier layer stay cached, but the bundle step itself now re-runs on backend-only commits (~1-3 min per build job on native builders). Scoping the value to the frontend input paths would keep that cache, but any missed input would silently reintroduce this exact bug, so the commit sha is the safer default.

rubenfiszel:
Both fixed in 89e99f7354.

**P1 (RHEL):** correct, and the blast-radius description is right too. `docker/RHEL{8,9}/Dockerfile` now declare `ARG WM_BUILD_VERSION=""` immediately before `RUN npm run build`. Swept all Dockerfiles in the repo: exactly three run `npm run build` (root, RHEL8, RHEL9) and all three now declare the arg, so there is no fourth path left unwired.

**P2 (ARG form):** adopted. One correction on the reasoning: the space form does receive a `--build-arg` — I tested `ARG WM_BUILD_VERSION ""` against `--build-arg WM_BUILD_VERSION=…` through buildx before the first push and the value arrived. Re-verified through the full Docker path on the `=` form, with node reading it rather than the shell: `node -e 'console.log(process.env.WM_BUILD_VERSION)'` inside a `node:24-alpine` stage prints the passed sha. The `=` form is still the better choice for consistency with `ARG features=""`.

`docker buildx build --check` on all three Dockerfiles reports no new warnings.

rubenfiszel:
✅ Review round clean @ 44c61a8c74e53fcee0b5bc9a7794f148bfd16cd4 — nit-only verdicts at 89e99f7354a15bf2403e90dca7e5322688420985 (codex: Good to merge; pi: Good to merge; claude: nit on the duplicated Dockerfile comment); nit addressed in 44c61a8c74.

PR #10317, opened 2026-07-24
feat: bind WAC approval urls to a named wait_for_approval step

rubenfiszel:
## Summary

Follow-up to WIN-2241 / #10314 (which made the approval **read** side channel-agnostic). This is the **write** side: a first-class way for a Workflow-as-Code developer to obtain the resume / cancel / approval-page URLs bound to one specific `wait_for_approval()` call, so an approval request can be routed through a custom channel (own email, own Slack/Teams message) that resumes exactly that step.

Until now the only SDK helper was `getResumeUrls()` / `get_resume_urls()`, which signs an **arbitrary random nonce**: the URLs it returns are not tied to any particular approval step, and they change on every call. The URLs the built-in approve/reject buttons use are computed by the worker at suspend time from `hash(step_key)` and were not reachable from the SDK.

### Design: option B (SDK helper + backend signer), not option A

Option A — having `wait_for_approval()` return its own URLs — does not work: the call suspends the workflow by throwing, so anything it "returns" only materializes *after* the approval has already been granted. Too late to notify anyone.

So: name the approval step, and ask the backend for that step's URLs before suspending.

```ts
const urls = await step("urls", () => getApprovalUrls("manager"));
await step("notify", () => sendEmail(urls.resume, urls.cancel));
const { value, approver } = await waitForApproval({ key: "manager", timeout: 3600 });
```

```python
urls = await step("urls", lambda: get_approval_urls("manager"))
await step("notify", lambda: send_email(urls["resume"], urls["cancel"]))
result = await wait_for_approval(key="manager", timeout=3600)
```

The SDK cannot derive the `resume_id` itself (it is a Rust-side hash, and duplicating it in two SDKs invites version skew), so the derivation stays in exactly one place — `windmill_common::wac::approval_resume_id` — shared by the worker that mints the inline buttons and the new endpoint that signs the ahead-of-time URLs.

### Authority is unchanged

`GET /w/{w}/jobs/wac_approval_urls/{job_id}/{step_key}` runs the same `resume_target_flow_path` + `check_scopes(jobs:run:flows:<path>)` gate as `get_resume_urls` and then delegates to the same `get_resume_urls_internal`; only the `resume_id` derivation differs. A caller that can reach this endpoint could already mint an equivalent signature via `resume_urls/{job_id}/{any_u32}`. `verify_suspended_secret` is untouched and still validates the URLs this path produces.

### Hash function

`approval_resume_id` uses SHA-256 truncated to `u32` rather than the `DefaultHasher` the worker used inline. The two producers now live in different processes, and std explicitly does not guarantee `DefaultHasher` stability across releases. Nothing persists a resume_id (it is carried in the URL and covered by the HMAC), and post-#10314 approval matching does not key on it, so switching is transparent — in-flight suspended jobs keep working with URLs already minted.

### Binding is enforced, not just derived

Approval rows are consumed oldest-unconsumed-first regardless of `resume_id` (WIN-2241 — required so Slack/Teams/the approval page, which sign random ids, keep working). So deriving the id from the step key is necessary but not sufficient: a workflow that mints URLs for several approvals up front would let whoever clicks first resolve the *earliest pending* step with their answer — WIN-2241's bug through a new door.

Enforced in two layers, for two different reasons.

**Consumption is the guarantee.** `prepare_checkpoint_for_resume` skips rows carrying another *minted* key's `resume_id`. This is race-free by construction — the consumer knows exactly which step it is resolving — and it is the only place that can be, since the API check cannot be atomic with the insert (a link can pass validation and then have its step time out before the row lands). Only ids of keys this workflow minted a URL for are excluded; every other `resume_id` stays eligible, so WIN-2241 holds unchanged for Slack, Teams, the approval page and resume-as-owner, which all sign random ids.

**The API check is the good error message.** `get_wac_approval_urls` records the step key in `workflow_as_code_status._minted_approval_keys` (upsert — a workflow can mint before any step has checkpointed), and the resume route rejects a resume whose `resume_id` matches a minted key while a *different* step is awaiting approval. Unbound resume_ids take exactly the path they do today. A link is valid **only** while its own step is awaiting approval. Accepting it at any other moment (including while the workflow is still running toward that step) banks a row that the next approval to be reached consumes, whichever step that is — the same misroute, deferred to the suspend timeout. Links can still be sent ahead of time; approvers simply cannot act before the workflow reaches the step.

Minting is restricted to WAC-shaped jobs in the caller's workspace: `v2_job_status` is keyed by job id alone, and `!!workflow_as_code_status` is what the run page keys its WAC timeline off, so an unscoped write could stamp another workspace's job or render an empty WAC timeline on a plain script. `getApprovalUrls()` also resolves the job id up to the WAC parent, so calling it from inside a `task()` body (where `WM_JOB_ID` is the child) still targets the workflow that will suspend.

Minting also refuses a step key whose `resume_id` collides with one already minted on the same job. A 32-bit id is not injective (`approval-12509` and `approval-81661` really do collide), and while that surface is unchanged from before this PR — the route parameter is a `u32`, and the `DefaultHasher` it replaces was masked to 32 bits — the binding check newly *depends* on the id identifying one key, so an ambiguous pair is refused at the point it would be created.

Note `resume` and `cancel` are step-bound; `approvalPage` is not — it opens the job's approval page, which acts on whichever approval is pending. Documented in both SDKs.

### Step keys no longer alias

`_allocKey` suffixed without checking whether the result was already handed out, so a second `step("dup")` and an explicit approval key `dup_2` both claimed `dup_2`. Both SDKs now track allocated keys and bump past collisions; allocation order is fixed by the workflow body, so replays reproduce identical keys. This also closes the same latent aliasing between two tasks/steps, which predates this PR.

### Duplicate keys fail loudly

`_allocKey` auto-numbers repeated names, so a mistyped/copy-pasted `waitForApproval({ key: "manager" })` twice would produce steps `manager` and `manager_2`. The workflow itself stays correct, but the URL the caller minted with `getApprovalUrls("manager")` addresses the *first* step: hitting it for the second approval collides on `resume_job.id = job_id ^ resume_id` and is rejected with `resume request already sent`, parking the workflow until timeout. Same trap for a named approval inside a loop, and for a key already taken by a task or `step()`.

Since an explicit key is an identifier callers mint URLs against, reusing one now throws (`WAC step key "manager" is already used in this workflow`). Unnamed approvals keep auto-numbering (`approval`, `approval_2`, …) unchanged, so this is additive — `key` did not exist before this PR.

## Changes

- `windmill-common/src/wac.rs`: new shared `approval_resume_id(step_key) -> u32` (SHA-256 truncated); `bun_executor.rs` now calls it instead of inlining a `DefaultHasher`.
- `windmill-api/src/jobs.rs`: new `GET /w/{w}/jobs/wac_approval_urls/{job_id}/{step_key}` → `get_wac_approval_urls`, same scope check as `get_resume_urls`, delegating to `get_resume_urls_internal`.
- `openapi.yaml`: `getWacApprovalUrls`.
- TypeScript SDK: `waitForApproval({ key })` and new `getApprovalUrls(stepKey, approver?)`; exported from `build.sh`.
- Python SDK: `wait_for_approval(key=...)`, `Windmill.get_approval_urls()` and module-level `get_approval_urls(step_key, approver)`.
- Duplicate explicit step keys now raise in both SDKs instead of being silently renamed to `<key>_2` (see below).
- `system_prompts/base/workflow-as-code.md` Approvals section rewritten around the bound URLs (+ regenerated `system_prompts/auto-generated/` and `cli/src/guidance/skills.gen.ts`).
- New integration test `backend/tests/wac_approval_urls.rs`.

## Test plan

- [x] `cargo check -p windmill-api -p windmill-worker -p windmill-common`
- [x] `cargo test -p windmill-common --lib wac::` — golden values pin `approval_resume_id` as a cross-process contract (worker and API can run different builds during a rolling deploy).
- [x] `cargo test --test wac_approval_urls` — 3 cases; a third covers the mint guards (blank key 400, cross-workspace 404, colliding key 400 naming the other key). (1) mints URLs for step `manager` against a real API server + Postgres, asserts they carry `approval_resume_id("manager")`, that a signature minted for one step is rejected when retargeted at another (`Invalid signature`), that the genuine URL resumes with no credential, and that the resulting `resume_job` row carries the posted value. (2) mints `legal` + `finance` up front, suspends on `legal`, asserts `finance`'s link is refused with 400 both while nothing is pending and while `legal` is pending, and that `legal`'s own link still resumes.
- [x] `pytest python-client/wmill/tests/test_workflow.py` — 4 new cases pin the explicit key, the duplicate-key error, an explicit key aliasing a suffixed step key, and unnamed auto-numbering (1 pre-existing unrelated failure: `test_preserves_function_name`, `@task` does not use `functools.wraps`).
- [x] `cargo test --test wac_approval_resume` — 2 cases: WIN-2241 read semantics unchanged, plus a new one pinning that consumption skips another step's bound row while still taking an unbound random-id row.
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api -p windmill-worker` — cache regenerated with all 747 EE entries preserved.
- [x] `typescript-client/build.sh` + `tsc --noEmit`: `JobService.getWacApprovalUrls` resolves, no new errors (19 pre-existing errors in this checkout from missing `@types/node` / lib target).
- [x] `system_prompts/check-freshness.sh` clean.
- [x] frontend `npm run check:fast` — only the pre-existing unrelated `modern-screenshot` module error.
- [x] **Full runtime WAC loop, driven by the real SDKs** (bun worker + API + Postgres). A preview WAC job with two sequential approvals:
  - the **real TypeScript `getApprovalUrls()`** (imported from the built client, not a copy) returns a URL **byte-identical** to the inline one the worker wrote at suspend — `resume_id` 2649449912 = the golden `approval_resume_id("approval")` — which is what actually validates the generated `JobService.getWacApprovalUrls` param mapping;
  - the **real Python `get_approval_urls()`** likewise matches byte-for-byte (after the fix below);
  - a link minted for the *other* step, POSTed while the first is pending → `400 … not currently awaiting approval`; once that step is pending, the same link is accepted;
  - resume through the TS-minted URL and **cancel** through the Python-minted URL → `{first: {via: typescript-sdk, approved: true}, second: {via: python-sdk, approved: false}}`, each approval carrying its own payload *and its own outcome*.
- [x] **Real `WorkflowCtx` key semantics** driven directly against the built TS module (the committed TS suite re-declares its own copy, so it cannot cover these): named key verbatim, duplicate rejected, unnamed auto-numbering, empty rejected, `toString`/`constructor` treated as ordinary keys, and an explicit key aliasing a suffixed step key rejected.
- [x] Python and bun WAC share one minting path — `python_executor.rs` calls `bun_executor::handle_wac_v2_output` — so the bun run covers both.

**A bug this found:** `httpx` encodes `params={"approver": None}` as `approver=`, so Python'''s URLs carried a trailing `?approver=` that TypeScript'''s did not. The signature still verified, but an unauthenticated resume recorded the approver as `""` instead of `"anonymous"`. Fixed by omitting the param when unset; both SDKs and the worker now agree byte-for-byte.

**Not covered, stated explicitly:** a worker suspending on a *custom-named* key (the worker resolves `windmill-client` from npm, which predates this PR — mitigated: the worker hashes whatever string it receives, proven for two distinct keys, and both SDKs''' key allocation is verified directly); the consumption-side filter firing live (the insert guard prevents creating a foreign bound row — covered at DB level); and the two TOCTOU windows, which need failpoints.

No frontend source changes (the regenerated `frontend/src/lib/gen` is gitignored), so no screenshots.

---
## Summary by cubic
Adds named, step‑bound approval URLs for Workflow‑as‑Code so workflows can notify approvers with links that resume exactly the intended `wait_for_approval` step. Enforces binding end‑to‑end and hardens key/URL validity while keeping Slack/Teams/approval page behavior unchanged (WIN‑2241).

- **New Features**
  - GET `/w/{workspace}/jobs/wac_approval_urls/{job_id}/{step_key}` returns `approvalPage`, `resume`, `cancel` for a specific `wait_for_approval(key=...)` (optional `approver`; same scopes as `get_resume_urls`).
  - Shared `windmill-common::wac::approval_resume_id(step_key) -> u32` (SHA‑256→`u32`) used by `windmill-worker` and the endpoint so ahead‑of‑time links match the inline buttons.
  - SDKs: TypeScript `getApprovalUrls(stepKey, approver?)`, `waitForApproval({ key })`; Python `get_approval_urls(step_key, approver=None)`, `wait_for_approval(key=...)`. OpenAPI and docs/prompts updated.

- **Bug Fixes**
  - Enforce binding: the resume route validates within the insert transaction and rejects links for steps not currently awaiting approval; worker consumption skips rows bound to other minted keys while leaving random‑ID channels eligible (WIN‑2241 intact).
  - Key safety: reject empty, duplicate, or invalid step keys (slashes or dot segments); detect and refuse `resume_id` collisions; minted‑key writes scoped to the job’s workspace.
  - Runtime hardening: deterministic TS key allocation with null‑prototype counters; atomic suspend counter decrement with the resume insert; Python omits a blank `approver` query param.

<sup>Written for commit 143ebd00e07d7778b9bc86a0d5dc111e15a18f75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10317?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
On the suggestion to add TypeScript allocator/key tests in `typescript-client/tests/workflow.test.ts`: declining, with reasoning.

That file is an inlined **copy** of the SDK (its header says "mirrors client.ts implementation") — it re-declares `WorkflowCtx`, `StepSuspend` etc. rather than importing from `client.ts`, because `client.ts` imports `./services.gen` which only exists after `build.sh` runs. A test added there would exercise a duplicate implementation I'd also be writing, and would keep passing if the shipped `client.ts` regressed. Per AGENTS.md ("a committed test must pin behavior a future change could plausibly break"), that's ceremonial.

The semantics under discussion are shared by both SDKs and are pinned against the **real** module in `python-client/wmill/tests/test_workflow.py` (`TestApprovalKeys`): explicit key used verbatim, duplicate key raises, explicit key aliasing a suffixed step key raises, empty key raises, unnamed approvals still auto-number. The TS guards are line-for-line equivalents.

Happy to revisit if the TS suite is ever wired to import the real `client.ts`.

rubenfiszel:
On the regression test for the stale-suspend-snapshot P1 (fixed in 202b347): **the fix ships, a test for it does not**, and I want to be explicit about why rather than leave "added a regression test" standing.

I wrote the test that was asked for — snapshot reads 0, workflow parks on the named approval, assert the resume drives `suspend` back to 0. It passed. Then I reverted the fix and **it still passed**, so it was pinning nothing: the test flips `suspend` to 1 *before* the request, so `get_flow_info_for_resume` already reads 1 and the old `if flow_info.suspend > 0` gate decrements too.

Reproducing the real interleave needs `suspend` to change *between* that read and the decrement, inside a single request. That is not reachable from an integration test without a failpoint (the repo has a `failpoints` feature, but wiring one into the resume path for this is more machinery than the one-line fix warrants). Per AGENTS.md — a committed test must pin behavior a future change could plausibly break — I deleted it rather than ship a green test that guards nothing.

The fix is verifiable by reading: the decrement condition is now evaluated atomically in SQL (`WHERE id = $1 AND suspend > 0`) inside the transaction that already holds the queue-row `FOR UPDATE` lock, so there is no snapshot left to go stale. The other 5 backend tests confirm no regression.

Separately, the **full runtime loop is now exercised** (see the updated test plan): a real bun worker suspends on a named approval, the API-minted URL comes back byte-identical to the worker's inline one (`resume_id` 2649449912 = the golden `approval_resume_id("approval")`), the other step's link is refused with 400, and the correct link resumes it — each of the two approvals ending up with its own approver's payload.

rubenfiszel:
Addressed the round-9 finding — you were both right, and following it through changed the outcome more than expected.

**The flakiness was real.** `wac_resume_decrements_a_suspend_set_after_the_snapshot` set `suspend = 1` *after* releasing the lock, so it raced the request's own decrement: if the request won, it decremented a zero, my write then set 1, and the test went red with no code change. It passed 8/8 for me only because a blocked request takes longer to wake than an immediate `UPDATE` — timing, not construction.

**The obvious fix made it useless.** Moving the write inside the lock-holding transaction makes it deterministic (6/6 green) but drops discrimination to **0/4** — because `get_flow_info_for_resume` takes `FOR UPDATE OF q` itself, so the request blocks *at the snapshot read*. Once the update commits with the lock release, the request reads `suspend = 1` and there is no stale snapshot left to detect. Deterministic and discriminating are mutually exclusive here without pausing execution between the read and the write — i.e. a failpoint.

So I deleted it (5692cb7), same as the binding-under-lock test earlier in this round for the same underlying reason. **Both race fixes stay in the code**; what is missing is automated proof of them, and I would rather say that than ship two green tests that assert nothing.

What remains committed is `wac_concurrent_colliding_mints_admit_exactly_one`, which I measured in both directions: **5/6 red** with the collision fix reverted, **8/8 green** with it. It is a concurrency *invariant* rather than an interleave, which is why it works without a failpoint.

Separately, the consumption-side filter — previously covered only by a direct-function test — is now verified end to end against a live worker. Staging the rolling-deploy case (an older API inserts a row bound to a later step), `approval` correctly skipped it and resolved `{approved: false}`, and `approval_2` then consumed it (`approver: bob`). The row is reserved for its own step, not lost.

If you would rather have the two interleave tests than not, say so and I will wire a failpoint into the resume path — it is the only way to get them, and it means adding the `failpoints` feature to `windmill-api`.

rubenfiszel:
Fixed the dot-segment nit rather than dismissing it — it is the same class of mismatch this PR exists to remove: `waitForApproval({key: ".."})` was accepted while `getApprovalUrls("..")` could never address it, since the key travels as one path segment.

Both SDKs now share one guard and reject the same set — `""`, whitespace, `.`, `..`, and anything containing `/` or `\` — with the same message, applied at all four entry points (`waitForApproval`/`wait_for_approval`, `getApprovalUrls`/`get_approval_urls`). Verified against the real built modules, not the inlined test copy: all six bad keys rejected on both sides, and `manager` still suspends normally, so ordinary keys are unaffected. Python coverage is a parametrized case over `["", "  ", ".", "..", "a/b"]`.

Validation at this head: 6 backend integration tests, 92 Python SDK tests (1 pre-existing unrelated failure — `@task` does not use `functools.wraps`), system-prompt freshness clean, TS typecheck at its pre-existing baseline.

rubenfiszel:
✅ Review round clean @ 143ebd00e07d7778b9bc86a0d5dc111e15a18f75 — nit-only verdicts at 5692cb7428ef99238f484a2942976d9badc74bca (claude: good to merge, pi: good to merge, codex: mergeable/nits); nit addressed in 143ebd00e07d7778b9bc86a0d5dc111e15a18f75

PR #10318, opened 2026-07-24
feat: record and replay raw app sessions step by step

rubenfiszel:
## Summary

Fixes WIN-2243. A raw app cannot demo itself the way a script or flow does: there is no run to replay. This adds a **session recorder** and a **step-by-step player**. Recording captures the app's DOM plus every user interaction with it; replaying walks the session one interaction at a time — the DOM the user acted on (target highlighted), then the DOM the app settled into.

Steps are the user's interactions: filling a field, picking a select option, toggling a checkbox or radio, sweeping a range/date/colour, clicking, submitting, pressing Enter/Escape, and app reloads. A step settles only once the DOM is quiet **and** any runnable job the interaction kicked off has answered, so a backend-driven result is part of the recording rather than a `running…` spinner.

The recording JSON is **standalone**: each frame is a complete HTML document with the stylesheets inlined as rendered and form state frozen into attributes. Nothing is stored on Windmill — the publisher downloads the JSON and hosts it wherever they like.

**Try it:** [one-click replay of a recorded session](https://ruben-win-2243-build-a-recor.windmill.pages.dev/replay?src=https://raw.githubusercontent.com/windmill-labs/agent-screenshots-internal/main/recordings/win-2243/ticket-demo-session.json) on the Cloudflare preview (no login).

player

## Where it shows up

- **Raw app editor → `…` → Record demo** — full-screen drawer: drive the app, stop, preview the replay, download the JSON. The drawer explains how to share it (the public `/replay` page, `?src=` autoload, embedding, hosting the JSON on S3/GitHub/anywhere public).
- **Publish flow (`/folders` → Publish to Hub) → step 2** — `Record demo` per raw app, saved against the Hub item so the Hub page can replay it. Consuming half: windmill-labs/windmillhub#160.
- **`/replay`** — public, no login, chrome-less and embeddable. Opens a local file or autoloads `?src=<url>`. Non-app recordings (script/flow/pipeline) are directed to `/pipeline_replay` in-workspace.

## `Share as iframe` is removed

#10310 already moved publishing to the folders page and removed the per-item `Publish to Hub` entries; this PR removes the publish flow's `Share as iframe` alongside it. It made an app anonymously executable and wired its public URL into the Hub's `external_embed_url`, which the recorded session replaces — and `external_embed_url` was only ever read for raw apps (low-code apps render natively on the Hub from their JSON).

Low-code apps lose nothing by it: the app editor's Deploy panel is the canonical control for the same decision — the login/no-login toggle, the secret URL, and the same iframe embed snippet — and `AppRow` still offers "Go to public page". That path is also the safer one: it is gated on `canSetAnonymous` and explains when a workspace protection rule forbids anonymous apps, whereas the publish-flow shortcut called `updateApp` with `execution_mode: 'anonymous'` directly and consulted no ruleset. Removing it closes that bypass and stops a Hub-publishing wizard from making an app public as a side effect.

record drawer

## Changes

- **`recording/rawAppSnapshot.ts`** — serializes a live (same-origin) app document into standalone HTML: live form state frozen into attributes, stylesheets inlined *as rendered* (linked sheets, CSS-in-JS, CSSOM mutations; media conditions preserved, `url()` re-resolved against the sheet), document scroll carried over, scripts/handlers/templates/meta-refresh dropped, and the interaction target stamped for the player's highlight. Also prepares a frame for replay: policy first in `<head>`, nothing that can execute, fetch or navigate.
- **`recording/rawAppRecording.svelte.ts`** — recorder store bound to the app iframe. One step per interaction, each with the DOM it acted on and the DOM once mutations settled. Pre-interaction frames come from `pointerdown` / `keydown` / `beforeinput` (so paste, cut, undo and pickers are covered), live until the step that uses them, and only a browser auto-repeat or a continuous drag folds into an existing step. Settling watches the runnable bridge so in-flight jobs are awaited (60s cap), and Stop drains a job the last step is still running rather than shipping its spinner as the outcome. Frames are content-deduplicated and only stored once a step claims one; the recording is capped (500 steps / 40M chars / bounded step text).
- **`recording/RawAppRecordingReplay.svelte`** — player: a timeline of clickable checkpoints placed at the time each interaction happened, a collapsible step list (closed by default, so the app gets nearly full width), the snapshot rendered at the recorded viewport size (scaled to fit), an Interaction/Result toggle, prev/next, and Play through to the end paced by the recorded gaps.
- **`recording/rawAppRecordingLoad.ts`** — `isAppRecording` (structure, cardinality, size) and a capped streaming `fetchRecording`, shared by both replay surfaces and by the Hub.
- **`workspaceSettings/RawAppRecordSession.svelte`** — the record surface used by both the editor menu and the publish flow. It renders through the same path the viewer does, so an app the publisher opted into sandbox isolation stays isolated and is refused with an explanation rather than recorded.
- **`workspaceSettings/deployToHubItems.ts`** — the publish flow's item shape and its pure predicates, split out of the session store so they can be tested without dragging in the API client and the editor bundle.
- **`hub_publish.rs`** — `POST /w/{workspace}/hub/raw_apps/{id}/recording`, the proxy the publish flow saves through (three-line mirror of the script/flow recording routes), plus its openapi entry.
- **Package exports** — `components/RawAppRecordingReplay.svelte`, `components/recording/rawAppRecordingLoad`, `components/recording/rawAppSnapshot`, so the Hub can mount the player. The player's import graph is closed (no `$app/*`, no stores, no API client).

## Privacy

Password values are masked at a fixed width (the length of a secret is information too). An app author can mark any element `data-wm-no-record`: its content is dropped from every snapshot and its step metadata, and it keeps only the attributes needed to hold its place in the layout — an allow-list, so an attribute nobody has thought of yet cannot leak. `class` and `id` are on that list by name only: their values are author-chosen free text (`id="salary-92000"`), so they are filtered to the tokens the snapshot's own surviving CSS selects on, which keeps the placeholder's shape without carrying the app's data. A marked stylesheet is not part of that vocabulary — honouring its selectors would launder the tokens it was marked to withhold. Redaction also covers a marked `<option>` picked in an unmarked `<select>`, a marked stylesheet, and a marked document root.

## Replaying untrusted markup

A recording can arrive from any `?src=` URL, so a snapshot renders in an empty-sandbox iframe (no scripting, opaque origin) carrying a `default-src 'none'` CSP injected structurally first via a real parser, with scripts, inline handlers, navigation attributes (any namespace), SMIL animation, templates and resource hints removed, and clicks disabled at the root. The cost is that images and fonts the recorder could not inline do not render on replay.

## Test plan

- [x] 20 unit tests (`vitest`), including a jsdom project for the DOM-bound utilities: no-record content/attributes/template/root/stylesheet redaction, password masking, target stamping past removed nodes, stylesheets snapshotted as rendered, and a replay frame that keeps the CSP first while retaining nothing executable, fetchable or navigable
- [x] Recorded end to end in the browser: fill / select / label-driven toggle / redacted field / clicks / app reload / backend runnable, replayed with timeline, highlight, Interaction/Result and autoplay
- [x] Verified across two full iframe reloads, keyboard-only interaction (Tab + typing, arrows on ranges/radios, Enter/Space activation), paste after a debounce, native picker sessions, typing interleaved with control changes, auto-repeat bursts, and label-wrapped buttons
- [x] Verified crafted hostile recordings: probe images blocked by CSP, viewport CSS-injection rejected at load, and SVG/`xlink`/SMIL/shadow-template navigation stripped
- [x] `data-wm-no-record` fields: typed secrets appear nowhere in the JSON
- [x] `/replay` confirmed to need no login (logged out, cleared storage, reloaded — the recording still played)
- [x] Bridge delivery verified against a live bundle iframe: the request is delivered only to the page window, the response only to the bundle window — so the watcher listens on both
- [x] `POST /hub/raw_apps/{id}/recording` reaches body validation instead of 404ing, while a bogus sibling path still 404s
- [x] Record → Stop → Record again in the editor drawer: the second run captures against the remounted preview
- [x] `policy.sandbox = true` on a raw app: the record drawer mounts no bundle at all and explains why; clearing it restores recording
- [x] A marked element's unstyled `class`/`id` tokens (`customer-acme-secret`, `salary-92000`) appear nowhere in the snapshot, while a token the CSS selects on survives
- [x] A marked stylesheet selecting a marked element's class: the token appears nowhere in the snapshot
- [x] `/replay?src=` pointed at a payload with a 5 MB `type`: the page shows the short generic error instead of echoing it
- [x] Clicked a 6s runnable and pressed Stop while it was still showing `running…`: Stop reported "Waiting for the job…", took 6.3s, and the step's Result frame holds the job's response (`result: {"ok":42}`) rather than the spinner

## Follow-ups

- The public `/replay` currently plays app recordings only; extending it to script/flow/pipeline recordings is in flight on a separate branch.
- Sandbox-isolated apps can't be recorded (their DOM is deliberately unreachable from the page).
- Scroll inside nested `overflow: auto` containers is not restored (no static-CSS equivalent, and replays are deliberately script-less); document-level scroll is.
- Images and fonts the recorder cannot inline do not render on replay, by CSP.


rubenfiszel:
### Try it on the Cloudflare preview

**One-click replay of a real recorded session** (recorded on this preview, logged in as `user@windmill.dev`):

https://ruben-win-2243-build-a-recor.windmill.pages.dev/pipeline_replay?workspace=user_admin&src=https://raw.githubusercontent.com/windmill-labs/agent-screenshots-internal/main/recordings/win-2243/ticket-demo-session.json

11 steps: fill, select, label-click toggle, a redacted field, two clicks, an app reload, then fill + click again. Use the step list, the Interaction/Result toggle, or Play.

preview-step6

**Record your own on the preview:**
1. Open https://ruben-win-2243-build-a-recor.windmill.pages.dev/apps_raw/get/u/user/win2243_recorder_demo?workspace=user_admin (any raw app works, as long as it is not sandbox-isolated)
2. Click **Record** (bottom-left), interact with the app, then **Stop & download**
3. Open https://ruben-win-2243-build-a-recor.windmill.pages.dev/pipeline_replay and drop the downloaded JSON in

The demo app (`u/user/win2243_recorder_demo` in `user_admin`) has an `data-wm-no-record` API-token field and a redacted button, so the recording shows the masking: the typed token appears nowhere in the JSON and the step reads `Filled input[text] (redacted)`.

rubenfiszel:
### Player: play-through fix + checkpoint timeline

Two follow-ups from trying it on the preview:

**Play stopped after one step.** Regression from moving the Interaction/Result switch to `ToggleButtonGroup`: the group also dispatches `selected` when playback sets the phase itself, and the handler treated every dispatch as a user click and paused. It now ignores a dispatch for the phase it is already on, so Play runs from the current position to the last step and stops there (pressing Play again restarts from the initial state).

**Timeline.** The steps are now also a timeline above the viewer: one checkpoint per interaction, positioned at the time it happened (with a minimum spacing so bursts stay clickable), a progress line up to the current step, and a tooltip with the step label and timestamp. Clicking a checkpoint jumps there. The step list auto-scrolls to the playing step.

timeline-playing

At the end of playback (last checkpoint, list scrolled, second ticket visible in the Result frame):

timeline-end

The demo link from the previous comment picks this up once the preview redeploys — the recording format is unchanged.

rubenfiszel:
### Round 4 addressed in `1ef68e4`

| Finding | Fix | Verified |
|---|---|---|
| P1 SVG `xlink:href` bypassed navigation stripping | navigation attributes are matched on `localName`, so namespaced ones go too; resource-hint `<link>`s dropped as well | crafted recording renders with no `xlink:href`, no `href`, no `preload` |
| P1 unvalidated `viewport` interpolated into `style` | loader rejects non-finite / non-positive / >20000 sizes, and the player clamps independently (it also renders recordings passed directly) | crafted `width: "1px; position:fixed; inset:0"` recording is refused with "Invalid app recording"; the player never mounts |
| P1 `commitFill` discarded the next control's pre-frame | the spent pointer frame is only cleared when it belongs to the field being filled | type-then-toggle now yields `fill` then `toggle`, with the toggle's Interaction showing `unchecked` and Result `checked` |
| P1 unbounded `frames` array | capped at `2 * MAX_RECORDED_STEPS + 1`, which is the most a legitimate recording can hold | |
| P2 `maskValue` leaked the secret's length (pi) | fixed-width mask | |
| P2 `style` in `REDACTED_ATTRS` contradicted the comment (claude) | `style` is stripped only when it embeds a `url()`; comment says why | |
| P2 `<link>` resource hints left live (claude) | removed in the sanitizer | |

**Not fixed — nested/horizontal container scroll (Codex P1).** Horizontal *document* scroll is now carried (`margin-left`), but scroll inside `overflow: auto` containers is not, and I don't think it should block: a replayed frame runs with scripting disabled and an opaque origin by design, and there is no static-CSS equivalent of `scrollTop` on a container. The alternatives all cost more than the defect: wrapping a container's children in a shifted div distorts flex/grid layout and any `>` selector; transforming each child breaks mixed text content; `autofocus`-into-view is disabled by `sandbox=""` (automatic features are gated on `allow-scripts`). The limitation is stated in the code comment next to the document-scroll shift and in the PR's follow-ups. If it turns out to matter in practice, the honest fix is a scroll-offset field per frame plus a player that sizes the iframe to the full document height and scrolls its own pane — a format change worth doing on evidence, not speculatively.

rubenfiszel:
✅ Review round clean @ 579ff5aa42881418a8ef5ee32edcb2dc19f48a44 — nit-only verdicts at 7d2e4f7362e148deba8b8666ed9e419bee57cf8b (pi: good to merge; codex and claude: mergeable, nits); nits addressed in 579ff5aa42 (Escape on controls is recorded again, disabled stylesheets are dropped rather than revived on replay, and `app_path`/`workspace` are bounded at load like the step text). `npm run check` clean, 12 tests green, and the session re-verified end to end in the browser.

rubenfiszel:
### Step list collapsed by default (`41c0b37551`)

The replay now opens with the app at nearly full width; the step list is behind a toggle in the header (leftmost button). Navigation without it: the timeline checkpoints, prev/next, and the step bar at the bottom.

Closed (default) — the recorded app renders at ~1:1:

panel-closed

Open:

panel-open

### Is the recording standalone?

Yes — `frames[]` holds complete HTML documents, not just interactions. Each frame has the app's CSS inlined (including CSSOM-only rules), live form state frozen into attributes, and no scripts; `steps[]` only references frames by index. The 9-step demo recording is 39 KB total, ~2.7 KB per frame.

Two caveats worth knowing:
- Images and fonts referenced by remote URL are not inlined, and the replay CSP blocks them — text, layout and CSS replay faithfully, remote images do not render.
- The CSS is inlined into *every* frame, so an app with a large framework stylesheet produces much bigger recordings (identical frames dedupe, but distinct ones each carry a copy). That is what the 40 MB cap is for; a shared stylesheet pool across frames would be the obvious optimisation if it becomes a problem.

rubenfiszel:
### Round addressed in `43cda6df7b`

- **P1 `<noscript>` leaked redacted content** — with scripting on its markup is a single text node, so the redaction pass cannot see into it while `outerHTML` writes it back out (and it would *render* on a script-less replay). `<noscript>` is now dropped like `<template>`, with a test.
- **Implicit Enter submission recorded three events** — the key press, the synthetic click the browser dispatches on the default submitter, and the submit. The submitter click is now folded like the submit already was, so one action is one step.
- **Independent key presses coalesced** — `kind === 'key'` merged any two key steps within the window, so Enter-then-Escape became one step labelled Escape. Only browser auto-repeats fold now, and a repeat that ends up starting its own step still captures a pre-frame.
- **frame-0 guard was inverted** (claude) — `about:blank` reports `readyState === 'complete'`, which is precisely the state the guard existed to defer; it now waits for actual body content.
- **`selected` restored for a redacted `<option>`** (claude) — dropping it made the snapshot claim the user picked the *first* option. A marked option's text is the answer and stays scrubbed; `selected` only says they picked something. `checked` stays dropped for a marked checkbox, where the boolean *is* the answer — the two differ, and the code says so.

### Not taken

**Filtering synthetic events by `isTrusted`** (cubic). Component libraries legitimately re-dispatch clicks (`el.click()` from a keyboard handler, custom widgets forwarding activation), and every one of those is a real user interaction; dropping them would silently lose steps in exactly the apps most likely to be recorded. The step cap and frame budget already bound what a noisy app can do to a recording.

**Stylesheet probe sampling** (cubic). Only *linked* sheets are cached at all, and reading every rule to invalidate would cost what the cache exists to save. A mutation to an unsampled rule of a linked sheet can be missed; that trade-off is stated in the code.

rubenfiszel:
✅ Review round clean @ f53777442e83e3bf65a88595e3c7c25391372dac — nit-only verdicts at e5c2641477af26d06a4f52e09b0cdf7a7bbb0c81 (codex: mergeable-with-nits, pi: good to merge); the nit (no regression coverage for `isAppRecording`, the gate for untrusted `?src=` payloads) is addressed in f53777442e with a table-driven test over a valid recording plus 14 boundary rejections.

Note for anyone reading the round history: the Claude reviewer failed to start ("encountered an error after 1s") on the last four rounds, so those rounds carry Codex and Pi verdicts only. Its findings from rounds 24-26 were all addressed.

rubenfiszel:
Re the latest Codex P1 (*`data-wm-no-record` preserves arbitrary allow-listed attribute values*) — **dismissing this one deliberately**, and the reasoning is worth recording since a variant of it has now come up twice.

The finding is technically accurate. I verified it: a redacted element keeps `hidden="salary-92000"`, `colspan="case-7"`, `width="secret-w"` verbatim, and `type="case-secret"` reaches the step label as `input[case-secret] (redacted)`.

It is not worth acting on. Reaching it requires an app author to write secrets into `hidden=` or `colspan=` — attributes whose values HTML gives no meaning to — on their own app, and then voluntarily record and publish it. A recording here is opt-in, author-driven and author-published; the threat model that makes attribute-value scrubbing worthwhile (untrusted markup captured without the author's involvement) is not this one. The previous round's `data-wm-no-record="…"` case was taken because the marker was the single attribute exempted from the allow-list and its value is inert downstream, so normalizing cost one line; that argument does not extend to attributes the replay actually renders from.

What I did fix from this round is the separate cubic P2, which is a real defect rather than a hypothetical: `styledTokens` did not decode CSS escapes, so a redacted element carrying utility classes (`md:flex` written `.md\\:flex`, `w-1/2` written `.w-1\\/2`) lost **all** of its layout styling on replay — the class attribute was dropped entirely. Fixed in 9abcb1c, with a test that fails without it.

Happy to revisit the attribute-value scrubbing if someone wants the stricter posture; it would be ~10 lines (empty the boolean attributes, clamp `type` to known input types).

PR #10321, opened 2026-07-25
fix: add missing is_operator guards to all workspace-object write handlers

rubenfiszel:
## Summary

The operator role is documented as read-only + execute-only (the `operator_settings` flags are all "whether operators can **view** X"), but only the script / flow / app / raw-app handlers enforced that on writes. RLS does not carry `is_operator` — only `is_admin` reaches `set_session_context` — and `check_scopes` is a no-op for a normal session token, so an operator could create and modify groups, folders, variables, resources, resource types, schedules and triggers through direct API calls.

Verified against a real operator on a dev instance before the fix: `groups/create`, `folders/create`, `resources/type/create`, `resources/type/update`, `schedules/create`, `schedules/update`, `http_triggers/create` (workspaced), `http_triggers/update`, `http_triggers/delete` all returned 2xx, and — the worst of them — `variables/delete g/all/*` and `resources/update_value g/all/*` succeeded, i.e. workspace-wide credential deletion and poisoning.

Fixes WIN-2245

## Changes

Added `OptJobAuthed::reject_operator_write(what)` in `windmill-api-auth` and applied it at the top of every affected write handler, before `check_scopes` / `check_deploy_rules` / `require_admin` / `user_db.begin`:

- **groups**: `create`, `update`, `delete`, `adduser`, `removeuser`
- **folders**: `create`, `update`, `delete`, `addowner`, `removeowner`
- **granular ACLs**: `acls/add`, `acls/remove` — the generic form of `folders/addowner`, writing `extra_perms` on every ACLable kind
- **variables**: `create`, `update`, `delete`, `delete_bulk`
- **resources**: `create`, `update`, `update_value`, `delete`, `delete_bulk`, and resource types `create` / `update` / `delete`
- **schedules**: `create`, `update`, `delete`, `setenabled`
- **triggers** (generic `TriggerCrud` router, all kinds): `create`, `update`, `delete`, `setmode`; plus `http_triggers/create_many` and native-trigger `create` / `update` / `delete`
- **trigger infrastructure**: `capture/set_config` (provisions external config and starts listeners), `postgres_triggers/publication/{create,update,delete}`, `postgres_triggers/slot/{create,delete}`

**Job tokens are exempt.** A job's token carries the operator's `is_operator`, and `wmill.setState` / `setVariable` / `setResource` go through `resources/create`, `resources/update_value`, `variables/create` and `variables/update` — a bare guard broke every operator-launched run that touched its own state (confirmed e2e: `401 Operators cannot create resources`). `reject_operator_write` lets a request through when `job_id` is set, which only happens on the token minted for job execution. What a job writes is decided by the script's author, not by the operator who ran it, and operators cannot author scripts or run previews.

## Scope boundaries

- Instance-group handlers and `schedules/setdefaulthandler` were left alone — they already call `require_super_admin`. `variables/encrypt` is not a persistence write.
- `setenabled` (schedules) and `setmode` (triggers) were not in the issue's list but are guarded: both rewrite who the object runs as (`email` / `permissioned_as`), a strictly larger capability than the update route next to them.
- **Not guarded, deliberately:** the capture session lifecycle (`capture/ping_config`, `capture/{id}` delete, `capture/move`) — all RLS-scoped to rows an operator can no longer create — and the EE `gcp_triggers/subscription/delete`, which today has *no* owner check at all. That is a pre-existing authz gap on an EE-only file and wants its own issue plus a companion PR rather than being bundled here.
- The operator-visible pages (`operator_settings.{schedules,variables,resources,folders,groups,triggers}`) still render create/edit/delete controls, which will now surface a 401 toast. Hiding them is a separate frontend follow-up; the settings are documented as view grants, so the backend behavior here is the intended one.

## Test plan

- [x] `backend/tests/operator_write_guards.rs` — table-driven over all 28 always-compiled guarded routes: operator gets 401 + `Operators cannot …`, a regular member never hits the guard, and an operator **job** token can still write its own state. Verified the test fails on the pre-fix code (`POST groups/create` returns 200).
- [x] `cargo check --features http_trigger,websocket,native_trigger,postgres_trigger` clean; `rustfmt --check` clean on all touched files.
- [x] Manual e2e on a dev instance with a real operator user, including a build with the trigger features: every route above 401s post-fix (`http_triggers/{create,update,delete,setmode,create_many}`, `acls/{add,remove}` on folder and group kinds, `capture/set_config`, `postgres_triggers/{slot,publication}/create`), an admin and a plain non-operator member are unaffected on all of them, and a deployed script calling `wmill.setState()` run by the operator succeeds again.

### DIFF
--- a/backend/tests/fixtures/operator_write_guards.sql
+++ b/backend/tests/fixtures/operator_write_guards.sql
@@ -1,0 +1,14 @@
+-- Fixture for the operator write-guard regression test (WIN-2245).
+-- Layered on top of `base` (which provides test-workspace and the non-operator
+-- `test-user-2`/SECRET_TOKEN_2). Adds an Operator member so we can assert that
+-- Operators cannot reach the group / folder / variable / resource / schedule
+-- write handlers.
+
+INSERT INTO password(email, password_hash, login_type, super_admin, verified, name)
+    VALUES ('operator@windmill.dev', 'not-a-real-hash', 'password', false, true, 'Operator User');
+
+INSERT INTO usr(workspace_id, email, username, is_admin, operator, role) VALUES
+	('test-workspace', 'operator@windmill.dev', 'operator-user', false, true, 'Operator');
+
+INSERT INTO token(token_hash, token_prefix, token, email, label, super_admin) VALUES
+	(encode(sha256('OPERATOR_TOKEN'::bytea), 'hex'), 'OPERATOR_T', 'OPERATOR_TOKEN', 'operator@windmill.dev', 'operator token', false);
--- a/backend/tests/fixtures/operator_write_guards.sql
+++ b/backend/tests/fixtures/operator_write_guards.sql
@@ -1,8 +1,5 @@
--- Fixture for the operator write-guard regression test (WIN-2245).
--- Layered on top of `base` (which provides test-workspace and the non-operator
--- `test-user-2`/SECRET_TOKEN_2). Adds an Operator member so we can assert that
--- Operators cannot reach the group / folder / variable / resource / schedule
--- write handlers.
+-- Adds an Operator member on top of `base` (which provides test-workspace and
+-- the non-operator `test-user-2`/SECRET_TOKEN_2).
 
 INSERT INTO password(email, password_hash, login_type, super_admin, verified, name)
     VALUES ('operator@windmill.dev', 'not-a-real-hash', 'password', false, true, 'Operator User');
--- a/backend/tests/operator_write_guards.rs
+++ b/backend/tests/operator_write_guards.rs
@@ -1,16 +1,8 @@
-//! Operators are a read-only + execute-only role, but only the script / flow /
-//! app handlers used to enforce that on writes. RLS does not carry
-//! `is_operator` (only `is_admin` reaches `set_session_context`) and
-//! `check_scopes` is a no-op for a normal session, so an Operator could create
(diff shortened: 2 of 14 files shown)


rubenfiszel:
Closing: the premise of WIN-2245 is inverted.

Every `is_operator` guard that exists on `main` is about **authoring or executing code** — `create_script`, `archive_script`, `create_flow`/`update_flow`, `create_app`/`update_app` (+ raw apps), `run_preview`, `run_inline_preview`, app previews, dependency jobs. Not one of them guards a data object.

The frontend says the same thing explicitly: `ItemsList.svelte` folds `&& !$userStore?.operator` into `canWrite` for scripts / flows / apps only. On the variables, resources, schedules, folders and groups pages the operator check gates **page visibility** (`operator_settings.<page>`) and every write control is gated on the plain `canWrite` permission check, with no operator exclusion.

So operators are ordinary workspace members for variables, resources, schedules and triggers, bounded by RLS and `extra_perms` like anyone else — the missing guards this PR added are absent by design, and adding them breaks intended behavior. The `g/all/*` "credential poisoning" framing in the issue isn't operator-specific either: `g/all` grants write to every member of group `all`, so the guard would not have addressed it.

One finding from the review rounds does stand on its own, independent of the operator question: **`gcp_triggers/subscription/delete` (EE, `windmill-trigger-gcp/src/handler_ee.rs`) has no owner check at all** — any workspace member can delete an external GCP subscription. Worth its own issue.

PR #10326, opened 2026-07-25
feat: run dbt projects as a first-class Windmill runtime

rubenfiszel:
## Summary

Run an unmodified dbt project as a first-class Windmill script, with dbt models becoming assets in the existing asset graph. Design doc committed as `docs/dbt-runtime.md` (first commit), sibling to `docs/pipelines-vs-dbt.md`: this is the adoption ramp, that is the long game.

Phases 1 to 3 of the spec. Phase 4 (`--defer`/`state:modified`, backfill integration, `wmill dbt import`) is out of scope.

**One dbt project is one Windmill script.** The script's content is a YAML descriptor; the project's own files ride with it as the script's module bundle, which the worker materialises into the job directory before invoking dbt. Nothing is cloned, so there is no repository resource, no ref, no commit and no clone cache.

```yaml
engine: dbt-core-1x            # or dbt-core-2x | fusion
profile:
  resource: $res:f/prod/warehouse
  target: prod
select: ["tag:nightly+"]
test_behavior: build
```

Importing an existing project is a copy, never a transformation. The bundle is written to disk verbatim, with the extensions dbt expects, so `wmill sync pull` produces a tree dbt itself can run with `--project-dir`:

```
f/analytics/
├── analytics.dbt.yaml              the descriptor (the script's content)
└── analytics__dbt/                 the module bundle: the project, unmodified
    ├── dbt_project.yml
    ├── models/staging/stg_orders.sql
    ├── models/marts/_marts__models.yml
    ├── macros/cents_to_dollars.sql
    ├── seeds/country_codes.csv
    └── snapshots/orders_snapshot.sql
```

```
cp -r my-dbt-project/. f/analytics/analytics__dbt/
wmill sync push
```

A team whose repository must stay canonical keeps it: git-sync points at that repository and pushes it into the workspace. What that buys over pointing at a repo directly is that the version of the project is the version of the script, so a deploy is atomic, a rollback is a redeploy, and the graph the deploy parsed is exactly what a run builds. Windmill is the runner and the viewer; editing stays in the local CLI-and-editor loop, which is where dbt development actually happens.

## The one architectural risk, resolved

The spec flagged this as the assumption most likely to reshape the phasing: every other language's asset parsing in `parse_assets_for_lang` is a pure function of script content, and dbt's needs the project on disk plus a dbt invocation.

Prototyped first, and it works. `parse_assets_for_lang` returns `None` for dbt; the **deploy-time dependency job**, which already runs on a worker with the engine available, materialises the bundle, runs `dbt deps` + `dbt parse`, ingests `manifest.json`, and writes the `asset` rows itself. Verified end to end before anything was built on top:

```
Relocking script
Reusing cached dbt_packages
09:31:26  Running with dbt=1.12.0
09:31:26  Registered adapter: postgres=1.11.0
Ingested 16 dbt nodes and 15 edges into the asset graph
```

A useful consequence: redeploy *is* the graph-refresh mechanism, with no new concept (decision 12). The only case that cannot be settled at deploy is a descriptor that is dynamic by construction (a `{{ }}` placeholder in `vars`, or a `$var:` value in `env`), and those re-ingest from each run's own manifest.

## Changes

**Phase 1 — run it**
- `ScriptLang::Dbt` across the language sites, both OpenAPI specs, the CLI-facing enums and the editor (init descriptor + dbt icon).
- `windmill-parser-yaml::dbt` — descriptor parse and the run signature (`select`, `exclude`, `vars`, `full_refresh`, plus one argument per `{{ placeholder }}` the descriptor interpolates).
- `dbt_executor.rs`: materialise the bundle, `dbt deps`, render `profiles.yml`, one `dbt build` per job, structured result, `dbt retry`.
- `dbt_engine.rs`: provisioning for all three engines, worker-global cache.
- CLI: `<script>__dbt/` as a dbt project's module folder, taken verbatim (a `.sql` model has no inferable dialect, and `.yml`/`.csv` are part of the project); the descriptor and lock stay beside the folder so it holds nothing but the project; directories dbt generates (`target-path`, `packages-install-path`, `clean-targets`, read from `dbt_project.yml`) are excluded from the bundle, from the sync diff and from staleness hashing; a module-only edit pushes its parent script.

**Phase 2 — graph**
- Migration: `AssetKind::Table` (`table://`) and the `dbt_node` / `dbt_edge` sidecar.
- `dbt_manifest.rs` — manifest → nodes, edges and asset usages. Models/snapshots/seeds are `w`, sources are `r`; internal `ref()`s are lineage inside the project, not usages of the script.
- `DbtDependencyLocks` in the lockfile: manifest digest, engine + version, and the profile's resolved relation root.
- A `project_digest` (sorted SHA256 over the bundle) stands where the commit used to: it keys the `dbt_packages` cache, which a `local:` dependency's content reaches no other way, and gates `dbt retry`, so an edited project cannot resume the previous attempt's failures.
- Graph API + canvas: per-model provenance on the asset nodes, model count on the script node.

**Phase 3 — live progress and ergonomics**
- The JSON event file is tailed during the run and each model's status recorded via `record_materialization` — `MaterializationStatus::Running` gets its first writer.
- Settled from `run_results.json` at the end, which fills in row counts and covers the engines that emit no node events.
- Structured per-model result (status, timing, rows, failing-row counts), so partial failure is legible without reading the log.

## Two things verified empirically, as asked

**1. ADBC adapter coverage in dbt Core 2.0 for postgres.** Postgres is **not** in the default supported set. `dbt debug` on 2.0.0-alpha.5:

```
[error] [InvalidConfig (dbt1005)]: Configuration Error: The 'postgres' adapter is not yet
supported by dbt Fusion. Supported adapters: snowflake, bigquery, databricks, redshift,
duckdb, salesforce, clickhouse. To use an experimental adapter, set the environment
variable DBT_ALLOW_EXPERIMENTAL_ADAPTERS=true.
```

With that variable set it works completely — `debug`, `parse`, `ls` and `build` all pass against Postgres. So **dbt-core-2x parity testing is not blocked**, but it requires an opt-in env var, which is why the descriptor gained an `env:` map. The e2e sets it and the run reports `"engine": "dbt-core-2x", "version": "2.0.0-alpha.5", "totals": {"success": 9, "error": 0}`.

Also confirmed while there, and it is the concrete evidence behind decision 1: a valid dbt 1.x project **fails to parse** on 2.x until its deprecations are resolved (`MissingArgumentsPropertyInGenericTestDeprecation`, `loaded_at_field`/`freshness` moved under `config:`). Shipping `dbt-core-1x` as the default is what keeps "runs an existing project unchanged" true.

**2. The concept mapping, against real manifest output rather than the types.** Both hold, with two caveats worth knowing:

| dbt | Windmill | Confirmed |
|---|---|---|
| `unique` / `not_null` / `accepted_values` / `relationships` | the four `// data_test` kinds | ✅ exact, via `test_metadata.name`; kwargs carry `column_name`, `values`, `to`+`field` |
| `materialized: table` / `seed` | `replace` | ✅ |
| `materialized: incremental` + `unique_key` | `merge` | ✅ |
| `materialized: incremental`, no key | `append` | ✅ |
| `{% snapshot %}` | `scd2` | ✅ |
| `materialized: view` / `ephemeral` | — | ❌ **no analogue.** A view produces a relation but performs no write; an ephemeral model produces no relation at all. Left unmapped rather than invented. |

- **Severity casing is inconsistent between engines**: 1.x echoes the author's casing (`warn`), 2.x uppercases (`WARN`). Stored as dbt spells it, compared case-insensitively, lowercased before it reaches the UI.
- **Column-level lineage is not in `manifest.json`** for either core engine — the spec's decision 14 assumed it falls out of the manifest, and it does not. Declared per-column *metadata* is ingested; column→column edges would need Fusion or SQL-AST parsing. Called out rather than faked.

## Decision 11 (`table://`) and its two traps

Both handled explicitly and covered by tests, since a regression in either is invisible — both nodes still render, they just stop being the same node:

- **Identifier canonicalization.** `parse_asset_syntax` is the single point where an asset URI becomes a graph key, so the rule lives there: strip the quote characters (`"`, backtick, `[`/`]`) from the schema and name, then ASCII-lowercase them — matching the case-insensitive identifier comparison the DuckDB paths already use. Six spellings of one relation (hand-written, dbt's pre-quoted `relation_name`, Snowflake's upper fold, Postgres's lower fold, backticks, brackets) canonicalize to one key.
- **Warehouse identity is the Windmill resource path**, like `datatable://` and `ducklake://` — never host/account/database. Documented limitation: two resources pointing at one physical warehouse do not unify.

## Decision 21, updated

Now that BigQuery and Snowflake are CE (#10324), the dbt policy mirrors the remaining native boundary rather than inventing one: **CE for every adapter, EE only for `mssql` and `oracle`**, the two warehouse `ScriptLang`s still behind a license. The runtime, manifest ingest, asset graph and all UI are CE.

The mechanism differs from the native languages, which gate at compile time — there is one dbt executor and the adapter is only known once the profile resolves, so it is a runtime check at both deploy and run, naming the adapter ("Microsoft SQL server is only available with an enterprise license") rather than degrading into a confusing connection failure. The check is `cfg!(feature = "enterprise") && LICENSE_KEY_VALID`: the OSS variant of that atomic is initialized to `true`, so reading it alone would wave a CE build through.

`docs/dbt-runtime.md` is updated for this.

## Deviations from the spec, stated rather than substituted

1. **`dbt retry` is an explicit run argument** (`dbt_command: retry`), not the automatic behavior of Windmill's generic retry. Windmill's retry has no per-language hook to switch the invoked command, and each attempt gets a fresh job dir. The previous run's `target/` is cached per (workspace, script) on the worker and restored for a retry, so the mechanism works — it just has to be asked for.
2. **Live per-model progress is dbt-core-1x only today.** 2.0.0-alpha.5 accepts `--log-format-file json` but still writes a text log, so those engines get per-model state from `run_results.json` when the invocation ends rather than during it. A no-op reporter there, not a second format-sniffing code path.
3. **`select`/`exclude` scope the graph, resolved by dbt.** A script deployed with `select: ["tag:nightly"]` was registering as the producer of every model, so two scripts splitting one project each claimed all of it. Fixed by asking dbt (`dbt ls --output json`) rather than interpreting the selector string — the grammar is dbt's, and reimplementing it is exactly the divergence this feature exists to avoid.
4. **Only dbt-core 2.x is bundled** in the full images (one adapter-agnostic binary). dbt-core 1.x needs a venv per adapter and the image cannot know which one a project uses, so the worker builds and caches it on first use. **The Fusion engine is never bundled** — fetched from dbt Labs at runtime, which is what keeps Windmill out of the redistribution its license restricts.
5. **Interpolation uses `{{ arg }}`**, the syntax `interpolate_template` actually implements; the spec's example wrote `${flow_input.date}`.

## Screenshots

The dbt project's lineage on the asset graph: `raw_customers` / `raw_orders` sources, the dbt script node badged `×2` (models it materializes), blue write edges to `customers` (`table · 3T`) and `orders_daily` (`incremental · 4T`), and the lineage edge into a native bash script that reads `table://…/customers`. (That screenshot predates the removal of the cascade — such a script is now a reader in the graph, not a subscriber; `# on table://…` is refused at deploy.)

dbt-asset-graph3

Mid-run, **taken before the cascade was removed**: the dbt script shows a success badge and a native downstream script is running. A dbt run no longer triggers anything, so today only the dbt job would appear here. Kept because it is the clearest shot of the run-page graph mid-flight.

dbt-run-live

The run form of a dbt script. Its arguments are derived from the descriptor by the script-create endpoint (`select`, `exclude`, `vars`, `full_refresh`, `dbt_command`, each defaulting to the descriptor's own value) plus one required argument per `{{ placeholder }}` the descriptor's `vars` interpolate.

dbt-run-form-schema

The pipeline editor is a split view: the folder's asset graph on the left, and clicking a node opens it in the embedded editor on the right. On the dbt project node that is the descriptor, with Test, logs and result — editing the project's run configuration and seeing the models it drives happen on one screen.

The folder's graph, with two dbt projects in it: sources (`raw_customers`, `raw_orders`), models badged with their materialization (`view`, `incremental`, `snapshot`, `seed`, `table`), `ref()` lineage between models, and each project's script node badged with the models it materializes.

dbt-graph-two-projects

Selecting a model shows its SQL as written, with the path of the file inside the bundle (`f/cptest/complex__dbt/models/marts/finance/fct_orders.sql`) and `read-only · edit locally`.

dbt-model-bundle

The graph draws no edge between a dbt project and its relations: one edge per model grows with the project and buries the lineage that matters (`ref()` between models, and native consumers). The association is a badge instead — hovering a model's dbt chip lights up the project, hovering the project lights up every model it materializes, and clicking the chip selects the project. Selecting a model shows its own SQL, read-only, with the file's path in the bundle so the edit is one `cd` away.

dbt-model-sql

dbt-canvas-plus-yaml

Home lists a dbt script in its own right — it is the artifact you edit — next to the `Pipeline` entry of the folder it belongs to. Other pipeline-member scripts stay folded into that entry; a dbt script outside a folder has no entry to fold into and would otherwise vanish from Home entirely.

dbt-home-both

Created from scratch in the browser — pick dbt in the language list, the starter descriptor is seeded, Deploy, then Run:

dbt-browser-run

### Browser verification (Playwright, this branch's frontend against a local backend)

Live run: per-model state while the job runs — green checks on finished staging models, spinners on the marts still building, `fct_boom` already failed and `fct_after` skipped.

dbt-live

The same run, completed: `7 passed 1 warned 1 failed 1 skipped of 10 nodes`, each node with its kind, relation and dbt's own message, and a warn-severity test surfacing without failing the job.

dbt-completed

Selecting a model shows its SQL and fully-qualified relation; **Preview rows** dispatches `dbt show` for exactly that node — the selector it posts is `stg_orders,package:slow_shop`, the front-anchored-FQN-free spelling this PR moved to.

dbt-preview

A `dbt retry` run of an older version: its inputs are only `dbt_command: "retry"`, the result carries the restored `invocation_args` (`select: ["fct_boom+"]`), and the graph is that run's own pinned two nodes rather than today's deployed eight.

dbt-retry-pinned

### The editor and the settings, re-verified on `fe6233027b`

The project browses as one tree — the descriptor at the project root, `models/marts` and
`models/staging` nested under it, directories before files. Selecting a model opens it and
the run button becomes **Build stg_orders**, which posts `select: ["stg_orders,package:slow_shop"]`
and materializes that model alone (one progress row, where the whole fixture is nine models
of `pg_sleep`). `dbt_project.yml` is the one file with no delete button: without it the
worker refuses the version outright.

dbt-editor-tree

`+` takes a project-relative path, so a new file lands in the folder it names
(`models/staging/stg_probe.sql`), and deleting it puts the tree back. Behind the popover,
the result of the per-model build: `1 passed of 1 nodes`.

dbt-new-file

dbt is **not offered as a language** for a new script: it is moving to an editor of its own
in the follow-up below, and a project bundle in the generic script editor is a worse version
of what that will be. The list ends at DuckDB / Ruby / R. Existing dbt scripts still open,
run and deploy, and their picker still shows `dbt`; `localStorage.setItem('dbt_language', 'true')`
offers it again.

dbt-language-withheld

The workspace's warehouses: a name, a resource and a target, and the sentence that a
project reaches one by NAME and carries no connection of its own.

dbt-warehouse-settings

A name too long to key an asset path on is refused where an admin can see it, rather than
silently skipping every model at ingest:

dbt-name-too-long

Also exercised in the browser, without a screenshot each: the empty state for a project that brings its own `profiles.yml` (no warehouse identity to key models on), the deployed-graph fallback on a preview run, the asset drawer linking a model back to the script that writes it, and the two navigation races — switching runs in the list, and dispatching one model's preview then another's while the first is still in flight (the panel keeps the second, `3 rows · stg_payments`). Every dbt endpoint returned 200; the only 404s in the session are pre-existing `native_triggers/integrations/*` probes unrelated to dbt.

## Two bugs the modules path surfaced

**A script's modules did not survive the worker's file-system cache.** The first fetch of a script version reads the database and carries its modules; every later fetch imports from the worker's cache directory, whose `RawScript::import` hard-coded `modules: None` and whose `export` never wrote them. A worker restart therefore started running the script without its own files, silently. This predates dbt and affects any script with a module bundle, which ran with its helper modules missing; for dbt it is fatal and fails with "carries no project". `modules.json` is now written on every export and required on import, so an entry written by an older version fails to import and is refetched rather than serving a stripped script for as long as the directory lives.

**The run path derived its project digest from nothing.** `handle_dbt_job` passed `None` for the modules, so every dbt script in a workspace digested identically, which let `dbt retry` resume a different project's `run_results.json`. Fixed by threading the bundle through, and pinned by a test: after editing a model, a retry is now refused with "the last dbt run on this worker was a different project, warehouse or engine".

## Test plan

All e2e scenarios were exercised for real against the local Postgres, using two fixture dbt projects: a 3-file one, and a "complex" one with two `model-paths`, nested schema YAML, a macro, a CSV seed, a snapshot, a singular test, an analysis, a `dbt_utils` dependency, a custom `target-path` and a custom `packages-install-path`. Scenarios 1, 3, 4, 5, 7, 10 and 11 were re-run on the module-bundle path after the pivot; 2, 6, 8 and 9 were verified before it, on machinery the pivot does not touch (engine provisioning, the per-run graph refresh and profile rendering all sit downstream of how the project reaches the job directory). Actual output:

| # | Scenario | Result |
|---|---|---|
| 1 | Happy path | ✅ both fixtures `success`; the complex one `{"total": 10, "success": 9, "warn": 1, "error": 0}` across 2 views, a table, a seed, a snapshot, an incremental model and 4 tests |
| 2 | Engine parity | ✅ same duckdb project on all three: `dbt-core-1x` (1.12.0), `dbt-core-2x` (2.0.0-alpha.5) and Fusion (2.0.0-preview.202) — build, a failing model with its downstream skipped, and `dbt retry`. Both Rust engines refuse a postgres adapter, so parity needed a duckdb fixture |
| 3 | Test severity | ✅ failing `warn` test → job `success`, `{"warn": 1}`. Failing `error` test → job `failure`, naming the failing node |
| 4 | Retry | ✅ a run failing on `fct_orders` retried with `dbt retry`: `Done. PASS=0 ERROR=1 TOTAL=1`, the 5 successful nodes not rebuilt. After editing a model, the retry is refused: "the last dbt run on this worker was a different project, warehouse or engine" |
| 5 | Graph ingest | ✅ the complex fixture deploys to 13 `dbt_node` rows / 8 `dbt_edge` rows and 6 `w` + 2 `r` asset usages; the 4 generic tests land with their kind, column and severity (including a `warn` one) |
| 6 | ~~Cross-boundary cascade~~ | **Removed.** A dbt run does not trigger downstream runs — see "No cascade from dbt" in `docs/dbt-runtime.md`. Verified as a non-behaviour: a full 6-model run starts nothing, where it previously dispatched `f/cptest/downstream` |
| 7 | Selection | ✅ descriptor `select: ["tag:nightly"]` built exactly `customers` + `orders_daily`; run-arg `{"select":["stg_orders"]}` built exactly `stg_orders`, `{"total": 1}` |
| 8 | Dynamic descriptors | ✅ a `{{ }}` placeholder in `vars` re-ingests the graph from the run's own manifest, so a model that placeholder enables appears in the same run that builds it |
| 9 | Both credential paths | ✅ resource-rendered `profiles.yml`, and the project's own file with `{{ env_var() }}` injection (tables in `jaffle_own`) |
| 10 | Caching | ✅ with `packages-install-path: vendor`, the deploy publishes a cache entry containing `dbt_utils` and both runs log `Reusing cached dbt_packages` |
| 11 | Import round-trip | ✅ `cp -r` a project into `<script>__dbt/`, `wmill sync push`, then `wmill sync pull` into an empty directory: `diff -r` against the source is identical, a second push reports `0 changes to apply`, and a module-only edit is reported as `[changed modules: macros/cents_to_dollars.sql]` and pushes its parent script once |
| 12 | Agent worker | ✅ two runs on a `MODE=agent` worker against the pinned EE: the graph publishes through `/api/agent_workers/dbt_graph` job-keyed, terminal per-model state lands in `materialized_partition`, retry state is written agent-locally. `dbt_run_progress` stays empty for agents by design — an agent has no DB to tail into |
| 13 | Pinned dependency resolution | ✅ a project declaring `dbt_utils` by range pins its `package-lock.yml` digest at deploy; a run restores the tree from the lock-keyed cache and a `dbt_utils` macro compiles; a worker whose resolution differs is refused with "dependencies resolve differently from the version deployed" |

Real bugs surfaced by running it, each fixed and (where it could regress silently) pinned by a test: profile `port`/`threads` quoted as strings when dbt's schema demands integers; the dbt child inheriting the worker's stdio so `handle_child` hung; asset paths keyed on `config.schema` (the custom *suffix*) instead of dbt's resolved schema; `select` leaking into what the script claims to produce; the worker cache dropping a script's modules; the run path digesting an empty bundle; and one push deploying a script once per changed module file, which raced the graph publication so a changed test never reached the sidecar.

**Committed as regression guards** — the ones pinning behavior a future change could break silently:
- `asset_parser`: every spelling of a relation canonicalizes to one key; the resource-path prefix keeps its case.
- `dbt_manifest`: the four materializations → the four strategies (and `view`/`ephemeral` → `None`); the four generic tests carry kind/column/severity; models become `table://` assets keyed on the resource path with dbt's *resolved* schema; models are written and sources read; edges come from `parent_map`; selection scopes ownership.
- `dbt_executor`: node events become materialization records with the same key the end-of-run reconciliation derives — they must agree or a run settles progress against a path no graph node has.
- `dbt_profiles`: numerics unquoted / credentials unable to break out of their scalar; only `mssql` and `oracle` are enterprise-gated.
- `dbt` descriptor: parse, shipped-default engine, run-argument signature.
- `dbt_executor`: the project digest is content-addressed and order-free (it gates retry state and keys the package cache, so a value depending on map ordering would evict and reject on every run); a project-supplied `packages-install-path` cannot escape the project, since both cache copies are rooted at it.
- `cache`: a script's modules survive a round trip through the worker's file-system cache, and an entry written without them fails to import so it is refetched.

Fusion is confirmed to be **fetched, never bundled**: no Dockerfile references it except the comment saying so, and the first run logged `Fetching the dbt Fusion engine from dbt Labs (not bundled with Windmill; subject to the dbt Fusion engine license agreement)...` before landing a 289 MB engine in the worker cache.

**Verified manually, not committed**: the e2e scenarios above. They need a live Postgres and a real dbt install, which no CI harness in this repo provides; the two fixture projects are scaffolding under AGENTS.md, so they are not committed. The manifest fixtures in the unit tests are captured verbatim from real `dbt parse` output on both engines, which is what makes the ingest contract testable without that harness.

Checks: `cargo check` (CE), `cargo check --features enterprise,private,license,parquet,python`, `cargo test` on the touched crates (642 lib tests passed), `npm run check` (one pre-existing unrelated error: `modern-screenshot` is in `package.json` but missing from this box's `node_modules`).

## Scope change: dbt does not trigger downstream runs

Earlier revisions cascaded from a dbt run: finishing a project woke the Windmill
scripts reading its marts. That is gone, deliberately.

dbt already orders its own DAG, so a cascade only ever added one edge — waking a
script that reads a mart — and only half of it could be expressed, since nothing
outside dbt can declare a `table://` write (`// materialize` takes DuckLake
targets only). Against that, dispatching *correctly* was not cheap: a run's
`select` can build any subset of the project, so the deploy-time write set is not
what ran, and narrowing it needs a per-job record the per-relation state table
cannot supply. Four of five findings in one review round came from that
machinery.

So `asset_dispatch` returns early for `ScriptLang::Dbt`, and its surface went
with it: `table://` is no longer an auto-trigger kind, the ingest derives no
subscriptions, and `# on table://…` is refused at deploy rather than persisting
a row that draws a cascade arrow and wakes nothing. Everything else stands —
models, `ref()` lineage, per-model run progress, the project tree, SQL previews.
Adding dispatch later is a decision about what a selective run should notify,
not plumbing.

## Left out on purpose: the embedded OpenAPI bundles

`openapi.yaml` documents all seven dbt routes. The generated bundles the server
embeds (`openapi-deref.yaml` / `.json`) are NOT regenerated here. They were last
built at #10410 and are 35 versions behind: they are missing **73** routes, only
7 of which are this PR's. Running `build_openapi.sh` produces ~14k changed lines
(the redocly version bump renumbers every YAML anchor) and folds four other
merged PRs' API surface into this diff. Regenerating them is its own PR.

## Follow-up: a graph for a project that has not been deployed

Tracked as [WIN-2272](https://linear.app/windmilldev/issue/WIN-2272/dbt-its-own-editor-with-an-explicitly-refreshed-model-graph).

Two review findings are also deliberately left for follow-up rather than landed here:
[WIN-2273](https://linear.app/windmilldev/issue/WIN-2273/dbt-canonicalize-a-module-path-before-the-add-file-checks)
(the add-file dialog compares raw strings, so `./dbt_project.yml` slips past the duplicate and
reserved-name checks) and
[WIN-2274](https://linear.app/windmilldev/issue/WIN-2274/script-hard-delete-and-dbt-graph-publication-take-their-locks-in)
(hard delete and `persist_ingest` take the `script` and graph-sidecar locks in opposite orders,
so a concurrent delete and publish can deadlock — Postgres aborts one with an error rather
than hanging). Re-ordering locks in the delete path is not a change to land unreviewed at the
end of a PR this size.

The editor shows no graph, because `dbt_node` / `dbt_edge` rows only ever come from a
`manifest.json` that dbt itself produced — at deploy (`dbt deps` + `dbt parse`, no warehouse
touched) and per run when the model set differs. A buffer that has never been parsed has
nothing to draw, and a hand-rolled `ref()` parse would key no relations and disagree with
dbt about `enabled`, macros and loops.

The follow-up moves dbt to its own editor and gives it an explicitly refreshed graph: a
parse-only job over the buffer, reusing the deploy's own path. The one structural change it
needs is a third provenance for graph rows — `GraphPublisher::Unversioned` never publishes,
and `get_dbt_run_graph` refuses to pin a job with no deployed version, both on purpose — so
a buffer parse must write rows keyed to its own job id, readable only through that job id.

---
## Summary by cubic
Run unmodified `dbt` projects as first‑class Windmill scripts with a hardened sandbox, live per‑model progress, and graphs pinned by script version and by run. Deploy‑time parsing uses the script’s own envs, and dbt runs do not auto‑trigger downstream scripts (lineage only).

- **Bug Fixes**
  - Sandboxing: keep the `nsjail` profile outside the job dir, mount only the engine (no `repos/`/`packages/` bleed), and clean staging on all exits.
  - Parsing/lineage: dependency jobs parse with script envs; package caches are env‑scoped; keep cross‑selection read edges.
  - Auth/graph: run graph moved to `/jobs/dbt_graph/{id}` and requires `assets:read` + job read + tag scope; `raw_code` gated by `scripts:read`; pinned reads take path/hash from the job; share‑link viewers see shape only; refuse publication for archived/deleted scripts.
  - Validation: refuse `{{ }}` placeholders that shadow built‑in args, non‑object `vars` overrides, unknown descriptor fields, and invalid lock/adapter versions; quote profile names/keys in rendered YAML.
  - Retry: auto node‑level retry with a job‑wide budget; durable and worker‑local state keyed by principal; compare a stable digest of resolved args; do not persist resolved secrets (store submitted args, re‑resolve on retry); clear stale/missing results; observe cancellation; survives renames; pre‑upgrade identities restore.
  - Preview rows: select by package (`<name>,package:<pkg>`), clamp row and byte limits, cache per model, pin to the run’s version, and flag relation drifts/renames.
  - Graph storage: version‑ and run‑keyed graphs with a snapshot marker and SHA‑256 digests; snapshot only when the graph differs; prune aged snapshots; retain newest‑50 deploys per path; keep archived graphs; record and publish the `published_relation_root`; snapshot/reads require job visibility and tag scope.
  - Progress: job‑keyed table for live per‑model state; finished runs colour from their own result; no read without job‑access.
  - Agents: `/api/agent_workers/dbt_graph/{workspace_id}` publishes graphs bound to the verified job; live progress stays SQL‑only.

<sup>Written for commit 593758ff92c3c66cfdaa00ca9550dd26b1b6e4b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
## ai_evals A/B: not run, and why

The `ai-chat` rule is "no context or behavior change ships without an `ai_evals`
A/B on the affected mode." I could not run one: `ai_evals` needs a provider API
key (`GEMINI_API_KEY` / equivalent) and this environment has none. Flagging
rather than skipping quietly — a maintainer with a key should decide whether to
run `flow` before merge.

What the Copilot-affecting change actually is, so that decision has something to
weigh:

- The only touched files are `openFlowZod.gen.ts` and `openFlow.json`, both
  regenerated verbatim from `openflow.openapi.yaml` via the existing
  `gen_openflow_schema.sh` → `minifiedOpenflowJson.sh` pipeline. No hand edits.
- The diff is two enum widenings: `language` gains `dbt`, asset `kind` gains
  `table`. Nothing else changed — no prompt, no context assembly, no tool
  definition, no schema shape.
- It is strictly permissive. Before it, Copilot's validator *rejected* flow
  modules carrying either value; after it, they validate. So the failure mode it
  removes is a false rejection, and it cannot make a previously-accepted flow
  fail.
- `providerKind.test.ts`, the existing guard over this generated schema, passes.

My judgement is that an A/B on `flow` would be measuring noise against an
additive enum change, but that is a judgement, not a run, and the rule does not
carve out additive changes — so the gap is real and recorded here rather than
asserted away.

rubenfiszel:
On the `openFlowZod.gen.ts` nit: `dbt` is deliberately absent from `RawScript.language`, and the generated file matches its source (`openflow.openapi.yaml`). `ScriptLang` still carries `dbt` — a deployed dbt script runs fine as a flow step by path. What is excluded is *inline* dbt in a flow, which the worker rejects (`worker.rs`: "Inline execution is not yet supported for this language"): a dbt script's content is a descriptor naming a git repo, so there is nothing to inline. Advertising it would let such a flow validate and then always fail at runtime — the round-36 finding this removal came from.

rubenfiszel:
Replying to the Codex review above (`forgeable retry ownership check`), which has now been raised on three consecutive heads. I believe it is incorrect, and the reasoning is checkable rather than a judgement call — but if any step below is wrong I would rather be shown it than leave the hole open.

**The claim.** `caller_may_resume` / `resume_refusal` compares `saved.created_by` with the retry job's `created_by`; both come from `display_username()`, so two users can mint tokens with the same label, collide on `created_by`, and resume one another's runs.

**Why the collision cannot be produced.** `created_by` is `username_override.unwrap_or(username)`, and `username_override` comes from the token label through `username_override_from_label` (`backend/windmill-api-auth/src/auth.rs:857`). That function has exactly one arm that yields a *bare* username:

| label | `username_override` |
|---|---|
| `webhook-*`, `http-*`, `email-*`, `ws-*`, `ephemeral-webhook-*` | the label verbatim (`webhook-foo`) |
| `ephemeral-script-end-user-<x>` | **`<x>`** — bare |
| `Ephemeral lsp token` | `lsp` |
| any other non-empty label | `label-<label>` |
| `session` / `ephemeral-script` / empty | `None` → the real username |

Every arm but one is prefixed, so it cannot equal a real username such as `alice`. The one that is bare is refused at token creation: `is_reserved_token_label` rejects `ephemeral-script-end-user-` in `create_token_internal` (`windmill-api-auth/src/lib.rs:920`), which the comment there notes is the single path every caller-supplied label reaches. A member therefore cannot mint a token whose `created_by` is another member's username.

That guard is part of this PR, added for precisely this class (it is what closed the impersonation half of GHSA-8x8x-88qc-qp4r for dbt), and `only_labels_that_become_a_bare_username_are_reserved` pins it.

**What the equality is doing.** It is not the authorization; it mirrors the first grant of `require_job_read_access` ("you can always read a job you launched"). When it does not match, the decision falls through to an RLS probe evaluated as the caller — built from `usr` + `get_groups_for_user` + `get_folders_for_user`, deliberately not `job_perms`, since those carry the identity the job runs *as* (for an `on_behalf_of` script, the owner, which would authorize everyone).

**Verified behaviour** with two non-admin members sharing one `on_behalf_of` principal:

| case | can B read A's run | B's retry |
|---|---|---|
| `u/<owner>` script shared via `extra_perms` | 403 | refused; A's marker value appears zero times in B's result |
| same script under `f/<folder>` with folder read | 200 | allowed, which is the documented sharing |
| A retrying her own run | 200 | allowed |

Two follow-ups from the same review round *were* real and are fixed in `12d28cd`: a pruned source job authorized the retry, and a `dbt_run_state` row with a NULL `job_id` skipped the check instead of failing closed.

If the intended attack is a label form I have not enumerated, or a path that reaches `created_by` without going through `create_token_internal`, please name it concretely — a label string and the resulting `created_by` — and I will fix it.

rubenfiszel:
Correction to my comment above, since one line of it was too strong and I would rather flag it than let it stand.

I wrote *"every arm but one is prefixed, so it cannot equal a real username"*. That is wrong as stated: the `Ephemeral lsp token` arm yields the bare string `lsp`, so two arms produce an unprefixed value, not one. (cubic's review flags the same thing at `backend/windmill-api-auth/src/auth.rs:854`.)

The conclusion is unchanged, but for a narrower reason than I gave: `lsp` is a **fixed** string, not attacker-chosen, so a member still cannot mint a token that resolves to another member's username. The residual is confined to a workspace that contains a user literally named `lsp` — for whom an `Ephemeral lsp token` would share `created_by`.

Reserving that label is not available as a fix: the editor mints it through the same `createToken` route for the language server (`frontend/src/lib/components/Editor.svelte:2001`), and `only_labels_that_become_a_bare_username_are_reserved` pins it as allowed for that reason. Closing it properly means not deriving a username from that label at all, which is a change to LSP job attribution and outside this PR.

So: the retry check is not forgeable to an arbitrary identity, which is what the original finding claimed; there is a `lsp`-shaped edge I had waved past, and it is now written down rather than implied.

rubenfiszel:
On the CLI finding from the last round (`isScriptModulePath` matching `__dbt/` anywhere, so `f/team__dbt/job.py` reads as a bundle): that rule predates this PR. `origin/main` has

```ts
export function isScriptModulePath(p: string): boolean {
  return normalizeSep(p).includes(MODULE_SUFFIX + "/");
}
```

so `f/team__mod/job.py` behaves identically today. This branch generalizes the same predicate to the dbt suffix rather than introducing the collision. Distinguishing a bundle from a same-named folder cannot be done from the path alone — it needs the local tree (a bundle has a sibling `<name>.dbt.yaml`, a `__mod` folder an inner `script.*`), which would change the contract of a pure helper used at five call sites including the sync diff. Left as is, and tracked as pre-existing rather than fixed here.

rubenfiszel:
Two findings from the local cold-review rounds, resolved as follows.

**Escaping `packages-install-path` (fixed, 262ba574).** Correct: the sanitized value only ever moved Windmill's cache, while `dbt deps` reads `dbt_project.yml` itself. It is now refused instead of substituted. Verified end to end — deploying a project with `packages-install-path: ../../escape` **and** a `packages.yml` fails its dependency job with:

```
Bad request: `packages-install-path: ../../escape` in dbt_project.yml points outside
the project; it must be a path within it, such as `dbt_packages`
```

**Retry racing a concurrent run of the same script (documented, not serialized).** Two ordinary concurrent `build`s of one incremental project have the same double-append exposure — dbt takes no cross-process lock — so this is a property of running a project twice at once rather than of `retry`. Windmill's mechanism for it is the script's concurrency limit, which covers retries with it; a lock held across a dbt execution would outlive worker deaths, cancellations and timeouts. Stated in `docs/dbt-runtime.md` under the retry section.

rubenfiszel:
Latest cold-review round, both P1s:

**Retry state rewritten under a renamed path (fixed, 76e66cff).** Correct — a rename leaves the old path *archived*, not deleted, so the `NOT deleted` guard still matched it. The save now requires a live version by the same test `clear_dbt_run_state_if_path_retired` retires state with (`deleted = false AND archived = false`), so the two agree on what "a live version at this path" means.

**Cross-caller argument disclosure on a shared `on_behalf_of` script (accepted, unchanged).** This is the residual documented under *"What a retry actually adds, and where that crosses a line"* in `docs/dbt-runtime.md`, and the decision to accept rather than gate it is deliberate and was taken explicitly on this PR. Two things worth stating precisely:

- The capability is not new to `dbt_retry_job`. Before a retry had to name a run, an unnamed `retry` resumed the script's last failure for that principal — no id needed. Requiring the id, and having an endpoint return the one the caller's own retry would land on, narrows what a caller can aim at; it does not widen who can resume.
- Gating it at the worker is what was tried and reverted: the worker has no caller identity, only `created_by`, which is `display_username()` and therefore supplied by a token *label* (GHSA-8x8x-88qc-qp4r is precisely about not trusting that name). The gate broke labelled CI-token retries. Authorizing `dbt_retry_job` as a job read has to happen at the submitting path, where the caller is real — that is written down in the docs as the design that would also let a hidden run be resumed by its author.

It takes all three of a `u/<owner>` path, `extra_perms` sharing and `on_behalf_of` for "may run as this principal" to be broader than "may read this principal's jobs", and what leaks there is the literal argument values of a run under an identity the caller may already execute as. References stay references, so no resolved secret is among them.

rubenfiszel:
On the `on_behalf_of` retry-disclosure finding, which three cold passes have now raised — split into its two halves, because they have different answers.

**Discovery: closed (3e3c2c45).** `get_dbt_resumable_for_script` now returns a job id only when the caller may READ that run (`require_job_read_access`, plus the app-embed cutoff added in 284c005f). A grantee of a shared `on_behalf_of` script who cannot open the owner's failed run is no longer told it exists or what its id is. Callers who can read the run — their own runs, folder-readable runs, which is what the prefill exists for — are unaffected.

**Resuming: accepted, deliberately.** If you *can* read the run, you can retry it, and the retry republishes that run's submitted arguments as `invocation_args`. That is the residual written down in `docs/dbt-runtime.md` under *"What a retry actually adds, and where that crosses a line"*, and it is the same capability as re-running the job you are already entitled to read.

Gating the *acceptance* of `dbt_retry_job` was implemented and reverted earlier in this PR on purpose: the worker has no caller identity, only `created_by`, which is `display_username()` and therefore set by a token **label** — the exact thing GHSA-8x8x-88qc-qp4r says not to trust. That gate denied every labelled token (a CI token becomes `label-<name>`, no workspace member) while still trusting a name. Doing it at the submitting path instead would mean the generic run route peeking into dbt-shaped arguments on every script run; the docs record that as the design that would also let a hidden run be resumed by its author, not as an oversight.

PR #10337, opened 2026-07-26
feat: build a React raw app from the script and flow detail pages

rubenfiszel:
## Summary

The **Build app** button on the script and flow detail pages used to scaffold a low-code app (a `verticalsplitpanescomponent` grid wired to a form + run button). It now scaffolds a **full-code React raw app** instead: the script/flow becomes a `runnableByPath` backend runnable, and `App.tsx` is generated from the runnable's schema as a plain React form that calls `backend.<runnable>(...)` and renders the result.

Fixes WIN-2247

## Changes

- New `frontend/src/lib/components/details/createRawAppFromScript.ts` — generates a React 19 raw app (`/App.tsx` + `/index.css` on top of `react19Template`) plus a `type: 'path'` runnable pointing at the script/flow. `createRawAppFromScript` / `createRawAppFromFlow` differ only in `runType`, which drives the runnable kind and the policy key (`<id>:script/...` vs `<id>:flow/...`).
- **Form generation** — one input per schema property in `schema.order`: text, password, number, checkbox, `enum` select (honouring `{ value, label }` pairs), JSON textarea for objects/arrays. Titles, descriptions, defaults and required markers come from the schema; `required` is enforced on the inputs, not just drawn.
- **Argument marshalling** — required args pass through directly (matching the non-optional type `genWmillTs` derives from the same schema); blank optional text/number/JSON send `undefined` so the runnable's own default applies. `__proto__` is emitted as a computed key, since `{ __proto__: v }` sets the prototype instead of creating the argument.
- **Resources** — `format: resource-*` props render a `$res:` text field and declare `allowUserResources` on the runnable, without which the backend replaces the submitted reference with a placeholder and the argument never resolves.
- **Secrets** — properties flagged `password` render masked and are declared `sensitive` on the runnable, which is what makes `updateRawAppPolicy` emit `sensitive_inputs` and the backend store `$encrypted:…` instead of plaintext job args.
- **Identifier safety** — argument names become JS bindings, so the allocator reserves every identifier the generated body references (its own locals *and* their setters, plus `JSON`, `Number`, `String`, `Error`, `undefined`) and every strict-mode-illegal binding (`eval`, `arguments`, keywords). A script argument called `result`, `setResult`, `class` or `undefined` still produces compilable code.
- **Theming** — the generated stylesheet paints an opaque surface and adapts via `prefers-color-scheme`. Without a background the iframe is transparent and the host page's dark theme shows through under dark text.
- Both detail pages hand the bundle off via `sessionStorage.rawAppImport` (the `/apps_raw/add` navigation is a hard reload for cross-origin isolation, so the in-memory `importStore` would be lost) and navigate to `/apps_raw/add`.
- Deleted `createAppFromScript.ts` — the low-code scaffold had no other callers. Creating a low-code app from scratch is untouched.

### Known limitation

Resource arguments are entered as a `$res:u/user/my_thing` path in a text field rather than through a resource picker. They resolve correctly, but the picker would be a nicer follow-up.

The generated app follows the viewer's OS `prefers-color-scheme`, not Windmill's in-app dark toggle — the toggle is a `dark` class on the host `<html>` and the raw-app iframe is a separate document that cannot read it. Matching it would mean propagating the theme through the raw-app runtime, which affects every raw app, not just generated ones.

## Screenshots

Script detail page → **Build app**, landing in the raw-app editor:

rawapp1

Generated app running the script, light and dark:

v2-light
v2-dark

Same from a flow detail page:

flow-dark

## Test plan

- [x] `npm run check` — 0 errors; `vitest` — 10 cases passing
- [x] Script detail → Build app → editor opens with the generated files and the `runnableByPath` runnable; esbuild build succeeds; running the form returns the script result
- [x] Flow detail → Build app → same, with `runType: 'flow'`
- [x] Deploy the generated app: `policy.triggerables_v2` gets `<id>:flow/u/admin/greet_flow`; the deployed app at `/apps_raw/get/…` runs the flow
- [x] Collision cases (`result`, `setResult`, `running`, `JSON`, `eval`, `class`, `undefined` args): distinct bindings, build succeeds, correct args sent
- [x] Labeled enum displays its labels and submits its values
- [x] Password + `__proto__` args — job record shows `"secret": "$encrypted:…"` and `"__proto__": 7`
- [x] Blank optional arg arrives omitted (`typeof n === 'undefined'`), not as `''`
- [x] Flow input named `user's-name` compiles and submits under that exact key; untouched optional resource is omitted (args: `{"user's-name": "ada"}`)
- [x] Flow input named `user-name` types correctly in the editor; enum value `a&amp;b` reaches the runnable undecoded
- [x] Resource arg resolves server-side (`{ host: "example.com", gotObject: true }`)
- [x] Dark and light rendering verified in the editor preview and on a deployed app


rubenfiszel:
✅ Review round clean @ 68db34918d99221af21169ee62d40f6a046acc9f — nit-only verdicts at f11b4c6aa06feb27bda5e261241e80fdc99ec2ad (codex: mergeable-with-nits, claude: mergeable-with-nits, pi: good to merge, cubic: 1 P2); both nits addressed in 68db34918d.

- Required array enums now emit `required` on the `<select multiple>` — verified in the browser: `checkValidity()` is false with nothing selected, the browser reports "Please select an item in the list." and the submit is blocked.
- Dropped the review-history sentence from the regression-test comment per AGENTS.md (comments record constraints, not chronology); the preceding sentence already states the invariant.

PR #10341, opened 2026-07-27
fix: enforce per-job authorization on cancel and force_cancel endpoints

rubenfiszel:
## Summary

`cancel_job_api` and `force_cancel` in `workspace_unauthed_service` did no per-job
authorization for authenticated callers. Every read endpoint in the same service
(`get/{id}`, `getupdate/{id}`, `completed/get/{id}`, …) goes through
`require_job_read_access` — which checks ownership, admin status, RLS visibility and the
flow ancestor chain — but the two cancel handlers skipped it entirely (aside from a narrow
app-embed-token check on `cancel_job_api`). Any authenticated workspace member who knew a
job UUID could cancel or force-cancel it, including runs hidden from their own job listing.

Both handlers now take the same gate as their read peers. The bulk `cancel_selection`
endpoint in `workspaced_service` was never affected — it resolves ids through RLS-scoped
`user_db` queries.

Fixes WIN-2248

## Changes

- `cancel_job_api` and `force_cancel` extract `Extension<UserDB>` and call
  `require_job_update_read_access` before any cancellation work. Access rules are now
  identical to reading the job: you launched it, you are an admin, or it is RLS-visible to
  you directly or through a flow ancestor. Anonymous callers are unchanged — they stay
  confined to `anonymous`-created jobs by `cancel_job`'s `require_anonymous`.
- `force_cancel` authorizes the job it *actually* kills. `cancel_job(force_cancel = true)`
  does not cancel the id in the URL: it walks up to the highest still-queued ancestor and
  cancels that. Since read visibility is inherited *down* the flow chain, gating on the
  requested id would have let a caller who can only see an inner step force-cancel a root
  flow hidden from them. The new `force_cancel_target` resolves the same ancestor the queue
  will pick, and the authorization check runs against it.
- The ancestor walk is bounded (a cyclic `parent_job` chain must not spin forever) but
  `cancel_job`'s own walk is not, so a chain longer than the bound would leave the two
  disagreeing about which job gets killed. Hitting the bound therefore fails closed with an
  error rather than authorizing an ancestor that may not be the one cancelled.
- Removed the inline app-embed-sentinel block in `cancel_job_api`. It is now exactly what
  `require_job_read_access` already enforces (`created_by == authed.username`, then
  `NotFound` for any embed-sentinel caller), so it only duplicated the invariant at a second
  site and cost an extra `SELECT created_by` per embed cancel. The embed confinement is
  unchanged and still pinned by the existing assertions in `jobs_read_auth.rs`.
- No `view_token` is passed at either call site: share links grant read access, never the
  right to kill someone else's run. Recorded in a comment since every sibling read handler
  forwards its token.

## Deliberate non-goals

- **Share-link UI.** A session that arrived via `?view_token=…` still renders the Cancel /
  Force-cancel buttons and now gets an error toast instead of silently cancelling a run it
  could only read. Hiding those buttons under a view token is a frontend follow-up.
- **`cancel_persistent_script_api`.** The adjacent `queue/cancel_persistent/{path}` route on
  the same router remains unauthorized beyond "logged in": it resolves jobs by
  `runnable_path` on the root DB handle with no ACL or RLS check. That is pre-existing and
  needs its own change (a script path is easier to guess than a job UUID), so it is left out
  of this PR rather than widened into it.

## Test plan

Backend regression coverage in `backend/tests/jobs_read_auth.rs` (fixture
`jobs_read_auth.sql`), all verified to fail without the corresponding fix:

- [x] A workspace member with no ACL on the runnable gets `403` on both
      `queue/cancel/{id}` and `queue/force_cancel/{id}` for another user's running job.
      Without the fix: `200 OK`, job cancelled.
- [x] The job's owner can still cancel it (guards against over-blocking).
- [x] A viewer who can read a queued sub-flow (folder ACL) but not its parent top flow is
      denied `queue/force_cancel` on that sub-flow. Without the ancestor resolution:
      `200 OK`, the hidden top flow is killed. The test first asserts the sub-flow *is*
      readable, so the denial can only come from the escalation guard.
- [x] Pre-existing app-embed assertions still pass: an embed token cancels the job it
      launched and gets `NotFound` for another user's.
- [x] `cargo test --features quickjs --test jobs_read_auth` — 1 passed.
- [x] `cargo test -p windmill-api-integration-tests --test jobs_unauthed` — 2 passed
      (route reachability for both cancel endpoints).
- [x] sqlx offline cache regenerated for the new `force_cancel_target` query (verified
      zero entries lost relative to `origin/main`).
- [x] Live check against a dev instance: an admin still cancels and force-cancels their own
      running job (`200`), and an unknown UUID returns a clean `404`.

No boundary test for the depth cap: reaching it needs 500+ nested queued flows, and the
behavior there is a denial rather than a bypass.

No frontend changes.


rubenfiszel:
✅ Review round clean @ 3fa1aea8eb2055bdafc303862c3e216601b38abc

PR #10343, opened 2026-07-27
feat(frontend): hide empty folders/users in home tree view

diegoimbert:
Fixes WIN-2250

## Summary

Since the lazy folder tree started injecting every workspace folder and user as a top-level node, folders and user namespaces containing zero runnables cluttered the home tree (e.g. auto-created `app_themes`/`app_groups`/`app_custom` folders). This PR filters them out server-side.

## Changes

- **`GET /w/{workspace}/folders/listnames`** (`backend/windmill-api-groups/src/folders.rs`): new optional `non_empty` query param. When true, only folders holding at least one non-archived script/flow/app are returned, via a `name IN (SELECT split_part(path,'/',2) FROM script/flow/app WHERE path LIKE 'f/%' ...)` semi-join — one scan per table instead of the per-folder correlated `EXISTS` probes sketched in the issue (a non-constant `LIKE` prefix can't use an index, and `_` in folder names would be a wildcard). Composes with the existing scope-filter and pagination clauses ($1 is simply reused).
- **`GET /w/{workspace}/users/list_usernames`** (`backend/windmill-api-users/src/users.rs`): same optional `non_empty` param filtering to users whose `u/<username>/` space holds at least one non-archived script/flow/app. Other callers (ShareModal, GroupEditor, etc.) are unaffected — the param defaults to false.
- Both filters run inside the caller's `user_db` transaction, so RLS applies: a folder/user whose items the caller cannot read counts as empty *for them*.
- **`backend/windmill-api/openapi.yaml`**: documents the new param on both operations.
- **`frontend/src/lib/components/home/ItemsList.svelte`**: `folderNamesRes` / `usernamesRes` now pass `nonEmpty: true`, so both the lazy tree injection and the owner chips hide empty owners. Exception: in the archived-only view the param is dropped — a folder whose items are all archived counts as empty, yet must stay reachable to browse its archived items (the resources now also depend on `archived`, so toggling refetches). The `archived` state declaration moved above the resources it now feeds.
- `TreeViewRoot.svelte` is untouched: pipeline folders are still injected unconditionally (a bundle-phase pipeline legitimately has no listed items), and the user's own `u/<username>` node is still always injected.

## Judgment calls

- The owner filter chips share the same fetches, so empty owners disappear from the chips too — consistent with the tree, and any owner present in loaded items still gets a chip from the item-derived fallback.
- A folder containing only library scripts (`auto_kind = 'lib'`) counts as non-empty even when "library scripts" are hidden client-side; tying the filter to that toggle didn't seem worth the extra param.
- Users: went with the backend `non_empty` param (issue offered that or frontend post-hoc collapse).

## Test plan

- [x] `GET /folders/listnames?non_empty=true` returns only folders with runnables (verified: workspace with 5 folders → only the one holding a script; a folder whose only script is archived is excluded)
- [x] `GET /users/list_usernames?non_empty=true` returns only users with runnables (verified: `admin` with a script listed, `ghost` without one excluded)
- [x] Home tree view shows only non-empty owners; expanding them still lazy-loads their items (screenshots below)
- [x] `npm run check` — 0 errors; backend compiles and serves under cargo watch with no warnings/errors in logs
- [ ] CI review round

Local review (fresh-context branch-diff-reviewer): Good to merge, no findings. Codex local pass skipped: devbox codex CLI is 0.132.0 < pinned 0.144.1.

## Screenshots

Workspace has 5 folders (4 empty) and 2 users (1 without runnables):

folders_page

Home tree view only shows `u/admin` and `f/occupied`:

home_tree

Expanding still lazy-loads the owners' items:

home_tree_expanded

https://claude.ai/code/session_01UomUzAvrC2cto1T2QUBNnT


diegoimbert:
Addressed the round-1 findings in 8a09343975:

- **Codex P1 (token scopes)**: the semi-join now embeds each domain's `scripts`/`flows`/`apps` read scope, via a shared `non_empty_owner_subquery` helper in `windmill-api-auth` used by both endpoints (empty grant → the branch matches nothing). Verified against the running backend: a token with `folders:read:*` but `scripts:read:f/none/*` gets `[]` from `listnames?non_empty=true` (while unfiltered listnames still lists all folders), and granting `scripts:read:f/occupied/*` makes exactly `["occupied"]` appear; same for `list_usernames?non_empty=true` with `u/admin/*`.
- **Codex P2 (demo short-circuit)**: the redacted placeholder is dropped when `non_empty=true` (it isn't a real user with runnables); the caller's own username is kept.
- **Claude P1 (stale owner lists)**: `reloadItems()` now refetches `folderNamesRes`/`usernamesRes`, so content mutations update the injected owner nodes. Verified in the browser: archiving the last script of `f/occupied` from the tree row menu removes the folder node without a page reload.
- **Claude P2 (struct visibility)**: both query structs are now private.
- **Claude P2 (perf, non-blocking)**: leaving the single-scan semi-join as is — `archived = false` already excludes superseded script versions from the scan (old versions are marked archived on deploy), each branch is bounded by the workspace's rows via the workspace index, and the alternative per-folder correlated probe is strictly worse (non-constant LIKE prefix can't use an index either, times N folders). Can revisit with an `EXPLAIN` on a large cloud workspace if it shows up in slow-query logs.

diegoimbert:
✅ Review round clean @ 93aab4b57c0a2a3877be9ba5bce875e2ec59d974 — nit-only verdicts at 8a093439758ce6ce2398449a4f740ba38b780837 (codex: Good to merge, pi: Good to merge, claude: nit-only); nits addressed in 93aab4b57c (hard assert on the SQL-interpolated path_prefix, order-independent first_bind at the folders.rs call site, plus unit tests pinning the fail-closed scope branches and bind numbering)

diegoimbert:
## Benchmark: `non_empty` endpoints

Seeded the dev workspace with synthetic data and measured end-to-end HTTP latency (20 requests each, admin token, `per_page=1000`).

**Scale 1 — 1,000 folders (500 non-empty), 300 users (150 non-empty), 34.5k script rows (11.5k current), 1k flows, 500 apps:**

| endpoint | min | p50 | avg | max |
|---|---|---|---|---|
| `folders/listnames` (baseline) | 7.0ms | 7.8ms | 8.0ms | 9.4ms |
| `folders/listnames?non_empty=true` | 22.1ms | 24.4ms | 24.2ms | 27.8ms |
| `users/list_usernames` (baseline) | 3.0ms | 3.6ms | 3.8ms | 5.5ms |
| `users/list_usernames?non_empty=true` | 4.9ms | 6.1ms | 6.2ms | 7.7ms |

**Scale 2 — 5,000 folders (2,500 non-empty), 94.5k script rows (31.5k current), after the CTE fix (7a2e9aad):**

| endpoint | min | p50 | avg | max |
|---|---|---|---|---|
| `folders/listnames` (baseline) | 6.6ms | 8.0ms | 7.9ms | 9.6ms |
| `folders/listnames?non_empty=true` | 17.3ms | 18.7ms | 19.4ms | 22.3ms |
| `users/list_usernames` (baseline) | 3.0ms | 3.9ms | 3.9ms | 4.9ms |
| `users/list_usernames?non_empty=true` | 5.4ms | 6.5ms | 6.8ms | 10.3ms |

### What the profiling found (and the fix it motivated)

`EXPLAIN ANALYZE` on the folders query at scale 1 showed the planner choosing a **nested loop that linearly probed the aggregated owner set once per folder row** (1,005 × 376 ≈ 378k join-filter comparisons, ~30ms of a 34ms execution) — the `ORDER BY name LIMIT` makes streaming folders in `folder_pkey` order look attractive, and the inline `IN (<union>)` gets flattened into that join. The union scan itself was only ~2.7ms. With fresh table stats the planner picks a hash join (~11ms) instead, so the bad shape is stats-dependent.

Commit 7a2e9aad fences the subquery in a `WITH owners AS MATERIALIZED (...)` CTE, which pins the good shape in both stats states: single scan per table → hash/merge join (4.4ms execution at scale 1, 12.6ms at scale 2, vs 34ms for the bad plan). The `list_usernames` query has no `ORDER BY`/`LIMIT`, never exhibits the pathology (1.2ms execution), and is left unfenced.

Verified after the fix: pagination is exact (2,501 non-empty folders → pages of 1000/1000/501/0), scoped-token results unchanged. Bench data cleaned up.

diegoimbert:
✅ Review round clean @ 2c1e227d294167b07e357b01c8293c9732c608b7 — nit-only verdicts at 7a2e9aadf1ccd90635ebb6dcf88215ceb72c7517 (codex: Good to merge, pi: Good to merge, claude: nit-only); nit addressed in 2c1e227d29: the owner-list refetches moved into an `onItemMutated()` wrapper bound to the mutation events, so scope changes (mount/sort/kind/archived) no longer duplicate the two listings. Verified in the browser: exactly one `listnames`/`list_usernames` request on home load and zero on a sort toggle, and archiving a folder's last script still drops its node live.

PR #10347, opened 2026-07-27
fix: show folder item counts on the homepage tree before expansion

Guilhem-lm:
## Summary

On the homepage tree view, a collapsed folder or user showed no item count — the number only appeared after you clicked to expand it. #10297 made the tree load each owner's contents lazily (so a chosen sort order stays globally correct on large workspaces), which left nothing that knew a collapsed owner's size; the header deliberately rendered a blank rather than a misleading `(0 items)`.

This adds a dedicated count endpoint so every owner shows its true total up front, and makes the number stable across expansion instead of jumping.

## Changes

- **New endpoint** `GET /w/{workspace}/runnables/count_by_owner`: one grouped `COUNT` per top-level owner prefix (`f/<folder>` / `u/<user>`), under the same visibility predicates as `runnables/list` — RLS via the `user_db` transaction, per-domain token path scopes, the archived-latest-version filter, library-script exclusion and the kind filter. The scope-predicate builder is extracted from a closure into a shared function parameterized by the number of preceding bind placeholders, so both handlers use one implementation.
- **Counts exclude pipeline member scripts**, matching the tree: those fold into their folder's single `Pipeline` row, so counting each one would promise more rows than expanding reveals. The frontend adds the `Pipeline` row back into the displayed total.
- **Header shows the server total in every state** (collapsed and expanded), so the count no longer changes when you open a folder — the loaded tree's own `items.length` only counts direct children, collapsing each subfolder to one entry.
- **Empty owners are inert**: an owner with no counted descendants and no `Pipeline` row renders without a chevron, ignores clicks, and no longer flashes a `Loading…` row before showing nothing. A pipeline-only folder stays expandable.
- **One request per change**: counts are keyed on the inputs the endpoint resolves (workspace, archived, library, kind) plus a mutation counter, so a row mutation refetches, a sort change — which cannot move a total — does not, and a superseded response is aborted rather than allowed to overwrite the current totals.
- If the counts request fails, headers fall back to the previous blank-until-expanded behavior rather than showing a wrong number or locking folders shut.

## Screenshots

Collapsed owners show their totals immediately; empty ones have no chevron:

pr-tree-collapsed-counts

Expanding keeps the same number:

pr-tree-expanded-and-inert-empty

## Test plan

- [x] `backend/tests/runnables_list_pagination.rs`: per-owner counts equal the owner-grouped tally of what `runnables/list` returns (unfiltered and per kind); a repeated kind does not double them; a pipeline member is not counted; a non-admin is counted only the owners they can read. Full file green (9/9).
- [x] Browser: collapsed folders/users show true totals with nothing expanded; the number does not change on expand (incl. a folder with a subfolder); an empty folder has no chevron and ignores clicks; switching the kind filter updates counts; one count request per filter change and none on a sort change.
- [ ] On a workspace with a deployed pipeline, confirm the pipeline folder's count matches the rows it expands to (`Pipeline` row + siblings).
- [ ] With a fine-grained token (e.g. `scripts:read:f/foo/*`), confirm `count_by_owner` returns only `f/foo`.

### DIFF
--- a/backend/tests/runnables_list_pagination.rs
+++ b/backend/tests/runnables_list_pagination.rs
@@ -481,3 +481,129 @@ async fn test_runnables_starred_pinned_first(db: Pool<Postgres>) -> anyhow::Resu
     );
     Ok(())
 }
+
+/// `owner -> count` from the count endpoint, sorted for stable comparison.
+async fn counts_by_owner_as(port: u16, query: &str, token: &str) -> Vec<(String, i64)> {
+    let url =
+        format!("http://localhost:{port}/api/w/test-workspace/runnables/count_by_owner?{query}");
+    let resp = authed(client().get(&url), token).send().await.unwrap();
+    assert_eq!(
+        resp.status(),
+        200,
+        "count_by_owner should succeed for {query}"
+    );
+    let body: serde_json::Value = resp.json().await.unwrap();
+    let mut out: Vec<(String, i64)> = body
+        .as_array()
+        .unwrap()
+        .iter()
+        .map(|c| {
+            (
+                c["owner"].as_str().unwrap().to_string(),
+                c["count"].as_i64().unwrap(),
+            )
+        })
+        .collect();
+    out.sort();
+    out
+}
+
+/// The same tally, derived from the rows `list_runnables` actually returns.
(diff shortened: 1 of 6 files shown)


Guilhem-lm:
Closing — this will be handled by @rubenfiszel instead.

PR #10349, opened 2026-07-27
fix(frontend): stop spurious asset analysis toasts in the flow editor

rubenfiszel:
## Summary

The flow editor asks *"Assets were detected in this step. Analyze entire flow for assets?"* whenever a raw-script step whose `assets` field is unset turns out to declare assets in its code. Nothing recorded that the question had already been asked, so the prompt reappeared on every step click. Worse, the selection watcher is re-created (and fires eagerly) every time the module list changes, so adding a step re-fired it for the currently selected step too — which is why building a flow with the AI chat, one step at a time, buries the user in toasts.

Steps created during the session were also treated like legacy steps: their `assets` field is unset until parsed, so each newly inserted step raised its own prompt, even though writing its assets only completes an edit the user has already made.

Fixes WIN-2251

## Changes

`frontend/src/lib/components/flows/FlowAssetsHandler.svelte`:

- **Offer the analysis once per editor session.** The prompt triggers a flow-wide action, so a second one carries no new information. Accepting it also carries over: modules analyzed later in the session are applied silently instead of waiting on a prompt that will not be shown again.
- **Only prompt for modules the flow was loaded with.** Ids are snapshotted at mount (the editor only mounts after the flow is loaded) and pruned when a module is deleted, since flow module ids are recycled. Steps added afterwards — by the AI chat, by *Duplicate*, by *Add step* — get their assets written silently.
- **Cache the inference per module, keyed on language *and* content, and replay it on a hit.** The selection watcher fires on creation and is re-created on every structural edit, so without a cache each add/remove re-ran `inferAssets` on the selected step. Replaying rather than early-returning keeps a cache hit idempotent, so a module whose `assets` alone was reset (undo/redo, restore to deployed, AI diff apply) still gets them re-derived — and `analyzeEntireFlow` needs no force flag. Cached values are copied before reaching the flow store, which would otherwise proxy them and let a later replay mutate the cache in place.
- Toast duration raised to 20s: it is the only entry point to "Analyze entire flow" and it is now offered once, so a 5s window was too easy to miss.
- Dropped the per-parse `console.log`.

Behaviour that is deliberately unchanged: editing a step's code still updates its assets silently, and "Analyze entire flow" still parses every raw-script step.

A persistent "Analyze flow for assets" affordance outside the toast would be a good follow-up; this PR does not add one.

## Test plan

Verified against a flow created through the API with three Python steps that reference `s3:///data/*.csv` and carry no `assets` field (a flow predating the assets feature), using a `MutationObserver` to count toasts.

- [x] Clicking through steps a → b → c → a: **3 toasts before, 1 after**
- [x] "Analyze entire flow" writes assets to every step (all three asset nodes appear in the graph)
- [x] Duplicating a legacy step: the new step's assets are applied silently, no second toast
- [x] Editing a legacy step's code applies its assets silently, as before
- [x] Toast stays up long enough to click (still visible ~11s after it fired)
- [x] Changing *only* a step's language through the OpenFlow YAML editor re-infers its assets (see screenshots)
- [x] "Restore to deployed" after accepting: it performs a full navigation, so the handler remounts with a clean session and the flow returns to its deployed state
- [x] `npm run check:fast` — only the pre-existing `modern-screenshot` error, present on `main` too
- [x] Prettier clean; file is ASCII text and the diff renders as 52/15 lines

## Screenshots

The prompt, fired once for the loaded flow:

repro-03

Still actionable ~11s later (previously auto-dismissed at 5s):

toast-duration

A step added during the session (`c2`, duplicated from `c`) gets its `c.csv` asset resolved silently — no prompt — while the loaded steps a/b/c stay untouched until the user accepts:

dup-04

Language-change invalidation. After accepting the analysis, step `a`'s language is switched `python3` → `bash` in the YAML editor with its content left byte-identical:

yaml-bash

On apply, `a` re-infers under the bash parser and its `a.csv` node clears, while the untouched python steps keep theirs — a content-only cache key would have replayed the stale python result:

lang-after


rubenfiszel:
✅ Review round clean @ 43adffb932e00f11e5f0599a5f17fc4020398147

PR #10351, opened 2026-07-27
feat: show per-owner runnable counts in the homepage tree (WIN-2253)

rubenfiszel:
Fixes WIN-2253

## What

The homepage tree listed **every** folder and user in the workspace as a root node, with a **blank** subtitle until you expanded one (its true size was unknowable without loading it). In a workspace with hundreds of folders, that is mostly noise: you scroll past folders you can see nothing in to reach the two you use.

This adds a per-owner count, fetched in one request, so the tree can label every node up front and drop the owners holding nothing.

## How

New endpoint `GET /w/{workspace}/runnables/counts?kinds=&include_without_main=` returning `{ "counts": { "f/analytics": 5, "u/admin": 2 } }`. Owners with nothing visible are omitted.

**It does not go through RLS.** RLS evaluates `split_part(path,'/',2) = ANY(regexp_split_to_array(current_setting('session.folders_read'), ','))` per row, which no index can serve, so any count under it degrades to a full workspace scan. Instead the endpoint runs on the plain pool and re-derives visibility from `path`:

- an owner is readable **whole or not at all** — admin, `f/<folder>` in the caller's read set, or the caller's own `u/<user>` space;
- its count is a byte-ordered prefix range, `path ~>=~ owner || '/' AND path ~<~ owner || '0'`.

That range is exact: every path under `owner` is `>= owner || '/'`, and since `'/'` (0x2F) is immediately followed by `'0'` (0x30) in ASCII, all of them are `< owner || '0'`. `~>=~` / `~<~` are the `text_pattern_ops` operators, which is what makes it indexable — the default opclass sorts by the database collation and cannot answer a byte-prefix range.

Admins are the exception: they read everything, so their owner set would be every folder *and* every member, and one prefix scan each would scale with headcount rather than with content. For them it is a single grouped scan per kind over the same indexes instead.

Hence the new migration:

- `idx_script_owner_prefix` — `(workspace_id, path text_pattern_ops) INCLUDE (auto_kind) WHERE archived = false`; `auto_kind` is an `INCLUDE` column so the library-script filter is answered from the index tuple and the scan stays index-only
- `idx_flow_owner_prefix` — same, `WHERE archived = false`
- `idx_app_owner_prefix` — same, no `archived` column on `app`
- `app_extra_perms` — GIN, matching the existing `script_extra_perms` / `flow_extra_perms`

The one thing `path` cannot express is an item **shared individually** out of an owner the caller is otherwise not a member of. A second pass recovers exactly those: one `extra_perms ?| ARRAY['u/<me>', 'g/<group>', …]` GIN scan per kind, grouped by owner. Owners already counted in pass 1 are readable whole, shares included, so merging by owner cannot double-count.

Fine-grained token scopes (`scripts:read:f/foo/*`) are pushed into both passes; the predicate builder is the one `list_runnables` already used, lifted out of its closure so both handlers share it.

### Frontend

`ItemsList` fetches the counts once per (workspace, kind, library, archived) scope and passes them down. `TreeView` shows the count instead of the blank placeholder. `TreeViewRoot` stops injecting an owner whose count is 0 — except pipeline folders, whose entry is the pipeline itself and whose member scripts are folded out of the runnable count. Owners that appear only in the counts (an individually-shared item in someone else's space) are now injected too, so they are reachable in the tree at all; previously they only showed up via search.

Counts are skipped in the archived view (the endpoint counts the live set) and outside lazy tree mode, where the tree groups already-loaded rows and its own counts are accurate. If the request fails the tree falls back to the previous behaviour.

The counts and the folder list are separate requests, so the tree holds its existing skeleton until the first counts response rather than rendering from whichever lands first — otherwise a slow counts response would draw every workspace folder and then prune it away. Both are issued in parallel with the listing, so in practice this costs no extra wait (in a local trace the counts landed 25 ms *before* the listing the tree was already waiting on). Only the first load gates: a scope change refreshes the counts in place without flashing the skeleton over a drawn tree.

The label counts what the node renders, so a folder with a data pipeline adds one for its Pipeline row — a folder holding a 5-step pipeline and one script reads "(2 items)" and expands to exactly the Pipeline row plus that script.

## Validation

`cargo check` clean, `npm run check` clean (0 errors). No new compile-time-checked queries, so the sqlx offline cache is unaffected.

**Correctness** — a synthetic workspace (500 folders, 100 user spaces, 53k runnables) with a non-admin `alice` holding: read on 2 folders, her own space, 30 scripts shared to `u/alice` in another user's space, and 20 apps shared to a group she belongs to. The endpoint returned exactly the same owner→count map as the equivalent query run under a real RLS session:

```
 f/fold1  | 100      {"f/fold1":100,"f/fold2":100,"f/fold42":20,
 f/fold2  | 100       "u/alice":1,"u/user7":30}
 f/fold42 |  20
 u/alice  |   1      RLS ground truth (left) vs /runnables/counts (right)
 u/user7  |  30
```

**Performance** — same fixture, `EXPLAIN ANALYZE`, warm:

| | |
|---|---|
| non-admin path, per-owner prefix counts (502 owners) | **7.3 ms** — Index Only Scan on all three tables |
| admin path, one grouped scan per kind | **9.3 ms** — Index Only Scan on all three tables |
| equivalent grouped count under RLS | 68 ms — Seq Scan per table |
| `extra_perms` share pass | 0.12 ms — Bitmap Index Scan on the GIN indexes |

**Manual** — driven in the browser with Playwright as both an admin and a non-admin; expanded folders load the items the count promised.

**Regression test** — `backend/tests/runnables_owner_counts.rs` asserts the endpoint's owner map equals the per-owner grouping of `/runnables/list` (which is RLS-enforced) for both a non-admin and an admin, over a fixture covering a folder grant, an own user space, a direct share and a group share out of an unreadable folder, a pipeline member, an operator, and a duplicated `kinds` value. It fails on each of the round-1 findings below.

## Screenshots

Admin, small workspace — `f/archive_2024` and `f/scratch` hold nothing and are gone; the rest are labelled.

| Before | After |
|---|---|
| before-admin-demo | after-admin-demo |

Non-admin in the 500-folder workspace. Before: four unlabelled nodes, one (`u/admin`) she can see nothing in, and her shared items unreachable. After: exactly the five owners she has something in — including `u/user7` (shared to her directly) and `f/fold42` (shared to her group), neither of which is in her folder list.

| Before | After |
|---|---|
| before-alice-perf | after-alice-perf |

A folder whose only content is a data pipeline: the member scripts are folded into the single Pipeline entry, so they are not counted as items and the folder is exempt from the empty-owner drop.

pipeline folder

## Review round 1

- **Pipeline members inflated the count** (Claude + Codex, P1) — the script predicate excluded only `auto_kind = 'lib'`. A folder holding a 6-step pipeline read `(6 items)` and expanded to a single Pipeline row. `'pipeline'` is now always excluded; the folder keeps its node (`TreeViewRoot` already exempts pipeline folders from the empty-owner drop).
- **Operator counts omitted flows and apps** (Codex, P1) — the handler copied a `kinds.retain(script)` that `main` has since removed from `list_runnables`. Rebased onto `main` and dropped it; only library scripts stay hidden from operators.
- **Duplicate `kinds` double-counted** (Codex, P1) — `kinds=script,script` built the subquery twice. The list is now deduplicated.
- **Pass 1 scaled with headcount for admins** (Claude, P2) — replaced with a single grouped scan per kind, see above.
- **The caller's own `u/<me>` node could vanish** (Claude, P2) — it was dropped when empty, breaking the invariant that your personal space is always in the tree. Now exempt from the drop.
- **The share pass's GIN bitmap is not workspace-scoped** (Claude, P2) — accepted and documented in the handler. Scoping it needs `btree_gin`, a contrib extension self-hosted installs can't be assumed to have, and the RLS policies already scan these same indexes the same way; `workspace_id` remains a recheck.

## Review round 2

Codex and Pi: "Good to merge". Claude: nit-only.

- **A mixed pipeline folder labelled one short** — the Pipeline row is rendered but its member scripts are excluded from the count. The label now adds one where that row renders, so it matches the expanded contents.
- **An empty own user space rendered blank rather than `(0 items)`** — an owner missing from the response holds nothing, so it now reads as 0 instead of unknown.

Also in this round, unprompted by the reviewers: the two-request reflow described above.

## Review round 3

Claude and Pi: "Good to merge". Codex: **cached closed owners bypass per-scope pruning** — confirmed by reproducing it.

Expand `f/billing`, collapse it, switch the kind filter to Apps: the folder has no apps, so the counts omit it and it should disappear — instead it stayed, labelled `(0 items)`, and expanding it showed a flow and a script from the previous scope. `reloadItems` re-fetches only the *open* owners and `treeKey` doesn't include `itemKind`, so a collapsed owner's rows survive into a scope they were never loaded for. It then appears in `grouped`, which short-circuits the empty-owner check, and stays marked `loaded`, so re-expanding never re-fetches.

Fixed by dropping the cached rows and load state of every owner that isn't currently open, since those are exactly the ones the reload skips. Applying the counts to already-grouped roots instead would only have fixed the zero case — an owner that *does* have items in the new scope would still expand to the old scope's rows under a correct count. The stale rows predate this PR (a collapsed owner's cache was never invalidated on a scope change); the count only turned a quiet inconsistency into a visible contradiction.

Verified after the fix: `f/billing` is dropped, re-expanding `f/analytics` under Apps lists only the dashboard app, and an owner left *open* across a scope change still stays open and refreshes in place.

## Review round 4

Claude and Pi: "Good to merge". Codex: **count-pruned owners stay registered as open** — a follow-on from the round-3 fix, and confirmed by reproducing it.

Dropping an owner destroys its `TreeView` without any collapse click, so its key stayed in `openOwners` and every later reload re-fetched something no longer on screen. Measured on the demo workspace — expand `f/billing` and `f/shared_infra`, switch to Apps so both are pruned, then keep changing kind:

| | before | after |
|---|---|---|
| switch to Apps (both still open when the reload starts) | 2 requests | 2 requests |
| then switch to Flows | 2 requests for hidden owners | **0** |
| then switch back to Apps | 2 requests for hidden owners | **0** |

Fixed with an `onDestroy` in `TreeView` that untracks the owner, so the reload fan-out follows what is actually rendered. This also covers an expanded owner sliced out of the rendered window by `nbDisplayed`. An owner left *open* across a scope change still refreshes in place and stays open — verified separately.

## Review round 5

Codex, Claude and Pi: all "Good to merge". No findings.

### DIFF
--- a/backend/migrations/20260727093955_runnable_owner_prefix_indexes.down.sql
+++ b/backend/migrations/20260727093955_runnable_owner_prefix_indexes.down.sql
@@ -1,0 +1,4 @@
+DROP INDEX IF EXISTS idx_script_owner_prefix;
+DROP INDEX IF EXISTS idx_flow_owner_prefix;
+DROP INDEX IF EXISTS idx_app_owner_prefix;
+DROP INDEX IF EXISTS app_extra_perms;
--- a/backend/migrations/20260727093955_runnable_owner_prefix_indexes.up.sql
+++ b/backend/migrations/20260727093955_runnable_owner_prefix_indexes.up.sql
@@ -1,0 +1,24 @@
+-- Indexes backing the homepage tree's per-owner runnable counts
+-- (/runnables/counts), which counts each `f/<folder>` / `u/<user>` prefix with a
+-- byte-ordered range `path ~>=~ owner || '/' AND path ~<~ owner || '0'`.
+-- text_pattern_ops is required: the default opclass sorts by the database
+-- collation, so a byte-prefix range cannot use it. `auto_kind` is an INCLUDE
+-- column so the library-script filter is answered from the index tuple and the
+-- count stays an index-only scan.
+-- Created CONCURRENTLY via the OVERRIDDEN_MIGRATIONS rewrite in windmill-api/src/db.rs.
+CREATE INDEX IF NOT EXISTS idx_script_owner_prefix
+    ON script (workspace_id, path text_pattern_ops)
+    INCLUDE (auto_kind)
+    WHERE archived = false;
+
+CREATE INDEX IF NOT EXISTS idx_flow_owner_prefix
+    ON flow (workspace_id, path text_pattern_ops)
+    WHERE archived = false;
+
+CREATE INDEX IF NOT EXISTS idx_app_owner_prefix
+    ON app (workspace_id, path text_pattern_ops);
+
+-- Same purpose as the existing script_extra_perms / flow_extra_perms GIN
+-- indexes: the counts endpoint finds items shared explicitly with the caller
+-- (or one of their groups) with a single `extra_perms ?| ARRAY[...]` scan.
+CREATE INDEX IF NOT EXISTS app_extra_perms ON app USING gin (extra_perms);
--- a/backend/tests/runnables_owner_counts.rs
+++ b/backend/tests/runnables_owner_counts.rs
@@ -1,0 +1,195 @@
+//! Regression test for `GET /w/{workspace}/runnables/counts`, which the homepage
+//! tree uses to label every folder / user node and to drop the empty ones.
+//!
+//! The endpoint runs off the non-RLS pool and re-derives visibility from `path`
+//! (admin / folder read set / own user space) plus one `extra_perms` pass, so it
+//! duplicates by hand what RLS otherwise enforces. That duplication is what can
+//! drift: a change to the RLS policies, or to how `authed.folders`/`groups` are
+//! populated, would silently make the counts over- or under-report. So the test
+//! pins them to the ground truth rather than to their own implementation — the
+//! counts must equal the per-owner grouping of `/runnables/list`, which is
+//! RLS-enforced, across every way an item can become visible.
+
+use sqlx::{Pool, Postgres};
+use std::collections::HashMap;
+use windmill_test_utils::*;
+
+fn client() -> reqwest::Client {
+    reqwest::Client::new()
+}
+
+fn authed(builder: reqwest::RequestBuilder, token: &str) -> reqwest::RequestBuilder {
+    builder.header("Authorization", format!("Bearer {}", token))
+}
+
+fn owner_of(path: &str) -> String {
+    path.split('/').take(2).collect::<Vec<_>>().join("/")
+}
+
+/// Per-owner counts derived from the RLS-enforced listing. Pipeline members are
+/// dropped to match what the tree renders (and what /counts promises): they are
+/// folded into their folder's single pipeline entry, never listed as rows.
+async fn counts_from_list(port: u16, token: &str) -> HashMap<String, i64> {
+    let base = format!("http://localhost:{port}/api/w/test-workspace/runnables/list");
+    let mut out: HashMap<String, i64> = HashMap::new();
+    let mut cursor: Option<String> = None;
+    loop {
+        let mut url = format!("{base}?per_page=100");
+        if let Some(c) = &cursor {
+            url.push_str(&format!("&cursor={c}"));
+        }
+        let resp = authed(client().get(&url), token).send().await.unwrap();
+        assert_eq!(resp.status(), 200, "list should succeed");
+        let body: serde_json::Value = resp.json().await.unwrap();
+        for it in body["items"].as_array().unwrap() {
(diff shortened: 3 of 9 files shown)


rubenfiszel:
✅ Review round clean @ f80dadd591f61c92e8401edadfbe3bcfb2057934

PR #10353, opened 2026-07-27
fix(cli): sync push no longer reports success on a script it never deployed

rubenfiszel:
## Summary

`wmill sync push` could print `Done! All N changes pushed`, exit 0, and leave the script absent from the remote — silent data loss in the primary deploy path, and a CI gate on the exit code goes green on a deploy that never happened.

The apply loop skipped every *added* `.script.yaml` / `.script.json` / `.script.lock` outright, on the assumption that the sibling content file in the same group carried the deploy. When the content file is not in the changeset (for instance filtered out of the sync by `excludes`), nothing was sent to the remote, yet the metadata file was still counted in the "N changes pushed" total. The `edited` branch never had this hole — it already routes metadata changes through `handleScriptMetadata`, which resolves the content file from disk.

The `added` branch now does the same. The deploy stays idempotent (`handleFile` short-circuits on `alreadySynced`, so a group holding both the content file and its metadata still deploys exactly once, in either processing order).

A metadata file that cannot be paired with exactly one script file is unpushable — either none exists, or several do and nothing says which one the metadata describes. Rather than aborting mid-flight, which would leave the push partially applied and drop unrelated changes queued behind it, the push records the failure, logs it, applies everything else, and reports `N of M changes pushed; K failed` with a **non-zero exit**. Only this content-resolution failure is soft (`UnresolvableScriptContentFileError`); a deploy the remote rejected still aborts, since a server rejection says nothing about whether the remaining changes are safe to apply.

Before / after, with a `.script.yaml` whose `.py` sibling is kept out of the sync by `excludes`:

```
# before
remote (test2) <- local: 1 changes to apply
+ script f/test/excluded.script.yaml
Done! All 1 changes pushed ...          # exit 0, script absent on the remote

# after
+ script f/test/excluded.script.yaml
Creating new script f/test/excluded ...
Created new script f/test/excluded (4ms)
Done! All 1 changes pushed ...          # exit 0, script deployed
```

With an unpushable script alongside two healthy ones that sort either side of it:

```
Created new script f/test/alpha (5ms)
Multiple script files found next to f/test/both.script.yaml: f/test/both.ts, f/test/both.py —
cannot tell which one the metadata belongs to. Keep exactly one.
Created new script f/test/zeta (4ms)

10 of 11 changes pushed to the remote workspace test named test; 1 failed:
  - f/test/both.script.yaml
# exit 1 — alpha and zeta are deployed, both is not
```

`--json-output` mirrors this with `"success": false` and a `failed` array.

## Changes

- `cli/src/commands/sync/sync.ts`: the `added` branch routes script metadata/lock changes through `handleScriptMetadata` before the blanket skip, mirroring the `edited` branch.
- `cli/src/commands/sync/sync.ts`: unresolvable script changes are collected into `failedChanges` instead of throwing; the final report becomes `N of M changes pushed; K failed` and the command exits non-zero (`success: false` + `failed[]` under `--json-output`).
- `cli/src/commands/sync/sync.ts`: that exit uses `process.exitCode` rather than `process.exit`, so a large `--json-output` payload is not truncated at the pipe buffer.
- `cli/src/commands/script/script.ts`: both of `findContentFile`'s unpairable branches (no candidate, several candidates) raise `UnresolvableScriptContentFileError` with a message naming the expected or clashing files; inputs that are neither flat nor module-entry metadata are rejected up front rather than silently resolving to themselves.
- `cli/src/commands/script/script.ts`: `handleScriptMetadata` and `findContentFile` gate module-folder metadata on `isModuleEntryPoint` rather than `isScriptModulePath`, so a module file nested deeper (`f/foo__mod/config/script.yaml`) is not mistaken for the script's metadata.
- `cli/test/sync_push_script_metadata_only.test.ts`: the metadata-only change reaches the remote (asserting deployed *content*, not just exit 0); an unpushable script fails the exit code without dropping the rest of the changeset.
- `cli/test/script_modules_unit.test.ts`: `findContentFile` resolves a module entry point, rejects a `script.yaml` nested below it, and signals every unpairable case with the class `sync push` catches.

## Test plan

- [x] Both integration regression tests pass; the first fails on `main` (push exits 0, script 404s).
- [x] Each new unit test verified to fail with its fix reverted.
- [x] `bun test test/*_unit*.test.ts` — 979 pass, 0 fail.
- [x] `bun test` across `sync_pull_push`, `sync_push_auto_metadata_repro`, `script_envs_sync`, `raw_app_sync`, `folder_missing_meta`, `gitsync_promotion`, `variable_resource_push` (+ the new file) — 94–105 pass, 0 fail.
- [x] `bun test test/folder_default_permissioned_as.test.ts test/labels_export.test.ts test/override_settings_behavior.test.ts test/folder_schedule_push.test.ts test/app_inline_script_delete.test.ts` — 50 pass, 0 fail.
- [x] Manual push against a local backend: flat layout, `__mod/` folder layout, a `__mod/config/script.yaml` module file, and both unpushable cases (missing and ambiguous) in text and `--json-output` mode.
- [x] Truncation check: a 1904-change `--json-output` push through a pipe emitted 432 KiB of valid JSON and exit 1; with `process.exit` it was cut off at exactly 65536 bytes.
- [x] `bunx tsc --noEmit` — no new errors in the touched files (the repo has pre-existing errors elsewhere).
- [x] `relative_imports_skip.test.ts` fails 4 Python lockgen tests identically with and without this change (local backend binary built without the `python` feature).

## Known limitations

- On Windows the folder-layout half of the routing is inert: `handleScriptMetadata`'s module-entry check compares against `/script.yaml` while `change.path` is `\`-separated. The flat `.script.yaml` case (the reported bug) is separator-independent and works. Pre-existing, not addressed here.
- `sync.ts` has other pre-existing `process.exit(1)` call sites, one of which (`missing_folders`) prints JSON first. Their payloads are bounded well below the pipe buffer, so they are left alone rather than widening this diff.

Fixes WIN-2254

### DIFF
--- a/cli/src/commands/script/script.ts
+++ b/cli/src/commands/script/script.ts
@@ -887,6 +887,14 @@ export async function findContentFile(filePath: string) {
       : filePath.endsWith("script.lock")
       ? filePath.replace(".script.lock", ext)
       : filePath.replace(".script.yaml", ext);
+  // Every branch above is a no-op on a path that is neither flat nor
+  // module-entry metadata, which would make toCandidate the identity function
+  // and "resolve" the input to itself.
+  if (!isModuleFolderMeta && !/\.script\.(yaml|json|lock)$/.test(filePath)) {
+    throw new Error(
+      `${filePath} is not a script metadata file — no script file can be resolved from it.`
+    );
+  }
   const candidates = exts.map(toCandidate);
 
   const validCandidates = (
--- a/cli/src/commands/script/script.ts
+++ b/cli/src/commands/script/script.ts
@@ -875,12 +875,12 @@ async function createScript(
 }
 
 /**
- * A script metadata file could not be paired with a script file on disk, so
- * nothing can be deployed for it. Distinct from a deploy that reached the
- * remote and was rejected: callers that can carry on with the rest of a
+ * A script metadata file could not be paired with exactly one script file on
+ * disk, so nothing can be deployed for it. Distinct from a deploy that reached
+ * the remote and was rejected: callers that can carry on with the rest of a
  * changeset catch this specifically.
  */
-export class MissingScriptContentFileError extends Error {}
+export class UnresolvableScriptContentFileError extends Error {}
 
 export async function findContentFile(filePath: string) {
   // Folder layout: __mod/script.yaml -> __mod/script.ts
@@ -899,7 +899,7 @@ export async function findContentFile(filePath: string) {
   // module-entry metadata, which would make toCandidate the identity function
   // and "resolve" the input to itself.
   if (!isModuleFolderMeta && !/\.script\.(yaml|json|lock)$/.test(filePath)) {
-    throw new MissingScriptContentFileError(
+    throw new UnresolvableScriptContentFileError(
       `${filePath} is not a script metadata file — no script file can be resolved from it.`
     );
   }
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ 92704bb1277d3a1782f7b13bb7665df84e8461ce

PR #10354, opened 2026-07-27
fix: edit-in-dev-workspace dead-ends, wraps, and misses the tree view

Guilhem-lm:
## Summary

Fixes the **"Edit in \<dev workspace\>"** affordance — the action that replaces **Edit** on item rows when a workspace has direct edits locked (`DisableDirectDeployment`). Six separate problems, from "the button is unusable" up to "the button goes nowhere":

1. **It dead-ended.** The destination URL was built at render time, so for any item prod has and the dev workspace doesn't, the button landed on a not-found page. A dev workspace is routinely behind its prod, so this hit a lot of items.
2. **The label wrapped to two lines**, blowing out the row height. The button had a hard `w-32` (128px) — even a 7-character workspace name doesn't fit — and the label was a bare text node, which has nothing to apply `text-overflow` to, so it could only wrap, never truncate.
3. **The tooltip only covered the label text**, leaving the icon and padding with no way to read a truncated name.
4. **The tree view never showed the button at all** — and instead showed a stale **Edit** linking straight into the locked workspace. `showEditButton` was never threaded from `ItemsList` through `TreeViewRoot` → `TreeView` → `Item`, so tree rows fell back to the `true` default.
5. **It didn't look like an Edit button** — fork icon in the lead position and a `default` variant in locked workspaces, so the row's primary action read as something other than "edit".
6. **It promised an update it couldn't deliver.** Once the prompt below existed, a user without deploy rights in the dev workspace got the full "Update and edit" flow and a 403 toast on confirm. Deploy rights are per *path*, not just per workspace, so this is not answerable from the workspace alone.

### Resolve the destination at click time

`editInFork.ts` grows `onEditInForkClick` (links) and `openEditInFork` (editor dropdowns, which open a new tab). Both check whether the item exists in the dev workspace and either navigate there or raise a prompt. The destination is the item's **editor** in the dev workspace (`/scripts/edit/…?workspace=<dev>`, and the equivalent for flows and both app kinds) — the button says "Edit", so it lands in the editor; `?workspace=` flips the workspace store on the way in. Modifier and middle clicks fall through to the raw `href` so open-in-new-tab still works, and an inconclusive check navigates anyway rather than blocking — the item page reports whatever is actually wrong.

The probe outlives the click, so the result is discarded unless the user is still where they asked from: a sequence counter drops a superseded click, and a pathname or workspace change abandons the answer entirely (the prompt is layout-global and `goto` is unconditional, so a late answer would otherwise hijack whatever the user moved on to). In `openEditInFork`, a `window.open` blocked because the probe outlived the click's transient activation now says so, rather than falling back to navigating in place and discarding whatever that editor was holding.

`UpdateDevWorkspaceModal.svelte` is that prompt, mounted once in the logged layout (state held in `editInForkModal.svelte.ts`) rather than per row. It offers **Update and edit** — deploying the single item across — or a link to the compare page to pull it alongside everything else the workspace is behind on. A failed update keeps the modal open, since the compare page is the way out when a lone item can't stand on its own (missing resource, resource type).

It speaks the **compare page's vocabulary**, because it is doing the compare page's `update` direction for one item: *"guilhem is behind on this item"*, *"Update guilhem with it"*, **Update and edit**. Calling the same operation a "copy" here and an "update" one click away made them read as two different mechanisms.

**Folders come along.** Deploying `f/<folder>/<name>` into a workspace that has no `<folder>` *succeeds* and leaves the item orphaned — it lands with no folder to carry its permissions, and the deploy reports success. The compare page never hits this because its diff lists the folder as its own item and deploys `folder:` entries first; a single-item update has to bring the folder itself, so `ensureFolder` deploys it when the target lacks it. That probe fails **closed** — deploying while the folder is in fact missing is the orphaning above, and nothing downstream would catch it. Anything else the item needs (resources, variables, resource types) is still the compare page's job, exactly as it is there.

Immediately before the write, the prompt re-checks that the item is still absent. The prompt is only up *because* it was absent, and the shared deploy silently switches to an update once it isn't — overwriting whoever landed it meanwhile. This narrows the window to one request rather than closing it; only a create-only write on the server can do that. An inconclusive re-check also refuses to write and leaves the prompt up for a retry.

### Check deploy rights per item, before offering the update

`checkItemDeployAccess(workspace, path)` answers, off a single `whoami` in the **target** workspace:

- the workspace-level gate (`checkDeployPermission`, shared with the session dock and compare page): operators can never deploy, and `RestrictDeployToDeployers` limits the rest to admins, `wm_deployers`, and per-ruleset bypasses;
- **write access to the item's own path** — the half the workspace-level rules don't cover. `u/<someone-else>/…` is refused; `f/<folder>/…` is refused when the caller isn't a writer of that folder, *unless* the folder is absent, since the deploy creates it with the deployer as owner;
- whether another identity may be preserved;
- who the caller *is in that workspace* — usernames are per-workspace, so the local one is the wrong name for a different target.

When it refuses, the prompt swaps the explanation for a warning and the button becomes **Open compare page**, deep-linked to that item. Every probe fails open, since the server still enforces on the actual write; the point is to stop offering an action that will 403.

### The identity is a choice, not a silent default

An item running on behalf of someone in prod would otherwise become the deploying user's on arrival. The prompt offers the same control the compare page does, under the same rules: shown only when the source item carries an `on_behalf_of` (`needsOnBehalfOfSelection`), picking anyone but yourself gated on admin/`wm_deployers` in the target, nothing preselected, and confirming blocked until a choice is made. There is no "keep the target's value" option here — the item is absent from the dev workspace, which is the premise of the prompt.

Three details that are easy to get wrong:

- **"me" is sent explicitly.** Sending nothing means "no preference", which lets the target folder's `default_permissioned_as` claim the item — so the option labelled *me* would have deployed it as somebody else.
- **The source's identity must be readable.** `getOnBehalfOfOrThrow` distinguishes "carries no identity" from "could not read it"; conflating them silently reassigns the item to the deploying user. Either failure — that one, or not knowing who *you* are in the target — raises an error alert and blocks the update rather than guessing.
- **Apps keep their identity in the app policy**, not a top-level field, and the shared deploy forwards the source policy untouched. Rewriting the policy on the way out is the only way a chosen identity reaches an app. This also closes the same gap on the **compare page**, which calls the same `deployItem` and previously discarded the pick for apps and raw apps (an earlier revision of this description listed that as a known gap; it is fixed here).

Both lookups are driven by a cancelling `$effect` rather than a `resource`, which assigns from whichever fetch settles last: requests outlive the prompt that started them, so cancelling and reopening otherwise let a stale answer decide this item's permissions. Results are tagged with the request they answer, and confirming is blocked until both land *for this item* — **Enter** is bound to confirm, so a fast one would deploy past the permission check and skip a required choice.

Reusing the compare page's `OnBehalfOfSelector` inside a dialog needed several fixes:

- `ConfirmationModal` takes a `confirmDisabled` prop, which also blocks its Enter binding.
- Its Enter/Escape handling is now popover-aware, per key: Enter always belongs to a popover (it both opens the trigger and chooses from the content), while Escape belongs to it only while it is *open* — otherwise Escape on a closed trigger would stop dismissing the dialog. Open/closed is read from `aria-controls` resolving to a mounted element, because the trigger's own `aria-expanded`/`data-state` are stale under `forceVisible`.
- The user picker reports its open state (`onPickerOpenChange`, fired from an effect so every close path counts) and the host suspends `keyListen` meanwhile: both handlers sit on `window`, where `stopPropagation` can't separate them.
- The popover's `z-[5001]` is now overridable via `contentClasses` (it was merged last, so callers couldn't win), and the user picker is portalled. A `ConfirmationModal` sits at `z-[9999]`, above `zIndexes.popover`, and its card is `transform`ed for the open transition — which makes it the containing block for the picker's `fixed` positioning. So both opened *behind* the dialog, and the picker was laid out inside the card rather than the viewport.

Supporting pieces: `Button.svelte` now calls `onClick` in its `<a>` branch too (a link button can legitimately want to intercept its own click); `CompareWorkspaces` gains `maskAppliesToUpdate` so an explicit `?items=` deep link preselects on the update direction, which a chat session's mask must not.

`(root)/+layout.svelte` also retries the workspace-list load. It's the only load for the session, and an empty `$userWorkspaces` degrades *silently* — the affordance falls back to a generic "Edit in fork", the no-direct-deploy alert loses its dev-workspace wording, the fork banner disappears. A transient failure shouldn't strand the tab in that state.

### Presentation

New `common/table/EditInForkButton.svelte` — the markup was identical in all three row components and got involved enough (conditional label, truncation, tooltip wrapper, click interception) that triplicating it wasn't worth it. `ScriptRow`, `FlowRow`, and `AppRow` now render `<EditInForkButton {itemType} {path} />`, and their now-unused `buildForkEditUrl` imports are gone. Its `href` is derived from the reactive workspace stores rather than from `buildForkEditUrl`, whose store reads are untracked — the href would otherwise stay frozen at mount while the label kept updating, so a row that outlived a workspace change offered to fork a workspace that already had a dev.

- **Styling matches the normal Edit button**: `variant="subtle"` unconditionally, and the **pen** as the icon — the button reads `[pen] Edit in guilhem`.
- **No fork icon anywhere on this action.** It was on the row button (alongside the pen), on the row `…` menu entries, and on the detail- and run-page buttons. In the menus it collided with **Duplicate/Fork**, which uses the same icon, making the two entries indistinguishable. The action is an edit, so it takes the pen at every site.
- **No wrapping**: `max-w-56` lets the button hug its content and cap out, and only the workspace name carries `truncate`, so `Edit in` always stays intact.
- **Tooltip on the wrapper `<div>`**, not on `<Button>` or the span. `Button.svelte` declares `title` as an explicit prop and renders it only in the `<button>` branch, so `<Button title={...}>` on this href-button was a silent no-op; the wrapper also covers the icon and the padding.
- **Fallback** (no canonical dev workspace) renders `[pen] Edit in fork`.

### Tree view

`TreeViewRoot` and `TreeView` gain a `showEditButton` prop (defaulting to `true`, matching `Item`), threaded through the recursive call into `Item`, and wired at the `TreeViewRoot` call site — mirroring what the flat list already did.

## Relationship to recent `main`

This branch is merged with `main` as of `9f3d15583a`. Two pieces of it landed there independently and this branch now defers to them rather than carrying a parallel version:

- **#10438** (`fix: make on_behalf_of control permissions for scripts and flows`) introduced `DeployItemParams.onBehalfOfPrincipal` and a `withPermissionedAs` wrapper for flow/script bodies — the same concept this branch had added as `onBehalfOfPermissionedAs`. Main's name and contract win; the app-policy rewrite described above sits on top, since #10438 left the app branch alone.
- **`?dir=update`** on the compare page (added for the fork banner's CTA) replaces the `?direction=update` this branch had introduced for the same purpose.

`ConfirmationModal` also picked up main's `ConditionalPortal`/overlay-host anchoring (#10477); the keyboard changes here are layered on that.

## Screenshots

Prod workspace (`ai-meta-e2e`) with a canonical dev workspace (`guilhem`), direct edits locked.

**Before** — `Edit in guilhem` wraps to two lines, fork icon leading, `default` variant:

before-wrap

**After** — `[pen] Edit in guilhem`, subtle, one line, no fork icon:

style-list

**Clicking an item the dev workspace doesn't have**, in the tree view — previously a not-found dead end, and before this PR these tree rows showed a plain **Edit** linking into the locked workspace. (Screenshot predates the icon and copy changes above; the prompt's current wording is in the next shot.)

update-modal

**The on-behalf-of choice**, on an item whose prod version runs on behalf of someone. Confirming is disabled until a user is picked:

on-behalf-of

**Long dev workspace name** (`Guilhem Development Sandbox Workspace`) — only the name ellipsizes:

style-long

**Fallback**, no canonical dev workspace:

style-fallback

**Tree view**, rows showing the affordance. Recaptured on top of #10519, which rewrote this component — before this PR these rows had no way into the dev workspace at all:

tree-view

**No deploy rights on the target**, as a non-admin whose account can read `f/shared` in the dev workspace but not write to it. The confirm button becomes a redirect rather than an action that would 403:

refusal

**Where that redirect lands** — the compare page in the update direction with just that item selected, so the request still has somewhere to go:

compare-preselected

The hover tooltip is an OS-rendered native `title` and can't be screenshotted; it was verified against the DOM instead.

## Test plan

No automated test is added: the presentation changes are a Tailwind/markup swap and a prop-threading fix, and the click-resolution logic is a thin layer over `checkItemExists` / `deployItem`. Verified in the browser via Playwright MCP.

The most recent full pass was 26 checks against `EE v1.778.0-12-g9f3d15583a`, so the server half of #10438 — which derives the on-behalf-of principal from the email and rejects mismatched pairs — was actually exercised. Every result was read back from the DOM or from Postgres after the action rather than inferred from the click succeeding.

After merging `main` (through `055a9c2690`, `EE v1.780.0`), the merge-sensitive checks were re-run: #10519 rewrote `TreeView.svelte`, which is where this PR threads `showEditButton`, and the merge conflicted there. The prop still reaches the recursive `TreeView` and the leaf `Item`, the tree still shows the affordance on every row, and clicking a prod-only item still raises the prompt with its identity picker and a disabled confirm. `npm run check`: 0 errors over 10,080 files.

**Destination**

- [x] Item present in both workspaces → click navigates straight to the item's editor in the dev workspace, and the workspace store switches
- [x] Item in prod only → click is intercepted, URL unchanged, **"dev is behind on this item"** prompt; confirming deploys it and lands in that item's editor in the dev workspace
- [x] Every rendered button is an editor URL carrying `?workspace=<dev>` — 45 buttons across `scripts`/`flows`/`apps_raw` in one expanded owner: 45 editor URLs, 0 non-editor, 0 missing the param
- [x] ⌘-click opens the raw `href` in a new tab and leaves the current page untouched — the interception defers to the anchor
- [x] Same anchor, plain click → raises the prompt instead
- [x] Folder-scoped item whose folder is missing in the dev workspace → the folder is created there with its owners *and* the item lands in it; without `ensureFolder` the item arrives orphaned and the update still reports success
- [x] Raw app whose bundle has no stylesheet → updates and opens in the dev workspace. Needs a backend containing #10364

**Permissions** (non-admin member of both workspaces)

- [x] No write access to the item's folder in the dev workspace → prompt shows **"You don't have write access to folder shared"**, button becomes **Open compare page**, no on-behalf-of selector
- [x] Operator in the dev workspace → **"You're an operator in this workspace — operators can't deploy"**, and the workspace-level reason takes precedence over the path one
- [x] **Open compare page** lands on `/forks/compare?workspace_id=…&mode=fork&dir=update&items=…`, on the **Update current (1)** side with that item preselected

**On-behalf-of**

- [x] Source item with no `on_behalf_of` → no selector, confirm enabled, deploys
- [x] Source item with one → selector shown, confirm **disabled**, warning displayed
- [x] **Enter** while unset does nothing — modal stays open, no create/update request fired
- [x] **Enter** once a choice is made confirms and deploys
- [x] Choosing another user → the deployed **app**'s policy carries that user in both halves (`u/e2edev` / `e2e-dev@windmill.dev`) while the source keeps `u/admin` / `admin@windmill.dev` — i.e. the pick reaches an app, which is what the policy rewrite is for
- [x] The *me* option is labelled with the caller's username **in the target workspace**
- [x] Escape with the popover open closes only the popover; Escape on the closed trigger closes the dialog; Escape in the stacked user picker closes only the picker
- [x] Popover and picker both render above the dialog, and the picker centres on the viewport rather than inside the dialog card
- [x] No regression on the compare page: it still renders, and its `?items=` + `dir=update` deep link preselects correctly

**Presentation**

- [x] No fork icon on this action anywhere: list view 15/15 pen and 0 fork, tree view 45/45 pen and 0 fork, `…` menu shows pen for **Edit in guilhem** next to fork for **Duplicate/Fork**, script detail page pen
- [x] List view: subtle (computed style has no background or border), one line
- [x] Long dev workspace name: 224px (`max-w-56`), 32px tall, `scrollWidth > clientWidth` on the name span only — `Edit in` never truncates
- [x] Full label resolves from every probe point on the button, so the tooltip covers the whole control
- [x] No canonical dev workspace: falls back to `[pen] Edit in fork`, un-truncated
- [x] Tree view, expanded owner node: 18 affordances / 0 plain Edit buttons after, vs 0 / 18 before — the "before" hrefs pointed into the locked workspace

That pass found one real bug, fixed in `780f483d40`: **Open compare page** dismissed the dialog and went nowhere. `confirm()` closed the prompt before reading the compare link, which is derived from the request being answered — so it recomputed to the empty string and `goto('')` was a no-op. Refused users were left exactly where they started with no way through.

Worth a reviewer's attention:

- Twice, early in that session, a first click after login fired the existence probe and then did nothing — no navigation, no prompt, which is the shape of the click-time guard bailing. It did not reproduce across ~10 later clicks, and sampling `workspaceStore` for 2s after load showed it stable, so the "store settles late" explanation is unconfirmed. A second click always worked.
- ⌘-click deliberately bypasses the probe (you cannot `preventDefault` a new-tab click without breaking it), so on a missing item it still opens a blank editor — the dead end the normal click now avoids.
- A **draft-only** item in prod (never deployed) can't be updated across — `deployItem` reads the deployed version, so it fails with `Flow not found at path …`. The prompt handles it as designed (stays open, error toast), but there is nothing to pull on the compare page either. Pre-existing and shared with the compare page.


Guilhem-lm:
All four addressed in `81413de13a`, each verified against Postgres rather than by reading. Thanks — #1 in particular was right and my earlier reasoning for not fixing it was wrong.

### 1. Final item write is not create-only — fixed

I'd argued this couldn't be closed without reimplementing the per-kind deploys. That was wrong: the frontend owns the `DeployProvider` the shared `deployItem` runs against, so it can refuse exactly the three writes the `alreadyExists` branch reaches for — `updateFlow`, `updateApp`/`updateAppRaw`, and a `createScript` carrying a `parent_hash`, which is what makes an otherwise identical create an update. A refusal sets a `conflict` flag on the result (not matched out of the error text — `deployItem` swallows every throw), and the prompt then opens their version instead of replacing it.

Verified by stubbing the *first* existence probe to report absent and letting the second through — precisely the race:

```
probeCalls: 2
toast:      "u/admin/e2e_conflict is already in dev, opening it"
dev:        "THEIR VERSION"   (1 version row — no version chained on top)
edit-in:    "PROD VERSION"
```

### 2. Copied folder ACLs retain source-local usernames — fixed

`owners` and `extra_perms` keys now go through the same source-username → email → target-username translation as the identity rules. An entry with no account in the target is dropped and reported, which can only make the folder narrower, and `create_folder` force-appends the caller to `owners`, so nobody is locked out of what they just created.

```
edit-in  {"g/all": true, "u/admin": true, "u/e2eonly": true}
dev      {"g/all": true, "u/admin": true}
```
> "f/e2eblock was created without access for u/e2eonly — no such user or group in dev"

### 3. Dropped matching folder rules still deploy under a fallback identity — fixed

Agreed, and the toast wasn't enough. Owners and identity rules now fail differently, because their failure modes differ: dropping an ACL entry only narrows access, but dropping a rule changes *who the item runs as*. An unresolvable identity rule now refuses the whole copy rather than proceeding.

I chose to refuse on *any* unresolvable rule rather than only one whose glob matches this item — replicating the server's glob semantics client-side to decide whether a silent identity substitution is safe is exactly the kind of subtly-wrong that this branch keeps finding. Verified nothing lands:

```
Could not update dev with f/e2eblock/e2e_block_script: f/e2eblock runs items on
behalf of u/e2eonly, which has no account in the target workspace. Bring the
folder across from the compare page first.

dev folders: (none)   dev scripts: (none)
```

### 4. Instance groups pass copied-rule validation — fixed

Confirmed in `groups.rs:135` — `list_group_names` is `SELECT name FROM group_ … UNION SELECT name FROM instance_group`, while folder rule resolution checks `group_` only. Swapped to `listGroups`, which reads the workspace's own rows.

---

On the two "manually exercise" suggestions: concurrent target creation and the unresolved matching folder identity are both covered above. The colliding-username case is the one I could not construct on this instance without inventing two accounts that share a username across workspaces; the translation is keyed on email throughout, so a namesake can never be selected, but I'm flagging that it's reasoned rather than executed.

PR #10355, opened 2026-07-27
feat: allow changing an account email in the superadmin settings

rubenfiszel:
## Summary

Fixes WIN-2255.

Instance settings could rename a user's instance-wide username but not move an account to a different email address, so an address change meant deleting the account and recreating it — losing the username, role, workspace memberships, drafts and tokens.

This adds a superadmin-only `POST /users/change_email/{email}` that moves the account **in place**: the `password` row is kept, so the instance-wide username (along with the role, login type and password hash) follows the new address, and every email-keyed row across the schema is repointed at it.

`audit` is deliberately not rewritten — it records who did what at the time.

## Changes

- **New endpoint** `POST /users/change_email/{email}` (`windmill-api-users`), superadmin-only (`require_super_admin` + `forbid_superadmin_job_token`), transactional. It repoints: `password` (→ `draft` via its `ON UPDATE CASCADE` fkey), `usr`, `workspace.owner`, `workspace_settings.slack_email`, `usage`, `email_to_igroup`, `workspace_invite`, `tutorial_progress`, `token`, `mcp_oauth_refresh_token`, `mcp_oauth_server_code`, `schedule`, `capture_config`, `azure_trigger`, `script`/`flow`.`on_behalf_of_email`, the queued rows of `v2_job`/`job_perms`, and the fork-deployment-request tables. Pending password-reset links (`magic_link`) and the external-JWT cache (`unique_ext_jwt_token`), which still assert the old address, are dropped.
- Rejections: malformed or over-long address, unchanged address, reserved `superadmin_*@windmill.dev` addresses, a destination already used by another account (compared case-insensitively, since login lowercases its input), and — when the new address exceeds the 50 chars that `workspace.owner` / `workspace_settings.slack_email` / `usage.id` hold — a user referenced by one of those columns.
- Tokens are kept valid rather than revoked; a `notify_token_invalidation` event is emitted per affected token prefix so every API server drops its cached authed user instead of serving the previous address.
- `TOKEN_EMAIL_CACHE` (`WM_END_USER_EMAIL`) gets a 60s TTL — nothing can invalidate it by token hash, so it would otherwise serve the previous address until LRU eviction.
- `VALID_EMAIL` in `windmill-common::users`, a strict subset of the `proper_email` CHECK constraint carried by `usr` / `workspace_invite`.
- **Frontend**: an `Email` section at the top of the instance-settings user Edit popup (`ChangeInstanceEmailInner.svelte`), with an alert spelling out what survives the move.

Two judgment calls worth flagging: the `v2_job` / `job_perms` rewrites are scoped to jobs still in `v2_job_queue` (those drive the permissions of a run that has not finished; neither column is indexed, so rewriting a busy user's whole history would hold the transaction's locks for minutes), and completed-job rows are left as history alongside `audit`.

## Screenshots

Edit popup as opened — the address is prefilled and no warning is shown until it is actually edited:

popup-untouched

Once the field differs from the current address, a warning framing this as a last-resort operation appears along with the enabled button:

popup-warning

After confirming — the list refreshes on the new address, username `olduser` preserved:

after-change

## Known limitations

Called out in the in-UI warning rather than enforced, since all three are legitimate for an
instance that has prepared for them:

- **SSO / SCIM instances must be updated at the identity provider first.** SSO login
  (`oauth2_ee.rs`) and SCIM sync (`scim_ee.rs`) both auto-create a `password` row for an
  unknown asserted email, so a login on the previous address after the move recreates it as a
  separate, empty account with a freshly generated username.
- **The account inherits what was already addressed to the destination** — `email_to_igroup`
  and `workspace_invite` rows on the new address are merged in rather than dropped, and an
  instance group can confer `super_admin`/`devops` via `apply_instance_role`.
- **Identity history splits across the two addresses**: audit rows and completed-job
  `permissioned_as_email` keep the previous one. The runs UI attributes by `created_by` (the
  username, which does not change), so visible run history stays coherent; email-keyed queries
  do not.

Moving your *own* account is rejected outright (backend and UI): every API server caches the
caller's identity behind their token until the invalidation event is polled, so the caller would
briefly authenticate as an address that no longer exists.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test users` — new `test_change_user_email` pins username preservation, `usr` membership and `workspace.owner` following the move, the old `password` row being gone, the exact and case-only collision 400s, the malformed-address 400, the unknown-user 404 and the non-superadmin 401
- [x] `cargo test -p windmill-common --lib users::` — `VALID_EMAIL` unit test
- [x] Playwright against the dev instance: moved a user with a workspace membership, an instance-group membership and usage counters through the Edit popup; confirmed in the DB that `password.username`, `usr`, `usage` and `email_to_igroup` all followed, and that the cascaded `draft` row did too
- [x] Confirmed against the running backend that a raw-email `permissioned_as` (schedule) and an app `policy.on_behalf_of{,_email}` both follow the move, and that a self-change is rejected
- [x] Confirmed a schedule owned by the moved user keeps a coherent identity: `schedule.email` follows, while `permissioned_as`/`edited_by` stay on the unchanged username that the tick actually resolves from
- [x] Verified over the API that the moved account logs in with the new address and is rejected on the old one, that its existing token keeps working and reports the new address once the invalidation lands, and that the rejection paths above return the stated status codes

---
## Summary by cubic
Allow superadmins to change a user's account email in place from instance settings or via API, preserving username, role, memberships, drafts, and tokens. Implements WIN-2255 to avoid delete/recreate flows.

- **New Features**
  - Added superadmin-only `POST /users/change_email/{email}` (transactional). Repoints all email-keyed rows, including: `password`/`usr`, tokens, invites, instance groups, `usage`, `workspace.owner`, `workspace_settings.slack_email`, runnables (`schedule`, `capture_config`, triggers), queued jobs, and app policy fields (`policy.on_behalf_of` and `policy.on_behalf_of_email`). Also updates any raw-email `permissioned_as` values in schedules/triggers/apps and folder `default_permissioned_as` rules (order preserved).
  - Validates input: well-formed email, not identical, blocks reserved `superadmin_*@windmill.dev`, case-insensitive collisions, rejects >50-char targets when referenced by short columns (including queued `v2_job.permissioned_as`), and rejects changing your own account.
  - Keeps tokens valid and emits `notify_token_invalidation` per token prefix; clears pending password resets and `unique_ext_jwt_token` cache.
  - Auth: sets `WM_END_USER_EMAIL` token→email cache TTL to 60s to avoid stale mappings.
  - Common: adds `VALID_EMAIL` regex aligned with DB constraints.
  - UI: Adds Email field in the superadmin Edit popup; disables self-change; shows a “last resort” warning (with SSO/SCIM guidance) only after the email is edited; refreshes list after change.
  - OpenAPI: documented the new endpoint.

<sup>Written for commit 6044c50922a1a91e9d663808dec668c5f8b9f99c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 6044c50922a1a91e9d663808dec668c5f8b9f99c

PR #10356, opened 2026-07-27
fix(ai): support OpenAI/Anthropic-compatible gateways in workspace AI settings

rubenfiszel:
## Summary

An AI resource pointing at an OpenAI-compatible or Anthropic-compatible gateway (rather than the vendor's own API) works in an AI agent flow step but is rejected with `Unauthorized` when the same resource is used for the workspace/instance AI settings. The two paths build their requests differently, and only the proxy path sends credentials in a shape gateways reject.

Reproduced against a live gateway with `base_url` `https://<gateway>/proxy/openai` and `https://<gateway>/proxy/anthropic/v1`:

| Provider | Before | After |
|---|---|---|
| OpenAI | `401 Unauthorized - Missing Authorization Header` | `200` |
| Anthropic | `401 Unauthorized - Both Authorization and X-API-Key headers are set, use only one` | `200` |

**OpenAI** — `is_azure` classified *any* base URL other than `https://api.openai.com/v1` as Azure, so a gateway received the Azure `api-key` header instead of `Authorization: Bearer`, plus an `/openai/v1/`-rewritten path (a 404 on its own once auth is fixed). The agent-step path never had this problem because `OpenAIQueryBuilder` always sends bearer auth.

**Anthropic** — the proxy sent `authorization: Bearer <key>` *and* `X-API-Key: <key>`. Gateways reject ambiguous credentials. `get_auth_headers` (the agent-step path) already sends exactly one; the proxy now matches it.

## Changes

- `is_azure` matches the endpoint instead of "base URL differs from the default": Azure-owned host suffixes (including sovereign clouds and API Management) or the `/openai/deployments` path that `openai_azure_base_path` documents, so Azure OpenAI behind a custom domain keeps its current treatment.
- `build_anthropic_proxy_request` sends one credential header: bearer for Vertex, `x-api-key` everywhere else.
- A resource that sets its own `authorization`/`x-api-key`/`api-key` header now suppresses the built-in one, on **both** the proxy and the agent-step path. Outgoing headers are appended, not replaced, so a user-supplied auth header previously produced two credentials on the wire; this is the escape hatch for an Anthropic-compatible endpoint that wants bearer auth. Non-credential headers such as `anthropic-version` are unaffected.

### Known limitations

Both are the deliberate cost of having no rule that separates an Azure endpoint from a gateway by URL alone. A resource in either situation should use the `customai` / `azure_openai` provider as appropriate.

- An OpenAI-compatible endpoint hosted *on* an Azure-owned domain (e.g. a self-hosted server behind API Management) is still read as Azure. Unchanged from before.
- **Behavior change on upgrade:** an `openai_azure_base_path` (or OpenAI-resource `base_url`) pointing at Azure OpenAI through a customer-owned domain *without* the `/openai/deployments` path — e.g. `https://ai.corp.example/openai/v1` — now gets bearer auth and the plain path instead of `api-key` + Azure URL building. The repro gateway's URL ends in `/openai`, so no path rule distinguishes the two.

## Test plan

- [x] `cargo test -p windmill-ai` (99 pass), `cargo check`, `cargo fmt --check` on the touched crates
- [x] End-to-end against a live gateway through `POST /api/w/<ws>/ai/proxy/...`: both providers 401 before the change and 200 after, with fresh (non-cached) completions
- [x] Workspace settings → Windmill AI → "Test key" for both providers in the browser: `POST /ai/proxy/responses`, `POST /ai/proxy/v1/messages` and `GET /ai/proxy/models` all 200
- [ ] Regression check by a reviewer with Azure access, since `is_azure` also feeds the agent-step chat-completions builder (`other.rs`) — exercise each in **both** workspace settings and an AI agent step:
  - [ ] Azure OpenAI, bare-host base and `…/openai/deployments/<id>` base: still resolves to the `/openai/v1` (or deployment) URL with the `api-key` header
  - [ ] Azure AI Foundry serving Claude: `POST /ai/proxy/v1/messages` still completes now that only `x-api-key` is sent (bearer is no longer sent alongside it)

### DIFF
--- a/backend/windmill-ai/src/ai_providers.rs
+++ b/backend/windmill-ai/src/ai_providers.rs
@@ -30,6 +30,30 @@ pub const OPENAI_BASE_URL: &str = "https://api.openai.com/v1";
 pub const DEEPSEEK_BASE_URL: &str = "https://api.deepseek.com/v1";
 pub const GOOGLE_AI_BASE_URL: &str = "https://generativelanguage.googleapis.com/v1beta";
 
+/// Hosts that serve the OpenAI API with Azure conventions: Azure OpenAI
+/// (`*.openai.azure.com`), AI Foundry (`*.services.ai.azure.com`,
+/// `*.cognitiveservices.azure.com`), their sovereign-cloud counterparts, and API
+/// Management fronting either.
+const AZURE_HOST_SUFFIXES: [&str; 4] = [".azure.com", ".azure.us", ".azure.cn", ".azure-api.net"];
+
+fn is_azure_host(base_url: &str) -> bool {
+    let authority = base_url
+        .split_once("://")
+        .map_or(base_url, |(_, rest)| rest)
+        .split(['/', '?', '#'])
+        .next()
+        .unwrap_or_default();
+    let host_port = authority.rsplit_once('@').map_or(authority, |(_, h)| h);
+    let host = host_port
+        .rsplit_once(':')
+        .map_or(host_port, |(host, _)| host)
+        .to_ascii_lowercase();
+
+    AZURE_HOST_SUFFIXES
+        .iter()
+        .any(|suffix| host.ends_with(suffix))
+}
+
 /// Empty string signals BedrockClient::from_env() to use the region from AWS environment/config
 /// (e.g., AWS_REGION or AWS_DEFAULT_REGION env vars, or ~/.aws/config)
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ 6812eb47c47070d9af7793c47b5189d0cbf73a6a

PR #10359, opened 2026-07-27
fix(ai): make the proxy and the AI agent step read a resource the same way

rubenfiszel:
## Summary

Follow-up to #10356. That fix landed in the AI **proxy** (workspace/instance AI settings) only, because the proxy and the **AI agent step** derive the endpoint and the credential header from the same resource independently. Nothing forced them to agree, so a resource could authenticate in one and 401 in the other — which is why the bug went unnoticed until it was reported.

A parity test now pins them together across the provider/platform matrix. Every divergence below was found by that test failing, not by inspection.

## Changes

**Parity test** (`providers/mod.rs`) — asserts both paths derive the same endpoint and the same credential header from identical `ProviderCredentials`, and apply the same rule for *dropping* the built-in credential. Verified as a real guard, not decoration: it fails on #10356's bug (`left: [("x-api-key", …)]` vs `right: [("authorization", …), ("x-api-key", …)]`) and on each fix below when reverted.

**Anthropic base URLs were read differently.** The proxy trimmed and re-appended `/v1`; the agent step appended `/messages` to the stored value. So `base_url = https://gw/anthropic` worked in workspace settings and 404'd in an agent step. `build_anthropic_api_url` accepts both forms on both paths. The proxy URL also no longer depends on the client-supplied `X-Anthropic-SDK` header — the same resource resolved differently per caller — so that header is removed from both ends.

**A trailing slash doubled it.** `base_url = https://gw/v1/` produced `https://gw/v1//chat/completions` in an agent step while the proxy trimmed it.

**An OpenAI resource pointed at Azure.** The proxy applied Azure's URL layout and `api-key` header; the agent step sent bearer auth and the plain path, because `OpenAIQueryBuilder` ignored `is_azure`. Now aligned with `OtherQueryBuilder`.

**An empty credential where the proxy sends none.** With no api key the agent step sent `Authorization: Bearer ` while the proxy omitted the header entirely. `retain_effective_credentials` gives both the same rule, so a keyless endpoint (self-hosted vLLM/ollama with auth off), a resource-supplied credential header, or mTLS all keep working.

**OAuth resources in agent steps.** A worker has no client credentials exchange, so the request went out with an empty credential and came back 401. It now fails with the reason — unless the resource carries the credential header *its provider actually reads*, in which case it runs. A credential-shaped header the provider ignores (an `x-api-key` used for routing by an OpenAI-compatible gateway) does not count.

## Behavior changes to note

- An Anthropic-compatible endpoint serving `/messages` at its configured root (rather than `/v1/messages`) becomes unreachable. No such endpoint is known; api.anthropic.com and the gateways in front of it all serve `/v1`.
- An `openai` resource pointed at a bare `*.azure-api.net` APIM that fronts a bearer-auth backend now gets Azure treatment in agent steps too. That rule was already decided for the proxy in #10356; such a resource should use `customai`.

## Test plan

- [x] `cargo test -p windmill-ai` (104 pass), `cargo check`, `cargo fmt --check` on touched files
- [x] `npx vitest run lib.anthropicRouting.test.ts` (8 pass) for the removed header
- [x] Each new guard verified by reverting its fix and watching the specific case fail
- [ ] Manual: an Anthropic resource whose `base_url` omits `/v1`, exercised in **both** workspace settings and an AI agent step — both should reach `/v1/messages`
- [ ] Manual: an `openai` resource on an Azure deployment URL (or `openai_azure_base_path`) in an agent step — should authenticate with `api-key` instead of 401ing on a bearer
- [ ] Manual: a keyless `customai` resource against a local no-auth OpenAI-compatible server in an agent step — must still run

`frontend/` changes are header-protocol only (no visible UI), so no screenshots. `npm run check:fast` reports 3 errors, all pre-existing on main in files this PR does not touch (stale generated client + a missing dep).

### DIFF
--- a/backend/windmill-ai/src/ai_providers.rs
+++ b/backend/windmill-ai/src/ai_providers.rs
@@ -224,6 +224,22 @@ impl AIProvider {
         }
     }
 
+    /// Build an Anthropic Messages API URL. The API lives under `<root>/v1/`, but a
+    /// resource may store the base URL with or without that `/v1` and a caller may
+    /// pass the path with or without it (the Anthropic SDK sends `v1/messages`,
+    /// `get_endpoint` asks for `messages`). Both forms resolve to the same endpoint
+    /// so a resource reads the same way from the proxy and from an agent step.
+    ///
+    /// This assumes the API is served under `/v1`, as api.anthropic.com and the
+    /// gateways in front of it are.
+    pub fn build_anthropic_api_url(base_url: &str, path: &str) -> String {
+        let base_url = base_url.trim_end_matches('/');
+        let root = base_url.strip_suffix("/v1").unwrap_or(base_url);
+        let path = path.trim_start_matches('/');
+        let path = path.strip_prefix("v1/").unwrap_or(path);
+        format!("{}/v1/{}", root.trim_end_matches('/'), path)
+    }
+
     /// Whether the URL is a bare scheme+host with no path component, e.g.
     /// `https://x.services.ai.azure.com` (vs `https://x.openai.azure.com/openai/deployments/y`).
     fn is_bare_host(url: &str) -> bool {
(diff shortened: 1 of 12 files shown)


rubenfiszel:
✅ Review round clean @ 1996afc584d094dcdb0b39c94320f7deaca732d6

PR #10361, opened 2026-07-27
feat: list draft-only runnables on the homepage again

rubenfiszel:
## Summary

Before #10297 the homepage called the per-kind endpoints with `include_draft_only=true`, so a script/flow/app you had saved as a draft but never deployed appeared in the list. The merged `GET /runnables/list` that replaced them selects `FROM script/flow/app` only, so those items silently disappeared from the homepage — list, tree, search and (since #10351) the per-owner counts alike. They remained reachable only through the Drafts drawer.

This brings them back as ordinary rows rather than as an appended block: a draft-only branch joins the UNION per kind, so drafts sort, search, page and count with everything else, at any position in the stream (the per-kind endpoints could only staple them onto page 0 with every filter disabled).

**A draft is listed under the path it says it will deploy to**, not the `u/<you>/draft_<uuid>` it is parked at: that typed path is what the row already displayed, and now also what search matches, what the name order sorts on, what the owner filter selects, and what the tree and the counts group it under. `path` itself stays the stored key, because that is what the editor link resolves.

## Changes

**Backend**
- `runnables.rs`: `draft_branch_sql(kind)` projects the caller's drafts at paths with no deployed row into the same column set as the deployed branches (`DISTINCT ON (path)` collapsing an owned draft and a legacy NULL-email one, as the per-kind endpoints do). Gated behind a new `include_draft_only` query param, off by default so picker callers stay deployed-only, and ignored for operators, in the archived view, and under a label filter (a draft carries no labels of its own).
- The search predicate and the owner (`path_start`) filter resolve through the typed path for draft branches; `sort_name` prefers it too.
- `/runnables/counts` grows the same flag and an `add_draft_counts` pass. It is deliberately not restricted to the readable owners the deployed passes walk: a draft is the caller's own and `/list` reads it off the non-RLS `draft` table, so scoping it to folder grants would hide a count for a row that still lists. Deployed and draft rows cannot overlap, so the two counts add.
- Library scripts and pipeline members follow the same rules as deployed rows, read off the `auto_kind` the editor stamped into the draft. It is not re-derived from the content, so a pipeline member errs toward listing as its own row rather than being folded away and hidden.

**Migration** — one index, `draft (workspace_id, typ, email)`. Every draft query is "this caller's drafts of this kind", and without those keys the existing `draft_user_listing_idx` hands back the caller's drafts of every kind for each per-kind branch to throw away.

**Frontend**
- `ItemsList.svelte` passes `includeDraftOnly: true` on the four list calls and the counts call, and the client-side fuzzy filter indexes a draft by its typed path.
- `treeViewUtils.ts` gains `treePath()` so the tree groups a draft where the counts count it.

## What was measured

`draft` is written on every editor autosave, so the index was sized against measurements rather than guessed. On a seeded workspace with 60k drafts, 8.5k of them one user's across all three kinds (3 runs each):

| Variant | Execution time |
|---|---|
| `(workspace_id, typ, email, path text_pattern_ops)` | 8.7 / 9.7 / 8.9 ms |
| `(workspace_id, typ, email)` | **8.4 / 8.5 / 8.7 ms** |
| no new index | 9.0 / 10.6 / 9.8 ms |

The `path` tail earns nothing: a draft is filtered by its typed path, which lives in the JSON and is only resolved after the `DISTINCT ON` dedup, so no index prunes on it. Time/name sort indexes were tried and dropped for the same reason — the dedup has to sort by path before the sort key applies, so an ordered index is never used for the order. "Draft-only" itself cannot be pushed into the index either: it means "no deployed row at this path", and Postgres rejects a subquery in an index predicate (`cannot use subquery in index predicate`); the anti-join prunes it at query time off the deployed tables' own `(workspace_id, path)` indexes.

## Test plan

- [x] `data-pipelines` (8 deployed, 10 draft-only): `/list` returns 18 with `include_draft_only=true` and 8 without; counts go from `{u/admin: 1}` to `{u/admin: 10, u/me: 1}` — `u/me` holds nothing but a draft and now gets a node and a chip.
- [x] Flat list, tree view and search each show the draft rows, with the existing "Draft only" badge and `*` marker the row components already had.
- [x] A draft stored at `u/admin/draft_typed_elsewhere` but typed as `f/orders/typed_into_folder`: counted under `f/orders`, returned by `path_start=f/orders/`, found by searching `typed_into`, and grouped under `f/orders (2 items)` in the tree.
- [x] Searching `wholesome` (a name that exists only in the typed path) finds the draft; before, both the server predicate and the client fuzzy filter matched only the `draft_<uuid>` stored path and returned nothing.
- [x] Drafts by other people on a deployed row still show: a coworker draft on `u/admin/auto_resolve_demo` renders the "CO Draft" badge, `draft_users` untouched.
- [x] Operators, the archived view and a label filter get no draft rows; `include_draft_only` unset returns exactly the previous result.
- [x] `cargo build`, `npm run check:fast`, `cargo fmt`, prettier clean.

## Screenshots

Flat list — draft-only rows among the deployed ones:

drafts-flat

Tree — `f/orders (2 items)` holds the pipeline plus a draft that is only *typed* into that folder:

drafts-tree-typed

Search by the typed name:

drafts-search

Someone else's draft on a deployed row, unchanged:

drafts-others

---
## Summary by cubic
Restores draft-only runnables to the homepage list, tree, search, and owner chips. They sort and paginate with deployed items using their typed path. Adds `include_draft_only` to `GET /runnables/list` and `GET /runnables/counts`, and updates the UI to use it.

- **New Features**
  - Backend: Adds a per-kind draft-only branch that joins the unified listing when `include_draft_only=true`. Search, owner (`path_start`), and name sort resolve via the typed path. Ignored for operators, in the archived view, and under a label filter. Drafts are excluded from the starred-first page (they can’t be starred). `/runnables/counts` supports the same flag and adds draft counts; not restricted to readable owners so badges match the list. Library scripts and pipeline members follow `auto_kind`.
  - Frontend: All list and counts calls pass `includeDraftOnly: true`. Tree, filters, and owner grouping use the effective typed path via `effectivePath()`. Fuzzy search indexes and displays the typed path. Owner chips use server counts, hide empty owners, prioritize your space, rank by count, and cap visible chips at 20 with a “…” toggle.
  - OpenAPI: Documents `include_draft_only` on both endpoints.

- **Migration**
  - Adds index `draft_kind_user_listing_idx` on `draft (workspace_id, typ, email)` (created concurrently). No breaking changes.

<sup>Written for commit 6ee9634b07cfab7f38c2b97502aa06d96cc722ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/migrations/20260727151319_draft_only_listing_indexes.down.sql
+++ b/backend/migrations/20260727151319_draft_only_listing_indexes.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX IF EXISTS draft_kind_user_listing_idx;
--- a/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
+++ b/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
@@ -1,0 +1,19 @@
+-- Index backing draft-only rows in the unified homepage listing
+-- (/runnables/list) and its per-owner counts (/runnables/counts).
+--
+-- Every draft query there is "this caller's drafts of this kind", so `typ` and
+-- `email` are key columns rather than filters: without them the existing
+-- draft_user_listing_idx hands back the caller's drafts of every kind and the
+-- branch throws most away (measured on a 5k-draft user: 1.4ms vs 4.1ms per
+-- kind). `path` sorts with text_pattern_ops so an owner-filtered listing
+-- (`path LIKE 'f/foo/%'`) can seek instead of scanning the whole slice — the
+-- default opclass sorts by the database collation and cannot serve a byte
+-- prefix.
+--
+-- Deliberately the only index added: `draft` is written on every editor
+-- autosave, and time/name sort indexes measured as pure overhead — the
+-- listing's DISTINCT ON (path) dedup has to sort by path before the sort key
+-- is applied, so an ordered index is never used for the order.
+-- Created CONCURRENTLY via the OVERRIDDEN_MIGRATIONS rewrite in windmill-api/src/db.rs.
+CREATE INDEX IF NOT EXISTS draft_kind_user_listing_idx
+    ON draft (workspace_id, typ, email, path text_pattern_ops);
--- a/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
+++ b/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
@@ -3,17 +3,16 @@
 --
 -- Every draft query there is "this caller's drafts of this kind", so `typ` and
 -- `email` are key columns rather than filters: without them the existing
--- draft_user_listing_idx hands back the caller's drafts of every kind and the
--- branch throws most away (measured on a 5k-draft user: 1.4ms vs 4.1ms per
--- kind). `path` sorts with text_pattern_ops so an owner-filtered listing
--- (`path LIKE 'f/foo/%'`) can seek instead of scanning the whole slice — the
--- default opclass sorts by the database collation and cannot serve a byte
--- prefix.
+-- draft_user_listing_idx hands back the caller's drafts of every kind and each
+-- per-kind branch throws most away.
 --
--- Deliberately the only index added: `draft` is written on every editor
--- autosave, and time/name sort indexes measured as pure overhead — the
--- listing's DISTINCT ON (path) dedup has to sort by path before the sort key
--- is applied, so an ordered index is never used for the order.
+-- Deliberately just these three columns. `path` would not help: a draft is
+-- listed, grouped and filtered under the path it says it will deploy to, which
+-- lives in the draft JSON and is only resolved after the DISTINCT ON dedup, so
+-- no index can prune on it. Nor can "draft-only" itself be pushed into the
+-- index — it means "no deployed row at this path", and Postgres rejects a
+-- subquery in an index predicate; the anti-join prunes it at query time off the
+-- deployed tables' own (workspace_id, path) indexes.
 -- Created CONCURRENTLY via the OVERRIDDEN_MIGRATIONS rewrite in windmill-api/src/db.rs.
 CREATE INDEX IF NOT EXISTS draft_kind_user_listing_idx
-    ON draft (workspace_id, typ, email, path text_pattern_ops);
+    ON draft (workspace_id, typ, email);
--- a/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
+++ b/backend/migrations/20260727151319_draft_only_listing_indexes.up.sql
@@ -6,13 +6,17 @@
 -- draft_user_listing_idx hands back the caller's drafts of every kind and each
 -- per-kind branch throws most away.
 --
--- Deliberately just these three columns. `path` would not help: a draft is
--- listed, grouped and filtered under the path it says it will deploy to, which
--- lives in the draft JSON and is only resolved after the DISTINCT ON dedup, so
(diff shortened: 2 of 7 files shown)


rubenfiszel:
Addressed both P1s, plus a perf fix and a migration-runner landmine found on the way.

**P1 — frontend categorization still used the parking path.** `treePath` covered only tree grouping and one reload filter. Replaced by `effectivePath()` in `treeViewUtils.ts`, now used at every site that *categorizes*: pipeline-member folder detection (`ItemsList:290`), lazy-owner replacement (`:445`), the open-owner reload filter (`:492`), the user-folder filter (`:516`), and the owner-chip tally (`:715`). `path` stays the storage identity for the editor link and the row key. Verified: with `Only u/admin and f/*` active, a draft parked at `u/admin/draft_<uuid>` but typed as `f/orders/typed_into_folder` now survives the filter and lands under `f/orders` in the chips and the tree, where before its storage path failed `startsWith('f/')`.

**P1 — app/raw_app pick was nondeterministic.** `DISTINCT ON (d.path)` ordered only by path and ownership, so a path holding both an `app` and a `raw_app` draft resolved arbitrarily — deciding `raw_app`, summary and draft_path, i.e. which editor the link opens. Both the list branch and the counts pass now order `d.path, (d.email IS NULL), d.created_at DESC`, matching apps.rs. Verified with both kinds at one path: the newest (`raw_app`) wins on every run and the row links `/apps_raw/edit/u/admin/draft_dual`.

**Perf.** Draft branches also ran in the starred pass, where they can never match — a draft-only path has no favorite row. Dropped them there and removed the favorite join from the draft branch entirely. That cut the added cost by two thirds.

Re-benchmarked on an extreme workspace (50k scripts, 5k flows, 200 members, 400 folders, 60k drafts across users), median of 7, flag off → on:

| | admin, 2000 *own* drafts | non-admin (RLS path), 300 own drafts |
|---|---|---|
| list, updated order | 15 → 23 ms | 18 → 19 ms |
| list, name order | 12 → 20 ms | 18 → 20 ms |
| search | 39 → 55 ms | 22 → 25 ms |
| owner filter | 17 → 23 ms | 22 → 24 ms |
| counts | 12 → 24 ms | 4 → 6 ms |

The cost tracks the caller's **own** draft count, not workspace size: 60k drafts belonging to 200 other users cost the non-admin nothing, and the RLS path is not disproportionately hit. Its counts stay cheaper than the admin sweep (4 ms vs 12 ms) because it walks owners with index-only prefix counts instead of scanning 55k rows. The 2000-own-drafts column is pathological — drafts are deleted on deploy.

**Migration-runner landmine.** The CONCURRENTLY rewrite in `db.rs` splits the migration file on `;` and treats any resulting chunk whose first line isn't `--` as a statement, so a semicolon *inside a comment* gets the trailing comment text executed as SQL (`syntax error at or near "the"`, then `unterminated quoted string` when I quoted the character while documenting it). The migration now has exactly one semicolon, with a comment saying why. Worth hardening the splitter separately.

PR #10362, opened 2026-07-27
fix(ai): resolve deployment-pinned Azure base URLs to the v1 surface

rubenfiszel:
## Summary

An Azure OpenAI base URL that names a deployment — the `https://<res>.openai.azure.com/openai/deployments/<id>` format the `openai_azure_base_path` instance setting documents — was appended to as-is. That targets Azure's legacy surface, which serves **only** with an `api-version` query string. Windmill never appends one, and the proxy path arrives through an axum path extractor that carries no query string, so a client SDK cannot supply it either.

The result: that base URL 404s on **both** paths, and always has.

Probed against a live Azure OpenAI resource with an empty body, so an existing route answers 400 and a missing one answers 404:

| URL | Result | |
|---|---|---|
| `…/openai/deployments/<id>/chat/completions` | **404 Resource not found** | what the proxy built |
| `…/openai/deployments/<id>/responses` | **404 Resource not found** | what the agent step built |
| `…/openai/deployments/<id>/chat/completions?api-version=…` | 400 | route exists — the query is mandatory |
| `…/openai/v1/chat/completions` | 400 → **200** with a real completion | what the proxy builds now |
| `…/openai/v1/responses` | 400 → **200** | what the agent step builds now |

## Changes

- A base containing `/openai/deployments`, pinned to a deployment or not, resolves to the resource root plus `/openai/v1/<path>`, like every other Azure shape. This replaces the `ends_with("/deployments")` branch, which handled only the unpinned form.
- `azure_foundry_root` recovers the root through the same helper, so a Foundry resource on such a base builds its Claude URL from the root too — otherwise this fix would have been one-sided, correcting GPT models while leaving Claude on the broken shape.
- The instance-settings help text said the URL makes every workspace "run on the specified deployed model". That surface never delivered it, and on the v1 surface the deployment comes from the request body, so the text now says where the model actually comes from.

## Notes for reviewers

- **The deployment id in the URL is now ignored**, and the model comes from each workspace's model list, whose entries must be Azure *deployment* names. Nothing regresses — the pinned URL 404'd unconditionally — but the failure shape moves from "404 on every call" to Azure's `DeploymentNotFound` if the model list holds Azure model names instead. That is strictly more actionable. Rewriting the body's `model` to match the URL would silently override an explicit user choice, which is worse.
- **Don't "clean up" a stored base URL to `…/openai/v1`.** Azure detection for a custom domain keys on the `/openai/deployments` marker, so shortening the URL loses `api-key` auth and 401s. Keep it as stored.
- The rewrite fires on any host containing `/openai/deployments`, not only Azure-owned ones, because custom-domain Azure resources are a supported case. A non-Azure facade serving that layout *without* requiring `api-version` would change behavior; no such endpoint is known, and gating on Azure host suffixes would drop the custom-domain Azure case this is meant to fix.

## Screenshots

Instance settings → Private Hub. The field is `hiddenIfEmpty`, so it renders only once set; shown here with an example value. The apostrophe, em dash and `Learn more` link all render correctly.

azure-base-path-help

## Test plan

- [x] `cargo test -p windmill-ai` (104 pass), `cargo check`, `cargo fmt --check`
- [x] Live probe against an Azure OpenAI resource: the old URLs 404, the new ones return 200 with generated content
- [x] `npm run check:fast` — 3 errors, all pre-existing on main in files this PR does not touch
- [x] Help text verified in the browser (screenshot above)
- [ ] Reviewer with Azure access: confirm an `azure_openai` resource stored as a bare host or `/openai` base is unchanged (those branches are untouched), and that a Foundry Claude deployment on a deployment-pinned base now resolves

---
## Summary by cubic
Routes Azure OpenAI deployment-pinned base URLs to the v1 surface so requests stop 404ing and succeed. Also fixes Foundry root detection and updates the instance setting help text.

- **Bug Fixes**
  - Any base containing `/openai/deployments` now resolves to `<root>/openai/v1/<path>`; the deployment id in the URL is ignored. The model comes from each workspace’s model list (use your Azure deployment names).
  - `azure_foundry_root` now strips the deployments path, so Foundry Claude URLs build from the resource root.

- **Migration**
  - No action needed; keep stored URLs as-is. Do not shorten them to `/openai/v1` or custom-domain Azure auth may break.
  - If model lists use Azure model names instead of deployment names, you’ll now see `DeploymentNotFound` instead of 404. Update entries to deployment names.

<sup>Written for commit 459684a453b5ed2b9f17788ac940254d33f8497f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-ai/src/ai_providers.rs
+++ b/backend/windmill-ai/src/ai_providers.rs
@@ -48,6 +48,13 @@ const AZURE_HOST_SUFFIXES: &[&str] = &[
 /// is the format `openai_azure_base_path` documents.
 const AZURE_DEPLOYMENTS_PATH: &str = "/openai/deployments";
 
+/// Where the deployment path starts in `base_url`, i.e. the end of the resource root.
+/// ASCII lowercasing preserves byte offsets, so the index is a char boundary in
+/// `base_url` and slicing on it cannot panic; `to_lowercase` would not be.
+fn azure_deployments_index(base_url: &str) -> Option<usize> {
+    base_url.to_ascii_lowercase().find(AZURE_DEPLOYMENTS_PATH)
+}
+
 /// Whether an OpenAI-API base URL is served by Azure, which authenticates with the
 /// `api-key` header and lays its routes out under `/openai/...`.
 ///
@@ -71,9 +78,7 @@ fn is_azure_endpoint(base_url: &str) -> bool {
     AZURE_HOST_SUFFIXES
         .iter()
         .any(|suffix| host.ends_with(suffix))
-        || base_url
-            .to_ascii_lowercase()
-            .contains(AZURE_DEPLOYMENTS_PATH)
+        || azure_deployments_index(base_url).is_some()
 }
 
 /// Empty string signals BedrockClient::from_env() to use the region from AWS environment/config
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 459684a453b5ed2b9f17788ac940254d33f8497f

PR #10364, opened 2026-07-27
fix: raw apps with no stylesheet were permanently un-deployable

Guilhem-lm:
## Summary

A raw app saved with **no `.css` bundle row at all** could never be deployed to another
workspace again. Deploy re-fetches both halves of the bundle from the source workspace, and
a missing stylesheet returned **404**, which was treated as fatal. Nothing self-healed, and
workspace cloning copies `app_bundles` row by row — so one hand-made demo app became **16
broken versions across 4 app paths and 10 workspaces** on the shared dev database.

The write side already treats `css` as optional; only the read side disagreed. Absent now
means empty.

**Why it surfaced only now:** the producer defect shipped with `ui code builder v0`
(`3c68fef2ab`, 2025-04-20) and sat harmless for 15 months — a style-less app renders fine in
its own workspace, and the only signal was a console 404 that read as dev noise. It became
*fatal* when cross-workspace deploy arrived (`4342c18541`, #8756, 2026-04-08), and became
*visible* when forks and AI sessions started moving these apps between workspaces: every
affected version dates from 2026-06-25 → 07-08.

Notably, both sibling write paths had independently grown the same defensive default —
the chat bridge (#9349, 2026-05-27) and the CLI (#9629, 2026-06-17). Two authors patched
their own call site against a bundle they didn't trust; neither traced it to the shared
source, and the one path that never got a default kept writing broken rows.

## Changes

- **Producer** — `RawAppEditor.svelte`: the `getBundle` reply now defaults `css` to `''`.
  `js` is deliberately *not* defaulted: the backend accepts any present `js` field
  regardless of length, so an empty one would publish a blank app over a working one. A
  missing/empty `js` now rejects with a clear error, matching `rawAppBundlerBridge`.
- **Backend** — `get_raw_app_data`: when no row exists and the request is for `css`, serve
  200 with an empty body instead of 404. This repairs every already-broken app for all
  clients with no data migration and no package release. A missing `.js` still 404s.
  Also silences a real side effect: the raw-app wrapper hardcodes a `<link rel="stylesheet">`,
  so until now every style-less app logged a 404 on *every page load*.
- **Reader** — `cli/windmill-utils-internal/src/deploy.ts`: new `getRawAppBundlePart` helper
  tolerating **only** a `.css` 404. A blanket `catch` would swallow auth/network failures and
  silently deploy an app with its styles stripped.
- **Tests** — a backend integration regression test and a CLI unit test for the helper. Both
  were mutation-tested: reverting the fix, or widening the tolerance to a blanket `catch`,
  fails them.

## Notes for reviewers

- **No screenshots**: the frontend change is inside a `postMessage` handler and has no visible
  UI effect. It was still verified in a real browser — see the A/B below.
- `windmill-utils-internal` is consumed by the frontend as a **published package**, so the
  deploy.ts fix is inert for the frontend until the package is bumped and republished. The
  backend fix is what covers the frontend today; layer 3 matters for older backends.
- Commit `2682414496` on `glm/nit-edit-button` patches this frontend-side only
  (`getRawAppDataAllowingMissingCss`). It becomes redundant once this lands and should be
  deleted when that branch merges.
- Pre-existing, not addressed here: `updateApp` in `RawAppEditorHeader.svelte` has no
  try/catch around `getBundle()`, so its failures fall through to the global
  unhandled-rejection toast rather than a scoped one.

## Test plan

- [x] Backend regression test passes; fails with the exact assertion when the fallback is removed
- [x] CLI unit tests pass (5/5); 3 fail against a blanket-`catch` mutant
- [x] Broken app on a patched backend: `.css` → 200 `text/css` 0 bytes (was 404)
- [x] Missing `.js` still 404s — tolerance does not leak
- [x] App with real CSS (1662 bytes) round-trips byte-identical
- [x] Cross-workspace deploy of a css-less app: aborts unpatched → **HTTP 201** patched
- [x] Controlled browser A/B: with the fix the editor sends `app|js|css` and a 0-byte css row
      appears; with it reverted the request carries only `app|js` and no row is written
- [x] `cargo check -p windmill-api`, `npm run check` (0 errors), `tsc --noEmit` all clean

### DIFF
--- a/backend/tests/raw_app_missing_css.rs
+++ b/backend/tests/raw_app_missing_css.rs
@@ -1,0 +1,172 @@
+//! Regression test for raw apps whose bundle has no stylesheet.
+//!
+//! `create_app_raw` stores the `css` blob only when the multipart field is
+//! present, so an app built from sources with no styles has no `css` row at
+//! all. The read side used to 404 for that, and cross-workspace deploy
+//! re-fetches both bundle parts and treats the 404 as fatal — which made such
+//! an app permanently un-deployable, with nothing to self-heal it.
+//!
+//! This test pins down:
+//!   - a bundle stored without a `css` part serves an empty stylesheet, not 404
+//!     (the fix; pre-fix this 404'd and broke every deploy of the app),
+//!   - a bundle that does have styles still serves them byte-for-byte (the
+//!     tolerance must not start blanking real stylesheets), and
+//!   - a missing `js` blob still 404s — the tolerance must not leak to the
+//!     half of the bundle whose absence really is broken.
+
+use reqwest::multipart;
+use sqlx::{Pool, Postgres};
+use windmill_test_utils::*;
+
+fn client() -> reqwest::Client {
+    reqwest::Client::new()
+}
+
+fn authed(builder: reqwest::RequestBuilder, token: &str) -> reqwest::RequestBuilder {
+    builder.header("Authorization", format!("Bearer {}", token))
+}
+
+fn app_part(path: &str) -> anyhow::Result<multipart::Part> {
+    let app = serde_json::json!({
+        "path": path,
+        "summary": "",
+        "value": {},
+        "policy": { "execution_mode": "publisher" },
+        "raw_app": true,
+    });
+    Ok(multipart::Part::text(app.to_string()).mime_str("application/json")?)
+}
+
+#[sqlx::test(fixtures("base"))]
+async fn test_raw_app_without_stylesheet_stays_readable(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+
+    let server = ApiServer::start(db.clone()).await?;
+    let port = server.addr.port();
+    let base = format!("http://localhost:{port}/api/w/test-workspace");
+
+    let js_body = "console.log('app')";
+
+    // 1. Create a raw app with NO css part at all — exactly what a client that
+    //    omits the field produces, and the state the 16 broken versions are in.
+    let form = multipart::Form::new()
+        .part("app", app_part("u/test-user/nocss")?)
+        .part("js", multipart::Part::text(js_body));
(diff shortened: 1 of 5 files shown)


Guilhem-lm:
Nits addressed in 38a72a3f07 — dropped the historical/ephemeral narration ("used to", "pre-fix", "the 16 broken versions") from the backend regression test's comments. Swept the rest of the diff for the same pattern; nothing else matched.

**Dismissing the frontend-screenshot nit**, with reasoning:

The only `frontend/` change is inside the `getBundle` `postMessage` handler. On the normal path it alters the *request payload* (a `css` part is now sent), not anything rendered — a screenshot of the editor before and after would be pixel-identical and would misrepresent the evidence as visual.

It was verified in a real browser, just not photographically. The controlled A/B was: same app, deployed twice.

| | multipart parts sent | `app_bundles` rows written |
|---|---|---|
| fix applied | `app`, `js`, `css` | `css` (0 bytes) + `js` |
| fix reverted | `app`, `js` | `js` only |

The one genuinely new *visible* state is the "Raw app bundler returned an empty JavaScript bundle." toast from the `js` guard. I tried to capture it and could not reach it through the UI: by the time the deploy drawer is filled in and submitted, a build has always settled, so the builder returns a real bundle and the resolve path wins. The guard is reachable when no build has succeeded yet — first build fails, user hits Deploy — which the happy-path drawer flow can't reproduce. Rather than stage a fake toast, I'm noting the limit: that branch is covered by reasoning and by the backend's own refusal to accept a bundle it would publish blank, not by a screenshot.

`cli/package-lock.json` (the `acorn` `"peer": true` churn flagged by the local Codex pass) is untracked by this PR — it predates the branch and is deliberately not in any of these commits.

Guilhem-lm:
✅ Review round clean @ 6ca55aecacba50864cb655cf2edfaf53eb5dca73

PR #10366, opened 2026-07-27
fix(wac): report a task failure the child round's body catches

rubenfiszel:
## Summary

A WAC v2 task failure was silently **not** reported when the workflow body wrapped the task call in a broad `try/catch`.

A `@workflow` dispatches a task by re-running the *entire* workflow body in a child job with `_executing_key` set, so only the target step runs its inner function. In that child round the raw exception thrown by the task body propagated into the workflow body's own `try/catch`. The body caught it, the child job **completed successfully** returning whatever the catch branch produced, and the parent recorded that as a normal successful step result. In the parent, `await failing_task()` then resolved with the child's value instead of raising, and the user's error handler never ran.

```ts
const boom = task(async function boom() { throw new TypeError("nope"); });
export default workflow(async function main() {
  try { await boom(); } catch (e) { return { caught: "yes" }; }
  return { caught: null };
});
```

Before: parent result `{"caught": null}`, child job **success** with result `{"caught":"yes"}` — the failure vanished.
After: child job **fails** with `TypeError: nope`, the parent records the `__wmill_error` marker, and the replayed parent's catch runs with a `TaskError`.

This is the structural twin of #10348, which stopped a broad catch from swallowing a `step()` **suspend**. Same reasoning: in the child round the executing step's exception *is* the round's result, not something the body may handle. Only the child round is affected — a parent round that legitimately catches a `TaskError` and retries is untouched, because the parking path is reachable only when `_executing_key` matches the step.

## Changes

- **`typescript-client/client.ts`** — `WorkflowCtx` parks the executing step's failure in `_pendingStepFailure` and re-throws it at the next SDK call (`_rethrowSwallowedSuspend` → `_rethrowSwallowed`, which now covers both parked kinds). `task()`'s `_execute_directly` branch routes a thrown task body through `_raiseStepFailure`, passing `StepSuspend` through untouched. New `_takePendingStepFailure()` hands the parked failure to the runner.
- **`backend/windmill-worker/src/bun_executor.rs`** — the generated WAC wrapper calls `ctx._takePendingStepFailure?.()` on both the return and the catch path and re-throws the parked error, so a catching body cannot report a `complete`. Optional-call, like the existing `_takePendingSuspend?.()`: an npm client that predates the hook keeps today's behavior instead of erroring.
- **`python-client/wmill/wmill/client.py`** — `_execute_directly` wraps a raised task body in `_StepFailure`, a `BaseException` subclass (so `except Exception:` cannot catch it, same device as `_StepSuspend`); `_run_workflow_async` unwraps it and re-raises the step's own exception, which the generated Python wrapper's `except BaseException` turns into a failed child job with the original name/message/traceback. No backend change is needed for Python.

The two clients are not equally airtight: TypeScript's parking + wrapper drain makes the failure unswallowable by *any* catch, while Python's relies solely on `BaseException`, so a bare `except:` (or `gather(..., return_exceptions=True)`) still erases it. That is the same exposure `_StepSuspend` has had since #10348, so it is not a regression, and the `_StepFailure` docstring now says so rather than implying parity.
- Regression test in each client's workflow suite.

### Options considered

1. **Park the failure so it escapes the body's catch** — chosen. Mirrors #10348; degrades gracefully in both directions of SDK/backend version skew.
2. *Have the child wrapper detect that the executing step raised, regardless of what the body returned* — the wrapper cannot know without SDK cooperation, so it reduces to option 1 anyway.
3. *Do nothing and document `except TaskError` / `e.name !== "TaskError"`* — leaves a silent data-loss default; the failure disappearing is not something a user can be expected to defend against.

## Test plan

Verified end to end on a dev backend (bun **and** python WAC jobs, parent + child rounds inspected), with the SDK changes actually installed — the published npm/PyPI packages are what jobs install, so the fixed clients were staged into an isolated `WINDMILL_DIR` cache and the TS one was checked against a real `tsdown` build of the modified `client.ts`.

- [x] Repro before the fix, both languages: parent `{"caught": null}`, child **success** returning `{"caught":"yes"}`
- [x] After: child **fails** with the task's own error (`TypeError: nope` / `ValueError: nope`, traceback pointing at the user's `raise`); parent result is the catch branch's value produced on the *replay* round, with `e.name === "TaskError"`
- [x] Retry-loop shape (body catches `TaskError` and re-dispatches with different args) still completes: `{"value":"ok on 3","errors":["TaskError: …","TaskError: …"]}` in both languages
- [x] Catch-then-continue (`catch {} ; await other()`) — the parked failure is re-thrown at the next SDK call, the child fails, and the parent still runs `other()` on replay
- [x] Uncaught failure still fails the parent with the nested child error; `Promise.all` / `asyncio.gather` fan-out with one failing task behaves the same
- [x] Backward compat: pristine (pre-fix) `windmill-client@1.773.0` against the new backend wrapper runs without error, falling back to the old behavior
- [x] `bun test typescript-client/tests/workflow.test.ts` → 90 pass; each new test verified to fail without its fix (the catch-then-continue guard races against a 500ms deadline, so a regression fails it instead of wedging the suite on the never-resolving thenable)
- [x] `pytest python-client/wmill/tests/` → 96 pass (1 pre-existing unrelated failure, `test_preserves_function_name`, fails identically on `main`)
- [x] `python system_prompts/generate.py` → no generated-file drift

### Review nits dismissed

- *The TS regression test exercises the test file's inline mirror of the SDK, not `client.ts` itself.* That is pre-existing structure — all 89 tests in `workflow.test.ts` run against the mirror because `client.ts` imports the codegen'd `src/services.gen` and cannot be imported by a standalone `bun test` without running `build.sh` first. Restructuring the suite is out of scope here; the shipped `client.ts` path is what the end-to-end runs above covered.

---
## Summary by cubic
Fixes a WAC v2 bug where a child round could swallow the executing task’s failure when the workflow body used a broad try/catch. The child now fails with the task’s error, and the parent receives a `TaskError` so user error handlers run as expected.

- Bug Fixes
  - TypeScript client: park executing-step failures in `_pendingStepFailure` and rethrow on the next SDK call; `task()` routes non-`StepSuspend` errors via `_raiseStepFailure`.
  - Bun wrapper: after return/catch, call `ctx._takePendingStepFailure?.()` and rethrow so a catching body cannot report `complete`; still honors `_takePendingSuspend?.()`. Optional calls for version skew.
  - Python client: wrap direct execution errors in `_StepFailure` (`BaseException`) and unwrap/re-raise in `_run_workflow_async`; no backend change needed.
  - Tests: added regression coverage in TS and Python, including a catch-then-continue test that ensures the failure is re-raised at the next SDK call (no child hang); verified graceful compatibility with `windmill-client@1.773.0`.

<sup>Written for commit 65c71324396d0cdc8d3dec84d8f48b77546a2007. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-worker/src/bun_executor.rs
+++ b/backend/windmill-worker/src/bun_executor.rs
@@ -1800,12 +1800,9 @@ pub async fn handle_bun_job(
         };
 
         // Kept comment-free — this string is written out per job.
-        // `_takePendingSuspend` returns a StepSuspend the body caught and swallowed
-        // (it is an `Error`), so honour it instead of reporting a `complete` whose
-        // step never reached the checkpoint. `_takePendingStepFailure` does the same
-        // for the exception raised by the step a child round executes directly: it is
-        // the round's result, not something the body may handle, so a broad catch must
-        // not turn it into a successful `complete`. Optional: npm clients may predate them.
+        // `_takePendingStepFailure` / `_takePendingSuspend` hand back what the body
+        // caught and swallowed; honour them instead of reporting a `complete` (see
+        // `_pendingStepFailure` in client.ts). Optional: npm clients may predate them.
         let wrapper_content = if is_wac_v2 {
             format!(
                 r#"
--- a/backend/windmill-worker/src/bun_executor.rs
+++ b/backend/windmill-worker/src/bun_executor.rs
@@ -1802,7 +1802,10 @@ pub async fn handle_bun_job(
         // Kept comment-free — this string is written out per job.
         // `_takePendingSuspend` returns a StepSuspend the body caught and swallowed
         // (it is an `Error`), so honour it instead of reporting a `complete` whose
-        // step never reached the checkpoint. Optional: npm clients may predate it.
+        // step never reached the checkpoint. `_takePendingStepFailure` does the same
+        // for the exception raised by the step a child round executes directly: it is
+        // the round's result, not something the body may handle, so a broad catch must
+        // not turn it into a successful `complete`. Optional: npm clients may predate them.
         let wrapper_content = if is_wac_v2 {
             format!(
                 r#"
@@ -1847,6 +1850,10 @@ async function run() {{
     try {{
         const result = await workflowFn(...argsArr);
         setWorkflowCtx(null);
+        const failed = ctx._takePendingStepFailure?.();
+        if (failed) {{
+            throw failed.error;
+        }}
         const swallowed = ctx._takePendingSuspend?.();
         if (swallowed) {{
             throw swallowed;
@@ -1874,6 +1881,10 @@ async function run() {{
                 return {{ type: "sleep", key: dispatch.key, seconds: dispatch.seconds }};
             }}
             return {{ type: "dispatch", mode: dispatch.mode ?? "sequential", steps: dispatch.steps ?? [] }};
+        }}
+        const failed = ctx._takePendingStepFailure?.();
+        if (failed) {{
+            throw failed.error;
         }}
         throw e;
     }}
--- a/python-client/wmill/tests/test_workflow.py
+++ b/python-client/wmill/tests/test_workflow.py
@@ -407,10 +407,8 @@ class TestChildMode:
         assert result["result"] == 14
 
     def test_child_cannot_swallow_the_failure_of_the_step_it_executes(self):
-        # The failure of the step a child round runs directly IS that round's
-        # result. If `except Exception` could catch it, the child would report
-        # a success returning "swallowed", the parent would record that as the
-        # step's value, and `await boom()` there would never raise.
+        # If `except Exception` could catch it, the child would report a success
+        # returning "swallowed" and the parent would record that as the step's value.
         @task
         async def boom():
             raise ValueError("nope")
--- a/python-client/wmill/tests/test_workflow.py
+++ b/python-client/wmill/tests/test_workflow.py
@@ -406,6 +406,26 @@ class TestChildMode:
         assert result["type"] == "complete"
         assert result["result"] == 14
 
+    def test_child_cannot_swallow_the_failure_of_the_step_it_executes(self):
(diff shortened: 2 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 65c7132439a4a01e3a2ac1a55b8cd2d0b78d2d4b — nit-only verdicts at 23f6c5a21937b51d2896fd54f23d376a9e3c8196 (codex "Good to merge", pi "Good to merge", claude "Mergeable, but should ideally address nits"); both nits addressed in 65c7132439a4a01e3a2ac1a55b8cd2d0b78d2d4b — added a catch-then-continue regression test pinning the `_rethrowSwallowed()` re-raise for both a task dispatch and a `sleep()` (raced against a deadline so a regression fails fast instead of hanging), and noted in the `_StepFailure` docstring and the PR body that Python's protection is narrower than TypeScript's (a bare `except:` still swallows, same as `_StepSuspend`).

rubenfiszel:
Correction to the marker above — the head SHA was mistyped. The correct one is:

✅ Review round clean @ 65c71324396d0cdc8d3dec84d8f48b77546a2007 — nit-only verdicts at 23f6c5a21937b51d2896fd54f23d376a9e3c8196; nits addressed in 65c71324396d0cdc8d3dec84d8f48b77546a2007.

rubenfiszel:
✅ Review round clean @ 65c71324396d0cdc8d3dec84d8f48b77546a2007

PR #10367, opened 2026-07-27
fix(wac): return the checkpointed value from step(), not the live object

rubenfiszel:
## Summary

A WAC v2 `step()` handed the workflow body a **different value in the round that ran it than in every replay of it**.

A `@workflow` body re-executes from the top every round. In the round that actually runs a `step()` body, the inline fast path POSTed the checkpoint and then returned the **in-memory** object; every later round returned the value read back from the checkpoint, i.e. its JSON form. So any value JSON does not round-trip changed type between rounds.

This is the same hazard class as [#10348](https://github.com/windmill-labs/windmill/pull/10348) fixed for *failures* — there the SDK deliberately raises the exception a replay would rebuild, so a handler cannot branch on something only one round has. The value path never got the same treatment, and its blast radius is wider: it affects every workflow returning a rich type, not just those catching errors.

**It does not merely crash later — it can deadlock the workflow.** A body that branches on the type dispatches a different task in the round that ran the step than in the replay, so the child job (spawned with `_executing_key` set) never reaches its own key and parks on the never-resolving promise. Reproduced on a dev backend, before this PR:

```
parent round 1:  BRANCH date   → dispatches isDate
child job:       BRANCH string → walks the notDate branch, never reaches its key
                 → parent and child both still running at 45s (force-cancelled)
```

After: `BRANCH string` in both rounds, child completes, workflow returns `"took the string branch"`.

## Changes

The fast path now returns exactly what it checkpointed — the analogue of "raise what a replay would rebuild".

- **`python-client`**: encode the payload once with `json.dumps(..., default=str)` — the encoder the worker wrapper already uses on the suspend path (`python_executor.rs`) — POST those bytes with `content=`, and return `json.loads(payload)["result"]`.
- **`typescript-client`**: encode once with the same `undefined → null` replacer the bun wrapper uses (`bun_executor.rs`), and return `JSON.parse(body).result`. A serialization failure (circular ref, bare BigInt) falls through to the suspend path as before.
- Docstrings on both `step()`s state the contract; `system_prompts/auto-generated/` regenerated so the agent-facing SDK docs carry it too.
- One round-parity test per SDK. The TS one imports the real `client.ts` (with the two generated modules stubbed) rather than `workflow.test.ts`'s inline mirror, which has no fast path and so could not pin this.

Two side effects, both fixing an arm-vs-arm disagreement that `wac.rs` already documents as an invariant:

- Python: a `datetime`/`set` result used to make `httpx`'s encoder raise, silently downgrading that step to a full suspend round. It now goes through the fast path — one fewer worker round-trip per such step.
- TypeScript: the fast path used to *drop* `undefined`-valued object properties while the wrapper kept them as `null`; the two arms now record the same value.

## Behavior change

Returning a non-JSON-native value from `step()` and using it **in the same round** now yields the serialized form.

This restores the pre-fast-path contract rather than inventing one: before #8807 (v1.683.0) `step()` always suspended, so it always returned the checkpointed value. Any code relying on the live object was already broken on every replay — and any workflow that dispatches a task, sleeps, or waits for approval after the step *does* replay. There is deliberately no flag to bring the live object back: `WM_WAC_INLINE_FAST_PATH=0` also yields the checkpointed value. Worth a release-note line.

## Test plan

- [x] `bun test typescript-client/tests/workflow_inline_fast_path.test.ts` — 5/5; 4 of them fail against the pre-fix client (`Date`, `Map`, `Set`, `undefined` prop)
- [x] `pytest python-client/wmill/tests/test_workflow.py` — 96 pass (the one failure, `test_preserves_function_name`, is pre-existing on `main`); the new parity case fails against the pre-fix client
- [x] E2E bun, dev backend, A/B/A on the installed SDK: original client observes `Date/Map` in the running round and `String/Object` in the replay; patched client observes `String/Object` in both
- [x] E2E bun branch-divergence: hangs before (above), completes after
- [x] E2E python (worker built with `--features python`): `OBSERVED str/tuple` then `str/list` before, `str/list` in both rounds after — and 2 worker rounds instead of 3
- [x] `npx tsc --noEmit` — no new errors in `client.ts` (only the pre-existing missing-`@types/node` ones)
- [x] `system_prompts/check-freshness.sh` — up to date

## Follow-ups (deliberately not in this PR)

- `backend/tests/python_jobs.rs::test_python_wac_v2_step_inline_fast_path` already asserts legacy ≡ fast-path equivalence and is the right home for an end-to-end guard (add a step returning a tuple and assert its observed type). It installs `wmill` from PyPI, so it can only land **after** the fixed SDK is published — it would be red until then.
- Neither SDK's test suite runs in CI today (the only `bun test` is `cli-tests.yml` over `cli/test/`), so these guards only fire locally.

## Residual, by design

A replay reads the value back out of `jsonb`, which normalizes beyond a local JSON round trip — object key order is not preserved, and integers past `u64` degrade to `f64` (serde_json without `arbitrary_precision`). Both predate this PR; it closes the type gap, not every last difference. The comments say so rather than over-claiming parity.

---
## Summary by cubic
Return the checkpointed JSON from `step()` in WAC v2 and apply the same JSON round trip to `task()` and standalone paths. This fixes type-based branch divergence and makes local runs match deployed behavior, with TS types that reflect the JSON shape.

- **Bug Fixes**
  - Fast path pre-serializes the payload and returns the decoded checkpointed result; `bigint`→string, top-level `undefined`/functions/symbols/classes→null; only the top-level result is normalized (nested unchanged). Non-finite numbers decode to `null` at runtime and remain typed as `number`.
  - Child-job `step_complete` results are normalized so deployed jobs can’t fail to parse. Standalone `step()` and `task()` round-trip results in both SDKs (incl. async Python tasks) for local/deployed parity. Tests pin round-parity and standalone-parity; system prompts regenerated.

- **Migration**
  - `step()` returns `Promise<Jsonified<T>>`; `task()` returns `JsonifiedFn<T>`. Update call sites that relied on live objects.
  - TypeScript: `Jsonified<T>` keeps `unknown` as `unknown`, converts `bigint` to `string`, and makes keys optional if `JSON.stringify` may omit them (e.g., unions with `symbol`/functions/classes). Non-finite numbers come back as `null`. No flag to restore live objects; `WM_WAC_INLINE_FAST_PATH=0` also returns the checkpointed (JSON) value.

<sup>Written for commit fa4262d41df852994ac647ffe27c56aaafe5a0ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

**Update — `Jsonified<T>` landed in this PR** (was listed as a follow-up).

`step<T>()` now returns `Promise<Jsonified<T>>`, a conditional type mirroring what `encodeCheckpointPayload` does at runtime:

```ts
const ts = await step("now", () => new Date())
ts.getTime()      // now a compile error — it used to typecheck and throw at runtime
ts.toUpperCase()  // ✅ it is a string
```

`Date` → `string`, `Map`/`Set` → `{}`, `undefined`/`void`/functions → `null`, `bigint` → `string` (the bun wrapper installs `BigInt.prototype.toJSON`), methods and symbol keys dropped, arrays/tuples/optionality/literal types preserved, `any` stays `any`. Exported from `windmill-client` so callers can name the shape.

Note this makes an *existing* mismatch visible rather than creating one: on every replay the value was already the JSON form while the type said `Date`. Previously `ts.getTime()` worked in the round that ran the body and threw in every replay; now it fails to compile, which is the point.

Scope calls:
- **`task()` / `taskScript()` / `taskFlow()` keep their current types.** Their results are also JSON round trips (they always come from the checkpoint), so the same lie applies — but wrapping `task()`'s return means transforming a whole function type, and the standalone (outside-a-workflow) path genuinely returns the live object. Left as a follow-up rather than widening this PR.
- **The standalone `step()` path is unchanged at runtime** — outside a workflow nothing is checkpointed and nothing replays, so `fn`'s value is returned untouched and the annotation only describes the in-workflow contract. Called out in a comment at the cast.

Verification: 7 type assertions colocated with the runtime cases in `workflow_inline_fast_path.test.ts`, plus a scratch matrix covering unions, index signatures, nested arrays-of-objects, and `@ts-expect-error` on `ts.getTime()`. `bun test` strips types and tsconfig only covers `src/`, so the command that actually checks them is in a comment above the block. `./build.sh && npx tsc --noEmit` is clean apart from the repo's pre-existing missing-`@types/node` errors.

---

**Update 2 — one rule everywhere.** The remaining carve-outs are gone: a value that crosses a `step()` or `task()` boundary now comes back as JSON in *every* path, and the types say so.

- **`task()` returns `JsonifiedFn<T>`** — `(...args: Parameters<T>) => Promise<Jsonified<Awaited<ReturnType<T>>>>`. Its result was already JSON 100% of the time (the workflow path reads it from the checkpoint, the WAC-v1 path reads it from the API), so the old `T` was never right, and unlike `step()` there was no round in which it was. Argument types stay checked; the one loss is a generic task's type parameters, which are meaningless across a JSON job boundary anyway.
- **Standalone `step()` and `task()` round-trip too** (both SDKs). Outside a workflow nothing is checkpointed and nothing replays, but returning the live object there means testing a workflow locally proves nothing about the deployed run. This matters most for Python, which has no compiler to catch the difference.
- **`taskScript()` / `taskFlow()` unchanged** — they return `(...args: any[]) => PromiseLike<any>`, and `Jsonified<any>` is `any`. Nothing to fix, noted so it doesn't read as an oversight.

The sharp edge, for whoever reads the diff: Python's `task` wrapper is **sync**, so an `async def` task hands back a *coroutine* at that point. Round-tripping it there would serialize `"<coroutine object …>"`; it has to happen after the await, inside a wrapper coroutine. There's a test pinning exactly that.

Re-verified end to end against a **real `tsdown` build** of the client installed into the worker's dependency cache (not a hand-patch): the branch-divergence workflow completes with `"took the string branch"`, and the round-parity workflow reports `String/Object` in both rounds. Suites: 147/147 TS, 97 Python (+1 pre-existing unrelated failure), type assertions clean, `system_prompts` fresh.


rubenfiszel:
✅ Review round clean @ 4ffd0b9590833d3775f418d02b13ca7e1af6adba — nit-only verdicts at 81b3c36c08995489534bed2e8d03ec5f6842342e (codex: mergeable/nits, claude: mergeable/nits, pi: good to merge); nits addressed in 4ffd0b9590

rubenfiszel:
Both round-6 nits are accurate diagnoses; resolving them differently than suggested.

**Non-finite numbers** — fixed as documentation + a runtime test (`e9ce03b`), not by widening the type. Making `Jsonified<number>` = `number | null` would force a null check on *every* numeric field of every step and task result, to model a value almost nobody returns and that already decodes to null identically on all four paths (fast, suspend, standalone, replay). That trade costs far more than it catches, and it is the same call `type-fest`'s `Jsonify` makes. The behavior is now pinned by `["non-finite", () => ({ nan: NaN, inf: Infinity }), { nan: null, inf: null }]` in the parity table and stated in the type's doc comment.

**Custom `toJSON()` returning a bigint** — dismissing: the two requirements are mutually exclusive. The replacer cannot distinguish a bigint that came out of someone's `toJSON()` from a plain bigint property — both reach it as the same value — so matching the wrapper exactly means dropping the bigint branch entirely, which reintroduces the round-2 P1 (`step("n", () => 1n)` throwing in the standalone path, where no `BigInt.prototype.toJSON` exists). Given the choice, the SDK being a superset is the better resolution: for `{ toJSON: () => 1n }` the fast and standalone paths return `"1"` as `Jsonified` promises, while the suspend arm throws — a pathological input (`toJSON` is contracted to return a JSON-representable value) failing loudly on one arm, rather than a common input throwing on all of them.

rubenfiszel:
✅ Review round clean @ e9ce03bd3d23bc4b95eab78a3815caf20353be05 — nit-only verdicts at 40570d6d8851986db29f96ec7764dde9cb394b0e (claude: good to merge, pi: good to merge, codex: mergeable/nits); nits addressed in e9ce03bd3d / dismissed with reasoning in the review reply above

rubenfiszel:
Addressed the round-8 nit in `6ca31693f0`: an omittable key is now **optional**, not merely `| undefined`.

Codex is right that the two are not the same — `{ tag: string | undefined }` still demands the property be present, so a runtime `{}` did not satisfy the declared type. `Jsonified` now splits the object branch into keys that always survive (required) and keys that merely might (optional), flattened back into one shape:

```ts
Jsonified<{ tag: string | symbol; keep: number }>  //  { keep: number; tag?: string }
const v: Jsonified<{ tag: string | symbol; keep: number }> = { keep: 1 };  // now typechecks
```

That assignment is in the test file as the assertion that would have caught the previous shape. Three cases, all pinned: a key that can *only* hold an unrepresentable value is gone entirely, one that might is optional, one that never can is untouched — and an already-optional key stays optional (`{ a: number; cb?: () => void }` → `{ a: number; cb?: null }`).

154/154 TS tests, type assertions clean, `system_prompts` fresh.

rubenfiszel:
Round-9 nits: one fixed in `fa4262d41d`, one dismissed.

**Construct-only functions — fixed.** Real gap, and I confirmed it before patching: `{ Klass: Foo, keep: 1 }` encodes to `{"keep":1}` because a class is a function at runtime, while its *type* has only a construct signature and so never matched `(...args: any[]) => any`. `JsonDropped` now includes `abstract new (...args: any[]) => any`, which covers classes and abstract classes both. Pinned by a type assertion and a runtime case in the parity table. The runtime side needed nothing — `checkpointableResult` already keys off `typeof value === "function"`.

**Overloaded tasks — dismissing.** Accurate, but TypeScript offers no mechanism: `Parameters<T>`/`ReturnType<T>` resolve to the last overload, and there is no way to map over a set of overloads and rebuild them. The only alternative is reverting `task()` to return `T`, which restores a type that is wrong 100% of the time (a task's result *always* crosses JSON) to preserve overload precision in a case that cannot survive that crossing anyway. Now documented on `JsonifiedFn` alongside the generic-erasure caveat, so the limitation is stated where someone would hit it.

155/155 TS tests, type assertions clean, `system_prompts` fresh.

rubenfiszel:
✅ Review round clean @ fa4262d41df852994ac647ffe27c56aaafe5a0ca — nit-only verdicts at 6ca31693f0b7399778597c25aca289a2fb7438f6 (claude: good to merge, pi: good to merge, codex: mergeable/nits); nits addressed in fa4262d41d / dismissed with reasoning in the reply above

PR #10368, opened 2026-07-28
fix(wac): one failure record for tasks and steps, in every round

rubenfiszel:
## Summary

Fixes WIN-2256.

A `@workflow` body is re-executed from the top on every round, so `except` / `catch` is
control flow that has to be deterministic: a handler that branches on the failure it
caught must be handed the same thing in the round that ran the failing body and in every
round replaying it. Two things broke that:

- a failed `step()` recorded `result: {"error": "<msg>", "type": "<Class>"}` while a failed
  task recorded the child job's `result: {"error": {"name", "message", "stack"}}` — one
  error class with two different structures inside;
- the round that ran the body chained the original exception onto `__cause__` / `cause`,
  and a replay, which rebuilds the failure from the checkpoint alone, had nothing to chain.
  The same handler saw `__cause__` set in one round and `None` in the next.

#10348 made step failures checkpointed, which is what brought both within reach.

## Approach

The failure record is decided in **one** function, `wac_failure_record` in
`windmill-common/src/wac.rs`, and everything routes through it:

- **task failures** — `result_processor.rs` calls it instead of assembling the record inline
- **step failures** — `persist_inline_checkpoint_delta` normalizes any `__wmill_error`
  payload before storing it, which covers both the API fast path and the worker suspend
  path, since they already shared that helper
- **the live round** — `POST /jobs/wac/inline_checkpoint/{job}` now returns the record it
  stored, and both SDKs raise from that rather than from their own copy

That last point is what makes the two rounds identical *by construction* rather than by two
codebases imitating each other. The first draft of this PR normalized client-side in each
SDK, and review found them already disagreeing about `name` and about `stack` formatting —
which is the argument for not doing it that way.

What a caught failure carries, from a task or a `step()` alike: `step_key`, `child_job_id`
(absent for a `step()`, which runs in the workflow job), a `message` that is the failure's
own message, and `result` = `{"error": {"name", "message", "stack"?}}`. Extra keys the
failing side attached to its error are passed through, so a custom error's fields survive.

## Changes

- `wac_failure_record` + `is_wac_failure` in `windmill-common/src/wac.rs`; the stack is
  capped at 8 KB, because `persist_inline_checkpoint_delta` rewrites the whole checkpoint on
  every step and an unbounded traceback would be re-serialized once per later step.
- `persist_inline_checkpoint_delta` returns the value it stored; the endpoint echoes it back
  as `{"failure": ...}` for failures only (a successful step's result can be large and the
  SDK already holds it).
- Both SDKs: raise from the echoed record, nothing on `__cause__` / `cause`, and the stack
  is printed to the job log on failure — dropping `__cause__` otherwise left an *uncaught*
  step failure with a job log whose deepest frame is inside the client.
- TS `error.name` now comes from `e.name`, matching what the bun/deno executors record for a
  child job; Python's stack matches the python executor's format (frames only, wrapper frame
  stripped).
- `typescript-client/wacError.ts`: the two functions that decide the shape moved to a
  dependency-free module so the test suite imports the shipped code instead of the inline
  mirror it used to copy. `@internal` in a JSDoc now excludes an export from the generated
  SDK reference, so these two don't show up as user-facing API.
- `system_prompts/base/workflow-as-code.md` documents the contract handlers may branch on.

## Backwards compatibility

SDKs in users' hands post the old `{"error": "<msg>", "type": "<Class>"}`. The normalizer
reads `type` as the `name` fallback, so their failures keep their error class instead of all
becoming `Error` — this was caught by running the **published** SDK against this backend,
and without it every step failure would have degraded on day one. Records checkpointed
before the upgrade replay unchanged; `result` is passed through, never parsed.

## Test plan

- [x] `cargo check` clean on windmill-common / windmill-api / windmill-worker
- [x] 6 new unit tests on `wac_failure_record`: task and step records agree, an unshaped
      child result still yields `name`/`message`, extra error fields survive, oversized
      stacks truncate, legacy `type` markers keep their class
- [x] `bun test typescript-client/tests/workflow.test.ts` — 91 pass
- [x] `pytest python-client/wmill/tests/test_workflow.py` — 96 pass (the one failure,
      `TestTaskDecorator::test_preserves_function_name`, fails identically on `main`)
- [x] **Live instance, published SDK + this backend**: a workflow with a failing `step()`
      and a failing `task()`, through dispatch, checkpoint and replay. Both records come out
      the same shape; this is what surfaced the `type` fallback above.
- [x] **Live instance, this SDK + this backend** (worker dependency cache patched under an
      isolated `WINDMILL_DIR`): the round that ran the body and the replay round now read an
      identical record. A/B/A against the published bundle brings the divergence back, so
      the change is what moved it.

## Not in this PR

Two findings from the same investigation, both pre-existing and shipped separately:
#10366 (a task failure the child round's body catches was not reported at all) and #10367
(a succeeding `step()` returns the live object in one round and its JSON round-trip in the
next). #10367 is still open and touches the same fast-path lines; whichever lands second
needs a rebase.

---
## Summary by cubic
Unifies WAC failure handling so tasks and `step()` return one deterministic failure record across live and replay, and makes SDK failure reporting robust and non-throwing. Fixes WIN-2256 by shaping the record once on the backend and echoing it to SDKs, with safe fallbacks.

- **Bug Fixes**
  - Backend: single `wac_failure_record`; normalizes posted step markers; `stack` capped at 8 KB (bytes); oversized `error.extra` dropped with `extra_omitted: true`; legacy markers stored untouched; inline checkpoint persists then echoes `{"failure": ...}` for failed steps; `persist_inline_checkpoint_delta` returns the stored record.
  - Worker: task failures now built via `wac_failure_record`, so task and step failures share the same fields (`step_key`, optional `child_job_id`, `message`, `result.error { name, message, stack?, extra? }`).

- **SDKs**
  - TypeScript: throws from the echoed record; falls back to the checkpoint when echo is absent; if a JSON body is present but unreadable, suspends so the next round reads the stored record. Failure reporting is hardened (safe against proxies, throwing accessors, circulars, non-finite numbers); only JSON-safe own props kept under `error.extra`, and the serialized snapshot that passed the probe is kept. Non-`Error` throws are normalized; suspend detection hardened via `isSuspendSignal`; step traces printed to job logs. New `wacError.ts` exposes `isSuspendSignal`, `stepErrorMarker`, and `taskErrorFromMarker` as `@internal`; shipped to `jsr`, excluded from docs/prompts, and the module surface trimmed to used exports.
  - Python: raises from the echoed record; same fallbacks and suspend behavior. Failure reporting is hardened and non-throwing; only JSON-safe custom attrs kept under `error.extra` (`json(default=_safe_str, parse_constant=lambda c: c)`); stacks match the executor (frames only, wrapper frame dropped).

<sup>Written for commit cd9ea675f65e15feaf08ef13a4974409eaa1bbc3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
On the unconditional stack print (claude, P2): keeping it deliberately.

A retried \`step()\` does dump one traceback per attempt into a run that ultimately succeeds, which is real noise. But the alternative loses the case the print exists for: dropping \`__cause__\` means an *uncaught* step failure otherwise reaches the log with its deepest frame inside the client and the user's failing line nowhere in it. As the review notes, there is no way to know at raise time whether the handler will catch, and a wrong guess in that direction is silent — the user sees a traceback that points at our SDK. Noise in a succeeding retry loop is visible and harmless by comparison, and the banner names the step so it reads as an attempt rather than a failure.

The stack is also in \`result.error.stack\` either way, so this only governs what reaches the log.

rubenfiszel:
✅ Review round clean @ cd9ea675f65e15feaf08ef13a4974409eaa1bbc3

PR #10371, opened 2026-07-28
fix: surface the real postgres error when data table migrations fail

rubenfiszel:
## Summary

Creating a schema (or any DDL) in a data table with migrations enabled could fail with:

```
Internal: Failed to ensure _wm_migrations table: db error @datatable_migrations.rs:235:13
```

`db error` is the entire message — the Postgres error is lost, so the report is undiagnosable.

Two distinct defects:

1. **The message is swallowed.** The pinned rust-postgres fork renders only the error *kind* in `Display` and exposes the cause solely through `source()` (upstream crates.io appends it), so every `format!("{e}")` on a `tokio_postgres::Error` prints `db error`. This regressed with the fork pin in #9106 (v1.700.0) and affects the whole data table surface, not just migrations (`Failed to query schemas: db error`, `Failed to query columns: db error`, …). The pg job executor was unaffected because it goes through `to_anyhow` → `Error::Anyhow`'s `{error:#}` chain rendering.

2. **The bookkeeping table needs CREATE even when it already exists.** `CREATE TABLE IF NOT EXISTS` checks CREATE on the schema *before* it checks existence, so a data table whose role only holds DML grants could never run migrations — not even against a `_wm_migrations` an operator had pre-created. Reproduced: `permission denied for schema public`. Since Postgres 15 `public` no longer grants CREATE to PUBLIC, so this is the default for any non-owner role on a bring-your-own-Postgres data table. Instance data tables are unaffected (`create_custom_instance_pg_database` runs an explicit `GRANT CREATE ON SCHEMA public TO custom_instance_user`).

## Changes

- Add `windmill_common::error::pg_error_message` — renders a `tokio_postgres::Error` as its Postgres message plus `detail`/`hint`, falling back to the `source()` chain for io/tls/protocol failures.
- Route every data table `tokio_postgres` error through it: `datatable_migrations.rs`, the schema/table/column listing and `create_pg_database` in `workspaces.rs`, and `pg_get_full_schema` in `query_builders.rs`.
- Probe with `to_regclass('_wm_migrations')` before issuing the `CREATE TABLE`, so a DML-only role keeps migrating against an existing bookkeeping table.
- On a genuine `insufficient_privilege`, append what to grant instead of leaving the user with a bare permission error.
- Integration test covering both halves (message reaches the caller; run succeeds on a pre-created table with a role that cannot create).
- New `test_datatable_connection` endpoint + **Test connection** action in workspace settings, reporting the connected role, its default schema, whether it can create tables/schemas, and the exact `GRANT`s for anything missing.

## Test plan

- [x] `cargo test --test datatable_migrations_grants` passes; the same test on unfixed code fails with the exact reported string (`db error @datatable_migrations.rs:235:13`).
- [x] Reproduced end to end against a data table whose role lacks CREATE: creating a schema in the DB Manager now shows the Postgres message and the required grant (screenshot below).
- [x] Happy path unaffected: created schema `analytics` on an owner-role data table — schema created, migration recorded in `datatable_migrations` and `_wm_migrations`.
- [x] With `_wm_migrations` pre-created and DML granted, `run_datatable_migrations` returns `{"applied":[]}` for the unprivileged role instead of failing.
- [x] Test connection exercised in the browser against a role with no `CREATE`: reports both missing grants; after running them, the same button reports healthy (screenshots below).
- [x] `npm run check:fast` clean; backend client regenerated from the updated `openapi.yaml`.

## Preventing the class of failure

Nothing in the data table configuration path checked connectivity or privileges — the first
signal was a Postgres refusal deep inside a migration, weeks after the resource was set up.
So this PR also adds `GET /w/{workspace}/workspaces/test_datatable_connection/{name}` and a
**Test connection** action on each data table in workspace settings. One side-effect-free
round trip reports the role the data table actually connects as, the schema its unqualified
statements resolve to, whether that role can create tables and schemas, and the exact
`GRANT`s for whatever is missing. Existing instances get the same diagnosis with no migration
or config change, since the probe is read-only against whatever is already configured.

### Test connection — missing privileges

dt-test-warning

### Test connection — healthy

dt-test-success

### Data table migration error toast

db manager error toast

---
## Summary by cubic
Surfaces real Postgres errors across data table and Postgres operations, fixes migrations for DML‑only roles by probing `_wm_migrations` before CREATE, and adds an admin‑only “Test connection” that reports exact, ready‑to‑run grants and search_path fixes.

- **Bug Fixes**
  - Added `windmill_common::error::pg_error_message` and used it across data table paths and Postgres helpers (migrations, schema/table/column listing, `pg_get_full_schema`, `create_pg_database`, custom instance DB setup) so users see the real Postgres message.
  - Probe with `to_regclass('_wm_migrations')` before creating the bookkeeping table; on `insufficient_privilege`, show an exact, quoted `GRANT` built from `current_user`/`current_schema()`; tests cover DML‑only roles and message formatting.
  - “Test connection” UI avoids stale reports by keying results by request ID; OpenAPI marks `schema` as nullable and required.

- **New Features**
  - Added `GET /w/{workspace}/workspaces/test_datatable_connection/{datatable_name}` and a “Test connection” button in Data Table settings (admin‑only). Returns `user`, nullable `schema`, `can_create_table`, `can_create_schema`, `migrations_table_exists`, `suggested_grants`, and `suggested_search_path`.
  - Suggestions are derived from `current_user`, `current_schema()`, and `current_database()` and shown even if `_wm_migrations` exists; all identifiers are correctly quoted.

<sup>Written for commit 0a58d3064e58ceb32448c8f83a9bf13a036315a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed all nits from the first round in 1b974c4689:

- **Codex P2 (unit coverage for `pg_error_message`)** — split the helper into `format_db_error` (message + detail + hint) and `error_source_chain`, and added unit tests for both branches in `windmill-common/src/error.rs`. The db-error path stays covered end to end by the integration test.
- **Claude P2 (extended-protocol probe vs #10352)** — the `to_regclass` probe takes no parameters, so it now goes through `simple_query` like `pg_get_full_schema`, with the pooler rationale noted at the site.
- **Claude P2 (adjacent `db error` sites)** — routed the three remaining `tokio_postgres` sites through the helper: the `custom_instance_user` grant error and the replication-grant log in `windmill-api-settings`, and the grant warning in `windmill-common/src/lib.rs`. Confirmed `{e:#}` did not help there (alternate-flag chain rendering is an anyhow behavior, not a `Display` one).
- **Claude P2 (non-atomic `CREATE ROLE`)** — the test's role creation is now `CREATE ROLE` + `EXCEPTION WHEN duplicate_object THEN NULL`, idempotent under concurrent runs on a shared cluster.

`cargo test -p windmill-common --lib error::` (6 passed) and `cargo test --test datatable_migrations_grants` (1 passed) are green.

rubenfiszel:
Both round-3 nits addressed in aefd7185cc:

- **Quoted identifiers in the grant hint** (Codex P2 / Claude P2) — the schema was interpolated raw while the role was quoted, so `Analytics` would fold to `analytics` and `my-schema` would not parse. Both now go through the existing `render_db_quoted_identifier(.., DbType::Postgresql)`, which also escapes embedded `"`. The test asserts the fully quoted statement.
- **Comment at `:273`** (Claude P2) — it claimed the hint "leads the message", which it doesn't; it leads the *explanation*. Reworded to state the constraint that actually matters (keep it ahead of the explanation, because the toast collapses past the first couple of lines). Also dropped the duplicated rationale from the test comment.

`cargo test --test datatable_migrations_grants` passes.

rubenfiszel:
✅ Review round clean @ aefd7185cc22441de5c6024653de1d6ba7feb34f

rubenfiszel:
Both findings fixed in b2f2f15d34.

**[P1] False success when table creation is denied** — correct, and the root of it was in the backend, not just the alert. `suggested_grants` suppressed the schema grant whenever `_wm_migrations` already existed, on the reasoning that nothing needed creating; but an existing bookkeeping table only spares that one table and says nothing about the tables a migration will go on to create. Two changes:
- The endpoint now suggests on the capability alone (`if !can_create_table`), so the grant is reported whether or not the bookkeeping table is there. `migrations_table_exists` is documented as informative only.
- The alert derives its status from `can_create_table && can_create_schema` rather than an empty grant list, and enumerates the three capability fields explicitly so a partial state reads correctly instead of collapsing to a single green/yellow verdict.

Verified against exactly the state you described (`can_create_table:false`, `migrations_table_exists:true`, `can_create_schema:true`), which previously produced an empty grant list and a green result:
```json
{"user":"dt_limited","schema":"public","can_create_table":false,"can_create_schema":true,
 "migrations_table_exists":true,
 "suggested_grants":["GRANT CREATE ON SCHEMA \"public\" TO \"dt_limited\""]}
```
The UI now shows a warning listing `Create tables in public: no / Create schemas: yes / Migration bookkeeping table exists: yes` plus the grant.

**[P2] Integration coverage for the endpoint** — added `test_datatable_connection_report` to `backend/tests/datatable_migrations_grants.rs`, covering: a non-admin gets 403 (the fixture gained a non-admin member for this); the unprivileged role reports both capabilities false with both quoted grants; the pre-created-bookkeeping state still reports `can_create_table:false` **and** still suggests the schema grant (this is the P1 regression guard); and granting both privileges clears the suggestions.

`cargo test --test datatable_migrations_grants` — 2 passed. `npm run check:fast` clean.

rubenfiszel:
All three nits fixed in 25892fd890, plus cubic's frontend race.

- **`CREATE ROLE` race** — you're right that b2f2f15 changed the odds rather than the theory: two `#[sqlx::test]`s in one binary now start within microseconds, and a genuine collision on `pg_authid`'s unique index raises `unique_violation` (23505), which the handler didn't catch. Now `EXCEPTION WHEN duplicate_object OR unique_violation`.
- **NULL `current_schema()`** — agreed, the `public` fallback advised a statement that fixes nothing: an unqualified `CREATE` on such a session fails with `no schema has been selected to create in` whoever holds the grant. The endpoint now suggests nothing in that case and lets `schema: null` carry the diagnosis; the alert explains the empty search_path and suggests `ALTER ROLE … SET search_path` instead.
- **DATABASE grant naming `pg_db.dbname`** — you marked this fine to leave, but it contradicted the principle stated two lines above it (every identifier comes from the server), so I took it: the probe selects `current_database()` and the grant uses that, falling back to the resource only if the column is somehow absent.
- **cubic's stale-panel race** — your read matches mine, the worst case was a correctly-labelled stale panel. It was two lines to eliminate, so a reply for a data table that is no longer the one being awaited is now discarded.

`cargo test --test datatable_migrations_grants` — 2 passed. `npm run check:fast` clean.

rubenfiszel:
Both round-7 nits fixed in 58e8210b98, plus the note.

- **Unquoted `ALTER ROLE`** — correct, and it is aefd718's defect re-introduced one layer up, which is worse for being rendered in `font-mono` next to a select-all block. Took the second option: the backend emits the whole statement as `suggested_search_path` (reusing the `quoted_user` it already had) and the frontend just prints it, so identifier quoting stays in one place instead of gaining a second escape site that can drift. Verified against a role whose search_path resolves to nothing:
  `"suggested_search_path":"ALTER ROLE \"dt_limited\" SET search_path = public"`
- **Prefix-only database assertion** — right, `starts_with` could not tell `current_database()` from the `unwrap_or` fallback, so renaming the alias would silently undo the previous commit with the suite green. Now asserts full equality against the test database name, and the later GRANT reuses that binding rather than re-deriving it.
- **Your "not a finding"** — took it; it was one token. The bookkeeping alternative is gated on `report.schema` too, since that advice does not work without a resolvable search_path (the migration path reaches `_wm_migrations` unqualified as well). Confirmed in the browser: in the NULL-schema state the note no longer renders.

null-schema report

`cargo test --test datatable_migrations_grants` — 2 passed. `npm run check:fast` clean.

rubenfiszel:
P1 fixed in 7010155d97 — you're right, and my previous guard was inadequate for exactly the sequence you named.

Keying on the data table name only rules out a *different* table's reply. In A→B→A both A requests are in flight and both satisfy `connectionCheck?.name === 'A'`, so the first-issued can land last and overwrite the newer result. Replaced with a monotonic request id:

```ts
let latestCheck = 0
const check = ++latestCheck
// … await …
if (check !== latestCheck) return
```

Only the last-issued request can match, in both completion paths, so the name check is subsumed rather than supplemented. I did not stage the timing manually — the property is structural (`latestCheck` is only ever incremented at issue time, and each continuation compares against its own captured value), so a delayed A cannot pass the check once a later request exists.

**Coverage gap you flagged** — added `test_datatable_connection_without_a_resolvable_schema`, which asserts `schema` is null, that no `ON SCHEMA` grant is suggested (no grant fixes an empty search_path), and the exact `suggested_search_path`. The role it uses is deliberately hyphenated — `wm-dtmig-noschema` — so the expected string only matches while the statement quotes the identifier; that is the defect that has now recurred twice, so it's worth a structural guard rather than a manual check. It needs a role of its own because `ALTER ROLE … SET search_path` is cluster-wide and would otherwise leak into the tests running in parallel.

`cargo test --test datatable_migrations_grants` — 3 passed. `npm run check:fast` clean.

rubenfiszel:
Nit fixed in c31728acee — correct, the contract was wrong rather than merely loose.

`schema` is an `Option<String>` with no `skip_serializing_if`, so the endpoint always emits the key and emits `null` when `current_schema()` is null (the new integration test asserts exactly that). Declaring it optional-but-string permitted omission, not null, so the generated client type was `schema?: string` — wrong in both directions. Now `nullable: true` **and** listed under `required`, per your second suggestion, since it is always present:

```ts
// before
schema?: string;
// after
schema: string | null;
```

Client regenerated; `npm run check:fast` clean against the stricter type (the frontend already tested truthiness, which handles null, so no call-site change was needed).

Round 9 verdicts on 7010155d97 were claude and pi "Good to merge", with this as the only outstanding item.

rubenfiszel:
✅ Review round clean @ c31728aceef69df31fef1d259ac46a131bd8a599

PR #10372, opened 2026-07-28
feat: bulk discard selected drafts on the compare & deploy page

Guilhem-lm:
## Summary

The Deploy-draft mode of the Compare & Deploy page (`/forks/compare?mode=draft`) only offered one bulk action: "Deploy N drafts". Cleaning up unwanted drafts meant discarding rows one at a time, with one confirmation each. This PR adds a bulk **"Discard N drafts"** button next to the deploy button, guarded by a single confirmation modal that explicitly lists — by path — every selected item that exists only as a draft, since discarding those is a permanent deletion rather than a revert.

It also fixes the selection gate to match the server's permission model: discarding your own (email-scoped) draft never requires write permission on the path, while deploying does. The former single `isSelectable` predicate is split into `isDiscardable` (own draft, not deployed this session, not a data-pipeline bundle, and — for legacy ownerless drafts only — writable) and `isDeployable` (`isDiscardable` + `can_write`). Each footer button counts its own eligible selection, so "Discard 5 / Deploy 4" can legitimately diverge when a drafted path lost writability; a footer hint explains the gap when it appears.

## Changes

- `CompareDrafts.svelte`: bulk "Discard N drafts" footer button (destructive styling) with a confirmation modal; permanent deletions (draft-only items) listed by path in the modal
- Sequential batch discard with per-row loading/failed status in the existing `deploymentStatus` map, a toast per failure, and a single workspace-drafts invalidation after the batch (`discardDraft` gains an `invalidate` opt-out so N discards don't trigger N overlapping list refetches)
- Selection gate split into `isDiscardable` / `isDeployable`; checkboxes, select-all, and the auto-select default now gate on discardability, with independent footer counts
- Legacy (ownerless) drafts stay write-gated in the UI, mirroring the server's discard check (`is_own_discard` excludes `legacy` in `drafts.rs`), with matching disabled-checkbox/row tooltips
- Footer hint when selected drafts are discardable but not deployable (no write permission on the path)

## Test plan

- [x] Bulk discard of draft-only items: modal lists them as permanently deleted; confirming removes the draft rows (verified in DB) and the items drop off the list
- [x] Bulk discard of a drafted-over deployed item reverts to the deployed version without touching it
- [x] Cancel leaves everything untouched
- [x] Network: N discards produce exactly N `drafts/update` POSTs and **one** final `drafts/list` refetch (not one per row)
- [x] Failure path: a failed row keeps a failed badge + toast while later rows still process
- [x] Buttons disable at 0 selected / while a deploy or discard is in flight
- [ ] `mine && !can_write` divergence (Discard count > Deploy count + footer hint) — not reproducible as a workspace admin (admins can write every path); verified by code inspection

## Screenshots

Deploy-draft mode with the new bulk Discard button next to Deploy:

compare-drafts-page

Confirmation modal listing draft-only items that will be permanently deleted:

bulk-discard-confirm-modal

### DIFF
--- a/frontend/src/lib/components/CompareDrafts.svelte
+++ b/frontend/src/lib/components/CompareDrafts.svelte
@@ -305,8 +305,8 @@
 		})
 	})
 	// Select all on the first non-empty load (acting on everything is the common
-	// intent);
-	// only once, so a refetch after a deploy doesn't re-select the leftovers.
+	// intent); only once, so a refetch after a deploy doesn't re-select the
+	// leftovers.
 	let hasAutoSelected = $state(false)
 
 	const deploymentStatus: Record<
@@ -499,6 +499,12 @@
 	let bulkDiscardItems = $state<Row[] | undefined>(undefined)
 	let discarding = $state(false)
 	const bulkPermanent = $derived((bulkDiscardItems ?? []).filter(isDestructiveDiscard))
+	// Third outcome the modal must cover: a draft-only item someone else also
+	// drafted isn't deleted — only this user's draft goes; the item survives via
+	// the other drafts.
+	const bulkSharedCount = $derived(
+		(bulkDiscardItems ?? []).filter((i) => i.draft_only && !isDestructiveDiscard(i)).length
+	)
 
 	function onDiscardSelectedClick() {
 		const toDiscard = visibleItems.filter((i) => selectedItems.includes(i.key) && isDiscardable(i))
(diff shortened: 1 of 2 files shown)


Guilhem-lm:
✅ Review round clean @ 26ea108a5417612de4b53d6b15c8899482a5522b — nit-only verdicts at 115c623b7ef0a10cfde09804f1d8a471ae2675d3 (codex: Good to merge, pi: Good to merge, claude: nit-only); nits addressed in 26ea108a54

PR #10373, opened 2026-07-28
fix(ai-agent): keep tool description through flow deployment

### LINKED ISSUE #10244: bug: AI Agent Tool description does not persist
### Describe the bug

The AI Agent Tool description field added in https://github.com/windmill-labs/windmill/issues/10036 appears to be reset after the flow is deployed.
The filed accepts text and keeps it for the draft, but once the flow is deployed the value is lost. In a request, the _tool name_ is being sent in the `description` field as before. If the flow is loaded for editing the field is empty. No value for `description` appears in exported flow.yaml .
No errors visible in the UI or the server logs.

### To reproduce

1. Define an OpenAI agent
2. Add a tool
3. Define non empty description for the tool
4. Deploy the flow
5. Open the flow for editing
6. Observe the tool description field being empty

### Expected behavior

The value set for the field should persist through deployment and should be used in the requests.

### Screenshots

_No response_

### Browser information

Microsoft Edge	150.0.4078.83

### Application version

CE v1.764.0

### Additional Context

_No response_

rubenfiszel:
## Summary

Fixes #10244. An AI agent tool's `description` (added in #10083) was silently wiped every time the flow was deployed: the field survived in the draft but came back empty after deploy, never reached `flow.yaml`, and the model kept receiving the tool *name* as its description.

Root cause is in the flow dependency job. To lock a tool's raw script, `lock_modules` projects each `AgentTool` down to a `FlowModule`, locks it, then rebuilt the tool with `tools[idx] = locked.into()`. `description` lives only on `AgentTool` — `FlowModule` has no counterpart — so that `From<FlowModule> for AgentTool` impl hardcoded `description: None`, and the round-trip dropped the user's text on every deployment.

## Changes

- Replace the lossy `From<FlowModule> for AgentTool` impl with `AgentTool::update_from_module`, which folds the locked module back into the existing tool in place and therefore keeps tool-level fields that have no module counterpart. Removing the `From` impl also stops the lossy conversion from being reintroduced by an innocent `.into()`.
- Update the single call site in `lock_modules`.
- Regression test pinning the `AgentTool -> FlowModule -> (locked) -> AgentTool` round-trip: the description survives and the locked value is adopted.

## Test plan

- [x] `cargo test -p windmill-types --lib agent_tool_keeps_description_through_locking`
- [x] End-to-end on a local dev instance: create a flow with an AI agent whose tool is a raw bun script and a non-empty tool description, let the dependency job deploy it, then re-read the deployed flow. Before the fix the tool comes back as `{"id","value","summary"}` with no `description`; after the fix `"description": "MY TOOL DESCRIPTION"` is still there.
- [x] Same flow reopened in the flow editor — the tool description field is populated instead of showing its placeholder.

## Screenshots

Deployed flow reopened in the editor, tool selected.

Before (description lost on deploy — field shows only its placeholder):

before-tool-description-lost

After (description persisted through deployment):

fixed-tool-description

Note: descriptions already lost by a previous deployment are not recoverable — they have to be re-entered once.

---
## Summary by cubic
Fixes a bug that cleared AI agent tool descriptions after flow deployment. Descriptions now persist through dependency locking and deployment.

- **Bug Fixes**
  - Replaced lossy `From<FlowModule> for AgentTool` with `AgentTool::update_from_module` to retain `description`.
  - Updated `lock_modules` in `windmill-worker` to use the new method and avoid `.into()`.
  - Added a regression test to ensure `description` survives and locked values are applied.

- **Migration**
  - Re-enter descriptions for tools deployed before this fix; previously lost text cannot be recovered.

<sup>Written for commit 7abe60914e132d64ae8d3ed504f5d463b6457a02. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-types/src/flows.rs
+++ b/backend/windmill-types/src/flows.rs
@@ -875,19 +875,17 @@ pub struct AgentTool {
     pub value: ToolValue,
 }
 
-// Convert FlowModule -> AgentTool
-impl From<FlowModule> for AgentTool {
-    fn from(flow_module: FlowModule) -> Self {
+impl AgentTool {
+    /// Fold a `FlowModule` that went through dependency locking back into the tool it came from.
+    /// `description` has no `FlowModule` counterpart, so it must be carried over from the
+    /// existing tool: rebuilding an `AgentTool` from the module alone drops it on every deploy.
+    pub fn update_from_module(&mut self, flow_module: FlowModule) {
         let module_value = serde_json::from_str::<FlowModuleValue>(flow_module.value.get())
             .unwrap_or(FlowModuleValue::Identity);
 
-        AgentTool {
-            id: flow_module.id,
-            summary: flow_module.summary,
-            // FlowModule has no dedicated tool description; it is carried on AgentTool only.
-            description: None,
-            value: ToolValue::FlowModule(module_value),
-        }
+        self.id = flow_module.id;
+        self.summary = flow_module.summary;
+        self.value = ToolValue::FlowModule(module_value);
     }
 }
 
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 7abe60914e132d64ae8d3ed504f5d463b6457a02

PR #10374, opened 2026-07-28
feat(ai-chat): add get_flow_run_details tool for per-step flow run results

Guilhem-lm:
## Summary

The global AI chat could only see a flow run's top-level result and logs — answering "what did each step of this run do?" required a chain of generic `getJob` API calls, one per step, recursing into every subflow. This PR adds a dedicated chat tool, `get_flow_run_details`, that returns a flow run's whole execution tree — per-step statuses and truncated results, including subflow steps, loop iterations, branches, and retries — in a single call, with an optional `step` parameter to drill into any one step's full result.

## Changes

**Backend**
- New endpoint `GET /w/{workspace}/jobs_u/get_flow_all_results/{id}`: the results twin of `get_flow_all_logs_structured`. Same recursive-CTE enumeration of the flow's job tree, joined against `v2_job_completed`/`v2_job_queue` instead of `job_logs`, so each entry carries status (`success|failure|canceled|skipped|suspended|running|queued`), duration, and a `result_prefix` truncated in-Postgres by `max_result_len` (default 2000, clamped to 30k). Works on running flows.
- Shared helpers extracted from the logs endpoint: `authorize_flow_tree_read` (scope tags, read access, view logging; also returns the enclosing run id when the requested job is itself a subflow step — `COALESCE(flow_innermost_root_job, parent_job)`) and `flow_tree_entry_label`.
- Server-side step resolution: `?step=b/c` / `?step=b[12]/c` resolves the address one indexed `parent_job + flow_step_id` lookup per segment (no tree enumeration) and returns the single resolved job; a miss returns a `step_error` diagnostic listing available step ids or per-iteration statuses.
- Retry attempts are distinguished from loop iterations: sibling jobs of a non-fan-out step (not forloop/whileloop/branchall/branchone/aiagent) are labeled `(attempt i/N)` and the ambiguous-step diagnostic says "was retried — N attempts" instead of "loop/branches".

**Frontend**
- New pure module `chat/global/flowRunTree.ts`: rebuilds the tree from the flat list, groups loop iterations/branches per step (loops render as a tally: `iterations/ok/failed_iterations`, showing ≤3 failed + the latest, rest elided), renders retried steps as the final attempt plus status-only `previous_attempts`, and shrinks result heads (700 → 250 → drop successes → 100) until the whole tree fits a 20k-char budget, keeping failure detail longest. Drill mode is one round trip returning the step's result capped at 12k chars.
- Tool registration + system-prompt guidance in `chat/global/core.ts`; on flow test-run/background-job failures, `shared.ts` appends a hint pointing at `get_flow_run_details` (gated to the chats that have the tool).

## Test plan
- [x] `cargo check` + `SQLX_OFFLINE=true cargo check` clean; sqlx cache updated (5 new entries, 1 stale removed)
- [x] `npm run check` 0 errors; 7/7 vitest unit tests on `flowRunTree.ts` (nesting, subflow rendering, loop tally + elision, retry-as-attempts, running iterations, root-job note, shrink ladder)
- [x] curl against fixture flows: tree with subflow + failing loop iteration; step drills `b/c`, `l[3]`, `a[3]` (retry); diagnostics for ambiguous (`l`, retried `a`), unknown (`z`, `b/z`), and out-of-range (`l[9]`) addresses; enclosing-run link from a subflow job
- [x] Playwright E2E in the sessions global chat: per-step breakdown of a run answered with one tool call; `b/c` and `l[3]` full-result drills; retried run correctly described as "no loop — 3 attempts"

## Screenshots

Drill-down: the model fetches two step results (subflow step `b/c` and failed loop iteration `l[3]`) via server-side address resolution:

flow-run-details-drill

Retry semantics: a step that exhausted its retry policy is reported as one step with 3 attempts, not a loop:

flow-run-details-retry


Guilhem-lm:
Addressed the round in d39b1c61f9:

- **Codex P1 (scope-tag bypass)**: `authorize_flow_tree_read` now returns the caller's scope tags and every descendant query re-applies them — the tree CTE's recursive member, the step resolver's sibling lookup, and the available-steps diagnostic. Verified live with a token scoped to `if_jobs:filter_tags:flow` against a `flow`-rooted run: the tree returns only the root + flow-tagged subflow job, and drilling a pruned step answers "No step \"a\" at this level. Available steps: b."
- **Codex P1 (branchone mislabel)**: the label no longer claims which branch ran — branchone children are labeled `(selected branch)` (branchone only enqueues its chosen branch, so `sibling_index` can't identify it). Pinned by a new Rust unit test.
- **Claude P1 (unbounded enumeration)**: the tree query now has a hard `LIMIT` of 2000 entries (+1 to detect overflow); the response carries `truncated: true` and the tool result notes that tallies may undercount. Step-drill mode is unaffected (no enumeration).
- **Claude P2 (`""` classification)**: all three paths now agree — an unresolved parent module type keeps the generic iteration wording; only a known non-fan-out type marks siblings as retry attempts. Covered by the new `flow_tree_entry_label` unit tests.
- **Claude P2 (skipped as unfinished)**: loop tallies now report `skipped` separately; `unfinished` counts only running/queued/suspended.
- **Claude P2 (`root_job` misnomer)**: renamed to `enclosing_job` across the response, OpenAPI schema, and frontend.
- **Pi nit (branchone rendering)**: partially addressed — the misleading "branch default" label is gone, and the tally's `type` field already carries `branchone`. Not adding a dedicated rendering arm: a normally-executed branchone renders as a plain step (only its chosen branch is enqueued, so there is no multi-node group), and a multi-node branchone group only arises when the branchone step itself was retried — too rare to justify a bespoke shape.

Also added per both reviews' test-coverage notes: Rust unit tests for `flow_tree_entry_label` (fan-out vs retry vs unresolved classification, branchone honesty) and `parse_step_segment` edge cases; frontend tests for skipped tallies and the truncation note (8 vitest + 2 cargo tests passing).

Guilhem-lm:
Round-2 nits addressed in f3e48f5535:

- **Claude P2 (LIMIT below the join)**: the row cap moved into a `limited` CTE before the `v2_job_completed`/`v2_job_queue` join, so at most 2000 rows ever reach the `LEFT(result::text)`/`length(result::text)` projection.
- **Claude P2 (silent scope pruning)**: the response now carries `scope_filtered: true` whenever the caller's token is tag-scoped; the tree prepends a note ("steps running on other tags are omitted") and step-resolution misses append the same caveat, so hidden steps are no longer indistinguishable from steps that never ran.
- **Pi P2 (logs twin parity)**: `collect_flow_log_entries` now applies the same descendant tag filter, closing the pre-existing logs-side gap — verified live: the same `if_jobs:filter_tags:flow` token that sees a pruned results tree now gets the identically pruned `get_flow_all_logs_structured` output.
- **Codex P2 (UTF-16 vs code points)**: result-size comparison and head slicing are now code-point-aware (`countCodePoints` / surrogate-safe cut) in both the tree and drill paths; a regression test pins that a 400-astral-char result reports `result_total_chars: 400` instead of silently hiding the truncation, and that the cut never leaves a lone surrogate.

Dismissed with reasoning:

- **Codex P2 (step addresses can't represent every module id)**: flow step ids are validated by the editor against `^[a-zA-Z][a-zA-Z0-9_]*$` (`frontend/src/lib/components/IdEditorInput.svelte:38`), so `.`, `/`, `[`, `]` cannot occur in ids created through the product — the address syntax has no collision with the actual id charset. Ids smuggled in via hand-crafted API flow payloads remain visible in the tree (with job ids) even if not drillable; supporting an escape syntax for them isn't worth the schema surface.
- **Codex P2 (missing ai_evals A/B)**: deliberately omitted for this change by maintainer decision earlier in development; behavior evidence is the Playwright E2E in the PR body (one-call tree answer, drills, retry semantics). Can be added as a follow-up if the tool's prompt wording needs tuning.

Guilhem-lm:
✅ Review round clean @ f3e48f5535fb6b52ff8b9d17409089f1231f7824 — nit-only verdicts at d39b1c61f91b30d30d4654c06981f24062e11a38; nits addressed in f3e48f5535 / dismissed in review replies

Guilhem-lm:
A final local Codex pass on the ready head surfaced two real P1s, fixed in 0db725b1d9:

- **P1 — sibling addressing relied on non-monotonic ULID order**: parallel iterations created in the same millisecond were shuffled (reproduced: a 20-iteration parallel loop had iterations 5 and 6 swapped, so `l[5]` returned 6). Sibling numbering now comes from the parent flow_status's `flow_jobs` array — the authoritative iteration order — in both tree CTEs and the step resolver, with id-order fallback for retries (which have no `flow_jobs`). Re-verified: all 20 iterations map 1:1 and `l[5]` returns 5.
- **P1 — classification read the mutable current flow definition**: `parent_module_type` resolved through `flow.value` (latest deploy) instead of the job's pinned version. All three lookups now join `flow_version` on the parent's `runnable_id` first (gated to `kind = 'flow'` so flownode ids can't collide). Verified: after redeploying the fixture's step `a` as a forloop, the old retried run still reports `rawscript` / "attempt" labels.
- **P2 — cli/package-lock.json drift**: not part of this PR — it's uncommitted local worktree drift (never staged or pushed), so the PR diff is unaffected.

Guilhem-lm:
✅ Review round clean @ e6272eedc12d1eaf0d0db03a2503ad68aa56f204 — nit-only verdicts at 0db725b1d9cd24c9c6133f878a4b55db40dedcc4 (codex: Good to merge, pi: Good to merge, claude: nit-only); nits addressed in e6272eedc1: the drill's flow_jobs ordinal lookup is now an uncorrelated join (array expanded once, hash-joined) and step diagnostics are bounded (statuses capped at 20 listed, out-of-range renders as a 1..N range).

PR #10375, opened 2026-07-28
fix(datatable): provision the replication user on managed postgres

rubenfiszel:
## Summary

On a custom-instance datatable, postgres triggers connect as `custom_instance_replication_user`. That role was created with `CREATE USER … WITH PASSWORD … REPLICATION`, which requires a real Postgres superuser on PG <= 15 (PG 16+ accepts `CREATEROLE` + the `REPLICATION` attribute). Managed Postgres never hands out a real superuser — on RDS the master user is a member of `rds_superuser` with `rolsuper = f` — so the role is never created there, and the failure only surfaced as a raw `SqlErr: must be superuser to create replication users` when a superadmin clicked "Refresh custom instance passwords".

The role is now created without the attribute, and replication capability is granted in a separate, catchable step: the `REPLICATION` attribute first, falling back to the provider role (`rds_replication`) when the attribute is rejected with `insufficient_privilege`. Two related gaps are closed at the same time: instances whose one-shot migration silently swallowed the failure now repair themselves at boot, and the setup wizard reports the outcome instead of only logging it.

## Changes

- `windmill-common/src/utils.rs`: split provisioning into "create the role + store its password" and "grant replication capability". The capability step tries `ALTER ROLE … REPLICATION` and, on `insufficient_privilege`, grants `rds_replication` when that role exists, otherwise re-raises. `ALTER ROLE custom_instance_user NOREPLICATION` is now guarded on the attribute actually being set, since the clause itself is superuser-only on PG <= 15. Both variants (rotate-always for the refresh endpoint, rotate-only-if-missing for the converge) are one statement and take `pg_advisory_xact_lock`, so concurrent replicas can't race into `CREATE USER`.
- `windmill-common/src/utils.rs`: new `ensure_custom_instance_replication_user`, idempotent and executor-generic.
- `windmill-api/src/live_migrations.rs`: converge the role on every boot for instances that have at least one custom-instance database, logging an actionable error when it cannot be provisioned. Instances without one are untouched and stay quiet.
- `windmill-api-settings/src/lib.rs`: the setup wizard provisions the role on the main pool (roles are cluster-wide) and reports `replication_user` / `replication_user_error` as its own step, replacing two silent `tracing::error!`s. The database stays usable for datatables when only this step fails, so the overall setup still reports success.
- `windmill-common/src/instance_config.rs`, `windmill-api/openapi.yaml`: mirror the two new log fields.
- `CustomInstanceDbWizardModal.svelte`: new step 7 with the commands it runs, the managed-Postgres note, and the underlying error inline.

## Screenshots

Step 7 on a superuser Postgres:

wizard-step7-ok

Step 7 when the role cannot be granted replication (forced locally by raising `insufficient_privilege` with no provider role present) — the step goes red and auto-expands with the Postgres error, the preceding steps stay green:

wizard-step7-fail

## Test plan

Verified locally against PG 18 (dev cluster) and by simulating a managed instance — a `CREATEROLE`-but-not-superuser role, a stub `rds_replication` granted to it with admin option, and a scratch database:

- [x] Old SQL as the simulated managed master: fails, no role created.
- [x] New SQL as the simulated managed master: role created without the attribute, granted `rds_replication` + `custom_instance_user`, password stored. Second run is a no-op and does not rotate.
- [x] Non-superuser with no provider replication role: raises, nothing left half-provisioned.
- [x] Superuser: `REPLICATION` attribute set, `custom_instance_user` left `NOREPLICATION`.
- [x] Through the API: wizard setup of a new custom-instance database reports step 7 green; with the attribute path forced to fail and a stub `rds_replication` present, the same button takes the fallback and grants it.
- [x] Boot converge: dropped the role and its stored password on an instance that has a custom-instance database, restarted the server, both reappear with no error logged.
- [ ] On RDS itself: setup a custom-instance database, confirm step 7 is green and `custom_instance_replication_user` is a member of `rds_replication`, then run a postgres trigger on a datatable in that database and confirm it streams (requires `rds.logical_replication = 1` on the parameter group).

No automated test: the provisioning path is cluster-global SQL, so an integration test would create and rotate the password of the real `custom_instance_replication_user` role for every other database on the same server — including a developer's dev instance. CI also always runs as a superuser, so it would only exercise the path that already worked.

---
## Summary by cubic
Fixes replication user provisioning for custom-instance datatables on managed Postgres (e.g., RDS) so triggers can open replication connections. Adds a setup wizard step and makes boot converge self-heal edge cases.

- **Bug Fixes**
  - Create `custom_instance_replication_user` without `REPLICATION`, then grant it; fall back to `rds_replication` on `insufficient_privilege`. `custom_instance_user` cleanup is now guarded and best-effort (errors don’t roll back provisioning).
  - New idempotent `ensure_custom_instance_replication_user`: atomic, advisory-locked provisioning; converged on boot when any custom-instance DB exists; self-heals a missing or JSON-null stored password.
  - Setup wizard provisions on the main pool and reports a new step with outcome and error details (`replication_user`, `replication_user_error`). UI explains managed-Postgres fallback and that datatables still work if this step fails.

<sup>Written for commit d523161a07f01ad6a756319446a1f1a772b781a9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10375?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-settings/src/lib.rs
+++ b/backend/windmill-api-settings/src/lib.rs
@@ -1506,6 +1506,10 @@ struct CustomInstanceDbLogs {
     db_connect: String,
     #[serde(skip_serializing_if = "String::is_empty")]
     grant_permissions: String,
+    #[serde(skip_serializing_if = "String::is_empty")]
+    replication_user: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    replication_user_error: Option<String>,
 }
 
 async fn list_custom_instance_pg_databases(
@@ -1705,29 +1709,23 @@ async fn setup_custom_instance_pg_database_inner(
             ))
         })?;
 
-    // The replication attribute lives on a dedicated role used by postgres trigger
-    // connections. The getter creates the role (with its stored password) when the
-    // migration couldn't.
-    if let Err(e) = windmill_common::utils::get_custom_pg_instance_replication_password(db).await {
-        tracing::error!("Failed to ensure custom_instance_replication_user exists: {e:#}");
-    }
-    if let Err(e) = client
-        .batch_execute(
-            "ALTER ROLE custom_instance_replication_user REPLICATION;
-             GRANT custom_instance_user TO custom_instance_replication_user;
-             ALTER ROLE custom_instance_user NOREPLICATION;",
-        )
-        .await
-    {
-        tracing::error!(
-            "Failed to grant replication permission to custom_instance_replication_user: {e:#}"
-        );
-    }
-
     logs.grant_permissions = "OK".to_string();
 
     drop(client); // /!\ Drop before joining to avoid deadlock
     windmill_common::shutdown_pg_connection(join_handle).await?;
+
+    // Roles are cluster-wide, so the dedicated role used by postgres trigger connections is
+    // provisioned on the main pool rather than on the new database. Reported as its own step
+    // rather than failing the setup: the database is usable for datatables without it, only
+    // postgres triggers are.
+    match windmill_common::utils::ensure_custom_instance_replication_user(db).await {
+        Ok(()) => logs.replication_user = "OK".to_string(),
+        Err(e) => {
+            tracing::error!("Failed to provision custom_instance_replication_user: {e:#}");
+            logs.replication_user = "FAIL".to_string();
+            logs.replication_user_error = Some(e.to_string());
+        }
+    }
 
     Ok(())
 }
--- a/backend/windmill-api-settings/src/lib.rs
+++ b/backend/windmill-api-settings/src/lib.rs
@@ -1716,8 +1716,8 @@ async fn setup_custom_instance_pg_database_inner(
 
     // Roles are cluster-wide, so the dedicated role used by postgres trigger connections is
     // provisioned on the main pool rather than on the new database. Reported as its own step
-    // rather than failing the setup: the database is usable for datatables without it, only
-    // postgres triggers are.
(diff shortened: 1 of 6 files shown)


rubenfiszel:
✅ Review round clean @ d523161a07f01ad6a756319446a1f1a772b781a9 — nit-only verdicts at cbff3d2c2e118eade26cc5240e3f013de107620a (codex: Good to merge, pi: Good to merge, claude: mergeable with nits); nits addressed in d523161a07.

- `ALTER ROLE custom_instance_user NOREPLICATION` is now wrapped in its own `EXCEPTION WHEN insufficient_privilege THEN NULL` so the cleanup can't roll back the role it just provisioned.
- Rotate condition now mirrors the reader with `AND value #>> '{}' IS NOT NULL`. Small correction to the finding's premise: `#>> '{}'` returns NULL only for a JSON `null`, not for any non-string (an object reads back as its JSON text), so the stranding case is specifically a `null` value — verified on PG 18, and the ensure block now rotates out of it.
- Comment negation and the operator-facing log line fixed; the log now names `rds_replication`/AWS RDS instead of implying generic managed-provider coverage.

PR #10378, opened 2026-07-28
[ee] feat: browse the S3 explorer folder by folder instead of a flat paged list

diegoimbert:
## Summary

The S3 explorer listed object keys **flat**, in pages of 1000 lexicographically-ordered keys, and rebuilt the folder tree client-side. A single folder holding many objects consumed the whole page, so every sibling that sorted after it stayed invisible until you clicked "Load more" repeatedly.

This switches browsing to **delimiter-style listing, one bounded page per request**: the server returns only the direct children of a prefix — subfolder prefixes in a new `folders` field, files in `windmill_large_files` — and the level pages via an opaque `marker`/`next_marker`, with folders and files both counting toward `max_keys`. The picker fetches a folder's children when you expand it, and a level with more entries than one page gets a **"Load more in this folder"** row.

On a bucket with 8 top-level entries where one folder holds 1500 files:

| | top-level entries visible in the first request |
|---|---|
| before | 2 (`README.md`, `aaa_huge_dataset`) — 1000 keys returned, all but one inside the same folder |
| after | all 8 |

## How the backend serves a page

`object_store` exposes delimited listing only as `list_with_delimiter`, which drains the whole level into memory and cannot be resumed — unusable for a stateless paged endpoint. All three cloud stores support exactly what we need natively (delimiter + page size + continuation token: S3 `max-keys`/`continuation-token`, GCS XML API likewise, Azure `maxresults`/`marker`), so this PR carries a **small additive fork of `object_store`** exposing it:

- `ObjectStore::list_delimited_page(prefix, token, max_keys) -> ListPage { common_prefixes, objects, next_token }`
- One native request per page on S3, GCS and Azure; `LocalFileSystem` lists names once and stats only the returned page; a drain-and-slice default implementation covers the remaining stores.
- The crate is vendored at `backend/vendor/object_store` (see “Fork logistics” below): 4 commits on top of the previously pinned rev `36752c9`, mirrored for upstreaming at https://github.com/diegoimbert/arrow-rs-object-store/tree/windmill/list-delimited-page.

The EE handler is then a thin pass-through (−127/+23 in the companion PR): no client-side delimiting, no buffering beyond one page, no fabricated resume keys. Work per request is bounded by `max_keys` **by construction**, paging a level of N entries costs `ceil(N/max_keys)` requests, and the `next_marker` is the store's own continuation token, treated as fully opaque.

Advanced LFS permission globs are written against full object keys and cannot be evaluated for a bare folder prefix, so those setups keep the flat listing — the response omits `folders`, which is how the client detects the fallback.

## Screenshots

Root level — all 8 top-level entries at once (previously only the first two were reachable):

root

Inside the 1500-file folder — "Load more in this folder" at the bottom of that level:

load more

After clicking it — `(1000+ items)` becomes `(1500 items)` and the row disappears:

fully loaded

## Behavior changes worth calling out

- **The "Folder prefix" box now actually filters.** It never did: `rootPath ?? (filter ...)` always took `rootPath`, which is `''` (not nullish), so the filter never reached the server. It is now sent as a real prefix; prefix matching is on whole path segments and the empty state says so.
- **`regexFilter` no longer hides folders** (app image picker). Lazy listing cannot know a folder's contents without opening it. Deliberate.
- Per-folder counts are direct children of the loaded level; `+` means more pages remain at that level. Subfolders surface as a level pages (they share the `max_keys` budget with files), and the picker synthesizes an ancestor from its path when deep-linking into a level that has not paged that far.

## Test plan

- [x] Fork: full `object_store` test suite passes (75 tests) including a new `list_delimited_page` default-impl test
- [x] `cargo test -p windmill-api shallow_page` — 4 tests over `InMemory` pin one-level listing, marker resume, exact page boundary, and a sibling named like a subtree bound
- [x] `cargo build --features quickjs,enterprise,license,private,parquet`; `npm run check` 0 errors
- [x] Against MinIO (native S3 path): root resolves in 1 request; the 1500-file folder pages to exactly 1500 over 2 pages with MinIO's real base64 continuation token round-tripping; `max_keys=3` pages folders+files together with no duplicates; a 3-file level at `max_keys=3` ends in 1 page with no phantom continuation
- [x] Browser: three-level deep-link expands with correct counts and previews; Load-more cycle `(1000+ items)` → `(1500 items)`
- [ ] Non-admin user on a storage with advanced LFS permission globs → flat-listing fallback, only permitted keys
- [ ] `GitRepoViewer` (always shallow via `rootPath`) browses level by level
- [ ] Azure Blob / GCS smoke test of the native paging path (fork code is exercised by object_store's own client tests; no live Azure/GCS credentials in this environment)

## Fork logistics

The fork is **vendored** at `backend/vendor/object_store` (a workspace member; `VENDORED.md` records the upstream base rev and the Windmill commits on top). No external repo is on the build path — deleting the development mirror at `diegoimbert/arrow-rs-object-store` cannot break a build. The feature commits remain candidates for an upstream PR to `apache/arrow-rs-object-store`; if that lands, the vendor directory goes away and the dependency returns to the upstream pin.

https://claude.ai/code/session_011HeHeGLWDuu62fxjBZA15P


diegoimbert:
Round notes for reviewers, since the last two rounds raced my pushes:

- **`selectItem` expand / `expandAncestors` generation guards** — these were added in `21de5c20cd`, but both Claude and Pi reviewed `54fabefc20` (the commit before it), so the nit is already addressed. Guards are at `S3FilePickerInner.svelte` lines 553 and 849.
- **`expandAncestors` relying on folders-complete-in-the-first-page** — good catch, that guarantee really did go away when the backend stopped using `list_with_delimiter`. Fixed in `28e9fab612`: an ancestor whose level has not been paged far enough to report it is now placed from its path, and dropped again if the load turns up nothing (so a delete that empties a folder does not leave a phantom row). Verified in the browser — a three-level chain expands with correct counts and the leaf previews.
- **Codex has no verdict for the current head**; its run for `21de5c20cd` was not green. Re-triggering below.

diegoimbert:
✅ Review round clean @ 71ba9fdd21683818513eb6326d79f00c6fed4d50 — nit-only verdicts at 28e9fab61237565df5863f53721b3c9ff0697330 (Pi: good to merge; Codex and Claude: mergeable with nits); nits addressed in 71ba9fdd21683818513eb6326d79f00c6fed4d50.

- **Codex P2 — stale ancestor load clears the current loading state.** `loadFiles` now re-checks `loadGeneration` after `expandAncestors` before clearing `fileListLoading`/`fileInfoLoading`, so a reload that started meanwhile owns the spinner.
- **Claude nit — synthesized ancestor below the first level never entered `displayedFileKeys`.** `createShallowNode` only pushes when the parent is the browsed root, so a synthesized mid-chain folder had its children revealed with no row above them. `expandAncestors` now pushes the folder itself as it opens the chain.

The head also moved past the reviewed SHA because `main` was merged into the branch twice during the round (`2acb03b`, `e8a081c`); my commits were rebased on top, not forced. Re-verified after the merges: 4 backend unit tests pass with `--features quickjs,enterprise,license,private,parquet`, `npm run check` reports 0 errors, and in the browser the three-level deep link (`bbb_reports/2026/q1/summary.json`) expands with correct counts and previews, while the 1500-file folder still pages `(1000+ items)` → Load more → `(1500 items)`.

diegoimbert:
Re the review note that the description defended `list_with_delimiter` code that was no longer there — correct, and rather than picking which side to fix, the head now does what the description *should* have promised all along.

`8cedf31a32` (+ companion `e89e9e0`) replaces the client-side delimiting entirely with a small additive fork of `object_store` exposing what S3/GCS/Azure all support natively: **one bounded, resumable, server-side-delimited page per request** (`list_delimited_page`). The EE walk — read-through threshold, subtree jumps, `head()` probes, fabricated resume keys — is deleted (−127/+23); the handler is a pass-through and `next_marker` is the store's own continuation token.

This also resolves the standing perf objections on their own terms rather than trading between them:
- first-page work is bounded by `max_keys` by construction (the drain objection);
- paging a level is `ceil(N/max_keys)` requests (the quadratic objection);
- a Hive-partitioned level is one delimiter request per page of partitions (the client-side-delimiting objection);
- Azure pages server-side via `maxresults`/`marker` — no offset emulation anywhere (the `list_with_offset` fallback objection).

Fork: https://github.com/diegoimbert/arrow-rs-object-store/tree/windmill/list-delimited-page — 2 commits on the previously pinned rev, full object_store suite green (75 tests). Needs mirroring to windmill-labs + repointing before merge; the feature commit is upstreamable. PR description rewritten to match the shipped design.

diegoimbert:
Round findings addressed at `35e88e851e`:

- **Codex P1 — unchecked page size can panic filesystem listings.** Two layers: the fork's default implementation now saturates the slice bound (`start.saturating_add(max_keys)`), and the EE handler clamps the caller-controlled `max_keys` to `[1, 1000]` at the API boundary. Exercised live with `max_keys=18446744073709551615` against a running server — clamped, no panic.
- **Codex P2 — object/prefix name collision dropped at a page boundary.** Real: `InMemory` can hold object `foo` next to prefix `foo/`, and a raw-path token skipped both on resume. The default implementation's token now carries an entry-kind discriminator (`o:`/`p:`), with a regression test paging that exact shape at `max_keys=1`.
- **Claude — lockfile churn.** The wholesale `cargo update` regeneration is reverted; the lockfile is the pre-change one with only the `object_store` source line repointed, validated with `cargo check --locked`. The remaining diff is the object_store entry plus windows-sys bumps pulled by its tree — the `tower`/`socket2`/`base64`/`itertools` downgrades are gone.
- **Codex P1 — personal fork.** Acknowledged and deliberate for now: I don't have rights to create repos under `windmill-labs` (fork attempt returned 403). Tracked in the PR body as a pre-merge requirement — someone with org access needs to mirror `diegoimbert/arrow-rs-object-store@windmill/list-delimited-page` (now at `a75c080`, 3 commits, 76 tests green) and repoint the two `Cargo.toml` lines + lockfile source.

Fork suite: 76 passed. EE `shallow_page` tests: 4 passed. MinIO end-to-end re-verified on the rebuilt binary (root in 1 request, 1500-file folder in 2 pages, native continuation tokens round-tripping).

diegoimbert:
Codex round addressed at `95417535b3aae29385f624261903f58649d73de8`:

- **Stale settle unlocking a current row (P2).** Both `finally` blocks (`loadMoreInFolder`, `selectItem` expand) now only remove their `loadingFolderKeys` entry when the captured generation is still current — a discarded request can no longer delete the fresh tree's guard for the same key. (`clearAndLoadFiles` already wipes the set, so stale keys cannot linger either.)
- **Footer overcount after deleting the last file in a synthesized folder (P2).** Dropping the phantom node now decrements `displayedCount` it had incremented via `createShallowNode`.

The remaining blocker on both Codex's and Claude's verdicts is the org mirror of the object_store fork, which needs someone with `windmill-labs` repo-creation rights — everything else in both reviews is resolved.

diegoimbert:
Both cosmetic P2s from the round at `95417535b3` addressed in `af98815e5c`:

- The wrapper delegation comment no longer claims `LocalFileSystem` pages natively — it forwards to the inner store's drain-and-slice default rather than shadowing it, same as every other delegation in that impl.
- `marker` now carries an OpenAPI description stating it is opaque (the store's own continuation token for shallow listings) and only valid for the same prefix and listing mode.

Every code finding from all seven rounds is now resolved. The single remaining blocker across all three verdicts is organizational: mirroring `diegoimbert/arrow-rs-object-store@windmill/list-delimited-page` (rev `a75c080`) into `windmill-labs` and repointing the two `Cargo.toml` entries + lockfile source line — which needs org repo-creation rights.

diegoimbert:
Latest review addressed at `662cc868c9` + `a3a7033c31`, and the fork-hosting blocker is resolved by **vendoring** rather than an org mirror:

- **Fork availability (the P1 all three verdicts blocked on).** The crate is now vendored at `backend/vendor/object_store` as a workspace member — no external repo is on the build path, so no personal (or org) repo can break a build. `VENDORED.md` records the upstream base rev and the Windmill commits; the development mirror stays only as the upstreaming vehicle.
- **H1 — quadratic filesystem paging.** `LocalFileSystem` now implements `list_delimited_page` natively: names enumerated once per request, sorted, resumed by token, with the per-entry `stat` done only for the returned page. The trait default also builds its page in a single pass (no more sort + HashMap + clone of the level). New crate test pins the native path; `cargo test -p object_store --lib` runs from this workspace now (78 tests green).
- **H2 — ancestor waterfall.** The root level paints as soon as its page lands, and the selected file's ancestor levels — all derivable from the key upfront — load concurrently instead of one await per path segment. Deep-linking `a/b/c/d/file` is now ~2 round-trips of latency with first paint after 1, instead of depth+1 in series.
- **M1 — dead "Load more" under `regexFilter`.** `loadShallowFolder` chases up to 5 pages when a page contributes nothing visible while a marker remains, so filtered folders don't render empty clicks; past the cap the row stays and the user decides.
- **M2 — per-click scans of the node map.** A `childrenByParent` index now backs `revealChildren`, ancestor expansion and `selectItem`'s re-add loop — no more `for..in` over the whole `$state` proxy per expansion.

Verified: 4 EE + 3 vendored-crate `list_delimited_page` tests green, `npm run check` 0 errors, MinIO end-to-end paging unchanged (root 1 request, 1500-file folder 2 pages), and both browser flows (three-level deep link with preview; `(1000+ items)` → Load more → `(1500 items)`) pass on the reworked code.

diegoimbert:
✅ Review round clean @ 2a56643ccd77b479a305f982f4b109ee1535e905 — nit-only verdicts at 662cc868c9336d5383194d91ae434003bb0112f8 (Pi: good to merge; Codex and Claude: mergeable with nits); nits addressed in 2a56643ccd77b479a305f982f4b109ee1535e905:

- **Orphaned synthesized descendants / phantom chain (Codex P2, Claude nit).** When any level of the chain proves stale, every synthesized node from that point down is now dropped — from `allFilesByKey`, the child index, the visible rows, and the loaded-items footer — not just the first one.
- **Duplicate background ancestor loads (Codex P2).** `expandAncestors` registers each in-flight level in `loadingFolderKeys`, so the row is locked (and shows the spinner) while its page loads; a click can no longer start the same first-page request twice and double the folder count. Loads already in flight from a click are skipped symmetrically.
- **Stale wrapper comment (Codex P2).** Now states only what is true of the code as it stands: the delegation forwards to the inner store's implementation instead of shadowing it.
- **`VENDORED.md` commit count (Claude nit).** No longer states a count that can rot.

PR #10384, opened 2026-07-28
fix: home kind filter no longer resets a fork or reloads the page

rubenfiszel:
## Summary

Two independent bugs on the home page's Script / Flow / App filter, found from one report ("selecting a fork then clicking a filter resets to root").

**1. The filter reverted the workspace to the root (and dropped the search text).**

`setQuery` (`frontend/src/lib/navigation.ts`) rebuilt the whole query string from a URL its caller passed in — and the home page passed `page.url`. SvelteKit refreshes `page.url` only on router-driven navigations: `replaceState`/`pushState` write the address bar and `page.state`, never `page.url`. Two params on that page are written exactly that way:

- `?workspace=<fork>` — `fixupUrlAfterWorkspaceSwitch` rewrites it with SvelteKit's shallow `replaceState` after a workspace switch.
- `?search=<text>` — `useSearchParams` writes it with raw `history.replaceState`.

So `page.url` still held the pre-switch `?workspace=<root>` and no `search` at all. Clicking a kind tab called `goto('?' + <stale params> + '&kind=script')`, landing on `?workspace=<root>&kind=script`; the logged layout's `onQueryChange` then read that param back and set the workspace store to the root.

`setQuery` now always sources the query from `window.location.search` — the address bar is the one URL that every writer, router-driven or shallow, keeps current. The `url` parameter is removed outright so a stale snapshot can't be handed in again.

**2. Every navigation was a full page reload on Cloudflare Pages deployments.**

Pre-existing, and reproducible on the `main` preview too, but it made the filter look broken. `frontend/static/_headers` applies COOP/COEP to `/*`, so `window.crossOriginIsolated` is true on *every* page of a Pages deployment. That tripped this branch in `(logged)/+layout.svelte`:

```js
} else if (toPath && (isEditorPath(currentPath) || window.crossOriginIsolated)) {
    navigation.cancel()
    window.location.href = navigation.to!.url.href
}
```

The branch exists to force one reload when *leaving* the cross-origin-isolated raw app editor, so its COEP header doesn't stick for the rest of the SPA session. With site-wide isolation that condition is true forever, and the reload it forces can't clear an isolation the next document asserts too — so every navigation became a page load for no benefit.

The path check alone is sufficient here. Of the four paths `needs_cross_origin_isolation` (`static_assets.rs:71`) matches, only `/apps_raw/edit` and `/apps_raw/add` are routes this layout governs: `/ui_builder/` has no SvelteKit route (iframe-only static bundle), and `/public/` and `/a/` sit outside `(logged)`. Entering the editor always forces a full load, so while the document is isolated its path *is* the editor path.

## Changes

- `setQuery(url, key, value, hash)` → `setQuery(key, value, hash)`; the query comes from `window.location.search`. It also no longer mutates the caller's URL object (it was mutating SvelteKit's `page.url` in place).
- Updated all 24 call sites: 11 trigger list pages, the home page's hub-tab kind toggle, and `ItemsList`. The trigger pages already passed `new URL(window.location.href)`, so they are unchanged in behavior; only the two `page.url` callers change.
- `ListFilters.svelte` untouched: it declares its own `setQuery` that shadows this one, writes via `pushState` rather than `goto`, and already reads the live URL.
- `(logged)/+layout.svelte`: the leaving-the-editor reload guard keys off the editor path only, not `window.crossOriginIsolated`.
- Regression test in `navigation.test.ts` pinning that params present only in the address bar survive into the `goto` target.

## Test plan

- [x] `npm run check` — 1 error, pre-existing and unrelated (`modern-screenshot` missing in this worktree)
- [x] `npx vitest run src/lib/navigation.test.ts --project server` — 5 passed
- [x] Full `--project server` suite: 4 failing files (copilot chat `ContextManager` / `app/core`, `raw_apps/utils`, dbtable `alterTable`) all pre-existing and unrelated — none reference `navigation`/`setQuery`
- [x] Local dev: with a fork selected and `foo` in the search box, clicking Scripts keeps `?workspace=wm-fork-…&search=foo&kind=script` and the sidebar stays on the fork. Reproduced the original failure first (it navigated to `?workspace=test&kind=script`, dropping both).
- [x] Local dev: `/routes` trigger page — the two chained `setQuery` calls still land both params
- [x] Cloudflare preview, filter click: no document reload, only the 2 expected `runnables/*` refetches
- [x] Cloudflare preview, plain sidebar nav (Runs → Home): no document reload
- [x] Cloudflare preview, **into** the raw app editor: still a hard document load, `crossOriginIsolated === true`, `new SharedArrayBuffer(8)` succeeds, tsserver workers running
- [x] Cloudflare preview, **out of** the editor: still a hard document load
- [x] Confirmed the reload is pre-existing: the same filter click on the `main` preview (windmill.pages.dev) also hard-reloads

## Screenshots

Home page in the fork, with the search text and the Scripts filter both applied — sidebar and banner still on `MyFork`:

home: fork + search + kind filter all preserved

## Follow-up (not in this PR)

- `frontend/static/_headers` isolates `/*` where the Rust server isolates only the routes that need it. Scoping it would also unbreak CORP-less cross-origin subresources on Pages deployments — but Cloudflare `_headers` cannot match on query string, so it can't express the `?wm_coep` opt-in for `/public/` and `/a/`. Left alone deliberately.
- Three other sites build a navigation target from `page.url` and share the staleness class: `runs/TimeframeSelect.svelte:16` (service logs; writes without navigating, so it corrupts the address bar rather than the store), `workspaceSettings/WorkspaceIntegrations.svelte:269`, `pipeline/[folder]/+page.svelte:172`.

### DIFF
--- a/frontend/src/lib/components/home/ItemsList.svelte
+++ b/frontend/src/lib/components/home/ItemsList.svelte
@@ -1361,7 +1361,7 @@
 					if (itemKind != 'all') {
 						subtab = v
 					}
-					setQuery(page.url, 'kind', v)
+					setQuery('kind', v)
 				}}
 			>
 				{#snippet children({ item })}
--- a/frontend/src/lib/navigation.test.ts
+++ b/frontend/src/lib/navigation.test.ts
@@ -1,5 +1,10 @@
-import { describe, it, expect } from 'vitest'
-import { buildFilterUrl } from './navigation'
+import { describe, it, expect, vi } from 'vitest'
+
+const goto = vi.hoisted(() => vi.fn(async (_path: string) => {}))
+vi.mock('$app/navigation', () => ({ goto: (path: string) => goto(path) }))
+vi.mock('$app/paths', () => ({ base: '' }))
+
+import { buildFilterUrl, setQuery } from './navigation'
 
 // Parse the (un-based) app path the builder returns against a dummy origin so we can
 // assert on pathname / searchParams / hash without caring about ordering.
@@ -48,5 +53,29 @@ describe('buildFilterUrl', () => {
 		const u = parse(buildFilterUrl('/schedules', { path: 'f/x' }, { hash: 'f/a/b' }))
 		expect(u.searchParams.get('path')).toBe('f/x')
 		expect(u.hash).toBe('#f/a/b')
+	})
+})
+
+describe('setQuery', () => {
+	// setQuery reads nothing but `window.location.search`; the node test env has no
+	// real `location`, so a stub carrying the query is the whole browser it needs.
+	function addressBar(search: string) {
+		Object.defineProperty(globalThis, 'location', {
+			value: { search },
+			writable: true,
+			configurable: true
+		})
+	}
+
+	it('preserves params that only the address bar knows about', async () => {
+		goto.mockClear()
+		addressBar('?workspace=wm-fork-x&search=foo')
+
+		await setQuery('kind', 'script')
+
+		const u = parse(goto.mock.calls[0][0])
+		expect(u.searchParams.get('workspace')).toBe('wm-fork-x')
+		expect(u.searchParams.get('search')).toBe('foo')
+		expect(u.searchParams.get('kind')).toBe('script')
 	})
 })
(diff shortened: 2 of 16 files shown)


rubenfiszel:
✅ Review round clean @ a63da9e4a54c02590de8489a7a8fa5b3933e28b9

PR #10387, opened 2026-07-28
feat(apps): support the frontend SDK in sandboxed raw apps

diegoimbert:
Stacked on #10377 (base branch `raw-apps-wmill-client-frontend`).

## Summary

#10377 shipped the viewer-permissioned `windmill-client` SDK for **unsandboxed** raw apps only. This lifts that restriction, so an app can have both sandbox isolation and frontend API access.

The blocker was never token delivery — it was that the published SDK forces `credentials: 'include'` on every generated-service call. Windmill's CORS answers `Access-Control-Allow-Origin: *` and never sets `allow_credentials`, so from a sandboxed app's opaque origin the browser rejects the request *before the token is consulted*. An opaque origin has no cookies to send anyway, so credentials buy nothing there.

Two things turned out **not** to be blockers, contrary to the follow-up note in #10377: the job-run routes already carry their own CORS layer (`jobs.rs:127`), as do `jobs_u/*`, `/resources` and `/users`. Only `/variables` was missing one.

## Changes

**`typescript-client`** — `WITH_CREDENTIALS` now mirrors one rule in both places: cookies only when there is no bearer token. `setClient()` sets it from the resolved token instead of hardcoding `true`, and the `build.sh` sed emits `!getEnv("WM_TOKEN")` instead of `true`, which is the path raw apps hit (they never call `setClient`). Also fixes `getEnv`'s bare `process?.env`, which throws `ReferenceError` where `process` is undeclared (web workers) — `build.sh` already injected the guarded form into `OpenAPI.ts`, the two copies had drifted.

**Backend** — `build_embed_token_response` no longer suppresses `sdk_scopes` for sandboxed apps; the sandboxed wrapper sets `window.process.env` from the existing `windmill:ctx` handshake, before it injects the bundle script (which is what makes the SDK's module-load auto-configuration work); `/variables` gets the CORS layer its sibling routers already have.

**Frontend** — `RawAppPreview.respondCtx` includes the `sdk` payload; `BASE_URL` is passed explicitly because `location.origin` is `"null"` on an opaque origin. The editor's "not available with sandbox isolation" warning becomes a note about the required client version.

## Trust model

Unchanged. The scopes, the consent prompt, the `raw_app_sdk` sentinel and the mint path are all mode-independent. Note the prompt matters *more* here: an unsandboxed app already runs with the viewer's whole session, so the token is strictly less than it had; a sandboxed app has nothing else, so this is its only reach.

## Known limitation

Bundles pin their SDK copy at deploy time, so **an existing sandboxed app must be redeployed** after the release carrying the client fix. Until then it fails with a CORS error — hence the editor note naming the required version (sourced from `__pkg__.version`, which tracks the monorepo release the client publishes from).

## Screenshots

Permission prompt on a sandboxed app:

prompt

After Continue — SDK calls from inside the opaque iframe:

app

## Test plan

- [x] Built the client locally and confirmed the emitted config: `WITH_CREDENTIALS: !getEnv("WM_TOKEN")`; at runtime `false` with a token, `true` without, and `setClient` follows the same rule
- [x] Sandboxed app now advertises `sdk_scopes` and mints on `?sdk_consent=true`
- [x] **A/B proving the client fix is load-bearing**: same app, same token — bundled against the published client → `SDK FAILED: Failed to fetch`; against the fixed client → `whoami` returns the viewer and `getVariableValue` returns the value
- [x] Frame is genuinely opaque: the wrapper's storage shim engaged (it only installs when real storage throws) and `window.parent.location` raises `SecurityError`; requests carry the bearer and **no cookie**
- [x] `/variables` CORS is what enables the variable read — control: `/scripts` (no layer) returns no `Access-Control-Allow-Origin`, `/variables` and `/resources` do
- [x] Regression: unsandboxed app still works unchanged
- [x] `cargo check --features quickjs`, `npm run check:fast`, `tsc` on the generated client

### DIFF
--- a/backend/windmill-api/src/apps.rs
+++ b/backend/windmill-api/src/apps.rs
@@ -639,6 +639,9 @@ async fn get_raw_app_data(
             .header(http::header::CONTENT_TYPE, "text/html; charset=utf-8")
             .header("X-Content-Type-Options", "nosniff")
             .header("Cross-Origin-Resource-Policy", "cross-origin")
+            // This URL carries the frontend-SDK handshake nonce, so it must not
+            // travel to whatever the bundle navigates to or links out to.
+            .header("Referrer-Policy", "no-referrer")
             .header(
                 http::header::CONTENT_SECURITY_POLICY,
                 "sandbox allow-scripts allow-forms allow-popups \
@@ -843,7 +846,13 @@ fn raw_app_wrapper_html(secret: &str) -> String {
       loadBundle();
     }
   });
-  try { window.parent.postMessage({ type: 'windmill:ready' }, '*'); } catch (_) {}
+  // Echo the handshake nonce from our own URL. It proves to the embedder that
+  // this is the document it loaded rather than one navigated into the frame
+  // afterwards, which is what gates the frontend SDK credential.
+  try {
+    var hs = new URLSearchParams(window.location.search).get('wm_hs') || undefined;
+    window.parent.postMessage({ type: 'windmill:ready', nonce: hs }, '*');
+  } catch (_) {}
   // Fallback for contexts that never send ctx (e.g. ctx-less rendering).
   setTimeout(loadBundle, 1500);
 })();
--- a/backend/windmill-api/src/apps.rs
+++ b/backend/windmill-api/src/apps.rs
@@ -827,30 +827,26 @@ fn raw_app_wrapper_html(secret: &str) -> String {
   }
   function applyCtx(d) {
     if (!d || d.type !== 'windmill:ctx') return;
-    {
-      window.ctx = d.ctx;
-      if (window.__wmStorageShim && d.storage) {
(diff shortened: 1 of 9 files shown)


diegoimbert:
✅ Review round clean @ 00121914373ffb639c8248d1e076a8ffe1594f40 — nit-only verdicts at 3dd117fff2ffd83ae187bd619658a67a67555ead (Codex: nits; Claude: nits; Pi: good to merge); all nits addressed in 00121914373ffb639c8248d1e076a8ffe1594f40: dropped the now-dead `contentWindow` fallback so a sender without a port gets nothing at all (previously it still received the viewer's identity), corrected the two comments that called the SDK token a sandboxed app's *only reach* (its policy-approved runnables still run through the bridge — it is the only **direct API** credential), and condensed the wrapper handshake comment plus a leftover brace.

PR #10388, opened 2026-07-28
feat: validate and make explicit the hub project's resource type export

rubenfiszel:
## Summary

Publishing a folder to the Hub derived its resource types from every `resource-<x>` arg format found in the selected items' schemas, and pushed all of them automatically. Two problems:

- `resource-<x>` is not a resource-type declaration. `resource-s3_object` (an S3 file input, not a resource type) and typo'd or stale formats were published as Hub resource types with an empty schema, plus a `f/<slug>/<x>` stub resource nothing can ever fill.
- Publishing a resource type's definition to the Hub happened as a silent side effect of bundling, with no way to opt out.

Now an input-derived type must be a resource type the workspace actually declares — the same bar `ArgInput` applies before rendering an arg as a resource picker — and publishing a type's *definition* is an explicit per-type opt-in, off by default.

## Changes

- `inputResourceTypes()` (moved from `deployToHubSession.svelte.ts` into the pure `deployToHubItems.ts` so it is testable) filters `resource-<x>` arg formats against the workspace's resource-type names, loaded once per session via `listResourceTypeNames`. Unknown formats produce neither a type push nor a stub resource. Types derived from a `$res:` reference are untouched — the resource itself declares its type.
- The catalog is only authoritative once non-empty: while it is still loading, or on a workspace that never synced the Hub's types, nothing is filtered rather than everything being dropped.
- New per-type **Export type definition** toggle in the Resource dependencies drawer, **off by default**. Only ticked types are pushed to `/hub/resource_types`.
- Stub resources are still synced for every dependency, so a fork still knows which credentials to fill and a hardcoded `$res:f/<slug>/…` in published code still resolves; only the type definition stays out of the Hub.
- Predeploy summary shows an `exported` badge per opted-in type and the count on the drawer button.

## Screenshots

Predeploy summary — nothing exported by default. The script under test takes four `resource-*` args (`postgresql`, `my_custom_api`, `s3_object`, `not_a_real_type`); only the two real resource types are listed:

predeploy-default-none-exported

Resource dependencies drawer — both toggles off by default:

drawer-toggles-off

After ticking `my_custom_api` only — badge on the chip, count on the button, `postgresql` left alone:

rt-summary2

## Test plan

- [x] `vitest run deployToHubItems.test.ts` — covers declared types, unknown formats, hidden formats, and the no-catalog fallback
- [x] Publish drawer on a folder whose script takes `resource-postgresql`, `resource-my_custom_api`, `resource-s3_object` and `resource-not_a_real_type`: only the two declared types are listed
- [x] Both toggles off on a fresh session; ticking one shows the `exported` badge and updates the drawer button count

---
## Summary by cubic
Make Hub resource type exports explicit and safe. Only workspace-declared types are detected from inputs, exporting a type’s definition is opt-in and captured at publish click; stubs still sync for all dependencies.

- **New Features**
  - Added “Export type definition” toggle per resource type in the Resource dependencies drawer (off by default).
  - Predeploy summary shows an “exported” badge per opted-in type and a count on the details button.
  - Only selected types are published; stub resources still sync for every dependency so forks know which credentials to fill and hardcoded $res paths resolve.

- **Bug Fixes**
  - Validate input-derived types against the workspace’s resource-type names; ignore unknown or hidden formats (`app_theme`, `state`, `cache`, `s3_object`), with `s3_object` excluded even when no catalog is available.
  - Skip validation when the type catalog is empty or still loading to avoid dropping legitimate types.
  - Snapshot the selected exports at publish click, before the draft request starts, so mid-deploy changes don’t affect the published set.

<sup>Written for commit dc59f8e781d41ae34ca3fb1cae045adaef1f7bb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/workspaceSettings/DeployToHub.svelte
+++ b/frontend/src/lib/components/workspaceSettings/DeployToHub.svelte
@@ -287,8 +287,9 @@
 									{/if}
 									<Tooltip>
 										Resource types the selected items depend on (whether passed as inputs or
-										referenced by a hardcoded path). Synced to the Hub so a fork knows what
-										credentials it needs to fill.
+										referenced by a hardcoded path). A stub resource of each type is synced to the
+										Hub so a fork knows what credentials it needs to fill. Publishing a type's own
+										definition is opt-in — tick it in the details drawer.
 									</Tooltip>
 								</span>
 								{#if s.dependencyTypes.length === 0}
@@ -303,6 +304,9 @@
 												: 'bg-surface'}"
 										>
 											{r.resource_type}
+											{#if s.exportedResourceTypes.has(r.resource_type)}
+												<Badge color="blue" size="xs">exported</Badge>
+											{/if}
 										</span>
 									{/each}
 									<Button
@@ -311,7 +315,9 @@
 										wrapperClasses="ml-auto"
 										onclick={() => resourceDrawer?.openDrawer()}
 									>
-										View details
+										{s.exportedDependencyTypes.length > 0
+											? `View details (${s.exportedDependencyTypes.length} type definition${s.exportedDependencyTypes.length > 1 ? 's' : ''} exported)`
+											: 'View details'}
 									</Button>
 								{/if}
 							</div>
@@ -830,11 +836,18 @@
 			<DrawerContent title="Resource dependencies" on:close={() => resourceDrawer?.closeDrawer()}>
 				<div class="flex flex-col gap-4">
 					<p class="text-xs text-secondary">
-						Resource types the selected items depend on. Each is synced to the Hub so a fork knows
-						what credentials it needs to fill. <span class="font-semibold">Input</span> means the
-						item takes the resource as a parameter;
+						Resource types the selected items depend on. A stub resource of each type is synced to
+						the Hub so a fork knows what credentials it needs to fill.
+						<span class="font-semibold">Input</span> means the item takes the resource as a
+						parameter;
 						<span class="font-semibold">hardcoded path</span> means the item pins a specific resource
 						path in its code.
+					</p>
+					<p class="text-xs text-secondary">
+						Publishing a type's own <span class="font-semibold">definition</span> (its schema and
+						description) is a separate, explicit choice: tick
+						<span class="font-semibold">Export type definition</span> only for custom types the Hub doesn't
+						already know. Standard types are already defined on the Hub and need no export.
 					</p>
 					{#if s.dependencyTypes.length === 0}
 						<span class="text-xs text-hint">No resource references in the current selection.</span>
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Round 1 dispositions (head `b9b7f5a8fc`):

**Fixed**
- **[codex P1] snapshot the opt-in before the first await** — `exportedResourceTypes` is now snapshotted at the top of `#deployAll` alongside the items/triggers/migrations snapshots, so toggling during an in-flight publish can no longer change what gets pushed.
- **[claude P2] `s3_object` slipping through the no-catalog path** — moved into `HIDDEN_RESOURCE_TYPES`, so it is excluded unconditionally rather than only when a catalog happens to be loaded. Test expectation updated.
- **[claude P2] silent `catch {}` on the catalog load** — now logs, and carries a comment on why `resourceTypesStore.getResourceTypes()` is deliberately not reused: its error path resolves to a non-empty `['error_fetching_names']`, which here would read as a real catalog and filter out every legitimate type.

**Not changing**
- **[claude P2] stub pushed for a type whose definition never reaches the Hub** — confirmed with the Hub owner: `/hub/resources` accepts a stub whose `resource_type` the project has not published, so leaving a type unticked does not fail the publish. This is the intended split: the stub tells a fork which credential to fill, the definition is the part that is opt-in.
- **[claude P2] opt-in not rehydrated on re-deploy** — the Hub project payload (`GET /hub/project`) returns items, not the project's resource types, so there is nothing to rehydrate from without a Hub-side change. Re-deploying with the toggle off does not remove an already-published type; the residual gap is that a *schema change* to an already-exported type needs a re-tick to propagate, which follows from making the export explicit per publish.

**[codex] manual verification of the race** — covered by construction rather than by a slow-publish repro: the snapshot is taken before any `await` in `#deployAll`, so no later mutation of `exportedResourceTypes` is observable by that run.

rubenfiszel:
Round 2 (head `7cafc83a28`):

**[codex P1] export selection snapshotted after the draft request** — fixed. `publishBundle()` now captures the set before `#createDraft()` and passes it into `#deployAll()`, so the window between clicking publish and the draft response is closed. The toggles are also `disabled` while `deploying`, so the UI no longer suggests a change that wouldn't apply.

Worth noting for a follow-up: `itemsSnapshot` / `triggersSnapshot` / `migrationsSnapshot` are still taken inside `#deployAll`, i.e. after the same await, so the item checkboxes and migration drafts keep the identical (pre-existing) window. Left alone here to keep this PR scoped to the resource-type export.

rubenfiszel:
✅ Review round clean @ dc59f8e781d41ae34ca3fb1cae045adaef1f7bb9 — nit-only verdicts at 7cafc83a2844948ef3128178a4e2a4f8591c64a3 (codex P2 + claude P2, both the same stale `publishBundle` comment); addressed in dc59f8e781d41ae34ca3fb1cae045adaef1f7bb9

PR #10390, opened 2026-07-28
fix: truncate strings on char boundaries to avoid panics on multibyte input

rubenfiszel:
## Summary

Starting a chat-enabled flow returned HTTP 500 when the **first** message of a new conversation contained multibyte UTF-8 (Cyrillic, CJK, emoji) and byte 25 fell inside a character: `get_or_create_conversation_with_id` truncated the conversation title with `&title[..25]`, which slices by **byte** index and panics on a non-boundary. The title is the raw user message, so any caller could panic the request handler.

Fixes GIT-936

```
thread 'tokio-runtime-worker' panicked at windmill-common/src/flow_conversations.rs:56:32:
byte index 25 is not a char boundary; it is inside 'а' (bytes 24..26) of `Нам нужно оснастить конференц-зал ...`
```

The same byte-slicing pattern existed at several more sites on user-controlled strings, fixed here too.

## Changes

- Add `windmill_common::utils::truncate_chars(s, max_chars) -> &str` (borrowed prefix) and `truncate_with_ellipsis(s, max_chars) -> String` — char-boundary-safe truncation, mirroring the existing `truncate_for_log` pattern in `sql_s3_input.rs`. Unit test covers the multibyte-at-boundary case.
- `flow_conversations::get_or_create_conversation_with_id`: truncate the conversation title by characters (25) instead of bytes.
- `instance_config::format_setting_value`: same fix for the 197-char setting-value preview.
- `windmill-api-debug::sign_expression`: same fix for the 200-char audit `resource` built from a user-supplied expression.
- **EE companion** (windmill-labs/windmill-ee-private#690): `audit_log` byte-sliced `username`, `email`, `span` and `resource` at 255 — all user-controlled, all a live panic on multibyte input longer than 255 bytes. Now counted and sliced by characters, which also matches the `varchar(255)` columns (Postgres `varchar(n)` counts characters, not bytes).

## Test plan

- [x] `cargo test -p windmill-common --lib utils::tests` and the `instance_config` tests pass
- [x] `cargo check -p windmill-audit --features private,enterprise,windmill-common/private`
- [x] End-to-end against a running backend, chat-enabled flow, `POST /api/w/test/jobs/run/f/<path>?memory_id=<uuid>`:
  - `"ааааааааааааа"` (13x U+0430 = 26 bytes, byte 25 mid-char) → **201** (was 500 + panic)
  - `"Нам нужно оснастить конференц-зал новым оборудованием"` → 201, title stored as `Нам нужно оснастить конфе...` (28 chars)
  - 13x `🙂` (4-byte chars) → 201
  - short ASCII / 12-char Cyrillic / existing-conversation paths → 201, unchanged
- [x] No panics in the server log for any of the above


rubenfiszel:
Nit round addressed:

- **`[ee]` title prefix** (codex P2) — already applied before this round was triggered; the title has read `[ee] fix: truncate strings on char boundaries…` since the EE ref was pinned. No change needed.
- **`truncate_with_ellipsis` length contract** (codex P2 / cubic P3) — fixed in a3ebe16: the doc comment now states that a truncated result is `max_chars + 3` characters, not `max_chars`.
- **"confirm all four `audit_ee.rs` fields were converted"** (claude) — confirmed. windmill-labs/windmill-ee-private#690 converts all four, and each is now **character**-counted (not byte-counted), matching the `varchar(255)` columns:

  ```rust
  // These map to varchar(255) columns, which count characters. Slicing by byte index
  // would panic when the cut lands inside a multibyte character.
  let mut username = author.username();
  if username.chars().count() > 255 { …; username = truncate_chars(username, 255); }

  let mut email = author.email();
  if email.chars().count() > 255 { …; email = truncate_chars(email, 255); }

  let mut span = author.token_prefix();
  if span.is_some_and(|s| s.chars().count() > 255) { …; span = span.map(|s| truncate_chars(s, 255)); }

  let mut resource = _resource;
  if resource.is_some_and(|r| r.chars().count() > 255) { …; resource = resource.map(|r| truncate_chars(r, 255)); }
  ```

  So the widening at `windmill-api-debug/src/lib.rs` (200 chars, up to 800 bytes) is safe: `resource` is bounded by characters against a character-counted column. Compiled with `cargo check -p windmill-audit --features private,enterprise,windmill-common/private`.
- **`truncate_for_log` in `sql_s3_input.rs`** (claude, explicitly "not a finding") — left alone. It uses a different suffix (`…`) and the suggested one-liner would append unconditionally; folding it in is a separate cleanup.

rubenfiszel:
✅ Review round clean @ a3ebe16010d943c79df1734fead539d6311007a4 — nit-only verdicts at 1ea9685015653c444a0b01efa28ed812b453223e; doc-contract nit addressed in a3ebe16010d943c79df1734fead539d6311007a4, the `[ee]`-prefix nit dismissed (already applied) and the EE-confirmation nit answered in review replies

PR #10398, opened 2026-07-28
fix: surface the real reason git sync settings saves are rejected

rubenfiszel:
## Summary

Enabling "Promote to prod via Git" on a dev workspace could fail with nothing but `ApiError: Bad Request` in the toast. The backend does return a precise reason (plain text, 400), but the generated API client throws `ApiError` with a canned per-status message (`400 → 'Bad Request'`) and keeps the server's explanation in `.body`, which the git sync toasts were dropping.

Two fixes: the git sync settings UI now shows the server's reason, and the dev-promotion parent-repo check names the actual mismatch instead of a generic sentence.

## Changes

- `frontend/src/lib/utils.ts`: add `apiErrorMessage(e)`, which prefers `.body` (the server's text) over `.message` (the canned status text). `workspaceSettings/projectInstall.ts` had a private copy of this and now imports it.
- `frontend/src/lib/components/git_sync/*`: the promotion toggles plus the sibling save / detect / remove / apply-settings toasts use it, so every git sync failure surfaces its reason. `PullWorkspaceModal`'s auto-save toast previously showed no reason at all.
- `windmill-common/src/git_sync_ee.rs` (EE): `dev_promotion_target_mismatch` returns *why* a dev workspace's promotion target does not match its parent's. `dev_promotion_target_matches_parent` becomes a thin wrapper, so the two deploy-path callers are unchanged. Only the normalized `host[:port]/path` key is exposed, never the raw resource URL (it can embed a token), and only the parent's branches **for that same repository** are listed, so the message stays actionable without enumerating repositories the caller may not otherwise see.
- `backend/windmill-api-workspaces/src/workspaces.rs`: `check_dev_promotion_targets_parent_repo` builds the message from that.
- Credential safety: `normalize_git_repo`'s key falls back to the raw URL when `url::Url::parse` fails (a `/` inside a token is enough), so the rendered repo gets its userinfo stripped first. The existing `sanitize_git_url` is not enough here: its `://[^/@]+@` regex cannot span the very `/` that made the parse fail.

Before / after for the most common case, a dev workspace whose git resource tracks a different branch than its parent:

> **before:** Promotion mode on a dev workspace must reuse the parent workspace's git repository (same URL and branch), but this repository is not one the parent tracks.
>
> **after:** Promotion mode on a dev workspace must reuse the parent workspace's git repository (same URL and branch). Repository 'u/admin/gh' targets 'github.com/org/proj' on branch 'staging', but the parent workspace 'proj' tracks it on branch 'master'. Set this repository's branch to match, or add branch 'staging' to the parent's git sync settings.

## Screenshots

Toast on a real failure (CE build rejecting promotion mode), before and after:

toast-before

toast-after

## Test plan

- [x] `cargo check --features enterprise,private` (windmill-common + windmill-api-workspaces)
- [x] `npm run check:fast`
- [x] `POST /api/w/admins/workspaces/edit_git_sync_repository` with `use_individual_branch: true` on a CE build returns 400 `text/plain` with `Bad request: Promotion mode is an Enterprise Edition feature`
- [x] Same call through the app's generated client in the browser: `${e}` yields `ApiError: Bad Request`, `apiErrorMessage(e)` yields the server's message (screenshots above)
- [x] `cargo test -p windmill-common --lib git_sync` (includes the new redaction test)
- [x] `vitest run src/lib/utils.test.ts` (4 new `apiErrorMessage` cases)
- [ ] On an EE instance with a dev workspace whose git resource branch differs from its parent's, the promotion toggle names both branches

---
## Summary by cubic
Shows the real server error in Git sync toasts and modals, and makes EE dev‑promotion validation explain the exact repo/branch mismatch while redacting credentials.

- **Bug Fixes**
  - Frontend: added a shared error helper and used it across detect/save/remove/apply flows, push/pull/autosave toasts, test job runs, and workspace settings (project install/migrations) so API errors show the server’s message instead of “Bad Request”.
  - Backend: promotion rejections now report the specific mismatch from `dev_promotion_target_mismatch`, naming the repo key and parent branches only for the same repository; if the parent doesn’t track it, the hint focuses on pointing at the parent’s repo. Drops the “add branch” suggestion when the dev repo has no branch. Raw URLs and unrelated repos are never shown, and credentials are stripped if URL parsing fails.

- **Dependencies**
  - Updated EE reference in `backend/ee-repo-ref.txt` to `85209cfccb07538ae4748d85b56e61c0bef4f606`.

<sup>Written for commit 3e47e4d4cb61d1cc5127b2f1efd6698e4ea40f62. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ c0b85e4eb6516ebfc7cd96eefa26d69d7f958214 — nit-only verdicts at c0b85e4eb6516ebfc7cd96eefa26d69d7f958214; nits addressed in 3e47e4d4cb (unactionable "add branch '<none>'" advice, last inline error-body copy in GitSyncContext) and in the PR title (`[ee]` prefix restored)

PR #10403, opened 2026-07-29
fix(forks): record fork changes that never reached the diff tally

### LINKED ISSUE #10401: bug:Changes to windmill scripts from .github/workflows actions (to sync forked workspaces) don't register as changes and hence can't be pushed upstream
### Describe the bug

We have a setup where we use a .github/workflows action to enable autosyncing of changes for any forked workspaces of our production workspace (but not for the production workspace/branch itself).

This makes working localy on dev branches easier (with easy testing in windmill) but protects the production workspace.
_(Unfortunatly we can't use the new github app sync functionality as our windmill is only available on a private network and can hence only be reached by first opening a tailscale connection on a github runner.)_

However any commits pushed into windmill workspace forks by a .github/workflows action don't appear as a change that can be pushed upstream to the parent (our production branch) inside Windmill. The deploy to {parent workspace} list is empty even though the fork has new scripts and changes to existing scripts. Only edits made in windmill to scripts can be deployed, any changes pushed by github actions into the workspace don't appear.

The current "fix" is to manually merge the github branch of the forked workspace into the github branch of the production workspace and then use the pull from repo button in the settings menu of our production branch to update production branch to have the changes. This is not a long term workable solution.

### To reproduce

1. Fork a windmill workspace and enable auto syncing from github using the .github/workflows action
2. Make changes to a script (or make a new script) in github
3. The script/change automaticaly appears in the forked branch
4. The script/change does not show in the "Deploy to {parent workspace name}" section of the "Compare & Deploy" menu.

### Expected behavior

I would expect new/changed scripts from the github action to show up in the ["Deploy to {parent workspace name}" section of the "Compare & Deploy" menu] just like making (a change to) a script in the windmill web UI makes that script show up in said list.

### Screenshots

Can provide if needed

### Browser information

Chrome

### Application version

EE v1.775.1

### Additional Context

Let me know if anything is unclear

rubenfiszel:
## Summary

Fixes #10401. Fixes WIN-2259.

A fork's **Compare & Deploy** list is not a live diff of the two workspaces — it reads
`workspace_diff`, a table maintained incrementally by `tally_deployed_object_changes` on
every deploy. `compare_workspaces` only ever re-evaluates rows that are already there;
nothing rescans. **A tally missed once makes the change invisible forever**, even though
the item is deployed and live in the fork.

This PR closes the two holes where that happens.

### 1. `ahead` was keyed off `deploy_to`, not the parent link (the reported bug)

The two directions of the tally used different sources of truth:

| direction | keyed off | when `workspace_settings.deploy_to` is NULL |
|---|---|---|
| parent → fork (`behind`) | `workspace.parent_workspace_id` | works |
| fork → parent (`ahead`) | `workspace_settings.deploy_to` | **records nothing at all** |

`parent_workspace_id` is the source of truth for fork-ness; `deploy_to` is the
user-editable prod/dev pairing setting, which an admin can clear from workspace settings
and which older forks may never have had set. A fork in that state has a permanently
empty "Deploy to &lt;parent&gt;" list while "Update current" keeps working — the exact
asymmetry reported in #10401, where changes pushed into a fork by a GitHub Action never
became deployable upstream.

The upstream side now tallies against both `deploy_to` and `parent_workspace_id`,
deduplicated, so a `deploy_to` pointing somewhere other than the parent keeps its own
diff and the counter is not double-incremented for the common case where they are equal.

### 2. A deploy whose lock job doesn't finish was never tallied

The new script/flow/app version is committed *before* the dependency (lock-generation)
job is pushed, but `handle_deployment_metadata` — and therefore the tally — only runs on
that job's success path. A failed, cancelled, or timed-out lock job leaves the item live
in the fork and untallied. This bites CI pushes hardest: a script authored directly in
git has no `.script.lock`, so it always goes through server-side lock generation, and
nobody watching a GitHub Action checks the Runs page.

The failure and cancellation paths now record the tally (the tally only — a broken lock
should not be pushed to the git repo).

## Changes

- `windmill-git-sync/src/git_sync_ee.rs` (EE companion PR): tally the upstream direction
  against `COALESCE`-style union of `deploy_to` and `parent_workspace_id`, deduplicated
- `windmill-worker/src/worker.rs`: after the dependency-job dispatch, tally when the job
  errored or was cancelled
- `windmill-worker/src/worker_lockfiles.rs`: new `tally_unfinished_dependency_deploy`,
  which rebuilds the `DeployedObject` from the job (reading `app_version.raw_app` to pick
  `app` vs `raw_app`) and skips jobs with no `runnable_id` (one-off preview lock jobs).
  Also called at the flow and app handlers' own cancellation early-returns, which return
  `Ok` with the in-worker `canceled_by` flag unset
- `windmill-git-sync/src/lib.rs` + `git_sync_oss.rs`: re-export `tally_deployed_object_changes`
  with a CE no-op stub

## Companion PR

windmill-labs/windmill-ee-private#693 — `backend/ee-repo-ref.txt` bumped to `3eec87b`.

## Test plan

Verified manually on an EE build (`--features quickjs,enterprise,private,license`) with a
`prodws` workspace and a `wm-fork-dev` fork, driving the same `POST /scripts/create?skip_if_noop=true`
call that `wmill sync push` makes, plus a real `wmill sync push` from a repo pulled with
`wmill sync pull`:

- [x] fork with `deploy_to = NULL`, script pushed with a lock present → **before:** 0 rows in
      `workspace_diff`; **after:** `ahead=1`, item listed under "Deploy to prodws"
- [x] fork with `deploy_to = 'prodws'` (normal fork), same push → still exactly `ahead=1`,
      no double-count from tallying both targets
- [x] parent-side deploy with the fork's `deploy_to = NULL` → `behind=1` (unchanged, confirms
      the asymmetry that motivated the fix)
- [x] script whose lock generation fails (unresolvable npm import) → **before:** 0 rows;
      **after:** `ahead=1`, `has_changes=true`, `exists_in_source=false`, listed as "New"
- [x] flow whose module lock generation fails → tallied
- [x] successful lock-gen deploys of a script / flow / app → still exactly `ahead=1`
- [x] `cargo check` (CE, no `private`) and `cargo check --features enterprise,private,license`
- [x] redeploy of an *existing* script (`parent_path == path`) with a failing lock job → `ahead=1`, not 2
- [x] genuine rename with a failing lock job → both old and new path tallied, once each
- [x] `SQLX_OFFLINE=true cargo check --tests` compiles (the offline cache covers the new test)
- [x] `cargo test -p windmill-api-integration-tests --features private,enterprise --test workspace_comparison` → 10 passed
- [x] regression guard verified by reverting the EE commit: fails with `left: None, right: Some(1)`
- [x] `./update_sqlx.sh` — four new query entries committed

## Screenshots

The fork's Compare & Deploy page listing a script whose lock job failed — previously this
list was empty:

compare-deploy-after-fix

## Known residual gap

A dependency job that is cancelled (or dropped) **while still queued** never reaches
`handle_queued_job`, so neither the success tally nor the fallback runs — the
`flow_version` / `app_version` / `script` row is already committed, so that change stays
untallied. Verified: cancelling a queued flow dependency job leaves `started_at NULL` and
no `workspace_diff` row. Same for a `SIGKILL`ed worker, or a `Connection::Http` agent
worker (which cannot run dependency jobs at all). Closing that would mean tallying at
deploy time in the API handlers and teaching the EE `handle_deployment_metadata` to skip
its own tally — a larger change than this fix warrants. As with every other case here, it
self-heals the moment the item is redeployed.

## Note on already-lost changes

This stops new changes from being dropped; it does not retroactively discover changes a
fork already lost. Those resurface as soon as the affected item is redeployed.

---
## Summary by cubic
Fixes missing fork diffs in Compare & Deploy by tallying fork→parent changes based on lineage, handling unset `deploy_to`, and recording tallies when lock jobs fail without double-counting. Addresses WIN-2259 and fixes #10401 so upstream deploys from forks are visible.

- **Bug Fixes**
  - Key the “ahead” tally to `workspace.parent_workspace_id`; fall back to `workspace_settings.deploy_to` only; dedup when both point to the parent.
  - Record a tally when dependency/lock jobs error or are canceled (scripts, flows, apps); rebuild the `DeployedObject`, read `app_version.raw_app`, and skip preview jobs with no `runnable_id`; prevent double-tally via a `deployment_tallied` flag.
  - Re-export `tally_deployed_object_changes` in `windmill-git-sync` with a CE no-op stub so OSS builds compile.
  - Add regression tests for forks without `deploy_to` and the dedup case; wait for async tally to settle; commit `.sqlx` offline cache.

- **Dependencies**
  - Bumped EE repo reference to the companion commit for the fork tally changes.

<sup>Written for commit cb373a51d412177c4545ce48cdfac23d16ba434e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ ab58f20fb95171d58cfc2e45a147dad8619449fd — nit-only verdicts at 59c0eaed156d707d3b35361206c4f3c02eb71b87 (codex: mergeable-with-nits, claude: good to merge, pi: good to merge); the codex nit (dedup assertion could observe a transient `ahead = 1` between the two upserts) is addressed in ab58f20fb9, which polls through a settling window. Verified the guard is real: temporarily dropping the dedup makes it fail with `left: Some(2)`.

PR #10404, opened 2026-07-29
fix(ai): route Azure OpenAI agent steps through the Responses API

rubenfiszel:
## Summary

Fixes WIN-2260.

The `AzureOpenAI` provider fell through to `OtherQueryBuilder` in `create_query_builder`, which targets `/chat/completions`. Newer Azure OpenAI deployments reject a `reasoning_effort` combined with function tools on that endpoint, so AI agent steps on those models fail. `AzureOpenAI` now uses `OpenAIQueryBuilder` like `OpenAI` does, targeting `/responses`.

`OpenAIQueryBuilder` already handles Azure through `is_azure()`: `build_azure_openai_url` resolves every stored base-URL shape (bare resource root, `/openai`, `/openai/v1`, `/openai/deployments/<id>`) to `<root>/openai/v1/responses`, and auth stays on the `api-key` header. This also aligns the agent step with the copilot chat, which already routes `azure_openai` through the Responses API (`frontend/src/lib/components/copilot/lib.ts`).

Azure serves the Responses API for only a subset of models and regions, and the routing decision cannot be made per model — on Azure the model name is a deployment name chosen by the user, which says nothing about the model behind it. So a resource outside that matrix would have regressed from a working `/chat/completions` call to a failing `/responses` one. Two things keep those resources working without giving up the fix for everyone else:

- **Per step**, a first call rejected with 400/404 re-routes to `/chat/completions` and stays there for the rest of the step. A rejection once the conversation is under way is not re-routed: by then a 400 is about the conversation (context length, content filter, tool schema), and re-routing would hide the real cause and pin a working `/responses` step to the endpoint this PR moves off.
- **Across steps**, the worker remembers the deployment so later steps start on `/chat/completions` directly — but only when the fallback actually answered *and* the rejection was a **404**, the one status that says the route is absent. Every 400 is ambiguous, since a deployment that does serve the route rejects tool schemas, blocked hosted tools and filtered content the same way, so a 400 re-routes that step alone and is not remembered. The cost of that conservatism is one rejected call per step for a deployment that answers 400 rather than 404, which is what this PR would do anyway without the cache.

## Changes

- `create_query_builder`: `AIProvider::AzureOpenAI` joins the `AIProvider::OpenAI` arm, so it builds an `OpenAIQueryBuilder`.
- New `QueryBuilder::supports_chat_completions_fallback(base_url)` (default `false`), implemented by `OpenAIQueryBuilder` as `is_azure(base_url)` — the only resources whose preferred surface may not be served. `create_chat_completions_query_builder` builds the replacement.
- `ai_executor`: the two rejections that are about the request shape rather than the model — `stream_options` and the route itself — are handled by one retry loop around the send. Each is retried once with that part dropped, and both decisions persist across the step's iterations, so a later iteration does not re-send a shape the endpoint already rejected. Errors name the endpoint that failed.
- `is_chat_completions_only` / `remember_chat_completions_only`: a worker-local cache keyed by endpoint **and** deployment name (Azure's support varies by both), holding only rejections, expiring after an hour so a deployment that gains Responses support is picked up. The factory stays pure — the cache is consulted in the worker, so the API proxy is unaffected.
- Parity test: the `AzureOpenAI` case now pins `responses` instead of `chat/completions`. The test asserts the proxy and the agent step agree on the endpoint and the credential header, so it fails if the routing regresses.

## Notes

- Web search is unchanged: Azure serves the same `{"type": "web_search"}` hosted tool on `/openai/v1/responses` for GPT-4-and-later deployments ([docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/web-search)), so moving Azure onto this endpoint *gains* working web search for agent steps, where `OtherQueryBuilder` silently ignored it. A subscription that blocks the tool rejects the request with a 400 instead; that step re-routes and completes without search, and nothing is remembered.
- `azure_foundry` deliberately stays on `/chat/completions` for non-Claude deployments: that resource kind also fronts model families (Llama, DeepSeek, Mistral, Phi) with no Responses API. Now that a probe-and-remember mechanism exists, applying it to Foundry is a plausible follow-up, but it needs its own decision and is out of scope here.
- The cache is per worker process, so each worker pays at most one probe per deployment per TTL. A shared or DB-backed probe would cost more than the single rejected call it saves.

## Test plan

- [x] `cargo test -p windmill-ai` — 105 passed, including the parity case pinning the new route and a guard that the cache is keyed per deployment. `cargo check -p windmill-worker` clean.
- [x] End-to-end against a mock endpoint, running real AI agent steps with an inline provider resource (`reasoning_effort: medium`):
  - `azure_openai`, endpoint serves Responses → one call, **`POST /openai/v1/responses`**, `api-key` header, Responses-shaped body (`input` / `instructions` / `reasoning: {effort}`). Before the change the same step went to `/chat/completions`.
  - `azure_openai`, endpoint 404s Responses → `/responses`, then **re-routed to `/chat/completions`** with the same `api-key` header and a well-formed completions body; step completed on the fallback. Two consecutive steps → the endpoint saw `/responses` **once**: a 404 is remembered.
  - Endpoint 400s `/responses` with `Invalid schema for function ...`, and separately with `The tool web_search is not supported for this subscription` → the step re-routes and succeeds both times, and the **next step probes `/responses` again**: no 400 is remembered.
  - `customai`, endpoint 400s `stream_options` with "Additional properties are not allowed", model calls a tool → call 1 with `stream_options` rejected, call 2 without it returns the tool call, call 3 (with the tool result) sent without `stream_options`. The pre-existing retry still works for every OpenAI-compatible provider and is not re-litigated per iteration.
- [ ] Verify against a real Azure OpenAI resource on a Responses-capable deployment with a function tool attached.

### DIFF
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -857,6 +857,7 @@ mod tests {
             messages,
             tools: None,
             model: "claude-sonnet-4",
+            base_url: "https://api.anthropic.com/v1",
             temperature: None,
             reasoning_effort: None,
             max_tokens: None,
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -857,7 +857,6 @@ mod tests {
             messages,
             tools: None,
             model: "claude-sonnet-4",
-            base_url: "https://api.anthropic.com/v1",
             temperature: None,
             reasoning_effort: None,
             max_tokens: None,
--- a/backend/windmill-ai/src/providers/mod.rs
+++ b/backend/windmill-ai/src/providers/mod.rs
@@ -75,30 +75,6 @@ lazy_static::lazy_static! {
 
 const CHAT_COMPLETIONS_ONLY_TTL: Duration = Duration::from_secs(3600);
 
-/// Whether a rejection means the route is absent for this deployment, rather than
-/// something about the request the step happened to build (a tool schema, a hosted tool
-/// the subscription blocks, the conversation itself), which a deployment that does serve
-/// the route rejects just as well. A 400 has to name the deployment to qualify, so a
-/// complaint about one part of the request is not read as a verdict on the endpoint.
-///
-/// Only a rejection that passes this may be remembered for later steps. Being wrong the
-/// other way is cheap: the route is still retried per step, it is just learned again.
-pub fn identifies_unserved_route(status: u16, body: &str, model: &str) -> bool {
-    if status == 404 {
-        return true;
-    }
-    let body = body.to_ascii_lowercase();
-    body.contains(&model.to_ascii_lowercase())
-        && [
-            "unsupported",
-            "not supported",
-            "does not exist",
-            "not found",
-        ]
-        .iter()
-        .any(|marker| body.contains(marker))
-}
-
 /// Whether this deployment is known not to serve the endpoint its provider prefers.
 pub fn is_chat_completions_only(base_url: &str, model: &str) -> bool {
     CHAT_COMPLETIONS_ONLY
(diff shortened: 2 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 8d164c051771791fc7f660da1a2aba18d1751d64 — nit-only verdicts at fc0ccec3594a337a06f48ebec07aabb0777e700d (codex: mergeable with nits; pi: good to merge; the Claude reviewer errored out without posting a verdict); the nit is addressed in 8d164c0517.

Codex's nit was right: `include_usage` was scoped inside the agent loop, so the comment overstated what persisted and every later iteration re-sent the `stream_options` shape the endpoint had already rejected. It now lives beside `query_builder`, so a rejected request shape and a rejected route both stay dropped for the rest of the step, and the comment is back within the 4-line rule.

Verified against a mock endpoint with a multi-iteration step (endpoint 400s `stream_options`, model calls a tool): request 1 with `stream_options` → 400, request 2 without it → tool call, request 3 (with the tool result) sent without `stream_options` — no repeated rejected round-trip.

rubenfiszel:
✅ Review round clean @ 4515e19a83cd63b419763528915bfc2ca9be61a5 — nit-only verdicts at c6c2650f8c20610eb231a42a226dc1dbef71aece (codex: mergeable with nits; claude: mergeable with nits; pi: good to merge); nits addressed in 4515e19a83.

Both comment nits were right and are fixed:
- `ai_executor.rs` reroute-flag declaration said the flag is set whenever the send re-routed, which stopped being true when c6c2650 narrowed the cache to a 404. It now describes what it does: set where the route is found absent, read once the fallback answered.
- `query_builder.rs` `supports_chat_completions_fallback` doc was 5 lines; now 4, same invariant.

No behavior change in this commit: `cargo check -p windmill-worker` clean, `cargo test -p windmill-ai` 105 passed.

On the recurring test-coverage note (no committed worker test for the fallback/cache state machine): that path is HTTP-status-driven and the crate has no mock server, so it is covered by the mock-endpoint runs in the test plan instead — 404 remembered, 400s not, `stream_options` retry intact across tool iterations. The remaining open box is verification against a real Azure resource, which needs credentials this environment does not have.

PR #10410, opened 2026-07-30
feat: make the fork lineage the only deploy relationship

rubenfiszel:
## Summary

`workspace.parent_workspace_id` and `workspace_settings.deploy_to` both answered "which workspace does this one deploy into", and nothing kept them in agreement. `deploy_to` (2023) predates forks (2025) by two years: it was the "link this workspace to a staging/prod workspace" setting for two otherwise unrelated workspaces. Fork creation and `attach_dev_workspace` then started seeding it to the parent, so the two overlapped without either being authoritative — every reader picked one, and they disagreed.

That has already produced shipped bugs. #10403 made the ahead tally lineage-first because a fork whose `deploy_to` was cleared recorded no ahead change at all. `detach_dev_workspace` clears the parent but leaves `deploy_to` pointing at the old prod (reproduced on a dev instance). `edit_deploy_to` accepted any string with no validation, and the settings page offered every workspace you belong to with nothing marking which one was the real parent.

This drops `deploy_to` and makes the fork lineage the only relationship. It also fixes job-tag routing, which was never lineage-aware at all.

## Job tags

Two paths derived a tag's workspace id and only one knew about forks:

- `per_workspace_tag` mapped any workspace with a parent to that parent.
- `$workspace` in an explicit tag always interpolated the **raw** id.

Measured on a dev instance by pushing a job and reading back `v2_job.tag`:

| workspace | `default_tags_per_workspace` | explicit `$workspace` tag | default tag |
|---|---|---|---|
| standalone | ON | `mytag-ws` | `bun-ws` |
| same, attached as dev of prod | ON | `mytag-ws` | **`bun-prod`** |
| ephemeral fork | ON | **`mytag-wm-fork-x`** | `bun-prod` |
| ephemeral fork | **OFF** | **`mytag-wm-fork-x`** | `bun` |

Two distinct bugs. A fork inherits its parent's scripts, so a script tagged `mytag-$workspace` that runs fine in prod produces `mytag-wm-fork-x` in the fork and queues forever — no worker is ever provisioned for a generated id, and this is not gated by `default_tags_per_workspace` (default `false`). Separately, attaching a dev workspace silently moved its default-tag jobs onto the parent's pool, idling any worker dedicated to it.

Both paths now go through `tags::tag_workspace_id`, which walks to the nearest ancestor whose id an admin would actually provision workers for:

| workspace | resolves to |
|---|---|
| root | itself |
| ephemeral fork | nearest servable ancestor |
| fork of a fork | past the intermediate generated id |
| dev workspace | itself |
| dev workspace with a `wm-fork-` id | keeps walking |
| fork under a dev workspace | that dev workspace |
| orphaned `wm-fork-*` (parent deleted) | itself |

Resolution only runs when the tag actually contains `$workspace`, so the job-push hot path keeps its previous cost.

## Migration

Legacy pairs fold into the lineage, then the column drops.

A pair becomes a **dev workspace** when it is the sole claimant on a target that has no dev workspace yet — that is what a long-lived staging paired with prod is, and it keeps its own tag domain and promotion mode, so nothing changes underneath it. Where several workspaces name the same target, or the target already has a dev workspace, `workspace_canonical_dev_idx` forces a **plain fork** instead; each is reported by name.

Pairs that cannot be expressed as lineage at all are skipped, reported with a reason, and left unlinked: a target that does not exist (`deploy_to` had no FK), a self-reference, a workspace that is already a fork, or a chain where the target itself has a target — which also covers mutual pairs.

`list_ws_specific_versions` is rewritten to walk `parent_workspace_id` instead of the `deploy_to` graph. Same traversal, different edge; it also closes a gap where forks were never considered related.

## Changes

- migration `20260730080304`: convert, rewrite the `ws_specific` traversal, drop the column
- `windmill-queue/src/tags.rs`: `tag_workspace_id` ancestor walk, shared by `per_workspace_tag` and both `$workspace` interpolation sites (`jobs.rs`, `worker_flow.rs`); attach/detach now invalidate descendants since resolution spans the chain
- `windmill-api-workspaces`: `edit_deploy_to` deleted; `get_deploy_to` returns the fork parent under the same response field, so the deploy drawer and the resource/variable editors are untouched; seeding removed from fork creation, dev attach, the settings clone, and the rename remap
- `windmill-api/src/workspaces_export.rs`, `openapi.yaml`, generated client: field and endpoint removed
- frontend: the target is read-only on a fork; the Deployment UI tab is hidden for roots, which have nothing to link
- `cli/src/core/settings.ts`: `deploy_to` no longer read or written; older settings files carrying it are ignored rather than erroring
- EE companion: the ahead tally drops its `COALESCE` fallback

`workspaces.edit_deploy_to` stays in the openapi audit-action enum — the endpoint is gone but historical audit rows still carry that action, and removing the variant would break filtering of existing history.

## Behavior change on upgrade

A pair converted to a dev workspace keeps its tags and promotion mode; on cloud its usage now attributes to the parent's billing root.

A pair converted to a plain fork additionally has `$workspace` resolve to the parent (regardless of `default_tags_per_workspace`, so workers on the old tag go unserved) and moves git-sync deploys to `wm-fork/**` without promotion mode. Re-attaching one of them as the dev workspace restores both.

## Cache and traversal

`FORK_PARENT_CACHE` is per-process, so `invalidate_fork_parent_cache` only ever cleared the process handling the request and replicas served stale lineage until the 300s TTL. That predates this PR, but resolution becoming ancestor-dependent widens which mutations land in that window, so lineage changes now broadcast over `notify_events` and every process clears the cache. The listener clears all of it rather than tracking ids: one mutation invalidates an unbounded descendant set, and re-resolving costs a single lookup per active workspace on the next push.

`list_ws_specific_versions` walks up through ancestors and down only to a **dev workspace**. Descending into plain forks would make a root fan out over its whole live fork subtree, and each member costs an identity lookup plus an RLS switch and probe; a dev workspace is the paired editable environment rather than a throwaway copy, so a prod still sees it. There is at most one per parent and attach rejects nested dev chains, so that edge stays bounded.

The walk also gained a visited-path guard. Its edges run both ways, so the recursion never converged — it bounced parent↔dev until the depth cap on every call, 33 rows for a two-member set, and did so on `main` too over the `deploy_to` graph. It now ends as soon as nothing new is reachable, which makes this function strictly cheaper than before rather than a tradeoff.

## Deliberately not changed

- `dedicated_worker_tag` still uses the raw workspace id. Resolving it to the parent would be worse: a dedicated worker has a specific script version pinned, so a fork's job would execute the parent's code.
- `$workspace` in concurrency keys and volume names — different semantics from tag routing; a fork arguably should have its own concurrency budget.

## Test plan

- [x] Migration on a seeded scratch DB: sole-claimant pair → dev workspace; two claimants on one target → both plain forks; target with an existing dev → plain fork; self-reference, dangling target, chain, mutual pair → skipped with reasons; pre-existing fork and dev workspace untouched
- [x] `cargo test -p windmill-queue --test tag_workspace_test` — seven lineage shapes including nested forks and a generated-id dev workspace
- [x] `npm run check` (frontend) — 0 errors; `npm run check` (cli) — 16 pre-existing errors, unchanged
- [x] Dev instance: `get_deploy_to` returns the parent for a fork and null for a root; `edit_deploy_to` 404s
- [x] Tag routing measured before and after for standalone / attached-dev / ephemeral-fork / root, with the setting on and off; a tag without `$workspace` passes through untouched
- [ ] Reviewer: confirm the dev-workspace-when-possible migration rule, and the plain-fork fallback for many-to-one pairs

## Screenshots

Fork settings after unification — the target is read-only and the item filters remain:

unified-fork-settings

EE companion: windmill-labs/windmill-ee-private#694 (must merge together — the old query loses its column).


rubenfiszel:
Round 4 responses.

**Fixed — `archive_workspace` sweep (Claude).** You were right, and I had the rationale backwards twice. `UPDATE workspace SET is_dev_workspace = false` lives inside `if let Some(prod) = dev_lock_parent`, so it only runs for an actual dev workspace. The sweep now sits in that same block, which makes it correct and stops issuing a `list_fork_descendants` query on every unrelated archive. (My previous comment there also claimed the tag walk skips archived ancestors — it has no `deleted` filter at all. Both comments described code I hadn't read.)

**Contesting — worker replicas retain stale tag-workspace cache entries (Codex).** Accurate, but pre-existing rather than introduced here. On `origin/main`, `FORK_PARENT_CACHE` is already `lazy_static` + `quick_cache` with a process-local `invalidate_fork_parent_cache`, so a worker replica has always kept a stale fork→parent mapping for the 300s TTL after any lineage mutation. This PR does not change that property; the invalidation calls it adds strictly reduce staleness in the process that performs the mutation.

What the PR does change is blast radius: resolution is now ancestor-dependent, so more mutation types (archive, rename, delete) and deeper descendants fall in the stale window. The failure mode and its 300s bound are unchanged.

A real fix has infrastructure ready — `windmill-common/src/notify_events.rs` with the `process_notify_event` dispatch in `main.rs` — and would be a channel constant, sends at the five-plus mutation sites, and a match arm. I'd rather do that as a follow-up than expand a PR that is already five rounds deep with a change to a shared cache mechanism that is orthogonal to the unification. Happy to fold it in if you disagree.

**Still open — `list_ws_specific_versions` fan-out (Claude, third round).** Deliberate. The traversal widened from a `deploy_to` pair to the ancestor chain plus live fork subtree because that is what makes the function correct for forks, which previously were never considered related at all. Narrowing it reintroduces that gap. Recorded in the PR body as an accepted cost.

**Not added — preprocessed-flow tag test (Codex, round 3).** The existing preprocessor tests in `tests/dedicated_workers.rs` all need a worker executing a preprocessor step end-to-end; reproducing that to assert one tag string is a large fixture for a one-line placement fix. The bug was where the call sat, not what it computed, so nothing lighter would catch a regression.

rubenfiszel:
✅ Review round clean @ 45f63b1b34b69a257e2106aab11bb37f0f96a6e6

rubenfiszel:
✅ Review round clean @ 255cc9396554fddf96898013b8a9a6ad5ccacdf3

PR #10411, opened 2026-07-30
feat(git-sync): dedicated base url for GitHub webhook delivery

rubenfiszel:
## Summary

Windmill registers its git-sync GitHub webhook at `<base_url>/api/w/{workspace}/github_app/webhook`. On instances that GitHub cannot reach on their browser-facing URL, there was no way to point the hook elsewhere, so delivery could never work and those repositories silently fell back to polling.

This adds an optional instance setting — **Instance settings → GitHub App → Webhook base url** — used to build the receiver URL, falling back to `base_url` when unset. It mirrors the existing `hub_base_url` / `hub_accessible_url` split: `base_url` stays the browser-facing URL (the check-run "details" link still uses it), while the new setting is only what GitHub must be able to reach.

The setting is expected to be configured at instance setup, before any GitHub App is connected — at which point there is nothing registered to move. Changing it later deliberately does **not** touch webhooks GitHub already holds; instead the settings page lists the workspaces whose hook still points at a previous receiver, and re-saving those workspaces' git sync settings moves them.

## Changes

- New global setting `github_app_webhook_base_url`, also exposed on the declarative `GlobalSettings` instance config, rendered in the GitHub App section. It applies to the managed and self-managed app alike, so it is not gated behind the self-managed toggle.
- `webhook_receiver_url` (EE) builds the receiver from the new setting, else `base_url`.
- `AutoPullSettings.webhook_url` records the receiver a live hook was registered with. It is server-owned: stripped from client input, preserved across both git-sync save paths, and cleared on workspace rename and dev-workspace attach alongside the other webhook fields.
- When a workspace saves its git-sync settings, a hook whose receiver no longer matches is re-registered. For a hook created before `webhook_url` existed there is no local copy, so Windmill asks GitHub for the URL it actually holds: matching hooks are recorded, mismatched ones replaced, and ones GitHub reports gone (404) re-created. Only an unreadable hook is left alone.
- Replacement registers the new hook before deleting the old one, and rolls the new one back if recording it fails — so a failure can never leave a hook live but untracked, nor drop the only hook a repository has.
- `GET /settings/github_app_stale_webhooks` (superadmin) reports repositories whose registered receiver differs from the configured one. Read-only; it never mutates a hook.
- Validation (`validate_webhook_base_url`) rejects values that cannot work once the receiver path is appended — wrong scheme, no host, whitespace, query string, fragment, trailing slash, and embedded credentials — on all three write paths (per-key API, bulk instance-config, declarative `sync-config`/operator). No rejection message echoes the submitted value, since these strings reach `sync-config` output and operator logs.

## Screenshots

Changing the setting while a workspace still holds a previous receiver:

stale-webhooks

## Test plan

Unit tests cover receiver construction, the validator (a shared accept/reject table pinned on both the Rust and TypeScript sides), and that no rejection message can carry a credential. An integration test asserts the declarative path validates before writing anything.

GitHub-facing behaviour was exercised against a mock GHES API (installation token, repo access, hook create/read/delete), asserting the URL actually registered:

- [x] Fresh repo with the override set registers the hook at the override host, not `base_url`
- [x] Hook registered before this change, still on the old URL → read from GitHub, re-registered at the override
- [x] Hook already on the current receiver → recorded locally, no delete/create
- [x] Hook deleted on GitHub (404) → a fresh one is registered, and it stops being re-probed
- [x] Override cleared → falls back to `base_url`
- [x] Changing the setting makes **no** GitHub calls, and the stale report lists the affected workspace
- [x] Re-saving that repository moves the hook and empties the report
- [x] Replacement create fails → the old hook is kept and still tracked
- [x] Recording the replacement fails → the new hook is rolled back, not left untracked
- [x] A repo that cannot register a hook settles to `mode = polling`, and that is persisted
- [x] Invalid values (bad scheme, bare host, query/fragment, whitespace, trailing slash, credentials) are rejected with 400 and the field shows the error inline
- [x] Verified in the browser: the setting saves, the warning appears after saving a changed URL, and it lists the workspace and repository


rubenfiszel:
Round-3 responses, plus two reviewer questions answered.

### Addressed in `34b66e9`

**Codex #1 — lock excluded persistence.** Correct, and the diagnosis was exact: `sync_repo_webhook` released the lock on return while `workspaces.rs` persisted later. The write-back now lives inside the critical section — `sync_repo_webhook` owns it (`reconcile_and_persist_locked`), and both workspace-save call sites no longer persist those fields themselves. `persist_reconciled_webhook_fields` is deleted; `mode` moved into the EE writer so the webhook→polling normalization is still durable (regression-checked, see below).

**Codex #2 / Claude P2 — declarative writer docs.** Added the `AUTHORIZATION:` contract, and corrected the overclaim: the doc now names the one rule it applies and states explicitly that `automate_username_creation`, `critical_alert_mute_ui` and `app_workspaced_route` are *not* inherited, so nobody assumes a new pre-write-hook rule propagates here for free. The "awaited, not detached" rationale is restated in terms of an operator tick not reporting success early, rather than "the process exits".

### Answering @claude's "was the operator actually routed?"

Yes — you were right to ask rather than assume, and right that the OSS-visible evidence pointed the other way. `windmill-operator/src/db_sync_ee.rs` now imports `sync_global_settings_declarative` and its `sync_global_settings` body is `sync_global_settings_declarative(db, &current, desired).await`. It's in EE `6dc266f`, which is why your checkout at `caf6abc` couldn't see it. Verified compiling with `--features enterprise,private,operator`.

### Deliberately not implementing: cross-replica serialization

Codex asks for the read → mutate → persist critical section to be *cross-replica*. I've closed the in-process half and am declining the distributed half, with the reasoning recorded on `repo_webhook_lock` rather than left implicit:

- The remaining exposure needs two API replicas reconciling the **same repository in the same instant**. The outcome is one extra hook on GitHub whose id Windmill doesn't track — the receiver drops its deliveries, since it matches no stored `webhook_id` — recoverable by deleting that hook or toggling auto-pull.
- The fix would be a Postgres advisory lock held across multi-second GitHub calls on a dedicated connection, and the pre-existing settings-save and poller paths would both have to take it. That trades a rare, recoverable duplicate for connection-pool exhaustion and a new deadlock surface, in paths this PR otherwise doesn't touch.

Happy to do it as a follow-up if you'd rather have it, but I don't think it belongs bolted onto this PR.

### Verification

10-case mock-GHES matrix, all passing. New this round: **mode normalization** (non-app-backed repo still settles to `polling` and persists after the write-back moved into EE) — the specific thing dropping `persist_reconciled_webhook_fields` could have broken. Retained: override honored; pre-existing hook re-pointed via a GitHub read; no churn when already correct; 404 left alone; fallback to `base_url`; setting-change-only re-point; idempotent re-save; failed DELETE keeps the hook with no duplicate and a surfaced error; and all malformed values rejected 400. Also re-ran both `sync-config` paths (invalid rejected with nothing written; valid change re-points).

`cargo check` clean on CE, `enterprise,private`, and `+operator`; unit tests pass; frontend clean; `.sqlx` shows only the one now-dead query removed.

rubenfiszel:
### @pi — your round-15 review is against a stale ref

Your review states it covers `94bcc005` → **`5aa0e514c`**, but the head at the time was **`9e1b41c236`** — `5aa0e514c` is five commits behind it. All four findings describe state that the commits after `5aa0e514c` had already changed, so I've verified each against the actual head rather than acting on them:

| Finding | State at `9e1b41c236` |
|---|---|
| P1 userinfo check removed | Present. `validate_webhook_base_url` rejects `!url.username().is_empty() \|\| url.password().is_some()`, added in `6b26ff1f` — which is *after* the commit you reviewed. |
| P1 declarative path passes non-strings | Rejected. `sync_global_settings_declarative` matches on the JSON shape and returns `"{key} must be a URL string, got {other}"` for bool/number/object/array. |
| P2 errors echo the submitted value | No branch interpolates it. Parse failure, host, query/fragment, whitespace and trailing slash all return static text; the scheme branch was the last one and is fixed in `abbcd841`. |
| P2 lockstep validator test deleted | Present at `frontend/src/lib/components/instanceSettingsWebhookBaseUrl.test.ts`, 20 cases, passing. |

You were right about all four *as of `5aa0e514c`* — that's genuinely what the code looked like then, and the userinfo leak in particular is the one @codex raised at that commit and I fixed in the next one. Nothing to change here; flagging it so the blocking verdict isn't read as describing the current branch.

### @codex round 15 — all three addressed

- **P1 cross-replica CAS contention** (`abbcd841` / EE `d830aa05`): you're right that comparing only `webhook_id` lets the stale sweep win and leaves the newer one with no rerun. On a failed write-back the reconcile now requests a fresh sweep through the coalescing queue, which re-reads the settings and converges on the newest receiver. Coalesced, so replica contention can't turn into a storm.
- **P2 scheme echoed**: fixed — `hunter2xyzzy://host` parses, so the scheme is attacker-chosen text like any other. Static message now, with that exact value added to `webhook_base_url_errors_never_echo_credentials`.
- **P2 sweep docs**: `reconcile_repo_webhooks` now says it retries failed attempts and may register a hook that never got created, instead of claiming it only re-points existing ones.

rubenfiszel:
### Scope correction: the reconcile lock and CAS are gone

@codex your R17 P1 (a newer polling-only save losing to an in-flight webhook-mode reconcile) is real, but it is **pre-existing on `main`** and I'm treating it as out of scope rather than fixing it here.

`origin/main` calls `sync_repo_webhook` from the same two post-commit save paths with **no** lock, **no** compare-and-set and no committed-state refresh — `grep -cE 'repo_webhook_lock|expected_webhook_id|IS NOT DISTINCT FROM'` on `origin/main:windmill-common/src/git_sync_ee.rs` returns `0`. Two overlapping git-sync saves of the same repository already race there today; this PR doesn't introduce or widen it.

I'd built that machinery for the instance-wide sweep, where contention was routine (a sweep touching every workspace, racing every save). With the sweep deleted, its only remaining job was a race that `main` already has, so I've removed it: the per-repository mutex, the under-lock committed refresh, the `webhook_id`/`mode`/`enabled` CAS, and the `mode_changed` conditional that existed only to make the CAS safe. That's another **157 lines deleted** and restores `main`'s concurrency posture exactly.

What I kept, because it belongs to behaviour this PR actually adds — re-pointing a hook whose receiver moved:

- **create-then-delete**, so a failed create can't leave a repository with no hook at all
- **rollback**, so a failed write-back can't leave a hook live on GitHub with nothing recording it
- the `EXISTS` guard on the write, so a write matching no repository is an error rather than a silent success (that's what triggers the rollback)

If save-vs-save is worth closing, it wants its own PR covering every `git_sync` writer — the poller and the workspace-lifecycle paths as well — not a half-measure bolted onto this one.

### Also addressed from R17

- **P2 stale sweep contract on `StaleWebhookRepo`** — my earlier deletion left a truncated sentence and an authorization contract claiming mutations it never performs. Removed.
- **P2 `registered_url` non-nullable** — legacy hooks legitimately serialize `null`; marked `nullable: true` and regenerated the client.
- **@claude nit** — the component rendered its own copy of the validation error the generic field already shows. Removed.

22-case mock-GHES matrix passing, three feature sets compile, 24 backend + 25 frontend + 1 integration test green, `.sqlx` +12/−0 against main.

rubenfiszel:
✅ Review round clean @ e2ed45cb9d9c1c88dc22f7d699b4f6e5bf35a117

PR #10412, opened 2026-07-30
fix: run the init script before dedicated workers install dependencies

rubenfiszel:
## Summary

A dedicated worker installs its language dependencies as soon as it is spawned, and that happened before the `INIT_SCRIPT` job had run. `INIT_SCRIPT` is the documented hook to prepare the host, so on a worker group that uses it to populate the CA store (`INIT_SCRIPT=update-ca-certificates`), the dependency install reached the network with a stale trust store and failed with `invalid peer certificate: UnknownIssuer`. Because a dedicated-worker startup failure sends the killpill, the whole worker process then exited, so the pod sat in CrashLoopBackOff while the same worker group without a dedicated worker was fine.

The window is not small: `create_dedicated_worker_map` was called before `queue_init_bash_maybe`, and the dedicated worker body runs in a `tokio::spawn`, so `uv` typically started a couple of seconds before `update-ca-certificates` finished. `uv` loads its trust store once per process, so every retry inside that process was doomed too.

Ordering alone cannot fix this: the init job is executed by the main job loop, which only starts once the dedicated workers have been spawned. Hence a gate.

## Changes

- `queue_init_bash_maybe` (and the rest of the `i_worker == 1` startup block) now runs before `create_dedicated_worker_map`, so the init job is queued before any dedicated worker exists, and a failure to queue it returns before spawning them
- New process-wide `INIT_SCRIPT_STATE` gate (`Pending` / `Completed` / `Aborted`, a `watch` channel) with `wait_for_init_script_completed`, awaited by the two boot-time spawn paths in `dedicated_worker_ee.rs` (`spawn_dedicated_worker`, `spawn_runner_group`) before they install anything
- Every path that gives up on the init script resolves the gate to `Aborted` (failed init job, failure to queue it, worker 1 leaving its loop), so waiters are released with `false` and skip the install instead of parking. Parking would hang teardown, which joins the dedicated worker handles
- First resolution wins, so a later shutdown cannot turn a failed init script into a success, and the wait takes the caller's own killpill receiver rather than a `resubscribe()` (which starts at the tail and would silently drop an already-sent killpill)

Companion EE PR: windmill-labs/windmill-ee-private#696

## Test plan

- [x] `cargo check --features enterprise,private` is clean
- [x] `cargo test -p windmill-worker init_script_gate` passes (pins that an aborted init script releases waiters with `false`, permanently). No feature flags: the gate is not feature-gated, and passing `--features private` alongside `-p windmill-worker` fails to build for an unrelated reason, since it does not propagate to `windmill-git-sync`
- [ ] With `INIT_SCRIPT=update-ca-certificates` and a dedicated worker configured, the worker logs `waiting for init script to complete before installing dependencies` before the dependency install, and the install then succeeds instead of failing with `UnknownIssuer`
- [ ] With `INIT_SCRIPT=exit 1` and `NUM_WORKERS=2`, the process exits promptly instead of running a doomed dependency install first
- [ ] SIGTERM a second or two after boot, while the dedicated workers are still parked on the gate, and the process exits rather than hanging on `all dedicated workers have exited`

---
## Summary by cubic
Ensure dedicated workers wait for the `INIT_SCRIPT` to finish before installing dependencies, avoiding stale CA store network calls and CrashLoopBackOffs. Prevents `UnknownIssuer` errors when `INIT_SCRIPT=update-ca-certificates` is used.

- **Bug Fixes**
  - Queue the init script before spawning dedicated workers; fail fast if queuing fails.
  - Add process-wide gate `INIT_SCRIPT_STATE` (Pending/Completed/Aborted) and `wait_for_init_script_completed`; dedicated workers wait before any install.
  - Resolve the gate from the post-processed outcome (accounts for `wm_failure`, non-agent servers); abort on init failure, queue failure, or if worker 1 exits early so waiters release and skip installs.
  - Resolve to `Completed` immediately when no `INIT_SCRIPT` is set.

- **Dependencies**
  - Update `ee-repo-ref` to the companion EE change for the init script gate.

<sup>Written for commit 2f9ed47d00962b375fd129607d9d6fc263f6391b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ fd631253e69c766a551ddf28047c713ab6bd5eb8

PR #10417, opened 2026-07-30
feat: let the merge UI target an arbitrary workspace

rubenfiszel:
## Summary

Fixes WIN-2266.

Folding `deploy_to` into the fork lineage left no way to deploy into a workspace that is
not the parent — the one-off custom migration. This puts that back in the merge UI, GitHub
style: the Compare & Deploy page accepts an arbitrary target workspace, behind a discrete
"Change target" control.

The tally that keeps `workspace_diff` current only ever records a fork against its parent,
so an arbitrary pair starts with no rows and would compare as identical. Rather than
tallying every possible pair, such a pair gets an explicit **full scan**: a new endpoint
enumerates every item on both sides as an undecided candidate, and the following comparison
is what evaluates them, keeps the differing rows and deletes the rest. That first comparison
is the expensive step (a per-kind query pair per candidate), which is why scanning is a
user-initiated action, its result is left in `workspace_diff` for later visits, and the UI
shows when it ran plus a Recompute button. An arbitrary pair also re-evaluates its remaining
candidates on every compare instead of trusting the cache — nothing else would ever
invalidate it — which keeps the list correct after a deploy at negligible cost once the
identical items have been pruned.

Direction is one-way (current → target): a cold scan has no deploy history saying which side
a change came from, so the update direction, conflicts and deployment requests are all
lineage-only concepts and stay hidden. The lineage comparison is untouched: the guards refuse
to scan a lineage pair (a one-way seed would overwrite the tally's direction) and attaching a
dev workspace drops any scan state the pair accumulated while unrelated.

Also folds the **Deployment UI** settings tab into **Dev workspace**, as the issue asked.

## Changes

**Backend**
- New `workspace_diff_full_scan` table recording when a non-lineage pair's candidate set was
  seeded; without it an empty `diffs` can't be told from "never computed".
- New `POST /w/{workspace}/workspaces/seed_full_diff/{target}`: enumerates every deployable
  item of both workspaces (one query per kind) into `workspace_diff` as `has_changes = NULL`,
  one-way. Requires admin of both sides; refuses self-comparison, a missing target, and a
  lineage pair.
- `compare_workspaces` learns the arbitrary case: it requires admin on both sides, returns
  `full_scan_at`, bypasses `skip_workspace_diff_tally` for a scanned pair, and re-evaluates
  every candidate instead of trusting the cache.
- A scan **replaces** the pair's candidate set rather than merging into it, and collapses the
  two kinds whose comparison keys on less than `(kind, path)` — a renamed data table migration
  (`(datatable, timestamp)`) and an app converted to a raw app (path) — onto this workspace's
  key. Without either, one logical item would list and deploy twice.
- `compare_two_apps` compares `app_version.raw_app`, so converting an app to a raw app is a
  change rather than comparing identical. This also fixes the plain lineage flow.
- Purge the new table alongside `workspace_diff` on fork create/delete and workspace rename,
  and drop the pair's diff state when `attach_dev_workspace` turns two unrelated workspaces
  into a lineage pair.

**Frontend**
- `CompareTargetPicker`: a subtle "Change target" popover listing the caller's workspaces,
  with the cost of retargeting spelled out and a way back to the parent.
- `/forks/compare?target=<ws>` drives the comparison; the merge card gains the unscanned
  prompt, the scan timestamp + Recompute, a one-way banner, and a "Removes in target" badge
  (opted into row by row) for items the current workspace lacks.
- A failed comparison renders the reason and keeps both a Retry and the picker reachable
  instead of a blank page; only the newest in-flight comparison may land, so a retarget or a
  post-deploy poll can't overwrite the current target's result.
- A recompute drops any selected row that has become "Removes in target", so a tick made while
  the item merely differed can't authorize a deletion in the target.
- A root workspace with no parent lands on a target-selection prompt instead of a dead end.
- Deployment requests and the fork datatable schema diff are hidden for an arbitrary target —
  both are anchored on the lineage.
- Workspace settings: the **Deployment UI** tab is gone, its deployable-items filters now live
  under **Dev workspace** (which also gained a "Merge into another workspace" button);
  `?tab=deploy_to` redirects.

## Screenshots

The normal lineage merge UI. Retargeting has no control of its own — the destination
badge is the picker, so the header stays as it was:

lineage

Clicking that badge opens the picker:

picker

An arbitrary target selected, diff not computed yet:

unscanned

After computing it — scan timestamp, Recompute, the one-way banner, and the opt-in
"Removes in target" row left unticked:

computed

The Deployment UI settings folded into Dev workspace: the lineage destination first,
then its filters, then the arbitrary-merge entry point last:

settings

A root workspace with no parent:

no target

A target the caller cannot compare — the reason plus both ways out:

error

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --features enterprise,private --test workspace_comparison` — 11 passed, including the new `test_full_diff_scan_against_arbitrary_workspace` (unscanned pair, scan pruning identical items, one-way direction, renamed-migration and app/raw-app collapse, candidate-set replacement on re-scan, admin-of-both refusal on both endpoints, lineage-pair refusal). The re-scan assertion was confirmed load-bearing by reverting the fix — it fails with the duplicate it guards.
- [x] `npm run check` — 0 errors; `SQLX_OFFLINE=true cargo check` validates the regenerated query cache
- [x] Manually: pick an arbitrary target from a dev workspace, compute the diff, deploy the
      selected items, confirm they landed in the target and dropped off the list; create a new
      item and confirm Recompute picks it up
- [x] Manually: retarget to a third workspace and confirm the card remounts with no carried-over
      selection or diff
- [x] Manually: select a differing item, delete it from this workspace, Recompute — the row
      returns as "Removes in target" and is unticked
- [x] Manually: as a user who is not an admin of the target, confirm the error and both recovery
      actions (Retry, Change target) render
- [x] Manually: `?tab=deploy_to` redirects to the merged Dev workspace tab; the lineage
      comparison still offers both directions

---
## Summary by cubic
Let the Compare & Deploy UI target any workspace, not just the parent. Adds a one‑way merge flow guarded by a full diff scan, with a cleaner picker and clearer messaging. Implements WIN-2266.

- **New Features**
  - Arbitrary targets: choose via the destination badge (now the same badge style with a chevron) or `/forks/compare?target=<ws>`. The flow is one‑way (current → target): hides “Update current”, shows a one‑way banner, and the compare view is reachable from Drafts when a target is set. Root workspaces get a target‑selection prompt.
  - Full scans: required outside the lineage. UI shows “Compute diff”, last scan time (`full_scan_at`), and “Recompute” (replaces the candidate set, preserves opt‑ins, and drops rows that flip to “Removes in target”). Only the newest in‑flight comparison may land; failures keep “Retry” and the picker reachable.
  - Diff correctness: collapses renamed datatable migrations and app/raw‑app conversions onto this workspace’s key; converting an app to a raw app counts as a change.
  - UX scope: items that exist only in the target show “Removes in target”, are not bulk‑selected, and are auto‑unselected if a recompute flips them. Deployment requests and the datatable schema diff are hidden for arbitrary targets. “Target workspace protection active” clarifies rule scope.
  - Backend: adds `workspace_diff_full_scan` and `POST /w/{workspace}/workspaces/seed_full_diff/{target}`. `compare_workspaces` supports arbitrary pairs, returns `full_scan_at`, re‑evaluates all candidates for non‑lineage pairs, requires admin on both workspaces, and refuses lineage pairs. Full scans and dev attachment are serialized with a per‑pair advisory lock; state is purged on fork lifecycle, workspace delete/rename, and when linking a dev workspace. OpenAPI documents the endpoint and `full_scan_at`.
  - Settings: “Deployment UI” moves under “Dev workspace”, which also links to “Merge into another workspace”; existing `?tab=deploy_to` links redirect.

- **Migration**
  - Run DB migrations.
  - To use: open Compare & Deploy, click the destination badge (or pass `?target=<ws>`), pick a workspace, then “Compute diff”. You must be an admin of both workspaces. Use “Recompute” to refresh or “Retry” after a failed comparison. Existing `?tab=deploy_to` links now land on “Dev workspace”.

<sup>Written for commit 03900f38e36762a1ace9ecb679353afac7250346. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 00b2189c58a063ad0702b15e0e4b39affef861de — nit-only verdicts at 85a01c3785458194773077d5f773e432d68dba41 (pi: good to merge; claude + codex: mergeable, nits only); both nits addressed in 00b2189c58a063ad0702b15e0e4b39affef861de

PR #10420, opened 2026-07-30
feat: lazily expand s3 explorer folders one level at a time

diegoimbert:
## Summary

The S3 file explorer currently lists the whole bucket 1000 keys at a time and rebuilds the tree client-side. This changes it to load **one folder level at a time**, with an individual "Load more" per folder for levels that overflow.

EE companion: windmill-labs/windmill-ee-private#697 (the listing module and endpoint are `*_ee.rs`, so they don't appear in this diff — `backend/ee-repo-ref.txt` is bumped to pull them in).

## Why not `object_store`

It offers delimited listing **or** paginated listing, never both: `list_with_delimiter` returns folders but drains every continuation token internally and buffers the whole level, and `list_with_offset` paginates but has no delimiter so it yields no folders. The EE module calls the provider REST APIs directly instead — S3, Azure Blob and GCS all bound folders and files together under one max-keys and expose a resume token.

## Changes

- `backend/windmill-api/openapi.yaml` — `GET /w/{workspace}/job_helpers/list_stored_files_paged` returning `{folders, files, next_page_token}`, plus `StorageFolder` / `StorageFile` schemas. The existing `list_stored_files` is untouched (still used by the CLI and the git-repo viewer).
- `frontend/.../S3FilePickerInner.svelte` — lazy per-level expansion with a per-folder "Load more"; a preselected file is revealed by loading each ancestor level in turn.
- `frontend/.../GitRepoViewer.svelte` — passes `lazyFolders={false}`; it overrides the list request with `listGitRepoFiles`, which has no paged counterpart, so it keeps the flat path.
- `backend/Cargo.toml`, `windmill-api/Cargo.toml` — `quick-xml` (already in the lock file transitively) for parsing the provider list responses.

### Implementation notes

- Tree order is a plain lexicographic sort of full keys: a child always starts with its parent's `/`-terminated prefix, so lexicographic order **is** depth-first. The "Load more" row is a sentinel key `prefix + '￿'`, which sorts last among exactly that folder's children (visible in the screenshot below — it sits after `file-099.txt` and before the next root-level sibling).
- Searching stays flat: with a non-empty filter the picker falls back to `list_stored_files`, since per-level listing has no cheap cross-bucket search.

### Drive-by bug fix

The filter box never worked. `prefix: rootPath ?? (filter.trim() != '' ? filter : undefined)` — `rootPath` defaults to `''`, which is not nullish, so `??` never fell through and the filter was never sent to the server; typing in it just refetched the same unfiltered first 1000 keys. Now tests `rootPath !== ''`. Happy to split this out if preferred.

## Behaviour changes

- The "N items (including inside folders)" counter is gone in lazy mode — it would only count what happens to be expanded. Replaced with "N items shown".
- Folders always show a chevron; whether one is empty isn't knowable without fetching it. A folder can also appear and expand to nothing when advanced permissions deny everything inside it (see the EE PR for why that direction of error is the safe one).

## Screenshots

Lazy expansion — one request per level (`prefix=folder1/`, then `prefix=folder1/nested/`):

3-nested

A folder with 130 objects loads 100, with "Load more" scoped to that folder:

4b-many-bottom

After "Load more" — 130 loaded, row gone:

5-loadmore

Non-empty filter falls back to flat search:

6-filter

The bucket root pages the same way (110+ top-level objects — this case regressed once and is now covered):

root-loadmore

## Test plan

- [x] Driven in a real browser against this branch's backend; network trace shows exactly one `list_stored_files_paged` per expanded level, correct `page_token` continuation, and `list_stored_files?prefix=folder1/` once a filter is typed
- [x] 130-object folder: 100 → "Load more" → 130, row disappears
- [x] Root level re-verified after the review fix: 100 shown → "Load more" → 117, row disappears
- [ ] `npm run check` (full) not yet run
- [ ] Git-repo viewer (flat path, `lazyFolders={false}`) not re-verified in browser

https://claude.ai/code/session_01JqAc8mz6Gu698kBbJJVwcT


diegoimbert:
Round 3 fixes pushed (`b85abd26f1` / EE `9b9d1d3`).

**P0 — GCS OAuth tokens crossing resource boundaries.** Real, and exactly as described: the cache was keyed on `client_email` straight out of the caller-supplied resource, and consulted *before* the private key was used for anything. A resource claiming another principal's email with a junk key would have collected that principal's cached token.

Both token caches are now keyed on a SHA-256 of the credential material itself (`credential_cache_key`), never on an identity label — GCS on the full service-account key, Azure on tenant + client + the federated assertion, which required reading the assertion before the cache lookup rather than after. Regression test added (`test_token_cache_key_tracks_credential_material_not_identity`): same claimed email with different key material must not share an entry, identical material still shares, and providers never collide.

**P1 — `ambient_aws_credentials` contract.** Documented: these are the instance's own credentials, returned in the clear and typically outranking the requesting user, so callers must authorize the target themselves and must never surface them in a response, log, or error, nor hand them to a caller-supplied endpoint. Points at `build_s3_client` as the preferred path.

**P2 — `max_keys` schema.** Now declares `minimum: 1`, `maximum: 1000`, `default: 100`, matching the handler's clamp.

@claude re: the two things you couldn't check from this checkout — both confirmed against the EE tree and a live server:
- `restricted_access` is now a plain `bool` (not `Option<bool>`) on `ListStoredFilesPagedResponse` and is in the schema's `required` list, so it cannot be omitted. Verified on a running EE backend: `"restricted_access": false`, and the picker lists normally with no restriction alert.
- `max_keys` is clamped server-side by `query.max_keys.unwrap_or(100).clamp(1, MAX_KEYS_PER_PAGE)` with `MAX_KEYS_PER_PAGE = 1000`; the schema now declares the same bounds.

diegoimbert:
Round 5 fixes pushed (`34e3f66965` / EE `763b5b5`).

**P0 — noncanonical filesystem prefixes bypassing LIST rules.** Real, and the same shape as the folder-visibility bug from an earlier round: the prefix addresses the storage *and* is prepended verbatim to every returned key, and those keys are what the permission globs match. `./secret/` read the real `secret/` directory while producing keys a `secret/**` deny rule never matches, so a later broad grant would surface exactly what the deny rule exists to hide.

Fixed by requiring a canonical prefix — empty, or delimiter-terminated with no empty, `.` or `..` segments — validated in `list_page` so it covers **every** provider, not just the filesystem, with the filesystem's own absolute-path guard kept as defence in depth. `test_validate_prefix_requires_canonical_form` pins `./`, `a/./b/`, `../etc/`, `a/../b/`, `/etc/`, `a//b/` as rejected and the canonical shapes as accepted. Confirmed on a running EE backend: all five aliases now return `Bad request: Invalid prefix … must be a canonical path`, while `""` and `a/` list normally.

**P2 — `next_page_token` emitting `null` outside its schema.** Added `skip_serializing_if = "Option::is_none"` to both the internal `StorageListPage` and the handler's response struct, so an exhausted level omits the field rather than sending `null`. That matches the declared `type: string` (not required) instead of widening the schema. Verified by paging `a/` to exhaustion: the last page's JSON has no `next_page_token` key at all. The frontend's `?? undefined` normalization still handles both spellings.

@claude on the two items you flagged as unverifiable from this checkout — the EE files are symlinks into `windmill-ee-private`, so `git diff` here can't show them. Both are done:
- **Virtual-hosted endpoints (Codex's P1 last round)**: `render_endpoint` is no longer used for S3 list URLs. `s3_bucket_base_url` now builds the base explicitly, placing the bucket in the host for virtual-hosted and in the path for path-style, for endpoints with or without a scheme. `test_s3_bucket_base_url_places_the_bucket_in_every_endpoint_shape` covers all four combinations plus the double-port case (`endpoint: "localhost:9000"` + `port: 9000` no longer yields `host:port:port`). Path-style listing re-verified end to end against MinIO, since that is the shape the existing behaviour relied on.
- **`next_page_token` schema/wire agreement**: closed by the `skip_serializing_if` above rather than by declaring `nullable`.

diegoimbert:
Round 7 fixes pushed (`9a5a63a205` / EE `9a743dc`).

**P1 — provider cursors disclosing LIST-denied names.** Correct, and my count-based filesystem cursor only covered the one backend. S3, Azure and GCS all derive their continuation token from the last key they returned, and LIST filtering happens *after* the provider produced it, so a page ending on a denied entry handed the caller that name in plaintext base64.

Page tokens are now **sealed at the API boundary** with the workspace key (`build_crypt`/`encrypt`/`decrypt`, the same path variables and drafts use) rather than per-provider. `list_page` still deals in raw provider tokens; the handler seals on the way out and unseals on the way in. Verified on a live server: the token no longer decodes as base64 JSON, paging `a/` still yields 2000 unique entries across two pages, and a token replayed against a different prefix is still rejected (400).

@claude that also answers your three confirmations: the seal is keyed from `build_crypt(db, w_id)` — the workspace key in Postgres, i.e. shared instance state, not a per-process key, so "Load more" works across nodes; the inner `PageToken` prefix/provider binding is unchanged and still enforced after unsealing (demonstrated by the 400 above); and the canonical-prefix rule is now `validate_filesystem_prefix`, called only from the filesystem branch.

**P1 — delete during "Load more" dropping the first page.** Real: in-flight requests were keyed by prefix alone, so the post-delete refetch joined a pending append and returned believing page one had been fetched. Added a per-folder epoch, bumped when a single level is invalidated on its own; a joined result is only reused when both the listing generation and that level's epoch still match.

**P2 — non-hermetic prefix test.** Fixed by construction rather than by mocking: `max_keys == 0` short-circuits *after* any pre-dispatch validation would have run, so reaching the empty page proves the prefix was accepted without contacting a provider.

diegoimbert:
Round 12 fixes pushed (`1a841a90cf` / EE `04f1798`).

**P1 — empty-segment prefixes returning nothing.** Correct. A prefix like `a//` fails `Path::parse`, falls back to `list(None)`, and the loop then applied the seek-path stopping rule to a stream that had not seeked — so the first unrelated key (`0.txt`) ended the search immediately. The loop now distinguishes the two cases: a non-matching key that sorts *after* the prefix ends the range, one that sorts *before* it is skipped.

Note I could not reproduce this end to end: MinIO rejects empty-segment keys outright (`XMinioInvalidObjectName`), so the scenario is only reachable on AWS S3/GCS, which accept them. I verified the seek path is unaffected (`a/file199`, `analytics/review`, `Screenshot` all unchanged) and that unparseable prefixes now traverse rather than bail.

**P2 — stale served spec.** Regenerated via `backend/windmill-api/build_openapi.sh`. Both `openapi-deref.yaml` and `.json` now carry `list_stored_files_paged`, its schemas, and the `search` parameter. **Most of that diff is not mine** — several `openapi.yaml` commits since `bfc3f5242a` did not regenerate, so the artifacts had drifted ~14k lines. It is in its own commit (`1a841a90cf`) so it can be dropped or split out if you would rather not carry that drift here.

@claude yes to both EE items from last round, fixed in `aa0cd03`:
- **`marker` ignored in the search path** — the search branch built its own stream and never consumed `object_stream`, which was the only thing holding the marker, so every "Load more" replayed page one and returned the same cursor. It now resumes from the caller's cursor (always a key that already matched, so it sits inside the prefix range) and performs the exact-key `head` lookup only on page one. Verified against a live bucket: 6 pages, 1111 keys, all unique, each page starting at a distinct key (`a/file1`, `a/file1179`, `a/file1359`, …).
- **`search_next_marker` derived before LIST filtering** — sealed at the API boundary with `seal_page_token`/`unseal_page_token`, the same mechanism the paged endpoint uses, and only for search so the CLI and git-repo viewer keep their raw markers. Verified the returned cursor no longer decodes as base64 JSON.

On your contract point — a terminal search page can still carry a `next_marker` when the last match lands exactly on `max_keys`, because the range end is only observable by reading one key past it. The following request then returns empty with no cursor. Happy to spend a request to make the terminal page authoritative if you would rather have that.

diegoimbert:
Both P2s check out against the code as written — confirming rather than fixing, and taking the third suggestion.

**P2a (branch chosen from the marker, not the needle)** — not reachable, because the stopping rule is no longer path-dependent. Round 13 extracted it into a pure `scan_step(key, needle, resume_after)` used by *both* branches:

```rust
fn scan_step(key: &str, needle: &str, resume_after: Option<&str>) -> ScanStep {
    if let Some(after) = resume_after { if key <= after { return ScanStep::Skip } }
    if key.starts_with(needle) { ScanStep::Collect }
    else if key > needle { ScanStep::Stop }      // only a key sorting *past* the needle ends it
    else { ScanStep::Skip }                       // before the range: keep walking
}
```

So the case you describe — needle `a//` unparseable, marker `0.txt` parseable, so the resume routes into the seek path — seeks to `0.txt` and then still walks: `1.txt` sorts before `a//` → `Skip`, `a//file` → `Collect`, `b.txt` → `Stop`. The round-12 bug can't come back one page later, because "first key that doesn't start with the needle ends the range" isn't the rule on either path anymore. That was the point of making it one function rather than two loops.

**P2b (empty page must carry an advancing marker)** — not reachable either, for a different reason: there is no per-request scan budget. The fallback loop exits only on `Stop`, on stream end, or on `matches.len() >= max_keys`. An empty page therefore means the stream genuinely ran out or sorted past the range, both of which are terminal, and `last_key`-only-on-`Collect` is the right marker for that. Verified live against MinIO paging at `max_keys=2` — 3 full pages, then one empty terminal page, 6 keys, no repeats, no replay.

**Third note (bound the skip to the longest parseable ancestor)** — taken, `9699308`. This was the real one: `list(None)` meant page *n* re-walked everything before the range, quadratic across "Load more". Now `logs/2026/.` lists under `logs/2026` instead of the bucket root.

Worth recording why the fallback exists at all, since it's narrower than it looks. `Path::parse` rejects exactly three things: empty segments, `.`/`..` as a whole segment, and ASCII control chars. For the empty-segment case the fallback can never actually return a match — `object_store` parses listed keys through the same `Path::parse` (`client/s3.rs:80`, `azure/client.rs:1091`) and the failure fails the whole page, not one entry, so a key like `a//x` isn't listable at all, on this branch or on main. The case that genuinely works is the partial-segment one — searching `logs/2026/.` for dotfiles — which is also the case the ancestor bound speeds up. Verified end-to-end: `logs/2026/.` returns 2026's dotfile and never lists `logs/2025/`.

diegoimbert:
Fixed in `fd44da1`, and it reproduced exactly as described.

`loadFiles`' flat branch is now wrapped the same way the lazy branch above it already was. Two details worth noting, since neither is a plain `try/catch`:

- The `finally` is **generation-guarded**. An unguarded one would let a superseded listing clear the spinner of the listing that replaced it — the debounce means overlapping searches are the normal case here, not an edge case. `loadFlatFiles` returns early rather than throwing when it loses the race, so the guard is what keeps the two exits from colliding.
- `fileInfoLoading` is cleared in the `catch` only, **not** the `finally`. On the `restricted_access` exit the preview pane is owned by an in-flight `loadFileMetadataPlusPreviewAsync`, and clearing it there would flip "Loading..." to "No file to preview" underneath a metadata load that is still running.

**Verified in the browser both ways**, driving the picker against a live MinIO bucket with `page.route('**/list_stored_files*', abort)` installed after the drawer opens — so the lazy path loads normally and only the search fails, which is the shape you described:

| | spinners visible | toast |
|---|---|---|
| without the fix | **1** (permanent) | 0 |
| with the fix | 0 | 1 |

Clearing the filter recovers in both cases, as you said. The counterfactual run was done by stashing just this file, so the difference is attributable to the change and nothing else.

failed-search-toast

On the two carried-over items: full `npm run check` is now run — **0 errors**, 65 warnings, all pre-existing and none in the touched files. The git-repo-viewer flat path and the storage-switch repro are still unrun; they're the two regression paths on `lazyFolders={false}`, and I'd rather they be exercised against a real git repo than simulated.

diegoimbert:
Good catch, and correctly attributed — this is only reachable *because* of the previous commit. Before it, a failed page was a permanent spinner, so there was no retry to get wrong; making it retryable is what exposed the stranded cursor. Fixed in `6f4bd77`.

Fixed at the increment site rather than in the catch, because `loadFiles` can't tell an initial load from a "Load more" and would have no idea what to roll back to:

```svelte
on:click={() => loadNextFlatPage()}
```

```ts
async function loadNextFlatPage() {
    page += 1
    const requested = page
    const ok = await loadFiles()
    // Only undo our own advance: another click may have moved it on meanwhile.
    if (!ok && page === requested) page = requested - 1
}
```

`loadFiles` now returns whether the listing completed. The `page === requested` check matters because neither flat button has a `!loadingMore` guard (that one is on the per-folder row), so two quick clicks can overlap and a blind `page -= 1` would undo the wrong one.

**Verified with exactly the repro you specified** — page one, abort the next flat request, then retry — reading the `marker` query param off the wire on each request:

| | after page 1 | failed page 2 | after retry | markers sent |
|---|---|---|---|---|
| without the fix | 1000+ | 1000+ | **1000+** (stuck) | `(none)`, `CmZ2Tv…`, **`(none)`** |
| with the fix | 1000+ | 1000+ | **2000+** | `(none)`, `CmZ2Tv…`, **`CmZ2Tv…`** |

The third marker is the whole finding: without the rollback the retry sends no marker and replays page one; with it, the failed page's cursor is reused and the keys that come back are the later ones. Counterfactual run by stashing just this file. I also re-ran the previous round's failed-search check against the refactor — still 0 spinners, 1 toast.

page2-retried

Still unrun, and I'd rather say so than simulate them: the git-repo viewer and the app image picker on `lazyFolders={false}`. Both want a real git repo / app rather than an injected listing.

diegoimbert:
Both fixed in `a01a8cb`.

**Codex P1 — stale rollback across generations.** Real, and reproduced. `loadNextFlatPage` now captures `listingGeneration` and requires it unchanged before rolling back:

```ts
const generation = listingGeneration
page += 1
const requested = page
const ok = await loadFiles()
if (!ok && generation === listingGeneration && page === requested) page = requested - 1
```

Verified with the sequence you specified — hold search A's page two open with a Playwright route, switch the filter to B (which bumps the generation and resets `page`), advance B to the same page number, *then* abort A. Because the sealed markers are deterministic, a replay is visible directly on the wire: I check whether B's post-abort request repeats a marker B already used.

| | B's markers before the abort | B's request after | replayed |
|---|---|---|---|
| without the guard | `(none)`, `HknrbnlO7H…` | `HknrbnlO7H…` | **yes** |
| with the guard | `(none)`, `HknrbnlO7H…` | `kwa9dWKtDO…` | no |

To make the race observable at all I had to shrink `maxKeys` to 50 locally — the test bucket only has ~1100 keys, so at 1000/page B ran out of pages before the interesting step and the last click issued no request. That change was local to the run; `maxKeys` is back to 1000 in the commit.

**Claude P2 — the delete replay loop.** Also real and mirrored: `deleteFileFromS3` re-drives `page = i + 1` by hand, and now that `loadFiles` returns `false` rather than throwing, a mid-replay failure let the loop keep going and exit with `page` permanently ahead of `listMarkers`. It now stops at the last page that actually loaded:

```ts
if (!(await loadFiles())) { page = i; break }
```

Being straight about coverage: **this one I have not exercised.** It needs a confirmation-modal delete while paged past page 0 *and* a failure partway through the replay, on a picker mounted with `allowDelete`, which the script-editor entry point isn't. The reasoning is the same invariant the other two fixes turn on (`listMarkers.length === page + 1`), but reasoning is exactly what has been wrong three rounds running here, so treat it as unverified rather than as verified-by-argument.

Worth noting the shape of the last three rounds: each found a defect introduced by the previous round's fix, all in this same error-handling arm — round 14's spinner fix made failures retryable, which exposed round 16's stranded cursor, whose fix exposed this round's cross-generation rollback. That's a sign the arm was genuinely unexercised, not that it's oscillating; each fix is narrower than the last and this round's two are now both pinned by the invariant rather than by a case analysis.

diegoimbert:
Both of you found the same thing and you were right — fixed in `0cdb5877`. `clearAndLoadFiles` now returns whether the fresh listing loaded, and the replay loop runs only if it did.

**This one I did exercise**, unlike last round's half of it. It needed three local-only patches to reach: the path is dead in a `lazyFolders={true}` picker (the early return takes it before `clearAndLoadFiles` clears the filter), so I temporarily gave the script editor's picker `lazyFolders={false} allowDelete` to stand in for the app image picker, and dropped `maxKeys` to 50 so ~1100 keys span enough pages. None of that is in the commit.

Sequence: page to page 2, expand the folder, select a file, then delete it with a route that fails **only** the first listing after the delete.

| | requests during the post-delete refresh | "Load more" after | later clicks |
|---|---|---|---|
| without the fix | `(none)`, `(none)`, `(none)` | **visible** | `(none)`, `(none)` — replaying page one |
| with the fix | `(none)` | hidden | — |

The three requests in the first row are the finding exactly: the failed seed, then two replay pages that each advanced `page` against an empty `listMarkers`, sending no marker. `listMarkers.length == page` then never holds again, so the cursor never recovers and every later click refetches page one. With the fix the loop is skipped, and `flatHasMore` stays false so the button isn't offered on a listing that isn't there — the state after a failed seed is just "nothing loaded", which is recoverable.

Worth noting what this says about the sequence: rounds 14→18 each found a defect created by the previous fix, and every one of them was a different way for `page` and `listMarkers` to disagree. They're now all expressions of one rule — `page` may only advance over a listing that actually landed — enforced at each of the three places that move it (`loadNextFlatPage`, the replay loop, and now the seed). I don't think there's a fourth site; `page` is assigned in exactly those three places plus the reset in `clearAndLoadFiles`.

Unchanged from last round: the git-repo viewer and the app image picker still haven't been exercised as themselves. The delete path above is the closest proxy I could build for the image picker, but it is a proxy — the real one carries a `regexFilter`, which I did not simulate.

diegoimbert:
One of the two P1s is real and is fixed in `a0e7143`. The other rests on a premise that doesn't hold — details below, with the check anyone can repeat.

## Codex P1 — superseded lazy load writes into the search that replaced it. Real, fixed.

Correct diagnosis: `loadFolderPage` resolves rather than throwing once its generation goes stale, so the lazy branch of `loadFiles` kept running after a filter change switched the picker to flat. Both halves now carry the guard the flat branch already had:

```ts
const generation = listingGeneration
...
if (generation === listingGeneration && selectedFileKey !== undefined && !emptyString(selectedFileKey.s3)) {
    await expandToKey(selectedFileKey.s3)
}
...
} finally {
    if (generation === listingGeneration) { fileListLoading = false; fileInfoLoading = false }
}
```

Exercised as far as this entry point allows: held the lazy root listing open with a route, typed a filter to supersede it, then released it. No browse entries leaked into the results, the search's own results stood, and no spinner was left behind. Being precise about coverage — that exercises the `finally` half. The `expandToKey` half needs a **preselected** file, which the script-editor picker never has, so that half is guarded-by-symmetry rather than reproduced.

## Pi P1 — `computeVisibleKeys` hides nested files. Not a regression; the premise is wrong.

The finding rests on "The old code showed **all** entries at once in flat mode." It didn't. On `origin/main`, same file:

```
$ git show origin/main:frontend/src/lib/components/S3FilePickerInner.svelte | sed -n '242p;247,249p'
                    collapsed: true, // folders collapsed by default
                if (i == rootPathNestingLevel && current_path.startsWith(rootPath)) {
                    displayedFileKeys.push(current_path)
                }
```

Folders already defaulted to `collapsed: true`, and only entries at `i == rootPathNestingLevel` were ever pushed — **root level only**. Nested files were never listed up front on main either; you always had to expand. `computeVisibleKeys` shows every entry whose ancestors are all expanded, which is a *superset* of what main displayed (root level, plus children of folders opened by the toggle, plus the preselected file's ancestors). It exists precisely because main's accumulate-on-push approach missed later pages' keys inside already-expanded folders — the bug reported on this PR.

Confirmed visually while testing the delete path with `lazyFolders={false}`: the picker opened showing exactly three root rows (`'dfsjjf (1 item)`, `Screenshot 2025-09-10…`, `a (48 items)`) and the files appeared on clicking `a (48 items)` — which is main's behavior.

The `lazyFolders` doc could read better, though: "listing every key up front" describes the *request* shape (one flat listing vs. per-level paged), not the collapse state, and it clearly invites this reading. Happy to reword it if you'd like — I left it alone rather than churn the prop doc on a disputed finding.

diegoimbert:
Fixed in `8c98c21` — you're both right that one check at the entry isn't enough. The generation is now threaded through `expandToKey` and `revealChild` and re-checked after every await, so a walk that loses its listing stops instead of finishing under whatever replaced it.

I built a repro for this one rather than reasoning about it: local-only, the script editor's picker opens with a preselected nested key (it normally has none), and every per-level request is delayed 1.8s so the walk is still going when the filter lands. Typing at 2.9s puts the supersede *after* the root load and *inside* the deep reveal — 2 paged requests instead of 1 confirms the walk actually got past the entry check.

**Confirmed:** without the mid-walk checks the stale walk issues **6** paged requests instead of **2**. It keeps loading levels for a listing nobody is looking at, and those loads date themselves to the replacement generation, so `stillCurrent()` inside `loadFolderPage` passes and they write.

**Did not reproduce:** the visible consequence you both predicted — browse entries grafted into the search results via the trailing `refreshDisplayed()`. With the checks removed I waited 25s, well past the walk's end, and the rendered rows were still exactly `"Screenshot 2025-09-10 at 17.53.21.png"` — the search's own result, nothing else. `refreshDisplayed` is unguarded and does overwrite `displayedFileKeys` from `allFilesByKey`, so I can't explain from the code why it stayed clean, and I'd rather report that than dress it up. What I can say is that the wasted round trips are real and measured, and that a walk continuing to mutate state for a discarded listing is worth stopping on its own terms.

So: fix landed, mechanism confirmed, stated impact only partly confirmed.

Small note on scale — that's five rounds where a fix exposed the next defect, and the last two were found by two reviewers independently, which suggests the area is now getting real scrutiny rather than that it keeps breaking. The generation rule is applied at every await in the lazy path now (`loadFolderPage`'s own guards, the `loadFiles` branch, `expandToKey`, `revealChild`), and the cursor rule at every site that moves `page`.

diegoimbert:
Fixed in `c245a21`, and it's mine — that `catch` isn't on `origin/main`, I added it earlier in this PR to stop a deleted key leaving the pane on "Loading..." forever. It cleared shared state without checking that its request still owned the pane.

```ts
console.error('Error loading metadata for', fileKey, e)
// Unless the pane has moved on: selections overlap, and a late failure for the
// previous file would blank the preview of the one now selected.
if (selectedFileKey?.s3 !== fileKey) return
fileMetadata = undefined
...
```

**Reproduced exactly as you described**: hold file A's `load_file_metadata` open with a route, select file B (which completes), then fail A.

| | pane blanked to placeholder while B is selected |
|---|---|
| without the guard | **true** |
| with the guard | false |

One thing your finding didn't mention that's worth recording: the **success** path has the same shape — it assigns `fileMetadata = { fileKey, ... }` with no ownership check, so a slow A resolving after B overwrites B's metadata with A's. That one is pre-existing (it's on `origin/main` unchanged), and it's a different fix, so I've left it alone rather than widen this PR. Flagging it so it isn't mistaken for something this branch introduced.

diegoimbert:
You found the same root cause from two directions, and it was the right one: I compared a **value** where I needed **request identity**. `d81a0d4` replaces the key comparison with a sequence number.

```ts
let metadataRequestId = 0
...
const requestId = ++metadataRequestId
...
} catch (e) {
    if (requestId !== metadataRequestId) return   // a later request owns the pane
    fileMetadata = undefined; filePreview = undefined; fileInfoLoading = false
}
```

This is one guard for both findings, which is what makes me think it's the right level:

- **Codex** — a same-key reload (storage switch calling `reloadContent`) now bumps the id, so the older request no longer speaks for the newer one. Key equality could never have distinguished those two.
- **Claude** — a request whose selection has moved to something that fetches no metadata is still the newest, so it clears `fileInfoLoading` instead of returning early. The key comparison got this backwards precisely because the flag is owned by the request while the key is owned by the selection — as you put it.

Verified both properties in the browser:

| | result |
|---|---|
| lone failing request (Claude's case, essence) — stuck on "Loading..." | **false** |
| hold A, select B, fail A — B's pane blanked | **false** |

Previous round's repro still passes, so the different-key race stays closed.

Still standing from my last comment, unchanged: the **success** path has the same shape (`fileMetadata = { fileKey, ... }` with no ownership check), it's pre-existing on `origin/main`, and I've deliberately left it out of this PR. With `metadataRequestId` now in place it would be a one-line follow-up if you want it — say the word and I'll do it here rather than leave it lying around, but I'm not widening the diff on my own initiative this late.

diegoimbert:
Real, and mine — fixed in `e6dfbc0`. The lazy branch refetches only the affected level and returns, so it never reached the reset the flat path gets for free from `clearAndLoadFiles`. Cleared at the one place the selection is cleared, so both branches are covered:

```ts
selectedFileKey = { s3: '', storage }
// The preview pane and its toolbar render from these rather than from the
// selection, so the deleted file stays previewed — and offers to download, move
// and delete itself — until they are cleared too.
fileMetadata = undefined
filePreview = undefined
```

Reproduced with a real delete (local-only `allowDelete` on the editor's picker, since that entry point doesn't offer deletion):

| | after deleting the previewed file |
|---|---|
| without the fix | preview + **Download / Move / Delete** toolbar still there for the deleted key |
| with the fix | pane falls back to "Select a file to preview" |

Your "offers to delete itself" detail is right and is the part that makes it more than cosmetic — the toolbar stays live against a key that no longer exists.

Pi and Claude both returned **Good to merge** on `d81a0d4` this round, so this is the only outstanding item as far as I can tell. Fixture file created for the test was deleted afterward.

diegoimbert:
Right again, and closing it took the follow-up I'd offered to leave out — so I've done it here rather than leave a known hole. `8a7913a`:

1. deleting bumps `metadataRequestId`, retiring whatever is in flight for that file;
2. the **success** path now checks the id (this is the pre-existing race I flagged two rounds ago — without it the bump does nothing, since only the failure path was guarded);
3. `loadFilePreview` takes the id too, so a preview trailing behind a retired metadata load can't write either.

Getting the repro honest took two attempts, worth recording because the first one looked like a pass:

- **First attempt, wrong**: hold the metadata request, delete, then release it. It came back **404 — the file is gone** — so it took the *catch* path, which clears. Result: "not reproduced", which would have been a false negative.
- **Correct**: the race is a response *computed before* the DELETE landed. Fulfilling the held request with a synthetic 200 body reproduces it exactly.

| | preview resurrected by the late response |
|---|---|
| without the retirement | **true** — pane and Download/Move/Delete toolbar back, for a key that no longer exists |
| with it | false — pane stays on the placeholder |

Codex and Pi both returned **Good to merge** this round, so this was the last open item I know of. Test fixture deleted afterward; the `allowDelete` patch used to reach the delete path was local only and is not in the commit.

diegoimbert:
Both of you again, same follow-on, and correct: retiring the request left nobody to report its outcome, so the flag stayed up. `861a8f3` clears it at the delete, which now owns it.

Verified with Codex's sequence — A previewed, B's metadata held open (A's toolbar is still live at that point, which is what makes the delete reachable), then delete A:

| | pane stuck on "Loading..." |
|---|---|
| without the clear | **true** |
| with it | false |

Note this is the fourth consecutive round on the same handful of lines, each one a consequence of the previous fix: clearing the pane → the response repopulated it → retiring the request → nobody cleared the flag. They're the three things a request owns (its data, its right to write, its loading flag), and the delete now takes all three at once rather than one per round. I don't see a fourth thing it owns, but I've said that before in this thread, so treat it as an expectation rather than a guarantee.

Pi was **Good to merge** this round; Codex's P1 and Claude's nit were the same item, now fixed. Fixtures deleted, `allowDelete` patch was local only.

diegoimbert:
Valid, and fixed in `c0623c4` — with one correction to the diagnosis that changes what needed doing.

The two artifacts are not in the same situation:

- **`openapi-deref.yaml`** — `Dockerfile:74` runs `build_openapi.sh` and `:117` copies the regenerated yaml into the build context *before* the Rust build, so the yaml baked into a shipped image is always fresh. The committed copy only affects a local `cargo build`.
- **`openapi-deref.json`** — regenerated in that same builder stage but **never copied**, so `include_str!` at `lib.rs:1244` embeds the committed file verbatim. That one is genuinely served stale, in every build, which is exactly the Scalar problem you describe.

So the fix is the json, and I spliced the operation in rather than regenerating: a full regeneration re-imports ~7k lines of drift that already exists between the committed artifact and the spec it derives from (I measured it — regenerating from *main's own* `openapi.yaml` differs from main's committed deref by 9,758 lines). That drift is what made this diff 17k lines in the first place.

Result: **+148 lines, no deletions.** Python round-trips the generator's output byte-for-byte (2-space indent, unescaped UTF-8), so the splice is textually indistinguishable from a regenerated file in the region it touches.

Your finding also caught something I'd have shipped broken. My first splice copied only the two `paths` entries, and `redocly lint` went from 4 errors to **6** — the response schema `$ref`s `#/components/schemas/StorageFolder` and `StorageFile`, which weren't in the committed `components`. Added both; back to 4 errors, all pre-existing (`operation-operationId-url-safe` on unrelated endpoints), with the per-rule breakdown unchanged and one extra warning of the same class the spec already trips several hundred times.

I could not run `openapi-generator-cli validate` locally — no Java here, it crashes during its own bootstrap — so CI's check is the one that counts. The refs now resolve, which is what it was failing on.

diegoimbert:
Fixed in EE `eb1a765` (ref bumped in `1d48613`). Confirmed the finding, and confirmed something more important while verifying it.

**The fix.** Both the root and the requested directory are canonicalized and compared, and an entry that resolves outside the root is skipped instead of followed. Symlinks that stay *inside* the root keep working — rejecting all symlinks would break legitimate layouts, and the security-relevant case is specifically the ones that leave.

Verified end-to-end against a real `FilesystemStorage` workspace with `escape -> /outside` and `alias -> ./inside` under the root:

| request | before | after |
|---|---|---|
| root level | `escape`, `alias`, `inside` | `alias`, `inside` |
| `prefix=escape/` | enumerated the target | `Bad request: Invalid prefix` |
| `prefix=alias/` | listed `ok.txt` | listed `ok.txt` (unchanged) |

Regression test added, and I checked it fails without the guards rather than trusting that it passes — it does, on the "escaping link listed" assertion.

**What you should know: this does not close the hole.** While verifying, I checked the pre-existing endpoints against the same root, and on `main` today:

```
GET list_stored_files          -> [... {"s3":"escape/secret.txt"} ...]
GET load_file_preview?file_key=escape/secret.txt  -> {"content":"secret\n", ...}
GET download_s3_file?file_key=escape/secret.txt   -> secret
```

The cause is upstream and deliberate: `object_store`'s `LocalFileSystem` sets `.follow_links(true)` (`local.rs:507,644`) and documents *"Symlinks that resolve to paths outside the root **will** be followed"*. So the escape predates this branch, and reading file *contents* through it — which is the part that actually matters — never went near my code.

I've deliberately not fixed that here. It belongs in the shared filesystem-client layer so listing, preview and download are covered together, it touches pre-existing behaviour several endpoints depend on, and a security fix of that shape deserves its own review rather than being appended to a 26-round feature PR. Happy to open it immediately as a separate PR — say the word.

Severity note for triage: `FilesystemStorage` is hidden from the settings UI and set via API only, and exploiting it requires write access to the storage root to plant the link. That lowers the practical urgency but doesn't change that the boundary isn't enforced.

PR #10423, opened 2026-07-30
fix: credit the token owner instead of the token label in the audit trail

rubenfiszel:
## Summary

When a job is launched with a user API token — including the tokens minted by the MCP OAuth flow — `ApiAuthed::display_username()` returned the token **label** rather than the token owner. That value is written to `v2_job.created_by` and to the audit row's `parameters.end_user`.

A token label is arbitrary user-controlled text that identifies no principal. For MCP OAuth tokens it is auto-generated as `mcp-oauth-<client_id>`, so it names an opaque OAuth client rather than a person.

This becomes an outright traceability hole when the target runnable has `on_behalf_of_email` set: the audit `username` **and** `email` columns are then taken by the on-behalf-of identity, and `end_user` holds only the label — so the human who actually made the request appears in **no column of the audit row at all**.

## Changes

- `username_override_from_label` now also returns whether the override names the token's *label* rather than the entity that acted, and `ApiAuthed` carries it as `username_override_is_token_label`. `display_username()` skips the override in that case, crediting the token owner.
- Provenance is carried, never re-derived from the value: the `ephemeral-script-end-user-` arm forwards a `created_by` verbatim, and `created_by` is an unconstrained varchar that may itself look like any of these shapes.
- **Only namespaces the server alone mints are trusted to name the acting entity** (`is_server_minted_label`), and `create_token` now rejects exactly those. Everything else is credited to the token owner — `webhook-`/`http-`/`ws-`, `email-*` and `Ephemeral lsp token` are all mintable through the public API and no mint produces them, so none of them names a principal. Their override *value* is unchanged, so `require_job_read_access` still matches it against the `created_by` of jobs launched while those shapes were trusted.
  - `email-*` in particular carries no trigger identity: the SMTP trigger builds its `email-*` override **server-side** (directly, on both the configured-trigger and tokened-address paths), so it never depends on the label round-trip. The worker's CE email-trigger quota gate keys on the resulting `created_by`, which those direct overrides still produce.
  - The guard is deliberately **narrower** than the `is_user_token` predicate used for relabelling: the editor mints `Ephemeral lsp token` on every TypeScript editor load and the debugger mints `debugger-token`, both through this handler, so reserving those would break both features.
- `set_acting_username_override` keeps the pair in sync, so assigning the override alone cannot leave a stale flag (the SMTP trigger path does exactly this mutation, and a stale flag there would also drop the job out of the CE email-trigger quota gate).
- `$ctx:username` in the app safe-args substitution goes through `display_username()` instead of re-implementing it.
- Audit logs table: the identity cell was a fixed-width horizontal scroll box, so a long override could scroll the username out of view entirely. The username is now pinned and the override truncates, both with `title` tooltips.
- Audit log details pane now shows the caller's token prefix (the `span` column), labelled `Job` for job-minted worker tokens. It was already recorded and API-returned but rendered nowhere.

## Known gap (pre-existing, reduced but not closed)

Trust is still derived from the label's *shape*, so **tokens minted before the new guard** keep a reserved label that still displaces the owner. There is no sound retroactive reclassification — legitimate server-minted rows are indistinguishable by shape, and native-trigger tokens without an expiration are long-lived, so relabelling wholesale would break live webhooks. Closing it needs trust grounded in **issuance** (a `token` column set at the mint helpers): tracked in #10426.

Net effect of this PR on that surface: from *every* label displacing the owner, down to that one case.

## Behaviour change to be aware of

`webhook-`/`http-` are what the webhook and HTTP-route panels pre-fill, so anyone who set a trigger up through the UI moves from `created_by = webhook-alice-x9k2` to `created_by = alice`. That is the intended correction — those tokens are user-minted and the owner is the accurate answer — but it is the widest-reaching part of this PR.

Separately, in the plain (non-on-behalf-of) case `created_by == permissioned_as`, so `push` emits no `end_user` and the row no longer distinguishes a token-driven operation from a browser session. The token is still identified by the audit `span` column (the token prefix, not user-controlled) — nothing in the audit UI rendered it, so this PR surfaces it in the details pane. On the Runs page there is no token indicator either way.

## Before / after

Six identical runs before and after: a generic labelled token, an MCP-OAuth token, and a `webhook-`-shaped token, each against a plain and an `on_behalf_of` script.

| # | token | script | shows |
|---|---|---|---|
| 16 | generic label | `on_behalf_of` | `svcacct (label-my-perso…)` — **caller absent** |
| 18 | mcp-oauth | `on_behalf_of` | `svcacct (label-mcp-oaut…)` — **caller absent** |
| 51 | generic label | `on_behalf_of` | `svcacct (alice)` |
| 53 | mcp-oauth | `on_behalf_of` | `svcacct (alice)` |
| 55 | `webhook-alice-x9k2` | `on_behalf_of` | `svcacct (alice)` |
| 50/52/54 | all three | plain | `alice` |

Row `14` is the control: a browser session against the same `on_behalf_of` script already recorded `svcacct (alice)` correctly, which is what shows the mechanism was right and only `display_username()` was wrong.

audit before and after

## Test plan

- [x] `cargo test -p windmill-api-auth --lib` — 39 passed. The new unit test drives the real `username_override_from_label` and pins each label class: generic and MCP labels → owner, the user-mintable `webhook-*` and `Ephemeral lsp token` → owner, the server-minted `ephemeral-webhook-*` → itself, and the passthrough collision `ephemeral-script-end-user-label-alice` → `label-alice` (not the owner). Verified to fail pre-fix.
- [x] `cargo test -p windmill-api-integration-tests --test token_label_idor` — 4 passed. `test_reserved_token_labels_not_creatable` pins the create guard; `test_trigger_token_labels_still_creatable` pins that the trigger-panel labels **and** the editor/debugger labels stay mintable; `test_legit_labeled_self_read_still_works` pins that a `label-*` `created_by` stays readable by its owner.
- [x] `cargo check --features enterprise,private,mcp,smtp --all-targets` — clean (the EE SMTP path needs `smtp`).
- [x] End-to-end on a local EE build: the six runs above; `POST /users/tokens/create` returning 400 for `ephemeral-webhook-forged` / `ephemeral-script-end-user-svcacct` / `session` / `mcp-oauth-forged` and 201 for `Ephemeral lsp token` / `debugger-token` / `webhook-legit` / `my-token`.
- [x] Opened a TypeScript script editor in the browser and confirmed the LSP token mints and type acquisition completes (`SETUP TYPE ACQUISITION` → `ATA done`), i.e. the create guard does not break language-server setup.
- [ ] Reviewer: confirm the `webhook-`/`http-` behaviour change is wanted, and whether the issuance-based follow-up should land here or separately.

Fixes WIN-2270


rubenfiszel:
Re: the persisted-reserved-label finding — deferring it deliberately, tracked in #10426.

The residual is real and I've documented it precisely in the PR description. Two things shape the call:

- It is **pre-existing and strictly reduced** by this PR, not introduced by it. Before this change *every* label displaced the owner; after it, only tokens minted before the new guard plus `email-*` do. Nothing gets worse.
- Closing it soundly requires grounding trust in issuance (a `token` column set at the mint helpers), because there is no safe retroactive reclassification: legitimate server-minted rows are indistinguishable by shape, and native-trigger tokens without an expiration are long-lived, so relabelling wholesale would break live webhooks. That is a migration plus every mint site plus JWT-claim handling — larger than this audit fix and with its own rolling-deploy surface, which is why it is a separate PR.

The `email-*` half cannot be closed by reserving the namespace at all: the trigger panel mints it client-side and the worker's CE email-trigger quota gate keys on `created_by.starts_with("email-")`.

Also fixed this round, from the other reviewer: the previous guard delegated to `is_user_token` wholesale, which 400'd the editor's `Ephemeral lsp token` (minted on every TypeScript editor load) and the debugger's `debugger-token`. The guard is now narrowed to `is_server_minted_label`, those two labels moved to the user-mintable/untrusted branch, and the test that had locked the regression in now asserts they stay creatable. Verified in the browser that type acquisition completes.

rubenfiszel:
✅ Review round clean @ fb4b64af88cfc91e440fb66461972e593814106c

rubenfiszel:
✅ Review round clean @ 346518eb991044b56940489dc22123428daf1d32

PR #10427, opened 2026-07-31
fix: apply default workspace dependencies to raw app runnables

rubenfiszel:
## Summary

`wmill generate-metadata` regenerated raw-app runnable locks without the default
Bun workspace manifest, so a pinned `windmill-client` drifted back to `latest`
on every run. Standalone Bun scripts and flow inline scripts with the same
import were unaffected.

Raw apps keep their runnables as files under `backend/`; `raw_app.yaml` only
carries a `runnables` map in the legacy layout. `generateAppLocksInternal` read
`appValue` straight from `raw_app.yaml` and fed that to
`filterWorkspaceDependenciesForApp`, so a file-based raw app filtered its
dependencies against an empty runnables map — the result was `{}` and the
dependency job ran with `raw_workspace_dependencies: null`.

### Regression window

Not a longstanding gap — this worked until v1.639.0:

- v1.590.0 (#7310) moved raw-app runnables into `backend/` files.
- v1.639.0 (#7993) introduced `filterWorkspaceDependenciesForApp` as a gate in
  front of lock generation. Before it, the raw workspace-dependency map was
  passed straight through to `updateRawAppRunnables`, so the on-disk layout was
  irrelevant and the pin was honored.

Raw apps in the legacy layout (runnables inline in `raw_app.yaml`) and normal
apps were never affected, which is why the feature still looks healthy in those
cases.

### Scope note

This grew past the original two-file fix during review. Chasing the root cause
surfaced a second, related defect — `loadRunnablesFromBackend` hardcoded `bun`
for a bare `.ts`, disagreeing with the assigner that wrote the file — and the
remaining call sites are fixed here rather than deferred.

## Changes

**The lock fix**

- `app_metadata.ts`: resolve raw-app runnables from `backend/` (falling back to
  `raw_app.yaml`) before workspace-dependency filtering. The two later, correct
  resolutions — the tree dry-run pass and the lock-generation branch — reuse
  that single load; having three independent answers to "where are the
  runnables", one of them wrong, is what produced this bug.

**The `defaultTs` round-trip**

`getLanguageExtension` writes a bare `.ts` only when the script's language
equals `defaultTs`, so in a `defaultTs: deno` workspace a bun runnable is
`test.bun.ts` and a plain `test.ts` means deno. Every reader hardcoded `bun`,
which relabeled the script and made the writer emit a duplicate `test.bun.ts`
beside `test.ts`. `defaultTs` is now threaded from each command entry, matching
how the pull side already feeds `newRawAppPathAssigner`:

- `sync.ts` → `pushObj` → `pushRawApp` (carries the effective, override-resolved
  `opts.defaultTs`)
- `app.ts` push and `pushRawAppCommand` → `pushRawApp`
- `dev.ts`: resolved once at startup, passed to `loadRunnablesFromBackend` (×3),
  `inferAllInlineSchemas`, and `inferRunnableSchemaFromFile`
- `lint.ts` left alone: it reads only `Object.keys(...).length`, which is
  language-independent

**Refactor**

- `pushObj`'s six trailing optional positionals collapse into a
  `PushObjOptions` object, removing a four-`undefined` call and naming
  `originalLocalPath` at the two sites that were passing it positionally.

**Test**

- `raw_app_workspace_deps_unit.test.ts`: pins that a file-based raw app's
  `backend/` runnable picks up `dependencies/package.json`.

## Test plan

- [x] `bun test test/raw_app_workspace_deps_unit.test.ts` — fails on `main`
      (app reported up-to-date after the default manifest appears), passes here.
- [x] `UNIT_ONLY=1 bun test test/*_unit*.test.ts` — 991 pass, 0 fail (re-run at
      `bec4de186a`).
- [x] `bunx tsc --noEmit` — no new errors; the two `sync.ts` JSZip errors are
      pre-existing on `main`.
- [x] End-to-end against a local backend + worker, `dependencies/package.json`
      pinning `windmill-client`, raw app runnable at
      `f/example.raw_app/backend/test.ts` importing it:
  - before: lock declares `"windmill-client": "latest"`, resolves `1.775.2`
  - after: lock declares the pinned version — byte-identical to the lock of a
    standalone Bun script with the same import
  - bumping the pin re-marks the app stale and regenerates; a second run reports
    `All metadata up-to-date`
- [x] Legacy raw app (runnables inline in `raw_app.yaml`) and normal app with an
      inline Bun script both still pick up the pinned dependency.
- [x] `defaultTs: deno` workspace: `test.bun.ts` → bun, pinned lock, no
      duplicate file, idempotent; plain `test.ts` → deno, no `package.json`, no
      duplicate. Baseline `031f17d665` produced
      `test.ts + test.bun.ts + test.lock + test.yaml` for the same input.
- [x] Full `wmill sync push` through `pushObj → pushRawApp` after the
      options-object refactor; server confirms the runnable landed as
      `type=inline language=bun` with the pin intact.

## Known deferred

`wmill app push`, `app push-raw` and `app dev` read top-level `defaultTs` via
`mergeConfigWithConfigFile`, which does not apply `workspaces[*].overrides`,
so a per-workspace `defaultTs` override is not honored by those three commands.
This is the existing CLI-wide convention — `wmill script push` resolves it the
same way and already mislabels a bare `.ts` standalone script in that config on
`main` — and `sync push` (the path that has to invert what pull wrote) is
correct, since it threads the effective value. Fixing only the two app commands
would leave `script push` inconsistent, so it belongs in a follow-up that
changes the convention everywhere.

Fixes GIT-939

### DIFF
--- a/cli/src/commands/app/app.ts
+++ b/cli/src/commands/app/app.ts
@@ -12,7 +12,7 @@ import * as wmill from "../../../gen/services.gen.ts";
 import { ListableApp, Policy } from "../../../gen/types.gen.ts";
 
 import { GlobalOptions, isSuperset } from "../../types.ts";
-import { getWmillYamlPath } from "../../core/conf.ts";
+import { getWmillYamlPath, mergeConfigWithConfigFile } from "../../core/conf.ts";
 import { readInlinePathSync } from "../../utils/utils.ts";
 import devCommand from "./dev.ts";
 import lintCommand from "./lint.ts";
@@ -402,7 +402,14 @@ async function push(
 
   if (isRawAppByName || hasRawAppYaml) {
     const { pushRawApp } = await import("./raw_apps.ts");
-    await pushRawApp(workspace.workspaceId, remotePath, absoluteFilePath);
+    const merged = await mergeConfigWithConfigFile(opts);
+    await pushRawApp(
+      workspace.workspaceId,
+      remotePath,
+      absoluteFilePath,
+      undefined,
+      merged.defaultTs,
+    );
     log.info(colors.bold.underline.green("Raw app pushed"));
   } else {
     await pushApp(workspace.workspaceId, remotePath, absoluteFilePath);
--- a/cli/src/commands/app/app_metadata.ts
+++ b/cli/src/commands/app/app_metadata.ts
@@ -41,7 +41,7 @@ import {
   newRawAppPathAssigner,
   SupportedLanguage,
 } from "../../../windmill-utils-internal/src/path-utils/path-assigner.ts";
-import { mergeConfigWithConfigFile, SyncOptions } from "../../core/conf.ts";
+import { getDefaultTs, mergeConfigWithConfigFile, SyncOptions } from "../../core/conf.ts";
 import { resolveWorkspace } from "../../core/context.ts";
 import { requireLogin } from "../../core/auth.ts";
 import { getNonDottedPaths } from "../../utils/resource_folders.ts";
@@ -471,10 +471,11 @@ async function updateRawAppRunnables(
   remotePath: string,
   appFolder: string,
   rawDeps?: Record<string, string>,
-  defaultTs: "bun" | "deno" = "bun",
+  defaultTs?: "bun" | "deno",
   noStaleMessage?: boolean,
   tempScriptRefs?: Record<string, string>
 ): Promise<string[]> {
+  const ts = defaultTs ?? (await getDefaultTs());
   const updatedRunnables: string[] = [];
   const runnablesFolder = path.join(appFolder, APP_BACKEND_FOLDER);
 
@@ -485,7 +486,7 @@ async function updateRawAppRunnables(
     // Folder may already exist
   }
 
-  const pathAssigner = newRawAppPathAssigner(defaultTs);
+  const pathAssigner = newRawAppPathAssigner(ts);
 
   for (const [runnableId, runnable] of Object.entries(runnables)) {
     // Only process inline scripts (runnableByName/inline with inlineScript)
@@ -845,7 +846,7 @@ export interface InferredSchemaResult {
 export async function inferRunnableSchemaFromFile(
   appFolder: string,
   runnableFilePath: string,
-  defaultTs: "bun" | "deno" = "bun",
+  defaultTs?: "bun" | "deno",
 ): Promise<InferredSchemaResult | undefined> {
   // Extract runnable ID from file path (e.g., "myRunnable.ts" -> "myRunnable",
   // "fetch_data.bun.ts" -> "fetch_data")
@@ -916,7 +917,7 @@ export async function inferRunnableSchemaFromFile(
     runnable?.type === "inline" ||
     runnable?.type === "runnableByName"
   ) {
-    language = getLanguageFromExtension(ext, defaultTs);
+    language = getLanguageFromExtension(ext, defaultTs ?? (await getDefaultTs()));
   } else {
     // Path-based runnable or unknown type - schema lives at the script/flow path
     return undefined;
@@ -981,8 +982,9 @@ export async function inferRunnableSchemaFromFile(
  */
 export async function inferAllInlineSchemas(
   appFolder: string,
-  defaultTs: "bun" | "deno" = "bun",
+  defaultTs?: "bun" | "deno",
 ): Promise<Record<string, any>> {
+  const ts = defaultTs ?? (await getDefaultTs());
   const schemas: Record<string, any> = {};
   const backendPath = path.join(appFolder, APP_BACKEND_FOLDER);
 
@@ -1002,7 +1004,7 @@ export async function inferAllInlineSchemas(
     const result = await inferRunnableSchemaFromFile(
       appFolder,
       fileName,
-      defaultTs,
+      ts,
     );
     if (result) {
       schemas[result.runnableId] = result.schema;
--- a/cli/src/commands/app/app_metadata.ts
+++ b/cli/src/commands/app/app_metadata.ts
@@ -194,7 +194,7 @@ export async function generateAppLocksInternal(
   const runnablesFolder = path.join(appFolder, APP_BACKEND_FOLDER);
   let rawAppRunnables: Record<string, any> = {};
   if (rawApp) {
-    rawAppRunnables = await loadRunnablesFromBackend(runnablesFolder);
+    rawAppRunnables = await loadRunnablesFromBackend(runnablesFolder, opts.defaultTs);
     if (Object.keys(rawAppRunnables).length === 0) {
       rawAppRunnables = (appFile as RawAppFile).runnables ?? {};
     }
--- a/cli/src/commands/app/app_metadata.ts
+++ b/cli/src/commands/app/app_metadata.ts
@@ -41,7 +41,7 @@ import {
   newRawAppPathAssigner,
   SupportedLanguage,
 } from "../../../windmill-utils-internal/src/path-utils/path-assigner.ts";
-import { getDefaultTs, mergeConfigWithConfigFile, SyncOptions } from "../../core/conf.ts";
+import { mergeConfigWithConfigFile, SyncOptions } from "../../core/conf.ts";
 import { resolveWorkspace } from "../../core/context.ts";
 import { requireLogin } from "../../core/auth.ts";
 import { getNonDottedPaths } from "../../utils/resource_folders.ts";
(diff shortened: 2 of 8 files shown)


rubenfiszel:
## Nit responses

**Dismissing the `opts.defaultTs` nit (@claude, @pi) — applying it re-breaks this PR's fix.**

Raw-app inline runnables only support `bun` and `python3`. The editor coerces
anything else before it reaches the API:

```ts
// frontend/src/lib/components/raw_apps/RawAppEditor.svelte:637-641
// Backend API only supports 'bun' and 'python3' for inline scripts
// Map TypeScript variants to 'bun'
if (language === 'nativets' || language === 'deno') {
  language = 'bun'
}
```

So a `.ts` file under a raw app's `backend/` is unconditionally bun — the
workspace `defaultTs` does not apply there, and the hardcoded default is correct
rather than an oversight. All six `loadRunnablesFromBackend` call sites omit it
for this reason.

Verified against a local backend + worker, `wmill.yaml` with `defaultTs: deno`,
a raw app runnable importing `windmill-client` pinned to `1.742.0` in
`dependencies/package.json`:

| | files written | lock |
|---|---|---|
| as merged (hardcoded `bun`) | `test.ts`, `test.lock`, `test.yaml` | `"windmill-client": "1.742.0"` ✅ |
| with `opts.defaultTs` passed | `test.ts`, `test.yaml` | **no lock generated at all** ❌ |

Passing `defaultTs` reads the runnable as `deno`, `deno` has no workspace-deps
manifest, so `filterWorkspaceDependencies` excludes `dependencies/package.json`
and the runnable ends up with no lock — the exact failure this PR fixes.

Separate pre-existing issue spotted while checking this, not touched here: in a
`defaultTs: deno` workspace the current code also writes a duplicate
`test.bun.ts` alongside `test.ts`, because `newRawAppPathAssigner(defaultTs)`
maps bun → `bun.ts` while the loader named the file `test.ts`. Unrelated to this
regression and unaffected by this PR either way — worth its own issue.

**Applied the two test nits** in 031f17d665:

- The tree-path point is fair. It cannot be tested offline: tree mode sources
  deps from `getMismatchedWorkspaceDeps()`, which is only populated by an
  `uploadScripts` round-trip. Both paths filter the same `appValue`, so
  resolving it correctly is what the test pins — now stated in the test header
  so nobody "improves" it into a backend-dependent test. The tree path is
  covered by the e2e verification in the PR description.
- Dropped the narrating comment; kept the one stating the invariant.

rubenfiszel:
✅ Review round clean @ 031f17d6659cb763fd102703625e9c90412f8a63 — nit-only verdicts at 5273b8415ccfbb79c395ae199076454ed77503cc (codex: Good to merge); test nits addressed in 031f17d665, `defaultTs` nit dismissed with evidence in the comment above

rubenfiszel:
## Correction: retracting my `defaultTs` dismissal — @claude and @pi were right

My previous comment dismissed the `opts.defaultTs` nit. That dismissal was wrong on
both the evidence and the reasoning. Applied in 7bd0f293b0.

**The claim was unsupported.** I quoted a comment in `RawAppEditor.svelte` saying the
backend only accepts `bun` and `python3`. That block feeds `getBackendRunnable` /
`getBackendRunnables`, an iframe SDK bridge — it is not the persistence path, so it
does not constrain what a stored runnable's language can be. The actual type is
`InlineScript & { language: ScriptLang }` (`raw_apps/utils.ts:12`), the full language
enum. I presented a code comment as a contract; it is not one.

**The experiment was mislabeled.** I hand-authored `backend/test.ts` in a
`defaultTs: deno` workspace while intending it as bun, then read "no `package.json`
applied" as breakage. But the writer convention is the opposite —
`getLanguageExtension(language, defaultTs)` returns `ts` only when
`language === defaultTs`:

| workspace | bun runnable | deno runnable |
|---|---|---|
| `defaultTs: bun` | `test.ts` | `test.deno.ts` |
| `defaultTs: deno` | `test.bun.ts` | `test.ts` |

So a plain `.ts` under a raw app's `backend/` means *defaultTs*, not *bun*. The CLI
would never write the file I hand-authored. Reading it back with `defaultTs` is the
symmetric, correct behavior; my test was measuring my own mislabeling.

**Re-run properly**, `defaultTs: deno`, correctly-named `test.bun.ts`:

| | files | lock |
|---|---|---|
| hardcoded `bun` | `test.bun.ts`, `test.lock`, `test.yaml` | `"windmill-client": "1.742.0"` |
| `opts.defaultTs` | `test.bun.ts`, `test.lock`, `test.yaml` | `"windmill-client": "1.742.0"` |

Identical — the compound extension is unambiguous either way. The two only diverge on
a plain `test.ts` in a deno workspace, and there the hardcoded `bun` is what causes the
duplicate-file corruption I misattributed earlier: the loader reports `bun`, then
`newRawAppPathAssigner('deno')` maps bun → `bun.ts` and writes a second file
`test.bun.ts` beside `test.ts`. Passing `defaultTs` removes it. So the duplicate I
reported as "pre-existing, unrelated" is in fact this nit's symptom, and this fixes it.

Also worth stating plainly: for `defaultTs: bun` — the default, and every fixture in
the PR description — `getLanguageFromExtension('ts', 'bun') === 'bun'`, so this is a
no-op. Re-verified: 991 unit tests pass, and the three bun-workspace e2e fixtures
produce byte-identical locks.

**Known remaining asymmetry, deliberately not fixed here:** the other five
`loadRunnablesFromBackend` call sites (`raw_apps.ts:380` in `pushRawApp`, three in
`dev.ts`, one in `lint.ts`) still omit `defaultTs`, so in a `defaultTs: deno`
workspace push would still read a plain `test.ts` as bun while lock generation now
reads it as deno. That is a pre-existing bug in its own right and fixing it means
touching the push path, which is outside this regression. Filing separately rather
than widening this PR.

rubenfiszel:
## Nit round addressed

**Fixed — orphaned `pushObj` JSDoc** (@codex P2, @claude P2) in bec4de186a. The
interface now sits above the doc block with per-field TSDoc, and the function
keeps its own JSDoc with `@param opts`. The three fields that were never
documented (`permissionedAsContext`, `wsSpecific`, `defaultTs`) now are.

**Deferring, explicitly — `wmill app push` / `app push-raw` use top-level
`defaultTs`** (@claude P2, carried). Confirmed the analysis: `push()` does
`opts = mergeCliWithEffectiveOptions(originalCliOpts, effectiveOpts)` after
`resolveEffectiveSyncOptions`, so `sync push` carries the override-resolved
value all the way to `pushRawApp`; `mergeConfigWithConfigFile` is a plain
`Object.assign` and does not.

Not fixing it here, for the reason @claude identified: it is the CLI-wide
convention, and `wmill script push` resolves `defaultTs` the same way before
feeding `inferContentTypeFromFilePath`, so a bare `.ts` standalone script is
already mislabeled in exactly that config on `main`. Patching only the two app
commands would make `script push` the odd one out and leave the convention
half-migrated. It wants its own PR that moves every standalone push command onto
the effective settings at once. Recorded under "Known deferred" in the PR body.

**PR body updated.** It described the original two-file change and claimed the
other call sites were being filed separately — both stale since 68e149d0b3. It
now covers the `defaultTs` threading, the `pushObj` refactor, the deno-workspace
verification, and the deferral above.

**Re-ran locally at bec4de186a**, since the previous figures predated three
commits: 991 unit tests pass, `bunx tsc --noEmit` clean apart from the two
pre-existing `sync.ts` JSZip errors, and a full `wmill sync push` through the
refactored `pushObj` lands the runnable as `type=inline language=bun` with the
pin intact.

rubenfiszel:
✅ Review round clean @ bec4de186a428cffb69cc154bdcf07aeae586f53 — nit-only verdicts at 0b01a1d98ab57c8aebee464b80a23ec2cf5d12a5 (codex + claude: Mergeable with nits; pi: Good to merge); JSDoc nit fixed in bec4de186a, effective-`defaultTs` nit dismissed with reasoning in the comment above

rubenfiszel:
## `wmill app dev` — verified, plus one correction and one pre-existing bug found

**Correction to the deferred note.** It named only `app push` / `app push-raw`
as reading top-level `defaultTs`. `dev.ts` resolves it through the same
`mergeConfigWithConfigFile`, so `app dev` has the identical
`workspaces[*].overrides` gap. PR body updated to say all three. No code change.

**Actually ran it** (I had only typechecked it before). Bun workspace, raw app
with a `backend/` runnable: dev starts, infers the runnable schema, generates
`wmill.d.ts` with `backend.test` / `backendAsync.test`, serves on the port, and
the watcher arms. Same in a `defaultTs: deno` workspace with a `test.bun.ts`
runnable.

**Pre-existing bug found while running it — not fixed here.** A code-only
runnable (a `.ts` in `backend/` with no sibling `.yaml`, which
`loadRunnablesFromBackend` documents as its auto-detect mode) is skipped by
schema inference, so `wmill.d.ts` types its args as `{}`:

```
backend/test.bun.ts  ->  Runnable test not found in backend folder or raw_app.yaml
                         test: (args: {}) => Promise<any>          # x: string dropped
+ backend/test.yaml  ->  Inferred schema for test
                         test: (args: { x: string }) => Promise<any>
```

Cause is the fallback chain in `inferRunnableSchemaFromFile`: when the sibling
YAML is absent it reads `raw_app.yaml`, and a missing `runnables[id]` there is
treated as "runnable not found" rather than as inline. The `{ type: "inline" }`
fallback the comment describes is only reachable when `raw_app.yaml` cannot be
read at all, which never happens in a real app — so it disagrees with
`loadRunnablesFromBackend`, which does auto-detect the same file as inline.

Reproduces identically on `81b23a2ba0` (this branch's base), so it predates this
PR and is unrelated to the workspace-dependency regression. Leaving it out to
keep this PR scoped; it deserves its own fix.

PR #10428, opened 2026-07-31
fix: add apps:run to the token scope picker and confine path-scoped app tokens

rubenfiszel:
## Summary

Running a deployed app's components was unreachable with any scoped token. `apps_u/execute_component` (and `apps_u/upload_s3_file`, which the S3 File Input component drives) are classified as `Run` actions by `RUN_PATH_ACTIONS`, so they require an `apps:run` scope — but the token scope picker only ever offered `apps:read` and `apps:write` for the Apps domain, and `ScopeAction::Write` does not include `Run`. Selecting *every* scope the UI offered still failed the route check, which makes `OptAuthed` resolve to `None` and surfaces as the misleading `publisher execution mode requires authentication`. This affects both regular and raw apps, since both run components through the same `apps_u` route.

Exposing `apps:run` also required closing a gap it would otherwise have inherited: the picker marks these scopes path-selectable, but `execute_component` / `upload_s3_file_from_app` only honored the path qualifier for app embed tokens. A picker-minted `apps:run:u/me/appA` carries no embed sentinel, so it reached *every* app in the workspace — including publisher-mode apps, whose runnables execute under the app author's identity. The path is now enforced for every scope-restricted caller.

Fixes WIN-2269

## Changes

- `windmill-api/src/token.rs`: the Apps domain exposes a third option, `apps:run` ("Run", path-selectable like read/write), so the scope can be granted through the supported UI.
- `windmill-api-auth/src/scopes.rs` + `lib.rs`: `apps:write` now satisfies a required `Run` action — at the route layer (`scope_grants_access`), resource-aware in `ScopeDefinition::includes`, and in `scope_contains` so an app-editor token can also mint the narrower `apps:run:<same path>` credential. Write can rewrite the app and its policy, so it covers running the same app's components; existing `apps:write` tokens keep working without also adding `apps:run`. Confined to the Apps domain — `jobs:write` still does not grant `jobs:run`.
- `windmill-api/src/apps.rs`: `execute_component` and `upload_s3_file_from_app` path-check `apps:run:<path>` for **every** scope-restricted caller instead of only app-embed tokens. `check_scopes` is a no-op for unscoped credentials, so cookie sessions, plain user tokens and anonymous public-app callers are unaffected. The app S3 read routes (`download_s3_file` and the `load_*` display ops behind `app_s3_on_behalf_and_provenance`) accept `apps:run:<path>` alongside `apps:read:<path>` via a shared `check_app_s3_read_scope`, so a run-scoped token can read back and render what its own runs produced; the provenance/policy gate behind them is untouched.
- Regression tests: the picker exposes `apps:run` path-selectably; the run routes accept `apps:run`/`apps:write` but not `apps:read`; and (in `tests/app_preview_auth.rs`) a path-qualified `apps:run|write:<path>` token is rejected on another app while still clearing the gate for its own.

## Screenshots

Apps domain in the token scope picker (`Limit token permissions` → Apps):

apps-run-scope-picker

## Test plan

Verified end-to-end against a local dev backend with two deployed publisher-mode apps (`u/admin/scope_app`, `u/admin/other_app`) wired to a deployed script:

- [x] `GET /api/tokens/list/scopes` returns `apps:run` (label `Run`, `requires_resource_path: true`) under Apps, and the picker renders it and emits `apps:run` when selected.
- [x] `apps:run:u/admin/scope_app` → `execute_component` on `scope_app` returns 200 and the job completes `success` with `trigger_kind = app`; on `other_app` it is `403 Required scope: apps:run:u/admin/other_app`.
- [x] `apps:write:u/admin/scope_app` → identical results (this was the reported failure).
- [x] `apps:read:u/admin/scope_app` → still rejected.
- [x] No over-blocking: unqualified `apps:run` reaches both apps; an unscoped session token reaches its app; anonymous callers are unchanged (still policy-gated).
- [x] `cargo test --test app_preview_auth`, `cargo test -p windmill-api-auth --lib scopes::`, `cargo test -p windmill-api --lib -- token::tests apps::embed_token_tests` (the app-embed scope matrix is unchanged). The new confinement test was confirmed to fail without the handler fix.
- [x] Parquet routes (dev backend rebuilt with `quickjs,parquet,private`): `upload_s3_file`, `download_s3_file` and `load_file_metadata` clear the scope gate for `apps:run`/`apps:write`/`apps:read` on their own app (falling through to the provenance gate) and 403 with the required-scope error on another app.
- [x] `cargo check --features parquet,quickjs`.

### DIFF
--- a/backend/tests/app_preview_auth.rs
+++ b/backend/tests/app_preview_auth.rs
@@ -22,7 +22,9 @@
 //!   - run mode against a deployed Viewer app rejects caller-supplied inline
 //!     `raw_code` whose sha is not publisher-pinned (CVE-2026-22683 residual:
 //!     the Viewer default-triggerable fallback let any caller / an operator run
-//!     arbitrary code as themselves, bypassing the content-hash pin).
+//!     arbitrary code as themselves, bypassing the content-hash pin), and
+//!   - a path-qualified `apps:run|write:<path>` token reaches only that app
+//!     (the route layer is resource-blind, so the handler must path-check).
 
 use serde_json::json;
 use sqlx::{Pool, Postgres};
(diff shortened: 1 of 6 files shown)


rubenfiszel:
✅ Review round clean @ e3799fa630bf5e62e0c76c259cd889edce646962 — codex and pi "Good to merge", claude nit-only at that SHA; the stale doc wording it flagged is fixed in beb710955c (comment text only, no behavior change).

PR #10431, opened 2026-07-31
feat: stamp webhook trigger_kind on token-driven job runs

rubenfiszel:
## Summary

Runs started directly through `/jobs/run*` leave `v2_job.trigger_kind` NULL, so the Runs
page cannot tell a run fired by a webhook, the CLI or an SDK apart from one a person
started in the browser. Before #10423 that was visible by accident — `created_by` was
`label-<token label>` for token-driven runs — and that PR removed it when it started
crediting the token owner instead.

This stamps **`webhook`** on those runs when the request carried an ordinary token, and
deliberately stamps **nothing** when it carried a browser-session token. The distinction is
derived server-side from the token label; the client cannot assert it (no query arg, header
or body field), because `trigger_kind` is authority-bearing — `apps.rs` uses
`trigger_kind = 'app'` as the app-origination marker for S3 provenance.

The fallback applies **only when no trigger built its own `TriggerMetadata`**, so the HTTP
route handler, the app runtime, the SMTP server, native triggers and schedules all still win.

## Why the browser session is left unstamped

`JobTriggerKind::Ui` is added to the enum, the OpenAPI schema and the frontend display maps,
but **nothing writes it yet**. Adding an enum value is a wire-format change against every
running binary, and neither decoder degrades — both fail hard:

```
sqlx  (DB workers, pull query):  'ui'  => Err(invalid value "ui" for enum JobTriggerKind)
serde (agent workers, JSON):     "ui"  => Err(unknown variant `ui`, expected one of ...)
```

A worker that predates the variant would claim the job, fail to decode it and strand it for
the zombie reaper — for *every* browser-started run, for the length of a rolling upgrade.
`webhook`, by contrast, is the oldest variant and readable by every supported binary, so it
is safe to write today. Shipping `Ui` now (decodable, unwritten) is exactly what a later
release needs before it can start stamping it.

Consequence, stated plainly: a NULL `trigger_kind` means "browser, or any run predating this
PR". The UI half is the absence of a marker, not a positive label.

## Stop this recurring: workers no longer decode the enum

The worker only ever asks two yes/no questions of `trigger_kind` (`is it a schedule` → the
schedule path for error/recovery handlers and re-scheduling; `is it a ci_test` → a reserved
variable) and stringifies it for a tracing span. It never enumerates the set — the strong
typing on the pull path was incidental, and it is what turns an unfamiliar label into an
unreadable row.

So `MiniPulledJob`, `MiniCompletedJob` and `QueuedJobV2` now hold the raw label
(`TriggerKindLabel`, a serde-transparent `String` with a sqlx `Decode` that reads the enum as
text) instead of `JobTriggerKind`. From this release on, a worker keeps running jobs stamped
with a kind added after it was built — through both decode paths. It does not rescue *this*
rollout (old binaries carry their own query and struct), which is why `ui` still waits.

## Changes

- `JobTriggerKind::Ui` + `as_str()`; `Display` delegates to it.
- `TriggerKindLabel` in `windmill-types`, used by the three worker/queue job structs.
- `ApiAuthed::is_session_token`, set from the token label where it is already in hand
  (`AuthCache::get_opt_job_authed`) rather than by re-querying `token` per run. New
  `windmill_common::auth::is_session_label` next to `is_server_minted_label`.
- `ApiAuthed::trigger_or_fallback(trigger)` — one place holding "a real trigger wins,
  otherwise attribute from the token" — applied at all 11 `/jobs/run*`, `run_wait_result*`,
  `run_and_stream*`, preview, `batch_rerun` and `restart` push sites.
- `trigger_kind` returned by the single-job GET (`#[sqlx(default)]`, like `is_retry`), so the
  run detail panel can render it. It was never exposed before, which is why the
  `trigger_kind` branch of `getTriggerInfo` was dead code. The runs *list* payload is unchanged.
- Frontend: the Trigger row renders whenever a kind was stamped and `getTriggerInfo` can name
  it; `JobDetailFieldConfig` now uses the shared `triggerDisplayNamesMap` instead of a stale
  local copy that rendered `asset`/`github` as "ASSET"/"GITHUB". `ui` is mapped for icon and
  display name but kept **out** of `jobTriggerKinds`, so the filter does not offer a kind that
  can only ever return nothing.

## Deliberate choices worth a reviewer's eye

- **`webhook` is the catch-all.** The CLI and the SDKs bucket as `webhook`, matching the
  `wm_trigger.kind = "webhook"` the preprocessor already reports for these routes. No
  `api`/`cli` value is added here.
- **`ui` will attribute, not prove.** `/users/refresh_token` mints a `session`-labelled token
  for any authenticated caller, so a member could obtain one for a script. Only the frontend
  calls it, so nothing is mislabelled in normal use, and the comments on `JobTriggerKind::Ui`
  and `is_session_label` say explicitly that nothing may be gated on it — in contrast to
  `App`, which is authoritative.
- **Not covered**: `/jobs/run/workflow_as_code/*` (a job calling back into the API with its own
  token — `webhook` would be actively wrong), and the dependency-job routes (`job_kind` already
  identifies them).
- **No backfill.** Existing rows stay NULL, and `not_in_nullable` keeps them included by a
  negated filter, so the `Hide schedules` preset (`job_trigger_kind: !schedule`) is unchanged.

## Test plan

- [x] `only_a_browser_session_is_left_unstamped`: the session label yields no stamp; every
      member-mintable label and a label-less token yield `webhook`.
- [x] `a_real_trigger_wins_over_the_token_fallback`: a supplied `TriggerMetadata` survives.
- [x] `an_unknown_trigger_kind_label_survives_deserialization`: the point of `TriggerKindLabel`.
- [x] Run started from the UI → no stamp; `curl` with a personal token → `webhook` (verified in
      the DB and in the browser).
- [x] A job stamped `ui` directly in the DB is pulled and **executed successfully** by a
      worker — the tolerant decode working on the real pull path.
- [x] A schedule that fires while the change is live still stores `trigger_kind = 'schedule'`,
      `trigger = <path>` — the fallback does not clobber it.
- [x] Filters against a DB of 5 NULL / 4 `ui` / 3 `webhook`: `webhook` → 3, `!webhook` → 9,
      `!schedule` → 12 (all rows, legacy NULLs included).
- [x] `cargo check --workspace --tests`, the same with
      `--features enterprise,private,native_trigger,http_trigger`, and `npm run check` (0 errors).

## Screenshots

Run started with `curl` + a personal token, and the same script started from the UI — the
first is marked, the second deliberately has no Trigger row (not an empty one):

run-detail-webhook

run-detail-ui-unstamped

Runs page filtered to `webhook`:

runs-filter-webhook

EE companion: windmill-labs/windmill-ee-private#700


rubenfiszel:
Pre-push local review (Claude branch-diff-reviewer + Codex) surfaced five findings; here is how each was resolved, so the CI round doesn't re-derive them:

- **[P1, Codex] `ui` is forgeable via `/users/refresh_token`.** Confirmed: that handler mints a `session`-labelled token for *any* authenticated caller. Only `frontend/src/routes/(root)/+layout.svelte` calls it, so nothing is mislabelled in normal CLI/SDK use — the exposure is a member who deliberately obtains a session token for a script. Fixed the *claim* rather than the endpoint (201bc22): `is_session_label` and `JobTriggerKind::Ui` now state that this attributes rather than proves, and that nothing may be gated on it — explicitly the opposite of `App`, which is the authoritative marker. Narrowing `/users/refresh_token` to session-only callers would change an auth endpoint's contract to harden a Runs-page label; I judged that the worse trade, but it is a one-line change if a reviewer disagrees. I did not add the suggested "personal token cannot be laundered into `ui`" test, because it would assert the opposite of the documented design.
- **[P2] Rolling-upgrade decode.** The worker pull query decodes `trigger_kind` into the Rust enum, so a worker predating `Ui` drops every job carrying it — and unlike the earlier `app`/`asset`/`freshness` additions this would hit every browser-started run. Gated the stamp on `MIN_VERSION_SUPPORTS_UI_TRIGGER_KIND` (1.776.0); nothing is written until the fleet can decode it. **Bump that constant if this lands in a different release.**
- **[P2] Empty "Trigger" row for unmapped kinds.** `ci_test` is stamped on real jobs and is in neither `triggerIconMap` nor the OpenAPI enum, so it would have rendered `Trigger  no value`. The presence rule now requires the kind to be one `getTriggerInfo` can name.
- **[P2] Stale duplicate display-name map.** `JobDetailFieldConfig.getTriggerDisplayName` was a second, incomplete copy of `triggerDisplayNamesMap` and rendered `asset`/`github` as "ASSET"/"GITHUB". Deleted in favour of the shared map.
- **[P2] `ui` missing from the `<kind>_trigger` reject list.** `POST /trigger/ui/resume_suspended_trigger_jobs/*` would have 500'd on a missing relation instead of 400'ing. Added to the arm alongside `App`.

rubenfiszel:
Nits from the round at `201bc22a2d27e8342e50b47758385059cd8c9434` are addressed in 20c0b5a791:

- **Duplicated fallback `match`.** Folded into one `ApiAuthed::trigger_or_fallback(trigger)`. All 11 push sites now read `authed.trigger_or_fallback(<trigger|None>).await`, the invariant lives in one place, and each push argument is back to a single line (which also keeps the conflict surface small against the branch changing `push`'s signature). Took the follow-on suggestion too: `a_real_trigger_wins_over_the_token_fallback` now pins that a supplied `TriggerMetadata` survives, for a session-token authed — the case that would silently re-attribute a scheduled run to the UI.
- **Version gate is worker-scoped.** Correct, and the code comment already said "worker" — the PR body said "fleet". Fixed the body: `MIN_VERSION` is the min over `worker_ping`, which only workers write, so an old API-server replica mid-rollout can still meet a `ui` row it cannot decode on the `jobs_export` path. Nothing to change in this PR — `MIN_VERSION`'s scope is a property of that machinery, not of this change — but the residual is now stated rather than implied.

On the suggested manual check of the deployed-app case: it holds by construction rather than by testing. `apps.rs::execute_component` calls `push` directly with its own `app_trigger`, and that call site is not in this diff (`git diff main...HEAD --name-only` has no `apps.rs`), so both deployed component runs (`app`) and app-editor previews (no stamp) behave exactly as on main — the new fallback is never reached there.

rubenfiszel:
✅ Review round clean @ 20c0b5a791607049cb0e197520f8000e0584b89b — nit-only verdicts at 201bc22a2d27e8342e50b47758385059cd8c9434 (codex: Good to merge, pi: Good to merge, claude: nit-only); nits addressed in 20c0b5a791

rubenfiszel:
Design changed after the first clean round — back to draft for a fresh one.

**What changed and why.** The previous revision stamped `ui` for browser sessions and guarded it with a `MIN_VERSION` constant. Pushing on *why* that guard was needed turned up a better answer: adding an enum value is a wire-format change against every running binary, and I verified both decoders fail hard rather than degrading —

```
sqlx  (DB workers, pull query):  'ui'  => Err(invalid value "ui" for enum JobTriggerKind)
serde (agent workers, JSON):     "ui"  => Err(unknown variant `ui`, expected one of ...)
```

— so the guard was the only back-compat available, not an alternative to it. Rather than ship a version constant that must be bumped at release and silently disables the feature if it isn't, this revision **only writes `webhook`** (the oldest variant, readable by every supported binary) and leaves the browser session unstamped. `JobTriggerKind::Ui` still ships — decodable, unwritten — which is precisely what a later release needs before it can start stamping it. The `MIN_VERSION_SUPPORTS_UI_TRIGGER_KIND` constant is gone.

**And the fragility is fixed at the root.** The worker only asks `is it a schedule` (→ schedule path for error/recovery handlers and re-scheduling) and `is it a ci_test` (→ a reserved variable), plus a tracing-span string. It never enumerates the set, so the enum typing on the pull path bought nothing and cost the ability to read unfamiliar rows. `MiniPulledJob` / `MiniCompletedJob` / `QueuedJobV2` now hold the raw label via `TriggerKindLabel` (serde-transparent `String`, sqlx `Decode` reading the enum as text). From this release on a worker keeps running jobs stamped with a kind added after it was built, through both decode paths — verified end-to-end by stamping a queued job `ui` in the DB and watching a worker pull and complete it.

**Accepted cost**, worth a reviewer's disagreement: a NULL `trigger_kind` now means "browser, or any run predating this PR". The UI half is the absence of a marker rather than a positive label, and the Runs filter deliberately does not offer a `ui` option that could only return nothing.

Nits from the previous round (single `trigger_or_fallback` helper, worker-scoped gate wording) are still addressed; the helper is now sync again since there is no `.met().await` left, which puts each push argument back to one line.

rubenfiszel:
Blocking finding from both Codex and Claude is fixed in 198c013 — they were right and my earlier reasoning was wrong.

I'd checked `git status backend/.sqlx` after the type change, saw it clean, and concluded no cache work was needed because the SQL text hadn't changed. It had: the `as "col: Type"` annotation lives *inside* the macro's query string, so switching `JobTriggerKind` → `TriggerKindLabel` changes the cache key. The old entries were orphaned and the new ones never generated, so `SQLX_OFFLINE=true` builds would have failed on all three queries.

Regenerated with `./update_sqlx.sh` (the full feature set, EE included). That run also wanted to delete ~50 unrelated entries — pre-existing drift in `main`'s cache, nothing to do with this branch — so I restored those and staged only the three swaps this change actually causes:

```
R  query-67e25a7c….json -> query-09ae1230….json   (get_mini_completed_job)
R  query-b3771b69….json -> query-9d331b63….json   (get_queued_job_v2)
R  query-d4211392….json -> query-d0df9d1f….json   (MiniPulledJob by id)
```

File count is unchanged at 2421, and nothing present in `main` is missing.

Verified the way CI does rather than by inspection — both offline builds are clean:

```
SQLX_OFFLINE=true cargo check --workspace --all-targets \
  --features all_sqlx_features,ee,deno_core,private,enterprise,mcp   # clean
SQLX_OFFLINE=true cargo check                                         # clean
```

rubenfiszel:
Rebased onto `main` and resolved; head is now `ea49c2b6`.

**EE.** windmill-ee-private#700 was squash-merged as `7de7daf`, so `backend/ee-repo-ref.txt` now points there. `main` had independently advanced its ref to `68aedf0`, which is an ancestor of `7de7daf`, so mine strictly supersedes it — that was the only true conflict. The `windmill-internal-app` bot had already pushed the same bump to this branch; I took its version of the file verbatim so the automation stays authoritative.

**The sibling audit PR.** #10433 ("carry the token label into job-run audit rows") landed on `main` and touches the same files as this branch — `windmill-api-auth/src/lib.rs`, `windmill-api/src/jobs.rs`, `windmill-api-jobs/src/execution.rs`, `windmill-queue/src/jobs.rs`. All four auto-merged, but a clean textual merge of adjacent changes to a 28-argument `push` call is exactly where an argument can silently shift, so I checked it rather than trusting it: `push`'s signature is unchanged in the merged tree (`audit_end_user` went elsewhere), the `trigger` argument still sits between `end_user_email` and `suspended_mode` at every call site, and `is_session_token`/`trigger_or_fallback` coexist with the new `display_username` audit logic.

Verified on the merged tree:
- `cargo check --workspace --tests`, the same with `--features enterprise,private,native_trigger,http_trigger`
- `SQLX_OFFLINE=true cargo check --workspace --all-targets --features all_sqlx_features,ee,deno_core,private,enterprise,mcp` — clean, so main's new queries and mine coexist in the cache
- `windmill-api-auth` (42), `windmill-types` (34) and `windmill-queue` test suites all pass
- `npm run generate-backend-client` + full frontend typecheck — 0 errors

Also folded in the nit from the last round (b5fdd63): `QueuedJob`/`CompletedJob.trigger_kind` now use `TriggerKindLabel` as well, so the API side no longer reintroduces the hard-decode failure this PR removes from the worker.

rubenfiszel:
✅ Review round clean @ ea49c2b69e006119eb5747af3df1daded805b176

PR #10433, opened 2026-07-31
fix: carry the token label into job-run audit rows

rubenfiszel:
## Summary

`push` builds its own `AuditAuthor` from `(user, permissioned_as)` and never sees the `ApiAuthed`, so it sets the audit `end_user` only for the genuine on-behalf case. Every other handler passes `&authed`, whose `AuditAuthorable` impl returns `username_override`, which is still `label-<label>` for a user token.

Before #10423 the two coincided, because `user` itself *was* `label-<label>`. That PR made `created_by` credit the token owner instead, so `jobs.run.*` rows stopped carrying `end_user` at all.

`list_audit` filters on `username = ? OR parameters->>'end_user' = ?`, so searching the audit log for a token label returns that token's `scripts.create` / `variables.create` / `resources.update` rows but omits its `jobs.run.*` rows. The runs stay identifiable, since `audit.span` still carries the token prefix, so this is an incomplete view in audit search rather than a loss of identifiability.

Observed on an EE instance before the fix, with one labeled token driving two operations:

```
op=variables.create   username=test-user  end_user="label-probe-tok"
op=jobs.run.preview   username=test-user  end_user=NULL
```

## Changes

- `windmill_queue::push` takes an explicit `audit_end_user: Option<&str>`, right after the `token_prefix` it already carries. The API call sites source it from `authed.username_override`, next to the `authed.token_prefix.as_deref()` they already pass.
- It takes precedence over the value `push` derives from `(user, permissioned_as)`, falling back to that when `None`. The invariant: the audit `username` is the identity the job runs as, and `end_user` identifies the caller behind it, the same split an app run already uses.
- When a runnable declares `on_behalf_of`, the run-as identity takes `username` and the label takes `end_user`, leaving the caller with no slot. `push` records it as a `created_by` audit parameter on exactly those rows, and `list_audit` matches it as a third term.
- Non-API callers (schedules, the reaper, native triggers, asset dispatch, the freshness watchdog, the EE cascade) pass `None` and are byte-identical.
- `execute_component` also passes `None`, because an app run's caller is whoever used the app, which `push` already derives from `username` there.

`created_by` on `v2_job` is left as the token owner, which #10423 deliberately made it. `require_job_read_access` is untouched, and nothing reads `end_user`.

## Behavior matrix

Token owner `alice`, generic user token labeled `mytoken`, so `username_override = "label-mytoken"`. Cells are the audit `username` · `end_user`, where `-` means absent. Bold marks a cell this PR changes relative to main.

| # | Case | pre-#10423 | main | this PR |
|---|---|---|---|---|
| 1 | Session cookie, plain script run | `alice` · `-` | `alice` · `-` | `alice` · `-` |
| 2 | Labeled token, plain script run | `alice` · `label-mytoken` | `alice` · `-` | `alice` · **`label-mytoken`** |
| 3 | Labeled token, runnable with `on_behalf_of: u/svc` | `svc` · `label-mytoken` | `svc` · `alice` | `svc` · **`label-mytoken`** (+ `created_by` = `alice`) |
| 4 | Labeled token, script deploy, `jobs.run.dependencies` | `alice` · `-` | `alice` · `-` | `alice` · **`label-mytoken`** |
| 5 | User-minted `webhook-…` token, script run | `alice` · `webhook-f/svc/s` | `alice` · `-` | `alice` · **`webhook-f/svc/s`** |
| 6 | Server-minted `ephemeral-webhook-…` token | `alice` · `ephemeral-webhook-…` | same | same |
| 7 | HTTP route / native trigger | `<owner>` · `<trigger identity>` | same | same |
| 8 | Deployed app, Publisher/Anonymous, alice | `appowner` · `alice` | same | same |
| 9 | Deployed app, Anonymous, no auth | `appowner` · `anonymous` | same | same |
| 10 | App editor preview (Viewer mode) | `alice` · `-` | same | same |
| 11 | App worker `ephemeral-script-end-user-alice` calling `/jobs/run` | `<runs-as>` · `alice` | same | same |
| 12 | Schedule | `<on-behalf user>` · `schedule-<path>` | same | same |
| 13 | Flow step of a labeled-token flow | `alice` · `label-mytoken` | `alice` · `-` | `alice` · `-` |
| 14 | Reaper / retry / error handler / asset dispatch / EE cascade | `<job.created_by>` · derived | same | same |

`v2_job.created_by` is a separate axis: #10423 changed it from `label-mytoken` to `alice` for rows 2, 3, 5 and 13, and this PR leaves that alone everywhere.

Reading the matrix:

- Rows 2, 3 and 5 restore the pre-#10423 row exactly.
- Row 4 is a new value rather than a restoration. Deploy-triggered dependency jobs pass `&authed.username`, never `display_username()`, so #10423 never perturbed them and they carried no `end_user` in either prior era. They now carry the label, matching the `scripts.create` row from the same request. The same applies to flow and app create/update, and to datatable migrations.
- Row 3 is the only row where the `end_user` slot was already occupied. The caller it displaces becomes `parameters->>'created_by'`, so that row stays searchable both by the token that fired it and by whoever fired it. Rows 6, 7 and 11 carry an override that already equals `user`, so the guard is `runs_on_behalf && audit_end_user.is_some_and(|e| e != user)` and they do not store the same string twice.
- Row 13 stays a gap. `worker_flow` pushes with `user = flow_job.created_by` and has no `ApiAuthed`, so #10423's change to `created_by` stopped the label reaching step rows, and nothing here can restore it because the label is not persisted on `v2_job`. A flow fired by a labeled token is searchable by label at the flow job but not at its steps.

## Indexing: measured, then deliberately dropped

`list_audit`'s identity filter has no supporting index. `audit_partitioned` carries only its primary key, a timestamp index, a workspace+timestamp index, and a partial login-activity index whose predicate a general username search does not imply. Adding `(workspace_id, username)`, `(workspace_id, (parameters->>'end_user'))` and `(workspace_id, (parameters->>'created_by'))` to each newly provisioned partition was implemented, benchmarked, and then dropped, because the planner does not use them.

Against 320k rows, on the audit page's actual query, searching a rare identity:

```
planner's choice   48.1 ms   MergeAppend over per-partition primary keys
forced bitmap       8.5 ms   BitmapOr across the three indexes
```

`list_audit` always ends in `ORDER BY id DESC LIMIT n`, and Postgres cannot produce ordered output from a three-way OR, so it prefers walking each partition's `(id, timestamp)` primary key backward and filtering. That is optimal when matches are dense and pathological when they are rare: a five-row identity among 320k rows scans the lot. The indexes are chosen only once the `ORDER BY` is removed, which no caller does.

They would therefore have cost three btree inserts per audit row, on a write path every operation goes through, for no read benefit. Making them pay off needs `id` as a trailing index column plus rewriting the OR as a UNION ALL of three separately-ordered, separately-limited scans. That is a change to the query builder rather than to this plumbing, so it is left as follow-up, along with the legacy `audit` table, which the same filter still scans until retention drains it.

## EE companion

windmill-labs/windmill-ee-private#699, which adds the `created_by` term to `list_audit` and passes `None` at the nine EE `push` call sites.

## Test plan

- [x] `cargo check --features quickjs --tests` and `cargo check --features enterprise,private`
- [x] `cargo test --features enterprise,private -p windmill-api-integration-tests --test audit`, 2 passed
- [x] `test_job_run_is_searchable_by_token_and_by_caller` was verified to **fail** with each half of the fix reverted independently: without the end-user precedence the label search returns `[]`, and without the `created_by` parameter the on-behalf run vanishes from the caller search
- [x] Workspace isolation re-checked directly: a row living only in another workspace, whose `end_user` and `created_by` match the search term, is not returned. `sql_builder` parenthesizes each `WHERE` fragment once there are two or more, so the added OR term stays inside the workspace predicate
- [ ] On an EE instance: mint a labeled token, fire a run against a script that declares `on_behalf_of`, then search the audit log for the label and for your own username. The same row should come back under both.


rubenfiszel:
✅ Review round clean @ 47dfaddb464221dd593a994c5c4b4f708602cf46 (codex: good to merge, pi: good to merge, claude: nit-only). The only nit was a stale read of the PR description: the duplicated `## Indexing` sections were rewritten before the round finished, and the body now carries a single one describing the benchmark that led to dropping the index commit. Head unchanged since the round.

PR #10436, opened 2026-07-31
feat: sync data table migrations to git, gated by a new object type

rubenfiszel:
Fixes GIT-942

## Summary

SQL migrations created for a data table never reached a linked git repository on
their own — only a manual "push to git" from workspace settings got them there.

**Why it never fired.** `should_skip_object_type` requires the deployed object's
`ObjectType` to be listed in the repo's `include_type`, but `datatablemigration`
existed only in the Rust enum: it was absent from the OpenAPI `GitSyncObjectType`,
from the git-sync filter UI, and from the CLI's `INCLUDE_TYPE_MAPPINGS`. No client
could put it there, so every data table migration deploy was dropped before a job
was enqueued — including for a workspace syncing `**` with every box ticked.

**Why it would still have committed nothing.** `DeployedObject::DatatableMigration`
is keyed by `<datatable>/<version>_<name>` (the `workspace_diff` key), but the
workspace export writes the files at
`migrations/datatable/<datatable>/<version>_<name>.{up,down}.sql`. The CLI turns the
item path into `--extra-includes` globs and a `git add` pathspec, so a job that did
get enqueued pulled nothing and staged nothing — `git add` failed with "pathspec did
not match any files", which the hub script swallows.

A third gap became reachable once migrations actually sync: renaming or deleting a
data table rewrote/removed `datatable_migrations` rows without emitting any deployed
object, so a linked repo would keep the pre-rename files forever.

## Data table migrations are now a first-class git-sync object type

`migrations/datatable/**` lives outside the `f/`/`u/` namespaces the include/exclude
**path** filters are written against — like `dependencies/`, `users/` and
`settings.yaml`, all of which already bypass those filters via
`get_ignore_regex_filter()`. Data table migrations now do the same, and a dedicated
`datatablemigration` object type governs them:

- `GitSyncObjectType` gains `datatablemigration` (`openapi.yaml` + the served
  `openapi-deref.{yaml,json}` + regenerated frontend client).
- A **Data table migrations** toggle in the git-sync filter settings, on by default
  in every `include_type` default the UI writes.
- `skipDatatableMigrations` in `wmill.yaml` (default `false` = synced), wired through
  `INCLUDE_TYPE_MAPPINGS`, both converter directions, `ignoreF`, and the tarball
  request as `skip_datatable_migrations` — so a repo that opts out never receives the
  files and therefore computes no diff for them in either direction.
- A migration backfills `datatablemigration` into every existing non-empty
  `include_type` (workspace-level and per repository). Without it every pre-existing
  workspace would read as opted-out until someone re-saved the settings form — which
  is exactly the reported bug. Absent/empty arrays mean "inherit" and are left alone.

**Push safety.** Because migrations bypass the path filters, a clone made before they
were synced sees every server-side migration as remote-only, and `pushMigrationFromDisk`
reads a locally absent `.up.sql` as "delete it". The working tree can't tell that apart
from a real deletion — git keeps no empty directory, and creating a migration locally
(`wmill datatable migrate new`) makes `migrations/datatable/` appear without anything
having been tracked. So the signal is git history: `gitRecordedDatatableMigrationPaths`
lists every path **the current branch** has committed under `migrations/datatable/`
(`HEAD`, not `--all` — a migration that only ever lived on another branch is not evidence
this branch tracked it), with the repo-root prefix stripped so it matches a sync diff's
cwd-relative paths. A recorded path's absence is a genuine deletion and flows through (it
stays recorded after the deletion is committed); a path this branch has never held is
never deleted on a guess — unattended runs (`--yes`, no TTY) and `--dry-run` keep those
migrations and say so, while a TTY push asks a deletion-specific question, defaulting to
no, once the user has seen the full change list. History that can't be consulted at all —
no repository, or a shallow clone whose truncated history would misreport paths as
never-seen — is reported as such and trusts nothing.

## Changes

- `DeployedObject::get_git_sync_path()` (new): the repo-relative path the sync item
  carries. Identical to `get_path()` for every kind except data table migrations,
  which gain the `migrations/datatable/` prefix. `get_path()` is untouched, so the
  `workspace_diff` key, the deployment-request anchor and `deployment_metadata` keep
  their shape.
- `get_ignore_regex_filter()` returns true for `DatatableMigration`.
- **[ee]** `git_sync_ee.rs`: the two path-filter checks and the `SyncItem` construction
  use `get_git_sync_path()`; `should_skip_object_type` is a plain `include_type` check
  again (the interim exemption from #702 is removed, since the type is now reachable).
- **[ee]** `folder_prefix_from_path` takes the first two segments of any path, not just
  `f/…` and `u/…`. The CLI derives a `group_by_folder` promotion branch the same way, so
  a `migrations/datatable/…` item was getting a per-item debounce key while the CLI put
  every one of them on the single `wm_deploy/<ws>/migrations__datatable` branch.
- `gitSyncIncludePattern`: a `datatable_migration` case expanding to the migration's
  `.up.sql` and `.down.sql`, instead of the script default (`${path}.*` plus a
  meaningless `${path}__mod/**`).
- `cascade_datatable_migration_renames_and_deletes` returns every migration path it
  moved or removed (both ends of a rename), and `edit_datatable_config` deploys them
  once its transaction commits.
- `validate_new_datatable_name` holds a name being **introduced** to `[A-Za-z0-9_.-]`
  starting alphanumeric. That segment passes through three matchers that all speak that
  charset — `transform_regexp` expands `**` to `[a-zA-Z0-9_\-./]*`, the CLI reuses the
  path as a minimatch pattern (where a leading `.` also needs `dot: true`), and the
  deploy stages it as a `git add` pathspec — so `sales[eu]`, `my db` or `.staging`
  match nothing in all three. Already-persisted names keep the historical rule, so a
  workspace carrying a legacy name can still save its settings and rename its way out;
  `record_datatable_migration_deployment` warns when it sees one.

## Screenshots

Git sync filter settings — the new toggle, next to Workspace dependencies:

gitsync-type-filters

## Test plan

Verified end-to-end against a local EE backend (`--features quickjs,enterprise,private,license`):

- [x] With `include_path: ["f/**"]` (the shipped default) and the type **unticked**: no job enqueued
- [x] With `include_path: ["f/**"]` and the type **ticked**: job enqueued, carrying
      `items: [{path_type: "datatable_migration", path: "migrations/datatable/mydb/<version>_<name>", …}]`
      — the path filter no longer gates migrations, the toggle does
- [x] `exclude_types_override: ["datatablemigration"]` enqueues nothing
- [x] `wmill sync git-deploy` into a repo scoped to `includes: ['f/**']` writes exactly the
      two `.sql` files and nothing else; the hub script's `git add wmill-lock.yaml '<path>**'`
      stages both and commits
- [x] Same run with the old un-prefixed path shape: 0 changes pulled and `git add` fails with
      "pathspec did not match any files" — both halves of the path fix are load-bearing
- [x] Deleting both files and re-running that `git add` stages `D` for both — the pathspec
      covers removals, which is the half the rename/delete cascade depends on
- [x] `skipDatatableMigrations: true`: the tarball omits them and the pull reports 0 changes
      (no phantom deletion)
- [x] Backfill migration on a seeded config covering all four shapes (workspace-level,
      per-repo non-empty, per-repo empty, repo with no `settings` key): the entry is added
      only where intended, array order and count preserved, idempotent on re-run, down reverses it
- [x] Toggle exercised in the browser (screenshot above)
- [x] `cargo test -p windmill-git-sync --features enterprise,private` (29 pass)
- [x] `cargo test -p windmill-api-workspaces --features enterprise,private datatable` (8 pass)
- [x] `bun test test/git_unit.test.ts test/gitsync_converter_unit.test.ts test/datatable_migrations_unit.test.ts`
- [x] `bun test test/include_flags_bypass_filtering.test.ts test/gitsync_settings_features.test.ts` (live backend)
- [x] `npm run check` — 0 errors; `cargo check --features enterprise,private`

**Not regenerated:** `openapi-deref.{yaml,json}` are only patched with the new enum value
rather than rebuilt. A full `build_openapi.sh` run produces a 12.9k-line diff (the checked-in
bundles are at `1.740.0` vs `openapi.yaml`'s `1.775.2` and every `&ref_N` renumbers), none of
it related to this change — that refresh belongs in its own PR.


rubenfiszel:
Round-2 nits addressed in acb83d4:

- **Leading-dot names** — real, fixed. `validate_new_datatable_name` now also requires the first character to be alphanumeric, which closes `.staging`, `.`, `...` and leading `-`/`_` in one rule. Test extended.
- **Legacy out-of-charset names are silently unsynced** — fixed. `record_datatable_migration_deployment` now emits a `tracing::warn!` naming the data table and the allowed charset when the name wouldn't survive `validate_new_datatable_name`, so the residual is observable per change instead of vanishing inside the path filter.

Dismissing the remaining two, with reasoning:

- **Delete/rename fan-out in individual-branch mode** — this isn't specific to data table migrations: promotion mode's contract is one branch per deployed object, so deleting a folder of 30 scripts fans out identically today. `group_by_folder` is the provided answer and the `folder_prefix_from_path` change in this PR is precisely what makes it collapse correctly for the new path shape. Changing the ungrouped fan-out would be a promotion-mode design change, not a fix to this bug.
- **No client-side name validation in `DataTableSettings.svelte`** — the settings form posts wholesale and surfaces server validation as a toast for every field it already validates server-side; singling out this one field for inline validation would be inconsistent, and the 400 names both the offending table and the allowed charset. Worth doing as part of a general pass on that form rather than here.

The `f/**` discoverability gap stays documented in the PR description as a follow-up: the filter behaving as configured is not a bug, and the delete-on-missing semantics of `pushMigrationFromDisk` are why bypassing it is unsafe.

rubenfiszel:
✅ Review round clean @ 1d5f40124f596c4945a2a6d4147ae7451800e35e

rubenfiszel:
Round 5 findings addressed in 59f0f83:

**P0/P1 — push from an untracked checkout deletes migrations.** Real, fixed. `dropUntrackedDatatableMigrationDeletions` drops `deleted` changes for data table migrations when the checkout has no `migrations/datatable/` directory at all — that means "this clone has never pulled them", not "delete them". It runs before `prettyChanges`, so the confirmation list matches what actually happens, and it logs what it skipped with the remedy (`wmill sync pull` first). Once the directory exists, real deletions flow through untouched. Unit test added pinning both branches, and the PR body's design note is rewritten to describe the new model rather than the old one.

**P2 — `openapi-deref.{yaml,json}`.** The enum value is now in both. I did *not* run `build_openapi.sh`: the checked-in bundles are at `1.740.0` against `openapi.yaml`'s `1.775.2`, so a full regeneration emits a 12.9k-line diff (version bump plus every `&ref_N` renumbering) that has nothing to do with this change. For the record, the cited precedent `dda5976` touched only `openapi.yaml`, `jobs.rs` and `lib.rs` — not the bundles — and 4 commits have modified `openapi.yaml` since the bundles were last updated, so per-PR regeneration isn't the current convention. The bundles are patched surgically (4 lines) and both still parse; a real refresh belongs in its own PR.

**P2 — screenshot / PR body.** Body rewritten: the stale "No frontend changes" line is gone and a `## Screenshots` section shows the new toggle in the git-sync filter settings.

**Test gaps.** `gitsync_converter_unit.test.ts` now covers the `datatablemigration` ↔ `skipDatatableMigrations` round-trip in both directions, and `datatable_migrations_unit.test.ts` covers the push-deletion scenario.

**On the EE half** (which reviewers can't see): the #702 exemption *is* removed — `should_skip_object_type` is a plain `include_type` membership check again, in windmill-labs/windmill-ee-private#703, and `ee-repo-ref.txt` points at its head. Verified at runtime, which is the check that matters: with `include_path: ["f/**"]`, creating a migration enqueues **no** sync job when the toggle is off and enqueues one when it is on. `folder_prefix_from_path` is also in that PR, matching the `migrations/datatable` branch the CLI test pins.

rubenfiszel:
Round 6 addressed in 191b733 — you were both right, and the fix is better than what I had.

Deleting the last migration removes `migrations/datatable/` (git keeps no empty directory), so the directory's absence genuinely cannot tell "never synced these" from "just deleted them all". Silently dropping the deletions was the wrong resolution: it made the destructive reading impossible and told the user something false about why.

Now the ambiguity is surfaced rather than guessed:

- `untrackedDatatableMigrationDeletions` only *identifies* the at-risk deletions; it no longer filters anything itself, and its doc comment states the ambiguity instead of implying it can't happen.
- The gate moved to just after the main "apply these N changes?" confirmation, so the user has already seen the full change list. On a TTY it asks a second, deletion-specific question defaulting to **no**; answering yes performs the deletion, so deleting the last migration from git still works.
- Non-interactive (`--yes`, CI) never deletes: it keeps the migrations and logs what it kept and the two ways forward (`wmill sync pull` to track them in git, or delete them from the workspace). That's the automated pipeline the P0 was really about.

The unit test now pins identification for both directory states; the confirmation branch is I/O, not logic worth mocking.

rubenfiszel:
Round 7 nits fixed in 12c16361c96285b031d900c2c594c3f138105e23 — both were real.

- **`--dry-run` disagreed with the push.** The gate sat below both dry-run returns (the JSON one and the pretty one), so a preview listed deletions a `--yes` push would then keep. The resolution now happens before the change list is printed or serialized: an unattended run (`--yes`, no TTY) or any dry-run settles it there, so the preview matches reality; a TTY push leaves the deletions in the list and asks after the user has seen it.
- **Contradictory doc comment** on `checkoutTracksDatatableMigrations`: it still claimed the directory's absence tells the two cases apart, which is exactly what round 6 established it cannot do. Rewritten to say what false actually means (ambiguous → confirm) and what true means (unambiguous deletion).

rubenfiszel:
Round 8 nits fixed in 9dc9189 — all three were real.

- **Split comment.** My earlier insertion anchored on the *middle* of the shared-UI comment, leaving `// Shared UI (the ui/ folder) is pushed out-of-band…` orphaned above the new block and `// and is excluded from the file diff…` dangling below it. The block moved above the comment, which is intact again.
- **Prompt counted files, not definitions.** A migration is one record across `.up.sql` + `.down.sql`, so "Delete 2 migration definition(s)" for a single migration was wrong in exactly the message a user makes a destructive decision from. Added `countDatatableMigrationRecords`, which dedupes by `(datatable, timestamp)`; both the prompt and the "keeping N" log now report records.
- **Stale push-safety description.** The PR body still described the round-5 drop-silently behavior. Rewritten to the current model: ambiguity is never guessed at — unattended runs and dry-runs keep the migrations and say so, a TTY push asks after showing the full change list.

rubenfiszel:
Round 9 nit fixed in bdb51aa: `countDatatableMigrationRecords` had landed between `untrackedDatatableMigrationDeletions`'s doc block and the function itself — the same comment/subject split as round 8, and this time it orphaned the "absent directory is ambiguous" rationale, which is the load-bearing part. The helper now sits above the whole doc block, so each comment is adjacent to what it documents.

rubenfiszel:
Round 10 P1 fixed in 686d73b — good catch, the directory heuristic was genuinely unsound.

`wmill datatable migrate new` creates `migrations/datatable/<dt>/`, so a pre-feature checkout that adds one local migration would have flipped the probe to "tracked" and then deleted every remote-only migration on the next `--yes` push. The working tree simply cannot answer "did this repo ever have that file".

Git history can. `gitRecordedDatatableMigrationPaths()` returns every path the repository has ever committed under `migrations/datatable/` (`git log --all --name-only`, one call, only computed when the push actually contains migration deletions), and the guard is now per-path rather than per-checkout: a recorded path's absence is a real deletion and flows through; a path the repo has never held is never deleted on a guess. No git history at all → nothing is trusted.

Verified against a scratch repo, all four states:

| state | recorded paths |
|---|---|
| fresh repo, no migrations ever | `[]` → all remote migrations protected |
| after `migrate new` (untracked file) | `[]` → **the hole you found is closed** |
| after committing it | `[the path]` → repo tracks it |
| after committing its deletion | `[the path]` → deletion still propagates |

That last row is what keeps round 6's fix intact: a committed deletion stays in history, so removing the last migration from git still deletes it on the workspace. Tests replaced to cover all four, including the `migrate new` case explicitly. `checkoutTracksDatatableMigrations` is deleted rather than left around.

rubenfiszel:
Round 11 addressed in 50fc5ce — all three were real.

- **P1, cross-branch history authorized deletions.** `--all` meant a migration that only ever existed on an unrelated branch vouched for deleting it from this one. Scoped to `HEAD`.
- **Shallow clones.** `fetch-depth: 1` truncates history, so "not found" meant nothing — and the message still claimed the repo had never tracked them. The result type is now `{kind:"known", paths} | {kind:"unknown", reason}`; a shallow clone (`git rev-parse --is-shallow-repository`) and a non-repository both return `unknown`, which trusts nothing *and* says why: "…shallow clone, so its history is truncated, so whether it ever tracked them cannot be established".
- **cwd below the git root.** `--name-only` prints repo-root-relative paths while the diff is cwd-relative, so every comparison missed. `--show-prefix` is stripped from each path (and non-matching paths dropped).

Verified against scratch repos:

| case | result |
|---|---|
| tracked on this branch | `known [path]` |
| deletion committed on this branch | `known [path]` → real deletion still propagates |
| orphan branch that never had them | `known []` → **cross-branch no longer authorizes** |
| `--depth 1` clone | `unknown (shallow clone, so its history is truncated)` |
| wmill root in a subdirectory of the repo | `known ["migrations/datatable/mydb/9_s.up.sql"]` → prefix stripped |

Tests updated to the new shape, including the `unknown` branch.

rubenfiszel:
Round 12 nit fixed in ab573bd: the two branches now carry their own remedy. `known` keeps "run `wmill sync pull` to track them in git, or delete them from the workspace"; `unknown` — which is the `actions/checkout` default — says to fetch the full history (`fetch-depth: 0`) so a real deletion can be told apart, since pulling doesn't help when the problem is truncated history.

rubenfiszel:
Round 13 nit fixed in 54d3833: the fetch-depth advice only fits the shallow case, so the remedy now travels with the cause. `unknown` carries `{reason, remedy}` and each producer sets its own:

- not a git repository → "Run the push from a git checkout of the synced repository"
- shallow clone → "Fetch the full history (for actions/checkout, fetch-depth: 0)"
- root unresolvable / history unreadable → "Check that git runs correctly in this directory"

The message composes reason + remedy, so no branch can hand out guidance that doesn't apply to it.

rubenfiszel:
Round 14 P1 fixed in d1735be. Sparse checkouts are the fourth way the working tree can disagree with history: `migrations/datatable/` stays recorded but is never materialised, so its absence would have read as a deletion. `core.sparseCheckout = true` now yields `unknown` ("this is a sparse checkout, so its working tree may not hold every tracked file" / "Include migrations/datatable in the sparse-checkout cone").

Deliberately conservative: any sparse checkout is treated as unattested, including a cone that does contain `migrations/`. Over-protecting costs a TTY confirmation; under-protecting costs migration definitions, and this is the fourth distinct bypass of a working-tree signal found in this PR.

Verified: a sparse clone of a repo with a committed migration returns `unknown`, while the same repo checked out normally still returns `known ["migrations/datatable/mydb/1_a.up.sql"]` — so real deletions are unaffected.

rubenfiszel:
Round 15 fixed in 8a86033 — both defects were in the guard I added last round.

- **Boolean parsing.** `git config --get` returns the raw string, so `core.sparseCheckout = 1` (or `yes`/`on`) sailed past `=== "true"` and the safeguard silently didn't fire. Now `git config --type=bool --get`, which normalises every git boolean form.
- **Remedy didn't resolve the state.** I flag *any* sparse checkout as unattested by design, so "include migrations/datatable in the cone" would have changed nothing. It now says "Run the push from a full (non-sparse) checkout".

Verified across every form: `true`/`1`/`yes`/`on` → `unknown` with the corrected remedy; `false` and an unset key → `known(1)`, so a normal checkout still authorises real deletions.

rubenfiszel:
Round 16 nit fixed in 34a292c: the `RecordedMigrationPaths` doc claimed `unknown` means "the history cannot be consulted at all", which is wrong for the sparse case — there the history reads fine and it's the working tree that doesn't mirror it. Rewritten around what the variant actually asserts (this checkout can't vouch for what it once held), naming both shapes and all five causes, with the caller's obligation stated once.

rubenfiszel:
✅ Review round clean @ 34a292c6231595a064f571a7a612a3556e25ef23

PR #10437, opened 2026-07-31
fix: honor on-behalf-of when a workflow step dispatches a script or flow

rubenfiszel:
## Summary

Three independent on-behalf-of defects found while diagnosing a report where a public script with "run on behalf of" set started running with the caller's permissions instead of its own, failing to read the variable it exists to read.

The main one: `handle_wac_v2_output` resolves the target runnable's on-behalf-of identity and then throws it away, pushing the child with the parent job's identity. So a script that opts into on-behalf-of runs as whoever called it when reached from a workflow-as-code step, while the same script reached through the run API runs correctly. Depending on which side is more privileged this either widens or narrows the child's permissions; any runnable using on-behalf-of to downscope is silently not downscoped.

## Changes

- `bun_executor.rs`: thread the `Option<OnBehalfOf>` returned by `script_path_to_payload` through the step-dispatch match instead of discarding it, and use it for the child push. The `flow` arm builds the same value from `on_behalf_of_email` + `edited_by` that `execution.rs` does. Module-relative (`./`) and `inline` steps pass `None`: they are the parent's own code, not a separate runnable. `created_by` still credits the caller, matching the run API, so the run page now shows the "permissioned as" line for these children. Python WAC routes into the same function and is covered.
- `utils_draft_deploy.ts`: send `preserve_on_behalf_of` on the script and flow draft-deploy branches. The app branch already did, with a comment explaining the hazard; scripts and flows were missed, so every draft deploy reset `on_behalf_of_email` to the deploying user via `resolve_on_behalf_of_email`. Still gated server-side by `can_preserve_on_behalf_of`, so a non-privileged deployer is unaffected.
- `variables.rs`: name the acting identity in the "Variable exists but you don't have access to it" 401, matching what the equivalent resource error already reports. Without it the error shows the grants but not who was denied, which is unreadable for on-behalf-of and token-scoped runs. Discloses strictly less than the resource version, which formats the whole `ApiAuthed`.

## Screenshots

The frontend change is request-payload logic in a deploy helper with no visible UI effect, so no screenshots.

## Test plan

- [ ] Run a WAC workflow whose step targets a script with `on_behalf_of_email` set, living in a folder the caller cannot read. The child job's `permissioned_as` / `permissioned_as_email` should be the target's identity, `created_by` the caller, and the variable read should succeed.
- [ ] Same with a `flow` dispatch step.
- [ ] Confirm `./` module-relative and inline steps still run as the parent.
- [ ] Deploy a script draft and a flow draft that carry an on-behalf-of email, as a `wm_deployers` member: the stored email should survive. Repeat as a plain member: it should flip to the deployer.
- [ ] Trigger the variable 401 and confirm the body names the acting username, email, groups and folders.

## Not addressed

Two related defects surfaced during the investigation and are deliberately out of scope:

1. `on_behalf_of_email` does not control permissions. `openflow.openapi.yaml` documents "will be run with the permissions of the user with this email", but `permissioned_as` is derived from the runnable's `created_by` / `edited_by`, so picking a specific user changes only the email. Fixing it means storing a real `permissioned_as` on script/flow, plus a migration and UI work. Triggers already store one.
2. A script version whose lock is pending or errored is invisible to `get_latest_script_hash`, which silently falls back to the previous locked version. A just-deployed on-behalf-of change therefore appears set in the editor while the runner still uses the old version. Flows have no such gate, which is why the two behave differently. Whether the runner should skip such a version or fail loudly is a behavior decision.

---
## Summary by cubic
Fixes on-behalf-of so workflow-dispatched scripts and flows run under their own identity. Auth errors now name the acting user, and draft deploys keep existing on-behalf-of.

- **Bug Fixes**
  - Workflow steps: run on-behalf-of scripts and flows under their own identity. Thread `OnBehalfOf` for `script`; compute it for `flow`. `./` module and `inline` run as parent. `created_by` remains the caller.
  - Draft deploys: keep existing on-behalf-of by sending `preserve_on_behalf_of` for scripts and flows when `on_behalf_of_email` is set (still gated by `can_preserve_on_behalf_of`). Tests added in `utils_draft_deploy.test.ts`.
  - Variable 401s: include “authed as” details (username, email, groups, folders).

<sup>Written for commit 75523c7765fcc1549cd4d1816b7dba9ec6e89b37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Both nits addressed.

**P2 (no regression test for the draft-deploy flag)** — fixed in 75d08cfa74. Three cases in `frontend/src/lib/utils_draft_deploy.test.ts`: script branch forwards `preserve_on_behalf_of: true` when the draft carries an `on_behalf_of_email`, script branch omits it otherwise, flow branch forwards it on `updateFlow`. 8/8 pass.

**P2 (WAC step target dispatched with no caller authorization check)** — dismissing, because it is pre-existing platform behavior rather than something this PR opens.

Flow steps take the identical shape today:

```rust
// windmill-worker/src/worker_flow.rs:6098
script_path_to_payload(
    &script_path,
    None,          // <- same missing db_authed
    db.clone(),
    &flow_job.workspace_id,
    Some(true),
)
```

and that path already consumes the returned `on_behalf_of` and pushes the child under the target's identity. So a user who can author a flow can already reach a runnable they cannot read and obtain its on-behalf-of identity; this PR brings WAC to parity with flows rather than creating a new class of elevation.

Adding the check to WAC alone would make WAC stricter than flows, which is inconsistent, and would break any existing workflow whose step references a runnable the caller cannot read. Whether step dispatch should authorize the caller's read access against the target is a real question, but it is a platform-wide decision that belongs in its own change covering both dispatchers, not a one-sided tightening bolted onto this fix.

Flagging it separately for @rubenfiszel rather than silently leaving it unstated.

rubenfiszel:
✅ Review round clean @ 75d08cfa74e70dde6430e9d2ba422c8c5a335d88 — nit-only verdicts at 97fb0ddf7d1b12f0cf1c9657adf98ba06fde3b65 (codex: Good to merge, pi: Good to merge, claude: nit-only); nits addressed in 75d08cfa74 / dismissed in the comment above

PR #10438, opened 2026-07-31
fix: make on_behalf_of control permissions for scripts and flows

rubenfiszel:
## Summary

For scripts and flows, `on_behalf_of_email` only ever supplied the **email**. The authorization identity was computed separately as `username_to_permissioned_as(script.created_by)` / `flow.edited_by`, and that is what `fetch_authed_from_permissioned_as` turned into username, is_admin, is_operator and folders for the RLS session. So picking "run on behalf of a specified user" changed the email but left the permissions as the deployer's.

The email was not inert: it fed the instance-superadmin lookup and `email_to_igroup`. The result was an incoherent hybrid — the target's superadmin flag and instance groups, with the deployer's username, workspace groups and folders. `openflow.openapi.yaml` documented the intended behavior ("The flow will be run with the permissions of the user with this email"), which the code did not implement.

This adopts the model triggers already use: **the permissioned_as is the only stored identity, and the address is derived from it.**

> [!WARNING]
> **Breaking, and it changes permissions at upgrade.**
> - `script.on_behalf_of_email` and `flow.on_behalf_of_email` are **no longer read**, but the columns stay and deploys keep writing them until every worker reports `1.776.0` (`MIN_VERSION_SUPPORTS_ON_BEHALF_OF_PRINCIPAL`) — a worker predating this release resolves runnables through those columns, and workers lag the server. A later release stops the write and drops them; that migration must re-run the backfill first. The API still returns `on_behalf_of_email` (derived), so clients are unaffected.
> - The migration **backfills** the principal from the address, which retires the old fallback. A runnable deployed before this change stops running as whoever last deployed it and starts running as the account its `on_behalf_of_email` always named — i.e. it starts doing what it always claimed to do. Rows whose address named nobody lose their on-behalf-of entirely.

## Changes

**Storage**
- Migration backfills `on_behalf_of` from the address — mirroring `users::permissioned_as_from_email` in SQL, including its superadmin and group branches — and does the same inside script/flow draft values. It leaves the address column in place for the compatibility window above; the down migration re-derives the address from the principal.
- Reads go through `ScriptHashInfo::on_behalf_of()` / `FlowVersionInfo::on_behalf_of()`, which resolve the address via the cached `get_email_from_permissioned_as`. Every dispatch site uses them: run-by-path, run-by-hash, run-wait-result, flow run, flow step dispatch, WAC step dispatch, schedules, the asset cascade, and the EE Slack/Teams/freshness/ducklake paths.

**Write path**
- `resolve_on_behalf_of` accepts either half. Only an address given → resolved to the account it names, and an address naming nobody is a 400 rather than a silently unrunnable identity. Both given and disagreeing → 400 (this check caught a real first-party bug during review). Not allowed to preserve → the caller's own identity.

**Read path** — `on_behalf_of_email` is still returned, derived in three places: `prefetch_cached_script` (the single funnel for script responses), the two flow-version handlers, and the workspace tarball. The tarball memoises per export, since the column is sparse and a workspace has few distinct principals.

**Simplifications this removed**
- `change_user_email` no longer touches script or flow: an address change does not move a username, so the principal is untouched. That deleted the sweeps and the six `g/`-vs-`@` guards that produced four review findings.
- Offboarding keys on the principal, which is unambiguous by construction — no guard needed to tell a user from a like-named group.
- The `NULL` → `created_by` fallback is gone; after the backfill there is no such row.

**Untouched** — apps keep both halves in `policy` (`on_behalf_of` + `on_behalf_of_email`) and are deliberately out of scope. The one guard retained is the app-policy one in `change_user_email`, precisely because apps *do* still store an address beside a principal.

**Unchanged on the wire** — the frontend and CLI needed no changes: responses still carry `on_behalf_of_email`, the CLI's `has_on_behalf_of` marker still derives from it, and the principal still never reaches the repo.

## Screenshots

The frontend change is a field rename (`on_behalf_of` → `on_behalf_of`), so the
selectors look the same; what is verified here is that they still round-trip through the renamed
field. Switching a script's identity to "me" in the editor and deploying stored
`on_behalf_of = u/admin` with the compatibility address `admin@windmill.dev` beside it, and the
reload below reads it back.

| script settings, after switching identity and redeploying | the identity picker |
|---|---|
| script | picker |

Flow settings, reading its stored principal back:

flow

## Test plan

- [x] `backend/tests/on_behalf_of.rs`: a recorded principal drives the job identity; an address-only deploy resolves to the same account; an address naming nobody is rejected; a mismatched pair is rejected; a flow update naming only the address keeps its principal.
- [x] `folder_default_permissioned_as`: a `g/wm_deployers` rule stores the group; a rule naming a raw address stores it verbatim.
- [x] `preserve_on_behalf_of` (9), `users` (3, incl. the group-collision guard for app policies).
- [x] `cargo check` CE + `--features enterprise,private` against EE ref `c71d604`; `npm run check:fast`; sqlx cache complete.
- [x] Migration verified against live data before applying: the same address resolved per-workspace (`u/alice` in one, `u/alice2` in another) and a legacy row went `NULL` → `u/alice`.
- [x] Live: the API returns the derived `alice@windmill.dev` next to `u/alice`, and the formerly-legacy script now runs with `permissioned_as = u/alice` instead of its deployer.

## Notes

`get_payload_tag_from_prefixed_path`'s flow branch still returns `None`, so a flow used as an error/recovery/success handler runs as the failing job's identity rather than its own. Left out deliberately: it changes already-configured handlers at upgrade, and belongs in its own change as the WAC equivalent did in #10437.


rubenfiszel:
Addressed in `3eb9e40`:

- **[Claude P1] stale generated system prompts** — `openflow.openapi.yaml` feeds `system_prompts/utils.py` as `OPENFLOW_SCHEMA`. Ran `python3 system_prompts/generate.py` (refreshing `flow.md`, `prompts.ts`, `skills/write-flow/SKILL.md`, `cli/src/guidance/skills.gen.ts`) and regenerated the minified copy the flow AI chat reads via `windmill-yaml-validator`'s `npm run gen` + `frontend/minifiedOpenflowJson.sh`. Verified the generator is idempotent on a second run. Good catch that this is on-topic rather than cosmetic — those prompts were still telling the model that `on_behalf_of_email` controls permissions.
- **[Codex P2] wrong omitted-principal documentation** — correct, and the two directions needed splitting rather than rewording. Response schemas now say the field is *absent on runnables deployed before it was recorded, which run as their last deployer*; request schemas say *omit it and it is derived from `on_behalf_of_email`*. Same split applied to `openflow.openapi.yaml` and the Rust request-field comments.
- **[Claude P2] `updateFlow` uncovered** — right: the mock returned `existsFlowByPath: false`, so deleting the clear from the update branch alone left the test green. It now drives all three branches (create-flow, update-flow, create-script) and asserts on each.

## [Codex P1] cached identities after a username rename — not taken, and not a regression

Real, but it predates this PR with identical severity, and the change makes the flow half strictly better.

`DEPLOYED_SCRIPT_INFO_CACHE` and `FLOW_INFO_CACHE` are plain `Cache::new(1000)` LRUs — no TTL, and `grep` finds no `.remove()`/eviction anywhere, only the `get`/`insert` pairs. They have always cached the identity:

| | before this PR | after |
|---|---|---|
| script | cached `created_by` → `u/{old}` after a rename | cached principal → `u/{old}` |
| flow | cached `edited_by` → `u/{old}`, **and `flow.edited_by` is never rewritten by the rename at all**, so the DB is stale too | cached principal → `u/{old}`, but the rename *does* rewrite the column (added here), so it self-heals once the entry ages out |

So the exposure is the same cache, the same key, the same lifetime — this swaps which cached string goes stale. Every other cached column (tag, timeout, `created_by`) has the same property, so a fix belongs in cache invalidation infrastructure (a `notify_*` channel like the one `IMPORTED_SCRIPT_HASH_CACHE` already uses) covering all of them, not in this PR.

The requested regression test would encode current behavior as correct, so I have not added one either.

## Still outstanding from both reviews

`--features enterprise,private` is green locally against `ee-repo-ref.txt`'s `5392616` (the EE branch rebased onto EE main `a786cd4`); CI runners carry a pre-bump EE checkout and can't verify it.

`wmill workspace merge` end-to-end: not run — it resolves its workspace from the git branch via `wmill.yaml`, which this checkout has no fixture for. Its identity-carrying path is `deployItem`, now covered by the unit test across all three branches, plus live cross-workspace API checks against two workspaces where the same person is `u/alice` / `u/alice2` (source principal → 400; cleared-key shape → stores `u/alice2`).

rubenfiszel:
`f687660`:

## [P1] external-superadmin principals — fixed

Correct, and it was an asymmetry in the function this PR introduced rather than anything pre-existing. `resolve_username_to_email` falls back to `password` for superadmins acting outside their workspaces; my inverse only searched `usr`, so the round trip broke in exactly one direction. The pair is *accepted* on write (the mismatch check goes through the forward function, which does resolve it), and then an email-only redeploy silently dropped it — reconstructing the hybrid identity.

`permissioned_as_from_email` now mirrors the forward function branch for branch:

```sql
SELECT COALESCE(
    (SELECT username FROM usr WHERE workspace_id = $1 AND email = $2),
    (SELECT COALESCE(username, email) FROM password WHERE email = $2 AND super_admin = true)
)
```

`COALESCE(username, email)` covers the disabled-`automate_username_creation` case, where the principal is the bare address — `username_to_permissioned_as` passes anything containing `@` through unchanged, so that round-trips too.

Regression added: an email-only preserve deploy naming `superadmin-external@windmill.dev` (in `password`, deliberately not in `usr` in the fixture) must store `u/superadmin-external`.

## [P1] deleted groups retain permissions — not taken here

Real, but it is pre-existing on four surfaces and this PR adds a fifth rather than creating it. `resolve_folder_default_permissioned_as` already writes `g/{group}` identities for:

| surface | call site |
|---|---|
| apps | `apps.rs:1883` |
| triggers | `windmill-trigger/src/handler.rs:513` |
| schedules | `windmill-api-schedule/src/lib.rs:74` |
| folder rules themselves | `folders.rs:23` |

`delete_group` (`windmill-api-groups/src/groups.rs:647-654`) removes only `usr_to_group` and `group_`, and `fetch_authed_from_permissioned_as`'s `g/` branch never checks existence — it takes the name and resolves folders from the ACL JSON, which still references it. So a deleted group already retains app/trigger/schedule folder access today.

Both candidate fixes are cross-cutting and change behavior for those existing surfaces: sweeping five tables plus app-policy JSON on group deletion, or rejecting nonexistent groups at auth time (which would start failing already-configured triggers rather than silently granting). Either is a defensible change; neither belongs in "make `on_behalf_of` control permissions for scripts and flows", and doing it half-way — for scripts and flows only — would leave the hole open where it already is.

Happy to open it as a follow-up. Flagging the auth-time option as the one I'd pick: it is a single place and fails closed for every kind at once.

rubenfiszel:
Both taken, in `9d2fbec`.

## [P1] synthetic group-email collisions

Right — the `group-*@windmill.dev` namespace is minted by `get_email_from_permissioned_as` but never reserved, so a real account can hold such an address. `permissioned_as_from_email` now resolves `usr` and the superadmin `password` fallback **first** and only falls through to the group interpretation when neither matches, so a real account always wins. A genuine group still round-trips: its synthetic address has no account behind it, so it reaches the group branch as before.

Regression added: a workspace user whose email is `group-ops@windmill.dev` while an `ops` group exists — an email-only preserving deploy must store `u/test-user-2`, not `g/ops`.

## [P1] forks retain principals for members they do not copy

Also right, and it is the same rule I applied to cross-workspace deploys earlier in this review — I just failed to carry it to fork cloning. A principal is workspace-scoped (`usr` is keyed by `(workspace_id, username)`), so copying one into a workspace with different membership can only produce an identity that names nobody.

`clone_scripts` and `clone_flows` now insert `NULL` for the column instead of copying it. The clone falls back to `created_by` — exactly what it did before the principal was recorded at all, so the fork regression you describe disappears rather than being traded for a different failure — and the fork re-establishes a real identity on its next deploy, where derive-from-email runs against the fork's own membership.

I deliberately did **not** apply the same treatment to `change_workspace_id` (`workspaces_extra.rs:387`): that renames a workspace and carries its `usr` rows along, so the principal stays valid there. The distinction is "does membership change", not "does the workspace_id change".

### Note on the pattern

Five of the seven P1s you have raised were real properties of this change and are now fixed; two (identity caches, deleted groups) I have declined with the evidence that they pre-date it on surfaces this PR does not touch, rather than expanding scope to half-fix them. If you disagree on either, the reasoning is in the two comments above and I will reopen them.

rubenfiszel:
All three taken, in `9f3d31b`. Each is the same field reaching a surface I had not followed it into.

**[P1] cloned drafts** — `clone_drafts` copied the value JSON verbatim, so a fork carried a principal naming a non-member and `deployDraft` then sent it as an explicit pair. It now strips the key from `script`/`flow` draft values while cloning, leaving the email so the deploy derives the fork's own principal. Same rule as `clone_scripts`/`clone_flows`, third surface.

**[P1] renamed users in drafts** — the rename rewrote the deployed columns but only `value.path` in drafts, so a draft naming `u/{old}` became undeployable (rejected as a pair naming nobody) where an email-only draft had been fine. It now rewrites `value.on_behalf_of_permissioned_as` alongside the path.

**[P1] `set-permissioned-as` and external superadmins** — the fix is to delete what I added rather than extend it. `lookupUsernameByEmail` only sees `usr`, so requiring it broke the very case the `password` fallback exists for. Both commands now send the email alone with `on_behalf_of_permissioned_as: undefined` and let the backend derive — which resolves members, groups and out-of-workspace superadmins alike. Net effect is less code than before this PR touched those commands.

For the record, this makes seven surfaces where the principal is workspace-scoped and had to be handled: deployed rows, drafts, the workspace-deploy request body (both `DeployProvider` adapters), fork clones, renames, email changes, and offboarding. The rule is uniform — *a principal is only meaningful in the workspace whose `usr` table defines it* — and the two remaining declines (identity caches, deleted groups) are the two places where the problem is not workspace scoping and pre-dates this change.

rubenfiszel:
Both taken, in `eb39245` — and you caught me overstating last round. I listed "renames, email changes, offboarding" as surfaces that handle the principal uniformly, but I had only swept *drafts* for renames. These are the other two.

- **email change** — the transaction now rewrites `on_behalf_of_email` and `on_behalf_of_permissioned_as` inside `script`/`flow` draft values alongside the deployed columns, so a draft cannot end up naming two different people.
- **offboarding** — same rewrite, keyed on the departing user's email and scoped to the workspace, since drafts are not confined to that user's paths and a draft on a shared path would otherwise keep running as them.

Regression added to the existing `test_change_user_email`, which already pins this exact class for schedules, app policies and folder rules: a `script` draft carrying the pair must come out holding only the new address. That test is the natural home — it exists because the same "identity reference left stale" bug has bitten those other surfaces before.

I have not added an offboarding regression: there is no offboarding test file to extend, and standing one up for a sweep that is textually the same statement as the two now covered would be scaffolding rather than a guard.

### Standing declines

Unchanged and still argued in the earlier comments: identity caches (no eviction path exists for *any* cached column, `created_by` included) and deleted groups (already open on apps, triggers, schedules and folder rules). Both are cases where the problem is not workspace scoping and pre-dates this change.

rubenfiszel:
Taken in `5d4e489`, and it reaches further than the two draft sweeps you cited — **the deployed columns have the same defect**, introduced by the same commit that added group support.

A runnable configured for group `ops` carries `{group-ops@windmill.dev, g/ops}`; that address validates because `get_email_from_permissioned_as("g/ops")` mints it. A user may also hold it — the collision case added in round 7. Every rewrite keyed on `on_behalf_of_email` therefore hits group-owned rows it has no business touching:

| site | before | consequence |
|---|---|---|
| `users.rs:1968` `UPDATE script SET on_behalf_of_email` | rewrote the email, left `g/ops` | mismatched pair → 400 on next deploy |
| `users.rs:1976` same for `flow` | ditto | ditto |
| `users.rs:2026` draft email sweep | ditto | ditto (your finding) |
| `offboarding.rs:842/884` script/flow | replaced **both** halves | group-owned runnable silently reassigned to the replacement user |
| `offboarding.rs:897` draft sweep | ditto | ditto (your finding) |

All six now carry `AND (… IS NULL OR … NOT LIKE 'g/%')`. The reasoning is the same in both operations and is stated once per file: a group's synthetic address does not change when a colliding user's email does, and a group-owned runnable does not run as a departing user, so neither operation should touch it.

The raw-email principal statements (`WHERE on_behalf_of_permissioned_as = $2`) need no guard — that column holds a bare address in the case they target, never `g/…`.

Regression added to `test_change_user_email`: with `test-user` holding `group-ops@windmill.dev`, a draft carrying `{group-ops@windmill.dev, g/ops}` must come out untouched while the ordinary user draft is rewritten. Both assertions live in the same test, so the guard and the sweep are pinned against each other.

rubenfiszel:
All five taken, in `7e3d163`. Claude's first nit was the important one.

## The group-collision regression was vacuous — thank you

The draft I added carried `group-ops@windmill.dev` while the address being changed was `test2@windmill.dev`, so the sweep's `WHERE value->>'on_behalf_of_email' = $2` never matched it and the assertion held with all six guards removed. Exactly the ceremonial test AGENTS.md warns about, and worse than none because it reads as coverage.

Replaced with `test_change_user_email_leaves_group_identities`, where the colliding user *is* the account being moved. Two scripts and two drafts share `group-ops@windmill.dev`; one pair is configured for `g/ops`, the other for `u/test-user-2`. After the move the group rows must keep the group's address and the user rows must take the new one. I verified it fails without the guard rather than assuming:

```
left:  [("u/test-user/g", Some("renamed@windmill.dev")), ("u/test-user/u", Some("renamed@windmill.dev"))]
right: [("u/test-user/g", Some("group-ops@windmill.dev")), ("u/test-user/u", Some("renamed@windmill.dev"))]
```

## [Codex P1] `g/alice@example.com` is a user, not a group

Correct. `VALID_EMAIL` permits `/` in the local part and `username_to_permissioned_as` passes an email-shaped username through verbatim, so a raw-email principal can begin with `g/`. My prefix-only guard classified it as a group: the email rewrite was skipped while the raw-principal rewrite still fired, leaving a mismatched pair.

Every guard is now `LIKE 'g/%' AND NOT LIKE '%@%'`, which mirrors `username_to_permissioned_as`'s own precedence — it tests for `@` first and only then applies prefixes.

## [Claude nit 2] the comment overstated, and the app sweep was unguarded

Both true, and the second is the substantive half: `UPDATE app SET policy = jsonb_set(…'on_behalf_of_email'…)` is email-keyed, and an app's `policy.on_behalf_of` can be `g/{group}` from a folder default. Guarded it rather than narrowing the comment, so the claim is now accurate.

## [Codex P2] offboard preview vs operation

Fixed — the preview's `obo_scripts` / `obo_flows` queries carry the same guard, so it lists exactly what will be reassigned instead of offering group-owned runnables the operation now skips.

## [Codex P2] raw-email principals undocumented

Fixed across `openapi.yaml` (5 descriptions), `openflow.openapi.yaml`, and the regenerated system prompts / minified OpenFlow / clients: the form is now documented as `u/{username}`, `g/{group}`, **or a bare email when the username is itself email-shaped**.

rubenfiszel:
Both right, same bug, fixed in `b0af399`. This one is entirely self-inflicted: I introduced it in `7e3d163` while fixing the previous round's finding.

Reshaping the guard from `(pa IS NULL OR pa NOT LIKE 'g/%')` to `NOT (pa LIKE 'g/%' AND pa NOT LIKE '%@%')` dropped the `IS NULL` arm, and SQL three-valued logic does the rest:

```sql
SELECT (NOT (NULL LIKE 'g/%' AND NULL NOT LIKE '%@%')) IS NULL;  -- t
```

`NULL`, not `true` — so `WHERE … AND <guard>` matched no row whose principal was unrecorded, i.e. every runnable deployed before this PR. `change_user_email` would have stopped rewriting their address and offboarding would have stopped reassigning them, which is a regression of behavior that was correct before I touched it.

All nine sites are back to an explicit `(… IS NULL OR NOT (…))`, with the reason noted where the rule is explained so the arm does not get optimized away again.

## The test gap that let it through

My round-11 test asserted only on rows that *have* a principal, so it was green with the bug present — the same shape of miss as the vacuous test one round earlier. It now carries a third row with `on_behalf_of_permissioned_as = NULL` alongside the group and user rows, all sharing the address being moved, and I verified the assertion catches the regression rather than assuming:

```
left:  [("…/g", "group-ops@…"), ("…/legacy", "group-ops@…"), ("…/u", "renamed@…")]
right: [("…/g", "group-ops@…"), ("…/legacy", "renamed@…"),   ("…/u", "renamed@…")]
```

Three rows, three outcomes, one statement: the group row is preserved, the user row moves, the legacy row moves. That is the whole contract of the guard in one assertion, and it now fails if any of the three is broken.

rubenfiszel:
✅ Review round clean @ 1a3321f2908b11578ad992ca007aad6bd571faa0 — nit-only verdicts at b0af3993079eea5f79f7500a565f4255caa690e2 (Codex: nits; Claude, Pi: good to merge); the overlong-comment nit is addressed in 1a3321f. All 17 CI checks pass on this head, including `check_ee`/`check_ee_full` against the pinned EE ref `5392616` — the EE half no CI reviewer could compile — and `check-freshness` for the regenerated system prompts.

rubenfiszel:
On the `ee-repo-ref.txt` finding at `7ab4c85e`: already fixed in that same commit — the review raced the push and reported against a head newer than the one it read. The merge did reset the pin to main's value, which dropped this PR's own EE companion commit; `7ab4c85e` merges EE main into the companion branch (windmill-ee-private#704, `c71d604..219b0b0`) and re-pins to `219b0b03`. The missing `windmill-trigger-azure` cache entry that came with it is regenerated.

On the rolling-upgrade window: this is a deliberate decision, not an oversight. A read-side fallback (derive the principal from the address when the principal is NULL) was implemented and then removed — server pods are only mixed for the minute or two a rollout takes, and carrying that fallback in the dispatch and GET paths for one release was judged not worth it. The exposure is bounded instead: the migration header requires the follow-up migration, the one that drops the column, to re-run the backfill *before* dropping, which repairs any row an old pod wrote. That makes the worst case "runs as caller until the next release" rather than permanent.

rubenfiszel:
Fixed the cached-address round-trip in `3a22f87`: `prefetch_cached_script`, `derived_on_behalf_of_email` and `resolve_folder_default_on_behalf_of` now use `get_email_from_permissioned_as_uncached`. That was a real inconsistency I introduced — the write path validates the pair uncached, so a cached read could hand the editor a live principal beside an address the account no longer holds, and the preserving redeploy would 400 until the entry expired. These three are the values that get round-tripped; the cached variant stays where the address only labels something.

On the mixed-version rollout window (fourth round): unchanged, and deliberate — see the reasoning in https://github.com/windmill-labs/windmill/pull/10438#issuecomment-5152091413. Server pods are mixed only for the minute or two a rollout takes, the read-side fallback was written and then removed as not worth carrying in the dispatch and GET paths for one release, and the exposure is bounded by requiring the migration that drops the column to re-derive first. `036b2cfc` also fixed the one place that would have made it permanent: the fork clone no longer clears an address whose principal was already absent in the source, since that address is what the later re-derivation needs.

PR #10441, opened 2026-08-01
[ee] refactor: stop reading the schedule's email column, deriving it from permissioned_as

rubenfiszel:
## Summary

Stacked on **#10440** (app policy), itself stacked on **#10438** (scripts and flows). Same model: **the permissioned_as is the only stored identity, and the address is derived from it.**

`schedule.email` was kept deliberately. #8439 dropped the address from all ten trigger tables and left schedule's behind "for backwards compat with old workers that don't know about permissioned_as". Nothing has used the value since — `push_scheduled_job` derives it from `permissioned_as` on every tick — while `set_enabled` wrote the **toggling user's** address next to an untouched principal. So the stored pair could already name two different people. It also is not the creator: that is `edited_by`, untouched here.

**This is phase 1 of 2.** It stops reading the column and makes every write derive from the principal, but **keeps the column**. Phase 2 drops it, and `MIN_VERSION_DERIVES_SCHEDULE_EMAIL` says when that becomes safe.

## Why the column stays, and how phase 2 gets scheduled

`get_schedule_opt` lists `email` in its SELECT on every released version, and it runs inside `commit_completed_job`'s transaction (`windmill-queue/src/jobs.rs:1364`) — **after** the `INSERT INTO v2_job_completed` and the queue-row delete, before the commit. Dropping the column would make that query fail on any not-yet-upgraded worker, rolling back the whole transaction: the job ran, but the database says it never completed and is still queued, so zombie recovery re-executes it. A rolling deploy would duplicate scheduled work.

The previous EE worker also passes the stored address straight to schedule error handlers, and API replicas on the previous version still serialize it from `schedules/get` and the workspace tarball. So the column is not just selected — its *value* is read. Every write here keeps it correct.

`MIN_VERSION_DERIVES_SCHEDULE_EMAIL` marks this release. Nothing gates on it at runtime; it exists because `vc()` asserts at compile time that a constraint stays above `MIN_KEEP_ALIVE_VERSION`, which is itself required to lag ≥50 minor versions. When this constraint stops compiling — *"Remove this obsolete constraint"* — no supported worker reads the column, and phase 2 is safe: drop it, delete the writes, delete the `change_user_email` sweep. The comment on the constant says exactly that.

## Changes

- `Schedule` loses its `email` field and every `SELECT … FROM schedule` drops it. A `ScheduleWithEmail` wrapper re-adds the **derived** address on the two read paths whose response shape carries one — `get_schedule` and the workspace tarball — so the wire shape is unchanged and no longer depends on the stored copy. It is `Serialize`-only: `Schedule` is flattened into it, and serde's flatten buffers through a representation `RawValue` (its `args`) cannot be read back from.
- `create_schedule` and `edit_schedule` keep writing the column, from the derived address.
- **`set_enabled` stops writing it.** It used to stamp the toggling user's address, which is how a schedule ended up with an address naming a different account than its principal. That is the behavioral fix in this PR.
- The tarball derives per schedule through the export's existing memo, so it is a handful of lookups rather than one per row.
- `with_derived_email` documents its authorization contract: it resolves an address for any `(w_id, schedule)` handed to it, so a caller acting for a user must already have established read access.
- EE: the schedule error handler derives the address it reports the failure as, instead of reading the column. The managed ducklake-maintenance rows keep writing it.

## Test plan

- [x] `preserve_on_behalf_of` (9), `folder_default_permissioned_as`, `schedule_push` (41). The existing assertions on the stored address still hold, and now pin write-through rather than a stored value a client could set.
- [x] New guard in `test_schedule_group_permissioned_as`: a schedule whose principal is `g/all` returns `email: group-all@windmill.dev` from `schedules/get` — a group has no address of its own, so the synthetic one can only come from deriving it.
- [x] `cargo check --tests` CE and `--features enterprise,private`; `SQLX_OFFLINE=true` clean for default, `ce_core` and the full EE set; `npm run check:fast`.
- [x] Live, with a `g/` principal: a schedule storing `g/testers` returns the derived address from `schedules/get`, and its queued occurrence lands `permissioned_as = g/testers`, `permissioned_as_email = group-testers@windmill.dev` in `v2_job`.

## Notes

No migration, and no frontend change: nothing in the UI reads `schedule.email`, and the CLI already deletes it from synced schedule files.

EE companion: https://github.com/windmill-labs/windmill-ee-private/pull/705


rubenfiszel:
✅ Review round clean @ 0508cc54bb45bbe7260acfbd51234042aeefd67f

rubenfiszel:
✅ Review round clean @ 9e0efa18819a9a99b35c10fb9a3df38f3fc93d73

rubenfiszel:
Superseded by #10455. This PR was auto-closed as merged when I force-pushed its head over its base branch by mistake; nothing was actually merged into main. #10455 carries the same commits, rebuilt on top of #10440.

PR #10446, opened 2026-08-01
fix: order script hard-delete and dbt graph publication locks consistently

rubenfiszel:
## Summary

Fixes [WIN-2274](https://linear.app/windmilldev/issue/WIN-2274). Review follow-up from #10326, deliberately left out of that PR because re-ordering locks in the delete path is not a change to land unreviewed.

Two writers touched `script` and the dbt graph sidecars in opposite orders:

- `persist_ingest` locks the `script` row (`SELECT … FOR UPDATE`), then replaces `dbt_node` / `dbt_edge` / `dbt_graph_snapshot`.
- Path deletion cleared the graph rows first, then deleted — and so locked — the `script` rows. Bulk deletion repeated the ordering.

Each could hold one side and wait on the other, so Postgres broke the cycle by aborting one transaction. The symptom was a rare failed delete or failed graph publish rather than a hang.

Fixed by taking the second option the issue proposes: drop the explicit clear and lean on the sidecars' foreign-key cascades to `script`. The cascade covers every graph sidecar, and the explicit clear was already redundant with it — the `DELETE FROM script` that follows fired the same cascade — so this removes an inverted lock rather than changing behaviour.

## Changes

- `delete_script_by_path` / `delete_scripts_bulk`: dropped the `clear_dbt_manifest` call; `dbt_node`, `dbt_edge` and `dbt_graph_snapshot` are reclaimed by their `ON DELETE CASCADE` FK to `script (workspace_id, hash)`.
- Both routes now clear `dbt_run_state` **after** the `DELETE FROM script` rather than before. It has no script FK so it still needs an explicit clear, but the deploy path (`create_script_internal`) locks the script row first and touches retry state later, so taking it ahead of the delete was the same AB/BA shape one line over.
- `delete_script_by_hash`: reordered `clear_dbt_manifest_version` ahead of `clear_static_asset_usage_by_script_hash`. Both graph publishers take the sidecars before the path's `asset` rows; this route took them the other way round, which deadlocks against the ingest that finds no warehouse identity. That route only soft-deletes the `script` row, so nothing cascades and its explicit clear stays.
- Removed the now-dead `clear_dbt_manifest` helper and updated the module's mutator contract.
- `docs/dbt-runtime.md`: recorded that hard-delete depends on the cascade, and which sidecars are and are not reachable by it.

`dbt_run_progress` is untouched: it has no key to `script` or to the job, is reclaimed only by its age sweep (`prune_run_progress`, called from every writer), and the delete path never cleared it.

## Test plan

- [x] Deadlock reproduced against Postgres with two lockstep sessions in the pre-fix order: `ERROR: deadlock detected … while deleting tuple in relation "dbt_node"`. With the delete going straight at `script`, both transactions commit and every sidecar is reclaimed.
- [x] FK/cascade definitions read from `pg_constraint` on a live database rather than assumed: all three graph tables carry `FOREIGN KEY (workspace_id, script_hash) REFERENCES script ON DELETE CASCADE`; `dbt_run_state` and `dbt_run_progress` have only a workspace FK. No RLS on any of them, so the cascade fires under the `user_db` role.
- [x] `deleting_the_script_cascades_to_every_sidecar` added — pins that deleting the `script` rows reclaims nodes, edges and markers for both a deployed version and a differing run snapshot. It replaces the test of the deleted helper.
- [x] `cargo test -p windmill-common --test dbt_graph_storage` — 7 passed, 0 failed.
- [x] `SQLX_OFFLINE=true cargo check` green for the workspace lib (`--features all_sqlx_features`) and for `windmill-common --all-targets`.

No regression test drives two transactions in lockstep. That needs two coordinated connections and would be flaky in CI, so the ordering invariant is held by the comments at the three call sites plus the cascade-coverage test above, which is what catches a sidecar added later without a cascade.

---
## Summary by cubic
Fixes WIN-2274 by aligning lock order between script hard-deletes and dbt graph publication to eliminate deadlocks. Hard-deletes now rely on `ON DELETE CASCADE` for graph sidecars, and retry state is cleared after the delete.

- **Bug Fixes**
  - `delete_script_by_path` and `delete_scripts_bulk`: delete from `script` first; drop `clear_dbt_manifest`; cascade reclaims `dbt_node`, `dbt_edge`, `dbt_graph_snapshot`.
  - Clear `dbt_run_state` after the `DELETE FROM script`; it has no script FK and still needs explicit cleanup.
  - `delete_script_by_hash`: clear `clear_dbt_manifest_version` before asset usage; keep explicit clear since this route soft-deletes.
  - Removed `clear_dbt_manifest` and updated the mutator contract; added cascade coverage test and docs noting cascade reliance.

<sup>Written for commit dbbe784b6cb032d2ca1069763206f25e6970c16b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ dbbe784b6cb032d2ca1069763206f25e6970c16b

PR #10448, opened 2026-08-01
feat: give dbt its own editor with an explicitly refreshed model graph

rubenfiszel:
## Summary

Follow-up to #10326. dbt shipped inside the generic script editor, which accommodates a project bundle by bolting a file tree onto its module tab strip — but its premise (one file, one signature, one body) is the wrong one, so dbt was withheld from the language picker behind `localStorage.dbt_language`.

This gives dbt its own editor and an explicitly refreshed model graph, and removes that gate.

A dbt script is a project: a file tree, a descriptor, run arguments and a model graph. The editor is those four. The graph is drawn from a manifest **dbt itself produced** — at deploy, or now by a `dbt parse` of the buffer being edited, on demand. Never from a `ref()` scan of our own: that yields names without relations (a node is `dbt://<warehouse>/<schema>/<table>`, and schema, database and alias come from configs layered with the profile) and disagrees with dbt about `enabled`, macro-built refs, loops and packages — two graphs of different fidelity in one product.

## The one structural change

Graph rows are keyed to a deployed version, deliberately, on both sides: `GraphPublisher::Unversioned` never publishes, and `get_dbt_run_graph` pins only when the job names a version.

A buffer parse is a third provenance — not the version's graph, not a run's snapshot of a version. The buffer differs from what is deployed (that is the point of refreshing it) and a project being written may have no deployed version at all, so its rows carry **no version** and are keyed to their own preview job, readable only back through that job id and never through the path. That keeps the property the `Unversioned` guard exists for: a parse publishes no path-keyed `asset` usages, so a `jobs:run` principal still cannot restate what a deployed project's graph says.

`script_hash` therefore became nullable on `dbt_node` / `dbt_edge` / `dbt_graph_snapshot`, and each primary key became two partial unique indexes — versioned rows keyed by version, editor rows by job alone. The composite FK to `script` is unchanged: `MATCH SIMPLE` is satisfied by a NULL, so a versioned row still cascades with its version and a version-less one is outside its reach. A partial arbiter must be named, so the marker's `ON CONFLICT` now repeats `WHERE script_hash IS NOT NULL`.

Retention: the newest 5 parses of one script **by one principal** keep their graph, dropped as each refresh lands. The principal is load-bearing rather than incidental — a preview's *path* is chosen by a caller who needs only `jobs:run`, so a count bounded per path alone would be a way to retire someone else's refreshes. `permissioned_as` is the execution principal the queue derived, which is why `dbt_run_state` keys on it too. The instance-wide age sweep every dbt run already performs still catches one refreshed once and left.

## Changes

**Backend**
- Migration `20260801121717_dbt_editor_preview_graph`: nullable `script_hash`, two partial unique indexes per table.
- `dbt_manifest.rs`: `replace_dbt_editor_graph` — version-less rows keyed to a parse job, idempotent by job, bounded per path. **No** digest suppression: the editor pins to its own job, so a buffer that happens to agree with the deploy must still leave something to pin to.
- `DBT_COMMANDS` gains `parse`. Not a run-form variant, exactly like `show` — each is a thing you do to the project in front of you — but accepted from a flow, the CLI and the API, which is what makes the refresh scriptable and testable.
- `handle_dbt_job` grows a parse-only branch: deps → parse → ingest, no build, no materializations, no retry state, no ownership. It uses `strict: false` like the deploy's own parse, so a `{{ placeholder }}` only a run can fill does not block a refresh. A job that *does* name a version writes an ordinary per-run snapshot instead.
- Read path: `PinnedRun.script_hash` is now optional; `live` joins with `IS NOT DISTINCT FROM`; the pinned scope keys on the pin rather than on the hash (an editor parse's models are precisely the ones `asset` has never heard of); `script_visible` is true for a version-less row, which is reachable only through the caller's own parse job. New `dbt_graph_ingested_at` on the graph response.
- `get_dbt_run_graph` pins a preview job **only when it parsed a graph of its own**; every other preview keeps answering with the workspace graph.

**Frontend**
- New `DbtEditor.svelte`, rendered by `ScriptBuilder` in place of `ScriptEditor` for dbt: project header (path, engine, warehouse), file tree, per-file Monaco, and a Models | Run pane over the job log. `Build <model>` when a model file is open, `Build project` otherwise — the bundle always travels, since dbt cannot run a subset of it.
- New `DbtModelGraph.svelte`: the graph, `Refresh models`, the provenance label and the empty state. The refresh carries the run form's `vars` (they steer `enabled`, schemas, aliases and relation identity, so a parse without them describes a different project than the run would build) and the script's worker tag and timeout, so a project reaching a private network parses on the worker that can reach it.
- The generic `ScriptEditor` loses its dbt branches; the shared file rules move to `dbt/projectFiles.ts`.
- The language gate (`processLangs`, `DefaultScriptsInner`) is gone. The **Generated UI** settings tab is hidden for dbt — that schema is derived server-side by `dbt_arg_schema` and anything refined there is overwritten by the next deploy.
- `infer.ts` reaches `parse_dbt` through the module namespace rather than a named import. This is a **pre-existing break from #10326**: `parse_dbt` is newer than the published `windmill-parser-wasm-yaml@1.770.0`, and a named import of a missing export is a link-time failure that takes the whole module down — every language's inference with it. It was unreachable while dbt was gated; removing the gate makes it the first thing a user hits. A package predating `parse_dbt` now degrades to "no schema inferred for dbt" instead. **Publishing a newer parser package is still worth doing separately.**

## Settled up front

- **A real `dbt_command: "parse"`**, so the refresh is scriptable and testable rather than a UI-only affordance.
- **A parse needs a resolvable warehouse** — `write_profiles` runs before it — so a misconfigured project fails the refresh the way it would fail a run. The empty state says why, and a failed refresh shows dbt's own message with a link to the job.
- **The provenance is labelled**: "parsed from the editor at 01:18 PM" against "as of last deploy", from `dbt_graph_ingested_at`. The two graphs are drawn identically, so leaving it unsaid would move the ambiguity the explicit refresh removes straight into the editor.

## Screenshots

The editor: project header, file tree, descriptor, Models pane.

02-dbt-editor

`Refresh models` on a project that has never been deployed or built — two models and their `ref()` edge, drawn from a parse of the buffer, labelled with where they came from.

03-refreshed

Selecting a model shows its SQL and jumps to its file.

04-node-selected

A warehouse that does not resolve fails the refresh, and the panel says why.

05-parse-failed

dbt back in the language picker, with no localStorage flag.

01-lang-picker

The Run tab: the descriptor's derived command block over the job log.

06-run-tab

Switching a dbt draft to another language leaves no dbt arguments behind — `inferSchema` honors `nlang` and clears arguments before the parse rather than after it, so a failed parse of the next language's template cannot hand dbt's command block to the editor that mounts next.

07-switched-to-python

## Test plan

Verified with real jobs against a local Postgres warehouse and dbt-core 1.12.0, on both worker kinds.

- [x] `parse` preview job on a DB-backed worker: 3 nodes / 2 edges ingested, rows land with `script_hash IS NULL` under the preview job id, `asset` untouched (0 rows — no ownership published).
- [x] `GET /jobs/dbt_graph/{parse job}` returns the models, their SQL, the `ref()` edge, the data-test badge, `dbt_snapshot_job` and `dbt_graph_ingested_at`; `GET /assets/graph` returns **nothing** for the same rows — the path guard holds.
- [x] Deploy still ingests the version's graph and publishes ownership (this is what the `ON CONFLICT ... WHERE script_hash IS NOT NULL` fix covers — without it every deploy 42P10'd).
- [x] `build` run materializes real tables; a `parse` of the deployed version stores no snapshot when its graph matches the deploy.
- [x] Agent worker (`ag-agentgrp-…`, `Connection::Http`): same parse publishes through `/api/agent_workers/dbt_graph/{workspace}` and reads back identically.
- [x] A parse whose `profile.warehouse` is not configured fails with `dbt warehouse 'nope' is not configured for this workspace (configured: main)`.
- [x] Playwright: create a dbt script from the picker, add a nested model, type a `ref()`, refresh, inspect a node, break the warehouse and see the reason.
- [x] A parse carrying `vars` runs and records `permissioned_as` on its marker; a redeploy ingests v2's graph and republishes ownership while v1's graph stays intact.
- [x] Another principal's parses under the same path do not retire this one's graph (`one_principal_cannot_evict_another_s_editor_graphs`).
- [x] `EXPLAIN` confirms the unpinned join still folds to a plain `script_path`/`script_hash` equality — the gated NULL arm costs the hot path nothing — while a pinned version-less read still matches its rows.
- [x] Switching a populated dbt editor to Python leaves no dbt `command` argument in the run form.
- [x] `cargo check --all-targets`, `cargo check --features enterprise,private`, `SQLX_OFFLINE=true` lib + test targets, `npm run check` (0 errors), 17 dbt storage/read tests.

## EE companion

windmill-labs/windmill-ee-private#707 — `publish_dbt_graph` accepts a version-less graph from an agent worker.


rubenfiszel:
Nit on the agent-worker payload dismissed with reasoning: the EE handler does not take `permissioned_as` from the payload — it reads it from the `v2_job` row it already verified, alongside `runnable_path`, so an agent cannot name another principal (or another script) there. `windmill-api-agent-workers/src/ee.rs` is a gitignored symlink into `windmill-ee-private`, so it isn't visible in this repo's diff; the change is in windmill-labs/windmill-ee-private#707.

The tab-switch nit was real and slightly worse than reported — the swap also unmounted the log panel, so switching to Models during a build lost the running job's log view. Both panes now stay mounted and are hidden instead, and Build switches to the Run tab so a build's output is where you're looking. Fixed in 52140c4c985d49e0fe3d49193c3f460f62745653.

rubenfiszel:
✅ Review round clean @ 9e2544d5f49229468fe455f54e6d6b4c9798f82e — nit-only verdicts at 72e49c5344d28c5ac057582077675a3b1a4b4526 (Claude and Pi: Good to merge; Codex: mergeable with nits); nit addressed in 9e2544d5f49229468fe455f54e6d6b4c9798f82e.

The Generated UI tab was hidden for dbt but its `TabContent` was not, so a `customUi` disabling Metadata and Runtime initialises `selectedTab` to `ui` and would have opened `ScriptSchema` for a dbt script with no trigger above it. The content is now gated too, and an effect falls back to whichever tab IS offered — the initialiser runs before the language is known, so it cannot decide this itself.

Final validation on this head: `npm run check` 0 errors, `cargo check --all-targets` clean, 17 backend dbt storage/read tests, 24 frontend dbt unit tests.

rubenfiszel:
✅ Review round clean @ 285d7c6df4f4f3e2715a38e19f888e76ec10de1c — nit-only verdicts at eb9f5b1cfe70d2c3850ec8886e4ffbfc275503e0 (Pi: Good to merge; Codex and Claude: mergeable with nits); both nits addressed in 285d7c6df4f4f3e2715a38e19f888e76ec10de1c.

- The dbt tab fallback walked straight to `triggers`, which a `customUi` may have disabled — it now follows the same chain the initialiser does, minus `ui`.
- `dbt_graph_snapshot.permissioned_as` was missing from `summarized_schema.txt`.

This head is the merge of `main` (1.776.0 + on_behalf_of permissions) plus those two fixes. Validation on it: `cargo check --all-targets` and `--features enterprise,private` clean, 17 backend dbt tests, 30 frontend unit tests, `npm run check` 0 errors. EE companion re-pinned at `7e424384` after merging EE main.

rubenfiszel:
✅ Review round clean @ 2a9968c71c172fe903bf144a027fd2b24a3ebd59

PR #10451, opened 2026-08-01
fix: repair the two CI jobs that fail on a release tag

rubenfiszel:
## Summary

Two workflows that only trigger on `v*` tags went red on `v1.776.0` ([backend-test-windows](https://github.com/windmill-labs/windmill/actions/runs/30713348717), [check-docs-links](https://github.com/windmill-labs/windmill/actions/runs/30713348758)). Neither failure is a product defect — one is a test that assumed a POSIX path separator, the other a docs link that is correct but points at a page that has not deployed yet.

**Windows integration tests.** `dbt_profiles::tests::postgres_forwards_its_root_certificate` hardcoded a forward slash in the expected `sslrootcert` value, while `render_profile` builds it with `profiles_dir.join(ROOT_CERT_FILENAME)`. On Windows that renders `"/tmp/p\\server-ca.pem"`, so the emitted profile was right and only the assertion was wrong (342 passed, 1 failed). The assertion now goes through the same `Path::join` + `yaml_scalar` pair, so it matches by construction on either platform while still pinning that the value is `profiles_dir`-joined, escaped rather than bare, and at the right key — a relative `ROOT_CERT_FILENAME` still fails it.

**Docs link check.** `https://www.windmill.dev/docs/getting_started/scripts_quickstart/dbt`, referenced from `DbtSettings.svelte`, 404s because the page exists only in the unmerged windmill-labs/windmilldocs#1625. The slug in the app is already the final one, so rather than retarget a link that is about to be correct, the checker gets a small `PENDING_DEPLOY` allowlist.

The allowlist is deliberately narrow in both directions, so it cannot become a permanent exemption:

- It suppresses **only a 404**, the one answer that means "not published yet". A timeout, 403, 429 or 5xx on the same URL is a real fault and still fails the run — suppressing those would also read as "still waiting" and defer the staleness check below by a release cycle.
- An entry that outlives its reason — the page went live, or nothing in `frontend/src` references the URL any more — **fails** the job and names the entry to delete. A line in a green log would never be read at release time, so nothing would force the entry out.

## Changes

- `backend/windmill-worker/src/dbt_profiles.rs`: build the expected `sslrootcert` value with `Path::join` + `yaml_scalar` instead of a hardcoded `/`, so the assertion holds on Windows.
- `.github/scripts/check-docs-links.mjs`: add `PENDING_DEPLOY`, an exact-URL map of links awaiting a docs deploy, with the reason as the value. A suppressed URL is reported as `⏳ waiting on a docs deploy`.
- `.github/scripts/check-docs-links.mjs`: restrict the exemption to a 404 response, so every other failure mode on an allowlisted URL still fails.
- `.github/scripts/check-docs-links.mjs`: fail the run on a stale entry — page live, or URL unreferenced — naming what to delete.
- `.github/scripts/check-docs-links.mjs`: reword the success line to `No broken docs links (N checked)`, since `All N docs links are reachable` is untrue once a 404 is suppressed.

## Test plan

- [x] `cargo test -p windmill-worker --lib dbt_profiles::` — 10 passed, 0 failed (Linux; the Windows rendering is what the next tag run exercises)
- [x] `node .github/scripts/check-docs-links.mjs` — 174 links checked, dbt link reported `not live yet (404)`, exit 0
- [x] Allowlisted URL returns 503 — exit 1, listed as a broken link
- [x] Allowlisted URL times out (status 0) — exit 1, listed as a broken link
- [x] Allowlisted page live (200) — exit 1, `the page is live`
- [x] Allowlisted URL unreferenced — exit 1, `nothing references it`, and the dbt link reverts to a hard failure, confirming suppression is exact-key
- [ ] Next `v*` tag: both workflows green

Once windmilldocs#1625 deploys, the next run of this job fails with `the page is live`, and the `PENDING_DEPLOY` entry should be deleted.

No frontend source changed (the `.mjs` is a CI script), so there is no UI surface to screenshot.

### DIFF
--- a/.github/scripts/check-docs-links.mjs
+++ b/.github/scripts/check-docs-links.mjs
@@ -15,6 +15,17 @@ const CONCURRENCY = 24
 const TIMEOUT_MS = 20000
 const RETRIES = 2
 
+// Links whose target page is written but not yet deployed on windmill.dev: the app
+// link is already the final slug, so a 404 is expected until the docs side ships.
+// Every entry is still fetched — one that starts resolving is reported so the entry
+// gets dropped, because a stale entry silently hides a genuinely dead link.
+const PENDING_DEPLOY = new Map([
+	[
+		'https://www.windmill.dev/docs/getting_started/scripts_quickstart/dbt',
+		'windmilldocs#1625 (dbt runtime quickstart)'
+	]
+])
+
 async function walk(dir) {
 	const out = []
 	for (const entry of await readdir(dir, { withFileTypes: true })) {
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 139fa9592c6a50451883678cc843e1e179b7dde2 — nit-only verdicts at 9310fb94f441748247f9f32235dbb37236324990 (codex + claude: mergeable/nits, pi: good to merge); the single nit both raised, that `PENDING_DEPLOY` suppressed every failure mode rather than the 404 the entry claims, is addressed in 139fa9592c. Pi argued the narrowing was not worth it since an allowlisted URL shares a host with the rest; going with the majority, because the case that actually costs something is path-specific — a 429 or timeout on that one URL after the docs deploy would read as "still waiting" and defer the stale-entry failure by a release cycle. Verified 503, timeout, 200-live and unreferenced all exit 1, and a plain 404 still exits 0.

PR #10456, opened 2026-08-02
fix: report a missing worker tag instead of spinning in data table UIs

rubenfiszel:
## Summary

Fixes WIN-2276.

Every data table / Database Manager operation runs as a Windmill job on a native tag
(`postgresql`). On an instance whose worker groups only declare the default tags, that job
is queued and never picked up — and the UI had no way to say so:

- `pollJobResult` (behind `dbOps`, the SQL REPL, table metadata, DDL) only advanced its
  retry counter on *exceptions*, so a job that stays queued polled forever.
- `getDbSchemas` used the blocking `run_wait_result` endpoint, which hangs for the full
  server timeout (10 min by default) and then reports `timeout after 600s`.
- `DBManagerContent` rendered a spinner whenever the schema was undefined, with no error
  branch — so even after the failure landed the drawer kept spinning.

The result was the reported symptom: an endless spinner on
`/workspace_settings?tab=windmill_data_tables&dbm=datatable~…`, with nothing pointing at
the missing worker.

## Changes

**Detection (`frontend/src/lib/components/jobs/missingWorker.ts`)**
- `missingWorkerTagOfQueuedJob(workspace, jobId)` reads the queued job's *actual* tag and
  asks `WorkerService.existsWorkersWithTags` whether any worker serves it. Never throws:
  an inconclusive lookup means "keep waiting", so an instance with `TAGS_ARE_SENSITIVE`
  (empty response) never gets a wrong diagnosis.

**Poller (`frontend/src/lib/components/jobs/utils.ts`)**
- A job that has never started is probed at 10s, then every 40s. A worker group booting is
  indistinguishable from an unserved tag in any single probe, so it takes **three
  consecutive empty probes (~90s)** before the caller is told. A running job is never
  touched, and no single negative reading acts on anything.
- What happens then depends on whether the job writes, because the two failure modes pull
  in opposite directions:
  - **Reads** (schema, metadata, count, select) stop waiting and report the tag, leaving the
    job *queued*. Cancelling them would delete the very backlog the autoscaler scales up on;
    with a 300s cooldown, a group returning from zero would never recover.
  - **Writes** (insert/update/delete, DDL, REPL SQL, migration SQL) are **never abandoned**.
    Reporting one as failed while it stays executable invites a duplicate retry, and
    cancelling it first cannot be done atomically from the client: a worker can claim it
    between the tag probe and the cancel, and a soft-cancelled statement may already have
    committed. Instead `onNoWorkerForTag` toasts why the operation is waiting, and the poll
    continues until the job actually resolves.
- The poll backs off (375ms → 750ms → 2s) instead of a fixed 375ms tick, so a minutes-long
  introspection or DDL no longer costs hundreds of round-trips.
- `failIfNoWorkerForTag: false` opts out; the Worker REPL uses it because its tag is a
  worker-name prefix pulled directly by the shell loop, not advertised in `worker_ping`.

**Reporting**
- `DBManagerContent` replaces the eternal spinner with an error panel + Retry, fed by both
  the schema and the table-metadata query and titled for whichever failed. The error branch
  is ordered **before** the cached-schema branch: `dbSchemas` is a store that survives a
  failed refetch, so the other order would hide a failure behind stale content.
- After 10s of either query loading, the drawer shows the missing-worker warning inline
  while the job is still queued, so the cause is visible long before the poller gives up.
- `getDbSchemas`, `loadTableMetaData` and `fetchSnowflakePrimaryKeys` moved from
  `runScriptPreviewAndWaitResult` to `runScriptAndPollResult`, so they inherit the check
  instead of blocking on the server-side timeout.
- `DataTablePreview` / `DucklakeResultPreview` distinguish "metadata read failed" from
  "table doesn't exist yet" instead of showing the latter for both.
- `loadAllTablesMetaData` no longer toasts (it always rethrows); each caller renders the
  error in its own pane, so a failure is reported once rather than twice. `DBTable`'s
  cell-edit path surfaces the message instead of discarding it.
- `AppDbExplorer` clears `columnDefs.loading` when metadata fails, instead of leaving the
  component's own spinner up forever.

**Proactive warning (`MissingWorkerTagAlert.svelte`)**
- Shown on the Data tables settings tab, in the migrations modal, and in the DB manager's
  slow-load state, so the question in the ticket ("will data tables work without a native
  worker?") is answered before anything is clicked. Worded as "no worker is running for
  this tag" rather than "this tag is not configured", since a correctly tagged group at
  zero replicas looks the same in `worker_ping` and queueing a job is what scales it up.
  Skipped when per-workspace default tags are on — the effective tag is then
  workspace-suffixed and resolves through fork ancestry, so probing the bare tag could warn
  about a tag that is in fact served.

## Screenshots

Data tables settings, on an instance whose worker group does not declare `postgresql`:

settings

Database Manager after 10s — the cause is named while the job is still queued:

db manager hint

After the confirmation window, for a read (job left queued):

db manager error

A write is never abandoned: this insert waited out a missing worker, then completed on its
own once the tag was served again.

write resolved

A failed *refresh* of an already-loaded manager reports the error instead of leaving the
stale schema on screen:

cached schema error

Migrations modal:

migrations

After adding `postgresql` to the worker group, unchanged behavior:

db manager healthy

## Test plan

- [x] `npx vitest run src/lib/components/jobs/missingWorker.test.ts` — 5 passed: a read
      whose tag stays unserved is reported without being cancelled; a write is never
      abandoned, reports once why it waits, and still resolves; a run of empty probes that
      recovers never gives up; a job queued behind a busy worker keeps waiting and still
      resolves; an inconclusive tag lookup counts as "served"
- [x] Removed `postgresql` from the worker group's tags: the DB manager names the tag
      inline at ~10s and reports the terminal error after the confirmation window
- [x] Confirmed in the DB that the *read* jobs from that attempt stay queued and uncancelled
      (`v2_job_queue`), so the backlog the autoscaler needs survives
- [x] Inserted a row on the same instance: the `WM_INTERNAL_DB_INSERT` job stays queued and
      uncancelled while the UI explains why. Restoring the worker tag then let that exact
      job run to `success`, and the grid picked the row up on its own — the write is neither
      lost nor duplicated
- [x] Loaded the manager successfully (schema cached), then removed the tag and hit Refresh
      — the error panel replaces the stale manager rather than hiding behind it
- [x] Data tables settings tab and migrations modal show the warning banner
- [x] Re-added `postgresql`: schema, table list, row reads, `CREATE TABLE` via migration
      and the migrations list all work as before

## Not covered

The `run_datatable_migrations` endpoint waits server-side (`run_wait_result_internal`), so
a migration run on an instance without a `postgresql` worker still blocks until the server
timeout. The banner in the migrations modal explains the cause up front; changing the
job-waiting semantics in the backend was left out deliberately, since a fail-fast there
would also break autoscaling-from-zero for every other `run_wait_result` caller.

### DIFF
--- a/frontend/src/lib/components/DBManagerContent.svelte
+++ b/frontend/src/lib/components/DBManagerContent.svelte
@@ -94,7 +94,13 @@
 	// owns its slot so neither can clear the other's error on a refetch.
 	let schemaError = $state<string | undefined>(undefined)
 	let colDefsError = $state<string | undefined>(undefined)
-	let loadError = $derived(schemaError ?? colDefsError)
+	let loadError = $derived(
+		schemaError
+			? { title: 'Could not load the database schema', message: schemaError }
+			: colDefsError
+				? { title: 'Could not load the tables of this database', message: colDefsError }
+				: undefined
+	)
 
 	let colDefs = resource(
 		() => [input, ws],
@@ -158,7 +164,7 @@
 	const SLOW_LOAD_MS = 10_000
 	let slowLoad = $state(false)
 	$effect(() => {
-		if (!dbSchemasPromise.loading) {
+		if (!isLoading()) {
 			slowLoad = false
 			return
 		}
@@ -200,8 +206,8 @@
 {#if loadError}
 	<div class="h-full w-full flex flex-col items-center justify-center gap-3 p-8">
 		<div class="max-w-2xl w-full flex flex-col gap-3">
-			<Alert type="error" title="Could not load the database schema" size="xs">
-				{loadError}
+			<Alert type="error" title={loadError.title} size="xs">
+				{loadError.message}
 			</Alert>
 			<div class="self-start">
 				<Button size="xs" color="light" startIcon={{ icon: RefreshCcw }} on:click={() => refresh()}>
--- a/frontend/src/lib/components/DBManagerContent.svelte
+++ b/frontend/src/lib/components/DBManagerContent.svelte
@@ -1,7 +1,9 @@
 <script lang="ts">
 	import { dbSchemas, workspaceStore, type DBSchema } from '$lib/stores'
-	import { sendUserToast, sortArray } from '$lib/utils'
-	import { Loader2 } from 'lucide-svelte'
+	import { sortArray } from '$lib/utils'
+	import { Loader2, RefreshCcw } from 'lucide-svelte'
+	import Alert from './common/alert/Alert.svelte'
+	import Button from './common/button/Button.svelte'
 	import { dbSupportsSchemas } from './apps/components/display/dbtable/utils'
 	import DbManager from './DBManager.svelte'
 	import {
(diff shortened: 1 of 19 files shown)


rubenfiszel:
Re: `DEFAULT_TAGS_WORKSPACES` allowlist in `MissingWorkerTagAlert` (round 2, P2 #2) — keeping the blanket skip on purpose.

The allowlist is a global setting, and the effective tag for an included workspace is not derivable client-side: `per_workspace_tag` resolves through fork ancestry and can append `-fork`, so a probe of `postgresql-<workspace_id>` would miss the ancestor-suffixed form and warn about a tag that is in fact served. A false "no worker serves this tag" banner on a correctly configured instance is worse than no banner, and there is no endpoint that exposes the resolved tag without pushing a job.

The cost of the skip is bounded: it only suppresses the *proactive* banner, on instances that have per-workspace default tags enabled globally. The in-flight check in `pollJobResult` reads the queued job's actual tag, so those workspaces still get the accurate diagnosis on the operation itself — which is the path the ticket is about.

Re: the waiting hint not covering a stuck metadata job (round 2, P2 #3) — `slowLoad` now watches both resources (fc1c6fa). The remaining sub-case is a *cached* schema plus a stuck metadata job, where the loaded manager renders and the hint has nowhere to go; that one surfaces through the error panel instead, which now takes precedence over the cached schema and is titled for the metadata failure specifically.

rubenfiszel:
✅ Review round clean @ 021222caaf290536503f5c45cf94077fc6165117 — nit-only verdicts at 021222caaf290536503f5c45cf94077fc6165117 (codex + claude: mergeable with nits, pi: good to merge). Both nits were the same one: the PR description still documented the cancel-then-report contract that 021222c replaced. The body was rewritten before those reviews landed and now describes the shipped behavior (writes are never abandoned or cancelled), including the test-plan line about the insert job. Head is unchanged since the round.

PR #10457, opened 2026-08-02
feat: add session recording to wmill app dev

rubenfiszel:
## Summary

`wmill app dev --recording` records a replayable session of the raw app you are building locally, without deploying it first.

The recorder itself is not re-implemented: `frontend/src/lib/components/recording/` (the raw-app recorder shipped with the app editor and the `/folders` publish flow) is bundled into the CLI by `cli/generate-dev-recorder.ts`, so a dev-server recording is the same format the `/replay` player, `/pipeline_replay` and the Hub player already read. The one adaptation is the `$state` rune: outside Svelte the store's fields are read through its getters, so the generator strips it and the values become plain `let`s. It refuses to generate on any other rune rather than emit a bundle that throws.

With `--recording`, the dev server serves a toolbar shell at `/__record` (what the browser is pointed at) that frames the app, which keeps the root URL its router and its own links expect. Framing is what makes this work: the recorder snapshots the app document on every mutation, so a toolbar living in that document would record itself.

Fixes WIN-2275

## Changes

- `cli/generate-dev-recorder.ts` (+ `bun run gen:dev-recorder`) bundles the frontend recorder into `cli/src/commands/app/devRecorderBundle.gen.ts`, recording the repo-relative sources esbuild pulled in and their hash.
- `cli/src/commands/app/devRecorder.ts`: the shell page (Record / Stop, live step count, Download JSON, Open in player) and the postMessage relay described below.
- `cli/src/commands/app/dev.ts`: the `--recording` flag and four routes, all gated behind it. `/__record` serves the shell (and is what `--recording` opens). `POST /__recordings` writes the recording to `<app folder>/recordings/recording-<ts>.json` (100 MB cap, matching what the player will load; cross-origin POSTs refused), `GET /__recordings/<file>` serves it back with `Access-Control-Allow-Origin` set to the instance origin so the player's `?src=` can fetch it, and `/__wm_recorder.js` serves the bundle.
- `cli/src/commands/app/wmillTsDev.ts`: the dev shim announces each runnable request and its answer to the shell, plus a ready message on every fresh module so the shell can settle calls a reload stranded (their WebSocket died with the old document; without this, Stop waits out the whole 60s job budget). A runnable call leaves the page over the dev server's WebSocket without touching the DOM, so without this the recorder settles on the spinner instead of the job's result. The shell relays the answer back into the frame, which is where `watchRunnableBridge` expects it (that is how the deployed runner delivers it). Inert outside an iframe, so plain `wmill app dev` is unaffected.
- `cli/src/commands/app/raw_apps.ts` and `cli/src/commands/sync/sync.ts`: `recordings/` is dropped from the app source push and from the sync diff. Both are needed: the push builds its own file set, while the differ walks the folder itself, so a recording left in the diff shows up as a change that pushing can never settle.
- Tests: `cli/test/dev_recorder_bundle_unit.test.ts` fails when the vendored bundle drifts from the frontend recorder, `cli/test/dev_recorder_routes_unit.test.ts` pins who may save a recording, what one may be named and that two saves never collide, and `cli/test/raw_app_recordings_skip_unit.test.ts` pins the root-only sync exclusion. `AGENTS.md` and the regenerated CLI docs point at the regeneration step.

## Screenshots

Recording in progress at `/__record`, after a backend runnable returned:

recording

Stopped and saved, with the player link:

saved

The same recording replayed on the instance's `/replay`, showing the step's outcome (the job's rendered result, not the spinner):

player

## Test plan

- [x] `wmill app dev --recording` in a raw app folder: Record, type into a field, toggle a checkbox, pick a select option, click a button calling `backend.greet()`, Stop. The click step's `after` frame holds `Hello, Ada!`, not `running…`, and its `before` frame holds `no result yet`.
- [x] Pressing Stop while a runnable is still in flight shows "Waiting for the last job to finish…" and the recording ends on the job's result.
- [x] A password field records as `••••••••` and its value appears in no frame.
- [x] Editing a source file mid-recording reloads the frame: a `navigate` step is recorded, the bridge rebinds, and subsequent interactions keep recording.
- [x] Reloading the frame while a 25s runnable is in flight: Stop completes in 0.1s (58s before the shell settled the stranded call).
- [x] An in-app link to `/` reloads the app inside the frame rather than nesting a second toolbar.
- [x] Record / Stop / Record again on the same page load: the second recording holds only the new steps.
- [x] The saved file replays on the instance's `/replay` via the toolbar's "Open in player" link (cross-origin fetch from the dev server).
- [x] Routes: a foreign origin sharing the dev server's port → 403, same-origin → 200, `GET /__recordings/../../../etc/passwd` (raw and percent-encoded) → 400, three concurrent saves → three distinct files, a 120 MB upload → a readable 413 with the dev server still serving.
- [x] A save that fails leaves Download JSON offered, so the recording is not lost.
- [x] Without `--recording`: `/` still serves the app unframed and `backend.greet()` still resolves.
- [x] `wmill sync push` on a folder holding a recording: the file is absent from the diff and from what a subsequent `wmill sync pull` writes back.
- [x] `bunx tsc --noEmit` clean on the touched files; `bun test test/*unit*.test.ts` 1047 pass.

---
## Summary by cubic
Add session recording to local app dev: `wmill app dev --recording` frames the app with a recorder toolbar and saves replayable sessions to `recordings/`. Recordings match the Windmill player format, bridge runnable responses so replays show job results, routes are hardened (same-origin only incl. preflight-free POSTs, validated names, 100 MB cap with graceful 413); the bundle hash is stable on CRLF checkouts; generated `.gen.ts` files are marked as generated; fixes WIN-2275.

- **New Features**
  - Shell at `/__record` frames the app at `/` with Record/Stop, live step count, Download JSON, and Open in player.
  - Routes: `/__wm_recorder.js`, `POST /__recordings` (same-origin only with a guard against preflight-free POSTs, unique names via `wx`, 100 MB cap), and `GET /__recordings/<file>` with CORS for the instance player `?src=`.
  - Dev shim relays runnable request/response via postMessage and announces reloads (`wmillDevReady`) so stranded jobs settle; replays show results, not spinners.
  - Privacy defaults: password inputs masked; mark sensitive elements with `data-wm-no-record`. `recordings/` is skipped by app source push and the sync diff (root-only, with Windows path normalization).

- **Migration**
  - After editing the recorder in `frontend/src/lib/components/recording/`, run `bun run gen:dev-recorder` from `cli/` to refresh the vendored bundle; the staleness check normalizes line endings for stable hashes on CRLF checkouts. Generated `.gen.ts` files are marked in `.gitattributes` to collapse diffs.
  - Replay saved files under `recordings/` at `<instance>/replay` via the toolbar link or with `?src=`.

<sup>Written for commit 57465a1bad39cf0a2558e1c3c000f6c1be75b603. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 1f7713466bd2cb570a12c1351070e159f9c71bd1

PR #10458, opened 2026-08-02
[ee] fix(groups): clean up workspace refs when instance groups are deleted

alpetric:
## Summary

Deleting an instance group left stale state behind in two places:

- the group name stayed in every workspace's `auto_invite` auto-assignment config (`instance_groups` array and `instance_groups_roles` object)
- users auto-added to workspaces via that group stayed in those workspaces

The two superadmin API paths that delete groups — `DELETE /api/groups/delete/{name}` and `POST /api/groups/overwrite` — did neither, so anything driving group lifecycle over the public API accumulated dangling references. Creating a group with the same name later silently re-acquires them.

Fixing that surfaced a **pre-existing data-loss bug in shipped code** affecting four call sites, not just the two this PR set out to change — see the comment below and the "Pre-existing bug" section.

Companion: windmill-labs/windmill-ee-private#708

## Changes

**Reference cleanup (the original fix)**

- Add `remove_instance_groups_from_workspace_settings`, stripping the given names from `auto_invite->instance_groups` and `->instance_groups_roles` across all workspaces. Shared by both call sites.
- `delete_igroup` and `overwrite_igroups` call it, plus `cleanup_removed_instance_groups` to drop workspace memberships whose `usr.added_via` names the deleted group.
- Both re-derive membership afterwards via `process_instance_group_auto_adds`, mirroring `edit_instance_groups`. `added_via.group` records only a user's *highest-precedence* group, so cleanup alone evicts members who still qualify through another configured group.
- `workspace_settings` is updated before the user cleanup at both sites, matching `edit_instance_groups`. The reverse order takes the same two locks the other way round and can deadlock against a concurrent workspace edit. Both still precede the `email_to_igroup` / `instance_group` deletes.
- Cleanup is gated on `private` alone, matching `edit_instance_groups`. CE ships `private` without `enterprise`, so gating on `enterprise` would scrub the workspace reference while stranding the orphaned users.
- The row filter uses `?|`: it returns false on a non-array value, where `jsonb_array_elements_text` raises `cannot extract elements from a scalar` and aborts the transaction. Not index-backed — the GIN index covers the `auto_invite` column, not this expression — which is acceptable at one row per workspace.

**Data-loss fix (EE companion)**

- `remove_users_from_instance_group_workspaces` now skips users who still belong to another instance group configured for that workspace, instead of deleting and re-adding them. It takes an explicit `removed_groups` argument so the "still qualifies" test cannot be satisfied by a group that is itself being removed, independent of whether the caller has already updated `workspace_settings` or `email_to_igroup`.
- Survivors keep their rows; callers' `process_instance_group_auto_adds` pass re-derives their role and `added_via`.
- `remove_user_igroup` gains the same reprocess pass, since it had no reconciliation of its own.

**Role re-derivation and serialization**

- All three SCIM member-removal paths in `scim_ee.rs` now re-derive workspace roles after a removal. Without this, preserving still-qualifying members left them holding the removed group's privileges — a member dropped from an admin group kept `is_admin`.
- `process_instance_group_auto_adds` takes `&impl AuditAuthorable` rather than `&ApiAuthed`, so SCIM (which has no `ApiAuthed`) can call it. It only ever used the value for audit attribution.
- `remove_users_from_instance_group_workspaces` locks the member's `usr` row before testing qualification. Two concurrent removals of different groups would otherwise each observe the other's still-present membership, both skip the deletion, and leave workspace access with no granting group.
- `delete_igroup`, `overwrite_igroups` and `remove_user_igroup` take a transaction-scoped advisory lock. Their internal orderings are dictated by correctness rather than one lock hierarchy, so serializing them removes deadlock between them instead.
- In `overwrite_igroups` the removal now runs after the imported memberships are installed. Judging qualification against pre-import membership deleted anyone who moved from a dropped group to a retained one, then re-added them stripped of their workspace data.

## Test plan

- [x] `test_remove_user_from_instance_group_rederives_role` — a member dropped from an admin group loses `is_admin` and is re-roled from the group they still belong to. Fails without the reprocess pass.
- [x] `test_overwrite_igroups_preserves_moved_member_data` — a member moved from a dropped group to a retained one in one import keeps their workspace data. Fails with the removal placed before the import.
- [x] `test_delete_instance_group_preserves_access_via_other_group` — a user in groups A (admin) and B (developer) keeps workspace access at B's role when A is deleted, `added_via` re-points at B, a user whose only path was A is removed, neither `auto_invite` field mentions A, **and their favorite and draft survive**. Verified to fail on each of those assertions with the corresponding fix disabled.
- [x] Existing `test_group_endpoints` still passes.
- [x] Manual verification against a running EE instance for `delete`, `overwrite`, and `edit_instance_groups`, including the data-survival case.
- [x] Deleting a group succeeds with a `workspace_settings` row whose `instance_groups` is JSON `null` (previously aborted the transaction).
- [x] Compiles and passes `SQLX_OFFLINE=true` with no features, `--features private`, and `--features enterprise,private`.

## Pre-existing bug this also fixes

`delete_workspace_user_internal` drops the member's drafts, favorites, inputs, captures, tokens, folder ownership, `usr_to_group` memberships and `extra_perms`. Every caller of `remove_users_from_instance_group_workspaces` deleted still-qualifying users and relied on a later pass to re-add them — restoring the `usr` row but nothing attached to it.

This was already reachable in shipped code, independently of this PR:

- `edit_instance_groups` — simply editing a workspace's instance-group config destroyed data for any member whose highest-precedence group was dropped
- `remove_user_igroup` — removing one user from one group did the same
- the three SCIM member-removal paths in `scim_ee.rs`

All are fixed by the shared change. Reproduced and confirmed fixed against a running instance for `edit_instance_groups`: before, a member's favorite and draft were silently destroyed; after, they survive and the role correctly downgrades.

## Known gaps (still pre-existing, not addressed here)

- `overwrite_igroups` does not remove workspace memberships for users dropped from a group that is *retained* in the payload — only fully-removed groups are reconciled.
- `scim_ee.rs::delete_group` keeps its own copy of the `auto_invite` cleanup SQL.


alpetric:
### Heads-up: this now fixes a pre-existing data-loss bug in shipped code

Addressing the P0 from the last Codex review turned out to require fixing a bug that exists **today, independently of this PR**, and that affects three call sites this PR does not otherwise touch.

`remove_users_from_instance_group_workspaces` deleted any workspace user whose `added_via.group` matched the group being removed, relying on a later pass to re-add those who still qualified. But `delete_workspace_user_internal` drops the member's **drafts, favorites, inputs, captures, tokens, folder ownership, `usr_to_group` memberships and `extra_perms`**. The `usr` row came back; everything attached to it did not.

Reproduced on a running instance against `edit_instance_groups` — untouched by this PR — with a user in groups A (admin) and B (developer), both configured for the workspace:

```
BEFORE: favorites=1 drafts=1 is_admin=t
  -> POST edit_instance_groups, dropping A and retaining B
AFTER:  usr=1  favorites=0  drafts=0        <-- silently destroyed
```

Nothing looks wrong afterwards: the member is still in the workspace, so the loss is invisible until someone goes looking for their drafts.

Affected paths, all pre-existing:

- `edit_instance_groups` — editing a workspace's instance-group config
- `remove_user_igroup` — removing a single user from a single group
- the three SCIM member-removal paths in `scim_ee.rs`

The fix is in the shared EE function (windmill-labs/windmill-ee-private#708): still-qualifying members are left in place rather than deleted and re-added, with an explicit `removed_groups` argument so a group on its way out cannot satisfy the "still qualifies" test. Same repro after the fix:

```
AFTER:  usr=1  favorites=1  drafts=1  is_admin=f  via=eig_b
```

Data intact, role correctly downgraded, `added_via` re-pointed at the surviving group.

Worth flagging separately because the blast radius is wider than this PR's title suggests, and because any instance that has edited instance-group configuration may already have lost member data — that is not recoverable by deploying this.

alpetric:
Superseded by #10686, which replaces the delta-patching approach here with a state-based reconciler. Kept unmerged for its review history — the branch (`change-57e45bf0`) stays as reference; its integration tests and the reference-stripping helper were carried over.

PR #10460, opened 2026-08-03
fix: return result.json and stdout results from sandboxed containers

rubenfiszel:
## Summary

Fixes GIT-948.

A `# sandbox <image>` (daemonless container / nsjail) bash script could not return a
structured result: `handle_docker_v2_job` returned the hardcoded string
`"sandboxed container (<image>) completed successfully"` no matter what the script
wrote, so neither documented bash result mechanism (`./result.json`, or the last line
of stdout) worked. That made image-sandbox jobs unusable as flow steps, app data
sources, or typed API responses.

### `./result.json` had nowhere to land

The container is chrooted into the extracted image rootfs. Writes to the image's
`WorkingDir` do flow back to `{job_dir}/rootfs/...`, but the most common case doesn't:
`WorkingDir: ""` → `/`, which is nsjail's own ephemeral tmpfs root. So the file the script
wrote was simply discarded. The fix bind-mounts a worker-created host file at
`{WorkingDir}/result.json` inside the jail.

The host side deliberately sits *outside* `{job_dir}/rootfs`, so the worker never reads
a path the image controls — a hostile image can't plant `result.json` as a symlink to a
host file and have the worker return its contents. The container can't swap it either:
unlinking a bind-mount point fails. Since the content is fully container-authored it is
parsed with `read_and_check_file`, not the `unsafe_raw`-based `read_file`.

### The mount destination is image-controlled, so it is contained

nsjail resolves a mount destination against its temporary root *before* `pivot_root`,
using ordinary path resolution, then creates it. I confirmed against raw nsjail that an
unverified destination makes it create the file on the **host**, as the worker user —
arbitrary file/directory creation anywhere the worker can write. Two properties contain
it:

- **The profile applies this mount directly after the rootfs binds, before every other
  mount.** Only the image rootfs can then influence the resolution — not the tmpfs `/tmp`,
  not `# volume` / same-worker shared mounts, which are applied later.
- **`result_mount_dst` walks the entire destination against the extracted rootfs on the
  host**, including the final `result.json` element: `..` is rejected, any symlink
  component is rejected, and any stat error other than `NotFound` is rejected (an
  unreadable mode-000 directory the jail's uid 0 could still traverse must not read as
  "fine"). The first absent component ends the walk — nsjail creates the rest as real
  directories under a parent already proven non-symlink. `/tmp`, `/proc`, `/dev` and
  `/sys` are rejected outright: mounted after this bind, a result there would be shadowed
  anyway, so the job says so rather than silently dropping it.

Relatedly, `proto_str` now escapes `{`/`}` so an image value can never survive as a
`{PLACEHOLDER}` token for a later substitution pass to expand — `{ENVARS}` expands to
*unquoted* directives, so that was a config-injection route (pre-existing via
`{WORKDIR}`, widened by the new mount).

### The last line of stdout wasn't captured

The bash executor gets this from its `wrapper.sh` (`tee` into `result2.out`), but
`docker_v2` runs the image's own command with no wrapper — and writing a wrapper script
into the image-controlled rootfs is exactly what this module avoids for sandbox-boundary
reasons. Instead, `handle_child` now records the last non-empty **stdout** line it already
streams to the logs and returns it in `HandleChildResult`. This requires no image tooling,
so it also works for shell-less entrypoint-only images.

Two properties that capture has to get right, and that plain bash gets for free from its
single `2>&1 | tee` pipe:

- **stdout only.** `child_joined_output_stream` merges the two pipes with a fair
  `stream::select`, so their relative order is arbitrary — and a piped stdout is
  block-buffered while stderr is not. Capturing the merged stream would let a stderr
  diagnostic beat the real result nondeterministically. `OutputLine` now carries which
  pipe a line came from, and only stdout feeds the capture.
- **unmasked.** `sensitive_log_masks` rewrites secrets *and appends a multi-line notice*.
  That belongs in logs, not in a result, so masking is now applied to the logged copy
  only and the captured line stays raw (with the same NUL scrub the log path applies,
  since Postgres rejects a NUL in `jsonb` too).

Note this is a visible behavior change for existing `# sandbox <image>` jobs that print
output: they now return their last stdout line instead of the canned completion message.
That is the point of the fix and matches plain bash — worth a release-note mention for
anyone branching on the old string.

## Changes

- `docker_v2.rs`: `result_mount_dst()` validates the image `WorkingDir` against the
  extracted rootfs and yields the in-jail destination (or `None` + a job-log warning);
  `render_result_mount()` renders the nsjail bind of `{job_dir}/result.json` → that
  destination. After `handle_child`, resolve the result as bash does: non-empty
  `result.json` → checked JSON result (malformed JSON fails the job with a clear message);
  else last stdout line → string result; else the completion message. `proto_str` escapes
  `{`/`}`.
- `run.docker.config.proto`: new `{RESULT_MOUNT}` placeholder, immediately after
  `{ROOTFS_MOUNTS}` and before every other mount.
- `handle_child.rs`: lines carry their pipe (`OutputLine`); `HandleChildResult` gains
  `last_line` (stdout-only, unmasked, NUL-scrubbed), captured next to the existing log
  append so it also works under `NO_LOGS_AT_ALL`. Masking now applies to the logged copy
  only. No other executor reads `last_line`, so nothing else changes behavior.
- `job_logger.rs`: the NUL scrub inside `append_with_limit` is extracted as `strip_nul` and
  reused by the capture.
- `docs/docker-v2-runtime.md`: document the result convention and when the bind is
  skipped; drop the now-stale "the script result is a completion message" limitation.

## Test plan

Exercised against a real worker (nsjail + crane, `cargo watch` backend), one job per case.
Hostile images are built with `crane mutate --workdir` / `crane append` and served from a
local `crane registry serve`.

Result resolution:
- [x] `printf '{"greeting":"hello"}' > ./result.json` on `alpine:latest` (`WorkingDir` `/`)
      → result `{"greeting":"hello"}`
- [x] `httpd:2.4-alpine` (`WorkingDir` `/usr/local/apache2`) → result `{"cwd_result":true}`
- [x] malformed `./result.json` → job fails with "the `./result.json` written by the
      sandboxed container is not a valid JSON result: …" (no `unsafe_raw` on bad input)
- [x] last-line fallback → result `"{\"greeting\":\"from stdout\"}"` (string, as in bash)
- [x] stdout result + stderr noise, repeated runs → stably returns the stdout line
- [x] secret-valued last line → masked (with notice) in the logs, returned **raw and
      uncorrupted** as the result
- [x] NUL byte in the last line → stripped, job completes (would otherwise be rejected as
      `jsonb`)
- [x] no output at all → unchanged `"sandboxed container (alpine:latest) completed successfully"`
- [x] shell-less entrypoint-only image (`hello-world`) → last line returned
- [x] failing container (`exit 3`) → still fails with the exit code and log tail
- [x] 20k output lines → correct last line, no worker panic/stack overflow

Escape containment — each first reproduced against raw nsjail to confirm it *does* create a
host path without the guard, then run as a real job with the guard. In every case the mount
is dropped, the warning appears in the job logs, and the target host directory is untouched:
- [x] `WorkingDir: /../../../../../..<host path>`
- [x] rootfs symlink `app -> <host path>` with `WorkingDir: /app`
- [x] rootfs symlink `result.json -> <host path>` with `WorkingDir: ""` (final component,
      the case a prefix-only walk missed)
- [x] `WorkingDir: /tmp/work` (the region same-worker shared mounts live in)

Regressions on the shared `handle_child` path:
- [x] plain bash, `# sandbox` bash (`result.json` + last line), a bun job, and `WM_STREAM: `
      result streaming (merged into the result, still excluded from the logs) all unchanged

- [x] `cargo test -p windmill-worker --lib docker_v2` — 8 passed, incl.
      `result_mount_dst_targets_working_dir`, `result_mount_dst_rejects_escapes` (`..`,
      relative paths, directory symlinks at two depths, final-component symlinks at two
      depths, and the `/tmp` `/proc` `/dev` `/sys` regions) and brace escaping in
      `proto_str_escapes_injection`

### DIFF
--- a/backend/windmill-worker/nsjail/run.docker.config.proto
+++ b/backend/windmill-worker/nsjail/run.docker.config.proto
@@ -46,6 +46,17 @@ gidmap {
 # and avoid it. Generated from the extracted rootfs.
 {ROOTFS_MOUNTS}
 
+# `./result.json` (relative to the image's WorkingDir) bound to a host file outside the
+# image rootfs, so the JSON result the container writes flows back to the job without the
+# worker ever reading a path the image controls.
+#
+# Placed immediately after the rootfs binds and BEFORE every other mount: nsjail resolves
+# a mount destination against its temporary root before pivot_root, so whatever is mounted
+# earlier can steer where this lands. Keeping only the rootfs ahead of it means the image
+# is the single influence on that resolution — and `result_mount_dst` verifies the whole
+# path against the extracted rootfs on the host. Empty when it can't be verified.
+{RESULT_MOUNT}
+
 # Pseudo-filesystems the image expects. /tmp honors the same instance settings as
 # every other nsjail job (nsjail_tmp_backing tmpfs/disk, nsjail_tmpfs_size_mb);
 # /dev gets the standard nodes; /proc comes from mount_proc (the jail's own pid ns).
@@ -104,12 +115,6 @@ mount {
 # rootfs binds and the tmpfs /tmp so a volume target overrides any colliding image
 # path and isn't shadowed by the tmpfs. Empty when there are no volumes.
 {SHARED_MOUNT}
-
-# `./result.json` (relative to the image's WorkingDir) bound to a host file outside
-# the image rootfs, so the JSON result the container writes flows back to the job
-# without the worker ever reading a path the image controls. Mounted last so it wins
-# over any colliding image path or `# volume`.
-{RESULT_MOUNT}
 
 iface_no_lo: true
 
--- a/backend/windmill-worker/nsjail/run.docker.config.proto
+++ b/backend/windmill-worker/nsjail/run.docker.config.proto
@@ -105,6 +105,12 @@ mount {
 # path and isn't shadowed by the tmpfs. Empty when there are no volumes.
 {SHARED_MOUNT}
 
+# `./result.json` (relative to the image's WorkingDir) bound to a host file outside
+# the image rootfs, so the JSON result the container writes flows back to the job
+# without the worker ever reading a path the image controls. Mounted last so it wins
+# over any colliding image path or `# volume`.
+{RESULT_MOUNT}
+
 iface_no_lo: true
 
 #{DEV}
--- a/backend/windmill-worker/src/docker_v2.rs
+++ b/backend/windmill-worker/src/docker_v2.rs
@@ -194,15 +194,27 @@ fn proto_str(s: &str) -> String {
 }
 
 /// In-jail destination for the `./result.json` bind — `{working_dir}/result.json` — or
-/// `None` when the image's `WorkingDir` cannot be proven to stay inside the jail.
+/// `None` when the image's `WorkingDir` cannot be proven to keep it inside the jail.
 ///
 /// nsjail resolves a mount destination against its temporary root *before* pivot_root,
-/// with ordinary path resolution: a `WorkingDir` that walks out of that root — via `..`,
-/// or via a symlink planted in the image rootfs — makes nsjail create, and mount over, a
-/// path on the *host*, as the worker user. So only a destination with no relative
-/// components and no symlink anywhere along its path inside the rootfs is accepted.
-/// `/tmp`, `/proc`, `/dev` and `/sys` come from the jail profile rather than the image,
-/// so there is nothing image-controlled to traverse under them.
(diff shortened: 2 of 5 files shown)


rubenfiszel:
✅ Review round clean @ d9609a9af20d049fec7e14d1e1124ed31c078133

PR #10463, opened 2026-08-03
fix: report why a native trigger service refused instead of a 500

### LINKED ISSUE #10461: bug:API error when attempting to edit flow after update
### Describe the bug

I wanted to confirm that a bug with the ui was fixed that I reported last week, so I updated the Windmill containers and then went to cause the issue again.
But when I went to edit a flow to see what happens now, I noticed this instead:
> Failed to load trigger configuration: ApiError: Internal Server Error
(see screenshots)

This causes existing flow's Nextcloud trigger to be read only for me. In the logs of the windmill server I found this:
(I've replaced the email for here, it showed my login email correctly in the logs)
> 2026-08-03T07:53:58.847845507Z [windmill-windmill_server-1] 2026-08-03T07:53:58.847642Z ERROR request: windmill-common/src/error.rs:306: Error: HTTP 403 error: {"ocs":{"meta":{"status":"failure","statuscode":403,"message":"Logged in account must be an admin, a sub admin or gotten special right to access this setting"},"data":[]}} @lib.rs:425:32 error=Anyhow { error: HTTP 403 error: {"ocs":{"meta":{"status":"failure","statuscode":403,"message":"Logged in account must be an admin, a sub admin or gotten special right to access this setting"},"data":[]}}, location: "lib.rs:425:32" } method=GET uri=/api/w/test/native_triggers/nextcloud/get/27 traceId="a74b590a-f553-46cb-8551-ed44f4764683" username="kroetsch" username="kroetsch" email="..." email="..." workspace_id="test" workspace_id="test" username="kroetsch" username="kroetsch" email="..." email="..." workspace_id="test" workspace_id="test" username="kroetsch" username="kroetsch" email="..." email="..." workspace_id="test" workspace_id="test"

The error doesn't really make sense to me because I am the one who installed the instance, therefore I am still the SuperAdmin for it (I checked the panel for that to confirm I still am) and I am also an Admin for the workspace.

I cannot create a new flow with a Nextcloud trigger either, as I get a similar error message upon deployment. But only the Nextcloud triggers seem affected, everything else deployed as configured in the editor.

I also tried creating a new SA user, to see if just my account is broken, but that one is affected too.

### To reproduce

1. be SuperAdmin and have flows with Nextcloud triggers from v1.775.1
2. update the containers
3. attempt to edit the existing flow's Nextcloud triggers

Sorry but I don't even know how to reproduce this. Chances are that if I were to nuke the containers and rebuild the instance, it may not even have this problem anymore.

### Expected behavior

I expected to still be able to edit the flows after the update.

### Screenshots

Image

### Browser information

Edge Version 150.0.4078.99 (Official build) (64-bit)

### Application version

Windmill CE v1.776.0-6-gbaefa1345b

### Additional Context

_No response_

rubenfiszel:
## Summary

Fixes the reporting side of #10461.

The reported failure is **not a Windmill bug**: every method of Nextcloud's `webhook_listeners`
OCS API carries `#[AuthorizedAdminSetting(settings: Admin::class)]`, so the *Nextcloud* account
behind the workspace integration must be a Nextcloud admin (or hold delegated admin rights for
the Webhooks setting). Being a Windmill superadmin or workspace admin is unrelated — which is
exactly where the reporter got stuck ("I am still the SuperAdmin for it"). An expired or revoked
token would have produced 401 (`NotLoggedInException`), so the 403 is a genuine authorization
failure on that instance.

What Windmill owns is everything about how that fact reached the user: nothing. The UI said
`ApiError: Internal Server Error`, the trigger became uneditable, and the only actionable text
lived in a server log. This PR makes provider rejections legible and non-fatal, and fixes three
defects found on the way.

## Changes

**Provider errors are typed and readable**
- New `ExternalApiError { service, status, detail, hint }` in `windmill-native-triggers`,
  produced by every `External::http_client_request` failure. It stays downcastable so internal
  callers can still branch on the status, and renders as one sentence for the user.
- Two defaulted `External` hooks: `describe_error_body` (Nextcloud unwraps the OCS envelope so
  the user reads `Logged in account must be an admin…` instead of the raw JSON) and `error_hint`
  (per service, per status — Nextcloud's 403 hint names the actual requirement and says in so
  many words that a Windmill admin is a different thing).
- `map_external_error` at every handler and route boundary: upstream 404 → `NotFound`,
  5xx/transport → `BadGateway`, other 4xx → `BadRequest`. The provider's status is deliberately
  **not** mirrored — a 401/403 answered by Windmill reads as a Windmill permission problem.
- Transport failures keep their source chain, so `Connection refused` survives instead of being
  swallowed by reqwest's `Display`.

**A rejection no longer takes the trigger down with it**
- `GET /native_triggers/{svc}/get/{id}` degrades: on any non-404 provider failure it returns
  200 with the stored `service_config` and a new `external_error` field, instead of failing the
  request outright. The editor opens on the last saved configuration with a warning banner.

**Two dead 404 checks, now live**
- `handler.rs` matched `Err(Error::NotFound(_))` and `nextcloud/external.rs` matched
  `Error::InternalErr(msg) if msg.contains("404")` — neither could ever match, because the error
  was an `Error::Anyhow`. So the "trigger no longer exists on external service" state was never
  recorded from a GET, and a trigger whose webhook had already been removed in Nextcloud could
  not be deleted from Windmill at all (500). Both now use `http_error_status`, matching how the
  GitHub service already did it.

**The OAuth refresh-and-retry is restricted to 401**
- 403 means the account is authenticated and still not allowed; minting a new token for the same
  account cannot fix it. Nextcloud rotates its refresh token on every use and the store-back runs
  in a detached task, so a 403 loop burned a rotation per request and buried the real reason. All
  three providers answer 401 for an expired or revoked token.

**Frontend**
- `NativeTriggerEditor` shows `external_error` as a warning above an editable form, and a hard
  load failure as an error alert with a Retry button *in place of* the form — a config that
  could not be loaded must not be savable. `openEdit` resets the form up front so a failure
  cannot leave the previously edited trigger's target and config behind.
- The Nextcloud event picker shows the actual API error instead of telling a user with a
  perfectly connected integration to go check their integration.

## Screenshots

Provider rejects the read (403) — editor opens on the saved config with the reason and the fix:

after-403-warning

Webhook gone on the service (404) — recorded as missing, with a working retry:

after-404-loaderror

Service unreachable — the transport cause survives, and the event picker says what really happened:

service-down

Healthy trigger, unchanged:

healthy

## Test plan

Exercised end to end against a mock Nextcloud (OAuth token endpoint, `cloud/user`,
`integration_windmill` events, and `webhook_listeners` CRUD with a switchable 200/403/404 mode),
with the integration connected through the real OAuth callback flow:

- [x] 403 on read → `200` with `external_error` carrying the OCS message + hint (was `500` with the raw envelope)
- [x] 403 on save → `400` with the same message (was `500`)
- [x] 403 on delete → `400`, and the message says the trigger was kept in Windmill
- [x] 404 on read → `404`, `native_trigger.error` set to "no longer exists on external service" (this path was unreachable before)
- [x] 404 on delete → trigger actually deleted (was `500`, leaving an undeletable row)
- [x] Service down → `502` on the events route, `Connection refused` preserved in the message
- [x] All four states verified in the browser via Playwright (screenshots above)
- [x] `cargo check --features native_trigger`, `cargo test -p windmill-api-integration-tests --test native_triggers` (21 passed), `npm run check` (0 errors)

Two regression tests added, pinning what was silently broken: a provider 403 reaches the user as
the service's own sentence plus the hint rather than a 500 with a JSON envelope, and a provider
404 is recognized by `http_error_status`/`map_external_error` — the contract both the
missing-trigger branch and the delete 404-swallow depend on.

**Note on a deliberate non-change:** any upstream 404 on the read path is treated as "the trigger
is gone", so a 404 from a wrong `base_url` or a disabled `webhook_listeners` app is reported as a
missing webhook. That matches how the GitHub service has always classified 404s, is self-healing
on the next successful read, and distinguishing the two would mean threading "did the body parse
as the service's own envelope" through the generic handler. Called out rather than silently left.
Per-service `error_hint`s are worded so they never assert *which* call was refused, since one is
appended to every failed call of that service.

Fixes #10461

---
## Summary by cubic
Show clear provider errors for native triggers instead of 500s, and keep triggers editable when providers reject or are unreachable. Treat timeouts, throttling, and refresh outages correctly; addresses #10461.

- **Bug Fixes**
  - Map provider failures to readable messages via `ExternalApiError` with per‑service hints; 5xx/408/429 and throttled 403s → BadGateway, other 4xx → BadRequest; preserve transport causes; unwrap Nextcloud OCS messages; detect throttled 403s via `is_transient_response` (GitHub: “rate limit” and “abuse detection mechanism”; Google: `usageLimits` domain).
  - Degrade reads only for provider errors: `GET /native_triggers/{svc}/get/{id}` returns 200 with stored config and `external_error`; upstream 404 marks the trigger missing; internal failures still error.
  - OAuth: refresh only on 401; classify a rejected refresh by OAuth error code (`invalid_grant`/`invalid_client`) or body (e.g., GitHub `bad_refresh_token` with 200); treat others as outages; do not mirror token endpoint status into trigger responses; show reconnect hints only on refused grants.
  - Routes/UI/OpenAPI/Tests: GitHub/Google/Nextcloud routes wrap errors with `map_external_error`; editor shows `external_error` above an editable form and hard load failures with a Retry button; Nextcloud event picker shows the real API error; OpenAPI adds nullable `external_data` and `external_error`; tests cover readable errors, 404 recognition, throttling vs. refusal (incl. GitHub “abuse detection”), provider‑only read degradation, and refresh refusal by code.

<sup>Written for commit fd1d12cdb3edbaf2d56f5f1813208df0ef953c84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-integration-tests/tests/native_triggers.rs
+++ b/backend/windmill-api-integration-tests/tests/native_triggers.rs
@@ -765,21 +765,29 @@ fn test_provider_404_is_recognized() {
 }
 
 /// Sending a user to reconnect their integration is only right when the token endpoint refused
-/// the grant. A busy or broken endpoint has them fix credentials that are fine.
+/// the grant; a busy or broken endpoint has them fix credentials that are fine. And a token
+/// endpoint's status must never reach the trigger's, where 404 means the webhook is gone and
+/// drops the row.
 #[test]
-fn test_only_a_refused_grant_blames_the_credentials() {
-    let rejected = |code: u16| {
-        refresh_failure_status(Some(StatusCode::from_u16(code).unwrap()))
-            == Some(StatusCode::UNAUTHORIZED)
-    };
-    assert!(rejected(400), "invalid_grant answers with 400");
-    assert!(rejected(401));
-    assert!(rejected(403));
-    assert!(!rejected(408), "a timeout is not a credential problem");
-    assert!(!rejected(429), "rate limiting is not a credential problem");
-    assert!(!rejected(503));
+fn test_refresh_failures_blame_only_the_grant_they_refuse() {
+    let status = |code: u16| refresh_failure_status(Some(StatusCode::from_u16(code).unwrap()));
+    for refused in [400, 401, 403] {
+        assert_eq!(status(refused), Some(StatusCode::UNAUTHORIZED), "{refused}");
+    }
+    for other in [404, 408, 429, 500, 503] {
+        assert_eq!(
+            status(other),
+            None,
+            "{other} from the token endpoint says nothing about the trigger"
+        );
+    }
     assert_eq!(refresh_failure_status(None), None);
+}
 
+/// A service that is busy or broken has not refused anything, and callers react differently to
+/// the two.
+#[test]
+fn test_transient_service_failures_are_not_refusals() {
     for transient in [408, 429, 503] {
         let err = nextcloud_error(StatusCode::from_u16(transient).unwrap(), "{}");
         assert!(
--- a/backend/windmill-api-integration-tests/tests/native_triggers.rs
+++ b/backend/windmill-api-integration-tests/tests/native_triggers.rs
@@ -23,8 +23,9 @@ use windmill_native_triggers::{
     google::{parse_stop_channel_params, should_renew_channel},
     http_error_status, list_native_triggers, map_external_error,
     nextcloud::NextCloud,
-    require_native_integration_use, store_native_trigger, store_workspace_integration, External,
-    ExternalReadFailure, HttpRequestError, NativeTriggerConfig, OAuthConfig, ServiceName,
+    refresh_failure_status, require_native_integration_use, store_native_trigger,
+    store_workspace_integration, External, ExternalReadFailure, HttpRequestError,
+    NativeTriggerConfig, OAuthConfig, ServiceName,
 };
 
 // ============================================================================
@@ -761,6 +762,31 @@ fn test_provider_404_is_recognized() {
         classify_read_failure(nextcloud_error(StatusCode::NOT_FOUND, "{}")),
         ExternalReadFailure::Missing
     ));
+}
+
+/// Sending a user to reconnect their integration is only right when the token endpoint refused
+/// the grant. A busy or broken endpoint has them fix credentials that are fine.
+#[test]
+fn test_only_a_refused_grant_blames_the_credentials() {
+    let rejected = |code: u16| {
+        refresh_failure_status(Some(StatusCode::from_u16(code).unwrap()))
+            == Some(StatusCode::UNAUTHORIZED)
+    };
+    assert!(rejected(400), "invalid_grant answers with 400");
+    assert!(rejected(401));
+    assert!(rejected(403));
+    assert!(!rejected(408), "a timeout is not a credential problem");
+    assert!(!rejected(429), "rate limiting is not a credential problem");
+    assert!(!rejected(503));
+    assert_eq!(refresh_failure_status(None), None);
+
+    for transient in [408, 429, 503] {
+        let err = nextcloud_error(StatusCode::from_u16(transient).unwrap(), "{}");
+        assert!(
+            matches!(map_external_error(err), Error::BadGateway(_)),
+            "{transient} should read as the service failing to serve, not refusing"
+        );
+    }
 }
 
 /// A service read degrades to the stored configuration, but only for the service's own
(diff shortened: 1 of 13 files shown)


rubenfiszel:
✅ Review round clean @ fd1d12cdb3edbaf2d56f5f1813208df0ef953c84

PR #10466, opened 2026-08-03
chore: auto-allow scratch file ops confined to /tmp

rubenfiszel:
## Summary

The `/tmp` permission entries in `.claude/settings.json` were written as `Bash(mkdir:/tmp/*)`, `Bash(cp:/tmp/*)`, `Bash(tar * /tmp/*)` and so on. In permission-rule syntax the colon is only meaningful as a trailing `:*` wildcard, so each was matched as a literal command string that no real invocation ever produces. All twelve rules were dead: `mkdir`, `cp`, `mv`, `touch`, `chmod`, `tar` and `unzip` prompted every time despite appearing to be allowed.

They are replaced by a single `PreToolUse` hook. A hook rather than working rules, because permission rules match a command *prefix* and so constrain only the first operand — `Bash(cp /tmp/:*)` would happily match `cp /tmp/x ~/.zshrc`. It is also the only mechanism that can cover `mv` and `chmod`, which sit in the `ask` list (`Bash(mv:*)`, `Bash(chmod:*)`) that outranks any `allow` rule.

The hook requires **every** path operand to resolve to an absolute path strictly under `/tmp`. Requiring the sources too, not just the destination, keeps it from becoming a way around the `Read(**/.env)` / `Read(**/secrets/**)` deny rules: copying a project file into `/tmp` would land the content somewhere `Read(/tmp/**)` permits.

It follows the shape of the existing `guard-rm-outside-tmp.sh` — single plain single-line command, deny-by-default token charset excluding every bash quoting/expansion/separator character, `realpath -m` to resolve `..` and existing symlinks — with three deliberate differences that review surfaced:

- **No globs.** Bash expands them only after the hook has decided, so `realpath` sees the unexpanded pattern: `chmod 600 /tmp/link*` canonicalizes to itself and passes, then expands onto a symlink pointing outside `/tmp`. `guard-rm-outside-tmp.sh` can allow globs because `rm` unlinks a symlink rather than following it; every command here follows one instead.
- **Absolute operands only.** Resolving a relative operand against the cwd makes any bare word look like a safe path whenever the cwd is under `/tmp`, while the tool reads it as an option — `tar P -xf …` is `--absolute-names`, not `./P`.
- **Options are a per-command allowlist, checked at every position.** GNU utils permute, so `cp /tmp/tree -RL /tmp/out` still enables dereferencing recursion, which copies the content of a symlink target outside `/tmp` into readable scratch. Plain `-r` and `-a` (which implies `-d`) recreate such a symlink as a symlink and stay allowed.

`tar`/`unzip` need their own parser, since their write destination arrives as a flag *value* (`-C`, `-d`) and a bundle like `-xzf` consumes the token after it. Flags are an allowlist, so `-P`/`--absolute-names` defer without needing enumeration. Extraction additionally requires a destination under `/tmp`, or a cwd already under `/tmp`, since members otherwise land in the checkout.

One residual risk is documented in the hook header rather than papered over: an archive containing a symlink pointing out of `/tmp`, followed by a write through it, can still escape, because tar applies member symlinks as it extracts. The archive must itself be under `/tmp` to reach that point, so this is a hazard only for archives fetched from an untrusted source into scratch.

## Changes

- Add `.claude/hooks/allow-fileops-in-tmp.sh`, a `PreToolUse` hook that auto-allows `mkdir`/`cp`/`mv`/`touch`/`chmod`/`tar`/`unzip` when every path operand is absolute and under `/tmp`, and makes no decision otherwise
- Register it on the `Bash` matcher alongside `guard-main-branch.sh` and `guard-rm-outside-tmp.sh`
- Delete the twelve dead `Bash(cmd:/tmp/*)` entries. No replacement allow rules: a prefix rule cannot express "every operand under `/tmp`", and any that tried would re-open the escape the hook exists to close
- `Read(/tmp/**)`, `Write(/tmp/**)` and `Edit(/tmp/**)` are untouched

Net effect is strictly more permissive than before for `/tmp` and unchanged everywhere else, since every rule removed already matched nothing.

## Test plan

- [ ] `bash -n .claude/hooks/allow-fileops-in-tmp.sh` parses
- [ ] Allowed, no prompt: `mkdir -p /tmp/x`, `touch /tmp/x/a`, `cp /tmp/x/a /tmp/x/b`, `cp -r /tmp/x/a /tmp/x/d`, `mv /tmp/x/b /tmp/x/c`, `chmod 700 /tmp/x/a`, `tar -czf /tmp/x/a.tgz /tmp/x/src`, `tar -xzf /tmp/x/a.tgz -C /tmp/x/out`, `unzip -o /tmp/x/a.zip -d /tmp/x/unz`, `unzip -l /tmp/x/a.zip`
- [ ] Escapes defer — **run these with cwd under `/tmp`**, which is what made them slip through: `tar P -xf /tmp/a.tar -C /tmp/out`, `tar xzf /tmp/a.tgz`, `cp /tmp/tree -RL /tmp/out`, `cp -r /tmp/a /tmp/b -L`
- [ ] Symlink and prefix escapes defer: `chmod 600 /tmp/link*`, `cp -RL /tmp/tree /tmp/out`, `mkdir /tmp/../etc/evil`, `touch /tmp/x ~/.bashrc`
- [ ] Outside `/tmp` still prompts: `cp /tmp/x ~/.zshrc`, `cp backend/.env /tmp/leak`, `chmod 777 /etc/passwd`, `mkdir /tmp`, `tar -xzf /tmp/a.tgz -C /home/rfiszel`, `tar -xzPf /tmp/a.tgz -C /tmp/o`, `unzip /tmp/a.zip -d /etc`
- [ ] Shell metacharacters defer: `mkdir /tmp/a; rm -rf /`, `mkdir $(echo /tmp/x)`, `mkdir "/tmp/a b"`
- [ ] Wrappers defer: `/bin/mkdir /tmp/x`, `timeout 5 mkdir /tmp/x`
- [ ] `rm` is unaffected and still handled by `guard-rm-outside-tmp.sh`

## Known deferrals

Absolute-operand-only means relative member names defer, including the idiomatic `tar -czf /tmp/a.tgz -C /tmp/src .` (on the `.`). That is the intended trade and it fails in the safe direction — use absolute operands, or accept one prompt.


rubenfiszel:
✅ Review round clean @ c5ef1d5c594fed7469fdbc262da28f68225322cb — nit-only verdicts at 9a544f911be8320bf0e9bccfcf1c65ffb208c56f (codex: mergeable/nits, claude: mergeable/nits, pi: good to merge); nits addressed in c5ef1d5c594fed7469fdbc262da28f68225322cb

PR #10470, opened 2026-08-03
feat: azure workload identity auth for mssql and postgres resources

rubenfiszel:
## Summary

Lets `ms_sql_server` and `postgresql` resources authenticate as the worker's **Azure
Workload Identity**, so a Windmill deployment on AKS can reach Azure SQL and Azure
Database for PostgreSQL Flexible Server with no password or hand-pasted AAD token stored
in the workspace.

**The mode is selected by a sentinel password, `ms_entraid`** — the same convention the
instance database already uses (`DATABASE_URL` with password `entraid` / `iamrds`,
`windmill-common/src/lib.rs`). That means **no resource type schema change on the hub**:
every database form already has a password field, so this reaches an existing deployment
the moment the worker is upgraded, rather than waiting on a hub schema round trip.

```jsonc
// ms_sql_server
{ "host": "myserver.database.windows.net", "dbname": "mydb",
  "password": "ms_entraid", "encrypt": true, "trust_cert": false }

// postgresql — user must be the Entra principal name the server knows
{ "host": "myserver.postgres.database.azure.com", "dbname": "mydb",
  "user": "windmill-worker-identity", "sslmode": "require", "password": "ms_entraid" }
```

The identity itself is the worker pod's: `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`,
`AZURE_FEDERATED_TOKEN_FILE` and `AZURE_AUTHORITY_HOST`, all injected by the Azure
workload identity webhook. The projected service-account token is exchanged with Entra ID
for an access token that replaces the password. Reaching two databases as two identities
means two worker groups, which is the same answer Kubernetes gives.

Fixes WIN-2286

## Changes

- New `windmill_common::azure_workload_identity`: reads the webhook-injected env, performs
  the client-credentials + client-assertion exchange, and caches the access token per
  identity and scope. Renewal is single-flighted per identity, expired entries are
  evicted, a failed renewal falls back to the still-valid token rather than failing the
  job (Entra ID throttles bursts), and a recent failure suppresses retries for 30s so a
  queue of jobs does not each pay the request latency.
- MSSQL: a resource whose password is `ms_entraid` authenticates with an Entra token
  (scope `https://database.windows.net/.default`) passed to tiberius as
  `AuthMethod::aad_token`, slotted into the existing
  `integrated_auth` → `aad_token` → user/password chain.
- Postgres: same sentinel, scope `https://ossrdbms-aad.database.windows.net/.default`,
  token used as the password. The native postgres resource previously only offered AWS
  RDS IAM auth.
- `pg_executor`: the ad-hoc `use_iam_auth` boolean becomes a `PgAuthMode` enum that also
  keys the connection cache, so a connection opened under one auth mode is never handed
  to a request asking for another. `use_iam_auth` combined with the sentinel is an
  explicit error rather than a silent precedence rule.
- `PgDatabase::connect_with_iam` and the new `connect_with_workload_identity` share one
  `connect_with_token` helper (TLS forced on, sslmode still selects verification depth).

### Judgment calls

- **Sentinel over new resource fields.** An explicit `workload_identity: true` boolean is
  more discoverable, but it is invisible until the hub schemas carry it, and `postgresql`
  lists `password` as required, so that route also needs the hub `required` relaxed. The
  sentinel needs neither and matches the existing instance-database convention. The cost:
  anyone whose password is literally `ms_entraid` authenticates as the worker identity
  instead of failing to log in. The job log prints `Using Azure Workload Identity
  (client id ...)` so the mode is never silent.
- **The postgres path is gated behind `enterprise`**, matching the existing IAM RDS auth;
  CE returns an explicit "requires Windmill Enterprise Edition" error. The MSSQL executor
  module is already `enterprise`-only, so it needed no extra gate.
- The Entra ID **scope is fixed per database kind** rather than configurable; sovereign
  clouds change the authority host (read from `AZURE_AUTHORITY_HOST`) but would also need
  a different scope, worth adding only if someone asks.

## Test plan

- [x] `cargo check` (CE) and `cargo check -p windmill-worker --features mssql,enterprise`
- [x] Unit tests: token endpoint built correctly from the scheme+trailing-slash form the
      webhook injects; cache key is per-scope; a token inside the refresh buffer is stale
      for serving but usable as the renewal-failure fallback; a recent failure is not
      retried by every queued caller; the sentinel password selects `PgAuthMode::WorkloadIdentity`
- [x] Real jobs on a worker with the webhook env simulated (`AZURE_TENANT_ID`,
      `AZURE_CLIENT_ID`, `AZURE_FEDERATED_TOKEN_FILE` pointing at a dummy projected token):
      MSSQL and Postgres resources with `password: "ms_entraid"` both reach Entra ID, and
      its rejection of the dummy assertion surfaces verbatim (`AADSTS5002723: Invalid JWT
      token...`), proving endpoint/form/parse/error paths end to end
- [x] Worker without the env vars → config error naming the missing var
- [x] MSSQL with no auth method at all → unchanged error, now mentioning the sentinel
- [x] Postgres with `use_iam_auth` plus the sentinel → explicit error
- [x] Postgres with a real password → still returns rows (regression on the `PgAuthMode`
      and cache-key refactor); on a CE build the sentinel returns the EE-gate error
- [ ] Not covered: a real AKS federated identity against a real Azure SQL / Azure
      PostgreSQL Flexible Server, for lack of Azure credentials here. Everything up to the
      token being handed to the driver is exercised above; the remaining hop can be
      checked independently by pasting an `az account get-access-token` token into the
      existing `aad_token` field.

### DIFF
--- a/backend/windmill-common/src/azure_workload_identity.rs
+++ b/backend/windmill-common/src/azure_workload_identity.rs
@@ -22,8 +22,9 @@ pub const AZURE_OSSRDBMS_SCOPE: &str = "https://ossrdbms-aad.database.windows.ne
 /// Renew an access token this long before it expires.
 const TOKEN_REFRESH_BUFFER: Duration = Duration::from_secs(5 * 60);
 
-/// A token with less life than this left is not worth handing to a connection.
-const TOKEN_MIN_LIFETIME: Duration = Duration::from_secs(30);
+/// A token with less life than this left is not worth handing to a connection: opening
+/// one takes up to 20 seconds, and it authenticates at the far end of that.
+const TOKEN_MIN_LIFETIME: Duration = Duration::from_secs(60);
 
 /// How long a failed renewal suppresses the next attempt, as long as the current token
 /// still works. Without it every queued job retries the exchange in turn, so a
@@ -172,7 +173,9 @@ impl WorkloadIdentityConfig {
             }
             Err(e) => {
                 *last_failure = Some(Instant::now());
-                match usable {
+                // Check again rather than trusting the pre-request value: a request that
+                // times out eats 20 seconds of whatever the token had left.
+                match slot.token_valid_for(TOKEN_MIN_LIFETIME) {
                     Some(token) => {
                         tracing::warn!("Keeping the current Entra ID token, renewal failed: {e:#}");
                         Ok(token)
--- a/backend/windmill-common/src/azure_workload_identity.rs
+++ b/backend/windmill-common/src/azure_workload_identity.rs
@@ -5,7 +5,7 @@
 //! Nothing long-lived is stored on the Windmill instance.
 
 use std::collections::HashMap;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant};
 
 use serde_json::Value;
@@ -22,9 +22,21 @@ pub const AZURE_OSSRDBMS_SCOPE: &str = "https://ossrdbms-aad.database.windows.ne
 /// Renew an access token this long before it expires.
 const TOKEN_REFRESH_BUFFER: Duration = Duration::from_secs(5 * 60);
 
+/// A token with less life than this left is not worth handing to a connection.
+const TOKEN_MIN_LIFETIME: Duration = Duration::from_secs(30);
+
 lazy_static::lazy_static! {
     /// Access tokens keyed by identity and scope, shared by every job on the worker.
-    static ref TOKEN_CACHE: Mutex<HashMap<String, CachedToken>> = Mutex::new(HashMap::new());
+    static ref TOKEN_CACHE: Mutex<HashMap<String, Arc<TokenSlot>>> = Mutex::new(HashMap::new());
+}
+
+#[derive(Default)]
+struct TokenSlot {
+    /// Read on the hit path, so it must never be held across the exchange.
+    token: RwLock<Option<CachedToken>>,
+    /// Held for the whole exchange, so a burst of jobs on a cold or expiring entry
+    /// makes one request to Entra ID instead of one per job.
+    refreshing: tokio::sync::Mutex<()>,
 }
 
 struct CachedToken {
@@ -32,9 +44,19 @@ struct CachedToken {
     expires_at: Instant,
 }
 
+impl TokenSlot {
+    fn token_valid_for(&self, remaining: Duration) -> Option<String> {
+        let cached = self.token.read().unwrap();
+        cached
+            .as_ref()
+            .filter(|cached| Instant::now() + remaining < cached.expires_at)
+            .map(|cached| cached.token.clone())
+    }
+}
+
 /// The federated credentials of the identity the worker authenticates as. The Azure
-/// workload identity webhook injects all of them as env vars; a resource may override
-/// them so one worker can reach databases behind distinct identities.
+/// workload identity webhook injects them as env vars; a resource may override the
+/// tenant and client so one worker can reach databases behind distinct identities.
 pub struct WorkloadIdentityConfig {
     tenant_id: String,
     client_id: String,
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Round 2 dispositions:

- **Codex P2 — failed refreshes retried by every queued caller**: fixed in 470c28e. The refresh mutex now guards the time of the last failed exchange; a caller that finds a recent failure and a still-usable token serves that token instead of re-hitting Entra ID (`REFRESH_RETRY_BACKOFF`, 30s). A cold slot still retries every time — there is nothing to fall back on. Unit test added for the decision itself (`should_attempt_refresh`), which is the part that can silently regress; the surrounding lock behavior isn't testable without injecting the exchange, which would cost more structure than it buys.
- **Pi P2 — blank `tenant_id`/`client_id` on `PgDatabase` key their own cache entry**: fixed in 470c28e with `empty_as_none` on both fields, matching the `MssqlDatabase` counterpart.
- **Claude P2 — duplicates `db_entra_ee::acquire_azure_token`**: acknowledged, not done here. That helper lives in `windmill-ee-private` behind `enterprise + private` and is hardcoded to the `ossrdbms` scope for the *instance* database URL; the new one is CE code reachable from a CE-compiled `windmill-common`, so consolidating means moving the EE helper into CE and re-pointing the instance-DB auth path at it. That is a separate refactor with its own EE companion PR, and doing it inside this PR would put the instance's own database connection at risk for no user-visible gain.

rubenfiszel:
✅ Review round clean @ 11eb77da113c3f78b0d41d0e10d5151df8dbd06e — nit-only verdicts at 470c28eeaf698e00a41f89510f540aae3897bd3d (Claude: good to merge, Pi: good to merge, Codex: mergeable with one nit); the remaining nit is addressed in 11eb77d — the fallback token's lifetime is now re-checked *after* the failed exchange instead of reusing the pre-request value, and `TOKEN_MIN_LIFETIME` is raised to 60s so a token handed to a connection outlives the 20s connect. The duplication-with-`db_entra_ee` nit is dismissed with reasoning in the comment above.

Not added: a regression test for refresh latency crossing the threshold — it needs an injectable clock/exchange, which is more structure than the one-line re-check warrants. `test_token_within_refresh_buffer_is_stale_but_usable` guards the threshold itself.

rubenfiszel:
Round 4 dispositions:

- **Claude P2 — the postgres path logs nothing when the sentinel switches the mode**: fixed in 6ece0c5. `do_postgresql` now appends `Using Azure Workload Identity` to the *job* logs, not just the worker logs, so the stated mitigation for an accidental sentinel holds on both databases.
- **Claude P2 — `ms_entraid` diverges from the `entraid` it cites**: deliberate, and the doc comment now says why. `DATABASE_URL`'s `entraid` is set by whoever runs the instance; this one sits in a field where users keep their own secrets, so a less password-shaped literal is worth the divergence. The analogy in the comment is to the convention, not to the literal.
- **Codex P2 — stale configuration contract in the PR description**: the description was rewritten for the sentinel contract before this round was triggered, and currently documents `password: "ms_entraid"` with the env-only identity and no resource fields (the removed booleans appear nowhere in it). Nothing to change; the reviewer appears to have read a pre-edit snapshot.
- **Codex P2 — MSSQL fallback lifetime not bounded by its connection path**: acknowledged, not changed. It is real in principle: unlike the postgres path, the MSSQL connect and Azure redirect are not wrapped in a timeout, so a stalled TCP/TDS handshake could outlive a fallback token that had just over `TOKEN_MIN_LIFETIME` left. In practice the window needs a *failed renewal* to coincide with a handshake stalling past 60s, the token goes out in the TDS login packet rather than at the end of the sequence, and the failure mode is one job erroring with an auth message. Adding a connect timeout to the MSSQL executor changes behavior for every existing MSSQL user, including ones on slow links with no Entra involvement, so it belongs in its own PR rather than riding along here.
- **Codex — no MSSQL sentinel unit test**: the MSSQL branch is a single `==` against the same shared constant the postgres test pins, and the auth chain is inline in `do_mssql` rather than in an extractable function; restructuring it for a test that asserts one comparison would cost more than it guards. The MSSQL sentinel path is exercised end to end in the manual run instead (real job, webhook env simulated, Entra ID reached).

rubenfiszel:
✅ Review round clean @ 6ece0c5bc084ea5ccf2a56f197de93340dde2199 — nit-only verdicts at a1724db32835b4701a09fa89c91fafef9e209bfb (Pi: good to merge; Claude and Codex: mergeable with nits); nits addressed in 6ece0c5 / dismissed with reasoning in the comment above.

rubenfiszel:
Round 5 on 6ece0c5bc084ea5ccf2a56f197de93340dde2199: Codex, Claude and Pi all **good to merge**, no findings.

PR #10474, opened 2026-08-03
feat(frontend): pick workspace members from a searchable instance user list

diegoimbert:
## Summary

Adding a user to a workspace required typing the full email by hand, with no way to discover
which instance users exist. The email field is now a `Select`: superadmins get the addable
instance users to pick from, and anyone can still enter an address that is not on the instance
yet through the "Add new" entry.

The list is never fetched in full — the picker asks for 10 users at a time and what is typed
narrows the query server-side, so a large instance costs the same as a small one.

## Changes

- New `GET /w/{workspace}/users/list_addable` (superadmin only) returning the instance accounts
  that can still be added to the workspace, `{email, username}` only:
  - every exclusion (existing members, disabled accounts, service accounts) is part of the query,
    so the `LIMIT` counts addable accounts rather than being eaten by rows the UI would drop;
  - optional `search` filters on email or username (`ILIKE`), with wildcards in the input escaped
    so `%` and `_` match literally;
  - service accounts live in `usr` only, so selecting from `password` leaves them out by
    construction.
- `AddUser.svelte` replaces the email `<input>` with `Select`, fed by a debounced (300ms), 10-row
  call issued only while the popover is open and only for superadmins — everyone else gets the
  free-text "Add new" path with no failing request. The list refetches after an add so the user
  just added disappears without closing the popover.

## Screenshots

Superadmin — addable instance users (members, disabled and service accounts already excluded by
the query):

v2-instance-user-list

Typing narrows the list server-side and offers "Add new" for anything unknown:

v2-filtered-with-add-new

Workspace admin who is not a superadmin — no listing, "Add new" only:

v2-non-superadmin

## Test plan

- [x] `test_list_addable_instance_users` (new): member/disabled/service-account exclusion,
      `per_page=1` still returns the addable account behind three members, username search,
      literal `_` and `%`, and 401 for a workspace admin who is not a superadmin
- [x] `cargo test -p windmill-api-integration-tests --test users` — 3 passed
- [x] `npm run check` — 0 errors; `cargo sqlx prepare` (offline cache updated)
- [x] UI as superadmin against a live backend: list opens, typing filters, picking `dave` and
      pressing Add adds the member and drops him from the still-open picker
- [x] Regression for the disabled-input finding: with an 800ms delay injected on every listing
      request, typing 8 characters at 120ms intervals keeps focus on the input, never disables it,
      and drops no keystrokes
- [x] UI as workspace admin (non-superadmin): zero `list_addable` requests, "Add new" still works

Local `codex` review was skipped — the CLI cannot authenticate `gpt-5.6-sol` on this box; the CI
Codex reviewer covers it.


diegoimbert:
✅ Review round clean @ 2063b0a80ae1161d9cf9971078e4557d719c6837

PR #10478, opened 2026-08-03
fix(frontend): show wall-clock execution time for workflow-as-code jobs

diegoimbert:
Fixes WIN-2288. Resurrects the approach from #9440 (closed without merging).

## Summary

A Workflow-as-Code (WAC) job suspends while the task jobs it dispatches run, and resumes via checkpoint replay. Its `duration_ms` therefore covers only the orchestration script's own compute, so a workflow that takes 15s end-to-end reports ~2s.

Measured on a WAC run that dispatches two 3s tasks:

```
                  id                  | duration_ms |          started_at           |         completed_at          | wallclock_ms
--------------------------------------+-------------+-------------------------------+-------------------------------+--------------
 019fc836-c662-b008-1c42-84ba45fa7e46 |        2058 | 2026-08-03 15:21:04.863893+00 | 2026-08-03 15:21:19.612341+00 |        14748   <- WAC
 019fc836-c659-7b0c-8499-9cea276e15a5 |        4046 | 2026-08-03 15:21:00.809029+00 | 2026-08-03 15:21:04.858230+00 |         4049   <- plain script
```

The fix is display-only. `duration_ms` deliberately stays as-is: cloud usage accounting (`apply_completed_job_cloud_usage`) and EE workspace fairness consume it as worker service time, so widening it to wall-clock would bill and throttle a suspended WAC root for idle time — and excluding WAC roots from those consumers instead would let arbitrary user code in the orchestration body go uncounted.

## Changes

- **`frontend/src/lib/utils.ts`**
  - `isWorkflowAsCodeJob(job)` — true iff `workflow_as_code_status` carries a `_checkpoint`. AI-agent jobs populate the same column but never write `_checkpoint`, so they keep their `duration_ms`.
  - `jobDisplayDurationMs(job)` — wall-clock (`completed_at - started_at`) for WAC jobs, `duration_ms` for everything else, and `duration_ms` as a fallback when the timestamps are missing or inverted.
- **`JobStatus.svelte`** — the "Successfully ran in / Failed after / Failed after …, resolved" badges use the helper.
- **`JobPreview.svelte`** — the completed-job duration badge uses the helper (via a `$derived`, which also drops the now-redundant `duration_ms` guard).
- **`JobDetailFieldConfig.ts`** — the `duration` field uses the helper.
- **`utils.test.ts`** — pins the WAC/AI-agent/plain-script branches and the fallbacks.

Note the helper matches WAC *task* jobs too, not only roots — that's why it isn't named `…Root`. For a leaf task job wall-clock and `duration_ms` differ by a few ms, and a task that itself dispatches tasks genuinely wants the wall-clock total.

Backend is unchanged. No EE companion change.

## Not covered

Two surfaces still show worker compute time for WAC roots, both because the fix can't reach them from the frontend alone:

- **The runs list (`RunRow.svelte`).** `list_jobs` hardcodes `workflow_as_code_status: None` (`backend/windmill-api-jobs/src/types.rs`), so the list payload carries no way to identify a WAC job. Calling the helper there would be dead code, so `RunRow` is left untouched; making it correct needs a discriminator added to the list response.
- **The schedule "Avg" figure** on `/schedules`, which is aggregated over `duration_ms` server-side.

Also worth knowing: `JobDetailFieldConfig`'s `duration` field is currently `false` in every entry of `categoryFieldPresence`, so that call site is correct but not rendered today.

## Screenshots

Same WAC run (`duration_ms = 2058`, real wall-clock 14.7s), before and after:

| Before | After |
|---|---|
| before-wac-detail | after-wac-detail |

`JobPreview` on `/schedules`, for a scheduled WAC run with `duration_ms = 20` and wall-clock 6257ms — reads "Ran in 6.258s" instead of "20ms":

after-jobpreview

## Test plan

- [x] `npm run check` — 0 errors (10040 files).
- [x] `npx vitest run src/lib/utils.test.ts` — 97 passed, including the new `jobDisplayDurationMs` cases.
- [x] Ran a WAC preview job dispatching two 3s tasks: run-detail badge went from **2.058s** (before) to **14.749s** (after), matching the DB wall-clock of 14748ms.
- [x] Ran a plain bun script alongside it: badge stayed **4.046s**, exactly its `duration_ms` — unchanged.
- [x] Ran the same workflow on a per-minute schedule and hovered its run marker on `/schedules`: `JobPreview` reads **"Ran in 6.258s"** where `duration_ms` is 20.
- [ ] `local-review-codex` not run — the devbox `codex` CLI is 0.132.0, below the 0.144.1 the skill pins. Claude-native `local-review` was run and its findings addressed.


diegoimbert:
Nits from the last round addressed in f7ed74699f: `isWorkflowAsCodeJob` is module-local now, and `utils.test.ts` is prettier-formatted.

**On the runs-list / detail-page disagreement — deferring deliberately.** Closing it means adding a WAC discriminator to the `list_jobs` payload, which is a backend + `openapi.yaml` change; that's a scope call for this frontend-only fix rather than a nit, so I'd rather it be its own PR than smuggled into a review round. The interim state is asymmetric but not a regression: the detail page is now right, the list under-reports exactly as it did before, and clicking through moves toward the correct number, never away from it. Filed as follow-up in the description's "Not covered" section.

**Exercised the failure path you flagged.** A WAC job that throws *after* its first checkpoint:

```
                  id                  | status  | duration_ms |          created_at           |          started_at           |         completed_at          | has_ckpt
--------------------------------------+---------+-------------+-------------------------------+-------------------------------+-------------------------------+----------
 019fc854-dbc4-55c5-a3f5-339720c122bd | failure |        5580 | 2026-08-03 15:53:52.323850+00 | 2026-08-03 15:53:52.342377+00 | 2026-08-03 15:53:57.922488+00 | t
```

`_checkpoint` is present and `started_at` is preserved, so the helper takes the wall-clock branch and the badge reads **"Failed after 5.58s"** — which here equals `duration_ms`, because a run that dies mid-orchestration never accumulates a compute-only total that diverges from wall-clock. No partial-status fallback weirdness, and no change versus `main` for this case.

**On the approval-suspension behavior note** — agreed, and intended: a WAC root that waits on a human now reports the wait in its badge, which is what a flow root already does. Making the two consistent is the point of the change.

diegoimbert:
Fixed in 5fec4ba36e — good catch, and you're right that my scope argument didn't cover it. The `LogViewer` chip had the full `getJob` payload in hand, so this was reachable where `RunRow` genuinely isn't.

Changed the two completed-job call sites:

- `routes/(root)/(logged)/run/[...run]/+page.svelte:1105`
- `runs/JobRunsPreview.svelte:243`

Left `JobRunsPreview.svelte:283` on the raw field: that's the still-running branch, where there's no `completed_at`, so the helper returns `duration_ms` unchanged — substituting there would be a no-op, same reasoning that keeps `RunRow` out of this PR.

Verified on the run detail page for the WAC job whose `duration_ms` is 2058:

```
HEADER BADGE : Successfully ran in 14.749s
LOGS CHIP    : 14749ms      (was 2058ms)
```

after-logs-chip

On your counterargument — "what the worker spent producing these logs" is a defensible reading, but the chip sits directly under the header badge on the same screen, and two different numbers for one job reads as a bug regardless of which is philosophically right. Matching them is the safer call.

`npm run check` is clean (0 errors) and the 97 `utils.test.ts` cases still pass.

PR #10481, opened 2026-08-03
perf: cut ai chat request size and cache the prefix on openrouter

rubenfiszel:
## Summary

Discussion #10471 reported ~45K input tokens on every AI chat request, including a bare "hello". Measuring the actual payload confirmed it and found two independent causes.

**Tool definitions dominated the payload.** The global chat sent 53 tool definitions totalling 97,968 characters, on top of an 10,951-character system prompt. A single tool, `write_trigger`, accounted for 43,571 of those characters: its `config` was a `z.union` over all eleven per-kind trigger request schemas, inlined into the definition and resent on every iteration whether or not triggers ever came up. The script and flow chat modes carried the same defect in `create_trigger` (38,690 characters).

**OpenRouter got no prompt caching.** `cache_control` breakpoints were injected only on the native Anthropic path (`usesAnthropicMessagesApi`). OpenRouter routes through the OpenAI-compatible path, so Anthropic-backed models there were billed the full prompt on every single iteration, where a direct Anthropic key pays a tenth after the first.

Measured effect on the global chat's static payload: ~108,900 characters down to ~66,800, a 39% cut on every request. Script and flow modes drop ~38K characters. On top of that, repeat turns through OpenRouter now read the tools plus system block from cache instead of paying full input price.

## Changes

- `write_trigger` keeps `config` open-ended in its tool definition; a new `get_trigger_schema` tool serves the one kind's schema on demand. The call is now validated against the schema for the declared `kind` instead of a union over all of them, so a config belonging to another kind is rejected rather than reaching the API as a corrupt draft, and the failure carries that schema so the model can correct without a second call.
- Same treatment for `create_trigger` in the script and flow modes, via `system_prompts/generate.py`. Its `fn` already re-validated against the per-kind schema, so the union was pure payload. The generator now also emits a `triggerConfigSchemas` map (the per-kind configs with `path` / `script_path` / `is_flow` omitted) that backs the workspace-mode `get_trigger_schema`.
- `usesOpenRouterPromptCaching()` gates on the routed model (`anthropic/*`), since OpenRouter documents `cache_control` passthrough only for Anthropic-backed upstreams. `prepareMessages` places breakpoints on the system message and the newest user turn; the chat loop opts in via a new `promptCaching` flag, one-shot calls (metadata, autocomplete) do not, because a cache write costs more than an uncached read when the prefix is never resent.
- Anthropic caches the full prefix in `tools -> system -> messages` order, so the system-message breakpoint also covers the tool definitions. That matters because OpenRouter does not document `cache_control` on tool definitions at all.

## Known limitation

Tool results accumulating within a single agent loop land after the last breakpoint and are still re-sent uncached each iteration. Covering them needs a breakpoint on a `tool`-role message, which OpenRouter documents only for text content blocks, so I did not risk 400-ing every OpenRouter chat on unverified behavior. Worth revisiting against a real key.

## Screenshots

No visible UI effect: this changes request construction and tool schemas only, with no component or styling changes.

## Test plan

- [x] `npm run check`: 0 errors
- [x] `frontend/src/lib/components/copilot/chat/global/core.test.ts`: 176 passed (2 new)
- [x] `frontend/src/lib/components/copilot/chat/shared.test.ts`: 51 passed (1 repointed)
- [x] `frontend/src/lib/components/copilot/lib.anthropicRouting.test.ts`: 10 passed (2 new)
- [x] `chat/anthropic.test.ts`, `chat/chatLoop.test.ts` unchanged and green
- [x] `python system_prompts/generate.py` re-run; the generated file is exactly what the generator emits
- [ ] With an OpenRouter resource on an `anthropic/*` model, run a multi-turn global chat and confirm from OpenRouter's usage view that later iterations report cached input tokens rather than full-price input
- [ ] In global mode, ask for a Kafka trigger and confirm the model calls `get_trigger_schema` before `write_trigger`; then send a deliberately wrong-kind config and confirm the error carries the schema and it self-corrects in the same turn
- [ ] Repeat the happy path from a script chat, where `create_trigger` requires confirmation, and check the confirmation card still shows the full config

Not verified locally: this devbox has no workspace provisioned and no AI provider key, so the live chat path went unexercised. The request shape is pinned at the unit level only.

Reported in #10471.

---
## Summary by cubic
Cuts AI chat request size by ~39% and enables cached prefixes on `OpenRouter` (Anthropic models) to lower input tokens and cost. Trigger configs are now fetched per kind on demand; later turns reuse the cached system+tools prefix.

- **New Features**
  - Added `get_trigger_schema` to global and workspace tools; `write_trigger`/`create_trigger` keep `config` open-ended and validate against the per-kind schema at call time.
  - Removed inlined per-kind unions from tool defs; the generator exposes `triggerConfigSchemas` and a workspace `get_trigger_schema` for on-demand serving.
  - Enabled prompt caching for `anthropic/*` via `OpenRouter`: add `cache_control` breakpoints on the system message and newest user turn, gated by `usesOpenRouterPromptCaching` and opted-in from the chat loop via `promptCaching`. One-shot calls stay uncached.

- **Bug Fixes**
  - `write_trigger` now leads error messages with a recovery hint to call `get_trigger_schema`, ensuring it stays within the 2k tool-error cap and is visible to the model.

<sup>Written for commit 06913d1d67ed3c18e41552786fd68c3dc5acdca2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10481?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
## `ai_evals` A/B (global mode, sonnet)

Ran per `.agents/skills/ai-chat/SKILL.md`. Baseline is the merge base (`0827fd285b`) with only the six changed copilot files reverted, so the sole difference is this PR. Same cases, same model, one attempt each.

| case | metric | base | head | delta |
|---|---|---|---|---|
| `global-test4-multi-artifact-notification-job` (creates an HTTP trigger) | `finalContextTokens` | 59,093 | **49,207** | **-16.7%** |
| | total tokens | 150,256 | **118,105** | -21.4% |
| | verdict | pass, judge 97 | pass, judge 97 | unchanged |
| `global-test1-script-create` + `global-test18-human-slack-resource-with-secret` (avg) | `finalContextTokens` | 38,097 | **26,520** | **-30.4%** |
| | total tokens | 192,126 | **94,025** | -51.1% |
| | verdict | 100% pass, judge 100 | 100% pass, judge 100 | unchanged |

`finalContextTokens` is the metric the skill says to optimize first, and it drops on every case, including the two that never touch a trigger, which is the point: the payload was fixed overhead paid by every request.

test4 is the case that matters for regression risk, since it is the one that has to reach `write_trigger`. It still passes with an identical judge score, so the `get_trigger_schema` round trip did not cost correctness.

Caveat: one attempt per case, so the cumulative token column is noisy (it scales with how many loop iterations the model happens to take). `finalContextTokens` is the stable one.

rubenfiszel:
✅ Review round clean @ 06913d1d67ed3c18e41552786fd68c3dc5acdca2 — nit-only verdicts at 06913d1d67ed3c18e41552786fd68c3dc5acdca2 (codex/claude/pi all non-blocking); the sole nit across all three was the missing `ai_evals` A/B, posted above after the round was triggered. Head unchanged since.

rubenfiszel:
## `ai_evals` A/B part 2: script and flow modes

The earlier A/B covered global only. `workspaceTools.ts:125` is `create_trigger`, which serves **script** and **flow**, so those were unmeasured. Ran all three existing `create_trigger` cases, same baseline method (merge base `0827fd285b` with only the six changed copilot files reverted), same model (sonnet).

| case | mode | `finalContextTokens` base → head | delta | pass base → head |
|---|---|---|---|---|
| `script-test3-create-current-script-http-trigger` | script | 26,632 → **17,622** | **-33.8%** | 1/1 → 1/1 |
| `flow-test16-create-current-flow-http-trigger` | flow | 45,718 → **36,795** (mean) | **-19.5%** | 1/4 → **4/4** |
| `flow-test18-implicit-http-trigger-intent` | flow | 36,227 → **27,270** | **-24.7%** | 1/1 → 1/1 |

`flow-test16` is averaged over 4 attempts per revision because it failed on the baseline; see below.

### The measured drop matches the static measurement

The per-case reduction is nearly identical across all three: **-9,010 / -8,923 / -8,957** tokens. `create_trigger`'s definition went from 38,690 to 711 characters, so 37,979 characters removed at ~4.2 chars/token is ~9,000 tokens. The eval independently reproduces the static number, which is the strongest evidence that the saving is exactly the fixed per-iteration overhead this PR targets.

Global does the same: its deltas (-9,886 and -11,577) line up with `write_trigger` dropping ~42,750 characters.

### On `flow-test16`'s pass rate

Baseline scored **1/4** (three attempts failed the `test_run_flow` notAllowed assertion), head scored **4/4**. I am not claiming this PR fixes that case. The plausible mechanism is that a much smaller context makes the model less likely to wander into a disallowed tool, but 4 attempts per side is too thin to establish it, and the case looks pre-existingly flaky. Reporting it because the raw pass counts differ and it would be misleading to average it away.

No case regressed on any metric in either mode.

PR #10483, opened 2026-08-03
fix: bind the fork banner's comparison to the workspace it describes

rubenfiszel:
## Summary

Follow-up to #10467. That PR gated the fork banner's CTA on `!loading && comparison !== undefined`, which closed the fork-switch window it was aimed at but left the underlying problem: **`comparison` is not bound to the workspace it describes**, so several paths can still answer for the wrong fork — or answer at all when there is no answer.

Found by the post-merge review of the last commit, which landed without a CI verdict (the merge cancelled the round).

## Changes

- **`comparison` is now scoped to a workspace.** `checkForChanges` captures the workspace it was issued for, clears the previous fork's rows before fetching a different one, and stamps `comparisonFor` on success.
- **Only the newest request writes.** A monotonic `requestSeq` drops superseded responses. Previously two overlapping fetches shared one `loading` flag and an unguarded assignment, so switching A → B could resolve out of order and leave fork A's counts painted over fork B — permanently, until the next switch.
- **A failed refresh no longer keeps the previous fork's rows.** The `catch` clears `comparison`; before, the banner body hid them behind the error branch while the CTA still routed on them.
- **A skipped tally is not "nothing to deploy".** `skipped_comparison` returns `diffs: []` with a zeroed summary, which is indistinguishable from a fork with nothing ahead; the CTA claimed pending updates next to text saying the changes can't be displayed.
- **`upToDate` / `showDraftsOnly` are gated too.** They read `comparison` directly and take precedence over the CTA, so a fork with drafts *and* real diffs showed the drafts CTA for the whole load window.
- **Leaving a fork invalidates what's in flight.** The non-fork branch cleared `comparison` but not the pending request, so a late response landed on a workspace with no fork banner — and dragged a cross-workspace `getCiTestResultsBatch` POST with it.
- **The CI summary is bound to its comparison.** Its counters are dropped with the comparison, and its own async write is guarded — the 3s poll outlives a fork switch, so a slow batch for the fork you left would otherwise repaint the current one's CI line.

All five collapse into one rule, which the helper now states once: the banner answers only from a comparison that is loaded, for this workspace, and not skipped.

## Test plan

- [x] Browser: fork with only a parent-only diff still reads "1 change behind" → "Review & Update fork" → opens `&dir=update`
- [x] Browser: switching from an update-only fork to an ahead-only fork shows that fork's own counts and lands on Deploy with its ahead row (the case the review predicted would still fail)
- [x] Browser: switching back and forth between the two forks shows each one's own counts and CTA with no carry-over
- [ ] Not exercised in-browser: the out-of-order response race and the error path — both need injected latency/failure; verified by inspection against the `requestSeq` guard

## Note

`frontend/src/lib/gen` is gitignored, so a stale local client shows up as type errors in `stores.ts` / `NativeTriggerEditor.svelte` (`Workspace.is_dev_workspace` and friends). `npm run generate-backend-client` clears them — not a repo breakage.

---
## Summary by cubic
Bind the fork banner’s comparison and CI summaries to the active workspace and invalidate in‑flight fetches during fork switches, so counts and CTAs always reflect the right fork.

- **Bug Fixes**
  - Scope `comparison` per workspace using `comparisonFor`; clear stale rows on switch.
  - Cancel in-flight on switch; sequence writes; reset CI counts; drop superseded CI summaries.
  - Treat in-flight, failed, or skipped comparisons as non‑answerable; gate CTA, `upToDate`, and drafts.
  - On fetch error, clear `comparison` to avoid showing the previous fork’s rows.
  - Distinguish skipped tallies from “nothing to deploy” to avoid misleading CTAs.

<sup>Written for commit 1d31fda9684a8acd222e37045b43177ff19b0bee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10483?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/ForkWorkspaceBanner.svelte
+++ b/frontend/src/lib/components/ForkWorkspaceBanner.svelte
@@ -162,11 +162,15 @@
 			.filter((d) => d.kind === 'script' || d.kind === 'flow' || d.kind === 'resource')
 			.map((d) => ({ path: d.path, kind: d.kind as 'script' | 'flow' | 'resource' }))
 		if (items.length === 0) return
+		// Counted for the comparison current at call time — the poll below outlives a
+		// fork switch, and a slow batch for the fork we left would repaint this one's.
+		const seq = requestSeq
 		try {
 			const batch = await ScriptService.getCiTestResultsBatch({
 				workspace: $workspaceStore,
 				requestBody: { items }
 			})
+			if (seq !== requestSeq) return
 			let passing = 0
 			let failing = 0
 			let running = 0
--- a/frontend/src/lib/components/ForkWorkspaceBanner.svelte
+++ b/frontend/src/lib/components/ForkWorkspaceBanner.svelte
@@ -34,25 +34,34 @@
 	const drafts = useWorkspaceDrafts(() => (isFork ? ($workspaceStore ?? undefined) : undefined))
 	const draftCount = $derived(drafts.count)
 
-	// What the banner is allowed to answer from. Anything else — in flight, failed, or a
-	// tally that was skipped outright — counts zero of everything, which is
-	// indistinguishable from "nothing to deploy" and would send the user to the wrong
-	// direction. Typed helpers avoid the `never`-inference quirk on `$state` in `$derived`.
+	// Every read of `comparison` that decides what the banner says or where its button
+	// goes must go through this: anything else — in flight, failed, or a tally skipped
+	// outright — counts zero of everything, which is indistinguishable from "nothing to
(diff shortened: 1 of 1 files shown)


rubenfiszel:
✅ Review round clean @ 1d31fda9684a8acd222e37045b43177ff19b0bee — codex, claude and pi all "Good to merge", no findings.

PR #10485, opened 2026-08-03
fix: confirm step delete consequences in a single dialog

Guilhem-lm:
## Summary

Deleting a step could raise two confirmation dialogs in sequence for one user action: when the step both emptied a group *and* had dependents, confirming "Remove group?" opened "Confirm deleting step with dependents", which had to be confirmed too. The second dialog appeared exactly where the first had been, so it read as the first click not registering.

The two consequences are now computed up front and confirmed together in one dialog that names all of them.

## Changes

- Collapsed `pendingDeleteConfirmation` and `pendingGroupAction` into a single `pendingModuleAction` (`label`, `stepCount`, `groups`, `dependents`, `confirm`, `cancel`), rendered by one `ConfirmationModal` instead of two.
- `requestDelete` no longer chains callbacks: it applies the plan immediately when there is neither an affected group nor a dependent, and otherwise raises one confirmation carrying both. `request.plan` already exposed `structureDelete.affectedGroups` and `dependents`, so nothing new is derived.
- The move entry point feeds the same state, so "move" and "delete" stay consistent — the move dialog is unchanged in content but now shares one code path.
- Title and button follow the action (`Delete step?` / `Move step?`) and pluralise on the number of steps; the body enumerates the groups being removed and the steps needing changes.
- Modal visibility is driven by a separate `moduleActionOpen` flag rather than by clearing the action. `ConfirmationModal` renders its children through a 100ms fade-out, so clearing the value blanked the dialog mid-transition (and, on the move path, flipped the title to the delete wording). Both handlers early-return when already closed, keeping confirm one-shot while the buttons remain clickable during the fade.

## Screenshots

Delete of a step that is both a group's only member and referenced downstream — previously two dialogs, now one:

delete-confirmation

Multi-select delete, pluralised:

multi-select

Move path, sharing the same dialog:

move-confirmation

## Test plan

Exercised in the flow editor against a local backend:

- [x] Step alone in a group + a downstream dependent → one dialog listing both; a single confirm deletes the step and removes the group
- [x] Dependents only, no group → dependents section only
- [x] Group only, no dependents → group section only
- [x] Multi-select two steps in a group with a downstream dependent → `Delete steps?` / `Delete steps` / "delete the steps?"
- [x] Move a step out of a group it empties → `Move step?`; confirming commits the move
- [x] Cancelling the move dialog clears the pending move (no stuck "Cancel move" state) and leaves the flow unchanged
- [x] Dialog text sampled every frame across the fade-out after confirming: stable, no blank or wrong-verb frame
- [x] Double-clicking confirm during the fade deletes exactly one step
- [x] `npm run check:fast` clean; `flowDeleteUtils.test.ts` passes

No test added: the delete-plan logic is untouched (and already covered by `flowDeleteUtils.test.ts`) — this is a presentation change in a Svelte component.

### DIFF
--- a/frontend/src/lib/components/flows/map/FlowModuleSchemaMap.svelte
+++ b/frontend/src/lib/components/flows/map/FlowModuleSchemaMap.svelte
@@ -307,19 +307,33 @@
 		}
 	}
 
-	type PendingDeleteConfirmation = {
-		plan: DeletePlan
-	}
-
-	type PendingGroupAction = {
-		groups: FlowGroup[]
+	// A single delete can have several consequences at once (emptied groups *and*
+	// dependent steps); they are confirmed together so one user action never raises
+	// more than one dialog.
+	type PendingModuleAction = {
 		label: 'delete' | 'move'
+		stepCount: number
+		groups: FlowGroup[]
+		dependents: Record<string, string[]>
 		confirm: () => void
 		cancel?: () => void
 	}
 
-	let pendingDeleteConfirmation: PendingDeleteConfirmation | undefined = $state(undefined)
-	let pendingGroupAction: PendingGroupAction | undefined = $state(undefined)
+	// The modal keeps rendering `pendingModuleAction` while it fades out, so visibility is
+	// driven by `moduleActionOpen` rather than by clearing the value — clearing it would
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
✅ Review round clean @ 91c6066b3dd492e51688237907df8b744824b1da

PR #10486, opened 2026-08-03
feat: open a session edit in the preview panel from the edits list

Guilhem-lm:
## Summary

The session chat bar's **Edits** popover lists every item the chat touched, but the only thing a row could do was open the review & deploy drawer. Previewing is the more common intent, so the row's primary action is now **open in the preview panel**, and reviewing becomes an explicit control: a `Diff` button per row, plus a pinned footer row for the whole set.

## Changes

**The popover**

- `SessionChangesBar.svelte`: a row click opens (or focuses) that item in the session's preview panel and dismisses the popover. Kinds the panel can't host fall back to the diff drawer focused on that item, so no row is inert.
- Each row carries a `Diff` button (plain label + the `Diff` +/− glyph) that opens the drawer focused on that item.
- `SessionStatusPopover.svelte`: new optional `footer` snippet, rendered below the scrolling list so it stays put as the list scrolls. The keydown host moved to a wrapper around both, so a footer carrying `data-status-row` joins the arrow-key order as the last stop. The Edits popover uses it for **Review & deploy all**, which opens the drawer unfocused. Artifacts and Jobs pass no footer and are unaffected.

**Routing**

- `sessionPreviewTabs.svelte.ts`: new `previewTargetForDeployKind(kind, path)` maps the review dock's layout `Kind` to a preview destination — script/flow/raw_app to their live editor, a legacy drag-and-drop app to an iframe over `/apps/edit`, a data pipeline to its folder's graph view. Everything else returns `undefined`, which doubles as the caller's "can this row be previewed?" test.
- Targets route on the item's **storage** `path`, not `displayPath`: a draft-only item parks at a synthetic `draft_<uuid>` path and the friendly name would 404.

**Two consolidations the review pass pulled out**

- `SessionPreviewTabs.openAndPulse()` now owns the "flash the tab only when focusing it was the sole visible outcome" rule. It was written out at three call sites, one of them with a looser condition — the `open_item_preview` tool handler used to flash even when un-collapsing the panel or switching tabs had already made the change visible.
- `$lib/pipelinePaths.ts` states the `f/<folder>/data_pipeline` bundle layout once. It was parsed independently in three places (two of them disagreeing on whether the trailing segment is validated) and hand-built in two more; the compare page, the home list, the pipeline editor, the pipeline route and the new mapping all go through it.

## Test plan

- [x] `vitest run sessionPreviewTabs.test.ts` — 64 passed, covering the pipeline-folder routing, which kinds get a preview, and all three arms of the pulse condition
- [x] Session with draft scripts: row click opens the item as a live editor tab and closes the popover; a second row opens a second tab rather than retargeting the first
- [ ] A schedule / trigger / variable row falls back to the diff drawer focused on that item
- [ ] A data-pipeline edit opens `/pipeline/<folder>`, not the bundle path
- [ ] Row `Diff` button opens the drawer on that item; footer opens it unfocused on the whole set
- [ ] Arrow-key from the last row lands on the footer and wraps back to the first


Guilhem-lm:
✅ Review round clean @ 32bf99b6835efcae161a5af2d2113775187ea842

Guilhem-lm:
Both P2s addressed in 5c45d7b31b, plus the non-finding.

**Pipeline row named by its bundle path** — the row (and its tooltip) now name `f/<folder>`, matching where the click actually lands and the same call `CompareDrafts` makes. Done as a display-only `rowLabel()` in `SessionChangesBar` rather than by changing `DeployItem.displayPath`: that field is also what `onItemDeployed` feeds to `renameModifiedItem` as the post-deploy mask key, so rewriting it would have changed a routing identity to fix a label.

**`data_pipeline` restated in `pipelinePaths.ts`** — the regex is now built from `PIPELINE_DRAFT_KIND`, so the parser and the builder can't drift.

**Non-finding (`openInPreview` closing the popover)** — you were right that it was redundant: its only caller arrives through `pick()`, which already closed. Removed, so `openDiff`'s comment about being the one path that *isn't* routed through `pick()` now says something true.

`check:fast` clean, 64/64 vitest.

On the manual verification you flagged as still open: the trigger-row fallback and the pulse-on-refocus were exercised in a real session; the pipeline row's untracked-chat path and a fork-session pass have not been, and are worth a look before merge.

PR #10487, opened 2026-08-03
perf: serve the rare schedule options on demand instead of inlining them

rubenfiszel:
## Summary

Follow-up to #10481, which removed the inlined per-kind trigger schemas. The same pathology remained in schedules: `write_schedule` (global mode) inlined the full 6,376-character schedule request schema and `create_schedule` (script and flow modes) inlined 5,883, on every request, whether or not a schedule ever came up. `retry` alone was 1,489 characters, 24% of the tool.

Common fields stay inline. The rest reach the same validation through an `advanced` object whose shape a new `get_schedule_schema` tool serves on demand.

`write_schedule`: **6,376 → 2,418 characters**. Global tool payload: **55,706 → 52,030**.

## Changes

- `write_schedule` and `create_schedule` carry only the fields a schedule request usually has (path, schedule, timezone, args, enabled, summary, description, on_failure/on_recovery/on_success, plus script_path/is_flow in global mode). Everything else goes through `advanced`.
- New `get_schedule_schema` in both tool sets, serving the non-common fields. The workspace-mode variant additionally filters `script_path` / `is_flow`, which come from the runnable rather than the model.
- `droppedOptionKeys` in `shared.ts` guards the merge. Every sub-field of `retry` is optional, so a guessed `{attempts: 2, seconds: 30}` validates clean and strips to `{}`. Without the guard that writes a schedule with no retry policy and reports success, and the shape is no longer in the tool definition for the model to check itself against. Now it fails with a message naming `get_schedule_schema`.
- New eval case `global-test29-schedule-with-retry-and-error-handler` pinning that the model performs the lookup and produces a real retry policy and worker tag.

## `open_page` deliberately untouched

It is the other large tool, but it is navigation rather than a domain: almost any conversation can end in "show me the failed runs". A lookup round trip would tax the most common action. It also already prunes its fields per workspace and permission through `OPEN_PAGE_FIELD_PAGES` / `allowedOpenPages`.

## `ai_evals` A/B (global, sonnet)

| case | `finalContextTokens` base → head | verdict |
|---|---|---|
| `global-test17` (plain schedule) | 21,418 → **20,258** (-5.4%) | 3/3 pass both sides |
| `global-test29` (advanced schedule, new) | 21,531 → 22,436 (+905) | 3/3 pass, judge 88-90, lookup 3/3 |

The honest trade: ~900 tokens saved on every iteration of every conversation, ~900 paid once in the rarer conversation that needs advanced schedule options. Break-even at one iteration, so positive, but a thinner win than the trigger change.

A judge dip on `global-test17` initially looked like a regression. It is not: every attempt on both sides satisfied the full checklist, and the numeric wobble tracks `u//` paths. **No case in `global.yaml` seeds a username**, so `prepareGlobalSystemMessage` falls back to an empty one and the path-convention block reads `u//<name>`. `globalEvalRunner.ts` has a `GlobalUserFixture` for exactly this and nothing uses it. That is pre-existing noise across all global evals and is not addressed here, since fixing it moves every case's baseline.

## Screenshots

No visible UI effect: tool schemas and request construction only.

## Test plan

- [x] `npm run check`: 0 errors
- [x] `chat/global/core.test.ts` 181 passed (2 new: `advanced` folds into the draft; a mis-shaped option is refused and no draft is written)
- [x] `chat/shared.test.ts` 51 passed (target-field assertion extended to the served schedule schema)
- [x] `global-test29` 3/3 on head, `get_schedule_schema` called every time
- [ ] In a script or flow chat, ask for a schedule with a retry policy on a deployed runnable and confirm the Retries tab shows the attempts requested; no automated test covers the `create_schedule` merge end to end
- [ ] Run `global-test29` against a second model to confirm the lookup is not Sonnet-specific

---
## Summary by cubic
Serve rare schedule options on demand via a new `get_schedule_schema` tool, trimming the inlined schedule schema while keeping common fields inline. This cuts context size each turn (`write_schedule` 6,376 → 2,418 chars; global payload 55,706 → 52,030) with a one-time ~+900 token lookup when advanced options are used.

- **New Features**
  - Added `get_schedule_schema` (global and workspace) to return the JSON schema for advanced options (e.g., `retry`, `paused_until`, `tag`); the workspace version omits `script_path`/`is_flow`.
  - `write_schedule` and `create_schedule` accept common fields directly plus an `advanced` object; the workspace version derives the runnable target.
  - Shared builder (`buildScheduleToolSchema` / `advancedScheduleShape`) generates the tool’s common-field schema and the on-demand advanced shape, keeping definitions in sync; new eval `global-test29-schedule-with-retry-and-error-handler`.

- **Bug Fixes**
  - Real schedule arguments now win over duplicates inside `advanced` for both `write_schedule` and `create_schedule`.
  - Dropped-option detection checks the entire request, rejecting mis-shaped or misplaced advanced options (nested typos, advanced fields passed at the top level, or a non-object `advanced`) with guidance to call `get_schedule_schema`.
  - Unknown keys are now called out as “not schedule fields” and no longer point to the schema lookup.

<sup>Written for commit 939c61cb047e2512bab880c122aa2abe016ac8b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ai_evals/cases/global.yaml
+++ b/ai_evals/cases/global.yaml
@@ -567,6 +567,40 @@
     - passes dry_run as true
     - leaves only the schedule draft for review
 
+- id: global-test29-schedule-with-retry-and-error-handler
+  prompt: |-
+    The workspace already has a report digest helper.
+    Schedule it every morning at 6am UTC with `dry_run` on, and have it retry twice
+    if it fails. Run it on our `nightly` worker group.
+    Draft only, I'll review before deploying.
+  initial: ai_evals/fixtures/frontend/global/initial/report_digest_script.json
+  runtime:
+    maxTurns: 8
+  validate:
+    draftCountExactly: 1
+    requiredDrafts:
+      - type: schedule
+        pathIncludes:
+          - digest
+        valueIncludes:
+          - f/evals/global/send_report_digest
+          - attempts
+          - nightly
+  toolExpect:
+    requiredToolsUsed:
+      - write_schedule
+      - get_schedule_schema
+    forbiddenToolsUsed:
+      - write_script
+      - write_flow
+      - deploy_workspace_item
+  judgeChecklist:
+    - finds the existing report digest helper rather than creating a new script or flow
+    - creates one schedule draft for that helper running daily around 06:00 UTC
+    - configures a retry policy with two attempts
+    - routes the schedule to the nightly worker tag
+    - leaves the schedule as a draft without deploying
+
 - id: global-test18-human-slack-resource-with-secret
   prompt: |-
     I'm preparing Slack notifications for eval failures.
--- a/frontend/src/lib/components/copilot/chat/global/core.test.ts
+++ b/frontend/src/lib/components/copilot/chat/global/core.test.ts
@@ -534,6 +534,56 @@ describe('global AI tools', () => {
 		).toBeUndefined()
 	})
 
+	// The same option is legal at the top level, where the bag-only check never saw it.
+	it('catches a stripped schedule option passed outside advanced', async () => {
+		await expect(
+			callGlobalTool('write_schedule', {
+				path: 'f/schedules/toplevel',
+				schedule: '0 0 6 * * *',
+				timezone: 'UTC',
+				script_path: 'f/scripts/run',
+				is_flow: false,
+				args: {},
+				retry: { constant: { attempts: 2, seconds_typo: 30 } }
+			})
+		).rejects.toThrow(/retry\.constant\.seconds_typo/)
+	})
+
+	// A non-object `advanced` spreads to index keys that zod strips, so the options the
+	// model meant to set vanish. Rejecting beats writing a schedule missing all of them.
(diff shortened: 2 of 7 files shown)


rubenfiszel:
✅ Review round clean @ 939c61cb047e2512bab880c122aa2abe016ac8b9 — nit-only verdicts at 940ab5db7359641b3b5db5e75b0ff5521704eb4b (codex: Good to merge, pi: Good to merge, claude: mergeable with one nit); that nit (the guard's error pointing non-schedule keys at get_schedule_schema) is addressed in 939c61cb04.

PR #10488, opened 2026-08-03
fix: flow step picker layout and single hover/keyboard highlight

Guilhem-lm:
## Summary

Two related problems in the flow step picker ("+" between steps):

**1. `Approval/Prompt` wrapped out of its row.** The left kind column had no width of its own — it was sized by its widest label. Firefox treats `/` as a soft break opportunity (Chromium does not), so the label split across two lines inside a rigidly `h-7` button. The underlying fragility came from two commits that each undid half of the other: `f33ae2a4b1` (#6955) removed `whitespace-nowrap` from `Button` *and* switched it to `min-h-*`, then `cfeb294308` (#6628) switched it back to a fixed `h-*` without restoring the nowrap. The column is now a fixed `w-40` with truncating labels, so it renders identically in every engine.

**2. Mouse and keyboard were two independent highlights.** Hovering and arrow-keying each drove their own selection *and* their own tooltip, so two rows could be highlighted with two tooltips open at once. They now share one index (`selectedByKeyboard`), and the highlight is a neutral `bg-surface-hover` rather than the accent blue, which reads better for a transient selection. The left kind column deliberately keeps its accent — it is a persistent category selector, not part of the index space.

Two non-obvious details fell out of unifying them:

- Rows report hover on **`mousemove`, not `mouseenter`**. Arrow keys smooth-scroll the list, which slides a different row under a stationary cursor and fires `mouseenter` — hijacking the selection mid-navigation (one ArrowDown jumped three rows). `mousemove` only fires on genuine pointer movement.
- `Popover` opens on its own hover independently of `forceOpen`, and with hover now driving `forceOpen` for every row the pointer crosses, its 300ms/100ms timers could not keep up with a fast sweep — **11 tooltips open at once**, 7 left stranded. `disablePopup={index !== selected}` makes it structural: a non-selected row unmounts its tooltip regardless of timer state.

Long labels elsewhere in the picker truncated too — `adobe_acrobat_sign` overflowed the fixed 144px integrations column, and the overflow crushed the app icon to zero width (Button merges `shrink-0` onto `startIcon.classes`, but hub app-icon components do not forward `class`).

## Changes

- `InsertModuleInner`: kind column gets a fixed `w-40`; popover default height 400px → 480px (the column had **zero** vertical slack at 10 items, and `stop` adds an 11th); `displayPath` threshold tracks the new default height, and its dead `width > 650` branch is dropped (the measured div is `w-[650px]`, so it could never fire)
- `TopLevelNode`: label truncates, chevron `shrink-0`; new `neutral` prop selects the hover-surface highlight for result rows while the kind column keeps the accent
- `FlowInputsQuick`: one shared index for all five row groups via `onHover`; offsets extracted to named deriveds instead of being re-spelled inline at each call site; `showAiRows` now gates both the AI rows and their two index slots (they previously disagreed, so on the Approval kind two keypresses landed on rows that render nothing)
- `PickHubScriptQuick` / `WorkspaceScriptPickerQuick`: neutral highlight, `mousemove` hover, `disablePopup` so only the selected row may open a tooltip; hub rows get the same `h-auto min-h-7 py-1` + `leading-tight` as workspace rows so two-line rows are not clipped
- `WorkspaceScriptPickerQuick`: two-line summary+path rows get `leading-tight` and auto height — they previously packed 32px of content into a 28px row
- `ListFiltersQuick` / `FlowInputsQuick`: integration and folder names truncate with a `title` tooltip
- `FlowScriptPickerQuick`: suppresses its own CSS hover only when it is wired into the shared index, so the standalone AI Sandbox row keeps its hover affordance
- `InsertModulePopover`: drops `!resize` from the popover content — the element computes `overflow: visible`, which CSS `resize` does not apply to, so the handle never did anything
- `GenAiQuick`: same hover guard as `FlowScriptPickerQuick`

## Screenshots

Default picker — `Approval/Prompt` on one line in the fixed-width column:

pr-1-picker

Hovering a result row — one neutral highlight, one tooltip, `adobe_acrob…` truncated with its icon intact:

pr-2-hover

## Test plan

- [ ] **Firefox**: open the step picker and confirm every kind-column label stays on one line and the active category is still accent-blue
- [ ] Park the cursor over the result list and arrow up/down: exactly one row highlighted and exactly one tooltip open at every step, and the selection moves by exactly one row per keypress
- [ ] Sweep the mouse quickly down the list: never more than one tooltip, none left stranded
- [ ] Select the Approval/Prompt kind, type a search term, arrow down repeatedly: no keypress lands on a blank row
- [ ] Hover the AI Sandbox "Claude Code" row: it still highlights on hover
- [ ] Open the **trigger** picker ("+" on a trigger node) in a workspace with enough scripts to grow past 480px — the only host where two-line summary+path rows render: both lines fit unclipped in the workspace list and in the Hub list below it

No automated tests: this is presentational Svelte with no extracted pure-logic utility worth pinning.


Guilhem-lm:
✅ Review round clean @ 11d95f56da9a15d60c91822bc21d6cfcd5d08133

PR #10497, opened 2026-08-04
fix: match openrouter model ids by parsed vendor, not raw prefix

rubenfiszel:
## Summary

Reported on #10471: OpenRouter publishes its own floating aliases with a `~` prefix
(`~anthropic/claude-sonnet-latest`) alongside the vendor-pinned ids
(`anthropic/claude-sonnet-5`). The model picker lists whatever `/models` returns, so
both forms are selectable — but `usesOpenRouterPromptCaching` gated on
`startsWith('anthropic/')`, so picking an alias silently dropped the `cache_control`
breakpoints added in #10481 and put the chat back on full input price every turn.

Verified against the live catalog: 11 of 338 OpenRouter models carry the prefix, 4 of
them Anthropic (`~anthropic/claude-{sonnet,opus,haiku,fable}-latest`).

The same class of mismatch reaches further than the alias: OpenRouter dot-versions its
ids (`anthropic/claude-sonnet-4.6`) while `MODEL_CONTEXT_WINDOWS` keys on dashes
(`claude-sonnet-4-6`), so every OpenRouter-routed Claude 4.6+ resolved to the 200K
`claude` fallback instead of 1M and the chat trimmed context five times earlier than
needed.

## Changes

- Add `parseModelId` to `modelConfig.ts`: one normalizer returning `{vendor, base, variant}`
  (strips a leading `~`, splits on the first `/`, drops a `:free`/`:thinking` suffix).
  The predicates now match on the parsed parts instead of raw substrings.
- `usesOpenRouterPromptCaching` gates on `vendor === 'anthropic'`, so aliases get
  cache breakpoints.
- `getKnownModelContextWindow` collapses `.` to `-` on both the query and the table
  keys, so one entry covers both spellings of a version.
- Add the missing `claude-opus-5` / `claude-sonnet-5` 1M entries (they were falling
  through to the 200K `claude` family entry).
- `requiresMaxCompletionTokens` and `reasoningRegistry`'s `baseModelId` use the shared
  parser instead of their own `split('/').pop()`.

Not fixed here: the `-latest` aliases carry no version, so version-keyed checks
(`/claude-(opus|sonnet)-4/` in the reasoning registry) can't match them and the effort
control stays hidden for those ids. That needs either an explicit alias-family map or
reading `supported_parameters` off the `/models` response we already fetch.

## Test plan
- [x] `npx vitest run src/lib/components/copilot/` — the two new guards
      (`~anthropic/claude-sonnet-latest` gets breakpoints; `anthropic/claude-sonnet-4.6`
      resolves to 1M) fail on the pre-fix code and pass after
- [ ] With an OpenRouter key: select `~anthropic/claude-sonnet-latest`, send two turns,
      confirm the second reports cached input tokens

## Screenshots

None. The only user-visible effect is the context-usage indicator's denominator for
OpenRouter-routed Claude models, which needs a configured OpenRouter provider + key to
render; this dev instance has no AI provider configured, so the change is verified at
the unit level only. No markup or styling changed.

---
## Summary by cubic
Switch model matching to a parsed vendor approach and normalize version separators for OpenRouter IDs. Restores Anthropic prompt caching for `~` aliases and ensures 1M context windows for dot-versioned Claude 4.6+ and Claude 5, while anchoring version matches so shorter versions don’t swallow longer ones.

- **Bug Fixes**
  - Parse model IDs into `{ vendor, base }` and gate caching on `vendor === 'anthropic'` so `~anthropic/claude-sonnet-latest` gets breakpoints.
  - Normalize `.` to `-` for context-window lookups; add 1M entries for `claude-opus-5` and `claude-sonnet-5`.
  - Anchor context-window matches only for version-suffixed entries with `(?!\d)` so `gpt-4.1` doesn’t claim `gpt-4-1106-preview`; family fallbacks still catch welded versions like `llama3.1`.
  - Use the shared parser in `requiresMaxCompletionTokens` and the reasoning registry to correctly strip prefixes/suffixes.
  - Tests cover alias caching, dot-version context windows, and the scoped digit guard.

- **Refactors**
  - Document the legacy `/thinking` suffix invariant in `parseModelId` comments.

<sup>Written for commit 0ee63524485c2e8d804eb0a49750e57f8688f282. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/copilot/lib.anthropicRouting.test.ts
+++ b/frontend/src/lib/components/copilot/lib.anthropicRouting.test.ts
@@ -259,25 +259,31 @@ describe('OpenRouter prompt caching', () => {
 		{ role: 'tool', tool_call_id: 't1', content: 'tool output' }
 	]
 
-	it('breaks on the system prompt and the newest user turn for Anthropic models', async () => {
-		const { getCompletion } = await import('./lib')
-		h.currentModel = { provider: 'openrouter', model: 'anthropic/claude-sonnet-5' }
+	// The `~` form is OpenRouter's own floating alias for the same vendor, so it has
+	// to reach the same gate — a prefix match on the raw id silently misses it and
+	// puts the chat back on full price every turn.
+	it.each(['anthropic/claude-sonnet-5', '~anthropic/claude-sonnet-latest'])(
+		'breaks on the system prompt and the newest user turn for %s',
+		async (model) => {
+			const { getCompletion } = await import('./lib')
+			h.currentModel = { provider: 'openrouter', model }
 
-		await getCompletion([...conversation], new AbortController(), undefined, {
-			promptCaching: true
-		})
+			await getCompletion([...conversation], new AbortController(), undefined, {
+				promptCaching: true
+			})
 
-		const sent = openaiCreate.mock.calls[0][0].messages
-		const ephemeral = { type: 'ephemeral' }
-		expect(sent[0].content).toEqual([
-			{ type: 'text', text: 'system prompt', cache_control: ephemeral }
-		])
-		expect(sent[1].content).toEqual([{ type: 'text', text: 'first', cache_control: ephemeral }])
-		expect(sent[3].content).toBe('tool output')
-		// The chat replays this same history through the Anthropic path, which rejects
-		// unknown fields on its own blocks, so the originals must come back untouched.
-		expect(conversation[0].content).toBe('system prompt')
-	})
+			const sent = openaiCreate.mock.calls[0][0].messages
+			const ephemeral = { type: 'ephemeral' }
+			expect(sent[0].content).toEqual([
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ 0ee63524485c2e8d804eb0a49750e57f8688f282 — nit-only verdicts at 6ce0b70ce5614d4c7665d40e4465a847143b923e (claude/pi: good to merge; codex: nit only); nit addressed in 0ee6352448

PR #10498, opened 2026-08-04
fix: bundle a vector emoji font so emoji scale with flow graph zoom

Guilhem-lm:
## Summary

An emoji in a flow step summary kept its unzoomed pixel size when the flow graph canvas was
zoomed, so at high zoom it looked far too small next to the words beside it.

The cause is in Gecko, not in our code. On macOS, Firefox resolves emoji to Apple Color Emoji,
which is a bitmap (`sbix`) font, and it rasterizes those glyphs at a size that ignores an
ancestor's `transform: scale()` — which is exactly how `.svelte-flow__viewport` implements canvas
zoom. Text outlines re-rasterize at the scaled size; the bitmap emoji glyph does not.

Confirmed Firefox-only: Chrome scales the emoji correctly at every zoom level (measured ink
height 22px → 36px going 1.0× → 1.6×, in step with the zoom). Narrowed on an isolated test page
to the transform itself — scaling the same label by `font-size` instead of by `transform` renders
correctly, and *every* variant that used `transform: scale()` reproduced the bug, including with
`will-change: transform`, `translateZ(0)`, and `text-rendering: geometricPrecision`. So no CSS
property on the existing text fixes it; the emoji has to stop being a bitmap glyph.

The fix bundles Noto Color Emoji, whose COLR glyphs are vectors and therefore scale like any
other text, and appends it to the app-wide font stack after Inter.

## Changes

- Bundle Noto Color Emoji in `frontend/static/fonts/`, split into 10 files by `unicode-range`
  exactly as Google Fonts serves it. Client-side, a chunk is only fetched when a page renders an
  emoji in that range — verified in the browser that loading the flow editor with one 🚀 fetches
  **1 of the 10 chunks** (118 KB), and a page with no emoji fetches none. Note the 2.0 MB total is
  nonetheless embedded in every binary and image: `static_assets.rs` serves `frontend/build/` via
  `rust_embed`, which bakes in all ten chunks regardless of what clients request.
- Add the ten matching `@font-face` blocks to `frontend/src/lib/assets/app.css`, next to the
  existing self-hosted Inter face, with the Firefox constraint recorded.
- `frontend/tailwind.config.cjs`: `main: ['Inter', '"Noto Color Emoji"', 'sans-serif']`. This is
  the single app-wide source of truth — it reaches `html` through the `addBase` call in the same
  file. Noto sits *after* Inter, so it only ever picks up codepoints Inter lacks.
- `frontend/src/routes/+layout.svelte`: same stack on the toast wrapper, which has its own scoped
  rule rather than inheriting.
- Include the SIL OFL 1.1 license text alongside the font files.

Beyond the reported bug, this also gives emoji a consistent appearance across operating systems
and fixes Linux clients, many of which ship no color emoji font at all and render tofu boxes today.

## Screenshots

Flow graph at default zoom and zoomed in, both after the change:

pr-default

pr-zoomed

These are Chrome, which never exhibited the bug — they are evidence of no regression, not of the
fix. The bug is a compositor-level artifact: Playwright's Firefox screenshots re-render from the
display list rather than reading back the compositor, so they showed the emoji scaling correctly
even before this change. The fix itself was verified by hand in a real Firefox.

## Known limitation (pre-existing, not introduced here)

`assets/app.css` is exported by `@windmill-labs/components`, but `files` is `["dist", "package"]`
and `svelte-package -o package` never copies `static/`. So the published stylesheet references
font files the tarball does not ship — that is already true on `main` for
`url('/Inter-Variable.woff2')`, and these `@font-face` blocks follow the same pattern.

For embedders this means the emoji chunks 404 and the browser falls back to the system emoji font,
i.e. exactly the pre-PR behaviour; `font-display: swap` means nothing blocks on it, and
`unicode-range` means only the ranges actually used are requested. Making the URLs
package-resolvable means moving both Inter and Noto under `src/lib/assets/` and changing how the
SvelteKit app resolves them — worth doing, but as its own change rather than inside an
emoji-scaling bug fix.

## Test plan

- [x] In **Firefox**, open a flow with an emoji in a step summary and zoom the canvas in and out —
      the emoji scales with the text at every zoom level. Confirmed by hand; this is the one check
      that cannot be automated, since Playwright's Firefox screenshots re-render from the display
      list and were green even before the fix.
- [x] On a page with one emoji, exactly one chunk is fetched, not all ten (measured via
      `document.fonts`); a page with no emoji fetches none.
- [x] Chrome renders the flow correctly with the new stack (no regression).
- [ ] Safari smoke test.
- [ ] Eyeball a couple of emoji-heavy non-flow surfaces (AI chat, item summaries, markdown) — the
      app-wide stack changes emoji appearance there too, which is intended but worth a look.
- [ ] Render a keycap sequence (`1️⃣`, `#️⃣`): Inter covers the digit and `#` while chunk 2 carries
      U+20E3 and U+FE0F, so the cluster spans two fonts. Expected to be fine via cluster-aware
      matching, but it is the one case where the Inter-first ordering could bite.


Guilhem-lm:
Round triage — thanks, several of these caught things I had asserted but not proven.

**Fixed** (`1ef4f2aa27`): the `unicode-range` comment said editing the ranges "breaks" the gating; editing them mis-gates it. Reworded.

**Verified rather than taken on faith:**
- `rust_embed` at `backend/windmill-api/src/static_assets.rs:34` is `#[folder = "frontend/build/"]`, so yes — all 2.0 MB is compiled into every binary and image regardless of client-side `unicode-range` gating. Good catch; the PR description had only made the client-side claim.
- The fonts are COLR, not a bitmap build, confirmed directly rather than inferred from size: `fontTools` reports tables `COLR`/`CPAL`/`glyf` with no `sbix` and no `CBDT`, version `2.051;GOOG;noto-emoji:20250818`. That is the premise of the whole fix, so it mattered to check.
- The license copyright string: name ID 0 in the shipped binaries is `Copyright 2022 Google Inc.`, which is what the bundled OFL file now carries. Note this differs from the `Copyright 2013 Google LLC` in the noto-emoji repo's root `LICENSE` — the binaries' own notice is the one OFL §2 asks to be redistributed.

**Not acting on, deliberately:**
- **App-wide scope.** This was an explicit decision, not a side effect of a Firefox fix. Scoping to `.svelte-flow__viewport` would fix the reported bug only; app-wide additionally gives emoji a consistent appearance across operating systems and fixes Linux clients, many of which ship no color emoji font and render tofu today. Signed off with that tradeoff understood.
- **Flag chunk (chunk 0, 709 KB / 35%).** Considered dropping it for binary size; keeping it, for the same consistency reason.
- **Exported stylesheet / `VITE_BASE_URL`.** Both are pre-existing properties of `assets/app.css`, not regressions: `main` already ships `url('/Inter-Variable.woff2')`, `files` is `["dist","package"]`, and `svelte-package -o package` never copies `static/`. Fixing it properly means relocating both Inter and Noto under `src/lib/assets/` with package-relative URLs and changing how the SvelteKit app resolves them — worth doing as its own change, not inside an emoji-scaling bug fix. Recorded in the PR description under "Known limitation".

Guilhem-lm:
✅ Review round clean @ 1ef4f2aa270256c60f6fbb658188aae04d271dc2 — nit-only verdicts at 1f0ca81ffe72349f7306630cea8f16ecb1c6868c; nits addressed in 1ef4f2aa270256c60f6fbb658188aae04d271dc2 / dismissed in the triage comment above

Guilhem-lm:
On the packaging P2 — sticking with leaving it, but the earlier dismissal was imprecise and the distinction Codex drew deserves acknowledging:

> Inter has the same pre-existing packaging gap, but these ten additional failed requests are introduced here.

That's right. The *pattern* is pre-existing (`main` already exports `app.css` with `url('/Inter-Variable.woff2')` under `files: ["dist","package"]`, and `svelte-package -o package` never copies `static/`), but the ten new 404s for npm consumers are genuinely added by this PR. They are non-blocking and async — `font-display: swap` renders the system fallback immediately, `unicode-range` means only the ranges actually used are requested, and the resulting appearance for embedders is exactly the pre-PR behaviour — but "pre-existing pattern" isn't the same as "no new cost", and my earlier triage comment elided that.

Still deferring the fix rather than folding it in: doing it properly means moving both Inter and Noto under `src/lib/assets/fonts/` with stylesheet-relative URLs so Vite fingerprints them for the app and `svelte-package` copies them into `package/assets/` for consumers. That fixes Inter's gap at the same time and is the right shape, but it changes the serving path of a font every page already depends on — not something to land inside an emoji-scaling bug fix.

PR #10499, opened 2026-08-04
feat: add multi-select and bulk actions to the Home page

### LINKED ISSUE #10449: feature: add multi-select and bulk actions to the Home page
## Problem

The Home page is the primary place for browsing and managing workspace scripts, flows, and apps, but every management action currently has to be performed one item at a time.

This makes routine cleanup and organization unnecessarily tedious. Examples include:

- discarding many abandoned draft-only items;
- moving several related items into another folder;
- archiving or deleting obsolete items;
- duplicating/forking a group of related scripts and flows.

#10372 recently added bulk draft discard to the Review & Deploy drafts page. That solves the draft cleanup use case once the user opens that dedicated page, but it is only accessible through the drafts banner and does not provide general bulk management from Home.

## Proposed solution

Add a multi-selection mode to the Home page, in both flat-list and tree views.

Selecting one or more items should display a contextual bulk-action bar with actions supported by the current selection.

Suggested actions:

- Move selected items to a folder
- Archive / unarchive selected items
- Delete selected items
- Discard selected drafts
- Duplicate or fork selected items into a target folder
- Rename selected items, where applicable

## Suggested interaction

Multi-selection should remain unobtrusive when it is not being used. Permanently visible checkboxes would add unnecessary visual noise to the Home page.

Suggested behavior:

- Keep the default Home rows visually unchanged.
- Reveal a selection affordance when hovering a row, for example by temporarily replacing its item icon with a selection control.
- Clicking that control enters selection mode without opening the item.
- Once selection mode is active, reveal selection indicators for the other rows and display a contextual bulk-action bar.
- Provide a “Select items” entry in the Home toolbar or overflow menu for discoverability.
- Support Shift-click for ranges and Cmd/Ctrl-click for non-contiguous selection.
- Allow Escape or a visible Cancel action to leave selection mode.
- Keep the normal row click behavior unchanged outside selection mode.

The contextual bar could display:

`3 selected · Move · Archive · Duplicate · More…`

Destructive actions such as Delete or Discard should be visually separated under “More…” or placed at the end of the bar.

Only actions compatible with the current selection should be enabled. Disabled actions should explain whether the limitation comes from permissions, item type, or item state.

## Destructive actions

Archive, delete, and draft discard should remain distinct actions:

- Discarding a draft on a deployed item only removes the draft.
- Discarding a draft-only item permanently removes that item.
- Deleting a deployed item should use the existing deletion/trash semantics.
- Archive should be offered as the safer reversible alternative when supported.

The confirmation modal should group affected items by operation and clearly list permanently deleted paths.

## Permissions and failures

- Items the user cannot modify should not be selectable for the relevant action.
- Mixed selections should clearly show which actions are available.
- Bulk operations should report failures per item rather than hiding partial success.
- The Home list and counts should refresh once after the operation.

## Possible implementation stages

1. Multi-selection foundation and contextual action bar
2. Move, archive, discard, and delete
3. Duplicate/fork
4. Advanced batch rename

## Why Home?

The dedicated Review & Deploy page is useful for deployment-oriented draft management. Home is the natural file-manager-like surface for organizing and cleaning up workspace content regardless of draft state.

rubenfiszel:
## Summary

Closes #10449.

The Home page is where workspace scripts, flows and apps are browsed, but every management
action had to be performed one item at a time. This adds a multi-selection mode with a
contextual action bar, in both the flat list and the tree view.

Selection stays out of the way until it is used: the default row is visually unchanged, and a
row's kind icon swaps to a checkbox on hover. Clicking that checkbox enters selection mode
without opening the item; from then on every row shows its checkbox and clicking a row card
toggles it. Shift-click selects a range, Escape or the bar's Cancel leaves.

Covers stages 1 and 2 of the issue's plan (selection foundation + move / archive / discard /
delete). Stages 3 (duplicate/fork) and 4 (batch rename) are not included.

## Changes

**Selection**
- `home/homeSelection.svelte.ts` — a `HomeSelection` class published through context, so the
  tree's nested levels don't have to carry it; `home/Item.svelte` is its only reader, which
  covers the flat list and the tree at once.
- Shift-click ranges resolve visual order by reading `data-home-row-key` back out of the DOM —
  the tree nests rows and pages them in lazily, so no single array holds the rendered order of
  both views. Either end being off-screen falls back to a plain toggle.
- `common/table/Row.svelte` gains an opt-in `rowSelection` prop: the kind icon and the checkbox
  are stacked in a fixed 16px box and swapped with `visibility`, so nothing shifts. The row's
  title link is suppressed while a selection is active. Rows without the prop (the compare &
  deploy layout, `Dev.svelte`) are unaffected.
- A "Select items" toolbar button enters selection mode before the first pick.

**Bulk actions** (`home/bulkActions.ts`, `home/BulkActionsBar.svelte`)
- Move to folder, Archive / Unarchive, Delete, Discard drafts. Each runner performs exactly
  what the matching per-row menu entry does, once per item, so a batch can never take a path
  the single-row action wouldn't.
- Delete and discard stay distinct, per the issue: discarding a deployed item's draft only
  removes the draft, discarding a draft-only item removes the item, and a draft-only item is
  therefore not deletable — there is nothing deployed at its path.
- Only actions compatible with the selection are enabled, and the reason the rest are out is
  surfaced on the disabled button and in the modal ("only scripts and flows can be archived",
  "you are not an owner of this path", …). The confirmation modal groups affected items by
  outcome and names every path that goes for good.
- Batches run item by item and never abort on a failure: each item gets its own outcome, and
  the modal stays open listing the failures with a Retry. Failures and rows the batch skipped
  as ineligible both stay ticked; only what actually changed leaves the selection.
- Move targets are limited to folders and user spaces the caller owns, so a move can't fail as
  a permission error that could have been prevented.

**Gating** — the feature is off for operators and wherever the list is rendered without edit
buttons (a protected workspace). Per-action gates mirror the row menus: script delete requires
workspace admin, flow delete/archive requires ownership, app delete requires write, move
requires ownership (what `MoveDrawer` enforces).

## Test plan

- [x] `frontend/src/lib/components/home/bulkActions.test.ts` — 7 cases pinning `blockedReason`
      per kind/state and `movedPath` preserving sub-path segments
- [x] `npm run check` — 0 errors (65 warnings, all pre-existing)
- [x] Flat list and tree view: hover affordance, click-to-enter, shift-click ranges (including
      across an expanded folder boundary in the tree), row-card click toggling instead of
      navigating, Escape to leave
- [x] Every action end-to-end against a seeded workspace, DB verified after each: archive +
      unarchive a script and a flow; move two scripts `f/alpha → f/beta`; delete two apps and a
      script; discard a draft-only script, a draft-only flow, and a draft on a deployed script
- [x] Mixed selections: an app in an Archive batch is counted out and the modal explains why
- [x] Partial failure (one delete forced to 500): the failed row and the skipped ineligible row
      both stay selected, the successful one leaves, and the modal lists the per-item error
- [x] Raw apps (an `app` row carrying `raw_app`): selectable, move and draft-discard both route
      through the right kind
- [x] Keyboard: Escape with a confirmation dialog open dismisses only the dialog; Enter confirms
      the dialog without ticking the highlighted row into the batch; ArrowRight still reaches a
      row's action buttons in both modes

## Screenshots

Hover reveals the selection control in place of the kind icon — the rest of the list is unchanged.
The toolbar entry is the icon on the left of the controls row:

hover-affordance

Move, with the resulting paths spelled out and the ineligible rows accounted for:

move-modal

Discard, grouped by outcome — what is removed for good vs. what keeps its deployed version:

discard-modal

Tree view, same affordance and bar:

tree-selection

---
## Summary by cubic
Adds multi-select to the Home list and tree with bulk Move, Archive/Unarchive, Delete, and Discard to manage items faster without changing the default row layout. Implements stages 1–2 of Linear #10449; duplicate/fork and batch rename are not included.

- **New Features**
  - Multi-select mode: checkbox replaces the kind icon on hover or via the “Select items” toolbar icon; row click toggles while active; Shift-click selects ranges across list and tree; Esc exits.
  - Bulk actions: Move, Archive/Unarchive, Delete, Discard drafts. Only eligible actions enable with clear reasons; the modal lists paths and per-item results with retry. Batches run item-by-item; failed/ineligible stay selected; move targets are limited to owned folders/spaces.
  - Keyboard: Enter toggles the highlighted row in selection mode; arrow keys still reach action buttons; the title link is suppressed while selecting.

- **Bug Fixes**
  - Keyboard safety: ignore Enter when a button/link has focus and let open dialogs own the keyboard; Esc exits selection without closing modals.
  - Selection robustness: selection survives filter changes and reloads; items that vanished or re-keyed are dropped; already-selected rows refresh their snapshot after list updates.
  - Checkbox reliability: added `onClick` to capture Shift for range select and re-assert checked state to avoid visual mismatch.

<sup>Written for commit 101c531ed679e0b10dff1deadafb73f4e3391eb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/common/checkbox/Checkbox.svelte
+++ b/frontend/src/lib/components/common/checkbox/Checkbox.svelte
@@ -14,6 +14,10 @@
 		class?: string | undefined
 		/** Change handler (controlled — the parent owns `checked`). */
 		onChange?: (e: Event & { currentTarget: EventTarget & HTMLInputElement }) => void
+		/** Click handler, for the cases that need the modifier keys a `change`
+		 * event doesn't carry (e.g. shift-click to select a range). Fires before
+		 * `change`, so calling `preventDefault` here suppresses both. */
+		onClick?: (e: MouseEvent & { currentTarget: EventTarget & HTMLInputElement }) => void
 	}
 
 	let {
@@ -22,17 +26,35 @@
 		disabled = false,
 		title = undefined,
 		class: className = undefined,
-		onChange
+		onChange,
+		onClick
 	}: Props = $props()
+
+	let inputEl: HTMLInputElement | undefined = $state()
+	let clicks = $state(0)
+	// A click flips the box before any handler runs, and flips it back when the
+	// handler prevents the default — both outside Svelte's diff. So a controlled
+	// checkbox whose value did not change across a click keeps the browser's
+	// guess. Re-assert the controlled value after every click, not only when it
+	// differs from the last rendered one.
+	$effect(() => {
+		clicks
+		if (inputEl) inputEl.checked = checked
+	})
 </script>
 
 <input
+	bind:this={inputEl}
 	type="checkbox"
 	{checked}
 	{indeterminate}
 	{disabled}
 	{title}
 	onchange={onChange}
+	onclick={(e) => {
+		onClick?.(e)
+		clicks++
+	}}
 	class={twMerge(
 		'rounded max-w-4 w-full',
 		// When disabled, grey it and let hover fall through to a wrapping trigger
--- a/frontend/src/lib/components/common/checkbox/Checkbox.svelte
+++ b/frontend/src/lib/components/common/checkbox/Checkbox.svelte
@@ -32,11 +32,9 @@
 
 	let inputEl: HTMLInputElement | undefined = $state()
 	let clicks = $state(0)
-	// A click flips the box before any handler runs, and flips it back when the
-	// handler prevents the default — both outside Svelte's diff. So a controlled
-	// checkbox whose value did not change across a click keeps the browser's
-	// guess. Re-assert the controlled value after every click, not only when it
-	// differs from the last rendered one.
+	// A click flips the box before any handler runs, outside Svelte's diff — so a
+	// controlled checkbox whose value does NOT change across a click would keep
+	// the browser's guess. Re-assert after every click, not only on a change.
 	$effect(() => {
 		clicks
 		if (inputEl) inputEl.checked = checked
--- a/frontend/src/lib/components/common/table/AppRow.svelte
+++ b/frontend/src/lib/components/common/table/AppRow.svelte
@@ -11,6 +11,7 @@
 	import { createEventDispatcher } from 'svelte'
 	import Button from '../button/Button.svelte'
 	import Row from './Row.svelte'
+	import type { RowSelection } from './rowSelection'
 	import InheritedLabels from '$lib/components/InheritedLabels.svelte'
 	import Badge from '../badge/Badge.svelte'
 	import {
@@ -48,6 +49,7 @@
 		menuOpen?: boolean
 		showEditButton?: boolean
 		keyboardSelected?: boolean
+		rowSelection?: RowSelection
 	}
 
 	let {
@@ -60,7 +62,8 @@
 		depth = 0,
 		menuOpen = $bindable(false),
 		showEditButton = $bindable(true),
-		keyboardSelected = false
+		keyboardSelected = false,
+		rowSelection = undefined
 	}: Props = $props()
 
 	const dispatch = createEventDispatcher()
@@ -112,6 +115,7 @@
 	canFavorite={!app.draft_only}
 	{depth}
 	{keyboardSelected}
+	{rowSelection}
 >
 	{#snippet badges()}
 		{#if app.execution_mode == 'anonymous'}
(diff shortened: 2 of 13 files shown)


rubenfiszel:
✅ Review round clean @ 101c531ed679e0b10dff1deadafb73f4e3391eb7

PR #10500, opened 2026-08-04
feat: deploy a raw app from its sources, bundling them on a worker

rubenfiszel:
## Summary

Follow-up to #10495, which merged while this was being built (so this branch is on main, not stacked). That PR stopped `/apps/update` from converting a raw app into a broken low-code one; it left MCP with **no way to change a raw app at all**, because the only raw-app deploy endpoint is `multipart/form-data` carrying an already-compiled js/css bundle, which the MCP tool generator can't express and an agent can't produce anyway.

This adds the missing half: **`POST /w/{workspace}/apps/update_raw_source/{path}`**, a JSON endpoint that takes the app's *sources* and compiles them server-side. It is exposed as the MCP tool `updateAppRawSource`, so an agent that reads an app with `getAppByPath` can now edit its `files` and deploy it.

## How the compile runs

As an ordinary bun job on a worker — no new job kind, no new executor:

```
POST /apps/update_raw_source  →  404 / kind guard (before paying for a compile)
                              →  push a bun job with the sources, wait for its result
                              →  update_app_internal(raw_app = true) + store the bundle
```

**The build is `wmill app bundle`, not a bundler of its own.** A new CLI subcommand exposes the build `wmill app push` already runs, and the job shells out to it, so an app deployed through the API is compiled by the same esbuild pass as the CLI and the editor — same entry-point pick, virtual `wmill` module, `/ui/` shared-UI resolution, Svelte/Vue plugins and production defines. A React app that the editor bundles to 187,124 bytes comes out of this endpoint at 187,140 (the 16 bytes are the edit itself). The alternative — reimplementing the build with `Bun.build` — was a third bundler to keep in step with the other two, and it silently diverged: no `/ui/` resolution, no framework plugins, and a bundle that needed its own `NODE_ENV` define to stop shipping React's dev build.

The job gets the logs, timeout, cancellation and attribution every job gets, and shows up in Runs as `bundle raw app`. Because the compile happens *before* the version is written, a build failure leaves the deployed app untouched and returns the compiler's own error:

```
Bad request: raw app bundling failed (job 019fcc07-…): Could not resolve: "./does-not-exist"
  App.tsx:1:22  import { nope } from "./does-not-exist"
```

which is exactly what an agent needs to fix its own code.

The CLI is pinned to the server's own version (`windmill-cli@{GIT_SEM_VERSION}`), which is the release the CLI ships alongside, so a deploy is compiled by the bundler that belongs to it and there is no pin to bump by hand. A dev build has no matching npm version: set `WM_RAW_APP_BUNDLER_CLI` to the whole command instead (that is how this was verified — `bun run cli/src/main.ts app bundle`).

## Changes

- `cli app bundle [folder] --out <dir>` — the existing `createBundle` exposed on its own. Result goes to files, not stdout, because the build logs to stdout as it runs.
- `apps_raw_bundle.rs` — pushes the bundle job and waits for it. The bundle comes back gzipped+base64 so a large app stays well inside `MAX_RESULT_SIZE_MB`, which a deployment can set far below the 500MB default.
- `apps_raw_bundler.ts` — writes the sources, `bun install --ignore-scripts`, runs the CLI, returns the bundle.
- `update_app_raw_source` handler + route, with the same operator/deploy-rule checks as `update_raw`.
- The kind guard from #10495 is factored into `reject_kind_change` so this endpoint can refuse before queueing a compile.
- `updateApp`'s MCP instructions now point at the new tool instead of saying raw apps can't be updated.

## Review findings addressed

From the round on the first revision, plus a local review:

- **Arbitrary code execution for an `apps:write`-only token (P0).** Bun executes `with { type: 'macro' }` imports at *bundle* time, so the previous `Bun.build` version ran caller-supplied code on the worker; esbuild has no macros, which removes the mechanism. Independently, the endpoint now requires **`jobs:run` as well as `apps:write`** — it makes a worker run a build on caller-supplied input, so it must not hand that to a token that couldn't already do it. `bun install --ignore-scripts` keeps lifecycle scripts out of it.
- **Compile queued before any authorization (P0, same finding).** The handler now 404s when there is no app at the path, so nothing compiles for a path the caller can't write.
- **`/ui/` shared-UI imports didn't resolve (P1).** The CLI's resolver plugin handles them.
- **Wait wasn't actually capped (P1).** `BUNDLE_TIMEOUT_SECS` bounds the run, not the wait; added the `check_queue_too_long` guard every other sync-wait surface uses, and the comment now says which is which.
- **Missing `tally_app_rename` on rename (P2).** Added, like the sibling handlers.
- **Kind pre-check read through the privileged pool (P2).** Now through `user_db`, so it can't disclose an app's existence or kind to someone who can't read it.
- **Job-result size cap (P2).** Bundle is gzipped in the result.

## Test plan

Verified against the dev backend with a real React raw app (`u/admin/zippy_app`), checking `app_version` / `app_bundles` rows after each call.

- [x] `POST /apps/update_raw_source` with edited sources → 200, new version `raw_app = true`, **187,140-byte** bundle stored and served, `bundle raw app` job succeeded
- [x] the served bundle is the CLI's esbuild output (`(()=>{var ud=Object.create;…`), not the previous `Bun.build` one, and contains the edit
- [x] `wmill app bundle` on the same sources locally → 187,124 bytes, i.e. the editor's build
- [x] the app's backend runnable works through the bundle's virtual `wmill` bridge — "Run 'a'" ran `u/admin/zippy_app/a` to `success`, result `42` (verified on the previous revision; the bridge comes from the same virtual module)
- [x] unresolvable import → 400 with the compiler's file/line, **no new version**
- [x] low-code app at the path → 400 from the kind guard, no bundle job queued
- [x] `cargo test -p windmill-api-integration-tests --test apps` — 2 passed; `SQLX_OFFLINE=true cargo check -p windmill-api` clean; CLI unit tests pass except two pre-existing `EDQUOT` failures in `resource_folders_unit.test.ts` (reproduced with this branch stashed)
- [ ] browser render of *this* bundle — the dev server kept failing module loads with `ERR_INSUFFICIENT_RESOURCES` on this box; the previous revision's bundle was verified rendering, and this one is byte-comparable to the editor's

Deployed from sources through the JSON endpoint (previous revision's bundle, same endpoint and flow):

source-deploy


rubenfiszel:
Round 2 addressed in d2c0a7f15c:

- **claude P1 — `bunx` isn't in the worker images.** Confirmed: the Dockerfiles copy the `bun` binary alone. Now `bun x`, which is the same resolver reachable through the binary that is there. Verified on this box's worker: the job resolved and downloaded `windmill-cli` rather than dying on ENOENT (it then failed on the point below, which is what proved the spawn works).
- **claude P1 — `jobs:run` 403s every MCP call.** Correct, and the tool this PR exists for was unusable. The MCP proxy mints one route scope per call and `includes` never crosses domains, so `jobs:run` is now minted alongside `apps:write` for this route (`extra_scopes_for_route` in `mcp/utils.rs`). The requirement stays — the endpoint makes a worker build caller-supplied sources.
- **codex P1 — documented entry points fail.** `app bundle` now picks the first of `index.tsx`/`index.ts`/`index.js` that exists, framework preference first, and errors listing what it looked for. Verified an `/index.ts`-only app bundles (it did not before).
- **codex P1 — `app bundle <folder>` deletes the caller's `dist`.** Real: `createBundle` `rmSync`s `<cwd>/<outDir>` and defaulted to `dist`. It now gets an outDir that is ours to delete. Verified a sentinel file in the caller's `dist/` survives.
- **codex P1 — public helper without an authorization contract.** `bundle_raw_app_sources` is `pub(crate)` and checks `jobs:run` itself.
- **codex/claude P2 — shared UI resolved differently from `app push`.** Now `<cwd>/ui`, same as push (and the same directory inside the job).
- **claude P2 — `bun install` keyed on the exact string `/package.json`.** Both spellings now.
- **claude P2 — build logs never reached the job.** Both spawns echo their output; a successful `bundle raw app` job now carries the install and build logs.

One thing I could not verify and it is worth being explicit about: **the stock-worker path needs a CLI release that contains `app bundle`.** The pin is the server's own release (`windmill-cli@{major}.{minor}.{patch}`, git-describe suffix dropped so a dev build resolves the last release rather than a nonexistent version). On this dev build it correctly fetched `windmill-cli@1.778.0` and failed with that CLI's usage output, because `app bundle` ships with this PR. So the release has to carry both halves; until then `WM_RAW_APP_BUNDLER_CLI` is the escape hatch, and that is how the end-to-end run below was done.

Re-verified end to end after the fixes: new version 27, `raw_app = true`, 187,140-byte bundle stored and served, build logs in the job.

rubenfiszel:
Round 3 addressed in fb9f3ce39e:

- **codex P1 — unbounded decompression.** Right: `MAX_RESULT_SIZE_MB` caps the compressed result, and `GzDecoder::read_to_string` had no bound on what it expands to, from sources the caller supplied. Now read through `.take(limit + 1)` and rejected past the limit, which is the same size `/apps/update_raw` accepts as a whole body.
- **codex P1 — package-less apps fail on slim workers.** Also right, and it would only have shown up on a slim image: with no manifest the job installed nothing, the CLI then shelled out to `npm`, which those images don't ship. It now creates the empty `node_modules` the CLI's check looks for, so nothing shells out.
- **claude nit — `bun x` needs npm reachability on an image that already ships `wmill`.** The job now prefers the installed `wmill` when there is one (`Bun.which`), which is the same release the server is, and falls back to the pinned `bun x windmill-cli@…`. `WM_RAW_APP_BUNDLER_CLI` still wins over both.
- **claude nit — `hasPackageJson` missed `./package.json`.** Normalises leading `./` too now.
- **claude nit — the minted-scope fix had no test.** `proxy_jwt_raw_app_source_deploy_also_gets_jobs_run` pins that this route mints `apps:write` *and* `jobs:run`; it fails if either goes missing. 5 passed under `--features mcp`.

One thing that surfaced on the way: regenerating the MCP tools produced a file that doesn't compile under `--features mcp`. The generator writes schemas with `json.dumps`, which escapes a non-ASCII character as `\\uXXXX` — valid JSON, invalid Rust (it spells it `\\u{XXXX}`). An em dash added to an openapi description since the last regen was enough. Fixed at the generator (`ensure_ascii=False`) rather than in the generated file, so the next regen doesn't hit it. This is why `cargo test --features …,mcp,…` matters here — a plain `cargo check` never compiles that module.

rubenfiszel:
Round 4 addressed in 2453723e03. The P0 changed my mind about the authorization design, so writing that out.

**codex P0 — caller-controlled Svelte compiler executes on the worker.** Verified: `loadSvelteCompiler` resolves `svelte/compiler` from the *app's* node_modules and `import()`s it, deliberately — compiler and runtime have to agree on internals, and that is the CLI's documented reason. `--ignore-scripts` doesn't stop an import. So bundling caller-supplied sources **is** code execution, and not only for Svelte.

That can't be fixed in the bundler without breaking the invariant the CLI is protecting, so the fix is the boundary: **I reverted the `jobs:run` minting I added in round 2.** The handler still requires `jobs:run`; the MCP proxy no longer hands it out for holding the tool. A scope-restricted MCP token now has to carry `jobs:run` itself, which is honest — using this tool runs a build. Unscoped/full tokens are unaffected, so the common setup still works. The tool instructions say so, and the unit test now pins that the route mints `apps:write` **only** (it would have caught the escalation).

- **codex P1 — twice the configured limit.** Right: each call got the full budget. They now share one — the css decode gets `limit - js.len()`.
- **claude P2 — preferred `wmill` unpinned with no gate or fallback.** Also right, and it would have broken every image built before the next release. The job now runs `wmill --version` and uses it only when it reports this server's release, else falls back to the pinned `bun x windmill-cli@X.Y.Z`.
- **claude P2 — `hasPackageJson` normalisation still disagreed with `write()`.** Both now go through one `target()` that does the same `path.join`, so `foo/../package.json` is seen and `.package.json` isn't mistaken for a manifest.
- **claude P2 — oversized bundle could surface as a 500 "not valid gzip".** Reads into `Vec<u8>` and checks the length before `String::from_utf8`, so the size error stays a 400.

`cargo check -p windmill-api --features mcp` and the 5 scope tests pass.

rubenfiszel:
Round 5 addressed in 42a1d274dc. Both P1s were right and the first one needed a design answer, not a patch.

**claude P1 — the tool 403s for every MCP token the product can create.** Correct: `ScopesPicker` in mcp mode emits only `mcp:…` parts, so a token that selects this tool can't satisfy the handler's `jobs:run`, and the only way to get one was hand-writing scopes through the API. Reverting the mint (round 4) closed an escalation but left the deliverable unreachable.

The fix is to let the requirement be **granted, visibly, at token creation** instead of minted behind the admin's back at proxy time:
- `x-mcp-extra-scopes: [jobs:run]` on the route in openapi, with the reason;
- the generator carries it into `mcpEndpointTools.ts` as `extraScopes`;
- `McpScopeSelector` adds those scopes to the token when the tool is selected (and in folder mode, where every endpoint is granted).

So selecting the tool grants `jobs:run` because using it runs a build — the admin sees the scope on the token, and the proxy's pass-through branch preserves it. The unit test still pins that the *proxy* mints `apps:write` only.

**codex P1 / claude P1 — the version probe accepts the CLIs it exists to reject.** Right on both counts: `wmill --version` prints its own version *and* npm's latest, so `.includes(server_version)` matched the wrong line — and the probe made an npm call, which is the cost the installed-CLI branch exists to avoid. The probe is gone. The job now just runs the installed `wmill app bundle`, and falls back to the pinned `bun x windmill-cli@X.Y.Z` **only** when the output is cliffy's "unknown command"/usage text; any other failure is a real build error and is reported as one.

Verified on this box, which happens to be exactly the flagged configuration — `wmill` 1.747.0 installed, predating `app bundle`. The job log shows the stale CLI printing `Usage: wmill app`, then the fallback running and `✅ Bundle created successfully`; version 28 stored with a 187,135-byte bundle. (The ANSI codes cliffy puts between `Usage:` and `wmill app` broke my first attempt at that match — stripped before testing now.)

**claude P2 — the css size error reported the remaining budget as the limit.** Now says how much of the shared budget is left. Also fixed the missing apostrophe in the tool instruction.

`cargo check -p windmill-api --features mcp` and `npm run check:fast` clean.

rubenfiszel:
Round 6 addressed in c3926bf9eb. Both reviewers found the same two holes in round 5's fix, and both were right — the grant reached the token but never reached the handler.

- **The proxy discarded the scope the selector had just granted.** `jwt_scopes_for_proxied_route`'s mint branch returns `[route_scope]` and nothing else, so a token holding `["mcp:endpoints:updateAppRawSource", "jobs:run"]` still arrived as `["apps:write"]` and 403'd. It now carries over an extra scope **only when the caller already holds it** — no escalation (round 4's finding stands), and the deliberate grant survives. `extra_scopes_for_route` is back for that, mirroring `x-mcp-extra-scopes`.
- **`mcp:favorites` / `mcp:all` exposed the tool while granting nothing.** Those modes emit a single scope part and my round-5 change only touched the endpoint-selection and folder branches. They now grant what the exposed tools require too.

The test pins both directions: a bare `mcp:` token still mints `["apps:write"]` alone, and one granted `jobs:run` gets `["apps:write", "jobs:run"]`. It would have caught this round's bug as well as round 4's.

5 scope tests pass under `--features mcp`; `npm run check:fast` clean.

rubenfiszel:
Round 7 addressed in 67571472e0 — by reverting rounds 5 and 6 on the token side. claude's P0 is correct and the fix I shipped two rounds ago was worse than the problem it solved.

**What went wrong.** Putting `jobs:run` on the MCP token widens the *token*, not the tool. These are ordinary bearer tokens, so `mcp:favorites` (the create-token default) and `mcp:all` (the OAuth consent default) came out able to `POST /jobs/run/preview` — arbitrary code execution — with nothing on the consent screen saying so. Strictly worse than what round 4 objected to.

**Where it landed.** Back to minting in the per-request JWT, which is the *confined* option and I should have argued that in round 4 rather than reverting:

- the JWT is created server-side for one proxied call and never reaches the client, so the MCP token stays `mcp:`-only and can't reach any job endpoint;
- the grant applies to this route and no other;
- a direct API caller still needs `jobs:run` on its own token, so an `apps:write`-only token gains nothing.

The residual is real and worth stating plainly: **ticking this tool grants the agent the ability to run a build on a worker**, because compiling an app's sources runs the app's own dependencies (the CLI loads the app's Svelte compiler). That is what the tool *is* — no mechanism makes it otherwise short of not offering it. The tool description says the sources are compiled on a worker; the mint site now carries the reasoning. If you'd rather that not be implied by the tool grant, the honest alternative is to drop the tool, not to move the scope around.

This also fixes both round-7 side effects for free: no token leaves the MCP scope editor (claude P2 / codex nit), and no unused `x-mcp-extra-scopes` plumbing remains in openapi, the generator or the TS tool list.

- **codex nit — `value` was optional on the source deploy.** Now `required: [value]`, so the schema matches the handler.

Test asserts the mint again (`["apps:write", "jobs:run"]` for a bare `mcp:` token); 5 pass under `--features mcp`, `npm run check:fast` clean.

rubenfiszel:
Round 8 addressed in 9908ba11e0.

**codex P0 — the default modes granted it implicitly.** Right, and it's the hole in my round-7 argument: "the admin consents by ticking the tool" only holds if they tick it. `mcp:favorites` (create-token default) and `mcp:all` (OAuth default) reach every endpoint without naming any, and their UI talks about scripts and flows.

The extra scope is now minted **only for a token that names this tool** in its `mcp:endpoints:` list. `mcp:all`, `mcp:favorites` and `mcp:endpoints:*` all mint `apps:write` alone, so the handler 403s for them — the capability is reachable only by an explicit selection. Test covers all four cases.

**claude P2 — the match was on a substring of a caller-influenced path.** Also right: a script named `f/apps/update_raw_source/x` proxies to `/jobs/run/p/f/apps/update_raw_source/x` and matched `contains(...)`. Nothing was gained by it, but a caller-chosen string was deciding what its own request got minted. Now matched on segments after `/w/{ws}/`, with a test that the spoofed run path mints `jobs:run:scripts` and nothing else.

**claude P2 — the doc oversold the JWT.** `create_jwt_token(..., 3600, ...)` is scope-bound for an hour, not route-bound. The comment now says what actually carries the argument: the JWT is built here and handed to the internal request, never to the client.

Also noting claude's round-8 observation, which settles the design question rounds 4–7 circled: `runScriptPreviewAndWaitResult` is already an MCP tool whose route deliberately derives the broad `jobs:run` for exactly this reason, so per-request minting is the existing mechanism for "a worker will run what you supply", not a new class of grant. With this round it is also strictly narrower than that precedent, since it requires the tool to be named.

5 scope tests pass under `--features mcp`.

rubenfiszel:
Round 9 addressed in 58fa13bc2e.

- **codex P1 — the compile ran before write authorization.** Correct: `deployed_app_kind` is a `SELECT`, so read access passed the preflight and a read-only member's sources were built on a worker before RLS refused the write. `require_is_writer` now runs before `bundle_raw_app_sources` — the same owner / folder-writer / `extra_perms` semantics the resource and variable editors use.
- **claude P2 — the consent surface never mentioned the worker.** The picker renders the tool's *summary*, and the consequence lived only in `x-mcp-instructions`, which the agent reads and the admin doesn't. The summary now says it: "update a raw app from its sources, compiling them on a worker (which runs the app's own dependencies to do so)". Regenerated into both artifacts.
- **claude P2 — the ANSI strip.** Worth correcting the diagnosis: the regex was matching a *literal ESC byte* that my earlier edit had written into the source, so stripping did work and `Usage:` was matching — but an invisible control character in a source file is exactly the fragility the finding is about, and it's why the code read as broken. It's `\\x1b\\[[0-9;]*m` now.
- **claude's note on re-implementing the grammar.** Taken: `selects_endpoint_tool` calls `parse_mcp_scopes` from `windmill-mcp`, which owns it. `mcp:all` pushes `*` into `endpoints` there, so the "named, not blanket" property still holds, and it follows the grammar if it grows a form. Same 5 tests pass.

Not doing, and saying why: **the tool stays listed to tokens the gate refuses.** Filtering the advertised list by whether the token named the tool would make the list depend on the token's own scopes, which is a change to how every MCP tool is advertised, not to this one. The first call returns `Required scope: jobs:run`, and the tool's summary now says what it needs — I'd rather leave that visible than special-case the listing here.

rubenfiszel:
Round 10 addressed in ee4b9f5ada. claude and pi are Good to merge; codex found one more, and it's right.

**codex P1 — group-owned apps were rejected by the preflight.** `require_is_writer` knows `u/` owners, `f/` folder writers and `extra_perms`, but not `g/<group>/` — while the app's own RLS (`see_member`, `FOR ALL`) grants group members write on those paths. So a group member who can update `g/team/app` through every other endpoint would have been refused by this one.

Rather than add a `g/` branch, the preflight now **asks the database**: an `UPDATE app SET path = path … RETURNING 1` through `user_db`, rolled back. If RLS allows the write, it returns a row; if not, it doesn't. That is the same predicate the real write is subject to, so it can't drift from the policies — which is exactly how this bug arose. `path = path` touches no column any trigger watches (the two on `app` are `UPDATE OF policy` with a `WHEN` clause, and `DELETE`), so the probe has no side effects.

Re-verified end to end after the change: version 29 deployed, `raw_app = true`, 187,136-byte bundle. Offline sqlx check clean with the one new cache entry.

On the two regression cases you asked for: the read-only case is what the probe returns false for, and the group case is what it now returns true for — but both need a second, non-admin user in the integration harness, which the apps fixture doesn't have. I'd rather rest it on the probe being RLS by construction than add a fixture that only exercises one of the two shapes; say the word if you want the fixture anyway.

Still explicitly unverified, unchanged from earlier rounds: the browser render of the latest bundle (the dev server on this box keeps dying with `ERR_INSUFFICIENT_RESOURCES` on that route). The bundle is byte-comparable to the editor's and an earlier revision's was verified rendering with its runnable working.

rubenfiszel:
✅ Review round clean @ 7abe5da0767e529a08ba52b4185be653f8b737ad — nit-only verdicts at ee4b9f5ada8929dd19b0d6f2aa14a9f0d8f257ea (codex: Good to merge, pi: Good to merge, claude: mergeable with one nit); the nit (write denial answered 401 where it means 403 — `Error::NotAuthorized` maps to UNAUTHORIZED, `PermissionDenied` to FORBIDDEN) is fixed in 7abe5da076.

PR #10501, opened 2026-08-04
fix: locate coursier artifacts by coordinate so private maven registries work

rubenfiszel:
## Summary

Java jobs failed to compile on any worker configured with a private maven registry, with a
`package … does not exist` error for every dependency even though the install phase reported
success.

`install()` fetched artifacts into a throwaway coursier cache and then moved them into the
shared repository by walking to a **hardcoded depth of 3**. Coursier lays its cache out as
`<scheme>/<host>/<registry url path…>/<group as path>/<artifact>/<version>/`, so depth 3 only
lands on the maven root when the registry URL has exactly one path segment — which is true of
Maven Central (`maven2`) and of almost nothing else:

| registry URL | depth 3 lands on | jars end up at | classpath expects |
|---|---|---|---|
| `…/maven2` | `maven2` | `repository/com/google/…` | ✅ |
| Nexus `…/repository/maven-public` | `repository` | `repository/maven-public/com/google/…` | ❌ |
| Artifactory `…/artifactory/libs-release` | `artifactory` | `repository/libs-release/com/google/…` | ❌ |
| root-hosted `https://maven.intra/` | `com` | `repository/google/code/…` | ❌ |

The classpath is built separately as `repository/<group as path>/<artifact>/<version>/*`, so
every entry pointed at a directory that did not exist — and `javac` ignores missing classpath
entries without a word. The failure surfaced only as a wall of compiler errors well after a
green "Fetching 19 packages… Done".

Verified on a real coursier tree fetched from a 3-segment registry: the old walk copies
`repositories/` to the repository root while the classpath looks for `com/google/…`.

The same jobs were also being restarted mid-resolution as false zombies, which this fixes too.

## Changes

**Locating the artifacts**

- Locate each artifact by its `<group as path>/<artifact>/<version>` coordinate instead of a
  fixed depth, so the number of path segments in the registry URL no longer matters.
- Match on path components rather than raw strings, so Windows' native separator cannot defeat
  the comparison.
- Try the longest coordinate first, so a group id ending in another one's coordinates
  (`com.org.foo:bar` over `org.foo:bar`) cannot claim the shorter one's directory.
- Only accept a coordinate directory that actually holds the artifact. Coursier leaves an
  **empty** directory at the coordinate under every repository it probed and 404'd on, and the
  default repositories are probed before the configured ones — so with a private registry and
  `no_default_maven` off, an internal artifact has an empty directory under Maven Central and
  the real one under the private registry, and which is walked first is `read_dir` order.
  Claiming the leftover would have cached an empty directory as the installed artifact, written
  its `.valid.windmill` marker, and (on EE) pushed it to S3 as the canonical tar for the fleet.
  Coursier's own bookkeeping next to a cached artifact (`.<name>.checked`, `.<name>.error`,
  checksums) is dot-prefixed, so only a non-hidden entry counts as the artifact having landed —
  a plain 404 leaves the directory truly empty, but a failed or partial download leaves those.

**Surfacing failures**

- Fail the job naming the artifacts when coursier reports success but an expected coordinate is
  absent from its cache. Previously `mark_success` wrote validity markers for artifacts that
  were never copied; the marker write then failed into `tracing::error!` only, invisible to the
  user, and every subsequent run silently re-fetched the whole dependency set.
- Reclaim the temporary fetch dir on every path out of `install()`, not just the success path —
  coursier failing against the registry is the likely case here and it bails before the
  postinstall step runs.

**False zombie restarts**

- Keep the job's ping fresh during java lockfile resolution: `resolve()` runs coursier through a
  raw `cmd.output()` (the lockfile has to be read from a clean stdout, so it cannot go through
  `handle_child`'s polling loop), so nothing refreshed `last_ping` and a resolution against a
  slow or distant registry was reaped as a zombie and restarted.
- Bound that same call with `resolve_job_timeout` and `kill_on_drop`, since the zombie reaper
  was incidentally the only thing turning a wedged resolution into a terminal state.
- Same heartbeat for the batch dependency install: `handle_child` only pings the job id it is
  handed, and that one is nil whenever output is buffered to a string, so nothing refreshed the
  ping for the whole install.

## Test plan

- [x] `cargo test -p windmill-worker --features java java_executor` — 4 unit tests over real
      coursier directory shapes: registry URLs with 0, 1 and 3 path segments; the empty
      404-leftover directory (constructed symmetrically so it fails regardless of `read_dir`
      order — confirmed it does fail without the guard); overlapping coordinates; and the
      missing-artifact error naming the right artifact.
- [x] Real java job (the default cowsay/gson/jackson template) run end to end on an EE build
      (`--features java,quickjs,enterprise,private`) against a **multi-segment private registry**
      (`maven_repos` = a 3-path-segment URL, `no_default_maven` on, zero "requires Enterprise"
      warnings in the job log, so the registry was genuinely used):

      | | repository root | job |
      |---|---|---|
      | before | `repositories` (a stray URL segment) | 8 javac errors, `cannot find symbol ObjectMapper` / `Cowsay` |
      | after | `com commons-cli javax org` | success, and `com/google/code/gson/gson/2.8.9/` holds `gson-2.8.9.jar` |

- [x] Same EE setup with a repository that 404s everything probed *before* the one that serves:
      the real jar lands (not the empty leftover) and all 19 `.valid.windmill` markers are written
      against populated directories.
- [x] Also run on a CE build against Maven Central — no regression on the default path, no
      leftover `tmp-fetch-*` dirs on either the success or the failure path, no worker errors.
- [x] `cargo check -p windmill-worker --features java --all-targets`

### DIFF
--- a/backend/windmill-worker/src/java_executor.rs
+++ b/backend/windmill-worker/src/java_executor.rs
@@ -525,16 +525,19 @@ async fn move_to_repository(
         let (mut subdirs, mut holds_artifact) = (vec![], false);
         let mut entries = tokio::fs::read_dir(dir).await?;
         while let Some(entry) = entries.next_entry().await? {
+            let name = entry.file_name().to_string_lossy().into_owned();
             if entry.file_type().await?.is_dir() {
-                subdirs.push(entry);
+                subdirs.push((name, entry));
             } else {
-                holds_artifact = true;
+                // coursier's own bookkeeping (`.<name>.checked`, `.<name>.error`, checksums) is
+                // dot-prefixed; only a downloaded artifact is not
+                holds_artifact |= !name.starts_with('.');
             }
         }
-        // every repository coursier probed and got a 404 from is left with an empty directory at
-        // the coordinate, so the match only counts where the artifact itself landed: the first
-        // repository tried is a default one, and claiming its leftover would cache an empty
-        // directory as the installed artifact
+        // a repository coursier probed and did not get the artifact from keeps a directory at the
+        // coordinate holding nothing but that bookkeeping, and default repositories are probed
+        // before the configured ones, so claiming the first match would cache an empty directory
+        // as the installed artifact
         if holds_artifact {
             if let Some(w) = wanted
                 .iter_mut()
@@ -545,8 +548,8 @@ async fn move_to_repository(
                 return Ok(());
             }
         }
-        for entry in subdirs {
-            below.push(entry.file_name().to_string_lossy().into_owned());
+        for (name, entry) in subdirs {
+            below.push(name);
             find_and_copy(&entry.path(), below, wanted).await?;
             below.pop();
         }
@@ -1194,13 +1197,15 @@ mod tests {
         let central = format!("{fetch_dir}/https/repo1.maven.org/maven2");
         let nexus = format!("{fetch_dir}/https/nexus.local/repository/maven-public");
         touch(&format!("{central}/com/corp/public/1.0/public-1.0.jar")).await;
-        create_dir_all(format!("{nexus}/com/corp/public/1.0"))
-            .await
-            .unwrap();
+        touch(&format!(
+            "{nexus}/com/corp/public/1.0/.public-1.0.jar.error"
+        ))
+        .await;
         touch(&format!("{nexus}/com/corp/internal/1.0/internal-1.0.jar")).await;
-        create_dir_all(format!("{central}/com/corp/internal/1.0"))
-            .await
-            .unwrap();
+        touch(&format!(
+            "{central}/com/corp/internal/1.0/.internal-1.0.jar.error"
+        ))
+        .await;
 
         let deps = vec![
             dep(&repository_dir, "com/corp", "public", "1.0"),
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ c6801374676d219b6a664512914856471b169ba7 — nit-only verdicts at 490ba28b3cae96986745ec40163f94de46852454 (codex: good to merge, pi: good to merge, claude: nit); nit addressed in c680137467

rubenfiszel:
Closed the verification gap: ran this end to end on an EE build (`--features java,quickjs,enterprise,private`) against a genuinely multi-segment private registry, which is the customer's configuration. `maven_repos` set to a 3-path-segment URL with `no_default_maven` on, and zero "requires Enterprise" warnings in the job log, so the registry was really in use.

| | repository root | job |
|---|---|---|
| before | `repositories` — a stray URL segment | 8 javac errors, `cannot find symbol ObjectMapper` / `Cowsay` |
| after | `com commons-cli javax org` | success, `com/google/code/gson/gson/2.8.9/gson-2.8.9.jar` present |

Also exercised the 404-leftover path on the same EE build, with a repository that 404s everything probed *before* the one that serves: the real jar lands rather than the empty leftover, and all 19 `.valid.windmill` markers sit next to populated directories.

The `Test plan` in the description is updated; no verification gap remains.

PR #10504, opened 2026-08-04
fix: open an AI session from the editor bar's AI button, on the step

rubenfiszel:
## Summary

Follow-up to #10493. The AI button in the inline code editor's toolbar (`EditorBar` →
`FlowInlineScriptAIButton`) still only knows how to open the **docked** chat pane. With AI
Sessions on — the default — that pane is never mounted, so the button did nothing; #10493
turned the dead click into a hidden button. Either way the editor bar lost its AI entry
point, in both places it appears:

- a flow step's inline script (`FlowModuleComponent`)
- a raw app's inline script (`RawAppInlineScriptEditor`)

It now does what the flow graph's toolbar button does — open a fresh AI session with the
item in the preview — except the session preview opens **on the step whose code the user
was editing**, instead of on the flow's metadata tab.

## Changes

**The hand-off carries a focus hint.** `OpenInSessionSource` gains `previewParams`, which
`openEditorInSession` appends to the tab URL. For a flow step that is
`?selected=<moduleId>` — the same param the full-page flow editor reads. It steers the
editor only: every tab resolver strips the query (`stripBase`), so the tab's identity stays
`(kind, path)` and the dedupe is unaffected.

**The in-process preview editor honors it.** A flow preview tab is not an iframe, so
`PreviewTabHost` reads `?selected=` off the tab URL (`parsePreviewSelectedId`) and hands it
to `FlowEditorView`, whose `selectedId` becomes `$derived` — that also clears a stale step
id when the tab is retargeted at another flow.

**A context carries the hand-off down to the toolbar.** The button sits four to five levels
below the component that owns the item (behind a recursive module wrapper), so
`openInSessionContext.ts` publishes it instead of threading a prop:

- `FlowBuilder` hoists the source it already builds for the graph toolbar and publishes it,
  adding `previewParams` when asked for a `moduleId`.
- `RawAppEditor` takes over `persistDraftForSession` from `RawAppEditorHeader` (it already
  has every input it needs), builds the source, publishes it, and passes it to the header —
  so one component owns the app's hand-off, as `FlowBuilder` does for flows. A raw app has
  no addressable sub-editor, so the `moduleId` is ignored there.

**The button wraps in `OpenInSessionButton`**, the same shape `ScriptEditor` already uses,
forwarding the toolbar's `btnProps` through `AIButton` so it keeps the editor bar's flat
styling. Unchanged behaviour: users opted out of the sessions beta still get the docked-chat
wand (the `fallback` snippet), nothing renders inside the session panel (no nested sessions),
and a workspace without AI configured gets an "enable Windmill AI" popover (`AIButton`'s, on
the sessions path — same as the flow graph's button and `ScriptEditor`).

**The hand-off is withheld under `disableAi`**, the same gate as the graph toolbar's
`showFlowAiButton`. Without it an SDK embed that turned AI off would gain an entry point that
navigates the host out to `/sessions` — the docked-chat button it replaces was implicitly
hidden there.

**Monaco is flushed before the hand-off.** The session preview loads the item from its
draft, and `persistDraftForSession` only flushes what the syncer has already parked — the
editor's own trailing debounce holds the last keystrokes for up to ~1s. That is tolerable for
the toolbar buttons that sit outside the code editor, but this one is *in* it, where "type,
then click" is the normal case. `EditorBar` hands the button `editor.flushPendingChanges()`
(the same call Ctrl/Cmd+S makes) and it runs before `beforeOpen`.

## Screenshots

Flow step editor. Before the click — the toolbar's AI button next to the editor settings:

40-nit-before

After the click — a fresh session whose preview opens the flow with **"second step" (`b`)
selected** and its code in view, rather than on the flow's metadata tab:

11-session

Raw app inline script — same button, opens the app in a session (no sub-editor to focus):

22-rawapp-session

Sessions beta opted out — unchanged: the legacy wand next to the docked chat pane:

30-optout

## Test plan

Driven in a real browser against the dev server (flow `u/admin/wand_demo` with two inline
steps, raw app `u/admin/wand_app` with one inline runnable):

- [x] Flow step `b` selected → toolbar AI button → lands on `/sessions`, preview shows the
      flow with `b` selected and its code loaded.
- [x] The flow graph's own AI button is untouched (still opens the flow on its default tab).
- [x] Raw app inline script → toolbar AI button → session preview opens the app.
- [x] Inside the session preview, the step's toolbar renders no AI button (no nested session).
- [x] Type into the step's Monaco and click the AI button in the same breath — the preview
      opens with the keystrokes included, so the flush closes the debounce window:

      crop-flush
- [x] `wm_sessions_beta_optout=1` → the toolbar falls back to the legacy wand and the docked
      chat pane, as before.
- [x] `npm run check` — no new errors. (`src/lib/rawAppDeploy.ts:81` `allow_kind_change`
      fails on `main` too: #10495 added the field to `openapi.yaml` without regenerating
      `src/lib/gen`. Untouched here.)
- [x] `vitest run src/lib/components/sessions/{sessionMode,previewRouter}.test.ts` — 39 passing.

Two unit cases added, pinning the round trip the feature rests on: `withPreviewParams`
produces the `?selected=` tab URL, and `parsePreviewSelectedId` reads it back **while
`resolvePreviewTab` still resolves the same editor** — the non-obvious invariant is that the
query must never leak into tab identity.

### DIFF
--- a/frontend/src/lib/components/EditorBar.svelte
+++ b/frontend/src/lib/components/EditorBar.svelte
@@ -1309,7 +1309,11 @@ JsonNode ${windmillPathToCamelCaseName(path)} = JsonNode.Parse(await client.GetS
 
 			{#if customUi?.aiGen != false}
 				{#if openAiChat}
-					<FlowInlineScriptAiButton {moduleId} btnProps={{ variant: 'subtle' }} />
+					<FlowInlineScriptAiButton
+						{moduleId}
+						flushEditor={() => editor?.flushPendingChanges()}
+						btnProps={{ variant: 'subtle' }}
+					/>
 				{/if}
 			{/if}
 
--- a/frontend/src/lib/components/FlowBuilder.svelte
+++ b/frontend/src/lib/components/FlowBuilder.svelte
@@ -172,9 +172,7 @@
 	// For preserve_on_behalf_of feature
 	let preserveOnBehalfOf = writable(false)
 	let savedOnBehalfOfEmail = writable<string | undefined>(savedFlow?.on_behalf_of_email)
-	let savedOnBehalfOfPermissionedAs = writable<string | undefined>(
-		savedFlow?.on_behalf_of
-	)
+	let savedOnBehalfOfPermissionedAs = writable<string | undefined>(savedFlow?.on_behalf_of)
 
 	// Keep savedOnBehalfOfEmail in sync when savedFlow is loaded asynchronously
 	$effect(() => {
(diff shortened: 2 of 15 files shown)


rubenfiszel:
✅ Review round clean @ 8567f255131fd7c465c9b0f53868d98776c899d7

PR #10506, opened 2026-08-04
feat: signpost the dbt migration path on the pipelines page

rubenfiszel:
> [!IMPORTANT]
> **Merge prerequisite:** the dbt docs page is not published yet — `https://www.windmill.dev/docs/getting_started/scripts_quickstart/dbt` returns 404. That is the canonical URL (already shipped on `main` by `workspaceSettings/DbtSettings.svelte:90`) and is kept deliberately so it self-heals on publish, but this PR puts it on a prominent button. Publish the page before merging.

## Summary

A dbt user landing on `/pipeline` had no signal that Windmill runs their existing project as-is: the page described native pipeline steps only, so the implied cost of adopting Windmill here was "rewrite the project as a pipeline". This adds an info alert stating that dbt is separately supported, with the links that make it actionable.

The alert is deliberately framed so dbt does **not** read as a flavour of data pipelines: a pipeline is Windmill's own abstraction, a dbt project stays a dbt project (dbt keeps owning its models, refs and tests), and the only overlap claimed is graph *visibility*. That claim is bounded on purpose — nothing native can write a `dbt://` relation (`// materialize` takes DuckLake targets only) and `AssetKind::Dbt` is excluded from `is_auto_trigger_kind`, so a `dbt://` read wires no cascade. It sits below the pipeline content for the same reason, so it isn't parsed as part of the definition of a pipeline.

## Changes

- New `PipelineDbtSignpost.svelte` under `frontend/src/lib/components/assets/AssetGraph/`: an info `Alert` titled "Prefer using dbt? Windmill also runs dbt".
- Three links out of it: **New dbt script** (`/scripts/add?lang=dbt`, which seeds the full dbt project scaffold), **dbt documentation** (external), and an inline link to `Settings → dbt` where the warehouse is configured.
- Rendered on `/pipeline` after the existing setup checklist and folder list, and hidden for operators — every affordance in it (create a script, configure a warehouse) is one an operator cannot act on, matching how the sibling `PipelineSetupSignpost` and the folder-create section are already gated.

## Screenshots

dbt-alert-light

dbt-alert-dark

(dark shot predates the final copy tweaks; layout and placement are unchanged)

## Test plan

- [x] `npm run check` passes with 0 errors (warning count unchanged from main)
- [x] `/pipeline` renders the alert in both light and dark themes
- [x] **New dbt script** lands on a fresh dbt draft with the `wm_dbt.yaml` / `dbt_project.yml` / `models/example.sql` scaffold
- [x] `Settings → dbt` resolves to `/workspace_settings?tab=dbt`
- [x] **dbt documentation** button resolves to the canonical docs URL with `target="_blank"` / `rel="noreferrer"` — the URL itself 404s until the docs page ships (see the note above)
- [ ] As an operator, the whole alert is hidden (gate mirrors the two existing ones on the page; not exercised with an operator account)

### DIFF
--- a/frontend/src/lib/components/assets/AssetGraph/PipelineDbtSignpost.svelte
+++ b/frontend/src/lib/components/assets/AssetGraph/PipelineDbtSignpost.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 	import { base } from '$lib/base'
-	import { userStore } from '$lib/stores'
 	import Alert from '$lib/components/common/alert/Alert.svelte'
 	import Button from '$lib/components/common/button/Button.svelte'
 	import DbtIcon from '$lib/components/icons/DbtIcon.svelte'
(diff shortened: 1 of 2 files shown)


rubenfiszel:
Re Codex's P2 (broken dbt documentation link) — confirmed and deliberately kept as-is.

\`https://www.windmill.dev/docs/getting_started/scripts_quickstart/dbt\` does 404 today; the page hasn't been published yet. Keeping the canonical URL rather than repointing, because:

- It is already the URL shipped on \`main\` by \`frontend/src/lib/components/workspaceSettings/DbtSettings.svelte:90\`, so swapping only this one would leave the product pointing two different places at dbt docs.
- It self-heals the moment the page publishes; a substitute would have to be found and reverted.
- There is no correct alternative today: the only live dbt URL on the site is \`/compare/dbt\` (Windmill-vs-dbt positioning, not documentation), and the parent quickstart index says nothing about dbt.

**Publishing that docs page is a prerequisite for merging this PR** — the alert puts the link on a prominent button, which is a worse 404 than the existing tooltip link in workspace settings.

rubenfiszel:
✅ Review round clean @ e3db64ad361892f05ffd428a19ca8b24353ca67c — Claude and Pi "Good to merge", Codex "Mergeable, but should ideally address nits"; the single nit (unpublished dbt docs page) is dismissed in the reply above and tracked as a merge prerequisite in the PR body. Head unchanged since the round.

PR #10507, opened 2026-08-04
feat: send prompt_cache_key on the openai responses api

rubenfiszel:
## Summary

Windmill never sent `prompt_cache_key` when calling `/v1/responses`, so prompt caching was unreliable on `gpt-5.6` and later. From those models on, the prefix hash alone no longer matches a cached prefix reliably: OpenAI states you "must set `prompt_cache_key` to use the more reliable matching for both implicit and explicit caching", and the key is combined with the prefix hash to route the request.

This is a cost regression rather than only a missed discount. On `gpt-5.6` and later, cache **writes** are billed on top of discounted reads, so every miss pays to re-write the prefix.

Both surfaces that speak the Responses API now send a key: the AI agent step (backend) and the copilot chat (frontend).

## Changes

- `ResponsesApiRequest` carries `prompt_cache_key`, fed from a new `BuildRequestArgs::prompt_cache_key`.
- Agent steps key on the step, not the run: `{workspace}:{runnable_path}:{flow_step_id}`. Every run of a step opens with the same system prompt and tool definitions, and each agent-loop iteration extends the previous one's prefix, so they all belong on one cache.
- Copilot chat keys on `{workspace}:{provider}:{model}:chat`.
- **Length bound.** OpenAI rejects a key over 64 characters (`Invalid 'prompt_cache_key': string too long`), and a runnable path alone can pass that. Both sides fold an over-long key into a digest of itself, keeping a readable head so the key stays traceable to its workspace. Without this, an ordinary path such as `f/internal_tools/customer_support_triage_agent` would 400 on the first call of every run and silently get *no* caching at all.
- **Retry, not a config knob.** If an OpenAI-compatible gateway refuses the field, the request is retried once without it and the refusal is remembered. Backend: a new branch in the existing `stream_options` retry ladder, checked *before* the chat/completions route fallback, since the endpoint serves the route and refuses one optional argument. Frontend: a third branch in the chat loop's existing "disable whichever optional feature the provider rejected" loop, cached per workspace+provider like the reasoning-summary probe.
- The key is deliberately **not** sent from `getOpenAIResponsesCompletionStream` / `getNonStreamingOpenAIResponsesCompletion`. Those are async-generator and one-shot paths whose callers cannot catch a rejection (invoking a generator runs none of its body), so a refused key there would be an unrecoverable failure. They are one-shot generations with no repeated prefix to route; the chat loop is the surface that has one, and it can retry.

## Verification

Exercised against a mock provider that records request bodies and rejects the field once, driving the real code paths end to end rather than only unit tests.

Backend, via an actual AI agent flow step:
```
req 1: prompt_cache_key='admins:u/admin/agent_retry_test/a:a'
req 2: prompt_cache_key=None          <- retried after the 400, job completed
```

With a path that exceeds the cap (87 chars raw), the key is bounded to exactly 64 and the job still completes:
```
prompt_cache_key='admins:f/internal_tools/custome:c0a367f1e3b26048346375bb9f57091e'
```

Frontend, driving the real chat loop in Chromium through the AI proxy:
```
req 1: /v1/responses  prompt_cache_key='admins:openai:gpt-5.6:chat'
req 2: /v1/responses  prompt_cache_key=None          <- retried, turn completed
```

## Screenshots

None — this change has no visible UI effect. It adds a field to an outbound request body and a retry around it; the chat renders identically. The browser-level evidence is the captured request bodies above.

## Test plan

- [ ] `cargo test -p windmill-ai --lib` (106 pass) and `cargo test -p windmill-worker --lib --features quickjs prompt_cache` (3 pass)
- [ ] `npx vitest run src/lib/components/copilot/chat/openai-responses.test.ts src/lib/components/copilot/chat/chatLoop.test.ts` (34 pass)
- [ ] `npm run check` — 0 errors
- [ ] Run an AI agent step against a real OpenAI `gpt-5.6` resource and confirm `cached_tokens` is non-zero on the second and later loop iterations
- [ ] Run one whose runnable path pushes the raw key past 64 chars and confirm no `Retrying request without prompt_cache_key` line appears in the worker log

## Follow-ups (deliberately not in this PR)

**`prompt_cache_retention` — worth doing separately, narrower than it looks.** OpenAI changed the default on 2026-05-29: without ZDR it is already `24h`, so setting it there is a no-op. Azure defaults `gpt-5.4` and older to `in_memory` (cleared after 5-10 min idle) and everything newer to `24h`. The population that gains anything is therefore **Azure OpenAI on `gpt-5.4` or older** — where the gain is real, since an agent step on an hourly schedule misses a 5-10 min cache every time.

It is not folded in here because it needs per-model, per-platform gating (newer Azure models reject `in_memory` outright) and because extended retention offloads KV tensors to GPU-local storage, which moves where data is retained. That is a compliance-visible change and belongs behind a deliberate setting, not a silent default in a PR about a routing key.

**`/chat/completions` — not worth doing, dropping it.** That surface is served by `OtherQueryBuilder` for mistral, deepseek, groq, togetherai, customai and non-Anthropic azure_foundry, none of which implement OpenAI's prompt cache; sending them an OpenAI-specific field risks 400s for no gain. The only OpenAI-family path there is the Azure fallback for deployments outside the Responses matrix — i.e. older models, which are not the ones that need the key. The intersection of "OpenAI-family, on chat/completions, and needs the key" is effectively empty.

---
## Summary by cubic
Send `prompt_cache_key` on OpenAI `/v1/responses` to restore reliable prompt caching on `gpt-5.6+` and cut costs by avoiding repeated cache writes. Adds stable, bounded keys for agent steps (backend) and Copilot chat (frontend), with safe fallbacks when gateways reject the field.

- **New Features**
  - Backend agent steps send a key derived from `{workspace}:{runnable_path}:{flow_step_id}` via `ResponsesApiRequest.prompt_cache_key`.
  - Copilot chat sends `{workspace}:{provider}:{model}:chat` for all turns via `getOpenAIResponsesCompletion`.
  - Streaming/one-shot generation paths do not send the key to avoid unrecoverable failures.

- **Bug Fixes**
  - Bound keys to 64 chars (OpenAI limit) with a digest suffix; truncation is char-safe to prevent 400s and lost caching.
  - If a gateway rejects `prompt_cache_key` (400), retry once without it and remember per `workspace+provider` (frontend) or step (backend) before any route fallback.

<sup>Written for commit 6e36e2be3926e8d40795d813bc87ec9d15596e98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -866,6 +866,7 @@ mod tests {
             user_message: "hello",
             attachments: None,
             has_websearch: false,
+            prompt_cache_key: None,
         };
 
         AnthropicQueryBuilder::new(AIProvider::Anthropic, AIPlatform::Standard)
--- a/backend/windmill-ai/src/providers/openai.rs
+++ b/backend/windmill-ai/src/providers/openai.rs
@@ -218,6 +218,12 @@ pub struct ResponsesApiRequest<'a> {
     pub max_output_tokens: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub text: Option<ResponsesApiTextFormat>,
+    /// From `gpt-5.6` on, the prefix hash alone no longer reliably matches a cached
+    /// prefix: the key is combined with it to route the request. Omitting it costs the
+    /// read discount and, since these models bill cache writes, pays to re-write the
+    /// prefix on every miss.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub prompt_cache_key: Option<&'a str>,
 }
 
 #[derive(Serialize, Debug)]
@@ -434,6 +440,7 @@ impl OpenAIQueryBuilder {
                 .map(|effort| ResponsesApiReasoning { effort: effort.to_string() }),
             max_output_tokens: args.max_tokens,
             text,
+            prompt_cache_key: args.prompt_cache_key,
         };
 
         serde_json::to_string(&request)
@@ -490,6 +497,8 @@ impl OpenAIQueryBuilder {
             reasoning: None, // Image generation models don't take a reasoning effort
             max_output_tokens: args.max_tokens,
             text: None, // No structured output for image generation
+            // A one-shot image prompt has no reusable prefix to route to a cache
+            prompt_cache_key: None,
         };
 
         serde_json::to_string(&request)
@@ -609,6 +618,7 @@ mod tests {
     use crate::query_builder::QueryBuilder;
 
     const SYSTEM_PROMPT: &str = "You are a helpful assistant";
+    const PROMPT_CACHE_KEY: &str = "test-workspace:f/agent:step_1";
 
     fn client() -> AuthedClient {
         AuthedClient::new(
@@ -641,6 +651,7 @@ mod tests {
             user_message: "hello",
             attachments: None,
             has_websearch: false,
+            prompt_cache_key: Some(PROMPT_CACHE_KEY),
         };
 
         OpenAIQueryBuilder::new(AIProvider::OpenAI)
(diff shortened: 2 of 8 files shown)


rubenfiszel:
✅ Review round clean @ 6e36e2be3926e8d40795d813bc87ec9d15596e98 — nit-only verdicts at 39491427825f6f4c436b2b044a7d30faeaf1c9a6 (codex: Good to merge, pi: Good to merge, claude: nit only); nit addressed in 6e36e2be3926e8d40795d813bc87ec9d15596e98

PR #10508, opened 2026-08-04
fix: log the db auth mode used and hint at the ms_entraid sentinel

rubenfiszel:
## Summary

Azure workload identity for `mssql` / `postgres` resources (#10470, released in v1.778.0) is selected by setting the resource password to the sentinel `ms_entraid`. When that string does not match, the code silently falls through to plain SQL Server authentication and sends the resource's `user` as a login name.

A report of "Entra ID auth does not work" turned out to be exactly that: the resource had the managed identity's client ID in `user` and something else in `password`, so the connection authenticated as SQL user `ee8f5073-…` and Azure rejected it with `18456`. Nothing in the job log said which auth mode had been chosen, so establishing that the Entra branch had never run took a full session of reading Azure SQL audit rows and hand-checking tiberius' LOGIN7 encoding against MS-TDS.

This does not change how workload identity authenticates. It makes the choice visible, and makes the near-miss recoverable.

**This does not fix that customer's connection** — their resource config was wrong and only they can fix it. What it fixes is that the next occurrence explains itself instead of costing a support round-trip. The underlying discoverability problem remains: `ms_entraid` is documented nowhere outside two Rust files, and that fix belongs in the docs repo.

## Changes

- `mssql_executor.rs`: every auth branch now names the mode it selected. Previously only `integrated_auth` and the `ms_entraid` branch logged; the `aad_token` and user/password branches were silent, so the absence of a line was ambiguous between "fell through to SQL auth" and "this code never logs here". The SQL-auth line includes the user, which is the field that was wrong in the report.
- `mssql_executor.rs`: new `login_error` appends a hint naming the sentinel when a login fails with `18456`, the connection used SQL auth, **and** the worker actually has Azure workload identity env vars. Gated on that last condition so the advice only appears where it is actionable. Replaces `to_anyhow(e).into()` on the two `Client::connect` error paths, which deliberately drops the `@file:line:col` suffix from those errors.
- `pg_executor.rs`: `PgAuthMode::log_name()` announces every non-default mode. Password auth stays silent, so ordinary Postgres jobs gain no log line. **Visible change for existing EE users: IAM RDS Postgres jobs now log `Using IAM RDS authentication` where they previously logged nothing.**
- Both paths compare the sentinel **trimmed**. A pasted value keeping its trailing newline previously went to the server as a literal password and failed as an ordinary bad login.
- The trim rationale lives on the `WORKLOAD_IDENTITY_PASSWORD` const, one place, rather than repeated at both call sites.

## Test plan

- [x] `cargo check -p windmill-worker --features mssql,enterprise` clean. Note `mod mssql_executor` is gated `#[cfg(all(feature = "enterprise", feature = "mssql"))]`, so `--features mssql` alone does **not** compile that file; only the combined set covers it. The CE build was checked for `pg_executor.rs`, which is not EE-gated.
- [x] `azure_workload_identity` unit tests pass, and `test_workload_identity_password_selects_the_auth_mode` was extended to pin the trim (`%20ms_entraid%0A`, which `parse_uri` percent-decodes to `" ms_entraid\n"`)
- [x] Real Postgres job, password auth: returns `[{"ok": 1}]`, no log line added
- [x] Real Postgres job with password `"  ms_entraid\n"`: selects workload identity and logs `Using Azure Workload Identity`, instead of sending the padded value as a password
- [ ] **Not exercised: the MSSQL path end-to-end.** No Azure SQL instance reachable from dev and `mssql` is not compiled into the dev backend. The MSSQL edits are the same trim logic proven on the Postgres path, two `append_logs` calls matching the two already in that function, and an error-formatting helper. Everything touched on that path is an error path or a log line except the trim, so it cannot break a connection that currently succeeds — but it is untested against a real server.

No frontend changes, so no screenshots. No `*_ee.rs` files modified.

### DIFF
--- a/backend/windmill-common/src/azure_workload_identity.rs
+++ b/backend/windmill-common/src/azure_workload_identity.rs
@@ -27,6 +27,10 @@ pub const AZURE_OSSRDBMS_SCOPE: &str = "https://ossrdbms-aad.database.windows.ne
 /// worker's identity instead of failing to log in, which is why it is deliberately less
 /// password-shaped than the instance's `entraid`: that one is set by whoever runs the
 /// instance, this one sits where users keep their own secrets.
+///
+/// Compare it trimmed. A near-miss is otherwise forwarded to the server as a real
+/// password, and the resulting rejection is indistinguishable from an ordinary bad
+/// login, so the mode having been intended at all leaves no trace.
 pub const WORKLOAD_IDENTITY_PASSWORD: &str = "ms_entraid";
 
 /// Renew an access token this long before it expires.
--- a/backend/windmill-worker/src/mssql_executor.rs
+++ b/backend/windmill-worker/src/mssql_executor.rs
@@ -64,6 +64,27 @@ lazy_static::lazy_static! {
     static ref RE_MSSQL_READONLY_INTENT: Regex = Regex::new(r#"(?mi)^-- ApplicationIntent=ReadOnly *(?:\r|\n|$)"#).unwrap();
 }
 
+/// A rejected login reports only the name it rejected, so a resource meant to reach the
+/// database as the worker's Azure identity, but whose password is not the sentinel, is
+/// indistinguishable from a wrong password. Point at the sentinel when the worker has an
+/// identity to offer, since that is the only case where the advice is actionable.
+fn login_error(e: tiberius::error::Error, sql_auth_user: Option<&str>) -> Error {
+    const LOGIN_FAILED: u32 = 18456;
+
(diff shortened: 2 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 40c215cd8985f78ae12fdaf37cfda0633f5ee910 — nit-only verdicts at 5780eecfa04f1ad54c9d08fec080dd5fbaa9e7e5 (codex: Good to merge, pi: Good to merge, claude: nits); nits addressed in 40c215cd89

PR #10510, opened 2026-08-04
feat: make createApp and updateApp the full-code app tools over MCP

rubenfiszel:
## Summary

Low-code apps are legacy, so `createApp` and `updateApp` — the app tools an agent gets over MCP — should be the full-code ones. Today they're the low-code ones, and an agent can't read an app at all.

The MCP app surface after this:

| tool | endpoint | |
|---|---|---|
| `listApps` | `/apps/list` | every app; `raw_app` tells the kinds apart |
| `getAppByPath` | `/apps/get/p/{path}` | the app's whole value, which a deploy needs |
| `createApp` | `/apps/create_raw_source` | **new endpoint** — create a full-code app from its sources |
| `updateApp` | `/apps/update_raw_source/{path}` | update one (#10500) |

The low-code `/apps/create` and `/apps/update/{path}` are no longer MCP tools. The endpoints are untouched — the editor, the CLI and git-sync still use them.

## How the names work

`createApp` and `updateApp` stay the operationIds of the low-code endpoints, because the frontend, the CLI and the published client are all keyed on them; renaming those would ripple through every caller for no gain. What changes is the name the *tool* carries: a new `x-mcp-tool-name` on the raw-source endpoints, which the generator uses in place of the operationId. So an agent sees `createApp`/`updateApp` and they are the full-code ones, while `AppService.updateApp` keeps meaning what it means.

## Changes

- **`POST /w/{workspace}/apps/create_raw_source`** — the create counterpart of `update_raw_source`: sources as JSON, compiled on a worker through `wmill app bundle`, then the app is created and the bundle stored. Same guards as the update path (non-operator, `apps:write:<path>`, `jobs:run` because the compile runs the app's own dependencies, deploy rules).
- **`listApps` and `getAppByPath` become tools.** Without a read tool the deploy tools aren't usable: `updateApp` takes the *whole* value, so an agent has to read it first — its own instructions said "read the current one with getAppByPath", which wasn't exposed. `listApps` carries no `value`, so listing stays cheap.
- **`x-mcp-tool-name`** in the generator, defaulting to the operationId.
- **The `jobs:run` gate covers both deploy routes** and matches on the names agents see. That gate is keyed on the tool name the token grants, so it had to follow the rename or `updateApp` would 403 for every token that named it.
- Each tool's instructions say which kind it is for: the read tools explain that `raw_app` distinguishes them and only a full-code app deploys through MCP; the write tools say low-code apps are legacy and have no MCP tool at all.

## Upgrade note — existing MCP tokens

The tool names are reused deliberately, and that has a consequence worth stating before merge: a token issued **before** this change whose scope names `mcp:endpoints:updateApp` (or `createApp`) selected the *low-code* tool, which couldn't run anything on a worker. After this it names the full-code deploy tool, so it gets `jobs:run` minted for that call and can have a worker compile app sources — i.e. run the app's own dependencies. Nobody re-consents; the picker's warning only reaches newly-issued tokens.

That was a deliberate call (MCP has one pair of app-write tools and these are them), not an oversight — the mint site says so, so nobody "fixes" it later. **Existing MCP tokens are worth reviewing when this ships.** If you'd rather old grants not carry over, the alternative is names an old token can't hold (`updateFullCodeApp`, or the `updateAppRawSource` this branch started with) — say so and it's a one-line change plus a regen.

## Test plan

Verified against the dev backend (`WM_RAW_APP_BUNDLER_CLI` pointing at the local CLI, since `app bundle` ships in the release that just merged and isn't on npm yet).

- [x] `POST /apps/create_raw_source` with a React app's sources → **201**, `raw_app = true`, 187,124-byte bundle stored — byte-identical to `wmill app bundle` on the same sources
- [x] a low-code value (`{grid: []}`) → 400 `no \`files\` to bundle`, before any job is queued (integration case, needs no worker)
- [x] a **non-admin member of `g/all`** creating at `g/all/…` reaches the compile, and gets 403 at another user's `u/…` path (integration case, needs no worker)
- [x] the same two outcomes straight from the `app` policies in psql: the group member's insert succeeds, the outsider's raises `42501`
- [x] generated tool list is exactly `listApps, getAppByPath, createApp, updateApp`, and `createApp`/`updateApp` point at the raw-source endpoints
- [x] the scope gate mints `jobs:run` for a token naming `createApp` or `updateApp`, and not for `mcp:all` — 6 scope tests pass under `--features mcp`
- [x] `cargo test -p windmill-api-integration-tests --test apps` — 2 passed; 6 scope tests; `cargo check -p windmill-api --features mcp`, `npm run check:fast` — clean

## Review findings addressed

- **create compiled before the checks that reject it** (claude, pi) — the path-already-taken and writer checks now run before the bundle job, matching what `update_raw_source` does and what its comment claims.
- **create refused `g/<group>/…` for a non-admin group member** (codex, claude) — that check was a hand-written restatement of the app RLS, and it missed the `see_member` policy that grants a group member write on their group's paths. It now probes the insert itself through `user_db` and rolls it back, the way the update path asks `can_write_app`, and a denial answers 403 like the update path rather than 400.
- **`listApps` couldn't page past the 1000-row default** (codex) — `page` and `per_page` are in the tool schema now.
- **`getAppByPath` can truncate a large app's value** (codex) — the 87,500-byte tool-result cap is global to MCP and out of scope to change here. The instructions now say what happens: send a truncated value back and the build fails rather than deploying something partial, so large apps belong in the editor or the CLI. A chunked read is the real fix and is a separate change.
- **Reusing the tool names upgrades existing tokens** (all three) — see the upgrade note above; kept deliberately.

Follow-up to WIN-2291 / #10500.

### DIFF
--- a/backend/.sqlx/query-b6f0ebbe674e1290afda7b5bed3b463f1d73d66b5924734274272509c98573db.json
+++ b/backend/.sqlx/query-b6f0ebbe674e1290afda7b5bed3b463f1d73d66b5924734274272509c98573db.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "INSERT INTO app (workspace_id, path, summary, policy, versions)\n         VALUES ($1, $2, '', '{}'::jsonb, '{}') RETURNING 1",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "?column?",
+        "type_info": "Int4"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Varchar",
+        "Varchar"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "b6f0ebbe674e1290afda7b5bed3b463f1d73d66b5924734274272509c98573db"
+}
--- a/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
+++ b/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
@@ -426,7 +426,11 @@ def find_mcp_tools(spec: Dict[str, Any]) -> List[Dict[str, Any]]:
             if isinstance(operation, dict) and operation.get('x-mcp-tool') is True:
                 # Extract tool information
                 tool = {
-                    'name': operation.get('operationId', f"{method}_{path.replace('/', '_').replace('{', '').replace('}', '')}"),
+                    # The name an agent sees. Defaults to the operationId, which
+                    # every generated client is keyed on, so an endpoint whose
+                    # tool should read differently overrides it here rather than
+                    # by renaming the operation.
+                    'name': operation.get('x-mcp-tool-name') or operation.get('operationId', f"{method}_{path.replace('/', '_').replace('{', '').replace('}', '')}"),
                     'description': build_tool_description(operation, method, path),
                     'instructions': operation.get('x-mcp-instructions', ''),
                     'path': path,
--- a/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
+++ b/backend/generate_mcp_endpoints_tools/generate_mcp_tools.py
@@ -443,7 +443,16 @@ def find_mcp_tools(spec: Dict[str, Any]) -> List[Dict[str, Any]]:
                     'include_query_params': operation.get('x-mcp-tool-include-query-params'),
                 }
                 tools.append(tool)
-    
+
+    # A tool name used to be an operationId, which OpenAPI already keeps unique.
+    # `x-mcp-tool-name` gives that up, and a duplicate would be silent: a token's
+    # `mcp:endpoints:<name>` resolves by first match, so which endpoint it reaches
+    # would depend on the order of this file.
+    names = [t['name'] for t in tools]
+    duplicates = sorted({n for n in names if names.count(n) > 1})
+    if duplicates:
+        raise ValueError(f"duplicate MCP tool name(s): {', '.join(duplicates)}")
+
     return tools
 
 def generate_typescript_code(tools: List[Dict[str, Any]], spec: Dict[str, Any], base_path: str = "") -> str:
--- a/backend/windmill-api-integration-tests/tests/apps.rs
+++ b/backend/windmill-api-integration-tests/tests/apps.rs
@@ -411,6 +411,24 @@ async fn test_raw_app_kind_is_not_flipped_by_update(db: Pool<Postgres>) -> anyho
         "expected the raw update of a low-code app to be refused"
     );
 
+    // Nor may the source endpoints compile a value that isn't a raw app's: both
+    // refuse up front, without queueing a bundle job no worker would pick up here.
+    let resp = authed(client().post(format!("{base}/create_raw_source")))
+        .json(&json!({
+            "path": "u/test-user/from_source",
+            "summary": "",
+            "value": { "grid": [] },
+            "policy": { "execution_mode": "publisher" }
+        }))
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(resp.status(), 400);
+    assert!(
+        resp.text().await?.contains("no `files` to bundle"),
+        "expected a low-code value to be refused before bundling"
+    );
+
     // The source endpoint refuses the same mismatch up front — it must not queue
     // a bundle job (which no worker would pick up here) to find that out.
     let resp = authed(client().post(format!("{base}/update_raw_source/{low_code_path}")))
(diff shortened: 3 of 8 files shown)


rubenfiszel:
Round 1 addressed in 6f4a8c3510.

- **claude + pi P1 — create compiled before the checks that reject it.** Right, and it contradicted the comment on the update path. The path-already-taken and writer checks now run before the bundle job, same ordering as `update_raw_source`.
- **codex P2 — `listApps` couldn't page past the 1000-row default.** My schema only exposed `path_start`; `page` and `per_page` are in it now.
- **codex P1 — `getAppByPath` can truncate a large app's value.** Real, and not fixable here: the 87,500-byte cap is global to every MCP tool result (`runner.rs:36`). Two things make it less bad than it reads — a truncated value sent back fails the build rather than deploying something partial, and the instructions now say so and point large apps at the editor or the CLI. A chunked/file-level read is the actual fix and belongs in its own change.
- **All three P0 — reusing the tool names upgrades tokens issued before this.** Confirmed and **kept deliberately**: a pre-existing token naming `mcp:endpoints:updateApp` now reaches the full-code tool and gets `jobs:run` minted for that call. MCP has one pair of app-write tools and these are them; the alternative is names an old token can't hold, at the cost of the plain `createApp`/`updateApp` an agent should see. The mint site records the decision so it doesn't read as an oversight later, and the PR body carries an upgrade note telling admins to review existing MCP tokens when this ships. Happy to switch to distinct names if you'd rather — one line plus a regen.

6 scope tests and 2 integration tests pass; `cargo check -p windmill-api --features mcp` and `npm run check:fast` clean.

rubenfiszel:
Round 2 addressed in 021b124eda.

**`createApp` refused `g/<group>/…` for a non-admin group member** (codex P0, claude) — real, and self-inflicted: I used the hand-written `require_is_writer` for the pre-compile check, which is the exact restatement of the app RLS that #10500 taught me to drop for the update path. The `app` table has a `see_member` policy that is `FOR ALL` on `g/<group>/…` when the group is in the session, so a group member may create there — `require_is_writer` doesn't know that.

The create path now asks the database the same way `can_write_app` does: a rolled-back `INSERT … RETURNING 1` through `user_db`. `app` has no insert trigger (its only two are `AFTER UPDATE OF policy` and `AFTER DELETE`), so the probe leaves nothing behind but a consumed sequence value. `42501` means refused; `23505` means the path was taken between the existence check and the probe, which isn't a permission problem, so the real insert reports the collision.

**Write denial answered 400 where the update path answers 403** (claude) — fixed, it's `PermissionDenied` now.

Both directions are pinned by a new worker-free integration case: the pre-compile check runs before the bundle, so an empty `files` says which side of it a request reached. As `test-user-2` (non-admin, member of `all`), `g/all/…` reaches the compile and `u/test-user/…` gets 403. Confirmed independently against the policies in psql — the group member's insert succeeds, the outsider's raises `42501`.

Checks: 3 app integration tests, 17 `windmill-api` scope/token tests under `--features mcp`, `cargo check -p windmill-api --features mcp`, `npm run check:fast` — all clean.

The three round-2 findings that stand unchanged are in the PR body: the token upgrade is the deliberate call you signed off on, `getAppByPath` truncation is the global 87.5KB MCP cap (documented in the tool instructions; a chunked read is a separate change), and `listApps` pagination was fixed in the previous round.

rubenfiszel:
Round 3 (local, fresh context) addressed in 27df269941. Codex and Pi both came back **good to merge** on 021b124eda; the local pass found three P2 nits, all fixed:

- **`x-mcp-tool-name` dropped the uniqueness the operationId gave for free.** A duplicate would have been silent: `mcp:endpoints:<name>` resolves by first match, so which endpoint a token reached would depend on the order of the yaml. The generator now raises on a duplicate. Verified by re-marking `/apps/create` as a tool: `ValueError: duplicate MCP tool name(s): createApp`.
- **The create body omitted `preserve_on_behalf_of` and `skip_draft_deletion`**, which `CreateApp` deserializes and `create_app_internal` honors. Documented, matching `/apps/create` and `update_raw_source`. They stay out of the MCP tool schema (`x-mcp-tool-include-fields` bounds it) — this is the spec catching up to the handler, not a widening of what an agent can send.
- **`listApps` told the agent to read `raw_app`, but `ListableApp.raw_app` is `skip_serializing_if = "is_false"`** — a low-code app has no such key, so an agent filtering on `raw_app === false` would classify nothing. The instruction now says the field is omitted rather than sent as false.

**One finding I did not act on**, because it doesn't reproduce. The review flagged `can_write_app` (from #10500) as answering 500 rather than 403 when RLS refuses, on the theory that an `UPDATE` passing `USING` but failing `WITH CHECK` raises `42501`. Checked against the live policies: for a row readable through a folder *read* grant, `see_folder_extra_perms_user_update` keys on `folders_write`, so the row is filtered out of the UPDATE rather than raising — `UPDATE 0`, which `can_write_app` already turns into 403. The `42501` arm is needed on the create side (an INSERT's `WITH CHECK` does raise, which is why it's there) and not on the update side.

Checks: `cargo check -p windmill-api --features mcp`, `npm run check:fast`, generated tool list still exactly `listApps, getAppByPath, createApp, updateApp` — all clean.

PR #10512, opened 2026-08-04
fix: let admins edit the dev workspace lock ruleset

rubenfiszel:
## Summary

A prod workspace paired with a dev workspace carries a reserved `dev_workspace_lock` protection rule. There was no way to change it: the rulesets tab listed it and offered Edit/Delete, but the API rejected both with

> Bad request: 'dev_workspace_lock' is a reserved protection-rule name managed by the dev workspace feature

and the dev-workspace panel's "Manage in Rulesets" button pointed at exactly that dead end. The only remaining lever was Detach, which drops the whole lock rather than one restriction.

The reservation exists because the pairing does name-based cleanup (attach creates the rule, detach deletes it), so a user rule sharing the name would get clobbered. That justifies reserving creation and deletion, but not updates.

Renaming was separately broken for **every** rule: the handler took the name from the path and ignored the body, so the editor reported success while the row kept its old name. That is fixed here too, with the reserved rule excluded since the feature locates it by name.

## Changes

**Backend**
- `update_protection_rule` no longer calls `reject_reserved_rule_name`. Creation and deletion stay reserved, so detach remains the way a pairing's lock is removed.
- `UpdateProtectionRuleRequest` gains an optional `name`, and the handler applies it: collisions are refused, `dev_workspace_lock` can be neither renamed nor renamed onto, and the audit entry records the previous name. Omitting the field means "leave the name alone", so older clients are unaffected.

**Frontend**
- The Rule Name field is disabled for `dev_workspace_lock` with an explanation, and the editor now sends the name so a rename applies.
- The rulesets tab hides Delete on the reserved rule and labels the row with where it comes from.
- `?rule=<name>` deep-links straight into a rule's drawer. The param is consumed on open so it does not re-trigger on later saves or tab switches.
- The dev-workspace panel names the rulesets enforcing each lock and links into them, instead of only reporting "blocked / allowed".
- That panel now also renders inside the dev workspace, showing the parent's protections. Its buttons name the parent (`Edit in <parent>`) because clicking them switches workspace, and they only appear for an admin **of the parent**, which is the tab that link leads to.

## Test plan

- [x] `cargo test --test protection_rules --features enterprise,private` passes 4/4, including two new tests: `test_dev_workspace_lock_rule_reservation` (create rejected, update accepted, rename rejected, delete rejected) and `test_protection_rule_rename` (rename applies and moves the row, collision and reserved-name targets refused)
- [x] `npm run check` clean
- [x] Prod side: dev-workspace tab names `dev_workspace_lock`, Edit opens its drawer directly
- [x] Rulesets tab: reserved row offers Edit only, no Delete
- [x] Reserved rule's name field is disabled; a normal rule's stays editable
- [x] Saving a renamed rule sends `name` in the request body
- [x] Dev side: panel shows the parent's protections, labels name the parent, Edit switches there with the drawer open
- [x] Dev side with no rules on the parent: "Manage in <parent>" switches to the parent (Codex finding, fixed in 3fd6de8e5b)

The rename round-trip is verified at the API level by the integration test. The browser check confirmed the editor sends `name`; it ran against a dev backend built before this change, so it could not observe the resulting rename.

## Screenshots

Dev-workspace tab on the prod side, naming the enforcing ruleset:

dev workspace panel

Edit opens the rule's drawer directly. The name is locked with an explanation; the restrictions are editable and now save:

reserved name locked

Rulesets row: Edit only, plus the provenance hint:

reserved row menu

Same panel seen from inside the dev workspace, reporting the parent's rules:

dev side panel

### DIFF
--- a/backend/.sqlx/query-adbe6ce27d46ba0d2d3ce87d875f60c179e5bccc54a36c8d5dcf04fd84445309.json
+++ b/backend/.sqlx/query-adbe6ce27d46ba0d2d3ce87d875f60c179e5bccc54a36c8d5dcf04fd84445309.json
@@ -1,0 +1,19 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n            UPDATE workspace_protection_rule\n            SET name = $1, rules = $2, bypass_groups = $3, bypass_users = $4\n            WHERE workspace_id = $5 AND name = $6\n        ",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Varchar",
+        "Int4",
+        "TextArray",
+        "TextArray",
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "adbe6ce27d46ba0d2d3ce87d875f60c179e5bccc54a36c8d5dcf04fd84445309"
+}
--- a/backend/tests/protection_rules.rs
+++ b/backend/tests/protection_rules.rs
@@ -421,3 +421,88 @@ async fn test_restrict_deploy_to_deployers(db: Pool<Postgres>) -> anyhow::Result
 
     Ok(())
 }
+
+/// The dev-workspace pairing owns `dev_workspace_lock` by name: attaching creates it, detaching
+/// deletes it. Only those two ends of the name are reserved. Updating the rule has to stay open,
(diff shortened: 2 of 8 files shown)


rubenfiszel:
✅ Review round clean @ bec20580db42fec1dcc99a9adc582c679291735c

PR #10514, opened 2026-08-04
fix: show which MCP endpoint tools a token scope will actually expose

rubenfiszel:
## Summary

The MCP token scope picker offered endpoint tools that the server then refused to list, with nothing in the UI explaining why. Two gaps, both reported from the field:

- **`runScriptByPath` / `runFlowByPath` are not governed by `mcp:endpoints:` at all.** The runner gates them on whether the token grants *any* script/flow path (`has_any`), so a token scoped `mcp:endpoints:getScriptByPath,runScriptByPath,listScripts` never gets `runScriptByPath` — while a token that happens to carry `mcp:flows:` gets `runFlowByPath` it never asked for. The picker listed both in the "API Endpoints" multiselect, so selecting one looked like it should work.
- **`runScriptPreviewAndWaitResult` / `deleteScriptByHash` are withheld from path-confined tokens.** They take no checkable path, so `endpoint_tool_in_scope` hides them once the token is limited to specific script paths. The picker let you select them anyway, and the Folders-mode preview listed them as available.

The gating is deliberate (an endpoint scope must not authorize running an arbitrary path), so this PR fixes the UI rather than the policy: the picker now mirrors the runner's listing rules and shows exactly what the token will expose.

## Changes

- New `frontend/src/lib/components/mcp/endpointScopePolicy.ts` — a TypeScript mirror of `endpoint_path_policy`, `endpoint_tool_in_scope`, `path_confined` (`backend/windmill-mcp/src/server/runner.rs`) and `parse_mcp_scopes` / `is_resource_allowed` / `has_any` (`backend/windmill-mcp/src/common/scope.rs`).
- `McpScopeSelector.svelte`:
  - The "API Endpoints" list no longer offers the run-by-path tools, since selecting them there does nothing. They get a dedicated **Run tools** section that reports `exposed` / `not exposed — select at least one script above, or add a script wildcard pattern`, driven live by the Scripts/Flows selections.
  - A warning when a selected endpoint will still be withheld because the token is confined to specific script paths.
  - The Favorites / All / Folders preview now lists only the endpoint tools that scope will actually expose (previously it listed every tool unconditionally — including run-by-path under Favorites, and the unconfinable pair under Folders).
  - The scope string became a `$derived` so the previews recompute in the same pass as the selection that changed them.
- Comment on the Rust `EndpointPathPolicy` enum pointing at the TS mirror, so the two stay in sync.

## Picked vs exposed

Every row below was walked through in the running app and read back off the rendered picker (39 endpoint tools total in this build).

| # | Scope | Picked | `runScriptByPath` | `runFlowByPath` | Unconfinable pair | Endpoint tools listed |
|---|---|---|---|---|---|---|
| 1 | Favorites only | — | not exposed | not exposed | exposed | 37 |
| 2 | All scripts/flows | — | exposed | exposed | exposed | 39 |
| 3 | Folders (`app_custom`) | — | exposed | exposed | **withheld** (path-confined) | 37 |
| 4 | Custom | nothing | not exposed | not exposed | n/a | — |
| 5 | Custom | script pattern `u/admin/*` | **exposed** | not exposed | n/a | — |
| 6 | Custom | script + flow patterns | exposed | exposed | n/a | — |
| 7 | Custom | flow pattern only | not exposed | **exposed** | n/a | — |
| 8 | Custom | script pattern + `runScriptPreviewAndWaitResult` | exposed | exposed | **withheld** + warning | — |
| 9 | Custom | same, pattern widened to `*` | exposed | exposed | exposed, warning clears | — |
| 10 | Custom + **read-only** | script pattern + `runScriptPreviewAndWaitResult` | section hidden | section hidden | pruned from selection | GET only |

Rows 1–3 are the preview-badge modes; rows 4–10 are the custom picker. Row 3 is the case the old UI got wrong in both directions at once: it advertised the unconfinable pair a folder token can never call. Row 10 drops the whole Run tools section because both run tools are POST.

## Screenshots

**Custom, nothing picked** — both run tools state what would enable them, instead of sitting silently in the endpoint list:

mcp-custom-empty

**Custom, `u/admin/*` script pattern + `runScriptPreviewAndWaitResult` picked** (rows 5 and 8) — `runScriptByPath` flips to exposed, `runFlowByPath` stays off, and the withheld endpoint is called out with the remedy that satisfies both gates:

mcp-custom-scope

**Favorites preview** (row 1) — 37 badges, no `runScriptByPath` / `runFlowByPath`. Before this PR all 39 were listed, including two the server would never hand out:

mcp-favorites-preview

## Test plan

- [ ] `frontend/src/lib/components/mcp/endpointScopePolicy.test.ts` pins the mirror against the two confusing cases (run-by-path gated on the script/flow scope, unconfinable tools withheld from path-confined tokens).
- [ ] Open Account settings → Generate MCP URL → **Custom**: with nothing selected both run tools read "not exposed"; typing `u/admin/*` into *Script wildcard patterns* flips `runScriptByPath` to "exposed" and leaves `runFlowByPath` alone; typing a flow pattern flips `runFlowByPath`.
- [ ] Select `runScriptPreviewAndWaitResult` while a script pattern is set — the withheld-endpoints warning appears.
- [ ] Switch to **Favorites only** — the endpoint preview no longer lists `runScriptByPath` / `runFlowByPath`; switch to **Folders** — it no longer lists `runScriptPreviewAndWaitResult` / `deleteScriptByHash`.
- [ ] Edit an existing MCP token whose scope contains `mcp:endpoints:runScriptByPath` — the no-op name is dropped from the endpoint list and the Run tools section explains the real gate.
- [ ] Regression guard for the endpoint-selection prune: a `*` endpoint scope must survive, since dropping it would emit a scope with no `mcp:endpoints:` part and silently revoke every endpoint tool. Verified by opening a `mcp:scripts:* mcp:flows:* mcp:endpoints:*` token in the edit modal and saving untouched — the scope round-trips intact.

### DIFF
--- a/backend/windmill-mcp/src/server/runner.rs
+++ b/backend/windmill-mcp/src/server/runner.rs
@@ -127,7 +127,9 @@ impl<B: McpBackend> Runner<B> {
 }
 
 /// How an endpoint tool interacts with the token's `mcp:scripts:` / `mcp:flows:`
-/// path scopes.
+/// path scopes. Mirrored in
+/// `frontend/src/lib/components/mcp/endpointScopePolicy.ts` so the scope picker
+/// only advertises tools this policy will actually list — keep the two in sync.
 enum EndpointPathPolicy {
     /// Executes the script/flow named by the `path` argument. Gated by the
     /// script/flow scope alone (an endpoint scope is not enough to run things):
--- a/frontend/src/lib/components/mcp/McpScopeSelector.svelte
+++ b/frontend/src/lib/components/mcp/McpScopeSelector.svelte
@@ -461,8 +461,6 @@
 	let exposedScriptCount = $state<number | undefined>(undefined)
 	let exposedFlowCount = $state<number | undefined>(undefined)
 
-	// Mirror the backend's is_resource_allowed: `*`, an exact path, or an `x/*`
-	// subtree (matching the folder itself or anything beneath it).
 	// Custom-mode counts are computed synchronously from the already-loaded
 	// allScripts/allFlows plus the current selections/patterns — no extra fetch.
 	$effect(() => {
@@ -670,8 +668,8 @@
 					<Alert type="warning" size="xs" title="Some selected endpoints will not be exposed">
 						<span class="text-xs">
 							Endpoints that cannot be limited to specific script paths are withheld from a token
-							restricted to a script selection: {withheldEndpoints.join(', ')}. Clear the Scripts
-							selection and patterns to expose them.
+							restricted to a script selection: {withheldEndpoints.join(', ')}. Widening the script
+							wildcard pattern to <code>*</code> exposes them, and keeps the run tools below available.
 						</span>
 					</Alert>
 				{/if}
--- a/frontend/src/lib/components/mcp/McpScopeSelector.svelte
+++ b/frontend/src/lib/components/mcp/McpScopeSelector.svelte
@@ -11,7 +11,9 @@
 	import {
 		endpointPathPolicy,
 		isEndpointExposed,
-		parseMcpScopeState
+		matchesAnyPattern,
+		parseMcpScopeState,
+		pruneEndpointSelection
 	} from '$lib/components/mcp/endpointScopePolicy'
 	import InfoIcon from 'lucide-svelte/icons/info'
 	import { SvelteMap } from 'svelte/reactivity'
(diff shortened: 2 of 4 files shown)


rubenfiszel:
✅ Review round clean @ 5d61f8b375e26b4e113e9416c1a91ebbae7d7d56 — nit-only verdicts at ab1f45fc068244cec37b2d7e186a500171c95bd3 (codex: Good to merge, pi: Good to merge, claude: Mergeable, but should ideally address nits); nits addressed in 5d61f8b375e26b4e113e9416c1a91ebbae7d7d56

PR #10516, opened 2026-08-04
feat: let the database manager run its jobs on a custom worker tag

rubenfiszel:
## Summary

Fixes WIN-2315.

Every Database Manager operation — schema introspection, table metadata, row reads and
writes, DDL, the SQL REPL — runs as a Windmill preview job tagged with the database
language (`postgresql`, `mysql`, `duckdb`, …). That tag decides which worker group picks
the job up, and the default groups are not always the right ones: a database on a private
network is often reachable only from one group, and #10456 already surfaced the case where
no worker serves the native tag at all. Until now the drawer had no way to say "run these
on that group instead" — the only fix was to widen a worker group's tags instance-wide.

This adds a per-database worker tag override to the drawer, and offers it exactly where
the failure shows up.

## Changes

**Setting the tag**
- New `DbWorkerTagPicker` wraps the existing `WorkerTagSelect` (custom tags + live
  worker-availability dots), with the language's native tag as the placeholder so it is
  clear what the jobs run on without an override.
- Both database drawers (the Database Manager and the raw-app `Data` drawer) get a
  `Worker tag` popover in their action bar; the button label shows the active override.
- `useDbManagerTag` (`dbManagerTag.svelte.ts`) keeps the override in local storage under
  `dbManagerWorkerTag:<workspace>:<db path>`. A database that needs a specific group needs
  it on every visit, not just the one where it was set. Tags set this session are held in
  one module-level map rather than per caller: local storage is not reactive, so a
  per-instance cache would leave a second drawer open on the same database running its
  jobs on the tag the first one just replaced.

**Hinting at it when the default tag doesn't work**
- The picker is rendered under the existing missing-worker warning (shown after 10s of a
  still-queued job) and in the load-error panel. A database no default worker can *reach*
  fails as an error rather than as a missing tag, so both paths offer the override.
- `MissingWorkerTagAlert` now probes the tag actually in use, so picking a tag no worker
  serves is reported against that tag.

**Plumbing**
- `tag` threaded through `dbTableOpsWithPreviewScripts` (count/select/insert/update/delete),
  `dbSchemaOpsWithPreviewScripts` (DDL + foreign-key/primary-key lookups),
  `getDucklakeSchema`, `loadAllTablesMetaData` (+ the Snowflake primary-key follow-up),
  `getDbSchemas` (`customTag`, already supported) and `SqlRepl`. All optional — every
  other caller is unchanged and keeps the native tag.
- Changing the tag is a dependency of the schema/metadata resources, so the manager
  refetches on the new tag immediately. Each load is stamped so a superseded one can
  neither report its failure nor overwrite what replaced it: a job already queued cannot
  be recalled (`resource` aborts a controller the poller behind these queries doesn't
  watch), and one left on an unserved tag only fails ~90s later — which is precisely the
  window in which the user picks a working tag from the hint.
- Data table migrations are run by the server on its own tag and are deliberately left
  alone.

## Screenshots

Local instance whose worker group does not serve `postgresql`, on a `postgresql` resource.

Queued too long — the missing-worker warning, now with the override right under it:

slowload-hint

The poller gives up — same override on the error panel:

error-hint

Picking a custom tag (availability dots come from `WorkerTagSelect`):

tag-dropdown

Same database, now loading on a tag a worker actually serves:

loaded-with-custom-tag

The action-bar popover, reachable whether or not anything is failing:

actionbar-popover

## Test plan

Exercised end to end against a local instance whose only worker serves the default tags
(not `postgresql`), on a `postgresql` resource, with `duckdb` declared as an assignable
custom tag so a real worker would pick the jobs up:

- [x] With no override, the drawer shows the missing-worker warning and then the error
      panel, both offering the picker.
- [x] Selecting the custom tag refetches immediately: schema, table list and 622 rows load.
      Every job in `v2_job` for that window carries the override
      (`LOAD_TABLE_METADATA`, the introspection query, `COUNT`, `SELECT`).
- [x] SQL REPL run (`SELECT * FROM _sqlx_migrations`) queued on the override.
- [x] DDL through the manager (`New table` → `wm_tag_probe`) queued on the override and
      the table was created.
- [x] The override survives a full page reload (local storage) and the action-bar button
      shows it; clearing it removes the entry and returns the jobs to `postgresql`.
- [x] Picking a working tag from the slow-load hint while the native-tag job is still
      queued: the manager stays loaded 140s later, past that job's no-worker verdict
      (regression test for the staleness guard — it flipped back to the error panel
      before the fix).
- [x] `npm run check` — 0 errors (65 pre-existing warnings, none in the changed files);
      prettier clean.

No unit test ships with this: the one logic module is rune-based, and this repo has no
vitest project that runs runes — `*.svelte.test.ts` is excluded from both the `server` and
`dom` projects, so the single existing rune test never runs. Adding a project for it would
pull that unrelated dormant file into CI, which is out of scope here.

- [x] Raw-app `Data` drawer, against a data table: same missing-worker hint and error
      panel, the picker there loads the schema on the override, and it is stored under its
      own key (`…:datatable://main` alongside `…:u/admin/local_pg`, neither leaking into
      the other).

Not exercised in a browser: two drawers open on the *same* database at once. Session
writes now live in a single module-level `$state` map that every instance derives from,
so there is no per-instance cache left to go stale; each drawer's write→derive→refetch
path was verified on its own.

### DIFF
--- a/frontend/src/lib/components/DBManagerContent.svelte
+++ b/frontend/src/lib/components/DBManagerContent.svelte
@@ -9,6 +9,7 @@
 		getLanguageByResourceType
 	} from './apps/components/display/dbtable/utils'
 	import DbManager from './DBManager.svelte'
+	import DbWorkerTagPicker from './DbWorkerTagPicker.svelte'
 	import MissingWorkerTagAlert from './jobs/MissingWorkerTagAlert.svelte'
 	import {
 		dbSchemaOpsWithPreviewScripts,
@@ -49,6 +50,9 @@
 		 *  navigation `$workspaceStore`; pass the acting workspace when embedded in
 		 *  a session preview whose workspace differs from the top nav. */
 		workspace?: string
+		/** Worker tag every job of this manager runs on, overriding the database
+		 *  language's native tag. Bound so the hints below can offer to set it. */
+		workerTag?: string
 	}
 
 	let {
@@ -62,7 +66,8 @@
 		selectedTables = $bindable([]),
 		disabledTables = [],
 		onImport,
-		workspace = undefined
+		workspace = undefined,
+		workerTag = $bindable()
 	}: Props = $props()
 
 	let ws = $derived(workspace ?? $workspaceStore)
@@ -103,12 +108,12 @@
 	)
 
 	let colDefs = resource(
-		() => [input, ws],
+		() => [input, ws, workerTag],
 		async () => {
 			colDefsError = undefined
 			if (!input) return
 			try {
-				return await loadAllTablesMetaData(ws, input)
+				return await loadAllTablesMetaData(ws, input, workerTag)
 			} catch (e) {
 				colDefsError = 'Error loading tables metadata: ' + ((e as Error)?.message || e)
 				return
@@ -117,7 +122,7 @@
 	)
 
 	let dbSchemasPromise = resource(
-		() => [input, ws],
+		() => [input, ws, workerTag],
 		async () => {
 			schemaError = undefined
 			if (!input) return
@@ -127,7 +132,8 @@
 					input.resourceType,
 					input.resourcePath,
 					ws,
-					(message: string) => (schemaError = message)
+					(message: string) => (schemaError = message),
+					{ customTag: workerTag }
 				)
 				if (!schema) {
 					schemaError ??= 'The schema query returned no schema'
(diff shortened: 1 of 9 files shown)


rubenfiszel:
✅ Review round clean @ 2a1f382f2260dcf44e60564c3f9a5e9893d691bb

PR #10517, opened 2026-08-04
fix: keep the token scope builder inside its panel when scopes get long

rubenfiszel:
## Summary

In the token scope builder, restricting `jobs:run:scripts` to a handful of script paths
made the per-scope **Restrict paths** buttons disappear.

The selected-scope chip renders the whole comma-joined path list as one unbreakable
string (`jobs:run:scripts:f/a/b,f/c/d,u/e/f`). Nothing capped its width, and two ancestors
propagated that width outwards instead of constraining it:

- `CreateToken`'s card carried `min-w-min`, so it sized itself to that chip and pushed the
  rest of the form past the right edge of the user-settings panel — the second grid column
  ("Write", "Run flows" and their "Restrict paths" buttons) ended up off-screen.
- `Modal`'s body wrapper is a `flex-1` item without `min-w-0`, so the same content
  stretched the "Edit token" modal past its `max-w`, with the same result.

## Changes

- `ScopeSelector.svelte`: factor the three copies of the scope chip into one `scopeChip`
  snippet that truncates its label (`min-w-0 max-w-full` + inner `truncate`) and exposes
  the full value via `title`.
- `CreateToken.svelte`: drop `min-w-min` from the token card so it stays bounded by the
  panel and its content shrinks instead.
- `Modal.svelte`: add `min-w-0` to the modal body wrapper so wide content can shrink
  rather than stretch the modal past its max width.
- `ScopeSelector.svelte`: the popover trigger reads "Restrict paths" only while a scope has
  no paths; once it has some it becomes "+ Add path", which is the action you actually want
  at that point (and what the report was looking for).

The scope strings sent to the API are untouched. One incidental unification: the snippet
applies the component's `disabled` prop to all three remove buttons, where the resource-path
chip previously had none — unreachable today (no consumer passes `disabled`) and the more
correct of the two behaviors.

## Screenshots

**Before** — 3 script paths on `jobs:run:scripts`; the right column and its "Restrict
paths" button are pushed out of the panel:

before

**After** — same state, chip truncates, both "Restrict paths" buttons stay reachable:

after

**Edit token modal, before** — same content stretched the modal past `max-w-3xl`:

edit-modal-before

**Edit token modal, after**:

edit-modal-after

**Button label** — "Restrict paths" while empty (Run flows), "+ Add path" once a scope has
paths (Run scripts):

add-path-label

**Narrow viewport (820px)** — still contained:

narrow

## Test plan

- [x] User settings → Tokens → "Limit token permissions" → Jobs → "Run scripts" →
      "Restrict paths": add several `f/…`/`u/…` paths and confirm nothing overflows and
      both "Restrict paths" buttons stay visible (measured `scrollWidth == clientWidth` at
      every level of the panel).
- [x] Same check in the "Edit token" modal on an already-scoped token.
- [x] Repeated at 820px and 420px viewport widths.

---
## Summary by cubic
Fixes overflow in the token scope builder and Edit token modal when long scope/path lists are selected, keeping buttons visible and the modal within its max width.

- **Bug Fixes**
  - `ScopeSelector.svelte`: added a reusable `scopeChip` that truncates long labels and shows the full value in `title`.
  - `ScopeSelector.svelte`: the path popover now shows "+ Add path" with a plus icon once paths exist; otherwise "Restrict paths".
  - `CreateToken.svelte`: removed `min-w-min` from the token card so content doesn’t push the panel wider.
  - `Modal.svelte`: added `min-w-0` to the body flex item so wide content shrinks instead of stretching the modal.

<sup>Written for commit e51cacfcc49b15b0fef025f806765727750d0912. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10517?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/common/modal/Modal.svelte
+++ b/frontend/src/lib/components/common/modal/Modal.svelte
@@ -138,7 +138,9 @@
 									>
 								{/if}
 								<div class="flex">
-									<div class="text-left flex-1">
+									<!-- min-w-0: without it this flex item takes its content's min-content width and
+									     stretches the modal past its max-width instead of letting content shrink. -->
+									<div class="text-left flex-1 min-w-0">
 										<div class="flex flex-row items-center justify-between">
 											<h3 class="text-emphasis text-lg font-semibold">{title}</h3>
 											{@render settings?.()}
--- a/frontend/src/lib/components/settings/CreateToken.svelte
+++ b/frontend/src/lib/components/settings/CreateToken.svelte
@@ -173,7 +173,9 @@
 </script>
 
 <div>
-	<div class="p-4 rounded-md mb-6 min-w-min bg-surface-tertiary">
+	<!-- Stays bounded by the panel width: a content-driven width (min-w-min) would let a long
+	     scope chip stretch this card and push the rest of the form out of view. -->
+	<div class="p-4 rounded-md mb-6 bg-surface-tertiary">
 		<h3 class="pb-2 font-semibold text-emphasis text-sm">{title}</h3>
 
 		{#if showMcpMode && !mcpOnly}
--- a/frontend/src/lib/components/settings/ScopeSelector.svelte
+++ b/frontend/src/lib/components/settings/ScopeSelector.svelte
@@ -490,6 +490,25 @@
 	fetchScopeDomains()
 </script>
 
+<!-- The label can be a long comma-joined path list: it has to truncate rather than widen its
+     container, which would push the per-scope "Restrict paths" buttons out of the panel. -->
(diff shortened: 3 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 0f1515fea46d4022ca90944c7af34344d05e304a

PR #10519, opened 2026-08-04
feat: show far more in the home tree view and say what is not loaded

### LINKED ISSUE #10502: feature: Lazy load scripts in Home page tree view
Lazy load scripts in windmill workspace Home page when in tree view (like [what was just added for s3 file explorer](https://github.com/windmill-labs/windmill/commit/386e115daaf850ac45d10f2240bacb5758d238e6))

Currently there is no way to know that not all scripts loaded yet in a home page subfolder.
Image
Only after scrolling all the way down (hidden below bottom of screen) do you see there is a "Load more" button
Image

I spent 30minutes trying to find my missing scripts (and was super confused because they did appear in Github) only to realise I had to scroll to the bottom of the page to see a tiny button to "Load more"...

(I was tempted to submit this as a bug report as current behaviour is seriously confusing.)

rubenfiszel:
## Summary

Fixes #10502. Follows up on WIN-2297 (tree-mode home).

Browsing the home tree, several caps hid rows without saying so, and the only way out of
each was a small link you had to find:

- the tree rendered **15 root nodes** and put `15 root nodes out of 202 · load 30 more` in
  muted text *below* the list frame — the reporter spent 30 minutes hunting for scripts that
  were simply sliced off;
- a subfolder under a partially-loaded owner showed the number of rows that happened to be
  loaded (`sub_00 (16 items)` when it held 133), which is indistinguishable from a complete
  count — and there was **no way to complete a subfolder** short of paging its entire owner;
- the per-owner pager said `Load more in f/bigfolder (100 loaded)` — a number with nothing
  to compare it to.

This makes the tree show much more per step, lets **every folder page within its own
prefix** at any depth, and says with both numbers wherever something is still missing.

## Changes

**Show a lot more per step**
- Root nodes: 15 → **100 per slice**, and the next slice reveals itself as you scroll
  (`IntersectionObserver` on the footer). A root node is a collapsed owner row that fetches
  nothing until expanded, so a large slice costs a row each and no request. Auto-reveal
  stops at 500 rendered — the rows aren't virtualized — after which the footer's button
  reveals the rest on demand.
- Per-prefix page size 100 → **300**, and the client-side per-node cap 15 → 30, so most
  folders arrive whole on the first click.
- A node that pages server-side now shows *all* its loaded rows when opened on its own, so
  there's one control to reach the rest instead of a client "Show more" in front of the
  server "Load more".

**Every folder pages itself, at any depth**
- `ownerLoad` is now keyed by the full path prefix a node covers (`f/x`, `u/x`, or
  `f/x/sub/…`) rather than only top-level owners; the listing endpoint already scopes on an
  arbitrary `path_start`, so no backend change. A subfolder's "Load more" fetches
  `path_start=f/x/sub/` and merges into the same store (dedup by item key; rows keep the
  server's order via `ord`).
- Once a subfolder has been loaded directly, its own pagination governs it: it reads
  `(133 items)` exact and drops its pager even while its owner still has pages left.
- Opening a subfolder issues **no** request — that would be one per subfolder. Only the
  explicit pager loads, and only when an ancestor reports unloaded pages.
- Only a forced refresh (re-sort / re-filter) replaces a prefix's rows; every other load
  merges. A nested prefix inherits rows an ancestor's pages already brought in, and one page
  of its own can cover fewer of them than are on screen — replacing made 700 rows become 300
  on the very click meant to add more.
- A replacing load drops the rows *and the load state* of everything under it, so a nested
  folder can never keep an "exact" count over rows an ancestor's page truncated. Reloads
  (sort/kind/archive/mutation) re-fetch the open prefixes shallowest-first, so a parent
  landing after its child can't wipe what the child just re-fetched.
- A subfolder's stream restarts at its first row, which its ancestor may already have loaded,
  so one click pages past what it already has (bounded at 5 pages) rather than fetching a
  page that adds nothing and looking broken. A folder already fully covered by its ancestor's
  pages therefore resolves from `(700+ items)` to `(700 items)` in one click.
- Collapsing a node at any depth removes it from the reload fan-out.

**Say what isn't loaded**
- Footer is now the last row of the tree frame, not a caption under it:
  `Showing 100 of 702 folders and users` + `Show more`. In selected-owner mode, where the
  tree groups the paged browse stream instead, it reads `Not all items are loaded yet`.
- Per-node pager states both totals: `Showing 300 of 800 items in f/bigfolder` /
  `Showing 50 items in f/bigfolder/sub_00`, with a `Load more` button that stays put and
  shows a loading state while the page is in flight (it used to vanish mid-fetch).
- Partially-loaded folders read `(N+ items)` instead of a count that looks exact.
- Selected-owner mode gets the owner counts too, so its node shows `(800 items)` rather than
  the 53 rows that happened to be in the loaded window. Counts are floored by the rows
  rendered beneath them (the count endpoint can miss an individually-shared item), matching
  what the owner chips already do. Search and label-filter modes still get no owner totals —
  there the rows are a filtered subset, so an owner total would mislead.

## Screenshots

Same workspace throughout: 700+ owners, `f/bigfolder` holding 800 scripts.

**Root list — before / after** (before: sliced at 15 with the pager below the frame; after:
100 per slice, auto-revealing, footer inside the frame)

BEFORE-roots

AFTER-roots-footer

*(the footer is photographed with auto-reveal stubbed out — in normal use scrolling to it
reveals the next slice, and the button takes over past the 500-row ceiling.)*

**Expanded folder — before / after** (subfolder counts `(16 items)` / `(17 items)` were just
"how many rows loaded"; now `(50+ items)`)

BEFORE-folder-open

AFTER-folder-open

**Owner pager — before / after**

BEFORE-folder-more

AFTER-folder-more

**New: the subfolder's own pager, and the folder completed by it** (`sub_00` goes from
`(50+ items)` to `(133 items)` on one click, while its siblings stay partial)

NEST-subfolder-pager

NEST-subfolder-complete

## Known limitations

A subfolder that got **no** rows from its owner's loaded pages has no node yet, so it can't
offer to load itself — paging the owner is still what surfaces it.

A folder whose size is an exact multiple of the page size shows one final "Load more" that
returns nothing (keyset pagination mints a cursor for any full page). The click is
self-correcting — the control disappears — and suppressing it from the owner *count* instead
was rejected: that count is decorative and can under-report, so trusting it would hide rows
rather than a button. Listing a folder's
immediate children up front (the S3 explorer's one-level-at-a-time model) would need a new
endpoint returning distinct subfolder prefixes plus their counts; that's the natural next
step, not this PR.

## Test plan

Driven with Playwright against a seeded workspace (702 folders/users, `f/bigfolder` = 800
scripts across 3 subfolders):

- [x] `npm run check` — 0 errors (65 pre-existing warnings, none in the touched files)
- [x] Root list renders 100 on load, auto-reveals to 500 by scrolling, then stops; `Show
      more` continues past it (500 → 600 of 702)
- [x] Expanding `f/bigfolder` loads 300 rows → `Showing 300 of 800 items in f/bigfolder`;
      `Load more` → 600 of 800
- [x] Subfolder pager: `Showing 50 items in f/bigfolder/sub_00` → one request with
      `path_start=f/bigfolder/sub_00/` → `(133 items)`, pager gone, siblings still `(50+ items)`,
      owner pager consistent at `383 of 800`
- [x] 283 rows rendered, 133 under `sub_00`, **zero duplicate paths** after a re-sort that
      re-fetched both the parent and the nested prefix
- [x] Re-sort with a nested prefix open re-fetches in depth order (global → `f/bigfolder/` →
      `f/bigfolder/sub_00/`) and the subfolder stays complete
- [x] A subfolder holding 700 of its owner's 900 items keeps all 700 rows through its own
      "Load more" (it dropped to 300 before the merge fix), and that one click resolves it to
      an exact `(700 items)`
- [x] A folder whose only direct child is a subfolder of 150 scripts reads `(150 items)`, not
      `(1 item)` — badges and pagers both count leaves
- [x] Collapsing the nested folder removes it from the fan-out (later reloads request only
      the global stream + the owner); re-opening it — including while its first load is still
      in flight — re-registers it without a second request, and it stays complete across a
      re-sort
- [x] Reopening an owner mounts no transient subtree (peak DOM rows == steady rows)
- [x] `Expand all` still issues at most 20 owner requests with the larger root slice, and
      scrolling to reveal the remaining ~600 roots issues **0** further requests
- [x] Selected-owner mode: `(800 items)` + `Not all items are loaded yet` / `Load more`;
      subfolders read `(16+ items)` there too
- [x] Search in tree view, and the flat (non-tree) list's own pager, unchanged
- [x] Dark mode

Fixes WIN-2297

### DIFF
--- a/frontend/src/lib/components/home/ItemsList.svelte
+++ b/frontend/src/lib/components/home/ItemsList.svelte
@@ -346,14 +346,15 @@
 		}
 	}
 
-	// Per-top-level-folder lazy loading for tree view: expanding a folder loads its
-	// items on demand (server-scoped to that folder), paginated within the folder,
-	// instead of relying on the global browse window. Keyed by folder name (the
-	// `f/<name>` node).
+	// Per-folder lazy loading for tree view: expanding a folder loads its items on demand
+	// (server-scoped to that folder), paginated within it, instead of relying on the
+	// global browse window. Keyed by the full path prefix a node covers — a top-level
+	// owner (`f/<name>` / `u/<name>`) or any folder under one (`f/<name>/<sub>/…`), since
+	// the listing endpoint scopes on an arbitrary `path_start`. That is what lets a
+	// subfolder be completed on its own instead of only by paging its whole owner.
 	//
-	// Every top-level namespace — a folder (`f/<name>`) OR a user (`u/<name>`) — is a
-	// lazily-loaded owner, keyed here by its full path prefix. Its items live in their
-	// OWN store (`treeOwnerItems`), never merged into the global browse arrays: those
+	// Rows for every prefix live in ONE store (`treeOwnerItems`), never merged into the
+	// global browse arrays: those
 	// advance by a single `serverCursor`, so injecting out-of-window rows there would
 	// make the flat stream non-contiguous and, once pagination reached those rows,
 	// duplicate them. Loading users lazily too (rather than sourcing them from the
@@ -367,20 +368,17 @@
 		loading: boolean
 		loaded: boolean
 		gen: number
-		// Total rows fetched for this owner so far (leaves + subfolder rows) — the tree
-		// node's own children count undercounts because rows nest into subfolders.
-		count: number
 	}
-	// Rows per request when expanding an owner in the tree. Larger than the flat list's
-	// page because a tree row is a single line and an owner is opened to see what it
+	// Rows per request when loading a prefix in the tree. Larger than the flat list's
+	// page because a tree row is a single line and a folder is opened to see what it
 	// holds — most folders come in whole on the first click.
 	const OWNER_PAGE_SIZE = 300
 	let ownerLoad = $state<Record<string, OwnerLoadState>>({})
 	let treeOwnerItems = $state<ItemType[]>([])
 	let treeGen = 0
-	// Owners the user currently has expanded (full path prefixes). A reload re-fetches
-	// only these: ownerLoad also retains collapsed owners as a cache, so keying reloads
-	// off its entries would re-request every owner ever opened in this scope.
+	// Prefixes currently loaded and on screen. A reload re-fetches only these: ownerLoad
+	// also retains collapsed nodes as a cache, so keying reloads off its entries would
+	// re-request every folder ever opened in this scope.
 	let openOwners = new Set<string>()
 
 	// The endpoint returns a unified `edited_at` per row; combinedItems derives every
@@ -393,14 +391,14 @@
 		} as unknown as ItemType
 	}
 
-	// `owner` is the full prefix: `f/<folder>` or `u/<username>`.
-	// `force` re-fetches an owner's first page even if already loaded (a re-sort /
-	// re-filter reload uses it to refresh the loaded rows in place).
+	// `owner` is the full prefix a node covers: `f/<folder>`, `u/<username>`, or any
+	// folder under one. `force` re-fetches its first page even if already loaded (a
+	// re-sort / re-filter reload uses it to refresh the loaded rows in place).
 	async function loadOwnerItems(owner: string, more = false, force = false): Promise<void> {
 		const ws = $workspaceStore
 		if (!ws || !$userStore) return
-		// Track the owner as open first — even a no-op call (re-expanding a cached owner)
-		// means it's expanded, so later reloads must refresh it.
+		// Track the prefix as open first — even a no-op call (re-expanding a cached node)
+		// means it's on screen, so later reloads must refresh it.
 		openOwners.add(owner)
 		const st = ownerLoad[owner]
 		// Only a load for the CURRENT generation blocks a new one. A load left in flight
@@ -414,9 +412,7 @@
 			hasMore: st?.hasMore ?? false,
 			loading: true,
 			loaded: st?.loaded ?? false,
-			gen,
-			// A fresh (non-more) load replaces this owner's rows, so its count restarts.
-			count: more ? (st?.count ?? 0) : 0
+			gen
 		}
 		const { orderBy, orderDesc } = sortToParams(sortOrder)
 		let res: { items: RunnableItem[]; next_cursor?: string }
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Heads-up: #10522 was built against the same issue (#10502) and lands on the same three
files. It converged independently on the same core decisions — an `OWNER_PAGE_SIZE`
constant at the same site, root nodes → 100, and moving the root window into
`TreeViewRoot` (dropping the `nbDisplayed` `$bindable`).

**This PR is the one to keep.** It answers the half of #10502 that #10522 doesn't —
*"there is no way to know that not all scripts loaded yet"* — via the in-frame footers,
`Showing N of M items in <owner>`, and `(N+ items)` on partially-loaded subfolders. It
also uses the design-system `Button` where #10522 kept raw `<div onclick>` rows. I'm
closing #10522 as superseded.

One idea from it that may still be worth taking, **judged against the per-prefix
pagination you have in flight rather than grafted blind** — in #10522, `effectiveMax`
became:

```js
let effectiveMax = $derived.by(() => {
	if (!isFolder(item) && !isUser(item)) return 0
	return collapseAll && !isSearching ? item.items.length : Math.min(item.items.length, showMax)
})
```

The rule: a node opened *by hand* renders every row already loaded for it — the click is
the request to see it — while "expand all" and search, which open every node without a
click, keep the `showMax` cap. Today that rule applies only to top-level lazy owners, so
a hand-opened subfolder with 120 rows already in memory reads `Showing 30 of 120 loaded
items` and needs four clicks to reveal rows that cost nothing to render: a client cap
stacked on top of the server page. The `!isSearching` guard is load-bearing — dropping
`isLazyOwner && ownerState != undefined` would otherwise uncap search too (it works today
only because `treeLazyMode` is false while searching, so `ownerLoad` is already
`undefined`).

Two notes from #10522's review round that apply here as well:

- `TreeView.svelte:69` — `opened` is `$state(true)` and only set false by the effect at
  line 111, which runs after the first render commits. So opening a node transiently
  mounts its whole subtree before collapsing it. Bounded today by the per-node cap; it
  gets more expensive as nodes render more. `$state(false)` looked safe (the effect
  assigns unconditionally, and search rendering is gated on `opened || isSearching`).
- Raising `OWNER_PAGE_SIZE` also multiplies the `reloadItems` fan-out, which fires on
  every row mutation and every sort/archive/kind change, not just on expand — worth
  exercising with several folders left open.

rubenfiszel:
✅ Review round clean @ 028416e2c1d3367df864c6deaeac28db66485934 — nit-only verdicts at 8d4bfaa9ce23c028e2a0c340a00e2e09ec2fe420 (claude: good to merge, pi: good to merge, codex: mergeable with nits); the comment-length nits are addressed in 028416e2c1.

PR #10520, opened 2026-08-04
feat: preview and edit steps inside expanded subflows

rubenfiszel:
## Summary

In the flow editor, expanding a subflow inlines its steps in the graph — but selecting one of those steps showed only *"Selected step is witin an expanded subflow and is not directly editable in the flow editor"*: no preview of the step, and no way to reach the flow that owns it.

Selecting such a step now shows a read-only preview of it (the same step viewer the flow graph viewer uses: step inputs, code, lockfile, iterator expression, branch predicates, nested flow graph) plus a breadcrumb of the subflow chain it lives in, an **Edit** button that opens that subflow in the flow editor drawer *with the step already selected*, and an open-in-new-tab link to `/flows/edit/<subflow>?selected=<step>`.

Fixes WIN-2316

## Changes

- `expandedSubflowStep.ts` — resolver for a `subflow:A:B:leaf` graph node id: follows each `Flow{path}` boundary through the flow it points at, level by level. Each boundary may itself sit inside a for-loop or branch (the node id only records subflow boundaries), so every level is walked with `findStepPath`. When the id points at something that isn't a module of the containing flow (an AI agent tool node, a stale selection) it still reports the containing flow, so the Edit action stays useful.
- `ExpandedSubflowStep.svelte` — the new panel, replacing the placeholder text. Resolutions are request-token guarded so a slower deep chain can't overwrite a newer selection, and the subflow modules it fetches are memoised for as long as the panel stays mounted (clicking through an expansion no longer refetches the chain per click).
- `FlowEditorDrawer.openDrawer` takes an optional step id to select on open; its "Edit in new tab & close drawer" action now follows the drawer's live selection.
- `FlowGraphV2.reloadExpandedSubflows()` refetches the steps inlined by every expanded subflow after one is deployed from the drawer, outermost first. Each expansion's path is re-derived from the current tree (the edited flow, or its refreshed parent expansion) rather than remembered from expand time, so a step that now points at a different flow stops rendering the old one; expansions whose step is gone are dropped, and one failing fetch no longer aborts the rest.

## Test plan

- [x] `npm run check` — 0 errors, no new warnings
- [x] `expandedSubflowStep.test.ts` — multi-level descent through boundaries nested in containers, plus the unresolvable-leaf fallback
- [x] Expand a subflow in the flow editor, select a step inside it → preview + breadcrumb + Edit
- [x] Two-level expansion → breadcrumb lists both flows, Edit opens the innermost one
- [x] Step inside a for-loop of an expanded subflow resolves; loop container itself previews its iterator
- [x] Edit → drawer opens the subflow at that step; deploy → toast, panel refreshes, inlined graph picks up the change without a reload
- [x] Repoint an intermediate subflow step at another flow and deploy → the nested expansion re-resolves to the new flow instead of keeping the old one

## Screenshots

Before — selecting a step inside an expanded subflow:

before

After — preview, breadcrumb, Edit and open-in-new-tab:

after

Two-level expansion (`f/test/inner › f/test/other`):

after-nested

Edit opens the owning subflow in the drawer, already on that step:

after-drawer

---
## Summary by cubic
Show a read-only preview and quick edit path for steps inside expanded subflows in the flow editor. Delivers WIN-2316 by letting users inspect a step and jump straight to the owning subflow.

- New Features
  - Selecting a step in an expanded subflow shows a read-only preview (inputs, code, lockfile, iterator, branch predicates, nested graph) with a breadcrumb of the subflow chain.
  - Edit opens the owning subflow in the drawer with that step selected, plus an open-in-new-tab link that preserves the selection.
  - Resolver walks nested subflow boundaries (across loops/branches) to the selected step; falls back to the containing flow if the leaf isn’t a direct module.
  - After deploying from the drawer, inlined steps are reloaded with paths re-derived; stale expansions are dropped and one failing fetch doesn’t stop others. Exposes reloadExpandedSubflows() for the panel to trigger.

- Bug Fixes
  - Hide the Edit button when no flow editor drawer is available.
  - “Open in new tab” links are base-prefixed, encode the workspace, and from the drawer follow the drawer’s current selection.
  - Prevent a pre-deploy response from repopulating the subflow module cache after it’s invalidated; slower resolutions can’t overwrite newer selections.
  - Expanded-subflow reloads are guarded: collapsing during reload won’t revive it, newer reloads supersede older ones, and a reload response applies only to the exact expansion it was fetched for.

<sup>Written for commit db0b25b50d74c1fedb94643adeacd1d7ffecbb7a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/flows/content/ExpandedSubflowStep.svelte
+++ b/frontend/src/lib/components/flows/content/ExpandedSubflowStep.svelte
@@ -56,7 +56,10 @@
 			let modules = modulesCache.get(key)
 			if (!modules) {
 				modules = (await FlowService.getFlowByPath({ workspace: ws!, path })).value.modules
-				modulesCache.set(key, modules)
+				// A response from before the last deploy must not repopulate the cache it cleared.
+				if (gen === cachedGeneration) {
+					modulesCache.set(key, modules)
+				}
 			}
 			return modules
 		})
--- a/frontend/src/lib/components/flows/content/ExpandedSubflowStep.svelte
+++ b/frontend/src/lib/components/flows/content/ExpandedSubflowStep.svelte
@@ -1,0 +1,112 @@
+<script lang="ts">
+	import { getContext, untrack } from 'svelte'
+	import { resource } from 'runed'
+	import { FlowService, type FlowModule } from '$lib/gen'
+	import { workspaceStore } from '$lib/stores'
+	import Badge from '$lib/components/common/badge/Badge.svelte'
+	import Button from '$lib/components/common/button/Button.svelte'
+	import Skeleton from '$lib/components/common/skeleton/Skeleton.svelte'
+	import FlowGraphViewerStep from '$lib/components/FlowGraphViewerStep.svelte'
+	import { ChevronRight, ExternalLink, Pen } from 'lucide-svelte'
+	import { sendUserToast } from '$lib/utils'
+	import type { FlowEditorContext } from '../types'
+	import { resolveExpandedSubflowStep } from '../expandedSubflowStep'
+	import { parseExpandedSubflowId } from '$lib/components/restartFromStepPath'
+
+	interface Props {
+		/** Graph node id of the selected step, of the form `subflow:<step>[:<step>...]:<leaf>`. */
+		selectedId: string
+		/** Called once the subflow was deployed from the drawer, to refresh the inlined graph. */
+		onSubflowUpdated?: () => void
+	}
+
+	let { selectedId, onSubflowUpdated }: Props = $props()
+
+	const { flowStore, flowEditorDrawer, opWorkspace } =
+		getContext<FlowEditorContext>('FlowEditorContext')
+	let opWs = $derived(opWorkspace?.() ?? $workspaceStore)
+
+	let leafId = $derived(parseExpandedSubflowId(selectedId)?.leaf ?? selectedId)
+
+	let step = resource([() => selectedId, () => opWs], async ([id, ws]) =>
+		resolveExpandedSubflowStep(
+			id,
+			untrack(() => $state.snapshot(flowStore.val.value.modules) as FlowModule[]),
+			async (path) => (await FlowService.getFlowByPath({ workspace: ws!, path })).value.modules
+		)
+	)
+
+	// While a new selection resolves, `step.current` still holds the previous step: hide the
+	// breadcrumb and the actions rather than pointing them at the step the user just left.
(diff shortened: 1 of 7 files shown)


rubenfiszel:
Re @pi's P1 (*incorrect parent lookup in `expandedSubflowPath`*): this rests on a misreading of `parseExpandedSubflowId`, so there is no bug to fix.

The review assumes `parseExpandedSubflowId('subflow:a:b')` → `{ subflowSteps: ['a', 'b'], leaf: 'b' }`. It actually returns `{ subflowSteps: ['a'], leaf: 'b' }` — the last segment is always the leaf and `subflowSteps` is `parts.slice(0, -1)` (`restartFromStepPath.ts:101-110`, pinned by `restartFromStepPath.test.ts:50`). So the `steps.length > 1 ? 'subflow:' + steps.join(':') : steps[0]` branch already yields `'a'` for `subflow:a:b` and `'subflow:a:b'` for `subflow:a:b:c` — exactly the mapping the review asks for. Claude's review independently verified the same mapping.

Verified end to end rather than only on paper: with `f/test/parent` → `b` (→`f/test/inner`) → `b` (→`f/test/other`) both expanded, editing the innermost step from the drawer and deploying keeps the nested expansion and refreshes it — node `subflow:b:b:z` relabels to "OTHER FLOW STEP v2" instead of disappearing.

To make the derivation harder to misread it now lives in the tested pure module as `expandedSubflowParentId`, with a unit test pinning `subflow:a:b` → `a` and `subflow:a:b:c` → `subflow:a:b` (078da2c837).

Also addressed in that commit: pi's P2 (the module cache is cleared when a deploy bumps the generation, instead of accumulating dead generations), Claude's P2 (`base` prefix on the new `/flows/edit/...` link — and on the drawer's `window.open`, whose query string this PR touches), and Claude's non-blocking nit (a failed refetch during `reloadExpandedSubflows` now raises an error toast rather than only a `console.error`).

rubenfiszel:
✅ Review round clean @ db0b25b50d74c1fedb94643adeacd1d7ffecbb7a

PR #10521, opened 2026-08-04
fix: require an explicit user for azure workload identity on postgres

rubenfiszel:
## Summary

A Postgres resource that selects Azure workload identity (password = `ms_entraid`) but leaves `user` unset silently logged in as `postgres`. Azure maps each Entra principal to a role of its own, created with `pgaadauth_create_principal('<identity_name>', …)`, so `postgres` is never that role — the connection fails with a server error naming a role the user never configured, which is a confusing place to start debugging from.

That mode now requires an explicit `user`, and says what to set it to:

> Azure workload identity authentication requires `user` on the resource. Set it to the Postgres role the worker's Entra principal is mapped to, as created by pgaadauth_create_principal.

The check runs before the token exchange, so a misconfigured resource fails immediately rather than after a round trip to Entra ID.

**Scoped to workload identity only.** IAM RDS keeps its `postgres` default: that path is older and a deployment may depend on it, whereas `ms_entraid` shipped in v1.778.0 and nothing can yet rely on its fallback.

This does *not* apply to MSSQL, where the same sentinel ignores `user` entirely — the fedauth token is the whole identity assertion there and no username is sent. Postgres still needs a role name in the startup packet.

## Changes

- `PgDatabase::entra_login()` — resolves the Entra login or returns an actionable `BadRequest`. Treats a blank or whitespace-only `user` as unset, so a resource created through the API or git-sync can't slip an empty string through.
- `connect_with_workload_identity()` validates the login before minting a token; `connect_with_token()` now takes the login explicitly so the two token modes can differ on whether a default exists.
- `PgDatabase::login_name()` — the single definition of the implicit `postgres` default used by the non-Entra paths. Replaces six copies of `user.as_deref().unwrap_or("postgres")` across `to_uri()`, `connect_with_iam()`, `pg_dump_database()`, `pg_import_dump()` and the DuckDB `CREATE SECRET … USER` builder.
- Both token modes name the presented login in the job log, which is the one thing the token itself doesn't carry:
  ```
  Using Azure Workload Identity (login <role>)
  Using IAM RDS authentication (login <role>)
  ```
  Password auth stays silent, as before.

## Test plan

- [x] `cargo test -p windmill-common --lib entra_login` — blank/whitespace/absent `user` is rejected; the non-Entra default stays `postgres`
- [x] `cargo test -p windmill-worker --lib log_name` — explicit login, the `postgres` default on IAM, no invented login for Entra, and silence for password auth
- [x] `cargo check --features enterprise,private` (both token modes are EE-gated)
- [x] `cargo check -p windmill-worker --features duckdb` (`duckdb_executor.rs` only compiles behind this flag)
- [ ] On an EE build, run a Postgres script against an `ms_entraid` resource with no `user` and confirm it fails with the message above instead of a server error about `postgres`
- [ ] Same resource with `user` set to the pgaadauth role: connects, and the log reads `Using Azure Workload Identity (login <role>)`
- [ ] A resource with `use_iam_auth: true` still connects and logs `Using IAM RDS authentication (login <role>)`

Not verified locally: the handshake itself. Both token modes need a real Azure/AWS-backed server, so only the guard, mode selection and log text were exercised here.

### DIFF
--- a/backend/windmill-api-workspaces/src/workspaces.rs
+++ b/backend/windmill-api-workspaces/src/workspaces.rs
@@ -2642,7 +2642,7 @@ pub(crate) async fn pg_dump_database(
 
     let host = &pg_db.host;
     let port = pg_db.port.unwrap_or(5432).to_string();
-    let user = pg_db.user.as_deref().unwrap_or("postgres");
+    let user = pg_db.login_name();
     let dbname = &pg_db.dbname;
 
     let mut cmd = tokio::process::Command::new("pg_dump");
@@ -2686,7 +2686,7 @@ pub(crate) async fn pg_dump_database(
 async fn pg_import_dump(target_db: &PgDatabase, dump_file: &DumpFile) -> Result<()> {
     let host = &target_db.host;
     let port = target_db.port.unwrap_or(5432).to_string();
-    let user = target_db.user.as_deref().unwrap_or("postgres");
+    let user = target_db.login_name();
     let dbname = &target_db.dbname;
 
     let mut cmd = tokio::process::Command::new("psql");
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ ca267812ba8179979456b71703ffd625490037c5

PR #10523, opened 2026-08-04
fix: bound ScopeSelector badge heights and correct three scope-chip defects

rubenfiszel:
## Summary

Follow-up to #10517, rebased on top of it. That PR stopped long scope chips from widening
their container; this one bounds the summary **vertically** and fixes three defects found
while exercising the component around it.

Fixes WIN-2318

## Changes

All in `frontend/src/lib/components/settings/ScopeSelector.svelte`.

### 1. The Selected Scopes summary grew without bound (the original issue)

Because a truncated chip is capped at exactly the container width, every path-restricted
scope occupies a full row of its own. Seven of them make the summary 176px tall, and it
renders up to 10 chips, so the worst case pushes the scope domain list and the token form's
Label / Expires In / action row well down the panel. The summary is now `max-h-32
overflow-y-auto`.

The per-domain header needed **no** cap in the end: #10517's truncation holds every chip to
one row, and the row count is bounded by the domain's own scope count (4 for Jobs, the
maximum in the scope table; 2 for most). A height cap there would nest a 64px scroller
inside the already-scrollable `max-h-96` domain list, swallowing wheel events that cross the
row — so that half of the issue was already fixed by #10517 as a side effect.

### 2. `scopeChip` bound the wrong disabled state

The shared snippet used the component-level `disabled`, but a scope card computes
`isDisabled = disabled || isScopeDisabled(...)`. A scope superseded by its `:write` sibling
greyed out its checkbox and its path button while the `x` on its path chips stayed live, so
those paths could still be destroyed from a card that looked read-only. The effective state
is now passed in.

### 3. Truncation hid the grant in the summary

An ellipsis is right for the tight per-domain header, but the Selected Scopes summary is
where the paths being granted are read, and `title=` is hover-only and not keyboard
reachable. Summary chips now wrap; the header and per-scope path lists keep truncating.

### 4. The domain checkbox silently widened path-restricted scopes

Ticking a domain checkbox re-added its write and run scopes bare, dropping any resource
paths configured on them — a token restricted to one path silently became a token for the
whole domain, and unticking did not bring the paths back. The checkbox already reads as
checked when those scopes are path-restricted, so it now carries the paths over. Both
branches of the `requires_resource_path` conditional it replaces pushed the same bare
value, so nothing was reading that flag.

Also corrects the path popover tooltip, which claimed every path added widens the scope's
reach — true only from the second path on, since the first replaces a bare full-access
scope with a path-restricted one.

## Screenshots

Same state in both: 7 scopes across Jobs / Scripts / Flows, most path-restricted, Jobs expanded.

**Before** (cap removed at runtime on this same build) — the summary runs 176px and pushes
everything below it down:

final-before

**After** — summary capped at 128px and scrolling, the rest of the form back in view. Note the
long `jobs:run:flows:...` chip wrapping to two full lines in the summary while the same scope
truncates in the tight `Jobs` header row below it:

final-after

## Test plan

Driven in a real browser (Playwright MCP) against the running dev stack.

**Token creation** (User settings → Tokens → *Limit token permissions*):

- [x] A path-restricted chip measures exactly the container width (747px of 747px) — this is what makes the summary's vertical growth unbounded after #10517
- [x] 7 path-restricted scopes → summary `scrollHeight 176 / clientHeight 128`: capped and scrolling, not growing
- [x] Long chip in the summary wraps to 2 lines at 747px wide showing its full value; the same scope still truncates in the domain header
- [x] Removing a wrapped multi-line chip from the summary still works
- [x] Zero elements with `overflow-y: auto|scroll` remain inside the `max-h-96` domain list; the Jobs header renders 3 chips at 20px with `scrolls: false`
- [x] Restrict `scripts:read` to a path, then tick `scripts:write`: the path chip's `x` is now `disabled` alongside the greyed-out checkbox and path button (it was clickable before, and clicking it destroyed the path)
- [x] Restrict `jobs:run:scripts` to `f/only_this_one/safe_script`, tick the Jobs domain box: the grant stays path-restricted instead of widening to bare `jobs:run:scripts`; scopes with no configured paths are still granted unrestricted
- [x] Untick → retick round-trip returns bare `jobs:run:scripts`, so full domain access stays reachable and no restriction gets stuck
- [x] Tooltip renders the corrected wording

**Token editing** (Tokens table → Edit token → *Limit token permissions*), the second consumer:

- [x] Modal holds at 900px with a long path-restricted chip, no horizontal document overflow, full chip text visible

**Checks:**

- [x] `npm run check` — 0 errors (65 pre-existing warnings, none in this file)
- [x] Local Codex review (same policy/model as CI) — "Good to merge", no findings
- [x] CI round on `f4e8b50`: Codex / Claude / Pi all "Mergeable, but should ideally address nits"; both P2 nits fixed in `f6f1072`

### DIFF
--- a/frontend/src/lib/components/settings/ScopeSelector.svelte
+++ b/frontend/src/lib/components/settings/ScopeSelector.svelte
@@ -535,7 +535,7 @@
 					>Administrator scope grants full access to all resources.</p
 				>
 			{:else}
-				<div class="flex flex-wrap gap-2">
+				<div class="flex flex-wrap gap-2 max-h-32 overflow-y-auto">
 					{#each selectedScopes.slice(0, 10) as scope}
 						{@render scopeChip(scope, 'Remove scope', () => removeSelectedScope(scope))}
 					{/each}
@@ -593,7 +593,7 @@
 							</div>
 
 							<div class="flex-1 min-w-0 flex flex-col gap-1">
-								<div class="flex items-center gap-x-2 gap-y-1 flex-wrap">
+								<div class="flex items-center gap-x-2 gap-y-1 flex-wrap max-h-16 overflow-y-auto">
 									<!-- svelte-ignore a11y_no_noninteractive_element_interactions -->
 									<label
 										for={`domain-${domain.name}`}
--- a/frontend/src/lib/components/settings/ScopeSelector.svelte
+++ b/frontend/src/lib/components/settings/ScopeSelector.svelte
@@ -598,7 +598,10 @@
 							</div>
 
 							<div class="flex-1 min-w-0 flex flex-col gap-1">
-								<div class="flex items-center gap-x-2 gap-y-1 flex-wrap max-h-16 overflow-y-auto">
+								<!-- No height cap here: truncation holds every chip to one row and a domain has a
+								     handful of scopes, so this row cannot run away. Capping it would nest a scroller
+								     inside the domain list, which then swallows wheel events crossing this row. -->
+								<div class="flex items-center gap-x-2 gap-y-1 flex-wrap">
 									<!-- svelte-ignore a11y_no_noninteractive_element_interactions -->
 									<label
 										for={`domain-${domain.name}`}
@@ -680,9 +683,9 @@
 															>
 																{resourcePathArray.length > 0 ? 'Add path' : 'Restrict paths'}
 																<Tooltip light>
-																	Restrict this scope to specific resource paths. Each path you add
-																	widens what the scope reaches. If no paths are specified, the
-																	scope gives full access.
+																	Restrict this scope to specific resource paths. With no paths
+																	listed it reaches everything; with paths listed it reaches exactly
+																	those.
 																</Tooltip>
 															</Button>
 														{/snippet}
--- a/frontend/src/lib/components/settings/ScopeSelector.svelte
+++ b/frontend/src/lib/components/settings/ScopeSelector.svelte
@@ -196,28 +196,22 @@
 					)
 			)
 
-			const writeScope = domain.scopes.find((s) => s.value === writeScopeValue)
-			if (writeScope?.requires_resource_path) {
-				selectedScopes = [...selectedScopes, `${writeScopeValue}`]
-			} else {
-				selectedScopes = [...selectedScopes, writeScopeValue]
-			}
+			// A path-restricted scope already counts as selected for this checkbox (see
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Both P2 nits addressed in `f6f1072`. On the second one I went further than the suggestion — flagging that explicitly since it diverges.

**Tooltip.** You're all right, and the wording was mine. "Each path you add widens" is only true from the second path on; the first replaces a bare full-access scope with a path-restricted one. Rather than patch the direction claim, the sentence now just states what each state means and lets the reader draw the direction:

> Restrict this scope to specific resource paths. With no paths listed it reaches everything; with paths listed it reaches exactly those.

**Domain-header scroller.** Both Claude and Pi suggested `overflow-hidden` + a `+N more` badge mirroring line 547. I started there, then found the badge can never render: the header shows `getSelectedScopesForDomain`, which is bounded by the domain's own scope count — 4 for Jobs (the maximum across the table), 3 for Apps, 2 everywhere else. And Jobs can only ever display 3 at once, because selecting `jobs:write` supersedes `jobs:read`. Any slice threshold that bounds the row is either never hit or throws away scopes users can currently see.

So I dropped the cap entirely instead. After #10517's truncation each chip is held to exactly one row, and the row count is bounded by the scope table, so the growth the cap was guarding against no longer exists — this is the half of WIN-2318 that #10517 already fixed as a side effect. The summary cap stays: that one is genuinely unbounded (10 chips, wrapping) and it is not nested inside another scroller.

Verified after the change: zero elements with `overflow-y: auto|scroll` remain anywhere inside the `max-h-96` domain list, and the Jobs header renders its 3 chips at 20px with `scrolls: false`.

**Also closed the gap Claude flagged as unverified** — the untick → retick round-trip. Restricting `jobs:run:scripts` to a path, then unticking and reticking Jobs:

```
restricted:   ["jobs:run:scripts:f/only_this_one/safe_script"]
afterTick:    ["jobs:write", "jobs:run:scripts:f/only_this_one/safe_script", "jobs:run:flows"]
afterUntick:  []
afterRetick:  ["jobs:write", "jobs:run:scripts", "jobs:run:flows"]
```

Full domain access is still reachable; there is no state where a restriction gets stuck.

And confirming the two things Claude could not check from its checkout: `npm run check` is 0 errors (65 pre-existing warnings, none in this file), and the `writeScopeValue` half of the new loop is indeed unreachable today for exactly the reason given — kept uniform as defence against a future domain that has both a path-capable write scope and run scopes.

PR #10530, opened 2026-08-05
fix(frontend): stop swallowing Ctrl/Cmd+Shift+S in the editors

rubenfiszel:
## Summary

Ctrl/Cmd+Shift+S was captured by Windmill and never reached the browser or OS, so whatever the user had bound to it (screenshot tool, devtools, extension) silently did nothing while an editor was open. Two independent layers were swallowing it.

**Page-level handlers.** `FlowBuilder` and `AppEditorHeader` normalize single-char keys with `event.key.toLowerCase()` in their keydown switch. That normalization exists so Ctrl+Shift+Z still reaches `case 'z'` for redo, but `case 's'` had no shift guard, so Ctrl+Shift+S fell into the save branch, flushed the draft and called `preventDefault()`.

**Monaco.** `monaco-editor` is aliased to `@codingame/monaco-vscode-editor-api`, which ships VS Code's keybinding set rather than the small standalone one. Ctrl+Shift+S resolved to `workbench.action.files.saveAs` and was swallowed inside every Monaco instance before the event could leave the editor. The command is meaningless here, so the binding is dropped at init. Probing neighbours confirms the shape: Ctrl+Shift+O and Ctrl+Shift+P are likewise taken (Go to Symbol, Command Palette) while unbound chords such as Ctrl+Shift+Q pass through untouched.

The second commit closes the sites that were taking Ctrl/Cmd+S without doing anything with it. Monaco stops propagation of the keydown, so a focused code editor never reaches the page-level handler; `Editor`/`SimpleEditor` already re-broadcast it as `wm-monaco-save-shortcut` for exactly this reason, but only `RawAppEditor` listened. In the app editor, Ctrl+S from an inline script or template editor was therefore a silent no-op — the draft only landed on the next autosave tick. The flow and script editors got there by accident, via the `formatAction` chain, which not every embedded editor wires up.

Ctrl/Cmd+S is unchanged in intent everywhere: flush the pending draft autosave (and format, in the script editor).

## Changes

- `FlowBuilder.svelte` / `AppEditorHeader.svelte`: require `!event.shiftKey` in the `case 's'` save branch
- `vscode.ts`: after vscode-api starts, `meditor.addKeybindingRule({ keybinding: CtrlCmd|Shift|KeyS, command: null })` to unbind Save As
- `ScriptBuilder.svelte`: lowercase the key in its switch, so Ctrl+S still saves with Caps Lock on (`event.key` is `S`), and exclude shift like the other two
- `FlowBuilder.svelte` / `ScriptBuilder.svelte` / `AppEditorHeader.svelte`: listen for `wm-monaco-save-shortcut` and flush, so Ctrl+S from any focused Monaco persists the draft and flashes the indicator. A redundant flush (script/flow, which also arrive via `formatAction`) is a no-op
- `TemplateEditor.svelte`: its Ctrl+S command was an empty function that swallowed the key and did nothing; it now materializes the code and re-broadcasts like the other editors
- `AppEditorHeader.svelte`: `preventDefault()` moved inside the guard, so the contexts that cannot act on the key (AI session pane, no draft path) leave it to the browser

The keybinding rule is scoped to the in-page Monaco instance shared by `Editor`, `SimpleEditor`, `TemplateEditor` and `DiffEditor`. The raw-app UI-builder iframe runs its own VS Code bundle and is deliberately untouched, so Ctrl+Shift+S still behaves like VS Code in there.

## Screenshots

No visible UI effect (keyboard shortcut routing only), so there is nothing to screenshot. Verified behaviourally instead, see below.

## Test plan

Verified in Chromium against a local dev server, reading `defaultPrevented` on a window capture listener after dispatch (capture is needed because Monaco stops propagation before the event reaches window):

| context | Ctrl+Shift+S before | Ctrl+Shift+S after | Ctrl+S after |
| --- | --- | --- | --- |
| flow editor | swallowed | passes through | saves |
| app editor | swallowed | passes through | saves |
| script editor, canvas focus | passes through | passes through | saves |
| script editor, code focus | swallowed | passes through | saves |

The "before" column is measured, not inferred: the fix was stashed, the page reloaded, and the swallow reproduced.

Second commit, measured the same way:

- App editor, edit then fire the Monaco re-broadcast: draft POST lands **0 ms after the event**, against **1516 ms** for the same edit left to the autosave debounce — the flush is real, not the debounce firing anyway.
- Script editor with the code editor focused: Ctrl+S emits exactly one `wm-monaco-save-shortcut`, and is still swallowed by Monaco (so the page-level handler could never have seen it).
- Caps Lock form of the shortcut (`key === 'S'`, no shift) now saves in all three editors; it was ignored in the script editor before.
- Ctrl+Shift+S still passes through in all three editors, with the canvas focused and with a code editor focused.

Manual checks worth repeating on review:

- [ ] Flow editor: Ctrl/Cmd+S saves the draft, Ctrl/Cmd+Shift+S reaches the browser/OS, Ctrl/Cmd+Shift+Z still redoes
- [ ] App editor: same three, plus Ctrl/Cmd+S from an inline script editor shows the autosave indicator
- [ ] Script editor with the caret inside the code editor: Ctrl/Cmd+S still saves and formats
- [ ] Raw app source editor: Ctrl/Cmd+Shift+S still handled by the embedded VS Code, unchanged

No test shipped: the diff is keydown guards, a keybinding-manager call and event wiring, none of which can be asserted without a full editor fixture.

### DIFF
--- a/frontend/src/lib/components/Editor.svelte
+++ b/frontend/src/lib/components/Editor.svelte
@@ -10,7 +10,7 @@
 	import { buildWsUrl } from '$lib/wsUrl'
 	import { sendUserToast } from '$lib/toast'
 
-	import { createEventDispatcher, onDestroy, onMount, untrack } from 'svelte'
+	import { createEventDispatcher, onDestroy, onMount, tick, untrack } from 'svelte'
 
 	// import libStdContent from '$lib/es6.d.ts.txt?raw'
 	// import domContent from '$lib/dom.d.ts.txt?raw'
@@ -1670,9 +1670,11 @@
 				// Monaco swallows the keydown (addCommand prevents default and
 				// stops propagation), so page-level Ctrl/Cmd+S handlers never
 				// see it. Re-broadcast as a window event so editors that flush
-				// a draft on the shortcut (raw apps) can react regardless of
-				// which Monaco has focus.
-				window.dispatchEvent(new CustomEvent('wm-monaco-save-shortcut'))
+				// a draft on the shortcut can react regardless of which Monaco
+				// has focus. Only after `tick()`: the autosave payload is parked
+				// by a `$effect`, so dispatching synchronously would make every
+				// listener flush the state from before `updateCode()`.
+				void tick().then(() => window.dispatchEvent(new CustomEvent('wm-monaco-save-shortcut')))
 			})
 
 			editor?.addCommand(KeyMod.CtrlCmd | KeyCode.Enter, function () {
--- a/frontend/src/lib/components/FlowBuilder.svelte
+++ b/frontend/src/lib/components/FlowBuilder.svelte
@@ -901,7 +901,9 @@
 				}
 				break
 			case 's':
-				if (event.ctrlKey || event.metaKey) {
+				// Shift excluded: the switch lowercases so Ctrl+Shift+S lands here
+				// too, and swallowing it would steal the browser/OS shortcut.
+				if ((event.ctrlKey || event.metaKey) && !event.shiftKey) {
 					saveDraft()
 					event.preventDefault()
 				}
--- a/frontend/src/lib/components/FlowBuilder.svelte
+++ b/frontend/src/lib/components/FlowBuilder.svelte
@@ -327,6 +327,15 @@
 		})
 	}
 
+	// Monaco swallows the keydown, so an editor with focus never reaches the
+	// window handler; Editor/SimpleEditor/TemplateEditor re-broadcast it
+	// (untyped event, hence the manual listener). A step's code editor also
+	// flushes through its `formatAction`, and a redundant flush is a no-op.
+	$effect(() => {
+		window.addEventListener('wm-monaco-save-shortcut', saveDraft)
+		return () => window.removeEventListener('wm-monaco-save-shortcut', saveDraft)
+	})
+
 	// Materialize a brand-new flow's draft before the session preview loads it by
 	// path — an untouched new flow never autosaved, so forcePersist is the only
 	// thing that creates the row. Gated to never-deployed: forcePersist skips the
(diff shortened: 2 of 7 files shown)


rubenfiszel:
Addressed the P1 in 61e1cbb254.

The diagnosis is right: the autosave payload is parked by a `$effect` (`userDraft.svelte.ts:811`), and the Monaco commands dispatched `wm-monaco-save-shortcut` synchronously, so every listener flushed before the effect had run.

Fixed at the dispatch sites rather than in each listener — `Editor.svelte`, `SimpleEditor.svelte` (both registrations) and `TemplateEditor.svelte` now `await tick()` before dispatching. That way the event means "the buffer is materialized and the mirror has parked it", so the invariant holds for every listener including the pre-existing `RawAppEditor` one, and future listeners can't reintroduce the bug. `ScriptBuilder.saveDraft` keeps its own `flushPendingChanges()` + `tick()`, which is still needed for the keydown path where Monaco never ran a command.

Verified end to end in the app editor, which is the path with no `tick()` of its own: created a background runnable (TypeScript inline script), typed `// ZZMARKERZZ` into its Monaco and pressed Ctrl+S, then inspected the request body rather than just its timing.

- POST fires **122 ms after the keystroke** and the body **contains the marker**.

That timing is itself the discriminator. Autosave was idle when I typed, so before the fix `pendingSaveOpts` was empty at flush time: the flush would have no-opped and the marker would only have reached the server on the debounce round (~1.5 s, measured separately on this same page). A 122 ms POST carrying the marker can only happen if the flush ran after the parking effect.

rubenfiszel:
✅ Review round clean @ 61e1cbb254a3f142aa6cd48d52b258e7f4d86ff4

PR #10531, opened 2026-08-05
fix: report python debugger dependency install failures instead of timing out

rubenfiszel:
## Summary

Starting a Python debug session on a cold dependency cache failed with `Debugpy command timeout: launch`, followed by a bare `ModuleNotFoundError`. Those are one bug, not two: the launch waits on `windmill prepare-deps` (up to 120s), every client above it gave up after 10s, and when preparation failed the script was run anyway so the only thing the user ever saw was the missing import.

Three linked defects:

1. `prepare_dependencies()` was a blocking `subprocess.run` called from an async DAP handler. It stalled the event loop for the whole install, which also stalls websocket keepalive.
2. Both `dap_debug_service.ts` and `dapClient.ts` capped every command at 10s, including `launch`. 10s < 120s, so a slow install always lost.
3. Preparation failures returned `None` and the script ran regardless, so the real reason never reached the user.

## Changes

- **Off the event loop**: `prepare_dependencies` now runs via `asyncio.to_thread`, and `handle_launch` awaits it while streaming `Preparing dependencies...` / `Still preparing dependencies... (Ns)` output events every 5s, so a slow launch is visibly working rather than frozen.
- **Per-command timeouts** instead of raising the global 10s: `launch` gets 180s in `dap_debug_service.ts` and in `dapClient.ts`; every other command keeps the 10s default. The timeout value is now in the error message.
- **Failures are reported**, with the distinction that matters:
  - the installer reporting failure (non-zero exit, `success: false`, or a 120s timeout) fails the launch with the reason, instead of running into a `ModuleNotFoundError`;
  - failing to reach the CLI at all (no JSON, unparseable output, missing binary) warns and still runs, because it says nothing about whether the script's imports are satisfied and must not brick a session that would otherwise work.
- **Consumes `PrepareResponse.stderr` if present, degrades if absent** so it can land in either order relative to the sibling branch adding that field. It is read both on failure and on a `success: true` response, because `prepare_python_deps_standalone` currently swallows a non-zero `uv pip install` into `tracing::warn!` and still returns success (`prepare_deps.rs:246`) — which is exactly the unreachable-index case. Until that producer change lands, this surfaces the installer's own error as a warning before the import fails, rather than saying nothing.
- **Connection close now rejects in-flight requests** in both clients. With `launch` at 180s, a debug server dying mid-launch would otherwise leave the caller awaiting for three minutes instead of ten seconds.
- **Bun/TypeScript sessions emit the same progress**, since `dapClient.ts` is shared across languages and that path also blocks the launch response on `prepare-deps` — without this it would sit silent for up to 180s where it used to error at 10s.
- `test_dap_server.py` had the identical 10s cap, which is why its `run_import_test()` could never reproduce this on a cold cache.

## Merge notes

Scoped to stay clear of the sibling branch editing `spawnProcess()` in `dap_debug_service.ts` — the edits here are at lines 492-560 and 805-815 and do not touch it.

## Test plan

Verified with a stub `windmill prepare-deps` driving a real DAP server and the real debug service over websockets:

- [x] **Slow success (12s install)** completes end-to-end, progress streams at 5s/10s, script output arrives. Against the pre-fix service the same run fails at 10.2s with `Debugpy command timeout: launch` — the reported error, reproduced and fixed.
- [x] **Event loop stays free**: websocket ping round-trips in 0.001s during preparation; pre-fix it took 4.0s (the remainder of the blocking call).
- [x] **`success: false` with `stderr`** fails the launch and shows the installer error; **without `stderr`** shows the `error` field alone (both sibling orderings).
- [x] **Non-zero exit / no JSON** reports the stderr rather than a generic failure.
- [x] **`success: true` carrying installer stderr** warns, then still runs the script.
- [x] **Missing `--windmill` binary** warns and still runs a stdlib-only script (no regression for working sessions).
- [x] **Python server killed mid-launch** surfaces `Debugpy connection closed` at 8.8s instead of hanging to 180s.
- [x] **Bun/TypeScript session** streams the same progress and completes a 12s launch.
- [x] `npm run check:fast` clean; `bun build` clean; `python -m py_compile` clean.

### Screenshots

None: the `frontend/` diff is a timeout map plus a promise-rejection on socket close in `dapClient.ts`, with no visual surface of its own. The user-visible strings this PR adds (progress lines, failure reasons) originate in the debugger servers and were exercised end-to-end over the real DAP protocol as above, but not in a browser — no backend or dev server is running for this worktree, and reaching the debug console needs both plus a configured debug-service URL.

---
## Summary by cubic
Fixes Python debugger launches that timed out on cold caches by moving dependency install off the event loop, streaming progress, and surfacing real `uv`/installer errors. `launch` now has a longer per-command timeout so slow installs don’t fail early, and failed launches are cleaned up.

- **Bug Fixes**
  - Run dependency preparation in a worker thread and keep websockets responsive; emit “Preparing dependencies…” and progress every 5s.
  - Use per-command timeouts: `launch` is 180s; others stay 10s. Include the timeout value in error messages.
  - Report installer failures (non‑zero exit, `success: false`, or 120s timeout) and stop the launch with the reason, instead of falling into `ModuleNotFoundError`.
  - If the CLI can’t be reached or output is unparseable/missing, warn and run anyway to avoid blocking valid sessions.
  - Read `stderr` from prepare responses even on success; strip routine `uv` progress lines (including build, download, and warning lines) and surface only meaningful installer errors.
  - Reject pending requests when the DAP or `debugpy` connection closes; on launch failure, send `terminated` immediately and clean up to reap the server and temp dir.
  - Bun/TypeScript sessions stream the same preparation progress and bound prepare‑deps at 120s to prevent indefinite hangs.
  - Test client increases `launch` timeout to 180s to cover cold-cache installs.

<sup>Written for commit 92774211bb7083d86b5a12bbbfe2114029a847c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10531?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/debugger/dap_debug_service.ts
+++ b/debugger/dap_debug_service.ts
@@ -805,6 +805,13 @@ class PythonDebugSession extends BaseDebugSession {
 			this.debugpyWs.onclose = () => {
 				logger.info('Debugpy WebSocket closed')
 				this.debugpyWs = null
+				// A Python server that dies mid-request must fail it now; otherwise the caller
+				// waits out the command timeout, which for `launch` is minutes.
+				const aborted = Array.from(this.pendingDebugpyRequests.values())
+				this.pendingDebugpyRequests.clear()
+				for (const pending of aborted) {
+					pending.reject(new Error('Debugpy connection closed'))
+				}
 			}
 		})
 	}
--- a/debugger/dap_debug_service.ts
+++ b/debugger/dap_debug_service.ts
@@ -489,6 +489,15 @@ abstract class BaseDebugSession {
 // Python Debug Session
 // ============================================================================
 
+const DEFAULT_DEBUGPY_TIMEOUT_MS = 10_000
+
+// `launch` waits on dependency preparation in the Python server, which allows `windmill
+// prepare-deps` up to 120s; anything shorter here reports a timeout while the install is
+// still legitimately running.
+const DEBUGPY_TIMEOUT_MS_BY_COMMAND: Record<string, number> = {
+	launch: 180_000
+}
+
 class PythonDebugSession extends BaseDebugSession {
 	private debugpyWs: WebSocket | null = null
 	private debugpySeq = 1
@@ -531,11 +540,13 @@ class PythonDebugSession extends BaseDebugSession {
 			arguments: args
 		}
 
+		const timeoutMs = DEBUGPY_TIMEOUT_MS_BY_COMMAND[command] ?? DEFAULT_DEBUGPY_TIMEOUT_MS
+
 		return new Promise((resolve, reject) => {
 			const timeout = setTimeout(() => {
 				this.pendingDebugpyRequests.delete(seq)
-				reject(new Error(`Debugpy command timeout: ${command}`))
-			}, 10000)
+				reject(new Error(`Debugpy command timeout: ${command} (after ${timeoutMs}ms)`))
+			}, timeoutMs)
 
 			this.pendingDebugpyRequests.set(seq, {
 				resolve: (value) => {
--- a/debugger/dap_debug_service.ts
+++ b/debugger/dap_debug_service.ts
@@ -1030,7 +1030,13 @@ sys.stdout.flush()
 			})
 		} catch (error) {
 			this.sendEvent('output', { category: 'stderr', output: `Failed to start Python: ${error}\n` })
+			// Claim the terminated event before cleanup kills the process, otherwise the
+			// `exited` handler sends a second one whose empty body erases this error.
+			this.terminatedSent = true
 			this.sendEvent('terminated', { error: String(error) })
+			// A Python server that refused the launch stays in its connection loop, so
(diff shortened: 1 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 92774211bb7083d86b5a12bbbfe2114029a847c6 — nit-only verdicts at b8c20498f861e77cfcede3895c537f95c77d50a4; nits addressed in 92774211bb (uv `Building`/`Built`/`Downloading`/`warning:` lines are install progress, not diagnostics)

PR #10532, opened 2026-08-05
fix: forward TLS trust roots to debug sessions and honor INIT_SCRIPT on windmill_extra

rubenfiszel:
## Summary

Rebased on main, which has since shipped most of what this PR originally carried: #10533 (index settings reach `prepare-deps`, install moved into the service, credentials kept out of the session), #10531 (install failures reported instead of timing out) and #10537 (the `PrepareResult` crash). Where main covers a change, its version is kept — including where it is better than what was here, such as passing registry settings to uv as CLI arguments and racing the read rather than only killing the child on timeout.

What is left is the part of the original report main does not cover: a debug session still cannot make its **own** outbound HTTPS calls through a TLS-intercepting proxy, `bun install` and hosts that spell their CA differently still cannot install, and the `windmill_extra` container still ignores `INIT_SCRIPT`.

## Changes

**Trust roots reach the debug session** (`debugger/env_passthrough.ts`, both session kinds). Main forwards proxy variables to a session; this adds `SSL_CERT_FILE`, `SSL_CERT_DIR`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE` and `NODE_EXTRA_CA_CERTS` alongside them. Installing the CA in the container's system store is not enough on its own: `requests` carries its own bundle and Node reads only `NODE_EXTRA_CA_CERTS`, so without these the install succeeds and then the script's first HTTPS call fails. Registry settings stay out, unchanged from main's reasoning.

`sessionProxyEnv()` moves out of `dap_debug_service.ts` into a module both sessions import, because the **TypeScript/Bun path built its own environment and got none of this** — not even the proxy variables main added. `dap_debug_service.ts` imports from `dap_websocket_server_bun.ts`, so the shared list cannot live in either.

**The venv is built against the interpreter that runs the script** (`prepare_deps.rs`, `dap_debug_service.ts`). `uv venv` was unpinned, so uv picked its own interpreter while the DAP server runs `DAP_PYTHON_PATH`; the resulting site-packages went on a *different* interpreter's `sys.path`. Pure-Python packages still import, which is why this hides, but anything with a compiled extension does not, and the error names the missing extension rather than the mismatch. `prepare-deps` now takes an optional `python_path` and passes it to `uv venv -p`, as the job path already does with `-p … --python-preference only-managed`.

**Host CA spellings are honored** (`prepare_deps.rs`). `INDEX_CERT` only reads `PY_INDEX_CERT`/`PIP_INDEX_CERT`; a host behind a MITM proxy usually configures its CA as `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` for its other tooling. Those are now the fallback chain, and `bun install` gets the bundle as `NODE_EXTRA_CA_CERTS`, the only spelling Bun reads — the environment is cleared, so nothing reached it before.

**`INIT_SCRIPT` on `windmill_extra`** (`docker/entrypoint-extra.sh`). Untouched by main and the other half of the original report: it was silently ignored, so `INIT_SCRIPT=update-ca-certificates` did nothing even though that is the documented pattern for `MODE=worker`. It now runs after `HOME` resolves and before any service starts, aborting with the script's exit code — coming up with an unprepared trust store instead fails every TLS handshake later, which is much harder to diagnose.

**One-word fix on main's timeout path**: `proc.kill()` becomes `proc.kill('SIGKILL')`. Measured directly — `prepare-deps` does not act on SIGTERM while uv is running, so the timeout left the installer going after the session gave up.

## Test plan

Verified against a fixture PEP 503 index served over HTTPS with a private CA, holding only `pip` and a package that does not exist on PyPI, driving the real service with a credential-bearing `PIP_INDEX_URL`.

- [x] `cargo check -p windmill-worker` and `cargo build --bin windmill` clean; `bun build` on both entrypoints; `py_compile`; `bash -n`.
- [x] **Both languages get the trust roots and no credentials.** Python and TypeScript sessions that dump their own environment see all five runtime variables; neither sees a registry variable or an unrelated variable planted in the service environment. The Python session additionally walks `__main__`'s globals and reads `/proc/self/environ` for the credential — both clean.
- [x] **The install still works through that split**: a script importing the fixture package returns `{"imported": "came-from-private-index"}`.
- [x] **The CA fallback is what carries it.** Cold cache, real binary, `PY_INDEX_CERT` unset: with no CA variable the install fails `Failed to install seed packages … Request failed after 3 retries`; with only `REQUESTS_CA_BUNDLE` it succeeds. Same for `CURL_CA_BUNDLE`.
- [x] **Compiled extensions import.** Reproduced the mismatch first: a 3.12 venv's site-packages on a 3.14 interpreter imports `msgpack` (it has a pure-Python fallback) but fails `xxhash` with `ModuleNotFoundError: No module named 'xxhash._xxhash'`. After the pin, a debug session running `import xxhash` returns a real digest and the venv is `python3.14`, matching the server.
- [x] **Disconnect reaps the installer.** Against a TCP tarpit that accepts and never replies, hanging up mid-install leaves no `prepare-deps` process (scan by `/proc/<pid>/exe`).
- [x] `docker/entrypoint-extra.sh`: `INIT_SCRIPT` runs before the service banner and its side effect lands; a script exiting 7 aborts with code 7 and starts no service; unset and empty remain no-ops.
- [x] uv semantics confirmed directly rather than assumed: `SSL_CERT_FILE` is honored *without* native TLS and *replaces* uv's roots (a private-CA-only bundle makes PyPI untrusted), native TLS is what adds the system store, and `--cert` exists on `uv pip compile` but not `pip install`/`venv` (it is hidden from `--help` on all of them).

No test is shipped: the remaining logic is one env allowlist and one fallback chain, both of which the scenario above exercises end to end.

Round-7 review findings are folded in: `SSL_CERT_DIR` now reaches uv (it is additive, unlike the bundle), the SIGKILL fix covers the Bun installer too, the Python installer is killed on disconnect, and the README states that the CA bundle replaces uv's roots so it must be complete.

## Not in scope

`prepare-deps` runs outside nsjail in both language paths (raw `spawn`, pre-existing since #7546), and killing the installer does not reap a `uv` grandchild it already spawned. Both predate this PR and are tracked separately.

A **Bun/TypeScript session never cleans up its temp directory**: after a session ends, `/tmp/windmill_debug_*` keeps `script.ts` and the `node_modules` symlink. Since `script.ts` survives, `unlink` does not run either, so the session's `cleanup()` is not reached at all on that path rather than merely failing on a non-empty `rmdir`. Pre-existing and unchanged by this PR (the Python path's equivalent is fixed above), but it wants its own investigation.


rubenfiszel:
Re the Pi review at head `144400cd38`: three of its four P1s (and the P2) describe code that is not what that commit contains. The mechanisms it reports as absent are all present at the reviewed head:

```
$ git show 144400cd38:debugger/dap_debug_service.ts | grep -nE "spawnProcess\(\{|PREPARE_DEPS_TIMEOUT_MS|disposed|SIGKILL|killPrepareDeps"
154:const PREPARE_DEPS_TIMEOUT_MS = 120_000
519:	private disposed = false
545:			this.prepareDepsProcess = spawnProcess({
556:				this.killPrepareDeps()
557:			}, PREPARE_DEPS_TIMEOUT_MS)
570:				logger.error(`prepare-deps timed out after ${PREPARE_DEPS_TIMEOUT_MS}ms`)
595:	private killPrepareDeps(): void {
600:				this.prepareDepsProcess.kill('SIGKILL')
1056:			if (this.disposed) {
1144:		this.disposed = true
1152:		this.killPrepareDeps()
```

The Codex and Claude reviews of the same head both quote `PREPARE_DEPS_TIMEOUT_MS`, so this looks like a stale read of the file rather than a real regression. These were the exact findings from the previous Codex round and they were fixed in `144400cd38`, verified by running it:

- **nsjail**: with nsjail enabled, the log line is `Spawning with nsjail: nsjail --config ... -- windmill prepare-deps`, and the session still installs from a private TLS index and runs.
- **Cancellation**: against a TCP tarpit that accepts and never replies, disconnecting mid-install kills the installer within 2ms and logs `Session was torn down during dependency preparation, not starting Python`. A scan by `/proc/<pid>/exe` afterwards finds no orphaned installer and no orphaned DAP server. SIGTERM was measured to leave it running, which is why cancellation uses SIGKILL.

**The fourth P1 is a fair catch and I am not fixing it here.** `DebugSession.prepareDependencies()` in `dap_websocket_server_bun.ts` does use a raw `spawn()` with no nsjail wrapping, no timeout and a local handle. That is pre-existing (unchanged since #7546, `git log -S` confirms) and not a regression from this PR, which touches that file only to spread `runtimeEnv()` into the debuggee. Fixing it means extracting the inline nsjail wrapper at lines 1669-1688 and adding timeout plus handle-ownership to the Bun session — a self-contained change that deserves its own PR rather than riding along here. Noting it explicitly so it is on record rather than silently dropped.

The two comment nits Codex and Claude raised (the timeout provenance, and a comment narrating the refactor) are fixed in `086c11ac30`.

rubenfiszel:
✅ Review round clean @ 086c11ac304a1b2581615b7917bd4dbd4620d10c

rubenfiszel:
✅ Review round clean @ c7b29b80d936ad19a7017010db49b0096942cc7c

PR #10533, opened 2026-08-05
fix: pass python index settings to prepare-deps and report failures

rubenfiszel:
## Summary

The debugger's dependency preparation (`windmill prepare-deps`) ran `uv` with only `PATH`, `HOME`, `UV_CACHE_DIR` and the proxy vars, passing no index, trusted-host, certificate or TLS arguments at all. Behind a private PyPI mirror or a TLS-intercepting proxy the Python debugger was therefore unusable, with no way to configure around it: this process has no database, so the `pip_index_url` / `pip_extra_index_url` instance settings the job path resolves are unreachable.

Two things also hid the failure. `uv pip install` errors were only logged as a warning while the response still claimed `success: true` with a venv path, so the caller ran the script against an empty venv and the user saw a bare `ModuleNotFoundError`. And the debug service replaces the child environment when spawning a session (`spawnProcess`: "Don't inherit all of process.env"), so registry settings never reached the CLI in the first place.

## Changes

**Backend**

- `prepare_deps.rs`: both uv invocations now receive `--extra-index-url`, `--index-url`, `--trusted-host` and `--native-tls`, plus `SSL_CERT_FILE`, `UV_HTTP_TIMEOUT` and `UV_INDEX_STRATEGY` in the environment, mirroring what `python_executor.rs` does for real jobs. `uv venv --seed` resolves pip from the index, so the venv step needs them too, otherwise a mirror-only network fails before any package is installed. uv has no `--cert` on `venv`/`pip install`, so the CA bundle goes through `SSL_CERT_FILE` exactly as the job path does.
- Values come from env only: `PY_INDEX_URL` / `PIP_INDEX_URL`, `PY_EXTRA_INDEX_URL` / `PIP_EXTRA_INDEX_URL` (comma-separated), and the existing `TRUSTED_HOST` (whitespace-separated), `INDEX_CERT`, `NATIVE_CERT` and `UV_HTTP_TIMEOUT` statics, which moved from `python_executor.rs` to `worker.rs`. They lived behind the `python` cargo feature while `prepare_deps` is ungated, so sharing them required the move; the parsing rules are unchanged and now have one definition. A variable that is set but empty no longer shadows its fallback name, which also fixes `PY_INDEX_CERT=""` exporting `SSL_CERT_FILE=""` to uv on the job path.
- A failed `uv pip install` now returns `success: false`. uv installs the set atomically, so the outcome was already an empty venv; the difference is that the reason is reported instead of discarded.
- `PrepareResponse` gains `install_stderr` carrying the installer's raw stderr, `skip_serializing_if = "Option::is_none"` so the JSON is byte-identical to before on success. `error` keeps the same full message it has always carried, so a caller that reads only `success`/`error` is unaffected.

**Debugger**

- Dependency installation for a Python session moves from `dap_websocket_server.py` into `dap_debug_service.ts`, which passes the resulting venv to the server with `--venv-path`. The registry settings routinely embed credentials, and the Python server executes the submitted script inside its own interpreter, so anything that process holds is recoverable by user code (`os.environ`, module globals, `/proc/self/environ`). The service never executes user code, so the credentials stop there. The server keeps its own `prepare-deps` path for standalone use when `--venv-path` is not given.
- The service reports a failed install to the client as an `output` event, so the reason reaches the user instead of a later `ModuleNotFoundError`.
- Proxy variables are forwarded into each session, since the debugged script needs them for its own outbound calls exactly as a job's script does on a worker. Both the upper and lower case spellings are forwarded, because the worker reads the lowercase names first.
- `debugger/README.md`: documents the variables, where to set them, and the failure reporting.

## Note on the sibling change

A separate branch teaches `dap_websocket_server.py` to surface `install_stderr` from the CLI response. That work still applies to the server's standalone path; in the service-driven path this PR reports the same information from the service. The response field is additive and the old shape still works on its own.

## Test plan

- [x] `cargo check` and `cargo check --features python`
- [x] `cargo test -p windmill-worker --lib prepare_deps` (wire-shape regression test)
- [x] `windmill prepare-deps` run directly: success response byte-identical to before; a nonexistent package returns `success: false` with uv's resolution error in `error` and `install_stderr`; an unreachable index surfaces the connection failure; comma-separated extra indexes and whitespace-separated trusted hosts both expand correctly
- [x] A `uv` wrapper capturing argv confirmed `uv venv --seed` and `uv pip install` each receive the full flag set with `SSL_CERT_FILE`, `UV_HTTP_TIMEOUT` and `UV_INDEX_STRATEGY` in the environment
- [x] An empty `PY_INDEX_URL` falls back to `PIP_INDEX_URL` instead of disabling it
- [x] Full debug session against a running `dap_debug_service.ts` with the registry settings and a marker secret set on the service only. The debugged script imported a package installed for it (`tabulate 0.10.0`) and found nothing through `os.environ`, `sys.modules['__main__']` or `/proc/self/environ`: `{"os_environ":[],"main_module_attrs":[],"proc_environ_hits":[],"proc_environ_has_secret":false}`
- [x] Same session with an unresolvable package: the client receives uv's full error as a stderr `output` event
- [x] The bun/TypeScript path still returns its previous response shape and its autoinstall test passes

### DIFF
--- a/backend/windmill-worker/src/prepare_deps.rs
+++ b/backend/windmill-worker/src/prepare_deps.rs
@@ -12,6 +12,7 @@ use regex::Regex;
 use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 
+use crate::worker::non_empty_env;
 use crate::{
     BUN_CACHE_DIR, BUN_PATH, HOME_ENV, INDEX_CERT, NATIVE_CERT, PATH_ENV, PROXY_ENVS, TRUSTED_HOST,
     UV_CACHE_DIR, UV_HTTP_TIMEOUT,
@@ -94,17 +95,11 @@ lazy_static::lazy_static! {
     /// This process has no database, so the `pip_index_url` / `pip_extra_index_url` instance
     /// settings the job path resolves are unreachable here: their env-var equivalents are the
     /// only registry configuration the debugger can see.
-    static ref PY_INDEX_URL: Option<String> = non_empty_var("PY_INDEX_URL").or_else(|| non_empty_var("PIP_INDEX_URL"));
-    static ref PY_EXTRA_INDEX_URL: Option<String> = non_empty_var("PY_EXTRA_INDEX_URL").or_else(|| non_empty_var("PIP_EXTRA_INDEX_URL"));
+    static ref PY_INDEX_URL: Option<String> = non_empty_env("PY_INDEX_URL").or_else(|| non_empty_env("PIP_INDEX_URL"));
+    static ref PY_EXTRA_INDEX_URL: Option<String> = non_empty_env("PY_EXTRA_INDEX_URL").or_else(|| non_empty_env("PIP_EXTRA_INDEX_URL"));
     /// uv defaults to `first-index`; the job path overrides it so a package missing from the
     /// first index is still resolved from the others. Same default here.
-    static ref PY_INDEX_STRATEGY: String = non_empty_var("UV_INDEX_STRATEGY").unwrap_or_else(|| "unsafe-best-match".to_string());
-}
-
-/// An env var declared but left empty (a common shape in compose/k8s manifests) must not
-/// shadow the fallback name it is checked against.
-fn non_empty_var(key: &str) -> Option<String> {
-    std::env::var(key).ok().filter(|v| !v.is_empty())
+    static ref PY_INDEX_STRATEGY: String = non_empty_env("UV_INDEX_STRATEGY").unwrap_or_else(|| "unsafe-best-match".to_string());
 }
 
 /// Simple loader that doesn't require Windmill API for relative imports
--- a/backend/windmill-worker/src/prepare_deps.rs
+++ b/backend/windmill-worker/src/prepare_deps.rs
@@ -12,7 +12,10 @@ use regex::Regex;
 use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 
-use crate::{BUN_CACHE_DIR, BUN_PATH, HOME_ENV, PATH_ENV, PROXY_ENVS, UV_CACHE_DIR};
+use crate::{
+    BUN_CACHE_DIR, BUN_PATH, HOME_ENV, INDEX_CERT, NATIVE_CERT, PATH_ENV, PROXY_ENVS, TRUSTED_HOST,
+    UV_CACHE_DIR, UV_HTTP_TIMEOUT,
+};
 use windmill_common::worker::write_file;
 
 const LOADER_BUILDER_CONTENT: &str = include_str!("../loader_builder.bun.js");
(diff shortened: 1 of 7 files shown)


rubenfiszel:
On the Codex P1 from the last round (reading the service's environment through `/proc` from a same-UID session), I've documented the boundary rather than changing the isolation model, and want to be explicit about why.

The finding is accurate: without `--nsjail`, a debug session runs as the same user as the service and can read `/proc/<ppid>/environ`. Two things bound how much this PR can do about it:

- It is the same property a worker has. A job's script never gets `PIP_INDEX_URL` in its own environment, but on an unsandboxed worker it can read the worker's `/proc/<pid>/environ` and find it there. Windmill's answer in both places is the sandbox: the shipped `nsjail.debug.config.proto` gives the session its own PID namespace with `mount_proc: true`, so the service isn't reachable from inside it.
- Closing it without a sandbox means changing how sessions are executed (a dedicated UID or an enforced namespace), or introducing a secrets-file config path so the credentials never sit in any process environment. Both are deployment-model changes for the debugger, well outside a change that makes `prepare-deps` honor index settings, and neither is something I should decide unilaterally.

What this PR does guarantee is the part that was in its control: the interpreter that executes user code never receives the settings, verified against `os.environ`, `sys.modules['__main__']` and `/proc/self/environ` with a marker secret set on the service. `debugger/README.md` now states the residual limitation and points at `--nsjail`.

The P2 is fixed in 5674b11: `sessionProxyEnv()` defaults `NO_PROXY` to `localhost,127.0.0.1` when a proxy is configured without a bypass list, matching `PROXY_ENVS`. Verified end to end - a session started with only `HTTPS_PROXY` set now sees `NO_PROXY=localhost,127.0.0.1`.

Also in 5674b11, from the Claude round: the 120s bound on the install is restored (`DAP_PREPARE_DEPS_TIMEOUT_MS`, configurable for slow mirrors). The first attempt at it did not work - killing the child leaves a grandchild holding the pipe, so the read stayed pending; the deadline now races the read. Confirmed with a deliberately hanging binary: the client gets a timeout message and the session starts without the dependencies instead of hanging. The unjailed-install trade-off Claude raised is now stated in the doc comment.

rubenfiszel:
Round 4 nits: the README one is fixed in cf412a6 (it now names `--nsjail --nsjail-config nsjail.debug.config.proto` and says the isolation comes from that config's PID namespace and `mount_proc`, not from the flag).

On the timeout killing only the CLI and not a stuck `uv` underneath it: correct, and I'm leaving it. Killing the whole group means spawning through `setsid` and signalling the negative pgid, which adds a hard dependency on `setsid` being present in the debugger image; if it isn't, every dependency install fails instead of just the rare stuck one. The leak it would prevent is also self-limiting rather than unbounded: `uv` has its own HTTP timeouts and retry budget (`UV_HTTP_TIMEOUT`, forwarded by this PR) and exits on its own, at which point the pipes close and the floating `read` resolves. The temp venv directory is not new either: `prepare-deps` creates `/tmp/windmill-deps/<uuid>` per call and nothing has ever cleaned those up, so a timed-out install leaves exactly what a successful one does.

If the group kill is wanted, it belongs with a cleanup story for `/tmp/windmill-deps` as a whole rather than bolted onto this path.

PR #10534, opened 2026-08-05
feat: allow a dev workspace to have its own dev workspace

rubenfiszel:
## Summary

Fixes WIN-2324.

A dev workspace could not be given a dev workspace of its own: the backend rejected any parent
that was itself a fork or dev, and the settings tab replaced its whole content with the "this is a
dev workspace" panel, so there was no form to attach or create one. This allows the nested shape
(`prod -> dev -> staging`) even though it is not the recommended pattern, and makes the UI stop
pointing past the workspace actually being worked in when you are inside one.

The environment label doubles as the deploy branch, and dev workspaces in a chain inherit the same
git-sync repositories, so the new guard is that **every dev workspace in a chain carries a distinct
label**. With only `dev` and `staging` available that caps a chain at two dev workspaces, which is
exactly the "dev of a dev" the issue asks for.

## Changes

**Backend** (`windmill-api-workspaces/src/workspaces.rs`)

- `ensure_dev_parent_is_root` → `ensure_dev_parent_can_host_dev`: a dev workspace may host a dev;
  only a throwaway fork still can't (its deploys go to its own `wm-fork/**` branch and it is
  discarded with its subtree).
- `attach_dev_workspace`: prod is no longer required to be a root. Since the candidate can now sit
  *above* prod, an explicit recursive ancestor check rejects a pairing that would close a
  parent↔child cycle. Dropped the guard that refused a candidate which already owned a dev — that
  is the nested chain this PR enables.
- New `reject_dev_label_taken_in_chain`, applied on create and attach: collects the dev workspaces
  in the resulting chain (the dev ancestors the new one lands under, the new one itself, and — on
  attach — the dev workspaces the candidate brings with it) and rejects any duplicate label.
- Teardown may not strand a nested dev (`reject_stranding_nested_dev`): its parent would end up in a
  shape that hosts no pairing, with no detach control in the UI and `delete_workspace` refusing it
  while the child lives. Detach only trips on a `wm-fork-` workspace (it keeps its parent and reads
  as a throwaway fork again; a prefix-less one returns to standalone and goes on hosting its dev),
  while archive trips on any dev workspace, since it soft-deletes whatever the id looks like.
- Label checks run after the shape checks so a cyclic attach is reported as a cycle rather than as a
  workspace clashing with itself.

**Frontend**

- `findWorkspaceRoot` stops climbing once it crosses a dev→dev edge, so a dev of a dev (and anything
  forked below it) treats its dev parent as the family head. Every root-scoped affordance follows:
  the sidebar scope picker, the scope chip, the fork-base list, the session workspace picker. The
  existing "stop at the first ancestor you can't see" behaviour is unchanged — that already covered
  the no-access-to-the-root half of the issue.
- `DevWorkspaceSetting.svelte` renders the parent pairing **and** this workspace's own dev section
  for a dev workspace, instead of only the former. The two panels read protection rules
  independently (extracted into a `useProtectionRules` factory, instantiated per side). The parent
  panel names the parent by what it is ("paired with dev workspace X", "Go to dev workspace")
  instead of always calling it the root.
- The attach form filters ancestors out of the candidate list (they would close a cycle), offers
  only the labels no dev workspace in the chain holds, and disables the attach when the candidate's
  own dev workspaces already clash with the chain above.
- `CreateWorkspaceInner.svelte` offers dev designation off a dev base too, with the same label
  narrowing.
- The deploy target and its item filters (`DeployToSetting`) are now rendered by
  `DevWorkspaceSetting` directly under the pairing line they restate, above the parent's protections
  panel — rather than after the whole component, where on a dev of a dev they read as belonging to
  the nested pairing below.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test nested_dev_workspace` — new integration
      tests pin the attach guards (cycle, label reuse from above, label reuse from the candidate's
      own subtree, and the nested attach succeeding) and the teardown-stranding refusals (detach and
      archive of a `wm-fork-` dev, archive of a prefix-less one, and the bottom-up order working).
- [x] `npx vitest run src/lib/utils/workspaceHierarchy.test.ts` — 8 passed, covering the dev-of-dev
      root boundary, `findWorkspaceAncestors` walking past it, and `devWorkspacesInChainAbove`.
- [x] `npm run check` — 0 errors (warning count unchanged at 65).
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features` and
      `cargo check -p windmill-api-integration-tests --all-targets`.
- [x] Manually built `prod -> prod-dev (dev) -> stgws (staging) -> wm-fork-leaf` on a local EE
      backend through the settings tab, as a superadmin and as a user who is only a member from
      `prod-dev` down. Confirmed live rejections: attaching the family root (cycle), reusing the
      parent's label, reusing a grandparent's label at the third level, attaching a candidate whose
      own dev already holds the label, and detaching or archiving a dev that still owns one.

## Screenshots

Chain used throughout: `prod → prod-dev (dev) → stgws (staging) → wm-fork-leaf`.

### Dev workspace settings, from the middle of the chain (`prod-dev`)

Before — nothing but the parent pairing; no way to give it a dev of its own:

before

After — the pairing, its deploy target and item filters, then the parent's protections:

after top

…and below it, this workspace's own nested dev workspace with its own protections panel:

after lower

### Dev workspace settings, from a dev of a dev (`stgws`)

Before — calls `prod-dev` a "root workspace" and offers "Go to root workspace":

before

After — names the parent correctly, and explains why the chain stops here (both labels are taken):

after

### Sidebar scope picker, in the leaf workspace, with access to everything

Before — the family head is the far `prod`, and an unrelated fork of it is listed alongside:

before

After — the family head is `prod-dev`, the workspace this subtree actually promotes into:

after

### Sidebar scope picker, in the leaf workspace, as a user with no access to `prod`

Unchanged — `findWorkspaceRoot` already stopped at the highest reachable ancestor, so this half of
the issue was already handled. Included to show the change does not regress it.

before
after

### Workspace selection page

Unchanged (it groups by true lineage, which is the right thing for a directory of every workspace
you belong to), and it already rendered the nested chain. Superadmin, then the member-from-the-
middle user:

all
mid

## Known limitations

Auto-pull does not reach a nested dev workspace: `resolve_fork_branch_target` routes an
environment-label branch to a *direct* dev child of the repo's workspace, and a nested dev is
parent-managed by a dev workspace that holds no auto-pull config of its own. The comment there now
says so rather than claiming dev workspaces cannot nest.


rubenfiszel:
✅ Review round clean @ d771de34b6c5514320a8525f85bf0a96f3542227 — nit-only verdicts at 22e19abd984152123244fad34bed11f58f64e40c (codex: fixture header said two families, defines three; claude and pi: good to merge); nit addressed in d771de34b6c5514320a8525f85bf0a96f3542227

rubenfiszel:
✅ Review round clean @ 9e6e68f425adeefb41e24b42f8bc004f6feeb20c

rubenfiszel:
✅ Review round clean @ 16c7ccc47ce11907154fe1e1862a99af94e51992

PR #10538, opened 2026-08-05
fix: point re-opened previews at the tab already showing them

AlexRV12:
## Summary

In AI sessions, re-opening something that is already in the preview panel gave no feedback. Clicking an artifact chip flashed the preview border to say "it's already here, look" — but clicking the Preview chip for an app, script, flow or page did nothing visible at all, so the click read as dead.

The pulse itself was firing for those cases. The flash ring simply did not exist for them: it was markup nested **inside** the `artifact` branch of `PreviewTabHost`, so an editor or iframe tab had nothing to light up.

Working on that surfaced a second, related bug: a run re-opened from the jobs tray spawned a duplicate tab instead of focusing the one already showing it.

## Changes

- **Render the flash ring for every tab kind.** Hoisted it out of the artifact branch to a sibling of every tab body. It has to be a sibling rather than a child — an editor's opaque background (or an iframe's document) paints over a ring drawn inside it.
- **Move the "should this flash?" decision into the model.** `SessionPreviewTabs` now snapshots the displayed tab around `open()` and `navigate()` and pulses only when the mutation left the panel showing exactly what it already showed. Two hand-rolled copies of that rule (artifacts in `sessionRuntime`, the tool-chip handler in `+page.svelte`) are gone, so every re-open path is covered: artifact chips, the chat's Preview chip, the `open_preview` / `open_page` tools, the jobs tray, the tab picker, breadcrumb picks and in-editor links.
- **`asOneChange(fn)`** for callers that reach a destination through their own sequence. `open_page`'s reuse branch does `select` → `navigate` → reveal; judged per call, the switch to a background tab would be followed by a `navigate` that finds nothing left to change and flashes a tab the user had just been moved to.
- **Fix duplicate run tabs.** Callers bake `?workspace=` into a run href, while a tab's observed `loc` has it stripped by `canonicalizeObservedLoc` — so the dedupe compared `/run/<id>?workspace=x` against `/run/<id>` and never matched. Both sides are now canonicalized in `open()`. Both trigger paths (jobs tray, in-iframe run click) bake the param in, which is why this belongs at the comparison rather than the call sites.
- **`retargetTab` no longer wipes a tab's editor-stamped label when the destination is unchanged.** Nothing re-stamps a tab that never left its item, so the wipe was permanent — a draft's label fell back to its `…/draft_<uuid>` storage path.

## Screenshots

Re-opening an app preview that is **already displayed**, via the tab-strip picker. `open()` returns `focused`, nothing about the panel changes, so the border flashes to say "it's already here". This is the case that could not flash at all before: the ring lived inside the `artifact` branch, so an app tab had nothing to light up.

| before / after the flash | mid-flash |
|---|---|
| 32-after | 31-flash |

Captured from a real pulse (the ring's class was observed flipping `opacity-100` → `opacity-0`), not a forced style.

## Test plan

- [ ] With a script, flow or app displayed in the preview, re-open that same item (chat Preview chip, `open_preview`, or the tab picker): the tab stays put and its border flashes once, with no second tab.
- [ ] Same for an artifact and for a plain page tab (Runs/Schedules).
- [ ] Open a run from the jobs tray twice: one tab, flashing on the second click, and the run page inside it scoped to the session's workspace.
- [ ] With two tabs open, re-open the one sitting in the background: it becomes active **without** flashing — the switch is itself the visible change.
- [ ] Collapse the panel, then re-open the tab it was showing: it reveals without flashing.
- [ ] Re-open via `open_page` a Runs tab that is open but not focused: switches to it, does not flash.
- [ ] Rename an unsaved draft, then re-pick it in the breadcrumb picker: the tab keeps its name instead of reverting to `…/draft_<uuid>`.

### DIFF
--- a/frontend/src/lib/components/sessions/PreviewTabHost.svelte
+++ b/frontend/src/lib/components/sessions/PreviewTabHost.svelte
@@ -143,8 +143,10 @@
 
 	let flashing = $state(false)
 	let flashTimer: ReturnType<typeof setTimeout> | undefined
-	// Guard against the effect's non-pulse reruns (tab/runtime changes) firing a flash.
-	let lastPulseNonce = -1
+	// Guard against the effect's non-pulse reruns (tab/runtime changes) firing a
+	// flash. Seeded from the current nonce: a pulse from before this host mounted
+	// is moot, the tab appearing is itself the change the flash would point at.
+	let lastPulseNonce = runtime?.previewTabs.focusPulse.nonce ?? -1
 	$effect(() => {
 		const pulse = runtime?.previewTabs.focusPulse
 		if (!pulse || pulse.nonce === lastPulseNonce) return
@@ -242,13 +244,6 @@
 		{:else if !runtime?.manager.artifacts.loading}
 			<div class="p-4 text-sm text-tertiary">This artifact is no longer available.</div>
 		{/if}
-		<!-- Overlay: the source editor's opaque bg would cover a ring on the container. -->
-		<div
-			class="pointer-events-none absolute inset-0 z-30 ring-2 ring-inset ring-border-accent transition-opacity duration-300 {flashing
-				? 'opacity-100'
-				: 'opacity-0'}"
-			aria-hidden="true"
-		></div>
 	</div>
 {:else if mounted}
 	<iframe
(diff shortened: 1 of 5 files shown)


AlexRV12:
✅ Review round clean @ c13088ab625cf8a19bd70eb9bb6ba6ea2e800e16 — nit-only verdicts at f3a0ba6976b1415ad2e5a3a984bcd682b0b6f11a (codex + claude "mergeable, nits", pi "good to merge"); both nits addressed in c13088ab62.

- **codex** — fullscreen-overridden collapsed tabs missed the pulse. Confirmed reachable: `fullscreen` is page state (`+page.svelte:320`) and is not reset on session switch, while `collapsed` is per-session and persisted, so a session whose preview is persisted collapsed renders anyway under fullscreen. The model now takes `setFullscreen()` and treats collapsed-but-fullscreen as displayed; regression test added.
- **claude** — `asOneChange` is synchronous-only (the verdict is read as soon as `mutate` returns). Stated in the doc comment. Also trimmed the `PreviewTabHost` ring comment down to the part that carries a constraint.

PR #10540, opened 2026-08-05
fix: translate folder principals when the compare page copies a folder

Guilhem-lm:
## Summary

Follow-up to #10354, which fixed how a folder is copied between workspaces on the one path it touched (`createFolderIfAbsent`). The **compare page** copies folders through the shared `deployItem`, whose folder branch built both the create and the update body from `owners`, `extra_perms` and `summary` alone — so it still had every defect that PR fixed.

Two consequences:

1. **`default_permissioned_as` and `labels` never crossed.** A copied folder applies no create-time identity rules, so an item deployed into it carrying no `on_behalf_of` of its own runs as whoever deployed it rather than as the principal the source folder would have chosen.
2. **`u/<username>` principals crossed verbatim**, and a username is per-(workspace, email). The same username in the target can be a different account, which then receives folder ownership or ACL access and can be selected by an identity rule. A CI reviewer rated this P1 when the identical code appeared in #10354.

Principals now resolve **source username → email → target username**, email being the only identifier stable across workspaces. Groups match by name, against two different sets depending on where the principal sits:

- an **identity rule** resolves against the workspace's own `group_` rows (`listGroups`), which is all `ensure_permissioned_as_exists` accepts;
- an **owner or ACL entry** resolves against `group_ ∪ instance_group` (`listGroupNames`), because `get_groups_for_user` folds instance groups into the `authed.groups` that `extra_perms` is matched against — an instance group is instance-wide and grants access in the target exactly as it did in the source.

Collapsing those two into one set (the first round of this PR did) either drops an instance-group owner on every deploy or writes a folder whose rule the server then refuses every item deploy into. Both halves are verified below.

The two kinds of unresolvable principal are separated because they fail differently:

- an **owner or ACL entry** is dropped and reported — that can only narrow the folder, and `create_folder` force-appends the caller to `owners`, so nobody is locked out;
- an **identity rule** refuses the copy. Dropping it runs items as the deployer; carrying it verbatim creates a folder the server then rejects every deploy into, because `create_folder` validates a rule's *shape* while `ensure_permissioned_as_exists` validates its principal's *existence* at item-create time.

### Why the fix lands twice

The shared package is the real home and fixing it also fixes `wmill workspace merge`. But the frontend consumes the **published** `windmill-utils-internal` (1.8.2), so the compare page cannot benefit until that version ships and the dependency is bumped. Until then it needs the mirror in `utils_workspace_deploy.ts` — the same arrangement #10354 used for `makeProvider`. **Follow-up for someone with npm credentials: publish the package, bump `frontend/package.json`, and delete `deployFolder` / `translatedFolderBody` / the folder branch of `getItemValue` from the frontend.** There is no publish workflow for this package (`npm_on_release.yml` only publishes the CLI and typescript-client), so it is a manual step.

### `perPage` documentation

Every paginated endpoint documented "default 30, max 100" while `paginate()` uses `DEFAULT_PER_PAGE = 1000` / `MAX_PER_PAGE = 10000`. That wrong comment is exactly what made the merged PR read groups a page at a time and miss them (a P1 there). Corrected on the shared `PerPage` parameter, including the caveat that the script and job listings use `paginate_without_limits` (default 10000, no maximum), and the two tracked artifacts derived from the spec regenerated with `backend/generate_mcp_endpoints_tools/generate_mcp_tools.py`.

## Changes

- `cli/windmill-utils-internal/src/deploy.ts` — folder branch translates every principal and carries `default_permissioned_as` + `labels`; new `principalTranslator` / `workspaceGroupNames` helpers; `listUsers` / `listGroups` added to `DeployProvider`; `droppedAccess` added to `DeployResult`; folder `getItemValue` surfaces the two new fields so the diff drawer shows what the deploy rewrites
- `cli/src/commands/workspace/merge.ts` — provider wiring, and dropped access reported per item
- `frontend/src/lib/utils_workspace_deploy.ts` — `deployFolder` mirror behind `deployItem`; `translatedFolderBody` extracted so `createFolderIfAbsent` and the new path cannot drift; folder `getItemValue` mirrored
- `frontend/src/lib/components/CompareWorkspaces.svelte`, `DeployWorkspace.svelte` — warn when a folder lands with less access than its source
- `backend/windmill-api/openapi.yaml` + regenerated `mcpEndpointTools.ts` / `auto_generated_endpoints.rs` — correct `perPage` description
- `cli/test/deploy_folder_principals_unit.test.ts` — pins the translation, the drop-and-report, and that an unresolvable rule writes nothing

## Verification

Verified against a real EE backend and Postgres, not by reading code. Scratch workspaces `scratchfoldersrc` → fork `wm-fork-scratchfolderdev`, with `alice@…` in both under **different** usernames (`alice_src` / `alice_dev`), `bob_src` source-only, group `scratchboth` in both and `scratchonlysrc` source-only. All fixtures deleted afterwards.

**Pre-change code** (fix stashed, same deploy replayed) — every defect reproduced:

```
owners  {u/alice_src,u/bob_src,g/scratchboth,u/admin}
extra_perms  {"u/bob_src": true, "u/alice_src": true, "g/scratchonlysrc": true, ...}
default_permissioned_as  []
labels  NULL
```

**Post-change, create path:**

```
owners  {u/alice_dev,g/scratchboth,u/admin}
extra_perms  {"u/alice_dev": true, "g/scratchboth": true, "u/admin": true}
default_permissioned_as  [{"path_glob":"jobs/**","permissioned_as":"u/alice_dev"},
                          {"path_glob":"**","permissioned_as":"g/scratchboth"}]
labels  {scratch,translated}
```

- **Update path**: re-deploying an existing folder translates identically, and *repaired* the folder the pre-change code had mis-copied.
- **Refusal**: a folder whose rule names `u/bob_src` was refused on both the create and the update path and never written to the target.
- **`wmill workspace merge`** produced byte-identical results, exercising the shared package directly rather than the mirror.
- **Instance groups** (second round): a folder owned by `g/<instance group>` keeps that owner and ACL entry across a deploy, while a folder whose *rule* names one is refused. The asymmetry is real, not theoretical — creating a script in such a folder returns `Folder 'igrule' default_permissioned_as rule '**' resolves to group 'g/scratch_igroup' which does not exist in this workspace`.
- **Deploy drawer** (second round): folder dependency deployed from the staging/prod drawer warns on dropped access and fails without writing when its rule cannot be resolved.

Checks: `npm run check` 0 errors · CLI `tsc` no new errors (18 pre-existing, baseline-diffed) · `cargo check -p windmill-api --features mcp` clean · CLI unit tests 1050 pass · `utils_workspace_deploy.test.ts` 12 pass.

## Screenshots

Compare & Deploy — a folder deployed with unresolvable access warns rather than narrowing silently, and one whose identity rule cannot be resolved fails outright:

folder-deploy-toasts

The folder diff drawer now shows `default_permissioned_as` and `labels` — previously it listed neither, so a reviewer could not see that the deploy was about to rewrite the folder's identity rules:

folder-diff-drawer

_(Both captured before a reviewer-prompted wording tweak to the refusal message — it now distinguishes a principal absent from the target from one whose account has been removed from the source, since "add them to the target" is useless advice for the latter.)_

## Test plan

- [ ] Compare page, parent → fork: deploy a folder whose `default_permissioned_as` names a user present in both workspaces under different usernames; confirm the fork's row carries the target's username, plus `labels`
- [ ] Same folder with an owner and an ACL entry that have no account in the target: deploy succeeds, warning names each dropped principal once
- [ ] Folder whose identity rule names someone absent from the target: row fails, target folder unchanged (and not created)
- [ ] Re-deploy an existing folder to confirm the update path translates the same way
- [x] Deploy-to-staging/prod drawer (`DeployWorkspace.svelte`), which pulls folders in as dependencies and shares this path: dropped-access warning fires, and a folder whose identity rule cannot be resolved shows `Failed` with nothing written to the target
- [x] Folder owned by an **instance group**, deployed parent → fork: the owner and its ACL entry survive
- [x] Folder whose identity **rule** names an instance group: refused, since the server accepts it at folder-create and then rejects every item created in the folder (confirmed directly against the API)
- [ ] One pass with a non-admin, non-operator member as the deployer: every folder deploy now reads the user list of *both* workspaces, and the shared dev database has RLS stripped, so it would not surface a permission difference

## Known trade-off

`principalTranslator` is rebuilt per folder, and both bulk paths deploy sequentially, so a deploy touching N folders pays N × (2 × `listUsers` + the group pages). Memoizing it for the duration of a run would need the translator threaded through `deployItem`'s signature; deploy sets are typically a handful of folders, so this is left as-is deliberately rather than overlooked.

### DIFF
--- a/backend/windmill-api/openapi.yaml
+++ b/backend/windmill-api/openapi.yaml
@@ -24471,7 +24471,7 @@ components:
         type: integer
     PerPage:
       name: per_page
-      description: number of items to return for a given page (default 30, max 100)
+      description: number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)
       in: query
       schema:
         type: integer
--- a/backend/windmill-api/src/mcp/auto_generated_endpoints.rs
+++ b/backend/windmill-api/src/mcp/auto_generated_endpoints.rs
@@ -326,7 +326,7 @@ pub fn all_tools() -> Vec<EndpointTool> {
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "label": {
                         "type": "string",
@@ -518,7 +518,7 @@ pub fn all_tools() -> Vec<EndpointTool> {
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "resource_type": {
                         "type": "string",
@@ -591,7 +591,7 @@ pub fn all_tools() -> Vec<EndpointTool> {
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "order_desc": {
                         "type": "boolean",
@@ -853,7 +853,7 @@ Creates a new version of an existing script when called with the same path and t
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "order_desc": {
                         "type": "boolean",
@@ -1089,7 +1089,7 @@ Creates a new version of an existing script when called with the same path and t
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "path_start": {
                         "type": "string",
@@ -1571,7 +1571,7 @@ Creates a new version of an existing script when called with the same path and t
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "all_workspaces": {
                         "type": "boolean",
@@ -1705,7 +1705,7 @@ Creates a new version of an existing script when called with the same path and t
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "trigger_kind": {
                         "type": "string",
@@ -2313,7 +2313,7 @@ You should get the schema of the script or flow before updating the schedule to 
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "args": {
                         "type": "string",
@@ -2378,7 +2378,7 @@ You should get the schema of the script or flow before updating the schedule to 
                 },
                 "per_page": {
                         "type": "integer",
-                        "description": "number of items to return for a given page (default 30, max 100)"
+                        "description": "number of items to return for a given page (default 1000, max 10000; the script and job listings default to 10000 and apply no maximum)"
                 },
                 "ping_since": {
                         "type": "integer",
--- a/cli/src/commands/workspace/merge.ts
+++ b/cli/src/commands/workspace/merge.ts
@@ -65,6 +65,7 @@ const provider: DeployProvider = {
   // Users and groups
   listUsers: wmill.listUsers,
   listGroups: wmill.listGroups,
+  listGroupNames: wmill.listGroupNames,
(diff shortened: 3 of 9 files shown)


Guilhem-lm:
Second round pushed in 9800051b00 — all four findings addressed.

**P1 (instance groups dropped)** — confirmed and fixed. `translateAccess` now resolves owners/ACL against `group_ ∪ instance_group` via `listGroupNames` (unpaginated server-side, so one extra call, not a paged one), while `translateRule` keeps `group_` alone. Verified both halves against a live EE backend: a folder owned by `g/<instance group>` keeps that owner and ACL entry across a deploy, and a folder whose *rule* names one is still refused — creating a script in such a folder really does 400 with `resolves to group 'g/scratch_igroup' which does not exist in this workspace`, so accepting it would have written a folder nothing can be deployed into.

**P2 (`create_folder` appends the caller — not on update)** — confirmed at `folders.rs:475` (`if !authed.is_admin`). `owners` is now omitted rather than sent as `[]` when every owner is untranslatable, so an admin deploy can't leave an owner-less folder; on create the server still appends the caller. The comment no longer claims a guarantee that only covers one branch.

**P2 (`default_permissioned_as: []` clears, `labels: undefined` preserves)** — picked mirroring for both: `labels: folder.labels ?? []`. A label or rule cleared at the source has to clear in the copy.

**Nit (pagination short-circuit)** — you're right, and I've restored it. I removed it in response to a pre-push review that worried about an instance capping `per_page` below the 1000 asked for; that can't happen, since `list_groups` goes through `paginate()` which caps at `MAX_PER_PAGE` (10000), and the constants have never been lower. The reason is now recorded in the comment so it doesn't get flipped a third time.

Also closed the two verification gaps you and Pi both flagged: the deploy-to-staging/prod drawer is now exercised in a browser (dropped-access warning fires; a folder whose rule can't resolve shows `Failed` with nothing written to the target), and the tests grew two cases — the instance-group split, and that an all-untranslatable owners list never blanks the target's.

Still open, deliberately: the non-admin deployer pass, and the per-folder rebuild of `principalTranslator` (noted as a trade-off in the description).

PR #10545, opened 2026-08-05
feat: register mounted CA certificates in windmill_extra at startup

rubenfiszel:
## Summary

`windmill_extra` ignored `RUN_UPDATE_CA_CERTIFICATE_AT_START`. The server and worker binary have honored it since forever (`backend/src/main.rs`, best-effort and default-off, with `RUN_UPDATE_CA_CERTIFICATE_PATH` to override the tool), so an operator who sets it across their deployment got it silently dropped on this one container — the same class of no-op as `INIT_SCRIPT` before #10532.

Follow-up to #10532, which left `INIT_SCRIPT=update-ca-certificates` as the only way to get a CA into this container's trust store.

## Changes

`docker/entrypoint-extra.sh` registers CA certificates before any service starts, by two routes:

- **`RUN_UPDATE_CA_CERTIFICATE_AT_START=true`**, with `RUN_UPDATE_CA_CERTIFICATE_PATH` overriding the tool. Same names, same default-off, same best-effort behavior as the server/worker, so one setting covers every container.
- **Certificates present in `/usr/local/share/ca-certificates/`** trigger the same action without the env var. Mounting a CA there is unambiguous intent and does nothing until registered, so requiring the operator to also discover a variable is what made this a support case in the first place.

Best-effort on purpose, unlike `INIT_SCRIPT`: a missing tool is skipped, a failing one warns and startup continues. The entrypoint explicitly supports arbitrary non-root UIDs, which cannot write `/etc/ssl/certs`, and a deployment that never needed a custom CA must still boot. `INIT_SCRIPT` keeps its abort-on-failure semantics and still runs after this, so it can do anything more involved.

The compose file gains a commented mount showing where the CA goes.

## What this does not fix, and why the docs say so

Registering the CA reaches only what reads the system trust store: Python's stdlib `ssl`, `curl`, `git`. It does **not** help the things a debug session actually installs with — uv verifies against its own bundled roots unless `PY_NATIVE_CERT`/`UV_NATIVE_TLS` is `true`, Bun and Node read only `NODE_EXTRA_CA_CERTS`, and `requests` carries certifi. All measured while doing #10532.

That is the trap worth avoiding: making the CA "just work" invites the conclusion that mounting it is sufficient, and then `pip` still fails with an untrusted certificate. `debugger/README.md` now states the split explicitly.

## Test plan

Exercised each path against the real entrypoint with a stub tool, services disabled:

- [x] `RUN_UPDATE_CA_CERTIFICATE_AT_START=true` runs the tool and logs success; `TRUE` is normalized too.
- [x] Tool exits non-zero → `WARNING: … may not own /etc/ssl/certs; continuing`, and the container proceeds to start services rather than dying on `set -e`.
- [x] Tool missing/not executable → logs `skipping CA update`, startup continues.
- [x] Certificates in the mount directory → runs without the env var; an **empty** directory stays completely silent.
- [x] Neither set → no CA output at all, behavior identical to today.
- [x] Ordering: CA update logs before `INIT_SCRIPT` runs, which logs before the service banner. `INIT_SCRIPT` still aborts with its own exit code (verified `exit 9` → container exit 9).
- [x] A `.pem` (or any non-`.crt`) in the mount directory warns that it will be ignored instead of reporting a bogus `CA certificates updated`; a `.crt` still auto-runs, including one nested in a subdirectory — the detection mirrors Debian's own `find -L "$LOCALCERTSDIR" -type f -name '*.crt'`, which is unbounded in depth.
- [x] A failing tool's own stderr reaches the log rather than a guessed cause.
- [x] `bash -n` on the entrypoint, compose file parses.

No test shipped: this is container startup wiring with no unit-testable surface, and the matrix above covers every branch.

---
## Summary by cubic
Enable automatic registration of mounted CA certificates in the `windmill_extra` container at startup, matching server/worker behavior. Prevents silent no-op when operators set CA update env vars across all containers.

- **New Features**
  - `docker/entrypoint-extra.sh` updates the system trust store before services start when either:
    - `RUN_UPDATE_CA_CERTIFICATE_AT_START=true` (override via `RUN_UPDATE_CA_CERTIFICATE_PATH`), or
    - `*.crt` exists under `/usr/local/share/ca-certificates/` (recursive, follows symlinks like Debian).
  - Best-effort: skips if the tool is missing, warns on failure, never blocks boot; runs before `INIT_SCRIPT`, which still aborts on failure. `docker-compose.yml` adds a commented mount; `debugger/README.md` explains system trust store limits and notes vars for tool-specific roots (`PY_NATIVE_CERT`, `UV_NATIVE_TLS`, `NODE_EXTRA_CA_CERTS`, `REQUESTS_CA_BUNDLE`).

- **Bug Fixes**
  - Avoid false success: only update when `*.crt` is present and the tool is executable; warn on non-`.crt` mounts. Detection mirrors Debian’s behavior (recursive, follows symlinks).

<sup>Written for commit 7d2d5c9eded8b78933837826ec99637215098931. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10545?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/debugger/README.md
+++ b/debugger/README.md
@@ -119,8 +119,8 @@ not enough on its own, since `requests` carries its own bundle and Node reads on
 only the service needs them.
 
 Registering that CA in the container's system store happens on its own: mount it into
-`/usr/local/share/ca-certificates/` and `windmill_extra` runs `update-ca-certificates` before
-starting any service. `RUN_UPDATE_CA_CERTIFICATE_AT_START=true` forces the same thing whether or not
+`/usr/local/share/ca-certificates/` **named `*.crt`**, the only extension `update-ca-certificates`
+reads, and `windmill_extra` runs it before starting any service. `RUN_UPDATE_CA_CERTIFICATE_AT_START=true` forces the same thing whether or not
 certificates are mounted there, and `RUN_UPDATE_CA_CERTIFICATE_PATH` overrides the tool, matching
 the server and worker. Both are best-effort: a UID that cannot write `/etc/ssl/certs` logs a warning
 and the container still boots. `INIT_SCRIPT` remains the hook for anything more involved, and unlike
--- a/debugger/README.md
+++ b/debugger/README.md
@@ -118,9 +118,18 @@ not enough on its own, since `requests` carries its own bundle and Node reads on
 `NODE_EXTRA_CA_CERTS`. Registry settings are deliberately not forwarded: they carry credentials and
 only the service needs them.
 
-To install that CA into the container's system store in the first place, set `INIT_SCRIPT` on the
-`windmill_extra` container (e.g. `INIT_SCRIPT=update-ca-certificates`). It runs before any service
-starts and aborts startup if it fails, the same hook a worker offers.
+Registering that CA in the container's system store happens on its own: mount it into
+`/usr/local/share/ca-certificates/` and `windmill_extra` runs `update-ca-certificates` before
+starting any service. `RUN_UPDATE_CA_CERTIFICATE_AT_START=true` forces the same thing whether or not
+certificates are mounted there, and `RUN_UPDATE_CA_CERTIFICATE_PATH` overrides the tool, matching
+the server and worker. Both are best-effort: a UID that cannot write `/etc/ssl/certs` logs a warning
+and the container still boots. `INIT_SCRIPT` remains the hook for anything more involved, and unlike
+the CA update it aborts startup when it fails.
+
+Note what the system store does *not* cover, which is most of what a debug session installs with:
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 7d2d5c9eded8b78933837826ec99637215098931

PR #10546, opened 2026-08-05
fix(debugger): confine prepare-deps under nsjail in both language paths

rubenfiszel:
## Summary

`windmill prepare-deps` was spawned with Bun's raw `spawn` in both the Python and the TypeScript
session, bypassing the nsjail wrapping the debugged script itself gets. With `ENABLE_NSJAIL=true`,
`uv pip install` builds source distributions — running their build backend's arbitrary Python — and
`bun install` runs postinstall scripts, so a package's own code executed unconfined next to the LSP,
multiplayer and gateway services in the `windmill_extra` container. Before the install was hoisted
into the service (#10533) it was launched by the already-jailed Python DAP server, so this was a
regression in confinement.

Second defect, verified on the base: killing the installer did not reap the `uv` (or `bun`) it had
already spawned. The grandchild was reparented to init and kept downloading, so the timeout and the
cancel-on-disconnect only half-worked.

Rebased on #10532, which landed meanwhile; its `python_path` pin, `sessionEnv()` allowlist and
SIGKILL semantics are kept as they are and exercised together with the jail below.

## Changes

**The installer is jailed in both languages.** The service's Python installer now goes through
`spawnProcess()` (`stdin` and an environment passthrough added to `SpawnOptions` for it), and the
Bun session wraps its own installer the same way. The two inlined nsjail command builders — one in
each file — collapse into a shared `nsjailWrap()`.

The installer keeps the service's whole environment rather than the minimal set `spawnProcess()`
gives a debuggee; that is how the registry credentials and CA settings reach `uv`/`bun`, and the
jail config's `keep_env` carries them across the boundary. The debugged script's environment is
untouched and still holds neither — re-checked, including `/proc/self/environ`.

**The installer runs in its own process group** (`detached`) and is signalled as a group by a new
`killProcessTree()`. It reads the group id back from `/proc` rather than assuming it, because a
group kill aimed at the service's own group would take down every service in the container; a child
without `detached` degrades to the plain kill. SIGKILL, as on the base, since `prepare-deps` does
not act on SIGTERM while uv is running. With nsjail on, the PID namespace tears the tree down
anyway — the group is what covers the unjailed default.

**Cancellation is finished off.** A kill is no longer reported to the client as an install failure —
it ends the read with nothing to parse, which the parse path blamed on the installer, so pressing
Stop produced `Failed to prepare dependencies: windmill binary produced no response`. The Bun
session, which answers `launch` only after installing, now answers it on that path too. The
standalone Bun server's close handler only dropped the session from its map, so its installer,
debuggee and temp dir were never cleaned up; it calls `cleanup()` now, like the unified service
always has. The teardown flag is scoped to a launch rather than the session, since `cleanup()` also
runs when a program finishes normally.

`debugger/README.md` records the two constraints a future change would break: the installer needs
`keep_env` (or an explicit carry of the registry/CA variables), and it needs its own process group.

## Test plan

Driven against the real service with `--nsjail --nsjail-config`, with the config's `/debugger` mount
repointed at the worktree and the `windmill` binary under `/tmp`. Before/after runs use the base
branch's sources unchanged, in the same harness.

- [x] **Python session, jailed, end to end.** `import tinydb` installs and runs: log line `Spawning
      with nsjail: nsjail --config … -- <windmill> prepare-deps`, then `tinydb version: 4.8.2` and
      `hostname: debugger` (the jail's UTS name) from the debuggee.
- [x] **Cold cache, so the jailed install really goes to the index** (`WINDMILL_DIR` repointed;
      6.8 MB of uv cache written from inside the jail). #10532's interpreter pin holds through it:
      the venv is `python3.14` and the debuggee reports `interp: 3.14.6`.
- [x] **TypeScript session, jailed, end to end.** `import { camelCase } from "lodash-es"` installs
      through the jailed `bun install` and prints `camelCase: helloJailedWorld`.
- [x] **The installer is actually confined.** Sampled mid-install: its `pid`, `mnt`, `user` and
      `uts` namespaces all differ from the service's. On the base with `--nsjail` on, all four match
      the service — the gap this fixes.
- [x] **Registry settings still cross the jail.** The jailed installer's `/proc/<pid>/environ`
      carries `PY_INDEX_URL` and `UV_HTTP_TIMEOUT`; the install fails against a tarpit index exactly
      as configured, so they are in force and not merely present.
- [x] **No credential regression.** A Python session dumping `os.environ` and `/proc/self/environ`
      sees no registry variable and not a planted service-only secret; same for a TS session.
- [x] **Disconnect reaps the tree**, both languages, jailed and unjailed: hanging up mid-install
      leaves no `nsjail`, no installer and no `uv`/grandchild. On the base the installer and `uv`
      both survive — and outlive the service itself.
- [x] **Timeout reaps the tree.** With `DAP_PREPARE_DEPS_TIMEOUT_MS=8000` against a stalled index:
      nothing survives the deadline, the client gets `dependency installation timed out after 8s`,
      and the session still starts without the dependency, as before.
- [x] **Standalone Bun server**: same disconnect check against `dap_websocket_server_bun.ts` run
      directly. On the base the installer and its grandchild survive.
- [x] **Two launches on one socket** both run, in both languages — the case the per-launch teardown
      flag makes valid.
- [x] **Real failures still reported.** An unresolvable import still surfaces uv's own message
      (`No solution found when resolving dependencies`) rather than a bare `ModuleNotFoundError`.
- [x] Both entrypoints bundle (`bun build`); `tsc --strict` reports the same 13 pre-existing errors
      as the base and no new ones. No Rust or frontend code changed, so `docs/validation.md` adds
      nothing here.

No test ships: `debugger/` has no CI harness (its `test_*` files are manual scripts needing a live
service), and the only pure function added is a string builder — a unit test for it would be the
ceremonial kind AGENTS.md says not to ship. The substance is process lifecycle and sandbox wiring,
which only a real run exercises.

### DIFF
--- a/debugger/README.md
+++ b/debugger/README.md
@@ -129,6 +129,14 @@ unsandboxed. Isolating sessions from the service takes `--nsjail --nsjail-config
 nsjail.debug.config.proto`: it is that config's PID namespace and `mount_proc` that put the service
 out of reach, not the flag on its own.
 
+The installer is jailed on the same terms, in both languages: `uv pip install` builds source
+distributions and `bun install` runs postinstall scripts, so a package's own code executes there
+too. It keeps the service's environment across that boundary — the config sets `keep_env`, which is
+how the settings above reach it — so replacing that with an allowlist would have to carry the
+registry and CA variables in explicitly. It also runs in its own process group, because `uv` and
+`bun` are grandchildren: signalling only the installer reparents them to init and they keep
+downloading, which would make the timeout and the cancel-on-disconnect half-measures.
+
 ### Frontend Integration
 
 ```svelte
--- a/debugger/dap_debug_service.ts
+++ b/debugger/dap_debug_service.ts
@@ -40,7 +40,12 @@ import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
 // Import the working Bun debug session from the standalone server
-import { DebugSession as BunDebugSessionWorking, type NsjailConfig } from './dap_websocket_server_bun'
+import {
+	DebugSession as BunDebugSessionWorking,
+	killProcessTree,
+	nsjailWrap,
+	type NsjailConfig
+} from './dap_websocket_server_bun'
 import { sessionEnv } from './env_passthrough'
 
 // ============================================================================
@@ -349,6 +354,17 @@ interface SpawnOptions {
 	cmd: string[]
 	cwd?: string
 	env?: Record<string, string>
+	stdin?: Blob
+	/**
+	 * Hand the child this process's whole environment rather than the minimal set below. Only
+	 * the dependency installer wants it: the registry credentials and CA settings it reads are
+	 * precisely what the minimal set exists to keep away from a debugged script.
+	 */
+	inheritEnv?: boolean
+	/**
+	 * Start the child in its own process group, so killProcessTree can signal what it spawns.
+	 */
+	detached?: boolean
 	stdout?: 'pipe' | 'inherit'
 	stderr?: 'pipe' | 'inherit'
 }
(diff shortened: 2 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 281ff2c783e0be43b6cb01bc05e4891c964b1221

PR #10547, opened 2026-08-05
fix(frontend): draw the tab strip's scroll bar instead of the native one

Guilhem-lm:
## Summary

The AI-sessions preview tab strip clipped its tabs at the top and left a wide gutter under them once the tabs overflowed. The strip is a fixed 32px (`h-8`), and `DraggableTabs` sized its scroll row to fit tabs + a 4px scrollbar reserve — but a *native* horizontal scrollbar claims layout height on top of that reserve, and how much is engine-specific:

| | scrollbar cost | tab clipped at top | gap under tab |
|---|---|---|---|
| Firefox 153 (`scrollbar-width: thin`) | **11px** | **3.5px** | **11.5px** |
| Chrome / Safari, overlay scrollbars | 0px | none | 5.5px |

`--wm-scrollbar-size: 4px` is a WebKit-only knob, so Firefox ignores it and spends 11px; 24px of tab plus an 11px bar cannot fit in a 32px strip under any padding arrangement. This is a regression from #10003, which replaced the melt scroll-area — whose 4px bar was *absolutely positioned "so toggling it never shifts layout"* — with a native `overflow-x: auto` row.

The fix hides the native bar and draws the thumb, so it costs no layout height and is 4px in every engine.

## Changes

- `DraggableTabs`: hide the native scrollbar (`no-scrollbar`) and render a 4px thumb positioned from `scrollLeft`/`scrollWidth`. Hover-revealed like the bar it replaces, draggable via pointer capture, and only rendered while the tabs overflow. `aria-hidden` — it mirrors scroll position rather than posing as a control at a 4px hit target, and the strip stays scrollable by wheel, trackpad and the existing arrow-key tab navigation.
- Tabs and thumb share one line: the tabs centre in the strip and the thumb sits on the strip's bottom edge, flush under them. The tabs land in the same place whether or not the strip overflows.
- Tab height `h-7` → `h-6` (28px → 24px) so the tabs clear the bar in a 32px strip.
- Strip default height `h-9` → `min-h-8`, matching the height the sessions caller already forced, so every strip has the same geometry; `min-h` rather than `h` so a taller `trailing` control can still grow it.
- `w-max` on the tab row: without it the row's box is pinned to the viewport and the tabs overflow *out* of it, so the `ResizeObserver` never sees a tab open and the thumb goes stale.
- Each dnd item wrapper is `flex`: an `inline-flex` tab in a block wrapper sits on a text baseline and rides ~1.5px off the row's centre.
- Comment fixes for the surfaces that no longer use `ScrollableX` for this (kitchen sink caption, `ScrollableX`, `app.css`), including the Firefox cost so the next caller doesn't repeat it.

`ScrollableX` itself is unchanged and still used by `Tabs`/`TabsV2` headers, code blocks and the chat — none of them are height-constrained, so nothing is clipped there.

## Screenshots

Sessions preview strip, 7 tabs overflowing, mid-scroll (Chromium, 4×).

**Before** — the native thumb sits below the strip's bottom border, over the previewed page:

pr-sessions-before

**After** — the thumb is inside the strip, flush under the tabs:

pr-sessions-after

Raw-app editor strip, now 32px and sharing the same geometry:

pr-rawapp-wide

Measured in the running app, sessions strip: `3.5 air · 24 tab · 0 · 4 bar · 1 border = 32`, identical with the tabs fitting (only the thumb stops rendering).

## Test plan

- [ ] **Firefox** (where the bug is visible): open a session, open enough preview tabs to overflow the strip — tabs are not clipped at the top, the thumb appears on hover, drags, and tracks the cursor
- [ ] Chrome: same strip, confirm the WebKit path did not regress
- [ ] Tabs land at the same height whether or not the strip overflows (widen the panel until they fit)
- [ ] Raw-app editor, split with preview: both strips are 32px, the split/rebuild buttons still sit clear of the pane border
- [ ] Drag-reorder still works, and dragging the thumb while a tab opens keeps the thumb under the cursor
- [ ] Drag a pane nearly shut: the thumb stays inside the strip instead of hanging out of it

### DIFF
--- a/frontend/src/lib/assets/app.css
+++ b/frontend/src/lib/assets/app.css
@@ -280,8 +280,9 @@
 	}
 
 	/* Subtle scrollbar: a thin, rounded thumb that only appears on hover, on both
-	   axes. Shared by ScrollableX (tab strips, code blocks) and the AI chat. Size
-	   via the `--wm-scrollbar-size` var (default 6px). Higher specificity than the
+	   axes. Shared by ScrollableX (code blocks, tab headers) and the AI chat. Size
+	   via the `--wm-scrollbar-size` var (default 6px) — WebKit only; Firefox sizes
+	   `thin` itself and spends ~11px of the box on it. Higher specificity than the
 	   app-wide `*::-webkit-scrollbar`, so it overrides it. */
 	.scrollbar-subtle {
 		scrollbar-width: thin;
--- a/frontend/src/lib/components/common/ScrollableX.svelte
+++ b/frontend/src/lib/components/common/ScrollableX.svelte
@@ -4,7 +4,9 @@
 	// Subtle horizontal scroll: native overflow (so wheel/trackpad/drag always
 	// work) with the shared `.scrollbar-subtle` thumb (thin, hover-revealed). Bar
 	// thickness is tunable via the `--wm-scrollbar-size` CSS var (pass through
-	// `style`), so denser callers (e.g. the tab strip) can shrink it.
+	// `style`) — WebKit only, so Firefox spends its own ~11px of the box on the
+	// bar whatever this says. Not for a height-constrained row: draw the thumb
+	// yourself there, the way the tab strip does.
 	let {
 		class: c = '',
 		style = '',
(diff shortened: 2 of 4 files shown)


Guilhem-lm:
✅ Review round clean @ 1dd63bf9f54996f8b83fce0016d149a5549c49e5 — nit-only verdicts at 4fe362cd52949f4343256e813d27a665fca8a4c3 (codex: Good to merge, pi: Good to merge, claude: nit); the one-sided `thumbLeft` clamp is fixed in 1dd63bf9f54996f8b83fce0016d149a5549c49e5.

PR #10548, opened 2026-08-05
feat: add a load all to the tree view's per-folder pager

rubenfiszel:
## Summary

The home page tree view loads a folder's runnables one server page at a time; its footer reads
`Showing 300 of 750 items in f/foo` with a single **Load more** button, so reaching an exact
count on a folder several pages deep takes a click per page. This adds a **Load all** next to it
that pages the prefix to the end in one click.

Paging a folder to completion can leave thousands of rows under one node, and the tree used to
mount every loaded row when a node was opened on its own — 9000 rows locked up the tab. A node
now renders at most 500 rows and the existing client `Show more` pager reveals the rest, so both
pagers can be on screen at once: one bounds what is rendered, the other what is fetched.

## Changes

- `Load all` in the per-prefix footer, alongside `Load more`. It reuses the same
  `loadOwnerItems` path with an `all` flag, so `loading` stays set for the whole run and both
  buttons are disabled until it finishes.
- `loadOwnerItems`'s third positional `force` boolean became an options object
  (`{ force?, all? }`) — the only other caller is the reload path.
- The run is bounded by `OWNER_LOAD_ALL_PAGES` (100) and aborts as soon as the node is
  collapsed; whatever it reached is committed, so the footer resumes from there.
- A mid-run fetch failure now persists the cursor/`hasMore`/`loaded` the run reached. Without
  `loaded` moving with the cursor, the retry is treated as a first load, restarts from no cursor
  and re-reads every page already merged.
- A node renders at most `LAZY_RENDER_MAX` (500) rows, and the server pager now also shows for a
  node opened on its own so it stays reachable past that ceiling. One `Show more` reveals another
  500 when the node holds that many, instead of 30.

## Known limitation (pre-existing)

A reload — any row mutation, or an in-place sort/archive/kind change — re-fetches each open
prefix's first page only, so a folder paged to the end falls back to one page. That is how the
tree has always behaved after repeated `Load more` clicks; `Load all` just makes it easier to
notice. Restoring the full depth on reload would mean replaying every page on every star or
archive click, which is a separate change.

## Screenshots

Footer with the new button, on a folder holding 750 items across 3 server pages:

shot-footer

After **Load all** — every row fetched, the server pager gone, rendering bounded by the client pager:

shot-after-load-all

## Test plan

Exercised against a live dev server with seeded folders (750, 9000, and a 2000-item nested
subfolder):

- [x] 750-item folder: `Load all` issues 2 requests, footer disappears, badge reads `(750 items)`
- [x] 9000-item folder: 30 requests in ~2s, all rows fetched, tab stays responsive (rendering
      capped at 500), `Show more` steps by 500
- [x] Nested subfolder whose rows came from an ancestor's pages: fills in without duplicating
- [x] Mid-run request failure (page 3 aborted via route interception): one toast, rows kept, and
      the retry's first request carries the saved cursor instead of restarting
- [x] Partially paged node shows both pagers with distinct labels (`loaded rows` vs `items in f/big`)
- [x] `Load more` unchanged — still advances exactly one page
- [x] "Expand all" unaffected: 30 rows per node, 61 rendered in total
- [x] `npm run check` — 0 errors, no new warnings in the changed files

---
## Summary by cubic
Adds a Load all button to the tree view folder pager to fetch all pages in one click. Caps per-node rendering to keep the UI responsive and makes long loads resumable; the pager stays visible during long runs.

- New Features
  - Added Load all next to Load more; fetches to the end in one run (max 100 pages), disables both buttons while loading (only the clicked one spins), keeps the pager visible mid-run, and stops if the node collapses. Footer resumes from the last cursor. Updated `loadOwnerItems` to use opts `{ force?, all? }` instead of a boolean.
  - Bounded rendering to 500 rows per node; client Show more reveals rows in 500-row steps and still respects the cap under “expand all”.

- Bug Fixes
  - On fetch failure, save cursor/hasMore/loaded so retries resume instead of restarting (including failed first loads).
  - Show more step now sizes by items held, not what’s rendered, so each click adds visible rows.

<sup>Written for commit 1269c095196bbfab4ae66a6b530728762cdf4042. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/home/ItemsList.svelte
+++ b/frontend/src/lib/components/home/ItemsList.svelte
@@ -483,11 +483,15 @@
 				// there. Left at where this run started, it would re-read every page already
 				// merged — and each of those dedups to nothing, so a "Load more" after a run
 				// that failed deep in the stream would spend its whole budget adding no rows.
+				// `loaded` moves with it: a node still marked unloaded is retried as a first
+				// load, which starts from no cursor and throws the saved one away.
 				const prev = ownerLoad[owner]
+				const advanced = nextCursor != undefined
 				ownerLoad[owner] = {
 					...prev,
 					cursor: nextCursor ?? prev?.cursor,
-					hasMore: nextCursor != undefined || (prev?.hasMore ?? false),
+					hasMore: advanced || (prev?.hasMore ?? false),
+					loaded: advanced || (prev?.loaded ?? false),
 					loading: false
 				}
 				sendUserToast(`Failed to load ${owner}: ${e?.body ?? e?.message ?? e}`, true)
--- a/frontend/src/lib/components/home/ItemsList.svelte
+++ b/frontend/src/lib/components/home/ItemsList.svelte
@@ -480,11 +480,9 @@
 			} catch (e: any) {
 				if (gen !== treeGen) return
 				// Keep the cursor the pages that did land reached, so the next click resumes
-				// there. Left at where this run started, it would re-read every page already
-				// merged — and each of those dedups to nothing, so a "Load more" after a run
-				// that failed deep in the stream would spend its whole budget adding no rows.
-				// `loaded` moves with it: a node still marked unloaded is retried as a first
-				// load, which starts from no cursor and throws the saved one away.
+				// instead of re-reading pages that now dedup to nothing. `loaded` moves with
+				// it: a node still marked unloaded is retried as a first load, which starts
+				// from no cursor and throws the saved one away.
 				const prev = ownerLoad[owner]
 				const advanced = nextCursor != undefined
 				ownerLoad[owner] = {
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ 1269c095196bbfab4ae66a6b530728762cdf4042 — nit-only verdicts at bc9b00fdee6575d32f46561b993c6fae5514eb0b (codex: good to merge, pi: good to merge, claude: mergeable with nits); both nits addressed in 1269c095196bbfab4ae66a6b530728762cdf4042.

- The per-prefix footer no longer unmounts mid-run under "expand all" (it was gated on `effectiveMax >= item.items.length`, which the first page falsified, taking the spinner with it). Verified: 8 samples across a throttled run all show the footer with a live count climbing 600 → 2700.
- Only the clicked button spins now; the other keeps its label and is disabled. Verified in both directions.

Also trimmed the two comment blocks flagged against the ≤4-line rule.

PR #10550, opened 2026-08-05
feat(debugger): install debug session deps from the instance registry settings

rubenfiszel:
## Summary

`bun install` inside `windmill prepare-deps` could not reach a private npm registry: the CLI runs
standalone with no database, so the `npm_config_registry` / `npmrc` instance settings never reached
it, and every bun/uv invocation `env_clear()`s. The debug service is where those settings can be
read, so it fetches them from a new endpoint and passes them down to the CLI over the stdin request
it already sends.

The credentials stop at the installer. They are never put in a debug session's environment, and the
files that carry them are written outside the tree a session can see, because a session executes
user-supplied code and resolves its `node_modules` symlink into the install directory.

The same channel replaces the Python side's dependence on the debug service's own environment for
`pip_index_url` / `pip_extra_index_url`, so the limitation `debugger/README.md` documented is gone
for both languages.

## Changes

- **`GET /api/debug/registry_config`** (`windmill-api-debug`) serves `npm_config_registry`, `npmrc`,
  `bunfig_install_scopes`, `pip_index_url`, `pip_extra_index_url` and `uv_index_strategy`, resolved
  exactly as a worker resolves them: workspace override (EE) over `FORCE_<env>` > `global_settings` >
  `<env>`, blank meaning unset.
- **Authorization**: the session's Ed25519 launch token, verified server-side, and only when its new
  `registry_config` claim is set. `sign_debug_request` sets it to `!authed.is_operator`, so the
  settings only reach a session whose author can already run a preview job. For npm that discloses
  nothing new (a worker leaves the same `.npmrc` / `bunfig.toml` in the directory the previewed
  script runs in); the Python index URL, which otherwise only appears as uv's argv, does become
  readable where it was not before. Each session is served only the settings its own language
  installs with, so a captured token cannot be replayed for the other language's credentials.
- **Enterprise gate**: a CE instance serves none of the private-registry settings and says why in the
  session output, matching `read_ee_registry` on a CE worker. `uv_index_strategy` is served on any
  edition, as it is to a worker. An `EPHEMERAL_TOKEN` index URL is withheld too: only a worker can run
  the command that substitutes it.
- **`prepare-deps`** takes an optional `registry` object on its stdin request (older debug service
  images that send none are unaffected) and configures `bun install` with it through
  `write_bun_registry_config`, extracted from `bun_executor::gen_bunfig` so both install paths write
  the same `.npmrc` / `bunfig.toml`.
- **Credential handling**: those files go to a `0700` directory under `/var/tmp/windmill-debug-registry`
  and are read from there via `--config` and `HOME`, never from the install directory, since
  `nsjail.debug.config.proto` bind-mounts all of `/tmp` into every session. That config now also
  mounts `/var/tmp` as a tmpfs, one instance per jail, so a jailed install's credentials are
  invisible to sessions and die with the jail even under the SIGKILL the service uses. A `Drop`
  guard removes them on every other exit, and the next install sweeps anything older than an hour.
- **Debug service**: new `debugger/registry_config.ts` fetches the settings for both the Bun and
  Python sessions, logs only which keys were set, and surfaces the endpoint's message to the client.
  A 401/403/404 is expected rather than an error. Both sessions re-check for teardown after the
  fetch, so a client that disconnects mid-fetch no longer gets an installer spawned for it. Both
  Dockerfiles that vendor the debugger copy the new file.

## Test plan

- [x] `cargo check` and `cargo check --features enterprise,private`; `cargo test -p windmill-api-debug`
      and `-p windmill-worker --lib prepare_deps`
- [x] EE instance + a local Verdaccio requiring auth for `@windmill-test/*`: a TypeScript debug session
      importing `@windmill-test/private-pkg` installs and returns its value (screenshot below)
- [x] Same via `npmrc` instead of `npm_config_registry`, and the same two settings on the worker path
      (a real bun preview job) after the `gen_bunfig` extraction
- [x] Workspace override replaces the instance value; a blank override means unset
- [x] Operator session: `registry_config` claim false, endpoint 401, nothing shown to the user;
      `nativets` / `bun` tokens get the npm settings, `python3` the pip ones, anything else nothing
- [x] Missing / malformed / expired / claim-tampered tokens are all rejected
- [x] CE build: settings withheld, session output explains it, install falls back to npmjs (screenshot)
- [x] Python: with `pip_index_url` set, `uv` resolves against it and fails there; cleared, it installs
      from PyPI
- [x] Credentials: during an install they exist only under `/var/tmp/windmill-debug-registry/<uuid>`;
      the directory is gone afterwards, and a stale one left by a kill is swept by the next install
- [x] Under `--nsjail`: a session's `/var/tmp` is an empty tmpfs while `/tmp/windmill-deps` is
      visible, and a jailed session still loads dependencies installed from the private registry
- [x] Client disconnected while the settings fetch was in flight (proxy delaying the endpoint by 4s):
      the session logs the teardown and no installer is spawned

## Screenshots

debug session installing from a private npm registry

CE instance reports the Enterprise gate in the session output

## Unrelated bug found while testing

`NPMRC_SETTING` is missing from the settings listener in `backend/src/main.rs`, so changing the
`npmrc` instance setting has no effect on jobs until the server restarts (`npm_config_registry` and
`bunfig_install_scopes` next to it are wired). Predates this branch and is left alone here; the
debugger is unaffected, since the new endpoint reads `global_settings` directly.


rubenfiszel:
✅ Review round clean @ b9c2bef9b8d2983d10146d25a9409dc041133f87

PR #10551, opened 2026-08-05
fix: keep the same_worker pin across a flow module that spawns no job

rubenfiszel:
## Summary
In a `same_worker` flow, the step following a module that completed **without spawning a job** — an empty branch, an empty for-loop, or a module already marked `Success` — is pushed as an ordinary queued job instead of being pinned to the current worker. It is then routed by tag and can be picked up by any worker in the pool.

That breaks the `./shared` directory contract, and worse: if the step's tag resolves to a worker group that cannot execute its language, the step fails instantly and takes the flow down. A worker in native mode, for example, rejects any non-native job with `Worker is in native mode and cannot execute non-native job with language 'python3'` in ~2ms.

**Cause.** Such a module hands the flow back over the `UpdateFlow` channel (`worker_flow.rs:3211` → `PushNextFlowJob::Done(Some(UpdateFlow { .. }))`). The receiver in `result_processor.rs` resumed the flow with a hardcoded `unrecoverable = true` — the same value for all four producers of that message. That flag means "the previous step's worker crashed (OOM/zombie)" and is true for only one of them.

The hardcoded `true` dates from 2024-09 and was inert until #9662 (v1.731.0) made `unrecoverable` gate the pin:

```rust
let continue_on_same_worker = !unrecoverable
    && (flow.same_worker || job_same_worker)
```

Since v1.731.0 the flag reaches this path and silently drops the pin. `continue_with_runners` (dedicated flow-module runners) is disabled by the same flag on the same path.

**Observed failure.** A `same_worker` flow looping over records: `[b (python3) → f (python3) → g (branchone) → e (nativets hub script)]`. On the iteration where `f` returned null, no branch matched and the empty default made `g` complete without a job. `e` was then pushed unpinned, routed by its `nativets` tag to the native-mode worker group — the only group serving that tag. That worker's result processor then advanced the flow and pinned the *next* iteration to itself, so every python3 step in it, plus the failure module, died on the native-mode guard.

## Changes
The flag moves onto the message, so each producer states its own case instead of the shared receiver assuming the worst:

- `backend/windmill-worker/src/worker.rs`: add `unrecoverable` to `UpdateFlow`.
- `backend/windmill-worker/src/worker_flow.rs`: `false` for the three producers that hand back a live flow (module completed without a job, schedule-overlap stop-early, flow-level `skip_expr`); `true` for the suspend-timeout producer, which is the one case no live step can recover from.
- `backend/windmill-worker/src/result_processor.rs`: pass the message's flag through instead of a hardcoded `true`.
- `backend/tests/worker.rs`: regression test — a `same_worker` flow whose branchone matches no branch and has an empty default asserts that `v2_job.same_worker` is still true on the following step.

`unrecoverable` also suppresses retry evaluation and forces failure-module routing (`worker_flow.rs:3737`, `:3758`). Keeping it `true` for the suspend-timeout producer means that behaviour is unchanged: only the pin decision changes, and only for the producers where no worker died.

## Test plan
- [x] `cargo test --features deno_core,quickjs --test worker test_same_worker_survives_empty_branch` passes
- [x] Same test fails with the fix reverted (`left: Some(false), right: Some(true)`), confirming it pins the regression
- [x] `--test flow_engine_parity` — 32/32 pass
- [x] `--test worker flow` and `--test retry` show no new failures (`test_complex_flow_restart`, `retry::test_pass`, `retry::test_fail_step_one` fail identically on `main` in this environment — missing `go` binary)

Fixes WIN-2326

### DIFF
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -587,6 +587,69 @@ async fn test_deno_flow_same_worker(db: Pool<Postgres>) -> anyhow::Result<()> {
     );
     Ok(())
 }
+
+#[cfg(feature = "deno_core")]
+#[sqlx::test(fixtures("base"))]
+async fn test_same_worker_survives_empty_branch(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+
+    let server: ApiServer = ApiServer::start(db.clone()).await?;
+
+    // No branch matches and the default is empty, so `a` completes without spawning a job and
+    // hands the flow back over the UpdateFlow channel. `b` must still be pinned to the worker.
+    let flow: FlowValue = serde_json::from_value(json!({
+        "same_worker": true,
+        "modules": [
+            {
+                "id": "a",
+                "value": {
+                    "type": "branchone",
+                    "branches": [{
+                        "expr": "false",
+                        "modules": [{
+                            "id": "c",
+                            "value": {
+                                "type": "rawscript",
+                                "language": "deno",
+                                "content": "export function main(){ return 1 }",
+                            }
+                        }]
+                    }],
+                    "default": []
+                }
+            },
+            {
+                "id": "b",
+                "value": {
+                    "type": "rawscript",
+                    "language": "deno",
+                    "content": "export function main(){ return 42 }",
+                }
+            }
+        ]
+    }))
+    .unwrap();
+
+    let job = run_job_in_new_worker_until_complete(
+        &db,
+        false,
+        JobPayload::RawFlow { value: flow, path: None, restarted_from: None },
+        server.addr.port(),
+    )
+    .await;
+    assert_eq!(job.json_result().unwrap(), json!(42));
+
+    let same_worker: Option<bool> = sqlx::query_scalar(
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ ad9c1262402d8508fb7bbe749963a65fdd94d061 — nit-only verdicts at 5a2ea5714d36d31b2c688b8421fa024eebea5155 (codex: good to merge, pi: good to merge, claude: nit); the suspend-path comment nit is addressed in ad9c1262402d8508fb7bbe749963a65fdd94d061, a comment-only amend.

rubenfiszel:
Tightened after self-review, back to draft for a fresh round.

The three producers that hand back a live flow now forward their caller's `unrecoverable` instead of hardcoding `false`. `unrecoverable = true` can reach `push_next_flow_job` on a live worker with a live result channel — the error fallback in `update_flow_status_after_job_completion` (`worker_flow.rs:224`) retries the internal update with `true` after a DB error — and hardcoding `false` would have dropped that flag across the UpdateFlow hop. Forwarding it means the only behaviour that changes is the pin, and only when the caller reported nothing broken.

`UpdateFlow` is `Debug, Clone` only, never serialized, so the new field cannot affect a mixed-version fleet.

PR #10552, opened 2026-08-05
fix: keep the same_worker pin when a suspend ends without approval

rubenfiszel:
## Summary

Fixes WIN-2327.

A disapproved or timed-out approval gate hands the flow back through the `UpdateFlow` channel with `unrecoverable: true`. That flag means *"the previous step's worker died"*, and six sites read it. Five of them happen to want what it does here — but `continue_on_same_worker` and `continue_with_runners` do not: the worker that ran the approval step is **alive**. Unpinning the error handler and routing it by tag breaks the `./shared` contract of a `same_worker` flow and can land it on a worker group that cannot run it. That is exactly the defect #10551 fixed for the three producers that hand back a live flow; the suspend producer was the one left behind.

This replaces the boolean with a `StepFailureKind` enum so the suspend producer can say *"worker alive, but this failure is not the module's to handle"* instead of overstating a worker death.

## The open question in the issue, answered differently than recommended

The issue proposed treating a resolved suspend as an ordinary step failure — honouring `retry` and `continue_on_error`. **I verified that empirically and it is wrong.** The failure is not recorded against the module carrying the `suspend`; it is recorded against the step the gate was holding back, which never ran:

```
"modules": [
  { "id": "a", "job": "019fd3a9-…", "type": "Success" },   <- the module with `suspend`
  { "id": "b", "job": "00000000-0000-…-000000000000", "type": "Failure" }
]
```

So `continue_on_error` and `retry` are read from `b` — a step with no job, no execution, and no error of its own:

- **`continue_on_error`** — honouring it made the engine *skip `b` entirely* and jump to `c` (verified: `a` Success, `b` Failure with a nil job, `c` Success). That is not "continue on error", it is "silently delete the step that was awaiting approval".
- **`retry`** — retrying `b` re-enters the gate for `a`, i.e. re-asks approvers who just declined, and would park the flow again.

There is already a purpose-built, UI-exposed setting for continuing past a gate: **`suspend.continue_on_disapprove_timeout`** ("Continue on disapproval/timeout" in the step's Suspend settings). It is configured on the correct module (`a`), and it makes `b` actually *run* with the error as its resume payload — the documented pattern being a `branchone` right after that inspects `resume.error`. That path returns before the code touched here, so it is unaffected.

So `SuspendNotApproved` deliberately keeps the error policy bypassed and changes only the worker pin. Net behaviour delta is one row:

| module config (gate disapproved / timed out) | before | after |
|---|---|---|
| `continue_on_error` on the gated step | failure module | **unchanged** |
| `retry` on the gated step | retry skipped, failure module | **unchanged** |
| neither | failure module | **unchanged** |
| `suspend.continue_on_disapprove_timeout` | continues to next step | **unchanged** |
| flow is `same_worker: true` | handler unpinned, routed by tag | **handler stays on the live worker** |

## Changes

- `StepFailureKind` (`Normal` / `SuspendNotApproved` / `Unrecoverable`) replaces the threaded `unrecoverable: bool` on `UpdateFlow`, `update_flow_status_after_job_completion(_internal)`, `handle_flow`, `push_next_flow_job` and `handle_job_error`.
- Two predicates split the axes the boolean conflated: `honors_step_error_policy()` (only `Normal`) gates `inc` / `should_continue_flow` / retry / failure-module routing; `keeps_worker_pin()` (not `Unrecoverable`) gates `same_worker` and dedicated runners.
- The suspend producer sends `SuspendNotApproved`; every other call site maps to the boolean it passed before, so worker-death semantics are untouched.
- `Unrecoverable` rather than `WorkerDeath`: the variant has two producers, and the flow-status-update error path runs on a live worker (review catch — the name would have asserted something false).

## Test plan

- [x] `SQLX_OFFLINE=true cargo check --features deno_core,quickjs` — clean
- [x] New `disapproved_suspend_keeps_same_worker_pin` (`backend/tests/suspend_resume.rs`): `same_worker` flow, self-disapproving gate, failure module. Asserts the handler ran **and** that its `v2_job` row has `same_worker = true`.
- [x] Confirmed it fails without the fix (`Some(true)` vs `Some(false)`), and also fails if `honors_step_error_policy()` is flipped for `SuspendNotApproved` — one test, both halves of the contract.
- [x] `--test suspend_resume` 5/5, `--test flow_engine_parity` 32/32
- [x] `--test worker unrecoverable` 3/3 — the #9662 worker-death tests are undisturbed
- [x] `--test worker` 52 passed / 5 failed; the same 5 fail on an unmodified tree (`Python requires the python feature to be enabled`) — pre-existing, baselined against `main` by stashing
- [x] Frontend: `npm run check` — 0 errors, no new warnings
- [x] Browser (Playwright, headless): approval step + Continue on error shows the hint; enabling "Continue on disapproval/timeout" hides it; turning the approval off hides it
- [ ] Manual: a `same_worker` flow whose approval step writes to `./shared` and whose failure module reads it back — should now find the file for both disapproval and timeout

## Editor hint (follow-up commit)

The same confusion is reachable from the UI: turning on **Continue on error** for a step that
has an approval reads as "the flow keeps going no matter what", when a disapproval or timeout
still stops it. The step settings now say so, and point at the toggle that does continue past a
gate. It shows only when both settings are on and `continue_on_disapprove_timeout` is off, so it
disappears once you act on it.

## Screenshots

continue-on-error-approval-hint

---
## Summary by cubic
Fixes routing after disapproved or timed‑out approvals: keeps `same_worker` and dedicated runners so the failure handler stays on the live worker and preserves `./shared`; addresses WIN-2327. Adds a flow editor hint that `continue_on_error` does not cover approval gates.

- **Bug Fixes**
  - Treat a resolved approval as `SuspendNotApproved`: keep `same_worker` and dedicated runners; route only to the failure module.
  - Do not apply the gated step’s `retry`/`continue_on_error`; use `suspend.continue_on_disapprove_timeout` to continue past a gate.
  - Same‑worker zombies are `Unrecoverable`, so the handler is unpinned only when the worker is gone.
  - Test ensures the failure module runs with `same_worker = true`.

- **Refactors**
  - Replace `unrecoverable: bool` with `StepFailureKind` (`Normal` / `SuspendNotApproved` / `Unrecoverable`) across flow update and worker paths.
  - Add `honors_step_error_policy()` and `keeps_worker_pin()` to control retries/continue_on_error and worker pinning.

<sup>Written for commit b2107b7ef622d7cac803faa4c45e66c29852c0cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10552?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 61e80a7e7c0cab819248156100b87eb44ab397be

PR #10554, opened 2026-08-05
perf: keep run status out of flow graph node and edge data

rubenfiszel:
## Summary

Every job-status poll re-rendered the whole flow graph, even though a run never changes the graph's shape.

Run status (`flowJob`, `suspendStatus`, per-step `flowModuleState`) was baked into every xyflow node's and edge's `data` — `graphBuilder`'s `addEdge` spread `...extra` into **every edge's** data, so each edge carried the flow job and the entire `flowModuleStates` record. One poll therefore produced new object identities for the entire graph, which makes xyflow re-measure every node (`updateNodeInternals` → forced `clientHeight` reflows) and makes Svelte destroy and re-create every edge.

Run status is presentation, not structure, so it no longer travels through the structural build. It goes through a `FlowRunStatus` context store that the renderers read directly, and `updateStores` reconciles nodes/edges by id so unchanged entries keep their object identity.

Measured on a 33-step branchone flow (5 branches), instrumented from the moment the run starts with a `MutationObserver` on `.svelte-flow__viewport` plus `PerformanceObserver('long-animation-frame')`:

| | before | after |
|---|---|---|
| DOM mutations during a run | 5,540 | **1,690** (−69%) |
| long animation frames | 7 | **2** |
| long-frame time | 873 ms | **271 ms** (−69%) |
| blocking time | 413 ms | **171 ms** (−59%) |
| edges replaced per rebuild | all (1,424) | **0** |
| nodes replaced per rebuild | all | 6 (Input/Result only, on `isRunning`/`success` flips) |

The two long frames that remain are the run's genuine structure changes — the 44→87 node expansion at start and the completion update — not the poll loop.

## Changes

- Add `graph/flowRunStatus.svelte.ts`: a context store holding `flowJob`, `suspendStatus` and a `SvelteMap` of per-module states, so one step finishing wakes only that step's readers. Absent context is tolerated, for the graphs that never show run status.
- `graphBuilder`: stop putting `flowJob`, `suspendStatus` and `flowModuleState` in node/edge data. Replace the `...extra` spread in `addEdge` with the three fields the edge renderer actually reads. `extra.flowModuleStates` stays only for the parts of the graph's *shape* a run decides — the expanded loop iteration and error-handler-marker anchoring.
- Point the 9 renderers (`BaseEdge`, `ModuleNode`, `InputNode`, `ResultNode`, `CollapsedGroupNode`, `FailureModuleNode`, `BranchAllStart`, `BranchOneStart`, `AIToolNode`, `GroupModuleIcons`) at the context.
- `FlowGraphV2`: provide and sync the store; reconcile nodes/edges with `reuseUnchanged` (deep-equal by id, the idiom already used for the sugiyama memo in that file), plus a WeakMap so `nodesWithOffset` does not undo the identity reuse.
- Key the `computeAIToolNodes` memo on the agent's `agent_actions`. Its only run-status invalidator used to be `flowModuleState` sitting in node data; without a replacement, tool-call nodes appearing mid-run would render only when a whole row filled up. Covered by a new case in `AIToolNode.test.ts`.

## Screenshots

Flow editor after a run — branch node green with its duration, per-step statuses intact:

graph-status

Run viewer — 8 green nodes with per-step durations:

run-viewer

## Also in this PR

One unrelated CSS fix, called out because it is not a perf change. The graph's zoom bar
(`.svelte-flow__controls.wm-flow-controls`) applied a bare `border`, which sets only a
width; its colour came from the base layer's `.border` rule pulled in by `@apply`. In the
production bundle those land as **two separate rules of equal specificity**, so if the
colour one is missing the only colour left is preflight's `*{border-color:#e5e7eb}` — a
white outline on the dark bar. The rule now names `border-border-light` and is
self-contained.

Reported against this PR's CF preview. I could not reproduce it locally in dev or in a
production build, before or after the change, so the fix is defensive: it removes a real
ordering fragility that matches the symptom, but it is not proven to be that symptom's
cause.

## Test plan

- [x] `npx vitest run src/lib/components/graph` — 28 passed
- [x] Run a flow from the editor: statuses, durations and colours still render; run viewer shows per-step status with 0 console errors
- [x] Re-measured mutations / long frames / blocking before and after on the same flow
- [x] Selection: click a step, Escape, Shift-click another — only the second stays selected (the `offsetNodeCache` regression)
- [x] Watched a live run in the run viewer: statuses progress step by step as the flow executes
- [ ] Dark mode on the CF preview: the zoom bar's border is not white
- [ ] Run a flow with an AI agent step making 4+ tool calls and confirm each tool node appears as the call is made, and the final count matches the step's detail panel
- [ ] Open a flow with an approval/suspend step in the editor preview and confirm the hourglass and "waiting for events" card still render on the edge and on a collapsed group
- [ ] Run a for-loop flow, switch the displayed iteration mid-run, and confirm the graph still follows `selectedForloopIndex` and statuses keep updating after the last poll

---
## Summary by cubic
Move run status out of node/edge data into a shared context and reuse node/edge identities so job-status polls no longer rebuild the graph. This cuts long frames and DOM churn, fixes selection clearing, and shows AI tool nodes as calls arrive without rebuilding the editor; also pins the zoom bar border color.

- **Refactors**
  - Added `graph/flowRunStatus.svelte.ts` context for `flowJob`, `suspendStatus`, and per-module state via `SvelteMap`; renderers now read status from context.
  - `FlowGraphV2`: provides/syncs the context; reconciles nodes/edges with `reuseUnchanged`; caches y‑offset mappings to preserve object identity.
  - `graphBuilder`: removed run-status from node/edge `data`; edges now carry only fields the renderer reads.
  - `AIToolNode`: memo keys on each agent’s `agent_actions` so tool nodes update mid‑run, but ignores them in the editor to avoid rebuilds; documented invariants (no per‑poll values in edge `data`, tool memo keying).

- **Bug Fixes**
  - Tool-call nodes appear as each call is made; added a test for calls arriving without moving nodes.
  - Clearing selection works again by resetting the offset cache.
  - Zoom bar uses `border-border-light` to avoid white outlines.

<sup>Written for commit b514c8eccf4375668a1093b3b49abc0c8e70de98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/graph/FlowGraphV2.svelte
+++ b/frontend/src/lib/components/graph/FlowGraphV2.svelte
@@ -103,6 +103,7 @@
 	import { computeNodeExtraSpace } from './nodeExtraSpace'
 	import type { ModuleActionInfo } from '$lib/components/flows/flowDiff'
 	import { setGraphContext } from './graphContext'
+	import { setFlowRunStatusContext } from './flowRunStatus.svelte'
 	import { computeNoteNodes } from './noteUtils.svelte'
 	import { Tooltip } from '../meltComponents'
 	import { getNoteEditorContext } from './noteEditor.svelte'
@@ -371,6 +372,18 @@
 		groupDisplayState
 	} as any)
 
+	const flowRunStatus = setFlowRunStatusContext()
+	$effect(() => {
+		flowRunStatus.flowJob = flowJob
+	})
+	$effect(() => {
+		flowRunStatus.suspendStatus = suspendStatus
+	})
+	$effect(() => {
+		readFieldsRecursively(flowModuleStates)
+		untrack(() => flowRunStatus.setModuleStates(flowModuleStates))
+	})
+
 	if (triggerContext && untrack(() => allowSimplifiedPoll)) {
 		if (isSimplifiable(untrack(() => modules))) {
 			triggerContext?.simplifiedPoll?.set(true)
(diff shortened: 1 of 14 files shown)


rubenfiszel:
✅ Review round clean @ e535eec1aa693352f899abcfb9da9695abac1666 — nit-only verdicts at e535eec1aa693352f899abcfb9da9695abac1666; the undocumented zoom-bar CSS fix is now described in the PR body under "Also in this PR", including that it is defensive and unproven as the cause of the reported preview symptom.

PR #10557, opened 2026-08-06
feat(flow-editor): show an agent's tool-call status without moving the graph

rubenfiszel:


rubenfiszel:
I went looking for a backend change to make an agent's actions available to the editor's **Test this step** gesture, on the assumption that a single-step agent run has no flow parent and so never persists `agent_actions`. That assumption is wrong — the data is already there, and already live. Verified by running a real agent with a tool call, not by reading.

**Why it already works.** `ModuleTest.svelte:142` runs an `aiagent` module through `runFlowPreview`, wrapping it as a single-module flow. So the agent job does have a flow parent: `direct_parent_job_kind` is `FlowPreview`, not `AIAgent` (`ai_executor.rs:698`), `flow_status_job` is `Some(...)`, and `update_flow_status_module_with_actions` takes its `Step::Step { idx: 0 }` arm rather than the `_ => {}` fallthrough.

Job tree from one such run:

```
flowpreview  019fd4c4-8211…                                  <- carries flow_status
aiagent      019fd4c4-8229…  flow_step_id=a  parent=…8211…
preview      019fd4c4-867f…  runnable_path=…/a/tools/b       <- the tool job
```

**The API already serves it, mid-run.** Polling `GET /w/{ws}/jobs_u/get/{flowpreview_id}` once a second:

```
t=2  QueuedJob     {"id":"a","type":"InProgress"}
t=3  QueuedJob     {"id":"a","type":"InProgress",
                    "agent_actions":[{"type":"message"},
                                     {"type":"tool_call","job_id":"…","module_id":"b","function_name":"mytool"}],
                    "agent_actions_success":[true,true]}
t=4  CompletedJob  3 actions, 3 successes
```

`JobLoader.svelte:534` refreshes `job.flow_status` from `getJobUpdates` on every poll, so `testState.testJob.flow_status.modules[0]` is already in the browser and already updating live.

**What's missing is one hop in `modulesTest.svelte.ts`.** `jobToGraphModuleState` forwards only `args/type/job_id/tag/duration_ms/started_at` and drops the rest of the module status:

```ts
} else if (testState.testJob) {
    const agentModule = testState.testJob['flow_status']?.modules?.[0]
    return {
        ...,
        agent_actions: agentModule?.agent_actions,
        agent_actions_success: agentModule?.agent_actions_success
    }
}
```

Two things to watch when wiring it: the `testState.loading` branch returns `{type: 'InProgress', args: {}}` before `testJob` is ever consulted, so live decoration during a single-step test needs that short-circuit relaxed; and `modules[0]` is right only because the preview flow this gesture builds has exactly one module.

**One genuine gap, out of scope here.** A *nested* agent — an agent used as another agent's tool — records nothing. Its parent is the outer agent job, which has no `v2_job_status` row, so `flow_status_job` is `None` and both writers fall through. Confirmed on a nested run: job `019fd4c5-6cc9…` (kind `aiagent`, parent = the outer agent job) has no status row and no `flow_status` in `v2_job_completed`, while the outer agent's own actions are recorded normally. So nested tool nodes will degrade to no decoration until that job kind gets somewhere to write. Worth a separate issue rather than a change here.

PR #10567, opened 2026-08-06
fix: stop destroying AI sessions in workspaces reached without a usr row

AlexRV12:
## Summary

AI sessions created in the **Admins workspace** were silently destroyed on every page load, along with their attached files and artifacts.

Sessions are client-side (IndexedDB), so the backend cannot delete one directly. Instead the client asks `POST /workspaces/session_workspace_status` whether each referenced workspace still exists and acts on the answer itself — and one of the three answers is an irreversible delete.

That endpoint derived reachability from a `usr` membership row alone. But membership is not the only way into a workspace: the authed extractor also admits a superadmin into **any** existing workspace without a `usr` row (`windmill-api-auth/src/auth.rs:451`), and the `admins` workspace is created by migration with **no** `usr` rows at all. So `admins` was unreachable-by-definition for everyone, forever, and every load reported `deleted` and wiped the sessions.

The frontend carried the same bad assumption independently: `putSession` refused to write any session whose workspace was missing from `userWorkspaces` — which is the membership list, so it dropped a superadmin's writes for the same reason the SQL did.

**Scope:** superadmins only. No other role can stand in a workspace without a membership row, and operators are blocked from AI Sessions outright. On a self-hosted instance, though, the operator usually *is* the superadmin.

## Changes

**Backend** — `session_workspace_status` now resolves reachability the way the authed extractor does:
- a superadmin reaches any workspace whose row exists, membership or not
- `workspace.id IS NULL` is checked **first**, so a hard-deleted workspace still reports `deleted` for superadmins too — without that arm the naive fix makes those sessions linger forever
- a soft-deleted workspace reports `archived` (reversible) rather than `deleted`
- non-superadmins are byte-for-byte unaffected: the new arm is gated on `AND NOT $3`

Request/response shapes are unchanged, so no `openapi.yaml` edit and no client regeneration.

**Frontend** — `putSession` now has two independent guards:
- a `deletedSessionIds` tombstone recording ids whose record was removed, so a caller holding a pre-delete reference (the chat-id seeder awaits mid-loop) cannot write back a record whose files and artifacts were already GC'd
- the pre-existing membership check, retained for its separate job (stopping a UI action persisting into an unavailable workspace) and gated on `get(superadmin) === false`, since `userWorkspaces` only answers that question for a confirmed non-superadmin

All store access funnels through `deleteSessionRow` / `putSessionRow`, so the "a write cannot commit behind a delete" invariant holds at every one of the seven call sites rather than only where it was first enforced.

## Screenshots

Same flow both times: open a session in the Admins workspace, type a prompt, hard-reload.

**Before — the session and its draft are gone**

before

**After — both survive the reload**

after

## Test plan

- [x] `backend/tests/session_workspace_status.rs` — crosses superadmin/plain user against all five meaningful cases (`admins`, non-member live, non-member archived, member, never-existed). Both new `CASE` arms verified load-bearing by reverting each and watching the test fail.
- [x] `sessionStateIndexedDb.test.ts` — the two windows the membership gate targets, plus the tombstone blocking a stale-reference write. Also verified by falsification.
- [x] `cargo test --test session_workspace_status`, `npm run check` (0 errors), `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`
- [x] Manual: session in `admins` survives a hard reload; screenshots above are the same script run against the reverted and fixed backends.

## Deliberate non-changes

- **`usr.disabled` is not added to the join.** The extractor requires `disabled = false`; this query doesn't check it, so a disabled member still gets `active`. Tightening it would flip those sessions to `deleted` and destroy them, with no recovery on re-enable. The failure direction here is a session lingering, not one being lost. The doc comment now states this asymmetry rather than claiming parity.
- **No test for the post-await tombstone re-check.** One was written and deleted because it passed with the re-check disabled: when the DB handle is already open, the write commits before reconciliation reaches its delete, so the assertion cannot discriminate. A real one needs an injectable `whenReady` seam that `sessionsDb` does not expose.


AlexRV12:
✅ Review round clean @ 8ef5887eed9cfcee713c743d4fdfddbce631d8ff

PR #10569, opened 2026-08-06
fix(frontend): shimmer highlight uses text-emphasis, not white

Guilhem-lm:
## Summary

The loading shimmer on AI-chat tool rows swept a **white** copy of the label across the muted one. In dark mode that reads as a highlight; in light mode a white sweep over a light surface is close to invisible. The highlight is now `text-emphasis`, which is the emphatic end of the text scale in whichever theme is active.

## Changes

- `ChatCollapsibleCard.svelte`: `labelText` takes a `highlight` flag; the band copy renders with `text-emphasis` appended (twMerge lets it win over `text-secondary` / any caller `labelClass`).
- Dropped the `filter: brightness(0) invert(1)` hack on `.shimmer-band` — it existed only to force the copy white without out-specifying the label's utility classes, which the flag now does directly.

Effect per theme:
- **light**: highlight goes from `#ffffff` to `#1d2430`, so the sweep now darkens against the muted label instead of washing out.
- **dark**: `text-emphasis` is `#f3f4f6`, visually the same as the previous white.

## Screenshots

None attached: the change is a colour swap inside an animated sweep, so a still frame conveys little and staging one means catching a tool call mid-flight. Verified by reading the rendered token values; the sweep itself is unchanged.

## Test plan

- [ ] In light mode, run an AI chat tool call and confirm the running row's label sweeps to a dark highlight and is legible.
- [ ] In dark mode, confirm the sweep looks unchanged from before.
- [ ] Confirm the reasoning card in `AssistantMessage.svelte` (the other `shimmer` caller) still sweeps while streaming.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/ChatCollapsibleCard.svelte
+++ b/frontend/src/lib/components/copilot/chat/ChatCollapsibleCard.svelte
@@ -41,8 +41,14 @@
 </script>
 
 <div class={twMerge('font-mono text-xs', className)}>
-	{#snippet labelText()}
-		<span class={twMerge('text-secondary font-medium text-2xs', labelClass)}>
+	{#snippet labelText(highlight: boolean)}
+		<span
+			class={twMerge(
+				'text-secondary font-medium text-2xs',
+				labelClass,
+				highlight && 'text-emphasis'
+			)}
+		>
 			{label}
 		</span>
 	{/snippet}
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
Re [Claude's P2](https://github.com/windmill-labs/windmill/pull/10569#issuecomment-5203542316) (no screenshot / in-browser verification) — dismissing, deliberately.

The author is reviewing this one live in their own browser as part of an interactive styling pass, so the still would document a frame they have already looked at directly rather than add evidence. The point about the `rgba(0,0,0,0.2)` mask floor is a good catch and correct: a running row's label now rests very slightly darker than a settled one in light mode, where it previously rested slightly lighter. That is the intended direction — a running row should read as more prominent — so it is noted, not fixed.

The three token-value corrections are accurate and change nothing: the active dark theme is `dark-3` (`#eeeff2`, not `dark`'s `#f3f4f6`), still visually indistinguishable from the white it replaces.

Guilhem-lm:
✅ Review round clean @ d5cd700e2fe312749fd6fec5004a28e41d674fc2 — nit-only verdicts at d5cd700e2fe312749fd6fec5004a28e41d674fc2; sole nit dismissed in review reply (head unchanged)

PR #10570, opened 2026-08-06
feat: offer more dev workspace environment labels

rubenfiszel:
## Summary

A dev workspace's environment label is its badge text and, when git-sync is configured without promotion mode, the name of the branch it deploys to. Only `dev` and `staging` were accepted, which also capped a promotion chain at two dev workspaces: `reject_dev_label_taken_in_chain` requires every dev workspace in a chain to carry a distinct label, so with a two-word vocabulary the third pairing had nothing left to claim. Teams wanting a `demo` or `uat` environment, or a chain deeper than dev → staging → prod, hit a dead end where the settings tab replaced the whole attach form with "Promote through the existing chain instead".

This extends the offered list to eight labels, ordered dev → prod:

```
dev, qa, test, uat, staging, demo, sandbox, preprod
```

A fixed list rather than free text, deliberately. The label has to be a usable single-segment git branch name, must not collide with the `wm-fork/**` and `wm_deploy/**` namespaces git-sync already writes, and must not be a repository's default branch. A curated list satisfies all three by construction, where free text needed a slug validator, a reserved-name check, and a guess at the remote's default branch to get the same guarantees.

The UI stays a picker. The only change is that it now has eight entries instead of a two-way toggle.

## Changes

- `DEV_WORKSPACE_LABELS` in `workspaces.rs` is the authoritative list; `normalize_dev_workspace_label` is a membership check against it. The list's length is also the deepest promotion chain.
- `reject_dev_label_matching_tracked_branch` additionally rejects a label that is the *namespace* of a tracked branch (label `uat` against tracked `uat/foo`). Git stores refs hierarchically, so `refs/heads/uat` cannot exist alongside `refs/heads/uat/foo` and every deploy would fail. This case was already reachable for `dev`/`staging` before this PR.
- The shared `DevWorkspaceLabelPicker` replaces the duplicated label row in the attach form and the create-fork form, offering only labels no dev workspace in the resulting chain already holds.
- Read-side helpers handle any stored label; `dev`, `staging`, `''` and `null` render exactly as before (`staging` still abbreviates to `stg`).
- OpenAPI carries the new enum, so generated clients stay as precise as they were.

## Screenshots

The picker in a chain that already holds `dev`, `staging` and `uat`, so it defaults to the next free label:

picker

Open, showing only the labels still free in that chain:

picker open

## Test plan

- [x] `dev_workspace_label_tests`: every offered label round-trips, the blank/`None` default, and off-list names are refused (`main`, `master`, `wm-fork`, `wm_deploy`, `UAT`, `feature/uat`)
- [x] `tracked_branch_blocks_dev_label`: the same-name and namespace-parent cases, plus the near-misses (`release` vs `release-main`, `uat` vs `pre/uat`)
- [x] `nested_dev_workspace` integration tests (7 passed), extended with a three-deep chain `tw-dev` (`dev`) → `spare` (`staging`) → `e-cand` (`uat`), which fails on `main` at `normalize_dev_workspace_label`
- [x] Manually built `demo → demo-dev (dev) → demo-stg (staging) → demo-uat (uat)` on a local EE backend, through the UI
- [x] The namespace guard verified live: a repo tracking `release/main` rejects the label `release`, while `release-main` attaches
- [x] Read side: badge, "This is a uat workspace paired with staging workspace demo-stg", "Git sync deploys to the uat branch"
- [x] `npm run check` (0 errors)
- [ ] **Not exercised**: a real deploy from a dev workspace on one of the new labels against a git-sync repo. The label reaches the branch name through the same `dev_workspace_branch` call as `dev`/`staging` (pinned by the existing `dev_workspace_label_branch_wins` test), but the branch-creation and auto-pull round trip were not run end to end.


rubenfiszel:
Addressed in 54eb332:

- **[P2] OpenAPI enum vs normalization** — fixed by tightening the backend rather than loosening the schema. `normalize_dev_workspace_label` no longer trims or aliases the empty string; only `None` defaults to `dev`, so the accepted set is now exactly the enum. A unit test pins `" uat "` and `""` as rejected. (The frontend only ever sends a list value or `undefined`, so nothing in-tree relied on the trim.)
- **[P2] stale integration-test comment** — rewritten to describe what the test pins now (a chain three deep), with the assert message renamed to match.
- **[P2, Claude] picker showed `stg`** — the dropdown now names labels verbatim. `devBadgeText` is the badge's abbreviation; offering `stg` for a branch called `staging` named something that does not exist.

**[P1] unpinned default branch — not implementing, reasoning below.**

The concern is real but I think the remedy is disproportionate, and I would rather say so than keep circling it.

The two proposed fixes are (a) resolve the remote's actual default branch, or (b) reject the pairing when it cannot be determined. (a) puts an outbound call with repository credentials inside a workspace-settings write, which is a new latency and a new failure mode for self-hosted instances behind a VPN or on GHES, and it is unresolvable for non-GitHub remotes. (b) rejects dev-workspace creation for every branchless repository, which breaks a configuration that works today — including the plain `dev` pairing that has always worked there.

On exposure: `dev` and `staging` were already offered before this PR and carry exactly this risk unchanged; a repository whose unpinned default is `dev` failed identically on `main`. What this PR newly exposes is `qa`, `test`, `uat`, `demo`, `sandbox`, `preprod` as default-branch names, and a repository defaulting to any of those is unusual. The failure is also not silent or destructive: the CLI refuses the deploy with `Fork branch 'X' equals the checked-out branch 'Y'; refusing to deploy a fork directly to the tracked branch. Use a different dev workspace label or tracked branch.` — which names the fix.

The guard that checks facts we actually have stays and was strengthened in this PR: a pinned tracked branch is rejected both when it equals the label and when the label is its namespace parent (`uat` against tracked `uat/foo`), which was a latent hole for `dev`/`staging` too.

Happy to reopen this if someone hits it in practice, or if there is a way to resolve the default branch offline that I have missed.

PR #10572, opened 2026-08-06
fix(cli): sync push crashed on edited fileset children; reject non-canonical fileset dirs

hugocasa:
## Summary

`wmill sync push` mishandled fileset resources (`is_fileset` resource types with `!inline_fileset` values) in two independent ways, together capable of erasing or deploying stale fileset content:

1. **Any push containing an *edited* fileset child crashed the CLI** — even in the canonical layout. In the apply loop's `edited` branch, `handleScriptMetadata`/`handleFile` ran before the `isFileResource`/`isFilesetResource` routing, so a child like `<resource>.fileset/query.sql` was treated as a standalone script: bare `.sql` has no script language, so `inferContentTypeFromFilePath` threw `Invalid language: .sql` and aborted the whole push mid-apply (a push with only edits deployed **nothing**). Children with `.ts`/`.py` extensions were silently deployed as standalone scripts instead. The same unguarded pattern in `generate-metadata` made `sync pull` warn and exit non-zero on such children.

2. **A non-canonical `!inline_fileset` pointer was silently half-accepted.** The diff engine only round-trips the canonical layout (`<resource-path>.fileset/` next to the resource yaml — what `sync pull` generates), but `pushResource` resolved any directory the yaml named, so the first push worked. Every later sync then planned full delete/re-add churn for all children, silently dropped adds under the custom directory (parent lookup fails → skipped without warning), and — when a stale copy of the canonical directory existed in the repo (e.g. committed `sync pull` output) — deployed stale content, resurrecting deleted files and erasing current ones.

Both were reproduced end-to-end against a dev backend before fixing; the fixes were verified the same way (transcript below).

## Changes

- `sync.ts` apply loop, `edited` branch: route `isFileResource`/`isFilesetResource` paths to their parent resource **before** the script handlers (matching the `added`/`deleted` branches), and turn the already-synced fall-throughs into `continue` + state write — the parent's push this run already resolved current child content from disk, and falling through ran `parseFromPath`/`pushObj` on a raw child file.
- `handleFile` returns `false` for file/fileset resource content files (defense in depth for every caller, incl. `wmill dev`); the standalone `wmill script push` now errors explicitly on such paths instead of no-op'ing with a success message.
- `generate-metadata`: exclude file/fileset resource content files from script candidacy (both walk paths), fixing spurious `Skipping …: Invalid language` warnings and non-zero `sync pull` exits.
- New `validateFilesetPointer` in `resource.ts`: **sync pushes** reject a pointer that isn't the **server-canonical** `<resource-path>.fileset` location with an actionable error (`git mv … && update the !inline_fileset value`). This is the only location the remote exporter renders children at — including for workspace-specific resources, so a `.ws_<branch>.fileset` directory is rejected too. The validation runs as a **fail-fast pre-flight** before any change is applied (deletes run first in the apply loop, so a mid-apply abort would leave a partial deploy); `resolveInlineContent` keeps a backstop check for the re-push paths. The standalone `wmill resource push` intentionally keeps accepting any pointer — it pushes one explicit file and no sync diff is involved.
- `isWorkspaceSpecificFile` returns `false` for fileset children. Its patterns match the workspace-name segment with `[^.]+`, and `[^.]` spans `/`, so in `<res>.fileset/edge/q.resource.file.sql` the run `fileset/edge/q` was read as a workspace name: the child was taken for another workspace's file, dropped by `elementsToMap`, and **silently never deployed** — no diff entry, no warning. Only the parent's metadata file carries the workspace suffix; children stay in the shared canonical directory.
- Routing precedence: the `edited` branch checks `isFilesetResource` before `isFileResource` (as `added`/`deleted` already did) — a fileset accepts arbitrary child names, so `<res>.fileset/q.resource.file.sql` satisfies both predicates and belongs to its fileset parent. The pointer pre-flight likewise excludes fileset children before treating a path as resource metadata, since a child may itself be named `*.resource.yaml`.
- New unit tests: `cli/test/fileset_sync_unit.test.ts` (routing guard + pointer validation, incl. normalization and workspace-specific rejection, and the fileset-child classification guard).

## Behavior change note (for the changelog)

`wmill sync push` now **fails with an explicit error** (also surfaced by `--dry-run`, all violations listed at once, before anything is applied) when a resource's `!inline_fileset` value points anywhere other than the server-canonical `<resource-path>.fileset`. Repos that relied on a custom fileset directory location were silently corrupting their sync (full delete/re-add churn, dropped adds, stale content) — they now get a clear message with the exact `git mv` to run. No action needed for repos using the layout `wmill sync pull` generates.

A consequence of the server-canonical contract: workspace-specific (branch-specific) resources get their own `<base>.<workspace>.resource.yaml` metadata file, but their fileset **children** always live in the shared `<base>.fileset/` directory — per-workspace fileset *content* is not supported (the remote exporter renders one canonical location for all workspaces). `findFilesetResourceFile` now resolves the workspace-suffixed metadata file for children in that canonical directory, so editing a child of a workspace-specific fileset pushes correctly.

## Demo (patched CLI against a dev backend)

```text
$ tree f
f
└── resources
    ├── sales_queries.fileset
    │   └── reports
    │       └── query_b.sql
    └── sales_queries.resource.yaml

$ # edit one child, add another, then push
$ wmill sync push --yes --plain-secrets
remote (admins) <- local: 3 changes to apply
~ resource f/resources/sales_queries.fileset/reports/query_b.sql
+ resource f/resources/sales_queries.fileset/reports/query_d.sql
Done! All 3 changes pushed to the remote workspace admins (26ms)
   # previously: crash — "Error: Invalid language: .sql", nothing deployed

$ # a non-canonical '!inline_fileset' pointer now fails fast with a clear error
$ wmill sync push --yes --plain-secrets
error: Resource f/resources/sales_queries uses '!inline_fileset f/sales.fileset', but a fileset
directory must live next to its resource file, at 'f/resources/sales_queries.fileset'. Move the
directory there (e.g. 'git mv f/sales.fileset f/resources/sales_queries.fileset') and update the
'!inline_fileset' value to match.
```

(Resource/file names in the demo are illustrative; the recorded run used a local scratch repo.)

## Test plan

- [x] `UNIT_ONLY=1 bun test test/*_unit*` — 1055 pass (8 new)
- [x] Manual: fileset children named `q.resource.file.sql` / `inner.resource.yaml` now appear in the diff, deploy on add, and update on edit (previously absent from the plan entirely); an `inner.resource.yaml` carrying an `!inline_fileset` value pushes as ordinary content instead of aborting the push
- [x] Manual, canonical layout vs dev backend: push with two edited children succeeds and updates the remote value (previously crashed with nothing deployed); deleted child is removed from the value; follow-up push plans no fileset changes
- [x] Manual: `sync pull` of a workspace with a fileset resource exits 0 with no `Invalid language` warnings (previously exit 1)
- [x] Manual: sync push with a non-canonical pointer fails fast with the actionable message; standalone `wmill resource push` with the same pointer still succeeds
- [x] Manual: `wmill script push <fileset child>` errors explicitly instead of reporting a false success
- [x] `bunx tsc --noEmit` introduces no new errors (pre-existing failures identical on main)

## Out of scope

- **Workspace-specific fileset diff mapping** (pre-existing): `elementsToMap` drops `<base>.fileset/*` children from local diffs when the parent is flagged workspace-specific, so a clean ws-specific fileset still plans full delete/re-add churn (the parent re-push keeps content correct, but plans are never clean). The fileset and specific-items features never composed in the diff engine; this PR fixes the crash, parent resolution, and metadata precedence for that combination, and the map handling is left as a follow-up rather than growing this fix into diff-engine surgery.

- Resource types with `format_extension: null` diff as edited on every push (unrelated churn observed while reproducing).
- `specific_items.ts` workspace-specific matching assumes canonical fileset paths — consistent with this PR's contract.

---
## Summary by cubic
Fixes `wmill sync push` for fileset resources by routing edited children to their parent and rejecting non‑canonical `!inline_fileset` pointers to prevent crashes, stale deploys, and dropped diffs (WIN‑2338).
Sync now validates pointers before dry‑run, prefers workspace‑specific metadata, and excludes fileset children from both workspace-specific and current‑workspace classifiers.

- **Bug Fixes**
  - Route fileset and single‑file resource content to their parent before script handlers; `wmill script push <fileset child>` now errors.
  - Exclude file/fileset content from `generate-metadata` script detection to remove "Invalid language" warnings and non‑zero `sync pull` exits.
  - Fileset child handling: never treat as workspace‑specific or current‑workspace to avoid drops/remaps; prefer workspace‑specific metadata over base; metadata lookup is platform‑separator safe.

- **Migration**
  - Sync pushes now fail fast if `!inline_fileset` is not `<resource-path>.fileset`. Move the directory next to the resource file and update the pointer. Standalone `wmill resource push` still accepts custom pointers.

<sup>Written for commit c39456beb1c6e4e3d6357169c3d417ab2103b3b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10572?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/src/commands/generate-metadata/generate-metadata.ts
+++ b/cli/src/commands/generate-metadata/generate-metadata.ts
@@ -3,6 +3,7 @@ import { Confirm } from "@cliffy/prompt/confirm";
 import { colors } from "@cliffy/ansi/colors";
 import { sep as SEP } from "node:path";
 import { GlobalOptions, isDatatableMigrationPath } from "../../types.ts";
+import { isFileResource, isFilesetResource } from "../../utils/utils.ts";
 import { SyncOptions, mergeConfigWithConfigFile } from "../../core/conf.ts";
 import { resolveWorkspace } from "../../core/context.ts";
 import { requireLogin } from "../../core/auth.ts";
@@ -56,6 +57,9 @@ async function walkLocalScripts(
       isFolderResourcePathAnyFormat(p) ||
       // Datatable migration `.sql` files aren't Windmill scripts.
       isDatatableMigrationPath(p) ||
+      // Neither are file/fileset resource content files (.sql, .ts, …).
+      isFileResource(p) ||
+      isFilesetResource(p) ||
       (isScriptModulePath(p) && !isModuleEntryPoint(p)),
     false,
     {},
@@ -225,6 +229,9 @@ function categorizeLocalFiles(
       !isFolderResourcePathAnyFormat(p) &&
       // Datatable migration `.sql` files aren't Windmill scripts.
       !isDatatableMigrationPath(p) &&
+      // Neither are file/fileset resource content files (.sql, .ts, …).
+      !isFileResource(p) &&
+      !isFilesetResource(p) &&
       !(isScriptModulePath(p) && !isModuleEntryPoint(p))
     ) {
       scripts.push(p);
--- a/cli/src/commands/resource/resource.ts
+++ b/cli/src/commands/resource/resource.ts
@@ -47,34 +47,22 @@ async function readFilesetDirectory(dirPath: string): Promise<Record<string, str
 }
 
 /**
- * A fileset directory must live next to its resource file at
+ * A fileset directory must live at the server-canonical location
  * `<resource path>.fileset` — that is the only layout the sync diff engine
- * can round-trip. Any other pointer breaks change detection: children are
- * planned as full delete/re-add churn and adds under the custom directory
- * are dropped, which manifests as erased or stale fileset content.
+ * can round-trip (remote state is always rendered there, including for
+ * workspace-specific resources). Any other pointer breaks change detection:
+ * children are planned as full delete/re-add churn and adds under the custom
+ * directory are dropped, which manifests as erased or stale fileset content.
  */
 export function validateFilesetPointer(
   dirPath: string,
   remotePath: string,
-  originalLocalPath: string | undefined,
 ): void {
   const normalize = (p: string) =>
     p.replaceAll("\\", "/").replace(/\/+$/, "");
   const pointer = normalize(dirPath);
-  const canonicalFromRemote =
-    normalize(remotePath.replaceAll(SEP, "/")) + ".fileset";
-  // For branch/workspace-specific metadata files the local file name differs
-  // from the server path, so the directory adjacent to the local file is also
-  // canonical.
-  const canonicalFromLocal = originalLocalPath
-    ? normalize(
-        originalLocalPath
-          .replaceAll(SEP, "/")
-          .replace(/\.resource\.(yaml|json)$/, ""),
-      ) + ".fileset"
-    : undefined;
-  if (pointer !== canonicalFromRemote && pointer !== canonicalFromLocal) {
-    const expected = canonicalFromLocal ?? canonicalFromRemote;
+  const expected = normalize(remotePath.replaceAll(SEP, "/")) + ".fileset";
+  if (pointer !== expected) {
     throw new Error(
       `Resource ${remotePath.replaceAll(SEP, "/")} uses '!inline_fileset ${dirPath}', ` +
         `but a fileset directory must live next to its resource file, at '${expected}'. ` +
@@ -111,7 +99,7 @@ export async function pushResource(
     if (typeof localResource.value === "string" && localResource.value.startsWith("!inline_fileset ")) {
       const dirPath = localResource.value.split(" ")[1];
       if (enforceCanonicalFileset) {
-        validateFilesetPointer(dirPath, remotePath, originalLocalPath);
+        validateFilesetPointer(dirPath, remotePath);
       }
       localResource.value = await readFilesetDirectory(dirPath.replaceAll("/", SEP));
     } else if (localResource.value["content"]?.startsWith("!inline ")) {
--- a/cli/src/commands/resource/resource.ts
+++ b/cli/src/commands/resource/resource.ts
@@ -91,6 +91,10 @@ export async function pushResource(
   localResource: ResourceFile,
   originalLocalPath?: string,
   wsSpecific?: boolean,
+  // Sync pushes reject non-canonical fileset pointers (the diff engine can
+  // only round-trip the canonical layout); the standalone `resource push`
+  // command pushes a single explicit file, where any pointer is fine.
+  enforceCanonicalFileset?: boolean,
 ): Promise<void> {
   remotePath = removeType(remotePath, "resource");
   try {
@@ -106,7 +110,9 @@ export async function pushResource(
   const resolveInlineContent = async () => {
     if (typeof localResource.value === "string" && localResource.value.startsWith("!inline_fileset ")) {
       const dirPath = localResource.value.split(" ")[1];
-      validateFilesetPointer(dirPath, remotePath, originalLocalPath);
+      if (enforceCanonicalFileset) {
+        validateFilesetPointer(dirPath, remotePath, originalLocalPath);
+      }
       localResource.value = await readFilesetDirectory(dirPath.replaceAll("/", SEP));
     } else if (localResource.value["content"]?.startsWith("!inline ")) {
       const basePath = localResource.value["content"].split(" ")[1];
--- a/cli/src/commands/resource/resource.ts
+++ b/cli/src/commands/resource/resource.ts
@@ -46,6 +46,44 @@ async function readFilesetDirectory(dirPath: string): Promise<Record<string, str
   return result;
 }
 
+/**
+ * A fileset directory must live next to its resource file at
+ * `<resource path>.fileset` — that is the only layout the sync diff engine
+ * can round-trip. Any other pointer breaks change detection: children are
+ * planned as full delete/re-add churn and adds under the custom directory
+ * are dropped, which manifests as erased or stale fileset content.
+ */
+export function validateFilesetPointer(
+  dirPath: string,
+  remotePath: string,
+  originalLocalPath: string | undefined,
+): void {
+  const normalize = (p: string) =>
+    p.replaceAll("\\", "/").replace(/\/+$/, "");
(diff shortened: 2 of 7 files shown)


hugocasa:
Round-3 findings addressed in b6b3603d3b:

- **Metadata precedence** (Codex P1b / Claude nit): `findFilesetResourceFile` now puts the workspace-suffixed candidates first, matching `findResourceFile` — the suffixed file is the workspace's authoritative metadata and wins over a coexisting base file. Unit-tested both directions.

- **Workspace-specific fileset diff mapping** (Codex P1a): deferred as a documented pre-existing limitation (see "Out of scope" in the PR body). `elementsToMap` dropping ws-specific `<base>.fileset/*` children from local diffs predates this PR — the fileset and specific-items features never composed in the diff engine. This PR already fixes the crash, parent resolution, and metadata precedence for that combination; content stays correct via the parent re-push, only the plan output churns. Extending `elementsToMap` is diff-engine surgery with regression risk for mainline sync, out of proportion for a combination the motivating incident doesn't use — follow-up material.

hugocasa:
Re: [P1] "Preserve canonical fileset children in workspace-specific diffs" (`elementsToMap`) — **deferring this, confirmed with maintainer.**

The gap predates this PR: the fileset and specific-items features never composed in the diff engine, so a workspace-specific fileset has always planned delete/re-add churn for its children. Content stays correct (each child change re-pushes the parent, which reads current disk state), only the plan output is noisy. Extending `elementsToMap` to map shared canonical children for workspace-specific parents is diff-engine surgery with regression risk for mainline sync, and the motivating incident does not involve workspace-specific resources — this PR keeps to the fixes that combination needs to be *safe* (no crash, correct parent resolution, workspace-suffixed metadata precedence) and leaves the map handling to a follow-up. Documented under "Out of scope" in the PR body.

Also fixed in cb566ac97d: the new metadata-lookup test asserted forward-slash paths and failed on the Windows CI runner; assertions now use the platform separator.

hugocasa:
Round-4 findings addressed in b4e3cfe415, plus a root cause neither reviewer reached.

**Both P1s were correct about the code but unreachable as described.** I built the exact scenario against a dev backend — a child at `<res>.fileset/edge/q.resource.file.sql` — and it never entered the diff at all: not as an add, not as an edit. Same for a child named `inner.resource.yaml`.

**Root cause:** `isWorkspaceSpecificFile` (`specific_items.ts:466`) builds its patterns with `[^.]+` for the workspace-name segment, and `[^.]` matches `/`. In `f/resources/data.fileset/edge/q.resource.file.sql` the run `fileset/edge/q` is therefore read as a workspace name, the child is classified as belonging to another workspace, and `elementsToMap` drops it. These children were **silently never deployed** — no diff entry, no warning, no error. Fixed by returning `false` for fileset children, which is the contract this PR establishes: only the parent's metadata file carries the workspace suffix, children stay in the shared canonical directory.

That fix is what makes the two reported findings live, so both are also fixed:

- **`.resource.file.*` routing** (`sync.ts:5128`): the `edited` branch now checks `isFilesetResource` before `isFileResource`, matching the `added` and `deleted` branches. Verified: editing `edge/q.resource.file.sql` now updates the parent resource's value on the remote.
- **Pre-flight metadata selection** (`sync.ts:4767`): fileset children are excluded before pointer validation. Confirmed load-bearing by removing the guard — an `inner.resource.yaml` whose body contains an `!inline_fileset` value then aborts the whole push (and `--dry-run`) with an error naming `…fileset/edge/inner.fileset`. With the guard it pushes as ordinary fileset content.

Remote state after add-then-edit of all three children:

```
edge/inner.resource.yaml  => "resource_type: c_files\nvalue: '!inline_fileset f/somewhere/else.fileset'\n"
edge/plain.sql            => 'select 2 as v;\n'
edge/q.resource.file.sql  => 'select 1 as v; -- EDITED\n'
```

New test pins that fileset children are never classified workspace-specific while the parent's suffixed metadata still is. 1055 unit tests pass; `tsc` shows only the two pre-existing JSZip errors in `sync.ts`.

hugocasa:
Round-5 P1 confirmed and fixed in c39456beb1.

Codex was right that the guard needed to cover `isCurrentWorkspaceFile` too, and the consequence there is worse than a drop. Reproduced against a dev backend with workspace `admins`: adding a child at `<res>.fileset/edge/inner.admins.resource.yaml` produced this plan —

```
~ resource f/.../data.fileset/edge/inner.resource.yaml     <- NOT edited locally
~ resource f/.../data.fileset/edge/q.resource.file.sql
```

The new child never appears under its own key. `elementsToMap` takes the `isCurrentWorkspaceFile` branch, remaps it to `fromWorkspaceSpecificPath(...)` = `edge/inner.resource.yaml`, and deploys its content over that untouched sibling.

With the exemption applied to both classifiers, the same repo plans and pushes correctly:

```
+ resource f/.../data.fileset/edge/inner.admins.resource.yaml
```

```
edge/inner.admins.resource.yaml => 'select 9 as v;\n'
edge/inner.resource.yaml        => "resource_type: c_files\nvalue: '!inline_fileset f/somewhere/other.fileset'\n"
edge/plain.sql                  => 'select 2 as v;\n'
edge/q.resource.file.sql        => 'select 1 as v; -- EDITED2\n'
```

Each child lands under its exact key, the sibling is intact, and the follow-up dry run is clean apart from the pre-existing `format_extension: null` resource-type churn noted under "Out of scope".

Test extended to cover `isCurrentWorkspaceFile` in both directions: a fileset child carrying the active workspace suffix is not current-workspace, while the parent's own suffixed metadata still is. 1055 pass.

PR #10573, opened 2026-08-06
feat: add public sharing option for job pages

rubenfiszel:
## Summary

Fixes WIN-2331.

The run page's **Share** button already mints a stateless read-only link, but it only works
for a logged-in member of the workspace. This adds a second sub-option that mints a *public*
flavor of the same token: anyone holding the link, logged in or not, lands on a minimal
read-only run page.

The public token is a **distinct HMAC domain** (`public_view_token`) rather than a reuse of
the existing one, so links already handed out keep granting only what they were minted for —
going public is always an explicit new mint, never a retroactive widening.

## Changes

**Backend**
- `generate_view_token` now takes a domain separator; `VIEW_TOKEN_DOMAIN` (member) and
  `PUBLIC_VIEW_TOKEN_DOMAIN` (public) are exported from `windmill-common::variables`.
- New `GET /w/{workspace}/jobs/job_public_view_token/{id}` (`getJobPublicViewToken`), gated by
  the same read access as the member flavor — whoever can read the job may publish it.
- `validate_view_token` returns the token's `ViewTokenAudience` instead of a bool. Either
  audience satisfies the authenticated ACL check (public is a superset of member); only
  `Public` satisfies the logged-out check.
- The ~14 copies of the logged-out gate (`else if created_by != "anonymous" { … }`) collapse
  into `require_opt_authed_job_read_access` / `require_unauthed_job_read_access`, which add
  the public-token grant uniformly across every `jobs_u` read handler that already accepted a
  `view_token` (get, args, logs, log file, results, timing, flow-tree logs/results, dispatch
  events, otel traces, job updates + SSE).
- `GetQuery` gained `with_public_view_grant`, consulted by `check_auth` alongside the existing
  logged-out restriction.
- The SSE stream's per-stream `anonymous_verified` latch is renamed
  `unauthed_read_authorized` and can be seeded by a valid public token at connection time.

**Frontend**
- The run page's Share button becomes a dropdown: *Copy link for members* (existing behavior)
  and *Copy public link*.
- New `/public_run/{workspace}/{id}?view_token=…` route outside the logged layout: status,
  inputs, and — for scripts — result + logs; for flows the full `FlowStatusViewer` (graph,
  timeline, per-step inputs/result/logs, live updates).

## What a public link does *not* grant

Cancel/force-cancel, resume, minting further tokens, and `get_flow_debug_info` are all
unchanged: they either require a session or are confined to anonymously-created jobs. A public
token is also still scoped to its job's flow subtree, and a tag-scoped caller cannot use one to
escape its tag filter.

## Screenshots

Share sub-menu on the run page:

share-dropdown

Public run page for a completed flow (logged out) — full flow viewer renders:

public-flow

Selecting a step in that graph loads its inputs, result and logs:

public-flow-step

A *running* flow updates live over SSE for the anonymous viewer:

public-flow-running

A script run:

public-script

## Test plan

- [x] `cargo test --test jobs_read_auth` — extended with a public-token block: a member token
      must NOT grant unauthenticated read; a public token grants the shared job and its whole
      flow subtree logged out; an out-of-scope or forged public token does not; minting the
      public flavor needs the same read access as the member one.
- [x] Manually, against the dev stack: minted both flavors for a flow run and a script run and
      probed the whole `jobs_u` read surface anonymously — public token 200s on all of it
      (including a nested step job), member token and no token both 400.
- [x] Anonymous `queue/cancel` of a running non-anonymous job with a valid public token is
      still refused, and the job kept running.
- [x] Drove the UI logged out with Playwright: completed flow, running flow (live SSE), script
      run, step selection, flow Logs tab, and the two failure paths (no token / tampered token)
      — no console errors, bounded retries.
- [x] The member share link (`/run/{id}?view_token=…`) still works for a logged-in member.

---
## Summary by cubic
Adds a public sharing option for run pages so anyone with the link can view a minimal, read-only run (flows update live over SSE). Addresses WIN-2331; public sharing is gated by a workspace rule and audited, and existing member-only links remain unchanged via a distinct HMAC domain.

- New Features
  - New endpoint: GET `/w/{workspace}/jobs/job_public_view_token/{id}` to mint public read-only tokens.
  - Share button is now a dropdown: “Copy link for members” and “Copy public link” (public option disabled unless admin or bypass user under the rule).
  - New public route: `/public_run/{workspace}/{id}?view_token=…` with status, inputs, result/logs, and a live flow viewer; page sets `noindex` and `no-referrer`.
  - Workspace rule `RestrictPublicRunSharing` to limit who can mint public links; surfaced in `RulesetEditor`, enforced server-side, and audited as `jobs.share_publicly`.
  - Public tokens cannot be minted by tag-scoped callers.
  - Flow subflow expansion now requires login; share-link viewers get a clear message.
  - Public run view is keyed on workspace, job id, and token to avoid stale or cross-link content during client-side navigation.
  - OpenAPI documented under `getJobPublicViewToken`.
  - Tests cover audience scope, rule enforcement and bypass, tag-scoped refusal, and unauthenticated reads.

- Refactors
  - `generate_view_token` now takes a domain; added `VIEW_TOKEN_DOMAIN` and `PUBLIC_VIEW_TOKEN_DOMAIN`.
  - `validate_view_token` returns the token audience (Member/Public) and supports subtree access.
  - Consolidated logged-out read gates into `require_unauthed_job_read_access` and `require_opt_authed_job_read_access`.
  - `GetQuery` gains `with_public_view_grant` to lift logged-out restrictions when a valid public token is present.
  - SSE gate renamed to `unauthed_read_authorized` and can be seeded by a public token at connection time.

<sup>Written for commit ee8bcccd9bcf412d8d1fb23f98504661dcf2f796. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10573?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/jobs_read_auth.rs
+++ b/backend/tests/jobs_read_auth.rs
@@ -30,7 +30,6 @@
 //!     the URL.
 
 use sqlx::{Pool, Postgres};
-use windmill_common::workspaces::invalidate_protection_rules_cache;
 use windmill_test_utils::*;
 
 const VICTIM: &str = "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa";
@@ -654,10 +653,6 @@ async fn test_single_job_read_authorization(db: Pool<Postgres>) -> anyhow::Resul
         status.is_success(),
         "admin should create the protection rule (got {status}): {body}"
     );
-    // Process-wide cache with a TTL: the earlier mints above populated it with an empty
-    // ruleset list, so drop it rather than wait the TTL out.
-    invalidate_protection_rules_cache("test-workspace");
-
     let (status, body) = get(
         &authed_base,
         &format!("job_public_view_token/{TOP_SECRET_FLOW}"),
@@ -690,11 +685,10 @@ async fn test_single_job_read_authorization(db: Pool<Postgres>) -> anyhow::Resul
         "an admin must bypass the rule (got {status}): {body}"
     );
 
-    // Later assertions in this test mint public tokens; drop the rule (and the cache
-    // entry it seeded) so they see the same workspace state as before.
+    // Later assertions in this test mint public tokens; drop the rule so they see the
+    // same workspace state as before.
     let (status, _) = delete(&format!("{rules_url}/no-public-runs"), Some("SECRET_TOKEN")).await;
     assert!(status.is_success(), "admin should delete the rule");
-    invalidate_protection_rules_cache("test-workspace");
 
     // ---- TAG-SCOPED token must not mint a token outside its allowed tags ----
     // SCOPED_DENO_TOKEN (test-user-2, scope `if_jobs:filter_tags:deno`) can read both
@@ -834,7 +828,7 @@ async fn test_single_job_read_authorization(db: Pool<Postgres>) -> anyhow::Resul
     )
     .await;
     assert!(
-        !status.is_success() || !body.contains("canceled"),
+        !status.is_success(),
         "a public token must not let an anonymous caller cancel the run (got {status}): {body}"
     );
     // The owner still cancels their own job (no over-blocking). Keep this last: it
--- a/backend/tests/jobs_read_auth.rs
+++ b/backend/tests/jobs_read_auth.rs
@@ -30,6 +30,7 @@
 //!     the URL.
 
 use sqlx::{Pool, Postgres};
+use windmill_common::workspaces::invalidate_protection_rules_cache;
 use windmill_test_utils::*;
 
 const VICTIM: &str = "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa";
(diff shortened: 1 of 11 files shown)


rubenfiszel:
✅ Review round clean @ ee8bcccd9bcf412d8d1fb23f98504661dcf2f796 — nit-only verdicts at ad9ed570de7ab5943bd826faa321c4c487dc2735 (codex + claude "mergeable, address nits", pi "good to merge"); nits addressed in ee8bcccd9b

Round 3 nits, all fixed:

- **Codex P2, stale view when only the workspace or token changes.** Correct — both guards were keyed on `jobId` alone, so the same UUID reached with a different workspace or `view_token` never reset. They now key on a `viewKey` of `workspace|jobId|token`. Verified the exact scenario in a logged-out browser: from a valid link to the *same* job UUID with the token removed, the run clears and the page shows "This run is not available"; navigating back with the token restores it.
- **Claude P2, vacuous anonymous-cancel assertion.** Correct — `cancel_job_api` returns the bare UUID on success, so `!body.contains("canceled")` was always true and the case would have passed through a regression. Tightened to `assert!(!status.is_success(), …)`.
- **Claude minor, cache-invalidation comment asserting a necessity that doesn't hold.** Confirmed `create_protection_rule` / the delete handler already invalidate (`windmill-api-workspaces/src/workspaces.rs:9347,9468`), so the two `invalidate_protection_rules_cache` calls were noise rather than belt-and-braces — dropped them and the comment instead of rewording.

One correction to the round-3 review: the deleted-job failure path *was* exercised, before that round. Deleting a completed flow job's `v2_job`/`v2_job_completed` rows and opening its public link lands on the "This run is not available" alert, both on direct load and on client-side navigation from a valid link — the concern about `JobLoader` being unmounted doesn't bite, because the initial load runs while `job` is still `undefined`, so the loader is mounted and sets `notfound`. The rule end-to-end in the UI was also driven with a real non-admin member: *Copy public link* is greyed out and its wrapper carries the explanatory `title`.

Still open by design, on the record: a public link cannot be revoked (documented in the description and the tooltip; a per-job or per-workspace mint epoch in the HMAC input is the follow-up).

PR #10574, opened 2026-08-06
feat: version history for session artifacts

AlexRV12:
## Summary

Every content change to an AI-session artifact is now kept as a version. The AI can list and read earlier versions, and the reader can pick one from a dropdown next to the artifact's title.

Fixes WIN-2325.

Rebased onto `origin/main` at `d638fab5f6`.

The UI was settled by exploring options first — twelve directions for the picker, then the chosen dropdown in detail, then the popover hint — and then refined against real usage.

## Changes

**Storage** (`artifactsDB.ts`)
- New `versions` object store (DB `1 → 2`), keyed `${artifactId}:${version}`, indexed by artifact.
- History is capped at 20 snapshots per artifact; pruning orders by parsed version number, since keys sort lexicographically and would otherwise drop `v9` before `v10`.
- Deleting an artifact — or a whole session — drops its versions in the same transaction.
- An artifact and the snapshots an edit produces are written in one `['items', 'versions']` transaction. Split across two writes, a row stamped v3 can outlive the v3 snapshot that failed to land beside it — and that state still *reads* as complete, because the current version is synthesized from the row, right up until the next edit overwrites the only copy of that content.

**Version rules** (`artifactsState.svelte.ts`)
- `create` writes v1; `update` bumps and snapshots **only when the content actually changed**, so a rename or an identical rewrite doesn't fill the picker with entries nobody can tell apart.
- Artifacts written before this change read as their own v1, and their original text is back-filled on *any* update — a rename stamps `version`, after which nothing would recognise them as pre-history.

**AI surface** (`artifactTools.ts`, `global/core.ts`)
- New `list_artifact_versions(id)`; `read_artifact` takes an optional `version`.
- `update_artifact` gains a required `change_note` — a short label the reader sees in the picker. A missing field throws in zod, which `processToolCall` returns to the model as a tool error, so it costs one retry and never loses the content. The stored note is trimmed and capped at 120 characters rather than schema-rejected, so an overlong label never discards a real content update.
- Reading a pruned version reports back pointing at `list_artifact_versions` rather than failing blankly.
- Both version reads are scoped to the calling session, as `update_artifact` already was, so an artifact id that crosses sessions cannot pull back another session's content. `list_artifact_versions` also joins `SESSION_PREVIEW_TOOL_NAMES`, which withholds it from the non-session global chat.

**Viewer** (`ArtifactVersionPicker.svelte` — new, `ArtifactViewer.svelte`)
- A version chip next to the title opens the history. Rendered **only above one version**, so an artifact the AI wrote once looks exactly as it did before.
- Rows are note-led and two-line: the change note gets the full width and up to two lines, with `v6 · 8m ago` on a quieter second line. One truncated line was not enough to tell two edits apart.
- The selected version is marked with a row background (`surface-accent-selected`), following `ScriptVersionHistory` and the rest of the codebase, rather than a small glyph.
- The menu header only says something when there is something to say — `20 most recent of 25 versions` once pruning has dropped history, otherwise just `Versions`. Version numbers run `1..latest`, so a plain count restates the rows.
- While pinned, the chip turns amber and a bar under the header states which version and how old, with "Back to latest". Copy, download, preview and source all follow the selected version.
- An explicit selection survives later AI edits: writing v8 does not yank a reader out of the v3 they chose.

**Artifacts popover** (`ArtifactsSegment.svelte`)
- Rows for artifacts with history read `v7 · 2m`. The version rides in the timestamp's already-reserved width, paid for by a new compact `TimeAgo` mode, so the kind badge keeps its slot (`ArtifactKind` is already `'md' | 'html'` and will grow) and the name loses none of its own.

**Shared** (`TimeAgo.svelte`, `userScopedDb.ts`)
- New `compact` mode implementing the `s/m/h/d` ladder the hand-rolled `ago()` helpers already use (`PipelineEventLog`, `PipelineActivityPanel`, `IndexerMemorySettings`).
- `userScopedDb`'s open lifecycle is rebuilt around one `Attempt` record per scoped database, replaced only once it has settled. An IndexedDB open cannot be cancelled and the browser queues opens per database, so discarding an in-flight one leaves it parked in that queue and any replacement waits behind it — never reaching its own `blocked` callback, and so never arming the timeout meant to save it. Three review rounds each found a different route into that hang (repeat call, user switch, `close()`); the record makes all of them structurally unreachable, and a test now pins the invariant itself by asserting a single open is issued however often callers give up. On top of that it gains `blocking`, `blocked` and `terminated` handlers plus a bounded wait for a blocked upgrade. This is the first version bump of `copilot-artifacts`, so it is the first time the failure can happen: a blocked open never settles, and nothing downstream has a timeout, so the upgrading tab would hang on `whenReady()` forever — leaving the artifacts panel spinning and an in-flight AI tool call unable to return. `blocking` yields the connection for every bump from here on; the grace period covers the tabs that predate it, since those hold v1 open without listening for `versionchange` and no amount of cooperation on this side can move them. The chat-history store shares this helper and is protected by the same fix.

## Screenshots

Capture d’écran 2026-08-10 à 12 13 48
Capture d’écran 2026-08-10 à 12 14 05
Capture d’écran 2026-08-10 à 12 14 26
Capture d’écran 2026-08-10 à 12 14 42
Capture d’écran 2026-08-10 à 12 17 03
Capture d’écran 2026-08-10 à 12 17 14

## Test plan

- [x] `npx vitest run src/lib/userScopedDb.test.ts src/lib/components/copilot/chat/artifacts/` — 59 passing
- [x] Every other `userScopedDb` consumer, since its lifecycle was rewritten: `HistoryManager.test.ts` + `src/lib/components/sessions/` — 227 passing, none of them edited. That suite carries one intermittent failure (`sessionStateIndexedDb.test.ts` → "setSessionPreviewSize persists a dragged width and round-trips it", ~1 run in 8); it reproduces with `origin/main`'s `userScopedDb.ts` swapped back in, so it predates this branch.
- [x] Real browser (not just `fake-indexeddb`): artifact and both versions survive a full page reload **and** a logout/re-login cycle, no page errors
- [x] `cd ai_evals && bun test` — 152 passing, 1 skipped
- [x] `npm run check` — 6 errors / 70 warnings, none in a file this branch touches: `Login.svelte` (whose last commit is the merge-base itself, #10598) plus `DevWorkspaceSetting.svelte`, `CreateWorkspaceInner.svelte` and `run/[...run]/+page.svelte`, all from a stale generated client (`getJobPublicViewToken`, `RestrictPublicRunSharing`, a widened workspace-kind enum)
- [x] Driven in a real browser at every step: chip appears only above one version; picking a version pins the chip, shows the bar and swaps preview *and* source; "Back to latest" restores; a pinned view survives a further AI edit; the popover row reads `v7 · 2m`; pruning verified end-to-end (25 edits → 20 rows, oldest kept is v6); a version landing under an open menu refreshes the rows without closing it; relative times checked at 5s / 59s / 61s / 473s / 3h / 3d; both light and dark themes
- [x] Models reliably fill `change_note` — covered by the `global-artifact-note-on-each-edit` eval case and run against two models
- [x] Manual two-tab check: load `main` in one tab and this branch in another, and confirm the second degrades to in-memory within the grace period instead of hanging (`fake-indexeddb` cannot hold a connection open across builds, so this is the one path the suite cannot reach)

## Bounds

History is capped per artifact at 20 snapshots, and fewer for a large one: retention is sized from the incoming snapshot against a ~1M-character budget, with a floor of 3. A count alone would not bound storage, since 20 snapshots of a max-size artifact is ~5 MB of history for a single document, and exhausting the origin's quota would silently take `putArtifact` down with it. Sizing from the incoming snapshot rather than totalling the stored history keeps pruning on `getAllKeys()`, so no version's content is deserialized on the write path.

Concurrent edits to one artifact from two tabs resolve last-write-wins, and the losing side's snapshot is overwritten along with its note. That is the same resolution the artifact content itself has had since artifacts landed — the AI path cannot hit it within a tab, as both providers dispatch tool calls through a sequential `for … await` loop.

### DIFF
--- a/ai_evals/adapters/frontend/core/global/evalArtifactHelpers.test.ts
+++ b/ai_evals/adapters/frontend/core/global/evalArtifactHelpers.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from "bun:test";
+import { createEvalArtifactHelpers } from "./evalArtifactStore";
+
+// A hand-written stand-in for SessionArtifactsStore (bun has no IndexedDB), so nothing
+// makes it follow that class. A method missing from it surfaces as a tool throwing
+// part-way through an eval run, which reads as a model failure rather than a harness one.
+describe("eval artifact store", () => {
+  it("exposes every method the artifact tools call", () => {
+    const { helpers } = createEvalArtifactHelpers();
+    for (const method of [
+      "create",
+      "get",
+      "update",
+      "remove",
+      "listForSession",
+      "listVersions",
+      "getVersion",
+    ]) {
+      expect(typeof (helpers.artifacts as any)[method]).toBe("function");
+    }
+  });
+});
--- a/ai_evals/adapters/frontend/core/global/evalArtifactStore.ts
+++ b/ai_evals/adapters/frontend/core/global/evalArtifactStore.ts
@@ -1,0 +1,93 @@
+// SessionArtifactsStore can't run here (bun has no IndexedDB, nor the compiled $state runes),
+// so mirror only the shape the artifact tools call, not its scoping or race handling.
+export const EVAL_SESSION_ID = "eval-session";
+export function createEvalArtifactHelpers() {
+  const items = new Map<string, Record<string, any>>();
+  // Snapshots per artifact id, oldest first — the version tools read history from here.
+  const history = new Map<string, Array<Record<string, any>>>();
+  let seq = 0;
+  const snapshotOf = (
+    artifact: Record<string, any>,
+    version: number,
+    note?: string,
+  ) => ({
+    key: `${artifact.id}:${version}`,
+    artifactId: artifact.id,
+    version,
+    name: artifact.name,
+    content: artifact.content,
+    savedAt: artifact.updatedAt,
+    note,
+  });
+  const store = {
+    create: async (sessionId: string, input: Record<string, any>) => {
+      const now = seq++;
+      const artifact = {
+        id: `eval-artifact-${now}`,
+        sessionId,
+        chatId: input.chatId,
+        kind: input.kind ?? "md",
+        name: input.name,
+        content: input.content,
+        createdAt: now,
+        updatedAt: now,
+        version: 1,
+      };
+      items.set(artifact.id, artifact);
+      history.set(artifact.id, [snapshotOf(artifact, 1)]);
+      return artifact;
+    },
+    get: async (id: string) => items.get(id),
+    update: async (
+      id: string,
+      input: Record<string, any>,
+      opts?: { sessionId?: string },
+    ) => {
+      const existing = items.get(id);
+      if (!existing) return undefined;
+      if (
+        opts?.sessionId !== undefined &&
+        existing.sessionId !== opts.sessionId
+      )
+        return undefined;
+      // Only a content change earns a version, as in SessionArtifactsStore.
+      const contentChanged =
+        input.content !== undefined && input.content !== existing.content;
+      const version = (existing.version ?? 1) + (contentChanged ? 1 : 0);
+      const updated = {
+        ...existing,
+        name: input.name ?? existing.name,
+        content: input.content ?? existing.content,
+        updatedAt: seq++,
+        version,
+      };
+      items.set(id, updated);
+      if (contentChanged) {
+        history.set(id, [
+          ...(history.get(id) ?? []),
+          snapshotOf(updated, version, input.note),
+        ]);
+      }
+      return updated;
+    },
+    remove: async (id: string) => {
+      items.delete(id);
+      history.delete(id);
+    },
+    listForSession: async (sessionId: string) =>
+      [...items.values()].filter((a) => a.sessionId === sessionId),
+    listVersions: async (id: string) =>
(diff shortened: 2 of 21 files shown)


AlexRV12:
Both P2 nits addressed in `41ca4c8bbc`.

**Stale compact timestamps** — fixed as described. `compact` was setting `isRecent = false` for hour- and day-based values, and the `onMount` interval clears itself for good on that flag, so a row froze for as long as the list stayed open. Now only the absolute `MM/DD` form (older than a month, and genuinely final) clears the timer; `m` / `h` / `d` keep advancing.

**Unenforced change-note length** — bounded, but by truncation rather than `.max(60)`. A schema rejection fails the whole `update_artifact` call, which means throwing away a real content edit over a cosmetic label and making the model resend the entire document to recover. That is a worse outcome than a long note, especially since the field is required and the picker truncates for display anyway. The note is now trimmed and capped at 120 characters before it is persisted, which bounds what every snapshot stores and what is replayed to the model — the actual concern raised — without putting the content update at risk. Test added for the overlong case asserting the update still succeeds.

PR #10581, opened 2026-08-06
fix: explain duckdb failures caused by job isolation

rubenfiszel:
## Summary

On a worker that isolates jobs (nsjail/unshare), the in-process DuckDB connection has no local filesystem access. Two consequences reached users as raw engine errors naming neither the policy nor a way out:

- DuckDB spills to disk through that same filesystem, so a query whose working set outgrew `memory_limit` (derived from the worker's cgroup) failed with `Permission Error: File system LocalFileSystem has been disabled by configuration` rather than an out-of-memory error;
- reading or writing a local file failed with that identical message.

## Changes

- Clear `temp_directory` alongside the other confinement settings, so an over-limit query fails as an out-of-memory error instead of a filesystem permission error (spilling is impossible either way).
- Append an explanation plus the `DISABLE_DUCKDB_FS_SANDBOX=true` opt-out to both failures. Hooked at `do_duckdb`'s central error funnel, so the prepare/typecheck path is covered too; the structured data-test payload keeps its variant.
- The `on_schema_change=sync` pre-pass opens its own DuckDB connection, so it now gets the same connection policy as the main query.

Public diff is the two gated call sites plus the EE ref; the logic lives in the EE companion.

## Test plan

- [x] `cargo check -p windmill-worker` with `duckdb,private,parquet,quickjs` and with the CE set (no `private`)
- [x] EE unit tests: 4 passing (`duckdb_isolation_ee`)
- [x] Real DuckDB jobs on a worker started with `ENABLE_UNSHARE_PID=true`: plain query returns rows; `read_text('/etc/hostname')` is blocked and carries the local-filesystem explanation; a query above `memory_limit` fails as `Out of Memory Error` with the explanation; `INSTALL shellfs FROM community` is refused
- [x] Same worker with `DISABLE_DUCKDB_FS_SANDBOX=true`: the identical over-limit query spills to the job dir and returns `{n: 3000000, bytes: 600000000}`, and the local read succeeds, so the opt-out is unaffected
- [x] Worker log clean after each run (no panic, no worker death)
- [ ] Not exercised: a real `// materialize` + `// on_schema_change=sync` job (this environment has no DuckLake catalog or workspace storage). The pre-pass change rests on the transform's unit tests and on it being the same transform the main path already runs.

---
## Summary by cubic
Clarifies DuckDB errors when job isolation blocks local filesystem access. Over-limit queries now fail with Out of Memory; local file access errors explain the isolation with an opt-out.

- **Bug Fixes**
  - Cleared DuckDB `temp_directory` to disable spill; over-limit queries now fail with Out of Memory instead of a filesystem error.
  - Added clear message and `DISABLE_DUCKDB_FS_SANDBOX=true` opt-out to spill and local FS read/write failures; applied via the central error path (incl. prepare/typecheck).
  - Applied the same isolation policy to the `on_schema_change=sync` pre-pass connection.

- **Dependencies**
  - Bumped EE ref for improved Out of Memory hint wording.

<sup>Written for commit 8f88935ad0fbcb8147ba7898bb353d64fd48e5ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10581?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ cc47893265c7969161de33591a99d2656a163341

Nit-only verdicts at 1e656e89143b9436998538f1526fe98e0977b3a6 (codex: nit, claude and pi: good to merge). The codex nit (the out-of-memory hint claimed the limit derives from worker memory, ignoring the `DUCKDB_MEMORY_LIMIT` override) is fixed in windmill-ee-private 55a8721c, carried here by the ee-ref bump cc478932.

PR #10586, opened 2026-08-07
fix(parser-py): keep first param when def main( line has trailing comment

hugocasa:
Fixes WIN-2337

## Summary
`filter_non_main` in the Python signature parser reconstructs the `main` signature by concatenating the remainder of the `def main(` line directly with the joined following lines, without a newline in between. When the `def main(` line carries a trailing comment (e.g. `def main(  # comment`), the first parameter on the next line was glued into the comment text and silently dropped from the generated schema.

## Changes
- Insert `remaining += "\n";` between the `def main(` line remainder and the joined continuation lines in both the `async def main(` and `def main(` branches of `filter_non_main` (`backend/parsers/windmill-parser-py/src/lib.rs`).
- Add regression test `test_parse_python_sig_trailing_comment_on_def_line` (verified it fails without the fix: only `test2` was parsed, `test1` was swallowed by the comment).
- Bump the pinned `windmill-parser-wasm-py` npm package from `1.693.1` to `1.782.0` in `frontend/package.json` and `cli/package.json` (plus `frontend/package-lock.json` and `cli/bun.lock`). The script editor and CLI infer the schema client-side from this prebuilt WASM package, so the backend fix alone does not reach the UI until the package is rebuilt and republished.
- The remaining hunks in `lib.rs` are whitespace-only rustfmt fixes to pre-existing misindented `otyp_inferred: false,` lines in the same file.

## Screenshots

The same script opened in the script editor, before and after the parser bump. `first_param` is missing from the generated UI on the left because the trailing comment swallowed it.

Before (`windmill-parser-wasm-py@1.693.1`):

win2337-before-fix

After (`windmill-parser-wasm-py@1.782.0`):

win2337-after-fix

## Test plan
- [x] `cargo test` in `backend/parsers/windmill-parser-py` — all 33 tests pass
- [x] New regression test fails against the unfixed `filter_non_main`, passes with the fix
- [x] Built the WASM package and loaded it in Node: the published `1.782.0` returns both params, `1.693.1` returns only the second
- [x] Exercised in the running script editor with both package versions (screenshots above)


hugocasa:
✅ Review round clean @ 0fc5edb0c00037003b1342627fa2f414c9252e2d

hugocasa:
✅ Review round clean @ e930dd1157ba7229a8568a170de02bce44319416

hugocasa:
parser is published

PR #10588, opened 2026-08-07
fix: bump the bundled DuckDB engine to 1.5.5

rubenfiszel:
## Summary

Bumps the bundled DuckDB engine from 1.5.2 to 1.5.5 (`duckdb` crate `1.10502.0` → `1.10505.0`). Only `windmill-duckdb-ffi-internal` depends on that crate; everything else reaches the engine through the FFI cdylib. That crate is excluded from the cargo workspace, so only its own lockfile moves — `backend/Cargo.lock` is untouched.

**The bump fixes a worker-killing bug that exists today.** At 1.5.2 the crate hands DECIMAL values back as `rust_decimal::Decimal`, whose mantissa is 96 bits. DuckDB DECIMAL runs to 38 digits, so a wider value makes the crate call the panicking `Decimal::from_i128_with_scale`. That panic escapes an `extern "C"` frame, which cannot unwind, so the process aborts — killing **every other job on that worker** and leaving the offending job `running = true` in `v2_job_queue`. One statement is enough, from any script author:

```sql
SELECT '1234567890123456789012345678.9012345678'::DECIMAL(38, 10);
```

```
thread '<unnamed>' panicked at rust_decimal-1.40.0/src/decimal.rs:471:23:
Number exceeds maximum value that can be represented.
thread caused non-unwinding panic. aborting.
```

Reproduced on a real worker pinned to the current 1.5.2 FFI. At 1.5.5 the crate carries its own i128-backed `Decimal` and the same query returns `"1234567890123456789012345678.9012345678"`.

## Changes

- `duckdb` `1.10502.0` → `1.10505.0` in `windmill-duckdb-ffi-internal`, with its lockfile updated (`libduckdb-sys` 1.10505 no longer pulls the reqwest/hyper stack, which is where the 861 removed lock lines come from).
- Adapt the FFI to three upstream API changes:

  | Upstream change | Adaptation |
  |---|---|
  | `duckdb::types::Value` is now `#[non_exhaustive]` | a `_` arm returning a named error rather than a plausible-looking rendering that would silently misrepresent the column |
  | `Value` gained `UHugeInt(u128)` and `Geometry(Vec<u8>)` | rendered as a string (matching `HugeInt`) and as bytes (matching `Blob`) |
  | `rust_decimal` became an optional feature | enabled, because the `decimal`/`numeric` job-argument path needs it to build a DECIMAL from a JSON number |

- Regression guard pinning that every DECIMAL width — and `UHUGEINT` — renders instead of aborting the worker, plus a CI step that actually runs the FFI crate's tests (it is excluded from the workspace, so `cargo test --all` never reached them).
- **EE companion**: adapt the DuckDB isolation transform to the fact that, since DuckDB 1.5.3, an extension install *and* load go through the VFS — so `SET disabled_filesystems='LocalFileSystem'` also blocks autoload. See "Isolation" below.

## Test plan

Every item below was exercised as a real preview job on a worker started on its own tag, against a CE server built with `duckdb,private,parquet,quickjs,mysql`. A 28-case battery was run first on the **current 1.5.2 build** and then on 1.5.5, and the two JSON reports were diffed case by case. **Every case is byte-identical between the two engines**, except the DuckLake snapshot ids, which increase monotonically because both runs append to the same catalog.

| # | Item | Result |
|---|---|---|
| 1 | Plain queries, all scalar/nested types, `TIMESTAMP[_S/_MS/_NS]`, `HUGEINT`, `INTERVAL`, `LIST`/`STRUCT`/`MAP`/`ARRAY`/`ENUM`/`UNION`/`BLOB` | identical |
| 1 | Type aliases (#7148, #6426): `JSON` (parsed, not stringified), `CREATE TYPE` aliases bare / in a list / in a struct, `NULL::JSON` | identical |
| 1 | Column order | identical; `result_columns` still `{zulu,alpha,mike,bravo,AAA}` as declared |
| 1 | All 9 `result_collection` strategies | identical; each of the 9 shapes distinct and correct |
| 1 | NUL bytes in results | identical (`'a'‖chr(0)‖'b'` → `"ab"`, `length` still 3), in scalars, lists and structs |
| 2 | `// prepare` typecheck path | identical. 9 authored statements → 5 setup (`SET`, `INSTALL`, `LOAD`, `CREATE OR REPLACE TEMP MACRO`, and the `ATTACH 'ducklake://…'` that expands to 3 more setup blocks) → **exactly 4 result entries** for the 4 user statements, so the count contract the frontend asserts is intact |
| 3 | `ATTACH` postgres | identical (1000 rows, `numeric` sums match) |
| 3 | `ATTACH` mysql | identical (against a real MySQL 8 container) |
| 3 | `ATTACH 'datatable://main'` | identical |
| 4 | S3 via the workspace proxy: `COPY … TO 's3://…'` (120k rows) then `read_parquet('s3://…')` | identical |
| 5 | **DuckLake** | **the pre-existing data kept working** — see below |
| 6 | `// materialize` + `// data_test not_null/unique` | identical (50k rows, 0 violations) |
| 6 | `// partitioned daily` + the `wm_partition` macro | identical (86 400 rows in the `2020-01-02` slice) |
| 6 | `// on_schema_change=sync` (its pre-pass opens its own connection) | identical; the added column is picked up on the second run |
| 7 | Isolation (`ENABLE_UNSHARE_PID=true`) | all 5 guarantees unchanged; extension autoload regressed and is addressed — see below |
| 8 | Error-string coupling | **no EE matcher change needed**; both literals unchanged |
| 9 | 1.5.3 / 1.5.4 / 1.5.5 release notes | read; SQL grammar, `DESCRIBE` output shape and the `Out of Memory` text are byte-identical upstream. The one relevant change is extension loading (item 7) |

### DuckLake: the pre-existing data kept working

A lake was created **on the 1.5.2 build** with a 250 000-row table — big enough that DuckLake wrote a real 4.9 MB parquet data file rather than inlining into the catalog. The engine was then rebuilt at 1.5.5 against that same catalog and storage:

- **Reads**: identical row count, sums, min/max timestamp, distinct-bucket count and `md5(string_agg(name ORDER BY id))` digest — `aa9c8b9c7a3ed1414b535dbdf890598b` on both engines.
- **Writes**: a 60 000-row `INSERT` (new parquet data file), an `UPDATE`, and a `DELETE` (delete file) all committed against the 1.5.2-authored table.
- **Schema evolution**: `ALTER TABLE … ADD COLUMN` / `DROP COLUMN` on the old table.
- **Catalog**: no migration happened. `ducklake_metadata` still reads `version=1.0`, `created_by=DuckDB 8a5851971f` (the 1.5.2 writer). The catalog schema — table list and `ducklake_data_file` columns — is unchanged; the only catalog diff is additive rows.
- **Rollback is safe**: after 1.5.5 had read, written and altered the lake, a worker pinned back to the **1.5.2** FFI read the same digest and successfully wrote again. Reverting this PR does not strand the data.

### Isolation: one tightening, one regression (fixed)

Since **DuckDB 1.5.3**, extension install *and* load go through the VFS (`FileSystem::GetLocal(db)` in `extension_helper.cpp` / `extension_install.cpp`). Once the isolation transform sets `disabled_filesystems='LocalFileSystem'`, DuckDB can therefore no longer autoload an extension — not even one already installed on local disk. Confirmed empirically both ways. **1.10504.0 does not avoid this**: the change is in 1.5.3, so 1.5.4 behaves the same. Only staying at 1.5.2 would, at the price of keeping the abort above.

Everything a script could reach through autoload — `read_parquet`, `COPY … TO … (FORMAT parquet)`, the JSON type, `AT TIME ZONE`, `read_xlsx`, `read_avro`, `iceberg_scan`, `delta_scan`, `INET`, FTS — started failing with `Extension Autoloading Error`, and there was no user-side workaround because the transform's allowlist refused a script's own `INSTALL avro;` too. The EE PR restores that reach in two parts:

- **Nameable.** A script that writes `INSTALL inet; LOAD inet;` gets it hoisted ahead of the confinement `SET`s — the same shape the transform already used for the extensions the `ATTACH` transforms name. `FROM community` is now honored only for `bigquery`, the one extension the platform installs from that repository; previously any allowlisted core name could be requested from the community namespace, where something else could answer to it.
- **Preloaded.** `parquet` and `json` are loaded unconditionally, because nothing names them: parquet backs every DuckLake read and write and the S3 paths, and the JSON type is used by result rendering and the data-test probes.

**The allowlist is not DuckDB's `AUTOLOADABLE_EXTENSIONS`, and deliberately so.** `disabled_filesystems` only covers readers that go through DuckDB's own VFS; an extension bundling a separate IO layer walks straight past it. Two of the 26 do:

| Extension | With the lock already set |
|---|---|
| `sqlite_scanner` | read a SQLite file at an arbitrary path, and **created** files under `/tmp` and `$HOME` (8192 bytes each). Reads are confined to valid SQLite files — `.env` and `/etc/passwd` both fail — so it is a file-**creation** primitive more than a read one |
| `ui` | `CALL start_ui()` **exec'd `/usr/bin/xdg-open`** and bound an HTTP server on `localhost:4213` |

Both are DuckDB-signed, so a signature is not the property that decides this. They stay refused; nothing in Windmill needs either.

Every name that *is* on the list was checked against a set `disabled_filesystems`, aimed at the two targets that matter on a worker — PID 1's environment and a shared cache directory other jobs read back:

| Sweep | Result |
|---|---|
| every path-taking function each allowlisted extension registers × `/proc/1/environ`, `/proc/self/environ`, `/etc/passwd` | 111 attempts, **0 escapes** |
| core readers (`read_text`, `read_blob`, `read_csv`, `read_json`, `read_parquet`, `glob`, `parquet_metadata`, …) × `/proc/1/environ` | all blocked |
| `COPY … TO` in every reachable format (parquet, json, xlsx, delta, iceberg, csv) → shared cache dir and `/tmp` | all failed closed, **no file created** |
| `ATTACH` forms (postgres, mysql, ducklake, iceberg, unity_catalog, delta, plain duckdb) → `/proc/1/environ` and the shared cache | all blocked, **nothing planted** |

`icu` is deliberately **not** preloaded. Loading it re-defaults `TimeZone` to the worker's host zone, which would have silently changed how every TIMESTAMPTZ renders on an isolating worker. Measured on a worker with `TZ=Europe/Paris`:

| | non-isolating | isolating, icu preloaded | isolating, as shipped |
|---|---|---|---|
| `(TIMESTAMPTZ '2026-07-01 23:13:42+00')::VARCHAR` | `…23:13:42+00` | `…01:13:42+02` ❌ | `…23:13:42+00` ✅ |
| same, after `INSTALL icu; LOAD icu;` | `Europe/Paris` | — | `Europe/Paris` ✅ |

So a script that wants ICU semantics names it and gets exactly the 1.5.2 behavior; one that does not is left alone.

One residual: an *implicit* autoload — `SELECT '127.0.0.1'::INET` with no `INSTALL` — now fails with DuckDB's raw `Extension Autoloading Error` and no guidance pointing at the hoist. The error funnel that would carry that hint (`explain_error`/`hint_for`) is added by #10581, which touches this same file; the hint belongs there once it lands rather than duplicated here.

Everything else is unchanged:

| Guarantee | 1.5.2 | 1.5.5 |
|---|---|---|
| `read_text('/etc/hostname')` blocked | ✅ | ✅ same message |
| `SET disabled_filesystems=''` / `RESET` cannot re-enable | ✅ | ✅ same message |
| `SET allow_community_extensions=true` refused | ✅ | ✅ same message |
| `INSTALL shellfs FROM community` refused | ✅ | ✅ |
| `SET temp_directory=''` + over-`memory_limit` query → `Out of Memory Error` | ✅ | ✅ same message |
| `ATTACH` postgres, S3 round-trip and DuckLake reads work under isolation | ✅ | ✅ |
| `INSTALL icu FROM community` (core name, community namespace) | allowed | ✅ refused |
| `INSTALL sqlite_scanner` / `LOAD ui` | autoloaded implicitly | ✅ refused |
| hoisted `INSTALL inet` / `INSTALL icu` / `INSTALL fts` then use it | n/a | ✅ works |

### Error strings `hint_for` matches

Both literals `duckdb_isolation_ee::hint_for` pins are byte-identical at 1.5.5, so no matcher change is needed:

```
Permission Error: File system LocalFileSystem has been disabled by configuration
Out of Memory Error: could not allocate block of size 256.0 KiB (285.9 MiB/286.1 MiB used)
```

### Found on the way, not fixed here

A remote `delta_scan('s3://…')` through the workspace S3 proxy fails on **any** worker, isolated or not: the proxy answers a missing object with `500` where `object_store` expects `404`, so delta-kernel's probe for `_delta_log/_last_checkpoint` retries ten times and gives up (`s3_proxy_ee.rs:506`). `delta` is on the allowlist and loads fine under isolation; what it then reads is gated by that separate bug. Verified identical on a non-isolating worker, so it is not caused by this PR — worth its own issue.

### For upgraders

DuckDB installs extensions under a version-scoped directory, and `setup_duckdb_connection` re-issues `INSTALL <ext>;` per job. Every deployment therefore re-downloads `httpfs`/`azure`/`postgres`/`ducklake`/… on its first DuckDB job after upgrading. That is invisible on a networked install, but an air-gapped or egress-restricted deployment running off a pre-seeded `v1.5.2` extension directory must re-seed it for `v1.5.5` — and, on isolating workers, seed `parquet` and `json` as well.

### Not fixed by this bump

- `SET temp_directory='s3://…'` plus a spilling query still SIGABRTs the worker at 1.5.5, exactly as at 1.5.2. That is handled separately.
- The mechanism behind that class of failure is untouched: there is no `catch_unwind` at the FFI boundary, so any panic raised in a Rust frame inside the cdylib still takes down every co-resident job rather than failing the one. This PR removes one known trigger (the DECIMAL one). Wrapping `run_duckdb_ffi`'s body would convert the class into a job-level error and is worth a follow-up — with the caveat that it only helps for panics on the way down, not one raised inside a callback invoked from DuckDB's C++.

### Checks

- [x] `cargo check -p windmill-worker --features duckdb`
- [x] `cargo test -p windmill_duckdb_ffi_internal` — 16 passed, including the new wide-number regression guard (and this now runs in CI)
- [x] `cargo test -p windmill-worker --features duckdb,private duckdb_isolation` — 3 passed
- [x] `/local-review` and `/local-review-codex`


rubenfiszel:
Both nits from the last round are addressed; one of them is invisible to CI, so noting it here.

**Unbounded error text in the `Value` wildcard arm** — fixed in 6f37588c48. It now names the *type* (`other.data_type()`) rather than dumping the value, and `row_to_value` prefixes the column name:

```
column "x": DuckDB type Foo is not supported by this worker's engine bindings;
cast the column to a supported type (e.g. VARCHAR) to return it
```

**Stale `1.10502` claim in `pipeline_advanced_ee.rs:41`** — fixed in the EE companion (windmill-labs/windmill-ee-private#716, commit `a3ab9e0`), which `ee-repo-ref.txt` now points at. Expect this one to keep reappearing: the CI reviewers check out `windmill-ee-private` **main**, not the ref this PR pins, so they read the pre-fix copy of every `*_ee.rs` file.

Rather than just bumping the digit, I re-verified all three claims that comment makes, against 1.5.5:

| Claim | How |
|---|---|
| first run of a new asset rolls back to "no table" | managed materialize into a fresh asset with a mid-scan cast failure — 0 rows in `ducklake_table` afterwards |
| …including on connection drop, the FFI's abort path | `kill -9` of the worker mid-materialize — 0 live table rows, 0 orphan parquet files |
| a `temp` write is allowed alongside the lake write in one transaction | materialize whose setup is `CREATE TEMP TABLE … AS SELECT` feeding the lake write |
| cross-catalog reads work inside it | materialize into lake `main` joining a table in a second attached lake (separate `METADATA_SCHEMA`) |

rubenfiszel:
✅ Review round clean @ 6f37588c481cf15e63457ce2d20b175c756e5514

rubenfiszel:
✅ Review round clean @ 948de26fc1510b7f9e5ee47749d073d617b7da61 — nit-only verdicts at 89cebeefd8e221d085cd004acc78e72b416919ec (codex and pi: good to merge); nits addressed since.

**Codex's delta/fts P1, confirmed against the current pin.** Both are back on the allowlist and both were checked against a set `disabled_filesystems` first: a local `delta_scan` is blocked by the VFS (bare path and `file://`) and reads fine unlocked, and `fts` builds and queries an index post-lock with no file IO at all. `unity_catalog`, `encodings` and `aws` came back on the same evidence.

**The automated pin Claude asked for** is in the EE companion at `fe7eb44`: `classify_gates_install_load` now asserts `delta`, `fts`, `iceberg`, `excel`, `avro` and `unity_catalog` stay nameable, alongside the existing assertions that `sqlite_scanner` and `ui` stay refused and that `confine` puts every preload before the lock.

One caveat carried in the PR body: a *remote* `delta_scan('s3://…')` still fails, but on **any** worker — the S3 proxy answers a missing object with 500 where `object_store` expects 404, so delta-kernel's `_last_checkpoint` probe retries out (`s3_proxy_ee.rs:506`). Verified identical on a non-isolating worker, so it is not caused by this PR.

PR #10590, opened 2026-08-07
feat(frontend): show prod and dev as sibling choices in the workspace picker

Guilhem-lm:
## Summary

Rework the workspace picker (`/user/workspaces`) around the prod/dev pair, plus two smaller declutters on that page and on the create-workspace form.

A dev workspace is a fork mechanically, but it isn't one conceptually: prod and dev are siblings — two environments of the same thing — and burying dev under a collapsed "1 fork" row turned that duality into a hierarchy. This restores it: the card header names the family, and the two environments sit side by side underneath as the actual choice. Genuine throwaway forks keep their collapsed section below.

## Changes

- **`WorkspaceCard` — prod/dev pair.** When a workspace has a canonical dev workspace, the header (icon, name, id) becomes inert and two cards render below it: **Production** and **Dev** (or **Staging**, following `dev_workspace_label`), each with its own `as <username>`, hover/selection state, disabled and archived treatment, and Unarchive button. The dev card also shows the dev workspace's id, which the header doesn't carry. With no dev workspace the card is unchanged — whole card clickable, identity line in the header.
- **`WorkspaceTreeView` — dev lifted out of the fork list.** The canonical dev child moves to a `_dev` slot on its parent instead of into `_children`. Forks branched off the dev workspace are lifted beside the parent's own forks: `findDefaultForkBase` bases a fork on the dev workspace whenever one exists, so that is where most of a family's forks hang, and they would otherwise disappear along with it. Two consequences handled: `workspacesWithChildren` (which drives `hasForks` and the Expand/Collapse-all button) is now read off the built tree rather than off `parent_workspace_id`, so a family whose only child is its dev workspace correctly reports nothing to expand; and keyboard navigation pushes the dev id right after its parent's, since it is always visible.
- **`WorkspaceTreeView` — families start collapsed.** Previously a prod workspace with a canonical dev under it was expanded by default, so fork-heavy instances opened part-unfolded. Manual chevron toggles, Expand-all, and search auto-expansion are unchanged; as a side effect the Expand/Collapse button label now starts correctly at "Expand".
- **`/user/workspaces` invites section.** The "Invites to join a Workspace" heading, the "Show workspace forks" toggle, the "You don't have new invites at the moment." line and both invite lists are gated on `invites.length > 0`. When there are only fork invites the section still renders, so the toggle can reveal them.
- **`CreateWorkspaceInner`.** On cloud, the auto-add/auto-invite block (label, toggle, Mode/Role sub-panel) is hidden when the account's mail domain isn't allowed, replacing the disabled toggle plus the "`<domain>` domain not allowed for auto-invite" message. It is also hidden until the `isDomainAllowed()` check resolves — and the check fails closed, because `createWorkspace` commits the workspace *before* `editAutoInvite`, so offering the toggle to a domain the backend then rejects would leave a created workspace behind a failed submit. Self-hosted is unaffected; the section always renders there.

`auto_invite` defaults to `false` and the create path only calls the invite API under `if (auto_invite)`, so hiding the control cannot change what gets submitted. The component is only ever used for workspace/fork *creation*, never as an edit form, so the guard can't hide an existing setting.

## Screenshots

`/user/workspaces` — "AI Meta E2E" shows the Production/Dev pair with its 12 forks collapsed below (those forks hang off the dev workspace and were previously reachable only by unfolding it); "Debug sessions" and "Diff testing" have no dev workspace and keep the plain clickable card:

wsnits-prod-dev-pair

Create workspace, self-hosted — the auto-add section renders as before (the hidden case is cloud-only and can't be reproduced on a local instance):

create-workspace-autoadd

## Test plan

- [ ] Prod workspace with a canonical dev child: the card shows Production and Dev side by side, clicking each enters the right workspace, and clicking the header does nothing.
- [ ] Same family with forks branched off the dev workspace: they appear in the family's collapsed fork section rather than vanishing with the dev workspace.
- [ ] Workspace with a dev child and no other forks: no fork row, and the Expand/Collapse-all button reflects that it has nothing to expand.
- [ ] Keyboard navigation walks prod → dev → forks (when expanded), and Enter on the dev row enters the dev workspace.
- [ ] Searching a dev workspace's name or id still surfaces its family.
- [ ] Zero pending invites: the "Invites to join a Workspace" heading and the "Show workspace forks" toggle are both gone; with a fork-only invite the section appears and the toggle reveals it.
- [ ] Create workspace, cloud, gmail-domain account: the auto-add section is absent rather than disabled; an allowed-domain account still gets the toggle.

### DIFF
--- a/frontend/src/lib/components/workspace/WorkspaceCard.svelte
+++ b/frontend/src/lib/components/workspace/WorkspaceCard.svelte
@@ -12,15 +12,17 @@
 	import WorkspaceIcon from './WorkspaceIcon.svelte'
 	import WorkspaceCard from './WorkspaceCard.svelte'
 	import { twMerge } from 'tailwind-merge'
-	import { devBadgeText } from '$lib/utils/devWorkspaceLabel'
+	import { devBadgeText, devLabelWord } from '$lib/utils/devWorkspaceLabel'
 
 	interface ExtendedWorkspace extends UserWorkspace {
 		_children?: ExtendedWorkspace[]
+		_dev?: ExtendedWorkspace
 		marked?: string
 	}
 
 	interface Props {
 		workspace: UserWorkspace & { marked?: string }
+		dev?: ExtendedWorkspace
 		isForked?: boolean
 		depth?: number
 		children?: ExtendedWorkspace[]
@@ -37,6 +39,7 @@
 
 	let {
 		workspace,
+		dev,
 		isForked = false,
 		depth = 0,
 		children = [],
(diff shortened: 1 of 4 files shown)


Guilhem-lm:
✅ Review round clean @ e71230308a3284d7216445cf335d1575b0f3fa48 — nit-only verdicts at e71230308a3284d7216445cf335d1575b0f3fa48 (codex: Good to merge, pi: Good to merge, claude: nits); nits addressed in d0c5e048bff05e86ce5c97a311aa59da7688383a

Guilhem-lm:
✅ Review round clean @ 4deb54a3b7ddf4c7c64849fd035106bfed459c99

PR #10591, opened 2026-08-07
fix: contain a duckdb FFI panic to the job that caused it

rubenfiszel:
Stacked on #10588 — review that first; this PR's diff against `main` will include it until it merges.

## Summary

DuckDB runs in-process via the FFI cdylib, so a Rust panic that unwinds out of one of its `extern "C"` entry points **aborts the worker process**. Every other job running on that worker dies with it, and the offending job is left `running = true` in `v2_job_queue` until the zombie sweep picks it up.

#10588 removed one trigger of this (a DECIMAL wider than 96 bits). It did not touch the mechanism, which turns any panic anywhere in the bindings into a multi-tenant outage. This catches the unwind at both entry points and returns the panic message through the error channel the worker already parses, so it fails one job like any other error.

**This is not a blanket safety net**, and the code says so: an abort raised inside DuckDB's own C++ — or a panic crossing a callback it invoked — never becomes a Rust unwind and still ends the process. It covers panics raised in Rust frames on the way down, which is what the bindings themselves produce.

## Changes

- `catch_ffi_panic` wraps the bodies of `run_duckdb_ffi` and `prepare_duckdb_ffi`. A caught panic becomes `ERROR "<json>"`, the same wire shape every other failure uses, so `run_duckdb_ffi_safe`'s existing `starts_with("ERROR")` branch handles it with no worker-side change.
- `ffi_error` factors out that encoding, which the two entry points were each open-coding.
- On the panic path `column_order_ptr` is left as the caller initialized it (null), so nothing reads an uninitialized pointer.
- The default panic hook still runs first, so the message and backtrace reach the worker log for diagnosis; the job error carries just the message.

## Test plan

Three unit tests pin the contract: a panic becomes an error, the message survives into it, the `ERROR ` prefix and JSON encoding are what the worker parses, and the happy path is unchanged. They run in CI via the FFI-crate test step #10588 adds.

**Real-worker verification.** Unit tests cannot show that the process survives, so I injected a temporary `panic!()` into `run_duckdb_internal`, built the cdylib, and ran jobs against a worker started with it:

| | before this change | with this change |
|---|---|---|
| job outcome | worker aborts, job stuck `running` | `execution error: panic in duckdb ffi run_duckdb_ffi: injected ffi panic for testing` |
| worker process | gone (exit 134) | same pid across **three** consecutive panicking jobs |
| `non-unwinding panic` in log | yes | 0 |
| rows left in `v2_job_queue` | 1 (zombie) | 0 |

The injection was then removed and the cdylib rebuilt; the 28-case battery from #10588 was re-run against the shipped artifact and is identical to that branch, so nothing on the normal path moved.

## Checks

- [x] `cargo test -p windmill_duckdb_ffi_internal` — 19 passed
- [x] `cargo check -p windmill-worker --features duckdb`
- [x] full 28-case job battery re-run, identical to #10588

---
## Summary by cubic
Contain DuckDB FFI panics to the triggering job so the worker stays up. Rust panics crossing `extern "C"` now return `ERROR "<json>"`, which the worker already parses.

- **Bug Fixes**
  - Wrapped `run_duckdb_ffi` and `prepare_duckdb_ffi` with `catch_ffi_panic`; a panic becomes a job error instead of aborting the process.
  - Added `ffi_error` for consistent `ERROR "<json>"` encoding; on panic, `column_order_ptr` stays null to avoid reading uninitialized memory.
  - Extract panic messages from `&(dyn Any + Send)`; only Rust panics in the bindings are caught, not DuckDB C++ aborts or panics across callbacks.

- **Dependencies**
  - Synced `ee-repo-ref` to the latest EE repository reference.

<sup>Written for commit 89d90a63e79085ee2c0167167a0baad827a7628b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ e7295a26195f42e8f7da20bdd678a68d2ab49152 — nit-only verdicts at a29c5e944c8f68ab5fa8011dc8cf0f8b28da3c02; the repeated backtrace overstatement in the test module is fixed in e7295a26.

PR #10593, opened 2026-08-07
fix: pin MCP OAuth token requests to the validated address

hugocasa:
## Summary

`windmill_common::ssrf::validate_url_for_ssrf` resolves a host and verifies every
address is public, then returns those addresses as a `ValidatedTarget` so the caller
can pin them onto the client that actually connects. A caller that validates and then
lets the connect re-resolve the hostname gets no benefit from the check.

Two MCP OAuth token requests were not holding up their end of that contract. Both post
to an endpoint taken from the MCP server's own OAuth discovery metadata, and both carry
the `client_secret`:

- the authorization-code exchange re-validated the endpoint and pinned, but paid for a
  second DNS lookup to do it
- the token **refresh** used a bare `reqwest::Client::new()` — unpinned, and following
  redirects, unlike every sibling client in this flow

`get_or_refresh_mcp_client` already resolved and checked the token endpoint on both its
cached and freshly-registered paths, and then dropped the result. It now keeps it, and
hands it out through a single accessor that returns the URL together with the client
pinned to it, so the two cannot drift apart.

## Changes

- `McpClientCredentials` carries the validated token-endpoint target. The field is
  private; `token_request()` is the only way to reach it and returns
  `(pinned client, url)` as a pair.
- The MCP OAuth token refresh (`oauth_refresh_ee.rs`) goes through that accessor, which
  also brings it the no-redirect policy and request timeout the rest of the flow uses.
- The authorization-code exchange (`mcp_oauth_ee.rs`) reuses the already-validated target
  instead of re-validating, removing one DNS lookup per callback.
- Adds the test that was missing underneath the whole guard: that a pinned client
  connects to the pinned address rather than resolving the host.

Behavior note for reviewers: the refresh path previously followed redirects and had no
timeout. A token endpoint that answers a refresh POST with a 3xx will now fail instead
of following it. `ALLOW_PRIVATE_MCP_SERVER_URLS` and IP-literal hosts are unaffected —
they yield an empty address list, which means "connect normally".

## Test plan

- [x] `cargo check --features enterprise,private,mcp,oauth2` — clean
- [x] `cargo test -p windmill-mcp --features server,auth` — new pinning test passes
- [x] Negative control: with the pin removed the new test fails in ~5s rather than hanging
- [x] `cargo test -p windmill-common --lib ssrf` — 19 passed
- [ ] Manual: connect an MCP server over OAuth, let the access token expire (or shorten
      `expires_in`), then trigger a call needing the token and confirm the refresh still
      succeeds against a real provider — this is the path whose HTTP client semantics changed

---
## Summary by cubic
Pins MCP OAuth token requests to the SSRF‑validated token endpoint address to block DNS rebinding. Unifies the refresh and auth-code flows on the same pinned, no-redirect client and removes an extra DNS lookup.

- **Bug Fixes**
  - `McpClientCredentials` now keeps the token endpoint URL and its validated target private; `token_request()` returns a (`pinned` `reqwest` client, url) so the pin and URL stay in sync.
  - `get_or_refresh_mcp_client` preserves the result of `validate_mcp_server_url_for_bad_request` and passes it through.
  - Authorization-code exchange reuses the validated target instead of re-validating (one less DNS lookup).
  - Token refresh now uses the pinned, no-redirect client with the shared timeout; 3xx responses error instead of following.
  - Test verifies the client connects to the pinned address and never resolves the host, with a bounded accept loop; drops the non-blocking reset on the accepted socket.

- **Dependencies**
  - Updated EE repository reference to the latest revision.

<sup>Written for commit 9eaadd1a8005bc251cc104a66fd9b9de79bcfe3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
✅ Review round clean @ d5bc91a60b152434c0462d95b44477bbd1d6a35d — nit-only verdicts at 822a81494c03874c0d883c0a1b1dfa57b86279c0 (codex: Good to merge; pi: Good to merge; claude: mergeable with nits); nits addressed in d5bc91a60b (windmill) and 729df45 (ee): token endpoint made private so the pinning invariant is enforced rather than documented, and the pinning test now resets the accepted socket to blocking so it does not read empty on macOS.

hugocasa:
✅ Review round clean @ a2f32000753ef2bee630c196f5055e99e7c6a565

PR #10596, opened 2026-08-07
feat: version resource values with history, diff and restore

hugocasa:
## Summary

Resource values had no history: editing one destroyed the previous value, with no way to see what changed or roll back. Flows, apps and scripts all keep version history; resources did not.

This adds append-only version history for every resource value, surfaced in the resource editor as a History drawer with a diff view and restore.

The motivation is reusable AI agents, which are `ai_agent` resources: without version history you cannot attribute a behaviour change to a prompt edit, because the prompt it replaced is gone. Versioning is workspace-wide rather than agent-specific because the primitive is useful for every resource type.

## Changes

- **`resource_version` table** mirroring `flow_version`'s shape, keyed `(workspace_id, path)` with a cascading FK so a rename carries history to the new path and a delete takes it along. RLS joins back to `resource`, so visibility follows the parent rather than restating its several policies. The migration seeds one version per existing resource so the first edit has something to diff against.
- **Recording lives in a database trigger**, not the API handlers. Resource values are written well outside them: variable renames rewrite a linked resource, workspace forks clone resources wholesale, native integrations upsert them, and trashbin restores reinsert them. A handler-level hook missed all of those, and a rename that changes path and value in one statement left the newest version holding the pre-rename value while the UI labelled it "Current".
- **`state` and `cache` are excluded** — both are rewritten by every job that calls `setState` or caches a result. Same exclusion workspace export already applies.
- **An unchanged value records nothing**, so no-op saves, renames and description edits don't accumulate rows. (Deleting a resource cascades its history away, so a trashbin restore starts a fresh history at v1 rather than continuing the old one.)
- **History is capped at 100 versions per path.** Unlike flows, resources are not only edited by humans: `setResource` from a scheduled script reaches the same write path.
- **Three routes** — list history, fetch one version with its value, restore. Restore writes the old value forward as a new version rather than rewinding, so history stays append-only and the restore is itself attributable and reversible.
- **Dangling references are surfaced, not blocked.** A version fetch reports `$var:`/`$res:` paths that no longer resolve so the UI can warn before restoring; refusing would make a version unrestorable for a reason outside the resource.
- **`prune_versions` gains a `"resources"` arm** for cloud parity with scripts, flows and apps.

## Screenshots

Version list, value view, and the dangling-reference warning:

v2-history-value

Diff against the current value:

v2-history-diff

Rows use the shared `VersionListItem` component, which `ScriptVersionHistory` was also converted to, so the two history drawers stay visually identical. The warning uses the standard `Alert` component.

## Test plan

- [ ] Edit a resource twice, open History, confirm both versions listed with author and timestamp
- [ ] Diff a version against current and confirm the changed fields are highlighted
- [ ] Restore an older version: value reverts, a new version is appended (history does not shrink), and both drawers close so a stale editor cannot save over it
- [ ] Re-save a resource without changing the value and confirm no version is added
- [ ] Rename a resource and confirm its history moves with it and no version is added
- [ ] Clear a resource's past versions and confirm only the current value remains and the resource itself is untouched
- [ ] Rename a variable linked to a resource and confirm the newest version matches the live value
- [ ] Confirm a `state` or `cache` resource accumulates no history
- [ ] Restore a version referencing a deleted variable and confirm the warning appears and restore still succeeds

## Notes

- Behaviour change worth flagging: `update_resource_value` on a nonexistent path previously returned 200 and silently did nothing; it now 404s. The handler is shared with restore, where a missing resource is a real error.
- Version rows hold the same value the `resource` row already holds and inherit its permissions, so the audience is unchanged. Values written through the UI keep secrets in linked variables, but one pushed by `wmill` or written by `setResource` can hold an inline credential — so rotating such a credential would leave the old one readable in the history. **Clear past versions** (in the History drawer, owner-gated) is the remediation: it drops every version but the current value.
- History captures the resource, not its transitive closure: two versions can be byte-identical while behaviour differs because a referenced variable changed. Documented in `docs/reusable-ai-agents.md`.
- Non-secret **variables** are the natural follow-up and the drawer was built to be reused for them. Secret variables should never be versioned — retaining rotated credentials would make rotation stop being remediation.

---
## Summary by cubic
Adds append-only version history for resource values with diff, restore, and an owner‑gated “Clear past versions” in the Resource Editor. Recording uses a DB trigger (covers direct writes), history is pruned by a periodic sweep, and the UI opens History unselected and diffs against the newest version for a consistent view.

- New Features
  - Resource history API: list versions, get one with value and missing `$var:`/`$jsonvar:`/`$res:` references, restore, and clear past versions; OpenAPI updated.
  - Workspace pruning supports `resources`, and Cloud quotas show prunable resource versions.

- Bug Fixes
  - Frontend: open History with no preselected version; discard stale loads when the path changes; reset the pane and disarm Clear confirmation when the selection moves; scope history to the selected workspace; treat the newest row as “Current”; never show a diff if the baseline can’t load and explain why; restore operates on the loaded version; diff baseline uses the newest version instead of the live row.
  - Backend: cap listings to 100 versions; trim history in the monitor; tighten RLS (SELECT-only) and write via a SECURITY DEFINER trigger; warn on high write rates (20/min) and harden in‑process counters; `update_value` now 404s on a missing resource.
  - CI: increase Node heap for `svelte-check` via `NODE_OPTIONS`.

<sup>Written for commit 3b25b5c407eeb08cffa51b0aaaa9f79ddf014d57. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
@pi Dismissing the `$jsonvar:` P1 — `$var:` is a *substring* of `$jsonvar:`, not a prefix, so `strip_prefix("$var:")` returns `None` for it and the iteration falls through to `$jsonvar:` correctly:

`"$jsonvar:u/x"` is `$ j s o n v a r :` — index 1 is `j`, where `$var:` needs `v`. The mismatch is at the second character, so no strip occurs. (The `var:` you're seeing does appear in the string, but at offset 5, which `strip_prefix` doesn't consider.)

Your coverage point was right though, so `test_missing_references` in `backend/tests/resource_versioning.rs` now pins exactly this: a version carrying `$jsonvar:` to a *present* variable, `$jsonvar:` to an absent one, `$var:` to a present one and `$res:` to an absent one, asserting only the two genuinely-absent references are reported. It passes, and it would fail if the prefix matching ever regressed to a substring test.

hugocasa:
Nits from round 9 addressed in 63d008a895 — all three reviewers' lists:

- **Nested `<button>`** (codex/pi/claude): `VersionListItem` now carries the card chrome on a `div` with the select `<button>` and an optional `action` snippet as siblings. `ScriptVersionHistory` passes "Run page" through `action`. Verified in the browser: no `Run page` button has a `<button>` ancestor, row click selects, and "Run page" navigates to that row's own hash.
- **`truncate` regression** (pi/codex/claude): restored on the row, with `min-w-0` on the inner button so the flex child can actually shrink. Verified with a 150-char deployment message — clips with an ellipsis instead of crossing the splitter.
- **Stale `.sqlx` entries** (codex found two; pi/claude one): both `query-6be4bf59…` and `query-b072e88b…` deleted. Neither is issued at HEAD.
- **Migration comment naming the old call site** (claude/codex): now names `trim_resource_versions`.
- **Clear gate on the wrong workspace** (codex/claude): `canClearSelected` now anchors on `$workspaceStore` — the workspace `$userStore` actually describes — instead of `effectiveWorkspace`, so an explicit `workspace` prop can no longer produce a verdict from the wrong membership. Verified Clear still shows for the ordinary same-workspace owner case.

**Cadence** (claude/codex): comment fixed rather than the period. `should_run` takes a `u8`, so a true hour would need the signature widened; ~20 min is a fine bound for a cap that only has to hold eventually, and it is the same cadence as the other heavy sweeps in that file. The comment now states it in iterations (`~20min at the default LISTEN_NEW_EVENTS_INTERVAL_SEC`) rather than asserting a wall-clock hour, since the interval is operator-tunable.

**Not adding the trim test** (claude/codex/pi all noted the gap): `backend/` is a bin-only crate — no `[lib]` target — so `backend/tests/*` cannot call `trim_resource_versions`. The alternative, re-issuing a copy of the DELETE from the test, would pin a duplicate rather than the function that ships, and adding a lib target is out of scope here. Flagging it as a known gap instead of shipping a test that can silently drift from the code it claims to guard.

hugocasa:
Round 10: claude and pi both "Good to merge". Codex raised three; two taken in 287c7247ad, one dismissed.

**Taken — history listing was unbounded** (`resources.rs`). `get_resource_history` now `LIMIT`s to `MAX_RESOURCE_VERSIONS`, which moved to `windmill-store` so the endpoint and the monitor sweep read one definition instead of two. Verified against a path seeded to 152 rows: the endpoint returns exactly 100.

**Taken — quadratic reference lookup** (`resources.rs`). `existing_vars`/`existing_res` are `HashSet`s now, so the filter is a hash lookup per reference rather than a scan of every resolvable path.

**Dismissed — retyping a resource into `state`/`cache`.** The premise is right, and I checked it rather than assuming: `update_resource` does apply `resource_type`, and `POST /resources/update/{path} {"resource_type":"state"}` succeeds against a running server. But there is no way to reach it from the product — the resource type is chosen at creation and the editor drawer has no control that changes it, so this needs a hand-written API call against a resource the caller already owns. That is not worth new DDL on this migration. The consequences are also bounded: the frozen rows stay subject to the per-path cap and the FK cascade still takes them on delete, so deleting and recreating the resource clears them.

Nits from the same round are addressed; the earlier `npm check` failure was an OOM in the runner (`Ineffective mark-compacts near heap limit`, no diagnostics emitted; an unrelated PR hit the same abort minutes later) and passed on a clean re-run.

hugocasa:
✅ Review round clean @ 287c7247ad8f88d1d9368693239956c3f579317d

hugocasa:
Re: **resource-type transitions break history invariants** — accepting this one deliberately, not overlooking it.

Changing `resource_type` is reachable only through `POST /resources/update/{path}`; nothing in the UI sends the field, so this needs a hand-written API call that both retypes a resource and changes its value.

Both directions are understood:

- **Into `state`/`cache`**: the trigger's `WHEN` excludes the row, so the value changes without recording. History freezes holding the last recorded value. The resource is now an internal type, whose history is a no-op by design.
- **Out of an excluded type without a value change**: nothing is recorded, so there is no baseline until the next edit. Less harmful than it sounds, since a `state` resource never had history to lose in the first place.

Both self-heal: the next value-changing write on a non-excluded type records normally and re-syncs.

Closing it would mean either trigger logic that cleans history on entering an excluded type and seeds it on leaving, or rejecting the transitions outright in the handler. Both add machinery to cover a path no product surface exercises, and the first was tried on this branch earlier and reverted as unnecessary. If we do close it later, the handler guard is the cheaper of the two.

The related consequence you raised — Clear retaining a stale row — only follows from that same transition, so it is bounded by the same reachability.

hugocasa:
✅ Review round clean @ 3b25b5c407eeb08cffa51b0aaaa9f79ddf014d57 — nit-only verdicts at 0d33b0a0015a3bccf84ec02b46b4c7c9ab4afd0e (codex: Good to merge, pi: Good to merge, claude: nits); nits addressed in 3b25b5c407eeb08cffa51b0aaaa9f79ddf014d57

PR #10598, opened 2026-08-07
fix(frontend): stack login options in one column below four

Guilhem-lm:
## Summary

On the login page, the third-party login buttons switched to a two-column grid as soon as there were 3 options, leaving an orphaned button on its own row. Instances with 2 or 3 configured providers now get a single stacked column; the two-column grid kicks in from 4 options.

## Changes

- `Login.svelte`: the grid switches to `grid-cols-2` above 4 options instead of above 2.
- The option tally now includes the SAML/SSO button, which was previously excluded from the count that drove the layout.
- Dropped the stray `mt-2` on the custom-provider and SSO buttons — they are grid children, so `gap-4` already spaces them and the margin made the stacked column unevenly spaced.

## Screenshots

Before — 3 providers, orphaned third button:

login-three-options-before

After — 3 providers stacked in one column:

login-three-options

After — 4 providers, two-column grid retained:

login-four-options

## Test plan

- [ ] Instance with 2 or 3 OAuth providers configured: `/user/login` stacks the buttons in one full-width column.
- [ ] Instance with 4+ providers: buttons still render in the two-column grid.
- [ ] Instance with SAML plus 3 OAuth providers (4 options total): the grid is two columns and the SSO button aligns with the rest.

### DIFF
--- a/frontend/src/lib/components/Login.svelte
+++ b/frontend/src/lib/components/Login.svelte
@@ -455,6 +455,8 @@
 	$effect(() => {
 		error && sendUserToast(escapeHtml(error), true)
 	})
+
+	let loginOptionCount = $derived((logins?.length ?? 0) + (saml ? 1 : 0))
 </script>
 
 <div class="bg-surface px-4 py-8 border sm:rounded-lg sm:px-10">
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
✅ Review round clean @ 41ebbe4932a072ad7a41505509bdb7017e667951 — all verdicts "Good to merge" at 4a0cbfc65bf2e52484db6f5a9921c8c24d4c62be; the only observation (redundant `w-full` on the custom-provider and SSO buttons, already applied unconditionally by Button) addressed in 41ebbe4932a072ad7a41505509bdb7017e667951.

PR #10600, opened 2026-08-08
S3BufferedReader.read(size) returns whole HTTP chunks, not size bytes — up to ~3.3 GB in memory via write_s3_file

tusharui:
Description
python-client/wmill/wmill/s3_reader.py — S3BufferedReader.read(size) loops range(size) times and appends the whole chunk from iter_bytes() each iteration. httpx streams yield ~64 KB (65536-byte) chunks, so size counts chunks: read(5) returns up to 5 chunks (~320 KB), violating the io.BufferedReader.read(n) contract of returning at most n bytes.
Production impact: bytes_generator() (used by write_s3_file) calls read(50 * 1024), so a single call can return up to 51200 × 64 KB ≈ 3.3 GB instead of a 50 KB slice (bounded by the file size). read() with no args and load_s3_file are unaffected.
Steps to reproduce (no network)
import sys
sys.path.insert(0, "python-client/wmill")
from wmill.s3_reader import S3BufferedReader

r = object.__new__(S3BufferedReader)
r._iterator = iter([b"A" * 10, b"B" * 10, b"C" * 10])
print(r.read(5))  # b"AAAAAAAAAABBBBBBBBBBCCCCCCCCCC" (30 bytes), expected b"AAAAA"
Expected
read(size) returns at most size bytes, buffering the remainder of the final chunk for the next call; read(-1) returns everything.
Environment
wmill 1.783.0, httpx >= 0.24
Introduced by the object_store migration in #3116.

### DIFF
--- a/python-client/tests/wmill_client_test.py
+++ b/python-client/tests/wmill_client_test.py
@@ -168,5 +168,54 @@ class TestParseS3Object(unittest.TestCase):
     def test_s3object_passes_through(self):
         self.assertEqual(wmill.parse_s3_object(S3Object(s3="x")), S3Object(s3="x"))
 
+class TestS3BufferedReaderRead(unittest.TestCase):
+    """Pure-unit tests for S3BufferedReader.read — no network/env needed.
+
+    iter_bytes() yields whole HTTP chunks (~64KB), so read(size) must slice
+    the last chunk and buffer the leftover for the next call.
+    """
+
+    def make_reader(self, chunks):
+        from wmill.s3_reader import S3BufferedReader
+
+        reader = object.__new__(S3BufferedReader)
+        reader._iterator = iter(chunks)
+        reader._buffer = bytearray()
+        return reader
+
+    def test_read_size_returns_exact_bytes(self):
+        reader = self.make_reader([b"AAAAAAAAAA", b"BBBBBBBBBB", b"CCCCCCCCCC"])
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Thanks a lot for this — the diagnosis is exactly right, and the reproduction made it easy to confirm. `read(size)` was counting httpx chunks rather than bytes, and `bytes_generator`'s `read(50 * 1024)` turned that into a whole-file buffer on any `write_s3_file(dst, load_s3_file_reader(src))` pipe.

We're taking this over to drive it to completion, using your branch as the base: #10623. Your commits are preserved there as-is, so you stay the author of the fix and are credited in the history and the PR description.

What we changed on top:
- Removed the duplicated `TestS3BufferedReaderRead` class from `python-client/tests/wmill_client_test.py`. That file isn't run by CI (`sdk-tests.yml` runs `pytest tests/` from `python-client/wmill`), and the same assertions already live in the new `test_s3_reader.py`.
- Trimmed the new test file to the four behaviours a future change could actually break, and made the `bytes_generator` guard call `bytes_generator` instead of reimplementing its loop.
- Aligned `peek()` with `io.BufferedReader.peek`, which does at most one read on the underlying stream, instead of looping until `size` bytes are buffered.

Closing this in favour of #10623 — please do follow along there, and thanks again for the contribution.

PR #10601, opened 2026-08-08
feat(sdk): add S3 connection settings helpers and cancelJob

tusharui:
Description
Adds missing S3/database connection-settings helpers to the TypeScript and Python SDKs so both clients expose the same surface, and fixes a gap where the TS SDK had no job-cancellation helper.
TypeScript client (typescript-client/)
- Re-enable duckdbConnectionSettings and polarsConnectionSettings, which were previously stubbed out behind a TODO(gb). They now call the v2 helpers endpoints, strip $res:/res:// prefixes via parseResourceSyntax, and honor an explicit workspace arg (falling back to the WM_WORKSPACE env var), matching the Python SDK's behavior.
- Add boto3ConnectionSettings, mirroring Python's __boto3_connection_settings: resolves the S3 resource via the s3_resource_info endpoint and builds the endpoint_url (including port when set) plus aws_session_token for OIDC/STS temporary credentials.
- Add cancelJob(jobId, reason?), mirroring Python's cancel_job, with a default reason when none is supplied.
- Add DuckDbConnectionSettings, PolarsConnectionSettings, and Boto3ConnectionSettings types in s3Types.ts, matching the s3_types.py models.
- Wire the new functions into the package's named and default exports in build.sh.
Python client (python-client/)
- Add database_url_from_resource(path) (class method + module-level convenience), mirroring the TS databaseUrlFromResource — builds a postgresql:// URL from a database resource. This closes the last remaining SDK parity gap in the database-helper surface.
Verification
- npx tsc --noEmit passes (client is rebuilt from openapi.yaml via build.sh).
- bun test — 167 pass.
- python -m pytest tests/ -q — 116 pass.
- Smoke-tested all new TS functions with a stubbed fetch: correct endpoints, $res:/res:// stripping, endpoint/port/token composition, and cancelJob request body verified. Smoke-tested database_url_from_resource against the mock API.
Impact
- Additive only; no breaking changes to existing exports or behavior.
- No new dependencies.

---
## Summary by cubic
Adds S3 connection settings helpers (DuckDB, Polars, boto3) and a job-cancel helper to the TypeScript SDK, plus a database URL helper to the Python SDK. Addresses review feedback with refined exports and safer URL construction.

- **New Features**
  - TypeScript SDK: added `duckdbConnectionSettings`, `polarsConnectionSettings`, `boto3ConnectionSettings` (v2 helpers, strips `$res:`/`res://`, optional `workspace`, builds endpoint with protocol/port and optional `aws_session_token`), and `cancelJob(jobId, reason?)`; added types `DuckDbConnectionSettings`, `PolarsConnectionSettings`, `Boto3ConnectionSettings`; exported in named and default exports.
  - Python SDK: added `database_url_from_resource(path)` to build `postgresql://` URLs with percent-encoded components and sensible defaults.

No breaking changes.

<sup>Written for commit 292929c57ad79b90d71b31c332fe7b7efa0148f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/python-client/wmill/wmill/client.py
+++ b/python-client/wmill/wmill/client.py
@@ -1297,6 +1297,25 @@ class Windmill:
             params=params,
         ).json()
 
+    def database_url_from_resource(self, path: str) -> str:
+        """Build a PostgreSQL connection URL from a database resource.
+
+        Args:
+            path: Path to the database resource in Windmill
+
+        Returns:
+            PostgreSQL connection URL string
+        """
+        resource = self.get_resource(path)
+        return "postgresql://{}:{}@{}:{}/{}?sslmode={}".format(
+            resource["user"],
+            resource["password"],
+            resource["host"],
+            resource["port"],
+            resource["dbname"],
+            resource["sslmode"],
+        )
+
     def request_interactive_slack_approval(
         self,
         slack_resource_path: str,
@@ -1722,6 +1741,19 @@ def get_result(job_id: str, assert_result_is_not_none=True) -> Dict[str, Any]:
     return _client.get_result(
         job_id=job_id, assert_result_is_not_none=assert_result_is_not_none
     )
+
+
+@init_global_client
+def database_url_from_resource(path: str) -> str:
+    """Build a PostgreSQL connection URL from a database resource.
+
+    Args:
+        path: Path to the database resource in Windmill
+
+    Returns:
+        PostgreSQL connection URL string
+    """
+    return _client.database_url_from_resource(path)
 
 
 @init_global_client
--- a/python-client/wmill/wmill/client.py
+++ b/python-client/wmill/wmill/client.py
@@ -1306,14 +1306,21 @@ class Windmill:
         Returns:
             PostgreSQL connection URL string
         """
+        from urllib.parse import quote
+
         resource = self.get_resource(path)
+        # Defaults mirror PgDatabase::to_uri(); quote each component so that
+        # credentials containing '@', '/', '?' etc. cannot reshape the URI.
+        host = resource["host"]
+        if not (host.startswith("[") and host.endswith("]")):
+            host = quote(host, safe="")
         return "postgresql://{}:{}@{}:{}/{}?sslmode={}".format(
-            resource["user"],
-            resource["password"],
-            resource["host"],
-            resource["port"],
-            resource["dbname"],
-            resource["sslmode"],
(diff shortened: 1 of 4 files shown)


rubenfiszel:
Thanks for this — re-enabling the helpers that had been sitting behind that `TODO(gb)` closes a real gap between the two SDKs, and the boto3 mirror of `__boto3_connection_settings` is faithful to the Python side.

We're taking this over to drive it to completion, using your branch as the base: #10624. Your commits are preserved there as-is, so you stay the author and are credited in the history and the PR description.

What we changed on top:
- An empty-string `s3_resource_path` reached the v2 endpoints as `Some("")` and 404'd looking for a resource named `""`. The Python client sends `json={}` in that case and gets the workspace default, and since `""` is the Python signature's own default it's exactly the call people port over. All four helpers now share one body builder that maps `""` back to absent.
- `S3Resource` in `openapi.yaml` was missing `port` and `token`, which is why the `as` cast was needed. The Rust struct in `windmill-types/src/s3.rs` has always had both, so we declared them on the schema and dropped the cast. The cast also asserted `accessKey`/`secretKey` were always present, but they're `Option<String>` server-side, so those are now optional on `Boto3ConnectionSettings`.
- Moved the boto3 composition into `s3Types.ts` so the CI test suite can import it without the generated client, and added regression tests for the endpoint/port and session-token branches.
- Gave the TypeScript `databaseUrlFromResource` the same percent-encoding and defaults as your new Python helper — it was interpolating raw and emitting `sslmode=undefined` when the resource had no `sslmode`. Also added it to the named exports; it was only reachable through the default export.
- Synced `build.jsr.sh`, which keeps its own copy of the export list, so the JSR package gets the new functions too.
- Regenerated `system_prompts/auto-generated/`. That's the one CI blocker: `check-system-prompts` triggers on both `typescript-client/**` and `python-client/wmill/wmill/client.py`, and it hadn't run here yet, so the failure wasn't visible on this PR.

Closing this in favour of #10624 — please do follow along there, and thanks again for the contribution.

PR #10604, opened 2026-08-10
fix: order workspace members and invites by email

Guilhem-lm:
## Summary

Toggling a member's role on the workspace settings Members tab reshuffled the list: the row you just edited jumped to the bottom.

`list_users` and `list_pending_invites` had no `ORDER BY`, so Postgres returned rows in heap order. An `UPDATE` rewrites the tuple at the end of the heap, and the settings page refetches the list right after the mutation — so the edited member came back last. Reproduced on a dev database:

```
before: (0,11) admin@windmill.dev | (0,70) test@mail.com | (0,77) new@mail.com
after an UPDATE on admin@windmill.dev:
        (0,70) test@mail.com | (0,77) new@mail.com | (0,117) admin@windmill.dev
```

Both list endpoints now sort by email, which also makes them deterministic for their other consumers (fork member settings, `OnBehalfOfSelector`, the CLI).

No frontend change: the members table already sorts non-instance-group members before instance-group ones, and `Array.prototype.sort` is stable, so the alphabetical order from the API carries through each block unchanged.

## Changes

- `list_users`: `ORDER BY email`
- `list_pending_invites`: `ORDER BY workspace_invite.email` (same class of instability — changing an invite's role rewrites its row too)
- `.sqlx` offline cache: two entries added for the new query text, the two pre-`ORDER BY` entries removed

## Test plan

- [x] Members tab: toggle a member's role and confirm the row stays in place (verified in the browser against a locally built backend — see screenshot)
- [x] API: after a role update, the endpoint returns `admin@windmill.dev new@mail.com test@mail.com` while the underlying heap order is `test@mail.com admin@windmill.dev new@mail.com`, i.e. the churn no longer reaches the response
- [x] Invites: created `zoe`, `alice`, `mike` out of order, then changed `alice`'s role — endpoint returns `alice mike zoe` before and after, while heap order is `zoe mike alice`

## Screenshots

Members tab immediately after toggling `new@mail.com` to Operator — the row stays second instead of jumping to the bottom:

members list order stays put after a role toggle

### DIFF
--- a/backend/.sqlx/query-3ce57583042b1daa2bd9d88502f6f4ab76c509dafee7371ac620c85e1b5d2498.json
+++ b/backend/.sqlx/query-3ce57583042b1daa2bd9d88502f6f4ab76c509dafee7371ac620c85e1b5d2498.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "SELECT\n            workspace_invite.workspace_id,\n            workspace_invite.email,\n            workspace_invite.is_admin,\n            workspace_invite.operator,\n            workspace.parent_workspace_id\n        FROM workspace_invite JOIN workspace ON workspace_invite.workspace_id = workspace.id\n        WHERE workspace_id = $1",
+  "query": "SELECT\n            workspace_invite.workspace_id,\n            workspace_invite.email,\n            workspace_invite.is_admin,\n            workspace_invite.operator,\n            workspace.parent_workspace_id\n        FROM workspace_invite JOIN workspace ON workspace_invite.workspace_id = workspace.id\n        WHERE workspace_id = $1\n        ORDER BY workspace_invite.email",
   "describe": {
     "columns": [
       {
@@ -42,5 +42,5 @@
       true
     ]
   },
-  "hash": "ce5c081bbf8322371a6ec95b6a0b033837d574a1ec72dfe42b666172daf10880"
+  "hash": "3ce57583042b1daa2bd9d88502f6f4ab76c509dafee7371ac620c85e1b5d2498"
 }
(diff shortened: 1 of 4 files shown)


Guilhem-lm:
✅ Review round clean @ b391c270f9f8c70772d3029c2d05a0908f17a6d9 — nit-only verdicts at 4e23cfa8edfcaf341b8c79b1d6463449696222bf (codex: Good to merge, pi: Good to merge, claude: nit-only); the orphaned pre-`ORDER BY` sqlx entries are deleted in b391c270f9f8c70772d3029c2d05a0908f17a6d9. The `sortedUsers` sort-per-row cost is a pre-existing shape and left as-is.

Guilhem-lm:
@pi the `sortedUsers` finding reads a stale PR body. The frontend change was dropped before this round was triggered — the body's Changes section has listed only the two `ORDER BY` additions and the `.sqlx` entries since before the `/review` comment, and the diff at c8853a0aef9a9069d6150c56549117b4b3831b3c contains no frontend files at all. The quoted "email tie-break in `sortedUsers`" bullet no longer exists.

The underlying observation is right and is why the change was dropped: the comparator's stable sort preserves the API's ordering within each block, so the backend `ORDER BY` alone is sufficient.

Your second point was valid and is now fixed — the unchecked "workspace mixing manual and instance-group members" test-plan item tested pre-existing grouping behaviour this commit doesn't touch, so I removed it. Test plan is now three checked backend/API scenarios.

Guilhem-lm:
✅ Review round clean @ c8853a0aef9a9069d6150c56549117b4b3831b3c — codex: Good to merge, claude: Good to merge, pi: nit-only (stale PR body, dismissed in the reply above; its valid test-plan point applied to the body, no code change).

PR #10606, opened 2026-08-10
fix(cli): attach the right job path to preview runs

rubenfiszel:
## Summary

Fixes WIN-2342.

`wmill script preview` and `wmill flow preview` derived the Windmill path a preview job runs under straight from the raw CLI argument, so the path only came out right when the argument was a bare workspace-root-relative path. Every other spelling produced a job with no path, a wrong path, or no job at all:

| invocation | before | after |
|---|---|---|
| `wmill script preview ./f/a/b.ts` | **nothing ran** — `validatePath` rejected the leading `.`, logged a warning and returned exit 0 | `f/a/b` |
| `wmill script preview /abs/repo/f/a/b.ts` | **nothing ran** (same) | `f/a/b` |
| `wmill script preview b.ts` from inside `f/a` | `ENOENT: no such file or directory` (reading wmill.yaml had already chdir'd to the root) | `f/a/b` |
| `wmill script preview f/a/b__mod/script.ts` | `f/a/b__mod/script` | `f/a/b` |
| `wmill script preview wm_dbt.yaml` from inside `f/a/proj__dbt` | **nothing ran** | `f/a/proj` |
| `wmill flow preview ./f/a/b.flow` | `./f/a/b` (steps ran as `./f/a/b/<step>`) | `f/a/b` |
| `wmill flow preview /abs/repo/f/a/b.flow` | backend 400: `Invalid path for preview job: /abs/…` | `f/a/b` |
| `wmill flow preview b.flow` from inside `f/a` | junk path from whatever the cwd made it | `f/a/b` |

A preview job carries no runnable of its own, so this path is the only identity it has: it is what `WM_JOB_PATH` reports, what the runs page links to, and what relative imports inside the previewed code resolve against. Scripts that read `WM_JOB_PATH` fail when it is empty or bogus, which is what the report describes.

## Changes

- `toSyncRootRelativePath` (`cli/src/core/context.ts`): re-expresses a user-supplied file/folder argument as a path relative to the sync root, so `./f/a/b.ts`, `/abs/repo/f/a/b.ts` and `b.ts`-from-inside-`f/a` all reduce to `f/a/b.ts`. The sync root comes from the directory holding `wmill.yaml`, not `process.cwd()` — the cwd only lands there once a config read has chdir'd into it, which `flow preview --remote` and fully-flagged invocations skip. The argument is resolved against the pre-chdir cwd first and the sync root second.
- A file that exists under neither reading is still real if the directory naming it exists — a dbt project's descriptor is deliberately absent — so that reading wins over the sync-root one, which is reserved for arguments the caller should reject.
- Symlinks are resolved only when the lexical result escapes the root (macOS' `/var` → `/private/var`, symlinked checkouts), so a symlinked file *inside* the tree keeps the path it is filed under; resolution goes through the directory when the file itself is not there.
- `assertRemotePath` (same file): a preview whose file has no place in the workspace tree now fails with an explanation naming the argument the user typed, instead of exiting 0 having run nothing (`script preview`) or pushing a made-up path to the backend (`flow preview`).
- `script preview` now derives the path with `scriptPathToRemotePath`, which handles the `__mod` folder layout, instead of an inlined split-at-first-dot. The same value feeds the plain preview, the codebase/bundle preview and relative-import resolution, which had drifted apart.
- `flow preview` normalizes its argument the same way and validates the derived flow path before it reads any files or uploads temp script refs.

One deliberate behavior change beyond the table: `wmill script preview f/myflow.flow/a.inline_script.ts` used to run under `f/myflow` — the *flow's* identity, not the step's — and is now refused. `wmill flow preview --step a` is the supported way to run one step, and gives it `f/myflow/a`.

## Test plan

- [x] `cli/test/preview.test.ts`: regression tests assert the job's own `WM_JOB_PATH` for `script preview` and for a `flow preview` step, across every argument shape (including `--remote` from a subdirectory, the one shape where nothing but the path resolution can move the cwd), plus the `__mod` folder layout and a non-zero exit for a file outside the tree. Both fail on `main` (empty stdout for the `./` script case, `./f/test/job_path_flow/a` for the flow case).
- [x] `cli/test/preview_path_unit.test.ts`: unit coverage for `toSyncRootRelativePath` / `assertRemotePath`, needing no backend — every argument shape, the deliberately-absent dbt descriptor, and that descriptor reached through a symlinked root.
- [x] Full `bun test test/preview.test.ts` against a cargo backend: 17 pass, 1 skip (EE-gated), 0 fail. `test-linux` and `test-windows` green in CI.
- [x] `UNIT_ONLY=1 bun test test/*_unit*.test.ts`: 1052 pass, 0 fail.
- [x] `bunx tsc --noEmit`: 38 pre-existing errors, unchanged from `main`.
- [x] Manually against a dev backend: every row of the table above, checking `WM_JOB_PATH` in the job result and `runnable_path` in `v2_job`, and the runs page listing the jobs under `f/test/foo` / `f/test/myflow` / `f/test/myflow/a`; plus `flow preview --remote` and `--step` from a subdirectory, a descriptor-less dbt project previewed by absolute symlink-aliased path (lands on `f/test/proj`), and both error messages.
- [x] `python system_prompts/generate.py` produces no drift (no command, option or description changed).

### DIFF
--- a/cli/src/commands/flow/flow.ts
+++ b/cli/src/commands/flow/flow.ts
@@ -13,7 +13,12 @@ import { mkdirSync, writeFileSync } from "node:fs";
 import { buildFolderPath, getMetadataFileName, loadNonDottedPathsSetting } from "../../utils/resource_folders.ts";
 
 import { requireLogin } from "../../core/auth.ts";
-import { resolveWorkspace, validatePath } from "../../core/context.ts";
+import {
+  assertRemotePath,
+  resolveWorkspace,
+  toSyncRootRelativePath,
+  validatePath,
+} from "../../core/context.ts";
 import { resolve, track_job, pollForJobResult } from "../script/script.ts";
 import { defaultFlowDefinition } from "../../../bootstrap/flow_bootstrap.ts";
 import { SyncOptions, mergeConfigWithConfigFile } from "../../core/conf.ts";
@@ -605,12 +610,17 @@ async function preview(
     log.setSilent(true);
   }
   const useLocalPathScripts = !opts.remote;
+  // Captured before the config read, which chdirs to the wmill.yaml root.
+  const cwdBeforeConfig = process.cwd();
   if (useLocalPathScripts) {
     opts = await mergeConfigWithConfigFile(opts);
   }
   const workspace = await resolveWorkspace(opts);
   await requireLogin(opts);
   const codebases = useLocalPathScripts ? listSyncCodebases(opts) : [];
+
+  const argPath = flowPath;
+  flowPath = toSyncRootRelativePath(flowPath, cwdBeforeConfig);
 
   // Normalize path - ensure it's a directory path to a .flow or __flow folder
   const isFlowDir = flowPath.endsWith(".flow") || flowPath.endsWith(".flow" + SEP)
@@ -632,6 +642,14 @@ async function preview(
   if (!flowPath.endsWith(SEP)) {
     flowPath += SEP;
   }
+
+  // The flow's windmill path (e.g. "f/cli_smoke/myrelflow"). It is what the
+  // preview job runs under, and the anchor for relative-import resolution:
+  // inline scripts in this flow are treated as living at
+  // "<flow_wm_path>/<step_id>", so "./util" resolves to
+  // "<flow_wm_path_parent>/util" — matching the keys in temp_script_refs.
+  const flowWmPath = stripFlowSuffix(flowPath).replaceAll(SEP, "/");
+  assertRemotePath(flowWmPath, argPath);
 
   // Read and parse the flow definition
   const localFlow = (await yamlParseFile(flowPath + "flow.yaml")) as FlowFile;
@@ -703,12 +721,6 @@ async function preview(
   // too — PathScript modules have already been rewritten to inline rawscript
   // when `useLocalPathScripts` is set, and tempScriptRefs covers relative
   // imports in inline scripts.
-  // Compute the flow's windmill path (e.g. "f/cli_smoke/myrelflow"). Used as
-  // the anchor for relative-import resolution: inline scripts in this flow are
-  // treated as living at "<flow_wm_path>/<step_id>", so "./util" resolves to
-  // "<flow_wm_path_parent>/util" — matching the keys in temp_script_refs.
-  const flowWmPath = stripFlowSuffix(flowPath).replaceAll(SEP, "/");
-
   if (opts.step) {
     await previewStep(opts.step, localFlow, flowWmPath, workspace, input, tempScriptRefs, opts.silent, opts.tag);
     return;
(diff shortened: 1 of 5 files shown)


rubenfiszel:
Round 1 addressed in 8412315515.

**Codex P1 — descriptor-less dbt preview from a nested directory.** Real, fixed. When neither reading of the argument exists on disk, the resolver now prefers the one whose *directory* exists, and only falls through to the sync-root reading when that is missing too. A dbt project's descriptor is deliberately absent, so `wmill script preview wm_dbt.yaml` from inside `f/test/proj__dbt` was mapping to a root-level `wm_dbt.yaml` and getting refused before `isMissingDbtDescriptor` ever saw the adjacent `dbt_project.yml`.

Pinned by a new unit test (`cli/test/preview_path_unit.test.ts`), which fails on the previous commit with `Expected: "f/test/proj__dbt/wm_dbt.yaml" / Received: "wm_dbt.yaml"`. Manually verified end-to-end: previewing that project from its own directory now pushes a job under `f/test/proj` (`v2_job.runnable_path`), reaching the dbt executor rather than the path guard.

**Claude P2 — `flow preview --remote` relies on the chdir inside `getWmillYamlPath()`.** Agreed it was an unstated invariant. It is now stated on `toSyncRootRelativePath` (resolving the sync root is what leaves the process in it, keep that in step if this stops going through `getWmillYamlPath`), and pinned: the flow test grew a fifth invocation, `--remote` from `<tempDir>/f/test`, which is the one shape where nothing else can move the cwd.

**Claude P2 — `assertRemotePath` refuses a dot-containing local folder.** Intended, no change. `wmill script preview f/myflow.flow/a.inline_script.ts` previously ran under `f/myflow` — the *flow's* identity, not the step's — which is the class of wrong path this PR exists to stop. `wmill flow preview --step a` is the supported way to run one step and gives it the correct `f/myflow/a`.

Also added the unit coverage both reviewers suggested for `toSyncRootRelativePath` / `assertRemotePath`, which needs no backend.

rubenfiszel:
Round 2 nit addressed in ad1d1e582b.

**Codex P2 — descriptor-less dbt previews through a symlinked absolute path.** Correct: `realpathSync` needs its target to exist, so for an absent descriptor reached through a symlink alias the catch fell back to the lexical `../…` path and `assertRemotePath` refused it before `isMissingDbtDescriptor` could run. Link resolution now goes through the directory naming the file and reattaches the basename.

Pinned by the unit case you asked for (symlinked root alias + absent dbt descriptor), which fails on the previous commit with `Received: "../wmill_alias_Js79Oe/link/f/test/proj__dbt/wm_dbt.yaml"`. Manually verified: previewing `<root-via-symlink>/f/test/proj__dbt/wm_dbt.yaml` from inside the sync root launches the job under `f/test/proj` (`v2_job.runnable_path`), reaching the dbt executor.

`test-linux` and `test-windows` were green on the round-2 head; the only source change since is the link-resolution line, and the new unit case is skipped on Windows (symlink creation needs a privileged process there).

rubenfiszel:
✅ Review round clean @ ad1d1e582b734f02f7b7ccd114e4fd29ec52d15f — nit-only verdicts at 8412315515a863356bbd15e2a9f93a1264fe2a39 (codex: mergeable with nits, claude: good to merge, pi: good to merge); nit addressed in ad1d1e582b

PR #10607, opened 2026-08-10
feat: keep duckdb spilling behind the local-filesystem fence

rubenfiszel:
## Summary

On a worker that isolates jobs, the DuckDB connection is fenced with
`SET disabled_filesystems='LocalFileSystem'`. DuckDB spills through that same filesystem,
so the fence also costs **out-of-core execution**: any query whose working set outgrows
`memory_limit` fails instead of going to disk. #10581 made that failure legible; this makes
it stop happening.

No combination of stock DuckDB settings gives "spill, but no other local file access":

- `allowed_directories` / `allowed_paths` enforce nothing while `enable_external_access` is
  true — `DBConfig::CanAccessFile` returns early.
- Turning `enable_external_access` off does preserve spilling, but the postgres and mysql
  extensions refuse to `ATTACH` on that flag alone, with no allowlist entry able to
  re-permit it. That would take DuckLake catalogs and datatables down with it.
- A remote `temp_directory` is impossible, and the attempt aborts the process (#10587).

So the capability has to come from the engine.

## The engine patch

A `lock_temp_directory` setting (BOOLEAN, global, irreversible). Setting it

1. fixes `temp_directory` in place — it can no longer be modified, and the setting itself
   cannot be unset or reset;
2. routes the buffer manager's temporary files through a local file system that
   `disabled_filesystems` does not apply to.

The two halves are only safe as a pair: the exemption is sound *because* the directory can
no longer be redirected. Without (1) a script could point `temp_directory` at an arbitrary
path and get a blind write primitive through the exemption. `allowed_directories` still
applies either way, and `temp_directory` was already treated as protected upstream —
`TempDirectorySetting` refuses to modify it when `enable_external_access` is false, and
startup adds it to `allowed_directories` so spilling survives that lockdown.

The isolation transform emits `SET lock_temp_directory=true` **before**
`disabled_filesystems`, while the local filesystem is still reachable, and after the FFI has
pointed `temp_directory` at the job dir. Every statement a script authored runs after all
three settings, so a script can never move the temp directory.

## Reading the engine patch

The patch is a `.patch` blob here, so to review it as a diff, read the upstream
submission: **https://github.com/duckdb/duckdb/pull/24694** — 14 files, +200/−20, against
`main`. (`https://github.com/rubenfiszel/duckdb/pull/1` is the older fork-to-fork view at
`v1.5.5`.)

## Where the patched engine comes from

`backend/windmill-duckdb-ffi-internal` now depends on
[windmill-labs/duckdb-rs](https://github.com/windmill-labs/duckdb-rs) branch
`windmill/duckdb-1.5.5`, pinned by rev: duckdb-rs 1.10505.0 (DuckDB 1.5.5) plus one commit,
rebased onto upstream duckdb-rs `main` ([PR](https://github.com/windmill-labs/duckdb-rs/pull/1)).
`duckdb.tar.gz` is byte-identical across that range, so the engine stays on 1.5.5.
`duckdb.tar.gz` is left byte-identical to upstream's and the patch is applied to the
extracted sources at build time, so `git diff` against `duckdb/duckdb-rs` shows the fork's
whole delta as text.

**The patch is filed upstream as [duckdb/duckdb#24694](https://github.com/duckdb/duckdb/pull/24694)**,
with [#24695](https://github.com/duckdb/duckdb/issues/24695) stating the problem without
prescribing `lock_temp_directory` as the shape — watch the issue for whether this fork can
be retired. Filing was a human call, because DuckDB's `CONTRIBUTING.md` asks contributors
not to submit LLM-generated pull requests, and asks outside contributors to discuss the
change on GitHub first and run CI on a fork before opening one. The patch is prepared in the
shape upstream takes changes — it carries `src/common/settings.json` regenerated with
`scripts/generate_settings.py`, clang-format 11.0.1, and two sqllogictests under
`test/sql/settings/` — and left as a `git format-patch` file that applies with `git am` to
`duckdb/duckdb` at `v1.5.5`. The vendored copy and #24694 are the same change; the
differences the base version forces are enumerated in `docs/duckdb-isolation.md`, which also
records the exit path (delete `crates/libduckdb-sys/windmill/`, go back to the crates.io
crate).

Rejected alternatives: vendoring a patched `libduckdb-sys` here would add a 6 MB binary
tarball to windmill's history per engine bump and would need the Dockerfile FFI stage to
copy more than `backend/windmill-duckdb-ffi-internal`; linking a prebuilt engine through
`DUCKDB_LIB_DIR` would drop the `bundled` build, so every developer and CI job would need a
patched `libduckdb` on the box before `build_dev.sh` could work.

### The patch may not move a struct member

Worth knowing before anyone rebases it. The engine loads the **prebuilt** extensions from
`extensions.duckdb.org`, which are compiled against the struct layouts of the release the
engine claims to be. The first revision declared its new `DatabaseInstance` member next to
`local_db_file_system`, which pushed `create_api_v1` — the extension C-API entry point — 8
bytes along. Every extension load then **hung forever** rather than failing.

The invariant is **no pre-existing member changes offset**, which is strictly stronger than
`sizeof` holding: the first attempt at the `DBConfigOptions` field went in next to
`temporary_directory`, kept `sizeof` at 920 by consuming interior padding, and still slid
four bools along by a byte each. Both fields are now placed to take space the structs
already had — `lock_temp_directory` at the end of the bool run, `temporary_db_file_system`
last in `DatabaseInstance` — and `clang++ -Xclang -fdump-record-layouts` diffs the two trees
to nothing but added lines. The recipe is in the fork's README; a rebase must run it rather
than compare sizes.

## Changes

- `windmill-duckdb-ffi-internal`: depend on the patched fork; two engine-level regression
  tests (below); `build_dev.sh` installs the cdylib through a rename, because overwriting it
  in place SIGBUSes any worker that has it mapped.
- `Dockerfile`: cache `$CARGO_HOME/git` in the FFI builder stage, which the git dependency
  now uses.
- **EE**: splice `SET lock_temp_directory=true` ahead of the fence; drop the
  `SET temp_directory='';` splice and `OOM_HINT` plus the `Out of Memory Error` arm of
  `hint_for`, which only existed to soften the loss of spilling; correct `LOCAL_FS_HINT`,
  which claimed queries cannot spill.
- `docs/duckdb-isolation.md`: why the fork exists, what was ruled out, and what stops the
  patch disappearing.

## Depends on

**#10588** (bump to DuckDB 1.5.5) has merged, so the only remaining dependency is **#10581**
(the diagnostics this partly supersedes), merged into this branch — `git diff main...HEAD` is
now this PR's own 9 files, of which `duckdb_executor.rs` is #10581's.

Complementary to **#10587**: that guard still covers non-isolating workers and the prepare
path, where `temp_directory` stays user-settable.

## What stops the patch disappearing silently

An unpatched engine compiles, links and passes every existing test.

- The fork's build script asserts a marker string is present after applying the patch, and
  panics if the patch no longer applies.
- `patched_engine_tests::locking_the_temp_directory_keeps_spilling_available` runs a query
  that must spill, first without the lock (asserting the fence stops it — which is what
  proves the query really spills) and then with it. It fails outright on a stock engine.
- `patched_engine_tests::prebuilt_extensions_still_load` guards the ABI finding above. It is
  a timeout, not an assertion, because the failure mode is a hang.
- The transform fails closed: on a stock engine `SET lock_temp_directory=true` errors, so
  every DuckDB job on an isolating worker fails immediately rather than quietly losing the
  ability to spill. **Rehearsed**, not just reasoned about — the FFI was repointed at the
  crates.io crate, rebuilt, and an isolated job run. It produced DuckDB's bare
  `unrecognized configuration parameter "lock_temp_directory"`, which is loud but says
  nothing about why, so `hint_for` now names the missing patch and the two ways out.

## Test plan

**DuckDB's own suite**, against the patched v1.5.5 tree: `1 017 053` assertions across `4216` test cases, all passing. It failed once first —
`test_reset.cpp` enumerates every setting and asserts a set/reset round-trip, which an
irreversible setting cannot satisfy, so `lock_temp_directory` joins the exclusion list
`disabled_filesystems`, `lock_configuration` and `enable_external_access` are already on.
Both new sqllogictests pass.

**FFI crate**: 18 tests pass, including the two new ones.

**Record layouts**: `DBConfigOptions` and `DatabaseInstance` dumped from the stock and
patched `src/include` and diffed — the only difference is added lines; no pre-existing
member changes offset.

**Real preview jobs** on a worker started with `ENABLE_UNSHARE_PID=true` on its own tag,
against a CE server built with `quickjs,duckdb,parquet,private`:

| # | Case | Result |
|---|---|---|
| 1 | `memory_limit`, `temp_directory`, `lock_temp_directory` as the transform set them | `488.2 MiB`, the job dir, `true` |
| 2 | 5M rows × 200-byte strings under the fence | ✅ 5 000 000 rows, 1 000 000 000 bytes |
| 3 | Same, reading `duckdb_temporary_files()` | ✅ **4 files, 205 422 592 bytes actually spilled** while `disabled_filesystems` is on |
| 4 | `read_text('/etc/hostname')` | ✅ blocked, with the corrected hint |
| 5 | `COPY (…) TO '/tmp/…'` | ✅ blocked |
| 6 | `INSTALL shellfs FROM community` | ✅ refused by the allowlist |
| 7 | `SET temp_directory='/etc/duckdb-escape'` | ✅ refused — the temp directory has been locked |
| 8 | `SET lock_temp_directory=false` | ✅ refused |
| 9 | `SET disabled_filesystems=''` | ✅ refused — cannot be re-enabled |
| 10 | `ATTACH '$res:…' AS pg (TYPE postgres)` | ✅ reads the catalog |
| 11 | Same, plus a spilling query in the same job | ✅ both, 5 temp files |
| 12 | DuckLake write, 250 000 rows | ✅ a real 3.7 MB parquet data file through the S3 proxy |
| 13 | DuckLake read back | ✅ same count, sum, bucket count and `md5` digest |
| 14 | DuckLake read while a second query spills | ✅ both, 5 temp files |
| 15 | `DISABLE_DUCKDB_FS_SANDBOX=true` worker | ✅ opt-out unaffected: local read succeeds, spilling succeeds |
| 16 | `// materialize ducklake://main/orders` + `// on_schema_change=sync`, 200 000 rows | ✅ materialized, snapshot 2 |
| 17 | Same script with a column added — exercises the sync **pre-pass**, which opens its own connection | ✅ column picked up, snapshot 3 |
| 18 | Read the materialized table back | ✅ 200 000 rows, 7 distinct values in the added column |
| 19 | **Rollback rehearsal**: same worker, FFI rebuilt against the stock crates.io engine | ✅ fails on the first statement, with the new explanation |
| 20 | Job-dir cleanup: cancel a job mid-spill (8 temp files, 1.6 GB live) | ✅ no `duckdb_temp_*` left, job dir gone |

Worker log clean after all of it — no panic, no abort, worker still alive.

The whole matrix was re-run after the `DBConfigOptions` field was relocated, so the numbers
above are from the shipped engine.

## Known follow-up

`configure_duckdb_resource_limits` sets `memory_limit` and `temp_directory` but not
`max_temp_directory_size`, which DuckDB defaults to ~90% of the temp filesystem's free
space. Spilling was impossible on an isolating worker before this PR, so a runaway query
just failed; now one job can fill the disk. **Observed, not hypothetical** — while testing
job-dir cleanup, a deliberately oversized `CREATE TABLE` spilled 8 files and died with
`IO Error: Could not write file` after exhausting the temp filesystem.

Left as a follow-up rather than guessed at here: it is parity with non-isolating workers and
with every other language runtime rather than a new class of risk, and picking the bound is
a policy call (a `DUCKDB_MAX_TEMP_DIRECTORY_SIZE` knob alongside `DUCKDB_MEMORY_LIMIT` is
the obvious shape). It does become reachable on this worker class for the first time
*because* of this change, so it should not get lost.


rubenfiszel:
EE companion: https://github.com/windmill-labs/windmill-ee-private/pull/720
Patched engine fork: https://github.com/windmill-labs/duckdb-rs/tree/windmill/duckdb-1.5.5 (rev `9f2d604`)
Upstream-ready patch (not filed, see PR body): [`crates/libduckdb-sys/windmill/0001-lock-temp-directory.patch`](https://github.com/windmill-labs/duckdb-rs/blob/windmill/duckdb-1.5.5/crates/libduckdb-sys/windmill/0001-lock-temp-directory.patch)

rubenfiszel:
### Round 3 — claude and pi are go; the one remaining codex nit is out of scope

codex flags `Cargo.toml:8` and `src/lib.rs:718` as five-line comment blocks over AGENTS.md's ≤4-line limit. Both predate this PR — they are #10588's, and codex sees them only because that PR is merged into this branch:

```
$ git show origin/duckdb-bump-1-5-5:backend/windmill-duckdb-ffi-internal/Cargo.toml | sed -n '7,11p'
# `rust_decimal` is optional upstream. It stays on because its ...   (5 lines, unchanged here)

$ git show origin/duckdb-bump-1-5-5:backend/windmill-duckdb-ffi-internal/src/lib.rs | sed -n '718,722p'
# `Value` is `#[non_exhaustive]`: a newer engine can hand back ...   (5 lines, unchanged here)
```

Editing another PR's comments from this one would just create a conflict, so leaving them. The two blocks this PR *did* add over the limit — the fork dependency comment and the `patched_engine_tests` module comment — were condensed in `f16b4b9`.

Round 2's two nits are both fixed: the extension-load guard now asserts instead of skipping when `CI` is set (a skip and a pass were indistinguishable in CI, so the ABI guard could have evaporated silently), and the misplaced `SPILLING_QUERY` doc comment now names where the limit actually lives.

Also closed the last manual gap both rounds asked about — **job-dir cleanup**: cancelled a job mid-spill with 8 temp files and 1.6 GB live in the job dir; no `duckdb_temp_*` survives and the job dir is gone. While setting that up, an oversized `CREATE TABLE` exhausted the temp filesystem and died with `IO Error: Could not write file`, so the `max_temp_directory_size` follow-up is now an observation rather than a hypothesis. It is written up in the PR body as a follow-up, with the reasoning for not fixing it here.

Staying a draft until #10588 and #10581 land, since it is stacked on both.

rubenfiszel:
### Rebased onto main now that #10588 has landed

- Merged `main` (which carries the squash-merged #10588). The EE side conflicted on `EXTENSION_ALLOWLIST`, since #716 grew it — resolved by taking main's newer list (`delta`, `unity_catalog`, `encodings`, `aws`, `fts` and the reworded rationale) and keeping the `lock_temp_directory` splice. `git diff main...HEAD` is now just this PR's 9 files.
- **Only remaining dependency is #10581.**
- The fork was rebased onto upstream duckdb-rs `main` and repinned (`42c650a`). `duckdb.tar.gz` is byte-identical across that range, so the engine is still DuckDB 1.5.5; main also carries *Fix DuckDB linking with custom target dirs*, which is relevant to the `CARGO_TARGET_DIR` this crate is built with.

### To read the engine patch as a diff

https://github.com/rubenfiszel/duckdb/pull/1 — `v1.5.5` vs `v1.5.5 + the commit`, both branches on a fork, 14 files, +194/−20. Nothing filed with duckdb/duckdb.

### Re-verified after the merge

18 FFI tests, 4 EE tests, and 17 real preview jobs on an isolating worker: 197 MB spilled under the fence, every escape refused, ATTACH postgres, DuckLake 250k round-trip (same `md5` digest as before the merge: `7a19940b525f69d6899a630f91cbc54e`), and the `on_schema_change=sync` pre-pass. Worker log clean.

PR #10612, opened 2026-08-10
feat: expose every runs filter on the open_page chat tool

Guilhem-lm:
## Summary

`open_page` could only take the Runs page to five of its filters (`path`, `status`, `schedule_path`, `job_kinds`, `user`), so a request like "show me the failed runs on worker X from the last 24 hours that a schedule didn't start" lost most of its criteria — the model either dropped them or burned turns trying to express them. The tool now accepts the page's whole filter set, and refuses the values and combinations the page can only fail silently on.

## Changes

- **New Runs params**: `folder`, `job_trigger_kind`, `label`, `tag`, `worker`, `concurrency_key`, `arg`, `result`, `search` (the page's broad free-text filter, `_default_` in the URL), `timeframe` (enum of the page's own relative windows), `min_ts` / `max_ts`, `resolved`, `show_skipped`, `show_future_jobs`, `all_workspaces`.
- **Multi-value and exclusion syntax is documented per field** — the page packs several values into one param as a comma list, with `!` for exclusion on the negatable ones and `*` wildcards on `label`/`tag`/`worker`.
- **List filters are normalized, not just documented.** The backend reads such a list with the polarity of its **first** item and matches the rest verbatim, so `"a, b"` would filter on `" b"` (and 400 for a trigger kind) and `"a,!b"` would quietly match `b` as an *inclusion*. Items are trimmed and a mixed include/exclude list is refused.
- **`folder` is single and positive, and enforced.** The page turns it into one `f/<folder>/` path prefix, so a comma or `!` lands inside the prefix and matches nothing; the handler rejects both rather than opening an empty page.
- **Combinations the page silently drops are refused.** A `concurrency_key` switches the page to its extended-jobs query, which has no `worker`, free-text or queue-status parameter at all — it would render those chips and apply none of them. Pairing `concurrency_key` with `worker`, `search`, or a `waiting`/`suspended` status now returns an error instead (the other four statuses map through that view correctly).
- **`all_workspaces` is gated** exactly as the page gates it (superadmin/devops in the `admins` workspace) — narrowed out of the advertised schema otherwise, and dropped fail-closed in the handler. `RUNS_FILTER_KEYS` is now built with the permissive schema so the key exists at all; the backend independently requires the `admins` workspace, so no authorization rests on this narrowing.
- **Malformed values fail closed**: a non-object `arg`/`result` JSON, an unknown `job_trigger_kind`, or an unparseable bound returns an actionable tool error. Left alone the page drops the first two without a word and 400s on the third behind a generic toast.
- **Absolute bounds normalized**: a bare `YYYY-MM-DD` is read as local midnight (matching the page's own date picker rather than shifting by the viewer's UTC offset — documented on both bounds, since on `max_ts` that excludes the named day), and an absolute bound suppresses `timeframe`, which the page ignores anyway.
- System prompt now tells the chat to carry over every filter the user described, not just the request's main clause.

## Screenshots

Runs page opened from a tool-built URL carrying every new filter (chip row scrolls; `min_ts`, `path`, `user` visible, timeframe pill reads the absolute bound):

runs-all-filters

A realistic multi-filter request — `timeframe` + `label` + `worker` + `!schedule`:

runs-multi-filter

## Test plan

- [x] `ai_evals` A/B on global mode, sonnet, 4 open_page cases (2 pre-existing controls + 2 new): **2/4 → 4/4**. Before the change the new cases fail — one burns all 6 turns and 122k tokens flailing at filters that don't exist (`global-openpage8-runs-label-and-worker`), the other silently omits the exclusion (`global-openpage9-runs-exclude-schedules`). After: both pass, and case 8 costs 40.8k. Per-iteration window tax for the added schema + prompt line, measured on the unchanged control cases: **~+1020 `finalContextTokens`** (19448 → 20471, 19437 → 20457).
- [x] Unit tests: the emitted param set equals the Runs filter schema's key set exactly (so a renamed or added page filter can't be silently swallowed by `buildFilterUrl`'s allow-list), the ten fail-closed rejections, list-item trimming, and the bare-date / timeframe-precedence rules. `core.test.ts` 190 passed, `AIChatManager.test.ts` 124 passed, `check:fast` clean.
- [x] Browser: loading `/runs` with every param applied renders chips for all of them (`path`, `user`, `folder`, `label`, `tag`, `worker`, `concurrency_key`, `status`, `show_skipped`, `resolved`, `job_trigger_kind: !schedule`, `arg`, `show_future_jobs: false`, free text) and the timeframe reads from `min_ts`.
- [ ] Reviewer sanity check: ask the chat for a multi-criteria runs view and confirm the opened page carries every criterion.

Not addressed here, deliberately: the Runs page's own searchbar offers `allowMultiple`/`allowNegative` on `folder` and lets you combine `concurrency_key` with `worker`/`search`/`waiting`/`suspended`, all of which fail the same silent way when typed by hand. Fixing that needs a page change (and, for the concurrency view, new API query params), so this PR only stops the tool from emitting those shapes.


Guilhem-lm:
✅ Review round clean @ a5b6e1c47f0fbd74abf5f660ea135e0351388844

PR #10614, opened 2026-08-10
fix(duckdb): cast list columns in quicksearch so tables containing them can be previewed

principalwater:
### Problem

A DuckLake (or DuckDB) table with a `LIST` or `ARRAY` column cannot be previewed at all. The grid fails, and so does its row count:

```
Binder Error: Cannot concatenate types VARCHAR, VARCHAR, BIGINT, INTEGER, ...,
VARCHAR[], DOUBLE, ..., BOOLEAN and TIMESTAMP WITH TIME ZONE - an explicit cast
is required

LINE 1: ... FROM "raw"."zenmoney__accounts" WHERE ($1 = '' OR CONCAT(' ', "id",
"changed", "user_id", ...
```

### Cause

Quicksearch concatenates every visible column. DuckDB's `concat` doubles as *list* concatenation, so a `LIST`/`ARRAY` argument next to a `VARCHAR` is ambiguous and rejected outright, where every other type is implicitly cast to text:

```
D SELECT CONCAT(' ', ['a','b']);
Binder Error: Cannot concatenate types VARCHAR and VARCHAR[] - an explicit cast is required
```

Only the list family is affected. Checked against the bundled engine (DuckDB 1.5.2), one column per type: `BOOLEAN`, every int width, `FLOAT`/`DOUBLE`/`DECIMAL`, `VARCHAR`, `BLOB`, `DATE`, `TIME`, `TIMETZ`, `TIMESTAMP`, `TIMESTAMPTZ`, `INTERVAL`, `UUID`, `JSON`, `BIT`, `ENUM`, `STRUCT`, `MAP` and `UNION` all concatenate; `VARCHAR[]`, `INTEGER[3]` and `VARCHAR[][]` are the three that fail.

### Fix

Cast the list columns, and only those, inside the quicksearch predicate. The comparison is textual either way, so no result changes, and the projection is left untouched: casting there would change the types the caller reads back.

Casting *every* column also fixes the error, but it rewrites the query for tables that never had a problem, and that string is not free to change: the frontend builder below is hashed into the policy of every deployed Database Studio app (`rawscript/<sha256(content)>` in `triggerables_v2`). A changed string turns a working app into `Path rawscript/... forbidden by policy` until it is redeployed. Restricting the cast to the columns DuckDB actually rejects keeps the emitted SQL byte-identical for every table without a list column, so only tables that are already broken see their query change.

### Note for the release notes

An app deployed *before* this fix, against a table that **has** a list column, swaps its binder error for `Path rawscript/... forbidden by policy` until it is redeployed. It was already unusable either way, so this is not a new breakage, but the error text changes. Tables without a list column are unaffected, which is what the narrow cast buys.

### Two code paths, both fixed

- **Backend, and the one that produced the error above.** The Database Manager does not send SQL; it sends a `-- WM_INTERNAL_DB_SELECT {...}` marker carrying the table and its column definitions, which `query_builders.rs` expands. That is the live path for DuckLake browsing.
- **Frontend.** `AppDbExplorer` builds SQL client-side for the app Database Studio component, with the same defect in `dbtable/queries/{select,count}.ts`.

They are independent implementations of the same predicate, so both are fixed; leaving either would fix one feature and not the other. Both are covered at `SELECT` and at `COUNT`, since fixing only one leaves the grid rendering while the row count still errors.

### Verification

`cargo test -p windmill-common --lib query_builders` → 187 passed. The one new test pins that a list column is cast and a scalar is not, in both the `SELECT` and the `COUNT`; it fails without the fix (confirmed by disabling the predicate and re-running). The pre-existing DuckDB assertions are untouched, which is the point: their SQL does not move.

Exercised end-to-end against a live DuckLake table (`accounts`, with `sync_id VARCHAR[]`) on a backend built with `--features quickjs,duckdb,parquet,private`:

| | result |
|---|---|
| before, same table, same engine | `Binder Error: Cannot concatenate types VARCHAR, VARCHAR, VARCHAR, VARCHAR[], DECIMAL(5,2), BOOLEAN and TIMESTAMP WITH TIME ZONE` |
| after, `SELECT` + `COUNT` | 3 rows, count 3, array column intact |
| after, search `Tinkoff` (scalar column) | 1 row |
| after, search `s2` (inside the array) | 1 row |

**Database Manager** (backend marker path): grid, row count and quicksearch on the array column:

![Database Manager preview of a DuckLake table with a VARCHAR[] column](https://raw.githubusercontent.com/windmill-labs/agent-screenshots-internal/main/shots/duckdb-cast/1786384768-wm-dbm-1.png)

Quicksearch "s2" matching inside the array column, count 1 of 1

**Database Studio app component** (frontend builder path): same table, quicksearch `s2` matching inside `sync_id`:

Database Studio app component, quicksearch s2 on the same table

The byte-identity claim was also checked the way it actually matters, against a **deployed** app rather than a draft: an app was deployed over a DuckLake table with no list column while the frontend was still at `main`, so its policy carried the pre-fix digests (`a:rawscript/941c1fa8…`, `a_count:rawscript/cdd20261…`). Loading that same deployed app with this branch's frontend renders and searches normally, with no `forbidden by policy`. Casting every column would have broken exactly this.

Deployed pre-fix app running on the patched frontend, quicksearch returning 1 of 1

Finally, a table mixing all the nested kinds (`tags VARCHAR[]`, `meta STRUCT(city VARCHAR, n INTEGER)`, `attrs MAP(VARCHAR, INTEGER)`) previews, and quicksearch on `Berlin`, which lives inside the **uncast** `STRUCT`, returns 1 of 1. Leaving `STRUCT` and `MAP` uncast costs nothing in search behavior. (DuckLake rejects fixed-size `INTEGER[3]` at write time, so that spelling is covered by the unit tests rather than end to end; it reaches this predicate only on plain DuckDB.)

Quicksearch matching inside an uncast STRUCT column on a table that also has a list and a map

### DIFF
--- a/backend/windmill-common/src/query_builders.rs
+++ b/backend/windmill-common/src/query_builders.rs
@@ -524,6 +524,27 @@ fn cols_to_simple(cols: &[ColumnDef]) -> Vec<SimpleColumn> {
         .collect()
 }
 
+/// DuckDB's `CONCAT` implicitly casts scalars — `VARCHAR`, `BIGINT`, `DOUBLE`,
+/// `BOOLEAN`, `DATE` and `TIMESTAMPTZ` all concatenate — but rejects nested
+/// types:
+///
+/// ```text
+/// D SELECT CONCAT(' ', ['a','b']);
+/// Binder Error: Cannot concatenate types VARCHAR and VARCHAR[] - an explicit
+/// cast is required
+/// ```
+///
+/// Quicksearch concatenates every visible column, so a single LIST, STRUCT or
+/// MAP column makes the whole table impossible to preview. Casting each column
+/// is also what a text search wants: the comparison is textual either way.
+fn duckdb_searchable_columns(columns: &[String]) -> String {
+    columns
+        .iter()
+        .map(|col| format!("CAST({} AS VARCHAR)", col))
+        .collect::<Vec<_>>()
+        .join(", ")
+}
+
 /// MSSQL `text`, `ntext`, and `image` types cannot be used with the `=` operator.
 /// This function wraps the column/param in CAST(...) when needed.
 fn mssql_needs_cast_for_eq(datatype: &str) -> bool {
@@ -1019,7 +1040,7 @@ pub fn make_select_query(
 
             let quicksearch = format!(
                 "($quicksearch = '' OR CONCAT({}) ILIKE '%' || $quicksearch || '%')",
-                filtered_columns.join(", ")
+                duckdb_searchable_columns(&filtered_columns)
             );
 
             query.push_str(&format!(
@@ -1188,7 +1209,7 @@ pub fn make_count_query(
             if !filtered_columns.is_empty() {
                 quicksearch_condition.push_str(&format!(
                     " ($quicksearch = '' OR CONCAT(' ', {}) LIKE CONCAT('%', $quicksearch, '%'))",
-                    filtered_columns.join(", ")
+                    duckdb_searchable_columns(&filtered_columns)
                 ));
             } else {
                 quicksearch_condition.push_str(" ($quicksearch = '' OR 1 = 1)");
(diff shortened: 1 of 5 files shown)


principalwater:

rubenfiszel:
Addressed the P2 in `ea008b6872`: `duckdbQuicksearchColumns.test.ts` next to the helper, asserting the cast case, the byte-identical no-list case (compared directly against `buildVisibleFieldList(columnDefs, 'duckdb').join(', ')`), and that ignored columns are excluded. 3 passing.

Also ran the manual check you flagged, since it is the failure mode this design exists to avoid. I deployed a Database Studio app over a DuckLake table with no list column **while the frontend was still at `main`**, so its policy carried the pre-fix digests (`a:rawscript/941c1fa8…`, `a_count:rawscript/cdd20261…`), then loaded that deployed app with this branch's frontend. It renders and quicksearches normally, no `forbidden by policy`. Screenshot and digest list added to the PR description.

On the release-note point: agreed, and worth stating plainly. An app deployed before this fix against a table **with** a list column swaps one broken state for another until it is redeployed. It was already unusable (binder error), so this is not a new breakage, but the error message changes.

Re-running the round: Codex could not complete on the last one (`bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`), so it had no verdict.

rubenfiszel:
### Round 2 @ `ea008b6872e58bb331d6cd3161f116af4db2992d`

| reviewer | verdict |
|---|---|
| Claude | Good to merge, no findings |
| Pi | Good to merge, no findings |
| Codex (CI) | unavailable, see below |
| Codex (local, same model/policy/effort) | **Good to merge**, no issues found |

**Codex in CI cannot review this PR because it is from a fork.** Both attempts died before reading anything:

```
warning: Codex could not find bubblewrap on PATH ... Codex will use the bundled bubblewrap in the meantime
bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted
```

This is not a transient failure and not specific to this change. Codex reviewed org-branch PRs #10619 and #10607 cleanly at 18:18, minutes between this PR's two failures, and the `codex-pr-review` run for `pr/duckdb-cast` sits at `action_required` (the fork gate). No fork PR in this repo has ever received a Codex review. The bundled bubblewrap cannot create a network namespace on the restricted fork runner. Worth a separate infra issue; re-running will not help.

As a substitute I ran `.agents/skills/local-review-codex` locally, which is the same `REVIEW.md` policy, `gpt-5.6-sol`, `xhigh` reasoning as the CI job: **Good to merge, no issues found.**

Codex's local pass asked for one scenario I had not covered, a table mixing `STRUCT` and `MAP` with a list. Done and added to the description: quicksearch on `Berlin`, which lives inside the **uncast** `STRUCT` column, returns 1 of 1, so leaving those types uncast costs nothing in search behavior. DuckLake rejects fixed-size `INTEGER[3]` at write time, so that spelling stays covered by the unit tests; it only reaches this predicate on plain DuckDB.

One correction to the Pi review for the record: it states the `]` sniff also fires on `MAP(VARCHAR, INTEGER[])` and `STRUCT(a INTEGER, b VARCHAR[])`. It does not. Checked against DuckDB 1.5.2, those render as `MAP(VARCHAR, INTEGER[])` and `STRUCT(x INTEGER, y VARCHAR[])`, both terminating in `)`. Only top-level lists end in `]`, so the heuristic is narrower than Pi assumed, which is the direction that protects byte-identity.

PR #10619, opened 2026-08-10
fix: raw app new-app modal ignores instance-level AI settings

rubenfiszel:
## Summary

Fixes WIN-2343.

The raw-app "New App setup" modal read the global `copilotInfo` store and rendered
**"AI is not configured for this workspace"** whenever `enabled` was false. That store
is empty until *some ancestor* fetches the config, so `enabled === false` conflates two
very different states: "this workspace has no AI providers" and "nobody has told us yet".

On 1.778.0 the second state was permanent for this route. The root layout only called
`loadCopilot` when the docked chat was enabled (`!disableAi`), and `disableAi` is `true`
for every non-operator once the AI Sessions beta gate is on — so the workspace AI config
was never loaded on any page that lacks its own `AiChatLayout`. Script and flow editors
mount their own (`ScriptWrapper` / `FlowWrapper`) and kept working, which is exactly the
reported asymmetry: AI configured at the instance level worked for scripts and flows but
the raw-app modal insisted it had to be set in workspace settings.

99e7661a1b (v1.779.0) untied the load from `disableAi`, which fixes the permanent case.
This PR removes the fragility that made it possible: the picker no longer trusts an
ancestor to have loaded the config, and never states AI is unconfigured until it has the
evidence. Without that, the alert still lies during the initial fetch — and this route
forces a full page reload, so the modal always opens on a cold store.

## Changes

- `RawAppTemplatePicker.svelte`: gate the AI section on the config actually describing the
  operating workspace (`copilotWorkspace === opWs`), not on `enabled` alone. While unknown,
  show "Loading AI settings..." instead of the alert.
- `RawAppTemplatePicker.svelte`: load the config for the operating workspace when the modal
  opens and nothing else has, so the claim owns its own evidence.
- `loadCopilot.ts`: share the in-flight request between callers loading the same workspace
  (the layout and the picker mount together) instead of firing one GET each. Only the newest
  load applies its result, so only the newest is shareable — a superseded workspace still
  gets a fresh request so it can win the token back.
- Reword the alert title to "AI is not configured." — with instance-level settings the
  config need not be a workspace one.
- Regression test for the `loadCopilot` sharing/token interaction.

## Screenshots

Instance-level AI configured (`global_settings.ai_config`), workspace `ai_config` NULL.
"Before"/"after-loading" are captured with `get_copilot_info` held open, which is the state
1.778.0 was stuck in permanently.

**Before** — the modal asserts AI is unconfigured while the config is merely unknown:

before

**After, config not yet known** — no claim either way:

after-loading

**After, config resolved from the instance settings** — the AI prompt is offered:

after-enabled

## Test plan

- [x] `npm run check` — 0 errors (70 pre-existing warnings, none in the touched files)
- [x] `npx vitest run src/lib/components/copilot/loadCopilot.test.ts src/lib/aiStore.test.ts` — 4 passed
- [x] Instance-level AI config set, workspace config empty → `/apps_raw/add` offers "Start with AI"
- [x] `get_copilot_info` held open → "Loading AI settings...", no alert (old code showed the alert)
- [x] No AI config anywhere → alert still shown after the fetch resolves
- [x] Exactly one `get_copilot_info` request on a cold load of `/apps_raw/add`, despite both
      the layout and the picker asking for it

---
## Summary by cubic
Fixes the raw-app “New App” modal showing “AI is not configured” before the config is known, respects instance-level AI settings, and shows an instance settings link for superadmins (even if they aren’t workspace admins). Addresses WIN-2343.

- **Bug Fixes**
  - Require a resolved workspace; only trust AI config when `copilotWorkspace === opWs`; show “Loading AI settings...” until known.
  - Load the operating workspace’s config on modal open; update alert title to “AI is not configured.”
  - Share in-flight `get_copilot_info` calls per workspace; keep only the newest request and allow a re-requested workspace to win back; remove the unused token guard.
  - Link workspace settings via `{base}/workspace_settings?tab=ai`; for `$superadmin`, also show an instance settings link even when not a workspace admin.

<sup>Written for commit fd5a70cb3cef6d13bb709a1545e39f4f0ddb985a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/copilot/loadCopilot.test.ts
+++ b/frontend/src/lib/components/copilot/loadCopilot.test.ts
@@ -1,0 +1,53 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { get } from 'svelte/store'
+
+vi.mock('./lib', () => ({
+	workspaceAIClients: { init: () => {} }
+}))
+
+let requests: { workspace: string; resolve: (config: any) => void }[] = []
+
+vi.mock('$lib/gen', () => ({
+	WorkspaceService: {
+		getCopilotInfo: ({ workspace }: { workspace: string }) =>
+			new Promise((resolve) => requests.push({ workspace, resolve }))
+	}
+}))
+
+import { loadCopilot } from './loadCopilot'
+import { copilotWorkspace } from '$lib/aiStore'
+
+const config = { providers: { anthropic: { resource_path: 'u/admin/anthropic', models: ['m'] } } }
+
+describe('loadCopilot', () => {
+	beforeEach(() => {
+		requests = []
+		copilotWorkspace.set(undefined)
+	})
+
+	it('shares one request between callers loading the same workspace', async () => {
+		const first = loadCopilot('ws')
+		const second = loadCopilot('ws')
+
+		expect(requests.map((r) => r.workspace)).toEqual(['ws'])
+		requests[0].resolve(config)
+		await Promise.all([first, second])
+		expect(get(copilotWorkspace)).toBe('ws')
+	})
+
+	it('lets a re-requested workspace win back over the one that superseded it', async () => {
+		const a1 = loadCopilot('a')
+		const b = loadCopilot('b')
+		// Sharing `a1` here would leave `a` unapplied: only the newest request's
+		// result is written, and `a1` already lost that race to `b`.
+		const a2 = loadCopilot('a')
+		expect(requests.map((r) => r.workspace)).toEqual(['a', 'b', 'a'])
+
+		requests[0].resolve(config)
+		requests[1].resolve(config)
+		requests[2].resolve(config)
+		await Promise.all([a1, b, a2])
+
+		expect(get(copilotWorkspace)).toBe('a')
+	})
+})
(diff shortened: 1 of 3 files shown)


rubenfiszel:
✅ Review round clean @ fd5a70cb3cef6d13bb709a1545e39f4f0ddb985a — nit-only verdicts at eedddca32c7e7d3a994216dadaa806ec8f9b5624 (codex: mergeable/nits, claude: mergeable/nits, pi: good to merge); the shared nit (instance-settings link nested inside the workspace-admin branch, so a superadmin who is only a member never saw it) is addressed in fd5a70cb3c and verified in the browser against a superadmin non-admin member.

PR #10620, opened 2026-08-10
fix(smtp): explain why a test email failed instead of 'deadline has elapsed'

rubenfiszel:
## Summary

Fixes WIN-2344.

Testing SMTP against `smtp.office365.com:587` with invalid credentials returned `deadline has elapsed` — a `tokio::time::Elapsed` string that says nothing about what went wrong, whether the credentials were the problem, or which toggle to flip.

Two things produced it:

- The test endpoint gave the whole session a **3s** budget. A remote provider needs longer than that for TCP + TLS + EHLO + AUTH, so the timeout fired before the server ever said why it refused.
- The budget wrapped `SmtpClientBuilder::connect()`, which performs the TCP connect, the TLS handshake, EHLO **and** authentication as one opaque future. Even with a longer budget, a stall anywhere in there is reported identically.

That second part matters more than the timeout value: Microsoft 365 answers credentials it refuses with **silence**, not a 535. Verified against the real host — `AUTH LOGIN` with a bogus user gets no reply at all, and the connection is eventually dropped. No timeout value can turn that into a useful message unless the failing *step* is known.

So the session is now driven step by step (connect → TLS → greeting → EHLO → auth → send), all steps sharing one deadline, and every failure names the step plus the setting that would fix it.

Before / after for the exact reported configuration:

```
deadline has elapsed
```
```
SMTP error while authenticating with smtp.office365.com:587: timed out after 20s: the
connection and TLS succeeded but the server never answered the authentication attempt,
which is how several providers (Microsoft 365 among them) treat credentials they refuse:
check the username and password, and that the mailbox is allowed to authenticate over SMTP.
```

## Changes

Backend (EE, `windmill-common/src/email_ee.rs` — companion PR windmill-labs/windmill-ee-private#721):
- Drive the SMTP session through mail-send's per-step primitives instead of `connect()`, so a failure carries the step it happened at. The wire conversation is unchanged for all three transports (plain, STARTTLS, implicit TLS).
- Map every `mail_send::Error` to a message naming the likely setting: TLS mode vs port (465 wants implicit TLS, 587/25 want STARTTLS), `ACCEPT_INVALID_CERTS` for self-signed certs, app-password / SMTP-AUTH guidance for rejected credentials, and STARTTLS advice when the server refuses it.
- One shared deadline across the steps, so labelling the steps cannot multiply the caller's timeout. It also bounds the message transfer, which previously had no timeout at all.

Backend (OSS):
- Raise the test-email budget from 3s to 20s — the previous value could not outlast a normal remote handshake.
- Return the SMTP diagnostic as the response body instead of `Internal: … @<source location>`.
- Report `501` when the build has no SMTP support at all, instead of the current silent "sent successfully" on a build whose send is a no-op.

Frontend:
- Show the failure in a persistent `Alert` under the test row instead of a 5s toast, read from `error.body` (which carries the message) rather than `error.message` (which does not).
- Spinner on the button while the test runs.

## Test plan

Exercised against a scriptable SMTP server and the real `smtp.office365.com`, with a backend built `--features quickjs,private,windmill-common/smtp`:

- [x] Successful send over plain SMTP, STARTTLS and implicit TLS (self-signed + `ACCEPT_INVALID_CERTS`); the server-side transcript is `EHLO → [STARTTLS → EHLO] → AUTH → MAIL → RCPT → DATA` in all three, identical to before
- [x] Send with no credentials configured (auth skipped, message delivered)
- [x] `smtp.office365.com:587` + invalid credentials → "timed out … never answered the authentication attempt"
- [x] Server replying `535` → "the server rejected the credentials (535 …)"
- [x] Server without STARTTLS → "the server refused STARTTLS (502 …)"
- [x] `smtp.gmail.com:465` with implicit TLS off → "enable 'Implicit TLS'"
- [x] Implicit TLS against a plaintext server → "disable 'Implicit TLS'"
- [x] Nothing listening → "Connection refused (os error 111)"
- [x] Non-SMTP port → "the reply could not be parsed as SMTP"
- [x] A build without SMTP support (`--features quickjs`) answers `501` with the "compiled without SMTP support" message instead of the previous silent success
- [x] Editing any SMTP field retires the failure alert, so it can never describe settings that are no longer in the form
- [x] `npm run check` (0 errors); `cargo check -p windmill-api-settings` in the CE, `private` and `private,smtp` configurations, plus `-p windmill-common` for the latter two

## Screenshots

Credentials shown are fabricated test values against a local dev instance.

Reported case — Microsoft 365 with invalid credentials:

wm-smtp-3-error

A server that does answer with 535:

wm-smtp-5-authfail

Success clears the alert:

wm-smtp-4-success

A build compiled without SMTP support, which previously reported "Test email sent successfully" while sending nothing:

wm-smtp-6-nosmtp

---
## Summary by cubic
Improves SMTP test email errors so admins see which step failed and how to fix it, instead of the vague “deadline has elapsed.” Addresses WIN-2344 and shows a clear, persistent error in the settings page.

- **Bug Fixes**
  - Backend: surface user-readable SMTP diagnostics in the response body (400) while preserving non-SMTP status codes; return 501 when SMTP isn’t compiled; raise the test email timeout from 3s to 20s; add `SMTP_ENABLED` (false in non-`private` builds).
  - Frontend: show failures in a persistent `Alert`, read `error.body`, add a loading spinner on “Send test email,” and retire the alert when settings change.

<sup>Written for commit bb6b81391d4d67e47454073e59058f4e9a767e41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10620?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 617fc594054cf06269baab91fb827d275ca6a61a — nit-only verdicts at 59241f3285798db0072f16f29f15512036345fe9 (codex: Good to merge, pi: Good to merge, claude: Mergeable, but should ideally address nits); all three nits addressed in 617fc59405:

- `cargo check -p windmill-api-settings --features private` and `--features private,windmill-common/smtp` both pass, so `SMTP_ENABLED` resolves through the `email_ee` glob re-export in the `private`-without-`smtp` arm (`worker_windows_core`) as well as the arm both published bundles use. The `private,smtp` value is `true` — the same build sends real mail and produces the new SMTP diagnostics at runtime.
- The `map_err` now matches only `Error::Anyhow` (what `send_email_plain_text` returns) and passes every other variant through with its own status code.
- The failure alert is keyed to the settings it describes, so editing any SMTP field retires it. Verified in the browser.

Also closed the reviewer's outstanding manual check: a backend built `--features quickjs` (no `private`) now answers 501 "This Windmill build was compiled without SMTP support" where it previously reported success — screenshot added to the PR body.

PR #10622, opened 2026-08-10
fix(cli): bundle Svelte raw apps by loading the app's ESM compiler

rubenfiszel:
Fixes GIT-958

## Summary

Since 1.775.2, `wmill sync push` fails to bundle any Svelte raw app under Node with `svelte.compile is not a function`, aborting the push on the first raw app it reaches.

1.775.2 started resolving the Svelte compiler out of the *app's* `node_modules` (so compiler and runtime agree on internals) via `createRequire(<app>/package.json).resolve("svelte/compiler")`. `require.resolve` answers with the **`require`** condition of Svelte's exports map:

```json
"./compiler": {
  "types": "./types/index.d.ts",
  "require": "./compiler/index.js",     // minified UMD/CJS bundle
  "default": "./src/compiler/index.js"  // ESM, named exports
}
```

Only the CJS loader can read that UMD file's exports — `cjs-module-lexer` can't see the assignments through the wrapper — so `import()`ing it under Node yields a namespace whose only keys are `default` and `module.exports`, and `svelte.compile` reads `undefined`.

The app-local resolution itself is right and is kept; the fix is to ask for the entry Node's *ESM* resolver would pick, since `require.resolve` takes no `conditions` option.

Bun masks this (it synthesizes named exports from `module.exports`), which is why the bun-run test suite never saw it while the published, Node-targeted CLI broke.

## Changes

- `cli/src/commands/app/bundle.ts`: read the app's `svelte/package.json` exports map and take the `./compiler` target the ESM resolver would pick (`node` / `import` / `default`, in declaration order), falling back to `require.resolve` when there is no usable exports map.
- Same file: unwrap a CJS `default` namespace in `loadSvelteCompiler`, so the `require.resolve` fallback still hands back a usable compiler.
- `cli/test/raw_app_svelte_compiler_resolution_unit.test.ts`: one regression case against a stub shaped like the real package (`require` → UMD, `default` → ESM), pinning which condition wins. It fails on the pre-fix code.

Out of scope: the reporter also noted the failure surfaces as an unhandled esbuild rejection with a raw Node stack trace rather than a handled push error. That error-handling papercut in the sync path is untouched here.

## Test plan

Verified with the npm-built CLI running under **Node** (`bun run src/main.ts` masks the bug), against a raw app whose only dependency is `svelte@5.56.8`:

- [x] `wmill sync push --yes` against a local backend: fails pre-fix with the exact reported error and Node stack trace; **succeeds** post-fix (5 files pushed, bundle 46.9kb).
- [x] `wmill app bundle`: same before/after result.
- [x] The deployed raw app renders in the Windmill UI and its `onclick` increments the counter — the compiler that produced the bundle agrees with the app's own runtime (screenshot below).
- [x] Compiler still resolves per app, not from the CLI: an app pinned to `svelte@5.45.2` loads `VERSION 5.45.2`, one pinned to `5.56.8` loads `5.56.8`.
- [x] `UNIT_ONLY=1 bun test test/*_unit*` — 1053 pass, 0 fail.
- [x] `bunx tsc --noEmit` — no new errors (18 pre-existing, unchanged with the diff reverted).

## Screenshots

Deployed Svelte raw app, pushed with the fixed CLI, after clicking the counter:

deployed-svelte-raw-app

## Note on `wmill app dev`

The issue reports `wmill app dev` as unaffected. It isn't: `dev.ts:581` calls the same `createFrameworkPlugins` → `loadSvelteCompiler`. On the pre-fix build under Node, `wmill app dev` fails with the identical `svelte.compile is not a function`; on this branch it builds and serves. There is no separate compiler-loading path to explain the difference.

### DIFF
--- a/cli/src/commands/app/bundle.ts
+++ b/cli/src/commands/app/bundle.ts
@@ -80,6 +80,51 @@ export function detectFrameworks(appDir: string): { svelte: boolean; vue: boolea
   }
 }
 
+/** What an `import()` matches — the point being that it never matches "require". */
+const ESM_CONDITIONS = ["node", "import", "default"];
+
+/**
+ * Walks one subpath of an exports map the way Node's ESM resolver would: first
+ * key in declaration order whose condition an `import()` matches wins.
+ */
+function esmConditionTarget(subpath: unknown): string | undefined {
+  if (typeof subpath === "string") return subpath;
+  if (!subpath || typeof subpath !== "object" || Array.isArray(subpath)) {
+    return undefined;
+  }
+  for (const [condition, target] of Object.entries(subpath)) {
+    if (!ESM_CONDITIONS.includes(condition)) continue;
+    const entry = esmConditionTarget(target);
+    if (entry) return entry;
+  }
+  return undefined;
+}
+
+/**
+ * `require.resolve` answers with the `require` condition, which Svelte maps at a
+ * minified UMD bundle. Only the CJS loader can read that file's exports, so
+ * `import()`ing it yields a namespace holding nothing but `default` and every
+ * named export reads undefined. The exports map is the only place to ask for the
+ * ESM entry instead — `require.resolve` takes no conditions.
+ */
+function resolveAppSvelteCompiler(appDir: string): string {
+  const requireFromApp = createRequire(
+    path.join(path.resolve(appDir), "package.json")
+  );
+  try {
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 1809326a74c9f24c7afb2f9aa9165e49de7860c6

PR #10624, opened 2026-08-10
feat(sdk): add cancelJob to the TypeScript client

rubenfiszel:
## Summary

Derived from #10601 by @tusharui, who is credited as the author of the `cancelJob` commit. Thank you for the contribution.

The Python client has had `cancel_job` for a long time; the TypeScript client had no way to cancel a job at all. This adds `cancelJob` and nothing else.

Scoped down from the original PR, which also re-enabled the S3/DuckDB/Polars/boto3 connection-settings helpers. Those are dropped here — see below.

## Changes

- `cancelJob(jobId, reason?)` in `typescript-client/client.ts`, hitting the same `jobs_u/queue/cancel/{id}` endpoint as Python's `cancel_job`, with a default reason when none is given.
- Exported from the npm package's named and default exports, and from the JSR package.
- Regenerated `system_prompts/auto-generated/`, which `check-system-prompts` requires for any change under `typescript-client/`.

## Test plan

- [x] `cd typescript-client && npm install && ./build.sh && npx tsc` — clean, 0 errors (`tsc` is what `prepublish` runs)
- [x] `cd typescript-client && bun test tests/` — 167 pass
- [x] `bash system_prompts/check-freshness.sh` — up to date
- [x] `cancelJob` present in all three export sites of the generated `src/index.ts`

Note that `sdk-tests.yml` runs on release tags and `workflow_dispatch` only, so the SDK suites do not execute on this PR.

Worth one manual run before merge: start a sleeping job, call `cancelJob(jobId)` from another script, and confirm the job reaches a cancelled state — with and without an explicit reason.

## Dropped from the original PR

The S3 connection-settings helpers (`duckdbConnectionSettings`, `polarsConnectionSettings`, `boto3ConnectionSettings`) and the Python `database_url_from_resource` are not included. Review turned up enough that they warrant their own change rather than riding along with a one-function addition:

- An empty `endPoint` is a valid resource state meaning plain AWS S3, and building `https://` from it hands boto3 a hostless endpoint. Both clients had this.
- An empty-string `s3_resource_path` reaches the backend as `Some("")` and 404s instead of falling back to the workspace default. Python omits the field; the TypeScript port did not.
- `S3Resource` in `openapi.yaml` is missing `port`, `token` and `expiration`, and the DuckDB response is missing `s3_bucket`, all of which the Rust structs return.
- `polars_connection_settings_v2` returns an untagged S3/Azure/GCS union whose arms share no field, so a single S3-shaped type is wrong for two of the three backends.

Separately, and unrelated to this PR: **`deno publish --dry-run` fails on `main`** with `TS1205`, because `build.jsr.sh` re-exports `Sql` as a value where it is `export type Sql = string`. JSR releases are broken today and that is worth its own fix.


rubenfiszel:
✅ Review round clean @ 1debbdeb7265eafbac6fcb85ede9de1e683e3f52

PR #10625, opened 2026-08-10
feat(triggers): nested filter groups and dotted paths

rubenfiszel:
## Summary

Trigger filters could only be a flat list combined by a single `filter_logic`, so a criterion like "A **and** B **and** (C **or** D)" was not expressible. A filter entry can now also be a **group** — `{"any_of": [...]}` or `{"all_of": [...]}` — nesting further entries, groups included. Kafka and WebSocket triggers share the filter engine, so both gain it.

The example from WIN-2345 now works exactly as it was written:

```yaml
filter_logic: and
filters:
  - path: account.id
    value: 1028
  - key: event
    value: message_created
  - key: message_type
    value: incoming
  - any_of:
      - path: in_reply_to_message.content_attributes.zeya__sent_by
        value: athosperformancecom/appointment_reminders/send_reminder_message
      - path: previous_public_message.content_attributes.zeya__sent_by
        value: athosperformancecom/appointment_reminders/send_reminder_message
```

A criterion addresses its field by `key` (a top-level field) or `path` (a dotted path into
nested objects), and a group nests criteria under `any_of` / `all_of` / `none_of`.

**Back-compatible.** `Filter` was already an untagged enum with `{key, value}` as its first variant, so every stored filter keeps parsing and meaning exactly what it did; groups are externally tagged (`any_of` / `all_of`), which cannot collide with a leaf. Top-level entries are still combined by the trigger's `filter_logic`. Nothing needs migrating.

**Faster than before, not slower.** Each leaf used to re-scan the whole message: N filters meant N streaming passes per message. Filters are now compiled once per connection — the set of top-level keys the whole tree references is collected up front — and each message is scanned **once**, borrowing the wanted keys as raw slices and skipping everything else as `IgnoredAny`. A field is only deserialized when the boolean evaluation actually reaches its leaf, so `and`/`or` short-circuiting still keeps a large object off the heap when a cheaper criterion already decided the verdict. So a 4-criterion filter goes from 4 scans per message to 1, nesting adds no per-message cost, and the message is never fully materialized.

Leaf semantics are unchanged: a key matches a top-level field of the message parsed as JSON, and the value must be equal to (or a superset of) it — nested fields are still expressed with an object value, as the editor's help text says.

### Judgment calls

- **An empty group is dropped, not evaluated.** It cannot be a constant: `true` makes an `or` list accept every message, `false` mutes an `and` list — and "Add group" + save produces one in two clicks. Dropping it leaves its siblings in force.
- **Filters that don't parse are rejected at save time** (400) rather than only logged at listen time. A group nests arbitrarily many criteria, so a single mistyped entry (`anyOf` instead of `any_of`) would silently widen the trigger by the whole subtree. Filters that *do* slip through are now logged instead of dropped in silence.
- **The editor caps nesting at 3 levels.** The backend has no such limit (serde_json's depth limit applies); deeper than that just stops being readable in a drawer.

### Judgment calls on the new fields

- **`path` is a separate field, not dots in `key`.** `key: "a.b"` already means the top-level field
  spelled that way, which flattened payloads really do have; teaching `key` to split would resettle
  what those triggers match. Both forms are pinned by a test.
- **`path` addresses objects only.** A path through an array does not match rather than guessing an
  element; array containment stays on the value side (`value: [{id: 1}]`). Worth noting this is the
  one deferral that is not purely additive — if paths ever traverse arrays, a filter that matches
  nothing today would start matching.
- **No escaping yet** for a segment containing a literal dot. Adding an escape or an array form later
  is additive, since a path without special characters behaves identically either way.
- **Negation is group-only.** The root's operator is the `filter_logic` column (`and|or`), which has
  no value for it, so `none_of` exists as a group and the root selector still offers two options.
- **A key the message does not carry satisfies `none_of`** — there is nothing there to match. The
  alternative would make every negative filter also silently require the field to be present.
- **Comparison operators** (`!=`, `in`, `exists`) are still deliberately out: that is the step where
  this becomes a query language and deserves its own design.

## Changes

**Backend**
- `windmill-trigger/src/filter.rs`: `FilterGroup::{AnyOf, AllOf, NoneOf}` and a `PathFilter` leaf as new `Filter` variants; `CompiledFilters` replaces `check_filters` — precomputes the referenced key set, parses each message in one pass (`KeysVisitor`), drops empty groups, and exposes `validate` for the handlers.
- Kafka (EE) and WebSocket listeners build a `CompiledFilters` per connection instead of a `Vec<Filter>`; both `validate_config`s now reject unparseable filters.
- `openapi.yaml`: new recursive `TriggerFilter` schema, `$ref`d from all six `filters` arrays.

**Frontend**
- `TriggerFilterList.svelte` (new, recursive) renders leaves and groups, each group with its own AND/OR/NONE selector; `TriggerFilters.svelte` becomes the section wrapper. `triggers/filters.ts` holds the node types and helpers.
- Each leaf gets a Key/Path toggle. Switching is lossless — a single-segment path behaves exactly like a key.
- "Add group" defaults to the opposite logic of its parent, which is the common case.

**Generated / tooling**
- `system_prompts/generate.py` maps `oneOf` to a Zod union and stays permissive at recursion cut-offs (it previously emitted a strip-mode `z.object({})`, which would have silently deleted nested payloads).
- `system_prompts/utils.py` resolves `$ref` inside array `items`, so a `$ref`d item type keeps its documented shape. This also recovers the shapes `initial_messages` and the MQTT `topics` had already lost.

## Screenshots

The WIN-2345 config from the issue, loaded verbatim — `path` criteria, `key` criteria and an `any_of` group:

the WIN-2345 filter config in the editor

A leaf's Key/Path toggle, an `any_of` group of paths, and a `none_of` group:

nested filter groups with key/path toggles

The toggle carries the distinction on hover, since the labels alone don't convey it:

the Path toggle's tooltip

## Test plan

Ran locally against a backend built with `--features kafka,websocket,enterprise,private`:

- [x] `cargo test -p windmill-trigger filter::` — 35 pass, incl. groups nested both ways, 3-level nesting, empty groups under AND and OR, legacy-and-group entries parsed side by side from raw config JSON, and first-occurrence-wins on a duplicated payload key (the invariant the single-pass rewrite could have broken).
- [x] `cargo check --features kafka,websocket,enterprise,private` clean.
- [x] **End-to-end through a real listener** (WebSocket trigger against an echo server, `filter_logic: and` + an `any_of` group): of three pushed messages, only the one satisfying `event == message_created` **and** one branch of the group started a job. The message matching neither branch, and the one failing the top-level leaf, did not.
- [x] **End-to-end for the empty-group case**: `filter_logic: or` with `[{any_of: []}, {key: event, value: wanted}]` — the `wanted` message ran, the `unwanted` one did not (before the fix, both would have).
- [x] **Short-circuiting is pinned by a test that fails without it**: a filter list naming a field whose value overflows `Value`'s f64 matches under `or` only if that field is never deserialized. Confirmed it fails against the eager version and passes against this one.
- [x] **`none_of` end-to-end through a real listener**: of four messages, the two matching a `none_of` branch were excluded, the one matching neither ran, and the one *missing the field entirely* ran. A control trigger without `none_of` confirmed the harness itself was live.
- [x] **`path` end-to-end**: the WIN-2345 config was created verbatim via the API, rendered correctly in the editor (screenshot above), and the Key/Path toggle round-tripped through a save.
- [x] **API round-trip**: a nested filter created via `POST /kafka_triggers/create` reads back identically; a `{"anyOf": ...}` typo is rejected with `400 filter #2 is neither a {key, value} criterion nor an any_of/all_of group`.
- [x] **UI round-trip**: opened a trigger whose stored filters contain `any_of`, confirmed the editor renders the group and its children, flipped it to AND, saved, and confirmed the stored config became `all_of` with children intact.

Fixes WIN-2345

EE companion: https://github.com/windmill-labs/windmill-ee-private/pull/722


rubenfiszel:
✅ Review round clean @ 754b391fb220a4f5394cf2d37b63d7bae7268573

rubenfiszel:
✅ Review round clean @ 231c461161f9d6e8ed87c979662f5aa5e44e297d — nit-only verdicts at e571c3bbcddad010a6bd65b90ba060442a8303dd (codex: good to merge, pi: good to merge, claude: nits); nits addressed in 231c461161f9d6e8ed87c979662f5aa5e44e297d (none_of added to the filter_logic description in all 6 copies + regenerated artifacts, and a test for the empty-path-segment rejection).

rubenfiszel:
✅ Review round clean @ 04148e2887411aa9b40b7a63c525b8ec3efc4eb8 — nit-only verdicts at b672c7a7216553e91156c2064d63b0b3036bf4da (codex: good to merge, claude: good to merge, pi: one nit); nit addressed in 04148e2887411aa9b40b7a63c525b8ec3efc4eb8 (a criterion naming both `key` and `path` is now rejected at save time instead of silently parsing as a `key` criterion, with a test).

rubenfiszel:
✅ Review round clean @ f8a3f5d9f0d9e12f4a05a6ca7fc50ef736cd97bc

PR #10629, opened 2026-08-11
fix(raw-apps): respect the instance .npmrc in the raw app editor

rubenfiszel:
Fixes WIN-2346

## Summary

The raw app editor installs npm packages in the browser, and that installer (in the pinned `ui_builder` artifact) hardcodes `unpkg.com` for version resolution and `registry.npmjs.org` for tarballs. An instance that configures a private registry through the `npmrc` instance setting is ignored entirely: the editor either pulls public packages when it should not, or — on a network where those hosts are unreachable — cannot install anything at all. The same applies to the deploy path, which bundles through a hidden `?mode=bundle` UI-builder iframe.

Confirmed on a dev instance before the change: with `npmrc` set, every install still went to `unpkg.com` / `registry.npmjs.org`.

This PR is the Windmill half of the fix. It adds the endpoints the in-browser installer needs and passes it the workspace so it can find them. The installer change is windmill-labs/windmill-code-ui-builder#27; **the `frontend/scripts/ui_builder_artifact.json` pin bump belongs on this branch once that PR is merged and its R2 artifact is published** — until then this is inert.

## Changes

**`backend/windmill-api-npm-proxy`**
- `GET /w/{w}/npm_proxy/config` → `{ registry_configured }`, so a client that would otherwise use the public CDNs knows whether to route through the proxy at all.
- `GET /w/{w}/npm_proxy/tarball/{package}/{version}` streams the package tarball from the configured registry, for installers that unpack it themselves.
- `resolve` now handles real semver ranges (`^1.2.3`, `19.x`, `1.2.3 - 2.3.4`, `>=1 <2`, `a || b`, bare partials), not just dist-tags and exact versions. Ranges are what `package.json` dependencies actually hold, so the endpoint was previously unusable for installs (and for ATA fallback). npm's grammar differs from the `semver` crate's in three ways the translation handles: space- vs comma-separated comparators, hyphen ranges, and bare partials (`1.2` is `1.2.x` to npm, `^1.2` to the crate).
- Packuments are cached in-process for 60s. Every endpoint needs the whole document and an install resolves then downloads each package, so a dependency would otherwise re-download a multi-megabyte packument several times per install.
- The four handlers' duplicated fetch-packument / download-tarball blocks are now two shared helpers.

**`frontend`**
- `RawAppEditor.svelte` and `rawAppBundlerBridge.ts` pass `workspace` on the UI-builder iframe URL.
- The bundler iframe drops `credentialless`. A credentialless frame gets an empty cookie jar, so its installer would be unauthenticated against `/api/w/*`; `/ui_builder/*` is itself served with `COEP=require-corp` (`static_assets.rs`, `needs_cross_origin_isolation`), so it embeds in isolated pages without it.

## Test plan

Verified on a dev instance with `npmrc` pointing at a token-gated mock registry that mirrors npmjs, and the ui-builder branch built locally into `frontend/static/ui_builder`:

- [x] Editor install of the React template: 13/13 package requests went to `/api/w/admins/npm_proxy/*`, 0 to npmjs.org/unpkg.com; the mock registry saw every packument and tarball request carrying the configured `_authToken`; the app builds and the preview renders.
- [x] Deploy: the bundle-mode iframe resolved through the proxy authenticated, `apps/create_raw` returned 201, and the deployed app renders.
- [x] Non-isolated parent (workspace home, where global AI chat bundles from): the non-`credentialless` bundle iframe loads, reaches `bundleRawAppReady`, and its `/npm_proxy/config` fetch is authenticated.
- [x] Fallback with `npmrc` cleared: one `/config` probe returns `registry_configured: false`, everything else goes to unpkg.com / registry.npmjs.org as before.
- [x] Range resolution matches `node-semver`'s `maxSatisfying` against the real `react` packument for `^19.0.0`, `19.x`, `18.0.0 - 19.0.0`, `19.0`, `19`, `>=18 <19`, `latest` (19.2.8 / 19.2.8 / 19.0.0 / 19.0.8 / 19.2.8 / 18.3.1 / 19.2.8 — identical on both sides).
- [x] Packument cache: 5 proxy requests for one package produce 1 packument fetch plus 1 tarball download at the registry.
- [x] `cargo check`, `cargo test -p windmill-api-npm-proxy`, `npm run check` (0 errors).

## Screenshots

Raw app editor with a private registry configured — packages resolved and installed through the proxy, build successful, preview rendering:

wm2346-editor-installs-via-proxy

The same app after deploying, bundled entirely from the private registry:

wm2346-deployed-app

---
## Summary by cubic
Route in-browser npm installs through the workspace npm proxy so the raw app editor and deploy path respect the instance `.npmrc`. Fixes WIN-2346 by avoiding `unpkg.com`/`registry.npmjs.org` when a private registry is configured.

- New Features
  - Added npm proxy endpoints: `/w/{w}/npm_proxy/config` and `/w/{w}/npm_proxy/tarball/{package}/{version}`; OpenAPI updated.
  - Version resolution matches npm semantics for tags, exact versions, and ranges (x-ranges, hyphen ranges, `||`, space-separated comparators) and keeps `v`-prefixed pins exact.
  - Packument caching (60s) is shared and byte-bounded (64MB), requests abbreviated packuments, and tarballs stream with immutable cache headers.
  - Bumped `ui_builder` artifact pin so the in-browser installer uses the proxy.

- Bug Fixes
  - Editor and bundle-mode installs now use the proxy when `.npmrc` is set; fall back to public CDNs when not.
  - UI builder iframes pass `workspace` and no longer use `credentialless`, keeping `/api/w/*` requests authenticated.

<sup>Written for commit 1d59318b97108569dd21578fb3dd4e10f484acb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10629?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -15116,6 +15116,7 @@ version = "1.783.0"
 dependencies = [
  "axum 0.8.9",
  "flate2",
+ "lazy_static",
  "reqwest 0.13.4",
  "semver 1.0.28",
  "serde",
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -15117,6 +15117,7 @@ dependencies = [
  "axum 0.8.9",
  "flate2",
  "reqwest 0.13.4",
+ "semver 1.0.28",
  "serde",
  "serde_json",
  "sqlx",
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -15117,6 +15117,7 @@ dependencies = [
  "axum 0.8.9",
  "flate2",
  "lazy_static",
+ "quick_cache",
  "reqwest 0.13.4",
  "semver 1.0.28",
  "serde",
--- a/backend/windmill-api-npm-proxy/Cargo.toml
+++ b/backend/windmill-api-npm-proxy/Cargo.toml
@@ -14,6 +14,7 @@ windmill-common = { workspace = true, default-features = false }
 axum.workspace = true
 flate2.workspace = true
 reqwest.workspace = true
+semver.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sqlx.workspace = true
--- a/backend/windmill-api-npm-proxy/Cargo.toml
+++ b/backend/windmill-api-npm-proxy/Cargo.toml
@@ -13,6 +13,7 @@ windmill-api-auth.workspace = true
 windmill-common = { workspace = true, default-features = false }
 axum.workspace = true
 flate2.workspace = true
+lazy_static.workspace = true
 reqwest.workspace = true
 semver.workspace = true
 serde.workspace = true
--- a/backend/windmill-api-npm-proxy/Cargo.toml
+++ b/backend/windmill-api-npm-proxy/Cargo.toml
@@ -14,6 +14,7 @@ windmill-common = { workspace = true, default-features = false }
 axum.workspace = true
 flate2.workspace = true
 lazy_static.workspace = true
+quick_cache.workspace = true
 reqwest.workspace = true
 semver.workspace = true
 serde.workspace = true
--- a/backend/windmill-api-npm-proxy/src/lib.rs
+++ b/backend/windmill-api-npm-proxy/src/lib.rs
@@ -13,8 +13,10 @@ use axum::{
 use serde::{Deserialize, Serialize};
 use sqlx::types::JsonValue;
 use std::collections::HashMap;
+use std::time::{Duration, Instant};
 use tower_http::cors::{Any, CorsLayer};
 use windmill_common::{
+    cache::Cache,
     error::{Error, JsonResult, Result},
     global_settings::{
         load_value_from_global_settings, NPMRC_SETTING, NPM_CONFIG_REGISTRY_SETTING,
@@ -165,6 +167,15 @@ fn build_registry_request(
     Ok(req)
 }
 
+/// Every endpoint here needs the whole registry document, and an install resolves then
+/// downloads each package, so without this each dependency of an app being edited would
+/// re-download a packument (megabytes, for popular packages) several times over.
+const PACKAGE_JSON_CACHE_TTL: Duration = Duration::from_secs(60);
+
+lazy_static::lazy_static! {
+    static ref PACKAGE_JSON_CACHE: Cache<(String, String), (JsonValue, Instant)> = Cache::new(500);
+}
+
 /// Fetch a package's registry document, together with the registry it came from so
 /// callers can validate tarball URLs against it.
 async fn fetch_package_json(
@@ -174,6 +185,14 @@ async fn fetch_package_json(
     let (registry_url, auth_token) = get_npm_registry(db)
         .await?
         .ok_or_else(|| Error::BadRequest("No private npm registry configured".to_string()))?;
+
+    let cache_key = (registry_url.clone(), package.to_string());
+    if let Some((package_json, fetched_at)) = PACKAGE_JSON_CACHE.get(&cache_key) {
+        if fetched_at.elapsed() < PACKAGE_JSON_CACHE_TTL {
+            return Ok((package_json, registry_url, auth_token));
+        }
+    }
+
     let package_url = format_registry_url(&registry_url, package, None, None);
 
     tracing::info!("Fetching package metadata from: {}", package_url);
@@ -194,6 +213,8 @@ async fn fetch_package_json(
         .json()
         .await
         .map_err(|e| Error::InternalErr(format!("Failed to parse package metadata: {}", e)))?;
+
+    PACKAGE_JSON_CACHE.insert(cache_key, (package_json.clone(), Instant::now()));
 
     Ok((package_json, registry_url, auth_token))
 }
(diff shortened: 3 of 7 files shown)


rubenfiszel:
Round 1 addressed in 083cf3776e.

**P1 packument cache (codex #2, claude)** — both parts fixed. The cache is now bounded by **bytes** via a `quick_cache` weighter (64 MB of registry-response length) rather than by document count, and stores `Arc<JsonValue>` so a hit is a pointer clone rather than a deep copy of the tree. Packument requests also now send `Accept: application/vnd.npm.install-v1+json`, npm's own abbreviated form, which is one to two orders of magnitude smaller and carries everything the handlers read; registries that do not implement it answer with the full document, which the byte bound then accounts for correctly. The one field the abbreviated document drops is `versions[v].main`, so `/filetree` reads it from the packaged `package.json` inside the tarball it already downloads — verified against npm's own metadata for a package whose main is not the default (`lodash@4.17.21` → `lodash.js`, `axios@1.7.7` → `index.js`).

**P2 unsupported range forms (claude)** — `>= 1.0.0` (operator detached from its version) and a `v` prefix both resolve now, with unit-test cases; verified end-to-end against the real `react` packument (`>= 18 < 19` → 18.3.1, `^v19.0.0` → 19.2.8, matching `node-semver`).

**P2 `/tarball` doc vs code (claude)** — now genuinely streams (`Body::from_stream`, chunked response confirmed) and sets `Cache-Control: private, max-age=31536000, immutable`.

**P2 `$workspaceStore` → `opWorkspace` (claude)** and **empty `workspace=` (claude)** — both done; the bundler bridge omits the param entirely when it has no workspace, so an omitting caller cannot silently fall back to the public registry.

**P1 artifact pin (codex #1)** — correct, and stated in the PR description: this branch is inert until windmill-labs/windmill-code-ui-builder#27 merges and publishes its R2 artifact, at which point the `frontend/scripts/ui_builder_artifact.json` bump lands here and the private-registry scenarios get re-run against the shipped pin. Splitting it that way is forced by the publish workflow (push-to-main on that repo), not a choice.

**P2 stale `openapi-deref.*` (codex #3)** — not this PR's to fix. Those bundles are regenerated occasionally rather than per-PR, and `build_openapi.sh` pins `@redocly/openapi-cli@latest`, so regenerating here produced ~16k changed lines: the tool version renumbers every YAML anchor globally and folds in every other merged PR's routes. A splice limited to these two operations is not possible. `openapi.yaml` (the source, and the input to `npm run generate-backend-client`) does carry both new operations. The regen belongs in its own PR.

rubenfiszel:
✅ Review round clean @ 2fa14a2a73f2b46e6a2507c5695ab6e6916226f7 — nit-only verdicts at 083cf3776ef36b1add7c215ae5fd03a9e3e9220c (codex + claude "mergeable, address nits", pi "good to merge"); nits addressed in 2fa14a2a73f2b46e6a2507c5695ab6e6916226f7.

- **`v`-prefixed exact version widened into a caret** (codex, claude) — the `v` strip moved out of `split_comparators` and into `resolve_version_spec`, ahead of both exact checks, so `v19.0.0` pins and `v19.0.99` returns `null` rather than silently resolving to the newest 19.x. Verified against the real `react` packument: `v19.0.0` → 19.0.0, `v19.0.99` → null, `^v19.0.0` → 19.2.8. Unit cases added for both.
- **`/filetree` gunzipped the tarball twice** (claude) — one pass now collects the file list and the manifest together. Re-checked that the entry point is still right for a package whose `main` is not the default: `lodash@4.17.21` → `lodash.js` (1054 files), `react@19.0.0` → `index.js` (27 files).
- **`RawAppEditor` still emitted `workspace=` empty** (claude, smaller note) — it now builds the iframe URL with `URLSearchParams` and omits the param when there is no workspace, matching the bundler bridge.

The remaining note on the weighter (encoded length as a proxy for the parsed tree, so the 64 MiB knob is really a few hundred MB of ceiling) is accurate and called out in the comment on the constant; with abbreviated packuments the absolute numbers are small, so I left the knob as is.

Re-verified after these changes: editor install 13/13 through the proxy with 0 public requests and the preview rendering, `cargo check` and the unit test clean, `npm run check` 0 errors.

The two scenarios @claude asked to fold into the post-pin pass — a non-`index.js` entry point opening correctly in the editor, and `/tarball` end-to-end now that it is chunked with no `Content-Length` — are noted for that run.

rubenfiszel:
The UI-builder PR is merged and published, so the pin is bumped in 1d59318b97 (`ui_builder_artifact.json` → `1522e43`, sha256 verified by the postinstall). **This PR is no longer inert** — the artifact it now pins is the one carrying the installer change.

Re-ran the private-registry scenarios against the shipped pin rather than a local build, including the two @claude asked to fold into this pass:

- **Editor install** — 13/13 package requests through `/api/w/admins/npm_proxy/*`, 0 to npmjs.org or unpkg.com, app builds, preview renders.
- **`/tarball` end-to-end, chunked** — 7 of those 13 are tarball downloads, served with `transfer-encoding: chunked` and no `Content-Length`. The installer's `fetch(...).arrayBuffer()` handles them fine; the extracted file counts match (`react@19.0.0` 27 files, `lodash@4.17.x` 1054).
- **Non-`index.js` entry point** — added `lodash: "^4.17.0"` to the app and imported `kebabCase`. Resolved through `/npm_proxy/resolve/lodash?tag=^4.17.0` → 4.18.1, downloaded through `/npm_proxy/tarball/lodash/4.18.1`, and the preview renders `hello-from-lodash`, i.e. the non-default `main` (`lodash.js`) resolved and the export executed. `/filetree/lodash/4.17.21` independently returns `default: lodash.js`, matching npm's own metadata.
- **Deploy** — `POST /apps/update_raw/…` 200 through the bundle-mode iframe, still authenticated, still no public requests.
- **Fallback with `npmrc` cleared** — exactly one `/npm_proxy/config` probe, the other 12 requests go to unpkg/npmjs as before, app builds.

Two follow-ups opened out of this work, both separate from the raw app editor:

- #10630 — the ATA backend-proxy fallback never fires when the CDN request *rejects* (blocked egress, DNS failure) rather than returning a non-ok status, because nothing catches it. Also fixes an over-limit path that returns a never-settling promise.
- A second PR (stacked on this one, since it needs `/npm_proxy/config`) to make the script editor's ATA prefer the proxy when a registry is configured, after adding a tarball cache — today each `/file` call re-downloads and re-gunzips the whole tarball, so flipping the order without that would be a large regression.

rubenfiszel:
✅ Review round clean @ 1d59318b97108569dd21578fb3dd4e10f484acb9 — codex, claude and pi all "Good to merge" on the head that includes the artifact pin bump.

PR #10630, opened 2026-08-11
fix(ata): fall back to the npm proxy when the CDN request fails outright

rubenfiszel:
## Summary

Type acquisition fetches from jsdelivr and falls back to `/api/w/{w}/npm_proxy/*`, which is what makes it work on an instance with a private registry and no route to the public CDNs. The fallback is gated on the helper *resolving* to an `Error`, but nothing catches a rejected `fetch` — so it only fires when jsdelivr **answers** with a non-ok status. On exactly the network where the fallback exists to help (blocked egress, DNS failure, refused connection), `fetch` rejects, the rejection propagates out through `ata/index.ts`, which has no try/catch around these calls, and the module gets no types at all. The proxy is never tried.

Three instances of the same class, all in `frontend/src/lib/ata/apis.ts`:

- `api()` — no `.catch()`, so a rejected request skips the fallback in `getNPMVersionsForModule`, `getNPMVersionForModuleReference` and `getFiletreeForModuleWithVersion`.
- `getDTSFileForModuleWithVersion()` — calls `fetch` directly and tests `res.ok`; a rejection propagates the same way.
- `backendProxyApi()` — has a `try`/`catch`, but `return limit(...)` inside a `try` in an `async` function adopts the promise *after* the block exits, so the catch never sees a rejection. Needs `return await`.

Also in `api()`: the over-limit guard returns `new Promise(() => new Error(...))`, an executor that never calls `resolve` or `reject`. That promise never settles, so once `resLimit` is exceeded every subsequent call hangs forever instead of erroring. `backendProxyApi` already returns the `Error` directly; this makes `api` match.

Found while working on #10629 (raw app editor honouring the instance `.npmrc`); separate bug, separate PR.

## Changes

- `api()`: `.catch()` returning an `Error` so callers reach their proxy fallback; over-limit returns a settled rejection-free `Error`.
- `getDTSFileForModuleWithVersion()`: `.catch(() => undefined)` on the jsdelivr fetch so a rejection falls through to the proxy branch.
- `backendProxyApi()`: `return await` so its existing `catch` actually covers the request.
- `apis.test.ts`: two cases pinning the fallback under a *rejecting* jsdelivr, for the metadata path and the `.d.ts` path.

## Test plan

- [x] `npx vitest run src/lib/ata/apis.test.ts` — 2 passed. Both fail with `TypeError: Failed to fetch` when `apis.ts` is reverted, confirming they pin the fix rather than passing vacuously.

No UI surface changes, so no screenshots: this is error-path plumbing behind Monaco's type acquisition.

---
## Summary by cubic
Ensures ATA falls back to the backend npm proxy when CDN requests reject or bodies fail to read, restoring type acquisition on networks without public CDN access. Also surfaces proxy errors, logs failed proxy d.ts fetches, and fixes a hang when the download limit is exceeded.

- **Bug Fixes**
  - api(): catch fetch rejections and return an Error; over-limit returns a resolved Error (no hang).
  - getDTSFileForModuleWithVersion(): use a `text()` helper to handle connect/body-read failures; then try the proxy with credentials; warn when the proxy d.ts fetch fails; return an Error if the proxy fails.
  - backendProxyApi(): use return await so try/catch sees rejections; log and surface proxy errors with cause.
  - text(): new helper that returns string or Error for both fetch and body-read failures.
  - Tests: added cases verifying proxy fallback for metadata and .d.ts when jsdelivr rejects.

- **Refactors**
  - Clarified comments on the proxy catch constraint and why `return await` is required.

<sup>Written for commit b4091e8f71781e4b5df1df1731ae4e5ac2c4423c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10630?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/ata/apis.test.ts
+++ b/frontend/src/lib/ata/apis.test.ts
@@ -1,0 +1,39 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { workspaceStore } from '$lib/stores'
+import { getDTSFileForModuleWithVersion, getNPMVersionsForModule } from './apis'
+
+/** jsdelivr unreachable (blocked egress, DNS failure), backend proxy answering. */
+function mockUnreachableJsdelivr(proxyBody: string) {
+	return vi.fn(async (input: RequestInfo | URL) => {
+		const url = String(input)
+		if (url.includes('jsdelivr')) throw new TypeError('Failed to fetch')
+		if (url.includes('/npm_proxy/')) return new Response(proxyBody, { status: 200 })
+		throw new Error(`unexpected request to ${url}`)
+	})
+}
+
+describe('ATA backend proxy fallback', () => {
+	afterEach(() => vi.unstubAllGlobals())
+
+	it('falls back to the proxy when the jsdelivr request rejects', async () => {
+		workspaceStore.set('test-workspace')
+		vi.stubGlobal(
+			'fetch',
+			mockUnreachableJsdelivr('{"tags":{"latest":"1.0.0"},"versions":["1.0.0"]}')
+		)
+
+		const versions = await getNPMVersionsForModule('lodash', { usage: 0 })
+
+		expect(versions).not.toBeInstanceOf(Error)
+		expect((versions as { versions: string[] }).versions).toEqual(['1.0.0'])
+	})
+
+	it('falls back to the proxy for a d.ts when the jsdelivr request rejects', async () => {
+		workspaceStore.set('test-workspace')
+		vi.stubGlobal('fetch', mockUnreachableJsdelivr('declare const x: number'))
+
+		const dts = await getDTSFileForModuleWithVersion('lodash', '1.0.0', '/index.d.ts')
+
+		expect(dts).toBe('declare const x: number')
+	})
+})
--- a/frontend/src/lib/ata/apis.ts
+++ b/frontend/src/lib/ata/apis.ts
@@ -41,8 +41,8 @@ const backendProxyApi = async <T>(endpoint: string, resLimit: ResLimit): Promise
 			})
 		)
 	} catch (e) {
-		// Reachable for a failed request only now that the `await` above is there. On an
-		// instance with no route to the CDN this is the only error an operator gets.
+		// Keep the cause: where the proxy is the only reachable source, this is the sole
+		// report of a failed acquisition, since callers only test the result for `Error`.
 		console.warn(`Backend proxy request to ${endpoint} failed`, e)
 		return new Error(`Backend proxy not available: ${e}`)
 	}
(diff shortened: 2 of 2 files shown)


rubenfiszel:
Both nits addressed in 72207cbe49.

- **`backendProxyApi`'s catch discarded `e`** — right, and it matters most on the instance this PR targets. It now logs (`console.warn`) and threads the cause into the returned message, matching what `api()`'s new catch does.
- **Same `return`-in-`try` trap at the old line 147, and an unguarded `res.text()` at 138** — both gone. `getDTSFileForModuleWithVersion` now routes each attempt through a small `text()` helper that catches around the whole request *and* the body read, so a connection dying mid-body is a returned `Error` rather than a rejection. That also removes the `res?.ok`/`undefined` dance the first commit introduced.

`npx vitest run src/lib/ata/apis.test.ts` still 2 passed, `npm run check` 0 errors.

rubenfiszel:
Fixed in b2b70f5a04. The comment described how the code got here rather than what constrains it:

```
- // Reachable for a failed request only now that the `await` above is there. On an
- // instance with no route to the CDN this is the only error an operator gets.
+ // Keep the cause: where the proxy is the only reachable source, this is the sole
+ // report of a failed acquisition, since callers only test the result for `Error`.
```

It now says why the cause has to be preserved — callers discard everything but the `Error`-ness of the result — which is what someone about to "simplify" the log or the message needs to know, and reads the same to someone who never saw the earlier version.

I left the comment on `return await` above it as is: "an async function adopts a returned promise after leaving the try block, so a rejection would escape the catch below" is the enduring reason the `await` cannot be dropped, not a note about the change.

rubenfiszel:
Second nit fixed in b4091e8f71 — and the diagnosis is exact. `text()` converts a rejection to a value, so `return await text(...)` handed the `Error` straight to the caller and the `catch` below became reachable only for `getBackendProxyUrl()` throwing. The cause survived inside the returned `Error`, but `ata/index.ts:218` and `:237` both log a fixed message and drop the value, so nothing printed it. This commit added a `console.warn` to `backendProxyApi` while silently removing the equivalent from the `.d.ts` path — exactly the asymmetry called out.

Now logs in the `instanceof Error` branch before returning, with a comment recording *why* the log has to live here rather than being left to the caller.

Both round nits are addressed: the drafting-history comment in b2b70f5a04, this one in b4091e8f71. `npx vitest run src/lib/ata/apis.test.ts` 2 passed, `check:fast` clean.

rubenfiszel:
✅ Review round clean @ b4091e8f71781e4b5df1df1731ae4e5ac2c4423c — codex and claude both "Good to merge" on this head; both earlier nits (drafting-history comment, lost `.d.ts` proxy failure log) addressed in b2b70f5a04 and b4091e8f71.

PR #10631, opened 2026-08-11
fix: confine job tokens to workspace-scoped API routes

hugocasa:
## Summary

`$WM_TOKEN` — the token a job is issued — is minted for one job in one workspace and carries that job's user privileges. It was accepted on API routes whose path names no workspace, which are instance-wide. This scopes it to the workspace it was minted for, which is the rule the token lookup already applies to workspace-bound API tokens on those same routes.

## Changes

- `check_job_token_for_global_route` (`windmill-api-auth/src/scopes.rs`): a JWT carrying a `job_id` claim is refused on a route with no workspace in its path, with a `403` naming the route and the alternative.
- Allowlist for the workspace-less routes that cross no workspace boundary. A route qualifies only when it answers from the caller's own account, from the request body, or with content identical for every workspace; never naming another workspace, never disclosing instance configuration.
  - Reads: `users/whoami` (the Python client's `whoami()` and *every* `wmill` CLI invocation), `users/email`, `users/usage`, `users/tutorial_progress`, `workspaces/allowed_domain_auto_invite`, the docs endpoints and the Hub proxies.
  - Rejected candidates, recorded in the code so they are not re-litigated: `users/list_invites` (its response carries a workspace id per invite), `users/tokens/list` (credential metadata of the borrowed identity), `users/exists/{email}` (an oracle over arbitrary addresses), `workspaces/list` / `workspaces/users`.
- Four POSTs stay reachable — `POST` for the sake of a request body, not because they commit anything of consequence: `settings/test_object_storage_config` (the resource editor's "Test connection" runs as a preview job that POSTs its config; the probe acts only on the store the body describes), `workspaces/exists` (`wmill workspace add` checks it before accepting credentials — the git-sync hub scripts run exactly this; note `workspace` carries no row-level security, so the boolean is instance-wide and discloses only whether an id is taken), `schedules/preview` (takes no `ApiAuthed` and opens no transaction — it returns nothing the caller did not send) and `users/tutorial_progress` (a UI bitfield keyed on the caller's own email).
- Enforced in `resolve_opt_job_authed`, so it covers every handler extracting `ApiAuthed`. MCP tokens are untouched: workspace-bound too, but they carry no `job_id` and deliberately publish instance-wide tools.
- `tests/wm_token_confinement.rs` covers both caps a `$WM_TOKEN` is held to (see below).

**Upgrade note:** a script that calls an instance-wide endpoint with `$WM_TOKEN` (instance settings, global user management, worker/config listing) must switch to an API token created for the user. Everything under `/api/w/<workspace>/` is unaffected.

The `wmill` CLI still works from a job: `requireLogin` calls `users/whoami` on every invocation and `workspace add` checks `workspaces/exists`, both reachable. `wmill workspace add --create` is refused (it reads a global setting, then creates a workspace) as is `wmill instance *`; both need an API token created for the user. Client-library calls are workspaced apart from the Python client's `whoami()`, which is reachable.

## Test plan

- [x] `cargo test --test wm_token_confinement --features enterprise,private,parquet,mcp` — 11 passed, covering both directions of the allowlist (the permitted routes answer, the rejected caller-scoped reads stay refused)
- [x] `cargo test --test mcp_token_exfil --test jobs_read_auth --test preserve_on_behalf_of --test app_preview_auth --test inline_preview_auth` — 15 passed, including #10124's own suite unchanged
- [x] `cargo test --test worker test_bash_wmill_variable_get` — the `wmill` CLI running under `$WM_TOKEN` (it calls `users/whoami` on every invocation)
- [x] Ran a bash job with `$WM_TOKEN` against a live instance: workspace-less routes `403`; `users/whoami`, the Hub proxies and its own workspace `200`; another workspace still `401`
- [x] Workers page loads in the UI (browser sessions are unaffected)

## Relationship to #10124

#10124 caps a job token's *privilege level*; this caps its *workspace scope*. Now that
it has merged, the two compose with **no production-code change on either side** — the
middleware check and the privilege gates are independent layers.

**Both cover** (this PR is redundant there): `settings/*`, `groups/*`, `service_logs/*`,
`configs/list_worker_groups` — #10124's job-aware `require_super_admin` /
`require_devops_role` / `require_instance_admin` already reject a job token on those.

**Still open on `main`, closed here:**
- `users/tokens/create` for a **non-elevated** identity. #10124's `forbid_elevated_job_token`
  gates on `is_devops_email`, so an unprivileged on-behalf identity still mints a token with
  `workspace_id NULL` — a durable credential valid in every workspace that identity belongs
  to. #10124 treats this as intentional ("nothing to launder"), which holds on the privilege
  axis but not on the scope axis.
- `workers/list` — #10124 only flips `has_devops_role` to false, which redacts `custom_tags`
  and the `last_job_*` columns; the worker rows still return.
- `workspaces/users`, `users/tokens/list`, and any workspace-less route added later with no
  instance-authority gate: #10124 enumerates gates (default-open), this is default-deny
  across the workspace-less surface.

**Only #10124 covers:** workspaced and stored-identity elevation paths, which a check keyed
on "the request path names no workspace" cannot see.

### Reconciling the test suites

The middleware answers a workspace-less route before the gate behind it runs, so 12
assertions #10124 added on such routes now observe `403` instead of `401`. Rejection is
what they assert, so the status was updated — but flipping alone would have left them
passing for a reason unrelated to the gate they were written for, so two were repaired:

- `list_worker_groups` asserted only that the body omits the static env value, which an
  error body satisfies trivially — it had gone vacuous *silently*, still passing. It now
  asserts the status, keeping the secret check as a second assertion.
- `require_super_admin` lost its only route that reaches the gate. `GET
  /api/w/{workspace}/users/list_addable` is gated solely by that call and names a
  workspace, so it passes confinement and pins the gate at `401`.

`403` is kept rather than reusing #10124's `401` so the layer that refused stays visible
(matching would have made those assertions pass silently), and so the `wmill` CLI's
`401` re-login fallback does not fire on a permission refusal.

`tests/wm_token_superadmin_guard.rs` → `tests/wm_token_confinement.rs`: the file now
covers both caps, so it is no longer named for either one alone.

### Resolved by the merge

Round 1's Codex P1 — the object-storage exception admitting a job token to an SSRF surface
— is now closed upstream on the half that mattered: #10124 moved `test_s3_bucket` to
`is_super_admin_authed`, so on Cloud a job token no longer skips URL validation. The
self-hosted path stays unrestricted by design (private-network object stores).

### Hub script impact

All 16 hub scripts the product references (`hubPaths.json`, `GIT_SYNC_PULL_SCRIPT_PATH`,
`LATEST_GIT_SYNC_SCRIPT_PATH`) were checked: none calls a workspace-less route directly.
The two that drive the CLI from a job — the git-sync init/push scripts — run
`wmill workspace add <id> <id> $BASE_URL --token $WM_TOKEN`, which needs `users/whoami`
and `workspaces/exists`; both are allowlisted, and `--create` is used by no hub script.

The hub also carries ~310 generated one-per-endpoint API wrappers
(`windmill/scripts/action/2xxx_*.fetch.ts`, `bunnative`, authenticating with `$WM_TOKEN`).
Nothing in the product references them and the batch is unmaintained. 29 of them already
fail on `main` against #10124's privilege gates; a further 28 are refused here, and those
are the exposure this PR is about — `users/tokens/create`, `workers/list`,
`workspaces/users`, `users/all_runnables`, `groups/list`.


hugocasa:
Round 1 responses.

**Codex P1 — object-storage exception bypasses job-token confinement.** Not applying, for three reasons; happy to be overruled.

1. It is not a regression from this PR. A job token reached `test_object_storage_config` unconditionally before it; the exception preserves that, it does not open it.
2. The half that matters — a superadmin-identity job token skipping `validate_object_storage_test` on Cloud — is already fixed in #10124, in this exact handler: `is_super_admin_email(&db, &authed.email)` becomes `is_super_admin_authed(&db, &authed)`, which rejects `job_id.is_some()`, so `restrict` becomes true for a job token there. Duplicating that here would collide with it.
3. The other half — self-hosted skipping validation for every caller — is deliberate, and the handler says why: the object store usually sits on the local/private network. Forcing validation for job tokens would break "Test connection" against private endpoints, which is the case the exemption exists for. Reversing that is a product decision, not something this scoping PR should make.

**Codex P2 — test accepts broken responses.** Fixed in cd537a8: asserts `422` exactly, so a route rename or mounting failure fails the test.

**Claude P1 — `wmill workspace add` 403s on `workspaces/exists`.** Confirmed and fixed in cd537a8: allowlisted as a second write exception, with a case pinning it. The handler returns a bare boolean under the caller's own RLS.

Note `wmill workspace add --create` is still refused, at `settings/global/automate_username_creation` and `workspaces/create`. Creating a workspace is the kind of instance-level action this PR intends to stop, and no in-product flow appears to do it from a job — flagging rather than allowlisting speculatively.

**Claude P2 — "persists nothing" overstates it.** Fixed in cd537a8: the comment now says the probe acts only on the store the request body describes and touches no Windmill state.

PR #10632, opened 2026-08-11
feat(ata): prefer the npm proxy when the instance configures a registry

rubenfiszel:
## Summary

Type acquisition in the script editor asks jsdelivr first and only falls back to `/npm_proxy/*` when jsdelivr fails. On an instance that configures its own npm registry that is the wrong way round:

- jsdelivr does not carry internal packages, so the fallback is the path that actually works, reached only after a failure.
- Every lookup discloses internal package names to a third-party CDN, which is close to the opposite of why someone points npm at a private registry.
- Worse than either: if a public package shares a name with an internal one, jsdelivr answers `200` and the editor silently shows the **public** package's types. Nothing errors, so the fallback never runs.

This flips the order when `/npm_proxy/config` reports a registry is configured, keeping the other source as fallback in both directions so neither a CDN outage nor a misconfigured proxy leaves the editor with no types.

**The ordering flip is not safe on its own**, which is why the backend half is in the same PR: `/file` is called once per `.d.ts`, and each call re-downloaded and re-inflated the entire tarball. `bun-types` alone is ~30 files. As a fallback that cost was rare; as the primary path it would have been a large regression.

## Changes

**Backend** (`windmill-api-npm-proxy`)
- Cache inflated tarballs, keyed by registry + package + version, byte-bounded (128 MB) with the same weighter as the packument cache, 5 minute TTL. `/filetree` and every `/file` for one package version now share one download and one inflate.
- `extract_tarball` replaces the two single-purpose extractors; `/file` reads from the extracted map instead of walking the archive.

**Frontend** (`ata/apis.ts`)
- `isRegistryConfigured()` asks `/npm_proxy/config` once per session. Not memoized before a workspace exists, which would pin the answer to `false`.
- `fromPreferredSource()` orders the two sources and keeps the other as fallback, replacing the ordering logic duplicated across the four fetchers. It treats a rejection or a synchronous throw as a failed attempt rather than letting it escape.

## Stacking

- Based on #10629, which adds `/npm_proxy/config`. Retargets to `main` when that merges.
- Overlaps #10630 in `apis.ts`: that PR fixes the same rejection-escapes-the-fallback class in the original structure. Whichever lands first, the other rebases onto it; this PR's `fromPreferredSource` keeps the behaviour either way.

## Test plan

Verified against a token-gated mock registry mirroring npmjs, with `npmrc` pointed at it:

- [x] **Ordering, real script editor** — opened a bun script importing `lodash`. Every ATA request went to `/api/w/admins/npm_proxy/*`: `resolve/lodash`, `resolve/bun-types`, `resolve/@types/lodash`, the three `filetree` calls, and ~32 `file` calls. Zero requests to jsdelivr. `ATA done`, types added to the runtime.
- [x] **Tarball cache** — `bun-types` needed 34 proxy requests and produced exactly 2 registry requests: one packument, one tarball. Same with a cold `debug@4.3.4`: 1 filetree + 10 `/file` reads → 1 packument + 1 tarball.
- [x] **`/filetree` still correct after the refactor** — `lodash@4.17.21` → `default: lodash.js`, 1054 files.
- [x] `npx vitest run src/lib/ata/apis.test.ts` — 3 passed (proxy-first when configured, CDN-first when not, fallback when the preferred source rejects).
- [x] `cargo check`, `cargo test -p windmill-api-npm-proxy`, `npm run check` (0 errors).

No UI surface changes, so no screenshots: this changes which host Monaco's type acquisition talks to.

---
## Summary by cubic
Prefer the backend npm proxy for type acquisition when a registry is configured, with the CDN as fallback. Cache the gzipped tarball and only retain manifests and `.d.ts` files to keep large packages fast and cacheable.

- **New Features**
  - ATA checks `/npm_proxy/config` once per session and prefers the proxy when set; CDN stays as fallback via `fromPreferredSource()`. The config probe is retried after failures and skipped until a workspace exists. Updated `getNPMVersionsForModule`, `getNPMVersionForModuleReference`, `getFiletreeForModuleWithVersion`, and `getDTSFileForModuleWithVersion`. Tests cover ordering, retries, and both fallbacks.

- **Bug Fixes**
  - Backend: move archive inflate/read to blocking threads via `tokio::spawn_blocking`. Cache the compressed archive plus only small files (`package.json` and `.d.ts`), capped to 4 MB per file and 128 MB total; single‑shard, byte‑weighted, 5‑minute TTL, 256 MB budget. Charge per‑file overhead and path bytes so archives with many empty or long‑named files aren’t “free.” Bound inflation to 1 GB and entries to 100k; stream large files on demand; always read back the manifest if not retained; strip the archive’s root generically. Stop retaining past the budget instead of refusing the package, serving the rest via on‑demand reads. Frontend: a failed `/npm_proxy/config` probe no longer pins the session to CDN‑first, and thrown fetch errors no longer skip the fallback; over‑limit resolves to `Error` instead of hanging.

<sup>Written for commit 42619397b6b7851b2599d20b60402b8a2de24319. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -15218,6 +15218,7 @@ dependencies = [
  "serde_json",
  "sqlx",
  "tar",
+ "tokio",
  "tower-http",
  "tracing",
  "url",
--- a/backend/windmill-api-npm-proxy/Cargo.toml
+++ b/backend/windmill-api-npm-proxy/Cargo.toml
@@ -21,6 +21,7 @@ serde.workspace = true
 serde_json.workspace = true
 sqlx.workspace = true
 tar.workspace = true
+tokio.workspace = true
 tower-http.workspace = true
 tracing.workspace = true
 url.workspace = true
--- a/backend/windmill-api-npm-proxy/src/lib.rs
+++ b/backend/windmill-api-npm-proxy/src/lib.rs
@@ -245,11 +245,37 @@ impl Weighter<(String, String, String), CachedTarball> for CacheWeighter {
     }
 }
 
+/// Single-sharded on purpose. quick_cache divides the weight budget across shards and
+/// declines an item heavier than one shard's share of it, so a sharded cache silently
+/// refuses exactly the packages that cost the most to fetch and inflate again — the very
+/// regression these caches exist to prevent. They are consulted a handful of times per
+/// editor session and hold only map operations under the lock, so the concurrency a
+/// single shard gives up is not worth that.
+fn byte_bounded_cache<Key, Val>(items: usize, bytes: u64) -> Cache<Key, Val, CacheWeighter>
+where
+    Key: Eq + std::hash::Hash,
+    Val: Clone,
+    CacheWeighter: Weighter<Key, Val>,
+{
+    let options = quick_cache::OptionsBuilder::new()
+        .shards(1)
+        .estimated_items_capacity(items)
+        .weight_capacity(bytes)
+        .build()
+        .expect("every cache option is set");
+    Cache::with_options(
+        options,
+        CacheWeighter,
+        Default::default(),
+        Default::default(),
+    )
+}
+
 lazy_static::lazy_static! {
     static ref PACKAGE_JSON_CACHE: Cache<(String, String), CachedPackageJson, CacheWeighter> =
-        Cache::with_weighter(500, PACKAGE_JSON_CACHE_BYTES, CacheWeighter);
+        byte_bounded_cache(500, PACKAGE_JSON_CACHE_BYTES);
     static ref TARBALL_CACHE: Cache<(String, String, String), CachedTarball, CacheWeighter> =
-        Cache::with_weighter(200, TARBALL_CACHE_BYTES, CacheWeighter);
+        byte_bounded_cache(200, TARBALL_CACHE_BYTES);
 }
 
 /// Fetch a package's registry document, together with the registry it came from so
@@ -792,7 +818,13 @@ async fn extracted_tarball(
 
 #[cfg(test)]
 mod tests {
-    use super::{extract_tarball_within, resolve_version_spec};
+    use super::{
+        byte_bounded_cache, extract_tarball_within, resolve_version_spec, CachedTarball,
+        ExtractedTarball, MAX_EXTRACTED_BYTES, TARBALL_CACHE_BYTES,
+    };
+    use std::collections::HashMap;
+    use std::sync::Arc;
+    use std::time::Instant;
 
     fn packument() -> serde_json::Value {
         serde_json::json!({
@@ -841,7 +873,9 @@ mod tests {
             header.set_size(content.len() as u64);
             header.set_mode(0o644);
             header.set_cksum();
-            builder.append_data(&mut header, path, content.as_slice()).unwrap();
+            builder
+                .append_data(&mut header, path, content.as_slice())
+                .unwrap();
         }
         builder.into_inner().unwrap().finish().unwrap()
     }
@@ -857,5 +891,24 @@ mod tests {
         // A tarball's transfer size bounds neither of these
         assert!(extract_tarball_within(&archive, 40, 16).is_err());
         assert!(extract_tarball_within(&archive, 1024, 1).is_err());
+    }
+
+    /// A cache that splits its budget across shards refuses anything heavier than one
+    /// shard's share, silently declining the packages worth caching most.
+    #[test]
+    fn the_largest_extraction_allowed_still_fits_the_cache() {
+        let cache = byte_bounded_cache(200, TARBALL_CACHE_BYTES);
+        let key = ("registry".to_string(), "big".to_string(), "1.0.0".to_string());
+
+        cache.insert(
+            key.clone(),
+            CachedTarball {
+                extracted: Arc::new(ExtractedTarball { names: vec![], contents: HashMap::new() }),
+                bytes: MAX_EXTRACTED_BYTES,
+                fetched_at: Instant::now(),
+            },
+        );
+
+        assert!(cache.get(&key).is_some());
     }
 }
--- a/backend/windmill-api-npm-proxy/src/lib.rs
+++ b/backend/windmill-api-npm-proxy/src/lib.rs
@@ -189,18 +189,46 @@ struct CachedPackageJson {
     fetched_at: Instant,
 }
 
-#[derive(Clone)]
-struct PackageJsonWeighter;
+/// Type acquisition asks for a package's file tree and then for each `.d.ts` in it, all
+/// out of the same archive, so without this one package costs a download and an inflate
+/// per file rather than per package.
+const TARBALL_CACHE_TTL: Duration = Duration::from_secs(300);
+const TARBALL_CACHE_BYTES: u64 = 128 * 1024 * 1024;
 
-impl Weighter<(String, String), CachedPackageJson> for PackageJsonWeighter {
+/// A package tarball, inflated once. Paths have the `package/` prefix stripped.
+struct ExtractedTarball {
+    /// In archive order, `/`-prefixed, for the file tree listing.
+    names: Vec<String>,
+    contents: HashMap<String, Vec<u8>>,
+}
+
+#[derive(Clone)]
+struct CachedTarball {
+    extracted: Arc<ExtractedTarball>,
+    bytes: u64,
+    fetched_at: Instant,
+}
+
+#[derive(Clone)]
+struct CacheWeighter;
+
+impl Weighter<(String, String), CachedPackageJson> for CacheWeighter {
     fn weight(&self, _key: &(String, String), cached: &CachedPackageJson) -> u64 {
         cached.bytes.max(1)
     }
 }
 
+impl Weighter<(String, String, String), CachedTarball> for CacheWeighter {
+    fn weight(&self, _key: &(String, String, String), cached: &CachedTarball) -> u64 {
+        cached.bytes.max(1)
+    }
+}
+
 lazy_static::lazy_static! {
-    static ref PACKAGE_JSON_CACHE: Cache<(String, String), CachedPackageJson, PackageJsonWeighter> =
-        Cache::with_weighter(500, PACKAGE_JSON_CACHE_BYTES, PackageJsonWeighter);
+    static ref PACKAGE_JSON_CACHE: Cache<(String, String), CachedPackageJson, CacheWeighter> =
+        Cache::with_weighter(500, PACKAGE_JSON_CACHE_BYTES, CacheWeighter);
+    static ref TARBALL_CACHE: Cache<(String, String, String), CachedTarball, CacheWeighter> =
+        Cache::with_weighter(200, TARBALL_CACHE_BYTES, CacheWeighter);
 }
 
 /// Fetch a package's registry document, together with the registry it came from so
@@ -481,7 +509,7 @@ async fn get_package_filetree(
 ) -> JsonResult<PackageFiletree> {
     let (package, version) = parse_package_and_version(package_version_path.to_path())?;
     let (package_json, registry_url, auth_token) = fetch_package_json(&db, &package).await?;
-    let tarball_bytes = fetch_tarball(
+    let extracted = extracted_tarball(
         &package_json,
         &package,
         &version,
@@ -492,9 +520,10 @@ async fn get_package_filetree(
 
     // The entry point comes from the packaged manifest rather than the registry document,
     // which carries no `main` in its abbreviated form.
-    let (files, manifest) = extract_tarball_files_and_manifest(&tarball_bytes)?;
-    let main = manifest
-        .and_then(|manifest| serde_json::from_str::<JsonValue>(&manifest).ok())
+    let main = extracted
+        .contents
+        .get("package.json")
+        .and_then(|manifest| serde_json::from_slice::<JsonValue>(manifest).ok())
         .and_then(|manifest| {
             manifest
                 .get("main")
@@ -502,6 +531,12 @@ async fn get_package_filetree(
                 .map(String::from)
         })
         .unwrap_or_else(|| "index.js".to_string());
+
+    let files = extracted
+        .names
+        .iter()
+        .map(|name| FileEntry { name: name.clone() })
+        .collect();
 
     Ok(Json(PackageFiletree { default: main, files }))
 }
@@ -514,7 +549,7 @@ async fn get_package_file(
 ) -> Result<String> {
     let (package, version, filepath) = parse_package_version_and_file(full_path.to_path())?;
     let (package_json, registry_url, auth_token) = fetch_package_json(&db, &package).await?;
-    let tarball_bytes = fetch_tarball(
+    let extracted = extracted_tarball(
         &package_json,
         &package,
         &version,
@@ -523,10 +558,13 @@ async fn get_package_file(
     )
     .await?;
 
-    // Extract the specific file from the tarball
-    let file_content = extract_file_from_tarball(&tarball_bytes, &filepath)?;
+    let content = extracted
+        .contents
+        .get(filepath.trim_start_matches('/'))
+        .ok_or_else(|| Error::NotFound(format!("File {} not found in tarball", filepath)))?;
 
-    Ok(file_content)
+    String::from_utf8(content.clone())
+        .map_err(|_| Error::BadRequest(format!("File {} is not valid UTF-8", filepath)))
 }
 
 /// Stream a package version's tarball, so in-browser installers can unpack it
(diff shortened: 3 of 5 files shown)


rubenfiszel:
Round 1 addressed in 3e54aeccca.

**P1 tarball cache is not memory-bounded (codex)** — correct, and the distinction matters: the weighter bounded what was *retained*, while the whole archive was already inflated in memory by then, and the weight counted only file contents. Extraction is now capped as it goes, before anything reaches the cache:

- 64 MB of inflated content and 20 000 entries per archive, both enforced during the walk. Each entry is read through `take(headroom + 1)` rather than sized from its header, which an archive is free to lie about, so an overrun is detected rather than trusted.
- The weight now counts path strings and map keys as well as contents, plus a fixed 64-byte per-entry overhead, so an archive of many empty files can no longer sit in the cache at near-zero weight.
- Regression test `extraction_is_capped_before_the_archive_is_held_in_memory` builds a small gzipped tar in-process and asserts both caps reject it, with the limits injected so the case stays cheap.

**Nit: a memoized `false` from a failed `/npm_proxy/config` pins the session to CDN-first (claude)** — fixed. Only a definitive answer is kept now; a transient failure clears the memo and logs, so the next module re-asks. Test added: first probe rejects and falls to the CDN, second succeeds and the proxy is preferred.

**Nit: `/filetree` retains the whole inflated tarball with no per-package size guard (claude)** — the same 64 MB / 20 000-entry cap covers it, and it applies to `/filetree` and `/file` whether or not the result is cached.

Re-verified after the change: `/filetree/lodash/4.17.21` still returns `default: lodash.js` with 1054 files, `/file/lodash/4.17.21/lodash.js` 200. `cargo test -p windmill-api-npm-proxy` 2 passed, `npx vitest run src/lib/ata/apis.test.ts` 4 passed, `cargo check` and `npm run check` clean.

On the test-coverage note: `cargo test -p windmill-api-npm-proxy --lib` runs fine here, so the new cap test is exercised; the linking problem looks specific to the review sandbox.

rubenfiszel:
Fixed in d829fac562, and the finding was right on the mechanism and the arithmetic. Verified against `quick_cache` 0.7.0 rather than taken on trust:

- `OptionsBuilder::build` defaults `shards` to `available_parallelism() * 4`, but `Cache::with_options` then halves the shard count while `estimated_items_capacity / shards < 32`. With 200 estimated items that lands on **exactly 4 shards** regardless of core count, so 32 MiB per shard.
- Admission is stricter still than the shard's weight capacity: `CacheShard::insert` rejects when `weight > self.weight_target_hot`, and `weight_target_hot = shard_weight_capacity * DEFAULT_HOT_ALLOCATION` where that constant is `0.97`. So the real ceiling was ~31.04 MiB against a 64 MiB extraction cap, and an over-ceiling item goes to `handle_insert_overweight`, which removes any existing entry and returns without inserting. Silent, exactly as described.

Both caches are now built through one `byte_bounded_cache` helper with `shards(1)`, so the per-shard budget is the whole budget and the ceiling is `budget * 0.97` — 124 MiB for tarballs against the 64 MiB extraction cap, comfortable headroom. A single shard is the right trade here: these caches see a handful of requests per editor session and hold nothing but map operations under the lock.

**The sibling packument cache had the same defect** and I fixed it in the same helper. It was `with_weighter(500, 64 MiB)` → 16 shards → ~3.88 MiB ceiling, so any registry that ignores the abbreviated `Accept` and answers with a full document for a large package would never have had it cached. That one shipped in #10629; since that merged while I was working, the fix rides here instead.

Test as requested: `the_largest_extraction_allowed_still_fits_the_cache` inserts an entry weighing `MAX_EXTRACTED_BYTES` into a cache built the production way and asserts it is still retrievable. It weighs from the stored `bytes` field, so it costs no allocation. I confirmed it is not vacuous by reverting the constructor to `Cache::with_weighter` — it fails on the `get` with the old sharding.

This branch is also rebased onto `main` now that #10629 is merged; the diff is down to this PR's own three commits.

Re-verified: `cargo test -p windmill-api-npm-proxy` 3 passed, `cargo check` clean, `npx vitest run src/lib/ata/apis.test.ts` 4 passed, `npm run check` 0 errors.

rubenfiszel:
Restructured in bdaee7005a. The finding is right, and it identifies a real regression I introduced: my 64 MiB extraction cap made large-but-legitimate packages *worse* than before this PR, turning a slow-but-working path into a per-file hard failure that never caches. Reproduced it — `next@16.3.0` errored out of `/filetree` before the fix.

The mistake was bounding the wrong thing. The cache held the **inflated** map, whose size the archive controls, so the only way to bound it was a cap on expansion — and any cap is a cliff for packages past it.

**The cache now holds the downloaded archive, not its inflated contents.** That bounds it by something already measured, the transfer size, and every request walks the archive holding one entry at a time, so peak memory per request is one entry regardless of package size:

- `TARBALL_CACHE` stores `Bytes` weighted by actual length, 256 MiB budget, single shard.
- `walk_tarball` streams entries to a visitor; `/filetree` collects names and reads only `package.json`, `/file` stops at the match.
- The size caps are gone. What remains is `MAX_ENTRY_BYTES` (32 MiB, one entry's body) and `MAX_ENTRIES` (100 000), both far above any real package, so they bound abuse rather than expressing a size policy. A package too large to keep in the cache is still served correctly, it just costs a walk per request.
- Trade-off, stated plainly: one inflate per request instead of one per package. That is CPU on an archive already in memory, against the download per file this PR exists to remove.

Verified exactly as you asked, against the package you named:

- `/filetree/next/16.3.0` → 8519 files, `default: ./dist/server/next.js`, in ~4s cold. It **errored** under the previous cap.
- 1 filetree + 6 `/file` reads on that package → **1 packument fetch and 1 tarball download** at the registry, and reading a real `.d.ts` from it (`/dist/pages/_app.d.ts`) returns 200 with no further download.
- No regression on small packages: `lodash@4.17.21` still `default: lodash.js` / 1054 files, `/file` 200, `/tarball` 200.
- Script editor ATA still resolves entirely through `/npm_proxy/*` with zero jsdelivr, `ATA done`, types in the runtime.

@claude's related nit — the extraction cap counted file bodies only, with entry paths retained twice and charged to neither budget — is moot under this design: nothing is retained but the archive, and the weight is its real length.

`cargo test -p windmill-api-npm-proxy` 3 passed, `cargo check` clean.

rubenfiszel:
Both findings addressed in defb93ca31, and the P1 was a repo-rule violation I should have caught myself — `AGENTS.md` says "Never block the async runtime. Use `spawn_blocking` for CPU-intensive work", and I had gunzip plus a full tar walk running inline in an async handler on what this PR makes the *primary* ATA path.

**P1, in three parts:**

- **Off the runtime.** All archive reading now goes through a `blocking()` helper wrapping `spawn_blocking`.
- **Skipped entries are bounded too.** Correct that `MAX_ENTRY_BYTES` only covered the matched file — a tar entry can only be advanced over by decompressing it, so everything before the match was inflating unbounded. A `BoundedRead` now wraps the decoder itself, so one cap covers the whole walk, entries skipped past included.
- **No repeated full-prefix scans.** An archive is read once, on the way into the cache, keeping its file list plus the entries a type request can actually ask for: the manifest and `.d.ts` files, which is exactly what `ata/index.ts` reads (`treeToDTSFiles` filters to `.d.ts`; the caller loop fetches each dependency's `package.json`). Everything else — bundles, maps, licences — is walked past and read back on demand, so `/file` for a `.d.ts` is now a map lookup, not a scan. Six concurrent `.d.ts` reads on `next@16.3.0` went from ~1.15s to **6ms**.

**P2:** the cache test used a 4 KiB budget instead of allocating 128 MiB.

Two things surfaced while verifying, both of which would have shipped as silent wrong answers:

1. **A size-only retention rule gets `next` wrong.** `next` orders `package.json` **3748th**, behind ~3700 licence files, so a byte budget was spent before reaching the manifest and `/filetree` silently reported `index.js` instead of `./dist/server/next.js`. That is why retention is by *what is read* rather than by size.
2. **`strip_prefix("package/")` was wrong for a whole class of packages.** npm roots its own tarballs at `package/`, but DefinitelyTyped roots at the type name — `@types/express` unpacks under `express/`. The proxy returned an **empty file tree** and 404s for every `@types/*` package. Pre-existing, but this PR makes the proxy primary, so it would have turned a working jsdelivr path into a broken one. Now the leading component is stripped whatever it is.

Verified:

- `@types/express@5.0.6` → 4 files and `index.d.ts` 200, where it was 0 files and 404 before. In the script editor, ATA now loads the entire `@types/lodash` declaration set through the proxy — hundreds of `.d.ts` — with zero jsdelivr and `ATA done`.
- `next@16.3.0` → 8519 files, `default: ./dist/server/next.js`, one packument and one tarball at the registry for filetree plus repeated file reads; a large non-retained file (`dist/server/next.js`) still reads 200 through the on-demand path.
- `lodash@4.17.21` unchanged: `default: lodash.js`, 1054 files.
- `cargo test -p windmill-api-npm-proxy` 3 passed, `cargo check` clean, `npm run check` 0 errors.

rubenfiszel:
Fixed in 821f9fed86. The finding is right and the arithmetic checks out: an empty `.d.ts` weighed zero by the old measure while still owning a `String` in `names`, a `String` key, a `Vec` body, three separate heap allocations and a map bucket. And because empty contents never advanced `retained`, the per-package budget never bound on them either — both halves of the same mistake.

Both are charged now:

- `retained_bytes()` adds a fixed `NAME_OVERHEAD` (64) per listed name and `RETAINED_FILE_OVERHEAD` (160) per retained file, on top of the string contents. Deliberately generous rather than precise — the point is that an entry cannot be free, not that the number is exact.
- The same per-entry cost counts against `RETAINED_BYTES` while reading, so an archive of many empty files exhausts the budget instead of being retained without limit.

Worst case is now bounded and honestly priced: at `MAX_ENTRIES` (100 000) a pathological archive is charged ~6.4 MiB of names plus ~16 MiB of retained files against the 256 MiB budget, rather than the near-zero it used to report.

Regression test `an_archive_of_empty_files_is_not_free` builds a 2000-entry archive of empty declarations and asserts the weight is at least `entries * (NAME_OVERHEAD + RETAINED_FILE_OVERHEAD)` above the archive itself — it fails on the old contents-only measure.

Re-verified the three archive shapes are unaffected: `next@16.3.0` 8519 files / `./dist/server/next.js`, `lodash@4.17.21` 1054 files / `lodash.js`, `@types/express@5.0.6` 4 files (its manifest carries an empty `main`, faithfully reported). `cargo test -p windmill-api-npm-proxy` 4 passed, `cargo check` clean.

rubenfiszel:
Both fixed in 5cbf945c91.

**P1 — declaration-heavy packages re-inflating per file.** The analysis is right, and the root cause is that I picked 32 MiB out of the air rather than from what packages actually hold. `aws-sdk@2.1692.0` carries ~51 MB across 442 declarations, so the budget ran out partway and the remainder fell through to a per-file archive scan.

The bounds are now sized against real packages instead: `RETAINED_BYTES` 32 MiB → **128 MiB**, `RETAINED_ENTRY_BYTES` 512 KiB → **4 MiB**, with a comment recording *why* those numbers (aws-sdk is the heaviest known at ~51 MB, so a 2.5× margin) and that the global bound is the cache's own byte budget, not a per-package cap. Anything past them still degrades to a read rather than an error.

Measured on the package you named, cold, at ATA's concurrency of six:

- `/filetree/aws-sdk/2.1692.0` → 2287 files, 442 declarations, `default: lib/aws.js`, 1.26s.
- **All 442 declarations fetched in 0.29s**, every one 200, against **one packument and one tarball** at the registry. No fallback scans, so none of the ~9.5 GB of projected inflation.

**P2 — retention budget hiding the manifest.** Correct, and it is the same wrong-entry-point failure I hit with `next`'s ordering, reachable again through the budget or through a manifest above the per-entry cap. `/filetree` no longer assumes: on a miss it reads `package.json` back out of the cached archive via `read_one_entry`, and only falls through to `index.js` when the package genuinely has no manifest. Regression test `a_manifest_past_the_bounds_is_still_readable` squeezes the per-entry bound so the manifest is excluded, then asserts it is still readable and still yields the declared `main`.

The retention bounds are now a `Retention` struct with a `PRODUCTION` constant, so the tests exercise the real code paths with tiny values instead of building 4 MiB fixtures — same reason the cache test uses a 4 KiB budget.

Unaffected: `next@16.3.0` 8519 files / `./dist/server/next.js`, `lodash@4.17.21` 1054 / `lodash.js`, `@types/express@5.0.6` 4 files. `cargo test -p windmill-api-npm-proxy` 5 passed, `cargo check` clean, `npm run check` 0 errors.

rubenfiszel:
All four addressed in 39b614510e.

**P1 (codex) — retention budget ignored path bytes.** Correct, and the sharper half is that `names` grows for *every* entry regardless of `worth_retaining`, so it was not bounded by anything: 50 000 entries with 8 KiB paths sat entirely outside the budget. Paths are now charged like any other bytes — once when the name is listed, again when it becomes a map key — and the check runs *before* each insertion rather than as a gate at the top of the loop, so the bound cannot be overshot by the entry that crosses it. Regression case in `an_archive_of_empty_files_is_not_free`: a single 2 KB path against a 512-byte budget is refused, which passed before.

**P2 (claude) — the manifest test asserted `read_one_entry` rather than the fixed arm.** Right, and it was already covered by the `dist/bundle.js` read-back in the sibling test. The lookup-then-fallback is now `fn entry_point(&PackageArchive)`, which the handler calls inside `blocking()`, and the test asserts through it. Verified it is not vacuous: replacing the `None => read_one_entry(…)` arm with `None => None` fails the test. It also now pins the other direction — a package with no manifest still gets `index.js`.

**P2 (claude) — "the cache budget bounds the total" ignored reads in flight.** Also right: `read_package_archive` builds its map on the blocking pool before `insert` ever weighs it, so resident memory is `in-flight reads × RETAINED_BYTES` plus the cache. The comment says that now instead of the wrong thing, and `const _: () = assert!(RETAINED_BYTES * 2 <= TARBALL_CACHE_BYTES);` keeps the two constants moving together, including the margin to the single shard's ~0.97 admission ceiling.

**P2 (claude) — three doc/code mismatches.** The retention rationale moved from `strip_archive_root` to `worth_retaining` where it belongs, with root-stripping keeping its own; "All three" → "Both", and the two size constants split so each documents itself; `small_files` no longer claims a `package/` prefix.

Re-verified after the budget accounting change, since it touches every read: `aws-sdk@2.1692.0` 2287 files / `lib/aws.js` with all **442 declarations in 0.26s** at concurrency six, `next@16.3.0` 8519 / `./dist/server/next.js`, `lodash@4.17.21` 1054 / `lodash.js`, `@types/express@5.0.6` 4 files. `cargo test -p windmill-api-npm-proxy` 5 passed, `cargo check` clean, `npm run check` 0 errors.

rubenfiszel:
Rebased onto `main` (which now carries #10629 and #10630), conflicts resolved, and the last round's nit fixed in 42619397b6.

**Conflict resolution** — `frontend/src/lib/ata/apis.ts`, where #10630 rewrote the same error handling this PR restructures. Resolved by keeping **all** of #10630's behaviour and layering the ordering on top, rather than taking either side wholesale:

- `backendProxyApi` keeps `return await` and the `console.warn` carrying the cause.
- `api()` keeps its `.catch()` and the settled over-limit `Error`.
- `text()` keeps the version that catches around both the request and the body read.
- `getDTSFileForModuleWithVersion` is the `fromPreferredSource` form, with #10630's proxy-failure log preserved inside the proxy thunk — that log is the only report of a failed acquisition, since `ata/index.ts` drops the value.

`apis.test.ts` conflicted as both-added; merged into six cases so neither side loses coverage: proxy-first when configured, CDN-first when not, re-ask after a failed config probe, proxy→CDN fallback (this PR's direction), and #10630's two CDN→proxy rejection cases for metadata and `.d.ts`. #10630's tests needed `workspaceStore.set`, which the module mock here supersedes, so they now run against the mock and gained an explicit `registry_configured: false` — making the scenario they always meant ("no registry, CDN unreachable") explicit rather than incidental.

**Nit: exhausting `RETAINED_BYTES` refused the package where it used to degrade.** Correct, and the distinction between the two charges is the right one: they mean different things and now behave differently.

- The **path** charge still errors — `names` must be complete or `/filetree` returns a wrong answer, not a slow one.
- The **contents** charge now `continue`s. The file stays reachable through `read_one_entry`, so stopping bounds memory exactly as well as refusing while leaving the package servable. It also no longer overshoots: the cost is checked before being added rather than after.

The two comments that said the opposite are corrected, and the test now asserts both halves — an over-budget archive still lists all four names with an empty `small_files`, while too many entries or too many path bytes are refused.

Verified after the rebase and the change: `cargo test -p windmill-api-npm-proxy` 5 passed, `cargo check` clean, `npx vitest run src/lib/ata/apis.test.ts` 6 passed, `npm run check` 0 errors, and the four archive shapes unchanged — `aws-sdk@2.1692.0` 2287 files / `lib/aws.js`, `next@16.3.0` 8519 / `./dist/server/next.js`, `lodash@4.17.21` 1054 / `lodash.js`, `@types/express@5.0.6` 4 files.

PR #10633, opened 2026-08-11
feat: AI agent evals: datasets, scored runs and comparison

hugocasa:
## Summary

Reusable AI agents ship as `ai_agent` resources, but an agent can only ever execute as a flow
step, and there is nowhere to keep the inputs that exposed a problem in production. So there is no
way to answer the only question that matters after editing one: is it better?

This adds datasets of cases, runs of a dataset against a chosen state of the agent, scorers that
measure those runs, and comparison between two of them.

Evals open as a modal from the two places an agent already is — the linked agent card in the
flow editor and the `ai_agent` row on `/resources` — each of which already knows which agent it
is about.

agent card

## The surface

**The home is the agent's history.** One table of every run of this agent, newest first, whichever
dataset each is of, with one badge per scorer. A list spanning datasets cannot hold every
dataset's scorers to look a name up, so a score carries its name and kind along with its number,
and thresholds are joined in per run and column — still one grouped query over `eval_score`
rather than a read of each run's cells.

Every column that ran is reported, including one that produced no number: a judge that failed on
every case of a run says `failed` rather than dropping out of the row, where it would read as a
column nobody had asked for. A column with no cells at all is left out — that one was added after
the run and has nothing to say about it.

Both moves the screen offers sit above the table: starting a run, and curating the set a run is
of. A dataset is a thing someone adds to between runs, so reaching it only through the run dialog
would be a detour through a run they are not making yet.

runs list

An agent that has never been run says so, with the one move it has where the first row would be.

no runs

**Run asks what to run**: which state of the agent, and which dataset. Latest resolves the agent
when the run executes, the way a flow step does. A past version is inlined as it was, which is the
only way to run one: a linked step resolves the resource live, so pinning is exactly what a
reference cannot express. Unsaved edits run the step's edits as they are when Run is pressed, and
are offered — preselected — only when the dialog was opened from an agent being edited; every older
version is one click further, in the overflow menu, since running one is a deliberate act.

run dialog

The dataset field is a way into the dataset as well as a choice of one, since what you are about
to measure against is exactly when you notice a case is missing from it. Editing or creating one
from here hands the screen to the drawer and takes it back afterwards, on the dataset that was
just made, with the version you had already chosen still chosen.

**A run is one flow** — a loop over the cases, each iteration answering and then scoring — held by
one job, so it can be watched, cancelled and rerun as the single thing it is. A case's scorers are
a parallel `branchall` inside its iteration, each branch named after the column it is, and each
allowed to fail on its own: one judge erroring is one cell, not a run. Scores are read as they
land rather than at the end, so a long run fills in while it goes.

Between the agent and the scorers sits one named step, `Assemble the run the scorers read`. It is
a step rather than an input transform because the payload is more than the agent's result: every
tool call carries the arguments, result, status and duration of the job that ran it and the schema
of the script version it ran, and the agent step's own status and duration are reported rather
than the iteration's, which is still running and would otherwise be measuring itself.

It is a `//native` step: one fetch returning JSON is what the native runtime is for, and it runs
in the worker's own isolate instead of starting a bun process per case. That puts it on the
`nativets` tag, which the seeded `native` worker group serves rather than the default one — the
compose file ships that worker, but an instance without it would queue every iteration rather
than fail, so the run dialog says so before you spend anything.

no native worker

eval flow

Opening one shows its cases against a baseline, offered as the run before it *of the same dataset*.
A case's panel is the case as that run executed it: what each column made of it and why, the
answer as it was recorded, and the job behind the whole iteration.

run view

**A run is permanent.** Written once and then only ever read: there is no writable experiment, no
partial rerun, and no cell that can be edited after the fact. That is also why the column header
over a run reports and nothing else: a control there that changed the columns would be editing the
past from the one place that must not.

A scorer that has nothing to measure on a case returns `{score: null}`, which is a verdict rather
than a failure: the cell shows `n/a`, keeps its reason, and is left out of the column's mean and
pass rate instead of counted as a zero. A column asking whether sources were cited has no verdict
on a case with nothing to cite, and Braintrust spells the same thing the same way. Written out
rather than merely absent, since a scorer that returns nothing at all is a scorer that is broken,
and that stays an error. `comment` is also read as `reason`, so a scorer written for LangSmith or
Langfuse keeps its rationale.

The template scorer uses it. It scores the one thing its column is about — whether the answer
matches the case's expected one — and reports how the agent got there as `checks` beside the
number rather than averaged into it: a right answer that was slow and a wrong answer that was fast
must not score the same. A case with no expected answer is left unmeasured, which also fixes a
defect in the old template, where an absent `expected` produced an empty needle and failed the
check. Below, the same case under both: the old column invents 0.83, the new one says n/a.

a case a scorer does not measure

**Scorers live with the dataset**, beside the cases they measure. A scorer is any runnable taking
`(input, output, expected)` and returning a number, a boolean or an object with a `score`, so
deterministic checks are scripts and LLM-as-judge is a reusable agent used as a scorer, and
scoring needs no engine of its own. The name is the dataset's own; the pass threshold is an
interpretation of scores already recorded, so moving it re-reads every run rather than running any
of them again.

The whole dataset is assembled before it exists — its name, its columns and its cases — and
created in one request. Nothing is written while it is being put together, and Save is what writes
an edit.

The cases are edited in the grid the data tables are edited in, rather than a list beside a form.
Curating a dataset is comparing cases against each other, which a pane showing one at a time
cannot do. A cell holding an expected answer keeps the text as typed and stores what it parses to,
so a structured answer can be written by hand and a half-written one is not reformatted under the
cursor. A dataset is capped at a thousand cases, which is also what one page of a listing returns,
so the editor always holds the whole set and Save always writes all of it.

Editing a cell needs a cell editor AgGrid does not have: its `agTextCellEditor` is an `<input>` and
cannot hold a newline, and its `agLargeTextCellEditor` opens a fixed ten-row box under the cell for
every edit. `MultilineCellEditor` sits exactly where the cell was and grows a line at a time, with
Enter to commit and Shift+Enter for a newline. It also keeps Escape from leaving the grid, which
would otherwise close the drawer around it. It is used by evals only for now; adopting it for the
data tables means picking it per column rather than as a default, since a `defaultColDef` would
replace the editors AgGrid picks for typed columns.

dataset drawer

Adding a scorer offers four entries rather than two: new or existing, for an AI judge or a code
scorer. Writing a scorer and picking one that already exists are different jobs, and which of the
two you are doing is the first thing the form needs to know.

A new judge is created with an output schema for `{score, reason}` rather than a prompt asking for
that shape politely: a model quoting the agent inside its own reason used to break the JSON around
it. Windmill delivers the schema whichever way the model takes it — a forced tool call for Claude
and Bedrock, the provider's own parameter otherwise — so there is no list of models to maintain
here. A verdict that arrives as near-JSON anyway is still salvaged.

new judge

Reuse picks from the scorers already measuring something in this workspace, each named with the
dataset it is a column of. Picking is not adding: the drawer's own button is what makes it a
column, as everywhere else in this surface.

reuse a scorer

## Navigation lives in the dialog header

`Modal` takes an optional `trail` of levels below its title, rendered as a breadcrumb whose
ancestors are the way back. The level you would return to wears a back chevron rather than getting
a control of its own beside the path: both would point at the same place, and the way back is a
level of the path rather than a second thing to find. The header is the one part of a dialog that
does not move, which is where a "where am I / how do I get back" affordance belongs; before this
it held a fixed title and each body placed its own back control somewhere in a toolbar.

Declarative rather than a push/pop stack: callers deep enough to need this already hold the state
that says where they are, so the dialog reads it instead of owning a stack they would have to keep
in step.

`Modal` also takes a `description` rendered under the title. In the header rather than at the top
of the body, because it describes the surface rather than whatever the body currently holds, and
because a body that scrolls would carry it away. Evals passes it only at the root: one level in is
about one run, not about what runs are.

Escape follows the trail — leaving a level rather than throwing away the surface someone navigated
into — and still closes at the root. That needs a dialog to know it is the surface being addressed,
so `Disposable` now answers `isTopmost()` and `Modal` asks before acting: it keeps Escape for
itself, so nothing was arbitrating between it and a drawer opened from inside it, and one key press
was moving both.

`Modal` also always portals now, as `Drawer` already did. Portalling only when an overlay host
happened to exist left it inheriting whatever stacking context an ancestor had, which is how a
dialog ended up painted under the nav.

Evals is the first caller: the runs list is the root and a run is a level inside it.

## Versions are numbered per resource

`resource_version.id` is one identity sequence for the whole table, so an agent saved nine times
read v4 … v24, and the gaps counted writes in workspaces the reader cannot see. Runs are named and
compared by their version, so the number has to be the resource's own count.

A new column holds it, backfilled per path and assigned on write as `MAX(version) + 1`. It needs
no lock of its own: the trigger runs inside the transaction that wrote `resource`, and a
concurrent write to the same path blocks on that row's lock — or on the primary key, for an insert
— before its own trigger can run. A unique index on `(workspace_id, path, version)` turns any
residual race into a hard error rather than two rows claiming v7.

Assigned on write rather than counted on read because both ways of deleting versions take the
oldest ones: the monitor's trim past `MAX_RESOURCE_VERSIONS`, and clearing a history down to its
current value. Counting the survivors would renumber a version a recorded run already names.

`id` stays how a version is addressed, so the history routes and restore are unchanged apart from
the path parameter being renamed to `id` — it never held a version number, and now that one
exists the old name reads as the wrong thing.

This also fixes the resource history drawer from #10596, which labelled versions by that id.

resource versions

## Edits to a linked agent live in the step, nowhere else

Editing an agent from a flow step forks it into the step, which is what makes the edits runnable
there. The step is the only copy: nothing is saved anywhere until you decide. **Save changes**
writes the edits to the agent and re-links the step; **Cancel** drops them and re-links, asking
first when there is something to drop; the unsaved-changes badge opens the diff, whose **Discard
changes** is Cancel without the question. Evals from the editing card run the edits as the step
holds them when Run is pressed — the request carries them, and the run records what it was handed,
inlined into its flow and hashed; evals from the linked card (and from `/resources`) run the
deployed agent, the same reading a linked step makes at run time. The agent's own resource draft,
the one the resource editor writes, is untouched by any of this and never read by evals.

Because the editor can unmount mid-session (another node selected, the step's tab switched), what it
needs to judge the edits travels with the session in `agentEditStore` rather than with the
component: the agent as deployed, in the same normalised form the step's configuration is compared
in. Without that baseline the badge could not tell an edit from the deployed value, which is the
difference between Cancel asking and Cancel dropping work unannounced.

editing card

dataset drawer: scorers and cases

cancel asks when there is something to drop

run dialog opened from the editing card

(Earlier revisions kept the edits as the agent's resource draft, first mirrored on every keystroke
and then written at commit points; each round of review found another reader of that second copy
that had to reason about whether a write had landed, failed or been refused. Editing a linked agent
in an editor of its own is a follow-up; until then the step is the one copy and there is no draft
to keep in step with it.)

## Changes

**Running** — a run is a `RawFlow` preview: a forloop over the cases whose body is the `aiagent`
step linked to the subject, followed by the scorers as a parallel `branchall`. That is the vehicle
`ModuleTest.svelte` already uses to test an agent step, so a case exercises the production
linked-agent branch of `ai_executor.rs` rather than a parallel one.

- `windmill-api/src/ai_evals/` (new): dataset, case and scorer CRUD, the run and results
  endpoints, and the cross-dataset run list.

**Where results live** — results are jobs; nothing is stored twice. `runnable_path` is stamped
with the agent's own path, so the existing job filter answers "every run of this agent" with no
new state, and `_eval` in the flow args records
`{subject: {kind, path, version}, dataset, experiment_id}` so a job opened cold from the runs page
explains itself.

**Storage** — `eval_dataset`, `eval_case`, `eval_experiment`, `eval_experiment_case`,
`eval_score`, added by a single migration expressing the final schema. Only the input side is
stored; what a run produced stays in `v2_job` and is read back when results are displayed. A run
records its case set by value, and `case_id` is a plain column rather than a foreign key, so
deleting a case does not rewrite the history of the runs that used it. Deleting a dataset (from the
drawer, behind a confirmation) cascades to its cases and runs through the foreign keys and leaves
their jobs alone.

Cases are capped at 256 KiB each, and a dataset at 1 000 cases and 16 MiB, refused rather than
truncated: attachments are S3 references rather than inline bytes, so a case near a cap is a
mistake rather than a use case.

**Permissions** — row-level security on `eval_dataset` decides who may read or write a dataset, so
an `extra_perms` grant works and the rule is not mirrored in Rust. Runs and scores carry a read
policy derived from their dataset and no write policy at all: they are written on the unrestricted
pool after the dataset row itself has been asked, with `SELECT ... FOR UPDATE`, whether this caller
may write it — which applies the dataset's UPDATE policies on top of its SELECT policies. Reading
a run's payload additionally requires the `jobs:read` scope, since it reads job arguments and
results.

**Flow editor** — the step picker can link a saved agent on insert, Cancel restores the link, the
agent card is simplified and carries the version its runs will be recorded against, and an agent's
in-progress edits stay in the step and can be run by evals as a subject of their own, as described
above.

**Docs** — `docs/ai-agent-evals.md`, plus a pointer from `docs/reusable-ai-agents.md`.

## Test plan

Exercised against a real backend with an OpenAI-compatible stub as the provider, so the agent path
ran for real rather than being compiled only. Evals need neither the `parquet` feature nor a
configured workspace object storage.

- [x] Dataset, case and scorer CRUD round-trip; removing a dataset row cascades to its cases and
      runs through the foreign keys and leaves the jobs those runs produced.
- [x] A run over a dataset scored by a script and by an agent-as-judge produces the expected
      numbers in every column, on the run and as badges in the list.
- [x] A dataset created with its scorers and cases in one request arrives holding all three, and
      an edit writes only on Save.
- [x] A column that failed on every case of a run is reported as failed rather than omitted, and a
      column added after a run stays out of that run's row.
- [x] Scores appear per column while a run is still going, and one scorer failing leaves the other
      columns of that case scored; a running run's badges update on the list even before any
      dataset has been opened.
- [x] Pinning a past version runs that version: v6 and v9 of the agent differ only in
      `max_iterations` (3 vs 5), and the agent jobs of a run pinned to v6 ran with 3.
- [x] A pass threshold moved after the fact re-reads the scores already recorded, without running
      anything again.
- [x] Version numbers are per resource and contiguous, in the run list, the run dialog and the
      resource history drawer; a rename carries them, clearing past versions does not renumber the
      survivors, and delete-and-recreate restarts at v1.
- [x] The down migration reverts cleanly, the restored trigger still records, and re-applying
      renumbers identically.
- [x] The header trail navigates: the back chevron returns to the level under the one you are on,
      Escape leaves one level and closes at the root, and a drawer or nested dialog opened over the
      modal takes Escape on its own without the level behind it moving. Measured: the header row
      sits at the same offset and keeps its height at the root and one level in, and a description
      shown at the root does not move the title.
- [x] Creating a dataset from the run dialog returns to it with that dataset selected; saving an
      edit made from there does too, on the renamed dataset; cancelling returns with the previous
      choice, and a pinned version survives either.
- [x] Saving the dataset drawer closes it, and reopening seeds it from the dataset as it now
      stands; New dataset on the runs list opens the same drawer.
- [x] A scorer created from the template scores a case with an expected answer, leaves one without
      unmeasured, and carries its checks either way — including the two that report a `detail`.
      Over the fixture: 1.00 and n/a, column mean 1.00 of one scored cell. The template round-trips
      out of the Rust const byte for byte and typechecks under `tsc --strict`.
- [x] A scorer returning `{score: null}` records the cell as not applicable rather than an error:
      the API reports `not_applicable` with `pending: false`, the column's mean is over the cases
      it measured (1.00 of one scored cell, not 0.50 of two), the harvest does not overwrite it on
      a later pass, and the table shows `n/a` with the scorer's reason on hover.
- [x] Cases edit inline: a cell writes through to the case, an expected answer typed as JSON is
      stored as JSON and one typed as prose as a string, a row added is dropped without being
      asked about, and a row the dataset already holds still asks.
- [x] The cell editor sits at the row's height for one line and grows 20px a line; Enter commits,
      Shift+Enter adds a line, Escape cancels the edit and leaves the drawer open.
- [x] A run pushed after naming the payload step carries the summary in its `raw_flow` and shows
      it in the flow graph, and the run still scores every case.
- [x] The payload step runs on a `native` worker: its jobs are tagged `nativets`, land on
      `wk-native-*`, complete in ~40-50ms and return the assembled payload rather than the
      `require deno_core` stub a worker without the runtime would produce. The run dialog warns
      when no worker serves the tag and is silent once one does.
- [x] The job's `runnable_path` and `_eval` args carry subject, version, dataset and experiment.
- [x] Both entry points open the modal (flow editor card, `/resources` row).
- [x] End-to-end on the final schema in a real browser: create a dataset, add a case, add a code
      scorer, run twice; answers land, scores are harvested into badges, and the comparison
      baseline is offered.
- [x] `cargo check` with and without `parquet`; `SQLX_OFFLINE=true` build; `npm run check`
      (0 errors, no new warnings); frontend unit tests.
- [ ] An oversized case and an over-full dataset are refused with a readable message.

## Not in this PR

Capturing a case from a real run (returning as a pick-from-recent-runs flow in the dataset
drawer), rescoring an existing run's stored answers with edited scorers, reproducing a host
flow's `tool_inputs` overrides in a run (the agent runs with its authored defaults), a model
matrix (one dataset across several models in one run), online evals on sampled production
traffic, a standalone route for reaching a dataset outside an agent context, and a link from a
score straight to the scorer's own job rather than to the iteration that contains it.


hugocasa:
On the P0 in the last Codex review ("deleting one dataset can delete another dataset's experiments" for `f/team/foo` vs `f/team/foobar`) — this one does not reproduce, so I have not changed the prefix.

`ObjectStore::list` matches on whole path segments, not raw string prefixes. Verified against MinIO (an S3 backend, so the raw-prefix `ListObjectsV2` path):

- created `u/admin/foo` and `u/admin/foobar`, each with an experiment
- listing `foo`'s experiments returned only `foo`'s
- deleting `foo` left `foobar`'s experiment intact

I have added a comment at the deletion site recording that invariant, since replacing the listing with string-prefix filtering *would* introduce exactly this bug.

The other five findings were real and are fixed in 89eaafe6d4.

hugocasa:
### On "dataset edits rewrite historical run columns" (Codex, round 25)

Answering rather than changing the behaviour in this PR.

**What happens.** The runs list and the results table build their columns from the dataset's
current `scorers`, so removing a scorer takes its column off the runs already recorded as well.
The `eval_score` rows are never touched: `datasets.rs` only rewrites `eval_dataset.scorers`, and
nothing deletes scores. What changes is what is rendered, not what is kept. Re-adding the same
script or agent mints a new id (`assign_scorer_ids` generates one when the client sends no valid
existing id), so it starts a fresh column rather than reconnecting to the old scores.

**Why it is deliberate.** A comparison table is worth having because every run in it is measured
by the same columns; deriving columns per run makes it ragged, which is a rendering question of
its own. `eval_score` also records only `scorer_id` and the scorer's config hash, with no name or
kind, so a column for a scorer the dataset no longer lists has nothing to label itself with. Making
history self-describing therefore needs scorer metadata recorded per run, not just a change of
query.

**What changed instead.** The removal confirmation used to say "the scores it already produced stay
on the runs that carry them", which reads as a promise that they stay visible. It now states the
display consequence at the point of action, and `docs/ai-agent-evals.md` records it alongside the
existing rule that a scorer added later has no cell in the runs that predate it.

**The likely follow-up**, deliberately not in this PR: keep a removed scorer in
`eval_dataset.scorers` marked as removed instead of dropping it, so it stops running on new runs
but keeps its id and its name and its historical column goes on rendering. `scorers` is `JSONB`,
so that needs no migration. It is a behaviour change worth reviewing on its own rather than an
addition at this point in this one.

hugocasa:
8f5eca9173 replaces the agent-draft mirror in `AgentResourceBar.svelte` rather than patching its next seam. The last rounds' draft findings (silent 409 stall, `draftLeftHere` outranking later fetches, the failure alert's remedy destroying the parked payload, `dropPending` after a failed flush dropping the only copy) were all readers of one asynchronous write: every keystroke was mirrored into the agent's resource draft, so the card on re-link, Edit's resume and the page-close flush each had to reason about whether that write had landed, failed or been refused.

Now the step is the only copy while you type. The draft is written at commit points only — Cancel, Evals from the editing card, Save changes, Discard — each `save({immediate: true})` with `last_sync`, awaited, outcome read from the syncer: a 409 raises `DraftSyncConflictModal` with the editor still open, a failure toasts and the editor stays open. Gone: the mirror effect, `draftLeftHere`, the `cardDraft*` state, the three post-Cancel alerts, the flush-before-read at Edit. The linked card's badge is the server's answer alone. Description section "An agent has one set of unsaved edits" rewritten to match; `docs/ai-agent-evals.md` too.

Exercised in a real browser: typing writes nothing; Cancel writes once and the linked card badges it; Edit resumes; Discard deletes; Save changes deploys and deletes; a newer draft from another tab makes Cancel raise the modal with the editor open, Load from server re-forks onto it; Evals-from-edit writes before the dialog opens; closing the tab mid-edit writes nothing.

hugocasa:
517ef6b7b8 changes the model again, and this time it is a removal rather than a rewrite: there are **no agent drafts** in this PR. The round-29 findings (Cancel deleting the draft after a remount; edits typed during the Cancel write) were both about the draft write, and the draft write is gone.

- Editing a linked agent forks it into the step as before; the step is the only copy of the edits. **Save changes** deploys; **Cancel** drops and re-links (with a confirm when there is something to drop); the unsaved-changes badge opens the diff, whose Discard changes is Cancel without the question.
- Evals from the editing card run the step's edits as they are when Run is pressed: the request now carries them (`EvalSubject.draft` for kind `agent_draft`), the server stamps `draft_hash` and the version and inlines them as before. Evals from the linked card and from `/resources` run deployed; the dialog offers *Unsaved edits* only when opened from the editing card (a caller-side getter, no server ask).
- Removed: `agent_draft_config` and every draft-table read (`subject_state` keeps only the version; the results endpoint drops `subject_has_undeployed_changes` / `subject_current_draft_hash`), `EvalsPane`'s subject poll and "stale run" line, the card's conflict modal / resume / `getDraft` read / the three alerts. Three orphaned `.sqlx` entries deleted (removal-only, no `prepare`).
- Kept from round 29's P1: the editing baselines ride in `agentEditStore` with the session, so a tab switch keeps the badge and Cancel's confirm honest. `agentEditStore.test.ts` pins it.
- `validate_subject` now refuses a configuration in the request for `agent` / `agent_version` and requires one for `agent_draft`; the docs say plainly what the server can and cannot assert about edits it is handed.

Exercised in the browser: Edit → type → tab switch keeps the badge → Cancel asks → drop re-links with nothing written anywhere; Edit again shows deployed; Cancel with no edits does not ask; Evals from the editing card offers `v11 + draft (current)` and the run request carries the edited prompt; Evals from the linked card offers no draft; Save changes deploys. A run of edits went through the new backend path end to end (`agent_draft`, v11, hash stamped, success).

hugocasa:
✅ Review round clean @ e3ab419cab337078adab528029bdeb64b456c6c6 — nit-only verdicts at 517ef6b7b86f6b3d04c66fafd84bade29fa0b032; nits addressed in e3ab419cab337078adab528029bdeb64b456c6c6 (diff badge made keyboard operable via Badge's `clickable`; docs no longer describe the deleted subject_state poller or a toggle-less dialog and state the in-memory edit-session cost; `subject_draft_hash` doc names its remaining reader; `validate_subject` also refuses an edits run without its edits, with a unit test)

hugocasa:
### Review triage — dismissed / deferred items

A few recurring review findings are intentionally not addressed in this PR; recording the rationale here so they don't keep getting re-raised each round.

**Dismissed (not a defect):**
- **"the `ai_evals` scope is offered but never enforced"** (Pi, Codex): it *is* enforced, by the route-scope middleware in `windmill-api-auth` (`check_scopes_for_route`), which maps `/w/{ws}/ai_evals/...` to the `AiEvals` domain by path + method. Verified with real scoped tokens: an `ai_evals:read` token gets **200** on `datasets/list` and **403** on `datasets/create`; an `ai_evals:write` token gets **200** on create; a `jobs:read`-only token gets **403** on list; and `experiments/run` correctly additionally requires `jobs:run`. The dataset handlers rely on the middleware rather than a per-handler `check_scopes`, which is the established pattern. A cold read of the diff can't see the middleware, hence the false positive.

**Deferred to a follow-up (beta-acceptable, cross-cutting):**
- **Per-dataset ACL API** — `eval_dataset` has `extra_perms` + RLS, but is not yet wired into the granular-ACL kind allowlist, so single-dataset sharing can't be configured via the ACL API. Datasets are still governed by folder/owner/group permissions; explicit per-dataset grants are the follow-up.
- **Agent-rename identity** — run history is keyed by the agent's path, so renaming the *agent resource* detaches its history. Fixing this needs a cross-subsystem change (hook resource-rename, or give experiments a stable agent id). Username renames *are* handled in this PR; the agent-resource-rename case is the follow-up.

**Now fixed (was previously flagged as an edge to defer):**
- The judge version/config race is resolved: `resolve_scorer` reads a judge's configuration and version in one caller-scoped snapshot, so the recorded version is always the one inlined.

hugocasa:
### Deferred to beta follow-up: MCP tool-output scoring

An MCP tool call runs inside the agent rather than as a child job, so its result and success state (which live in a later \`role: \"tool\"\` message, not a job row) are not yet surfaced to scorers — a scorer can inspect script/agent tool results but not MCP ones, and can't distinguish a failed MCP call. This is a documented known-limitation for the beta (see the comment at \`ai_evals/payload.rs\`); surfacing MCP results — including a persisted success flag from the executor — is a follow-up. Scoped to agents that use MCP tools.

This joins the other explicit beta follow-up — **agent-resource-rename identity** (run history is keyed by the agent path, so renaming the agent resource detaches its history; the *username*, *offboarding*, and *workspace-fork* lifecycle paths are all handled in this PR).

hugocasa:
### Deferred to beta follow-up (decided)

Recording the remaining review items intentionally left for follow-up, so they aren't re-raised each round:

- **Narrow concurrency races** — e.g. a dataset deleted and recreated at the same path during the seconds of a launch, so the run attaches to the replacement (`read_dataset_and_cases` releases its `FOR UPDATE` before the experiment insert). These require precisely-timed concurrent mutations on the same object mid-operation; accepted as beta known-limitations rather than held for a lock-across-the-launch redesign.
- **Per-dataset ACL API** — `eval_dataset` has `extra_perms` + RLS, but is not wired into the granular-ACL kind allowlist, so individual dataset sharing can't be configured via the ACL API yet. Datasets remain governed by folder/owner/group permissions; explicit per-dataset grants are a follow-up.
- **MCP tool-output scoring** — MCP calls run inside the agent (no child job), so their result/success aren't surfaced to scorers yet (see `ai_evals/payload.rs`).
- **Agent-resource-rename identity** — run history is keyed by the agent path, so renaming the *agent resource* detaches its history. (The *username*, *offboarding*, and *workspace-fork* lifecycle paths are all handled in this PR.)

Everything else surfaced by review — the RLS/atomicity refactor, scorer authorization + deployed-hash pinning, the workspace-lifecycle coverage above, canonical path validation, and load/launch error handling — is addressed and verified.

hugocasa:
Re: Pi's nit "the `ai_evals` token scope is registered but never enforced" — this is a false positive; the scope **is** enforced, just not per-handler.

Scope enforcement for workspaced routes happens in the auth middleware, not in each handler. `windmill-api-auth/src/auth.rs:835` runs `check_scopes_for_route(authed.scopes, path, method)` on every request from a scoped token, before the handler. `extract_domain_from_route` maps `/api/w/{ws}/ai_evals/...` to `ScopeDomain::AiEvals` (`scopes.rs:405`), so a scoped token lacking `ai_evals:*` is rejected at the layer. That's why no `ai_evals` handler calls `check_scopes` — the same as most other domains; `resources`/`variables` add *per-path* predicates on top because they filter by object path, which eval datasets don't need (RLS already scopes them by identity).

Verified empirically against a running instance with real scoped tokens:

- token scoped `jobs:run:scripts:*` → `GET /w/admins/ai_evals/datasets/list` → **403 "Access denied. Required scope: ai_evals:read"**
- token scoped `ai_evals:read` → same GET → **200**
- token scoped `ai_evals:read` → `POST /w/admins/ai_evals/datasets/create` → **403 "Access denied. Required scope: ai_evals:write"** (read cannot write)

So the scope the UI advertises does gate these routes, and read/write are distinguished correctly. No change needed here.

The two Codex nits from this round are fixed in 89a925f745: the scorer picker now filters to deployed script versions (matching `get_latest_script_hash`), and the doc line no longer promises baseline rescoring (which isn't implemented).

hugocasa:
Addressed the latest round in 5e1a5025b3 (browser-verified on a running instance):

- **P2 — `loadError` never cleared (EvalsPane)**: reset at the start of both loaders, so a recovered read clears the "Could not load evals" state.
- **P2 — 0–1 `pass_if` not enforced in the column-header drawer (EvalScorers)**: `thresholdError` now rejects out-of-range values, matching `assign_scorer_ids`. Verified: entering `5` disables Save, `0.7` re-enables it.
- **P1 — judge/agent picker ignored the acting workspace (AddScorer:440)**: the agent `ResourcePicker` now receives `{workspace}`. Verified it lists the acting workspace's ai_agent resources.
- **P2 cleanups**: removed the now-dead `MAX_DATASET_PATH_CHARS`, corrected the stale `missing_in_baseline` doc comment, and moved the two newly-added tests below the doc comments that describe their neighbors.

Consciously not doing these here:

- **P1 — "Any script" picker offers undeployed scripts (AddScorer:442)**: left as-is for now. The launch already fails fast with a clear, actionable message (`Scorer script <path> is not deployed or not readable`), so this is a UX papercut rather than a silent failure. The fully-correct fix touches the shared `ScriptPicker`/listing (which exposes only `has_deploy_errors`, not pending-lock state) and is deferred rather than special-cased in the eval form.
- **P1 — provider picker ignored the acting workspace (AddScorer:333)**: deferred. `AIProviderPicker` has no `workspace` prop, so this means threading a new prop through a shared copilot component (resource listing + model discovery); scoped as a follow-up since it only bites in fork/session eval usage.
- **P2 — expand/collapse implemented as a raw `<div role="button">` (AgentResourceBar)**: deferred. The row wraps a link and three buttons, so it can't become a single `<button>`; the design-system-correct fix is a separate disclosure control, left for a follow-up.

Pi's "`ai_evals` scope registered but not enforced" is a false positive — enforcement is in the auth middleware (`check_scopes_for_route`, auth.rs:835), verified earlier with real scoped tokens (`jobs:run` → 403 `ai_evals:read`; `ai_evals:read` → 200 read / 403 write).

hugocasa:
Addressed the latest Codex round in 7044b8c055 (and 81fe8ba30c before it):

- **Finding 1 (P1) — dataset not reselected in the Run dialog**: fixed. The drawer now runs `onCreated`/`onRenamed` (which move the pane's selected dataset) **before** `closeDrawer()`, so the stepped-aside Run dialog reopens reading the saved selection rather than a stale one. Verified end-to-end on a running instance: creating a dataset from the Run dialog now returns with that dataset selected and Run enabled (previously returned unselected). The rename path uses the same reordering.
- **Finding 3 (P2) — stale rescoring claims**: fixed. Reworded the OpenAPI `missing_in_baseline` description and the `eval_score` migration comment so neither advertises a rescore-existing-runs operation that isn't implemented (the `_sqlx_migrations` checksum was reconciled for the edited migration).

Deferring **Finding 2 (P1) — deployed runs indistinguishable across agent deletion+recreation** as a beta known-limitation. It requires deleting an agent resource and recreating it at the same path (version history cascades, restarts at v1); the old path-keyed experiments then attach to the recreated agent. The proper fix is a design change — an immutable incarnation identity for deployed subjects instead of keying on path — which is the same class as the launch/delete race and per-dataset ACLs already deferred. Recorded rather than special-cased.

Also from the prior round: the `settle_unspawned_cases` and `new_run` doc/code mismatches were corrected, the shared eval `loadError` was split per-loader, a `0` pass threshold is no longer dropped in AddScorer, and the subject version/hash reads are now caller-scoped (consistent with `subject_state`).

hugocasa:
Addressed the two Codex findings from this round:

- **Finding 2 (P1) — dataset rename authorization: fixed** in 93a0759e95. Good catch that the policy was copied from the standard tables — it was, faithfully. `eval_dataset`'s `see_extra_perms_user_update` has the same `USING((extra_perms->>'u/<user>')::boolean)` / implicit-`WITH CHECK` shape as `resource`, `app`, etc. The piece we lost in the RLS refactor is the Rust-side guard those tables pair with it: `update_resource` (windmill-store/src/resources.rs:1853) calls `require_owner_of_path(&authed, path)` on the rename branch, which restricts renames to the path **owner** (or admin), so a mere extra_perms write-grantee can't move an object into another namespace. `update_dataset` now does the same — content edits (summary/scorers/cases) stay available to write-grantees; only the rename is owner-gated. Verified by compile + code path; runtime API exercise was blocked by DB-connection saturation in the shared dev env making token auth intermittently fail on all routes (whoami included), unrelated to the change.

- **Finding 1 (P1) — stale agent edit overwriting a newer deployment: deferred** as a beta known-limitation. It's a concurrent-edit lost-update (two edit sessions of one agent; one saves v2, the other runs/saves a stale v1 fork) — the concurrency-race class already deferred on this PR. The proper fix is optimistic concurrency on the agent save (preserve the fork's base version, reject or rebase when the deployed version advances), a design change to the agent-edit flow.

Claude's nits (16 MiB cap undocumented, `deployedConfig`/`forkedConfig` redundancy, no test for the `eval_case` RLS write policy) are noted; happy to fold in the doc + redundancy ones.

hugocasa:
Addressed this round's nits in 27cab03875:

- **Claude nit 1 — malformed `ai_agent` value 400s the whole results/list page: fixed.** `readable_agent_state` now maps a `config_to_draft` failure to `None` instead of propagating it, matching its own contract ("None when … not an agent"). A bad row leaves the "since deployed" label unresolved on the results page and the run list rather than failing the entire response; `resolve_scorer`'s judge arm still refuses (None → "not a readable ai_agent resource"), which is correct for a malformed judge you can't run.
- **Claude nit 2 — failed tool call carried twice at 2× the judge cap: not a doubling in the judge's context (comment clarified).** `render_tool_calls` (payload.rs:382) matches `(Some(error), _) => "error: …"` — for a failed call it renders `error` and *not* `result`, so the judge's context (which is `results.<payload>.rendered`) carries the payload once and bounded. `result` remains on the raw call only for a script scorer, which reads `.error` to detect the failure (template.rs:151) and `.result` for what it returned. Reworded the comment to say that, per your "make the comment match the intent" note.
- **Codex nit — eval datasets missing from the granular ACL registry:** this is the per-dataset ACL sharing item already deferred as a beta known-limitation (no dataset endpoint mutates `extra_perms` yet; the RLS supports it for when the sharing surface lands).

Note: nit 1's runtime exercise was again blocked by the shared-env DB-connection saturation making token auth intermittently fail on all routes; verified by compile + unit tests + the all-callers reasoning above.

hugocasa:
✅ Review round clean @ 82d777a12a29fd04447e096850e7df79a6e974c2 — nit-only verdicts at af8f0b713b / ec3c9aa4b2 (Pi: good to merge; Codex, Claude, cubic: mergeable with nits); nits addressed in ec3c9aa4b2, fdec87c37b, 82d777a12a. Dismissed: cubic's Create-during-scorer-add race on 82d777a12a (the nested Add-scorer drawer's click-away closes it before Create can be pressed, and the created judge stays reusable), and the SelectDropdown wrapper-class item (the change restores main's own behaviour — SettingsSearchInput on main already shows the category line; the diff was read against the previous PR head).

PR #10634, opened 2026-08-11
fix: harden custom env var name handling in the nativets/bun prologue

hugocasa:
## Summary

Custom workspace environment variable **names** are spliced into the generated
NativeTS / Bun worker prologue (`const {name}` and `process.env['{name}']`), but
only the value was ever escaped. A non-identifier name could therefore alter the
generated program rather than just supplying data. This hardens name handling at
the point the prologue is built, and adds a light guard at the write path.

Every other executor passes env vars to the OS via `Command::envs` (real process
env, not generated code), so those paths are unaffected.

## Changes

- Add `escape_js_single_quoted` and `is_valid_js_identifier` / `can_bind_as_prologue_const`
  helpers in `windmill-common`.
- `worker.rs` (standalone NativeTS) and `bun_executor.rs` (`//native`): escape the
  name as a string literal in `process.env['…']`, and emit the `const {name}`
  binding only when the name can safely bind (valid identifier, not a reserved word,
  not a name the prologue already declares). Names that can't bind are still exposed
  via `process.env['…']`.
- `set_environment_variable`: reject non-identifier names for **new** names only, so
  editing the value of a pre-existing name still works. Deletion stays unrestricted.
- Document the name constraint on the endpoint in `openapi.yaml`.

## Test plan

- [x] `cargo check` clean under `quickjs` and `quickjs,deno_core`
- [x] Unit tests for the new helpers (`windmill-common variables::tests`, 3 passing)
- [x] Live API: new non-identifier / injection-shaped names → 400; valid new name → 200;
      value edit of a pre-existing non-identifier name → 200; deletion unrestricted
- [x] Modeled the exact generated prologue in V8 (Node) with hostile names: no code
      executes, values preserved, reserved/owned names produce a valid (non-erroring)
      prologue
- [x] Ran a real standalone NativeTS job and a `//native` Bun job in a `deno_core`
      worker with hostile env-var names planted directly in `workspace_env`
      (`_p=1;globalThis.PWNED_CONST=1//` and `a'];globalThis.PWNED_STR=1;process.env['b`):
      both jobs completed successfully, the hostile names came back verbatim as
      `process.env` data, no injected global was set, normal names still bind, and the
      worker log showed no panic

---
## Summary by cubic
Prevents code injection via custom env var names in the NativeTS/Bun worker prologue. Names are escaped and only bound as constants when safe; reserved words and strict-mode forbidden bindings (`eval`/`arguments`) are excluded, `async` is allowed, and unsafe new names are rejected by the API.

- **Bug Fixes**
  - Escape both keys and values in generated `process.env['…']` assignments.
  - Guard `const {name}` with `can_bind_as_prologue_const`: allow identifiers (including `async`); exclude JS reserved words, strict-mode forbidden (`eval`/`arguments`), and prologue-owned names (`process`, `BASE_URL`, `BASE_INTERNAL_URL`); otherwise use `process.env` only.
  - Validate on write: block new non-identifier names; allow edits/deletes of existing names; skip the existence check for valid names.
  - Add `escape_js_single_quoted`, `is_valid_js_identifier`, `can_bind_as_prologue_const` in `windmill-common`; update `worker.rs` and `bun_executor.rs`; document the constraint in `openapi.yaml`; add unit tests.

<sup>Written for commit 86683d258371c40489ef2a1afd5512ec22091b6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10634?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.sqlx/query-6b9d8d54771a850a86787172bbf65804aefcf7d8d4cc9b71a9d70c86ea380d1e.json
+++ b/backend/.sqlx/query-6b9d8d54771a850a86787172bbf65804aefcf7d8d4cc9b71a9d70c86ea380d1e.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT EXISTS(SELECT 1 FROM workspace_env WHERE workspace_id = $1 AND name = $2)",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "exists",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "6b9d8d54771a850a86787172bbf65804aefcf7d8d4cc9b71a9d70c86ea380d1e"
+}
--- a/backend/windmill-api-workspaces/src/workspaces.rs
+++ b/backend/windmill-api-workspaces/src/workspaces.rs
@@ -4677,26 +4677,26 @@ async fn set_environment_variable(
 
     match value {
         Some(value) => {
-            // The name is spliced into the NativeTS/Bun worker prologue. The
-            // worker escapes it regardless, so this write-time check is a friendly
-            // guard against new non-identifier names, not the injection defense.
-            // Enforce it only for names that don't already exist so a value edit of
-            // a grandfathered non-identifier name (the edit UI resubmits the name)
-            // isn't rejected with no in-product way to fix it. Deletion (the None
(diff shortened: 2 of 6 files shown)


hugocasa:
✅ Review round clean @ 86683d258371c40489ef2a1afd5512ec22091b6e

PR #10638, opened 2026-08-11
feat(npm-proxy): keep package files on disk and in the object store

rubenfiszel:
## Summary

Implements the design in `docs/npm-proxy-storage.md` (in this PR, first commit).

The npm proxy kept everything in the API process's heap: ~320 MiB of budgets plus a transient that scaled with request concurrency rather than with any budget. On a 2 GB container that is a large share of the box for content that is public, immutable and re-fetchable.

Package archives now go to disk, and to the instance object store when one is configured, following the tiering `windmill-worker`'s dependency cache already uses: **local disk, then object store on a local miss (which fills disk), then the registry**. Packuments stay in memory at 16 MiB, because they are small, hot, mutable, and exist to save a round trip that a store of their own would reintroduce.

What is stored is the **extracted files, not the archive**: the manifest plus the `.d.ts` files, which is what `ata/index.ts` reads. Caching the tarball as the served unit would still force a full download and inflate to reach one declaration, 50 MB per file for `aws-sdk`.

```
<ROOT_CACHE_DIR>/npm_proxy/<registry hash>/<package>/<version>/
    manifest.json      the file list and the entry point
    files/<path>       each retained .d.ts and package.json
```

## Notable in the implementation

- **Nothing large is ever resident.** Extraction streams each entry straight into the package directory, so a read holds one entry rather than the whole retained set. That removes the transient the budgets could not bound.
- **A warm package costs no round trip.** The registry is resolved from settings; the packument is fetched only on a miss. Fetching it first, to reach the tarball URL, left a fully cached package hitting the registry once a minute.
- **The manifest is exempt from the retention budget.** `next` orders its `package.json` 3748th, so a budget spent on the declarations ahead of it would default the entry point, which is a wrong answer rather than a slow one.
- **Publishing is atomic.** Files land in a sibling temp directory and are renamed into place, so a concurrent reader never sees a half-written package and a killed process leaves nothing that looks complete. Losing the rename race is success: the other writer published the same immutable content.
- **Archive paths are validated.** Entry paths are registry-supplied and now become paths on our disk, so anything but a normal relative component is refused (`../`, absolute, `..` mid-path).
- **Eviction.** An LRU sweep by access time runs at most every 10 minutes after a write, trimming to 2 GiB. An unbounded cache directory on a small container disk would be a worse failure than the memory it replaces.
- **Registry-scoped keys.** The `npmrc` setting can be repointed at a registry serving a different artifact under the same name and version, so the registry URL is hashed into both the disk path and the object key.
- **CE gets the disk tier alone**, since `windmill-object-store` is `parquet`-gated and its call sites also gate on `enterprise`. The object store is a pure addition on top.

## Test plan

Backend running against a token-gated mock registry and MinIO.

- [x] **Disk tier.** Four package shapes correct and unchanged: `aws-sdk@2.1692.0` 2287 files / `lib/aws.js`, `next@16.3.0` 8519 / `./dist/server/next.js`, `lodash@4.17.21` 1054 / `lodash.js`, `@types/express@5.0.6` 4 files. On disk: `aws-sdk` holds 443 files in 50 MB, all of it out of the heap.
- [x] **Cold process, warm disk.** Killed the API (verified the port was released, after a first attempt that silently did not), restarted, then four filetrees and a `.d.ts` read: **0 registry requests**.
- [x] **Object store push.** Wiped disk, one filetree, and `npm_proxy/<hash>/lodash/4.17.21.tar` appears in the bucket.
- [x] **Object store pull.** Wiped disk only, kept the bucket: `@types/express` filetree and its `index.d.ts` served with **0 registry requests**, disk repopulated by the pull (tar extracted in 19 ms).
- [x] **On-demand fallback.** `lodash.js` is neither manifest nor declaration, so it is not retained; the request falls through to a walk of a freshly fetched archive and returns 200.
- [x] `cargo test -p windmill-api-npm-proxy` 10 passed, `cargo check` clean on CE and on `enterprise,parquet,quickjs`.

Not exercised: `--features private`, which fails to build in this worktree on an unrelated `windmill-api-settings` SMTP feature mismatch. The object store path was verified under `enterprise,parquet`, which is what gates it.

### Re-verified after `eb85c6f9f8`

Counted registry hits by parsing the mock registry's JSON log (`len(entries)`); an earlier
`wc -l` over that endpoint returned 0 for everything, since it answers on one line.

- **Cancelled mid-pull.** `timeout 0.03 curl` on a cold `aws-sdk` filetree: the client is cut
  off (`rc=124`), the package still publishes whole, `manifest.json` present, no scratch left
  behind, and `lib/core.d.ts` reads back at 200. Cancelling earlier (`0.01`/`0.02`, still in
  the download) leaves nothing published and nothing behind.
- **Object store pull.** Disk wiped, bucket kept, packument cache expired: all four packages
  served 200 with **0 registry requests**, four package directories restored, no scratch left.
- **Cold fill.** Disk and bucket both wiped: served from the registry and pushed back, four
  tars in the bucket.

### Re-verified after `04473688de`

Reproduced the reported failure on a build with the fix stashed, then again with it applied,
against the same warm bucket and the same directory:

| | pre-fix | post-fix |
|---|---|---|
| `/filetree` after `manifest.json` is removed from a cached package | 500, twice | 200, first try |
| registry requests | 0 (the pull short-circuits) | 0 (repaired from the bucket) |
| destination repaired | no, permanently | yes |

A retained-file read through the repaired tree answers 200, and no scratch directory is left
behind.

### Re-verified after `510d1297c4`

Eviction now detaches with a `rename` and deletes the detached copy, so the live path only
ever holds a whole package or nothing. Re-ran the tiering end to end on the new binary: four
packages served 200 from a wiped disk with **0 registry requests**, four package directories
restored, no scratch left behind, and a retained declaration reads 200.

Still manual-only, unchanged: the `enterprise,parquet` branch has no automated coverage
(push, pull, backpressure, cancellation). The upload-backpressure bound has not been
distinguished by a run against a throttled store, only by reading `WriteMultipart`.

### Re-verified after `519d3bdbd2`

- **One registry snapshot per miss.** Fully cold (disk and bucket wiped): two packages at 200
  on 4 registry requests (packument plus tarball each), both tiers repopulated, `/metadata`,
  `/resolve` and a declaration read all 200.
- **A pulled package reads as recently used.** `tar` restores archived mtimes, so a pulled
  tree carried the time it was first filled: measured 16:02 on every file after a 16:15 pull.
  The manifest, which is what the sweep takes as recency, is now stamped at publication and
  reads 16:15.

Not staged: a registry switch landing between the two resolutions of a single miss, which
needs a pause hook inside the request. That fix is structural (the fetch helper can no
longer resolve the setting at all) rather than covered by a test.


rubenfiszel:
All six addressed in e0cd8466d7. Two of them were mine to own: the design doc in this same PR says disk writes must be best effort, and the code propagated their errors.

**P1 cache-write failures made the proxy unavailable (codex #2, claude).** Correct, and it turned cache pressure into an editor outage. `cached_package` now returns `PackageFiles::Disk | Memory`: when the directory cannot be created, a write fails, or the publish fails, the same walk fills an in-memory map for that request and the endpoint keeps answering. Routing both through one `extract_to` sink means the fallback cannot drift from what would have been stored. Verified by `chmod 500` on the cache root: `/filetree` and `/file` still return 200 with correct content, three `npm proxy cache is not writable, serving without it` warnings logged. Before the fix these were 500s.

**P1 object-store extraction blocked the runtime (codex #3).** `extract_tar` is async in signature and synchronous inside. Pulls now stream to a temp file and unpack on a blocking thread.

**P1 transfers were fully buffered (codex #4).** Pulls collected the whole object into `Bytes` and pushes built the whole tar in a `Vec`, recreating exactly the request-scaled heap this PR removes. Pulls stream to a file; pushes build the tar on disk and upload it with `WriteMultipart` in 8 MiB parts, serialised behind a semaphore since none of it is on a request's path.

**P1 the 2 GiB cap was not enforced (codex #1).** Four separate holes, all real: the object-store pull path returned before the sweep, failed extractions left their directories behind, ten minutes of writes could overshoot, and `measure` counted file lengths. Now: the pull path sweeps, a `Scratch` guard removes its directory on drop, the sweep also removes `.tmp.`/`.upload.` debris it finds, 256 MiB written forces an early sweep, and measurement uses allocated blocks, which matters exactly for the many-tiny-declarations packages.

**P2 doc/code path mismatch (codex #5)** and **P2 narrating test comments (codex #6)** fixed.

**Claude's second finding, the on-demand fallback re-downloading:** accepted as a real cost, not fixed, and now stated in the doc rather than left implicit. It only triggers for a path that is neither manifest nor declaration, which `ata/index.ts` never requests. Making it cheap means either caching archives on disk (the size problem this design exists to avoid) or reintroducing a memory archive cache; both seem worse than a rare slow path. Happy to be argued out of that.

**Something the sweep test caught that I would have shipped wrong:** the LRU was keyed on access time, and cache volumes mount `relatime`, so atime does not move on a read. It was silently a FIFO. It now keys on `mtime` and a cache hit explicitly touches the manifest. The test fails against the old code, evicting the recently used package.

One bug found while re-verifying: the pull created its temp file before the parent directory existed, so on a wiped cache every pull failed and silently fell back to the registry. That is why an earlier run showed a stray registry request.

Re-verified end to end against the mock registry and MinIO: cold fetch populates disk and pushes 3 objects; wiping disk only, `@types/express` and its `index.d.ts` are served with **0 registry requests** and disk is repopulated by the pull; all four package shapes unchanged. `cargo test -p windmill-api-npm-proxy` 11 passed, `cargo check` clean on CE and on `enterprise,parquet,quickjs`.

rubenfiszel:
Addressed in 3f4b5be523. Both reviewers converged on the same thing and they are right: **my previous fix reintroduced the regression it was meant to prevent.** Filling `fallback` on every extraction put the per-request heap back on the happy path, so the degrade fix undid the point of the PR.

The map is now built only when disk has actually failed, by walking the archive a second time. The archive is already in hand, the path is rare, and the alternative (accumulating as you go, in case) is exactly what was wrong. A successful disk fill allocates nothing beyond one entry at a time.

The rest, all real:

- **Concurrent pulls collided on one temp archive.** `scratch.path.with_extension("tar")` replaced the uuid rather than appending, so `<dir>.tmp.<uuid>` became `<dir>.tmp.tar` for every pull. The download and the unpacked tree now both live inside the scratch directory, which makes them per-pull and takes the download with the drop.
- **Failed temp archives bypassed cleanup and eviction.** Same fix: the tar is inside the scratch directory rather than a sibling, so `Scratch`'s drop removes it, and it can no longer sit outside what the sweep walks.
- **Tiny files bypassed the early sweep.** The write accounting charges a 4 KiB block minimum per file, matching what the sweep measures. A package of many tiny declarations summed to almost nothing before and would never have tripped it.
- **`/filetree` raced eviction.** A package swept between the lookup and the manifest read returned a 500. A missing manifest is now a miss to repopulate, bounded to one retry. `/file` already degraded correctly, since a missing file falls through to the on-demand walk.

Verified end to end after the changes:

- Cold: four packages populate disk and push 4 objects, **0 stray temp files** left behind.
- Six concurrent `/filetree` for the same uncached package: all 200, which is the collision the shared temp path would have produced.
- Unwritable cache root: still 200 with correct content.
- Disk wiped, object store intact: `aws-sdk` filetree and a `.d.ts` served with **0 registry requests**.
- All 442 `aws-sdk` declarations at concurrency six: 442×200 in 0.28s.
- `cargo test -p windmill-api-npm-proxy` 11 passed, `cargo check` clean on CE and `enterprise,parquet,quickjs`.

Single flight remains open and is now stated in the doc: concurrent misses for the same package each fetch and extract. Safe, just wasteful.

rubenfiszel:
All seven addressed in the head just pushed.

**P1 oversized manifests lost the entry point (codex #1).** Correct, and it is the third time this file has found a way to default `main`. The manifest now has its own ceiling (`manifest_bytes`, 64 MiB) checked before the entry cap skips it, because that cap is about how much to keep while the manifest decides which answer is right. Test case added: a manifest above `entry_bytes` still yields the declared `main`.

**P1 failed upload archives bypassed cleanup and the cap (codex #2, claude).** The pull was fixed last round and the push was not, which also made `is_debris`'s `.upload.` clause dead code, since nothing creates such a *directory* and `sweep` only walks directories. The upload tar is now built inside a `Scratch`, so both early returns after `spawn_blocking` clean up on drop and the sweep can see it.

**P1 multipart uploads buffered the whole tar (codex #3, claude).** `WriteMultipart::write` queues without waiting, so the 8 MiB read buffer bounded nothing and the doc comment's "uploaded in parts" was not true. `wait_for_capacity(CHUNK)` before each write, matching the two other call sites in the repo.

**`pull_from_object_store` published without checking for a manifest (claude).** The failure mode is worse than it sounds: a truncated object publishes, `read_manifest` returns `None`, the retry re-pulls, `atomic_publish_dir` returns `Ok` because the destination exists, and `Ok(true)` short-circuits the registry, so `/filetree` 500s until eviction. It now checks the unpacked tree and treats a manifest-less object as a miss, which the retry then repairs from the registry.

Three carried-over items, all fixed:
- `escape_segment` did not reject `..`, so `package_dir(reg, "..", "..")` resolved to the registry directory. Dots-only segments are percent-encoded now, with a test.
- `cached_package` hardcoded `manifest.json` while `store.rs` owns `MANIFEST_FILE`; it goes through `store::has_manifest` now, so a rename cannot silently turn the hot-path hit check into a permanent miss.
- `sweep_if_due(0)` after a pull counted nothing, so object-store hits on a wiped cache only trimmed on the clock, never on the 256 MiB trigger that exists for exactly that burst. The pull now reports what it unpacked.

Verified: cold populate of four packages pushes 4 objects with **0 stray temp files**; disk wiped, `aws-sdk` served from the object store with **0 registry requests**; all four shapes unchanged. `cargo test -p windmill-api-npm-proxy` 12 passed, `cargo check` clean on CE and `enterprise,parquet,quickjs`.

rubenfiszel:
All five addressed in af6fa6e4df.

**The multipart backpressure was a no-op (claude).** Confirmed against the vendored source: `wait_for_capacity` polls for fewer than N `UploadPart` *tasks* in flight, so passing `CHUNK` (8388608) never waited and the whole tar was still queued. It takes a part count now (`UPLOAD_PARTS_IN_FLIGHT = 2`), and the constant says what the unit is so the next reader does not repeat it. Good catch: the fix read correctly and did nothing.

**P1 sweeps deleted active scratch directories (codex #1).** Real, and worse than described: a fresh scratch sat at the version level, so it was not only classified as debris, it was also a normal **eviction candidate**. A scratch directory is now never an eviction candidate, and is removed only once an hour stale. My own test caught the eviction half after I had written the fix for the debris half.

**P1 pulls blocked runtime workers (codex #2).** `measure` walks and stats the whole tree and `atomic_publish_dir` can recursively remove the loser's, both scaling with a package's file count. Unpacking, measuring and publishing all run on the blocking thread now.

**P1 manifests bypassed write accounting (codex #3).** `publish` writes the file list and returned nothing, so a package of mostly non-retained entries advanced the sweep trigger by almost nothing while writing tens of MiB. `publish` reports its size and the caller charges it, block-rounded.

**P1 unsafe paths degraded like a broken disk (codex #4).** The sharp part is the consequence: a traversal came back as `BadRequest` from `write_file`, `cached_package` read it as an unwritable cache, and re-extracted into a `HashMap` **without validation**, so the unsafe path was served in the filetree and never cached. Validation moved into `extract_to`, before either sink, and refuses the archive. An archive carrying a traversal is malicious; a full disk is not; they must not share a path.

Worth noting for the test: `tar::Builder` refuses to *write* a `..` entry, so the fixture cannot be built with the safe API. The guard is tested directly instead, which is the honest version rather than a fixture that proves the builder validates.

Verified: four packages fetched **concurrently** from cold all return 200 and read back correctly, 4 objects pushed, **0 stray temp directories**, which is the race the sweep fix targets. `cargo test -p windmill-api-npm-proxy` 13 passed, `cargo check` clean on CE and `enterprise,parquet,quickjs`.

rubenfiszel:
✅ Review round clean @ 519d3bdbd24e2c8463a3c35dadd0945ec156e489 — nit-only verdicts at 28f14c719b178660a9ab26ea03418536e0b25cfd (codex: object-store pulls retain stale LRU timestamps; claude: good to merge, doc missing the one-snapshot rule); both nits addressed in 519d3bdbd24e2c8463a3c35dadd0945ec156e489

PR #10644, opened 2026-08-11
fix: bound postgres result collection so an oversized result cannot OOM the worker

rubenfiszel:
## Summary

Follow-up to #10641, which did the same for DuckDB and has landed — this now sits directly
on `main` and reuses the `MAX_SQL_RESULT_SIZE` machinery introduced there.

A Postgres step that returns a large result set OOM-kills the worker **process**, taking
down every job colocated on it. A guard existed but sat behind `if *CLOUD_HOSTED`, so it
never fired on a self-hosted or dedicated deployment.

Bounding it needed more than flipping that condition: the collection path buffered
everything before anything could check it, and three separate places measured a payload
while the worker paid for a container.

## Changes

- **The guard applies off-cloud**, in both the JSON and raw-output paths, against the
  `MAX_SQL_RESULT_SIZE` cap from #10641. The cloud *limit* is unchanged at
  `MAX_RESULT_SIZE * 4`. Postgres routes through a new `max_sql_result_size()` that picks
  between the two; DuckDB deliberately keeps reading the cap directly, since its limit is
  worker-survival only and not subject to the cloud product cap.
- **Rows are consumed one at a time instead of `try_collect::<Vec<Row>>()`.** The old
  order buffered every row *before* the first check ran, so the cap had nothing left to
  protect. Each row's wire buffer is now dropped as its converted form is appended.
  `build_envelope(Vec<Row>)` becomes `RawOutputEnvelopeBuilder` so the raw path streams
  too.
- **Rows are charged for their containers.** The raw-output counter summed only non-null
  string lengths, so `SELECT NULL FROM generate_series(…)` accumulated
  `Vec<Option<String>>` per row against a total that never moved.
- **Serialization is bounded, not estimated.** JSON escaping expands text — one control
  character becomes the six-byte escape `\u0001` — so a value that fit the cap in memory
  could allocate several times it on the way out, after the check. `to_raw_value_within`
  refuses to allocate past the remaining budget instead of discovering the size
  afterwards. Rows are charged `max(sizeof_val, serialized_len)`, so the in-memory
  estimate still governs wherever it is the stricter of the two.

Because Postgres needs the same machinery on every build, `SQL_RESULT_SIZE_FRACTION`,
`MIN_MAX_SQL_RESULT_SIZE`, `parse_byte_size` and `MAX_SQL_RESULT_SIZE` lose the
`#[cfg(feature = "duckdb")]` gate they were introduced behind.

`format_byte_size` now only uses a unit that divides the limit evenly, falling back to a
plain byte count. The error claims its figure can be set as `MAX_SQL_RESULT_SIZE`
verbatim, and rounding broke that: a 1.5MB limit rendered as `1MB`, which both misreports
it and lowers it if pasted back. The memory-derived default is rarely a whole number of
MB, so that was the common case. (The DuckDB copy of this helper in the FFI crate still
rounds — same fix, but it belongs in a DuckDB-scoped change.)

### Serialization cost

Bounding the writer means the bytes come back as a `String` rather than straight from
`serde_json::to_string`, and the safe `RawValue::from_string` **re-parses every row** to
validate what serde_json just wrote. `serde_json::value::to_raw_value` — what the old code
called — skips that by construction. Measured per row (1.0.151, `preserve_order`, release):

| per row | `to_raw_value` (old) | via `RawValue::from_string` | via `from_string_unchecked` |
|---|---|---|---|
| wide 20-col row | 606 ns | 1104 ns (+82%) | 622 ns (+3%) |
| 1KB text row | 297 ns | 519 ns (+75%) | 281 ns (−5%) |

The UTF-8 check accounts for ~2% of that; the re-parse is the whole regression. So the
`String::from_utf8` stays safe and only the `RawValue` construction is unchecked, leaving
one invariant to justify: `to_writer` returned `Ok`, so the buffer holds one complete,
well-formed JSON value. serde_json `debug_assert!`s exactly that by re-parsing in debug
builds, so tests keep verifying what release trusts.

**This also fixes DuckDB.** The FFI crate carries its own copy of `to_raw_value_within`
and pays the same per-row re-parse on `main` today; #10641 accepted it on the stated
grounds that "the unchecked constructor is not public", which was not the case —
`RawValue::from_string_unchecked` was added in serde_json 1.0.151, already the pinned
version at that commit. Both copies are fixed here, and both crates' `serde_json`
requirement moves from `^1` to `^1.0.151` to match what the code now needs.

### Measured

| scenario | before | after |
|---|---|---|
| ~600MB result, 200MB cap | all rows buffered before the first check | refused, peak 258 MiB (baseline 124) |
| 19MB cell of `U+0001`, 20MB cap | **succeeded**, peak **995 MiB** | refused, peak 200 MiB |
| same via `-- raw_output` | unbounded | refused, peak 226 MiB |
| 10M all-NULL raw_output rows, 20MB cap | counter stayed at 0 | refused, peak 172 MiB |

## Notes for reviewers

- **Behavior change for self-hosted**: queries that returned very large results and
  happened to survive will now fail. The cap scales with worker memory and is
  overridable, so it should only bite jobs already close to killing their worker. A
  worker with no cgroup reading to scale from is still uncapped, as before.
- **Cloud's limit is unchanged, but what a row is charged against it is not.** Counting
  containers costs ~24B per cell plus ~24B per row, and the JSON path now charges
  `max(sizeof_val, serialized_len)`. A wide row of short cells therefore weighs several
  times what it used to against the same 8MB cap, so a cloud `-- raw_output` job on a
  100-column table could start failing where it previously passed. That is the same
  correction as the all-NULL fix — those rows were free before — but it is a real change
  in reach, not only off-cloud.
- The cap bounds what is *collected*, not the process. `-- raw_output` holds the
  accumulated text and then serializes it, so that path peaks near 2x the cap; the
  derived default is a fraction of the worker's budget to leave room for that.
- The envelope is budgeted against the whole cap rather than the remainder, because its
  rows were already charged as text and double-charging would reject results that passed
  every row-level check.

## Test plan

- [x] `cargo test -p windmill-worker --lib` — 200 pass
- [x] Connection reuse after a refusal: an oversized query aborts the row stream
      mid-result, and the next small query on the same resource and worker returns
      correct rows over the same cached connection (JSON and `-- raw_output` both)
- [x] `cargo check -p windmill-worker --features duckdb` — clean (the shared machinery is
      no longer feature-gated, so both builds are checked)
- [x] Regression tests: null/empty cells still count toward the cap, escaping cannot
      outgrow the budget, every rendered limit round-trips to the exact byte count
- [x] **Real DuckDB jobs** through the changed FFI path, against a release-profile dylib
      (so serde_json's `debug_assert!` is compiled out and the JSON validity is genuinely
      trusted rather than re-checked): nested `LIST`/`STRUCT`/NULL rows, a row of control
      characters + backslash/tab + `é☃`, and a 5MB result refused under a 1MB cap. Every
      result is stored into a `jsonb` column, so Postgres re-parsed and accepted each one
      — independent confirmation that the unchecked `RawValue` is well-formed. Worker
      healthy throughout, zero panics.
- [x] Real Postgres jobs on a dev backend. Under `MAX_SQL_RESULT_SIZE=1MB`: small,
      empty, and `-- raw_output` small results succeed; 5MB JSON, 5MB `-- raw_output`,
      200k all-NULL `-- raw_output`, and a 900KB `chr(1)` cell (whose escaped form is
      ~5MB) are all refused with the new error, and a deliberately non-round
      `MAX_SQL_RESULT_SIZE=1500000` is quoted as `1500000B` rather than a rounded `1MB`.
      **The worker survived every refusal and kept serving jobs**, zero panics.
- [x] The memory-derived default, exercised by running the backend under a 3 GiB cgroup
      (`systemd-run --user --scope -p MemoryMax=3G`): the cap derives to 483183820B and a
      600MB result is refused quoting exactly that, with the worker surviving and
      returning to baseline. This is the case the rendering fix matters most for — the old
      renderer would have called that limit `460MB`, which is 838,860 bytes below it.
      Note an unconstrained host (no cgroup limit, `memory.max = max`) derives no cap at
      all and is unaffected by this PR, which is why the cgroup was needed to test it.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -433,7 +433,9 @@ tower = "^0"
 tower-http = { version = "^0.6", features = ["trace", "cors", "catch-panic"] }
 tower-cookies = "^0.11"
 serde = "^1"
-serde_json = { version = "^1", features = ["preserve_order", "raw_value"] }
+# 1.0.151 introduced RawValue::from_string_unchecked, which the SQL executors use
+# to avoid re-parsing every collected row.
+serde_json = { version = "^1.0.151", features = ["preserve_order", "raw_value"] }
 serde_yml = "0.0.12"
 uuid = { version = "^1", features = ["serde", "v4", "js"] }
 thiserror = "^2"
--- a/backend/windmill-duckdb-ffi-internal/Cargo.toml
+++ b/backend/windmill-duckdb-ffi-internal/Cargo.toml
@@ -21,7 +21,9 @@ duckdb = { git = "https://github.com/windmill-labs/duckdb-rs", rev = "7190adfcf5
 regex = "1"
 rust_decimal = "1.37.2"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "^1", features = ["preserve_order", "raw_value"] }
+# 1.0.151 introduced RawValue::from_string_unchecked, which row collection uses
+# to avoid re-parsing every row.
+serde_json = { version = "^1.0.151", features = ["preserve_order", "raw_value"] }
 
 [lib]
 crate-type = ["cdylib"]
--- a/backend/windmill-duckdb-ffi-internal/src/lib.rs
+++ b/backend/windmill-duckdb-ffi-internal/src/lib.rs
@@ -910,10 +910,6 @@ fn interpolate_named_args<'a>(
 /// for any reason except the budget. A caller passing a type with a fallible
 /// `Serialize` would have its error silently retold as a size limit.
 ///
-/// The bytes come back through `RawValue::from_string`, which re-parses to
-/// validate what `serde_json` just wrote. `to_raw_value` skipped that, so this
-/// costs one extra pass per row — the price of getting the bytes through a writer
-/// we can bound, since the unchecked constructor is not public.
 fn to_raw_value_within<T: serde::Serialize>(value: &T, budget: usize) -> Option<Box<RawValue>> {
     struct Budgeted {
         buf: Vec<u8>,
@@ -921,6 +917,12 @@ fn to_raw_value_within<T: serde::Serialize>(value: &T, budget: usize) -> Option<
     }
     impl std::io::Write for Budgeted {
         fn write(&mut self, bytes: &[u8]) -> std::io::Result<usize> {
+            self.write_all(bytes)?;
+            Ok(bytes.len())
+        }
+        // `Vec<u8>` overrides this too: the default implementation loops over
+        // `write`, and serde_json emits a great many small pieces per row.
+        fn write_all(&mut self, bytes: &[u8]) -> std::io::Result<()> {
             if bytes.len() > self.left {
                 return Err(std::io::Error::new(
                     std::io::ErrorKind::WriteZero,
@@ -929,7 +931,7 @@ fn to_raw_value_within<T: serde::Serialize>(value: &T, budget: usize) -> Option<
             }
             self.left -= bytes.len();
             self.buf.extend_from_slice(bytes);
-            Ok(bytes.len())
+            Ok(())
         }
         fn flush(&mut self) -> std::io::Result<()> {
             Ok(())
@@ -938,9 +940,15 @@ fn to_raw_value_within<T: serde::Serialize>(value: &T, budget: usize) -> Option<
 
     let mut writer = Budgeted { buf: Vec::new(), left: budget };
     serde_json::to_writer(&mut writer, value).ok()?;
-    String::from_utf8(writer.buf)
-        .ok()
-        .and_then(|s| RawValue::from_string(s).ok())
+    let json = String::from_utf8(writer.buf).ok()?;
+    // SAFETY: `to_writer` returned `Ok`, so `json` holds one complete, well-formed
+    // JSON value with no surrounding whitespace. Running out of budget is the only
+    // way a partial write happens, and it takes the `?` above instead of reaching
+    // here. The safe constructor re-parses every row to learn the same thing, which
+    // measured ~1.8x the cost of serializing it in the first place; serde_json
+    // itself builds a `RawValue` this way in `to_raw_value`, and `debug_assert!`s
+    // the invariant by re-parsing in debug builds.
+    Some(unsafe { RawValue::from_string_unchecked(json) })
 }
 
 /// How much a row may still expand to, carried through every value it contains.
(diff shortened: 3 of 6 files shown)


rubenfiszel:
✅ Review round clean @ 3407d9f1e2b494e21d8640e6cfc453fae72f3e2c — nit-only verdicts at ea52a395b55e12e9f45f65b6e6e94bb9772dca06 (Codex: Good to merge; Claude: one P2 on a stale `parse_byte_size` doc rationale); nit addressed in 3407d9f1e2, which also reverts an unrelated rustfmt reflow of `log_context_for_job` that the format hook had reintroduced.

Also ran the extra check Claude suggested: the memory-derived default exercised for real, by running the backend under `systemd-run --user --scope -p MemoryMax=3G`. The cap derives to `483183820B`, a 600MB result is refused quoting exactly that, and the worker survives and returns to baseline. That is the case the rendering fix matters most for — the old renderer called that same limit `460MB`, 838,860 bytes below it. Worth noting for anyone reading the test plan: a host with no cgroup memory limit (`memory.max = max`) derives no cap at all, so the guard is inert there; the cgroup is what made this testable.

rubenfiszel:
Addressing the Codex P2 (unexercised DuckDB FFI result path) — fair catch, the DuckDB evidence was `cargo check`/unit tests only while the real-job runs were all Postgres. No code change; the gap was verification, so here it is.

Built the FFI dylib (`build_dev.sh`, **release** profile — so serde_json's `debug_assert!` is compiled out and the unchecked `RawValue` is genuinely trusted rather than re-validated) and ran the backend with `--features quickjs,duckdb`:

| DuckDB job | result |
|---|---|
| `SELECT 1, 'plain', NULL, [1,2,3], {'k':'v'}` | ✅ `[{"n":1,"s":"plain","lst":[1,2,3],"nul":null,"strct":{"k":"v"}}]` — nested LIST/STRUCT and NULL all intact |
| control chars + backslash/tab + `é☃` | ✅ `\u0001\u0002` escaped, backslash/tab escaped, UTF-8 preserved |
| `repeat('x',1000) FROM range(5000)` under `MAX_SQL_RESULT_SIZE=1MB` | ✅ refused, quoting the 1MB limit with the COPY hint |

Worker healthy throughout, zero panics.

Worth noting the strongest part of that evidence is not my inspection: `v2_job_completed.result` is a **`jsonb`** column, so Postgres itself parsed and accepted every payload on insert. A malformed `RawValue` from the unchecked construction would have failed the insert rather than round-tripping. That is independent, machine-checked confirmation of the `unsafe` invariant on a build where nothing else is checking it.

✅ Review round clean @ 10a5c9863335d33f2c0560a206dd5d85855383fa — verdicts at this same head: Claude "Good to merge", Codex "Mergeable, but should ideally address nits" with the single P2 above, now addressed by verification (no commit, head unchanged).

PR #10645, opened 2026-08-11
feat: bound how much disk a single duckdb job can spill

rubenfiszel:
## Summary

DuckDB runs in-process in the worker, and since #10607 spills normally on an isolating
worker: the patched engine's `lock_temp_directory` freezes `temp_directory` and in exchange
exempts the buffer manager's spill files from `SET disabled_filesystems='LocalFileSystem'`.

Nothing bounded how much they spill. One query could fill the worker's disk, taking every
job colocated on it down with it. DuckDB has `max_temp_directory_size` for exactly this, but
setting it alone is a guardrail against accidents rather than a bound: `SET
max_temp_directory_size='100GB'` was accepted from any script statement, so a query could
lift the cap and carry on. A cap a script can raise is not a cap.

So this does both halves: the FFI sets the cap, and the engine patch freezes it under the
same lock that already freezes the directory. Unset means unchanged behaviour, so
self-hosted installs are unaffected unless an operator opts in. The cloud sets `6GiB` per
worker pod (windmill-labs/infra-aws#60, merged).

## Changes

- **`configure_duckdb_resource_limits`** emits `SET max_temp_directory_size` from
  `DUCKDB_MAX_TEMP_DIRECTORY_SIZE`, immediately before it takes the lock. The value is read
  from the environment rather than added to the FFI signature: the cdylib is versioned
  separately and agent workers ship it out of band, so a signature change would cost an ABI
  bump and a coordinated redeploy for what is a config value. It is carried on
  `ResourceLimits` so the SQL-emitting function stays a pure function of its input and the
  env is read once, at the FFI boundary.
- **Engine patch**: repins the `duckdb-rs` fork to `7190adf`, which adds
  `windmill/0002-lock-the-temp-directory-size-cap.patch` — `lock_temp_directory` now makes
  `max_temp_directory_size` immutable too, `SET` and `RESET` alike.
- **Tests**: `spilling_past_the_configured_cap_fails_the_query_and_cleans_up` pins that a
  configured connection stops at its cap, names the setting when it does, and leaves no
  spill behind (with a generous-cap control, since DuckDB compresses temp files and the
  failure is only evidence of the cap if the same query gets through without it).
  `locking_the_temp_directory_makes_it_immutable` gains the two cap statements.
- **Error messages**, added in review: DuckDB's cap error ends by suggesting `PRAGMA
  max_temp_directory_size='10GiB'`, which the lock now refuses — advice that was actionable
  before this PR and is now a dead end. `decode_ffi_error` appends a correction. The engine
  emits the same text for its *own* default cap, so `spill_cap_hint` takes the attribution
  from the worker's config: with the var set it names it, unset it says the worker is low on
  disk rather than blaming a variable nobody set. The cap's `SET` is also issued on its own
  statement, so an unparseable value names the env var instead of producing wording
  indistinguishable from a bad `DUCKDB_MEMORY_LIMIT`.
- **`docs/duckdb-isolation.md`**: the new half of the lock, the env var, the cloud default
  and its sizing, and the units gotcha.

### On the upstream patch

`0001` is filed as [duckdb/duckdb#24694](https://github.com/duckdb/duckdb/pull/24694) and is
under review. `0002` is kept as a **separate commit rather than an amendment** so `0001`
stays byte-identical to what that PR carries, and `0002` applies to it verbatim as a second
commit whenever a human chooses to push it. It is **windmill-only for now** — filing
upstream was a human decision on purpose (duckdb's `CONTRIBUTING.md` asks contributors not
to submit LLM-generated pull requests), so this is a deliberate, declared divergence rather
than drift. Both `docs/duckdb-isolation.md` and the fork's `windmill/README.md` say so and
say what to do when it is pushed.

### Sizing note for the cloud

The cloud worker pods carry **no `ephemeral-storage` limit**, so what the cap must stay
clear of is the node's eviction threshold rather than a per-pod one. They run a single worker
each (`NUM_WORKERS` unset, default 1), so no division is needed.

I initially sized this against node capacity alone and got it wrong; measured on the live
cluster instead:

| node | nodefs | used at rest | cloud worker pods | their ephemeral-storage |
|---|---|---|---|---|
| `ip-10-2-15-45` | 99.9G | 45.3G (55% free) | 6 | 33.4G |
| `ip-10-2-20-115` | 99.9G | 59.1G (41% free) | 5 | 27.6G |
| `ip-10-2-27-31` | 99.9G | 56.8G (43% free) | 3 | 14.6G |

A worker pod already holds **4.7–6.9 GiB** of language-runtime caches and job dirs before any
spill. So the budget is not "6GiB against ~89GiB free" — on the fullest node above, all five
pods spilling to the cap simultaneously would land at roughly 11% free, i.e. near the
kubelet's default `nodefs.available<10%` eviction threshold. That is the tail rather than the
common case (it needs five concurrent max-spill DuckDB jobs on one node, each in a 2Gi pod),
and it is still strictly better than today, where a single query can take all the free space
by itself. But 6GiB is sized against a mostly-full node, not an empty one — worth knowing
before raising it, and worth re-measuring if the runtime-cache baseline grows.

`6GiB`, not Kubernetes' `6Gi` — DuckDB accepts `KiB`/`MiB`/`GiB`/`TiB` and `KB`/`MB`/`GB`/`TB`
and rejects anything else outright.

## Test plan

Layout invariant (`docs/duckdb-isolation.md`: no pre-existing struct member may change offset):

- [x] `clang++ -Xclang -fdump-record-layouts` diff of `0001` vs `0001+0002` — **byte-identical**;
      this patch adds no struct member.
- [x] vs stock v1.5.5 — only the two *added* lines from `0001` and its `sizeof` growth; no
      member moved.
- [x] `prebuilt_extensions_still_load` passes (`INSTALL httpfs; LOAD httpfs;` returns, no hang).

DuckDB built from source with both patches applied:

- [x] `test/sql/settings/*` — 388 assertions, all pass (includes the extended
      `lock_temp_directory.test`).
- [x] `test/sql/storage/temp_directory/*` — 86 assertions, all pass (upstream's own
      `max_swap_space` suite).
- [x] `Test RESET statement for ClientConfig options` — 226 assertions, all pass.

FFI crate: `cargo test --release -p windmill_duckdb_ffi_internal` — 21/21 pass.

Real preview jobs on a worker built with `--features duckdb,private,parquet,quickjs` and
started with `ENABLE_UNSHARE_PID=true` (isolation transform active), `DUCKDB_MEMORY_LIMIT=32MB`,
`DUCKDB_MAX_TEMP_DIRECTORY_SIZE=20MB`:

- [x] settings read back as `cap=19.0 MiB, locked=true, tmpdir=<job dir>` — the env var
      reaches the connection.
- [x] under the cap: query spilled **15,204,352 bytes** (`duckdb_memory()`) and returned the
      right answer.
- [x] over the cap: `Out of Memory Error: failed to offload data block of size 256.0 KiB
      (18.9 MiB/19.0 MiB used). This limit was set by the 'max_temp_directory_size' setting.`
      — job dir left at **0 bytes**.
- [x] `SET max_temp_directory_size='100GB'` → `Permission Error: Modifying
      'max_temp_directory_size' is disabled - the temp directory has been locked`.
- [x] `SET temp_directory='/tmp/…'` → same refusal; `read_text('/etc/hostname')` still fenced.
- [x] worker log clean (no panic / stack overflow / SIGBUS) and the worker alive after every
      run, including the failures.
- [x] `DUCKDB_MAX_TEMP_DIRECTORY_SIZE=6GiB` (the cloud value) reads back as `6.0 GiB`.
- [x] env unset → `90% of available disk space`, DuckDB's default: self-hosted unaffected.
- [x] the over-cap error now carries the correction (`the PRAGMA suggested above is refused`).
- [x] `DUCKDB_MAX_TEMP_DIRECTORY_SIZE=6Gi` (the k8s spelling) fails closed with `Error applying
      DUCKDB_MAX_TEMP_DIRECTORY_SIZE='6Gi': Parser Error: Unknown unit for memory: 'gi'`.

No frontend surface.

---
## Summary by cubic
Bound DuckDB temp spill per job to prevent a single query from filling a worker’s disk. The cap comes from `DUCKDB_MAX_TEMP_DIRECTORY_SIZE`, is locked, and over-cap errors include a clear hint; cloud default is 6GiB, self‑hosted unchanged unless configured.

- **New Features**
  - Set `max_temp_directory_size` from `DUCKDB_MAX_TEMP_DIRECTORY_SIZE`, then lock it with `temp_directory`; `SET`/`RESET` are blocked.
  - Over‑cap queries fail cleanly and remove temp files; the worker appends a hint and corrects DuckDB’s PRAGMA advice. Hint attributes the limit to the env var when set, or to DuckDB’s default (low disk) when unset.
  - Spill files remain usable behind `disabled_filesystems` after the lock.
  - Repinned to `duckdb-rs` `7190adf` with two engine patches; no FFI signature change.
  - Docs: added cloud sizing guidance based on real pod disk usage and unit notes; cloud default 6GiB, self‑hosted unchanged unless configured.

- **Bug Fixes**
  - Do not blame `DUCKDB_MAX_TEMP_DIRECTORY_SIZE` when unset; the hint now points to DuckDB’s default and low disk.
  - Apply the cap in its own `SET` so parse errors clearly name `DUCKDB_MAX_TEMP_DIRECTORY_SIZE`.

<sup>Written for commit b47084928f476a73b1260317610c35b967648369. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10645?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-duckdb-ffi-internal/Cargo.lock
+++ b/backend/windmill-duckdb-ffi-internal/Cargo.lock
@@ -472,7 +472,7 @@ dependencies = [
 [[package]]
 name = "duckdb"
 version = "1.10505.0"
-source = "git+https://github.com/windmill-labs/duckdb-rs?rev=475383b71cde3d009ca6eaa0cd733faa52bc7393#475383b71cde3d009ca6eaa0cd733faa52bc7393"
+source = "git+https://github.com/windmill-labs/duckdb-rs?rev=7190adfcf58624e7e9a7a2a1000bac18fed64c19#7190adfcf58624e7e9a7a2a1000bac18fed64c19"
 dependencies = [
  "arrow",
  "cast",
@@ -814,7 +814,7 @@ checksum = "3eaf3ede3fee6db1a4c2ee091bf8a8b4dccdc6d17f656fb07896ee72867612f2"
 [[package]]
 name = "libduckdb-sys"
 version = "1.10505.0"
-source = "git+https://github.com/windmill-labs/duckdb-rs?rev=475383b71cde3d009ca6eaa0cd733faa52bc7393#475383b71cde3d009ca6eaa0cd733faa52bc7393"
+source = "git+https://github.com/windmill-labs/duckdb-rs?rev=7190adfcf58624e7e9a7a2a1000bac18fed64c19#7190adfcf58624e7e9a7a2a1000bac18fed64c19"
 dependencies = [
  "cc",
  "diffy",
--- a/backend/windmill-duckdb-ffi-internal/Cargo.toml
+++ b/backend/windmill-duckdb-ffi-internal/Cargo.toml
@@ -11,10 +11,10 @@ chrono = "0.4.41"
 # identical DECIMAL. Building one by hand would mean reimplementing that
 # f64 → (mantissa, scale) inference.
 #
-# Not crates.io: a fork of duckdb-rs 1.10505.0 carrying one patch to the bundled engine —
+# Not crates.io: a fork of duckdb-rs 1.10505.0 carrying two patches to the bundled engine —
 # see docs/duckdb-isolation.md for why, and for the exit path back to the crates.io crate.
 # Pinned by rev, not branch, so `cargo update` cannot move the engine under us.
-duckdb = { git = "https://github.com/windmill-labs/duckdb-rs", rev = "475383b71cde3d009ca6eaa0cd733faa52bc7393", features = [
+duckdb = { git = "https://github.com/windmill-labs/duckdb-rs", rev = "7190adfcf58624e7e9a7a2a1000bac18fed64c19", features = [
     "bundled",
     "rust_decimal",
 ] }
--- a/backend/windmill-duckdb-ffi-internal/src/lib.rs
+++ b/backend/windmill-duckdb-ffi-internal/src/lib.rs
@@ -68,7 +68,7 @@ pub extern "C" fn run_duckdb_ffi(
     collect_first_row_only: bool,
 ) -> *mut c_char {
     let resource_limits = match (ptr_to_opt_str(memory_limit), ptr_to_opt_str(temp_directory)) {
-        (Ok(m), Ok(t)) => Ok(ResourceLimits { memory_limit: m, temp_directory: t }),
+        (Ok(m), Ok(t)) => Ok(ResourceLimits::from_ffi(m, t)),
         (Err(e), _) | (_, Err(e)) => Err(e),
     };
     let (r, column_order) = match convert_args(
@@ -195,7 +195,7 @@ pub extern "C" fn prepare_duckdb_ffi(
     temp_directory: *const c_char,
 ) -> *mut c_char {
     let resource_limits = match (ptr_to_opt_str(memory_limit), ptr_to_opt_str(temp_directory)) {
-        (Ok(m), Ok(t)) => Ok(ResourceLimits { memory_limit: m, temp_directory: t }),
+        (Ok(m), Ok(t)) => Ok(ResourceLimits::from_ffi(m, t)),
         (Err(e), _) | (_, Err(e)) => Err(e),
     };
     let r = match convert_prepare_args(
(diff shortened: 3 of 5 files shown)


rubenfiszel:
✅ Review round clean @ b6f4a19304504a93ea466fe674d5110932bb5a4b — nit-only verdicts at ea3aea9118ca961071749a2c2ba86a417350c695 (claude: Good to merge; codex: nits only); nits addressed in b6f4a19304504a93ea466fe674d5110932bb5a4b

PR #10651, opened 2026-08-12
fix(frontend): skip reserved ids when auto-assigning flow module ids

### LINKED ISSUE #10610: bug: Auto assigning module ids past 'ar' breaks the flow
### Describe the bug

The automatic id assignment for modules works correctly only up to 'ar' . The next module created receives the id of 'as', and so does every module created after it. If these modules are normal flow modules this results in "error parsing the flow". If these modules are the ai agent tools no warnings or errors are displayed, but the nodes are invisible and cannot be edited in the visual editor.

### To reproduce

1. Create a flow
2. Add a node and set its id to 'ar'
3. Add one more node
4. Observe the new node receives the forbidden id 'as'
5. Add third node
6. Observe "error parsing the flow" in the visual editor. Observe the third node having duplicate 'as' as its id in the YAML editor.

### Expected behavior

The automatic id assignment does not make the flow uneditable or broken.

### Screenshots

_No response_

### Browser information

_No response_

### Application version

v1.783.0

### Additional Context

This is likely due to how ids are handled in flowModuleNextId.ts between nextId() and autoIdNumber() . The former seem to generate prevous id + 1 with no regard to forbidden ids, then the latter returns 'undefined' as it encounters the forbidden value, causing the id to never increment.
'as' is just the first of forbiddenIds alphabeticaly, probably would break the same way with other values from the list.

rubenfiszel:
## Summary

Fixes #10610.

`nextId` returned `numberToChars(max)` without checking whether that id is reserved. Reserved ids never raise `max` — `autoIdNumber` returns `undefined` for them — so once the highest auto id reaches `ar`, its successor `as` gets assigned anyway, and then handed out again on every later call. Two modules with id `as` breaks the flow with "There are duplicate modules in the flow at id: as", and silently hides AI agent tool nodes so they can't be edited.

This is not an exotic edge case: `as` is the 45th auto-generated id, so a flow (or an agent's tool list, which shares the same id space) that grows past 45 steps hits it without anyone renaming anything. The other reserved ids the base-26 sequence can produce sit at the same kind of distance — `bg` (59th), `do` (119th), `if` (240th), `in` (248th), then `ctx`, `for`, `new`.

The fix skips forward past any reserved or already-taken id before returning, so the sequence jumps `ar` → `at`. `getNextId` in `idUtils.ts` already treats `forbiddenIds` as taken; this brings `nextId` in line with it.

Flows already poisoned by this keep their duplicate `as` and still need the id fixed by hand in the YAML editor — nothing repairs them on load.

Diagnosis and original patch by @eeshsaxena in #10650; reopened here because the CLA check blocks that branch.

## Changes

- `nextId` skips reserved and already-taken ids before returning one.
- Regression test pinning the successor of `ar` to `at`.

## Test plan

- [x] `vitest run src/lib/components/flows/flowModuleNextId.test.ts` — 6/6 pass.
- [x] Browser repro on a dev server: add a step, set its id to `ar`, add two more steps. On `main` the third add breaks the editor with "There are duplicate modules in the flow at id: as"; with the fix the ids run `ar`, `at`, `au` and the graph renders normally.

## Screenshots

**Before** (on `main`) — the third step re-uses `as` and the editor dies:

before

**After** — `as` is skipped, ids run `ar`, `at`, `au`:

after

---
## Summary by cubic
Prevents duplicate flow module IDs by skipping reserved and already-used IDs when auto-assigning. This avoids editor crashes and hidden agent tool nodes; the sequence now goes `ar` → `at` (skips `as`) and continues normally.

- **Bug Fixes**
  - `nextId` builds a set of taken IDs from the graph and `flowState`, then skips reserved and used IDs until it finds a valid one.
  - Added a regression test asserting `ar` → `at`.

- **Migration**
  - Existing flows with duplicate `as` need a manual fix in the YAML editor; no auto-repair.

<sup>Written for commit 95d8552e97914747d825e0d5ccc2432ff98d0423. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10651?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/lib/components/flows/flowModuleNextId.test.ts
+++ b/frontend/src/lib/components/flows/flowModuleNextId.test.ts
@@ -51,8 +51,7 @@ describe('nextId', () => {
 		expect(nextId(state, flowWith(ids))).toBe('c')
 	})
 
-	// A reserved id never raises the max, so assigning it ("as", the successor of "ar")
-	// would hand it out again on every later call and leave the flow with duplicate ids.
+	// "as", the successor of "ar", is the first reserved id the sequence reaches.
 	it('skips a reserved id instead of assigning it', () => {
 		expect(nextId(stateWith(['failure']), flowWith(['ar']))).toBe('at')
 	})
--- a/frontend/src/lib/components/flows/flowModuleNextId.test.ts
+++ b/frontend/src/lib/components/flows/flowModuleNextId.test.ts
@@ -50,4 +50,10 @@ describe('nextId', () => {
 		const state = stateWith([...ids, 'process', 'my_step', 'failure'])
 		expect(nextId(state, flowWith(ids))).toBe('c')
 	})
+
+	// A reserved id never raises the max, so assigning it ("as", the successor of "ar")
+	// would hand it out again on every later call and leave the flow with duplicate ids.
+	it('skips a reserved id instead of assigning it', () => {
+		expect(nextId(stateWith(['failure']), flowWith(['ar']))).toBe('at')
+	})
 })
--- a/frontend/src/lib/components/flows/flowModuleNextId.ts
+++ b/frontend/src/lib/components/flows/flowModuleNextId.ts
@@ -25,13 +25,23 @@ function autoIdNumber(key: string): number | undefined {
 
 // Computes the next available id
 export function nextId(flowState: FlowState, fullFlow: OpenFlow): string {
-	const allIds = dfs(fullFlow.value.modules, (fm) => fm.id)
+	const allKeys = dfs(fullFlow.value.modules, (fm) => fm.id).concat(Object.keys(flowState))
+	const takenIds = new Set(allKeys)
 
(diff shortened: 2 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 95d8552e97914747d825e0d5ccc2432ff98d0423 — Codex and Claude both returned "Good to merge" at 9bd4a12ca6a8dcf77ce3961e7abd622ecd75ed65; the only follow-up is 95d8552e97, a comment-only nit from the local Codex pass (the test restated an invariant already documented beside the implementation).

PR #10656, opened 2026-08-12
feat: let the global AI chat call connected MCP servers as the user

hugocasa:
## Summary

The global AI chat could only call Windmill's own API. This lets it act on external systems through MCP servers the user has connected — GitHub first — **under that user's own credentials**, not a shared bot identity.

Most of the machinery already existed: a real MCP client (`windmill-mcp`, rmcp/streamable-HTTP, SSRF-validated), an `mcp` resource type whose `token` is a `$var:` pointer, and `GET /w/{ws}/resources/mcp_tools/{path}`. Two gaps: nothing could *call* a tool over HTTP (execution lived only in the worker's AI-agent step), and the chat's tool registry had no way to read a resource.

"As the user" comes from path scoping plus the backend resolving the token under the caller's RLS, so the chat can only reach what the user can.

### Why not the MCP OAuth connect flow for GitHub

GitHub's remote MCP server publishes `/.well-known/oauth-protected-resource/mcp` pointing at `https://github.com/login/oauth`, but that authorization server advertises **no `registration_endpoint`**, so Dynamic Client Registration is unavailable. Windmill then falls back to CIMD, sending `<BASE_URL>/api/mcp/oauth/client-metadata.json` as the `client_id` — GitHub responds `404` to that, because it resolves the id against registered apps rather than fetching the document.

Even if GitHub did support CIMD, the fallback would still be unusable for most instances: it requires GitHub to fetch a metadata document from `BASE_URL`, which a localhost or firewalled Windmill cannot serve. So the GitHub login goes through Windmill's existing instance-configured `github` OAuth connect (or a PAT) instead. That also means this ships in CE: the `mcp` cargo feature is in `oss_core` and needs neither `private` nor `enterprise`.

## Changes

**Backend**
- Extract `connect_mcp_client` in `mcp_tools.rs` (scope check → resource read under the caller's RLS → OAuth refresh → token via `get_value_internal` → connect), shared by the list handler and the new one.
- Add `POST /w/{ws}/resources/mcp_call_tool/{path}`, scoped `resources:write:{path}` (the auth middleware maps POST → Write before the handler runs, so a read scope would be strictly weaker than the middleware). The MCP result is forwarded verbatim, including a tool that ran but reported `isError`.
- Fix the `getMcpTools` response schema in `openapi.yaml`: it declared `parameters`, but rmcp serializes `name` / `description` / `inputSchema` / `annotations`. Neither operation is marked `x-mcp-tool`, so Windmill's own MCP catalog is unchanged.

**Chat**
- New `global/mcpTools.ts`: `search_mcp_tools`, `call_mcp_read_tool`, `call_mcp_write_tool`, modeled on the existing `apiCatalogTools.ts` facade. Search returns name/description/param-names only; a tool's full input schema enters context only after a call fails; results are capped at 20k chars and MCP text content is flattened.
- The read/write split keys off `annotations.readOnlyHint`. A **missing** annotation is treated as mutating, so an unannotated tool can never skip the confirmation. Both call tools are built from one body so the two sides of that check cannot drift.
- A connected server stays **off until it is turned on** in the Connections drawer (`mcp/enabledServers.ts`, stored per browser, keyed by workspace **and account** so a shared browser never hands the next person tools the previous one enabled). This is a scoping choice, not a security control — browser storage is editable by whoever holds the browser. What a server can reach is bounded by the backend: the token resolves under the caller's RLS, so the chat only reaches what that user already could. Being able to read an `mcp` resource in a shared folder is not the same as wanting the chat to act through it, and each enabled server puts its tool descriptions in the model's context. Connecting one through the drawer turns it on, since connecting is the act of choosing it.
- Tools are registered per session in `AIChatManager.configureGlobalMode` **only when at least one server is enabled**, so a workspace without one pays no per-iteration schema cost. Same for the system-prompt block.
- One unreachable server no longer blanks out search for the others; it is reported under `unavailable`.

**Connecting a server (shared by the chat and AI agent steps)**
- New `mcp/McpConnect.svelte`: one connect UI used by **both** the chat drawer and the AI agent step, so the two surfaces no longer have different flows for the same resource type. It asks for a server first and only then for a credential: enter a url (or pick a suggestion), let discovery say what the server supports, then either sign in or paste a token. Never both at once. It is always open in both surfaces, since connecting is what they are for; a connection remounts it rather than collapsing it.
- `mcp/registry.ts` holds a short list of known servers (GitHub, Notion, Linear, Sentry, Stripe) shown as compact suggestions beside the url field, each with its provider icon and a note on what it accepts as a static token. A suggestion is an entry rather than a string because it also pins how the server hands out credentials, which a typed url cannot say. Every entry's url and `auth` kind was verified against the server's live discovery documents.
- A **typed url resolves to its registry entry by host**, so pasting `https://api.githubcopilot.com/mcp/` reaches GitHub's instance-OAuth connect instead of a dynamic registration its server cannot answer. Only a picked suggestion locks the url field; a url you typed stays editable.
- **Both sign-in paths behave identically.** An instance-OAuth-app server and a dynamic-registration server both show their scopes inline and offer one button that opens the popup. A server that advertises no OAuth says so and the token field takes its place, with the reason (no OAuth app on the instance, or no enterprise license) stated where the token is asked for.
- The path field says where the connection lands and what it becomes ("Signing in saves the connection at this path, as an MCP resource"), and its resource path is suggested from the host (`mcp.sentry.dev` becomes `sentry_mcp`) until you edit it. A path outside your own user space warns that anyone who can read it can use the connection, since its tools run as the account the token belongs to.
- `McpServerOAuthConnect.svelte` (moved out of `flows/content/`) drives the **MCP server's own** OAuth: discovery, dynamic client registration and scope selection. `McpConnect` renders the chrome and the action around it. The two are not alternatives — the parent also handles the case where the credential comes from an OAuth app registered on the instance, which is the only path GitHub supports.
- `McpConnections.svelte`: the chat's `+` menu carries an **MCP connections** submenu, one row per connected server with its provider icon and a switch, plus a way to connect another; the drawer behind it lists everything, hosts the connect card, and disconnects. Menu rows read through the live list rather than the row object they were built from, because a background refresh replaces those objects and a switch bound to the old one silently stops repainting.
- Disconnect goes through a confirmation that says what it does — it deletes the resource, and the chat and any flow pointing at it lose the server. It strips the `token` reference before deleting, because `delete_resource` cascade-deletes every variable a resource value references — a plain delete would have destroyed the credential the provider resource still uses.
- Provider icons are resolved from the server url (registry first, then Windmill's icon set by host name) and loaded through `import.meta.glob`, so a workspace with two connections downloads two icons rather than the whole 230-icon barrel. `listResource` deliberately returns no resource values, so the url costs one read per row; the resolved provider is cached in `localStorage` against the row's `edited_at`, which makes every later open paint immediately and re-read only rows that changed.
- Both connect components take `workspace` as a **required** prop. A forgotten one used to fall back to whichever workspace the ui was showing, which silently created the connection in the wrong place inside a forked session chat.

**Instance settings**
- Each OAuth connect now shows its **Redirect URL** (`{BASE_URL}/oauth/callback/{client}`) with a copy button, next to the client id and secret. Registering an OAuth app previously meant knowing that format, and getting it wrong fails only at the end of the flow with a provider-side error.

## Screenshots

The connect card sits at the top of the connections drawer, with the servers below it:

connect drawer

Choosing the token path swaps the credential rather than stacking a second one, and says what that server takes:

connect token

Connections live in the chat's `+` menu, one switch per server, which stays open while you flip them:

mcp menu

Disconnecting says what it deletes:

disconnect confirmation

The same connect UI backs the AI agent's MCP tool step, under the resource picker:

mcp tool step

In CE, everything works except the MCP server's own OAuth, which is EE: a server that would need it goes straight to the token field and says why. GitHub's sign-in is unaffected, since it runs through the instance's own OAuth app.

ce

Leaving your own user space warns what that means:

shared path

The redirect URL to register with the provider:

redirect url

Calling tools: a read runs unprompted, a mutation waits for confirmation.

write confirmation

read result

## Test plan

- [ ] `cargo test --features mcp --test mcp_token_exfil` — the new POST case pins the confused-deputy property on the new route: a developer who can read the `mcp` resource but not its token variable is denied at the variable-RLS gate *before* any connection is attempted, and an admin still resolves the token (no over-blocking).
- [ ] `cd frontend && npx vitest run src/lib/components/copilot/chat/global/` — 268 pass, 10 new. They pin the security-relevant cases: mutating tool refused on the read path, **unannotated** tool refused on the read path, read-only tool refused on the write path, confirmation only on write, plus partial-failure search.
- [ ] `cargo check --features mcp` and `npm run check` (0 errors).
- [ ] Open the plug in the chat toolbar, connect GitHub, and confirm an `mcp` resource lands with a `$var:` token and arrives already toggled on. Then ask something GitHub-shaped: a read tool should run with no prompt and a mutating one should show the confirmation card first.
- [ ] Toggle a server off and confirm the chat stops advertising its tools without the list reloading.
- [ ] Paste a URL with no suggestion selected: discovery should run on blur, and the token fallback should stay available whether or not it finds OAuth.
- [ ] Disconnect a server and confirm the token variable and any provider resource pointing at it still work.

## What was verified, and what wasn't

Exercised end to end in the running app against a **local stub MCP server** (two tools, one `readOnlyHint: true`, one `false`) driven by a scripted model: the read tool ran with no prompt, the write tool parked on the confirmation card until confirmed, and the stub confirms the bearer token actually arrived. The cascade-delete fix was verified against the live backend — a plain delete destroys the shared variable, strip-then-delete preserves it.

The connect UI was driven in a browser on both surfaces: the drawer (suggestion and typed url, both sign-in paths, the token fallback, connect-and-reset, disconnect confirmation) and the AI agent's MCP tool step. Live discovery against Notion, Linear and Sentry returns their real scopes and reaches the provider's consent screen.

Both sign-in paths were then completed by hand against the real providers, which is what the automated environment could not do: **GitHub** through the instance's OAuth app (the CE path), and **Stripe** through the MCP server's own dynamic registration (the EE path). They share only the card around them — different endpoint, different callback message, different correlation key, different `account` bookkeeping — so one does not stand in for the other.

The real GitHub server was then driven from the chat on that connection: a read tool ran without a prompt, and a mutating one waited on the confirmation card and created an issue once confirmed. So the read/write gate has been exercised against a real server's own annotations as well as against a stub that sets `readOnlyHint` both ways.

**Not verified:** the `ai_evals` A/B below, which is the one item deliberately left.

**No `ai_evals` A/B was run** — deliberately skipped rather than overlooked. The property it would check holds structurally: with no MCP server enabled the tool list and system prompt are byte-identical, which the existing `global/core.test.ts` prompt assertions pin. For users **with** a server enabled the per-iteration cost is real and unmeasured — worth a benchmark before this leaves draft.

## Known limitations

- Every call opens a fresh MCP connection (`from_resource` does DNS + `initialize` + `tools/list`; no pooling), so ~2 extra round trips per call. A short-lived client cache keyed by workspace + path + token digest is a follow-up.
- The default `github` connect scopes are `["workflow", "repo"]`; GitHub's MCP server also advertises `read:org` / `read:user`, so org and user tools may 403 until an admin widens the scopes in instance settings. `oauth_connect.json` is deliberately left alone — changing it would broaden every GitHub connect on the instance.
- `readOnlyHint` is a hint from the MCP server, so a hostile server could mark a mutating tool read-only to skip the confirmation. That is the same trust the user already extends by choosing the server URL.
- Reconnecting over a preserved OAuth token deletes and recreates its variable in two requests, so a failure between them loses the credential that disconnect deliberately kept. `EditVariable` carries no `account`, which is why it cannot be patched in place. The existing resource connect (`AppConnectInner`) has the same property; the fix for both is a transactional connect endpoint, which is a follow-up rather than this PR.


hugocasa:
Disposition of this round, at `1054bffb3d`.

**Fixed**

1. **Invalidated tool lists still authorizing calls.** Right, the generation check only kept the stale answer out of the cache. `loadServerTools` now retries once and fails the call if the generation moved again, rather than handing back a listing that may describe the server that was replaced. Test drives a `getMcpTools` that invalidates while in flight and asserts the call never reaches `callMcpTool`.
3. **Connect operations not pinned to the server.** Fixed: the whole target (`workspace`, `path`, `url`, `name`, `label`) is snapshotted when the operation starts and threaded through `createMcpResource`, so editing the url or the suggestion during a save or a popup cannot bind a credential to a different server. This also covers Claude's nit, which is the same defect.
2. **Popup completions shared across connectors.** Both connectors now keep the popup handle and require `event.source === popup` on top of the client / server-url match. Also fixed a real ordering bug in the same area: the dynamic connector's storage fallback called `cleanup()` before checking the url, so another server's completion stranded the flow. The cross-tab `localStorage` path still correlates only by client / server url: the callback pages carry no nonce, and they are shared machinery for every connect on the instance, so adding one is a change outside this PR.
5. **Caps not hard bounds.** Correct, escaping is server-controlled. The cap is now measured on the serialized result and the body is shrunk until it fits. Test sends 60k backslashes through and asserts the model-facing string is at most 20k.

**Not fixing here, deliberately**

4. **Non-atomic OAuth reconnect.** Real, and not fixable in the frontend: it needs a transactional endpoint. The window is one failed request wide and the recovery is to connect again. Worth noting the existing OAuth connect this mirrors (`AppConnectInner`) has the same property and additionally holds the access token in memory between its steps, so this is the house behaviour rather than a regression. The right fix is a `POST /w/{ws}/mcp/connect` doing collision check plus variable plus resource in one transaction under the caller's RLS, which would cover both flows. Follow-up, not this PR.

6. **`ai_evals` global A/B.** Deliberately skipped, as the description states. With no server enabled the tool list and system prompt are byte-identical, which the existing `global/core.test.ts` assertions pin; for users who do enable a server the per-iteration cost is real and unmeasured, and that is recorded in the PR body rather than papered over.

**On the divergence, since three rounds have pushed on it:** compared with `AppConnectInner`, this card is stricter on callback correlation (that one accepts any same-origin `success`) and on workspace pinning (that one reads the live store at every step, which is not safe here because a session chat operates on a workspace other than the one displayed), and looser on variable collisions (that one refuses every existing variable; this one refuses any it did not stamp, which is what makes reconnect work after a disconnect deliberately preserves the token).

**Verification.** A real GitHub OAuth sign-in will be run by hand before this leaves draft. It is the one path not exercised here: this instance has no OAuth app configured, and its `base_url` is http, which the dynamic-registration providers refuse. Everything else in the connect flow has been driven in a browser, including the token path end to end after each of these rewrites.

hugocasa:
Closing the loop on the verification note above: both OAuth paths have now been completed by hand against the real providers. **GitHub** through the instance's OAuth app (CE path, `McpConnect.finishProviderOAuth`, `account.client = github`) and **Stripe** through the MCP server's own dynamic registration (EE path, `McpServerOAuthConnect`, `account.client = mcp` plus `mcp_server_url`). Those are separate code paths that share only the card around them, so both were needed.

That leaves calling the real GitHub server's own tools as the one unexercised item; the read/write gate itself has been driven against a stub that sets `readOnlyHint` both ways. The PR body has been updated to match, and the non-atomic OAuth reconnect is now recorded under known limitations rather than left implicit.

hugocasa:
Disposition at `922eee9111`.

**Fixed**

- **Scopes not snapshotted with the popup.** `pending` now carries a copied scope array, used both to build the authorization url and to create the account, so editing the list behind an open popup can no longer record a grant that was never asked for. That one fails at refresh rather than at connect, which is the worst place to find it.
- **Stale server list left installed** (Claude). Dropping the late answer was not enough on its own: `refreshMcpServers` now clears when the operating workspace moved, instead of going on advertising the previous workspace's paths.
- **`event.source` gate missing on the dynamic connector's error branch** (Claude). Same gate as `McpConnect` now, above both branches.
- **Cached annotations outliving a resource revision** (half of the enablement finding). `McpServer` carries `edited_at` and the tool cache is keyed `workspace:path:revision`, so a resource edited or recreated outside this flow cannot answer with the previous server's `readOnlyHint`. This is the half that guards the confirmation gate, and it costs nothing. Regression test added.

**Not fixing: keying the enabled set by revision**

Deliberate. It closes a narrow gap (delete an mcp resource through the Resources page, create a *different* server at the same path, and it inherits the previous one's enablement) at the cost of every ordinary edit silently switching a server off in every chat until the user turns it on again, including edits our own reconnect makes. The enabled set is a per-browser preference, not the security boundary — what a server can reach is bounded by the token resolving under the caller's RLS — so the trade does not buy enough to be worth the surprise. Raised with the maintainer and settled that way.

CI is green on this head, and both OAuth paths plus real GitHub read and write calls have now been exercised by hand.

hugocasa:
Disposition at `6f6f621c97`.

**Fixed**

- **Stale revision at call time** (codex P1) — moved the gate off the frontend cache entirely rather than validating the snapshot. `call_mcp_tool` now takes `read_only`, set only by the path that skipped the user's confirmation, and checks it against `available_tools()` from the handshake it just performed — so the classification always describes the server about to be called, never a listing of whatever the resource pointed at earlier. No extra round trip (the client already lists on connect), and a mismatch fails the call with a message that sends the model to the write tool, which prompts. Re-fetching `edited_at` before the call would have cost a request and stayed TOCTOU.

  To be explicit about what this does *not* do: `readOnlyHint` is still the server's own claim, so a server the user connected can still lie. That boundary is unchanged and documented; what closes here is the mismatch between the server that was classified and the server that is called.

- **Callback lost on remount** (codex P2) — the URL field, the suggested-provider buttons and the token-mode swap are disabled while a dynamic-OAuth popup is open, so nothing can destroy the connector that holds the pending listener.

- **Stale comment** (claude) — the sentence claiming two connectors can't be told apart is gone; the `event.source` gate above it does exactly that.

**Verified**

Against a local stub MCP server: a read-only tool asserted read-only runs (200), a mutating tool asserted read-only is refused (400) with no `tools/call` reaching the server, and the same mutating tool with no assertion runs. Then through the chat itself — `call_mcp_read_tool` sends `read_only: true`, and the confirmed write call sends no assertion. The popup guard was exercised against a stub advertising OAuth with dynamic client registration: with the popup open, the url input and both mode switches are disabled.

Added one frontend regression test (the read facade sends the assertion, the write facade does not).

PR #10659, opened 2026-08-12
test: fix repro_diffname CLI flake by draining the app dependency job

rubenfiszel:
## Summary

`cli/test/repro_diffname.test.ts` flakes on the linux CI runner ([run 31574940152](https://github.com/windmill-labs/windmill/actions/runs/31574940152/job/94044847310)) with a bare `Expected: 0 / Received: 1` on the final `sync push`.

Root cause is a race with the app dependency job, not the code under test:

1. `POST /apps/create_raw` queues an `AppDependencies` job in the same transaction (`backend/windmill-api/src/apps.rs`), and the worker later writes the generated inline-script locks back into the app value (`backend/windmill-worker/src/worker_lockfiles.rs`).
2. The test pulled immediately, so the local copy had no `a.lock` / `b.lock`.
3. When the job landed *between* the pull and the push, `sync push` saw a real diff (`- raw_app .../a.lock`) instead of a no-op, so it took the bundle-rebuild branch in `cli/src/commands/app/raw_apps.ts`.
4. Rebuilding runs `npm install`, and this fixture — a raw app created straight through the API — has no `package.json`:

```
📦 node_modules not found, running npm install...
npm error enoent Could not read package.json: .../diffname_app.raw_app/package.json
npm install failed with exit code 254
```

The window is ~600ms wide, which is why only the slower runner hits it. The earlier Windows fix (#8811) excluded lock files from the precondition, which papered over the same race one step earlier.

## Changes

- Wait for the deployment jobs to drain (existing `waitForDeploymentJobs` helper) after creating the app, so the pull captures the locks and the final push is the no-op the test intends to assert.
- Assert the pulled file list including `a.lock` / `b.lock`: now that lock generation is deterministic, a lock that never lands fails as a named precondition instead of resurfacing later as an opaque npm error.
- Add a `runCLI` helper that throws with the CLI's stdout/stderr on a non-zero exit — the previous `expect(result.code).toEqual(0)` swallowed the CLI output, which is why CI showed nothing but `Received: 1`.

## Test plan

- [x] `bun test test/repro_diffname.test.ts` passes.
- [x] Reproduced the flake deterministically by making the worker poll lazily (temporarily overriding the hardcoded `SLEEP_QUEUE` in `cli/test/cargo_backend.ts`) to steer the dependency job into the pull→push window: before the fix `SLEEP_QUEUE=400` and `800` fail with the npm ENOENT, after the fix `300/400/800/900` all pass.
- [x] Full `bun test test/` on linux: 1490 pass, 36 skip, 0 fail. This matters because the suite shares one backend whose reset never clears the queue — `repro_diffname` runs after the tests that leave enabled schedules parked in `v2_job_queue`, and the helper's `job_kinds` filter keeps the wait from being held by them.
- [ ] Windows CI job green (the runner where the earlier variant of this race was first seen).

---
## Summary by cubic
Fixes a flaky CLI test by waiting for the app dependency job to finish before pulling, so the final `sync push` is a no-op instead of triggering a bundle rebuild and `npm` error.

- **Bug Fixes**
  - Wait for deployment jobs (`waitForDeploymentJobs`) after creating the app so pull includes generated lock files and avoids the race.
  - Assert pulled files include `a.lock` and `b.lock` to fail early if locks are missing.
  - Add a `runCLI` helper that throws with stdout/stderr on non-zero exit to make failures actionable.

<sup>Written for commit 00bcdcd19471a946b25b52e79fa393b0c9c58d1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/test/repro_diffname.test.ts
+++ b/cli/test/repro_diffname.test.ts
@@ -1,7 +1,7 @@
 import { test, expect } from "bun:test";
 import { writeFile, readdir } from "node:fs/promises";
 import path from "node:path";
-import { withTestBackend } from "./test_backend.ts";
+import { withTestBackend, type TestBackend } from "./test_backend.ts";
 import { addWorkspace } from "../src/commands/workspace/workspace.ts";
 
 /**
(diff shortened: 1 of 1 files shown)


rubenfiszel:
✅ Review round clean @ 00bcdcd19471a946b25b52e79fa393b0c9c58d1a — nit-only verdicts at b724b8dacf98df949061e566385da76af21e5810 (codex: nit, claude: good to merge); nit addressed in 00bcdcd19471a946b25b52e79fa393b0c9c58d1a. CLI Tests green on this head: test-linux, test-windows, build-check, test-summary all pass.

PR #10660, opened 2026-08-12
fix: clear orphaned usr_to_group rows on service account creation

rubenfiszel:
## Summary

Creating a service account failed with `duplicate key value violates unique constraint "usr_to_group_pkey"` whenever an orphaned `usr_to_group` row already existed for that username.

`usr_to_group` has no FK to `usr`, and the `leave_workspace` handlers delete the `usr` row without touching group memberships, so the mappings survive. Recreating that username as a service account then collided on the `all`-group insert, which — unlike the deployers insert ten lines below and `auto_add_user` — had no `ON CONFLICT DO NOTHING`.

Fixes WIN-2349.

## Changes

- `backend/windmill-api-workspaces/src/workspaces_ee.rs` (EE): in `create_service_account`, delete any stale `usr_to_group` rows for the username before inserting the requested memberships, and add `ON CONFLICT DO NOTHING` to the `all`-group insert to match the sibling call sites.
- `backend/windmill-api-integration-tests/tests/workspaces.rs`: EE-gated regression test.
- `backend/ee-repo-ref.txt`: bumped to the companion EE commit.
- `backend/.sqlx/`: dropped the cache entry for the old `INSERT` text, now orphaned. No `prepare` run was needed — the offline key is `sha256(query)`, and both the new `DELETE` and the `ON CONFLICT` `INSERT` already have entries from other call sites (verified byte-for-byte, including parameter types). The new test uses non-macro `sqlx::query`, so it adds no cache entries either.

### Note on scope

The issue asked only for `ON CONFLICT DO NOTHING`. That alone is not enough: the same orphan condition also leaves stale rows for *other* groups, and since auth resolves group membership by `(workspace_id, username)`, swallowing the conflict would let a recreated service account silently inherit `wm_deployers` or any custom privileged group — where the old behavior was a hard error. Two independent reviewers flagged this, so the `DELETE` is included.

It is safe at that point: the `INSERT INTO usr` immediately above has just succeeded under the table's primary key, proving no live user holds this username in this workspace, and the whole handler runs in one transaction. The `DELETE` is workspace-scoped (`WHERE workspace_id = $1 AND usr = $2`), which the test pins with a second workspace. The `ON CONFLICT DO NOTHING` is kept both for consistency with the five sibling inserts and because a concurrent writer can still race a row in between the two statements.

### Regression test

`test_create_service_account_drops_orphaned_group_memberships`, `#[cfg(feature = "private")]` so it compiles and runs under CI's `cargo test --features enterprise,…,private,… --all`. It plants orphaned `all` + `wm_deployers` + `secrets` rows for a username with no `usr` row, creates the service account with `add_to_deployers: false`, and asserts the resulting memberships are exactly `{all}` — plus a live user of the same username in a second workspace whose memberships must be untouched.

Verified it fails without the fix, in both directions:

| Handler state | Failure |
|---|---|
| Original | `400` — `duplicate key value violates unique constraint "usr_to_group_pkey"` |
| `ON CONFLICT` only, no `DELETE` | `left: ["all", "secrets", "wm_deployers"]` vs `right: ["all"]` — *service account inherited stale memberships* |
| This PR | passes (full `workspaces` suite: 7 passed, 0 failed) |

### Manual end-to-end

Against a local backend built with `--features quickjs,enterprise,private`, driving the real `POST /api/w/{ws}/workspaces/create_service_account`:

| Case | Orphans planted | `add_to_deployers` | Result | Final memberships |
|---|---|---|---|---|
| 1 | `all`, `wm_deployers`, `secrets_team` | `false` | 201 | `all` |
| 2 | `all`, `secrets_team` | `true` | 201 | `all`, `wm_deployers` |
| 3 | none | `false` | 201 | `all` |
| 4 | username held by a **live** user | `false` | 400 `user with username … already exists` | live user's `all` intact |

Case 4 confirms the pre-existing duplicate-username guard still rejects live collisions.

### Follow-up: WIN-2350

The same privilege-inheritance hole stays open on the far more common **rejoin** path: `leave_workspace` deletes from `usr` without clearing `usr_to_group` (`windmill-api-users/src/users.rs:3281`, `windmill-api-workspaces/src/workspaces.rs:8154`), and the re-add path (`users.rs:1215-1235`, `workspaces.rs:8487`, `workspaces_ee.rs:265`) inserts with `ON CONFLICT DO NOTHING` and no cleanup — so a regular user who leaves and is later re-added silently regains every group they had, `wm_deployers` included. The admin delete-user path already cleans up correctly. That is a latent privilege-escalation vector rather than a cleanliness issue, but it is a separate change on CE code paths and is deliberately out of scope here. Filed as [WIN-2350](https://linear.app/windmilldev/issue/WIN-2350/re-added-workspace-user-silently-regains-all-previous-groups-including).


rubenfiszel:
Thanks — all three addressed.

**P2 — `DELETE` scoping.** Confirmed workspace-scoped; the query is `DELETE FROM usr_to_group WHERE workspace_id = $1 AND usr = $2` (cache key `526b0c0a…`, `Text`×2), not the unscoped `WHERE usr = $1` variant. You were right that this is invisible from the public diff and that picking wrong would compile clean, so it is now pinned by test rather than by assertion: the new test puts a **live** user of the same username in a second workspace and asserts their memberships survive untouched.

**P2 — no-test justification was wrong.** You are correct and I was not: `backend-test.yml:298` runs `cargo test --features enterprise,…,private,… --all`, and `windmill-api-integration-tests` declares both features. Test added in 8876aeca69 — `test_create_service_account_drops_orphaned_group_memberships`, `#[cfg(feature = "private")]`. I verified it actually guards something by running it against both broken states:

| Handler state | Result |
|---|---|
| Original | `400` — `duplicate key value violates unique constraint "usr_to_group_pkey"` |
| `ON CONFLICT` only, no `DELETE` | `left: ["all", "secrets", "wm_deployers"]` vs `right: ["all"]` |
| This PR | passes (full `workspaces` suite: 7 passed, 0 failed) |

It uses non-macro `sqlx::query`, so it adds no `.sqlx` entries and does not drag in the `prepare --all-targets` hazard.

**P2 — rejoin path.** Verified and agreed; the description now states it as a latent privilege-escalation vector rather than a cleanliness follow-up. Left out of this PR because it is a separate change on CE code paths (`users.rs:1215-1235`, `workspaces.rs:8487`, `workspaces_ee.rs:265`) plus the two `leave_workspace` handlers, and it wants its own test matrix. Filing it as a ticket rather than leaving it only in a PR body.

Re: your merge gate — agreed, `check_ee`/`check_oss` are the only mechanical check on the hand-edited `.sqlx`, and they are re-running on the new head.

rubenfiszel:
Follow-up filed as [WIN-2350](https://linear.app/windmilldev/issue/WIN-2350/re-added-workspace-user-silently-regains-all-previous-groups-including) — *Re-added workspace user silently regains all previous groups, including wm_deployers*.

rubenfiszel:
✅ Review round clean @ 8876aeca69be87552d16ae9285093898f06440d5

Both reviewers returned `Good to merge` with no findings on this head, and all CI checks are green — including `cargo_test` (17m36s), the only run of the new EE-gated test outside my local box, and `check_ee`/`check_ee_full`, the mechanical gate on the hand-edited `.sqlx` cache.

PR #10665, opened 2026-08-12
fix: authorize GET /concurrency_groups/{job_id}/key per job

rubenfiszel:
## Summary

`GET /api/concurrency_groups/{job_id}/key` required authentication (#10221) but no
authorization: any authenticated user of the instance could read the concurrency key of
**any** job, in any workspace, from its UUID. A concurrency key embeds the workspace id,
the runnable path and any `$args[...]`-templated argument values
(`wsa/script/u/admin/concurrent_demo`), so this leaked cross-workspace metadata.

The endpoint is now gated on read access to the job itself, i.e. exactly what the run
page that consumes it already requires. Fixes WIN-2353.

## Changes

- `get_concurrency_key` moves from `windmill-api-jobs` to `windmill-api`, where the shared
  `require_job_read_access` gate lives, so this route uses the same predicate as every
  other job-read route (share read links, tag scopes, flow-lineage visibility and the
  created-by fast path all come for free) instead of a second, divergent check.
- The route carries no workspace, so the extracted `ApiAuthed` is not workspace-scoped
  (no groups/folders, `is_admin` only for superadmins). The handler resolves the job's
  workspace from `v2_job`, re-resolves the caller's token against that workspace via
  `AuthCache`, and evaluates access there.
- A caller outside the job's workspace and a caller naming a job that does not exist get
  the same `404`, so the route is not an instance-wide job-existence oracle. Members reach
  `require_job_read_access` and get its `403`, which discloses existence only inside a
  workspace they belong to.
- `require_job_read_access` is now `pub(crate)`; `ApiAuthed` stays an explicit handler
  extractor because this service is nested after `route_layer(from_extractor::<ApiAuthed>())`
  and would otherwise be unauthenticated again.
- Integration test pinning each direction: same-workspace non-viewer → 403 and no key,
  cross-workspace → 404 identical to an unknown UUID, owner → 200 and the key.
- OpenAPI: documents the access requirement and the 403/404 semantics.

Resulting matrix (verified against a local instance, workspaces `wsa`/`wsb`):

| caller | before | after |
|---|---|---|
| unauthenticated | 401 | 401 |
| member of another workspace | 200 + key | 404 |
| any caller, unknown job UUID | 200 + `null` | 404 |
| workspace member who cannot see the run | 200 + key | 403 |
| same member holding a share read link | 200 + key | 200 + key |
| the run's owner / a workspace admin | 200 + key | 200 + key |

No frontend change: the run page only calls this for a run it has already loaded, and the
share-link page keeps working through the existing `X-View-Token` interceptor.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test concurrency_groups` (2 passed)
- [x] Manual matrix above with `curl` against a local backend, including share read links
      (`X-View-Token` header and `view_token` query param) and a forged token (403)
- [x] Inaccessible job and random UUID return byte-identical responses
- [x] Run page still renders the "Concurrency: …" link for an authorized viewer
      (`GET /api/concurrency_groups/{id}/key => 200`, checked in the browser)
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api` passes with the added `.sqlx` entry
      (and fails without it, so the entry is the one CI needs)

### DIFF
--- a/backend/.sqlx/query-9724f8493bba07ffd480c71d65be1b3f5b18fde66d33f33c7c3136452c20d4eb.json
+++ b/backend/.sqlx/query-9724f8493bba07ffd480c71d65be1b3f5b18fde66d33f33c7c3136452c20d4eb.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT workspace_id, created_by FROM v2_job WHERE id = $1",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "workspace_id",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "created_by",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid"
+      ]
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "9724f8493bba07ffd480c71d65be1b3f5b18fde66d33f33c7c3136452c20d4eb"
+}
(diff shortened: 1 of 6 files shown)


rubenfiszel:
✅ Review round clean @ dd913abf7638feab42731220dde2f2ac57b400ea

PR #10666, opened 2026-08-12
feat: add memory limits to the go build subprocess

rubenfiszel:
## Summary

Go compilation had no memory bound. `go build` and `go mod init|tidy|download` are spawned with `env_clear()` and only receive `PATH`, `GOPATH`, `HOME`, `GOCACHE` and the proxy vars, so nothing configured on the worker reached them and the container memory limit was the only control. A compilation that outgrows the cgroup gets the worker process OOM-killed, taking every job colocated on it down with it.

This gives the toolchain a budget derived from the worker's own memory limit, and forwards the Go tuning knobs from the worker group config to the compilation step (previously they only reached the run step).

`GOMEMLIMIT` is a *soft* limit: the GC works harder as the heap approaches it rather than failing the allocation, so a pathological build turns into a slow build instead of a dead worker.

## Changes

- `GO_BUILD_LIMITS` (`windmill-worker/src/worker.rs`) resolves a build budget: `GO_BUILD_MEMLIMIT` (`512MB`, `2GiB`, … / `0` / `off`) if set, else 75% of the worker's memory reading. A worker with no reading to scale from stays unlimited.
- The budget is split into `GOMEMLIMIT` **and** `GOMAXPROCS`, applied as an atomic pair. `GOMEMLIMIT` caps one process, but a build is a driver plus up to `-p` (= `GOMAXPROCS`) compilers that each inherit it, so a per-process cap alone bounds nothing — half of the pair is no limit, whichever half is missing. Cores are put to work while the budget affords each compiler a comfortable share; below that the process count holds and the share absorbs the difference, down to an absolute floor.
- `GOMEMLIMIT`, `GOGC` and `GOMAXPROCS` set on the worker group are forwarded to all three toolchain spawn sites. The allowlist is deliberately narrow: the built binary is cached under a hash of the source and lockfile alone (and shared through the object store), so a var that alters codegen — `GOFLAGS`, build tags — would let two worker groups disagree about the contents of one cache entry. A worker group that pins either half of the pair owns both, and the derived pair steps aside entirely.
- Forwarded values are passed through verbatim, not normalized. Parsing `512MB` into bytes for the toolchain was considered and rejected: the run step already forwards the same variable verbatim, so a spelling the Go runtime rejects (it takes a bare byte count or `KiB`/`MiB`/`GiB` only) would compile fine and then abort on the binary it produced — confirmed by running it. Failing both steps identically is the clearer signal.
- `go build` passes `-p` explicitly. `GOMAXPROCS` only supplies its *default*, and a `GOFLAGS=-p=…` persisted in the toolchain's own env file outranks that — `env_clear()` does not disable it, since the command restores `HOME`. A flag on the command line is applied last, so the parallelism the aggregate assumes is the one the build gets. It is only passed when the installed value is a positive integer, so a worker group's typo stays as harmless as the Go runtime makes it.
- The `go mod init|tidy|download` steps get the whole budget instead of a compiler's share: they are a single process with nothing else running, and their fetch queue is sized from `GOMAXPROCS`, so pinning it would slow them without bounding anything.
- The limits actually handed to the toolchain are logged once per distinct setting, so an operator can tell what is in force — including when a worker group's own values are the ones installed.

Ordering note: the forwarded envs are applied *before* each command's explicit `.env` calls, so the worker's own `PATH`/`GOPATH`/`GOCACHE`/`HOME` still win — the same precedence the run step already has.

## Cost of the limit (measured)

The split was chosen by measurement. Cold build cache, dependency-heavy script (aws-sdk-go-v2 s3, 36 modules), 24-core host; within each table every row holds the same aggregate.

At a 2.25 GiB budget (a 3 GiB worker):

| split | wall | peak RSS |
|---|---|---|
| unlimited | 4.97s | ~650 MB |
| `GOMAXPROCS=8` alone | 5.25s | 611 MB |
| 256MB x 8 compilers | 7.46s | 411 MB |
| **384MB x 5 compilers** | **6.92s** | 450 MB |
| 768MB x 2 compilers | 10.38s | 643 MB |
| 1.1GB x 1 compiler | 23.93s | 612 MB |

At a 768 MiB budget (a 1 GiB worker):

| split | wall |
|---|---|
| 384MB x 1 compiler | 26.63s |
| 256MB x 2 | 14.37s |
| 192MB x 3 | 10.96s |
| **128MB x 5** | **9.20s** |
| 96MB x 7 | 9.38s |

Both budgets peak at the same *process count*, not the same share: what a build is sensitive to is how much of it overlaps, and pinning parallelism alone costs almost nothing (+5%). So the split puts one compiler on each core while the budget affords them a comfortable share, holds a minimum process count when it cannot, and only then lets the share fall to an absolute floor. Dropping to a single compiler — the naive way to give the serial link phase a real share — is ~3x worse.

What remains is ~40% on a cold, dependency-heavy build on a memory-tight worker, paid once per script version since the built binary is cached locally and in the object store. An ordinary build never approaches the cap.

## Test plan

Verified against a running worker (`cargo check` and unit tests do not exercise a spawn path), with the backend inside a 3 GiB cgroup via `systemd-run --user --scope -p MemoryMax=3G`:

- [x] Derived path (3 GiB cgroup, 24 cores): log reads `Go compilation limited to GOMEMLIMIT=402653184 GOMAXPROCS=5`, job compiles and runs
- [x] Small-worker budget (`GO_BUILD_MEMLIMIT=768MB`, what a 1 GiB worker derives): `GOMEMLIMIT=134217728 GOMAXPROCS=5` — five compilers, not one
- [x] `GO_BUILD_MEMLIMIT=300MiB` on the worker: resolves and applies, job compiles and runs
- [x] Worker group `GOMEMLIMIT=512MiB` + `GOGC=50`: honored, job compiles and runs
- [x] Worker group `GOMAXPROCS=3` alone: derived pair steps aside and the log says so (`Go compilation limited to GOMAXPROCS=3`), claiming no bound it does not install
- [x] Worker group `GOMEMLIMIT=bogus` / `512MB`: `go mod init` aborts with `malformed GOMEMLIMIT` — proves the value reaches the compilation subprocess, and matches how the same value already fails the run step
- [x] Worker group value takes precedence over `GO_BUILD_MEMLIMIT`
- [x] No memory reading available (plain host, `memory.max=max`): unlimited, job compiles and runs
- [x] Persisted `GOFLAGS=-p=0` in the worker's `HOME`: build succeeds with the derived `-p`, and fails with `go: -p must be a positive integer: 0` when no `-p` is asserted — the bypass is reachable, and the flag closes it
- [x] Per-step limits: `Go dependency resolution limited to GOMEMLIMIT=2415919104` alongside `Go compilation limited to GOMEMLIMIT=402653184 GOMAXPROCS=5`
- [x] Timing matrices above, on the real toolchain
- [x] `cargo check -p windmill-worker --all-targets`; unit tests pin the budget split (floor, `off`/`0`, unparseable fallback) and the env merge (either-half-pinned, whitespace, empty allowlist entry)

Not exercised: a build large enough to OOM a worker outright. The aggregate is arithmetic over `GOMEMLIMIT` × `GOMAXPROCS`, and non-Go allocations in the toolchain sit outside it, so the budget is a close bound rather than a hard one. Below ~1 GiB of worker memory the floor under the share wins and the aggregate exceeds the budget — deliberately, since every split that respects it there measured worse.

Fixes WIN-2354

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,6 +120,14 @@ A crash like this takes down every job on that worker, not just yours, so check 
 backend log after the run rather than only the job's own status. If you cannot run one,
 say which path went unexercised instead of implying it was verified.
 
+### A worker runs one job at a time
+
+`NUM_WORKERS > 1` falls back to 1 outside native mode (`backend/src/main.rs`, unless
+`I_ACK_NUM_WORKERS_IS_UNSAFE`), and native workers only serve the lightweight tags in
+`NATIVE_TAGS` — no `go`, `python3`, `dependency` or `flow`. So anything sizing a per-job
+resource budget (memory, CPU, temp space) is sharing the worker with the worker process
+itself, not with a second script job.
+
 ## Banned Patterns
 
 ### `$bindable(default_value)` on optional props
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,13 +120,17 @@ A crash like this takes down every job on that worker, not just yours, so check 
 backend log after the run rather than only the job's own status. If you cannot run one,
 say which path went unexercised instead of implying it was verified.
 
-### A worker runs one job at a time
+### How many jobs a worker runs at once
 
 `NUM_WORKERS > 1` falls back to 1 outside native mode (`backend/src/main.rs`, unless
-`I_ACK_NUM_WORKERS_IS_UNSAFE`), and native workers only serve the lightweight tags in
-`NATIVE_TAGS` — no `go`, `python3`, `dependency` or `flow`. So anything sizing a per-job
-resource budget (memory, CPU, temp space) is sharing the worker with the worker process
-itself, not with a second script job.
+`I_ACK_NUM_WORKERS_IS_UNSAFE`), so a worker serving script tags — `go`, `python3`,
+`dependency`, `flow`, … — runs one job at a time: a per-job resource budget (memory, CPU,
+temp space) shares the worker with the worker process alone.
+
+Native mode is the exception, and budgets for its tags must divide by its concurrency:
+it forces 8 workers, and `NATIVE_TAGS` includes executors that already claim a per-job
+share of the worker's memory (`postgresql` and `mysql` through `MAX_SQL_RESULT_SIZE`), so
+up to 8 of those run against the same limit at once.
 
 ## Banned Patterns
 
--- a/backend/windmill-common/src/worker.rs
+++ b/backend/windmill-common/src/worker.rs
@@ -1528,6 +1528,12 @@ pub fn get_affinity_cpus() -> Option<usize> {
 
 #[cfg(windows)]
 pub fn get_affinity_cpus() -> Option<usize> {
+    // The 1CU cap is a policy rather than a bandwidth quota, so it is the whole
+    // answer here as it is for `get_vcpus` and `get_memory` — a consumer that reads
+    // this as the hardware count would raise the worker back above the cap.
+    if *LIMIT_WINDOWS_TO_1CU {
+        return Some(1);
+    }
     let mut sys = System::new();
     sys.refresh_cpu_all();
     Some(sys.cpus().len()).filter(|cpus| *cpus > 0)
--- a/backend/windmill-common/src/worker.rs
+++ b/backend/windmill-common/src/worker.rs
@@ -1497,6 +1497,42 @@ pub fn get_cpu_period() -> Option<i64> {
     Some(100000)
 }
 
+/// CPUs the process is allowed to run on, ignoring any bandwidth quota — the count
+/// Go's `NumCPU` reports. `available_parallelism` cannot stand in for it: that folds
+/// the quota in, so a fraction of a CPU makes it report a single-core machine.
+#[cfg(not(windows))]
+pub fn get_affinity_cpus() -> Option<usize> {
+    // "Cpus_allowed_list:\t0-7,16-23"
+    let status = parse_file::<String>("/proc/self/status")?;
+    let list = status
+        .split("Cpus_allowed_list:")
+        .nth(1)?
+        .lines()
+        .next()?
+        .trim();
+
+    let cpus = list
+        .split(',')
+        .map(|range| {
+            let (first, last) = range.split_once('-').unwrap_or((range, range));
+            let (first, last) = (first.trim().parse::<usize>(), last.trim().parse::<usize>());
+            match (first, last) {
+                (Ok(first), Ok(last)) if last >= first => Some(last - first + 1),
+                _ => None,
+            }
+        })
+        .sum::<Option<usize>>()?;
+
+    (cpus > 0).then_some(cpus)
+}
+
+#[cfg(windows)]
+pub fn get_affinity_cpus() -> Option<usize> {
+    let mut sys = System::new();
+    sys.refresh_cpu_all();
+    Some(sys.cpus().len()).filter(|cpus| *cpus > 0)
+}
+
 #[cfg(not(windows))]
 fn get_memory_from_meminfo() -> Option<i64> {
     let memory_info = parse_file::<String>("/proc/meminfo")?;
--- a/backend/windmill-common/src/worker.rs
+++ b/backend/windmill-common/src/worker.rs
@@ -1472,6 +1472,31 @@ pub fn get_vcpus() -> Option<i64> {
     (sys.cpus().len() * 100000).try_into().ok()
 }
 
+/// The window `get_vcpus`'s quota is spent over, in the same microseconds. Only
+/// their ratio is a number of CPUs, and the window is configurable — 100ms is
+/// merely its usual value.
+#[cfg(not(windows))]
+pub fn get_cpu_period() -> Option<i64> {
+    if Path::new("/sys/fs/cgroup/cpu/cpu.cfs_period_us").exists() {
+        // cgroup v1
+        parse_file("/sys/fs/cgroup/cpu/cpu.cfs_period_us")
+    } else {
+        // cgroup v2: `cpu.max` is "<quota|max> <period>"
+        let cgroup_path = get_cgroupv2_path()?;
+        parse_file::<String>(&format!("{cgroup_path}/cpu.max"))?
+            .split_whitespace()
+            .nth(1)?
+            .parse()
+            .ok()
+    }
+    .filter(|period| *period > 0)
+}
+
+#[cfg(windows)]
+pub fn get_cpu_period() -> Option<i64> {
+    Some(100000)
+}
+
 #[cfg(not(windows))]
 fn get_memory_from_meminfo() -> Option<i64> {
     let memory_info = parse_file::<String>("/proc/meminfo")?;
--- a/backend/windmill-worker/src/go_executor.rs
+++ b/backend/windmill-worker/src/go_executor.rs
@@ -95,26 +95,45 @@ pub const GO_OBJECT_STORE_PREFIX: &str =
 /// entry — including the one shared through the object store.
 const GO_TOOLCHAIN_TUNING_ENVS: [&str; 3] = ["GOMEMLIMIT", "GOGC", "GOMAXPROCS"];
 
+/// The two shapes a Go toolchain invocation takes, which spend the budget
+/// differently: a build fans out into a driver plus compilers, while the module
+/// steps are one process with nothing else running.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum GoToolchainStep {
+    Build,
+    Mod,
+}
+
+impl GoToolchainStep {
+    fn label(&self) -> &'static str {
+        match self {
+            GoToolchainStep::Build => "Go compilation",
+            GoToolchainStep::Mod => "Go dependency resolution",
+        }
+    }
+}
+
 /// Memory and parallelism settings for the Go toolchain subprocesses, which
 /// otherwise inherit nothing (they are spawned with a cleared environment).
 ///
 /// Must be applied before the explicit `.env` calls of each command so the
 /// worker's own `PATH`/`GOPATH`/`GOCACHE`/`HOME` win over a worker group setting
 /// the same names, as they do on the run step.
-fn go_toolchain_envs() -> Vec<(String, String)> {
+fn go_toolchain_envs(step: GoToolchainStep) -> Vec<(String, String)> {
     let worker_config = windmill_common::worker::WORKER_CONFIG.load();
-    let envs = merge_go_toolchain_envs(&worker_config.env_vars, *GO_BUILD_LIMITS);
-    log_go_toolchain_limits(&envs);
+    let envs = merge_go_toolchain_envs(&worker_config.env_vars, *GO_BUILD_LIMITS, step);
+    log_go_toolchain_limits(step, &envs);
     envs
 }
 
 /// A slow compilation is the symptom of a limit set too low, and nothing else would
 /// tell an operator that one is in force or what it resolved to. Reports what the
-/// toolchain is actually given, since a worker group can pin values of its own —
-/// on change rather than per job, because it can also do so at runtime.
-fn log_go_toolchain_limits(envs: &[(String, String)]) {
+/// toolchain is actually given, since a worker group can pin values of its own, and
+/// once per distinct setting rather than per job, since it can move them at runtime.
+fn log_go_toolchain_limits(step: GoToolchainStep, envs: &[(String, String)]) {
     lazy_static::lazy_static! {
-        static ref LAST_LOGGED: std::sync::Mutex<Option<String>> = std::sync::Mutex::new(None);
+        static ref LOGGED: std::sync::Mutex<std::collections::HashSet<String>> =
+            Default::default();
     }
 
     let limits = envs
@@ -122,24 +141,24 @@ fn log_go_toolchain_limits(envs: &[(String, String)]) {
         .map(|(k, v)| format!("{k}={v}"))
         .collect::<Vec<_>>()
         .join(" ");
+    let line = if limits.is_empty() {
+        format!("{} is unlimited", step.label())
+    } else {
+        format!("{} limited to {limits}", step.label())
+    };
 
-    let Ok(mut last_logged) = LAST_LOGGED.lock() else {
+    let Ok(mut logged) = LOGGED.lock() else {
         return;
     };
-    if last_logged.as_deref() == Some(limits.as_str()) {
-        return;
+    if logged.insert(line.clone()) {
+        tracing::info!("{line}");
     }
-    if limits.is_empty() {
-        tracing::info!("Go compilation is unlimited");
-    } else {
-        tracing::info!("Go compilation limited to {limits}");
-    }
-    *last_logged = Some(limits);
 }
 
 fn merge_go_toolchain_envs(
     worker_envs: &HashMap<String, String>,
     derived: Option<GoBuildLimits>,
+    step: GoToolchainStep,
 ) -> Vec<(String, String)> {
     // Values reach the toolchain as the worker group wrote them, the same way they
     // reach the run step: a `GOMEMLIMIT` the Go runtime rejects then fails both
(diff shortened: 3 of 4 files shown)


rubenfiszel:
Both reviewers landed on the same defect and it was real — fixed in df4991af2f, along with the two P2s.

**P1 / atomic pair (codex, claude).** Correct: a worker group pinning `GOMEMLIMIT` alone left `go build` free to use every CPU, and pinning `GOMAXPROCS` alone left the derived per-process cap divided by a process count that no longer applied. Half of the pair is not a weaker limit, it is no limit. Either half configured now hands the whole policy to the worker group and the derived pair steps aside. `GOGC` is orthogonal and still rides along.

**P2 / trimmed values (codex).** Correct, and it defeated the stated contract: `" 512MiB "` compiled as `512MiB` and then aborted the run step, which forwards the untrimmed original. Trimming is now only used to test emptiness (an allowlisted name the worker never set resolves to `""` and must not shadow the derived pair); the value itself is forwarded byte-for-byte.

**P2 / the serial link phase gets `budget/(p+1)` (claude).** Real, and worth more than an argument, so I measured it rather than reasoning about it — cold cache, aws-sdk-go-v2 s3 (36 modules), 24 cores, every split holding the same 2.25 GiB aggregate:

| split | wall | peak RSS |
|---|---|---|
| unlimited | 4.97s | ~650 MB |
| `GOMAXPROCS=8` alone | 5.25s | 611 MB |
| 256MB × 8 | 7.46s | 411 MB |
| **384MB × 5** | **6.92s** | 450 MB |
| 768MB × 2 | 10.38s | 643 MB |
| 1.1GB × 1 | 23.93s | 612 MB |

Two things fall out. Pinning parallelism is nearly free (+5%), so that half of the pair is not what costs. And the tempting fix — fewer, fatter processes so the linker gets a real share — is dramatically *worse*, because it serializes the compile phase. The shape of the split is therefore right; only the floor was wrong, so it moved 256MiB → 384MiB (the measured optimum, and the constant now carries that rationale). What remains is ~40% on a cold dependency-heavy build on a memory-tight worker, paid once per script version since the binary is cached. I did not find a way to give the serial link phase more than the compile phase: `GOMEMLIMIT` is one env var inherited by the whole tree.

**Nit / tripled rationale (claude).** Trimmed. The invariant is stated once at the derivation site (`resolve_go_build_limits`); the merge site points at it and the struct fields just name the variables.

**Coverage (both).** Added a unit test for the env merge: either-half-pinned, `GOGC` alongside, whitespace preserved, empty allowlist entry not shadowing, and no-derived-limits.

One thing I did not change, flagging it since it looks adjacent: a worker-group `GOMEMLIMIT` in a spelling the Go runtime rejects (`512MB`) fails the build. That is not new behavior introduced here — the same value already aborts the run step today — and normalizing only the build side is what produces the confusing outcome of compiling successfully and then dying on the binary it produced. Verified both ways on a real worker before settling on verbatim.

rubenfiszel:
Both nits fixed in 69311fdff7 — the second one turned out to be a real regression I introduced, and measuring it changed the design again.

**The log described values it no longer controlled (both).** Correct, and it was the unaddressed half of the previous round's P1. The log moved from `GO_BUILD_LIMITS` to the merge site and now prints what is actually handed to the toolchain, on change rather than per job (a worker group can move it at runtime, so `Once` would go stale). Verified on a worker with a group pinning `GOMAXPROCS` and nothing else — precisely the configuration that used to lie:

```
Go compilation limited to GOMAXPROCS=3
```

No memory bound claimed, because none is installed. The derived case reads `Go compilation limited to GOMEMLIMIT=402653184 GOMAXPROCS=5`.

**The 384 MiB floor serialized small workers (claude).** This was the more interesting one. You called it defensible; it isn't, and the suggestion to time a 1 GiB budget is what showed why. Same harness, 768 MiB budget (a 1 GiB worker), same aggregate in every row:

| split | wall |
|---|---|
| 384MB × 1 compiler ← what the floor selected | 26.63s |
| 256MB × 2 | 14.37s |
| 192MB × 3 | 10.96s |
| **128MB × 5** | **9.20s** |
| 96MB × 7 | 9.38s |

So the floor was picking the worst split available on exactly the workers this PR is meant to protect — a ~3x regression, not a ~40% one. The curve also flattens at 5 compilers here, and the 2.25 GiB matrix from the last round peaked at 5 compilers too. Both budgets want the same *process count*, which means the constant I had been tuning was the wrong one: the share isn't what the build is sensitive to, the overlap is.

Reworked accordingly — one compiler per core while the budget affords each a comfortable share (384 MiB), never fewer than 6 processes, never more than the cores can run, with a 128 MiB absolute floor under the share:

```rust
let cap = vcpus.max(1) + 1;
let processes = (budget / GO_BUILD_TARGET_MEMLIMIT).clamp(MIN_GO_BUILD_PROCESSES.min(cap), cap);
let memlimit = (budget / processes).max(MIN_GO_BUILD_MEMLIMIT);
```

`MIN_GO_BUILD_PROCESSES.min(cap)` keeps `clamp`'s min ≤ max on few-core workers. This reproduces every measured optimum: 2.25 GiB/24 cores → `384MB × 5`, 768 MiB/24 cores → `128MB × 5`, 3 GiB/4 cores → `614MB × 4` (unchanged), and a generous worker still gets one compiler per core. Verified on a live worker at both budgets; the 1 GiB-equivalent case now logs `GOMEMLIMIT=134217728 GOMAXPROCS=5`.

Test updated for the new splits, including the small-budget case that used to serialize.

rubenfiszel:
Both fixed in 0a9d6b946b. The `GOFLAGS` finding was real — reproduced it before and after.

**P1 / persisted `GOFLAGS` outranks the derived parallelism (codex).** Confirmed exactly as described: `env_clear()` doesn't disable the toolchain's own env file because the command restores `HOME`.

```
$ printf 'GOFLAGS=-p=0\n' > $HOME/.config/go/env
$ HOME=$HOME go env GOFLAGS
-p=0
```

`GOMAXPROCS` only supplies the *default* for `-p`, so a persisted `-p=32` would have run 32 capped compilers against arithmetic that assumed 5. `go build` now passes `-p` explicitly, which is applied after `GOFLAGS` (verified: with `-p=32` persisted, a command-line `-p 0` is what errors).

Proved both directions on a live worker with `GOFLAGS=-p=0` planted in the worker's `HOME`:

- derived pair installed → `-p 5` on the command line → **build succeeds**
- worker group pinning `GOMAXPROCS=many` (non-numeric, so no `-p` is asserted) → **`go: -p must be a positive integer: 0`**

The second is the pre-fix behavior, so it also demonstrates the bypass was reachable rather than theoretical. `-p` is only passed when the installed `GOMAXPROCS` parses as a positive integer — a worker group's value is forwarded verbatim, and a typo should stay as harmless as the Go runtime makes it instead of failing the build.

**Nit / the module steps are single-process (claude).** Agreed and separable, as you said — `go mod init|tidy|download` is one process with nothing else running, so it now gets the whole budget rather than a compiler's share, and no `GOMAXPROCS` pin (nothing to bound, and it only throttles the fetch queue on a network-bound step). On the tested worker:

```
Go dependency resolution limited to GOMEMLIMIT=2415919104
Go compilation limited to GOMEMLIMIT=402653184 GOMAXPROCS=5
```

The log dedupe became a set keyed on the rendered line, since the two steps otherwise alternate every job.

**On the aggregate framing below ~1 GiB** (your "worth knowing"): correct, and the PR description overstated it — under that point the floor wins and the aggregate exceeds the budget. Reworded rather than changed, since the alternative measured worse.

Tests added for the module-step split and for `-p` being asserted only on a numeric value.

rubenfiszel:
Both fixed in 2d627dd02c.

**P1 / `GOMAXPROCS=08` (codex).** Confirmed — the two parsers really do disagree:

```
$ go build -p 08 .
invalid value "08" for flag -p: parse error
$ GOMAXPROCS=08 go version
go version go1.26.5 linux/amd64
```

So copying the operator's spelling into `-p` would have broken a worker-group configuration that works today, on every uncached build. `-p` now carries the *parsed* number rather than the string it arrived in, and parsing is `u32` so a value outside the flag's range falls back to omitting `-p` instead of failing the build. Verified on a live worker with the group pinning `GOMAXPROCS=08`: job compiles and runs. Regression case added.

**Nit / `budget` handed to the module step unfloored (claude).** Correct, and it inverted the exact invariant the floor exists for — below ~171 MiB of worker memory `go mod download` would have been capped tighter than any of the six compilers beside it. Floored at the source (`budget.max(MIN_GO_BUILD_MEMLIMIT)` in the resolver) rather than at the call site, so the rule stays in one place; the resolver test's small-worker case now pins it.

rubenfiszel:
Fixed in c4f9e51610 — and this one was worse than "the flag accepts it", so thanks for pushing on it.

Reproduced both halves. The runtime ignores an out-of-range value:

```
$ GOMAXPROCS=2147483648 ./probe
GOMAXPROCS = 24        # NumCPU, i.e. the setting had no effect
```

and `go build` does not:

```
$ timeout 40 systemd-run --user --scope -q -p MemoryMax=512M go build -p 2147483648 .
exit=137               # SIGKILL — OOM-killed by the cgroup
```

So the code as written would have converted a worker-group setting that is *inert today* into an OOM kill of the worker on the first uncached build — the exact failure this PR exists to prevent, introduced by the fix for the previous round's finding.

`-p` now parses the value as `i32`, which is what the runtime reads, so the flag agrees with `GOMAXPROCS` on both edges: `08` → `-p 8` (the runtime reads 8; the flag would reject the spelling), `2147483648` → no `-p` at all (the runtime ignores it, so the build keeps the default the runtime itself would have used). Regression case added at `i32::MAX + 1`.

Verified on a live worker with the group pinning `GOMAXPROCS=2147483648`: job compiles and runs, no OOM, worker healthy afterwards.

rubenfiszel:
Both fixed in 4095fa70c8. Checked each claim against the runtime first, and one half of the first finding did not hold:

```
$ GOMAXPROCS="8 "  ./probe   →  GOMAXPROCS = 24    # ignored, as you said
$ GOMAXPROCS="+8"  ./probe   →  GOMAXPROCS = 8     # accepted, so no divergence there
$ GOMEMLIMIT=off go version  →  ok                 # `off` is a value Go takes
```

**P1 / whitespace.** Confirmed and fixed. `"2147483647 "` is inert for the runtime but `.trim().parse()` accepted it, so it would have become an authoritative `-p 2147483647` — the same OOM as last round through a different door. The value is now read with the runtime's own grammar (optional sign, ASCII digits, nothing else) rather than trimmed, so anything the runtime discards leaves `-p` off.

The leading `+` half I left as-is: both parsers accept it, so there is nothing to reconcile, and it canonicalizes to `8` on the way to the flag anyway. Test asserts that rather than assuming it.

**P2 / `off` logged as active.** Right, and it applies to `GOMEMLIMIT` too, since Go accepts `off` there as well — the message claimed a bound where the operator had explicitly removed one. The log now reports rather than characterizes: `Go compilation runs with GOMEMLIMIT=off`, and `… runs with no limits` when nothing is installed.

Verified on a live worker in a 3 GiB cgroup: group `GOMAXPROCS="2147483647 "` → job compiles and runs, no `-p`, worker healthy afterwards; group `GOMEMLIMIT=off` → job compiles and runs, logged as above. Regression cases added for the whitespace-padded value and the leading `+`.

rubenfiszel:
Fixed in d632ef77e8 — and the first attempt at it was wrong in a way only the live run caught, which is worth recording.

**The finding is real.** `get_vcpus()` returns the raw quota in µs, so dividing by a hardcoded `100_000` both assumed the period and floored the ratio. A 1500m worker resolved to one vCPU and therefore `GOMAXPROCS=1` — the serialization cliff from round 2, reintroduced through the CPU side.

`worker_vcpus()` now reads the period too (new `get_cpu_period()` in `windmill-common`, mirroring `get_vcpus`'s v1/v2 handling: `cpu.cfs_period_us`, or the second field of `cpu.max`) and rounds the ratio up, as Go's own container-aware `GOMAXPROCS` does.

**The part that needed the worker to catch:** my first version clamped the result with `.min(available_parallelism())`. That looks like the affinity bound Go applies — but Rust's `available_parallelism` is *itself* quota-aware and floors:

```
unconstrained    → 24
CPUQuota=150%    → 1
CPUQuota=250%    → 2
```

So the clamp put back the exact flooring the fix removes, and the job still logged `GOMAXPROCS=1` under a 1.5-vCPU quota. The unit test passed throughout — it takes the CPU count as a parameter, so it could not see it. The quota is now the answer whenever there is one to read, and `available_parallelism` only fills in when there is not.

Verified end to end under `systemd-run -p MemoryMax=3G -p CPUQuota=150%` (cgroup confirmed as `cpu.max: 150000 100000`):

- before: `Go compilation runs with GOMEMLIMIT=1207959552 GOMAXPROCS=1`
- after: `Go compilation runs with GOMEMLIMIT=805306368 GOMAXPROCS=2`

Two compilers plus the driver, 768 MB each, job compiles and runs. Unit tests added for the fractional quota, a non-default period, and the no-quota fallback.

rubenfiszel:
Fixed in 5417b58fc2. Checked the claim against the runtime rather than the doc, and it holds exactly:

```
unconstrained  → GOMAXPROCS = 24
CPUQuota=150%  → 2
CPUQuota=50%   → 2      # floor, not ceil(0.5)
CPUQuota=25%   → 2
```

So a sub-CPU quota was the one case where the pinned parallelism was *stricter* than what Go would have chosen for itself, and it produced fully serial builds — the outcome the whole split is designed to avoid.

`effective_vcpus` now floors at two whenever the machine has two CPUs to give, and stays at one for a genuinely single-CPU worker.

That floor needs a CPU count the quota has not already been folded into, which is the same trap as the last round: `available_parallelism` reports **1** under a 50% quota on this 24-core host, so using it would have made the floor `min(1, 2) = 1` and changed nothing. Added `get_affinity_cpus()` (parses `Cpus_allowed_list` from `/proc/self/status`, i.e. `sched_getaffinity` — what Go's `NumCPU` reports), with `available_parallelism` kept only as a fallback.

Verified under `systemd-run -p MemoryMax=3G -p CPUQuota=50%`: `Go compilation runs with GOMEMLIMIT=805306368 GOMAXPROCS=2`, job compiles and runs, worker healthy. The old code computes `GOMAXPROCS=1` and `GOMEMLIMIT=1207959552` for that same cgroup, so the log doubles as proof of which binary ran. Test cases added for the fractional quota on multi-CPU affinity and on genuine single-CPU affinity.

rubenfiszel:
Fixed in 1fa7fea8c7. Correct: `LIMIT_WINDOWS_TO_1CU` synthesizes `get_vcpus() = 100000` against a real hardware count, so the new floor read that as "one CPU on a multi-core machine" and raised it to two — quietly overriding an explicit cap.

Fixed at the source rather than in the floor: `get_affinity_cpus()` now returns `Some(1)` under the cap on Windows, exactly as its two siblings in that file already do (`get_vcpus` → 1 vCPU, `get_memory` → 2 GB). The cap is a policy, not a bandwidth quota, so every helper reporting the worker's resources should report it; special-casing the floor instead would leave the next consumer of that count to rediscover the same thing. The one-CPU allowance then flows through unchanged and `GOMAXPROCS`/`-p` stay at 1.

Regression case added at `effective_vcpus(Some(100_000), Some(100_000), 1)`.

One caveat I want to be explicit about: **I could not compile the Windows path locally.** `cargo check --target x86_64-pc-windows-{msvc,gnu}` fails in this checkout on a native C dependency (`aws-lc-sys`), not on this code, so the `#[cfg(windows)]` branch is unverified beyond review. It is three lines mirroring the guard pattern already used twice in the same file under the same cfg, so the symbol and shape are established — but CI's Windows build is the first real compile of it, and I have not run a Go job on a Windows worker.

rubenfiszel:
✅ Review round clean @ 1fa7fea8c7cd3c8b62c052cd43cecc7cb5bb874a

rubenfiszel:
Fixed in 9cc7bd0cdb — the finding is right and the wording was worse than imprecise, since it pointed the wrong way for exactly the executors that have a per-job budget today.

Native mode forces 8 workers, and `NATIVE_TAGS` carries `postgresql` and `mysql`, which claim `MAX_SQL_RESULT_SIZE` (15% of worker memory each, peaking near twice that). Eight of those against one limit is ~120% committed before the peak factor — so "sized against the worker process, not a second job" is precisely the assumption a native-tag budget must not make.

The section now splits the two cases: script tags run one job at a time and may size against the worker process alone; native mode is called out as the exception, with its concurrency and the `MAX_SQL_RESULT_SIZE` precedent named so the next budget divides by 8 rather than rediscovering this.

The Go limits themselves are unaffected — `go` is not a native tag, so the one-job premise they rest on still holds — but the general rule was the part being written down, and it needed the exception.

PR #10671, opened 2026-08-12
feat: add EXIT_AFTER_N_JOBS worker mode for environment cleanup

rubenfiszel:
## Summary

Adds `EXIT_AFTER_N_JOBS`, an env variable that shuts the worker process down once it has
executed that many jobs, so that a supervisor (docker restart policy, kubernetes, systemd)
restarts it and no process runs more than that many. It is meant for deployments that cannot
sandbox jobs with nsjail and rely on the process lifetime to isolate one execution from the
next.

A worker restarting every job would otherwise churn the workers list: the worker name carries
a random suffix, and that name is the primary key of `worker_ping`, so every restart registers
a brand new worker and leaves the previous row behind. In this mode the suffix is derived from
the hostname and the worker index instead, so the restarted process reclaims its row: same
entry in the workers list, `started_at` of the first start, and a `jobs_executed` count that
keeps climbing across restarts.

Fixes WIN-2356.

## Changes

- **`EXIT_AFTER_N_JOBS`** (`windmill-common/src/worker.rs`): after N jobs, the worker sends the
  process killpill. It sends rather than breaks so the main loop still drains the same-worker
  jobs it owns — a `same_worker` flow runs to its end, past the limit, instead of parking — and
  the background processor persists the results before the process goes away. The exit code is
  0. A value that does not parse refuses startup rather than silently disabling the mode.
- **Only jobs that age the environment count** (`dirties_worker_env`): the worker's own init and
  periodic scripts (recognized by the path this very worker queues them under, not by their
  routing tag), flow orchestration jobs, noop/identity/unassigned steps, the warmup job, jobs
  rejected before the executor ran, and results served from the cache all leave the budget
  alone, so the worker never burns a restart cycle without having run a user job. Jobs handed to
  a dedicated worker or a flow runner execute outside the main loop and are not counted either;
  the worker warns at startup when it is configured with dedicated workers, or with
  `NUM_WORKERS > 1` (where the shared environment means the first worker to reach the limit
  takes the others down with it).
- **Stable worker name** (`resolve_worker_suffix`): when `EXIT_AFTER_N_JOBS` is set, the suffix
  becomes `{instance_name}-{digest(hostname, worker index)}` instead of
  `{instance_name}-{random}`. The digest covers the full hostname, so two hosts whose names end
  on the same segment do not collide, and folding the index in keeps the name to exactly one
  suffix segment — what the interactive shell tag strips off. Unset (the default), naming is
  unchanged.
- **`WORKER_SUFFIX`** for the one case the hostname cannot disambiguate: several worker
  processes of the same worker group on one host. The label is appended to the stable suffix and
  validated (ASCII letters/digits/underscores, at most 64 characters), and the assembled worker
  name is checked against the 255 characters `worker_ping.worker` holds, so a bad value stops
  the process at startup rather than at the directory it creates or the ping it `expect`s.
- The initial ping returns the row's `jobs_executed` and the worker seeds its counter with it,
  and its `ON CONFLICT` branch — until now effectively unreachable — refreshes everything
  describing the process. `wm_version` in particular: a reclaimed row still reporting the version
  of the process that died would hold the instance-wide `MIN_VERSION` back forever;
  `current_job_id` likewise, which would skew the zombie/OOM diagnostics that read it.
  `started_at` and `jobs_executed` are the only columns carried over.
- Documented in the README env variable table; both variables are in `ENV_SETTINGS` so they show
  in the startup config log.

## Test plan

Verified against a local worker process (`MODE=worker`) run repeatedly against the dev instance,
while the dev standalone backend kept running with the default configuration.

- [x] `EXIT_AFTER_N_JOBS=1` with 3 queued jobs: each process ran exactly one job, exited 0, and
      left the rest queued; all 3 succeeded over 3 runs.
- [x] Worker group with an init script + `EXIT_AFTER_N_JOBS=2`: the init script ran and did not
      consume the budget (init + 2 user jobs, 1 job left queued).
- [x] Continuity: across the restarts, one single `worker_ping` row, `started_at` unchanged from
      the first start, `jobs_executed` climbing — against 5 rows accumulated by the dev backend
      restarting with random names over the same period.
- [x] Hard kill: `SIGKILL` mid-job, row left pointing at that job and stamped with a stale
      `wm_version`, then restart — the reclaimed row came back with `current_job_id` cleared and
      the current `wm_version`, while `started_at` and `jobs_executed` survived.
- [x] 2-step flow with `EXIT_AFTER_N_JOBS=1`: the worker exited after step `a` with the flow
      parked and step `b` queued, and the restarted worker completed the flow.
- [x] Same-worker 2-step flow: both steps ran in the same process (drained past the limit), the
      flow completed before exit, and the row's final count matched the process's own.
- [x] Cache hit: a second run of a `cache_ttl` script logged the cache skip, succeeded, and did
      *not* consume the limit.
- [x] `NUM_WORKERS=2`: both workers got distinct stable names, the warning fired once, and the
      first to hit the limit shut the process down.
- [x] Refused at startup, exit 1: `EXIT_AFTER_N_JOBS=1x`, `WORKER_SUFFIX=slot-a`.
- [x] Unset: the dev backend keeps drawing random suffixes and keeps running after a job.
- [x] `cargo test -p windmill-common --lib utils::tests`,
      `cargo test -p windmill-worker --lib exit_after_n_jobs`,
      `SQLX_OFFLINE=true cargo check` on both crates and the binary.

Docs for the new variables on windmill.dev are not part of this repo and still need an entry.


rubenfiszel:
Round 1 addressed in 226b41ed36.

**codex [P1] stable worker-name collisions for same-host processes** — reintroduced `WORKER_SUFFIX`, which replaces the random part of the name (`wk-{group}-{instance}-{label}`) for hosts running several worker processes of one worker group; the hostname alone cannot tell those apart. The label is confined to a single name segment (`-` is rewritten to `_`, and the worker index is appended as `_{index}`) so it does not shift the interactive shell tag, which is what the previous iteration of this knob got wrong.

**codex [P1] user jobs with reserved tags bypass the limit** (from the pre-push local Codex pass) — the worker's own init/periodic scripts are now recognized by the path this very worker queues them under (`init_script_{worker}`, `periodic_script_{worker}_{ts}`), not by their routing tag alone. New test: `user_jobs_count_whatever_tag_they_are_routed_with`.

**codex [P2] drained same-worker jobs not persisted in the count** — the forced ping is no longer tied to the first `killpill_tx.send()`, so every job counted from the limit onwards (the drained same-worker steps) reaches the row. Verified on a same-worker 2-step flow: the row now ends at 12, matching the process's own final count.

**codex [P2] agent-worker restarts reset the persisted count** — not fixed, documented instead (README + `insert_ping`). The count would have to come back on the agent ping response, whose type is part of the agent↔server wire contract: a server returning a body where agents deserialize `()` breaks every agent older than it, and the initial ping is `expect`ed at worker start. Agent workers do keep their row and `started_at`.

**claude [P2] README overstates what a restart resets** — reworded: a container restart resets the process, not the container filesystem, so caches and `/tmp` survive it. Same row now also states the one-process-per-hostname assumption and points at `WORKER_SUFFIX`.

**claude [P2] pre-execution rejections consume the budget** — jobs cancelled or errored before the executor ran (`canceled_by`, `pre_run_error`) and the `Unassigned*` kinds no longer count.

**claude [P2] reclaimed row keeps the dead process's `current_job_id`** — the `ON CONFLICT` branch now nulls `current_job_id` / `current_job_workspace_id`. Verified by seeding a stale job id on the row and restarting the worker.

rubenfiszel:
Round 2 addressed in ed716da909.

**codex [P1] distinct `WORKER_SUFFIX` values could collapse to one identity** — the label is now validated instead of rewritten: anything outside `[A-Za-z0-9_]` is refused at startup (`Error: WORKER_SUFFIX must only contain ASCII letters, digits and underscores, got 'slot-a'`, exit 1), so the mapping from label to worker name is injective and still confined to one name segment. `create_labelled_worker_suffix` returns a `Result` and both call sites in `main.rs` propagate it; the test covers the rejected shapes.

**codex [P2] cache hits consumed the budget** — `handle_queued_job` now returns a distinct `JobOutcome::CompletedFromCache` for the cached short-circuit, and the counter takes the outcome into account: a job answered from the cache, or one another worker had already completed, does not age the environment. Verified end to end: a second run of a `cache_ttl` script logged `Job skipped because args & path found in cache and not expired`, succeeded, and the worker did *not* hit the limit (it exited via `EXIT_AFTER_NO_JOB_FOR_SECS` instead).

**codex [P2] a final non-dirty drained step was still missing from `jobs_executed`** — the forced ping moved out of the `dirties_env` branch: once the limit is reached, every job the drain runs (Identity/skipped steps included) re-arms it.

**claude [P2] nothing detects two live processes sharing a stable name** — not implemented, and I'd argue the suggested form would misfire: this mode forces a final ping *right before* exiting, so the row a restarting worker reclaims is always seconds old, and a freshness check would warn on every single restart — the normal path of the feature. A false-positive-free version would have to detect a foreign write (e.g. the main-loop ping noticing the row no longer holds what this process last wrote); worth doing if the collision proves to happen in practice, but it does not belong in the hot ping path on speculation.

**claude cosmetic — `WORKER_SUFFIX` README wording** — reworded ("Pins the last part of the worker name, which is otherwise random, so that ..."), and the row now also states the allowed characters.

rubenfiszel:
Round 3 addressed in ff530d678b.

**codex [P1] malformed job limits silently disabled recycling** — a non-empty `EXIT_AFTER_N_JOBS` that does not parse now refuses startup (`validate_worker_lifecycle_env`, called from `windmill_main` before any worker registers). Verified: `EXIT_AFTER_N_JOBS=1x` → `Error: EXIT_AFTER_N_JOBS must be a positive integer (or 0 to disable), got '1x'`, exit 1. `0` still means disabled, as documented.

**codex [P1] labelled worker identities could still collide** — the label is now appended to the stable suffix instead of replacing it, so the digest (over the full hostname *and* the worker index) is what anchors the identity: `wk-default-rub-a71d9_slot_a`. `worker-east-1` and `worker-west-1` with the same label get different digests, and `(slot, 2)` no longer reaches `(slot_2, 1)`. The test asserts all three non-collisions plus the unchanged interactive-shell prefix.

**codex [P2] README overstated isolation above 1** — reworded to "no process ever runs more than that many (set it to 1 for a process per job)".

**claude [P2] `AlreadyCompleted` treated as "nothing ran"** — correct, and a real inversion: that variant is raised when the queue row disappears *while the child is running*, i.e. the strongest case for recycling. `ran_on_this_worker` now excludes only `CompletedFromCache`, with a test pinning all four variants.

**claude [P2] `(label, index)` injectivity** — same fix as codex's second finding: the index rides in the digest rather than being appended.

Manual checks this round: malformed limit refused at startup; `WORKER_SUFFIX=slot_a` produced `wk-default-rub-a71d9_slot_a`; unit tests and `SQLX_OFFLINE=true cargo check` (both crates and the binary) pass.

rubenfiszel:
Round 4 addressed in 7da2f706cc.

**codex [P1] unbounded `WORKER_SUFFIX` could prevent startup** — the label is now capped at 64 characters, alongside the character check, so it cannot push the worker name past the `VARCHAR(255)` primary key or the worker directory's path component. It is refused before the process registers anything, rather than surfacing at the initial ping (which the worker `expect`s). Covered by the existing rejection table.

**codex [P2] the documented limit ignored same-worker draining** — README now says the limit holds "bar the steps of a same-worker flow it has started, which it always finishes", which is what the drain does (and what the round-2 manual run showed: a two-step same-worker flow with a limit of 1 completes in the one process).

**claude — Good to merge**, no findings.

Also ran the hard-kill scenario both reviewers had listed as the last untested corner: `SIGKILL` on a worker mid-job, row left pointing at that job and stamped with a stale `wm_version`, then restart. The reclaimed row came back with `current_job_id` cleared and the current `wm_version`, while `started_at` (from the first start, hours earlier) and `jobs_executed` (14) survived.

rubenfiszel:
Round 5 addressed in f2a2ded885.

**codex [P1] the 64-character cap only bounded the label** — right: `WORKER_GROUP` is unvalidated too, so the assembled name is what has to be checked. `checked_worker_name` now builds `wk-{group}-{suffix}` and refuses anything past the 255 characters `worker_ping.worker` holds, at the single site where the worker name is assembled in `main.rs`, before any directory is created or ping sent. The label cap stays as the targeted error for the common case. New test pins the boundary exactly (a name of 255 is accepted, 256 refused), deriving the group length from the assembled overhead rather than hardcoding it.

**claude — Good to merge**, no findings.

rubenfiszel:
✅ Review round clean @ f2a2ded885d2ebd62f29694de3e7d2c93bd54a16

PR #10673, opened 2026-08-12
feat: auto-build binaries to object storage on deployment

rubenfiszel:
## Summary

Fixes WIN-2355.

Compiled languages (Rust, Go, C#) cache their binary in the instance object store, but only ever build it lazily, on some worker's first run of the script. That worker pays a full `cargo build --release` / `go build` / `dotnet publish` inside the job it was asked to run.

This adds an opt-in instance setting — gated on instance object storage being configured, with an optional worker tag — that compiles a script right after it is deployed and pushes the binary to object storage, so no run ever pays the compile.

## How it works

A deploy queues a **binary-build job** (`JobPayload::BuildBinary`, which runs as a `dependencies` job: it needs the same worker-side toolchain and job dir, but generates no lock and touches no script row). The job re-derives exactly the cache key the run path uses, skips if the object store already has it, otherwise builds and uploads.

It is a separate job rather than inline work because a release build takes minutes and the deploy must not block on it — and because the tag lets the build land on a pool that has the toolchain and matches the platform of the runtime workers (the cache key includes OS/arch, so a binary built elsewhere is never reused).

Both deploy paths queue it:
- deploys that generate a lock, from the dependency job that just wrote it;
- deploys that supply one (CLI, git-sync), from `create_script_internal`, which queues no dependency job at all — pushed on the deploy's own transaction so the build can never read the script version back before it is committed.

## Changes

- **Settings**: `auto_build_binary_on_deploy` (bool) and `auto_build_binary_tag` (string), with `AUTO_BUILD_BINARY_ON_DEPLOY` / `AUTO_BUILD_BINARY_TAG` env fallbacks. Both in the instance-settings UI under Object Storage, and in the declarative instance config. Off by default; inert unless instance object storage is configured.
- **`windmill-queue/src/binary_prebuild.rs`** (new): the single decision point — supported language, non-empty lock, min-version, setting — plus the job it hands back for either call site to push.
- **`JobPayload::BuildBinary`**: maps to `JobKind::Dependencies` but is deliberately excluded from two dependency-job queue policies — it gets its own `build_binary:{ws}/{path}` concurrency key (sharing the lock slot would park the *next* deploy of that path behind a compile) and does not take the dependency-job priority jump ahead of ordinary jobs on its tag.
- **Per-language prebuilds**: `prebuild_rust_binary`, `prebuild_go_binary`, `prebuild_csharp_binary`. Each checks the **object store**, not the local disk, before skipping, and fails if the artifact did not reach the object store — `save_cache` logs and swallows a failed upload, which is fine for a run that has the binary locally but would otherwise latch a permanent skip here.
- **Cache keys unified**: `rust_cache_key` / `go_cache_key` / `csharp_cache_key` are now the single definition used by both the run path and the prebuild, so the two cannot drift into keying different slots.
- **`build_go_binary` extracted** from `handle_go_job` (behavior-preserving; `create_args_and_out_file` now runs after the build instead of before, which the generated `main.go` does not care about — it only opens `args.json` at run time).
- **Min-version gated** (1.789.0): the build job is a `dependencies` job distinguished by a job arg, and a worker predating the arg would run it as a real lock generation.

## Notable judgment calls

- **Scope is Rust/Go/C#** — the languages whose artifact lands in the instance object store. Bun bundles are already prebuilt inline by the dependency job; Java's compile output is a jar directory and was left alone.
- **Build failures surface as failed jobs** on the script's path. A script that locks but does not compile now fails visibly at deploy time instead of silently at first run. Builds inherit the instance job timeout, same as a cold-compile run would.
- **Compiling executes author-controlled build steps** (Cargo build scripts, MSBuild targets, cgo) with exactly the isolation the cold build of a first run has — nsjail for Rust when job isolation is on, none for the Go and C# compilers. Deploying is therefore now the trigger for that, on the configured pool, rather than first run. Called out in the tag setting's description.

## Test plan

Verified end-to-end against a local EE build (`enterprise,private,parquet,rust,csharp`) with instance object storage configured:

- [x] **Go, lock-generating deploy** → dependency job, then build job; object at `linux_x86_64_gobin/<hash>`; run with the local cache cleared logs `loaded bin/bundle from object store` and returns the right result.
- [x] **Go, pre-locked deploy** (CLI/git-sync shape, no dependency job) → build job queued from the create-script path; same object-store load on the run.
- [x] **Rust** → build job, `linux_x86_64_rustbin/<hash>`, run loads from object store instead of compiling.
- [x] **C#** → build job, `linux_x86_64_csharpbin/<hash>`, run loads from object store. (Execution then fails on this box only because it has .NET 10 while Windmill targets `net9.0` — a normal C# run fails identically here.)
- [x] **Multi-file script** — a C# script with a companion `Helper.cs` module builds at deploy time (companion modules are written into the build dir with the same call the run path uses).
- [x] **Build failure is contained** — a C# script that locks but does not compile produces a failed build job carrying the compiler error, and leaves the deployed script row intact (`lock` written, `lock_error_logs` null, not archived).
- [x] **Idempotent** — redeploying identical content returns `{"status":"skipped","reason":"binary already in cache"}`.
- [x] **Tag honored** — with a tag set, the build job queues on it and stays queued until a worker advertises it; unset, it lands on the language tag.
- [x] **Concurrency** — build jobs take `build_binary:{ws}/{path}`, not `dependency:{ws}/{path}`, and carry no priority boost.
- [x] **Off by default** — with the setting off (or absent), no extra job; a Bun deploy never queues one.
- [x] **Env fallback** — `AUTO_BUILD_BINARY_ON_DEPLOY=true AUTO_BUILD_BINARY_TAG=gobuild` with no rows in `global_settings` queues a build job on `gobuild`.
- [x] `cargo check` for `windmill-{types,queue,common,worker,api-scripts}` with `rust,csharp,parquet,enterprise`, and the full binary with `quickjs,enterprise,private,parquet,smtp,rust,csharp`; `npm run check` clean.
- [x] Instance settings UI: both settings render, save, and appear in the declarative-config diff.

No new `sqlx::query!` calls, so the offline cache is unchanged.

## Screenshots

auto-build settings

---
## Summary by cubic
Builds Rust, Go, and C# binaries at deploy time and uploads them to instance object storage to eliminate first‑run compile cost. Previously binaries compiled on first run; now deploys queue a background `dependencies` job marked build‑only, so deploys don’t block and later deploys aren’t parked behind compiles. Addresses WIN-2355.

- Centralizes eligibility in `windmill-queue::binary_prebuild`; queued from dependency jobs and from `create_script_internal` for pre‑locked deploys, pushed within the deploy transaction to avoid races.
- Worker/build path: materializes companion modules before compiling; unifies Rust/Go/C# cache keys across run and prebuild; adds `prebuild_rust_binary`, `prebuild_go_binary`, `prebuild_csharp_binary`; extracts Go build to `build_go_binary` and moves `args.json` generation post‑build without runtime change; skips with a reason if the worker cannot reach object storage and requires `ensure_pushed_to_object_store`.
- Build jobs: separate concurrency key `build_binary:{ws}/{path}` and no dependency priority boost; do not stamp `lock_error_logs`, re‑tally deployments, or restart dedicated workers; min version gate 1.789.0 so older workers don’t treat the job as real lock generation; keeps the build‑job marker internal (not user‑provided and not required on the agent wire).
- Settings/UI: adds instance settings `auto_build_binary_on_deploy` (bool) and `auto_build_binary_tag` (string) with env fallbacks (`AUTO_BUILD_BINARY_ON_DEPLOY`, `AUTO_BUILD_BINARY_TAG`); visible in settings UI and declarative config; off by default; inert on Pro or without instance object storage.

Rollout
- Upgrade all workers to >= 1.789.0.
- Configure instance object storage; feature is inert on Pro or without object storage.
- Enable via `auto_build_binary_on_deploy`; optionally set `auto_build_binary_tag` to route builds to a pool with the required toolchains and matching OS/arch.

<sup>Written for commit 15031f0431cf065794ec6bc413172e8c56dc66ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Known pre-existing limitation (not introduced here)

The binary cache key is derived from the script's main content and lock, not from its companion modules — so editing only a module does not invalidate the cached binary. That is true of the run path today and the prebuild deliberately keys identically; changing it would invalidate every cached binary on upgrade, so it is left for a separate change.


rubenfiszel:
On the two round-2 nits:

**`lock_error_logs` reachable from a build job** — real, fixed in 3473854. The `cache::script::fetch` error arm sat above the build-job early return, so a transient read error during a build would have stamped `lock_error_logs` on a script that deployed fine and surfaced in the editor as a broken deploy. The predicate is now computed before the fetch and guards that one `UPDATE`.

**`MIN_VERSION_SUPPORTS_BINARY_PREBUILD = 1.789.0` "looks one release high"** — dismissing this one: it reads `1.788.0` from the base commit this branch was cut at, but `origin/main` has since taken `chore(main): release 1.788.0` (b39860235c) and `backend/Cargo.toml` on main is now `1.788.0`, so a merge from here ships in 1.789.0. The constant is correct.

I did take the useful half of that finding: `binary_prebuild_job` now logs when the min-version gate blocks a build, checked *after* the setting so it only fires for someone who actually enabled the feature. An enabled setting that silently does nothing was otherwise undiagnosable.

rubenfiszel:
✅ Review round clean @ 3473854a7621e25a22891b9652d0280e326f2740 — nit-only verdicts at d3882ed05e (codex: "Good to merge"; claude: "Mergeable, but should ideally address nits"); nits addressed in 3473854 / dismissed in review reply above.

**On the red `Backend only integration tests` check:** `test_flow_lock_all` is flaky right now and unrelated to this PR. Evidence:

- It panics at the identical line (`tests/worker.rs:4518`) on an unrelated branch, `ruben/win-2354-add-memory-limits-to-go-build-subprocess` (run 31639396767).
- It passed on this branch at 4820340b, and passed on a rerun of the identical commit d3882ed0.
- The test waits for only the *first* completed job and then immediately asserts every flow module has a lock; the module left unlocked is always the Python one, whose uv lock generation is the slowest, so it loses the race under load.
- Nothing in this diff runs in that path: flow module locking goes through `handle_flow_dependency_job` / `capture_dependency_job`, neither of which this PR touches, and `binary_prebuild_job` returns before any DB access for Python.

rubenfiszel:
Merged `origin/main` (conflict was in `go_executor.rs`: main's Go toolchain memory/parallelism limits from #10666 land inside the exact build step this branch extracted into `build_go_binary` — resolved so both survive, verified by `test_go_job` passing locally).

## `cargo_test` failure: root-caused, not related to this PR

I reproduced it locally and it is **not** a race or a regression here — `test_flow_lock_all` depends on three external registries and any one of them failing sinks it.

When a module's lock generation fails, `lock_modules` records the error, writes the flow **with `lock: null` on that module**, and the dependency job then fails. The test waits for that job, ignores whether it succeeded, and asserts every module has a lock — so a registry failure surfaces as a bare module dump with no hint of the cause. That's why the CI failure was unreadable.

The fixture's modules pull `wmill` from PyPI, `deno.land/x/windmill@v1.50.0` (→ `std@0.165.0`), and `rsc.io/quote` + the Go client. In CI the Python module was the one that lost; locally it was deno, and I confirmed the cause is external:

```
$ curl -sS --compressed https://deno.land/std@0.165.0/node/internal/buffer.mjs
curl: (61) Unrecognized or bad HTTP Content or Transfer-Encoding
```

`deno.land` is serving that file with a corrupt brotli encoding right now — curl rejects it, which is exactly deno's `brotli error`.

## What I changed

1. **Report the real error.** The test now checks the dependency job succeeded and panics with the job's own error, so the next failure names the registry and the module instead of dumping a struct.
2. **Retry once** under a fresh path, absorbing a momentary registry blip. A real breakage fails both attempts and reports itself.
3. **Moved the deno module off `deno.land/x`**, which is being sunset, to `jsr:@std/encoding` — a maintained registry. Any remote specifier exercises the same lock-generation path.

Before (3) the test failed deterministically on this machine; after, it passes — 3/3 runs, plus `test_go_job` green.

PR #10674, opened 2026-08-13
fix: close SSRF bypasses in git URL validation (fail-open DNS, redirects)

rubenfiszel:
## Summary

`validate_git_url` in `backend/windmill-store/src/resources.rs` guards the `git ls-remote` probes against SSRF, but two bypasses let a caller reach an address the guard never approved. Both are reachable from the authenticated `get_git_commit_hash` route (`resources:read`) and from the background auto-pull poller running as `SUPERADMIN_SYNC_EMAIL`, and git's stderr is returned to the caller, so each doubles as a status-code oracle.

1. **Fail-open on DNS error.** The hostname branch was `if let Ok(addrs) = lookup_host(..)`, so a resolution error silently skipped the private-IP check entirely. A host that does not resolve at check time went straight through to git. `windmill-common/src/ssrf.rs` already fails closed here (`map_err(..)?` plus an explicit empty-answer rejection); this now mirrors it.

2. **Redirects followed.** All three probes ran with git's default `http.followRedirects=initial`, which still follows the *initial* request's redirect — exactly the vector. A validated public host could 302 the probe onto a private or link-local address (e.g. `169.254.169.254`) with nothing re-validating the new target. `windmill-common/src/ssrf.rs` documents the same house rule for HTTP clients (validate, then use a no-redirect client).

Fixes WIN-2357.

## Changes

- `validate_git_url` fails closed on DNS: a `lookup_host` error and an empty answer are both rejected, matching `validate_url_for_ssrf`'s shape and wording.
- New `git_probe_command()` builds every remote probe with `-c http.followRedirects=false`; all three `ls-remote` call sites (`get_repo_latest_commit_hash`, `get_git_repo_head_for_autopull`, `get_git_repo_fork_heads_for_autopull`) go through it.
- `run_git_probe_for_url()` retries once against the `.git` form when the remote answers the URL as given with a redirect. This is the one legitimate redirect that occurs in practice, and refusing redirects outright would otherwise break it — see below. The retry only extends the path, so it cannot leave the host `validate_git_url` already cleared, which is precisely what following the redirect would not guarantee.
- `git_probe_stderr()` decodes the three probes' stderr in one place and, for a redirect the retry could not recover, appends the remedy: point the resource at the address the remote redirects to.

## The redirect trade-off, measured

Refusing redirects is a behavior change for remotes that rely on them, so I checked which real ones do (`curl` against `info/refs?service=git-upload-pack`):

| Case | Behavior | Status here |
|---|---|---|
| `https://gitlab.com/group/project` (no `.git`) | **301** → `.../project.git` | **Recovered** by the `.git` retry — verified end-to-end |
| `https://github.com/org/repo` (no `.git`) | 200, no redirect | Unaffected |
| Renamed repos (`denoland/deno_std`, `facebook/jest`) | 200, no redirect — GitHub serves the old path directly | Unaffected |
| `http://host/repo.git` upgraded to https | 301 (GitHub and GitLab both) | Fails with the remedy in the message |

The `.git`-less GitLab form is the one that would have silently broken existing auto-pull setups (it is what GitLab's address bar gives you), which is why it is recovered rather than only documented. The remaining break is a plain `http://` remote that upgrades to https; following that redirect is the hole being closed, so the fix is to store the https URL, and the error says so.

## Deliberately not in scope

DNS-rebinding pinning. The resolved addresses are now materialized but still discarded, so git re-resolves at connect time; closing that needs `http.curloptResolve` keyed on the URL's real port (the validation lookup hardcodes `:443`). Pre-existing, tracked separately.

## Test plan

- [x] `cargo test -p windmill-store --lib` — 26 pass, including four new tests (redirect refused, `.git` retry, `dot_git_url` only ever extends the path, DNS fail-closed).
- [x] Mutation-checked: deleting the `-c http.followRedirects=false` arg fails the redirect test in 0.01s on "git followed the redirect to the unvalidated target"; the DNS test fails without the `map_err`; the `no_proxy` env in the fixture is load-bearing (verified the test fails without it under `http_proxy=http://127.0.0.1:1`).
- [x] Redirect behavior confirmed against real git 2.55 outside the tests: default follows the 302 and hits the target; with the flag git stops and the target logs no request.
- [x] End-to-end on a running backend via `GET /api/w/<ws>/resources/git_commit_hash/<path>`:
  - `https://gitlab.com/gitlab-org/cli` (no `.git`, the 301 case) → `200 {"commit_hash":"8198d4c8…"}`
  - `https://gitlab.com/gitlab-org/cli.git` → same sha
  - `https://github.com/windmill-labs/windmill` (no `.git`) → `200`
  - `http://github.com/windmill-labs/windmill.git` → `400 … error: 301 (the remote redirects, and redirects are not followed; set the repository URL to the address it redirects to)`
  - unresolvable host → `400 Failed to resolve git URL host '…'`
- [x] `cargo check` clean.

### DIFF
--- a/backend/windmill-store/src/resources.rs
+++ b/backend/windmill-store/src/resources.rs
@@ -3279,10 +3279,15 @@ fn git_probe_stderr(stderr: Vec<u8>) -> String {
 fn dot_git_url(url: &str) -> Option<String> {
     let url = url.trim().trim_end_matches('/');
     let lower = url.to_lowercase();
-    if !lower.starts_with("http://") && !lower.starts_with("https://") {
-        return None;
-    }
-    if lower.ends_with(".git") || url.is_empty() {
+    let after_scheme = lower
+        .strip_prefix("https://")
+        .or_else(|| lower.strip_prefix("http://"))?;
+    // Only a non-empty path may be extended. With nothing after the authority,
+    // appending lands inside it instead — `https://host` becomes the unvalidated
+    // host `https://host.git`, `https://host:8443` a bogus port — which is the hop
+    // onto an unchecked address that refusing redirects exists to prevent.
+    let path = after_scheme.split_once('/')?.1;
+    if path.is_empty() || path.ends_with(".git") {
         return None;
     }
     Some(format!("{}.git", url))
@@ -4117,6 +4122,15 @@ mod tests {
             dot_git_url("https://gitlab.com/group/project/").as_deref(),
             Some("https://gitlab.com/group/project.git")
         );
+        assert_eq!(
+            dot_git_url("https://user:tok@gitlab.com:8443/group/project").as_deref(),
+            Some("https://user:tok@gitlab.com:8443/group/project.git")
+        );
+        // A pathless URL has only its authority to extend, so `example.com` would
+        // become the unvalidated host `example.com.git`. Never retry those.
+        assert_eq!(dot_git_url("https://example.com"), None);
+        assert_eq!(dot_git_url("https://example.com/"), None);
+        assert_eq!(dot_git_url("http://example.com:8080/"), None);
         // Nothing to retry: already the `.git` form, or no HTTP redirect surface.
         assert_eq!(dot_git_url("https://gitlab.com/group/project.git"), None);
         assert_eq!(dot_git_url("git@github.com:user/repo"), None);
--- a/backend/windmill-store/src/resources.rs
+++ b/backend/windmill-store/src/resources.rs
@@ -3250,6 +3250,23 @@ fn git_probe_command() -> Command {
     git_cmd
 }
 
+/// Decode a failed probe's stderr. A remote that answers with a redirect (a plain
+/// `http://` URL upgraded to https, say) only surfaces as a bare curl status here,
+/// since `git_probe_command` refuses to follow it — name the remedy, which is to
+/// point the resource at the address the remote redirects to.
+fn git_probe_stderr(stderr: Vec<u8>) -> String {
+    let stderr =
+        String::from_utf8(stderr).unwrap_or_else(|_| "Failed to decode stderr".to_string());
+    if stderr.contains("returned error: 30") {
+        format!(
+            "{} (the remote redirects, and redirects are not followed; set the repository URL to the address it redirects to)",
+            stderr.trim_end()
+        )
+    } else {
+        stderr
+    }
+}
+
 async fn run_git_probe(mut git_cmd: Command, what: &str) -> Result<std::process::Output> {
     git_cmd.kill_on_drop(true);
     match tokio::time::timeout(GIT_PROBE_TIMEOUT, git_cmd.output()).await {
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Thanks — addressed all three nits, and the third one turned out to be half right in a way that mattered.

**Nit 3 (redirect breakage is wider than described).** I measured it rather than reasoning about it, and the renamed-repo claim doesn't hold: GitHub serves renamed repos' `info/refs` with a **200**, no redirect (checked `denoland/deno_std`, renamed to `denoland/std`, and `facebook/jest`), and so does `/repo` without the `.git` suffix. But **gitlab.com 301s `.../group/project` → `.../group/project.git`**, and that one is real — `https://gitlab.com/gitlab-org/gitlab` is what GitLab's address bar hands you, so refusing redirects would have silently stopped auto-pull for those resources. Confirmed with real git:

```
$ git -c http.followRedirects=false ls-remote https://gitlab.com/gitlab-org/cli HEAD
fatal: unable to access '...': The requested URL returned error: 301
```

That's more than a changelog line, so it's fixed rather than documented: `run_git_probe_for_url()` retries once against the `.git` form when the remote answers with a redirect. The retry only ever extends the path, so it stays on the host `validate_git_url` already cleared — which is exactly the guarantee following the redirect doesn't give. Verified end-to-end through `git_commit_hash`: the `.git`-less GitLab URL now returns `200 {"commit_hash":"8198d4c8…"}` with redirects still refused, and an `http://` remote that upgrades to https still fails, now with the remedy in the message. The PR body has the full measurement table.

**Nit 1 (hint pinned by no test).** `test_git_probe_refuses_redirects` now asserts on `git_probe_stderr(output.stderr)` and checks for both the status and the remedy text, so rewording either side fails the test.

**Nit 2 (tests depend on the resolver).** The fail-closed test now says so in its failure message ("does this resolver synthesize records for NXDOMAIN?") instead of reading like a validation bug, and `test_validate_git_url_allows_valid_urls` carries a comment that it needs outbound DNS. That dependency is inherent to fail-closed, so I kept it rather than mocking the resolver.

**Very minor (stack array across `.await`).** Fixed — `vec![0u8; 4096]`.

**Informational (`ansible_executor.rs`).** Agreed it's not a finding, and I've dropped the "no `Command::new(\"git\")` anywhere else in `backend/`" claim from the PR body since it implied more coverage than it had.

**DNS-rebinding pinning** stays out of scope, called out explicitly in the body.

rubenfiszel:
Good catch, and it was worse than the report — fixed in affd4d05b1.

`dot_git_url` now requires a **non-empty path** after the authority; anything else returns `None` and no retry happens. Writing the regression case for it immediately turned up a second variant the host-equality check I first tried would have let through:

```
dot_git_url("http://example.com:8080")  ->  Some("http://example.com:8080.git")
```

The host compares equal there (`extract_host_from_git_url` splits at the last `:`), so only the structural "must have a path" condition rejects both shapes. That is what shipped:

```rust
let path = after_scheme.split_once('/')?.1;
if path.is_empty() || path.ends_with(".git") {
    return None;
}
```

Regression cases added for `https://example.com`, `https://example.com/`, and `http://example.com:8080/`, plus a credentials-and-port case (`https://user:tok@gitlab.com:8443/group/project`) confirming the retry still only extends the path when there is one.

Verified end-to-end through `git_commit_hash` on a running backend:

- `https://gitlab.com/gitlab-org/cli` (has a path, 301s) → `200 {"commit_hash":"8198d4c8…"}` — retry still works
- `https://gitlab.com` (pathless, 302s) → `400 … error: 302 (the remote redirects, and redirects are not followed; …)` — the retry does **not** fire, so `gitlab.com.git` is never dialed

26 unit tests pass, `cargo check` clean.

rubenfiszel:
✅ Review round clean @ affd4d05b18517c37c844e2aea431fb8cef6114f

PR #10677, opened 2026-08-13
fix: expand AZURE_DEVOPS_TOKEN placeholder in backend git probes

rubenfiszel:
## Summary

Git-sync repository URLs may carry `AZURE_DEVOPS_TOKEN(<path/to/azure/resource>)` where a
credential belongs. The hub sync scripts (`hub/28871` push, `hub/28890` pull/init,
`hub/28216` repo-viewer clone) expand it in TypeScript by minting an Azure DevOps access
token from that `azure` resource's client credentials.

The backend's own git probes shell out to `git` directly and never learned the convention,
so they hand git the literal placeholder. Its embedded `/` truncates the authority and curl
rejects the result:

```
head check failed: Bad request: Error getting git repo commit hash: fatal: unable to access
'https://AZURE_DEVOPS_TOKEN(f/…/…)@dev.azure.com/…': URL rejected: Bad hostname
```

For an Azure DevOps repository configured this way, deploy→git push and manual pull work,
but automatic pull polling fails every tick and the message lands in the repository card as
"Last sync failed". This teaches the backend probes the same expansion.

## Changes

- `resolve_azure_devops_url` expands the placeholder: reads the referenced `azure` resource
  through the caller's own `DbWithOptAuthed`, then mints a token against the same endpoint,
  grant and resource id the hub scripts use. A missing resource, a missing
  `azureTenantId`/`azureClientId`/`azureClientSecret`, or a rejected token request each give
  a specific error instead of curl's "Bad hostname".
- Wired into all four probes: the auto-pull head check, the branch-less `ls-remote --symref`
  default-branch resolution, fork-branch head listing, and the authed `git_commit_hash`
  endpoint. Expansion runs *before* `validate_git_url`, so the SSRF and host checks now see
  the real authority rather than a malformed one.
- The placeholder is rejected unless it occupies the userinfo of an http(s) URL. Elsewhere
  the minted token would be spliced into a component git echoes verbatim in failure
  messages and that credential redaction does not cover.
- Minted tokens are cached in-process keyed by a **digest of the credentials**, never the
  resource path, so a cache hit cannot serve a token to a caller who could not read the
  resource. Without it every poll tick would mint a fresh token.
- `redact_git_url_credentials` strips userinfo from the two probe errors that interpolated
  the raw URL (`No commits found for reference '…' in repository '…'`, `No HEAD found in
  repository '…'`). Those strings are persisted as `last_pull_status.error` and rendered in
  the UI, so for any token-bearing repo — a `$var:` PAT today, a minted token now — they
  leaked a live credential.

## Notes for review

- **Not covered:** `windmill-worker/src/ansible_executor.rs:933` has the same gap for ansible
  git repos. Different crate and auth model (worker-side `AuthedClient`, not a DB handle);
  deliberately left out of this PR.
- The authed `git_commit_hash` endpoint checks `resources:read:<git repo path>` scope, but
  the `azure` resource it now reads is not separately scope-checked. RLS via `UserDB` still
  applies, and this matches how nested `$var:`/`$res:` interpolation already behaves —
  diverging here would be inconsistent.
- Endpoint, grant and resource id are copied verbatim from the hub scripts so a repository
  that authenticates for a sync job authenticates for these probes too. Sovereign clouds
  (Azure Government/China) are not supported, matching the hub scripts.

## Test plan

- [x] `cargo check -p windmill-store --features private`
- [x] `cargo test -p windmill-store --features private azure_devops` — placeholder parsing:
      resource path containing `/`, no placeholder, unterminated, in the path (rejected),
      non-http scheme (rejected), `user:token` userinfo (accepted)
- [ ] **Not exercised at runtime.** No Azure DevOps repository or Azure AD app registration
      was available, and the auto-pull poller is `#[cfg(feature = "private")]` behind an
      Enterprise license check. The token exchange itself has never run — it is copied from
      the hub script that already performs it in production. Needs a manual pass against a
      real Azure DevOps repo through both the commit-hash lookup and the auto-pull head
      check before merge.

---
## Summary by cubic
Expands AZURE_DEVOPS_TOKEN(...) in backend git probes and redacts credentials from probe output. Previously probes sent the literal placeholder and failed with “Bad hostname”; now we mint an Azure DevOps token, substitute it, validate the URL, confine the host, and scrub secrets.

- Applies to: auto-pull head check, default-branch resolution (`ls-remote --symref`), fork-branch head listing, and the authed commit-hash endpoint. Expansion runs before URL validation; `validate_git_url` rejects unexpanded placeholders.
- Placeholder rules: only allowed as HTTPS userinfo and only for Azure DevOps hosts (`dev.azure.com` or `*.visualstudio.com`). Rejected elsewhere with a specific error.
- Token minting: Azure AD client-credentials flow; in-process cache keyed by a digest of credentials with expiry margin and fallback lifetime. No mint when the destination URL would be rejected.
- Authorization: enforced at write time on `create_resource`, `update_resource`, and `set_resource_value`. Only workspace admins can configure `AZURE_DEVOPS_TOKEN(<path>)`, and the referenced `azure` resource must exist and be readable; errors name the referenced path.
- Error hygiene: scrub userinfo from git stderr and redact credentials in persisted messages (“No commits found…”, “No HEAD…”).

Action: For Azure DevOps repos, put the placeholder in the HTTPS credentials segment to `dev.azure.com` or `*.visualstudio.com`, for example: https://AZURE_DEVOPS_TOKEN(path/to/azure/resource)@dev.azure.com/…; only workspace admins can set this, and they must have read access to the referenced `azure` resource.

<sup>Written for commit d4ae33bfd6668f03fef00367313c524cc2392218. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/asset_trigger_dispatch.rs
+++ b/backend/tests/asset_trigger_dispatch.rs
@@ -244,6 +244,7 @@ fn make_mini(id: Uuid, runnable_path: &str) -> MiniCompletedJob {
         cache_ttl: None,
         cache_ignore_s3_path: None,
         runnable_settings_handle: None,
+        build_binary_only: false,
     }
 }
 
--- a/backend/windmill-store/src/resources.rs
+++ b/backend/windmill-store/src/resources.rs
@@ -3623,7 +3623,11 @@ pub async fn authorize_azure_devops_reference(
         return Ok(());
     };
 
-    require_admin(authed.is_admin, &authed.username)?;
+    if !authed.is_admin {
+        return Err(Error::PermissionDenied(format!(
+            "Only a workspace admin can point a git repository URL at AZURE_DEVOPS_TOKEN({resource_path}): background sync mints that credential under an identity that bypasses resource permissions"
+        )));
+    }
 
     let dba = DbWithOptAuthed::from_authed(authed, db.clone(), Some(user_db.clone()));
     let readable =
--- a/backend/windmill-store/src/resources.rs
+++ b/backend/windmill-store/src/resources.rs
@@ -1140,6 +1140,19 @@ async fn create_resource(
     }
     let authed = maybe_refresh_folders(&resource.path, &w_id, authed, &db).await;
 
+    authorize_azure_devops_reference(
+        &authed,
+        &db,
+        &user_db,
+        &w_id,
+        resource
+            .value
+            .as_deref()
+            .and_then(|v| serde_json::from_str::<serde_json::Value>(v.get()).ok())
+            .as_ref(),
+    )
+    .await?;
+
     let mut tx = user_db.begin(&authed).await?;
 
     let update_if_exists = q.update_if_exists.unwrap_or(false);
@@ -1821,6 +1834,18 @@ async fn update_resource(
     sqlb.returning("path");
     let authed = maybe_refresh_folders(path, &w_id, authed, &db).await;
 
+    authorize_azure_devops_reference(
+        &authed,
+        &db,
+        &user_db,
+        &w_id,
+        ns.value
+            .as_deref()
+            .and_then(|v| serde_json::from_str::<serde_json::Value>(v.get()).ok())
+            .as_ref(),
+    )
+    .await?;
+
     let mut tx = user_db.begin(&authed).await?;
 
     if let Some(npath) = ns.path.clone() {
@@ -2098,6 +2123,8 @@ async fn set_resource_value(
     {
         return Err(Error::PermissionDenied(msg));
     }
+    authorize_azure_devops_reference(authed, db, user_db, w_id, value.as_ref()).await?;
+
     let mut tx = user_db.clone().begin(authed).await?;
 
     // `RETURNING resource_type` rather than a second lookup: the advisory below has to know the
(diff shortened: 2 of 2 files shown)


rubenfiszel:
✅ Review round clean @ d4ae33bfd6668f03fef00367313c524cc2392218 — nit-only verdicts at b54c8dc7c074209085a805c4a389d55eccd1db66 (codex: Good to merge; claude: mergeable, nits); nit addressed in d4ae33bfd6.

Note for whoever merges: the Azure token exchange has still never been executed. Every round here was static review. The unchecked box in the test plan needs a real Azure DevOps repo and app registration before merge.

PR #10686, opened 2026-08-13
fix(groups): replace instance-group delta-patching with a state-based reconciler

alpetric:
## Problem

Workspace membership granted via instance groups was maintained by hand-written delta patches scattered across seven mutation paths (instance-group API, workspace settings, SCIM). Each path re-implemented its own notion of who should be added, re-roled, or removed, and several got it wrong:

1. **Dangling references**: deleting an instance group (`DELETE /api/groups/delete/{name}` or `POST /api/groups/overwrite`) left its name in every workspace's `auto_invite.instance_groups[_roles]`. A group created later with the same name silently re-acquired the mapping.
2. **No user cleanup at all** on those two paths — members stayed in workspaces after their group was deleted.
3. **Data loss on membership shrink** — pre-existing in shipped code, see below.
4. **SCIM `delete_group`** kept its own copy of the reference-stripping SQL.
5. An overwrite import that retained a group but dropped members never cleaned those members up, and a member removed from their highest-precedence group kept a stale elevated role.

## Design

Instance-group workspace membership is a projection of two base tables: `email_to_igroup` (who is in which group) and `workspace_settings.auto_invite.instance_groups[_roles]` (which groups a workspace auto-assigns, at which role). The desired state per workspace, for users with `added_via.source = 'instance_group'`:

- member **iff** they belong to ≥ 1 group configured for that workspace
- role = highest precedence among those groups (admin > developer > operator)
- `added_via.group` = the group granting that highest role

One EE function, `reconcile_workspace_instance_groups(workspace_ids, tx, audit_author)`, computes that state and applies the diff: adds missing members (`auto_add_user`), fixes role/`added_via` drift in place (`update_workspace_user_internal`), and removes no-longer-qualified members (`delete_workspace_user_internal`). Users with any other `added_via.source` (manual, domain, service_account) are invisible to reconciliation.

Every mutation path becomes: mutate base tables → collect affected workspace ids → reconcile. Collapsed paths:

- `delete_igroup`, `overwrite_igroups`, `remove_user_igroup`, `add_user_igroup` (this repo)
- `edit_instance_groups` (EE)
- SCIM `update_group` member adds/removals (full-replace, patch-add, patch-remove, filter-remove) and `create_group` (EE)
- the SCIM reconcile-orphans pass (`reconcile_instance_group_membership`) is deleted — the reconciler subsumes it

The delta helpers `cleanup_removed_instance_groups`, `remove_users_from_instance_group_workspaces`, `process_instance_group_auto_adds` and `process_single_instance_group_user` are deleted. SCIM `delete_group` now uses the shared `remove_instance_groups_from_workspace_settings` helper instead of its own JSONB SQL.

**Note the no-downgrade rule is not a special case**: `role = max(groups)` does not decrease when a lower group is added and does decrease when the top group is removed. The existing `test_no_role_downgrade_when_added_to_lower_group` passes untouched.

**Renames**: SCIM group renames previously stranded the old name in `workspace_settings` references (inert under the delta model, but the state reconciler would read a stale name as "no members" and evict everyone granted through it on the next reconcile). Both SCIM rename sites now follow the rename through workspace references via a new `rename_instance_group_in_workspace_settings` helper and reconcile.

**Concurrency**: no global advisory lock. The reconciler takes one transaction-scoped advisory lock per workspace (in sorted order), so concurrent mutations reconciling the same workspace serialize, and the later transaction re-derives state from the earlier one's committed writes.

**cfg gates**: the reconcile calls are gated on `private` alone (not `enterprise`) — CE builds ship `private` without `enterprise`, and gating cleanup on `enterprise` would strand CE users. The previous `add_user_igroup`/`remove_user_igroup` sync blocks were gated on `all(private, enterprise)` and are now `private`.

**Upgrade path**: instances that hit the old bugs carry members whose granting group no longer exists, in workspaces still referencing that name — inert under the delta model, but the state reconciler would evict them destructively on the first reconcile. A reversible migration (`preserve_orphaned_instance_group_members`) converts those members to `added_via.source = 'manual'` (original group kept under `migrated_from_instance_group`) and strips the dangling references, so nothing is lost at upgrade.

## Pre-existing data-loss bug (shipped code)

`remove_users_from_instance_group_workspaces` deleted the workspace user whenever `added_via.group` named the removed group — a field that only records the member's *highest-precedence* group — and callers then re-added still-qualifying members. The deletion (`delete_workspace_user_internal`) destroys the member's drafts, favorites, inputs, captures, tokens, folder ownership, `usr_to_group` rows and `extra_perms`. This was reachable from `edit_instance_groups`, `remove_user_igroup`, and three SCIM paths: any member who still qualified through another configured group was deleted and re-created with all workspace-scoped state destroyed. This PR stops further loss by updating still-qualifying members in place; it cannot recover state already destroyed.

## Test plan

All run with `--features private,enterprise`:

- [x] `test_delete_instance_group_preserves_access_via_other_group` — deleting the top group demotes a dual-group member in place (favorite + draft survive), evicts the single-group member, and strips both `auto_invite` fields
- [x] `test_remove_user_from_instance_group_rederives_role` — a member dropped from an admin group loses `is_admin` and is re-roled from the group they still belong to
- [x] `test_overwrite_igroups_preserves_moved_member_data` — a member moved from a dropped group to a retained one in one import keeps their workspace data
- [x] `test_overwrite_igroups_reconciles_retained_group_membership` (new) — an import that retains groups but drops members demotes/evicts accordingly, in place
- [x] `test_group_endpoints`
- [x] `test_scim_group_rename_follows_workspace_references` (new, EE) — a SCIM rename rewrites workspace references and re-points members' `added_via`, so no later reconcile evicts them
- [x] full `instance_group_auto_add` suite (13 tests), including the untouched `test_no_role_downgrade_when_added_to_lower_group`
- [x] `SQLX_OFFLINE=true cargo check` with no features / `private` / `enterprise,private`

Companion EE PR: windmill-labs/windmill-ee-private#726

Supersedes #10458 (kept open as reference; its review history documents the bugs above).

https://claude.ai/code/session_01TYi5MGYwUL6gjYMgQVY2Yx


alpetric:
Addressed the review in 232fc1da75:

- **P1 (upgrade-path eviction)**: added reversible migration `20260813195023_preserve_orphaned_instance_group_members`. Members whose granting group no longer exists in `instance_group` are converted to `added_via.source = 'manual'` (keeping the original name under `migrated_from_instance_group` so admins can prune deliberately), and dangling names are stripped from `auto_invite.instance_groups[_roles]`. Verified the transform empirically against fabricated pre-fix state (orphan converted, live member and live reference untouched). The new `test_reconcile_ignores_non_instance_group_members` pins the protective property: non-`instance_group` sources — including migrated rows — are never re-roled or removed by reconciliation.
- **P2 (JSON-`null` roles)**: the roles expression in `remove_instance_groups_from_workspace_settings` is now guarded with `jsonb_typeof(...) = 'object'`.
- **P2 (repeated rationale)**: the `private`-not-`enterprise` note now lives once, on `workspaces_referencing_instance_groups` (the shared entry point); call sites keep only their site-specific ordering constraints.
- **P2 (per-user mutation cost)**: accepted as the cost of the state model — admin-action rates and one settings row per workspace keep it bounded, and bulk membership syncs should go through `overwrite`/SCIM full-replace (one reconcile per request) rather than N single-user calls. No code change.
- **`ee-repo-ref.txt`**: confirmed `fcf923c` is the head of the EE companion branch (based on the previous ref `2ba6a2a`), not EE `main`.
- The concurrency suggestion is left as-is: deterministic cross-transaction interleaving isn't practical in this harness; the lock-ordering constraint is documented on the reconciler.

alpetric:
Both follow-ups addressed in e0b29eda71:

- **P1 (retained-group orphans)**: the migration predicate is now the full set the reconciler would evict — members with no `email_to_igroup` row in any group their workspace configures — which subsumes the deleted-group case and the retained-group membership drops. Verified empirically on fabricated state covering all three classes: deleted-group orphan and retained-group-dropped orphan both convert to manual (with `migrated_from_instance_group`), a still-qualifying member is untouched. The SRF is applied through a `jsonb_typeof` CASE so a JSON-`null` `instance_groups` cannot abort the migration.
- **P2 (test doesn't pin the guard)**: the fixture row now uses username `shielded_legacy` (differs from the instance-derived `shielded`, so an unguarded `auto_add_user` would create a second row instead of no-op'ing on a username collision), and the test asserts `count(*) = 1` for the email plus that the surviving row is the original one. On the EE side the guard exists — the reconciler's add branch checks `EXISTS (SELECT 1 FROM usr WHERE workspace_id = $1 AND email = $2)` irrespective of source before calling `auto_add_user`, same contract as the code it replaced — and the strengthened test now fails if that check is ever removed.

alpetric:
Third-pass nits addressed in the latest commit:

- **Statement order**: the reference strip now runs before the member conversion, so the conversion predicate reads the same post-strip config the reconciler will — a dangling name with stray `email_to_igroup` rows can no longer spare a member from conversion and then strand them.
- **Migration test**: `test_preserve_orphaned_members_migration` fabricates all three member classes (deleted-group orphan, retained-group-dropped orphan, still-qualifying member) and re-executes the migration's statements via `raw_sql`, asserting both conversions, the spared member, and the reference strip.
- **Audit**: documented on both settings helpers that the mutation is instance-scoped and deliberately recorded by the caller's global igroup audit event rather than per workspace.

On the scope caveat: the EE side (reconciler internals, SCIM conversions, `edit_instance_groups`) was reviewed by a separate fresh-context pass with the EE worktree checked out at the companion branch — it confirmed the source-agnostic add-branch guard and the sorted lock ordering, and its findings (SCIM rename reference-stranding) were fixed in windmill-labs/windmill-ee-private#726 with a regression test.

alpetric:
Fourth-pass nit addressed: the migration test now seeds `ghost_grp` — referenced by the workspace, carrying a membership row, but with no `instance_group` row — which is exactly the shape that discriminates the two statement orders. With strip-first, ghost@ is converted and preserved (asserted); with convert-first they would be spared on the doomed reference and stranded, so the reordering is now guarded.

alpetric:
Codex round addressed in b937a214b7 (+ EE c96b7af):

- **P1 (lock ordering)**: confirmed real — `delete_igroup` acquired `password` row locks (instance-role recompute) before the reconciler's per-workspace advisory lock, while `add_user_igroup` and the SCIM member arms did the reverse, still holding the transaction-scoped advisory lock while touching `password`. Every mutation path now makes reconciliation its **last locking step** (instance-role writes first), and the SCIM multi-op PATCH collects touched groups and reconciles once after the op loop, so interleaved role writes between operations can't reintroduce the inversion. The invariant is documented on `reconcile_workspace_instance_groups` itself, which is the place someone would break it. On the suggested barrier-controlled concurrency test: deterministic cross-transaction interleaving isn't practical in this harness, so the ordering is pinned by contract + review rather than by test.
- **P1 (authorization contract)**: already present — the reconciler's doc comment states: *"Grants and revokes workspace access without checking the caller's rights: the workspace list comes from the caller. Callers must have established superadmin, or admin on every workspace passed, beforehand. `authed` is audit attribution only, not an authorization check."* The EE side isn't visible to this repo's checkout, which is likely why it read as missing; no change needed (companion PR: windmill-labs/windmill-ee-private#726).

alpetric:
Follow-up confirmed and fixed in d073dc8d8b (+ EE de68943), with a stronger design than the suggested per-email lock:

The no-op observation was right, and it generalizes — `apply_instance_role` takes no lock when the role is unchanged, and some paths (`edit_instance_groups`, `delete_igroup` of a role-less group) never touch `password`/`token` for the affected users at all, so no "row locks first" discipline can be made complete. Instead the per-workspace **advisory lock is now first in the lock hierarchy**: every mutation path acquires it up front (`lock_workspaces_referencing_instance_groups`, sorted order, before any `workspace_settings`/`password`/`token`/`usr` row lock), and the reconciler re-acquires it idempotently. Concurrent mutations of the same workspace serialize at that single acquisition point before taking any row lock, so there is no ordering left to invert — including for no-op role updates, multi-op SCIM patches, and paths that skip role updates entirely. Rename paths additionally lock workspaces referencing the target name before rewriting their settings rows.

alpetric:
Pi/Codex convergent finding fixed in the latest commit (+ EE 29a1e98): `lock_workspaces_referencing_instance_groups` now precedes the `email_to_igroup` write in all three flagged paths — `add_user_igroup`, `remove_user_igroup`, and SCIM `create_group` — so every mutation path now acquires the advisory lock before its first row lock, with no exceptions. As Pi noted, hoisting is semantically free: the affected-workspace set is a function of `workspace_settings` only. The single consistent order is now advisory → `email_to_igroup` → `password`/`token` → reconcile, everywhere.

alpetric:
Codex's discovery-race finding confirmed and fixed in 153b43126e (+ EE f0f3419):

The race was real — `remove_user_igroup` could snapshot "workspaces referencing G" before a concurrent `edit_instance_groups` added G to a workspace; the edit's reconcile then derived membership from the not-yet-deleted `email_to_igroup` row, both committed, and the workspace kept a member no configured group still grants.

Rather than a third advisory-lock keyspace, the **`instance_group` row itself is now the group-level mutex** (`SELECT … FOR UPDATE`), acquired ahead of the workspace advisory locks in every mutation path: the single-user paths lock it via their existing not-found lookup, `delete_igroup` via its `instance_role` fetch, `overwrite_igroups` locks all rows in sorted order, the SCIM branches take a transaction-scoped `FOR UPDATE` by id, and `edit_instance_groups` locks its configured groups (sorted) before touching settings. The full hierarchy — group rows (sorted) → workspace advisory locks (sorted) → all other row locks — is documented on the reconciler. SCIM `create_group` deliberately takes no group lock: a group that doesn't exist yet cannot pass `edit_instance_groups`' existence check, so no snapshot can race it.

With the group mutex held, the replay serializes: whichever transaction wins, the loser recomputes from committed state and the workspace converges on the correct projection.

alpetric:
Both findings addressed in the latest commit (+ EE ad184fa):

- **P1 (overwrite locking unseen rows late)**: real — per-row `FOR UPDATE` can't cover rows committed after its scan, so the unqualified deletes could lock a freshly created group's row while already holding workspace locks. `overwrite_igroups` now takes `LOCK TABLE instance_group IN EXCLUSIVE MODE` as its group-mutex phase, first in the transaction: the import replaces the whole catalog, so the whole-table lock is the honest granularity. It conflicts with every other path's `instance_group` writes and `FOR UPDATE` (so nothing can slip in between scan and delete, and the `edit_instance_groups` G→W / W→G cycle cannot form) while leaving plain reads unblocked. Taken before the workspace locks, it slots at the top of the existing hierarchy without reordering anything else.
- **P2 (stale hierarchy docs)**: both helper docs (`lock_workspaces_referencing_instance_groups`, `lock_instance_group_workspaces`) now state the actual order — `instance_group` row locks → workspace advisory locks → all other row locks — and point at the reconciler's contract as the canonical statement.

PR #10687, opened 2026-08-13
refactor: combine the per-minute counters onto one shared helper

hugocasa:
## Summary

The backend had four independent implementations of the same thing: a `DashMap` keyed by a string holding `{ count, minute_bucket }`, where a read resets the count once the wall-clock minute has rolled over. Written at different times, they had drifted:

| Site | Key | Count | Evicts stale keys? |
|---|---|---|---|
| `login_rate_limit.rs` (per-IP + per-account) | `String` | `i32` | yes, every 256 attempts |
| `public_app_rate_limit.rs` | workspace id | `i32` | **no** |
| `users.rs` (`TOKEN_CREATE_RATE_LIMIT`) | username | `i32` | **no** |
| `resources.rs` (`RESOURCE_WRITE_RATES`) | `(workspace, path)` | `u32` | yes, every 256 writes |

Two consequences worth fixing rather than just deduplicating:

- The two sites without eviction grow one entry per distinct key for the process lifetime. Bounded by workspace count and username count respectively, so not dramatic, but unbounded in principle and clearly unintended given the other two do sweep.
- The non-obvious hazard in this pattern (the `entry()` guard holds a lock on its `DashMap` shard, and `retain()` takes every shard, so keeping the guard alive across the sweep deadlocks the request handler) was documented and handled in exactly one of the four. Anyone adding eviction to the others by copying the shape had to rediscover it.

Behavior at each call site is preserved, with two deliberate exceptions: the two previously-unevicted maps are now bounded, and the login per-account map is now swept every 256 recorded failures rather than every 256 login attempts (it used to share an eviction counter with the per-IP map). That map only ever gains entries on failures, so its bound is unaffected.

## Changes

- Add `windmill-common/src/per_minute_counter.rs`: `PerMinuteCounter<K>`, with a public surface sized to exactly what the callers need.
  - `count()` — current-minute count without recording one, `Borrow<Q>` so the login account peek stays allocation-free
  - `increment() -> u32` — record one, return the new count (for the resource advisory's `== N` crossing)
  - `try_increment(key, limit) -> bool` — record unless the key already reached `limit`; returning `bool` rather than a `Result` keeps each caller's own error message and status at the call site
  - Both mutating methods delegate to a private `bump_at(key, limit, minute)`, which owns the rollover reset, the increment, and the every-256-events sweep. It takes the minute explicitly so the tests can drive rollover and eviction without sleeping.
  - The shard-lock/`retain` deadlock constraint now lives here as the single comment guarding it.
- Rewrite the four call sites onto it, deleting their local entry structs, eviction counters, and duplicated bucket logic.
- Public app limit: the limit arrives from the workspace setting as `i32`, so clamp with `limit.max(0) as u32` — a wrapped negative would become an effectively unlimited allowance, where a non-positive limit must reject every execution.
- Login limits: per-IP uses `try_increment`, the per-account pre-check uses `count()`, `record_login_failure` uses `increment()`. The global limit stays on its atomic pair — it is unkeyed, so a map would buy nothing.
- Drop `dashmap` as a direct dependency of `windmill-api-users` and `windmill-store`; the counters were their only use. `windmill-api` keeps it for `s3_log_batching`.

Left alone deliberately: `cloud_usage.rs`'s `UsageBuffer` (monthly counts flushed to Postgres on a timer) and `s3_log_batching.rs` (batch state, not counts). Neither is this shape, and folding either in would mean generalizing the helper past what any caller wants.

No frontend change, so no screenshots.

## Test plan

- [x] `cargo check -p windmill-common -p windmill-store -p windmill-api-users -p windmill-api`, plus a full `cargo check`
- [x] `cargo test -p windmill-common --lib per_minute_counter` — three tests pinning what the call sites depend on: a refused `try_increment` does not record, the bucket resets on the next minute, and the sweep drops keys older than the previous minute while keeping the previous one
- [x] Resource write advisory, live: 25 writes to one resource within a minute produced exactly one `resource written 20 times in a minute` warning, at write 20 — the `==`-not-`>=` crossing behavior is intact
- [x] Public app rate limit, live: workspace limit set to 1, anonymous `execute_component` returned `200` then `429` then `429`, then `200` again after the minute rolled over followed by `429`
- [x] Login limits, live: backend restarted with `LOGIN_RATE_LIMIT_PER_IP=2` and `LOGIN_RATE_LIMIT_PER_ACCOUNT=2`. Per-IP gave two `400`s then `429`. Per-account gave two `400`s then `429` **from a different IP on each attempt**, confirming the account counter fired rather than the IP one. A valid admin login from a fresh IP still returned `200`, and the backend log showed no panics.

---
## Summary by cubic
Combines four per-minute `DashMap` counters behind a shared `PerMinuteCounter` to standardize rollover/eviction and remove a shard-lock/retain deadlock risk. Old behavior had inconsistent eviction and ad‑hoc locking; new behavior adds bounded eviction to all counters and consistent minute resets. Side effects: public app and token creation counters now evict stale keys; negative public app limits now reject all; login per-account eviction is driven by failures only.

- Introduces `PerMinuteCounter<K>` with `count`, `increment`, and `try_increment(limit)`; resets on minute boundaries and evicts stale keys every 256 calls (including refusals).
- Refactors login limits: per-IP uses `try_increment`, per-account pre-check uses `count` (failures call `increment`), global limit stays atomic.
- Public app limit clamps with `limit.max(0) as u32` to avoid wrap-around; non-positive limits reject all.
- Moves token creation rate limit to `PerMinuteCounter` with `u32` counts.
- Keeps resource write advisory semantics; now uses `PerMinuteCounter` and logs once at the threshold crossing.
- Dependencies: drops `dashmap` from `windmill-api-users`; keeps `dashmap` in `windmill-store` for the Azure DevOps token cache and in `windmill-api` for `s3_log_batching`.
- Config: `LOGIN_RATE_LIMIT_PER_IP` and `LOGIN_RATE_LIMIT_PER_ACCOUNT` parse as `u32`; negative values are ignored (fallback to defaults).

<sup>Written for commit c07a8d09763e7d61898a27658d539d47f0115dba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
✅ Review round clean @ c07a8d09763e7d61898a27658d539d47f0115dba

PR #10688, opened 2026-08-13
feat: track token cost in AI sessions and chats

hugocasa:
## Summary

AI chats already parse provider token usage on every path and accumulate it across a turn — `chatLoop.ts` even calls the total "suitable for cost accounting" — but the number was thrown away after one anonymous telemetry event. Nothing was persisted, nothing was shown, and nobody could answer "what did this chat cost" or "what is this workspace spending on AI".

This records that spend per workspace and surfaces it in settings: an admin **AI usage** panel for the whole workspace, and a self-scoped **Your AI usage** panel in user settings.

Two gaps had to close first:

- `ChatTokenUsage` folded cache-read and cache-write tokens into `prompt`. Fine for the context gauge, wrong for money — a cached prefix reads at a fraction of the input rate and writes at 1.25x, and a long chat is mostly cache.
- `stream_options.include_usage` was only set for openai / azure_openai / azure_foundry / googleai, so chats on OpenRouter, Groq, DeepSeek, Mistral and TogetherAI reported **zero** tokens. Nothing to price, and no context gauge either.

**Estimated vs real cost.** Cost is an estimate — token counts times a price table — and is marked as such with a leading `~`, the way Cloudflare's AI Gateway labels its own, with a per-model override for negotiated rates. OpenRouter is the exception: it returns what it actually charged (`usage.cost`), and that figure wins over the estimate and renders without the `~`. A line aggregating both sources takes the `~`, since it is only partly an invoice.

**The backend stores tokens, not money.** Rates are applied when usage is read, from the built-in table plus the workspace's effective overrides. Nothing in the chat computes a price. One price table instead of two, no client-computed money trusted into the database, and correcting a rate fixes the history rather than only future rows — visible in the test plan below, where an override changes an already-recorded total.

**Spend is keyed by workspace throughout.** The client buckets each report by the workspace the chat operates on, posts it to that workspace's route, and the row's primary key carries the workspace id — so a chat continued after a workspace switch bills each half to its own workspace, and each panel prices only its own rows with its own rates.

**Rates resolve with the rest of the AI config.** `model_pricing` lives in `AIConfig`, so a workspace on instance defaults inherits the instance's rates along with its providers, and a workspace that declares its own providers uses its own. That is the existing all-or-nothing rule (`has_providers()`), not a new one — the same way `max_tokens_per_model` already behaves — which does mean a workspace with its own provider keys does not pick up instance-level negotiated rates and restates them in its own card.

**Unpriced is a supported state.** A model the table doesn't cover resolves to "no rate" and points at the override, rather than being guessed at. Newer revisions of a tracked family (`gpt-5.6`, `gemini-3.1`) are listed explicitly as unpriced so they can't silently inherit an older revision's rate.

## Changes

**Backend**
- New `ai_token_usage` table: per (workspace, day, user, provider, model, session) token counts, plus a nullable provider-reported cost. Deliberately not `feature_usage`, which is anonymous telemetry that leaves the instance and is pruned at 60 days.
- `POST /w/{workspace}/ai/usage` records a batch (any member; attribution comes from the session, never the payload, and the route carries a 64 KiB body limit). `GET` returns `{buckets, truncated}` grouped by day / user / model — `scope=workspace` (default) is admin only, `scope=self` filters to the caller's own email so a member can see what they are costing without being shown their colleagues'. `truncated` so a caller summing the rows can tell the total is partial.
- No per-session grouping: a session is identified by a client-generated id whose name lives only in the browser that made it, so a bucket keyed on one is a label nobody can resolve. `session_id` is still stored, at the grain the client batches on, should sessions ever gain a server-side name.
- `AIConfig.model_pricing`: per-`provider:model` rate overrides in USD per 1M tokens, alongside the existing `max_tokens_per_model`. Rates are bounded on both write paths — the typed workspace handler and the untyped instance settings blob, which bypasses it — and again on read.

**Frontend**
- `modelPricing.ts`: list prices for the families we can track, resolved with the same matcher machinery as the context-window table (so `anthropic/claude-opus-4.8`, `claude-opus-4-8-20260101` and `…:thinking` all land on one entry). Each entry carries its own cached-read rate, which varies by family (0.1x on Anthropic and gpt-5, 0.25x on gpt-4.1/o-series, 0.5x on gpt-4o). An override that leaves a cache rate unset inherits that model's own published multiple, and a model with no built-in entry has none to inherit, so its cached tokens bill at the full input rate — an unstated discount reads as none rather than as some other vendor's. `inheritedCacheRates` is the single implementation of that rule, shared by the resolver and the editor, so the placeholder a rate field shows is the figure that will actually be charged.
- `ChatTokenUsage` gains `cacheRead` / `cacheWrite` / `cost`; all three provider parsers fill them, and `billedTokens` derives the uncached share under both provider conventions.
- `include_usage` extended to the OpenAI-compatible providers that document it (`customai` deliberately excluded — arbitrary endpoint); OpenRouter additionally sends `usage: {include: true}` for the real-cost path.
- The chat loop hands each completed provider response to an `onUsage` callback, so a tool loop counts every request and a turn that fails partway still records what it spent. `AIChatManager` reports token counts through a batched, `keepalive` reporter — it keeps no running cost and persists nothing on the chat record.
- UI: an **AI usage** panel with a range picker and day/user/model grouping, priced with the workspace's *effective* rates so an instance-inheriting workspace agrees with its own chats; the same panel self-scoped in user settings; and a **Model pricing** card below it, where every model takes input, output and both cache rates, the cache fields showing their inherited value as a placeholder.
- Failures report what actually happened: only a 403 on the workspace scope claims a permission problem, so an expired session or a database hiccup no longer tells an admin to go find access they already hold.

## Out of scope

AI agent flow steps. `AIAgentResult.usage` already carries a `TokenUsage`, so pricing them is additive — but it needs the price table readable from the Rust worker, which is a separate decision (duplicate it, or serve it over the API).

## Screenshots

Workspace settings → AI, grouped by model. `$0.099` is what OpenRouter returned; the `~` figures are estimated from the rate table, and a model the table does not cover reads `no rate` rather than `$0.00`. Because some spend is unrated, the headline is marked `(partial)`, and hovering it explains that unrated models are not counted:

ai usage workspace

When *every* model in the range is unrated there is no figure to show, so the headline reads `—` rather than `$0`, which would claim the period was free:

all unpriced

User settings, self-scoped — same panel, only your own rows, and no "By user" grouping. The title names the workspace, since the user settings drawer opens from anywhere and carries no other workspace context. The total differs from the workspace one above because it excludes the other member's usage:

your ai usage

Model pricing, below the usage it explains. Every model takes all four rates; the cache fields show what an empty one falls back to, so `claude-haiku-4-5` reads 0.1 and 1.25 against its input rate of 1, while an unpriced Gemini model has no published multiple to inherit and would fall back to its input rate:

model pricing settings

## Test plan

- [x] `cargo check` and `SQLX_OFFLINE=true cargo check` on `windmill-api`, `windmill-api-settings`, `windmill-ai`
- [x] `svelte-check`: 0 errors across 10111 files
- [x] `vitest` on `components/copilot`: 965 pass; the 6 failures (`ContextManager`, `app/core`) are pre-existing on `main`, confirmed on a clean tree
- [x] API: batch upsert accumulates across posts and pre-sums duplicates within one batch; invalid provider/model rows skipped without a 500; a 300 KB body → 413; all three `group_by` values, with `session` now rejected as 400 alongside any other unknown value; `days=1` covers today only; 1200 sessions → 1000 buckets with `truncated: true`
- [x] Scoping: `scope=bogus` → 400; a non-admin member gets 403 on the default scope and 200 on `scope=self`, seeing only its own row while the admin's workspace view shows both; `POST` as that member is attributed to its own email
- [x] Workspace keying: the same admin posting to `test` and to `admins` produces two disjoint listings, neither showing the other's models
- [x] Rate validation: workspace path rejects `input:-5` and `1e9` (400); instance settings path rejects the same, `model_pricing: []`, and a non-object `ai_config` (400) while `null` still clears (200)
- [x] Real chat on Anthropic `claude-haiku-4-5`: recorded 3 uncached input + 24k cache-write, priced at the cache-write rate — which the old collapsed field would have got wrong; a second turn read back exactly those 24,016 tokens as `cache_read` plus a 26-token incremental write
- [x] Real chat on OpenAI `gpt-5-mini`: second turn hit OpenAI's automatic cache and recorded `input 222 / cache_read 19,712` from a `prompt_tokens` of 19,934 — the include-cached convention, which under Anthropic's would have double-counted 19,712 tokens
- [x] Real chat on Google AI `gemini-2.5-flash`: 18k tokens recorded and rendered `no rate`, confirming an unpriced model reports usage without claiming it was free
- [x] Real chat on OpenRouter `anthropic/claude-haiku-4.5`: panel shows `$0.099` with no `~`, matching the `reported_cost_nano_usd` that reached the DB; the response also carried a real `cache_write_tokens` of 22,358, confirming the field name the unit test pins
- [x] Real multi-tool-call turn: recorded `requests = 2`, one per provider response (it was 1 before the `onUsage` change)
- [x] Setting an override (haiku input $1 → $10) moved a recorded chat from `$0.030` to `$0.300` and the workspace total from `$0.928` to `$1.20` — a rate correction applies retroactively
- [x] Usage panel absent from instance settings, present in workspace settings and user settings
- [x] Instance inheritance: with an instance `model_pricing` of $100/1M for `claude-opus-5`, an inheriting workspace's 7000-in/300-out row priced `~$0.730` rather than the $0.0425 the built-in table gives, confirming the panel uses effective rather than built-in rates; `get_copilot_info` returns that pricing for the inheriting workspace and `None` for one with its own providers
- [x] Row counts over a 45-day backfill: 7/30/90-day ranges return 14/37/53 buckets and render 7/30/46 rows. Buckets exceed rows because the API groups by (day, provider, model) and the panel folds to one line per day, so the 1000-bucket cap is reached before the visible row count suggests; `truncated` surfaces it
- [x] Layout: at a 700px viewport the panel controls and the rate fields wrap without horizontal overflow, and a model name keeps a readable width instead of truncating to an ellipsis
- [x] Unpriced-model override: 1M uncached + 1M cached input at $2/1M bills $4.00, not the $2.20 an inherited 0.1x ratio would have given; stating `cache_read` explicitly overrides that
- [x] A day mixing a provider-reported model with an estimated one renders `~$0.067` rather than an unmarked figure, so a partly-estimated total never reads as an invoice; a line mixing a priced model with an unpriced one keeps the `~` and omits the unpriced model's spend

All four providers above were exercised against live APIs, and the by-model panel reconciles to the recorded rows exactly. Not exercised locally: the remaining OpenAI-compatible providers that `include_usage` was extended to (Groq, DeepSeek, Mistral, TogetherAI), for want of keys.


hugocasa:
✅ Review round clean @ 5c9d88f0486bccd54b077f8c1e59c8e6cc02983b — nit-only verdicts at 2e3e1b80d9b39c0356180d65cb12664ac1dc9065 (codex nits, claude and pi "Good to merge"); nits addressed in 5c9d88f (schema summary entry for `ai_token_usage`, Gemini pricing note corrected to match its unpriced state, Sonnet 5 rationale compressed to the 4-line limit)

hugocasa:
Fixed in `6fece029df`:

- **[P1] cleared cache rate kept the old value** — confirmed by reproduction before fixing (set `cache_read: 0.5`, cleared the field, saved, and the DB held `0.5` while the field showed the inherited `0.1` placeholder). An emptied cache field now deletes the key so it inherits again. Added an `onblur` resync too, since the root cause was broader: any value the state refuses left the DOM showing what was typed while state kept the old number.
- **[P1] partial totals shown as exact** — the headline now renders `~$0.142 total (partial)` whenever a bucket is unpriced, with a `title` explaining that unrated models are not counted. Following the existing convention in `instanceSettings/ObjectStoreConfigSettings.svelte`, which marks incomplete storage figures the same way. The all-unpriced case renders `—` rather than `$0`, since `$0` asserts zero spend where `—` asserts nothing.
- **[P2, pi] orphaned `resolveModelPrice` doc** — moved back onto the function.

**[P2] raw `<button>` for Reset — respectfully dismissing.** The blanket "never use raw HTML elements" rule is contradicted by the codebase for this exact case: `instanceSettings/RetentionPeriodOverrides.svelte` renders its inline add/remove actions as raw `<button type="button">` with `text-blue-500 hover:underline`, reserving `Button` for primary actions. Reverting one model's rate is a minor inline action; a bordered `Button` gave it the same visual weight as *Save AI settings*, which is what prompted the change. Happy to revisit if there's a design-system link component I missed.

hugocasa:
Both nits fixed in `eee162b7c3`:

- **rejected value's error outlived the value** — the blur resync now clears `errors[key]`. Verified: typing `5000` shows the out-of-range message while the bad value is on screen, and tabbing away reverts the box to `1` with the message gone, rather than leaving a valid-looking row flagged red.
- **`clearCacheRate` accepted `'input' | 'output'`** — narrowed to a `CacheField` alias, so deleting a required field is no longer expressible rather than merely prevented by the call site.

hugocasa:
**[P1] newer GPT revision variants inherit GPT-5 pricing** — confirmed and fixed in `9dbb5e237d`.

Reproduced before fixing, via `resolveModelPrice('openai', …)`:

| id | was | now |
|---|---|---|
| `gpt-5.4-mini` | `gpt-5` at $1.25 | unpriced |
| `gpt-5.5-pro` | `gpt-5` at $1.25 | unpriced |
| `gpt-5.4-nano` | `gpt-5` at $1.25 | unpriced |
| `gpt-5.6` | unpriced | unpriced |
| `gpt-5-mini` | $0.25 | $0.25 |

The explicit `gpt-5.1`–`gpt-5.6` null entries were meant to cover exactly this, but they are themselves subject to `strictVariants`: `gpt-5.4-mini` cannot match the `gpt-5.4` entry because `-mini` makes it a sub-model, so it fell through to `gpt-5`. Papering it with more entries would have the same hole for every future `-mini`/`-pro`/`-nano`, so the guard now rejects a following short numeric segment under `strictVariants`:

```
strictVariants ? '(?!-\\d{1,3}(?:$|-))' : ''
```

Short deliberately: a date is digits too (`-20251101`, `-2026-01-01`) and has to stay a decoration. Guarded on `strictVariants` so the context-window table keeps inheriting, where an approximate window is safe and a wrong price is not.

Regression tests added for `gpt-5.4-mini` and `gpt-5.5-pro`, plus one pinning that `gpt-5-2026-01-01` still resolves at $1.25 so the new guard cannot quietly start eating dates. `modelPricing.test.ts` 13/13, `modelConfig.test.ts` 4/4, full copilot suite 967 passed with the same 6 pre-existing `ContextManager`/`app/core` failures that fail on a clean tree.

hugocasa:
All three P1s fixed. `93a81efefe` (frontend) and `22cbbd80b3` (providers).

**[P1] legacy `/thinking` overrides never match** — confirmed and fixed. `setCopilotInfo` strips the suffix before the chat sees a model, so usage is reported under `claude-haiku-4-5`, while the editor iterated the raw config and saved under `claude-haiku-4-5/thinking`. Planted that exact config: two rate rows before, one (`claude-haiku-4-5`) after. Stripping can collapse two configured slots onto one model, so the list dedupes.

**[P1] Bedrock chats record no usage** — fixed and **verified against live AWS**. `bedrock_sse_chunks_for_event` now emits a terminal usage chunk for the `Metadata` event. A real chat through `eu.anthropic.claude-haiku-4-5-20251001-v1:0` recorded:

```
aws_bedrock/eu.anthropic.claude-haiku-4-5-20251001-v1:0  in=3  cr=0  cw=22599  out=1  req=1   →  ~$0.028
```

Before the fix that row did not exist at all. The 22,599-token cache write also exercises the split: Bedrock counts cache apart from `inputTokens`, the chunk folds them into `prompt_tokens` and reports them under `prompt_tokens_details`, and the client recovers `input=3` by subtracting. Unit test added on the converter. Also checked that regional inference-profile ids (`eu.`/`us.` prefixes) still resolve against the price table.

**[P1] Gemini omits cached and thinking tokens** — fixed, thinking half **verified live**. Asked Gemini to think and answer with one number; the visible reply was `391` (one token) and the recorded row shows `out=54`. Those are thinking tokens, billed at the output rate and previously discarded, so every thinking reply was recorded at a fraction of its cost. `cachedContentTokenCount` now maps to `prompt_tokens_details.cached_tokens`; that half is covered by a unit test but **not** proven live, since Gemini's implicit caching did not trigger during the test turns (`cr=0`). Regression test asserts both classes.

**[P2] matcher documentation** — updated; it claimed `strictVariants` cannot catch a version bump, which the previous commit made it do.

Both nits (duplicated usage-reporting rationale in `chatLoop.ts`) addressed.

hugocasa:
Round non-blocking: Codex and pi both nits-only, claude's run errored before posting.

**Fixed in `39d4cbf4e0`** — duplicated Gemini token-accounting rationale (raised by both). The disjointness invariant and its live figures now sit once on `gemini_prompt_tokens`, where someone would break it; the test's assertions already pin the numbers.

**Dismissing pi's P2 — `gemini-2.5-flash` is not tiered.** The finding states Flash is "tiered exactly like Pro: input $0.30→$0.60 and output $2.50→$10.00 above 200K". Google's pricing page does not show that. Checked twice, the second time aimed squarely at the tiering question:

> **Gemini 2.5 Flash** — Input "$0.30 (text / image / video) $1.00 (audio)", Output "$2.50", Context caching "$0.03 … $1.00 / 1,000,000 tokens per hour"
> **Token Threshold** — "Neither model displays pricing variations based on prompt length thresholds like the 200k-token distinction found in other models (Gemini 3.1 Pro, Gemini 2.5 Pro)."

The 200k threshold is real, and it is exactly why Pro is left unpriced and Flash is not — that is the distinction the comment draws, and it holds. If pi has a source showing a Flash threshold I would rather be corrected than ship a wrong rate, since this is precisely the "confidently wrong number" case the design rule is about.

One genuine nuance the finding didn't raise: Flash's **audio** input is $1.00 against $0.30 for text, and the per-model shape carries a single input rate, so an audio-heavy request would understate. The chat sends text, so no path reaches it today.

**Re-dismissing the raw `<button>` for Reset** (carried from two rounds ago, accepted by claude after it checked the precedent itself): `instanceSettings/RetentionPeriodOverrides.svelte:99-101` renders its inline actions as exactly `<button type="button" class="text-2xs text-blue-500 hover:underline">`, reserving `Button` for primary actions. Same directory, same purpose, same shape.

Verification on this head: `check:fast` clean, `modelPricing.test.ts` 14/14, `windmill-ai` gemini/bedrock tests pass, and the panel prices real recorded Gemini usage (`googleai/gemini-2.5-flash 54k in / 58 out → ~$0.016`).

PR #10690, opened 2026-08-13
fix: refresh AI provider model defaults and capability metadata

hugocasa:
## Summary

Audit of every AI provider's model catalog against the providers' current
documentation. Four providers were recommending models that no longer exist, the
newest families were missing entirely, and several capability lookups resolved
current models to the wrong window, output cap or reasoning ladder.

The recommendation lists (`defaultModels`) are pruned to what providers actually
serve today. The capability tables (context window, vision, reasoning) keep their
rows for retired ids — those can still sit in a saved workspace config, where a
stale row is harmless but a missing one degrades the request.

`defaultModels[0]` is the model a new workspace is created with, so each list
leads with the balanced tier rather than the frontier model, matching the
existing convention.

## Changes

**Retired models removed from recommendations**
- Anthropic: `claude-3-5-haiku-latest` (retired 2026-02-19)
- DeepSeek: `deepseek-chat`, `deepseek-reasoner` (retired 2026-07-24)
- Groq: `llama-3.3-70b-versatile`, `llama-3.1-8b-instant` (shutdown 2026-08-16)

**Current families added**
- OpenAI / Azure OpenAI: the `gpt-5.6` family (sol / terra / luna)
- Anthropic: `claude-sonnet-5`, `claude-opus-5`
- Google AI: `gemini-3.6-flash`, `gemini-3.5-flash`, `gemini-3.5-flash-lite`
- DeepSeek `deepseek-v4-flash`, Mistral `mistral-medium-latest`, Groq gpt-oss/qwen,
  Bedrock Claude 5, and a refreshed Azure AI Foundry list

**Capability metadata corrections**
- `claude-opus-5` was matched by no reasoning pattern, so the effort control was
  silently unavailable for it; `claude-mythos-5` was unhandled
- `claude-opus-5` / `claude-sonnet-5` were missing the `xhigh` level, and
  `gpt-5.6` was missing `xhigh` and `max`
- `gpt-5.6` resolved to the 400K `gpt-5` window instead of its own 1.05M
- Gemini 3.x fell through to the 8192 output catch-all instead of 64K; Opus was
  capped at 32K where its siblings get 64K (Opus 4.1 and older correctly stay at 32K)
- `mistral-medium-latest` / `-3.5` resolved to the 128K fallback instead of 256K
- Claude 5 thinks when the `thinking` field is omitted, so turning reasoning off
  was a no-op. Off now sends an explicit `thinking: {type: "disabled"}`, following
  the existing DeepSeek sentinel pattern. Fable/Mythos reject the disable, and
  Bedrock has no route for it, so neither offers an off switch.

**Backend**
- Bedrock prompt caching was allowlisted only up to Claude 4.5, so caching was
  silently off for every 4.6+/5 model. Those families are matched by prefix
  because Bedrock publishes bare, `-v1` and `-v1:0` spellings of the same id.

## Screenshots

Workspace AI settings with the refreshed Anthropic and Google AI defaults —
`claude-sonnet-5` and `gemini-3.6-flash` lead their lists, and `claude-opus-5`
resolves to a 64000 output budget (previously 32000):

ai-settings-defaults

wm-ai-settings-anthropic

## Test plan
- [x] `npx vitest run src/lib/components/copilot/` — 957 passing; the 6 failures are
      pre-existing on `main` (`app/core.test.ts` mock, `ContextManager.test.ts`)
- [x] `cargo test -p windmill-ai --features bedrock` — 127 passing, including new
      coverage that every Claude 4.6+ id spelling reaches the caching prefix match
      and that Opus 4.5's dated id is not swallowed by it
- [x] Workspace AI settings exercised in the browser: model lists render the new
      defaults and the per-model output limit reflects the corrected cap
- [ ] With provider keys: confirm an "off" selection on `claude-opus-5` sends
      `thinking: {type: "disabled"}` and that a Bedrock Claude 5 chat completes
      with cache points enabled

---
## Summary by cubic
Refreshes provider model catalogs, defaults, and capability metadata. Fixes two behavior gaps: Bedrock prompt caching now covers Claude 4.6+/5 (incl. AU geo), and reasoning “off” on Claude 5 translates to an explicit disable; older Claude keeps omission. A shared off sentinel keeps this translation consistent across routes.

- **Catalog and capabilities**
  - Prunes retired recommendations; adds current families (OpenAI `gpt-5.6` sol/terra/luna, Anthropic `claude-sonnet-5`/`claude-opus-5`, Google `gemini-3.6`/`3.5`, DeepSeek `deepseek-v4-flash`, Mistral `mistral-medium-latest`, Groq `openai/gpt-oss`) and refreshed Azure AI Foundry and Bedrock Claude 5 ids. New workspaces default to `claude-sonnet-5`, `gemini-3.6-flash`, `gpt-5.6-terra`.
  - Corrects windows and caps: `gpt-5.6` = 1,050,000; Mistral Medium 3.5/`-latest` = 256,000; Gemini 3.x output cap = 64,000; Sonnet/Haiku/Fable/Mythos and Opus 4.5+ = 64,000 (Opus 4.1 and older = 32,000).
  - Updates reasoning ladders: Claude 5 and `gpt-5.6` include `xhigh`/`max`; Mistral Medium 3.5 accepts `high` only; `deepseek-chat` remains excluded.

- **Backend and routing**
  - `aws_bedrock`: enables prompt caching by family prefixes for Claude 4.6+/5 (incl. Fable 5), normalizes `au.` geo, and avoids swallowing Opus 4.5 dated ids. Translates the shared off sentinel (`none`) to `thinking: {type: "disabled"}` for Claude 5; 4.6–4.8 keep omission.
  - Anthropic/`azure_foundry`/`openrouter`: map the same off sentinel to a disable for Claude 5; 4.6–4.8 keep omission; Fable/Mythos reject disables. Drop temperature when adaptive thinking is on; preserve it when disabled. Shared off sentinel is defined once and covered by tests across routes.
  - No migration required.

<sup>Written for commit 64aa0facbfef2a6c3d6e4936741bdc1d2c73f52b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10690?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-ai/src/ai_bedrock.rs
+++ b/backend/windmill-ai/src/ai_bedrock.rs
@@ -113,6 +113,7 @@ const BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_IDS: &[&str] = &[
 /// model (`anthropic.claude-sonnet-4-6`, `...-4-6-v1`, `...-4-6-v1:0`), so they
 /// are matched by family prefix rather than by exact id.
 const BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_PREFIXES: &[&str] = &[
+    "anthropic.claude-fable-5",
     "anthropic.claude-opus-4-6",
     "anthropic.claude-opus-4-7",
     "anthropic.claude-opus-4-8",
@@ -135,7 +136,7 @@ fn normalize_bedrock_model_id(model: &str) -> String {
         .unwrap_or(model)
         .to_ascii_lowercase();
 
-    for prefix in ["global.", "us.", "eu.", "apac."] {
+    for prefix in ["global.", "us.", "eu.", "apac.", "au."] {
         if let Some(normalized_model) = model.strip_prefix(prefix) {
             return normalized_model.to_string();
         }
@@ -1267,6 +1268,8 @@ mod tests {
             "global.anthropic.claude-opus-4-8",
             "anthropic.claude-opus-5",
             "eu.anthropic.claude-sonnet-5-v1:0",
+            "au.anthropic.claude-sonnet-5",
+            "anthropic.claude-fable-5",
         ] {
             assert!(
                 bedrock_model_supports_prompt_caching(model),
--- a/backend/windmill-ai/src/ai_bedrock.rs
+++ b/backend/windmill-ai/src/ai_bedrock.rs
@@ -109,6 +109,18 @@ const BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_IDS: &[&str] = &[
     "anthropic.claude-3-5-sonnet-20241022-v2:0",
 ];
 
+/// Claude 4.6 and later are published under several id spellings for the same
+/// model (`anthropic.claude-sonnet-4-6`, `...-4-6-v1`, `...-4-6-v1:0`), so they
+/// are matched by family prefix rather than by exact id.
+const BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_PREFIXES: &[&str] = &[
+    "anthropic.claude-opus-4-6",
+    "anthropic.claude-opus-4-7",
+    "anthropic.claude-opus-4-8",
+    "anthropic.claude-opus-5",
+    "anthropic.claude-sonnet-4-6",
+    "anthropic.claude-sonnet-5",
+];
+
 fn build_default_cache_point() -> aws_sdk_bedrockruntime::types::CachePointBlock {
     aws_sdk_bedrockruntime::types::CachePointBlock::builder()
         .r#type(aws_sdk_bedrockruntime::types::CachePointType::Default)
@@ -135,6 +147,9 @@ fn normalize_bedrock_model_id(model: &str) -> String {
 pub fn bedrock_model_supports_prompt_caching(model: &str) -> bool {
     let normalized_model = normalize_bedrock_model_id(model);
     BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_IDS.contains(&normalized_model.as_str())
+        || BEDROCK_PROMPT_CACHING_SUPPORTED_MODEL_PREFIXES
+            .iter()
+            .any(|prefix| normalized_model.starts_with(prefix))
 }
 
 fn append_cache_point_to_system_prompts(system_prompts: &mut Vec<SystemContentBlock>) {
@@ -1241,6 +1256,25 @@ mod tests {
         ));
     }
 
+    /// Claude 4.6+ ships under bare, `-v1` and `-v1:0` spellings of the same id,
+    /// so every one of them has to reach the prefix match.
+    #[test]
+    fn bedrock_prompt_caching_supports_claude_4_6_and_later_id_spellings() {
+        for model in [
+            "anthropic.claude-sonnet-4-6",
+            "anthropic.claude-sonnet-4-6-v1:0",
+            "us.anthropic.claude-opus-4-6-v1",
+            "global.anthropic.claude-opus-4-8",
+            "anthropic.claude-opus-5",
+            "eu.anthropic.claude-sonnet-5-v1:0",
+        ] {
+            assert!(
+                bedrock_model_supports_prompt_caching(model),
+                "{model} must support prompt caching"
+            );
+        }
+    }
+
     #[test]
     fn bedrock_prompt_caching_rejects_unsupported_or_opaque_model_ids() {
         assert!(!bedrock_model_supports_prompt_caching(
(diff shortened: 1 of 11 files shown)


hugocasa:
Round 1 addressed in bcb12de194. Notes on what I changed vs. dismissed:

**Fixed (codex P1 — Bedrock caching coverage):** added `au.` to the geo prefixes
`normalize_bedrock_model_id` strips, and `anthropic.claude-fable-5` to the caching
prefix list, with both spellings pinned in the test. Good catch — an AU inference
profile failed both the exact-id and prefix match.

**Dismissed (codex P1 — Azure Foundry Mythos `max`):** this is not introduced by
this PR. On `main`, `anthropicReasoningLevels` already ends in
`return ['low', 'medium', 'high', 'max']`, and `reasoningProviderFamily` already
folds every Foundry Claude deployment into that family — so `max` is offered for
Foundry Claude today, Mythos included, with or without this diff. What this PR
adds for Mythos is `xhigh`, not `max`.

Separately, `claude-mythos-5` is Project Glasswing (invitation-only, Anthropic
direct); I found no evidence it is deployable on Foundry at all. If Foundry does
reject `max` for some Claude deployments that is a real but pre-existing bug in
the shared ladder, and scoping it wants the actual Foundry capability table rather
than a Mythos-shaped special case bolted on here.

**Claude nits, all four fixed:**
- Reverted the Mistral ladder to `['high']`. The widening rested on a secondary
  source, and this PR promotes `mistral-medium-latest` to `defaultModels[0]` — not
  the place to relax a constraint the file records as live-verified. Kept only the
  dot-tolerant id match, so `mistral-medium-3.5` and `-3-5` both resolve.
- Restored the `deepseek-chat` exclusion, consistent with the retired-id rationale
  used in `MODEL_CONTEXT_WINDOWS`.
- Dropped `qwen/qwen3.6-27b` from the Groq defaults rather than shipping a default
  with no vision or context metadata. `openai/gpt-oss-120b` is Groq's other
  documented migration target for `llama-3.3-70b-versatile` and already has both.
- Rewrote the disable-branch comment to state the constraint (the disable must not
  carry an effort) instead of describing the outcome.

Also added the coverage gap Claude flagged: the off-path divergence is now pinned
in both directions — native Anthropic Opus 4.8 returns `'none'` while the same
model on Bedrock stays omission.

The credentialed checks (Anthropic `off` on the wire, Bedrock cache reads) remain
unchecked in the test plan; I have no provider keys here.

hugocasa:
**Dismissing the round-2 P1 (Bedrock Claude Opus 5 transport).** The premise is
contradicted by AWS's own model card for the model:
[Claude Opus 5 on Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html).

Its **APIs supported** table marks **Converse ✅**, alongside Invoke ✅ and
Messages ✅ — Converse is not excluded. The page's own *Converse API* sample uses
the exact id this PR recommends:

```python
client = boto3.client('bedrock-runtime', region_name='us-east-1')
response = client.converse(
    modelId='global.anthropic.claude-opus-5',
    messages=[{'role': 'user', 'content': [{'text': '...'}]}],
)
```

Windmill does send Bedrock chats through `converse` / `converse_stream`
(`providers/bedrock.rs:375,681,984`), so the transport matches. No change made.

The same page independently confirms three other things in this PR:
- **Prompt caching: Yes**, min 512 tokens per checkpoint, 5m + 1h TTL — the
  backend prefix addition is correct.
- **Geo inference ids** include `au.anthropic.claude-opus-5`, which is the `au.`
  normalization gap fixed in bcb12de194.
- **"Adaptive thinking is on by default; can be disabled but effort is capped at
  high when thinking is disabled"** — exactly the constraint the frontend disable
  branch encodes (send the disable with no effort).

On that last point, Bedrock's `canDisable` stays `false` for the 5 family here for
a codebase reason rather than a provider one: `bedrock_thinking_fields`
(`providers/bedrock.rs:417`) unconditionally emits
`{"thinking": {"type": "adaptive"}, "output_config": {"effort": …}}` for any
effort, so there is no disable route on that path today. Bedrock supports it;
Windmill's Converse mapper doesn't yet. That is a worthwhile backend follow-up,
and it is the same defect class the Claude reviewer flagged for the AI-agent step
(`providers/anthropic.rs:657` maps any effort to adaptive thinking) — both are
outside this diff.

Leaving this in draft rather than flipping to ready: the Codex verdict is
blocking, and I am refuting it rather than fixing it, so the call on whether that
refutation stands is @hugocasa's.

hugocasa:
✅ Review round clean @ 64aa0facbfef2a6c3d6e4936741bdc1d2c73f52b — nit-only verdicts at d3e611e79b36b0af2d25e42b5ac76147b5cf70cf (codex: mergeable-with-nits; claude, pi: good to merge); nits addressed in 64aa0facbf

PR #10694, opened 2026-08-14
fix: keep non traffic-serving processes out of coordinated restarts

rubenfiszel:
## Summary

`announce_server_started` writes a `background_task_state` row keyed on `INSTANCE_NAME`, which is
`rd_string(5)`, a fresh random value per process. The `ON CONFLICT (name)` therefore never fires,
and every process start inserts a row nothing ever reuses.

`run_server` is reached by every non-agent process, worker-mode ones included, so this ran in
workers too. With `EXIT_AFTER_N_JOBS=1`, one process start per job, the table grows by a row per
job: roughly 700k rows steady state at 100k jobs/day, all of them scanned by
`check_any_server_started`'s `name LIKE 'server_heartbeat:%'`. The 7-day sweep is the only bound,
and it only runs in server mode.

The row exists so a coordinated restart can hold a shutdown open until a peer is confirmed
serving. That matters for an API server: if every replica is down at instant T, a request arriving
at T fails outright, with no queue and no retry. It does not matter for a worker, an indexer or an
MCP process, none of which answers traffic that outlives the process. A worker group sitting empty
for a moment costs queue latency rather than work: `v2_job_queue` rows are durable, and the
killpill is read between jobs (`windmill-worker/src/worker.rs:3222`) so the job in flight still
drains and persists its result.

So the fix is to keep those processes out of the mechanism entirely rather than to make their rows
cheaper, and to bound the remaining growth from server restarts on top.

Worth noting the old peer gate never staggered anything: non-first processes ignore their computed
`delay` and all release on the same first-peer poll, so `restart_worker_group` already restarted
the whole group together after the first worker returned. What changes for a worker group is a
window of roughly one second with no worker up, instead of one worker staying up while the rest
shut down.

## Changes

- `announce_server_started` is called only when `server_mode` (`Mode::Server` or
  `Mode::Standalone` with the server enabled). Worker, indexer and MCP processes no longer write
  a heartbeat row at all.
- `spawn_graceful_killpill` takes `server_mode`. Non-server processes skip
  `coordinate_restart_delay` and the peer wait and take the 3s drain delay. They no longer write
  `_restart_coordination` entries either, so a worker can no longer claim the `is_first` slot and
  leave every server waiting on a peer that answers nothing.
- The rows that servers do still write are keyed `server_heartbeat:{HOSTNAME}` instead of
  `server_heartbeat:{INSTANCE_NAME}`, so a restart in place reuses one row. `owner` stays the
  per-process `INSTANCE_NAME`, which is what lets `check_any_server_started` tell a peer's start
  from its own when several processes share a host and therefore a row.
- The `ON CONFLICT` branch also refreshes `started_at`. It never fired before, so there is no
  prior behaviour to preserve.

Pre-existing instance-keyed rows are left to the 7-day sweep. They cannot affect a restart
decision meanwhile, since `check_any_server_started` filters on `updated_at > not_before`.

## Test plan

Verified on a local instance, CE debug build. `_restart_coordination` was cleared between runs,
since its entries stay non-stale for 120s and otherwise make every process believe it is not
first.

- [x] **Workers write no row.** 5 consecutive `MODE=worker DISABLE_SERVER=true` starts: the
      `server_heartbeat:%` count stayed at 1 throughout, that one row belonging to the separate
      standalone dev backend. On main the same run went 1 to 6.
- [x] **`restart_worker_group` still works, without a peer.** Two supervised workers in one group
      on distinct hosts. Both logged `Scheduling worker group restart requested graceful shutdown
      in 3s (first_to_restart=true)`, both were listening for jobs again 4s after the
      notification, no heartbeat rows were written, and `_restart_coordination` was left empty.
- [x] **Servers still coordinate (regression check).** Two `MODE=server` processes plus an
      `expose_metrics` change. `srvhost1` took `first_to_restart=true` and was back up at `:49`;
      `srvhost2` logged `healthy peer detected after 5.0s` and restarted at `:53`, so the
      availability overlap held and it did not fall through to the 120s ceiling.
- [x] **Server rows are reused.** 3 restarts of the dev backend: `owner` went
      `YaulL`, `9d6Pj`, `6ABHp`, `ajo6y` with the total `server_heartbeat:%` count staying at 1.
- [x] **Shared host.** Two servers under one hostname, sharing a single row. The non-first still
      logged `healthy peer detected after 5.0s`, since the first one's restart rewrote `owner` to
      a new instance name. This is the case the per-process `owner` exists to protect.

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1691,7 +1691,8 @@ async fn process_notify_event(
         "restart_worker_group" => {
             if worker_mode && payload == *WORKER_GROUP {
                 tracing::info!("Restart requested for worker group '{payload}'");
-                spawn_graceful_killpill(tx, db, 30, "worker group restart requested").await;
+                spawn_graceful_killpill(tx, db, 30, "worker group restart requested", server_mode)
+                    .await;
             }
         }
         "notify_webhook_change" => {
@@ -1997,8 +1998,14 @@ async fn process_notify_event(
                     reload_otel_tracing_proxy_setting(conn).await;
                     if worker_mode {
                         tracing::info!("OTEL tracing proxy setting changed, restarting worker");
-                        spawn_graceful_killpill(tx, db, 30, "OTEL tracing proxy setting change")
-                            .await;
+                        spawn_graceful_killpill(
+                            tx,
+                            db,
+                            30,
+                            "OTEL tracing proxy setting change",
+                            server_mode,
+                        )
+                        .await;
                     }
                 }
                 REQUIRE_PREEXISTING_USER_FOR_OAUTH_SETTING => {
@@ -2009,12 +2016,20 @@ async fn process_notify_event(
                 }
                 EXPOSE_METRICS_SETTING => {
                     tracing::info!("Metrics setting changed, restarting");
-                    spawn_graceful_killpill(tx, db, 30, "metrics setting change").await;
+                    spawn_graceful_killpill(tx, db, 30, "metrics setting change", server_mode)
+                        .await;
                 }
                 EMAIL_DOMAIN_SETTING => {
                     tracing::info!("Email domain setting changed");
                     if server_mode {
-                        spawn_graceful_killpill(tx, db, 30, "email domain setting change").await;
+                        spawn_graceful_killpill(
+                            tx,
+                            db,
+                            30,
+                            "email domain setting change",
+                            server_mode,
+                        )
+                        .await;
                     }
                 }
                 EXPOSE_DEBUG_METRICS_SETTING => {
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ b1c482517898623856b074bdc73e009510352435 — nit-only verdicts at f8b35ee3628adf9fbd7d28d8bf814719859acfbd (codex + pi "Good to merge", claude nit-only); nit addressed in b1c4825178

rubenfiszel:
Round 2 on `36b56fcc23`: codex and claude both **Mergeable with nits**, pi **Good to merge**. Both nits were on the same doc comment and are fixed in `fd16d9ae26`.

- Codex was right that "the job in flight still finishes" overpromises: `handle_docker_job` stops the container on killpill receipt (`bash_executor.rs:537`). The claim is now limited to `v2_job_queue` durability.
- Claude was right that "none of them answers traffic that outlives the process" is false for indexer (it is the only process serving `/api/srch/...` in a split deployment) and MCP (it serves the gateway routers). The comment now records the worker case as the rationale, without the universal claim.

On the earlier finding that removing worker coordination discards restart staggering: there was no staggering to discard. Non-first processes never sleep their computed `delay`, they poll `check_any_server_started` and all release on the same first-peer hit (`main.rs:2352-2377`), so `restart_worker_group` already restarted the group together once the first worker returned. What actually changes is that group capacity now dips to zero for about a second instead of one worker staying up. That is an accepted trade here: the queue rows are durable, so the cost is queue latency rather than lost work.

**One path went unexercised.** `MODE=indexer` is the other non-server mode that reaches `spawn_graceful_killpill`, and I did not run one: indexer mode panics without the `tantivy` feature, which this dev baseline does not compile, and building it would have meant a full rebuild through the shared backend pane. The claim decomposes into two things that are established separately: `server_mode` is false for `Mode::Indexer` by direct reading (`main.rs:787-791`), and the `!server_mode` branch was exercised with workers (3s drain, no coordination entries, no heartbeat row). Both halves hold, but no real indexer process was restarted.

PR #10695, opened 2026-08-14
feat: stream audit logs in batches when a page is slow to load

rubenfiszel:
## Summary

The audit logs page had no equivalent of the runs page's "load 1 by 1" mode: a slow page of logs
was one request you could only wait out (the old 4s toast offered to drop to 25 rows per page,
which throws away the other 75). It can now stream the requested page in batches, so rows land
progressively, the batch size is adjustable mid-flight, and the load can be stopped with the rows
already fetched left on screen.

The entry points mirror the runs page exactly: a load that takes more than 4s offers "Stream by
batches of 25", and a batched load whose batches are themselves slow offers "Stream 1 by 1".

Fixes WIN-2376

## Changes

**Backend** (EE companion: windmill-labs/windmill-ee-private#727)
- `before_id` query param on `GET /w/{workspace}/audit/list`: `id < before_id` on the existing
  `id DESC` ordering. Batches after the first resume from the previous batch's last id instead of
  a growing `OFFSET`, which the `(id, timestamp)` primary key on `audit_partitioned` serves
  directly. Optional, so existing callers are unaffected.

**Frontend**
- New `useAuditLogsLoader` composable owns audit log loading (it used to live inside
  `AuditLogsFilters`, which also meant the mobile filter popover ran a second competing load of
  the same query). It streams a page as `ceil(perPage / batchSize)` requests, capped at 1000 rows
  per page, and exposes progress, batch size, reload and stop.
- The runs page's batch-progress row is extracted to `BatchLoadProgress.svelte` and reused; on the
  audit page it renders inside the table footer, so appearing mid-load shifts nothing.
- The timeline is fed the settled result rather than every intermediate batch — regrouping it can
  fire extra requests to fill in missing job spans.

## Test plan

- [x] `npm run check` clean; `computeFirstBatch` unit test covers the page-offset arithmetic
- [x] Page 2 streamed in batches of 30 (a size that divides neither the page nor the offset)
      returns exactly the same 100 contiguous ids as one unbatched page-2 request
- [x] Requests observed: `page=4&per_page=30` (aligned offset, 10 rows dropped) then
      `per_page=30&before_id=…` cursors
- [x] Stop mid-stream halts the request chain and keeps the rows already loaded; stopping before
      the first batch of a *new* query clears the previous query's rows instead of showing them
- [x] `?perPage=100000` is clamped to one 1000-row request rather than thousands of batches
- [x] `perPage=25` still gets a slow-load warning (offering 1-by-1)
- [x] Batch size input is bounded to `perPage - 1`, so it cannot silently end streaming
- [x] Filters, Refresh, Previous/Next and per-page still issue exactly one request chain each
- [x] Runs page renders the extracted progress row unchanged

## Screenshots

Slow load offers to stream (audit logs page):

toast

Streaming by batches of 25, progress in the table footer:

progress

Streaming 1 by 1:

one by one

Runs page, same row after the extraction:

runs

---
## Summary by cubic
Streams audit logs in batches when a page is slow, matching the Runs page’s progressive mode. Previously one slow page was a single long request (or dropping to 25 rows); now rows render progressively, batch size is adjustable, Stop preserves loaded rows, and failures clear stale results and reset paging. Implements WIN-2376.

- Backend/API: optional before_id on GET /w/{workspace}/audit/list enables keyset pagination on id DESC; OpenAPI updated; EE reference updated; existing callers are unaffected.
- Frontend: new `useAuditLogsLoader` streams up to 1000 rows/page, exposes progress, restream-with-batch-size, and stop; loading moved out of `AuditLogsFilters` to avoid duplicate loads; `BatchLoadProgress.svelte` extracted and reused by Runs; the timeline consumes the settled result only.
- UX and resilience: after 4s offer “Stream by batches” (and “Stream 1 by 1” if batches are slow); while streaming, periodic toasts offer “Stop”; perPage is clamped to 1000 and batch size ≤ perPage − 1; Stop is ignored once a load settles; if a new query is stopped or fails before first rows arrive, stale rows are cleared and paging is reset.
- Pagination correctness: first batch aligns via computeFirstBatch; `hasMore` is set only once the page completes; unit test added.

<sup>Written for commit 3ad62376670233932682bbcd201de2ffe72ec25f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10695?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

Left in draft: the change carries an EE companion PR and adds a query parameter to `openapi.yaml`, so it wants a human look before ready. Review round is clean (nit-only verdicts, nits addressed).


rubenfiszel:
✅ Review round clean @ a5525639156eca14a1642ce1c81b932b930ca5f6 — nit-only verdicts at d202722e7a6cb6f4faffc68804076b71f42f927e (codex/claude/pi all "Mergeable, but should ideally address nits"); nits addressed in a5525639156eca14a1642ce1c81b932b930ca5f6 (stop is ignored once a load has settled, and a failed load resets paging instead of keeping the previous query's).

PR #10696, opened 2026-08-14
feat: add trigger_history table with source tracking

rubenfiszel:
## Summary

Fixes WIN-2378.

Every schedule and trigger mutation is now recorded in a new `trigger_history` table:
**who** did it, **what** changed, and **from what kind of client**. All three are derived by
the server at write time — no caller supplies any of them.

`source` is one of:

| value | how it is decided |
|---|---|
| `cli` | the request declares `X-Windmill-Client: cli`, which `wmill` now sets globally |
| `ui` | no declared client, but the request carries the session token minted at browser login |
| `api` | a token that is not a session token: user scripts, CI, third-party clients |
| `worker` | no request at all — the server disabling a schedule or trigger on its own after a failure |

`changes` is a compact `{field: {old, new}}` map, computed by diffing a `to_jsonb` snapshot of
the row before and after the write, minus bookkeeping columns (`edited_at`, `edited_by`,
`extra_perms`, `server_id`, `last_server_ping`, `email`). A create keeps only the fields that
were actually set; a delete carries no diff (the trashbin already holds the full row). A
payload over 32 KiB is replaced by the list of field names it would have held, so a large
`args` cannot write a multi-megabyte row.

## Changes

**Backend**
- `trigger_history` migration: append-only table, indexed by `(workspace_id, trigger_kind, path, id DESC)`.
  RLS mirrors the *path* half of a live trigger's own policies (`u/` owner, `g/` member,
  `f/` folder reader), so a row quoting a schedule's `args` never reaches someone who could not
  read the schedule. Narrower than the live row on one point: `extra_perms` grants have no
  counterpart, since the history must outlive the row's ACL.
- `windmill-common/src/trigger_history.rs`: the `TriggerSource` / `TriggerOperation` enums, the
  request-client task-local, the snapshot/diff helpers and the insert.
- Recorded from: schedule create / update / setenabled / delete and the workspace-wide default
  error-handler override (`windmill-api-schedule`); the generic `TriggerCrud` create / update /
  delete / setmode, which covers every trigger kind (`windmill-trigger`); and the two
  server-initiated disables — a schedule the queue disables after `QuotaExceeded`/`NotFound`
  (`windmill-queue`, `windmill-worker`) and a trigger the listener disables on error
  (`windmill-trigger`). The worker paths write on their own connection so a history failure can
  never poison the job-completion transaction.
- `GET /w/{workspace}/triggers_history/list` (`trigger_kind`, `path`, paginated), read through
  the RLS pool and gated by a new read-only `triggers_history` token-scope domain. Its own
  domain rather than an alias of `schedules` / `*_triggers`: one listing spans every kind, and a
  history row quotes the whole trigger row, so reading it is an explicit grant rather than a side
  effect of being able to read the trigger.

**CLI**
- `wmill` sets `X-Windmill-Client: cli` on every request, alongside the existing
  `X-Windmill-Deploy-Origin`.

**Frontend**
- A history button in the shared trigger editor toolbar opens a drawer listing the modifications
  of that trigger, with operation and source badges. Opt-in per editor via a `triggerKind` prop,
  so kinds `trigger_history` does not record never show it.

## What is recorded, and what is not

Twenty-odd non-test files write these tables, so "record every write" is not a boundary anyone
can hold. The line this PR draws, now stated in the module docs so the next route knows which
side it is on:

**Recorded** — authoring a single trigger through its own surface: create (including
`http_triggers/create_many`), update, delete, enable/disable/suspend, restore from the trashbin,
the workspace-wide default-handler override, and the server disabling a trigger after a failure.
Adding a route that authors a trigger means adding a `record` call to it; nothing enforces that,
because the alternative — a database trigger — cannot see who or which client asked, and would
fire on every listener ping.

**Not recorded**, deliberately:

- **Cascades of renaming or deleting something else** — a script/flow rename rewriting
  `script_path` (`triggers::update_triggers_script_path`), a user being removed rewriting
  ownership. The event belongs to the runnable or the user, not the trigger.
- **Workspace-level bulk operations** — archive, fork clone, cross-workspace deploy. They move
  whole workspaces; a row per path would say nothing the workspace event does not.
- **Runtime housekeeping** — clearing `paused_until` / `error` after a run, consumer-offset state
  (Kafka `reset_offsets` / `commit_offsets` write `reset_offset` and `server_id`), the managed
  ducklake-maintenance schedule. Same category as the `server_id` / `last_server_ping` columns the
  diff already drops.
- **Native triggers** (`nextcloud`, `google`, `github`) — a separate, EE-gated subsystem keyed by
  external id rather than path. Not recorded, and they do not show the history button.

## Proposed follow-up (not in this PR)

The history drawer is a new entry point, so whether anyone finds it is an open question. A
`feature_usage` counter — register `("trigger_history", "open")` in `FEATURE_USAGE_KINDS`, log it
on drawer open with `key` = the trigger kind, and name it in the `InstanceSettings` disclosure —
would answer it. Left out here because it changes the telemetry disclosure copy; say the word and
I'll add it.

## Review round 1 fixes

- **Token scopes** (P1, all three reviewers). The endpoint enforced RLS but no scopes, so a
  scoped token could read every diff its *user* could see. It now needs `triggers_history:read`,
  with `check_scopes` on an explicit `path` and `build_scope_path_predicate` filtering the rows
  otherwise. (Without a registered domain the route also 400s for any scoped token, which is why
  the scope domain is part of the fix rather than just the handler.)
- **Forged history rows** (P1, Codex). `TriggerCrud::update_trigger` returns `Result<()>` and
  several impls do not check `rows_affected`, so an update aimed at a nonexistent or invisible
  path reached the recorder having changed nothing — and the insert policy is `WITH CHECK (true)`.
  Recording is now gated on the row actually being there after the write.
- **Pool-exhaustion shape** (P2, Claude). The queue's auto-disable took a second pooled
  connection while the job-completion transaction was held; it now records through a savepoint on
  that transaction. Both server disables are also gated on `rows_affected`, so a concurrent delete
  no longer produces a disable row for a path someone else may now own.
- **Payload cap** (P1, Codex). The 32 KiB limit cloned everything and then serialized the result
  just to measure it, and hand-built `changes` bypassed it entirely. Size is now accumulated as
  the diff is built, and capped again at the write itself.
- **`error` in the ignore list** (P2, Claude). Every trigger update clears `error`, so editing a
  previously-failed trigger recorded a spurious `error: {old: …, new: null}`.
- **Incomplete bulk diff** (P1 Codex / P2 Claude+Pi). The workspace-wide handler override rewrites
  up to five columns per schedule but recorded only the handler path. It now records all of them,
  documented as the write rather than a delta.
- **Authorization contract** (P1, Codex) and doc accuracy nits (Pi) on the public helpers.

## Review round 2 fixes

Round 2 came back "mergeable, but should ideally address nits" (Claude), with three nits, all fixed:

- **Scope filter ran after `LIMIT/OFFSET`.** A scoped token got short pages whose very shortness
  counted the rows it may not read. The grants now go into the `WHERE` via
  `build_scope_path_filter`, following `data_metrics.rs`.
- **The handler *reset* branch recorded values the `UPDATE` never wrote.** Clearing the workspace
  handler hardcodes `NULL` / `false` in SQL while the request still carries the form's other
  fields; `changes` is now built from whichever branch actually ran.
- **The third server-disable writer was ungated.** `worker_flow.rs` recorded on any `Ok`; it now
  checks `rows_affected` like the queue and listener paths.
- **A restore from the trashbin was not recorded** (P1, Codex). `restore_trash_item` reinserts the
  row directly, so after restoring a schedule or trigger its history still ended at `delete` for
  an object that exists again. It now appends a `create` built from the trashed snapshot, in the
  same transaction — no extra read.

## Design notes

- A rename is recorded as one `update` at the new path with `path` among the changed fields,
  rather than a delete plus a create. Rows written before the rename stay at the old path.
- `X-Windmill-Client` is attribution, not authority — exactly like `X-Windmill-Deploy-Origin`.
  Nothing reads a history row to make an access decision, so a caller lying about its client
  only mislabels its own row.

## Screenshots

The clock icon in the editor toolbar opens it — same slot on every kind (schedules and all
`TriggerCrud` triggers share `TriggerEditorToolbar`), immediately left of the AI-session icon and
before the enable/disable toggle. Only the surrounding buttons differ per kind:

trigger-history-button

The change viewer renders the payload's shape rather than the payload. `changes` only ever
arrives as one of three things — `null` (a delete), `{truncated_fields: [...]}`, or
`{field: {old?, new}}` — so it is parsed into that closed set once and every entry is laid out
the same way: one row per field, aligned in columns, marked `+` added / `−` removed / `✎` changed,
with scalars inline and a null on either side reading as added/removed rather than as a change to
`null`.

trigger-history-structured

Object and array values (`args`, `retry`, `*_extra_args`) stay one line until expanded, then show
before and after side by side:

trigger-history-expanded

## Test plan

Exercised end-to-end on a local instance (backend `--features quickjs,http_trigger,websocket`):

- [x] Schedule create with `X-Windmill-Client: cli` → `source=cli`; update/setenabled with the
      browser session token → `source=ui`; setenabled with a plain API token → `source=api`.
- [x] Real `wmill schedule enable` run from source → `source=cli`.
- [x] HTTP trigger create / setmode / rename-update / delete → one row each, the rename showing
      `path: {old, new}`, the delete showing no diff.
- [x] Schedule pointing at a deleted script → the queue disables it → `source=worker`,
      `username` unset, the error in `changes`.
- [x] RLS verified directly against the policies: a non-member sees nothing; a group member sees
      `g/` paths; a folder reader sees `f/` paths; a workspace admin sees everything.
- [x] Drawer driven with Playwright on both the schedules and HTTP-routes pages (screenshots
      above), including expanding an object-valued field.
- [x] `parseChanges` unit-tested on all three payload shapes (`npx vitest run`, 2 pass).
- [x] Invariant proven by injecting a constraint that fails the history insert (above); backend
      unit tests 5 pass, including the NUL guard.
- [x] All three server-initiated disable writers now exercised end-to-end: the queue path
      (script schedule), the `worker_flow` fallback (flow schedule, forced by failing only the
      first insert), and the trigger listener (websocket).
- [x] `tx.begin()` confirmed to emit `SAVEPOINT`, not `BEGIN`, by reading the statements Postgres
      received during a real auto-disable: one connection, one `COMMIT`.
- [x] Scoped tokens, after the round-1 fix: a token scoped to one path lists only that path,
      is refused (403) on another path, and a token holding `schedules:read:*` but no
      `triggers_history` scope is refused at the route.
- [x] An update aimed at a nonexistent trigger path records no history row.
- [x] A 60 KB `args` update stores the field-name list (172 bytes), not the payload.
- [x] The default-handler override records all five error-handler columns, and *clearing* it
      records the NULL/false the write actually produced rather than the form values the request
      still carried.
- [x] With the scope filter in SQL, a path-scoped token gets full pages of its own rows
      (`per_page=3` → 3 rows, page 2 → the remainder) instead of a page truncated by a
      post-fetch filter.
- [x] Deleting and restoring both a schedule and an HTTP trigger records `delete` then `create`,
      the trigger one under `trigger_kind = http`.
- [x] `http_triggers/create_many` records one `create` per trigger.
- [x] With the preimage locked, a concurrent commit between snapshot and mutation no longer leaks
      into the diff (see round 4).
- [x] `cargo check --features quickjs`; `cargo check` with every trigger feature plus
      `private,enterprise`; `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`;
      `cargo test -p windmill-api-auth` (42 pass); `npm run check` (0 errors); CLI unit tests
      (1063 pass).

## Review round 3

Pi: **good to merge**. Codex and Claude each found more *unrecorded write paths* — the third round
running in a row where the finding was "here is another call site". That is the shape problem the
per-call-site design has, so rather than patch one more and invite a fourth round, this round
draws the line above and writes it into the module docs. `http_triggers/create_many` moved inside
it (it is authoring); the rest are named as out, with reasons.

## Review round 4

Claude and Pi: **good to merge**. Codex found one real race, now fixed: `snapshot_row` took the
preimage with a plain `SELECT`, so another request committing between the snapshot and the
mutation put its change into this caller's diff under this caller's name. The preimage is now
taken `FOR UPDATE`. The lock is on the trigger row and the callers touch the job queue only
afterwards, which is the order the schedule paths already document.

Verified by holding a `FOR UPDATE` lock on a schedule from a second session, changing its cron
there, and firing `setenabled` against it: the request blocks, and its history row contains only
`enabled: {old: false, new: true}` — the concurrent cron change is not attributed to it.

## Review round 5

Pi: **good to merge**. Claude: nit only. Codex found one more real race, fixed:

- **The auto-disable claimed a transition it did not make.** `try_schedule_next_job` reads the
  schedule as enabled, but a user can disable it before the disable `UPDATE` runs; without a
  predicate that `UPDATE` still reports one affected row, so the `rows_affected` gate let the
  worker record `enabled: {true → false}` on top of the user's own event. Both auto-disable
  updates now carry `AND enabled = true`, so the write and the history agree. Verified: replaying
  the race against the predicate yields `UPDATE 0`, and the normal auto-disable still records.
- **Nit (Claude):** round 4's `FOR UPDATE` starts the row lock before
  `TriggerCrud::update_trigger`, and the postgres / gcp / azure impls do remote work before their
  own `UPDATE` — so the lock now spans that call. Left as is (narrowing it means a preimage inside
  each impl, two of them EE) and written into `snapshot_row`'s doc, since that doc is what the
  next person reasons from.

## Review round 6 — clean

Claude and Pi: **good to merge**. Codex: nit only (the same rationale repeated at four call
sites); the invariant now lives once on `TriggerHistoryEvent::server_disable` and the copies are
gone.

Findings across the six rounds: 7 → 3 → 1 (plus the boundary question) → 1 → 1 → 0 blocking.

## Server-initiated disables: the disable wins, a lost row is loud

When the server disables a trigger it could not run, two things want to be true and cannot both be
guaranteed: the trigger ends up disabled, and the history says who did it. **The disable wins.**

A trigger left enabled reads as healthy while never firing again — and for a flow schedule nothing
comes back to retry, because it arms its next occurrence when the flow *starts*, so once the
runnable is gone that code is never reached again. Enabled-and-dead is silent. Disabled without an
audit row is not: the trigger is visibly off and its own `error` column says why.

So each writer commits the disable on its own and records after it. On the success path both land
together — the queue path shares one commit through a savepoint, which also keeps a failed insert
from poisoning the transaction that completes the job. Where the disable has already committed the
insert is **retried** (3 attempts, backing off) before giving up, since by then a retry cannot
invent a row for a mutation that did not happen. Only if it still fails is the trigger left
disabled without its row, and that is reported to the workspace error handler and the critical
alert channel.

The listener's retry is **spawned rather than awaited**, which testing showed is not optional:
`disable_with_error` runs inside a `tokio::select!` whose ping branch exits the instant the disable
clears `server_id`, so every await after the disable is a cancellation point. An inline retry was
cancelled after its first attempt — and took the alert with it, silently. Detaching it is safe
precisely because the disable is durable by then.

Supporting the same goal: `changes` is stripped of NUL bytes at the write (jsonb rejects them and
`changes` quotes caller-supplied text) and capped, so the insert has almost no way to fail in the
first place.

**Verified by breaking the insert on purpose** — `CHECK (source <> 'worker') NOT VALID` on
`trigger_history`, across all three writers:

| writer | result with the insert broken |
|---|---|
| queue path (script schedule) | `enabled = f`, `error` set, 0 history rows |
| `worker_flow` fallback (flow schedule) | `enabled = f`, `error` set, 0 history rows |
| trigger listener (websocket) | `mode = disabled`, `error` set, 0 history rows |

Each also raised `Disabled … but could not record it in the trigger history: …` in the `alerts`
table. With the constraint dropped, the healthy path disables *and* records as normal.

The retry was verified separately with a trigger that rejects only the first two inserts for a
path (a sequence tells the attempts apart, since `nextval` survives rollback): the trigger ended
`disabled` with its row present, tagged `landed-on-attempt-3`. Before the spawn fix the same probe
produced one attempt, no row and no alert.

## Left in draft

Deliberately, not for lack of a clean round. Four of the "ask a human first" triggers apply at
once: a migration under `backend/migrations/`, a new API endpoint with an `openapi.yaml` change and
a regenerated client, a new token-scope domain (`triggers_history`) in the auth path, and edits to
the job-completion path in `windmill-queue` / `windmill-worker`. Any one of those is worth an
owner's eye before this is marked ready.


rubenfiszel:
Round 3 addressed in f1d0573.

**Fixed:** `http_triggers/create_many` now records a `create` per trigger, same as the single-create route (verified: a two-trigger bulk create with `X-Windmill-Client: cli` writes two `create` rows, `source=cli`, with the initial-shape diff).

**Scoped out, with the reasons now in the module docs** rather than patched one call site at a time. Three rounds in a row the finding has been "here is another write path" — Kafka `reset_offsets`, the fork clone, the ducklake-maintenance schedule, and ~20 non-test files write these tables in total. Per-call-site wiring cannot make "every write is recorded" true, and the alternative that could — a database trigger — cannot see who or which client asked, and would fire on every listener ping. So the boundary is now stated instead of implied:

- **Recorded:** authoring a single trigger through its own surface (create incl. bulk, update, delete, enable/disable/suspend, trashbin restore, the workspace-wide default-handler override) plus the server disabling one after a failure.
- **Not recorded:** cascades of renaming/deleting something else (a script rename rewriting `script_path`, a user removal rewriting ownership); workspace-level bulk operations (archive, fork clone, cross-workspace deploy); runtime housekeeping (`paused_until`/`error` clearing, Kafka `reset_offset`/`server_id`, the managed ducklake schedule) — the same category as the `server_id`/`last_server_ping` columns the diff already drops.

@claude the two paths you named (fork clone, ducklake-maintenance schedule) are in that list now.

rubenfiszel:
Both round verdicts above were on `f519da1e` and predate the two commits that fix them.

**Codex P1 — flow fallback strands enabled schedules.** Fixed in `e3e6fc7`, before that review posted. The fallback no longer wraps the disable and the insert in one transaction: the disable commits on its own via `.execute(db)`, and the history row is recorded after it. A failing insert can no longer roll the disable back, so nothing is stranded enabled-and-unarmed. `1adc514` then added 3 retries with backoff, and only a persistent failure alerts.

**Pi P2 — the doc's "retried on the next occurrence".** Also fixed in `e3e6fc7`, which rewrote that module doc; it now says the opposite (the disable wins, a lost row is reported).

**Codex P2 / Pi P2 — raw `<button>`.** Valid and now fixed in `82ca9628`: the expander uses `Button` with `variant="subtle"`, `unifiedSize="2xs"`. Verified in the browser that the grid alignment and the expand/collapse still work.

One thing that came out of testing the retry, worth flagging for reviewers because it is not obvious from the diff: in the listener, `disable_with_error` runs inside a `tokio::select!` whose ping branch exits the moment the disable clears `server_id`. Every await after the disable is therefore a cancellation point — an inline retry was cancelled after attempt 1 **and took its alert with it**. The recorder there is now spawned rather than awaited (safe, since the disable is durable by then). Probe evidence: rejecting the first two inserts yields a row tagged `landed-on-attempt-3`; rejecting all of them yields a disabled trigger plus a `could not record it in the trigger history` alert.

`cargo_test` is red on `test_bun_nobundling_transitive_require` (worker timed out after 60s). It fails the same way on `main` — see run 31803525622 on `main` @ `6d03784d` — so it is not from this branch.

rubenfiszel:
All of round 7 addressed in `eaefa36`.

**Pi P1 / Codex P1 — disable history can attach to a replacement trigger.** Real, and mine: making the disable commit on its own put the insert on a second pooled connection, so the row lock was released before the row was written. All three server-disable writers now put the `UPDATE` and the record in **one transaction**, with only the insert in a savepoint (`record_in_disable_tx`). The lock is held across the pair, so a delete-and-recreate at the same path cannot slip in between — while the disable still wins, since a failing insert rolls back to the savepoint and the outer transaction still commits.

That also removes the two things the previous shape needed and Pi flagged: the retry (its 200/400ms sleeps widened the very window at issue, and it had no idempotency key, so an ambiguous commit could duplicate a row) and the detached spawn (the listener's cancellation problem disappears once the work is one short transaction — the ping loop can no longer observe a cleared `server_id` mid-flight, because nothing is committed until the end).

**Codex P1 — stale listener disable history.** The listener's `UPDATE` now carries `AND mode <> 'disabled'`, so a user disabling first yields `rows_affected = 0` and no worker row. Verified: pre-disabling the trigger produced only the `create` row, no `worker` row.

**Codex P1 — non-atomic bulk override.** `set_default_error_handler` now runs its six branch `UPDATE`s and `record_bulk` on one transaction and commits once.

**Claude / Pi P2 — stale docs.** The module overview no longer claims the client kind is unforgeable (it is caller-declared; `of_request` always said so). `strip_nuls` and its test no longer assert the reversed atomicity policy. The policy section now describes the shape as it is.

Verified end-to-end after the rewrite: listener disable records; a user winning the race produces no worker row; the bulk override writes 9 rows across 9 paths in one commit; the queue path (script schedule) records with the error in the diff.

`cargo_test` remains red on `test_bun_nobundling_transitive_require`, which fails identically on `main` (run 31803525622 @ `6d03784d`).

rubenfiszel:
Round 8 addressed in `693bde0`. Pi: good to merge. Codex P1 + the shared nit:

**P1 — cancelable history-loss alert.** Correct, and the same cancellation trap one step later: the commit makes the cleared `server_id` visible, the biased `loop_ping` branch finishes, and `disable_with_error` is dropped — taking the awaited alert with it. So a savepoint-failed insert could have left the trigger disabled with neither a row nor a durable signal. The alert is now spawned. Safe to detach: it is a report, not a mutation, and the disable is already committed.

Verified with the insert broken: trigger `disabled`, and `Disabled websocket trigger u/admin/g2_ws but could not record it in the trigger history: …` reached the `alerts` table.

**P2 (Codex) / nit (Claude) — duplicated rationale.** The lock/savepoint/"disable wins" contract now lives once, on `record_in_disable_tx`. The four call sites carry only what is local to them: the queue path notes that `tx` is the job-completion transaction, `worker_flow` notes it is the last chance to disable, the listener notes why the alert is spawned.

Re-verified after the rewrite: listener disable records normally; the alert lands when the insert fails. `cargo_test` remains red on `test_bun_nobundling_transitive_require`, which fails identically on `main`.

rubenfiszel:
Round 9: Claude and Pi both **good to merge**. Codex's P1 fixed in `2b5a03c9`.

**P1 — history drawer used the navigation workspace.** Real. `triggerWorkspace.ts` is an explicit seam whose own doc says "every workspace-scoped backend call / navigation inside the trigger subtree reads its workspace via `getTriggerWorkspace`", and the new button read `$workspaceStore` directly — so in an AI session editing a forked workspace the drawer would have queried the nav workspace instead. It now uses the house idiom the siblings use:

```ts
const triggerWs = getTriggerWorkspace()
const wsId = $derived(triggerWs?.() ?? $workspaceStore)
```

Behavior outside such a host is unchanged by construction (no resolver → falls back to `$workspaceStore`); verified in the browser that the drawer still loads normally. `npm run check` clean.

PR #10698, opened 2026-08-14
perf: declare a settings pass instead of reading one setting at a time

rubenfiszel:
## Summary

`initial_load` (`backend/src/monitor.rs`) read several dozen instance settings one at a time, each its own `SELECT value FROM global_settings WHERE name = $1`. Measured on a CE debug build against a localhost Postgres: **50 serialized round trips** before a worker reaches `listening for jobs`, ~28 for a server's pass. On localhost that is ~20 ms and invisible; against a real database it is 50x the RTT per process start, which `EXIT_AFTER_N_JOBS` turns into a per-job cost. An agent worker had it worse: ~36 sequential HTTP requests.

The call sites now declare what they will read, and the pass fetches it in one go.

```rust
let mut pass = SettingsPass::new();
pass.settings(&[OAUTH_SETTING, BASE_URL_SETTING], false, move |v| async move { … });
pass.option_setting(JOB_DEFAULT_TIMEOUT_SECS_SETTING, "JOB_DEFAULT_TIMEOUT_SECS", JOB_DEFAULT_TIMEOUT.clone());
if *CLOUD_HOSTED { pass.setting(CONCURRENCY_KEY_MAX_QUEUED_SETTING, false, …); }
pass.action(write_maven_settings_xml());
pass.run(conn).await;
```

## Why declaring, rather than a list of names to prefetch

Demand is exact: the same `if server_mode` / `if *CLOUD_HOSTED` / `cfg` branches that used to guard a read now guard a declaration, so the fetch asks for what this process actually needs and there is no list to drift. It also covers `global_settings`' dynamically named rows (`workspace_dependencies_map_rebuilt:<workspace_id>`, one per workspace, no cleanup path) for free: nothing declares them, so nothing fetches them, and the query stays on the primary key.

## Changes

- `SettingsPass` with three ordered step kinds: `.setting` (one read), `.settings` (several one piece of work needs together, which is what `base_url` needs since it consumes `oauths` and `base_url` at once), and `.action` (non-setting work that must keep its position in the sequence).
- `.run(conn)` fetches once, then replays appliers in declaration order. Ordering invariants survive unchanged, including the workspace-fairness knobs applying before the enabled flag.
- Transports: one `WHERE name = ANY($1)` over a database connection; a concurrent fan-out over the existing per-setting endpoint for `Connection::Http`. A real batch endpoint can replace the latter behind `fetch_settings_batch` without touching a call site.
- The `reload_*` functions keep working per setting for `process_notify_event`, sharing their apply half with the pass. The six no caller was left using are removed.

## The three states a batch has to distinguish

A per-setting read only ever produced two outcomes at a given call site, so the batch has to be explicit about a third, and both bugs found while building this came from that:

- **A read that failed is not a setting that is unset.** `load_workspace_fairness_max_percent` documents why: a transient DB error must leave the atomic alone, while a cleared setting must restore the default so deleting it in the admin UI takes effect. Collapsing both into "absent" would let one failed query reset fairness and the queue caps across a cluster. A failed read is now left out and skips its applier.
- **A setting an agent worker may not request still reads as unset**, which is what `load_value_from_global_settings_with_conn(.., false)` returned for it. Skipping it instead would have left `disable_password_login` and friends never storing their default; it coincided with the static initial values, so nothing observably broke.

## Measurements

| | before | after |
|---|---|---|
| worker startup | 50 queries | **1** |
| `initial_load`, server mode | ~28 | **1** batch |
| agent worker | ~36 sequential HTTP requests | **1 concurrent round** |

`jwt_secret` was the one read that could not be batched, because it decided from its own read whether to generate and upsert a secret, so a stale value let a booting replica overwrite one another replica had just written. `get_or_create_jwt_secret` moves that decision into the statement (`ON CONFLICT ... WHERE jsonb_typeof(value) <> 'string' RETURNING value`), so first writer wins, a usable secret is never overwritten, and the read can join the batch. The `WHERE` also means a normal startup performs no write, which matters because `notify_global_setting_change` fires on every write to this table.

`exec` → `listening for jobs` is unchanged in the mean, because it is dominated by a ~390 ms blocking HTTPS call to `hub.windmill.dev` — a separate finding, not touched here.

## Test plan

- [x] Worker start: 50 → 2 queries, stable over repeated runs; the 27 settings the pass resolves are byte-identical before/after (diffed the `Loaded setting …` log lines).
- [x] Server start: `initial_load` down to one batch; the remaining singles are `main.rs` reads that happen before `initial_load` is called and were in the original count too.
- [x] Agent worker (EE build, `MODE=agent` against a `MODE=server` instance): loads the same setting values over HTTP, reaches `listening for jobs, agent mode`. The 34 batched requests each show 25–33 ms latency and all complete within the same ~7 ms window, i.e. one concurrent round rather than 34 sequential ones.
- [x] Live notify path: setting `job_default_timeout` and clearing it in `/#superadmin-settings` both reach the running worker (`Number(777)`, then `to None`).
- [x] A real bash job runs to completion on the rebuilt backend.
- [x] `cargo check` CE and `--features enterprise,private,parquet,quickjs`; `SQLX_OFFLINE=true` for both `windmill-common --all-targets` and the bin.
- [x] `windmill-common/tests/global_settings_batch.rs`: the batch returns only rows that exist and only names that were asked for, and reports a failed read as an error rather than an empty result that would read as every setting unset.

## Note

`backend/src/main.rs` still builds a throwaway max-2 pool via `db_connect::initial_connection()` and drops it before `connect_db`, costing one extra Postgres handshake per process start. Measured at ~0.9 ms on a localhost database (pgbench: 0.915 ms/txn with `-C` versus 0.008 ms on a persistent connection). It is not removable safely: `connect_db`'s `max_connections` depends on `num_workers`, which for a worker that is not native-from-env is read from the database, and sqlx fixes `max_connections` at construction with no way to move an established connection between pools. The alternatives are keeping both pools alive (the exhaustion the comment at `main.rs:1024` guards against) or decoupling pool sizing from the database, which changes how native mode resolves. Left alone deliberately.


rubenfiszel:
✅ Review round clean @ 0fea48b23cb7e4538b66b8e574da82bf8fd165d3

rubenfiszel:
Round on `b1aa0fc846`: Claude and Pi "Good to merge", Codex nit-only (dangling `reload_setting` doc links), fixed in `fa48f2acfb`.

Left in draft deliberately for a human read of the step sequence in `initial_load` before merging: three rounds surfaced ordering bugs in this area (audit-partition retention, the audit-export anchor, and the object-store OIDC issuer all read settings the pass had not applied yet), so the ordering of declarations and `action` steps is the part most worth a second pair of eyes.

rubenfiszel:
✅ Review round clean @ b1aa0fc8468a3eb3e8434bb4c8d1890f3979753a — nit-only verdicts at that SHA (Claude and Pi "Good to merge", Codex nit-only); the dangling doc links are addressed in fa48f2acfb.

rubenfiszel:
✅ Review round clean @ de82ee65abb0f62408becc06cd097e75dc24203f

PR #10700, opened 2026-08-14
perf: back off the interactive worker shell under EXIT_AFTER_N_JOBS

rubenfiszel:
## Summary

Part of a series on worker process startup cost, motivated by `EXIT_AFTER_N_JOBS` (at
`EXIT_AFTER_N_JOBS=1` there is one full process start per job, so anything the worker does once
per process becomes a per-job cost).

Three items were investigated; two produced a change.

**(a) The interactive worker shell never backed off.** `start_interactive_worker_shell` is a
second pull loop, on the worker's own name prefix, that backs off from ~500ms to 15s only after
`TIMEOUT_TO_RESET_WORKER_SHELL_NAP_TIME_DURATION` (120s) without a shell job. A worker recycled
after N jobs never lives that long, so it polled its shell tag twice a second for its entire life,
every life. It now starts in the backed-off state, and latches to the fast cadence as soon as it
actually serves a shell job.

The shell backs the workers-page REPL, so it is not removed: a worker under `EXIT_AFTER_N_JOBS`
still pulls its shell tag once per process start, and the first command of a REPL session waits at
most 15s. That is the same latency a plain worker already has today once it has been idle past
those 120s; the change only makes a recycled worker start there instead of never getting there.

**(b) Init and periodic scripts re-run on every restart, invisibly.** They are deliberately not
counted towards the limit (they are what prepares the fresh environment), which is correct and
unchanged. But with `EXIT_AFTER_N_JOBS=1` and an init script configured, every job is now preceded
by a pushed-and-executed init job whose full runtime lands on that job's latency, and a periodic
script fires once per process start whatever its interval says. Nothing told the operator. Two
startup warnings now do, alongside the existing `num_workers`/dedicated-worker ones.

**(c) Pool sizing: no change, measured.** The worker pool is `min_connections(0)` with a cap of
`DEFAULT_MAX_CONNECTIONS_WORKER + (num_workers - 1)` = 5. A per-job process opens 3 of those (plus
1 for the throwaway initial pool, which is another change's territory), and the three are genuine
concurrency: config/global-settings loading at startup, and the ping loop overlapping the first
pull. The cap is not the binding constraint, so lowering the default saves nothing and only
serializes those users onto fewer connections. Reported rather than churned.

## Changes

- `interactive_shell_nap` extracted from the shell loop and given the process start instant
  explicitly, so the 120s backoff is anchored to process start rather than to a seeded
  `last_executed_job`, and a worker with `EXIT_AFTER_N_JOBS` set starts already napping.
- Startup warnings when `EXIT_AFTER_N_JOBS` is set together with an init script or a periodic
  script, in the existing `EXIT_AFTER_N_JOBS` warning block.
- README `EXIT_AFTER_N_JOBS` row: says that the uncounted init/periodic scripts therefore re-run on
  every restart, and that the worker shell starts backed off.
- One unit test on the nap policy (the backoff anchor is exactly what a "simplification" here
  breaks).

## Measurements

Local CE debug build, Postgres on localhost, worker on its own tag.

Interactive-shell pulls over 180s idle:

| | before | after |
|---|---|---|
| `EXIT_AFTER_N_JOBS=1` | 486 | **12** |
| unset | 486 | 486 (unchanged, same per-20s profile: ~80/20s for 120s, then ~1/15s) |

Full `EXIT_AFTER_N_JOBS=1` run against a real queued job, exec -> `listening for jobs` and
last-job-completed -> process exit, 3-5 runs each: startup 197-538ms before / 208-524ms after (the
spread is the hub call and background load, not this change), shutdown 214-219ms before /
214-216ms after. Joining the interactive shell during shutdown takes 27µs: the 15s nap sits in a
`tokio::select!` on the killpill, so it does not stall teardown.

REPL behaviour under `EXIT_AFTER_N_JOBS=1`: first command 14.1s, second 0.28s, third 0.28s. On an
unmodified binary, a plain worker left idle 130s answered its first shell command in 3.6s, i.e.
from the same 0-15s window.

With an init script and a periodic script configured, both warnings fire once at startup, one init
job and one periodic job are pushed per process start, and a 1s init script pushed the process's
total life from ~750ms to ~1840ms with the user job recording `wait_time=1608ms`.

## Test plan

- [x] `cargo test -p windmill-worker interactive_shell_nap` passes
- [x] `EXIT_AFTER_N_JOBS=1` worker, 180s idle: 12 shell-tag pulls (was 486)
- [x] plain worker, 180s idle: 486 shell-tag pulls, fast for the first 120s then ~1/15s, matching
      the pre-change profile
- [x] `EXIT_AFTER_N_JOBS=1` worker against a real queued job: startup and shutdown timings
      unchanged, shell join takes microseconds
- [x] REPL against an `EXIT_AFTER_N_JOBS=1` worker: first command served within 15s, following
      commands immediate
- [x] worker group with an init script and a periodic script: both warnings fire, both jobs are
      pushed per process start

---
## Summary by cubic
Backs off the interactive worker shell and adds startup warnings under `EXIT_AFTER_N_JOBS`. Before: the shell always polled ~2Hz until 120s idle (recycled workers never got there) and init/periodic re-ran silently; now only a live shell session gets the fast cadence, the loop naps 5s before the first command (15s with `EXIT_AFTER_N_JOBS`) and 15s after 2 minutes idle, and we warn that init adds per-batch latency and periodic runs once per process start.

- Extracts `interactive_shell_nap`: anchors backoff to process start, gives sub-second cadence only after the first served shell job, and starts backed off when `EXIT_AFTER_N_JOBS` is set. First REPL command may wait up to 5s (plain workers) or 15s (`EXIT_AFTER_N_JOBS`); subsequent commands are immediate while the session is active.
- Emits startup warnings when `EXIT_AFTER_N_JOBS` is set and the worker group has an init or periodic script: init runtime is added to each batch’s latency; periodic runs once per process start.
- Updates the `EXIT_AFTER_N_JOBS` README row to document these behaviors.
- Adds a unit test for the nap policy.

<sup>Written for commit 541d2ab89494fed658a2d2b0b276e89643a4181b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 936445d328b34e4ca8b152ec2bbee035fbc380e6 — nit-only verdicts at c167a25a172833ae9fe17a18eae05d31604ca875 (codex and pi "Good to merge", claude nit-only); both nits addressed in 936445d328b34e4ca8b152ec2bbee035fbc380e6

PR #10703, opened 2026-08-14
fix(agents): stop the scratch-dir guards prompting on quoted text

rubenfiszel:
## Summary

Follow-up to #10702. The guards now own the prompt for `rm`, `mv` and `chmod`, and `runs_verb`
decides whether a command runs one. That scan is textual, so it read *text about* a verb as a
verb and prompted. Two shapes hit within minutes of the merge, both captured from real
transcripts and now in the table:

- `gh pr create --body "$(cat <<'EOF' … )"` — the body mentions `` `rm` `` in backticks, and
  backticks are a separator so that `` `rm -rf x` `` is caught. Markdown prose became a command.
- `claude -p "…⏎1: rm -rf /tmp/a"` — the token `1:` matched the `[0-9]*` arm, which exists for
  `timeout 5 rm`, so the scan walked past it into quoted prose.

Which commands are *allowed* is unchanged. This is mostly a false-positive fix, with one real
reduction in coverage worth naming: a verb written into a heredoc body no longer prompts, so
`cat > /tmp/s.sh <<'EOF' … rm -rf ~/x … EOF` is silent where it used to ask. That protection was
thin anyway, since the `bash /tmp/s.sh` that follows never prompted either.

## Changes

- `strip_heredoc_bodies`: drop heredoc bodies before scanning, since a body is data. Dropping
  one needs both a delimiter that could really open a heredoc and a line that terminates it, so
  a `<<` inside prose (`echo "cat <<EOF"`) or in an arithmetic shift leaves the commands around
  it intact.
- The `[0-9]*` skip now applies only past a wrapper, where a bare number is a duration.
- The post-wrapper scan is bounded, but only ordinary words are charged against it. Flags, env
  assignments and durations are already known not to be commands, and charging them ended the
  scan before the verb in `env A=1 … F=6 rm` and `sudo -E -H -u root FOO=1 BAR=2 rm`.
- Braces are words rather than separators. Splitting on them cut `xargs -I {} … rm` in half and
  stranded the `rm` in a segment that no longer knew a wrapper preceded it.

## Test plan

- [x] `bash .claude/hooks/test-hooks.sh` — 60 rows
- [x] The two field commands replayed verbatim from the transcripts: both `none`, previously `ask`
- [x] Every reviewer repro prompts: `env -i` with 5 assignments, `sudo -E -H -u root FOO=1 BAR=2`,
      `xargs -a f -d d -E e -I {} -L 1 -n 1`, and the quoted `<<` lookalike
- [x] Still prompts where it must: a real command after a heredoc, an unterminated `<<` followed
      by a real `rm`, and every row from #10702

## Known limitations

`runs_verb` scans text rather than parsing shell, so two things are true by design:

- **It does not track quote state across lines.** A command that opens a quote, mentions
  `<<WORD` at the end of a line inside it, closes the quote later, contains a destructive verb
  and also contains a line that is exactly `WORD` will not prompt. Closing that needs a
  quote/`$( )` lexer, a larger risk than the case it covers, since its failure mode is silent.
  When it does happen the guard falls silent rather than allowing anything, so the auto-mode
  classifier still evaluates the command.
- **A heredoc body is only treated as data for a known reading consumer** (`cat`, `tee`, `gh`,
  `jq`, `python`, …). Anything else — including a shell, which executes the body — keeps the
  body in the scan. An unrecognized consumer therefore costs an extra prompt, which is the
  direction that fails safely.


rubenfiszel:
Fixed in 1a4c8622c0: the post-wrapper bound is gone entirely, so `sudo -u "root" rm -rf /etc`,
`sudo -u 'root' rm`, `env -u A … -u G rm` and `env A=1 … F=6 rm` all prompt again. Both bounds I
tried (a word count, then stopping at the first quoted word) failed the same way, because a
wrapper's option value cannot be told from an argument. Checking the two field transcripts that
motivated this PR shows the bound was never what fixed them — the `[0-9]*` narrowing was — so it
bought nothing and cost three false negatives. Table row added for the quoted form.

## Dismissing the multiline quoted heredoc lookalike

Not fixing this one, deliberately:

```bash
echo "cat <<EOF
prose"
rm -rf /etc
EOF
```

Closing it means tracking quote state across lines, and it can't be done by counting quotes:
`gh pr create --body "$(cat <<'EOF'` has an odd number of `"` before the `<<` and *is* a real
opener, because `$(` starts a fresh context. Doing it correctly is a small shell lexer with a
quote/`$( )` stack inside a guard whose failure mode, when wrong, is silent. Five rounds have
established that every one of these findings is the same root cause — a textual scanner meeting a
construct it doesn't parse — and each fix has been safe-direction, but a lexer is a different
class of risk than the bug it closes.

The residual: a command must open a quote, mention `<<WORD` at a line end inside it, close the
quote later, contain a real destructive verb, and contain a line that is exactly `WORD`. When it
happens the guard falls silent rather than allowing anything — the auto-mode classifier still
evaluates the command, as it did for `rm` everywhere before these hooks existed.

Happy to build the lexer if the maintainer would rather have it; flagging the tradeoff instead of
making that call silently.

rubenfiszel:
Both round-9 findings are fixed in 5ecdd28c4d, and the fix inverts the check rather than
extending it. Naming shells to avoid (`bash`, `sh`, …) could never be complete — `ash`, `rbash`,
`busybox sh`, or a `bash \` line continuation that hides the word entirely all executed a body
the guard was calling data. It now requires positive evidence of a *reading* consumer (`cat`,
`tee`, `gh`, `jq`, …) before dropping a body, checking the command the `<<` belongs to and
anything the body is piped into. An unrecognized consumer costs a prompt instead of losing one.

Verified: `bash \`+`<<'EOF'`, `ash`, `busybox sh`, `sudo -s`, `(bash <<'EOF')`, `cat <<'EOF'|bash`
all prompt; `cat <<'EOF' > f` and `cat <<'EOF' 2>&1 | tee f` stay silent; both original field
transcripts stay silent. 85 rows in `.claude/hooks/test-hooks.sh`.

Merging with the two documented limitations in the description rather than running a tenth round.
Nine rounds found real defects every time, but the last few have been narrowing variants of one
root cause — a textual scanner meeting shell constructs it doesn't parse — and closing that class
properly means a real lexer inside a guard whose failure mode, when buggy, is silent. That is a
worse trade than the residual. Noting for the record that 5ecdd28c4d itself has not been through
a review round.

PR #10708, opened 2026-08-14
fix: drop sampling params on Claude models that reject them

hugocasa:
## Summary

Follow-up to #10690, verified against the live Anthropic Messages API.

Claude Opus 4.7/4.8 and the 5 family removed the sampling params: any
`temperature` is a hard 400 — `` `temperature` is deprecated for this model `` —
**whatever the thinking mode**. Both provider paths only dropped it under
adaptive thinking, so two request shapes 400 today.

## The two failure modes

Probed against `api.anthropic.com` (bodies emitted by the code itself, then POSTed):

| Request | Result on merged main |
|---|---|
| `claude-opus-5`, reasoning off, `temperature` set | `400 temperature is deprecated for this model` |
| `claude-opus-4-8`, no reasoning at all, `temperature` set | `400 temperature is deprecated for this model` |

The first is a regression from #10690: the off sentinel added
`thinking: {type: "disabled"}` and kept `temperature`, on the mistaken premise
that sampling params are only rejected alongside *adaptive* thinking. They are
rejected by the *model*.

The second predates #10690 (the default `None` arm passed `temperature`
straight through) and hits any agent step on Opus 4.7/4.8 that sets a
temperature and leaves reasoning unset.

## Changes

- `providers/mod.rs`: `anthropic_model_rejects_sampling_params(model)` — one
  home for the model list (Opus 4.7/4.8, Opus 5, Sonnet 5, Fable, Mythos).
- `providers/anthropic.rs`: `anthropic_thinking_config` takes the model and
  drops `temperature` for those models on every arm, not just adaptive.
- `providers/bedrock.rs`: same rule via `bedrock_temperature`, applied at all
  three Converse call sites. Non-Anthropic Bedrock models (Nova, Llama) keep
  their sampling params.

## Verification

Bodies were emitted by the real code path and POSTed to the live API:

| Body the code builds | Live result |
|---|---|
| `opus-5` `thinking=disabled`, no effort, no temp | **OK** |
| `opus-5` `thinking=adaptive` `effort=xhigh`, no temp | **OK** |
| `opus-4-8` no thinking, no temp | **OK** |
| `sonnet-4-6` `thinking=disabled`, `temperature=0.5` | **OK** (4.6 still accepts sampling) |

Same probe also confirmed two things #10690 asserted from documentation only:
- `thinking: {type: "disabled"}` is accepted on Opus 5 and Sonnet 5, and returns
  no thinking block — the off switch works.
- `output_config.effort: "none"` is rejected (`Input should be 'low', 'medium',
  'high', 'xhigh' or 'max'`), and a disable paired with `xhigh` is rejected
  (`Use effort 'high' or below, or enable thinking`) — the sentinel translation
  and the no-effort-with-disable rule were both necessary.

## Test plan
- [x] `cargo test -p windmill-ai --features bedrock` — 131 passing, including
      per-model sampling coverage on both paths and a non-Anthropic Bedrock case
- [x] Live probe against `api.anthropic.com` (table above)
- [ ] Bedrock leg of the sampling rule is inferred from the shared model, not
      probed — no AWS credentials here

---
## Summary by cubic
Drops `temperature` on Claude models that removed sampling params across Anthropic and Bedrock to prevent 400s when reasoning is off or unset. Also aligns Bedrock behavior: Sonnet 5 cannot disable thinking on Bedrock, unlike the native Anthropic API.

- Centralizes detection in `anthropic_model_rejects_sampling_params` (Opus 4.7/4.8, Opus 5, Sonnet 5, Fable, Mythos). Bedrock application inference profiles with opaque IDs keep `temperature`.
- Anthropic: `anthropic_thinking_config` now takes the model and drops `temperature` on rejecting models on every path; adaptive thinking still drops it on all models.
- Bedrock: `bedrock_temperature` applies the same drop rule in streaming, non‑streaming, and builder paths. Non‑Anthropic Bedrock models (e.g., `amazon.nova-pro-v1:0`) keep `temperature`; Sonnet 4.6 still accepts it. Sonnet 5 cannot be disabled on Bedrock; the native provider still supports disable.
- Verification: live probes cover `claude-opus-5` (disabled/adaptive) and `claude-opus-4-8`; tests exercise both providers, a non‑Anthropic case, per‑model temperature behavior, and build the disable request via the resolver.
- UI: updates Gemini default model id to `gemini-3.1-pro-preview` and reflects the Bedrock Sonnet 5 disable rule in the reasoning registry.

<sup>Written for commit 940a80d8255cc8500840c251afa57d0a44250b20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10708?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -1,3 +1,4 @@
+use super::{anthropic_model_rejects_sampling_params, REASONING_OFF_SENTINEL};
 use crate::{
     ai_google::parse_data_url,
     ai_providers::{AIPlatform, AIProvider},
@@ -12,7 +13,6 @@ use crate::{
 };
 use async_trait::async_trait;
 use http::Method;
-use super::{anthropic_model_rejects_sampling_params, REASONING_OFF_SENTINEL};
 use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
 use windmill_common::{client::AuthedClient, error::Error};
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use http::Method;
-use super::REASONING_OFF_SENTINEL;
+use super::{anthropic_model_rejects_sampling_params, REASONING_OFF_SENTINEL};
 use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
 use windmill_common::{client::AuthedClient, error::Error};
@@ -165,10 +165,10 @@ pub struct AnthropicOutputConfig {
 }
 
 /// Resolve the thinking config, effort and sampling params for a reasoning
-/// selection. Adaptive thinking rejects sampling params, so temperature only
-/// survives when thinking is off or explicitly disabled (Anthropic returns a
-/// hard 400 otherwise).
+/// selection. Temperature is dropped both under adaptive thinking, which
+/// rejects it, and on the models that removed the sampling params outright.
 fn anthropic_thinking_config(
+    model: &str,
     reasoning_effort: Option<&str>,
     temperature: Option<f32>,
 ) -> (
@@ -176,6 +176,9 @@ fn anthropic_thinking_config(
     Option<AnthropicOutputConfig>,
     Option<f32>,
 ) {
+    let temperature = (!anthropic_model_rejects_sampling_params(model))
+        .then_some(temperature)
+        .flatten();
     match reasoning_effort {
         // The disable sentinel is not an effort token — Anthropic's vocabulary
         // is low..max and rejects it. The disable carries no effort either:
@@ -688,7 +691,7 @@ impl AnthropicQueryBuilder {
         }
 
         let (thinking, output_config, temperature) =
-            anthropic_thinking_config(args.reasoning_effort, args.temperature);
+            anthropic_thinking_config(args.model, args.reasoning_effort, args.temperature);
 
         // Build request based on platform
         if self.is_vertex() {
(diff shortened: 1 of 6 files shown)


hugocasa:
Nits from Claude and Pi all fixed in 6aec9ab36a:
- rewrote the three stale "adaptive thinking rejects sampling params" comments at
  the Converse call sites, and the stale one in the off-sentinel test
- split the sampling assertions out into
  `bedrock_temperature_drops_sampling_params_per_thinking_mode_and_model`
- `mod.rs` now says exactly what was probed (`claude-opus-5`, `claude-sonnet-5`,
  `claude-opus-4-8` reject; `claude-sonnet-4-6` accepts) and attributes Opus 4.7 /
  Fable / Mythos to Anthropic's migration guide rather than to a probe
- `cargo fmt` — no added line exceeds `max_width = 100` (the remaining long lines
  in `bedrock.rs` predate this branch)

**On the P1 (opaque Bedrock application inference profiles):** the observation is
correct and I've documented it in the matcher, but I don't think it should block.

It is not introduced here. An application profile backed by Opus 4.8 already 400s
on merged main for exactly the same reason — the `None` arm passes `temperature`
through — so this PR strictly reduces the failing surface rather than leaving it
unchanged. Nothing that works today regresses.

It is also the codebase's existing, reviewed stance for these ids:
`bedrock_model_supports_prompt_caching` cannot classify them either, and its test
pins an application-inference-profile ARN returning `false`. Both features degrade
on opaque ids because the id carries no model.

Resolving the backing model properly means a `GetInferenceProfile` call per
request (network + latency on the hot path) or threading the ARN the frontend
already fetches — `fetchAvailableModels` reads `inferenceProfileSummaries[].models[].modelArn`
— through the saved provider config into the backend. That is a real feature, not
a fix, and it would also fix prompt caching on the same ids. Worth its own issue;
I'd rather not bolt a per-request AWS lookup onto this.

The limitation is now stated at the matcher so the next reader hits it before the
400 does.

hugocasa:
✅ Review round clean @ 6e72ee9f5baa1fd5392b746dd40b406f7fef0f21 — nit-only verdicts at 6aec9ab36ae04d53c36dd9fad4524e13eda50e57 (codex, claude: mergeable-with-nits; pi: good to merge); nits addressed in 6e72ee9f5b

Both were real misses on my part:
- `anthropic.rs` still carried the refuted "only rejected alongside adaptive thinking" comment — and worse, the test under it asserted an `opus-5` body **with** `temperature`, which is exactly the shape the live API rejects. It now builds the body through `anthropic_thinking_config` and asserts temperature is absent, so it pins the real contract instead of a shape that 400s.
- dropped the three copies of the same rationale at the Converse call sites; it belongs once, on `bedrock_temperature` (AGENTS.md: state each invariant once, at the place where someone would break it).

hugocasa:
Extended the live verification to the other two providers I have keys for. One
more broken default found and fixed in 22aff6bce0.

**Google AI — `gemini-3.1-pro` 404s.** It is not a servable id; only
`gemini-3.1-pro-preview` resolves. It predates #10690 (I kept it and moved it up
the list rather than introducing it), so this has been a dead recommendation for
a while. Every other Google AI default resolves.

**Verified live, all of it previously assumed:**

| Check | Result |
|---|---|
| OpenAI: all 9 defaults exist | ✅ `gpt-5.6-sol/terra/luna`, `gpt-5`, `gpt-5-mini`, `gpt-4o`, `gpt-4o-mini`, `o4-mini`, `o3` |
| gpt-5.6 ladder `low/medium/high/xhigh/max` | ✅ all accepted — the `xhigh`/`max` this PR series added are real |
| gpt-5.6 `none` (the off token) | ✅ accepted — `canDisable: true` is right |
| gpt-5.6 `minimal` | ✅ **rejected** — and the registry correctly does not offer it |
| Google AI defaults exist | ✅ except `gemini-3.1-pro` (fixed here) |
| Gemini 3.x context / output | ✅ `in=1048576 out=65536` — matches the 1M window and 64K output rows |
| Gemini `minimal` on flash | ✅ accepted |
| Gemini `minimal` on pro | ✅ **rejected** (`Thinking level MINIMAL is not supported for this model`) — matches `geminiReasoningLevels` |

So the OpenAI and Gemini ladders, windows and output caps are now measured rather
than inferred, and both matched.

**Still unverified — no keys:** Mistral, DeepSeek, Groq, OpenRouter, Together AI,
Azure OpenAI, Azure AI Foundry, AWS Bedrock. Changes there were deliberately
conservative (Groq's defaults are Groq's own documented migration target; the
Mistral ladder was reverted to its original `['high']`; DeepSeek's `deepseek-chat`
guard was restored; Together AI untouched), but they rest on documentation.

Note `npm run check:fast` reports 3 pre-existing errors on this branch's base
(`TriggerHistoryButton.svelte`, `useAuditLogsLoader.svelte.ts`) — main added
trigger-history endpoints and this worktree's generated client is stale. Confirmed
identical with my changes stashed.

hugocasa:
Checked the remaining unverified surface. Two more corrections, both found by
reading AWS's own model cards rather than by inference.

**1. Bedrock Sonnet 5 cannot disable thinking** (940a80d825) — and this one
contradicts what #10690 shipped. AWS's card is explicit:

> Claude Sonnet 5 — **Reasoning:** Supported (adaptive thinking is **always on and
> cannot be disabled**; effort level is configurable)

versus Opus 5 on the same platform:

> Claude Opus 5 — **Reasoning:** Supported (adaptive thinking is on by default;
> **can be disabled** but effort is capped at high when thinking is disabled)

So the two Claude 5 models differ *on Bedrock*, and Sonnet 5 differs between
Bedrock and the native API — where I live-verified that `thinking: {type:
"disabled"}` is accepted and returns no thinking block. `canDisableReasoning` and
`explicitOffToken` now split `anthropic` from `aws_bedrock`: Bedrock offers the
off switch for Opus 5 only, the native provider for both.

**2. Prompt caching in #10690 is confirmed, not inferred.** Both cards state
`Prompt caching supported: Yes` (4,096 min tokens, 5m + 1h TTL, `system`/
`messages`/`tools`) for Opus 4.8 and Sonnet 5, and both list `Converse` under
APIs supported. The merged allowlist and the Converse transport are correct.

The cards also confirm the geo prefixes this series normalizes — `us.`, `eu.`,
`au.` for both, plus a `jp.` on Opus 4.8 that `normalize_bedrock_model_id` does
**not** strip. That is pre-existing (the list predates this series) and only costs
prompt caching on JP inference profiles, so I have left it; worth a follow-up
alongside the opaque application-profile gap.

[Opus 4.8 card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-8.html) ·
[Sonnet 5 card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-5.html)

PR #10714, opened 2026-08-16
perf: cap resource content sent to the search modal

rubenfiszel:
## Summary

Fixes WIN-2387.

The global content search (`Ctrl+K` → `#`) froze the browser on workspaces with many large
resources. `GET /w/{workspace}/resources/list_search` returned **every** readable resource's
full JSONB value — a row `LIMIT 1000` was the only bound, none on size — and the client then
re-serialized all of them with `YAML.stringify` and kept both copies in memory for the life of
the modal.

Measured on a workspace of 1000 resources totalling 59 MB of values (localhost, so the transfer
number is a floor, not a ceiling):

| | before | after |
|---|---|---|
| response payload | 61.1 MB | 4.0 MB |
| transfer | 1253 ms | 217 ms |
| client-side serialization | 986 ms blocking `YAML.stringify` | none |
| strings retained for the search haystack | 63.8 MB | 3.8 MB |

The endpoint now returns the value already rendered as pretty-printed JSON, capped at 4000
characters per row, plus a `truncated` flag. The search preview only ever shows a handful of
lines, so nothing visible is lost; the frontend searches that text directly instead of
rebuilding it.

## Changes

- `list_search_resources` selects `left(jsonb_pretty(value), 4000)` with a `truncated` flag
  instead of the raw value. The `LATERAL … OFFSET 0` fences the subquery so the planner cannot
  inline `jsonb_pretty` into both the `left()` and the `length()` and serialize each value twice.
- `ContentSearchInner` takes the API text as the search haystack — no `YAML.stringify` for
  resources (flows and apps are unchanged).
- Truncation is surfaced rather than silent: a "Truncated" badge with a tooltip on each affected
  result, and a line under the corpus counts saying how many resources are only partly searched
  (a resource whose only match sits past the cap drops out of the results with nothing else to
  show for it).
- Regression test on the existing `list_search` integration test: an oversized fixture resource
  comes back at exactly the cap with `truncated: true`, a small one at `truncated: false`, and a
  `null` value as `""` rather than `null`.

### Deliberate response-shape change

`listSearchResource`'s `value` goes from arbitrary JSON to a capped string, and gains a required
`truncated`. This is a breaking change to a documented endpoint, so calling it out explicitly:
it is a UI-support route ("list resources for search") whose only consumer in this repo is the
search modal, and nothing in the CLI or the Python/TS SDKs calls it. `get_value` remains the way
to read a resource's actual value. I chose a clean break over an opt-in legacy mode because the
legacy mode *is* the defect being fixed. **Left in draft for a human call on that**, per the
"changes openapi.yaml" rule in the `pr` skill.

## Screenshots

Content search with large resources — capped JSON preview, per-result "Truncated" badge, and the
partial-search notice:

content-search-truncated

Tooltip on the badge:

tooltip

Same workspace with no oversized resources — no badges, no notice:

content-search-no-truncation

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test resources test_resource_endpoints`
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`
- [x] `npm run check` (0 errors; the 70 warnings are pre-existing and none in the changed file)
- [x] Driven in the browser against a seeded 1000-resource / 59 MB workspace: modal opens without
      a freeze, fuzzy matching still works on JSON text (including YAML-flavoured queries like
      `host: db-42`), badge + tooltip + notice render, and both the truncated and non-truncated
      corpora were exercised.

---
## Summary by cubic
Caps resource content returned by the global content search to a 4000‑char, pretty‑printed JSON string with a required `truncated` flag. Previously the endpoint sent full JSONB per resource and the client re‑serialized it, causing multi‑MB payloads and freezes; now the client searches the capped text directly.

- Backend: `list_search_resources` selects `left(jsonb_pretty(value), 4000)` and `truncated`, with `LATERAL (…) OFFSET 0` to avoid double serialization. `COALESCE` returns `""` for nulls. OpenAPI changes `value` to a string, requires `truncated`, and documents the 4000‑char cap.
- Frontend: `ContentSearchInner` removes `YAML.stringify` for resources, searches the provided text, shows a per‑result “Truncated” badge with tooltip, and pluralizes a notice counting partially searched resources. Flows/apps unchanged.
- Tests: Adds an oversized resource fixture and asserts the cap, `truncated`, and `""` for nulls.
- WIN‑2387: Meets the requirement to keep the search modal responsive on large workspaces.

**Migration**
- Consumers of `GET /w/{workspace}/resources/list_search` must treat `value` as a capped string and read `truncated`. Do not rely on the prior raw JSON shape; use `get_value` to fetch full resource content.

<sup>Written for commit e4a88aa7e471c90af52027b6d7859efb026fb976. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Review round on \`afd73abe23\`: **codex** Good to merge, **pi** Good to merge, **claude** nit-only (hard-coded plural in the truncation notice; the \`4000\` literal duplicated between \`openapi.yaml\` and the Rust const). Both nits fixed in \`e4a88aa7e4\` — the notice now reads "1 of those resources is … only its beginning" in the singular (verified in the browser), and the const carries a pointer to the openapi description so the two move together.

Deliberately **not** flipping to ready: this changes a documented endpoint's response shape in \`openapi.yaml\`, which wants a human call (see the "Deliberate response-shape change" section in the description). Everything else is verified — integration test, offline \`cargo check\`, \`npm run check\`, and the UI driven against a seeded 1000-resource / 59 MB workspace.

PR #10720, opened 2026-08-17
feat(secrets): Apple Keychain secret backend

principalwater:
## What

A `SecretBackend` implementation backed by an Apple Keychain. Each secret becomes a generic password item, so for those variables the database holds no ciphertext at all — a stolen database dump does not contain them.

Off by default behind a new `keychain` cargo feature.

## Why two transports

Windmill is frequently not the process that can talk to a keychain. In the common self-hosted setup on a Mac, Windmill runs in a Linux container: it can neither load `Security.framework` nor reach the host's keychain. A macOS-only native implementation would therefore be unusable in exactly the deployment where people have a keychain.

So the backend has two transports:

* **`native`** — calls the Security framework directly. For a macOS-native Windmill process.
* **`helper`** — runs an operator-provided command. The container reaches the host through it.

The helper contract is deliberately tiny: two appended arguments and one stream.

```text
<helper...> get <service>       -> value on stdout, exit 0
                                -> exit 44 when the item is absent
<helper...> set <service>       <- value on stdin, exit 0
<helper...> delete <service>    -> exit 0, also 0 when already absent
```

Exit 44 is the only "not found" signal; every other non-zero exit is an error and the helper's stderr is surfaced verbatim. That distinction is the point — a locked keychain must not look like an empty one, or callers accept an empty string as a valid secret and fail much later somewhere unrelated.

## Configuration

```json
{
  "type": "AppleKeychain",
  "service_prefix": "windmill",
  "transport": { "mode": "native" }
}
```

```json
{
  "type": "AppleKeychain",
  "service_prefix": "windmill",
  "transport": {
    "mode": "helper",
    "command": ["/usr/local/bin/windmill-keychain"]
  }
}
```

Items are named `<service_prefix>/<workspace_id>/<path>`, so several workspaces coexist in one keychain and an operator inspecting it by hand can tell which workspace an item belongs to.

## Deliberate choices worth reviewing

* **No fallback to the database**, in either transport. The Enterprise arms in `vault_oss.rs` warn and fall back; doing that here would write plaintext into the very place this backend exists to avoid. Their behaviour is untouched.
* **The OSS resolver now reads the configured backend** instead of returning `DatabaseBackend` unconditionally. It accepts only backends compiled into the build; the Enterprise variants report that plainly. This is the one change that touches a product boundary rather than just adding code, and I would rather have it discussed than merged quietly — if you would prefer the selection to stay Enterprise-gated and this backend to be reachable another way, say so and I will rework it.
* `security-framework` is declared only under `cfg(target_os = "macos")` and only with the feature on, so the helper transport builds everywhere without it. The crate is already in `Cargo.lock` transitively.
* Blocking Security framework calls run in `spawn_blocking`: they are short, but a locked keychain can block for as long as a dialog is up, and that must not stall a runtime worker.
* `errSecItemNotFound` is matched by its numeric code rather than by message text, which is localised.

## Not in this PR

* Docs. Happy to add them wherever you would like them to live.
* Migration of existing database-stored secrets into a keychain. The existing migration machinery is Enterprise-side; tell me whether that path should be reused.
* macOS CI. There is no macOS runner in `.github/workflows` today, so the native transport is compile-tested locally (arm64, macOS 26.4) and the helper transport is covered by tests that run anywhere.

## Testing

`cargo check -p windmill-common --no-default-features` and the same with `--features keychain` both pass, on macOS/arm64.

Six tests exercise the helper contract against a real `/bin/sh` helper rather than a mock of our own parsing: newline handling, a deliberate trailing blank line, absent item versus empty value, error passthrough, the argv/stdin split (the value must never appear in argv), and that a `get` does not hang when the helper reads stdin to EOF.

```
test secret_backend::keychain::tests::get_trims_one_trailing_newline ... ok
test secret_backend::keychain::tests::get_keeps_a_deliberate_blank_line ... ok
test secret_backend::keychain::tests::absent_item_is_not_found_rather_than_empty ... ok
test secret_backend::keychain::tests::other_failures_surface_stderr ... ok
test secret_backend::keychain::tests::set_passes_the_value_on_stdin_and_the_service_in_argv ... ok
test secret_backend::keychain::tests::get_does_not_hang_when_the_helper_reads_stdin ... ok
test result: ok. 6 passed; 0 failed
```

I also could not find a CONTRIBUTING or CLA file in the repository — if contributions need a CLA signed, point me at it and I will.

---
## Summary by cubic
Stores secrets in an Apple Keychain via a new backend and routes reads by the stored marker instead of the configured backend. OSS now honors `SECRET_BACKEND_SETTING` and fails closed: unsupported or unbuilt backends are errors rather than silent fallbacks.

- Implements `SecretBackendConfig::AppleKeychain` with `KeychainSettings { service_prefix, account?, transport: Native | Helper { command } }` and the `$keychain:<path>` marker.
- Builds behind the root-crate `keychain` feature; `security-framework` is optional and only linked on macOS; helper transport builds everywhere.
- Resolver:
  - Loads `SECRET_BACKEND_SETTING` in OSS; accepts Database and, when `keychain` is built, Apple Keychain. EE backends error; Apple Keychain errors if configured without the feature.
  - Reads by marker (`$vault:`, `$azure_kv:`, `$aws_sm:`, `$keychain:`); a marker/config mismatch errors and names both.
- Variable lifecycle:
  - Store writes through the backend and persists `$keychain:` when selected.
  - Delete and rename handle any external marker; rename now refuses to proceed if the marker does not belong to the configured backend.
  - Shared helpers `external_marker_prefix` and `marker_for_backend` remove duplicated prefix logic.
- Helper transport hardening:
  - Contract: `<helper> get|set|delete <service>`; `set` via stdin, `get` to stdout; exit 44 means “not found” for `get` only.
  - Concurrent stdin/stdout, overall timeout with kill-on-drop, stdout bounded at source; `get` returns stdout byte-for-byte, `set`/`delete` ignore stdout; stderr is discarded.
- No fallback to the database for Apple Keychain.

**Operator enablement**
- Build with `--features keychain`; without it, configuring Apple Keychain errors.
- Choose transport: Native on macOS with an unlocked keychain, or Helper with a command implementing the contract above.
- Set `SECRET_BACKEND_SETTING` to Apple Keychain. No migration of existing DB-stored secrets is included.

<sup>Written for commit c1785f4ed9e585d98e86b94ebff455a93e0b0796. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Thanks for the PR. At the moment we're not taking PRs anymore given AI made it a lot easier to generate code, we prefer discussing the needs and use-cases and then we decide or not to create the PR. We haven't made it official yet, apologies about it if you spend lots of time on this. That being said, I'm genuinely curious about the use-case. Do you run Windmill directly on your laptop ?

principalwater:
> Thanks for the PR. At the moment we're not taking PRs anymore given AI made it a lot easier to generate code, we prefer discussing the needs and use-cases and then we decide or not to create the PR. We haven't made it official yet, apologies about it if you spend lots of time on this. That being said, I'm genuinely curious about the use-case. Do you run Windmill directly on your laptop ?

Thanks for the straight answer - no apology needed, and I'd rather hear the policy than have the PR sit there.

To your question: not a laptop. It's a Mac Studio that runs continuously as a small home server, and Windmill runs on it in containers under OrbStack. It's the orchestrator for everything on that machine - a DuckLake data lake, health and activity collectors, a handful of small VPS nodes - so it holds most of the secrets in the house: API tokens, an SSH key for the fleet, database passwords.
The use case is specifically the single-machine self-hosted shape. With the database backend, the workspace encryption key and the ciphertext live in the same Postgres, on the same disk. That's fine against a lot of things, but it means any copy of that database leaving the machine carries every secret with it - and a backup leaving the machine is exactly what you want to be doing. Meanwhile macOS already ships a secret store that the rest of the host tooling on that machine uses anyway, so Windmill was the one component keeping its secrets somewhere else.

That's the whole motivation. Not "keychain is nicer" - it's that on a single-machine deployment the database backend and the backup of that database are the same artifact.

One design note, since it's the part that isn't obvious: the reason there are two transports is that Windmill on a Mac almost always runs in a Linux container, where there is no Security framework and no path to the host keychain. A macOS-only native implementation would therefore be unusable in precisely the deployment where a keychain exists. So the native path is there for a macOS-native process, and a `helper` transport runs a command the operator provides - two appended arguments and one stream, nothing more. In my setup that command is an SSH call to a forced-command script on the host, restricted to the `windmill/` namespace, and it's the transport actually in use.

I should also say plainly: after an independent review of my own branch, the PR as it stands is not mergeable regardless of your policy. The backend is not yet threaded through the full variable lifecycle - `store_secret_value` doesn't know the marker, and delete, rename and workspace fork still assume the Vault-shaped checks - so it only works for reads I wire up by hand. I've fixed the contained defects locally (a real deadlock in the helper pipe handling, exit 44 being accepted as a successful write, a trailing newline being trimmed off secrets, and a non-exhaustive match that broke the EE build), but the lifecycle work is the substantial part and I haven't done it.

So I'm not asking you to merge anything. If the use case above sounds worth having in Windmill, I'm happy to move this to a discussion or an issue and let you decide the shape, including "the helper transport alone, no platform-specific code", which honestly covers most of the value and would work on any OS. If it doesn't fit your roadmap, that's an entirely fine answer too; I'll keep running it on a fork and you can close this.

Either way, thanks for building the thing - it's been running my house for months!)

principalwater:
@cubic-dev-ai review this

principalwater:
@rubenfiszel, following up on my last comment, I said the branch wasn't mergeable and that the lifecycle work was missing; I've now done it and run the thing for real, and three of the four bugs I found only showed up once it was actually serving a live instance. Pushing them here because two of them are not specific to this backend — they affect Vault, Azure Key Vault and AWS Secrets Manager today.

**The one worth your attention regardless of what you decide about this PR imho:**

`get_secret_value` dispatches on the *configured* backend rather than on how the value was stored. So the moment an external backend is switched on, every pre-existing secret becomes unreadable: those values are still ciphertext in `variable.value`, but the read goes to the external store and comes back not-found. I hit this on my own instance — 23 secret variables started returning 404 and the jobs that read them stopped, until I set the backend back to `database`.

It also makes migration impossible in principle, since migrating a secret starts by reading it. I've changed it to route on the marker in the stored value, with a mismatched marker becoming its own error naming both sides instead of a silent lookup in the wrong store. If you'd rather not take the rest of this PR, this part still seems worth fixing on its own.

Two smaller ones in the same family:

- `is_vault_backend_configured` had an OSS copy returning `false` outright. It gates `delete_secret_from_backend`, so deleting a secret variable removed the row and left the secret behind in the backend. Correct while OSS had no external backend; not correct once one exists.
- the rename branch was gated on `is_vault_stored_value`, which only matches `$vault:`. With any other backend the block was skipped whole: the row moved, the secret stayed under the old name, the new path read nothing.

I also removed the marker if-chain that had been written out three times (two in `windmill-store`, one in `windmill-api`) in favour of one helper, because adding a backend previously meant remembering all three and forgetting one rewrites a value with the wrong prefix.

On the backend itself: it now goes through the whole lifecycle. Verified on a live instance rather than in tests: create puts the value in the keychain and leaves only `$keychain:<path>` in the database; read returns it byte-for-byte including a trailing newline; rename moves the keychain item and empties the old name; delete removes both the row and the item. Then I migrated all 23 of my own secrets, each verified individually before moving to the next, and the scheduled jobs kept running.

Your bot's review was useful and I've addressed all of it - including, embarrassingly, comments I'd left in Russian in `Cargo.toml` and `vault_oss.rs`. One correction to my PR description: I wrote that I couldn't find a CLA (that was botty one). There is a CLA assistant configured; the check failed on this PR for an unrelated reason (GitHub returned 429 while fetching the action). So CLA was successfully signed within previous PR by me.

**And to be clear about what I'm asking.** Nothing, still. You've said you'd rather discuss use cases first, and that's fine - I'm not trying to route around it by improving the code. If the use case is interesting, I'd happily open a discussion and let you decide the shape; the version I'd argue for is the `helper` transport alone, without any platform-specific code, which works on any OS and covers most of the value. If it doesn't fit, close this and I'll keep running the fork - but do take a look at the read-routing one either way, since it bites anyone switching a backend on an instance that already has secrets.

principalwater:
@cubic-dev-ai review this

PR #10722, opened 2026-08-17
feat: audit the icon library against brand guidelines

Guilhem-lm:
## Summary

Every icon component in `frontend/src/lib/components/icons` audited against its brand's own published guidelines, for correct artwork, current colours, and readability on both app surfaces (light `#fbfbfd`, dark `#2e3441`).

The library had drifted: marks in colours their owner retired years ago, one drawn as an app tile the brand doesn't use, one that wasn't the brand's mark at all, and a cross-component CSS bug that made GitLab and Funkwhale change colour depending on which other icons happened to be mounted.

The branch also carries the resource-type description work it was named after: descriptions now reach the drawer, the form and the table, and searching a resource type matches and ranks on them.

### 1 — Audit against brand guidelines

- **127 per-theme pairs** applied as `text-[#light] dark:text-[#dark]`, only where the brand publishes a reversed or dark variant. `twMerge` where the component exposes a `class` prop, so callers passing `w-6 h-6` don't wipe the colour.
- **303 of 304 brand icons record their source** in a comment above the `<svg>`, including the rule where one exists — Google forbids recolouring, Cal.com is deliberately greyscale, Oracle reserves the MySQL dolphin for licensees.
- **`BRAND_COLORS.md`** — generated from the components (colours from the fills, sources from the comments, contrast computed), so the table can't drift from the code.
- **Readability**: marks effectively invisible on a surface went **13 → 1** on dark and **9 → 4** on light. Every remaining one is blocked by trademark terms, not unfixed.
- **Artwork replaced** where a first-party or CC0 source existed: PayPal is now the real three-colour monogram (a flat fill was losing the overlap that defines it), Stripe is the bare S rather than an S in a rounded-square container, and `gcloud` resolves to Google's actual mark instead of a generic hexagon.
- **Concept icons** (`CACertificate`, `DbIcon`, `WebdavIcon`, `Asset*`, `BcryptIcon`) inherit `currentColor` instead of hardcoding a colour.
- **`/kitchen_sink/icons`** — a gallery rendering every icon on both surfaces at once, which is how the rest of this was verified.

#### The CSS bug

Ten icons embedded a `<style>` inside their `<svg>`. Svelte only scopes a component's **top-level** style block, so these were emitted as document-global CSS under names like `.st0` and `.cls-2` — which four icons each defined with different colours. The worst offender was our own `WindmillIcon`, rendered from the root logged-in layout, putting `.st0 { fill:#ffffff }` on every page of the app. Class names are now namespaced per icon.

### 2 — Icons for resource types that had none

19 hub resource types fell back to a generic `Boxes` glyph. **11 now have a real vector mark** from a first-party source (`beamer`, `campayn`, `codat`, `comapeo_server`, `klaviyo`, `matteroom`, `mollie`, `motimate`, `paychex`, `terra`, `vectara`).

The other 8 publish their mark only as raster, or publish a wordmark rather than a mark. Rather than upscale a PNG or invent an SVG, they use **`BrandLetterIcon`** — the brand's initial in the brand's colour, on a rounded square in dark mode. The same fallback covers the marks whose owner reserves them: `AdobeAcrobatSign`, `Mysql`, `LeonardoAi`, `WeatherApi`, `WebScrapingAi`, `LocalContexts`, `GlobalForestWatch`, `Actimo`, `AdRapid`, `AeroWorkflow`.

Callers pass literal classes (`bgClass="fill-transparent dark:fill-[#584CCC]"`), because Tailwind's scanner cannot see a class name built from a prop.

### 3 — Provenance closeout

Settles the 54 records whose author rated itself below "high" and which no verifier ever reached, plus the 8 icons with no source at all. **46 confirmed as already correct**, 11 citations corrected, 3 colours corrected — and two changes refuted and reverted by the adversarial pass:

- **`Mysql`** — the comment had the colour-to-shape mapping inverted. Rasterising the first-party asset shows `#00758F` paints the dolphin and "My" while `#F29111` paints "SQL", not the reverse.
- **`AdobeAcrobatSignIcon`** — a "corrected" citation rejected on evidence. It claimed the canonical URL 404s; three fetches returned HTTP 200 with a real Adobe SVG whose stylesheet is `.a{fill:#584ccc}`.

### 4 — Desaturated trigger marks, as separate components

Once brand marks became coloured, a trigger picker mixing lucide glyphs (`Webhook`, `Route`, `Database`) with brand marks (Kafka, GCP, AWS) read as two different sets of controls.

**This is done with 10 dedicated components in `icons/triggers/`, not a CSS override.** An earlier attempt used an `.icon-mono` utility that redirected descendant fills to `currentColor`; it is gone, because coercion cannot work in general:

- lucide icons are outlines (`fill="none"` + `stroke`), so a blanket `fill: currentColor` fills them in solid;
- marks that set their own `text-[#hex]` class ignore a fill rule entirely — MQTT and NATS stayed purple through it;
- gradient-based marks can't express a monochrome variant at all without being redrawn.

Nine variants are the pre-audit artwork painted with `currentColor`. `GoogleCloudIcon` is the exception and is **greyscale, not monochrome** — four grey tones, because the GCP mark loses its shape as one flat colour.

Where each is used:

| Surface | Marks |
|---|---|
| Sidebar trigger list, capture dropdown, add-trigger dropdown, flow-graph badge **menu** | `icons/triggers/` variants |
| Per-kind trigger pages, flow-graph badge itself, tables, global search, resource picker, chat tool cards | full colour |

`triggers/utils.ts` exposes both maps: `triggerIconMap` and `triggerIconMapMono` (a spread of the first with the 10 overrides), so a surface picks by importing one or the other and a missing variant falls back to colour rather than breaking. The rule is written down in `icons/index.ts`.

### 5 — Resource-type descriptions in the UI

- **Drawer titles** name the type: "Add a Supabase resource", with its icon beside the title (`AppConnectDrawer`, `ResourceEditorDrawer`, `LightweightResourcePicker`, via `addResourceTitle`).
- **`ResourceForm`** shows the type's description at the top as rendered markdown, and explains what the path controls (a `u/` path is private, an `f/` path follows the folder's permissions).
- **Resources table** — the description column is fixed at two clamped lines with the full text on hover, so a long description can't make one row three times the height of its neighbours; the *Shared globally* badge sits flush right with no cell separator.
- `GfmMarkdown` gained a `prose` prop over the shared `markdownProse` stack; `DropdownV2` gained `menuClass`, because it portals its menu and a wrapper can't style it.

### 6 — Search resource types by description, ranked

Resource types are named by abbreviation (`gdrive`, `gcal`, `s3`), so the product name only appears in the description — searching "google" used to find none of the Google integrations. Search now covers name + description, which makes ranking necessary: 14 types match "google", most of which only mention it in passing.

`resourceTypeMatchRank` sorts name matches above description matches, ties broken on where the match starts:

```
"google"   before: anthropic, bigquery, firebase, gcal, gcloud, gcloud_storage, …
           after : googleai, bigquery, firebase, gcal, gcloud, gcp_service_account, …
"postgres" before: neondb, postgresql        after: postgresql, neondb
"slack"    before: signature_auth, slack     after: slack, signature_auth
```

Applied to all three resource-type searches: the Resource Types tab (whose bare term searched the name only until now), the add-resource drawer, and the schema-narrowing picker. Guarded by `resourceTypeMatchRank.test.ts`.

Harmonising the four unrelated search mechanisms behind these surfaces is tracked separately in WIN-2400, not in this PR.

## Screenshots

`/kitchen_sink/icons` — each tile shows the icon on the light surface and the dark surface side by side. Ansible inverts, Anthropic and AssemblyAI flip black→white, concept icons stay grey.

icons-gallery-light

Dark page toggle. The note in the count line is a real limitation, not a bug: with the page dark, `dark:` classes also match inside the light half, so Anthropic's light tile goes white-on-white. Read the pairs with it off.

icons-gallery-dark

## Test plan

- [ ] `/kitchen_sink/icons` — pairs differ between halves; the gallery includes `triggers/`, so each variant sits next to the colour mark it shadows
- [ ] Toggle the app theme and confirm marks stay legible in the resource picker and AppConnect drawer
- [ ] Render GitLab, Funkwhale, Appwrite, YNAB and SurrealDB **on the same page** — before this they fought over `.st0` / `.cls-2` and took whichever won the cascade
- [ ] Trigger surfaces split correctly: sidebar list, capture dropdown, add-trigger dropdown and the flow-graph badge *menu* are desaturated; the trigger pages, the badge itself, tables and global search are in colour
- [ ] The runs/triggers row strip (`RowIcon`): Kafka, NATS, MQTT, AMQP and SQS now render in brand colour beside GCP and Azure, instead of five of the seven being greyed
- [ ] Search "google" in the Resource Types tab, the add-resource drawer and a resource-type arg picker — `googleai` first, `anthropic` well down the list
- [ ] A resource with a long description doesn't make its table row taller than the others; hovering shows the full text
- [ ] `npm run check` passes

## Known gaps

Being explicit rather than implying this is finished:

- **The resource-type description, table and search changes were not exercised in a browser** from this session — no backend was running for this worktree. Verified by type-check, unit test and by ranking the real 272-type hub corpus offline. Only the icon work is screenshotted.
- **Smartsheet** artwork has no CC0 or first-party source and keeps its current drawing.
- **28 icons carry an inert `fill="currentColor"`** — harmless, but they don't respond to a `color` prop the way the rest do.


Guilhem-lm:
@codex re the bundle P1 — deferring it deliberately, tracked as [WIN-2402](https://linear.app/windmilldev/issue/WIN-2402/load-the-resource-type-icon-barrel-lazily).

The finding is right about the cost and I measured it: the 89 added components are ~55 KB gzip of source, and the eager consumers are the hot resource surfaces (`IconedResourceType`, `ResourcePicker`, `home/ListFilters(Quick)`, `flows/pickers/PickHubScript(Quick)`, `AuthSettings`).

Two things put the fix outside this PR:

- **The barrel was already eager.** 231 of the 320 entries were static imports before this branch, so making it lazy fixes a pre-existing architecture problem, not a regression this PR introduces. Doing it here would fix 320 while the PR is nominally about 89.
- **It is not a mechanical rewrite.** Every call site renders synchronously (`{@const SvelteComponent = APP_TO_ICON_COMPONENT[x]}`), so going lazy needs an async resolver plus a placeholder frame in dense lists — a visible change that wants its own review, not an addendum to a 390-file diff.

The 89 are not new artwork either: they are components that already shipped in the repo but were never mapped, so wiring them in is what stops those resource types rendering the generic `Boxes` glyph. Dropping them back out would give up most of the coverage this PR is for.

`components/mcp/providerIcon.ts` already does the `import.meta.glob` version for the chat's server icons, and its comment is the rationale WIN-2402 will follow.

Guilhem-lm:
@pi-agent re the bundle P1 — same answer I gave @codex above: deferring deliberately, tracked as [WIN-2402](https://linear.app/windmilldev/issue/WIN-2402/load-the-resource-type-icon-barrel-lazily), which cites `providerIcon.ts` as the model exactly as you do.

Agreed on the mechanism and the measurement. The reason it isn't in this PR:

- **231 of the 320 entries were already eager before this branch.** Making the map lazy fixes a pre-existing problem — it would be a 320-entry change riding inside a PR whose icon delta is 89.
- **The 89 aren't new artwork.** They are components that already shipped in the repo but were never mapped, so wiring them in is precisely what stops those resource types rendering the generic `Boxes` glyph. Dropping them back out gives up most of what this PR is for.
- **The fix needs its own review pass.** Values become filename strings, an `AppIcon` resolves them through `import.meta.glob`, and every consumer renders a frame late — visible in dense lists like `ResourcePicker` and `ListFilters`, so it wants browser verification rather than being appended to a 390-file diff.

Your round also caught two things that are fixed rather than deferred: the `AwsIcon` / `GoogleIcon` sizing regression @codex flagged is in `d9c06f9f55` (both take `size` again, keep `width`/`height`, and accept a `class` so `RowIcon`'s grey still applies) — I swept every audited icon for the same prop-contract break and those two were the only ones.

PR #10726, opened 2026-08-17
feat: add empty state cards to list pages

Guilhem-lm:
## Summary

Empty tables across the workspace showed a bare line of text — "No resources found", "No Kafka triggers" — with no way to act on it. That is the first thing you see in a new workspace. This replaces those with a dashed card carrying the item's icon, a sentence explaining what the thing is for, and a button that opens the same editor the page header opens. The card's button uses the `default` variant: the page header keeps the one accent CTA per view, and a centred neutral button doesn't compete with it.

Also fixes a pre-existing animation bug found while testing: lazily-mounted trigger drawers snapped open the first time instead of sliding.

## Changes

**New shared components**
- `common/emptyState/EmptyState.svelte` — dashed card, 32px icon, title, description, optional action button. Replaces the copy-pasted "nothing found" markup.
- `common/tabs/TabFade.svelte` — keyed crossfade for tab content (~140ms). Both panels share one grid cell so the outgoing one does not push the incoming one down mid-fade; the inner panel keeps `min-w-0` so wide tables still scroll inside their own container.

**Pages using it**
- Resources — per-tab copy for Workspace / States / Cache / Theme, `Add a resource` opens the connect drawer. The Theme tab and its tooltip now say themes are CSS for the legacy low-code app editor and are added from an app's CSS panel, not from this page.
- Variables — workspace tab plus the custom contextual variables section, each with its own CTA (contextual gated on admin, matching the header button).
- Folders — the empty message used to be a row *inside* the table, leaving a header row over nothing; the card now replaces the table. The create-folder popover was extracted into a snippet shared by the header and the card, parameterized by placement and variant so it centres under the empty-state button and renders neutral there.
- 13 trigger pages — schedules, HTTP routes, postgres, websocket, email, kafka, nats, mqtt, amqp, sqs, gcp, azure and native triggers, each with its own icon and CTA.

Pages whose filters are applied server-side (resources, variables, schedules) distinguish "nothing here yet" from "the filters excluded everything" and render a `SearchX` card for the latter, with no CTA. `user_folders_only: false` does not count as an active filter: it narrows nothing when off, and treating it as active would hide the create CTA in an empty workspace. The filtered-empty case on the other trigger pages already had `NoItemFound` and is untouched.

**Drawer animation fix**
The Drawer slides by toggling a CSS class, and a CSS transition needs the closed state painted once to have a start value. The 13 trigger editor wrappers lazily mount their drawer and then call `openNew()` after only `await tick()`, which flushes the DOM but does not wait for a paint — so the panel was created and opened within one frame and snapped. New `utils/paint.ts` exports `tickPainted()`, which waits two animation frames (one is not enough: a rAF callback runs *before* the paint it is scheduled against — measured, single-frame still produced zero transitions). It also absorbs the near-identical `waitForAnimationFrame` that `CreatedResourceActionDrawers` was carrying, keeping that call site's single-frame semantics and its SSR guard.

## Screenshots

resources-empty

variables-empty

Filtered-empty — no CTA, since the fix is to widen the filter rather than create:

variables-filtered

folders-empty

schedules-empty

kafka-trigger-empty

## Test plan

- [x] Empty workspace: resources, variables, contextual variables, folders, schedules, routes, kafka render the card; each CTA opens its editor
- [x] Remaining trigger pages (postgres, email, websocket, sqs, amqp, mqtt, nats, gcp, azure) verified by reading the rendered card text and icon size from the DOM
- [x] Filtered-empty renders the `SearchX` card instead of the CTA, on resources, variables, schedules and the resource-types tab
- [x] Schedules with one deployed schedule + a filter matching nothing shows "No schedules found", not "No schedules yet"
- [x] `?user_folders_only=false` still shows the real empty state
- [x] Populated tables unaffected; tab crossfade measured (old panel to 0 by 90ms, new to 1 by ~220ms) with no layout jump
- [x] Trigger drawers animate on first open — `transitionrun: transform` now fires on the first click, on schedules and kafka
- [ ] `native_triggers/[service_name]` empty state — could not exercise: the page short-circuits to "triggers are not available" without a connected integration
- [ ] Restricted states that offer no create button (workspace with `DisableDirectDeployment` and no bypass, non-admin on email triggers, restricted demo workspace on folders) — verified by reading the gates, not exercised; each keeps its own explanatory affordance and none could create before this PR either


Guilhem-lm:
✅ Review round clean @ 4083a78193d1f4ab3fb8ed32abfa6be9f6f641f0 — nit-only verdicts at ef0deecd9e70fee01df2365c85183b3ca44a982e (codex: good to merge, pi: good to merge, claude: nits); nits addressed in 4083a78193

PR #10727, opened 2026-08-17
docs: make herdr the documented worktree manager

hugocasa:
## Summary

The `AGENTS.md` dev environment section still assumes webmux and tmux throughout, so an agent in a
herdr worktree goes looking for `$WEBMUX_WORKTREE_PATH` and `webmux/runtime.env`, finds neither,
and is left with no documented way to discover its own ports or database. Now that herdr is what
the team runs, this makes it the documented manager and keeps webmux as a short fallback for
whoever has not moved.

This PR was opened in August, when herdr was still being trialled alongside webmux, and documented
the two side by side. Reshaped: herdr leads, the symmetric table is gone, and the shell-script
refactor it also carried is dropped (both of its callers, `scripts/worktree-env` and
`scripts/post-create.sh`, are webmux-only paths — herdr's hook writes its own `.env.local` by
design). Docs only now.

## Changes

- `AGENTS.md`: herdr's markers and pane commands (`$HERDR_ENV`, `herdr pane list` / `pane read`,
  `herdr --skill`), with the webmux/tmux equivalents kept in one short paragraph
- `AGENTS.md`: `.env.local` is documented as where the per-worktree values actually live, read
  with `cat` rather than the file-read tool, which denies every `.env.*` path. `DATABASE_URL` is
  written with an `export` prefix that a `^DATABASE_URL=` grep misses
- `AGENTS.md`: the database-naming example uses the herdr layout, where the directory keeps the
  full branch name and the 63-character truncation cuts mid-word
- `AGENTS.md`: the fallback discovery snippet matched `git branch --show-current | tr - _`, which
  under herdr returns nothing for any branch with a `/` in it or any name past the truncation
  limit. It now matches the worktree directory, truncated the way Postgres truncates it
- `backend/AGENTS.md`: the restart recipe was pure `tmux send-keys`, so none of it worked under
  herdr; each step now gives the herdr command with the webmux one alongside
- `.agents/skills/pr/SKILL.md`: "webmux oneshot" is just "a one-shot run" now

## Test plan

- [x] Every herdr claim checked against the running CLI: `HERDR_ENV=1`, `HERDR_WORKSPACE_ID`,
      `herdr pane list --workspace`, `herdr pane read --source recent-unwrapped --lines`,
      `herdr pane process-info --pane`, `herdr pane send-keys`, `herdr pane run`, `herdr --skill`
- [x] `.env.local` in three live herdr worktrees carries all six documented fields
- [x] The new database-discovery snippet returns the right database in a herdr worktree; the old
      one returns nothing there
- [x] The vite `REMOTE` snippet still resolves in a herdr worktree
- [x] `WM_CLONE_DB` confirmed to come from the `windmill.worktree` plugin's `config.env`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,9 +64,9 @@ they set disjoint environment variables, so check which one you are in:
 
 See `backend/AGENTS.md` to restart the backend with different cargo features. Under herdr, `herdr
 --skill` prints a full reference for inspecting and driving panes and agents, and the plugins that
-provision worktrees live in `windmill-labs/windmill-herdr` — clone it and run `./setup.sh` to
-install them and the keybindings they need. The commands below are for a plain checkout with
-nothing running.
+provision worktrees live in the private `windmill-labs/windmill-herdr` — clone it and run
+`./setup.sh` to install them and the keybindings they need. The commands below are for a plain
+checkout with nothing running.
 
 - **Backend**: `cargo run` from `backend/` (API at http://localhost:8000)
 - **Frontend**: `REMOTE=http://localhost:8000 npm run dev` from `frontend/` (port 3000+)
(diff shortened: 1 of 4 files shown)


hugocasa:
✅ Review round clean @ f551c48c2d742af9319348b475eb25420212c602 — nit-only verdicts at 1fece42e00af4ada5ff434983f17e13780caa31d (codex: good to merge, pi: good to merge, claude: mergeable with nits); the capture-pane nit is addressed in f551c48c2d742af9319348b475eb25420212c602

PR #10729, opened 2026-08-17
feat: a wizard for importing a hub project, and finishing what the import cannot

Guilhem-lm:
Importing a hub project used to be one page that inherited whatever workspace happened to be active. The hub cannot know which workspace you want — it only ever links to *an instance* — so the question has to be asked here. `/projects/import` now asks it, and then finishes the job: the destination, the import, and the setup an import cannot do for you.

## Merge windmill-labs/windmillhub#180 first

This reads one hub endpoint, `GET /projects/<slug>`, which that PR adds. Merging this first is not dangerous — the project card degrades to "Could not read `<slug>` from the hub" and the import still completes, because it goes through the backend's own workspace-scoped proxy — but step 1 looks bare until the hub ships.

**Deploy order is the opposite.** The hub's button points at `/projects/import`, a route that only exists here, so a hub deployed ahead of this puts every click on a 404. Windmill must not lag behind the hub.

## Nothing happens until you ask

The wizard's state is a plain value in the URL — `lib/importWizard/plan.ts`. Steps 1 and 2 only ever *describe* an import, so the back button, the stepper and the Back control are the same operation: a URL change. None can strand a half-created workspace, because there is no state anywhere else to unwind.

`lib/importWizard/execution.svelte.ts` is the only code that acts on a plan, and it reports as an ordered task list:

```
✓ Create workspace calendly-2
✓ Fetch the project from the hub      — 9 items
✓ Import the items                    — 2 apps, 4 scripts, 2 resources
✓ Run 1 data table migration
```

The same list is drawn *before* the run from the plan alone (`plannedTasks`), so step 3 states what it is about to do and then fills in the same rows rather than replacing a paragraph with a list.

**Nothing is stored anywhere.** The plan is in the URL, what already exists is asked of the instance (`lib/importWizard/probe.ts`), and what is in flight is in memory. A reload mid-run re-probes rather than re-creating: `workspace.owner` — set at `INSERT INTO workspace`, already on `listWorkspaces` — says whether an id that exists is one this user made, which is what makes adopting it instead of creating it safe.

## The four steps

1. **Destination** — a new workspace or one you already have. Creating is gated on the same permission the workspace picker uses, and now correctly offers it to a superadmin on an instance with `CREATE_WORKSPACE_REQUIRE_SUPERADMIN` set, which the old check hid from everyone.
2. **Workspace** — the real `WorkspaceTreeView`, so forks nest under their parent with the same search, colours and roles as `/user/workspaces`. For a new one: name and id, auto-suffixed `-2`, `-3` when the prefill is taken, and a username only where `automate_username_creation` is off.
3. **Import** — the folder, the checklist, and the run.
4. **Set up** — *optional, and the point of the PR.* An import cannot supply what a project needs to actually run: a data table is a named database connection the workspace owns, and every resource arrives as an empty stub because the hub never publishes credentials. This step configures both, then finishes. Skippable, with an alert saying what will not work.

## Step 4 in detail

**Data tables.** A project ships migrations that recreate its tables. Where the destination already has a table of that name, they run during step 3 behind a review drawer showing the SQL. Where it does not, they defer to here: the row opens `AddDataTableWizard` — the same one workspace settings uses — which provisions the database, and the migration then runs as the last row of that wizard's own checklist. The SQL is shown before anything runs it.

**Credentials.** Each blank resource is listed with its integration icon, its type's display name, and which required fields are missing. Where the instance has an OAuth client for that type, the button is **Connect** and the flow writes into the existing stub (`AppConnectInner` gained an opt-in `fillPath`); otherwise it opens the resource editor. Rows are marked done rather than removed — a checklist line that vanishes when you complete it reads as something going wrong.

## Elsewhere in the diff

- **A first-time signup no longer loses the project.** `rd` survives an OAuth sign-in, but a first-time account was sent to onboarding and the destination dropped. Onboarding now honours it, through `toSameOriginRelativePath`.
- **Sign-in kept dropping the destination** — `rd` was honoured verbatim only for `/user/workspaces`, so anyone with more than one workspace landed on the picker: the page this replaces, asking the question it was about to ask.
- **`/projects/install` redirects** to the wizard, query intact, since hubs upgrade on their own schedule.
- `FolderPicker` takes an optional `workspace`, so it can list and create in a workspace the app has not switched to. `CenteredModal` takes an optional `subtitleSnippet`. Both additive.
- The wizard renders at `+page@(root).svelte`, outside the app chrome, because `(logged)` waits on a `$userStore` that a not-yet-created workspace cannot provide.

## Known gaps

- **Scopes are per-provider, not per-project.** They come from `backend/oauth_connect.json` and are identical for every project. Calendly's `freeBusy` call needs broader Google Calendar scope than the registry grants, so availability fails while booking works, and the user must widen it by hand. Closing it means the export declaring what each resource needs — a hub change.
- **Retry replays the whole bundle.** `probe.ts` ships the reads that would make it import only what is missing; wiring them up changes what Retry means and is left for a follow-up.
- Step 3's counts come from the hub summary while the export is fetched during the run. Two sources for one number, agreeing today.

## Checks

`svelte-check`: 0 errors, warnings unchanged from `main`. 46 unit tests across the plan, the executor's abandonment and resume behaviour, workspace ids, the probe's SQL and path inference, and migration SQL generation.

Walked end to end against a real backend: both destinations, a new workspace created and deleted through the UI, the setup step configuring a data table and running its migration, a reload mid-run resuming without a second create, and the reload-then-Back routes refusing to re-import.


Guilhem-lm:
/review

Round 2. Addressed from the previous round:
- **P1 username never validated** — `planProblem` + the Continue gate now run `validateUsername` and require non-empty when the instance does not derive it.
- **P1 leaving mid-run did not stop the run** — `ImportExecution.abandon()` stops at the next phase boundary and keeps the workspace parked so the link resumes; the review-drawer teardown now resolves `'abort'` instead of `false`, so an orphan can no longer import items without their migrations.
- **P2 workspace name length** — `WORKSPACE_NAME_MAX_LENGTH` checked in `planProblem`.
- **P2 stale `hubAppIcon` JSDoc** — removed.

Added tests for the parking decision and both validation gates.

Guilhem-lm:
/review

Round 3. All nine from round 2 addressed in `4c6528774a`:

**P1** hub data-table names as raw HTML → escaped (same class as the round-1 SVG finding).
**P1** setup read unretargeted paths → retargets like the import does, and filters to the import folder so a crafted export can't name a path outside it.
**P1** abandonment only checked between phases → `installProject` takes a `stopped` callback, checked before every write loop.
**P1** failed setup migration reported success → rethrown so the wizard's checklist marks it failed.
**P1** completed run lost on Back from setup → the page hands its execution back via `resume`, adopted under the same plan tag.
**P1** blank and `global` workspace ids reached step 3 → both refused at the Continue gate.
**P1** `onboardingDestination` weaker than the shared helper → now uses `toSameOriginRelativePath` (`/\\evil.com` was passing).
**P2** `size="xs2"` → `unifiedSize="2xs"`.
**P2** two doc comments reattached to the functions they describe.

Tests: 53 passing, including `global` id rejection.

Guilhem-lm:
/review

Round 4. Both round-3 findings addressed in `ff8326c3c0` — both were regressions from my round-2 fixes:

**P1 `resume` adopted a stale run.** The seed tagged the run with the plan being rendered, so the guard was true by construction. `ImportExecution.planTag` now carries the plan the run was actually made for.

**P1 abandoning marked the run `done`.** `installProject` returns early on `stopped` the same way it returns on success, so the tail could not tell them apart — a run stopped after 3 of 10 items reported `done — 3 items` with no error, and the resumed step offered Continue instead of Retry. The tail now checks the flag and leaves the run failed and retryable.

`abandon.test.ts` was the hand-written copy you called out; it now drives a real `ImportExecution` with the install seam mocked and abandons from inside the write loop. 60 tests passing.

Guilhem-lm:
/review

Round 5. All of round 4 addressed in `a07a63ae5c`.

**Codex P1 unreviewed setup SQL** — each unconfigured data-table row now shows the SQL before "Set up" runs it. Read-only rather than editable: the risk named was executing unseen SQL against a database that may hold unrelated objects, and this closes that. Say if you want full parity with step 3's editor.
**Codex P1 finish before the setup check settles** — the button reads "Checking…" and is disabled until `listDataTables` returns.
**Codex P1 reload on step 4 → Back re-imports** — Back exists only while the page still holds the run, which is exactly the non-reloaded case.
**Codex P2 stranded `run()` JSDoc** — moved back onto `run()`.
**Pi P2 fork named `global`** — only the effective id is checked now; test added.
**Pi P2 abandoned run left `migrate` on `running`** — reset to pending.

61 tests passing; svelte-check clean. Verified in a browser: step 4 shows the disclosure and expanding it reveals the CREATE TABLE statements.

Guilhem-lm:
/review

Round 6. Both reviewers landed on the same remaining gap, fixed in `ee2a50022a`.

Round 5 gated the setup step's own Back button, which turned out to be the least used of three routes back into step 3. The stepper's "Import" tab reached it (`importIsRunning()` is false after a reload), and `onFinish` pushed step 4 over step 3, so the browser's own Back pointed at the same place.

`ImportWizardSteps` now takes a `lowestStep`, which the page raises to 4 exactly when the run is gone, and the step-3 → 4 transition replaces rather than pushes.

Verified against a real hard reload on step 4:
- stepper "Import" tab → stays on `step=4`, toast explains why
- browser Back → lands on `step=2`, no runnable import offered

Also added the migration-phase abandonment assertion — that no task is left on `running` when a run stops. 62 tests passing, svelte-check clean.

Guilhem-lm:
/review

Round 7. The round-6 P2 was right and worth spelling out: my migration-phase test asserted over a branch it never ran. The mock never called `onMigrationsStart`, and with `migrations: []` in the export the `migrate` row was never appended — so "no task is left running" held because no task existed. The comment was wrong too; `onMigrationsStart` fires at the head of the migration loop, past every item loop.

Fixed in `7a6991e3dd`: the mock mirrors the real order (item loops → `onMigrationsStart` → migrations, `stopped` checked before each write), a second hook abandons once the row is running, the export ships a migration so the row exists, and the test asserts it exists before asserting its status.

Verified by deleting the fix: fails with `expected 'running' not to be 'running'`, passes with it restored.

62 tests passing, svelte-check clean at the branch baseline.

Guilhem-lm:
### Changes that reach beyond project imports

Audited every shared component this PR touches. Five changes affect existing flows, and
they are easy to miss among the wizard code:

| change | reach | what it does |
|---|---|---|
| `validateWorkspaceId` rejects `global` | 3 callers of `CreateWorkspaceInner`, incl. onboarding | New rule — `main` has none. The backend's `check_w_id_conflict` refuses it while `existsWorkspace` reports it free, so today you reach the last step before the create fails. |
| `RadioCard` selected background | all 5 call sites | `bg-surface-selected` → `bg-surface-accent-selected`, matching FileExplorer / TriggersTable / RunnableRow / ArtifactVersionPicker. Repaints every existing radio card. |
| `FolderPicker` records ownership before `loadFolders()` | all 4 callers | `loadFolders` derives `write` from membership, so a just-created folder currently comes back `(read-only)` until the next load. |
| onboarding honours `?rd=` | anyone finishing onboarding with that param | `main` always went to `/user/workspaces`. Same-origin-validated via `toSameOriginRelativePath`. |
| `AppConnectDrawer`: `handedOff` + Back condition | 3 callers | **Two live bug fixes**, see below. |

The `AppConnectDrawer` pair are user-visible bugs on `main` today, not new-feature code:

- **Blocked OAuth popup.** The hand-off reactive statement re-runs both when `rtToLoad`
  changes and when `appConnectInner` binds — and it rebinds on every opening, since the
  drawer destroys its content on close. A second `open()` runs `next()` again, walking a
  drawer opened on a resource type past the Connect button into `window.open`. A popup
  opened from a reactive effect rather than a click is blocked, leaving "Finish connection
  in popup window" with no popup. Guarded with a `handedOff` flag — a flag rather than the
  last resource type, because the resources page opens the drawer with no argument and
  `undefined === undefined` would skip the hand-off entirely.
- **Back button on a list you never chose from.** Opening the drawer for one resource type
  skips step 1, so `{#if step > 1}` offered Back into a type picker the user never saw.

Verified inert despite wide reach: `CenteredModal` (19 callers), `ResourceEditorDrawer`
(12 — `closeDrawer()` still fires before the await, so timing is unchanged), `SetupChecklist`
(5), `AppConnectInner` (4 — every change gated on `fillPath`), `oauthRegistry.ts` (extraction
checked equivalent to the inline originals), `Login.svelte` / `login/+page.svelte` (a path
added to an existing OR). `projectInstall.ts` has no callers outside the wizard now that the
old install page is gone.

### Not a breaking change for older hubs

`/projects/install` is deleted as a page but kept as a 307 redirect to `/projects/import`
with the query string intact, so a self-hosted hub still pointing at the old URL keeps
working. The version coupling is one-directional: the Hub release needs this, this does not
need the Hub release.

PR #10733, opened 2026-08-17
fix: derive a raw app's policy on deploy, and default an omitted execution_mode

rubenfiszel:
## Summary

Fixes GIT-966.

The MCP `createApp` tool exposes `policy` as required but every field in it as optional, so a client can legitimately send `"policy": {}`. Two things were wrong with that:

1. **It didn't deserialize.** `Policy::execution_mode` had no serde default, so the request 400'd before reaching the handler. That's the reported bug.
2. **Fixing only that would have been worse.** `triggerables_v2` is the allowlist `execute_component` matches every run against, keyed by `<component>:rawscript/<sha256 of the inline code>`. A model can't compute those keys, so an accepted empty policy deploys an app whose every runnable fails at run time with `Policy is missing triggerables` — a loud 400 becomes a silent broken app.

So the server derives the policy itself on the raw-app **source** endpoints (`create_raw_source` / `update_raw_source`, which exist only for MCP), and only defaults `execution_mode`.

**The derivation is the frontend's own, vendored into the bundle job.** `frontend/src/lib/components/raw_apps/rawAppPolicy.ts` is what the app editor and `wmill app push` already run; `cli/generate-app-policy.ts` bundles it to `backend/windmill-api/src/apps_raw_policy.gen.js`, which is `include_str!`d into the server binary and prepended to the job script. Same pattern AGENTS.md documents for `devRecorderBundle.gen.ts`.

It rides in the job rather than being read off the `wmill` the job already runs, because the images install `windmill-cli` **unpinned** (`Dockerfile:293`, `DockerfileSlim:77`, `DockerfileSlimEe:77`) and `docker-image.yml` / `npm_on_release.yml` are independent workflows on the same tag — so image vX generally carries CLI v(X-1), and falling back to `bun x windmill-cli@<release>` needs npm reachable at deploy time, which is exactly what the bundler's installed-CLI branch exists to avoid. Vendoring means the derivation always matches the server, on any image, offline. It also sidesteps the stale published `@windmill-labs/shared-utils@1.0.12` (see Notes).

`execution_mode` stays in Rust, because it is permissioning rather than derivation:

- **Create** resolves an unstated mode to `publisher`, never `anonymous` — the issue suggested reusing `ExecutionMode::default()` (which was `Anonymous`), but that would let any caller publish a publicly-executable app by leaving a field out.
- **Update** keeps the *deployed* mode when the caller states none. `Policy::execution_mode` is now a method over an `Option`, so "not stated" is distinguishable from "stated `publisher`".

## Changes

- `cli/generate-app-policy.ts` + `bun run gen:app-policy` vendor the derivation; `cli/test/app_policy_bundle_unit.test.ts` rebuilds it and fails on drift, and exercises the committed artifact for the key format and `sensitive_inputs`.
- The bundle job derives `triggerables_v2` and **rejects runnables it can't classify**, naming them, instead of dropping them into a policy that would refuse them at run time. An explicitly empty entry is an unconfigured runnable and still needs no grant.
- The policy merge moved *inside* `update_app_internal`, under a single `FOR UPDATE` read of the app row that now serves the inherited mode, the already-anonymous check, and the carried-over policy. Previously the read was outside the transaction and before a long bundle job, so a concurrent policy edit was clobbered.
- A source deploy that sends no policy no longer re-points `on_behalf_of` at the updater; submitting a policy is what claims the app's execution identity.
- `Policy::execution_mode` is an accessor over `Option<ExecutionMode>`; `#[default]` moves from `Anonymous` to `Publisher`.
- `openapi.yaml` documents that omission means `publisher` on create and "keep the deployed mode" on update (the schema is shared by both tools); MCP tool schemas regenerated.

## Why no test caught this

The raw-source endpoints have integration tests, but they all stop *before* the bundle — the harness has no worker, so they assert the up-front rejections and treat reaching the compile as success (`"a group member must reach the compile"`). Everything after "queue the bundle job" was untested, and the policy is produced exactly there. Nothing anywhere calls `execute_component`, so the consequence of a bad policy had no coverage at all.

## Test plan

E2E against a local backend (`--features quickjs,mcp`), with the box's **installed `wmill` at 1.747.0 — 43 releases behind the server** — doing the bundle, which is the version-skew case this design has to survive. App under test carries three runnables: an inline script with static + sensitive + user fields, a deployed script by path with a static field, and a deployed flow.

- [x] `createApp` **through the MCP tool** with `"policy": {}` returns the created path (was an `ExceptionGroup` from the 400).
- [x] Stored policy carries all three derived keys plus `sensitive_inputs`, which the CLI's pinned `shared-utils` would have dropped.
- [x] **The derived policy authorizes**: running with `{"tenant": "ATTACKER-OVERRIDE", "pinned_secret": "ATTACKER"}` returned `{"tenant": "acme-corp", "pinned_secret": "from-the-publisher", "viewer_arg": "from-viewer"}`. Same for the path runnable.
- [x] A **viewer-supplied** field marked sensitive is stored as `$encrypted:…` in `v2_job.args`. (Static fields bypass `safe_args` and are never encrypted — pre-existing, unchanged.)
- [x] A runnable missing its `type` discriminator now fails the deploy naming it, instead of deploying a dud; a sibling valid runnable and an explicitly null entry are unaffected.
- [x] `update_raw_source` with changed sources and **no policy**: triggerables re-keyed to the new sha, and a planted `on_behalf_of: u/someone_else` survives. An update that *does* send a policy still claims the identity.
- [x] An app pinned to `viewer`, updated with `{"policy": {"triggerables_v2": {}}}`, stays `viewer`.
- [x] An app with no runnables deploys with `triggerables_v2: {}`.
- [x] `cargo test -p windmill-api --lib apps::`, `cargo check --features private,enterprise`, `bun test test/*_unit*` (1065 pass), `npm run check` (0 errors).

## Notes for review

- **`@windmill-labs/shared-utils@1.0.12` is stale and `wmill sync push` is affected** — separate from this PR, which no longer depends on it. The published package predates `sensitive_inputs`/`delete_after_secs` (#8975, 2026-04-29) and inline-script `tag` (#9005, 2026-05-01); the version constant in `frontend/sharedUtils/vite.sharedUtils.config.js` was last touched 2026-02-28 and there is **no publish workflow for it** in `.github/workflows`. So every raw app pushed via CLI/git-sync since April has a policy missing those three: sensitive viewer inputs aren't encrypted, retention is ignored, and inline jobs lose their worker tag. Worth a republish (bump that constant, then the `cli/package.json` pin) independently of this PR.
- There is also a stray `console.log('hex', …)` in `rawAppPolicy.ts` that now shows up in the bundle job's log.
- No screenshots: the frontend diff is generated MCP metadata plus comment removals.


rubenfiszel:
✅ Review round clean @ 5e2be9768d3b453f159de2d9aa41c5ff58998479

PR #10737, opened 2026-08-18
test: settle the backlog before the capped audit-export drain

rubenfiszel:
## Summary

Fixes the `ee::audit_s3_export::audit_s3_export_tests::audit_export_end_to_end` flake on `main` ([run 32082482567](https://github.com/windmill-labs/windmill/actions/runs/32082482567/job/95548012447)):

```
backlog row 20 lost during capped drain (floor must be carried, not advanced, while draining): exported []
```

Test-only, and EE-side: the whole change is in `backend/src/ee.rs` (symlinked from `windmill-ee-private`), so this PR's own diff is just the `ee-repo-ref.txt` bump. The production exporter `export_audit_logs_to_object_store` is untouched.

EE companion PR: windmill-labs/windmill-ee-private#730

The backlog-drain sub-test anchors a checkpoint at the current snapshot xmin, writes a 10-row backlog (6 recent rows, 4 fabricated 10-day-old rows in an old partition), forces the per-run xid cap to 3, and ran the exporter a fixed 40 times before asserting every row landed.

XIDs and the snapshot xmin are cluster-wide, so a neighbor `sqlx::test` running its migration set pins the xmin at the anchor for hundreds of ms, and all 40 back-to-back runs then process an empty `[prev, cur)` interval. Worse, a run that sees `cur_xmin == prev_xmin` has `draining == false`, so it advances the stored floor to the recent in-flight timestamp, which permanently prunes the fabricated 10-day-old rows. One stalled run poisons the whole drain. That floor advance is sound in production (a real audit row's `timestamp` is its inserting transaction's `xact_start`), so this is purely an artifact of the fabricated timestamps.

## Changes

- `backend/ee-repo-ref.txt`: bump to the EE commit carrying the fix.
- (EE) Anchor the seeded checkpoint at the snapshot **xmax** (first unassigned xid) instead of the xmin, so the crawl length is bounded by the sub-test's own inserts rather than by however far a neighbor's pinned xmin lags behind the xid counter. Measured span drops to 12 xids / 4 capped runs, stable across runs.
- (EE) Wait for the whole backlog to be settled cluster-wide (`wait_until_settled`) before the first capped run, so every run is genuinely draining and the floor can never advance mid-drain.
- (EE) Derive the run count from the settled snapshot (`(x1 - x0) / 3 + 2`, breaking early once all ids land) instead of a fixed 40, and report the run count and xid span in the assert message so a future failure is self-diagnosing.

## Test plan

- [x] `cargo test --bin windmill --features private,enterprise,parquet` green (16/16, `--test-threads=10`)
- [x] Flake reproduced with the old code under a pulsing `BEGIN; SELECT txid_current(); pg_sleep(0.4); COMMIT;` loop on the same cluster: 2/8 red, including the exact CI signature `exported []` and a partial `exported [20, 21, 22]`
- [x] Same harness with the fix: 30/30 green (0.4s and 1.5s holds)
- [x] Guard still bites: reverting the production floor carry (`next_oldest_inflight_ts = oldest_inflight_ts`) is 4/4 red with the new anchor

---
## Summary by cubic
Stabilizes the EE audit S3 export end-to-end test by bumping `backend/ee-repo-ref.txt` to the EE change that settles the backlog before a capped drain. Previously the test anchored at snapshot xmin and could advance the floor during stalled runs; it now anchors at xmax, waits for cluster-wide settlement, and derives the run count so drains complete without advancing the floor.

**Review notes**
- Diff is a single EE ref bump; test changes live in the EE repo.
- Production exporter `export_audit_logs_to_object_store` is untouched.
- No migration or rollout actions.

<sup>Written for commit dcf0fcee7adf9a7370c2b00e2c55f67fd6542adc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 6091a5e6f7aeffc385dfb487efe0b2e9fe49c0a1

PR #10738, opened 2026-08-18
fix: show runtime-detected assets in a run's Assets tab

rubenfiszel:
## Summary

Fixes GIT-963. A run's **Assets** tab showed "No assets found" even when the workspace `/assets` page attributed an S3 write to that exact run.

Runtime-detected assets (SDK S3 calls, resources passed as arguments) are recorded in the `asset` table with `usage_kind = 'job'` and `usage_path = <job id>`. The run page never read them: `JobAssetsViewer` only inferred assets statically from the script/flow source plus the job arguments, so anything whose path is only known at runtime was invisible — which is exactly the reported case (a workflow-as-code script building an S3 key from `get_root_job_id()` and loop values).

The recorder attributes an asset to the job that performed the operation, which for a flow step or a workflow-as-code task is *not* the job the user opened, so the new endpoint aggregates over the whole job tree: a WAC parent run now reports what its tasks wrote, and a flow run what its steps did.

Review round on the current head is clean — codex, claude and pi all "Good to merge", no open findings. Note this adds a public API endpoint and changes `openapi.yaml` / the generated client, so those are the parts worth a careful human read.

## Changes

- **New endpoint** `GET /w/{workspace}/jobs/run_assets/{id}` (`backend/windmill-api/src/jobs.rs`): runtime assets of a job and its descendants (recursive CTE on `v2_job.parent_job`), with per-asset access types merged across the tree.
  - It lives with the job routes, next to `run_progress`, for the reason that one documents: `asset` has no RLS, and `require_job_read_access` is what applies the view token, a scoped token's tag filter and the app-embed cutoff.
  - The tree is walked on the privileged pool after that gate — as `collect_flow_log_entries` does — because the gate grants access the caller's own RLS does not have (a share link, or a job they launched that runs as someone else); re-filtering through `user_db` would hand a share-link viewer the empty tab this PR exists to fix. A scoped token's tag filter is re-applied when selecting which jobs' assets are read, since the gate only checked the root.
  - Merging across jobs treats a missing access type as "that job recorded none" rather than "unknown", so a parent that only saw a resource in its arguments does not erase the write its step recorded.
  - The list is capped at 1000 assets with a `truncated` flag, because a fan-out run touches as many assets as it has steps and the whole list would otherwise land in one response and one DOM list. The per-asset access types are merged in SQL (`bool_or` over the group) so the cap counts assets rather than asset rows — retention keeps up to ten job rows per asset, so a row-counted cap would have delivered a tenth of what it advertised.
- **`JobAssetsViewer.svelte`**: merges those runtime assets with the existing static inference (runtime wins, falling back to the statically inferred access type when the recorder had none), renders the Read/Write access type, shows a skeleton while the request is in flight, says when the list was capped, and says in the empty state that runtime recording is asynchronous and retained only for the most recent runs that touched an asset.
- **Integration tests** (`backend/windmill-api-integration-tests/tests/assets.rs`): tree aggregation and access-type merging (including the missing-type case), a member who cannot read the run getting 403, a share-link viewer getting the whole tree, and a tag-scoped token reading in-scope descendants but not out-of-scope ones. Each authorization and merging assertion was confirmed to fail against the pre-fix code.

Not changed, and worth knowing: runtime assets are inserted by a batched loop that flushes every 120s and pruned to the 10 most recent runs per asset, so a just-finished run can take a couple of minutes to show and an old one may never show. That is the same behaviour the workspace `/assets` drawer already carries an alert for; the empty state now says so rather than the tab polling for it.

## Screenshots

Workflow-as-code run whose `writeFile` task writes an S3 object at a runtime-computed path — the parent run's Assets tab now lists it with its access type (previously "No assets found"):

wac-run-assets

Empty state for a run that touched nothing:

empty-run-assets

## Performance

Measured with `EXPLAIN (ANALYZE)` on a synthetic workspace of 305k `usage_kind='job'` asset rows and 40k jobs. Every shape plans the same way — recursive walk on `ix_v2_job_parent_job`, then a nested loop probing `idx_asset_usage (workspace_id, usage_path, usage_kind)`. No hash join over the asset table and no sequential scan at any size, so cost tracks the size of the run rather than the size of the workspace.

| Run | Query |
|---|---|
| single job, 305k workspace assets | 0.28 ms |
| 2000-level deep chain | ~6 ms |
| 201-job flow, 200 assets | 0.85 ms |
| 50 001-job flow, 45 000 assets | ~455 ms |

The extreme row is a forloop with 50k steps each writing an object. `ORDER BY` means the cap does not save the join there — the ~455 ms is 45k index probes plus a top-N sort — but it does bound the response and the DOM. For scale, the flow-tree endpoints (`get_flow_all_logs*`, `get_flow_all_results`) run the same recursive enumeration over the same subtree, ~48 ms for that 50k-job tree, so the tree walk is not a new class of cost; the asset fan-out is. All of them are user-triggered rather than page-load work — those behind "Download all logs" and an AI-chat tool, this one behind selecting the Assets tab, which is the only thing that mounts `JobAssetsViewer`.

The cap follows `get_flow_all_results`, which bounds the same enumeration at `FLOW_ALL_RESULTS_MAX_ENTRIES = 2000` and reports it with a `truncated` flag on the response object.

The read gate adds a `created_by` lookup plus `require_job_read_access`, whose ancestor probe is memoized in `JOB_READ_ACCESS_CACHE`; `require_job_within_run_scope` early-returns for unscoped callers, which is the common case.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test assets` — 3 tests pass; the authorization and merging assertions were each re-run against the pre-fix code (RLS-scoped walk / no descendant tag filter / absorbing merge) and fail there.
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api --features private,parquet`, `npm run check` (0 errors).
- [x] End to end on a dev instance (`--features quickjs,private,parquet`, filesystem workspace storage): a bun script and a workflow-as-code script each writing an S3 object at a runtime-built path. Before: both runs' Assets tabs empty while `/assets` listed the object against the run. After: the leaf run, the WAC task's own run, and the WAC parent run all list it as a Write.
- [x] `GET /jobs/run_assets/{id}` with an `X-View-Token`, and for an unknown job id (empty list, matching `run_progress`).


rubenfiszel:
✅ Review round clean @ 1e7f3059e0b1a707c6572bd049208f9a976b1427 — nit-only verdicts at a9c9f0eec22c3ce732a0aa9a23bac485067e1cc4 (codex: recursive tag filter pruned in-scope nested jobs; claude: runtime-first dedupe could drop a statically inferred access type, and the empty state showed while the request was in flight); all three addressed in 1e7f3059e0b1a707c6572bd049208f9a976b1427.

rubenfiszel:
✅ Review round clean @ 0809b368fcf3065c3c5c0d826586f1b44bdcad10 — codex, claude and pi all "Good to merge", no open findings.

PR #10741, opened 2026-08-18
fix(cli): keep script settings on push and repair the up-to-date check

rubenfiszel:
## Summary

Follow-up to #10731, which fixed the same class of bug in the script editor. The CLI builds its `createScript` body by listing fields explicitly, and five script settings were never on that list, so `wmill script push` / `wmill sync push` (and every git-sync deploy that runs the CLI) sent them as absent and the backend persisted `NULL` on the new version:

- `delete_after_secs`
- `cache_ignore_s3_path`
- `debounce_args_to_accumulate`
- `max_total_debouncing_time`
- `max_total_debounces_amount`

`create_script`'s no-op guard compares these against the parent, so the push is not filtered out as a redundant deploy: it lands and clears the setting. A dev to git to prod pipeline therefore wipes retention and debounce configuration on prod every time the script content changes.

None of them was ever CLI-managed (`delete_after_use`, the predecessor of `delete_after_secs`, was not in the body either), so nothing regressed here; the fields simply never round-tripped, and the loss is silent.

`cache_ignore_s3_path` had a second gap: the workspace tarball never exported it, so it did not reach the metadata file at all and could not be managed from git-sync in either direction.

## Changes

- `cli/src/commands/script/script.ts`: send the five settings in the `createScript` body, and compare them when deciding whether a push is a no-op. `delete_after_secs` deliberately skips `nonePositiveInt`: 0 means "delete immediately after completion", not "unset".
- `backend/windmill-api/src/workspaces_export.rs`: export `cache_ignore_s3_path` in the script metadata. It is `skip_serializing_if = "Option::is_none"`, so only scripts that set it gain a key.
- `cli/test/script_runtime_settings_sync.test.ts`: integration regression test, alongside the existing `script_envs_sync.test.ts` which covers the same class of bug for `envs`.

## Test plan

- [x] `bun test test/script_runtime_settings_sync.test.ts` passes; fails with the CLI change reverted (`Expected: 900, Received: undefined`) and fails with the export change reverted (`Expected to contain: "cache_ignore_s3_path"`).
- [x] `bun test test/script_envs_sync.test.ts test/sync_push_script_metadata_only.test.ts` still green (6 pass, 0 fail with the new test).
- [x] Manual round trip against a local instance: created a script carrying all eight settings, `sync pull`, edited the content, `sync push`. Before: `delete_after_secs` 900 to NULL, `cache_ignore_s3_path` true to NULL, `debounce_args_to_accumulate` ['a'] to NULL, `max_total_debouncing_time` 60 to NULL, `max_total_debounces_amount` 5 to NULL. After: all preserved.
- [x] Settings-only edit (`delete_after_secs` 900 to 120) now deploys instead of being reported as up to date, and `delete_after_secs: 0` survives as 0.
- [x] `bunx tsc --noEmit`: no new errors.

---
## Summary by cubic
Keeps script retention, debouncing, and cache S3-path settings on CLI pushes and workspace exports, and fixes the “up to date” check so metadata-only edits deploy and unchanged bunnative scripts are skipped. Previously, pushes omitted these fields and exports skipped `cache_ignore_s3_path`, saving NULL on new versions; the check missed `language`/`labels`/numeric `priority` and treated bunnative as a change.

- CLI: send and compare `delete_after_secs` (0 means delete immediately), `cache_ignore_s3_path`, `debounce_args_to_accumulate`, `max_total_debouncing_time`, `max_total_debounces_amount`; compare `language` with bunnative normalized to bun, `labels`, and `priority` by value; include `delete_after_secs` in the check; read `cache_ignore_s3_path` from the typed response.
- Export: include `cache_ignore_s3_path` in script metadata when set.
- Tests: add integration tests for settings round-trip and the push “up to date” path, including bunnative; surface the `scripts/create` response on fixture failure.

**Rollout**
- For `cache_ignore_s3_path`, deploy backend and CLI, then re-pull the workspace before pushing; a checkout from a pre-fix server lacks the key and the first push after upgrading the CLI can clear it until re-pulled.

<sup>Written for commit 860506fb5ff5d792663d70ce41cd5dc065607178. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10741?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Rollout note

For `cache_ignore_s3_path` only, both halves have to be deployed before a repo is safe: a checkout taken against a pre-fix server has no key in `.script.yaml`, so the first push after upgrading the CLI still clears it until the workspace is re-pulled. The other four fields were already exported, so the CLI half covers them on its own. Worth a release-note line.

## Why the test does not cover the debounce bounds

A backend built without git tags reports a bare commit as its version (`CE 2506cf6`), `GIT_SEM_VERSION` then falls back to `0.1.0`, and `get_min_version` uses that when no other worker has pinged. Every version constraint is therefore unmet on such a build, and `guard_script_from_debounce_data` refuses any script carrying debounce settings with `Debouncing is unavailable due to some workers being behind ... at least 1.566.0`.

That is exactly what CI builds (`actions/checkout` fetches no tags), so a fixture that sets any debounce field fails at `scripts/create` there while passing locally. The committed test therefore asserts `delete_after_secs` and `cache_ignore_s3_path`, which are not version-gated; the three debounce fields travel through the same object literal and were verified by hand against a local instance, as described in the test plan.

Worth its own follow-up: no CLI test can exercise a version-gated feature today. Adding `fetch-tags: true` to the CLI test workflow's checkout would fix the class, but it can flip other version-gated paths that currently run as "fleet is old" in CI, so it does not belong in this PR.

## Now carries #10743 as well

#10743 was stacked on this branch and has been merged into it, so this PR is the single one left for `main` and covers both halves:

- the five settings missing from the `createScript` body and from the up-to-date comparison (this PR's original scope)
- the repaired up-to-date comparison itself: `priority` compared by value instead of against a boolean coercion, plus `language` and `labels`, which the body sends but nothing compared

Their ordering constraint is satisfied by construction now that they are one branch, in that order.

With #10742 in `main`, the `(remote as any)` cast is gone: `cache_ignore_s3_path` is read off the typed response.


rubenfiszel:
On the nit (`cache_ignore_s3_path` missing from the `Script` response schema, which forces the `as any` cast): that gap is real and is fixed in #10742, which declares the field on the `Script` schema and is green on all three reviewers. Keeping the cast here is deliberate: the generated clients are not checked in, so a plain property access would fail CI typecheck on this branch until #10742 is in `main`.

Merge order is #10742, then this PR with the cast dropped and its comment removed, then #10743. I verified that edit on a local branch with all three merged: `Boolean(remote.cache_ignore_s3_path)` typechecks once the spec change is present.

rubenfiszel:
Addressed the round:

- **P1, `bunnative` never matches (claude)**: confirmed and fixed. The comparison now normalizes `remote.language === "bunnative"` to `bun`. That is lossless because `create_script` re-derives the language from the `//native` annotation in the content (`scripts.rs:1327-1336`), and the content is compared three lines above, so adding or removing the annotation still deploys. New test `an unchanged bunnative script is not redeployed` pins it: with the normalization reverted it fails.
- **P2, settings-only phase races the dependency job (both reviewers)**: the retention test now settles with `waitForDeploymentJobs` and re-pulls before editing the setting, so the deploy is caused by the setting rather than a stale lock. Verified the phase now pins the comparison term: removing `typed.delete_after_secs == remote.delete_after_secs` from the gate (leaving the body field in place) fails it with `Expected: 0, Received: 900`, which it did not before.
- **Secondary, `deno` scripts when `wmill.yaml` has no `defaultTs`**: left alone. The export writes them as plain `.ts` and the CLI infers `bun`, so the body has always converted them; that is a round-trip gap in the exporter and the config warning, not something this comparison can fix, and it behaves exactly as it does on `main` today.

Local run on this head: 87 pass, 3 skip, 0 fail across the seven push/sync integration files.

rubenfiszel:
✅ Review round clean @ 860506fb5ff5d792663d70ce41cd5dc065607178

PR #10743, opened 2026-08-18
fix(cli): repair the script push up-to-date comparison

rubenfiszel:
## Summary

`pushScript` short-circuits with "Script X is up to date" when the local script already matches the remote. That comparison has been dead since it was written: it reads

```ts
typed.priority == Boolean(remote.priority)
```

An unset local priority against a null remote priority is `undefined == false`, which is false, so the branch could not be taken. Any priority other than exactly `1` fails too (`3 == true` is false); `1 == true` is the only value that passes, which is why the short-circuit appears to work when you happen to test with it.

The consequence is that every CLI push of an unchanged script deploys a new version and spawns a dependency job. Two settings the body sends were also never compared, so with the gate live they would silently never deploy: `labels`, and the `language` inferred from `defaultTs` (flipping `defaultTs` changes the language of an unchanged `.ts` file).

## Stacked on #10741

This PR targets `fix/cli-script-settings-round-trip`, so the diff shown is incremental and the ordering both reviewers flagged is structural rather than a note: making the gate live while `delete_after_secs`, `cache_ignore_s3_path` and the debounce bounds are still absent from the body and the comparison would turn their clobber-on-deploy into a silent skip. GitHub retargets this to `main` when #10741 merges.

Because #10741's changes are in scope here, this branch also drops the `priority: 1` fixture and its now-false comment from `script_runtime_settings_sync.test.ts`: with the comparison repaired, that test no longer needs a priority to reach the gate.

## Changes

- `cli/src/commands/script/script.ts`: compare `priority` by value, and add `language` and `labels` to the comparison. Those three are now the complete set of fields the push body carries, apart from `deployment_message`, which is per-deploy audit metadata that the backend's own no-op check ignores as well.
- `cli/test/script_push_up_to_date.test.ts`: integration test covering both directions, an untouched script skipping and a changed one deploying, including a skip at `priority: 2` (the value the old comparison got wrong, unlike `1`).

## Known gap, not addressed here

A `bunnative` script exports to the same extension as `bun`, so `language` infers back as `bun` and this comparison is permanently false for it. That is the pre-existing round-trip gap (a push already rewrites such a script's language to `bun`), not a regression from this diff, and it wants its own fix.

## Test plan

- [x] `bun test test/script_push_up_to_date.test.ts` passes; fails without the fix (`Expected to contain: "is up to date"`). Run three times in a row after adding the dependency-job drain, green each time.
- [x] Change-detection matrix against a local instance: with the gate live, an untouched script reports up to date, and a lone change to each of summary, description, tag, cache_ttl, timeout, priority, concurrent_limit, concurrency_time_window_s, concurrency_key, debounce_key, debounce_delay_s, restart_unless_cancelled, visible_to_runner_only, envs, labels, kind, content and language still deploys and lands on the remote. 18/18.
- [x] `bun test` over the push/sync integration suite (`sync_pull_push`, `script_envs_sync`, `sync_push_script_metadata_only`, `sync_push_auto_metadata_repro`, `folder_schedule_push`): 84 pass, 3 skip, 0 fail.
- [x] `bunx tsc --noEmit`: no new errors.


rubenfiszel:
Addressed the round:

- **P1, dependency-job race in the test**: the test now drains with `waitForDeploymentJobs(backend)` before every `sync pull`, same as `sync_pull_push.test.ts`. Ran it three times in a row, green each time.
- **P1, merge ordering**: this PR is now based on `fix/cli-script-settings-round-trip` (#10741) rather than `main`, so the ordering is enforced by the branch rather than by a note in the description. It also drops the `priority: 1` fixture and its stale comment from #10741's test, which is only valid once this change is in.
- **Codex test coverage ask**: added a skip assertion at `priority: 2`, the case the old comparison got wrong where `1` accidentally worked. Unset priority was already covered by the baseline.
- **P2, `bunnative`**: agreed it is pre-existing and out of scope here; recorded in the description as its own follow-up.

PR #10744, opened 2026-08-18
fix(agents): prove scratch file ops per command segment

hugocasa:
## Summary

The two scratch-dir PreToolUse guards only reasoned about a single plain one-line command and deferred on everything else. Since #10702 that deferral is an explicit `ask`, and a hook `ask` is not a permission rule: it prompts in every mode, `bypassPermissions` included, and no allow rule or "don't ask again" can silence it. A trailing `; rm -f probe.ts` was therefore enough to stall a session — and in headless or subagent runs, where nobody can answer, it becomes a hard denial.

Both guards now split the command into segments and prove each `rm` / `mv` / `chmod` / `cp` / `mkdir` / `touch` / `tar` / `unzip` invocation on its own operands, so chaining and line breaks carry no weight of their own. The fileops guard also gains the checkout as a root, so an in-repo `mv` or `chmod +x` is as auto-allowable as the in-repo delete already was.

Replayed over 6,672 real Bash commands taken from local transcripts: **prompts drop from 58 to 23**, with 14 `allow` decisions and no command auto-allowed that `main` did not already auto-allow.

## Changes

- `lib-guarded-verb.sh`: new shared primitives — `split_segments`, `segment_runs_verb`, `segment_tokens`, `apply_cd`, and `path_class` (the root classifier both guards share, which also holds the `.git` / `.claude` / `.env` / secret-glob carve-outs). `runs_verb` keeps its previous behavior, rebuilt on those.
- `guard-rm-outside-tmp.sh`: proves each `rm` segment; relative operands resolve against the working directory a `cd` moved to **and** the one the command started in, since a `cd` can fail at runtime.
- `allow-fileops-in-tmp.sh`: same segment treatment; every operand of one operation must land in a single root (a sibling checkout is a different root); one write per line; directory form (`cp x dir`) is left unproved because the tool then writes a path the command does not name.
- A decision covers the whole command line, so `allow` is emitted only when every segment is one the guard proved or a plain absolute `cd`. A line that also runs something unexamined returns **no decision** and falls back to the normal permission flow, rather than letting a trailing `rm -f /tmp/x` auto-approve its neighbours.
- `test-hooks.sh`: the decision table grows from 79 to 128 rows.
- `AGENTS.md`: the scratch bullet now describes what the guards actually do, plus a new bullet preferring Edit/Write over shell edits — the guards and the permission classifier match `Bash` and nothing else, so a heredoc edit arrives as a prompt.

## Review history

Eight review rounds (seven Codex, one Claude) found nine P0/P1 issues, each reproduced through the hooks before and after its fix. They are worth listing because they are what the design now defends against:

| Round | Issue | Resolution |
|---|---|---|
| 1 | `cd /tmp/nonexistent; rm -rf .git` auto-allowed — the `/tmp` early return preceded the `.git` carve-out | carve-outs apply in every root; operands proved against both candidate directories |
| 1 | `mv backend/server.pem backend/server.txt` renamed a key out of the `Read(**/*.pem)` deny glob | `path_class` refuses every path `.claude/settings.json` denies reading |
| 2 | `cd /tmp >~/.zshrc; rm -f /tmp/a` — a redirect rode along on an unparsed `cd` | a `cd` the guard cannot read whole drops the line off the allow path |
| 2 | `cp -r /tmp/tree /tmp/live; cp /tmp/payload /tmp/live/link` planted a symlink then wrote through it | one write per line (deletes still compose — `rm` unlinks rather than follows) |
| 3 | `CDPATH` — `cd ssh` is not `$cwd/ssh` | `apply_cd` takes absolute destinations only |
| 3 | ``rm -rf /tmp/a/`printf ../../etc` `` proved only the literal half of a split word | a command carrying a substitution is never proved |
| 4 | every checkout shared one `repo` class, so `mv ~/other-repo/x .` looked same-root | the class names the git root |
| 5 | `cp frontend/apps_ee.rs backend/windmill-api/src` wrote through an `*_ee.rs` symlink into the EE checkout | directory form left unproved |
| 6 | the derived name came from the resolved path, not the operand as written | (subsumed by the above) |
| 8 | a `cd` into a file hid the directory form, reopening round 5 | a relative operand after a `cd` is not proved for `cp`/`mv` |

The trusted surface that remains is small enough to state in one line: a command is proved only if it carries no substitution, and every segment is either a guarded verb whose operands all resolve inside one root, or a plain absolute `cd`.

## Test plan

- [ ] `bash .claude/hooks/test-hooks.sh` — 128 rows, from the main checkout **and** from a linked worktree (`path_class` treats a tree root differently depending on whether `.git` is a file or a directory)
- [ ] In a live session, confirm no prompt for: a two-`rm` cleanup on one line, an in-repo `mv` of a scratch file, `chmod +x` on a script
- [ ] Confirm a prompt still appears for: `rm -rf` outside `/tmp` and any checkout, `rm -rf "$VAR"`, `xargs rm`, `cd /etc && rm -rf x`
- [ ] Confirm `cp file dir/` now prompts — the cost the directory-form rule deliberately introduces

## Note on scope

The guards prompt in `bypassPermissions` too, which is what makes them the last check for anyone whose harness boots into yolo. Gating the deferral on `permission_mode` would take yolo prompts to zero, but with most sessions in bypass that would delete the guard for nearly everyone while leaving code in the tree that looks like protection — so this PR deliberately does not do it.

---
## Summary by cubic
Proves scratch file ops and deletes per command segment and per root, replacing the “single plain line” heuristic. Previously chaining, wrappers, or redirects forced an unconditional prompt and a trailing “; rm …” could stall headless runs; now each `rm`/`cp`/`mv`/`chmod`/`mkdir`/`touch`/`tar`/`unzip` segment is validated, and a line is allowed only if every segment is proved (or a resolvable absolute `cd`). Prompts drop 58→23 with no new auto-allow vs `main`.

- Treats `/tmp` and each git working tree under `$HOME` as distinct roots; `.git`, `.claude`, `.env`, `secrets/`, `*.pem`, `*.key`, `credentials.json`, and `.secret*` never auto-allow; sibling checkouts are separate roots.
- `allow-fileops-in-tmp.sh`: allows `mkdir`/`cp`/`mv`/`touch`/`chmod` when all operands of a segment stay within one root; `tar`/`unzip` remain `/tmp`-only with an explicit `/tmp` destination (or cwd in `/tmp` with no `cd` seen); refuses command substitution and relative `cd`; literal path operands only; one write per line; directory-form `cp/mv` prompt; relative operands are valid only if they land in the same root from both the pre- and post-`cd` cwd.
- `guard-rm-outside-tmp.sh`: proves each `rm` segment; globs auto-allow only under `/tmp`; relative operands must be valid from both candidate cwds after `cd`.
- Shared lib: centralized parsing and rationale (`split_segments`, `segment_tokens`, `segment_runs_verb`, `has_substitution`, `apply_cd`, `path_class`). A decision covers the whole line; any unexamined neighbor yields no decision and falls back to normal permissions.
- Tests/docs: matrix grows to 128; `AGENTS.md` documents roots, one-write-per-line, naming destinations, and preferring Edit/Write over shell edits.

**Rollout/Migration**
- Name destination files explicitly; directory-form `cp`/`mv` now prompt.
- Keep at most one write per line; split chained writes across lines.
- Use absolute paths or an absolute `cd`; relative `cd` or ambiguous working dirs will prompt.
- Keep `tar`/`unzip` extractions in `/tmp` and pass `-C /tmp/<dir>` if not already in `/tmp`.
- Auto-allow requires literal path operands; substitutions, `~`, quoted/variable paths, wrappers (`xargs`, `timeout`, etc.), or mixed segments will prompt.

<sup>Written for commit 5ecbd15fcabd0ef805e5b9a68f237187f2a54f7e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10744?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/hooks/allow-fileops-in-tmp.sh
+++ b/.claude/hooks/allow-fileops-in-tmp.sh
@@ -17,12 +17,14 @@
 # they can only constrain the FIRST operand. `cp /tmp/x ~/.zshrc` matches a `cp /tmp/` prefix,
 # and requiring every operand is the point.
 #
-# One operation may not straddle the two roots, sources included. A copy out of a checkout into
-# /tmp would be a read-exfiltration path around the `Read(**/secrets/**)` / `Read(**/*.pem)`
-# deny rules, since the content lands where `Read(/tmp/**)` allows it to be read back; keeping
-# every operand of one operation inside a single root closes that without restating the deny
-# list here. The checkout root itself is what makes an in-repo `mv` or `chmod` auto-allowable:
-# deleting a file there has never prompted, and moving or chmod-ing one is not the graver act.
+# One operation may not straddle two roots, sources included, and a sibling checkout is a
+# different root — `path_class` names the git tree, not just its kind. A copy out of a checkout
+# into /tmp would be a read-exfiltration path around the `Read(**/secrets/**)` / `Read(**/*.pem)`
+# deny rules, since the content lands where `Read(/tmp/**)` allows it to be read back, and one
+# out of a repo the Read tool is not confined to would do the same for that repo. Keeping every
+# operand of one operation inside a single root closes both without restating those rules here.
+# The checkout root itself is what makes an in-repo `mv` or `chmod` auto-allowable: deleting a
+# file there has never prompted, and moving or chmod-ing one is not the graver act.
 #
 # Deny-by-default tokenizing, in the same spirit as guard-rm-outside-tmp.sh: every path token
 # must consist only of alphanumerics and `. _ / -`. That set contains none of the characters
@@ -235,7 +237,7 @@ check_fileops_segment() {
 
     cls=$(operand_class "$t") || defer "\`$t\` is outside /tmp and not inside a git checkout in \$HOME"
     # Every operand of one operation stays in one root: see the exfiltration note above.
-    [ -n "$seen_class" ] && [ "$cls" != "$seen_class" ] && defer "\`$t\` puts this $verb across both /tmp and a checkout"
+    [ -n "$seen_class" ] && [ "$cls" != "$seen_class" ] && defer "\`$t\` puts this $verb across two roots"
     seen_class="$cls"
     path_operand=1
   done
--- a/.claude/hooks/allow-fileops-in-tmp.sh
+++ b/.claude/hooks/allow-fileops-in-tmp.sh
@@ -1,17 +1,28 @@
 #!/usr/bin/env bash
-# PreToolUse allowance for scratch file ops: auto-allow a single, plain, single-line
-# `mkdir` / `cp` / `mv` / `touch` / `chmod` / `tar` / `unzip` whose every path operand
-# resolves under /tmp. Anything else makes no decision (exit 0) and falls back to the normal
-# permission flow, except for `mv` and `chmod`: those get an explicit `ask`, the only prompt
-# they get (see lib-guarded-verb.sh).
+# PreToolUse allowance for scratch file ops: auto-allow `mkdir` / `cp` / `mv` / `touch` /
+# `chmod` whose every path operand resolves inside one of the roots `path_class` recognizes —
+# under /tmp, or inside a git working tree under $HOME — and `tar` / `unzip` confined to /tmp.
+# Anything else makes no decision (exit 0) and falls back to the normal permission flow, except
+# for `mv` and `chmod`: those get an explicit `ask`, the only prompt they get (see
+# lib-guarded-verb.sh).
+#
+# The command is read one segment at a time, so chaining and line breaks carry no weight of
+# their own: `mkdir -p /tmp/a && mv /tmp/b /tmp/a` is two operations, each proved on its own
+# operands. A decision covers the whole command line, so `allow` is emitted only when every
+# segment is one of these verbs proved here or a `cd` that resolved. A line that mixes a
+# proven op with some other command makes no decision instead and leaves that line to the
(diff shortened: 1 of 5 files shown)


hugocasa:
✅ Review round clean @ 5ecbd15fcabd0ef805e5b9a68f237187f2a54f7e — nit-only verdicts at 5a58edb782191b0388af22c335ed67b17ea8d34c (codex: good to merge, pi: good to merge, claude: nits); nits addressed in 5ecbd15fca

hugocasa:
Ran the decision table from a linked worktree as well as the primary checkout (test plan box 1). Three rows failed there — `rm -rf $CWD`, `chmod -R 777 $CWD`, and the chained `chmod` — because they hardcoded the primary-checkout expectation: a linked worktree's root IS auto-allowable (its `.git` is a pointer file, so the history lives in the main repo and survives), a primary checkout's is not. Test bug, not a code bug; `path_class` behaved as designed in both.

Those rows now follow the checkout the suite runs in, which also pins the linked-worktree half of that rule for the first time. `ALL PASS` from both kinds at 7d88a37dcf5b84a571ab88c50d1b1141d2234a58.

PR #10747, opened 2026-08-18
style: use subtle Button for raw app preview toolbar actions

Guilhem-lm:
## Summary

The preview tab bar in the raw app editor carried five hand-rolled `<button>` elements with the design system's *default* variant baked in as literal Tailwind classes. They read heavier than the tab bar they sit in, and the ES build label rendered bold because the raw button set no font weight and inherited one from the tab bar.

They now use the shared `Button` with `variant="subtle"`.

## Changes

- `RawAppEditor.svelte`: the bundler switch, element inspector, rebuild, open-in-new-window and split-view buttons are `<Button variant="subtle" unifiedSize="sm">`. The split-view toggle appears on both pane tab bars, so six call sites in total.
- `unifiedSize="sm"` keeps the previous 28px height and `text-xs`, and adds `font-normal` — that is what un-bolds the ES build label.
- Icon buttons use `iconOnly` + `startIcon`, so the icon size comes from the size scale (13px, was a hardcoded 14px).
- The two toggles (inspector, split view) use the `selected` prop instead of a hand-written `bg-surface-accent-selected text-accent border-border-selected` branch.
- `title`, `aria-label` and `aria-pressed` carry over unchanged.

## Screenshots

Before — bordered, bold `esbuild` label:

toolbar-before

After — subtle, regular weight:

toolbar-after

Selected state still reads (split view pinned):

toolbar-after-selected

In context:

rawapp-after-full

## Test plan

- [x] `svelte-check` — 0 errors
- [x] Open a raw app editor, confirm all five toolbar buttons render borderless at the same size and position as before
- [x] Toggle split view and confirm the button takes the accent-selected state
- [x] Click the bundler switch and confirm it flips esbuild ⇄ rolldown

### DIFF
--- a/frontend/src/lib/components/raw_apps/RawAppEditor.svelte
+++ b/frontend/src/lib/components/raw_apps/RawAppEditor.svelte
@@ -7,6 +7,7 @@
 	import RawAppYamlEditor, { type RawAppYamlUpdate } from './RawAppYamlEditor.svelte'
 	import type Drawer from '../common/drawer/Drawer.svelte'
 	import Alert from '../common/alert/Alert.svelte'
+	import { Button } from '../common'
 	import { AppService, type Policy, WorkspaceService } from '$lib/gen'
 	import DiffDrawer from '../DiffDrawer.svelte'
 	import { deepEqual } from 'fast-equals'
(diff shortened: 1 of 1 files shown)


Guilhem-lm:
✅ Review round clean @ a913200beccf01def69d182e134b71064d25e997 — nit-only verdicts at f513a5cfc83bf8a4938e31a41971919606bbee89 (codex: Good to merge, pi: Good to merge, claude: nit); nit addressed in a913200beccf01def69d182e134b71064d25e997

PR #10752, opened 2026-08-18
fix: make the listScripts parent_hash filter valid SQL

rubenfiszel:
## Summary

`GET /w/{workspace}/scripts/list` returned a 400 with `syntax error at or near "any"` whenever the optional `parent_hash` filter was supplied. The filter was built as `and_where_eq("any(parent_hashes)", hash)`, which emits `any(parent_hashes) = <hash>` — `ANY()` is only valid on the right-hand side of a comparison in PostgreSQL, so the endpoint could never serve the parameter it documents.

Fixes GIT-969

## Changes

- `backend/windmill-api-scripts/src/scripts.rs`: build the `parent_hash` predicate as `<hash> = ANY(parent_hashes)` via `and_where` + `bind`, matching the surrounding SQL-builder conventions. The hash is an `i64` newtype, so `bind` interpolates a bare integer literal.
- `backend/tests/list_scripts_parent_hash.rs`: new regression test driving the endpoint over HTTP — builds a two-version lineage plus an unrelated script, then asserts the filter returns the descendants of a hash, returns nothing for the head version, and that the unfiltered listing is unaffected.

No API-surface change: the `parent_hash` query parameter and its OpenAPI description already existed, and hash parsing and authorization are untouched.

## Test plan

- [x] `cargo test --features quickjs --test list_scripts_parent_hash` passes; reverting the one-line fix makes it fail with the exact `syntax error at or near "any"` from the report.
- [x] `--test script_auto_parent_archived --test runnables_list_pagination` still pass (the neighbouring script-listing coverage).
- [x] Exercised on a running dev instance: `scripts/list?parent_hash=<v1>` returns the v2 script, `parent_hash=<head>` returns `[]`, an unfiltered list is unchanged, and combinations with `show_archived`, `first_parent_hash` and `path_start` all return 200. A hash whose `i64` value is negative (`ffffffffffffffff`) also returns 200, confirming the bare integer literal is parenthesised safely by the builder.

## Note on reach

Creating a new version archives the parent row and `scripts/list` defaults to `archived = false`, so `parent_hash=<old hash>` surfaces the still-active descendant rather than every intermediate version; `show_archived=true` narrows to the newest row per path and so does not recover them either. Both are pre-existing properties of the surrounding filters, not introduced here — recorded so nobody expects this parameter to walk full version history.

---
## Summary by cubic
Fixes invalid SQL in GET /w/{workspace}/scripts/list when the optional parent_hash filter is used, so the endpoint now returns scripts whose parent_hashes contain the given hash. Previously it emitted ANY(parent_hashes) = <hash> and returned 400; now it builds <hash> = ANY(parent_hashes). Fixes GIT-969.

- Builds the predicate via the SQL builder with a bound value: "? = ANY(parent_hashes)".
- Adds an HTTP regression test that creates a two-version lineage and verifies: descendants of v1 are returned, head has no descendants, and unfiltered results are unchanged.
- No API changes or migrations. Existing list semantics remain: with defaults, only active descendants are shown; show_archived does not traverse full history.

<sup>Written for commit bab0154f4406fca4b6725c7958ed05fd731c9d39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10752?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/list_scripts_parent_hash.rs
+++ b/backend/tests/list_scripts_parent_hash.rs
@@ -1,0 +1,118 @@
+//! Pins the `parent_hash` filter of `GET /w/{workspace}/scripts/list`: the
+//! request must succeed and return exactly the scripts whose `parent_hashes`
+//! array contains the given hash. The filter is assembled into a raw SQL string,
+//! so a malformed predicate is only caught by executing the query.
+
+use serde_json::json;
+use sqlx::{Pool, Postgres};
+use windmill_test_utils::*;
+
+fn client() -> reqwest::Client {
+    reqwest::Client::new()
+}
+
+fn authed(builder: reqwest::RequestBuilder, token: &str) -> reqwest::RequestBuilder {
+    builder.header("Authorization", format!("Bearer {}", token))
+}
+
+fn new_script(path: &str, content: &str) -> serde_json::Value {
+    json!({
+        "path": path,
+        "summary": "",
+        "description": "",
+        "content": content,
+        "language": "deno",
+        "schema": {
+            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "type": "object",
+            "properties": {},
+            "required": []
+        }
+    })
+}
+
+async fn create(port: u16, script: serde_json::Value) -> anyhow::Result<String> {
+    let resp = authed(
+        client().post(format!(
+            "http://localhost:{port}/api/w/test-workspace/scripts/create"
+        )),
+        "SECRET_TOKEN",
+    )
+    .json(&script)
+    .send()
+    .await?;
+    let status = resp.status();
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ bab0154f4406fca4b6725c7958ed05fd731c9d39

PR #10757, opened 2026-08-19
feat: rework the resource type list in the add-resource drawer

Guilhem-lm:
## Summary

The "Add a resource" drawer listed 273 resource types as bordered chips in three columns, labelled by raw type name (`adobe_acrobat_sign`, `bigquery`), with each type's description reachable only by searching for words inside it. Picking a type meant scanning a wall of identifiers.

Each type is now a row carrying its product name, its type name, and its description, in a single borderless column under a sticky search field, drivable from the keyboard.

## Changes

- **Rows instead of a 3-column chip grid.** Product name + mono type name on the first line, description on the second, truncated to one line with the full text as a tooltip when it is cut off. Borderless, aligned with the drawer's own padding, 700px drawer (was 800px).
- **Display names.** `resourceTypeDisplayName` infers a product name from the type name (`mysql` → MySQL, `adobe_acrobat_sign` → Adobe Acrobat Sign, `c_acme_api` → Acme API), with a per-word table for acronyms and internal capitals and a whole-name table for the dozen that don't compose (`gdrive` → Google Drive). Checked against all 273 types on a hub-synced instance. It also feeds the drawer title, so `resourceTypeLabel` is gone.
- **Search ranks name → display name → description**, each field preferring an opening match over a word-start over a mid-word one. `microsoft` used to return `azure` first; it now leads with `ms_sql_server` and `ms_teams_webhook`. While a filter is active the native-database-first grouping is dropped, since it outranked the search itself (`sql` led with `mysql` over `ms_sql_server`).
- **Keyboard navigation.** The search field takes focus on open; ↑/↓ walk every row across all sections as one list; Enter opens the highlighted type. Hover and the arrow keys move one shared highlight — the variant's own hover is off so two rows are never lit at once.
- **A "Custom resource types" section** above the rest, for types made in this workspace (the `c_` prefix the resources page adds, or a non-`admins` `workspace_id`). Hidden when there are none.
- **Fix: the Google sign-in button on step 1.** `isGoogleSignin` was tied to `step == 1`, correct in 2022 when step 1's footer button started the OAuth flow. Today a row click advances to step 2, so the branded button only ever appeared after pressing Back from a Google type. It now tracks the control that actually starts the sign-in: `step == 2 && !manual`.

`ResourceTypePicker` (the schema-narrowing picker) keeps its 3-column grid — it renders in narrow panels where a description column doesn't fit — but picked up the same borderless, left-aligned rows.

## Screenshots

drawer-list

drawer-search

## Test plan

- [ ] Open Resources → Add resource: rows show name, type name and description; the search field has focus.
- [ ] Type `microsoft`: `ms_sql_server` and `ms_teams_webhook` come before the azure rows, and the first is highlighted. Enter opens it.
- [ ] Hold ↓ with the mouse parked over the list: the highlight advances one row at a time and does not jump back under the cursor. Tab into a row, then ↓: focus follows the highlight, so Enter opens the lit row.
- [ ] Hover a row with a clipped description: the full text appears as a tooltip. A short description gets none.
- [ ] In a workspace with a `c_`-prefixed type, the "Custom resource types" section renders above the others; with none, it is absent.
- [ ] Pick Google Sheets, then press Back: step 1 shows the normal "Next" button, not the Google sign-in image. With an instance Google OAuth client configured, picking it from the OAuth section shows the Google button on step 2.

## Notes

Verified against a local instance throughout, including the OAuth path (by configuring a dummy `gsheets` client and clearing it afterwards) and the custom section (by creating a `c_` type). `npm run check:fast` clean; `resourceTypeMatchRank.test.ts` covers the new display-name ranking tier and `resourceTypeDisplayName` itself (8 tests).

The local Codex review could not run — that account is over its usage limit until Aug 20 — so only the Claude branch review ran locally; its findings are addressed in 5c8b2ded39.

### DIFF
--- a/frontend/src/lib/components/AppConnectDrawer.svelte
+++ b/frontend/src/lib/components/AppConnectDrawer.svelte
@@ -57,7 +57,7 @@
 		step = 1
 		dispatch('close')
 	}}
-	size="800px"
+	size="700px"
 	{disableChatOffset}
 >
 	<DrawerContent
--- a/frontend/src/lib/components/AppConnectInner.svelte
+++ b/frontend/src/lib/components/AppConnectInner.svelte
@@ -4,7 +4,13 @@
 	import { userStore, workspaceStore } from '$lib/stores'
 	import LabelsInput from './LabelsInput.svelte'
 	import IconedResourceType from './IconedResourceType.svelte'
-	import { resourceTypeSearchText, sortResourceTypesByMatch } from './resourceTypeDisplay'
+	import {
+		isCustomResourceTypeName,
+		resourceTypeDisplayName,
+		resourceTypeMatchRank,
+		resourceTypeSearchText,
+		sortResourceTypesByMatch
+	} from './resourceTypeDisplay'
 	import {
 		OauthService,
 		ResourceService,
@@ -15,7 +21,7 @@
 	} from '$lib/gen'
 	import { emptyString, truncateRev, urlize } from '$lib/utils'
 	import oauthConnectRegistry from '$oauth_connect_registry'
-	import { createEventDispatcher, onDestroy } from 'svelte'
+	import { createEventDispatcher, onDestroy, tick, untrack } from 'svelte'
 	import Path from './Path.svelte'
 	import { Button, RadioCard, Skeleton } from './common'
 	import ApiConnectForm from './ApiConnectForm.svelte'
@@ -36,6 +42,7 @@
 	import SyncResourceTypes from './SyncResourceTypes.svelte'
 	import Label from './Label.svelte'
 	import ResourcePathHint from './ResourcePathHint.svelte'
+	import { twMerge } from 'tailwind-merge'
 
 	interface Props {
 		step?: number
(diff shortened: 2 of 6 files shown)


Guilhem-lm:
✅ Review round clean @ 9921b3a39c7e9255f4fbb4c2fe077f61798c8d96 — codex "Good to merge"; claude and pi nit-only. Nits addressed in 28e403b75a3916428ae8666fcce64937b94db2aa (whitespace-only query now counts as no search; `resourceTypeArticle` no longer says "an NATS"; the lightweight picker's wrapper gets a height so the step-1 list fills it the way it fills the drawer).

One caveat on that last one: the lightweight picker renders from the app editor's user-resource input and the flow AI-agent resource bar, and I could not get either open to exercise it — the container shape it now has is the one verified in the drawer, but that path went untested.

Guilhem-lm:
✅ Review round clean @ 744643e9bbf83ad63fc242f4cf2249b22583b11a — codex, claude and pi all "Good to merge".

PR #10758, opened 2026-08-19
fix: use constant-time comparison for SUPERADMIN_SECRET authentication

rubenfiszel:
## Summary

`SUPERADMIN_SECRET` was compared with Rust string equality (`==`), which short-circuits on the
first differing byte — a CWE-208 (Observable Timing Discrepancy). Practical remote exploitation is
infeasible (network jitter and TLS overhead dominate the signal, CVSS 3.7 Low), but every other
secret comparison in the codebase already uses `constant_time_eq` — HTTP trigger auth
(`http_trigger_auth.rs`), the Azure webhook signature (`mod_ee.rs`), MCP PKCE (`oauth_server.rs`) —
so this was an inconsistency in the one place that gates instance-wide superadmin access.

Both comparison sites now use `constant_time_eq`. Behavior is otherwise unchanged: this is a
byte-for-byte equivalent substitution, not a semantic change.

Fixes WIN-2406

## Changes

- `backend/windmill-api-auth/src/auth.rs` — `AuthCache::get_opt_job_authed` now compares the
  configured secret with `constant_time_eq`. `Option::map(..).unwrap_or(false)` collapsed into
  `is_some_and(..)`, which is identical for `None`.
- `backend/windmill-api-users/src/users.rs` — `global_whoami` now compares
  `std::env::var("SUPERADMIN_SECRET")` with `constant_time_eq`. `Result::is_ok_and` returns
  `false` for both `VarError::NotPresent` and `VarError::NotUnicode`, matching the old
  `.ok() == Some(token)`, which mapped both `Err` variants to `None`.
- `backend/windmill-api-auth/Cargo.toml`, `backend/windmill-api-users/Cargo.toml` — add
  `constant_time_eq.workspace = true` (the crate was already a workspace dependency; both
  additions are unconditional, so there is no feature-combination gap).

A repo-wide sweep (CE and the EE repo) found no other site that compares the secret *value* — the
remaining `SUPERADMIN_SECRET` hits are `SUPERADMIN_SECRET_EMAIL` identity checks against a
hardcoded constant, plus the single producer in `windmill-api/src/lib.rs` that feeds
`AuthCache::new`.

No test is shipped. The change is a comparison swap with no new behavior to pin; a unit test
asserting `constant_time_eq` would test the dependency, and a timing assertion would be flaky.
Existing integration coverage in `windmill-api-integration-tests/tests/{users,token_hash}.rs`
already exercises `/whoami`.

## Test plan

Verified end-to-end against a backend rebuilt with `SUPERADMIN_SECRET` set (all checks passed):

- [x] `auth.rs` site — `GET /api/w/admins/users/whoami` with the correct secret returns 200 as
      `superadmin_secret@windmill.dev` (`is_super_admin: true`).
- [x] `auth.rs` site — the same call with a same-length wrong secret, a different-length wrong
      secret, and no token each returns 401.
- [x] `users.rs` site — `GET /api/users/whoami` with the correct secret returns 200 with
      `login_type: "superadmin_secret"` and `super_admin: true`.
- [x] `users.rs` site — same-length wrong secret, one-byte-off-at-the-end wrong secret, and a
      proper prefix of the correct secret each return 401.
- [x] `users.rs` negative branch — an authenticated token whose email has no `password` row
      (service account, planted via SQL and removed afterwards) falls through to the
      service-account branch with `super_admin: false`, confirming the comparison returns false
      rather than short-circuiting into the superadmin payload.

---

**Left in draft: touches auth/token paths, wants a human look before ready.** The review round at
`711d8cd1431993a56e94303495970ee5b4235571` came back clean on all three reviewers (codex, claude,
pi), and two local passes found nothing — but per `.claude/skills/pr`, changes on auth/permission/
token paths are flipped to ready by a human rather than unattended. The clean-round marker is
posted, so flipping skips a redundant CI round.


rubenfiszel:
✅ Review round clean @ 711d8cd1431993a56e94303495970ee5b4235571

All verdicts "Good to merge" (codex, claude, pi). Left in draft deliberately — see the PR description.

PR #10760, opened 2026-08-19
feat: surface execution usage in the sidebar and explain what an execution is

Guilhem-lm:
## Summary

Users read "executions" as a job count and are surprised by the real number, which meters a second of compute. This makes that explicit everywhere the UI prints an execution count, and adds a usage meter to the sidebar for the quota that will bind first.

Everything here is cloud-only (`isCloudHosted()`), so self-hosted instances are unaffected.

## Changes

- **New `SidebarUsage.svelte`**, at the bottom of both sidebar surfaces above the Windmill mark: a progress ring in the collapsed rail, a labelled bar when expanded, and a modal breaking down every quota. Clicking opens the modal; it does not navigate.
  - Free tier: meters the per-user and per-workspace 1,000-execution caps and shows whichever is proportionally closest to its cap. Reaching a cap is red — jobs stop.
  - Team/Enterprise: meters workspace usage against the executions the workspace's seats already include (seats × 10,000). Going over is yellow, never red — it adds a billed seat rather than stopping jobs.
  - Visible to every non-operator member. Only the "See plans" / "See billing" button is admin-gated.
- **Execution hint on the existing free-tier counters** in the workspace, user and settings menus, so the number is explained where it already appears.
- **`Item.tooltip` was inert on disabled dropdown rows**: `DropdownSubmenuItem` rendered the ⓘ inside the disabled `<button>`, which swallows hover, and the row's own `title` shadowed any wrapper title. Both renderers now fall back to a wrapper `title`, the way `DropdownV2Inner` already intended. This also fixes the pre-existing inert ⓘ on FlowBuilder's disabled "Diff" item.
- **`isPremiumStore` is now `boolean | undefined`**, cleared while the active workspace's tier is resolving. It previously held the previous workspace's tier across a switch — invisible while it only gated affordances, wrong once a surface renders a number from it. Truthiness checks read `undefined` exactly like the old `false` default, so no consumer changes.

## Data access

The paid-plan numbers deliberately avoid `premium_info`: it calls `require_admin` and its route only exists when `STRIPE_KEY` is set, so non-admin members would see nothing. `getWorkspaceUsage` reads the same `usage` row with no admin gate and is already loaded into `workspaceUsageStore`; `list_users` is member-open (admin only for the cloud `demo` workspace) and gives the seat count. No new data is exposed — workspace usage was already shown to all members in the workspace menu, and `/users/list` was already reachable.

This does add one `listUsers` call per page load on paid cloud workspaces. Happy to defer it until the modal opens if that's the wrong trade.

## Proposed follow-up (not in this PR)

`feature_usage` telemetry on the modal open, per `AGENTS.md`. Left out so it's a deliberate call — say the word and I'll add it along with the backend allowlist entry.

## Screenshots

Free tier, expanded sidebar — the meter above the Windmill mark:

rail-expanded

Collapsed rail — progress ring:

rail-collapsed

Modal, free tier — both quotas, hard cap:

modal-free

Modal, paid plan — usage against included executions:

modal-paid

## Test plan

Everything below is cloud-only; a dev instance renders nothing unless `isCloudHosted()` is stubbed.

- [ ] Free workspace: the rail shows whichever of the two counters is proportionally higher; the modal lists both bars.
- [ ] Free `demo` workspace: only the user counter appears.
- [ ] Paid workspace: the bar reads against seats × 10,000 and the seat count matches `/workspace_settings?tab=premium`.
- [ ] Non-admin sees "Ask a workspace admin to change the plan" and no button; admin sees "See plans" / "See billing".
- [ ] Switch between a paid and a free workspace repeatedly — the cap must always match the workspace you are on, never flash the other tier's number.
- [ ] Collapsed rail: the ring fills proportionally and its tooltip reads correctly.
- [ ] Settings → user-email submenu on a free cloud workspace: hovering the greyed "N/1000 user execs" row shows the execution hint.
- [ ] Self-hosted: the sidebar bottom is pixel-identical to main.


Guilhem-lm:
Addressed the three P1s in `3343687985`:

- **Fork seats** — seats now come from the billing root (walked via `findWorkspaceAncestors` to a parentless workspace); when the root isn't visible from the fork, or `listUsers` on it 403s for a fork-only member, the paid meter stays hidden rather than showing a fork-sized cap. Verified on `wm-fork-pleasurable-fork` (1 member, root `ai-meta-e2e` has 2): the cap reads 20,000, not 10,000.
- **Tier cleared too late** — `isPremiumStore.set(undefined)` moved above the first await in `updateUserStore`, so it no longer survives the `getUserExt` round-trip.
- **Stale workspace usage** — `loadUsage()` now clears on a workspace change and drops a response that lost the race, matching what the tier got.

Also, from the Claude and Pi rounds: `!$isPremiumStore` read an unknown tier as free, which flashed the free-tier counters during a paid→paid switch. The three counter sites test `=== false` now, so they wait for a known tier.

On finding 4 (use the design-system `Button`) — I'd like to push back. The sidebar's own row primitive, `MenuButton.svelte`, renders a raw `<button>`/`<a>` rather than `Button`, and every neighbouring row in this rail goes through it; a `Button` here would be the odd one out visually. `Button` also has no `aria-label` prop and no rest-spread, so it can't carry the accessible name the Claude round asked for on the collapsed ring — adopting it would trade one finding for another. Happy to add an `aria-label` prop to `Button` and switch if you'd rather have the consistency.

Guilhem-lm:
Addressed round 2 in `9668d60e49`.

**P1 — unresolved usage read as zero.** Fixed at the source: `usageStore` / `workspaceUsageStore` are `number | undefined`, `loadUsage()` assigns each endpoint separately with its own catch (so one failing leaves the other's number intact, which `Promise.all` had regressed), and a quota is listed only once its own usage, tier and cap are known. The legacy counters render an em dash instead of a fabricated 0.

Verified with the endpoints aborted — note the workspace path is `/api/w/{ws}/workspaces/usage`, not `/api/w/{ws}/usage`:
- both usage endpoints aborted → meter **hidden** (previously `0/1,000`)
- workspace usage aborted, user usage OK → falls back to the user quota only, workspace quota dropped
- unblocked → meter returns

**P2 (Codex) — comment length.** `billingRoot`'s rationale condensed to 4 lines.

**P2 (Pi) — fork-affordance flicker.** The three gates now use `$isPremiumStore !== false`, so an unknown tier holds the affordance rather than hiding it. Agreed this was PR-introduced by the clear.

@claude — thanks for the correction on `Button`: it does have a rest-spread, so `aria-label` would pass through and that half of my reasoning was wrong. The consistency argument is the one that holds.

Guilhem-lm:
Round 3 nits addressed in `0b67acfbfc`.

**@claude — the `text/plain` catch was exact.** Confirmed against the live endpoint: `GET /api/w/admins/workspaces/usage` → `content-type: text/plain; charset=utf-8`, body `13500`. With a 13,500 usage planted and a paid tier, the bar read `13500/20,000` before and `13,500 of 20,000` after. My own verification had missed it because I was assigning fake numbers directly into the stores, so the value was a real `number` and `fmt()` worked; the free tier hides it since the cap is 1,000. Fixed with `Number(usage)` at both `.then`s as suggested.

**@codex — on the fork gates**, you and Claude landed on opposite defaults for the same three lines, so I split the difference along the axis that actually matters. Your strongest point was the failure case: `getIsPremium` is best-effort, so a rejection left the tier unknown *for the session*, and `!== false` kept the fork controls enabled on a free workspace indefinitely — clicking through to a backend rejection. That's now closed: the catch falls back to `false`, so a failed fetch reads as free tier.

I kept `!== false` for the *pending* window. That window is one round-trip, it restores the pre-PR behavior on a paid→paid switch (the flicker Pi flagged, which this PR introduced by clearing the store), and retracting a button someone is about to click is worse than briefly offering one. If you'd still rather have it hidden while pending, the workspace-keyed tier cache you suggested is the version I'd write — happy to do it, it just wants its own commit.

Everything else from the round is unchanged.

Guilhem-lm:
Round 4 addressed in `8a7368dc38`.

**@claude / @codex — you were both right, and my round-3 fix was the wrong shape.** Collapsing an unknown tier to `false` fixed the gates by lying to the meter: on a paid workspace whose `getIsPremium` rejects, real five-figure usage rendered against the 1,000 hard cap, red, under "jobs stop running for the rest of the month". Same class of defect as the two rounds before it, moved from the usage half to the tier half.

Fixed by separating the two consumers instead of picking one default for both. `isPremiumStore` never asserts a tier it doesn't know; a failure sets `premiumFetchFailed`, and affordances read a derived `maybePremium` (`premium !== false && !failed`). So:

| | meter | fork affordances |
|---|---|---|
| pending | hidden | held (no paid→paid flicker) |
| failed | hidden | hidden (fails closed) |
| free | free caps | hidden |
| paid | seat caps | shown |

Verified with `**/is_premium**` aborted and 13,500 usage planted on the workspace: meter **hidden**, no red bar, fork affordance **hidden**. Both halves of the previous trade, without the trade. (Also confirmed `is_premium` itself returns 200 `false` normally, so the ordinary path is untouched.)

**Stale seat cap** — no shared invalidation exists today; `WorkspaceUserSettings` refetches into a local list. Rather than wire a cross-cutting store for a background meter, the cap re-resolves when the modal opens, which is when the number is being read rather than glanced at. The rail figure can still lag a membership change until then, or until a reload or workspace switch. Happy to wire real invalidation if you'd rather — it wants its own commit and touches user management.

Guilhem-lm:
Round 5 addressed in `55a2ef2524`. All three of you found the same gap from different sides, so I fixed the cause rather than adding a third refresh site.

**The cause:** the counters were written in exactly one place — `loadUsage()` in the root layout, reached only on a workspace change. That's why @claude's finding and @codex's are the same bug: the modal could re-resolve the cap but not the count (the layout owns it), and the rail could never re-resolve either.

`loadUsage()` moved to `lib/usage.ts` as `refreshUsage()`, next to the stores it writes. The layout calls it on workspace change as before; the meter now calls it on modal open, so both numbers refresh — @claude, your point that the staleness landed on the wrong side is right, the numerator is what moves.

**Membership invalidation** (@codex's P1) — `workspaceMembershipVersion` in `stores.ts`, bumped in `WorkspaceUserSettings.listUsers()`, which every mutation on that page already routes through. `SidebarUsage` re-resolves seats on it without clearing, so the bar doesn't blank. Neither side owns the other and there's no polling.

Verified live on a paid-stubbed `admins`: toggling a member Developer→Operator fired **two** `users/list` calls — the settings page's own refetch and the sidebar's — where before there was one. The displayed cap correctly stayed 20,000, since 1 developer + 1 operator still rounds to 2 seats; roles restored afterwards.

That leaves the rail lagging only for a mutation made outside the app entirely (CLI, another session), which reload/switch already covers.

Guilhem-lm:
Round 6 addressed in `b2cc43d6e2`.

**@codex — both races are real and both are mine.** The workspace id doesn't order two requests for the *same* workspace, and this PR is what created the second in-flight request on each path: usage through A→B→A or a modal-open refresh landing on one already running; seats through a membership bump arriving mid-request. Each refresh now takes a generation and writes only if it's still the newest — `generation` in `usage.ts`, `seatsGeneration` in `SidebarUsage`.

**@claude — right that the doc claimed more than the code did**, and the fix was worth more than a reword. `listUsers()` now bumps only on an observed change to the member set, never on the first read (which has nothing to compare against), and the store comment says "seen to have changed" rather than "after a mutation".

Measured on a paid-stubbed `admins`, counting `users/list` requests:

| | before | after |
|---|---|---|
| opening the users tab | 5 | **4** |
| toggling a member's role | 2 | **2** |

The extra request on a plain read is gone; the invalidation still fires on a real change. Roles restored afterwards.

Guilhem-lm:
Round 7 addressed in `c7658f3633`.

**@codex — the seat filter was a genuine correctness bug, and worth spelling out which way it errs.** `count_paid_seats` (`windmill-common/src/workspaces.rs:773`) counts `FILTER (WHERE ... NOT disabled AND NOT is_service_account)`; `loadSeats` counted every row. So the cap was *too large* — it conceals a real overage rather than inventing one, which is the more expensive direction to be wrong in. Now filtered on the same two fields, and both are in the membership signature so enable/disable re-resolves the cap.

Verified on `admins` with a service account added: 3 `usr` rows, cap reads **20,000** (2 billable developers). Unfiltered it would have read 30,000. Row removed afterwards.

Worth flagging separately: **`PremiumInfo.svelte` has this same bug** — its `developerNb`/`operatorNb` come from an unfiltered `listUsers`, so the seat count an admin sees at `?tab=premium` is also inflated relative to what the backend bills. I've matched the backend rather than the billing page, so the two now disagree where service accounts or disabled members exist. Deliberately out of scope here; happy to file it or fix it in a follow-up.

**The tier generation guard** is the third instance of the same pattern — you were right that it was the one refresh still ordered by workspace id alone.

**@claude — the workspace-keyed signature is in.** You were right that the bump had moved rather than gone: `users` survives a workspace switch on that page, so `previous` held A's members against B's list. The captured signature is now qualified by the workspace it came from, so a switch reads as a different workspace rather than a membership change, and the store comment's claim holds.

Guilhem-lm:
Round 8 nit fixed in `c8235ec457` — all three of you caught the same thing, and you were right that I'd made it worse.

@claude's phrasing is the precise one: I put the workspace *inside* the compared value (`${workspace}:${signature}`), so on a switch the two strings could never match and the version bumped **unconditionally** — where before the key it bumped only when the member lists genuinely differed. The workspace is the key now rather than part of the payload:

```ts
const previous = lastSeen?.workspace === workspace ? lastSeen.signature : undefined
```

A different workspace has nothing to compare against, so it re-baselines silently, which is the same `previous === undefined` branch as a first read.

Measured after the change on a paid-stubbed `admins`: opening the users tab still costs **4** `users/list` calls (no sidebar refetch), a role change still costs **2** (the invalidation fires). Roles restored. I couldn't drive an in-app workspace switch through the melt picker programmatically, so that path is verified by reading rather than measuring — it takes the identical `previous === undefined` branch as the mount case above.

This was a nit-only round, so no new round for this fix.

Guilhem-lm:
Pushed `da1d2a2609`, a refactor rather than another fix: the usage and tier fetches now sit in `resource` from runed, which `frontend/AGENTS.md` prescribes for async data and ~80 files here already use.

Looking back at this branch's findings, most were one bug repeated. Each of these values belongs to a workspace but lived in a bare store, so every writer and reader had to re-derive "does this still describe what I'm rendering?" by hand — and the sites that forgot produced three stale-workspace overwrites, three A→B→A races, and two placeholders (`0` executions, `false` tier) that rendered as real data because an in-band value stood in for "not known". @claude flagged exactly this in the pre-PR review, quoting the AGENTS.md line for `loadSeats`, and I hand-rolled a token guard instead. Six rounds re-deriving what the primitive gives for free.

`resource` supplies all three as behaviour: superseded fetches discarded, value reset when its key changes, loading and error as states rather than magic values. Seats key on `{ billingRoot, membershipVersion }`, so a switch and an added member both re-resolve.

Removed: `generation`, `seatsGeneration`, `premiumFetchGeneration`, `usageFetchedFor`, `premiumFetchedFor`, and the clear-and-compare around each fetch. What's left is one publish site asserting the value still carries the active workspace. Net 146 insertions / 139 deletions, and the deletions are the parts that kept going wrong.

Re-verified every scenario from the earlier rounds rather than trusting the type-check — all on a cloud-stubbed instance:

| scenario | result |
|---|---|
| free tier | `Your executions this month: 0 of 1,000` |
| paid root, 13,500 planted | `13,500 of 20,000` (text/plain coercion intact) |
| paid fork | `0 of 20,000` — still the root's seats, not the fork's 1 |
| both usage endpoints abort | meter hidden, no fabricated `0` |
| tier aborts | meter hidden, fork affordance hidden (fails closed) |
| role change | 2 `users/list` calls — invalidation still fires |

Roles and planted rows restored afterwards.

Worth noting for whoever picks up the follow-up: `grep -rn 'Generation'` finds the same hand-rolled pattern in `S3FilePickerInner`, `ResourceVersionHistory`, `workspaceComparison`, `pipelineHistory` and others. Not touching them here.

Guilhem-lm:
Fixed in `7099685117`. **@pi — you were right and my claim was false.** I read `runed@0.36.0`'s `resource` rather than take it on trust, and `baseFetcher` does `current = result` unconditionally after the await; the only ordering is an `AbortController` whose signal the generated client cannot consume. So the commit message, the module docstring and my PR comment all asserted a property the library does not have. Corrected in the code and the commit.

**@claude — `loading` was doing the job it can't do**, exactly as you said, and the click path is worse than a blank bar: both the rail meter and the `<Modal>` sit inside `{#if isCloudHosted() && tightest}`, so opening the modal unmounted the modal.

Both come down to one thing: I removed the generation guards on the strength of a property the primitive doesn't provide, then used `loading` as a stand-in for it. Values now carry the scope they describe, and `scopedValue` keeps the newest one matching the active scope — a superseded answer neither publishes nor erases, and a re-read leaves the display alone. Also fixed the account-wide counter keying on the workspace (@claude's P2) — its key is the account now, so a switch is not a key change.

Verified rather than argued, since I got this wrong once already:

- **The blanking test was validated against the bug before trusting it.** With the `loading` gate restored as a probe, sampling 60ms after the click reads `GONE`; without it, ten consecutive samples read `ok` and the modal stays mounted.
- `scopedValue` is extracted to `lib/utils/scopedValue.ts` with four unit tests pinning the ordering properties — superseded answer ignored, value held across a re-read, key change reads as unknown. This is the one piece of pure logic in the change and it's the piece five rounds kept breaking, so it earns a test.
- Paid root `13,500 of 20,000`, paid fork `0 of 20,000` (root's seats), free tier, both-usage-abort hidden, tier-abort hidden with affordances closed: all unchanged.

I could not stage a true in-page A→B→A through the melt workspace picker programmatically, so that specific interleaving is covered by the unit tests rather than the browser.

Guilhem-lm:
Fixed in `ab87668fad`. All three of you found the same gap and it's a real one: the tag carried *what* a value described but not *when* it was asked for, so two fetches for one scope were indistinguishable and the older won if it landed last.

@claude's framing of why the obvious fix is wrong is the useful part — widening the tag to the resource key (`{root, version}`) would close it by blanking the bar on every membership change, which is precisely what the previous commit removed. So the issue order travels alongside the scope rather than inside it: `tagged()` stamps each request as it's issued, and only a strictly newer answer for the current scope replaces the held one.

**My earlier tests passed while missing this**, because they only exercised cross-key cases. Two new ones cover it, and I checked they fail against the key-only guard before trusting them:

```
× ignores an answer overtaken by a newer one for the same key
× stamps issue order even when responses resolve inverted
Tests  2 failed | 4 passed (6)     ← key-only guard
Tests  6 passed (6)                ← with issue order
```

The second one drives `tagged()` with two promises resolved in inverted order, so it pins the stamp-at-issue property rather than just the comparison.

Also fixed the `seatsResource` comment (@claude's P2) — it still asserted the "superseded response is discarded" property I retracted two commits ago, and ran to 8 lines over the ≤4 rule.

Browser re-check unchanged: paid root `13,500 of 20,000`, paid fork `0 of 20,000`, and ten consecutive samples after clicking the meter show it and the modal staying mounted.

Guilhem-lm:
Round 10, addressing @codex. Claude and Pi came back good-to-merge; taking the three Codex findings in turn. Fixed in `1514b2cb38`.

**P1 #2 (late user list suppresses seat invalidation) — real, fixed.**

`lastSeen` was assigned unconditionally after the await, so a response for a workspace already left overwrote the baseline for the workspace on screen; the next real change there compared against a baseline never taken for it, re-baselined silently, and skipped the bump. The `users = await …` on the same line had the same hole in the other direction — an overtaken list could paint over a newer one. Both now go through one check, using the issue order the branch already established:

```ts
const seq = ++usersIssued
const list = await UserService.listUsers({ workspace })
if (lastSeen && seq < lastSeen.seq) return
users = list
lastSeen = { workspace, signature, seq }
```

Per-workspace signatures would have fixed the baseline but not the list assignment, so this covers both.

**P1 #1 (superseded results publish before the newest settles) — not a defect; the proposed change would regress it.**

I checked runed's source rather than reasoning from the API: on error `baseFetcher` sets `error = e` and leaves `current` untouched. So what stands after a failed refresh is the last **successful** value. That is the intended policy, and it is the same one that made `loading` unusable as a gate — blanking a meter because a refresh failed is the defect the earlier rounds were spent removing.

The seat example doesn't separate the two policies. Removing a member bumps `workspaceMembershipVersion`, so request 1 is the *pre-removal* read and request 2 the post-removal one. Under today's code request 1 is accepted and, when request 2 fails, the pre-removal cap stands. Under "track the latest issued sequence" request 1 is rejected and `held` keeps the value from before it — which is also pre-removal. Same number on screen either way, so the change buys nothing there.

Where it does differ is first load: request 1 succeeds, request 2 fails, and the meter renders nothing despite a good number having arrived. That's strictly worse than showing it.

The policy is now stated where someone would undo it (`scopedValue.ts:20-24`). No test: a rejected fetch never produces a `Tagged`, so `scopedValue` is never called with one and the suggested case would assert nothing at that level.

**P2 #3 (overlong comments) — trimmed.**

Both `scopedValue` docstrings and the `listUsers` rationale are now four lines each.

Guilhem-lm:
Round 11: Claude and Pi good-to-merge, both confirming the `listUsers` ordering fix and the P1 #1 pushback independently. @codex's remaining P2 is real and fixed in `18a6c36252`.

**P2 (unresolved tier shown as "Premium plan") — fixed.**

My own regression: widening `isPremiumStore` to `boolean | undefined` left the `{:else}` catching the unresolved state. Both branches under that block assert a plan, so the whole block is now gated on the tier being known — which also stops the bordered divider from rendering empty while it resolves:

```svelte
{#if isCloudHosted() && $isPremiumStore !== undefined}
```

Verified on the running instance with the tier stubbed to resolve slowly, measuring from navigation start rather than from the tool call (the first attempt read 300ms that were really 5s):

| tier state | menu open | "Premium plan" | free counters |
| --- | --- | --- | --- |
| unresolved (5.2s, stub fires at 8s) | ✅ | ❌ | ❌ |
| resolved `false` | ✅ | ❌ | ✅ |
| resolved `true` | ✅ | ✅ | ❌ |

And the control that makes those readings mean something — reverting just the guard reproduces the bug: "Premium plan" at 304ms on a workspace whose tier resolves to `false`.

Grepped the other four `isPremiumStore` consumers while I was there: `WorkspaceMenu:425` and `SettingsMenu:236` test `=== false`, and `SidebarUsage`'s truthiness branches sit behind an explicit `=== undefined` guard. `UserMenu` was the only one reading `undefined` as paid.

Guilhem-lm:
Round 12: Pi good-to-merge. On @codex's two findings — the test is added in `86e3894615`, the behaviour is unchanged, and here is why.

**P1 (A→B→A accepts the original response) — working as intended; test added so it's pinned rather than argued.**

The bypass fires only when `held` describes a *different* scope, i.e. when there is no value for the current scope to order against. Walking it: A's read (seq 1) is in flight, we leave for A→B, B (seq 2) is held, we return to A. The late A answer is then the only value that describes A, so it stands. The read issued on returning (seq 3) supersedes it on `seq`, and once that has landed the late answer is rejected — `held.key === key` and `1 > 3` is false. There is no ordering where a lower-seq answer displaces a higher-seq one for the same scope.

"Overwrites B" isn't the failure it looks like: `held` is a single slot for the scope on screen, and once the key is `a` nothing renders B's value regardless of what occupies the slot.

The remaining case — fresh A fails, late A stands — is the last-successful-value policy from round 11, which you and both other reviewers settled: rejecting the late answer leaves the meter blank while a perfectly good number for the current scope is in hand.

New test (`scopedValue.test.ts:41-50`), exactly your scenario:

```ts
expect(held('b', { key: 'b', seq: 2, value: 2 })).toBe(2)
expect(held('a', { key: 'a', seq: 1, value: 10 })).toBe(10)   // only value describing A
expect(held('a', { key: 'a', seq: 3, value: 30 })).toBe(30)   // returning read supersedes
expect(held('a', { key: 'a', seq: 1, value: 10 })).toBe(30)   // and it cannot come back
```

7/7 passing. Under "preserve ordering across scope transitions" the second line would be `undefined` — a blank meter — which is the behaviour I'm declining, now visible in the test rather than in a thread.

**P2 (raw button) — keeping it, matching the rail it sits in.**

The sidebar rail is built from raw buttons, not design-system ones: `SidebarContent.svelte:563` (`hover:bg-surface-hover`, `rounded`, custom `-mx-1` chrome), `SidebarContent.svelte:699`, `DarkModeToggle.svelte:25`, and two in `WorkspaceMenu.svelte`. The meter is the same species — a full-width rail row wrapping a progress bar and a two-column label — and `Button`'s padding and label layout would fight both the collapsed 16px ring and the expanded bar. AGENTS.md's "follow established patterns in the codebase" points the same way here as its component rule points the other; I've taken the local one. I did add the explicit `type="button"` that `SidebarContent:564` uses.

Keyboard activation comes free with the native element: it's focusable and Enter/Space fire `onclick`, in both rail states. Mouse activation opening the modal was verified in an earlier round.

Guilhem-lm:
Round 13: Claude and Pi good-to-merge, both re-checking the round-12 ordering test against the guard rather than reading it for shape. @codex's P2 is right and is fixed in `31b75b3e9f`.

**P2 (deprecated `Button.size`) — fixed.** `unifiedSize="sm"` instead of `size="sm"`. Checked the rendered result rather than just the prop: the plan button comes out `h-7` / `text-xs`, the same height and weight as the modal's own Close button sitting next to it. `SidebarUsage` has no other `size=` left.

That was the only deprecated-prop instance in the diff.

Guilhem-lm:
Round 14: Codex and Pi both good-to-merge, no findings. @claude's P2 is right and my claim was wrong — fixed in `af07d83560`.

**P2 (plan button vs the modal's Cancel) — confirmed by measurement, fixed.**

I asserted parity from the plan button's own classes without measuring the neighbour. Measured both with the modal open, which settles it:

| | before (`unifiedSize="sm"`) | after (`unifiedSize="md"`) | Cancel |
| --- | --- | --- | --- |
| height | 28px | **32px** | 32px |
| classes | `px-2 h-7` | `px-4 h-8` | `px-3 py-[7px]` |
| top | — | 540 | 540 |

Your class-level trace was exactly right: `h-7` against `px-3 py-[7px]`'s 16+14+2. `md` is the unified size that lands on 32px, so the pair matches without putting a deprecated prop back, and without migrating `Modal`'s Cancel — which, as you say, is every `kind="button"` modal in the app and belongs in its own change. Residual difference is font weight (500 vs Cancel's 600), inherent to the two prop systems and not worth chasing.

Also taking the correction on the AGENTS.md citation: the deprecated-prop sentence I quoted across rounds isn't in `frontend/AGENTS.md`. The rule comes from the `@deprecated` JSDoc on the props and the `svelte-frontend` skill's prop list. Noted so it doesn't get quoted at the wrong file again.

Guilhem-lm:
Three follow-ups landed since round 15 (`1251a63e4a`, `d115923042`). This is now an **`[ee]` PR** — companion: windmill-labs/windmill-ee-private#735.

**1. `feature_usage` telemetry — added** (it was a deferred proposal in the PR body).

`usage_meter` / `opened`, keyed by plan tier plus the quota that was tightest: `free:user`, `free:workspace`, `paid:workspace`. The full set is a type beside the call site so the vocabulary reads in one place. No identifying data — no workspace, path, or user.

Both halves of the gate verified, since the failure mode here is silent:

- The browser posts `{"feature":"usage_meter","kind":"opened","key":"free:user","entity_id":"","value":1}` to `/w/<ws>/workspaces/log_feature_usage`.
- Against the **running EE image**, `SELECT count(*) FROM feature_usage WHERE feature = 'usage_meter'` is **0** — its registry predates the entry, so the post 204s and records nothing. That is the trap `docs/feature-telemetry.md` warns about, reproduced rather than assumed.
- With the entry in `FEATURE_USAGE_KINDS`, `is_recordable_event("usage_meter", "opened", ...)` accepts both keys and still rejects an unregistered kind (checked by compiling `windmill-common --features private` against the symlinked EE file).

Disclosure updated in both `InstanceSettings.svelte` passages: the plan tier and quota are now named alongside the AI model identifiers and hub script names.

**2. `PremiumInfo.svelte` seats — fixed here**, as agreed rather than deferred. It computed seats from an unfiltered `listUsers`, counting disabled members and service accounts that `count_paid_seats` (`windmill-common/src/workspaces.rs:770`) does not bill, so the billing page overstated seats against the invoice. Same filter as the sidebar cap now.

Note the diff is deliberately six lines: the repo's formatter reflows this whole file on any edit (it predates the current Prettier config), so I applied the hunk without the ~100 lines of unrelated churn.

**Not exercised:** the `PremiumInfo` page itself. `getPremiumInfo` requires admin *and* only mounts when `STRIPE_KEY` is set, and the dev instance has no Stripe key — so the arithmetic change is unrun in a browser. It mirrors `fetchSeats` in `SidebarUsage`, which is exercised.

**3. `ee-repo-ref.txt`** points at the EE commit. `write_latest_ee_ref.sh` printed "Directory not found" and left the old hash in place, so it is written by hand — flagging it since the doc says to check.

Guilhem-lm:
Round 16: Codex and Pi good-to-merge. @claude's P2 and its caveat are both right — fixed in `4af3debc34`, plus one thing that turned out to matter more than the nit.

**P2 (the telemetry doc's own inventory) — fixed.** `docs/feature-telemetry.md` now reads 21 actions across nine features, with `usage_meter` in the list. You're right that the sentence is load-bearing — it's what the doc uses to argue most of the product is uninstrumented — and the file itself notes this copy has drifted once already.

**The seat framing — taking the correction.** I wrote that the billing page "overstated seats against the invoice". The repo can't establish that: `count_paid_seats`'s own doc says it counts provisioned members rather than the active-user population billing meters, and the Stripe subscription quantity isn't computed here or in `stripe_ee.rs`. What the change actually establishes is that both in-repo seat counters now agree on the same basis — which was the round-7 divergence — and that `count_offline_potential_seats` states that basis mirrors which accounts billing charges for. I'm not carrying the stronger claim forward.

**The race you flagged as pre-existing — fixed, because it defeats the seat fix.** `loadPremiumInfo` reads `users` after its own await and nothing recomputes the rows when the list lands, so whenever `premium_info` won the race the page rendered 0 developers / 0 operators / 0 seats and kept them. A more accurate filter over a list that hasn't arrived is still zero. Now:

```js
listUsers().catch(console.warn).then(loadPremiumInfo)
```

`.catch` rather than a bare `.then` chain on purpose: the two calls used to be independent, so a failing `/users/list` cost only the seat rows, and sequencing them naively would have made it cost the whole billing page.

**EE ref confirmed**, since you flagged it as resting on my word: `c5f415bb` is the head of `glm/improve-execution-display` on windmill-ee-private, and EE `main` is `d30af67` — so the ref points at the branch, not main. That's the `docs/enterprise.md` step-5 check for the `write_latest_ee_ref.sh` fallback.

**Still unexercised:** the `PremiumInfo` page. Your suggested substitute — a workspace with a disabled member or service account, watching the Developers/Operators rows drop — needs the page to render at all, and `premium_info` 404s on this instance because no `STRIPE_KEY` is set. So the seat arithmetic and the sequencing fix are both unrun in a browser, and I'd rather say that twice than imply otherwise.

Guilhem-lm:
Round 17: Claude and Pi good-to-merge. @codex's P2 is right — fixed in `348f15194f`.

I retracted the invoice-parity claim in the thread last round but left it standing in the code comment, which is the copy that outlives the thread. Now:

```
// Same basis as the backend's `count_paid_seats`, which excludes disabled members
// and service accounts: counting them here would put this page at a different
// seat count from the rest of the product.
```

Scoped to what the repo can show — agreement with `count_paid_seats` — rather than to the invoice, which nothing here computes.

Left the sibling comment in `SidebarUsage.svelte:53-55` as it stands: it says those accounts "are not billed", which `count_offline_potential_seats` states in as many words ("mirroring which accounts billing charges for"), and it makes no claim about seat quantities on an invoice.

Guilhem-lm:
Merged `main` into the branch — `7b57c356f1`, `mergeable=MERGEABLE`.

**The conflict was `backend/ee-repo-ref.txt`, and neither side was the answer.** Main's ref had advanced to `d30af67`, four EE commits ahead of this branch's base — including the `require_super_admin` fix (GHSA-hfh4-cx4h-3fcr). Keeping my hash would have rolled the EE side back past that; taking main's would have dropped the `usage_meter` registration and left the telemetry silently unrecorded, which is the exact failure this PR spent a round demonstrating.

So the EE branch is rebased onto EE `main` and the ref points at the result:

- windmill-labs/windmill-ee-private#735: `c5f415b` → **`bbc48fa`**, carrying both the security fix and `("usage_meter", "opened")` — confirmed the entry survived the rebase.
- `ee-repo-ref.txt` → `bbc48fae6b73b6d72fe2e125e6003794a4ece167`.

`check:fast` and the seven `scopedValue` tests pass on the merge commit. Main's rewrite of `isCloudHosted()` (guarding `globalThis.window?.location?.hostname`, since `BROWSER` alone is true in SSR bundles and vitest's server project) merged cleanly and doesn't change the meter's gating.

**One retraction to retract.** @claude flagged in round 14 that the deprecated-prop sentence I'd cited as `frontend/AGENTS.md` wasn't in that file, and I accepted it. It is in `main` — added by `ac27d0200d` (#10772), which landed while this branch was 15 commits behind. The citation was correct against `main`, just not against the stale copy either of us was reading. The fix it prompted (`unifiedSize`) was right regardless.

PR #10766, opened 2026-08-19
fix(agents): unbreak the scratch-dir permission guards on macOS

Guilhem-lm:
## Summary

The two `PreToolUse` guards under `.claude/hooks/` — `guard-rm-outside-tmp.sh` (`rm`) and
`allow-fileops-in-tmp.sh` (`mkdir`/`cp`/`mv`/`touch`/`chmod`/`tar`/`unzip`) — auto-allow an
operation whose every operand it can prove lands in a safe root, and `ask` otherwise. On macOS
they proved nothing at all: **every** operation prompted, including the `rm /tmp/x` the design
exists to wave through.

Three separate macOS incompatibilities, all in the path-resolution layer:

- Both guards canonicalized with GNU `realpath -m`. BSD realpath rejects `-m` outright
  (`realpath: illegal option -- m`), so every operand fell to `defer "cannot resolve"`.
- `/tmp` is a symlink to `/private/tmp`, so a canonicalized scratch path never matched the
  literal `/tmp` the `tmp` class was anchored to.
- `${ops[-1]}` is a fatal error under the bash 3.2 macOS ships, aborting the guard mid-decision.

`test-hooks.sh` went from **26 failures to ALL PASS** on macOS. Linux is unaffected by all three:
`realpath -m` is still tried first, `TMP_ROOT` resolves to plain `/tmp`, and the index arithmetic
is equivalent on bash 4+.

Two behaviour changes ride along, both motivated by what actually prompts during a session:

- **A `mcp-cache` class** for the Playwright MCP browser caches (`~/Library/Caches/ms-playwright`
  and `ms-playwright-mcp`, `~/.cache/…` on Linux). Nothing prunes them — they reached 10G on one
  machine — and clearing one costs a browser re-download and nothing else. They sit outside every
  checkout, where no other class reaches them.
- **`~/` and `$HOME/` expansion**, so the natural spelling of a path outside every checkout can be
  proved rather than deferring on the character whitelist.

## Changes

- `lib-guarded-verb.sh`: add `canon_path`, which tries `realpath -m` and falls back to python3's
  `os.path.realpath` (same non-strict, symlink-resolving semantics). Trying rather than probing
  keeps the cost off the Bash calls that never reach a path check.
- `lib-guarded-verb.sh`: anchor the classes to resolved `TMP_ROOT` / `HOME_ROOT`, both computed
  with `cd -P` so sourcing the library spawns no process.
- `lib-guarded-verb.sh`: add the `mcp-cache` class to `path_class`, matching the root itself so
  `rm -rf` on the whole directory works.
- `lib-guarded-verb.sh`: add `expand_home_prefix`, applied in both guards and in `apply_cd`.
  `~user/`, `"$HOME/x"` and every other `$VAR` still fail the charset check by construction.
- `lib-guarded-verb.sh`: match the exclusion list (`.git`, `.claude`, `.env`, `secrets/`, `*.pem`,
  `*.key`, `credentials.json`, `.secret*`) against a lowercased copy of the path. APFS is
  case-insensitive by default, so `.GIT` and `.git` are one directory — and that list is the
  entire protection for those paths, since the `repo:` class already allows deletes everywhere
  else in a checkout. Caught by review; it fails open exactly on the platform this PR activates.
- `guard-rm-outside-tmp.sh`: globs auto-allow in `mcp-cache` as well as `tmp`. Both are roots
  where `rm -rf <root>` is already allowed wholesale, so matching inside one grants nothing more.
  In a checkout literal operands are still required — an expansion there could reach `.git` or a
  dotfile the literal checks never see.
- `allow-fileops-in-tmp.sh`: replace `${ops[-1]}` with index arithmetic for bash 3.2.
- `test-hooks.sh`: rows for the new root, the home prefix on both guards, and a case-variant of
  each denied kind.
- `AGENTS.md`: update the scratch-policy paragraph — it told agents a `~` could never be proved
  and listed only two roots.

## Test plan

- [ ] `bash .claude/hooks/test-hooks.sh` → `ALL PASS` on macOS (bash 3.2 / BSD realpath / APFS)
- [ ] `bash .claude/hooks/test-hooks.sh` → `ALL PASS` on the Linux dev box. The suite branches on
      checkout kind (`ROOT_SOLO`/`ROOT_CHAINED`) but not on platform — `path_class` matches both the
      `Library/Caches` and `.cache` spellings everywhere, and the suite now has a row for each — so a
      Linux run still cannot exercise the BSD-realpath, bash-3.2 or APFS-case-fold branches
- [ ] In a live session `rm -rf ~/Library/Caches/ms-playwright-mcp` completes with no prompt, and
      `rm -rf ~/Documents` still prompts
- [ ] `rm -rf <repo>/.GIT` and `chmod -R 777 <repo>/.GIT` still prompt (the case-fold regression)

No frontend surface, so no screenshots. No `*_ee.rs` touched, so no companion PR.

### Reviews run before opening

- `local-review` (cold branch-diff-reviewer): found the case-insensitivity P0, fixed in ee77cd9351.
- `local-review-codex`: **could not run** — the `codex` CLI hit its account usage limit
  (resets Aug 20 08:53). CI's Codex reviewer runs with its own credentials and is unaffected.

### Round 1 (head ee77cd9351) — addressed in 5886959510

- Codex P1 / Pi P2: `ms-playwright-mcp*` matched siblings such as `ms-playwright-mcp-backup`,
  because a case pattern's `*` spans the `-` too. Each root is now named exactly and again with
  `/*`, mirroring the `ms-playwright` entries. Negative rows added on both guards.
- Pi P2: `rel_operand` inspected the raw token, so a `~/`-spelled operand read as relative and
  spuriously deferred a `cp`/`mv` after a `cd`. Now tested against the expanded token.
- Claude P2: no row exercised the Linux `~/.cache/ms-playwright*` arm of the class. Added, and
  the test-plan claim about the suite branching on `.cache` vs `Library/Caches` is corrected above.

### DIFF
--- a/.claude/hooks/allow-fileops-in-tmp.sh
+++ b/.claude/hooks/allow-fileops-in-tmp.sh
@@ -244,7 +244,9 @@ check_fileops_segment() {
     [ -n "$seen_class" ] && [ "$cls" != "$seen_class" ] && defer "\`$t\` puts this $verb across two roots"
     seen_class="$cls"
     ops+=("${resolved#*$'\n'}")
-    case "$t" in /*) ;; *) rel_operand=1 ;; esac
+    # Against the expanded token, since `~/a` is cwd-independent and only reads as relative
+    # before `expand_home_prefix` has run.
+    case "$(expand_home_prefix "$t")" in /*) ;; *) rel_operand=1 ;; esac
     path_operand=1
   done
 
--- a/.claude/hooks/allow-fileops-in-tmp.sh
+++ b/.claude/hooks/allow-fileops-in-tmp.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 # PreToolUse allowance for scratch file ops: auto-allow `mkdir` / `cp` / `mv` / `touch` /
 # `chmod` whose every path operand resolves inside one of the roots `path_class` recognizes —
-# under /tmp, or inside a git working tree under $HOME — and `tar` / `unzip` confined to /tmp.
+# under /tmp, inside a git working tree under $HOME, or in an MCP browser cache — and
+# `tar` / `unzip` confined to /tmp.
 # Anything else makes no decision (exit 0) and falls back to the normal permission flow, except
 # for `mv` and `chmod`: those get an explicit `ask`, the only prompt they get (see
 # lib-guarded-verb.sh).
@@ -28,9 +29,10 @@
 # file there has never prompted, and moving or chmod-ing one is not the graver act.
 #
 # Deny-by-default tokenizing, in the same spirit as guard-rm-outside-tmp.sh: every path token
-# must consist only of alphanumerics and `. _ / -`. That set contains none of the characters
+# must consist only of alphanumerics and `. _ / -`, the one exception being the leading `~/` or
+# `$HOME/` that `expand_home_prefix` rewrites first. That set contains none of the characters
 # bash uses for quoting, expansion, or command separation ($ ` ~ { } ( ) ' " \ ; & | < >), nor
-# any glob character, so all of those forms fail by construction. `realpath -m` then resolves
+# any glob character, so all of those forms fail by construction. `canon_path` then resolves
 # `..` and existing symlinks, so `/tmp/link` pointing at /etc/passwd is caught.
 #
 # `tar` and `unzip` keep the stricter rule — /tmp only, and absolute operands only — because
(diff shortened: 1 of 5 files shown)


Guilhem-lm:
✅ Review round clean @ be8436f90083785db873ddae4958bf62f0b0dc2f — nit-only verdicts at 588695951086f9a09e2f1287b3ad1c7a26581e23 (codex: Good to merge, pi: Good to merge, claude: nits); nit addressed in be8436f90083785db873ddae4958bf62f0b0dc2f

PR #10767, opened 2026-08-19
fix: scope git-sync concurrency key per repository

rubenfiszel:
## Summary

Every git-sync push callback was pushed with the concurrency key `{workspace_id}:git_sync`
(concurrent limit 1, time window 0), so all repositories configured on a workspace shared one lane
even though they are distinct remotes that can never race on a ref.

Losing that race is expensive, because a concurrency-limited job is not woken when the slot frees.
`apply_concurrency_limit` re-queues it to a fixed estimate:

```
inc  = avg_duration(last 10 jobs on that key) + 100ms + time_window
next = min_started_at + (windows_full + 1) * inc
```

The average is pooled across every repo on the key, so a workspace with one large repo whose sync
takes minutes pushes the other repos' syncs minutes into the future, and the estimate does not
shrink when the first push finishes early. Observed shape: a small repo's sync landing 7+ minutes
after the deploy that triggered it, which reads as "the second repo never synced".

Promotion mode already keyed its *debounce* per repo (#9145), but its concurrency key did not, so
two promotion repos deploying the same object also shared a lane.

## Changes

- `concurrency_key_append` now carries the repo's `git_repo_resource_path` in every git-sync
  **push** path, giving `{workspace}:git_sync:{repo}` in workspace-wide mode and
  `{workspace}:git_sync:{repo}:{path_type}:{ref}` per promotion branch:
  - batch dispatcher, promotion bucket and workspace-wide arm
  - both env-label mirror pushes for a dev workspace in promotion mode (batch and single-object)
  - the legacy non-debounced per-item dispatch
  - the single-object dispatcher
- **Pulls deliberately stay on the workspace-wide lane.** A pull writes workspace objects, so the
  shared lane is what stops two repos applying creates/updates/deletes to the same scripts and
  flows concurrently. The PR-diff dry run stays there too, since the diff it computes reads a
  workspace that must not be mid-apply.
- One writer per repo is unchanged, and promotion buckets keep one writer per branch.
  `group_by_folder` and empty-path callbacks fall back to the repo lane, since the branch they
  push is not knowable at dispatch time.

- The lane id trims the `$res:` prefix, so a repository stored in both forms (the whole-config
  settings endpoint does not require it, unlike the per-repository one) shares one lane instead of
  racing itself.
- `windmill-queue/src/jobs.rs` caps the key at the column width, but measured the bare key while
  `resolve_concurrency_key` prepends `{workspace_id}/` on cloud builds. `concurrency_key.key` is
  `VARCHAR(255)` and its INSERT runs inside `push`, so an over-long key fails the push and the sync
  job is never created at all — a `tracing::error!` and nothing in the queue. Putting the repo in
  the key made that reachable, so the cap now reserves the prefix, as `resolve_debounce_key`
  already does.

EE companion PR: windmill-labs/windmill-ee-private#734

### Known trade-offs

- A push can now overlap a pull, where the shared lane previously serialized them. The push then
  commits a snapshot of a half-applied pull, which the next push settles. Serializing both would
  require pushes back on one workspace lane, which is the problem this PR fixes; a proper
  reader/writer gate is the real answer and is out of scope here.
- Lane identity is the repo *resource path*, not the resolved remote. Two resources aliasing the
  same URL and branch would now push concurrently and one could fail non-fast-forward; previously
  they serialized. Keying on the resolved URL was rejected because that URL can embed a token and
  concurrency keys appear in job logs. That configuration already double-pushes identical content,
  so rejecting duplicate URL+branch pairs at settings-save time is the better place to handle it.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test workspace_dependencies_git_sync --features enterprise,private`
      (8 passed, green on an immediate re-run)
- [x] New test fails with the EE change reverted
- [x] Long-repo-path test fails with the cap disabled: `value too long for type character
      varying(255)`, push aborted, zero callbacks queued
- [ ] Configure two repositories on a workspace in workspace-wide mode, deploy a script, and
      confirm both sync jobs run instead of one waiting out the other
- [ ] Enable auto-pull on two repositories, push to both remotes at once, and confirm the two pull
      jobs still serialize

The new test pins one concurrency lane per repo in both workspace-wide and promotion mode. It uses
repo paths carrying a random per-run id because concurrency settings are written through a
filesystem-backed cache keyed by the settings hash: a hash any earlier run inserted is assumed to
be in the database already, so under `#[sqlx::test]` (fresh database, shared cache) the row never
lands and the key cannot be read back.


rubenfiszel:
Nits from the round addressed in `067b20d`:

- **Prefix-reservation arithmetic untested** — extracted `git_sync_concurrency_key(workspace_id, suffix, reserved_prefix_len)` as a pure fn with unit tests in `windmill-queue`. Passing the reservation as a parameter makes it testable regardless of the `cloud` cfg: one test pins that a suffix which fits the bare 255 but not the prefixed one gets hashed, the other that distinct repos stay distinct after hashing.
- **No pull-lane regression test** — added `test_pull_stays_on_the_workspace_lane`, which calls `enqueue_git_pull_job` directly and asserts the stored key is exactly `test-workspace:git_sync`. That is the invariant the whole safety argument rests on, so it is worth pinning.
- **Test name covered half its assertions** — renamed to `test_two_repos_get_distinct_concurrency_lanes` and updated the doc comment to name both phases.
- **EE comment invariants unverifiable** — the EE diff lives in the companion PR (windmill-labs/windmill-ee-private#734); the reviewer's EE checkout is pinned at the base ref, so it cannot see it from here. The four comment sites flagged were all rewritten in that PR: `enqueue_git_pull_job`'s doc contract now states the workspace-wide lane and why, and the mirror/lane comments were updated with the change.

On the six-line comment at `jobs.rs`: it moved into the new fn's doc comment, which is the right home for it.

Not addressed, deliberately: the `$res:`/raw-path asymmetry between the debounce key and the concurrency key. As the review notes, that asymmetry is the safe direction (over-debouncing is a delay, under-serializing is a non-fast-forward), and changing the debounce key would alter behavior #9145 pinned.

PR #10768, opened 2026-08-19
fix: explain the 6-field cron format when a schedule is rejected

rubenfiszel:
## Summary

Pushing a schedule whose cron is a standard 5-field crontab string was rejected with a bare parser error and no indication of what Windmill expects:

```
Bad request: cron: 0 2 * * *
         ^
```

Windmill cron expressions carry a leading seconds field, so every 5-field crontab line fails, and neither parser says so: the `cron` crate (v1) just points a caret at the expression, and croner (v2) reports `Pattern must consist of five or six fields ... and optional second` while the API parses with seconds required, which contradicts the actual rule.

`ScheduleType::from_str` now appends the expected format to cron parse errors, and for a 5-field expression it also suggests the equivalent 6-field one:

```
Bad request: cron: 0 2 * * *
         ^

Windmill cron expressions have 6 fields and start with seconds: 'sec min hour day-of-month month day-of-week'. The 5-field crontab syntax is not accepted; prepend a seconds field, e.g. '0 0 2 * * *'.
```

This is the shared parser, so the hint reaches the CLI (`wmill schedule push`), the API, and the schedule preview alike.

## Changes

- `six_fields_hint` in `windmill-common/src/utils.rs`, appended to the parse error of both the v1 (`cron`) and v2 (croner) branches of `ScheduleType::from_str`.
- The example is only offered when the 5-field expression prefixed with a `0` seconds field actually parses, so an expression that is invalid for other reasons is not given a bogus rewrite. It is phrased as the mechanical prepend it is rather than as the same schedule.
- On v1 the example is withheld whenever the weekday field is restricted, because that is where v1 parts ways with crontab on two counts: it numbers weekdays from Sunday = 1 (`0 2 * * 1-5` prepended runs Sun-Thu, not Mon-Fri), and it intersects day-of-month with day-of-week where crontab unions them (`0 2 1 * MON` prepended runs only when the 1st is a Monday, verified against the preview endpoint). With the weekday unrestricted the remaining fields carry their crontab meaning, so that is the case where an expression can be handed back. croner reads both the crontab way, so v2 keeps the example throughout.
- The hint is suppressed on the seconds-optional parse (the worker re-parsing a stored schedule), where croner does accept 5 fields, and on expressions that already have 6 or more fields, so unrelated errors are unchanged.
- The probe that decides whether the example parses calls the two parsers directly instead of re-entering `from_str`, so a synthesized expression never produces an ERROR log line for a string nobody submitted. The croner flag setup both paths need is now one helper.

## Not covered here

The schedule editor discards the server's message for any cron 400 and renders a static `Invalid cron syntax` (`CronInput.svelte`). That surface can't show this hint anyway: `formatCron` pads any expression shorter than 6 fields with trailing `*`, so a pasted `0 2 * * *` is silently sent as `0 2 * * * *` (hourly at :02) rather than rejected. Both are pre-existing and want their own change.

## Test plan

- [x] `cargo test -p windmill-common --lib utils::tests` — the message reaches the caller for `None` / `v1` / `v2`, the example is withheld on v1 for both a numeric weekday and a weekday alongside a day-of-month while v2 keeps it, and an unrelated 6-field error stays untouched.
- [x] `POST /api/schedules/preview` on a dev instance with `0 2 * * *` (no `cron_version`, and `v2`), `0 2 * * 1`, `0 2 * * MON`, `0 2 1 * MON`, `0 2 1 * *`, a 3-field expression, garbage, and a valid `0 0 2 * * *`.
- [x] The two divergences the v1 rule guards against, confirmed against the preview endpoint: `0 0 12 * * 1-5` runs Sun-Thu on v1 and Mon-Fri on v2; `0 0 2 1 * MON` runs only when the 1st is a Monday on v1, and every Monday or 1st on v2.
- [x] The reported flow end to end: `wmill schedule push` of a YAML with `schedule: "0 2 * * *"` now prints the actionable error instead of the bare caret.

### DIFF
--- a/backend/windmill-common/src/utils.rs
+++ b/backend/windmill-common/src/utils.rs
@@ -909,6 +909,35 @@ pub enum ScheduleType {
     Cron(cron::Schedule),
 }
 
+/// Both cron parsers reject the standard 5-field crontab syntax without naming the missing
+/// leading seconds field, and croner even advertises five fields as valid while we parse
+/// with seconds required. Only that seconds-required parse is 6-field, so the hint stays
+/// off the seconds-optional path, where five fields really are accepted.
+fn six_fields_hint(schedule_str: &str, version: Option<&str>, seconds_required: bool) -> String {
+    let fields = schedule_str.split_whitespace().collect::<Vec<_>>();
+    if !seconds_required || fields.len() >= 6 {
+        return String::new();
+    }
+    let suggestion = if fields.len() == 5 {
+        // The suggestion has 6 fields, so parsing it can only recurse one level deep.
+        let with_seconds = format!("0 {}", fields.join(" "));
+        match ScheduleType::from_str(&with_seconds, version, seconds_required) {
+            Ok(_) => format!(
+                " The 5-field crontab syntax is not accepted, did you mean '{}'?",
+                with_seconds
+            ),
+            Err(_) => String::new(),
+        }
+    } else {
+        String::new()
+    };
+    format!(
+        "\nWindmill cron expressions have 6 fields and start with seconds: \
+         'sec min hour day-of-month month day-of-week'.{}",
+        suggestion
+    )
+}
+
 impl ScheduleType {
     pub fn find_next(
         &self,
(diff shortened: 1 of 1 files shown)


rubenfiszel:
✅ Review round clean @ f8e35179008190d9a41e7eac8f1940b32d51279f

PR #10769, opened 2026-08-19
feat(cli): deduplicate identical script lockfiles (dedupeLockfiles)

rubenfiszel:
## Summary

A workspace whose dependencies come from a `dependencies/<file>` resolves to the same lock for every script of that language, so the repo carries thousands of byte-identical `.script.lock` files. One dependency bump rewrites all of them, and every open branch then conflicts on all of them.

`dedupeLockfiles: true` in `wmill.yaml` (opt-in, default off) stores that lock once, in a file named after the dependency file it came from, referenced from each `.script.yaml` through the `!inline` indirection the metadata already uses:

```
dependencies/requirements.in         ->  locks/requirements.in.lock
dependencies/team_a.requirements.in  ->  locks/team_a.requirements.in.lock
dependencies/package.json            ->  locks/package.json.lock
```

A dependency bump becomes a one-file diff. Fixes WIN-1756.

## Design

**Identity comes from the dependency file** — never from the lock's content, and never from which scripts happen to be in view. That is what makes the pass stateless: a sync narrowed to a single script (which is what the git-sync deploy callback does, via per-item `extraIncludes`) computes the same name as a full one, because the lock that script holds *is* that dependency file's lock — the worker resolves it from the file alone and ignores the script body.

Two consequences worth knowing:

- **A lockfile lives exactly as long as its dependency file.** Asking that, rather than "does any script still read it", is what lets a narrowed sync leave the rest of the workspace alone.
- **Two cases keep a private `.script.lock`**: a script whose annotation is `extra_` or carries inline dependencies (its lock folds in its own imports), and one naming several dependency files at once (its lock is no single file's). Scripts with no workspace dependency file behind them are untouched.

The name is a suffix append rather than an extension swap, so the mapping is reversible: strip `.lock` and you have the dependency filename, which is how `isSharedLockPath` decides whether a file is Windmill's. `locks/vendor.lock` names no dependency file, so a repo that already keeps lockfiles there keeps them, and with the option off `locks/` is ignored entirely.

## Changes

- `cli/src/utils/lock_dedup.ts` (new): the pure planner (`computeSharedLockPlan` → `{writes, deletes}`, applied to a sync map or to the working tree).
- `compareDynFSElement` collapses the **remote** map in both directions, so local and remote stay comparable and neither pull nor push sees a phantom diff.
- `sync push` fans a shared-lock change out to the scripts that read it (nothing else queues them — their own metadata is identical on both sides), after `buildTracker` so `--auto-metadata` does not regenerate their locks.
- `updateScriptLock` writes a regenerated lock straight to the shared file, so a bump never lands a private copy that a later pass has to collapse.
- `generate-metadata` and `push --auto-metadata` run a whole-tree normalization pass, re-hashing what they rewrite (never a script whose generation failed) so nothing reads as stale next run; it also runs on the up-to-date path, which is the state a repo is in when the option is first turned on, and reports instead of writing under `--dry-run`.
- A script lockfile deletion is deferred to the end of a pull and skipped if the metadata on disk still reads it — a conflict resolved as "preserve local" must not be left with a dangling `!inline`.
- Workspace-dependency vocabulary (`workspaceDependenciesPathToLanguageAndFilename`, the annotation parser) moves to `script_common.ts`, beside the languages it describes; `metadata.ts` re-exports it.
- `wmill.yaml` option + regenerated `wmill.schema.json`; `wmill lint --locks-required` resolves a shared reference from the sync root.

Scope: standalone scripts (flat and `__mod` layouts). Flow and app inline-script locks use folder-relative references and are untouched; dbt is excluded. `locks/` is created lazily — a workspace with nothing to share never grows the directory.

## Test plan

Unit (`cli/test/lock_dedup_unit.test.ts`, `UNIT_ONLY=1 bun test test/*_unit*.test.ts` — 1085 pass):

- [x] the scripts of a dependency file collapse into its lockfile; each named set gets its own
- [x] a bump changes exactly one file; the pass is idempotent
- [x] one script in view reaches the same answer as the whole workspace
- [x] `extra_`/inline annotations and multi-file scripts keep a private lock; a stale committed lock is outvoted
- [x] a shared lockfile outlives its scripts but not its dependency file
- [x] `isSharedLockPath` claims only names a dependency file gives (`locks/vendor.lock` is not ours)
- [x] dotted script paths, module layout, `lock: ''`, languages that need no lock

Exercised against a running backend (workspace with two bun dependency sets, python scripts, one divergent):

- [x] `sync pull` writes one lockfile per dependency file and removes the duplicates; a second pull and a push both report 0 changes
- [x] bumping one dependency set touches that set's lockfile only — no `.script.yaml` changes
- [x] editing a shared lockfile and pushing deploys it to exactly the scripts that read it (verified on the server)
- [x] narrow pull (`--extra-includes`, the git-deploy shape) is a no-op and leaves the shared files intact
- [x] `generate-metadata` on an up-to-date tree converts it and is a no-op on the next run; `--dry-run` leaves every file byte-identical
- [x] `push --auto-metadata` with a bumped shared lock runs 0 dependency jobs
- [x] a stateful pull with a conflict resolved as "preserve local" keeps the lockfile that metadata still references
- [x] with the option off, `locks/*.lock` files are left untouched by pull and push

### DIFF
--- a/cli/src/commands/generate-metadata/generate-metadata.ts
+++ b/cli/src/commands/generate-metadata/generate-metadata.ts
@@ -412,6 +412,44 @@ export async function rehashOnly(
   return counts;
 }
 
+/**
+ * Normalize the tree's shared lockfiles, unless the run asked to stay inside one
+ * folder: shared lockfiles are workspace-wide, so the pass reads and rewrites
+ * metadata outside it, which `--strict-folder-boundaries` promises not to do.
+ */
+async function maybeDedupeLockfiles(
+  opts: GlobalOptions & SyncOptions & { strictFolderBoundaries?: boolean },
+  workspace: Workspace,
+  codebases: SyncCodebase[],
+  ignore: (p: string, isD: boolean) => boolean,
+  rawWorkspaceDependencies: Record<string, string>,
+  tree: DoubleLinkedDependencyTree,
+  folder: string | undefined,
+): Promise<void> {
+  if (!opts.dedupeLockfiles) return;
+  if (folder && opts.strictFolderBoundaries) {
+    log.info(
+      colors.yellow(
+        `Skipping lockfile deduplication: it spans the whole workspace, and --strict-folder-boundaries keeps this run inside "${folder}".`,
+      ),
+    );
+    return;
+  }
+  await beginLockfileBatch();
+  try {
+    await dedupeLockfilesOnDisk(
+      opts,
+      workspace,
+      codebases,
+      ignore,
+      rawWorkspaceDependencies,
+      tree,
+    );
+  } finally {
+    await flushLockfileBatch();
+  }
+}
+
 export async function generateMetadata(
   opts: GlobalOptions & {
     yes?: boolean;
@@ -611,6 +649,10 @@ export async function generateMetadata(
   // === Show stale items and confirm ===
   if (filteredItems.length === 0) {
     log.info(colors.green("All metadata up-to-date"));
+    // Turning `dedupeLockfiles` on in a repo whose metadata is already current
+    // is exactly the case where nothing is stale, and the conversion still has
+    // to happen — the sync compares against a deduplicated remote either way.
+    await maybeDedupeLockfiles(opts, workspace, codebases, ignore, rawWorkspaceDependencies, tree, folder);
     return;
   }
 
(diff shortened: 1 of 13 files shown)


rubenfiszel:
✅ Review round clean @ 4f15a73fc9e79bc0ddfa92424cd9c2333572cd17 — nit-only verdicts from all three reviewers at that SHA (codex/claude: mergeable-with-nits, pi: good to merge); nits addressed in 2baa81c094 (parse the active metadata twin's `lock` field instead of matching raw text, exclude metadata under `locks/` from the ownership scan, correct the `isNeverWalkedDir` rationale).

PR #10774, opened 2026-08-20
fix: ground the chat's AI agent provider in the workspace's models

Guilhem-lm:
## Summary

Asking the AI chat to build a flow with an AI agent step produced a step that failed on every run. The chat got the provider `kind` and `resource` right and **invented the `model` id** from its own training data — nothing in any prompt, tool or validation told it which models the workspace's AI resources actually serve. Three runs of a chat-written flow failed with:

```
API error calling https://api.anthropic.com/v1/messages: 404 Not Found
{"type":"not_found_error","message":"model: claude-3-haiku-20240307"}   @ai_executor.rs:1337
```

for `claude-3-haiku-20240307`, `claude-3-5-haiku-20241022` and `claude-3-5-sonnet-20241022`. The API key and resource were fine; those models simply do not exist for that key.

The workspace AI settings play no runtime role here — `ProviderWithResource` (`backend/windmill-ai/src/types.rs`) takes `resource` and `model` verbatim, and nothing in `windmill-ai`/`windmill-worker` reads `workspace_settings.ai_config`. So the fix reads the models from each resource's own endpoint, the same source the AI settings model picker uses: `GET /w/{workspace}/ai/proxy/models` with `X-Resource-Path`, which works for any AI resource the user can read, settings-configured or not.

Two mechanisms: the model list goes into the flow instructions so the chat picks a real id, and a write-time check rejects one that isn't.

## Changes

- **`flow/aiAgentProviderCatalog.ts` (new)** — lists the workspace's AI provider resources and calls `fetchAvailableModels` for each. Reads `getCopilotInfo` for that workspace (not the `copilotInfo` store, which tracks the *navigated* workspace) to order configured resources first and surface the default model. Cached 5 min per workspace, max 8 resources, 10s per listing; never rejects — on failure it resolves to an empty catalog rather than blocking a write.
- **`flow/aiAgentProviders.ts` (new)** — the pure half: the prompt section and `validateAiAgentProviders`. Blocks a provider that is a bare `"$res:..."` string, one missing `kind`/`resource`/`model`, a resource that is not an AI resource of the workspace, a `kind` that disagrees with the resource type, and a model outside a live listing.
- **Proxy resources are not held to the official names.** `hasCustomEndpoint` flags a resource with a non-empty `base_url`/`baseUrl`, a `platform` other than `standard`, or kind `customai` — a gateway names its models as it likes and may accept aliases it does not list. There an unlisted model is reported as a warning on the tool result instead of blocking, and the prompt marks the resource. A read failure counts as a custom endpoint, so a model is never wrongly rejected.
- **Validation is wired into all four write paths** — `write_flow` and `patch_flow_json` (global mode), `set_flow_json` and `patch_flow_json` (flow mode) — through `aiProviders` / `aiProviderWarnings` on the validation context. The error names the models the resource does serve, so a wrong guess self-corrects in one round-trip.
- **`get_instructions(subject: "flow")` is now async** and appends the live provider list and the provider shape. Measured at 1.1k characters for a one-resource workspace against 48.8k for `getFlowPrompt()`, so it rides along with the flow reference rather than needing a subject of its own.
- **The chat asks rather than guesses when the choice is open.** With exactly one AI resource *and* a workspace default model it uses that default and names it in its reply. With several resources, or no default, the prompt directs it to `askUserQuestion` for the resource and the model before writing the step. Flow mode gets the same section but never the `askUserQuestion` line, since `flowTools` has no such tool — there it is told to pick and name its choice instead.
- **`flows/aiAgentModules.ts` (new)** — `forEachAiAgentModule`, the AI-agent traversal, moved into a types-only leaf module. `collectProviderlessAgentIds` and `collectInvalidAgentToolNames` (from #10756, which had grown its own copy as `visitAgentModules`) now share it, and validation can traverse a flow without pulling the editor-side agent tool helpers — which drag in Monaco — into the node test environment.

## Screenshots

Global session chat in a workspace with two Anthropic resources: it asks which resource and which model before writing, then writes the step with the answers.

agent-provider-question

## Test plan

Exercised against a running EE instance (workspace with a real Anthropic resource), not only type-checked:

- [x] The catalog returns the resource's live model list (10 ids), its settings-configured models and the workspace default
- [x] `validateEditableFlowJson` rejects `claude-3-haiku-20240307` — the id from the original failing runs — and accepts `claude-sonnet-4-6`
- [x] A fresh session asked to build an agent flow calls `get_instructions`, writes `{"kind":"anthropic","model":"claude-sonnet-4-6","resource":"$res:u/admin/anthropic_windmill_codegen"}` first try, and the flow runs green (preview job returned its answer)
- [x] With a second Anthropic resource present, a fresh session asks which resource, then which model, then writes the step with the answers
- [x] A resource with `base_url` set to a gateway accepts an unlisted model; the same model on the provider-hosted resource is rejected
- [x] `npm run check` — 0 errors; `vitest` over `chat/flow` and `components/flows` — all pass, including #10756's tool-name tests through the shared traversal
- [ ] Reviewer: confirm the extra `getResourceValue` per AI resource (for `base_url`) is acceptable — it returns the resource value, including an inline API key, to the browser the same way the resource editor does

**Not covered:** providers on reusable `ai_agent` resources. A step that links to a saved agent inherits its provider, and nothing grounds or validates the model there — only providers written directly on flow modules go through this.

## Review round (local two-axis review, addressed)

**Correctness — an incomplete catalog could reject a valid provider.** `checkProviderValue` treated a non-empty catalog as exhaustive, so a resource missing from it was rejected as "not an AI provider resource of this workspace". Two silent ways to hit that with a perfectly good resource: the 8-resource cap truncating the list, and `listResource` failing (which left only the `ai_config`-named resources in the catalog — the very thing this PR argues is not authoritative). The catalog now carries `resourcesAreComplete`; an unknown resource only blocks when it is true, and the truncation logs what it dropped.

**Flow mode was enforcing without grounding.** It validated provider configs but never got the model list, so its only guidance stayed the `"model": "gpt-4o"` example in `flow-base.md`. `changeMode` now appends the catalog section to the flow-mode system message when it resolves; the chat loop re-reads `systemMessage` each iteration, so a send that beats the fetch picks it up on the next one, and a stale append is dropped by identity check.

Also: removed the static provider-shape lines from TypeScript that `system_prompts/base/flow-base.md:103-105` already carries verbatim; dropped an exported cache-clearer with no callers and an unread `configuredModels` field; a model list that is a fallback rather than a live listing now says so instead of being presented as authoritative; messages say `Step "<id>"` per `CONTEXT.md`, which lists *module* as a word to avoid.

## ai_evals A/B (`.claude/skills/ai-chat/SKILL.md`)

The harness could not see this change at all — `listResource` was stubbed to `[]` for benchmark workspaces, so an eval run got an empty catalog and rendered no provider section. `BenchmarkWorkspaceRunnables` now takes an `aiProviders` seed, served through `listResource` / `getResourceValue` / `getCopilotInfo`, with two fixtures (one provider; two providers).

Two new global cases, `--model sonnet`, same cases both sides:

| case | before (`574775d50c`) | after |
| --- | --- | --- |
| `global-ai-agent-step-uses-workspace-model` | **fail**, judge 30 | **pass**, judge 97 |
| `global-ai-agent-step-asks-which-provider` | **fail** (never called `askUserQuestion`) | **pass** |
| `global-test3-flow-create` (control) | pass, judge 97 | pass, judge 97 |

Before, with two AI provider resources seeded in the workspace, the assistant wrote:

```json
{"kind": "openai", "resource": "$res:f/ai_providers/openai", "model": "gpt-4o"}
```

— the placeholder resource from the docs example, which does not exist in that workspace, plus an invented model. That is the reported defect, reproduced in the benchmark.

Window cost of the added section: final context 37,518 → 37,717 tokens on the one-provider case (~200 tokens); the two-provider case lands at 38,003.

## Review round 1 (`ce37f67b53`)

All three reviewers flagged that the flow-mode prompt named `askUserQuestion`, a tool only global mode installs — introduced when this branch gave flow mode the catalog, and contradicted by the PR body, which is fixed above. `formatAiAgentProvidersPrompt` now takes `canAskUser`; flow mode passes false and gets "pick the one that best fits and name it in your reply".

Two further false-rejection paths, both from Codex, one also from Claude:

- **A filtered listing is not grounds for calling a model wrong.** `fetchAvailableModels` is the model picker's helper: it keeps only `gpt-`/`o`/`codex` ids for OpenAI and Azure OpenAI (so a `ft:` fine-tune never appears), keeps text models and inference profiles for Bedrock, and rewrites Google AI ids. Marking those `modelsAreLive` made validation reject models the endpoint serves. Authority is now its own flag, `modelsRuleOutOthers`, true only for kinds whose listing passes through unfiltered and endpoints that are not gateways. Anthropic — the reported bug — still blocks.
- **A cached catalog no longer rejects a resource created after it was built.** `getAiAgentProviderCatalogFor` re-reads the catalog once with the cache bypassed before an unknown resource can block, so "create an Anthropic resource, then build an agent flow" in one session works instead of failing for the rest of the 5-minute TTL.

Claude's P2 on cost is fixed by the same resolver: it collects the AI agent steps first and fetches nothing at all when a flow has none, which is most flow writes.

Claude's P2 on the eval harness was right that the A/B did not cover the headline behaviour: `/ai/proxy/models` had no benchmark handler, so every run took the `modelsAreLive: false` branch. The harness now serves the seeded listing (and defines a `location` for the frontend code that builds the URL from it). That handler did not actually match until `992f9274` — see round 3. Re-run on that harness, same three cases, `--model sonnet`:

| case | before (`574775d50c`) | after |
| --- | --- | --- |
| `global-ai-agent-step-uses-workspace-model` | **fail**, judge 30 | **pass**, judge 97 |
| `global-ai-agent-step-asks-which-provider` | **fail** (never called `askUserQuestion`) | **pass** |
| `global-test3-flow-create` (control) | pass, judge 97 | pass, judge 97 |

Still open from the round, deliberately: deleting a resource leaves a stale catalog that permits a now-invalid reference until the TTL lapses. That fails permissive, which is the direction every other unknown in this file is resolved in.

## Review round 2 (`13af889f6f`)

Codex blocked on one finding, Claude and Pi were non-blocking. All five are fixed.

**P1 — instance-level AI settings were offered as workspace resources.** `getCopilotInfo` returns the *effective* config, so a workspace with no AI settings of its own gets the instance settings back, whose `resource_path`s live in the `admins` workspace. The catalog was adding those to its candidates even when `listResource` never returned them, so the prompt offered a `$res:` reference that resolves against the flow's own workspace at run time and fails with "Resource not found" — the same class of broken step this PR exists to prevent. Candidate selection moved into a pure `selectAiAgentProviderCandidates`, which keeps only resources the workspace itself lists and uses the AI settings for ordering alone; it has the catalog-loading test Codex asked for, covering exactly the instance-fallback case.

Nits, all addressed:

- `patch_flow_json` in flow mode had lost its `Invalid JSON after replacement:` wrapper when the parse moved out of the `try` to gate the catalog fetch. The parse has its own wrapper again, and validation errors now propagate with their own diagnosis instead of being relabelled as JSON errors — matching what global mode already did.
- The cache bypass fired on every rejected write, so a model retrying an invented path rebuilt the whole catalog each time. It is now rate-limited per workspace (30s), which keeps the freshly-created-resource fix without the repetition.
- Anthropic's listing pages at 20 and `fetchAvailableModels` ignores `has_more`, while the AI proxy routes on the path alone so no caller can ask for more. A listing that fills a page is no longer treated as exhaustive — only a shorter one can rule a model id out.
- A workspace with zero AI provider resources said nothing at all: the prompt section was empty and validation stayed silent, so the model invented a path and the write was accepted. An authoritative empty catalog now tells the model there are none and to get one created, and a provider written anyway is reported (not blocked, since there is nothing to suggest instead).

Eval cases re-run on this head: 3/3.

## Review round 3 (`992f92740e`)

Codex blocked on two, Claude and Pi were non-blocking. All six are fixed.

**P1 — a provider's model listing was rendered into the chat's context verbatim.** A resource can point at a gateway someone else controls, and its `/models` response landed in a system prompt (flow mode) and in `get_instructions` output (global), so an id containing newlines or backticks could append instructions to the chat's own context, and an unbounded one could exhaust it. `sanitizeModelIds` now keeps only bounded, single-line, id-shaped strings — `claude-sonnet-5`, `meta-llama/Llama-3.3-70B-Instruct-Turbo`, `anthropic.claude-haiku-4-5-20251001-v1:0`, `ft:gpt-4o:acme::abc123` all survive — applied to the live listing and the fallback alike, with the hostile cases tested.

**P1 — the instance default model was pinned onto a workspace-local resource of the same kind.** `defaultModel` matched on provider kind, so a workspace with no AI config of its own took the *instance* default (whose resource lives elsewhere) and labelled it the default for a local resource whose endpoint serves something else. The default is now tied to the exact `resource_path` its provider is configured with, and dropped when that resource is not one of the workspace's own.

Nits, all addressed:

- `formatAiAgentProvidersPrompt`'s doc comment still said "empty string when the workspace has no AI provider resource" — the opposite of what the new empty-catalog branch does, and the line callers read to decide append-or-skip.
- The zero-resource report was a settled fact wrapped in the generic "which could not be ruled out" hedge, and dropped the remedy the prompt side carries. Findings the catalog established now say so plainly and carry the instruction.
- Anthropic's 20-per-page limit was applied to every provider, so OpenRouter, Together and Mistral permanently warned for a truncation that cannot happen to them. Scoped to `anthropic`.

**Pi found that the eval `/ai/proxy/models` handler never matched.** A global eval registers its workspace under a `mkdtemp` path, so the workspace id carries slashes, and the handler's `[^/]+` could not span it — every run fell through to the network, failed, and took the offline fallback. The claim in round 2 that eval runs exercised the live listing was therefore wrong; it is corrected above. The pattern is greedy now, with a unit test (`aiProviderModels.vitest.ts`) pinning a path-shaped workspace id, and a new case that can only pass off the live listing: the fixture's AI settings configure `claude-sonnet-5` alone while the resource serves Opus too, and the prompt asks for the Opus model.

Four cases on this head, `--model sonnet`: 4/4, including `global-ai-agent-step-uses-listed-model` writing `claude-opus-5` (judge 95) — an id reachable only through the seeded listing.

## Review round 4 (`8761f3a0ea`)

All three reviewers converged on one finding, and it was created by the round-3 fix: `sanitizeModelIds` capped a listing at 200 entries, and that cap was invisible to `modelsRuleOutOthers`, so an OpenRouter or Together resource treated a truncated list as exhaustive and rejected models it actually serves.

That is the fourth round in a row to find the same class of mistake — a secondhand list treated as the whole truth — so this one is fixed by construction rather than at the site:

- Models are now a `ModelListing` (`{ ids, complete }`), built only by `sanitizeModelListing`. Every way the list can fall short of what the endpoint serves — a filtered listing, a page limit, the entry cap, an unusable entry dropped, or no listing at all — is expressed as `complete: false` in that one function, and `rulesOutOtherModels` is the single derivation validation consults. A new truncation cannot leave the list looking exhaustive without going through it.

Codex's other two, both also from the last commit:

- **The workspace default model bypassed sanitization.** Live and fallback listings were shape-checked, but `aiConfig.default_model.model` was copied verbatim into the prompt, reopening the injection path the same commit closed. It is held to the same shape now, and dropped when it fails.
- **The pre-validation traversal could throw on malformed input.** `forEachAiAgentModule` iterated `v.branches` without an array check, and all four write paths now run it before Zod, so `"branches": {}` surfaced as `branches is not iterable` instead of the flow-schema error. Guarded, with a test.

244 frontend tests, `npm run check` clean, eval set 4/4 on this head.

## Review rounds 5 and 6 (`ef0a373d77`)

**Round 5** was the first non-blocking one: all three "Mergeable, but should ideally address nits", and the nits were stale prose rather than behaviour. The one that mattered was a comment in the `listResource` catch claiming AI-settings resources still get catalog entries — untrue since round 2, and a reader who "fixed" the code to match it would have reintroduced the instance-fallback leak that round removed. Also a test comment naming `modelsRuleOutOthers`, the field the round-4 refactor deleted.

Round 5 is also where the `ModelListing` refactor was checked by someone other than its author: Pi walked every blocking condition in `checkProviderValue` and found no remaining false-rejection path, and Claude verified the cap boundary (the `kept.length === MAX` check sits before the push, so a 200-entry listing keeps `complete: true` and a 201-entry one does not).

**Round 6**: Claude and Pi both "Good to merge"; Codex found two more.

- **A path match does not prove provenance.** `getCopilotInfo` returns the effective config, resource paths are workspace-scoped, and an instance-configured path can coincide with a local resource's. The default model is no longer offered on a path match alone — it is offered only when that resource's own listing confirms it serves that model, which holds whichever config the path came from. Otherwise there is no default and the prompt asks (or names its choice, in flow mode).
- **An untrusted listing was parsed before it was bounded.** The catalog lists models without anyone asking — opening a chat is enough — and `fetchAvailableModels` parsed, mapped and sorted the whole response before the 200-entry cap applied. A hostile gateway could exhaust the tab. `fetchAvailableModels` takes an optional `maxBytes`, and the catalog passes 2 MB; the reader refuses mid-stream rather than after. The picker's own behaviour is unchanged (no cap passed).

On the test suite: `copilot/lib.toolCalls` and `copilot/lib.anthropicRouting` have timeout-sensitive tests that hinge on how long `await import('./lib')` takes, and one of them (`truncated arguments`) fails on `main` as well. Immediately after a stash churns mtimes, two more fail on a cold transform; with a warm cache this branch produces the same single failure as `main`, three runs in a row. The suites this PR owns — `chat/flow`, `components/flows`, `boundedJson` — are 246/246 green, and `npm run check` is 0 errors over 10,251 files.

## Review rounds 7–10 (`4b49974b57`) — clean

**Round 10: Codex, Claude and Pi all "Good to merge".**

Rounds 7 through 9 were one rule, stated wrong three times in a row, each fix exposing the next facet:

- **Round 7** (Codex P1, Claude nit): requiring `models.complete` for the workspace default meant OpenAI, Azure, Google AI and Bedrock — filtered by `fetchAvailableModels`, so incomplete by design — could never have one. Presence in a live listing proves a model is served; completeness is only needed to rule an absent id *out*. Those are opposite directions.
- **Round 8** (Codex P1): the presence check read `models.ids`, capped at 200, so an OpenRouter default past entry 200 was invisible. The cap was never the memory bound — the 2 MB byte cap is — and the prompt only renders 25 ids, so the membership set is now 5000 and the two numbers do their own jobs.
- **Round 9** (Claude nit): raising that cap switched on blocking for five aggregator kinds for the first time. OpenRouter appends routing suffixes (`:online`, `:nitro`, `:floor`) to any listed model without listing the combinations and carries no `base_url`, so `customEndpoint` never covered it — `openai/gpt-4o:online` is served and would have been rejected, with an error naming 25 of 400 ids so the model could not self-correct. `ALIASING_MODEL_LISTING_KINDS` keeps those permissive; Anthropic, the reported bug, still blocks.

Also dropped a test both Codex and Claude flagged as scaffolding: it handed `formatAiAgentProvidersPrompt` a `defaultModel` that was already set, so it passed with the bug it was added for restored.

### Verified on the running instance at this head

```
resourcesAreComplete: true
defaultModel:  { kind: anthropic, model: claude-sonnet-4-6 }
option:        $res:u/admin/anthropic_windmill_codegen — live, complete, 10 ids, not a gateway
prompt tail:   "Unless the user asks for something else, use kind `anthropic` with model
                `claude-sonnet-4-6` — the workspace default. Name the model you used in your reply."
```

247 tests green on the suites this PR owns, `npm run check` 0 errors over 10,251 files, eval cases 4/4.

### DIFF
--- a/ai_evals/adapters/frontend/aiProviderModels.vitest.ts
+++ b/ai_evals/adapters/frontend/aiProviderModels.vitest.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from 'vitest'
+import {
+	handleBenchmarkApiFetch,
+	hasBenchmarkApiHandler,
+	registerBenchmarkWorkspaceRunnables,
+	unregisterBenchmarkWorkspace
+} from './mockBackend'
+
+// A global eval run registers its workspace under a mkdtemp path, so the workspace id the
+// frontend interpolates into the models URL carries slashes. A handler that assumed a single
+// path segment silently fell through to the network, and every run took the offline fallback.
+const WORKSPACE = '/tmp/wmill-frontend-global-benchmark-abc123'
+const RESOURCE = 'f/evals/global/anthropic_main'
+const URL_FOR = (workspace: string) =>
+	`http://benchmark.local/api/w/${workspace}/ai/proxy/models`
+
+describe('benchmark /ai/proxy/models', () => {
+	it('serves the seeded listing for a workspace id that is a path', async () => {
+		registerBenchmarkWorkspaceRunnables(WORKSPACE, {
+			aiProviders: [
+				{ path: RESOURCE, kind: 'anthropic', models: ['claude-sonnet-5', 'claude-opus-5'] }
+			]
+		})
+		try {
+			expect(hasBenchmarkApiHandler(URL_FOR(WORKSPACE))).toBe(true)
+			const response = handleBenchmarkApiFetch(URL_FOR(WORKSPACE), {
+				headers: { 'X-Resource-Path': RESOURCE, 'X-Provider': 'anthropic' }
+			})
+			await expect(response.json()).resolves.toEqual({
+				data: [{ id: 'claude-sonnet-5' }, { id: 'claude-opus-5' }]
+			})
+		} finally {
+			unregisterBenchmarkWorkspace(WORKSPACE)
+		}
+	})
+})
--- a/ai_evals/adapters/frontend/mockBackend.ts
+++ b/ai_evals/adapters/frontend/mockBackend.ts
@@ -71,6 +71,7 @@ export interface BenchmarkWorkspaceAiProvider {
 	path: string
 	/** Resource type, which for AI resources is the provider kind (`anthropic`, `openai`, ...). */
 	kind: string
+	/** What this resource's `/ai/proxy/models` listing returns. */
 	models?: string[]
 	/** Set to point the resource at a gateway rather than the provider's own API. */
 	base_url?: string
@@ -1164,6 +1165,9 @@ function parseBenchmarkRequestBody(
 /** True when `handleBenchmarkApiFetch` has an answer for this `/api/...` url.
  * Any other relative fetch must keep its normal (non-benchmark) behavior —
  * intercepting it with a synthetic 404 sends the model into retry loops. */
+// Not anchored: the frontend builds this URL from location.origin, so it arrives absolute.
+const BENCHMARK_AI_MODELS_PATH = /\/api\/w\/([^/]+)\/ai\/proxy\/models$/
+
 export function hasBenchmarkApiHandler(url: string): boolean {
 	const path = url.split('?')[0]
 	return (
@@ -1172,7 +1176,8 @@ export function hasBenchmarkApiHandler(url: string): boolean {
 		BENCHMARK_RUN_BY_PATH.test(path) ||
 		/^\/api\/w\/[^/]+\/jobs\/queue\/list$/.test(path) ||
 		path === '/api/embeddings/query_hub_scripts' ||
-		path.startsWith('/api/scripts/hub/get_full/')
+		path.startsWith('/api/scripts/hub/get_full/') ||
+		BENCHMARK_AI_MODELS_PATH.test(path)
 	)
 }
 
@@ -1182,6 +1187,17 @@ export function handleBenchmarkApiFetch(url: string, init?: RequestInit): Respon
 	const path = url.split('?')[0]
 	if (path === '/api/workers/list') {
 		return Response.json(BENCHMARK_WORKERS)
+	}
+	// The provider's own model listing, which grounds an AI agent step's model id. Keyed by the
+	// resource the caller names, so two seeded providers can serve different models.
+	const aiModels = BENCHMARK_AI_MODELS_PATH.exec(path)
+	if (aiModels) {
+		const headers = new Headers(init?.headers)
+		const resourcePath = headers.get('X-Resource-Path') ?? ''
+		const seed = benchmarkWorkspaceRunnables
+			.get(decodeURIComponent(aiModels[1]))
+			?.aiProviders?.find((entry) => entry.path === resourcePath)
+		return Response.json({ data: (seed?.models ?? []).map((id) => ({ id })) })
 	}
 	if (/^\/api\/w\/[^/]+\/jobs\/queue\/list$/.test(path)) {
 		return Response.json([])
--- a/ai_evals/adapters/frontend/mockBackend.ts
+++ b/ai_evals/adapters/frontend/mockBackend.ts
@@ -5,6 +5,7 @@ import type {
 	Flow,
 	Job,
 	ListableApp,
+	ListableResource,
 	ListableVariable,
 	Script
 } from '../../../frontend/src/lib/gen'
@@ -64,6 +65,21 @@ export interface BenchmarkWorkspaceVariable {
 	ws_specific?: boolean
 }
 
+/** An AI provider resource of the benchmark workspace, as an AI agent step would reference it.
+ * `models` stands in for the provider's model listing, which no eval run can reach. */
+export interface BenchmarkWorkspaceAiProvider {
+	path: string
+	/** Resource type, which for AI resources is the provider kind (`anthropic`, `openai`, ...). */
+	kind: string
+	models?: string[]
+	/** Set to point the resource at a gateway rather than the provider's own API. */
+	base_url?: string
+	/** Models the workspace AI settings selected for this provider. */
+	configuredModels?: string[]
+	/** Marks this provider's first configured model as the workspace default. */
+	isDefault?: boolean
+}
+
 export interface BenchmarkWorkspaceJob {
 	/** Stable id so a case prompt can reference a specific run (e.g. for get_job_logs). */
 	id?: string
@@ -80,6 +96,7 @@ export interface BenchmarkWorkspaceRunnables {
 	flows?: BenchmarkWorkspaceFlow[]
 	apps?: BenchmarkWorkspaceApp[]
 	variables?: BenchmarkWorkspaceVariable[]
+	aiProviders?: BenchmarkWorkspaceAiProvider[]
 	datatables?: BenchmarkDatatableSeed[]
 	jobs?: BenchmarkWorkspaceJob[]
 }
@@ -244,6 +261,59 @@ export function listBenchmarkVariables(workspace: string): ListableVariable[] | 
 	}
 	// The list route never decrypts.
 	return (runnables.variables ?? []).map((seed) => buildBenchmarkVariable(workspace, seed, false))
+}
+
+/** AI provider resources of a benchmark workspace, shaped like `ResourceService.listResource`
+ * rows (which carry no value). Null when the workspace is not a benchmark one. */
+export function listBenchmarkAiProviderResources(workspace: string): ListableResource[] | null {
+	const runnables = benchmarkWorkspaceRunnables.get(workspace)
+	if (!runnables) {
+		return null
+	}
+	return (runnables.aiProviders ?? []).map((seed) => ({
+		workspace_id: workspace,
+		path: seed.path,
+		resource_type: seed.kind,
+		value: null,
+		is_oauth: false,
+		is_linked: false,
+		is_refreshed: false,
+		extra_perms: {},
+		edited_at: BENCHMARK_TIMESTAMP
+	}))
+}
+
+/** The value of a seeded AI provider resource. Only the endpoint fields are modelled — a key is
+ * never needed, because no eval run calls the provider through this resource. */
+export function getBenchmarkResourceValue(
+	workspace: string,
+	path: string
+): Record<string, unknown> | null {
+	const seed = benchmarkWorkspaceRunnables
+		.get(workspace)
+		?.aiProviders?.find((entry) => entry.path === path)
+	if (!seed) {
+		return null
+	}
+	return seed.base_url ? { base_url: seed.base_url } : {}
+}
+
+/** The AI settings of a benchmark workspace, as `WorkspaceService.getCopilotInfo` returns them. */
+export function getBenchmarkAiConfig(workspace: string): Record<string, unknown> | null {
+	const seeds = benchmarkWorkspaceRunnables.get(workspace)?.aiProviders
+	if (!seeds) {
+		return null
+	}
+	const providers: Record<string, unknown> = {}
+	let defaultModel: { model: string; provider: string } | undefined
+	for (const seed of seeds) {
+		const models = seed.configuredModels ?? seed.models ?? []
+		providers[seed.kind] = { resource_path: seed.path, models }
+		if (seed.isDefault && models[0]) {
+			defaultModel = { model: models[0], provider: seed.kind }
+		}
+	}
+	return { providers, ...(defaultModel ? { default_model: defaultModel } : {}) }
 }
 
 export function getBenchmarkVariableByPath(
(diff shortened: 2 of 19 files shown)


stefannut (commented):
### Architectural & Frontend Review: AI Agent Provider Grounding & Catalog Caching

Great enhancement to ground `aiagent` copilot flow modules within the workspace's configured AI resources and available model inventory.

#### Key Review Points:

1. **Resilient Catalog Resolution**:
   - The fail-open pattern in `aiAgentProviderCatalog.ts` (resolving to an empty catalog with fallback rather than rejecting the promise) is crucial. It ensures transient provider network hiccups do not stall the user's flow builder session.
   - `CACHE_TTL_MS` (5 min) paired with `MODELS_TIMEOUT_MS` (10s) provides a good balance between fresh model discovery and snappy UI transitions.

2. **Resource Throttling (`MAX_PROVIDER_RESOURCES = 8`)**:
   - Capping concurrent provider queries prevents browser request saturation in enterprise workspaces with dozens of configured LLM API keys.

3. **Type Safety**:
   - Clean integration of Svelte store typing (`AiAgentProviderCatalog`, `AiAgentProviderOption`) and `$lib/gen` OpenAPI schema bindings.

LGTM! Clean, defensive implementation.

Guilhem-lm:
✅ Review round clean @ 4b49974b5789878c2eac30c006e1de5cddb938da

PR #10776, opened 2026-08-20
feat: add WM_ROOT_WORKSPACE, the closest dev or prod workspace of a job

rubenfiszel:
## Summary

Adds `WM_ROOT_WORKSPACE`, a contextual variable naming the workspace a job belongs to *as an environment*: the nearest ancestor-or-self that is a root ("prod") workspace or a dev workspace. It equals `WM_WORKSPACE` for both of those, and for a throwaway fork it names the standing workspace the fork descends from — so a script running in an ephemeral fork can still key external resources, logging, or routing off the environment it came from.

Resolution is a single recursive walk of `workspace.parent_workspace_id`, stopping at `parent_workspace_id IS NULL OR is_dev_workspace`.

**Note this is deliberately not the same question as two neighbouring resolvers**, and the divergence is pinned by tests:

- `get_billing_workspace_id` walks all the way to the parentless root. A fork under a dev workspace bills to prod but *belongs to* the dev environment.
- `tags::tag_workspace_id` walks past a dev workspace whose id carries the generated `wm-fork-` prefix, because nothing provisions workers for such an id. A dev workspace is still an environment, so `WM_ROOT_WORKSPACE` stops there.

## Caching

The ask was that lookups be rare, so this is one lookup per workspace per *invalidation*, not per job:

- `ROOT_WORKSPACE_CACHE` has a 5 min TTL, but the TTL is only a backstop — correctness rests on the sweep. `windmill_queue::tags::invalidate_fork_parent_cache` and `apply_fork_lineage_change` now drop this cache too, so it inherits every existing lineage-change call site and the `notify_fork_lineage_change` cross-process broadcast with **no new call sites to keep in sync**.
- `create_fork` was the one lineage mutation not routed through that entry point; it now is, so a pre-creation lookup can't pin a new fork to itself.
- A *failed* lookup is not cached at all (matching `lookup_tag_workspace`'s documented reasoning): the fallback value is indistinguishable from a legitimate answer, so caching one DB blip would make every job in a fork misreport its environment for the whole TTL. It serves the last value this process actually resolved instead of degrading to `w_id`.
- An agent worker consumes no `notify_event`, so no sweep ever reaches it and the TTL is its whole invalidation story. It takes a 60s TTL — the same staleness bound `FORK_ANCESTOR_CHAIN_CACHE` and `BILLING_WORKSPACE_CACHE` already accept — rather than the long one only a broadcast-fed process has earned.

## Changes

- `windmill-common/src/workspaces.rs` — `ROOT_WORKSPACE_CACHE`, `root_workspace_id` (cached, `Connection`-aware), `lookup_root_workspace_id` (uncached SQL)
- `windmill-common/src/variables.rs` — the `WM_ROOT_WORKSPACE` contextual variable
- `windmill-queue/src/tags.rs`, `src/main.rs`, `windmill-api-workspaces/src/workspaces.rs` — invalidation wiring
- `windmill-common/tests/root_workspace.rs` — resolution + invalidation regression guards
- **EE companion (windmill-labs/windmill-ee-private#736)** — an agent worker has no database, so `Connection::Http` needs a route: `GET /api/w/{w_id}/agent_workers/root_workspace`, mirroring `custom_envs`. Internal agent routes are not in `openapi.yaml`, so no spec change is owed.
- `frontend/src/lib/{process,deno_fetch}.d.ts.txt` — Monaco autocompletion stubs

## Screenshots

The only frontend change is the two Monaco `.d.ts` stubs, whose visible effect is completion and hover in the script editor:

monaco-wm-root-workspace-completion

## Open question for review

`WM_ROOT_WORKSPACE` is the name from the original request, and it is a permanent user-facing surface. Review flagged that "root workspace" already means *parentless workspace* elsewhere in this module (`get_billing_workspace_id`'s doc, the attach/detach comments), and this resolver returns a dev workspace when the fork sits under one. I kept the requested name and cross-referenced `get_billing_workspace_id` from the doc comment so nobody reaches for the wrong one — but if `WM_ENV_WORKSPACE` / `WM_BASE_WORKSPACE` reads better, now is the only time to change it.

## Test plan

Exercised on a running instance against a real `prod → dev → forks` hierarchy built through the EE fork/attach endpoints, with actual bash jobs (not just type checks):

- [x] `rwtest-prod` (root) → `rwtest-prod`; `rwtest-dev` (dev) → `rwtest-dev`
- [x] fork under the dev workspace → `rwtest-dev` (not prod — the case that diverges from tag resolution)
- [x] fork under prod → `rwtest-prod`
- [x] promoting a fork to dev flipped its value **on the very next job**, confirming the sweep rather than a TTL expiry
- [x] the whole table repeated through a **real agent worker** (`MODE=agent`, `worker=ag-…`) per `docs/agent-worker-e2e.md`, exercising the `Connection::Http` branch and the new EE route. The fork case returned the dev workspace, which the silent fallback could not have produced.
- [x] `cargo test -p windmill-common --test root_workspace` and `-p windmill-queue --test tag_workspace_test`
- [x] `cargo check` with `enterprise,private,license`; `cargo sqlx prepare` adds one entry, zero deletions


rubenfiszel:
✅ Review round clean @ 420a86dd60fb812c0fda58aeb8b08df601e894ca

PR #10777, opened 2026-08-20
feat: inline login errors and a narrower single-column login card

Guilhem-lm:
## Summary

A wrong password used to be reported by a toast that faded after a few seconds, far from the fields that caused it. The login card now reports credential failures inline, the way a login form normally does: a message under the field it is about, red borders on that field, and one short shake when you retry the same credentials. Errors with no field to attach to — an SSO failure from `?error=`, a blocked popup, a missing SAML config — still toast.

Around that, the card got a layout pass: one column of provider buttons at any count, a narrower card anchored to the top of the page, an `OR` divider between the two ways in, the official icon-library brand marks at lucide's proportions, and copy that only offers sign-up when a third-party login is configured. The browser now also remembers which method last worked and leads with it.

The logged-out pages were harmonized while we were in there: instead of each centering a large Windmill mark, one top-left header brands them all. The login page is the exception — it keeps a mark in the middle above "Log in to Windmill".

`/kitchen_sink/login` is a new DEV-only page rendering all fourteen states of the card from fixed configs with no API calls, so states that need a configured provider can be reviewed without one.

### The login card

- **Inline credential errors** — `InputError` under the field, `error` borders on it, `role="alert"` plus `aria-invalid`/`aria-describedby` so a failure still reaches a screen reader. The message is keyed to the credentials that produced it: editing either field retires it, and a failure that arrives while you are typing is attributed to what was submitted. Its line is reserved (`min-h-5`), so appearing and clearing shifts nothing.
- **Per-field messages** — "Enter your email." reddens and sits under Email, "Enter your password." under Password, "Invalid email or password." under Password with both fields flagged.
- **Error text is allowlisted** — `loginErrorMessage` (`src/lib/loginError.ts`, with tests) maps `Invalid login`, the 429 and password-login-disabled cases, and falls back to a generic sentence for anything else. Without it a `SqlErr` (also a 400, carrying the sqlx error and a source path) or a proxy's HTML error page would be printed verbatim to an unauthenticated visitor.
- **Shake on retry only** — 0.2s, ±2px, linear, `motion-safe:`; it fires when the message is already on screen, and replays on every retry (a forced style recalculation between removing and re-adding the class, or the browser coalesces both and replays nothing).
- **Last used method** — `src/lib/lastLoginMethod.ts` records what worked, only on confirmed success: the password path writes after the API accepts it; OAuth and SAML park the method and it is promoted once a session exists. That method leads the card — a provider moves to the top of the list with a badge on its corner, the password form takes the top of the card already open — so a returning user lands on the way in they used last time.
- **Layout** — provider buttons always one column; card `max-w-xl` → `max-w-sm`; anchored to the top rather than vertically centered, so opening the form or showing an error grows the card downward instead of sliding it; `OR` divider between providers and the form; the third-party toggle disappears once the form is open; "Forgot password?" sits under the Log in button.
- **Copy** — buttons read "Continue with Google" and so on; the primary button and the heading both say "Log in"; the heading is "Log in or sign up to Windmill" only when a provider or SAML is configured, since `login_externally` creates the account on first third-party login. A whitelabelled instance uses its own name.
- **Buttons on the unified scale** — `unifiedSize` on every button in the card, replacing the deprecated `size` prop and `on:click`.

### Icons

- Provider buttons use the icon-library marks instead of the monochrome `icons/brands/*` glyphs, which are deleted (`Discord.svelte` stays, still used).
- All eight provider marks got a viewBox sized so the artwork fills 90% of the box — lucide's proportion. Vendor SVGs ship arbitrary padding (Google none, GitHub 10%), so they rendered at visibly different sizes; `MicrosoftIcon`/`GitlabIcon` also gained `size`/`class` and finally honour the size their call sites pass.

### The logged-out pages

- `LoginPageHeader` is now a top bar: mark + wordmark on the left, dark-mode toggle and version on the right. A whitelabelled instance shows its own name without the mark.
- `CenteredModal` dropped its centered logo, so its 15 pages move together; pages that only wait (logging out, redirecting) get a small spinner beside the title, which the spinning logo used to provide.
- forgot-password and reset-password lost their centered marks and a duplicate dark-mode toggle, and match the login page's width and rhythm.

### Dev

- `/kitchen_sink/login` renders every state; `Login.svelte` takes an opt-in `preview` prop whose effect is gated on `import.meta.env.DEV`, so it is inert in a production bundle whatever a caller passes, and the page redirects to `/` outside dev.

## Screenshots

Default card, password-only instance:

pr2-login-default

Wrong password — message under the field it is about, both fields flagged, no layout shift:

pr2-login-error

Last used a provider — badged on the corner and moved to the top of the list:

pr2-lastused-provider

Last used email and password — the form leads the card, already open, providers under the divider:

pr2-lastused-password

`/kitchen_sink/login` — every state of the card:

pr2-kitchen-sink

## Test plan

- [ ] Wrong password: message under the password field, both fields red, no toast, nothing on the page moves. Submit again unchanged → the card shakes once. Edit the password → message clears; the next failure shows it again without a shake.
- [ ] Press Enter with only one field filled: "Enter your email." / "Enter your password." under the empty field, only that field red.
- [ ] Correct password logs in; return to `/user/login` and the form is open with a "Last used" state (badge on the provider case).
- [ ] With an OAuth provider configured: heading reads "Log in or sign up to Windmill", buttons stack one per row as "Continue with X", the `OR` divider sits between them and the form, and the toggle disappears once the form is open.
- [ ] Log in through a provider, then return to the login page: that provider is first in the list with the badge on its top-right corner.
- [ ] Password-only instance: heading reads "Log in to Windmill", no divider.
- [ ] An SSO failure (`/user/login?error=...`) still toasts.
- [ ] Short viewport (~420px tall) with several providers: the page scrolls and the top of the card stays reachable.
- [ ] The other logged-out pages (workspaces, first-time, accept invite, logout, CLI auth, checkout) show the top-left brand and no centered mark; a page that only waits shows a spinner beside its title.
- [ ] `/kitchen_sink/login` in dev renders all states and never calls the API; in a production build it redirects to `/`.

Unexercised: this worktree has no OAuth/SAML provider configured, so the SSO states were verified through the kitchen sink and the persistence round trip through the password path rather than against a live provider; the production-build redirect was not run.


Guilhem-lm:
✅ Review round clean @ 4d22506d37aff1749bb94c8a59a20015a2e59c7f

PR #10778, opened 2026-08-20
feat: support application default credentials for gcp pub/sub triggers

hugocasa:
## Summary

GCP Pub/Sub triggers can now authenticate with the instance's **application default credentials** instead of a `gcloud` resource, for operators who would rather not hand Windmill a downloadable service-account key (GKE Workload Identity, a metadata-server identity, or `GOOGLE_APPLICATION_CREDENTIALS`).

ADC is selected by leaving the resource unset. Because those credentials are the *server's* own identity and no workspace ACL covers them — unlike a `gcloud` resource, whose ACL decides who may use it — **selecting ADC requires workspace admin**. That mirrors who picks ambient credentials for workspace object storage today.

Two smaller things ride along, both in the same code path:

- **Cross-project subscriptions.** `subscriptions/list` returned bare ids, so a subscription living in a different project than its topic was silently re-qualified with the client's own project and pointed at something that did not exist. It now returns fully qualified names, and subscriptions are stored fully qualified so one Pub/Sub subscription has exactly one representation.
- **Workload Identity Federation.** `google-cloud-pubsub` is pulled in with default features, which compiles out the `external_account` token source; such an ADC file was rejected as an unsupported account type. Enabling the feature is a one-line dependency change.

## Changes

- `gcp_resource_path` is now nullable; `NULL` means application default credentials. New optional `project_id`, needed because some ADC credentials name no project and because it selects which project topics are listed from.
- `require_admin_for_default_credentials` guards every path where a caller *chooses* ADC: trigger create/update, the capture config, test-connection, topic listing, subscription listing, subscription deletion. Deliberately **not** applied to the listener or the push-request handler, which read an already-persisted choice rather than accepting one.
- `GcpIdentity` carries only the service-account email and certificate URL, never the credentials themselves.
- Subscriptions are canonicalized to `projects/<project>/subscriptions/<id>` on write; the unique index keys off that, so a bare id and the qualified form of the same subscription can no longer both be created.
- Enables the `external-account` cargo feature and resolves the impersonated service account from `service_account_impersonation_url` so authenticated push works under WIF.
- Documents `GOOGLE_APPLICATION_CREDENTIALS` in the README, including that the `executable` credential source is unsupported.

## Verification

Verified against **production Google Pub/Sub**, not just an emulator:

| Path | Result |
|---|---|
| ADC pull: real token → real subscription → real message → job | passes |
| ADC authenticated push: Google-signed OIDC token accepted | passes |
| Forged / missing / wrong-key push tokens rejected, no job created | passes, incl. a valid `kid` with a bad signature |
| Workspace-admin gate on all ADC entry points | 403 for non-admins; resource mode unaffected |
| Cross-project topic/subscription via fully qualified names | passes (emulator) |
| Capture ("Test") config, both credential modes | passes (emulator) |
| WIF credential type reaches Google STS; impersonation email resolved | passes |

**Not verified:** the metadata-server branch of `adc_service_account_email()`, which needs a GCP-hosted process; and the STS *success* path for WIF, which is upstream crate code and needs a real workload identity pool.

## Test plan

- [ ] Non-admin cannot create or edit an ADC trigger, and the UI disables the option for them
- [ ] Admin can create an ADC pull trigger; publishing to the topic runs the job
- [ ] An existing resource-backed trigger still loads in "Service account" mode with its resource selected
- [ ] Reopening an ADC draft keeps ADC selected
- [ ] Switching credential mode clears the previously selected topic and subscription

## Screenshots

**Service account (default).** The credential toggle is new; the `gcloud` resource picker is where it always was, with the optional Project ID beneath it.

service account mode

**Application default credentials.** No resource to pick: the picker disappears and Test connection / Topic become available immediately.

application default credentials mode

**Non-admin opening a trigger an admin created in ADC mode.** It loads in ADC mode, the toggle is disabled, and the reason saving is unavailable is stated rather than surfacing as a bare 403.

non-admin view

**Subscription picker.** Subscriptions are listed fully qualified and labelled with their project, so the two `orders-consumer` subscriptions in different projects are distinguishable — the cross-project case this PR fixes.

subscription picker

## EE

Companion PR: windmill-labs/windmill-ee-private#738

---
## Summary by cubic
Support application default credentials (ADC) for GCP Pub/Sub triggers so workspaces can use the server’s identity instead of uploading service‑account keys. Old behavior required a `gcloud` resource; new behavior sets `gcp_resource_path = NULL` to select ADC, with admin gating wherever callers choose or enable ADC.

- Schema/behavior
  - Make `gcp_resource_path` nullable (`NULL` selects ADC); reject empty strings. Add optional `project_id`.
  - Accept bare or fully qualified topic/subscription names; store `subscription_id` fully qualified (400 chars). List endpoints return fully qualified names.
  - Update unique index to dedupe on fully qualified `subscription_id` + workspace using `COALESCE(gcp_resource_path, '')`.
  - Enforce admin-only when selecting ADC on create/update, capture test, topic/subscription listing and deletion, and when enabling a trigger. Listeners/push handlers use the stored choice and are not gated. Enabling authorizes under a row lock to avoid races.
  - Enable Workload Identity Federation by building `google-cloud-pubsub` with the `external-account` feature; the `executable` credential source remains unsupported.

- API/UI and rollout
  - New ADC endpoints: GET `/w/{workspace}/gcp_triggers/topics/list`, POST `/w/{workspace}/gcp_triggers/subscriptions/list`, DELETE `/w/{workspace}/gcp_triggers/subscriptions/delete`.
  - Existing `.../topics/list/{path}` and `.../subscriptions/list/{path}` accept optional `project_id`.
  - Editor: adds a Service account vs Application default credentials toggle (permission keyed off the loaded mode), a Project ID field, labels subscriptions with project, clears selections on mode switches, skips ADC-only listing for callers who cannot use ADC, and surfaces admin-gate errors when enabling.
  - Migration: apply `20260820073246_gcp_trigger_application_default_credentials`. Rebuild the backend with `google-cloud-pubsub` `external-account`. If using ADC, ensure the server has ADC available (`GOOGLE_APPLICATION_CREDENTIALS`, metadata server, or WIF). Existing resource-backed triggers continue to work unchanged.

<sup>Written for commit 72503923785a7034e326d3a8378246c72b369a4c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
✅ Review round clean @ 0945dbd7ff4882d334edbf60919cfe8e804612c3

PR #10784, opened 2026-08-20
feat: upgrade bun to 1.4.0 and demote deno in the language picker

rubenfiszel:
## Summary

Two independent changes:

1. **Bun 1.3.10 → 1.4.0** in every image that ships a bun binary, plus the CI pins so tests run against the bun we ship.
2. **Deno demoted in the language picker** — moved from position 3 to last, and relabelled `TypeScript (Deno)` → `Deno`, ahead of eventual deprecation.

Bun 1.4.0 is a full Zig→Rust rewrite of the runtime and was released 2026-08-20, so there is no patch release behind it yet. It was exercised end-to-end against a live worker before opening this (see Test plan).

## Changes

**Bun pins**
- `Dockerfile`, `docker/DockerfileSlim`, `docker/DockerfileSlimEe`, `.github/DockerfileBackendTests`: `oven/bun:1.3.10` → `oven/bun:1.4.0`
- `bun-version: 1.3.10` → `1.4.0` in `backend-test.yml`, `backend-test-windows.yml`, `git-sync-test.yml`, `ai-evals-test.yml`, `ai-agent-tests.yml`. Without this CI would keep testing 1.3.10 while the images ship 1.4.0, so a 1.4-specific regression would escape CI.
- `debugger/Dockerfile`: `oven/bun:1` → `oven/bun:1.4.0`. This image drives `prepare-deps` against the same DB-stored lockfiles as the workers, so a floating tag there is exactly where version skew would bite.
- `docker/DockerfileCli` (`oven/bun:slim`) and the `bun-version: latest` pins in `cli-tests.yml`, `sdk-tests.yml`, `npm_on_release.yml`, `publish_extra.yml` are left floating — pre-existing, and out of scope here.

**Deno demotion**
- `frontend/src/lib/scripts.ts`: `deno` moved to the end of `scriptLanguagesArray` and relabelled `Deno`. That array is the default picker order for every surface that falls back to `Object.keys(defaultScriptLanguages)` (script editor, flow step, app inline script). Workspaces with a saved `$defaultScripts.order` keep their own order, unchanged.
- `frontend/src/lib/components/assets/AssetGraph/pipelineLanguages.ts`: same relabel, and moved after Go so the pipeline picker matches.

## Compatibility notes

- **Lockfiles.** Bun 1.4 writes `bun.lock` v2, or **v3** when `package.json` carries nested or
  version-scoped overrides. Bun 1.3 does not fail on either: it prints
  `error: Unknown lockfile version`, then installs from `package.json` alone. Because the
  dependency job writes every dependency as `"latest"`, the lockfile is the *only* pin, so an
  older worker reading a newer lock resolves whatever is newest rather than drifting inside a
  range.

  **This PR closes that.** Below `MIN_VERSION_SUPPORTS_BUN_LOCKFILE_V2` (1.794.0) `gen_bun_lockfile`
  plants an empty `lockfileVersion: 1` lockfile before `bun install` (so the runtime "no lockfile
  found" path seeds one too, though that copy is discarded with the job dir — only the dependency
  job stores anything). Bun keeps the version
  the lockfile already had and raises it only when the dependencies genuinely need newer syntax —
  so the stored lockfile is a real v1 written by bun, not a header rewritten by us. If bun raises
  it anyway, the job fails with an actionable error instead of storing a lock older workers would
  silently misread. Once every worker is on 1.794.0+ the seeding stops and v2/v3 flow through
  untouched, with no follow-up change.

  Verified end-to-end on a bun 1.4.0 worker with the fleet below the gate:

  | case | outcome |
  |---|---|
  | flat deps | stored lock is `lockfileVersion: 1`, script runs green |
  | workspace deps with a nested override | deploy refused: *"bun produced a v3 lockfile, which workers older than 1.794.0 cannot read…"* |
  | gate temporarily lowered below the fleet | stored lock is `lockfileVersion: 2` — seeding correctly disengages |
  | existing v1 lock + bun 1.4 | pins honoured exactly, lock left at v1 (no mass re-lock on upgrade) |

- All bun flags this codebase passes still exist in 1.4: `install --save-text-lockfile --no-cache`,
  `run --preserve-symlinks -i --prefer-offline -r`, `--config=<bunfig>`, `--minimum-release-age`.

## Screenshots

Script settings language picker — Deno no longer at position 3:

language-picker-top

…and now last, labelled `Deno`:

language-picker-deno-last

## Test plan

Exercised on a live worker by pointing `BUN_PATH` at a 1.4.0 binary and re-running the same suites against the same database on both versions.

- [x] 18-case runtime suite (npm deps incl. pinned versions, `windmill-client` SDK, `node:*` builtins, `Bun.*` APIs, web streams, worker_threads, TS syntax, CJS interop, large results, args/env): identical 15/18 on 1.3.11 and 1.4.0 — same three failures on both, none introduced by 1.4
- [x] Deployed scripts: relative imports (`./dep`, `./nested/deep`), absolute windmill imports (`/f/.../dep`), and a two-step flow passing data between bun steps — 5/5 on both versions. This is what exercises the `loader.bun.js` `onResolve`/`onLoad` plugin, which behaves identically.
- [x] Lockfile compatibility in both directions, using a lockfile pulled from app.windmill.dev, with version-drift detection; both branches of the version gate exercised through the real dependency job (see above)
- [x] Job latency improves — median `duration_ms` over 9 runs: trivial job 31→23ms, `node:*` builtins 32→25ms, `windmill-client` SDK 52→37ms, 3 npm deps 64→61ms
- [x] Deno picker verified in a running browser: absent from position 3, last in the list, labelled `Deno`
- [ ] Docker images build and the CI workflows pass on 1.4.0
- [ ] Not exercised: raw-app/browser `bun build`, codebase-bundle scripts, dedicated workers, and the 1.4 `trustedDependencies`/`nativeDependencies` change (packages with postinstall scripts)

### Pre-existing failures, confirmed not caused by this bump

Both reproduce identically on bun 1.3.x and are **not** regressions from 1.4:

- `//nodejs` scripts mis-bundle certain npm import shapes. `import { chunk } from "lodash"` produces a bundle ending `…process.exit(1)} = pkg;` which node's ESM loader rejects; `import * as _ from "lodash"` bundles but yields `G.chunk is not a function`. `import _ from "lodash"` (default) works, as do dayjs, zod, uuid and mime-types in both named and default form — so this is specific to certain package/import-shape combinations, not to npm deps in general. Reproduced on app.windmill.dev (bun 1.3.10, node v20.20.2) and locally on 1.3.11 and 1.4.0 (node v26.4.0), so it is independent of both the bun and node version.
- `better-sqlite3` crashes bun with `NAPI FATAL ERROR: Error::New napi_get_last_error_info` on x64 — same panic on 1.3.11 and 1.4.0.

### Not yet verified

- Legacy binary `bun.lockb` scripts (the `write_lock` branch in `bun_executor.rs`) under 1.4's
  narrowed `trustedDependencies` matching, which no longer trusts truncated-name-hash entries —
  postinstall lifecycle scripts for those packages may stop running.
- A real two-worker skew test on one instance (re-lock on 1.4, run from a 1.3 worker). The
  isolated `bun install` measurement above predicts silent drift; it has not been reproduced
  through the full dependency-job path.

## Release note

Two things worth calling out to operators:

- **During rollout**, a bun script whose dependencies need overrides or catalogs will *hard-fail to
  deploy* until every worker is on 1.794.0, where it previously succeeded. That is deliberate — the
  alternative is silently resolving different dependency versions on the older workers — and the
  error says what to do. Note that an instance which deliberately pins a worker group to an older
  release never crosses the gate, so for them the failure is indefinite rather than transient.
- **Once the gate opens**, the next re-lock of each bun script rewrites its `bun.lock` as v2, so
  git-sync users see a lockfile format churn across the repo at that point rather than now.

### DIFF
--- a/.github/DockerfileBackendTests
+++ b/.github/DockerfileBackendTests
@@ -42,7 +42,7 @@ RUN wget https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON_VER
 RUN /usr/local/bin/python3 -m pip install pip-tools
 
 # Bun
-COPY --from=oven/bun:1.3.10 /usr/local/bin/bun /usr/bin/bun
+COPY --from=oven/bun:1.4.0 /usr/local/bin/bun /usr/bin/bun
 
 # Install windmill CLI
 RUN bun install -g windmill-cli \
--- a/.github/workflows/ai-agent-tests.yml
+++ b/.github/workflows/ai-agent-tests.yml
@@ -61,7 +61,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.10
+          bun-version: 1.4.0
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/ai-evals-test.yml
+++ b/.github/workflows/ai-evals-test.yml
@@ -75,7 +75,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.10
+          bun-version: 1.4.0
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/backend-test-windows.yml
+++ b/.github/workflows/backend-test-windows.yml
@@ -73,7 +73,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.10
+          bun-version: 1.4.0
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -58,7 +58,7 @@ jobs:
           go-version: 1.21.5
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.10
+          bun-version: 1.4.0
       - uses: actions/setup-node@v4
         with:
           node-version: "20"
--- a/.github/workflows/git-sync-test.yml
+++ b/.github/workflows/git-sync-test.yml
@@ -133,7 +133,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.10
+          bun-version: 1.4.0
 
       - uses: denoland/setup-deno@v2
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -287,7 +287,7 @@ COPY --from=windmill_duckdb_ffi_internal_builder /windmill-duckdb-ffi-internal/t
 
 COPY --from=denoland/deno:2.2.1 --chmod=755 /usr/bin/deno /usr/bin/deno
 
-COPY --from=oven/bun:1.3.10 /usr/local/bin/bun /usr/bin/bun
+COPY --from=oven/bun:1.4.0 /usr/local/bin/bun /usr/bin/bun
 
 # Install windmill CLI
 RUN bun install -g windmill-cli \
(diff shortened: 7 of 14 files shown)


rubenfiszel:
✅ Review round clean @ 692bfb4b361fba17e456185ec81bfafbdbebbcbc — nit-only verdicts at 3d7610f69a04a8c03a57e22d7a458db1ab582f37 (codex + claude "mergeable, should ideally address nits", pi "good to merge"); nits addressed in 692bfb4b36 (None-arm now warns instead of storing silently, drafting-history phrasing trimmed, stray rustfmt hunk removed). CI green on this head: build, build_ee, cargo_test, git_sync_e2e, run_integration_test all pass.

PR #10786, opened 2026-08-20
fix: make workspace preprocessor scripts selectable in flow preprocessor steps

rubenfiszel:
## Summary

A flow's preprocessor step could not pick a workspace script, and the empty picker gave no clue why.

The classic step panel (`FlowInputs.svelte`) keeps a local `kind` that stays `'script'` for the
preprocessor slot and passed it to both pickers and to the inline-template generator. Two
consequences:

- The Workspace list showed **action** scripts and never preprocessor ones, and the panel defaulted
  to the **Hub** tab even though the hub publishes no preprocessor script (`HubScriptKind` has no
  such value).
- `initialCode()` keys the preprocessor template off `kind`, not `subkind`, so "Inline new …" seeded
  a plain `main()` action stub instead of a `preprocessor(event)` one, for every language.

That panel is reached whenever `preprocessor_module` is an `identity` step (imported flows,
CLI/YAML, AI-authored flows) rather than from the "Add preprocessor step" button.

Separately, the quick insert popover filters correctly on `kind=preprocessor` but says only "No
scripts found." when a workspace has no such script. Since a workspace script is only eligible once
its **Script kind** is set to **Preprocessor**, and nothing in either picker said so, an empty list
read as a broken feature. Both pickers now say what makes a script eligible.

## Changes

- `FlowInputs.svelte`: derive `scriptKind` (`'preprocessor'` for the preprocessor slot, else `kind`)
  and key the inline template, language filter, headings, doc link and pre-made list off it. The
  preprocessor slot renders the workspace picker directly, with no Hub toggle.
- `WorkspaceScriptPicker.svelte`: accept `'preprocessor'`; show the empty state whether or not a
  search is typed (it previously rendered an empty bordered list), and add the eligibility hint.
- `WorkspaceScriptPickerQuick.svelte`: same hint under "No scripts found.".
- `FlowModuleWrapper` / `FlowModuleComponent` / `EditorBar`: widen the `scriptKind` union to carry
  `'preprocessor'` instead of laundering it through a union that excluded it, and clamp it to
  `'script'` only at `PickHubScript`, so the step editor's Library drawer keeps a populated hub tab
  while its workspace tab correctly lists preprocessor scripts.
- `FlowModuleComponent`: derive the editor bar's kind from the preprocessor slot rather than from
  `scriptKind`, which only records how a step was created this session and is back to `'script'` on
  any remount. Without this, a reloaded inline preprocessor got an action-script Library list and,
  worse, Reset Content replaced a valid `preprocessor(...)` body with a `main()` stub that cannot run
  under the preprocessor entrypoint.
- `getResetCode()`: route preprocessors through `initialCode` before the per-language short-circuits.
  Those all return a `main` stub without ever consulting `kind`, so php was the only
  preprocessor-capable language whose reset was correct; deno, bun and python3 still produced a
  `main()` body. Pinned by a case in `script_helpers.test.ts` (verified to fail without the guard).

## Screenshots

Before — preprocessor step listing **action** scripts, Hub tab default, "action script" labels:

before

After — preprocessor labels, workspace-only picker, the preprocessor script listed:

after

Empty state in the classic panel and in the quick insert popover:

classic empty

quick empty

Reloaded (remounted) inline PHP preprocessor: Library lists preprocessor scripts, and Reset Content
keeps the preprocessor entrypoint:

remount library

remount reset

Reset Content on a reloaded **Python** preprocessor, the case the per-language short-circuit missed:

reset python

Inline PHP preprocessor now seeded with the preprocessor template (was the `main()` action stub):

inline template

## Test plan

Exercised against a local instance with a `preprocessor`-kind Python script and an action script in
one workspace, plus a second workspace with neither.

- [x] Preprocessor step's "Use pre-made preprocessor script" lists the preprocessor script, not the
      action script, and shows no Hub tab
- [x] Picking it wires the step to that path with the `preprocessor` entrypoint and a working test panel
- [x] "Inline new preprocessor script" seeds the preprocessor template (checked PHP and Python)
- [x] Empty workspace shows the eligibility hint in both the classic panel and the quick popover,
      with no empty bordered list
- [x] Step editor's Library drawer: Hub tab populated, Workspace tab filtered to preprocessor
- [x] Reload a flow with a saved inline PHP preprocessor: Library's Workspace tab lists only the
      preprocessor script, and Reset Content seeds the preprocessor template rather than `main()`
- [x] Same for a saved inline Python preprocessor (php was the only language the first pass fixed)
- [x] `npx vitest run src/lib/script_helpers.test.ts`: 9 passed
- [x] No regression on action and failure steps (kind toggle, Hub/Workspace toggle, hub list intact)
- [x] `npm run check`: 0 errors, no new warnings

One test added, in `script_helpers.test.ts`: `getResetCode` is pure logic with an existing sibling
test file, and the reset regression is exactly the kind a future language addition could reintroduce.
The Svelte wiring itself stays untested per `docs/validation.md` (pure UI, covered by type-checking).

### DIFF
--- a/frontend/src/lib/components/EditorBar.svelte
+++ b/frontend/src/lib/components/EditorBar.svelte
@@ -94,7 +94,7 @@
 		// editor). `undefined` = not applicable; `false` makes the badge red
 		// even if the main function parses.
 		validAssets?: boolean | undefined
-		kind?: 'script' | 'trigger' | 'approval'
+		kind?: 'script' | 'trigger' | 'approval' | 'preprocessor'
 		template?:
 			| 'pgsql'
 			| 'mysql'
@@ -580,7 +580,13 @@
 <Drawer bind:this={scriptPicker} size="900px">
 	<DrawerContent title="Code" on:close={scriptPicker.closeDrawer}>
 		{#if pick_existing == 'hub'}
-			<PickHubScript bind:filter {kind} on:pick={onScriptPick}>
+			<!-- The hub publishes no preprocessor script, so that kind falls back to the action list
+			     rather than showing an empty hub. -->
+			<PickHubScript
+				bind:filter
+				kind={kind == 'preprocessor' ? 'script' : kind}
+				on:pick={onScriptPick}
+			>
 				<ToggleHubWorkspace bind:selected={pick_existing} />
 			</PickHubScript>
 		{:else}
--- a/frontend/src/lib/components/flows/content/FlowInputs.svelte
+++ b/frontend/src/lib/components/flows/content/FlowInputs.svelte
@@ -44,6 +44,11 @@
 					? 'approval'
 					: 'script'
 	)
+	// The preprocessor slot shows no kind toggle, so `kind` stays 'script' there. Everything that
+	// tags a script (inline template, pre-made list) must key off this, never off `kind`.
+	let scriptKind: 'script' | 'failure' | 'approval' | 'trigger' | 'preprocessor' = $derived(
+		preprocessorModule ? 'preprocessor' : kind
+	)
 	let pick_existing: 'workspace' | 'hub' = $state('hub')
 	let filter = $state('')
 
@@ -56,7 +61,7 @@
 	)
 
 	function displayLang(lang: SupportedLanguage | 'docker', kind: string) {
-		if (preprocessorModule) {
+		if (kind === 'preprocessor') {
 			return canHavePreprocessor(lang as SupportedLanguage)
 		}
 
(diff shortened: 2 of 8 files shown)


rubenfiszel:
✅ Review round clean @ 7e7973400bc7f3fd1a48846eee288196b12f289e — nit-only verdicts at 5764380c7ee84fd5b253e795e57fc36c6d5e0bc8 (claude and pi: Good to merge; codex: nit); nit addressed in 7e7973400b, a comment-only change with no behaviour delta.

PR #10788, opened 2026-08-20
fix: make named and namespace npm imports work under //nodejs

### LINKED ISSUE #10785: bug: //nodejs scripts mis-bundle some npm import shapes (lodash named/namespace imports)
## Summary

Under the `//nodejs` annotation, certain npm import shapes produce a broken bundle. The two known failures both involve `lodash`, and they fail in two *different* ways — one corrupts the generated bundle, the other produces a bad namespace object.

This is **not** version-specific: it reproduces identically on bun 1.3.10 / node v20.20.2 (app.windmill.dev) and on bun 1.3.11 and 1.4.0 / node v26.4.0 locally.

## Reproduction

New script, language TypeScript (Bun), then Test:

```ts
//nodejs
import { chunk } from "lodash";
export async function main() { return chunk([1, 2, 3, 4], 2); }
```

**Actual:**

```
exit code for "bun run": 1, last log lines:
…await B.writeFile("result.json",JSON.stringify(D)),process.exit(1)} = pkg;
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:213:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:320:5)
Node.js v20.20.2
```

The whole minified program is emitted as the left-hand side of an assignment ending `= pkg;`, which is not valid — node's ESM loader throws at instantiate time.

**Expected:** `[[1,2],[3,4]]`.

Live job on app.windmill.dev: https://app.windmill.dev/run/01a020f2-7e71-0e82-8ef6-19324905b09c?workspace=user-test

### Second, distinct failure — namespace import

```ts
//nodejs
import * as _ from "lodash";
export async function main() { return _.chunk([1, 2], 1); }
```

Bundles successfully, then fails at runtime with `TypeError: G.chunk is not a function`.
Job: https://app.windmill.dev/run/01a020f3-411a-a66b-5646-b394af9636e7?workspace=user-test

## Scope — what works vs what doesn't

Measured on app.windmill.dev. It is **not** "any npm dep breaks `//nodejs`", and CJS-vs-ESM does not explain it (mime-types is CJS and works):

| script (all prefixed `//nodejs`) | result |
|---|---|
| `import { chunk } from "lodash"` | ❌ bundle ends `= pkg;`, ESM loader rejects |
| `import * as _ from "lodash"` | ❌ `TypeError: G.chunk is not a function` |
| `import _ from "lodash"` (default) | ✅ |
| `import { lookup } from "mime-types"` (CJS, named) | ✅ |
| `import dayjs from "dayjs"` | ✅ |
| `import { z } from "zod"` | ✅ |
| `import { v4 } from "uuid"` | ✅ |
| no imports, or only `node:*` | ✅ |
| the same lodash named import **without** `//nodejs` (plain bun) | ✅ |

## Where to look

`LoaderMode::Node` in `backend/windmill-worker/src/bun_executor.rs` writes `node_builder.ts`, which bundles the wrapper with:

```js
Bun.build({
  entrypoints: ["<job_dir>/wrapper.mjs"],
  outdir: "./",
  target: "node",
  plugins: [p],                       // windmill relative-resolver
  external: fileNames,                // dir names under node_modules
  minify: true,
})
```

The passing and failing cases differ precisely in the CJS named-export interop the bundler has to synthesize, so the `external` + `minify` + resolver-plugin interaction is the place to start. Worth checking whether dropping `minify` or narrowing `external` changes the output.

## Notes

Found while e2e-testing the bun 1.4 upgrade (#10784) and explicitly ruled out as a regression from that bump — it predates it and is unrelated.

rubenfiszel:
## Summary

Fixes #10785.

Under the `//nodejs` annotation the wrapper is bundled with `Bun.build({ target: "node", external: <dir names under node_modules> })`, so npm packages stay out of the bundle and **node**, not bun, resolves them at runtime. Node's ESM loader only exposes the named exports of a CommonJS package that `cjs-module-lexer` can detect statically, and lodash & co. assemble `module.exports` while they run. So:

- `import { chunk } from "lodash"` never instantiated. Node's diagnostic prints the offending source line — the whole minified bundle, on one line — followed by its `import pkg from 'lodash'; const {chunk} = pkg;` hint, which is what made the log look like a bundle ending in `= pkg;`.
- `import * as _ from "lodash"` instantiated but gave a namespace holding nothing but `default`, hence `TypeError: _.chunk is not a function`.

The bundle now imports such a package as a namespace and reads the names off `default` (i.e. `module.exports`), which is exactly what node's own error message suggests doing by hand.

Only packages node loads as CommonJS are rewritten — the classification follows node's own rule (the resolved file's extension, and for `.js` the `type` of the closest `package.json`) — so an ESM package's named imports stay the live bindings node gives them. Default-only imports (`import _ from "lodash"`) and side-effect imports are left untouched as well, since node already resolves those to `module.exports`; that keeps the clear instantiation error when an ESM package has no `default` export.

## Status

**Left in draft, and duplicated by #10792.** That PR fixes the same issue in the same `build_loader` `LoaderMode::Node` branch by routing CommonJS packages through a generated `.cjs` shim so bun synthesizes the interop, rather than rewriting the emitted bundle as this PR does. The two conflict; only one should land. The trade-off: this PR never changes which file node loads (the `import` condition still picks the entry) at the cost of a post-pass over the bundle, while #10792 is much smaller but switches shimmed packages to their `require` entry.

Independently of that, this touches shared worker infrastructure (the bun executor's node loader, on the path of every `//nodejs` job), so it wants a human look before it goes ready rather than an unattended flip.

Review rounds: the round at `b0152ef` was clean — claude and pi "Good to merge", codex nit-only (JSON modules classified as CommonJS), which `8de64d4` fixes.

## Changes

- `backend/windmill-worker/node_cjs_interop.js` (new): rewrites the import statements of external CommonJS specifiers in the emitted bundle into a namespace import plus `k in ns ? ns[k] : ns.default?.[k]` lookups; namespace imports get a proxy that falls through to `default` for keys the namespace lacks. Import statements are located on a copy of the bundle whose string/template/comment/regex literals are blanked out (offset-preserving, so generated code carrying an import statement inside a string is not rewritten), and a `Bun.Transpiler.scanImports` cross-check leaves the bundle untouched if the statements found don't match the parsed ones.
- `bun_executor.rs`: `LoaderMode::Node`'s `node_builder.ts` applies that rewrite to `wrapper.js` after `Bun.build`. A failure to rewrite logs and keeps the original bundle rather than failing the job. This is also the loader used by `//nodejs` dedicated workers and by `//nodejs //nobundling`; the deployed-script `LoaderMode::NodeBundle` path bundles its dependencies and was never affected.
- `backend/tests/bun_jobs.rs`: regression test that runs the real `LoaderMode::Node` loader over a CommonJS package whose exports only exist once it has run, then executes the bundle with node. It fails without the fix with the exact `Named export 'greet' not found` error from the issue.

## Test plan

- [x] `//nodejs` + `import { chunk } from "lodash"` returns `[[1,2],[3,4]]` as a real preview job
- [x] `//nodejs` + `import * as _ from "lodash"` returns `[[1],[2]]`
- [x] every passing case in the issue's matrix still passes as a real job: lodash default, `mime-types` named, `dayjs` default, `zod` named, `uuid` named, no imports, `node:*` only, and the same lodash named import without `//nodejs`
- [x] `import * as wmill from "windmill-client"` and `import { setState, getState } from "windmill-client"` still work under `//nodejs`
- [x] deployed `//nodejs` script (bundle cache path) and deployed `//nodejs //nobundling` script both run
- [x] ESM packages unaffected: `nanoid` named + namespace, `chalk` default, and an ESM package that mutates an exported binding still reads live through a named import (0 → frozen without the CommonJS gate, correct with it)
- [x] scoped packages fixed, `lodash/fp.js` subpath fixed
- [x] a script holding `;import{y}from"lodash";` inside a string literal is rewritten correctly rather than corrupted
- [x] `cargo test --test bun_jobs` — 35 passed

---
## Summary by cubic
Fixes named and namespace npm imports under `//nodejs` by rewriting external CommonJS imports to a namespace import with a `default` fallback and explicit missing‑export errors. Previously, named imports from CJS packages like `lodash` failed and namespace imports exposed only `default`; now both work without changing default‑only or side‑effect imports.

- Runs post‑bundle in `LoaderMode::Node` only: rewrites `wrapper.js` for externals Node loads as CommonJS; `LoaderMode::NodeBundle` is unchanged.
- Classifies targets with Node’s resolver (`import.meta.resolve`) and Node’s format rules (extensionless entries follow the nearest `package.json` `type`); falls back to Bun when needed; skips Node builtins; never rewrites JSON or other non‑CommonJS module formats.
- Preserves import attributes. CJS named bindings are snapshots; namespace imports merge keys from the namespace and `default`. Missing named exports throw a `SyntaxError` at module init.
- Masks literals and cross‑checks with `Bun.Transpiler.scanImports`; on mismatch, logs and skips to avoid corrupting the bundle.
- Adds a regression test for dynamic CJS exports, live ESM export mutation, conditional exports that hand Bun CJS but Node ESM, and an extensionless ESM entry.

<sup>Written for commit 8de64d4c4a6bc00a84ef75eb9ae48553522f7123. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -1284,8 +1284,10 @@ fn test_generate_bun_bundle_propagates_exit_status() {
 /// see the named exports of a CommonJS package that builds `module.exports`
 /// dynamically (lodash & co.), so a named import fails to instantiate and a
 /// namespace import yields nothing but `default`. The bundle must import such a
-/// package as a namespace and read the names off `default` instead — while
-/// leaving an ESM package's named imports as the live bindings node gives it.
+/// package as a namespace and read the names off `default` instead, while
+/// leaving alone anything node loads as ESM — including a package whose
+/// conditional exports hand bun a different file than they hand node — so its
+/// named imports stay the live bindings node gives them.
 #[test]
 fn test_node_loader_cjs_named_export_interop() {
     use std::process::Command;
@@ -1294,39 +1296,63 @@ fn test_node_loader_cjs_named_export_interop() {
     let temp_dir = tempfile::tempdir().unwrap();
     let dir = temp_dir.path();
     let dir_str = dir.to_str().unwrap();
+    let write_pkg = |name: &str, files: &[(&str, &str)]| {
+        let pkg_dir = dir.join("node_modules").join(name);
+        std::fs::create_dir_all(&pkg_dir).unwrap();
+        for (file, content) in files {
+            std::fs::write(pkg_dir.join(file), content).unwrap();
+        }
+    };
 
     // A CommonJS package whose exports only exist once it has run, which is
     // what defeats the lexer.
-    let pkg_dir = dir.join("node_modules").join("dyn-cjs-pkg");
-    std::fs::create_dir_all(&pkg_dir).unwrap();
-    std::fs::write(
-        pkg_dir.join("package.json"),
-        r#"{ "name": "dyn-cjs-pkg", "version": "1.0.0", "main": "index.js" }"#,
-    )
-    .unwrap();
-    std::fs::write(
-        pkg_dir.join("index.js"),
-        r#"
+    write_pkg(
+        "dyn-cjs-pkg",
+        &[
+            (
+                "package.json",
+                r#"{ "name": "dyn-cjs-pkg", "version": "1.0.0", "main": "index.js" }"#,
+            ),
+            (
+                "index.js",
+                r#"
 const api = {};
 ["greet"].forEach((k) => { api[k] = (s) => k + " " + s; });
 module.exports = api;
 "#,
-    )
-    .unwrap();
+            ),
+        ],
+    );
 
     // An ESM package whose exported binding changes after evaluation.
-    let esm_dir = dir.join("node_modules").join("live-esm-pkg");
-    std::fs::create_dir_all(&esm_dir).unwrap();
-    std::fs::write(
-        esm_dir.join("package.json"),
-        r#"{ "name": "live-esm-pkg", "version": "1.0.0", "type": "module", "main": "index.js" }"#,
-    )
-    .unwrap();
-    std::fs::write(
-        esm_dir.join("index.js"),
-        "export let count = 0;\nexport function bump() { count++; }\n",
-    )
-    .unwrap();
+    write_pkg(
+        "live-esm-pkg",
+        &[
+            (
+                "package.json",
+                r#"{ "name": "live-esm-pkg", "version": "1.0.0", "type": "module", "main": "index.js" }"#,
+            ),
+            ("index.js", "export let count = 0;\nexport function bump() { count++; }\n"),
+        ],
+    );
+
+    // Conditional exports that give bun CommonJS and node ESM: only node's
+    // answer says whether the bundle may snapshot the bindings.
+    write_pkg(
+        "dual-cond-pkg",
+        &[
+            (
+                "package.json",
+                r#"{ "name": "dual-cond-pkg", "version": "1.0.0",
+                     "exports": { ".": { "bun": "./bun-cjs.js", "node": "./node-esm.mjs", "default": "./node-esm.mjs" } } }"#,
+            ),
+            ("bun-cjs.js", "module.exports = { count: 0, bump() {} };\n"),
+            (
+                "node-esm.mjs",
+                "export let count = 0;\nexport function bump() { count++; }\n",
+            ),
+        ],
+    );
 
     std::fs::write(
         dir.join("main.ts"),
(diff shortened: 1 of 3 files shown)


rubenfiszel:
Duplicate effort: #10792 fixes the same issue (#10785) in the same `build_loader` `LoaderMode::Node` branch, with a different mechanism — a resolver plugin routing CommonJS packages through a generated `.cjs` shim so bun synthesizes the interop, instead of this PR's post-pass over the emitted bundle. Only one of the two should land; they conflict textually and semantically.

PR #10792, opened 2026-08-21
fix: make //nodejs scripts import CommonJS npm packages correctly

rubenfiszel:
## Summary

Fixes GIT-975.

Under the `//nodejs` annotation, `import { chunk } from "lodash"` produced a bundle node's ESM loader rejects at instantiate time, and `import * as _ from "lodash"` produced a namespace object without the package's methods (`TypeError: _.chunk is not a function`).

Root cause: `LoaderMode::Node` passed `external: <every top-level directory name under <job_dir>/node_modules>` to `Bun.build`, so a bare import of an npm package survived verbatim into the ESM output that node then has to resolve at runtime. Node can only expose the named exports of a CommonJS dependency that `cjs-module-lexer` finds statically, and lodash's UMD wrapper defeats it, so the named import fails to link and the namespace object carries only `default`. `mime-types` worked because the lexer *can* see its `exports.x = ...` assignments, which is why the failure looked package-specific rather than CJS-wide.

The fix replaces the `external` option with a bun plugin that decides, per `node_modules` package, whether a CommonJS shim is provably safe:

- a package proven **CommonJS** is routed through a generated `.cjs` shim, so bun synthesizes the interop itself while the package is still `require`d from `<job_dir>/node_modules` at runtime rather than bundled;
- **everything else** stays a plain external, byte-for-byte the shape that shipped before this PR.

"Proven" means node's `import` and `require` land on the same CommonJS file no matter which export conditions are live. Since `--conditions` in a job's `NODE_OPTIONS` can enable a condition of any name, no denylist can be complete, so **any** condition disqualifies a package: only subpath keys and `default` are permitted anywhere in `exports`, plus packages with no `exports` at all, which resolve through `main`, where conditions never reach. Three review rounds each found a different condition that selected a different entry, so the rule is categorical rather than enumerated. The worst case is "not fixed for that package", never "a package that worked now breaks".

Nothing is newly bundled into the output, so native addons and packages that read their own files at runtime keep working the same way.

## Changes

- `backend/windmill-worker/node_externals.bun.js` (new) — the `windmill-node-externals` bun plugin: leaves every `node_modules` package external except the ones it can prove CommonJS, which resolve to a generated `<job_dir>/.wm_node_cjs/*.cjs` shim. CommonJS-ness comes from the resolved file extension plus the same nearest-`package.json` `type` walk node performs, and only for manifests whose entry cannot vary by condition. It lives in its own file, interpolated like the existing `RELATIVE_BUN_BUILDER`, so the rule can be unit-tested and so the block stops needing `format!` brace escaping.
- `backend/windmill-worker/src/bun_executor.rs` — `build_loader`'s `LoaderMode::Node` branch drops `external: fileNames` and registers that plugin instead.
- `backend/tests/bun_jobs.rs` — a job-level regression test (`//nodejs` script with a named and a namespace import of lodash, plus a named import from ESM-only `nanoid` to pin the stays-external branch), and a unit test running `entryDependsOnCondition` under bun against a table of manifest shapes.

## Test plan

- [x] `cargo test --test bun_jobs` — 36/36 pass; the new `test_bun_nodejs_cjs_named_and_namespace_import` fails on `main`'s builder and passes with the fix.
- [x] `cargo test --features quickjs,private,enterprise --test dedicated_workers test_dedicated_flow_bun_nodejs` — the dedicated-worker `//nodejs` path also builds through `LoaderMode::Node`.
- [x] Real jobs on a local instance, before and after. Before: `import { chunk } from "lodash"` fails with the reported `… = pkg;` ESM-loader error, `import * as _ from "lodash"` fails with `TypeError: G.chunk is not a function`. After: `[[1,2],[3,4]]` and `[[1],[2]]`.
- [x] Real jobs for the shapes that already worked: lodash default and `lodash/chunk.js` subpath, `mime-types`, `dayjs`, `zod`, `uuid`, `axios`, ESM-only `nanoid`, `windmill-client` (namespace and named imports, plus a real `setVariable`/`getVariable` roundtrip), a `/f/...`-imported windmill script that itself imports lodash, `//nodejs //nobundling`, and a deployed `//nodejs` script (the `NodeBundle` path, unchanged).
- [x] A 25-shape matrix built through the real generated builder and executed under node, across **node v20.20.0** (the managed instance's version, no `require(esm)`) and **v26.4.0**, with **bun 1.3.11** and **bun 1.4.0** (the version #10784 upgrades to) as the builder. All pass in every combination.
- [x] Fixtures for each way the classifier can be fooled, each compared against `main`: a `{"bun": …, "node": …}` package (shimming it gave `ERR_REQUIRE_ASYNC_MODULE`), a `{"import": "./import.cjs", "require": "./require.cjs"}` package whose entries return different values (shimming it silently returned the `require` entry), a `{"development": "./dev.mjs", "default": "./index.cjs"}` package run under `NODE_OPTIONS=--conditions=development` (shimming it gave `ERR_REQUIRE_ASYNC_MODULE`), a CJS subpath long enough that the flattened shim basename exceeded the filesystem limit (`ENAMETOOLONG`, a hard build failure), a symlinked `node_modules` package, and `main` (CJS) + `module` (ESM) with no `exports`. Every one now matches or improves on `main`.
- [x] Dynamic `await import("lodash")` — also broken before this change, fixed by it.

Not verified: a `//nodejs` job on a **Windows** worker. The generated builder transpiles against `loader.bun.windows.js` and the path handling is drive-letter safe by construction, but no job was run there.

## Known limitation: a shimmed package is invisible to async ESM loader hooks

A shimmed package is reached by `require()`, so node's **async** ESM hooks
(`module.register()` / `--experimental-loader`, reachable because a script's `envs` can set
`NODE_OPTIONS` off `CLOUD_HOSTED`) no longer see it. Measured on node v26.4.0 with a hook
redirecting `lodash`, against `import _ from "lodash"` built both ways:

| interception mechanism | before | after |
| -- | -- | -- |
| `register()` / `--experimental-loader` (async, ESM-only) | intercepted | **not intercepted** |
| `module.registerHooks()` (sync, node >= 22.15) | intercepted | intercepted |
| `--require` + `Module.prototype.require` patch (what APM auto-instrumentation uses) | intercepted | intercepted |

This is a property of the mechanism, not an oversight, and neither remedy is available:

- **Preserving ESM resolution cannot fix the bug.** Node's ESM loader cannot expose lodash's
  named exports at all — that *is* the root cause. Any fix must obtain the CommonJS exports
  object through `require`, and node does not route `require` through async ESM hooks by design.
- **Refusing the shim when hooks are configured is not soundly detectable.** The bundler runs
  with `env_clear()` and only `common_bun_proc_envs` (`bun_executor.rs:1032-1037`), so it never
  sees the job's `NODE_OPTIONS`; and the bundle cache key is content + lock + transitive import
  versions with no env component (`bun_executor.rs:1367`), so a bundle built for a job without
  `NODE_OPTIONS` is reused verbatim by a later job that sets it. A build-time gate would be
  silently stale — the same class of bug rounds 1-3 were about.

Narrowing the shim to only the import shapes that are broken today (leaving default-only imports
external, so nothing that works today changes) would bound this to scripts that currently crash.
It needs an import-clause classifier — `Bun.Transpiler.scan()` reports specifiers but not clauses,
so it would be hand-rolled — and a mis-parse re-breaks the bug on the common path to protect an
exotic one. Recorded as a deliberate call for the human review this PR is waiting on, not done.

**Left in draft deliberately.** Round 6 is clean on all three reviewers (Codex, Claude, Pi all "good to merge" at `d767b1c3d0`), round 5's Codex P1 being the loader-hook trade-off documented above. It is left in draft anyway because this changes how every non-cached `//nodejs` job links its npm dependencies, which is shared worker infrastructure rather than a self-contained fix. Three earlier rounds each found a real regression in this same classifier, so it wants a human look before it goes ready rather than an automatic flip on a clean round.

### DIFF
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -540,7 +540,9 @@ export function main() {
 }
 
 /// Node's ESM loader cannot see the named exports of a CommonJS package that cjs-module-lexer
-/// fails to analyze, so `//nodejs` scripts must not leave such packages as plain externals.
+/// fails to analyze, so `//nodejs` scripts must not leave such packages as plain externals. The
+/// ESM-only import guards the other half: those must stay external, since `require`ing one throws
+/// on the node versions without require(esm).
 #[sqlx::test(fixtures("base"))]
 async fn test_bun_nodejs_cjs_named_and_namespace_import(db: Pool<Postgres>) -> anyhow::Result<()> {
     initialize_tracing().await;
@@ -550,9 +552,10 @@ async fn test_bun_nodejs_cjs_named_and_namespace_import(db: Pool<Postgres>) -> a
     let content = r#"//nodejs
 import { chunk } from "lodash";
 import * as lodash from "lodash";
+import { nanoid } from "nanoid";
 
 export function main() {
-    return [chunk([1, 2, 3, 4], 2), lodash.chunk([1, 2], 1)];
+    return [chunk([1, 2, 3, 4], 2), lodash.chunk([1, 2], 1), typeof nanoid()];
 }
 "#
     .to_owned();
@@ -578,7 +581,10 @@ export function main() {
         .json_result()
         .unwrap();
 
-    assert_eq!(result, serde_json::json!([[[1, 2], [3, 4]], [[1], [2]]]));
+    assert_eq!(
+        result,
+        serde_json::json!([[[1, 2], [3, 4]], [[1], [2]], "string"])
+    );
     Ok(())
 }
 
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -2515,6 +2515,65 @@ export function main(name: string) {
     Ok(())
 }
 
+/// A shimmed package is reached by `require()`, so the shim is only safe when nothing can make
+/// node's `require` land on a different entry than the `import` the classifier looked at. Bun's
+/// own conditions, `import`/`require`, and `--conditions` in the job's NODE_OPTIONS can each do
+/// that, so any condition at all disqualifies a manifest.
+#[test]
+fn test_node_externals_refuses_conditional_exports() {
+    use std::process::Command;
+    use windmill_worker::{BUN_PATH, NODE_EXTERNALS_PLUGIN};
+
+    let temp_dir = tempfile::tempdir().unwrap();
+    let dir = temp_dir.path();
+    let plugin = NODE_EXTERNALS_PLUGIN.replace("JOB_DIR", &dir.to_string_lossy());
+    std::fs::write(dir.join("externals.js"), plugin).unwrap();
+    std::fs::write(
+        dir.join("probe.js"),
+        r#"
+import { entryDependsOnCondition } from "./externals.js";
+const manifests = JSON.parse(await Bun.file("manifests.json").text());
+console.log(JSON.stringify(manifests.map((m) => entryDependsOnCondition(m))));
+"#,
+    )
+    .unwrap();
+    // Every entry a shimmable package may have, then every way one can vary by condition.
+    let manifests = serde_json::json!([
+        serde_json::Value::Null,
+        "./index.js",
+        { ".": "./index.js" },
+        { ".": { "default": "./index.js" }, "./sub": "./sub.js" },
+        { ".": { "bun": "./bun.js", "default": "./index.js" } },
+        { ".": { "import": "./esm.js", "require": "./cjs.js" } },
+        { ".": { "development": "./dev.js", "default": "./index.js" } },
+        { "./sub": { "node": "./node.js", "default": "./index.js" } },
+        { ".": [{ "browser": "./browser.js" }, "./index.js"] },
+    ]);
+    std::fs::write(
+        dir.join("manifests.json"),
+        serde_json::to_string(&manifests).unwrap(),
+    )
+    .unwrap();
+
+    let output = Command::new(BUN_PATH.as_str())
+        .args(["run", "probe.js"])
+        .current_dir(dir)
+        .output()
+        .expect("failed to run bun");
+    assert!(
+        output.status.success(),
+        "probe failed: {}",
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    let verdicts: Vec<bool> =
+        serde_json::from_slice(String::from_utf8_lossy(&output.stdout).trim().as_bytes()).unwrap();
+    assert_eq!(
+        verdicts,
+        vec![false, false, false, false, true, true, true, true, true]
+    );
+}
+
 /// Tests for RELATIVE_BUN_BUILDER (loader_builder.bun.js)
 /// These tests verify Bun's behavior for import scanning and package.json generation.
 /// Purpose: Catch regressions when upgrading Bun versions.
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -539,6 +539,49 @@ export function main() {
     Ok(())
 }
 
+/// Node's ESM loader cannot see the named exports of a CommonJS package that cjs-module-lexer
+/// fails to analyze, so `//nodejs` scripts must not leave such packages as plain externals.
+#[sqlx::test(fixtures("base"))]
+async fn test_bun_nodejs_cjs_named_and_namespace_import(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+    let server = ApiServer::start(db.clone()).await?;
+    let port = server.addr.port();
+
+    let content = r#"//nodejs
+import { chunk } from "lodash";
+import * as lodash from "lodash";
+
+export function main() {
+    return [chunk([1, 2, 3, 4], 2), lodash.chunk([1, 2], 1)];
+}
+"#
+    .to_owned();
+
+    let job = JobPayload::Code(RawCode {
+        hash: None,
+        content,
+        path: None,
+        language: ScriptLang::Bun,
+        lock: None,
+        concurrency_settings: windmill_common::runnable_settings::ConcurrencySettings::default()
+            .into(),
+        debouncing_settings: windmill_common::runnable_settings::DebouncingSettings::default(),
+        cache_ttl: None,
+        cache_ignore_s3_path: None,
+        dedicated_worker: None,
+        modules: None,
+        tag: None,
+    });
+
+    let result = run_job_in_new_worker_until_complete(&db, false, job, port)
+        .await
+        .json_result()
+        .unwrap();
+
+    assert_eq!(result, serde_json::json!([[[1, 2], [3, 4]], [[1], [2]]]));
+    Ok(())
+}
+
 #[sqlx::test(fixtures("base"))]
 async fn test_bun_nobundling_mode(db: Pool<Postgres>) -> anyhow::Result<()> {
     initialize_tracing().await;
--- a/backend/windmill-worker/node_externals.bun.js
+++ b/backend/windmill-worker/node_externals.bun.js
@@ -89,7 +89,9 @@ function cjsShim(specifier) {
       ".cjs";
     mkdirSync(cjsShimDir, { recursive: true });
     // The local binding is load-bearing: bun collapses a bare `module.exports = require(x)` back
-    // into a passthrough external import, which is the shape that breaks node.
+    // into a passthrough external import, which is the shape that breaks node. A shimmed package
+    // is reached by `require`, so node's async ESM hooks (`register()`) no longer see it; its
+    // sync `registerHooks()` and `--require` interception still do.
     writeFileSync(
       shim,
       "const mod = require(" + JSON.stringify(specifier) + ");\nmodule.exports = mod;\n"
--- a/backend/windmill-worker/node_externals.bun.js
+++ b/backend/windmill-worker/node_externals.bun.js
@@ -10,13 +10,10 @@ try {
   installedPackages = readdirSync(nodeModulesDir);
 } catch (e) {}
 
-// The shim reaches the package through require(), the classifier through bun's resolution, and the
-// two only ever agree on a manifest that cannot select a different entry for either of them. Which
-// conditions are live is not knowable here: bun applies its own, `import`/`require` split node's
-// two ways in, and `--conditions` in the job's NODE_OPTIONS adds arbitrary more at runtime. So a
-// condition of any name disqualifies the package. Subpaths and `default` always select the same
-// entry, and a package with no `exports` at all resolves through `main`, which conditions never
-// reach.
+// The shim reaches the package by require() and the classifier by bun's resolution, which agree
+// only where no condition can steer them apart. Which conditions are live is unknowable here (bun
+// applies its own, and NODE_OPTIONS `--conditions` adds any name at runtime), so any condition at
+// all disqualifies the package; subpaths and `default` select one entry whatever is live.
 export function entryDependsOnCondition(exports) {
   if (Array.isArray(exports)) {
     return exports.some(entryDependsOnCondition);
@@ -103,11 +100,9 @@ function cjsShim(specifier) {
 }
 
 // Node only sees the named exports of a CommonJS dependency that cjs-module-lexer finds
-// statically, which fails on packages such as lodash, so leaving those as plain externals breaks
-// `import { x } from "pkg"` and `import * as pkg from "pkg"`. A generated CommonJS shim makes bun
-// synthesize the interop while the package itself is still required at runtime. Only a package
-// proven CommonJS gets one: requiring an ESM entry throws on the node versions without
-// require(esm), so everything else keeps the plain external it would have had.
+// statically, which fails on lodash and the like, so a plain external breaks `import { x }` and
+// `import * as x` from it. Routing it through a CommonJS shim makes bun synthesize the interop.
+// Only proven-CommonJS packages get one: requiring an ESM entry throws without require(esm).
 export const nodeExternals = {
   name: "windmill-node-externals",
   setup(build) {
--- a/backend/windmill-worker/node_externals.bun.js
+++ b/backend/windmill-worker/node_externals.bun.js
@@ -1,0 +1,127 @@
+import { dirname } from "node:path";
+import { mkdirSync, readFileSync, readdirSync, writeFileSync } from "node:fs";
+
+const jobDir = "JOB_DIR";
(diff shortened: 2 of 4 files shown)


rubenfiszel:
Round 1 addressed in f041852fbb. Thanks — the Codex P1 was real and I reproduced it before fixing.

**[P1, codex] Classification followed bun's export conditions, not node's.** Confirmed with a synthetic package:

```json
{ "exports": { ".": { "bun": "./bun.cjs", "node": "./node.mjs", "default": "./node.mjs" } } }
```

`Bun.resolveSync` returns `bun.cjs`, so the old code classified it CommonJS and shimmed it, while node resolves `node.mjs`. With top-level await in that entry the job died with exactly the predicted `ERR_REQUIRE_ASYNC_MODULE` — on node 20 **and** node 26 — where the plain external had worked.

Fixed by refusing to classify from bun's resolution whenever the manifest can steer bun somewhere node will not follow: a `bun`/`bun-macro` condition anywhere in `exports`, or a top-level `bun` field. The classifier is now tri-state (`cjs` / `esm` / `unknown`) and only `cjs` gets a shim; `unknown` keeps the plain external it had before, so the fallback is always the pre-PR behaviour. The fixture now returns `"from-node-esm"` on both node versions.

I deliberately denied only the conditions bun applies and node does not, rather than allowlisting node's. `browser`, `deno`, `worker` and friends appear in a lot of real manifests and are applied by *neither* resolver here, so allowlisting would have dropped the shim (and logged) for a large share of packages that the PR is meant to fix. Within the conditions both apply, "bun says CJS" implies the shim's `require` lands on a CommonJS entry, which is the direction pi worked through independently.

**[P2, claude] Silent unknown → external fallback.** The `unknown` branch now logs to the job's build output:

```
could not tell whether '@wmtest/dual-cond' is CommonJS or ESM, leaving it external: named imports from it may not resolve under node
```

Logged once per specifier, only on `unknown` — a correctly-classified ESM package stays quiet.

**[P2, local reviewer] Shim rewritten once per import statement.** Correct; `mkdirSync`/`writeFileSync` sat outside the memo. Shim generation is memoized on the specifier now — three `import ... from "lodash"` statements in one script produce one file instead of three.

**Test coverage — the ESM-stays-external branch.** Both claude and the local reviewer asked for this, so the regression test now also imports a named export from ESM-only `nanoid` and asserts on it. Worth being precise about what it buys: on node ≥ 22.12 `require(esm)` succeeds for a TLA-free package, so a misclassification there would still pass. It is a real guard on older runtimes (the managed instance runs node 20, where it would be `ERR_REQUIRE_ESM`), which is why it is worth the one line, but the stronger evidence is the matrix below.

**Re-verification after the change** — 21 import shapes through the real generated builder, each built and then executed under node, against **node v20.20.0** (no `require(esm)`) and **node v26.4.0**, and with both **bun 1.3.11** and **bun 1.4.0** as the builder. All pass in every combination: lodash named/namespace/default/subpath, `mime-types`, `dayjs`, `zod`, `uuid`, dual `axios` (default + named), ESM-only `nanoid`/`node-fetch`/`chalk`, scoped CJS and scoped ESM packages, the `bun`-condition fixture above, dynamic `await import(...)` of both a CJS and an ESM package, builtin-only and import-free scripts. Plus `cargo test --test bun_jobs` (35/35) and the real-worker runs from the PR body re-run on this head.

**Windows** remains verified only to the extent that the generated `node_builder.ts` transpiles against `loader.bun.windows.js` and the path handling is drive-letter safe by construction (forward-slashed `job_dir`, importer normalised before the `/.wm_node_cjs/` test, `dirname` walk terminating on `C:\`). I have no Windows worker to run a job on, so that path is unexercised rather than confirmed.

rubenfiszel:
Duplicate effort: #10788 fixes the same issue (#10785) in the same `build_loader` `LoaderMode::Node` branch, by rewriting the named/namespace imports of CommonJS packages in the emitted bundle instead of shimming them through `require`. Only one of the two should land.

rubenfiszel:
Round 2 addressed in 07e7c8ccd2. The Codex P1 was real again, and worse than the write-up suggested — thanks.

**[P1, codex] The shim switched conditional exports from the import entry to the require entry.** Reproduced with the suggested fixture, `"exports": {".": {"import": "./import.cjs", "require": "./require.cjs"}}`, where the two entries return different values:

| | `import m from "@wmtest/dual-cjs"; m.which()` |
|---|---|
| `main` | `"import-entry"` |
| f041852 (round-2 head) | `"require-entry"` ❌ |
| 07e7c8cc | `"import-entry"` |

So it was not merely a hypothetical: a default import silently returned a value from the wrong entry, with no error anywhere. That is worse than the named-import breakage this PR set out to fix.

**I did not take the "load the already-resolved import target" option, because it breaks a different thing.** Requiring the absolute path `Bun.resolveSync` returned looks like it preserves the entry exactly, but `resolveSync` realpaths through symlinks — for a symlinked `node_modules` entry it returns the store path, not the path under `<job_dir>/node_modules`. The wrapper runs as `node --preserve-symlinks wrapper.mjs` (`bun_executor.rs:2477`), under which a module's dependencies resolve relative to the path it was loaded by, so a shim requiring the realpath would resolve the package's transitive deps from the store instead of the job dir. That is exactly the failure `test_bun_nobundling_transitive_require` exists to pin. Requiring the *specifier* is the correct thing under `--preserve-symlinks`, and I added a symlinked-package fixture to the matrix to keep that honest.

So I took the other option: the classifier now refuses any package whose entry depends on which condition asks for it. One predicate covers all three ways that can happen — `bun`/`bun-macro` (bun applies, node never does) and `import`/`require` (node's own two ways in). The shim is only ever generated when node's `import` and `require` provably land on the same CommonJS file, which is the property the shim needs and the one that was missing.

This costs some coverage: packages using `import`/`require` conditions no longer get a shim. It costs less than it looks like, because the classic dual package has an ESM `import` branch, which was already classified `esm` and left external. What still gets fixed is the shape the bug actually lives in — CommonJS packages with no `exports` map, or an `exports` map with no condition split. lodash, `mime-types`, and every fixture in the matrix below still resolve.

**[P2, claude] The warning fired on healthy packages.** Right, and the divergence rule above would have made it much worse — `axios`, `zod` and most modern manifests would now print it on every successful build. The deliberate refusal is silent; the log survives only on the genuine-failure path (unreadable manifest, `resolveSync` throwing), reworded to match:

```
could not inspect 'pkg' to pick its module format, leaving it external: <error>
```

A build of a script importing lodash, axios, nanoid, zod and the `bun`-condition fixture together now prints nothing and generates exactly one shim.

**[observation, claude] `main` (CJS) + `module` (ESM) with no `exports`.** Checked it, and the concern does not hold: `Bun.resolveSync` returns the `main` entry, not `module`. This shape is therefore *fixed* by the PR rather than silently missed — `main` fails with the lexer error, this head returns the value from the CJS entry, which is the same file node's `import` loads. Added as a fixture.

**Matrix re-run on this head** — 24 shapes, each built through the real generated builder and executed under node, on **node v20.20.0** and **v26.4.0**, with **bun 1.3.11** and **1.4.0** as builder. New since round 1: the divergent-CJS-conditions fixture, a symlinked `node_modules` package, and the `main`+`module` fixture. All pass. Plus `cargo test --test bun_jobs` (35/35) and the real-worker runs re-run on a backend rebuilt from this commit.

rubenfiszel:
Round 3 addressed in 85ac33ec73. Both Codex findings reproduced on `07e7c8cc` before fixing.

**[P1, codex] Runtime-only export conditions.** Reproduced: a package with `{"development": "./dev.mjs", "default": "./index.cjs"}` was shimmed from the `default` entry, and under `NODE_OPTIONS=--conditions=development` the shim's `require()` hit the ESM entry and died with `ERR_REQUIRE_ASYNC_MODULE`, where `main` returns `"development-esm"` fine.

This is the third round in a row where a condition I had not enumerated turned out to select a different entry, so I stopped enumerating. **Any condition of any name now disqualifies a package** — only subpath keys and `default` are permitted anywhere in the `exports` tree. That is airtight rather than one more patch: `--conditions` can enable an arbitrary name, so no denylist can be complete, while `default` and subpaths select the same entry no matter which conditions are live, and a package with no `exports` at all resolves through `main`, which conditions never reach.

It costs coverage again, and it is worth saying plainly what the fix now does and does not cover. Shimmed: CommonJS packages with no `exports` map, or an `exports` map with no conditions. Left exactly as before this PR: everything conditional. lodash, `mime-types` and the rest of the reported bug population are in the first group, which is why the fix still does its job.

**[P1, codex] `ENAMETOOLONG`.** Also real, and it was a hard build failure rather than a runtime one:

```
error: ENAMETOOLONG: name too long, open '…/.wm_node_cjs/deep_pkg_ssss…_mod_js_u0i53lt3vasa.cjs'
Failed to build node bundle
```

for a CJS subpath whose components are each legal but whose flattened basename is not. The readable part of the shim name is now truncated to 64 characters; the hash is what was keeping specifiers apart anyway. That fixture now builds and returns its value.

**[P2, pi] No committed test for the refusal.** Agreed, and your framing of the risk is exactly right: the plausible future break is a "simplification" back to an enumerated denylist, which fails silently by loading the wrong entry.

Rather than a fixture package, I made the rule itself testable. The plugin moved out of the inline Rust `format!` template into `backend/windmill-worker/node_externals.bun.js`, interpolated the same way `RELATIVE_BUN_BUILDER` already is, which also drops a block of `{{`/`}}` escaping. `entryDependsOnCondition` is exported from it, and `test_node_externals_refuses_conditional_exports` runs it under bun against a manifest table: the four shapes a shimmable package may have, and the five ways one can vary by condition (`bun`, `import`/`require`, a custom `development`, a `node` condition under a subpath, and a condition nested in an array fallback).

I checked it discriminates rather than just passes — replaying the round-3 predicate against the same table returns `[…,false,false,false]` for the last three, so a revert to it fails the test. It is a pure-function test with no fixture tree, which is why it is worth committing where a package fixture was not.

**Matrix re-run on this head** — now 25 shapes including the two new fixtures, on node **v20.20.0** and **v26.4.0**, with bun **1.3.11** and **1.4.0**. All pass, plus `cargo test --test bun_jobs` (36/36) and the real-worker runs on a backend rebuilt from this commit. The extraction is behaviour-identical: the matrix was run against the rendered builder before and after the move.

rubenfiszel:
Round 4 clean: Claude and Pi both "good to merge", Codex nit-only.

**[P2, codex] Constraint comments over the four-line limit.** Fixed in 3509302be3 — both blocks are down to four lines with the constraint intact (why bun's resolution and node's `require` only agree without conditions; why a plain external breaks named imports from a lexer-hostile CommonJS package). No block in the file exceeds four lines now. Matrix (25 shapes, node 20 + 26) and `cargo test --test bun_jobs` (36/36) re-run after the edit.

**Staying in draft rather than flipping to ready.** The round was clean, but this changes how every non-cached `//nodejs` job links its npm dependencies, which is shared worker infrastructure rather than a self-contained fix — and rounds 1 through 3 each turned up a real regression in this same classifier (a `bun`-conditioned entry, a silently swapped `import`/`require` entry, a runtime `--conditions` entry). The rule is categorical now rather than an enumeration, so I do not expect a fourth, but that history is exactly why this wants a human look before it is marked ready rather than an automatic flip.

**Still unexercised: Windows.** Codex and I agree this is the remaining gap. The generated builder transpiles against `loader.bun.windows.js`, and the path handling is drive-letter safe by construction (forward-slashed job dir, importer normalised before the `/.wm_node_cjs/` test, `dirname` walk terminating on `C:\`), but no `//nodejs` job has actually been run on a Windows worker. Whoever picks this up with access to one should run a `//nodejs` script with a named and a namespace import of lodash there.

rubenfiszel:
Round 5: Claude and Pi "good to merge"; addressing the Codex P1.

**[P1, codex] Generated shims bypass runtime loader hooks.** Reproduced, and the premise checks out — `build_envs` takes the script's own `envs` list, so off `CLOUD_HOSTED` a script can set `NODE_OPTIONS` itself. Built `import _ from "lodash"` (the shape that works on `main`, so the one with prior behaviour to lose) both ways and ran each under node v26.4.0 with a hook redirecting `lodash`:

| interception mechanism | before | after |
| -- | -- | -- |
| `register()` / `--experimental-loader` (async, ESM-only) | intercepted | **not intercepted** |
| `module.registerHooks()` (sync, node >= 22.15) | intercepted | intercepted |
| `--require` + `Module.prototype.require` patch | intercepted | intercepted |

So the bypass is real but confined to **async ESM hooks**. Sync `registerHooks()` and `--require` interception — which is what APM auto-instrumentation actually uses — still see a shimmed package, because bun emits `createRequire(import.meta.url)("lodash")` and that goes through the CJS loader.

I could not act on either remedy you name, and I think both are unavailable rather than merely awkward:

- **"Preserve ESM resolution"** cannot coexist with the fix. Node's ESM loader cannot expose lodash's named exports at all — that *is* the bug. Any fix has to obtain the CommonJS exports object via `require`, and node does not route `require` through async ESM hooks by design. I tried the shapes that might dodge this (an ESM shim re-exporting the default; a lexer-friendly CJS re-export file left external) and each still needs a `require` to produce the object bun hangs its runtime getters off.
- **"Leave these imports external when runtime loader hooks are configured"** is not soundly detectable. The bundler runs `env_clear()` with only `common_bun_proc_envs` (`bun_executor.rs:1032-1037`), so it never sees the job's `NODE_OPTIONS`; and the bundle cache key is content + lock + transitive import versions, with no env component (`bun_executor.rs:1367`), so a bundle built for a job without `NODE_OPTIONS` is reused verbatim by a later job that sets it. A build-time gate would be silently stale — the same failure shape as rounds 1-3.

What *would* bound it is narrowing the shim to the import shapes that are broken today, leaving default-only and side-effect imports external so nothing that works today changes behaviour at all. I did not do it: it needs an import-clause classifier, `Bun.Transpiler.scan()` gives specifiers but not clauses (checked), so it would be hand-rolled — and a mis-parse in the "looks default-only" direction re-breaks the reported bug on the common path to protect an exotic one. That trade is a judgement call rather than a mechanical fix, so I have written it up in the PR body with the measurements above and left it for the human review this PR is already waiting on.

Documented the constraint at the shim site in d767b1c3d0 and added a "Known limitation" section to the PR body. `cargo test --test bun_jobs` 36/36 after the edit; real `//nodejs` jobs on a local instance re-run for the named, namespace, default, subpath, dynamic, `mime-types`, `nanoid`, `zod`, `uuid`, `dayjs` and no-import shapes, all green.

Staying in draft, for the reason already in the description: this changes how every non-cached `//nodejs` job links its npm dependencies.

PR #10794, opened 2026-08-21
fix: resolve a script path to its new version as soon as the lock lands

rubenfiszel:
## Summary

Fixes WIN-2441.

Deploying a script and running it by path right after could execute the **previous** version.

`get_latest_deployed_hash_for_path` caches `(workspace, path) -> hash` in `DEPLOYED_SCRIPT_HASH_CACHE` with a 60s TTL, and only a version with a non-null `lock` is runnable. The lock is written asynchronously by a dependency job, so a deploy has two steps: the version row is inserted with `lock = NULL` (the path still resolves to the previous version, which is correct), and later the dependency job writes the lock (the path should now resolve to the new version).

Nothing made that second step visible promptly:

- `script_insert_trigger` was gated on `WHEN (NEW.lock IS NOT NULL)`, so a deploy that needs a lockfile emitted no invalidation event at all until the lock `UPDATE`.
- Invalidation travels through the `notify_event` table, polled every `LISTEN_NEW_EVENTS_INTERVAL_SEC` (10s by default), so the eviction lands up to a poll interval after the version became runnable.
- Anything that resolved the path during the lock window re-cached the previous version for the full 60s TTL, and nothing evicted that entry when the lock landed.

Measured on a dev instance before the fix: a run by path 8s after the lock landed still executed the previous version. Callers could not work around it, since `getScriptByPath` reports the new hash and a non-null lock while runs still get the old one.

## Changes

- **Notify on every script insert** (`backend/migrations/20260821084723_script_insert_notify_before_lock.up.sql`). The deploy is what makes the cached answer suspect, not the lock `UPDATE` that follows it, so other replicas now start invalidating from the deploy instead of from the lock. Every consumer of `notify_runnable_version_change` resolves without a lock predicate except `DEPLOYED_SCRIPT_HASH_CACHE`, so an earlier eviction only makes a warm cache behave like a cold one.
- **Don't hand a 60s TTL to an answer a dependency job is about to change.** `get_latest_deployed_script_hash` returns the runnable hash plus `pending_lock`: the newest version of the path is still waiting on its lockfile. While that holds, the cache entry gets a 2s TTL, so the first resolution after the lock lands reads the new version without waiting for an event to arrive. Both subqueries ride `index_script_on_path_created_at`. `pending_lock` stops believing in a dependency job that reported an error, or that is more than 10 minutes old and clearly not coming, so a version that will never become runnable cannot pin the path to the short TTL forever.
- **Evict this process' entries at the two mutation points**, so a single-server instance never waits at all: the API after a deploy commits, and the worker after it writes the lockfile. The worker drops only the runnable-hash cache, since the import-side caches ignore the lock and their answer already moved when the row was inserted. `notify_runnable_version_change` still carries both to the other replicas.
- **MCP `createScript` / `updateScript` instructions** no longer say a new version "can take up to a minute" to take effect. They now describe what actually happens: the lock is generated async, only a locked version is runnable, so poll `getScriptByPath` until `lock` is non-null.

### Residual window

Cross-replica invalidation is still bounded by the `notify_event` poll interval (`LISTEN_NEW_EVENTS_INTERVAL_SEC`, 10s by default): a replica holding an entry cached *before* the deploy keeps it until its next poll. What changes is that the deploy now starts that clock instead of the lock, so for any script whose dependency job takes longer than a poll interval (i.e. anything with real dependencies) the eviction has already happened by the time the version becomes runnable, and the short pending-lock TTL takes it from there. On the replica that served the deploy, and on any single-process instance, the window is zero.

## Test plan

- [x] `cargo test -p windmill-common --test latest_deployed_script_hash --test notify_events` (new: the `pending_lock` state machine across locked / pending / landed / failed / abandoned; extended: the insert trigger now fires for a version deployed without a lock).
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`, and `cargo check --features quickjs`.
- [x] **Reproduced and fixed end to end.** Deploy v1 (`return 1`), run it by path, deploy v2 (`return 2`), wait for its lock, then run by path in a loop. Before: the run executed v1 for 8.1s after the lock landed. After: the very first run executes v2 (`run_wait_result` returns `2`).
- [x] **Two replicas, same DB.** Warmed replica B's cache through a run by path, deployed a version whose lockfile lands later, let B poll the deploy event and re-resolve while the version was still pending, then wrote the lock. B flipped to the new version 2.1s later, having processed zero notify events in that span, i.e. the short pending-lock TTL closed it rather than the poll.
- [x] `EXPLAIN` confirms both subqueries use `index_script_on_path_created_at (workspace_id, path, created_at DESC)`.
- [ ] Not covered by an automated end-to-end test: the integration harness sets `DEPLOYED_SCRIPT_CACHE_DISABLED`, since these caches are process-global and shared across parallel tests, so a test of the cache behaviour would have to opt them back on for everyone. Exercised manually as above instead.

Left in draft: adds a migration and changes `openapi.yaml`, so it wants a human look before ready.

---
## Summary by cubic
Resolves WIN-2441: running a script by path right after deploy could execute the previous version. Old: path→hash caches updated only after the lock UPDATE and the next event poll; new: we notify on INSERT, use a short TTL while the new version waits for its lock, and evict locally at deploy commit and on lock write so the path resolves to the new version as soon as it’s runnable. Cross-replica residual window stays bounded by `LISTEN_NEW_EVENTS_INTERVAL_SEC` (10s default); single-process or the deploying replica flips immediately.

**What changed**
- DB: `script_insert_trigger` now always calls `notify_runnable_version_change` on INSERT (was gated on `lock IS NOT NULL`).
- Cache resolution: new `get_latest_deployed_script_hash` returns `{ hash, pending_lock }`; if `pending_lock = true`, cache TTL is 2s (was always 60s). Pending stops if `lock_error_logs` is set or the version is older than 10 minutes.
- Invalidation: after deploy commit and on lockfile write, evict local path→hash via `invalidate_latest_script_hash_caches` / `invalidate_deployed_script_hash_cache`. Event handling now evicts both deployed- and import-side latest-hash caches for scripts.
- API/docs: MCP `createScript`/`updateScript` now tell clients to poll `getScriptByPath` and stop on either outcome: lock non-null = ready; `lock_error_logs` set = lock failed and that version will never run.

**Rollout / migration**
- Apply migration `20260821084723_script_insert_notify_before_lock`.
- No backfill. Optionally lower `LISTEN_NEW_EVENTS_INTERVAL_SEC` to shrink the cross-replica window. Clients need no code changes but should follow the documented poll-and-stop guidance above.

<sup>Written for commit 8e196937265d8b7bfd8ae73604364a6e09700931. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ ba0cb188ffe49f13655b2a8e0309e513ee2aa508 — nit-only verdicts at ba0cb188ffe49f13655b2a8e0309e513ee2aa508 (codex + claude: mergeable with nits; pi: good to merge); nits addressed in 8e196937265d8b7bfd8ae73604364a6e09700931.

The MCP `createScript`/`updateScript` instructions now give the poll an exit on failure (`lock_error_logs` set means the lockfile failed and that version will never run), and the worker-side cache comment is back under the 4-line limit.

Left in draft deliberately: this adds a migration (`script_insert_trigger` loses its `WHEN (NEW.lock IS NOT NULL)` gate) and changes `openapi.yaml`, so it wants a human look before ready rather than an automatic flip on a clean round.

PR #10796, opened 2026-08-21
fix: keep raw-app files within their app folder on sync pull

hugocasa:
## Summary

A raw app's `value.files` keys are app-root-relative paths (`/index.tsx`, `/src/util.ts`) that `wmill sync pull` writes to disk relative to the app's folder. Nothing constrained those keys to resolve *within* that folder, so a key with a `..` segment could resolve elsewhere. This adds containment at both the deploy boundary and the pull boundary so a raw-app file always lands inside its own app folder.

## Changes

- **CLI (`cli/src/commands/sync/sync.ts`)** — extract `rawAppFilePathWithinFolder()`, which resolves each `value.files` key under the app folder and rejects any key that resolves outside it (via a `path.relative` containment check). This is the write-side guard on pull.
- **Backend (`backend/windmill-api/src/apps.rs`)** — add `validate_raw_app_file_keys()`, called from `create_app_internal`/`update_app_internal` (guarded on `raw_app`, before any persistence). It validates keys against the value *as it will be stored and read*: it strips NUL escapes first (matching the storage INSERT) and parses with `serde_json::Value` (last-duplicate-wins, matching how a puller's JSON parse resolves duplicate fields), so a key that looks fine on the wire but not once stored is still rejected. All four raw-app deploy entry points funnel through these two functions.
- **Tests** — Rust unit tests for `validate_raw_app_file_keys` (accept, reject, NUL-obfuscated, duplicate-field, no-op) and CLI unit tests for `rawAppFilePathWithinFolder`.

## Test plan

- [x] Backend rejects a raw-app create/update carrying an out-of-folder key with HTTP 400 (literal, NUL-obfuscated, and duplicate-`files` variants), before storing a version
- [x] A legitimate raw app with nested files (`/src/util.ts`) deploys and pulls unchanged
- [x] `wmill sync pull` of a value planted directly in the DB with an out-of-folder key aborts without touching a sentinel file outside the app folder
- [x] `cargo test -p windmill-api raw_app_file_key` and the CLI unit tests pass

---
## Summary by cubic
Prevents path traversal in raw-app file keys and runnable IDs during deploy and wmill sync pull. Previously, .. segments (including Windows-style \) could escape the app folder; now deploys and pulls reject such paths, and unparseable values fail closed. Addresses WIN-2186.

- Backend: validates value.files keys and value.runnables IDs after NUL stripping with last-duplicate-wins JSON semantics; rejects any .. segment split on / or \; returns HTTP 400 if the value cannot be parsed for validation.
- CLI: adds rawAppPathWithinFolder() and uses it for files and runnable IDs; resolves under the app folder (or backend subfolder) and errors if the path escapes containment. Strips only a single leading / from file keys to match backend validation.
- Tests: unit tests cover traversal (including Windows separators), NUL-obfuscated keys, duplicate fields, unparseable values, runnable-ID escapes, and CLI containment.

**Migration**
- Deploys with out-of-folder file keys or runnable IDs now return HTTP 400, and wmill sync pull errors. Remove .. segments and keep paths app-root-relative (e.g., /src/util.ts, runnable ID a/b).

<sup>Written for commit cb04a3f4272c47eff66ba99d9a0e9f6f8fefcab9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10796?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Thanks all — addressed the round in `2fd7da084bba79f209c223637a73b136a0978adb`:

**Runnables traversal (Claude P1)** — fixed. `value.runnables` ids also name a written file (`<backend>/<id>.yaml`), so an id like `../../../etc/evil` escaped the same way. Both layers now cover it: the CLI helper is generalized to `rawAppPathWithinFolder(baseFolder, relPath)` and the runnable-id write goes through it, and the backend `validate_raw_app_path_keys` now polices `runnables` keys alongside `files`. Added a Rust test (`rejects_traversal_in_runnable_ids`) and a CLI case; verified live that a `..` runnable id is rejected with 400 and a normal id still deploys.

**Fail-open on parse error (Pi P2)** — fixed. The validator now returns `Err` when `serde_json::from_str::<Value>` fails (e.g. a value nested past serde's recursion limit that `Box<RawValue>` accepts but `Value` rejects), rather than accepting an un-inspectable value. Added `rejects_value_too_nested_to_inspect`.

**Comment duplication (Claude P2)** — trimmed the redundant call-site comment and the test-file header; the two function doc comments are the single home for the invariant.

**Symlink containment (Codex P0)** — I don't think this belongs in this PR, and I'd like a maintainer's call before merging. Reasoning: `value.files`/`value.runnables` writes create *regular* files (`writeFile` with string content) — the pull never creates a symlink. So the symlink escape requires a symlink to already exist in the victim's checkout, which an attacker whose only capability is `apps:write` (the threat model this PR closes) cannot plant; introducing a checkout symlink needs write access to the git repo itself, at which point this vuln adds nothing. Real-path/symlink containment is a legitimate hardening, but it's a general property of *every* `writeFile` in `sync pull` (scripts, flows, apps alike), pre-existing and outside this diff. I'd prefer to track it as a separate follow-up rather than expand this focused fix into the sync-pull write loop — happy to do that PR if you'd rather. @hugocasa

PR #10797, opened 2026-08-21
fix: require item read scope on workspace tarball export

hugocasa:
## Summary

Scoped-token enforcement on the workspace tarball export was coarser than on the equivalent per-item routes. The archive collects resource and variable `value`s, which `get_resource_value` / `get_variable` gate on their own domain, while the export handler only asked for the scope the route itself needs. This aligns the two.

Because a whole-workspace export cannot be confined to a path, the requirement is the **unqualified** domain scope: `ScopeDefinition::includes` treats a path-qualified grant as not satisfying an unqualified requirement, so a path-scoped token passes `skip_resources` / `skip_variables` instead of exporting the whole workspace. This subsumes the narrower plain-secret-only `variables:read` gate from #9712 — every input that check denied is still denied.

Unscoped tokens are unaffected, since `check_scopes` no-ops on them. That covers UI session tokens and job `$WM_TOKEN`s (`create_jwt_token` mints them with `scopes: None`), so git-sync is unchanged.

## Changes

- `workspaces_export.rs`: require `resources:read` when the resource block runs, and `variables:read` when the variable block runs, replacing the plain-secret-only gate
- `backend/tests/workspace_export.rs`: `test_tarball_export_gates_values_on_item_scopes` pins the denial for a bare `workspaces:read` token and for a path-qualified grant, that skipping both kinds still succeeds, and that unqualified item-read scopes still yield the values

## Test plan

- [x] `cargo test --test workspace_export` — both tests pass
- [x] `cargo check -p windmill-api` — clean, no warnings
- [x] Exercised on a running instance with real tokens minted via `/api/users/tokens/create`: bare `workspaces:read` gets 403 naming `resources:read`; with `skip_resources=true` it gets 403 naming `variables:read`; with both skips it gets 200; a path-qualified grant gets 403
- [x] `workspaces:read,resources:read,variables:read` gets 200 with values present, and `plain_secret=true` still decrypts
- [x] Unscoped admin token unchanged, with and without `plain_secret`
- [x] `wmill sync pull` with `users:read,workspaces:read` fails cleanly with `Permission denied: Required scope: resources:read` before writing anything (no partial sync, no truncated tar); granting `resources:read` + `variables:read` on the same token makes the pull complete with the resource and variable files written
- [x] A real job's `$WM_TOKEN` still gets 200 with values present, confirming the git-sync path is unaffected

---
## Summary by cubic
Require item-read scopes for workspace tarball exports and accept `*` as a whole-domain grant for enforcement and delegation. Aligns the export with per-item routes and preserves the unscoped-only requirement for including the workspace key.

- Export requires `resources:read` when resources are included and `variables:read` when variables are included; use `skip_resources=true` and/or `skip_variables=true` to bypass those checks.
- Wildcard path grants (`resources:read:*`, `variables:read:*`) satisfy the unqualified requirement and may delegate `resources:read` / `variables:read`; path-scoped grants do not.
- Including the workspace key requires admin and an unscoped token; this check runs before item-scope gates.
- Unscoped tokens (UI sessions and job `$WM_TOKEN`) remain unaffected.
- Migration: scoped API clients must add `resources:read` and `variables:read` for full exports, or set both skip flags. Addresses WIN-2187.

<sup>Written for commit bd10bc2ec515b4d2aa75a08cc9e5e9899984f367. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10797?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
✅ Review round clean @ 2d167f26fd55efdc973e372651bd8d59cf89c1dd

PR #10803, opened 2026-08-21
fix: name the requested storage when a workspace storage lookup finds nothing

rubenfiszel:
## Summary

Reading an s3 object whose `storage` field names a storage the workspace does not have failed with:

```
500 Internal Server Error - Internal: No files storage resource defined at the workspace level
```

That sentence describes a workspace with **no** storage configured at all. It sends the reader to the workspace settings — where the storage is configured and looks perfectly correct — and never mentions the `storage` field that actually caused the failure. The value that reached `get_large_file_storage` was invisible in the message, so the only way to find it was to guess.

Two defects on that path:

1. **The requested storage was never named.** Thirteen call sites collapsed both causes — "no storage by that name" and "no workspace storage at all" — into one message.

2. **`_default_` did not resolve.** `_default_` is this codebase's own name for the primary storage (`windmill_object_store::DEFAULT_STORAGE`): it is what `get_secondary_storage_names(include_default=true)` returns, what the `/s3_proxy/{storage}/…` URL carries, and what the clients fall back to. `get_large_file_storage` looked every non-`None` name up under `secondary_storage`, where `_default_` can never exist — so an s3 object carrying the primary storage's own name reported "no files storage" and failed outright. `s3_proxy_ee.rs`, `windmill-api-workspaces/src/workspaces_extra.rs` and `windmill-common/src/workspaces.rs` all already normalize it; `job_helpers` did not.

## Changes

- **`workspace_storage_not_found(storage: Option<&str>)`** in `windmill-object-store` builds the message from the storage that was asked for. A named one now reports:

  > No files storage named 'exports' is defined at the workspace level. A storage name must be one of this workspace's secondary storages, or `_default_` (equivalently, nothing at all) for the primary one.

  The wording stays neutral about where the name came from — callers pass an s3 object's `storage`, a request field, or a trigger's stored config. It is a `BadRequest` — the name is caller input — while the "no workspace storage at all" case stays the internal error it is today. Neither may be `NotFound`, and neither message may contain "not found": `S3FilePreview.svelte`'s `isNotFoundError` reads both as "this object has not been written yet" and renders a calm empty state, which would swallow the message on the one surface that displays it. The constraint is recorded on the helper and asserted by the test.

- **Routed every workspace-storage call site through it**: the ten in `job_helpers_ee.rs`, the three siblings that took a caller-supplied `storage` but still reported it as "the workspace has no datasets storage" (`duckdb_connection_settings_v2`, `s3_resource_info`, `test_connection` — the last being the workspace-settings *Test connection* button), and the static-asset HTTP trigger. The two `apps.rs` upload/delete sites always pass `storage: None`, so their message is already accurate and they are unchanged.

- **`get_large_file_storage` filters `_default_`** before the `secondary_storage` lookup, matching the three modules that already do — as does the worker's own resolver (`windmill-worker/src/common.rs`, `get_workspace_s3_resource_path`), the one module the original sweep missed.

- **`_default_` is reserved as a storage name.** That filter is only sound if no secondary storage can carry the name — one that did would be shadowed, silently moving the workspace's reads *and writes* to the primary bucket. `secondary_storage` was an unvalidated `HashMap<String, _>` and the API accepted `_default_`, so `edit_large_file_storage_config` (the only route that creates one) now rejects it with a 400, and the test pins that. `DEFAULT_STORAGE` moved from `windmill-object-store` to `windmill-types::s3` so the validation can reference it without the optional `parquet` dependency; `windmill-object-store` re-exports it, so every existing call site is unchanged.

- **`sign_s3_objects` no longer signs when the storage does not resolve.** `get_workspace_s3_resource_and_check_paths` early-returns `Ok((None, None))` *before* running `check_lfs_object_path_permissions`, so an unresolvable `storage` skipped the `S3Permission::READ` check while the signature was minted anyway. Not exploitable today — redemption goes through the on-behalf read, which fails at resolution — but the class was invisible until this change.

  The guard tests `get_large_file_storage(...).is_none()`, not the resolver's `None` return, because that return *also* means "resolved but denied" — an RLS miss, or an instance `S3_BUCKETS_WORKSPACE_RESTRICTIONS` policy. Those ran the permission loop and must keep signing exactly as before; only the early-return case skips it. (An earlier revision of this PR guarded on the resolver's `None` and made a bucket-restricted workspace fail to sign where it previously succeeded.)

- **`S3FilePreview.svelte`'s `isNotFoundError` no longer matches a bare `404` substring** in the body. The named-storage message is the first on that path to interpolate a caller-supplied string, so a storage named `archive404` would otherwise render as the "asset not yet materialized" empty state — the exact failure the helper's contract exists to prevent. `err.status` already covers the real case, and every backend not-found serializes as `"Not found: …"`, which the remaining substring still matches.

- **`_default_` signatures are now redeemable.** The signer wrote `&storage=_default_` into the HMAC verbatim while `s3_proxy_ee.rs::get_object` normalizes `_default_` away before rebuilding the message, so the two HMACs never matched and the URL failed with `Invalid signature`. A second validator caller (`apps.rs`, the `apps_u` file routes) passed `storage` unnormalized, so fixing only the signer would have left that path broken. Both sides now build the fragment through one `s3_signature_storage_fragment` helper — the signer is CE and the validator EE, and a divergence surfaces only as `Invalid signature`, which nobody would trace back here.

### Behaviour changes beyond the message

- `storage=_default_` now resolves instead of failing. As a consequence the S3 permission check, which that same early return used to skip for `_default_`, now runs. That is a tightening, not a regression.
- An unknown storage **name** goes from HTTP 500 to 400 on the ten `job_helpers_ee.rs` sites, and from 404 to 400 on the three sibling sites.
- On those same three sibling sites, "**no workspace storage at all**" goes from 404 to **500** — the one status move in the unfavourable direction, taken so all thirteen sites agree. Inert in practice: `S3FilePickerInner.reloadContent` is the only consumer and it catches any error without inspecting the status.
- **Upgrade behaviour, accepted deliberately:** the `_default_` reservation is forward-only. A workspace that *already* has a secondary storage named `_default_` — which the API used to accept — has its reads and writes re-pointed at the primary storage on deploy, with no migration. Judged not worth a migration: the name has to be typed deliberately (the UI proposes `storage_N`), and no such configuration is known to exist. Flagged here rather than left to be discovered.
- `sign_s3_objects` is a `try_join_all` batch, so one object with an unresolvable `storage` now fails the whole request rather than signing the rest — consistent with how a permission denial in the batch already behaves.
- **On deploy only:** a signature minted *before* the deploy with an explicit `storage: "_default_"` and redeemed *after* it on the `apps_u` routes fails with `Invalid signature` until it expires (≤12h; a page reload re-signs). That pairing was the one place such a signature used to verify — on `/s3_proxy/` it was already broken.

## Test plan

Verified against a local EE instance (`quickjs,private,enterprise,parquet,python`) with MinIO-backed workspace storage, on both a workspace and a fork of it:

- [x] `GET /job_helpers/download_s3_file?storage=_default_` — was `500 No files storage resource defined at the workspace level`, now `200` streaming the object's bytes
- [x] `?storage=bogus` — was the same generic 500, now `400` naming `bogus` and saying what a valid storage name is
- [x] no `storage` param — `200`, unchanged
- [x] `GET /job_helpers/test_connection?storage=_default_` — was a 404 "no datasets storage", now `200`
- [x] `POST /job_helpers/v2/duckdb_connection_settings` with `storage: "bogus"` — now names it
- [x] under `S3_BUCKETS_WORKSPACE_RESTRICTIONS` excluding the workspace, `sign_s3_objects` still returns 200 for a configured storage (the resolver's `None` is a denial, not a missing storage) while a bogus `storage` is still refused
- [x] `cargo check` clean with no features (`windmill-api-workspaces`, `windmill-object-store`), with `parquet` alone (the OSS `get_large_file_storage` stub path) and with `private,parquet` — confirming the `DEFAULT_STORAGE` move keeps non-`parquet` builds compiling
- [x] `cargo test --features enterprise,private,parquet --test workspace_storage_by_name --test sign_s3_objects_authz --test app_s3_onbehalf` — 9 passed

`backend/tests/workspace_storage_by_name.rs` pins `_default_` resolution and the named-storage error (status and wording included, so the previewer constraint above is guarded), against a `FilesystemStorage` LFS so it needs no object store. `sign_s3_objects_authz.rs` gains the `_default_` mint→redeem round trip and the unresolvable-storage refusal; they live inside the existing test function because `s3_proxy_ee.rs`'s `S3_RESOURCE_CACHE` is process-global and keyed by `(workspace, storage)`, so separate test functions sharing the one fixture workspace serve each other stale filesystem roots.

### Frontend

The one frontend change is a predicate narrowing in `S3FilePreview.svelte` with no visible UI effect on its own — it only stops an unrelated error from being misrendered as the existing "asset not yet materialized" state — so there are no screenshots. `svelte-check` is clean on the file.


rubenfiszel:
Round 2 nits addressed in `bfb0b94` (nit-only round, so no fresh full round):

- **s3-object-specific advice was wrong at four call sites** (codex + claude + pi). `test_connection`, `s3_resource_info` and `duckdb_connection_settings_v2` take the name as a request field, and the static-asset trigger reads it from stored config with an anonymous visitor as the reader. The message is now neutral about provenance: it names the bad value and states what a valid one is.
- **`isNotFoundError`'s bare `'404'` substring** (claude). Real hole — the named branch is the first message on that path to echo a caller-supplied string, so a storage named `archive404` would land in the "asset not yet materialized" state. Dropped the substring; `err.status` covers the real case and every backend not-found still serializes as `"Not found: …"`.
- **`sign_s3_objects` comment** (claude) now says the `None` also means "denied on the resource" once a `user_db` is threaded in, so the next person doesn't silently turn a denial into a storage-name error.
- **Duplicated/overlong invariant comments** (codex) trimmed to one statement at the helper, with the test docs saying only what they guard.

Left in draft deliberately: this touches `*_ee.rs` and a permission path, and two calls are worth a human look — the 500→400 / 404→400 status moves, and `sign_s3_objects` now failing a whole `try_join_all` batch on one unresolvable `storage`.

Worth flagging separately: **CI did not review the EE half.** The Claude reviewer noted its `windmill-ee-private` checkout is pinned at the pre-change ref rather than the one this PR bumps to, so it was reading the old `job_helpers_ee.rs`. The EE side is the bulk of the diff (13 call sites plus the `_default_` filter) and has only had a local review.

rubenfiszel:
Round 3 addressed in `6f6104c`.

**Fixed**

- **[codex P2] A storage name containing "not found" collapsed into the preview empty state.** Real hole. Dropping the substring test outright isn't available: a genuinely missing object reaches the frontend as a **500** whose message merely contains "not found" (`load_file_metadata_internal` wraps the object-store error), so that substring is the only signal for the case the empty state exists to serve. Instead `isNotFoundError` now short-circuits on **400** — a request whose own input was wrong is never an object that hasn't been written yet — which closes the hole without breaking missing-object detection.
- **[codex P2] Static-asset trigger returned 400 for a server-side misconfiguration.** Agreed. Added `workspace_storage_misconfigured`: same message, server-error class, for callers whose storage name comes from stored config rather than the request. The visitor can't correct a name they never supplied.
- **[claude] `windmill-worker/src/common.rs` left un-normalized.** Good catch — `get_workspace_s3_resource_path` was the one resolver the sweep missed, and the new `DEFAULT_STORAGE` rustdoc is exactly where a future reader would be told it can't happen. Now filtered like the others.

**Not fixing: [codex P1 / claude P2] legacy `_default_` secondary storages.**

Deliberate call by the maintainer, not an oversight. A secondary named `_default_` has to be typed by hand (the UI proposes `storage_N`), no such configuration is known to exist, and the reservation stops any new one. A migration was considered and judged not worth its cost against a state believed empty. The upgrade behaviour is now spelled out in the PR description rather than left to be discovered.

Worth noting the finding is real and was found by testing, not theory — an earlier revision of this PR *did* silently reroute such a workspace, which is why the reservation exists at all.

rubenfiszel:
Round 4 nits addressed in `d4a8533` (nit-only round).

**[claude] The recorded invariant didn't match what protects the code.** Correct, and the sharpest finding of the round. "Neither message may contain 'not found'" holds only for the unnamed branch; the named one interpolates a caller-supplied name. What actually saves it is the **400**, which `isNotFoundError` now short-circuits on. Rewrote the doc to state the real constraint per variant — named ⇒ must stay 400; unnamed ⇒ message must not read as missing — and added an explicit "not for any route the asset previewer reads" warning on `workspace_storage_misconfigured`, which is exactly the 500-with-interpolated-name shape that would reopen the hole for its next caller. Also `#[track_caller]` on it, so static-asset failures point at `handler.rs` again rather than the helper.

**[codex] Browser verification.** Done, and it corrected the finding. `archive-not-found` does **not** trip the substring — it is `"not found"` with a space, and that name has hyphens. Driven against the live API from the browser:

| storage | status | before fix | after |
|---|---|---|---|
| `archive not found` | 400 | swallowed | shown |
| `not found` | 400 | swallowed | shown |
| `archive-not-found` | 400 | shown | shown |

So the hazard is real but needs a literal space, and the 400 short-circuit is load-bearing (the response does carry `status: 400`, which the fix depends on). Both comments repeated the wrong example and are corrected.

Scope note on that verification: I could not reach `S3FilePreview`'s render branch through the UI — the file picker bails earlier on `testConnection` and shows its own "not connected" state, and the asset-graph pane needs an asset row the list would not surface. So this exercised the predicate against the real serialized error in a live browser, not the painted component. Stating that plainly rather than letting "verified in browser" stand unqualified.

**[codex] Comment length** trimmed to the ≤4-line rule.

Tests: 10/10. `svelte-check` clean. Builds verified for `parquet` alone, `private,parquet`, and `windmill-worker --features parquet`.

rubenfiszel:
Round 5 nits addressed in `cac0a45` — Codex and Pi were **Good to merge**; both of Claude's nits were correct.

- **`#[track_caller]` only helped the named branch.** The unnamed one used plain `Error::InternalErr`, which carries no location, so a static-asset trigger in a workspace with *no* storage lost the `@handler.rs` stamp it has on `main`. A small regression I introduced while fixing the location for the other branch. Both arms now go through `internal_err`.
- **The test comment stated the pre-fix invariant** the rustdoc had just been corrected away from. Rewrote it to name the real one (the 400 is what protects this), and dropped the neighbouring `!body.contains("not found")` assertion — as noted in round 4 it passes only because the fixture name is `nope`, so it reads as a guard while guarding nothing. An assertion that looks load-bearing but passes by accident of fixture data is worse than none.

Tests 10/10, `cargo check -p windmill-api --features parquet` clean, rustfmt clean. Nit-only fixes, so no further full round.

PR #10804, opened 2026-08-21
fix: size the ephemeral job token to the job timeout it must serve

rubenfiszel:
## Summary

On Windmill Cloud, a premium workspace's job may run `MAX_TIMEOUT * 6` (5400s), but the
ephemeral job token was minted for `SCRIPT_TOKEN_EXPIRY`, which defaulted to `MAX_TIMEOUT`
(900s). The token is issued once when the job is pulled and never refreshed under a running
script, so any premium job that called back into the Windmill API after ~15 minutes got a 401 —
after it had already burned the compute.

Observed in production on app.windmill.dev: a workspace running long web-automation scripts had
jobs fail at 17–37 minutes with

```
.../api/w/<ws>/resources/get_value_interpolated/f/db/redis: 401, Not authorized: Unauthorized
```

Only paying workspaces can run long enough to hit this — a free workspace is capped at
`MAX_TIMEOUT` and its token expired with it.

The token lifetime is now derived per workspace from the same ceiling the timeout uses, so it
cannot come out shorter than the run the instance permits. It is resolved per workspace rather
than globally so a free workspace, which cannot run that long, does not get the wider token.

Note the ceiling only bounds a single child process: `resolve_job_timeout` is called per
`handle_child`, so lock resolution, dependency install and bundling each carry a full timeout of
their own before the run phase gets its own. Pull-to-finish wall time is therefore not bounded by
the ceiling at all. The token gets one setup phase's worth of slack (`x2`) to cover the realistic
case; a pathologically slow install can still outlive it, and the doc comment says so rather than
claiming a guarantee the code does not hold.

## Changes

- `windmill-common/src/worker.rs`: new `max_job_duration_secs(premium)` owns the premium
  multiplier, replacing an inline `*MAX_TIMEOUT_DURATION * 6 //30mins` (the comment was also
  wrong — 900x6 is 90 min, not 30). Uses `saturating_mul`.
- `windmill-common/src/auth.rs`: new `job_token_expiry_secs(db, w_id)` resolves the token
  lifetime from that ceiling plus setup headroom, via the already-cached `get_team_plan_status`.
  On a lookup failure it assumes the wider ceiling, matching `resolve_job_timeout`, so the token
  can never come out shorter than the timeout the job is held to.
- `SCRIPT_TOKEN_EXPIRY` becomes `SCRIPT_TOKEN_EXPIRY_OVERRIDE` — an explicit operator override
  rather than the default. The env-var key is unchanged; this is a Rust-symbol rename only.
- All three minting sites go through the new helper: job pull (`windmill-queue/src/jobs.rs`),
  flow-step refresh (`windmill-worker/src/result_processor.rs`), zombie handler
  (`backend/src/monitor.rs`).
- Deleted `MAX_TIMEOUT_DURATION`, whose last consumer this PR removed. It is `pub`, so no
  dead-code warning would have fired, leaving a second and now-incomplete way to express the
  job ceiling.

## Test plan

- [x] `cargo check` and `cargo check --features cloud` clean
- [x] `cargo test -p windmill-common --lib auth::` — 6/6, including the new
      `job_token_outlives_the_longest_job_it_may_serve`, which pins the invariant that regressed
- [x] Reproduced on a running instance (`CLOUD_HOSTED=true`, `TIMEOUT=60`, premium workspace,
      job sleeps 150s then calls `resources/get_value_interpolated`):
      pre-fix `t=0 -> 200`, `t=150 -> 401`
- [x] Premium workspace, post-fix: token TTL 720s (60x6x2), `t=0 -> 200`, `t=150 -> 200`
- [x] Free workspace, post-fix: token TTL 120s (60x2) — not widened to the premium ceiling — and
      the job is correctly killed by its own 60s timeout
- [ ] Confirm on cloud that a premium job running past 15 minutes no longer 401s

Note when reproducing: `jsonwebtoken` validates `exp` with a 60s leeway, so a gap smaller than
that will pass either way and prove nothing.

---
## Summary by cubic
Sizes the ephemeral job token to each workspace’s job ceiling and adds bounded setup headroom, preventing premium jobs from 401-ing mid-run. Old: token TTL = `MAX_TIMEOUT` (900s). New: token TTL = max run duration per workspace + min(max run, 1h); this keeps self-hosted tokens ~7d while giving long cloud runs enough slack.

- Adds `max_job_duration_secs(premium)` and `job_token_expiry_secs(db, w_id)`; per-workspace resolution uses cached plan status and falls back to the wider ceiling on lookup failure to avoid under-sized tokens.
- Caps setup slack at 1h so very long ceilings (e.g., self-hosted 7d) don’t balloon credentials; free cloud workspaces get only setup slack, not the premium ceiling.
- Replaces token minting sites (job pull, flow-step refresh, zombie handler) to use `job_token_expiry_secs`.
- Renames internal symbol to `SCRIPT_TOKEN_EXPIRY_OVERRIDE` (env var key `SCRIPT_TOKEN_EXPIRY` unchanged). Drops `MAX_TIMEOUT_DURATION`; `resolve_job_timeout` now uses `max_job_duration_secs`.
- Test ensures token TTL always exceeds the job ceiling.

**Rollout**
- No config change required; override env var unchanged.
- Verify on cloud that premium jobs running >15 minutes no longer receive 401s on API calls.

<sup>Written for commit f30525fb0d1c356965cc1734f116d57dad656966. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10804?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ f30525fb0d1c356965cc1734f116d57dad656966 — nit-only verdicts at 4392adaf42ac676b3ffb27dd9fd19a90992737e6 (codex/claude nits, pi good to merge); nits addressed in f30525fb0d.

- **Self-hosted widening (claude P2)** — real, fixed. The `x2` took the self-hosted ephemeral JWT from 7d to 14d for slack that buys nothing when each phase already has 7 days of its own. Slack is now additive and capped at 1h: premium cloud 5400+3600, free cloud 900+900, self-hosted 604800+3600. Also clamped to `u32::MAX` so a saturated value can't reach the `as i64` cast in `create_jwt_token` and mint an already-expired token.
- **Comment policy (codex P2)** — fixed. The 12-line doc is down to 2 lines stating the constraint once; the setup rationale moved to the site it governs; the drafting-history comment on the test is gone.
- **`ansible_executor.rs` contradicts "must go through this" (claude P2)** — doc softened rather than converting ansible, which has no premium flag at that point and whose +24h slack already dwarfs the ceiling.
- **Stale `see create_token` pointer (claude P2)** — now points at `job_token_expiry_secs`.

Scope note on verification: the local e2e ran at `TIMEOUT=60`, where the ceiling is under the 1h cap, so the measured 720s/120s are unchanged by this commit — but that run does **not** exercise the cap. At production values the cap is what keeps self-hosted at 7d+1h, and that path is covered by reading and the unit test's invariant, not by an executed run.

PR #10808, opened 2026-08-22
feat: show the date on the runs dashboard chart axes

rubenfiszel:
## Summary

The x axis of both runs dashboard charts labelled ticks with bare clock times (`6PM`, `8PM`), so a window that never reached a day boundary carried no date at all and you could not tell which day a point belonged to.

Sub-day ticks stay bare clock times, except the leftmost tick and the first tick of each day, which spell out the date (`Aug 21 8PM`, `Aug 22`). Day-or-coarser axes are unchanged.

## Changes

- New `frontend/src/lib/components/chartjs-wrappers/timeTicks.ts`: `timeTicksWithDate(min, max)` returns the x-axis `ticks` config for a chart.js time scale — a tick callback that prefixes the date on the leftmost tick and on each day's first tick, picking the clock format from the tick spacing.
- Chart.js major ticks are enabled only when the axis spans more than one calendar day. Majors pin autoSkip's grid to the day boundaries, which is what keeps a wide axis aligned on (and dated at) whole days — but they also un-pin it from tick 0, whose date is the only one an axis inside a single day has.
- `RunChart.svelte` (Duration) and `ConcurrentJobsChart.svelte` (Concurrency) use it in place of their inline `ticks` options.
- Unit tests for the label rules and the major-tick gate.

## Screenshots

30 h window crossing midnight — leftmost date plus the dated day boundary:

30h

2 h window inside one day — the case that previously showed no date anywhere:

2h

28 day window — unchanged, day ticks already carried dates:

28d

Concurrency tab, same 30 h window:

concurrency

## Test plan

- [x] `npx vitest run src/lib/components/chartjs-wrappers/timeTicks.test.ts`
- [x] `npm run check:fast`
- [x] Runs dashboard, Duration and Concurrency tabs, with 12 jobs spread over 30 h: 3 min, 2 h aligned and unaligned, 8 h, 30 h aligned and unaligned, and 28 day windows all name a date; each window that crosses midnight labels the boundary
- [x] Drag-zoom into a few hours inside one day: the zoomed axis still names the date

---
## Summary by cubic
Show the date on the runs dashboard x-axes so sub-day ranges are unambiguous. Previously ticks showed only times (e.g., 6PM); now the leftmost tick and the first tick of each day include the date, while day-or-coarser axes remain unchanged. Major ticks are enabled only when the axis spans more than one day to align day boundaries without hiding the leftmost dated tick.

- Add `timeTicksWithDate(min, max)` in `frontend/src/lib/components/chartjs-wrappers/timeTicks.ts` to generate `chart.js` time-scale ticks; use it in `RunChart.svelte` and `ConcurrentJobsChart.svelte`.
- Gate `major` ticks to multi-day ranges to keep whole-day alignment on wide windows and preserve the only dated tick on single-day windows.
- Add `timeTicks.test.ts` to cover label formatting and the major-tick gate; no API or data changes.

<sup>Written for commit bc74ee102b235f209223831704debc655425c445. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ bc74ee102b235f209223831704debc655425c445

PR #10809, opened 2026-08-22
fix: keep ai chat messages when leaving the page mid-generation

hugocasa:
## What's broken

Start an AI chat turn, leave the page before it finishes, come back: everything except your first message is gone.

The chat's agent loop runs entirely in the browser tab. A turn was written to storage exactly twice: once when your message went in, once when the turn finished. Everything in between lived only in memory. Close or reload the tab in that window and the whole turn went with it.

The worst part is that the work was real. In the repro, the model had already created a script draft. After the reload that draft was sitting in the workspace, but the conversation had no idea it existed.

This is why it felt intermittent: if the turn happened to finish before you came back, the end-of-turn save caught everything and nothing looked wrong.

## What this changes

A turn is now saved *while it runs*, not just at its edges. Concretely, that means calling the same `saveChat` that already existed, at more moments, with a transcript assembled from the turn's in-flight state.

**When it runs.** A 2s timer. Each tick compares a small fingerprint (messages collected, transcript rows, answer text received) and returns immediately if nothing changed. So a turn parked on a confirmation ticks and writes nothing; an active turn writes at most once every 2s. There is also one forced write when the page goes hidden, which is the last moment a write can still complete before a close or reload.

**What it costs you.** At most a couple of seconds of a turn, instead of all of it.

## What this does *not* change

- **The generation still stops when the tab dies.** There's no server-side loop to resume. You come back to a stopped turn that you can continue by sending a message, not to one still running.
- **Navigating inside Windmill was already safe** and still is. The chat lives in a module singleton and session runtimes are cached, so switching pages or session tabs never interrupted anything.

## The details it has to get right

Each of these is small, and each one was a bug caught in review. The first is the one worth a careful look.

| | |
|---|---|
| **Stop before the commit** | The end-of-turn code merges the turn into the transcript and *then* saves. A checkpoint landing in that window would append the same messages again and, queued behind the real save, win — producing duplicate `tool_call` ids that providers reject. So checkpointing stops the moment the loop returns, not in `finally`. |
| **Settle in-flight tool cards** | A card saved mid-run restores spinning forever, and a staged confirmation restores with Run/Reject buttons nothing is listening to. This also fixes the same gap on the existing Stop path. |
| **Keep the finished half of a batch** | A model can request several tools in one message, and they run one at a time. Dropping an unfinished batch also drops the calls that already completed, whose side effects are already real. Unfinished calls get a synthesized "interrupted" result instead. |
| **Carry buffered screenshots** | A screenshot only becomes a message once its whole batch finishes, so a batch saved mid-run would otherwise reference an image the model can't see. |
| **Capture text as it arrives, not as it's painted** | A hidden tab pauses the animation frames the typewriter runs on, so painted text freezes while text keeps arriving. |
| **Don't store a stale context size** | It measures the pre-turn history, so keeping it would under-report a restored chat by the whole partial turn — enough to skip the compaction its next send needs. |
| **Don't swap the transcript mid-turn** | Switching conversation from the History menu while a turn runs made it commit onto a different transcript. The entries are now disabled while generating. |

## Screenshots

A multi-tool turn interrupted by a reload, restored in full:

restored-transcript-top

A turn that was parked on a tool confirmation when the page went away. The staged call comes back as a settled card showing what it was about to run, with no Run/Reject footer:

restored-after-reload

## Test plan

Driven against Anthropic through the real `/ai/proxy` on a local instance.

- [x] Reload mid tool-batch: all 17 model messages and 3 drafts restored, and a follow-up answers from the restored history
- [x] Reload while parked on a confirmation: full transcript plus an interrupted card, no dead buttons, model correctly recalls what it had done
- [x] Reload mid streamed answer: 10,221 characters of context restored, and the visible bubble matches it exactly
- [x] Backgrounded tab with animation frames stubbed out: capture keeps advancing (5560 → 8979 chars) where it used to freeze
- [x] Clean turn with repeated hide events: no duplicated or unpaired tool calls, real context size preserved
- [x] Follow-up on a conversation restored from a synthesized batch close: provider returns 200, not a rejection
- [x] Every regression test confirmed to fail when its own fix is reverted
- [x] `npm run check` 0 errors; 1004 passing in `src/lib/components/copilot/chat/` (6 failures there are pre-existing on main, in `ContextManager.test.ts` and `app/core.test.ts`)
- [x] Write cost measured in-browser: 22 ms per checkpoint on a 1.7 MB transcript, 32x larger than the biggest real chat, capped at one write per 2s

## Note for later

The checkpoint composes a transcript rather than appending to the live one, because a turn's messages aren't committed until its outcome is known. A cleaner shape would append at tool-batch boundaries, which needs a "step completed" event from the provider parsers. That's a change to the loop's contract and is deliberately not in this PR.

### DIFF
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -612,8 +612,8 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 									{#each pastChats as chat (chat.id)}
 										<button
 											class="text-left flex flex-row items-center gap-2 justify-between hover:bg-gray-100 dark:hover:bg-gray-700 rounded-md p-1 disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-transparent dark:disabled:hover:bg-transparent"
-											disabled={aiChatManager.loading}
-											title={aiChatManager.loading
+											disabled={aiChatManager.loading || aiChatManager.sendInFlight}
+											title={aiChatManager.loading || aiChatManager.sendInFlight
 												? 'Stop the current answer to switch conversation'
 												: undefined}
 											onclick={() => {
--- a/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
+++ b/frontend/src/lib/components/copilot/chat/AIChatDisplay.svelte
@@ -611,7 +611,11 @@ the panel, or the Escape-to-stop focus check would wrongly reject them. -->
 								<div class="flex flex-col">
 									{#each pastChats as chat (chat.id)}
 										<button
-											class="text-left flex flex-row items-center gap-2 justify-between hover:bg-gray-100 dark:hover:bg-gray-700 rounded-md p-1"
+											class="text-left flex flex-row items-center gap-2 justify-between hover:bg-gray-100 dark:hover:bg-gray-700 rounded-md p-1 disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:bg-transparent dark:disabled:hover:bg-transparent"
+											disabled={aiChatManager.loading}
+											title={aiChatManager.loading
+												? 'Stop the current answer to switch conversation'
+												: undefined}
 											onclick={() => {
 												loadPastChat(chat.id)
 												close()
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -2282,9 +2282,6 @@ export class AIChatManager {
 		snapshot?: {
 			/** Result to synthesize for the calls of a batch caught mid-execution. */
 			interruptedToolContent: string
-			/** True when partialReply has not been folded into a message yet, so it
-			 *  cannot be the stale duplicate the dedup below exists to drop. */
-			textIsLive: boolean
 			/** Images a tool has produced that the turn has not yet turned into a
 			 *  message (see appendPendingToolImages). */
 			bufferedImages?: ChatCompletionMessageParam
@@ -2294,8 +2291,9 @@ export class AIChatManager {
 			? closeInterruptedToolBatch(collectedMessages, snapshot.interruptedToolContent)
 			: truncateToToolPairedPrefix(collectedMessages)
 		// partialReply can be stale — equal to text already committed inside the
-		// prefix — so only append when new. Live text is exempt: identical text can
-		// legitimately recur across iterations, so content alone cannot judge it.
+		// prefix — so only append when new. A snapshot is exempt: it passes only
+		// live streaming text, which is never in the prefix, and identical text can
+		// legitimately recur across iterations where content alone cannot judge it.
 		const lastCommittedText = [...prefix]
 			.reverse()
 			.find(
@@ -2303,7 +2301,7 @@ export class AIChatManager {
 					m.role === 'assistant' && typeof m.content === 'string' && !!m.content.trim()
 			)?.content
 		const keptPartialReply =
-			!!partialReply.trim() && (!!snapshot?.textIsLive || partialReply !== lastCommittedText)
+			!!partialReply.trim() && (!!snapshot || partialReply !== lastCommittedText)
 		// Images sit between the batch that produced them and whatever the model
 		// said next, matching where appendPendingToolImages puts them live.
 		const tail = snapshot?.bufferedImages ? [...prefix, snapshot.bufferedImages] : prefix
@@ -3025,10 +3023,6 @@ export class AIChatManager {
 		// that never became one.
 		const collectedMessages: ChatCompletionMessageParam[] = []
 		let partialReply = ''
-		// Whether partialReply is already in collectedMessages. Parsers call
-		// onMessageEnd both before pushing (text giving way to a tool call) and
-		// after (a finished message), so only the capture site can tell.
-		let partialReplyCommitted = false
 		// Once an outcome branch (commit/restore) took over, a later throw (e.g.
 		// from saveChat) must not make the catch commit the turn a second time.
 		let turnOutcomeHandled = false
@@ -3052,12 +3046,15 @@ export class AIChatManager {
 			// nothing and the rate follows steps taken rather than time.
 			const shape = `${collectedMessages.length}:${this.displayMessages.length}:${streaming.length}`
 			if (!force && shape === checkpointedShape) return
+			// Live text only. Text the parsers have flushed is theirs to push, and
+			// they do so before the tool execution a checkpoint is likely to land in
+			// — so reading it here would mean re-appending what the transcript
+			// already holds. The abort path still recovers it via partialReply.
 			const { messages, keptPartialReply } = this.interruptedTurnMessages(
 				collectedMessages,
-				streaming || partialReply,
+				streaming,
 				{
 					interruptedToolContent: INTERRUPTED_TOOL_RESULT,
-					textIsLive: !!streaming || !partialReplyCommitted,
 					// A screenshot's image becomes a message only once its whole batch
 					// does, so a batch closed mid-flight would restore a result
 					// announcing a screenshot the model cannot see. Read without
@@ -3410,8 +3407,6 @@ export class AIChatManager {
 						// deduped in commitInterruptedTurn.
 						if (this.currentReply) {
 							partialReply = this.currentReply
-							const last = collectedMessages[collectedMessages.length - 1]
-							partialReplyCommitted = last?.role === 'assistant' && last.content === partialReply
 						}
 						if (this.currentReply || this.currentReasoning) {
 							this.displayMessages = [
--- a/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
+++ b/frontend/src/lib/components/copilot/chat/AIChatManager.svelte.ts
@@ -159,6 +159,11 @@ function prefersInstantReveal(): boolean {
 // schema changes from mode switches, and the estimate's chars/4 error.
 const COMPACTION_TRIGGER_RATIO = 0.8
 const COMPACTION_TARGET_RATIO = 0.7
+// How often a running turn is offered to the mid-turn checkpoint (see
+// sendRequest). The whole transcript is rewritten on each accepted checkpoint,
+// so this bounds the write rate; it also bounds how much of a turn a tab
+// closed without warning can lose.
+const CHECKPOINT_INTERVAL_MS = 2000
 // Flat per-image token estimate for a downscaled (≤1568px) vision image. Used instead
 // of chars/4 on the base64 data URL, which would overcount by ~50x.
 const IMAGE_TOKEN_ESTIMATE = 1200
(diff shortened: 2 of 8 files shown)


hugocasa:
Round 1 addressed in 2404eb9634. Verified each fix fails the new test when reverted, and exercised all of it against Anthropic on a local instance.

**Codex P1 — partial multi-tool batches not checkpointed.** Real, and the more valuable half of the two. `truncateToToolPairedPrefix` drops a batch until every call in it has a result, so a batch where one `write_script` finished and a second call was still running lost the finished step *and its already-real side effect*. Added `closeInterruptedToolBatch` (`chatLoop.ts`), used by the checkpoint only: it keeps the batch and synthesizes a result for each unfinished call, matching what those cards already show the reader. `commitInterruptedTurn` still truncates — a turn being committed for a follow-up is still live and nothing has been lost yet, so the model is better off rerunning the batch than reading invented results. Confirmed on a real parallel batch: 6 real results retained, 2 synthesized, 0 unpaired, 0 duplicated.

**Codex P1 / Claude P2 / Pi P2 — streaming under-capture.** Both halves fixed. `currentReply.length` now joins the fingerprint, so the poll keeps re-capturing the answer as it grows (measured live: 3001 → 3916 → 4393 → 4947 → 5439 → 6477 chars across successive polls; it froze at the first slice before). And the forced hide path now calls `replyReveal.flush()` before reading, so the typewriter backlog is not dropped — the polled path deliberately does not flush, since that would snap the reveal to the end on a page the reader is still watching. After a mid-stream reload: 10,221 chars of model context and a visible bubble of exactly the same 10,221 chars.

**Pi — the `CHECKPOINT_INTERVAL_MS` comment overstated the guarantee.** Correct, and it is now true rather than reworded: with the reply in the fingerprint the interval really does bound the loss.

**Claude P2 — the shape dedup assumes it is the record's only mid-turn writer.** Accepted as a known bounded window rather than fixed here. `#persistBackgroundJobs` / `#persistModifiedItems` do write the live display against pre-turn `this.messages`, and the dedup can suppress the repair. But it self-heals on the turn's next advance and unconditionally at turn end, so only a tab that dies inside that window is affected — and today that window loses the entire turn regardless, so this is strictly better, not a regression. Fixing it properly means making those two writers turn-aware, which is a different change from this one; happy to do it as a follow-up if you'd rather it not wait.

Also noted from the Claude review and left alone as pre-existing: `loadPastChat` neither cancels nor gates on `loading`, so a turn running across a history switch already writes into the newly-loaded chat. The checkpoint surfaces it sooner but does not create it.

hugocasa:
Both round-2 nits fixed in b4ae11866c — the second one turned out to be load-bearing, so this went in as code rather than a dismissal.

**Pi — the poll does not bound loss in a backgrounded tab.** Correct, and it invalidated the guarantee I had just claimed. A hidden tab pauses `requestAnimationFrame`, so `TypewriterReveal` stalls: text keeps arriving into the buffer while `currentReply` freezes, the fingerprint stops changing, and every later poll short-circuits — with the single hide flush already spent.

Fixed along the lines suggested, and it made the code simpler rather than more complex: the checkpoint now reads `currentReply + replyReveal.pending` (a new non-mutating getter for received-but-unpainted text) instead of the painted text. That advances as tokens arrive regardless of visibility, so the fingerprint and the persisted text are both complete. It also let the `force → replyReveal.flush()` from round 1 be **deleted**: no path in the checkpoint disturbs the typewriter now, which incidentally resolves the tab-switch snapping you'd separately dismissed as cosmetic.

Verified with rAF stubbed out to model a genuinely backgrounded tab: the persisted answer went 5560 chars at hide → 6002 → 7086 → 7580 → 8307 → 8979 across polls. Before the fix it froze at 5560.

**Codex — stale-text dedup can drop a repeated live reply.** Correct. The dedup exists because `partialReply` may already be inside the prefix, but that reasoning never applied to live streaming text, which provably is not. Two iterations around a tool call saying the same sentence twice would have lost the second from both model and display history. `interruptedTurnMessages` now takes a `snapshot` options object carrying `textIsLive`, and live text skips the dedup. Regression test added; it fails without the fix as `expected { role: 'tool', …(2) } to match object { role: 'assistant', …(1) }`.

Also added a unit test pinning that `pending` reports the un-revealed tail without painting or ending the animation — the property that makes it safe to call from a poll on a page the reader is still watching.

`npm run check` 0 errors; 201 tests green across `AIChatManager` / `chatLoop` / `typewriterReveal` / `HistoryManager`.

hugocasa:
Both round-3 findings fixed in b5d2d957c4.

**Codex P1 — buffered tool screenshots lost from a checkpointed half-run batch.** Confirmed, and it is a defect this PR introduced rather than one it exposed: `closeInterruptedToolBatch` (added in round 1) keeps a batch that is still executing, so the restored history could carry `take_screenshot`'s "Screenshot attached below" result while `appendPendingToolImages` had not yet turned the buffered image into a message — announcing a screenshot the model cannot see. Before round 1 the whole batch was dropped, so it was at least self-consistent.

Fixed by snapshotting the buffer non-destructively: `pendingToolImagesMessage` is split out of `appendPendingToolImages` so both build the same message, and the checkpoint reads `pendingToolImages` without draining it — the live turn still owns that buffer. The image is placed after the batch that produced it, matching where the live path puts it. Regression test added; without the fix it fails as `expected [] to have a length of 1`.

**Pi — `textIsLive` only covered actively-streaming text.** Right, and the residual window is real: between `onMessageEnd` capturing `partialReply` and the parser pushing it to `addedMessages`, the text is uncommitted but `currentReply` is already empty, so a segment identical to an earlier one would still have been deduped away.

Replaced the heuristic with a fact rather than widening it: `onMessageEnd` now records `collectedMessages.length` alongside `partialReply`, and the checkpoint treats the text as live while that count is unchanged — i.e. while the parser has not folded it into a message. That covers both the streaming case and the window you identified, and it no longer relies on "live" being a proxy for "not yet committed".

`npm run check` 0 errors. 1001 tests pass across `src/lib/components/copilot/chat/`; the 6 failures there are pre-existing in `ContextManager.test.ts` and `app/core.test.ts` and reproduce on a clean checkout of main.

hugocasa:
All three round-4 nits addressed in 1de2fb1ea5.

**Claude — `partialReplyCollectedLen` recorded after the push, so committed text was treated as live.** Correct, and it was a defect my round-3 fix introduced. @pi-review and Claude reached opposite conclusions on reachability: Pi argued the post-push site is unreachable from a checkpoint because the path to `stopCheckpoints()` carries no macrotask; Claude argued the duplicate is real. Rather than bet on "no macrotask can interleave here", I removed the dependency.

Both call-site orderings exist, which is what made a single length comparison wrong: `anthropic.ts:225` (and `openai-responses.ts:449`) flush *before* any push, when text gives way to a tool call, while `anthropic.ts:349` pushes *first*. Only the capture site can tell the two apart, so it now decides: `onMessageEnd` records whether the last collected message already carries this exact text, and `textIsLive` reads that boolean. No cross-await length invariant, correct under either ordering.

**Pi — the flush-window branch was untested (`partialReplyCollectedLen` stayed at `-1` in every test).** Right, and worth fixing on its own terms: the branch was dead under the suite. The new test drives the real `config.callbacks.onMessageEnd()` at both orderings in one turn and asserts the text appears exactly once either way. Reverting the fix fails it as `expected [...] to have a length of 1 but got 2` — i.e. it now pins Claude's duplicate too.

**Codex — overlong comments.** Fair; several blocks had grown to 5–9 lines of narration across rounds. Condensed to the durable constraint at each site, per AGENTS.md.

`npm run check` 0 errors. 1002 passing in `src/lib/components/copilot/chat/`; the 6 failures are the pre-existing `ContextManager.test.ts` / `app/core.test.ts` ones that reproduce on a clean tree.

hugocasa:
All three of you found the same thing this round, and you were right: `partialReplyCommitted` is computed at flush time and goes stale the moment the parser pushes that text, so a checkpoint during tool execution re-appended the preamble.

Rather than patch the flag a third time, I removed the mechanism. **The checkpoint no longer reads `partialReply` at all** — only live streaming text (`currentReply + replyReveal.pending`).

The reasoning: flushed text is the parser's to push, and it pushes before tool execution begins — so during the long window a checkpoint actually lands in, that text is already in `collectedMessages` and reading it back could only ever duplicate it. What this gives up is the sub-second window between the tool-boundary flush and the message-event push, where a checkpoint would now miss the preamble; the next checkpoint has it. The abort path still recovers that text via `partialReply` in `commitInterruptedTurn`, which is where it was always genuinely needed.

That deletes `partialReplyCommitted`, the `textIsLive` option, and the `onMessageEnd` bookkeeping added in rounds 3–4 — net -38 lines in `AIChatManager`. The dedup in `interruptedTurnMessages` is back to serving only the commit path, which is the one caller whose text can be a stale duplicate.

The round-4 test pinned the mechanism rather than the behavior, so it is replaced by one that pins the invariant you each described: a text-then-tool-call turn in parser order, checkpointed while the tools run, asserting the preamble appears exactly once. Reverting to `streaming || partialReply` fails it as `to have a length of 1 but got 2`.

Verified on a local instance against Anthropic: sampled every checkpoint across a full preamble-then-tool turn (22 samples) — no assistant text ever appeared twice in a persisted transcript.

`npm run check` 0 errors; 1002 passing in `src/lib/components/copilot/chat/` (the 6 failures are the pre-existing `ContextManager` / `app/core` ones).

hugocasa:
Round 6 addressed in f3798f0467.

**Codex / Claude — the guard was incomplete.** Right, and the file already said so: `loading` only rises after a send's attachment upkeep, which is exactly why `sendOrQueue` gates on `sendInFlight` as well (`:1803-1806`). I guarded on `loading` alone and reintroduced the window that precedent exists to close. Now both, in the method and in the disabled state of the History entries.

New test for that specific window — it starts a send, asserts `loading === false && sendInFlight === true`, then attempts the switch. With the old `loading`-only guard it fails as `expected "loadPastChat" to not be called at all, but actually been called 1 times`.

**Pi / Codex — overlong comment.** Fair, and the second time this round-type has caught me on it. Down to four lines, pointing at `sendOrQueue` rather than restating its reasoning.

Also worth correcting something in my earlier framing of this guard: the History popover is inside `{#if !hideHeader}` and `SessionWrapper.svelte:455` passes `hideHeader`, so sessions never render it. This path is reachable from the legacy docked chat only — still a real corruption path, but not the primary surface, and I described it as broader than it is.

`npm run check` 0 errors; 1004 passing in `src/lib/components/copilot/chat/` (the 6 failures are the pre-existing `ContextManager` / `app/core` ones). Verified in the browser on the docked chat: entries disabled with the tooltip while generating, enabled again once the turn ends.

hugocasa:
✅ Review round clean @ f3798f0467ccf1735ea259eaedfa631c3f4d8bc6

hugocasa:
Closed the two gaps I'd left unverified.

**Buffered screenshots — now verified end to end, and this was the right thing to check.** Previously only unit-tested with a faked `attachToolImage` and a mocked `saveChat`, so nothing had exercised the real tool or the blob round trip.

Built a raw app (`f/shot/demo`), got a genuine parallel batch `assistant(take_screenshot, test_run_script)`, and parked the second call on a confirmation so the batch sat half-finished with the image buffered. A checkpoint captured exactly the intended shape: a synthesized result for the parked call plus the images message from `pendingToolImagesMessage`, carrying `wm-image:a1b4bf1a-…`. After reloading in that state, the model answered questions it could only answer by seeing the image — heading wrapping onto two lines, two words, red button. Wire-confirmed on the round trip: an 80 KB base64 PNG in the request body.

Two things that were reasoned rather than observed are now observed: the blob survives dehydrate → store → rehydrate on the checkpoint path, and repeated checkpoints reuse the same blob id instead of accumulating against `MAX_IMAGES_PER_CHAT`.

**OpenAI — previously untested, now verified.** Everything before this was Anthropic, and I'd flagged non-Anthropic replay as an open risk since the Responses API has its own round-trip rules. Ran a gpt-5.1 turn that parked mid-batch, reloaded, and sent a follow-up: `POST /ai/proxy/responses` → **200**, answered correctly from the restored context. The checkpoint had one real tool result and one synthesized, and no `_anthropicContent` — so this exercised the non-Anthropic path properly rather than incidentally.

Still untested, and I'd rather name it than imply otherwise: **switching provider between save and restore** (a chat saved under one provider continued on the other), which takes a reconstruction fallback. Pre-existing behaviour, not introduced here.

Setup note for anyone reproducing: `take_screenshot` needs a raw app preview, and a fresh worktree has no `static/ui_builder/` — run `node frontend/scripts/untar_ui_builder.js` first or the preview renders blank.

PR #10812, opened 2026-08-24
feat(datatables): add a down migration from the migration viewer

diegoimbert:
## Summary

A data table migration with no `down` could only be viewed, never made revertable — the migration viewer's Down tab was a dead end reading "No down migration". This adds the missing authoring step: the empty state now offers **Add a down migration**, which swaps in an editable SQL editor with a Save button.

The backend `upsert_datatable_migration` guard refuses any edit to a migration that has already run, because its `_wm_migrations` record would no longer match its SQL. Filling in an absent `code_down` is the one edit that keeps that record true — the up that ran is untouched — and it is the only way to make an already-applied migration (the `pg_dump` initial snapshot, most notably) revertable. That single case is now exempt; rewriting the name, the up, or an already-recorded down stays refused.

Because that exemption fires only while `code_down` is NULL, attaching a down is a one-shot edit on an applied migration. Two things follow from that, and both are part of this PR:

- The UI refuses to save a draft that holds no statement, so two stray clicks can't freeze an empty transaction in as the down (which would then "revert" the migration by clearing its `_wm_migrations` row while leaving the schema in place).
- The write re-tests the read it was judged on. The exempt paths — add-a-down, and the pre-existing unchanged re-push — skip the run lock and the applied-check, so without a compare-and-set a request working from a stale read could revert a concurrent addition's down, restore an old `code_up` over a locked rewrite, or overwrite a version created in the meantime.

## Changes

- `DataTableMigrationsButton.svelte`: the Down tab's empty state gains an "Add a down migration" button that swaps in a Monaco SQL editor (prefilled with the same `BEGIN; … END;` frame the new-migration modal uses) plus Cancel / Save. Save calls `upsertDatatableMigration`, echoing the migration's existing name and `code_up` unchanged so only the down is added, then shows the saved down read-only. It stays disabled until the draft contains a statement outside the transaction frame and comments — with `leadingChangeSync` on the editor so that enables on the keystroke rather than after the trailing debounce.
- `datatable_migrations.rs`: the applied-migration guard in `upsert_datatable_migration` now also lets through an upsert whose only difference is a `code_down` where there was none, and every lock-skipping write is conditioned on the row still matching what was read (`WHERE NOT $7 OR (name = $8 AND code_up = $9 AND code_down IS NOT DISTINCT FROM $10)`), returning a 400 otherwise.
- New integration test pinning the four sequential arms of the guard.

## Screenshots

Empty state with the new button:

no-down

Editor after clicking it — Save is disabled while the template holds no statement:

editor

Saved — the tab falls back to the read-only view:

saved

## Test plan

- [x] `cargo test --test datatable_migration_add_down` — passes, and fails with the guard condition reverted
- [x] `cargo test --test datatable_migrations_grants` — existing data table migration tests still pass
- [x] `npm run check` — 0 errors; offline `cargo check -p windmill-api-workspaces` green against the regenerated `.sqlx`
- [x] Manual, against a live data table with one applied down-less migration: added a down through the new editor, saved, then reverted the migration through the API and confirmed the table it created was dropped
- [x] Manual: Save is disabled on the untouched template and enables on the first keystroke of a statement
- [x] Manual: rewriting the up of an applied migration, rewriting a down it already has, and dropping its down are all rejected with 400
- [x] Concurrency: six simultaneous additions to the same applied migration — one 200, five 400s, stored down is the winner's. The CAS predicate itself verified directly against Postgres across all seven arms (stale up / stale down / observed-no-row-but-row-exists → 0 rows; fresh read, add-down, lock-held, brand-new insert → 1 row), since it is unreachable without genuine concurrency.

Not covered by an automated test: the CAS. Every interleaving yields "exactly one winner" with or without the predicate, so a test would assert an invariant that holds either way and would only catch a regression by luck.


diegoimbert:
Round 2 follow-up.

**Codex P1 — concurrent down additions could overwrite each other.** Real, fixed in 81a9e1a's successor: the add-down exemption skips the run lock and the applied-check, so two requests could both observe `code_down = NULL` and the later write would silently replace an already-recorded down. The write now re-tests that premise where Postgres can serialize it — `ON CONFLICT … DO UPDATE … WHERE NOT $7 OR datatable_migrations.code_down IS NULL`, with `$7` set only for the exempt path — and a 0-row result returns a 400 telling the caller to reload. Verified against a live applied migration with six simultaneous additions: exactly one 200, five 400s, and the stored down is the winner's.

I did not add a test for it. Any interleaving yields "exactly one 200" whether or not the predicate is there, so an integration test would assert an invariant that holds either way and would only catch the regression by luck — a flaky test, not a guard. The predicate is enforced by the database; the manual six-way check above is the evidence.

**Claude nit — no server-side rejection of a blank `code_down`.** Dismissing. A comment-only `.down.sql` is legitimate input on the `wmill sync push` path this endpoint primarily serves, so rejecting it at the API would break a valid workflow to defend against a UI misclick that the UI now blocks directly (`hasStatement`). Pi reached the same conclusion independently.

**Pi nit — missing `leadingChangeSync` on the new editor.** Dismissing: that prop does not exist. `SimpleEditor.svelte` has no `leadingChangeSync` in its props, `grep -rn leadingChangeSync frontend/src` returns nothing, and neither `flows/header/FlowYamlEditor.svelte` nor `raw_apps/RawAppYamlEditor.svelte` sets it — so the quoted JSDoc and the cited precedents aren't in this tree. The underlying observation is fair (`bind:code` materializes on a 200 ms trailing debounce, so Save enables shortly after the last keystroke rather than instantly), but that is `SimpleEditor`'s normal behaviour everywhere it is used, and adding a new prop to a shared editor component is out of scope here. Confirmed in the browser that Save is disabled on the untouched template and enabled after typing a statement.

diegoimbert:
Round 3 follow-up — Codex P1 and Claude's P2 are the same gap, now closed.

`only_adds_down` and the unchanged-equality arm both assert three things about the row they read (`name`, `code_up`, `code_down IS NULL`) and both then write without the lock; only the third was re-tested. The compare-and-set now covers the whole observed read:

```sql
WHERE $7::text IS NULL
   OR (datatable_migrations.name = $7::text
       AND datatable_migrations.code_up = $8::text
       AND datatable_migrations.code_down IS NOT DISTINCT FROM $9::text)
```

`$7..$9` carry the row as it was read, and are NULL exactly when the lock is held (`existing.filter(|_| _run_lock.is_none())`) — so a locked rewrite still writes unconditionally, while every lock-skipping upsert can only land on the row it was judged against. Both scenarios Codex named now fail closed with a 400: the stale add-down that would restore an old `code_up`, and the unchanged re-push that would erase a down added in between.

Verified the predicate directly against Postgres, since it is unreachable without genuine concurrency:

| observed vs. actual row | rows affected |
|---|---|
| stale `code_up` | 0 |
| stale `code_down` (saw NULL, row now has one) | 0 |
| fresh read, rewriting a recorded down | 1 |
| fresh read, adding a down | 1 |
| lock held (`$7` NULL) | 1 |

And re-ran the six-way concurrent addition against the live API: one 200, five 400s, stored down is the winner's.

Still no test for this, for the reason Claude states in its round-3 coverage note: the predicate can't be reached without real concurrency, so a test would assert an invariant that holds with or without it.

Pi's `leadingChangeSync` nit is a re-post of round 2's; dismissed there — the prop does not exist in this tree (`grep -rn leadingChangeSync frontend/src` is empty, and neither cited precedent file sets it).

diegoimbert:
Round 4 follow-up, plus a correction I owe Pi.

**I was wrong to dismiss `leadingChangeSync`, and my reason for dismissing it was wrong too.** I grepped my working tree — which was 319 commits behind `main` — and concluded the prop didn't exist. It does: `SimpleEditor.svelte:144`, implemented at `:451`, and both YAML editors set it. Pi's premise was correct and my "the prop does not exist in this tree" was an artifact of a stale checkout, not a fact about the codebase. The branch is now merged up to `main` and the prop is set on the down-draft editor; verified in the browser that Save enables on the keystroke rather than after the 200 ms trailing debounce.

**Claude P2 — `$7 IS NULL` conflated "lock held" with "no row observed".** Fixed as suggested: `$7` is now the boolean "re-test?" (`_run_lock.is_none()`) and `$8..$10` carry the observed row from `existing`, so a `None` observation makes the equality NULL and fails closed. That closes the last arm — a version created and applied concurrently is now refused instead of overwritten.

Predicate re-verified directly against Postgres, all seven arms:

| observed vs. actual row | rows affected |
|---|---|
| stale `code_up` | 0 |
| stale `code_down` | 0 |
| observed no row, row exists (concurrent create) | 0 |
| fresh read, rewriting a recorded down | 1 |
| fresh read, adding a down | 1 |
| lock held (`$7` false) | 1 |
| brand-new insert, no conflict | 1 |

Merged `origin/main` into the branch (it was 319 behind, so CI was reviewing a merge ref I had never built). Re-validated on the merged tree: `cargo test --test datatable_migration_add_down` passes, `npm run check` is 0 errors after regenerating the backend client, offline `cargo check -p windmill-api-workspaces` is green, and the full UI flow was re-driven end to end.

Note for anyone re-running `update_sqlx.sh` here: it needs the EE worktree at the pinned `ee-repo-ref.txt` **and** hand-made symlinks for `*_ee.rs` files added since the worktree was created (`feature_usage_ee`, `duckdb_isolation_ee`, `storage_list_ee`). Without them `cargo sqlx prepare` fails partway and drops thousands of cache entries.

No Codex verdict was posted for the previous head, so this round needs a fresh one.

diegoimbert:
✅ Review round clean @ a4a17034ab719e84a8990a890a72cfa97c12c8ad

PR #10816, opened 2026-08-24
fix(ai-chat): keep the composer usable while a question is pending

diegoimbert:
## Summary

In AI Sessions, when the agent called `askUserQuestion`, the main chat composer was disabled until the question was answered. The question card also grabs focus on mount (for its arrow-key navigation), so focus was pulled out of the composer and clicking it back did nothing — a dead end unless you answered from the card or pressed Escape to abort the turn.

The composer now stays enabled while a question is parked, and a plain-text send from it answers the question instead of being queued behind a turn that only the answer can resume.

## Changes

- `AIChatDisplay.svelte`: drop `hasActiveUserQuestion` from the composer's `disabled`; pass the parked question's `tool_call_id` down instead.
- `AIChatInput.svelte`: a plain-text draft sent while a question is parked resolves it via `handleUserQuestionAnswer`. Placeholder reads "Answer the question above", and the send/stop button flips back to Send once there is text to ship (an empty composer keeps Stop). Drafts carrying attachments can't ride a text answer, so they fall through to the existing queue and flush when the turn resumes.
- `AIChatManager.handleUserQuestionAnswer` now returns whether the answer reached a waiting resolver. A card restored from history still looks parked but its promise left with the old page; the composer holds the only copy of the typed text, so it only clears the draft on a `true`.
- `shared.ts`: `pendingUserActionDetail` — the same backward scan as `pendingUserAction`, carrying the blocked card's `tool_call_id`.
- `MessageDraft.hasAttachments`.

## Test plan

- [x] `shared.test.ts`, `AIChatManager.test.ts`, `messageDraft.test.ts` — 180 passed
- [x] Live session on `/sessions` with Sonnet actually calling the tool: composer enabled, click + type + Enter answers the question, card collapses to `Asked: … — <text>`, turn resumes
- [x] Same with `multiSelect: true`
- [x] Answering by clicking a card choice still works (unchanged path)
- [x] Before the fix, the same script recorded `textarea.disabled = true` and typing produced nothing

## Screenshots

Before — question pending, composer disabled, typing does nothing:

pr-before-2-typed

After — composer live, placeholder "Answer the question above", Send button once typed:

pr-after-2-typed

After — Enter delivers it as the answer and the turn resumes:

pr-after-3-answered

## Notes

`local-review` ran and its two P2 nits are fixed in 38eb0ad. `local-review-codex` was skipped: the local `codex` CLI is 0.132.0 and the skill requires >= 0.144.1.


diegoimbert:
✅ Review round clean @ 197ae1bffe78b33915de08c63e639f05ec9107db — nit-only verdicts at 10faa5a4d548ac5e8763c85f5b128bc7eb15e509 (codex: nits, claude: good to merge, pi: good to merge); Codex's comment-length nit addressed in 197ae1b.

Also ran the two manual checks Codex asked for:
- **Collapsed large paste sent as an answer**: pasted a 14-line blob (collapses to a `Pasted 14 lines` token) plus trailing text while a question was parked — the readback shows all 14 lines expanded followed by the typed text, so `expanded(chatDraft(…))` behaves the same on the answer path as on a normal send.
- **Reload with a parked question**: on `/sessions`, reloading mid-question does not restore the question card at all — only the user message comes back, and the composer correctly reads "Ask followup", not "Answer the question above". So the affordance can't lie in this path; the `loading` gate and the boolean return from `handleUserQuestionAnswer` remain as guards for any path that does restore a parked-looking card (e.g. `loadPastChat`), which I could not reproduce.

PR #10818, opened 2026-08-24
fix: unify billable seat counting and prevent fork subscriptions

hugocasa:
## Summary

Two faults in cloud billing, both of which caused workspaces to be invoiced for something the product does not charge for.

**1. Seat counting disagreed between the product and the billing job.** A workspace's billable seat count is computed in the backend (`count_paid_seats`), in the workspace settings billing panel, in the sidebar usage meter, and in the out-of-repo billing job. The three in-product ones excluded service accounts and disabled members; the billing job excluded only disabled members. Since `create_service_account` takes `operator` from the caller rather than forcing it, a non-operator service account landed in the developer tally, so an affected workspace was invoiced for a seat every in-product surface said it did not have. The docs already state service accounts do not consume a seat.

The billing job has been corrected separately. This PR removes the two in-product duplicates by having the meters read the count from the backend rather than recomputing it from the member list.

**2. A fork could carry its own subscription.** A fork/dev workspace draws its plan from its billing root and meters its executions there, so it is never invoiced separately — stated in the docs and enforced everywhere in the runtime. But nothing stopped one carrying a Stripe subscription of its own, and the settings page replaced its billing panel with the parent-managed notice, so an admin in that state could not reach the portal to cancel it. Attaching an already-subscribed workspace is now refused, checkout from a fork is refused, and a leftover subscription is surfaced with a working portal link.

## Changes

- `billable_seats` helper returns the developers/operators/seats breakdown; `count_paid_seats` delegates to it. The rule stays in Rust — the billing job keeps its own SQL, and the doc comment records that the two must change together.
- New `GET /w/{workspace}/workspaces/billable_seats`, readable by any workspace member, resolving a fork to its billing root server-side. The breakdown is returned only when the seats counted are the requested workspace's own; a fork gets the total alone, since a fork member need not be a member of the root.
- `attach_dev_workspace` refuses a candidate that already has a plan of its own. Scoped to candidates not already under the same billing root, so re-designating a renamed dev workspace is not blocked by a billing state the attach did not cause. Placed after the admin-of-candidate check so the refusal cannot be used to probe an arbitrary workspace's plan. Enterprise arrangements are deliberately not covered: they set `premium` without a plan and have no self-serve portal, so refusing there would be a dead end.
- `stripe_checkout` refuses a forked workspace (EE, see companion PR).
- The fork billing tab shows a leftover subscription with a portal link instead of asserting there isn't one.
- `SidebarUsage` and `PremiumInfo` read the endpoint instead of recounting the member list. This also fixes the meter staying hidden for a fork member who is not a member of the root.

Note: several hunks in `PremiumInfo.svelte` are the format hook reflowing pre-existing text, not logic changes.

## Screenshots

Fork billing tab, with the leftover subscription surfaced and the portal link that is the entire remedy for fault 2:

fork-billing

Root workspace seat panel (4 members billable out of 5, one being a service account):

root-seats

Sidebar meter for a fork member, resolving the root's seats server-side:

fork-sidebar-meter

## Test plan

- [x] `cargo test -p windmill-common --features cloud --test billing_workspace` — seat weighting, service-account and disabled exclusion, and the developers/operators breakdown.
- [x] `cargo test -p windmill-api-workspaces --features cloud,enterprise,private attach_billing` — the attach guard refuses a candidate with a plan and permits one with no plan or no settings row.
- [x] `cargo check --all-targets --features cloud,enterprise,private,stripe,quickjs` clean; `npm run check` 0 errors; sqlx offline cache validated.
- [x] Exercised against a running backend built `quickjs,cloud,enterprise,private,stripe` with `CLOUD_HOSTED=true`: `billable_seats` on a root returns the full breakdown and excludes a planted non-operator service account; on a fork it returns the total alone; attaching a subscribed candidate 400s while an unsubscribed one succeeds; checkout from a fork 400s; an admin of the parent who is not a member of the candidate gets a generic 403 rather than the plan-specific message.

## Known gaps

**Checkout race.** Creating a Stripe checkout session does not set `workspace_settings.plan`, so an admin could open checkout, attach the workspace while `plan` is still null, and complete the session afterwards — recreating the double-billed state. The backend never observes subscription creation, so the guard belongs in the webhook flow that records it (refuse to set `plan`/`premium` for a workspace with a `parent_workspace_id`), not here. The settings-page warning makes the resulting state visible and self-serviceable in the meantime.

**Neither new test runs in CI.** Both this PR's guard test and the pre-existing `windmill-common/tests/billing_workspace.rs` suite are gated on `feature = "cloud"`, which the `cargo_test` job does not enable — so they only run locally under `--features cloud`. I tried adding `cloud` to that job's feature list and reverted it: the feature is not behaviour-neutral, because `jobs.rs:6434` reserves a root prefix on the git-sync concurrency key under `#[cfg(feature = "cloud")]` rather than behind `*CLOUD_HOSTED`, so three `workspace_dependencies_git_sync` tests that assert the bare `{workspace}:git_sync` key fail. Making those cloud-aware and turning the feature on in CI is worth doing, but it is its own change rather than something to fold into a billing fix.


hugocasa:
✅ Review round clean @ 15e305e4088952b39f357f763e702bc99cd0cdc2 — nit-only verdicts at 6e7dd598c379e61b86ce61058b860309dcd181c9; nits addressed in 15e305e408

hugocasa:
✅ Review round clean @ 4073fd6501f057cb4445e94d056f814111a2d82d — nit-only verdicts at 6e7dd598c379e61b86ce61058b860309dcd181c9; nits addressed in 15e305e408, plus 4073fd6501 reverting the CI test feature change (that file is now identical to main). Supersedes the earlier marker naming 15e305e408.

PR #10819, opened 2026-08-24
fix: key build artifact caches on a runnable's inline modules

hugocasa:
## Summary

Windmill's build-artifact caches are global and content-addressed: the bun bundle, the
rust/go/csharp binaries and the java jar are each looked up by a hash of the runnable's own
content plus its lockfile. A runnable's inline modules were not part of that hash, even though
`write_module_files` puts their content in the job dir, where the build inlines or compiles it
into the very artifact being cached.

Two runnables whose main content and lockfile match therefore shared one cache slot no matter
how their modules differed, so whichever built first decided what the other ran — across
workspaces, the cache being global.

## What the cache name is made of

Everything that was already in the name is unchanged, byte for byte. The only addition is the
module map, and only for a runnable that has one.

| Language | Base fed to the hash | Changed by this PR |
|---|---|---|
| bun | `content` + `lock` + transitive relative imports + temp script refs + workspace-registry suffix | no |
| rust | `code` + `requirements` (registry suffix appended after) | no |
| csharp | `code` + `requirements` + `DOTNET_TARGET_FRAMEWORK` (registry suffix after) | no |
| go | `code` + `{:?}MaybeLock` + `"v2"` | no |
| java | `code` + `requirements` (registry suffix after) | no |

| Module field | In the name | Why |
|---|---|---|
| path | yes, length-prefixed | names the file written into the job dir |
| content | yes, length-prefixed | this is what gets bundled/compiled in |
| `language` | no | `write_module_files` writes content at path and nothing else, so equal path+content give an identical job dir |
| `lock` | no | deploy regenerates it *after* the parent has prebuilt; naming it would strand every prebuilt artifact |
| map order | normalised (sorted by path) | iteration order must not move the name |

| Runnable | Name before | Name after | Cache impact |
|---|---|---|---|
| no modules (the overwhelming majority) | `<sha256>` | `<sha256>`, byte-identical | none — every cached artifact stays reachable |
| has modules | `<sha256>`, shared with the row above | `mod-<sha256>` | rebuilt once; these are exactly the ones the fix must re-key |

## Changes

- Add `artifact_cache_name` in `windmill-worker/src/worker.rs`, and derive the name through it in
  every artifact cache a module can reach: `compute_bundle_local_and_remote_path` (bun),
  `compute_rust_hash`, `csharp_cache_key`, `go_cache_key`, and java's `compute_hash`.
- Thread `modules` to both the run path and the deploy-time prebuild for each, so the two still
  derive the same name.
- Seal the base with a hash before appending the module block rather than concatenating the two.
  The base ends in caller-supplied bytes on a preview (its own source *and* lockfile), and plain
  concatenation would let one spell out another runnable's block and reach its slot.
- Put module-bearing artifacts in a `mod-` namespace. It is a prefix on the stored name, not on
  the hash preimage: a preimage is chosen by the caller, so a versioned preimage is one they could
  have fed the old hash beforehand. `calculate_hash` emits hex, so the namespace is unreachable
  from a module-free name however the source and lockfile are crafted.

  **Accepted residual:** keeping the module-free name stable is two-sided. A *pre-existing*
  multi-file artifact also sits under the bare name, so a single-file runnable with matching
  content and lockfile can still load one. That is the original issue rather than new exposure,
  and once this ships nothing can be stored there with module content again. Abandoning the
  keyspace would close it at the cost of one rebuild per deployed Rust/C#/Java script
  instance-wide; that trade was weighed and declined. An operator who wants it closed can purge
  the per-language object-store prefixes.
- Skip the deploy-time bun prebundle for a multi-file script: the dependency job has no
  module-writing step, so a bundle built there resolves relative imports remotely rather than from
  the modules, and must not be cached under a name that says otherwise.
- EE (`windmill-api-agent-workers`): the server precomputes the bun bundle path for agent workers,
  which never derive it themselves, so `get_code_and_lock` must resolve modules exactly as the
  worker does. It now (a) falls back to `_MODULES` in the args for *every* job kind when stored
  modules are absent, not only previews, and (b) serves a preview carrying a `runnable_id` from
  the script data — code, lock and modules — as the worker does, instead of from the job's raw
  code. Resolving them any other way leaves the agent building content the precomputed name does
  not cover.

Nothing on the agent-worker wire changed: `PrecomputedAgentInfo` is untouched, and agents consume
the precomputed `{local, remote}` verbatim, so an agent on an older version cannot disagree with
the server about where an artifact lives.

## Test plan

- [x] `cargo check` and `cargo test` clean under `-D warnings` (as CI runs them) on default,
      `quickjs,rust,csharp,java`, and `enterprise,private`
- [x] `cargo test -p windmill-worker` — 221 pass, including unit tests pinning the encoding's
      unambiguity, the namespace separation, the base seal, lock-independence, and the unchanged
      module-free name. The seal test is mutation-checked: it is the only test that fails if the
      seal is reverted to a bare base.
- [x] **All five languages exercised end to end**, each in two workspaces sharing a body and
      lockfile but differing in module content, and each checked against a deliberately unfixed
      build:

      | | unfixed | fixed |
      |---|---|---|
      | bun | one workspace ran the other's module | each ran its own |
      | rust | one workspace ran the other's module | each ran its own |
      | go | one workspace ran the other's module | each ran its own |
      | csharp | one workspace ran the other's module | each ran its own |
      | java | one workspace resolved to the other's cache entry | each ran its own |

- [x] Repeat runs still hit the cache in every language
- [x] A module-free bun script cached *before* this change still logs
      `BUN BUNDLE SNAPSHOT EXECUTION` on its first run after it — existing artifacts stay reachable
- [x] **Object store** (`enterprise,private,parquet` against MinIO): each workspace's bundle lands
      under its own `bun_bundle/mod-…` key, and with the local cache emptied each run pulls *its
      own* artifact back — the cross-worker channel, not just local disk
- [x] **EE agent worker**: a real agent (`ag-…`) running a deployed bun script, checked against an
      unfixed server. Unfixed, a caller-supplied `_MODULES` landed in the module-free slot and both
      subsequent clean runs returned the injected value; fixed, the injected run gets its own slot
      and clean runs return the deployed helper.
- [x] **Under nsjail** (`isolation=nsjail`): the two workspaces derive distinct `mod-` names, push
      to the object store, and each reloads its own on a repeat run

## Not covered

The agent preview-by-hash path is fixed but could not be exercised: it needs a multi-file script
that has a lockfile, and deploying one fails today in `gen_bun_lockfile` with
`Failed to find relative import`, because the dependency job has no module-writing step. That is
pre-existing and out of scope here (it is also why the deploy-time prebundle is skipped for
multi-file scripts). The equivalent on the deployed-script path was exercised on a real agent.


hugocasa:
On the retained-artifact residual raised this round (codex P1, claude/pi P2) — this is a deliberate trade, now stated in both the code and the PR description rather than left implicit.

Abandoning the keyspace does close it, and an earlier revision of this PR did exactly that (`m1-` on every artifact). It was reverted: the cost is one rebuild for every deployed Rust/C#/Java script instance-wide, borne almost entirely by runnables that never had modules and were never at risk, to remediate a compromise that would have had to happen before the upgrade. Reaching a stale entry also requires a runnable whose content and lockfile match byte-for-byte and which calls into a module it does not declare.

What ships instead: post-fix nothing can be stored under a module-free name with module content in it, so the residual is bounded to what is already on disk and cannot grow. An operator with reason to suspect prior exploitation can purge the per-language object-store prefixes (`bun_bundle/`, `<target>_gobin/`, `<target>_csharpbin/`, `java_jar/`, the rust prefix), which is a cheaper and more targeted remedy than forcing every instance to rebuild.

The other finding from this round was a genuine gap and is fixed in 8d8065e: rewriting the tests last round dropped the only guard on the base seal. `a_module_bearing_runnable_cannot_absorb_another_ones_block` restores it, using the concrete two-module-bearing forgery vector, and it is the only test in the suite that fails when the seal is reverted to a bare `base`.

hugocasa:
Both findings from this round are fixed in fddf1d4 / ee efce983.

**pi (P1) — the EE preview branch resolved modules differently from the worker.** Correct, and it was the branch I got wrong: for a preview carrying a `runnable_id`, the worker serves code, lock *and* modules from the script data (`worker.rs`, `JobKind::Preview`), not from the raw code the job carries. The precompute now mirrors that, gated on `hash_to_codebase_id` exactly as the worker is, and falls back to the raw code plus `_MODULES` only for an inline preview. That also removes a pre-existing divergence in the same branch, where the name was derived from `raw_code` while the worker built from the fetched script.

Worth recording that this path is currently **latent** rather than reachable: it needs a multi-file script that has a lockfile, and a multi-file bun script cannot get one today — deploying one fails in `gen_bun_lockfile` with `Error locking file: ... Failed to find relative import`, because the dependency job has no module-writing step. That is pre-existing and out of scope here (it is also why this PR skips the deploy-time prebundle for multi-file scripts). So the fix is forward-looking, and I could not demonstrate it end to end; the equivalent on the deployed-script path *was* demonstrated on a real agent worker, pre-fix and post-fix.

**claude (P2) — the `_MODULES` fallback was documented as preview-only.** The comment said "For preview jobs" while the code applies to every job kind, and the EE precompute now depends on that being true for all of them. The comment now says so and points at `get_code_and_lock`, so the coupling is visible from the side that would break it.

hugocasa:
✅ Review round clean @ c88cd5bdcb045544fb399b8207099744d4374723 — all three reviewers "Good to merge" at fddf1d4a826227e41c5652ed9136f7c277ee1d9e; head moved only to drop two imports left unused by the rename (c88cd5b), which is what CI's `-D warnings` flagged. No behaviour change since the reviewed SHA, and CI is green on this head.

PR #10820, opened 2026-08-24
fix: reset pooled connections left in an untracked transaction

rubenfiszel:
## Summary

A production incident had **every flow on one worker pod failing for ~20 minutes** with:

```
Internal: retrieval of args from state: error returned from database:
current transaction is aborted, commands ignored until end of transaction block
```

The same `25P02` surfaced at ~10 unrelated call sites (job pull, flow status, job logs, variables, scripts, the workspace error handler), some as critical alerts. The pod never restarted — it healed on its own after `max_lifetime` recycled the pool.

**Root cause.** `jobs.rs:4394` (the failing pull) is a plain pool query with no transaction of its own, so it can only report `25P02` if the connection came out of the pool already inside a failed transaction. It did:

1. sqlx's `PgTransactionManager::begin` builds its `Rollback` drop-guard *before* sending `BEGIN`, but `start_rollback` is a no-op while `transaction_depth == 0` — and depth is incremented only *after* the round-trip await (`sqlx-postgres-0.8.6/src/transaction.rs:31-38`). A future cancelled during that await (a disconnecting API client, a `tokio::time::timeout`, an aborted task) leaves the server session inside an open transaction while sqlx believes it is not in one.
2. `Floating::return_to_pool` calls `raw.ping()`, which for Postgres is a bare `wait_until_ready` and never inspects the `ReadyForQuery` transaction-status byte — so the connection goes **back into the pool still in a transaction**.
3. The next borrower's statements silently run inside it and hold their row locks indefinitely → lock waits on `worker_ping`, then a 4-way deadlock (`40P01`).
4. That error flips the session to `idle in transaction (aborted)`, and every unrelated query on it fails with `25P02` from then on.
5. Nothing heals it: depth is still 0 so no `ROLLBACK` is ever queued, and `idle_in_transaction_session_timeout` never fires on a constantly-used connection. Only `max_lifetime` (30 min) recycles it.

`PgConnection::in_transaction()` exists but is `pub(crate)`, and the public `Connection::is_in_transaction()` reads sqlx's *client-side* depth — which is exactly the value that is wrong here — so a free check genuinely is not available to callers.

## Changes

- `windmill-common/src/db.rs` — new `connection_reset` module: a pool hook that issues `ROLLBACK`, ending both a leaked-open and an aborted transaction. Rolling back on **every** checkout costs a round trip per query (measured at ~37% of throughput on small ones), so it is **armed** only once Postgres has reported a state that proves a connection is carrying leftover transaction state, for a 60s window. Steady state is one relaxed atomic load, and the 0 → armed transition logs a `warn!` so the window is attributable.
- Hook runs on **`before_acquire`**, not `after_release`: the connection that reports the first `25P02` is released *before* its error has been converted, so a release-side hook lets exactly that connection back into the idle queue uncleaned.
- `windmill-common/src/error.rs` — arm from `From<sqlx::Error>` and from `to_anyhow` (a downcast, so it is a type-id compare), which between them cover the paths a `sqlx::Error` takes on its way out of a query.
- `windmill-common/src/tracing_init.rs` — filter `sqlx::postgres::notice` to `ERROR`. A `ROLLBACK` on a clean session makes Postgres answer `there is no transaction in progress`, which sqlx logs at WARN; sqlx only ever talks to Windmill's own database, so notices on that target are never operator-actionable.
- `windmill-common/tests/connection_reset.rs` — regression test.

Net effect: a 20-minute pool-wide outage becomes a single failed query.

## Test plan

- [x] `cargo test -p windmill-common --test connection_reset` passes
- [x] Test is load-bearing — verified it fails with the customer's exact error when the hook is removed, and also fails when the arming wiring is removed
- [x] Steady-state cost measured over 3 interleaved rounds: 209,913 → 210,100 ops/s (**+0.09%**, i.e. noise). Armed window: 132,645 ops/s (−36.8%), only after a `25P02` has already proven the pool is broken
- [x] `before_acquire` vs `after_release` verified against a real pool: with the arming losing the race, `after_release` leaves the next borrower broken while `before_acquire` cleans the connection
- [x] Exercised on a running backend: multi-step flow with a nested for-loop completes (`success: true`, `result: [2,4,6]`)

The leak itself is a genuine sqlx cancel-safety hole and still happens; this bounds its blast radius from 30 minutes to one query. Closing it entirely would mean routing all ~188 `begin()` call sites through a cancel-safe wrapper, which is worth doing separately.

---
## Summary by cubic
Reset pooled Postgres connections left in an untracked or aborted transaction to contain pool-wide failures. Before: a connection could re-enter the pool mid-transaction and poison borrowers, cascading into `25P02` until `max_lifetime`. Now: after the first `25P02`, we `ROLLBACK` on acquire for 60s so only the triggering query fails.

- Wire the pool `before_acquire` hook to `windmill_common::db::connection_reset::reset_before_acquire` so cleanup happens on checkout and never re-queues a poisoned connection.
- Arm the reset automatically when `25P02` is seen via `From<sqlx::Error>` and `to_anyhow`; remove per-site arming calls.
- Steady state has no extra RTT; during the 60s window each checkout adds one RTT and logs a single `warn!` on arming.
- Filter `sqlx::postgres::notice` to ERROR across all sinks, including the OTEL logs bridge, to suppress benign “no transaction in progress” notices while the window is armed.
- Add regression test `backend/windmill-common/tests/connection_reset.rs` to pin arming and cleanup behavior.

<sup>Written for commit a0ee61992cb0b2ec70795bdafe9498fec32452f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10820?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## On detection coverage (addressing the round-1 P1)

An earlier revision added a `note_sqlx_error` call at the flow-args query, the one path from the incident report that formats its `sqlx::Error` into a message instead of converting it. That call has been **removed**, deliberately.

That query is 1 of **39** sites using the same lossy idiom (`sqlx result → .map_err(|e| Error::internal_err(format!("…{e:#}")))`). Patching one of them is arbitrary — it implies a per-site contract that does not exist, and would mislead the next reader into adding more. (The ~10 sites that `match` on `sqlx::Error` variants are already fine: they test one specific code and fall through to `Error::from` for everything else, so a `25P02` there arms normally.)

It is also not load-bearing, because **arming is process-wide, not per-query**. The flag covers a pool shared by everything in the process, so it does not matter which caller notices a poisoned connection, only that one does. `pull()` returns `windmill_common::error::Result` and uses `?` on its query — that is `jobs.rs:4394`, the exact line from the incident — so it converts, and it runs every few tens of milliseconds on every worker. For arming to be missed outright, every `25P02` in the process would have to land only on lossy sites while the job poller hammers the same pool.

A missed site therefore delays arming until the next converting query, not indefinitely. The reasoning now lives in the `connection_reset` module docs so it does not have to be rediscovered.

### DIFF
--- a/backend/src/db_connect.rs
+++ b/backend/src/db_connect.rs
@@ -153,6 +153,15 @@ pub async fn connect(
         pool_options = pool_options.idle_timeout(Duration::from_secs(60));
     }
     pool_options
+        // Clears transaction state sqlx is not tracking, so a connection whose session was
+        // left inside a transaction is not handed to the next borrower. See
+        // `windmill_common::db::connection_reset` for how such a session comes about and why
+        // this only runs once one has been observed.
+        .after_release(|conn, _| {
+            Box::pin(windmill_common::db::connection_reset::reset_on_release(
+                conn,
+            ))
+        })
         .after_connect(move |conn, _| {
             if worker_mode {
                 Box::pin(async move {
--- a/backend/src/db_connect.rs
+++ b/backend/src/db_connect.rs
@@ -153,12 +153,12 @@ pub async fn connect(
         pool_options = pool_options.idle_timeout(Duration::from_secs(60));
     }
     pool_options
-        // Clears transaction state sqlx is not tracking, so a connection whose session was
-        // left inside a transaction is not handed to the next borrower. See
+        // Clears transaction state sqlx is not tracking, so a session left inside a
+        // transaction is cleaned before the connection is handed to a borrower. See
         // `windmill_common::db::connection_reset` for how such a session comes about and why
         // this only runs once one has been observed.
-        .after_release(|conn, _| {
-            Box::pin(windmill_common::db::connection_reset::reset_on_release(
+        .before_acquire(|conn, _| {
+            Box::pin(windmill_common::db::connection_reset::reset_before_acquire(
                 conn,
             ))
         })
--- a/backend/windmill-common/src/db.rs
+++ b/backend/windmill-common/src/db.rs
@@ -331,12 +331,14 @@ pub mod connection_reset {
     }
 
     fn arm() {
-        let previous = RESET_UNTIL_MS.fetch_max(
-            now_ms() + RESET_WINDOW.as_millis() as u64,
-            Ordering::Relaxed,
-        );
-        // Entering a window costs throughput, so it must be attributable in the logs.
-        if previous == 0 {
+        let now = now_ms();
+        let previous =
+            RESET_UNTIL_MS.fetch_max(now + RESET_WINDOW.as_millis() as u64, Ordering::Relaxed);
+        // Every window costs throughput, so each one must be attributable in the logs —
+        // hence the comparison against `now` rather than against zero, which would report
+        // only the first incident a process ever sees. Re-arming inside a live window is
+        // the common case and stays quiet.
+        if previous <= now {
             tracing::warn!(
                 "a pooled connection was found in an aborted transaction; rolling back on \
                  checkout for the next {}s",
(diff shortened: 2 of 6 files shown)


rubenfiszel:
Superseded by #10823, which fixes the cause instead of containing it.

This PR healed a pooled connection already stuck in an aborted transaction. #10823 patches sqlx so `Pool::begin` is cancel-safe on Postgres, so the connection never reaches that state — and it carries a test that fails if the patch is ever dropped, which is a stronger guard than this PR's best-effort, process-wide arming.

Branch `fix/pool-connection-tx-reset` is left pushed. Worth reviving if a cause outside `begin` turns up: a pooler in front of Postgres, or a multi-statement `BEGIN; …; COMMIT;` batch that fails mid-way (reproduced, but Windmill does not run one on its own pool today). Upstream report: transact-rs/sqlx#4393.

PR #10822, opened 2026-08-24
fix: make Pool::begin cancel-safe across the backend

rubenfiszel:
## Summary

Independent of #10820, which fixes the same incident from the other end: that PR *heals* a pooled connection whose session was left inside a transaction sqlx is not tracking; this one stops that state arising. Either can merge first — they touch no files in common.

`Pool::begin` is not cancel-safe. `PgTransactionManager::begin` builds its `Rollback` drop-guard before sending `BEGIN`, but `start_rollback` is a no-op while `transaction_depth == 0`, and depth is raised only *after* the round-trip await (`sqlx-postgres/src/transaction.rs:31-38`). A future cancelled in between — a disconnecting API client, a `tokio::time::timeout`, an aborted task — leaves the session inside a transaction that nothing will ever roll back, and `return_to_pool`'s `ping` (a bare `wait_until_ready`) does not look at the `ReadyForQuery` transaction-status byte, so that connection goes back into the pool. The next borrower's statements then run inside it and hold their locks indefinitely, which is how a transient database hiccup turned into a 20-minute outage.

Upstream has known since 2022 ([launchbadge/sqlx#2054](https://github.com/launchbadge/sqlx/issues/2054), labeled `bug`), fixed it for **SQLite only** ([#2057](https://github.com/launchbadge/sqlx/issues/2054)), and closed the issue. Verified against sqlx `main` today: Postgres still has it.

**The fix** is to run the `begin` on a task. Dropping a `JoinHandle` detaches its task rather than aborting it, so a cancelled caller still leaves the `Transaction` to be built and then dropped — and dropping one at depth 1 is exactly what queues the `ROLLBACK`.

## Changes

- `windmill-common/src/db.rs` — new `BeginCancelSafe` trait, implemented for `Pool<Postgres>` only, so it cannot be misapplied to a `Transaction` (nested `begin` is a `SAVEPOINT` and is already guarded by a non-zero depth).
- Swept every pool-level `begin()` in the backend onto it — 49 files. `tx.begin()` (savepoints) and `migrator.connection().begin()` are deliberately untouched.
- `UserDB::begin` and the two `Acquire` impls in `db.rs` go through it too, so RLS transactions are covered.
- `windmill-common/tests/begin_cancel_safe.rs` — regression test.

No `*_ee.rs` file is touched, so there is no companion EE PR.

## Test plan

- [x] `cargo test -p windmill-common` — 531 unit tests + all integration tests pass
- [x] New test is load-bearing: swapping `begin_cancel_safe()` back to `begin()` fails it with `left: Some("SELECT pg_backend_pid()")`, `right: Some("ROLLBACK")` — i.e. it detects that the work died with its caller instead of completing. It polls the future exactly once via `now_or_never()` rather than using a short timeout, because tokio's timer granularity lets the begin win that race (a `Duration::ZERO` timeout does **not** cancel, which is worth knowing before "simplifying" it)
- [x] Mechanism verified directly against a real pool: cancelling a plain `begin` leaves the session `idle in transaction`; cancelling the spawned one leaves it `idle`
- [x] Cost measured at −9.6% on a tight begin+commit loop on localhost, where a begin+commit is ~27µs and the spawn is ~1–3µs of fixed overhead. Against a real 1–2ms round trip that is under 0.2%
- [x] Exercised on a running backend: workspace create, script deploy, group create, script run (`21`), multi-step flow with a nested for-loop (`[2,4,6]`)

## Note on the import placement

The trait import is added as the first `use` in each file rather than in sorted position. `cargo fmt` is not enforced in CI and the tree is not currently fmt-clean (there are pre-existing diffs in files this PR never touches), so running it would have buried a mechanical sweep in unrelated churn.

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -5,6 +5,7 @@
  * Please see the included NOTICE for copyright information and
  * LICENSE-AGPL for a copy of the license.
  */
+use windmill_common::db::BeginCancelSafe;
 use anyhow::Context;
 use monitor::{
     load_base_url, load_otel, reload_critical_alerts_on_db_oversize,
@@ -2401,7 +2402,7 @@ async fn coordinate_restart_delay(
 
     let now = chrono::Utc::now();
 
-    let mut tx = db.begin().await.context("begin restart coordination tx")?;
+    let mut tx = db.begin_cancel_safe().await.context("begin restart coordination tx")?;
 
     // Serialize access across all instances
     sqlx::query("SELECT pg_advisory_xact_lock($1)")
--- a/backend/src/monitor.rs
+++ b/backend/src/monitor.rs
@@ -1,3 +1,4 @@
+use windmill_common::db::BeginCancelSafe;
 #[cfg(feature = "oauth2")]
 use std::collections::HashMap;
 use std::{
@@ -1837,7 +1838,7 @@ async fn cleanup_scheduled_job_deletions(db: &Pool<Postgres>) {
 
     let mut total_deleted = 0u64;
     for batch_num in 0..MAX_BATCHES {
-        let mut tx = match db.begin().await {
+        let mut tx = match db.begin_cancel_safe().await {
             Ok(tx) => tx,
             Err(e) => {
                 tracing::error!("Error starting transaction for scheduled job deletion: {e:?}");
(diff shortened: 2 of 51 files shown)


rubenfiszel:
Superseded by #10823.

This PR routed ~173 `Pool::begin` call sites through a spawn-based cancel-safe wrapper. #10823 gets the same guarantee from a 4-line patch to sqlx itself, which is strictly better: it cannot be forgotten at a new call site, costs nothing at runtime, and deletes itself when upstream carries the fix (transact-rs/sqlx#4393).

Branch `fix/cancel-safe-begin` is left pushed in case the patch route stalls.

PR #3980, opened 2024-06-26
chore(main): release 1.355.0

rubenfiszel:
:robot: I have created a release *beep* *boop*
---

### Features

* add wmill cli image ([fa2f732](https://github.com/windmill-labs/windmill/commit/fa2f732182c8ac103c3fdcb0cfe5eecba2474382))
* **frontend:** Add context for inner components of list, groups and … ([#3974](https://github.com/windmill-labs/windmill/issues/3974)) ([73175ac](https://github.com/windmill-labs/windmill/commit/73175acc018651eb0d3c161c6176e96576178813))

### Bug Fixes

* **frontend:** add tooltip to schedule pausing ([#3984](https://github.com/windmill-labs/windmill/issues/3984)) ([08476a3](https://github.com/windmill-labs/windmill/commit/08476a379456039cc8569125ca3535008d6f7bea))
* hide draft only items in pickers and from ops ([#3977](https://github.com/windmill-labs/windmill/issues/3977)) ([dd80483](https://github.com/windmill-labs/windmill/commit/dd804839f07aa56f6707fa122fa8e8bbcef81f6e))
* parallel branchall disordered skip failure retrieval ([#3975](https://github.com/windmill-labs/windmill/issues/3975)) ([e9b310e](https://github.com/windmill-labs/windmill/commit/e9b310eee0d73f3518c85b572e05473c449959a0))
* remove useless metrics ([#3962](https://github.com/windmill-labs/windmill/issues/3962)) ([f29c46a](https://github.com/windmill-labs/windmill/commit/f29c46a282e7cb1a9f093a3089e4eefd24a7bc1e))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit d55d73447d1c8ca8a282f212511cae4b70de7234  | 
|--------|--------|

### Summary:
Release version 1.355.0 introduces new features and bug fixes, with an update to `version.txt`.

**Key points**:
- Release version 1.355.0
- **Features**:
  - Add `wmill` CLI image (`fa2f732`)
  - **frontend**: Add context for inner components of list, groups, etc. (`73175ac`)
- **Bug Fixes**:
  - **frontend**: Add tooltip to schedule pausing (`08476a3`)
  - Hide draft-only items in pickers and from ops (`dd80483`)
  - Fix parallel branchall disordered skip failure retrieval (`e9b310e`)
[...]

PR #10823, opened 2026-08-24
fix: patch sqlx so a cancelled BEGIN cannot poison a pooled connection

rubenfiszel:
## Summary

A production incident had **every flow on one worker pod failing for ~20 minutes** with `current transaction is aborted, commands ignored until end of transaction block` (`25P02`), surfacing at ~10 unrelated call sites at once — job pull, flow status, job logs, variables, scripts, the workspace error handler. The pod never restarted; it healed on its own once `max_lifetime` recycled the pool.

**Root cause is upstream.** `Pool::begin` is not cancel-safe on Postgres:

1. `PgTransactionManager::begin` raises the `transaction_depth` its rollback-on-drop guard keys on only *after* the `BEGIN` round trip. A future cancelled in between — a disconnecting API client, a `tokio::time::timeout`, an aborted task — leaves the session inside a transaction with nothing queued to end it. (Normally that window is sub-millisecond; a degraded database widens it, and the pod logged `db_latency=8315ms`.)
2. `Floating::return_to_pool` calls `ping()`, which on Postgres is a bare `wait_until_ready`: it drains the `ReadyForQuery` but never inspects its transaction-status byte. So the connection is judged healthy and returned to the pool.
3. The next borrower's statements run inside that transaction and hold its locks indefinitely → lock waits on `worker_ping` → a 4-way deadlock (`40P01`).
4. That error turns the session into `idle in transaction (aborted)`, and every unrelated query on the connection fails with `25P02` from then on.
5. Nothing recovers it: the depth is still zero so no rollback is ever queued, and `idle_in_transaction_session_timeout` never fires on a connection in constant use. Only `max_lifetime` (30 min) does — which matches the ~20 minutes observed.

Upstream has known since 2022 ([launchbadge/sqlx#2054](https://github.com/launchbadge/sqlx/issues/2054), labeled `bug`), fixed it for **SQLite only** and closed it. #2819 later added the Postgres `Rollback` guard, but it is a no-op at depth 0 — the pool case. #3980 (shipped in 0.9.0, titled *"Correctly `ROLLBACK` transaction when dropped during `BEGIN`"*) restructures the generic `Transaction::begin`, but for Postgres it still funnels into the same depth-gated `start_rollback`. **I tested 0.9.0 directly: it still leaks**, so bumping does not fix this.

## Changes

- `backend/Cargo.toml` — point the `sqlx` family at [windmill-labs/sqlx@6bdaee94](https://github.com/windmill-labs/sqlx/commit/6bdaee94fa62a01561125646da3f99eb341f2457), which is `v0.8.6` plus one commit: claim the transaction depth *before* the `BEGIN` round trip and unwind it if the `BEGIN` did not take, so the drop-guard has something to act on.

  The whole family moves together deliberately. `sqlx-postgres` depends on `sqlx-core` by path inside the sqlx workspace, so patching it alone yields two incompatible `sqlx-core`s and `Postgres` stops implementing the `Database` the macros expect. The lock still resolves to exactly one `sqlx-core`.
- `backend/windmill-common/tests/sqlx_begin_cancel_safe.rs` — guards the patch. Dropping it (an sqlx bump that forgets to rebase, a deleted `Cargo.toml` line) still compiles, so a test is what notices.

This is the same mechanism already used for `tokio-postgres`/`postgres-types`/`postgres-protocol`, which are pinned to the MaterializeInc fork for a change upstream declined to merge.

## Test plan

- [x] `cargo test -p windmill-common` — 531 unit + all integration tests pass
- [x] New test is load-bearing: with the patch entries removed it fails with `left: Some("idle in transaction")`, `right: Some("idle")`
- [x] `cargo tree` confirms `sqlx-postgres` resolves to the fork, and the lock has exactly one `sqlx-core` (unchanged count from main)
- [x] Patch verified not to alter transaction semantics: commit persists, drop rolls back, explicit rollback works, nested savepoint rolls back inside a committing outer transaction
- [x] Exercised on a running backend: script deploy, group create, script run (`27`), multi-step flow with a nested for-loop (`[2,4,6]`)

## Follow-ups

- Upstream report filed with the reproduction; drop this patch the moment upstream carries the fix — it is one line in `Cargo.toml` plus the test.
- Bumping to sqlx 0.9 is worth doing on its own merits (`sqlx.toml`, advisory-lock cancel-safety) but is deliberately **not** coupled here: it does not fix this bug, and it needs 114 `AssertSqlSafe` call sites plus a Postgres connection-option escaping change that wants its own testing.

---
## Summary by cubic
Patches the `sqlx` family to a fork so a cancelled `Pool::begin` on Postgres no longer leaves the session in an open transaction and poisons a pooled connection. Previously, cancelling `BEGIN` could return a connection still in a transaction; later queries deadlocked or failed with `25P02` until the pool recycled it.

- backend/Cargo.toml: patches all `sqlx` crates (`sqlx`, `sqlx-core`, `sqlx-postgres`, `sqlx-macros`, `sqlx-macros-core`, `sqlx-mysql`, `sqlx-sqlite`) to the `windmill-labs/sqlx` fork that claims transaction depth before the `BEGIN` round trip; `Cargo.lock` reflects git sources and keeps a single `sqlx-core`.
- Adds `backend/windmill-common/tests/sqlx_begin_cancel_safe.rs` to guard the patch; the test is ignored by default. When changing `sqlx`, run: cargo test -p windmill-common --test sqlx_begin_cancel_safe -- --ignored. Updates `docs/validation.md` to point to this.
- No schema or app code changes. Remove the patch once upstream ships a fix; the test will catch regressions.

<sup>Written for commit 7cb7941d0f13743c7eac3d192818449fa9f9b74c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10823?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes WIN-2444

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -11849,8 +11849,7 @@ dependencies = [
 [[package]]
 name = "sqlx"
 version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fefb893899429669dcdd979aff487bd78f4064e5e7907e4269081e0ef7d97dc"
+source = "git+https://github.com/windmill-labs/sqlx?rev=6bdaee94fa62a01561125646da3f99eb341f2457#6bdaee94fa62a01561125646da3f99eb341f2457"
 dependencies = [
  "sqlx-core",
  "sqlx-macros",
@@ -11862,8 +11861,7 @@ dependencies = [
 [[package]]
 name = "sqlx-core"
 version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee6798b1838b6a0f69c007c133b8df5866302197e404e8b6ee8ed3e3a5e68dc6"
+source = "git+https://github.com/windmill-labs/sqlx?rev=6bdaee94fa62a01561125646da3f99eb341f2457#6bdaee94fa62a01561125646da3f99eb341f2457"
 dependencies = [
  "base64 0.22.1",
  "bigdecimal",
@@ -11901,8 +11899,7 @@ dependencies = [
 [[package]]
 name = "sqlx-macros"
 version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2d452988ccaacfbf5e0bdbc348fb91d7c8af5bee192173ac3636b5fb6e6715d"
+source = "git+https://github.com/windmill-labs/sqlx?rev=6bdaee94fa62a01561125646da3f99eb341f2457#6bdaee94fa62a01561125646da3f99eb341f2457"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -11914,8 +11911,7 @@ dependencies = [
 [[package]]
 name = "sqlx-macros-core"
 version = "0.8.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19a9c1841124ac5a61741f96e1d9e2ec77424bf323962dd894bdb93f37d5219b"
+source = "git+https://github.com/windmill-labs/sqlx?rev=6bdaee94fa62a01561125646da3f99eb341f2457#6bdaee94fa62a01561125646da3f99eb341f2457"
 dependencies = [
  "dotenvy",
  "either",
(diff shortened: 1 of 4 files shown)


rubenfiszel:
✅ Review round clean @ 3507a05cfeb5f1237eba9df5b928ac58e2306204 — nit-only verdicts at 761c5e0f3ecc2723dd8ce5a42863824c8f451c5d (pi: good to merge; codex and claude: mergeable with nits, no P0/P1); nits addressed in 3507a05cfeb5f1237eba9df5b928ac58e2306204 — dropped the migration run for a test that touches no schema, replaced the fixed 3.5s sleep with polling until the session stops being `active`, loosened the over-specified `state == "idle"` assertion to the absence of `idle in transaction`, and de-duplicated the rationale so it lives only in `Cargo.toml`. Re-verified the guard still fails without the patch entries.

PR #10825, opened 2026-08-25
feat: run and test scripts from the AI chat through an argument form

AlexRV12:
Fixes WIN-2360

## Summary

Adds a `run_script` tool to the AI chat. The model proposes a deployed script and its arguments,
the chat renders the real argument form inline, and nothing runs until the user presses Run. The
same card then holds the run: logs while it works, the result when it lands, and the error or the
cancellation if that is how it ends. `test_run_script`, which previews the draft the model is
writing, meets the same card.

The form is the same `SchemaForm` the script page uses, so the deployed schema drives it:
required fields, enums, defaults, pickers, validation. That is also what makes it the
confirmation: neither run tool carries `requiresConfirmation`, because a yes/no card asking
whether to run says less than the form that shows what will run.

**It is now the only route to a deployed run.** `runScriptByPath` is excluded from the API
catalog the way the authoring endpoints already are, and `call_api_endpoint` refuses it by name,
pointing at `run_script` — left reachable it was the way around the form, confirming an opaque
request body instead of the arguments. That closes the last gap, but it also means the tool can
no longer be session-only: the side-panel chat used to be denied `run_script` on the grounds of
scope, and denying it now would leave that chat with no way to run a deployed script at all. Both
surfaces render the form, so both get the tool. The prompt keeps main's gate on it, unchanged in
strength — deployed runs happen only when the user explicitly asks for one — and now also tells
the model to read the deployed schema first, because a draft's inputs are not the deployed ones.

**What is in the form is what runs.** A proposed argument is not consent, but it is also not
noise: the worker takes the arguments its own signature names, and the stored schema is only a
description of that signature — stale, absent, or silent about `**kwargs`. So nothing is dropped
for being undeclared. Instead every proposed value is brought to the shape its declared type
gives it, before the form opens:

- A value the field's own widget would read is converted to what the widget shows. `"12"` in a
  `number` becomes `12`, `"false"` in a `boolean` becomes `false`, `3` in a `string` becomes
  `"3"`. This is the case that mattered: `ArgInput` renders its own reading of a wrong-typed
  value and never writes that reading back, so an untouched form submitted `"12"` after
  displaying a filled-looking `12`, and `"false"` after displaying a toggle that was on.
- A value with no reading at all is emptied, and named on the card and to the model. `"abc"` in a
  `number` has nothing to show, so the field opens empty; an empty field is otherwise the caller
  having sent nothing, which is why it is said out loud. The model is told to re-read the schema
  and match the declared type.
- A variable or resource reference is left alone in every slot but one. `$var:`/`$res:`/
  `$jsonvar:` are resolved by the job, so the declared type describes what the job receives and
  never the string standing in for it — `Number('$var:…')` is `NaN`, and coercing would destroy
  it.
- A non-array in a `dynmultiselect-` slot is emptied, and that is the one slot where a reference
  is emptied too. Not for being unreadable: `MultiSelect` maps over the value as it renders, so
  anything else throws mid-render and takes the whole card down, Cancel with it. Deferred
  resolution says nothing about that — the widget has to draw before anything resolves — which is
  why the reference rule stops here and only here.
- Nested and container values are left to the widget that renders them, and disabled fields are
  reset to their schema default exactly as `RunForm` has always done; that logic moved into
  `job_args.ts` unchanged, and `RunForm` still calls it.

*A secret's reference is not a secret.* A password-typed argument holding `$var:<path>` is kept:
naming a workspace variable is how a secret is meant to reach a job, the value never leaves the
variable, and the field's own widget already treats an incoming reference as filled. Only a
literal is dropped. `$jsonvar:` is deliberately not honoured — those paths are minted from what
the user typed into the form, so a caller naming one is naming a secret it was never shown. Both
tool descriptions say so, since a capability the model is not told about is one it will not use.

*At every depth, for the readers.* The card is persisted, a result string goes back to the
model, and the model reads it again next turn. So password-typed and `contentEncoding: base64`
values are stripped from the proposal and redacted from card and result: a secret literal is a
value the user never chose, a `$var:` reference names something the card cannot show them, and
prefilled bytes are bytes the transcript then carries unbounded. Those fields open empty, and
the real file bytes reach `runScriptByPath` and nowhere else. The walk descends on the shape of
the value rather than on the keys a declaration happens to carry, so a secret sitting in a
`oneOf` branch nobody opened is reached too.

**Both run tools meet one card.** `test_run_script` used to be a plain tool row that ran the
draft and reported back in text; it now opens the same form on the draft's schema and holds its
preview run in the same card. The tools differ in three things: where the schema comes from, how
the job starts, and whether a `dynselect-` field may fetch its options — the test form withholds
the `helperScript` entirely, because `deployed` would resolve the wrong version and `inline`
would hand the unreviewed draft to the backend to execute on mount, which is the one thing this
form exists to ask about first.

Both are answerable by the bypass posture, which is the user's standing answer to confirmations
rather than an exception to them — a form parked under it is a card nobody is watching. An
auto-answered form is never mounted: the card is attached already settled, so no fields appear
for the frame before the answer lands, and the schema they would have been built from never
reaches the transcript. The posture answers for consent, not for information: when the form is
the only source for a value the run needs — a secret or file field, which is never the model's to
send, or a required field it left empty — it opens anyway, because there is no standing answer to
a question nobody has asked. "Required and empty" now descends into a nested object that declares
its own `properties` and `required`, where the form would have rendered those fields and the
shape is unambiguous; `oneOf` and free-form objects are left alone, since guessing "needs the
user" there would park the test-and-iterate loop the posture exists to serve. `''` counts as
unanswered because the mounted form says so — `ArgInput` marks a required empty scalar invalid
and disables Run — so treating it as filled would let the bypass start a run the form itself
would have refused.

A host with no form to mount refuses the run outright when the form is the only thing that could
have answered; the posture cannot answer a question it was never asked, and a run carrying an
unanswered field would otherwise start missing an argument, including the secret just stripped
out of the model's own proposal. A schema declaring no field at all is no exception to the
confirmation: the form opens with no fields, and its Run button is still the confirmation — an
empty form asks, where the tool used to just run.

**The card is a tool call row, not a panel of its own.** It collapses like every other tool
call, opens the preview with the same chip every other row uses, and carries its status as the
duration alone, ahead of the label so the chevron stays next to what it opens. The row names the
script rather than its path — the summary users see on its own page, falling back to the path for
a script that has none — with the verb receding so the subject reads first. The duration wears the
hue of the badge the jobs bar shows for the same job, a step darker than that badge's own ink:
drawn for a tinted ground, it lands brighter than the name on this transparent one. Below the row a fixed-height body sits under a tab strip —
Inputs, Logs, and whichever of Result / Error / Cancelled / Not run the run ended on — showing
the run page's own `JobArgs` and `DisplayResult` rather than a second renderer invented for the
chat. A `JSON` toggle in the strip swaps that body for the raw tool call at the same height, so
switching does not resize the card under the cursor.

**A run moves the card three times, and each move is legible.** A tab appears as the run
produces it and the selection lands on it; landing both in one frame is what made the card hard
to follow. The tab now widens in first and the selection follows once it has landed, travelling
as one bar that slides between tabs rather than a border each tab turns on, with the body fading
over the same 200ms so the two halves of a change arrive together. The third tab cannot be
reserved ahead of time — its name is only known once the run ends (Result, Error, Cancelled, Not
run) — so it arrives carrying its final label instead. Everything collapses to the current
instant behaviour under `prefers-reduced-motion`, and a card that mounts already settled, which
is every card in a restored transcript, renders at rest.

**Result streaming reaches the card.** `getJob` returns logs but never the partial result, so a
streaming script showed nothing until it landed. Both the inline wait and the background poller
now read through `getJobUpdates`, the only endpoint carrying `new_result_stream`, and the card
switches to the Result tab at the first chunk and renders the partial with the same
`DisplayResult` stream view the run page uses. Whether a script streams is not knowable in
advance — the worker decides it by testing what `main` returned — so the first chunk is the
earliest signal there is. Logs became incremental in the same change: offset-based rather than
refetched whole on every tick.

**Plan mode holds the whole card, not just Run.** Two of the form's writes never waited for
Run: a password field mints its ephemeral variable on the first keystroke, and a `dynselect-`
argument runs its entrypoint as soon as it has a `helperScript`. So plan mode disables every
input, withholds the `helperScript`, and disables Run while leaving Cancel live. It is read
live rather than at mount, because the picker moves while a form sits pending. Refused at
submit the form stays open: nothing was blocked for good, and leaving plan mode is the way on.

**A stored transcript never asks for input nothing can answer.** Some saves fire mid-turn and
store a transcript nothing will resume. They wrote `displayMessages` verbatim, so a card that
was loading came back still asking, with no callback behind it. Every writer now stores the
same settled snapshot and carries the background jobs along with it, so a card left loading has
something to poll, and the restore settles every loading card the poller will not reach. Writer
and reader ask one shared predicate so they cannot drift.

**What this form deliberately does not fix.** Below the top level it behaves exactly like every
other script form. Descending further means resolving which `oneOf` branch is open, merging
declarations two branches make for one key, and telling a declared structure from a free-form
object — each a place to be wrong in the one direction that costs the user something, by
rewriting an argument the user typed between pressing Run and the job. Five gaps are left to
their own changes, all product-wide rather than specific to this form:

- `ArgInput` still renders a wrong-typed scalar with no error and Run enabled, everywhere else in
  the product: the schedule editor, the flow editor, `/run/<path>`. This form coerces ahead of
  it, so the gap is closed here and only here. Fixing it at the source means teaching `ArgInput`
  to write its own reading back — with the same reference guard, since a `$var:` in a number slot
  renders blank and must not be rewritten — and that is a change with its own blast radius.
- **`MultiSelect` throws on any non-array value, and nothing but this form guards it.** The
  argument coercion above keeps such a value out of the chat's card, but that is a guard at one
  caller, not a fix. `DynamicInput` accepts `any` and normalizes only `undefined`
  (`DynamicInput.svelte:30-40,57-61`), and `MultiSelect` treats every non-null value as valid
  before calling `.map` on it (`MultiSelect.svelte:75,91`). This reproduces on `main` with no AI
  involved: the script page seeds its form from the URL hash with `JSON.parse`
  (`scripts/get/[...hash]/+page.svelte:353-364`), so opening
  `/scripts/get/<hash>#tenants="$var:u/admin/x"` on a script with a `dynmultiselect-` argument
  throws in render and leaves the page with no form, no summary and no Run button. Verified on
  this branch, where `DynamicInput` and `MultiSelect` are byte-identical to `main` in everything
  but a `disabled` prop. The fix belongs in `DynamicInput`, which knows the format is multiple —
  not in a `MultiSelect` fallback of `Array.isArray(value) ? value : []`, which paints an empty
  widget while leaving the bad value bound and submittable.
- **A malformed `format: "ai-provider"` object throws in render the same way**, and this form does
  not guard it either: `coerceArgsToSchema` keeps the object because it is not a scalar, and
  `AIProviderPicker` checks only that the outer value is an object
  (`AIProviderPicker.svelte:22-27`). An object-valued `model` reaches `model.endsWith(...)`
  (`reasoningRegistry.ts:30-33`) and an object-valued `resource` reaches `value?.startsWith(...)`
  (`ResourcePicker.svelte:189`), neither of which is a type guard. Pre-existing on `main` for any
  form over a schema carrying that format; the known producer is the AI-agent flow schema
  (`flowInfers.ts:8-14`).
- A `DynSelect`/`DynMultiselect` argument runs a helper job of the script when its field mounts,
  before Run and undone by no Cancel. Opening `/run/<path>` has always done this; the fix is
  deferring the first fetch inside `DynamicInput`, which fixes every caller at once, rather than
  gating this one card. Test runs already withhold the helper entirely, for a different reason.
- `processSecretArgs` walks only the top level, so an *object*-typed password nested under
  another argument is queued as its own value. String passwords are unaffected at any depth.
  `RunForm`, `ScriptEditor`, `FlowPreviewContent` and `DbtEditor` all call it the same way, so
  every run form in the product sends that literal. This branch does widen the gap on the read
  side: `stripSecretArgs` and `redactSecretArgs` are now recursive, so such a value is hidden
  from the card and the model while still reaching the job. Closing it means changing a helper
  six surfaces depend on, and one that mints ephemeral variables — its own PR.

## Changes

- `run_script` in the global toolset, rendering `RunArgsFormDisplay` inline in both the sessions
  chat and the side panel; the tool description tells the model to open the form even for
  arguments it cannot supply (a required file is otherwise a request it would explain in chat and
  never act on).
- `runScriptByPath` joins `COVERED_ENDPOINTS`, so `search_api_endpoints` hides it and
  `call_api_endpoint` refuses it pointing at `run_script`. `runFlowByPath` stays: a flow has no
  form tool yet.
- The system prompt's run bullet leads with the default (`test_run_script` … `for any run
  request, an existing script included`) and gates `run_script` on the user naming the deployed
  version, with the deployed-schema read ahead of the call for both it and `runFlowByPath`.
  `run_script`'s own description says the same thing: before this it read *use this when the user
  asks to run or execute something*, contradicting the prompt, and the description was winning.
- `job_args.ts` holds the shared `enforceDisabledDefaults` and its toast, `coerceArgsToSchema`
  (convert, or empty and report), and the recursive secret/file stripping and redaction.
  `RunForm` calls the same helpers.
- `RunScriptCard` renders the whole call as one `ChatCollapsibleCard` row: the duration in
  `headerLeft`, coloured with the ink of the status badge the jobs bar shows for the same job,
  the standard `ToolPreviewCard` chip in `headerRight`, and a fixed-height tabbed body holding
  the form, then `JobArgs`, the logs, and `DisplayResult`. Logs open on their last line and
  follow the tail while the job writes. Cancel sits in a bottom-right action row, where the form
  keeps Run.
- `runThroughForm` in the global toolset is the one path both run tools take: coerce the
  proposal, strip what is not the model's to send, open the form, re-gate plan mode, run, and
  report. `run_script` and `test_run_script` are thin callers over it, differing in the schema,
  the job starter, whether the form may load dynamic options, and whether YOLO may answer.
  `loadScriptForEdit` returns the draft's schema alongside its content so the test path has one
  without a second fetch.
- `createJobUpdateReader` in `shared.ts` reads `getJobUpdates` incrementally, and both
  `pollJobCompletion` and `AIChatManager`'s detached poller drive one, so a run that outlives
  its turn keeps streaming into its card. `ToolDisplayMessage.resultStream` carries the partial
  and the terminal patch clears it.
- `AIChatManager`: pending run-form callbacks, a manager-owned form draft shared by the card and
  the preview pane, mid-turn saves storing settled cards together with their background jobs,
  restored cards settled when nothing can resolve them, and Stop honoured while the deployed
  schema is still being read. A run whose job already exists is recorded as started, not
  cancelled.
- A native `runform:` preview slot in the sessions panel (`previewRouter`, `sessionPreviewTabs`,
  `PreviewTabHost`, `RunFormPreviewSlot`) mounting the chat's own form on the same tool call, so
  Run in the panel is Run in the chat and the tab follows the call to `/run/<jobId>` once the job
  exists. It reports on the existing `ai_session` / `tab` `feature_usage` pair, with `run_form`
  as its key.
  A submitted form whose job never starts — a path the user cannot execute, a script deleted
  since the model read it — leaves that tab open on "This run form is no longer available",
  since only a job retargets it and only Cancel closes it. Deliberate: the failure is already
  reported on the card, and closing a tab under someone is the more surprising move, so the tab
  is theirs to dismiss. The wording stays general on purpose — a throw from `startJob` does not
  say whether the server queued the job, so anything more specific would be a guess.
- `DynamicInput` accepts `disabled` and `ArgInput` passes it, so a disabled dynamic input is
  now disabled wherever `SchemaForm disabled` already flowed: the run page and the flow editor
  as well as this card.
- `Tabs` gains an opt-in `slidingIndicator` (with `indicatorClass`, defaulting to the neutral
  border colour this card takes as is), drawing the selection as one bar that measures the
  selected `Tab` and slides between them. Off everywhere else, so the ~100 existing strips keep
  the border they have. `Tab` gained one attribute,
  `data-tab-selected`, for the bar to measure against: a `Tab` decides its own selection, and
  the mark moving is the only signal that survives prefix and `otherValues` matching.
- Scroll fades on the form, the card body and the raw tool blocks: bottom only, drawn over the
  scroller and only while there is something below, sized `min(2.5rem, 35%)` so a short block
  dims a line or two rather than half of itself.
- The card's own UI text is set in `font-main`: it lives inside a tool row, which is
  `font-mono` throughout, and a field label or a tab name is not code. The pane copy of the same
  form always read that way, so this is what makes the two hosts match.
- Escape is inert inside the form except on the action row, so a stray press cannot discard
  typed arguments.
- `ChatCollapsibleCard` gains `labelPrefix`, set before the label and left unemphasised so a
  caller's `labelClass` lifts the subject alone. The run card passes the verb through it and the
  runnable's name as the label.
- The card's own per-call state — the raw-JSON view and the tab the user picked — is keyed by
  `tool_call_id`, like the collapse state beside it: the transcript is keyed by index, so a bare
  flag hands one card's view to whatever run reuses the slot.
- The run form's in-flight flag moved to `AIChatManager`, keyed by call. It lived on the
  component, which unmounts when the form moves between the card and the preview panel — coming
  back with Run and Cancel re-armed while `processSecretArgs` was still in flight.
- Arguments persisted with the card are bounded: past the size `JobArgs` would refuse to render
  anyway, the transcript stores the `WINDMILL_TOO_BIG` marker the runs page already speaks and the
  card reads them off the job. The job still receives them in full. Substituted only once there is
  a job id to read from, so a run that never starts keeps the proposal.
- Both run tools are marked `bypassedByAutoAccept`, so the YOLO tooltip lists them among what the
  posture skips. Dropping `requiresConfirmation` had quietly removed them from that list while the
  posture still answered their forms.
- `formNeedsUser` decides whether the posture has anything to answer with: a stripped secret or
  file, or a required field — at the top level or inside a declared nested object — with neither a
  proposed value nor a default, keeps the form open under the bypass. It is what keeps
  `run_script`'s own description honest, since that description tells the model the form is where
  the user supplies what it cannot.
- The no-form-host refusal is asked of the posture rather than of the tool, and only when the form
  is the only thing that could answer. Keyed on the tool it would pass every run tool through, so
  a chat with nowhere to mount a form would run a deployed script on the model's arguments alone,
  in any posture.
- Two new eval cases. `global-test33-run-deployed-script-with-form` pins that the model reads
  the deployed schema before running and prefills the form from it.
  `global-test34-run-with-secret-from-variable`: a deployed script with a
  secret argument and a workspace variable holding the value, asserting the call references the
  variable rather than inventing a literal.
- Unit tests for the coercion, the stripping traversal, the settle-on-restore predicate, the
  preview-tab routing, the arguments a test run previews, the YOLO asymmetry between the two
  forms, the confirmation a schema-less test run now waits for, the refusal of `runScriptByPath`
  through the catalog, and the bound on persisted arguments. `job_args.test.ts` owns what each
  step of the argument pipeline does; one integration case crosses all of them through
  `run_script`, since deleting a call from `runThroughForm` leaves every unit test passing.

## Test plan

- [x] Ask a session to run a deployed script: the model reads the deployed schema first, the card
      opens on it prefilled where the model could infer, and no job exists until Run.
- [x] A required base64 file: the field opens empty, the job receives the real bytes, and
      neither the card nor IndexedDB holds any base64.
- [x] A password field, plain and nested and in an array item: the job's args carry
      `$var:`/`$jsonvar:`, the card records `<hidden>`, and the transcript holds neither the
      typed secret nor the variable path.
- [x] Ask a session to run a script whose secret argument is held in a workspace variable: the
      call references the variable, the form opens with it filled, and a literal the model
      proposes instead is still dropped.
- [x] Make the model send `count` as `"7"` on a `number` field: the form shows `7` and the job
      receives the number `7` — not the string it was sent, and not the nothing the previous
      filter would have dropped it to.
- [x] Make the model send `count` as `"a lot"`: the field opens empty and marked required, the
      card names it on its own line, and the model reads back a shape problem rather than
      concluding the script has no such input.
- [x] Make the model send a `$var:` reference for a `dynmultiselect-` argument of a deployed
      script: the form mounts, the multiselect opens empty and is named on the card, Run and
      Cancel both work, and the job receives the empty array the field displayed. Before the
      guard this threw in `MultiSelect` mid-render and left the card frozen on its pre-form view
      with no Run and no Cancel, recoverable only by reloading the page.
- [x] Ask the model to send an argument the schema does not declare: it reaches the form and the
      job, because the worker's signature is the authority and the stored schema may be stale,
      absent, or silent about `**kwargs`.
- [x] Run a streaming script: the card switches to Result at the first chunk and fills in as the
      answer arrives, including after the wait detaches and the background poller takes over.
- [x] Run a script that logs for a while: the Logs tab opens on its last line and follows the
      tail, and Cancel stops the job and settles the card as cancelled rather than failed.
- [x] Switch tabs and toggle JSON on a settled card: the card keeps one height throughout, and
      the raw view is the same tool call every other row shows.
- [x] Move a pending form to the preview panel and back: the typed values follow in both
      directions, only one form is mounted at a time, and pressing Run in the panel is the chat
      pressing Run.
- [x] Switch to plan mode with a form open: every input goes read-only, dynamic ones included,
      Run disables and Cancel stays live; switching back restores what was typed.
- [x] Reload mid-turn with a form pending: the card comes back settled rather than still
      asking. Reload with a run's job in flight: the card resolves from the polled job, and a
      settled card comes back with its logs and result intact.
- [x] Open a script with a dozen-odd arguments: the fields scroll under a height cap and Run
      stays visible.
- [x] `runScriptByPath` is gone from `search_api_endpoints` and refused by name through
      `call_api_endpoint` with a pointer at `run_script` (pinned by unit test).
- [x] A chat with no form to mount refuses a deployed run whose form is the only thing that could
      answer for it, and runs one the posture can answer on its own (pinned by unit test).
- [x] Under YOLO, a deployed run whose schema declares a secret, or requires a field the model
      left empty — at the top level or inside a declared nested object — still opens its form and
      waits; one with everything supplied runs with no form mounted at any point.
- [x] Ask a session to test a draft: the same card opens on the draft's schema, the row reads
      Test / Testing / Tested rather than Run, and a `dynselect-` field falls back to a plain
      editor instead of running the deployed script's helper. Under YOLO the same request runs
      with no form mounted at any point, and so does a deployed run under the same posture.
- [x] Watch a run advance: the Logs tab widens in and the bar follows it, then the outcome tab
      does the same carrying its own name, and clicking between tabs afterwards moves the bar to
      each one. The bar keeps the selected tab's width as its label changes — a growing log count,
      Result becoming Error.
- [x] Ask a session to test a script that takes no arguments while the model proposes one: the
      card opens an empty form and no job exists until Run. Under YOLO the same call runs, and the
      autonomy tooltip names `test_run_script` among what it bypasses.
- [x] Toggle a settled card to raw JSON, collapse it and open it again: it renders rather than
      throwing. Load a different chat: the card in that slot is not left in raw view.
- [x] Read the row: it names the script, not its path, with the path on the preview chip.

Driven end to end on a local instance with a real model, against deployed fixtures covering no
arguments, required arguments, a locked argument, a free-form `oneOf` branch, a base64 file, a
real `DynMultiselect`, a script that streams its result, one that logs on a timer and one that
always throws, with the job's arguments asserted against `v2_job.args` rather than the UI. Two
guards are pinned by unit test instead: stripping a model-proposed file, since no model emits
base64 on request, and a secret under `array.items.oneOf`, a shape no parser emits. Each new
guard was proven able to fail before being trusted — disabling the reference check fails exactly
the reference tests, reverting the dyn-multiselect ordering fails exactly the reference half of
its crossing test while the object half still passes, and breaking each of the three pipeline
calls in `runThroughForm` in turn — `coerceArgsToSchema`, `stripSecretArgs`, `stripFileArgs` —
fails the integration case that crosses them.

`npx vitest run src/lib/components/copilot/chat src/lib/components/sessions
src/lib/components/job_args.test.ts` gives 1,329 passing; the 94 failures sit in three files
(`AIChatManager.test.ts`, `ContextManager.test.ts`, the app toolset) that fail the same way on
`main`. `npm run check` reports 18 errors across 10,326 files, all pre-existing on `main` in
files this branch does not change — `src/lib/gen/` is stale there relative to
`SuperadminSettingsInner.svelte`, `AppEditorHeaderDeploy.svelte`, `GuestActivityList.svelte`,
`workspace_settings`, `a/[...path]` and `public/[workspace]`, and is fixed by
`npm run generate-backend-client` on main.

**Eval A/B.** `global` mode on `sonnet`, eleven shared cases, branch head against the
merge-base rather than current `main` so unrelated commits stay out of the delta. Pass rate is
8/11 on both sides. The fixed cost of the new tool surface reads cleanly off the five cases that
took an identical tool trajectory on both sides — `test10` +498, `test32` +498, `test6` +498,
`test30` +495, `test3` +458 — so **roughly +486 tokens of window occupancy per iteration**, about
1.9% of a 26k window. The other six cases swing between -11,781 and +10,072 because the model took
different paths; that is trajectory noise, not tax.

Both new cases pass. Two failures (`global-test9`, `global-planmode2`) fail identically on both
sides, and `global-test31` flipped in both directions across runs on the same `run succeeded`
check with no relevant code between them, so it is environmental here rather than a signal.

One genuine divergence: `global-run-arg-resource-reference` passed before and fails now. The model
reads the deployed schema and passes the resource reference exactly as the case asserts
(`$res:f/evals/global/github_main`), then calls `run_script` where the case requires
`test_run_script`. Four rounds across three wordings — including one naming the counter-example
outright, *a bare "run X" is not that* — produced a byte-identical trajectory every time. The
prompt and `run_script`'s description now agree with each other, which they did not before; the
case predates `run_script` existing, and whether a bare "run this" should mean the draft or the
deployed version is still open.

## Screenshots

Every state the card can be in, in dark except where noted, taken on the current head. The
scripts are `f/qa/wide_form` (fourteen arguments), `f/qa/log_stream`, which prints a line per
tick, `f/qa/stream_answer`, which streams its answer a word at a time, `f/qa/fails`, which throws
on demand, and `f/qa/number_only`, whose one argument is a number.

**Waiting on you.** The deployed schema's own widgets, prefilled where the model could infer.
Only the fields scroll, under a height cap, so Run never leaves the fold, and what is still
below fades rather than being cut.

the run form in a card

**A value the field has no reading of.** The model was told to send `count` as `"a lot"` on a
`number` field. There is nothing to show, so it opens empty — and the card says so, because an
empty field is otherwise the caller having sent nothing. Required and empty, so Run is off until
the user fills it in. A `"7"` in the same slot would have been converted to `7` silently, with
nothing to report.

an argument emptied for having no reading

**Moved to the preview panel.** The panel mounts the chat's own form on the same tool call, so
Run there is Run in the chat, and both render one draft: what was typed in the card is already
in the panel, and closing the tab brings it back with the panel's edits. The row says where its
inputs went rather than showing a second copy of them.

the same form open in the preview panel

**The run in the preview panel**, which the tab holding the form follows into once Run is
pressed there, keeping its place in the strip. The chat card fills in beside it.

a run open in the preview panel

**Succeeded**, opening on the outcome in the run page's own renderer.

a succeeded run showing its result

**What it ran with**, as the runs page tabulates it — the password argument recorded as
`<hidden>`, because the card is persisted and read back by the model.

the inputs tab of a settled card

**As raw JSON**, the shape every other tool card in the chat is read in, at the same height the
tabs had.

the card as raw JSON

**Running.** The elapsed time leads the row; logs land on the softer surface program output is
shown on everywhere else, opened on their last line and following the tail. Cancel sits where Run
did, and the jobs tray follows the same job.

a run in flight with its logs

**Cancelled while it was running.** Being stopped is an outcome, not a failure: the duration
says so in its own colour, the tab is named for it, and it carries the same result the run page
shows for a cancelled job.

a run cancelled mid-flight

**Streaming.** The answer is arriving, so the card is already on the Result tab, rendered by the
same `DisplayResult` the run page streams into. The model's turn has ended and the run has not:
the card and the jobs tray keep following it.

a result streaming into the card

**Landed**, in light, with what was streamed now the result.

the streamed run settled, light

**Failed**, with the error where the result would be.

a failed run showing its error

**A test run of a draft**, in the same card, saying so in its own tense.

a settled test run of a draft

**Plan mode entered with a form open**: every input goes read-only, Run disables, Cancel stays
live, and the card says which way out.

a run form held read-only by plan mode

**Cancelled before it started.** Nothing ran, so there are no logs and nothing to preview: the
card says so and points at the inputs it would have used.

a run cancelled before starting


AlexRV12:
Superseded by #11001 — same final state, rebased on current `main` and rebuilt as 10 scoped commits instead of 78 incremental fixes, with a shorter description. No code change between the two heads (`git diff` is empty).

AlexRV12:
Superseded by #11001 — same final tree, rebuilt as 10 scoped commits on current `main` with a shortened description. Closing this one; the branch stays for reference.

PR #10827, opened 2026-08-25
chore: move the compose stack to postgres 18

rubenfiszel:
## Summary

Moves the `db` service in the root `docker-compose.yml` from `postgres:16` to `postgres:18`.

From 18 on, the official image stores the cluster in a major-version subdirectory
(`/var/lib/postgresql/18/docker`) and declares its volume at `/var/lib/postgresql`, so the mount
point moves with it. That is a breaking change for every existing stack. The compose file carries
only the constraint a reader needs before touching the mount; the migration runbook lives in a gist
so it can be corrected in one place:

**https://gist.github.com/rubenfiszel/4f2807b0fc685865f6e97d352681b9a8**

## Performance: PG18 removes a scalability cliff

Earlier revisions of this description reported a flat regression. That was wrong, twice, and the
corrections are worth recording because both were measurement errors rather than PostgreSQL
behaviour:

1. A first pass showed PG18 1.7-2.7x *faster*. Each postgres image ships its own `pgbench`, so
   `docker exec <pg16> pgbench` vs `docker exec <pg18> pgbench` compared **clients**, not servers.
2. A second pass showed a flat ~8-23% regression. Those runs used a queue where every row was
   runnable, which is only one of the states this query runs in.

Measured properly (identical CPU-pinned containers, one `pgbench` binary over TCP, real Windmill
schema, both clusters verified to the same physical state, `VACUUM FULL` + `FREEZE` + `REINDEX`
between runs), the worker pull query with 8 tags over 200k queued rows:

| queue state | PG16 | PG18 | ratio |
| --- | --- | --- | --- |
| 40k jobs runnable now | 48,133 | 41,167 | 0.86 |
| backlog present, nothing runnable | 4,995 | 53,590 | **10.7** |

**PG18 is ~14% slower when there is work waiting, and ~10.7x faster when the queue holds a backlog
with nothing runnable** — the state an instance with scheduled jobs, delayed retries and suspended
flow steps sits in most of the time, and the state every worker polls in a loop.

### Why, and a correction

Both effects are the same mechanism, and it is **skip scan**, which I previously reported as not
applying here. It does. `queue_sort_v2` is `(priority DESC NULLS LAST, scheduled_for, tag)` and
`priority` is NULL for ~99% of rows, so the leading column is unconstrained:

- **PG16** cannot begin the scan without it and walks the whole index — 645 buffers, 1.4ms — before
  concluding nothing is runnable. That is the 4,995 tps cliff.
- **PG18** skip-scans the two distinct `priority` values (`Index Searches: 2`) and answers in 12
  buffers, 0.024ms.

The same machinery costs a constant +6 buffers per scan when a row is available immediately, which
is the 14%. Bisected on a synthetic table, `LIMIT 1` buffers:

| index / predicates | PG16 | PG18 |
| --- | --- | --- |
| plain `(a,b)`, matching ORDER BY | 1 | 1 |
| partial `(a DESC NULLS LAST, b, c)`, no qual | 4 | 4 |
| ...+ qual on a trailing column | 4-11 | 10-17 |

Ruled out individually: plan choice, protocol overhead, row locking, `SKIP LOCKED`,
`data_checksums`, `io_method` (`worker` and `sync`), prefetch (`effective_io_concurrency=0`), and
index bloat.

### The 14% is recoverable, if we want it

An `OFFSET 0` fence keeps `tag` out of the index condition while leaving the `priority` skip scan
intact. On the real schema:

| | PG16 | PG18 |
| --- | --- | --- |
| runnable work, current query | 48,133 | 41,167 |
| runnable work, fenced | 45,153 | **47,477** |
| nothing runnable, fenced | 4,970 | 51,447 |

PG18 + fence is at or above PG16 in **every** state. Verified that the fenced form still locks
correctly: two concurrent sessions pick different rows under `FOR UPDATE SKIP LOCKED`.

It costs PG16 ~6%, so it would want gating on server version rather than applying unconditionally.
`make_pull_query` builds this SQL at runtime and the backend already reads `SELECT version()` at
startup (`backend/src/main.rs:830`), so that is mechanically straightforward. **Not proposed in this
PR** — it touches the hottest query in the system and deserves its own change with its own review.

## Changes

- `db` image `postgres:16` → `postgres:18`
- volume mount `db_data:/var/lib/postgresql/data` → `db_data:/var/lib/postgresql`
- 7-line comment stating the constraint and linking the runbook

The volume name is deliberately unchanged. A `db_data` holding a 16 cluster makes the 18 image exit
with an explanatory error, so an operator who bumps without migrating gets a loud failure and
untouched data, rather than Windmill booting against an empty database and looking wiped.

## The two things that make the migration non-obvious

**Windmill's data is not confined to the `windmill` database.** `create_custom_instance_database`
(`windmill-common/src/lib.rs:1425`) and the workspace-fork endpoint run `CREATE DATABASE` on the
main pool, so instance datatables, DuckLake catalogs and `wm_fork_*` databases sit in the same
cluster. `pg_dump -Fc windmill` carries none of them and nothing errors. Verified: **0** references
to the sibling databases in that dump.

**RLS policies are granted to cluster-level roles.** `windmill_admin` / `windmill_user` are not in a
database dump, and `pg_restore` skips every policy whose grantee is missing unless
`--exit-on-error` is passed. Measured: **338 policies → 0**. Those roles are live;
`windmill-queue/src/schedule.rs` does `SET LOCAL ROLE windmill_user`.

`pg_dumpall` of the whole cluster solves both and keeps role passwords.

## What was tested

- **Schema** — all 627 migrations apply cleanly on 18; `pg_dump -s` byte-identical to 16 apart from
  the version banner and 18's named-`NOT NULL` syntax.
- **Fresh stack** — real `ghcr.io/windmill-labs/windmill:main` (CE v1.796.0) server + worker on
  `postgres:18` from this file: migrations applied, worker registered, Python job ran to `success`.
- **The gist runbook, verbatim, with sibling databases present** — 16 stack on the previous compose
  file, a script, a job, plus `wm_datatables` and `wm_fork_admins_lake_deadbeef` (500 rows each).
  After: 3/3 databases, 500/500 rows each, 2 scripts, 2 completed jobs, 365 RLS policies, 4 roles
  with passwords, and the pre-upgrade script ran to `success`.
- **`vacuumdb --all --analyze-in-stages`** — user-table `pg_statistic` 0 → 314 / 0 → 2 / 0 → 2. A
  single-database `ANALYZE` left the siblings at 0.
- **Skipping the migration** — container exits 1 with the docker-library explanation,
  `depends_on: service_healthy` keeps the server from starting, `PG_VERSION` still reads 16.
- **Performance** — as above, on both a synthetic reproducer and the real schema, across queue
  depths 0 / 100 / 5k / 200k and readiness 0% / 5% / 20% / 50% / 100%.

## Other PG18 features assessed

- **`uuidv7()`** — no gain, job ids are already ULIDs.
- **`NOT NULL NOT VALID`** — useful going forward: add `NOT NULL` to a large job table without the
  validation scan under `AccessExclusiveLock`.
- **pg_upgrade carries planner statistics** — helps 18→19, not this dump/restore.
- **checksums by default / async I/O / virtual generated columns** — no measurable effect here.
- **A tag-leading index** was tested as an alternative and is catastrophic (740 tps): the ORDER BY
  then requires a sort. The current index shape is correct.

## Not in scope

`start-dev-db.sh`, `integration_tests/`, `examples/deploy/*` (14/16) and the pinned CI service
containers are untouched. `backend-test.yml` uses an unpinned `image: postgres`, today 18.6
(identical image id to `postgres:18`), so the backend suite already runs on 18. Four example stacks
are on postgres:14, EOL 2026-11-12; worth its own PR.

## Test plan

- [x] 627 migrations apply cleanly on `postgres:18`
- [x] schema dump on 18 matches 16 modulo dump syntax
- [x] fresh stack boots and runs a Python job to success
- [x] gist recipe reproduces state exactly, incl. sibling databases, 365 RLS policies, role passwords
- [x] script created pre-upgrade runs post-upgrade
- [x] `vacuumdb --all` populates statistics in every restored database
- [x] bumping without migrating fails loudly and leaves the 16 cluster intact
- [x] perf characterised across queue states on synthetic and real schema, cause bisected
- [x] fenced variant verified to preserve `FOR UPDATE SKIP LOCKED` semantics
- [x] `docker compose config` validates; gist link returns 200

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,15 +8,46 @@ x-logging: &default-logging
     compress: "true"
 
 services:
+  ## UPGRADING AN EXISTING STACK: a db_data volume written by postgres:16 holds a
+  ## cluster that 18 cannot read. The container exits with an explanatory error
+  ## instead of coming up blank, so nothing is lost before you migrate. Postgres 16
+  ## is supported upstream until Nov 2028, so staying on `image: postgres:16` with
+  ## the old `db_data:/var/lib/postgresql/data` mount remains a valid option.
+  ##
+  ## To move the data across, dump first while the old 16 container is still the
+  ## one running (git checkout the previous docker-compose.yml if you already
+  ## replaced it), with only db up:
+  ##   docker compose stop windmill_server windmill_worker windmill_worker_native
+  ##   docker compose exec -T db pg_dumpall -U postgres --globals-only --no-role-passwords > globals.sql
+  ##   docker compose exec -T db pg_dump -U postgres -Fc windmill > windmill.dump
+  ## Check both files, then switch to this file and start 18 on an empty volume:
+  ##   docker compose down
+  ##   docker volume ls | grep db_data     # then remove the one this project owns
+  ##   docker volume rm <that_volume>
+  ##   docker compose up -d db
+  ##   docker compose exec -T db psql -U postgres -d postgres < globals.sql
+  ##   docker compose exec -T db pg_restore -U postgres -d windmill --no-owner --exit-on-error < windmill.dump
+  ##   docker compose up -d
+  ## Loading globals.sql is not optional. Windmill grants its row-level security
+  ## policies to the cluster-level windmill_admin and windmill_user roles, which a
+  ## database-only dump does not carry, and pg_restore silently drops every policy
+  ## whose grantee role is missing. The one error to expect is psql reporting
+  ## `role "postgres" already exists`, which the fresh cluster bootstraps itself.
   db:
     deploy:
       # To use an external database, set replicas to 0 and set DATABASE_URL to the external database url in the .env file
       replicas: 1
-    image: postgres:16
+    image: postgres:18
     shm_size: 1g
     restart: unless-stopped
     volumes:
-      - db_data:/var/lib/postgresql/data
+      # From 18 on the official image keeps the cluster in a major-version
+      # subdirectory (/var/lib/postgresql/18/docker), so the mount has to be the
+      # parent directory: that is what lets pg_upgrade see an old and a new
+      # cluster inside a single mount point. Mounting the pre-18 .../data path
+      # instead makes the image exit rather than start, which is what turns a
+      # stale 16 cluster into a loud failure instead of an empty instance.
+      - db_data:/var/lib/postgresql
     expose:
       - 5432
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,15 +27,19 @@ services:
   ##   docker volume ls | grep db_data     # then remove the one this project owns
   ##   docker volume rm <that_volume>
   ##   docker compose up --wait db
-  ##   docker compose exec -T db psql -U postgres -d postgres < cluster.sql
-  ##   docker compose exec -T db psql -U postgres -d windmill -c ANALYZE
+  ##   docker compose exec -T db psql -U postgres -d postgres < cluster.sql > restore.log 2>&1
+  ##   grep -i '^ERROR' restore.log      # only the two expected ones, see below
+  ##   docker compose exec -T db vacuumdb -U postgres --all --analyze-in-stages
   ##   docker compose up -d
-  ## `database "windmill" already exists` and `role "postgres" already exists` are
-  ## expected; the fresh cluster bootstraps both, and psql carries on to load into
-  ## them. ANALYZE is not optional: a restored cluster starts with no planner
-  ## statistics, which reads as "the upgrade made Windmill slow" until autovacuum
-  ## catches up. `up --wait` rather than `up -d` because psql would otherwise race
-  ## initdb on a fresh volume.
+  ## `up --wait` because psql would otherwise race initdb on a fresh volume. psql
+  ## does not stop on error and the old volume is already gone by then, so grep the
+  ## log rather than trust its exit code; `database "windmill" already exists` and
+  ## `role "postgres" already exists` are the two the fresh cluster causes itself.
+  ## vacuumdb spans every restored database, which a plain ANALYZE would not: each
+  ## comes back with no planner statistics, and a fork database is where the stall
+  ## shows. Postgres triggers reading a database in this cluster come back disabled
+  ## with "replication slot ... no longer exists", since slots are never dumped;
+  ## re-saving the trigger recreates the slot.
   db:
     deploy:
       # To use an external database, set replicas to 0 and set DATABASE_URL to the external database url in the .env file
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
   ## Windmill keeps datatable, DuckLake and wm_fork_* databases beside it and grants
   ## its RLS policies to cluster-level roles, and a single-database dump loses both
   ## silently. Full procedure, and why 16 is still a valid choice until Nov 2028:
-  ## https://gist.github.com/rubenfiszel/4f2807b0fc685865f6e97d352681b9a8
+  ## https://www.windmill.dev/docs/advanced/self_host#upgrade-postgresql-to-18
   db:
     deploy:
       # To use an external database, set replicas to 0 and set DATABASE_URL to the external database url in the .env file
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,19 +27,23 @@ services:
   ##   docker volume ls | grep db_data     # then remove the one this project owns
   ##   docker volume rm <that_volume>
   ##   docker compose up --wait db
+  ##   docker compose exec -T db psql -U postgres -d postgres -c 'DROP DATABASE windmill'
   ##   docker compose exec -T db psql -U postgres -d postgres < cluster.sql > restore.log 2>&1
-  ##   grep -i '^ERROR' restore.log      # only the two expected ones, see below
+  ##   grep -i '^ERROR' restore.log      # only `role "postgres" already exists`
   ##   docker compose exec -T db vacuumdb -U postgres --all --analyze-in-stages
   ##   docker compose up -d
-  ## `up --wait` because psql would otherwise race initdb on a fresh volume. psql
-  ## does not stop on error and the old volume is already gone by then, so grep the
-  ## log rather than trust its exit code; `database "windmill" already exists` and
-  ## `role "postgres" already exists` are the two the fresh cluster causes itself.
-  ## vacuumdb spans every restored database, which a plain ANALYZE would not: each
-  ## comes back with no planner statistics, and a fork database is where the stall
-  ## shows. Postgres triggers reading a database in this cluster come back disabled
-  ## with "replication slot ... no longer exists", since slots are never dumped;
-  ## re-saving the trigger recreates the slot.
+  ## `up --wait` because psql would otherwise race initdb on a fresh volume. The
+  ## DROP is what lets the dump's own `CREATE DATABASE windmill` run: POSTGRES_DB
+  ## has already made an empty one, and loading into it would keep the new cluster's
+  ## encoding and collation instead of the dumped ones, for that database only.
+  ## psql does not stop on error and the old volume is already gone by then, so grep
+  ## the log rather than trust its exit code; the fresh cluster bootstraps `postgres`
+  ## itself, so that role is the one error left. vacuumdb spans every restored
+  ## database, which a plain ANALYZE would not: each comes back with no planner
+  ## statistics, and a fork database is where the stall shows. Postgres triggers
+  ## reading a database in this cluster come back disabled with "replication slot
+  ## ... no longer exists", since slots are never dumped; re-saving the trigger
+  ## recreates the slot, then re-enable it.
   db:
     deploy:
       # To use an external database, set replicas to 0 and set DATABASE_URL to the external database url in the .env file
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,25 +14,28 @@ services:
   ## is supported upstream until Nov 2028, so staying on `image: postgres:16` with
   ## the old `db_data:/var/lib/postgresql/data` mount remains a valid option.
   ##
-  ## To move the data across, dump first while the old 16 container is still the
-  ## one running (git checkout the previous docker-compose.yml if you already
-  ## replaced it), with only db up:
-  ##   docker compose stop windmill_server windmill_worker windmill_worker_native
-  ##   docker compose exec -T db pg_dumpall -U postgres --globals-only --no-role-passwords > globals.sql
-  ##   docker compose exec -T db pg_dump -U postgres -Fc windmill > windmill.dump
-  ## Check both files, then switch to this file and start 18 on an empty volume:
+  ## Migrating means moving the WHOLE CLUSTER, not the windmill database. Windmill
+  ## creates instance datatable, DuckLake and wm_fork_* databases alongside it, and
+  ## grants its row-level security policies to cluster-level roles; a single-database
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Round 1 addressed in 876516510b.

**Codex P1, readiness race** — fixed. Both `up -d db` calls are now `up --wait db`, which uses the healthcheck already defined on the service. My own testing had an explicit wait loop that the written recipe didn't, so this was a real gap between what I ran and what I shipped.

**Claude P1, the recipe loses every database except `windmill`** — correct, and the more serious of the two. Confirmed in the code (`create_custom_instance_database` at `windmill-common/src/lib.rs:1425` and the fork endpoint both run `CREATE DATABASE` on the main pool) and then empirically: `pg_dump -Fc windmill` contains **0** references to sibling databases, so the following `docker volume rm` destroys them with nothing to signal it. Took the suggested fix, `pg_dumpall` for the whole cluster restored with `psql`, which also folds in the RLS/roles problem and keeps role passwords, so `--no-role-passwords` and the `custom_instance_replication_user` caveat are both gone.

Re-ran the whole upgrade with `wm_datatables` and `wm_fork_admins_lake_deadbeef` present (500 rows each) alongside the canary script and job. Everything round-trips: both sibling databases with their rows, 365 RLS policies, 4 roles with passwords intact, and the pre-upgrade script runs after the upgrade. Table in the PR body.

**P2 stop list misses `windmill_indexer`** — fixed, now `docker compose down && docker compose up --wait db`, no service names.

**P2 no `ANALYZE`** — fixed, added as an explicit step with the reason. Verified it repopulates `pg_statistic`.

**P2 runbook length** — trimmed, but not resolved, and I agree with the underlying point. It is still ~20 lines, and "a recipe in one place can be corrected once rather than after every self-hoster has curled a copy" is exactly right, as this round demonstrates. The blocker is only that the windmilldocs page does not exist yet and I did not want to ship a comment pointing at a dead link. Offered in the PR body: happy to open that PR and cut this to four lines plus a link.

**FYI on CI coverage** — one correction to the consistency note: `backend-test.yml` doesn't pin, it uses `image: postgres`, which today resolves to 18.6 with an image id identical to `postgres:18`. So the backend suite has been exercising 18 already. The pinned 14/16 workflows are real divergence, just not a total absence of coverage.

**On the ~8% pull-query regression** — agreed it should be a deliberate decision rather than a side effect, which is why it leads the PR description. Flagging for @rubenfiszel rather than deciding it here.

rubenfiszel:
Round 2 addressed in cc88fd8aac.

**Codex P1, replication slots** — taken, with a scope correction. I checked the two reviewers disagreed here, so I read the code rather than pick a side. Both are partly right: `listener.rs:186` returns `BadConfig(ERROR_REPLICATION_SLOT_NOT_EXISTS)` and `disable_with_error` does set `mode = 'disabled'`, so Codex is right that the trigger stops. But it is not silent, the `error` column carries the remediation text verbatim ("Recreate a new replication slot or select an existing one in the advanced tab..."), and `handler.rs:185-197` recreates the slot on save when it is missing, so it is recoverable without deleting the trigger. It is also narrow: only triggers reading a database *inside this cluster*, not external Postgres.

That makes it a documented expected outcome rather than a step to add, so the recipe gains two lines saying triggers come back disabled and re-saving fixes them, instead of a slot inventory procedure.

**P2, `ANALYZE` only covers `windmill`** — correct, and I had measured this wrong last round. My "803 rows in `pg_statistic`" counted catalog stats, which are there from `initdb` regardless. Counting only `public` tables:

| database | after restore | after `vacuumdb --all --analyze-in-stages` |
|---|---|---|
| windmill | 0 | 314 |
| wm_datatables | 0 | 2 |
| wm_fork_admins_lake_deadbeef | 0 | 2 |

So the siblings were getting nothing. Replaced with `vacuumdb -U postgres --all --analyze-in-stages` as suggested.

**P2, restore no longer detects errors** — correct. Now redirects to `restore.log` and greps it. Note it needs `2>&1`, not a pipe, since psql writes errors to stderr. Verified the log catches exactly the two documented lines and nothing else.

**Comment length** — this is now 32 lines and I am not going to pretend that is fine. Two consecutive rounds of corrections to a recipe embedded in a config file is the argument for moving it, as you both noted. Waiting on @rubenfiszel for the windmilldocs page; the moment that exists this block becomes four lines and a link.

Everything above was replayed on a real 16 → 18 upgrade with the sibling databases present, not just reasoned about.

rubenfiszel:
Round 3 nit addressed in ebe5fce07b. Codex and Pi were both "Good to merge"; this was the one nit.

**The `windmill` database keeps the new cluster's attributes** — correct mechanism, and I could measure the part the review environment couldn't. The two images' `initdb` defaults are **identical**, so on stock images nothing actually changes:

| | postgres:16 | postgres:18 |
|---|---|---|
| encoding | UTF8 | UTF8 |
| collate / ctype | en_US.utf8 | en_US.utf8 |
| locale provider | libc | libc |
| `datcollversion` | 2.41 | 2.41 |

Both are Debian 13.6 with the same glibc. So the consequence is conditional on someone setting `POSTGRES_INITDB_ARGS` or the base images diverging later, exactly as you said.

Took the fix anyway, because it costs one line and is strictly better: `DROP DATABASE windmill` before the restore lets the dump's own `CREATE DATABASE` run, so the main database is no longer the asymmetric one, and the expected-error list goes from two to one, which tightens the `grep`. Verified end to end on the real upgrade: 1 error line (`role "postgres" already exists`), all four databases now carrying the dumped attributes, 365 RLS policies and the sibling rows intact.

**Accuracy note on the trigger sentence** — took it. `disable_with_error` writes `mode = 'disabled'` and the update path takes `mode` from the request, so re-saving recreates the slot but leaves the trigger off. Changed to "re-saving the trigger recreates the slot, then re-enable it" so the sentence doesn't stop half a step short.

Leaving this in draft rather than flipping to ready. Not because of the review, which is clean, but because two decisions above it are @rubenfiszel's: whether to take the ~8% worker-pull-query regression now when 16 is supported to Nov 2028, and whether to move this runbook to windmilldocs so the block collapses to four lines and a link.

rubenfiszel:
Swept the rest of PG18 against this codebase rather than the release notes. Two additions and two warnings.

**LISTEN/NOTIFY is ~19% faster.** Windmill uses it for cross-replica cache invalidation. `notify_buffers` is new and configurable in 18 (default 16 x 8kB, previously hardcoded at 8). Measured with one pgbench client against both servers:

| | PG16 | PG18 | ratio |
| --- | --- | --- | --- |
| `pg_notify`, no listeners | 174,466 | 224,331 | 1.29 |
| `pg_notify`, 4 listeners attached | 126,717 | 151,147 | **1.19** |

(The first pass on this said 1.8x. That was the pgbench-client trap again, from `docker exec`. Controlled, it is 1.19-1.30x.)

**Skip scan reaches exactly two indexes, and they are the right two.** Swept every multicolumn btree index in the schema for the shape that benefits, a low-cardinality leading column that queries leave unconstrained. Only five qualify, and three are housekeeping (`autoscaling_event`, `healthchecks`, `worker_group_job_stats`). The other two are `queue_sort_v2` and `queue_suspended`. Everything else leads with `workspace_id`, which queries always constrain, so skip scan cannot apply. That is why the win is concentrated in the queue rather than spread thin.

**Warning 1: `queue_suspended` does NOT get the win.** `make_suspended_pull_query`'s `(suspend <= 0 OR suspend_until <= now())` cannot become an index condition, so the scan walks the whole partial index on both versions. With 200k suspended rows: PG16 200,839 buffers / 24.4ms, PG18 80,860 buffers / 29.6ms. PG18 touches fewer buffers but is slightly slower in wall time, and `Index Searches: 1` confirms no skip. This is a Windmill query issue, version-independent, and on a large instance it is worth more than the PG18 delta either way. Separate issue material.

**Warning 2: do not enable `idle_replication_slot_timeout`.** New in 18, default 0 (off), and it should stay off here. Windmill's Postgres triggers own logical replication slots (`windmill-trigger-postgres/src/handler.rs`). The setting invalidates slots with no connected consumer, so a server outage or maintenance window longer than the timeout would invalidate every trigger's slot at once and disable the triggers. The default is correct; this is a note for anyone tuning the DB.

**Smaller, unmeasured:** `vacuum_max_eager_freeze_failure_rate` (0.03, new) spreads freezing across normal vacuums instead of anti-wraparound storms, which suits append-mostly tables like `v2_job_completed` and the `audit` partitions. `autovacuum_vacuum_max_threshold` (100M, new) caps the 20%-of-table trigger, but only bites above roughly 500M rows. Both are plausible operational wins that I did not benchmark, since they need a long-running workload to show.

**Ruled out:** `uuidv7()` (job ids are already ULIDs), virtual generated columns (the schema has one, STORED), `enable_self_join_elimination` (no matching query shapes). `RETURNING OLD/NEW` could tidy some of the 34 `UPDATE ... RETURNING` sites but buys nothing at runtime.

rubenfiszel:
Two changes since the last round, both from @rubenfiszel's review of where this content should live.

**The gist is gone; the runbook is in the docs.** GitHub gists are owned by user accounts and never by organisations (`GET /orgs/windmill-labs/gists` is a 404, and every gist's `owner.type` is `User`), so a gist was the wrong home for the only migration instructions a self-hosted operator gets. It also read as unofficial and hung on one person's account. The compose comment now points at the self-host docs page:

https://www.windmill.dev/docs/advanced/self_host#upgrade-postgresql-to-18

Companion PR: windmill-labs/windmilldocs#1704. It also fixes a sentence on that page that this change makes false, "updating the database image is safe too", which was true within a major version and is exactly what would walk someone into the failure.

**Ordering:** windmilldocs#1704 has to merge and deploy before this one, or the link in the compose file 404s for anyone who pulls it. Flagging that rather than leaving it implicit.

Head is now e1e3a4b561. Starting a fresh round, since the last clean one was at ebe5fce07b and two commits have landed since (the runbook extraction and this link change).

rubenfiszel:
Round 4 addressed in 5658c33b32.

**Claude's nit was the most valuable finding of the whole review, and it is now fixed.** It is the one path in this change that fails quietly, and I had not tested it. Reverting only the image tag, leaving the 18-style mount in place, puts 16 in a state where `PGDATA` is a subdirectory of the volume rather than its root. Verified end to end:

```
before:            USER DATA THAT MUST NOT VANISH
volume root:       PG_VERSION = 16
after reverting only the image tag:
  container state: running exit=0        <- boots happily
  data_directory:  /var/lib/postgresql/data
  select v from important:  ERROR: relation "important" does not exist
```

The original cluster is still at the volume root, untouched, so it is recoverable, but Windmill comes up blank and looks wiped. Every other path in this change was designed to fail loudly; this was the exception, and the trimmed comment had stopped pairing "16 is still valid" with "revert the mount too". Four lines added saying exactly that.

**Codex's P1, deploy the runbook first.** Agreed on the sequencing and it is already recorded in the PR body and in windmilldocs#1704. One correction to the risk, though, since it changes whether this blocks: the link cannot 404. The page it points at exists today, and the fragment is client-side and never reaches the server:

```
/docs/advanced/self_host                          -> HTTP 200
/docs/advanced/self_host#upgrade-postgresql-to-18 -> HTTP 200
```

So the worst case before windmilldocs#1704 deploys is a reader landing on the self-host page without auto-scrolling to a section that is not there yet, on a page that is already the right page for the topic. That is a degraded link, not a broken one. I would still merge the docs PR first and have said so in both, but I do not think it needs to gate this one on correctness grounds. Happy to be overruled.

There is no code change that would resolve it either way: the alternatives are inlining the runbook again, which round 3 correctly asked me to stop doing, or linking nothing.

PR #10828, opened 2026-08-25
fix(debugger): parse bun 1.4's UUID inspector token

rubenfiszel:
## Summary

The `test_extra` job in `publish_extra.yml` has failed on every run since the bun 1.4.0 bump (`d85050f505`) — 1.794.1, 1.795.0 and 1.796.0 all fail with the same two cases:

```
- Debugger TypeScript execution: Missing expected output. Got: ["Preparing dependencies..."]
- Debugger breakpoint support: Timeout waiting for event: stopped
```

Bun 1.4 changed the inspector URL's token from an alphanumeric string to a hyphenated UUID. The debugger scrapes that URL out of bun's stderr banner with `/ws:\/\/[\d.]+:\d+\/[a-z0-9]+/i`, which stops at the first hyphen:

```
banner:    ws://127.0.0.1:9742/848c719d-a52e-4610-8e94-99cd60f34af9
scraped:   ws://127.0.0.1:9742/848c719d          -> HTTP 404
```

The inspector answers the truncated path with a 404, so the WebSocket handshake fails with *"Expected 101 status code"* and no TypeScript debug session has been able to attach since. Both failing tests are that one cause: execution never starts, so there is no output and no `stopped` event.

Two things kept it from being diagnosable from the CI log. A close before the script was running was reported as a normal script termination, so the client saw a plausible `terminated` and never learned the connection had failed. And the debuggee was never reaped — `--inspect-wait` blocks until a debugger attaches, so each failed attach leaked a bun process.

## Changes

- Match the whole URL path, and only once whitespace proves it complete. A stderr chunk can end mid-URL, which would otherwise be read as a complete but truncated URL — the same 404, arrived at a different way. Backward-compatible with bun 1.3's token format.
- Report any close before execution starts as the connection failure it is, naming the handshake error. The socket opening is not the dividing line: the setup commands run over an open socket and none of them reject when it drops (`sendInspectorCommand` only has its own timer), so `Inspector.initialized` being answered is what separates a finished script from a lost connection.
- Reap the debuggee when the launch fails.
- Pair the `Failed to start Bun` output with the `terminated` event it explains, so a run that already reported its result cannot afterwards be told it failed to launch.
- `docker/test_windmill_extra.ts`: queue events that arrive before their waiter registers. The server sends `initialized` immediately behind the `initialize` response, so the client could drop it and then time out waiting for it — a latent flake independent of the bun bug, which I hit locally.

No new test: `docker/test_windmill_extra.ts` is already the regression guard for this, and it is the suite that was failing.

## Test plan

Verified locally by running the CI suite against the debug service with the image's exact bun (1.4.0), `REQUIRE_SIGNED_DEBUG_REQUESTS=false` as CI passes, and a real `windmill` binary so the `prepare-deps` path runs:

- [x] All 4 debugger tests pass, 16 consecutive runs (was 2/4 failing deterministically)
- [x] Reproduced the root cause first: 40/40 handshake failures under bun 1.4.0, 0/40 under 1.3.11
- [x] Backward compatible — suite green with bun 1.3.11's non-hyphenated token
- [x] Close before open (the 404 path), via a stub advertising an unreachable inspector: exactly one `terminated` carrying the real error, and the process reaped
- [x] Close mid-setup, via a stub that answers `Inspector.enable` then drops: one `terminated` with the connection error, immediately rather than on a 10s timer
- [x] Close during `Inspector.initialized`, via a stub that answers every setup command then drops: one `terminated` with the connection error. Confirmed this is a real regression detector — with the flag set one line earlier the same stub yields a silent `terminated {}`

**`test_extra` cannot run on this PR.** `publish_extra.yml` triggers only on `v*` tags and `workflow_dispatch`, so the job will first execute on the next release tag after merge. Note for anyone tempted to dispatch it on this branch to check: `publish_extra` has no branch guard and would push the branch build to `ghcr.io/windmill-labs/windmill-extra:latest`.

LSP and multiplayer tests went unexercised locally (building the image pulls `windmill-ee-slim` and installs Go/gopls/pyright, which would risk filling this box's root partition); they pass in CI today and nothing here touches them.

### DIFF
--- a/debugger/dap_websocket_server_bun.ts
+++ b/debugger/dap_websocket_server_bun.ts
@@ -1652,8 +1652,12 @@ export class DebugSession {
 			await this.startBunProcess(cwd)
 		} catch (error) {
 			this.sendEvent('output', { category: 'stderr', output: `Failed to start Bun: ${error}\n` })
-			this.terminatedSent = true
-			this.sendEvent('terminated', { error: String(error) })
+			// A socket that drops mid-handshake reports the termination from onclose and fails the
+			// launch; that first event carries the result, so it must not be overwritten here.
+			if (!this.terminatedSent) {
+				this.terminatedSent = true
+				this.sendEvent('terminated', { error: String(error) })
+			}
 			// --inspect-wait blocks until a debugger attaches, so a bun we failed to attach to
 			// waits forever unless it is reaped here.
 			await this.cleanup()
--- a/debugger/dap_websocket_server_bun.ts
+++ b/debugger/dap_websocket_server_bun.ts
@@ -1652,7 +1652,11 @@ export class DebugSession {
 			await this.startBunProcess(cwd)
 		} catch (error) {
 			this.sendEvent('output', { category: 'stderr', output: `Failed to start Bun: ${error}\n` })
+			this.terminatedSent = true
 			this.sendEvent('terminated', { error: String(error) })
+			// --inspect-wait blocks until a debugger attaches, so a bun we failed to attach to
+			// waits forever unless it is reaped here.
+			await this.cleanup()
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 50e4b9e26ea1c978aa8462255b2b800fcfba5b3d

PR #10829, opened 2026-08-25
feat(frontend): flag the fork-compare datatable schema diff as legacy

diegoimbert:
## Summary

On the fork compare page, the "Datatable schema changes" section only appears for forked data tables that have **not** opted in to migrations — it is the legacy fallback that diffs schemas directly instead of tracking them as migrations. Nothing on screen said so, so the section read like a normal part of the compare flow.

This adds a warning `Alert` at the top of that section explaining why it is showing.

## Changes

- `DatatableSchemaDiff.svelte`: render a warning `Alert` ("Legacy schema comparison") above the diff list, stating that the section shows because the affected data table(s) have not opted in to migrations, that migrations track schema changes as regular deployable items, and that the opt-in belongs in the parent workspace — with the caveat that a fork inherits the flag only at creation, so the current fork keeps using the legacy diff.
- Added an `applicableLabel` `$derived` for the singular/plural wording, driven by the existing `applicableCount`.

No behavior change: the section's visibility condition (`applicableCount > 0`) and every diff/migration action are untouched.

## Screenshots

fork-compare legacy datatable alert

## Test plan

- [ ] On a fork whose data table has not opted in to migrations, open Compare & Deploy and confirm the warning renders above the diff list
- [ ] Opt that data table in to migrations in the fork and confirm the whole section (alert included) disappears
- [ ] With two or more applicable data tables, confirm the copy reads "these data tables have"

### DIFF
--- a/frontend/src/lib/components/DatatableSchemaDiff.svelte
+++ b/frontend/src/lib/components/DatatableSchemaDiff.svelte
@@ -41,7 +41,9 @@
 	// migrations, its changes flow through the normal item diff instead, so it is
 	// excluded here and the whole section hides once none remain.
 	let applicableCount = $state(0)
-	let applicableLabel = $derived(applicableCount === 1 ? 'this data table' : 'these data tables')
+	let applicableLabel = $derived(
+		applicableCount === 1 ? 'this data table has' : 'these data tables have'
+	)
 	let expandedDatatables: Set<string> = $state(new Set())
 
 	// Drawer state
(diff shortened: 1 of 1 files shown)


diegoimbert:
Nits from the round on `88bbab8`, addressed in `9eb2f33`:

**"Migrations are disabled" mislabels compatibility-enabled tables** (codex #2) — fixed. The copy now says the data table(s) "have not opted in to migrations", which is exactly what the section's filter tests (`migrations_enabled !== true`), instead of asserting an effective disabled state the backend may not agree with.

**Remediation points at the parent while the fork's own flag drives visibility** (codex #1, claude, pi) — copy reworded, directive intentionally kept on the parent. Pointing users at the fork's own settings is not the behavior we want here: opting the fork in on its own hides this section without the parent tracking migrations, which is the wrong end state. The alert now spells out the consequence instead of implying the section will clear: *"Opt in from the parent workspace's data table settings — a fork only inherits the flag when it is created, so this fork keeps using the legacy diff."*

**Alert renders in the "No changes detected" empty state** (claude, P2 nit) — dismissed. That state is where the explanation is most needed: the section is standing chrome whenever an applicable data table exists, and with no diff rows there is nothing else on screen to explain why it is there. Collapsing it by default would hide the one sentence the change exists to deliver.

diegoimbert:
✅ Review round clean @ 9eb2f33fe47685b06b3234459b3f3ee0c62d45e8

PR #10830, opened 2026-08-25
fix(datatables): stop a fork's pg_dump restore from failing silently

diegoimbert:
## Summary

Forking a data table could fail partway through the SQL pg_dump generated, report success, and leave the fork's database half-restored.

`pg_import_dump` ran `psql --file <dump>` with no error handling at all. psql's default is to report a failed statement on stderr, **carry on with the rest of the script, and exit 0** — and the exit status was the only thing the import was checked against, so a broken restore was indistinguishable from a good one.

Reproduced against an instance data table (source: 2 tables, 1 view, 145 rows; target already holding a conflicting `invoices` table):

```
psql exit status = 0   <- what the API keyed its success on
psql:before.sql:13:  ERROR:  unrecognized configuration parameter "transaction_timeout"
psql:before.sql:70:  ERROR:  relation "invoices" already exists
psql:before.sql:107: ERROR:  column "customer_id" does not exist
psql:before.sql:110: ERROR:  relation "public.unpaid" does not exist
psql:before.sql:291: ERROR:  syntax error at or near "1"
psql:before.sql:337: ERROR:  column "customer_id" referenced in foreign key constraint does not exist
psql:before.sql:351: ERROR:  permission denied to change default privileges
objects in the target after the "successful" import: customers, invoices
```

The fork wizard showed the step as done; the data table was missing its view, its indexes, its foreign keys and every invoice row.

## Changes

- **`--set ON_ERROR_STOP=1 --single-transaction` on the psql import.** The restore stops at the first error, exits non-zero with the failing statement, and rolls back — the target is left as it was and the import can simply be retried.
- **`--no-owner` on every import's dump.** A restore replays as the target's own connection user, which can never hand ownership to the source's roles (`must be member of role "postgres"`, or the role simply not existing on another server) — harmless while errors were ignored, fatal once they stop the restore. Dropping the statements changes nothing else: what the restoring user creates, it owns.
- **`--no-privileges` only around an instance data table** (either side of the import). Windmill plants `custom_instance_user` grants in an instance database at creation, and that user owns neither the source's objects nor the target schema, so nothing else can replay them — this PR's own repro shows one (`permission denied to change default privileges`). Windmill gives the target its privileges when it creates it. Between two `$res:` databases the ACLs are user intent and are kept: dropping them would restore a `SECURITY DEFINER` function with `EXECUTE` back on `PUBLIC`.
- The other two `pg_dump_database` callers are unchanged; both flags are opt-in through a new `PgDumpOptions`.
- **The target's known GUCs are read through psql, not a tokio-postgres connection.** libpq negotiates TLS for `sslmode=prefer` and an unset mode, where `PgDatabase::connect` hands a TLS-only server a plaintext socket — the probe has to reach exactly the servers the restore itself can.
- **Preamble `SET`s the target server does not recognize are commented out.** The image ships the latest postgres client, whose dump preamble sets GUCs an older server does not have (`transaction_timeout` is PG 17+). They only tune the session the dump was taken in, but one of them failing is now enough to abort the whole restore. They are blanked in place, three bytes each, so the offsets of a multi-gigabyte data section stay put.

## Test plan

Verified against a PG 16 server with a PG 17 client — i.e. exercising the preamble path, which without the fix aborts every restore at line 13 — on an instance-backed data table:

- [x] fork schema + data → 200, 2 tables + view, 25 + 120 rows
- [x] fork schema only → 200, 2 tables + view, 0 rows
- [x] restore breaks partway (target holds a conflicting table) → 500 carrying `ERROR: relation "invoices" already exists`, target rolled back to exactly its prior contents
- [x] driven through the fork wizard end to end (fork a workspace → "Clone schema and data" → Start): the cloned database came out with both tables, the view and all 145 rows
- [x] resource-backed import across roles (the DB Manager path): source objects owned by `postgres`, target connecting as an ordinary `wm_import_user` that is not a member of it → 200, `secrets owned by wm_import_user`, both rows imported, and the `SECURITY DEFINER` function still locked down (`proacl` `{wm_import_user=X/wm_import_user}`, `has_function_privilege('public', 'peek()', 'execute') = f`, PUBLIC cannot select `secrets`). Dumping the same source with `--no-privileges` instead leaves `proacl` NULL and PUBLIC able to execute — the regression the scoping avoids
- [x] instance data table → `$res:` target owned by an ordinary role (the mirror case): 200, `customers owned by wm_import_user`, 7 rows — before scoping `no_acl` on the source too this aborted on Windmill's own `ALTER DEFAULT PRIVILEGES`
- [x] the psql probe returns a real list (342 GUCs as `custom_instance_user`, `client_encoding` present, `transaction_timeout` absent on PG 16) — an empty probe would silently comment out every preamble `SET`
- [x] `cargo test -p windmill-api-workspaces` (19 passed), `cargo fmt`

Two unit tests pin the preamble handling: that the in-place rewrite comments only the GUCs the server lacks while preserving byte offsets, and that the replay filter keeps everything but the meta-commands and session timeouts. Both check that a `SET` inside a function body, past the preamble, is never touched.

## Also: a generated initial migration was not replayable at all

`generate_initial_datatable_migration` snapshots a data table's schema with pg_dump and stores it as a migration's `code_up`, stripping only backslash meta-commands. Replaying that snapshot on another database — which is the only thing it is for, since it is recorded as already installed on the source — did not work:

```
psql:code_up.sql:158: WARNING:  no privileges were granted for "public"
psql:code_up.sql:165: ERROR:  permission denied to change default privileges
replay exit=3
```

Line 165 is `ALTER DEFAULT PRIVILEGES FOR ROLE postgres ... TO custom_instance_user`, which **Windmill itself plants** in every instance database at creation. It fails even replaying onto the same server as the same user, and takes the whole migration with it. The stored SQL also carried `SET transaction_timeout = 0;`, so a server older than the client that generated it would fail on line 12 instead.

The snapshot now dumps with `no_owner` unconditionally — the replaying user owns none of the source's objects — and `no_acl` on the same rule the import path uses: only around an instance data table, whose planted grants nothing else can replay. A resource-backed data table's ACLs are the user's own and are kept. `strip_unreplayable_dump_lines` drops both pg_dump's psql meta-commands (as before) and the preamble's session timeouts, the GUCs that come and go across versions. Only the preamble is filtered, so an object's own body is left exactly as it is. `SET check_function_bodies = false` is deliberately kept: without it a plpgsql function whose body names a table created later in the dump fails to restore.

Verified end to end:

- [x] generated snapshot has no `OWNER TO`, no `GRANT`, no `ALTER DEFAULT PRIVILEGES`, and no session timeouts — 141 lines, down from 171
- [x] replaying it into an empty database as `custom_instance_user`: exit 0, `customers, orders, recent` and the `count_orders()` plpgsql function all created (it names `orders`, declared later in the dump — the case `check_function_bodies` covers)
- [x] through the real migration job runner, against a second data table pointed at an empty database: `{"applied":[{"version":20260827092128,"name":"initial"}]}`
- [x] resource-backed data table with a `SECURITY DEFINER` function whose `EXECUTE` is revoked from `PUBLIC`: the snapshot keeps the `REVOKE`, replays with exit 0, and `has_function_privilege('public', 'peek()', 'execute')` is still `f` on the replayed database

---

**One judgement call worth a reviewer's eye.** When an instance data table is the *source* of an import into a `$res:` database, its ACLs are dropped along with Windmill's own planted grants — so a `REVOKE ... FROM PUBLIC` a user set through the SQL editor would not reach the target. The alternative is a UI-reachable import that fails with an unactionable `permission denied to change default privileges` and no workaround, so this side of the trade was chosen deliberately rather than on a reviewer's request.


diegoimbert:
✅ Review round clean @ d0e4ea82c566566ff97e106aa09866459af70fb5 — Codex, Claude and Pi all "Good to merge", no nits outstanding.

PR #10831, opened 2026-08-25
fix(frontend): key the GitHub App installation selector on installation_id

hugocasa:
## Summary

A workspace can hold several GitHub App installations for the same org: `install_from_workspace` and the GHES assign path dedup on `installation_id` only, so re-installing an app or pulling one in from another workspace leaves two entries sharing an `account_id`.

The GitHub Account ID dropdown used `account_id` as the `<option>` value **and** as the `.find()` key, so both entries were indistinguishable. Picking the live installation resolved to whichever came first, which could be a stale, token-errored one whose `repositories` array is empty, leaving the repository dropdown with nothing in it. `RepositorySelector`'s "load more" pagination matched the same way and appended the wrong installation's page.

Everything now keys on `installation_id`, which is unique per installation. The label still shows the org name, with the installation id appended only for orgs that appear more than once.

Fixes WIN-2448

## Changes

- `frontend/src/lib/githubApp.ts`: `selectedGHAppAccountId: string` → `selectedGHAppInstallationId: number`. Dropped `getRepositories()` and the `GitHubRepository` type it returned — both unreferenced, and `getRepositories()` carried a third copy of the same `account_id` lookup.
- `frontend/src/lib/components/GitHubAppIntegration.svelte`: the select binds and looks up by `installation_id`; the placeholder option now actually shows when nothing is selected (it was `value=""` against an `undefined` state, so the browser fell through to the first enabled installation); duplicated org names get ` (<installation id>)` appended so the two entries are distinguishable.
- `frontend/src/lib/components/RepositorySelector.svelte`: `accountId: string` prop → `installationId: number`, and `loadMoreRepositories` matches on it.
- Switching installation now remounts `RepositorySelector` (`{#key}`) and clears the selected repository. `RepositorySelector` snapshots `initialRepositories` into local state at mount and tracks its own page cursor, so without the remount a switch kept showing the previous org's repositories and a stale "Loaded N of M" count.

## Test plan

Exercised against a local instance with the installations endpoint stubbed to return two installations for one org (`acme-corp`: `111111` token-errored with no repos, `222222` healthy) plus a second org:

- [x] Options carry unique values; the duplicated org reads `acme-corp (111111, token error)` / `acme-corp (222222)`, the unique org stays plain `other-org`
- [x] Nothing preselected on open — the placeholder shows instead of the errored installation
- [x] Selecting the healthy duplicate populates the repository dropdown (this is the reported bug; before the fix it was empty)
- [x] Switching org refreshes the repository list and clears the previous selection, re-disabling Apply
- [x] "load more" with `total_count` 5 / `per_page` 3 appends the right installation's page 2 (3 of 5 → all 5); under the old lookup it would have hit the empty errored installation and appended nothing
- [x] Apply writes `url` + `is_github_app` onto the resource
- [x] `svelte-check`: 0 errors

## Screenshots

Two installations for `acme-corp`, the live one selected.

Before — the errored installation is picked up and the repository dropdown is empty:

before-duplicate-org

After — the two entries are distinguishable and the repositories load:

after-duplicate-org

## Not in scope

Local review surfaced an adjacent pre-existing backend issue, left alone here because it is EE backend behaviour rather than a regression from this diff: `installation_callback` (`git_sync_ee.rs`) upserts by replacing **every** array element whose `account_id` matches. A workspace that already holds two entries for one org and then receives a fresh callback for that org would collapse both onto the same `installation_id`, which duplicates the `{#each}` key. That key expression is unchanged by this PR, but this PR is what makes `installation_id` the identity of the selection, so it is worth a separate look.

### DIFF
--- a/frontend/src/lib/components/GitHubAppIntegration.svelte
+++ b/frontend/src/lib/components/GitHubAppIntegration.svelte
@@ -55,6 +55,19 @@
 			)
 	)
 
+	// Org names that appear on more than one installation, so the dropdown can
+	// tell those entries apart.
+	let duplicatedAccountIds = $derived(
+		new Set(
+			githubState.workspaceGithubInstallations
+				.filter(
+					(installation, _, array) =>
+						array.filter((other) => other.account_id === installation.account_id).length > 1
+				)
+				.map((installation) => installation.account_id)
+		)
+	)
+
 	let showGitHubApp = $derived(
 		resourceType === 'git_repository' &&
 			$workspaceStore &&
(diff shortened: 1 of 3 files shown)


hugocasa:
✅ Review round clean @ 98ff933a35d37fc57b923bb1f4b9bf3e2af04d0e

PR #10832, opened 2026-08-25
feat(ai-sessions): make a running session visible and safe across browser tabs

hugocasa:
## Summary

An AI session lives in the browser tab that opened it, so a session open in two tabs is two independent copies of one conversation. Both can send, both append to the same chat id, and whichever `saveChat` lands last silently drops the other's turn — after its tool calls have already run against the workspace.

This makes a session usable from more than one tab: one tab drives a turn, the others show what is running and stay out of the way until it lands.

A watching tab keeps showing its own transcript, draws the ordinary loading indicator and the prompt the run is working on, and re-reads the shared record once the turn ends. Nothing that crosses between tabs carries a transcript.

## How it works

- **One driver per session.** `sessionRunOwner` holds one position per session (`idle` / `driving` / `watching` / `catchingUp`) and every cross-tab gate is derived from it, so they cannot disagree. A turn is bracketed by a Web Lock, which is held by the tab rather than by a record someone has to clean up, so a driver that dies mid-turn releases it and the next send succeeds.
- **A status heartbeat, not a transcript.** The driving tab broadcasts `{ loading, compacting, blockedOnUser, loadingLabel, planModeActive, userMessage }` once a second. Every field is of fixed size, `userMessage` by truncation at the source — it is the one field whose length a user sets. A watching tab renders the same loading indicator a local turn shows, so a run reads as live without any of it having to travel.
- **The prompt is echoed, not filed.** `userMessage` is drawn as the user bubble it stands in for and is never written to `displayMessages` or to the store, so a watching tab can say *what* is running without holding a transcript it does not own. It is pinned per run, searched only past where the transcript stood when the run began: the first status goes out at guard entry, before the send has appended the user message, and reading whatever is last there would echo the previous turn's prompt and then swap it. The echo is suppressed once the transcript carries the real message, and cleared by the catch-up that brings it in.
- **Turn end is a re-read.** `turn-end` names the chat the driver finished on (a `/clear` rotates it mid-turn), and the watching tab re-reads that record from the shared IndexedDB store — transcript, API-format history, context usage, edits mask, background jobs. That re-read is what releases the tab to drive the next turn.
- **The composer is locked while another tab drives**, and unlocks on its own once the re-read lands. Locking rather than queueing is deliberate: a watching tab holds the conversation from before the driver's turn, and a turn sent from it would reach the model as a conversation the driver has already moved past, then persist over what that tab saved.

Records and artifacts travel between tabs by id; the receiver re-reads. Broadcast delivery order and IndexedDB commit order are independent, so shipping copies would let an older write land on top of a newer one. Stop is the one control that crosses, because it hangs off `loading` rather than off a card.

## Changes

- `sessionRunOwner.svelte.ts` (new) — one run position per session, the Web Lock, and the dead-driver reaper.
- `sessionSync.svelte.ts` (new) — the `BroadcastChannel` between tabs of one session. Carries record ids, run status, turn-end, and control messages. No message carries a transcript.
- `sessionRuntime.svelte.ts` — posts run status while driving, applies it while watching, pins the run's prompt, and performs the turn-end re-read with a bounded retry when the store is briefly unreadable.
- `AIChatManager` / `AIChatDisplay` / `AIChatInput` — gate sending, saving, compaction, history switching and deletion on `runHeldElsewhere`; disabled composer treatment; draw the run indicator and echoed prompt when this tab holds no transcript for the running turn.
- `sessionState.svelte.ts` — a record another tab wrote is merged into the object this tab holds rather than swapped for it, so the debounced writers keep the object they captured and the preview tabs keep the live editor's transient labels.
- Artifacts (`artifactsState`) — a tab that did not write an artifact re-reads it, so a plan awaiting approval has its document in both tabs.

## Non-secure origins

Web Locks is secure-context only. Every HTTPS deployment and `localhost` takes the real lock; an instance served over plain HTTP falls back to refusing while another tab's run is visibly on screen, which is not mutual exclusion and cannot be made into it — a driver throttled in a hidden tab goes silent long before its turn ends, and nothing over the channel distinguishes a frozen tab from a closed one. Accepted rather than solved, and stated at the code.

Verified on a plain-HTTP non-loopback origin (`isSecureContext: false`, `navigator.locks` absent): `BroadcastChannel` and IndexedDB are not gated, so watching, the echoed prompt and the turn-end catch-up behave exactly as on `localhost`.

## Screenshots

Two tabs on one session. Tab A is driving; tab B is watching.

**While A runs** — B keeps its own transcript, draws the prompt A is working on and the loading indicator with its elapsed timer, and the composer is disabled reading "Running in another tab":

watching tab during a remote turn

**After A's turn ends** — B has re-read the record and holds the complete turn (reasoning, tool output, prose), composer unlocked, and the echoed prompt has given way to the real message:

watching tab after catch-up

**While A is parked on a question** — B says where the run is waiting rather than implying work is in progress:

watching tab while the driver waits on an answer

## Test plan

Driven against a real Anthropic provider, two tabs on one session in one browser profile:

- [x] B locks at the first heartbeat (t=0s) when A sends, and draws A's prompt from that point.
- [x] B's transcript stays flat for the whole turn — 700 to 701 characters, the 1 being the indicator's elapsed-time counter. Nothing carries a transcript, and this is the measurement that says so.
- [x] B unlocks and its transcript jumps to 1957 characters in one step: the catch-up re-read, not a stream.
- [x] The prompt appears exactly once after catch-up, so the echo gives way to the real message rather than doubling it.
- [x] B ends holding the same conversation as A.
- [x] Tapping the `BroadcastChannel` from a third tab: the run posts no prompt on its first frame and the pinned prompt thereafter, never the previous turn's. An identical prompt sent twice in a row still echoes.
- [x] An open tab shows "Waiting for your answer in the other tab" while A is parked, and no interrupted tool call.
- [x] A second tab opening mid-session finds the session and initialises against the committed store.

Unit coverage: run position transitions and lock refusal (`sessionRunOwner.test.ts`), the send/save/queue refusals and their unwinding (`AIChatManager.test.ts`), cross-tab artifact put and remove (`artifactsState.test.ts`), delete-behind-checkpoint ordering (`HistoryManager.test.ts`), remote-record merge and the silent mirrored delete (`sessionState.test.ts`, `sessionStateIndexedDb.test.ts`), catch-up ordering and supersession (`latestWins.test.ts`).

`npm run check` and the frontend suites: no errors or failures in files this branch touches; the ones on the branch are `main`'s (enumerated in-thread).

## Accepted limitations

A run parked on a question or a tool confirmation can be answered only in the driving tab. The watching tab shows that the run is waiting on you, but the card and the resolver waiting on the answer exist only where the turn is running.

This is why the watching tab does not trust its own transcript for that state: `loading` now describes the driver's run while `messages` is this tab's own conversation from before it, and a card restored from history still looks parked. Both derivations that pair the two are gated on the run being local.

A tab that *loads or reloads* while the driver is parked on a question is a separate case, and it reads the question as `Interrupted`. The mid-turn checkpoint stores open tool calls closed-as-interrupted so a crash can be recovered from it, which is what that state means only if the writing tab died; a tab mounting meanwhile reads the same record and cannot tell the two apart. Left as is: suppressing that write costs the last-chance checkpoint that exists to stop a closed tab losing a turn's progress, and a reader cannot separate these rows from genuinely interrupted ones in earlier turns. It resolves itself when the turn ends and the tab re-reads. Tabs already open when the question appears are unaffected, since they never re-read mid-turn.

**A frozen driver reads as a closed one.** The reaper's only signal is silence, and a tab the OS has frozen past the reaper's threshold is silent. Watchers stop showing its run early; the Web Lock is still held, so a send from them is refused rather than double-driven. The cost is a stale indicator, never a second driver.

**The lock handoff has a sub-heartbeat window.** A tab that mounts between the driver's last heartbeat and its turn-end, and whose user sends inside that window, can take the freed lock while still holding pre-turn history. Closing it means making a re-read of the persisted chat a precondition of acquiring the lock — a change to the handoff contract that belongs in its own PR, not a guard at a site.

**Conveniences are removed, not debugged.** The echoed prompt sits above this feature's bar (one master, visible progress, visible waiting-for-input, convergence at turn end). A future defect in it is answered by deleting the convenience, not by extending its machinery — the sidebar draft cue already went this way once the prompt field's new ownership left it reading a signal that no longer meant a draft.


hugocasa:
On the frozen-tab probe finding (`sessionRunOwner.svelte.ts`): fixing the other two, but I don't think this one is closeable and I'd rather name the limit than pretend otherwise.

A tab the browser has frozen runs no script at all, so it answers a probe exactly the way a closed tab does. Nothing sent over a BroadcastChannel can tell those apart, and no shared-storage heartbeat helps either — a frozen tab stops writing it for the same reason. Mutual exclusion on this path needs a primitive the origin doesn't have.

What the probe does buy is the case that actually happens: a *throttled* driver. A hidden tab's timers drop to as little as once a minute, so its heartbeat stops long before its turn does, and the reaper retires it well inside a live run. Message delivery is a task rather than a timer, so that tab still receives and answers.

Worth noting the baseline: before this PR two tabs on one session could both send at any time, with no coordination at all. On a lockless origin this is strictly better than that, never worse — and it is exactly why secure origins take the Web Lock instead. The guarantee is documented as best-effort in the PR body's Degradation section, and the code comment at `bestEffort` now says the same thing in place.

hugocasa:
Update on the lockless-origin findings across the last few rounds (the reaping/probe cluster): the driver probe is now **removed**, and the limitation is documented instead.

Rationale: the only origins that reach that path are served over plain HTTP *and* are not `localhost` — every HTTPS deployment and every `localhost` session is a secure context and takes the real Web Lock. That shape is used here for local testing, not for running Windmill, so it does not warrant machinery that cannot actually deliver the guarantee.

And it can't: a probe separates a throttled tab from a closed one, but not a frozen tab from a closed one, and browsers freeze hidden tabs on roughly the schedule they throttle them. The probe bought a slice of the window in exchange for a pre-flight delay on every send, and generated three findings of its own across two rounds. Net −102 lines removing it.

What remains on that path is stated plainly in `bestEffort`'s doc comment and in the PR body's Degradation section: no mutual exclusion, which is what the session had before this PR anyway. Please read further findings there as accepted rather than unaddressed.

hugocasa:
On the catch-up retry not converging for a permanently failed IndexedDB open — settling this rather than flipping it a third time, since the two reviews have pushed opposite ways on the same line across rounds.

It keeps retrying, and the gate holds. The alternative is releasing the gate, which is the P1 from the previous round: this tab would then send a mirrored transcript paired with pre-run history, so the driver's completed turn is missing from what reaches the model and its record gets overwritten. Mirror frames carry `displayMessages` and never `messages`, so a store read is the only thing that can make the pair whole again — there is no bounded number of attempts after which sending becomes safe.

The cost of holding is one IndexedDB read every five seconds for a session that was mirroring, until the tab closes or the runtime is disposed. And a browser whose IndexedDB never opens has no session history, no artifacts and no session records either, so a locked composer is not the thing that is broken for that user.

Reasoning is now in the code at `scheduleCatchUpRetry` so it does not get re-litigated.

hugocasa:
On the `#inGuardedRun` / concurrent-send finding — declining it for this PR, with the reasoning rather than a shrug.

It is a real window, and the raw-app route reaches it: `RawAppEditorView` gates on `manager.loading` but not `sendInFlight`, while `#inGuardedRun` is set inside the guard body and `loading` only rises later inside `sendRequestImpl`.

But it predates this branch. On `main`:
- `runGuard` does not exist at all (0 occurrences in `AIChatManager.svelte.ts`)
- `RawAppEditorView`'s gate is byte-identical: `if (runtime.manager.loading) queueMessage(...) else sendRequest(...)`

So the second concurrent loop on one manager — sharing its abort controller and transcript — is reachable on `main` today through exactly this window. What this PR adds is a guard that closes that hazard everywhere except the same pre-existing window. It strictly narrows the bug; it does not create or widen it.

Fixing the class properly means replacing the ambient `#inGuardedRun` with an explicit reentrancy flag on the call, which requires classifying all nine `this.sendRequest(` sites in the manager as reentrant or not — getting one wrong turns a legitimate nested send (the queued-message flush, the background-job auto-resume) into a refused turn. That is its own change to the send path's concurrency model and belongs in its own PR, not appended to a cross-tab one that has already converged.

Also noting for the record: the Pi review on this PR is evaluating head `23d324c` and quotes the plan-mode "Your message stays queued" toast, which was deleted several commits ago and does not exist on HEAD.

hugocasa:
Addressed this round. On the rollback-deletion P1 specifically, I dug into how this codebase treats an unavailable IndexedDB rather than adding a mechanism, and the answer changed the fix.

`userScopedDb.whenReady()` states its own contract: it returns `undefined` when nobody is logged in yet, the open failed, or another connection is holding up a schema upgrade — *"degrade to in-memory; never rejects, never hangs"*. That is a deliberate design choice, not an oversight, and the call sites follow it: of 19 `if (!db)` guards under `components/`, 17 degrade silently (`return`, `return []`, `return 0`). The only two that report unavailability are the two this PR added.

So a dropped delete is the house behaviour, and the real error was mine: `await`ing `deletePastChat` implied a barrier the write queue never offered. The comment now says what it actually provides — ordering on the write queue, not a guarantee — and names the residue: a watcher whose own read still succeeds can find the rolled-back transcript. That needs the store to be unavailable for the driver's delete but available for the watcher's read, which are separate tabs and separate handles. Where the watcher's read also fails, the catch-up gate already holds rather than releasing.

I did not add a `rolledBack` flag to the turn-end message for this. It would be a wire-format change to remove a dependency on a write that this codebase drops everywhere else, and the gate already covers the correlated case.

Also fixed from this round: a refused synthetic auto-resume left `this.instructions` set, which disarmed every later auto-resume for the session (now cleared, with a regression test); the no-composer fallback dropped paste blobs while keeping their tokens (now expanded into the parked text); and the catch-up comments are trimmed to the AGENTS.md limit — 13 lines to 4, with the duplicates on the helper and the test removed.

hugocasa:
## Thread cleanup after the redesign

`e13c713969` replaced transcript mirroring with a run-status heartbeat, which deleted the code a large share of the open threads were written against. I've resolved 23 of them so the remaining ones are actually actionable:

**Moot — the code no longer exists** (11): frame prefix/splice validation, `applyMirror` writing into the watcher's live manager, the `AttachedImage.url` strip, the `committed` flag on turn-end, `pastes`/`result` payload stripping, the tail overwriting complete messages, the `/clear` rotation branch, the positional restore, and the three driver-probe threads. `sessionMirrorPayload.ts` and both its test files are gone; so are `MirrorMsg`, `ResyncRequestMsg`, `applyMirror` and `answerResync`.

**Fixed** (12): the resend reservation leak, the `watching` vs `watching|catchingUp` gate (now one check at the single entry to `withSessionRunLock`), the dropped programmatic prompt, both `prependText` threads, `restoreRefusedSend` restoring an empty prompt, the auto-resume unwind, the fallback comment, the unconditional `instructions` write, and the `restartGeneration` mutate-before-refuse (hoisted above every `draft.take()` in `AIChatInput`).

**Still open and still mine to answer** — not resolved: the `setInterval` heartbeat in a hidden tab, the mask write dropped rather than deferred, `reloadChat`'s `'unavailable'` retry, delete-during-init, the `cancel-request` turn identity, the artifact broadcast-before-commit pair, the inherited `cursor-not-allowed`, and the two test-isolation P3s. Plus the three findings just filed against the new design, which I agree with and am fixing: the `loading` flag re-arming question affordances in a watching tab, the now-unreachable tool-confirmation/question-answer routing, and the unread `chatId` on `RunStatusMsg`.

If I've resolved something you still consider live, reopen it — the triage is mine, not a claim that a reviewer was wrong.

hugocasa:
## Thread sweep @ 46b77fe363

Resolved 19 more threads, each verified against the current head rather than closed on age. Fourteen remain, and they are the real list.

**Fixed since filed** — the cross-tab queue flush and its toast (`flushQueuedMessage` no longer exists in `sessionRuntime`), `mirroredPlanMode` retirement (the posture rides the watching position and dies with it), `lastTurnAcceptsFollowUp` (deleted), the "New chat" control (now visibly disabled), the no-runtime map entry, Stop turn-identity, the answer-with-attachment Stop/Send flip (`pendingQuestionToolCallId` is gated on `runIsLocal`, so it is always undefined in a watching tab and the composer is disabled outright), plus all seven from the last two rounds.

**Superseded by the redesign** — the stuck-on-generating and unfiltered-`turn-end` P1s: the reaper releases a watcher after `DRIVER_SILENCE_MS`, and `canCompleteCatchUp()` settles a tab with no runtime instead of parking it in a state nothing can leave.

### Still open (14)

| Severity | What | Status |
|---|---|---|
| P1 | `#inGuardedRun` concurrent send | Declined — pre-existing; `runGuard` has zero occurrences on `main` and the gate it replaces is byte-identical there |
| P1 | `whenReady() === undefined` read as "no store" | Declined — degrade-silently is house style, 17 of 19 guards do it |
| P2 | Artifact write/delete broadcast before IndexedDB confirms | Open — the one I would still like a second opinion on: it is cross-tab divergence, not local degradation |
| P2 | Discard/deploy mask dropped, not deferred | Answered in code: the turn-end re-read reseeds the mask, so a deferred write would be overwritten anyway |
| P2 | Heartbeat `setInterval` in a hidden tab | Accepted and documented — on a secure origin the lock query blocks a false reap; plain HTTP has no backstop, by design |
| P2 | `reloadChat` `'unavailable'` retry never converges for a permanently dead store | Deliberate — polling beats releasing the gate and sending pre-run history |
| P2 | Reaping on silence hands the run back | Same trade as above, stated at `bestEffort` |
| P3 | `cursor-not-allowed` inherited by chip buttons | Open, cosmetic |
| P3 | Test module-global leakage (`onDriverLost`, `vi.resetModules`) | Open — real hygiene, and adjacent to the fixture bug that bit round 14 |

Session-deleted-during-`initRuntime` (cubic P1) is fixed in `46b77fe363`: every await in `initRuntime` is now followed by a liveness check, so a disposed runtime's continuation cannot write — `setSessionChatId` would otherwise have put back a record the user just deleted.

Reopen anything you disagree with; the triage is mine, not a claim the reviewer was wrong.

hugocasa:
## Round @ `d441f3c5a3` — all three nits addressed in `ad8fdb24bf`

Nit-only verdicts from Codex, Claude and Pi; nothing from Cubic. All three are fixed.

**`this.instructions` left staged after a refused send** (Claude P2, Pi P2 — found independently)

This is the one that mattered, and it was a real regression from this branch. Dropping the
`restartGeneration` precheck meant the field is staged and then left set when the send is refused;
only `sendRequestImpl` clears it, and a refusal never gets there.

I had considered this when it was first raised and talked myself out of it, on the reasoning that a
staged prompt is equivalent to the user having an unsent draft. That was wrong on a fact I hadn't
checked: `AIChatInput.svelte:76` documents that `aiChatManager.instructions` *only carries
programmatic prompts* — the composer's text is separate local state. So two readers take a non-empty
value as user activity:

- `#maybeAutoResumeFromJobs` — background-job auto-resume goes dead for the session
- `SessionPicker.hasDraft` — a sidebar cue that never clears

Both are accidentally right while the restored text sits in the composer, and both go wrong the
moment the user clears it. Fixed with one line at the hand-off in `restoreRefusedSend`, which covers
`restartGeneration`, `fix` and `askAi` together (and `sendRequestImpl`'s own pre-flight refusals).

The existing test was asserting `instructions === 'first'` — i.e. pinning the leftover. It now
asserts the hand-off instead: the prompt lands on the queue and the field is released.
Mutation-verified (`expected 'first' to be ''`).

**Duplicated rationale in a test comment** (Codex P2) — trimmed; the invariant stays at the
production parameter.

**`bestEffort`'s opening sentence** (Claude P3) — it described a refusal that moved to
`withSessionRunLock` two commits ago. Reworded.

### Not addressed, deliberately

**Lock-handoff catch-up.** A tab that mounts inside the sub-heartbeat window *and* whose user sends
in it can take the freed lock holding pre-turn history and overwrite the completed turn. Real, and
the loss persists rather than being repaired by the turn-end re-read. Not fixed here because the fix
is "re-read the persisted chat when acquiring the lock", which changes the handoff design rather
than patching a site — out of scope for this PR by decision, and worth its own change.

**`npm run check` errors.** 17 on this branch, all pre-existing on `main` (`ServiceLogsInner`,
`PremiumInfo`, `SidebarUsage`, `hubProject`, `projects/import`, and `aiStore` / `EditableTextarea`
which arrived with #10020). None are in this diff; `main` has components referencing a client that
was never regenerated. Not regenerating it here — that belongs on `main`, not in a frontend PR.

585 frontend tests green.

hugocasa:
## Round @ `ad8fdb24bf` — the P1 and both P2s addressed in `3358fe6376` + `8789f5b9af`

**Retry/edit could blank a concurrent winner's prompt** (Claude P2) — fixed at the ownership
level rather than with the suggested guard. The guard compares by value, and the one case that
matters defeats it: a double-fired retry stages the same text the winner is running on, so
equality holds and the clear still lands mid-turn. Instead, every sender now passes its prompt in
`options.instructions` (retry/edit, AI Fix, Ask AI, auto-resume, the raw-app hand-off page), the
field is written only by `sendRequestImpl`, and `restoreRefusedSend` reads only its own options
and clears nothing. A refusal can no longer read — or unwind — text it does not own, which also
closes the sibling edge where the loser handed the *winner's* text back to the composer.
Auto-resume's conditional unwind is deleted with the staging that made it necessary. New test pins
that a refusal parks its own prompt and leaves a concurrent winner's field untouched.

**Overlapping turn-end catch-ups** (Codex P1) — catch-ups are now serialized per session,
newest-wins: a small `latestWins` primitive runs them strictly in order, skips one superseded
before its turn, and hands the running one a `superseded()` probe checked after `reloadChat`, so
a stale catch-up stops before `loadPastChat` can re-persist the chat the driver already left. All
three entry points (turn-end, driver-lost, the unavailable-store retry) queue through it. The
primitive carries its own suite (ordering, supersession skip, in-flight probe, failure isolation)
— the requested overlapping-catch-up case, pinned at the layer that owns the ordering.

This closes the last unversioned async write-back in the sync layer: `applyRemoteSessionPut`
(read tokens + local-write sequence), `applyRemoteArtifact` (session re-check after its await) and
`setGeneratedSessionSummary` (chatId + placeholder re-check) already carried the discipline.

**Duplicated staged-prompt rationale in the retry test** (Codex P2) — trimmed to the assertions.

Also: the PR description's limitations section now records the accepted edges discussed in-thread
(frozen driver, sub-heartbeat lock handoff) and the policy that the echoed prompt and sidebar
draft cue are conveniences to be removed, not debugged, if a future finding lands on them.

Verified beyond the suites: real-provider run on a live instance — composer send, edit-resend
(the path whose prompt routing changed), and a two-tab turn with the watcher catching up to the
completed turn in one re-read. 564 session/chat tests green; `npm run check` clean for touched
files.

hugocasa:
## Round @ `8789f5b9af` — both P1s and the nits addressed in `3e25ae7b00` + `9815c6c0a1`

**Catch-up gate released early** (Codex P1) — the `superseded()` probe now also runs after
`loadPastChat`, before `caughtUp` is set, so a catch-up overtaken during image hydration leaves
the gate to the newer one instead of unlocking the composer over a transcript missing that turn.
Every await in `applyTurnEnd` is now followed by a probe before any write or gate release. (Pi
dismissed this one as redundant with the tail chain, but the chain only orders the *writes* — the
early return here is about the *gate*, which the overtaken task's `finally` released.)

**Stale retry supersedes a newer turn-end** (Pi P1) — closed structurally rather than checked
for: `queueCatchUp` cancels any pending unavailable-store retry, and a catch-up that hits
`'unavailable'` schedules its own with its own chat id. A retry can no longer exist once a newer
turn-end has queued, so there is nothing left to steal newest-wins. On the requested test: the
scenario is now unreachable rather than guarded, and the primitive's suite already pins the
supersession semantics; a runtime-level harness (faked channel + stubbed HistoryManager) would be
the first of its kind in that layer, which both reviews agreed is a fair place to stop.

**The unenforced field contract** (Claude P2) — took the enforcement option: `sendRequestImpl`'s
assignment is unconditional (`options.instructions ?? ''`), so the prompt is exactly what the
options carry and a caller that stages anyway fails loudly at the empty-draft check. All test
fixtures that staged the field are converted to option-passing — the suite now runs through the
enforced path, and the three remaining direct writes are deliberate pins (refusals and
explicit-empty sends leave the field alone). The impl comment now states the same contract as the
field doc. Audited before flipping it: no production writer of the field remains outside
`sendRequestImpl`, and every caller passes `instructions` (the `AIChat.svelte` passthrough has no
caller).

**Comment nits** (Codex P2, Claude P2/P3) — the two test comments repeating production rationale
are trimmed; the auto-resume guard and `SessionPicker.hasDraft` comments no longer call the field
the user's unsent draft; the resend-context rationale moved next to the `contextOverride` option
it documents.

Verified: 455 session/chat tests green through the enforced path; the two-tab `/clear` rotation
driven live (watcher lands on the fresh empty chat, both tabs agree); `npm run check` unchanged
(the 17 errors are main's, enumerated earlier).

hugocasa:
## Round @ `9815c6c0a1` — the one finding all three of you converged on, fixed in `0c6568a189`

**The probe-less catch** (Codex P1, Pi P1, Claude P2 — the same bug, found independently by all
three) — the `catch` block now makes the same `superseded()` check the `'unavailable'` path
already sits behind, so a superseded catch-up whose `loadPastChat` threw leaves the gate and the
retry to the newer catch-up instead of arming one with its stale chat id. With this, every path
out of `applyTurnEnd` — both awaits, both store-failure answers — is behind a probe; there is no
way left for a superseded task to write, settle, or re-queue.

**The failure case that never failed** (Claude P2) — real: both `run` calls were synchronous, so
the throwing task was superseded before it ran and the `try`/`catch` went unexercised. A tick
between them lets it actually throw, and the case now asserts the error was seen as well as the
follow-up task running.

**On the requested runtime-level test** (Pi, Codex): with the guard in place the scenario is
unreachable rather than guarded-against — a superseded task cannot arm a retry on any path — and
the primitive's suite pins the supersession semantics the wiring leans on. A
`sessionRuntime`-level harness (faked channel, stubbed `HistoryManager`) remains the first of its
kind in that layer; stopping short of it is the call all three reviews have endorsed in previous
rounds, and it holds better now that the failure path is exercised at the primitive.

**Held deliberately — `hasDraft` (Claude P3):** agreed the cue now reads a field it can barely
see, and the PR body's conveniences rule points at deletion rather than documentation. Removing a
visible sidebar cue is a product call beyond this round's findings, so it waits for an explicit
decision rather than riding a fix commit.

166 tests green across the touched suites; `npm run check` unchanged.

hugocasa:
## Round @ `0c6568a189` — Claude's P2 fixed in `07b033f458`; the held draft-cue deletion rides along in `054b65cdba`

**The disposed-runtime catch-up** (Claude P2) — an identity check (`runtimes.get(sessionId) !==
runtime`) now sits beside each supersession probe, so a catch-up whose session was deleted
mid-read returns instead of re-arming the orphaned manager's job poller or re-taking the store
handle disposal closed. It is the same discipline `initRuntime` already carries, applied to the
one async writer in the layer that still lacked it.

**`hasDraft` deleted** (Claude P3, previous round — held for an explicit decision, now made) —
the cue read a field that, under the new ownership, could only ever surface a pre-flight-leak
artifact; the status dot already covers the in-flight case. Removed per the PR body's conveniences
rule (the body's policy line now records the rule having fired), along with the unused icon
import. The unread badge is unchanged, just unnested from the wrapper the cue shared.

With Pi and Codex both at **Good to merge** with no findings this round, this should be the
confirming pass. 455 tests green through the touched suites; `npm run check` unchanged.

hugocasa:
Superseded by #10916 — same feature rebuilt on a deliberately minimal design (stateless heartbeat/turn-end messages + idempotent IndexedDB re-read).

PR #10835, opened 2026-08-25
feat: configurable expiry for presigned s3 public url signatures

diegoimbert:
## Summary

`sign_s3_objects` hardcoded a 12-hour expiry for presigned S3 object signatures, so callers had no way to mint shorter-lived (or longer-lived) public URLs. This adds an optional `expiry_secs` to the request body and threads it through the TypeScript and Python SDKs.

Omitting the field keeps the previous behaviour exactly (12h). Out-of-range values are **clamped** to `[60, 604800]` (1 minute – 7 days) rather than rejected, so an over-eager caller gets a safe URL instead of an error. The spec deliberately does not declare `minimum`/`maximum` — that would tell a validating client the request is invalid, while the server actually accepts it and clamps.

The expiry is now computed once per request instead of once per object, so all objects in a batch share one `exp`.

No change was needed on the validation side — `validate_s3_signature` already signs `exp` into the HMAC message and rejects `exp < now`.

Fixes WIN-2450

## Changes

- `backend/windmill-api/src/apps.rs`: `S3TokenRequestBody.expiry_secs: Option<i64>`, defaulted to 12h and clamped to `[60, 604800]`; `exp` hoisted out of the per-object loop.
- `backend/windmill-api/openapi.yaml`: optional `expiry_secs` on `/w/{workspace}/apps/sign_s3_objects`, documented as clamped.
- `typescript-client/client.ts` (+ `client.d.ts`): `signS3Objects`, `signS3Object`, `getPresignedS3PublicUrls`, `getPresignedS3PublicUrl` take an optional `{ expirySecs }` option.
- `python-client/wmill/wmill/client.py`: `sign_s3_objects`, `sign_s3_object`, `get_presigned_s3_public_urls`, `get_presigned_s3_public_url` (methods and module-level wrappers) take an optional `expiry_secs`. The request body omits the field entirely when it is `None`, since the spec declares it optional but not nullable.
- `backend/tests/sign_s3_objects_authz.rs`: a second test pins the default and both clamp bounds against the returned signature's `exp`.
- `system_prompts/auto-generated/**` and `cli/src/guidance/skills.gen.ts`: regenerated via `python system_prompts/generate.py` (they embed the SDK signatures).

No frontend source changes — `frontend/src/lib/gen` is generated and gitignored, and nothing in `frontend/` calls these functions, so there is no visible UI effect to screenshot.

## Test plan

Validated against a local backend built with `--features quickjs,parquet,private` (the endpoint's permission check is EE-only, so a CE build can't exercise it):

- [x] `cargo test --features enterprise,private,parquet --test sign_s3_objects_authz` — 2 passed
- [x] `typescript-client`: `bash build.sh` + `tsc -p tsconfig.json` — clean against the regenerated `services.gen`
- [x] End-to-end against `POST /w/admins/apps/sign_s3_objects`, decoding `exp` from the returned `presigned` string:

  | request | resulting `exp - now` |
  |---|---|
  | no `expiry_secs` | 43200 |
  | `"expiry_secs": null` | 43200 |
  | `"expiry_secs": 300` | 300 |
  | `"expiry_secs": 1` | 60 (clamped) |
  | `"expiry_secs": 604800` | 604800 |
  | `"expiry_secs": 99999999` | 604800 (clamped) |

- [x] Two objects in one request share the same `exp`
- [x] `GET /w/admins/s3_proxy/...?exp=..&sig=..` accepts the fresh signature (fails later, only on unconfigured storage) and returns `400 Invalid signature: MAC tag mismatch` when `exp` is tampered with
- [x] A signature minted with `expiry_secs: 60` is rejected with `400 Signature expired` once its 60s elapse

**Note:** the `local-review-codex` pass was skipped — the devbox `codex` CLI is 0.132.0 and the skill pins >= 0.144.1; a global npm upgrade is not permitted here. The Claude-native `local-review` pass did run, and its two findings (spec `minimum`/`maximum` vs. clamping, and the missing clamp regression test) are addressed above.


diegoimbert:
✅ Review round clean @ a35265a094246c1f84146bf47de4ece1aabbd564 — nit-only verdicts at 3f9fba5010a39567fb9071754dd07fb01a83118e (codex: nit, claude: good to merge, pi: good to merge); the codex nit (Python SDK serializing `expiry_secs: null` against a non-nullable spec field) is addressed in a35265a094246c1f84146bf47de4ece1aabbd564, which omits the field when it is `None`.

PR #10838, opened 2026-08-25
fix(cli): keep svelte component styles in the raw-app bundle

hugocasa:
## Summary

A `<style>` block in a raw app's `.svelte` component never reached the bundle the CLI produces. The plugin compiled with `svelte.compile(source, { filename })` and used only the returned `js`; Svelte 5 defaults to `css: "external"`, which hands the styles back on a separate `css` field that this esbuild `onLoad` never emits. The compiled markup still carried its `svelte-<hash>` scoping class, so the app rendered with the class present and no rule matching it. No build error, no warning.

The raw-app editor's in-browser bundler already compiles with `css: "injected"`, so the same app looked correct in the editor and lost its styling the moment the CLI built it. This makes the CLI match.

Both default Svelte templates ship a `<style>` block (`frontend/src/lib/components/raw_apps/templates.ts`, `cli/src/commands/app/new.ts`), so `wmill app new --framework svelte5` followed by `wmill app push` reproduced it directly.

Scope: `wmill app bundle`, `wmill app push` and `wmill app dev` all share this plugin. `POST /apps/create_raw_source` does too, since the server-side bundle job shells out to `wmill app bundle` — that path picks the fix up once the CLI is published, since the job runs `bun x windmill-cli@<server version>`.

Not a regression: the line has been unchanged since the plugin was introduced in v1.587.0.

Global CSS was never affected — `index.css` goes through esbuild's own CSS loader and still lands in `bundle.css`, which is what made this easy to miss.

## Changes

- `cli/src/commands/app/bundle.ts`: compile `.svelte` files with `css: "injected"` so each component's styles ride in its js output. The sibling `compileModule` handler for `.svelte.ts` / `.svelte.js` is unchanged — rune modules cannot carry a `<style>` block and `compileModule` has no `css` option.
- `cli/test/raw_app_svelte_module_unit.test.ts` → `raw_app_svelte_plugin_unit.test.ts`: renamed to match what it now covers, and extended with a `component styles` case that reuses the existing fixture. It asserts the invariant that broke: whatever `svelte-<hash>` class the markup carries must have a matching rule in the bundle.

`bundle.css` keeps its current contract. `createBundle` already returns `""` when esbuild emits no CSS output, `bundle_command.ts` writes an empty `bundle.css`, and both `dev.ts` and the backend's `/app_bundles/<id>.css` route already handle a missing or empty stylesheet — component CSS was previously dropped entirely rather than routed there, so nothing downstream loses an input.

## Test plan

- [x] `cd cli && bun run test:unit` — 1095 pass, 0 fail
- [x] New test fails on the unfixed compile options and passes with them (verified by reverting the option in place)
- [x] `wmill app bundle` on the default Svelte template: `h1.svelte-n50uah {font-size:1.5rem;}` present in `bundle.js`
- [x] `wmill app new --framework svelte5` + `wmill app push`, then open the deployed app: computed `font-size` is 24px, was 32px
- [x] `wmill app dev`: served bundle carries the rule
- [x] `POST /apps/create_raw_source` against a server run with `WM_RAW_APP_BUNDLER_CLI` pointed at this CLI: stored js bundle contains the rule
- [x] Svelte app with a `<style>` block and no `index.css`: builds with an empty `bundle.css` and no error
- [x] `bunx tsc --noEmit` reports no errors in the touched files (the ~12 files that do error are pre-existing and untouched)


hugocasa:
✅ Review round clean @ cdcb35a60fef70e389ac1e9e536bea6007879cd4

PR #10849, opened 2026-08-26
fix: force HTTP router rebuild on trigger-change notification

hugocasa:
## Summary

Fixes WIN-2452.

The HTTP trigger router cache detects changes with a PostgreSQL sequence, `http_trigger_version_seq`. `increase_trigger_version` calls `nextval` on the writing transaction, and sequences are non-transactional: the bumped version is visible to every other session as soon as it runs, while the trigger row is not visible until commit.

If a refresh lands in that window — the 60s background loop, or the empty-cache fallback — it caches the *new* version against the *old* rows. When the transaction then commits and `notify_http_trigger_change` is processed, `refresh_routers` sees `version == cached_version` and skips. Every later version-gated refresh skips too, so the new route returns 404 until a pod restart or an unrelated version bump.

`refresh_routers` now takes a `force` flag. The notify handler passes `force = true`: the notify event is only readable after the writing transaction commits, so the rows are guaranteed visible there. The 60s loop, the workspaced-route setting handler and the on-demand fallback keep the version-check optimization (`force = false`) — the setting handler is safe because it bumps the sequence itself, in autocommit, after swapping the atomic.

**Not one rebuild per row.** Every changed row emits its own notify event, so forcing per event would turn a bulk trigger change into one full rebuild per row, where the version gate previously collapsed them into one. The poll loop coalesces `notify_http_trigger_change` events within a batch: the first *successful* forced rebuild reads every row the batch committed, so the rest are dropped.

**A failed forced rebuild keeps its retry.** Its change is already inside the cached version, so nothing version-gated would rebuild. Within the batch, the coalescing flag stays clear and the next event retries immediately. Across batches — the common one-trigger-one-event-one-batch shape — `invalidate_routers` bumps a counter that the periodic refresh compares against the count the cache was built with. A rebuild that started *before* the invalidation publishes the older count, so it cannot pass its own stale rows off as covering that invalidation; the next refresh still rebuilds. Routes already loaded keep being served throughout.

Also fixes a latent deadlock on a line this PR already touches: the empty-cache fallback in `get_http_route_trigger` held its read guard across `refresh_routers`, which takes the write lock. Reachable only when the startup load errored, where it would hang the request instead of erroring.

## Changes

- `refresh_routers(db, force)` skips the version comparison when `force` is set (`backend/windmill-trigger-http/src/lib.rs`)
- `notify_http_trigger_change` handler forces the rebuild; the other three call sites stay version-gated (`backend/src/main.rs`, `backend/windmill-api/src/triggers/http/handler.rs`)
- The notify poll loop processes at most one *successful* `notify_http_trigger_change` per batch; `process_notify_event` reports whether the rebuild succeeded (`backend/src/main.rs`)
- `invalidate_routers` records a rebuild the cache still owes, as a counter the cache stores alongside its version so an older in-flight rebuild cannot clear it (`backend/windmill-trigger-http/src/lib.rs`)
- Drop the read guard before the write-locking refresh in the empty-cache fallback (`backend/windmill-api/src/triggers/http/handler.rs`)
- Regression test covering both recovery paths (`backend/windmill-trigger-http/tests/refresh_routers.rs`)

## Test plan

- [x] `cargo check` and `cargo check --features http_trigger`, both warning-free
- [x] `cargo test -p windmill-trigger-http --test refresh_routers`. Both guards are load-bearing: dropping `force ||` fails it at the force assertion, dropping the invalidation comparison fails it at the invalidation assertion.
- [x] End-to-end against a backend built with `--features quickjs,http_trigger`, poisoning the cache the way the race does (bump the sequence with no row, let the 60s loop cache that version, then commit a trigger row without a further bump). Pre-fix binary: `404 Trigger not found` indefinitely. With the fix: `Refreshed HTTP routers (trigger change)` and the route answers `201`.
- [x] Three rows committed at once: three `notify_http_trigger_change` events, one rebuild, all three routes answer `201`.
- [x] In-batch retry after a failed rebuild: renamed `http_trigger_version_seq` away so every rebuild errors, then committed two rows — both events attempted the rebuild (2 attempts / 2 errors), so a failure does not consume the batch's retries. Restoring the sequence and committing a third row recovered all three routes.
- [x] Cross-batch recovery after a failed rebuild: same break, single row (single-event batch). The rebuild failed and the route 404'd; restoring the sequence with **no further trigger change** healed it in ~54s via the backstop, and the backstop then went quiet rather than rebuilding on every tick. The build without the invalidation stayed 404 past two backstop ticks.
- [x] Delete propagation in the poisoned state: deleting one of three triggers without a sequence bump took that route to `404` while the other two stayed `201`.
- [x] Workspaced-route setting change still rebuilds through the version gate: flipping `http_route_workspaced_route` moved the live route from `/api/r/test/racy` to `/api/r/admins/test/racy`.

Not exercised: the exact interleaving where an in-flight rebuild publishes after an invalidation. It needs a pause inside the build, so it is covered by construction (the published count is captured before the rows are read) and by the test's invalidation assertions rather than by a timed run.

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1825,14 +1825,12 @@ async fn process_notify_event(
         #[cfg(feature = "http_trigger")]
         "notify_http_trigger_change" => {
             tracing::info!("HTTP trigger change detected: {}", payload);
-            match windmill_api::triggers::http::refresh_routers(db).await {
-                Ok((true, _)) => {
+            // The notify event is only readable once the writing transaction has committed, so
+            // the trigger rows are guaranteed visible here: force past the version check, which
+            // a concurrent refresh may already have satisfied with pre-commit rows.
+            match windmill_api::triggers::http::refresh_routers(db, true).await {
+                Ok(_) => {
                     tracing::info!("Refreshed HTTP routers (trigger change)");
-                }
-                Ok((false, _)) => {
-                    tracing::warn!(
-                        "Should have refreshed HTTP routers (trigger change) but did not"
-                    );
                 }
                 Err(err) => {
                     tracing::error!("Error refreshing HTTP routers (trigger change): {err:#}");
@@ -2059,7 +2057,7 @@ async fn process_notify_event(
                         tracing::error!(error = %e, "Could not reload http route workspaced route setting");
                     }
                     #[cfg(feature = "http_trigger")]
-                    match windmill_api::triggers::http::refresh_routers(db).await {
+                    match windmill_api::triggers::http::refresh_routers(db, false).await {
                         Ok((true, _)) => {
                             tracing::info!(
                                 "Refreshed HTTP routers (http workspaced route setting change)"
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1843,6 +1843,7 @@ async fn process_notify_event(
                 }
                 Err(err) => {
                     tracing::error!("Error refreshing HTTP routers (trigger change): {err:#}");
+                    windmill_api::triggers::http::invalidate_routers_version().await;
                     return false;
                 }
             };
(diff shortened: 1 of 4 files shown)


hugocasa:
✅ Review round clean @ 2546d61a255c0f81e64b319b582099b6a5c6e014

PR #10851, opened 2026-08-26
fix(cli): say where a sync push deleted variable or resource went

hugocasa:
## Summary

`wmill sync push` deletes a remote variable or resource that has no local file, and says nothing more about it. Both handlers `move_to_trash` first — the item is restorable for three days from workspace settings → Trash (`migrations/20260326000000_trashbin.up.sql`, CE since v1.665.0) — but the CLI is the one surface that never mentioned it. The report behind this concluded the deletion was final and that there was nothing to restore.

A push that deleted any now ends with where they went:

```
Deleting variable f/test/v1.variable.yaml
Deleting resource f/test/r1.resource.yaml
Done! All 2 changes pushed to the remote workspace git980g (30ms)
1 variable and 1 resource deleted. Restore within three days from the workspace trash (Workspace settings -> Trash).
```

Counted by object, so a file resource's two files read as the one deletion they are.

No behaviour change: every deletion applies exactly as before.

Fixes GIT-980

## Scope

Only the notice. Two bugs this turned up in the data-table migration guard, which reads git history the same way, are split out into #10883. Separating "you deleted this on purpose" from "this was provisioned on the instance and the repo never owned it" needs the branch's git history, and being right about the answer needs object identity across a file resource's two files, workspace-specific names, quote-path handling and shallow/sparse detection — around 130 lines for a claim the trashbin already softens to a three-day undo.

So a `sync push` that removes an instance-provisioned credential is still silent about the repo never having owned it: you find out from the trashbin afterwards, not from the dry run beforehand.

## Test plan

Verified end to end against a live instance with `wmill` built from this branch:

- [x] Push deleting a variable and a resource: both deleted, notice printed, and `GET /trash/list` confirms both objects present.
- [x] File resource counted as one resource, not two files.
- [x] `--json-output` unchanged: `log.setSilent(true)` silences the notice, so the JSON the git-sync hub script parses is byte-identical.
- [x] `bunx tsc --noEmit` (only pre-existing `JSZip | TarAsZip` errors) and `UNIT_ONLY=1 bun test test/*_unit*` (1098 pass, 0 fail).


hugocasa:
Round 3 nits addressed in 8abcd96, except `--full-history`, which I'm declining — with a different reason than last round, since the design changed underneath it.

**Fixed:**
- The warning lists the offending paths instead of counting them (codex P2 #1). A push deleting a tracked *and* a never-tracked resource said "1 resource" and identified neither; now it names them. This is the one that actually mattered — the whole feature is reporting, so an unidentifiable subject defeats it.
- It no longer opens "This branch's history…" outside a git checkout, which contradicted the reason it then gave, and the pronouns that disagreed with a plural count are gone.
- The history walk is skipped under `--json-output`, where both notices are silenced and the result had no reader.
- Stale hold-back wording removed from the `describeSecretBearingChanges` doc comment and the test file header.

**Declining `--full-history`:** `gitRecordedPaths` is shared with `gitRecordedDatatableMigrationPaths`, and that caller still uses it for *prevention* — it holds data-table migration deletions back. Adding `--full-history` there would make that guard delete migrations that only ever lived on a merged side branch: a behaviour change to a feature this PR doesn't otherwise touch, in a PR whose whole claim is that it changes no behaviour. The alternative, parameterising it so the two callers disagree about what "recorded" means, is a worse thing to have to explain than the narrow inaccuracy it fixes.

The inaccuracy is real and I'm not disputing it: a path added *and* removed on a merged side branch is pruned by TREESAME, so the warning would call it never-tracked. Worth a follow-up that weighs it for the migration guard on its own terms, rather than smuggling it in here.

hugocasa:
✅ Review round clean @ 8abcd96f31c3d4b2a7bc46490eb9bab9ffd7462e — nit-only verdicts at 47060e23a1e8a878ce833d98a7ca1e162ef2daa3; nits addressed in 8abcd96 / `--full-history` dismissed in review reply

hugocasa:
@cubic-dev-ai both findings were valid and are fixed in 490cd23 — thanks, the first one was a real false positive in my own logic.

**1 (companion files):** correct, and the root cause was that the tracked set was built from the *candidates* rather than from history. A file resource whose `.resource.yaml` stays on disk while its content file is deleted produced exactly the false "never tracked" warning you describe, since the only file proving the repo owned the object wasn't among the deletions being judged.

Fixed by building the tracked set from the whole recorded history instead. That also turned the workspace-specific lookup around, which is a simplification: history is now normalized to base paths (the form the changeset already carries) via `fromWorkspaceSpecificPath`, instead of each candidate being searched for under two names.

Verified live: a file resource with only `.resource.yaml` committed, content file deleted — the deletion applies with no warning.

**2 (one line per object):** fixed, one line per `secretBearingObjectPath` with its kind. A push with a never-tracked file resource (two files) plus a never-tracked variable now reads:

```
Nothing in this branch's git history accounts for these deletions, so they were
provisioned outside this repository rather than deleted from it:
  - resource f/test/never_conf
  - variable f/test/never_var
```

4 file changes, 2 object lines. Naming the object rather than a file also happens to be the more useful form, since that is the unit the backend deletes.

New regression test covers the companion case; the workspace-specific test was inverted to match the new mapping direction.

hugocasa:
@cubic-dev-ai false positive, but a reasonable read of the commit in isolation.

The `core.quotePath` change isn't being reverted into a bug — it moved to #10883, together with the no-commits diagnosis, because both fix `gitRecordedDatatableMigrationPaths` and this PR no longer touches git history at all. `cli/src/utils/git.ts` on this branch is now byte-identical to `main`:

```
$ git diff origin/main origin/hugo/git-980-bug-wmill-sync-push-hard-deletes-variables-and-resources -- cli/src/utils/git.ts
(no output)
```

So there's no regression relative to `main` here; the fix landed in the other PR, and it got better on the way — your point about quoted paths was raised there too, and `core.quotePath=false` turned out to cover only non-ASCII. #10883 now uses `-z`, which also handles `"` and `\`.

hugocasa:
@codex fixed in 91f5a8f — the header described the change ("never said", "now prints") instead of the invariant. It now states what the tests protect: the classifier has to agree with the apply loop on which deleted files are a variable or a resource (a path the loop skips must not be counted, or the notice announces a deletion that never happened) and that the unit is the server-side object, so a file resource's two files are the one deletion they cause.

PR #10852, opened 2026-08-26
feat: make the flow graph step panel resizable and detachable

Guilhem-lm:
## Summary

The flow viewer inside the raw app editor was unusable: the graph rendered at 55% width with the Test panel glued to its right, and the step-details panel took a fixed third of what remained. This makes the step panel resizable, and below a width breakpoint moves it out of the layout entirely — the same Auto / Attached / Detached rule the flow editor already uses, off the same breakpoint.

Because the placement rule was already factored out into `flows/panelPlacement.ts` and `FlowGraphV2` already renders the placement picker, most of this is wiring the viewer into machinery that exists rather than new mechanism.

## Changes

**Raw app editor**
- The backend-runnable panel splits the viewer and the Test panel into separate full-width tabs (Flow/Script · Test · Inputs) instead of a 55/45 split. Running from the by-path toolbar switches to Test, so results appear where you are looking.

**Flow graph viewer** (`FlowGraphViewer`, shared by `/flows/get`, `FlowViewer`, `FlowStatusViewerInner` and the app/raw-app runnable panels)
- The step panel is a resizable `<Pane>` (66/34) instead of a fixed `grid-cols-3` cell.
- Below `MODAL_PANEL_BREAKPOINT` (1280) it detaches: the graph takes the full width and the panel opens in a dialog on double-click, or on a click on the already-selected step. Same gestures and the same standing hint as the editor.
- Width is measured on the component, not the window — the viewer is routinely embedded in a pane far narrower than the screen.
- Provides the `flowPanelDetach` context, which is what surfaces the Auto / Attached / Detached picker in the graph's control bar. The picker is also in the dialog header, since the control bar is behind it once detached.
- The `sm`-breakpoint gate on the step panel is gone — a `Pane` keeps its width even when its content is `display: none`, so the breakpoint now decides whether the pane exists.

**Supporting**
- `flows/stepLabel.ts`: the step's display name, previously a 25-line `{#if}` chain inside `FlowGraphViewerStep`'s markup, extracted so the panel and the dialog title can't disagree. The `failure` / `preprocessor` id checks stay between the composite types and the leaf script types, so an error handler reads "Error handler" and not "Inline bun script".
- `FlowGraphViewerStep`: `class` and `hideHeader` props, so a host that pads, scrolls and names the panel itself doesn't get it twice. Defaults are unchanged for the other four callers.
- `Modal`: `titleBadgeFirst`, for a badge that identifies the subject (a step id reads before the name) rather than qualifying it (a "Beta" tag reads after). The plain header row also gets the `pr-8` that the breadcrumb row already had under `kind="X"`, so a long title or anything in `settings` stops running under the close button.
- `FlowGraphV2`: the control bar carries the opaque surface. `surface-hover` is a translucent token, so a hovered button that only swapped its own background composited that tint straight onto the graph.
- `/dev/flow_path_viewer`: a dev harness rendering `FlowPathViewer` against a seeded fixture flow, with toggles for `noSide`, `fillAvailableHeight` and the container size, for iterating on the viewer without hunting for a flow that exercises every node type.

## Screenshots

Attached — resizable pane, picker in the control bar:

attached

Detached below the breakpoint — graph full-width, standing hint bottom-left:

detached

The step dialog, titled by id and summary:

dialog

## Test plan

- [ ] `/dev/flow_path_viewer` — drag the splitter; the graph re-lays out and the inputs table stops clipping its `Format` / `Required` columns.
- [ ] Same page, "Raw-app pane (55%)" — the panel detaches; double-click a step, then Escape, then single-click the selected step to reopen.
- [ ] Widen the window past 1280 with the dialog open — it closes and the pane comes back.
- [ ] Control bar picker → Detached with a step selected: the dialog opens on that step, not empty. Attached from the dialog header puts it back.
- [ ] Hover any control-bar button: the hover shade is opaque, not a window onto the graph.
- [ ] Raw app editor → a workspace-flow backend runnable: Flow / Test / Inputs are separate full-width tabs; Run from the toolbar lands on Test.
- [ ] `/flows/get/<path>` and the flow editor's subflow step still render.

## Not covered

- Two `FlowPathViewer` drawers (`ScriptPicker`, the runnable toolbar's Expand) are 900px and 1200px, both under the breakpoint, so they now open detached. Not exercised.
- No `feature_usage` instrumentation. The editor's equivalent has `useFlowPanelPlacementTelemetry`; adding it here needs a backend allowlist entry and disclosure copy, so it is left out deliberately.

## Review round 1 fixes

- **The graph is no longer re-created when the panel moves.** It was rendered from two `{#if}` branches, so every docked↔modal transition destroyed and rebuilt `FlowGraphV2` — losing pan, zoom and the selected node — and every embed under 1280 mounted it twice, since width starts at 0 and 0 resolves to docked. It now lives in one persistent `Pane` sized 100 or 66, with only the step pane conditional, matching `FlowEditor.svelte:343`. Verified by tagging the `.svelte-flow` element and watching it survive both transitions.
- **`FlowModuleComponent.svelte:1080` now passes `noSide`.** The flow editor's own subflow step was detaching and shadowing the editor's `flowPanelDetach` context — the earlier "Not covered" note wrongly claimed the nested case was unaffected, which was only true of `FlowGraphViewerStep.svelte:314`.
- **`wasVisible` no longer counts a pane that renders nothing.** Under `hideDefaultInputs` with no selection, Detached opened an empty dialog.
- **The step dialog is actually 896px wide.** `max-w-4xl` lost the cascade to `Modal`'s own `sm:max-w-lg`; `sm:max-w-4xl` wins it.
- **The dev harness can't clobber a real flow.** It stamps every flow it deploys and refuses to overwrite one it didn't create, and the whole route is behind `import.meta.env.DEV`.
- **`stepLabel.test.ts`** pins the ordering the extraction had to preserve — summary first, `failure`/`preprocessor` between the composite and leaf types, and the parenthesised suffixes.

The step dialog at its intended width:

dialog

https://claude.ai/code/session_016eu85JrwsztjfXGNQAATRz

### DIFF
--- a/frontend/src/lib/components/FlowGraphViewer.svelte
+++ b/frontend/src/lib/components/FlowGraphViewer.svelte
@@ -7,6 +7,13 @@
 	import { writable } from 'svelte/store'
 	import { twMerge } from 'tailwind-merge'
 	import { Pane, Splitpanes } from 'svelte-splitpanes'
+	import { MousePointerClick } from 'lucide-svelte'
+	import Modal from './common/modal/Modal.svelte'
+	import Badge from './common/badge/Badge.svelte'
+	import FlowPanelPlacementPicker from './flows/common/FlowPanelPlacementPicker.svelte'
+	import { useFlowPanelMode } from './flows/flowPanelMode.svelte'
+	import type { FlowPanelDetachContext } from './flows/types'
+	import { stepLabel } from './flows/stepLabel'
 
 	import FlowGraphViewerStep from './FlowGraphViewerStep.svelte'
 	import FlowGraphV2 from './graph/FlowGraphV2.svelte'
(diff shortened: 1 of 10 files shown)


Guilhem-lm:
✅ Review round clean @ 80d6fb7eb358cc1a90868e3a0e40f1c2ef804445 — nit-only verdicts at 80d6fb7eb358cc1a90868e3a0e40f1c2ef804445 (codex + pi "Good to merge", claude nit-only); nit addressed in 259b878751a1cab192b7eaab8a3dfc37e5f3f6ec

PR #10856, opened 2026-08-26
fix: nested template literals in step inputs, and unresolvable $args tags

hugocasa:
## Summary

Two independent bugs that meet in the same failure: a flow step whose input is built with a nested template literal, routed by a dynamic worker tag.

**1. Template mode corrupted nested template literals.** Converting a template-mode input value into a JS expression escaped *every* backtick, including ones inside `${...}`, where a backslash is a syntax error. Typing

```
--input ${flow_input.iter.value}${results.config ? ` --config ${results.config}` : ""} --host ${flow_input.hostname}
```

stored an escaped backtick for each inner one, and the step failed with `QuickJS evaluation error: unexpected token in expression: '\'`. The identical text typed in javascript mode worked.

**2. That failure was then invisible.** When a step's input transforms fail, `push_args` becomes empty, so a step tagged `worker-$args[hostname]` interpolated against nothing and was queued on a tag no worker serves. The step never ran, never failed, and never timed out — the expression error was never reported. The step just sat there.

## What changes for existing setups

Nothing that worked before behaves differently. Every change below only affects a job that was already broken — an expression that didn't parse, or a job nobody could pull — with one class of exception, from the `pre_run_error` dispatch guard.

Neither hand-off path ever read `pre_run_error`, so a job whose input transforms failed used to **run with the empty arguments left by that failure and report success** — measured, see the test plan. Three kinds of job were reaching those paths and now fail instead: a step of a dedicated flow, a step on a dedicated runnable (exempt from the reroute, but it still arrived on its dedicated tag), and a squashed for-loop iteration on flow runners — with `skip_failures` on, that iteration could report success. All three are strictly better failing than running on inputs that never evaluated. Note that when such an iteration fails, the loop's runners are dropped (`handle_all_job_kind_error` completes with `flow_runners: None`) and the rest of the loop falls back to ordinary queued jobs.

Tag resolution itself is untouched. A tag that resolves in part or not at all is queued exactly as it is today: `gpu$args[suffix]` with no `suffix` still means the `gpu` queue, `worker-$args[hostname]` with no `hostname` still becomes `worker-` and still sits unpulled. Whether a leftover literal is a real queue depends on which tags workers advertise, which `push` cannot know.

### Frontend

- `escapeTemplateBackticks` / `unescapeTemplateBackticks` (`frontend/src/lib/utils/templateLiteral.ts`) ask acorn whether the text already reads as one template literal. If it does it needs no escaping at all, which is what makes a nested template work; if it does not, every backtick is escaped, which is the previous behaviour. Requiring a single `TemplateLiteral` spanning the whole source is what rejects a body that closes its own literal early, so the author's literal text can never become evaluated code. Replaces the three blanket backtick-escaping sites and the inverse in `codeToStaticTemplate`.
- Because acceptance requires every backtick to sit inside a `${...}`, and blanket-escaping such a text always produced an escaped backtick in expression position (a syntax error), **the set of inputs that behave differently is exactly the set the old code broke** — no previously-working expression can change meaning.
- `unescapeTemplateBackticks` also heals an expression escaped before nested templates were handled, instead of showing its backslashes and adding one more on every save.
- Running a step test now reports an input expression that failed to evaluate, instead of silently sending `undefined`. The reactive re-evaluations that fire on every flow edit stay quiet.
- Adds `acorn` (MIT) as a frontend dependency. Measured on builds either side: the entry chunk is unchanged at 22.1 KB gzip, total app JS goes 13.62 → 13.66 MB gzip, and acorn is isolated in its own 32.6 KB chunk reachable only from the flow-editor and app routes. A fast path skips the parser entirely for any value with no backtick in it, which is every ordinary value.

### Backend

- A flow step whose input transforms failed is pushed on the **parent flow's tag** when its own tag reads `$args[...]`. Such a step is already going to fail, but it still has to reach a worker to be marked failed, and `push_args` is empty on that path — so its own tag can only interpolate to a queue nobody serves, whatever the arguments were meant to be. The flow's tag is provably served: a worker is running the flow on it right now. The step then fails like any step: the module shows `Failure`, the flow's error handler runs, and the reported error is the expression error rather than anything about tags.
- Two kinds of step are excluded, because their tag never reaches the queue and rewriting it would be noise: a `same_worker` (or flow-runner) step, handed over by job id (`get_same_worker_job` has no tag predicate), and a step on a dedicated *runnable*, whose tag `push` replaces with the dedicated tag — read through a new `JobPayload::is_dedicated_worker()` rather than by duplicating `push`'s logic.
- **A job carrying a `pre_run_error` is no longer handed to a dedicated runner or a flow runner.** Both dispatch by path and return before the `pre_run_error` check at `worker.rs:4487`, so such a job used to run with the arguments that survived the failure. This is pre-existing, but the reroute above would have routed one more case into it, so the guard is what keeps the reroute safe rather than a drive-by fix. See *What changes for existing setups* for everything it covers.

## Screenshots

Step input in javascript mode, showing the expression generated from the same template text.

Before — inner backticks escaped, expression no longer parses:

before

After:

after

## Test plan

Exercised against a local instance; results in the table below.

- [x] Type a nested template literal into a step input in template mode, switch to javascript mode: no stray backslashes; switch back: the text round-trips unchanged
- [x] Flow step with a failing input expression and a tag reading `$args[...]`: step `Failure`, error handler runs, reporting the expression error
- [x] An invalid step input expression surfaces its error from the test panel's Run button, verified in the browser (it was silently `undefined` before on every entry point but the graph button)
- [x] The same step with a literal tag no worker serves: still queues, unchanged
- [x] The same step in a `same_worker` flow: still runs green
- [x] A tag that resolves to nothing with the args intact, on a flow step and on a direct push: untouched, queues exactly as on main
- [x] `gpu$args[suffix]` with no `suffix`: still tagged `gpu`; `worker-$args[hostname]` with no `hostname`: still tagged `worker-`; `hostname: "host1"`: still tagged `host1`
- [x] Differential of the old blanket escaping against the new implementation over 30 template values and 33 stored expressions: they differ only where the old output was unparseable JS, and no previously-parseable stored expression breaks on a view/save cycle

| case | before | after |
|---|---|---|
| bad expression + tag reading `$args[...]` | queued forever, error never reported | step `Failure`, handler ran, expression error shown |
| bad expression + literal tag no worker serves | queued forever | queued forever (unchanged) |
| bad expression, `same_worker` step | ran fine | ran fine (unchanged) |
| bad expression, step of a dedicated flow / dedicated runnable / squashed loop iteration | **reported `success`** with the argument `undefined` | fails with the expression error |
| tag resolving to nothing, args intact | queued forever | queued forever (unchanged) |
| `gpu$args[suffix]` / `worker-$args[hostname]` / resolvable tag | as configured | as configured (unchanged) |

The dedicated-flow row was measured too, on an EE build (`--features quickjs,enterprise,private`) run as `MODE=worker DEDICATED_WORKER=<ws>:flow/<path>` against the same database, with a `skip_failures` loop whose step input expression fails. Same flow, same worker, guard reverted and restored:

| | status | result |
|---|---|---|
| without the guard | `success` ×3 | `"ran with undefined"` |
| with the guard | `failure` ×3 | `could not transform input using an expr: …` |

So the step did not merely run with empty arguments — it ran and was recorded **green**. That is the one change an existing setup can notice: a dedicated flow that has been quietly succeeding on inputs that never evaluated will start failing.

Checks: `npm run check` 0 errors, vitest green, `cargo check --features enterprise,private` clean. No SQL changed.


hugocasa:
Re the mixed literal+nested backtick nit (`templateLiteral.ts`): confirmed real, but it is not a regression — the old blanket escaping produces byte-identical output for `literal ` text ${flag ? `nested` : ""}`, so this PR neither introduces nor worsens it. Fixing it properly needs per-interpolation span detection (locating each `${...}` end with the parser and escaping only the literal spans), which is a larger change than this PR should carry. Documented as a known limitation on the function, along with the workaround: escaping the literal backtick by hand makes the whole value parse, and it is then kept verbatim.

The two nits from the Claude review are fixed in aaf3a24: a value corrupted before nested templates were handled now heals through `unescapeTemplateBackticks` instead of compounding a backslash per edit, and the comments no longer describe the removed walker.

hugocasa:
cubic's P1 is confirmed and fixed in 968acfb. Measured before the change: stored `${"a\\\\`"}` evaluates to `"a\\`"` but came back from the editor as `${"a\\`"}`, which evaluates to `"a`"` — the author's backslash dropped on a view/save cycle.

Adopting the suggested restriction (heal only when the stored text does not already parse) gives 0 meaning changes across the corpus, against 1 for the previous rule, and lets the textual `everyBacktickEscaped` check go away entirely.

It does give up two earlier *cosmetic* requests, which is the deliberate trade: a legacy value that stayed valid under the old rule (`${"\\`"}`) and one holding an authored backslash now display their backslash rather than being cleaned up. Both still evaluate correctly and round-trip stably; the two forms are indistinguishable from the text alone, so cleaning them up means guessing, and guessing is what dropped a real character. Correctness over display.

cubic's two P2s, for the record:

- `templateLiteral.ts:38` — a static value with a backslash immediately before a backtick loses a character. Real, but pre-existing and not a regression: the old rule stored `a\\`` (dropping the backtick), the new one stores `a\`` (dropping the backslash). Fixing it properly means escaping backslashes too, which changes the stored format for every existing value. Documented on the function alongside the mixed literal+nested limitation.
- `ModuleTest.svelte:65` — blocking the run when a transform fails rather than warning is a UX decision rather than a defect, so I've left it for the author to call.

hugocasa:
Not taking cubic's remaining two P2s. Reasoning, with the measurement behind it.

**Priority for this change is that no expression that evaluates today evaluates differently.** Differential over 196 template values (every pairwise combination of 14 fragments: backticks, backslashes, nested templates, regex literals, comments, quotes, escaped `\${`), comparing the old blanket escaping against the shipped implementation by *evaluated result* rather than by string:

| escape direction | |
|---|---|
| old valid, result unchanged | 152 |
| old broken, now works | 44 |
| old valid, result changed | **0** |

View/save cycle over every expression either rule could have stored: **0** change meaning.

**`templateLiteral.ts:38` — backslash immediately before a backtick in a static value.** Not a regression. Measured: the value `a\\`` stored under the old rule is `a\\\\``, which does not parse at all — the field was a syntax error. The new rule stores `a\\`` and yields `"a`"`, consuming the backslash as a JS escape. So this shape sits in the 44 that the old code broke and this one fixes, not in the changed set. Making it preserve the backslash means escaping backslashes as well, which changes the stored format for every existing value — exactly the kind of change the priority above rules out. Documented as a limitation on the function, next to the mixed literal+nested one.

Worth noting how narrowly this can be reached at all: `escapeTemplateBackticks` sees a static value only at `InputTransformForm.svelte:252`, gated on `isCodeInjection` so the value is already becoming an expression, or at `:691` when the author explicitly switches to javascript mode. A value with no backtick returns unchanged on the first line. A static value that stays static is never touched.

**`ModuleTest.svelte:65` — skip the run when a transform fails.** A UX decision rather than a defect, and out of scope here. Running still surfaces the failure (the toast names the argument and the parse error), and a real flow run now fails such a step server-side regardless. Happy to see it as its own change if the maintainers want the stricter behaviour.

PR #10862, opened 2026-08-26
fix: keep the thrown value identifiable in the bun and deno error wrappers

rubenfiszel:
## Summary

The Bun wrapper that turns a thrown value into a job error was not defensive about what a script actually threw:

- `throw null` crashed it on `e.message`.
- A thrown object with a reference cycle crashed the `JSON.stringify` that writes `result.json`. This is the practically important case: several widely used HTTP clients throw errors with circular references (an axios error hangs `response` off `request` and back), so any adapter that lets one propagate lost it entirely.
- `throw "a string"` survived, but was spread into a character index map (`{"0":"n","1":"u",...}`).

The blast radius is error *reporting*, not correctness: the job still fails, nothing is lost or corrupted, no worker dies. What is lost is why it failed, since the result carries a stack trace from Windmill's own wrapper instead of the script's error. That makes the real cause unrecoverable from run history and leaves a flow error handler unable to branch on the error class.

The Deno wrappers share the same unguarded `e.message` read, so `throw null` crashed them too, and a thrown non-object produced an empty `{}`.

Fixes GIT-984.

## Changes

A shared `JS_ERROR_SERIALIZER` const in `backend/windmill-worker/src/common.rs`, injected into all five generated wrapper templates (Bun main, Bun WAC v2, Bun dedicated worker / runner group, Deno main, Deno dedicated worker):

- `wmToErrorObject` reads `message` / `name` / `stack` behind a guard, so a non-object throw cannot crash the wrapper, and it no longer spreads a non-object's indices into `extra`. A thrown non-object becomes `{ message: <string form>, name: "ThrownValue" }`, so `throw null` reports `null` and `throw "boom"` reports `boom` verbatim, at any length.
- `wmSerializeError` attempts the plain `JSON.stringify` the wrappers already did, and only if that throws retries with `extra` replaced by `"[extra omitted: not serializable]"`, falling back to the thrown value's own string fields. So the output is byte-identical wherever the old wrapper produced one, and the job reports the script's error instead of the wrapper's stack wherever it did not.
- A thrown object exposing none of `message` / `name` / `stack` reports `Object.prototype.toString.call(e)`, a type tag with no contents, where Deno previously stored a bare `{}`. It fires whether or not `extra` is present, so a cyclic object thrown on Bun still carries an identifying field rather than the marker alone.
- `extra` skips properties that read back `undefined`, so a thrown object whose getters throw still reports an identifiable error rather than an empty `{"extra":{}}`.
- `collectExtra` is true only for the two wrappers that already collected the thrown value's own properties: the Bun main wrapper and Bun WAC v2. The Bun dedicated worker / runner group and both Deno wrappers emitted `{ message, name, stack }` and keep doing so. Giving them `extra` would newly persist whatever an HTTP client hangs off its errors, credential-bearing headers included, into the run history of every existing script on those paths. That parity is a separate decision, not part of stopping the wrapper from crashing.

A thrown `Error`, and a thrown object carrying its own `message`, are reported exactly as before. What changes is only what used to be unreportable: a value that crashed the wrapper now yields the fields it does expose, and one exposing none of them gains the type tag rather than an empty object.

The helper adds ~2 KB to the per-job `wrapper.mjs` in the job's temp directory, and nothing else. It does not reach the cached bundle: that is built from `main.ts` alone (`bun_executor.rs:1003`) and the wrapper imports it, which is why `bun_executor.rs:2194` rewrites the wrapper's import to `./main.js` afterwards. Verified by grepping the on-disk bundle cache. The helpers are function declarations that only run inside `catch`, so the success path is unaffected; parse cost measured against bun startup is below the noise floor.

## Test plan

- [x] `cargo test -p windmill --test bun_jobs --test worker` (68 tests, all pass) plus `wac_approval_resume` / `wac_approval_urls`. Two regression tests, `test_bun_job_non_error_throws` and `test_deno_job_non_error_throws`, pin the throws each runtime actually got wrong, including the Deno no-`extra` invariant, a real-shaped cyclic `AxiosError` carrying sentinels in a header, in `config.auth` and in an mTLS private key, and on Deno a `throw { cause: axiosErr }` wrapper, none of which may appear in the result.
- [x] Ran real jobs against this worktree's backend for `throw null`, a thrown string, a thrown plain object, `throw { cause: axiosErr }`, a cyclic `AxiosError`, a thrown `Error` and a successful return, on **bun** and on **deno**, plus the cyclic cases through a **WAC v2** workflow. Every case reports the script's own value; none reports the wrapper, and the success path is unchanged. The sentinel credential appears zero times in every stored result, on all three paths.
- [x] Generated the **dedicated worker** wrappers (Bun and Deno) from the real codegen functions and drove them over their `execd:` line protocol for the same cases: each `wm_res[error]:` line stays a single line and carries the thrown value.
- [x] Ran the pre-fix wrapper logic and the new one side by side, in both `collectExtra` modes, asserting `extra` parity on every acyclic shape and no sentinel on every previously-crashing one: plain objects, errors with scalar and with nested-acyclic properties, the cyclic `AxiosError`, `throw { cause: axiosErr }`, a cyclic secret-bearing plain object, an array cycle, a BigInt property, a function property, and a key-sensitive `toJSON`. Also the awkward primitives: symbol, bigint, `Object.create(null)`, a getter that throws, a `toString` that throws. All three runtimes agree byte for byte.

Not exercised: the `dedicated_workers` integration suite, which needs `--features private,enterprise`. The local EE worktree sits one commit ahead of the `backend/ee-repo-ref.txt` pin, and that extra commit calls a `windmill-common` helper this branch does not have, so the EE build fails before reaching any test. That is local worktree state, unrelated to this change, which touches no EE file. The two dedicated-worker wrappers were verified instead by generating them from their real codegen functions and driving the line protocol directly, as above.

## On the credential findings

Four review rounds each found a different credential reachable once cyclic graphs serialize: an `Authorization` header, then structured `config.auth = { username, password }`, then an mTLS private key under `config.httpsAgent.options.key`, then a `wmErrDescribe` helper of mine rendering a `throw { cause: err }` wrapper's whole contents into `message` on the paths that collect no `extra`. The first two I answered by extending a blocklist of key names. That was the wrong shape: a blocklist over a library's internal object graph can never be complete, and each round produced the next key name.

The blocklist, the size cap and `wmErrDescribe` are all gone, replaced by a condition that is checkable rather than enumerable:

> A thrown object's contents reach the result only through the one plain `JSON.stringify` the wrappers already did.

That makes the exposure question decidable instead of a matter of coverage, and it closes both routes that could carry contents. Where the old wrapper produced a result, the output is byte-identical. Where it crashed, the job reports the thrown value's own `message` / `name` / `stack` plus the marker. A thrown object with none of those fields reports a type tag. So this PR persists nothing that `main` did not, for any thrown value, whatever its keys are named.

I checked that as a property rather than by example, running the pre-fix wrapper logic and the new one side by side in both `collectExtra` modes over: plain objects, errors with scalar and with nested-acyclic properties, a real-shaped cyclic `AxiosError` carrying sentinels in a header, in `config.auth` and in an mTLS key, a `throw { cause: axiosErr }` wrapper, a cyclic secret-bearing plain object, an array cycle, a BigInt property, a function property, and a key-sensitive `toJSON` returning a different graph for the `"extra"` key than for the root. Every acyclic case is byte-identical; every previously-crashing case yields the marker; the sentinel count is zero throughout, under bun, node and deno alike.

Two notes behind that:

- `wmSerializeError` does not probe and then re-serialize. A probe in a separate context can disagree with the real serialization, because `toJSON` receives the key it is being serialized under.
- The pre-existing exposure is untouched, deliberately: an *acyclic* HTTP error already serialized `config` / `request` / `response` into `extra` on `main`, credentials included. Redacting that applies to results, logs and returned values alike, changes what existing scripts read out of `extra`, and wants its own change rather than the blocklist that failed to converge here.

The cost is that an axios error no longer carries `extra.response.status`. That is parity with `main`, which carried nothing at all for those errors, and `message` / `name` / `stack` still identify the failure. A bounded, cycle-safe projection of transport errors would be a genuine improvement, but it is an enrichment rather than part of stopping a crash.

## Left in draft

This regenerates the wrapper for every Bun and Deno job, so the blast radius is wider than the two-file diff suggests. Per the repo's PR policy that is a change to shared worker infrastructure and wants a human look before it goes ready, so it stays a clean draft rather than flipping on a green round.

### DIFF
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -414,9 +414,9 @@ async fn test_bun_job_non_error_throws(db: Pool<Postgres>) -> anyhow::Result<()>
     }
 
     // An HTTP client links its errors both ways, which is what a plain JSON.stringify
-    // could not serialize. `extra` is reported only when it serializes on its own, so the
-    // request the client hangs off the error, credentials included, stays out of the
-    // result exactly as it did when the wrapper crashed.
+    // could not serialize. Only that same stringify reports `extra`, so the request the
+    // client hangs off the error, credentials included, stays out of the result exactly as
+    // it did when the wrapper crashed.
     {
         let job = bun_job(
             r#"
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -436,6 +436,40 @@ export function main() {
         );
     }
 
+    // Serializing a cycle is what first brings an HTTP client's whole request into
+    // `extra`, so the transport credentials on it must not reach the stored result.
+    {
+        let job = bun_job(
+            r#"
+export function main() {
+    const config: any = { url: "/x", headers: { Authorization: "Bearer sentinel-tkn" } };
+    const request: any = { path: "/x", _header: "GET /x HTTP/1.1\r\nAuthorization: Bearer sentinel-tkn\r\n\r\n" };
+    const response: any = { status: 401, config, request };
+    request.res = response;
+    const e: any = new Error("Request failed with status code 401");
+    e.name = "AxiosError";
+    e.config = config;
+    e.request = request;
+    e.response = response;
+    throw e;
+}
+"#,
+        );
+        let completed = run_job_in_new_worker_until_complete(&db, false, job, port).await;
+        assert!(!completed.success);
+        let result = completed.json_result().unwrap();
+        let error = &result["error"];
+        assert_eq!(
+            error["message"],
+            serde_json::json!("Request failed with status code 401")
+        );
+        assert_eq!(error["extra"]["response"]["status"], serde_json::json!(401));
+        assert!(
+            !error.to_string().contains("sentinel-tkn"),
+            "credential reached the result: {error}"
+        );
+    }
+
     Ok(())
 }
 
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -442,7 +442,11 @@ export function main() {
         let job = bun_job(
             r#"
 export function main() {
-    const config: any = { url: "/x", headers: { Authorization: "Bearer sentinel-tkn" } };
+    const config: any = {
+        url: "/x",
+        headers: { Authorization: "Bearer sentinel-tkn" },
+        auth: { username: "svc", password: "sentinel-tkn" },
+    };
     const request: any = { path: "/x", _header: "GET /x HTTP/1.1\r\nAuthorization: Bearer sentinel-tkn\r\n\r\n" };
     const response: any = { status: 401, config, request };
     request.res = response;
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -401,29 +401,6 @@ async fn test_bun_job_non_error_throws(db: Pool<Postgres>) -> anyhow::Result<()>
         assert_eq!(result["error"]["message"], serde_json::json!("null"));
     }
 
-    // Circular object: several HTTP clients throw errors that link two objects
-    // both ways, which a plain JSON.stringify cannot serialize.
-    {
-        let job = bun_job(
-            r#"
-export function main() {
-    const a: any = { name: "circular", message: "boom" };
-    a.self = a;
-    throw a;
-}
-"#,
-        );
-        let completed = run_job_in_new_worker_until_complete(&db, false, job, port).await;
-        assert!(!completed.success);
-        let result = completed.json_result().unwrap();
-        let error = &result["error"];
-        assert_eq!(error["message"], serde_json::json!("boom"));
-        assert_eq!(
-            error["extra"]["self"]["self"],
-            serde_json::json!("[Circular]")
-        );
-    }
-
     // A thrown string is a string, not an index map of its characters.
     {
         let job = bun_job(r#"export function main() { throw "nur ein string"; }"#);
@@ -436,8 +413,10 @@ export function main() {
         );
     }
 
-    // Serializing a cycle is what first brings an HTTP client's whole request into
-    // `extra`, so the transport credentials on it must not reach the stored result.
+    // An HTTP client links its errors both ways, which is what a plain JSON.stringify
+    // could not serialize. `extra` is reported only when it serializes on its own, so the
+    // request the client hangs off the error, credentials included, stays out of the
+    // result exactly as it did when the wrapper crashed.
     {
         let job = bun_job(
             r#"
@@ -446,6 +425,7 @@ export function main() {
         url: "/x",
         headers: { Authorization: "Bearer sentinel-tkn" },
         auth: { username: "svc", password: "sentinel-tkn" },
+        httpsAgent: { options: { key: "-----BEGIN PRIVATE KEY-----sentinel-tkn" } },
     };
     const request: any = { path: "/x", _header: "GET /x HTTP/1.1\r\nAuthorization: Bearer sentinel-tkn\r\n\r\n" };
     const response: any = { status: 401, config, request };
@@ -467,10 +447,26 @@ export function main() {
             error["message"],
             serde_json::json!("Request failed with status code 401")
         );
-        assert_eq!(error["extra"]["response"]["status"], serde_json::json!(401));
+        assert_eq!(error["name"], serde_json::json!("AxiosError"));
+        assert_eq!(
+            error["extra"],
+            serde_json::json!("[extra omitted: not serializable]")
+        );
         assert!(
             !error.to_string().contains("sentinel-tkn"),
             "credential reached the result: {error}"
+        );
+    }
+
+    // An acyclic thrown object still reports its own properties, unchanged.
+    {
+        let job = bun_job(r#"export function main() { throw { code: 42, hint: "kein Error" }; }"#);
+        let completed = run_job_in_new_worker_until_complete(&db, false, job, port).await;
+        assert!(!completed.success);
+        let result = completed.json_result().unwrap();
+        assert_eq!(
+            result["error"]["extra"],
+            serde_json::json!({ "code": 42, "hint": "kein Error" })
         );
     }
 
--- a/backend/tests/bun_jobs.rs
+++ b/backend/tests/bun_jobs.rs
@@ -364,6 +364,81 @@ export function main() {
     Ok(())
 }
 
+/// A script can throw a non-object, or an object with a reference cycle. The error the
+/// job reports must still identify what was thrown, never the wrapper's own failure to
+/// read or serialize it.
+#[sqlx::test(fixtures("base"))]
+async fn test_bun_job_non_error_throws(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+    let server = ApiServer::start(db.clone()).await?;
+    let port = server.addr.port();
+
+    fn bun_job(content: &str) -> JobPayload {
+        JobPayload::Code(RawCode {
+            hash: None,
+            content: content.to_owned(),
+            path: None,
+            language: ScriptLang::Bun,
+            lock: None,
+            concurrency_settings: windmill_common::runnable_settings::ConcurrencySettings::default(
+            )
+            .into(),
+            debouncing_settings: windmill_common::runnable_settings::DebouncingSettings::default(),
+            cache_ttl: None,
+            cache_ignore_s3_path: None,
+            dedicated_worker: None,
+            modules: None,
+            tag: None,
+        })
+    }
+
+    // `throw null`: the wrapper must not dereference the thrown value.
+    {
+        let job = bun_job("export function main() { throw null; }");
+        let completed = run_job_in_new_worker_until_complete(&db, false, job, port).await;
+        assert!(!completed.success);
+        let result = completed.json_result().unwrap();
+        assert_eq!(result["error"]["message"], serde_json::json!("null"));
+    }
+
+    // Circular object: several HTTP clients throw errors that link two objects
+    // both ways, which a plain JSON.stringify cannot serialize.
+    {
+        let job = bun_job(
+            r#"
+export function main() {
+    const a: any = { name: "circular", message: "boom" };
+    a.self = a;
+    throw a;
+}
+"#,
+        );
+        let completed = run_job_in_new_worker_until_complete(&db, false, job, port).await;
+        assert!(!completed.success);
+        let result = completed.json_result().unwrap();
+        let error = &result["error"];
+        assert_eq!(error["message"], serde_json::json!("boom"));
+        assert_eq!(
+            error["extra"]["self"]["self"],
+            serde_json::json!("[Circular]")
+        );
+    }
+
+    // A thrown string is a string, not an index map of its characters.
+    {
+        let job = bun_job(r#"export function main() { throw "nur ein string"; }"#);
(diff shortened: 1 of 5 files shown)


rubenfiszel:
Pushing back on the P1 from the last Codex round ("Bun circular errors can persist credentials"), because its premise does not hold.

The finding assumes that when serialization failed, the credential was not persisted. It was. `console.error(e)` is the first statement of the Bun catch block, it runs before any serialization, and this PR does not touch it. I reproduced the pre-fix wrapper verbatim (old `let err = { message: e.message, ... }` + `Object.getOwnPropertyNames` + bare `JSON.stringify`) against an axios-shaped error carrying `Authorization: Bearer WMSENTINEL123`:

```
exit=1
sentinel in STDERR (job logs) on the OLD wrapper: 1 occurrence
serialization: crashed on JSON.stringify
result.json: not written
```

So on `main` today that header is already in the job logs, under the same job, the same readers and the same retention. What this PR changes is that the same already-logged data also reaches the `result` field. That is not a new exposure boundary.

Two further points:

1. Bun `extra` collection is untouched by this PR. It already serialized `config`/`request`/`response` in full for every non-circular thrown object. The PR adds no key to `extra` and removes none.
2. The remedies suggested (redact sensitive fields, or disable `extra`) both reach well past this bug. Disabling `extra` on the Bun main path would be a behavior regression for anyone reading `error.extra` today, and a key-name blocklist is incomplete by construction, so it would give false assurance. Redaction of job results is a real topic, but it applies to results, logs and returned values alike, and belongs in its own change rather than inside a crash fix.

For the record, the earlier rounds did find two real regressions I had introduced, and both are fixed: I had wrongly turned on `extra` for the Deno wrappers and then for the Bun dedicated worker / runner group, neither of which ever reported it. Every wrapper now keeps its exact previous field set, and `test_deno_job_non_error_throws` pins the no-`extra` invariant with a sentinel authorization property.

Leaving this as a draft for a human call on the point above.

rubenfiszel:
Superseding my earlier pushback on the credential P1: the finding is right, and the fix is in `f8d691e714`.

The pushback claimed `console.error(e)` already logs the credential pre-fix, so the crash was not what kept it out. Measuring rather than asserting it, that is only half true. Bun's inspect is depth-limited, so a real `AxiosError` logs:

```
config: { url: "/x", method: "get", headers: [Object ...] },
request: { path: "/x", _header: "GET /x HTTP/1.1\r\nAuthorization: Bearer <secret>\r\n\r\n", ... },
```

The flat `request._header` string does reach the log, which is the occurrence my earlier sentinel test hit. The structured `config.headers.Authorization` does not: it is masked as `[Object ...]`. So for the structured credential the crash genuinely was the only thing keeping it out of a stored result, and serializing cycles does newly persist it. My earlier rebuttal generalised from the one shape that happened to sit at depth 1.

Both bounds live in the replacer that already existed for cycles:

- transport-credential keys (`authorization`, `proxy-authorization`, `cookie`, `set-cookie`, `x-api-key`, `api[-_]?key`, `_header`) serialize as `"[redacted]"` at any depth;
- `extra` past 100 KB serialized becomes `"[extra omitted: too large]"`, so a `socket`/`agent`/buffer graph cannot produce a multi-MB result.

I deliberately did not extend the list to `token`/`password`/`secret`. Those are ordinary field names in user data, and redacting them changes what existing scripts read out of `extra` for a broader policy call that wants its own change rather than riding along on a crash fix. Happy to widen it if you would rather.

The useful error is untouched: the axios case still reports `Request failed with status code 401` with `extra.response.status === 401`. Verified on real bun, deno and WAC v2 jobs, sentinel absent from all three stored results, and pinned by a new case in `test_bun_job_non_error_throws`.

rubenfiszel:
Reworked the fix in `21bba1b762` in response to this round's mTLS finding.

Three rounds each surfaced a different credential reachable through `extra` once cyclic graphs serialize: an `Authorization` header, then structured `config.auth = { username, password }`, then `config.httpsAgent.options.key`. I answered the first two by extending a blocklist of key names. The third made it clear that was the wrong shape, and this review's own suggestion names the alternative ("or avoid serializing transport-client internals"): a blocklist over a library's internal object graph cannot be complete, and each round would have produced the next key name.

The blocklist and the size cap are gone, replaced by a condition that is checkable rather than enumerable:

> `extra` is reported only when it serializes on its own, which is exactly when the pre-existing bare `JSON.stringify` would have succeeded.

Where the old wrapper produced a result, `extra` is byte-identical to what it produced. Where it crashed, `extra` is `"[extra omitted: not serializable]"` and the job reports `message`/`name`/`stack` instead of the wrapper's own stack, which is the bug. So the PR persists nothing `main` did not, for any thrown value, whatever its keys are named — no key list to audit.

I checked that as a property, not by example: ran the pre-fix wrapper logic and the new one side by side over plain objects, errors with scalar and nested-acyclic props, a real-shaped cyclic `AxiosError` with sentinels in a header, in `config.auth` and in an mTLS key, a self-referential object, an array cycle, a BigInt prop and a function prop. Every acyclic case identical; every previously-crashing case yields the marker; sentinel count zero.

Also resolves the nits from the same round: the "walk is capped" wording (no cap now), the doc paragraph that argued with the review thread rather than recording a constraint, the `wmErrDescribe` paragraph, and Pi's point that `auth`/`passphrase` name-matching newly hid same-named non-credential fields.

Known cost, called out in the body: an axios error no longer carries `extra.response.status`. That is parity with `main`, which carried nothing at all for those errors. A bounded cycle-safe projection of transport errors would be a genuine improvement, but it is an enrichment rather than part of stopping a crash, and it wants its own change.

rubenfiszel:
Both P1s addressed in `40a44eefc2`, and `wmErrDescribe` is gone rather than patched.

**The describe fallback (Codex P1 #2, Pi P1).** You are both right, and the `throw { cause: err }` framing is what made it clear. `wmErrDescribe` was my own addition, not part of the crash fix, and it was a second ungated route that reintroduced exactly the exposure I had just closed for `extra`. Removed. A thrown object exposing none of `message`/`name`/`stack` now reports `Object.prototype.toString.call(e)` — a type tag, no contents — where Deno previously stored a bare `{}`.

**The probe (Codex P1 #1).** Also right, and the fix removes the two-context problem rather than hardening it. `wmSerializeError` no longer probes and re-serializes; it attempts the plain `JSON.stringify` once and, only if that throws, retries with `extra` replaced. There is now a single serialization context, so a key-sensitive `toJSON` has nothing to disagree with. Pinned by a case whose `toJSON` returns `{ok:1}` for the root and a credential-bearing cycle for the `"extra"` key: it comes out identical to the pre-fix wrapper.

The invariant is now one line covering both routes:

> A thrown object's contents reach the result only through the one plain `JSON.stringify` the wrappers already did.

Re-ran the side-by-side property check in both `collectExtra` modes over plain objects, errors with scalar and nested-acyclic props, the cyclic `AxiosError` with sentinels in a header / `config.auth` / an mTLS key, `throw { cause: axiosErr }`, a cyclic secret-bearing plain object, an array cycle, BigInt and function props, and the key-sensitive `toJSON`. Every acyclic case byte-identical, every previously-crashing case yields the marker, sentinel count zero — under bun, node and deno. Confirmed on real bun / deno / WAC v2 jobs and by driving both generated dedicated-worker wrappers over the `execd:` protocol.

**Claude's nits.** The `[Circular]` pin is moot: the cycle-safe replacer is gone with `wmErrDescribe`, since nothing renders a cyclic graph any more. The "reports nothing" case is fixed by the type tag. The doc block was rewritten to state the invariant instead of describing the old behaviour.

**On `test_deno_job_non_error_throws`**: Pi's suggested guard is in — it throws `{ cause: <cyclic axios-shaped error with a sentinel> }` on the Deno path and asserts the sentinel is absent. It replaces the benign `{ code: 42 }` describe case, which was asserting behaviour that no longer exists.

rubenfiszel:
✅ Review round clean @ d995522ddd3fddac0539e4de3ab99bdc44c15257 — nit-only verdicts at 40a44eefc23859ff540745e1475c36cc0cd669ec (Codex: Good to merge; Claude and Pi: Mergeable with nits); nits addressed in d995522ddd / one dismissed below.

Fixed:

- **The type tag could not fire on the two Bun `collectExtra: true` wrappers** (Claude). Right, and my previous note calling that case fixed was wrong: the condition was `Object.keys(err).length === 0`, which `extra` alone satisfied, so a cyclic object thrown on Bun reported only the marker. It now keys off whether any of `message`/`name`/`stack` was found, so that case reports `{"name":"ThrownValue","message":"[object Object]","extra":"[extra omitted: not serializable]"}`. Bun and Deno agree again.
- **"reaches a result exactly when it did before" overstated the code** (Claude). Two narrow directions do not hold: `wmErrProp` skips a throwing getter where main crashed, and `wmErrString` coerces a non-serializable `message`/`name`/`stack` where main crashed. Both widen what is reported, neither renders new contents. Reworded to the half that holds: *"nothing reaches a result that the bare call did not already put there"*.
- **Truncated doc clause** (Pi). `"no blocklist of key names being complete enough to"` now reads `"...and no blocklist of key names is complete enough to render that graph safely"`.
- **Stale test comment** at `bun_jobs.rs:417` describing the removed probe (Pi, dismissed but correct). Reworded to the actual mechanism.

Dismissed:

- **A key-sensitive `toJSON` regression test** (Codex). Agreeing with Claude's adjudication: that class was reachable only because the old code serialized in two contexts. With a single serialization there is no second context to disagree with, so a test would pin the absence of a design rather than behavior, which is what AGENTS.md's "ship only the tests the PR needs" rules out. The property run covers it.

Re-verified after the change: side-by-side parity against the pre-fix wrapper in both `collectExtra` modes (every acyclic case byte-identical, every previously-crashing case yields the marker, sentinel count zero) under bun, node and deno; real bun / deno / WAC v2 jobs; 35 bun + 33 worker tests green; `cargo fmt` clean.

**Staying a draft deliberately.** This regenerates the wrapper for every Bun and Deno job, which the repo's PR policy puts in the "shared worker infrastructure" bucket that wants a human look before ready. The round behind it is clean; the reason is in the PR body.

rubenfiszel:
✅ Review round clean @ a1d7c530bde2ff4bf12ae7b3a4f931fb90f82893 — verdicts at d995522ddd3fddac0539e4de3ab99bdc44c15257 (Codex: Good to merge; Pi: Good to merge; Claude: Mergeable with nits); nits addressed in a1d7c530bd.

This round existed because my previous nit fix changed behaviour (the type tag firing when `extra` is present) that no reviewer had seen, so the earlier marker was resting on a different head. It came back clean, and Claude's two nits were both about that gap:

- **Nothing pinned the type tag on the two Bun `collectExtra: true` wrappers** — the one thing `d995522` changed. Pinned now by two assertions on the existing acyclic `throw { code: 42, hint }` case, which is exactly the changed condition (extra present, no `message`/`name`/`stack`). No new job: per AGENTS.md that is the smallest setup that exercises the new logic.
- **"Every wrapper keeps its exact previous field set" no longer held.** Corrected in the body to say what is actually true: a thrown `Error`, and a thrown object carrying its own `message`, are reported exactly as before; what changed is only what used to be unreportable.

This last commit is test-and-docs only, no wrapper change, so the runtime behaviour reviewed at `d995522` is what ships.

**Staying a draft deliberately.** This regenerates the wrapper for every Bun and Deno job, which the repo's PR policy puts in the "shared worker infrastructure" bucket that wants a human look before ready. The round behind it is clean; the reason is in the PR body.

PR #10863, opened 2026-08-26
perf: index the suspended-job resume test instead of filtering it

rubenfiszel:
## Summary

`make_suspended_pull_query` runs on every worker's poll loop. Its resume test,
`(suspend <= 0 OR suspend_until <= now())`, is a disjunction, and a disjunction cannot become
an index qual — so `queue_suspended` could only *order* the scan while the test itself ran as a
heap `Filter`. Every poll walked the entire partial index and fetched one heap tuple per
suspended row, just to conclude there was nothing to resume.

Under the `suspend_until IS NOT NULL` guard that disjunction is exactly
`CASE WHEN suspend <= 0 THEN '-infinity' ELSE suspend_until END <= now()` — a job already
holding its resume events reads as one that came due infinitely long ago. That form is
immutable, so it can be an index column. `queue_suspended_v2` replaces the two dead filter
columns with it and keeps `(priority DESC NULLS LAST, created_at)` leading, so the scan stays
ordered and never sorts. It stays a single index scan, so `FOR UPDATE SKIP LOCKED` is untouched.

### Measured

200k suspended flows, one client driving all three servers over TCP; reseed +
`VACUUM (ANALYZE, FREEZE)` + `REINDEX` between every variant, with `relpages`/`relallvisible`
asserted equal.

| state | PG14 | PG16 | PG18 |
|---|---|---|---|
| nothing due (steady state) | 201431 buf / 130 ms → **1431 / 5 ms** | 201431 / 65 ms → **1431 / 3 ms** | 101452 / 47 ms → **1443 / 4 ms** |
| a few due | 10077 / 6.5 ms → **77 / 0.3 ms** | 10077 / 3.0 ms → **77 / 0.15 ms** | 6098 / 3.1 ms → **89 / 0.24 ms** |
| half due / all approved / all timed out | 6 / 0.07 ms → **6 / 0.09 ms** | 6 / 0.04 ms → **6 / 0.04 ms** | 27 / 0.07 ms → **18 / 0.08 ms** |
| all due, no matching tag | 201431 / 122 ms → **201431 / 126 ms** | parity | parity |

Heap fetches in the steady state go from one per suspended row to zero. No state regresses on
any version, with `priority` all-NULL, sparse, or uniform over 0–100.

### Why the scan is filtered in-index rather than bounded

A range qual needs an index prefix and so does the `ORDER BY`, and
`priority DESC NULLS LAST, created_at` conflicts with `resume_time <= now()`. Both alternatives
were measured and both trade the win into another queue state: the expression leading is O(1)
when nothing is due but forces a top-1 sort costing 27–61 ms whenever most of the queue *is*
due; the expression second gets PG18 skip-scan plus incremental sort, but only while `priority`
has several distinct values — with `priority` all-NULL, the common case, the single group
degenerates into the same full sort (101451 buffers / 51 ms). Third position never sorts, which
is what makes it the only shape with no regression. A second index does not help: the planner
always picks the never-sorting one, because `LIMIT 1` makes it look O(1).

Getting to a genuinely bounded scan would mean ordering by
`(priority DESC NULLS LAST, resume_time, created_at)` — earliest-*ready* rather than
earliest-*created* within a priority. That changes which job is picked, so it is not in this PR.

## Changes

- `make_suspended_pull_query`: the `OR` becomes the equivalent `CASE` expression. The
  `suspend_until IS NOT NULL` guard stays — it is what makes the two equivalent, and without it
  every ordinary queued row (`suspend = 0`, `suspend_until IS NULL`) would match.
- New migration replacing `queue_suspended` with `queue_suspended_v2`, carrying the expression
  as its third column.
- The migration leads with `DROP INDEX IF EXISTS queue_suspended_v2` rather than relying on
  `IF NOT EXISTS`. It runs CONCURRENTLY via the `OVERRIDDEN_MIGRATIONS` rewrite, and an
  interrupted concurrent build leaves the index present but *invalid* — `IF NOT EXISTS` would
  skip rebuilding it on the retry and the following `DROP` would then retire the only usable
  index. Raised by both pre-push review passes.
- Retiring `queue_suspended` is a **separate** migration. sqlx records a migration only after
  all its statements run, so a process that dies before the record is written replays the
  build — and the leading `DROP` above would then be destroying a *valid* index rather than an
  interrupted one, leaving the pull unindexed for the length of the rebuild. Splitting it means
  the build can only ever be replayed while `queue_suspended` is still there to cover it.
  Raised by Codex in review round 1.
- `backend/tests/suspended_pull_index.rs` pins the plan. The drift this PR fixes is silent
  (right answer, slow), so nothing functional would catch a regression.
- One line at the `migration_sql.split(';')` site in `db.rs`: the split is naive, so a `;`
  anywhere in an overridden migration — inside a comment included — cuts mid-statement.

## Test plan

- [x] Predicate equivalence: 48-case truth table over `suspend` ∈ {-5,-1,0,1,2,1000} ×
      `suspend_until` ∈ {NULL, ±infinity, now()±1µs, now(), past, +14d} — 0 mismatches.
- [x] Row equivalence: 225 seeded queue states on PG16 and PG18, old query+index vs new,
      identical row picked every time.
- [x] `FOR UPDATE SKIP LOCKED` unchanged: a concurrent session gets a *different* row, and a
      third session's `FOR UPDATE NOWAIT` on the first row errors — same as baseline. No
      materialized CTE is involved.
- [x] Live on a running instance: approval resume (`suspend <= 0`), timeout resume
      (`suspend_until <= now()`), and parallel-foreach parallelism gating (`suspend = k`,
      14-day park) all resume correctly; `pg_stat_user_indexes` shows the pull loop reading
      ~2 index tuples across 54 scans.
- [x] `suspend_resume` (5), `wac_approval_resume` (2), `zombie_flow_recovery` (5) pass.
- [x] The new test fails when the query is reverted to the `OR` form.
- [x] Migration applies through the CONCURRENTLY override on a live backend
      (`indisvalid = true`), and down/up round-trips in a transaction.
- [x] Retry safety: with `queue_suspended_v2` forced invalid, replaying the three statements
      rebuilds it valid *before* the old index is dropped.

---
## Summary by cubic
Makes the suspended-job pull evaluate its resume test inside the index instead of as a heap filter. `queue_suspended` held the relevant columns but not the test, so every worker poll fetched one heap tuple per suspended row. `queue_suspended_v2` (migration `20260826202939`) replaces the two dead filter columns with one indexable `CASE` expression, keeping `(priority DESC NULLS LAST, created_at)` leading so the scan stays ordered and never sorts. It stays a single index scan, so `FOR UPDATE SKIP LOCKED` is untouched. The legacy index is retired by a separate migration (`20260826214706`) once the build is recorded.

### Measured

200k suspended flows, one client driving all three servers over TCP; reseed + `VACUUM (ANALYZE, FREEZE)` + `REINDEX` between every variant, with `relpages`/`relallvisible` asserted equal.

| state | PG14 | PG16 | PG18 |
|---|---|---|---|
| nothing due (steady state) | 201431 buf / 130 ms → **1431 / 5 ms** | 201431 / 65 ms → **1431 / 3 ms** | 101452 / 47 ms → **1443 / 4 ms** |
| a few due | 10077 / 6.5 ms → **77 / 0.3 ms** | 10077 / 3.0 ms → **77 / 0.15 ms** | 6098 / 3.1 ms → **89 / 0.24 ms** |
| half due / all approved / all timed out | 6 / 0.07 ms → **6 / 0.09 ms** | 6 / 0.04 ms → **6 / 0.04 ms** | 27 / 0.07 ms → **18 / 0.08 ms** |
| all due, no matching tag | 201431 / 122 ms → **201431 / 126 ms** | parity | parity |

Heap fetches in the steady state go from one per suspended row to zero. No state regresses on any version, with `priority` all-NULL, sparse, or uniform over 0–100.

### Why the scan is filtered in-index rather than bounded

A range qual needs an index prefix and so does the `ORDER BY`, and `priority DESC NULLS LAST, created_at` conflicts with `resume_time <= now()`. Both alternatives were measured and both trade the win into another queue state: the expression leading is O(1) when nothing is due but forces a top-1 sort costing 27–61 ms whenever most of the queue *is* due; the expression second gets PG18 skip-scan plus incremental sort, but only while `priority` has several distinct values — with `priority` all-NULL, the common case, the single group degenerates into the same full sort (101451 buffers / 51 ms). Third position never sorts, which is what makes it the only shape with no regression. A second index does not help: the planner always picks the never-sorting one, because `LIMIT 1` makes it look O(1).

Getting to a genuinely bounded scan would mean ordering by `(priority DESC NULLS LAST, resume_time, created_at)` — earliest-*ready* rather than earliest-*created* within a priority. That changes which job is picked, so it is not in this PR.

## Changes

- `make_suspended_pull_query`: the `OR` becomes the equivalent `CASE` expression. The `suspend_until IS NOT NULL` guard stays — it is what makes the two equivalent, and without it every ordinary queued row (`suspend = 0`, `suspend_until IS NULL`) would match.
- New migration builds `queue_suspended_v2`, carrying the expression as its third column. It leads with `DROP INDEX IF EXISTS queue_suspended_v2` rather than relying on `IF NOT EXISTS`: it runs CONCURRENTLY via the `OVERRIDDEN_MIGRATIONS` rewrite in `db.rs`, and an interrupted concurrent build leaves the index present but *invalid*, which `IF NOT EXISTS` would skip rebuilding on retry.
- Retiring `queue_suspended` is a **separate** migration. sqlx records a migration only after all its statements run, so a process that dies before the record is written replays the build — and the leading `DROP` above would then destroy a *valid* legacy index rather than an interrupted one, leaving the pull unindexed for the length of the rebuild. Splitting it means the build can only ever be replayed while `queue_suspended` is still there to cover it.
- `backend/tests/suspended_pull_index.rs` pins the plan. The drift this PR fixes is silent (right answer, slow), so nothing functional would catch a regression.
- One comment at the `migration_sql.split(';')` site in `db.rs`: the split is naive, so a `;` anywhere in an overridden migration — inside a comment included — cuts mid-statement.

## Test plan

- [x] Predicate equivalence: 48-case truth table — 0 mismatches.
- [x] Row equivalence: 225 seeded queue states on PG16 and PG18, old query+index vs new — identical row picked every time.
- [x] `FOR UPDATE SKIP LOCKED` unchanged: a concurrent session gets a *different* row, and a third session's `FOR UPDATE NOWAIT` on the first row errors — same as baseline.
- [x] Live on a running instance: approval resume, timeout resume, and parallelism gating all resume correctly; `pg_stat_user_indexes` shows the pull loop reading ~2 index tuples across 54 scans.
- [x] `suspend_resume` (5), `wac_approval_resume` (2), `zombie_flow_recovery` (5) pass.
- [x] The new test fails when the query is reverted to the `OR` form.
- [x] Migration applies through the CONCURRENTLY override on a live backend (`indisvalid = true`), and down/up round-trips in a transaction.
- [x] Retry safety: with `queue_suspended_v2` forced invalid, replaying the statements rebuilds it valid *before* the old index is dropped.

<sup>Written for commit 3bb2ae0be19d011460bf50bf44c51d8a3de4dcaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10863?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
On the second nit — `.github/workflows/backend-test.yml:26` running an untagged `image: postgres` — I'm dismissing rather than fixing, for two reasons.

Pinning the service image changes the Postgres major for *every* backend test, not just this one. That's a CI-wide decision with its own fallout (unrelated suites shifting behaviour on a downgrade from `latest`), and it doesn't belong to a PR about one index.

The narrower framing also doesn't hold up: `Index Cond` vs `Filter` for an indexable expression on an index column isn't a planner heuristic that drifts between majors — it's whether the qual is matchable at all. I measured this shape on 14, 16 and 18 and it lands in `Index Cond` on all three. If a future major *did* change that, the suspended pull would silently go back to a heap fetch per suspended row, which is precisely the regression this test exists to catch. A red `suspended_pull_index` on a Postgres bump is the signal working, not a false positive — and the failure message now names `version()` so it reads as such.

The formatting fragility you flagged first was the real half, and that's fixed: the test parses `EXPLAIN (FORMAT JSON)` and reads `Index Cond` / `Filter` as keys, so it no longer depends on EXPLAIN's line layout.

PR #10865, opened 2026-08-26
[ee] fix: index job logs as plain text so every line stays searchable

rubenfiszel:
## Summary

Job logs were run through `serde_json::to_string` before being added to the tantivy document. That encodes every newline as a literal `\n`, and tantivy's default tokenizer splits on the backslash but not the `n` — welding it onto the following word. **The first term of every log line after the first was unsearchable**, silently, for as long as the index has existed.

`args` and `result` genuinely are JSON and keep their encoding; only `logs` is plain text.

## Changes

- Index `job_logs` verbatim in `build_job_doc` instead of JSON-encoding it (`windmill-indexer/src/completed_runs_ee.rs`)

## Test plan

Measured against a real index, on a job whose logs are `alphamarker betamarker\ngammamarker deltamarker`:

| search term | before | after |
|---|---|---|
| `alphamarker` (line 1) | 1 hit | 1 hit |
| `gammamarker` (line 2, first word) | **0 hits** | 1 hit |
| `ngammamarker` (the welded artifact) | 1 hit | 0 hits |

- [x] Terms at the start of a continuation line are found
- [x] Terms on the first line still found (no regression)
- [x] The welded `n`-prefixed artifact no longer matches
- [ ] Existing indices pick this up on their next rebuild; no schema change, so no forced wipe

Companion PR: windmill-labs/windmill-ee-private#TBD

---
## Summary by cubic
Bumps the EE ref to index job logs as plain text instead of JSON-encoding them, so every line stays searchable.

- Previously, encoding newlines as `\n` welded the `n` to the following word, making the first term of each continuation line unsearchable.
- Adds a test pinning that a term starting a log continuation line stays searchable.
- Documents the tokenizer constraint once, at the site that enforces it.
- No schema change, so existing indices pick up the fix on their next rebuild.

<sup>Written for commit b32457d8acf987bec6e1e375d00e3e91a319d9ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Both nits addressed — one fixed, one I think is inaccurate.

**P2 regression guard — fixed** in windmill-ee-private@4657e39 (ref bumped here in d15b274). Extended `test_job_indexing_pipeline` exactly as suggested: `job1`'s fixture is now `"job1 is logging the word transparent\nmulticadence tail"` and a new assertion requires `multicadence` to return 1 hit. No new fixtures.

I checked it's a real guard rather than a passing test: with the fix reverted it fails `left: 0, right: 1`, and passes once restored.

**P2 no automatic reindex — I don't think this holds.** The index self-heals; no operator action is needed on a default instance.

`refresh_jobs` calls `monitor_job_index` at the top of every cycle, which calls `delete_old_jobs_from_index` to prune every document outside the effective window — `min(max_index_time_window_secs, retention)`, defaulting to 7 days. So mangled documents age out on their own, and jobs indexed after the deploy are correct from the start. Within one window period the whole index is consistent.

The genuine edge case is narrower than "existing data stays broken": an instance configured to **never** prune — `max_index_time_window_secs` and retention both `0`, which is the `(false, false) => 0` arm of the `effective` closure — keeps the old tokens indefinitely. That's the only configuration where `should_clear_job_index` is required, and I've said so in the PR body rather than in release notes, since it doesn't apply to a default deployment.

The mixed-index window you describe is real but bounded and self-resolving, so I don't think it warrants operator instructions.

PR #10866, opened 2026-08-27
docs(api): document large completed job result placeholder

### LINKED ISSUE #10860: bug: results above ~8 MB are silently replaced with WINDMILL_TOO_BIG while the job reports success
### Describe the bug

When a script returns a result above roughly 8 MB, the result is replaced with the literal string `"WINDMILL_TOO_BIG"` and **the job is still marked `success: true`**.

A downstream flow step consuming that output receives the string `"WINDMILL_TOO_BIG"` where it expected its data, with nothing anywhere indicating that a substitution happened. There is no error, no warning in the job, and the run shows green.

### To reproduce

Script (`bun`):

```ts
export async function main(mb: number) {
  return { blob: "x".repeat(mb * 1024 * 1024) };
}
```

Run it with increasing `mb` and read the result back via `GET /api/w/{workspace}/jobs_u/get/{id}`:

| `mb` | `success` | `result` |
|---|---|---|
| 1 | true | `{"blob":"xxx…"}` — intact |
| 2 | true | intact |
| 3 | true | intact |
| 4 | true | intact |
| 6 | true | intact |
| **8** | **true** | **`"WINDMILL_TOO_BIG"`** |
| 10 | true | `"WINDMILL_TOO_BIG"` |
| 50 | true | `"WINDMILL_TOO_BIG"` |
| 200 | true | `"WINDMILL_TOO_BIG"` |

The cutoff sits between 6 and 8 MB, which is consistent with an 8 MiB (8,388,608 byte) cap on the serialised result.

Durations stay plausible throughout (200 MB completed in 1,831 ms), so nothing about the run suggests the payload was dropped.

### Expected behavior

Any of the following would be an improvement, roughly in order of preference:

- Fail the job when the result cannot be stored, so the flow's error path runs.
- Keep `success: true` but make the substitution visible in the job — a dedicated field or a log line, rather than only a magic string in the result position.
- At minimum, document the limit and the sentinel value.

Marking the job successful while silently swapping the payload for a sentinel is the combination that makes this hard to catch: the failure surfaces later, in a different step, as a type error or as corrupted downstream data.

### Screenshots

_No response_

### Browser information

_Not applicable — reproduced via API._

### Application version

`CE v1.796.0-12-gffdf17ef8d`, image `ghcr.io/windmill-labs/windmill:main`, digest `sha256:f8513becee64ac47b0099a84adde4bb90962f396768070671b0ab9790e343f87`. Stack from the repo's own `docker-compose.yml`.

### Additional Context

The practical case this came from is document processing: a step reads a scanned invoice and passes the extracted content to the next step. A large scan crosses the threshold, the extraction step reports success, and the booking step receives `"WINDMILL_TOO_BIG"`.

Possibly related to #6855, which reports large base64 results filling Postgres TOAST tables — that one is about the cost of storing large results, this one is about what happens when they are not stored.

I could not find the sentinel string documented anywhere; searching the issue tracker for `WINDMILL_TOO_BIG` returns nothing.

bpiper02:
## Summary

Documents the `WINDMILL_TOO_BIG` placeholder returned in `CompletedJob.result` for large results.

The API already exposes a separate completed-job result endpoint for retrieving the full result, but the placeholder behavior was not documented in the OpenAPI schema.

## Why

This addresses the minimum documented-improvement path described in #10860: make the sentinel value and recovery path explicit for API consumers.

## Testing

- `git diff --check`
- Documentation-only OpenAPI change; no runtime behavior changed

Fixes #10860

I have read the CLA Document and I hereby sign the CLA.

---
## Summary by cubic
Documents the `WINDMILL_TOO_BIG` placeholder returned in `CompletedJob.result` for large results, so API consumers know to use the completed job result endpoint to retrieve the full result. Documentation-only change without runtime behavior impact. Fixes #10860.

<sup>Written for commit c2843786f759306404959cef050126fe94bbb92f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10866?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api/openapi.yaml
+++ b/backend/windmill-api/openapi.yaml
@@ -27991,9 +27991,9 @@ components:
         args:
           $ref: "#/components/schemas/ScriptArgs"
         result:
-	   description: |
-	     For large results, this may be the placeholder string 'WINDMILL_TOO_BIG'.
-	     Use the completed job result endpoint to retrieve the full result.
+          description: |
+            For large results, this may be the placeholder string 'WINDMILL_TOO_BIG'.
+            Use the completed job result endpoint to retrieve the full result.
         logs:
           type: string
         deleted:
--- a/backend/windmill-api/openapi.yaml
+++ b/backend/windmill-api/openapi.yaml
@@ -27990,7 +27990,10 @@ components:
           type: string
         args:
           $ref: "#/components/schemas/ScriptArgs"
-        result: {}
+        result:
+	   description: |
+	     For large results, this may be the placeholder string 'WINDMILL_TOO_BIG'.
+	     Use the completed job result endpoint to retrieve the full result.
         logs:
           type: string
         deleted:


bpiper02:
I have read the CLA Document and I hereby sign the CLA.

bpiper02:
recheck

hugocasa:
Hi thanks for the contribution! Can you make sure to use spaces instead of tabs for the indentation?

diegoimbert:
/ai use spaces instead of tabs for indentation

PR #10867, opened 2026-08-27
[ee] feat: serve run and log search from any instance, not just the indexer

rubenfiszel:
## Summary

Only the indexer holds the in-memory tantivy `IndexReader`, so `/api/srch/*` answered on that pod alone and deployments had to pin ingress routing to it (WIN-1968) — the `reverse_proxy /api/srch/*` line commented out in our own `Caddyfile`. Search now works on every instance: the indexer advertises a reachable base URL on the lock row it already heartbeats, and an instance without a reader forwards the request there.

Forwarding replays the caller's own `Authorization`/`Cookie`, which the indexer re-authenticates independently — the proxying instance does not vouch for anyone. `require_devops_role` still runs on the proxying instance first, so a non-devops caller is rejected before any outbound request is made.

## Changes

- `concurrency_locks.owner_addr`: the indexer stamps `{owner}|{url}` on the lock it already refreshes. Pairing the address with its writer means a foreign writer invalidates it — an indexer from a build predating the column refreshes `last_locked_at` without touching `owner_addr`, so on a rollback the row would otherwise keep a dead pod's address and look live forever.
- The advertised port comes from `DEFAULT_PORT`, now shared from `windmill-common` rather than duplicated. The server binds `port_var.unwrap_or(DEFAULT_PORT)`, so deriving the address from `PORT` alone advertised nothing on a default deployment.
- `INDEXER_ADVERTISED_URL` overrides the derived address (needed on Kubernetes, where a bare pod hostname only resolves through a headless service). Non-`http(s)` values are rejected — the address receives credentials.
- An `x-windmill-indexer-proxy` marker stops two readerless instances bouncing a request between them.
- A connection failure returns `BadGateway` with the generic routing message; the address is logged, not returned, since the workspace route is reachable by operators.
- Explicit 120s forward timeout instead of inheriting `HTTP_CLIENT`'s 20s, so the proxied and direct paths behave the same.
- Removes `backup_index_to_store` and four helpers, dead since the directory started mirroring writes to object storage as they happen.

## Test plan

- [x] `test_resolve_advertised_url` (8 cases) and `test_addr_for_owner` (5 cases), no real env vars
- [x] Indexer with **`PORT` and `BACKEND_PORT` both unset**: binds 8000, writes a real `owner_addr`, and a separate readerless server returns hits — the default-deployment shape
- [x] Indexer log shows the forwarded request arriving as the **caller's** identity, not the proxy's
- [x] Rollback simulated in SQL (foreign `owner`, fresh `last_locked_at`, address intact): readerless instance stops forwarding and returns the routing message; restoring the owner returns it to 200
- [x] Unreachable address: HTTP 502, address in the server log and absent from the response body
- [x] Unauthenticated request rejected before forwarding; loop marker yields a clean error, no second hop
- [ ] Not exercised: a real multi-pod cluster. Verified on one host, so the `INDEXER_ADVERTISED_URL` path a cluster would use is covered only by its unit tests.

Companion PR: https://github.com/windmill-labs/windmill-ee-private/pull/749

---
## Summary by cubic
Fixes WIN-1968 by letting any instance answer run and log search: a server without an in-memory index reader now forwards `/api/srch/*` to the indexer instead of only the indexer pod answering, which forced deployments to pin ingress routing to it.

- The indexer writes its reachable base URL to `concurrency_locks.owner_addr` on the lock it already heartbeats, and readerless instances read that row and forward when the lock looks live.
- Forwarding replays the caller's `Authorization`/`Cookie`; `require_devops_role` still runs first, and the indexer re-authenticates independently.
- An `x-windmill-indexer-proxy` marker prevents two readerless instances from bouncing a request; a failed forward returns `BadGateway` with the address only in server logs.
- The forward timeout is explicitly 120s so the proxied and direct paths behave the same.
- The advertised URL derives its port from `PORT`, then `BACKEND_PORT`, then `DEFAULT_PORT` (now shared from `windmill-common`), and `--help` documents that fallback.
- Removes `backup_index_to_store` and four helpers dead since directory writes started mirroring to object storage.

**Migration**
- Adds `owner_addr` to `concurrency_locks`; because the address is stored with its writer's identity, an older indexer that takes over the lock invalidates the stale address instead of leaving a dead pod's address looking live.
- `INDEXER_ADVERTISED_URL` overrides the derived address for Kubernetes, where a bare pod hostname resolves only through a headless service; non-`http(s)` values are rejected since the address receives credentials. The compose example sets it explicitly, and the reference deployment no longer pins `/api/srch/*` to the indexer.

<sup>Written for commit 936780945405e6176bb102a538d85118759655e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10867?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Codex: **Good to merge** on this head. Claude's remaining nit is already fixed — it's a stale-checkout artefact, not an outstanding issue.

**The nit asks for `without_url()` on the `resp.bytes()` path.** That landed in windmill-ee-private@17bb4b0, which is exactly what `backend/ee-repo-ref.txt` points to on this head:

```rust
let body = resp.bytes().await.map_err(|e| {
    tracing::error!("Failed to read the indexer response from {log_target}: {}", e.without_url());
    Error::BadGateway(NO_INDEX_READER_ROUTING_MSG.to_string())
})?;
```

Claude's own round-1 review flagged why it can't see this: its `windmill-ee-private` checkout is pinned at the base ref `b58ad414` and it can't fetch the branch, so every round it has reviewed the EE half as it stood *before* this PR. Worth knowing more generally — on this PR the Claude reviewer has only ever seen the migration, the sqlx cache, `DEFAULT_PORT` and the help text. The forwarding handler, `resolve_advertised_url`, `addr_for_owner` and the loop marker have been reviewed by Codex alone.

**On the regression test Codex asked for last round** (a truncated response authenticated with `?token=SECRET`): not added, deliberately. Both failure paths now return the *constant* `NO_INDEX_READER_ROUTING_MSG`, so nothing dynamic remains that could carry a token into the response, and asserting a constant is constant would need a fake upstream to prove nothing. `forward_log_target_drops_the_query_string` covers the one place that still composes a string from the URI. Flagging it rather than quietly skipping.

Both reqwest error paths in the function are now sanitized, and I swept the rest: the remaining `{e}` formats are `spawn_blocking` and serde errors that carry no URL.

PR #10876, opened 2026-08-28
feat: enable Anthropic prompt caching on Vertex AI agent steps

hugocasa:
## Summary

The Anthropic query builder skipped `cache_control: {"type": "ephemeral"}` whenever the platform was Google Vertex AI, so AI agent steps running Claude models through Vertex never got prompt caching. Vertex serves the same Messages API (`vertex-2023-10-16`) and supports explicit cache breakpoints, and the frontend copilot path already applies `cache_control` for every Anthropic platform, so the backend exclusion was unintended.

The three `is_vertex()` guards are removed, and Vertex agent steps now emit the same cache breakpoints the standard Anthropic platform already gets.

Fixes WIN-2464

## Changes

- `backend/windmill-ai/src/providers/anthropic.rs`: drop the `is_vertex()` guards around `cache_control` on the system prompt, the last custom tool definition, and the last content block of the last message.
- Add `sets_cache_breakpoints_on_every_platform`, which asserts all three breakpoints for both `AIPlatform::Standard` and `AIPlatform::GoogleVertexAi`.
- Add a `DISABLE_ANTHROPIC_PROMPT_CACHING` env flag that suppresses all three breakpoints, for Vertex projects that have prompt caching disabled (see below).

The change is scoped to Anthropic-on-Vertex: `AIPlatform::GoogleVertexAi` only reaches `AnthropicQueryBuilder` through `AIProvider::Anthropic`. Azure Foundry is constructed with `AIPlatform::Standard` and Bedrock routes to `OtherQueryBuilder`, so neither took the guarded branch before or after. The request still carries at most 3 breakpoints, under Anthropic's limit of 4.

## Test plan

- [x] `cargo test -p windmill-ai --lib` — 112 tests pass, including the new one. The new test was confirmed to fail when a guard is restored, so it is a real regression guard.
- [x] End-to-end against the running backend: ran an AI agent step (`preview_flow`, `kind: anthropic`, `platform: google_vertex_ai`) pointed at a local stub standing in for the Vertex endpoint, and inspected the request the backend actually sent to `…/models/claude-sonnet-4-5:streamRawPredict`. The body carried `anthropic_version: vertex-2023-10-16` plus `cache_control: {"type":"ephemeral"}` on `system[0]`, on the last entry of `tools`, and on the last content block of the last message. Both the tool-less and the with-tool variants completed successfully.

---
## Summary by cubic
Enables Anthropic prompt caching on Vertex AI agent steps so Claude models on Vertex get the same cache breakpoints as the standard Anthropic platform. Previously the backend skipped `cache_control` on Vertex even though Vertex supports explicit breakpoints and the frontend already applies them. Fixes WIN-2464.

- Removes the `is_vertex()` guards that suppressed `cache_control` on the system prompt, the last custom tool, and the last message content block.
- Adds a regression test asserting all three breakpoints on both `Standard` and `GoogleVertexAi` platforms.
- Adds a `DISABLE_ANTHROPIC_PROMPT_CACHING` escape hatch, which suppresses breakpoints on every Anthropic platform — not just Vertex — for Google projects that have explicit prompt caching disabled.

<sup>Written for commit d05be770a355865acce58a5db8390fce4b760915. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Vertex projects with prompt caching disabled

A Google Cloud project can have explicit prompt caching turned off by request to Cloud support, after which Vertex rejects any request carrying cache breakpoints. `DISABLE_ANTHROPIC_PROMPT_CACHING=true` turns the breakpoints off instance-wide so such a project can still run Anthropic agent steps. It follows the existing `ALLOW_PRIVATE_AI_BASE_URLS` pattern in `ai_providers.rs` and defaults to off, so nothing changes for the ordinary case.

An instance-level flag was chosen over a per-resource field because the constraint is a property of the GCP project, and a resource-level toggle would mean changing the AI provider resource type, which lives in a separate repo.

## Verification limits

The request body the backend emits for Vertex is verified end-to-end against a local stub, with and without the flag. Real Vertex API responses (and therefore actual `cache_read_input_tokens` accrual across tool round-trips) are unexercised, since that needs a GCP project with Vertex Claude access that the dev environment does not have.

The flag itself has no unit test: it is a `lazy_static` read from the process environment, so a test cannot vary it within one test binary. It is covered by the runtime check above instead.

### DIFF
--- a/backend/windmill-ai/src/ai_providers.rs
+++ b/backend/windmill-ai/src/ai_providers.rs
@@ -24,6 +24,14 @@ lazy_static::lazy_static! {
         .ok()
         .map(|v| v == "true" || v == "1")
         .unwrap_or(false);
+    /// A Google Cloud project can have explicit prompt caching turned off (by request to
+    /// Cloud support), and Vertex then rejects any request carrying cache breakpoints.
+    /// Such a project needs this to run Anthropic agent steps at all.
+    pub static ref DISABLE_ANTHROPIC_PROMPT_CACHING: bool =
+        std::env::var("DISABLE_ANTHROPIC_PROMPT_CACHING")
+            .ok()
+            .map(|v| v == "true" || v == "1")
+            .unwrap_or(false);
 }
 
 pub const OPENAI_BASE_URL: &str = "https://api.openai.com/v1";
--- a/backend/windmill-ai/src/ai_providers.rs
+++ b/backend/windmill-ai/src/ai_providers.rs
@@ -24,9 +24,11 @@ lazy_static::lazy_static! {
         .ok()
         .map(|v| v == "true" || v == "1")
         .unwrap_or(false);
-    /// A Google Cloud project can have explicit prompt caching turned off (by request to
-    /// Cloud support), and Vertex then rejects any request carrying cache breakpoints.
-    /// Such a project needs this to run Anthropic agent steps at all.
+    /// Drops the cache breakpoints from agent-step requests on every Anthropic platform,
+    /// not just the one that motivates it: a Google Cloud project can have explicit prompt
+    /// caching turned off (by request to Cloud support), and Vertex then rejects any request
+    /// carrying breakpoints. An instance that sets this to unblock such a project also gives
+    /// up caching on its direct-Anthropic and Foundry resources.
     pub static ref DISABLE_ANTHROPIC_PROMPT_CACHING: bool =
         std::env::var("DISABLE_ANTHROPIC_PROMPT_CACHING")
             .ok()
--- a/backend/windmill-ai/src/providers/anthropic.rs
+++ b/backend/windmill-ai/src/providers/anthropic.rs
@@ -636,11 +636,7 @@ impl AnthropicQueryBuilder {
             vec![AnthropicSystemContent {
                 r#type: "text".to_string(),
                 text,
-                cache_control: if self.is_vertex() {
-                    None
-                } else {
-                    Some(CacheControl::ephemeral())
-                },
+                cache_control: Some(CacheControl::ephemeral()),
             }]
         });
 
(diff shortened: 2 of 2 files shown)


hugocasa:
Addressed the P1 in `1b93ab9a10`.

Added `DISABLE_ANTHROPIC_PROMPT_CACHING`, an instance-level flag following the existing `ALLOW_PRIVATE_AI_BASE_URLS` pattern in `ai_providers.rs`. When set, all three breakpoints are suppressed, so a Vertex project with explicit caching disabled can still run Anthropic agent steps. Default is off, so the ordinary case is unchanged.

I went with an instance-level flag rather than a per-resource field because the constraint is a property of the GCP project, and a resource-level toggle would mean changing the AI provider resource type, which lives in a separate repo.

Verified against a stub standing in for the Vertex endpoint: with the flag set, the body the backend sends to `…/models/claude-sonnet-4-5:streamRawPredict` contains zero `cache_control` occurrences (system, tools, and messages all clean) and the agent step completes; without it, all three breakpoints are present. No unit test for the flag itself — it is a `lazy_static` read from the process environment, so a test binary cannot vary it.

hugocasa:
✅ Review round clean @ d05be770a355865acce58a5db8390fce4b760915 — nit-only verdicts at 1b93ab9a1017978587ef1298a6abc367164c79c7 (codex: Good to merge; claude: Mergeable, address nits); nit addressed in d05be770a3.

The nit was correct: `DISABLE_ANTHROPIC_PROMPT_CACHING` is read unconditionally in `build_request`, so it drops breakpoints on every Anthropic platform, not only Vertex. Kept the behaviour (the name promises a global switch, and an operator wanting one shouldn't have to set a Vertex-shaped flag) and rewrote the doc comment to say so outright, including that an instance setting it also gives up caching on its direct-Anthropic and Foundry resources.

PR #10879, opened 2026-08-28
feat: flag a schedule whose runs outlast the interval between its slots

### LINKED ISSUE #10861: bug: effective schedule silently changes when a job runs longer than its interval
### Describe the bug

When a scheduled job takes longer than its own interval, the effective cadence changes and nothing says so. The schedule keeps displaying the configured cron expression, and there is no warning anywhere that runs are happening at a different rate than configured.

To be explicit about what is **not** being reported: the absence of pile-up is correct and desirable. Windmill schedules the next run only after the previous one finishes, so a slow job does not stack up or double-process. That part is good design. The report is about the silence around the resulting drift.

### To reproduce

Script (`bun`) that takes 50 s:

```ts
export async function main() {
  await new Promise(r => setTimeout(r, 50000));
  return { done: new Date().toISOString() };
}
```

Schedule it with `*/20 * * * * *` (every 20 seconds), enabled, UTC.

Observed over ~4.5 minutes:

```
2026-08-26T20:43:20.026761Z   (first)
2026-08-26T20:44:20.041157Z   gap 60s
2026-08-26T20:45:20.013396Z   gap 60s
2026-08-26T20:46:20.027231Z   gap 60s
```

Average gap: **60 s**, against a configured 20 s. The behaviour is deterministic — the 50 s job finishes past the next two cron slots, so the next matching slot is always 60 s out.

Reading the schedule back afterwards:

```
schedule = '*/20 * * * * *'
```

Unchanged, with no indication that the effective rate is one third of it.

### Expected behavior

Make the divergence visible. Any of these would do:

- A warning when a run's duration exceeds the schedule interval.
- An indicator on the schedule showing the effective cadence next to the configured one.
- A field on the schedule recording the last observed interval.

### Screenshots

_No response_

### Browser information

_Not applicable — reproduced via API._

### Application version

`CE v1.796.0-12-gffdf17ef8d`, image `ghcr.io/windmill-labs/windmill:main`, digest `sha256:f8513becee64ac47b0099a84adde4bb90962f396768070671b0ab9790e343f87`. Stack from the repo's own `docker-compose.yml`.

### Additional Context

This is a condition that arises on its own over time rather than at setup. A job that initially takes two minutes and grows to twelve as data volume increases will quietly stop keeping its hourly cadence, and the only trace is timestamps drifting apart.

Two practical consequences for anyone relying on schedules:

- **The number of runs in a period cannot be derived from the cron expression.** Anything that counts scheduled executions has to count actual runs.
- For monitoring, the useful signal is the *interval between consecutive runs*, not whether the schedule is enabled — an enabled schedule running at a third of its configured rate looks healthy by every other measure.

hugocasa:
## Summary

Closes #10861.

A script schedule queues its next run only once the previous one has completed. So a run that takes longer than the gap between two slots pushes the next run out, the schedule quietly runs less often than its cron says, and nothing anywhere reports it: the cron expression still reads the same, and the effective cadence is recorded nowhere.

This flags it on the schedules page. It does not change how schedules run, and does not add an overlap policy for script schedules.

The check is the direct one: the page already loads the last 20 runs of every schedule, with their durations, to draw the sparkline. If the median run outlasts the interval between slots, the next run necessarily lands on a later slot. For a script schedule that isn't an approximation of drift, it is exactly the condition that causes it.

The one thing the page can't work out for itself is the interval, since that needs the cron evaluated, so `list_with_jobs` now returns it. It reports the *shortest* gap over the next few occurrences: under an irregular expression a run can outlast the tight gaps while still fitting inside the mean, and the tight ones are the ones it skips.

## Changes

- `list_with_jobs` returns `interval_s` (shortest gap the cron asks for) and `queues_next_run_at_start`. Both are computed from the schedule row already being read; no extra query, and no change to any endpoint that carries a schedule's configuration.
- `runsOutlastingInterval` in `frontend/src/lib/components/schedules/scheduleDrift.ts`: the check, over the runs the page already has, returning how long those runs have been taking so both call sites can state the figure.
- A warning in the schedule editor, under the cron and beside the panel listing the events it is missing. It is fed by an optional `getRunsSample` prop, so the drawer costs no fetch of its own and simply shows nothing when opened from somewhere that has no runs to hand (a script or flow's triggers panel).
- A warning triangle beside that schedule's runs on the schedules page, with the short version on hover, so a workspace can be scanned without opening each schedule.

Not flagged: flows and scripts with a skip handler, which queue the next run as the previous one *starts*, so a long run doesn't push the following one out; disabled schedules; and schedules with fewer than three recent runs, so that one slow run isn't read as a change of cadence.

## Screenshots

In the schedule editor, under the cron and next to the events it is missing.

editor warning

On the schedules page, beside the runs it describes. Four schedules on the same slow script; only the enabled plain-script one is marked.

list triangle

## Test plan

Exercised on a local instance with the issue's setup: a script that sleeps 10s, on a sub-minute cron.

- [x] The plain script schedule is flagged on both surfaces; the endpoint reports `interval_s: 3` against a median run of 10054ms
- [x] A flow schedule on the same script is not flagged (`queues_next_run_at_start: true`)
- [x] A script schedule with a skip handler is not flagged, for the same reason
- [x] Disabled schedules are not flagged, and neither is one with no runs yet
- [x] A daily cron reports `interval_s: 86400`, and the unit test pins an irregular cron to its shortest gap rather than its mean
- [x] Both numbers in the warning render through one formatter, so the comparison reads correctly past a minute (`2m 10s` against `2m 0s`, not `2m` against `2m 0s`)

## Note on the history

The first twelve commits on this branch implemented this server-side instead: sampling each schedule's job history, recomputing the cron slot each past run was queued for, and comparing consecutive slots. It measured more (it caught drift from runs starting late, not just running long, and worked for flows) but the shape cost far more than it returned — a scan budget, an index question, a trigger-history invalidation boundary, and two P1s from attaching a runtime measurement to the endpoint that serves a schedule's configuration. It is reverted in `2966c7a` and replaced by the version above; the diff against main is only this version.


hugocasa:
Closing in favour of a smaller change built around the signal the issue actually named.

This branch measured the wrong thing twice. It started by sampling each schedule's job history server-side and recomputing the cron event every past run was queued for, which cost a scan budget, an index question, a trigger-history invalidation boundary and two P1s — all consequences of attaching a runtime measurement to the endpoint that serves a schedule's configuration. It was then rewritten (2966c7a) as a much smaller check comparing recent run durations against the interval, which is sound but infers the skip from one of its causes: a run that is quick yet waits for a worker drifts identically and stays silent.

The replacement detects the skip directly, from the runs the schedules page already loads: consecutive runs sitting at least two intervals apart means at least one scheduled event was skipped. That is what #10861 asked for ("the number of runs in a period cannot be derived from the cron expression"), it covers both causes rather than one, and taking the *minimum* gap keeps a pause or a disabled stretch from ever reading as drift.

A fresh PR follows. Reopening this one is not worth it: seventeen commits here implement approaches that no longer exist.

PR #10880, opened 2026-08-28
feat(frontend): group the agent form and edit saved agents as drafts

hugocasa:
## Summary

Two changes that turned out to be one. The AI agent step form was a flat dump of `AI_AGENT_SCHEMA`'s eleven properties labelled with their snake_case keys; it now groups into **Model / Messages / Tools / Output** and hides optional fields until they hold something. And editing a *saved* agent no longer forks its config into the flow step: the step stays linked and the edits become an ordinary **resource draft**, autosaved, surviving a reload, listed on Compare & Deploy, deployed by the code that already deploys resource drafts.

The second needed the first. Once the edits live in the resource rather than in a step, the agent needs an editor that is not a step panel, and that editor is the same grouped form.

### The agent form

- `agentFormFields.ts` is a registry of what each field is called, which group it belongs to, and **what an absent key does at run time**. That last point is the rule for whether a field shows: not `schema.default`, which has disagreed with the executor. The add-field menu states it (`Default: text`, `Default: off`, `Default: 10`).
- Core fields are always present; the rest are added from a menu and removed with an `x`. Visibility is sticky, so emptying a box never makes its row vanish under the cursor.
- The same form serves the step panel, the standalone editor, and (filtered) a linked step.

### Editing a saved agent

- `agentDraft.svelte.ts` composes the existing `useTriggerDraftSync` over `itemKind: 'resource'` and writes the same `ResourceState` shape the generic editor writes. That is what makes agent drafts appear on the deploy page with no new code.
- `AgentEditorHost.svelte` builds a synthetic single-module flow so the agent can be edited outside any flow. Reached from the agent card in a flow step, from the resources page row, and from a `#/resource/<path>` deep link.
- Evals moved inside the editor, the one place an agent is edited. `AgentEvalModal` and the resources-page dropdown item are gone.
- The fork machinery goes with it: `agentEditStore.svelte.ts`, `flowStoreRefresh.svelte.ts`, and the `forkMarker` / `sameSession` / `discardedOnLinkSnapshot` guards that existed only to manage a homeless edit. Unlink keeps its `forkFromResource`, which genuinely wants the config folded into the step.

### Tools

- A tool opens in a **drawer** beside the agent rather than replacing it, titled by the tool's name.
- Tools can be added and deleted from the agent's Tools section, through the same insert path as the graph's `+ Tool` (extracted to `flowStateUtils`, so the graph and the editor share one implementation).
- A saved agent's tool inputs offer **AI or static only**. Its inputs live on the resource and cannot name any one flow's `flow_input` or `results`; a flow binds those on the step through `tool_inputs`.
- At flow level, a linked agent's tool inputs are collapsible.

### Behaviour fixes

| Fix | Detail |
|---|---|
| Opening a step no longer dirties the flow | `ArgInput` materialises `schema.default` for every row it renders, so the old form wrote six spurious keys just by being looked at |
| Data loss on `output_type` | Four fields carried `showExpr: "fields.output_type === 'text'"`, and `InputTransformForm` *clears* a field whose `showExpr` is false. A step that never materialised `output_type` had its temperature and memory wiped. Now `!== 'image'` |
| `streaming` defaults to on | `unwrap_or(true)` in `ai_executor.rs`, matching the schema, so absent and on mean the same thing. See **Upgrade notes** — this reaches more existing agents than it looks like |
| Nested agent used as a tool | Its own configuration no longer reads as AI-filled; only `user_message` is, matching `AI_AGENT_TOOL_SCHEMA` on the server |
| Failed-result badge | No longer shown on a step that never ran |
| Test form | Hides unset agent config, the same rule the add-field menu uses |
| Save to workspace | Was building an invalid script path from the editor's synthetic flow path |

### Shared code

- The graph's module create/insert moved to `flowStateUtils` and both callers use it.
- `InputTransformPickers.svelte` + `inputTransformEnv.svelte.ts` hold the variable-picker plumbing and the S3-configured lookup that `InputTransformSchemaForm` and the agent form both need (~120 duplicated lines removed).
- `refreshFlowStateStore` is gone; it only wrapped `reanchorAgentEditsAcross`, whose mechanism this PR deletes.

## Screenshots

**The standalone agent editor** — grouped form, tool roster, test panel, version badge.

01-agent-editor

**A tool, in a drawer beside the agent.**

02-tool-drawer

**Add a field**, grouped, each stating what leaving it unset does.

03-add-a-field

**The same form as a flow step.**

04-flow-step-form

## Test plan

Needs one `ai_agent` resource with a couple of script tools, and a flow with a step linked to it.

**The form**
- [x] Open an AI agent step. Fields are grouped, and only the ones actually set are shown.
- [x] Add a field from the menu, then remove it with the `x`. Removing leaves no diff behind.
- [x] Empty a text field without leaving the step. Its row stays.
- [x] Open a step and close it without typing. The flow is **not** marked dirty.
- [ ] Open an agent that never set `output_type`. Temperature and memory keep their values (they used to be wiped on load).
- [x] Set `output_type` to `image`. Temperature and memory clear, deliberately: memory is ignored for image output, and temperature is still forwarded to the request, so leaving a hidden value would change the run.

**The draft**
- [x] Edit a linked agent from its step. The step's `agent` stays set throughout, and the unsaved-changes banner appears.
- [x] Close the editor, reload the page, reopen. The edits are still there. This is the thing the fork could not do.
- [x] Discard. The form returns to the deployed value and the banner clears.
- [x] Deploy. The version badge advances, the banner clears, and the host graph still shows the agent's tools.
- [x] With a draft outstanding, open Compare & Deploy. The agent draft is listed and deploys.
- [x] History → restore an earlier version. The editor closes, as the generic resource editor does. Reopen it: the restored config is what shows, with no unsaved-changes banner.
- [ ] Open an agent whose resource carries a key the form does not render (a `user_message` default, say). Opening it must **not** put a `*` on the resources row, and a deploy must leave that key untouched.

**The editor**
- [x] Open the same agent from the resources page, and from `#/resource/<path>`. Same editor, same draft.
- [x] Open a tool. It opens in a drawer titled by the tool's name, over the agent.
- [x] Add a tool, and delete one.
- [x] On a tool input, confirm only AI and static are offered, with no connect button.
- [x] Run the agent from the test panel on the right.
- [x] Open Evals from the header, run against unsaved edits, confirm the run records as `agent_draft`.
- [x] `...` menu on a tool → **Save to workspace**. The script lands at `<agent path>/<tool id>`.
- [x] Resources page, `ai_agent` row → **Edit as JSON**. Opens the JSON editor and does **not** create a draft.

**Regression**
- [x] A non-agent resource still opens the ordinary form, with no "Edit as JSON" item.
- [x] Variable picker (`$`) still works on an ordinary flow step's inputs.
- [x] Unlink still folds the agent's config into the step.

## Upgrade notes

**Agents that never wrote a `streaming` key now stream.** That is a wider set than "new agents". A step materialises the key only when its Inputs tab is opened in the flow editor, and `InputTransform::Static` carries `#[serde(default = "default_null")]`, so the seeded `{"type":"static"}` placeholder reaches the worker as `None` exactly like an absent key. Agents with no effective key therefore include: flows pushed by the CLI or git-sync, created through the REST API, or imported from the hub; flows built by the copilot; every saved `ai_agent` resource (`inputTransformsToAgentConfig` omits unset keys); every nested agent tool (`createAiAgentTool` seeds a placeholder for each non-seeded key); chat-mode auto-created agents; and any step whose Inputs tab was never opened.

What does **not** change is the request to the provider: `stream` is already hardcoded true for text output in every provider, and only `StreamEventProcessor::new()` vs `new_silent()` differed. What does change:

- the job result gains a `wm_stream` field holding the newline-delimited event log, including `ToolResult` events with their full tool output embedded. Downstream steps reading `results.<id>`, and webhook or API consumers of a flow ending in an agent, will see it.
- each run writes one row per streaming event into `job_result_stream_v2`.

Set `streaming` to `false` explicitly on a step to keep the old behaviour.

**Anyone mid-edit in an "Edit saved agent" session at upgrade time** gets an unlinked step with the agent's config inlined — the session was in-memory only and its store is deleted here. Nothing is lost; the step must be re-linked by hand.

## Notes

- **Deliberately out of scope**, each its own follow-up: **renaming an agent**, which is a whole story of its own and stays entirely outside this PR. The editor offers no path field, and the two pre-existing gaps a rename leaves are untouched here: linked steps keep pointing at the old `value.agent`, and `eval_experiment.subject` (a plain string with no FK) strands the agent's runs. Renaming stays reachable exactly as it was, through the generic resource editor and the API. Also out of scope, both about nesting: linking a saved agent as another agent's tool, and tools belonging to a tool that is itself an agent. The editor adds an agent as a tool, and that agent stays part of the one being edited: it offers no *Save as reusable agent* there, and one linked from a flow is shown but not openable, so a second editor over a second draft is never reachable from inside the first. It also gives such an agent no tools of its own (its roster has no add control without a graph behind it), and a grandchild authored elsewhere cannot be opened here — the drawer addresses one tool, not a path of them.
- **Parity, not a gap**: an agent in another user's `u/` folder that you can edit through `extra_perms` cannot be *run* from the editor. `require_path_read_access_for_preview` authorizes a preview by namespace and never reads the item's ACL, so it rejects the path whatever lives there. `ScriptEditor` previews under the script's own path and behaves identically; the editor previews under the agent's path for the same reason, and so that a test run and its history name the same thing.
- **Known, narrow**: after *Save to workspace*, Discard clears the draft and leaves the deployed value untouched, but the editor keeps showing the converted tool until a reload. A plain field edit discards correctly.
- **Pre-existing, untouched**: the generic `ResourceForm` materialises schema defaults into an existing resource value, so merely opening any resource in it creates a draft. This is why `ai_agent` rows open the agent editor, and why the escape hatch opens JSON rather than that form.


hugocasa:
Re the `noDeployed` dead-code nit: it is reachable — the AI chat writes resource drafts that carry `resource_type`.

`createResourceToDraftState` (`copilot/chat/global/core.ts:4619`) sets `resource_type: args.resource_type ?? base?.resource_type`, and `RESOURCE_SPEC` / `writeResourceDraft` (`:5082-5093`) persist it through `writeDraft(..., 'resource', ...)` — the same draft kind and row `useAgentDraft` reads. So asking the chat to create an `ai_agent` produces a draft-only agent whose draft names its type, which is exactly what the `noDeployed` branch serves.

Walked it on a local instance: a draft-only `ai_agent` opens in the agent editor with Deploy (and no Evals, per the `draftOnly` gate), and deploying it creates the resource with `resource_type = ai_agent`. The "cannot bootstrap" step in the reasoning holds only if this module is the sole writer of `resource_type` into a resource draft, and it isn't.

Leaving the branch, the accessor and the Evals gate as live code.

hugocasa:
Re the deploy-time type check remaining racy — correct, and not being fixed in this PR.

The window is now one HTTP round-trip between the GET and the update, and exploiting it needs another writer to delete the path and recreate it as a different resource type inside that window. More to the point, this is not a regression: the save path this editor replaces did exactly the same thing, non-atomically — `existsResource` → `getResource` → `updateResourceValue` as three separate calls (`AgentResourceBar.svelte` on main, comment: *"the drawer's path check is debounced, so a fast save can reach here with an unrelated resource at the path — never clobber a resource of another type"*). The pre-check restores parity with that; it does not claim to close the race, and the code comment says so.

Closing it properly means an `expected_resource_type` on the update endpoint: a field on `EditResource`, a conditional `and_where_eq("resource_type", …)` in `update_resource` (`windmill-store/src/resources.rs`), plus `openapi.yaml` and a client regen. That is a backend/API change that applies to every resource editor, not just this one, and it wants its own PR and review rather than riding along in a frontend change.

Deferring deliberately. Claude and pi are both "Good to merge" on this head.

hugocasa:
✅ Review round clean @ 3b5a9e758cb0b0d992590feba402aa641289f483 — Pi "Good to merge"; Codex's remaining P1 (duplicate-id row identity) dismissed below.

Codex and cubic both flag that with two tools sharing an `id`, selection opens the first match and delete removes both. Dismissing it: the condition is unreachable through the product, and where it can be reached by hand, main behaves the same way.

Tool ids come from `insertAgentTool` → `insertNewModuleAtIndex` → `emptyModule` → `nextId(flowState, flowStore.val)`. `nextId` builds `takenIds` from `dfs(fullFlow.value.modules, fm => fm.id)`, and `dfs` descends into `value.tools` for an `aiagent` module (`dfs.ts:29-33`), so every existing tool id is taken — MCP and websearch tools included, which is the case that matters since those get no `flowState` entry. The roster has no duplicate action, and `copyId` (graph-only) uses the same `dfs`. There is no UI path that mints a colliding id.

The only way in is authoring the resource value by hand: "Edit as JSON", or a resource update that repeats an id. On that input the flow editor is already ambiguous in the same direction — `removeFlowModule` resolves through `findFlowNode`, which takes the first match, and every module lookup in the editor is first-match-by-id.

The failure originally raised in this thread — a duplicate id throwing `each_key_duplicate` and leaving the editor unopenable — is fixed by the position key in 3b5a9e758c. What remains is resolution order on deliberately malformed data, which is bounded and non-destructive beyond the entries that share an id.

PR #10883, opened 2026-08-28
fix(cli): stop the migration guard misreading its own git history

hugocasa:
## Summary

`gitRecordedDatatableMigrationPaths` answers "has this branch ever tracked this data table migration?" — a push keeps a migration on the remote when the answer is no, on the grounds that this clone may simply never have synced it. Two ways it answered wrongly:

- **`core.quotePath` is on by default**, so a path holding a non-ASCII byte comes back C-quoted — `"migrations/datatable/dt/1_caf\303\251.up.sql"` — matching nothing the caller holds, and reading as never recorded. A migration the user committed and deleted would be kept instead of deployed.
- **`git log HEAD` fails on a repository with no commits**, which fell into the generic `its history could not be read` / `Check that git runs correctly in this directory`. Git runs fine; nothing is committed yet, and that is now what it says (with `Commit the migrations you sync` as the remedy).

Both are one-line-ish fixes inside the existing function. No new callers, no behaviour change beyond the two wrong answers.

## Where this came from

Found while building windmill-labs/windmill#10851 on top of this function. That PR no longer touches it, so these are split out rather than shipped as unrelated scope.

## Test plan

- [x] New `cli/test/git_recorded_migrations_unit.test.ts` runs real git in a temp repo — the answer is a property of git's output, not of logic testable without it.
- [x] Both new cases **fail against `main`** and pass with the fix; the third (a committed migration stays recorded after its deletion is committed) guards the pre-existing behaviour they must not break.
- [x] `bunx tsc --noEmit` clean for `utils/git.ts`, `UNIT_ONLY=1 bun test test/*_unit*` 1098 pass / 0 fail.

---
## Summary by cubic
Fixes `gitRecordedDatatableMigrationPaths` misreading its own git history so it no longer treats migration paths git would C-quote (non-ASCII, `"`, `\`, control characters) as untracked and no longer reports a repo with no commits as broken git. Previously paths were parsed from newline-separated output under the default `core.quotePath`, and `git log HEAD` failed on empty repos; now paths are read raw NUL-separated via `-z` and an empty repo returns "no commits yet" with the remedy to commit. Adds temp-repo unit tests covering both cases and the existing behavior.

<sup>Written for commit 16a13d510ec95f60de0e21718962b457fa467180. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/src/utils/git.ts
+++ b/cli/src/utils/git.ts
@@ -170,20 +170,12 @@ export function gitRecordedDatatableMigrationPaths(): RecordedMigrationPaths {
     };
   }
 
-  // `core.quotePath` (on by default) C-quotes any path with a non-ASCII byte, which
-  // matches nothing the caller holds and reads as "never tracked".
+  // `-z`, not `core.quotePath=false`: that setting only stops git C-quoting non-ASCII
+  // bytes, leaving a path holding `"`, `\` or a control character quoted and matching
+  // nothing the caller holds. `-z` emits every path raw, NUL-separated.
   const r = spawnSync(
     "git",
-    [
-      "-c",
-      "core.quotePath=false",
-      "log",
-      "HEAD",
-      "--format=",
-      "--name-only",
-      "--",
-      "migrations/datatable",
-    ],
+    ["log", "HEAD", "--format=", "--name-only", "-z", "--", "migrations/datatable"],
     { encoding: "utf8", stdio: "pipe", maxBuffer: 64 * 1024 * 1024 },
   );
   if ((r.status ?? 1) !== 0) {
@@ -195,7 +187,7 @@ export function gitRecordedDatatableMigrationPaths(): RecordedMigrationPaths {
     };
   }
   const paths = new Set<string>();
-  for (const line of (r.stdout ?? "").split("\n")) {
+  for (const line of (r.stdout ?? "").split("\0")) {
     const p = line.trim();
     if (p.length === 0) continue;
     if (prefix.length > 0) {
--- a/cli/src/utils/git.ts
+++ b/cli/src/utils/git.ts
@@ -156,9 +156,34 @@ export function gitRecordedDatatableMigrationPaths(): RecordedMigrationPaths {
   }
   const prefix = (prefixOut.stdout ?? "").trim();
 
+  // `git log HEAD` fails on a repository with no commits, which is not the same as
+  // git being broken and must not be reported as such.
+  const head = spawnSync("git", ["rev-parse", "--verify", "--quiet", "HEAD"], {
+    encoding: "utf8",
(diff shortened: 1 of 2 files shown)


hugocasa:
Both addressed.

**@claude — the quoting fix was one line short, and you're right.** I checked it in a scratch repo:

```
$ git -c core.quotePath=false log HEAD --format= --name-only -- migrations/datatable
migrations/datatable/dt/1_café.up.sql
"migrations/datatable/dt/2_a\"b.up.sql"
"migrations/datatable/dt/3_a\\\\b.up.sql"
```

`core.quotePath=false` only covers non-ASCII; `"` and `\` stay C-quoted and keep reading as never-recorded. Switched to `-z` (raw, NUL-separated), which covers all three, and split the parser on NUL. The test now commits all three shapes and still fails against `main`.

**@codex — on the third test**, I'd argue it's no longer re-exercising unchanged behaviour: the `-z` change moved the parser from newline-split to NUL-split, so "a committed migration stays recorded after its deletion is committed" is what catches a parser that silently returns nothing. It was fair when you raised it against the `core.quotePath` version, which didn't touch parsing. I've added a comment saying what it guards so it doesn't read as ceremony.

hugocasa:
Closing — not worth carrying.

Both fixes are correct and tested, but neither is reachable enough to justify the review cycles:

- **The quoting fix** can't be produced by the product at all. `wmill datatable migrate new` rejects any name outside `^[a-zA-Z0-9_-]+$`, and the backend rejects a data table name outside ASCII alphanumerics plus `_ - .` — with an error that says explicitly it's so migrations can be synced to a git repo. Only a hand-written migration file gets you a path git would quote (`parseDatatableMigrationPath` accepts any name via `(.*)`).
- **The no-commits diagnosis** only mislabels a message. The guard already keeps the migrations, which is the safe outcome; the sentence just sends you to debug git when the fix is `git commit`. The normal bootstrap order (`wmill sync pull` before `push`) avoids it, and a fresh clone always has commits, so CI never sees it.

Recorded here in case either becomes real: the `-z` fix (not `core.quotePath=false`, which only covers non-ASCII and leaves `"` and `\` quoted) and the `rev-parse --verify HEAD` check are on branch `hugo/git-980-datatable-migration-history-fixes`, with a temp-repo test that fails against `main`.

PR #10884, opened 2026-08-28
feat: rework the evals dataset drawer and run navigation

Guilhem-lm:
## Summary

A UX pass over the AI evals dialog, plus two reusable pieces it needed. Nothing
here changes the backend or the API — it is entirely `frontend/`.

The dialog now behaves like a two-level surface: the runs list and one run's cells
are pages that slide, each carrying its own controls, and the level you are on is
never rebuilt while you are moving to it. Both levels answer the keyboard.

## Changes

**Dataset creation and the run dialog**
- The empty runs table always offers **New evaluation**, even with no dataset. Creating
  a dataset happens inside the run dialog, which already handled the no-dataset case.
- The dataset drawer opens *over* the run dialog instead of replacing it, so the
  dialog keeps its state. `stepAside()` and the `resumeRunDialog` bookkeeping are gone.
- Starting a run leaves you on the list instead of jumping into an empty results grid.
  The new row fills in live on the existing 2s poll.

**Reading a run in progress**
- The Cases column shows `answered/total` while a run is going, read from the flow's
  `flow_jobs_success` via `getJobUpdates` — no new endpoint, and no per-cell fetch.
- One spinner per row, on the case count. The per-scorer spinners are gone: they read
  as "this judge is slow" when what is unfinished is the run.

**Paged dialog (new reusable component)**
- `common/modal/PagedContent.svelte`: pages laid over each other, mounted on first
  visit and never unmounted, so a page keeps its state, its loaded rows and its scroll
  position. Opt-in `warm` builds every page one frame after mount, which keeps the
  first navigation from paying for a mount inside the transition.
- The outgoing page is dropped at once and the incoming one slides and fades in;
  a symmetric fade leaves the old page ghosted over the new one for the whole slide.
- `Modal` gained an opt-in `paginated` and a breadcrumb fade. Both default off — of
  25 `Modal` callers, `AgentEvalModal` is the only one passing `trail` or `description`.

**Keyboard navigation**
- `←` / `→` step between the two levels; `↑` / `↓` move through the runs, and `Enter`
  or `→` opens the highlighted one.
- One highlight serves the pointer and the keyboard, the way a melt menu does it:
  hovering a row moves the highlight there, so the arrows carry on from where the
  pointer left off rather than running a second, invisible cursor. It wears
  `bg-surface-hover` — nothing is *chosen* until Enter, so it must not read as a
  selection.
- The highlight is tracked by run id, not row index: the list is newest-first and the
  poll prepends to it, so an index would come to mean a different run mid-keypress.
- A chevron appears on the highlighted row, which also separates "open this run" from
  the row's own dataset button.
- `Enter` still belongs to whatever is focused if that thing uses it, so a highlighted
  row does not swallow the press on **New evaluation** or a row's dataset button.
- Focus follows the page: `inert` on the page being left drops focus on the body, so the
  arriving page takes it. Focus is only ever taken from the body or from inside the pages —
  never off dialog chrome the user is actually on.
- Both handlers listen at `window`, so they ask two questions the DOM cannot: is my
  overlay host the visible one (session preview tabs stay mounted when hidden), and is
  my surface still on top rather than under a drawer opened since. The second is a new
  `topmostSurface` context that `Modal` and `Drawer` both provide.

**Editing cases**
- `EvalCasesGrid` moved off ag-grid onto `DataTable` + a new
  `common/EditableTextarea.svelte` (the multiline counterpart to `EditableInput`).
  This drops an ag-grid instance and its bespoke cell editor from this surface.
- Cases paginate at 25/page, since a dataset holds up to 1000 and every row mounts two
  editors — ag-grid virtualised, `DataTable` paginates.
- Adding a case focuses its Question; a dataset cannot be *created* with no cases,
  and the empty table says so with an accent CTA.
- `TextInput` gained an opt-in `autosizeParams`, forwarded to the `autosize` action.

**Styling**
- Scorers render as cards matching `RadioCard`'s layout; the cases table is a raised
  surface sized to its rows; the case side panel animates with `AnimatedPane` and is
  laid out as cards rather than rules.
- The case result renders through the same prose stack as a chat answer
  (`markdownProse.sm`) instead of `GfmMarkdown`'s legacy `prose-xs` fallback.

## Screenshots

pr-01-runs-list

pr-02-dataset-drawer

pr-03-empty-cases

## Test plan

- [ ] Open Evals on an `ai_agent` resource with no dataset — the table offers
      **New evaluation**, and the run dialog offers **New dataset** without closing.
- [ ] Create a dataset: **Create dataset** stays disabled until a case exists, and
      **Add a case** focuses the Question cell.
- [ ] Type a multi-line question — the cell grows with it, and clicking in and out
      does not shift the text by a pixel.
- [ ] Clear a case's Expected value and save — the payload must omit `expected`
      entirely, not send `""`. (`""` makes the default script scorer score a hard 0
      instead of excluding the case.)
- [ ] Add more than 25 cases and page through them; a case added to a full page still
      opens for typing.
- [ ] Click into a run: the slide starts on the click frame, the first open is no
      slower than the rest, and a run of a *different* dataset never shows the
      previous dataset's questions or scorer columns.
- [ ] `↑` `↓` move the highlight, hovering a row takes it over, `→` and `Enter` open
      the highlighted run, `←` returns. Pick a non-default **Compare to**, go back to
      the list and reopen the same run — the comparison must survive.
- [ ] With the dataset drawer open over the dialog, arrows must not steer the list
      underneath, and `Escape` in a case cell must leave the edit without closing the
      drawer.
- [ ] Open a case: the right panel slides rather than appearing, and a markdown answer
      renders like a chat answer.

## Known gaps

- `Modal` has no focus trap of its own, so `Tab` from the last control in a page can still
  leave the dialog. Pre-existing, and not something this PR narrows or widens.
- `EditableTextarea` is close to a fork of `EditableInput`; a `multiline` prop on the
  latter would have avoided two copies of the same state machine. They already
  diverge (`edit()` defers 50ms there, one frame here; Escape stops propagation here).
- Inside a `Modal`, `EditableTextarea`'s Enter belongs to the dialog unless the caller
  passes `enterConfirms={false}` — documented on the component rather than enforced.


Guilhem-lm:
✅ Review round clean @ 03b880c968fe270f386fa24e625c31cf958f16a3

Guilhem-lm:
✅ Review round clean @ 70327a21910563b6205bba87724662ed12d36741

PR #10885, opened 2026-08-28
fix: keep raw-app editor selection consistent across sidebar and tabs

Guilhem-lm:
## Summary

In the raw-app editor, opening a backend runnable from the tab bar left the frontend file
highlighted too, so the sidebar showed two rows selected at once.

The cause: `activeTabId`, `selectedRunnable` and `selectedDocument` were three independent pieces
of state written from a dozen places, and each writer had to remember to clear the others.
`activateTab` cleared `selectedRunnable` when a file tab opened but not `selectedDocument` when a
runnable tab did. This PR removes that class of bug by making one function the only writer of all
three, so they cannot disagree about what is selected.

Two more desyncs fell out of the same root cause and are fixed here:

- Toggling split-with-preview while Preview was active moved the active tab to the last
  file/runnable without setting the selection, leaving a runnable tab active over an **empty
  editor pane**.
- Deleting or renaming a runnable left its tab open and active over an empty pane. The stale-tab
  cleanup read `void runnables`, which does not observe `delete runnables[id]` on a `$state`
  proxy — enumerating the keys does.

Folders are also no longer selectable in the file tree. A folder can't be opened in an editor, so
selecting one highlighted a row nothing on screen corresponded to — the same "selected thing we
can't show" problem one level down. Clicking a folder now only expands it.

### The selection funnel

- `select(next: EditorSelection, opts)` in `RawAppEditor.svelte` is the single writer of
  `activeTabId`, `selectedRunnable` and `selectedDocument`. `activateTab` is a thin wrapper that
  maps a tab id to a selection. Every caller goes through it: both tab bars, `closeTab`'s
  fallback, `toggleSplit`, the sidebar, the chat tools (`setFrontendFile` / `setBackendRunnable`),
  the UI Builder iframe's `setActiveDocument` message, and the mount-time default file.
- Deleted the `$effect` that mirrored `selectedRunnable` into the tab system — the sidebar now
  calls `select` directly, so there is no assign-and-wait-for-the-effect step.
- Split `iframeDocument` out of `selectedDocument`. `selectedDocument` is now purely what is
  selected; `iframeDocument` tracks what document the UI Builder iframe has open, which has to
  outlive switching to a runnable so a reload reopens it.
- Sidebar props are one-way (value + callback) instead of `bind:`, per `frontend/AGENTS.md`. A
  child can no longer write half the selection.
- The iframe's `setActiveDocument` follows the document switch only while a file tab is the
  visible surface; behind a runnable tab it backfills a tab without stealing the selection.

### One existence predicate

`wmill.ts` is generated inside the iframe from the runnables' dts: the sidebar lists it and the
iframe can open it, but it is never a key of `files`. Three separate sites needed to answer "does
this document still exist?", and a plain `files[path] !== undefined` test silently discards it —
which closed the user's `wmill.ts` tab.

- `isOpenableDocument(path)` is now that single test, shared by the stale-tab cleanup,
  `populateFiles` (replaying the open document after a files change) and `applyEntry`
  (history undo/redo).
- `WMILL_TS_PATH` moved to `raw_apps/utils.ts` so the sidebar's `extraNodes` entry and the
  predicate can't drift apart.

### File tree

- `FileTreeNode`: clicking a folder toggles expansion only. Folder rows gain "New file" and "New
  folder" menu entries, wired to the pre-existing but never-invoked `onAddFile`/`onAddFolder`
  props — folder selection used to be the only way to seed an empty folder.
- `FileExplorer`: the add-file anchor is one `parentFolderOfSelection()` helper, and neither
  `handleRename` nor `handleDelete` reports a folder path as the new selection any more.
- Renaming a folder remaps the selection by prefix (`old/a.ts` → `new/a.ts`). Without it,
  suppressing the folder path left `FilesetEditor` editing a key that no longer exists.

### Behaviour change worth flagging

"Create a folder, then press + file" used to nest the new file inside that folder, because the
folder became the selection. It now creates next to the selected file (or at the root), and the
folder row's own "New file" is the way to nest. This applies to the raw-app sidebar and to the
fileset editor in resource forms, which shares `FileExplorer`.

## Screenshots

Before — a file tab, then the runnable tab: both `App.tsx` and `a` highlighted.

before

After — the same sequence: only `a` is highlighted.

after

After — a folder click only expands it, and the folder row carries its own create actions.

folder-menu

## Test plan

Exercised in a browser against a raw app in a scratch workspace:

- [x] Sidebar runnable → file tab → runnable tab → sidebar file: exactly one row highlighted at
      every step, and the active tab always matches the pane content.
- [x] Create a runnable, rename the selected runnable, delete the selected runnable: the tab
      follows the rename, and the delete closes the tab and falls back to a real selection instead
      of an empty pane.
- [x] Rename a file in the tree: the old tab closes, the new one is active and selected.
- [x] Preview active in single-pane mode clears both highlights; toggling back to split opens the
      last file/runnable with its content.
- [x] With a runnable tab active, the UI Builder reporting a document switch backfills a file tab
      but leaves the runnable selected.
- [x] With a runnable tab active, a files change still replays `setFilesAndSelect` for the
      document the iframe has open (the `iframeDocument` split).
- [x] `wmill.ts` open, then add / rename / delete a runnable: the tab stays open and selected
      (it is not a key of `files`, so the old cleanup closed it).
- [x] `wmill.ts` open, then undo/redo through history: the replay reopens it instead of dropping
      the document.
- [x] Folder click expands/collapses only; creating a folder does not move the selection; the
      folder menu's "New file" creates inside it, expands it and selects the new file.
- [x] Rename a folder holding the selected file: the selection follows to the new path.
- [x] Same folder checks in the fileset editor of a resource form, which shares `FileExplorer`.

No automated tests: frontend-only Svelte components, and the change extracts no pure-logic utility
(`select` reads component state directly), so per `REVIEW.md` there is nothing here the repo tests.


Guilhem-lm:
✅ Review round clean @ 853e84f2cf36b1fbb5311aa5482de00b04b15779

Guilhem-lm:
✅ Review round clean @ b9954eb60a18dab4c5209fc51b052596c31ff36b — Good to merge from codex, claude and pi at 9532c44322ca8afd68862b27f188b2694aa3048e; cubic's remaining P2 (optional `onDelete` making the row's Delete a silent no-op) addressed in b9954eb60a by making the callback required.

PR #10891, opened 2026-08-29
fix: register every rotated service log file exactly once

rubenfiszel:
## Summary

Service log files are registered in the `log_file` table by `send_logs_to_object_store` in `backend/src/monitor.rs`. Some minutes were silently never registered, and both search engines read their work list from that table — the tantivy count index and the Parquet retrieval store alike — so those log lines were permanently unsearchable.

This is a pre-existing bug, not a regression: it predates the Parquet store work and affects search coverage on every deployment. This PR is the first time the file-selection logic moves.

### Cause

`find_two_highest_files` never computed the two highest files. It only remembered the previous running maximum:

```rust
if file_name > highest_file.clone().unwrap_or_default() {
    second_highest_file = highest_file;   // only updated when a new max appears
    highest_file = Some(file_name);
}
```

So the "second highest" was really the largest name read *before* the global maximum's position in the directory listing — and readdir order is filesystem-dependent, not sorted. Which minutes got registered came down to where the newest entry happened to land:

- **tmpfs** (the default `WINDMILL_DIR=/tmp/windmill`) returns the newest entry first, so the global max is read at position 0 and the second-highest is `None` on **every** tick: not a single minute is ever registered.
- **ext4** hashes names. Simulating an hour of the rotate-and-prune pattern, 28 of 60 ticks selected the wrong file; a live 10-minute run lost 5 of 10 rotated minutes.

Because `LAST_LOG_FILE_SENT` is a monotonic watermark, a minute walked past is skipped permanently. Restarts added a second hole: the watermark reset to `None`, so files a killed process left behind were never registered, and the graceful-shutdown flush stamped the still-open file with `Utc::now()` (seconds included) rather than the file's own minute.

## Changes

- Sort the directory listing by the minute parsed out of each filename instead of scanning for a running maximum, so readdir order no longer decides which files are seen. Names that do not parse are dropped rather than falling back to `Utc::now()`, which could poison the watermark.
- Send every file below the newest one that is past the watermark, oldest first, so a multi-file backlog drains instead of only the single second-highest file.
- Stop at the first failure rather than moving on: both indexers walk `log_file` with a `log_ts > watermark` cursor, so a row that lands after a newer one is never picked up by either.
- Seed the watermark at startup from `log_file`, so files left behind by a killed run are recovered instead of staying unsearchable. The newest registered minute is deliberately excluded: the shutdown flush registers the file that is still open, and `tracing-appender` reopens that minute in `append` mode, so a restart inside it would otherwise strand everything written afterwards. Re-sending restores the full object and sums the counters (the additive upsert adds only the lines this run counted for that minute); see the known limitation below for what it does not restore.
- Serialize the periodic uploader against the shutdown flush with a mutex. Both now walk the same files, so without it both could clear the watermark for one file and count its lines twice.
- Bound the shutdown flush at 15s, matching the pool close that follows, so a backlog against a slow object store cannot hold the process past its termination grace period. Anything left over is registered by the next run's catch-up.
- The shutdown flush now uses each file's own parsed minute and covers every pending file, not just the current one.
- Skip files past `SERVICE_LOG_RETENTION_SECS` during catch-up: `delete_expired_items` deletes `log_file` rows past that same cutoff, so a run coming back from a long outage would otherwise register rows its own cleanup drops on the next pass, after both indexers had paid to parse them.

## Known limitation (pre-existing, not introduced here)

Both indexers walk `log_file` with a **single global** `log_ts >` cursor (`service_logs_ee.rs:380` — no `hostname` predicate), and registration upserts on `(hostname, log_ts)`. So a row rewritten in place, or a crash-recovery row inserted below a cursor another host already advanced, is never re-read: the minute reaches the object store and the service-logs listing but not full-text search.

This is the indexers' cursor design, not something this PR introduces — on the same-minute restart `main` produces a *second* row at `Utc::now()` second precision that the cursor also skips, and on multi-node crash recovery `main` registers nothing at all. Closing it needs a cursor-visible version column plus changes to both indexers (EE), which belongs in a follow-up rather than in a file-selection fix. Discussion: https://github.com/windmill-labs/windmill/pull/10891#issuecomment-5461605607

Catch-up rows also carry `ok_lines = 0` — a freshly started process counted none of a previous run's lines — which only flattens those bars in the service-logs histogram. Content is unaffected: the indexer parses the object itself rather than trusting the counts.

## Test plan

Unit tests (`cargo test --bin windmill log_file_listing`) pin the ordering: a listing handed back newest-first (what tmpfs actually does) still yields every file ascending, the newest is held back as still-being-written, and non-log names are dropped.

End-to-end on `MODE=standalone+search` with an isolated `WINDMILL_DIR` and a filesystem instance object store, comparing `select file_path from log_file order by log_ts` against the files on disk:

- [x] **Before**: over 10 minutes on ext4, 5 of 10 rotated minutes missing from both `log_file` and the object store (`08-28`, `08-31`, `08-34`, `08-36`, `08-37`). On tmpfs, nothing is ever registered.
- [x] **After, ext4**: 19 rotated files over 21 minutes, zero missing.
- [x] **After, tmpfs**: 19 rotated files over 21 minutes, zero missing.
- [x] **Crash recovery**: `kill -9` mid-minute; the file the dead process was writing is registered on the first tick after restart. The minute during which the process was down has no file and is correctly absent.
- [x] **Same-minute restart**: graceful stop at `09:19:07`, restart at `09:19:24`. The reopened file is re-sent complete — the object-store copy goes from 7,485 to 77,421 bytes — and `ok_lines` sums to 2248 = 276 (pre-shutdown) + 1972 (post-restart), with no double counting.
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`

No frontend changes.

---
## Summary by cubic
Fixes service log file registration so every rotated log minute is registered exactly once, closing permanent search gaps that previously depended on filesystem readdir order.

- Sorts rotated log files by the minute in their name instead of scanning for a running maximum.
- Registers every file below the newest one, oldest first, so a backlog drains in order.
- Seeds the sent watermark from `log_file` at startup to recover files left behind by a killed run.
- Skips files past the retention cutoff on catch-up so a backlog from a long outage isn't re-uploaded.
- Serializes the periodic uploader and shutdown flush with a mutex so the same file is never counted twice.
- Bounds the shutdown flush at 15s and uses each file's own parsed minute instead of `Utc::now()`.
- Drops filenames that don't parse as log minutes rather than poisoning the watermark.

<sup>Written for commit 1f8a527687a405422f97f0cb7b7b25de5a93a829. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10891?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
## On the Codex P1 — "replayed log files remain invisible after the indexer advances"

The mechanism is real and I verified it directly: `IndexedLogTrackerData.max_log_ts` is a **single global** cursor persisted to the object store, and the work-list query is `FROM log_file WHERE log_ts > $1 ORDER BY log_ts ASC LIMIT $2` with no `hostname` predicate (`windmill-indexer/src/service_logs_ee.rs:380`). Registration is `ON CONFLICT (hostname, log_ts) DO UPDATE`, so a row rewritten in place is never revisited. No argument there.

What I dispute is that it is a defect *in this PR*. On every scenario the finding names, this branch is equal to or better than `main`:

| scenario | `main` | this branch |
|---|---|---|
| same-minute restart: shutdown flush registers | `(host, 09:19:07)` — `Utc::now()`, second precision, matching no file's minute | `(host, M)` — the file's own minute |
| after restart, minute `M` is registered as | a **second** row `(host, 09:19:00)`, *below* the `09:19:07` the indexer consumed | the same row upserted in place, *equal to* the cursor |
| post-restart tail indexed? | **no** | **no** |
| multi-node crash recovery | rows never inserted at all — not in the object store, not listed, not retrievable | object uploaded and row listed/retrievable via `get_log_file`; only full-text search is missed |

So the residual is the indexers' global-cursor design — a row rewritten in place is never re-read — which predates this change. The cold Claude pass reached the same conclusion independently and did not treat it as blocking.

**Why I am not closing it here.** The two ways to close it are both worse or out of scope:

1. *Cursor-visible version / invalidation* (what the finding asks for). That is a `log_file` schema addition plus changes to both indexers in `windmill-ee-private` — a separate change with an EE companion PR, not a file-selection fix.
2. *Never replay* — i.e. don't register the still-open file at shutdown. That loses the final partial minute of **every terminating pod**, which in k8s is every deploy (a new pod gets a new hostname and a fresh `WINDMILL_DIR`, so no later run ever picks it up). Strictly worse than a minute that is stored but unindexed, and it would re-open the P1 raised against the first revision of this branch.

Meanwhile the bug being fixed is total: on tmpfs — the default `WINDMILL_DIR=/tmp/windmill` — `second_highest` is `None` on *every* tick, so **not one minute is ever registered**; on ext4 a live run lost 5 of 10 rotated minutes. Holding that behind an indexer redesign trades a systematic, verified data-loss fix for a narrow residual that this branch already improves.

**What I did change in response** (`1f8a527`):
- `send_current_log_file_to_object_store` → `flush_pending_log_files_to_object_store`, since it now flushes every pending file (the other reviewer's P2).
- The `init_last_log_file_sent` doc comment no longer implies recovery reaches search. It now says the re-send restores the object and sums the counters, and that whether the indexers read it again depends on their single non-per-hostname `log_ts >` cursor.
- Noted that catch-up rows carry zero line counts (this run counted none), which only flattens their bars in the UI.

The cursor work is worth a follow-up issue against the indexer; I'd rather file it there than bolt a version column onto this fix.

PR #10893, opened 2026-08-29
feat: day-partition the service log index and expire whole chunks

rubenfiszel:
## Summary

Replaces the single rolling tantivy index behind the service-log per-host count with **one tantivy index per UTC day**. Expiry becomes an unlink of a whole chunk instead of a `delete_query` whose tombstones sit on disk until a merge happens to pick the segment up, and a day that has rolled over is immutable, so it is merged down to a single segment once and never rewritten.

The substance is in windmill-ee-private#753 (symlinked `*_ee.rs`); this side carries the EE ref and one change in `backend/src/main.rs`.

## What the prize actually is

This started as a measurement, because nobody had measured steady-state bloat — every index size quoted so far came from a single-pass build followed by `wait_merging_threads()`, i.e. a perfectly merged index with zero tombstones.

Simulated 14-day run at 2M docs/day with a 7-day window (tantivy 0.27, 100 hosts, 288 refresh passes/day, 150 MB writer budget, production schema and ingest cadence), sampling the index hourly across the steady-state window:

| | on-disk | vs today |
|---|---|---|
| **today**: one rolling index, soft delete + `LogMergePolicy` | **637 MB** — 10.8% of docs dead on average, peak `du` 1070 MB | — |
| the same live rows, indexed once and naturally merged | 544 MB | −15% |
| day chunks, sealed but not force-merged | 604 MB | −5% |
| **day chunks, force-merged on seal** | **491 MB** | **−23%** |

Priced for whole-day retention rounding — a chunked index holds 7–8 days where the rolling one holds exactly 7, so 7.5 on average — the steady-state saving is **~17%**. Repeating the run at 4× the volume (8M docs/day, 56M live docs, tantivy 0.26) gave the same shape: 2408 MB → 1890 MB.

So the tombstone bloat on its own is ~15%, not 2×. What makes the switch worth it is that most of the remaining win is only available to an immutable chunk: a rolling index can never be force-merged, because the next commit refragments whatever a merge just joined.

Two other measured effects:

- **Object-storage write volume.** `S3BackedMMapDirectory` writes through on every file write, so bytes written to the directory are bytes uploaded: 894 MB/day today against 560 MB/day chunked, because a rolling index keeps re-merging data a sealed day never touches again.
- **The cheap alternative does not work.** tantivy's `del_docs_ratio_before_merge` defaults to 1.0, which means deleted documents never trigger a merge at all. Setting it to 0.3 was measured on the same run and changed nothing: 623 MB vs 619 MB.

Query fan-out is not a cost at this size — the unscoped count over 7 chunks takes 12.8 ms against 14.2 ms for the equivalent single index, because the chunks have fewer segments between them than the rolling index does.

## Merged with #10892 / ee-private#754

That PR moved the ingest cursor out of the index into the columnar store and made an empty index repairable by rebuilding from it. It replaced everything this branch had carried for the same purpose — the v3 watermark adoption, the store-versus-local comparison, the clear-path exemption, the root tracker — so all of it is gone. The transition gap this PR previously documented as a known cost went with it: a layout bump is now just another empty index, which the rebuild fills from parquet.

The rebuild walks the store's own `dt=` partitions, so it opens each day's chunk once, in order, and seals it on the way past.

## Changes

- `backend/ee-repo-ref.txt` → windmill-ee-private#753.
- `backend/src/main.rs`: move the log-indexer writer into the indexer future instead of cloning it. Sealing a chunk takes sole ownership of its tantivy writer, which a second live handle would silently prevent.

## Test plan

- [x] `cargo check --features enterprise,private` and `--features enterprise,private,parquet,tantivy` clean
- [x] `cargo test -p windmill-indexer --features enterprise,private,parquet` — 18 pass, including two new unit tests for the chunk-selection predicates
- [x] Exercised on a local instance (`MODE=standalone+search`, `--features quickjs,private,tantivy`, instance object storage on a local filesystem) over 10 days of backdated seeded log files:
  - a chunk per day locally and mirrored to the object store; every rolled-over day a single segment, today's still open
  - counts correct per day, per range, and in aggregate — the per-day counts sum exactly to the whole-range count
  - a term outside the window returns a well-formed empty bucket list
  - shrinking the window expires whole chunks: objects deleted, directory unlinked, counts drop accordingly
  - **wiping the index with the store cursor left in place rebuilds all seven day chunks from parquet, each sealed to one segment, with counts unchanged**
  - raw log files are deleted once ingested, and the three outside the window are left alone
  - deleting one local chunk and restarting pulls back only that chunk
  - a planted unopenable chunk directory is reclaimed by the expiry sweep
  - restarting on a stale local copy while the store had moved on keeps counts exact rather than doubling
  - the Service Logs page host picker renders from the fan-out and retrieval still resolves

## Note for reviewers

A pre-existing bug spotted while reviewing, not touched here: in `refresh_log_index`, when `store_rows` reaches `LOG_STORE_MAX_BUFFERED_ROWS` the buffer is drained with `std::mem::take` and the `store_rows.last()` immediately after is `None`, so that one line is indexed with empty searchable text. Unchanged from #10886; wants its own fix.

https://claude.ai/code/session_01EeUmYWCJeaaHutiHLfZBQX


rubenfiszel:
Round 4 findings addressed in `fa2f26f` (EE `ebe0031`):

- **[codex P1] day-boundary checkpoints stranded already-ingested raw log files** — fixed. A completed day's `ingested` paths are now deleted at the checkpoint that moves the cursor past them, instead of at the end of the whole batch. Left until the end, a later day failing stranded them: the next pass resumes after the cursor and never sees those files again.

- **[pi P2] rebuild read a narrower window than `expire_chunks` keeps** — fixed, and thank you, this was a real under-report. The rebuild floored at `now - retention` while expiry keeps whole days, so the oldest chunk came back holding only the hours after that instant while the chunk itself stayed alive. It now reads back to that day's midnight. `rebuild_window_secs` is extracted and unit-tested at three times of day, asserting the floor is midnight, is the day expiry keeps, and is never narrower than the retention window. The stale `monitor_log_index` reference in that doc comment is gone too — the trim is `expire_chunks`.

- **[claude nit] `ee-repo-ref.txt` points at an EE branch commit that goes dangling once #753 is squash-merged** — this is the documented EE workflow rather than a defect: `docs/enterprise.md` step 5 has the OSS PR pin its own EE branch head so CI builds against the EE code under review, and the `sync-ee-ref` workflow re-points it at merged EE main afterwards. #10896 was that workflow doing its job, not a repair of a mistake. Pinning merged EE main here instead would build this PR against EE code that does not contain it.

Re-verified end to end on the merged head: index wiped with the store cursor intact rebuilds all seven day chunks from parquet, six sealed to one segment and today's open, counts unchanged at 140/host aggregate and 20/host per day, per-day sums equal the whole-range count. 21 unit tests pass; `cargo check` clean on `enterprise,private` and `enterprise,private,parquet,tantivy`.

rubenfiszel:
Round 5 findings addressed in `0b4b822` (EE `f553810`). Both were real and both were mine.

**[P1] Interrupted empty-index rebuilds persist as complete.** Correct, and the consequence is worse than a stale count: the days the rebuild never reached stay missing permanently. A rebuild seals each day as it walks past, so an interrupted one leaves committed chunks — the index is then non-empty and, the cursor never having moved, not behind either, which is exactly neither rebuild trigger. Upstream's single-writer rebuild committed once at the end, so cancellation left nothing behind; sealing per day is what opened this.

Fixed with a marker written before the first chunk commit and removed after the last, checked as a third rebuild trigger. Reproduced the bug and the fix as an A/B on the same partial index (four of seven days removed from disk *and* object storage, so the chunk pull cannot mask it):

- without the marker: stays at 3 chunks / 60 hits a host, and no later pass repairs it
- with the marker: next pass rebuilds all 7 chunks, 140 hits a host, six sealed to one segment, marker cleared

**[P2] Rebuild cutoff drifts past midnight.** Also correct — `drop_all_chunks()` runs between the two clocks, so the duration landed after the midnight it was meant to be. The floor is an absolute instant now, and `for_each_index_batch` takes the cutoff rather than a window, so nothing between computing it and the scan can move it. `rebuild_floor` is unit-tested at three times of day for midnight alignment, the right day, and never being narrower than retention.

On the test-coverage note: I did not add a cancellation regression test. Driving a real killpill mid-rebuild needs a live tantivy index and an object store, which this suite does not stand up, and the marker's value is in *detection*, which the A/B above exercises directly. The arithmetic half is unit-tested.

rubenfiszel:
Round 6 addressed in `8e4adb1` (EE `84167f7`). All three of you found the same hole and you were right — the marker was local-only, which is the wrong place for it.

The reasoning I missed: a rebuild is interrupted most often by the very thing that also replaces the disk. So the likeliest route into the bug is the one a local marker cannot cover — replica A is cut short mid-rebuild having dropped and partially rewritten the chunks *in the store*, replica B comes up on a fresh disk, pulls the half-written set, finds the index non-empty and the cursor level, and settles on it permanently. Same failure as round 5, reached by the more probable path.

The marker now lives in the object store beside the chunks a rebuild drops and rewrites there. A failed read of it ends the pass rather than being read as absence, for the same reason an unreadable cursor does.

Verified against exactly that scenario: three of seven days plus the marker left in object storage, **nothing on local disk at all** — the replica rebuilds all seven days, counts return to 140 a host, and the marker is cleared. Before this it would have kept the three and reported 60.

`cargo check` clean on `enterprise,private` and `enterprise,private,parquet,tantivy`; 21 unit tests pass.

rubenfiszel:
Round 7 addressed in `6798804` (EE `dc795ac`). Both P1s were real.

**Stale local chunks survive replica handoff.** Correct, and it is a guard I had and deleted: when the cursor moved into the store in #754 I removed the local-vs-store comparison along with the watermark machinery, without noticing that one job of it survived. The local tracker is still written on every checkpoint, so it still knows where *this* replica got to; when it sits behind the store's cursor, every day from its own last position on is stale on this disk and `pull_missing_chunks` will not replace a day it already has. Those days are dropped before anything opens them, so the pull brings the current copies down. Days before that point are sealed and identical either way, so they stay.

Verified as an A/B rather than by inspection: snapshot a replica's index, let the instance index 5 rows of a distinct term into the shared store, then restore the snapshot (tracker behind the cursor) and restart. The new term comes back at 5 hits where the stale chunk had none, and the sealed days are untouched at 140 a host.

**Interrupted explicit clears leave a valid partial index.** Also correct. The request is now reset after the deletion completes rather than before it starts, so an interrupted clear repeats instead of leaving a subset that every trigger reads as finished. Verified a full clear still runs and rebuilds all seven days; I did **not** drive an actual interruption mid-deletion — that needs a fault injection point this code does not have — so that half rests on the ordering change, which is the whole of the fix.

**On the marker-vs-`drop_all_chunks` nit:** it is safe, and by prefix rather than by luck. `drop_all_chunks` deletes `{index_prefix}/dt=YYYY-MM-DD` per day; the marker is `{index_prefix}/.rebuilding`, a sibling of those, not under any of them. It matters on the rebuild-failure path, where `drop_all_chunks` runs *with* the marker already written and has to leave it — which is what makes a failed rebuild retry rather than settle. The happy path writes the marker after the drop, so it is not load-bearing there.

`cargo check` clean on both feature sets; 21 unit tests pass.

rubenfiszel:
Round 8 addressed in `f7468e0` (EE `5bec54d`), and this one changed the shape rather than adding a guard.

**[P1] Missing local trackers leave stale chunks trusted.** Correct, and the detail that matters is the one you added: *a rebuild never writes the tracker*, so chunks-without-a-tracker is the ordinary post-rebuild state, not an anomaly. My round-7 guard returned early on exactly that.

This was the fifth route to the same failure — a partial or stale chunk set, non-empty, level with the cursor, accepted permanently — after the interrupted rebuild, the local-only marker, the handoff, and the interrupted clear. Patching a fifth route would have been the wrong move, so the rule is now stated once and positively:

> With a cursor in the store, a local chunk survives only up to where the local tracker says this replica indexed. With no tracker, none of them do.

and init records what the settled set covers, which is what gives the next start something to vouch with — and what the rebuild path was missing. `pull_missing_chunks` then fills whatever was dropped from the store, which is authoritative because every local write goes through to it.

Verified by handing a **tracker-less** snapshot back after another replica indexed 7 rows of a distinct term into the shared store: all 7 come back, the sealed days stay at 140 a host, and a tracker is present afterwards.

**[P2] Comments over four lines.** Both condensed — the rebuild-marker doc from eleven lines to four, and the clear-ordering comment from five to four.

On the regression test you asked for: the missing-tracker branch is a two-line predicate whose bug was *when it returns early*, and a unit test over it would restate the branch rather than pin behaviour — the thing worth pinning is that a tracker-less replica recovers, which needs an object store and a live index, so I exercised it as the A/B above instead. Happy to be argued out of that.

`cargo check` clean on both feature sets; 21 unit tests pass.

rubenfiszel:
Round 9 nits closed in `ebe2b9f` (EE `3e79901`).

**[pi] `discard_stale_local_chunks` doc still six lines** — condensed to four.

**[claude] is the init tracker write ordered after the pull is durably complete?** It is, and the ordering is load-bearing rather than incidental, so worth spelling out:

```
discard_stale_local_chunks()   // drop anything unvouched
open_local_chunks()
pull_missing_chunks()          // any failure -> Err, aborting init
                               // handoff    -> Ok(false) -> return Ok(None)
log_tracker.save_data_to_file()   // only reached when the pull completed
```

Every failure inside `pull_missing_chunks` propagates as `Err` — I made partial restores abort init back in round 2 for this reason — and a handoff returns early before the write. So a partial pull never gets a tracker written over it. A process death mid-pull is the same: no tracker on the next start, so the guard distrusts every local chunk and the pull redoes the lot. The vouching claim is only ever recorded once it is true.

---

✅ Review round clean @ ebe2b9fcac27ecd76caaf1077b8c3daf200d9c01 — nit-only verdicts at f7468e04449b59ca26d56368680400c114d34cea (codex: good to merge); nits addressed in ebe2b9f / claude's dismissed in the reply above.

Leaving this in draft for @rubenfiszel to flip: it touches `*_ee.rs` with a companion PR, which `.claude/skills/pr` puts in the ask-before-ready bucket.

PR #10895, opened 2026-08-29
fix: scope app deployment history to the workspace and its app

rubenfiszel:
## Summary

Fixes WIN-2443.

`POST /api/w/{workspace}/apps/history_update/a/{app_id}/v/{version}` addressed a
deployment message by `(app_id, app_version)` without tying either to the calling
workspace:

- the app lookup was `SELECT path FROM app WHERE id = $1`, so any app id from any
  workspace resolved;
- `app_version.id` is a global sequence and the supplied version was never checked
  against the resolved app, so any version id in the instance could be referenced;
- `get_app_history`, `get_latest_version` and `list_apps?with_deployment_msg=true`
  joined `deployment_metadata` on `av.id = dm.app_version` alone, so a row written
  under another workspace surfaced in the victim's deployment history.

Together, any authenticated user could write an arbitrary deployment message into
another workspace's app history by calling the endpoint with their own app id and the
victim's version id.

## Changes

- `update_app_history` resolves the app path with a single query that joins
  `app_version` and filters on `a.workspace_id`, so both the app and the version must
  belong to the caller's workspace; otherwise it 404s (message now names the version
  too).
- `get_app_history` and `get_latest_version` scope the `deployment_metadata` join with
  `AND dm.workspace_id = a.workspace_id`.
- `list_apps` scopes the same join with `AND dm.workspace_id = app.workspace_id`, so a
  pre-existing injected row cannot leak through the list either.
- New integration test `app_history_cross_workspace` covering both halves, with a
  second-workspace fixture.

The only caller of the endpoint, `DeploymentHistory.svelte`, passes an app id and
version taken from the current workspace's own history, so it is unaffected.

### Not covered: same-workspace residual rows

The join is scoped by workspace, not by path. The pre-fix endpoint also allowed
planting `(own_workspace, own_app_path, coworkers_version_id)`, and such a row still
satisfies `dm.workspace_id = a.workspace_id`, so it can still surface on a co-worker's
history in that same workspace. Nothing can create one after this PR — this is residual
data only. Tightening the join with `AND dm.path = a.path` would be a regression:
nothing rewrites `deployment_metadata.path` on app rename, so it would hide the
messages of every renamed app's older versions.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test app_history_cross_workspace`
      passes; it fails on the pre-fix code (write half: 200 instead of 404) and fails
      again with only the read-side join scoping reverted (injected message surfaces).
- [x] `cargo test -p windmill-api-integration-tests --test apps` still passes.
- [x] `SQLX_OFFLINE=true cargo check -p windmill-api --features quickjs` and
      `SQLX_OFFLINE=true cargo check -p windmill-api-integration-tests --all-targets`.
- [x] On a running instance with two workspaces: `history_update` with a foreign app id
      or a foreign version returns 404, the legitimate update returns 200 and appears in
      its own workspace's history; a `deployment_metadata` row planted under the other
      workspace for the victim's version no longer appears in the victim's
      `history/p`, `get_latest_version`, or `apps/list?with_deployment_msg=true`.
- [x] Updating the deployment message of a **non-latest** version returns 200 and
      persists, and `apps/list?with_deployment_msg=true` keeps showing each app's own
      latest-version message.
- [x] UI: the app editor's Deployment History panel lists the deployment message and
      updating it through the pencil control still writes to the correct workspace.

---

Left in draft: this changes an authorization path, so it wants a human look before
ready. The review round is clean (see the marker comment).

### DIFF
--- a/backend/.sqlx/query-a462d8661fee58129ec52952ee08a33fcc2e22b47d352030d55e1375c885cd36.json
+++ b/backend/.sqlx/query-a462d8661fee58129ec52952ee08a33fcc2e22b47d352030d55e1375c885cd36.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "SELECT path FROM app WHERE id = $1",
+  "query": "SELECT a.path FROM app a JOIN app_version av ON av.app_id = a.id\n        WHERE a.id = $1 AND a.workspace_id = $2 AND av.id = $3",
   "describe": {
     "columns": [
       {
@@ -11,6 +11,8 @@
     ],
     "parameters": {
       "Left": [
+        "Int8",
+        "Text",
         "Int8"
       ]
     },
@@ -18,5 +20,5 @@
       false
     ]
   },
-  "hash": "abb56dec024600d0d7a63212f63c3296196898003707f2cb3a890690173a51a5"
+  "hash": "a462d8661fee58129ec52952ee08a33fcc2e22b47d352030d55e1375c885cd36"
 }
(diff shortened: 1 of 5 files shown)


rubenfiszel:
✅ Review round clean @ efa2b98c47a4e924c7558134318c195a9bda5db3 — codex and pi "Good to merge" at c3a98c880abdf00bf61b970b609f9028f70c1c45; claude nit-only at the same SHA, nits addressed in efa2b98c47a4e924c7558134318c195a9bda5db3 (vacuous negative assertion) and in the PR description (same-workspace residual rows, documented rather than tightened — a path-scoped join would hide renamed apps' older messages).

Leaving this as a clean draft rather than flipping to ready: it changes an authorization path, so it wants a human look first.

PR #10897, opened 2026-08-29
[ee] perf: add service log documents to the index one batch at a time

rubenfiszel:
## Summary

The service log indexer's ingest called `index_log()` once per log line, doing one
`tokio::task::spawn_blocking` **per document**, i.e. a scheduler round trip per line. This batches
the tantivy `add_document` calls into the same batches that already get flushed to the parquet
store, which is what `rebuild_index_from_store` in the same file has always done.

Two bugs surfaced while restructuring, and they are the stronger reason to take this:

- **The mid-pass flush returned without rolling the writer back.** If the store flush at the
  `LOG_STORE_MAX_BUFFERED_ROWS` boundary failed, the documents already added stayed pending in the
  shared writer, the cursor never moved, and the next pass re-read the same files and re-added
  them, so the next commit published both copies and inflated the per-host counts in the Service
  Logs UI. The final flush already rolled back; this makes the two sites identical.
- **One line per batch was indexed with an empty body.** `index_log` read `store_rows.last()` for
  the searchable text, but at a flush boundary that runs immediately after `std::mem::take`, so it
  returned `None` and the line was indexed unsearchable.

## Measured

Same fixture both sides (45 files x 5,000 JSON lines = 225,000 lines per pass, 4 hosts, filesystem
object store), timing from the `log_file` query returning to the last batch being published:

| | median | mean | range | n |
|---|---|---|---|---|
| before | 36.85 us/line | 37.18 | 34.78-40.29 | 6 |
| after | 23.30 us/line | 23.28 | 23.09-23.51 | 6 |

Per-row breakdown of the new path over 906,882 rows: parquet arrow/ZSTD encode and write
5.12 us/row, batched `add_document` **3.39 us/row**, object GET plus `parse_line` plus
`searchable_text` plus doc build 14.44 us/row.

So the index step went from about 17.3 to 3.39 us/line, landing where `rebuild_index_from_store`
sits (5.5 to 6.3 us/row): the per-document hop was the whole gap on that step. The pass as a whole
is only 1.6x because the parse and the parquet encode are about 19.5 us/line that the rebuild path
never pays. This is a correctness change with a throughput side effect, not the other way round.

## Changes

- `index_log` becomes `build_doc`: pure, returns a `TantivyDocument`, no blocking hop.
- New `add_documents`: one `spawn_blocking` for a whole batch, taking the writer lock once.
- New `flush_batch`: publishes the store rows and the same lines' documents together, so a batch
  lands in both or in neither.
- `whole_file_read` is replaced by `pending_files`. A file joins it once fully read into the
  buffer, and only moves to `ingested` (the raw-object delete list) after the batch holding its
  last line has been published. Every failure path abandons the pass, so a file whose lines are
  half-flushed is never deletable.
- The mid-pass flush now calls `rollback_writer` before returning, as the final flush already did.
- The mid-pass store object keeps the key the per-line ingest produced (`commited + n - 1`, so
  `-p199999`, `-p399999`, and so on). `append_partition` dedups the manifest by exact name, so
  renumbering these would leave an old partial object beside the retry's copy and permanently
  double those rows in both retrieval and any later index rebuild.

Note that `add_document` fails only when the writer's worker threads are dead
(`index_writer_status.is_alive()`), never per document. The old "log it, skip this file, continue"
could therefore never do what it looked like it did: on the only reachable failure it would skip
every remaining file and then die at the commit's `??` anyway. Abandon-and-retry is what it was
pretending to be, and there is no poison-document wedge risk in the new shape.

## Test plan

No automated test: `service_logs_ee.rs` has no test module and there is no indexer integration
harness, so guarding the promotion rule would mean standing up a tantivy writer, an object store
and a `Pool<Postgres>` for a bookkeeping assertion. Per AGENTS.md that is a ceremonial test, so
this ships none. Verified on a running instance instead:

- [x] 225,000-line pass crosses the 200,000-row boundary (`Wrote 25000 log lines` in the final
      flush confirms the mid-pass flush fired), index holds exactly 225,000 documents, all 45 raw
      files deleted, `total_lines` in the ingest state matches the lines planted. Across the whole
      session, 1,350,000 planted and 1,350,000 indexed.
- [x] Files straddling the batch boundary are still deleted exactly once.
- [x] **Forced mid-pass flush failure** (store tree made read-only): pass aborts with
      `Error publishing a log batch`, the rollback runs, and afterwards all 45 raw files are still
      present, the cursor is byte-identical to before the pass, nothing is deleted, and the index
      doc count is unchanged, i.e. the roughly 200,000 documents added before the failure were
      dropped.
- [x] **Same-version retry** after that failure: 900,000 documents for 900,000 planted lines, no
      duplicates, nothing missing. Without the rollback fix this would have been 1,100,000.
- [x] **Cross-version retry** (the upgrade window for the object-key change): ran a 225,000-line
      pass on the base `main` build and `SIGKILL`ed it the instant the mid-pass object appeared.
      Base wrote `ing-...-p199999.parquet`, no ingest state, all 45 raw files intact. Swapped in
      this head and let it retry. The partial object was overwritten in place rather than
      duplicated:

      ```
      dt=2026-08-29/mode=worker/manifest.json
          ing-20260829T203954626-p199999.parquet   197575
          ing-20260829T204454626.parquet            27425
          TOTAL                                    225000
      ```

      225,000 store rows and 225,000 index documents for 225,000 planted lines. With a renumbered
      key the worker partition would have held 422,575.
- [x] `cargo check --features enterprise,private,parquet,tantivy` and `--features private,tantivy`
      both clean.

Not addressed here, and pre-existing in #10892: the mid-flush `pass_id` is the running max
`log_ts` at the boundary, so a retry where an out-of-order file shifts that boundary already writes
a differently-named partial object today. `per-hostname-service-log-index-cursor` is independently
moving these names to be content-derived, which is the durable fix for the general case.


rubenfiszel:
Confirming what the current head names that object, since the runner can't fetch `windmill-ee-private` — and the Codex P1 from the previous round was real and is fixed.

**The fix is the narrowest one you suggested**: the mid-pass key is now `commited + n - 1`, which reproduces the pre-batching sequence byte-for-byte rather than restarting at a batch multiple.

I ran the cross-version scenario you both asked for rather than arguing it:

1. Built the **base** (`main`) EE code, planted 45 files x 5,000 lines = 225,000, and `SIGKILL`ed the process the instant the mid-pass object appeared. Base wrote `ing-20260829T203954626-p199999.parquet` — confirming the `p199999` sequence empirically, not just from reading the code. No `ingest_state.json`, so the cursor never advanced; all 45 raw files survived.
2. Swapped in **this head**, rebuilt, and let it retry the interrupted pass.

Manifests afterwards:

```
dt=2026-08-29/mode=worker/manifest.json
    ing-20260829T203954626-p199999.parquet   197575   <- overwritten in place, not duplicated
    ing-20260829T204454626.parquet            27425
    TOTAL                                    225000
dt=2026-08-29/mode=standalone/manifest.json
    ing-20260829T203954626-p199999.parquet     2425
```

225,000 store rows for 225,000 planted lines, one manifest entry for the partial batch, and the index holds exactly 225,000 documents. With the `-p0` numbering this would have been 197,575 + 197,575 + 27,425 = 422,575 in the worker partition.

On the adjacent point in your footnote: agreed, and I've deliberately not touched it. The mid-flush `pass_id` is the running max `log_ts` at the boundary, so a retry where an out-of-order file shifts the boundary still writes a differently-named partial object. That is pre-existing in #10892 and unchanged here; my fix closes only the numbering divergence this PR would otherwise have introduced. Worth noting that `per-hostname-service-log-index-cursor` is independently moving these names to be content-derived ("name store objects by content", "include the source file in the store object name"), which is the durable fix for the general case.

On the second-buffer nit from the earlier round: not taking it. Building the documents inside the blocking hop from `&rows` would change behaviour, because the index document's timestamp is the **file** timestamp while `parse_line` sets the row's to the line's own `timestamp` field — so docs built from rows would silently change time-range filtering. Preserving the current semantics needs per-row `host_key` and `file_ts` anyway, and the zip-alignment hazard that introduces sits in the path that governs raw-file deletion. Not a good trade for the memory.

A re-round won't move this finding, since the CI runner still can't read the EE code the finding is about.

rubenfiszel:
✅ Review round clean @ 5c8b249aa328aae59364bd0d9167af74cadfd7b6

Codex and Pi both returned "Good to merge" at this head. Claude's round-2 verdict was "Should
address issues", but it was explicit that it could not fetch the EE commit (`windmill-ee-private`
is outside the CI job's permission set), so it verified the mechanism against the base and asked
the author to confirm what the current head names the mid-pass object. That confirmation, plus the
cross-version test it asked for, is in the comment above: the head reproduces the base key
(`-p199999`), the base's stranded partial object is overwritten in place rather than duplicated,
and the totals are exact. Head unchanged since the round. Flipping to ready at the author's
request.

rubenfiszel:
Superseded by windmill-labs/windmill#10898 (companion: windmill-labs/windmill-ee-private#758), which carries the two bug fixes from this PR without the batching.

Closing rather than rebasing. #10893 day-partitioned the index, so `index_log` now resolves `writer_for_day(timestamp.date_naive())` before adding. A buffered batch can straddle days, which turns "one blocking hop per batch" into "one hop per (batch, day)": a different shape from the one built and measured here, not a conflict resolution. And every number in the description above was measured against a base that no longer exists, so the throughput comparison, the forced mid-pass failure and the cross-version retry would all have to be redone.

The 1.6x was never the reason to take this. The reasons were the two bugs, and both are independent of the batching:

- the mid-pass flush returning without `rollback_active()`, which duplicates every line of the batch on the next pass
- `store_rows.last()` after `std::mem::take`, which indexes one line per batch with an empty body

Both are still live on `main` and both are in #10898, which is about 25 lines against current `main` and needs no re-measurement.

One thing that disappears with the batching: the `commited + n - 1` object-key fix existed only to stop this PR renumbering the mid-pass store objects. Without the batching there is no renumbering, so the upgrade-window hazard and the cross-version test requirement go with it.

The per-document `spawn_blocking` is still there on `main` and is still worth removing eventually. Better done once `per-hostname-service-log-index-cursor` has landed and the ingest loop has stopped moving.

PR #10901, opened 2026-08-31
fix: stop the release version bump from re-resolving every dependency

rubenfiszel:
## Summary

The v1.799.0 release never produced Docker images: both `build` and `build_ee` failed to compile, and `ghcr.io/windmill-labs/windmill{,-ee}:1.799.0` are still 404 (so `:latest` remains 1.798.1).

Root cause is in the release path itself, not in the code being released. `change-versions.yml` ran `cargo generate-lockfile` after bumping the version, which rebuilds `backend/Cargo.lock` from scratch **at the latest available version of every package**. In the 1.799.0 release commit that produced a 377-line lockfile diff, including `argon2` 0.5.3 → 0.6.0 — permitted by the then-current `argon2 = "^0"` requirement, and an API break that stopped `windmill-api/src/users_ee.rs` compiling (`password_hash::SaltString` gone, `hash_password` down to one argument). `check_ee_full` and `cargo_test` went red on the release PR; the tag was cut anyway and the image build failed.

`argon2` has since been pinned to `0.5`, which unblocked `main`, but the mechanism that pulled the breaking bump in is untouched and would fire again on any future `^0` dependency. This PR closes that mechanism: a release now changes version strings only, never dependency resolution.

Every backend workspace member is named `windmill*` and shares one version, so their lockfile entries can be rewritten in place — exactly what the script already does for the `windmill-parser-wasm` workspace lock. With that widened sed, the `cargo generate-lockfile` step is unnecessary as well as harmful.

The main-branch `run_integration_test` failures since 1.799.0 are the same incident, not a separate bug: `integration_tests/build.sh` pulls `ghcr.io/windmill-labs/windmill-ee:$(cat version.txt)` and gets `manifest unknown` because that image was never published. It resolves itself once a release publishes images again.

## Changes

- `.github/change-versions.sh`: widen the `backend/Cargo.lock` version rewrite from just the root `windmill` package to every `windmill*` entry, so the lockfile stays internally consistent without re-resolution. Records the invariant the sed depends on — every `name = "windmill*"` entry in that lock is a workspace member on the shared version, which `windmill_duckdb_ffi_internal` (pinned at `0.1.0`) satisfies only because `exclude` keeps it out of the lock.
- `.github/workflows/change-versions.yml`: drop the `cargo generate-lockfile` step and the `actions-rs/toolchain@v1` setup it needed (that action is archived, so this removes it from the release path as a side benefit).
- Delete `.github/change-versions-mac.sh`: unreferenced anywhere in the repo, and a second divergent copy of the release-bump path is precisely the hazard this PR is closing. It had also rotted — both of its multiline lock seds are silent no-ops under BSD `sed` (no `-z`, so a newline never enters the pattern space), it writes `v$VERSION` into `cli/src/core/constants.ts` where the real script writes `$VERSION`, and it misses `typescript-client/jsr.json` and `windmill-yaml-validator/package.json`.

## Test plan

Verified by running the patched script over a clean extraction of `main` at version `1.800.0`:

- [x] `backend/Cargo.lock` diff is exactly 79 version lines and zero dependency changes (it was 1064 lines under `cargo generate-lockfile`); `argon2` stays at 0.5.3
- [x] `cargo metadata --locked --offline` accepts the resulting lockfile without rewriting it — i.e. it is complete and needs no re-resolution at Docker build time
- [x] the counterfactual confirms both halves are needed: with the old narrow `name = "windmill"` sed, `--locked` fails with `cannot update the lock file`, so dropping the workflow step alone would have been a regression
- [x] the bump-and-revert round trip used by `integration_tests/build.sh` (to `<ver>-dev` and back) leaves `backend/Cargo.lock` and `backend/Cargo.toml` byte-identical — an improvement, since the old sed left 78 of the 79 members stale in between
- [x] no other workflow, Dockerfile, or script runs `cargo generate-lockfile` / `cargo update`, so nothing undoes the fix elsewhere
- [ ] confirm on the next `version.txt` release PR that the bot's auto-commit touches only the 79 `windmill*` version lines in `backend/Cargo.lock` and no third-party `[[package]]` block, then that the release Docker build compiles

Note that 1.799.0 itself cannot be recovered in place — that tag's lockfile still pins argon2 0.6.0, so re-running its build fails again. Recovery is a new release cut from `main`, which builds green today, and which also unsticks `:latest` / `:main` and the main-branch integration test.

---
## Summary by cubic
Stops the release version bump from re-resolving every dependency, which is what caused the 1.799.0 Docker images never to be published.

`cargo generate-lockfile` ran after the version bump, resolving every package at its newest version. For 1.799.0 that pulled `argon2` 0.5.3 → 0.6.0, a breaking change that failed the build, so neither `build` nor `build_ee` produced images and `:latest` stayed at 1.798.1. The bump now edits only the `windmill*` version lines in `backend/Cargo.lock` and never re-resolves dependencies.

**Bug Fixes**

- Widens the lockfile sed to rewrite every `windmill*` entry (not just root `windmill`), so `cargo metadata --locked --offline` accepts it without rewriting.
- Drops the `cargo generate-lockfile` step and the `actions-rs/toolchain@v1` action from the `change-versions.yml` workflow.
- Removes `.github/change-versions-mac.sh`, an unreferenced and broken duplicate of the release-bump path.
- 1.799.0 can't be recovered in place, but a new release cut from `main` publishes images and unsticks `:latest`, `:main`, and the `run_integration_test` on main.

<sup>Written for commit 12e1400f6ca74d9da579ff88bcd60b023353ad13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ 12e1400f6ca74d9da579ff88bcd60b023353ad13

PR #10902, opened 2026-08-31
fix: upgrade argon2 to 0.6 and migrate the password hashing API

rubenfiszel:
## Summary

#10890 pinned `argon2 = "0.5"` to stop the v1.799.0 release build failing, which unblocked `main` but left us held off 0.6 with the code still written against the old API. This lifts the pin and does the migration.

(The release-automation half of that incident — the release PR re-resolving every dependency, which is how the breaking bump got in — is #10901, separately.)

argon2 0.6 / password-hash 0.6 changed four things we depend on:

- `SaltString` moved into the new `phc` crate
- `password_hash::rand_core` is now behind a non-default feature
- `PasswordHasher<H>` is generic, and `hash_password(&self, password)` generates the salt itself
- `impl PasswordVerifier<str> for Argon2` — verify takes the PHC string directly

The migration makes both call sites smaller: we no longer hand-roll salt generation, and we no longer parse the stored hash before verifying it.

`Argon2::default()`'s cost parameters are byte-identical between 0.5.3 and 0.6.0 (m=19456, t=2, p=1, output 32), so newly minted hashes keep exactly the shape they had.

## Changes

- `backend/Cargo.toml`: `argon2 = "0.5"` → `"0.6"`, and the pin comment now states the durable constraint (argon2's 0.x minors are API-breaking, so this stays minor-pinned rather than taking the `^0` its neighbours use) instead of describing the old workaround.
- `backend/Cargo.lock`: argon2 0.5.3 → 0.6.0, password-hash 0.5.0 → 0.6.1, `phc` 0.6.1 added. Only argon2's transitive closure moved; no unrelated dependency rode along.
- `backend/windmill-api-users/src/users.rs`: drop the `PasswordHash::new` parse on the login path and verify against the PHC string directly. Adds a unit test.
- EE `windmill-api/src/users_ee.rs` (companion PR): drop `SaltString::generate(&mut OsRng)` and let argon2 pick the salt.

### Salt generation is equivalent

Worth stating explicitly since the diff deletes an explicit RNG call. password-hash 0.5's `SaltString::generate` filled a 16-byte salt from the supplied `CryptoRngCore` (`OsRng`); 0.6's `hash_password` fills a 16-byte salt via `getrandom::fill`. Same length, same OS CSPRNG. The failure mode improves: `OsRng::fill_bytes` panicked on RNG failure, whereas `try_generate_salt()?` returns an error that surfaces as `Error::InternalErr`.

### Dropping the explicit parse also fixes a small login-path leak

A corrupted stored hash previously produced `Error::internal_err(e.to_string())` — an HTTP 500 handing the PHC parser's message back to an unauthenticated caller. It now falls into the wrong-password branch: `users.login_failure` audit entry, rate-limit increment, 400 "Invalid login". The parse still happens, just inside argon2.

## Test plan

Exercised against a running backend (`--features quickjs,private`) with a real database, not just type-checked:

- [x] login as `admin@windmill.dev`, whose stored hash is the migration-seeded **argon2 0.5-era** `$argon2id$v=19$m=4096,t=3,p=1$...` — succeeds, so no existing instance is locked out
- [x] wrong password on that account rejected with HTTP 400
- [x] create a user, minting a fresh hash through the EE `hash_password`: comes out `argon2id v=19 m=19456,t=2,p=1`, identical to what 0.5.3 produced
- [x] log in as that user — hash-then-verify round-trips entirely within 0.6
- [x] wrong password for that user rejected with HTTP 400
- [x] two users given the same password get different salts (`LZIjntYIg96hy3r1xju3Ww` vs `2K0+bJ2a+VQKxYWDTvy3mQ`) — the property most at risk from delegating salt generation to the crate
- [x] `cargo check --features all_sqlx_features,private` clean on toolchain 1.97.0
- [x] `cargo test -p windmill-api-users --lib` passes

The shipped test pins the one property this class of change can silently break: a hash minted by an older argon2 must keep verifying. It goes through the same `PasswordVerifier<str>` call shape as `login`, needs no fixtures or database, and its negative assertion guards against a vacuous pass. The hashing half stays uncovered by automated tests because it is EE-gated and is now a three-line delegation to the crate; it is covered manually above.

---
## Summary by cubic
Lifts the `argon2` pin from 0.5 to 0.6 and migrates the password hashing API to match, so existing hashes keep verifying and the code no longer hand-rolls salt generation or parses stored hashes on login.

**Migration**
- `argon2 = "0.6"` stays minor-pinned because 0.x minors are API-breaking.
- `Cargo.lock` updates only argon2's transitive closure; no unrelated dependencies moved.
- The EE companion PR drops the explicit `SaltString::generate` call, but the 16-byte salt from the OS CSPRNG is unchanged.
- Hash shape is byte-identical, so hashes minted by 0.5 keep verifying.

**Behavior change**
- A corrupted stored hash previously returned an HTTP 500 with the parser's error message; it now falls into the wrong-password branch with an audit entry and 400.
- Tests pin both sides of that change: a unit test verifies the migration-seeded admin hash still verifies, and an integration test asserts the 400 on an unparseable stored hash.

<sup>Written for commit 9d2d32603ec0ff05ec82d373c65446c4b0b6c76d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10902?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ a7e6158279ef40afa87926d705c96e72c82a06fd — nit-only verdicts at a7e6158279ef40afa87926d705c96e72c82a06fd; nit addressed in 7c46112c76 (tightened the existing integration assertion from `400 || 401 || 500` to `assert_eq!(400)`).

Verified the tightened assertion is a real guard rather than a tautology: with the login change temporarily reverted to the pre-PR `PasswordHash::new` shape, `test_user_endpoints` fails with `login: unexpected status 500 Internal Server Error`; restored, all 5 tests in the file pass.

Also resolves the one thing the review flagged as unverifiable from its checkout: CI `check_ee_full` passed on this PR, which resolves `ee-repo-ref.txt` to EE commit `02a89fc` and compiles CE+EE together with `--features all_sqlx_features,private`.

PR #10903, opened 2026-08-31
feat: expose request headers to scripts invoked via MCP

### LINKED ISSUE #10887: feature(mcp): expose incoming HTTP headers to scripts invoked via the native MCP endpoint
## Problem

When a script is invoked through the native MCP endpoint, the script has no way to read the HTTP headers of the incoming MCP request. HTTP triggers expose headers, the MCP endpoint doesn't.

This matters for multi-user setups. We run a self-hosted instance as an MCP server exposing a set of tools that wrap an internal ticketing system's REST API. Many tools are user-scoped ("show my tickets", "comment on my request", attribute this ticket to the requesting person). Since scripts can't see who's on the other end of the MCP connection, we currently pass the user identity as a tool parameter that the LLM fills in, fenced with prompt rules like "never invent or accept a different address". It works, but identity traveling through the model is spoofable by prompt injection, no matter how strict the descriptions are.

The obvious fix would be identity pass-through via headers: the MCP client sends `Authorization: Bearer <per-user token>` or something like `X-User-Id`, and the script resolves the caller server-side from a verified value. Other MCP servers do exactly this (e.g. Combodo's standalone MCP server for their ITSM product takes a per-user bearer token). With Windmill we can't, because the header never reaches the script.

The only workaround we found is a custom gateway that terminates MCP itself and re-invokes scripts via webhook/HTTP triggers with injected headers — which throws away the best part of the native MCP support (tool schemas generated from script signatures) and adds another component to operate.

## What I'd like

Some supported way to read incoming request headers from a script when invoked via MCP, e.g. `wmill.get_request_headers()` returning a dict (empty outside HTTP/MCP contexts), or an injected context parameter similar to how resources work.

A few things probably worth thinking about:

- Should be opt-in / allowlisted per script or workspace, so the Windmill MCP token's own `Authorization` header doesn't leak into every script by default
- Header values shouldn't show up in job args or logs
- If it's done as a parameter, it needs to be excluded from the generated MCP tool schema — the LLM must not see or fill it

## Alternatives considered

- LLM-supplied identity parameter with strict schema rules — current approach, works day-to-day but is a soft control
- Custom MCP gateway in front of Windmill — duplicates what the native endpoint already does well

Happy to provide more detail on the setup or test a preview build.

hugocasa:
## Summary

Closes #10887. A script or flow invoked through the native MCP endpoint could not read the HTTP headers of the incoming request, so a multi-user MCP server had no way to learn who was on the other end. Caller identity had to travel as a tool parameter the LLM fills in, fenced with prompt rules — a soft control, forgeable by prompt injection.

This delivers the request to a runnable's **preprocessor**, the way every other trigger does. The model supplies the tool's arguments, which become `event.body`; the headers arrive in `event.headers`, a key the model never writes. That separation *is* the guarantee — an identity the model cannot address is one prompt injection cannot forge — and it needs no allowlist, no schema rewriting and no argument filtering to hold.

| Runnable | Receives |
|---|---|
| V2 preprocessor | `event.kind == "mcp"` with `body`, `headers`, `tool_name` |
| V1 preprocessor | `wm_trigger.kind == "mcp"`, `wm_trigger.mcp.headers`, main args unchanged |
| No preprocessor | only what the model sent, unchanged |

V2 preprocessors are also a fix: before this, a `preprocessor(event)` invoked over MCP received the raw tool args and saw `event === undefined`.

A runnable that wants a header without a preprocessor is asking for it to arrive as an ordinary parameter — one the model can also write, which would make the value unattributable. Adding a preprocessor is what makes the request readable, and it can map `event.headers` onto whatever `main` expects.

There is nothing to configure. A preprocessor receives the request, the way a webhook or HTTP trigger preprocessor does; a runnable without one receives only what the model sent.

`Authorization`, `Cookie` and `Proxy-Authorization` are the exception: they are withheld. That is a deferral, not a security boundary — a webhook preprocessor does receive them, and the asymmetry that would justify withholding them here does not survive scrutiny. Nothing needs them yet, and this is the direction that stays reversible: releasing one later is additive, while withdrawing one after runnables have started reading it is not. If someone asks, it becomes a setting.

## Which tools deliver headers

**Only the per-item script and flow tools** — the ones MCP generates from the workspace's own scripts and flows (`s-u_team_my__script`, `f-u_team_my__flow`). Those call `prepare_push_args` directly, so this PR owns the whole path from request to job arguments.

The endpoint tools generated from the OpenAPI catalogue — including `runScriptByPath` and `runFlowByPath` — do **not**. They reach a runnable by proxying over a second HTTP request to the webhook run route, and that route builds the preprocessor event itself from the headers of the proxy's own request. Making those forward the caller's headers means copying headers across the hop and second-guessing what the shared webhook code will do with them, which is a seam worth its own PR rather than a rider on this one.

Consequently **multi-workspace (gateway) tokens do not receive headers at all**, since that mode lists only endpoint tools.

## Changes

- `prepare_push_args` is now format-aware, reusing `get_runnable_format` from `windmill-trigger`. `TriggerKind::Webhook` reaches the same generic arm MCP wants, so no new `TRIGGER_KIND` variant and no migration.
- `McpRequest` carries the caller's `HeaderMap` and the invoked tool name from `Runner::extract_context` through to `run_script` / `run_flow`.
- MCP CORS mirrors the request's header list instead of a fixed one, so a browser MCP client can send a custom header. Nothing is granted by echoing it: origin is `Any`, so browsers never attach credentials.

No new public surface: no query parameter, no setting, no UI, no telemetry. Six backend files, no frontend and no EE change.

## Scope of the guarantee

- A token that can `updateScript` / `updateFlow` can rewrite the runnable itself, so the guarantee presumes an MCP token whose scopes exclude authoring.
- **The guarantee covers `event.headers`, not the runnable's `main`.** Scheduled runs push with `apply_preprocessor: false` (`windmill-queue/src/schedule.rs`) and previews push `JobPayload::Code`, which has no such field — so both call `main` directly with stored arguments and the preprocessor never runs. A script whose preprocessor fills a `main` parameter from `event.headers` can therefore be run with that parameter chosen by the caller, through `createSchedule` / `updateSchedule` / `runScriptPreviewAndWaitResult` on a token that has them. This is how schedules and previews behave for every entrypoint, not something MCP introduces. Note it is not specific to headers: `runScriptPreviewAndWaitResult` runs caller-supplied code, so any token reaching it already has arbitrary execution in the workspace. Scope the token to the scripts and flows it needs where a runnable relies on preprocessor-derived identity.
- **The guarantee covers the header's value, not its presence.** On an `mcp:all` connection `list_tools_single` publishes the per-runnable tools *and* `runScriptByPath` / `runFlowByPath`, and only the first delivers the caller's headers. Which one the model picks is the model's choice, so a preprocessor that quietly defaults a missing header runs as that fallback whenever it picks run-by-path. Nothing is forged — a downgrade, not a bypass — but a preprocessor relying on identity should treat a missing header as an error rather than defaulting, or the token should be scoped so run-by-path is not published.
- The preprocessor event says `kind: "mcp"`, but the job row still records `trigger_kind = webhook`. `job_trigger_kind` has no `mcp` value — adding one is the migration this PR deliberately avoids — and MCP runs reach the queue through the same `run_wait_result_*_by_path_internal` path a webhook does, which is what stamps it. Unchanged by this PR, which sets no `trigger_kind` at all.

## On the withheld credentials

Worth being precise, because the reasoning is not "these are dangerous":

- A webhook preprocessor already receives the caller's `Authorization`, so "the caller's bearer reaches a script other people can edit" is an accepted shape in Windmill, not something MCP would introduce.
- The case for withholding is that **nobody has asked for them**. Resolving caller identity — what #10887 is about — needs a custom header like `X-User-Id`, not the bearer.
- `Cookie` has the strongest case of the three: a same-origin browser MCP client attaches it automatically, so the caller never chose to send it. `Proxy-Authorization` is hop-by-hop by RFC 9110 §7.6.1.
- The known cost: an MCP client authenticating through the OAuth flow sets `Authorization` itself, so an operator who wants the caller's Windmill token in a preprocessor cannot mirror it into a differently-named header. If that user appears, this becomes a setting.

No screenshots: the diff touches no frontend file. Operator-facing documentation of what a preprocessor now receives is handled separately from this PR.

## Test plan

Exercised against a backend rebuilt with `--features quickjs,mcp` — MCP routes are feature-gated and absent from a default build — started as `INCLUDE_HEADERS=x-user-id` so the instance-wide webhook setting was live throughout.

- [x] V2 preprocessor, direct tool: `event.kind == "mcp"`, `tool_name` set, caller's `x-user-id` present in `event.headers`, `authorization` / `cookie` / `proxy-authorization` absent
- [x] V1 preprocessor, direct tool: `wm_trigger.kind == "mcp"`, `caller` resolved from the header, main args still arrive
- [x] No-preprocessor runnable, direct tool: `x_user_id` stays empty despite `INCLUDE_HEADERS=x-user-id` — no header is bound as a parameter
- [x] `Authorization: Bearer <mcp-token>` authenticates an MCP request with no `?token=`, so the caller's bearer does arrive as a request header — and is withheld from the preprocessor rather than passed through
- [x] `runScriptByPath`: the preprocessor sees only the proxy hop's own request (`accept`, `accept-encoding`, `authorization`, `host`, `user-agent`) — no `x-user-id`, no `cookie`
- [x] A stale `?include_header=` on an existing MCP URL is ignored rather than rejected, so old URLs keep working

Automated: `backend/windmill-api-integration-tests/tests/mcp_preprocessor_headers.rs` drives a real MCP `tools/call` against a deployed V2-preprocessor script on a live worker, sending `X-User-Id: alice@corp.example` alongside a model-supplied argument of the same name set to `attacker@evil.test`. It asserts the header lands in `event.headers`, the model's value lands in `event.body` and cannot reach `event.headers`, and the three connection credentials are absent. Confirmed to fail when header delivery is removed and when the withheld list is emptied.


hugocasa:
Round 1 addressed in `954533b`.

**P1 — normalized header-name aliases (codex).** Real: `x-user-id` and the distinct header `x_user_id` both normalised to one name, and which won depended on header iteration order. `McpIncludeHeaders` now stores the exact HTTP name alongside the derived parameter name, and `allowed_headers` looks up by the exact name. Two spellings of one parameter collapse to the first entry, so the mapping stays one-to-one. Verified live: an `X_User_Id` header against an `x-user-id` allowlist now yields the parameter's default, not the alias value.

**P1 — nested endpoint arguments (codex) / `runScriptPreviewAndWaitResult` (claude).** Both real, and worse than run-by-path: preview accepts a deployed `script_hash` plus a nested `args` map, so a model could replay a protected runnable with a forged identity; `createSchedule`/`updateSchedule` persist one to fire later. Top-level stripping could never reach either. The withdrawal now covers every endpoint tool that runs a runnable with model-supplied arguments — `runScriptByPath`, `runFlowByPath`, `runScriptPreviewAndWaitResult`, `createSchedule`, `updateSchedule` — in `list_tools` and at call time. Verified: all five refuse on a header-forwarding connection and are absent from the listing (43 tools vs 48 without an allowlist).

**P2 — gateway advertises tools it refuses (pi, claude).** `list_tools_multi` now takes the allowlist and applies the same filter, so the two modes agree.

**P2 — UI copy (codex, claude).** The hint now covers both delivery shapes (parameter vs `event.headers`) and states that authentication headers are excluded and that the run/preview/schedule tools are withdrawn while the field is set.

**Test coverage (pi, claude).** Added the credential-boundary tests both of you flagged as the highest-value gap: `allowed_headers` refuses a credential header even when explicitly allowlisted, `preprocessor_headers` never carries one while passing everything else through, and the underscore alias does not feed an allowlisted parameter. Plus withdrawal-set and alias-parsing tests. 91 `windmill-mcp` + 33 `mcp::utils` tests pass.

**Not changed — preprocessors receive every non-credential header (claude P2).** Deliberate, and the maintainer's explicit call when the design was settled: it is parity with HTTP-trigger preprocessors, which is what makes headers outside the allowlist reachable at all. Credentials are excluded, which is the part that differs from a webhook. Happy to narrow it to the allowlist if you'd rather, but it would remove the only route to a header the operator did not enumerate in advance.

**Residual limit worth naming:** a token that can `updateScript` can rewrite the runnable itself, so the guarantee presumes an MCP token scoped away from authoring. Noted in the PR body.

hugocasa:
Round 2 nits addressed in `6bb7e98`. All three verdicts were non-blocking; fixing them all rather than dismissing any.

**`list_tools_multi` took the allowlist but never applied it (claude, pi, codex).** A real miss — I threaded the parameter and both call sites and left the loop body alone. The filter is now in the loop. Verified on a real gateway URL: 37 tools with `&include_header=`, 43 without, none of the five withdrawn ones advertised.

**Duplicate headers went first-wins (claude).** `HeaderMap::get` returns the first value where the old `iter()` + `insert` loop returned the last, so a gateway that *appends* an identity header would leave the caller's copy in front. Rather than pick a side, a repeated allowlisted header is now not forwarded at all and logs a warning — the value is not attributable, so the parameter falls to its default. Verified: sending `X-User-Id` twice yields `""`, not the attacker's copy.

**Malformed allowlist entry failed open (claude).** `include_header=x-user-id;x-tenant` produced one entry matching no header *and* stripping no schema property, leaving the parameter model-settable on a connection believed locked down. `parse` is now fallible, validating each name against the RFC 9110 token set, and the middleware 400s — consistent with the unparseable-query rejection above it. Verified: `400 Invalid include_header: 'x user id' is not a valid header name`.

**Withdrawal list had no exhaustiveness guard (claude).** Now derived from `endpoint_path_policy` for `RunByPath(_) | Unconfinable(_)` plus the two schedule names, and pinned by a catalogue test in the same shape as `every_path_addressed_script_or_flow_tool_is_path_confined`. It earned its keep immediately: it flagged `deleteSchedule`, which sharpened the predicate to "hands a runnable a model-composed argument map" rather than "touches a run or schedule path".

**Stale `allowed_headers` doc comment (claude)** and **UI copy omitting the V1 `wm_trigger.mcp.headers` shape (codex, pi)** — both corrected.

**Endpoint parameter collisions fail at call time (codex).** Left as is, matching pi's dismissal: it only fires when an auto-generated endpoint has a parameter named exactly like an allowlisted header, which no current tool does, and it fails closed rather than forging.

Tests: added the two claude named as the gaps that let the `list_tools_multi` miss through green — a duplicate-header tie-break test and the catalogue guard — plus malformed-entry rejection. 91 `windmill-mcp` + 36 `mcp::utils` pass; `cargo check --features enterprise,private,mcp` clean.

hugocasa:
Round 3 nits addressed in `0314a9f`. All three verdicts non-blocking again; four of the five findings were corrections to my own round-2 edits, so I fixed them rather than dismissing any.

**`Unconfinable(_)` over-withdrew `deleteScriptByHash` (claude, pi).** A functional regression I introduced: the variant means "executes arbitrary code **or** affects a script without a checkable path", so deriving from it swept in a tool that runs nothing. The derivation is gone — the set is named again, with a doc comment explaining why it is not derived, and the catalogue guard is what keeps a future addition from slipping through. Verified: `deleteScriptByHash` is listed on a header-forwarding connection again (43 tools), while the five that do run a runnable stay withdrawn.

**`carries_runnable_args` was spliced into another test's doc comment (claude).** My editing error — the path-confinement test's rationale is restored and the helper has its own.

**Guard blind spot (claude).** `additionalProperties` was only consulted when `properties` was empty, so `{"properties": {"script_hash": …}, "additionalProperties": true}` would have read as harmless. Both are now checked independently. The `/jobs/run` / `/schedules/` prefilter stays — widening it would flag unrelated tools — but the helper's doc says what it covers.

**Transport-owned parameters left in the published schema (codex).** Correct and worth fixing: `$user` is a valid header name whose property publishes as `user`, so matching the raw parameter name left it advertised. The strip now compares in the published key space via `apply_key_transformation`, so the schema and the call-time strip agree. No spoofing was possible either way — reverse transformation already discarded the value — but the schema contradicted the contract.

**Duplicate-header hardening covered one arm (claude).** Extended to `preprocessor_headers`, so one request no longer resolves differently depending on whether the runnable has a preprocessor. Verified live: sending `X-User-Id` twice yields `""` on both arms.

**Stale rationale on `strip_transport_owned_endpoint_args` (claude)** — rewritten to describe what it now is, defence in depth behind the withdrawal.

Tests added for the two behavioural fixes: `a_punctuated_header_name_is_stripped_under_its_published_key` and `a_repeated_header_is_dropped_from_the_preprocessor_event_too`. 92 `windmill-mcp` + 39 `mcp::utils` pass; `cargo check --features enterprise,private,mcp` clean; full behaviour matrix re-run with no regressions.

Not starting a fourth round: three consecutive non-blocking verdicts, and the findings have gone P0 → P1 → nits → nits-on-my-own-edits. Leaving this in draft for a human look rather than self-flipping, since it touches an EE file, token/auth paths and adds public surface (a URL query parameter and an `mcp` preprocessor event kind).

hugocasa:
Round 4 addressed in \`eba4423\`. One blocking finding, three nits, all fixed.

**[P1] `include_header` parsing was unbounded and pre-auth (codex).** Correct and the one finding here that mattered. `extract_include_headers` is layered *outside* `route_layer(from_extractor::<ApiAuthed>())`, so an unauthenticated request reaches the parse, and the dedup scan is quadratic in the entry count with nothing capping it — a query string of a few hundred KB of short unique names turns into billions of comparisons on a runtime thread with no `.await` to yield at. `parse` now refuses a value over 1024 characters or 32 entries before doing any of the work; with the count capped the linear scan is bounded at 32 comparisons per entry, so the dedup stays a `Vec`. Regression test `an_oversized_allowlist_is_refused`.

**[P2] The duplicate-header drop blinded the preprocessor on multi-hop requests (claude).** A real regression from my own round-3 edit, and the finding I'd have least liked to ship. The attributability argument is about an *allowlisted* header — two copies, no way to tell which the proxy set. The preprocessor arm has no allowlist, so the same `retain` was also dropping `X-Forwarded-For`, `Via`, `Forwarded`, `Accept`: list-valued fields where repetition is normal, and exactly the ones a preprocessor reads. It now drops a repeat only for allowlisted names. The existing test was widened to pin both halves (`x-user-id` sent twice disappears, `x-forwarded-for` sent twice survives).

**[P2] The two strips matched in different key spaces (claude).** Pi read this as already consistent; it isn't, for a *second* script. With `include_header=$user`, the schema strip removed the published key `user`, but `transform_call_args` reverse-transformed first and compared against `$user` — so a different script with a genuine `user` parameter had it hidden from the model while a model-supplied `user` was still honoured. No forgery either way ($user never feeds `user`), but the schema contradicted the call-time behaviour. Both halves now go through one predicate, `McpIncludeHeaders::owns_published_key`, so they agree by construction rather than by matching each other's arithmetic. Test `a_key_hidden_from_the_schema_is_also_refused_from_arguments`.

**[P2] Test gaps and a stale doc (claude).** `deleteScriptByHash` — the name that regressed in round 2 — is now in the withdrawal test's negative list alongside `deleteSchedule`, with a comment saying why it is the boundary case. `run_by_path_arguments_cannot_carry_a_header_fed_name` still described the pre-withdrawal route; renamed to `endpoint_arguments_cannot_carry_a_header_fed_name` and its doc rewritten to describe the strip as the layer *under* withdrawal.

`cargo test -p windmill-mcp --features server` 91 passed; `cargo test -p windmill-api --features mcp,quickjs mcp::` 39 passed.

Also merged `origin/main` (`2f62ed3`) — the PR had been unmergeable on `backend/ee-repo-ref.txt` since main advanced its EE pin, which is why no round ran between round 3 and now. EE companion merged main too and is at `4853f10`; the ref points there.

hugocasa:
Round 6 addressed in `eb40d1d` and `6877a56`. One P1, and the credential question settled the other way.

**[P1] The schedule handlers carried three more argument maps than the strip reached (codex, claude, pi — independently).** Correct, and my miss: teaching the strip to descend into `args` closed the direct half and left `on_failure_extra_args` / `on_recovery_extra_args` / `on_success_extra_args` open, so `on_failure: "u/team/protected"` with a forged `x_user_id` in its extra args would have bound normally at the handler. The strip is now driven off each tool's own body schema — every property declaring `additionalProperties: true` — rather than a list of names, so a structured object like `retry` is untouched and a catalogue that grows another argument-bearing field is covered without an edit in `windmill-mcp`. The catalogue guard was the deeper problem: it asked whether a tool was *classified*, not whether the strip *reached* every map it carries, which is exactly why this got through. `the_strip_reaches_every_argument_map_a_classified_tool_carries` now asserts the latter against the live catalogue. Verified live: `createSchedule` with a forged name in both maps persists `args={"ticket_id": "T-1"}`, `on_failure_extra_args={"ticket_id": "T-2"}`.

**[P2] Credentials — reversed, and the reviews are why.** The maintainer had asked for parity with webhooks and HTTP triggers, and my argument against it at the time was weak (a webhook route is authenticated too, and `?include_header=authorization` there already binds the caller's token). Two arguments in this round were better than mine and changed the decision: on a webhook the *caller* chooses the callee, while over MCP the *model* does, so an unconditional forward hands the caller's Windmill token to whichever script the model routes to — including one steered there by prompt injection; and the run-by-path arm would put this server's own minted proxy JWT in `event.headers`.

The rule is now **a credential travels only when named**: withheld by default, forwarded if `?include_header=authorization` asks for it. That keeps the webhook parity the maintainer wanted for anyone who deliberately opts in, and drops the part MCP cannot safely copy. Naming one credential does not release the others. Verified live against a preprocessor that dumps its header names: without the allowlist neither `authorization` nor `cookie` appears while `user-agent` / `x-user-id` do; with `include_header=authorization`, `authorization` appears and `cookie` still does not.

**[P2] The gateway UI couldn't configure the mode this arm exists for (claude).** Fair hit — `WebhookBody`'s own doc justifies itself with "multi-workspace mode runs scripts only through these two tools", and the field was still gated on `!isAllWorkspaces`, so that mode needed a hand-edited URL. Ungated, along with the URL parameter.

**[P2] Duplicate-drop alias (codex).** Now matched on the exact allowlisted header name rather than the normalised parameter, so a repeated `x_user_id` is not dropped when only `x-user-id` was named.

**Dismissing [P2] published-key collisions (codex).** This is the inverse of the round-5 finding I implemented at claude's request: matching in the published key space is what fixes `$user`, and matching per-tool originals is what codex now wants. Reaching it needs a punctuated allowlist entry *and* a second tool declaring the collapsed name; both spellings are refused at call time either way, so nothing is forgeable. Swinging the seam a third time is churn, not convergence — happy to revisit if a third reviewer lands on codex's side.

**Not fixed, and worth naming:** on the run-by-path arm a preprocessor still sees the *proxied* request's headers — this server's minted JWT as `authorization`, and none of the caller's non-allowlisted headers. That predates this PR (run-by-path has always proxied with a minted bearer) and lives in shared webhook code, so it is not something this PR should reshape.

`cargo test -p windmill-mcp --features server` 91 passed; `cargo test -p windmill-api --features mcp,quickjs mcp::` 40 passed; `svelte-check` 0 errors.

hugocasa:
Round 7 addressed in `18fe91a`. One P1 and two nits, all real.

**[P1] The strip deleted endpoint control arguments on a name collision (codex).** Correct, and it turns an allowlist typo into a broken connection: the top-level `retain` ran for every endpoint tool and dropped any key matching an allowlisted name, so `?include_header=path` deleted `runScriptByPath`'s target before authorization, and `?include_header=workspace-id` deleted the workspace selector that multi-workspace mode routes on — every call on that connection failing rather than being protected.

The strip now separates the two kinds of argument a tool carries. A free-form body (run-by-path) *is* the runnable's argument map, so its extra keys are stripped, but path parameters, query parameters, declared body fields and the synthetic `workspace_id` are the endpoint's own controls and survive. A structured body is not stripped at the top level at all; only the free-form maps it declares are. Verified live: on `?include_header=path`, `runScriptByPath` still resolves `u/admin/hdr_plain` and the job runs with `{"ticket_id": "T-path"}`. Regression test `an_allowlist_colliding_with_an_endpoint_control_leaves_it_alone` covers both collisions.

**[Nit] The catalogue guard was tautological (claude, pi).** Both right, and it is the more useful of the two findings: the test computed the expected set with the same `additionalProperties` predicate `free_form_arg_map_keys` uses, so it compared the function to a copy of itself and could not fail whatever either did. It now keys off what the catalogue *says* a field is — the `"arguments to pass to the script or flow"` description — which is independent of how the strip finds it, so a field described as runnable arguments but declared some other way fails the test. Checked it actually bites by breaking the strip on purpose: `["createSchedule.on_failure_extra_args", "updateSchedule.on_failure_extra_args"]`, then reverted.

**[Nit] The tooltip's credential sentence was wrong on the run-by-path arm (claude).** Fair, and sharpened by the ungating in the same round: for an all-workspaces token run-by-path is the only route to a runnable, so "reach a runnable only if you name one here" was wrong for every call on a gateway connection — a named `authorization` never survives the proxy hop. The copy now says that route never forwards authentication headers and scopes the opt-in sentence to the rest.

**[Nit] Stale comment at the preprocessor arm (claude).** Rewritten to match what `preprocessor_headers` now does.

`cargo test -p windmill-mcp --features server` 92 passed; `cargo test -p windmill-api --features mcp,quickjs mcp::` 40 passed; `cargo check --features mcp,quickjs` clean; `svelte-check` 0 errors.

hugocasa:
Round 8 addressed in `831f3b1`. One P1 and two nits.

**[P1] `workspace_id` was exempted from the strip in a mode where it is not a control (codex).** Correct, and a hole I introduced in round 7 by hardcoding the exemption. `workspace_id` routes the call in multi-workspace mode — `call_tool_multi` reads it after the strip and removes it before the body is built, so it never reaches a runnable there. In single-workspace mode the workspace comes from the URL and nothing routes on it, so it is just another model-supplied key that the free-form body forwards; exempting it there let a model set a transport-owned parameter of that name. The exemption is now a parameter rather than an assumption, `false` at the `call_tool_single` site and `true` at the multi one.

Verified live on a script taking `workspace_id`, all three directions:

| connection | job args |
|---|---|
| `?include_header=workspace-id`, model sends `workspace_id: "attacker"`, no header | `{"ticket_id": "T-ws"}` |
| same, with `Workspace-Id: prod` | `{"ticket_id": "T-ws", "workspace_id": "prod"}` |
| no allowlist (nothing transport-owned) | `{"ticket_id": "T-ws", "workspace_id": "attacker"}` |

`an_endpoint_control_survives_the_strip_only_where_it_is_one` pins all three.

**[Nit] The catalogue guard skipped the free-form tools (claude).** Right — `runScriptByPath` and `runFlowByPath` declare no `properties`, so they fell through the loop that walks them, and those are exactly the two whose arguments live in the body root. The guard now asserts that a tool classified `WebhookBody` really does declare a free-form body, which is what makes the strip's top-level branch engage; the catalogue is the independent signal, so a body changed to structured upstream fails here.

**[Nit] The tooltip was still wrong on the run-by-path arm (claude, codex).** Third attempt, and the previous two were both too broad. What is true: that route shapes a webhook event rather than an MCP one, and the proxy hop owns `Authorization` so a named one is not forwarded — but `cookie` and `proxy-authorization`, if named, are. The copy now says exactly that and no more: a preprocessor there receives a webhook event, and a named `Authorization` is not forwarded to it. The opt-in sentence sits with the direct-path description it actually describes.

`cargo test -p windmill-mcp --features server` 92 passed; `cargo test -p windmill-api --features mcp,quickjs mcp::` 40 passed; `cargo check --features mcp,quickjs` clean; `svelte-check` 0 errors.

hugocasa:
Round 9 nits addressed in `4cfa95c`. Both non-blocking, both mine.

**Prettier failed on `CreateToken.svelte` (claude).** The tooltip block landed misindented when I rewrote the copy, which would have broken `npm run lint` in CI. Reformatted with the svelte plugin; the diff is whitespace only, no wording change.

**Stale workspace-scoped claim in the PR description (codex).** Right — the description still said the field was gated to workspace-scoped tokens, which round 7's ungating undid. Corrected to say it is offered for gateway tokens too, with the reason: multi-workspace mode reaches runnables only through run-by-path, which is exactly the arm the forwarding covers.

All three verdicts were non-blocking at `831f3b1` (pi: good to merge; codex and claude: nits only), so this is a nit-only push rather than a new round. `svelte-check` 0 errors, `prettier --check` clean.

hugocasa:
## Redesign: headers now reach a runnable only through its preprocessor

Round 10 came back clean (codex and pi "good to merge"), but the maintainer asked whether restricting delivery to preprocessors would simplify this. It does, substantially, and the review history is the argument for it.

Of the eleven blocking findings across ten rounds, ten were the *named-argument* path's attack surface — the `runScriptByPath` bypass, header-name aliasing, nested `args`, `on_*_extra_args`, `workspace_id`, endpoint control arguments, the unbounded allowlist parse. All of them exist because a header bound to a runnable *parameter*, which creates a name the model must never set, which turns every route that reaches a runnable's arguments into a door to defend. The preprocessor arm produced exactly one finding in the same period.

Delivering only through the preprocessor removes the class rather than the instances: the model's arguments become `event.body` and the headers are `event.headers`, a separate key of a struct the server builds, so there is nothing to strip and no second door to close.

**Deleted:** `strip_transport_owned_params`, the header logic in `transform_call_args`, `strip_transport_owned_endpoint_args`, `runnable_args_carrier` / `RunnableArgsCarrier`, `free_form_arg_map_keys`, `endpoint_control_keys`, `body_is_free_form`, `owns_param` / `owns_published_key`, both catalogue guards, and the `McpIncludeHeader` name-pair split. `?include_header=` survives only as the credential opt-in.

**Diff: 1401 → 603 insertions.** `runner.rs` 543 → 138, `tools.rs` 105 → 10.

**What this costs:** a runnable that wants the request now needs a preprocessor. Without one, a header could only arrive as an ordinary parameter — one the model can also write, which is exactly the unattributable value this feature exists to avoid. A preprocessor can map `event.headers` onto whatever `main` expects.

Re-verified on a running instance: v2 preprocessor gets `kind: "mcp"` and the header; a runnable without a preprocessor gets the model's arguments unchanged and no binding; credentials are absent unless named; `runScriptByPath` still carries the caller's header across the proxy hop. Full matrix in the description.

Also addressing the round-10 nit: `("mcp", "header_passthrough")` is confirmed registered at the new pin `135a9eb`, which the main merge moved the ref to.

hugocasa:
Round 11 addressed in `98c7437`, and the P1 is documented rather than fixed — reasoning below.

**[P1] Preview and the schedule tools reach `main` with the preprocessor skipped (claude).** Verified both halves of the claim in code — `windmill-queue/src/schedule.rs` pushes with `apply_preprocessor: false` in four places, and `run_preview_script` pushes `JobPayload::Code(RawCode)`, which has no such field — and then reproduced it end to end. On an `mcp:all` connection, `runScriptPreviewAndWaitResult` against the *deployed* hash of a script whose preprocessor fills `caller` from `event.headers["x-user-id"]`, with `caller` forged in `args` and the real header set to `victim@corp.example`:

```
{"kind": "x", "tool": "x", "caller": "attacker@evil.test", "ticket_id": "P-1"}
```

So the finding stands: `event.headers` is unforgeable, and a `main` parameter *derived* from it is only as trustworthy as the token's scopes.

I'm documenting rather than reinstating the withdrawal, for the reason claude reaches in the finding itself: delivery is now unconditional, so there is no "header-forwarding connection" left to gate a withdrawal on, and withdrawing three tools from every MCP token would regress connections that never touch this feature. It is also not something MCP introduces — a scheduled run of an HTTP-trigger script skips its preprocessor the same way. The mitigation is scoping: omit the schedule and preview endpoint scopes on a token whose runnables rely on preprocessor-derived identity, which the scope picker already supports per endpoint.

The summary overstated things and is corrected: the separation is the guarantee *for anything delivered through a preprocessor*, with a pointer to the limit. "Scope of the guarantee" now carries it as a fourth entry, naming the two push sites and the mitigation.

**[P2] `PROXY_OWNED_HEADERS` was sized for an allowlist (claude).** Right, and it became load-bearing the moment forwarding stopped being allowlist-driven: a four-name list was fine when only named headers rode along, and wrong once the caller's whole set does. It now also excludes the hop-by-hop fields (RFC 9110 §7.6.1) and the `content-*` / `accept-encoding` values `reqwest` sets for the body it is actually sending — a `content-length` copied from the caller's request describes the wrong body. Regression test `the_proxy_hop_never_forwards_this_connections_framing`.

**[P2] Orphaned doc comments (codex, claude, pi).** My deletion pass removed test functions and left their doc comments attached to unrelated tests. Cleaned up in `tools.rs` and `utils.rs`, and the surviving `the_proxy_hop_never_forwards_its_own_credential_header` doc now describes both reasons the hop owns a header rather than only the credential one.

`cargo check --features mcp,quickjs` clean; `windmill-mcp` 81 passed; `windmill-api` 36 passed.

hugocasa:
Round 12 nits addressed in `0cd6e607`. All three verdicts were non-blocking.

**[P2] Gateway mode recorded no telemetry (codex).** Correct, and it was the wrong mode to miss: `header_passthrough` was logged only by `prepare_push_args`, which the run-by-path arm never reaches — so multi-workspace connections, whose *only* route to a runnable is run-by-path, delivered headers and counted nothing. Now counted at the proxy site under the key `proxy`, alongside `preprocessor` for the direct arm. The disclosure copy is corrected to match what the counter actually records.

While adding it I noticed the redesign had left header forwarding unconditional across all ~45 endpoint tools, when only the two run routes hand the proxied request to a runnable. Scoped via `delivers_request_to_runnable`, derived from the catalogue's own path (`/jobs/run/`) rather than a name list. Preview is deliberately excluded: it pushes `JobPayload::Code`, skips the preprocessor, and a header forwarded there would have nowhere to arrive.

**[P2] The doc sweep reached three of seven sites (claude, codex, pi).** Fair — I fixed the ones my deletion script had orphaned and missed the ones that were merely *stale*, which is the harder half. All seven now: `create_http_request`'s doc no longer claims `?include_header=` is what binds the forwarded headers (nothing binds them any more — the run route just sees them); `extract_include_headers` and `McpRequestQuery` describe a credential-release list rather than a parameter allowlist; its inline comment no longer refers to unstripped schemas; `McpRequest.include_headers` likewise; and the orphan on `proxy_jwt_unscoped_caller_keeps_none` is gone.

**[P2] `cargo fmt` unrun (claude).** Formatted the six touched files with `rustfmt` directly rather than `cargo fmt -p`, which reformats the symlinked EE sources and dirties that checkout. Diff is whitespace only in the untouched regions.

Re-verified after the change that run-by-path still carries the caller's header: `x-user-id` present in the preprocessor's header map on a proxied call. `cargo check --features mcp,quickjs` clean; `windmill-mcp` 81; `windmill-api` 36; `prettier --check` clean.

hugocasa:
Round 13 addressed in `c368d94`. One P1, three nits.

**[P1] Run-by-path could bind headers as parameters after all (codex).** Correct, and it defeated the point of the redesign on exactly one path. Forwarding was gated on the *route* rather than on the target: `build_headers`' no-preprocessor branch unions the instance-wide `INCLUDE_HEADERS` (`args.rs:637`), so on an instance that configured one for webhooks, a no-preprocessor `runScriptByPath` would receive the forwarded header as a `main` argument. That is both the parameter binding this design exists to remove and the `INCLUDE_HEADERS` inheritance the PR says it does not take.

Forwarding now resolves the target's format first (`headers_for_proxied_run`) and copies nothing unless the runnable has a preprocessor to receive it. An unresolvable format fails closed. Verified on a backend started with `INCLUDE_HEADERS=x-user-id`:

| call | job args |
|---|---|
| `runScriptByPath` → no-preprocessor script, `X-User-Id` sent | `{"ticket_id": "NP-1"}` — nothing bound |
| `runScriptByPath` → preprocessor script, same connection | `x-user-id` present in `event.headers` |

**[Nit] `proxy` counted calls where `preprocessor` counts deliveries (claude).** Fixed by the same change rather than separately: the arm now only forwards when there is a preprocessor to deliver to, so both keys count a delivery. Comment says so at the log site.

**[Nit] `delivers_request_to_runnable` had no test and fails silently (claude).** Right, and it is a substring match on a catalogue path — a rename upstream would stop forwarding with no symptom but a preprocessor quietly losing its headers. `exactly_the_run_routes_deliver_the_request_to_a_runnable` pins it to `["runScriptByPath", "runFlowByPath"]` against the live catalogue.

**[Nit] Stale comment and prettier (codex, claude).** The framing-header test's doc ended with "Only relevant since forwarding stopped being allowlist-driven", which describes this PR's own history rather than the code — removed. And claude was right that the reworded disclosure lines did not match prettier's wrapping; matched to its output exactly, leaving the rest of that already-unformatted file alone.

`cargo check --features mcp,quickjs` clean; `windmill-mcp` 81; `windmill-api` 37.

hugocasa:
## Round 15 response

**codex P1 (runnable-version race) — dismissing, with the evidence.**

The direct arm is a literal copy of the plain webhook route. `jobs.rs:7533-7543`:

```rust
let args = args
    .to_args_from_runnable(&authed, &db, &w_id, RunnableId::from_script_path(path), …)  // resolves format by path
    .await?;
run_wait_result_script_by_path_internal(db, run_query, script_path, …)                  // resolves the script by path again
```

Same two-step, same window, and it is what every webhook and HTTP-trigger run has always done (`args.rs:348`, `trigger_helpers.rs:482`). Claude reached the same conclusion independently this round and added the severity point: a deploy landing inside the window yields a *mis-shaped* argument, so the job errors. Nothing is forged — the model controls neither the event the server builds nor when someone deploys. Pinning the version through both hops means changing the shared trigger path for every entrypoint, which is a different PR, and would make MCP the only trigger that pins a version instead of running latest.

**On the proxy-specific half**, which is genuinely this PR's: yes, if the target loses its preprocessor between my check and the second hop's, a forwarded header could be bound as a parameter on an instance that configured `INCLUDE_HEADERS`. The value bound is the *real* caller header, not a model-chosen one, and triggering it needs a deploy — and a token that can deploy already voids the guarantee, which is the first entry in "Scope of the guarantee". The reverse direction (no preprocessor → preprocessor) forwards nothing, so the preprocessor sees the hop's request: degraded, not unsafe.

This is the fourth round this race has been raised. It is inherited, low-severity, and the fix belongs in the shared path, so I am not going to keep patching around it here.

**claude P2 (nothing pins the inclusion) — fixed** in 9c2581a. Both surviving tests asserted exclusions and would have passed under the old deny list too. `a_caller_credential_that_is_not_the_hops_own_still_travels` asserts `cookie` and `proxy-authorization` do ride the hop, and I confirmed it bites by re-adding a deny list and watching it fail.

**claude nit (two blank-line deletions in `runner.rs`) — fixed** in the same commit; that file's diff against `main` is now only the feature.

**claude P2 (the disclosure left the product with the field)** — open, and correctly raised. Before this PR an MCP preprocessor received nothing; now it receives the caller's `authorization` and `cookie`, and the token dialog that said so is gone. That is a docs change rather than a code one, and it is with @hugocasa.

**Stale in claude's review** (it read `f40eb5c`): `ee-repo-ref.txt` is back to `f2a3115`, unchanged from `main`. The `mcp`/`header_passthrough` counter was dropped, the EE companion is closed, and the PR now touches no EE and no frontend file.

hugocasa:
## Round 17 response — all four nits fixed in 2a705a5

**Vacuous credential assertions (codex P2 #1, claude P2 #2, pi)** — correct, and it is the same class of mistake as the tautological catalogue guard earlier in this PR. The request carried only `Authorization`, so two thirds of the loop asserted the absence of something never sent, and my sabotage check passed because it only ever exercised `authorization`.

The test now sends `Cookie` and `Proxy-Authorization` too, and I verified each name independently by dropping it from `WITHHELD_FROM_PREPROCESSOR` one at a time:

| dropped | result |
|---|---|
| `authorization` | FAILED — appears in `header_names` |
| `cookie` | FAILED — appears in `header_names` |
| `proxy-authorization` | FAILED — appears in `header_names` |

**Stale CORS contract (codex P2 #2, claude P2 #3)** — fixed. The quantifier was the wrong word; the argument that carries `mirror_request()` is that a custom name cannot be enumerated, and the comment now says only that.

**Overlong credential rationale (codex P2 #3)** — cut from nine lines to four, keeping the durable constraint (not a boundary; withheld because nothing needs them and releasing later is additive) and dropping the deliberation.

**The model can pick the route that drops the identity (claude P2 #1)** — real, and now recorded in "Scope of the guarantee" rather than left implied. `list_tools_single` publishes both the per-runnable tools and `runScriptByPath`, only the first delivers headers, and the choice is the model's. It is a downgrade rather than a bypass, so the guidance is that a preprocessor relying on identity should treat a missing header as an error rather than defaulting. Closing it properly means collapsing the proxy seam, which is the follow-up this PR deliberately scopes out.

hugocasa:
✅ Review round clean @ 2a705a5b002e7741b07c8122e8d015c15665d7d3 — nit-only verdicts at 29c299e282f5be3ec6d93febb0f4f515db7450bd; nits addressed in 2a705a5b00

PR #10894, opened 2026-08-29
fix: track outstanding service log files on the rows themselves

rubenfiszel:
## Summary

The service log ingest walked `log_file` with a cursor over `log_ts` — when a line was *written*, not when its row *appeared*. Rows do not arrive in that order, so a row registered after the pass had gone by its minute sits below the cursor and is never read. Nothing repairs it: the row stays in `log_file` and its lines stay out of search until retention drops them.

Three ways it happens, all the same bug:

- **A delayed upload.** The uploader sends rotated files every 10s and stops at the first failure. Any object-store hiccup delays one host's registrations while the others keep the cursor moving, and the delayed rows land underneath it.
- **A host that has just started.** A new pod's first file covers the minute it started in. If another host registered that minute and a pass ran in between, `log_ts > cursor` excludes it — and its later files push the cursor further, so nothing looks below again.
- **A batch cut mid-minute.** `LIMIT` splits rows sharing one `log_ts`; the cursor advances to that minute and the rest are skipped.

The codebase already works around this. `monitor.rs` gives up on the rest of a host's backlog on any single upload failure, rather than registering out of order.

## The change

**No ordering fixes this.** An arrival sequence fails the same way: `nextval` is allocated before its `INSERT` commits, so a row can take a lower value and commit *after* a higher one has already moved the cursor past it — the same permanent skip, on a narrower trigger.

Which rows are outstanding is a property of the rows, so `log_file.indexed_at` records it. The pass reads `WHERE indexed_at IS NULL ORDER BY log_ts, hostname LIMIT n` and marks what it took. A row that becomes visible late has no mark to be behind, so there is nothing to overtake. It also collapses the two query branches into one — the bootstrap case is just "everything is outstanding" — and the partial index `WHERE indexed_at IS NULL` makes the scan proportional to what is left to do rather than to the retention window.

`max_log_ts` stays as derived output: the index window, the rebuild range and the chunk staleness check are in event time and need a timestamp.

## What the watermark no longer tells you

A pass reading only rows registered behind the watermark leaves it where it was, so anything using it as a progress signal goes blind:

- **What the index has caught up to** is a line count, which advances on every ingest.
- **The event-time projection** kept for a replica on the previous version is biased to "behind" in exactly that case, or it reads a lagging index as synchronised and its next checkpoint erases the record. The state is written in both shapes because the indexer deploys with `RollingUpdate` deliberately — `Recreate` was measured to cause a full search outage for the restart, windmill-helm-charts#674.

## Upgrade

[...]

PR #10904, opened 2026-08-31
fix: harden the service log indexer's recovery and read paths

rubenfiszel:
## Summary

Three follow-ups to #10894, all in the service log indexer. They are separate concerns that share one file, so they are here together rather than as three PRs that would conflict with each other.

## A corrupt cursor no longer takes the server down

`read_ingest_state` propagated a parse failure, and every caller propagates in turn — one of them at startup. A single unparseable `ingest_state.json` therefore stopped a whole server from booting, over a file only the log indexer reads.

It is not hypothetical or hard to produce: any partial write leaves one, and a timestamp the type cannot parse is enough. I hit it three times while testing #10894 and lost most of an afternoon to it before recognising what it was.

An unparseable cursor now reads as **absent**, which is also the honest answer — nothing can be said about what the index holds, which is exactly what a missing cursor means, and the rebuild that answers it writes a good one back. A failure to *fetch* still propagates: retrying reaches the object, whereas calling it absent would rebuild the whole window over a network blip.

## The queue's writes are covered against a real database

Every reviewer on #10894 said the same thing: the conversion, the mark and the release were verified by hand on a running instance and by nothing else. That is how several of their bugs survived a round — the shapes are easy to reason about wrongly, and cheap to run for real.

`#[sqlx::test(migrations = "../migrations")]` already exists in this repo and the indexer's own tests already use an in-memory object store, so the harness was mostly a matter of putting the two together. The conversion and the mark move out of the middle of their callers to be reachable at all.

Three tests: the sentinel split either side of the window and its idempotence, that marking takes only the pairs it names and not a row sharing a host or a minute, and that releasing takes only files the store vouches for and only rows still queued.

## A read pays for the dedupe only when it can hold repeats

Collapsing repeats costs an aggregate over the partition in front of what is otherwise a streaming top-N, on the interactive search route. #10894 took that cost unconditionally and recorded it as measured — 68–98 ms for 444 hits on a ~19k-row store.

A line belongs to one partition, so two copies of it can only be two objects within the same one. `plan_files` already reads every manifest, so it now reports whether any single partition contributed more than one object, and the three read paths collapse only when it did. After compaction has merged a day there is one object in it, which is the steady state for everything but the day being written.

**Measured** on 20,000 rows in one compacted object, the same query planned both ways, best of three:

| plan | |
|---|---|
| `select` — what a single-object partition now does | **16.3 ms** |
| `distinct_on` — what every read did before | **342.7 ms** |

About 21× on the steady state, and the aggregate is the whole difference: identical data, identical filter and limit, only the collapse removed. The day being written still pays it, which is correct — that is the day whose objects can hold repeats.

## Test plan

- [x] `cargo test -p windmill-indexer --features enterprise,parquet,private` — 31 pass, including four database-backed ones
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features,private`
- [x] Builds under the feature set the server runs (`quickjs,tantivy,private`), binary included

Each new test was checked to fail when the thing it guards is removed, rather than assumed to guard it:

| guard | mutation that must fail it |
|---|---|
| unreadable cursor reads as absent | propagate the parse error instead → fails with the error that took the server down |
| marking is exact on `(hostname, log_ts)` | `WHERE hostname = …` alone → fails; `WHERE log_ts = …` alone → fails |
| retrieval collapses repeats | remove the dedupe from `read_log_file` → fails with the line rendered twice |
| the rebuild's scan collapses repeats | remove the dedupe there → fails with two lines becoming four |
| the overlap discriminator | one object reports no overlap, two report overlap |

The database tests place rows relative to the statement's own `NOW()`, never an absolute date, so they cannot start failing when the calendar moves past a hardcoded window.

**Not covered:** the spill path under memory pressure, which needs a partition larger than the query budget.


rubenfiszel:
Measured the third item rather than leaving it as a claim about plan shape. On 20,000 rows in one compacted object, the same query planned both ways, best of three:

| plan | |
|---|---|
| `select` — what a single-object partition now does | **16.3 ms** |
| `distinct_on` — what every read did before | **342.7 ms** |

About 21× on the steady state. Identical data, identical filter and limit, only the collapse removed — so the aggregate is the whole difference. The day being written still pays it, which is correct: that is the day whose objects can hold repeats.

I had written "skips the aggregate" without a number, which was a claim about the plan and not a demonstrated saving. It is in the description now.

rubenfiszel:
Both fixed.

**The conversion test was a time-bomb** (codex and pi, independently). Entirely right, and it is the worse kind of test failure: it passes today and starts failing in CI once the calendar moves past the window, long after the change that appears to have caused it. Rows are now placed a number of seconds back from the statement's own `NOW()` — the same clock the conversion compares against — and the one test that names a row by its timestamp reads back the timestamp the row actually got, rather than assuming one.

**Two orphaned `.sqlx` entries** (claude). Correct, and exactly two. Moving a query into a function changes its indentation, and sqlx keys on the literal text including whitespace, so the extraction left the pre-move copies behind. I checked them against what `cargo sqlx prepare` actually emits rather than a text search — the search I tried first reported seven, five of them live queries it could not match through line continuations.

On the caveat about the EE code not being reviewable from the checkout: that is the same constraint as on #10894, and the substance here is again entirely in `windmill-indexer`. The three changes are small and independent if that helps read them from the diff alone — a `match` arm in `read_ingest_state`, two queries lifted out of their callers with tests, and a boolean threaded from `plan_files` to three read sites.

rubenfiszel:
**Pi: good to merge.** Both of codex's nits fixed, and the first was sharper than a nit.

**The pair-exactness test did not test pair exactness.** Exactly right: the two rows shared neither half of the key, so it would have passed with a host-only *or* a timestamp-only predicate — the two ways that statement can be wrong. It now places a row sharing the host at another minute and a row sharing the timestamp on another host, and asserts over every row rather than the two it names. I checked both mutations fail it:

```
WHERE hostname = c.hostname                      -> FAILED
WHERE log_ts   = c.log_ts                        -> FAILED
both                                             -> ok
```

**The comments narrated the outage rather than stating the rule**, which is the AGENTS.md line about describing the code as it is. Rewritten. The caller's was worse than style — it still said `read_ingest_state` answers `None` only for a genuinely absent object, which the parse fallback in this very PR had made untrue, so a reader would have taken the wrong invariant from it.

**One thing added beyond the round**: the rebuild's scan now has its own dedupe case. That read is where double-indexing came from before the dedupe existed, and after this PR it shares a conditional with the other two — so it deserves a guard rather than inheriting confidence from a sibling. Mutation-checked: remove the dedupe there and it fails with two lines becoming four.

31 tests, `SQLX_OFFLINE` workspace check clean, and the binary builds under the feature set the server runs.

rubenfiszel:
Both nits fixed — same category, comments that no longer described the code.

Pi's was the load-bearing one: the `select!` arm still gave "an unreadable cursor" as its example of a pass that gives up, which *this PR* made untrue. That case now reads as absent and is answered by a rebuild, so the comment was teaching the next reader an invariant the code had stopped holding. It names a store it cannot reach instead.

Codex's two test comments described when their bugs existed rather than what the code does, and the fixture's odd timestamp is now described by what makes it odd — an offset the field's type will not parse — rather than by what it once broke.

31 tests, `SQLX_OFFLINE` workspace check clean.

✅ Review round clean @ 80602635d2d5f93ca7be33504c27a0b9f289e921 — nit-only verdicts at 25301cd149dc (claude good to merge, codex and pi nit-only); nits addressed in 80602635d2d5

PR #10906, opened 2026-08-31
perf: add service log documents to the index one batch at a time

rubenfiszel:
## Summary

The service log ingest called `index_log()` once per log line, doing one
`tokio::task::spawn_blocking` **per document**, i.e. a scheduler round trip per line. This adds the
tantivy documents in the same batches the parquet store rows are already flushed in, which is what
`rebuild_index_from_store` in the same file has always done.

## Measured

Same fixture both sides (45 files x 5,000 JSON lines = 225,000 lines per pass, 4 hosts, filesystem
object store), timed from the `log_file` query returning to the last batch being published:

| | median | mean | range | n |
|---|---|---|---|---|
| before | 34.18 us/line | 34.12 | 33.81-34.32 | 6 |
| after | **23.24 us/line** | 23.23 | 23.16-23.30 | 6 |

1.47x, about 11 us/line. A per-row breakdown taken on an earlier base put the batched
`add_document` at 3.39 us/row against a parquet encode of 5.12 and a parse plus fetch of 14.44, so
what the batching removes is the hop, not the tantivy work, and the remaining ~19 us/line is the
parse and the encode that no batching can help.

The useful way to read this is headroom rather than latency. At the 300s refresh period a pass is
capped near 8.8M lines before, and near 12.9M after: the change buys roughly 47% more ingest
ceiling before the indexer starts falling behind. Below that ceiling it saves about 3 seconds on a
300s cycle, which nobody sees.

## Changes

- `index_log` becomes `build_doc`: pure, returns a `TantivyDocument`, no blocking hop.
- New `add_documents`: one `spawn_blocking` for a whole batch, taking the writer lock once.
- New `flush_documents`: routes a batch to the day currently open, and **errors** rather than
  dropping documents if no day is open. That arm is unreachable today, but the caller checkpoints
  on `Ok`, which marks the rows indexed and deletes their raw files, so the invariant is asserted
  rather than assumed.
- `index_docs`, a document buffer parallel to `store_rows`, drained by `flush_documents` after
  every successful `flush_rows` at all three sites: the day boundary, the between-files
  `LOG_STORE_MAX_BUFFERED_ROWS` cut, and the end of the pass. So a batch reaches the store and the
  index together, or neither.
- `writer_for_day` is now called once per file rather than once per line inside `index_log`. This
  is what opens the day that the boundary check and every flush read back off `self.active`.
- `whole_file_read` is gone and `ingested.push(path)` is unconditional: nothing inside the line
  loop can fail any more, and `ingested` is only ever acted on after a checkpoint that now follows
  a document flush.
- `index_docs` also cuts **inside** a file at `LOG_STORE_MAX_BUFFERED_ROWS`, which the store rows
  deliberately do not. The store's cut has to be reproducible across a retry because its object
  names are derived from what it holds; documents have no such constraint, since a failed pass
  rolls the writer back wholesale and the retry re-adds from the same rows. That bound is what
  stops one oversized file, which the between-files cut cannot bound, from being resident twice.

Incidentally fixes a pre-existing under-count: the old code pushed the row to `store_rows` and
*then* called `index_log`, so a failed `index_log` left a row in the store that `total_lines` never
counted, skewing `index_is_behind()`.

Known trade-off, bounded rather than removed: within its cut `index_docs` still holds a second copy
of each line's `searchable_text` alongside the `LogRow` it was built from, so peak ingest memory
rises by roughly one batch of line bodies. Building the documents inside the blocking hop from
`&rows` would remove that but change behaviour, since the document's timestamp is the **file**
timestamp while `parse_line` sets the row's to the line's own; preserving current semantics needs
per-row `host_key` and `file_ts` anyway, and the zip-alignment hazard that introduces sits in the
path that governs raw-file deletion. The within-file cut above is what keeps the increase bounded
instead of proportional to the largest file.

## Test plan

No automated test: the existing `mod tests` in this file covers only pure helpers and none of them
reach `refresh_log_index`, while a test of the ingest loop needs a live tantivy writer, an object
store and a `Pool<Postgres>`. Per AGENTS.md that is a ceremonial test, so this ships none. The
honest statement is that "documents reach the index before the checkpoint marks the rows" is
unguarded by CI and rests on review. Verified on a running instance instead:

- [x] Single-day pass, 225,000 lines crossing the between-files buffer cut: index holds exactly
      225,000 documents, all raw files deleted, zero rows left with `indexed_at IS NULL`, and
      1,350,000 planted matched 1,350,000 indexed across six passes.
- [x] **Midnight-spanning pass** so all three flush sites fire in one pass: 22 files before
      midnight and 23 after. The day-boundary flush and its sealing checkpoint ran (22 raw files
      deleted at the seal, 23 at the end), both `dt=` partitions were created, and the totals are
      exact: 225,000 overall, 110,000 in the first day's window, 115,000 in the second's.
- [x] **Forced flush failure** (store tree made read-only mid-pass): the pass aborts, the rollback
      runs, and afterwards all 45 raw files are still present and all 45 rows are still
      outstanding.
- [x] **Retry after that failure**: exactly 1,575,000 documents for 1,575,000 planted lines, no
      duplicates, nothing left outstanding.
- [x] **One oversized file**, 500,000 lines in a single file, i.e. 2.5x the buffer bound and the
      case the between-files cut cannot bound: the within-file document cut fires twice and the
      index holds exactly 500,000 documents.
- [x] `cargo check --features enterprise,private,parquet,tantivy` clean.

Note for reviewers: a time-range count on the exact boundary timestamp reads 5,000 short, because
`scope_query_to_time_range` uses `Bound::Excluded` and drops the file sitting on the bound. That is
pre-existing query semantics, unchanged here; moving the bound one second out returns the full
count.

---
## Summary by cubic
Updates `backend/ee-repo-ref.txt` to the EE commit that batches service log document indexing, so tantivy documents are added in the same batches as the parquet store rows are flushed, instead of one `spawn_blocking` per line. Measured ingest time drops from ~34 us/line to ~23 us/line, which raises the ingest ceiling by ~47% before the indexer starts falling behind.

**Changes**
- Turns `index_log` into a pure `build_doc` and adds batch `add_documents`/`flush_documents` calls.
- Flushes documents and store rows together at day boundaries, buffer cuts, and end of pass; if no day is open, the batch errors instead of dropping documents.
- Fixes an existing under-count where a failed `index_log` left a row in `store_rows` that `total_lines` never counted, skewing `index_is_behind()`.
- Adds a within-file document buffer cut at `LOG_STORE_MAX_BUFFERED_ROWS` so one oversized file cannot double memory.
- Peak ingest memory rises by roughly one batch of `searchable_text` line bodies.

**Verification**
- No automated test was added; manual verification covered exact counts, a midnight-spanning pass, a forced flush failure with retry, and an oversized file.
- A time-range count on the exact boundary timestamp reads 5,000 short because `scope_query_to_time_range` uses `Bound::Excluded`; that is pre-existing and unchanged.

<sup>Written for commit a351f068eaea10df435cdd898a1c00be3ff1ab4f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ b5dd05baced6251735941696c84894ca5a039519 — nit-only verdicts at f684452134161fb62b4dc3cb84649ff13bf7d2fa; nits addressed in b5dd05ba (EE f9a0b98)

Round verdicts at f6844521: Codex "Mergeable, but should ideally address nits", Pi "Good to merge", Claude posted no verdict because it could not read the change (see below).

**Codex's P2 is fixed.** It was right, and self-inflicted: its own earlier P1 asked for a within-file bound on the document buffer, and once I added that, the comments claiming the store and index buffers always flush together stopped describing the code. Both are reworded to state what actually holds, which is that documents may reach the writer ahead of their rows but never later than the checkpoint that marks those rows indexed.

**For the Claude reviewer**, which correctly reported it could not review this: the EE checkout in the runner is pinned at the base commit and fetching the branch is outside the job's permission set, so its inability is a sandbox limit rather than a finding. It asked for one confirmation, `docs/enterprise.md` step 5, that the ref is the tip of the EE branch and not main. Confirmed:

```
EE branch tip: 1f6e6c0cdd07c4857fde67e18a0092d6fdf74b97
EE remote tip: 1f6e6c0cdd07c4857fde67e18a0092d6fdf74b97
ee-repo-ref:   1f6e6c0cdd07c4857fde67e18a0092d6fdf74b97
```

(all three have since moved together to `f9a0b98` with the comment fix). The invariants it recorded from the base file are the right ones, and they are the ones the change preserves: `flush_documents` runs after each successful `flush_rows` and therefore before every `checkpoint`, while `self.active` still holds the day the buffered documents belong to. The companion PR is windmill-labs/windmill-ee-private#762, where the diff is readable.

Two findings from the local Codex pass are deliberately not addressed here, both because they are not this change: a P0 about `wmill sync pull --keep-deleted` deleting local dbt descriptors (`cli/src/commands/sync/sync.ts`, from #10878, already on main and worth someone's attention on its own), and a P2 about comments referencing the pull request description and counting review rounds, which came in with #755.

PR #10911, opened 2026-08-31
fix: keep a local dbt descriptor under sync pull --keep-deleted

rubenfiszel:
## Summary

`wmill sync pull --keep-deleted` could delete a local dbt descriptor (`<project>__dbt/wm_dbt.yaml`), losing the warehouse, command and run arguments it held — the exact outcome the flag exists to prevent.

A dbt project's descriptor is optional, and both sides spell "this project names no descriptor" as empty content rather than as an absent file. So a descriptor the remote no longer names reaches the pull as an `added`/`edited` change with empty content, never as a `deleted` one. `dropDeletions` (added in #10878) filters on `name === "deleted"`, so it left the change in place and the apply loop's `rm` ran anyway. The dry-run preview reinforced this, listing it as `~ script f/…/wm_dbt.yaml`.

## Changes

- Extract the descriptor-removal condition out of the pull apply loop into `removesDbtDescriptor(change)`, so the apply loop and `dropDeletions` read one predicate and cannot drift — that drift is what caused this.
- `dropDeletions` now counts a descriptor removal as a deletion on pull, so `--keep-deleted` strips it.

One deliberate asymmetry: **push is untouched.** There the descriptor is a script's *content*, so an empty one updates the remote script in place rather than deleting a remote item. Making `push --keep-deleted` refuse it would change what push means.

Both change shapes count, not just `edited`. A stateful pull compares `.wmill` rather than the working tree, so a descriptor absent from that map still arrives as `added` while a real file with real settings sits on disk — and the apply loop deletes from the working tree. An earlier revision of this PR counted only `edited` (to keep the `keeping N item(s)` line from over-reporting on a first clone) and reintroduced the data loss on that path, silently. The safe rule is worth the cosmetic over-count.

## Test plan

Three unit tests in `cli/test/dbt_optional_descriptor_unit.test.ts` (the two pull ones fail without the fix; the push one pins the deliberate asymmetry). Verified end-to-end against a local backend:

- [x] Stateless: project pushed with no descriptor, then a local `wm_dbt.yaml` with warehouse/command/args — `sync pull --yes --keep-deleted` keeps it and reports it as kept
- [x] Stateful (`stateful: true`): `.wmill` not knowing the project while the working tree holds a populated descriptor — the file survives, and is now reported instead of silently removed
- [x] `sync pull --yes` without the flag still removes it — unchanged
- [x] Fresh clone of a descriptor-less project: no stray `wm_dbt.yaml`
- [x] A real remote descriptor still round-trips: push, delete locally, pull restores it
- [x] Non-dbt `--keep-deleted` both directions: remote script survives a push, local-only file survives a pull
- [x] CLI unit suite 1100/1100; `tsc --noEmit` unchanged from baseline (18 pre-existing errors, same files)

### DIFF
--- a/cli/src/commands/sync/sync.ts
+++ b/cli/src/commands/sync/sync.ts
@@ -3175,15 +3175,42 @@ export function untrackedDatatableMigrationDeletions<
 }
 
 /**
+ * Whether a pull change removes a local dbt descriptor. A dbt project's
+ * descriptor is optional and the remote spells "this project names none" as
+ * empty content, so its removal reaches the apply loop as an add or an edit
+ * whose content is `""` — a `deleted` change is never produced for it.
+ */
+function removesDbtDescriptor(change: Change): boolean {
+  return (
+    isDbtDescriptorPath(change.path) &&
+    ((change.name === "added" && change.content === "") ||
+      (change.name === "edited" && change.after === ""))
+  );
+}
+
+/**
  * `--keep-deleted`: strip every deletion from the changeset, in place, so the
  * sync only adds and updates. A path missing on one side is not on its own
  * evidence that it should go from the other — a partial clone, a scoped
  * checkout or an item authored in the UI all read as deletions here.
+ *
+ * A dbt descriptor the remote no longer names counts too, but only on a pull,
+ * where applying it removes a local file. A push removes nothing: the descriptor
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ 47df3253b9cd750758e1ec597df3bb3df1c61dd7

PR #10912, opened 2026-08-31
fix: connect to dev server instead of localhost

nferch:
## What does this PR do?

`wmill app dev` defaults to listening on localhost, which is good from a security perspective. You can make it listen to a different address with the `--host` parameter, however this doesn't fully work as the Websocket connection is hardcoded to use localhost.

In my particular case my web browser is running on a separate machine from my editor and OpenCode, so this is necessary to be able to access the dev server across the network. This is somewhat of a security concern, but I have already accepted some of that risk by electing to use the `--host` parameter.

## Related issue

---
## Summary by cubic
Fixes the dev server WebSocket connection so it works when the dev server runs on a non-local host or behind an HTTPS proxy/tunnel.

- WebSocket URL now derives from `window.location`; the unused `port` parameter was dropped from `wmillTsDev`.
- Scheme uses `wss` over HTTPS to avoid mixed-content blocking.

<sup>Written for commit 36826b839fee2c4b37c5a8f673d5deb3ab8166e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


nferch:

PR #10914, opened 2026-08-31
feat(ai-chat): make reusable skills ai_skill resources you select per workspace

rubenfiszel:
## Summary

Skills for the AI chat move off a bespoke `ai_skill` table onto ordinary Windmill resources, and which ones the chat may use becomes a per-user, per-workspace choice instead of "every skill in the workspace, for everyone, always".

The old design had three problems that all came from the storage choice:

- **No selection.** Every skill in the workspace was listed in every user's system prompt, in every session. At the 100-skill cap that is a lot of unfiltered context.
- **Admin-only authoring.** A non-admin could not write a skill for themselves.
- **No platform affordances.** No ACL or folder, no version history, and absent from workspace export, git-sync and the CLI — so skills did not move dev→prod, were not reviewable in a repo, and could not be restored.

Resources already solve all three. A skill is now a resource of type `ai_skill`, a *file* resource type (`format_extension = 'md'`) whose `value.content` is the SKILL.md body, with the description on the resource itself so the assistant can judge relevance before loading the instructions.

## Screenshots

The `+` menu, where a skill is turned on for you:

skills menu

The drawer, with the authoring flows ported from the deleted workspace-settings card:

skills drawer

New skill, opening on an editable SKILL.md sample rather than placeholder text — the path's last segment tracks the frontmatter `name` until you edit the path yourself, and Reset restores the sample:

new skill

Only selected skills appear as `/` commands:

slash picker

## Changes

**Storage.** A migration seeds the `ai_skill` resource type under `admins`, copies every `ai_skill` row to `f/skills/<name>` (creating the folder with the old admin-only-write ACL), and drops the table. A destination that is already taken lands on `<name>_migrated` rather than being skipped and then dropped.

**Removed.** `windmill-api/src/ai_skills.rs`, its route nest, the `ai_skills` token scope and `ScopeDomain`, the four openapi paths, and the `AiSkillsSettings` workspace-settings card. Skills are covered by the existing `resources` scope.

**Selection.** `createEnabledPathsPreference` in `copilot/chat/enabledPathsPreference.ts`, keyed `workspace:email` in localStorage — now shared with the MCP `enabledServers` module, which was the same code.

**Chat.** Only selected skills reach the system prompt, listed by resource path. `read_skill` takes a path and enforces the selection rather than trusting the prompt listing. `/name` expands to the path; when two folders answer to the same name the picker inserts the full path and submission refuses to guess.

**`format_extension` plumbing.** `ai_skill` is the first *shared* resource type to use `format_extension` — the column has existed since `20240910151523_resource_type_file_format` but was only ever reachable for workspace-local types, so no hop in the chain carried it. This PR teaches `HubResourceTypeRaw`, the sync's comparison and upsert, and the `wmill` CLI. `update_resource_type` also never wrote the column at all, so an edit could not set or clear it; it now can (doubly-optional, so absent leaves it alone and `null` clears it).

## Companion PRs

| Repo | PR |
|---|---|
| `windmill-ee-private` (telemetry counters) | windmill-labs/windmill-ee-private#764 |
| Hub service (`format_extension` column + endpoints) | windmill-labs/windmillhub#193 — **merged, released as v2.13.0, deployed** |
| `windmill-integrations` (push payload + the type) | windmill-labs/windmill-integrations#180 |

## Telemetry

Two `feature_usage` pairs, registered in the EE PR above: `('ai_session','skill_toggle')` keyed `on`/`off`, and `('ai_session','skill_read')`. No skill names or paths — those are user content.

## Test plan

- [x] `cargo check` and `cargo check --features enterprise,private`
- [x] `npm run check` — 0 errors (the 4 remaining are main's, in the Hub-project wizard)
- [x] Frontend suites pass, including `mcpTools.test.ts` after the shared-preference refactor
- [x] Migration applies, and down→up round-trips content; with an unrelated resource planted at `f/skills/deploy-check` the skill lands at `deploy-check_migrated` and the planted resource is untouched
- [x] Driven in the browser against a live DeepSeek model: asked in plain language, the model chose `read_skill` on its own and followed the SKILL.md body; `/triage-bug …` did the same via the path expansion
- [x] A readable-but-unselected skill never reaches the prompt (verified by inspecting the outgoing request body)
- [x] Folder import against a directory of `SKILL.md` files; with a non-skill resource planted at the destination the import is refused rather than clobbering it (400, resource intact)
- [x] `format_extension` round trip: a stubbed hub serving `"md"` reaches the DB through `wmill hub pull`, and serving `null` clears it; a real sync against the deployed hub imports 272 types and leaves a local `ai_skill` at `md`
- [ ] CI review round

## Note for reviewers

Skills stay behind the AI Sessions beta gate. The migration copies existing rows, but a skill that was implicitly on for everyone is off until each user selects it — the enabled set is browser-local, so it cannot be pre-seeded.

---
## Summary by cubic
Moves AI chat skills off the bespoke `ai_skill` table onto `ai_skill` resources, making skill selection a per-user, per-workspace choice instead of injecting every workspace skill into every prompt. Skills gain folder ACLs, version history, export, git-sync, and CLI support, and anyone with resource write access can author them. The skill list is paginated, descriptions are truncated for the prompt, and every skill action, including the picker drawer, targets the workspace the chat runs in; refreshes for a workspace the chat has left are discarded.

**Migration**

- The migration seeds the `ai_skill` resource type under `admins`, copies existing skills to `f/skills/<name>`, and leaves the `ai_skill` table in place for a later release to drop; its down migration removes only the system-seeded type.
- The old `ai_skills` API scope and endpoints are removed; use the existing `resources` scope.
- Skills are off by default; each user selects the ones they want, and the selection is browser-local.
- `ai_skill` is a file resource type (`format_extension = 'md'`); hub sync and resource-type edits can set or clear the extension, sync refuses it on fileset types, and `read_skill` rejects unselected paths.
- The migration guards folder owners, and the import reports skills it had to skip because a destination was taken.

<sup>Written for commit 2d311f3cd1d89fbdaad3e9f9656ba43ece31b1b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
✅ Review round clean @ e37b9b28a0e5808fd068cd8214df578afd222507 — nit-only verdicts at e37b9b28a0e5808fd068cd8214df578afd222507 (codex + claude "Mergeable, but should ideally address nits", pi "Good to merge"); nits addressed in 2d311f3cd1d89fbdaad3e9f9656ba43ece31b1b6

Issue #10913, opened 2026-08-31
bug: supplied script locks never populate lock_hash, amplifying importer relocks


ryan-gray-db:
## Describe the bug

Creating a script with a caller-supplied nonempty `lock` stores the lock on `script` but does not upsert `lock_hash(workspace_id, path, hash_script(lock))`. Worker-generated locks do upsert that row.

`try_skip_relock` intentionally treats a missing imported-script hash as changed (`COALESCE(dm.imported_lockfile_hash = lh.lockfile_hash, false)`), so every affected importer relocks on every upstream change. The five-second debounce does not prevent already-dispatched jobs from overlapping. On public/CE builds, the dependency concurrency key in `windmill-queue` is assigned and enforced only behind the `private` feature, so same-path jobs can execute concurrently.

This remains reproducible on current main after v1.800.0.

## To reproduce

1. Create an imported script through `POST /api/w/{workspace}/scripts/create` with a nonempty supplied `lock`.
2. Query `lock_hash` for that workspace/path. No row is created.
3. Create a second script that imports the first and let its dependency-map edge be recorded.
4. Update the imported script repeatedly within the debounce window.
5. Observe that the importer cannot satisfy the relock-skip predicate. On a CE worker fleet, multiple dependency jobs for the same workspace/path can execute at once.

A production incident had 147 affected dependency-map edges across 32 imported scripts, blocking relock skipping for 82 importers. In one hour those importers received 2,132 dependency jobs; 2,088 executed relocks and none skipped. One importer reached seven concurrent relocks.

## Expected behavior

- The script-create transaction persists the Rust `hash_script(lock)` value whenever a nonempty supplied lock is accepted, before relative-import processing is scheduled.
- Relative-import relocks for one workspace/importer path cannot overlap in public builds.
- After generation, an unchanged lock/modules result repairs `lock_hash` and dependency-map snapshots but does not rewrite the script row, emit deployment metadata, or fan out another dependent cascade.

## Validated patch

A rebased public patch is available at:

- Branch: https://github.com/ryan-gray-db/windmill/tree/codex/eng-9106-windmill-lock-hash
- Compare: https://github.com/windmill-labs/windmill/compare/main...ryan-gray-db:windmill:codex/eng-9106-windmill-lock-hash
- Commit: https://github.com/ryan-gray-db/windmill/commit/a0553b7d9

The patch:

- upserts supplied-lock hashes in the create transaction;
[...]

PR #10915, opened 2026-09-01
fix: record supplied script lock hashes so importers can skip relocking

hugocasa:
## Summary

Creating a script with a caller-supplied `lock` — a CLI push, a git-sync deploy, any
`POST /scripts/create` carrying a lockfile — stored the lock on `script` but never wrote the
matching `lock_hash(workspace_id, path, hash_script(lock))` row. Only worker-generated locks did.

`try_skip_relock` deliberately treats a missing hash for an imported script as *changed*
(`COALESCE(dm.imported_lockfile_hash = lh.lockfile_hash, false)`), so no importer of such a script
could ever satisfy the skip predicate. Every deploy of the imported script relocked every importer,
forever.

## Changes

**Record the lock's hash in the create transaction**
(`windmill-api-scripts/src/scripts.rs`), at both places that accept one.

- **On create.** Including the empty lock a codebase or a language with no lock generation carries
  — the worker writes `hash_script("")` in that situation, and a path going from a real lock to an
  empty one has to stop matching what its importers recorded. Only a lock left to a dependency job
  is skipped, because that job writes it.

- **On the `skip_if_noop` path the CLI uses.** That returns before the insert and queues no
  dependency job, so a path pushed but never edited gets a row from neither writer. This is what
  makes the fix reach scripts that are already deployed: the importer side of the comparison,
  `dm.imported_lockfile_hash`, snapshots whatever `lock_hash` holds at relock time
  (`ScopedDependencyMap::patch`), so while an imported script has no row every relock of an
  importer re-records NULL and that importer can never skip — however often it is relocked, and
  whichever of its imports triggered it. A stable shared helper deployed once with a supplied lock
  never takes the create path again, so without this write its importers stay unskippable
  indefinitely.

**Record `lock_hash` for a workspace clone** (`windmill-api-workspaces/src/workspaces.rs`). A clone
never had one, so every fork started in exactly the state this PR fixes and cost one redundant
relock round per importer.

Derived from the locks the clone holds, not copied from the source's rows. A source row is only as
current as the last write to it, and the bug this PR fixes is precisely what leaves one behind: an
install upgrading from before this carries paths whose worker-generated `H1` was never replaced
when a supplied `L2` was deployed. Copying that row hands the fork a hash describing a lock the
path no longer holds — an importer relocked there resolves against `L2` but records `H1`, and a
later deploy of `L1` then reads as unchanged and makes it skip. Hashing the clone's own locks
cannot express that, and it needs no snapshot reasoning about what the source did mid-clone.

`DISTINCT ON (path) … ORDER BY path, created_at DESC` takes the live version per path, the idiom
the rest of the file uses, since nothing enforces one live row per path — the path-conflict check
at `scripts.rs:1260` takes no row lock, so two concurrent creates of a fresh path both pass it.

Both writes go through one `record_lock_hashes` in `windmill-dep-map`, next to the `lock_hash`
reads that `ScopedDependencyMap` already does. The other two writers in the codebase
(`worker_lockfiles.rs:478`, `workspace_dependencies.rs:100`) keep their own copy of the upsert on
purpose: each folds it into the statement that writes the lock itself, which is what makes the lock
and its hash land together. These two have no such statement to fold into. Since `script.lock` is
unbounded and a workspace holds one per script, the clone streams its rows and hashes each lock as
it arrives rather than materializing them.

The upsert skips the write when the hash is unchanged
(`WHERE lock_hash.lockfile_hash IS DISTINCT FROM EXCLUDED.lockfile_hash`). That is the common case
for the no-op push, which is the mode a git-sync of an unchanged workspace runs in, so without it
every sync cycle would cut a row version per script for nothing. Repeated paths are collapsed
last-wins before the insert: Postgres rejects an entire statement that resolves a conflict on one
key twice, and the `WHERE` does not rescue it.

`dependency_map.imported_lockfile_hash` is deliberately *not* copied: it records what an importer
resolved against when it was last locked, which nothing at clone time can establish for the version
the clone received. Left NULL, every importer relocks once and re-anchors both sides to what the
clone actually holds.

No change to lock generation, and none to what a relock does once it decides to run. The only
behavior that moves is which relocks are skipped.

## Not here: serializing same-path relocks

The issue's second bullet asks for relative-import relocks of one path not to overlap, on the
grounds that the dependency-job concurrency key is `private`-gated. Every shipped build already
has it: the images build with `features=ce` or `features=ee`, and
`ce_core = ["oss_core", "private", "operator"]`, so each dependency job gets
`concurrency_key = dependency:{workspace}/{path}` at limit 1, enforced at pull.

The only build without it is the `oss` feature set, which also sets `no_auth` and where `private`'s
absence removes debouncing and concurrency limiting wholesale — `update_concurrency_counter` is a
stub that always allows, `apply_concurrency_limit` is never called, `maybe_debounce` is never
called. A bespoke lock on this one path would patch a hole in a wall with much larger ones. If
same-path overlap is real on a current fleet it is the `MIN_VERSION_SUPPORTS_DEBOUNCING`
rolling-upgrade window, and it belongs in that gate rather than beside it.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test scripts --test workspaces` (8 + 6) —
      includes a new test covering both writes: that a supplied lock's `hash_script` value lands in
      `lock_hash` on create, and that a no-op push repairs the row for a path that has none. It
      fails on `main`, and on this branch with either write removed, with "no rows returned by a
      query that expected to return at least one row".

      The clone derivation is **not** covered by a test: the only fork fixture is inside
      `test_workspace_endpoints` behind `#[cfg(feature = "enterprise")]`, so it would not run in
      the CE job. It was exercised by hand instead, per the next item.
- [x] `cargo test --features python,private,quickjs --test relock_skip --test dependency_map
      --test script_modules` (2 + 8 + 3), all unmodified.
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`, `--all-targets` for
      the crate with the new test query, and `cargo check --features enterprise,private,native_trigger --tests`.
- [x] The pre-upgrade stale row, on a running instance. `f/t/p` deployed with `LOCK_TWO`, its
      `lock_hash` then set by hand to `hash(LOCK_ONE)` — the state an install carries out of the
      bug this PR fixes. Forking the workspace records `hash(LOCK_TWO)` for that path in the fork,
      matching the lock the fork actually holds, while the source keeps its stale row until its
      next deploy. Copying the source's row would have carried `hash(LOCK_ONE)` across.
- [x] On a running instance, a three-level bun chain `imported → importer → importer2`: creating
      `f/test/imported` with a supplied lock writes its `lock_hash` row, and the importer's
      dependency-map edge records that hash instead of NULL, so its next dependency job logs
      `Skipping relock - imported lockfiles unchanged` where it previously always relocked. A real
      dependency change still cascades through all three levels.
- [x] Workspace fork of the same chain: the fork's `lock_hash` rows match the source's, its
      dependency-map edges come across with no hashes, and redeploying the imported script in the
      fork twice logs `Relocking script` then `Skipping relock`.
- [x] A fork of 253 locked scripts records exactly 253 rows, none lost or duplicated, and the
      fork's hash for the stale path equals what the create path independently wrote for another
      script carrying that same lock.
- [x] Memory: a workspace of 150 scripts holding 4 MB of lock each (600 MB aggregate), forked while
      sampling the API's `VmRSS` — peak 131 MB against 108 MB at rest, so ~23 MB for 600 MB of
      locks, and all 150 rows recorded.
- [x] The upsert's two guards, against the live table: a repeated path aborts the statement with
      `ON CONFLICT DO UPDATE command cannot affect row a second time` both with and without the
      `WHERE`, so the dedup is what prevents it; re-recording the same hash reports `INSERT 0 0`
      while a changed one reports `INSERT 0 1`. Redeploying a path from one supplied lock to
      another still moves its row.
- [x] Empty lock: redeploying a path from a bun script with a supplied lock to a `postgresql` one
      moves its `lock_hash` from the real lock's hash to `hash_script("")`, matching a freshly
      deployed `postgresql` script.

Relative-import bundles are unaffected by importers now skipping: the bun bundle cache key is
`content + lock + (path, deployed hash) for the whole transitive relative-import closure`, so a new
version of an imported script changes the key whether or not its importer relocked.

Pre-existing and untouched here: nothing anywhere deletes a `lock_hash` row, on `main` or on this
branch, so a row outlives the archive or rename of the path it names. This PR adds rows for one
more class of path rather than changing that lifecycle; cleaning it up means touching the archive,
rename and delete paths and belongs on its own.

Also pre-existing: `lock_hash` has one row per path, and no writer checks that the version it
describes is the path's live head. Concurrent creates of one path both succeed — the advisory lock
at `scripts.rs:1196` only covers `auto_parent`, and the path-conflict check takes no row lock, so
racing two `POST /scripts/create` at a fresh path leaves two live versions every time (5 of 5
locally). `created_at` is the transaction's start time while the surviving `lock_hash` is decided
by commit order, so a transaction that starts first and commits last leaves the row describing a
lock that is no longer the head, and an importer resolving against the head records a hash that
belongs to the other version.

The same hole is reachable on `main` through the worker: `worker_lockfiles.rs:474` writes
`(workspace_id, script_path)` with the hash of whatever version its own dependency job locked, and
two live versions mean two such jobs finishing in an order unrelated to `created_at`. That window
is wider than this one, since the jobs are asynchronous rather than two statements in one
transaction.

Not addressed here because the fix is not in this table: while a path holds two live versions every
"the live version" read in the codebase is arbitrary, so guarding only `lock_hash` treats a symptom.
Closing it means serializing same-path creates — extending that advisory lock to every create —
which changes deploy concurrency for all scripts and wants its own change and its own testing.

---

### 1. An unchanged relock still redeploys and fans out

The issue's third bullet, and the larger half of the amplification. Skipping the *generation* is
not the same as skipping the *deploy*: a relock that skips still writes a new script version,
records deployment metadata, commits to git-sync and queues a dependency job for every importer —
so one upstream change walks the whole graph in cheap-but-real jobs and duplicate commits.

The shape of the fix: after generation, compare the result against the version the relock was
cloned from. `clone_runnable` mints a fresh version at pull time whose `lock` is NULL, so the
comparison has to reach through `parent_hashes[1]` rather than at the job's own version — the
reported patch compared against the latter, which is why its convergence path could never match.
When lock and module locks are equal, refresh `lock_hash` and the dependency-map snapshot and stop.

Measured on `relock_skip_on_workspace_deps_redeployment`, instrumented to count completed jobs:
an unchanged workspace-dependency redeploy goes from **18 jobs to 5**; a changed one still fans out.
(That number belongs entirely here, not to this PR — the fixture deploys its scripts without a
supplied lock, so the create path is never exercised in it.)

Needs deciding as part of it: it changes observable behavior. No git-sync commit for a version
byte-identical to its predecessor is the point, but CI tests are a separate question — a test names
the item it covers, so a test on an importer is not among the ones its imports' deploys trigger,
and an unchanged lock only says the dependency set held, not that the imported code did.

### 2. A skipped relock overwrites module lockfiles with the parent script's

Pre-existing, and reachable today. On a skip, `try_skip_relock` returns the lock the *path* last
deployed; the per-module loop asks with the script's own `runnable_type`, so what comes back is the
parent script's lockfile and it is assigned to every module. Any multi-file script importing a
worker-locked script has already had this happen the first time its relock skipped. This PR widens
it to importers of supplied-lock scripts, which previously could never skip.

Narrower than it reads: `module.lock` has exactly one consumer, `handle_wac_v2_output`, where a WAC
v2 step naming `./<module>` dispatches it as its own job with that lock. An ordinary run installs
the parent script's lockfile, and `artifact_cache_name` hashes module *content*, not module locks.
So it bites a WAC v2 module dispatch whose module has dependencies outside the parent's closure —
where the parent statically imports the module, the parent's lock is a superset and it works.

The fix is to return the module's own last deployed lock on a skip rather than the path's:
`modules -> $key ->> 'lock'` off the same row the script arm already selects (`AND lock IS NOT NULL
ORDER BY created_at DESC LIMIT 1` — the clone's own inherited copy is *not* safe, a concurrent
relock may already have replaced it). That keeps the skip and repairs existing bad data. Making
modules simply never skip also fixes it, but gives up the optimization for every module-bearing
script, which is the opposite of what this line of work is for.

### 3. A fork under one snapshot

The clone runs at READ COMMITTED, so its steps read different source states. Deriving `lock_hash`
from the clone's own rows sidesteps that for this table, but `SET TRANSACTION ISOLATION LEVEL
REPEATABLE READ` on the fork transaction would settle it for every step, and would make
`dependency_map.imported_lockfile_hash` safe to copy — recovering the one relock round per importer
this PR gives up. It needs serialization-failure handling on a long multi-step transaction, so it
is its own change.


rubenfiszel:
Two things adjacent to this, from reproducing #10913 end to end on a local instance before finding this PR. Both are small; neither blocks it.

**1. Workspace clone/fork reproduces the same bug, and this PR does not cover it.**

`clone_workspace_runnable_dependencies` (`backend/windmill-api-workspaces/src/workspaces.rs:7143`) copies `dependency_map` without `imported_lockfile_hash`, and never copies `lock_hash` at all. Every fork and workspace clone therefore starts in exactly the state this PR fixes: no hash on either side of the comparison, for scripts whose locks were copied verbatim.

With the create-path fix here that is no longer "forever" (the first relock re-snapshots the map and converges), but it still costs one redundant relock round per importer after every fork. Copying both sides removes it:

```rust
    // `lock_hash` and `dependency_map.imported_lockfile_hash` are one comparison split over
    // two tables: a relock is skipped when an importer's recorded hash still matches the
    // imported path's current one. The clone copies each script's lock verbatim, so both
    // sides have to come along or every importer in the clone relocks forever.
    sqlx::query!(
        "INSERT INTO lock_hash (workspace_id, path, lockfile_hash)
         SELECT $1, path, lockfile_hash
         FROM lock_hash
         WHERE workspace_id = $2",
        target_workspace_id,
        source_workspace_id
    )
    .execute(&mut **tx)
    .await?;
```

plus adding `imported_lockfile_hash` to the `dependency_map` insert's column list and its `SELECT`.

Verified on a running instance, forking a workspace holding `u/admin/importer` importing `u/admin/imported`:

- before: the fork's `lock_hash` is empty and its `importer -> imported` edge is `NULL`, so redeploying `imported` with an unchanged lock still relocks the importer.
- after: both tables match the source, and the fork's importer logs `Skipping relock - imported lockfiles unchanged` on its first cascade.

**2. `.filter(|lock| !lock.is_empty())` in `create_script_internal` leaves a stale row behind.**

`lock` is `Some("")` when the script has a `codebase`, and when its language is not in the lock-gen list. A path that goes from a real lock to an empty one keeps its old `lock_hash`, and that is the one direction that can make an importer wrongly *skip* rather than merely relock too often. The worker writes `hash_script("")` in the same situation (`worker_lockfiles.rs:470`), so dropping the filter and keeping only the `None` arm (where a dependency job is about to write the row) matches it:

```rust
if let Some(lock) = lock.as_deref() {
```

---

Separately, confirming your note about the reported patch's convergence path being dead: `clone_runnable` goes through `clone_script`, which sets `lock: None` (`backend/windmill-common/src/scripts.rs:395`), so comparing the generated lock against the freshly cloned version could never match. I reviewed that branch earlier and missed it.

hugocasa:
Both taken, in 69f5d97 — thanks, and the second one is the sharper of the two: I had the filter backwards. Dropping it and keeping only the `None` arm is right, since that is the one case where a dependency job is about to write the row.

Verified both on a running instance.

**Clone/fork.** Forking a workspace holding the `imported → importer → importer2` chain from the rest of this PR's verification:

| | `lock_hash` rows | `dependency_map` edges with a hash |
|---|---|---|
| source | 4 | 3 / 7 |
| fork | 4 | 3 / 7 |

and the fork's `lock_hash` is byte-identical to the source's, path for path.

**Empty lock.** Deploying `f/test/emptylock` as bun with a supplied lock, then redeploying the same path as `postgresql` (no lock generation):

- after bun + lock: `8857522219566086108`
- after postgresql: `3476900567878811119`
- a fresh `postgresql` script at another path: `3476900567878811119`

so the row moves to `hash_script("")` and agrees with what the worker writes, where before it kept the stale real-lock hash — the direction that makes importers wrongly skip.

Also in that push: Codex found a race in the cancellation watcher I added for the relock wait. It polls every 5s, so a cancel committed in the same interval the advisory lock came free was dropped along with the watcher when the acquire branch won the `select!`. There is now one more check after the wait settles; racing a cancel against the lock release in the same instant still completes the job as `canceled`.

hugocasa:
Right, and the fix isn't derivable at clone time — dropped the `imported_lockfile_hash` copy in 32d82eb, kept the `lock_hash` one.

The two halves are not symmetric. `lock_hash` describes the imported path as the clone holds it, and it is the side an importer's own relock reads in order to record a hash at all — without it every dependency-map snapshot in the clone records NULL and nothing ever skips. `imported_lockfile_hash` records what an importer resolved against *when it was last locked*, and nothing visible at clone time distinguishes a hash consistent with the cloned importer from one a concurrent relock moved on. So the failure directions differ too: a hash older than the cloned scripts costs one relock, a newer one makes the importer skip a relock it needed. Only the second is unrecoverable, and only the second comes from copying that column.

Verified on the fork of the same chain:

| | `lock_hash` rows | dm edges with a hash |
|---|---|---|
| source | 6 | 3 / 7 |
| fork | 6 (identical to source) | 0 / 7 |

and it converges the way the comment claims — redeploying `f/test/imported` in the fork twice, the importer logs `Relocking script` the first time and `Skipping relock - imported lockfiles unchanged` the second, ending at 3/7 hashes like the source.

@rubenfiszel this keeps your point 1 for the half that fixes "relocks forever" and gives up the half that saved the one extra round per importer. Happy to look at recovering that separately if it's worth it — it needs the clone to derive the pair from what it copied rather than from the source tables, and `hash_script` being a Rust `DefaultHasher` means that can't be a `SELECT`.

hugocasa:
Dropped the relative-import relock serialization in 0f5ce24, along with everything that existed to support it — the shared advisory-lock module and its test, the native-trigger lock refactor done to share it, the ping heartbeat and the cancellation watcher.

It was redundant on every build anyone ships. The published images build with `features=ce` or `features=ee`, and `ce_core = ["oss_core", "private", "operator"]`, so `windmill-queue` already gives each dependency job `concurrency_key = dependency:{workspace}/{path}` at limit 1, enforced at pull. The only build without it is the `oss` feature set — which also sets `no_auth`, and where `private` being absent removes debouncing and concurrency limiting wholesale: `update_concurrency_counter` is a stub that always allows, `apply_concurrency_limit` is never called, `maybe_debounce` is never called. A bespoke lock on this one path was patching a hole in a wall with much larger ones, using a mechanism nothing else in the codebase uses.

It was also where the churn was: six of the eight findings across the four rounds were on that machinery, and each fix created the next.

The diff goes from 764 insertions across 15 files to 396 across 10. What remains is the reported bug and what it needs — the lock hash recorded on create, a converged relock not redeploying or cascading, module locks off the skip path, and `lock_hash` carried into a workspace clone. Re-verified end to end after the removal: the importer still skips and stops, `importer2` still gets no job, and the 28 tests across `relock_skip`, `dependency_map`, `script_modules`, `scripts` and `workspaces` are green.

If same-path overlap turns out to be real on a current fleet, it is the `MIN_VERSION_SUPPORTS_DEBOUNCING` rolling-upgrade window, and it belongs in that gate rather than beside it.

hugocasa:
On the concurrent-create finding: the mechanism is right, but it is a pre-existing property of `lock_hash` rather than something this PR introduces, so I have documented it instead of fixing it here.

`lock_hash` holds one row per path and **no writer anywhere checks that the version it describes is the live head** — including `worker_lockfiles.rs:474` on `main`, which writes `(workspace_id, script_path)` with the hash of whatever version its own dependency job locked. Two live versions therefore mean two dependency jobs finishing in an order unrelated to `created_at`, which is a wider window than two statements inside one create transaction.

The precondition is not narrow: the advisory lock at `scripts.rs:1196` only covers `auto_parent`, and the path-conflict check takes no row lock, so racing two `POST /scripts/create` at a fresh path leaves two live versions **5 times out of 5** locally. `created_at` is the transaction start time while the surviving `lock_hash` is decided by commit order, so the inversion you describe is reachable.

I am not fixing it in this PR because the fix is not in this table. While a path holds two live versions, every "the live version" read in the codebase is arbitrary — `scripts.rs:3327`, `:4198` and `try_skip_relock` all just take `ORDER BY created_at DESC LIMIT 1` — so head-guarding `lock_hash` alone treats a symptom of a broken invariant. Closing it properly means serializing same-path creates by extending that advisory lock to every create, which changes deploy concurrency for all scripts and deserves its own change and its own testing.

hugocasa:
Final pass at `7c1836f`, verified independently of the bot rounds.

**Scope.** Of the issue's three asks, this PR takes the first (the create-path `lock_hash` write, plus the no-op push repair and the fork derivation). The second is already covered on every published image: both `features=ce` and `features=ee` include `private`, which is what gates the per-path `dependency:{workspace}/{path}` concurrency key at limit 1. The third (a converged relock not redeploying or fanning out) is the larger half of the amplification and is deferred on purpose; a skipped relock still writes deployment metadata and cascades, so the job count of the incident only shrinks with that follow-up.

**Verified on a running instance built from this branch:**
- Create with a supplied lock writes `hash_script(lock)`; the same lock at two paths hashes identically; a `postgresql` script writes `hash_script("")`.
- No-op push repairs a deleted row with no new version, and after the last commit an unchanged push leaves the row's `xmin` untouched while a missing row is still repaired.
- An importer's edge records the imported hash instead of NULL. Four sequential redeploys of the imported script gave skip / relock / skip / relock exactly as the lock stayed the same or changed.
- The create path and the worker agree: pushing the importer back with its worker-generated lock produced the same hash the worker had written.
- Fork: one row per live locked script, edges with NULL hashes, and a source row deliberately set to a bogus value came out in the fork as the hash of the lock the fork actually holds.
- `test_create_script_persists_supplied_lock_hash` passes locally at `7c1836f`.

**On the concurrent-create finding:** the two-live-versions state comes from the create race itself, and the worker's own writer has the same last-write-wins property for it, so this PR extends an existing property to one more writer rather than adding a new one. CLI and git-sync pushes take the `auto_parent` advisory lock and cannot reach it. Serializing every create is the actual fix and belongs in its own change, as documented in the description.

One open question outside this PR: the reporter measured 4 to 7 concurrent same-path relocks on a fleet that should be running a `private` build. Worth confirming which image they run before treating the issue's second bullet as closed.

PR #10917, opened 2026-09-01
[ee] feat: detect and warn when a schedule skips occurrences

hugocasa:
## Summary

You set a script to run every 10 minutes. It quietly runs every 30, and Windmill never tells you.

This happens whenever a run takes longer than its own interval, or waits for a free worker. Windmill does not run schedules on a fixed grid: it keeps one queued occurrence at a time and computes the next one from the clock when the previous finishes. So a slow or delayed run pushes the whole schedule out, and the occurrences in between never happen. No log line, no failed job, no record of any kind.

Measured on a dev instance, a 25 s script on a 10 s cron:

```
ran at: 08:20:20 → 08:20:50 → 08:21:20 → 08:21:50
```

Two of every three runs silently gone. From the UI, that schedule looks perfectly healthy: every run it did make succeeded. Today the only way to find this is to notice stale data downstream and work backwards.

## What users get

**On the schedules list**, a warning next to schedules that are losing runs, alongside the existing paused and error indicators.

**In the schedule editor**, the numbers behind it, so the fix is obvious rather than a guess. The two numbers separate the two causes, which need opposite fixes: a large wait means not enough workers, a long run means the job outgrew its interval. There is deliberately no one-word label, because both usually contribute and any dominance rule would mislabel the mixed case.

**For operators (EE)**, an alert when a scheduled run is still going and has already blown past its own next occurrence, so a wedged schedule pages someone instead of waiting to be noticed. No thresholds to tune and no new instance setting: the check compares each schedule against its own configured cadence, so it is equally meaningful for a daily and a per-minute schedule, and delivery rides the existing critical alert channel.

## Screenshots

Both signals on the schedules row (the schedule was mid-overrun when this was taken):

late-and-skipped

The editor detail, showing the real 25s-on-10s case recovered from job rows. The `1000+` row is the gap across a disable/re-enable with the baseline cleared, which is what `occurrence_baseline_at` exists to suppress:

editor-occurrences

## Which cause is caught, and when

| | overrun (ran longer than the interval) | late start (waited for a worker) |
|---|---|---|
| plain script schedule | caught | caught |
| script with `retry` or `dynamic_skip` | nothing to catch | caught |
| flow schedule | nothing to catch | caught |

Where occurrences are **serialized**, the successor is pushed when the current run finishes, so the gap it opens is governed by wait and duration together. Where occurrences **overlap**, the successor is pushed when the current run *starts*, so only the wait can move the grid forward; a long run loses nothing there, which is the design intent.

Counting is retrospective: a gap becomes visible once the *next* occurrence exists. A schedule wedged right now shows nothing until it moves, which is what the live indicator and the operator alert cover.

## What this is not

Not a change to how schedules execute. Nothing starts overlapping, nothing gets queued up, no missed run is retroactively fired. This is diagnosis only. Making the overrun behaviour configurable (skip / overlap / queue) is a much larger change to the execution model, and it needs this diagnosis first to know which way people actually want it to go.

---

## How it works

`push_scheduled_job` is not touched, so nothing in the per-run hot path changes.

The counting is free because the history is already in the database. `v2_job.created_at` of occurrence N+1 is bit-identical to `completed_at` of occurrence N (the push happens inside the completing transaction, and both timestamps resolve to the same transaction start time), so `scheduled_for = find_next(created_at)` holds **exactly**. The whole occurrence sequence is reconstructible from rows that already exist and are already indexed.

The only write is `occurrence_baseline_at` on `schedule`, advanced at create / edit / `set_enabled(true)` / `rearm_schedule`. Those are all cold paths, so it costs nothing per run. It excludes gaps that span a pause, a cron change, a re-enable or a reconciler re-arm, with one mechanism. It is nullable with no backfill: NULL reads the whole history, which beats blinding a daily schedule for 20 days.

### Why this is affordable on a list page

Measured on croner 2.2.0, release build: `find_next` costs 165–330 ns for ordinary expressions (3.3 µs for the worst tried, `0 0 3 29 2 *`), and walking 1000 consecutive occurrences 150–650 µs.

Deciding *whether* a run skipped anything is one `find_next` per gap: across a full page of 1000 schedules × 20 occurrences that is ~39,000 calls, about 10 ms, hard bounded and independent of how badly the schedules are behaving. Counting *how many* needs a walk, which does not fit that budget, so the list answers the cheap question and the editor answers the expensive one for one schedule at a time. No step budget, no degradation heuristic.

### Correctness hazards handled

- **`status = 'skipped'` occurrences must not be filtered out**, or holes in the sequence manufacture phantom skips. `list_schedule_with_jobs` carries `AND status <> 'skipped'`; against a `dynamic_skip` schedule, 4 actual root occurrences were visible as **0** through that filter. Per `CONTEXT.md` a polling flow's steady state is also `skipped`, so this is the common case. The reconstruction reads an unfiltered array from a second LATERAL, leaving the displayed bar chart byte-identical.
- **Order by `created_at`, not `completed_at`** — flows overlap, so completion order is not creation order.
- **The overrun signal is not simply "scripts".** A script schedule carrying `retry` or `dynamic_skip` is pushed as a `SingleStepFlow`, which `is_flow()` reports as a flow, so it re-arms at step 0 entry and its occurrences overlap. The scope is `NOT is_flow AND retry IS NULL AND dynamic_skip IS NULL`.
- **Flow schedules hold two or more concurrent root rows in `v2_job_queue`**, so the live query aggregates per `(trigger, runnable_path)` rather than assuming one row.
- **`no_flow_overlap` and `dynamic_skip` do not manufacture phantom skips**: both leave a completed row that reconstruction reads like any other.
- **sqlx types LEFT JOIN columns as NOT NULL** off the column definition, so the occurrences query needs explicit `?` overrides or the still-queued occurrence decodes wrong.

### Rejected alternatives

- **Counting inside `push_scheduled_job`.** Exact and nearly free, but a failure there fails job completion (`jobs.rs:1411` returns the error), leaving the schedule unarmed until the reconciler catches it.
- **An events table.** Buys frequency and click-through, costs an `INSERT` in the push transaction plus retention and RLS. Reconstruction gives the same data for free.

## Changes

- `occurrence_baseline_at` migration on `schedule`, nullable, no backfill
- `reconstruct_occurrences` and `is_overdue` in `windmill-common`, both pure, with unit tests
- `list_schedule_with_jobs` gains `skipped_runs`, `runs_examined`, `running_late`
- new `GET /w/{workspace}/schedules/occurrences/{path}` for the per-run detail
- baseline written at create / edit / `set_enabled(true)` / `rearm_schedule`
- schedules-row indicator and a "Recent occurrences" section in the schedule editor
- `schedule_overrun_alerts` wired into the monitor pass (EE implementation in windmill-labs/windmill-ee-private#772)
- design doc at `docs/schedule-missed-occurrences.md`

## Test plan

Unit tests (6, in `windmill-common`), all passing:

- [x] the measured 25s-on-10s sequence reconstructs to `skipped 2` per gap
- [x] a hole in the input manufactures a phantom skip (pins why the query must not filter `skipped`)
- [x] a gap older than the baseline is not comparable
- [x] a long gap saturates at the walk cap rather than walking forever
- [x] detect-only reports the gap without counting it
- [x] an occurrence is overdue exactly once it outlives its successor

Exercised on a running instance, not just type-checked:

- [x] the occurrences endpoint recovered yesterday's real job rows exactly: `08:20:20 → 08:20:50 → 08:21:20 → 08:21:50`, wait 3–39 ms, duration ~25,030 ms, `skipped_after: 2` each
- [x] enabling a schedule wrote `occurrence_baseline_at` and pushed the occurrence in the same transaction (identical timestamps), and the list then reported `0 examined` — the re-enable gap correctly excluded
- [x] `running_late` observed flipping true ~10 s into each 25 s run and back to false the moment it completed, with `skipped_runs` incrementing as the gap became measurable
- [x] badge, popover and editor section verified in the browser (screenshots above)
- [x] EE alert exercised against a binary built with `--features enterprise,private`: alert fired, no duplicate on the following pass, recovered on the pass after the condition cleared, `healthchecks` row flipping `false → true`
- [x] `cargo check` clean for CE, `enterprise`, and `enterprise,private`; `SQLX_OFFLINE=true` clean for CE and `enterprise,private`; `npm run check` 0 errors


hugocasa:
Re the Claude review nit (edit erases `skipped_at`, now `last_missed_at`): intended. Editing or toggling a schedule clears the whole state, including the 7-day muted badge, because the state describes the previous configuration and the edit is the user acknowledging it. The other findings are addressed in bddd6def41: `fetch_optional`, the alert waits for the push to commit (`FOR SHARE`), a failed `find_next` leaves the streak untouched, and an integration test covers the streak and its reset.

hugocasa:
Re Codex P1 (a rolled-back alert task can validate a retry's committed streak): accepted as a known residual. It needs the first attempt's spawned task to not even reach the database before the 3s retry backoff elapses, i.e. pool starvation, and its cost is one duplicate critical alert. The realistic case, a push that keeps failing across the flow path's retries, is what `FOR SHARE` covers: verified with the `Push` failpoint on the third late run while the outer transaction commits, the streak stays at 2 and no alert is raised, the committed retry raises exactly one, and the next on-time run recovers (and acknowledges) it. Tying the alert to the spawning transaction id would close the residual, but not for its cost here.

Also in 078ec0aa29: the recovery alert with the `schedule:{path}` resource (Claude P2), `last_missed_at` now stores the last missed slot (Codex P2 / Claude nit), and the counting helpers are renamed to *missed* (Claude nit).

hugocasa:
Round 4 (on 2b502d6), addressed in 28ca33dbcd by simplifying rather than patching: the late run alert now follows the sibling critical alerts (trigger disable/failure, job error handler failure): one `report_critical_error(msg, db, Some(w_id), None)` per streak, acknowledged from the alerts feed. That removes the recovery, the resource key and `acknowledge_late_run_alert`, so Claude's P2 (`AND NOT acknowledged` on cloud), Codex's P1 (the public mutator's authorization contract) and Codex's acknowledgement race no longer apply. The capped count and timestamp are documented as approximate in openapi (Codex P2). The `FOR SHARE` guard stays for the flow path's retries, re-verified with the `Push` failpoint.

hugocasa:
✅ Review round clean @ 28ca33dbcd4861e1920ac8e1742577ee12d7dd06

PR #10919, opened 2026-09-01
feat(ai-sessions): replace the context panel with an assistant settings modal

rubenfiszel:
## Summary

A session chat's capabilities come from several places the user controls independently — skills they turned on, MCP servers they connected, custom instructions set in two places, files and folders they linked, and a tool set that changes with the mode. Each outlives the moment it was set, and none of it was visible together: the `+` menu toggles skills and MCP but never shows what is *currently* in effect, and nothing surfaced the tools or the instructions at all.

This replaces the read-only context popover with an **Assistant settings modal** — one place that both accounts for what the chat carries and is where you change it. The skills drawer, the MCP drawer and the model dropdown's Parameters flyout are gone; their contents are sections of the modal.

EE companion: windmill-labs/windmill-ee-private#765

## The modal

A sliders trigger sits in the composer next to the model pill. Five sections, each a `SidebarNavigation` row with a count:

- **Tools** — `manager.tools` + `manager.planMode.tools`, i.e. exactly what the chat loop sends. Fuzzy search via the shared `SearchItems` (uFuzzy), highlighting matches in the name *and* the description — two passes, since `marked` covers one haystack. A row opens a detail page with the full description and the tool's argument schema through `SchemaViewer`. Read-only: tools follow the mode and the connected servers.
- **Skills** — every `ai_skill` resource, with the per-user switch, and an editor for the SKILL.md behind the row.
- **Instructions** — the workspace and user prompt blocks as two tabs, edited in place, one Save writing whichever changed.
- **MCP connections** — each connection with its switch, a **Manage connection** detail page backed by the shared `ResourceEditor`, and a connect form. The destructive action is now named **Delete**, because it deletes the resource; the switch is the reversible one.
- **Files & folders** — linked folders and files, each with why it is not readable when it is not (`needs access`, `indexing…`, `failed`), and an unlink. Restores the row dropped when the old panel went, using @rubenfiszel's `countReadyAttachments` from #10919's base — a folder counts by its readable children, never its aggregate status.

Skills, MCP and Tools are list/detail pairs built on `PagedContent`, so Left/Right steps between the pages and Escape backs out one level rather than closing the modal.

## Shared components (review these on their own)

- **`common/listRow/ListRow.svelte`** (new) — the row shape the resource-type picker already used, extracted: icon, label, one truncated line, optional trailing controls. Renders a `Button` when nothing trails it and a `div` + inner label button when something does, since a button cannot nest in a button. Now used by the picker and all three chat lists.
- **`common/listRow/listHighlight.svelte.ts`** (new) — the picker's keyboard highlight, extracted: arrow keys move a lit row while focus stays in the search field, Enter opens it. Keeps the parts that are easy to get wrong — resting index on filter change, and refusing to let a scroll under a resting pointer drag the highlight back.
- **`AppConnectInner.svelte`** — switched to both; loses its row markup, its local `titleIfTruncated` and its whole highlight state machine (−136 lines).
- **`SidebarNavigation.svelte`** — gains `{@attach arrowTabNav()}`, so instance settings, workspace settings and the setup wizard get Up/Down navigation too.
- **`McpConnect.svelte`** — the path field now says what the path decides: under `u/<you>` the connection is yours alone; in a folder, everyone in it can use the connection **and its token**, which lives in a secret variable at the same path. Verified against the RLS policies and a second user.
- **`Section`**, **`SimpleEditor`**, **`Modal2`** — secondary description colour; line-number gutter matched to `Editor.svelte` plus a model-swap ordering fix; an additive `closeOnEscape` so a nested surface can win.

## Telemetry

`logFeatureUsage('ai_session', 'context_panel', { key, workspace })` over a closed key set. The `(feature, kind)` pair is registered in the EE companion, so `backend/ee-repo-ref.txt` is re-pinned.

## Screenshots

Tools — every definition the chat sends, fuzzy-searchable:

tools

A tool opened: the description in full, and the arguments it takes:

tool detail

Skills, each with the switch that decides whether this chat carries it:

skills

Instructions — the workspace and user blocks, edited in place:

instructions

MCP connections:

mcp connections

Files & folders, with why one is not readable when it is not:

files and folders


rubenfiszel:
Round 1 addressed in 354a5f638b. Taking the findings in turn.

**Fixed**

- **Stale user instructions in the glance (codex P1 / claude P2).** The `instructions` derived now also depends on `aiChatManager.systemMessage`, which the manager reassigns whenever it re-reads the prompts — the path `update_user_instructions` takes. That is the only signal the localStorage-backed user half has, so the trigger tooltip no longer lags the popover until it is opened.
- **Untested `summarizeTools` (codex P2).** Added a test pinning the sorted order and the missing-description fallback. The sort is the part with a real failure mode: the panel re-derives this every turn, so losing it would reshuffle the list under the reader.
- **Duplicate overlong rationale (codex P2).** The "why not `changeMode`" constraint now lives once, at the `sessionRuntime` call site where someone would break it; the manager-side comment is one line saying the method is local.
- **PR description drift (claude P2).** Corrected — it described the `changeMode(GLOBAL)` routing that review round 0 rejected, and omitted the `workspace` option.

**Not changing, with reasons**

- **Workspace-scoped instructions during a session switch (codex P1).** `getCustomPromptParts(AIMode.GLOBAL)` is the same call `configureGlobalMode` / `rebuildGlobalSystemMessage` use to build the system prompt, with no workspace argument in either. So the panel cannot disagree with what the prompt carries — which is its whole contract — and gating it on `operatingWorkspace` would make it disagree in the other direction, reporting instructions the assistant is not actually being given. The transient the finding describes self-corrects: `copilotInfo` is a dependency of the derived, so the panel updates the moment `loadCopilot()` resolves. If the prompt builder should be workspace-scoping its custom prompts, that is a real question about `getCustomPromptParts`, and a behaviour change to the assistant rather than to this panel.
- **Telemetry disclosure (codex P2).** The disclosure reads "counts of which product features are used, including AI provider and model identifiers, the names of public hub scripts used, ...". The enumerated items are the counters that carry something beyond a plain count; this one's keys are a closed set of five section names with no identifier in them, so the leading clause already covers it and the copy stays accurate. The claude reviewer reached the same conclusion independently on this head. Happy to add a clause if you would rather the list be exhaustive.

rubenfiszel:
Round 2 addressed in bf2ee5a8cb — all three fixed, no dismissals.

- **Context refresh / send race (P1).** Correct and the sharpest finding so far: the refreshes carry generation counters, so opening the panel during a send invalidated the send's own refresh — its `Promise.all` returned without applying results, and the panel's then rewrote tools and the system prompt under a running turn. `onOpenChange` now refreshes only when `!loading && !sendInFlight`, the same gate used at `AIChatManager.svelte.ts:1867` and `:3975`. Mid-turn the panel shows the values the turn was actually given, which is the honest reading anyway — and a panel that reports should not be able to change what it reports on.
- **Stale hover glance after a user-prompt save (P2).** Fixed at the source rather than in the panel: `AIChatModelSettings.save()` now calls `rebuildGlobalSystemMessage()` after writing user prompts, the same thing `update_user_instructions` does for the same block. The manager's own system message was equally stale until the next send, so this is a little wider than the tooltip.
- **Empty folders reported as usable (P2).** A folder now needs a readable child, not just a `ready` aggregate — an empty or all-binary one keeps a placeholder that `readyFiles()` filters out, so it counted as an attachment while exposing nothing.

`npm run check` is clean and the suites are unchanged (the 6 failures in `ContextManager.test.ts` / `app/core.test.ts` reproduce on the base commit `7c04aabaf9`).

Worth flagging for the next round: two of these three followed from earlier rounds' fixes — the race from adding refresh-on-open in round 0, the tooltip gap from the freshness dependency in round 1. The recurring shape is that this state is materialized lazily and refreshed from several call sites, so a read-only observer of it is easy to get subtly wrong. The fixes have stayed small and local, and claude and pi both returned "Good to merge" on the previous head, so I don't think it warrants reshaping — but if round 3 turns up another finding of the same family rather than a new one, that judgement should be revisited rather than patched again.

rubenfiszel:
Round 3 nits fixed in 683f99d98b. Both correct, and the first was mine to own.

- **Mixed-status folders reported unavailable.** Round 2's fix for "empty folders counted as usable" over-corrected: requiring the aggregate `status === 'ready'` excluded a folder with one ready child and one indexing/failed one, which the assistant can already partly read. Now counted the way `readyFiles()` sees it — `files.some(f => f.status === 'ready')` — which covers the empty/all-binary case and the mixed case in one predicate.
- **Overlong refresh invariant.** Compressed to four lines, keeping the generation-counter rationale.

Since that predicate has now been wrong in both directions, I pulled it out as `countReadyAttachments` in `agentContext.ts` with tests for the three folder shapes (partly readable → counted, empty → not, all-failed → not) and for standalone files. That is the third bug in this one rule, so it earns a guard rather than another comment.

`npm run check` clean, 9 tests in `agentContext.test.ts`, suites otherwise unchanged.

On the convergence question I raised last round: this round was non-blocking from every reviewer that posted (codex nits-only, pi good to merge), and the two nits were a straightforward over-correction plus a comment length — not a new family of finding. So I don't think the seam needs reshaping.

Guilhem-lm:
Round 4 addressed in `aa1ef58643`. One fixed, one dismissed with evidence.

**Fixed — P2, the refused `setMcpEnabled` write.** Correct and the sharper of the two. `preference.setEnabled` returns `false` when there is no account key or the write throws, and both MCP toggles painted `enabled` regardless, so the chat carried a different set than the switch showed with nothing to say so. Both call sites now check and toast, matching `AssistantSkillsSection.toggle` three lines away and `SkillsMenu.#toggle`.

**Not changing — P1, the prompt entries in SCRIPT/FLOW/APP chats.** Those chats do not exist. `AIChatDisplay` has exactly one mount, and it is a session:

```
AIChatDisplay ← AIChat.svelte:152 ← SessionWrapper.svelte:504 ← routes/(root)/(logged)/sessions/+page.svelte:895
```

`<AIChat` appears once in the tree, and `sessionRuntime.svelte.ts:335` settles the mode:

> Sessions always operate in GLOBAL mode (workspace-item tools across the session's workspace). The page-level gate already requires the global-AI flag, so this is always available here. Mode is locked and the dropdown is hidden in the chat UI.

The `AIMode.APP` branch cited as proof is evidence of the opposite: it is left from when the chat was embedded in the editors. `RawAppEditor.svelte:873` still calls `changeMode(AIMode.APP)`, but on the singleton manager with no `AIChatDisplay` rendered against it. The editors' AI buttons now navigate to a session — driving one in a browser lands on `/sessions?session_name=…`, not an in-editor panel.

So `promptSettings={aiChatManager.mode !== AIMode.GLOBAL}` would be an expression that is always `false`, and the entries it would restore have no composer to appear in. The reachable version of this finding was real and is already fixed: `d0f81ed5cc` restored the submenu for the logged-in home composer (`home/HomeAIChat.svelte:212`), which mounts `AIChatModelSettings` standalone with no settings modal — that is the second mount point, and the one that mattered.

This is the third round to reach this finding from the same in-file evidence, so rather than re-argue it I have put the mount chain in a comment at the call site (`AIChatDisplay.svelte`), so the next reader does not have to re-derive it.

**Note taken — the `context_panel` telemetry kind.** Fair: the name now contradicts the surface. Leaving it, with the reason recorded next to `record()` — the pair is registered EE-side under that name and events already exist against it, so a rename spans `feature_usage_ee.rs`, `ee-repo-ref.txt` and the recorded history for a label.

Also ran `prettier` over the files flagged as unformatted. `npm run check` clean, 15 tests pass.

Guilhem-lm:
Pi and Codex round addressed in `bd8ea58c06`. All three fixed — the P1 both reviewers found independently was mine, and it is the interesting one.

**P1 — workspace instructions written into a staged fork's parent.** Correct, and it is the failure the guard exists for. I added `blockedByPendingFork` to Skills, then extended it to MCP a round later, and never asked which *other* sections write workspace-scoped data. Instructions was the fourth writer: `editCopilotConfig({ workspace: ws })` with `ws` resolving to the parent before the first send, so an admin saving here edited the live parent's shared AI config — and, as Codex notes, every chat still in that parent would then be given it.

The workspace half is now read-only while a fork is staged, saying which workspace it would have written to, and `saveWorkspace` refuses with a toast rather than relying on the field being disabled. Exercised against a real staged fork rather than reasoned about:

```
Acting on guardtest (new)
alert:      "…so these would be saved to "ai-meta-e2e" and given to every chat already in it."
field:      readOnly
Save:       disabled
ai-meta-e2e custom_prompts, before and after: None
```

**P1 — editability read the navigation workspace's role.** Sharp, and I had not considered it. `$userStore.is_admin` belongs to `$userStore.workspace_id`, but this section acts on `ws`, which in a session is the operating (possibly forked) workspace. So an admin of the target but not of the nav workspace saw a read-only field they could in fact edit, and the reverse case offered an editor whose save would be refused. The role is now resolved for the target with `getUserExt(ws)` alongside the settings read, with `is_super_admin` still short-circuiting. Both awaits in `saveWorkspace` are pinned to one `target`.

**P2 — comments too long and describing drafting history.** Fair, and squarely against the root `AGENTS.md` rule I am supposed to be following. Both trimmed to the durable constraint: the `AIChatDisplay` one now states the mount and the mode lock in three lines without narrating what the file used to be, and the telemetry note says only that the kind is the pair registered in `feature_usage_ee.rs` and that an unregistered pair is dropped silently.

**Unrelated red check.** `cargo_test` is failing on `main` as well, at `38fc0d3a` (11:18 today), same test and same assertion — `test_mcp_preprocessor_receives_the_callers_headers`, `Invalid requirements, expected to find //bun.lock split pattern in reqs. Found: ||`. Not from this branch, whose only backend change is the `ee-repo-ref.txt` re-pin.

`npm run check` clean, 15 tests pass.

Guilhem-lm:
All three addressed in f35a8661a6.

**1. [P1] Stale target-workspace role** — the resolved membership is now keyed to the workspace it was read for (`targetRole = { workspace, user }`), and `$userStore` is only consulted when its `workspace_id` matches the target. Until the target's role resolves, both fields stay read-only with `Checking your access to <workspace>.` as the stated reason, so a non-admin never sees an editable workspace field that the write would reject.

**2. [P1] Save discarding edits typed mid-flight** — the field is `readonly` while `saving`, so the two-request round trip has no window to swallow keystrokes. Verified by stalling `get_settings` for 2.5s: the textarea reports `readOnly: true` for the whole flight, returns to editable after, and the persisted value is the submitted one. `load()` keeps the other approach on purpose — it starts on its own rather than from a click, so it preserves a newer draft instead of blocking someone who opened the tab to type.

**3. [P2] AI-settings shortcut opening the navigation workspace** — the link now carries the target: `/workspace_settings?workspace=<ws>&tab=ai`.

`svelte-check`: 0 errors. `agentContext.test.ts`: 5 passed.

Unrelated to this branch: `AIChatManager.test.ts` is red here and on `main` — #10941's `workspaceDisabled` guard reads `copilotInfo`, which the test's `$lib/aiStore` mock does not export. The test file is byte-identical to main's and this branch touches neither symbol. `cargo_test` is likewise red on main at `38fc0d3a` for a bun-lockfile parse error.

Guilhem-lm:
All three nits addressed in ca56d98fa4.

**1. [P2] Deferred reload after a dirty target switch** — the effect now tracks `dirty` (as `clean`) and keys off `loadedWorkspace`, so a reload skipped to protect a draft happens as soon as that draft is saved or reverted, against whatever `ws` is by then. Parking the section clears `loadedWorkspace`, which keeps the reload-on-each-visit behaviour.

Exercised in the browser with `fetch` instrumented: typed into User instructions, left the section and came back — **zero** requests, draft intact; cleared the draft — `users/whoami` + `workspaces/get_settings` fired immediately. Before this commit that second pair never came.

I did not invalidate the in-flight load on the skip path: the same skip runs on the ordinary "first load still in flight when the user starts typing" case, and cancelling there would leave the panel with no workspace prompt and nothing to retrigger it. A load resolving for the old workspace is already harmless — the draft guard keeps the text, and `targetRole` is workspace-keyed, so it reads as unresolved rather than as the wrong role until the retry lands.

**2. [P2] Raw button in `ListRow`** — the trailing-control branch now renders `Button` (`variant="subtle"`, padding and hover zeroed, since both belong to the row around it), and `triggerableByAI` comes from `Button`'s own `aiId`/`aiDescription`. Verified against an MCP row: label control 415px wide at x=12 inside a 536×56 row, `padding: 0px`; the ellipsis still opens its menu without navigating, and the label still opens the detail page.

**3. [P2] Overlong role invariant** — down to four lines, keeping the nav-workspace-vs-`ws` and unresolved-role constraints.

**On `listHighlight` coverage**: no vitest project in this repo runs rune-bearing test files. `server` includes `src/**/*.{test,spec}.{js,ts}` but excludes `*.svelte.{test,spec}`, and `dom` only takes `*.dom.test.ts`, so `useListHighlight` (which needs `$effect.root` and therefore a `.svelte.ts` file) cannot be tested without adding a project. `npx vitest run src/lib/components/flows/flowDiffManager.svelte.test.ts` returns "No test files found" — that pre-existing rune test never runs either. Adding that infrastructure is a separate change from this PR; happy to file it.

`svelte-check`: 0 errors. `agentContext.test.ts`: 5 passed.

Guilhem-lm:
Four findings from this round (Codex, Pi, cubic ×2) addressed in e3ed338c4b.

**[P2, Codex] Page navigation stealing ArrowLeft from the skill View/Edit toggle** — real, and the root cause is `PagedContent`, not the skills panel: its window handler ignored `event.defaultPrevented`, so any nested control that answers the arrows got overridden. It now returns early on `defaultPrevented`, which covers toggle groups, menus and sliders alike rather than special-casing one selector.

Measured in the browser: dispatching ArrowRight on the `View` radio reports `defaultPrevented: true` — so pre-fix the page navigated too — and after the fix focus moves `View → Edit` with the editor page still live (`inert` unchanged). ArrowLeft from `Edit` moves back to `View`, and the page arrows still work: from the editor's back button, ArrowLeft returns to the list (live page flips `editor → list`).

**[P2, cubic] `roleResolved` never true when the role lookup fails** — correct. `getUserExt(...).catch(() => undefined)` stores a record with no user, which read as "still checking" forever. Split into three states instead of collapsing two: `roleResolved` (the read for this workspace came back), `roleUnknown` (it came back empty and the nav store is not the authority), and `isAdmin`. A failed read is still read-only, but says `Could not read your access to <ws>. Reopen this section to try again.` rather than claiming a role it could not read.

I did not take the suggested one-liner as-is: `targetRole?.workspace === ws` alone would report a failed read as a definite non-admin ("Only workspace admins can edit"), which is equally wrong for an admin whose lookup failed transiently.

Verified with `users/whoami` stubbed to 500: with the nav workspace equal to the target, the field stays editable through the nav-store fallback and no blocker appears — a failed read does not lock out an admin. The cross-workspace sentence itself I could not exercise locally: it needs a session acting on a *committed* fork workspace, and the fork here has no committed session. Flagging that rather than implying I saw it.

**[P2, cubic] `type="button"`** — right, `Button` spreads `rest` onto the element and sets no native type. Added on both `ListRow` branches, not just the new one.

**[P2, Pi] The deferred-reload comment was 5 lines** — trimmed to 4.

`svelte-check`: 0 errors. `agentContext.test.ts`: 5 passed.

Guilhem-lm:
Correct, and it was a regression from the previous round's `defaultPrevented` guard — that guard turned a harmless redundancy into dead arrow keys. Fixed in 649447e8bb.

Reproduced first, on `097d5e212b`: open a skill, ArrowLeft back to the list, then ArrowRight — the event came back `defaultPrevented: true` and the detail page did **not** reopen. The Tools instance mounts first, `listening()` only asks whether the overlay host is active and on top (nothing about which section is on screen), so it consumed the key and its own `navigate()` then discarded it as inactive.

Took the first of your two suggestions, since it removes the class of bug rather than one instance: a parked section no longer hands `PagedContent` a navigator at all — `onNavigate={active ? navigate : undefined}` at the three call sites, so the hidden instance returns on its existing `!onNavigate` line before any `preventDefault`. The now-dead `if (!active) return` inside each `navigate()` is gone, leaving one statement of the rule instead of two, and `PagedContent`'s `onNavigate` doc now says a host that stays mounted while hidden must withhold it while hidden.

After the fix, same script: Skills opens, ArrowLeft returns to the list, ArrowRight reopens the detail, and the hidden Tools section is still on its own list. Tools' own Left/Right still work, so the first-mounted and later-mounted ends both behave.

`svelte-check`: 0 errors. `agentContext.test.ts`: 5 passed. Prettier clean.

PR #10929, opened 2026-09-01
feat: guest app execution mode, a role that takes no seat

rubenfiszel:
## Summary

Adds a fourth app `execution_mode`, `guest`: **login required, workspace membership not**. Someone the identity provider authenticates can open an app opted into it, and the app's runnables execute as the publisher exactly as in `publisher` mode.

The point is licensing. A guest gets **no `password` row and no `usr` row**, so the row-based seat counters never see one. What they cost instead is a single rule, `guest_seats`: the first **100 distinct guest emails over a trailing 30 days are free** on every plan; past that an Enterprise plan meters them at **four guests to one seat**, and every other plan and build (Community, Pro) stops admitting new emails until the count drops (a returning guest is always let back in). The companion EE PR adds the before/after seat figures so the pricing change can be decided on measured numbers rather than a guess.

Two explicit gates, both off by default:
- **Per app**: the `Public URL` toggle becomes an `Access` button group: **Members / Guests / Public**. Widening to Guests is gated by a new `RestrictGuestAppDeployment` protection rule, sharing one helper with the existing anonymous rule.
- **Per plan**: the allowance, not a refusal. `guest_admission` (in `windmill_common::workspaces`, next to `guest_app_admits`) is checked once where a session is minted, under a transaction-scoped advisory lock so concurrent strangers count each other and the cap is exact: a metered plan admits everyone; a capped one refuses a stranger once the window holds 100 emails, with the sentence the visitor reads ("This instance has reached its limit of 100 guests over 30 days. Guest sign-in beyond that needs an Enterprise license."). `guests_are_metered` is the only place the plan is read (Enterprise on an `enterprise` build; every other build is capped, since the CE image compiles the OAuth callback that mints). The switch, discovery and the auth door do not consult the plan: an admitted guest stays admitted for its session, and the cap only ever decides about the next stranger.
- **Per instance**: a superadmin switch, `guest_access_disabled` in the global settings (a toggle at the top of the new Guests list), which every workspace switch sits under. Read from the table by the same gates as the workspace switch (the mint, discovery, the auth door on every guest request), so turning it off stops every guest on the instance on their next request and hides every app from discovery, whatever a workspace or an app says. `REQUIRE_PREEXISTING_USER_FOR_OAUTH` stays what it is, a gate on provisioning an account, which a guest never does.
- **Per workspace**: `workspace_settings.guest_access_enabled`, checked where a guest session is *minted* (`guest_app_admits`, the single source of truth for "both gates") and then **once, at the auth door, for every request a guest makes** (`AuthCache::get_opt_job_authed`), not per handler, where each guest-reachable route would have to remember it. Deliberately not only at deploy time: an app carries its `execution_mode` in its definition, so git-sync and the CLI push `guest` past every UI. Turning it off stops every guest session and every token derived from one on their next request.

## Who is a guest, and what gets counted

The rule is decided once, at sign-in, from rows that already exist. There is no promotion step, no promotion endpoint and no promotion counter.

- **The identity has an account** (a `password` row, or a `usr` row in *any* workspace): it gets an ordinary session token, exactly as before this PR. Nothing about guests applies to it. If it is not a member of the app's workspace, the app refuses it ("You are signed in, but this app is not open to you"): an account holder is never turned into a guest, and never gets a second, cheaper identity. The rule holds after the mint too: the auth door re-reads "no account" together with the workspace switch on every guest request (`guest_session_stands`), so an account provisioned later, or in a race with the mint, ends the guest session on its next request.
- **The identity has no account at all**: it gets a `guest_session` token pinned to the workspace and scoped to that one app, and **no rows are written**: no `password`, no `usr`, no invite. Every seat counter (telemetry, offline-license cap, fork billing, the superadmin user list) counts rows, so a guest is invisible to all of them without a single counting rule changing.

"Promotion" is only the second case later doing an ordinary sign-in: `/api/oauth/login/<client>` takes a `guest_app` parameter, and a sign-in *without* it drops the same identity into the provisioning that was already there, which writes the `password` row, and from then on the existing counters see it like anyone else. One code path, so `REQUIRE_PREEXISTING_USER_FOR_OAUTH` and `auto_invite` cannot drift between "first sign-in" and "promotion". A sign-in that *did* carry a guest target the app no longer admits (the switch or the mode flipped during the provider round-trip) fails with a 403 rather than falling through: the page promised no account, and provisioning one there would take a seat nobody asked for.

The only new counting is `guest_activity` (email × workspace × day, 60-day retention). It is the set the allowance is counted on, instance-wide (a licence is per instance, and one email is one guest however many workspaces it opens), and it feeds three places through one `GuestUsage`: the **offline licence seat check** (`ActiveSeatCounts.guest_seat_count`, so on an Enterprise plan guests past 100 count against the key exactly as operators do, four to a seat, and can push an instance over it), the **telemetry** (`seats_after` now includes those seats; `seats_before = accounts + 0.5 × distinct guests`, what the old policy would have charged had those people been operators), and the **UI**: a superadmin "Guests" sub-tab under Users listing the window's distinct emails with their workspaces and first/last seen, and a usage line ("N of 100 free guests used in the last 30 days", then the meter or the cap) on the workspace Guests card and under the editor's Guests rung. The per-workspace `billable_seats` the invoicing job mirrors is deliberately left without guests: they are not attributable to one workspace.

The word *account* in the first bullet means "anywhere", which is stricter than "member of this workspace": someone with an account in workspace A opening a guest app in workspace B is refused as a non-member, not admitted as a guest of B. The licensing case is people with no account, and letting existing accounts ride as guests elsewhere would be the loosening; the mint's eligibility check (`has_any_account` in `login_externally` and `create_guest_session_token`) is the one line that encodes it, so the other reading is a deliberate one-line change, not a redesign.

## Changes

**CLI and git-sync**
- The access mode is the one policy field a tracked `app.yaml` keeps, and it is now a tri-state: `public: true` (anonymous), `guests: true` (guest), neither (publisher), carried by `markAccessFromPolicy` on pull and `executionModeFromAppFile` on push, for low-code and raw apps and the sync bundle alike. A pull then a push deploys the mode that was pulled; `cli/test/app_access_mode_unit.test.ts` pins the round trip. (Before, push regenerated `execution_mode` from a boolean, so `guest` came back as `publisher` with no diff to show for it.)

**Authorization**
- `ExecutionMode::Guest`, and a `guest` sentinel scope confining a session to the app surface (`guest_route_denied`, reusing the reviewed `app_embed` allowlist plus the embed-token mint the embedder page needs).
- A new `AuthCache` arm resolving a token with no `usr` row into an identity, keyed on a **server-minted label** (`GUEST_SESSION_LABEL`), never on a scope, which is caller-supplied. The label is reserved in both `is_server_minted_label` (so `/users/tokens/create` refuses it) and `is_user_token` (so `/users/tokens/update_label` refuses relabelling *into* it: a user token has no workspace pin, and one that could be relabelled this way would authenticate everywhere). Empty groups and folders, so every ACL denies independently and the token's scopes are its whole grant.
- A token minted *by* a guest (the embed token for a sandboxed low-code app, the SDK token for a sandboxed raw app) is a guest twice over: the same label (so it resolves at all) **and** the `guest` sentinel (so the workspace switch, the rate limit and the `whoami` role all still apply to the one credential handed to untrusted app JS). Both mints take these from one helper, and its expiry is capped at the *minting session's own expiry*, not a fresh interval: that expiry is a guest's only revocation.
- A guest-labelled token's scopes are immutable: `update_scopes` refuses it. The scopes are the guest's entire confinement, and after promotion the same email owns a real account that could otherwise strip them.
- The label is the single source of truth: `AuthCache` injects the `guest` sentinel on any guest-labelled token, so a credential that resolves as a guest is governed as one whatever its scopes say; a derived token is no longer one forgotten `push` away from being ungoverned.
- A guest is someone with **no account at all**: no `password` row (deactivated ones included: the sign-in path's own lookup filters on `disabled = false`, so a SCIM-offboarded account would otherwise read as absent and walk back in) and no `usr` row anywhere (what a service account has instead of a password).
- Guest tokens are pinned to their workspace (`token.workspace_id`), so an `apps:run:<path>` scope cannot unlock a same-path app elsewhere. A consequence worth knowing: a guest therefore cannot authenticate on any workspace-less route, which is intended and recorded at the pin.
- `authorize_non_member_viewer` factors the anonymous/RLS gate shared by `get_public_app_by_secret`, `get_app_embed_token` and the EE custom-path handlers; behaviour is unchanged for every non-guest caller. Whether a guest may act through an app at all (its mode must still be `guest`) is decided once in `get_on_behalf_details_from_policy_and_authed`, the resolver every on-behalf path (component runs, S3 reads, uploads) goes through, so an app flipped back to members-only closes to its existing guests everywhere at once. Which app a guest may act through is `guest_caller_for_mode`, applied once the policy is loaded on the run, S3 read and S3 upload paths alike: on an anonymous app the guest proceeds as itself (the run and the reads that follow it must carry one identity, or the page can never fetch the result it started); anywhere else a session for another app is refused.
- Fixes a latent bug found in review: `mint_app_embed_token` passed the `app_embed` sentinel to `ensure_scopes_within_caller`, which rejects one-part scopes, so *any* scope-restricted caller (a guest is the first) could not mint an embed token. Now mirrors `mint_raw_app_sdk_token`.

**Entry**
- Unauthenticated `apps_u/guest_entry/{secret}` and (EE) `apps_u/guest_entry_by_custom_path/{path}`: what a signed-out visitor reads to learn that signing in would let them in. 404 unless both gates say yes, so they say nothing about apps that are not open to guests.
- The sign-in card is offered by `PublicAppFrame` (the gate that actually fires first) on both the share-secret and custom-path routes, and the frame is not mounted until guest-entry discovery has settled, so a configured auto-login cannot race ahead and provision an account. It stops offering sign-in once one has happened and not helped, rather than looping. Safari's popup fallback polls a workspace-scoped route for a guest, since a pinned token cannot answer the global one. Each protocol carries the guest target through the IdP round-trip in its own channel and hands it to `login_externally` explicitly: OAuth in a `SameSite=Lax` cookie written by `/api/oauth/login`, bound to the flow's OAuth state (`<state>|<workspace>/<path>`) so a callback can use only the target of the flow whose state it validated, whatever other flows the same browser started meanwhile; no callback ever clears it, since a callback sees only its own request's snapshot of the jar and a clear from a slow one would land on a newer flow's entry, while a stale entry can never match a newer state (`Lax`, not the `None` of the existing `close` cookie: browsers drop `None` without `Secure`, so on a plain-http base URL the target would vanish and the callback would silently provision an account, which is exactly what the end-to-end pass below caught); SAML as a `guest_app` query parameter of `RelayState`, its native state channel, which the ACS lifts out; the ACS is a cross-site POST no cookie can be relied on to reach. The frame renders the app in parallel with guest-entry discovery; only the sign-in card waits for it, and only a confirmed 404 selects the ordinary sign-in. Any other failure retries, then fails closed with a visible error rather than offering a sign-in that would provision an account. A guest holding a session for one app who opens another guest app of the same workspace is offered sign-in (a 403 there is a session for the wrong app, which a fresh sign-in replaces). A SAML guest sign-in that cannot carry its target in `RelayState` does not start. An account holder who is not a member is told so on arrival (the frame probes for an account session whenever the app denies it) rather than being offered a sign-in that would loop. A sign-in the server refuses (the cap, an app no longer open to guests, an account that cannot be a guest) is relayed by the popup to the page that opened it and shown above the card in the server's words; before, the popup closed on it and the page saw nothing.
- Guest executions fall under `public_app_execution_limit_per_minute` alongside anonymous ones.
- `require_job_read_access` stops a guest after the launched-by-me grant exactly as it stops an app embed token, so a job share token whose audience is the workspace's members reads nothing for a guest, and a missing job and a forbidden one look the same.
- `logout` under `INVALIDATE_ALL_SESSIONS_ON_LOGOUT` deletes `guest_session` rows with the `session` ones, and the guest label is excluded from `username_override_from_label` like `session` is, so a guest's audit rows carry the email rather than `label-guest_session`.
- A sandboxed raw app's declared `frontend_sdk_scopes` are filtered to what a guest's session can actually delegate before the consent prompt is shown, so the prompt never promises a scope the mint would refuse.

**Migration**: `workspace_settings.guest_access_enabled`, plus a `guest_activity` table shaped like `feature_usage` (day-bucketed, `day` index, 60-day retention). Deliberately not an audit operation: the seat scan is served by a partial index whose predicate names the login operations literally, and `audit_partitioned` is partitioned, where `CREATE INDEX CONCURRENTLY` is unsupported; adding a guest operation there means a locking rebuild on the largest table an instance has.

## Screenshots

wm-guest-access-toggle

wm-guest-workspace-setting

superadmin-guests
workspace-guests-card
refusal-relayed

instance-switch-off

## Test plan

`backend/tests/app_guest_execution_mode.rs` (17 tests, passing on CE and EE feature sets; the upload one needs `parquet`, which CI builds with) and `backend/tests/app_guest_allowance.rs` (its own binary, because it flips the process-wide licence plan):
- [x] the allowance: with 100 emails in the window, a capped plan refuses a stranger with the visitor-facing message and lets a returning guest back in; the superadmin list and `guest_usage` report 100, unmetered, no seats; on an Enterprise plan the stranger is admitted and the 101st guest is one whole seat (`billable_guests: 1`, `guest_seats: 1`); plus a unit test on the rounding of `guest_seats`
- [x] a self-declared `guest` scope grants nothing: `users/tokens/create` refuses the label, `users/tokens/update_label` refuses relabelling into it, and a token carrying the scopes under another label does not authenticate as a non-member
- [x] the auth door enforces the workspace switch: a guest cannot authenticate at all while guests are off, and is through once on
- [x] a guest scoped to app A cannot run app B, even one open to guests
- [x] a guest-minted embed token (sandboxed app) resolves as a guest, does not outlive the session that minted it, is closed by the workspace switch, and cannot be rescoped by the same email holding a promoted session, nor can the guest session itself
- [x] a guest-labelled token with no `guest` sentinel is still reported and confined as a guest
- [x] a deactivated account is refused a guest session
- [x] a guest uses an anonymous app as itself: the component run starts and the guest reads its result back (the run and the read are one identity)
- [x] a guest cannot upload outside an app's `s3_inputs` policy: an app path with no row is refused by the confinement, and its own app without such a policy refuses it as it would an anonymous caller (the legacy branch uploads with the caller's own standing, which a guest has none of)
- [x] an app path carrying a character the scope grammar reserves (`:`, `,`, `*`) is refused at the mint before anything else, since the path is spliced into the session's scopes; discovery does not advertise such an app and the Guests mode cannot be deployed on one. Anything else (spaces, `@`, the shapes app paths legitimately take) is literal to that grammar and admitted
- [x] Guests mode cannot land on such a path however it gets there: set at creation, set on update, or a rename of an app already in that mode (checked against the destination, under the app-row lock so a policy update landing alongside cannot slip past); a path with a space is fine
- [x] a workspace-ID rename copies the guest switch with the rest of the settings and moves the workspace's `guest_activity` rows to the new id; a hard delete removes them with the other per-workspace rows
- [x] a guest cannot read a job it did not launch, and gets "not found" for it: it stops where an app embed token stops, before the share-token and ACL grants a member would get
- [x] the superadmin instance switch closes an issued session at the door and hides every app from discovery, whatever the workspace says, and reopens them when turned back on
- [x] an account created after the mint ends the guest session at the door on its next request
- [x] a guest reaches its own app and is denied `jobs/list`, `scripts/list`, `flows/list`, `variables/list`, `apps/list`, `resources/get_value`; `whoami` reports `role: guest` rather than falling through to the non-member superadmin shape
- [x] a guest token pinned to one workspace does not authenticate against another
- [x] `guest_entry` needs both the app mode and the workspace switch, and closes again when the switch is turned off

Exercised against a running instance (not only type-checked): guest opens its app 200 / denied on a members-only app 403 / all six confinement routes 403; workspace switch flips `guest_entry` between 200 and 404; clicking a rung persists and flips the gate; the public app page as a guest makes exactly four API calls, all 200, and renders on behalf of the publisher.

End to end through a real identity-provider round-trip (EE build with `oauth2`, a local mock OIDC provider registered as a login provider, headless Chromium): a signed-out visitor on the share link gets the guest card, clicks the provider, and the app renders on behalf of the publisher with **no `password` row, no `usr` row, a `guest_activity` row, a `guest_session` token pinned to the workspace** and `whoami` reporting `role: guest`; the same on the custom-path route; `/api/users/email` 401 and another workspace 401 from that session; the workspace switch turns the session into a 401 and back; revoking guest access between the login start and the callback (driven with curl, cookies carried) is refused with 403 and provisions nothing; a sandboxed anonymous custom-path app renders (anonymously) in a browser holding another app's guest session; the telemetry payload reports `guest_count: 2, seats_before: 2.0, seats_after: 1.0`; promotion (the same provider from `/user/login`) creates the account and lands on workspace selection; the promoted, non-member account is told the app is not open to it on arrival; a deactivated account through the guest link is refused and no session is minted. `saml_ee::tests::guest_target_is_lifted_out_of_relay_state` pins the `RelayState` contract; the ACS itself is exercised only by that unit test, as the SAML IdP round-trip has no local stand-in.

guest-app-as-guest

## Review history

Nine CI review rounds. Pi reached "good to merge" at round six; Claude has been at nit level since round three; Codex's last two rounds were a race and comment length. The substantive findings across those rounds, all fixed and each pinned by a test: the guest grant keyed on a forgeable scope (P0), the relabel and rescope twins, guest-derived tokens (embed, raw-app SDK) not governed as guests, the disabled-account re-entry, the sign-in gate firing before discovery, SAML dropping the target, and the workspace switch and app mode each being re-checked per handler instead of once at a choke point. The last two are the structural fixes: the switch is enforced once at the auth door, the mode once at the on-behalf resolver.

## Known gaps, deliberately not in this PR

- **The cap is exercised through the mint and the API, not through a browser on a Pro key** (the dev instance holds an Enterprise key). The refusal reaches the visitor the same way every sign-in refusal now does: the popup relays the server's sentence to the page that opened it, which shows it above the sign-in card; that relay is exercised live with the "no longer open to guests" refusal.
- **The SAML ACS is covered by a unit test of the `RelayState` split, not a live IdP round-trip.** The OAuth path is exercised end to end (above); SAML's differs only in where the target rides.
- **Someone who already has an account but is not a member of the target workspace cannot open a guest app.** See "Who is a guest" above: the page says so plainly instead of re-offering sign-in, and the UI copy says "without a Windmill account" rather than "anyone who signs in". Widening this is a product decision about what *guest* means, worth taking deliberately.


rubenfiszel:
Re the Codex nits on ba2c5a3dfa:

- **Overlong invariant comments** — tightened in 7dd5ccd374 (the `AuthCache` guest arm and the `create_guest_session_token` doc).
- **Test feature gate omits `parquet`** — dismissed: `a_guest_cannot_upload_outside_a_policy` already carries its own `#[cfg(feature = "parquet")]` (line 466), so without `parquet` it is compiled out rather than hitting the stub; the module gate covers what every other test needs.
- **Keyless Enterprise build: server admits guests, UI disables the controls** — dismissed as the codebase's existing convention rather than this PR's: an empty `$enterpriseLicense` makes every EE control in the UI read as CE (git-sync, SAML, S3 rules alike), while `get_license_plan` reads a missing key id as Enterprise for every server-side plan gate. Aligning the two is an instance-wide change, not a guest one.

rubenfiszel:
✅ Review round clean @ 7dd5ccd3741e9fc6829792ba475fc86dc28f0765 — nit-only verdicts at ba2c5a3dfa8431747caabb3b86db6cd4fa2b3886 (Claude and Pi: good to merge); nits addressed in 7dd5ccd3741e9fc6829792ba475fc86dc28f0765 / dismissed in the reply above

hugocasa (commented):
## Code review

**Verdict: address issues before merging.** The headline problem is that `guest` mode does not survive a git-sync / `wmill sync push` round-trip, which is precisely the property the PR's architecture is justified by.

### 1. [P1] The CLI silently demotes a `guest` app to `publisher`

On push, both app paths do:

```ts
app.policy.execution_mode = publicApp ? "anonymous" : "publisher";
```

`cli/src/commands/app/app.ts:238`, `cli/src/commands/app/raw_apps.ts:537` — and `publicApp` comes from `isExecutionModeAnonymous` (`app.ts:110`), so it is only ever true for `anonymous`.

The pull side compounds it: `sync.ts:1399` sets `app.policy = undefined` and only preserves `public: true` for `anonymous` (same at `sync.ts:1419` for raw apps). `guest` is not representable in the tracked `app.yaml` at all, so a pull/push cycle turns guest access off with no diff to show for it.

This contradicts the PR's own premise — "an app carries its `execution_mode` in its definition, so git-sync and the CLI push `guest` past every UI gate" — which is the stated reason the workspace switch is enforced at the auth door rather than at deploy time. Fixing it needs a tri-state carried through pull/push (an `isExecutionModeGuest` counterpart), not the current boolean `publicApp`.

### 2. [P1] Non-member account holders can open a guest app, contradicting the PR description and the UI copy

In `authorize_non_member_viewer` (`backend/windmill-api/src/apps.rs`):

```rust
if matches!(mode, ExecutionMode::Guest) {
    if is_guest { check_scopes(authed, || format!("apps:read:{}", app_path))?; }
    return Ok(true);   // any authed caller, guest or not
}
```

For a `Guest`-mode app, **any** authenticated caller short-circuits to authorized before the RLS read-probe runs — including an account holder who is a member of a different workspace entirely. They then run the app as the publisher, since `get_on_behalf_details_from_policy_and_authed` treats `Guest` like `Publisher`.

The PR description lists the opposite as a deliberate known gap ("Someone who already has an account but is not a member of the target workspace cannot open a guest app"), and the editor and sign-in-card copy promise guests are people "without a Windmill account". Either the check is missing or the description and copy are wrong. It changes who can open a guest app instance-wide, so it is worth settling before merge.

### 3. [P2] Guests are admitted ahead of `REQUIRE_PREEXISTING_USER_FOR_OAUTH`

`oauth2_ee.rs` guest branch in `login_externally`. On an instance whose superadmin set that flag, any workspace admin can now admit arbitrary IdP-authenticated identities by flipping `guest_access_enabled` and setting one app to `guest`. The comment's reasoning (the flag gates *provisioning*, and a guest provisions nothing) is defensible, but there is no instance-level control that turns guests off — the three gates are the license plan, a workspace-admin switch, and a per-app mode. Either honor the flag here, add an instance setting, or update the flag's documented meaning.

### 4. [P2] `logout` does not revoke a guest token under `INVALIDATE_ALL_SESSIONS_ON_LOGOUT`

`backend/windmill-api-users/src/users.rs:665` — that branch deletes `WHERE email = … AND label = 'session'`, so a `guest_session` row survives, `email` comes back `None`, and no `users.logout` audit row is written. The cookie is still cleared so the UX is unaffected, but given the PR's own position that "the expiry is a guest's only revocation", logout was the one remaining user-driven revocation. (The default branch deletes by `token_hash` and is fine; disable/delete-user and SCIM offboarding do unfiltered `DELETE FROM token WHERE email = $1`, also fine.)

### 5. [P2] Every guest carries `username_override = "label-guest_session"`

`backend/windmill-api-auth/src/auth.rs:937` — the catch-all arm of `username_override_from_label` excludes `"session"` and `"ephemeral-script"` but not the new label. `display_username()` still returns the email (the `is_token_label` flag is true), so `v2_job.created_by` is correct, but the audit `end_user` field records the literal `label-guest_session` for every guest action, and `require_job_read_access` does a pointless extra `permissioned_as_email` lookup on any job whose `created_by` matches it. `guest_session` is the direct analogue of `session` and belongs alongside it.

### 6. [P2] Comment-policy compliance

AGENTS.md: "State each invariant once, at the place where someone would break it, in ≤4 lines… don't repeat the same rationale at multiple sites." The "a guest has no `usr` row, no `password` row, therefore no seat" rationale is restated at roughly a dozen sites (the migration, `create_guest_session_token`, `guest_session_scopes`, `ExecutionMode::Guest`, `GUEST_SESSION_LABEL`, `is_guest_access_enabled`, `guest_access_licensed`, the `AuthCache` guest arm, `whoami`, `authorize_non_member_viewer`, `has_guest_sentinel`, the settings field, the test module doc, and three OpenAPI descriptions), and several individual comments run 12–15 lines. Worth collapsing to one canonical statement plus pointers.

### What checked out as sound

- **The label/scope split holds.** `GUEST_SESSION_LABEL` is the grant, reserved in both `is_server_minted_label` (blocks `/users/tokens/create`) and `is_user_token` (blocks relabelling into it); `update_scopes` refuses guest-labelled rows. Every site branching on `has_guest_sentinel` only ever narrows — a forged sentinel grants nothing it would not already have, and the two early-check skips are re-checked once the policy loads.
- **The workspace pin is real.** The token lookup's `(workspace_id IS NULL OR workspace_id = $2)` yields NULL, not true, when `w_id` is absent, so the `w_id.is_none()` arm of `get_opt_job_authed_inner` is unreachable for guests. Derived embed/SDK tokens are pinned too and cannot extend lifetime (`MIN(expiration)` cap, errors closed when there is no parent).
- **The switch is enforced at a single choke point.** `is_guest_access_enabled` runs outside `get_opt_job_authed_inner`, so it is never served from `AUTH_CACHE`; `get_opt_job_authed` is the only entry point (`get_authed` delegates to it). DB errors on that read fail closed.
- **Route confinement.** `guest_route_denied` sits on top of the reviewed `app_embed` allowlist; the `apps_u/*` routes a guest can reach are all in `unauthed_service()` and grant nothing over anonymous. By-id job reads fall through to the RLS probe, which denies with empty groups and folders.
- **Cookie/protocol handling.** `SameSite=Lax` is correct for the OAuth callback (top-level GET) and wrong-by-necessity for the cross-site SAML ACS POST, which is why the target rides `RelayState`; `login()` overwrites the cookie with `""` on every start, so an abandoned guest visit cannot turn a later plain sign-in into a guest session.
- Migration, retention delete, `summarized_schema.txt`, OpenAPI additions and dev-dependency placement are all correct. No `crypto.randomUUID`, no drafting-history comments.

### Test coverage

`backend/tests/app_guest_execution_mode.rs` is well targeted — each test pins a distinct invariant (label-not-scope, workspace pin, the door re-reading the switch, cross-app confinement, derived-token label + expiry cap + immutable scopes, label-without-sentinel, disabled-account refusal, anonymous-app single identity, upload confinement, `guest_entry` needing both gates) and none reads as scaffolding under the "ship only the tests the PR needs" rule. The `saml_ee` `RelayState` unit test is the right shape for a path with no local IdP stand-in.

Two gaps worth naming:

- Nothing pins that `guest` survives a CLI round-trip. A `cli/test/` case asserting that pulling and re-pushing an app with `policy.execution_mode: guest` leaves the deployed mode unchanged would have caught issue 1, and is the regression guard that fix needs.
- `login_externally`'s branch selection (account holder → ordinary path; no account + admitting target → guest; no account + revoked target → 403) is exercised only by the manual end-to-end pass described in the PR. A test over `has_any_account` × `guest_target` × `guest_app_admits` would be cheap and pins the one place an account could be laundered into a guest identity.

**Still worth exercising manually:** the git-sync round-trip behind issue 1 (deploy in Guest mode, `wmill sync pull` then `push`, confirm the deployed policy still reads `guest` and the share link still shows the guest sign-in card); the Pro-plan refusal in both directions (toggle disabled with the Enterprise hint, `POST /workspaces/edit_guest_access` 400, `guest_entry` 404 on an app already set to `guest`, `get_settings` reporting `guest_access_enabled: false` for a workspace whose column is `true`); and opening guest app B while holding a session for guest app A on an instance with `INVALIDATE_ALL_SESSIONS_ON_LOGOUT` set, which interacts with issue 4.

rubenfiszel:
Re the review at ba2c5a3 (thanks, the CLI one was the big miss), all addressed in 2ce9118f60 except where noted:

1. **CLI demotes guest to publisher**: fixed. The access mode is now a tri-state in `app.yaml` (`public: true`, `guests: true`, neither), carried by `markAccessFromPolicy` on pull and `executionModeFromAppFile` on push for low-code apps, raw apps and the sync bundle; `cli/test/app_access_mode_unit.test.ts` pins the round trip.
2. **Non-member account holders can open a guest app**: dismissed, with the reasoning. The `usr`-row check is at the auth door, not in `authorize_non_member_viewer`: a token for an account with no `usr` row in the route's workspace resolves to `None` in `get_opt_job_authed_inner` (only superadmins and guest-labelled tokens survive that arm), so by the time `authorize_non_member_viewer` sees `Some(authed)` on a guest app it is a member, a guest, or a superadmin. Exercised live: the promoted, non-member account gets 401 on the embed-token mint and the page shows "You are signed in, but this app is not open to you". The one identity that does reach it without a row is an external JWT, whose key is root by design. Since 2ce9118f60 the door also re-reads "no account" for every guest request, so a guest session cannot outlive an account either.
3. **Guests ahead of `REQUIRE_PREEXISTING_USER_FOR_OAUTH`**: fixed by honouring the flag. A guest is the very identity it keeps out, and it is the instance-level control you asked for; an instance that sets it has no guests. Written up in the PR body.
4. **Logout under `INVALIDATE_ALL_SESSIONS_ON_LOGOUT`**: fixed, `guest_session` rows go with the `session` ones.
5. **`label-guest_session` as `username_override`**: fixed, excluded next to `session`.
6. **Repeated rationale**: collapsed to one canonical statement on `ExecutionMode::Guest` (plus the mint, where someone would break it); the other sites now point at it.

Test gaps: the CLI round trip is pinned (above); `login_externally`'s branch selection is EE code with no OSS test seam, so it stays on the manual pass, which now also covers the revoked-target refusal reaching the page.

rubenfiszel:
Nits from the round at f626dc312f, addressed in 07dfa0ba8b: `is_guest_access_enabled` removed (its callers had moved to `guest_app_admits` and `guest_session_stands`, orphaned sqlx entry deleted); the instance setting is named once through `GUEST_ACCESS_DISABLED_SETTING` and read with `(value #>> '{}')::boolean` so a string-encoded value counts too (the direct jsonb cast did work, live and in the test, but this is sturdier); the Guests tab shows a skeleton or a retry alert instead of the Users table while its list is missing; the instance toggle is disabled during its write and remounts on the stored value after either outcome, so two quick clicks cannot cross and a refused write cannot leave it showing the click; the workspace card saves the guest switch first, so on Community and Pro a dirty Enterprise-only field beside it cannot swallow the write.

✅ Review round clean @ 07dfa0ba8baf1023efaa20fe8b5d175be2740e8f — nit-only verdicts at f626dc312fabe5ea9332482bc3780fc91eb9e68b; nits addressed in 07dfa0ba8baf1023efaa20fe8b5d175be2740e8f

rubenfiszel:
Nits from the round at 47dbd96a92, addressed in 0241ba52b4: the deploy-time guard now checks the path the app ends up on (a rename-only update included, keeping the deployed mode) and refuses a leading slash as well, and `deployment_rule_for_mode` has its doc back. Dismissed: cubic's note that three of `loadGuestPage`'s four callers ignore its boolean; the toggle is the one that needs it.

✅ Review round clean @ 0241ba52b483e274b284682c06231f7dd771fa7c — nit-only verdicts at 47dbd96a92f9ef891a4b3e9759e759a299dba2e9; nits addressed in 0241ba52b483e274b284682c06231f7dd771fa7c

rubenfiszel:
Nit from the round at cb5d00b83b (Codex and Pi: good to merge), addressed in 6169076d70: `guest_activity` now follows a workspace rename and goes with a workspace delete, like the other per-workspace rows; the rename test asserts the move.

✅ Review round clean @ 6169076d70f36b509f940aa19412fda3c41ea5d7 — nit-only verdicts at cb5d00b83b5407562d06823030a5b93e05d071e6; nits addressed in 6169076d70f36b509f940aa19412fda3c41ea5d7

rubenfiszel:
Nits from the round at 943789533d (Codex and Pi: good to merge), addressed in 6081e286c9: the workspace-scoped `guest_activity` delete now says it also moves the instance-wide allowance count, and the allowance test asserts the mint writes the `guest_activity` row that count is built on.

✅ Review round clean @ 6081e286c969da11aadf887d46b7c5b6553bde08 — nit-only verdicts at 943789533dc71004757abb1fb421c2b8bc148f16; nits addressed in 6081e286c969da11aadf887d46b7c5b6553bde08

rubenfiszel:
The P2 on 6081e286c9 (the "mint records the guest" assertion was satisfied by the seed) is fixed in c7c421b3b8: the allowance is now seeded a day old, still inside the window, so a `guest_activity` row dated today can only be the mint's own; the assertion checks that row.

✅ Review round clean @ c7c421b3b8ca39bba6e3a22cba71153997401b61 — nit-only verdicts at 943789533dc71004757abb1fb421c2b8bc148f16; nits addressed in c7c421b3b8ca39bba6e3a22cba71153997401b61

PR #10930, opened 2026-09-01
feat: make the home Build with AI composer dismissible, quiet the rest of the home page

rubenfiszel:
## Summary

The home page led with three things competing for attention: a full-bleed AI composer, a card-styled tutorial banner, and two Accent buttons. This makes the composer dismissible, shrinks the tutorial banner to an inline row, and enlarges the empty state.

The composer is also hidden entirely in workspaces that are meant to be run rather than authored in: those where the `DisableDirectDeployment` protection rule is active. That covers both a prod paired with a dev workspace (the pairing creates the `dev_workspace_lock` ruleset) and any workspace an admin has locked down. The gate is deliberately a property of the workspace, not of the caller: a session started from a locked prod would be steered into the paired dev workspace, and an admin bypasses the lock outright, but neither makes prod the place to start authoring.

Three commits: the composer feature, the tutorial banner restyle, and the empty state.

## Changes

**AI composer (`HomeAIChat.svelte`, `CloseButton.svelte`)**

- A close button at the top-right of the composer dismisses the whole block (title, textarea, model picker, example prompts). Collapsed, the only thing left is a subtle `Build with AI` button in the left slot of the row that already carries CLI / MCP and Hub, so the home page reduces to one quiet line.
- The state persists per browser under the `home-ai-composer-collapsed` local setting, via the existing `getLocalSetting` / `storeLocalSetting` helpers.
- The composer and its reopen button are both hidden when `DisableDirectDeployment` is active. Unresolved rules read as unlocked, so the far commoner unlocked workspace never pops the composer in mid-load.
- The "no AI provider" blur and overlay now cover the input alone instead of the whole block. The title, the example prompts and the connect row stay legible and usable without a provider, and the overlay carries its own secondary `Hide` next to `Configure AI` so there is exactly one dismiss control per state.
- The blurred wrapper is `relative` in both states: `blur-sm` is a filter, which makes an element the containing block for its absolute children, so the send button and model picker would otherwise shift when the blur turns on.
- The typewriter placeholder effect stops while collapsed instead of driving an unrendered input.
- `CloseButton` gains an optional `title` prop forwarded to `Button`, so an icon-only close button can carry an accessible name.

**Tutorial banner (`TutorialBanner.svelte`)**

- From a full-width bordered card to a single inline row: no fill, no border, no subtitle, no duplicated graduation-cap icon.
- `View tutorials` becomes `Tutorials` and drops from Accent to the design system's secondary tier. The home page had two Accent buttons where the guidelines allow one; `accent-secondary` is marketing-only.
- The prompt shortens to `First time?`. The `New tutorial available!` state is unchanged.
- Left-aligned, on the same gutter as the tab group and item list below it, and moved above the composer.
- The hand-rolled `<button>` dismiss becomes `CloseButton`, per frontend/AGENTS.md on raw HTML elements. The CTA moves off the deprecated `size` prop to `unifiedSize`.

**Layout (`+page.svelte`, `NoItemFound.svelte`)**

- The `mt-20 mb-16` wrapper around the composer moved into `HomeAIChat`, which now picks its own vertical spacing. Hero margins around a bare connect row were just empty page, so the collapsed, operator and run-only views get `mt-8 mb-2`, which puts about the same gap under that row as above it.
- The empty-workspace line goes from `text-xs` to `text-lg`, keeping its `text-hint` weight and colour.

## Screenshots

Expanded, with the tutorial row above it and the larger empty state below:

expanded

Collapsed, persisted across reloads:

collapsed

No AI provider configured: blur and overlay on the input only, `Hide` alongside `Configure AI`:

disabled

Workspace with `DisableDirectDeployment` active, composer and reopen button both gone:

prod-hidden

Dark mode:

banner-dark

## Test plan

- [x] Click the close button: composer, example prompts and title all go, one row remains
- [x] Reload: still collapsed, no textarea rendered
- [x] Click `Build with AI`: composer, typewriter placeholder and example tags return, setting cleared
- [x] Add a `DisableDirectDeployment` protection ruleset to the workspace: composer and reopen button both gone, `Edits restricted` badge shows
- [x] Unset the workspace AI config: title and example prompts stay sharp, only the input is blurred, and the overlay's `Hide` collapses the composer
- [x] Tutorial banner renders as one inline row above the composer, `Tutorials` navigates, the X dismisses and toasts
- [x] Dark mode and a 900px viewport, both composer states
- [x] `npm run check:fast` clean for the changed files

No test is included: the change is a boolean plus a localStorage key and some Tailwind classes, with no pure-logic utility a future change could break silently.

https://claude.ai/code/session_01QjJhxHHqRqyEX7HsbPjetn

### DIFF
--- a/frontend/src/lib/components/common/CloseButton.svelte
+++ b/frontend/src/lib/components/common/CloseButton.svelte
@@ -10,10 +10,12 @@
 		Icon?: any | undefined
 		class?: string
 		id?: string | undefined
+		/** Names the button: it has no text, so without this it reads as "button" to a screen reader. */
+		title?: string | undefined
 		onClick?: () => void | undefined | any
 	}
 
-	let { noBg = false, small = false, Icon, class: className, id, onClick }: Props = $props()
+	let { noBg = false, small = false, Icon, class: className, id, title, onClick }: Props = $props()
 
 	const dispatch = createEventDispatcher()
 </script>
@@ -22,6 +24,7 @@
 	on:click={() => (dispatch('close'), onClick?.())}
 	on:pointerdown={(e) => e.stopPropagation()}
 	{id}
+	{title}
 	startIcon={{ icon: Icon ?? X }}
 	iconOnly
 	unifiedSize="sm"
--- a/frontend/src/lib/components/home/HomeAIChat.svelte
+++ b/frontend/src/lib/components/home/HomeAIChat.svelte
@@ -20,23 +20,45 @@
 
 <script lang="ts">
 	import TextInput from '$lib/components/text_input/TextInput.svelte'
-	import { ArrowUp, ExternalLink, Globe2, KeyRound, PlugZap, Settings } from 'lucide-svelte'
+	import {
+		ArrowUp,
+		ExternalLink,
+		Globe2,
+		KeyRound,
+		PlugZap,
+		Settings,
+		WandSparkles
+	} from 'lucide-svelte'
 	import Button from '../common/button/Button.svelte'
 	import { Badge } from '../common'
+	import CloseButton from '../common/CloseButton.svelte'
 	import { startSessionWithPrompt } from '../sessions/sessionSwitch.svelte'
 	import { copilotInfo, copilotWorkspace } from '$lib/aiStore'
 	import { loadCopilot } from '$lib/components/copilot/loadCopilot'
 	import { aiUserDisabled, hubBaseUrlStore, userStore, workspaceStore } from '$lib/stores'
 	import { HOME_SHOW_HUB } from '$lib/consts'
 	import { base } from '$lib/base'
+	import { getLocalSetting, storeLocalSetting } from '$lib/utils'
+	import { isRuleActive } from '$lib/workspaceProtectionRules.svelte'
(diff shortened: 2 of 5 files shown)


rubenfiszel:
Re: Pi's P2 on `TutorialBanner.svelte` — dismissing, the condition it asks for is already met.

The change is intentional and rode along by request. It is called out in the PR description: a paragraph in the Summary, a bullet in Changes naming the exact `variant="accent"` → `variant="default"` swap and why `accent-secondary` was not used, a line in the Test plan, and a screenshot. Pi appears to have read the body before that edit landed.

The rationale, for the record: the brand guidelines put `default` at the secondary tier ("Secondary actions and most UI interactions") and allow only one Accent button per view. The home page had two, `View tutorials` and `New`, so this is a hierarchy fix rather than a stray edit.

rubenfiszel:
✅ Review round clean @ b61b90f35d0a23f62d209310b688ed49fdb316c6 — nit-only verdicts at b61b90f35d0a23f62d209310b688ed49fdb316c6; the single nit (unrelated `TutorialBanner` variant change) is dismissed in a review reply: it is intentional and already documented in the PR description.

PR #10936, opened 2026-09-02
[ee] feat(bedrock): add OIDC role assumption as a fourth auth mode

hugocasa:
## Summary

AWS Bedrock resources could authenticate three ways: a bearer API key, explicit IAM keys, or the server's ambient IAM role. All three make every Bedrock call look the same on the AWS side, so CloudTrail cannot tell which job or which user drove the spend.

This adds a fourth mode: an optional `oidcRoleArn` on the Bedrock resource. When set, Windmill mints an OIDC token for the caller against audience `sts.amazonaws.com`, exchanges it for temporary IAM credentials through STS `AssumeRoleWithWebIdentity`, and signs the Bedrock request with those. The STS role session name carries the caller, so CloudTrail attributes the request:

- AI agent steps: session `windmill-ai-job-<first 8 chars of job id>`, with the job-claim token `/oidc/token` issues.
- Copilot / chat proxy: session `windmill-ai-copilot-<username>`, with a workspace-claim token whose `sub` is `<email>::<workspace>`.

Same OIDC role-assumption pattern already used for S3 and SQS triggers, and enterprise-only for the same reason: the OIDC signing key lives in EE.

Precedence is `apiKey` > IAM keys > OIDC role > environment, so a resource can carry a role ARN without it ever overriding keys someone set explicitly. An OIDC role requires a region: the assumption resolves to explicit IAM keys, and the SDK client built from those has no ambient-region fallback the way the environment mode does, so an unset region is refused with a clear message instead of producing a client pointed at an empty region.

## Design notes

Two deviations from the ticket sketch, both because of where the OIDC signing key lives:

- **No `OidcRole` variant on `BedrockAuthConfig`.** `create_bedrock_client` gets only `ProxyBuildArgs`, which has no `DB` and no `AuthedClient`, so it cannot mint a token — the variant would be unreachable. Instead the precedence rule lives in one shared helper (`bedrock_oidc_role_to_assume`) used at the two boundaries that *can* mint a token, and both Bedrock client builders keep seeing concrete IAM keys.
- **The proxy exchange runs after the workspace credential cache, not inside `resolve_provider_credentials`.** `AI_REQUEST_CACHE` is keyed by `(workspace, provider)`; caching STS output there would hand one user's session to every other user in the workspace, which defeats the per-user attribution this exists for. The workspace cache now holds the role ARN (no secrets), and a second cache keyed by `(workspace, role ARN, user email)` holds the temporary credentials, re-assumed 2 minutes before AWS expires them. The worker holds the equivalent state for the length of one job, so a tool-heavy agent loop assumes once rather than on every model turn.

## Changes

- `windmill-ai/src/ai_bedrock.rs`: `bedrock_oidc_role_to_assume` (the precedence rule), `assume_role_with_oidc_token` (STS exchange, reusing `windmill_common::auth::aws`), `refresh_bedrock_oidc_credentials` (the worker's assume-once-per-job path), `aws_role_session_name` (AWS's `[\w+=,.@-]{2,64}` constraint), `AssumedRoleCredentials`.
- `windmill-ai/Cargo.toml`: the `bedrock` feature now pulls `windmill-common/aws_auth` for the STS client.
- `windmill-ai/src/credentials.rs`, `types.rs`, `windmill-api/src/ai.rs`: `oidc_role_arn` on `ProviderCredentials`, `ProviderResource` and `AIStandardResource`, serde alias `oidcRoleArn`, resolved through `$var:` like the other credential fields.
- `windmill-api/src/ai.rs`: `resolve_bedrock_oidc_credentials` in the proxy path plus the per-user assumed-role cache. Gated so a build without `enterprise,openidconnect,private` returns a clear error rather than a signing failure.
- `windmill-worker/src/ai_executor.rs`: assumes the role once per job and refreshes only near expiry; `BedrockQueryBuilder::execute_request` is unchanged and still only sees concrete keys.
- `frontend/src/lib/components/BedrockCredentialsCheck.svelte`: points at the OIDC option as the way to get per-job/per-user attribution, shown only where a license makes it usable.

## Merge order

Three PRs, and the sequencing matters:

1. windmill-labs/windmill-ee-private#770 — the OSS build resolves `*_ee.rs` through `backend/ee-repo-ref.txt`, which this PR bumps to that branch's head.
2. windmill-labs/windmill-integrations#181 — adds `oidcRoleArn` to the Bedrock resource-type schema. Until it ships, the resource form renders no field for it, and the copy this PR adds points at something a user cannot fill in.
3. This PR.

## Screenshots

Bedrock resource form with the new `oidcRoleArn` field and the updated credentials-check copy:

bedrock-oidc-resource-form

## Test plan

Exercised against a local backend built with `quickjs,bedrock,private,enterprise,openidconnect,license`, with `AWS_ENDPOINT_URL_STS` pointed at a local STS stub that records what was asked for and hands back fake temporary credentials. For the agent-loop run, a local Bedrock ConverseStream stub answered turn 1 with a tool call and turn 2 with text so the loop actually iterated. For the proxy runs, Bedrock was the real AWS endpoint, so `UnrecognizedClientException: The security token included in the request is invalid` is the proof that the stub's credentials — session token included — actually signed the request.

- [x] Copilot proxy: STS received `AssumeRoleWithWebIdentity` for the resource's role, session `windmill-ai-copilot-admin`, token `aud=sts.amazonaws.com`, `sub=admin@windmill.dev::admins`, with `email` and `workspace` claims. AWS then rejected the stub credentials as expected.
- [x] Proxy cache: two consecutive requests, one STS call.
- [x] Worker, multi-iteration: an AI agent step with a tool ran two model turns and finished successfully; both Bedrock requests were signed with the assumed-role key and session token, and STS was called **once** for the whole job. Session `windmill-ai-job-01a0614d` matched the job id, with the job-claim token.
- [x] Precedence: a resource with IAM keys *and* `oidcRoleArn` made no STS call, and the Bedrock request was signed with the explicit key.
- [x] Fallback unchanged: a resource with only `region` made no STS call and failed at the ambient-environment lookup as before.
- [x] Region guard: `oidcRoleArn` with an empty region returns `AWS Bedrock resources that assume a role through OIDC must set a region`.
- [x] Session names: a whole job UUID rides through (`windmill-ai-job-<32 hex>`, 48 chars); a name that would exceed AWS's 64 keeps a prefix plus a digest of the full identity, so two long identities sharing a prefix stay distinct in CloudTrail (unit-tested).
- [x] `cargo check` (no features), `cargo check -p windmill-api --features enterprise,openidconnect,bedrock` (the `check_oss_full` shape, no `private`), `cargo check --features enterprise,private`, `cargo check --features bedrock,enterprise,private,openidconnect,license,quickjs`, `cargo test -p windmill-ai --features bedrock` (138 passed), `svelte-check` (0 errors).

### Not exercised

Everything above ran against a local STS stub, so the real AWS leg is unverified: I have no AWS account here to create a role with a trust policy. What a reviewer with one should check is that **both** callers can assume the same role, because their token subjects differ by construction — `<email>::<workspace>` from the copilot proxy, `<email>::<path>::<flow_path>::<workspace>` from an AI agent step. A trust policy conditioned on only one shape will leave the other failing at assume time. Worth checking the instance-level `ai_config` case too, where the resource lives in `admins` but the token's workspace claim is the caller's workspace.

## Left in draft

Review round is clean at `bceb740deb297460c6c19f18319cc04fb746f9f4` — Codex, Claude and Pi all "Good to merge". Kept in draft anyway: it touches an EE file (companion PR windmill-labs/windmill-ee-private#770), it is an auth/credential path, it depends on the merge order above, and the real-AWS leg is unverified. Ready to flip once someone has looked.


hugocasa:
✅ Review round clean @ bceb740deb297460c6c19f18319cc04fb746f9f4 — Codex, Claude and Pi all "Good to merge". Deliberately left in draft (EE companion PR, auth/credential path, cross-repo merge order, real-AWS leg unverified); see the PR description.

hugocasa:
Thanks — fixed two of the three in `7c38551bdb`.

**1. `BEDROCK_ASSUMED_ROLE_CACHE` not invalidated (P3)** — fixed. `invalidate_ai_request_cache_for_workspace` now clears both caches, and the existing `invalidates_all_cached_providers_for_workspace` test is extended to pin it.

Worth noting the framing though: clearing this cache does not revoke anything. An STS session that has already been minted stays valid at AWS until it expires no matter what Windmill drops, and the worker deliberately holds one for the length of a job. The cache key also contains the role ARN, so rotating to a different role already misses, and a user who lost access fails the RLS check in `resolve_provider_credentials` before reaching the cache. The fix is worth having because two caches behind one `invalidate` call should both clear — not because it closes a permissions window.

**2. Not single-flight (P2)** — not fixing. `AI_REQUEST_CACHE`, immediately above it and reached on the same request, has the identical race: concurrent misses each resolve credentials before the first insert lands. The cost here is a couple of redundant `AssumeRoleWithWebIdentity` calls that converge on the next request, far below any STS limit. Adding a keyed mutex to this one cache would harden it against a pattern its sibling shares, which is the kind of local inconsistency that confuses the next reader more than the duplicate calls cost. Happy to revisit if both get it together.

**3. Region not mentioned in the hint (P3)** — fixed, the copy now says to set a region alongside `oidcRoleArn`.

Verification for 1: `cargo check` with and without the `bedrock` feature, the extended unit test, and a runtime check against an STS stub — two proxy calls produced one STS call, then `edit_copilot_config`, then a third call produced a second STS call. On the old code that third call reused the cached session.

PR #10938, opened 2026-09-02
feat: bring gitlab repositories to parity for git sync

hugocasa:
## Summary

Git sync's managed features (instant pull via webhook, promotion and fork PRs, the PR diff check, the repo picker) were all gated on `is_github_app`, a boolean on the `git_repository` resource. That flag conflated two things: *the server holds a credential for this repository*, and *the managed features are available*. Only the first is GitHub-specific.

This gates the features on the first instead, and adds GitLab as a second provider behind that seam. A GitLab repository whose project access token is handed to Windmill through the picker gets webhooks, merge requests, MR diff previews and credential renewal, with no app to install. GitHub App repositories behave exactly as before, and so does every repository whose credential the server does not hold: a token written into the URL, GitHub or GitLab, is a plain git remote that syncs on deploy and by polling.

## How the credential is stored

The token lives encrypted on `workspace_settings.git_credentials`, keyed by the repository it was issued for (scheme, host, port and path), and the resource's `url` carries no credential. Each entry also records its provider, so a fork resolving the credential knows which host it talks to before any check of its own has run.

It started in a secret Windmill variable, on the reasoning that a long-lived credential belongs in the object built for secrets. That missed forks: `clone_variables` copies secret variables *with their values*, so every fork got its own copy of the token, and the first rotation revoked the copies the other forks were still using. A fork's `workspace_settings` row is built from a fixed short column list, so it inherits nothing and resolves its ancestors' credential instead — the same shape app-backed repositories already rely on. One stored copy is what makes rotation correct: the owner rotates, every fork sees the replacement immediately, and a fork resolving an ancestor's credential reports `rotatable: false` so it never races to renew a token it does not own.

A stored credential is refused for any other URL. The `url` field it would authenticate is writable by anyone with write on the resource path, so keyed by resource instead, repointing one would send the token to a host of the writer's choosing. The scheme is part of the key so an `https` credential is never served over `http`.

A token written into the URL, inline or in a `$var:` the resource points at, keeps working exactly as it does on `main`: the server does not know it is there, so it registers no webhook, opens no merge request, and neither reports nor renews its expiry. Managed means held; there is no third state.

### Why credential removal is not wired to git sync settings

Removing a repository from git sync settings, or clearing them, leaves its stored credential in place. The credential belongs to a repository, not to any one resource or git-sync entry: several resources can name the same repository, forks resolve it through their ancestors, and a resource outlives its git-sync membership. Deleting it on one of those events would silently stop every other user of the repository authenticating.

Deletion is not wired to resource deletion either, for the same reason, which matches how the existing credentials behave:

- **A token in the URL** lives inside the resource, or in a `$var:` variable that `delete_resource`'s existing linked-variable cascade already removes. Nothing separate to clean up.
- **A GitHub App installation** is keyed by installation, not by resource. Deleting every git repository resource in a workspace leaves the installation — and its cached tokens — until someone removes it explicitly.

So for a server-held credential the precedent is that its lifetime is independent of any one resource. An orphaned entry is inert: it is write-only over the API, and the token endpoint fails at the resource fetch long before it reaches the credential. An explicit way to remove a stored credential belongs with a UI affordance for it, the way the App has "remove installation", rather than being inferred from an unrelated action.

## Changes

- `is_github_app` gates replaced with "the server holds a credential for this repo, and knows the provider" (`result_processor.rs`, `git_sync_ee.rs`, the deploy check marker).
- New `workspace_settings.git_credentials` column, plus the credential store, the repository binding, and ancestry resolution for forks.
- GitLab provider: project listing, webhook receiver authenticated by the per-hook secret token, merge request creation, MR diff preview notes, and token introspection.
- `GET /w/{workspace}/git_sync/credential/origin?path=` answers whether the workspace holds the credential for a repository or borrows an ancestor's. The repository card and the resource editor read it instead of a marker on the resource, which went stale on export, import, repoint and fork.
- Credential renewal in `monitor.rs` via GitLab's self-rotation endpoint, with an expiry ladder on the repository card. Rotation is never retried — GitLab revokes the old token on success and treats a replayed one as reuse, revoking the whole family including the live replacement.
- The two webhook handlers are folded into one: the providers now differ only in how a delivery is parsed into a push or PR event, and every decision after that is shared.
- GitLab picker in the resource form, alongside the GitHub App button.

## Notes for review

- **No commit statuses on GitLab.** A status there files as a job inside the customer's pipeline, `allow_failure` is ignored, and an external pipeline can gate merging. The PR diff preview goes in an upserted MR note instead.
- **Two `GitHubAppIntegration.svelte` buttons move from the deprecated `size="xs"` to `unifiedSize="sm"`.** Only so the new GitLab button next to them is not a different height.
- **The `/github_app/token` route now also serves a stored credential.** The endpoint already resolves and authorizes the job's repository, so putting the GitLab branch inside it avoids a second path that could drift on the authorization check. The route name is now broader than what it says; renaming it would churn the generated client for no behavioural gain.
- **Workspace admin is equivalent to holding the token.** An admin of the workspace, or of any fork below it, can point a repository at a sync script they wrote and have that job request the credential — as they can for an App installation token. This storage keeps the token out of the variables API and out of every descendant's own storage; it is not a boundary against those admins.

## Screenshots

The GitLab button in the `git_repository` resource form. Pick the instance and paste a project
access token, list the projects it reaches, and choose one: the token is stored against that
repository when you pick it, and the resource gets the plain remote URL.

gitlab-picker

A repository whose token Windmill holds, with the replacement flow. The URL carries no credential,
so this is what tells you the token is Windmill's to keep and renew.

managed-credential-alert

## Known limitations, deliberately not addressed here

Both are instances of a class of problem that exists on `main` today, so fixing either for GitLab alone
would leave the providers inconsistent. They are better as their own changes.

- **A workspace key rotation does not re-encrypt server-held git credentials.** `set_encryption_key`
  walks the `variable` table only, so anything encrypted with the workspace key elsewhere is
  stranded. `git_sync.webhook_secret` and user drafts are already in that set on `main`;
  `git_credentials` joins it here, and after a rotation the token has to be entered again. The
  fix belongs with those other two.
- **Storing or replacing a credential writes no audit entry.** Editing git sync settings does;
  the credential route should too. Four lines, in a follow-up.
- **Saving git sync settings asks the host once per repository** for its credential and its hook,
  in the request path. A workspace with many repositories on a slow host feels it. The maintenance
  pass converges either way.
- **A fork keeps its recorded credential error** after the workspace holding the token replaces
  it, until the fork's next settings save or the next maintenance pass, at most 40 minutes. The
  fork syncs with the new token throughout; only its managed features wait. Clearing it at once
  would need a walk down the fork tree on every credential write.
- **The resource drawer keeps the URL it opened with** after a save that changes it, so the
  Replace token control stays disabled until the drawer is reopened. Cosmetic; the drawer's
  fetched-resource cache is not refreshed on save, which predates this change.
- **Repointing a repository's `url` leaves the old webhook and the old credential status in
  place.** Both are reconciled on a git-sync settings save, not on a resource edit. The old project
  keeps its hook and the new one gets none until the settings are saved again; polling covers the
  gap. The recorded status is keyed by the resource path while the credential is keyed by the
  repository, so when the new repository also has a token stored in the workspace it is judged by
  the old one's status until that save or the next maintenance pass, at most 40 minutes: an
  MR-on-deploy in that window is skipped, or opened before the new token's first check. With one
  stored token per workspace a repoint resolves to no credential and no status. Binding the status
  to the repository identity would touch the status struct, the OpenAPI schema and the client;
  refreshing on resource edit would put an EE hook in the store crate. A settings save after a
  repoint is the instruction either way.

## Test plan

- [ ] Instant pull: push to the tracked branch, confirm the webhook enqueues a pull and the change lands.
- [ ] Polling fallback: delete the hook, confirm the poller picks the push up.
- [ ] Forged and absent webhook tokens are rejected with no job enqueued.
- [ ] Deploy in promotion mode opens an MR carrying the diff in a note.
- [ ] Fork sync at depth 1 and depth 2 both pull into the right workspace.
- [ ] Credential renewal inside the window rewrites the stored token and sync continues.
- [ ] Repointing a resource's `url` at another repository resolves to no credential.
- [ ] Removing a repository from git sync settings leaves its resource authenticating.


hugocasa:
Answering the maintenance-pass scalability question from the review, since it needs the EE side to settle:

- A repo whose URL carries no credential (every GitHub App repo) costs one variable read and no HTTP — `gitlab_api_base` / `url_credential` return `None` before any request.
- The pass no longer writes on every visit. `persist_repo_credential_status` compares against the stored status (ignoring `checked_at`) and skips the update when nothing changed, so an unchanged repo — including every repo with nothing to track — costs no `workspace_settings` write at all. That was a real problem in the reviewed head: it rewrote the shared `git_sync` blob every 40 minutes per repo to record nothing, racing other writers.
- Both GitLab calls go through `gitlab_client()`, which sets a 10s connect and 20s total timeout, so one unreachable host cannot stall the pass indefinitely.

There is still no provider filter in the query and the loop is sequential; at the scale of git-sync repos per instance that is a bounded cost with the write and HTTP removed, and adding a per-repo throttle like `AUTO_PULL_LAST_POLL` is easy if it ever needs one.

hugocasa:
On the Codex P1 "project access tokens are renewed through the wrong endpoint": this was verified live rather than from the docs, on GitLab CE 19.3.1 (project token and group token) and on gitlab.com (project token on `wmill-group/test-sync`). `POST /api/v4/personal_access_tokens/self/rotate` returns **200** for all three, with a new token id and the requested `expires_at`; the old token 401s afterwards and the new one keeps its project membership. A project or group access token *is* a personal access token held by the bot user GitLab mints for it, which is why the personal-token self routes accept it. The dedicated `/projects/:id/access_tokens/self/rotate` route is the documented one but not required. Recorded in `gitlab_rotate_self`'s doc comment; not changing.

On the webhook body size: the GitLab receiver now authenticates the secret before parsing the body, so an unauthenticated request pays for the decrypts and nothing else. The body limit itself stays the instance default, as it does for the GitHub receiver.

The stale-status finding and Claude's nit are addressed on `d879dc5c6d`: the managed-feature gates now also require a held or borrowed credential, so a repoint drops them immediately; and a fork checks the licence before reading `borrowed` as renewed. A token without `api` is not refused up front: it fails the first deploy and the hook registration loudly on the parent, which is enough.

hugocasa:
On the two Codex P1s from the `ddae66ba9f` round, addressed on `c91da8fdf3`:

- **Replacing a failed token** now re-checks every git-sync repository naming that repository in the same request, so the recorded error clears at once. Verified live: planted an error, stored a fresh token, status healthy in the response's own write.
- **Maintenance ordering**: the budget, the least-recently-checked ordering and the six-hour rewrite throttle are gone rather than fixed. The budget existed only to keep the pass under the monitor join's deadline; the pass now runs as its own task, so the deadline no longer applies and the advisory lock alone prevents overlap, the same shape as the auto-pull poller. A status now records every check.

The drawer keeping its opened URL after a save is listed under known limitations; the export comment is fixed on `3d115bca00`.

hugocasa:
On the `c91da8fdf3` round:

- **Codex P1, forks keeping a stale error after the holder replaces its token**: not fixing. The fork syncs with the new token immediately; only its managed features wait for its next settings save or the maintenance pass, at most 40 minutes. Clearing it at once would need a walk down the fork tree on every credential write. Listed under known limitations.
- **Claude P2, the spawn comment**: reworded on `0e4653b043` to say what detaching buys (a deadline a long sweep would hit) and what remains (process shutdown, which a rotation almost never coincides with). Not restructuring for it.
- `cargo check --features quickjs,private,parquet` passes against the current EE ref.

hugocasa:
Re the Codex findings on `35b7ed1`:

**P1 (repoint reuses the old status)** — acknowledged, and deliberately left as a documented limitation rather than fixed here. It only arises when a resource is repointed from one repository to another that *also* has a token stored in the same workspace; with one stored token per workspace, the usual case, a repoint resolves to no credential and no status (pinned by `a_fork_reads_an_ancestors_credential_for_the_bound_repository_only`). The window closes on the next git-sync settings save or the maintenance pass (≤40 min), and the deploy push itself works with the new repository's token throughout. It is the same edit, trigger and recovery as the "repoint leaves the old webhook in place" limitation already listed, so that bullet now covers the status too. The two fixes on the table (repository identity stored with the status and compared on read; status refresh from the generic resource-update handler) both widen the PR's surface for an edit that already needs a settings save to get its webhook.

**P2 (stale card after Replace token)** — dismissed: the repository card does not open the resource drawer. Replace token lives only in the Resources page drawer, and returning to the git-sync tab remounts `GitSyncSection`, which calls `loadSettings()` again, so the card shows the refreshed status. The only stale view is a second browser tab, which a reload fixes. Wiring `onReplaced` in `ResourceEditor` would have nothing to call, as the editor holds no git-sync state.

hugocasa:
Re the Codex findings on `72caffaf3d`:

**P1 (two-connection pools)** — fixed in `543da5bec2` / ee `a96c47c`. The per-repository advisory lock was redundant: `rotate_git_credential_if_due` has one caller, the maintenance pass, and the pass is already serialized across servers by the instance-wide lock, while a manual replacement racing a rotation is caught by the `FOR UPDATE` read and `expect_unchanged_token` at the write-back. It is removed, so the pass retains one connection like the neighbouring monitor tasks, and repositories are checked sequentially on top of that.

**P1 (credentialed repoint)** — same finding as the previous round; covered by the "Repointing a repository's `url`" limitation in the description, with the reasoning for not binding the status to the repository identity in this PR.

**P2 (stale card after Replace)** — same as the previous round: the repository card never opens the resource drawer, and returning to the git-sync tab remounts `GitSyncSection`, which reloads settings. `ResourceEditor` holds no git-sync state an `onReplaced` callback could refresh.

**P2 (picker beyond 50 projects)** — by design. The docs and the helper text under the token field recommend one project access token per repository, which reaches exactly one project. A group token covering a large group can still be stored for any project: create the resource with the URL, then use Replace token in the drawer, which checks the token against that project by name.

hugocasa:
Re the Codex nits on `72a0268b35`:

**P2 (picker admin check follows the navigated workspace)** — same terms as the GitHub App button beside it, which gates on `$userStore` the same way (`GitHubAppIntegration.svelte:74`); the picker was written to match it so the two read as one choice. Fixing it for GitLab alone would leave the pair inconsistent, and in both cases a mismatch ends in a 403 from the target workspace rather than a wrong write. Left as is.

**P2 (callback docs describe a deferred write)** — fixed in the docs commit above: both pass-through docs now say what `GitLabIntegration` actually does, store first, then fire the callback so the form stops filing the URL as a secret.

Also in this push: the maintenance pass holds its instance-wide lock as a transaction-scoped advisory lock, the pattern `reconcile_unarmed_schedules` uses, so a sweep that dies before unlocking cannot strand the lock in the pool.

hugocasa:
Re the Codex P1 on `e8938fc18b` (lock expires mid-sweep): correct, and fixed in the commit above. The lock transaction now runs `SET LOCAL idle_in_transaction_session_timeout = 0` before taking the advisory lock, so the pool's ten-minute idle timeout does not end it while the sweep is out talking to git hosts. `SET LOCAL` is transaction-scoped: verified against Postgres that the setting reads `0` inside the transaction and the pool's `10min` again after rollback, so the connection goes back to the pool with its normal timeout, and a dropped transaction still releases the lock.

PR #10942, opened 2026-09-02
feat: restore owner and label filter chips on the homepage

rubenfiszel:
## Summary
The home search/filter revamp (#10020, shipped in v1.800.0) moved the homepage's owner chips (`u/<user>` / `f/<folder>`) and label chips into the FilterSearchbar's preset dropdown, so scoping the list went from one visible click to opening the searchbar and picking a preset. This brings the on-page chips back as a single row under the header, wired to the same URL-synced `owner` / `label` filter keys the searchbar uses, so a chip click, the searchbar tag and the query parameter are one state.

## Changes
- `ItemsList.svelte`: render owner and label chips on one line with two `ListFilters` instances in `inline` mode inside a shared flex-wrap row. Each is bound to its searchbar key through a Svelte 5 function binding (`bind:selectedFilter={() => ownerFilter, setOwnerFilter}` and the label equivalent). `queryName` points ListFilters' mount-time URL read at the same `?owner=` / `?label=` param the filter instance syncs; no `syncQuery`, the filter instance owns the URL.
- Each group caps at 10 chips with the existing "…" overflow toggle, the selected chip always visible. Owners keep their existing order (your own space first, then most populated). Labels are now ranked by how many loaded rows carry them, ties alphabetical, instead of plain alphabetical.
- Clearing a chip deletes the key rather than writing `null`, so no `owner: null` / `label: null` tag lingers in the searchbar. A URL-supplied value no loaded row carries still gets a chip (ListFilters' selected-prepend), so it can be cleared from the page.
- `ListFilters.svelte`: optional `color`, `icon` and `inline` props. `inline` renders the chips as `display: contents` so several instances can share one row; the defaults leave the 15 other call sites unchanged. The label group passes `color="blue" icon={Tag}` for the pre-revamp look.
- The row hides in content-search mode. The searchbar owner/label presets stay, so both entry points keep working.

## Test plan
- [x] `npm run check:fast` and prettier pass for both files; Svelte autofixer reports no issues.
- [x] With 12 folders and skewed labels (billing on 5 scripts, ops on 3, one tag each): one row shows `u/admin`, nine folders and "…", then `billing`, `ops`, `tag01`…`tag08` and "…". Expanding shows every chip, collapsing returns to 10.
- [x] Clicking `billing` narrows to the five billing scripts, the URL gains `label=billing`, the searchbar shows `label: billing`; clicking again clears it.
- [x] Clicking a folder chip narrows the list to that folder, the URL gains `owner=f%2F…`, the searchbar shows the matching tag; clicking the selected chip again restores all rows and drops the param.
- [x] Loading `?owner=u%2Fadmin&label=ops` directly selects both chips on mount and shows only the matching rows; browser Back restores URL, chips and searchbar tags together.
- [x] Loading `?label=zzz` (no such label) shows a `zzz ✗` chip alongside the real ones, with an empty list.
- [ ] Reviewer: on a workspace with labeled runnables in a few folders, toggle chips and clear from the searchbar side; confirm chips and tags stay in sync both ways.

## Screenshots
Default state: owners then labels on one wrapped row, each group capped at 10 with its overflow toggle:

home-chips-oneline

`billing` selected: list narrowed to the five billing scripts, chip shows the clear mark, searchbar carries the matching tag:

home-chips-oneline-billing

https://claude.ai/code/session_01AbM6X8fYEQWKMTjUyp6aqy

### DIFF
--- a/frontend/src/lib/components/home/ItemsList.svelte
+++ b/frontend/src/lib/components/home/ItemsList.svelte
@@ -40,6 +40,7 @@
 		type FilterSchemaRec
 	} from '$lib/components/FilterSearchbar.svelte'
 	import NoItemFound from './NoItemFound.svelte'
+	import ListFilters from './ListFilters.svelte'
 	import ToggleButtonGroup from '../common/toggleButton-v2/ToggleButtonGroup.svelte'
 	import ToggleButton from '../common/toggleButton-v2/ToggleButton.svelte'
 	import FlowIcon from './FlowIcon.svelte'
@@ -76,8 +77,8 @@
 
 	// FilterSearchbar schema — `_default_` is the free-text search; the rest mirror the
 	// boolean/kind list filters. Owner and label scoping are offered as searchbar presets
-	// (searchPresets); label is also a row of on-page chips (labelChips) writing the same
-	// key. `content` is a distinct mode: it swaps the list for the client-side
+	// (searchPresets) and as rows of on-page chips (ListFilters / labelChips) writing the
+	// same keys. `content` is a distinct mode: it swaps the list for the client-side
 	// content-match view below (usable on any instance, not EE-gated).
 	let searchFilterSchema = $derived({
 		_default_: { type: 'string' as const, hidden: true },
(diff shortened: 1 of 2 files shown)


rubenfiszel:
✅ Review round clean @ e646fabac8bd4c4f37f43b8fff637525614ddc82 — nit-only verdicts at a33130de2ca31720909dbef1ff385e128cad6b64; nits addressed in e646fabac8 (dropped the no-op `small` prop, fixed the stale `labelChips` comment, consolidated the shared-state rationale into the markup comment)

PR #10943, opened 2026-09-02
fix: register the job token with the sensitive log masking system

hugocasa:
## Summary

The log masking system covered secrets fetched through `get_value_internal` and `$encrypted:` args, but not the job's own token. A script that echoed `$WM_TOKEN` wrote it verbatim into logs that are persisted to the database and, when configured, to object storage — and the token stays usable well past the run that produced it.

`handle_queued_job` now registers the token with the masking registry for the job it is about to run, so it is redacted like any other registered secret. NativeTS needed a second fix: it runs V8 in-process and delivers `console.log` output through its own channel, so it never reached the masking in `handle_child::write_lines` and the registration alone was a no-op for it.

Masking stays a log concern. The raw value still reaches the job's result, which is deliberate — that is a data channel a later flow step may consume, not display.

## Changes

- `worker.rs`: hold the masking lifecycle at the job boundary. `RunningJobMasks` registers the job and its token on entry to `handle_queued_job` and unregisters on drop, so the poller, the interactive worker shell and inline AI agent tools are all covered by construction — the first draft registered around the poller's call site only, and the other two paths logged the token unmasked. Unregistering earlier than the old call site costs nothing: the writes that follow go through `append_logs`, which never consulted the registry.
- `sensitive_log_masks.rs`: cache the compiled Aho-Corasick automaton per job, invalidated when a new secret is registered. Every job now carries at least one registered value, where before the per-batch snapshot was skipped entirely for the majority of jobs that touched no secret — without the cache a chatty job would rebuild the automaton once per log batch (~90µs and ~126KB each, measured).
- `sensitive_log_masks.rs`: add `JobMasker`, for a sink that masks line by line rather than in batches. A bare `snapshot` per line would re-arm the security notice on every one; this keeps it to once per distinct set of secrets while still picking up secrets registered mid-run, and falls back to the masks it last saw for a sink still emitting after the job is unregistered.
- `windmill-runtime-nativets/src/lib.rs`: mask each chunk as V8 produces it, on the producing side of the log channel. That loop is joined before the job completes, so the job's secrets are always still registered — unlike the detached task that drains into `append_logs`, which outlives the job and would write a late secret raw. Masking there also covers the `windmill:job_log` tracing emission that EE forwards job logs on, which a channel-side mask missed.

  Stream/log routing is decided before masking, so a secret-bearing `WM_STREAM:` chunk cannot spend the one-shot notice on text that is then discarded. A stream chunk itself stays raw, because `merge_result_stream` can make it the job's result; this deliberately differs from `handle_child`, which extracts its stream from the masked text.
- Tests: `backend/tests/job_token_log_masking.rs` (bash, plus a nativets case gated on `deno_core` the way the CI test build is), and unit tests pinning the automaton cache invalidation, the notice cadence, and the post-unregistration fallback.

## Known remaining sinks

Two log paths still persist an unmasked token. Both predate this change, neither is reachable from a plain `cargo run`, and both are noted here rather than fixed so the scope stays honest:

- **Dedicated workers / runner groups.** Dispatch returns before `handle_queued_job` is ever called, and their logs go through `append_logs` with no mask. The credential there is not the per-job `WM_TOKEN` at all but a worker-lifetime token minted with `super_admin = true`, so it does not fit the per-job mask registry and needs its own fix in an EE file. Worth its own ticket ahead of this one on severity.
- **`docker_v2`.** Streamed container logs are appended unmasked. The token only reaches them if the user's own script passes it into the container.

The durable answer for those is to centralize masking at the `append_logs` funnel rather than adding it sink by sink. That is deliberately not in this PR: it would change the log output of every executor and demands re-verifying each, and it does not close the dedicated-worker gap at all, since that credential never enters the per-job registry.

## Test plan

- [x] Bash job logging `--token $WM_TOKEN` shows `jwt*****xxx` plus the masking notice in the UI and in the persisted `job_logs` row; same for a flow step.
- [x] NativeTS job logging `process.env.WM_TOKEN` is masked, verified on a real `deno_core` build run as a tag-scoped worker so the quickjs worker could not race for the job; A/B confirms the full JWT was persisted without the fix.
- [x] NativeTS job that emits a secret-bearing `WM_STREAM:` chunk *and* logs the token: the log line is masked and still carries the notice, confirming the stream chunk no longer swallows it. (The job then fails on `invalid input syntax for type json` when a non-JSON stream becomes the result — reproduced identically with no secret involved, so it is pre-existing and unrelated.)
- [x] A job that echoes the token, fetches a secret variable mid-run, then echoes both: token masked in the first batch, token and variable both masked after — confirms the automaton cache invalidates correctly.
- [x] `cargo test --features deno_core --test job_token_log_masking` passes both cases; each fails with the raw token in the log when its own masking is removed.
- [x] `cargo test -p windmill-common --lib sensitive_log_masks` passes; each test fails when the invariant it pins is broken.
- [x] `cargo test --test worker bash` (the three pre-existing bash job tests) still passes.

Left in draft: touches shared worker infrastructure and a token path, so it wants a human look before ready.

https://claude.ai/code/session_01P316wKe2QCYNcdsx1PwAJ3

### DIFF
--- a/backend/tests/job_token_log_masking.rs
+++ b/backend/tests/job_token_log_masking.rs
@@ -62,8 +62,7 @@ async fn test_job_token_masked_in_persisted_logs(db: Pool<Postgres>) -> anyhow::
 
 /// nativets runs V8 in-process and persists `console.log` output through its own
 /// channel, so it is masked by a different mechanism than the bash case above and
-/// needs its own guard. Tagged `deno` because a standalone nativets job defaults to
-/// the `nativets` tag, which the default worker tag set does not advertise.
+/// needs its own guard.
 #[cfg(feature = "deno_core")]
 #[sqlx::test(fixtures("base"))]
 async fn test_job_token_masked_in_nativets_logs(db: Pool<Postgres>) -> anyhow::Result<()> {
@@ -84,7 +83,7 @@ async fn test_job_token_masked_in_nativets_logs(db: Pool<Postgres>) -> anyhow::R
             .into(),
         debouncing_settings: windmill_common::runnable_settings::DebouncingSettings::default(),
         modules: None,
-        tag: Some("deno".to_string()),
+        tag: None,
     }))
     .run_until_complete(&db, false, port)
     .await;
--- a/backend/tests/job_token_log_masking.rs
+++ b/backend/tests/job_token_log_masking.rs
@@ -1,0 +1,61 @@
+/*
+ * The job's own token (`$WM_TOKEN`) stays valid well past the job it was minted
+ * for, and job logs are persisted to `job_logs` and optionally to object storage,
+ * so a script that echoes the token would otherwise park a live credential in
+ * durable storage. `run_worker` registers the token with `sensitive_log_masks`
+ * for the job it pulled; this pins that the persisted log carries the masked form.
+ */
+
+use sqlx::{Pool, Postgres};
+use windmill_common::{
+    jobs::{JobPayload, RawCode},
+    scripts::ScriptLang,
+};
+use windmill_test_utils::*;
+
+/// Prefix of a serialized job token: `jwt_` plus the base64 of a JWT header.
+/// The masked form keeps only `jwt` + the last three characters, so it never matches.
+const RAW_TOKEN_PREFIX: &str = "jwt_ey";
+
+#[sqlx::test(fixtures("base"))]
+async fn test_job_token_masked_in_persisted_logs(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+    let server = ApiServer::start(db.clone()).await?;
+    let port = server.addr.port();
+
+    let job = RunJob::from(JobPayload::Code(RawCode {
+        hash: None,
+        content: "echo \"running with --token $WM_TOKEN\"".to_string(),
+        path: None,
+        lock: None,
+        language: ScriptLang::Bash,
+        cache_ttl: None,
+        cache_ignore_s3_path: None,
+        dedicated_worker: None,
+        concurrency_settings: windmill_common::runnable_settings::ConcurrencySettings::default()
+            .into(),
+        debouncing_settings: windmill_common::runnable_settings::DebouncingSettings::default(),
+        modules: None,
+        tag: None,
+    }))
+    .run_until_complete(&db, false, port)
+    .await;
+    assert!(job.success, "job should have succeeded");
+
+    let logs =
+        sqlx::query_scalar::<_, Option<String>>("SELECT logs FROM job_logs WHERE job_id = $1")
+            .bind(job.id)
+            .fetch_one(&db)
+            .await?
+            .unwrap_or_default();
+
+    assert!(
+        !logs.contains(RAW_TOKEN_PREFIX),
+        "an unmasked job token reached the persisted logs: {logs}"
+    );
+    assert!(
+        logs.contains("secret value was masked"),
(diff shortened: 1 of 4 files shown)


hugocasa:
Both nits from the last round are addressed in `d13fb9f` (comment-only).

**The `WM_STREAM:` chunk reaching `windmill:job_log` raw** is deliberate, and the comment now says why rather than leaving it to be re-derived. A stream chunk is result data, not a log line: it never reaches `job_logs`, and `merge_result_stream` can make it the job's result. Masking it would be masking a result, which is the one thing this PR consistently does not do — the job result carries the raw token too, for the reason in the description. So it stays raw wherever it goes, the log sink and the mirror alike. This is unchanged from `main`, where that sink received every nativets chunk raw.

**The masker test's rationale** no longer names a mechanism. I could not actually confirm that the memory-limit path leaves the producing loop running past unregistration — `handle` is not awaited there, but the inner current-thread runtime is dropped when the enclosing future completes, which would cancel the task rather than let it run on. The retention is defensive, so the test now states the invariant it pins and nothing more, and `lib.rs` says "normally" instead of "always".

hugocasa:
Round status on `d13fb9f`: Claude is **Good to merge**. The Codex CI job did not post for this head (its workflow shows `skipped`), so I ran the same reviewer locally — same policy, model and reasoning effort — and it returns one P1: the dedicated-worker path.

**Declining that one, deliberately.** It is the first entry in "Known remaining sinks" above, and it is not this fix wearing off — it is a different credential. `handle_queued_job` is never entered for a dedicated job (dispatch `continue`s at `worker.rs:3621`), but more to the point, `create_worker_token` mints a **worker-lifetime** token shared by every job that worker runs, with `super_admin = true`. `SENSITIVE_MASKS` is keyed by job id, so there is nothing to register it against; covering it needs a new "mask for the life of this worker" concept plus a companion `windmill-ee-private` PR, because `dedicated_worker_ee.rs` is a symlink into that repo. That is its own ticket, and on severity it outranks this one — a long-lived superadmin token in persisted logs is worse than the ephemeral job token this PR closes.

Two independent Claude passes reached the same conclusion on the same evidence, so I am leaving it rather than growing this diff into the EE repo. Happy to be overruled if you would rather hold this until the dedicated-worker fix lands alongside it.

**CI is red for unrelated reasons.** `Backend check` and `Backend only integration tests` fail on `spurious network error ... exit status: 128` fetching git dependencies (`sqlx`, `tiberius`, `object_store`, `php-parser-rs`, `duckdb`). I re-ran both and they failed the same way; `main` and every other branch pushed in the last hour fail identically, so it is a GitHub-side outage rather than anything in this branch. Locally `cargo test --features deno_core --test job_token_log_masking` and `cargo test -p windmill-common --lib sensitive_log_masks` both pass.

Staying a draft on purpose (see the last line of the description).

hugocasa:
CI update, correcting my last comment: **`Backend only integration tests` is green on a re-run.** That is the workflow that builds with `deno_core` and runs the suite, so both new tests in `backend/tests/job_token_log_masking.rs` — including the nativets case — passed in CI, not just locally.

`Backend check` is still red, but all three of its jobs (`check_oss`, `check_oss_full`, `check_ee`) die at `failed to load source for dependency postgres-protocol`, i.e. during dependency resolution, before a single crate is compiled. Two re-runs, same failure on a rotating set of git dependencies. The green integration run is the counter-evidence that matters: it resolved the same dependency graph and compiled the same code, so there is nothing here for a code change to fix.

PR #10945, opened 2026-09-02
feat: offer known Google scopes as checkboxes in the oauth connect dialog

alpetric:
## Summary
Users connecting a Google integration on a shared instance type extra scopes as free text (for example `gmail.modify`) and then hit Google's "This app is blocked" page when the instance's OAuth client is not verified for that scope. This adds an advisory `scope_options` list to the seven Google registry entries and renders it as checkboxes in the connect dialog, so users pick from the scopes the instance's client is meant to carry instead of guessing scope URLs. Free-text entry stays for self-hosters with custom clients.

The `gforms` entry keeps its legacy `https://www.googleapis.com/auth/forms` default: accounts connected before scopes were stored per account refresh with the registry default, so changing it could make Google reject their refresh. The Forms API v1 scopes (`forms.body`, `forms.body.readonly`, `forms.responses.readonly`) are offered as options instead, alongside the legacy one, so new connections can pick the correct set.

## Changes
- `backend/oauth_connect.json`: `scope_options` on gsheets, gdrive, gmail, gcal, gforms, gcloud, gworkspace (defaults plus least-privilege/read-only and, for Gmail, readonly/compose/modify/labels). Defaults are unchanged.
- `backend/windmill-oauth/src/lib.rs`: a comment on `OAuthConfig::scopes` noting the JSON-only `scope_options` key. The key is deliberately not modelled in Rust: the backend never reads it, serde ignores it, and modelling it would touch every exhaustive `OAuthConfig` literal including the EE Slack client.
- `frontend/src/lib/components/OauthScopes.svelte`: optional `options` prop rendered as `Checkbox` rows above the free-text list. Ticked options and free-text rows are tracked as their own state, so a row can be typed through an option's exact value without ticking or unticking anything, ticking an option absorbs a row holding the same value, and the counter counts free-text rows only (hidden when there are none). The toggle takes its target from that state rather than the input's `checked`, because `Checkbox` re-asserts its `checked` prop on every click and the input already reads the old value again by the time `change` fires.
- `frontend/src/lib/components/AppConnectInner.svelte`: passes `registryEntry()?.scope_options`.

Companion: windmill-labs/windmill-integrations#182 adds the key to the registry validator type.

## Test plan
- [x] `svelte-check`: 0 errors; registry validator (`validate-oauth-connect.ts`) passes for gmail/gforms/gdrive
- [x] Exercised on a local instance (`cargo run --features private,oauth2` + vite dev, Gmail tile configured) with Playwright: defaults pre-ticked (`gmail.send`), ticking `gmail.modify` and unticking/re-ticking `gmail.send` work, a custom row typed through the exact `gmail.labels` value leaves that box unticked and the row survives, and the read-only list after closing the editor shows the ticked options plus the custom row. The first browser run caught the `Checkbox` re-assert issue described above; the jsdom harness had not.
- [ ] Complete a real Google connection with a ticked scope and confirm the authorize URL carries it (needs a client verified for that scope; pending the verification this PR supports)

## Screenshots
Connect dialog for Gmail with the option checkboxes (defaults pre-ticked):

connect-gmail-2-checkboxes

`gmail.modify` ticked, plus a custom row:

connect-gmail-3-modify-ticked

Read-only view after closing the editor:

connect-gmail-4-readonly-after

https://claude.ai/code/session_01Kv72vdCDggCZEjJSmNCwnX

### DIFF
--- a/backend/oauth_connect.json
+++ b/backend/oauth_connect.json
@@ -86,11 +86,9 @@
   "gforms": {
     "auth_url": "https://accounts.google.com/o/oauth2/v2/auth",
     "token_url": "https://oauth2.googleapis.com/token",
-    "scopes": [
-      "https://www.googleapis.com/auth/forms.body",
-      "https://www.googleapis.com/auth/forms.responses.readonly"
-    ],
+    "scopes": ["https://www.googleapis.com/auth/forms"],
     "scope_options": [
+      "https://www.googleapis.com/auth/forms",
       "https://www.googleapis.com/auth/forms.body",
       "https://www.googleapis.com/auth/forms.body.readonly",
       "https://www.googleapis.com/auth/forms.responses.readonly"
--- a/backend/oauth_connect.json
+++ b/backend/oauth_connect.json
@@ -29,6 +29,10 @@
     "auth_url": "https://accounts.google.com/o/oauth2/v2/auth",
     "token_url": "https://oauth2.googleapis.com/token",
     "scopes": ["https://www.googleapis.com/auth/spreadsheets"],
+    "scope_options": [
+      "https://www.googleapis.com/auth/spreadsheets",
+      "https://www.googleapis.com/auth/spreadsheets.readonly"
+    ],
     "extra_params": {
       "access_type": "offline",
       "prompt": "consent"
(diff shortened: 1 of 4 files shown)


alpetric:
✅ Review round clean @ 9a2147f5657241558d237c41601e37313fe992a1 — nit-only verdicts at 2cb791ea3ee1ee4f04070e6337faa33b07fea428; nits addressed in 9a2147f565 (gforms default kept, counter hidden when empty); Claude 'Good to merge' at 9a2147f565. Codex never ran on this draft (auto-review skips drafts; /codex command runs failed on a GitHub git error).

PR #10949, opened 2026-09-02
feat: add retention cleanup for the otel_traces table

hugocasa:
## Summary

The `otel_traces` table backing the HTTP request tracing proxy grew unbounded. Its migration
created a time index on `start_time_unix_nano` for retention, but no cleanup ever ran, so an
instance with tracing enabled accumulated one row per HTTP request made from a job script,
forever. Every comparable table (jobs, audit logs, service logs) already has a retention window
and a sweep in the monitor loop; this gives spans one too.

The window is a new `otel_traces_retention_secs` instance setting, defaulting to 7 days, swept in
`delete_expired_items` alongside the neighbouring retention sweeps.

It is deliberately independent of the job retention period rather than derived from it: the two
answer different questions (how long a job's record is kept vs. how long its per-request
telemetry is worth storing), and spans are far more numerous than the jobs they hang off, so an
operator usually wants to expire them sooner. The consequence, stated in the setting's help text,
is that a span can outlive its job or be swept while the job remains.

## Upgrade note

**The first monitor tick after upgrade deletes every span older than 7 days.** The table was
previously unbounded, so an instance that has had HTTP Request Tracing on for a while will lose
request traces it could previously still show in the job detail view, with no way to recover
them. That is the point of the change and the default is the same order as the neighbouring
retention windows, but it is a one-way trim worth a line in the release notes. Operators who want
a longer window should set `otel_traces_retention_secs` before upgrading.

## Changes

- `otel_traces_retention_secs` global setting, read at startup and reloaded live on change.
- Batched sweep in `delete_expired_items`, following the existing service-log sweep: 10 batches
  of 10k rows per pass, so a retention change that expires millions of rows drains across ticks
  rather than inside one 600s-bounded monitor pass.
- The batch predicate is `ctid`, not the `(trace_id, span_id)` primary key. `EXPLAIN ANALYZE` on
  a 300k-row table showed the key form plans a Hash Semi Join whose outer side is a **Seq Scan of
  the whole table** once rows actually qualify, which is exactly the cost the batching exists to
  avoid; `ctid` plans as a Tid Scan and touches only the rows it deletes. In steady state the
  subquery uses `otel_traces_time_idx` and finds nothing for ~3 buffer hits.
- `otel_traces` added to the hourly `VACUUM` list, next to `log_file` — same bulk-delete churn
  profile as the sibling sweep.
- Extracted `clamp_retention_secs` in windmill-common, now shared with the pre-existing
  `set_service_log_retention_secs` rather than duplicating its 15-line clamp and the reasoning
  behind it. Behaviour for service logs is unchanged, and its existing test still passes.
- Setting exposed on the OTEL/Prom instance-settings tab, EE-badged.

### Why the EE badge needed a new `ceMaxSeconds` field

Both reviewers asked for `ee_only` on the new setting, and it belongs there — only the EE proxy
captures spans. But on a `seconds` field `ee_only` was doing double duty: it also imposed a
30-day ceiling on a build without a license. That ceiling is real and backend-enforced for
`retention_period_secs` (`CE_MAX_RETENTION_PERIOD_SECS`), and meaningless everywhere else it
happened to apply.

Inheriting it here would have been wrong twice over: a CE build still sweeps spans an EE-era
instance left behind, and `clamp_retention_secs` accepts the same 100-year range on either
edition, so the UI would have silently contradicted the backend. So the cap is now an explicit
`ceMaxSeconds`, set only on `retention_period_secs`, and the badge is just a badge. Verified on a
real CE build (license removed): job retention still clamps 90 days to 30, trace retention
accepts 90 days.

## Test plan

- [x] `cargo test --bin windmill otel_traces_retention` — DB-backed test pinning that only spans
      past the window are deleted. Mutation-checked: changing the nanosecond multiplier makes it
      fail.
- [x] `cargo test -p windmill-common --lib retention` — the pre-existing service-log clamp test
      still passes after the shared-helper extraction.
- [x] Live on a running instance: planted spans at known ages, set the window to 3600s through
      `POST /api/settings/global/otel_traces_retention_secs`, and confirmed the monitor deleted
      exactly the spans past it and logged `deleted 1234 expired otel trace spans`.
- [x] Live through the UI: the field renders on the OTEL/Prometheus tab, edits to 2 days show
      correctly in the Review-changes diff, save persists, and after restart the sweep used the
      saved 2-day window (3-day-old spans gone, 1-day-old kept).
- [x] Clearing the field falls back to the 7-day default live, with no restart (9-day-old spans
      swept, 3-day-old kept).
- [x] CE build (license temporarily removed): EE badge renders, `retention_period_secs` still
      clamps to 30 days, `otel_traces_retention_secs` accepts 90 days.
- [x] `EXPLAIN (ANALYZE, BUFFERS)` on a 300k-row table, both with and without expired rows
      present.

### Not exercised

Spans were planted with SQL rather than captured through the EE tracing proxy, which needs an
EE build and a job making HTTP requests. The writer is unchanged EE code; the column's unit
(nanoseconds since epoch) is confirmed from three independent places — the proto/migration
comment, `store_span` in `otel_tracing_proxy_ee.rs`, and the frontend's
`new Date(ns / 1_000_000)`.

## Screenshots

Deliberately omitted. The only UI change is one field on the instance-settings page, and this
repo's PR guidance forbids uploading screenshots of instance settings to the public screenshot
host (the same page renders `jwt_secret` and `license_key`). The field is a standard `seconds`
input labelled "HTTP Request Tracing retention in secs" with an EE badge, on the OTEL/Prom tab
directly below "HTTP Request Tracing".

https://claude.ai/code/session_01NLhUaCPpLRAa29rSZDjS28


hugocasa:
✅ Review round clean @ eb6af5f4735847267e9055278c98025a88d29224

PR #10951, opened 2026-09-02
fix(cli): make a sync push into a fork converge on schedules and inline names

rubenfiszel:
## Summary

A git-sync **Pull from repo** into a fork workspace (the pull job runs `wmill sync push` from the fork branch into the fork) reported `All 19 changes pushed`, and the next pull listed the same 19 rows. Nothing was dropped this time (unlike #10473): the content did land, but each row was a difference the push can never make go away, so the list is permanent and reads like a silent no-op. Two of the four shapes are the CLI's to fix:

1. **Schedules in a fork.** A fork clones every schedule disabled, the backend refuses to enable one whose path the parent also has (`fork-conflict:schedule:<parent>`), and for those paths the fork's tarball export writes the *parent's* `enabled` into the file (#9476). Between #8976 and #9476 the fork export omitted `enabled` entirely, so a fork branch can carry schedule files with no `enabled` or a stale one. On a push into the fork the differ then reports every such file as edited forever (`isSuperset` considers the push up to date), and when the file says `enabled: true` against a disabled fork copy the push calls `setScheduleEnabled`, hits the refusal, and aborts the whole pull job.
2. **Inline-script names.** `ZipFSElement` names a flow's rendered inline-script files from the step summary. A checkout that names the file differently (the flow.yaml `!inline` reference still resolves) shows a delete + `flow.yaml` edit + add on every push while the resolved flows compare equal, so `pushFlow` reports "up to date" and the three rows never clear. The existing "preserve original `!inline` filenames" `extractCurrentMapping(...)` call ran on the *remote* modules, whose content is script source and never an `!inline` reference, so it was always empty.

The other two shapes are on the checkout side and unchanged: a flow step lock file missing from git (the server relocks the flow after every push, so the `.lock` "deleted" row returns) and a stray file inside a `.flow/` folder.

## Changes

- `cli/src/commands/sync/sync.ts`: `push()` resolves `parentOwnedScheduleEnabled(workspaceId)`: `undefined` unless the target is a fork (`listUserWorkspaces` parent id, or the `wm-fork-` id prefix); otherwise a predicate saying whether a schedule file's `enabled` belongs to the parent, built from `listSchedules` on the parent. When the parent cannot be listed (a job token is scoped to the fork), every schedule is treated as the parent's, so nothing is toggled and nothing aborts. A fork-only schedule keeps toggling from the file.
- `cli/src/commands/sync/sync.ts`: `compareDynFSElement` takes that predicate and compares matching schedule files without `enabled` (YAML and JSON).
- `cli/src/commands/sync/sync.ts`: `ZipFSElement` takes an optional `localFlowInlineMapping(flowDir)`; `push()` feeds it the checkout's `flow.yaml` `!inline` references (module id → file, via `extractCurrentMapping`), so the remote render uses the same file names as git. `pull()` is untouched: a rename on pull converges after one pull. The dead remote-side `extractCurrentMapping` call is removed. Both functions are now exported for the unit test.
- `cli/src/commands/schedule/schedule.ts`, `cli/src/types.ts`: `pushSchedule(..., enabledOwnedByParent)` (threaded through `PushObjOptions`) drops `enabled` from the update when the fork already has the row, which also skips the `setScheduleEnabled` call. Creating a schedule the fork does not have keeps the file's value. The stale comment saying the fork export "strips" `enabled` is replaced by the one constraint that guard now serves.
- `cli/test/push_diff_convergence_unit.test.ts`: a schedule the parent has compares without `enabled` (differing or absent), still reports other field changes, and a fork-only schedule still reports an `enabled` change; a remote flow rendered with the checkout's names produces no rows for identical content and a single `edited` for a real content change.

- Review follow-ups (368cad5bad): the parent lookup runs only when `includeSchedules` is on, after the options merge; a `fork-conflict:schedule` answer to `setScheduleEnabled` is logged and skipped instead of failing the push (the parent listing is RLS-scoped, the backend's check is not); a checkout `!inline` name equal to another step's summary-derived name makes that flow render the export's way, so two steps never share one path.

- Hardening (f092836443): when a fork schedule's flag is set aside while the file disagrees, or the backend refuses an enable with `fork-conflict`, the CLI says so on stderr through a new `log.warnAlways` that survives `--json-output` (the git-sync job runs with that flag, and `log.warn` is silenced after startup there). `checkoutInlineNames` ignores a reference that leaves the flow folder (`..` segments or an absolute path), so a remote step can never be rendered onto another item's path.
- b4dcf0490c: a create into a fork lands disabled on the backend whatever the request says, so a fork-only schedule the file wants enabled is enabled right after creation (one push converges) while a parent-owned one is created disabled with a notice. A `wm-fork-*` workspace whose parent was deleted keeps its prefix but nothing defers to a parent any more: when the listing says it has no parent it is handled as a normal workspace, so its schedules toggle from git again.
- Known limit, unchanged behavior: a fork-scoped job token cannot call `listUserWorkspaces` or the parent's `listSchedules` (both 401), so in the git-sync job a throwaway fork is detected by its `wm-fork-` prefix and every schedule is treated as the parent's (fork-only schedules are not toggled from git there, with a notice). A dev workspace (custom id with a parent) is not detected on that path and keeps today's behavior; closing that needs the parent id on a workspace-scoped endpoint, a backend follow-up.

## Test plan

- [x] `UNIT_ONLY=1 bun test test/push_diff_convergence_unit.test.ts` passes; full `bun run test:unit`: 1102 pass, 0 fail
- [x] `bun run check`: the same 18 pre-existing errors as `main`, none in the touched files beyond the pre-existing `JSZip | TarAsZip` ones at the two `ZipFSElement` call sites
- [x] Manual, against a local CE backend with a parent workspace, a fork created via `create_fork`, and a checkout pulled from the parent (dotted layout, `includeSchedules`): with the fork's step renamed and the checkout keeping the old file name, `wmill sync push --workspace <fork>` lists one `edited` content row instead of the delete/edit/add triple, applies the content, and a second run lists nothing for that flow
- [x] Manual: parent schedule disabled while the checkout says `enabled: true` (previously aborted the push with `fork-conflict:schedule`), and a checkout schedule file with no `enabled` key against an enabled parent (previously "edited" on every run): the push succeeds, the fork's copies stay disabled, and the second dry run lists neither
- [x] Manual: a schedule that exists only in the fork, `enabled: true` in the checkout, is enabled by the push
- [x] Manual: a push into a non-fork workspace still applies `enabled` from the file; `sync pull` unchanged
- [ ] Shipping needs the usual chain: publish `windmill-cli`, republish hub script 13952 (git-sync init) with the new pin, bump `GIT_SYNC_PULL_SCRIPT_PATH` and `gitInitRepo` (auto-pull runs the same pinned script)

https://claude.ai/code/session_01JcofduXAs9FT948Aj78V8m

---
## Summary by cubic
Fixes WIN-2478: previously, `wmill sync push` into a fork could repeatedly report schedule and inline-script filename changes or abort on a schedule conflict. It now preserves parent-owned schedule flags, enables fork-only schedules on create, and renders inline scripts with safe checkout filenames so repeated pushes converge.

- Parent schedule ownership comes from workspace and schedule listings when available; fork-scoped jobs fall back to `wm-fork-` detection, while forks without a parent treat schedules as local.
- `fork-conflict:schedule` responses no longer fail the push, and warnings remain visible with `--json-output`.
- Inline references outside the flow folder are ignored, and colliding checkout names fall back to unique export-derived names.
- Non-fork pushes and `sync pull` keep their existing behavior; unit tests cover schedule changes, inline matching, collisions, path traversal, and OS-specific paths.

**Release**

- Publish `windmill-cli`, republish hub script 13952 with the new pin, and bump `GIT_SYNC_PULL_SCRIPT_PATH` and `gitInitRepo`.

<sup>Written for commit b4dcf0490c817f0a64e2ec0892db184aaa463972. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10951?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes WIN-2478

### DIFF
--- a/cli/src/commands/schedule/schedule.ts
+++ b/cli/src/commands/schedule/schedule.ts
@@ -106,7 +106,8 @@ export async function pushSchedule(
   path: string,
   schedule: Schedule | ScheduleFile | undefined,
   localSchedule: ScheduleFile,
-  permissionedAsContext?: PermissionedAsContext
+  permissionedAsContext?: PermissionedAsContext,
+  enabledOwnedByParent?: boolean
 ): Promise<void> {
   path = removeType(path, "schedule").replaceAll(SEP, "/");
   log.debug(`Processing local schedule ${path}`);
@@ -122,6 +123,14 @@ export async function pushSchedule(
 
   // Strip CLI-only boolean marker before sending to API
   delete (localSchedule as any).has_permissioned_as;
+
+  // In a fork, the file's `enabled` is the parent's for a path the parent
+  // also has (see sync push's `parentOwnedScheduleEnabled`): the fork's own
+  // flag stays as it is. A schedule the fork does not have yet is created as
+  // the file says, there being no fork state to keep.
+  if (enabledOwnedByParent && schedule) {
+    delete localSchedule.enabled;
+  }
 
   const preserveFields: { permissioned_as?: string; preserve_permissioned_as?: boolean } = {};
   if (permissionedAsContext?.userIsAdminOrDeployer) {
@@ -153,13 +162,9 @@ export async function pushSchedule(
           ...preserveFields,
         },
       });
-      // Tarball export from a fork strips `enabled` from schedule YAMLs so
-      // the fork→parent git-sync round-trip can't flip the parent's state.
-      // Skip the secondary setScheduleEnabled call when the local YAML
-      // doesn't carry `enabled` — sending `{ enabled: undefined }` would
-      // serialize to `{}` and the backend (`SetEnabled.enabled` is required)
-      // would reject the request. Preserving the target's existing flag is
-      // exactly the round-trip-safe behavior.
+      // No `enabled` in the file (absent from the YAML, or set aside above)
+      // leaves the remote flag alone: `SetEnabled.enabled` is required, so
+      // `{ enabled: undefined }` would be rejected rather than ignored.
       if (
         localSchedule.enabled !== undefined &&
         localSchedule.enabled !== schedule.enabled
--- a/cli/src/commands/schedule/schedule.ts
+++ b/cli/src/commands/schedule/schedule.ts
@@ -129,6 +129,14 @@ export async function pushSchedule(
   // flag stays as it is. A schedule the fork does not have yet is created as
   // the file says, there being no fork state to keep.
   if (enabledOwnedByParent && schedule) {
+    if (
+      localSchedule.enabled !== undefined &&
+      localSchedule.enabled !== schedule.enabled
+    ) {
+      log.warnAlways(
+        `Schedule ${path} stays ${schedule.enabled ? "enabled" : "disabled"}: the file says ${localSchedule.enabled ? "enabled" : "disabled"}, but in a fork that flag is the parent workspace's`
+      );
+    }
     delete localSchedule.enabled;
   }
 
@@ -189,7 +197,7 @@ export async function pushSchedule(
           if (!conflict) {
             throw e;
           }
-          log.warn(
+          log.warnAlways(
             `Schedule ${path} left ${schedule.enabled ? "enabled" : "disabled"}: the parent workspace '${conflict.parentWorkspaceId}' has the same schedule, so its flag is the parent's to set`
           );
         }
--- a/cli/src/commands/schedule/schedule.ts
+++ b/cli/src/commands/schedule/schedule.ts
@@ -126,8 +126,7 @@ export async function pushSchedule(
 
   // In a fork, the file's `enabled` is the parent's for a path the parent
   // also has (see sync push's `parentOwnedScheduleEnabled`): the fork's own
-  // flag stays as it is. A schedule the fork does not have yet is created as
-  // the file says, there being no fork state to keep.
+  // flag stays as it is.
   if (enabledOwnedByParent && schedule) {
     if (
       localSchedule.enabled !== undefined &&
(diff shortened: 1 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 4d54ce892330d11660d7af5235a9529a3053d88e — nit-only verdict at 0becd05e57 (Claude: gate the parent lookup on includeSchedules, tolerate fork-conflict on enable, keep rendered names unique), nits addressed in 368cad5bad, then Claude: Good to merge at 368cad5bad; 4d54ce8923 only makes the new test fixtures use the OS path separator (test-windows was red on them). Codex CI never posted a verdict on any head: its job dies at `git fetch --no-tags origin main +refs/pull/N/head` with `fatal: could not read Username for 'https://github.com'` (runner git-access fault, same as #10931 today), so `/codex` cannot help. Substituted the same review locally (`local-review-codex`: REVIEW.md policy, gpt-5.6-sol, xhigh) on this head: **Good to merge**, no findings. CI on this head: CLI tests (linux + windows) and git_sync_e2e green.

rubenfiszel:
Post-ready hardening, cold-reviewed locally since the CI Codex/Pi runners still cannot fetch from github.com: f092836443 (stderr notices that survive `--json-output`, checkout inline names kept inside the flow folder), b4dcf0490c (fork-only schedule enabled on create so one push converges; a fork with no parent on record owns its schedules again). Cold Claude review (branch-diff-reviewer vs origin/main): mergeable with nits, all three addressed. Cold Codex review (`local-review-codex` vs origin/main) raised one P1 with two halves: the orphaned-fork half is fixed in b4dcf0490c; the dev-workspace half is the documented limit above (a fork-scoped job token cannot learn its parent from any workspace-scoped route, so a custom-id dev workspace keeps today's behavior on the git-sync job path until the backend exposes `parent_workspace_id` there). CI on each head: CLI tests linux+windows and git_sync_e2e green.

PR #10954, opened 2026-09-03
feat: guest JWT entry for embedded apps

rubenfiszel:
## Summary

Adds a second way in for a guest (windmill#10929): a JWT the embedding customer's own backend mints and signs, carried on the app's share link and verified against a key the workspace admin configured. It needs no identity-provider round-trip, so it works inside an iframe where popups and third-party cookies do not. Bearer prefix `jwt_guest_`, stateless: verified per request, cached until `exp`, no `token` row.

It is deliberately narrower than the existing external JWT scheme (`jwt_ext_`, instance-level `JWT_EXT_PUBLIC_KEY` / `JWT_EXT_JWKS_URL`, whose claims can assert admin, groups and folders). A guest key can only ever mint guests, whatever the token says.

> Rebased onto the parent's new guest policy (`license-app-only-implicit-promotion`): the Enterprise-plan gate was replaced by a free allowance (100 emails / 30 days) with metering on Enterprise and a hard cap elsewhere, plus a superadmin instance switch. The JWT arm now calls `guest_admission` transactionally and the door re-reads `guest_session_stands`; the plan gate on the key config is gone.

## Who a JWT guest is

The same identity a signed-in guest gets in windmill#10929: someone with no Windmill account, no `usr` row and no `password` row anywhere, so no seat on any counter, confined to the one app the token names. Every gate a signed-in guest passes applies unchanged:

- **The instance allowance** (`guest_admission`): guests are free up to 100 distinct emails over a trailing 30 days; past that an Enterprise plan meters them (four to a seat) and every other plan or build refuses a new email. A returning guest is always let back in. The JWT arm calls it inside the transaction that records `guest_activity`, so the advisory lock spans the count check and the row that changes it. There is also a superadmin instance switch (`guest_access_disabled`), folded into `guest_app_admits`/`guest_session_stands`.
- **The workspace switch** (`guest_access_enabled`): enforced once at the auth door for every guest request, JWT or not, so turning guests off closes the JWT entry on the next call.
- **The app mode**: the app named by `app_path` must be in `guest` execution mode (`guest_app_admits`).
- **No account at all**: an email that already has a `password` or `usr` row is refused, not downgraded to a guest (the same `has_any_account` rule as the signed-in mint).

The resolved identity is the same shape as the DB guest arm: `username = email`, operator, no groups, no folders, `scopes = guest_session_scopes(app_path)` plus the `guest` sentinel. `whoami` reports `role: guest`.

## What a token must carry

Claims honoured, and nothing else: `email`, `workspace_id`, `app_path`, `exp` (mandatory). `nbf` and `iat` are validated when present, and the accepted lifetime is capped at 24h so a long-lived token cannot be minted by mistake. Extra claims are ignored. `app_path` is mandatory: the token opens that one app. The claim's `workspace_id` must equal the route's workspace, and a workspace-less route never accepts the token.

Algorithms: RS256/384/512, PS256/384/512, ES256/384. HS\* (a symmetric shared secret) is refused: a PEM public key has no symmetric form, and the JWKS path rejects it explicitly.

## Why the key is per workspace

The verification key is a per-workspace setting, workspace-admin owned (guests are free up to the instance allowance on any plan, so configuring their key needs no licence): a PEM public key, or a JWKS URL whose keys are fetched and refreshed the way `ExternalJwks` does it. A `kid` selects the JWKS key; a PEM key ignores `kid`. New `workspace_settings.guest_jwt_public_key` / `guest_jwt_jwks_url` columns (at most one non-null, a DB CHECK enforces it). The JWKS URL is validated against private ranges and the connect is pinned to the validated address, so a workspace admin cannot turn it into an SSRF.

This is why it is narrower than `jwt_ext_`: that key is instance-level and its claims can assert admin, groups and folders; a guest key is scoped to one workspace and can only ever produce a guest.

## The counting consequence

A JWT guest is recorded in `guest_activity` (once per email, workspace and day, cached so it is not one upsert per request) and **not** in `unique_ext_jwt_token`. A new `jwt_entry` column marks the row, and the EE telemetry reports a `guest_jwt_count` next to `guest_count` so the two entries can be told apart; `seats_before` keeps counting both. An audit `users.login_guest` is written the first time a token is seen, with the entry kind (`jwt`) in the parameters.

The policy this encodes, taken deliberately: an app-only user a customer routes through `jwt_ext_` today is counted in `unique_ext_jwt_token_count`. The same user through a guest JWT becomes a free guest. That is the intended pricing change, and the `guest_jwt_count` split is what lets it be measured rather than guessed.

## Derived tokens

A token minted *by* a JWT guest (the embed token for a sandboxed low-code app, the SDK token for a sandboxed raw app) is capped at the minting credential's expiry, since that expiry is a guest's only revocation. A signed-in guest session is a `token` row, and `guest_derived_token_constraints` looks the expiry up there. A JWT has no row, so its `exp` is carried through the auth cache on a new `ApiAuthed.credential_expiry` field and the mint caps on that. It never falls back to a fresh interval.

## Share link

`/public/<workspace>/<secret>/<jwt>` already puts a trailing segment into `OpenAPI.TOKEN` as `jwt_ext_<jwt>` on the embedder side only (the JWT must never reach the opaque iframe, where app-authored code could read it). A guest JWT rides a distinct URL form, `/public/<workspace>/<secret>/guest.<jwt>`, recognised by the `guest.` prefix (a `.` can't appear in a secret or path segment, so an existing ext-JWT link whose custom path ends in a `guest` segment is untouched) and sent as `jwt_guest_<jwt>`. The ext-JWT behaviour is byte-for-byte unchanged; the viewer iframe URL uses the bare secret, so no JWT reaches it either way.

The custom-path route (`/a/<path>`) carries an ext JWT today, so it gets the same treatment: `/a/<path>/guest.<jwt>`.

## Changes on the parent branch

The feature needed a few additive changes to windmill#10929's code, made as normal commits here rather than rewriting the parent:

- `ApiAuthed.credential_expiry`, a credential's own expiry when it has no `token` row (a JWT guest). `None` everywhere else.
- `guest_derived_token_constraints` prefers `credential_expiry` when set, and falls back to the existing prefix lookup for a signed-in guest session.
- `guest_session_scopes` moved from `windmill-api-users` to `windmill-api-auth::scopes` so the mint and the JWT auth arm share one copy.
- `has_any_account` factored into `windmill-common::users`, replacing the inline check in the signed-in mint and reused by the JWT arm.
- The signed-in guest mint's `users.login_guest` audit now carries `entry=idp`, the counterpart of the JWT arm's `entry=jwt`.

## Test plan

Integration tests in `backend/tests/app_guest_jwt_entry.rs` (`#![cfg(all(feature = "enterprise", feature = "private"))]`, same fixtures as `app_guest_execution_mode.rs`), run with `--features quickjs,enterprise,private`. Fixed EC P-256 test keys; tokens signed in-test with `jsonwebtoken`:

- [x] a valid token opens its app, runs a component as the publisher, reads the run back, reports `role: guest`, and leaves exactly one `guest_activity` row across several requests
- [x] refused: wrong workspace, wrong key, expired (past the leeway), HS256, an email with an account, an app not in guest mode
- [x] the workspace switch gates it at the door: off refuses, on admits, off again closes it
- [x] a workspace-less route refuses it
- [x] a sandboxed app's derived embed token expiry is at or before the JWT's

Exercised the JWKS path live too (backend run with `ALLOW_PRIVATE_GUEST_JWKS_URLS` so a localhost stub could stand in for the issuer): configured a JWKS URL, a `kid`-carrying token opens the app as a guest, no-kid and unknown-kid tokens are refused, and with the issuer stopped a burst of requests is served from the positive cache in ~0.03s with no outbound fan-out.

Exercised on a running EE instance (`--features quickjs,private,enterprise,oauth2`): configured a PEM key on a workspace through the settings UI, minted a token with `openssl`, opened the share link signed out, confirmed the app renders on behalf of the publisher with `whoami` reporting `guest`, no `password`/`usr` rows and one `guest_activity` row; the same with `sandbox: true`, whose derived embed token is capped at the JWT's `exp`.

### Embed section: guest JWT as the primary embed method

The app deploy drawer, for a guest-mode app, now explains embedding with a guest JWT: mint a `jwt_guest_` token in your backend and append `guest.<jwt>` to the app URL, with a copyable iframe snippet pre-filled with this app's `workspace_id` and `app_path`. It names both key sources (the workspace's guest JWT key, or the instance issuer `JWT_EXT_JWKS_URL` off cloud) and the accepted algorithms.

guest-jwt-embed

### Workspace settings: the Guests card and the guest JWT key

A PEM public key or a JWKS URL (which must be https), with the accepted claims and algorithms, and the "leave empty to fall back to the instance issuer (`JWT_EXT_JWKS_URL`)" note — hidden on cloud.

guest-jwt-settings

### A guest opening the share link signed out

The page renders on behalf of the publisher, the header shows the JWT's email as the guest identity, and it runs `on behalf of admin@windmill.dev`. No `password` or `usr` row is written.

guest-jwt-render

_Audit note: this dev instance does not persist audit rows (0 rows for every operation), so the `users.login_guest` write is exercised but not verified in the DB, same as the signed-in mint on the parent branch._


rubenfiszel:
Addressed round-1 findings in 77b24b20cd: lowercased email for the account gate (and `has_any_account`), capped the auth-cache entry for a guest JWT at 5 min so a rotated key revokes and the second-day activity row is written, DB-atomic `xmax` audit dedup, JWKS body size cap + alg-family selection for alg-less RSA keys + negative caching so an unreachable issuer is not fanned out, and the issuer caveat in the settings copy. Added unit tests for `jwk_algorithms` and a JWK-derived key verifying a real token, plus mixed-case and over-24h refusal cases. Also drove the JWKS path live: a `kid` token opens the app as a guest, no-kid/unknown-kid are refused, and with the issuer down a burst of requests is served from cache in ~0.03s with no outbound fan-out.

rubenfiszel:
Round 2 fixes in 98e8a594f2: JWKS refresh is now single-flight (a per-URL lock, so a burst against a cold/stale entry triggers one fetch, not one per request); a remembered empty (negative) entry now reports 'issuer unreachable' rather than 'kid not in the JWKS'; the PEM field uses the design-system `TextInput` textarea variant instead of a raw `<textarea>`; and `has_any_account` no longer sits between `username_to_permissioned_as`'s doc comment and its body. Verified live: a concurrent burst of 10 against a cold JWKS cache returns 200 in ~0.04s.

rubenfiszel:
Round 3 fixes (388768f8c6, 6b3dedbe49, 629ac4ba92):
- **Codex P1 (email claim):** `verify()` now requires a plain, bounded email address (an `@`, no `/`, no whitespace, ≤254 bytes). This closes both the principal-confusion angle (a `group-` prefix or missing `@` would map to a `g/`/`u/` principal via `username_to_permissioned_as`) and the over-255-byte overflow you found (the `guest_activity` insert failed while the guest was still admitted uncounted). Refused before authentication; tested with group-shaped, slash, and oversized emails.
- **Claude nit (stale-while-revalidate):** a stale-but-good JWKS entry is now served immediately while a refresh runs off the request path, so a slow or hanging issuer no longer stalls guests at the 15-minute TTL boundary. Only a cold or negative entry blocks, still single-flight.
- **Test coverage:** added `a_concurrent_cold_burst_makes_one_jwks_fetch`, a stub issuer that counts inbound connections while ten concurrent `cached_jwks` calls run, asserting exactly one fetch. This pins the single-flight property the last rounds were about.

rubenfiszel:
Merged the parent's new guest policy (`license-app-only-implicit-promotion` @ f626dc312f) and re-integrated:
- The Enterprise-plan gate is gone. Guests are free up to 100 emails / 30 days, then metered on Enterprise and hard-capped elsewhere; a superadmin `guest_access_disabled` switch is folded into `guest_app_admits`/`guest_session_stands`.
- The JWT arm now calls `guest_admission` **inside the transaction that records `guest_activity`** (the advisory lock spans the count check and the row), and the door re-reads `guest_session_stands` (switch, instance switch, no account) per request, so the arm needs nothing extra there.
- The plan gate on `edit_guest_jwt_key` and the `isEnterprisePlan` gate on the settings key config are removed (guests are free on any plan); the footer now saves a key-only change.
- New test `a_guest_jwt_is_capped_like_a_signed_in_guest` in `app_guest_allowance.rs` (its own binary, which flips the plan): with the window full on a capped instance, a stranger's JWT is refused (401) and a returning guest's is admitted.

Also carries the earlier round-4 fixes (all preserved through the merge): `VALID_EMAIL` reuse + 254-byte bound, the `app_path` scope-metacharacter guard, base64url `isJwt`, JWKS single-flight + stale-while-revalidate. EE companion updated to b011775 (guest_jwt_count merged alongside the parent's guest_seats). All 8 unit + 7 integration tests pass; backend and frontend compile clean.

Note: the JWT path refuses a capped stranger with a 401 (the visitor-facing sentence is logged, since the auth arm returns `Option` and cannot carry a message); the signed-in path returns the message, pinned by the parent's allowance test.

rubenfiszel:
Round 5 fixes (b8838564ff):
- **Audit-aborted tx (P1):** `admit_and_record_guest_jwt` now commits the `guest_activity` row first, then audits best-effort on a separate connection. The EE audit writer swallows an `audit_partitioned` failure but the failing statement aborts its transaction, so auditing before the commit could roll the activity row back while the arm returned success. Now the activity is durable before the audit runs.
- **Custom-path `guest` collision (P1):** the share-link marker is now the prefix `guest.` glued to the token (`/a/<path>/guest.<jwt>`, `/public/<ws>/<secret>/guest.<jwt>`). A `.` can't appear in a custom-path or secret segment, so an external-JWT link whose path ends in a `guest` segment (`/a/foo/guest/<jwt>`) is read as `foo/guest` + `jwt_ext_` as before. Removed the unused `SHARE_LINK_SEGMENT` const.
- **Allowance test race (P2):** moved the JWT allowance case to its own binary `app_guest_jwt_allowance.rs`, since `set_plan` flips a process-global license key that a shared binary would race on under `--test-threads`.

Three test binaries pass (app_guest_allowance 1, app_guest_jwt_allowance 1, app_guest_jwt_entry 5); frontend and backend compile clean.

rubenfiszel:
Round 6 nits addressed (4899e51703):
- **IdP-then-JWT audit (Codex P2):** the JWT `users.login_guest` audit is now gated on the `jwt_entry` false→true transition, decided atomically by a CTE in the upsert (`RETURNING (NOT COALESCE((SELECT jwt_entry FROM prior), false))`), so it fires once a day even when an IdP sign-in created today's row first.
- **has_any_account (Codex/Pi/Claude):** the signed-in mint now calls the shared `has_any_account` helper too, so the two copies no longer diverge on lowercasing and the PR body's claim is now accurate.
- **Refused-JWT advisory lock (Claude):** a refused guest JWT is negative-cached 30s, so a replayed bearer past the cap no longer takes the instance-wide allowance lock every request.
- Fixed the two stale `guest/<jwt>` header comments (→ `guest.<jwt>`), the entry-test cfg rationale, the blank lines from the const removal, and prettier-formatted the settings card.

sqlx cache regenerated for the CTE (offline check green); 8 unit + 7 integration tests pass across three binaries.

rubenfiszel:
Round 7 nits addressed (e2c6e659ea):
- **Audit atomicity (Codex/Claude):** the audit gate is a conditional upsert `DO UPDATE ... WHERE NOT guest_activity.jwt_entry RETURNING 1` read with `fetch_optional` — the row comes back exactly once per email per day, decided by the conflicting tuple, so it keeps `xmax = 0`'s atomicity (no double-audit when two first requests race on a metered instance) and still audits the first JWT after an IdP row exists. Replaces the snapshot-based CTE.
- **Negative cache (Claude):** only a real `PermissionDenied` allowance refusal is cached; a transient DB error denies the one request but no longer locks the email out for 30s.
- **JWKS key_ops (Codex):** a key whose `key_ops` omits `verify` is refused (unit-tested).

sqlx cache regenerated (offline green); 9 unit + 6 integration tests pass.

rubenfiszel:
Round 8 fix (last Codex finding): `fetch_jwks` now parses each JWK on its own via `serde_json::Value` and skips any that does not model as a jsonwebtoken signing key, so a set carrying an encryption key (`alg: "RSA-OAEP"`) alongside a signing key no longer fails whole-set deserialization. Unit-tested with a mixed RSA-OAEP + P-256 set (only the signing `kid` survives). Claude and Pi were already good to merge.

rubenfiszel:
✅ Review round clean @ 1845a9c10fe043b38ab6f97e2b490dfc87ca2775 — Codex, Claude and Pi all "Good to merge".

rubenfiszel:
Merged the parent again (`license-app-only-implicit-promotion` @ 2b4631cedc, which merged main). Re-integrated the two items from the latest parent review:
- **Canonical `app_path`:** `guest_session_scopes` now returns `Result` and calls `windmill_common::utils::check_proper_path` first (the path is spliced into a scope whose parser reads `,` as a separator and `*` as a wildcard). My moved copy in `windmill_api_auth::scopes` carries that guard, and the JWT arm's `verify()` refuses a non-canonical `app_path` claim with the same check (replacing the earlier ad-hoc metacharacter test).
- **require_job_read_access:** a JWT guest gets the launched-by-me job-read stop automatically once the sentinel is set — nothing extra needed.

EE companion updated to 5827618 (my guest_jwt_count + credential_expiry over the new parent EE). CE+EE compile clean; 10 unit + 6 integration tests pass.

rubenfiszel:
Round-10 findings addressed and parent re-merged (`license-app-only-implicit-promotion` @ 0241ba52b4).

**Codex P1 (CE settings can skip guest JWT key updates):** `saveDefaultAppSettings` now saves the guest JWT key right after guest access, ahead of the Enterprise-only default-app and rate-limit writes, so a refused write can no longer swallow a valid key change on CE.

**Claude nits:** the `guest_jwt_refusals` suite now opens with a positive control (a token valid against its exact fixture is admitted 200), so a broken setup fails loudly instead of passing the eleven 401s vacuously; `enable_guests` asserts its status like the other setup helpers; `verify()`'s doc now names the email and app_path rules.

**Pi nit:** the Guests card summary now names the JWT entry alongside the IdP path.

**Parent correction (app_path guard):** switched from `check_proper_path` to `windmill_common::auth::is_scope_literal_path` in both the relocated `guest_session_scopes` and the JWT arm's `verify()` — app paths may carry spaces and `@`, so only the scope-reserved `:`, `,`, `*` and a leading `/` are refused. `require_job_read_access` needs nothing extra.

CE+EE compile clean; 10 unit + 6 integration tests pass (the wildcard-`app_path` refusal now trips the new predicate).

rubenfiszel:
Round-11 addressed.

**Codex P1 (a private RSA PEM could be stored as a verification key):** `decoding_key_from_pem` now refuses any PEM containing `PRIVATE KEY` before it is parsed or persisted. It is the single choke point for both `edit_guest_jwt_key` (save time) and per-request verification, so a private key can neither be stored nor served back through the settings response. New unit test `a_private_pem_is_refused` pins it with a real EC private key (which the parser would otherwise accept) and an RSA-private label.

**Claude P2 nit (stale `canonical`):** the JWT arm's comment and refusal log no longer say "canonical" — the log now reads `guest JWT app_path cannot be scoped`, matching the reserved-character guard. The relocated `guest_session_scopes` doc regained the sentinel/label distinction for both callers.

11 unit + 5 entry tests pass; the valid PEM path (PUB1) still saves and authenticates. The wider auth.rs diff is rustfmt bringing adjacent lines into compliance.

rubenfiszel:
Round-13 addressed (codex/cubic P1: label-only guard bypassable).

The label check was insufficient: jsonwebtoken 8.3 decides public vs private from the PEM label alone and never inspects the DER, so private DER relabelled `PUBLIC KEY` (or a PKCS#1 key under `RSA PUBLIC KEY`) would still be stored and served back through the settings response.

`decoding_key_from_pem` now parses the decoded DER as a `spki::SubjectPublicKeyInfo` before the key is persisted or used — a structure only a public key satisfies, so private-key DER is refused whatever its armor says. This is the single choke point for both `edit_guest_jwt_key` and per-request verification. Added `spki` (with the `pem` feature) as a windmill-common dep; `Cargo.lock` gains one edge, no version churn.

The regression test now uses **complete, valid** private keys — the same EC PKCS#8 key relabelled `PUBLIC KEY`, and a full RSA PKCS#1 key relabelled `RSA PUBLIC KEY` — so the SPKI parse is what refuses them, not malformed DER (the flaw in the previous test). A real RSA public key is asserted `ok` as a positive control. 11 unit + 5 entry tests pass; the valid EC public key (PUB1) still saves and authenticates.

rubenfiszel:
Round-14 nits addressed (claude, pi, cubic P2 — all the same PKCS#1/wrapping compatibility point; codex was already Good to merge).

The SPKI guard decoded the body with the strict RFC 7468 decoder (64-column wrapping) and accepted only SPKI, so a legitimately-formatted public key wrapped otherwise, or a PKCS#1 `RSA PUBLIC KEY`, was refused where jsonwebtoken would have parsed it.

`decoding_key_from_pem` now decodes the body leniently (as jsonwebtoken does, tolerating any wrapping) and then requires the DER to be a public-key structure — an SPKI (RSA or EC) **or** a PKCS#1 RSA public key. Private material satisfies neither, so the round-13 relabelled-private bypass stays closed. Added `pkcs1` as a windmill-common dep. Test adds a one-line (non-64-column) public key as a positive control; the three private-key rejections still hold. 11 unit + 5 entry tests pass.

rubenfiszel:
Round-15 addressed (codex P2 JWKS material; claude stale-comment nits; pi was Good to merge).

**Codex P2 (malformed JWKS accepted at config time):** `parse_jwks_keys` filtered on metadata only, but jsonwebtoken carries `n`/`e`/`x`/`y` as strings and defers decoding to auth time. It now keeps a key only if `DecodingKey::from_jwk` actually decodes it, so a JWKS whose only key is malformed is refused at save time (the empty set errors) instead of persisting a URL whose tokens all fail later. It is the single source for both `edit_guest_jwt_key` and per-request verify. New unit test `a_jwks_with_only_malformed_key_material_is_refused`; the mixed-use test now carries real coordinates so its signing key survives the material check.

**Claude nits:** the two test comments that credited the SPKI parse alone now name the full guard (SPKI or PKCS#1 RSA public).

12 unit + 5 entry tests pass.

rubenfiszel:
On codex's remaining P2 (base64url-valid but structurally-impossible JWK material) — a deliberate boundary rather than a fix:

The round-15 change keys config-time validation to `DecodingKey::from_jwk`, which is **exactly the decode the auth path already performs** at parse time (`jwks_key_for`). So config-time and auth-time now validate identically — that is what closes the original 'save succeeds, every token fails later' gap for malformed material.

jsonwebtoken 8.3 never checks on-curve-ness or modulus validity at parse; it surfaces those at **signature verification** (ring for EC, the rsa path for RSA). A base64url-valid but cryptographically-invalid key therefore cannot silently 'work' — it fails loudly at the first token's verification, on a key the workspace admin controls and is testing.

Making config-time validation stricter than the auth parse itself would mean constructing full EC/RSA keys at save time (p256/p384 for on-curve, and `rsa` — which is only an *optional* windmill-common dependency, so not available on the OSS-compiled `guest_jwt.rs` path — for modulus checks). That is inconsistent with where jsonwebtoken itself draws the line, for marginal value given the loud first-token failure. I'm keeping the boundary at decode-level validation, matching the auth path.

pi and claude both verified the round-15 fix closes the gap and are Good to merge.

rubenfiszel:
Merged the parent again (`license-app-only-implicit-promotion` @ 6169076d70; EE unchanged at 6bfe1e25).

The parent adds guest_activity handling on workspace **rename** (`UPDATE guest_activity SET workspace_id ...`) and **delete** (`DELETE FROM guest_activity ...`), and reads the deployed mode under the app-row lock on a rename. The rename/delete queries touch only `workspace_id`, so the `jwt_entry` column this PR adds rides along untouched. Conflict resolved as before: `guest_session_scopes` stays relocated to `windmill_api_auth::scopes` (its refusal message synced with the parent's); the parent's re-added local copy is dropped.

Production build compiles clean under `enterprise,private`; 12 unit + 5 entry + 1 allowance guest tests pass.

Note: a full `cargo check --tests` surfaces a pre-existing EE unit-test bug unrelated to this PR — `src/ee.rs:1382` calls `seats_consumed(c.author_count, c.operator_count)` (two args) while the function has taken `&ActiveSeatCounts` since the guest-seat feature; every production call site already uses `seats_consumed(&seats)`. It is present verbatim on this PR's EE base (6bfe1e25), so it is not in this PR's diff.

rubenfiszel:
Addressed the round P1 (workspace rename silently dropping the guest JWT key) and merged the EE OAuth-cookie change.

**P1 fix:** `change_workspace_id` copies `workspace_settings` by an explicit column list that omitted `guest_jwt_public_key` and `guest_jwt_jwks_url` — added both to the copy (and updated the `.sqlx` cache). The parent's rename test now also asserts the guest JWT key travels, so the silent drop has a guard. Synced `verify()`'s refusal message with the other two sites (leading `/` clause).

**EE:** merged the parent's OAuth guest-cookie change (`<state>|<ws>/<path>`, bound to the flow state via `guest_target_for_state` / `clear_guest_target_for_state` in `oauth2_ee.rs`); it is self-contained in that EE file and CE compiles against it unchanged. ee-repo-ref bumped to the merged EE head. The CE parent branch is unchanged at 6169076d70 (the change was EE-only).

12 unit + 5 entry + 1 allowance + the rename test pass. (The unrelated pre-existing EE `seats_consumed` arity bug in `src/ee.rs:1382` is still on the EE base, not in this diff.)

rubenfiszel:
Merged the CE parent (`license-app-only-implicit-promotion` @ 3776040dd6 — its only change was bumping ee-repo-ref for the EE-only OAuth-cookie change; kept this branch's ee-repo-ref at the EE companion head c96012d8, my guest_jwt EE changes on top of that same EE parent).

Also addressed cubic's P3: the app_path refusal messages (verify() and guest_session_scopes) now name the empty-path case that is_scope_literal_path also rejects.

Prior round's P1 (workspace rename dropping the guest JWT key) is fixed in 71122d53cd. 12 unit + 5 entry + 1 allowance + rename tests pass; CE compiles against the new EE.

rubenfiszel:
Addressed the JWKS-cache retention finding (codex P1 / pi P2 / claude P2) and merged the latest parent (OSS 4434d3d94b, EE 00bf1dde9c).

**JWKS cache bound:** `parse_jwks_keys` now refuses a set larger than `JWKS_MAX_KEYS = 50`. The byte cap (1 MiB/response) and entry-count cap (200) left keys-per-entry unbounded — a densely packed body parses into thousands of keys, so a workspace admin rotating ~200 URLs could retain hundreds of MB. Capping the count at the source bounds retention (≤ 50 keys × 200 entries); real issuers publish a handful of keys, so it is invisible to any legitimate config. New unit test `an_oversized_jwks_is_refused`.

**Merge:** the OSS parent change was an ee-repo-ref bump; the EE change was in `duckdb_isolation_ee.rs` (unrelated). CE compiles against the new EE (dc5cba65). ee-repo-ref bumped accordingly. The prior round's empty-path message nit (cubic P3) is also in.

13 unit + 5 entry + 1 allowance + rename tests pass.

rubenfiszel:
Reworked the JWKS bound to close codex P1 and cubic P2 / claude nit together.

The key-count cap had two problems the round named: it did not bound retained **memory** (`from_jwk` decodes `n`/`e`/`x`/`y` with no length limit, so 50 keys could still carry ~1 MiB → the ~200 MB ceiling codex named), and it counted **raw entries**, so a valid mixed-use set (many encryption keys, few signing) was refused wholesale.

`parse_jwks_keys` now caps the **serialized size of the usable, retained keys** at `JWKS_MAX_RETAINED_BYTES = 64 KiB`, measured **after** filtering. So the bound is on what a cache entry actually retains (cache ceiling 200 × 64 KiB ≈ 12.8 MiB), and a large mixed-use set is judged by its signing keys, not its raw entry count. A real set retains a few KB. Test replaced: `a_jwks_retaining_too_many_bytes_is_refused` uses 600 real-coordinate keys so it exercises the bytes path, not the empty-set path.

13 unit + 5 entry + 1 allowance + rename tests pass.

rubenfiszel:
Bounded every dimension of the JWKS cache (closes codex P1 on the URL, claude P2 on per-key overhead, and cubic P2 on the raw-entry basis), and merged the latest parent (OSS 6081e286c9, EE 32841072aa).

`parse_jwks_keys` now caps, all measured **after** filtering to usable keys:
- **key count** `JWKS_MAX_KEYS = 50` — bounds the per-`Jwk` fixed cost (claude's many-tiny-keys case);
- **retained bytes** `JWKS_MAX_RETAINED_BYTES = 64 KiB` — bounds material (`from_jwk` decodes `n`/`e`/`x`/`y` with no length limit);

and `fetch_jwks` caps the **URL length** `MAX_JWKS_URL_LEN = 2048` — the caches key on the URL string (unbounded `TEXT` column), and fetch_jwks is the choke point `edit_guest_jwt_key` validates through, so an overlong URL is never stored or cached (codex's P1). Both caps are on usable keys so a large mixed-use set is judged by its signing keys (cubic). Per-entry ceiling ≈ 2 KiB URL + 64 KiB material + 50 small structs; nothing is unbounded now.

Three new tests: `too_many_usable_keys_is_refused`, `keys_retaining_too_many_bytes_is_refused`, `an_overlong_jwks_url_is_refused`. 15 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Bounded the guest PEM key length (codex P1) and tightened two cap tests (claude P2).

**PEM cap:** `decoding_key_from_pem` now refuses a PEM longer than `MAX_GUEST_PEM_LEN = 8 KiB` before decoding — like the JWKS URL, the key is admin-set into an unbounded `TEXT` column and reparsed on every guest-JWT request, so a well-formed key with an oversized modulus could otherwise be stored and amplify. It is the same choke point `edit_guest_jwt_key` validates through, so an oversized key is never stored. 8 KiB covers RSA-16384 SPKI (~2.8 KiB) with margin. New test `an_oversized_pem_is_refused`.

**Test nits (claude):** `an_overlong_jwks_url_is_refused` now asserts the `longer than` error specifically (a bogus URL would fail the fetch regardless), and `too_many_usable_keys_is_refused` asserts `usable signing keys` rather than a substring the byte-cap message also contains.

Both stored admin inputs to the guest-JWT config (URL, PEM) are now length-bounded; the JWKS cache is bounded on URL, key count, and key bytes. 16 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Fixed the whitespace bypass (codex P1 / claude P2 / cubic P2): the PEM cap now measures the **untrimmed** input, since `edit_guest_jwt_key` stores what the admin sent. A key padded with 8 KiB of leading/trailing whitespace is now refused before storage rather than trimmed down to a small key that slipped the check. Test covers a whitespace-padded key.

Both stored admin inputs to the guest-JWT config are length-bounded before storage (URL in fetch_jwks, PEM in decoding_key_from_pem, both at the choke point the save path validates through). 16 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Bounded the guest JWT token length (codex P1) and corrected the URL-cap doc (claude P2).

**JWT cap:** `verify_for_workspace` now refuses a token longer than `MAX_GUEST_JWT_LEN = 8 KiB` at its top, before the key lookup, `decode_header`, verification, or caching. So an oversized token can't be decoded unauthenticated, and can't be cached at full size if it verified. Added to the refusals suite (`oversized token`).

**Doc (claude):** `MAX_JWKS_URL_LEN`'s comment no longer claims `cached_jwks` checks before caching; the bound holds because the save path validates the URL through `fetch_jwks`, so only a bounded URL is ever stored and cached.

Every input to the guest-JWT config and request path is now length-bounded: JWKS URL, JWKS response, JWKS key count and key bytes, the PEM key (untrimmed), and the JWT token. 16 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Fixed the repeated-prefix bypass (cubic P2) and the vacuous test (cubic P3 / claude nit).

**Prefix bypass:** the auth arm used `trim_start_matches(BEARER_PREFIX)`, which strips *every* `jwt_guest_` prefix, so a repeated-prefix bearer shrank to a valid short token that verified and was cached under the full oversized bearer key. Now it strips exactly one prefix (`strip_prefix`) and bounds the raw bearer length before verifying or caching (the cache keys on the whole bearer).

**Test:** the refusal case now mints a **valid signed** token over the cap — without the guard it would deserialize into GuestJwtClaims (the extra claim ignored) and verify, so it 401s only because of the cap — plus a repeated-prefix bearer. Both fail if either guard regresses.

5 entry + 16 unit + 1 allowance + rename pass.

rubenfiszel:
Addressed the two remaining nits (codex was Good to merge).

- **cubic P2:** the oversized-guest-bearer check now runs **before** the cache key is built, so an overlong bearer is rejected before it is copied and hashed for the lookup.
- **claude P2:** corrected the `strip_prefix` comment — it prevents a repeated-prefix bearer from reducing to a valid token cached under a non-canonical key; the separate length guard handles the oversized case.

5 entry + 16 unit + 1 allowance + rename pass.

rubenfiszel:
CI `cargo_test` was red on a pre-existing EE compile error unrelated to this feature: `src/ee.rs:1382` (`seat_count_membership`) called `seats_consumed` with two i64s, but it has taken `&ActiveSeatCounts` since the guest-seat feature (every production call site already passes it that way). Fixed in the EE companion (`count_offline_potential_seats` returns `ActiveSeatCounts`, so `seats_consumed(&c)`); ee-repo-ref bumped. `cargo test` now compiles and the test passes. No change to the guest-JWT feature.

rubenfiszel:
Addressed codex P1 (stale JWKS keys trusted indefinitely) and claude's logging nit.

**Stale keys fail closed:** `JwksEntry` now tracks `fetched_at` (last successful fetch), preserved across stale re-serves (which bump `expires_at`). Keys older than `JWKS_MAX_STALE = 1h` are no longer served on any path (fast path and the post-lock recheck); a too-old entry falls through to a blocking refresh that fails closed. So once an issuer revokes a `kid` or goes unreachable, its old keys stop minting guest JWTs within an hour rather than forever. New test `stale_jwks_keys_stop_being_served_past_the_grace_window` (dead loopback port so refreshes fail): within the window stale keys still serve, past it the request errors.

**Logging (claude):** the oversized-bearer refusal now logs like the other guest refusals instead of returning a bare None.

17 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Addressed codex P1 (plaintext HTTP) and claude's test-race nit.

**HTTPS required:** `validate_guest_jwks_url` now requires `https` — an on-path attacker could otherwise replace a plaintext JWKS response and forge accepted guest JWTs. `http` is allowed only under the existing `ALLOW_PRIVATE_GUEST_JWKS_URLS` operator opt-in (dev/loopback). New test `a_plaintext_http_jwks_url_is_refused_by_default` (asserts `DisallowedScheme`).

**Test race (claude):** the two JWKS tests that mutate the process-wide `ALLOW_PRIVATE_GUEST_JWKS_URLS` now lock a shared `TEST_ENV_LOCK`, mirroring the ssrf tests, so a concurrent run can't clear it from under the single-flight test.

Note: the previous `cargo_test` red was two flaky **relock** tests (`relock_noop.rs` — 'worker timed out', 'never reached the row lock'), unrelated to this feature; all guest-JWT tests passed. 18 unit + 5 entry + 1 allowance + rename pass.

rubenfiszel:
Addressed the HTTPS-message nit (codex P2 + claude P2).

Added a distinct `SsrfValidationError::HttpsRequired` variant ("URL must use https") so an `http://` JWKS URL without the opt-in no longer returns the shared `DisallowedScheme` whose text says http is permitted. Corrected `validate_guest_jwks_url`'s doc (the flag now opts the scheme in too, not just private ranges) and the Guests-card copy now says a JWKS URL must be https. `a_plaintext_http_jwks_url_is_refused_by_default` asserts `HttpsRequired`.

18 unit + 5 entry + 1 allowance + rename pass.

PR #10966, opened 2026-09-03
fix: deploy a relocked script version only when its lock changed

rubenfiszel:
## Summary

Follow-up to #10913 / #10915. Recording supplied lock hashes made importers *skip lock
generation* when nothing they import changed, but a skipped relock still deployed: every
relative-import dependency job for a script minted a new version, recorded deployment metadata,
and queued a dependency job for each of its own importers, whether or not the lock it ended up
with differed from the one it replaced. One deploy of a shared helper therefore walked the whole
downstream graph with new versions carrying byte-identical locks, and each hop re-walked its
own importers. The follow-up report on #10913 measured this on a fleet upgraded to the #10915
fix: 4,400 dependency jobs in 16 minutes across 94 importer paths, every generated lock
byte-identical to the one it replaced, and a queue of 665 that needed every schedule and
webhook paused to drain.

This PR moves the minting of a relocked script version from pull time to the end of the
dependency job, and mints it only when the generated lock or module locks differ from the live
version's. An unchanged result writes nothing and the walk stops there. Two smaller defects on
the same path go with it: the version is inserted with `created_at = clock_timestamp()` so a
child can no longer sort before its archived parent, and two relocks of one path serialize on
the head's row lock instead of one of them failing with `Non-archived script with path ... not
found`.

Closes the cascade half of #10913. Flows and apps keep their pull-time clone; see *Not here*.

## What happened on `main`

A relative-import dependency job for a script went through four places, and each of them
assumed the deploy was already decided:

1. **Push** (`trigger_dependents_to_recompute_dependencies`): for every importer recorded in
   `dependency_map`, push a `Dependencies` job with `triggered_by_relative_import`, debounced
   five seconds per path.
2. **Pull** (`clone_runnable` in `windmill-queue`): before the worker looks at anything,
   `clone_script` inserts a new version of the path with `lock = NULL`, `parent_hashes[1]` set
   to the live version, and archives the live version. The job's `runnable_id` becomes that new
   hash. From this moment the path's newest version has no lock, which the runtime resolver
   treats as a pending deploy for up to ten minutes.
3. **Generate** (`capture_dependency_job`): `try_skip_relock` compares the importer's recorded
   `imported_lockfile_hash` for each import against the current `lock_hash` row. When they all
   match it returns the path's last deployed lock and skips generation. This is what #10915
   made reachable for supplied-lock imports.
4. **Deploy** (`handle_dependency_job`): unconditionally write the lock onto the new version,
   upsert `lock_hash`, invalidate caches, call `handle_deployment_metadata` (a git-sync commit
   and a `deploymentcallback` job on builds that have it), call `process_relative_imports`
   (re-record this importer's edges, then push a job for each of *its* importers), trigger CI
   tests, and queue a binary prebuild.

So "skip" only ever meant skipping step 3. Reproduced on a three-script bun chain
`f/t/a` imported by `f/t/b` imported by `f/t/c`, with a comment-only redeploy of `f/t/a`:

| path | generation | new version | lock identical to parent | fanned out |
|---|---|---|---|---|
| f/t/a | relock (its own deploy) | yes | yes | to b |
| f/t/b | skipped | yes | yes | to c |
| f/t/c | skipped | yes | yes | no importers |

Three jobs and two pointless versions for a chain of three. The amplification in the report
comes from the shape of a real graph: a path reachable through several import chains gets one
job per chain unless the pushes land inside the same five-second debounce window, and each of
those jobs re-walks everything below it. The number of jobs grows with the number of distinct
root-to-node paths, not the number of nodes.

Two more defects come from step 2 minting the version before the outcome is known:

- **A NULL-lock window on every relock.** Until step 4 writes the lock, the path's newest
  version has none. `get_latest_deployed_script_hash` reports `pending_lock` for it and runs
  either wait or resolve to the previous version, for a change that in the unchanged case
  never materializes.
- **Two relocks of one path race on the clone.** `clone_script` selects the live version
  `FOR UPDATE`. A second clone that blocks on the same row re-evaluates it after the first
  commits, finds it archived, and fails with `Non-archived script with path ... not found`,
  which is the "preprocessing failure" the report counted 283 of. And when the second
  transaction *began* first but locked second, its `now()` predates the first's, so the live
  child it inserts has an earlier `created_at` than its archived parent. Every "latest version"
  read that orders by `created_at` alone then returns the archived parent:
  `GET /scripts/get/p/{path}` does, run-by-path resolves to the archived hash, and the
  workspace archive that `wmill sync pull` downloads drops the path entirely (it takes
  `max(created_at)` per path, then requires `archived = false`). All three were confirmed by
  planting the inverted timestamps locally.

## The fix

**Scripts are no longer cloned at pull.** `clone_runnable` keeps the flow and app arms and
leaves the script's `runnable_id` alone. The `base_hash` job arg it used to add was read by
nothing.

**The worker relocks the live version and decides afterwards.** In `handle_dependency_job`, a
job carrying `triggered_by_relative_import` resolves the path's live version at job start
(`live_head_hash`) rather than trusting the hash captured at push, since a deploy can land
during the debounce delay. It generates against that version's content as before. Then
`commit_relock` opens a transaction, re-reads the live version with `fetch_script_for_update`,
and takes one of three exits:

- **Superseded**: the live version is not the one the lock was generated for, because a
  deploy landed mid-generation. Nothing is written for it. Instead the job queues one more
  relative-import relock of the path, through the same push the fan-out uses, so the version
  that is live now gets locked against the current imports. That covers both a user deploy
  (which carried its own lock or queued its own dependency job, so the re-run usually ends
  Unchanged) and a sibling relock on a build without per-path serialization, where the
  discarded result may have been the fresher one and nothing else would have relocked the
  path. The job succeeds with `"Lock generation superseded by a newer version"`.
- **Unchanged**: the generated lock equals the live version's and the module locks equal too.
  Nothing is written to `script`. The `lock_hash` row is still upserted, under the same row
  lock so a deploy that takes the lock next cannot have its hash overwritten by this one (a
  version deployed before lock hashes were recorded has none, and its importers cannot skip
  until it does). The importer's dependency-map edges are refreshed so its next skip check
  compares against the imports' current hashes, and the job returns
  `"Lock unchanged, no new version deployed"`. No deployment metadata, no git-sync commit, no
  CI-test trigger, no prebuild, and no jobs for this path's importers.
- **Deployed**: `deploy_relocked_version` inserts the new version *complete*, with the lock,
  module locks and `parent_hashes`, archives the live one, and upserts `lock_hash` in the same
  transaction. The rest of the deploy runs as before against the new hash: cache invalidation,
  deployment metadata, `process_relative_imports`, CI tests, prebuild.

A generation failure on a relative-import relock still leaves what a failed deploy leaves: a
version with no lock that carries `lock_error_logs`, so the failure shows on the script while
runs keep resolving to the last version that has a lock. It goes through the same
`commit_relock`, so a version that landed meanwhile is left alone. A job that cannot even read
the live version no longer stamps `lock_error_logs` anywhere: on `main` that landed on the
clone, and here it would land on the version runs resolve to and take it out of resolution.

The worker's fallback tally for failed dependency jobs (the fork "Compare & Deploy" counters)
assumes every such job left a deployed version behind, which held when the version was minted
at pull. A relative-import relock now claims that tally up front and hands it back only when
its failure path did mint the error-carrying version, so a failure before any commit tallies
nothing.

The deploy's own dependency job (the one `create_script` queues when no lock is supplied) is
untouched: it still writes the lock onto the version the deploy created.

**`created_at` is stamped at insert time.** `deploy_relocked_version` writes
`created_at = clock_timestamp()`. The row lock on the live version is what orders one relock
after another, and a transaction that began first but locked second inserts after the first
committed, so its timestamp is later. With `now()` it was the transaction's start time and
could be earlier.

**Same-path relocks serialize instead of failing.** `fetch_script_for_update` inside
`commit_relock` still blocks on the row lock, and when the holder archives that row Postgres
re-checks only that row against `archived = false` and returns nothing: the successor is not
in the blocked statement's snapshot. That is the mechanism behind the `NotFound` failures on
`main`. `commit_relock` reads once more on an empty result, which is a fresh statement and
sees the successor, and then takes the Superseded exit and requeues (or Unchanged, when the
successor already carries this lock). On builds with the `private` feature the
`dependency:{workspace}/{path}` concurrency key already keeps two such jobs from running at
once; this closes the window for builds and pushes without it, and for a user deploy holding
the row while a relock reaches its commit.

**Module locks.** The per-module skip used to hand every module the *parent script's* last
lock (`try_skip_relock` was asked with the script's type and ignored the module lock it was
given), which on `main` silently overwrote module locks on every skipped relock, and here would
have made the Unchanged comparison fail for every multi-file script. It now returns the
module's own last lock, and a module that never had one generates instead of skipping. The
relocked version's hash covers the module locks the row stores.

Supporting changes:

- `refresh_dependency_map` is split out of `process_relative_imports` so the Unchanged exit can
  re-record edges without fanning out. `process_relative_imports` calls it and is unchanged for
  its other callers.
- `ScriptModule` derives `PartialEq` so module locks compare structurally.
- `maybe_queue_binary_prebuild` takes the hash to build rather than reading `job.runnable_id`,
  which is now the pre-relock version.

## What changes observably

| situation | `main` | this PR |
|---|---|---|
| Import redeployed, importer's lock unchanged | new version, git-sync commit, `deploymentcallback`, CI tests, prebuild, one job per importer of the importer | nothing written to `script`; `lock_hash` and edges refreshed; walk stops |
| Import redeployed, importer's lock changed | new version minted before generation, lock written later | new version inserted complete at the end |
| Runs during a relock | newest version has no lock for the duration; `pending_lock` | live version stays the newest; runs never see a lockless head |
| Relock fails | clone gets `lock_error_logs` | same shape, minted at the end |
| Deploy lands mid-relock | two live versions, or the relock's clone wins over the user's deploy | relock result discarded (`Superseded`); the deploy's own lock stands |
| Two relocks of one path overlap | second fails `Non-archived script ... not found`; possible `created_at` inversion | second serializes on the row lock; `created_at` monotonic along the lineage |

Git-sync users will notice that an import's redeploy no longer produces one commit per
importer for versions identical to their predecessors. CI tests that reference an importer are
no longer triggered by a redeploy of one of its imports that left the importer's lock alone;
the tests referencing the import itself still are. A test names the item it covers, and an
unchanged lock says the dependency set held, so this is the intended reading, but it is a
behavior change.

## Measured

Same three-script chain on the running instance, before and after, comment-only redeploy of
the leaf:

| | dependency jobs | versions minted on b, c | c reached |
|---|---|---|---|
| `main` | 3 | 1 + 1 | yes |
| this PR | 2 | 0 + 0 | no |

With a real dependency change on the leaf (adding an npm import) the walk still runs in full:
`a`, `b`, `c` each get one job and `b`, `c` each get one new version whose lock carries the new
package, with the archived parent older than its live child. A fourth importer deployed with a
supplied lock and a broken npm import got a failed relock: one new version with no lock and
`lock_error_logs` set, its parent archived, and run-by-path still resolving to the parent.

## Not here

- **Read-side ordering.** Fourteen "latest version" reads on `script` order by `created_at`
  with no `archived` filter (`get_script_by_path`, `get_latest_deployed_script_hash`, the
  workspace archive, and others). Instances that already hold an inverted pair from the race
  above keep returning the archived parent until those reads prefer live rows; that is a
  separate change across `windmill-api`. `create_script` still stamps `now()` too, and has the
  same theoretical inversion through its own row lock.
- **Multi-file bun scripts cannot be locked without a supplied lock.** Observed while trying
  to verify the module-lock change live: a bun script created through the API with a
  `modules` entry and `import { greet } from "./helper.ts"` fails its own dependency job
  with `Failed to find relative import ... /f/t/helper.ts`, because lock generation never
  writes the module files (the comment above `prebundle_bun_script` says as much) and the
  bundler resolves the import as a workspace path. Pre-existing and independent of this PR,
  which is why the module-lock change is covered by the compiled test suite and code reading
  rather than a live relock.
- **Flows and apps** still clone at pull and deploy unconditionally. They are importers only,
  never imported, so a relock of one fans out to nothing; what remains is one version and one
  deployment-metadata write per relock, without the cascade.
- **Why same-path relocks overlapped on the reporting fleet.** On CE and EE images
  `push` assigns `dependency:{workspace}/{path}` at limit 1 to every dependency job and the
  pull applies it before the clone, so overlapping clones mean the key was not on those jobs.
  A push from a binary built without `private`, or a `worker_ping` row under five minutes old
  advertising a version below 1.566.0, would both do that. This PR makes the overlap harmless
  either way, but the gate itself is unchanged.

## Test plan

- [x] New `backend/tests/relock_noop.rs`: a bun chain `a` imported by `b` imported by `c`. A
      comment-only redeploy of `a` produces exactly two dependency jobs (`a`'s own and one for
      `b` whose logs say `Lock unchanged: no new version deployed`), no job for `c`, and no new
      version on either. A redeploy of `a` adding `lodash@4.17.21` produces one job each for
      `a`, `b`, `c`, a second version on `b` and `c` whose lock contains the package, archived
      parent older than live child. Fails on `main` at the first assertion (three jobs, two
      versions).
- [x] Two more tests in `relock_noop.rs`, both deterministic. `relock_waiting_on_a_deploy_
      requeues_for_its_successor` holds the importer's head row `FOR UPDATE` from the test,
      triggers its relock, waits until `pg_stat_activity` shows the relock blocked on that lock,
      then archives the head and inserts a successor before committing: the relock must succeed,
      log the supersede, and the requeued relock must find the successor's lock current, leaving
      exactly the successor live. `multi_file_importer_relock_is_a_no_op_too` (python) deploys a
      two-file importer and checks that an import's content-only redeploy leaves it at one
      version with its module's own lock intact.
- [x] `cargo test --features python,private,quickjs --test relock_skip --test dependency_map
      --test script_modules --test relock_noop` (2 + 8 + 3 + 3). `relock_skip`'s
      workspace-dependency test waited for ten jobs in its first two steps, which was the size
      of the cascade an unchanged redeploy used to produce; it now gets exactly the five direct
      importers, so those two thresholds moved to five.
- [x] The live checks in *Measured*, on this branch's backend, including the failed-relock
      importer: run-by-path resolves to its archived parent's hash and `GET /scripts/get/p/`
      shows the error-carrying version.
- [x] The Superseded exit, live: an importer given a heavy dependency of its own, the
      worker's bun cache cleared so its relock generates cold for several seconds, and a new
      version of it deployed with a supplied lock 11 ms after that relock started. The relock
      logged `Version <hash> was deployed while this lock was generated; discarding it and
      queueing a relock of that version` and wrote nothing; the requeued relock ran five
      seconds later against the user's version, found its supplied lock stale (the import had
      gained a dependency) and deployed a child of it carrying the new lock. Four versions in a
      clean lineage, one live.
- [x] Two workers on this build (no `private`, so no per-path serialization), two deploys of
      `a` 300 ms apart each adding a dependency: the first relock of `b` resolved the live head
      after both had landed and produced the lock with both packages, and the second relock, 154
      ms later on the same worker, found it unchanged. The relocks did not overlap by timing, so
      the concurrent commit path is covered by the deterministic waiting-relock test above
      rather than by this run.
- [x] `cargo sqlx prepare` for the one changed query; the cache diff is one added and one
      removed entry.
- [x] `cargo_test` on this PR fails to compile `windmill-api/src/scim_ee.rs` with dead-code
      errors from the SSO groups commit `79426a1a68`; main's own last four `backend-test` runs
      fail the same way, so it is not this branch's to fix.
- [x] Two cold-context reviews before opening (`local-review`, `local-review-codex`). Their
      findings are folded in: the unchanged path writes `lock_hash` under the row lock rather
      than after releasing it, a relock claims the failure tally up front, the Superseded exit
      re-queues instead of discarding (a sibling relock never queues one for its own path),
      and the live head is read from the database rather than through the per-worker script
      cache, which would have pinned a lockless entry under the head's hash for the life of
      the worker when the head's own dependency job had not written its lock yet.
- [x] First CI round (Codex and Claude, same P1): a relock that had to wait on the head's row
      lock got an empty result and failed with `NotFound` instead of taking the Superseded
      exit. Fixed with the re-read described above, along with the two module-lock nits.
- [x] Second CI round: Codex and Pi good to merge, Claude nit-only (the re-read was one deep
      and could lose the same race to a second writer). The re-read is now a bounded loop.

Left in draft on purpose: this changes the job pull path and the dependency handler, which is
shared worker infrastructure, so it wants a human look before it goes ready.

https://claude.ai/code/session_01GdEb6gzCZ2qXmAQJAeMf9W


rubenfiszel:
✅ Review round clean @ 196401c01fa96d1bc81bb86162bfbf756abcc327 — nit-only verdicts at 57931f7c3ef25b9eee0fde879633546e67a18941 (Codex and Pi good to merge, Claude nit-only); nit addressed in 196401c01fa96d1bc81bb86162bfbf756abcc327

rubenfiszel:
✅ Review round clean @ 157542765ec55294e85c80c89d37219985308bbd — verdicts at 5b111be002ccf4750a997256155f90882e88ab82 (Codex, Pi and Claude good to merge); tests added in 157542765ec55294e85c80c89d37219985308bbd, no source change since the round

PR #10974, opened 2026-09-04
feat(otel): support standard OTEL resource attribute env vars

hugocasa:
Resolves WIN-2480.

## Problem

Three ways of setting OTEL resource attributes were silently ignored.

1. The resource was built with `Resource::builder_empty()`, which runs **no** resource detectors, so anything injected through the standard `OTEL_RESOURCE_ATTRIBUTES` env var (`k8s.pod.uid`, `k8s.container.name`, `service.namespace`, `service.instance.id`, …) was parsed by nobody and dropped.
2. `service.name` was hardcoded to `windmill-<mode>`, so `OTEL_SERVICE_NAME` was read by the SDK and then immediately overwritten.
3. `service.version` was hardcoded to the git version, with no env var reading it.

For 2 and 3 there was no workaround: because both attributes are set in code they also outrank `OTEL_RESOURCE_ATTRIBUTES` (see precedence below), so there was no route to set them by any means.

This matters for deployments that identify each pod from its own labels, the normal Kubernetes pattern:

```yaml
- name: OTEL_SERVICE_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['app.kubernetes.io/name']
```

`fieldRef` resolves per pod, so each component still gets its own `service.name`.

## Fix

- Use `Resource::builder()`, which seeds from the SDK's `EnvResourceDetector`, so `OTEL_RESOURCE_ATTRIBUTES` is picked up.
- Read `OTEL_SERVICE_NAME` and `OTEL_SERVICE_VERSION` explicitly, mirroring the existing `OTEL_HOST_NAME` override.
- An env var set to the empty string counts as unset. The downward API yields `""` for a label the pod does not carry, so a missing label falls back to Windmill's value instead of blanking the attribute; the spec requires this for `OTEL_SERVICE_NAME` anyway.

`OTEL_SERVICE_VERSION` is **not** an OTel env var — the spec defines only `OTEL_SERVICE_NAME` — but deployments set it expecting it to work because it sits next to the one that does, and setting `service.version` in code blocks the `OTEL_RESOURCE_ATTRIBUTES` route.

## Precedence

Per the [OTel resource spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md):

> The SDK MUST extract information from the `OTEL_RESOURCE_ATTRIBUTES` environment variable and merge this, **as the secondary resource**, with any resource information provided by the user, i.e. the user provided resource information has higher priority.

⚠️ "User" there means **the application using the SDK** — Windmill's own code — not the person operating the deployment. The rule reads: *attributes set by the instrumented application outrank the environment variable.*

So for a given attribute key:

**dedicated env var** > **Windmill's built-in value** > **`OTEL_RESOURCE_ATTRIBUTES`**

`OTEL_SERVICE_NAME` also outranks a `service.name` inside `OTEL_RESOURCE_ATTRIBUTES`, as the spec requires.

## What each attribute resolves to

| attribute | dedicated override | default when unset |
|---|---|---|
| `service.name` | `OTEL_SERVICE_NAME` | `windmill-<mode>` |
| `service.version` | `OTEL_SERVICE_VERSION` | git version |
| `host.name` | `OTEL_HOST_NAME` | instance hostname |
| `deployment.environment.name` + `deployment.environment` | `OTEL_ENVIRONMENT` | first label of the base URL |
| `telemetry.sdk.{name,language,version}` | — | from the SDK |
| any other key (`k8s.*`, `service.namespace`, `service.instance.id`, …) | — | set it via `OTEL_RESOURCE_ATTRIBUTES` |

`OTEL_HOST_NAME` and `OTEL_ENVIRONMENT` already existed; `telemetry.sdk.*` is newly exported because it comes with `Resource::builder()`. Of these, only `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` are OTel-standard variables.

## Verification

Run against a real `otel/opentelemetry-collector-contrib` (OTLP gRPC receiver → `debug` exporter at `verbosity: detailed`), so the attributes below are the collector's own parse, not a hand-decode.

Env, shaped the way a Kubernetes deployment would set it, including a competing `service.name` inside `OTEL_RESOURCE_ATTRIBUTES`:

```
OTEL_SERVICE_NAME=windmill-workers
OTEL_SERVICE_VERSION=1.796.0
OTEL_RESOURCE_ATTRIBUTES='k8s.pod.uid=8a1f-abc-123,k8s.container.name=windmill,k8s.namespace.name=prod,
                          k8s.deployment.name=windmill-workers-default,service.namespace=acme,
                          service.name=should-lose-to-otel-service-name'
```

Collector output:

```
Resource attributes:
     -> host.name: Str(thedevbox)                              # Windmill default, no override set
     -> telemetry.sdk.language: Str(rust)
     -> telemetry.sdk.name: Str(opentelemetry)
     -> telemetry.sdk.version: Str(0.30.0)
     -> deployment.environment.name: Str(localhost)            # Windmill default
     -> deployment.environment: Str(localhost)
     -> service.name: Str(windmill-workers)                    # OTEL_SERVICE_NAME beats the injected one
     -> service.version: Str(1.796.0)                          # OTEL_SERVICE_VERSION
     -> k8s.container.name: Str(windmill)
     -> k8s.deployment.name: Str(windmill-workers-default)
     -> k8s.namespace.name: Str(prod)
     -> k8s.pod.uid: Str(8a1f-abc-123)
     -> service.namespace: Str(acme)
```

Every `k8s.*` attribute and `service.namespace` is dropped entirely before this change. Confirmed on both the metrics and logs pipelines; traces take the identical `otlp_service_resource(...)` value but were not separately exercised, since the `full_job` span only fires on job completion.

Two regression tests in `backend/tests/otel.rs` pin the merge precedence, `OTEL_SERVICE_NAME` outranking `service.name` in `OTEL_RESOURCE_ATTRIBUTES`, and the dedicated overrides including empty-means-unset.

## What this does and does not change for the stock Helm chart

Worth knowing before pointing anyone at the two `OTEL_SERVICE_*` variables — with `windmill-helm-charts` as shipped:

- `app.kubernetes.io/version` is **never set on any pod**: the `windmill.labels` helper that defines it is not included by any template. A `fieldRef` to it yields `""`, which falls back to the git version — so `OTEL_SERVICE_VERSION` wired that way is a no-op. (Were the helper wired up it would emit `.Chart.AppVersion`, pinned in `Chart.yaml` and stale against whatever `windmill.tag` actually runs.)
- `app.kubernetes.io/name` is per-component but not per-worker-group — every worker pod carries `windmill-workers`, with the group in a separate `workerGroup` label. So it renames `windmill-server` → `windmill-app` and `windmill-worker` → `windmill-workers` without adding granularity. Worker groups do accept custom labels (`workerGroups[].labels`), so a deployment that sets its own can get something more specific.

The substantive win here is `OTEL_RESOURCE_ATTRIBUTES`, which is what carries `k8s.*` through.

## Upgrade impact

These variables were previously ignored, so an instance that already has them set in its environment — without changing anything — sees different telemetry after upgrading:

| already set | before | after |
|---|---|---|
| `OTEL_SERVICE_NAME` | ignored | **`service.name` changes** from `windmill-<mode>` to that value |
| `OTEL_RESOURCE_ATTRIBUTES` | ignored | its attributes now appear on every metric, log and span |
| *(nothing)* | — | `telemetry.sdk.{name,language,version}` newly present |

The first is the one that can bite: `OTEL_SERVICE_NAME` is a standard variable, so it may already be set namespace-wide for other workloads, or injected automatically by the OpenTelemetry Operator. Dashboards, alerts or saved queries keyed on `service.name = "windmill-server"` would silently start missing data. Worth a changelog line.

`OTEL_RESOURCE_ATTRIBUTES` is additive and cannot override anything Windmill sets, so the risk there is cardinality rather than correctness — see the `target_info` note below.

One smaller change to an existing variable: `OTEL_HOST_NAME` set to the empty string previously produced `host.name=""` and now falls back to the hostname. Deliberate — a Kubernetes downward-API `fieldRef` to an absent label yields `""`, and blanking the attribute is never what that means.

## Notes for reviewers

- **Pre-existing, unrelated failure in the same suite.** `test_root_job_span_created_on_success` asserts `job_kind` and `created_by` on the EE `full_job` span. Those assertions came from #8718, whose own description says it covered only the OSS `job` / `job_postprocessing` spans and left the `full_job` span to a follow-up EE change that was never made — so the assertion has never passed. It goes unnoticed because CI's `cargo test` feature list has no `otel`, and the file is `#![cfg(all(feature = "otel", feature = "enterprise"))]`. Left alone here rather than deleted, since the right fix is the missing EE propagation, not removing the evidence. Worth its own ticket.
- **This suite does not run in CI**, for the reason above, and no `cargo check` job passes `--all-targets`, so it is not even type-checked. Adding `otel` to the backend-test feature list would fix that but must come after the stale assertion above. Run locally with `cargo test --features enterprise,private,otel --test otel -- --test-threads=1`.
- `#[serial_test::serial]` orders the two new tests against each other but not against the 13 unannotated tests in the binary. Harmless while the file never runs in CI; if `otel` is added to CI (which runs `--test-threads=10`), the others need `#[serial_test::parallel]` first.
- `telemetry.sdk.*` is additive for trace and log backends. In a Prometheus-via-OTLP setup resource attributes become `target_info` labels, so that series gains three; existing `on(instance, job) group_left(...) target_info` joins keep working since the join keys are untouched.
- `otlp_service_resource` became `pub` so the integration tests can call it, matching the other items the suite already uses from that module.

https://claude.ai/code/session_013yUJaBDHPgjZPodP2u4aqR


hugocasa:
Round clean at `a29ba3812c` — Codex and Claude both "Good to merge", Pi mergeable with one nit.

**Dismissing Pi's nit** (the `otel` suite not running in CI) rather than fixing it here: adding `otel` to the backend-test feature list turns CI red on the pre-existing `job_kind`/`created_by` assertion documented in the description, whose correct fix is the EE `full_job` propagation that #8718 promised and never landed. Both halves belong in that follow-up ticket, not this PR.

**Left in draft deliberately** — this is an EE change (companion windmill-labs/windmill-ee-private#779) and wants a human look before ready.

One thing worth a maintainer's attention: Claude reports that across all seven rounds the EE checkout available to the review job has stayed at the pre-PR base `b9aeffa8`, not the `a123db94` this PR's `ee-repo-ref.txt` points at — so the substantive change has not been visible to it in any round, and its verdicts cover the OSS test file and description only. The ref itself is correct and pushed (`a123db94` is the head of the EE branch and reachable on origin), so this looks like the review workflow's EE checkout not honouring `ee-repo-ref.txt`. The EE half was reviewed locally instead (Claude and Codex both pointed at the EE worktree explicitly, both clean) and exercised end to end against a real OTel Collector.

hugocasa:
✅ Review round clean @ 98da21a6422fd2cb53dc2fd01bbe3fbbf69809f0 — verdicts at `a29ba3812c93e9d0c9ce4ed963159088dbf71c62`: codex "Good to merge", claude "Good to merge", pi "Mergeable, but should ideally address nits" (the nit — the `otel` suite not running in CI — dismissed with reasoning in the comment above).

The head moved only by `98da21a642`, a merge of `origin/main` resolving the `backend/ee-repo-ref.txt` conflict (main advanced its EE pin to `81edd138`; EE main was merged into the EE branch first, so the pin is now the resulting EE merge commit `62efa909`). No source change: the net diff vs main is unchanged at `backend/ee-repo-ref.txt` + `backend/tests/otel.rs`.

Re-validated at this head rather than assuming the merge was inert:
- Local Codex review under the same `REVIEW.md` policy at xhigh, scoped so the EE diff was the primary subject: **Good to merge**, no findings.
- `cargo check -p windmill-common --features enterprise,private,otel` — clean.
- `cargo test --test otel --features enterprise,private,otel` — 14 passed, 1 failed; the failure is the pre-existing `test_root_job_span_created_on_success` ("missing job_kind attribute") from #8718, untouched by this diff and already recorded as a follow-up. The two tests added here pass.

PR #10975, opened 2026-09-04
feat: durable dbt state per environment, and `--defer` onto it

rubenfiszel:
## Summary

`dbt retry` worked off two artifacts and only one of them was durable: `dbt_run_state`
holds `run_results.json` keyed by the executing principal, while `manifest.json` lived on
worker-local disk under a four-generation cache. That is enough to resume the last run and
nothing else — the next run of a project usually lands on a worker holding neither
artifact — so `--defer`, which resolves an unbuilt `ref()` through a manifest, had nothing
to read.

This adds a durable dbt state per environment and wires deferral onto it. Both stages ship
here because a store with no reader is dead weight: stage 2 turned out to be a modest
addition (a `build` command-block field, one materialisation the retry path already did,
and the per-engine flag split below).

Spec updated in `docs/dbt-runtime.md`: new decision 25, a "Durable state per environment"
section, decision 13 rewritten (the manifest IS stored now, once per environment), phase 4
narrowed to what is genuinely left.

## Changes

- **`dbt_environment_state`** (new migration): one row per (workspace, script path,
  environment), holding `manifest.json` and `run_results.json` from the last **successful**
  run. Separate from `dbt_run_state`, which answers a different question — last run whatever
  its outcome, keyed by principal, for `dbt retry`.
- **Environment** is `<warehouse>|<target>|<schema>|<database>`, where the target is the one
  dbt actually runs (the descriptor's, else the workspace warehouse's, else the project's own
  `profiles.yml` default). The last two are in the key because deferring resolves relation
  NAMES: a repointed warehouse resource or a moved `profile.schema` keeps the first two while
  putting every relation somewhere else, and a manifest has no other way to notice. A move
  reads as an environment nothing has published yet, which is what it is.
- **Who publishes**: a successful `build` whose graph becomes what the script owns
  (`GraphRefresh::publishes_ownership`) — the same condition and the same reason. An
  invocation that scoped its own model set describes where the *caller* put those relations.
  That is also what makes the ordinary flow work with no second concept: a deferring run is
  by construction a narrowed or re-varred one, so it reads the state and does not replace it.
- **Where the blob goes**: inline in the row under `DBT_STATE_INLINE_MAX_BYTES` (8 MiB), the
  INSTANCE's object storage above it with the row keeping the key — where every other internal
  worker artifact already lives, and deliberately not the workspace bucket, which members write
  through `job_helpers/*` with a caller-supplied key (only `volumes/` is reserved there). A
  manifest under that bucket is one any member could replace, and the next deferring run would
  hand dbt an attacker-chosen `defer_relation` while holding the script's warehouse
  credentials. Inline is what makes this work with no storage configured at all; past the
  ceiling with none, the job log says so and names the setting — the run still succeeds.
- **Publication is atomic.** Each one writes its own object keys
  (`…/<digest of path and environment>/<job>/<artifact>`) and the row switches to them in one
  statement, so an upload never overwrites an artifact the committed row still names. The
  objects a commit displaces are dropped afterwards — never before, since a reader that has
  already read the row is about to fetch them — and a reader that loses that race re-reads the
  row once. What a publication uploaded and then could not commit is dropped on the way out.
  Publishers of one environment serialize on `pg_advisory_xact_lock`, which covers the first
  publish too, where there is no row to lock.
- **The write's guard names the version and the job kind.** The live dbt script at the path
  must be the one this job ran or a later version of it (`hash = $n OR $n = ANY(parent_hashes)`),
  and it **holds that row** (`FOR SHARE`) for the rest of the publication so a rename, archive
  or delete either waits for it or is seen by it. The kind is checked too, because a preview
  carries a caller-supplied `script_hash` into `runnable_id` — the version alone would let
  anyone who may run a job publish arbitrary content as a deployed script's state.
- **A run that deferred never publishes**, whatever narrowed it: it built some of the relations
  its manifest names and resolved the rest out of the state it read.
- **A project-owned `profiles.yml` that templates its schema or database is refused a
  deferral** — dbt renders those and Windmill does not, so two renderings would share one
  environment key. Plainly absent is left alone; that is the adapter's default.
- **A `show` defers too** — it compiles the model it previews, so a model whose upstream this
  environment built and this run did not is exactly the deferral case.
- **A build recovered by the automatic in-job node retry publishes its manifest without
  results** — `run_results.json` is then the retry's, naming only the nodes it redid. The
  manifest is a function of the project, so deferral is unaffected.
- **Retention**: none, deliberately, and the migration says why. One row per script per
  environment, replaced in place, whose reader is every later run of that script — so it goes
  with the script instead (`clear_dbt_script_state` / `clear_dbt_script_state_if_path_retired`,
  renamed from `*_dbt_run_state` now that they carry both halves). A **rename** moves both:
  the path is only a prefix of an object key and the row is what names an artifact, so a script
  created at the old path cannot publish over a moved row.
- **`defer`** is a field on the `build` command block, defaulting to the descriptor's own
  `defer:`. Per-run rather than descriptor-only because the run that publishes an
  environment's state and the run that defers to it are two invocations of ONE script.
- **One materialisation path**: `write_state_dir` writes the artifacts a `--state` directory
  holds, and the two `dbt retry` restores (worker-local generation, database row) go through
  it as well.
- **`--state` is also a retry's own argument**, which is the trap this had to work around.
  `dbt retry` reads the run it RESUMES from `--state`; handed the deferral's directory it
  resumes the successful run stored there and rebuilds nothing (silently, on dbt-core 1.x —
  verified). So a retry on 1.x takes `--defer-state`, and a run that deferred is refused a
  retry on the Rust engines, which have no such flag, rather than rebuilding its nodes with
  every unbuilt `ref()` resolving into the schema this run writes into. The automatic in-job
  node retry is dropped there for the same reason and says so in the log.
- The state directory is passed **relative** (`wm_dbt_state`, beside `wm_target` in the job
  directory): dbt records the invocation's flags into `run_results.json`, so an absolute path
  would name the job directory of the run being resumed, gone by then.
- The result gains `deferred_to` — what a deferring run built against is otherwise
  unrecoverable, since the next successful run of that environment replaces the state.

## Two engine facts found while wiring this up

Both are pre-existing and documented in `docs/dbt-runtime.md` rather than fixed here:

- **`dbt retry` on dbt-core 2.x restores neither the resumed invocation's `--vars` nor its
  deferral** — it re-parses with the current (empty) ones, so a retry of a run that overrode
  `vars` rebuilds into the descriptor's schema rather than the run's. Independent of deferral
  and predates it; the refusal above stops the deferring case from being how it is discovered.
- **`dbt show` on either Rust engine prints a bare JSON array** where dbt-core frames it as
  `{"node": …, "show": […]}`, which `run_show` is written against — so a preview there fails to
  parse with or without `defer`. The deferral itself resolves correctly under it (verified: the
  row comes back, it is the framing that is rejected).
- Neither Rust engine reached dbt's own service-backed State (`--manage-state`) on any run
  measured here, so no flag is passed to disable it.

## Why this is not keyed by principal, unlike `dbt_run_state`

dbt records the invocation's flags into `run_results.json`, and Windmill resolves `$var:` /
`$res:` before dbt sees them — which is exactly why the retry state is per-principal. Here they
cannot be one caller's: a publishing run added nothing of its own, and a descriptor
interpolating a `{{ }}` placeholder into `vars` never publishes at all, so what is recorded is
the descriptor's own arguments — the script's content, which anyone entitled to run it may
already read. Stated in the code at the publish site, because widening the publish condition
would end it.

## Test plan

Exercised on this worktree's instance against a real dbt project (`stg_a` → `mart_b`, plus a
`generate_schema_name` macro reading `var('schema_suffix')`) and a real Postgres / DuckDB
warehouse:

- [x] **dbt-core-1x**: full build publishes the state (535 KB manifest, 4.7 KB results);
      a narrowed run into `analytics_dev` with `defer: false` fails with
      `relation "analytics_dev.stg_a" does not exist`; the same run with `defer: true`
      succeeds, builds `analytics_dev.mart_b` off `analytics.stg_a`, and returns
      `deferred_to`. The deferring run does **not** replace the state.
- [x] **`dbt retry` of a deferring run** (1.x): resumes exactly the failed node and still
      defers — the failure is the model's own `division by zero`, not relation-not-found.
      The same via the automatic in-job `retry_failed_nodes`.
- [x] **`test_behavior: after_all`**: the test phase runs under the deferral too.
- [x] **dbt-core-2x** (DuckDB): same publish / control / defer sequence; a retry of a
      deferring run is refused with the engine named.
- [x] **fusion** (DuckDB, 2.0.0-preview): publish and defer both work.
- [x] **Object storage**: with `DBT_STATE_INLINE_MAX_BYTES=1024` and a filesystem instance
      store, both artifacts land in it, the row keeps only the keys, the deferring run reads
      them back, and a republish overwrites in place (still two objects). The published
      environment key reads `main|prod|analytics|dbt_wh_defer` — the inherited target, which
      was empty before the effective-target fix.
- [x] **Recovered by the in-job node retry**: a build failing on a model whose missing relation
      is created during the retry delay succeeds, and publishes `manifest_key` with both
      results columns NULL.
- [x] **No storage, past the ceiling**: the run succeeds and the job log names the setting.
- [x] **Under nsjail** (`DISABLE_NSJAIL=false`, `isolation=nsjail` in the job log): the full
      build publishes to the instance store and the narrowed deferring run reads it back and
      builds `analytics_dev.mart_b` off the published relation.
- [x] **Displaced objects are dropped**: after a republish the store holds exactly the new
      pair, keyed by the new job id, and the row names them.
- [x] **`show` with `defer`** returns rows for a model whose upstream this run did not build;
      the same `show` with `defer: false` fails. On dbt-core 2.x and fusion the deferral
      resolves correctly (the run builds off the published relation) but `run_show`'s parser
      rejects their output — see the engine note below.
- [x] **Rename**: the row travels with the script, a deferring run at the new path uses the
      state published before the rename, and the next full build publishes under the new path's
      digest with the displaced objects dropped.
- [x] **A preview claiming the deployed path and hash** builds successfully and leaves the
      published row untouched.
- [x] **A deferring run publishes nothing** even submitted with no overrides at all — the case
      `publishes_ownership` cannot see, since a descriptor-level narrow `select` needs none.
- [x] **Nothing published yet**: refused with the environment named and the one action to take.
- [x] `cargo check` with and without `parquet`; `cargo test -p windmill-parser-yaml`,
      `-p windmill-worker --lib`, `-p windmill-common --test dbt_graph_storage`;
      `SQLX_OFFLINE=true` check of lib and test targets; `npm run check:fast` (4 pre-existing
      errors, none in touched files).

**`cargo_test` is red on this PR, and on `main`.** Run 33884307602 (branch `main`, 14:32 UTC,
before this branch was pushed and with no dbt changes) fails the same two tests with the same
panics: `relock_waiting_on_a_deploy_requeues_for_its_successor` at `tests/relock_noop.rs:275`
("b's relock never reached the row lock") and `relative_import_relock_deploys_only_when_the_lock_changed`
at `windmill-test-utils/src/lib.rs:384` after >60s. Every run of that workflow in the last hour
is failing, across branches. Both pass locally on this branch
(`cargo test --test relock_noop` → 2 passed; CI runs a third, `multi_file_importer_relock_is_a_no_op_too`,
which passes there too).

All of the above was re-run after the first review round, and the nsjail-sandboxed path — the
one gap the first round flagged — is now exercised rather than reasoned about.

## Screenshots

The `defer` toggle on the generated run form, and the descriptor a new dbt script starts with:

deferfield

dbt-template3


rubenfiszel:
Round 3 addressed in `049628036f`.

**Fixed:**
- **[Codex P1] a deferring build could replace the state it read.** Correct, and the case `publishes_ownership` cannot see is a *descriptor-level* narrow selection — no caller override, so it published. `publish` now also requires `inv.deferral.is_none()`. Verified: with the state at job `01a06f47-5716…`, a deferring build submitted with no overrides at all leaves the row on that same job.
- **[Codex P2, Claude P2, Pi] `publication_lock` overflow.** `u64::from_str_radix(…) as i64`, plus a unit test that 64 distinct environments get 64 distinct keys — which is what would have caught it.
- **[Codex P2] `db.begin()` failure leaked the uploaded objects.** Routed through the same cleanup path.
- **[Claude, not inlined] the destination delete in `move_dbt_script_state`** leaves its objects behind; the clause is now next to that delete rather than only on `clear_dbt_script_state`.

**Not fixed — [Claude P1] `dbt ls` and `--defer`.** dbt-core's `list` does carry it:

```
$ dbt ls --help | grep -i defer
  --defer / --no-defer            If set, resolve unselected nodes by
  --defer-state DIRECTORY         Override the state directory for deferral
```

and it parses — `dbt ls --defer --state …` reaches credential resolution, where a genuinely unknown option does not:

```
$ dbt ls --no-such-option …
Error: No such option '--no-such-option'.
```

`dbt-sa-cli ls --help` and `fusion ls --help` list `--defer` and `--state` too. It is also already exercised: every `defer: true` build in the matrix carries a `select`, so `resolve_selection` ran on all three engines and the builds succeeded. Happy to drop `--defer` from the `ls` (only `--state` is what a `result:` selector needs) if you would rather carry less, but it is not a break.

**Also verified since the last round**, both things the round asked for: `show`+`defer` on dbt-core 2.x and fusion — the deferral resolves correctly there, though `run_show`'s parser rejects those engines' bare-array output with *or* without `defer` (pre-existing, now recorded in `docs/dbt-runtime.md`); and a rename of a script with object-backed state — the row travels keeping its `manifest_key`, a deferring run at the new path reads it back, and the next full build publishes under the new path's digest and drops the old objects.

rubenfiszel:
Round 5 addressed in `0dd836eef9`.

**Fixed:**
- **[Codex P1] generated profile templates share one environment key.** Correct — I had only covered the project-owned `profiles.yml`. A `dbt_profile` resource is one block of the user's own file copied through unchanged, and `profile.schema` is written as given, so both can carry a template. The check now reads the rendered profile's schema/database too.
- **[Claude nit] the guard was read-side only.** It now refuses publication as well: a published template would sit under a key a literal profile shares, so de-templating later would make that stale manifest readable as the new location's. Verified both sides — a full build under a `{{ env_var('DBT_SCHEMA', 'analytics') }}` schema succeeds and publishes nothing, and a deferring run of it is refused by name. The remedy text no longer mentions `profile.schema`, which was inert on the path that could reach it.
- **[Codex P2] the documented key layout** now carries `<job>.<nonce>`.

**Not fixed — [Pi P2] flow/app-invoked builds can never publish.** They can. A flow step naming a deployed dbt script by path is an ordinary `script` job carrying that script's own hash, not a `FlowScript` — `FlowScript` is inline flow code, which a dbt project (a module bundle) is not, and which has no deployed version to publish for anyway. Verified end to end: a one-step flow over `u/admin/deferprobe_renamed` produced

```
kind    | runnable_id          | runnable_path
script  | 5267659369627786680  | u/admin/deferprobe_renamed
```

(`5267659369627786680` = `0x491a7c49946c1db8`, the deployed hash) and the step published `dbt_environment_state` normally. I have added a clause to `docs/dbt-runtime.md` saying so, since it is worth stating explicitly.

rubenfiszel:
✅ Review round clean @ 6ff2706ff7d993dffa172fe30b50b4e9dca7cf96 — nit-only verdicts at 96899e7a573080884c0afa21eba218cf6a7eaf47 (Pi: good to merge; Codex and Claude: nit-only, both naming the same stale `docs/dbt-runtime.md` sentence); nit addressed in 6ff2706ff7d993dffa172fe30b50b4e9dca7cf96

rubenfiszel:
Merged `origin/main` (head `ddf2603c9e`) and moved the EE worktree to the ref that merge brought in (`3b89bfc` → `1a10132e`), so the earlier clean-round marker no longer applies — a fresh round is running against the merged head.

**`cargo_test` is red for a reason outside this PR, and it is not the `relock_noop` flake this time.** The EE tree at the ref `main` pins does not compile its own test target:

```
error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> src/ee.rs:1382
1382 |  assert_eq!(seats_consumed(c.author_count, c.operator_count), 3.5);
note: function defined here
   --> windmill-common/src/ee.rs:160
 160 |  pub fn seats_consumed(seats: &ActiveSeatCounts) -> f64 {
```

Both files are EE symlinks, and at `windmill-ee-private@1a10132e` — the exact commit `backend/ee-repo-ref.txt` names on `main` — the caller is still the two-argument form while the definition has moved to `&ActiveSeatCounts`. It reproduces on every branch, not just this one:

| run | branch | failure |
|---|---|---|
| 33942296993 | `dbt-prod-state-defer` | `E0061` at `src/ee.rs:1382` |
| 33941801243 | `dbt-materialize-target` | same |
| 33940487027 | `fix-relock-noop-test-runtime` | same |

It needs a fix in `windmill-ee-private` (or an `ee-repo-ref.txt` bump past it), which is someone else's change — nothing in this PR touches seats or licensing. `cargo check` passes because the mismatch is inside a `#[cfg(test)]` block, so only the test target is affected.

rubenfiszel:
✅ Review round clean @ ab3f76564a7e980b1680abf33b2897bc9e7af0ff — verdicts at ddf2603c9e40ed61ebe0523f56e20fe67136f28a were Codex: good to merge, Pi: good to merge, Claude: nit-only (one docs sentence that excluded `load`'s second error exit); nit addressed in ab3f76564a7e980b1680abf33b2897bc9e7af0ff

rubenfiszel:
Merged `origin/main` (head `98c4f43301`) and moved the EE worktree to the bumped ref (`1a10132e` → `fb1c5c10`).

**Both previous blockers are gone**, confirmed in run [33953741307](https://github.com/windmill-labs/windmill/actions/runs/33953741307):
- the EE `seats_consumed` arity break — `src/ee.rs:1382` is now `seats_consumed(&c)`;
- `relock_noop` — all three tests pass (`relock_waiting_on_a_deploy_requeues_for_its_successor ... ok`).

**`cargo_test` is still red, on exactly one test, and it is main's:**

```
test_mcp_preprocessor_receives_the_callers_headers
  windmill-api-integration-tests/tests/mcp_preprocessor_headers.rs:160
  assertion `left == right` failed: preprocessor event kind:
  {"error":{"name":"ExecutionErr","message":"execution error:
   Invalid requirements, expected to find //bun.lock split pattern in reqs. Found: ||"}}
```

That is the only failing test binary in the run — everything else passes. The same single failure, same assertion, same message, is on `main` itself:

| run | branch / sha | result |
|---|---|---|
| [33952526800](https://github.com/windmill-labs/windmill/actions/runs/33952526800) | `main` @ `b5ae12bdf2` | same one failure |
| [33953713570](https://github.com/windmill-labs/windmill/actions/runs/33953713570) | `dbt-profiles-env-secrets` | same |
| [33953741307](https://github.com/windmill-labs/windmill/actions/runs/33953741307) | this PR | same |

It was masked until now: before `b5ae12bdf2` the suite died at the EE compile error, so this test never ran. The lock the preprocessor script is handed is empty (`||` are the message's own delimiters), so `bun_executor` refuses it — the fixture is not producing a bun lock. Nothing in this PR touches bun, locks, MCP or preprocessors; the diff is dbt plus the `clear_dbt_run_state` → `clear_dbt_script_state` rename.

This one needs an owner on the main side. Everything this PR is responsible for is green.

rubenfiszel:
✅ Review round clean @ db37ff2f8f3844add61a2b9e68d6b06ca606b99d — verdicts at 7e900a5417edcf88f92089f097c9d07fd1ca5d21 were Codex: good to merge, Pi: good to merge, Claude: nit-only (a miscounted byte length in the worked example of the key encoding, and the encoded key reaching the user in the "nothing published yet" refusal); both addressed in db37ff2f8f3844add61a2b9e68d6b06ca606b99d — the refusal now reads `warehouse `main`, target `prod`, relations in `dbt_wh_defer.analytics``.

rubenfiszel:
Both addressed in `588417d505`.

**[Codex P1] hash deletion and the conditional cleanup are now one transaction.** Confirmed the read: `delete_script_by_hash` ran its `UPDATE script SET … deleted = true` with `.fetch_one(&db)` — the pool, so committed — while every cleanup below it, including `clear_dbt_script_state_if_path_retired`, ran in `tx`. In that window the path has no live version and a concurrent deploy can take it; the `NOT EXISTS` guard then sees the *new* script and keeps the deleted project's state for it. The update moves to `.fetch_one(&mut *tx)`, which is exactly what `archive_script_by_hash` twenty lines above already does.

That moves the statement from the superuser pool to the RLS-scoped `user_db` transaction, so I exercised the route rather than reasoning about it — `POST /scripts/delete/h/<hash>` on the running instance returns 200 and the row lands `deleted=t, archived=t`.

**[Claude / cubic] the retirement guard was pinned by nothing.** Correct, and it is the finding I would have wanted flagged: `environment_state_follows_the_script` moved the only row to `MOVED` and *then* called the conditional clear on `PATH`, so the assertion afterwards could not fail whatever the predicate said. That step is gone, and `state_goes_only_once_no_live_version_is_left` pins both directions against a real database — a second live version keeps the state, the last one leaving takes it.

I checked the test actually bites rather than assuming: inverting `NOT EXISTS` to `EXISTS` in `clear_dbt_script_state_if_path_retired` fails it (`test result: FAILED. 0 passed; 1 failed`), and it passes again once restored.

Also in this push: the `origin/main` merge that resolves the `docs/dbt-runtime.md` conflict. Main's #10978 took decision **25** for "Cascade direction", so deferral is renumbered **26**; the two `decision 25` references in `dbt_manifest.rs` and `assets.rs` are #10978's own and still point at the right row.

rubenfiszel:
Addressed in `499b6a826d`.

**[Codex P1] `archive_script_by_path` fixed — and I swept the class rather than the instance.** This is the second round finding the same shape, so I checked every route in `scripts.rs` that clears dbt state against the write above it:

| route | write | before |
|---|---|---|
| `archive_script_by_path` | `UPDATE script SET archived` | `&db` — **fixed** |
| `archive_script_by_hash` | `UPDATE script SET archived` | `&mut *tx` ✓ |
| `delete_script_by_hash` | `UPDATE … deleted = true` | `&mut *tx` (fixed last round) |
| `delete_script_by_path` | `DELETE FROM script` | `&mut *tx` ✓ |
| bulk delete | `DELETE … path = ANY($2)` | `&mut *tx` ✓ |

So `archive_script_by_path` was the last one, and the set is uniform now. Exercised on the running instance rather than reasoned about: with a state row planted for `u/admin/deferprobefus`, `POST /scripts/archive/p/…` returns 200, every version lands `archived = t`, and the state row is gone — archive and clear in one transaction.

**[Claude] the stale comments and the double `11.`.** Both real, and the first is my own rename's debt: two sites beside `clear_dbt_script_state` still called it the *retry* state, disagreeing with the third I had updated. All three now name both halves. The verification list's duplicate `11.` came out of the merge — #10978 inserted "7. Declared write" above it — so deferral is `12.`

rubenfiszel:
✅ Review round clean @ 499b6a826db51d645409578a60c772939594783c — unanimous at this head: Codex "Good to merge", Claude "Good to merge.", Pi "Good to merge." (cubic skipped).

rubenfiszel:
✅ Review round clean @ 27c8faaf635972626661f918fbc400adbebe1ea0 — nit-only verdicts at 2a9e3f7d44069ade79fdb97177fdd4f894dbe61c (Codex: good to merge; Pi: good to merge; Claude: nit-only, that the safety argument for the exempted empty selection left with the origin-based condition it justified); addressed in 27c8faaf635972626661f918fbc400adbebe1ea0, which restates it at the site and names the descriptor refusal it now depends on.

PR #10977, opened 2026-09-04
feat: ingest dbt column lineage and real column schemas from the engine's parquet index

rubenfiszel:
## Summary

Ingest and storage for dbt column-level lineage and real column schemas, from the engine's
own static analysis. `manifest.json` carries neither: what does is the parquet index an
engine writes under `dbt compile --static-analysis strict --write-index --write-lineage` —
`dbt.column_lineage.parquet` (column-to-column edges, each labelled `copy`, `mod` or `scan`)
and `dbt.node_columns.parquet` (every column of every node, typed and ordered, rather than
only the ones an author documented).

Opt-in per project (`column_lineage: true` in the descriptor), because strict analysis is a
stricter SQL dialect than the build's, so it is a separate pass with its own `--target-path`,
never a flag on the build. Off, nothing changes; on, a project that cannot be analyzed keeps
exactly the graph it had. 1.x has no such flag and degrades silently.

**Scope.** The endpoint that draws a column trace is a follow-up, on
`dbt-column-lineage-surface`. What ships here is the pass, the storage, the gating — and the
one user-visible half, `column_schema`, which rides the asset graph the details pane already
fetches. See "Why this is split" below.

## Changes

- **Worker**: a separate `dbt compile --static-analysis strict --write-index` pass under half
  the job's remaining wall clock, then a bounded parquet decode. The compile owns the job's
  semantics and may `Err` (a cancellation or the job's deadline must fail the run); reading
  the artifact owns none and cannot fail anything. A pass stopped by the output ceiling
  reports as *truncated*, not complete; an unreadable index is reported apart from a missing
  one, since the two send a reader looking in different places.
- **The pass takes the build's own `--full-refresh`.** `is_incremental()` branches on it, so
  an incremental model reading `{{ this }}` compiles its self-join — and any `ref()` inside
  that branch — only when the flag is absent. The pass previously used the descriptor's
  default and so could store lineage for SQL a full-refresh run never executed. The flag now
  comes from one place shared with the build, and a run that overrides it gets its own graph
  rather than standing as the version's.
- **Storage**: `dbt_column_edge`, keyed like `dbt_edge` (path, version, job) with the same
  composite FK to `script`, so a version's lineage dies with the version and a run's snapshot
  with the sweep. `dbt_node.column_schema` holds the typed column list, beside `columns`
  rather than merged into it — `columns` stays what the author *declared*.
- **`column_schema` on the asset graph**, gated like the model's SQL: a column-level view is
  the shape of what the author wrote, so a share-link viewer entitled to a dbt run gets its
  relations and `ref()` edges and neither the SQL nor the columns.

## Why this is split

The API surface kept generating review findings, and the last one is why it moved rather than
getting a fourth fix here: a dbt selection's trace stops at the first project boundary, and
fixing it means expanding owners to a fixpoint. Today the caller's gate is decided **once**,
for the owners of the asked-for relation — that is the invariant the authorization design
rests on. Expanding transitively pulls in edges from projects the caller may have no grant
on, so the gate has to be re-applied to each newly discovered project inside the expansion.
That is the same shape as the leak four reviewers independently caught in the pinned arm of
this PR, and it is the one class where being wrong is a data leak rather than a wrong diagram.
It wants its own review.

Nothing here depends on it: the edges are stored and waiting, and the typed column list is
already visible.

## Why a table, not `dbt_edge.column_lineage` JSONB

Hanging the links on the `ref()` edge beneath them would inherit its clone, prune, clear and
cascade paths for free. It does not work: **a model reading `{{ this }}` gets column lineage
from itself to itself, and `parent_map` has no self-loop**, because a model does not `ref()`
itself — those pairs have no `dbt_edge` row to attach to. Measured, not hypothetical: an
incremental selecting from `{{ this }}` (`coalesce(p.dbl, s.dbl)`, `p.up as prev_up`) makes
Fusion emit `inc2.up → inc2.prev_up` with kind `copy`, for a node pair `dbt_edge` has no row
for. Under the JSONB shape that edge is silently dropped.

## Behaviour verified against the real engines

- **A failed strict pass still writes the index**, holding every edge of the models that did
  analyze — so the artifact is read whatever the exit status, and partial lineage is normal.
- **The flag is not the capability.** `dbt-core` 2.0.0-alpha.5 accepts `--write-index` and
  `--write-lineage`, declares views over both tables in its own `views.sql`, and writes
  neither parquet; only Fusion does today. An experimental adapter (postgres under
  `DBT_ALLOW_EXPERIMENTAL_ADAPTERS`) turns static analysis off and says so only in a warning
  on an otherwise *successful* compile — so the job log carries the engine's stderr whenever
  no index appears.
- **`scan` is stored but never served.** It means the column was read to produce the ROW
  rather than the value (a join key, a predicate, a `group by`), so it reaches every output
  column of its model. Keeping it in the table lets the follow-up's "show indirect" view ask
  for it without redeploying every project.

## Screenshots

**There is no column-lineage view in this PR, and the screenshot is not one.** The edges are
ingested and stored; what renders is the column *schema*. The trace that draws the edges is
the follow-up.

What changes is the dbt editor's model panel (`/scripts/edit/<path>` → Models → click a
model). Before, it could only list the columns an author wrote into `schema.yml`; for the
jaffle project that is four columns across four models, untyped. After, it lists every column
each model actually produces, typed and in emission order — fifteen — with descriptions still
coming from the declared `columns` where they exist:

| model | before | after |
|---|---|---|
| `customers` | `customer_id`, `full_name` | `customer_id Int32`, `full_name Utf8`, `doubled_spend Int32`, `order_total Float64`, `order_count Int64` |
| `stg_customers` | — | `customer_id Int32`, `first_name Utf8`, `last_name Utf8`, `lifetime_spend Int32` |
| `stg_orders` | — | `order_id Int32`, `customer_id Int32`, `amount Int32` |
| `big_spenders` | — | `customer_id Int32`, `full_name Utf8`, `order_total Float64` |

The pipeline page's asset pane shows a dbt relation's SQL but has no columns section, so it
is unaffected; adding one there is a natural piece of the follow-up, which already touches
that pane.

pr1-column-schema

## Test plan

- [x] `cargo test -p windmill-common --test dbt_graph_storage` — storage, prune, clear and
      sweep paths carry the column edges (12 tests).
- [x] `cargo test -p windmill-api-assets --test dbt_pinned_graph` — the pinned graph's
      authorization and the `ref()` edges it resolves (5 tests).
- [x] `cargo test -p windmill-worker --lib dbt` — including the new case pinning that a run
      overriding `full_refresh` gets its own graph, and that the form echoing the
      descriptor's own value is not read as an override. Verified to fail without the fix.
- [x] `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features` and
      `-p windmill-api-assets --all-targets`; `svelte-check` clean.
- [x] Exercised on a running instance with a real dbt Fusion project (jaffle shop, postgres
      adapter): a real `dbt build` succeeds, so the pass's separate `--target-path` leaves the
      build's own `run_results.json` intact; the deploy ingests 32 direct edges and typed
      schemas for four models; the details pane renders them.


rubenfiszel:
Round 1 addressed in `15300f854e` — all three verdicts' findings, plus the two local pre-push reviews. Summary of what changed and why:

**Codex P1 (deadline)** and **Claude P1/Codex P1 (unbounded decode)** are both real and both fixed at the root rather than papered over. `run_capturing` is split: `run_captured` returns the exit status *in the value*, so the pass downgrades only a failed **compile** — a cancellation, the job's deadline or the output ceiling now propagate and fail the job, which is what they always should have done. And the pass may spend at most half the job's remaining wall clock, so an optional annotation can no longer starve the `dbt build` that follows it. The parquet is decoded a row at a time with `MAX_COLUMN_EDGES` enforced *during* the decode.

**Claude P1 (`dbt_column_edges` unbounded on a polled endpoint)**: the API now serves `copy`/`mod` only, under a `LIMIT`. Your observation that `scan` also crowds the 200k storage cap was the sharper half — `attach_column_index` now orders direct kinds first, so the cap gives up `scan` before it evicts the lineage the trace draws. `scan` stays in the table so a future "show indirect" view needs a query change rather than every project being redeployed.

**Claude P1 (fork)**, found by the local pre-push review: `clone_dbt_graph` copied every other dbt sidecar table and not this one. Worse than it looks — the cloned digest covers the column edges, so a dynamic run in the fork matched it and stored nothing, leaving the lineage gone until a redeploy. Fixed, along with the missing `column_schema` in its `dbt_node` column list.

**`column_schema` gating**: Pi considered and dismissed this, Claude flagged it; I've taken Claude's reading. `columns` gates the *declared* names, so leaving `column_schema` open meant enabling the pass widened what a share-link viewer sees. Both it and the column edges now take the same gate as the model's SQL. `dbt_edges` stays ungated: it draws relations the caller already sees in `asset`.

**Digest drift** (Pi dismissed as benign, Claude flagged): fixed anyway, since it was one line — the section is hashed only when non-empty, so a project that never asked for the pass keeps the digest it has.

**Pi's telemetry nit** is addressed in the PR description rather than in code: `AGENTS.md` asks for a *proposal* the reviewer can keep or drop, and step 1 of the recipe registers the pair in `feature_usage_ee.rs`, which would pull an EE companion PR into an otherwise CE-only change. The concrete vocabulary is written out there.

Also: `--target-path` no longer relies on beating `DBT_TARGET_PATH` (both are set), `run_capturing`'s doc comment is back on `run_capturing`, and `summarized_schema.txt` has the new table and column.

Re-verified on a live instance after the fixes: deploy still ingests 32 edges, the graph endpoint now returns 9 (`copy`/`mod` only), the trace still renders in both the dbt editor and the pipeline pane. 12 storage + 16 unit + 10 frontend tests pass; `SQLX_OFFLINE` check clean across all four touched crates.

rubenfiszel:
Round 2 addressed in `c0aeed2d18`. All four findings were real, and all four were regressions from the round-1 fixes — worth saying plainly, since two of them undid the thing the round-1 fix was for.

**Editor buffer trace (Codex #1, Pi P1)** — the sharpest catch of either round. I copied `script_visible` from the node query and dropped its `script_hash IS NULL` arm, so `sc.hash = NULL` was never true and every version-less row was filtered out: the editor rendered typed columns and no lineage, in exactly the pane the PR screenshots. Fixed with the same arm the node query has, and now pinned by an assertion in `dbt_pinned_graph.rs` — the file that exists for this class of bug, which I should have extended in round 1 rather than after being told. Verified on the running instance: a Refresh models on the buffer now shows "parsed from the editor at …" *and* its trace.

**Phase budget (Codex #2)** — correct, and my round-1 fix was the wrong shape. Handing a shortened deadline to the poller made my artificial budget indistinguishable from a real cancellation, so a slow-but-valid analysis aborted the build it only annotates. The runner gets the full remaining deadline back (cancellation and the job's own deadline must still fail the job) and the budget is now a `tokio::time::timeout` race around the whole pass, where expiring is my budget and nothing else. Dropping that future kills the child, which `run_captured` already spawns with `kill_on_drop`.

**Decode cap (Codex #3, and Pi's "the sort is a no-op" nit is the same fact)** — right on both counts. The cap counted parquet rows before any filtering, so `scan` and out-of-graph rows could spend it before a drawn edge was read, and the later sort could then only reorder what survived. It now counts what the closure *keeps*, runs direct kinds in a first pass, and is handed the graph's own node set so the budget can't go on rows that could never be stored. The sort in `attach_column_index` stays for the caller that doesn't come through the reader — the agent-worker wire — and its comment says so now.

**Digest (Codex #4)** — also right, and my round-1 fix didn't work: `column_schema: None` still serialized as `"column_schema":null` inside the nodes, so every digest changed regardless. `skip_serializing_if` on that field is the actual fix.

Pi's other nit — `collect`'s doc still claiming "every failure returns `None`" — is now false and corrected: the doc distinguishes what degrades (wrong engine, failed analysis, missing artifact, budget) from what fails the job (cancellation, deadline, output ceiling).

33 backend tests + 10 frontend pass; `SQLX_OFFLINE` check clean across the four touched crates; deploy and buffer-parse both re-exercised on a live Fusion project.

rubenfiszel:
## Stopping here: this wants a different shape, not a fourth round of fixes

Round 3 came back Pi "Good to merge", Codex "should address" (2×P1), Claude 3×P2. But
counting verdicts is the wrong way to read it. **Every round-3 finding is a consequence of
a fix from round 1 or 2, not a defect the feature had on its own:**

| Round-3 finding | Introduced by |
|---|---|
| P1 the `spawn_blocking` decoder outlives the phase timeout | round 2's timeout fix |
| P1 `LIMIT 20000` runs before the Rust auth/render filters | round 1's cap meeting round 1's gating |
| P2 decision 14 contradicts itself — "the pass never fails a run" vs three named ways it does | round 1's text vs round 2's |
| P2 the 1 MB stdout ceiling is a *pass* condition wearing *job* error clothes | round 2's error split |
| P2 the sort's comment names the agent-worker wire as "a caller that does not come through the reader" — it does | round 2's comment |

Three of those are cases where the **justification written for the previous fix is itself
wrong**. And the findings land in the same three files every round: `dbt_column_index.rs`,
`windmill-api-assets/src/lib.rs`, `dbt_manifest.rs`. `AGENTS.md` names this exactly — "each
review finding fixing an assumption the previous fix broke" — and says to stop and re-derive
rather than patch again. So: no round-4 fixes. Here is what the simpler shape is and what
switching costs.

### Seam 1 — column edges belong *on* `dbt_edge`, not in a table of their own

`windmill-api-assets/src/lib.rs` produced a finding in all three rounds: unbounded response,
ungated `column_schema`, missing `script_hash IS NULL` arm, `LIMIT` before filtering. They
are one mistake wearing four hats — I wrote a **fourth near-copy** of the `live`/`chosen` CTE
query, and every invariant the existing three encode had to be re-derived by hand. I got
several wrong, and the ones I got right I got right by copying.

A column edge is precisely *the column-to-column links between one parent/child node pair* —
which is what a `dbt_edge` row already is. As `dbt_edge.column_lineage JSONB`
(`[{from, to, kind}]`):

- **the fourth query disappears.** The existing `dbt_edge` query returns them in its
  `SELECT`, already carrying the NULL-hash arm, the pinned/unpinned scoping, the folder gate
  and the `resource_type <> 'test'` filter.
- **the separate `LIMIT` disappears**, so applying it before authorization becomes
  impossible — the edges returned are already the bounded set.
- **the fork/prune/clear gaps disappear.** `dbt_edge` is already in `clone_dbt_graph`,
  `prune_dbt_run_graphs`, `clear_dbt_manifest_version` and the cascade. The round-1 fork bug
  could not have happened.
- **the gating decision disappears** — it inherits the edge's.

That is three of the fourteen findings so far made *structurally impossible* rather than
fixed.

The load-bearing assumption is that a column edge never connects a node pair that has no
`ref()` edge. Checked on the probe project: 4 distinct column-edge pairs, 4 `parent_map`
pairs, identical sets. One probe is not a proof — an ephemeral chain or a `scan` through a
source is where I would look for a counterexample, and that check should come before
committing to this.

### Seam 2 — `dbt_column_index.rs` is three contracts in one function

Five findings across three rounds (unbounded decode, swallowed cancellation, wrong budget
mechanism, cap before filtering, detached decoder) plus two P2s about its own comments. The
function is simultaneously a subprocess runner with job-lifecycle error semantics, a
memory-bounded decoder, and a best-effort degrader — so every fix to one perturbs another.
The timeout fix *created* the detached-decoder bug; the stdout ceiling is a pass-level
condition that can only be reported as a job-level error.

Split by error contract instead:

- `compile_index(...) -> Result<CompileOutcome, Error>` — only a cancellation or the job's
  deadline can `Err`; the exit status **and** the output ceiling are values on
  `CompileOutcome`. The phase budget wraps this and only this, so nothing survives it.
- `read_index(dir) -> ColumnIndex` — infallible, bounded, no job semantics at all, no
  timeout to outlive.

### What switching costs

The branch is unmerged, so there is no data migration — but it is a replacement migration, a
rewrite of the ingest to group edges by node pair, deleting the new API query and extending
the existing one, reshaping the frontend type and its test, regenerating the sqlx cache, and
**three rounds of review discarded**. Call it a day, plus review restarting from zero.

### What is *not* in question

The probe findings and the feature's premise survive either shape, and they are the
expensive part of this PR: that the index exists at all, that `dbt-core-2x` has the flags and
writes nothing, that `--static-analysis strict` is a stricter dialect that still writes a
partial index, that `lineage_kind` has three values and an open set, that an experimental
adapter disables analysis silently. So does the descriptor gate, the degradation behaviour,
and the rendering — which is verified working on a real Fusion project, including the
editor-buffer path.

### Recommendation

Reshape seam 1 and seam 2, then re-review. I have not marked this ready and have not landed
a round-4 patch. If you would rather take the current shape and accept the two open P1s as
follow-ups, say so and I will fix them — but I do not think a fourth round converges, and
the third round is the evidence.

rubenfiszel:
Reshaped in `542135f52b`, per the decision to take both seams. What changed and what the counterexample hunt settled:

**The JSONB idea is dead, and the PR description now records why** so the next person can see it without re-running the probe. A self-referential incremental produces column edges for a node pair `dbt_edge` has no row for — `parent_map` has no self-loop, because a model does not `ref()` itself. I pushed past the easy version of this: a watermark (`where id > (select max(id) from {{ this }})`) yields only `scan`, which we discard anyway, so I built an incremental that *selects* from `{{ this }}` and got **`inc2.up → inc2.prev_up`, kind `copy`** — a drawn edge meaning "this column carries the previous run's value". JSONB would drop it silently on incremental models. Two things the same hunt confirmed safe: Fusion contracts column lineage through ephemerals exactly as our ingest contracts `parent_map` (checked on a two-level chain), and a source reached only via a join predicate and a `where` appears in both.

**Seam 2 — the worker pass, split by error contract.** `compile_index` owns the JOB's semantics: only a cancellation or the job's deadline `Err`s, while a non-zero exit, the output ceiling and the phase budget are outcomes on a `Compiled`. `read_index` owns the ARTIFACT's: infallible, bounded, no job semantics. The budget wraps the compile **alone**, which is what kills round 3's detached-decoder P1 at the root — a decode can no longer outlive the timeout that said the build would get the rest. The stdout ceiling also becomes a value (`Overflow::Truncate`) rather than a job error, which was the other P2: it is the pass's condition, not the job's, and `dbt show` — whose output *is* its answer — keeps `Overflow::Fail`.

**The CTE duplication is gone.** Both kinds of edge now come from one statement over a `UNION ALL`'d edge source, so the `live`/`chosen` CTEs *and* the version and editor-buffer join conditions — the part that actually bit, when the `script_hash IS NULL` arm went missing — exist once. The union is at the source rather than a join precisely because of the self-pair above.

**Round-3 P1 #2 fixed structurally.** The scope filter moved into SQL via the existing `ScopePathFilter` (whose own doc names this failure mode), and the cap is a `LIMIT` on the column branch after the scope filter, the visibility check and the graph joins — so no row a caller may not read, and none outside the graph on screen, can spend it.

Also fixed: decision 14 contradicted itself about whether the pass can fail a run (it now states plainly what degrades and what propagates), and the `attach_column_index` sort comment named the agent-worker wire as a caller that skips the reader, which it does not.

33 backend tests + 10 frontend pass; `SQLX_OFFLINE` check clean across four crates; deploy, editor-buffer refresh and a real `dbt build` all re-exercised on a live Fusion project, with 9 `ref()` edges and 9 column edges (`copy`/`mod` only) served and the trace rendering.

rubenfiszel:
## Round 4: the same signal, a second time. Stopping again rather than patching.

Round 4 is Pi "Mergeable, nits", Codex "Should address" (2×P1, 2×P2), Claude 2×P2. Pi calls
the reshape "a real improvement over round 3" and it is — the error-contract split and the
unified query are both verified correct by two reviewers independently. But the test that
matters is not the verdict line, it is whether the churn *stopped*. It did not.

**Five findings. All five in the three files that produced findings in every previous round.
Four of five are consequences of my own earlier fixes.**

| Round-4 finding | File | Introduced by |
|---|---|---|
| P1 response cap precedes `rendered` and dedup | `windmill-api-assets/src/lib.rs` | round-1 cap, third time at this seam |
| P1 decode outside the budget, two full passes over the file | `dbt_column_index.rs` | round-2's two-pass cap fix, kept by round 4 |
| P2 truncated compile reported as complete | `dbt_column_index.rs` | round-4's own `Overflow::Truncate` fix |
| P2 decision 14 *still* self-contradictory | `docs/dbt-runtime.md` | round-3's finding, my round-4 fix edited one of the two paragraphs |
| P2 cap comment overstates `script_visible` | `windmill-api-assets/src/lib.rs` | round-4's own cap placement |

Claude's wording on the doc one is the tell: *"the paragraph round 3 flagged is still here, and
this commit made it wrong in a new way."* And I found the `script_visible` gap myself in
self-review before the verdicts landed — which is corroboration, not mitigation: it is the same
seam producing the same class of defect for the third round running.

### What the recurring seam actually is

Every cap finding — round 1 (unbounded), round 3 (cap before authorization), round 4 (cap
before `rendered` and dedup, cap before visibility) — is the same shape: **the column-edge
result is assembled by a pipeline whose filters are split across SQL and Rust, and the bound
sits somewhere in the middle.** Each round I move one more filter to the correct side of the
bound, and the next round finds the next filter that is still on the wrong side. `rendered` and
the dedup cannot move into SQL without re-deriving in SQL the asset set that a *different*
query built — which is the duplication problem that caused the other half of the churn.

The bound exists at all only because column lineage rides on the folder-wide asset-graph
response — the one a run page polls — while it is *rendered* for exactly one selected asset at
a time.

### The shape that dissolves it

**Serve column lineage from its own endpoint, keyed to the selected asset**, rather than as a
field on the graph response:

- No cap is needed, so there is no bound for filters to be on the wrong side of. The result is
  naturally bounded by one asset's columns.
- The run page's poll stops carrying it.
- `rendered` stops existing as a concept here: the caller asks about an asset it is already
  rendering.
- Scope and visibility apply to one script's edges, checked once, with nothing to order them
  against.

The other two seams stay as they are — the error-contract split and the storage decision are
both confirmed sound by this round, and the counterexample that killed the JSONB idea still
holds.

**Cost:** a new endpoint plus its openapi and client regen, the frontend fetching on selection
instead of reading the field (so the trace gets a loading state), removing `dbt_column_edges`
from the graph response, and a fifth review cycle. Roughly half a day.

### The honest alternative

Land it as it stands with the two P1s fixed narrowly — move `rendered`/dedup into the query as
best I can, bound the decode — and accept that this seam will keep producing findings whenever
someone touches it. That is a legitimate choice if column lineage on the graph endpoint is
wanted sooner rather than better; the feature *works*, and every reviewer agrees the rest of
the change is sound.

I am not choosing between these. Four rounds in, with the same seam failing in the same way
each time, the call is yours. I have landed nothing since `542135f52b` and am holding a small
uncommitted fix (the `script_visible` gate moved into SQL, which addresses Pi's P2 and part of
Codex's #2) rather than adding another layer while the question is open.

rubenfiszel:
Round 7 verdicts on `fcdcad1810` — codex: *should address* (1×P1); claude: *mergeable, nits* (3×P2); pi: *mergeable, nits* (1×P2). All three name the same core issue, and every finding is in the frontend seam (`pipeline/[folder]/+page.svelte`, `dbtColumnLineage.svelte.ts`) — none in the backend files that produced findings in earlier rounds.

Fixed in `fab2b93299`, before flipping to ready:

- **Every dbt boundary a selection reaches is expanded, not the first.** A ducklake table derived from two unconnected dbt relations now issues one request per boundary and merges the components; previously the second rendered as a leaf. Verified live: selecting such a table produces two `assets/column_lineage` requests and both upstream chains render.
- **The component cache only skips a relation it actually asked about.** Claude's point: a relation two projects describe has an owner row in each, so a component fetched for one carries it as an endpoint without the other's half — treating "appears in the graph" as "resolved" would hide exactly the cross-project edges the server's relation-keyed walk exists to merge. (Pi dismissed this one as safe on equivalence-class grounds, which holds within a single project but not for the two-owner case; taking the stricter reading costs some caching and no correctness.)
- The producer-graph comment that still said "gated to a ducklake selection" after the predicate was widened to `ducklake || dbt`.

Flipping to ready per instruction rather than on a clean round: codex's verdict on `fcdcad1810` was blocking, so the automatic `ready_for_review` review will run. The one gap both Pi and I would flag for a human: `read_index_blocking`'s budget logic has no unit test — it needs a synthetic parquet with a filled direct bucket followed by a scan tail, a heavier fixture than the rest of the PR carries, and the decode path is covered end-to-end against a real Fusion engine instead.

PR #10978, opened 2026-09-04
feat: let `// materialize` declare a `dbt://` warehouse-relation write

rubenfiszel:
## Summary

The shared asset graph was a picture in one direction: a native script could *read*
a dbt mart (the body parsers resolve `"dbt://…"` from a string literal) but nothing
outside dbt could *declare a write* to a warehouse relation — `// materialize`
accepted DuckLake targets only. So an ingestion script and the dbt project consuming
its output were two disconnected pictures, and `docs/dbt-runtime.md` refused a
`# on dbt://<mart>` subscription outright at deploy on the reasoning that an edge
which can never fire is worse than an error saying so. That reasoning was sound; this
PR removes its cause.

Identity is untouched — the physical relation, never the producing tool — so the
ingestion script and the dbt model reading that relation land on the same graph node.

## Changes

- **`// materialize manual dbt://<warehouse>/<schema>/<name>`, from any language.**
  `manual` is not a mode here but the only mode: nothing generates warehouse DDL, so
  the script issues its own write and Windmill records the outcome. The non-`manual`
  spelling is refused rather than silently degraded, so nobody writes it expecting
  generated DDL. The target must be a full three-segment relation, and `<warehouse>`
  is resolved against the workspace's configured warehouses for the same reason a
  descriptor's `profile.warehouse` is — a name no warehouse answers to strands the
  write on a node nothing else reaches. The DuckDB-only rule stays, scoped to
  `ducklake://` targets.
- **The write records.** `record_declared_warehouse_write` (windmill-worker
  `worker.rs`) lands the same `materialized_partition` row a DuckLake target does —
  `materialized` / `failed` with the error, keyed by the resolved `partition` when the
  script is `// partitioned`. It sits in the generic job path rather than an executor
  because the DuckLake write *engine* is DuckDB's while this declaration is anyone's.
  Agent workers post it through the existing job-scoped API route.
- **Neither annotation is accepted on a dbt script.** A project's writes come from
  its manifest and its graph ingest republishes that path's trigger and asset rows
  wholesale, so either annotation would deploy something the dependency job then
  silently removes.
- **The `# on dbt://` refusal is narrowed, not removed.** Refused only when *every*
  script that writes the relation is a dbt one (`sole_dbt_producer`) — the one shape
  that still cannot fire, since `asset_dispatch` continues to return early for
  `ScriptLang::Dbt`. "Nothing produces it yet" is deliberately not that shape: a
  subscriber may be deployed before its producer, as for every other asset kind, and
  refusing there would break deploy-order-independent syncs. A dbt script may not
  subscribe at all — its graph ingest clears its own `dbt://` trigger rows, so
  accepting one would deploy an edge the dependency job then silently removes.
- **The ordering deploy cannot catch is logged.** A subscription accepted while
  nothing produced the relation, then claimed by a dbt project deployed afterwards, is
  named in that project's own deploy log (`dormant_dbt_subscriptions`) rather than
  left silently dormant.
- **AI guidance.** The pipeline copilot prompt (`copilot/chat/pipeline/core.ts` and
  `system_prompts/base/pipeline-base.md`, plus the regenerated bundle) told the model
  `// materialize` was DuckDB-only and rejected on any other target, which would have
  steered users away from this feature. It now distinguishes the managed DuckLake
  write from the warehouse-relation declaration.
- **Docs.** `docs/dbt-runtime.md`: decision 11 no longer claims the scheme names an
  exclusive producer; "No cascade from dbt" becomes "No cascade *from* dbt" with a new
  subsection on declaring the write and which subscriptions are refused; decision-log
  row 25 records the amendment. `docs/ducklake-materialization.md`'s "`// materialize`
  is DuckDB-only" bullet is scoped to `ducklake://`.

The forward direction — dispatching *out of* a finished dbt run — is unchanged and
stays a separate item: a run's `select` can build any subset of the project, so it
needs a per-run write set to be correct, which the per-relation state table cannot
supply.

## Known limitation

The record is written from the normal execution path, which a job handed to a
**dedicated worker or a flow runner** never enters — exactly as those bypass
`// partitioned` resolution and the `WM_PIPELINE` env today. Such a run performs its
write and still cascades (the fan-out reads the deploy-time `asset` rows) but records
no row, so the relation shows no last writer. Closing it is one change for every
runtime pipeline annotation, not this one; stated in the function doc and in
`docs/dbt-runtime.md`.

- **The canvas draws the edge.** The pipeline editor suppressed every
  `// on dbt://…` overlay, which was right while the deploy refused all of them;
  it would now hide the author's own annotation for exactly the case this PR adds.

## Screenshots

The canvas change: the pipeline editor with an unsaved `# on dbt://main/analytics/agent_rel`
in the buffer (right) and the arrow it now draws (left) — `agent_rel` appears with a
dashed unsaved trigger edge into `f/dbtdemo/consumer`, while `raw_orders` loses the
edge the buffer no longer subscribes to. Before this PR the overlay was suppressed
for every `dbt://` kind, so the author's own annotation drew nothing.

pipeline canvas draws an unsaved dbt:// subscription

The other `frontend/` change is a copilot prompt string (`chat/pipeline/core.ts`),
which has no visible UI.

## Test plan

Exercised on a running instance (postgres warehouse, dbt-core 1.12 / dbt-postgres):

- [x] A `postgresql` script with `-- materialize manual dbt://main/analytics/raw_orders`
      creating and filling that relation, a dbt project whose model reads it as a
      `source`, and a subscriber on `dbt://main/analytics/raw_orders`. The graph
      response and the assets UI show one lineage — `ingest → raw_orders → matdemo →
      orders_summary` — with `raw_orders` carrying the script as Write and the project
      as Read on one node; the pipeline canvas draws the same chain (dbt runnable
      hidden, as designed).
- [x] Running the ingestion script lands `materialized_partition (dbt,
      main/analytics/raw_orders, materialized, <job id>)` and dispatches the
      subscriber — visible in the run page's Dispatch table with the child job. A run
      that fails lands `failed` with the executor's error.
- [x] The dbt project then builds `orders_summary` off the relation the script wrote.
- [x] Deploy-time: subscribing to `dbt://main/analytics/orders_summary` (dbt-only
      producer) is refused naming the project; subscribing to the natively-produced
      relation and to an unproduced one are both accepted; non-`manual`, a two-segment
      target, an unconfigured warehouse, an `s3://` target, a `ducklake://` target from
      a non-DuckDB script, and a dbt script subscribing are each refused with their own
      message.
- [x] Deploying a dbt project that claims a relation an accepted subscriber was
      waiting on logs `dbt://main/analytics/late2 → u/admin/on_future2` in the
      dependency job.
- [x] `// data_test` beside a `dbt://` target is refused at deploy (nothing would
      run those probes on a warehouse relation).
- [x] A dbt script is refused both annotations, while a descriptor whose header
      carries ordinary prose about materialization is not a false positive.
- [x] `cargo test -p windmill-common -p windmill-parser -p windmill-types`;
      `cargo test -p windmill --test asset_trigger_dispatch --features
      quickjs,enterprise,private,license,scoped_cache` (9/9).
      `backend/windmill-common/tests/dbt_producer_rules.rs` pins both spellings of
      the sole-producer rule against each other;
      `backend/windmill-api-integration-tests/tests/dbt_materialize_target.rs` pins
      the deploy contract and the canonical key the three sides have to agree on.

`cargo_test` is red on this branch and on `main` itself for the same two
`relock_noop` tests (`worker timed out` at `windmill-test-utils/src/lib.rs:384`,
e.g. main run 33884307602, before any dbt change) — a known pre-existing flake, not
this PR.

- [x] The agent-worker (`Connection::Http`) branch, which a plain `cargo run` never
      takes, on a real agent worker (`docs/agent-worker-e2e.md`: EE build,
      `MODE=server` on its own port, `MODE=agent` serving the script's tag). A
      `# materialize manual dbt://main/analytics/agent_rel` bash script ran on
      `ag-agentgrp-…` and landed `(dbt, main/analytics/agent_rel, materialized,
      <job id>)` through the job-scoped HTTP route; a failing sibling landed
      `failed` with the exit status.

## Not included

No `feature_usage` counter: this is an annotation with no UI affordance, and `dbt`
carries no registered feature vocabulary yet, so instrumenting it would be its own
change (plus an EE companion for `FEATURE_USAGE_KINDS`).

---


rubenfiszel:
On the concurrent-ingest race (Codex, head `f7f79d45`): the interleaving is real and I've recorded it rather than serialized it.

The window is a subscriber deploy whose producer check has run, a dbt ingest committing and running its warning query, and the subscriber's trigger row landing after both. Neither side then reports the edge. My `sole_dbt_producer` doc claimed the race only resolves toward refusing, which was wrong for that direction — corrected in `dff574d3` along with `docs/dbt-runtime.md`.

I did not take the suggested per-relation serialization. Both sides would have to share that lock, and `persist_ingest` takes `script … FOR UPDATE` (`dbt_executor.rs:3278`) *before* its own `pg_advisory_xact_lock` (`:3418`). A deploy that acquired relation locks first would invert that order and deadlock the API deploy path against the worker's ingest — a real, user-visible failure traded for a cosmetic dormant edge. A post-commit recheck was the other option; it cannot refuse anything by then and only reaches the server log, so it buys nothing the docs don't.

The exposure is bounded the same way as the other orphaning the deploy cannot catch (a native producer dropping its `// materialize`): the next deploy of that project emits the warning, and the canvas shows the edge meanwhile. Both are now written down together as accepted, with the deadlock reasoning, rather than left as something a reader would assume was handled.

rubenfiszel:
✅ Review round clean @ 3d5136e1572e13fd7044de923c00491f6b80bab6

All three reviewers returned **Good to merge** on this head (round 13).

`cargo_test` is red for a main-branch breakage, not this PR: the job fails to compile `windmill-api/src/ee.rs:1382`, where `seats_consumed(c.author_count, c.operator_count)` no longer matches the `seats_consumed(&ActiveSeatCounts)` signature that #10929 introduced. The identical E0061 is on main's own latest backend-test run. Before that landed, the same job was failing on the two `relock_noop` tests (`worker timed out` at `windmill-test-utils/src/lib.rs:384`), also reproducible on main run 33884307602. Either way the run aborts before it reaches this PR's crates, so the new tests never execute in CI — they were run locally instead:

- `cargo test -p windmill-common -p windmill-parser -p windmill-types` (incl. the 8 `dbt_producer_rules` cases)
- `cargo test -p windmill-api-integration-tests --test dbt_materialize_target`
- `cargo test -p windmill --test asset_trigger_dispatch --features quickjs,enterprise,private,license,scoped_cache` (9/9)
- `vitest run resolveGraph.test.ts` (46/46)

Every other check is green.

PR #10982, opened 2026-09-04
feat: report resource type picks to the hub and rank pickers by popularity

hugocasa:
## Summary

The Hub gained a pick counter for resource types (`GET <hub>/resource_types/<id>/pick` increments it, `GET <hub>/resource_types/picked` ranks on it), but nothing in Windmill fed it — the counter only ever saw picks made on the Hub site itself. This wires it up, and then spends the signal: the flow step picker listed ~216 integrations A→Z and the add-resource drawer ~273 resource types the same way, so the handful anyone actually uses sat wherever the alphabet put them.

Both lists now lead with what is popular on the Hub, then with what the workspace itself uses, then alphabetical. Every Hub signal is optional: a Hub that predates these routes, or a private one that has none, degrades silently to the local ordering, and a pick that cannot be recorded is dropped rather than surfaced.

Companion PR (the `picks` field this reads off `/integrations/list`): windmill-labs/windmillhub#200.

## Changes

**Backend** — three routes on the workspaced resources service:
- `POST /w/{ws}/resources/type/hub/pick/{name}` reports a pick. Windmill stores no Hub ids, so it resolves the name against a process-wide cache of `<hub>/resource_types/list` (keyed on the Hub base url, since that is a live instance setting) and then calls the Hub. It never fails the caller: an unreachable Hub, a Hub without the route, and a workspace-local type all return `{"success": false}` with a 200. POST and `resources:write` because it changes state on the Hub under the instance's own credentials — the sibling `/type/*` routes are metadata reads that a `resources:run` app-embed token may make, and `resource_metadata_route_allowed` matches that prefix, so a GET here would have let untrusted app JavaScript drive Hub counters through us.
- `GET /w/{ws}/resources/type/hub/picked` projects `<hub>/resource_types/picked` down to `[{name, picks}]`, returning `[]` on any failure so an older Hub reads as "no Hub signal". The Hub counts picks in a bigint, which its driver serialises as a string, so it decodes through `maybe_number`.
- `GET /w/{ws}/resources/type/resource_counts` counts the workspace's resources per type through `user_db`. Guarded by `resources:read`, unqualified — a count per type is aggregate, so there is no path to narrow it by, and a path-scoped token gets nothing rather than a total spanning paths it cannot read.

**Frontend** — new `pickerPopularity.ts` holds the two cached signals plus one comparator, in four tiers: **whether this workspace already holds a resource of the type**, then Hub picks, then the local count, then the name.

Used-here leads rather than merely breaking Hub ties because the counts are on incomparable scales — a Hub pick count is global and unbounded, a local one is usually single digits. Ranked the other way, local usage only sorts the slice where Hub counts are *equal*, which (they being distinct integers) is just the tail nobody has picked. That looks fine on a Hub with few picks and silently stops mattering as one fills up, so the ordering would drift away from the workspace's own stack with no change to this code. The Hub still decides within each half, so both signals are honoured without a weighting constant to tune.
- The flow step picker (`PickHubScriptQuick.svelte`, and `PickHubScript.svelte` behind the script editor's Library drawer) sorts its integration chips with it, reading `picks` off the `listHubIntegrations` rows and treating an absent field as zero.
- The add-resource drawer (`AppConnectInner.svelte`) sorts the browse list with it. The native-database-first grouping in "Others" and the search ranking are untouched: `sortResourceTypesByMatch` already owned the order while searching, so only the browse order changes.
- Saving a resource reports the pick.

Both signals resolve to nothing on an instance that has switched the Hub off, so a closed environment spends no request on `hub.windmill.dev` to order a list. The resting comparator is the alphabetical floor rather than a no-op: the OAuth connect list arrives in a server-side `HashMap`'s iteration order, so leaving it unsorted before the signals land would render it differently between processes.

`PickHubScript.svelte` carries a large whitespace-only re-indentation: the edit hook runs prettier and that file was not prettier-clean on `main`. `git diff -w` shows the real change is ~15 lines.

No `feature_usage` instrumentation: this is an ordering change plus a counter the Hub already measures directly.

## Screenshots

Driven against a Hub deliberately set up so the two tiers disagree: it ranks `airtable` highest of all (900 picks) and the workspace owns none, while the workspace holds `stripe` ×3, `slack`, `github`, `ably` and `postgresql`.

Flow step picker. The four the workspace uses lead — `slack` (Hub 500) and `github` (250) ordered by Hub picks, then `stripe` (Hub 0, 3 local) and `ably` (Hub 0, 1 local) — and the two it doesn't, `airtable` (900) and `gmail` (40), sit below them. Hub-first ordering would have opened on `airtable`:

step-picker

Add-resource drawer, same Hub. The native database group keeps its place (`postgresql` leads it, the workspace has one), then the used types Slack → GitHub → Stripe → Ably, then Airtable and Gmail, then A→Z:

resource-drawer

The step picker against a Hub with no `picks` field and no `/resource_types/picked` — the whole Hub tier collapses and it falls back to used-here then alphabetical, with no error:

step-picker-old-hub

## Test plan

- [ ] Point `hub_base_url` at a Hub that ranks a type the workspace does **not** own above ones it does; confirm both pickers still lead with the workspace's own, Hub-ordered among themselves, and place the Hub's favourite below them.
- [ ] Save a resource of a Hub type from the drawer; confirm that type's `picks` incremented on the Hub.
- [ ] Point `hub_base_url` at a Hub that 404s `/resource_types/picked` and `/resource_types/<id>/pick`; confirm both lists still render (local-then-alphabetical) and saving a resource still succeeds with no error surfaced.
- [ ] Create a few resources of one type and confirm it rises above equally-unpicked types in both lists.
- [ ] Turn on `disable_hub` and confirm opening the drawer and saving a resource make no Hub request at all.
- [ ] `npx vitest run src/lib/components/pickerPopularity.test.ts` from `frontend/`, and `cargo test -p windmill-store --lib hub_picks_tests` / `cargo test -p windmill-api --lib embed_scopes_allow_app_routes_and_deny_the_rest` from `backend/`.

---
## Summary by cubic
Sends resource type picks to the Hub and ranks the flow step picker and add-resource drawer by popularity instead of alphabetical order.

**Behavior**
- Both lists lead with types the workspace already holds, then Hub pick count, then local usage, then name.
- Local usage is counted per integration, mapping a type like `discord_webhook` to `discord` via the Hub's type-to-app listing.
- Saving a resource reports the pick to the Hub.
- Search keeps its existing relevance ranking; popularity only orders the browse lists.
- An unreachable Hub, a Hub predating the routes, or one disabled via `disable_hub` falls back to the local ordering with no error.

**Backend**
- Adds three routes: `POST /w/{ws}/resources/type/hub/pick/{name}`, `GET /w/{ws}/resources/type/hub/info`, and `GET /w/{ws}/resources/type/resource_counts`.
- Hub reads are cached; a failed index read is remembered for a minute and a successful one for an hour, so an unreachable Hub doesn't cost a timeout on every picker open.
- The pick route is a POST scoped to `resources:write` because it changes Hub state under the instance's credentials; an embed-token test asserts that path is denied.
- The `PickHubScript.svelte` diff is mostly whitespace; `git diff -w` shows the real change.

<sup>Written for commit eba033c5a911d6b777d758d59459f9fcf0305f10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hugocasa:
Round 1 addressed in 078c8d4f1c.

**Codex P1 — local usage ignored resource type → app mappings.** Real, and a consequence of the tier change in the previous commit: with used-here now the *leading* tier, a name mismatch stopped being a tiebreak miss and started deciding which half of the list an integration lands in.

Fixed at the source rather than with a lookup table. The backend already cached `/resource_types/list` to resolve pick ids; it now keeps each type's `app` too, and `GET .../resources/type/hub/picked` became `GET .../resources/type/hub/info`, returning `{name, app, picks}` for every hub type. The frontend totals local counts per integration through that mapping, falling back to the type's own name when the hub has none — which is also what an unreachable hub and a workspace-made `c_` type leave every entry with. The two hub reads degrade independently on purpose: a hub with no `picked` route still supplies the mapping, which is the part that decides tier membership.

Verified against a stub hub mapping `discord_webhook`/`discord_bot_configuration` → `discord` and `ms_teams_webhook` → `msteams`, in a workspace holding one of each and no `airtable`: the picker now leads `discord`, `msteams`, then `airtable` (900 picks) and `zulip`. Before the fix `airtable` led and neither Discord nor MS Teams reached the used tier. Regression test added as Codex asked (`totalLocalCountsByApp`), covering the mapping, the several-types-one-integration case, and the unmapped fallback.

**Codex P2 — stale ordering comment** at `AppConnectInner.svelte`: fixed, it described the pre-tier ordering.

**Claude P2 — `back()` never loaded the signals**: fixed, one line.

**Claude P2 — serialised fetches**: both step pickers now `Promise.all` the integrations list and the local counts.

**Claude P2 — no bound on pick reporting.** Not fixed here, deliberately. Any workspace member with unqualified `resources:write` can loop the route, and the outbound call carries the instance's `X-uid`, so the hub cannot distinguish it from a genuine pick. A per-instance throttle on this side would only slow a single instance down; the counter is global and public, so the dedup that actually protects it belongs on the hub, keyed on that uid. Tracking it there rather than adding a limiter here that would give false assurance.

The secondary note on the same finding — concurrent picks each pulling `/resource_types/list` on a cold cache — is real but bounded: it needs simultaneous first-saves within the same process after a 1h TTL expiry, and each is one request. Happy to add single-flight if you'd rather not carry it.

PR #10984, opened 2026-09-05
test: fit the relock no-op tests inside the 60s worker cap

rubenfiszel:
## Summary

`relock_noop::relative_import_relock_deploys_only_when_the_lock_changed` and
`relock_noop::relock_waiting_on_a_deploy_requeues_for_its_successor` have been failing
intermittently on every open backend PR, with `worker timed out: Elapsed(())` at
`windmill-test-utils/src/lib.rs:384`, usually preceded by `has been running for over 60 seconds`.

`in_test_worker` wraps the spawned worker in `tokio::time::timeout(60s, ..)` and `select!`s it
against the test body with `biased`. A healthy worker future never resolves, so that timeout is in
practice a hard 60-second wall-clock cap on the whole test, and when it trips the harness blames
the worker rather than the test. **That cap is not touched here** — it is long-standing (#6573)
and guards every backend test. The two tests were simply too slow, so this makes them fit.

Reproduced on a run of `main` itself, not caused by any one branch.

## Where the time went

In one CI run of the `relock_noop` binary, all three of its tests started together. The file's
Python test finished in 8.3s; the two Bun tests blew the cap on the same runner in the same
minute. Two costs, measured locally with checkpoint timestamps inside the test bodies:

- **The blind drain.** `wait_for_jobs` ended with a `timeout(2s, completed.next())` loop, so every
  call paid a flat 2 seconds of pure sleeping regardless of what happened. Test 1 calls it five
  times and test 2 three times: **10s and 6s of the (then) 11.5s and 7.4s runtimes were that
  sleep.** It is also a weak check — under load 2s may not be long enough for a spurious job to
  show up at all.
- **Bun.** Every deploy and relock spawns `bun run build.js` plus a prebundle. Locally that is
  cheap; on CI it is what separates the Bun tests from their Python sibling. Note test 2 has no npm
  dependency at all and was still the one that timed out, so this is per-job Bun cost, not the
  `lodash` fetch.

## Changes

All in `backend/tests/relock_noop.rs`.

- **Deploy Python scripts instead of Bun ones.** What these tests pin is language-independent —
  relock cascade, lock comparison, requeue-on-supersede — and Python is the cheapest runtime that
  still produces a real lockfile out of relative imports. The "dependency changed" step swaps
  `lodash@4.17.21` for the PyPI package `tiny`, which `backend/tests/python_jobs.rs` already
  resolves earlier in the same `cargo test` run. The file is now `#![cfg(feature = "python")]` —
  `backend-test.yml`, the job that runs these tests, builds with `python`, and the third test in
  the file was already gated that way. The trade is real and worth naming: a build without the
  `python` feature no longer runs these two.
- **Replace the 2s blind drain with a fixpoint.** `wait_for_jobs` now waits until `v2_job_queue`
  is empty. A dependency job queues its fan-out (`process_relative_imports`) *before* it completes,
  so an empty queue is a real fixpoint rather than a lull — quicker *and* firmer than sitting out a
  fixed delay, since a spurious job is now observed as a queued row instead of hoped-for within 2s.
- **Shrink the lock-wait poll** from 300 × 100ms (a 30-second budget on its own) to 20ms polling
  under a shared `WAIT_BUDGET` ceiling. Measured, that step takes **45–67ms**: a's content-only
  redeploy has the same requirement set as its predecessor, so `uv_pip_compile` hits
  `pip_resolution_cache` and no `uv` process runs at all. I also confirmed the poll observes the
  real thing rather than passing vacuously — at the moment it fires, `pg_stat_activity` shows
  b's relock `active`/`Lock`/`transactionid` on the `fetch_script_for_update` SELECT, while the
  test's own lock-holding transaction sits `idle in transaction`/`Client` and is correctly not
  matched.
- **Bound the counted wait.** Each `completed.next()` now has a `WAIT_BUDGET` timeout reporting
  `only {i} of {count} jobs completed`, so a missing job is named instead of hanging until the
  worker cap trips (see the mutation results below).
- **One `WAIT_BUDGET` (20s) for all three waits** instead of three separate magic numbers, with the
  constraint that matters written next to it: it has to stay clear of the 60s cap, or a stuck step
  reports as a worker timeout instead of saying what it was waiting on.

## Measurements

`before` and `after` binaries built from the same feature set (`quickjs,python`) and run
**interleaved within each rep** — before, then after, back to back — so load drift cancels rather
than favouring whichever variant happened to run in a calm minute. Load average is sampled at the
start of each variant's block and recorded per rep. This box has 24 cores and four sibling agent
sessions were running their own builds and backend test suites on it; **that ambient load is the
busy-runner condition**, so no synthetic load generator was used (adding one would only have
starved the sibling sessions). Seconds:

| round | rep | loadavg (before / after) | test 1 before → after | test 2 before → after | all 3 before → after |
|---|---|---|---|---|---|
| A (quiet) | 1 | 4.3 / 5.9 | 11.52 → **1.77** | 7.85 → **1.38** | 12.58 → **1.66** |
| A (quiet) | 2 | 5.6 / 5.1 | 11.57 → **1.49** | 7.43 → **1.33** | 11.66 → **1.64** |
| A (quiet) | 3 | 5.0 / 3.2 | 11.42 → **1.48** | 7.28 → **1.32** | 11.64 → **1.63** |
| A (quiet) | 4 | 3.0 / 2.4 | 11.55 → **1.56** | 7.68 → **1.31** | 11.65 → **1.57** |
| A (quiet) | 5 | 2.7 / 2.3 | 11.53 → **3.56** | 7.26 → **3.10** | 12.23 → **3.92** |
| A (quiet) | 6 | 8.9 / 20.8 | 14.19 → **3.42** | 10.09 → **3.09** | 14.00 → **3.61** |
| B (busy) | 1 | 31.8 / 33.9 | 18.22 → **4.25** | 9.39 → **3.68** | 16.44 → **4.18** |
| B (busy) | 2 | 33.4 / 34.5 | 14.06 → **1.98** | 10.15 → **1.76** | 14.16 → **2.95** |
| B (busy) | 3 | 31.4 / 20.3 | 14.75 → **1.62** | 7.50 → **1.57** | 11.68 → **2.06** |
| B (busy) | 4 | 19.2 / 15.6 | 13.69 → **1.61** | 7.64 → **1.51** | 12.23 → **1.77** |
| B (busy) | 5 | 14.8 / 10.0 | 12.56 → **1.49** | 7.30 → **1.36** | 11.71 → **1.69** |
| C (busy) | 1 | 33.2 / 33.9 | 21.72 → **4.56** | 9.34 → **4.15** | 14.61 → **5.43** |
| C (busy) | 2 | 33.7 / 36.1 | 14.45 → **3.43** | 9.36 → **3.39** | 18.05 → **3.86** |
| C (busy) | 3 | 34.9 / 46.6 | 16.21 → **2.32** | 21.05 → **1.48** | 14.11 → **1.79** |
| C (busy) | 4 | 43.4 / 26.1 | 14.98 → **2.78** | 7.83 → **3.73** | 19.75 → **2.15** |
| C (busy) | 5 | 24.6 / 23.4 | 11.73 → **16.59** | 8.03 → **50.05** | 16.48 → **2.09** |
| C (busy) | 6 | 35.2 / 22.5 | 12.04 → **1.49** | 10.65 → **1.36** | 12.07 → **1.68** |

### The margin is the point

Speed is only the mechanism; **headroom under the 60s cap is the actual claim that the flake is
gone rather than rarer.**

- On the five genuinely quiet reps (loadavg ≤ 6 on both blocks), **before** ran 7.3–12.6s — a
  ~5x margin — and **after** runs 1.3–3.9s, a **~15–45x margin**.
- Under ambient load, **before** ran 11.7–21.7s (~3x margin, and on CI it went past 60s outright);
  **after** stays at or under 5.4s in 16 of 17 reps.
- Repetition in the CI shape (whole binary, `--test-threads=3`): **45 consecutive runs at loadavg
  14–36, 45/45 pass, max 9.8s, median ~2.5s** — still a ~6x margin at the worst run of 45.

A ~3–5x margin is what CI load erases; ~15x is not.

**The one honest outlier**: round C rep 5, `after`, test 2 took 50.05s (it still passed) at the
peak of sibling contention on this box's shared Postgres — note its `all3` in the very same rep was
2.09s. Log analysis of the slowest run I could capture attributes the tail to Postgres, not to the
test: `SLOW_QUERY … total duration: 1.81s` on the job-queue read, while four other test suites were
concurrently creating and dropping `sqlx::test` databases in the same cluster. CI runs one Postgres
for one job, so that specific contention does not exist there — but it is one outlier in ~60 runs,
not zero, and it belongs in the record.

## Does it still pin the same behaviour?

Checked by breaking the production code from #10966 and confirming the tests fail:

| mutation in `worker_lockfiles.rs` | result |
|---|---|
| `commit_relock` never returns `Unchanged` (always deploys) | all 3 tests FAIL in 1.5s — test 1 on `the leaf's own job and one no-op relock of its importer, and nothing for c` |
| `commit_relock` re-reads the head once instead of `0..4` | test 2 FAILS on `only 2 of 3 jobs completed` |

The second one is why the bounded counted wait was added: before it, that regression reported as a
60s `worker timed out`; now it reports in 23s with the reason.

## Deliberate coverage trade

These tests no longer drive the relock path with Bun/TS. That is intentional, not incidental:
`commit_relock`, `requeue_relock` and the fan-out are language-agnostic, and Bun relative-import
lock generation itself stays covered by `backend/tests/bun_jobs.rs` and
`backend/tests/dedicated_workers.rs`.

This unblocks the backend test job for the currently open PRs.

## Test plan

- [x] `cargo test --features quickjs,python --test relock_noop` — 3/3 pass
- [x] Pre-PR review: `local-review-codex` (`gpt-5.6-sol`, xhigh) clean; `local-review` returned two
      P2 nits (a comment that argued against the deleted implementation, and a lock-wait deadline
      cut below what the happy path warranted) — both addressed above
- [x] 45 consecutive runs of the whole binary at `--test-threads=3` under load, 0 failures
- [x] Both properties re-verified by mutating the production code and watching the tests fail
- [x] Compiles and is a no-op (0 tests) without the `python` feature
- [x] `rustfmt --check` clean on the changed file
- [ ] CI backend-test green

### DIFF
--- a/backend/tests/relock_noop.rs
+++ b/backend/tests/relock_noop.rs
@@ -10,6 +10,11 @@ use windmill_test_utils::*;
 
 const W: &str = "test-workspace";
 
+/// Ceiling on any one wait below. Everything here settles in well under a second, so this only
+/// bounds a step that is stuck, and it has to stay clear of the 60s cap `in_test_worker` puts on
+/// the whole body: past that the harness reports a worker timeout instead of what was waited on.
+const WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(20);
+
 const A: &str = "def main():\n    return 'a'\n";
 const A_COMMENTED: &str = "# same dependencies, different content\ndef main():\n    return 'a'\n";
 const A_WITH_TINY: &str = "import tiny\n\ndef main():\n    return 'a'\n";
@@ -101,15 +106,14 @@ async fn wait_for_jobs(
     count: usize,
 ) {
     for i in 0..count {
-        tokio::time::timeout(std::time::Duration::from_secs(20), completed.next())
+        tokio::time::timeout(WAIT_BUDGET, completed.next())
             .await
             .unwrap_or_else(|_| panic!("only {i} of {count} jobs completed"));
     }
     // Then let anything else that was queued run out, so a job the assertions say must not
     // exist would have shown up here. A dependency job queues its fan-out before it completes,
-    // so an empty queue is a fixpoint rather than a lull, and waiting for one is both quicker
-    // and firmer than sitting out a fixed delay.
-    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(20);
+    // so an empty queue is a fixpoint rather than a lull.
+    let deadline = std::time::Instant::now() + WAIT_BUDGET;
     loop {
         while let Ok(Some(_)) =
             tokio::time::timeout(std::time::Duration::from_millis(20), completed.next()).await
@@ -277,7 +281,7 @@ async fn relock_waiting_on_a_deploy_requeues_for_its_successor(
                 .unwrap();
 
             // b's relock skips generation and reaches its commit, where it waits on the lock.
-            let deadline = std::time::Instant::now() + std::time::Duration::from_secs(15);
+            let deadline = std::time::Instant::now() + WAIT_BUDGET;
             let mut waiting = false;
             while !waiting && std::time::Instant::now() < deadline {
                 waiting = sqlx::query_scalar(
--- a/backend/tests/relock_noop.rs
+++ b/backend/tests/relock_noop.rs
@@ -16,7 +16,8 @@ const W: &str = "test-workspace";
 /// rather than surfacing as a worker timeout.
 const WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(30);
 
-/// How often the waits below re-check, and the floor on one turn of the drain loop.
+/// How often the waits below re-check. The drain reads the queue once per completion it waits
+/// this long for, so it doubles as the floor on one turn of that loop.
 const POLL: std::time::Duration = std::time::Duration::from_millis(20);
 
 const A: &str = "def main():\n    return 'a'\n";
@@ -131,9 +132,6 @@ async fn wait_for_jobs(
             tokio::time::Instant::now() < deadline,
             "the queue never emptied"
         );
-        // The stream ends instantly once it is closed, so this is what keeps the loop from
-        // spinning on the queue read until the deadline.
-        tokio::time::sleep(POLL).await;
     }
 }
 
--- a/backend/tests/relock_noop.rs
+++ b/backend/tests/relock_noop.rs
@@ -10,10 +10,14 @@ use windmill_test_utils::*;
 
 const W: &str = "test-workspace";
 
-/// Ceiling on any one wait below. Everything here settles in well under a second, so this only
-/// bounds a step that is stuck, and it has to stay clear of the 60s cap `in_test_worker` puts on
-/// the whole body: past that the harness reports a worker timeout instead of what was waited on.
-const WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(20);
+/// Budget for one wait below, counted completions and drain together. Even against a cold cache
+/// these settle in a few seconds, so it only bounds a step that is stuck, and it stays under the
+/// 60s cap `in_test_worker` puts on the whole body so the panic names what was being waited on
+/// rather than surfacing as a worker timeout.
+const WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(30);
+
+/// How often the waits below re-check, and the floor on one turn of the drain loop.
+const POLL: std::time::Duration = std::time::Duration::from_millis(20);
 
 const A: &str = "def main():\n    return 'a'\n";
 const A_COMMENTED: &str = "# same dependencies, different content\ndef main():\n    return 'a'\n";
@@ -105,20 +109,17 @@ async fn wait_for_jobs(
     completed: &mut (impl futures::Stream<Item = uuid::Uuid> + Unpin),
     count: usize,
 ) {
+    let deadline = tokio::time::Instant::now() + WAIT_BUDGET;
     for i in 0..count {
-        tokio::time::timeout(WAIT_BUDGET, completed.next())
+        tokio::time::timeout_at(deadline, completed.next())
             .await
             .unwrap_or_else(|_| panic!("only {i} of {count} jobs completed"));
     }
     // Then let anything else that was queued run out, so a job the assertions say must not
     // exist would have shown up here. A dependency job queues its fan-out before it completes,
     // so an empty queue is a fixpoint rather than a lull.
-    let deadline = std::time::Instant::now() + WAIT_BUDGET;
     loop {
-        while let Ok(Some(_)) =
-            tokio::time::timeout(std::time::Duration::from_millis(20), completed.next()).await
-        {
-        }
+        while let Ok(Some(_)) = tokio::time::timeout(POLL, completed.next()).await {}
         let queued: i64 = sqlx::query_scalar("SELECT count(*) FROM v2_job_queue")
             .fetch_one(db)
             .await
@@ -127,9 +128,12 @@ async fn wait_for_jobs(
             return;
         }
         assert!(
-            std::time::Instant::now() < deadline,
+            tokio::time::Instant::now() < deadline,
             "the queue never emptied"
         );
+        // The stream ends instantly once it is closed, so this is what keeps the loop from
+        // spinning on the queue read until the deadline.
+        tokio::time::sleep(POLL).await;
     }
 }
 
@@ -293,7 +297,7 @@ async fn relock_waiting_on_a_deploy_requeues_for_its_successor(
                 .await
                 .unwrap();
                 if !waiting {
-                    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
+                    tokio::time::sleep(POLL).await;
                 }
             }
             assert!(waiting, "b's relock never reached the row lock");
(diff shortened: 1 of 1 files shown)


rubenfiszel:
Round 1 (head `69f0e12`): codex **Good to merge**, pi **Good to merge**, claude **Mergeable, nits**. All three of Claude's findings are addressed in `2875691`:

**1. `WAIT_BUDGET` was per-`next()`, so the doc comment overclaimed.** Fixed as suggested: one `deadline` computed at `wait_for_jobs` entry, shared by the counted loop (`timeout_at`) and the drain. A `wait_for_jobs(.., 3)` now spends at most one budget, not `3 × budget + drain`, so the comment is true of the constant as written.

**2. "20s is tight for a cold `uv` bootstrap."** Fair, and it was the one number I had no measurement behind — so I measured it rather than guessing. Windmill's whole cache root is `WINDMILL_DIR`, so pointing it at a fresh directory gives a genuinely cold runner:

```
rm -rf /tmp/coldcache && WINDMILL_DIR=/tmp/coldcache \
  cargo test --features quickjs,python --test relock_noop -- --test-threads=3
```

That downloaded the interpreter, the uv cache and `tiny` from scratch (126MB of `cache/py_runtime`, `cache/uv`, `cache/python_3_12` created), with all three tests in parallel:

| | cold, 3 tests in parallel |
|---|---|
| longest single `wait_for_jobs` | **4.17s** |
| whole binary | **9.12s** |

So the cold bootstrap costs ~4s, not the 25–35s the finding assumed. I still raised the budget — **20s → 30s**, ~7x the measured cold worst case. I did not go to 35–40s because the budget's only job is to turn a hang into a named panic, and it has to fire *before* the 60s cap to do that; past ~30s a stuck first wait starts racing the cap it exists to pre-empt. 30s buys the headroom the finding asked for while keeping that property.

**3. Busy-poll if `completed` ends (`Ok(None)`).** Real — the inner `while let` returns instantly on a closed stream and the outer loop would hammer `SELECT count(*)` until the deadline. Added a `POLL` sleep at the end of the loop body as the floor on one turn, with a comment saying why it is there.

Re-verified after the change: 6/6 warm runs green, plus a fresh cold-cache run green in 3.80s.

rubenfiszel:
✅ Review round clean @ 0184470b97d47dde0fa22ff287c9c214f24059ba — all three reviewers returned "Good to merge" at 2875691783 (codex, claude, pi); the only item in that round was Claude's correction to its own round-1 nit, addressed in 0184470b97d47dde0fa22ff287c9c214f24059ba.

That round-2 note was right: `listen_for_uuid_on` is a `stream::unfold` that hits `.expect("lost database connection")` rather than returning `None`, so the closed-stream case my `POLL` sleep guarded against is unreachable — the inner `timeout(POLL, ..)` was already the per-turn floor. Rather than keep a defensive line justified by a state that cannot occur, 0184470 deletes the sleep and folds the floor into the `POLL` doc comment. Comment and dead-guard only, no behaviour change: `2 insertions(+), 4 deletions(-)` in the one test file.

Re-verified on the new head: 5/5 runs of the whole binary at `--test-threads=3` green (2.9–12.4s at loadavg 4–8), `rustfmt --check` clean.

rubenfiszel:
⚠️ **`cargo_test` is red on this PR for a reason unrelated to this change — `main` is red the same way.**

The job never reaches a test; it fails to compile:

```
error[E0061]: this function takes 1 argument but 2 arguments were supplied
  --> src/ee.rs:1382:20
1382 |  assert_eq!(seats_consumed(c.author_count, c.operator_count), 3.5);
note: function defined here
  --> windmill-common/src/ee.rs:160:8
160 | pub fn seats_consumed(seats: &ActiveSeatCounts) -> f64 {
```

Evidence it is not from this branch:

- This branch's only diff vs `main` is `backend/tests/relock_noop.rs`; it touches no EE file and does not modify `ee-repo-ref.txt`.
- `main`'s own `backend-test` run at `fce635d3c` ([run 33917980155](https://github.com/windmill-labs/windmill/actions/runs/33917980155)) fails with the identical error, and the four `backend-test` runs on `main` before it are also red.

Root cause, for whoever picks it up: `main`'s `ee-repo-ref.txt` pins EE `1a10132e`. At that commit `seats_consumed` takes `&ActiveSeatCounts` (it gained `guest_seat_count`), and every production caller was updated — `windmill-common/src/ee.rs:297,407,574,644`, `stats_ee.rs:147,148` — but the `#[cfg(test)]` caller at `src/ee.rs:1382` still passes the old two `i64`s. `cargo check` does not compile `#[cfg(test)]` blocks, which is why `check_ee` / `check_oss` are green and only `cargo test` fails. The one-line fix is `assert_eq!(seats_consumed(&c), 3.5);` in the EE repo (the assertion value stays 3.5 as long as that fixture adds no guests).

**Consequence for this PR:** the flake fix here is necessary but not sufficient to get `backend-test` green — nothing compiles, so no test runs, mine included. The change is verified locally instead: 60+ runs including 45 consecutive at `--test-threads=3` under load, a cold-cache run, and two production-code mutations that make the tests fail as intended. I have left the "CI backend-test green" box unchecked rather than claim a signal CI cannot currently produce.

PR #10985, opened 2026-09-05
fix: meter WAC compute per segment, not the whole sleep

rubenfiszel:
## Summary

A Workflow-as-Code (WAC v2) parent job that parks — on `sleep()`, on `waitForApproval()`, or
while its dispatched children run — kept `v2_job_queue.running = true` **and its original
`started_at`**. When such a job was later completed by a path that supplies no worker-measured
duration, `commit_completed_job` falls back to `now() - started_at`, so `duration_ms` covered the
entire park. Two such paths exist and both hit parked WAC parents:

- a cancel — `cancel_single_job` → `add_completed_job_error(..., None)`, and the worker's
  cancelled-at-pull path (`handle_queued_job` returns before it starts timing);
- the WAC child-failure handler in `result_processor.rs`.

On cloud that duration is the billing input. A workspace whose scheduled sweep spawns WAC
children that sleep for days showed **31,286,371 computations** for a month in which ~600 jobs
actually ran; 31,252,823 of those came from 200 jobs bulk-cancelled mid-sleep.

Fixing that alone exposed the mirror-image problem. A WAC parent resumes with a **fresh timer**
on every pull, and only the final invocation reaches `add_completed_job` — so the meter saw the
last segment and nothing else. A workflow that burned 30s, slept, burned 30s, slept, then
finished in 5s was billed for 5 seconds.

So this PR does both: the park no longer counts, and each **segment** is metered as it ends
rather than one of them being metered at completion. The sum is the compute the workflow
actually used, and it accrues while the workflow runs instead of only when it finishes — a
workflow that is cancelled or never completes is still billed for what it ran.

## Changes

- Clear `started_at` when a WAC parent parks, via a new `suspend_wac_parent` used by all three
  suspend arms. It pairs the update with `workspace_id` and errors rather than silently matching
  no row — a no-op on the dispatch arm is unrecoverable, since the children push next and
  decrement a `suspend` that was never set. It returns the segment that just ended.
- Extract `meter_execution_seconds` from `apply_completed_job_cloud_usage`, and charge the
  finished segment at each of the four points a WAC parent stops running (child dispatch,
  approval, sleep, inline checkpoint). The final segment is still charged by the ordinary
  completion path, so the two together are the whole workflow.
- Metering happens where the parent really parks, never on the dispatch push-failure rollback —
  that path goes on to complete the job, which would charge the same segment twice.
- Restore `started_at` in that rollback from the in-memory pull value, so a job that genuinely
  ran doesn't land in `v2_job_completed` with a null start.
- Add a `suspend` branch to the Runs list time column: a parked parent has no `started_at`, so
  it would otherwise read "Waiting executor" next to the Suspended hourglass.

`v2_job_completed.duration_ms` still reports the last segment — unchanged, and unchanged for
every non-WAC job. Only the meter aggregates.

Billing is not the only consumer of the old wrong duration: `workspace_fairness_ee.rs` clamps
occupancy by `completed_at - duration_ms`, so a bulk-cancelled parked parent also poisoned the
overload signal. That is fixed by the same change.

No consumer of `v2_job_queue.started_at` regresses: all three zombie sweeps filter
`suspend_until IS NULL`, `force_complete_zombie_job` already does `COALESCE(q.started_at, now())`,
`add_outstanding_wait_time` is gated on `!was_suspended_job`, the runs filters fall back to
`v2_job.created_at`, and `v2_job_completed.started_at` is nullable (already null for ~94k
completed jobs a month in production).

## Test plan

Exercised against a local backend running with `CLOUD_HOSTED=true` and `--features cloud`, so the
usage meter itself ran. Every row is a measured delta on a quiet workspace's `usage` counter,
using a workflow that burns ~2.5s, parks, then burns ~2.5s again.

| scenario | lang | billed | decomposition |
|---|---|---|---|
| sleep -> resume -> success | bun | 5 | push + 2s + 2s |
| sleep -> resume -> success | python | 5 | push + 2s + 2s |
| approval -> approve | bun | 5 | push + 2s + 2s |
| approval -> disapprove | bun | 5 | push + 2s + 2s |
| cancel while parked (900s sleep) | bun | 3 | push + 2s; the park bills nothing |
| child dispatch -> success | bun | 6 | parent push + parent segment 2s + child push + child 2s + final segment 0 |
| dispatch push failure | bun | 3 | push + 2s, charged **once** |

Two of those carry the argument:

- **Child dispatch (6).** The parent's first segment is 2.5s of real compute that never appears
  as a row in `v2_job_completed`, because intermediate segments produce no completion. It is
  billed. On `main` this workflow bills 4 and drops it.
- **Push failure (3).** The only path where a segment could be charged twice — the suspend
  captures it, then the rollback lets the job complete for real. Charged once, because metering
  happens only where the parent genuinely parks. Its `duration_ms` of 2534 also confirms the
  rollback restores the pulled segment start rather than the rollback moment.

Mid-flight metering was observed directly, not inferred: with the job still parked
(`suspend = 1`, `started_at` NULL, zero rows in `v2_job_completed`), the counter had already
moved by the push plus the first segment.

Python goes through the same code: `python_executor.rs` delegates to
`bun_executor::handle_wac_v2_output`, and the worker log shows
`"WAC v2 parent job sleeping for 120s"` from `bun_executor.rs` with `language="python3"`.

Other checks:

- [x] `cargo test --test wac_suspend_started_at` — fails on main, passes here
- [x] `cargo check` against a live database, with and without `--features cloud`, and
      `cargo check -p windmill-worker --features cloud` / `-p windmill-queue --features cloud`
      standalone (both were broken on `main` before the feature-forwarding commit)
- [x] `npm run check` (0 errors)
- [x] Runs list shows "Suspended (created ...)" for a parked parent (screenshot below)

Not covered: an approval whose timeout expires with no decision, and a worker dying mid-segment.

## Screenshots

runs-suspended

Top row is the parked WAC parent — violet Suspended hourglass with "Suspended (created 96s ago)".
Below it, the resume probe that slept and dispatched children, completed in 0.01s.

https://claude.ai/code/session_01AzS5d5Dc3GL49kWZPQAQCg


rubenfiszel:
✅ Review round clean @ c3cb06b9343088a2b1f9e90e3301928dccf70238

rubenfiszel:
✅ Review round clean @ abcc25e0bc6bffb91fd64170a06db01e00015169 — nit-only verdicts (codex and pi "Good to merge", claude nit-only); the remaining nit is addressed in c018783aa8, which adds `windmill-common/cloud` to the worker's cloud feature.

PR #10990, opened 2026-09-05
fix: stand the WAC park down for a cancel that beat it to the row

rubenfiszel:
## Summary

Cancelling a Workflow-as-Code (WAC v2) parent is a two-step dance between the API and the
worker, and it had no interlock.

`cancel_single_job` soft-cancels a running parent by writing `canceled_by`,
`scheduled_for = now()` and `suspend = 0`, then relies on the worker to notice `canceled_by`
on the next pull. Meanwhile the worker, finishing a segment, parks the parent through
`suspend_wac_parent`, which wrote `suspend = $3, suspend_until = now() + …` unconditionally.

If the cancel commits first and the park second, the park **overwrites** `suspend = 0`. The
cancel is not lost — `canceled_by` stays set — but nothing acts on it until the row becomes
pullable again, which on the sleep arm means waiting out `suspend_until`. A `sleep(86400)`
makes the user's cancel appear to do nothing for a day.

This is pre-existing, not introduced by #10985 — `main` before that PR had the same
unconditional update. #10985 reduced its cost (the park is no longer billed) but did not
close it.

## Changes

- `suspend_wac_parent` now takes the row's lock with `SELECT … FOR UPDATE` before deciding,
  and returns a `WacPark` enum instead of just the segment length:
  - `Parked(Option<i64>)` — the normal case, carrying the finished segment for metering;
  - `Cancelled(CanceledBy)` — a cancel is already on the row, so **nothing is written** and
    the cancel's `suspend = 0` survives;
  - the row being gone at all stays a hard error, as before.
- The three suspend arms in `bun_executor.rs` (dispatch, approval, sleep) return early on
  `Cancelled` without committing, so the checkpoint, the timeline entries and the approval
  metadata are all rolled back and — on the dispatch arm — no child is ever pushed against a
  parent that never parked. Nothing had been committed yet, so dropping the transaction is
  the whole rollback; the heavier push-failure block at `bun_executor.rs:3150` exists only
  because that path has committed work to undo.
- `wac_cancelled_mid_segment` hands the cancel to the completion path: it sets the worker's
  `canceled_by` (which is what makes the job land as `canceled` rather than `failure`, since
  the row was cancelled after this worker pulled the job) and returns the canonical
  `{"name": "Canceled", …}` payload.
- Deliberately **no** `end_wac_segment` on that path. `started_at` is untouched, so the
  ordinary completion charges the segment via `now() - started_at`; metering here too would
  bill it twice. The existing dispatch push-failure rollback is the precedent.
- The `inline_checkpoint` arm is unchanged: it writes `running = false` rather than a
  suspend, so a concurrently-cancelled row stays pullable by the normal pull query either
  way.

## Test plan

`backend/tests/wac_suspend_started_at.rs` gains `wac_suspend_stands_down_for_a_cancel`,
which plants a cancelled queue row, calls the park helper, and asserts the row does not come
back parked. It fails without the fix (`a cancelled parent must not park, got
Parked(Some(30002))`).

Exercised on a running instance with a temporary 25s delay wedged between the segment ending
and the park (since removed) so the race is deterministic:

| scenario | result |
|---|---|
| bun `sleep(600)`, cancel inside the window, **guard disabled** | still queued 242s later: `suspend=1`, `suspend_until=+10min` — the bug |
| same, guard on | `canceled` 23s after the cancel, `duration_ms=26566` (the real segment, not 600s) |
| python `sleep(600)`, cancel inside the window | `canceled` 21s after the cancel, `duration_ms=25068` |
| bun dispatch, cancel inside the window | parent `canceled`, **0 children pushed**, nothing queued, no checkpoint left behind |
| cancel while already parked | leaves the queue within 5s — unchanged pre-existing behaviour |
| uncontended bun `sleep(5)` / dispatch / `waitForApproval` + resume | all `success` |
| uncontended python `sleep(5)` | `success` |

- [ ] `cargo test --features quickjs --test wac_suspend_started_at` (2 passed)
- [ ] `SQLX_OFFLINE=true cargo check -p windmill-worker --features quickjs` and `--features quickjs,cloud`
- [ ] sqlx offline cache updated: 2 entries added, 1 orphan removed

https://claude.ai/code/session_01GHfNFFJh3ozZYgyyoaEepu

---
## Summary by cubic
Fixes a race where a cancel that lands while a WAC v2 parent finishes a segment could be overwritten by the park, leaving the job stuck until the suspend window expires. The park now locks the row, checks for an existing cancel, and stands down so the job completes as canceled instead.

**Bug Fixes**
- `suspend_wac_parent` locks the row with `SELECT … FOR UPDATE` and returns a `WacPark` enum instead of just the segment length.
- If a cancel is already on the row, the park is skipped so `suspend = 0` survives.
- The dispatch, approval, and sleep arms return early on `Cancelled` without committing, so no child is pushed against a parent that never parked.
- `wac_cancelled_mid_segment` sets the worker's `canceled_by` and returns the canonical `Canceled` payload, and no `end_wac_segment` is called to avoid double billing.
- The `Canceled` payload now comes from a shared `canceled_result` helper in `jobs.rs`, also used by `canceled_job_to_result`.
- Adds a test that plants a cancelled queue row and asserts the park helper stands down.

<sup>Written for commit 411edc2c87b9e80622d6b5e6c8e184dfc5677508. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/10990?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Thanks — nit taken. `wac_cancelled_mid_segment` now calls a shared `windmill_queue::canceled_result(reason, canceler)`, and `canceled_job_to_result` is a two-line wrapper over it, so the payload shape lives in one place (411edc2c87).

On the other note, about the description saying the completion charges via `now() - started_at`: I think the description is right and the `$9` reading is not, for this path. `process_completed_job`'s failure branch passes `duration` literally as `None` to `add_completed_job_error` (`result_processor.rs:1754`), so `COALESCE($9, now() - started_at)` falls through to the `started_at` arm — `duration` is only threaded onward for the flow-step branch below it. The measured runs agree: the already-parked cancel came back with `duration_ms = 4`, which is `now() - started_at` against a `started_at` the pull had just re-stamped, not a worker-measured segment.

That is also why the test asserts `started_at.is_some()` on the `Cancelled` path — leaving it set is what makes the charged duration the segment that actually ran. Happy to be shown otherwise if I've misread the branch.

PR #10997, opened 2026-09-06
fix(frontend): no phantom draft when opening a CLI-pushed script

diegoimbert:
## Summary

Opening the script editor on a deployed script and leaving it untouched could persist a user draft. The editor re-infers the schema from the code as soon as it mounts and writes the result into the script being edited. The autosave then compared that script against the raw stored row, so any script whose stored schema differed from what the frontend parser emits (a CLI push carrying a schema written by another parser version or by hand, e.g. one lacking the `originalType` key on string args) diverged on the very first open and was saved as a draft. Scripts deployed from the UI never hit it because their stored schema is already the editor's own output.

The deployed baseline the autosave `discardIf` compares against is now normalized the way the mounted editor normalizes it, so an unedited open compares equal and the autosave resolves to a no-op.

## Changes

- New `frontend/src/lib/scriptEditorSchema.ts`: `schemaAsEditorMounts` runs the same `inferArgs` + partition-arg injection the editor applies at mount on a copy of the stored schema; `injectPartitionArg` moves here from `ScriptEditor.svelte` so both sides share one implementation.
- `scripts/edit/[...path]/+page.svelte`: the `deployedBaseline` used by the autosave `discardIf` carries that normalized schema instead of the raw stored one (load token re-checked after the await).
- `ScriptEditor.svelte`: imports `injectPartitionArg` from the shared module instead of defining it locally.
- `DbtEditor.svelte`: infers the descriptor schema on mount, as `ScriptEditor` does, so the mounted dbt editor and the normalized baseline agree; previously it only re-inferred on the first descriptor edit.

With auto-save off the same `discardIf` path parks a delete, so these scripts also stop raising a phantom "unsaved changes" prompt on navigate-away.

## Test plan

Exercised on a running instance by driving the frontend with headless Chromium:

- [x] Push a script from the CLI whose `script.yaml` schema lacks `originalType`, open it in the editor, navigate back: before the fix a draft row was created and the home list showed it as a draft; after the fix the autosave posts a delete (`value: null`) and no draft row exists.
- [x] Same for a script with freshly generated CLI metadata and for a UI-deployed script: no draft either way.
- [x] Typing an edit in the editor still creates a draft; deleting the edit again removes it.
- [x] dbt script with a stale stored schema (empty, while the parser emits the `command` arg): untouched open leaves no draft; a descriptor edit then revert, and a model-file edit then revert, both end with the draft deleted.
- [x] `npm run check`: 0 errors. Changed files pass prettier.

Not exercised: a duckdb `// partitioned` script and a preprocessor-kind script through the new baseline path (they go through the same moved code the editor already used).

## Screenshots

CLI-pushed script (schema without `originalType`) open in the editor, untouched:

editor-cli-pushed-script

Home after navigating back: no draft indicator on `f/clitest4/old`:

home-after-back-no-draft

Local Codex review skipped: the `codex` CLI is not installed on this machine.

https://claude.ai/code/session_01QpqVdiTiqVGCvgmpBzL6m3


diegoimbert:
Addressed the Codex P1 in 02877aa884: rather than exempting dbt from the baseline normalization (which would have left the descriptor edit-and-revert case with a no-op draft, since `DbtEditor` re-infers on that edit), `DbtEditor` now infers the descriptor schema on mount the way `ScriptEditor` does, so the mounted editor and the baseline agree for every language. Exercised on a dbt script with a stale stored schema: untouched open leaves no draft; a descriptor edit followed by a revert, and a model-file edit followed by a revert, both end with the draft deleted.

diegoimbert:
Addressed the round-2 Codex P1 in 4e559e1895: `schemaAsEditorMounts` now retries the inference once when the first attempt throws (same `content && language` condition as `ScriptEditor`'s mount retry), and `DbtEditor`'s mount inference got the same single retry so all three sides land in the same place after a transient wasm-init failure. Re-exercised the untouched-open case on the bun and dbt fixtures after the change: no draft row either way. The transient first-parse failure itself was not simulated.

diegoimbert:
✅ Review round clean @ 4e559e1895226d48a54b6c2aa449341973345abf

PR #11000, opened 2026-09-07
fix(cli): keep permissioned_as on single-item push, as sync push does

hugocasa:
> **Left in draft on purpose:** this changes which identity deployed items run as, across five commands and every repo on `syncBehavior: v1`. Reviews are clean (Codex + Claude both "Good to merge" @ `135aa13`), but a run-as identity change wants a human look before it goes ready rather than an agent's judgement call.

## Summary

`wmill schedule push <file> <remote_path>` reset a schedule's `permissioned_as` to whoever ran the push, every time, with no way to opt out (GIT-996). A schedule that must keep running as a specific service identity needed a `wmill schedule set-permissioned-as` follow-up after every single push.

The preserve-on-update logic in `pushSchedule` was already correct, but it is gated on a `permissionedAsContext` that only `wmill sync push` ever built — the standalone command passed `undefined`, so the branch never fired. The same omission was present in the standalone `app push`, `flow push`, `script push` and `trigger push`, so all five are fixed together.

## Changes

- `buildPermissionedAsContext` in `cli/src/core/permissioned_as.ts`: the whoami lookup that `sync push` did inline, now shared. Returns `undefined` below `syncBehavior: v1`, which is where the old inline gate sat, so both push paths resolve ownership identically. `sync push` now calls it instead of building the context itself.
- All five standalone `push` commands build and pass the context.
- `readEffectiveSyncBehavior` in `cli/src/core/conf.ts`: resolves `syncBehavior` through the current branch's `workspaces.<name>.overrides` the way `sync push` does, rather than reading only the top level of `wmill.yaml`.
- Each push pins its file argument with `path.resolve` before reading the config: `readConfigFile` moves the cwd to the `wmill.yaml` root when the file sits in a parent directory, which would otherwise break a relative path passed from a subdirectory.
- Regression test driving the real `schedule push` command over a temp `wmill.yaml`, asserting both sides of the gate.

## Decisions

- **Gated on `syncBehavior: v1`, matching `sync push`.** That is the flag that already versions this behaviour (`wmill init` writes it for new repos), and it is what makes the pull/push pair coherent: a v1 pull strips `permissioned_as` from the file, so the value can only come back from the remote at push time. Below v1, both push paths keep reassigning ownership, unchanged. A v0 repo can still pin an owner by putting `permissioned_as` and `preserve_permissioned_as: true` in the YAML itself.
- **Raw apps are not covered.** `pushRawApp` always sends a freshly generated policy and takes no context, on this path or `sync push`'s, so `wmill app push` on a `.raw_app` folder still reassigns `on_behalf_of`. Left as-is so the two paths stay in parity; the code comment says so.
- **Non-admin pushers are out of scope.** Preservation stays gated on admin / `wm_deployers` in every pusher, and the backend enforces the same rule regardless of what the CLI sends, so a non-admin `schedule push` still reassigns ownership exactly as it did before this PR. `sync push` warns first via `preCheckPermissionedAs`; wiring an equivalent pre-check into the single-item commands would mean synthesizing `Change` entries and is a separate change.

## Test plan

- [x] `cd cli && bun test test/*_unit*.test.ts` — 1106 pass, including the new guard; the new test fails on the pre-fix code with `permissioned_as: undefined`
- [x] `bunx tsc --noEmit` — no new errors (38 pre-existing, unchanged)
- [x] Reproduced the bug on the unpatched code: schedule owned by `u/svc`, push as `admin`, owner reset to `u/admin`
- [x] Exercised against a running instance with `syncBehavior: v1` — `schedule`, `flow`, `app`, `script` and `trigger` push each keep the remote's `u/svc` / `svc@windmill.dev` owner while applying the local edit
- [x] Below v1 and with no `wmill.yaml` at all: unchanged behaviour, no crash
- [x] `syncBehavior: v1` set only under `workspaces.<branch>.overrides`: now honoured
- [x] Relative file path from a subdirectory below the `wmill.yaml` root, with explicit `--base-url/--workspace/--token`: pushes and preserves
- [x] `sync push` unaffected: still preserves, and still refuses a non-admin push with the `preCheckPermissionedAs` message

---

Follow-up from review: `syncBehavior` is now resolved from the workspace being pushed to — the entry `--workspace` names, else the one matching the resolved profile's remote + workspace id, else the git branch — the same order `sync push` uses. `resolveWsNameForConfigFromFlags` and `inferWsNameFromProfile` moved from `sync.ts` to `conf.ts` so both paths resolve it through one implementation rather than two that must stay in step. `readEffectiveSyncBehavior` reads the config with `warnIfMissing: false`, so a push in a checkout with no `wmill.yaml` doesn't gain a new warning.

### DIFF
--- a/cli/src/commands/app/app.ts
+++ b/cli/src/commands/app/app.ts
@@ -409,7 +409,8 @@ async function push(
   if (isRawAppByName || hasRawAppYaml) {
     const { pushRawApp } = await import("./raw_apps.ts");
     const merged = await mergeConfigWithConfigFile(opts);
-    // Raw apps have no policy, so no ownership to preserve.
+    // Raw-app ownership preservation is not implemented on either push
+    // path: sync push hands pushRawApp no context either.
     await pushRawApp(
       workspace.workspaceId,
       remotePath,
--- a/cli/src/commands/app/app.ts
+++ b/cli/src/commands/app/app.ts
@@ -426,7 +426,7 @@ async function push(
       undefined,
       await buildPermissionedAsContext(
         workspace.workspaceId,
-        await readEffectiveSyncBehavior(),
+        await readEffectiveSyncBehavior(opts, workspace),
       ),
     );
     log.info(colors.bold.underline.green("App pushed"));
--- a/cli/src/commands/app/app.ts
+++ b/cli/src/commands/app/app.ts
@@ -12,7 +12,11 @@ import * as wmill from "../../../gen/services.gen.ts";
 import { ListableApp, Policy } from "../../../gen/types.gen.ts";
 
 import { GlobalOptions, isSuperset } from "../../types.ts";
-import { getWmillYamlPath, mergeConfigWithConfigFile } from "../../core/conf.ts";
+import {
+  getWmillYamlPath,
+  mergeConfigWithConfigFile,
+  readEffectiveSyncBehavior,
+} from "../../core/conf.ts";
 import { readInlinePathSync } from "../../utils/utils.ts";
 import devCommand from "./dev.ts";
 import lintCommand from "./lint.ts";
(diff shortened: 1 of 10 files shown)


hugocasa:
✅ Review round clean @ 135aa13d1415149c5239a38f7f0dfbcb614fde03

PR #11001, opened 2026-09-07
feat: run and test scripts from the AI chat through an argument form

AlexRV12:
## Summary

The AI chat can run a deployed script (`run_script`) or test a draft (`test_run_script`) by putting the script's own argument form in a chat card. The model prefills what it can infer, the user edits and presses Run, and the card carries the call from there through logs to the result. The form is the confirmation: the script itself runs only when the user submits, with one deliberate exception below — a `dynselect-` field fetches its options when the form appears.

No secret reaches the job as a value. A sensitive argument becomes an ephemeral secret variable and the job is given `$var:`/`$jsonvar:` instead, on both postures: the mounted field mints as the user types, and `runThroughForm` mints in its stead for anything that reaches the run still literal. Proposed file bytes are emptied for the user to attach, since a form prefilled with them puts unbounded base64 into the stored transcript, and the card marks them by size.

The card shows the reference rather than a marker, as the run page does for the same job: whoever reads it is whoever's prompt produced the value, so hiding it would only make the card misdescribe what ran. A cancelled card settles on what the form actually held, for the same reason. Only the model-facing result still redacts, and only a literal — a secret the user typed into the form has not been in its context before.

In an AI session the same form can be opened in the preview pane, which then follows the call to `/run/<jobId>` once the job exists.

## Changes

- `job_args.ts` gains the run form's argument hygiene, all pure and unit-tested: `coerceArgsToSchema` (convert to the declared type, or empty the field and report it), `dropUndeclaredArgs` (an argument the schema never names has no widget on any run surface in the product, so it is not sent and the model is told to re-read the schema), `enforceDisabledDefaults` (extracted from `RunForm`, which now calls it), and recursive file stripping and secret/file redaction.
- `processSecretArgs` becomes the one place that decides how a sensitive argument reaches a job. It now covers string secrets as well as objects, walks nested and `oneOf`-branch fields through the same `mapLeaves` the stripping uses, and normalises a `$var:` named for a field that cannot hold a string. `PasswordArgInput` mints through it instead of carrying its own copy — which fixes the top-level-only gap for its four other callers (`RunForm`, `ScriptEditor`, `FlowPreviewContent`, `DbtEditor`) too. Three rules it has to hold for those callers as much as for the chat: an empty field mints nothing (`ArgInput` synthesises `''` for every untouched string, so the alternative is a variable per run for a field nobody filled in); two leaves are told apart by their whole path, since a property name can itself contain a dot; and the arguments are snapshotted on the way in, since every caller binds a form that stays editable across the mints and a leaf is addressed by its path — an array row moved in between would otherwise take another row's reference.
- A form that never ran settles its card on its own arguments rather than the proposal it was published on, so the Inputs tab names the reference the field had already minted — matching what the user was looking at, and what a run writes onto the card when it starts. Both of the form's exits take that snapshot: its Cancel button, and the turn being stopped from the action row or the composer. Anything that never minted, an object secret or a keystroke ahead of the mint, reads `<hidden>` there instead of the value; an empty field stays empty, for the reason `processSecretArgs` mints nothing for one. The literal stays only in `runForm.args`, which is what the form reopens on after a reload.
- `run_script` and `test_run_script` in the global toolset, both routed through one `runThroughForm`. `runScriptByPath` joins `COVERED_ENDPOINTS` so the API catalog points at the tool instead.
- Plan mode is re-checked at both points on that path where a write would follow: before the form is opened at all, and again before the job starts. `processToolCall` gates these tools once, ahead of the schema fetch, and a form reaching the screen after that mints a proposed secret on mount — which the later gate is too late to unmake. A `dynselect-` field's helper job is withheld the same way, from the form itself, since that one does wait for a mount.
- The system prompt defaults to `test_run_script` for any run request and gates `run_script` on the user naming the deployed version; `run_script`'s description now says the same thing, where before it contradicted the prompt.
- `RunScriptCard` renders the whole call as one tool call row: the form while pending, then logs and result in a tabbed body, with the elapsed time in the header.
- `AIChatManager` keeps a run's card and its job together across mid-turn saves and reloads, so a detached run is not left spinning and a restored card is not left actionable.
- A `runform:` preview slot in the sessions panel mounts the chat's own form on the same tool call, so Run there is Run in the chat and one draft backs both.
- Incremental job output via `getJobUpdates`, the only endpoint carrying `new_result_stream`, so the card streams logs and a partial result.
- Two supporting bits usable elsewhere: `Tabs` gains an opt-in sliding selection indicator, and `DynamicInput` now honours `disabled` (a pre-existing gap, so a disabled dynamic field was editable on every form in the product).
- A test run's `dynselect-`/`dynmultiselect-` fields offer options from the draft it previews, via `{ source: 'inline', code, lang }` — the same helper shape the script editor's own test panel uses. `deployed` would answer for the wrong version, and is stale or absent for a draft never deployed. The code rides on `RunFormDisplay` only while the form is mounted, dropped with the schema once it settles so no card carries a copy in history.
- Two eval cases covering a deployed run and a secret supplied by variable reference.

## Pre-existing defects this surfaces but does not fix

All reproduce on `main` without the chat, and each wants its own PR:

- `MultiSelect` throws on any non-array value; `/scripts/get/<hash>#tenants="$var:u/admin/x"` takes the page down. This form coerces ahead of it, which is a guard at one caller, not a fix.
- A malformed `format: "ai-provider"` object throws the same way, in any form over a schema carrying that format.
- A `dynselect-`/`dynmultiselect-` field runs a helper job when it mounts, before Run and undone by no Cancel. `RunForm.svelte:317` has always done this on the script page, and the fix belongs in `DynamicInput`. See below for what that means for a test run, which is new here.
- **That helper job stores the form's other arguments, a plaintext secret included.** `SchemaForm.svelte:359` hands the helper `{ ...args, [thisField]: undefined }` and `DynamicInput.svelte:107` forwards it, so on a form whose password field still holds a literal the helper is queued with that literal, and it lands in `v2_job.args` permanently and readable by anyone who can view that run. Measured on a fixture with a password field plus a `dynselect-` field, opened at `#api_token="…"` on the plain script page, no chat involved: two helper jobs carried the literal and a third, queued one second later once `PasswordArgInput` had minted, carried `$var:u/admin/secret_arg/…` and worked identically. Not gated on the helper's signature — a helper declaring only `(filterText: string)` still produced job args containing `api_token`. The chat reaches it the same way, since a form here opens on the model's proposal; minting ahead of the form was tried and reverted for the reason given below. The fix belongs in `DynamicInput`, which changes every form in the product and wants its own PR.
- `ArgInput` renders a wrong-typed scalar with no error and Run enabled everywhere else in the product. Repro with no chat involved: open `/scripts/get/f/qa/number_only#count="abc"` on a script whose only argument is a required `number`. The field renders empty, Run stays enabled, and the job runs with `count: "abc"`. Every producer feeding `RunForm` has this — a shared run link, a saved input, "Run again" on a job whose schema has since changed. `coerceArgsToSchema` guards the chat's args against it; the fix belongs in the form, where all of them would get it.
- `ArgInput.validateInput`'s email / ipv4 / ipv6 / uuid branches (`ArgInput.svelte:447-454`) set `error` but never assign `valid`, so an invalid email shows its error with Run still enabled. `ArgInput.svelte:142` already flags the design ("valid and error should be one coherent state").

## The one thing that runs before Run

A `dynselect-`/`dynmultiselect-` argument fetches its options by executing the named helper when the field mounts, so opening a run form queues that job before the user has pressed anything, and Cancel does not undo it. For a deployed run this is the deployed script's own helper, exactly as on the script page. For a **test** run it is the draft the model just wrote, which is new in this PR: displaying the form executes model-authored code.

Accepted deliberately. The job is an ordinary preview run permissioned as the user — no privilege escalation, and it can reach nothing the user could not reach themselves — while the alternative is a free-text box for a value that has a real, discoverable option set, which for a chat user who never opens the script is worse than the risk it removes. Withheld entirely under plan mode, and whenever the draft's code did not come through.

Worth knowing alongside it: under the auto-accept posture no form mounts, so no helper runs and the model's proposed value for a `dynselect-` field is submitted unvalidated — the same as `main`, which runs whatever the model proposed with no form at all.

**The helper-job secret leak listed above is reachable from here, and this PR does not fix it.** A helper is queued with every argument the form holds, so a form that arrives holding a literal credential puts it into a job the user never confirmed. Minting ahead of the form was tried and reverted: it turns an object secret into a `$jsonvar:` string, and the nested `SchemaForm` that an object argument renders as normalises that straight back to `{}`, so Run submits empty credentials. Stripping the argument instead would permanently hand `undefined` to a picker that legitimately needs the token, on every form in the product.

The fix belongs in `DynamicInput`, holding the mount-time helper run while a field still holds raw text, and wants its own PR: it is `main`'s behaviour on the script page too, where a user typing a secret loses the same race.

## What the posture decides, and what the form decides

An earlier revision of this branch tried to work out, without mounting the form, whether the form would have been satisfied — `applySchemaDefaults` filling declared defaults, `formNeedsUser`/`requiredUnanswered` looking for required keys with no value. That is a second implementation of `ArgInput` and `SchemaForm`, and it has to be *complete* over the whole schema language to be safe. It never was: four review rounds landed on it in turn (nested defaults, `null`, a required child of a supplied optional object, a required key read through `Object.prototype`), each fix a new special case, with `oneOf` branches, array items and every constraint in `validateInput` still unmodelled.

It is gone. The posture answers whether to ask, and the form answers everything else:

- **Ask permission** — the form mounts, applies its own defaults, runs its own validation, and mints secrets on submit.
- **Auto-accept** — no form mounts, so the three rules that hold against *every* caller are applied without one: an argument the schema does not declare is not sent, a disabled one is the schema's to set, and a sensitive one becomes a reference. Nothing beyond those is derived — no defaults filled, no validation, no guess at whether Run would have lit up. `main` applies none of the three (`testRunScriptByPath` is `normalizeTestRunArgs(args.args)`, i.e. `args ?? {}`, straight into `executeTestRun`), so a model-proposed credential reaches the job in plaintext there.

The one thing the missing form must still do is turn a proposed secret into a reference, or a credential the model supplied would be stored in `v2_job.args` in plaintext for everyone who can read that run — which is what `main` does today. So `processSecretArgs` becomes the single place that decides how a secret reaches a job, and both paths call it: `PasswordArgInput` mints through it, and a form-less run calls it in the widget's stead. It also gained the cases that were split across the two implementations — string secrets (previously only `PasswordArgInput` knew about those), nested and `oneOf`-branch secrets (the top-level-only gap listed above), and a `$var:` named for a field that cannot hold a string, which the worker resolves to text and which now reads the same variable as `$jsonvar:`.

An earlier revision also emptied a proposed secret out of the form, to keep it out of the stored transcript. That was dropped: `stripSecretArgs` only ever sees the *model's* proposal, and any model proposal is already in that same IndexedDB record as `tool_calls[].function.arguments` and `_anthropicContent` — searching the store for a proposed secret finds it in the user's prompt, the tool call, its native-format twin and the model's prose, and there is no store one reaches that the other does not (`HistoryManager` makes no API calls and there is no server-side session table). Emptying the field removed a duplicate from a document that already held the original, and charged the user a retype for it. What the user types into the form was never the concern: that plaintext lives in component state, is minted on the way out, and is never persisted.

Net effect of the cleanup: 314 lines added, 482 removed.

## Eval A/B

`global` mode on `sonnet`, eleven shared cases, branch head against the merge-base (`670404ffe2`), one run each: 10/11 before, 8/11 after. The fixed cost of the new tool surface reads off the six cases with an identical tool trajectory (+477, +482, +488, +488, +500, +517), so **about +490 tokens of window occupancy per iteration**, ~1.9% of a 26k window. The other five swing on trajectory, not tax. Both new cases pass.

Measured at an earlier head. Nothing since has added or removed a tool or a parameter, so the per-iteration cost stands; two model-facing lines did change (an undeclared argument is now reported, and the disabled-field line names what the run used), which would move a pass rate rather than the occupancy.

Re-run at the current head on thirteen cases, `sonnet`, one run each: 11/13, with both run-form cases passing and average final context 25,474 tokens. The two failures are `global-test9-human-weekday-trial-job` (max turns) and `global-planmode2-sketches-while-planning`, the same pair the paragraph below records as failing on `main` too. Not a paired A/B — the earlier run's case list is not recoverable from `ai_evals/results/`, so the denominators differ and only the absence of a new failure class carries over.

The pass-rate gap is variance rather than the branch. Re-running the two cases that failed only on the after side, three times per side: `global-test31-draft-test-run-not-deployed` passes 3/3 on both, and `global-test9-human-weekday-trial-job` passes 1/3 on both, failing on `main` the same way it does here (max turns, and the draft-count check). `global-planmode2-sketches-while-planning` fails on both sides.

## Test plan

- [x] Run a deployed script from a session: schema read first, form prefilled, no job until Run.
- [x] A required base64 file: field opens empty, the job gets the real bytes, the card and the model get a size marker.
- [x] A proposed literal secret is minted by the field on mount: the form reads `Linked to variable u/admin/secret_arg/…` with no user action, and the job never sees the literal.
- [x] Cancel: the card's Inputs tab and `displayMessages[].parameters` carry that same reference, not the literal. Checked in IndexedDB — the literal is left only in `runForm.args`, which the form reopens on, and in the prompt and tool call that carried it in.
- [x] Escape from the form's action row, which stops the turn instead of cancelling the form: same reference on the tab, same five copies in IndexedDB.
- [x] Run through the chat form after the snapshot went in: the job's stored args read `token: "$var:u/admin/secret_arg/…"` and the script reads the real value back (`token_len` matches). `$state.snapshot` is what the snapshot uses — `structuredClone` throws `DataCloneError` on the reactive proxy the form binds.
- [ ] Not re-driven since the revert: the password field after moving the form to the preview pane, and after a reload onto a pending form.
- [x] Script page, string secret: typed into the field, the job stores `$var:u/admin/secret_arg/…` and the script reads the real value back (`token_len` matches).
- [x] Script page, object secret: the job stores `$jsonvar:…` and the script receives `creds_type: "object"`, not a string.
- [x] A wrong-typed argument is converted; one with no reading is emptied and reported.
- [x] A `dynmultiselect-` field given a `$var:` reference opens empty instead of taking the card down.
- [x] Cancel starts no job; entering plan mode while the form is open starts no job.
- [x] Test a draft: same card on the draft's schema, reading Test / Testing / Tested.
- [x] Move a live form between card and preview pane mid-edit; edits survive both ways, one form mounted at a time.
- [x] Run from the pane: the tab becomes the run page in place, the card fills beside it.
- [x] Reload mid-run and mid-form; no duplicate job, no card left spinning.

Covered by unit tests but not driven in the browser: plan mode arriving during the schema fetch, where no form must open and no ephemeral variable be created; two secret fields whose paths a joined name would conflate, which must each keep their own reference; an untouched optional password field, which must mint nothing and must not be marked as hidden either; and a settled card keeping a minted reference while redacting a secret that never minted, on both of the form's exits.

Known: a submitted form whose job never starts leaves its preview tab open on "This run form is no longer available", for the user to close. Deliberate, since the failure is already on the card and a throw from `startJob` cannot say whether the server queued the job.

`npx vitest run src/lib/components/copilot/chat src/lib/components/sessions src/lib/components/job_args.test.ts src/lib/components/secretArgUtils.test.ts` gives 1,423 passing. The 6 remaining failures are in `ContextManager.test.ts` and `app/core.test.ts`, neither of which this branch touches, and both fail the same way on `main`. `AIChatManager.test.ts` was among them at 88 failures: its `$lib/aiStore` mock omits `copilotInfo`, which `sendRequest` reads first, so every test driving a real turn died there — including one this PR adds. Restoring the export to the mock makes the whole file green. `npm run check` reports 22 errors across twelve files, none of them touched by this branch — they arrived with `main`.

## Screenshots

Card states, dark except where noted. Fixtures are `f/qa/wide_form` (fourteen arguments), `f/qa/log_stream`, `f/qa/stream_answer`, `f/qa/fails` and `f/qa/number_only`.

**Waiting on you** — the deployed schema's own widgets, prefilled where the model could infer. Only the fields scroll, so Run never leaves the fold.

the run form in a card

**A value the field has no reading of** — `count` sent as `"a lot"` on a `number` field opens empty and says so, since an empty field otherwise reads as nothing sent. A `"7"` would have been converted silently.

an argument emptied for having no reading

**Moved to the preview panel** — one draft backs both, so what was typed in the card is already here, and closing the tab brings it back with the panel's edits.

the same form open in the preview panel

**The run in the panel**, which the tab follows into once Run is pressed, keeping its place in the strip.

a run open in the preview panel

**Succeeded**, in the run page's own renderer.

a succeeded run showing its result

**What it ran with.** This shot predates the change above: the password argument now reads as its `$var:` reference, the way the run page shows it for the same job.

the inputs tab of a settled card

**As raw JSON**, the shape every other tool card is read in.

the card as raw JSON

**Running** — elapsed time leads the row, logs open on their last line and follow the tail, Cancel sits where Run did.

a run in flight with its logs

**Cancelled mid-flight** — an outcome, not a failure, carrying what the run page shows for a cancelled job.

a run cancelled mid-flight

**Streaming**, and the same run settled in light.

a result streaming into the card

the streamed run settled, light

**Failed**, with the error where the result would be.

a failed run showing its error

**A test run of a draft**, in its own tense.

a settled test run of a draft

**Read-only under plan mode**, and **cancelled before starting**, which never ran and says so.

a run form held read-only by plan mode

a run cancelled before starting

https://claude.ai/code/session_01ERed9zo2oJjpSczzMayaNh


AlexRV12:
Thanks — three of the four findings are addressed, and the first P1 is being kept deliberately. Detail below.

**[P1] Dynamic inputs execute code before the user confirms the run — kept, deliberately.**

The finding is accurate, including the part that stings: this is new for test runs. The previous revision withheld the helper for `kind === 'test'` for exactly this reason, and the commit that added `{ source: 'inline', code, lang }` reversed that decision.

Keeping it, for two reasons:

- The helper job is an ordinary preview run permissioned as the user (`run_dynamic_select` pushes it with `authed`, and `$var:`/`$res:` resolve through that job's client). There is no privilege boundary crossed: the code reaches nothing the user could not reach themselves, and cannot escalate. What is lost is the timing of consent, not the scope of what can happen.
- A `dynselect-` field whose options never load is a free-text box for a value that has a real, discoverable option set. The user of this form has typically never opened the script, so removing the picker makes the form unusable for its main case — a worse outcome than the risk it removes.

What we have changed is the claim, not the behaviour: the summary no longer says "nothing runs until the user submits", there is a dedicated section spelling out what executes on form display and why it is accepted, and the code comments at `RunArgsFormDisplay.svelte` and both tool registrations in `core.ts` now say the same rather than asserting the form is the only thing that runs.

**[P1] Auto-accept treats schema defaults as supplied without applying them — fixed** in `ffee9f7d3f`.

New `applySchemaDefaults` in `job_args.ts`, applied in `runThroughForm` *before* stripping, so a secret or file carrying a default still opens empty instead of being handed its default back. Both paths now build the same arguments, so the bypass and the mounted form cannot diverge. Regression test added as you asked (`sends a schema default the model omitted when yolo skips the form`); reverting only the `applySchemaDefaults` call makes it fail with `retries: 3` absent from the request body.

**Also fixed:** `schemaForTestRun` treated a stored `properties: {}` as a declared schema and skipped inference, so a form could come up empty for code that takes arguments (`5cbbbd5495`).

**One gap we found while reviewing this, left for its own PR:** under auto-accept no form mounts, so no helper runs and the model's proposed value for a `dynselect-` field is submitted unvalidated against options that were never fetched. Forcing the form open for those fields, the way a stripped secret does, is a behaviour change that wants its own think rather than a patch here.

AlexRV12:
Second round: one fixed, one answered, one pushed back on with what I measured on a running instance.

**[P1] YOLO still omits nested schema defaults — fixed.**

`applySchemaDefaults` now descends on exactly the condition `requiredUnanswered` uses to credit a nested default as answered (a declaration naming its own `properties` and `required`, not a `oneOf`, against a non-array object value), so the two cannot disagree about which fields the form would have filled. Regression test added in `job_args.test.ts`: removing the recursive branch fails it with `region: 'eu'` absent, and a second case pins that a supplied value and a `oneOf` declaration are both left alone.

**[P1] Sensitive schema defaults refill fields after being stripped — working as intended, I think.**

A schema `default` is stored in the script's metadata in plaintext, visible to anyone who can read the script, and the system prompt tells the model to read that schema before proposing arguments (`core.ts:1348`). So a "default password" is already in the model's context and on the script's page before any of this runs; removing it from the form protects nothing.

What `stripSecretArgs` is for is narrower: not letting a value **the model proposed** stand as the user's consent. A schema default is not the model's proposal — for a deployed run it is the script author's, and `/scripts/get/<hash>` prefills it through this same `ArgInput`. Blanking it only in the chat form would make one form in the product hide a value every other surface shows.

The case that did matter is the auto-accept path, where no form is mounted and the user never sees the field. That is already closed: `applySchemaDefaults` runs before stripping, so a sensitive key carrying a default lands in `strippedKeys`, `formNeedsUser` returns true, and the form opens rather than the run starting on a value nobody chose.

**[P2] An interrupted pre-job card still renders as an active run — not reproducible as described.**

The finding assumes the stored card has `submitted: true`. It does not today: `#patchRunForm` mutates `displayMessages` without persisting, and the checkpoint snapshot is written by `#persistBackgroundJobs`, which only runs once a job is registered — after the id arrives.

Exercised on a running instance, with job creation delayed 20s to widen the window:

- Press Run, stop the turn, **reload**: the card reads "Cancelled — This run was cancelled before the script started" with a "Not run" tab. Not "Running".
- Press Run, stop the turn, **stay on the page**: the card settles to "Ran" with its result. Stopping the turn aborts the model stream but not the in-flight `startJob`, so the id arrives, the job registers, and the poller finishes it.

So the state described is latent rather than live — it becomes reachable the moment any save lands inside that window, and the transcript layer already composes a "while starting, check the runs page" message for it that no component renders.

What is live there is the opposite defect: a run that **did** start is labelled as cancelled before starting. The job existed (`v2_job` row, `f/qa/number_only`, created within the window). Fixing that properly means persisting `submitted` at Run **and** rendering the indeterminate state, because either alone is worse than today — persisting alone produces exactly the stuck "Running" this finding describes. That pair is a change to the submit/persist path rather than a rendering tweak, so it is going in its own PR rather than at the end of a review round.

AlexRV12:
Both points fixed, and I want to state the contract explicitly so this seam stops moving.

**Fixed**

- **null.** `ArgInput.computeDefaultValue` guards with `value == undefined || value == null`, so it fills both. `applySchemaDefaults` now does too. Test: `fills a null the same as an absent value` — reverting the branch fails it with `retries: null` where `3` was expected.
- **Optional nested defaults.** The recursion required the nested declaration to carry a `required` array, mirroring `requiredUnanswered`. That mirror was too tight: `requiredUnanswered` needs `required` because it answers "is a required field unanswered", which is not a constraint on defaults. It now descends into any non-`oneOf` declaration naming `properties`, against a non-array object value. Test: `fills a missing default at the depth the form would` now covers an all-optional nested object — re-adding the `required` check fails it with `opts: {}`.

The invariant between the two is now a superset rather than an equality, which is the direction that is safe: `applySchemaDefaults` descends everywhere `requiredUnanswered` credits a nested default as answered, and further. Narrower is the failure mode — that is what lets the posture skip a form over a default nothing applied — and the comment at the recursion says so.

**What this function deliberately does not do**

`computeDefaultValue` does more than apply `default`: with no default it synthesises `''` for a string, `false` for a boolean, `[]` for a list, the first member of a required enum, and a uuid where `x-auto-generate` is set. None of that is replicated here, and I do not intend to replicate it. The contract is: auto-accept sends the model's proposal plus every **declared** default, not a byte-identical copy of what a mounted form would submit.

That is defensible rather than a shortcut, because of how these schemas are built. `infer.ts:570` adds a parameter to `required` only when it has no default:

```ts
if (!arg.has_default && !schema.required.includes(arg.name)) {
    schema.required.push(arg.name)
}
```

So in an inferred schema a `default` means the field is optional *and* the code carries that same default — `def main(retries=3)`. Omitting the argument makes the worker apply 3 anyway, so `{}` and `{ retries: 3 }` produce the same run. The divergence is only observable on a hand-edited schema whose default differs from the code's, and the same fact makes `requiredUnanswered`'s required-with-a-default branch near-dead for inferred schemas.

If the goal were byte-identical arguments, the right fix would be extracting `computeDefaultValue` into a pure function both paths call, rather than growing a second implementation of it here. That touches `ArgInput`, which every form in the product renders, so it belongs in its own PR rather than at the end of a review round.

AlexRV12:
Fixed.

`requiredUnanswered` now runs two passes instead of walking `required` alone: required keys missing at the current level, then a descent into every object the caller actually supplied — required parent or not, since supplying the object is what puts its fields on the form. The descent condition is the one `applySchemaDefaults` already uses (a non-`oneOf` declaration naming `properties`, against a non-array object value), so what one fills the other inspects, and dropping the old `Array.isArray(declared.required)` requirement means a level with no `required` of its own no longer stops the walk before its grandchildren.

Regression test added to `opens a run form under yolo for an empty or nested-missing required field`: an optional `config` whose own schema requires `api_key`, proposed as `{ config: {} }`, must open the form. Restricting the descent to required keys fails it — the card comes back `submitted: true` where the test expects no bypass.

On scope, so this is on the record rather than discovered later: `formNeedsUser` models one failure — a required key with no value — while the mounted form's Run button is gated on `isValid` across all of `ArgInput.validateInput` (`:405-461`), which also refuses a number outside `minimum`/`maximum`, an invalid email, IPv4, IPv6 or UUID, a value failing a custom `pattern`, and a non-array where an array is declared. `oneOf` branches and the contents of array items are not walked by either function either. So auto-accept can still start a run the visible form would have refused, and closing that class properly means both paths calling one implementation of `ArgInput`'s defaulting and validation rather than a second copy of its rules living here. That extraction touches the component every form in the product renders, so it is not going in at the end of a review round — but it is the actual fix, and I would rather name it than keep narrowing the approximation one finding at a time.

AlexRV12:
Confirmed and fixed.

Verified by running rather than by reading: a schema with `required: ['constructor']` and a proposal of `{}` came back `submitted: true` — the bypass firing on `Object.prototype.constructor` read as an answer. `requiredUnanswered` now reads through `Object.hasOwn` at both the required-key check and the descent.

On why this was the only site: every other schema-keyed read in this path already builds on `Object.create(null)` — `coerceArgsToSchema`, `enforceDisabledDefaults`, `applySchemaDefaults` and `mapLeaves` each copy into a null-prototype bag before indexing, for the `__proto__` reason documented at `job_args.ts:20-22`. `requiredUnanswered` was the one function inspecting a bag it had not built, since `applySchemaDefaults` returns `{ ...result }` and the spread restores the prototype. Fixing it where the question is asked rather than by returning a prototype-less object keeps the guarantee local to the check, where one spread downstream cannot silently undo it.

Regression added to `opens a run form under yolo for an empty or nested-missing required field`: a required `constructor` the model omits must open the form. It fails before the change with `expected true to be undefined`.

Suites: `core.test.ts` + `job_args.test.ts`, 257 passing. `npm run check`: 0 errors, 82 pre-existing warnings.

One clarification against my previous reply, since the two are easy to conflate: this is not an instance of the validation-approximation class I named there. That one is about `formNeedsUser` modelling fewer failures than `ArgInput.validateInput` does, and its fix is the shared-implementation extraction. This was a plain property-read bug, and it is closed.

AlexRV12:
Fixed, and exercised in the browser on both hosts.

The check is now gated on the action row rather than on the panel, so the row stops the turn wherever it is mounted:

```ts
if (aiChatManager.hasPendingRunForm) {
	if (!active?.closest('[data-run-form-actions]')) return
} else if (!focusOnChat) return
```

**Exactly one case changes.** Over `hasPendingRunForm` / focus-in-action-row / `focusOnChat`, the handler's decision is identical to before in five of six cells; the sixth is pending-form + action-row + focus-outside-the-panel, which was the dead case and now cancels. Nothing else can reach it: `data-run-form-actions` is on one element in the tree (`RunArgsFormDisplay.svelte:224`), which only exists while a form is mounted, so an editor, drawer, modal or nav link can never satisfy the new branch. `RunFormPreviewSlot` and `PreviewTabHost` have no Escape handler that could lose the keypress.

**Verified on a running instance** (sessions chat, `f/qa/wide_form`, ask-permission posture), four presses:

1. Card action row → `cancelling request: {reason: user_cancelled}`, card settles to "Cancelled". Unchanged.
2. Pane action row, form moved with the card's Preview button (`panelEl.contains(row) === false`) → now the same cancel, the card settles, the tab closes cleanly and no job starts. Before the change this produced no console output at all.
3. Pane field with `ops@acme.test` typed in → value and form preserved, turn continues. The protective half still holds.
4. No pending form, focus on a sidebar link outside the panel → no cancel, the turn ran to completion. The branch I restructured still rejects Escape from outside the chat.

No unit test: the handler is a window listener over two mount points and a manager, and the fixture needed to assert one keypress would be larger than the logic it guards. `npm run check` 0 errors; suites unchanged against the branch baseline (the 94 failures are the three files that fail the same way on `main`, none of which imports this component).

AlexRV12:
Reworked rather than patched, because the last four findings here were all the same defect.

**The P1 is real.** Verified by running, not reading: a script declaring `creds` as a sensitive object, run with `creds: "$var:u/admin/qa_json_probe"`, returned `{"creds_type": "string", "creds_keys": null}`. `windmill-worker/src/common.rs:283` maps `$var:` to `json!(string)`; only `$jsonvar:` (`:292`) parses. And the shape is not hypothetical — `schema/PropertyEditor.svelte:302` renders an "Is sensitive" toggle for `type == 'object'` and `ScriptSchema.svelte:22` passes `showSensitiveToggle`, so a script's own arg editor produces it.

**But the one-line fix would have been the fifth special case.** `applySchemaDefaults` and `formNeedsUser`/`requiredUnanswered` were a second implementation of what `ArgInput`/`SchemaForm` do when a form mounts, and a predicate like that has to be *complete* over the schema language to be safe. Rounds 3, 4, 5 and 7 each found one more way the copy disagreed, and `oneOf` branches, array items and every constraint in `validateInput` were still unmodelled. So the copy is deleted.

The decisive fact is what `main` does: `testRunScriptByPath` is `normalizeTestRunArgs(args.args)` — literally `args ?? {}` — straight into `executeTestRun`. `main` strips nothing, defaults nothing and predicts nothing; the whole prediction layer was scope this PR invented. Now the posture answers whether to ask, and the form answers everything else. Auto-accept runs the model's arguments as sent, which is `main`'s behaviour.

**The one thing the missing form must still do is mint.** Otherwise a proposed credential is stored in `v2_job.args` in plaintext for anyone who can read that run — which is what `main` does today. So `processSecretArgs` becomes the single place that decides how a secret reaches a job, and both paths call it: `PasswordArgInput` mints through it while the user types, a form-less run calls it in the widget's stead. Folding the two implementations together also closed what each was missing on its own — string secrets (only the widget knew about those), nested and `oneOf`-branch secrets (the top-level-only gap this PR had listed as a known pre-existing defect, which affected `RunForm`, `ScriptEditor`, `FlowPreviewContent` and `DbtEditor` too), and this finding, since routing on declared type is now done in one place rather than by a hand-written check.

210 lines added, 394 removed.

**Verified on a running instance**, since the risk here is regressing paths that exist on `main`:

- Script page, string secret: typed a token, the widget minted `$var:u/admin/secret_arg/…`, the job stored the reference, and the script read the value back (`token_len: 17`, the length of the plaintext).
- Script page, object secret: the job stored `$jsonvar:…` and the script received `creds_type: "object"` — the probe that produced `"string"` above.
- Chat under ask-permission: the form opens with `creds` emptied and "A secret or a file, so it opened empty for you to fill in: creds".
- Suites unchanged against the branch baseline (94 failures in the three files that fail the same way on `main`); `npm run check` 0 errors.

Not exercised end to end: the chat's auto-accept branch, because `autoAcceptToolConfirmationsAvailable` is false on this instance. It is covered by two unit tests in `core.test.ts` — that a run the posture answers starts on the model's arguments, and that a proposed secret reaches `startJob` as a reference and never as a literal.

**A correction I owe on my own earlier comment.** I wrote that the mounted form's Run is gated on `validateInput` including "an invalid email, IPv4, IPv6 or UUID". That is wrong. Those branches (`ArgInput.svelte:449-457`) set `error` but never assign `valid`, so Run stays enabled — confirmed on `f/qa/wide_form`: type a valid email, edit it to `nope-not-an-email`, and Run is still clickable with "invalid email address" showing. Whether Run is disabled depends on `valid` being stale-`false` from an earlier "Required" pass, which makes it order-dependent. `ArgInput.svelte:142` already carries the note *"this should not exist, valid and error should be one coherent state"*. It is a pre-existing defect in every form in the product and out of scope here, but it is worth having on the record — not least because it means "make the toolset match the form" was never a well-defined target.

AlexRV12:
Follow-up to the previous comment: the secret *stripping* it described is now gone too, and this is the reasoning, since it reverses something I asserted there.

I claimed emptying a proposed secret out of the form was load-bearing for the stored transcript. It is not. I checked the actual stores rather than reasoning about them, using a session where the model had proposed `s3kret-from-chat`:

- Searching every IndexedDB database found the value in exactly one store, `copilot-chat-history` / `chats`, at six paths: the user's own prompt, the model's `tool_calls[].function.arguments`, its `_anthropicContent[].input` twin, the model's prose repeating it back, and the two `displayMessages` copies of the prompt and the prose.
- It was **not** in `displayMessages[].runForm.args`, because stripping had removed it — that is the one site stripping controls, out of six.
- `HistoryManager` makes no API calls and there is no server-side session table, so `actualMessages` and `displayMessages` live in the same record of the same browser-local store. There is no store one reaches that the other does not.

So `stripSecretArgs` only ever sees the model's proposal, and a model proposal is already in that record by construction. Emptying the field removed a duplicate from a document that already held the original, and charged the user a retype for it. The case that *would* matter — a user typing a password into the form and the plaintext being stored instead of a reference — was never stripping's to handle: that value lives in `PasswordArgInput`'s component state, is minted on the way out, and is never persisted.

**Verified that dropping it is safe**, since prefilling a password field is now load-bearing where before the field always opened empty. Loading `f/qa/wide_form` with `api_token="prefilled-plaintext"` in the args: the widget minted on mount with no user action (`generated $var:u/admin/secret_arg/UYIWuXO6spPb`), the bound args became `api_token="$var:u/admin/secret_arg/UYIWuXO6spPb"`, and the field showed "Linked to variable". The literal never reaches the job. Across the whole database, `select count(*) from v2_job where args::text like '%<the secret>%'` is 0, while 16 jobs carry `secret_arg` references.

What that leaves: file bytes are still emptied — that rationale is size, not secrecy, since a form prefilled with base64 puts unbounded data into the stored transcript — and secrets are still redacted to `<hidden>` in the persisted card and in everything the model reads back.

The real remaining leak is the one this made visible: the model's proposed secret sits in the stored tool call and is re-sent to the provider every turn. That is fixable at the storage boundary with the `redactSecretArgs` that already exists, and unlike stripping it would close the channel that actually carries the value. It is not in this PR; I would put it above the `ArgInput` extraction on the follow-up list.

Suites unchanged against the branch baseline (94 failures in the three files that fail the same way on `main`); `npm run check` 0 errors.

AlexRV12:
## Reverting the pre-form minting — rationale for the next round

Both reviewers were right about the object secret, and the fix is to remove its cause rather than work around it. The next push reverts the change that introduced it. Reviewers: this is deliberate, and here is the reasoning so the round does not re-litigate it.

### What was wrong

The previous head moved `processSecretArgs` ahead of publishing the form, so the form opened on `$var:`/`$jsonvar:` references. That closed the pre-confirmation helper-job leak for a `password: true` **string**, and broke a `password: true` **object**: the reference is a string, `ArgInput` renders an object-with-properties through a nested `SchemaForm`, and that form's `$effect.pre` normalises any non-object `args` to `{}`. The reference is erased on mount and Run submits empty credentials, silently.

### Why not fix it in place

Three shapes were considered and all three are worse than reverting.

**Mint strings early, objects at submit.** Splits one rule into two by declared type, and leaves the object case leaking exactly as it does today. The asymmetry would have to be explained at every future call site.

**Teach the widget to hold a reference for a non-string field.** This is a real bug and it is `main`'s, not this branch's: `SchemaForm` and `ArgInput` are unchanged here, so a "Run again" link carrying `$jsonvar:` into an object-with-properties field is erased on the plain script page today. Fixing it means changing how every form in the product renders object arguments. That is its own PR, not this one.

**Strip sensitive leaves from what `SchemaForm` hands `DynamicInput`.** Suggested during review and rejected on inspection. Today a helper that declares the secret as a parameter re-runs once the widget has minted, receives the reference, and the worker resolves it — so the helper works. Stripping it would hand that helper `undefined` permanently, so a `dynselect-` that lists options for an authenticated API would return nothing, on every form in Windmill, forever. That trades a bounded leak for an unbounded functional regression.

### What the revert leaves

Minting returns to submit time, which is what `main` does everywhere. Against `main` this PR is still strictly better on the surface that matters: `main`'s chat has no run form at all and sends a model-proposed credential straight into the run job's arguments as a literal, every time; here the run job gets a reference on both postures.

What remains is the pre-existing helper-job exposure already documented under "Pre-existing defects this surfaces but does not fix", and its real shape is narrower than the earlier text implied:

- a value picked from a variable is a reference in the form, so it is a reference in the helper job — never exposed, on any path
- a typed password starts minting on the first keystroke and the helper's re-run is debounced a second behind it, so typing is almost always safe
- the exposure needs a form that **arrives** holding plaintext (a shared run link, "Run again", a saved input, or the chat's proposal), because the first helper run happens at mount before anything can mint
- a credential **object** is the durable case, since those are not minted until submit

The correct fix is at the shared site: hold the mount-time helper run while a sensitive field still holds raw text, and let it fire once the value is a reference. That keeps the helper working, fixes the script page as well as the chat, and covers strings and objects alike. It changes how every form in the product loads its dynamic options, so it belongs in its own PR against the script page where a person can actually reach it.

### What is kept from the reverted commit

Only the pre-form minting and the `PasswordArgInput` "A value is set" branch go. These stay, since they were separate fixes that happened to travel in the same commit:

- `holdsSecret` no longer treats `''` as a secret, so an untouched optional password field mints nothing (this was a regression against `main` reaching `RunForm`, `ScriptEditor`, `FlowPreviewContent` and `DbtEditor`)
- the persisted card redacts file bytes, so the auto-accepted path cannot store base64 past the size guard
- the card shows a reference rather than `<hidden>`, matching what the run page shows for the same job

### Still open, not addressed by this push

`ai_evals_global` fails because the wasm parser cannot load on that runner, so schema inference throws, `schemaForTestRun` returns an empty schema, every argument is dropped and the model loops to max turns. The parser failure is environmental and pre-exists this branch. The code-side question is whether a failed inference should refuse the run outright rather than present a form reading "This script takes no arguments" with Run enabled. Comments welcome on that specifically.

AlexRV12:
**[P2] Interrupted pre-job cards remain "Running" — already answered in round 2, and the answer still holds.**

Full reasoning and the instance test are in https://github.com/windmill-labs/windmill/pull/11001#issuecomment-5573783556. Nothing on the relevant paths has changed since, so re-summarising briefly rather than re-litigating:

The explanatory `content` this finding refers to is written only by the `runState === 'starting'` branch of `settledToolDisplay`, which requires `runForm.submitted` to be true in the snapshot being persisted. It never is. `#patchRunForm` (`AIChatManager.svelte.ts:2025-2039`) mutates `displayMessages` in memory without saving, and of the three `#interruptedSnapshot` callers, `#persistBackgroundJobs` only runs once a job is registered — by which point `started` is true, so the state is `'started'`, not `'starting'`.

Exercised on a running instance with job creation delayed 20s to widen the window:

- Press Run, stop the turn, reload: the card reads "Cancelled — This run was cancelled before the script started", with a "Not run" tab. Not "Running".
- Press Run, stop the turn, stay on the page: the card settles to "Ran" with its result. Stopping the turn aborts the model stream but not the in-flight `startJob`, so the id arrives, the job registers, and the poller finishes it.

So the state is latent, not reachable. It becomes live the moment any save lands inside that window, which is why the message exists at all.

Note that this round's wording drops the `submitted: true` premise the round-2 version stated explicitly. That does not change the analysis: the explanatory content the finding asks to be rendered is produced by that premise alone.

Deferred deliberately, for the reason given in round 2: the honest fix is a pair — persist `submitted` at Run **and** render the indeterminate state — because persisting alone manufactures exactly the stuck "Running" card described here. That is a change to the submit/persist path rather than a rendering tweak, so it ships in its own PR.

AlexRV12:
Answering the round on `d0b5276`. One finding fixed, one dismissed with a correction, plus the body. Reasoning first so the next round does not re-derive it.

### [P2] The persisted card keeps the proposed literal — fixed, but not the way it was suggested

The diagnosis is right and the fix I pushed is a different one, because restoring `redactSecretArgs` at publish would have replaced one wrong value with another.

Verified in a browser against a running instance. A form opened on a proposed literal is minted by the field at mount, so while the form is up it reads `Linked to variable u/admin/secret_arg/lx4DvcsV0oa6`. Press Cancel and the Inputs tab read `token: "canary_pw_42"`. The two surfaces described the same argument differently, and `<hidden>` would not have reconciled them.

The asymmetry is that a run writes its arguments onto the card once it has them and a cancellation never reaches that write, so `parameters` keeps the copy made when the card was published. `#settleRunForm` now makes the cancel path do what the run path already does, using the draft the form was editing, through the same redaction: a reference stands, anything still literal becomes `<hidden>` — an object secret, which only mints at submit, or a keystroke that beat the mint. It also fixes a smaller thing: a cancelled card used to show the model's proposal rather than the user's own edits.

After the fix, the same scenario reads `token: $var:u/admin/secret_arg/lx4DvcsV0oa6` on the Inputs tab, clickable, as the run page shows it for a real job.

Deliberately **not** redacting `runForm.args` alongside it. With no live session the form's draft is seeded from that copy, so `<hidden>` there would come back as the field's value on a reload and be minted as a secret whose value is the string `<hidden>`. Dumping IndexedDB for the canary confirms the split: five copies remain, four of them the prompt and the tool call that carried the value in, and one `runForm.args` the form reopens on. The copy this finding is about is gone.

`AIChatManager.svelte.ts:1917` is corrected in the same commit — it now says file arguments, which is what is actually redacted there.

Unit test added, and confirmed to fail before the fix with `token: 'hunter2'`.

### [P2] "A value is set" — not restoring it, and the premise is wrong

The claim is that without that branch a set password field is indistinguishable from an empty one. It is not. `ArgInput` renders a `Linked to variable <path>` line next to the field whenever the value is a `$var:`/`$jsonvar:` reference, as a button that opens that variable for editing. `SchemaForm` constructs the `VariableEditor` it needs and passes it into every `ArgInput` it renders, so this holds in the chat card and in the preview pane alike — the surface the finding says is newly reachable.

Confirmed on the instance: the open form showed the masked box and, under it, `Linked to variable u/admin/secret_arg/lx4DvcsV0oa6` as a live link. So a set field is labelled, its variable is named, and it is one click from being edited.

That leaves the branch as duplicate signalling in a component shared by every auto-generated form in Windmill, which is not a change worth making from a nit on this PR.

### [P2] Interrupted pre-job cards remain "Running" (Codex)

Answered separately above: https://github.com/windmill-labs/windmill/pull/11001#issuecomment-5599303104. Still latent rather than reachable, for the same reason.

### The description

Rewritten. It described the reverted pre-form minting as current in five places, including a checked test-plan item. The secret paragraph, the changes bullet, the helper-job section and the test plan now describe the code as it stands, and the helper-job leak is stated as reachable and unfixed here rather than as closed.

AlexRV12:
Both nits fixed in `9444615`, both confirmed on a running instance. Each was a consequence of the previous commit rather than pre-existing, and each is now covered by a test proven to fail without its fix.

### [P2] Stop/Escape bypasses the cancellation settlement

Correct, and the previous commit only ever covered one of the form's two exits. `cancel()` resolves and clears `#runForms` directly — deliberately, since `settledToolDisplay` settles every card of a stopped turn at once and is the only thing that can tell a run which reached the server from one which did not — so the draft died with that loop and the card kept the proposal.

`#settledFormArgs` now holds the snapshot and both exits take it: `#settleRunForm` for the form's own Cancel, and the `cancel()` loop for the turn being stopped. There is no third exit — `#runForms` is resolved or cleared in exactly those two places.

Driven on the instance. Form opened on a proposed literal, minted at mount to `u/admin/secret_arg/4te1X1WVtkqK`, then Escape from the action row (console confirms `cancelling request: user_cancelled`, so this is the `cancel()` path and not the Cancel button). The settled Inputs tab reads `token: $var:u/admin/secret_arg/4te1X1WVtkqK`, and the note keeps its edit. Dumping IndexedDB for the canary leaves the same five copies as the Cancel path: the prompt and tool call that carried it in, and the one `runForm.args` the form reopens on.

### [P2] An empty sensitive field is redacted as though it held a value

Correct, and wider than described: this is not only a *cleared* optional password. `ArgInput` synthesises `''` for every untouched string, so any optional password the user never touched was marked `<hidden>` — on the model-facing summary already, and on the card since the previous commit.

The sharper form of it is that the two halves of one rule disagreed. `processSecretArgs` excludes `''` explicitly, on the grounds that an empty field holds nothing to mint; `redactSecretArgs` then marked the same field as holding something. `redactSecretArgs` now preserves an empty value, so the two agree, and its doc comment says so rather than restating the mechanism.

### Test coverage

The two cases asked for, both verified to fail against the unfixed code first:

- `AIChatManager.test.ts` — `manager.cancel()` after editing a form: the card keeps the edit and the reference. Fails without the fix with `{ token: 'hunter2', note: 'hello' }`.
- `job_args.test.ts` — an empty secret stays empty. Fails without the fix with `{ tok: '<hidden>' }`.

502 passing across the five affected suites. `check:fast` is unchanged at 29 errors, none in a file this branch touches.

### The description

One more stale sentence found and fixed while updating it: the pre-existing-defects section still said the helper-job leak could not be reached from the chat "since a form here opens on references". That has not been true since the revert, and it now says so, pointing at the same explanation as the section below it.

AlexRV12:
Fixed in `0b6a97f`, in a different place from the one suggested, and with a correction on the severity.

### The defect is real and it is new to this branch

Verified rather than taken on faith. All three links hold: `RunArgsFormDisplay` passes `disabled={planMode}` to the form and not `submitting`, so fields stay editable while the mints are in flight; `enforceDisabledDefaults` shallow-copies, so nested arrays are the same references the form edits in place; and `processSecretArgs` walks twice keyed by `JSON.stringify(path)` with the awaits in between.

It is also ours. On `origin/main` `processSecretArgs` iterated top-level `schema.properties` only and assigned `result[key]` inside the same loop iteration — one pass, no path keys, arrays never visited. Main's worst case is a mid-await edit leaving a variable holding the pre-edit value under the right key; cross-row misattribution is not reachable there. Making the walk recursive and path-keyed is what introduced it, and since this PR routes `RunForm`, `ScriptEditor`, `FlowPreviewContent` and `DbtEditor` through that same function, the new failure mode reached every run form in the product rather than only the chat.

### Fixed in the function, not at the call site

The suggestion was to snapshot at the call site or disable the form. `processSecretArgs` is the function that reads its argument twice across an await, all five callers hand it a live-bound object, and a sixth added later would get it wrong the same way — so the snapshot goes at the top of `processSecretArgs`. One place, every caller, including the four outside the chat.

Not `structuredClone`: it throws `DataCloneError` on a Proxy, and the chat binds a `$state` proxy, so that version would have broken the form outright. Confirmed in a browser before committing. It uses `stateSnapshot` (`$state.snapshot`), which is the codebase's existing idiom for exactly this and unwraps the proxy.

The field-disabling is deliberately not included. See below.

### P1 overstates the reachability

Measured on a running instance: creating one ephemeral secret variable takes 7, 9 and 10 ms. The mints are sequential, so the window is roughly N x 9 ms — about 18 ms for two credential rows, 45 ms for five, 90 ms for ten. Human reaction time to register that a click landed is an order of magnitude longer than that, so a deliberate drag-reorder inside the window is not something a person can perform. On a remote deployment a round trip is nearer 50-150 ms, which puts ten rows on a poor connection into the low seconds; that is the only regime where it is reachable at all, and it needs the user to click Run and then immediately go back to reorder a row.

The fix is in regardless, because it is one line, it closes the class rather than this instance, and the failure is silent — no error, and the job permanently records the wrong secret against the wrong row. But the same measurement is why the field-disabling is not: at 20-50 ms it would be an invisible flicker, and the Run and Cancel buttons already disable during a submit.

### Test coverage

`secretArgUtils.test.ts` gains a case that reorders an array of credential rows while the mints are pending and asserts each row keeps its own secret. Keyed by a row's own name rather than its index — the first draft keyed by index and passed against the unfixed code, because substitution by position lands the first-minted reference on index 0 either way. Against the unfixed code the corrected version fails with `expected 'SECOND' to be 'FIRST'`.

Also driven end to end after the change, since a unit test uses plain objects and never exercises the proxy path: a run through the chat form stores `token: "$var:u/admin/secret_arg/…"` in `v2_job.args` and the script reads the real value back, `token_len` matching the typed token.

503 passing across the five affected suites. `check:fast` unchanged at 29 errors, none in a file this branch touches.

PR #11004, opened 2026-09-07
feat(git-sync): sync extra_perms for variables

rubenfiszel:
## Summary

`extra_perms` (granular ACLs) already round-trip through `wmill sync pull` / `push` for
flows, scripts, apps and raw apps (#9162). Variables were left out in both directions:
the tarball hardcoded `ExtraPermsBehavior::Drop` for them, and the CLI's `VariableFile`
had no `extra_perms` field, so a variable's ACLs were invisible to git-sync.

This extends the existing opt-in mechanism to variables, following the flow pattern
exactly: perms are read from the tarball only under `?preserve_extra_perms=true`, and are
pushed through the dedicated `/acls/*` endpoints rather than the variable update body, so
a perm-only edit never rewrites the variable row.

## Changes

- `backend/windmill-api/src/workspaces_export.rs`: variables are serialized with
  `new_kinds_extra_perms` instead of `ExtraPermsBehavior::Drop`, so a variable's ACLs
  appear in the tarball when the caller passes `preserve_extra_perms=true` (which
  `wmill sync pull` does) and only when the map is non-empty — an ACL-less variable's
  yaml is byte-identical to before, so no repo sees a one-time noisy diff.
- `cli/src/commands/variable/variable.ts`: `VariableFile` gains `extra_perms`, and
  `pushVariable` strips it from the body sent to `updateVariable`/`createVariable` and
  applies it via `applyExtraPermsDiff(workspace, "variable", ...)`. The up-to-date
  short-circuit no longer returns early, so ACLs are synced even when the variable body
  is unchanged. An absent `extra_perms` key still means "no opinion" — a checkout that
  predates this change can never revoke UI-managed grants.
- `backend/windmill-api-groups/src/granular_acls.rs`: an ACL add/remove on a `variable`
  now dispatches `DeployedObject::Variable`, like the existing folder/app/raw_app/script/
  flow arms. Without it the repo would go stale on a UI perm change, and the next
  automated `sync push` — now that the yaml carries `extra_perms` — would revoke that
  grant. This makes the export change safe rather than merely complete.
- Comments naming the kinds the `preserve_extra_perms` flag governs were updated to
  include variables.

## Test plan

Added `cli/test/variable_resource_push.test.ts` → "extra_perms round-trips and pushes via
/acls/* without rewriting the variable": pulls a variable with a group grant, asserts the
yaml carries `extra_perms`, downgrades the grant, pushes, and asserts the new perms landed
while `edited_at` and `value` are untouched; then removes the `extra_perms` key entirely
and asserts a push leaves the remote ACLs alone.

Exercised manually against a local instance:

- [x] `sync pull` writes `extra_perms` for a variable with ACLs; a variable without ACLs
      keeps its previous yaml exactly.
- [x] Export without `preserve_extra_perms` still drops the field.
- [x] `sync push` grants, downgrades and revokes; `edited_at` unchanged for perm-only
      edits; audit log records `variables.grant_acl`, not a second `variables.update`.
- [x] Create path (new local yaml with `extra_perms`) and combined value+perms edit.
- [x] Single-file `wmill variable push`, including a secret variable.
- [x] `pull` → `push` round-trip is idempotent (0 changes).
- [x] EE build with git-sync configured against a local Gitea: granting a variable ACL
      through the API dispatches a git-sync job carrying
      `[WM] Variable '<path>' changed permissions` with `path_type: variable`, and the
      repo receives the updated `<path>.variable.yaml`.
- [x] `cli/test/variable_resource_push.test.ts` (12 pass) and
      `cli/test/sync_pull_push.test.ts` (68 pass) green.

Not exercised: the non-admin path where `/acls/*` 403s and `applyExtraPermsDiff` logs a
red non-fatal line — that behavior is shared with flows/scripts/apps and unchanged here.

**Left in draft**: this touches a permission path (granular ACL sync and the git-sync
dispatch for it), so it wants a human look before going ready.

https://claude.ai/code/session_01Hvv5B8VP5Di4dbcCiVyZyE

---
## Summary by cubic
Syncs variable `extra_perms` through `wmill sync pull` / `push` so granular ACLs round-trip for variables like they already do for flows, scripts, and apps.

- Variable ACLs are now included in the tarball under `preserve_extra_perms=true`; ACL-less variables keep byte-identical yaml.
- Perm-only edits go through the `/acls/*` endpoints, so the variable row (value, `edited_at`) is never rewritten.
- A missing `extra_perms` key means "no opinion", so checkouts that predate this change can't revoke UI-managed grants.
- Variable ACL changes now dispatch git-sync deployment metadata so the repo stays current after a UI perm edit.

Added `cli/test/variable_resource_push.test.ts` → "extra_perms round-trips and pushes via /acls/* without rewriting the variable": pulls a variable with a group grant, asserts the yaml carries `extra_perms`, downgrades the grant, pushes, and asserts the new perms landed while `edited_at` and `value` are untouched; a missing `extra_perms` key leaves remote ACLs alone, while a present-but-empty map revokes them.

<sup>Written for commit 05499d56a5584e19b73a9a16993d3d55e93f77e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/11004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-groups/src/granular_acls.rs
+++ b/backend/windmill-api-groups/src/granular_acls.rs
@@ -318,6 +318,19 @@ async fn add_granular_acl(
             )
             .await?
         }
+        "variable" => {
+            handle_deployment_metadata(
+                &authed.email,
+                &authed.username,
+                &db,
+                &w_id,
+                DeployedObject::Variable { path: path.to_string(), parent_path: None },
+                Some(format!("Variable '{}' changed permissions", path)),
+                true,
+                None,
+            )
+            .await?
+        }
         _ => (),
     }
 
@@ -523,6 +536,19 @@ async fn remove_granular_acl(
                     &w_id,
                     DeployedObject::Flow { path: path.to_string(), parent_path: None, version: 0 },
                     Some(format!("Flow '{}' changed permissions", path)),
+                    true,
+                    None,
+                )
+                .await?
+            }
+            "variable" => {
+                handle_deployment_metadata(
+                    &authed.email,
+                    &authed.username,
+                    &db,
+                    &w_id,
+                    DeployedObject::Variable { path: path.to_string(), parent_path: None },
+                    Some(format!("Variable '{}' changed permissions", path)),
                     true,
                     None,
                 )
--- a/backend/windmill-api/src/workspaces_export.rs
+++ b/backend/windmill-api/src/workspaces_export.rs
@@ -346,7 +346,7 @@ pub(crate) struct ArchiveQueryParams {
     default_ts: Option<String>,
     /// Settings format version: "v1" (default) returns legacy flat format, "v2" returns grouped format
     settings_version: Option<String>,
-    /// Opt-in: include `extra_perms` on flow / script / app rows. Default `false`
+    /// Opt-in: include `extra_perms` on script / flow / app / variable rows. Default `false`
     /// so cross-workspace tarball imports do not carry over ACLs referring to
     /// identities that may not exist in the target workspace. `wmill sync pull`
     /// passes `true` to surface ACLs in the git-tracked yaml.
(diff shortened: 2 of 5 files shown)


rubenfiszel:
✅ Review round clean @ 05499d56a5584e19b73a9a16993d3d55e93f77e3 — nit-only verdicts at 2c6140915b66707b59d6acbf364bb8a1ee0bfd92 (codex: Good to merge; claude: nit only); nit addressed in 05499d56a5584e19b73a9a16993d3d55e93f77e3

PR #11008, opened 2026-09-07
fix(flows): stop re-evaluating skip_if once a loop is in progress

### LINKED ISSUE #11007: bug: native while-loop stops after one iteration when combined with skip_if on the outer module
### Describe the bug

A `whileloopflow` module whose **outer** module has `skip_if` (to skip the whole loop based on a preceding step's result) and whose **inner** body module has `stop_after_if` (to decide when to stop iterating) stops after exactly **one iteration**, even when the `stop_after_if` expression evaluates to `false`.

Root cause (found by reading `backend/windmill-worker/src/worker_flow.rs`): `skip_if` is meant to be a one-time entry gate for a module, but the flow stays "at" the same step for the whole lifetime of an in-progress loop, so `push_next_flow_job` re-evaluates `module.skip_if` on every iteration, not just on first entry.

`previous_id` is computed once from the static step index (`flow.modules[step_idx - 1].id`) and keeps pointing at the true preceding top-level step (e.g. `open_mr`) for as long as the flow is "at" the loop's step. But the `results` proxy in `backend/windmill-jseval/src/lib.rs` has a shortcut: when the requested id equals `__previous_id`, it returns `previous_result` (the last completed job's result) directly instead of fetching the real stored result for that id:

```js
if (name === __previous_id && typeof previous_result !== 'undefined') {
    return Promise.resolve(previous_result);
}
```

On the first entry into the loop, `previous_result` really is `open_mr`'s result, so this shortcut is correct. But once the loop is `InProgress`, the *last completed job* is an inner iteration of the loop's own body, not `open_mr` — yet `previous_id` is still (incorrectly, in this context) `"open_mr"`. So `results.open_mr` silently resolves to the last iteration's result instead of `open_mr`'s real result, `skip_if`'s expression evaluates against the wrong value, and the loop's own module gets marked as skipped after a single iteration.

### To reproduce

Import this flow (adjust workspace/permissions as needed) and run it:

```json
{
  "modules": [
    {
      "id": "open_mr",
      "value": {
        "type": "rawscript",
        "language": "python3",
        "content": "def main():\n    return {\"status\": \"mr-aperta\"}\n",
        "input_transforms": {}
      }
    },
    {
      "id": "poll_fix_loop",
      "value": {
        "type": "whileloopflow",
        "skip_failures": false,
        "modules": [
          {
            "id": "poll_iter",
            "value": {
              "type": "rawscript",
              "language": "python3",
              "content": "def main(i=0):\n    return {\"done\": i + 1 >= 6, \"iter\": i + 1}\n",
              "input_transforms": {
                "i": {"type": "javascript", "expr": "flow_input.iter.index"}
              }
            },
            "stop_after_if": {
              "expr": "result.done === true",
              "skip_if_stopped": true
            }
          }
        ]
      },
      "skip_if": {
        "expr": "results.open_mr.status !== 'mr-aperta'"
      }
    }
  ]
}
```

1. Run the flow.
2. `open_mr` succeeds with `{"status": "mr-aperta"}`, so `skip_if`'s expression (`results.open_mr.status !== 'mr-aperta'`) should evaluate to `false` and the loop should run to completion.
3. `poll_iter` runs once (`{"done": false, "iter": 1}` — `done` is `false`, so `stop_after_if` should be `false` too).
4. The flow stops there. `poll_fix_loop`'s module status shows `"skipped": true`, and the overall flow result is just `{"done": false, "iter": 1}` instead of 6 iterations.

Reproduced deterministically (4/4 runs) on both `v1.801.0` and current `main` (`670404f`, one commit past `v1.804.0`). A version without `skip_if` on the outer module doesn't trigger this — only the combination of outer `skip_if` + inner `stop_after_if` does.

### Expected behavior

The loop should run until `poll_iter`'s `stop_after_if` (`result.done === true`) is actually `true`, i.e. 6 iterations, with `poll_fix_loop` not marked as skipped.

### Application version

CE v1.804.0-1-g670404f (also reproduced on CE v1.801.0)

### Additional context

I have a fix (PR incoming): skip re-evaluating `module.skip_if` in `push_next_flow_job` once `status_module` is already `FlowStatusModule::InProgress` — `skip_if` only makes sense as an entry gate, and a module that's already running can't meaningfully be "skipped" anyway. Verified locally: 4/4 runs fail without the fix, 3/3 runs pass with it (built from source, `cargo check -p windmill-worker` clean).

We hit this in production as a while-loop nested after several real preceding steps in a larger flow (poll pipeline status / code review, retry on failure, stop when clean) and ended up working around it by moving the loop into a single trusted Python step instead of a native `whileloopflow`, before tracking down the actual cause.

davidemodolo:
Fixes #11007

## What

`skip_if` is meant as a one-time entry gate for a module, but the engine re-evaluates it on every iteration of an in-progress `whileloopflow`/`forloopflow` (the flow stays "at" the same step for the loop's whole lifetime).

`previous_id` is derived once from the static step index and keeps pointing at the module preceding this step for as long as we're still "at" it. Once the loop is `InProgress`, the last completed job is an inner iteration, not that preceding module — but the `results` proxy in `windmill-jseval` aliases `results.<previous_id>` to the last job's result as a shortcut. So a `skip_if` referencing the true preceding step's result silently reads the wrong value on the second and later evaluations, and can flip the loop's own module to `skipped` after a single iteration even though nothing should stop it.

## Fix

Skip the `skip_if` check once `status_module` is already `InProgress`: it only makes sense as an entry gate, and a module that's already running can't meaningfully be skipped anyway.

## Testing

- `cargo check -p windmill-worker`: clean.
- Built from source and ran the minimal repro in #11007 against both the unpatched and patched binary:
  - Without the fix: 4/4 runs stop after 1 iteration, `poll_fix_loop` marked `skipped: true`.
  - With the fix: 3/3 runs complete all 6 iterations, `poll_fix_loop` marked `skipped: false`.

---
## Summary by cubic
Fixes #11007. `skip_if` was re-evaluated on every iteration of an in-progress while/for loop, reading the wrong job's result and potentially flipping the module to skipped. Now the check only runs in entry states (`WaitingForPriorSteps`, `WaitingForEvents`, `WaitingForExecutor`), preserving the one-time entry gate; a restart-at-iteration also resumes without re-gating. Adds a regression test covering the loop-entry case.

<sup>Written for commit 93a7d9c459bb173629889c011a440c37d86a0e65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/11008?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/worker.rs
+++ b/backend/tests/worker.rs
@@ -5637,6 +5637,83 @@ async fn test_whileloop_propagates_inner_iterator_eval_failure(
 
 #[cfg(all(feature = "quickjs", feature = "python"))]
 #[sqlx::test(fixtures("base"))]
+async fn test_whileloop_skip_if_evaluated_once_at_entry(db: Pool<Postgres>) -> anyhow::Result<()> {
+    initialize_tracing().await;
+
+    // Regression test for #11007: `skip_if` on a while-loop module must be
+    // evaluated once, at loop entry, using the preceding step's result.
+    // Re-evaluating it on every iteration aliases `results.first` to the
+    // previous iteration's own result instead, which here lacks `.ok` and
+    // makes `skip_if` incorrectly turn true after the first iteration.
+    let port = 123;
+    let flow: FlowValue = serde_json::from_value(serde_json::json!({
+        "modules": [
+            {
+                "id": "first",
+                "value": {
+                    "type": "rawscript",
+                    "language": "python3",
+                    "content": "def main(): return {\"ok\": True}",
+                },
+            },
+            {
+                "id": "outer",
+                "value": {
+                    "type": "whileloopflow",
+                    "skip_failures": false,
+                    "modules": [
+                        {
+                            "id": "inner",
+                            "value": {
+                                "input_transforms": {
+                                    "i": {
+                                        "type": "javascript",
+                                        "expr": "flow_input.iter.index",
+                                    },
+                                },
+                                "type": "rawscript",
+                                "language": "python3",
+                                "content": "def main(i): return i",
+                            },
+                        },
+                    ],
+                },
+                "skip_if": { "expr": "!results.first.ok" },
+                "stop_after_if": {
+                    "expr": "result >= 2",
+                    "skip_if_stopped": false,
+                },
+            },
+        ],
+    }))
+    .unwrap();
+    let job = JobPayload::RawFlow { value: flow, path: None, restarted_from: None };
+
+    let cjob = RunJob::from(job).run_until_complete(&db, false, port).await;
+
+    assert!(cjob.success, "flow should succeed");
+
+    let outer_module = get_module(&cjob, "outer").expect("outer module status");
+    match outer_module {
+        windmill_common::flow_status::FlowStatusModule::Success { skipped, flow_jobs, .. } => {
+            assert!(
+                !skipped,
+                "while-loop must not be skipped: skip_if should only run once, at entry"
+            );
+            assert_eq!(
+                flow_jobs.map(|v| v.len()),
+                Some(3),
+                "while-loop should run 3 iterations before stop_after_if halts it"
+            );
+        }
+        other => panic!("expected outer module to be Success, got {other:?}"),
+    }
+
+    Ok(())
+}
+
+#[cfg(all(feature = "quickjs", feature = "python"))]
+#[sqlx::test(fixtures("base"))]
 async fn test_stop_after_all_iters_if_bad_expr_parallel_branchall(
     db: Pool<Postgres>,
 ) -> anyhow::Result<()> {
--- a/backend/windmill-worker/src/worker_flow.rs
+++ b/backend/windmill-worker/src/worker_flow.rs
@@ -3891,7 +3891,14 @@ async fn push_next_flow_job(
 
     drop(resume_messages);
 
-    let is_skipped = if let Some(skip_if) = &module.skip_if {
+    // `skip_if` is a one-time entry gate: once `InProgress` (looping), the last
+    // completed job is an inner iteration, not `previous_id`'s module, so
+    // re-evaluating here would alias `results.<previous_id>` to the wrong result.
+    let is_skipped = if let Some(skip_if) = module
+        .skip_if
+        .as_ref()
+        .filter(|_| !matches!(status_module, FlowStatusModule::InProgress { .. }))
+    {
         let idcontext = get_transform_context(&flow_job, previous_id.as_str(), &status);
         let skip_if_res = compute_bool_from_expr(
             &skip_if.expr,
--- a/backend/windmill-worker/src/worker_flow.rs
+++ b/backend/windmill-worker/src/worker_flow.rs
@@ -3891,14 +3891,18 @@ async fn push_next_flow_job(
 
     drop(resume_messages);
 
-    // `skip_if` is a one-time entry gate: once `InProgress` (looping), the last
-    // completed job is an inner iteration, not `previous_id`'s module, so
-    // re-evaluating here would alias `results.<previous_id>` to the wrong result.
-    let is_skipped = if let Some(skip_if) = module
-        .skip_if
-        .as_ref()
-        .filter(|_| !matches!(status_module, FlowStatusModule::InProgress { .. }))
-    {
+    // `skip_if` is a one-time entry gate, so only first-entry statuses evaluate it.
+    // Once the module is looping, the last completed job is an inner iteration, not
+    // `previous_id`'s, and re-evaluating would alias `results.<previous_id>` to it.
+    // A restart-at-iteration also enters as `InProgress`: it resumes without re-gating.
+    let is_skipped = if let Some(skip_if) = module.skip_if.as_ref().filter(|_| {
+        matches!(
+            status_module,
+            FlowStatusModule::WaitingForPriorSteps { .. }
+                | FlowStatusModule::WaitingForEvents { .. }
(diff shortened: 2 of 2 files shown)


davidemodolo:

rubenfiszel (changes requested):
Thanks for the detailed writeup in the description, the diagnosis is correct and the fix is the right shape. I traced the surrounding paths and the core change is sound; two things I'd like addressed before merge.

## What I verified as clean

- No normal re-entry loses its `skip_if`. Retries rewrite `status_module` back to `WaitingForPriorSteps` (`worker_flow.rs:3757`) and restore `arc_last_job_result` to the preceding step's result, suspend/approval resume arrives as `WaitingForEvents`, and a freshly pushed step is `WaitingForPriorSteps`/`WaitingForExecutor`. All still pass the gate. `Success` returns early at the top of the function, and `Failure` only reaches this point on the failure-module path where `module`/`step` have already been rewritten.
- No sibling call site has the same unguarded shape. `sleep` (`:3719`) and the BranchOne/BranchAll predicates (`:5561`, `:5679`) are already restricted to first-entry statuses. `stop_after_if` / `stop_after_all_iters_if` run through `get_id_ctx_for_expr` with `previous_id = ""` (`:332`), so the `results.<previous_id>` aliasing cannot fire there. `skip_if` really was the only `previous_id`-bearing predicate reachable with a non-fresh status.
- Dropping the per-iteration `get_transform_context` + jseval is also a small win on every sequential loop/branch-all iteration whose module carries `skip_if`.

## Requested change 1: `InProgress` is also the *entry* state for restart-at-iteration

`InProgress` is not only reached by continuation. A flow restarted at a specific iteration or branch (`branch_or_iteration_n >= 1`, which the run page's nested "Re-start from" also sends) has its target module **constructed** as `FlowStatusModule::InProgress` before the run starts: `create_restarted_module` in `backend/windmill-queue/src/jobs.rs:7356` (BranchAll) and `:7401` (ForloopFlow).

On that run, the very first `push_next_flow_job` sees `InProgress`, so `module.skip_if` is now never evaluated for a restarted loop or branch-all. Before this PR it was evaluated exactly once there, with a *correct* context: `arc_last_job_result` comes from `get_previous_job_result` and `previous_id` really is the preceding module, so the aliasing bug this PR fixes does not apply on that path.

Not re-checking the gate on a run the user explicitly asked to resume is arguably the better behavior, so I'm not asking you to preserve the old semantics. But right now it is an unremarked and untested change to that path (`backend/tests/restart_at_step_api.rs:57` restarts with `branch_or_iteration_n: 1`), and the new comment's stated reason is factually wrong for it: on a restart entry there is no last completed job at all.

Either of these resolves it:

- Extend the comment to cover the second entry, e.g. "...or resuming a restart-at-iteration, where the gate was already decided by the original run", or
- Make the condition express "this module has already pushed a child in this run" rather than "status is `InProgress`".

Worth noting the two sibling gates in the same file already use the positive allowlist form for exactly this distinction (`WaitingForPriorSteps | WaitingForEvents | WaitingForExecutor` at `:5561` and `:5679`, and the `sleep` gate at `:3719`). Matching that shape would be more consistent, though it lands on the same behavior for the restart case.

## Requested change 2: a regression test

This is cheap here and the harness already exists. `backend/tests/worker.rs:5557` (`test_whileloop_propagates_inner_iterator_eval_failure`) is a direct template: it builds an inline `whileloopflow` `FlowValue` and runs it with `RunJob::from(JobPayload::RawFlow { .. }).run_until_complete(&db, false, port)` under `#[cfg(all(feature = "quickjs", feature = "python"))]`.

The repro from #11007 maps onto it almost verbatim: a first step, an outer `whileloopflow` carrying `skip_if: results.<first_step>...`, an inner step with `stop_after_if` bounding the loop at ~3 iterations, then assert the loop module ends with `skipped == false` and `flow_jobs.len() == 3`. Without the fix that fails at 1 iteration with `skipped: true`, so it pins exactly the behavior this PR changes. One such test is enough; no need for a matrix.

If you go the "explicitly intend the restart behavior" route on point 1, a second case in `backend/tests/restart_at_step_api.rs` restarting at `branch_or_iteration_n: 1` on a loop whose module has `skip_if` would pin that contract too, but that one is optional and a maintainer call.

davidemodolo:
Thanks for the review.

**Requested change 1**: rewrote the condition as a positive allowlist matching the `BranchOne`/`BranchAll` predicate gates (`WaitingForPriorSteps | WaitingForEvents | WaitingForExecutor`), and updated the comment to state the restart-at-iteration case separately instead of folding it into the aliasing reason, which doesn't apply there as you noted.

**Requested change 2**: added `test_whileloop_skip_if_evaluated_once_at_entry` in `backend/tests/worker.rs`, modeled on `test_whileloop_propagates_inner_iterator_eval_failure` as suggested. Confirmed it fails on the pre-fix code (loop marked `skipped: true` after one iteration) and passes with the fix.

hugocasa (approved):
LGTM

davidemodolo:
@rubenfiszel friendly bump; this has hugocasa's approval and it's been a few days since the re-request. Both requested changes were addressed on 93a7d9c (positive allowlist matching the sibling gates + the regression test). Let me know if anything else is needed.

PR #11010, opened 2026-09-07
fix: let a draft-only schedule, trigger or resource be deleted

rubenfiszel:
## Summary

A schedule that has never been deployed — it exists only as a per-user row in the `draft` table — is listed on the schedules page, because `list_schedules` synthesizes rows for such paths when `include_draft_only=true`. Its DELETE route, however, returned `NotFound` whenever the `schedule` table had no row, so the entry was visible and impossible to remove:

```
DELETE /api/w/<ws>/schedules/delete/u/me/my_draft_schedule
Not found: Schedule u/me/my_draft_schedule not found   HTTP 404
```

The gap is not schedules-only. Fourteen list pages pass `includeDraftOnly`; auditing each backend delete found the same strand in the generic trigger handler (covering all 11 trigger kinds) and in `delete_resource`. Variables were already correct — their delete tolerates a missing row and clears drafts at the path.

Nothing was needed on the frontend: the row already renders a "Draft only" badge, a disabled enable/disable toggle and a greyed-out Runs button. Only the delete was broken.

## Changes

- Add `delete_draft_only_for_path` in `windmill-common/src/user_drafts.rs`: deletes the caller's own draft at a path with no deployed row. A `NOT EXISTS` against `UserDraftItemKind::deployed_table()` means a deployed row's draft is never touched, so a route can call it on its not-found branch without having to distinguish "absent" from "hidden by RLS".
- Call it on the not-found branch of `delete_schedule` and the generic `delete_trigger`.
- Call it in `delete_resource` ahead of both the transaction and the deployment-rule check. Ahead of the transaction because that route's not-found branch (`not_found_if_none`) sits mid-transaction with the linked-variable cascade already staged; ahead of the deployment rules because discarding one's own draft is not a deployment — gating it there would leave the row undeletable in a workspace with `DisableDirectDeployment`, while the same discard is already ungated on the drafts routes.

### Authorization

The helper is scoped to `email = $4`, the caller's own row. It deliberately stops short of the list synthesis, which additionally surfaces **legacy** (`email IS NULL`, pre-per-user-drafts) rows: those are owned by nobody and keep their write gate, which `update_draft` documents and enforces via `require_can_write_path`. Discarding a legacy draft stays on the `update_draft { legacy: true }` and admin-only migrate routes that apply that gate. Requiring no permission check here is the same reasoning `update_draft` uses to exempt an own-discard: an email-scoped row belongs to the caller, who can always discard it even after losing write access to the path.

**Known residual:** a *legacy* draft-only row still appears in these lists and still 404s on Delete. That is strictly smaller than before (every draft-only row 404'd previously; now only legacy ones do) and a correctly-gated door exists for it, but the symptom is indistinguishable from "the fix didn't land". Closing it properly means making `require_can_write_path` reachable from `windmill-api-schedule` / `windmill-trigger` / `windmill-store`, which today would be a dependency cycle on `windmill-api` — deliberately out of scope here.

## Test plan

- [x] `cargo test -p windmill-api-integration-tests --test schedules --test protection_rules --test resources --test drafts` — all green
- [x] `test_delete_draft_only_schedule`: drives the real route stack (save draft → list with `include_draft_only=true` → DELETE → assert the `draft` row is gone → DELETE again → 404), plus a legacy (`email IS NULL`) row that must survive its DELETE with a 404
- [x] `test_protection_rules` §4b: with `DisableDirectDeployment` active, a non-admin can still delete their draft-only resource, while creating a deployed one at the same path stays blocked
- [x] Both new assertions falsified: reintroducing `OR email IS NULL` fails the legacy assertion; moving the fast path back below `check_deploy_rules` fails the protection-rule assertion
- [x] Exercised on a running instance — schedule and resource draft-only deletes return 200 with the draft row gone; deployed schedules and resources carrying drafts still delete normally; the generic trigger path checked against `http_triggers` after restarting the backend with `--features quickjs,http_trigger`
- [x] Driven through the UI with Playwright: the draft-only row disappears from `/schedules` on Delete and stays gone across a reload
- [x] Legacy row checked live: DELETE returns 404 and the row survives

No `.sqlx` regeneration: the new query is the untyped builder, since the table name comes from the `deployed_table()` enum rather than a literal.

https://claude.ai/code/session_012SV5kjTis3AFtTx2nW2VRi

---
## Summary by cubic
Fixes deletion of draft-only schedules, triggers, and resources so entries visible in lists can actually be removed instead of returning 404 and leaving a stuck row.

- Adds `delete_draft_only_for_path`, which deletes the caller's own draft only when no deployed row exists at that path.
- Calls it on the not-found branch of `delete_schedule` and `delete_trigger`, and ahead of deploy-rule checks in `delete_resource`, so discarding a draft isn't gated as a deployment.
- Legacy drafts (`email IS NULL`) still 404 on delete and remain gated on the drafts routes; this is a known residual, smaller than before since every draft-only row previously 404'd.
- Tests cover the real route stack for schedules and resources, including `DisableDirectDeployment` workspaces and legacy-row survival.

<sup>Written for commit 31f1a1bb8c3c6c689c0ede7f9eeaa73ab0652b82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/11010?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/windmill-api-integration-tests/tests/protection_rules.rs
+++ b/backend/windmill-api-integration-tests/tests/protection_rules.rs
@@ -124,6 +124,72 @@ async fn test_protection_rules(db: Pool<Postgres>) -> anyhow::Result<()> {
     assert!(!resp.status().is_success(), "Non-admin should be blocked from flows: {}", resp.status());
 
     // ========================================
+    // 4b. ...but a draft-only resource stays deletable: nothing is deployed at
+    // its path, so its DELETE is a draft discard rather than a deployment.
+    // ========================================
+
+    let draft_only_path = "u/test-user-2/draft_only_resource";
+    let resp = authed(
+        client().post(format!("{base}/drafts/update/resource/{draft_only_path}")),
+        "SECRET_TOKEN_2",
+    )
+    .json(&json!({ "value": {
+        "path": draft_only_path,
+        "value": { "a": 1 },
+        "resource_type": "c_test",
+        "description": ""
+    }}))
+    .send()
+    .await?;
+    assert_eq!(
+        resp.status(),
+        200,
+        "Non-admin should save a draft: {}",
+        resp.text().await?
+    );
+
+    let resp = authed(
+        client().delete(format!("{base}/resources/delete/{draft_only_path}")),
+        "SECRET_TOKEN_2",
+    )
+    .send()
+    .await?;
+    assert_eq!(
+        resp.status(),
+        200,
+        "Draft-only delete should not be gated by the deploy rules: {}",
+        resp.text().await?
+    );
+
+    let remaining: i64 = sqlx::query_scalar(
+        "SELECT count(*) FROM draft WHERE workspace_id = 'test-workspace' AND path = $1 \
+         AND typ = 'resource'::DRAFT_KIND",
+    )
+    .bind(draft_only_path)
+    .fetch_one(&db)
+    .await?;
+    assert_eq!(remaining, 0, "the draft should be gone");
+
+    // The gate itself is still there for a DEPLOYED resource at the same path.
+    let resp = authed(
+        client().post(format!("{base}/resources/create")),
+        "SECRET_TOKEN_2",
+    )
+    .json(&json!({
+        "path": draft_only_path,
+        "value": { "a": 1 },
+        "resource_type": "c_test",
+        "description": ""
+    }))
+    .send()
+    .await?;
+    assert!(
+        !resp.status().is_success(),
+        "Non-admin should still be blocked from creating a resource: {}",
+        resp.status()
+    );
+
+    // ========================================
     // 5. Admin bypasses protection rule
     // ========================================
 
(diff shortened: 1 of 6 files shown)


rubenfiszel:
✅ Review round clean @ 54920b4f979941dbf5f91ea2b13d99126b7e05bd — nit-only verdicts at 54920b4f (Claude: Good to merge; Codex: mergeable with one P2 on comment length / drafting history; Pi posted no findings); the P2 is addressed in 31f1a1bb8c3c6c689c0ede7f9eeaa73ab0652b82, comments only, no behavior change.

PR #11013, opened 2026-09-08
feat: retry a workflow-as-code task from its task options

rubenfiszel:
## Summary

A workflow-as-code task had no way to retry: one flaky HTTP call failed the whole workflow, and the only answer was a hand-written `for` loop with `try`/`catch` and `sleep` — the shape our own docs recommend. Whole-workflow retries already exist (put a retry on the schedule or the trigger, or on the flow step when the workflow runs as one), so this fills the per-task hole, which is where every comparable engine puts retry.

`TaskOptions.retry` / `@task(retry={...})` now re-dispatches a failed task:

```ts
const callApi = task(fetchOrders, { retry: { attempts: 3, delay: 30, multiplier: 2 } });
```
```python
@task(retry={"attempts": 3, "delay": 30, "multiplier": 2})
async def call_api(payload: dict): ...
```

`attempts` counts retries after the first failure, the same way the flow and schedule retry settings count them. `delay` is the seconds before the first retry, `multiplier` grows it per attempt (1, the default, keeps it constant), `max_delay` caps it.

No backend change. The policy is never persisted: `WorkflowCtx._nextStep` / `_next_step` runs one pass per attempt and re-derives from the replay checkpoint alone which attempt comes next, so a checkpointed failure with attempts left either raises a `mode: "sleep"` suspend for the backoff or dispatches the next attempt, and only the exhausted case rejects into the workflow body. The backoff is therefore a durable sleep: a retrying task holds no worker while it waits.

Keys for the attempts after the first hang off the first attempt's key (`fetch#2`, backoff sleeps `fetch#retry2`) instead of being allocated from the name counter. Allocating would shift the keys of steps dispatched beside it: in `Promise.all([t(1), t(2)])` a retry of the first call would take `t_2` and read the second call's result as its own. They are still claimed through the allocator, since `step()` names are arbitrary strings and a step really can be called `fetch#2`.

Known limitation, documented on `TaskRetry` and in the generated SDK docs: a workflow sleeps once per round, so tasks backing off in the same fan-out wait one after another rather than together (three `delay: 30` retries from one `parallel()` come back after 90s). Removing that needs a per-step `scheduled_for` on `WacStepDispatch`, which is a backend change and out of scope here. Retries with no delay all go out in one round.

## Changes

- `typescript-client/client.ts`: `TaskRetry` interface, `retry` on `TaskOptions`, the attempt loop and backoff suspend in `WorkflowCtx._nextStep`, `_attemptKey` / `_retryBackoff`
- `python-client/wmill/wmill/client.py`: the same in `WorkflowCtx._next_step`, plus a `retry` keyword on `task`, `task_script` and `task_flow`
- Tests in both SDKs against the real client: the backoff sequence through to exhaustion, a retry that succeeds, and the sibling-key guard
- `system_prompts/generate.py` emits `TaskRetry` alongside `TaskOptions`, so the generated agent prompts carry the policy rather than a dangling type name; generated files regenerated
- `docs/wac-sdk-e2e.md`: the SDK-injection recipe was missing a step — a bun job keeps running the bundle snapshot it built from the published SDK until `$WINDMILL_DIR/cache/bun` is dropped

## Test plan

- [x] `bun test tests/` in `typescript-client` (170 pass)
- [x] `PYTHONPATH=. uv run --frozen --python 3.14 pytest tests/ -q` in `python-client/wmill` (115 pass)
- [x] `bash system_prompts/check-freshness.sh`
- [x] End to end on a real worker for **both** SDKs, with the built clients injected into the worker cache and every scenario run once against the published client and once against the patched one:
  - bun, `attempts: 2, delay: 5, multiplier: 2`, always failing: 3 child jobs 5s and 10s apart, then the workflow failed. Parent held a worker for 20ms of a 15.5s run, so the backoff really is a durable sleep.
  - bun, `attempts: 1`, no delay, fails once then succeeds: retried immediately and returned its value.
  - bun, `attempts: 1, delay: 5`, task called **without `await`**: 2 attempts 5s apart (the published client, and an earlier draft of this change, ran one and reported the workflow complete).
  - python, `attempts: 2, delay: 5, multiplier: 2`, always failing: same 3 attempts 5s and 10s apart, checkpoint keys `boom`, `boom#2`, `boom#3`, `boom#retry2`, `boom#retry3`, parent held a worker for 54ms of a 16s run.
  - Run page timeline for both clients reads `boom → sleep (5s) → boom#2 → sleep (10s) → boom#3`, so `#` in a step key travels the checkpoint, the `_step/<key>` timeline entries and the child `_executing_key` intact.

---
## Summary by cubic
Adds per-task retry to workflow-as-code tasks so a flaky call no longer fails the whole workflow. Previously only whole-workflow retries existed; failed tasks now re-dispatch with a configurable backoff via `retry` on `TaskOptions` and `@task`.

- `attempts` counts retries after the first failure; `delay`, `multiplier`, and `max_delay` shape the backoff, and an explicit `multiplier: 0` is honored so later retries go out immediately.
- The backoff is a durable sleep, so a retrying task holds no worker while waiting.
- All attempt and backoff keys are claimed up front from the first attempt's key, so retries never shift sibling step keys and an inline `step()` cannot alias an attempt.
- `attempts` is capped at 100, and unknown keys or out-of-range values are rejected where the policy is written.
- Tasks backing off in the same fan-out wait serially, because a workflow sleeps once per round.
- Retries also fire for task calls the body never awaits, since the runner flushes pending steps each round.
- Requires no backend change; the policy is never persisted and each round re-derives the next attempt from the checkpoint.
- The SDK-injection recipe in `docs/wac-sdk-e2e.md` now also drops the bun bundle cache so patched clients reach bun jobs.

<sup>Written for commit e347fd63e6f2c15c7e5de9a211431ebf18b1b530. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/windmill-labs/windmill/pull/11013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rubenfiszel:
Addressed round 1.

**P1 (codex) / P2 (claude) — attempt keys could alias a step allocated later.** Both reviewers hit the same ordering: `const p = t(1); await step("t#2", …)`. The decoy is invisible to `_usedKeys` when the retry derives its key, but present in `completed` from an earlier round, so the attempt read the step's value.

Fixed by claiming every key the call can ever use (attempts and backoffs) up front at the first attempt, rather than deriving them as attempts are spent. Whichever of the two the body allocates second is then the one renamed, identically in every round — instead of it depending on how many attempts had run by then. I did not take the `#`-rejection route: `step()` names are already in use in deployed workflows, and turning one into a hard error mid-run seemed worse than renaming it.

Verified on a real worker, the scenario codex asked for: retrying `t` called before `step("t#2", …)`, first attempt failing. Result `{"t": {"x": 3, "calls": 2}, "decoy": "not an attempt"}`, two child jobs, checkpoint keys `t`, `t#2`, `t#2_2` — a real retry ran, and neither step consumed or re-executed the other. Against the published client the same script fails with one child and no retry.

Regression tests for both orderings in both SDKs.

**Nit — untyped `retry` dict in python.** Rather than a `TypedDict` (invisible at runtime, and `pyproject` still declares `^3.7`), unknown keys now raise at decoration: `@task(retry={"max_delay_s": 300})` fails immediately instead of silently backing off on a policy nobody wrote.

**Nit — fan-out scale.** The doc now states the cost as the sum of every backoff pending in the round, growing with both fan-out width and `attempts`, instead of illustrating with N=3.

**Nit — exhaustion error doesn't say retries ran.** Not taking this one: `Task 'x' failed` is only the *fallback* message, used when the failure marker carries none, so a message mentioning attempts would almost never be the one a user sees. The `TaskError` already carries `step_key` of the failing attempt (`boom#3`), which names the attempt directly, and the run page lists every attempt as its own step.

Also re-ran the backoff scenario end to end after the refactor: 3 attempts 5s and 10s apart, keys `boom`, `boom#2`, `boom#3`, `boom#retry2`, `boom#retry3`, parent holding a worker for 17ms of the run.

rubenfiszel:
Addressed round 2.

**P1 — unbounded `attempts` preallocates keys.** Correct, and my doing: since attempts claim their keys up front, `attempts: Infinity` is now a hang rather than a very patient workflow, and a large finite count allocates two keys per retry on every replay. `attempts` is now rejected where the policy is written (`task`, `taskScript`, `taskFlow`, and the python `@task` family) unless it is a whole number in 0..100, and `_nextStep` clamps as well, so a policy that reached it another way cannot spin the loop. Tests in both SDKs cover `10_000`, `-1`, `2.5`, `Infinity` and `NaN`.

**P1 — unawaited exhaustion reports success.** Not taking this one: it is main's behaviour for unawaited tasks, not something this PR introduces. `_nextStep` has always handed a failed step back as a lazy rejecting thenable, so a task the body never awaits has never surfaced its failure — the body returned `"done"` without ever looking at the task, and the runner reports what the body returned.

Measured, rather than reasoned: the unawaited scenario in this PR's test plan, run against the **published** `windmill-client@1.805.0` (no retry support at all), completes as `success | "done"` with one failed child. Same script, same worker, patched client: two attempts five seconds apart, still `success | "done"`. The retries are what this PR adds; the unobserved terminal failure is unchanged on either side.

Making an unawaited failure fail the run is a change to unawaited-task semantics generally. Doing it only when a retry policy is present would be worse — adding `retry` to a task would silently change whether its failure can kill the workflow — and doing it for all unawaited tasks would change behaviour for workflows that fire-and-forget deliberately. Either belongs in its own PR, and I'm happy to open one if you'd like it.

No e2e re-run this round: the only behaviour change is a guard on a value that previously had none, and the retry paths themselves are byte-identical for any policy that was valid before.

rubenfiszel:
Addressed round 3 — all four taken.

**P1 — python swallowed an explicit `multiplier: 0`.** Real parity bug: `retry.get("multiplier") or 1` read a deliberate `0` as the default, so `{attempts: 2, delay: 30, multiplier: 0}` slept 30s before both retries in python and 30s then none in TS. Now an explicit `None` check, and a parity regression test in both SDKs pinning 30s before the first retry and no sleep round before the second.

**P2 — the 0..100 bound wasn't documented.** Added to `TaskRetry.attempts` and to the python `task` docstring, so the generated SDK prompts carry it too.

**P2 — the key-allocation comment blocks were 10 lines against AGENTS.md's ≤4.** Condensed to 4 in both SDKs; the invariant is unchanged.

**Nit (claude) — `retry` without `attempts` was a silent no-op in python and a hard error in TS.** `attempts` is now required in python as well, so a policy that would never retry is rejected where it is written rather than quietly doing nothing.

Suites: bun 177 pass, pytest 123 pass, `check-freshness.sh` clean.

No e2e re-run: the zero-multiplier path is arithmetic with unit coverage on both sides, and nothing else here is behavioural.

On the unawaited-exhaustion P1 I pushed back on last round: it now has its own worktree and session (branch `wac-unawaited-task-failure`), briefed with the measurement, the seam (`_takePendingSuspend`-shaped) and the three semantics decisions it needs from a human — how loud, whether deliberate fire-and-forget keeps an escape hatch, and one SDK or both.

rubenfiszel:
Round 4 came back nit-only from both reviewers. Both nits addressed.

**codex — zero-multiplier behaviour unexercised on a worker.** Fair; I'd skipped it as arithmetic. Ran it on both SDKs, `{attempts: 2, delay: 5, multiplier: 0}`, always failing:

| | attempt 1 | attempt 2 | attempt 3 |
|---|---|---|---|
| bun | 09:26:57.059 | 09:27:02.531 (+5.5s) | 09:27:02.703 (+0.17s) |
| python | 09:26:57.302 | 09:27:02.597 (+5.3s) | 09:27:02.767 (+0.17s) |

Checkpoint keys on both runs: `boom`, `boom#2`, `boom#3`, `boom#retry2` — one backoff key, not two. The second backoff is `5 * 0`, so no sleep step is created at all and the third attempt goes out immediately. That is the timeline the review asked for, and the two clients agree.

**claude — the unknown-key rule was stated twice in the python docstring**, and rode into nine generated prompt sites. Removed the duplicate; the surviving sentence keeps the required-`attempts`, range and unknown-key rules in one place. Generated files regenerated.

Head is now `e347fd63e6f2c15c7e5de9a211431ebf18b1b530`.

rubenfiszel:
✅ Review round clean @ c4d409f8d3ae16ee6e51d04653faa1edae19fd84 — nit-only verdicts at that round SHA; nits addressed in e347fd63e6f2c15c7e5de9a211431ebf18b1b530 (docstring dedupe) and by the zero-multiplier worker run reported above

PR #11017, opened 2026-09-08
feat: report a WAC task failure the workflow body never awaited

rubenfiszel:
## Summary

In workflows as code, a task called without `await` is still dispatched, still runs, and still
fails: the failure lands in `completed_steps` and the child job shows red on the run page. But the
body never observes it, so the round reports `complete` and **the workflow goes green**.

```ts
export const main = workflow(async () => {
  notify();            // fails
  return 'done';
});
```

Before: `success | "done"`, with one red child. After: same result, plus a line in the job log
naming the task and its error.

This is longstanding behaviour, not a regression. Both runners already frame the case as a
mistake rather than a feature — `// Flush any unawaited tasks (e.g. forgotten await on last
statement)` in the generated `wrapper.mjs` and in the python client — so if the flush exists to
rescue a forgotten `await`, swallowing that task's failure is the wrong default.

**Only the TypeScript client is affected.** Python's `@task` wrapper is a sync `def` calling
`_next_step`, which raises on the `__wmill_error` marker whether or not the call was awaited, so a
python workflow already fails this case. TypeScript's `_nextStep` instead hands back a *rejecting
thenable*, and nothing ever drives it.

The result deliberately stays green: failing the run would turn workflows that pass today red on
upgrade, which is a semantics change worth making separately if at all. This PR only makes the
failure visible.

## Changes

- `WorkflowCtx` records a failed task's error when it hands back the rejecting thenable, and drops
  it again on the first `.then()` — so `await` and `Promise.allSettled` both count as observing it,
  including when the body holds the handle and awaits it further down.
- `_warnUnobservedTaskFailures()` reports whatever is left, on stdout so it stays in sequence with
  the other `--- WAC:` markers (the two streams are merged without preserving order).
- The generated wrapper calls it on both paths that end a round for good — the `complete` return
  and the failure rethrow — but not on a round that only dispatches more steps, which would report
  the same failure again after the next round re-registers it from the checkpoint.
- A child round replays the whole body to reach one step, so it re-registers every checkpointed
  failure; the ctx suppresses reporting there rather than duplicating it into every child job's log.
- Registration sits after the retry check merged from #11013, so only the attempt actually handed
  back to the body counts — a failure a retry recovered from is never reported — and it is keyed by
  the task's base key rather than the final attempt's (`notify`, not `notify#3`).
- No escape hatch for deliberate fire-and-forget: python has shipped without one, and the result
  stays successful either way. Note the handle is a bare thenable, so the idiom is
  `Promise.resolve(notify()).catch(…)` rather than `notify().catch(…)`.

## Test plan

- [x] `bun test typescript-client/tests/` — 181 pass, including the new
      `workflow_unawaited_failure.test.ts`: reported when unobserved, silent when awaited later,
      silent when a retry recovered from it, silent in a child round. The retry case was checked
      against a deliberately broken client (registration hoisted above the retry check) to confirm
      it actually fails.
- [x] Real bun worker, SDK injected per `docs/wac-sdk-e2e.md`, run against the published client and
      then the patched one, and re-run after merging main:
  - unawaited failing task — published: `success | "done"`, log has only `--- WAC: complete ---`;
    patched: same result, log gains
    `--- WAC: task 'notify' failed but was never awaited, so the workflow result does not reflect it: boom from unawaited task ---`
  - failing task awaited inside `try`/`catch` — `success | "caught: …"`, no warning (the
    false-positive case the design turns on)
  - unawaited failure and then the body throws for an unrelated reason — run fails with the
    unrelated error *and* the unawaited failure is reported, with no duplicate in the child's log

## Out of scope, found while testing

An unawaited task dispatched *alongside* another task hangs its own child job forever, on the
published client and unrelated to this change: the child round's unawaited call strands its
`step_complete` in an unhandled rejection, and the body then parks on the never-resolving thenable
that non-matching steps are given by design. The single-task shape in this PR is unaffected.
Worth its own issue.


rubenfiszel:
Nits addressed in 37f682b, except one I'd like to dismiss.

**Placement invariant documented away from its enforcement point** (both reviewers) — agreed and
fixed. The rationale now sits in the block comment above the wrapper in `bun_executor.rs`, naming
all four suspend modes plus the trailing-dispatch return, so someone adding a fifth mode meets it
where they'd break it. `client.ts` keeps only what the method itself does.

**Runner-level regression test** (Codex P2) — dismissing, and the two reviews disagree here:
Claude reached the opposite conclusion ("the wrapper is a generated string with no cheap unit
seam"). The only TS-side test possible is one that mirrors the wrapper's `run()` in the test file
and asserts against the mirror. That guards a copy, not the shipped artifact: the copy can stay
green while the real `format!` string drifts, which is a worse failure mode than no test, and
`AGENTS.md` calls that shape scaffolding. I wrote exactly that mirror while developing and deleted
it. There is no Rust seam that compiles the wrapper either.

Instead I closed the gap Claude actually identified — the retry-backoff round resting on
inspection — by exercising it on a real bun worker. A fire-and-forget task with
`{ retry: { attempts: 1, delay: 2 } }` failing every attempt:

```
--- WAC: sleep(notify#retry2, 2s) before retrying notify ---
--- WAC: task 'notify' failed but was never awaited, so the workflow result does not reflect it: boom from unawaited task ---
--- WAC: complete ---
```

`success | "done"`, exactly one warning, named `notify` rather than `notify#2`, nothing from the
backoff round, and both failed child jobs' logs clean. That is the `sleep` early return holding at
runtime rather than by reading.

Also, on Claude's note that the reported test run predated the merge: `bun test
typescript-client/tests/` is green at 181 across 6 files on the merged tree, and I checked the new
retry test is load-bearing by hoisting the registration above the retry check — it fails with a
spurious `task 'retried' failed` line, then passes again on revert.

rubenfiszel:
✅ Review round clean @ 8baef72c8787ed00e364cb2d0fe71b80f888ee29 — nit-only verdicts at 8baef72c8787ed00e364cb2d0fe71b80f888ee29; nits addressed in 37f682b6c538cd39c328f5e85540a04a4fb602b4 / dismissed in review replies

PR #11019, opened 2026-09-08
feat: let a worker group override the dependency cache object store

rubenfiszel:
## Summary

A worker group can now point its **dependency cache** at its own object store instead of the
instance one. This is for workers that sit far from the instance bucket, or cannot reach it at
all — a regional pool that would otherwise pull every venv, bundle and compiled binary across a
WAN on each cold start.

The override is deliberately narrow. What may move to a group's own bucket is what is both
**reconstructible** and **group-local**: pip/uv venvs, language bundles, and compiled
Go/Rust/C#/Java binaries. Every artifact there can be rebuilt from the lockfile, so a group that
cannot find one just rebuilds it — a miss costs time, never correctness.

Everything else keeps using the instance object store, for two distinct reasons:

- **The server also reaches it** — job results, service and job logs, codebases uploaded at deploy
  time, app assets (`/app_bundles/…`, written and served by the API). A worker-local redirect
  would point the worker at a bucket the server never wrote to.
- **It outlives any one group** — AI agent conversation memory (`memory/{ws}/{conversation}/…`) is
  worker-only on both sides, so it *could* have moved, but must not: it is keyed by conversation,
  not by worker, and agent steps are dispatched by tag. Step 1 on group A and step 2 on group B
  would make step 2's read a NotFound, which is indistinguishable from "no history" — the agent
  would silently forget. It is also the only copy, not a cache: the DB write is a fallback for a
  failed put, not a second copy to fall back to on read.

Enterprise-only, and applied live: changing it hot-swaps the store on every worker in the group
with no restart.

## Changes

**New store, resolved per call site**
- `windmill-object-store`: a second store (`CACHE_OBJECT_STORE_OVERRIDE`) alongside
  `OBJECT_STORE_SETTINGS`, plus a `CACHE_OBJECT_STORE_OVERRIDDEN` flag held apart from it.
  `get_cache_object_store()` returns the override when one is configured and the instance store
  otherwise; `get_object_store()` is unchanged for everyone else. The AWS-OIDC refresh path that
  `get_object_store()` had is factored into `resolve_object_store()` and shared by both.
- A configured-but-unbuildable override resolves to **no store**, not to the instance store — the
  flag is what makes that distinction possible. Falling back would write a group's cache into the
  bucket the operator explicitly redirected it away from. The dependency cache degrades to
  local-disk-only until the store builds.
- `reload_cache_object_store_override()` builds and clears it; refuses on the Pro plan (clearing
  any override loaded before a downgrade) and under `DISABLE_S3_STORE`, mirroring
  `reload_object_store_setting`. It returns `ObjectStoreReload::Later` for a build failure that
  may pass, and the caller retries once after 10s — the same shape `initial_load` already uses for
  the instance store. A malformed config returns `Never`, since it reads the same on every retry.
  Beyond that retry the backstop is the periodic settings reload, which is 12h apart by default.
- `resolve_object_store` no longer writes a finished OIDC refresh unconditionally: if a reload
  installed a different store while the credentials were being minted, the refresh is stale and
  the newer store wins. This also fixes the same race on the pre-existing instance-store path.

**Config plumbing**
- `object_store_cache_config` on `WorkerConfigOpt` / `WorkerConfig` (raw JSON: `windmill-common`
  cannot depend on the crate that parses it, and comparing the raw value is what tells a reload the
  store changed). `WorkerConfig`'s hand-written `Debug` redacts it — it holds bucket credentials.
- `reload_worker_config` hot-swaps the store when the value changes, outside the
  config-unchanged early return so a failed build is retried. No killpill: nothing about the
  override needs a worker restart.
- Typed field on `WorkerGroupConfig` for config-as-code.

**Call sites switched** to `get_cache_object_store()`: `global_cache.rs` (`load_cache`,
`save_cache`, `exists_in_cache`, `exists_in_object_store`, `object_store_available`),
`python_executor.rs` (piptar push/pull), `universal_pkg_installer.rs`.
Deliberately **not** switched: `bun_executor.rs:1173`, which downloads a codebase the *server*
uploaded at deploy time.

**API**
- The obfuscation `list_worker_groups` did inline is now a shared `obfuscate_worker_config`, which
  also masks the new bucket credentials (`access_key` / `secret_key` / `accessKey` /
  `serviceAccountKey`) — and `get_config` now goes through it too. It previously returned the raw
  config to any devops user, so it would have handed out the whole bucket; it also handed out
  `env_vars_static` in the clear, which this closes as well.
- `update_config` restores masked credentials from the stored row before writing. A devops user
  who is not an instance admin edits from the obfuscated view, so a plain save would otherwise
  store `*****` as the secret and take the group's cache offline silently. (This is why the new
  field uses a fixed sentinel: `env_vars_static` masks to a variable-length run of `*`, which
  cannot be told from a real value, so **that pre-existing clobber is not fixed here** — worth a
  separate change.)
- No new endpoint or schema: the value rides in the existing worker-group config blob. CE is
  already gated — `update_config` strips non-whitelisted keys on non-enterprise builds.

**Frontend**
- A "Dependency cache object storage" section in the worker-group drawer, reusing
  `ObjectStoreConfigSettings`.
- That component gained `showInstanceStorageTools` (its storage-usage and log-cleanup panels act
  on the *instance* store, so they must not appear here) and `disabled`. Its config types moved to
  a module script so callers can name them.

## Screenshots

Worker group config drawer → new section, off and on:

pr-cache-override-off

pr-cache-override-on

## Test plan

Exercised on a running worker (`--features quickjs,enterprise,private,parquet`) with two
filesystem object stores — one as the instance store, one as the group override:

- [x] **Override set** — a bun bundle lands in the group bucket; the instance bucket receives only
      service logs.
- [x] **Override set, local cache cleared** — the job logs `loaded bin/bundle from object store`
      and returns its result, i.e. the read path uses the override too.
- [x] **Override unbuildable** (`root_path` under `/proc`) — the job logs
      `backed by EE distributed object store: false`; the cache stays local and **neither** bucket
      is touched.
- [x] **Override removed** — falls back to the instance bucket live, with no worker restart
      (`Worker group object store cache override removed…` in the log).
- [x] UI driven with Playwright: section renders, "Overridden" badge appears, the config
      round-trips through Apply into the worker-group config row.
- [x] As a **devops non-superadmin**: both `list_worker_groups` and `get_config` return masked
      credentials, and saving that masked view back (changing only the bucket) keeps the real
      `access_key`/`secret_key` in the stored row.
- [x] `cargo check` at `quickjs`, `quickjs,parquet`, `quickjs,enterprise,private,parquet`, and
      `windmill-common --features instance_config_schema`.
- [x] `cargo test -p windmill-object-store --features parquet` — 42 pass, including a new test
      pinning that a configured-but-unbuilt override yields no store rather than the instance one.

## Notes for reviewers

- Deploy-time prebuilds run on whichever worker holds the `dependency` tag. If that worker is in a
  group without the override while the running workers are in one with it, the prebuilt binary is
  not found — the group rebuilds it once and pushes to its own bucket, so it self-heals, just not
  on the first run.
- Agent workers never read their group's config (only their token), so the override does not reach
  them.
- The frontend has no Pro-vs-Enterprise signal at all (`enterpriseLicense` is the only store), so
  a Pro instance can fill this section in and the worker will refuse it — exactly as it can today
  for the instance object storage setting. Left alone rather than invented here.
- No `feature_usage` telemetry yet; say the word if a `worker_group` / `cache_store_override`
  counter is wanted.

Reviewed locally before opening with both `/local-review` (Claude, cold context) and
`/local-review-codex` (`gpt-5.6-sol`, xhigh), then through four CI review rounds. 16 findings in
total, all addressed; the last round is clean from both reviewers. The two exceptions are noted
above as pre-existing and deliberately out of scope (the `env_vars_static` mask clobber, and the
absence of any Pro-vs-Enterprise signal in the frontend).

Unrelated gap found while tracing the memory path, worth its own issue: the object-store-aware
`memory_ee::delete_conversation_memory` has no caller — the API's conversation delete only removes
`ai_agent_memory` rows — so on an instance with object storage configured, deleting a conversation
leaves its `memory/…` objects in the bucket indefinitely.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -15845,6 +15845,7 @@ dependencies = [
  "reqwest 0.13.4",
  "serde",
  "serde_json",
+ "serial_test",
  "sqlx",
  "tempfile",
  "tokio",
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1944,6 +1944,13 @@ async fn process_notify_event(
                     if let Err(e) = reload_license_key(&db.into()).await {
                         tracing::error!("Failed to reload license key: {e:#}");
                     }
+                    // The worker-group cache override is Enterprise-only, and nothing else
+                    // re-reads the plan for it: the periodic settings pass runs ahead of
+                    // reload_license_key, so it would see the plan this event just replaced.
+                    #[cfg(feature = "parquet")]
+                    if worker_mode {
+                        crate::monitor::reload_cache_object_store_override_with_retry(db).await;
+                    }
                 }
                 DEFAULT_TAGS_PER_WORKSPACE_SETTING => {
                     if let Err(e) = load_tag_per_workspace_enabled(db).await {
--- a/backend/src/monitor.rs
+++ b/backend/src/monitor.rs
@@ -5129,37 +5129,6 @@ pub async fn reload_worker_config(db: &DB, tx: KillpillSender, kill_if_change: b
                 .as_ref()
                 .is_some_and(|dws| !dws.is_empty());
 
-        // Outside the `!=` block below so a failed build is retried by a later settings pass,
-        // which sees an unchanged config.
-        #[cfg(feature = "parquet")]
-        if wc.object_store_cache_config != config.object_store_cache_config
-            || windmill_object_store::cache_object_store_override_failed().await
-        {
-            let settings = config.object_store_cache_config.clone();
-            if matches!(
-                windmill_object_store::reload_cache_object_store_override(db, settings.clone())
-                    .await,
-                ObjectStoreReload::Later
-            ) {
-                // The whole group's dependency cache is local-only until this builds, and the
-                // settings pass that would otherwise retry is 12h away, so try again soon —
-                // an AWS OIDC store cannot mint its first token until the server is serving.
-                let db = db.clone();
-                tokio::spawn(async move {
-                    tokio::time::sleep(Duration::from_secs(10)).await;
-                    // The group config may have been edited during the wait; installing the
-                    // settings this retry captured would then pin the worker to a store the
-                    // group no longer asks for, and clear the flag the newer one needs.
-                    if WORKER_CONFIG.load().object_store_cache_config == settings
-                        && windmill_object_store::cache_object_store_override_failed().await
-                    {
-                        windmill_object_store::reload_cache_object_store_override(&db, settings)
-                            .await;
-                    }
-                });
-            }
-        }
-
         if **wc != config || has_dedicated {
             if kill_if_change {
                 if has_dedicated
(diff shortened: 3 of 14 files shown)


rubenfiszel:
✅ Review round clean @ 48a04290e1c6d2f90c72d8b0adf462d583b91585

PR #11020, opened 2026-09-08
feat: put a data table's connection under Postgres roles

diegoimbert:
First of four PRs. This one is the security boundary, so it stands on its own.

## Summary

A data table backed by the instance database resolved to exactly one Postgres connection —
`custom_instance_user` — for everyone who could reach it at all. There was no way to say *this job
reads, that one writes, this one never sees the salaries table*.

A **data table role** is now a real Postgres login on the cluster, defined once for the instance by
a superadmin and named exactly as they named it (one `analytics`, shared by every instance
database). A script that declares `-- role analytics` connects **as** `analytics`, and Postgres
decides what it may touch — grants are ordinary SQL. Windmill answers only *may this caller ask for
this role*, from the tenant lists on the data table entry: `u/alice`, `g/analysts`, `f/finance`, `*`.

A data table with no `permissions` block behaves exactly as it does today.

## Changes

**Instance role catalog** — a `datatable_role` table, one row per role. Deliberately not a
`global_settings` key: that row has generic read, list, write and CLI round-trip paths that know
nothing about what they carry, and every one of them is a way to leak the passwords or to
overwrite the catalog with a copy that has none. A row nothing generic touches has none of those,
and two concurrent creates are two inserts rather than a read-modify-write over one document.
Superadmin CRUD in `windmill-api-settings`. Create is `CREATE ROLE <name> LOGIN
PASSWORD …; GRANT <name> TO custom_instance_user` and nothing more; disable is `NOLOGIN`; rename
carries a password reset (`RENAME` discards an md5 hash) and moves nothing else, because tenants key
by a generated id. Delete disables the role in its own commit, then reassigns and drops what it
owns in every registered database, and only then forgets it. Each database's pass commits on its
own, so a failure part-way leaves a disabled, retryable catalog entry, never an enabled role already
stripped in some databases nor a live login nothing names.

**One chokepoint** — `get_datatable_resource_from_db(db, w_id, name, role, DatatableAccess)`
replaces the `_unchecked` resolver at every site that acts for a caller. `DatatableAccess` carries
the identity explicitly and `NoIdentity` fails closed. Role precedence: `-- role` annotation >
`?role=` on the reference > the entry's default > `admin`. The role **logs in as itself** — never
`SET ROLE`, which a script could `RESET ROLE` its way out of. A reference's `?role=` is split off
only after looking for an entry stored under the whole string: names could contain `?` before they
were restricted, and such an entry keeps resolving as itself (Postgres executor, DuckDB `ATTACH`,
the clone checks). An agent worker cannot read the workspace, so there the strict parse applies and
such a name is refused rather than misread.

**Forks point instead of copying** — `clone_workspace_data` copies `workspace_settings` wholesale,
so a keep-original fork used to get a byte-identical entry naming the parent's database, which a
fork admin (not a superadmin, unable to repoint it) could edit to grant themselves `admin` there. A
fork's instance-backed entry is now `{reference: {workspace_id, datatable}}`: nothing local to edit.
`reference`, `permissions` and `forked_from` are all carried across a settings save from the stored
entry rather than taken from the request, so no payload can forge or drop one. Tenants are evaluated
**as a member of the governing workspace**, by email; a `g/` or `f/` from a foreign workspace means
nothing there and is refused. A clone fork keeps its own copy of `permissions` and governs it. A settings save that drops a governing
entry returns the fork pointers it leaves resolving to nothing, worked out from the stored entries
rather than a client-sent list, so `wmill sync push` warns about them as the settings page does. A database under roles is
reached only through an entry carrying them, and roles follow an entry only through a declared
rename, which a settings sync never sends. So a save that newly points an entry without roles at a
database under roles, in this workspace or another, is refused rather than leaving that database
open to everyone as `admin`: a rename to a new name, an existing name repointed, or a second entry.

**Cloning a data table under roles is refused** — `pg_dump` carries no roles and the import runs
`--no-privileges`, so a copy would arrive with the parent's tenants and none of the grants behind
them: every role but `admin` denied by Postgres, in an entry that reads as configured. Refused at
both endpoints a clone uses, before any database is created. The ACL replay that lifts this is the
follow-up PR's subject; dropping `permissions` from the copy instead would hand the parent's rows
to everyone in the fork.

**Owner-gated** — the operations that see the whole database whatever the roles grant: editing the
roles, a migration that declares none (it now runs as `admin`, matching what it is gated as),
schema export, the
full-schema read, database create/import/drop, the connection test, and the schema snapshot a clone
takes of its parent.

**Postgres triggers and roles don't mix** — a replication stream reads every row of every table
whatever the roles grant, and its listener checks access only when it connects. So a data table
carries one or the other. A Postgres trigger cannot be created on, or edited to point at, a data
table under roles; one that reaches such a data table anyway (re-enabled, or cloned into a fork) is
disabled by its listener with the reason, as a missing replication slot is, and a capture is refused
the same way. The other way round, roles cannot be turned on while an enabled trigger or a live
capture reads the data table — its own, or a fork's through its pointer — and the refusal names each
one to disable. The two sides serialize on one advisory lock: roles going on hold it exclusive
while they look for streams, and every path that can start one (trigger create, edit and enable,
capture setup and ping) holds it shared while it commits. So a trigger enabled at the same moment
is either seen by the look or refused by its listener, which connects after roles are committed.
A listener also outlives its trigger being disabled, or its capture's client going quiet, until
its next heartbeat notices, so one that pinged in the last 15 seconds — the window a server holds
a listener for — still counts as streaming, and the refusal says to allow it that long to stop.

**`permissions` does not travel** — stripped from `get_settings`, `get_public_settings` and the
workspace export. Tenants name principals of one workspace (`g/analysts` in dev is a different
group from `g/analysts` in prod), and a settings push is not where an access decision should be
made. Principal deletes and renames cascade into every tenant list in the same transaction.

**Callers** — both executors, the agent-worker endpoint (EE; it carries the job id, since it
authenticates as infrastructure), the schema/table API (now resolving as the caller's role, with
`has_schema_privilege` filtering so a restricted role is not shown schemas it cannot enter), the
migrations gate, the asset parsers (`?role=` is not part of the asset path), and both SDKs plus the
regenerated prompts. `system_prompts/generate.py` also learned to render the bare `*` before
keyword-only args — without it `datatable(name, *, role)` read as all-positional.

**Telemetry** — `("datatable", "role_created")`, `("datatable", "roles_toggled")` (`on`/`off`) and
`("datatable", "role_connection")` (`named`/`default`, only on a data table under roles), with the
instance disclosure copy updated. Verified landing in `feature_usage` on an EE build.

## The UI ships but is not mounted

`DataTableRolesSection.svelte` (the instance role catalog) and
`DataTablePermissionsButton.svelte` (the per-table roles drawer) are complete and reviewed, but
their call sites in `DataTableSettings.svelte` are commented out. Creating roles is only useful
once there is a way to grant them privileges, which arrives with the ACL editor; left mounted they
would offer half a feature. Each call site carries a note pointing the follow-up PR at them.

The API surface is unaffected — `/datatable_permissions/{name}` and `/settings/datatable_roles`
are routed and enforced either way. This removes the way in through the UI, not the feature.

Screenshots below are from before they were unmounted, and are what the follow-up restores.

Data table settings, with the superadmin-only instance role catalog:

dt-settings

The roles drawer on the owning workspace:

dt-roles-drawer

The same drawer from a fork, naming the workspace that governs it:

dt-roles-fork

## Test plan

Exercised on a running instance (backend + worker, `postgresql` tag served), not just type-checked:

- [x] `CREATE ROLE` lands in `pg_roles` with `GRANT analytics TO custom_instance_user`; reserved and
      malformed names refused
- [x] With `SELECT` on `orders` and nothing on `salaries`: a job under the default role reads
      `orders` and gets `permission denied for table salaries` **from Postgres**
- [x] All four selection paths: default role, `-- role admin`, `?role=analytics`, annotation beating
      the URI role; an unknown role is an error, not a fallback
- [x] A caller no tenant covers is refused by Windmill before a connection is opened, for both the
      default role and `admin`
- [x] Same through DuckDB `ATTACH 'datatable://main?role=analytics'` (backend rebuilt with `duckdb`)
- [x] Migrations: a no-role migration is refused for a non-admin and runs as `admin` for an admin; a
      `-- role`-declaring one passes the gate and is then refused by Postgres
- [x] Turning roles on is refused while an enabled trigger reads the data table through a fork's
      pointer, and the refusal names it; disabling the trigger lets roles on
- [x] On a data table under roles (backend rebuilt with `postgres_trigger`): creating a Postgres
      trigger and its connection test are refused, and an enabled trigger planted there is
      switched to `disabled` by its listener with the reason instead of being retried
- [x] Roles going on wait for a trigger enable in flight, then refuse naming it; with the lock
      stubbed out the save goes through before the trigger commits
- [x] A disabled trigger whose listener pinged in the last 15 seconds still keeps roles off; once
      its ping is older they go on. Without the heartbeat clause roles go on at once
- [x] A stored name containing `?` resolves as itself, `main?role=analytics` still splits, and an
      unknown parameter is still refused. Without the exact-name lookup the stored name is misread
      as a role query; DuckDB `ATTACH` unit tests cover the same through its pattern
- [x] A settings save dropping a governing data table names the forks it strands without being
      sent `deleted_datatables`; without the server-side derivation the list is empty. The CLI
      from source (`wmill sync push --include-settings`) printed the warning naming the fork
- [x] Sync-style saves that newly point an entry without roles at a database under roles are
      refused, with the roles left in place: a rename to a new name, an existing name repointed,
      and another workspace's entry. Without the check the first goes through; declared renames
      still carry roles
- [x] A role delete failing on an unreachable database leaves the role disabled in the catalog
      and in `pg_roles`; without the disable step it stays enabled
- [x] A keep-original fork gets a pointer; a fork admin can use the data table, cannot edit its
      roles, and a hostile settings save naming the parent's database leaves the pointer intact
- [x] Export / full-schema / connection-test refused for a non-admin member on a permissioned data
      table
- [x] Deleting a user, group and folder strips them from every tenant list; deleting a role drops
      the Postgres login and strips it from every workspace, falling the default back to `admin`
- [x] Rename keeps tenants (they key by id) and the connection still authenticates; disable makes
      the role unusable
- [x] `permissions` absent from `get_settings`, `get_public_settings`; the enable-time sweep names a
      second workspace planted on the same database
- [x] UI driven in a real browser (screenshots above), `npm run check` clean
- [x] `cargo check` for `quickjs`, `enterprise,private`, and `-p windmill-worker --features duckdb`;
      `SQLX_OFFLINE=true cargo check --workspace --features all_sqlx_features`
- [x] Unit tests (annotation, ATTACH, asset path, tenant matching, entry shape, role-name
      validation) and eighteen integration tests in `datatable_roles.rs`
- [x] A role named but not spelled exactly — `-- Role x`, `-- role: x`, `-- role x;`,
      `-- role x -- why`, `?Role=x`, `?role=`, and the `row_to_json` retry — resolves to the role
      or is refused, never to the data table's default
- [x] Copying a data table under roles is refused at both `create_pg_database` and
      `import_pg_database`, leaving no `wm_fork_*` database and no registry entry

Not exercised here, stated rather than implied:

- The clone flow end to end. The dev shell's Postgres client trailed the server (17.7 against
  18.6), which this PR fixes by pinning `postgresql_18` in `flake.nix` — a real `pg_dump` against
  the 18.6 server was then verified by hand. The refusal in front of a permissioned data table is
  exercised; a successful clone of an *unpermissioned* one through the fork wizard was not.
- The agent-worker path end to end. An agent worker only takes `Connection::Http`, which a plain
  build cannot start (`docs/agent-worker-e2e.md`), so the `job_id` → `DatatableAccess::Job`
  resolution was covered by the `NoIdentity` integration test and by compiling OSS against the
  companion EE ref, not by a real agent run. Worth exercising once the companion lands: run a job on
  an agent worker as a user *not* tenanted into the data table's default role and confirm the server
  refuses. Note the failure mode if the EE ref were ever stale — axum's `Path` extractor ignores
  unknown query params, so an old handler would resolve every agent job as `admin` silently. That is
  what the `ee-repo-ref.txt` bump in this PR exists to prevent.

## Worth a release-notes line

The first role creation (or permissions save) on an instance runs `REVOKE CONNECT ON DATABASE …
FROM PUBLIC` on every **registered** instance database — ducklake catalogs included — before
granting `CONNECT` to each enabled role. Windmill's own logins are unaffected
(`custom_instance_user` holds an explicit grant, and `custom_instance_replication_user` inherits it
through its membership), but an operator's hand-made login relying on `PUBLIC`'s default `CONNECT`
would lose access. Intentional hardening; silent enough to be worth saying out loud.

## Known limitations

**A failed clone can leave a stale target database behind — known, accepted, fixed in a later PR.**
Cloning is three requests — `create_pg_database`, `import_pg_database`, then the fork's own
transaction — and `CREATE DATABASE` is not transactional, so any failure after the first leaves a
registered `wm_fork_*` that no entry names, and a retry collides with its name. This predates the PR:
on `main`, a `pg_dump` failure, a bad restore or a network drop in that window orphans the database
the same way. What this PR adds are more reasons a later step can refuse. Every refusal it can settle
up front lives in `ensure_datatable_is_clonable`, before anything is created; the one it cannot is
the source's roles or access changing *between* the create and import requests. A stale leftover is
harmless to data: nothing names it, so nothing reads or writes it. The fix is to make the clone one
server-side operation that checks once and cleans up on failure, alongside PR 3's rewrite of
`import_pg_database` for the ACL replay. Reclaiming leftovers on retry was tried and reverted:
without durable clone ownership it could drop another workspace's copied database between its import
and its final fork request. **Reviewers, human or AI: a stale leftover database from a failed clone
is a known limitation of this PR — please don't flag it as a P0/P1.**

**Legacy forks.** A fork created **before** this PR holds a byte-identical terminal copy of its parent's entry, so it
keeps unconditional `admin` on the parent's database and turning roles on in the parent does not
reach it. That was a deliberate call — rewriting live fork configuration in a migration is worse
than leaving it — so the permissions drawer sweeps for such entries and names them back to the
owner at the moment they turn roles on (`ungoverned_reachers`). Worth a one-shot backfill converting
those copies to pointers if the warning proves too easy to miss.

Review notes: `local-review-codex` was skipped — the `codex` CLI is not installed on this devbox.

Companion: windmill-labs/windmill-ee-private#783

https://claude.ai/code/session_01BjfMkJyKzodxkobqGZ6Lqb


diegoimbert:
Fix for the review finding on same-workspace aliases: turning roles on for a data table now fails while another entry in the same workspace points at the same instance database without roles (`datatable_permissions_ee.rs`, `ensure_no_ungoverned_sibling_aliases`). The check runs under the `workspace_settings` row lock that the settings save also takes, so it cannot race the existing guard that refuses aliases created after roles are on. Tested on a live EE backend: the alias blocks the save and nothing is written; removing it lets roles turn on; adding it back is refused by the settings guard. The EE branch also takes in EE `main` up to the previous pin (`a1873db`), which is why `ee-repo-ref` moves to `5361477`.

diegoimbert:
The "? names" P1 can be ignored, nobody would do that

PR #11026, opened 2026-09-08
feat(nativets): bound fetch on a peer that never answers

rubenfiszel:
## Summary

Bound Native TypeScript `fetch()` calls that never receive response headers. A silent peer currently holds a healthy worker until the job timeout, which can default to seven days on self-hosted instances. The new response timeout defaults to 900 seconds and stops once headers arrive, so streaming bodies and long downloads remain unrestricted.

## Changes

- Configure the default with `WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS`; override it per script with `//fetch_response_timeout <seconds>`. Zero disables the timeout. Both normal and native dedicated workers use the setting.
- Preserve caller cancellation, WebIDL `RequestInit` conversion, request bodies, fetch arity, required-argument errors, and already-aborted promise settlement. Capture promise and abort operations before user code can replace them.
- Clamp timer milliseconds to `i32::MAX` to avoid Deno timer overflow. Timeout messages omit query strings and fragments and explain the configuration options.
- Keep the 900-second default above the synchronous execution wait limit (600 seconds by default). Raising the hot-reloaded wait limit may also require raising the fetch environment setting and restarting workers.

The deadline includes connection establishment, request upload, and waiting for response headers. Long polls and slow uploads exceeding it need a larger override or zero. Bodies that stall after headers remain outside this timeout.

## Test plan

- [x] Runtime regression suite: 14 fetch cases, 2 annotation cases, and the isolated environment configuration test.
- [x] Build the backend with `deno_core` and with `deno_core,enterprise,private` at the pinned EE commit.
- [x] Real normal-worker jobs: silent response fails around 2 seconds; the same worker runs a follow-up job; a 4-second streaming body completes; zero allows an 8-second wait.
- [x] Real native dedicated-worker jobs: per-script and environment defaults fail silent responses around 2 seconds; streaming completes; zero overrides a 2-second environment default and allows an 8-second wait; the same worker runs follow-up jobs.
- [x] Real job with public abort helpers replaced: timeout still fires and the same worker accepts the next job.
- [x] Full backend check including tests with `deno_core,enterprise,private`.
- [x] Real synchronous child-job call: a 400-second child completed successfully; the outer job waited 456 seconds including queue time under the default 900-second fetch deadline.
- [x] Final CI checks on `d2c53ce728`: OSS/EE builds, Rust client, integration suite, Pages, CLA, and automated review checks passed. Codex verdict: **Good to merge**. GitHub still requires reviewer approval.

Fixes WIN-2485

### DIFF
--- a/backend/tests/nativets_dedicated.rs
+++ b/backend/tests/nativets_dedicated.rs
@@ -12,7 +12,7 @@ mod prewarmed_isolate_tests {
     use windmill_worker::{build_loader, LoaderMode, BUN_PATH};
 
     fn default_annotation() -> NativeAnnotation {
-        NativeAnnotation { useragent: None, proxy: None }
+        NativeAnnotation::default()
     }
 
     /// Bundle a TypeScript script into JS suitable for `PrewarmedIsolate`.
--- a/backend/windmill-runtime-nativets/src/lib.rs
+++ b/backend/windmill-runtime-nativets/src/lib.rs
@@ -197,10 +197,10 @@ pub struct NativeAnnotation {
 /// Covers everything up to the response headers and stops there, so a body may
 /// then stream for any length of time. `src/runtime.js` holds the semantics.
 ///
-/// Must stay above `TIMEOUT_WAIT_RESULT` (default 600): `run_wait_result`
-/// long-polls with no headers until that ceiling, so a script running another
-/// job synchronously would otherwise time out client-side while the server is
-/// still legitimately holding the request open. Raising one means raising both.
+/// Must exceed `TIMEOUT_WAIT_RESULT` (default 600), which holds synchronous job
+/// calls open without headers. Raising that hot-reloaded instance setting may
+/// also require raising `WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS` in the deployment
+/// and restarting workers: this environment value is cached for the process.
 pub fn default_fetch_response_timeout_secs() -> u64 {
     static SECS: LazyLock<u64> = LazyLock::new(|| {
         std::env::var("WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS")
--- a/backend/windmill-runtime-nativets/src/lib.rs
+++ b/backend/windmill-runtime-nativets/src/lib.rs
@@ -26,7 +26,7 @@ use std::{
     cell::RefCell,
     path::PathBuf,
     rc::Rc,
-    sync::{Arc, Mutex},
+    sync::{Arc, LazyLock, Mutex},
 };
 
 // Re-export deno_telemetry for use by windmill-worker's otel proxy
@@ -182,10 +182,28 @@ struct LogString {
     pub s: mpsc::UnboundedSender<String>,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct NativeAnnotation {
     pub useragent: Option<String>,
     pub proxy: Option<(String, Option<(String, String)>)>,
+    /// `//fetch_response_timeout <seconds>`: per-script override of
+    /// [`default_fetch_response_timeout_secs`]. `Some(0)` disables it for this
+    /// script; `None` leaves the default in force.
+    pub fetch_response_timeout_secs: Option<u64>,
+}
+
+/// How long `fetch()` waits for a response to begin, in seconds; `0` disables.
+///
+/// Covers everything up to the response headers and stops there, so a body may
+/// then stream for any length of time. `src/runtime.js` holds the semantics.
+pub fn default_fetch_response_timeout_secs() -> u64 {
+    static SECS: LazyLock<u64> = LazyLock::new(|| {
+        std::env::var("WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS")
+            .ok()
+            .and_then(|x| x.trim().parse::<u64>().ok())
+            .unwrap_or(300)
+    });
+    *SECS
 }
 
 /// Serializes V8 isolate creation as defense-in-depth against concurrent
@@ -401,7 +419,7 @@ pub fn transpile_ts(expr: String) -> anyhow::Result<String> {
 }
 
 pub fn get_annotation(inner_content: &str) -> NativeAnnotation {
-    let mut res = NativeAnnotation { useragent: None, proxy: None };
+    let mut res = NativeAnnotation::default();
 
     let anns = inner_content
         .lines()
@@ -414,6 +432,13 @@ pub fn get_annotation(inner_content: &str) -> NativeAnnotation {
             res.useragent = Some(ann.trim_start_matches("useragent").trim().to_string());
         } else if ann.starts_with("proxy") {
             res.proxy = capture_proxy(ann.trim_start_matches("proxy").trim());
+        } else if ann.starts_with("fetch_response_timeout") {
+            // A typo falls back to the default, never to "no timeout".
+            res.fetch_response_timeout_secs = ann
+                .trim_start_matches("fetch_response_timeout")
+                .trim()
+                .parse::<u64>()
+                .ok();
         }
     }
     res
@@ -554,6 +579,16 @@ pub(crate) fn create_nativets_runtime(
     let ops = vec![op_get_static_args(), op_log()];
     let ext = Extension { name: "windmill", ops: ops.into(), ..Default::default() };
 
+    // deno_web's setTimeout puts its delay through `webidl.converters.long`,
+    // which wraps at 32 bits: past i32::MAX ms (~24.8 days) the delay comes out
+    // negative and fires immediately, so an over-generous setting would abort
+    // every fetch on the spot. Cap rather than wrap.
+    let fetch_response_timeout_ms = ann
+        .fetch_response_timeout_secs
+        .unwrap_or_else(default_fetch_response_timeout_secs)
+        .saturating_mul(1000)
+        .min(i32::MAX as u64);
+
     let fetch_options = deno_fetch::Options {
         root_cert_store_provider: NATIVE_ROOT_CERT_STORE_PROVIDER.clone(),
         user_agent: ann.useragent.unwrap_or_else(|| "windmill/beta".to_string()),
@@ -625,10 +660,15 @@ pub(crate) fn create_nativets_runtime(
     }
 
     // Per-isolate JS init that can't run in the snapshot (runtime.js executes at
-    // snapshot-build time): currently seeds performance.timeOrigin via
-    // setTimeOrigin(), which must read this isolate's wall clock.
+    // snapshot-build time): the wall clock behind performance.timeOrigin and the
+    // fetch response timeout are both per-isolate values.
     js_runtime
-        .execute_script("<wm_init>", "globalThis.__wmInitPerIsolate()")
+        .execute_script(
+            "<wm_init>",
+            format!(
+                "globalThis.__wmInitPerIsolate({{ fetchResponseTimeoutMs: {fetch_response_timeout_ms} }})"
+            ),
+        )
         .map_err(windmill_common::error::to_anyhow)?;
 
     Ok(CreatedRuntime { js_runtime, log_receiver, memory_limit_rx })
--- a/backend/windmill-runtime-nativets/src/lib.rs
+++ b/backend/windmill-runtime-nativets/src/lib.rs
@@ -196,12 +196,17 @@ pub struct NativeAnnotation {
 ///
 /// Covers everything up to the response headers and stops there, so a body may
 /// then stream for any length of time. `src/runtime.js` holds the semantics.
+///
+/// Must stay above `TIMEOUT_WAIT_RESULT` (default 600): `run_wait_result`
+/// long-polls with no headers until that ceiling, so a script running another
+/// job synchronously would otherwise time out client-side while the server is
+/// still legitimately holding the request open. Raising one means raising both.
 pub fn default_fetch_response_timeout_secs() -> u64 {
     static SECS: LazyLock<u64> = LazyLock::new(|| {
         std::env::var("WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS")
             .ok()
             .and_then(|x| x.trim().parse::<u64>().ok())
-            .unwrap_or(300)
+            .unwrap_or(900)
     });
     *SECS
 }
--- a/backend/windmill-runtime-nativets/src/runtime.js
+++ b/backend/windmill-runtime-nativets/src/runtime.js
@@ -40,6 +40,8 @@ const ORIGINAL_FETCH = fetch.fetch;
 // invocation".
 const setTimeoutUnbound = timers.setTimeout;
 const clearTimeoutUnbound = timers.clearTimeout;
+// Captured before user code shares the isolate and could redefine it.
+const PromiseReject = Promise.reject.bind(Promise);
 
 // Installed per isolate by __wmInitPerIsolate; 0 disables. Only a backstop
 // for the impossible case of fetch running before that init.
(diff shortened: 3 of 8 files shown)


rubenfiszel:
All five addressed in `5e02236`. Thanks — the P1 and the P2 on the error message were both real.

**P1 — customer attribution and identifiers.** Correct and my mistake; this repo is public and I put real identifiers in it. The test file is rewritten with no narrative, no `BC`, no timestamps, and neutral values (`/orders`, `{ id: 1 }`). The commit was amended rather than followed up, so the identifiers are gone from the branch history rather than just from its tip, and the PR description is rewritten the same way. It reads as a technical motivation now — 7-day fallthrough, zombie detection keying off a ping that stays fresh.

**P2 — the message asserted something the timer can't know.** You're right, and it mattered more than the wording: the window really does cover DNS, connect, TLS and request upload, so "the connection was established but the server sent no response" was false for a connect-phase abort, and a slow large upload was a rollout-risk class I'd missed. The message is now "no response headers arrived within Ns (this covers connect, request upload and the wait for the server to start replying; once a response begins it is never interrupted)", and slow uploads are listed alongside long-polling in the PR's rollout risk. I couldn't make upload progress *reset* the timer — deno_fetch exposes no upload-progress hook to the JS layer — so the honest fix was to stop over-claiming and name the escape hatch in the message.

**P2 — annotation naming.** Agreed on the problem, took a different name than you suggested: everything is now `fetch_response_timeout` (`//fetch_response_timeout`, `WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS`, `fetch_response_timeout_secs`). `idle` would have matched the env var but inherits the same inaccuracy as the old message — upload activity doesn't reset it, so it isn't an idle timeout. "Response timeout" says what it actually measures and implies the right thing about the body. Renamed while it's still unreleased, as you say.

**P2 — `process::exit(0)`.** Fair, and green-washing a run is worse than the test is worth. Rather than drop it, I made the peer close after a bounded delay so nothing stays wedged, which let the counterfactual become a normal non-ignored test: it asserts the request is still pending past where the timeout would have fired, then unwinds. It now runs in CI and can actually guard `a_peer_that_never_answers_is_given_up_on` from decaying, which was the thing it couldn't do before.

**P2 — comment volume.** Trimmed. The `runtime.js` block went from ~28 lines to 3, the doc comment on the resolver to 4, and the clamp comment states only the `webidl.converters.long` wrap, once. Dropped "Past the cap, ask for 0 (disabled) instead" — you're right that it described behaviour the code doesn't have.

Also dropped `annotations_still_compose` and `only_the_leading_comment_block_counts` per your test note, and folded the rest into two.

**On your two manual-verification asks** — neither is done, and I'd rather say so than imply otherwise. The tests drive `PrewarmedIsolate` directly, so the job-failure and log-plumbing path around it is still unexercised, and so is `//fetch_response_timeout 0` reaching a real dedicated worker. Both need a running worker rather than the harness. Worth doing before this ships, not something I can claim from here. The `WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS` gap you noted stands too — the `LazyLock` makes it process-global, so a test can't isolate it.

rubenfiszel:
/claude Re-review at 3f97784. The prior review's five findings are all addressed (see the comment above): identifiers and customer attribution removed from the test file, the commit message and the PR body; the error message no longer claims the connection went silent, since the window also covers connect/TLS/upload; everything renamed to fetch_response_timeout for consistency with the env var; the counterfactual test no longer calls process::exit and now runs in CI; comments trimmed. Please focus on whether those are genuinely resolved, and on the JS wrapper's correctness against the deno_fetch/deno_web APIs it depends on.

rubenfiszel:
Both addressed in `6fa9201`.

**`{...init, signal}`.** Real bug, and the sharper half of it is the one I'd missed entirely — I was thinking about `Object.create({method:"POST"})`, which is rare, but `fetch(url, "POST")` going from a loud `TypeError` to a **silent GET** is the version that would actually bite someone. I reproduced the inherited-member case before fixing it (`GET /inherited` where the script asked for POST), so it was not theoretical.

Fixed as suggested: `Object.create(init, { signal: … })`, with a `typeof` guard that hands a non-dictionary straight back to deno_fetch for its own `TypeError`. Two things worth noting from doing it:

- The guard has to sit **before** the timer is armed. My first attempt returned early from the non-object branch after `setTimeout`, which leaked an armed timer holding the event loop open with nothing attached to abort.
- `Object.create` is captured at module scope next to the timer references, since user code shares the isolate and could redefine it before calling `fetch`.
- Checked the one place deno_fetch does use own-keys — `init.headers || ObjectKeys(init).length > 0` at `23_request.js:405`. `signal` is an own property of the inheriting object, so that branch still evaluates the same way it did with the spread.

Both cases now have tests: `an_init_whose_members_are_inherited_is_not_flattened` and `a_non_dictionary_init_still_fails_loudly`.

**The 600s long-poll ceiling.** This is the finding I'm most glad you caught, because it's the kind that surfaces months later as a mystery. Verified it: `execution.rs:275` is `TIMEOUT_WAIT_RESULT.read().await.clone().unwrap_or(600)`, and it holds the request open with no headers for the duration — so a 300s client deadline would kill a legitimate sync run at exactly the point the server considers it healthy. Shipping a client timeout at half the server's own ceiling isn't a documentation problem.

Default raised to **900s**, and rather than only noting it in the PR I put the constraint on `default_fetch_response_timeout_secs` where someone changing it would see it: it must stay above `TIMEOUT_WAIT_RESULT`, and raising one means raising both. Still bounds a hang to 15 minutes against the 7 days it replaces.

Also took the `response_timeout_env.rs` note — the doc comment now says outright that adding a second test there breaks the isolation.

**Still not done, same as before:** the three manual checks. The real-worker log/job-failure path, `//fetch_response_timeout 0` on the dedicated-worker call site, and now your third — a sync execution around ~400s to see the 300/600 mismatch from the other side, though raising the default to 900 is me picking which side to ship rather than measuring it. All three need a running worker.

rubenfiszel:
Fixed in `00a1dc9`, and you were right — I reproduced it before changing anything:

```
Cannot read private member #body from an object whose class did not declare it
    at Init.get body (file:///eval.ts:9:17)
    at Array.RequestInit (ext:deno_webidl/00_webidl.js:816:31)
    at new Request (ext:deno_fetch/23_request.js:325:44)
```

Both of my attempts were wrong in opposite directions. Spreading dropped inherited and non-enumerable members and downgraded a non-object init to a silent GET; `Object.create` fixed those and broke accessor receivers. That pattern is the actual lesson: **there is no safe way to carry a WebIDL dictionary across by hand**, so the fix is to stop carrying it.

`fetch()`'s own first act is `new Request(input, init)`. Doing that in the wrapper hands the caller's init to the same constructor, read exactly as it would be without the wrapper — accessors keep their receiver, inherited and non-enumerable members are read, and a non-dictionary raises deno_fetch's own `TypeError`. Our signal then travels in an init we own. It also deleted code rather than adding it: `req.signal` is deno's resolution of `init.signal` over an input Request's signal, so the hand-rolled version of that rule is gone.

The one thing this does change is that the Request gets constructed twice, so I checked the cost rather than assuming it: `cloneInnerRequest` carries method, headers, redirect mode, `clientRid` and blob entry, and step 41 *proxies* the body rather than consuming it — a static body is a shallow `{body, consumed}` copy sharing its bytes, and a stream gets `pipeThrough(new TransformStream())`, a one-chunk pass-through, not a buffer. `a_request_input_keeps_its_body_and_headers` pins that a `Request` input keeps method, headers and body through the extra construction.

Tests for both of your points: `an_accessor_backed_init_reads_against_its_own_receiver` (private-field getters for both `method` and `body`), and I confirmed it fails with the exact error above when I put `Object.create` back, so it isn't vacuous.

Suite is 11 e2e + 2 annotation + 1 env, green under `--features enterprise` at `--test-threads=10`, plus the 21 smoke tests and `otel_e2e`.

On your pre-merge asks — the streaming one is already covered as an automated test (`a_streaming_body_outliving_the_timeout_is_not_cut_off`: headers inside 2s, body streaming past it, full body asserted). The two real-worker ones are not done and still need a worker: normal and dedicated nativets jobs against a silent endpoint, confirming the `TimeoutError` reaches the job log and the slot takes the next job.

For context, `cargo_test` came back green on the previous head; this push re-runs it.

rubenfiszel:
Fixed in `6ee5650`. You were right on both halves, including that my comment was false — deno_fetch's outer `fetch` is `function fetch(...)`, not async, and 26_fetch.js says why in as many words:

```js
// WPT has a test that aborted fetch is settled in the same tick.
// This means we cannot wrap the promise if it is already settled.
```

So the comment claimed to match something it was actually breaking. Reproduced it before changing anything — with a pre-aborted signal the ordering was `queued-after,fetch` where plain deno_fetch gives `fetch,queued-after`.

The wrapper is now non-async, with an early return that hands deno's settled rejection straight back when the signal is already aborted, and arms no timer there since there is no response to wait for. Construction still has to reject rather than throw, which is what the `async` was buying, so that is now an explicit try/catch returning `Promise.reject` (captured at module scope alongside the timer references). `an_already_aborted_fetch_settles_in_the_same_tick` pins the ordering.

Suite is 12 e2e + 2 annotation + 1 env, green under `--features enterprise` at `--test-threads=10`, plus the 21 smoke tests, `otel_e2e`, and a full-workspace `cargo check --tests`. `cargo_test` was green on `6fa9201`; this push re-runs it.

On your three pre-merge items: the third (a ~400s sync wait against the 900s default) is the one I'd most want run, since raising the default from 300s to 900s was a judgement call about which side of `TIMEOUT_WAIT_RESULT` to sit on rather than something I measured. The other two need a running worker as well. None of them are done and I'm not claiming otherwise.

rubenfiszel:
Both fixed in `de2de43`.

**Primordials.** Fair, and the real flaw was the inconsistency: this file already captures `setTimeout`, `clearTimeout` and `Promise.reject` at module scope for exactly this reason, so an ordinary `.then` lookup was the odd one out. Now captured alongside them.

**Signature.** `init` without a default made `fetch.length` 2 where the standard says 1, and the empty-call branch forwarded two explicit `undefined`s — so deno raised `Invalid URL: 'undefined'` instead of `1 argument required`. Forwarding through `ReflectApply` preserves the count.

`the_wrapper_keeps_fetch_s_own_shape` covers all three. Worth recording how it got there, because my first version was partly worthless:

- Patching `Promise.prototype.then` and then `await`ing under the patch measures the wrong thing — V8 drops its await fast path and treats the promise as a plain thenable, so the test failed for a reason unrelated to the wrapper. The patch has to span only the call, which is where the wrapper reaches for `.then`.
- I then reverted each fix in turn to check the assertions were live. Arity and primordials failed as expected; **the argument-count assertion passed anyway**, because `new Request(undefined, undefined)` throws a `TypeError` too. Asserting the error *type* pinned nothing. It now asserts the message distinguishes `1 argument required` from `Invalid URL: 'undefined'`, and I re-ran the revert to confirm it catches.

Suite is 13 e2e + 2 annotation + 1 env, green under `--features enterprise` at `--test-threads=10`, plus 21 smoke, `otel_e2e`, and a full-workspace `cargo check --tests`. `cargo_test` was green on `6fa9201`.

The manual items are unchanged and still not done — they need a running worker. Of them, the ~400s sync-execution check against the 900s default is the one I'd actually want run, since that number was a judgement call about which side of `TIMEOUT_WAIT_RESULT` to sit on.

rubenfiszel:
✅ Review round clean @ d2c53ce7287bd51a4669d18717b6dafd52d60334 — nit-only verdicts at de2de43eadf1ffcf628a5c64d25a2a896f241ea2; nits addressed in e2908df095 and d2c53ce728. Real normal and native dedicated worker checks now pass, including timeout failures, follow-up jobs, streaming bodies, zero overrides, and a 456-second synchronous child-job wait. Final CI checks are running.

rubenfiszel:
Both P2s are on the branch as of `d2c53ce`.

**Abort intrinsics.** Correct, and the same inconsistency as the `.then` one — I'd hardened the Promise and timer operations and left the abort ones as ordinary lookups. `AbortSignal.any` and `AbortController.prototype.abort` are now captured alongside them.

**The `RequestInit` comment.** This one was my regression: the round-1 review flagged the identical AGENTS.md violation, I fixed it, and it grew back to eight lines when I rewrote the block for the accessor fix. Back to the WebIDL constraint in four.

Verifying the assertions turned up something worth recording. Reverting each fix under its test:

- `AbortSignal.any` capture — caught.
- `AbortController.prototype.abort` capture — **not caught**. `abort` is only reached when the timer *fires*, and the shape test's fetch succeeds in milliseconds, so the patched prototype was restored long before anything touched it. That needed its own test (`the_timer_aborts_through_a_captured_intrinsic`) that patches `abort` out, points at a silent peer, and lets the timeout elapse under the patch. Reverting the capture does fail that one.

Third time in this PR a test passed for a reason unrelated to what it claimed to pin — after the clamp test and the argument-count one. Worth doing routinely, not when convenient.

Also picked up @rubenfiszel's `e2908df` on the `TIMEOUT_WAIT_RESULT` note, which is the P2 from the review above: it now says which knob is which, since `TIMEOUT_WAIT_RESULT` is hot-reloaded instance state a superadmin can raise from the settings page while `WINDMILL_FETCH_RESPONSE_TIMEOUT_SECS` is a process env var behind a `LazyLock` needing a worker restart. Raising only the reachable one re-creates the mismatch a notch higher, which was exactly the point.

Suite is 14 e2e + 2 annotation + 1 env, green under `--features enterprise` at `--test-threads=10`, plus 21 smoke, `otel_e2e`, and a full-workspace `cargo check --tests`.

Manual items unchanged and still not done, in your stated order: the ~400s sync-execution check against the 900s default first, since that number was chosen rather than measured; then the job-log plumbing on a real worker; then `//fetch_response_timeout 0` on the dedicated path.

rubenfiszel:
The worker-level verification is now complete, using an isolated migrated database and backend builds from this branch. The PR description contains the updated checklist.

- Normal Native TypeScript: silent fetch failed after 2,024 ms with `TimeoutError`; the same worker completed the next job. A streaming response completed after 4,013 ms, and a zero override allowed an 8,015 ms response wait.
- Native dedicated worker (`//native`, pinned EE implementation): both the annotation and a 2-second environment default timed out silent requests. A 4-second body streamed to completion. `//fetch_response_timeout 0` overrode that environment default and completed an 8-second response wait. Follow-up jobs succeeded on the same worker. Repeated successfully after rebuilding the abort-helper fix.
- The abort-helper fix was also exercised through a real normal-worker job with `AbortSignal.any` and `AbortController.prototype.abort` replaced: timeout fired after 2,023 ms, then the same worker completed its next job.
- A real synchronous child-job request completed successfully under the default 900-second fetch deadline: the child slept 400 seconds; the outer job waited 456,325 ms including queue time (`01a0815e-32ef-9ea3-37d9-8236e1bbd3fd`).

Local validation also passed: 14 fetch regression cases, 2 annotation cases, 1 environment case, backend builds with `deno_core` and `deno_core,enterprise,private`, and `cargo check --tests --features deno_core,enterprise,private`.

The temporary servers are stopped. Final CI is running on `d2c53ce7287bd51a4669d18717b6dafd52d60334`.

rubenfiszel:
Verification record reconciled: the completed checks are documented in https://github.com/windmill-labs/windmill/pull/11026#issuecomment-5586696087 and in the PR description. That later record supersedes the concurrent 14:22 reply stating they remained outstanding. All listed normal-worker, native dedicated-worker, streaming, zero-override, recovery, and 456-second synchronous-wait checks were executed successfully.

rubenfiszel:
Reconciling the record, since two comments a minute apart contradict each other and the later one is the wrong one.

The manual checks **are** done — the 14:21 comment is accurate. The 14:22 comment's closing paragraph ("manual items unchanged and still not done") was written before that result landed and repeated a status I'd been carrying forward for several rounds without re-checking. Disregard that paragraph; nothing else in it is affected.

So all three outstanding items are now covered, on both normal and native dedicated workers: timeout failures surfacing as `TimeoutError`, the worker taking a follow-up job afterwards, streaming bodies completing past the deadline, `//fetch_response_timeout 0` disabling the limit, and a **456-second synchronous child-job wait** completing.

That last one is the result that matters most here, because it's the only one that tested a number rather than a behaviour. 900s was chosen against `TIMEOUT_WAIT_RESULT`'s 600s default, not measured — and 456s sits above the 300s this PR originally shipped, so it is a case the first version would have killed while the server was still legitimately holding the request open. The raise wasn't hypothetical.

Every review finding across the four rounds is addressed, and the manual verification behind the one judgement call is done.

PR #11029, opened 2026-09-08
fix: chain redeploys onto a retired path's version history

rubenfiszel:
## Summary

`wmill sync push` to a workspace whose scripts had been archived failed with
`A script with same hash (hence same path, description, summary, content) already exists!`
for every unchanged script.

When a sync push applies a file deletion before the corresponding script update, the head at the
target path is archived by the time the update arrives. Parent resolution in
`create_script_internal` only looked for a *live* head (`AND archived = false`), found none, and
left `parent_hash = None`. The version hash covers `parent_hash`, so a parentless deploy of
unchanged content hashes to exactly what the path's own first version hashed to, and the
uniqueness guard rejects it.

The collision needs the path's *first* version to hold the content being pushed, which is why it
is near-deterministic in workspaces where every script has a single version (freshly forked or
CI-provisioned) and sporadic in long-lived ones. Because a remote rejection aborts the whole
push, one affected script fails the entire sync — excluding it just promotes the next one. The
archive has already committed by then, so the failure is sticky: the path stays retired and the
retry fails the same way.

## Changes

- **Deploys at a retired path adopt its lineage tip.** With no live version at the path,
  resolution now takes the most recent version there and adopts it if nothing already descends
  from it. Because the parent is part of the hash, that makes the redeploy unique instead of
  colliding, and it is what the path's history says the deploy is: its next version, not its
  first.

  The "nothing descends from it" condition is load-bearing: a rename leaves its *source* path
  holding a version whose child lives at the destination, and giving that one a second child
  forks a lineage the guard downstream requires to be linear — the exact 400 that #9172 fixed.
  That case still starts a fresh lineage; its regression test is unchanged and still passes.

  The descendant lookup runs on the unscoped pool rather than the caller-scoped transaction. A
  rename can move the child into a folder the caller cannot see, and an RLS-hidden child forks
  the lineage just as hard as a visible one.

- **Reached by any parentless deploy, not only an `auto_parent` one.** The CLI sends no parent at
  all for a path its listing no longer shows, which is where a retried push lands once the
  archive has committed; the UI's new-script flow and MCP do the same. Verified before the
  widening that this door still returned the original 400. Guarded on nothing being live at the
  path, so a parentless deploy onto a *live* one still meets the `(None, Some(_))` arm's
  `Path conflict for {path} with non-archived hash {hash}` refusal rather than silently
  superseding a head the caller never read.

- **The adopted version supplies lineage, not grants.** Chaining lands in the `(Some(_), _)` arm,
  which copies the parent's `extra_perms` into the new row — previously `json!({})` for every
  parentless deploy. Since a retired path can be reused by a *different* script, that would start
  it holding an ACL nobody granted it, including one an admin `delete/h` had purged. Perms are
  reset to `{}` when the parent was adopted rather than named, which preserves today's behaviour
  for parentless deploys exactly; a caller-supplied `parent_hash` still carries them, which is
  what unarchive restores.

- **The candidate is locked before the descendant probe.** A competing deploy chaining onto the
  same candidate takes `FOR UPDATE` on it before inserting, so holding the row is what serializes
  the two. Probing first asks about a child that is still uncommitted, adopts anyway, and leaves
  the pair sharing a `parent_hashes[1]` — a fork the caller-scoped check further down cannot be
  relied on to catch, since the competitor's destination may be RLS-hidden.

- **The linear-lineage guard now runs unscoped, for every parented deploy — not just the ones
  above.** It previously asked through the caller-scoped transaction, but linearity is a property
  of the lineage, not of what a caller may read: a child in a folder they cannot see reads as
  absent and the fork goes in. It is sound only because it already runs while holding the
  parent's `FOR UPDATE`. The conflict message names the clashing hash **only when the caller can
  already read that row** — a hash alone reads a script's content back through `raw/h/{hash}`,
  which authorizes nothing per script, so echoing an invisible one would hand over a way to
  fetch it. Callers who could always see the row get the message unchanged.

- **`delete_after_secs` added to `impl Hash for NewScript`.** It was the one persisted field the
  no-op comparison checked but the version hash did not.

## Known limitation

One branch of the same failure remains: when no candidate qualifies (the rename case above), the
deploy starts a fresh lineage, and a fresh lineage has no parent to vary its hash — so re-pushing
content byte-identical to the path's original first version still 400s. Fixing that means
reviving the archived version rather than deploying a new one, or relaxing the linear-lineage
invariant for a child living at another path; both are larger changes than this.

## Test plan

- [x] `test_auto_parent_adopts_archived_head_instead_of_colliding` — reverting the fallback makes
      it fail with the exact reported 400.
- [x] `test_parentless_redeploy_adopts_the_lineage_but_not_the_grants` — pins the widened entry
      point and the perms reset. Fails on the pre-widening code with the reported 400; with the
      widening but without the perms reset, fails on the grants assertion instead.
- [x] `test_redeploy_chains_past_a_deleted_version` — a `delete/h` tombstone keeps the hash it was
      deployed under even with its content wiped, so excluding soft-deleted versions from the
      candidate query reintroduces the 400. Verified both ways.
- [x] Pre-existing `test_auto_parent_starts_fresh_lineage_when_all_versions_archived` (the #9172
      rename case) still passes, as do `on_behalf_of` (which covers `skip_if_noop`),
      `list_scripts_parent_hash`, and the 8 tests in `windmill-api-integration-tests/tests/scripts.rs`
      (including `test_auto_parent_resolves_parent_hash` and
      `test_update_script_does_not_revive_an_archived_path`).
- [x] Exercised on a dev instance with the real CLI: `sync push` → delete a script file → push
      (the path archives) → restore the file unchanged → push (redeploys, chained onto the retired
      version) → push again (0 changes). No phantom versions.
- [x] Live check of the grants rule: a version carrying `{"u/someone_else": true}`, archived, then
      redeployed parentless — the new live row chains onto it and carries `{}`.
- [x] `delete_after_secs` round-trip: the value persists, changing it cuts a new version, and
      unchanged pushes short-circuit.
- [x] Re-posting an identical create still reports "same hash already exists" rather than the
      linear-lineage message.
- [x] `cargo check` clean.

Not covered by an automated test: the RLS-hidden-descendant path, which needs a second user with
folder-scoped permissions to construct.

## Note on scope

The linked ticket also proposed moving the hash-uniqueness guard above the `skip_if_noop`
short-circuit. That was implemented, then reverted: the collision it was meant to unblock is not
reachable (every insert path stores `parent_hashes[1]` equal to the `parent_hash` that went into
the row's hash, so such a collision is already rejected by the linear-lineage guard, which sits
above the no-op check), and moving it demonstrably regressed the error for a plain duplicate
deploy from "same hash already exists" to "the lineage must be linear". The no-op check is not
involved in this bug at all: with the parent resolved to `None` the deploy takes the
`(None, None)` arm, so `is_noop_deploy_against_parent` never runs and its position cannot matter.

https://claude.ai/code/session_01KMhCSFeWSezY61g6woQiz2


rubenfiszel:
✅ Review round clean @ 7119400d686cd3105e3d380fc8307ce694c3c6cf

PR #11037, opened 2026-09-09
feat: add a dismissible instance-wide announcement banner

rubenfiszel:
## Summary

Adds an instance-wide announcement banner so an operator can broadcast a maintenance window (or an incident) to everyone using the managed cloud. It is a new `instance_banner` global setting: superadmins edit it in Instance settings → Core, every authenticated user can read it, and it renders as a strip above every page until it is turned off.

**Cloud-only, and it costs nothing elsewhere.** The banner component is mounted only when `isCloudHosted()`, so a self-hosted instance makes no request, arms no timer and registers no listener for it, and the setting itself is `cloudonly` so it never appears in their Instance settings. On the cloud it polls every 60s (visible tabs only) and on tab focus, so a session left open all day still hears about the window; a superadmin publishing from the settings drawer sees it immediately.

Viewers can dismiss it, per browser. The dismissal is keyed to a fingerprint of the announcement's content, so editing the message brings it back for everyone who dismissed the previous one; an announcement marked non-dismissible ignores stored dismissals entirely, so escalating an existing notice to mandatory reaches the people who already dismissed it.

## Changes

- `INSTANCE_BANNER_SETTING` and `validate_instance_banner` in `windmill-common/src/global_settings.rs`. Rejects a non-object value, wrong-typed fields, an unknown severity, an over-long message or link label, and any link that is not an absolute http(s) URL — the link becomes the href of an anchor shown to every user, so `javascript:`/`data:` must never reach the browser. An enabled banner with no message is deliberately allowed: it renders as nothing, and every write path is a bulk settings save, so rejecting it would fail an admin's unrelated edits over a half-typed announcement.
- Validation wired into both write paths: the settings pre-write hook (HTTP) and `sync_global_settings_declarative` (the CLI's `instance sync-config` and the operator), which bypasses the hook entirely. Validation stays unconditional — a write can happen on any deployment even though the banner only renders on the cloud.
- `instance_banner` added to the non-superadmin read allowlist in `get_global_setting`. Reads stay behind `ApiAuthed`; writes stay superadmin-only.
- `InstanceBanner.svelte` renders the strip in the content column, in flow so it pushes the page down rather than covering it, using the existing Alert palette for info/warning/critical. Rendered only on the cloud, and skipped for embeds and OAuth callbacks where an instance-wide announcement would land inside someone else's page. Overlapping loads are sequenced by generation so a slow earlier fetch cannot put a retracted announcement back on screen.
- `instanceBanner.ts` holds the pure logic — resolving the stored value into what renders, the visibility rule, and the form validity check shared with the setting's `isValid`. It reads every field defensively: the setting is a raw `global_settings` row, so its shape is only as good as the writer that last touched it.
- `InstanceBannerSetting.svelte` is the editor, with a preview that runs the same resolver the banner does, so what the admin sees is what the instance gets.
- `invalidCategories` now skips `cloudonly` settings off-cloud. A field the build never renders must not be able to block Save — otherwise an invalid config-synced value would leave the category permanently unsaveable with nothing on screen to fix.

## Screenshots

Warning severity with a link, light theme:

pr-banner-light

Same banner, dark theme:

pr-banner-dark

The setting in Instance settings → Core, with the live preview:

pr-banner-setting

## Test plan

Exercised against a running instance, not only type-checked. The cloud-only paths were verified by temporarily forcing `isCloudHosted()` true, then reverting.

- [x] Off-cloud: banner never renders, zero `instance_banner` requests on a page load (network log), and the setting card is absent from Instance settings — same as the pre-existing `cloudonly` `request_size_limit_mb`
- [x] On the cloud path: banner renders, the setting card and its live preview appear
- [x] `javascript:`, `data:`, relative and hostless links rejected on save; unknown severity, wrong-typed fields, over-long message and link label rejected; an enabled banner with an empty message accepted and rendering nothing
- [x] A non-superadmin reads `instance_banner` (200) but is still denied `license_key` (401) and any write (401), and sees the banner in their own workspace
- [x] Banner appears, dismisses, stays dismissed across a full reload, and returns when the message changes
- [x] An already-dismissed announcement flipped to non-dismissible reappears, with no dismiss button
- [x] Overlapping loads: held the first `instance_banner` GET for 5s, published a newer announcement, let a second GET resolve, released the stale response — the newer announcement stayed
- [x] Save is disabled with a `javascript:` link and enabled once it is a valid https URL
- [x] Turning the banner off retracts it live; severity variants and the optional link render correctly in light and dark themes
- [x] A malformed value planted straight into `global_settings` (the config-sync shape) leaves the settings form rendering rather than throwing
- [x] `cargo test -p windmill-common --lib` (555 passed), `cargo test --test instance_config declarative_sync`, `npx vitest run instanceBanner.test.ts`, `npm run check` (no new errors)

### DIFF
--- a/backend/tests/instance_config.rs
+++ b/backend/tests/instance_config.rs
@@ -1485,3 +1485,44 @@ async fn declarative_sync_rejects_an_unusable_webhook_base_url(db: Pool<Postgres
         "the other settings in the same apply must not have been written either"
     );
 }
+
+/// Same contract for the announcement banner: this path owns its validation, and a value
+/// that lands here unchecked reaches every user's browser. A rejected banner must not be
+/// half-applied either.
+#[sqlx::test(fixtures("base"))]
+async fn declarative_sync_rejects_an_unusable_instance_banner(db: Pool<Postgres>) {
+    clear_settings_and_configs(&db).await;
+    let before = count_global_settings(&db).await;
+
+    let mut desired = BTreeMap::new();
+    desired.insert(
+        "base_url".to_string(),
+        serde_json::json!("https://wm.example.com"),
+    );
+    desired.insert(
+        "instance_banner".to_string(),
+        serde_json::json!({ "enabled": true, "message": "down", "link": "javascript:alert(1)" }),
+    );
+
+    let err = windmill_common::instance_config::sync_global_settings_declarative(
+        &db,
+        &BTreeMap::new(),
+        &desired,
+    )
+    .await
+    .expect_err("a javascript: banner link must fail the sync");
+    assert!(
+        err.to_string().contains("instance_banner"),
+        "the error should name the offending setting, got: {err}"
+    );
+
+    assert_eq!(
+        count_global_settings(&db).await,
+        before,
+        "validation must run before anything is applied"
+    );
+    assert!(
+        get_global_setting(&db, "base_url").await.is_none(),
+        "the other settings in the same apply must not have been written either"
+    );
+}
--- a/backend/windmill-api-settings/src/lib.rs
+++ b/backend/windmill-api-settings/src/lib.rs
@@ -59,11 +59,11 @@ use windmill_common::{
         CRITICAL_ALERT_MUTE_UI_SETTING, CUSTOM_TAGS_SETTING, DEFAULT_TAGS_WORKSPACES_SETTING,
         DISABLE_HUB_SETTING, EMAIL_DOMAIN_SETTING, ENV_SETTINGS,
         GITHUB_APP_WEBHOOK_BASE_URL_SETTING, HTTP_ROUTE_WORKSPACED_ROUTE_SETTING,
-        HUB_ACCESSIBLE_URL_SETTING, HUB_BASE_URL_SETTING, MAX_RETENTION_OVERRIDE_WORKSPACES,
-        RETENTION_PERIOD_SECS_OVERRIDES_SETTING, RUFF_CONFIG_SETTING,
-        WORKSPACE_FAIRNESS_DURATION_SECS_SETTING, WORKSPACE_FAIRNESS_ENABLED_SETTING,
-        WORKSPACE_FAIRNESS_MAX_PERCENT_SETTING, WORKSPACE_FAIRNESS_MIN_TOTAL_SETTING,
-        WS_BASE_URL_SETTING,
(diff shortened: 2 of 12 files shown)


rubenfiszel:
Round 1 addressed in f70ddae121:

- **[P1] stale banner responses overwrite newer state** — fixed. `load()` now stamps each call with a generation and only assigns `banner` if it is still the latest. Verified in a browser by holding the first `instance_banner` GET for 5s, publishing a newer announcement, letting a second GET resolve, then releasing the stale response: the newer announcement stays on screen.
- **[P2] Unicode-aware character limits** — fixed. Truncation is now by code point, matching Rust's `chars().count()`, with a test asserting a 500-emoji message the backend accepts survives whole. The fingerprint deliberately keeps hashing the *untruncated* message: it is an identity token for "is this the announcement the viewer dismissed", and two messages differing only past the cap should not collide into one dismissal.

**Not addressing: the missing `feature_usage` proposal.** This is a superadmin-only broadcast toggle, and per `docs/feature-telemetry.md` the bar is "would the answer change a decision" — a count of how many superadmins flipped a maintenance banner would not. Instrumenting it also is not free here: step 1 of the recipe registers the `(feature, kind)` pair in `feature_usage_ee.rs`, which lives in `windmill-ee-private`, so it would turn this into an EE companion PR. I raised the decision to skip it explicitly rather than silently; happy to add it if @rubenfiszel wants the signal.

rubenfiszel:
✅ Review round clean @ f70ddae121240c5d9f9d9edd82639021cedbd50d — nit-only verdicts at that SHA (codex: mergeable with nits, claude: good to merge); the nit (banner link errors not wired into settings validity) is addressed in d7b55e2cb8ef8164905680d0f1a3e8c4dccefda0. Verified in the browser: Save settings is disabled with a `javascript:` link and enabled once it is a valid https URL.

Left in draft on purpose: this widens the non-superadmin read allowlist in `get_global_setting`, which is an authorization surface, so it wants a human look before ready rather than an automatic flip.

rubenfiszel:
✅ Review round clean @ 69edadfb5d2d350a6044d5223a2d8e0169787553

Both reviewers: Good to merge.

Still in draft on purpose — this widens the non-superadmin read allowlist in `get_global_setting`, an authorization surface, so it wants a human look before ready rather than an automatic flip.

PR #11038, opened 2026-09-09
feat: batch chained DDL statements into a single migration

diegoimbert:
## Summary

The datatable DDL guard (SQL REPL and postgres script editor) prompted once per DDL statement, so
a script chaining several schema changes asked the same question N times and produced N separate
migrations. It now prompts once per **run of adjacent DDL statements** and turns that whole run
into a single migration, applied in one `BEGIN; … END;` transaction.

A non-DDL statement breaks the run: anything after it may depend on it, and the statements that
stay behind must keep their relative order — so `CREATE; ALTER; INSERT; CREATE; DROP;` yields two
prompts (the two schema-change runs), not five.

## Changes

- `sqlDdl.ts`: `splitSqlRuns(code)` splits a script into alternating DDL / non-DDL runs.
- `sqlDdl.ts`: `joinSqlStatements(statements)` joins statements back into a runnable block,
  `;`-terminating each. A statement ending in a `--` comment gets its terminator on its own line —
  appended on the same line it would be commented out and the statement would run into the next
  one.
- `DdlMigrationGuard.svelte`: `guard()` iterates runs instead of statements — one prompt per run,
  "Create a migration" prefills the modal with the whole run inside one transaction, "Run anyway"
  keeps the whole run inline, cancel still aborts.
- Prompt copy pluralizes: "Schema changes detected — These N consecutive statements are
  schema-changing (DDL) … Create a single migration for them instead?"
- Comment/doc touch-ups where the guard and its telemetry described a single statement.

No API, schema or public-surface change; the exported `guard()` signature and return contract are
unchanged.

## Screenshots

Running `CREATE TABLE t1 (id int) -- keeps its terminator\n; ALTER TABLE t1 ADD COLUMN name text; INSERT INTO t1 VALUES (1, 'a'); CREATE TABLE t2 (id int); DROP TABLE t2;`
in a datatable SQL REPL — one prompt per DDL run, one migration for the first run, a second prompt
for the run after the `INSERT`:

wm_ddl2_1_prompt1

wm_ddl2_2_migration_modal

wm_ddl2_3_prompt2

## Test plan

- [ ] `npx vitest run src/lib/components/sqlDdl.test.ts` — run grouping, the non-DDL break, and the
      terminator-after-a-line-comment case
- [ ] On a datatable, run the script above in the DB manager REPL: two prompts, each showing its
      two statements `;`-terminated
- [ ] "Create a migration" on the first prompt writes one migration with both statements inside one
      `BEGIN; … END;`; running it creates `t1(id, name)`
- [ ] "Run anyway" on the second prompt leaves `INSERT …; CREATE TABLE t2 …; DROP TABLE t2;` in the
      editor, in the original order, and runs them

https://claude.ai/code/session_01N8jvLAnZGMDQTm3Q8WQ3u6


diegoimbert:
On the P1 about adjacent DDL needing a commit boundary (e.g. `ALTER TYPE … ADD VALUE` followed by a use of the new value): acknowledged and accepted as-is, not fixed.

Proving a run transaction-safe would mean modelling per-statement transaction rules for every dialect, and the failure mode here is cheap and visible: the batched body is shown in the migration editor before anything runs, and a failed run is reverted with the database's own error, so splitting the migration by hand is a one-step fix the user can see. Recorded as a comment on `splitSqlRuns` in 1923fae558 so the constraint is stated where someone would otherwise "fix" the grouping.

PR #11039, opened 2026-09-09
fix: stop a new AI session adopting a legacy sidebar chat

Guilhem-lm:
## Summary

Clicking "Open in AI session" from an editor could open the new session on **someone else's
conversation** — the script in the preview tab was right, but the chat on the left was a
conversation held earlier in the docked sidebar chat.

`ensureChatIdsSeeded` was a one-shot migration from the original sessions PR (#9034): it paired
saved chats carrying no `sessionId` ("untagged") with sessions carrying no `chatId`, **by index in
two independently-sorted lists**, so conversations held before sessions shipped would still appear
under sessions. Two of its assumptions no longer hold:

- **"A session with no chat is an old session."** `openInSession` creates the session and then
  seeds its preview tab, and `setSessionTabs` → `persistTouched` strips `transient` — so by the
  time the sessions page inits the runtime, the brand-new hand-off session looks pre-existing.
  Being prepended to `sessionState.sessions`, it sits at index 0 and is first in line for the
  newest untagged chat, which `initRuntime` then loads into it.
- **"An untagged chat is a leftover from before sessions."** The docked sidebar chat still ships
  behind the beta opt-out, and only the session runtime calls `historyManager.setSessionId` — so
  every conversation held there today is written untagged and keeps feeding the migration.

That same PR had already patched this symptom once, by skipping `transient` sessions; the hand-off
slips past that clause because it persists its session before the seed runs.

Rather than tighten the heuristic again, this deletes it. Sessions and legacy chats are separate
things and pairing them by list position cannot be made correct. Pre-sessions conversations remain
in storage and reachable through the legacy chat's own History list.

## Changes

- Delete `ensureChatIdsSeeded` and its `seedPromise` one-shot from `sessionState.svelte.ts`, plus
  the now-unused `HistoryManager` type import.
- Drop its single call site from `initRuntime` in `sessionRuntime.svelte.ts`. A session still
  assigns and reloads its own `chatId` there, so nothing else changes.
- Update two comments that referenced the deleted migration: a test comment that justified itself
  by citing it (the behaviour it pins — recovery never lands on a persisted session — still holds,
  because such a session can hold a conversation of its own), and `persistTouched`'s note about
  non-touch writers.

**Users who already ran the seeding keep their pairing on disk**, and this deletion does not
repair it: a session that adopted a chat still loads it (`initRuntime`'s `if (session.chatId)`
branch), and because the seeding also stamped `sessionId` on that chat, it no longer appears in the
legacy chat's History. `/clear` in the affected session moves it onto a fresh chat. The deletion
stops new occurrences; it does not restore the pre-migration state.

## Screenshots

Before — hand-off from the script editor lands on a sidebar-chat conversation the user never had
in this session:

grafted-chat-repro

After — same steps, same history: the session opens empty, on its own chat:

after-fix-handoff-empty-chat

## Test plan

Both preconditions matter: the bug needs an untagged chat in history **and** the one-shot unspent
in that page load (it ran at the first session runtime init), which is why visiting `/sessions`
first hides it.

- [ ] Hold a conversation in the docked sidebar chat (switch back to legacy chat), then switch
      sessions back on.
- [ ] Hard-reload a script editor page — don't open `/sessions` after the reload — and click
      "Open in AI session". The session opens empty, with the script in the preview tab.
- [ ] Reopen an existing session that has a conversation: it still loads its own chat.
- [ ] The sidebar conversation is still listed in the legacy chat's History.
- [ ] `npx vitest run src/lib/components/sessions/ src/lib/components/copilot/chat/HistoryManager.test.ts` — 324 pass.

Fixes WIN-2489

### DIFF
--- a/frontend/src/lib/components/sessions/sessionRuntime.svelte.ts
+++ b/frontend/src/lib/components/sessions/sessionRuntime.svelte.ts
@@ -38,7 +38,6 @@ import {
 import {
 	commitSessionWorkspace,
 	deleteSession as deleteSessionState,
-	ensureChatIdsSeeded,
 	getEffectiveWorkspaceId,
 	materializeTransient,
 	sessionState,
@@ -927,7 +926,6 @@ async function initRuntime(runtime: SessionRuntime, session: Session) {
 	// Restore linked files persisted for this session (live handles re-grant on send;
 	// snapshots restore directly). Non-transient sessions persist immediately.
 	await manager.attachedFiles.restore(session.id, !session.transient)
-	await ensureChatIdsSeeded(manager.historyManager)
 
 	// Keep the session record's chatId following the manager's active chat: a
 	// "/clear" rotation or a history switch would otherwise leave it pointing at
--- a/frontend/src/lib/components/sessions/sessionState.svelte.ts
+++ b/frontend/src/lib/components/sessions/sessionState.svelte.ts
@@ -38,7 +38,6 @@ export function syncWorkspaceTo(workspaceId: string | undefined): void {
 }
 import { WorkspaceService } from '$lib/gen'
 import { sendUserToast } from '$lib/toast'
-import type HistoryManager from '$lib/components/copilot/chat/HistoryManager.svelte'
 import { onUserChange } from '$lib/userScopedStorage'
 
 // A destination the session preview can open as an editor: a workspace item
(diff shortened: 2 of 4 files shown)


Guilhem-lm:
✅ Review round clean @ 026dacfbd8e8e0fb9716c9d78066b246c6eaff6a

